PR #5445, opened 2021-05-10
docs: fix slider example indentation [skip ci]

elgiano:
## Purpose and Motivation

One example had excessive indentation, which prevented code-block execution from the Help Browser

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Slider.schelp
+++ b/HelpSource/Classes/Slider.schelp
@@ -245,32 +245,31 @@ a.action = nil;
 subsection:: Use Slider for triggering sounds
 code::
 
-	(
-		s.waitForBoot({
-			
-			SynthDef(\pluck, { |out, freq=55|
-				Out.ar(out,
-					Pluck.ar(WhiteNoise.ar(0.06),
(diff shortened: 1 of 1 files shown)


telephon:
what has changed?

elgiano:
Just indentation, because with wrong indentation the code doesn't execute when pressing Ctrl+Enter on the Help Browser

telephon (approved):

telephon:
> with wrong indentation the code doesn't execute when pressing Ctrl+Enter on the Help Browser

This seems to be a bug in the help browser which would be nice to get rid of.

elgiano:
> This seems to be a bug in the help browser which would be nice to get rid of.
In principle I agree with you, but the current behavior is actually consistent with how the Document editor works. In this case the problem was that a code block is not recognized as a block because there is whitespace before the opening `(`. And this works the same in scide's Documents, opening parens for blocks need to be the first character on a line, otherwise the block is not recognized.

PR #5456, opened 2021-05-23
Fix data race in CondVar timeout

patrickdupuis:
## Purpose and Motivation

This PR fixes a data race that can happen when a `CondVar` is signaled immediately as it is timing out. The problem is easy to reproduce:

```supercollider
(
var cond = CondVar();
var test = false;
var waitTime = 1.0;

fork {
    waitTime.wait;
    cond.signalOne;
};

fork {
    cond.waitFor(waitTime, { test });
    "waitFor timed out".postln;
};
)
```

An error is thrown because the time out mechanism is trying to wake a thread that has already been woken by the signal. `this.prRemoveWaitingThread(waitingThread);` returns `nil` in this case, so `this.prWakeThread(wokenThread);` throws an error having received a `nil` value.

It doesn't look like this change has any adverse effects. `TestCondVar.sc` is still passing. I tried to write a test for this issue, but couldn't figure out a way to write one using `assertNoException`. If anyone knows how I can write a test for this, please let me know and I'll add it to this PR.

## Types of changes

- Bug fix

## To-do list

- [X] Code is tested
- [X] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Core/CondVar.sc
+++ b/SCClassLibrary/Common/Core/CondVar.sc
@@ -86,7 +86,7 @@ CondVar {
 			// if wokenThread is nil, we asssume a signal happened
 			// immediately prior to timeout and the thread is already awake
 			// no call to prWakeThread happens in this case
-			wokenThread !? { this.prWakeThread(wokenThread) };
+			if(wokenThread.notNil) { this.prWakeThread(wokenThread) };
 		};
 
 		timeoutThread.play(thisThread.clock);
--- a/SCClassLibrary/Common/Core/CondVar.sc
+++ b/SCClassLibrary/Common/Core/CondVar.sc
@@ -79,11 +79,14 @@ CondVar {
 
 			didNotTimeout = false;
 
-			// wokenThread must be non-nil. If our waiting thread is signaled before
-			// this timeout fires, the timeout routine will be stopped after prWait
-			// below and we won't get here.
+			// If our waiting thread is signaled before this timeout fires
+			// the timeout routine will be stopped after prWait below
+			// and we won't get here.
 			wokenThread = this.prRemoveWaitingThread(waitingThread);
-			this.prWakeThread(wokenThread);
+			// if wokenThread is nil, we asssume a signal happened
+			// immediately prior to timeout and the thread is already awake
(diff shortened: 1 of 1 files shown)


joshpar (approved):

SCClassLibrary/Common/Core/CondVar.sc
```diff
@@ -79,11 +79,14 @@ CondVar {
...
-			// below and we won't get here.
+			// If our waiting thread is signaled before this timeout fires
+			// the timeout routine will be stopped after prWait below
+			// and we won't get here.
 			wokenThread = this.prRemoveWaitingThread(waitingThread);
-			this.prWakeThread(wokenThread);
+			// if wokenThread is nil, we asssume a signal happened
+			// immediately prior to timeout and the thread is already awake
+			// no call to prWakeThread happens in this case
+			wokenThread !? { this.prWakeThread(wokenThread) };
```
jamshark70, line 89: I'll offer this only as a comment, not a "requested change," but... as I posted recently on the forum, I find more and more that it improves readability and future maintainability by spelling out the logic rather than using funky operators. In my opinion, I'd like to see funky operators used less and named methods more.

If I were making this change, I would write:

```
if(wokenThread.notNil) { this.prWakeThread(wokenThread) };
```

Incidentally:

```
{ a !? { b } }.def.dumpByteCodes
BYTECODES: (7)
  0   12       PushInstVar 'a'
  1   8F 1B 00 01 ControlOpcode 1  (5)
  5   13       PushInstVar 'b'
  6   F2       BlockReturn

{ if(a.notNil) { b } }.def.dumpByteCodes
BYTECODES: (7)
  0   12       PushInstVar 'a'
  1   8F 1B 00 01 ControlOpcode 1  (5)
  5   13       PushInstVar 'b'
  6   F2       BlockReturn
```

So the more explicit style, in this case, has no performance cost (identical byte codes), but would have benefits for future readers -- not much of a downside.

patrickdupuis: I'll re-write it as an if statement for clarity :+1:

patrickdupuis:
@joshpar @dyfer this is ready to be merged

Issue #5360, opened 2021-02-16
Some tests fail when run in qpm


dyfer:
## Environment

* SuperCollider version: current develop
* Operating system: Linux, macOS
* Other details (Qt version, audio driver, etc.): test suite run with qpm

## Steps to reproduce

Instructions from https://github.com/supercollider/supercollider/pull/5332#issuecomment-776672228

- install qpm with `pip2`
  - `pip2 install git+https://github.com/supercollider/qpm.git@topic/xvfb-off`
  - note that we are temporarily using the `topic/xvfb-off` branch of qpm
- install API quark 
  - with `python2 /usr/bin/qpm quark checkout API -l <some path>`
    - default checkout path only makes sense on macOS (`~/Library/Application Support`)
  - OR by checking it out `git clone --depth=1 https://github.com/supercollider-quarks/API`
- On Linux, add path with scscyth to the PATH, e.g.
`export PATH=/path/to/supercollider/build/Install/bin:$PATH`
- create a copy of the json file (it will be updated with test results)
`cp /path/to/supercollider/testsuite/scripts/gha_test_run_proto.json /path/to/supercollider/testsuite/scripts/run/gha_test_run.json`
- run with `python2 /usr/bin/qpm test.run -l <json copy> --path <sclang> --include <API path> <testsuite/classlibrary path>`

## Expected vs. actual behavior

Some tests fail only when run in qpm. See #5332 for more information.

When fixing the tests, please remove appropriate entries in [gha_test_run_proto.json](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/gha_test_run_proto.json)

Let's use the lists below to keep track of fixed tests.

- [ ] TestBus.test_get: UnitTest sometimes fails when run in qpm
- [ ] TestCondition.*: UnitTest fails when run in qpm
- [ ] TestDocumentEnvironment.*: UnitTest creates issues when run in qpm
- [ ] TestDoneActions.test_freeSelfAndResumeNext: UnitTest fails when run in qpm
- [ ] TestEnvGen_server.test_shapeHold_setEndValue: UnitTest fails when run in qpm on Linux
- [ ] TestEnvGen_server.test_trig_gate_nonsustain_env: UnitTest fails on supernova in qpm on Linux
- [ ] TestEnvironmentDocument.*: UnitTest fails when run in qpm
- [ ] TestLinkClock.*: UnitTest has unknown issues when run in qpm
[...]

Issue #5359, opened 2021-02-16
Some tests have bugs


dyfer:
## Environment

* SuperCollider version: current develop
* Operating system: Linux, macOS
* Other details (Qt version, audio driver, etc.): tests running locally or in CI

## Steps to reproduce
- add `supercollider/testsuite/classlibrary` to the interpreter's include paths
- reboot interpreter
- run designated test, e.g. `TestCoreUGens.run`

## Expected vs. actual behavior

After implementing our full test suite in the CI, it turned out that there's a number of tests that are currently broken, either due to bugs in tests, or due to test themselves failing. See #5332 for more information.

When fixing the tests, please remove appropriate entries in [gha_test_run_proto.json](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/gha_test_run_proto.json)

Tests with bugs:
- [ ] TestCoreUGens
- [ ] TestDocument
- [ ] TestEnvGen_server.test_zero_gate_n_off_not_sent
- [ ] TestFilterUGens
- [ ] TestFilterUGens
- [ ] TestMixedBundleTester
- [ ] TestNode_Server
- [ ] TestProxyNodeMap
- [ ] TestServer_boot

Tests failing (might be due to error in the test, or the test catching an actual bug):
- [ ] TestNodeProxy.test_copy_nodeMapIsCopied
- [ ] TestNodeProxyBusMapping:test_audiorate_mapping
- [ ] TestNodeProxyBusMapping:test_audiorate_mapping_elasticity
- [ ] TestNodeProxy_Server.test_schedAfterFade_notBeforeQuant
- [ ] TestNodeProxy_Server.test_synthDef_isReleased_afterFree
- [ ] TestPanAz
- [ ] TestSCDoc.test_helpSourceDirs_includedExtensions
- [ ] TestTempoClock.test_nextTimeOnGrid_okAfterMeterChange
- [ ] TestUnitTest.test_setUpClass_already_happened

patrickdupuis:
When working on fixing unittests, is it a good idea to have SC built `Debug`? or, does it not matter?

dyfer:
I don't know...

patrickdupuis:
ok, I'm currently going through the entire collection of tests in the test suite to see whick ones fail on my system. Hopefully, some of them are easy fixes.

dyfer:
All tests in this PR should be relatively easy to fix, unless they actually catch a bug that's more involved. Tests from #5360 will be harder to debug...

[...]

PR #5457, opened 2021-05-24
Use CondVar instead of UnitTest.wait in tests

### LINKED ISSUE #5366: Some tests broke due to a change in UnitTest.wait
## Environment

* SuperCollider version: current `develop`

## Issue

A handful of UnitTests broke when `UnitTest.wait` changed in https://github.com/supercollider/supercollider/commit/2aa3374e1e52b7db3f7708498810dc6032a1d082

This breakage is pretty easy to spot. You'll see an error like:
```
^^ ERROR: Message 'test' not understood.
Perhaps you misspelled 'dist', or meant to call 'test' on another receiver?
```

These tests need to be modified so that calls to `UnitTest.wait` are passed a Condition instead of a Function or Boolean.

- TestFilterUGens
- TestNode_Server
- TestServer_boot
- TestMixedBundleTester
- TestCoreUGens
- TestDocument

See discussions in #4659 and #5363
Related issue #5359

patrickdupuis:
## Purpose and Motivation

Fixes #5366

This PR removes all but one use of `UnitTest.wait` inside our test suite (the exception being testing that method itself). `CondVar` is used as a preferred alternative. I've updated the documentation, and added comments and notes pointing users towards `CondVar`.

I've also re-written the implementation of `UnitTest.wait`. This method no longer makes use of `Condition` as this Class is likely going to be deprecated in the future (or at least no longer be used inside the SC class library).

## Types of changes

- Documentation
- Bug fix
- Breaking change (maybe?)

## To-do list

- [X] Code is tested
- [X] All tests are passing
- [X] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/UnitTesting/UnitTest.sc
+++ b/SCClassLibrary/Common/UnitTesting/UnitTest.sc
@@ -232,21 +232,22 @@ UnitTest {
 		^boolean
 	}
 
-	// waits for condition with a maxTime limit
-	// if time expires, the test is a failure
-	wait { |condition, failureMessage, maxTime = 10.0|
-		var dt = 0.05;
-		var limit = max(1, maxTime / dt);
+	// this method should be avoided if at all possible
+	// it's better to use CondVar directly in tests instead
+	wait { |predicate, failureMessage = "", maxTime = 10.0|
+		var condvar = CondVar();
+		var waitDur = 0.1;
+		var limit = max(1.0, maxTime / waitDur);
 
 		while {
-			condition.test.not and: { limit >= 0 }
+			(limit >= 0) and: { condvar.waitFor(waitDur, predicate).not }
 		} {
 			limit = limit - 1;
-			dt.wait;
 		};
 
-		if(limit < 0 and: failureMessage.notNil) {
-			this.failed(currentMethod,failureMessage)
+		// consider test failed if limit is surpassed
+		if(limit < 0) {
+			this.failed(currentMethod, failureMessage)
 		}
 	}
 
--- a/testsuite/scripts/gha_test_run_proto.json
+++ b/testsuite/scripts/gha_test_run_proto.json
@@ -8,7 +8,13 @@
             "suite":"TestBus",
             "test":"test_get",
             "skip":true,
-            "skipReason":"UnitTest sometimes fails when run in qpm (FIXME)"
+            "skipReason":"UnitTest fails when run in qpm (FIXME)"
+        },
+        {
+            "suite":"TestBus",
+            "test":"test_getn",
+            "skip":true,
+            "skipReason":"UnitTest fails when run in qpm (FIXME)"
         },
         {
             "suite":"TestCondition",
@@ -18,7 +24,7 @@
         },
         {
             "suite":"TestCoreUGens",
-            "test":"*",
+            "test":"test_ugen_generator_equivalences",
             "skip":true,
             "skipReason":"UnitTest has bugs (FIXME)"
         },
@@ -26,7 +32,7 @@
             "suite":"TestDocument",
             "test":"*",
             "skip":true,
-            "skipReason":"UnitTest has bugs (FIXME)"
+            "skipReason":"Document tests are only valid inside the IDE"
         },
         {
             "suite":"TestDocumentEnvironment",
@@ -59,34 +65,10 @@
             "skipReason":"UnitTest fails on supernova in qpm on Linux (FIXME)"
         },
         {
-            "suite":"TestEnvGen_server",
-            "test":"test_shapeHold_setEndValue",
-            "skip":true,
-            "skipReason":"UnitTest fails on when run in qpm on Linux (FIXME)"
-        },
-        {
-            "suite":"TestFilterUGens",
-            "test":"*",
-            "skip":true,
-            "skipReason":"UnitTest has bugs (FIXME)"
-        },
-        {
-            "suite":"TestFilterUGens",
-            "test":"*",
-            "skip":true,
-            "skipReason":"UnitTest has bugs (FIXME)"
-        },
-        {
             "suite":"TestLinkClock",
             "test":"*",
             "skip":true,
             "skipReason":"UnitTest has unknown issues when run in qpm (FIXME)"
-        },
-        {
-            "suite":"TestMixedBundleTester",
-            "test":"*",
-            "skip":true,
-            "skipReason":"UnitTest has bugs (FIXME)"
         },
         {
             "suite":"TestNodeProxy",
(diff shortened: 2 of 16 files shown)


dyfer:
Thanks @patrickdupuis !

Could you remove the "skip" entries for the fixed tests from the `gha_run_test_proto.json`?
E.g. remove the following: https://github.com/supercollider/supercollider/blob/6c49680cbb2e8157dccdaf8e08354eb6d6c64953/testsuite/scripts/gha_test_run_proto.json#L19-L24
as well as other fixed tests. Once you do, these tests will already run in CI in this PR.

SCClassLibrary/Common/UnitTesting/UnitTest.sc
```diff
@@ -232,21 +232,22 @@ UnitTest {
...
+	// this method should be avoided if at all possible
+	// it's better to use CondVar directly in tests instead
+	wait { |predicate, failureMessage = "", maxTime = 10.0|
+		var condvar = CondVar();
+		var waitDur = 0.1;
+		var limit = max(1.0, maxTime / waitDur);
 
 		while {
-			condition.test.not and: { limit >= 0 }
+			(limit >= 0) and: condvar.waitFor(waitDur, predicate).not
```
telephon, line 243: better replace by:
```
(limit >= 0) and: { condvar.waitFor(waitDur, predicate).not }
```
this will result in inlined code.

patrickdupuis: Interesting, thanks! I'll re-write it that way.

patrickdupuis:
@dyfer I've updated the list of skipped tests in the CI. Let's see what happens in GA :)

dyfer:
> @dyfer I've updated the list of skipped tests in the CI. Let's see what happens in GA :)

I don't think we can run `TestDocument` in the current configuration since we are not running the IDE, so these should stay skipped. The only other test that's failing is `TestUnitTest:test_setUpClass_already_happened`

SCClassLibrary/Common/UnitTesting/UnitTest.sc
telephon: thanks!

patrickdupuis:
I've added those tests back to the skipped list.

testsuite/scripts/gha_test_run_proto.json
```diff
@@ -8,7 +8,13 @@
             "suite":"TestBus",
             "test":"test_get",
             "skip":true,
-            "skipReason":"UnitTest sometimes fails when run in qpm (FIXME)"
+            "skipReason":"UnitTest fails when run in qpm (FIXME)"
+        },
+        {
+            "suite":"TestBus",
```
dyfer, line 14: `TestBus` are weird. They show as if they fail, but they actually don't register as a failure in the end. So I don't think they need to be skipped (as they weren't before), this is something to address in the testing framework itself...

patrickdupuis: I was seeing a failure on macOS (or linux?) earlier. I think the `test_get` and `test_getn` are ones we should skip just to be safe.

testsuite/scripts/gha_test_run_proto.json
```diff
@@ -176,13 +146,19 @@
...
             "skip":true,
-            "skipReason":"UnitTest fails (FIXME)"
+            "skipReason":"test fails when run the first time after interpreter boot"
         },
         {
             "suite":"TestWebView",
-            "test":"*",
+            "test":"test_findText_found",
+            "skip":true,
+            "skipReason":"test likely doesn't work inside the CI"
```
dyfer, line 155: Is this comment coming from your tests? We should be able to run GUI classes fine, as long as they are not IDE-only. IIRC my experience was that these tests _sometimes_ failed, so again something between the testing framework, CI environment and the test itself.

patrickdupuis: Yeah, this is my comment. The 2 "findText" tests are commented in the source. It says they only work if the window is being displayed.

testsuite/scripts/gha_test_run_proto.json
dyfer: My personal preference would be not to touch tests unrelated to the CondVar change. `TestBus` passes at least sometime:
https://github.com/supercollider/supercollider/runs/2651269477?check_suite_focus=true#step:8:14863
https://github.com/supercollider/supercollider/runs/2651269468?check_suite_focus=true#step:9:15393

But if you want really want to keep this change, this fine too.

testsuite/scripts/gha_test_run_proto.json
dyfer: But we can "display" windows in CI, can't we? We do run a virtual framebuffer device (on Linux); on macOS the graphic server is always available AFAIU. The test is using `Window` so not an IDE class.

testsuite/scripts/gha_test_run_proto.json
patrickdupuis: For this PR, I'd like to get to a point where are running tests pass everytime(-ish). There's more work needed to be done in follow up PRs.

testsuite/scripts/gha_test_run_proto.json
patrickdupuis: When I looked at the logs, one the "findText" tests was passing and the other not. When I looked at the source, it wasn't clear to me why either would fail. That's why I marked them both as skipped. We can work on this issue some other time.

patrickdupuis:
Should we re-run the CI a few times just to make sure everything is passing?

testsuite/scripts/gha_test_run_proto.json
dyfer: I agree that they should be skipped, but I was commenting on your change to the reason it's skipped - I thought the original comment was more accurate. Maybe not a big deal either way.

patrickdupuis: Ok, yeah I agree. All of the comments here should be re-evaluated. I'll change it back when I squash the commits to clean up the history.

patrickdupuis:
I cleaned up the commit history. Let's run the CI one last time before merging.

patrickdupuis:
@dyfer the test that's failing for Linux hasn't happened before. I'd rather just ignore it and merge the PR. What do you think?

dyfer:
> @dyfer the test that's failing for Linux hasn't happened before. I'd rather just ignore it and merge the PR. What do you think?

I'd be fine ignoring it, but I have seen `TestSoundFile_Server:test_cue_playNow` failing before, albeit randomly. Up to you whether you want to add skipping it, or merging this as is.

patrickdupuis:
That test isn't touched by any changes made in this PR, so I say we ignore it.

dyfer (approved):
Thanks!

dyfer:
I've updated lists of tests in #5359 and #5360

PR #4598, opened 2019-09-27
Classlib: Buffer:write should use Buffer:writeMsg

jamshark70:
## Purpose and Motivation

I got really rather cross just now when I was running some code against supernova and a `Buffer:write` failed because the buffer's numFrames was a float rather than an integer.

Upon investigation, I found that `Buffer:writeMsg` correctly casts float to integer:

```
		^["/b_write", bufnum, path,
				headerFormat, sampleFormat, numFrames.asInteger, startFrame.asInteger,
				leaveOpen.binaryValue, completionMessage.value(this)];
```

... but `Buffer:write` does not follow the convention established in other buffer methods of calling the corresponding `*Msg` method to get the message to send. `write`'s duplicated message code *omits* the cast, so it may fail against supernova:

```
		server.listSendMsg(
			[\b_write, bufnum, path,
				headerFormat, sampleFormat, numFrames, startFrame,
				leaveOpen.binaryValue, completionMessage.value(this)]
			)
```

The PR replaces the copy/pasted message code with a call to writeMsg.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [-] Updated documentation (no need)
- [x] This PR is ready for review

PR #5460, opened 2021-05-24
Fix bad argTemplate in Buffer.getn

stevearc:
## Purpose and Motivation

There is a bug in `Buffer.getn` that causes it to not call the `action` if the index passed in is a float. This happens because the OSC message generated casts the value to an integer, but the `argTemplate` does not.

#4598 did a large pass over the Buffer methods to try to make them all properly cast to integer in the right places, but it appears that this location was missed.

To repro the issue (and verify that it is fixed):
```sc
(
s.boot;
s.waitForBoot({
  b = Buffer.alloc(s, 10);
  s.sync;
  b.getn(0.0, 10, {|msg| msg.postln}); // Doesn't print anything without this fix
});
)
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Control/Buffer.sc
+++ b/SCClassLibrary/Common/Control/Buffer.sc
@@ -455,7 +455,7 @@ Buffer {
 			// [/b_setn, bufnum, starting index, length, ...sample values].
 			// We want the sample values, which start at index 4.
 			action.value(message[4..]);
-		}, \b_setn, server.addr, argTemplate: [bufnum, index]).oneShot;
+		}, \b_setn, server.addr, argTemplate: [bufnum, index.asInteger]).oneShot;
 		server.listSendMsg(msg)
 	}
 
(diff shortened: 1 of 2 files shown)


joshpar (changes requested):
More of a light request than a hard one. This is a great catch - thanks! Can you please include a couple of tests for this? One for passing in a float and another for an integer, and finally some garbage like a string that we would expect to fail?

stevearc:
@joshpar Thanks for the review! I've added a regression test to `TestBuffer_server` to handle the float case. There's a test right above it that handles the integer case. I'm not convinced that we should add a test for passing in garbage, since the behavior in that case is undefined. Since there's no explicit contract it feels wrong for a test to be enforcing anything, but I could be convinced otherwise.

joshpar (approved):

joshpar:
Thank you!

PR #5461, opened 2021-05-24
Fix test setUpClass and tearDownClass not being called

patrickdupuis:
## Purpose and Motivation

This PR fixes an issue where a test class' `setUpClass` and `tearDownClass` functions aren't being called. To reproduce the issue, try running the following:

```supercollider
// reboot the interpreter before running this
UnitTest.runTest("TestUnitTest:test_setUpClass_already_happened")
```

`setUpClass` and `tearDownClass` are being run on `UnitTest` not `TestUnitTest`. To fix this, I've called `prRunWithinSetUpClass` on `class` instead of `this` in a couple places inside `UnitTest`.

This seems to work, however, I'd appreciate having another pair of eyes on this. I can't tell if this change will have any unforeseen side-effects, or if the same change needs to be made elsewhere.

## Types of changes

- Bug fix

## To-do list

- [X] Code is tested
- [X] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -494,7 +494,7 @@ jobs:
       SCLANG: ${{ github.workspace }}/${{ matrix.sclang }}
       SCRIPT_PROTO: ${{ github.workspace }}/testsuite/scripts/gha_test_run_proto.json
       SCRIPT_RUN: ${{ github.workspace }}/testsuite/scripts/run/gha_test_run.json
-      QPM_URL: git+https://github.com/supercollider/qpm.git@topic/xvfb-off
+      QPM_URL: git+https://github.com/supercollider/qpm.git@topic/testclass-setup-teardown
     steps:
       - uses: actions/checkout@v2
         with:
--- a/SCClassLibrary/Common/UnitTesting/UnitTest.sc
+++ b/SCClassLibrary/Common/UnitTesting/UnitTest.sc
@@ -14,8 +14,8 @@ UnitTest {
 			var classkey = class.asString[4..]; // drop Meta_
 			var methtests = class.findTestMethods.collectAs({ | method |
 				method.name.asString -> {
-					this.prRunWithinSetUpClass {
-						class.new.runTestMethod(method)
+					class.prRunWithinSetUpClass {
+						class.new.runTestMethod(method);
 					}
 				}
 			}, Dictionary);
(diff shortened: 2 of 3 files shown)


patrickdupuis:
#5359

patrickdupuis:
There's an issue here. It's not so much that the test fails, but how it is run by qpm. I'm looking into it.

patrickdupuis:
As we can see [here](https://github.com/supercollider/qpm/blob/d3f72894e289744f01f3c098ab0a474d5190315e/qpm/scscripts/test_runner.scd#L169), qpm runs its individual test cases by calling `runTestMethod`. This method expects `setUpClass` and `tearDownClass` to be run manually. It says so in the comment [here](https://github.com/supercollider/supercollider/blob/a2347ebfd966497dc3f667dcfa3390581c0cf389/SCClassLibrary/Common/UnitTesting/UnitTest.sc#L80).

So, this means that SC's test runner is not compatible with the `setUpClass` and `tearDownClass` features that we added to UnitTest recently :( I'm going to see if there's an easy solution. Maybe it's as simple as calling those methods in the qpm test_runner script?

patrickdupuis:
@telephon If I recall correctly, you worked on this class setup/tearDown. Is there a reason why `runTestMethod` can't call these itself? Something like:

```supercollider

	// run a single test method of this class
	// this assumes that setUpClass has been called
	// and that tearDownClass is called afterwards
	runTestMethod { | method, report = true |
		this.class.forkIfNeeded {
			this.class.setUpClass;
			this.setUp;
			currentMethod = method;
			this.perform(method.name);
			this.tearDown;
			this.class.tearDownClass;
			if(report) { this.class.report };
		}
	}
```

patrickdupuis:
It should be possible to call the class setUp/tearDown in qpm. This PR is now waiting on fixes to qpm.

patrickdupuis:
The issue on the qpm side has been tentatively fixed. This is ready for review.

dyfer:
I'd be fine with merging this and seeing if there are any issues in the longer run. Looks fine on the face of it.

Do you think it would be worthwhile to add a comment around `class.prRunWithinSetUpClass` that either briefly explains the change or links this issue? I know that pollutes the source, but I feel it might be helpful to have a visible trace of such changes, at least for some time.

patrickdupuis:
I'd merge it as is. I don't think we need to comment anything. The two methods where `this` became `class` are more consistently written and easier to understand now. Other places where `this` is used are ok since those methods are safe to call on `UnitTest` or any of its subclasses (`run` method for example). There isn't a similar mix of using `this` and `class` in those methods.

dyfer (approved):
Thanks!

PR #5439, opened 2021-05-03
GHA: run hdiutil again if it fails

dyfer:
## Purpose and Motivation

We are getting random build failures on macOS while using the `hdiutil` command to create the archive, see https://github.com/supercollider/supercollider/runs/2495736121?check_suite_focus=true#step:14:15 

I couldn't find the reason of these failures... so for now I propose that we just run the command again in case it fails. 

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

Issue #5278, opened 2020-12-07
Improve macOS installation instructions clearer to avoid translocation


brianlheim:
## Environment

* SuperCollider version: i believe this affects all recent versions
* Operating system: macOS
* Other details (Qt version, audio driver, etc.): N/A

## Description

i've noticed that a lot of people are noticing and/or complaining about [app translocation](https://lapcatsoftware.com/articles/app-translocation.html) on macOS. it's easy to see why this happens -- the official release .zip decompresses into a folder containing the .app and a set of examples, so many people assume the right thing to do is drag the entire folder into `/Applications`. however, this doesn't remove the quarantine attribute, because the .app itself was not dragged into the folder. as far as i can tell, the .app itself must be moved into `/Applications` to remove the quarantine attribute. the other way i know of to remove this attribute is to use the `xattr` command line tool which we should of course not require people to do themselves.

we should make things easier for people by communicating the correct installation steps. i'm not sure what the best thing would be, mainly just wanted to log this issue so we can be sure to address it for 3.12.0. some ideas off the top of my head are (1) better documentation, (2) releasing an installer (.dmg or something else), (3) only releasing a zipped .app.

dyfer:
A common way to do this is to create a .dmg image (without making an installer) and providing shortcut to `/Applications` in the .dmg, possibly with a background image that hints the user to drag the app to `/Applications`. I've seen this on multiple occasions. Example:
![Screen Shot 2020-12-09 at 9 20 33 PM](https://user-images.githubusercontent.com/1589177/101724967-907a3200-3a64-11eb-962a-2e4ce82a1702.png)

PR #5462, opened 2021-05-25
DMG: add background

dyfer:
## Purpose and Motivation

Continuation of the work addressing #5278

In the pre-github times (SC 3.6) supercollider was distributed as a DMG with a background image. I've updated our DMG creation, adding the previously used background image (modified to accommodate larger windows size with more icons). I was not trying to change the original graphic design.

I've included the GIMP file used to resize the original image, hoping it will make future changes easy.

What do you think?

Screen Shot 2021-05-24 at 17 53 42

I've also changed the dmg's compression to zlib (from bzip2), which seems to be much faster to both create and mount, at the cost of being ~20% larger. **Please let me know what you think about the archive size increase.**

EDIT: I've also partially reverted #5439. It seems that `hdiutil`'s failure happened after the image was created, when trying to unmount it, so I think we can safely ignore it, instead of running hdiutil again.

## Types of changes

- New feature (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer:
I downloaded it on another machine and can confirm that it looks as expected. Here's the [build artifact](https://github.com/supercollider/supercollider/suites/2820410065/artifacts/62803441).

Any and all feedback welcome!

joshpar (approved):

PR #5463, opened 2021-05-26
class library: add tidy as a UGen method

### LINKED ISSUE #5459: BPF and BBandPass blowing unrecoverably
## Environment

* Operating system: MacOS

## Steps to reproduce

```supercollider
//blow the filter with dumb values (they don't need to be that stupid but that is the best way to reproduce
{BBandPass.ar(WhiteNoise.ar(),0,0)}.play
//stop the synth
//play something
{PinkNoise.ar(1)}.scope
```

We get the synth to work visually in the scope (it wiggles) but no audio is coming out of the left channel anymore. The only way to recover the blown audio channel is to restart the synth.

It does the same with BPF.

telephon:
Fixes #5459

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/UGen.schelp
+++ b/HelpSource/Classes/UGen.schelp
@@ -330,6 +330,10 @@ Limits the receiver range to one of the four clip modes, see code::linlin:: abov
 method:: checkBadValues
 Wraps the receiver in a link::Classes/CheckBadValues:: UGen with the corresponding code::id:: and code::post:: flag.
 
+method:: tidy
+Remove any non-standdard output from the signal, such as nan or inf. Wraps the receiver in a link::Classes/Sanitize:: UGen.
+
+
 method:: if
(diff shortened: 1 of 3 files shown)


muellmusik:
Hi Julian, I'm not sure I understand the need for this? It seems just a slightly more concise version of sanitize?

telephon:
Yes, it is. I thought that that was it when you said:
>Ah, then I'd definitely prefer to add tidy.

muellmusik:
I'm sorry Julian! I meant to say prefer _not_ to! End of term and I'm a bit knackered. 

I generally think we shouldn't add synonyms, especially when there's no new functionality or advantage. But I'm very sorry if you've done this work because of my mistyping! I certainly won't object if others want it to go forward!

telephon:
It is an honor :) Typos can be very powerful in programming, but I think this one was [the least expensive typo in history](https://en.wikipedia.org/wiki/Mariner_1#Launch_failure).

Issue #5383, opened 2021-03-04
Replace stutter, Pstutter, etc. with non-ableist alternatives


mossheim:
## Motivation

Per discussion at https://scsynth.org/t/terminology-question/3372, using "stutter" this way is ableist and actually also not a good description of what these features even do. We should replace them with clearer and more inclusive alternatives.

## Description of Proposed Feature

- Deprecate these functions and classes and replace them with new ones:
  - Dstutter
  - Pstutter
  - `Array:-stutter`
  - `List:-stutter`

## Plan for Implementation

Info on deprecation is here: https://github.com/supercollider/supercollider/wiki/git-workflow-and-guidelines#deprecation. Currently-deprecated things [here](https://github.com/supercollider/supercollider/tree/develop/SCClassLibrary/deprecated) may serve as a good guide/inspiration. Deprecating the actual UGen might prove difficult, this probably needs more discussion and can maybe be split into a separate issue if needed? The rest should be easy.

brianlheim:
Other comments from that thread which are relevant to this discussion:

From @Sciss:

> There are three occurrences: `.stutter` on collections, `Pstutter` (patterns), and `Dstutter` (demand UGen).
> 
> There is this equivalence 
> ```supercollider
> [4, 5, 6].stutter(3)
> [4, 5, 6].collect(_.dup(3)).flatten
> [4, 5, 6].dup(3).flop.flatten
> ```
> The second "bracket" is due to SC lacking a `flatMap` operator (I guess it would be called `flatCollect`). Putting these terms together is probably not feasible. I would also not use terms like 'multiply', 'repeat', or 'times' that could be confused with arithmetic operations, or operations on the entire collection. Somehow the operation _on each element_ must be conveyed. And there is no `Each` tradition in SuperCollider (AFAIK), so there is no precedence of saying `dupEach` or `repeatEach`.
> 
> I guess `reproduce` could work, even though it's another metaphor, given that we imagine a bottom-up multiplication on the basis of individual elements of a population? In terms of a non-interpolating expansion, perhaps `stretch` could work as well, if it wasn't already used by patterns. Maybe `dilate`? Or if you think of a volume (balloon), perhaps `inflate`.
> 
> __Edit:__ Or if staying with the idea of getting stuck: `hang` (taken by `Condition`), `stick` (bit unclear), `stall`.

From gurk (not sure if this is the same person as www.github.com/gurk):

> Just had a thought while reading, may as well share.
> 
[...]

PR #5470, opened 2021-06-06
Use more inclusive language for repetitions

jrsurge:
## Purpose and Motivation

As discussed in #5383 and our continuing push for greater inclusivity in the project, this adds alternatives for the creation and utilisation of repetitions in user code:

| Original        |  Alternative   |
|---------------|---------------|
| `.stutter`        |  `.dupEach`   |
| `Dstutter`      | `Ddup`          |
| `Pstutter`       | `Pdup`          |
| `PdurStutter` | `Psubdivide` |

Due to the long-time use of the original versions, this PR does not deprecate them.
The use of the original versions will, however, produce a warning recommending the use of the alternatives.

## Implementation details
* The original methods now print a warning in favour of the new alternatives.
    * This may not actually be desired, particularly for the patterns (which will likely warn on each subsequent call?)
    * The Help files have been modified to recommend the new alternatives, so the warning might not be required
* The original methods then redirect to the new alternatives. This is for two reasons:
    1) It ensures that any future development is carried out on the new alternatives.
    2) It ensures that we only need to maintain one set of implementations.
* The documentation and classlibrary have been modified to use the new alternatives.

This PR needs to be carefully reviewed and tested, I do not know the pattern system well enough to ensure I haven't broken anything - but the examples still seem to work, so I think it's ok.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
    * The unit tests still pass, but more testing would be good
- [x] All tests are passing
- [x] Updated documentation


SCClassLibrary/Common/Streams/FilterPatterns.sc
```diff
@@ -589,6 +589,14 @@ Pbindf : FilterPattern {
 
 
 Pstutter : FilterPattern {
+	var <>n;
+	*new { arg n, pattern;
+		"The use of Pstutter is not recommended. Please use Pdup instead.".warn;
```
telephon, line 594: This will create hundreds of warnings in all kind of projects and force their authors and/or users to do this replacement. In this case, I would prefer if we would allow the language use to change actively because of the correct considerations rather than through pressure.

telephon (commented):
Thanks a lot for moving forward! There is one issue to be further discussed, I've added it to where it is implemented.

SCClassLibrary/Common/Streams/FilterPatterns.sc
dyfer: I would also support not generating warnings in this PR, and instead count on the change happen in a more frictionless manner over a possibly longer period of time.

HelpSource/Classes/Psubdivide.schelp
```diff
@@ -0,0 +1,81 @@
+class:: Psubdivide
+summary:: partition a value into n equal subdivisions
+related:: Classes/Pdup
+categories:: Streams-Patterns-Events>Patterns>Repetition
+
+description::
+
+A filter pattern designed for a stream of float durations.
+
+Subdivides each duration by each stutter and yields that value stutter times. A stutter of 0 will skip the duration value, a stutter of 1 yields the duration value unaffected.
```
dyfer, line 10: Do we want to use the work `stutter` here?

dyfer:
One request I'd have is that every class that now returns a different one on creation, has that clearly indicated in the help (probably as a `note:: ::`)

SCClassLibrary/Common/Streams/FilterPatterns.sc
jrsurge: Thanks for the comments, apologies for the delay in responding, I've been thinking about this one a lot.

Just to clarify, is this comment regarding all of the warnings, just `Pstutter`, or the types that now warn?

I agree that this will be noisy, but I think it's important that we have a warning of some sort. 

Firstly, users will never receive the old types, which on its own deserves a warning:
```supercollider
Pstutter()
// -> a Pdup
```

And secondly, I'm worried that existing users have little reason to look at the docs for this sort of thing, so only new users will find and use the alternatives, but maybe that is part of a larger discussion on how we communicate this sort of change.

With that in mind, what do you think of the following?
```supercollider
Pstutter : FilterPattern {
   classvar constructNew;

    *initClass {
        constructNew = { |n,pattern|
            "The use of Pstutter is not recommended. Please use Pdup instead.".warn;
            constructNew = {|n,pattern| Pdup(n,pattern)};
            constructNew.(n,pattern);
        };
    }

    *new { |n, pattern|
        ^constructNew.(n,pattern);
    }
}
```
This might look a little odd at first, but it's a strategy pattern, swapping out the behaviour at runtime on first call. 

It prints the warning once and then because there's no conditional, we don't pay for it again. it's as close to zero-cost as I could think of? Also, I'm not keen on `constructNew` as the function name, but it's the best I could come up with so far, any suggestions appreciated!

Many thanks!

HelpSource/Classes/Psubdivide.schelp
jrsurge: Oops, that's definitely a copy-paste error - thanks for catching!

joshpar:
What about calling it ‘deprecatedNew’ instead? This way, the naming doesn’t imply that this is a general new constructor example, but just something that is around until the deprecation happens. 

/*
Josh Parmenter
www.realizedsound.net/josh
*/

telephon:
I'm not sure if we should ever really deprecate the old ones. At least I feel a bit uneasy to enforce the new use by technical means. 

Also, it will affect many compositions.

SCClassLibrary/Common/Streams/FilterPatterns.sc
joshpar: perhaps 'suggestNew' is a classier name that helps clarify the intent.

dyfer:
Just checking in - what else do we need in this PR?

bla3kw3b:
I agree with @telephon ... those post window warnings are an eyesore... 

Plus it won't mend any broken hearts, to post warnings for already completed works...

dyfer:
I think the idea is to post the warning only once (per class library compilation) which IMO would be fine.

bla3kw3b:
> post the warning only once (per class library compilation)

A warning for every launch... right after method overrides ?

muellmusik:
I think we should avoid a semi-permanent warning posted for every user (including those using the new classes) on every compile. 

If we’re not satisfied with the soft incentives of documentation and returning a different class, then I think it may be time to implement a warnOnce method, that warns only once on first usage (could be this session or even once on this machine). This sort of problem has come up before.

dyfer:
Just to be clear, the warning would not be posted _on_ every compile, but on the first use of the discouraged class _after the recompile_. Sounds like a reasonable soft discouragement for me - for active code one has the option to simply update classes to use the recommended ones, which would make the warning go away, and make users use the recommended classes, which IMO is desirable.

telephon:
>but on the first use of the discouraged class after the recompile. 

So you mean when the class is used in some compiled class?

> and make users use the recommended classes, which IMO is what we ultimately want.

I have suggested that we shouldn't post warnings when you call code that has the old classes but instead recommend it in the documentation.

dyfer:
> > but on the first use of the discouraged class after the recompile.
> 
> So you mean when the class is used in some compiled class?
> 

I meant that the warning would be posted when the class is first called in the user code (first invocation after recompile). If you look at @jrsurge 's [implementation](https://github.com/supercollider/supercollider/pull/5470#discussion_r650429519) isn't it quite clear how it would behave?

> > and make users use the recommended classes, which IMO is what we ultimately want.
> 
> I have suggested that we shouldn't post warnings when you call code that has the old classes but instead recommend it in the documentation.

I understand your position. I'd however personally support the single warning (as proposed in this PR), on top of the suggestion in the docs.

dyfer:
@telephon is the currently proposed solution satisfactory for you? I'd love to see this PR move forward as this is the last one we're waiting for before the 3.12 release. Thanks!

telephon:
@dyfer most of it is perfect, thank you.

As I have said, I just have an issue with the warnings like this: https://github.com/supercollider/supercollider/blob/06d6873d90cafd9d153be576c51214b8f5cdbd4d/SCClassLibrary/Common/Streams/FilterPatterns.sc#L594

I think we should do without enforcement in this case.

dyfer:
> I think we should do without enforcement in this case.

Just to make sure we're on the same page - are you against the warning even if it's posted only _once_ on the _first use_ of the class? 

Visual noise aside, `Pstutter()` will now return `a Pdup` (and `Pdurstuter` will return`Psubdivide`) which is arguably an unexpected behavior. A warning at least gives heads up about that.

joshpar:
I wouldn't call it 'enforcement' - I'd call it informative since a level of introspection is lost:
a = Pstutter() ...
a.class // > Pdup

jamshark70:
FWIW:

- I would not agree to a warning on every use. (The code Julian cited does this, but it's also unofficially out of date because there was a concrete proposal to cut the messages down to one per session.)
- I would agree to a warning on first use in the session. (Here, I'm not sure where the miscommunication lies: The question in https://github.com/supercollider/supercollider/pull/5470#issuecomment-865255248 was "How do you feel about a warning once per session?" and the reply was, "I'm against warning upon every use.")
- I think it would be puzzling to post nothing at all, ever.

```
Pstutter : FilterPattern {
   classvar constructNew;

    *initClass {
        constructNew = { |n,pattern|
            "The use of Pstutter is not recommended. Please use Pdup instead.".warn;
            constructNew = {|n,pattern| Pdup(n,pattern)};
            constructNew.(n,pattern);
        };
    }

    *new { |n, pattern|
        ^constructNew.(n,pattern);
    }
}
```

This would be fine with me.

joshpar (approved):

dyfer (changes requested):
I've looked around, run through examples etc. Looks good to me! I just found 2 small things to correct.

HelpSource/Classes/Pdup.schelp
```diff
@@ -0,0 +1,91 @@
...
+		0.12.wait;
+	}
+}.play;
+)
+
+
+// Voss algorithm for pink noise
+// note that pattern.stutter(n) is a synonym for Pdup(n, pattern)
+a = Pwhite();
+c = (2 ** (0..10)).sum { |n| a.stutter(n) };
```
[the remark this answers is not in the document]
jrsurge: thanks for the careful review - should be fixed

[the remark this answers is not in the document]
dyfer: thanks!

dyfer (approved):
Thanks @jrsurge and everyone else for your work and input on this!

PR #5474, opened 2021-06-09
Made Supernova bind to the specified address, instead of always liste…

### LINKED ISSUE #4864: supernova binds to 0.0.0.0 by default
## Environment

* SuperCollider version: 3.11/develop
* Operating system: macOS

## Steps to reproduce

```supercollider
Server.supernova;
s.boot;
```
in terminal
```
lsof -i UDP
```
result:
```
...
supernova 34622 marcin   19u  IPv4 0xdd06c08b2563682f      0t0  UDP *:57110
...
```

It seems that `supernova` binds its UDP or TCP socket to `0.0.0.0` by default, similarly to what `scsynth` used to do: https://github.com/supercollider/supercollider/issues/4496. See this issue for more details and discussion.

Currently `scsynth` binds to `127.0.0.1`; `supernova` should probably be doing the same.

EDIT: according to #4496, a fix for `supernova` would be more involved (adding a new command line argument).

laserbat:
…ning on all interfaces.

## Purpose and Motivation

Fixes #4864

Supernova currently binds to 0.0.0.0 regardless of address provided in arguments. This PR makes it honor user supplied address, as well as binding on 127.0.0.1 by default.

As @brianlheim pointed out in #4864, this fix can potentially cause issues if there are multiple supernova servers running on same port but different interfaces. I think it's worth to create a separate issue to discuss how this can be fixed, possibly in a different PR. 

## Types of changes

- New feature
- Breaking change

This PR changes the default (insecure) behavior, by making supernova listen on 127.0.0.1 when no address is supplied. The security improvement, as well as making supernova behave more like scsynth should justify this change.

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Server.schelp
+++ b/HelpSource/Classes/Server.schelp
@@ -386,6 +386,10 @@ method:: hasShmInterface
 Returns true if a link::Classes/ServerShmInterface:: is available. See also link::Classes/Bus#Synchronous control bus methods::.
 The shared memory interface is initialized after first server boot.
 
+warning::
+Currently, the shared memory interface treats server port as an unique identifier. In the rare case when there are multiple servers running on the same machine and listening on the same port, this leads to the shared memory interface of one of the servers being overwritten by the other.
+::
+
 method:: serverBooting
 code::true:: if the server is booting, code::false:: otherwise.
 
--- a/HelpSource/Classes/ServerOptions.schelp
+++ b/HelpSource/Classes/ServerOptions.schelp
@@ -98,7 +98,7 @@ The IP address that the server's TCP or UDP socket is listening on. The default 
 warning::
 Until SuperCollider 3.10.3, this was set to code::"0.0.0.0":: (listen on all network interfaces). However, this is a dangerous default configuration — for most users working on laptops connected to WiFi, this means that anyone on your local network can send OSC messages to the server. code::"0.0.0.0":: is only useful if you are running networked server/client, and only safe if your networking is properly configured.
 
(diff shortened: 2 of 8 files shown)


laserbat:
Opened an issue to discuss the potential shmem collision: #5475

laserbat:
Just realized that this fix doesn't work with ipv6 and that I should validate address early, as suggested in @brianlheim's comment. Will correct this ASAP.

laserbat:
I fixed ipv6 issue and corrected my code formatting. Hopefully it'll pass the tests

dyfer:
Thank you! This works as advertised, tested on macOS.

I'd like for someone else to also look over this PR before approving/merging.

laserbat:
Cool! Glad it works on macOS as well. Since you added it to 3.12.0 milestone, should I write that version in the updated documentation? I used "3.12.x" as a placeholder.

dyfer:
@laserbat good point about the docs. Just FYI we weren't planning on getting any new PRs into 3.12, but since it changes the default behavior (and tightens security, and makes supernova behave the same as scsynth) I think it would be desirable to include it. 

In that case yes, the docs could be changed stating that the old behavior was in place "Before SuperCollider 3.12" (or 3.12.0). I also think "before" is better than "until" since reading it I wouldn't be sure if the latter means including or excluding that version. But that's a really minor thing, feel free to phrase it the way sounds good to you.

laserbat:
I see. Thanks for clarifying this, I will update the documentation in a few minutes :)

joshpar (approved):
Thank you! I think this looks fine to me.

PR #5483, opened 2021-06-17
QWebView: added ability to set scrollPosition of QcWebView

paum3:
## Purpose and Motivation
WebView scrollPosition setter was not impemented even it was in docs. 

## Types of changes
This small example demonstrates what it does:

```
(
    ~webview = WebView();
    ~webview.url = "https://ferokiraly.com";
    View(bounds:900@700).layout_(VLayout(~webview)).front;
)


~webview.zoom_(1)
~webview.scrollPosition
~webview.scrollPosition_(0@504) // scroll it  !
~webview.scrollPosition // 0@504
~webview.zoom_(2)
~webview.scrollPosition_(0@0) // scrolls correctly back to top
~webview.zoom_(1) // zooms back


```

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/QtCollider/widgets/QcWebView.h
+++ b/QtCollider/widgets/QcWebView.h
@@ -116,8 +116,12 @@ public:
     Q_PROPERTY(QSizeF contentsSize READ contentsSize)
     QSizeF contentsSize() const { return page() ? page()->contentsSize() : QSizeF(0, 0); }
 
-    Q_PROPERTY(QPointF scrollPosition READ scrollPosition)
+    Q_PROPERTY(QPointF scrollPosition READ scrollPosition WRITE setScrollPosition)
     QPointF scrollPosition() const { return page() ? page()->scrollPosition() : QPointF(0, 0); }
+    void setScrollPosition(QPointF p) {
+        if (page())
+            page()->runJavaScript(QString("window.scrollTo(%1, %2);").arg(p.rx()).arg(p.ry()));
+    }
 
     Q_PROPERTY(bool audioMuted READ isAudioMuted WRITE setAudioMuted)
     bool isAudioMuted() const { return page() ? page()->isAudioMuted() : false; }
--- a/SCClassLibrary/Common/GUI/Base/QWebView.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWebView.sc
@@ -253,7 +253,7 @@ WebView : View {
 	contentsSize { ^this.getProperty('contentsSize') }
 
 	scrollPosition { ^this.getProperty('scrollPosition') }
-	scrollPosition_ { |point| ^this.setProperty('scrollPosition', point) }
+	scrollPosition_ { |point| ^this.setProperty('scrollPosition', point / this.zoom) }
 
 	audioMuted { ^this.getProperty('audioMuted') }
 	audioMuted_ { |muted| this.setProperty('audioMuted', muted) }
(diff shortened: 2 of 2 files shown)


paum3:
Hi guys, is there some issue not to review this pull? The aim is to bring vim keys to scdoc which is I think a great feature for those who use nvim as IDE. This pull brings the main part to it- ```scrollPosition``` even in docs, never implemented. So as soon as this pull will be accepted I can continue to send a new pull with vim keybindings in ```HelpBrowser```. Thank you all.

dyfer:
Sorry for the delay on this! It's just that there aren't many active reviewers lately. 
I'll do my best to get to this at some point. Thanks for the PR and for your patience!

QtCollider/widgets/QcWebView.h
```diff
@@ -116,8 +116,12 @@ public Q_SLOTS:
     Q_PROPERTY(QSizeF contentsSize READ contentsSize)
     QSizeF contentsSize() const { return page() ? page()->contentsSize() : QSizeF(0, 0); }
 
-    Q_PROPERTY(QPointF scrollPosition READ scrollPosition)
+    Q_PROPERTY(QPointF scrollPosition READ scrollPosition WRITE setScrollPosition)
     QPointF scrollPosition() const { return page() ? page()->scrollPosition() : QPointF(0, 0); }
+    void setScrollPosition(QPointF p) {
+        if (page())
+            page()->runJavaScript(QString("window.scrollTo(%1, %2);").arg(p.rx()).arg(p.ry()));
```
joshpar, line 123: is this really the correct way to do this? running JS to set a scroll position feels... ok? strange? Curious if there isn't another way to do this more 'natively'

dyfer: I was wondering about the same and found on the qt forum that indeed suggesting to use JS, as webengine view doesn't seem to provide such functionality atm.

paum3: yes, it is strange I agree but @dyfer is correct.

dyfer:
Thanks @paum3 
I tested this and it works as expected.
One comment/request I have is: would you consider moving calculation based on zoom to the cpp side? I fell like it would be more appropriate to get and set the same values to/from qt.

paum3:
> Thanks @paum3
> I tested this and it works as expected.
> One comment/request I have is: would you consider moving calculation based on zoom to the cpp side? I feel like it would be more appropriate to get and set the same values to/from qt.

paum3:
uups, closed by mistake. so reopened.

@dyfer I am not sure about moving zoom calculation into cpp side for these reasons:
- there is already z variable ```zoom``` defined on sc side, so I felt like this has some logic to use it (and easyness + cleariness)
- if somebody would like to experiment with WebView and zoom funcionality, there will be no 'hardcoded' algorithms on cpp side
- and last, I have to be honest, I have looked into cpp code, and it look like not so easy trick to implement this

So it is all up to you how you decide. If you feel that my implementation is wrong (I am amaterur programmer) I have no choice but believe you and then I can close this pull. On the other side I have to recapitulate the reasons of this pull:

-  WebView scrollPosition setter was not impemented even it was in docs.
-  it is crucial for my next step to SC community to add 'vim like keys' navigation  to the docs which I found useful when using vim and Qt docs. I am using it already in my custom fork and can't be without it - so maybe it could be good for others.

fero

dyfer:
Thanks @paum3 

Your contribution is definitely useful and I'd love to pull it in.
After reading your comment I too have looked at the cpp side and indeed maybe it's not as straightforward to use zoom there as I have thought. While I feel like it would be more "proper" to get and set the same values from Qt and avoid  calculation in the language, it is better to have this implementation than to not have it implemented at all.

dyfer (approved):
I think the current implementation is fine. In the future we might look into moving calculation of scroll position based on zoom to the Qt (cpp) side. If there are no objections, I'd merge this next week or so.

PR #5490, opened 2021-06-28
Fix insulting phrasing on Tutorial

capital-G:
## Purpose and Motivation

I think describing the playing of a bassist as *idiotic* should not be part of the documentation.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Tutorials/A-Practical-Guide/PG_Cookbook08_Swing.schelp
+++ b/HelpSource/Tutorials/A-Practical-Guide/PG_Cookbook08_Swing.schelp
@@ -87,7 +87,7 @@ Ppar([
 
 (
 q = Ppar([
-	// walking bass (by an idiot bass player who only chooses notes randomly)
+	// walking bass (by a bass player who only chooses notes randomly)
 	Pbind(
 		\octave, 3,
 		\degree, Pwhite(0, 7, inf),


jamshark70:
> I think describing the playing of a bassist as *idiotic*...

Actually, you've put your finger on the reason why the phrasing is not really a problem.

It's *a* bassist. Not bassists in general.

The latter would be insulting to a large class of potential readers. The former refers to an imagined person.

In English, commas are semantically significant in conjunction with "who" or "which" subordinate clauses.

- "A bassist who chooses notes randomly" indicates an individual member of the class of all bassists, and this individual member chooses notes randomly.

- "A bassist, who chooses notes randomly, ..." indicates that the entire class of bassists is distinguished by the behavior of choosing notes randomly.

In jazz circles, choosing notes randomly (without awareness of harmonic context or the other players' inflections) would be considered... naïve to the point of unprofessionalism. I could easily imagine jazz players using the term "idiot" for such a player.

If this phrasing is insulting... insulting to whom? If it refers to bassists in general, then yes, it would be insulting -- but the absence of a comma rules out this interpretation. It simply isn't correct to read it that way. If it instead refers to a specific, hypothetical bassist, and one feels insulted, then one is voluntarily identifying with musical behavior that would be rejected in the referenced musical culture -- which... why would one do that?

I won't stand in the way of this change, but I do wish to register that its likely justifications evaporate under scrutiny.

muellmusik (approved):

muellmusik:
> If this phrasing is insulting... insulting to whom?

Crazy free jazz players perhaps. In any case this change does no harm, and the inclusion of 'idiot' is negative and adds no value I think, so I've reviewed and approved.

jamshark70:
> In any case this change does no harm, and the inclusion of 'idiot' is negative and adds no value I think, so I've reviewed and approved.

Sure.

My reason for posting is that consideration of offensive language really needs to take context into account. Racist, sexist, homophobic, ableist, etc. speech is offensive *because of* context. In these cases, we often don't need to be explicit about the context because the large majority of us in the SC community understand it already. In ambiguous cases such as this one, I think it's helpful to be clear and direct about the context -- but this was not done here; it was simply assumed.

Unfortunately, we live in a world where racist etc. speech exists, *and* in a world where a USC professor can refer to the Mandarin demonstrative pronoun 那个 and then be suspended, because the Mandarin word happens to sound superficially like an American racist slur (despite the lack of even the most tenuous etymological connection between them). The latter incident is a real case and extremely troubling. I would like us not to follow along with that case's precedent of fact-free assumption of guilt. The reasoning here seems to be "I saw a word and I didn't like it" and I maintain that this should not be sufficient.

However, I don't mind the specific change, so, go ahead and merge.

capital-G:
Thanks for the discussion, I think this is an interesting topic worth discussing.

> In jazz circles, choosing notes randomly (without awareness of harmonic context or the other players' inflections) would be considered... naïve to the point of unprofessionalism. I could easily imagine jazz players using the term "idiot" for such a player.

Thats nice for the jazz players but I think a musical framework like SuperCollider should be open to any kind of sonic expression and not be judgemental about certain kind of playing styles of an instrument.
At least not in this insulting way of describing it as *idiotic*, maybe *unusual*.

This is also especially a turnoff when trying to learn a new insturment or framework and you are happy about finally  some sounds but there is already a judgement about its sonic properties given by the tutorial.

> If this phrasing is insulting... insulting to whom? If it refers to bassists in general, then yes, it would be insulting -- but the absence of a comma rules out this interpretation. It simply isn't correct to read it that way. If it instead refers to a specific, hypothetical bassist, and one feels insulted, then one is voluntarily identifying with musical behavior that would be rejected in the referenced musical culture -- which... why would one do that?

I could also ask why one would want to stay in the referential musical culture?

dyfer:
Thanks for the changes and for the discussion.
I recognize @jamshark70 's comments about this phrase possibly not being overly offensive etc.
IMO there's no need to use such a judgmental adjective in this context so I'll merge this.

jamshark70:
Prefacing with -- I fully agree with the merge, thanks to capital-G for doing it and Marcin for merging it. (I spoke of context -- one reasonable context might be, suppose an SC user is a relative of an autistic or Down's syndrome person, and had to witness bullying with the term "idiot" -- it is fair that this user might not want to be casually reminded of it in our docs. So it's not hard to move beyond "it's offensive" to "it's offensive *because*...")

> Thats nice for the jazz players but I think a musical framework like SuperCollider should be open to any kind of sonic expression... 

I don't think that sonic styles found in documentation examples should ever be taken as somehow normative, should they?

It's a tutorial chapter about swing rhythm, with an example emulating walking bass (rhythm)... it's not really much of a stretch to demonstrate swing by reference to jazz. What *is* a stretch would be to suppose that readers would necessarily internalize the stylistic reference and think that code for swing is not useful for anything other than jazz. I have a bit more faith in our users' creativity and iconoclasm than that :wink:

PR #5497, opened 2021-07-11
remove duplicate example from PdurStutter.schelp

JaimeClover:
## Types of changes

- Documentation

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/PdurStutter.schelp
+++ b/HelpSource/Classes/PdurStutter.schelp
@@ -48,21 +48,6 @@ Routine({
 	})
 }).play(TempoClock.default);
 )
-
-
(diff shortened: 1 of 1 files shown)


telephon (approved):

dyfer:
Thanks @JaimeClover 
That example is actually slightly different than the one above, is it worth removing it?

JaimeClover:
Ahh my mistake! I could have sworn they were identical, but now I see the difference. I agree the example should stay because it demonstrates what happens when n=0. Sorry for the noise, and thanks for all your work on SC!

PR #5499, opened 2021-07-11
Better function flop implementation that works with string ellipsis and keyword arguments

### LINKED ISSUE #5498: makeFlopFunc fails with ellipsis arguments
## Steps to reproduce

```supercollider

f = { |a, b ... c| [a, b, c] }.makeFlopFunc;
f.(1, [2, 3], [4, 5], [6, 7]);
// should be [ [ 1, 2, [ 4, 6 ] ], [ 1, 3, [ 5, 7 ] ] ]

// but throws an error:
ERROR: syntax error, unexpected ELLIPSIS, expecting NAME or WHILE
  in interpreted text
  line 1 char 17:

  #{ arg  a, b, ... c; [ a, b, ... c ].flop }; 
                ^^^
```

telephon:
## Purpose and Motivation

Reimplements `flop` by `makeFlopFuncString`, which returns a flopped function s that can be used with keyword arguments. This also fixes #5498. Ellipsis arguments didn't work before, and for all usual cases there tests have been added.

## Types of changes

- Bug fix
- New Functionality

There is a possibly breaking change, that functions without arguments will also flop: `{ 3 }.flop.value`returns `[3]` and not `3`. just like  in the case of arrays `[3].flop` returns `[[3]]` and not `[3]`. It is probable that the original behaviour caused more trouble, because it is unexpected. This change is up to discussion and can be separated from the rest of the changes if so desired.

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/HelpSource/Classes/Function.schelp
+++ b/HelpSource/Classes/Function.schelp
@@ -281,14 +281,7 @@ f.value(2, 3);
 
 method::envirFlop
 
-like flop, but implements an environment argument passing (valueEnvir).
-Less efficient in generation than flop, but not a big difference in evaluation.
-
-code::
-f = { |a| if(a > 0) { a + 1 } { -inf } }.envirFlop;
-e = (a: [20, 40]);
-e.use { f.value }
-::
+This method is kept for backward compatibility. Use code::flop:: instead.
 
 
 method::inEnvir
--- a/SCClassLibrary/Common/Core/Function.sc
+++ b/SCClassLibrary/Common/Core/Function.sc
@@ -248,13 +248,33 @@ Function : AbstractFunction {
 		}
 	}
 
-	makeFlopFunc {
+	makeFlopFuncString {
+		var ellipsisArgName, valueBlock, i, argBlock;
 		if(def.argNames.isNil) { ^this };
 
-		^interpret(
-			"#{ arg " ++ " " ++ def.argumentString(true) ++ "; "
-			++ "[ " ++ def.argumentString(false) ++ " ].flop };"
-		)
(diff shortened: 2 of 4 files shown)


telephon:
Open question: do we want function.flop to use `flopInterpret`? Disadvantage: slower in creation. Advantage: behaves exactly like the original in terms or argument passing.

joshpar:
> Open question: do we want function.flop to use `flopInterpret`? Disadvantage: slower in creation. Advantage: behaves exactly like the original in terms or argument passing.

@telephon - yes. I think it should. My gut says that methods like this aren't being called thousands of times per second... and even then if they are, the 'slower' still may not be a problem... So I suggest the change to function.flop.
However - we could also consider retaining the original method's implementation with a new method name (floptimized?). I don't really think that is necessary, but if the speed really is a concern it wouldn't be a bad approach. 

I vote for updating function.flop, then I'll approve and merge.

joshpar (changes requested):
use makeFlopFuncString for function:flop

telephon:
ok, good – I'll do this.

dyfer:
Hello @telephon , any update on this?

telephon:
Thanks for returning to this. I'd like to do this in August.

telephon:
I've found a problem with ellipsis args, will try to fix asap.

telephon:
OK, that is resolved, ready for review.

telephon:
(I've updated the `Purpose and Motivation` text at the beginning).

joshpar (approved):

PR #5523, opened 2021-07-25
Remove sc-el from build and build instructions

jxa:
## Purpose and Motivation

SuperCollider currently bundles language modes for emacs, in addition to vim, gedit, etc. This is convenient for users on linux platforms, but this convenience comes with a cost. The mode can not be installed independently without running cmake or manually editing the source code. A number of issues have been opened on the `scel` repo, and no clear instructions are given. https://github.com/supercollider/scel/pull/38 is my proposal to address these installation issues. 

Package managers for emacs have come a long way in the last 10+ years since the decision was made to bundle language modes with SC. Today emacs users have an expectation that they can install packages via a couple of standard methods, usually a package on [MELPA](https://github.com/melpa/melpa) or direct from a git repo via [straight.el](https://github.com/raxod502/straight.el).

With this PR I propose that we remove `sc-el` from the build and distribution of SuperCollider. Instructions can be provided for users to easily install the mode with the package manager of their choice. If the community desires it, a linux package can be created that wraps the library in the current way. The important thing that I'm trying to accomplish is to decouple the build of the emacs package from the build of SuperCollider, so that it can better serve the needs of the emacs user base and can evolve independently, outside of the SC release schedule.

I asked emacs users to [weigh in on this topic](https://scsynth.org/t/future-of-the-emacs-library/4221/17) and no objections have yet been raised. Mostly folks are excited to see some attention to the emacs mode.

## Types of changes

- Documentation: removes the `DSC_EL` flag from compile instructions
- Breaking change: build of the supercollider-emacs linux package will no longer happen. Existing installs will continue to work, but won't receive updates

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

## Alternatives

- We could continue to create this package, but we should remove the `sclang-vars.el.in` file and build step so it can be independently distributed for non-linux users.


jxa:
After discussion with @dyfer and @joshpar I'm going to attempt a dual-build in order to support existing linux install workflows as well as package-based installs

gusano:
This looks great

PR #5507, opened 2021-07-14
Add wiki content to doc

gusano:
## Purpose and Motivation
Following the discussion with @dyfer in https://github.com/supercollider/supercollider.github.io/issues/247, this adds the `UserFAQ` page from the website WIKI to the doc.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] Updated documentation
- [x] This PR is ready for review

PR #5539, opened 2021-08-10
Update errors and outdated usages in Control Structures help

jamshark70:
## Purpose and Motivation

Updates Control Structures help:

1. At one point we had a code style guideline document. (Aside: Has this been deprecated intentionally? https://supercollider.github.io/contributing/ links to https://supercollider.github.io/development/code-style-cpp.html but "Sorry this page does not exist =(.") In that document, we recommended trailing-closure syntax for control flow structures. But, Control Structures help consistently recommended e.g. `if(expr, trueFunc, falseFunc)` which... basically nobody writes it like that anymore. So I updated all the syntax examples to put trailing-closure syntax first, and present others as alternatives.

2. The `switch` section claims that an equality `==` test is always used. But, if it's inlined, the hash lookup is based on identity. This has actively misled at least one user (https://scsynth.org/t/strange-switch-behavior/4362). So I explained the difference and provided an example.

3. The `switch` section says that floats will be inlined (correct) but doesn't call attention to the problems with using floats with `switch`, misleading the aforementioned user into believing that floats simply transparently work with `switch`. So I added a note and an example.

4. The `do` section was a little unclear about binary infix notation for a series expression + do, e.g. `(1..5) do: { /* body */ }`.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/Control-Structures.schelp
+++ b/HelpSource/Reference/Control-Structures.schelp
@@ -59,10 +59,10 @@ code::
 while { testFunc body } { loopFunc body };
 
 // alternate
-while (testFunc, bodyFunc);
+while({ testFunc body }, { loopFunc body });
 
 // possible but harder to read, so, not preferred
-testFunc.while( bodyFunc );
+{ testFunc body }.while({ loopFunc body });
 ::
 
 Example
@@ -194,7 +194,7 @@ Object implements a switch method which allows for conditional evaluation with m
 
 The switch statement will be inlined if the test objects are all Floats, Integers, Symbols, Chars, nil, false, true and if the functions have no variable or argument declarations. The inlined switch uses a hash lookup (which is faster than nested if statements), so it should be very fast and scale to any number of clauses.
 
-note::Hash lookup by definition implies testing emphasis::identity:: rather than equality: a code::switch:: construction that is not inlined will test code::==::, while one that is inlined will test code::===::. See the examples.::
+note::Hash lookup by definition implies testing emphasis::identity:: rather than equality: a code::switch:: construction that is not inlined will test code::==::, while one that is inlined will test code::===::. See the examples. One implication is that strings should be avoided: code::switch(text) { "abc" } { ... }:: may not match, even if code::text == "abc"::. Symbols are preferred.::
 
 note::Floating point numbers may sometimes appear to be equal while differing slightly in their binary representation: code::(2/3) == (1 - (1/3)):: is false. Therefore floats should be avoided in code::switch:: constructions.::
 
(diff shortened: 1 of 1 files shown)


thresholdpeople:
The style guides still work if accessible via the wiki link directly on this github page. Not sure if the hostname and location was inserted automatically for both supercollider.github.io and/or github.com/supercollider/supercollider... ?
https://github.com/supercollider/supercollider/wiki/Code-style-guidelines
https://github.com/supercollider/supercollider/wiki/Cpp-code-style-guidelines

dyfer:
The code style pages were moved from the old location to the new wiki, the old link is simply outdated.
BTW we are in the process of [migrating the old WIKI](https://github.com/supercollider/supercollider/pull/5507), i.e. the WIKI at supercollider.github.io will go away, with the relevant information being transferred to our [current WIKI](https://github.com/supercollider/supercollider/wiki) or the the help files themselves.

HelpSource/Reference/Control-Structures.schelp
```diff
@@ -11,63 +11,69 @@ section:: Basic Control Structures
...
 Syntax
 code::
-while ( testFunc, bodyFunc );
-::
---or--
-code::
+while { testFunc body } { loopFunc body };
+
+// alternate
+while (testFunc, bodyFunc);
```
dyfer, line 62: I think this should also be loopFunc instead of bodyFunc? 
Thinking about beginner users, how about we do `while ({ testFunc body }, { loopFunc body })` for similarity with the previous example?

HelpSource/Reference/Control-Structures.schelp
```diff
@@ -116,57 +125,109 @@ Do is used to iterate over a link::Classes/Collection::. Positive Integers also
...
 ::
 ::
 
 method:: switch
 
 Object implements a switch method which allows for conditional evaluation with multiple cases. These are implemented as pairs of test objects (tested using if this == test.value) and corresponding functions to be evaluated if true.
+
 The switch statement will be inlined if the test objects are all Floats, Integers, Symbols, Chars, nil, false, true and if the functions have no variable or argument declarations. The inlined switch uses a hash lookup (which is faster than nested if statements), so it should be very fast and scale to any number of clauses.
 
+note::Hash lookup by definition implies testing emphasis::identity:: rather than equality: a code::switch:: construction that is not inlined will test code::==::, while one that is inlined will test code::===::. See the examples.::
```
dyfer, line 197: Optional: maybe this sentence could also mention that this typically means using symbols instead of strings or something like that?

HelpSource/Reference/Control-Structures.schelp
jamshark70: Agreed -- better to be consistent. Done.

HelpSource/Reference/Control-Structures.schelp
jamshark70: Makes sense -- done.

dyfer (approved):
Thanks!

PR #5549, opened 2021-08-19
create duplicated window only on forceCreate

### LINKED ISSUE #5545: Many meters we have.
## Environment

* SuperCollider version: 3.12
* Operating system: Mac OS 10.13.6

## Steps to reproduce

Press Cmd+M more than once. 
Move a meter. 
See! 
More than one meter.

I dare to dream of a world where SuperCollider does not open up more than one server meter. 
Instead SC would show me the one I already have open when I press Cmd+M.

capital-G:
## Purpose and Motivation

Fixes #5545

## Types of changes

- Documentation
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -107,10 +107,10 @@ Window {
 		reuseExisting = false;
 
 		if( reuseExisting, {
-			var existingWindows = Window.allWindows.select({|w| w.name == name});
-			if( existingWindows.size > 0 ) {
+			var existingWindow = Window.allWindows.detect({|w| w.name == name});
+			if( existingWindow.notNil ) {
 				"Window \"%\" already exists".format(name).warn;
-				^existingWindows[0].front;
+				^existingWindow.front;
 			}
 		});
 
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -104,7 +104,16 @@ Window {
 		resizable = true,
 		border = true,
 		server,
-		scroll = false;
+		scroll = false,
+		forceCreate = false;
+
+		if( forceCreate.not, {
+			var existingWindows = Window.allWindows.select({|w| w.name == name});
+			if( existingWindows.size > 0 ) {
+				"Window \"%\" already exists".format(name).warn;
+				^existingWindows[0].front;
+			}
+		});
 
 		if( bounds.isNil ) {
 			bounds = Rect(0,0,400,400).center_( Window.availableBounds.center );
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -105,9 +105,9 @@ Window {
 		border = true,
 		server,
 		scroll = false,
-		forceCreate = false;
+		checkForExistingWindow = false;
 
-		if( forceCreate.not, {
+		if( checkForExistingWindow, {
 			var existingWindows = Window.allWindows.select({|w| w.name == name});
 			if( existingWindows.size > 0 ) {
 				"Window \"%\" already exists".format(name).warn;
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -106,8 +106,12 @@ Window {
 		scroll = false,
 		reuseExisting = false;
 
+		// declare variable here for performance reasons, see
+		// https://github.com/supercollider/supercollider/pull/5549#issuecomment-980518930
+		var existingWindow;
+
 		if( reuseExisting, {
-			var existingWindow = Window.allWindows.detect({|w| w.name == name});
+			existingWindow = Window.allWindows.detect({|w| w.name == name});
 			if( existingWindow.notNil ) {
 				"Window \"%\" already exists".format(name).warn;
 				^existingWindow.front;
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -105,9 +105,9 @@ Window {
 		border = true,
 		server,
 		scroll = false,
-		checkForExistingWindow = false;
+		reuseExisting = false;
 
-		if( checkForExistingWindow, {
+		if( reuseExisting, {
 			var existingWindows = Window.allWindows.select({|w| w.name == name});
 			if( existingWindows.size > 0 ) {
 				"Window \"%\" already exists".format(name).warn;
--- a/HelpSource/Classes/Window.schelp
+++ b/HelpSource/Classes/Window.schelp
@@ -38,6 +38,8 @@ METHOD:: new
 		This is a dummy argument which is here to provide compatibility with SwingOSC and has no effect.
 	argument:: scroll
 		A Boolean indicating whether this window will add scrollbars if its contents exceed its bounds. If this is set to code::true::, then link::Classes/View#-resize:: settings will be ignored for contained views. The default is false.
+	argument:: checkForExistingWindow
+		A Boolean for checking if a window with the same name already exists. If such a window exsits no new window will be created, the existing window with the same name will be taken to the front and a warning will be displayed.
 
 METHOD:: allWindows
 
--- a/HelpSource/Classes/Window.schelp
+++ b/HelpSource/Classes/Window.schelp
@@ -38,7 +38,7 @@ METHOD:: new
 		This is a dummy argument which is here to provide compatibility with SwingOSC and has no effect.
 	argument:: scroll
 		A Boolean indicating whether this window will add scrollbars if its contents exceed its bounds. If this is set to code::true::, then link::Classes/View#-resize:: settings will be ignored for contained views. The default is false.
-	argument:: checkForExistingWindow
+	argument:: reuseExisting 
 		A Boolean for checking if a window with the same name already exists. If such a window exsits no new window will be created, the existing window with the same name will be taken to the front and a warning will be displayed.
 
 METHOD:: allWindows


SCClassLibrary/Common/GUI/Base/QWindow.sc
```diff
@@ -104,7 +104,16 @@ Window {
 		resizable = true,
 		border = true,
 		server,
-		scroll = false;
+		scroll = false,
+		forceCreate = false;
+
+		if( forceCreate.not, {
+			var existingWindows = Window.allWindows.select({|w| w.name == name});
```
telephon, line 111: if you like, you could replace this by 

`alreadyExists = Window.allWindows.any({|w| w.name == name })`

SCClassLibrary/Common/GUI/Base/QWindow.sc
```diff
@@ -104,7 +104,16 @@ Window {
...
+			if( existingWindows.size > 0 ) {
+				"Window \"%\" already exists".format(name).warn;
```
telephon, line 113: if you have `forceCreate = false` you don't want warnings, I think.

telephon (changes requested):
I'd move this into a specific method e.g. called `*newIfNeeded` or move the code to where it is required. It is pretty short, so it would be fine as well.

jamshark70:
I have some reservations about this approach.

The problem is specific, but the proposed solution is a/ general and b/ changes default behavior for *all* code that doesn't take the new argument into account. This is a recipe for breaking compatibility.

For instance, SynthDesc:makeGui has a hardcoded, fixed window title, and it doesn't override `forceCreate: true` -- so, where it used to be possible to open any number of synth makeGui's, for the same or different SynthDefs, suddenly it would be possible to open one and only one.

I would favor a localized solution -- just fix the given problem. ServerMeter should keep track of whether it has a window open -- if there is, then `ServerMeter.new` would only `.front` the existing window. I don't see any reason to touch global window creation behavior for this issue.

capital-G:
> I would favor a localized solution -- just fix the given problem. ServerMeter should keep track of whether it has a window open -- if there is, then `ServerMeter.new` would only `.front` the existing window. I don't see any reason to touch global window creation behavior for this issue.

I think making this a global setting for window creation is the proper way to do when implementing this from scratch, but making this a retrograde fix is indeed problematic as there are so many paths in SC that one does not know - e.g. I never heard of SynthDesc before, I always used JitlibGUI before which would greatly benefit from such a behavior - when livecoding I tend to have 10 or 20 copies of the same window on top of each other in the heat of the moment.

I honestly do not expect this to get merged, therefore it is still a draft. I just wanted to show another user who opened the issue on how to fix this by oneself. I think fixing this solely for ServerMeter could also fix the problem, but for me the issue is the global spawning of multiple windows - I know there are different opinions on this and therefore I think this won't get merged.

SCClassLibrary/Common/GUI/Base/QWindow.sc
capital-G: I would have sought for something like `SELECT * FROM windows WHERE name='' LIMIT 1`.
Using any here is indeed clearer but one needs to scan over the window array two times this way - with the approach above there is only one scan needed - but I am happy to change this for readability.

SCClassLibrary/Common/GUI/Base/QWindow.sc
capital-G: Depends on if `forceCreate = false` gets default because this would prevent windows form popping up. Indicating this behavior, the prevention of a window creation, to the user via a warning is indeed intended.

jamshark70:
I'm not 100% opposed to a `*newIfNeeded` method -- in fact, that would address my concern about backward compatibility by leaving `*new` alone.

I have very real doubts about making the default behavior to suppress duplicate windows. A change in default behavior is risky. I'm going to have to vote against that. You can then change ServerMeter to use the nondefault behavior -- problem solved -- everyone else's code stays OK.

We've had questions/issues before about adding arguments to existing methods -- I'd prefer a new method over editing the existing one. Strongly prefer. I think Julian is correct to suggest that.

telephon:
> I have very real doubts about making the default behavior to suppress duplicate windows. A change in default behavior is risky.

One reason for me to be sceptical is that the identity check is the name of the window, which is a little ad hoc.

jamshark70:
> I honestly do not expect this to get merged, therefore it is still a draft...

I think it's easily fixed -- just move the new logic into a `*newIfNeeded` or (`*newOnce` ?) method, and revert `*new` to the way it was before. Then ServerMeter does `window = Window.newIfNeeded( /* same arguments */ )` and you have a/ a globally-available feature; b/ a localized bugfix; c/ no broken compatibility.

I think it's important to be careful about backward compatibility, but that concern doesn't invalidate this PR -- it only calls for a slightly different form of it.

dyfer:
AFAIU the implementation here is still WIP, but I just wanted to point out that there are some [tests failing](https://github.com/supercollider/supercollider/pull/5549/checks?check_run_id=3375553338#step:9:19289) which might be due to proposed changes. These might be helpful to ensure backwards compatibility :)

telephon:
> I honestly do not expect this to get merged, therefore it is still a draft. I just wanted to show another user who opened the issue on how to fix this by oneself. I think fixing this solely for ServerMeter could also fix the problem, but for me the issue is the global spawning of multiple windows - I know there are different opinions on this and therefore I think this won't get merged.

If there is a well-named method to open a window only once and you use that, I am confident that it will be merged.

dyfer:
I like the idea of preventing the opening of multiple server meters.

I am also in favor of localized solution. When creating GUIs manually, it feels like a heavy-handed assumption that the user wants the same window if it has the same name (especially for empty names) - even if this is an optional feature. If `JitlibGUI` also have this issue, I'd suggest fixing it there as well...

Maybe `*newIfNeeded` is a reasonable approach, but I'm still a bit uneasy about using the name string as the window identifier. Even if it works in this case, it seems conceptually unreliable.

Also, some tests are failing and they seem to be related to server GUI, so they indeed seem to catch some backward compatibility issue with these changes. I haven't looked exactly what the issue is, but I wanted to point this out for future work on such extension.

capital-G:
I changed the default behavior to no-check if such a window exists - this at least allows to introduce this argument when the check is necessary (although I used this as default behavior for 6+ months and never occurred any problems within the standard library).

Now we can introduce bit-by-bit PRs where the default behavior is changed, e.g. on server meter etc.

I wrote a test class for this which is currently not included b/c I have problems w/ my sclang interpreter in my dev environment.

Would be happy to get feedback if this is acceptable or not.

SCClassLibrary/Common/GUI/Base/QWindow.sc
```diff
@@ -103,7 +103,16 @@ Window {
 		resizable = true,
 		border = true,
 		deprecatedServerArgument,
-		scroll = false;
+		scroll = false,
+		checkForExistingWindow = false;
```
telephon, line 107: renaming suggestion: `reuseExisting = false`. This makes clear what is done when it exists.

SCClassLibrary/Common/GUI/Base/QWindow.sc
telephon: >"Window \"%\" already exists".format(name).warn;

_It is never good to warn when there is no danger._
We should assume that it was intentional if `reuseExisting` is set to true, right?

Better:
`"Reusing existing Window \"%\"".format(name).postln`

telephon (changes requested):
@dyfer had an issue with checking for the name if it is unspecified, but I think we can care for this down the line, when applying the fix.

SCClassLibrary/Common/GUI/Base/QWindow.sc
capital-G: I think the warning is justified as it marks that a statement of a user is changed in the background to some different behavior (e.g. calling `s.meter()` twice is probably an antipattern and therefore good for a warning - in case this was intended that the statement won't get executed is also worth a warning). It also depends where one is coming from, [in e.g. JS there is a trend](https://github.com/xojs/xo) to stop execution if the linter finds unformatted code or unused variables so every warning stops your program (the *nürnberger trichter* of coding? but at the same time I found this extremely productive once you do not have to manage coding styles yourself anymore. one step beyond this is maybe [vigil](https://github.com/munificent/vigil)).

I am totally on the side for more warnings in SC as I personally had many moments of frustration caused by the muteness of some SC internals - maybe the proper angle could be on how warnings should be obeyed.

In general I think SC could really benefit form some kind of advanced logging mechanism (see e.g. https://docs.python.org/3/howto/logging.html#logging-flow which is a good basic logging workflow for a standard library IMO) which could be adjusted depending on the situation or liking of anybody.

SCClassLibrary/Common/GUI/Base/QWindow.sc
capital-G: thanks, this is indeed much better

capital-G:
I think the comment of @dyfer is indeed valid and shows an example where the default behavior of allowing only one window per name is existing can be tricky (although it is the same old problem of state management all along?) - but by switching this feature on explicitly resolves this issue (at least for me)

are there still remarks that would oppose on merging this as is which means as optional parameter during creation? it allows that we can discuss switching default behavior on a per gui creation basis independently, e.g. `s.meter()`

jamshark70 (changes requested):
Couple of things:

- Declaring `existingWindows` inside an if branch will produce compiler warnings for users who have set the inline warning flag to true, so we try to avoid this in the main class library.
- It's more efficient in this case to use `detect` instead of `select` -- we don't need to create an array here.

```
var existingWindow;

if(reuseExisting) {
    existingWindow = Window.allWindows.detect { ... comparison ... };
    if(existingWindow.notNil) { ^existingWindow };
};
```

capital-G:
>     * Declaring `existingWindows` inside an if branch will produce compiler warnings for users who have set the inline warning flag to true, so we try to avoid this in the main class library.

But why does this create a warning? Doesn't this undermines the idea of scopes in a programming language?
I understand that this could result in declaring a variable only within a branch but as this branch is a function on its own it has its own scope and therefore I can not access variable exclusive to this function outside of it.

I switched from select to detect, thanks for the hint.

jamshark70:
> But why does this create a warning? Doesn't this undermines the idea of scopes in a programming language?

A full discussion of this is well outside of the scope of a PR review.

The requirement to declare variables at the top of a function has been extensively discussed on the user forum. There is interest in changing it (I wouldn't mind if it changed), but I'm not aware of anyone stepping up to do the work yet.

A couple of `if` examples to illustrate:

```
// A. Declare variable within `if` block
bench { 1000000.do { if(true) { var x = 10.rand; x } } }
time to run: 0.65018633099999 seconds.

// B. Declare variable within loop block, and use within `if`
bench { 1000000.do { var x; if(true) { x = 10.rand; x } } }
time to run: 0.11717177500009 seconds.
```

So this test shows an over 5x speed improvement by avoiding a variable declaration within the `if` block.

In A, the variable forces the compiler to treat the branch as an independent closure. The process of evaluating the conditional, then, is 1/ push the Boolean; 2/ push the true Function; 3/ if there's a false Function, push it as well; 4/ do a standard method call for `if` (which creates a new stack frame etc., and will `value` one of the functions, which creates yet another stack frame).

In B, the branch(es) can be rendered inline, so the process is 1/ push the Boolean; 2a/ if false, jump forward over all the byte codes for the true branch (to the false branch, or the next statement); 2b/ if true, step forward into the true branch, and skip over a false branch if one was provided. There are no new stack frames, nothing to unwind -- avoiding the overhead of the standard method call.

I think we don't want to get in the habit of letting `if` in the main class library be 5+ times slower, so we're a bit picky about inlining as much as possible. (Code outside of the main repository is free to do whatever.)

capital-G:
Thanks for the hint @jamshark70 , was not aware of the inner mechanics of sclang :)
Added your recommendation.

adcxyz:
In case it helps, the adclib quark has the WinBounds class,
which remembers preferred window locations by name, 
and can move or remake windows if they were closed.

```
Quarks.install("adclib");
thisProcess.recompile;

// WinBounds first code examples: 
// make 3 windows
Window.closeAll;
~names = [\abc, \bcd, \cde];
~names.do {|name, i| Window(name, Rect(20, i + 3 * 100, 180, 80)).front };

// store their current positions
WinBounds.storeAll;
// ... they are here now:
WinBounds.stored;

Document("WinBounds_stored.scd", "// save?\n" ++ WinBounds.saveString);

// jumble
Window.allWindows.do(_.bounds_(Rect(800.rand, 800.rand, 300, 300)));
// restore one
WinBounds.restore("abc"); // restore one window by name
WinBounds.restoreAll;     // restore all of them

// WinBounds can also be use to store window-making functions

// store a makeFunc
WinBounds.addMake(\WinBounds_Test, { Window("WinBounds_Test", Rect(10, 400, 200, 100)).front });
WinBounds.makeFuncs.cs;
// run it to make the window
WinBounds.make(\WinBounds_Test);
// store the bounds
WinBounds.store(\WinBounds_Test);

// close the test window
Window.find(\WinBounds_Test).close;
// -> showOrMake remakes it
WinBounds.showOrMake(\WinBounds_Test);

/// WinBounds uses some extensions to the Window class :
Window.find(\abc).do(_.front);
Window.find(\nopey).do(_.front);
Window.find(\c).do(_.front);
Window.findAll(\b).do(_.background_(Color.rand));
​
Window.find(\a).moveTo(200).front;
Window.find(\bcd).moveTo(200, 200).front;
Window.findAll(\c).do(_.front);
​
~locs = Window.getAll.postcs;
Window.findAll(\c).do(_.moveTo(800.rand, 800.rand));
Window.setAll(~locs, true);
​
Window.getAllCS;
​
Window("test").front;
Window.show("test").notNil;
```

joshpar:
@capital-G - I see you don't expect this to be merged - is that still the case? If so - can we close this PR?

capital-G:
@joshpar  Actually I think this is still a very useful feature and would love to get this included and move this out of the halt. I use the feature myself since over a year and find it especially useful in live environments and although it may not too helpful for SC veterans it is helpful for newcomers to SC [see #5545 where this originated from]

I always struggle to find out what is the best way to contribute to SC - as there is a discussion about a PR (as it should be and which I find really interesting) it is difficult to moderate a compromise between opposing directions if you don't know the involved authorities [I am used to much more direct terms in contributions such as rejected b/c ... / please fix ... / approved]. As a beginner I don't really see myself in a position to lead a discussion of experts on inner mechanics of SC and it also makes it harder if I don't even know *who* could decide in case of contradictions as it may not be possible to find a solution which pleases everyone.

capital-G:
@telephon @jamshark70 any steps that are missing or I can do to get this merged? :)

telephon (commented):
... sorry, this review was pending, I wasn't aware of that. Posting it now, let me know what you think.

capital-G:
Thanks for your feedback @telephon 

My main motivation behind this PR was to adapt a scheme that I use on my private installation which globally does not allow to spawn multiple windows of the same name at all - this comes handy if you live code and also use GUIs for certain behavior.
I can understand that the behavior to ban multiple windows of the same name may be too radical for the multitude of SC  users but at the same time its always a good time to cut with old ties - therefore I would like to adapt this behavior on a step by step basis [with the associated discussion for each], e.g. `NdefGui` or `s.meter` - the PR would lay a foundation to build upon this.

I was not aware of the inconsistency of `s.meter` and `s.freqscope` [and I ask myself why there is this discrepancy?], yet following the proposed change to adapt the behavior of `s.scope` for `s.meter` [which is based on a `nil check` of a variable in the scope of the `Server` object] looks flawed to me b/c IMO the behavior is not attached to the e.g. the server (meter) but is a behavior of the actual window itself. Therefore the server meter should be able to control the behavior [via args] but should not care about any implementation specifics of a window as this allows for better separation of concerns and more re-usability (see https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

To still allow for multiple server meter windows I would propose [on another PR] to change

* `Server.meter` spawns a window via `ServerMeter.new(... reuseExisting: true)`
* `ServerMeter.new` has arg `reuseExisting: false` and calls `Window(... reuseExisting: reuseExisting)`

Multiple ServerMeters would still be possible for advanced usages but we would it make easier for beginners to not spawn *n* meters in the heat of live coding or learning.

Concerning my motivation for the warning: From my programming experience I would agree to the phrase *avoid doing something implicit* and *if* you "violate" it: do not be silent about it.
From my SC learning experience I found 2 things hard: 1) is the client/server dualism, 2) that there is lots of stuff going on behind the scenes which is really hard to keep track of when you start SC, which often does implicit stuff that you may not wanted.
Other language I use spam you with lots of warnings which I always find very helpful because it gives me a hint which is helpful for novices but not necessary for advanced users.
Also: Due to its inflation it leads to an understanding that warnings should be seen as a guideline which is also allowed to be broken if you know what you are doing.

Regarding the window: `s.meter` should spawn a window as having stateless functions is preferable - doing this a 2nd time is caught and redirected by the proposed behavior and therefore the warning gets shown that not another window was spawned [which is the idea of the stateless `s.meter` function].

E.g. not raising a warning and having activated my "do not spawn multiple windows of the same name" allows me to understand that

```supercollider
s.meter;
-> a ServerMeter
ServerMeter.new(server: s, numIns: 0, numOuts:  0)
WARNING: Window "localhost levels (dBFS)" already exists
-> a ServerMeter
```

may did something which was not my intention and can act accordingly.
I would be fine with removing the warning but would prefer not to in memory of my own sc learning time in which I would have preferred to have such indicators.

telephon:
>I can understand that the behavior to ban multiple windows of the same name may be too radical for the multitude of SC users but at the same time its always a good time to cut with old ties 

Don't misunderstand this: it is very good not to have duplicate windows for things that shouldn't duplicate. The main point is that we should make a difference when the window represents different objects. There is already the pattern that we have singletons for a given server (like `ServerMeter`), but you can have as many as you need (perhaps you are using different screens or embed the views in different guis, etc).

So either the objects need to make sure that the window names are in fact _unique_ for each of these objects, or they just have to keep the window as an instance variable and that one can then be a unique one. The latter case might have less spill.

>Concerning my motivation for the warning: From my programming experience I would agree to the phrase avoid doing something implicit and if you "violate" it: do not be silent about it.

OK, fair enough. But you should use `postln` instead of `warn` if there is nothing wrong. You know what happens if someone shouts "fire!" too often without reason ...

Note that we also have a convention of setting a `verbose = true` or `verbosity = 1` in various classes. This is not a consistent schema across the library – that would be very useful to have, but a bit of work.

capital-G:
> So either the objects need to make sure that the window names are in fact _unique_ for each of these objects, or they just have to keep the window as an instance variable and that one can then be a unique one. The latter case might have less spill.

But then each class has to duplicate the same behavior instead of having a single instance taking care of this. It also gets more difficult when spawning windows dynamically within a `scd` file.
Also it is still possible to deactivate this behavior (as a matter of fact with the proposed change it needs to be explicitly turned on) for each class/window creation.

> Note that we also have a convention of setting a `verbose = true` or `verbosity = 1` in various classes. This is not a consistent schema across the library – that would be very useful to have, but a bit of work.

Good point, this lack lead to something I was thinking about a while, will try to express this in another PR/Issue.
I always try to separate user output and sclang output and only found warnings to be the only way to separate them clearly.

telephon:
> Also it is still possible to deactivate this behavior (as a matter of fact with the proposed change it needs to be explicitly turned on) for each class/window creation.

OK, so let's take a look at the next step. What you want to do with this is that you have a global variable that will make all windows with the same name unique? Or do you want to change the instance variable in each use case? E.g. in `ServerMeter` change 

```supercollider
			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
			false);
```

to

```supercollider
			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
			false, reuseExisting: true);
``` 

And the same for all other cases?

telephon:
Hm, then we would need to guarantee that the servers are named. Two unnamed servers will steal each other's windows:

```
a = Server(addr: NetAddr("127.0.0.1", 57441));
b = Server(addr: NetAddr("127.0.0.1", 57223));

a.meter;
b.meter;
```

It is an explicit feature for servers to be possibly unnamed:

```
name_ { |argName|
		name = argName.asSymbol;
		if(named.at(argName).notNil) {
			"Server name already exists: '%'. Please use a unique name".format(name, argName).warn;
		} {
			named.put(name, this);
		}
	}
```

capital-G:
Why not change the levels window name to

```supercollider
window = Window.new("% (%:%) levels (dBFS)".format(server.name, server.host, server.port),
...
```

jamshark70:
I apologize to say this, but I remain skeptical of this approach.

Let me try another way to say it --

IMO (though I suspect it wouldn't be hard to find professional computer scientists who agree), in general, with rare exception, it makes sense to decouple program logic from GUI interaction. That is, everything that the program is supposed to do should be possible to do by issuing instructions (in SC, method calls). Then the GUI is a shell on top of this, calling methods in the program for user interaction, and receiving notifications for visual updates -- but *not* interfering in program logic.

This principle avoids many kinds of confusion.

The desired program logic here is that the user requests a server meter, and, if one already exists, reuse the old one. The program-logic instruction for this is `s.meter`. If I do this twice, then I get two windows.

So the program logic is faulty (according to the desired behavior being discussed -- although the current behavior isn't inherently unreasonable either).

IMO (and sorry again for stating it bluntly), if GUIs should not *interfere* in program logic, then **it is not `Window`'s job to fix bad program logic.**

The objection that's been raised to this point is "But then each class has to duplicate the same behavior instead of having a single instance taking care of this." Yes, exactly. Program logic should be implemented correctly, according to spec, independent of the GUI. It shouldn't be up to Window to refuse to create a new window; IMO it actually *is* the responsibility of the code requesting the window to avoid requesting duplicates.

IMO it would be better to model server-view singletons with a factory class belonging to a server -- `ServerViewFactory(s)` -- then `s.meter` would do `this.factory.meter` and the factory would know if one already exists or not.

jamshark70:
I should also say that I won't press very hard on this if others disagree with me -- it's fine to disagree! But I do think the principle of "GUI-as-shell" makes the most sense in the most cases and is difficult to refute.

capital-G:
I fully agree that the behavior of a window should only be concern of the Window class and IMO this is what I suggested as the proposed change allows to check if a name of such a Window already exists and instead `.front` the existing window, making it an additional argument when spawning windows and therefore a feature and not a fix. But currently each class needs to keep the state of its windows which intermingles logic, gui and state management not at a single place but at n places.

Well - I am really unsure on how to get the proposed change in and albeit I personally think this feature would benefit SC its probably best just to close the PR? :/

jamshark70:
> I fully agree that the behavior of a window should only be concern of the Window class and IMO this is what I suggested as the proposed change allows to check if a name of such a Window already exists and instead .front the existing window

Again, I might not mind this as a separate method, not modifying `*new`. (In that case, it could even be an extension, and you wouldn't have to ask anyone for permission.)

But then it also occurred to me -- the usual workflow, when creating a window, is that you get the object from Window.new, and then put stuff in the window.

But if you might get an old window instead of a new one, then putting stuff into the window might break, or even completely replace (in the case of `w.layout = ...`), the existing display.

By returning `^existingWindow.front`, the caller has no way to know whether or not it's OK to put things into the returned window -- because in both cases (new or old window), the returned value is a Window -- "I got a window back, but I don't know if it really belongs to me."

Perhaps return nil?

telephon:
Maybe we need a WindowDef class that works a bit like this?

```supercollider
WindowDef : Window {
	classvar <>all;
	*initClass { all = IdentityDictionary.new }
	*new { |name, func|
		var window = all.at(name);
		if(window.isNil) {
			window = super.new;
			all.put(name, window);
		} {
			window.initWindow( window.name, .window.bounds, window.resizable, window.border, window.scroll);
		};
		func.value(window)
		^window
	}
	
}
```

adcxyz:
The WindowDef idea sounds like a good way to go. 
Actually, the WinBounds class (in adclib quark) does pretty much what this PR proposes, 
with a somewhat clumsy code interface: 
```
WinBounds.addMake(name, func); // prepare the make function
WinBounds.showOrMake(name); // show or make it
```
This would be much more elegant with 
```
Wdef(name, func); // declare the make func, and make it immediately (?) 
Wdef.prime(name, func); // only declare the make func, don't make it yet
```
Wdef(name).show; // show or make it
```

**Comment by adcxyz** [comment]
by convention, the make func sets the window name to the Wdef's name, 
so uniqueness just behaves as with OSCdefs etc.

**Comment by capital-G** [comment]
> By returning `^existingWindow.front`, the caller has no way to know whether or not it's OK to put things into the returned window -- because in both cases (new or old window), the returned value is a Window -- "I got a window back, but I don't know if it really belongs to me."

Well, only if you activate the feature that you want to actively overwrite an existing window this would occur, but by default this behavior is set to not overwrite existing or reuse existing windows (`reuseExisting=false`).

> Maybe we need a WindowDef class that works a bit like this?

Yeah, but don't we have already have this behavior mimicked by the proposed change as we simply allow to use the name of a `Window` like a key within a WinDef? I find it more intuitive to use the actual Window name as an Identifier as it mimics my visual identification of the GUI and not introduce an intermediate class which would allow for a divergence of both- or are there usecases where one would want to diverge between the WinDef name and the GUI name?

In case wanting to show 2 ServerMeters of the same server I do not see how a WindowDef class would make a difference from the proposed change?

**Comment by adcxyz** [comment]
FWIW, here is a working sketch for a Wdef class.
```
// instead of s.meter, write:
Wdef(\levels, { s.meter.window });
--- Wdef.sc ---
/*
// A def class for Window

Wdef.all.clear;

Wdef(\x, { Window("anon") });
Wdef(\x).dump
Wdef(\x).hide;
Wdef(\x).show;
Wdef(\x).close;
Wdef(\x).show;

Wdef(\x).bounds = Rect(0,0, 200, 200);
Wdef(\x).close;
Wdef(\x).show;

Wdef(\x, { Window.new.background_(Color.red) });

*/
Wdef {
	classvar <all;
	var <name, <func, <bounds, <window;

	*initClass {
		all = ();
	}

	*new { |name, func, bounds, show = true|
		var found;
		name = name.asSymbol;
		found = all[name];
		if (found.notNil) {
			if (bounds.notNil) {
				found.bounds = bounds;
			};
			if (func.notNil) {
				found.func = func;
			};
			^found
		} {
			^super.newCopyArgs(name, func, bounds).init(show);
		};
	}

	init { |show|
		all.put(name, this);
		if (show) { this.show };
	}

	hasWindow {
		^window.notNil and: { window.isClosed.not }
	}

	func_ { |argFunc|
		if (argFunc.notNil) {
			func = argFunc;
			if (this.hasWindow) {
				window.close;
				this.makeWindow
			}
		}
	}

	bounds_ { |argBounds|
		if (argBounds.notNil) {
			bounds = argBounds;
			if (this.hasWindow) { window.bounds = bounds }
		}
	}

	show {
		if (this.hasWindow) {
			this.bounds_(bounds);
			window.front
		} {
			this.makeWindow
		}
	}

	close {
		if (this.hasWindow) { window.close }
	}

	hide {
		if (this.hasWindow) { window.minimize }
	}

	makeWindow {
		window = func.value(this);
		if (this.hasWindow) {
			window.name = name;
			this.bounds_(bounds);
			bounds = bounds ? window.bounds;
			window.front
		}
	}
}
---
```

capital-G:
BUMP

So is there a chance that this could get merged if it gets written as a `WindowDef`? It was proposed by the developers of Jitlib but I know people who don't like the `def` approach.

Also the main idea motivation of this PR was a user complain about spawning too many server meters by accident, so the default behavior of spawning such a window should be adapted.
So its not about adding a `WindowDef` class but adapting it throughout the standard library as well.

I know that there are exotic setups where e.g. many ServerMeters are needed, but I think this is only relevant to advanced users that would know how to still spawn multiple windows (by e.g. documenting how to deviate from this default behavior) - this is more intended for the "newcomers" to SC.

jamshark70:
> So is there a chance that this could get merged if it gets written as a WindowDef? It was proposed by the developers of Jitlib but I know people who don't like the def approach.

I was skeptical of the original proposal, but I would be okay with WindowDef (because 1/ the user has control over the matching key and 2/ the matching key is independent of the display).

> So its not about adding a WindowDef class but adapting it throughout the standard library as well.

I would say, using WindowDef *where appropriate* in the standard library.

To be honest, though, if the rule is going to be one server : one meter, IMO ServerMeter is responsible for this logic. It isn't the job of the GUI system to compensate for the underlying class failing to control its instances properly.

Program flow A: User requests a second ServerMeter; SC gives back the existing ServerMeter and doesn't open new windows.

Program flow B: User requests a second ServerMeter, which starts to initialize but then fails to get a new window, whereupon it has to backtrack and return... something... from some other set of stored instances... which, if you have stored instances to be able to return one back to the user, then you probably should have checked this before trying to open a window.

A is direct, to the point, and ultimately will be easier to maintain.

GUI should *reflect* a well functioning system, not be a central point of influence in the fundamental logic.

I realize that we might have to agree to disagree here -- but, I haven't changed my mind that Window (or WindowDef) is not the right place to control ServerMeter instances.

dyfer:
Thanks for your work on this @capital-G and thanks to @telephon @jamshark70 @adcxyz for the review. In the light of the discussion that followed, I don't believe that there is a consensus that would allow this PR to move forward. If that is the case, please feel free to close this.

PR #5559, opened 2021-09-01
replace aiff with wav as the default value for recHeaderFormat

### LINKED ISSUE #5557: Switch away from using AIFF as default recHeaderFormat
## Motivation

When installing SuperCollider from fresh, it will by deffault recording the server output in the .aiff format unless changed ala `s.recHeaderFormat_("wav")` in a startup file. 

AIFF is a more MacOS centric format and it just doesn't work as well on Linux for example when converting files, importing/exporting from other programs etc. On Linux (and Windows?) users have to intervene to change this default to remedy the situation and I think it should be the other way around - if you want a more platform specific suffix/headerformat it should be something demanding intervention - just my 2 cents.

Change the default to "wav"

## Plan for Implementation

Change this line:
https://github.com/supercollider/supercollider/blob/e4743da23948f6ce1cfe7750273814d0f489cc20/SCClassLibrary/Common/Control/Server.sc#L98

RhnSharma:
## Purpose and Motivation

Fixes #5557

## Types of changes
Enhancement 

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

Hi @madskjeldgaard, how does this look?
Let me know if this needs any changes.
Thanks

### DIFF
--- a/HelpSource/Classes/DiskOut.schelp
+++ b/HelpSource/Classes/DiskOut.schelp
@@ -82,7 +82,7 @@ x = Synth("bubbles");
 b= Buffer.alloc(s, 65536, 2);
 
 // create an output file for this buffer, leave it open
-b.write("~/diskouttest.aiff".standardizePath, "aiff", "int16", 0, 0, true);
+b.write("~/diskouttest.wav".standardizePath, "wav", "int16", 0, 0, true);
 // create the diskout node; making sure it comes after the source
 d = Synth.tail(nil, "help-Diskout", ["bufnum", b]);
 // stop recording
@@ -99,7 +99,7 @@ b.free;
 x = Synth.basicNew("help-Diskin-2chan");
 m = { |buf| x.addToHeadMsg(nil, [\bufnum,buf])};
 
-b = Buffer.cueSoundFile(s,"~/diskouttest.aiff".standardizePath, 0, 2, completionMessage: m);
+b = Buffer.cueSoundFile(s,"~/diskouttest.wav".standardizePath, 0, 2, completionMessage: m);
 )
 x.free; b.close; b.free; // cleanup
 ::
@@ -118,7 +118,7 @@ s.sendMsg("/s_new", "bubbles", 2003, 1, 1);
 s.sendMsg("/b_alloc", 0, 65536, 2); // Buffer number is 0
 
 // create an output file for this buffer, leave it open
-s.sendMsg("/b_write", 0, "~/diskouttest.aiff".standardizePath, "aiff", "int16", 0, 0, 1);
+s.sendMsg("/b_write", 0, "~/diskouttest.wav".standardizePath, "wav", "int16", 0, 0, 1);
 
 // create the diskout node
 s.sendMsg("/s_new", "help-Diskout", 2004, 3, 2003, "bufnum", 0);
--- a/HelpSource/Classes/Recorder.schelp
+++ b/HelpSource/Classes/Recorder.schelp
@@ -115,7 +115,7 @@ method::numChannels
 a number of sound file channels that is used always when using this recorder, unless a different one is specified in the link::#-record:: method. When not set, we use link::Classes/Server#-recChannels::.
 
 method:: recHeaderFormat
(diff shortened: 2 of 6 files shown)


dyfer:
Thanks @RhnSharma 
I personally don't oppose this change, but I think it would be good to get input from multiple devs whether this is desirable.
As for the PR itself, the documentation in `Server.schelp` (maybe elsewhere as well? haven't looked closely) should also be updated.

telephon:
As far as I am concerned, it is ok. I am not sure about possible downsides of `wav` though. I think there used to be a difference in the way headers could be extended, but I don't know.

dyfer:
I've just realized: what about files >4GB? Currently, AFAICT, `wav`'s data is being written above that level, but the header does not support that size (some programs will read the whole file, some will only read the size up to what's specified in the header). I'm not sure what's the state of supporting RF64/BWF to get around 4GB limit with `wav` files. I think the 4GB limit is something we should consider when making changes to the default format.

jamshark70:
> FWIW the 4GB limit might also apply to AIFF

https://manual.audacityteam.org/man/export_formats_supported_by_audacity.html :

"Size limits for WAV and AIFF PCM-encoded files
"WAV and AIFF files are limited to a maximum size of 4GB (this is a general restriction and not an Audacity one)."

So we don't lose anything by switching to WAV.

I think this change is fine esp. for the benefit of Windows users. (FWIW, #5557 reports problems opening AIFF files in Linux, which I have never experienced, so I found that a puzzling remark.)

madskjeldgaard:
> I think this change is fine esp. for the benefit of Windows users. (FWIW, #5557 reports problems opening AIFF files in Linux, which I have never experienced, so I found that a puzzling remark.)

This is generally true - sorry, I should have specified (was in a hurry when reporting - bad idea). The only program really that struggles with this is `lame`, it simply refuses to use aiff. Of the programs I use daily - `sox` and Reaper seem ok with AIFF for example. I don't use audacity and ardour so I don't know how they behave. Still, I think the argumentation holds - that having something more universal than an apple format as the default would be cool

jamshark70:
> The only program really that struggles with this is lame, it simply refuses to use aiff.

FWIW I couldn't reproduce this: `lame -V5 SC_210703_134352.aiff SC_210703_134352.mp3` produced a valid mp3 file.

But again, I don't mind the change for the sake of Windows players.

It's more common to find audio apps that don't like 32-bit float samples, but we shouldn't change that -- floats prevent irreversible clipping.

dyfer:
If we are to move ahead with this change, the documentation should be updated as well.

dyfer:
To clarify: I'm now also OK with this change, but the PR needs to have the documentation updated, as indicated earlier.

smrg-lm:
WAV is a simpler format and don't have extended 80 bit IEEE Standard 754 floating point number just for the sample rate... Which is a technological mystery for me.

joshpar (changes requested):
Thanks! Can you please add changes for documentation?

dyfer:
Hello @RhnSharma 
Would you be able to add documentation changes to this?

dyfer:
Okay, I'm going to push documentation changes for this PR

joshpar (approved):

dyfer:
Please note:
- I did make changes to the `Non-Realtime Synthesis` guide replacing `aiff` with `wav`
- `Buffer -write` still uses `aiff` as the default format; as this was not the objective of this PR, I did not change that
- `ProxySpace` recording uses `aiff` by default (present in the documentation in the main repo, but the classes are in the `JITLibExtensions` quark; again, I haven't touched these

madskjeldgaard:
> Please note:
> - I did make changes to the `Non-Realtime Synthesis` guide replacing `aiff` with `wav`
> - `Buffer -write` still uses `aiff` as the default format; as this was not the objective of this PR, I did not change that
> - `ProxySpace` recording uses `aiff` by default (present in the documentation in the main repo, but the classes are in the `JITLibExtensions` quark; again, I haven't touched these

Thanks a bunch Marcin !

dyfer:
All right, seems like there are no objections, I'll merge this. Thanks!

Issue #5568, opened 2021-09-07
Server.remote does not work as documented: raises an error when options is nil


dreinsch:
## Environment

* SuperCollider version: 3.12 (since https://github.com/supercollider/supercollider/commit/c1eb86075c77293c29b9cb6020f06d7e5132f9fb)
* Operating system: should occur on all systems, tested on Windows 10
* Other details (Qt version, audio driver, etc.): -

## Steps to reproduce

```supercollider
~remote = Server.remote('remote')
```

## Expected vs. actual behavior

expected: remote server gets created even without providing `options`. 
actual: Method raises an error as `options` is nil and has no message `protocol`

```
ERROR: Message 'protocol' not understood.
RECEIVER:
   nil
ARGS:
CALL STACK:
	DoesNotUnderstandError:reportError
		arg this = <instance of DoesNotUnderstandError>
	Nil:handleError
		arg this = nil
		arg error = <instance of DoesNotUnderstandError>
	Thread:handleError
		arg this = <instance of Thread>
		arg error = <instance of DoesNotUnderstandError>
	Object:throw
		arg this = <instance of DoesNotUnderstandError>
	Object:doesNotUnderstand
		arg this = nil
		arg selector = 'protocol'
		arg args = [*0]
	Meta_Server:remote
		arg this = <instance of Meta_Server>
		arg name = 'sc3nb_remote'
		arg addr = nil
		arg options = nil
		arg clientID = nil
		var result = <instance of Server>
	< closed FunctionDef >  (no arguments or variables)
	Interpreter:interpretPrintCmdLine
		arg this = <instance of Interpreter>
		var res = nil
		var func = <instance of Function>
		var code = "~remote = Server.remote('sc3..."
		var doc = nil
		var ideClass = <instance of Meta_ScIDE>
	Process:interpretPrintCmdLine
		arg this = <instance of Main>
^^ ERROR: Message 'protocol' not understood.
RECEIVER: nil

```

solution: create default instance of ServerOptions like described in the [Server.remote docs](https://doc.sccode.org/Classes/Server.html#*remote)

This currently causes a issue in sc3nb https://github.com/interactive-sonification/sc3nb/issues/15. A workaround would be creating the default ServerOptions and providing them to `Server.remote`

Thank you very much for developing SuperCollider!

jamshark70:
The `options` are not the problem here.

It is mandatory to supply a NetAddr to locate the remote server. Your example doesn't -- hence, error.

I think, not a bug.

[...]

Issue #3310, opened 2017-11-24
Proposal: Split duties between Server and ServerProcess


adcxyz:
In the discussion on #3286 and complicated PR #3299, 
@jamshark70 had the insight that ServerStatusWatcher should actually be responsible for handling the entire admin for the server process: booting, watching status, and quitting. 

I have started to explore this idea in a branch: 
https://github.com/adcxyz/supercollider/tree/server_ServerProcess

This is a massive refactoring of Server, ServerStatusWatcher based on #3299, 
and looks promising so far - not at all finished yet, so not a PR yet.

Now the ServerProcess class replaces ServerStatusWatcher 
and handles boot, quit, and running status of a server process.
The status code is copied and adapted from ServerStatusWatcher,
boot and quit-related methods are copied and refactored from Server.

The Server class now focuses only on 'content': allocation, synths, controls, etc.

Disclaimers : 
* This shifts a lot of code around in ways that github cannot track well
* A lot of things are renamed for clarity.   I made an effort to support 
much of the earlier code interface, but this is not complete at all. 
---
```
*** Pluses:
* all boot steps are in a single boot routine
* logic is a lot clearer and easier to read
* uses conditions to speed up booting and trigger waiting tasks
 
//////// WORKING ALREADY : ///////////
// These tests all run:
TestServer_boot.run;
TestServer_clientID.run;
TestServer_clientID_booted.run;

// * ca. 40% faster booting on macOS
s.boot;
s.quit;

// * boot has onComplete and onFailure arguments:
s.boot(onComplete: { "READY".postln; }, onFailure: { "failed".postln; }, timeout: 5);

// * most standard boot variants work
s.reboot;
s.quit;
s.waitForBoot { Env.perc.test };

// * remote server works:
// boot scsynth by hand
Server.killAll;
unixCmd( (Server.program + s.options.asOptionsString).postcs);
r = Server.remote(\remote);

// * test for missing scsynth works:
Server.program = "noValidProgramName";
s.quit;
s.boot; // fails very quickly

//////// TODO: /////////////////

Server.trace; // enables extensive posts on all boot steps for debugging 
// * this complains ...
s.quit.boot;

// * UnitTests for every if-branch in boot method!

// * fix remote method
r = Server.remote(\remote);

// * Volume sendSynthDef actions accumulates with every boot
(see post window after a few boots)

// * this should run into timeout I think
Server.program = "sleep 10";
s.quit;
s.boot(timeout: 5); // fails very quickly

[...]

PR #5569, opened 2021-09-09
class library: refactor functionality: connectToServerAddr

telephon:
## Purpose and Motivation

While looking for a fix of #5568, a small refactoring was recommended.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Control/Server.sc
+++ b/SCClassLibrary/Common/Control/Server.sc
@@ -389,14 +389,7 @@ Server {
 	}
 
 	*remote { |name, addr, options, clientID|
-		var result;
-		result = this.new(name, addr, options, clientID);
-		if(options.protocol == \tcp) {
-			addr.tryConnectTCP({ result.startAliveThread }, nil, 20)
-		} {
-			result.startAliveThread;
-		};
-		^result;
+		^this.new(name, addr, options, clientID).connectToServerAddr({ this.startAliveThread })
 	}
 
 	init { |argName, argAddr, argOptions, argClientID|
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks so much! 
(especially that the linked issue was introduced by my commit...)
- Maybe it would be good to add the documentation for the new method in the help...?
- Do you think this should be cherry-picked for 3.12 or is it fine to stay on develop?

telephon:
I'm still usure whether `connectToServerAddr` is the best name for the method – it seems misleading when taken out of context. Whe only connect if necessary (for tcp), otherwise we directly call the onComplete message.

In fact, I would have thought it good to pack the `startAliveThread` into it, but the method `boot` passes this only when the argument `startAliveThread = true` (sorry, this is all a bit complicated to explain).

dyfer:
> In fact, I would have thought it good to pack the `startAliveThread` into it, but the method `boot` passes this only when the argument `startAliveThread = true` (sorry, this is all a bit complicated to explain).

Yeah... the whole `Server` class and boot behavior was mean to be rewritten, but it seems it never happened. Not sure how much we want to change before the (hopefully) rewrite later one, and how much we should resort to minimal patching.

telephon:
@adcxyz did a rather large rewrite long ago, but as far as I know, it somehow was never taken up and merged.

adcxyz:
> @adcxyz did a rather large rewrite long ago, but as far as I know, it somehow was never taken up and merged.

Yes, that was discussed here : https://github.com/supercollider/supercollider/issues/3310
and the related fork that sketched this out: https://github.com/adcxyz/supercollider/tree/server_ServerProcess

telephon:
Ah good, thanks for pointing to this. As you wrote, the branches have diverged. What do you think: is the current PR doing harm (I can minimize impact at the cost of beauty :))

telephon:
what do you think?

adcxyz (approved):
looks good to me, clearer than before.

PR #5580, opened 2021-09-27
fixes #5579

### LINKED ISSUE #5579: /g_head is not sending corresponding /n_move notification
## Environment

* SuperCollider version: 3.12.1 (also 3.10.2)
* Operating system: Debian 10

## Steps to reproduce

```supercollider
s.quit;
s.boot;
r = RootNode(s);
r.dumpTree;
/*
NODE TREE Group 0
   1 group
*/
s.dumpOSC;
(
f = { |msg, time, addr|
    if(msg[0] != '/status.reply') {
        "r: %\n".postf(msg);
    }
};
thisProcess.addOSCRecvFunc(f);
);
a = Group.tail(r); // r: [ /n_go, 1000, 0, 1, -1, 1, -1, -1 ]
b = Group.head(r); // r: [ /n_go, 1001, 0, -1, 1, 1, -1, -1 ]
/*
NODE TREE Group 0
   1001 group
   1 group
   1000 group
*/
a.moveToHead(r);   // NO RECEPTION
r.dumpTree;        // ... despite updated tree:
/*
NODE TREE Group 0
   1000 group
   1001 group
   1 group
*/
a.moveToTail(r);   // r: [ /n_move, 1000, 0, 1, -1, 1, -1, -1 ]
```

The `/g_head` command should be followed by the notication `r: [ /n_move, 1000, 0, -1, 1, 1, -1, -1 ]`.

Sciss:
- g_head should always fire an n_move reply,
  even if the parent group remains the same.
  This is in line with the other commands such
  as g_tail, and ensures that a client can build
  a copy of the tree structure by observing the
  server's n_move notifications.

## Purpose and Motivation

fixes #5579

## Types of changes

- Bug fix

## To-do list

- [X] Code is tested
- [ ] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/server/scsynth/SC_MiscCmds.cpp
+++ b/server/scsynth/SC_MiscCmds.cpp
@@ -1213,15 +1213,14 @@ SCErr meth_g_head(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRe
         if (!node)
             return kSCErr_NodeNotFound;
 
-        Group* prevGroup = node->mParent;
+        // Group* prevGroup = node->mParent;
 
         Node_Remove(node);
-
         Group_AddHead(group, node);
 
-        if (group != prevGroup) {
-            Node_StateMsg(node, kNode_Move);
-        }
+        // if (group != prevGroup) {
+        Node_StateMsg(node, kNode_Move);
+        // }
     }
     return kSCErr_None;
 }
@@ -1245,7 +1244,7 @@ SCErr meth_g_tail(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRe
 
         // if (group != prevGroup) {
         Node_StateMsg(node, kNode_Move);
-        //}
+        // }
     }
     return kSCErr_None;
 }
--- a/server/scsynth/SC_MiscCmds.cpp
+++ b/server/scsynth/SC_MiscCmds.cpp
@@ -1213,14 +1213,10 @@ SCErr meth_g_head(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRe
         if (!node)
             return kSCErr_NodeNotFound;
 
-        // Group* prevGroup = node->mParent;
-
         Node_Remove(node);
         Group_AddHead(group, node);
 
-        // if (group != prevGroup) {
         Node_StateMsg(node, kNode_Move);
-        // }
     }
     return kSCErr_None;
 }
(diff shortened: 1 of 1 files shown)


Sciss:
(there is one CI test failure, which AFAIK is not related to this PR)

telephon:
Hm, looking at other commands like `meth_n_before`, this will also only send if something has changed. I think that was the original idea of the interface. 

Probably we need to keep this in a uniform manner across the whole interface.

Sciss:
The point is, the server should send a notification if _something_ changed. Of course it can opt to not send a message, if the node remains in the same location. But comparing the parent is insufficient to determine change for `g_head` and `g_tail`. Right now the two behave inconsistently, and the reason for changing `g_tail` was that it was omitting information to the clients. Therefore, fixing `g_head` makes the situation consistent and symmetric, irrespective of whether `n_after` and `n_before` need adapation as well. This could be a second fix are carefully checking the situation. I think this PR is very straight forward.

Sciss:
Julian, which version are you looking at? Current `meth_n_before` _always_ notifies client:

```cpp
            Node_Remove(prevNode);
            Node_AddBefore(prevNode, prevTarget);
            Node_StateMsg(prevNode, kNode_Move);  // always sent
```

telephon:
Ah ok, I see, I was looking at what you had already changed.

telephon (approved):
It probably was an oversight indeed.

dyfer:
Thank you @Sciss, this is great.
Before merging, I'd like to consider whether we really want to leave the commented out code in?
In any case I don't think that [this whitespace change](https://github.com/supercollider/supercollider/pull/5580/files#diff-288168ff6fe75140f3bb6e1f098394e04d5a4580b59613cd211928012b5394b4L1248) should be a part of this PR.
If we decide that we remove the checks instead of commenting them out, we could also - optionally! - consider removing the [already commented out check](https://github.com/supercollider/supercollider/pull/5580/files#diff-288168ff6fe75140f3bb6e1f098394e04d5a4580b59613cd211928012b5394b4L1246) as part of this PR.

Sciss:
either way; I just think it would be good to have g_head and g_tail symmetric; feel free to remove comments in both cases.

dyfer:
All right @Sciss @telephon should I push a change to the fix branch to remove unused/commented out code?

Sciss:
yes, please; thank you!

dyfer:
Are we good to merge this after the commit I added?

dyfer:
Thanks!

PR #5601, opened 2021-10-19
Plugins: Fix AudioControl and InFeedback processing for an extra cycle

vitreo12:
## Purpose and Motivation

I have noticed a weird behavior of the `AudioControl` and `InFeedback` UGens.

If you run this code, you'll notice that the last audio cycle will run twice, generating a "double" dropout:

```SuperCollider
(
s.waitForBoot({
	SynthDef(\test, {
		Out.ar(\out.ir(0), { SinOsc.ar(2) * EnvGen.ar(Env([1, 1], [0.2]), doneAction:2) })
	}).add;

	s.sync;

	~bus = Bus.audio(s);

	Pbind(
		\instrument, \test,
		\dur, 1,
		\out, ~bus.index,
		\addAction, \addToHead
	).play;

	~reader = { \in.ar }.play(args:[ \in, 'a' ++ ~bus.index]);

	{ In.ar(0) }.plot(0.5);
})
)
```

image

This happens because the `AudioControl` UGen (as well as the `InFeedback` one) only check if the bus has been touched in the current audio cycle. In my opinion, it should probably take into consideration if the bus has been touched in the previous audio cycle aswell, in order to avoid this "repetition" effect.

This PR fixes this issue. It basically allows `AudioControl` and `InFeedback` to work just like `In` when the `~reader` `Synth` is later in the chain, and just like `InFeedback` when the `~reader` `Synth` is before in the chain (reading the previous audio cycle).

`Pbind (\addToHead) -> ~reader:`
image (1)

`-> ~reader Pbind (\addToTail) ->:`
image (2)

## Types of changes

- Bug fix

## To-do list

- [X] Code is tested
- [X] All tests are passing
- [X] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestIOUGens.sc
+++ b/testsuite/classlibrary/TestIOUGens.sc
@@ -22,7 +22,7 @@ TestIOUGens : UnitTest {
             var inFeedback = InFeedback.ar(bus.index);
             inFeedback - in;
         }.loadToFloatArray(0.5, server, { | data |
-            this.assertFloatEquals(data.sum, 0.0, "InFeedback is not equal to In", report:true);
+            this.assertFloatEquals(data.sum, 0.0, "InFeedback should be equal to In when no other signal has been written to its bus", report:true);
             condition.unhang;
         });
 
@@ -30,7 +30,7 @@ TestIOUGens : UnitTest {
     }
 
     //Test that AudioControl is equal to In when write / read are in the right order
-    test_AudioControl_equals_In {
+    test_AudioControl_equals_In_for_read_after_write {
         var condition = Condition();
         var bus = Bus.audio(server);
         var sumBus = Bus.audio(server);
@@ -56,7 +56,7 @@ TestIOUGens : UnitTest {
 
         server.bind({
             sum.loadToFloatArray(0.5, server, { | data |
-                this.assertFloatEquals(data.sum, 0.0, "AudioControl is not equal to In", report:true);
+                this.assertFloatEquals(data.sum, 0.0, "AudioControl should be equal to In for read after write", report:true);
                 condition.unhang;
             });
 			in.play(server);
@@ -68,7 +68,7 @@ TestIOUGens : UnitTest {
     }
 
     //Test that InFeedback is equal to In + a delay of BlockSize when write/read are one cycle apart
-    test_InFeedback_equals_delayed_In {
+    test_InFeedback_equals_In_for_write_after_read {
         var condition = Condition();
         var bus = Bus.audio(server);
 
@@ -79,7 +79,7 @@ TestIOUGens : UnitTest {
             var delayedIn = DelayN.ar(In.ar(bus.index), blockSizeMs, blockSizeMs);
             inFeedback - delayedIn;
         }.loadToFloatArray(0.5, server, { | data |
-            this.assertFloatEquals(data.sum, 0.0, "InFeedback is not equal to delayed In", report:true);
+            this.assertFloatEquals(data.sum, 0.0, "InFeedback should be equal to In + a delay of BlockSize when write/read are one cycle apart", report:true);
             condition.unhang;
         });
 
@@ -87,7 +87,7 @@ TestIOUGens : UnitTest {
     }
 
     //Test that AudioControl is equal to In + a delay of BlockSize when write/read are one cycle apart
-    test_AudioControl_equals_delayed_In {
+    test_AudioControl_equals_In_for_write_after_read {
         var condition = Condition();
         var bus = Bus.audio(server);
         var sumBus = Bus.audio(server);
@@ -114,7 +114,7 @@ TestIOUGens : UnitTest {
 
         server.bind({
             sum.loadToFloatArray(0.5, server, { | data |
-                this.assertFloatEquals(data.sum, 0.0, "AudioControl is not equal to delayed In", report:true);
+                this.assertFloatEquals(data.sum, 0.0, "AudioControl should be equal to In + a delay of BlockSize when write/read are one cycle apart", report:true);
                 condition.unhang;
             });
 			delayedIn.play(server);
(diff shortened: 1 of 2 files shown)


vitreo12:
Sure, I just added some tests. Let me know if they are enough :)

telephon:
I think these failing tests are just timeouts, @vitreo12. It would be good to get this working, perhaps you can see if there is something in your tests that causes it?

vitreo12:
Hi @telephon! All the tests I have implemented complete on my local Linux build, I don't know why they would hang on MacOS. Could anyone with MacOS maybe try them out and report what is going wrong? 

The test class is called `TestIOUGens` and the functions are: `test_InFeedback_equals_In`, `test_AudioControl_equals_In`, `test_InFeedback_equals_delayed_In` and `test_AudioControl_equals_delayed_In`.

dyfer (changes requested):
Hello @vitreo12 
I have a request regarding the tests: it seems that they use `Silence.ar`, which does not exist in the standard class library. Could you replace it with `Silent.ar`?

dyfer:
@telephon @joshpar aside from the tests, do the changes in this PR look good to you?

vitreo12:
> Hello @vitreo12 I have a request regarding the tests: it seems that they use `Silence.ar`, which does not exist in the standard class library. Could you replace it with `Silent.ar`?

Sure, just pushed in the changes.

testsuite/classlibrary/TestIOUGens.sc
```diff
@@ -0,0 +1,127 @@
...
+        var condition = Condition();
+        var bus = Bus.audio(server);
+
+        {
+            var writer = Out.ar(bus.index, SinOsc.ar(1));
+            var in = In.ar(bus.index);
+            var inFeedback = InFeedback.ar(bus.index);
+            inFeedback - in;
+        }.loadToFloatArray(0.5, server, { | data |
+            this.assertFloatEquals(data.sum, 0.0, "InFeedback is not equal to In", report:true);
```
telephon, line 25: The report should be clearer, something like: "InFeedback should be equal to In when no other signal has been written to its bus"

testsuite/classlibrary/TestIOUGens.sc
```diff
@@ -0,0 +1,127 @@
...
+            Silent.ar;
+        };
+
+        var sum = {
+            In.ar(sumBus.index);
+        };
+
+        server.bind({
+            sum.loadToFloatArray(0.5, server, { | data |
+                this.assertFloatEquals(data.sum, 0.0, "AudioControl is not equal to In", report:true);
```
telephon, line 59: I'd suggest the message to say: "AudioControl should be equal to In for read after write".

testsuite/classlibrary/TestIOUGens.sc
```diff
@@ -0,0 +1,127 @@
...
+        }.loadToFloatArray(0.5, server, { | data |
+            this.assertFloatEquals(data.sum, 0.0, "InFeedback is not equal to delayed In", report:true);
+            condition.unhang;
+        });
+
+        condition.hang;
+    }
+
+    //Test that AudioControl is equal to In + a delay of BlockSize when write/read are one cycle apart
+    test_AudioControl_equals_delayed_In {
```
telephon, line 90: suggestion: `test_AudioControl_equals_In_for_write_after_read`

telephon (commented):
I see only some changes to be done in the naming of tests and messages. This is a very important fix, thanks a lot.

smrg-lm:
**Please review if this change is necessary** or something else is failing because the given example is not good, it works as expected if the `test` ugen doesn't define a control input for the output bus, try replacing it by:

```
	SynthDef(\test, {
		Out.ar(0, SinOsc.ar(2) * EnvGen.ar(Env([1, 1], [0.2]), doneAction:2))
	}).add;
```

vitreo12:
> **Please review if this change is necessary** or something else is failing because the given example is not good, it works as expected if the `test` ugen doesn't define a control input for the output bus, try replacing it by:
> 
> ```
> 	SynthDef(\test, {
> 		Out.ar(0, SinOsc.ar(2) * EnvGen.ar(Env([1, 1], [0.2]), doneAction:2))
> 	}).add;
> ```

That's because `AudioControl` or `InFeedback` are used for inner-Bus communication. With your code, you're just writing directly to the audio output.

smrg-lm:
@vitreo12 I was wrong testing and the alarms sounded, apologies.

vitreo12:
@telephon I have changed the names accordingly :)

dyfer (approved):

dyfer:
Tests look good do me now.
@telephon feel free to merge if you're happy with the PR otherwise.

smrg-lm:
Apologies for being annoying, I was worried because the use case is wrong, because the order of execution, and it makes the issue smell, but it smells good so far. I hope I'm not mistaken again: this change creates a gap in the audio in the following situation which is maybe better than what was happening before.

```
(
SynthDef('source', { arg out, freq=220;
	Out.ar(out, SinOsc.ar(freq));
}).add;

SynthDef('input', { arg a_in;
	Out.ar(0, a_in);
}).add;
)

b = Bus.audio();
x = Synth('source', [out: b, freq: 440]);
y = Synth('input', [a_in: b.asMap]);
z = Synth('source', [out: b]);

z.free;  // one block gap before x writes again.
```
After:

despues

Before:

antes

I'm not sure what could be the desired behavior in this particular case but the feedback was not intended to be used like this anyway.

telephon:
That is a good point. It is actually a bit tricky what the best behaviour should be.

Possibiities:

1. all audio is read with time delay (this accumulates in multiple routing)
2. all audio is read with time delay only if there is at least one writer downbus, otherwise without (this may cause a discontinuity if the downbus writer stops)
3. the audio from upbus is read with no time delay and the audio from downbus is read with time delay
4. once the inFeedback has received its flag that there are writers downstream, it should continue to play delayed, irrespective of whether the downstream writers ended.

joshpar:
As noted elsewhere, there likely isn't a way to satisfy ALL cases, but I do think this satisfies ones that make more sense than my initial implementation of this. I'm good approving it, but will wait until this weekend to merge in case @telephon or @dyfer feel like anything else can still be resolved.

joshpar (approved):

telephon (approved):

Issue #5488, opened 2021-06-25
unit commands executed before unit initialized


Spacechild1:
## Environment

* SuperCollider version: SC 3.11

## Description

When the Server dispatches a `/s_new` message, the units for the Synth are allocated, but the constructors are not called yet.
If the Server subsequently dispatches a `/u_cmd` message for one of the units, the unit command is invoked immediately, although the unit hasn't been fully initialized, leading to undefined behavior.

In the user-registered unit command method there's no direct way to know whether the unit has been initialized or not.

I hit this problem in VSTPlugin and my solution has been to
1) (ab)use `specialIndex` by setting it to a magic number in the constructor
2) temporarily store unit commands inside the unit if the constructor hasn't run yet (by checking `specialIndex`)
3) in the constructor, check for pending unit commands and run them.

I think this really should be dealt with in the Server instead.

PR #5610, opened 2021-11-12
Fix possible crash with unit commands

Spacechild1:
## Purpose and Motivation

We want to be able to send unit commands right after Synth creation (at the same time stamps), but with the current naive implementation (both scsynth and supernova) the unit command will be sent before the unit constructor has run. This will always lead to undesirable behavior and it can easily crash, e.g. when accessing an uninitialized pointer member. There are ways to circumvent this issue (see VSTPlugin) but they are ugly and error-prone.

Fixes https://github.com/supercollider/supercollider/issues/5488

## Changes

When we receive a unit command, we first check if the Graph has been prepared (i.e. all Unit constructors have run).
If yes, we can execute it immediately (= current behavior); if not, we put the unit command on a linked list inside the Graph and only execute it after the Graph has been prepared.

I have added the opaque member `void* mPrivate` to the *end* of the `Graph` struct. This naturally changes the struct *size* but does not change the struct *layout*. Since plugins generally do not care about the size of the `Graph` struct, this should be safe. This means it is not necessary to bump the plugin API version.

I have also added an example plugin `UnitCmdDemo` which demonstrates the usage of Unit commands, similar to the existing plugin command demo in `UIUGens.cpp`. Since they are not UI UGens, I have moved both to a dedicated new file `DemoUGens.cpp`.

You can use this plugin to test the code:

```
SynthDef(\u_cmd_test, { UnitCmdDemo.kr }).add;

u = Synth(\u_cmd_test, nil, s);
// unit command will be dispatch immediately
s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.9); // will print the value
u.free;

(
// create a Synth and send unit commands at the same timestamp.
// the unit commands will be queued (= the new behavior).
s.bind {
	u = Synth(\u_cmd_test, nil, s);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.9);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.2);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.5);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.1);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.4);
}
)

(
// special case: the Synth is destroyed at the same timestamps,
// so the unit commands will be queued, but never executed.
// they are, however, correctly freed in the Graph destructor.
s.bind {
	u = Synth(\u_cmd_test, nil, s);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.9);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.2);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.5);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.1);
	s.sendMsg(\u_cmd, u.nodeID, 0, \setValue, 0.4);
	u.free;
}
)
```

I also discovered a little bug in `Graph_Dtor` (missing check for `Graph_NullFirstCalc`).

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer:
Thanks @Spacechild1 !

I appreciate all the comments and information in this PR and contributed code. Unfortunately, I likely won't be able to understand all aspects of it...

Having said that:
- I do want to merge this, but I'd like to have an extra pair of eyes on this before we approve it
- I am nervous about touching the `Graph` struct. Your explanation helps and I can see this build still works with existing plugins, but again I'd like at least one more person to take a look to make sure we are not missing something w/r/t compatibility and plugin API
- Finally, after building this and trying to create a SynthDef with `UnitCmdDemo.kr`, I am getting `exception in GraphDef_Recv: UGen 'UnitCmdDemo' not installed.` Looking into the source, it seems that on macOS `UIUGens.mm` [is compiled instead of](https://github.com/Spacechild1/supercollider/blob/unit_cmd_fix/server/plugins/CMakeLists.txt#L99) `UIUGens.cpp`... Not sure how best to solve this, but that seems to be the reason the demo plugin is not present in my (macOS) build.

Spacechild1:
> I'd like at least one more person to take a look to make sure we are not missing something w/r/t compatibility and plugin API

I think @scztt would be the right person for this.

> Looking into the source, it seems that on macOS UIUGens.mm is compiled instead of UIUGens.cpp

Thanks, I didn't notice this because I only tested on Windows and Linux.

Actually, it doesn't really make sense that `UnitCmdDemo` should be in `UIUGens.cpp` resp. `UIUGens.mm`. The only reason I put it there is because `pluginCmdDemo` is already there (for whatever reason). I think I could just move both `UnitCmdDemo` and `pluginCmdDemo` into a new dedicated file. Maybe something like `DemoUGens.cpp`?

dyfer:
> I think I could just move both `UnitCmdDemo` and `pluginCmdDemo` into a new dedicated file. Maybe something like `DemoUGens.cpp`?

Yes, I like the idea of putting duplicated code in a new file. Could the functions for async execution (`cmdStaveN` etc) be also moved to a new file? Or maybe better to do this in a separate PR...

Spacechild1:
> Could the functions for async execution (cmdStageN etc) be also moved to a new file? Or maybe better to do this in a separate PR...

Yes, these functions belong to `pluginCmdDemo`. I think `UIUGens.cpp` resp. `UIUGens.mm` should only contain the code for the actual UI UGens (`KeyState`, `MouseX`, `MouseY`, `MouseButton`)

Spacechild1:
done!

joshpar:
@scztt - you are mentioned in this convo as someone who may be good to look at these changes... do you have time per chance? No worries if not, but I'd appreciate your eyes as well if you have some time in the next few days.

joshpar (approved):

dyfer (approved):
After discussion on the dev meeting with @joshpar and @jrsurge we think this is good to go.

Spacechild1:
Thanks! :-)

dyfer:
Thank you!

PR #5621, opened 2021-11-24
classlib: protect EventPatternProxy.prNext

### LINKED ISSUE #5597: Pdef errors inside stream need to .stop.clear to restart
## Environment

* Operating system: Linux

## Steps to reproduce

```supercollider
Pdef(\play, 'garbage').play;
// ERROR: Message 'playAndDelta' not understood. Then
Pdef(\play).isPlaying
// -> true
Pdef(\play, Pbind(\instrument, \default)).play
// nothing happens
Pdef(\play).stop.clear;
Pdef(\play, Pbind(\instrument, \default)).play;
// now it plays
```

I would expect Pdefs to stop on errors, or at least to be available for a new definition to take effect immediately. Instead, after an error occurs, Pdefs are left in playing state while their player is no longer running, making new definitions uneffective. Calling `.stop.clear` on the Pdef resets it and makes it ready to work again, but this is a hacky trick considering that the Pdef gives no feedback about its unworkable state: it seems to accept a new definition, but it doesn't play, leaving users quite clueless until they figure out on their own that they have to reset it.

## Proposed fix
Catch errors inside EventStreamPlayer:prNext, and stop the player before throwing them
https://github.com/supercollider/supercollider/blob/0ff1fe45db3dbab937596f42d678112eff104399/SCClassLibrary/Common/Streams/Stream.sc#L486
```supercollider
try {
    nextTime = outEvent.playAndDelta(cleanup, muteCount > 0);
} { |error|
    this.removedFromScheduler;
    error.throw;
};
```

elgiano:
to ensure proper cleanup of Pdefs, PatternProxies and EventPatternProxies if events occur during their players' .prNext
## Purpose and Motivation
fixes #5597 

## Types of changes
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

There are probably too many tests, my idea was to test the "feature": Pdefs, Tdefs, PatternProxies, EventPatternProxies and TaskProxies should all be able to restart playing after an error occurs in their players, when a new source is provided. However, I didn't touch PauseStream, which is used as a player by Tdef and TaskProxy, and indeed, my tests for those two classes would be successful also without this fix.

### DIFF
--- a/SCClassLibrary/Common/Streams/Stream.sc
+++ b/SCClassLibrary/Common/Streams/Stream.sc
@@ -483,7 +483,11 @@ EventStreamPlayer : PauseStream {
 			this.removedFromScheduler;
 			^nil
 		}{
-			nextTime = outEvent.playAndDelta(cleanup, muteCount > 0);
+			protect {
+				nextTime = outEvent.playAndDelta(cleanup, muteCount > 0);
+			} { |err|
+				if (err.notNil) { this.streamError };
+			};
 			if (nextTime.isNil) { this.removedFromScheduler; ^nil };
 			nextBeat = inTime + nextTime;	// inval is current logical beat
 			^nextTime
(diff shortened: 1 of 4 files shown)


telephon:
In principle yes, but `protect` has a substantial overhead which is too much for a method like `prNext`. This method is called at a very high rate.

elgiano:
closed in favor of #5623

PR #5623, opened 2021-11-25
class library: protect PauseStream and EventStreamPlayer

### LINKED ISSUE #5597: Pdef errors inside stream need to .stop.clear to restart

telephon:
## Purpose and Motivation

This will guarantee that an error thrown within a PauseStream and EventStreamPlayer correctly notify this player.

This will simplify code in various places where Pprotect was used.

This fixes #5597.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Streams/Stream.sc
+++ b/SCClassLibrary/Common/Streams/Stream.sc
@@ -346,15 +346,21 @@ PauseStream : Stream {
 		isWaiting = true;	// make sure that accidental play/stop/play sequences
 						// don't cause memory leaks
 		era = CmdPeriod.era;
-		clock.play({
-			if(isWaiting and: { nextBeat.isNil }) {
-				clock.sched(0, this);
-				isWaiting = false;
-				this.changed(\playing)
-			};
-			nil
-		}, quant.asQuant);
-		this.changed(\userPlayed);
+		protect {
+			clock.play({
+				if(isWaiting and: { nextBeat.isNil }) {
+					clock.sched(0, this);
+					isWaiting = false;
+					this.changed(\playing)
+				};
+				nil
+			}, quant.asQuant);
+			this.changed(\userPlayed);
+		} { |result|
+			if(result.isException) {
+				this.streamError
+			}
+		};
 		^this
 	}
 	reset { originalStream.reset }
@@ -444,7 +450,15 @@ EventStreamPlayer : PauseStream {
 
 	init {
 		cleanup = EventStreamCleanup.new;
-		routine = Routine{ | inTime | loop { inTime = this.prNext(inTime).yield } };
+		routine = Routine{ | inTime |
+			protect {
+				loop { inTime = this.prNext(inTime).yield }
+			} { |result|
+				if(result.isException) {
+					this.streamError
+				}
+			}
+		};
 	}
 
 	// freeNodes is passed as false from
--- a/testsuite/classlibrary/TestPatternProxy.sc
+++ b/testsuite/classlibrary/TestPatternProxy.sc
@@ -366,4 +366,28 @@ TestPatternProxy : UnitTest {
 
 
 	}
+
+	test_Pdef_error_recovery {
+		var test = false;
+		Pdef.clear;
+		Pdef(\x).quant_(0).play(quant: 0);
+		Pdef(\x, (play: {  SinOsc.foo }, dur: 0.001)); // correctly throws an error
+		0.1.wait;
+		Pdef(\x, (play: {  "----".postln; test = true })).play(quant:0);
+		0.1.wait;
+		this.assert(test, "an error shouldn't block a playing Pdef");
+	}
+
+	test_Tdef_error_recovery {
+		var test = false;
+		Tdef.clear;
+		Tdef(\x).quant_(0).play(quant: 0);
+		Tdef(\x, { 0.01.wait; SinOsc.foo }); // correctly throws an error
+		0.1.wait;
+		Tdef(\x, { "----".postln; test = true }).play(quant:0);
+		0.1.wait;
+		this.assert(test, "an error shouldn't block a playing Tdef");
+	}
+
+
 }


elgiano:
Thanks a lot @telephon! 
I had a look at it, tried it out and it seems like it works well! I can't evaluate on efficiency though, so I'll leave those considerations to you and others :)

It looks good to have an 'outer' protect instead of many inner ones, and actually this makes me wonder: could this be enough to remove the Pprotects from there? Now players catch stream errors by themselves, and call their `streamError` by themselves, so there might be no need anymore for `TaskProxy::-asProtected` or the `Pprotect` in `TaskProxy::-playOnce`!

SCClassLibrary/Common/Streams/Stream.sc
```diff
@@ -346,15 +346,21 @@ PauseStream : Stream {
...
-		clock.play({
-			if(isWaiting and: { nextBeat.isNil }) {
-				clock.sched(0, this);
-				isWaiting = false;
-				this.changed(\playing)
-			};
-			nil
-		}, quant.asQuant);
-		this.changed(\userPlayed);
+		protect {
```
jamshark70, line 349: I'm a bit unclear on the purpose of this `protect`? This is executing nothing in the user pattern -- everything here is under control of the player.

I think protecting inside the Routine is fine (provided that it passes tests).

jamshark70, line 349: Hmm... I was all set to say, OK, we're catching errors in dependants, but in the end, I have to stick with my original assumption that `protect` here is unnecessary.

The dependants are being notified *after* the PauseStream has updated all of its state variables. So (as far as I can see), an error in a dependant will not leave the PauseStream in an inconsistent state.

For example:

```
(
t = Task {
	inf.do { |i| i.postln; 1.wait }
};

c = SimpleController(t)
.put(\playing, { c.remove; 1.flurghle });

t.play;
)

^^ ERROR: Message 'flurghle' not understood.
RECEIVER: 1


0.0
1.0
2.0
3.0...
```

The `playing` dependant is scheduled on the clock, not under control of `protect`, so the error is not caught, `streamError` is not called, and... the Task continues on its merry way.

That is, an error in user code encountered when starting an idle PauseStream has no ill effect on the player -- so it's strictly the user's responsibility.

However, adding `protect` here introduces an inconsistency:

```
(
t = Task {
	inf.do { |i| i.postln; 1.wait }
};

c = SimpleController(t)
.put(\userPlayed, { c.remove; 1.flurghle });

t.play;
)

^^ ERROR: Message 'flurghle' not understood.
RECEIVER: 1

... waiting... hm... nothing is happening...

t.isPlaying  // false, why?
```

The fact that there is different error handling behavior for `\playing` and `\userPlayed` dependants doesn't seem intended.

Simplest solution is just to drop this one. It isn't doing any good, while creating new confusion.

Also `EventStreamPlayer` overrides PauseStream's `play` method -- that would be irrelevant after removing this `protect`, but as given, the PR wasn't quite complete.

jamshark70 (changes requested):

telephon:
I haven't understood yet.

```supercollider


t = Task { 1.wait; 1. flurghle; 1.wait }.play;

// wait until error
t.isPlaying // true
```

It should not be considered playing.

Routine is correct:
```supercollider
t = Routine { 1.yield; 1. flurghle; 1.yield }.play;

t.isPlaying // false
```

telephon:
… so maybe, the whole mess boils down to a simple one line change in `PauseStream`?

```supercollider
    isPlaying { ^stream.notNil }
```

should be

```supercollider
    isPlaying { ^stream.isPlaying }
```
?

telephon:
OK for testing, these are convenient examples.

All three should post `true`.

```supercollider

(
fork {
	var test = false;
	Pdef.clear;
	Pdef(\x).quant_(0).play(quant: 0);
	0.1.wait;
	Pdef(\x, (play: { SinOsc.foo }, dur: 0.001)); // correctly throws an error
	1.wait;
	Pdef(\x).isPlaying.postln;
	Pdef(\x, (play: {  "----".postln; test = true }));
	0.1.wait;
	test.postln;
}
)


(
fork {
	var test = false;
	Pdef.clear;
	Pdef(\x).quant_(0).play(quant: 0);
	0.1.wait;
	Pdef(\x, Pbind(\foo, Pfunc { 1.snokle })); // correctly throws an error
	1.wait;
	Pdef(\x).isPlaying.postln;
	Pdef(\x, (play: {  "----".postln; test = true }));
	0.1.wait;
	test.postln;
}
)


(
fork {
	var test = false;
	Tdef.clear;
	Tdef(\x).quant_(0).play(quant: 0);
	Tdef(\x, { 0.01.wait; SinOsc.foo }); // correctly throws an error
	0.1.wait;
	Tdef(\x, { "----".postln; test = true });
	0.1.wait;
	test.postln;
}
)
```

telephon:
But I can't remove the `Pprotect` from the streams.

There is something wrong with the following:

```supercollider
r = Routine { 1.yield; 1.flummi; };
x = Routine { |inval| r.next.wait; r.next; };
t = PauseStream(x).play; 
// wait a sec
t.isPlaying // true, should be false
```

jamshark70:
> I haven't understood yet.
> 
> t = Task { 1.wait; 1. flurghle; 1.wait }.play;

AFAICS, you 'protect'ed the Routine in EventStreamPlayer, but not in PauseStream.

So Task was never fixed.

I think it's a reasonable idea but it has to be done consistently throughout the PauseStream hierarchy.

jamshark70:
```
Task : PauseStream {
	*new { arg func, clock;
		var new = super.new(Routine({ |inval|
			protect {
				func.value(inval)
			} {
				new.streamError
			};
		}), clock);
		^new
	}
	storeArgs { ^originalStream.storeArgs
		++ if(clock != TempoClock.default) { clock }
	}
}
```

With this, then:

```
t = Task { 1.wait; 1.flurghle; 1.wait }.play;

// wait

t.isPlaying
-> false
```

jamshark70:
https://github.com/jamshark70/supercollider/commit/52f9c415728616eca17339ef5d27db7742ec53e7

This version removes the "protect" calls from both PauseStream:play and EventStreamPlayer:play, as neither of these will have any effect on the stream being played.

There remained, in these methods, the possibility of changing some state variables and then trying to evaluate the `quant`. If the `quant` is invalid, that could leave the object in an inconsistent state. So I moved the quant calculation to the head of the method. If it's going to fail, it will fail before updating the object's state, hence, no inconsistency.

PauseStream `refresh` now wraps the user-supplied stream within another Routine, for `protect`.

Task `refresh` re-overrides this behavior -- Task creates its own Routine, which can use `protect`.

EventStreamPlayer keeps the `protect` in the init method.

All three of the test examples above are passing in this version.

> @jamshark70 I haven't understood your example (mainly because I never use dependence so I don't really have a feeling about what it does.)

Dependants are a secondary concern here.

The primary concern is that `protect` can handle errors only from operations that are performed synchronously. Anything scheduled on a clock is not synchronous, therefore exempt from `protect`... but it's exactly the activities scheduled on the clock that we want to protect.

Therefore, `protect` as it was used in the `play` method has no effect at all on the things we wanted to protect -- so I think it was a superfluous change and should be removed.

You found this yourself, when you tested Task and found that the routine wasn't protected. It might have been tempting to assume that `protect { clock.play ... }` would take care of all of it, but `clock.play` escapes the protected context.

So then I asked myself, what possible things could happen in the `play` method, synchronously, that we might want to protect? The only things out of the direct control of the PauseStream object are the dependants, and (I realized later) `quant`.

One of the `changed` notifications occurs within a scheduled function, so it's exempt from "protect" as written. The other notification is under control from "protect" -- hence the discrepancy.

telephon:
@jamshark70 to come back to your point about external point:
>That is, an error in user code encountered when starting an idle PauseStream has no ill effect on the player -- so it's strictly the user's responsibility.

Yes, of course. But if the stream was indeed halted, the player should not pretend it is still playing.

I agree that the current PR is not the solution. Let's see if we find one nevertheless?

jamshark70:
> But if the stream was indeed halted, the player should not pretend it is still playing.

The `changed` calls follow the real work of playing, and the stream runs on the clock separate from the `changed` calls. So AFAICS there is no way for an error in a dependant to stop the stream.

> I agree that the current PR is not the solution. Let's see if we find one nevertheless?

The current PR is very close. Did you check my variant of it? https://github.com/jamshark70/supercollider/commit/52f9c415728616eca17339ef5d27db7742ec53e7 I believe it addresses just about everything (pending testing of course).

telephon:
Ah that solution is much better – thank you. A bit of code says more than thousand words. 

Can you make it a pull request? Then I'll close this one here.
I can add the tests in a separate one.

jamshark70:
OK, will do.

PR #4725, opened 2020-01-06
Topic/midi out use logical time

sirlensalot:
## Purpose and Motivation

Improve `MIDIOut` jitter by using logical time to schedule MIDI events.

## Types of changes

- New feature
  - really a performance improvement, doesn't change any functionality

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation
- [ ] This PR is ready for review

Caveats:
- this improvement only targets CoreMIDI on OSX. Since it doesn't harm the other implementations (IOS, PortMidi) it is conceivable that this change can be it's own PR.
- Tested on my machine with noticable, if subjective, improvement in scheduling reliability.

PR #5631, opened 2021-12-02
Revert #4725 - MIDI use logical time on macOS

### LINKED ISSUE #5622: MIDI issues on macOS
## Environment

* Operating system: macOS 10.15 (Catalina), macOS 11.6 (Big Sur)

## Steps to reproduce

```supercollider
MIDIClient.init;
m = MIDIOut.newByName("X-TOUCH COMPACT", "X-TOUCH COMPACT"); // testing with a midi cotroller
m.latency; // reports the default 0.2

m.noteOn(0, 30, 127); // this turns button LED on
// after the first start of SC after the boot, there's about 2s delay for the message to be sent (Catalina and Big Sur)
m.noteOn(0, 30, 0); // turn it off

// put computer to sleep for about a minute
m.noteOn(0, 30, 127); // now the message is delayed by over 40 seconds (the time the laptop was physically asleep) - Big Sur only
```

## TL;DR
As much as I'd like to keep the improved MIDI timing introduced in #4725, that PR seems to introduce issues that IMO are hard to be solved. Unless someone has an idea how to tackle them, I'd suggest to revert #4725 (and possibly release 3.12.2, if desired).

#4725 improved the relative timing of MIDI messages, but introduced 2 (3) new issues:
1. all MIDI messages are delayed approx. 2s after the first run of `MIDIClient.init` after SC starts; re-running `MIDIClient.init` removes that delay. This happens on Catalina and on Big Sur
2. MIDI messages are additionally delayed by the time the computer was asleep since running `MIDIClient.init`. This happens on Big Sur only
3. If we use logical time for MIDI messages, latency calculation should probably be revised (right now the latency value is just added to the resulting logical timestamp). Depending how we deal with the above issues, this point might be moot if we revert #4725.

Details:
#4725 introduces definition of "starting time" using both SC thread time and time from `AudioHostGetTime()`. This seems like a good approach in principle. The issue is that starting with Big Sur, the time from `AudioHostGetTime()` does not advance when the computer is asleep, while the thread clock does advance. This conceptually similar to the issue solved in #5298 (also see explanation [here](https://github.com/supercollider/supercollider/issues/5168#issuecomment-749277265)) since that issue also dealt with comparing times from different references. Since MIDI message on macOS expects the time as reported from `AudioHostGetTime()`, I don't know how we could reliably calculate logical time, since the SC thread time does advance while the computer is asleep. I guess we could conceivably re-set that offset every time system wakes from sleep, but I'm not sure how to go about it or whether that's the best solution.

Regarding problem `1.`, there seems to be some fluctuation in thread timing that occurs only on the first MIDI initialization after SC starts. I don't know where that comes from.
I tried posting the offset between `AudioHostGetTime()` and SC thread when initializing MIDI and that offset is consistently off by approx. 2s between the first and the subsequent runs of `MIDIClient.init`. The subsequent runs report offsets that differ only by few ms, which is to be expected, I think, and is not an issue. Only that initial offset is considerably large. (See below for examining that delay)

I was trying to fix the issue with clock diverging when computer goes to sleep, but that didn't work since we ultimately need to use the time in relation to `AudioHostGetTime()` to schedule MIDI events. However, my code does post initial time difference between SC clock and system nanosecond clock, where one can observe the difference in offsets between the first and the subsequent initializations of MIDI client: https://github.com/dyfer/supercollider/tree/topic/macos-midi-01

dyfer:
This reverts commit c6790a9168be1972724034e79f83d53ef4f0399c, reversing changes made to 13be6e8202674a98d0e20b58201ec10763515de3.

## Purpose and Motivation

Fixes #5622

See linked issue for explanation and details.
I also cherry-picked CI commit that fixes the macOS legacy build.

I'm suggesting to merge this into `3.12` branch in order to release `3.12.2` with this fix, but I'm open to suggestions about this.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -326,6 +326,8 @@ jobs:
       - name: install qt using aqtinstall
         uses: jurplel/install-qt-action@v2
         if: matrix.qt-version
+        env:
+          DEVELOPER_DIR: '' # remove developer dir which causes installation to fail
         with:
           modules: 'qtwebengine'
           version: ${{ matrix.qt-version }}
--- a/lang/LangPrimSource/SC_CoreMIDI.cpp
+++ b/lang/LangPrimSource/SC_CoreMIDI.cpp
@@ -72,8 +72,6 @@ MIDIClientRef gMIDIClient = 0;
 MIDIPortRef gMIDIInPort[kMaxMidiPorts], gMIDIOutPort[kMaxMidiPorts];
 int gNumMIDIInPorts = 0, gNumMIDIOutPorts = 0;
 bool gMIDIInitialized = false;
-UInt64 gCoreAudioInitTime = 0, gLogicalInitTime = 0;
-
 // cp
 static bool gSysexFlag = false;
 static Byte gRunningStatus = 0;
@@ -637,9 +635,6 @@ int prDisconnectMIDIIn(struct VMGlobals* g, int numArgsPushed) {
 
     return errNone;
 }
-
-UInt64 logicalTimeToNanos(struct VMGlobals* g) { return slotRawFloat(&g->thread->seconds) * 1000000000; }
-
 int prInitMIDI(struct VMGlobals* g, int numArgsPushed);
 int prInitMIDI(struct VMGlobals* g, int numArgsPushed) {
     // PyrSlot *a = g->sp - 2;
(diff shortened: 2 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -326,6 +326,8 @@ jobs:
       - name: install qt using aqtinstall
         uses: jurplel/install-qt-action@v2
         if: matrix.qt-version
+        env:
```
jamshark70, line 329: This appears to be a new change, not a reversion.

jamshark70 (changes requested):
Just noting an unrelated change that leaked in. As for the CoreMIDI changes, I can't comment.

.github/workflows/actions.yml
dyfer: As I noted above

> I also cherry-picked CI commit that fixes the macOS legacy build.

This is a fix from #5627 - if we want to merge this into 3.12 and make another release, we need that fix for the macOS legacy build.

jamshark70: OK, I've withdrawn that review. Thanks (and sorry that I missed / didn't understand the initial comment).

joshpar (approved):

PR #5640, opened 2021-12-10
Node proxy seti

nuss:
This pull request adds a new method `seti` to `NodeProxy`, modelled after `seti` in `Synth`. The method allows the user to set controls in a multi-channel NodeProxy based on a channel index.
Additionally it adds a new `NodeProxy role` `\seti` that allows independent sequencing on a per channel basis.

- add `seti` method to `NodeProxy`
- add NodeProxy role `\seti`
- documentation for method `seti` and NodeProxy role `\seti`
- unit tests for method `seti`

- New feature

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
@@ -307,19 +307,21 @@ NodeProxy : BusPlug {
 		this.set(*args)
 	}
 
-		seti { arg ... args; // pairs of key, index/indices (array) and value/values (array)
+	seti { arg ... args; // pairs of key, index/indices (array) and value/values (array)
 		var msg = Array.new(args.size div: 3 * 2);
-		var key, offset, value, hasControlKey;
+		var key, offset, value, controlKeys, hasControlKey;
 		var controlKeysValues, controlSize, controlIndex, controlValues;
+
+		controlKeys = this.controlKeys;
+		controlKeysValues = this.controlKeysValues;
 
 		forBy(0, args.size-1, 3, { |i|
 			key = args[i];
 			offset = args[i+1].asArray;
 			value = args[i+2].asArray;
-			hasControlKey = this.controlKeys.includes(key);
+			hasControlKey = controlKeys.includes(key);
 
-			if (hasControlKey) {
-				controlKeysValues = this.controlKeysValues;
+			if(hasControlKey) {
 				controlIndex = controlKeysValues.indexOf(key);
 				controlValues = controlKeysValues[controlIndex + 1];
 				controlSize = controlValues.size;
@@ -330,7 +332,8 @@ NodeProxy : BusPlug {
 				msg.add(key).add(controlValues);
 			};
 		});
-		if (this.isPlaying) {
+
+		if(this.isPlaying) {
 			server.sendBundle(server.latency, [15, group.nodeID] ++ msg.asOSCArgArray);
 		}
 	}
--- a/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
@@ -307,6 +307,34 @@ NodeProxy : BusPlug {
 		this.set(*args)
 	}
 
+		seti { arg ... args; // pairs of key, index/indices (array) and value/values (array)
+		var msg = Array.new(args.size div: 3 * 2);
+		var key, offset, value, hasControlKey;
+		var controlKeysValues, controlSize, controlIndex, controlValues;
+
+		forBy(0, args.size-1, 3, { |i|
+			key = args[i];
+			offset = args[i+1].asArray;
+			value = args[i+2].asArray;
+			hasControlKey = this.controlKeys.includes(key);
+
+			if (hasControlKey) {
+				controlKeysValues = this.controlKeysValues;
+				controlIndex = controlKeysValues.indexOf(key);
+				controlValues = controlKeysValues[controlIndex + 1];
+				controlSize = controlValues.size;
+				offset.do { |o, j|
+					controlValues[o % controlSize] = value.wrapAt(j);
+				};
+				nodeMap.set(*[key, controlValues]);
+				msg.add(key).add(controlValues);
+			};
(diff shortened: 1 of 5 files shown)


nuss:
Sorry, commits in this pull requests have become a bit of a mess as wanted to make sure that really everything belonging to it goes in. Maybe I should clean it it up but I'm unsure how I should do it. Can anyone give me some advice how to do it? Nevertheless everything has gone into the pull request as I intended it.

testsuite/classlibrary/TestNodeProxy.sc
```diff
@@ -385,8 +385,40 @@ TestNodeProxyBusMapping : UnitTest {
...
+		keysValues = proxy.getKeysValues;
+		this.assertEquals(keysValues[0][1], [200, 212, 200, 328, 200], "the 'freq' arg array should have been set to [200, 212, 212, 328, 200] in NodeProxy 'proxy'.");
+	}
+
+	test_seti_wrapAt {
+		var keysValues;
+		proxy.seti(\freq, [1, 23, 57], [345, 145]);
+		keysValues = proxy.getKeysValues;
+		this.assertEquals(keysValues[0][1], [200, 345, 345, 145, 200], "the 'freq' arg array should have been set to [200, 345, 345, 145, 200] in NodeProxy 'proxy'.");
+	}
```
telephon, line 423: It would be good to have one test that checks if the values in the node map ar correct.
 In `set ` this works: `Ndef(\y).set(\freq, Ndef(\x))`
So could you also check if passing a `NodeProxy` via `seti` works?

telephon (changes requested):
The code looks good! Just some small change requests.

SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
```diff
@@ -307,6 +307,34 @@ NodeProxy : BusPlug {
...
+		var controlKeysValues, controlSize, controlIndex, controlValues;
+
+		forBy(0, args.size-1, 3, { |i|
+			key = args[i];
+			offset = args[i+1].asArray;
+			value = args[i+2].asArray;
+			hasControlKey = this.controlKeys.includes(key);
+
+			if (hasControlKey) {
+				controlKeysValues = this.controlKeysValues;
```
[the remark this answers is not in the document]
nuss: This is fixed now with my latest commit

testsuite/classlibrary/TestNodeProxy.sc
nuss: I've added one more test that addresses that.

nuss:
hmm... that test failure on Linux seems unrelated, though I can't see my TestNodeProxy_seti tests having been run in the results

telephon (approved):

telephon:
the failure looks like a time out.

telephon:
Reran the test,  here is the point of failure (both linux and macOS):
```
 [!] TestNodeProxySeti:test_seti_nodeMap (0.0s) (Did not complete.)
           ! Error: "ERROR: BusPlug:setBus: bus can't be set to nil."
```

nuss:
[!] TestNodeProxySeti:test_seti_nodeMap (0.0s) (Did not complete.)
           ! Error: "ERROR: BusPlug:setBus: bus can't be set to nil."

... I don't understand this right now @telephon . The test passes on my machine (Ubuntu Studio). The one thing I was not sure about is the case when a value `nil` gets passed in (as you already noted earlier). but I'm not testing for that.

telephon:
I also don't see the reason. `setBus` is called in `defineBus`, but there the bus is definitely not `nil`.

You anyhow may want to try an wait (`server.sync`) after you assign the sources to the proxies, that might be something one could try.

nuss:
Aha. I'm a bit busy now (not sure how much I'm gonna do over the holidays). But I'm gonna try.

telephon:
I don't see the reson why this should fail. Maybe just comment that test out for a try. We can add it later.

nuss:
Yeah, it's a bit weird. I can only see this here in the test logs. I've tried to run all tests locally but can't find any trace of `test_seti_nodeMap`. Is it even executed? Executing only tests in `TestNodeProxySeti` works fine though and I got no failures (of course).

telephon:
maybe some timeout

dyfer:
For the future: if a test mysteriously fails in GHA, it should instead be added to the [list of skipped tests here](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/gha_test_run_proto.json). Thanks!

nuss:
@ dyfer - sorry, I had no idea `gha_test_run_proto.json` exist. As I'm the author of this pull request, should add the failing test again, edit ``gha_test_run_proto.json` and make another pull request? The test never failed for me on my local machine but chokes in GHA.

dyfer:
No worries, I had no idea either about `gha_test_run_proto.json` until I had a deep dive a few months ago. 
If you're willing to make another pull request re-adding the test and adding the "skip" entry to `gha_test_run_proto.json` that would be great. 
Additionally, maybe we could take a look as to why this test fails (if you have time). Looking at it, it's unclear whether the test should boot the server or not. Is it supposed to run without booted server? If so, then I don't know what's wrong. If not, then I see the server is not being booted by the test and also I suspect we might need `server.sync` after creating a new source... (?)

nuss:
Ok, I've re-added the test and added it also to `gha_test_run_proto.json`. The pull request is here: #5687.
I would certainly like to find out why the test fails but I'm a bit clueless where to start. As I said already earlier the test passes locally on my machine but fails in GHA. The test does *not* require a server to be booted.

dyfer:
Thanks for the new PR and the explanation.
Looking at the failed test, it seems the there's also an error `ERROR: Meta_Bus:control: failed to get a control bus allocated. numChannels: 1 server: localhost` - maybe the server object used runs out of available buses?

telephon:
I've taken a look, I think we need to make sure all tests use a fresh server. I'd like to refactor the tests and introduce a `ServerUnitTest` class that takes care for the server state, like this: https://github.com/supercollider/supercollider/compare/develop...telephon:topic-unit-test-refactoring-for-server

PR #5646, opened 2021-12-18
BusPlug.value fix for UGens

### LINKED ISSUE #5645: CombN fools Ndef numChannels auto-sizing down to 1 (via BusPlug.value)
## Environment

* Operating system: all

## Steps to reproduce

```
Ndef(\a2) { VarSaw.ar([333, 999]) }
Ndef(\ok) { LPF.ar(Ndef(\a2), 1111) }
Ndef(\ok).numChannels // -> 2
Ndef(\bug) { CombN.ar(Ndef(\a2)) }
Ndef(\bug).numChannels // -> 1
```

Ideally, the last line should return 2 instead of 1, as if one had written `Ndef(\nobug) { CombN.ar(Ndef.ar(\a2)) }`,

```
Ndef(\nobug) { CombN.ar(Ndef(\a2).ar) } // .ar "manually" added
Ndef(\nobug).numChannels // -> 2
```

## Fix proposed

This was discussed in a rather overly verbose [thread](https://scsynth.org/t/ndef-simple-channel-expansion-question/4840) on the scsynth forum, but in a nutshell, a fix is to make sure `BusPlug.value` *always* auto-sizes by passing nil to `this.ar` when it sees a UGen (or UGen class) as `something`, because `.numChannels` is always 1 for UGens, unlike say for another Ndef:

```
+ BusPlug {
	value { | something |
		var n;
		if(UGen.buildSynthDef.isNil) { ^this }; // only return when in ugen graph.
		if(something.notNil) {
			if(something.isKindOf(UGen.class) or: { something.isKindOf(UGen) }) {
				// UGen.numChannels always 1, so let InBus.new1 ultimately called by this.ar/.kr
				// auto-detect numChannels by keeping n=nil for the call to this.ar/.kr below.
				// Certainly need to do this in the case of UGens classes that can't be
				// flagged by something.isUGen, which is an instance method,
				// e.g. CombN.isUGen == false while CombN.ar.isUGen == true
				//("is UG or UGC" + something + n).postln
			} {
				n = something.numChannels
			}
		};
		//"bpv2: % is_UGen:% is_UGen.class:% n:% ".format(something, something.isKindOf(UGen), something.isKindOf(UGen.class), n).postln;
		^if(something.respondsTo(\rate) and: { something.rate == 'audio'} or: { this.rate == \audio }) {
			this.ar(n)
		} {
			this.kr(n)
		}
	}
}
```

avdrd:
## Purpose and Motivation

Fixes #5645 

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing

### DIFF
--- a/SCClassLibrary/JITLib/ProxySpace/BusPlug.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/BusPlug.sc
@@ -107,13 +107,21 @@ BusPlug : AbstractFunction {
 	value { | something |
 		var n;
 		if(UGen.buildSynthDef.isNil) { ^this }; // only return when in ugen graph.
-		something !? { n = something.numChannels };
+		if(something.notNil) {
+			if(something.isKindOf(UGen.class) or: { something.isKindOf(UGen) }) {
+				// UGen.numChannels always 1, so let InBus.new1 ultimately auto-size them
+				// by keeping n=nil for the call to this.ar/.kr below.
+				// Also: CombN.isUGen == false while CombN.ar.isUGen == true
+			} {
+				n = something.numChannels
+			}
+		};
 		^if(something.respondsTo(\rate) and: { something.rate == 'audio'} or: { this.rate == \audio }) {
 			this.ar(n)
 		} {
 			this.kr(n)
 		}
-
+		
 	}
 
 	composeUnaryOp { | aSelector |
--- a/SCClassLibrary/JITLib/ProxySpace/BusPlug.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/BusPlug.sc
@@ -121,7 +121,6 @@ BusPlug : AbstractFunction {
 		} {
 			this.kr(n)
 		}
-		
 	}
 
 	composeUnaryOp { | aSelector |


telephon:
There is an alternative solution, maybe you could try if that works for you?

https://scsynth.org/t/ndef-simple-channel-expansion-question/4840/24?u=julian


you'd add this to `BusPlug`:

```supercollider
asAudioRateInput {
		^this.ar(nil)
	}
```

avdrd:
True, it would fix it in this case, with less typing. But there can be other call paths that call `BusPlug.value` with a UGen `something`. Trying to find and fix all the call sites in all UGens seems much more brittle than making `BusPlug.value` expand on UGen `something`. Can you see a case where the latter would be semantically wrong?

telephon:
OK, let's think this through here.

The original implementation was based on the assumption that if you write `SomeUGen.ar(X)`, thenn `X` will be passed the message `asUGenInput`. This is a system wide thing. Now since this is not the case everywhere, we have two options:

1. remove the feature in BusPlug (that is, just return whatever the BusPlug is initialised to already). PRO: easier to understand, more consistent with what `ar` and `kr`do. CON: may break code that relies on the feature.
2. the interface `asAudioRateInput` is incompletely implemented. Sometimes it passes a class (`CombN`), sometimes not (`DelayN`). Also, `asAudioRateInput` is not defined correctly for `BusPlug`. Btw. also `CombL` calls `asUGenInput`, but this comes later, so `asAudioRateInput` breaks it.

So in order to avoid breaking code, we could make these three changes:

```supercollider
// in BusPlug (this alone fixes the current issue):
asAudioRateInput {
		^this.ar
	}

// in DelayN, BufDelayN (in a separate pull request)
in.asAudioRateInput(this)

// in AbstractFunction (optional: but this is a better interface)
asAudioRateInput { |for|
		var result = this.asUGenInput(for);
		^if(result.rate != \audio) { K2A.ar(result) } { result }
	}

```


>True, it would fix it in this CombN case, with less typing. But there can be other call paths that call BusPlug.value with a UGen something.

It wouldn't make sense to pass in a UGen/UGen-class to `.value` directly. It is just via `asUGenInput` that this is passed and we have to catch that. In the future, we could remove the feature that `.value` checks the rate of the input completely, and just do th above (1.). But this needs communicating, becaus it is a (minor, but still) breaking change.

avdrd:
This more of a meta-comment, but the way I see it SC is stuck where it's at precisely because of this issue: incompletely specified interfaces (and that's an understatement), but every core dev seems to worry about breaking the implied promises in these that someone might have relied on in some putative program.

joshpar:
You're welcome to join our dev meetings and help support changes like this. We have deprecation paths that new functionality can replace old / faulty existing classes and methods, and you are welcome to come help support those efforts.

avdrd:
> It is just via asUGenInput that this is passed and we have to catch that. 

Of course, it's your library, so you can come up with whatever (brittle)
"don't call this directly, call that instead" conventions you want, but
you'd have to document them if you expect anyone else to use them,
including yourselves sometime later, because you apparently forgot them.

telephon:
> Of course, it's your library, so you can come up with whatever (brittle)
> "don't call this directly, call that instead" conventions you want

I was not aruing from that point of view. It is a question of systematicity in the class library on the whole.

telephon:
I don't think this is a good solution. We should check if the existing fix in BusPlug 
```
asAudioRateInput {
		^this.ar
	}
```
resolves the issue?

joshpar:

telephon:
I had written:

>I don't think this is a good solution. We should check if the existing fix in BusPlug … resolves the issue?

so there needs to be some exchange about this.

dyfer:
IIUC https://github.com/supercollider/supercollider/pull/5679 addressed this...? If not, feel free to reopen.

Issue #4664, opened 2019-12-07
[build] MacOS incremental builds broken


brianlheim:
## Environment

* SuperCollider version: current dev/3.10
* Operating system: macOS 10.14.6
* Other details (Qt version, audio driver, etc.): Xcode 11.0, CMake 3.15.3

## Steps to reproduce

1. build the SuperCollider target once
2. modify a file in the SC IDE target, forcing it to rebuild
3. rebuild the SuperCollider target

## Expected vs. actual behavior

instead of the build completing normally, it appears that the sclang target already existing in the app package causes the (AFAICT, unavoidable) codesigning after building the package to fail. here's the full output:

```
=== BUILD TARGET SuperCollider OF PROJECT SuperCollider WITH CONFIGURATION Debug ===

Check dependencies

Ld build/editors/sc-ide/Debug/SuperCollider.app/Contents/MacOS/SuperCollider normal x86_64
    cd /Users/brianheim/git/sc2
[...]

PR #5650, opened 2021-12-21
Explicitly add ad hoc code signing on macOS

### LINKED ISSUE #5603: Codesigning issues on macOS with M1 (ARM64) processors
## Environment

* SuperCollider version: `develop` branch
* Operating system: macOS 11.6
* Other details (Qt version, audio driver, etc.): M1 MacBookPro, XCode 13.0

## TL;DR

Apple requires _all_ binaries to be signed in order to run on Apple ARM64 processors (M1 family), including the binaries built locally. XCode does sign all the binaries when building from their IDE. We currently [disable codesigning in CMake](https://github.com/supercollider/supercollider/pull/5362) due to issues with signing dependencies.

## Steps to build SuperCollider on M1 MBP

**Note: On ARM64 Qt5 from homebrew does not provide the WebEngine, so we need to disable its use** (#5584)

- Install all the dependencies using homebrew, as indicated in the `README_MACOS.md`
- Configure and build as usual, except without QtWebEngine
```sh
cmake -G Xcode -D CMAKE_PREFIX_PATH=`brew --prefix qt5` -D SUPERNOVA=ON -D SC_USE_QTWEBENGINE=OFF ..
cmake --build . --target install --config RelWithDebInfo
```
The build should finish successfully. However, the signatures of the binaries are invalid. To rectify that, we need to re-sign all the binaries
- Sign the binaries and libraries
```sh
# assuming you are in the build folder
codesign --deep --force --verify --verbose --timestamp --sign "Apple Development" Install/SuperCollider/SuperCollider.app Install/SuperCollider/SuperCollider.app/Contents/MacOS/*  Install/SuperCollider/SuperCollider.app/Contents/Resources/*  Install/SuperCollider/SuperCollider.app/Contents/Frameworks/*
```

Now you should be able to run your build.

We need to sign everything with our developer certificate (ad-hoc signing doesn't seem to be an option). If you only have a single certificate, you should be able to use "Apple Development" as the identity. Otherwise check the existing certificates using `security find-identity -v -p codesigning` and use the SHA of the chosen certificate instead of "Apple Development".

## Possible solutions

I'm guessing the proper way to "fix" this is to re-integrate codesigning into our CMake workflow, but I'm not sure how to achieve that with all the dependencies, passing the credentials etc.

Comments always welcome!

dyfer:
Fixes #5603

On macOS all binaries needs to be signed when run on M1 hardware (arm64 / aarch64)
When building, the binaries are signed ad hoc, however it turns out that `macdeployqt`, as well as manipulations with `install_name_tool` invalidate these signatures.
I propose the solution that runs ad hoc signing after the `macdeployqt` step.

I'm open to suggestions whether this is the desired solution and if there's a better name for the introduced CMake flag.

The PR includes the following changes:
- A new option is added to `CMakeLists.txt` to trigger ad-hoc codesigning after building (`SC_CODESIGN_AFTER_DEPLOY`, defaults to `ON`, macOS only)
- The following components are signed: binaries `SuperCollider`, `sclang`, `scsynth`, `supernova`, all `.framework` and `.dylib` files in `Contents/Frameworks/`, `DiskIO_UGens` plugins, and `SuperCollider.app` itself (last). 
  - `SuperCollider` binary is signed with the `--deep` flag to avoid `code object is not signed at all` error; signing other files is done without that flag
  - Signing `DiskIO_UGens` was needed due to a rather obscure issue with codesigning - the servers would not boot, but only after incremental builds, while running fine on clean builds. Signing this specific ugen file seems to have fixed that. I made a note of it in the commit message.
- I added `--deep` flag to build-time codesigning for the `SuperCollider` target. This seems to me like a more proper solution to #4664 than turning off build-time codesigning altogether. 
  - Since the incremental builds do work now with the build-time codesigning turned on, I decided to turn OFF `SC_DISABLE_XCODE_CODESIGNING` by default, as I don't think it's needed anymore.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,6 +248,9 @@ option(SC_ABLETON_LINK "Build with Ableton Link support" ON)
 if(APPLE)
     # codesigning breaks increamental builds in Xcode 11+
     option(SC_DISABLE_XCODE_CODESIGNING "Disable codesigning in XCode" ON)
+    # however, codesigning is needed on arm64 to run the app locally at all
+    # but it is being broken by macdeployqt, hence we sign after deploying the app
+    option(SC_CODESIGN_AFTER_DEPLOY "Run ad hoc codesigning after creating the app bundle" ON)
     # verify_app fails with official Qt; we'll still use it for Qt from homebrew
     option(SC_VERIFY_APP "Run verify_app on the app bundle" OFF)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,10 +246,10 @@ endif()
 option(SC_ABLETON_LINK "Build with Ableton Link support" ON)
 
 if(APPLE)
-    # codesigning breaks increamental builds in Xcode 11+
-    option(SC_DISABLE_XCODE_CODESIGNING "Disable codesigning in XCode" ON)
-    # however, codesigning is needed on arm64 to run the app locally at all
-    # but it is being broken by macdeployqt, hence we sign after deploying the app
+    # codesigning used to break increamental builds in Xcode 11+, but it's now fixed
+    option(SC_DISABLE_XCODE_CODESIGNING "Disable codesigning in XCode" OFF)
+    # codesigning is also being broken by macdeployqt, but it is needed on arm64
+    # to run the app locally at all, hence we sign after deploying the app
     option(SC_CODESIGN_AFTER_DEPLOY "Run ad hoc codesigning after creating the app bundle" ON)
     # verify_app fails with official Qt; we'll still use it for Qt from homebrew
     option(SC_VERIFY_APP "Run verify_app on the app bundle" OFF)
--- a/editors/sc-ide/CMakeLists.txt
+++ b/editors/sc-ide/CMakeLists.txt
@@ -547,6 +547,20 @@ if(APPLE OR WIN32)
                 ")
         endif()
 
+        if(SC_CODESIGN_AFTER_DEPLOY)
+            string(APPEND VERIFY_CMD "
+                message(STATUS \"Running ad hoc code signing\")
+                execute_process(COMMAND 
+                    sh -c \
+                    \"find ${CONTENTS_DIR} \
+                    \\\\( -name \\\"*.dylib\\\" -o -name \\\"*.framework\\\" \
+                    -o -name \\\"SuperCollider\\\" -o -name \\\"sclang\\\" \
+                    -o -name \\\"scsynth\\\" -o -name \\\"supernova\\\" -maxdepth 2 \
+                    \\\\) -exec codesign --force --deep --sign - {} \\\;\"
+                )
+            ")
+        endif()
+
         if(SC_VERIFY_APP)
             string(APPEND VERIFY_CMD "
                 message(STATUS \"Verifying app\")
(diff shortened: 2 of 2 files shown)


joshpar:
I think this is probably going to be the way to go. However, I think this PR should also include changes to the Mac build notes.

dyfer:
My original implementation turned out to be faulty, since I was trying to gather a list of files in the installation location before they were created. 

Subsequently I went through 3 different iterations of possible solutions (adding cmake commands to the string executed at the end, using `ls` to gather files and pass that to `codesign` and finally using `find` to get a list of files) with TONS of escaping, and I think the current solution is reasonable. I'll clean up the PR and add note in macOS build instructions before marking this as ready for review.

Failures in the CI seem unrelated to this PR BTW

dyfer:
I am going to also investigate adding the "linker-signed" flag which _might_ solve some of the problem without re-signing.

dyfer:
> I am going to also investigate adding the "linker-signed" flag which _might_ solve some of the problem without re-signing. https://gitlab.kitware.com/cmake/cmake/-/issues/21854

This turns out to be available only on macOS 11+ and I'm not sure if it would fix the issue with macdeployqt touching binaries (to be tested though - I'm not on M1 currently).

In the meantime I think I found a more proper solution to #4664 - original solution was to disable codesigning altogether; that didn't seemed like the best idea to me and I found that running `codesign --deep` on one binary (`SuperCollider` itself, i.e. the IDE) solves the original problem. For now I'm putting this change in this PR but maybe we'll want to consider it separately?

Finally, I don't know how/what to document here. This doesn't seem to change anything about building locally (aside from making it work on M1) and AFAIU does not change anything about signing the releases either. 

Before this PR local builds always had invalid signatures. While code signatures for local builds were not required to run them on earlier macOS systems (before Big Sur and M1), making this change for all builds seems fine to me, as it results in binaries with valid (albeit still ad-hoc) signatures.

dyfer:
I am happy with this PR now. I updated the first comment to give a clear description of the current changes.

Here's what changed from the previous version:
- codesigning step a little more granular - `codesign --deep` is only run on `Contents/MacOS/SuperCollider`, while all other files are signed without the `--deep` flag
  - I did this since I've come across information that using the `--deep` flag unnecessarily is not recommended
- I did investigate using `-o linker-signed` flag for codesigning, but eventually did not include it here
  - this was only relevant for `libfftw3f.3.dylib`, which we manually correct with `install_name_tool` after assembling the app
  - it would result in adding a re-signing step before running `install_name_tool`, but we re-sign all the libraries at the end so I decided this would be superfluous
  - the only downside of the current solution is that we get a warning towards the end of the build process: `install_name_tool: warning: changes being made to the file will invalidate the code signature in: (...)/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/Frameworks/libfftw3f.3.dylib`
- I've added codesigning for the `DiskIO_UGens` plugin. Other plugins don't seem to need to be re-signed, but this plugin does, however the issue only transpires on incremental builds. Originally I observed that that the server would not boot after an incremental build, unless codesigning in the build process was disabled. Eventually I tracked this down to this one plugin that needed to be re-signed in order to make incremental builds functional again (without disabling the codesigning in the build process)

dyfer:
Spoke too soon, something is wrong...

dyfer:
It turned out, that I needed to sign all the ugens after all

joshpar (approved):

PR #5651, opened 2021-12-21
CMake: allow building without Qt on macOS

dyfer:
## Purpose and Motivation

Previously, when selecting `-D SC_QT=OFF` build option on macOS, the configure step would fail due to references to `SuperCollider` target in other targets: sclang, scsynth, supernova and the plugins. 

This PR allows to run the configure step, and thus build SC without Qt on macOS.

(I vaguely remember this issue has been mentioned before, but I can't find it in the open issues)

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/CMakeLists.txt
+++ b/lang/CMakeLists.txt
@@ -350,9 +350,12 @@ if(LTO)
 endif()
 
 if(APPLE)
-    add_custom_command(TARGET sclang POST_BUILD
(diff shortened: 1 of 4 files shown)


joshpar (approved):

dyfer:
Well, this is a draft. Do not merge :)
I think we need to put appropriate lines inside checks whether `scappbindir` etc exists

joshpar:
@dyfer - ah - good point! thanks for mentioning that.

dyfer:
Ready for review!

joshpar (approved):

Issue #5660, opened 2021-12-27
FlowVar.value returns \unbound if `next` is called more than once on the Routine that is hosting it


avdrd:
## Environment

* SuperCollider version: 3.12.1
* Operating system: all

## Steps to reproduce

```
(
~fv = FlowVar.new;
r = Routine {
	"Routine starts".postln;
	v = ~fv.value + 0.1;
	"Routine got its value % to work with".format(v).postln;
	v.alwaysYield;
};
)


r.next
// Routine starts
// -> hang
r.next
// Routine got its value unbound to work with
// -> unbound
```

## Expected vs. actual behavior

I'd have expected it to hang again until the value of the FlowVar is actually set.

To give a bit more motivation why someone would try something like this: imagine you write a selector Routine that reads data in Unix-select style from several other routines (or streams), meaning some of the latter can return \hang and some not, and your selector routine stops e.g. at the first non-hang (real) data it receives and yields that bit. With that approach it's inevitable that a routine that gave a \hang may be called again while it's still supposed to hanged, but with FlowVar as presently implemented it will return \unbound on subsequent calls.

## Proposed fix (work in progress)

For my use I've changed it like

```
+ FlowVar {
	value {
		while { value == \unbound } { // mod
			condition.wait
		};
		^value
	}
}
```

So now with the above setup it does re-hang multiple times until a value is actually set with `value_`.

```
r.next
//Routine starts
//-> hang
r.next
//-> hang
r.next
//-> hang
~fv.value = 12
// Routine got its value 12.1 to work with
```

Actually that is not entirely a good fix. Because `Condition.wait` will add multiple entries of the same thing to its queue, in that while loop, so thereafter on `Condition.signal` (called by `FlowVar.value_`), the same routine will be scheduled on the clock and then run several times. `alwaysYield` in the above example unfortunately masked this issue, but it becomes apparent if the routine does a `loop { /* side-effect like postln */ something.yield }` instead.  Here's that kind of example:

```
(
~fv = FlowVar.new;
r = Routine { loop {
	"Routine resumes".postln;
	v = ~fv.value + 0.1;
	("Routine got its value:" + v).postln;
	v.yield;
}};
)
```
With the current FlowVar this latter example actually alternates between \hang and \unbound at every `r.next`

Just with the not so great fix I proposed above, it does something like

```
r.next
r.next
~fv.value = 145
// Routine got its value: 145.1
// Routine resumes
// Routine got its value: 145.1
```

[...]

PR #5670, opened 2021-12-29
CondVar.prWait: fix duplication in waitingThreads

### LINKED ISSUE #5668: Calling CondVar.wait multiple times from the same routine results in the same routine being ran multiple times on CondVar.signalAll
This is a replication of #5667 using "workflow 1" of CondVar, meaning `while` loop in the caller.

## Environment

* Operating system: all

## Steps to reproduce

```
(
c = CondVar.new;
t = false; // CondVar only supports "external" tests
r = Routine {
	"Routine started.".postln;
	while {
		postln("t ==" + t);
		t == false // actual test of while
	} {
		"t is false. Routine waiting at CondVar.".postln;
		c.wait; // (t) for 2nd wf
		"Routine resumed from wait.".postln;
	};
	loop {
		"t is true. Routine can do its work.".postln.yield;
	};
})

r.next
// Routine started.
// t == false
// t is false. Routine waiting at CondVar.
// -> wait

r.next
// Routine resumed from wait.
// t == false
// t is false. Routine waiting at CondVar.
// -> wait

t = true;
c.signalAll; // you don't really expect more than one copy of yourself...
// -> a CondVar
// Routine resumed from wait.
// t == true
// t is true. Routine can do its work.
// t is true. Routine can do its work.
```

So we got the exact same bad result as when using Condition, meaning our routine ran twice.

Someone might argue this can be "fixed" by calling just `signalOne` in this case, but that isn't a feasible approach for fixing something like FlowVar, since it makes no sense for the latter to signal just one of the threads that called `.value` on the FlowVar when latter is set with `.value_`, i.e. the semantics there mandate a signalAll (which is the semantics of `Condition.signal`, currently called by FlowVar, if that needs saying).

### LINKED ISSUE #5669: Calling CondVar.wait(predicate) results in the same routine being ran multiple times on CondVar.signalAll, proportional to the number of prior `next` calls (on the routine) that returned \wait
This is a replication of #5667 and #5668 with "workflow 2" of CondVar, meaning `while` loop internal to `CondVar.wait`, which allows a predicate to specify the (internal while) `CondVar.wait` test. Arguably this is even a more serious bug than the other two linked because in this "workflow 2" of CondVar it's even less apparent to the `CondVar.wait` caller that multiple internal yields and multiple queue additions of the same routine may have happened inside `CondVar.wait`.

## Environment

* Operating system: all

## Steps to reproduce

```
(
c = CondVar.new;
t = false; // CondVar only supports "external" tests
r = Routine {
	"Routine started and waiting at CondVar with predicate { t }.".postln;
	c.wait({ t }); // needs to be in a function to work
	"Routine resumed from wait({ t }).".postln;
	postln("t ==" + t);
	loop {
		"Routine can do its work.".postln.yield;
	};
})

r.next
// Routine started and waiting at CondVar with predicate { t }.
// -> wait

r.next
// -> wait

t = true;
c.signalAll;

// Routine resumed from wait({ t }).
// t == true
// Routine can do its work.
// Routine can do its work.
```

I would have expected the routine to only run once after being signaled.

## Also

Therefore one alas can't just fix #5660 by using CondVar as it stands, instead of Condition in the guts of FlowVar.

I'm quite interested in what @mossheim thinks about (all) this since they apparently designed CondVar, if I'm not mistaken. In particular the 2nd solution listed in #5667 "fixes" section would certainly apply to `CondVar.wait(predicate)` as well. It's alas less obvious to me right now what to do about `CondVar.wait` without any predicate specified, i.e. how to fix #5668.

avdrd:
## Purpose and Motivation

Fixes #5669 & fixes #5668 as well.

It would enable a simple fix for #5660 as well, simply by switching to using CondVar over there instead of Condition, after #5670 is applied.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Core/CondVar.sc
+++ b/SCClassLibrary/Common/Core/CondVar.sc
@@ -63,8 +63,14 @@ CondVar {
 	deepCopy { this.shouldNotImplement(thisMethod) }
 
 	prWait {
+		var tp = thisThread.threadPlayer;
 		this.prSleepThread(thisThread);
-		\wait.yield
+		\wait.yield;
+		waitingThreads.remove(tp);
+		// ^^ Sadly, prRemoveWaitingThread takes the wrong argument for what we
+		// need to ensure corectness here if the threadPlayer of thisThread
+		// changes while this method has yielded control to one that does
+		// something like that at the yield above.
 	}
 
 	// Returns true iff we were woken via signal (and not timeout)


avdrd:
Actually just this commit introduces a subtle bug with `threadPlayer`s. Because  with it `wait` does its own removing from `waitingThreads`, if you have legit multiple copies of the same `threadPlayer` in `waitingThreads` (because a threadPlayer can legitimately own multiple routines), `signalOne` will remove the next one, *in addition* to the one removed in `wait`, so you'll have a double-removal problem in those cases.

The real fix is to have a map from threads->threadPlayers as I have [said](https://scsynth.org/t/and-reason-why-condition-allows-duplicate-entries-in-its-waitingthreads) on the forum. But since the SC community doesn't seem to care or grok my uses cases, I won't bother supplying the real fix for this here, but I'll "fork" CondVar and FlowVar to my own classes, perhaps to be provided as a quark if there's 3rd party interest.

PR #5671, opened 2021-12-29
FlowVar.condition switchover to CondVar; add reset method

### LINKED ISSUE #5660: FlowVar.value returns \unbound if `next` is called more than once on the Routine that is hosting it

### LINKED ISSUE #5663: FlowVar.reset
## Motivation

I found it really irritating to have to use `v = Ref(FlowVar())` and write `v.value.value` all over the place when dereferencing just so I can reset a FlowVar (with `v.value = FlowVar()`).

```
+ FlowVar {
	reset {	value = \unbound }
}
```

## Plan for Implementation
PR unless someone objects to this simple idea.

avdrd:
## Purpose and Motivation

FlowVar switchover to using CondVar for its `condition`. Partially fixes #5660, meanig the first example in there because `CondVar.wait` implements its own condition-re-checking `while` loop, when supplied with a predicate. This PR will also help fix the 2nd example that I alas put the same #5660 bug report then when #5670 is merged, fixing and underlying issue in CondVar (that also exists in Condition).

I've also added a `reset` method in the same commit that resolves #5663 since I didn't see any objections to the enhancement request. It's in the same commit  because the proximity of the lines changed would likely have created a gratuitous merge conflict otherwise. 

## Types of changes

- Bug fix
- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Core/Condition.sc
+++ b/SCClassLibrary/Common/Core/Condition.sc
@@ -48,17 +48,18 @@ FlowVar {
 	}
 	init { arg inVal;
 		value = inVal;
-		condition = Condition { value != \unbound };
+		condition = CondVar.new;
 	}
 	value_ { arg inVal;
 		if (value != \unbound) {
 			Error("cannot rebind a FlowVar").throw
 		};
 		value = inVal;
-		condition.signal;
+		condition.signalAll;
 	}
 	value {
-		condition.wait
+		condition.wait { value != \unbound };
 		^value
 	}
+	reset { value = \unbound }
 }


avdrd:
I don't care about this PR anymore because I don't use this class anymore, having written my own substitute for it, rather having to explain here over and over why the changes are desirable.

PR #5676, opened 2022-01-01
Topic fix  some tests

telephon:
## Purpose and Motivation

NodeProxy tests that failed because of wrong assumptions in the test.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/testsuite/classlibrary/TestNodeProxy.sc
+++ b/testsuite/classlibrary/TestNodeProxy.sc
@@ -229,10 +229,10 @@ TestNodeProxy : UnitTest {
 	}
 
 	test_copy_nodeMapIsCopied {
-		var oldProxy = proxy.source_({ Silent.ar });
+		var oldProxy = proxy.source_({ Silent.ar }).set(\freq, 7817);
 		var newProxy = oldProxy.copy;
 
-		this.assertEquals(oldProxy.nodeMap, newProxy.nodeMap);
+		this.assertEquals(oldProxy.nodeMap.at(\freq), newProxy.nodeMap.at(\freq));
 	}
 
 	test_copy_independent {
@@ -267,7 +267,7 @@ TestNodeProxy : UnitTest {
 		timeout.stop;
 		resp.free;
 
-		this.assertEquals(result, proxy.source, "after the crossfade from a ugen function to a value the bus should have this value");
+		this.assertFloatEquals(result, proxy.source, "after the crossfade from a ugen function to a value the bus should have this value");
 
 		server.quit;
 		server.remove;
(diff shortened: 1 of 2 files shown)


testsuite/classlibrary/TestNodeProxy.sc
```diff
@@ -267,7 +267,7 @@ TestNodeProxy : UnitTest {
 		timeout.stop;
 		resp.free;
 
-		this.assertEquals(result, proxy.source, "after the crossfade from a ugen function to a value the bus should have this value");
+		this.assertFloatEquals(result, proxy.source, "after the crossfade from a ugen function to a value the bus should have this value");
```
dyfer, line 270: Did `test_control_fade` fail before? If not, what's the rationale for this change?

testsuite/classlibrary/TestNodeProxy_Server.sc
```diff
@@ -71,6 +71,7 @@ TestNodeProxy_Server : UnitTest {
 		// ... and also afterwards, we force an update
 		server.sendStatusMsg;
 		server.sync;
+		0.8.wait;
```
dyfer, line 74: Is this really the only way to fix this? This is a rather lengthy wait. Could this be addressed by sending some other message to the server and running another server.sync? Or maybe trying the shorter wait time? 

If you remove the exclusion from `gha_test_run_proto.json` we'll be able to see how the shorter wait time performs in the CI (which is arguably not tuned for audio/low-latency performance).

telephon: I agree. I find it weird that this doesn't work faster, but it has to do with the time the server takes to report the deallocation of the synth def.

telephon: I've found another solution: just fix the wrong order in the `sync` …

adcxyz (approved):

dyfer:
Hey @adcxyz 
Maybe I should've mark this as "changes requested", my bad, but if you look at my comments, some of them were not addressed. I'd prefer to wait for outstanding discussion to be resolved before merging a PR, thanks!

adcxyz:
@dyfer 
Sorry, read the discussion too quickly, and thought @telephon had addressed your concerns.
What is still open, enabling these tests again, 
and anything else?

dyfer:
No worries @adcxyz 
Yes, the main point was to remove fixed tests from the "skipped" list in `gha_test_run_proto.json`. Aside from that I asked about the change in `test_control_fade` but that change is minor so I assume it's fine.

dyfer:
`test_synthDef_isReleased_afterFree` got enabled in #5702

PR #5680, opened 2022-01-02
Find jack using cmake's FindPkgConfig

### LINKED ISSUE #5678: does not detect jack if provided by pipewire-jack-audio-connection-kit-devel
[EDIT] not sure about the "bug" label. Still posting the issue in the hope that someone knows a work-around.

## Environment

* SuperCollider version: 3.12.2-rc1
* Operating system: Fedora GNU/Linux

```
Package pipewire-jack-audio-connection-kit-devel-0.3.40-1.fc35.x86_64 is already installed.

$ pkgconf jack --cflags
-D_REENTRANT 
$ pkgconf jack --libs
-L/usr/lib64/pipewire-0.3/jack -ljack -ljacknet 
$ pkgconf jack --print-provides
jack = 1.9.17
```

related: https://bugzilla.redhat.com/show_bug.cgi?id=1966869

## Steps to reproduce

```
$ cmake -DSC_QT=OFF -DCMAKE_INSTALL_PREFIX=/opt/supercollider/3.12.2-rc1   ../source-sc-3.12.2-rc1
```

expected: build from source succeeds

actual:
```
$ cmake -DSC_QT=OFF -DCMAKE_INSTALL_PREFIX=/opt/supercollider/3.12.2-rc1   ../source-sc-3.12.2-rc1
-- The C compiler identification is GNU 11.2.1
-- The CXX compiler identification is GNU 11.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- SuperCollider Version: 3.12.2-rc1
-- Building from branch HEAD, commit hash is 38b28f95e
-- Build type defaulting to "RelWithDebInfo"
-- Not compiling SC IDE, because it requires Qt but SC_QT is False
-- Found Pthreads: /usr/lib64/libpthread.a  
-- Using bundled boost
-- Using bundled yaml-cpp
-- HIDAPI components:
--     linux hidraw
-- libudev stable: 1
-- Found UDev: /usr/lib64/libudev.so
--    include: /usr/include
--     hidapi_parser
-- Building with HID support
-- Found Sndfile: /usr/lib64/libsndfile.so  
-- Found X11: /usr/include   
-- Looking for XOpenDisplay in /usr/lib64/libX11.so;/usr/lib64/libXext.so
-- Looking for XOpenDisplay in /usr/lib64/libX11.so;/usr/lib64/libXext.so - found
-- Looking for gethostbyname
-- Looking for gethostbyname - found
-- Looking for connect
-- Looking for connect - found
-- Looking for remove
-- Looking for remove - found
-- Looking for shmat
-- Looking for shmat - found
-- Looking for IceConnectionNumber in ICE
-- Looking for IceConnectionNumber in ICE - found
-- Found fftw3f: /lib64/libfftw3f.so
-- FFT library (scsynth): FFTW
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.8.0") 
-- Checking for module 'jack'
--   Found jack, version 1.9.17
CMake Error at server/scsynth/CMakeLists.txt:43 (message):
  Jack selected as audio API, but development files not found
```

dvzrv:
## Purpose and Motivation

Replace the custom `find_package()` cmake integration for jack with the use of `pkg_check_modules()` (part of cmake's [FindPkgConfig](https://cmake.org/cmake/help/v3.12/module/FindPkgConfig.html)), as it is more straight forward to use this way and does not rely on inconsistent behavior due to calls to `find_path()` and `find_library()` (which disregard the output of pkg-config).

Fixes #5678

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dvzrv:
The failure in the test suite seems unrelated to this PR.

dvzrv:
There are more uses of the legacy FindJack.cmake implementation in the portaudio submodule. I don't really know what to do about those yet, but maybe they can be fixed upstream and removed there as well.

dyfer (approved):
Thanks! Looks good to me. To test, I built with "ancient" Jack on macOS and still works fine.

I wonder if this changes anything on Windows but @sonoro1234 is probably the only person who knows/cares about that...

sonoro1234:
I cant test now: laptop broken

sonoro1234:
I think it should try to work without pkgconfig installed also.

dyfer (changes requested):
After the comment from @sonoro1234 I also think it would be good to not fail if pkgconfig is not available (if I understand correctly that this is the current behavior?)

dvzrv:
After conversing with @falktx on this topic, he is of the opinion, that on Windows an application should not link directly against libjack.dll, but instead use `LoadLibrary` to wrap it.
The implementation that is about to be removed should therefore be replaced by something akin to https://github.com/DISTRHO/DPF/tree/develop/distrho/src/jackbridge on Windows, while it is completely fine to use pkgconfig in all other cases.

Related to this and seeing that there are no CI targets to even build against jack2 on Windows, could you elaborate on how this worked in the past, or whether this ever worked in the past? I have no Windows system to test on and am generally not really interested in the platform, so my question would also be: Is there even a support guarantee for it?

dyfer:
> (...)  on Windows an application should not link directly against libjack.dll, but instead use LoadLibrary to wrap it.

I agree that this would be optimal. There were some previous conversations about the different backends being a runtime option, where this would be needed anyway. However I don't see that we'll be implementing this shortly (although contributions are always welcome of course).

> Related to this and seeing that there are no CI targets to even build against jack2 on Windows, could you elaborate on how this worked in the past, or whether this ever worked in the past? I have no Windows system to test on and am generally not really interested in the platform, so my question would also be: Is there even a support guarantee for it?

There's no support guarantee and we do not test for this. IIUC correctly @sonoro1234 has used this configuration and knowing that I don't want to break his setup. However, I don't want to advocate for functionality that's too obscure either... If we can maintain some way of _optionally_ providing Jack include/library paths as cmake config flags I'd support that, but if everyone agrees that requiring `pkgconfig` is the right solution, I can be persuaded.

dvzrv:
>  However, I don't want to advocate for functionality that's too obscure either... If we can maintain some way of optionally providing Jack include/library paths as cmake config flags I'd support that, but if everyone agrees that requiring pkgconfig is the right solution, I can be persuaded.

It appears as if the direct linking against libjack.dll is not the right way to do it on Windows (but I am far from being an expert on that topic), so pkgconfig integration there would not necessarily help anyways I guess.
Given that the FindJack.cmake file introduces inconsistent behavior in regards to the assumptions made with plain pkgconfig integration and generally adds more complexity, I would still advocate for removing it. However, I can not estimate how much time and effort it would take to implement something for Windows, that uses `LoadLibrary` (disregarding the effort it would take to deal with a runtime selection of backends on all operating systems, as I believe that should really be a different topic).

In regards to the cmake config flags, I think it could be saner to use `find_library()` etc. on Windows specifically instead (although [pkgconf](https://github.com/pkgconf/pkgconf/) can also be built for Windows AFAIK - @sonoro1234 did you ever use that for header/library detection on Windows?), but leave a note, that it should be replaced with `LoadLibrary` in the future.

sonoro1234:
> did you ever use that for header/library detectio

Never

cmake_modules/FindJack.cmake
```diff
@@ -1,87 +0,0 @@
...
-      /usr/local/include
-      /opt/local/include
-      /sw/include
-  )
-  
-  find_library(JACK_LIBRARY
-    NAMES
-      jack
-    PATHS
-      ${_JACK_LIBDIR}
```
dvzrv, line 47: @sonoro1234 is this how you would provide the search path on Windows for this? e.g. `cmake -D _JACK_LIBDIR=/path/to/jack -D _JACK_INCLUDEDIR=/path/to/jack`?

sonoro1234:
Got new laptop:

Building for AUDIOAPI jack:

```
-- Audio API (scsynth): jack
-- Found fftw3f: C:/supercolliderrepos/x86/fftw/libfftw3f-3.dll
-- FFT library (supernova): FFTW
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
CMake Error at C:/cmakes/cmake-3.16.3-win64-x64/share/cmake-3.16/Modules/FindPkgConfig.cmake:511 (message):
  pkg-config tool not found
  C:/cmakes/cmake-3.16.3-win64-x64/share/cmake-3.16/Modules/FindPkgConfig.cmake:643 (_pkg_check_modules_internal)
  server/supernova/CMakeLists.txt:152 (pkg_check_modules)


-- Audio API (supernova): jack
-- Configuring incomplete, errors occurred!
```

dvzrv:
> Got new laptop:
> 
> Building for AUDIOAPI jack:

For the sake of the argument, can you try again after installing [pkgconf](https://github.com/pkgconf/pkgconf/#compiling-pkgconf-and-libpkgconf-with-cmake-usually-for-windows)?

sonoro1234:
> For the sake of the argument, can you try again after installing [pkgconf](https://github.com/pkgconf/pkgconf/#compiling-pkgconf-and-libpkgconf-with-cmake-usually-for-windows)?

I guess it could work, but the point is: I was able to build without pkgconf earlier, why having an extra dependency is seen as an improvement?

I never have need pkgconf for  building any program in windows.

I think it should try to work without pkgconfig installed also. (In case it is not found, try with find_package)

dyfer:
@dvzrv would it be possible to make a change so that jack include/library paths can be passed as cmake config flags, in which case pkgconfig is not used?

dvzrv:
> @dvzrv would it be possible to make a change so that jack include/library paths can be passed as cmake config flags, in which case pkgconfig is not used?

Yes, I'll look into that as soon as time permits. It shouldn't require a lot of changes though I hope.

dyfer:
Hello @dvzrv 
I wanted to loop back to this, would you be able to look into finishing this PR?

dyfer:
Hello @dvzrv, I wanted to loop back on this. We might be gearing up for new release (RC) in the next few weeks (TBD), are you planning to work on this PR?

dyfer:
Any update on this @dvzrv ?

dvzrv:
> Any update on this @dvzrv ?

@dyfer Sorry, I have forgotten about this.

I have now rebased and included a hack for Windows (also adding a NOTE), which allows setting `JACK_LINK_LIBRARIES` and `JACK_INCLUDE_DIRS` via providing the `_JACK_LINK_LIBRARIES` and `_JACK_INCLUDE_DIRS` options to cmake.

@sonoro1234 This will require testing on your side, but it worked for me when testing on Linux (without the check for `WIN32` of course). My assumption this should "just work", but I think this is rather hacky and as noted by jack upstream should not be done this way. If you have time adding a proper weakref solution for windows in the future (plus CI integration), that'd be ace!

server/scsynth/CMakeLists.txt
```diff
@@ -38,10 +38,16 @@ if(NOT AUDIOAPI MATCHES "^(jack|coreaudio|portaudio|bela)$")
 endif()
 
 if(AUDIOAPI STREQUAL jack)
-  find_package(Jack)
-  if(NOT JACK_FOUND)
-    message(FATAL_ERROR "Jack selected as audio API, but development files not found")
+  # NOTE: this is a hack for Windows, which should rely on a weakref link against jack instead
```
dyfer, line 41: IMO the suggestion here is only one of possible solutions.
IIUC, the workaround (or hack, whatever) is provided to be able to build with jack _without relying on PkgConfig_. While regular linking might be suboptimal on Windows, it should technically work at least when building and running locally. And even if weak linking was used, we'd still need to provide jack headers, which might also need to be found somehow, if they are not provided with the library that provides weak linking.

This is not very important and the comment can stay as is. However, IMO, a more accurate note would state something like `this is a workaround for Windows if PkgConfig is not available`

dyfer (approved):
I managed to build this on Windows with MinGW 8.1 using following cmake invocation:
```
cmake -G "MinGW Makefiles" -D SUPERNOVA=OFF -D SC_USE_QTWEBENGINE=OFF -D AUDIOAPI=jack -D _JACK_LINK_LIBRARIES="c:/Program Files/JACK2/lib/libjack64.lib" -D _JACK_INCLUDE_DIRS="c:/Program Files/JACK2/include" ..
cmake --build . --config RelWithDebInfo --target install
```
I couldn't then run scsynth as I was getting an error, but it looks like this has something to do with my build environment, as I got the same result when building with PortAudio backend (and still MinGW):
```
>Install\SuperCollider\scsynth.exe -u 57110
Mingw-w64 runtime failure:
  VirtualProtect failed with code 0x677
```
Since this seemed to build fine, and we can't even build with jack backend with MSVC (due to dependency on `sys/time.h` I think), and the build I attempted completed, I assume this is enough to make it work for those who need it.

I'll wait a bit in case @sonoro1234 has any input on this and I'll plan to merge it in a day or two.

dvzrv:
Will reschedule the failing tests. I believe they are flaky (as I only changed the comment...).

dyfer:
Yes, no worries about these test, the failure is not related to this PR. One day we might improve reliability of the test suite...
Thanks again for the PR!

dyfer:
Merging. Thanks so much @dvzrv !

PR #5686, opened 2022-01-09
ndef uses embedded specs for guis

### LINKED ISSUE #5515: NdefGui (EnvirGui) ignores local specs from NamedControls
## Environment

* Operating system: Arch Linux
* Other details (Qt version, audio driver, etc.): scnvim

## Steps to reproduce

```supercollider
// In this gui, the range of the \fff parameter 10 to 4000.
Ndef(\1, {SinOsc.ar(freq: \fff.kr(200, spec: Spec.specs[\freq]))}).gui;

// Checking local specs:
Ndef(\1).specs;

// -> ( 'fff': a ControlSpec(20, 20000, 'exp', 0, 440, " Hz") )

```

The problem seems to be in EnvirGui's getSpec method:
https://github.com/supercollider/supercollider/blob/de60e3405b417779bd35c19add1c6b2cca3bd194/SCClassLibrary/JITLib/GUI/EnvirGui.sc#L324-L335

As a user I don't think having to install JITLibExtensions to get correct behaviour is the expected way for this to work. 

I experimented with overriding the method to adding a check to see if the object contains a spec and if so use it:

```
+ EnvirGui {
  getSpec { |key, value|
    var spec = specs[key] ?? object.specs.at(key) ?? { key.asSpec };
    if (spec.isNil) {
      spec = Spec.guess(key, value);
      specs.put(key, spec);
    };
    ^spec
  }
}
```

This seems to fix it for me - but not sure if this breaks other things.

adcxyz:
## Purpose and Motivation

Fixes #5515.

Currently, NdefGui does not find controlspecs embedded in synth functions, 
thus NdefGui cannot use such specs for its slider ranges.
With this PR, core JITLib NdefGui finds these specs and uses them. 

## Types of changes

- Bug fix

```
// Ndef with an embedded spec:
n = Ndef(\1, { Saw.ar(freq: \fff.kr(200, spec: [50, 500, \exp])) * 0.1 });
// before this PR, this gui uses guessed range of 10-4000
n.gui;

// with this PR, ndef finds the embedded spec:
n.findFirstSpecFor(\fff);
// also in the specs method
n.specs;
// and the gui uses the 50 - 500 range given in the embedded spec:
n.gui;
```
Note: this fix works for plain SC without JITLibExtensions quark;
JITLibExtensions update to use it is simple, and ready to go. 

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -667,11 +667,8 @@ SynthDef {
 	}
 
 	findSpecFor { |controlName|
-		^if(metadata[\specs].isNil) {
-			nil
-		} {
-			metadata[\specs][controlName]
-		}
+		var specs = metadata[\specs];
+		^specs !? { specs[controlName] }
 	}
 
 	specs {
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -666,6 +666,14 @@ SynthDef {
 		^desc
 	}
 
+	findSpec { |name|
+		^if(metadata[\specs].isNil) {
+			nil
+		} {
+			metadata[\specs][name]
+		}
+	}
+
 	specs {
 		if(metadata[\specs].isNil) { metadata[\specs] = () };
 		^metadata[\specs]
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -666,11 +666,11 @@ SynthDef {
 		^desc
 	}
 
-	findSpec { |name|
+	findSpecFor { |controlName|
 		^if(metadata[\specs].isNil) {
 			nil
 		} {
-			metadata[\specs][name]
+			metadata[\specs][controlName]
 		}
 	}
 
--- a/SCClassLibrary/JITLib/ProxySpace/ProxyInterfaces.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/ProxyInterfaces.sc
@@ -413,8 +413,8 @@ SynthDefControl : SynthControl {
 		bytes = control.bytes; // copy cached data
 	}
 
-	findSpec { |name|
-		^synthDef.findSpec(name)
+	findSpecFor { |controlName|
+		^synthDef.findSpecFor(controlName)
 	}
 	specs {
 		^synthDef.specs
--- a/SCClassLibrary/JITLib/ProxySpace/ProxyInterfaces.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/ProxyInterfaces.sc
@@ -413,6 +413,9 @@ SynthDefControl : SynthControl {
 		bytes = control.bytes; // copy cached data
 	}
 
+	findSpec { |name|
+		^synthDef.findSpec(name)
+	}
 	specs {
 		^synthDef.specs
 	}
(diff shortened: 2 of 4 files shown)


SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -665,6 +665,14 @@ SynthDef {
 		^desc
 	}
 
+	findSpecFor { |controlName|
```
telephon, line 668: maybe we want to avoid a double lookup?
```supercollider
var spec = metadata[\specs];
^spec !? { spec[controlName] }
```

SCClassLibrary/JITLib/ProxySpace/ProxyInterfaces.sc
```diff
@@ -413,6 +413,9 @@ SynthDefControl : SynthControl {
 		bytes = control.bytes; // copy cached data
 	}
 
+	findSpecFor { |controlName|
+		^synthDef.findSpecFor(controlName)
+	}
 	specs {
 		^synthDef.specs
 	}
```
telephon, line 421: You could even add one for `SynthControl`:

```supercollider
findSpecFor { |controlName|
var specs = this.synthDesc.specs;
^specs !? { specs[controlName] }
}
```

Then you can use the specs of `Synth`s played like `Ndef(\x, \default)`

SCClassLibrary/Common/Audio/SynthDef.sc
adcxyz: yes, better, thanks

SCClassLibrary/JITLib/ProxySpace/ProxyInterfaces.sc
adcxyz: Should SynthDesc also have its own  findSpecFor method?
SynthDesc is used in many other places, so could be useful, no?

SCClassLibrary/Common/Audio/SynthDef.sc
adcxyz: done

adcxyz:
ah, just saw that this portaudio file slipped into the PR by accident!
 - any ideas how to remove it?

SCClassLibrary/JITLib/ProxySpace/ProxyInterfaces.sc
telephon: yes, that seems reasonable.

telephon:
I would try this:

copy the commit SHA of the version before the commit, then do a `git reset <SHA>`. Then you can use `Discard Changes` in the GitHubDesktop to throw away that portaudio file change. Then commit. And then do a force push (`git push --force`).

well, at least what I would try.

telephon (approved):

joshpar:
@adcxyz can you please resolve formatting issues? After that I will merge this in.

joshpar:
ooops! accidentally closed it! reopened for reformatting.

dyfer:
It seems that the problem is not with the formatting itself, but with the state of this branch (there were no changes to cpp files; also there's a rouge submodule change here). As far as I can tell, this branch needs to be rebased on the develop. @adcxyz would you do that or should I do it and force push?

joshpar:

telephon:
To me, this looks good. It would be better if `findSpec` were added to `SynthDesc` too, but fine like this as well.

dyfer:
In the interest of moving this forward, I've rebased this on `develop` and will force push it now.

dyfer:
Looks like I was able to fix the submodule change. This should be clean to merge now.

adcxyz:
@dyer @telephon 
thanks :-)

PR #5687, opened 2022-01-09
Add UnitTest test_seti_nodeMap again

nuss:
The test passes locally but failes on GHA. Therefore it's been added to
gha_test_run_proto.json

Signed-off-by: Stefan Nussbaumer <st9fan@gmail.com>

## Purpose and Motivation

The test `test_seti_nodeMap` has always passed on my local machine but fails in GHA. So I've added the test to `gha_test_run_proto.json`

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/scripts/gha_test_run_proto.json
+++ b/testsuite/scripts/gha_test_run_proto.json
@@ -154,11 +154,11 @@
             "skip":true,
             "skipReason":"UnitTest sometimes fails when run in qpm"
         },
-	{
-	    "suite":"TestNodeProxySeti",
-	    "test":"test_seti_nodeMap",
-	    "skip":true,
-	    "skipReason":"test passes on local machine but fails in GHA for unknown reasons"
-	}
+        {
+            "suite":"TestNodeProxySeti",
+            "test":"test_seti_nodeMap",
+            "skip":true,
+            "skipReason":"test passes on local machine but fails in GHA for unknown reasons"
+        }
     ]
 }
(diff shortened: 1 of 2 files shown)


testsuite/scripts/gha_test_run_proto.json
```diff
@@ -153,6 +153,12 @@
             "test":"test_findText_notFound",
             "skip":true,
             "skipReason":"UnitTest sometimes fails when run in qpm"
-        }
+        },
+	{
```
telephon, line 157: is the formatting correct? YOur code seems to be inlined differently.

dyfer: ah good catch @telephon , the new addition has tabs while the rest of the file uses spaces. @nuss could you change the new code to spaces please?

nuss: hmm... I've seen that too in the diff. However, formatting looks correct in my forked repo: https://github.com/nuss/supercollider/blob/NodeProxy-seti/testsuite/scripts/gha_test_run_proto.json 
I've tried to keep my entry in line with other entries (indenting with spaces, not tabs). So, if formatting's not right I don't really know what's wrong.

nuss: ... sorry, I thought I had formatted my entry with tabs (in vim) but there were some tabs left.

dyfer (approved):
Looks good, thanks!

dyfer:
`test_synthDef_isReleased_afterFree ` got enabled in #5702

PR #5699, opened 2022-01-16
class library: NodeProxy converts mixed outputs instead of failing

telephon:
Instead of failing, we better convert a mixed output that contains an audio rate UGen and non-audio-rate UGens, because this is hard to understand. There is an informative warning.

## Purpose and Motivation

See discussion here: https://scsynth.org/t/why-does-multichannel-expansion-not-work-for-mul-in-nodeproxy/5232

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing


adcxyz (approved):
good idea to catch this and convert.

PR #5702, opened 2022-01-19
Fix multiple UnitTests

dyfer:
## Purpose and Motivation

This PR fixes a number of issues with UnitTests, particularly for running them in the CI:
- most (all?) tests using a server now use a unique server instance
  - the use of default server, particularly if it was not shut down, has led to multiple tests failing, some without reporting failure in the CI (i.e. with the [Did not complete](https://github.com/supercollider/supercollider/runs/4808924636?check_suite_focus=true#step:8:12377) message)
  - `TestFunction -test_plot` arguably was hiding the server booting most covertly...
- a number of tests was fixed
- the link to the API quark, used by QPM for testing was updated to point to the address that's specified in the quarks directory
  - I'm not sure if this was causing any of the issues we experience in qpm or not...
- we now have ~~3483~~ **3507** passing tests _with no tests that did not finish_
  - the number of tests mysteriously failing in CI has greatly dropped; it appears that most of these failures were due to issues with booting the server or using the default server
  - there's a handful of tests still excluded, I tried updating lists in #5359 and #5360
- I've made [some changes to qpm](https://github.com/supercollider/qpm/compare/topic/testclass-setup-teardown...supercollider:topic/ignore-persistent-ui-error) and pointed our CI at its new branch, most notably increasing overall test suite timeout from 10 to 20 minutes (macOS tests slightly exceed 10 minutes right now)

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -590,7 +590,7 @@ jobs:
 
           # install API quark
           mkdir $QUARKS_PATH && cd $QUARKS_PATH
-          git clone --depth=1 https://github.com/supercollider-quarks/API
+          git clone --depth=1 https://github.com/crucialfelix/API
 
           # make working copy of the script
           cp $SCRIPT_PROTO $SCRIPT_RUN
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -543,7 +543,7 @@ jobs:
       SCLANG: ${{ github.workspace }}/${{ matrix.sclang }}
       SCRIPT_PROTO: ${{ github.workspace }}/testsuite/scripts/gha_test_run_proto.json
       SCRIPT_RUN: ${{ github.workspace }}/testsuite/scripts/run/gha_test_run.json
-      QPM_URL: git+https://github.com/supercollider/qpm.git@topic/testclass-setup-teardown
+      QPM_URL: git+https://github.com/supercollider/qpm.git@topic/ignore-persistent-ui-error
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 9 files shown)


nuss (approved):

dyfer:
This PR is now ready for review :)

dyfer:
`TestServer_boot` and `TestServer_clientID*` tests were effectively fixed in #5715, so I re-enabled them here.
Ready for further review :)

.github/workflows/actions.yml
```diff
@@ -590,7 +590,7 @@ jobs:
 
           # install API quark
           mkdir $QUARKS_PATH && cd $QUARKS_PATH
-          git clone --depth=1 https://github.com/supercollider-quarks/API
+          git clone --depth=1 https://github.com/crucialfelix/API
```
joshpar, line 593: is there a repo (other than someone specifically, who could just destroy it) that we can use as a test? I don't think the crucial lib is going away, but it stood out a little

joshpar (changes requested):
Basically approved but want to consider the crucialfelix bit... otherwise, fine from my side to merge in

.github/workflows/actions.yml
dyfer: Using the API quark is part of the whole qpm design (for using json files). I'm not a fan of using any quark for testing, but if we use one I think it's better if it's the currently maintained version.

IMO the long term solution is to re-consider using qpm and how we run tests to possibly not have to use a quark altogether.

joshpar (approved):

PR #5708, opened 2022-01-28
Possibly Fixes #2954 

### LINKED ISSUE #2954: \.isPrimitiveName is true
## Problem

`Symbol:isPrimitiveName` is supposed to test for a string that begins with `_`, but it returns true for an empty symbol.

## Solution

`Symbol:isPrimitiveName` calls the undocumented primitive method `Symbol:isPrefix`, which appears to be the source of this bug. it's found in `prSymbolIsPrefix` in `PyrSymbolPrim.cpp`. it compares the strings to see if theyre equal up to the minimum of their two lengths, so it can return a false positive if the receiver is shorter than the argument.

also, whoever's fixing this should take the opportunity to document `Symbol:isPrefix`.

Brandon-Yip2:
The situation presented in #2954 stemmed from Symbol:isPrimitiveName testing a string and returning true given an empty symbol. Checking the minimum length between the two symbol and making sure it isn't 0 might fix this. The prefix method still works as intended but now doesn't accept empty symbols as input.

### DIFF
--- a/lang/LangPrimSource/PyrSymbolPrim.cpp
+++ b/lang/LangPrimSource/PyrSymbolPrim.cpp
@@ -43,7 +43,7 @@ int prSymbolIsPrefix(struct VMGlobals* g, int numArgsPushed) {
     int32 alen = slotRawSymbol(a)->length;
     int32 blen = slotRawSymbol(b)->length;
     length = sc_min(alen, blen);
-    if (memcmp(slotRawSymbol(a)->name, slotRawSymbol(b)->name, length) == 0) {
+    if (memcmp(slotRawSymbol(a)->name, slotRawSymbol(b)->name, length) == 0 && length > 0) {
         SetTrue(a);
     } else {
         SetFalse(a);


telephon (approved):
looks good!

dyfer:
Thanks @Brandon-Yip2 
This seem to work as intended - `''.isPrimitiveName` now returns `false`.
@mossheim suggested that a desired fix would include adding documentation for `Symbol -isPrefix`. I can only guess what should go in there, as I haven't dealt with this part of the project much.
@telephon do you think that this PR should also include above mentioned documentation or should we merge as is?

telephon:
It doesn't have to include this extra documentation. But it would be a good occasion.

Here is the bit I would propose to add:

```

method::isPrefix

Answer whether the symbol is a prefix of another one

code::
'a'.isPrefix('all'); // true
'z'.isPrefix('all');  // false
::


```

dyfer:
All right, I'll merge and I'll make a PR for documentation

PR #5710, opened 2022-01-31
Deprecate QuartzComposerView

dyfer:
## Purpose and Motivation

Quartz Composer bindings are not available in Qt 6. We currently plan to drop Quartz Composer once we move to Qt 6, so we'll deprecate it in SC 3.13.

## Types of changes

- Deprecation

## To-do list

- [x] Code is tested
- [x] Updated documentation


telephon (approved):

wtkns:
not sure if this could be related?

ERROR: Superclass 'View' of class 'QuartzComposerView' is not defined in any file.

https://github.com/supercollider/supercollider/issues/5731

PR #5711, opened 2022-01-31
Topic update seminaire

telephon:
## Purpose and Motivation

Improvements to the aural comprehension application "séminaire musical" in the examples folder.

There were bugs fixed in time and beats test.

## To-do list

- [x] Code is tested

### DIFF
--- a/examples/research_and_tools/SeminaireMusical.scd
+++ b/examples/research_and_tools/SeminaireMusical.scd
@@ -307,12 +307,11 @@ f_neu = {
 		// beatings
 		5, {
 			referenceToneIsPlaying = false;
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks! This looks good to me, though I'm not very familiar with the original.
Just FYI, when piano tone is selected (e.g. with "notes") and then one switches to "beatings", then I get an error and a large (DC?) signal after trying to listen to any of the beatings. Otherwise works well AFAICT

telephon:
ah thank you for testing – I've fixed it, should work now.

dyfer (approved):
Thanks!

Issue #1455, opened 2015-05-10
Readline/Command line mode not working on Windows


bagong:
The situation is somewhat different in the MinGW and VS build:
- in the MinGW build you get a proper prompt and your input is echoed, but nothing happens. None of the control keys does anything too
- in the VS Studio build sclang crashes when started in cmd-line mode. It happens at a very late stage of the startup process. You still get the full list of sclang boot messages, but before the prompt displays it crashes.

bagong:
I have inspected this with more detail now. I have to say first, the situation is quite different from the one described above. If you use readline 5.0.1 as provided by GnuWin32 rather than a readline vs6 from MinGW. With the earlier it *almost* works in the MinGW-32 build (in the 32-bit VS build you get a link error).

In order to get input, you have to keep your finger on a key for something like half a second. Then the keypress is registered (if you wait too long, the key repeats), if you press a key briefly/naturally, it will usually register after the second keystroke. Things like enter, ctrl-x and ctrl-t work (if you wait on the key). On ctrl-t you notice most clearly that there is a wait time, before the key registers.

ctrl-d (quit) works, but crashes sclang

Exiting sclang (ctrl-D)
sc3> main: waiting for input thread to join...
main: quitting...
empty
cleaning up OSC
terminate called without an active exception

And then the system pops in and tells that this was not a nice ending.

There is a readline-Windows related "fixme" in SC_TerminalClient::readlineInit()

    // FIXME: Implement the code below on Windows
    // #ifndef _WIN32
    // Set our handler for SIGINT that will clear the line instead of terminating.
    // NOTE: We prevent readline from setting its own signal handlers,
    // to not override ours.
    rl_catch_signals = 0;
    struct sigaction sact;
    memset( &sact, 0, sizeof(struct sigaction) );
    sact.sa_handler = &sc_rl_signalhandler;
    sigaction( SIGINT, &sact, 0 );
    // #endif

If you try to compile the code commented out in Windows, you get following errors:

    C:\Users\Rainer\Projects\sc\supercollider\lang\LangSource\SC_TerminalClient.cpp:507: error: aggregate 'SC_TerminalClient::readlineInit()::sigaction sact' has incomplete type and cannot be defined
      struct sigaction sact;
    C:\Users\Rainer\Projects\sc\supercollider\lang\LangSource\SC_TerminalClient.cpp:508: error: invalid application of 'sizeof' to incomplete type 'SC_TerminalClient::readlineInit()::sigaction'
[...]

PR #5712, opened 2022-01-31
Readline on Windows

### LINKED ISSUE #2171: Windows / readline / MSVC
As part of the effort of bringing the Windows version on par to Linux and Mac, I have been having a look at the readline stuff for MSVC.

I have seen that there is a library called linenoise-ng, that we could use for this. I have been doing some initial testing and it looks promising. The only potential problem is that it does not have a mode to give it chars as they come (like the `rl_callback_read_char`). It has a similar interface to the vanilla `readline()`.

As I said I have been doing some testing, and it works fine. I have changed the original code:

```
mStdIn.async_wait( [&] (const boost::system::error_code & error) {
        if(error)
            onInputRead(error, 0);
        else {
            DWORD bytes_transferred;

            ::ReadFile(GetStdHandle(STD_INPUT_HANDLE),
                       inputBuffer.data(),
                       inputBuffer.size(),
                       &bytes_transferred,
                       nullptr);
            onInputRead(error, bytes_transferred);
        }
    });
```

to:

```
char* result = linenoise("sc3>");
    if (result){
        strncpy(inputBuffer.data(), result, inputBuffer.size());
        inputBuffer[strlen(result)] = kInterpretPrintCmdLine;
        inputBuffer[strlen(result) + 1] = 0;
        DWORD bytes_transferred = strlen(result) + 1;
        onInputRead(boost::system::error_code(), bytes_transferred);
    }
```

And everything seems to work fine.

Now I have a question: what is the reason to have readline work in async mode? Since it is running in its separate thread, couldn't we just use the direct `readline()` function call? I am probably missing something here...

If this is possible, we could probably take this further and have a common version of readline-like functionality using the linenoise-ng library?

Any thoughts? If people are interested I can publish my branch or do a PR.

@timblechmann, I see you were involved in this part of the code? Any insights?

dyfer:
## Purpose and Motivation

I checked what's the current status of GNU Readline on Windows and I found out that it is available through vcpkg. After small adjustments I was able to add it to our Windows build. ~~However, there is still an issue with the keyboard input, see below.~~ I think I also fixed the keyboard input. (My initial findings were inaccurate: I think the issue was with trying to read input when there was nothing to read.) 

Readline from vcpkg currently works with msvc builds only, since build errors occur when it's being built with mingw in vcpkg.

This will fix #2171 I think.

(outdated comments hidden below)
<details>

~~Outstanding issue with `stdin`~~

~~Even though readline seems to work fine, the keyboard requires pressing a key _twice_ in order for it to be registered. This is consistent with what @bagong [reported some time ago](https://github.com/supercollider/supercollider/issues/1455#issuecomment-194192283#:~:text=In%20order%20to,the%20key%20registers.).~~

~~I have investigated the issue a little bit and it seems that the [callback in .async_wait](https://github.com/supercollider/supercollider/blob/develop/lang/LangSource/SC_TerminalClient.cpp#L544-L555) gets called only every other keystroke (`mStdIn` is a [boost asio object_handle](url)). I couldn't find many examples online how to read asynchronously from stdin on windows using boost asio, so I'm leaving this issue for now. Maybe somebody has ideas how to fix this? @bagong or @jrsurge ? :)~~

~~(Please note, this issue is not related to a [fixme referred to by @bagong](https://github.com/supercollider/supercollider/blob/develop/lang/LangSource/SC_TerminalClient.cpp#L521) AFAICT)~~
</details>

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Windows builds compile with readline in GitHub Actions
- [x] Issue with asynchronous read of stdin is fixed
- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -491,7 +491,10 @@ jobs:
 
           mkdir $BUILD_PATH && cd $BUILD_PATH
 
-          cmake -G "${{ matrix.cmake-generator }}" -A "${{ matrix.cmake-arch }}" -D CMAKE_PREFIX_PATH="$Qt5_DIR;$VCPKG_INSTALLATION_ROOT/installed/${{ matrix.vcpkg-triplet }}" -D SUPERNOVA=ON -D SC_USE_QTWEBENGINE=${{ matrix.use-qtwebengine }} -D SC_ABLETON_LINK=${{ matrix.ableton-link }} -D CMAKE_BUILD_TYPE=Release .. # build type is specified here for MinGW build
+          EXTRA_CMAKE_FLAGS=
+          if [[ -n "${{ matrix.vcpkg-triplet }}" ]]; then EXTRA_CMAKE_FLAGS="-D CMAKE_TOOLCHAIN_FILE=$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake $EXTRA_CMAKE_FLAGS"; fi
+
+          cmake -G "${{ matrix.cmake-generator }}" -A "${{ matrix.cmake-arch }}" -D CMAKE_PREFIX_PATH=$Qt5_DIR -D SUPERNOVA=ON -D SC_USE_QTWEBENGINE=${{ matrix.use-qtwebengine }} -D SC_ABLETON_LINK=${{ matrix.ableton-link }} -D CMAKE_BUILD_TYPE=Release $EXTRA_CMAKE_FLAGS .. # build type is specified here for MinGW build
 
       - name: build
         shell: bash
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -488,13 +488,11 @@ jobs:
         shell: bash
         run: |
           export PATH=$Qt5_DIR/bin:$PATH
+          export VCPKG_ROOT=$VCPKG_INSTALLATION_ROOT
 
           mkdir $BUILD_PATH && cd $BUILD_PATH
 
-          EXTRA_CMAKE_FLAGS=
-          if [[ -n "${{ matrix.vcpkg-triplet }}" ]]; then EXTRA_CMAKE_FLAGS="-D CMAKE_TOOLCHAIN_FILE=$VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake $EXTRA_CMAKE_FLAGS"; fi
-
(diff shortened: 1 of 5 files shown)


joshpar (approved):

dyfer:
@joshpar I think @jrsurge had some comments about the CMake integration for this one. It might not be implemented the most proper way... but then again, this seems to work fine.

joshpar:
> @joshpar I think @jrsurge had some comments about the CMake integration for this one. It might not be implemented the most proper way... but then again, this seems to work fine.

Thanks @dyfer - I'm not merging anything yet, but I agree... it looks about right to me, but if there is a change someone else thinks is important, that's fine. I'll target pulling approved PRs in at the dev meeting on 2-28-2022 if no one else objects.

dyfer:
Hi @jrsurge 
In follow up on our conversation during the last dev meeting, I looked more closely at vcpkg documentation and tested how it behaves on a local windows machine. Here are my findings (please correct me if I'm wrong):
- vcpkg integration (enabled with `vcpkg integrate install`) provides projects in Visual Studio with information about the installed libraries, but AFAICT it _does not_ provide system-wide package discovery when using cmake from the command line
  - neither the integration command, nor the bootstrap script sets any environment variables AFAICT; on my system vcpkg executable is not in PATH and `VCPKG_ROOT` is _not_ set
- it seems that the most "automatic" way for cmake to discover vcpkg is the combination of 1) setting `VCPKG_ROOT` and 2) adding the following to project cmake, as shown [here](https://vcpkg.io/en/docs/users/integration.html#using-an-environment-variable-instead-of-a-command-line-option) (probably more elegant than specifying toolchain file manually):
```cmake
if(DEFINED ENV{VCPKG_ROOT} AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
  set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
      CACHE STRING "")
endif()
```

I also have some observations about the [manifest mode](https://vcpkg.io/en/docs/users/manifests.html). Since we are using both vcpkg and other means to install dependencies on Windows, and since on other platforms we require users to install dependencies manually before building, I feel that using manifest mode for readline only is an overkill right now. I would also be happy to re-consider if if we switch to using vcpkg for other libraries.

I propose to proceed as follows:
- add the mentioned above CMake entry to our `CMakeLists.txt` (probably for `WIN32` only?)
- provide following instructions to users who want to build with readline support:
  - install vcpkg (including runnig the bootstrap script)
  - install readline using vcpkg
  - set `VCPKG_ROOT` env variable
- implement it appropriately in CI, which will be easy, since:
  1. I've already added readline vcpkg installation
  2. With the changes to CMakeLists the configure step would not need to have any additional flags
     - GHA runners do have `VCPKG_ROOT` already set

What do you think?

dyfer:
After exchanging DMs with @jrsurge I proceeded with suggested changes.

I decided to check for `VCPKG_ROOT` without a platform check, as I think we might want to be able to use VCPKG on other platforms at some point.

The PR is now ready for review.

joshpar:
@jrsurge - If you have time to take a look, please do. i think this looks fine though. If we haven't heard back from you by April 23rd we'll merge with my approval

Issue #5634, opened 2021-12-07
Memory leak: UGen Ctors get called over and over again when they throw exceptions (e.g. alloc failed)


elgiano:
## Environment

* SuperCollider version: 3.12
* Operating system: Linux
* Other details (Qt version, audio driver, etc.): Jack

## Steps to reproduce
I encounter this using PartConv, but I suspect it's something more general.
```supercollider
// allocate a big IR buffer
b = Buffer.alloc(s, PartConv.calcBufSize(2048, (numFrames: 44100 * 300)).asInteger);
// this is too much for my 8GB ram, trying to play PartConv causes alloc failed
x = { PartConv.ar(Silent.ar, 2048, b) }.play;
// if x runs for a little while, it eventually prints "server disconnected shared memory interface"
x.free;
// then if I try to allocate anything else, I still get "alloc failed"
{DelayN.ar(Silent.ar, 1)}.play
```

on supernova:

```supercollider
{DelayN.ar(Silent.ar, 100)}.play
// -> Failed to allocate memory for  ugen.
// Server 'localhost' exited with exit code 0.
// server 'localhost' disconnected shared memory interface
```

## Expected vs. actual behavior
I would expect `x.free` to cleanup as if the "too big" memory allocation never happened, so that I can still allocate a smaller amount of memory.
What I observe is that on failed allocation, the UGen constructor runs repeteadly (once for every time the alloc failed error is printed), and the destructor never runs (I observed this by adding a few Print statements in UGens constructors and destructors). Destructors are not called even when the synth or the whole Node Tree is freed.
This is by the way how error messages are "flooding" (relating to #3563): the error occurs again and again at every processing block.

I observed this looping behavior also with DelayN, but I can't get to the "disconnected shared memory interface" point with DelayN.

On supernova the server crashes when I try to allocate too much memory.... I assume it might be linked but it looks also like a different issue to me.

elgiano:
What I've figured out so far:

- When a new Synth is created, scsynth calls Graph_New, which sets the graph's mCalcFunc to Graph_FirstCalc
- Graph_FirstCalc calls all the UGens constructors, and then sets the graph's mCalcFunc to Graph_Calc
- If any of the Ctors fails by throwing an exception (such as the alloc failed error), Graph_FirstCalc exits without changing the graph's mCalcFunc, so FirstCalc gets called again at the next processing block, calling again all constructors

[...]

PR #5713, opened 2022-02-03
Remove RTAlloc exceptions, and review all plugins' RTAlloc/RTFree

### LINKED ISSUE #782: plugins: review error handling of memory allocations
many plugins don't perform null pointer checks to ensure that rt memory allocation actually succeeds. instead of a nice error message, the effect is usually a server crash

### LINKED ISSUE #3563: Error message flooding on failed RTAlloc with JACK
running the following code:

    { DelayN.ar(DC.ar(0), 1000) }.play;

attempts to allocate 1000 seconds of delay line memory. it doesn't crash the server (unlike some ugens -- see #782). it prints an error message:

    JackDriver: exception in real time: alloc failed, increase server's memory allocation (e.g. via ServerOptions)

the problem is not the message, but the rate -- it *floods* the post window as long as the ugen is running. it should only print once.

elgiano:
## Purpose and Motivation
This PR is backed by a RFC: https://github.com/supercollider/rfcs/pull/18

Fixes #3563 and #5634
Partially fixes #782 (sc3-plugins are not affected)

If real time allocation fails during a UGen constructor, an exception is thrown and the enclosing synth can't exit a loop trying to construct all its UGens over and over again, possibly leading to memory leaks. During the discussion, it was proposed to remove the throwing mechanism entirely, mainly because incompatible with plugin boundaries (sorry for the sloppy terminology here, for more details, please see the RFC).

This PR follows the program from the RFC:
NOTE: turns out `RTFree` doesn't need extra NULL checks, as they are already included in `AllocPool:Free`. The new macro `RTFreeSafe` proposed in the RFC is thus omitted.
- [x] Remove the throw from `RTAlloc`, return `NULL` instead
- [x] Handle possible `NULL` returns and eventually re-throw an exception in all places in `sclang` and `scsynth` where it's relevant
- [x] Introduce macros for safer `RTAlloc`, implement them in all server plugins, everywhere `RTAlloc` or `scfft_create` are used. The new macros are:
- `ClearUnitOnMemFailed`: disables unit's processing function, sets `mDone=true`, prints a nice error message and returns from the enclosing function (to exit early from Ctors when RTAlloc fails)
- `ClearUnitIfMemFailed`: calls the previous macro if the passed pointer is NULL. Can be used passing multiple pointers by chaining them with `&&` operator.
- [x] Update documentation for devs and plugin writers

## Types of changes

- Bug fix
- Documentation

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/include/plugin_interface/FFT_UGens.h
+++ b/include/plugin_interface/FFT_UGens.h
@@ -62,6 +62,17 @@ static inline SCComplexBuf* ToComplexApx(SndBuf* buf) {
 
 struct PV_Unit : Unit {};
 
+// Ordinary ClearUnitOutputs outputs zero, potentially telling the IFFT (+ PV UGens) to act on buffer zero, so let's
+// skip that:
+static inline void FFT_ClearUnitOutputs(Unit* unit, int wrongNumSamples) { ZOUT0(0) = -1; }
+
+#define ClearFFTUnitIfMemFailed(condition)                                                                             \
+    if (!(condition)) {                                                                                                \
+        Print("%s: alloc failed, increase server's RT memory (e.g. via ServerOptions)\n", __func__);                   \
+        SETCALC(FFT_ClearUnitOutputs);                                                                                 \
+        unit->mDone = true;                                                                                            \
+        return;                                                                                                        \
+    }
 
 #define sc_clipbuf(x, hi) ((x) >= (hi) ? 0 : ((x) < 0 ? 0 : (x)))
 
--- a/include/plugin_interface/SC_Unit.h
+++ b/include/plugin_interface/SC_Unit.h
@@ -80,6 +80,17 @@ enum { kUnitDef_CantAliasInputsToOutputs = 1 };
 // set the calculation function
 #define SETCALC(func) (unit->mCalcFunc = (UnitCalcFunc)&func)
 
+#define ClearUnitOnMemFailed                                                                                           \
+    Print("%s: alloc failed, increase server's RT memory (e.g. via ServerOptions)\n", __func__);                       \
+    SETCALC(*ClearUnitOutputs);                                                                                        \
+    unit->mDone = true;                                                                                                \
+    return;
+
+#define ClearUnitIfMemFailed(condition)                                                                                \
+    if (!(condition)) {                                                                                                \
+        ClearUnitOnMemFailed                                                                                           \
+    }
+
 // calculate a slope for control rate interpolation to audio rate.
 #define CALCSLOPE(next, prev) ((next - prev) * sc_typeof_cast(next) unit->mRate->mSlopeFactor)
 
--- a/HelpSource/Guides/WritingUGens.schelp
+++ b/HelpSource/Guides/WritingUGens.schelp
@@ -161,8 +161,6 @@ Unit generator plugins are called from the real-time context, which means that s
 to avoid audio dropouts.
 
 definitionList::
-## Memory Allocation || Do not allocate memory from the OS via code::malloc:: / code::free:: or code::new::/ code::delete::.
-                        Instead you should use the real-time memory allocator via code::RTAlloc:: / code::RTFree::.
 ## STL Containers || It is generally not recommended to use STL containers, since they internally allocate memory. The only
                      way the STL containers can be used is by providing an Allocator, which maps to the allocating functions of
                      the server.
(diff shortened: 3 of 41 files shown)


include/plugin_interface/SC_Unit.h
```diff
@@ -80,6 +80,17 @@ enum { kUnitDef_CantAliasInputsToOutputs = 1 };
...
+#define ClearUnitOnMemFailed                                                                                           \
+    Print("%s: alloc failed, increase server's RT memory (e.g. via ServerOptions)\n", __func__);                       \
+    SETCALC(*ClearUnitOutputs);                                                                                        \
+    unit->mDone = true;                                                                                                \
+    return;
+
+#define ClearUnitIfMemFailed(ptr)                                                                                      \
+    if (ptr == nullptr) {                                                                                              \
+        ClearUnitOnMemFailed                                                                                           \
+    }
```
Spacechild1, line 92: The macro expects a pointer, but you sometimes use it with a boolean expression, e.g. https://github.com/supercollider/supercollider/pull/5713/files#diff-06296acaab6425bef65d947ed70f147408a421060d32ed7601c9b18654bdf802R143
You shouldn't really compare a `bool` with `nullptr`.

I think it's better to just take a boolean expression instead of a pointer, so it works cleanly with both:
```
#define ClearUnitIfMemFailed(condition)   \
    if (!condition) {                     \
        ClearUnitOnMemFailed              \
    }
```

Spacechild1:
> This PR affects SC's plugin_interface and will thus introduce breaking changes

I don't see any breaking changes... `RTAlloc` and `RTRealloc` now just has the expected behavior.

Spacechild1:
> Introduce macros for safer RTAlloc/RTFree, implement them in all server plugins, everywhere RTAlloc, RTFree, scfft_create or scfft_destroy are used. The new macros are:
RTFreeSafe: calls RTFree only if the passed pointer is not NULL

I just looked again at the code of the `AllocPool::Free` and I noticed that calling it with `NULL` is well defined and has no effect. Same for the `rt_free` implementation in Supernova. So the `RTFreeSafe` macros and the check before `scfft_destroy` are not necessary...

I am very sorry for putting you on the wrong track!! I have no idea why I thought it would be necessary...

elgiano:
You're right, but I also always had in mind that those checks for RTFree were necessary. I guess I was on the wrong track on my own already :) I can easily remove that part from the PR. My best guess is that what created those crashes that I was explaining with freeing null pointers, was actually freeing uninitialized (garbage) pointers. So I'll also double check that path!

Thanks

elgiano:
> I don't see any breaking changes... `RTAlloc` and `RTRealloc` now just has the expected behavior.

I think I got a bit confused here, so could you help me clarifying these two points?
1. I'm only introducing new `#defines`, so this isn't actually an API extension, right?
2. Isn't removing the `throw` from `AllocPool:Alloc` an API change?

include/plugin_interface/SC_Unit.h
elgiano: Good point, thanks, and I also see we need parenthesis around `ptr` (or `condition`) in the `if` check.

I'm afraid my next question here can be a bit bikeshedding, but I'd like to evaluate two options here, to make this cleaner:
1. Rename `ClearUnitIfMemFailed` to something like `ClearUnitIfConditionFailed`, `ClearUnitIfFalse`, or `CheckMemOrClearUnit`, AssertOrClearUnit`, or anything else that would better describe that we just check a boolean condition there, and if it's false we consider it an allocation failure and clear the unit accordingly.
2. Leave the `if` to the user, and only introduce `ClearUnitOnMemFailed`. The drawback is that we loose authority on the if check.

Spacechild1: Ad 1: Since `ClearUnitIfMemFailed` calls `ClearUnitOnMemFailed` under the hood and the latter prints an error message about the RT memory pool, it's good to keep the `MemFailed` part and not use something generic like `ClearUnitIfFalse`.

Ad 2: First I thought the macro wouldn't be necessary, but it actually is quite handy.

Spacechild1:
> My best guess is that what created those crashes that I was explaining with freeing null pointers, was actually freeing uninitialized (garbage) pointers.

That would make sense.

> I'm only introducing new #defines, so this isn't actually an API extension, right?

Macros, inline functions, C++ helper classes, etc. belong to the plugin side and are not visible to the Server. *Adding* such things is generally not a problem.

> Isn't removing the throw from AllocPool:Alloc an API change?

In theory yes, but in our case, throwing an exception has always been the wrong behavior. Plugin authors expected `RTAlloc` to return `NULL` on failure. To quote from `AnalogEcho_Ctor` in `AnalogEcho.cpp`:
```
// This check makes sure that RTAlloc succeeded. (It might fail if there's not enough memory.)
// If you don't do this check properly then YOU CAN CRASH THE SERVER!
// A lot of ugens in core and sc3-plugins fail to do this. Don't follow their example.
if (unit->buf == NULL) {
```

include/plugin_interface/SC_Unit.h
elgiano: Great, I'll just fix the parenthesis then and leave the name as it is!

elgiano:
I removed the API change and API extension tags since we are considering this a bugfix, and it doesn't add or change the InterfaceTable

include/plugin_interface/SC_Unit.h
joshpar: @Spacechild1 - do you feel this is resolved now? Looks right to me.

Spacechild1: Yep, looks good! Thanks for the heads-up.

Spacechild1:
> However, they don't set the done flag and print their own error message.

Some do: https://github.com/supercollider/supercollider/blob/29550ee26aa55e0e1d2ea8f7ded5b0c59519e648/server/plugins/Convolution.cpp#L320
https://github.com/supercollider/supercollider/blob/29550ee26aa55e0e1d2ea8f7ded5b0c59519e648/server/plugins/DelayUGens.cpp#L772

Apart from that, there are many cases where resource acquisition (mostly Buffers) or any other kind of initialization error also sets `mDone` to `true`. Here's one example: https://github.com/supercollider/supercollider/blob/29550ee26aa55e0e1d2ea8f7ded5b0c59519e648/server/plugins/DelayUGens.cpp#L738.
Just grep for `mDone = true` and you will find more instances. Some of them are in macros, so they are actually used in multiple places.

> Does it make sense to give this new additional meaning to mDone?

As demonstrated above, it's not really a new additional meaning, so it should be fine. `mDone` just tells whether the UGen has stopped playing, but it doesn't actually tell the reason.

> Shall we make them call ClearUnitIfMemFailed, for consistency?

IMO yes!

dyfer:
hi @elgiano 
You might want to rebase your PR on the latest `develop` to resolve the conflict...

include/plugin_interface/FFT_UGens.h
```diff
@@ -62,6 +62,17 @@ static inline SCComplexBuf* ToComplexApx(SndBuf* buf) {
 
 struct PV_Unit : Unit {};
 
+// Ordinary ClearUnitOutputs outputs zero, potentially telling the IFFT (+ PV UGens) to act on buffer zero, so let's
+// skip that:
+static void FFT_ClearUnitOutputs(Unit* unit, int wrongNumSamples) { ZOUT0(0) = -1; }
```
Spacechild1, line 67: You might use `static inline` instead of just `static`, like in the functions above.

---

As a side note:

Specifying free functions in header files as `static` or `static inline` is just a hack for C.
For C++, the more idiomatic (and efficient) specifier would be just `inline`. Note that the meaning of `inline` is different for C and C++!
For some nice explanations, see https://stackoverflow.com/a/1759575/6063908 and https://stackoverflow.com/a/68146072/6063908

Since this header is really a C++ header (it uses C++ classes), we could actually substitute all those `static inline` with `inline`, but that's not the topic of this PR :-)

Spacechild1:
> Is this all right? And, sorry for not understanding still, but is it now an API change? @Spacechild1 :shy_emoji:

Looks good! No API change in sight :-)

> I've actually left mDone = true in the macro. On a further look it seemed intended...

Yes, see also my comment above https://github.com/supercollider/supercollider/pull/5713#issuecomment-1053720433

include/plugin_interface/FFT_UGens.h
Spacechild1, line 67: You might use `static inline` instead of just `static`, like in the functions above.

---

As a side note:

Specifying free functions in header files as static or static inline is just a hack for C. For C++, the more idiomatic (and efficient) specifier would be just inline. Note that the meaning of inline is different for C and C++! See also https://stackoverflow.com/a/1759575/6063908 and https://stackoverflow.com/a/68146072/6063908

Since this header is really a C++ header (it uses C++ classes), we could actually substitute all those static inline with inline, but that's not the topic of this PR :-)

Note that you can still take the address of an `inline` function! For an explanation, see https://stackoverflow.com/a/41737096/6063908

elgiano: Thanks for the details! That clarified a great deal some misconceptions I had about inlining!
I put it as `static inline` for consistency for now

elgiano:
I've added a note about FFT UGens in the docs as well, and squashed commits the way I would like it to be merged in the end.

Tests fail sometimes spuriously, it looks like it's quite fragile code. I've run them on my machine multiple times without a fail (Arch Linux with PipeWire, 48k/2048). I've seen them passing on CI on mac when I set `server.options.blockSize = 512` in the test class, to match the driver's block size displayed on CI logs, but then it failed on linux (driver's block size = 1024). I've tried a number of driver's block sizes on my machine (512, 1024, 2048) and it always passes. @dyfer would you have any insight about it?

dyfer:
> @dyfer would you have any insight about it?

Unfortunately debugging tests in the CI is routinely cumbersome.

Some data points to consider: GH runners have no audio hardware. On macOS the installed sound device is the "SoundFlower" virtual loopback soundcard, which I'm guessing one should be able to access at various buffer sizes (and sample rates?). On Linux we use JACK with dummy interface: `jackd --no-realtime -d dummy`. Not sure if any of these issues are directly impacting the tests. Looking at your tests, nothing seems wrong at the first glance...

One test would be to try running the whole test suite script locally to see if there are any issues when all the tests are run in sequence:
- add testsuite to the compiled paths in SC (which I'm guessing you have already)
  - install API quark if it's not installed
- create a copy of the [testing json file](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/gha_test_run_proto.json) (it will be overwritten by the test runner)
- point to that copy at the beginning of [test_runner.scd](https://github.com/supercollider/qpm/blob/topic/ignore-persistent-ui-error/qpm/scscripts/test_runner.scd#L1) file
- execute the contents of `test_runner.scd`

include/plugin_interface/FFT_UGens.h
Spacechild1: `inline` is probably one of the most confusing keywords in C++ because it has very little to do with actual code inlining :-)

dyfer:
What's the status of this PR @elgiano and @Spacechild1 ?

elgiano:
For me it's ready, waiting for review

joshpar (approved):

PR #5718, opened 2022-02-08
testsuite: Document: add missing skips

elgiano:
## Purpose and Motivation

Fixes errors occurring while testing the Document class outside the IDE. 

`Document` is not defined outside the IDE, so UnitTests needs to skip, and most of them do. This PR adds a few missing skips.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/testsuite/classlibrary/TestDocument.sc
+++ b/testsuite/classlibrary/TestDocument.sc
@@ -59,7 +59,6 @@ TestDocument : UnitTest {
 
 	test_document_getText_retrievesText {
 		var doc, str;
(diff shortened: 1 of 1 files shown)


dyfer:
Hi @elgiano 
These tests still fail (but not error out I think) when run outside of the IDE. Is this intentional?
Other tests do not fail, but I wonder if this is the desired behavior. Should skipping based on the environment be a responsibility of the test itself or of the test _runner_? I created a new issue to discuss this and related matters: https://github.com/supercollider/supercollider/issues/5719

joshpar (approved):

PR #5727, opened 2022-02-16
fix Integer#wrap docs

mxw:
## Purpose and Motivation

follow-up from #5725 

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Integer.schelp
+++ b/HelpSource/Classes/Integer.schelp
@@ -221,11 +221,11 @@ copymethod:: Float -fold
 
 method:: wrap
 
-Wrap this around [lo, hi] such that it falls in range. Equivalent to (this % (hi - lo)) + lo.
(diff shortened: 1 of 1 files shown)


jamshark70:
I think the way you've documented it is fine -- I'd approve.

It's a little uncomfortable, but if it were lo <= x < hi as Float is, then users would complain "but it isn't outputting my upper bound" (in fact, we had exactly that question when someone was doing Float:wrap with a series of whole-valued floats) -- and Float really *must* be lo <= x < hi (because, if it's lo <= x <= hi, then you need to know what is the smallest real number > hi which is a cute mathematical joke) -- there's no good solution. I think JMc's approach here is about as close to "least surprise" as you can get.

jamshark70 (approved):

telephon:
Thank you!

mxw:
@jamshark70 i think i expected a range [like, this) even for Integer by analogy with Array#wrapAt, because i think of array index ranges as being [0, size). anyway, i guess you can't please everyone :)

i do hope Wrap matches the behavior of Float, though.

Issue #3524, opened 2018-02-18
optimize .linlin, .explin, .lincurve, .curvelin


snappizz:
`.linlin`, `.explin`, `.lincurve`, `.curvelin` should be implemented as native UGens for efficiency.

`.linexp` is currently implemented as a native ugen `LinExp`, but `LinLin` is a pseudo-ugen, and `ExpLin`/`CurveLin`/`LinCurve` don't exist.

thanks to @LFSaw for bringing this one up on slack.

the efficiency hits of the pseudo-ugen versions are quite noticeable for embedded systems.

LFSaw:
oh, the thanks should go to @sensestage since it was her bringing it to my attention :)

vivid-synth:
Would be great for non-sclang client devs too :smile:

snappizz:
seems like nobody else is interested in this, so i'll reopen this when i get around to it.

Issue #3371, opened 2017-12-31
PureUGen usage is inconsistent


jamshark70:
```
SynthDef(\x, { SinOsc.ar(440); Out.ar(0, Pulse.ar(440)) }).children
-> [ a Pulse, an Out ]
```

^^ As one would expect, the PureUGen SinOsc is not connected to any output, and it's optimized away.

```
SynthDef(\x, { Pulse.ar(440); Out.ar(0, SinOsc.ar(440)) }).children
-> [ a Pulse, a SinOsc, an Out ]
```

^^ But, one would expect Pulse to be a PureUGen... but it isn't, and the optimization does not take place. I can't think of a good reason why. "A Pure UGen is a UGen, which does not access any shared resources like busses, buffers or random number generators." Pulse is just math, right? Same for Saw.

I did a sweep through UGens that are not subclasses of PureUGen and found over 170 that may be candidates to make "pure." Math operators, basic signal processors (XFade2 is not pure...?), info UGens.

**Point of discussion:** I'm taking a slightly more expansive definition of "pure," namely that it's a UGen which does not have any side effects -- that is, it's only importance is its output signal. RecordBuf is obviously not pure, because it modifies a buffer. PlayBuf, by the definition in the help file, accesses a buffer... but it only reads from the buffer. It's side-effect-free. In `FreeSelf.kr(Done.kr(aUGen))`, `Done` only reports the state of another UGen, but it doesn't mess with the synth's running state: pure. `FreeSelf` cannot be pure.

Side effects include:

- Writing to buffers, buses, random seeds or disk files. (I also think `DiskIn` and `VDiskIn` have a side effect: advancing through the disk file.) Note that this inherently makes all PV units non-pure, because you cannot rely on them not changing the buffer.
- Freeing or pausing a synth (e.g., EnvGen, Line etc. can't be pure).
- Sending OSC.
- Posting (Poll, CheckBadValues).

Some questions:

- Is it a side effect to *use* a random number generator? It does change the RNG's state for other synths... but, the point of an RNG is that you shouldn't be relying on deterministic output. I think WhiteNoise may reasonably be considered pure.
  - So I've provisionally suggested that chaos generators should be pure, but I haven't checked their implementation.
- `Demand`, `Duty`, `TDuty`... I think these should not be pure, in case they share any demand-rate unit instances with other Demand readers. But a case could be made that these units are for output.
  - What about `Dseq` etc.? They are internally stateful, but so is SinOsc, and we only care about the output.

[...]

PR #5730, opened 2022-02-20
Replace LinLin's pseudo-UGen implementation with a C++ UGen

tdug:
## Purpose and Motivation

I saw [this](https://github.com/supercollider/supercollider/issues/3524) by @nhthn, and also want to practice writing UGen functions. 

Please check my math on this. I did my best to follow the example in the SimpleNumber class `.linlin`. I also haven't written a UGen before, so thorough review would be appreciated. I'm happy to write out the rest of the UGens in the issue I've linked above.

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Audio/Line.sc
+++ b/SCClassLibrary/Common/Audio/Line.sc
@@ -26,18 +26,13 @@ LinExp : PureUGen {
 	}
 }
 
-LinLin {
+LinLin : PureUGen {
+	checkInputs { ^this.checkSameRateAsFirstInput }
 	*ar { arg in=0.0, srclo = 0.0, srchi = 1.0, dstlo = 1.0, dsthi = 2.0;
-		var scale  = (dsthi - dstlo) / (srchi - srclo);
-		var offset = dstlo - (scale * srclo);
-
-		^MulAdd(in, scale, offset)
+		^this.multiNew('audio', in, srclo, srchi, dstlo, dsthi)
 	}
 	*kr { arg in=0.0, srclo = 0.0, srchi = 1.0, dstlo = 1.0, dsthi = 2.0;
-		var scale  = (dsthi - dstlo) / (srchi - srclo);
-		var offset = dstlo - (scale * srclo);
-
-		^(in * scale + offset)
+		^this.multiNew('control',  in, srclo, srchi, dstlo, dsthi)
 	}
 }
 
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -2327,8 +2327,8 @@ void LinLin_next(LinLin* unit, int inNumSamples) {
 }
 
 #ifdef NOVA_SIMD
-static inline void LinLin_next_nova_loop(float* out, const float* in, int inNumSamples, nova::vec<float> srchi,
-                                         nova::vec<float> srclo, nova::vec<float> dsthi, nova::vec<float> dstlo) {
+static inline void LinLin_next_nova_loop(float* out, const float* in, int inNumSamples, nova::vec<float> srclo,
+                                         nova::vec<float> srchi, nova::vec<float> dstlo, nova::vec<float> dsthi) {
     const int vecSize = nova::vec<float>::size;
     int unroll = inNumSamples / (2 * vecSize);
 
@@ -2352,7 +2352,7 @@ FLATTEN static void LinLin_next_nova(LinLin* unit, int inNumSamples) {
     float* out = OUT(0);
     float* in = IN(0);
 
-    LinLin_next_nova_loop(out, in, inNumSamples, unit->m_srchi, unit->m_srclo, unit->m_dsthi, unit->m_dstlo);
+    LinLin_next_nova_loop(out, in, inNumSamples, unit->m_srclo, unit->m_srchi, unit->m_dstlo, unit->m_dsthi);
 }
 
 FLATTEN static void LinLin_next_nova_kk(LinLin* unit, int inNumSamples) {
@@ -2364,7 +2364,7 @@ FLATTEN static void LinLin_next_nova_kk(LinLin* unit, int inNumSamples) {
     float dstlo = ZIN0(3);
     float dsthi = ZIN0(4);
 
-    LinLin_next_nova_loop(out, in, inNumSamples, dstlo);
+    LinLin_next_nova_loop(out, in, inNumSamples, srclo, srchi, dstlo, dsthi);
 }
 
 #endif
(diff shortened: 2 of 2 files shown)


server/plugins/LFUGens.cpp
```diff
@@ -2302,6 +2312,169 @@ void LinExp_Ctor(LinExp* unit) {
     unit->m_rrminuslo = unit->m_rsrcrange * -srclo;
     LinExp_next(unit, 1);
 }
+////////////////////////////////////////////////////////////////////////////////////////////////////////
```
tdug, line 2315: This section was shamelessly copied from `LinExp` above, with the calculations changed. I believe I worked out all of the kinks.

Can anyone give me insight into what the `m_` prefix is on some of the variables created in the constructor? I don't know what the `m_` is short for, but I think it's just the initial value of the arguments passed into the UGen.

Reviewers:
Please look out for any cases where I missed changing a `LinExp` to `LinLin`. I missed a couple in my first iteration, but I think I got them all.
I think I got rid of every instance of the variables created in `LinExp_Ctor`. Please verify this.

joshpar: m_ prefixes are members and is needed to keep style consistent across UGens, please

telephon (commented):
Thanks, this looks very good.
One remark: the title of the PR "PureUGen for LinLin" is misleading.
A `PureUGen` is a UGen "which does not access any shared resources like busses, buffers or random number generators.". So sure `LinLin` is a PureUGen, but has always been one. 

The thing is that LinLin used to be a "pseudo-UGen".
So better write something like: "Make LinLin a UGen"

SCClassLibrary/Common/Audio/Line.sc
```diff
@@ -26,18 +26,13 @@ LinExp : PureUGen {
 	}
 }
 
-LinLin {
+LinLin : PureUGen {
+	checkInputs { ^this.checkSameRateAsFirstInput }
```
telephon, line 30: This may change behaviour: so far we could write `LinLin.kr(SinOsc.ar(440), -1, 1, 3, 4)`.

telephon (commented):
Sorry, another comment. Isn't there a trade off here in terms of optimization?
When I have a scaling function `f = { |x, c| x.linlin(0, 1, c.neg, c) }`, and I call it with 0 as an argument, like this `f.(SinOsc.kr, 0)` (this may happen in some automatic SynthDef building process. e.g.), then the old implementation would eliminate the whole calculation, but the new one wouldn't. Also `MulAdd` has a number of optimizations built in that may help to find the right combination of operations. Once all this is cast into a UGen, it can't be optimized for (or at least isn't in the new implementation).

jamshark70:
> Isn't there a trade off here in terms of optimization?

It would be worth benchmarking this, say, by creating a SynthDef with a couple hundred different `linlin`s, and seeing how much CPU is used for them both.

I'm not competent to review the C++, but I did have a go at writing a UGen once to speed up a specific math formula. *If* the UGen is vectorized, it's likely to be much faster; if not, there would be no advantage. The advantage here is likely to be smaller than in my case, because MulAdd is already well optimized.

One bad thing about the pseudo-UGen version is something like this:

```
(
SynthDef(\test, { |lo = 20, hi = 100|
	var lfos = LFDNoise3.kr(Array.fill(20, { Rand(0.1, 0.5) }));
	var rangeMapped = lfos.linlin(-1, 1, lo, hi);
	Out.kr(0, rangeMapped)
}).dumpUGens
)

[ 0_Control, control, nil ]
[ 1_-, control, [ 0_Control[1], 0_Control[0] ] ]
[ 2_/, control, [ 1_-, 2 ] ]
[ 3_-, control, [ 0_Control[1], 0_Control[0] ] ]
[ 4_/, control, [ 3_-, 2 ] ]
[ 5_-, control, [ 0_Control[1], 0_Control[0] ] ]
[ 6_/, control, [ 5_-, 2 ] ]
...
```

We get 20x the `-` and `/` operators. They will all be the same! The SynthDef optimizer is not smart enough to recognize duplicate math-op UGens, so in this case, the code looks like there's nothing wasteful going on, but in fact there is. A C++ UGen would speed this up a lot, I think, because the coefficient calculation would be a lot faster (and could be recalculated only if the input changed, as many UGens already do).

> the old implementation would eliminate the whole calculation, but the new one wouldn't.

A good point, but not hard to fix.

```
LinLin {
	*ar { arg in=0.0, srclo = 0.0, srchi = 1.0, dstlo = 1.0, dsthi = 2.0;
		if(dstlo == dsthi) { ^dstlo };
		^this.multiNew('audio', in, srclo, srchi, dstlo, dsthi)
	}
}
```

timblechmann:
> > Isn't there a trade off here in terms of optimization?
> 
> It would be worth benchmarking this, say, by creating a SynthDef with a couple hundred different `linlin`s, and seeing how much CPU is used for them both.
> 
> I'm not competent to review the C++, but I did have a go at writing a UGen once to speed up a specific math formula. _If_ the UGen is vectorized, it's likely to be much faster; if not, there would be no advantage. The advantage here is likely to be smaller than in my case, because MulAdd is already well optimized.

a few points:
* the synthdef implementation will allow arbitrary mixing of rates for all 4 range parameters
* the synthdef implementation will perform constant/rate folding. e.g. it will try to replace the division (slow) by multiplications by scalars (or constants). if `srchi` and `srclo` are control-rate, it will compute `srchi - srclo` at control-rate, not at audio-rate. if `srchi - srclo` is constant-rate `1`, it will completely eliminate the division
* the c++ implementation will be more efficient if it uses less ugens, which could be the case when all signals are audio-rate
* the c++ implementation will result in faster synthdef building, because there are less opportunities to rewrite the ugen graph.

as a rule of thumb: if a c++ ugen only performs arithmetic operations, it is hard to beat the synthdef due to the ugen graph optimisation passes and will likely be slower in most use cases. `LinLin` is a prime example of a class which should be implemented as pseudougen for efficiency reasons.

jamshark70:
While @timblechmann 's points are generally well taken, .linlin is poorly optimized for parallel chains with different inputs but the same bounds. The example I posted has 20 `-` ops and 20 `/` ops, where one of each would do. I believe this could be fixed by 1/ properly marking all deterministic, side-effect-free UGens as PureUGen (#3371, which I started but dropped for some reason) and 2/ caching PureUGens in a MultiLevelIdentityDictionary, so that later attempts to create a matching instance would reuse the instance from cache.

> it will try to replace the division (slow) by multiplications by scalars (or constants)

Actually, it doesn't do this. (The logic is sound, but not implemented.)

```
[ 0_Control, control, nil ]
[ 1_-, control, [ 0_Control[1], 0_Control[0] ] ]
[ 2_/, control, [ 1_-, 2 ] ]
```

If it really did as you were saying, the third unit would be `[ 2_*, control, [ 1_-, 0.5 ] ]`.

capital-G:
@tdug do you still want to get this merged? Would be a good addition IMO :)

dyfer:
ello everyone, is the consensus here that we should close this, leaving the linlin implementation as is?

jamshark70:
I tend to agree to let #6405 optimize the pseudo-UGen.

capital-G:
Lets close this in this case - re-open if necessary.

PR #5743, opened 2022-03-11
class library: guarantee that SetBuf gets an array

### LINKED ISSUE #5722: Minor bug SetBuf
Although `SetBuf` is not a commonly used ugen (I don't know its use cases), its constructor should call `asArray` on `values` parameter otherwise it will set the wrong size for the command (0 instead of 1) if a single number is provided. Currently `LocalBuf` seem to be the interface for this ugen and its `set` method does `SetBuf(this, values.asArray, offset);`. It could be better to make `values = values.asArray` inside `SetBuf` constructor instead so it will work for any other case. `SetBuf` could be said to be expecting an array but a single number will not throw any error and create the command anyway.

## Example

```
(
d = SynthDef('test', {
	var buf = LocalBuf(64, 1);
	buf.set(7); // [ 1_LocalBuf, 0, 1, 7 ] (buffer, offset, size, values*)
	SetBuf(buf, 7); // [ 1_LocalBuf, 0, 0, 7 ]
});
d.dumpUGens
)
```

## Environment

* SuperCollider version: all
* Operating system: all

telephon:
this fixes #5722

## Purpose and Motivation

See discussion on #5722.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Audio/BufIO.sc
+++ b/SCClassLibrary/Common/Audio/BufIO.sc
@@ -190,6 +190,7 @@ MaxLocalBufs : UGen {
 
 SetBuf : WidthFirstUGen {
 	*new { arg buf, values, offset = 0;
+		values = values.asArray;
 		^this.multiNewList(['scalar', buf, offset, values.size] ++ values)
 	}
 }


dyfer (approved):
Looks good!

PR #5744, opened 2022-03-11
class library: loadRelative – improve behaviour of error

telephon:
When a file is unsaved or a path can't be found for other reasons, `loadRelative` and `resolveRelative` warn the user. So far, they have thrown an error, but it seems better to warn with more details and bail out instead.

## Types of changes

- New feature
- Breaking change (possibly)

## To-do list

- [x] Code is tested
- [x] All tests are passing


dyfer (approved):
Looks good, thanks!

PR #5747, opened 2022-03-19
Refactor SC_ComPort, add support for receiving raw UDP socket communication

scztt:
## Purpose and Motivation

Implementing a SuperCollider language server (https://github.com/scztt/LanguageServer.quark) requires that sclang be able to send and receive network communication that is not in the form of OSC messages. The primary motivation of this change is to introduce the `recvRawMessage`, which can be used to receive non-OSC content arriving at a TCP port. However, the existing socket communication code is fragile and inflexible - I've tried to clean up and generalize where possible to make it easier to introduce new network processing capabilities in the future.

Once this pull request is merged, I will use it as the basis for adding a JSON-RPC communication channel, which is required for a final proper LSP implementation.

Pure refactoring changes with no change to application behavior are marked as "No functional change".

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


scztt:
There are no unit tests for any of the socket code at the moment. There should be (in that, there should be unit tests for EVERYTHING 🙃), but implementing unit test for this code area feels like a very large scope expansion of the PR. If there's consensus that this should be fully unit tested before merging, I'm open to ideas about how to build tests in a lightweight way.

Note that lots of the socket code is essentially duplicated between the client and the server. I didn't attempt to solve this problem in this PR, but fixing this is probably a good next step after this PR is merged.

scztt:
I'm a little confused by the lint failures - the suggested diffs shows in the lint GHA results differ from the way clang-format formats these files. Could this be related to a clang-format version difference?

telephon:
If the changes do not touch any existing part of the system, unit tests could be a separate issue.

davidgranstrom:
@scztt I'm trying out this branch but I'm having some trouble connecting to the TCP port. 

I'm opening the port like this:

```
(
thisProcess.openTCPPort(9999);
thisProcess.openPorts().debug('openPorts'); // openPorts: Set[ 9999, 57120 ]
thisProcess.recvRawfunc = {|time, replyAddr, msg|
    msg.debug('msg');
};
)
```

And then trying to connect to it using `netcat`: 
```
❯ nc -v localhost 9999
nc: connectx to localhost port 9999 (tcp) failed: Connection refused
nc: connectx to localhost port 9999 (tcp) failed: Connection refused
```

Please let me know if there's something missing/wrong from the code above, or any pointers into debugging this!

I'm using macOS 12.3 with the M1 Pro chip.

scztt:
> @scztt I'm trying out this branch but I'm having some trouble connecting to the TCP port.

This was a mistake in the naming - this does NOT in fact open a TCP port but just a UDP port with a non-OSC handler. Fixed in the PR. This example now works for me:

```
(
thisProcess.openUDPPort(9999, \raw);
thisProcess.openPorts().debug('openPorts'); // openPorts: Set[ 9999, 57120 ]
thisProcess.recvRawfunc = {|time, replyAddr, msg|
    msg.debug('msg');
};
)
```

```
nc -u -v localhost 9999                
```

scztt:
Ping - any updates on merging this?

ranjithshegde:
> Ping - any updates on merging this?

Been using this branch for a while now. @scztt would you please rebase on master? Otherwise it breaks compatibility with latest version of libsndfile

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -111,127 +251,91 @@ void SC_UdpInPort::handleReceivedUDP(const boost::system::error_code& error, std
...
         return;
     }
 
-    OSC_Packet* packet = (OSC_Packet*)malloc(sizeof(OSC_Packet));
-
-    packet->mReplyAddr.mProtocol = kUDP;
-    packet->mReplyAddr.mAddress = remoteEndpoint.address();
-    packet->mReplyAddr.mPort = remoteEndpoint.port();
-    packet->mReplyAddr.mSocket = udpSocket.native_handle();
+    std::unique_ptr<char[]> data(new char[bytesTransferred]);
```
jrsurge, line 259:
```suggestion
    auto data = std::make_unique<char[]>(new char[bytesTransferred]);
```

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -111,127 +251,91 @@ void SC_UdpInPort::handleReceivedUDP(const boost::system::error_code& error, std
...
     // FIXME: handle max connections
     // FIXME: backlog???
 
     startAccept();
 }
 
-void SC_TcpInPort::startAccept() {
-    SC_TcpConnection::pointer newConnection(new SC_TcpConnection(ioService, this));
+void TCP::startAccept() {
+    Detail::TCPConnection::pointer newConnection(new Detail::TCPConnection(ioService, mPortNum, mHandlerType));
```
jrsurge, line 284: I think `std::make_shared` makes this more clear - and given that we aren't modifying the `shared_ptr` itself, from what I can see it can maybe even be const?

```suggestion
    const auto newConnection = std::make_shared<Detail::TCPConnection>(ioService, mPortNum, mHandlerType);
```

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -241,49 +345,38 @@ void SC_TcpClientPort::handleLengthReceived(const boost::system::error_code& err
         return;
 
     // msglen is in network byte order
-    OSCMsgLength = sc_ntohl(OSCMsgLength);
-    data = (char*)malloc(OSCMsgLength);
+    mOSCMsgLength = sc_ntohl(mOSCMsgLength);
+    mData.reset(new char[mOSCMsgLength]);
```
jrsurge, line 349:
```suggestion
    mData = std::make_unique<char[]>(new char[mOSCMsgLength]);
```

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -65,16 +66,136 @@ void stopAsioThread() {
...
+                   });
+}
+
+void TCPConnection::handleMsgReceived(const boost::system::error_code& error, size_t bytes_transferred) {
+    if (error) {
+        mData.reset();
+        return;
+    }
+
+    assert(bytes_transferred == mOSCMsgLength);
```
jrsurge, line 176: I always like to add a comment to an assert to explain:
- why I might have hit this
- what am I supposed to do if this gets hit

from context here, it's hard for me to know where SC went wrong

lang/LangPrimSource/SC_ComPort.h
```diff
@@ -20,103 +20,122 @@
...
+    void initHandler(HandlerType type);
 
-    void handleLengthReceived(const boost::system::error_code& error, size_t bytes_transferred);
+    void handleReceivedUDP(const boost::system::error_code& error, std::size_t bytes_transferred);
+    void startReceiveUDP();
 
-    void handleMsgReceived(const boost::system::error_code& error, size_t bytes_transferred);
+    int mPortNum;
+    HandleDataFunc mHandleFunc;
+    boost::array<char, kTextBufSize> mRecvBuffer;
```
jrsurge, line 85: is `boost:array` the same as `std::array`?

lang/LangPrimSource/SC_ComPort.h
```diff
@@ -20,103 +20,122 @@
...
-
     void handleMsgReceived(const boost::system::error_code& error, size_t bytes_transferred);
+    void initHandler(HandlerType);
 
-    boost::asio::ip::tcp::socket socket;
-    boost::asio::ip::tcp::endpoint endpoint;
-
+    HandleDataFunc mHandleFunc;
+    int32 mOSCMsgLength;
+    std::unique_ptr<char[]> mData;
```
jrsurge, line 134: why does TCP use a `std::unique_ptr<char[]>` but UDP use a `boost::array`?

jrsurge (changes requested):
Sorry for the delay in reviewing this!

Huge thanks for all the work, it looks great.
I had a few thoughts and questions about the code, but I've tried to put github suggestions where applicable.

Thanks!

SCClassLibrary/DefaultLibrary/Main.sc
```diff
@@ -83,24 +84,56 @@ Main : Process {
 		OSCresponder.respond(time, replyAddr, msg);
 	}
 
+	recvRawMessage { arg time, replyAddr, recvPort, msg;
```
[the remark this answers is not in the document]
scztt: Resolved - I switched both of the recv methods for consistency, but didn't do a full pass in this file to avoid mixing in unrelated changes.

SCClassLibrary/DefaultLibrary/Main.sc
```diff
@@ -1,6 +1,7 @@
 Main : Process {
 	var <platform, argv;
 	var recvOSCfunc, prRecvOSCFunc;
+	var >recvRawfunc, >prRecvRawFunc;
```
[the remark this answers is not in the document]
scztt: Shouldn't be - resolved.

lang/LangPrimSource/OSCData.cpp
```diff
@@ -757,41 +776,61 @@ void init_OSC(int port) {
...
+    int port;
+    int err = slotIntVal(b, &port);
+    if (err)
+        return err;
+
+    std::unique_ptr<InPort::UDPCustom> newTCPPort;
+
+    try {
+        SetTrue(a);
+        newTCPPort.reset(new InPort::UDPCustom(port, HandlerType::Raw));
```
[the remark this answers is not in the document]
scztt: Good catch, fixed.

SCClassLibrary/DefaultLibrary/Main.sc
```diff
@@ -83,24 +84,56 @@ Main : Process {
 		OSCresponder.respond(time, replyAddr, msg);
 	}
 
+	recvRawMessage { arg time, replyAddr, recvPort, msg;
+		// this method is called when an raw message is received.
```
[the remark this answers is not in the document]
scztt: Only following the comment in recvOSCmessage - I agree, these basically just repeat the method name :) - should I remove both?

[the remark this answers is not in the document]
scztt: TBH how these two methods fit into the larger environment is still unclear - if you didn't know SC internals, there's no way to know that these are called from C++ when messages are received (something you'd never know by searching the SCClassLibrary source). Maybe instead I add a better explanation here for both of these?

lang/LangPrimSource/SC_ComPort.cpp
scztt: This is a reasonable change, resolved.

lang/LangPrimSource/SC_ComPort.h
scztt: This should be `std::array`. Resolved.

lang/LangPrimSource/SC_ComPort.h
scztt: I wanted to keep existing code as close as possible to the original behavior / minimize possibilities for subtle change  since this has very little test coverage. This can definitely be changed to std::array, but probably it should be done in a separate change and tested as such. Maybe I'm just being extra paranoid. :)

scztt:
@jrsurge Thanks for the extremely thorough review!

dyfer:
Thanks @scztt - while we are still in the process of reviewing this, could you please make sure the title reflects the functionality of the PR? Shouldn't it be "UDP socket" as opposed to "USP socket"?

lang/LangPrimSource/SC_ComPort.h
```diff
@@ -20,103 +20,122 @@
...
 
 const int kTextBufSize = 65536;
 
-class SC_UdpInPort {
-    int mPortNum;
-    boost::array<char, kTextBufSize> recvBuffer;
+enum class HandlerType { OSC, Raw };
 
-    boost::asio::ip::udp::endpoint remoteEndpoint;
+using HandleDataFunc = std::function<void(std::unique_ptr<char[]>, size_t)>;
```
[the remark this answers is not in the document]
scztt: In this case, we are actually handing off ownership of this data, so std::span doesn't capture it? This could be re-worked as `std::string`, but probably that's more of a refactoring change after this PR.

lang/LangPrimSource/OSCData.cpp
```diff
@@ -288,7 +288,7 @@ static int netAddrSend(PyrObject* netAddrObj, int msglen, char* bufptr, bool sen
     using namespace boost::asio;
 
     if (IsPtr(netAddrObj->slots + ivxNetAddr_Socket)) {
-        SC_TcpClientPort* comPort = (SC_TcpClientPort*)slotRawPtr(netAddrObj->slots + ivxNetAddr_Socket);
+        OutPort::TCP* comPort = (OutPort::TCP*)slotRawPtr(netAddrObj->slots + ivxNetAddr_Socket);
```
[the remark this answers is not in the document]
scztt: The general guidance I've seen is that static_cast is more appropriate for a void* <> non-void* cast. Fixed.

lang/LangPrimSource/OSCData.cpp
```diff
@@ -713,35 +714,53 @@ void PerformOSCMessage(int inSize, char* inData, PyrObject* replyObj, int inPort
...
+                PerformOSCMessage(inPacket->mSize, inPacket->mData.get(), replyObj, inPortNum, time);
             }
         }
     }
     gLangMutex.unlock();
+}
+
+void ProcessRawMessage(size_t inSize, std::unique_ptr<char[]> inData, ReplyAddress& replyAddress, int inPortNum,
+                       double time) {
+    gLangMutex.lock();
```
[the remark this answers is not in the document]
scztt: Yeah I thought about that - but this is the pattern used ALL over the code base. It would be a lot easier to fix all of those in a single PR rather than as a piecemeal change in this unrelated PR.

lang/LangPrimSource/SC_ComPort.cpp
scztt: Sadly doesn't work. `std::make_unique` doesn't have a variant that takes an existing pointer (makes sense, make_unique is supposed to do the allocating(), and a raw `char[]` doesn't have a constructor that takes the size of the array to allocate (which would fit the make_unique pattern).
Anyway, this should all be replaced with storage other than char[], but that's a bigger change that should be separated from this.

scztt:
@jrsurge All your comments should now be resolved!

scztt:
Ping - anything left to do on this pr? As far as I can tell it's ready for merge.

jrsurge (approved):
Apologies for the delay!

Huge thanks for this - it looks good to me, thanks for responding so thoughtfully to the review, I appreciate I asked a lot of questions.

Thanks!

PR #5750, opened 2022-03-30
test: shadowing parent event with envirPut nil

### LINKED ISSUE #5749: environment: ~test = x does not behave the same as currentEnvironment.put(\test, x)
## Steps to reproduce

```supercollider

e = Environment.new;
currentEnvironment = e;

// these two lines should do the same:
~test = 56; e; // Environment[ (test -> 56) ]
e[\test] = 56; // Environment[ (test -> 56) ]

// but they don't do the same in the case of nil
~test = nil; e; // Environment[ (test -> nil) ]
e[\test] = nil; // Environment[  ]

// the reason is that envirPut won't do the same as IdentityDictionary:put 
\test.envirPut(nil); e;
e.put(\test, nil);

// IdentityDictionary:put has this extra code when the primitive bails out for nil:
put { arg key, value;
		_IdentDict_Put
		value ?? { this.removeAt(key); ^this };
		^this.primitiveFailed
		<...>
	}


// the same should be in place for Symbol:envirPut


```

```
e = Environment.new;
currentEnvironment = e;

e[\test] = nil; // Environment[  ]  // as expected
~test = nil; e; // Environment[ (test -> nil) ] // should be  Environment[  ]
```

This is annoying when one tries to get through to the parent environment by setting an overridden key to nil.

telephon:
## Purpose and Motivation

Fixes #5749

Note that this might break code, so it needs discussion.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangPrimSource/PyrListPrim.cpp
+++ b/lang/LangPrimSource/PyrListPrim.cpp
@@ -471,6 +471,9 @@ int prSymbol_envirPut(struct VMGlobals* g, int numArgsPushed) {
 
     if (!ISKINDOF(dict, class_identdict_index, class_identdict_maxsubclassindex))
         return errFailed;
+    
+    if (IsNil(b))
+        return errFailed; // will call removeAt
 
     int err = identDictPut(g, dict, a, b);
     if (err)
--- a/SCClassLibrary/Common/Core/Symbol.sc
+++ b/SCClassLibrary/Common/Core/Symbol.sc
@@ -86,7 +86,11 @@ Symbol {
 	}
 	envirPut { arg aValue;
 		_Symbol_envirPut
-		currentEnvironment.put(this, aValue);
+		if(aValue.isNil) {
+			currentEnvironment.removeAt(this)
+		} {
+			currentEnvironment.put(this, aValue)
+		};
 		^aValue
 	}
 
--- a/testsuite/classlibrary/TestEnvironment.sc
+++ b/testsuite/classlibrary/TestEnvironment.sc
@@ -1,0 +1,15 @@
+TestEnvironment : UnitTest {
+
+	test_envirPutNilRemovesEntry {
+		var envir = Environment.new(parent: (test: \defaultValue));
+		envir.use {
+
+			this.assertEquals(~test, \defaultValue, "parent environment should be in scope if not shadowed");
+			~test = \overriddenEntry;
+			this.assertEquals(~test, \overriddenEntry, "parent environment should be shadowed");
+			~test = nil;
+			this.assertEquals(~test, \defaultValue, "parent environment should be in scope again if not shadowed any more");
+		};
+	}
+
+}


lang/LangPrimSource/PyrListPrim.cpp
```diff
@@ -471,6 +471,9 @@ int prSymbol_envirPut(struct VMGlobals* g, int numArgsPushed) {
 
     if (!ISKINDOF(dict, class_identdict_index, class_identdict_maxsubclassindex))
         return errFailed;
+
```
dyfer, line 474: It seems that there's an extra space here that triggers the linting failure

telephon: extra whitespaces before the line break … (thanks!)

dyfer:
@telephon just FYI since there was no discussion on this (especially as it's potentially a breaking change), I don't think this will make it into 3.13. But if you feel we should consider it, let me know.

nhthn (changes requested):
Two recommendations here:

1. the test is pretty cryptic to me, would like to see it split into multiple tests cases possibly
2. user-facing documentation of Environment should be updated to clearly explain this behavior (including the change in behavior) as it is rather subtle

elgiano:
Hi @nhthn would you have time to check if your points have been addressed?

@telephon for the test I think I preferred symbol values to numbers, something like \parentEnvValue and \childEnvValue, like it was before, it makes it more self-explanatory IMO.

other than this it looks good to me!

elgiano:
Hi @nhthn would you have time to check if your points have been addressed?

@telephon for the test I think I preferred symbol values to numbers, something like \parentEnvValue and \childEnvValue, like it was before, it makes it more self-explanatory IMO.

other than this it looks good to me!

elgiano:
@telephon probably I'm missing something, but on a second thought: why shouldn't we remove Symbol primitives and  call `currentEnvironment.put(this, aValue);`?

```supercollider
envirGet {
  ^currentEnvironment.at(this)
}
envirPut { arg aValue;
  currentEnvironment.put(this, aValue);
}
```

telephon:
(sorry for the late reply, I missed this one).

> @telephon probably I'm missing something, but on a second thought: why shouldn't we remove Symbol primitives and call `currentEnvironment.put(this, aValue);`?

Well, the byte codes indicate that the replacement would have one more step (`PushClassVar`). I don't know if it matters.


```
{ ~test = 9 }.def.dumpByteCodes;

BYTECODES: (7)
  0   40       PushLiteral Symbol 'test'
  1   2C 09    PushInt 9
  3   B0       TailCallReturnFromFunction
  4   C2 3D    SendSpecialMsg 'envirPut'
  6   F2       BlockReturn
-> < closed FunctionDef >


{ currentEnvironment.put(\test, 9) }.def.dumpByteCodes;

BYTECODES: (9)
  0   50 01 PushClassVar
  2   40       PushLiteral Symbol 'test'
  3   2C 09    PushInt 9
  5   B0       TailCallReturnFromFunction
  6   C3 03    SendSpecialMsg 'put'
  8   F2       BlockReturn
-> < closed FunctionDef >

```

There is no consistent difference in benchmark.
```
bench { 10000.do { ~test = 9.rand } };
bench { 10000.do { currentEnvironment.put(\test, 9.rand) } };
```

elgiano:
Thanks for the dumps and benchs!

I'm not familiar enough with the interpreter or GC to evaluate the difference, it seems negligible to me but I wouldn't say it for sure. After all, this was designed to be like this almost from the start.

Removing symbol envir primitives would reduce code duplication, the two functions are almost the same, as envirPut actually refers to identDictPut internally, doing nothing more than retrieving currentEnvironment in cpp (in `currentEnvironment.put`, PushClassVar is passing the env as an argument to the primitive, while envirPut is retrieving `currentEnvironment` internally). 

Another potential advantage would be that pseudo-classes could then override envirPut:
```supercollider
q = (
	envirPut: { |self, k,v| [k, v].postln; self[k] = v },
)

q.use {|self| self.envirPut(\test, 9) }
// if we remove symbol primitives, this would call the override too
q.use { ~test = 9 }
```

What do you think?

telephon:
> What do you think?

I tend to agree. Maybe @jamshark70 could comment on the byte code question – after all it is a primitive very commonly called in situations where efficiency is most important.

elgiano:
Thanks @jamshark70.
However why should we call `currentEnvironment.put(this, aValue);` only if `_Symbol_envirPut` fails? Couldn't we call it directly in any case and get rid of `_Symbol_envirPut`?

My reasoning is:
- They do the same thing: `_Symbol_envirPut` is calling `identDictPut` internally, and so does `_IdentDict_Put`.
- if `Symbol_envirPut` fails, it calls `currentEnvironment.put` , where `_IdentDict_Put` would also fail, and I dont think this call is necessary. 
- The only difference between the two methods seems to be that `_IdentDict_Put` needs an extra `PushClassVar` op, in order to pass `currentEnvironment`, while `Symbol_envirPut` gets it directly in cpp without needing it as an argument. Do you think this difference is worth the code duplication?

telephon:
I think it is fine to discuss such questions with a relevant case in mind, so good here.

jamshark70:
OK, so first @telephon -- yesterday I had read only the conversation, not the code -- now I see that the PR implements the same as my comment, so my comment was a bit redundant.

Second @elgiano --

> My opinion is that if we remove Symbol_envirPut, we would actually remove an extra flow of dispatch...

OK, I needed more time to ponder it.

Actually, changing the `~xyz = 111` bytecode might improve performance in my typical use case, because in my own code, a majority of the time, `~xyz` is in a Proto object, not an Environment directly. So, when I assign to an environment variable, I'm going through the fallback *a lot*. Calling `.put` directly on the Proto would skip both the `envirPut` dispatch and the fallback.

What about `envirGet`, then? Could switch that to call `.at` directly (and avoid another fallback that I'm also going through *a lot*).

I guess it's really a matter of testing and risk assessment. The current PR applies a known execution flow to a situation where it currently doesn't apply, but should -- low risk. The alternate proposal is to change bytecode emission in the compiler, which is higher risk. Thorough testing would mitigate the risk. Provided that the compiler revision is done properly, I suppose the risk is pretty low (but the risk of doing the compiler revision incorrectly is not zero).

PR #5761, opened 2022-04-19
libsndfile: use a macro instead of redefining the struct

### LINKED ISSUE #5756: libsndfile version 1.1.0 breaks building from source
## Environment

* SuperCollider version: b92a9073bce1c7871553e089bbf1bf58381fcaf4
* Operating system:Mac OS X Monterey 12.2.1
* Other details (Qt version, audio driver, etc.): libsndfile v1.1.0 from Homebrew

## Steps to reproduce
```
cmake --build . --target install --config RelWithDebInfo 
```

```
In file included from /Users/stratosbichakis/workspaces/learn/SuperCollider/server/scsynth/SC_CoreAudio.cpp:23:
In file included from /Users/stratosbichakis/workspaces/learn/SuperCollider/server/scsynth/SC_SequencedCommand.h:35:
In file included from /Users/stratosbichakis/workspaces/learn/SuperCollider/common/SC_SndFileHelpers.hpp:35:
/usr/local/opt/libsndfile/include/sndfile.h:356:31: error: typedef redefinition with different types ('struct sf_private_tag' vs 'struct SNDFILE_tag')
typedef struct sf_private_tag   SNDFILE ;
                                ^
In file included from /Users/stratosbichakis/workspaces/learn/SuperCollider/server/scsynth/SC_CoreAudio.cpp:23:
In file included from /Users/stratosbichakis/workspaces/learn/SuperCollider/server/scsynth/SC_SequencedCommand.h:32:
In file included from /Users/stratosbichakis/workspaces/learn/SuperCollider/include/plugin_interface/SC_World.h:31:
/Users/stratosbichakis/workspaces/learn/SuperCollider/include/plugin_interface/SC_SndBuf.h:25:28: note: previous definition is here
typedef struct SNDFILE_tag SNDFILE;
```

dyfer:
## Purpose and Motivation

Fixes compatibility with `libsndfile` >=1.1.0
Fixes #5756 

See https://github.com/supercollider/supercollider/issues/5756#issuecomment-1094378766 for previous discussion.

Thanks @Spacechild1 for providing this fix.

I also had to add additional Windows includes in `SC_HiddenWorld.h`, similarly to `SC_SndFileHelpers.hpp`. 

Please review carefully, as I have limited experience with this aspect of the project.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/plugins/DiskIO_UGens.cpp
+++ b/server/plugins/DiskIO_UGens.cpp
@@ -111,7 +111,7 @@ void DiskIOMsg::Perform() {
     sf_count_t count;
     switch (mCommand) {
     case kDiskCmd_Read:
-        count = buf->sndfile ? sf_readf_float(buf->sndfile, buf->data + mPos * buf->channels, mFrames) : 0;
+        count = buf->sndfile ? sf_readf_float(GETSNDFILE(buf), buf->data + mPos * buf->channels, mFrames) : 0;
         if (count < mFrames) {
             memset(buf->data + (mPos + count) * buf->channels, 0, (mFrames - count) * buf->channels * sizeof(float));
             World_GetBuf(mWorld, mBufNum)->mask = mPos + count;
@@ -126,17 +126,17 @@ void DiskIOMsg::Perform() {
             memset(buf->data + mPos * buf->channels, 0, mFrames * buf->channels * sizeof(float));
             goto leave;
         }
-        count = sf_readf_float(buf->sndfile, buf->data + mPos * buf->channels, mFrames);
+        count = sf_readf_float(GETSNDFILE(buf), buf->data + mPos * buf->channels, mFrames);
         while (mFrames -= count) {
-            sf_seek(buf->sndfile, 0, SEEK_SET);
-            count = sf_readf_float(buf->sndfile, buf->data + (mPos + count) * buf->channels, mFrames);
+            sf_seek(GETSNDFILE(buf), 0, SEEK_SET);
+            count = sf_readf_float(GETSNDFILE(buf), buf->data + (mPos + count) * buf->channels, mFrames);
         }
         break;
     case kDiskCmd_Write:
         // printf("kDiskCmd_Write %d %p\n", mBufNum, buf->sndfile);
         if (!buf->sndfile)
             goto leave;
-        count = sf_writef_float(buf->sndfile, buf->data + mPos * buf->channels, mFrames);
+        count = sf_writef_float(GETSNDFILE(buf), buf->data + mPos * buf->channels, mFrames);
         break;
     }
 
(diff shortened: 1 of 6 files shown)


server/plugins/DiskIO_UGens.cpp
```diff
@@ -126,17 +126,17 @@ void DiskIOMsg::Perform() {
             memset(buf->data + mPos * buf->channels, 0, mFrames * buf->channels * sizeof(float));
             goto leave;
         }
-        count = sf_readf_float(buf->sndfile, buf->data + mPos * buf->channels, mFrames);
+        count = sf_readf_float(GetSndFile(buf), buf->data + mPos * buf->channels, mFrames);
```
joshpar, line 129: this may not matter, but I wonder if it is cleaner to store the result of GetSndFile(but) into a variable and pass that around, rather than calling GetSndFile over and over again?

joshpar (changes requested):
@dyfer - I think this mostly is fine, but the readability of the change feels a little off to me. that being said, if you feel ok with it, I can go with it as well.
I think in other places, macros use all upper case? Maybe that would help this from looking like a method / class call? GETSNDFILE instead? 
Let me know what you think. This is a stylistic question more than a functional one.

server/plugins/DiskIO_UGens.cpp
Spacechild1: IMO it's fine because `GetSndFile` is just a getter macro and not a real function call. I find it quite readable. After all, it's just 3 characters longer than the original code.

Spacechild1:
> I think in other places, macros use all upper case? 

It depends. For example, the macros in `SC_InterfaceTable.h` are camel case. IMO, `GETSNDFILE` would be fine as well. Dunno.

server/scsynth/SC_HiddenWorld.h
```diff
@@ -37,6 +37,15 @@
...
 
+#ifndef NO_LIBSNDFILE
+#    ifdef _WIN32
+// This needs to be defined on every sndfile.h include, see SC_SndFileHelpers.hpp
+#        include <windows.h>
+#        define ENABLE_SNDFILE_WINDOWS_PROTOTYPES 1
+#    endif
+#    include <sndfile.h>
+#endif
+
```
Spacechild1, line 48: You could also just include `SC_SndFileHelpers.hpp` instead.

dyfer:
> @dyfer - I think this mostly is fine, but the readability of the change feels a little off to me. that being said, if you feel ok with it, I can go with it as well. I think in other places, macros use all upper case? Maybe that would help this from looking like a method / class call? GETSNDFILE instead? Let me know what you think. This is a stylistic question more than a functional one.

I also wasn't convinced by the readability. I'll change it to all caps.

server/scsynth/SC_World.cpp
```diff
@@ -968,9 +968,9 @@ void World_Cleanup(World* world, bool unload_plugins) {
 
 #ifndef NO_LIBSNDFILE
         if (nrtbuf->sndfile)
-            sf_close(nrtbuf->sndfile);
-        if (rtbuf->sndfile && rtbuf->sndfile != nrtbuf->sndfile)
-            sf_close(rtbuf->sndfile);
+            sf_close(GETSNDFILE(nrtbuf));
+        if (rtbuf->sndfile && rtbuf->sndfile != GETSNDFILE(nrtbuf))
```
dyfer, line 972: @Spacechild1 is this check correct now? I wasn't sure here

Spacechild1: rtbuf->sndfile && rtbuf->sndfile != nrtbuf->sndfile is perfectly ok.

dyfer: Thanks! Updated.

server/plugins/DiskIO_UGens.cpp
dyfer: I've changed it to `GETSNDFILE`

joshpar (approved):

PR #5772, opened 2022-05-06
Help: Linux: Clarify behavior of MIDIOut objects with uid

jamshark70:
## Purpose and Motivation

Linux MIDIOut users frequently get confused when `MIDIOut.newByName` pseudo-connections don't appear in patchbay GUIs.

This documentation change explains that the mechanism bypasses visible connections altogether.

Previously, the documentation (written by myself) claimed that "one-to-one connections" were made. That's misleading -- targeting a device by uid is a completely separate mechanism.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/MIDIOut.schelp
+++ b/HelpSource/Classes/MIDIOut.schelp
@@ -141,9 +141,15 @@ m.connect(1);  // connect to MIDIClient.destinations[1]
 
 WARNING:: The teletype::port:: argument to link::Classes/MIDIOut#*new:: has an entirely different meaning in Linux, compared to macOS and Windows. If user code calls this method emphasis::and:: cross-platform compatibility is needed, it is the user's responsibility to handle Linux separately. User code can check the platform using code::thisProcess.platform.name:: (which returns one of code::\osx::, code::\linux:: or code::\windows::). Or, for compatibility, use link::#*newByName:: instead.::
 
-link::Classes/MIDIOut#*new:: optionally takes a second argument, for the uid of the MIDI destination device. This establishes a direct connection to the target, which is not visible in Qjackctl. The teletype::port:: argument is irrelevant in this case; the MIDIOut instance is always one-to-one with the target.
+link::Classes/MIDIOut#*new:: optionally takes a second argument, for the uid of the MIDI destination device. If the uid is non-zero, then: 1/ all connections in the ALSA MIDI patchbay are ignored, and 2/ individual messages will be sent directly to the specified device.
 
-link::Classes/MIDIOut#*newByName:: should create one-to-one MIDIOut connections on all platforms.
+link::Classes/MIDIOut#*newByName:: locates a device by name and populates the uid in the new MIDIOut object. Thus uid should be non-zero for all MIDIOut objects created using this method.
+
+note::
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/MIDIOut.schelp
```diff
@@ -141,9 +141,15 @@ m.connect(1);  // connect to MIDIClient.destinations[1]
...
 
-link::Classes/MIDIOut#*new:: optionally takes a second argument, for the uid of the MIDI destination device. This establishes a direct connection to the target, which is not visible in Qjackctl. The teletype::port:: argument is irrelevant in this case; the MIDIOut instance is always one-to-one with the target.
+link::Classes/MIDIOut#*new:: optionally takes a second argument, for the uid of the MIDI destination device. If the uid is non-zero, then: 1/ all connections in the ALSA MIDI patchbay are ignored, and 2/ individual messages will be sent directly to the specified device.
 
-link::Classes/MIDIOut#*newByName:: should create one-to-one MIDIOut connections on all platforms.
+link::Classes/MIDIOut#*newByName:: locates a device by name and populates the uid in the new MIDIOut object. Thus uid should be non-zero for all MIDIOut objects created using this method.
```
dyfer, line 146: >  "Thus uid should be non-zero for all MIDIOut objects (...)"

On first reading I had an impression that "should" referred to requiring the user to provide these. What do you think about changing it to something along the lines of "Thus uid will be set to non-zero (...)"? Or maybe even refer to it as "internally used uid"? (not sure about that last one)

HelpSource/Classes/MIDIOut.schelp
```diff
@@ -141,9 +141,15 @@ m.connect(1);  // connect to MIDIClient.destinations[1]
...
+
+note::
+A non-zero uid (whether specified in teletype::MIDIOut.new::, found by teletype::MIDIOut.newByName::, or set manually later) does not create or use any ALSA MIDI connections. It is not a "connection" at all; rather, it is a parameter that causes outgoing MIDI messages to bypass ALSA MIDI connections and go directly to one specific device. Because it is not a connection, there is nothing to show in MIDI patchbay interfaces; users should not expect to see connections for these MIDIOut objects.
```
dyfer, line 149: Is there a reason to use `teletype` instead of `code` for the inline snippets here?

dyfer:
Thanks for this! It is a very useful clarification, especially for casual Linux users.

HelpSource/Classes/MIDIOut.schelp
jamshark70: I guess I was thinking -- if it's a full, executable code snippet such as `MIDIOut.new(0).connect(1)`, then code colorizing makes sense. If it's just identifying a method, then it doesn't need to be "code" but monospaced type is still useful.

I can change it but I'm curious, why does `teletype:: ::` exist if we're always supposed to use `code:: ::` instead? :wink:

HelpSource/Classes/MIDIOut.schelp
jamshark70: Fixed.

HelpSource/Classes/MIDIOut.schelp
dyfer: I haven't been part of the conversations about the origins of teletype. I simply use `code:: ::` for all inline code in the help. I wouldn't say there's any discouragement for using `teletype` - I personally would use it only for non-sc code (very rarely).

I don't think it's worth it to go in and change other uses of `teletype:: ::` :) thanks!

dyfer (approved):
Thanks!

PR #5783, opened 2022-05-20
Update Linux build matrix

dyfer:
## Purpose and Motivation

Our build matrix had no builds on Ubuntu 22.04 and only one on 20.04. This update brings building using newer compilers and newer OSs.

~~Please note, Linux builds with `clang-11`, `clang-12` and `clang-13` fail to build (see https://github.com/dyfer/supercollider/runs/6531974521?check_suite_focus=true) and are thus disabled. I don't know whether the issue has to do with the OS configuration/missing packages, or with our build system... However, clang builds seem to be fine on macOS (I think locally I've been building using all or most of these clang versions throughout the last few years) so I'm hoping this is not an issue with our codebase.~~

~~EDIT: After discussion with @jrsurge he decided to look into the reason for these failures. Let's wait for a proposed fix before merging this PR.~~

EDIT: it turns out that issue with clang builds was due to missing libraries; we needed to install proper versions of the `libc++` libs in CI and this PR now provides that.

I've also updated the Qt5 package name for Ubuntu, which seems to work fine in all the OS versions we use here (18.04, 20.04, and 22.04).

EDIT: ~~I also updated the tests to run on Ubuntu 20.04. This was semi-intended change, but I think it's better to test on a slightly newer system anyway...~~ That seemed to be failing for some reason when installing QPM. Fixing that is definitely out of scope for this PR, so I reverted to testing with 18.04.

I updated ~~"tested with" and~~ "known to support" entries in our `README.md`. The "guaranteed support" section should be updated too, but I think it's out of scope for this PR.

## Types of changes

- New feature (?)
- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -83,7 +83,7 @@ jobs:
             cxx-compiler: 'g++-10'
             use-syslibs: false
             shared-libscsynth: false
-            artifact-suffix: 'linux-bionic-gcc10' # set if needed - will trigger artifact upload
+            artifact-suffix: 'linux-focal-gcc10' # set if needed - will trigger artifact upload
 
           # - job-name: 'focal gcc10 use system libraries' # disabled - boost version incompatible
           #   os-version: '20.04'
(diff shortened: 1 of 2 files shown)


dyfer:
I've rebased the PR to resolve the merge conflict in the README.md.

joshpar (approved):

PR #5786, opened 2022-05-24
Fix vorbis, add mpeg formats (including mp3)

### LINKED ISSUE #5735: Broken Vorbis write support due to lack of OGG (write) support
## Environment

* Operating system: Windows 10

## Steps to reproduce

```
SoundFile.openWrite("C:/tmp/tst.ogg", "vorbis", "int16", 1, 44100)
```

Always returns `nil`.

Copying my comment from another thread:

The reason why Vorbis writing doesn't work in SC (and probably doesn't in SC on any OS) is that Vorbis [is a subtype](https://libsndfile.github.io/libsndfile/api.html). The main type needs to be set to OGG.

```
Major formats.
SF_FORMAT_WAV | 0x010000 | Microsoft WAV format (little endian).
SF_FORMAT_OGG | 0x200000 | Xiph OGG container

Subtypes.
SF_FORMAT_PCM_S8 | 0x0001 | Signed 8 bit data

SF_FORMAT_VORBIS | 0x0060 | Xiph Vorbis encoding.
SF_FORMAT_OPUS | 0x0064 | Xiph/Skype Opus encoding.

```

But SC has [no support](https://github.com/supercollider/supercollider/blob/18c4aad363c49f29e866f884f5ac5bd35969d828/common/SC_SndFileHelpers.hpp#L95) for OGG (writing/constant) whatsoever. SC attempts to treat Vorbis as a "major format" instead of subtype.

Vorbis (and Opus since libsndfile 1.0.31 included with SC 3.12 supports it) needs to be moved to `sampleFormatFromString` and support for OGG added to `headerFormatFromString`.

Vorbis (and even Opus) read support is present and working in SC 3.12.1 with the libsndfile that ships with the official windows binaries, only write write support is broken. (I haven't bothered updating to 3.12.2 since it apparently only fixes a bug on Mac.)

dyfer:
## Purpose and Motivation

This PR fixes write support for OGG Vorbis files, as well as adds support for MPEG layer 1/2/3 formats (if libsndfile provides that). 

Fixes #5735

A few points to consider:
- The check for vorbis/opus/mpeg support is equivalent to libsndfile _version_ check - it checks whether library version supports a given format, but not whether the support is actually compiled in. IMO it's still better than resorting to manual switches.
- I've added documentation to SoundFile about the new formats but also more details about the previously supported formats. I'm looking forward to any feedback on this
- As indicated in the helpfile, mpeg support is provided in macOS and Windows binaries, but for Linux and local builds it depends on the installed libsndfile. Particularly, on macOS libsndfile from homebrew does not have the MPEG support compiled in (even though it is updated to the required version)

## Types of changes

- Documentation
- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/SoundFile.schelp
+++ b/HelpSource/Classes/SoundFile.schelp
@@ -219,41 +219,56 @@ This is a link::Classes/String:: indicating the header format which was read by 
 definitionList::
 ## read/write header formats: ||
 table::
-## "AIFF" || Apple/SGI AIFF format
-## "WAV","WAVE", "RIFF" || Microsoft WAV format
-## "Sun", "NeXT" || Sun/NeXT AU format
-## "SD2" || Sound Designer 2
-## "IRCAM" || Berkeley/IRCAM/CARL
-## "raw" || no header = raw data
-## "MAT4" || Matlab (tm) V4.2 / GNU Octave 2.0
-## "MAT5" || Matlab (tm) V5.0 / GNU Octave 2.1
-## "PAF" || Ensoniq PARIS file format
-## "SVX" || Amiga IFF / SVX8 / SV16 format
-## "NIST" || Sphere NIST format
-## "VOC" || VOC files
-## "W64" || Sonic Foundry's 64 bit RIFF/WAV
-## "PVF" || Portable Voice Format
-## "XI" || Fasttracker 2 Extended Instrument
-## "HTK" || HMM Tool Kit format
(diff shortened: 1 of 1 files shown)


dyfer:
I think this is ready for review. Any feedback is appreciated!

joshpar (approved):

PR #5789, opened 2022-05-28
[docs] VOsc, VOsc3 - update help docs, fix examples

mtmccrea:
## Purpose and Motivation

Previously, buffer IDs were set manually, which made it break with just small modifications (like plotting the output, which may use those buffers).  Help doc Description was also outdated and redundant.

Update the help files of `VOsc` and `VOsc3` to:

- Use `Buffer:*allocConsecutive`.
- Remove redundant `\b_alloc` messages.
- Modify examples so they all work.
- Generally cleaned up variables, comments, etc. and free resources.
- Clarify Description and update links.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer (approved):
Thanks @mtmccrea !

PR #5794, opened 2022-06-05
Topic/ezknob align number

madredeuz:
## Purpose and Motivation
I use a lot the GUI and most of time I prefer to center the number in EZKnob, so I thought it is a good idea to include this option.

## Types of changes

- New feature

## To-do list

- [ x] Code is tested
- [ x] All tests are passing
- [ ] Updated documentation
- [ x] This PR is ready for review

### DIFF
--- a/SCClassLibrary/Common/GUI/Base/EZKnob.sc
+++ b/SCClassLibrary/Common/GUI/Base/EZKnob.sc
@@ -130,6 +130,9 @@ EZKnob : EZGui {
 
 	centered{ ^knobView.centered }
 
+	align_ { arg aligment; numberView.align_(aligment) }
+	align{ ^numberView.align }
+
 
 	setColors{arg stringBackground,stringColor,numBackground,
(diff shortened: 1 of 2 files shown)


dyfer:
Hi @madredeuz 
Thanks for the PR!
I do like the view of centered number in EZKnob in certain cases, but I'm not sure if singling out that setting is desirable. And if it was singled out, it should probably be called more explicitly like `.numberAlign` or something like that... In that case it is getting closer to writing `.numberView.align`, which is already available. I wonder how others feel about this.

BTW this branch also contains a commit from your other PR, if this is go through, you'd need to remove the other commit e.g. through rebasing and force pushing.

madredeuz:
Hi @dyfer, sorry, is the first time I work with git and I had a few problems this morning, so probably I did something wrong uploading the file. 

You are right and there is really no need to add align. This is a hack I did long time ago and didn´t noticed I could use numberView.align, I always check documentation and sometime I missed things, so I think we can close it and leave it as it is.

Thanks

PR #4312, opened 2019-02-18
class library: do not convert value before rate check

telephon:
For deriving the controls of a `ProxyNodeMap`’s values, we cannot convert
them to control inputs before rate checking, because this will return
symbols for busses that have no explicit rate information (like `[\a1,
\a2]`).

This was introduced in 1b88d80c7dd761170fbf970f813466feac9cddcf.

Thanks to James Harkins for analysis (see #4311).

Issue #4311, opened 2019-02-18
ProxyNodeMap:controlNames faulty logic breaks NodeProxy audio patching


jamshark70:
## Environment

* SuperCollider version: 0685eb153
* Operating system: Ubuntu Studio 18.04

## Steps to reproduce

https://github.com/supercollider/supercollider/commit/1b88d80c7dd761170fbf970f813466feac9cddcf broke this.

The addition of `value = value.asControlInput;` into ProxyNodeMap:controlNames means that the `<<>` and `<>>` operators are broken for audio-rate NodeProxy inputs. You can use them once. After that, the target proxy's nodeMap will return a *control*-rate ControlName for an input that was defined at audio rate.

```supercollider
s.boot;
p = ProxySpace.new.push;

~out = { \in.ar(0!2) }; ~out.play(vol: 0.2);

~osc = { |freq = 440| Saw.ar(freq).dup };
~out.controlNames.detect { |cn| cn.name == \in };  // 'audio', [0, 0]
~osc <>> ~out;
~out.controlNames.detect { |cn| cn.name == \in };  // 'control', [ a(n), a(n+1) ]
~osc <>> ~out;  // 2nd time, "no matching input"
```

## Expected vs. actual behavior

ProxyNodeMap tries to determine the input's rate according to the value stored in the node map.

After mapping `~osc <>> ~out`, `~out`'s node map contains a reference to `~osc` for its `\in` input.

```
~out.nodeMap[\in]
-> NodeProxy.audio(localhost, 2)
```

At this point, the old logic in ProxyNodeMap:controlNames would ask the source NodeProxy for its rate, returning `\audio`, and everything was fine.

The new logic attempts to account for "arbitrary objects," so it calls `.asControlInput` on the node map value.

```
~out.nodeMap[\in].asControlInput
-> [ a6, a7 ]
```

... and assumes that `.rate`, on this, will do something sensible. But in fact, there is no logic in SC to identify the rate of a bus-mapping string.

```
~out.nodeMap[\in].asControlInput.rate
-> scalar
```

It looks to me like a case of shortcuts being taken in the object design. If the node map is going to be a source for ControlName objects, then it should store ControlName objects alongside the values -- that is, you really need a dictionary of `name -> ProxyInputData(controlName, value)`. Instead, it's storing only values, and then attempting to recover information from the values (when that information more properly comes from the specification of the input, directly).

Possibly revert 1b88d80c7 until there is a proper fix. The current situation is unusable.

[...]

PR #5797, opened 2022-06-07
Fix NodeMap

dyfer:
## Purpose and Motivation

This PR fixes an issue with `NodeMap -controlNames`. The fix was proposed by @telephon.

The test that was failing because of this issue (`TestProxyNodeMap -test_nodeMap_controlNamesConvertsObjectToControlInput`) is now also enabled in our test suite.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
@@ -135,7 +135,7 @@ ProxyNodeMap : NodeMap {
 			var rate = if(value.respondsTo(\rate) and:{
 				value.rate == \audio
 			}) { \audio } { \control };
-			res = res.add(ControlName(key, nil, rate, value))
+			res = res.add(ControlName(key, nil, rate, value.asControlInput))
 		};
 		^res
 	}
--- a/testsuite/scripts/gha_test_run_proto.json
+++ b/testsuite/scripts/gha_test_run_proto.json
@@ -83,12 +83,6 @@
             "skipReason":"UnitTest fails when run in qpm on Linux (FIXME)"
         },
         {
-            "suite":"TestProxyNodeMap",
-            "test":"test_nodeMap_controlNamesConvertsObjectToControlInput",
-            "skip":true,
-            "skipReason":"UnitTest has bugs (FIXME)"
-        },
-        {
             "suite":"TestSCDoc",
             "test":"test_helpSourceDirs_includedExtensions",
             "skip":true,


SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
```diff
@@ -135,7 +135,7 @@ ProxyNodeMap : NodeMap {
 			var rate = if(value.respondsTo(\rate) and:{
 				value.rate == \audio
 			}) { \audio } { \control };
-			res = res.add(ControlName(key, nil, rate, value))
+			res = res.add(ControlName(key, nil, rate, value.asUGenInput))
```
telephon, line 138: should be `asControlInput`, because the `ControlNames` are all used outside a UGen graph.

jamshark70:
I remember an issue with audio rate mapping that I think is related to this, so be sure to test this very carefully.

https://github.com/supercollider/supercollider/pull/4312/files *removes* `value = value.asControlInput;` from the method being discussed here.

But the new change adds `asUGenInput` later in the execution flow. It might be OK, then, or it might not.

`TestNodeProxy:test_map_rates` might be the (or a?) relevant unit test, but it's probably worth throwing some other test cases at it.

SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
dyfer: Thanks for catching this! Sorry for incorrectly copying your solution.

dyfer:
> EDIT: At least, the test case in #4311 is not broken after this change.

As I wrote in the comment under the linked issue, I'm not sure what the broken state looks like. The example code from the issue behaves the same when run on a recent `develop`, as well as in this PR, so I think this PR doesn't break it.

jamshark70:
> I'm not sure what the broken state looks like.

The broken state is shown in the description of #4311 -- if you `<>>` map an audio rate control twice, the control's rate is reported incorrectly and you get an error message "no matching input."

This was fixed, so you won't see the problem in current develop.

My concern was about a regression issue, but, running the #4311 reproducer doesn't show the error after the patch.

Sometimes it isn't clear in JITLib what, exactly, a method should be doing internally. #4311 was that controls were being constructed with a conversion in the wrong place. Fixing that by removing the conversion caused a different issue (ControlName defaults were not converted) -- where, even then there was some confusion over asControlInput vs asUGenInput. That's the other part of the reason for my concern. I am confident that the change in this PR would *not* to regress back to the error in #4311. But I'm not 100% confident that there wouldn't be other fallout.

dyfer:
> The broken state is shown in the description of #4311 -- if you `<>>` map an audio rate control twice, the control's rate is reported incorrectly and you get an error message "no matching input."

@jamshark70 Yes, I was not getting the "no matching input" error. However, what's _not_ described (and not clear to someone not using JITlib regularly...) is what should be the result of running that line - what should be returned? Should there be any change to the sound? etc etc :)

Anyway, thanks for confirming this doesn't break that case.

As this fix was proposed by @telephon, I hope that the two of you can figure out if this PR is good or not :)

telephon:
I do see a potential problem, and I am not sure how to resolve it.

1. nodeProxy.controlNames should definitely return valid ControlNames. 
2. If a value of the NodeMap is a NodeProxy (representing a mapping), what value the ControlName should have?

Right now, we were thinking, let's convert it with `asControlInput`, which for a NodeProxy will return the correct `busArg`, i.e. something like `a32` or `c61`.

But I see places in `NodeProxy` where we use `ControlName` to transport and assign values. An example is `getKeysValues`, which is used in a number of other places. So my change would propagate into those places.

While I am not a typing fanactic, here is an example where a bit of more explicit typing would have been helpful.

So (1) should be explicitly negated. These are "special" ControlNames, whose values may be node proxies or other objects. I can't do this now, saving for later.

telephon:
Thank you for the discussion – small thing, but this saved some troubles! Here is a PR that fixes the test instead:  #5801

PR #5804, opened 2022-06-21
clang-format in docker

capital-G:
## Purpose and Motivation

Installing clang-format is a rather tedious work, especially if there are no binaries for the specific version on the OS.
This PR allows to run the lint/format process inside a docker container in 2 simple statements on all major platforms (windows, linux and macos) by running

* `docker build -t clang-format .` - to build an image with the proper clang-format binary - takes around 1 minute and shall be executed in `tools/`
* `docker run -v ${PWD}:/sc/ clang-format formatall` - to create a container from the image, e.g. in the root dir of the repo, and run formatall

The documentation (https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions) is not yet modified but is also not attached to code. I want to wait til I get approval for inserting this and I am unsure who has privileges to modify the wiki.

Edit: Rewrote git history of this branch to fix some git errors and switched from alpine docker image to ubuntu 18.04 which represents the state of the CI pipeline and also has the benefit that it has a binary available.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 RUN apt-get update && \
     apt-get install --yes \
@@ -6,10 +6,6 @@ RUN apt-get update && \
         git \
         clang-format-8 && \
     rm -rf /var/lib/apt/lists/*
-
-# add python3.6->python3 alias
-# clang needs python2 which already takes the python name
-RUN ln -s /usr/bin/python3.6 /usr/bin/python3
 
 ENV SC_CLANG_FORMAT=/usr/bin/clang-format-8
 ENV SC_CLANG_FORMAT_DIFF=/sc/tools/clang-format.py
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,0 +1,19 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && \
+    apt-get install --yes \
+        python3.6 \
+        git \
+        clang-format-8 && \
+    rm -rf /var/lib/apt/lists/*
+
+# add python3.6->python3 alias
+# clang needs python2 which already takes the python name
+RUN ln -s /usr/bin/python3.6 /usr/bin/python3
+
+ENV SC_CLANG_FORMAT=/usr/bin/clang-format-8
+ENV SC_CLANG_FORMAT_DIFF=/sc/tools/clang-format.py
+
+WORKDIR /sc/
+
+ENTRYPOINT [ "python3", "/sc/tools/clang-format.py" ]


capital-G:
I somehow got 2 commits (https://github.com/supercollider/supercollider/pull/5804/commits/a2e58ce6af14f573d26abdf5d56824b212aba205 and https://github.com/supercollider/supercollider/pull/5804/commits/608a4fdcb0c002a14c6d021062c254c2957205a0) that are not associated with this PR into it.
If this gets a chance of merging I would remove them [although, how? - `git reset --soft` and a rewrite?]

capital-G:
I saw that the CI pipeline uses `ubuntu-18.04` https://github.com/supercollider/supercollider/blob/647d7bb23d6211e3d681a29b2f196924d403e290/.github/workflows/actions.yml#L16-L33
If there is a possibility of a merge I would translate this to ubuntu to be more close to the CI.

tools/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:18.04
```
tdug, line 1: I may suggest using a more recent version, if possible. `22.04` is now the most recent LTS version.

capital-G: Yeah, normally I would agree to this but the build pipeline also relies on 18.04

https://github.com/supercollider/supercollider/blob/647d7bb23d6211e3d681a29b2f196924d403e290/.github/workflows/actions.yml#L15-L18

and 20.04 is the last version to have clang-format-8 as a binary-package (https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=clang-format-8&searchon=names), otherwise one has to build the version oneself which takes around 10 minutes to build it within the container and by using this version we can skip this energy consuming task. Also 18.04 still receives security updates up to 2028: https://ubuntu.com/about/release-cycle

capital-G:
Bump - what is missing for a merge?

dyfer:
Hello @capital-G 
Sorry for the delay.
Please bear with me, as I'm not very familiar with docker.

I did install it and it runs fine. I think I understand the logic of this and I was about to ask about handling of paths, wouldn't it be better to provide the path to the python script directly etc, but instead here is my suggestion:

Would it be possible to change this so that it _only_ provides clang-format? My thinking is that it would be great if this could integrate seamlessly into the current python scripts. If we could just use it to get clang-format, then my preference would be to set that docker command as `SC_CLANG_FORMAT` [environment variable](https://github.com/supercollider/supercollider/blob/develop/tools/clang-format.py#L570) and use python scripts directly (i.e. using system interpreter, not the one in the container). That way the whole formatting workflow could stay as is, with no changes, IIUC. I'd still like to see this documented in the wiki of course, but then the documentation would be limited to 1) building docker image and 2) setting the env variable. What do you think?

Other minor comments:
- maybe call the image `clang-format-8`?
- I think we should go with ubuntu 20.04. The reason we still use 18.04 in the CI is that I encountered issues when I tried upgrading it. The SC support policy is that we officially support two major OS versions back, so technically that's 22.04 and 20.04 currently. I'd just go with the newest version that provides `clang-format-8`.

capital-G:
> Would it be possible to change this so that it _only_ provides clang-format?

So something like a [virtual environment](https://realpython.com/python-virtual-environments-a-primer/) in python where certain environment variables to paths of binaries are replaced?
The problem here is that a Docker container is like a VM and not just a namespace for binary execution - this means that we need to mount any files of the host to the container if the container should have access to it (which is essential to this procedure) and this makes it really hard b/c we need to know at all times where the source files are and need to be sure that 

Also I think having a fixed python version within the container is an advantage as it provides a fixed working set, allowing to be sure to support specified python versions.

> * maybe call the image `clang-format-8`?

Good point - the name of the image is specified during build time, so something for documentation.

> * I think we should go with ubuntu 20.04.

Implemented in https://github.com/supercollider/supercollider/pull/5804/commits/b1aa97f08fbc4153bad8e9f9837b30dab1bb800c

> I'd probably like it to have the dockerfile moved one level down, maybe inside clang-format-docker or something like that

Are you sure? I don't prefer folders with a single file and at the same time I don't think that people get confused by this file if they don't know what Docker is.

dyfer:
> So something like a [virtual environment](https://realpython.com/python-virtual-environments-a-primer/) in python where certain environment variables to paths of binaries are replaced? 

I'm not sure if this is the best metaphor. I'd just like it be be able to be picked up by the existing python scripts.

> The problem here is that a Docker container is like a VM and not just a namespace for binary execution - this means that we need to mount any files of the host to the container if the container should have access to it (which is essential to this procedure) and this makes it really hard b/c we need to know at all times where the source files are and need to be sure that

A quick search about this provides some clues about `docker exec` or `docker exec -i`. Are you sure it's not possible to use it as if it was just a binary? Would you be willing to experiment to try that? 

I think it would be really beneficial to incorporate this into python scripts. Ideally, so that all formatting procedures described in the wiki (e.g. cmake target) can be done with the docker version. (Again, if that's at all possible. I'm at a disadvantage here, haven't had worked with docker much).

> > I'd probably like it to have the dockerfile moved one level down, maybe inside clang-format-docker or something like that
> 
> Are you sure? I don't prefer folders with a single file and at the same time I don't think that people get confused by this file if they don't know what Docker is.

Yes, that would be my preference... but maybe let's have someone else chime in on this.

capital-G:
> A quick search about this provides some clues about `docker exec` or `docker exec -i`. Are you sure it's not possible to use it as if it was just a binary? Would you be willing to experiment to try that?

I have some professional experience with Docker and from my experience it is best to quickly get into the fixed state of a Docker container as possible in order to decrease the number of possible states to support.
There is of course `docker exec` but this is more used to access a responsive shell-like environment for debugging. You propose something like Flatpack or Snap which is not available outside of Linux.

The more interesting question to me would be what scenarios are not covered by the proposed approach which then could be documented as well.

Essentially

```shell
python3 clang-format.py <command-to-run>
```

can be replaced by

```shell
docker run -v ${PWD}:/sc/ clang-format <command-to-run>
```

or if something else need to be run in the environment

```shell
docker run --entrypoint <BINARY_TO_RUN> -v ${PWD}:/sc/ clang-format <ARGS_FOR_BINARY>
```

dyfer:
> I have some professional experience with Docker and from my experience it is best to quickly get into the fixed state of a Docker container as possible in order to decrease the number of possible states to support. 

I can see how this is beneficial.

> There is of course `docker exec` but this is more used to access a responsive shell-like environment for debugging. 

I don't have any contrargument for this, with my limited experience. But if it could be used the way I descried bearlier, I'd still consider it.

> You propose something like Flatpack or Snap which is not available outside of Linux.

Well, that wasn't my intention. But I also have limited experience with Flatpak or Snap. I was trying to figure out a way that's the most transparent with the current procedures.

> The more interesting question to me would be what scenarios are not covered by the proposed approach which then could be documented as well.

I want to be sure that we can support all the procedures described in the [wiki](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions#reformatting-as-you-work-normal-usage).
IIUC, in this PR you provide a process for running `clang-format.py`. However, we also have `lint.py` and `format.py` scripts generated at the build time, and there are cmake targets. Is it possible to be able to use docker with them all?

capital-G:
> I don't have any contrargument for this, with my limited experience. But if it could be used the way I descried bearlier, I'd still consider it.

I mean you can shell script anything but I have not come across it yet. The main problem will be that clang needs access to the actual cpp files which needs to be mounted at the proper location at runtime to the container - so if the python script calls clang from different (relative) directories you need to keep account for this as you need to mount the files (relatively) within the container - seems rather error prone to me.

> IIUC, in this PR you provide a process for running `clang-format.py`. However, we also have `lint.py` and `format.py` scripts generated at the build time, and there are cmake targets. Is it possible to be able to use docker with them all?

It allows you to run clang-format without installing it on your system. On my machine (macOS) it took over 1 hour and multiple GB to install it because no binary was available for the required clang-format version - with Docker it boils down to 1 minute setup and you don't have to install clang on your system and it works on all major OS the same (Windows/Linux/macOS) and you can easily use fixed versions.

As I understand `format.py` and `lint.py` are just redirects to `tools/lint_or_format.py` with different args. Those scripts could add a `--docker` flag (or create new files to use docker)?
Maybe introducing a `-DDOCKER=yes` flag for cmake or is it better to create new targets?
Basically one would need to take a look everywhere where the script is called and replace it with a docker call if the flag is set. Would this be a good addition to this PR? I don't have much experience with cmake.

dyfer:
> > I don't have any contrargument for this, with my limited experience. But if it could be used the way I descried bearlier, I'd still consider it.
> 
> I mean you can shell script anything but I have not come across it yet. 

I was really imagining this could make it simple - and again, correct me if I'm wrong.

Our Python scripts can pick up clang-format path from and environment variable `SC_CLANG_FORMAT` (e.g. `export SC_CLANG_FORMAT=/what/ever/path/clang-format-8`. Couldn't we do something like `export SC_CLANG_FORMAT="docker exec blah"` and have the scripts work with it transparently? Wouldn't that solve the need to mount files within the container?

> 
> > IIUC, in this PR you provide a process for running `clang-format.py`. However, we also have `lint.py` and `format.py` scripts generated at the build time, and there are cmake targets. Is it possible to be able to use docker with them all?
> 
> It allows you to run clang-format without installing it on your system. 

Yes, I understand that. But your solution is valid for only one of _multiple_ ways we run formatting. My point is that it should account for all existing workflows - or we should change those, but that would be a separate issue.

> As I understand `format.py` and `lint.py` are just redirects to `tools/lint_or_format.py` with different args. Those scripts could add a `--docker` flag (or create new files to use docker)? Maybe introducing a `-DDOCKER=yes` flag for cmake or is it better to create new targets? Basically one would need to take a look everywhere where the script is called and replace it with a docker call if the flag is set. Would this be a good addition to this PR? I don't have much experience with cmake.

Adding a flag could be one solution. Or maybe using another environment variable that would be picked up by the scripts (the way they pick up clang-format path now).

---
The main point here is that this needs to work with existing workflows and have as little technical debt as possible.

However, it is conceivable that the best solution would be to use newer clang-format. Unfortunately people who set up formatting originally and had the vision for it are not active anymore, and I don't have the same depth of experience with formatting large projects. If you think we should consider upgrading clang-format we use, then we could start a new issue to discuss this and halt working on this PR until there's a consensus about the best way to move forward.

capital-G:
> Couldn't we do something like `export SC_CLANG_FORMAT="docker exec blah"` and have the scripts work with it transparently? Wouldn't that solve the need to mount files within the container?

Well `docker exec blah` would not have access to any files on the host filesystem without mounting the proper directory.
If the python script calls`clang-format` in `tools/` the mounting point would need to be different than `clang-format` in `./`, therefore we would need to calculate the proper relative file path based upon where `clang-format` was called - probably possible in Unix but as we need to support Windows as well. I wouldn't go there. 
To me it seems more easier to just change the invocation of the script than going through each step where the script invokes clang-format. Going through the [wiki](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions#reformatting-as-you-work-normal-usage) I don't see any workflow that is not possible/reflected by modifying the script invokation.
Basically the script can be called with any args via

```shell
docker run -v ${PWD}:/sc/ clang-format any args for the script
```

> My point is that it should account for all existing workflows

Which formation steps are done that are not reflected by the script? GitHub actions could also be changed to use the Docker container, see e.g. another PR https://github.com/supercollider/supercollider/pull/5487/commits/c68d134d93159a2075fab6da563171f952996397

> The main point here is that this needs to work with existing workflows and have as little technical debt as possible.

I would argue that we already have this technical debt as clang-8 is deprecated (at least [no official macOS binaries](https://releases.llvm.org/download.html#8.0.1)?) and it seems it is difficult to update in an easy manner. Docker would have the advantage that the used version is also manifested as code (via a Dockerfile) and not rely on the clang version of the host machine.

> If you think we should consider upgrading clang-format we use, then we could start a new issue to discuss this and halt working on this PR until there's a consensus about the best way to move forward.

grafik

I did not know upgrading clang would be an option. It always seemed to me its better a topic that you don't discuss :D
Still, clang is a really clunky/heavy dependency and as CPP/cmake has not (?) a proper dependency management it seems best to rely on Docker images/containes for administration tasks which provide a fixed reference environment which can easily be distributed and versions can be upgraded via code (see above).

I have some experience enforcing formatting procedures in Python projects but not in a CPP environment.
I myself use [pre-commint](https://pre-commit.com/) on [e.g.](https://github.com/capital-G/musikinformatik-sose2021/blob/main/.pre-commit-config.yaml) my repos, there is also [a hook for clang-format](https://github.com/pre-commit/mirrors-clang-format) which relies on a [python/pip distributed clang-format-8 binary](https://pypi.org/project/clang-format/#history), making the installation really easy (or so it seems to me - it boils down to having Python installed and call `pre-commit install`). It would also be possible to [use docker within a pre-commit hook](https://pre-commit.com/#docker), allowing for arbitrary tasks in a fixed environment, but this would require python (and docker) as a development dependency.

I really enjoy the comfort and restrictiveness that comes along with `pre-commit` (you can not commit without passing the added tests and the tests are installed and configured by the repo config file automatically and can be updated) but I don't have the impression that this would fit the *not-so-strict* SC community - but maybe I have a false impression of the SC community?

dyfer:
> > Couldn't we do something like `export SC_CLANG_FORMAT="docker exec blah"` and have the scripts work with it transparently? Wouldn't that solve the need to mount files within the container?
> 
> Well `docker exec blah` would not have access to any files on the host filesystem without mounting the proper directory. If the python script calls`clang-format` in `tools/` the mounting point would need to be different than `clang-format` in `./`, therefore we would need to calculate the proper relative file path based upon where `clang-format` was called

Thanks for explanation.
> Going through the [wiki](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions#reformatting-as-you-work-normal-usage) I don't see any workflow that is not possible/reflected by modifying the script invokation. Basically the script can be called with any args via
> 
> ```shell
> docker run -v ${PWD}:/sc/ clang-format any args for the script
> ```


> Which formation steps are done that are not reflected by the script? GitHub actions could also be changed to use the Docker container, see e.g. another PR [c68d134](https://github.com/supercollider/supercollider/commit/c68d134d93159a2075fab6da563171f952996397)

Possible - sure. Reflected? I don't think so? E.g. from the wiki:
```
Reformatting as you work (normal usage)
If you are developing a C++ feature or bug fix, a typical workflow might look like this:

Write some code, build, test
Use your editor's integration or run ./format.py in your build directory to format your code
Commit your changes with git
```

How would you need to change the `./format.py` script for this to work?
There are also cmake targets, which I haven't looked into.
While you did offer to update the wiki after merging this, it would be helpful to have a draft of proposed changes (e.g. how the commands would change), which would be helpful for me to test. 


> > The main point here is that this needs to work with existing workflows and have as little technical debt as possible.
> 
> I would argue that we already have this technical debt as clang-8 is deprecated (at least [no official macOS binaries](https://releases.llvm.org/download.html#8.0.1)?) and it seems it is difficult to update in an easy manner. Docker would have the advantage that the used version is also manifested as code (via a Dockerfile) and not rely on the clang version of the host machine.

Yes, I get the advantage of having the docker image. But docker would be yet another (soft) dependency.
(MacOS Binaries are available through homebrew I think, but that probably won't last forever)

> > If you think we should consider upgrading clang-format we use, then we could start a new issue to discuss this and halt working on this PR until there's a consensus about the best way to move forward.
> 
> I did not know upgrading clang would be an option. 

I don't know whether it is. But I think this could still be discussed. As I said, I don't have the long-term vision for maintaining formatting, unfortunately.


> I have some experience enforcing formatting procedures in Python projects but not in a CPP environment. I myself use [pre-commint](https://pre-commit.com/) on [e.g.](https://github.com/capital-G/musikinformatik-sose2021/blob/main/.pre-commit-config.yaml) my repos, there is also [a hook for clang-format](https://github.com/pre-commit/mirrors-clang-format) which relies on a [python/pip distributed clang-format-8 binary](https://pypi.org/project/clang-format/#history), making the installation really easy (or so it seems to me - it boils down to having Python installed and call `pre-commit install`, after this the linter will be run over the files before each commit on all changed files). It would also be possible to [use docker within a pre-commit hook](https://pre-commit.com/#docker), allowing for arbitrary tasks in a fixed environment, but this would require python (and docker) as a development dependency.

Whoa, so could we get clang-format-8 through python? Maybe that's the simplest way? Python is already required for running formatting scripts...

> I really enjoy the comfort and restrictiveness that comes along with `pre-commit` (you can not commit without passing the added tests and the tests are managed by the repo config file and can be updated and configured via it) but I don't have the impression that this would fit the _not-so-strict_ SC community - but maybe I have a false impression of the SC community?

I can't speak for the community, but at some point I was interested in pre-commit hooks, though I haven't tried them.
Would they have to be enabled for all git users? Or is this something a user could opt-in to (opt-out of)? If that would have to be for everyone, the that's probably not a desirable change.

capital-G:
I created a new branch https://github.com/capital-G/supercollider/tree/init-pre-commit where clang-format is managed by a pre-commit hook, making the format scripts obsolete?
I could not use 8.0.1 so I updated to 14.0.6 (see below) - running the linter of these files returns in 199 changed files, most of them have minor changes like

```diff
diff --git a/common/SC_List.h b/common/SC_List.h
index 45505a510..1684099a6 100644
--- a/common/SC_List.h
+++ b/common/SC_List.h
@@ -34,7 +34,7 @@ A doubly linked list template.
 
 template <class T> class Link {
 public:
-    Link(): mNext(this), mPrev(this) {}
+    Link(): mNext(this), mPrev(this) { }
 
     T* Prev() { return static_cast<T*>(mPrev); }
     T* Next() { return static_cast<T*>(mNext); }
```

or some changed line-breaks. Or maybe there is a clang-format-config file I did not pick up?
One advantage: If we decide to use pre-commit we can simply bump the clang-format version within the pre-commit config file and it will be picked up by everyone with the new config file automatically, allowing to have different kind of clang-format versions per branch because the version is stated in code.

> How would you need to change the `./format.py` script for this to work?

There are some possible scenarios, I would suggest

* check installed clang-format-8 format - if available use this
* check if docker is installed - use docker container instead by using command posted above
* else fail and post a link on setup of clang-format workflow

or switch to pre-commit, making the format scripts obsolete.

> Whoa, so could we get clang-format-8 through python? Maybe that's the simplest way? Python is already required for running formatting scripts...

Yes, [pypi provides a clang-format binary](https://pypi.org/project/clang-format/#history) but the 8.0.1 version is only a linux binary (https://pypi.org/project/clang-format/8.0.1/#files), only newer files (10.0+) also have a windows/macOS binary. Maybe we can ask the maintainer to add a binary for 8.0.1 as well but maybe we should adapt a flexible versioning of clang-format?

> I was interested in pre-commit hooks, though I haven't tried them.
> Would they have to be enabled for all git users? Or is this something a user could opt-in to (opt-out of)?

To activate pre-commit you need to run `pre-commit install` in the git-pepo root dir, otherwise the pre-commit config file is ignored and no hooks are run.
Yet only files you change get checked, so maybe its best to make pre-commit optional but advised, but let pre-commit run as github-action nonetheless.

If you want to commit files pre-commit checks the files according to its defined hooks - if something is off, e.g. linting, the commit will not be executed, instead it will automatically modify the *offending* files (in this case lint it accordingly), allowing you to view the proposed changes via git diff (like above) and accept/modify them by adding them to the still open commit as well - once all checks passed the commit will be executed. You can ignore steps via an ENV variable or just perform `pre-commit uninstall` to completely make your repo free of any pre-commit hooks.

dyfer:
After the dev meeting on 2022.09.11 I think we'd like to explore a possibility of upgrading the clang-format version that we use before considering this PR. Thank you for your patience!

capital-G:
> After the dev meeting on 2022.09.11 I think we'd like to explore a possibility of upgrading the clang-format version that we use before considering this PR. Thank you for your patience!

Glad to hear on the update of clang-format, although I think this PR is rather fitting for rolling out a new version of clang as it automates any setup regardless of the host OS and allows to pinpoint the version to use as code.
May I propose the other way around, first integrate this PR and then pull up the version? If this PR or the pre-commit one gets merged updating clang is as easy as changing

* https://github.com/capital-G/supercollider/blob/init-pre-commit/.pre-commit-config.yaml#L15 if pre-commit is considered
* https://github.com/capital-G/supercollider/blob/b1aa97f08fbc4153bad8e9f9837b30dab1bb800c/tools/Dockerfile#L7 - for docker based workflow (this PR)

dyfer:
> May I propose the other way around, first integrate this PR and then pull up the version?

At this time we are really trying to avoid adding further dependencies (like docker) to the project. Thank you for your understanding.

capital-G:
IMO if I have to choose between docker and clang8 as dependency I would choose docker any day because it is platform independent, scriptable/stateless and a generic tool which can also keep other dev dependencies such as python fixed - not to mention how difficult it is to install a specific version of clang if it is not available via the package manager of choice (took me over 1 hour last time b/c brew had no binaries and it compiled it from scratch).
Having non-deterministic output in different versions of clang-format is very problematic, making it mandatory to use fixed versions.

Issue #5754, opened 2022-04-06
[FR] Expand Git interface to include submodule handling and use it in Quarks


madskjeldgaard:
## Motivation
The Git class currently doesn't support submodules which prevents for example using submodules in Quark repos.

## Description of Proposed Feature

Add a submodule method to `Git` and add submodule recursion to the `Quark` install step, eg adding the `--recurse-submodules` git flag to the clone command. This will download any submodules if the quark has them.

PR #5806, opened 2022-06-29
Recurse submodules by default

madskjeldgaard:
This makes git.clone recurse submodules by default so that it's possible to make Quarks with git submodules registered.

## Purpose and Motivation

Fixes https://github.com/supercollider/supercollider/issues/5754

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Quarks/Git.sc
+++ b/SCClassLibrary/Common/Quarks/Git.sc
@@ -8,12 +8,14 @@ Git {
 	*new { |localPath|
 		^super.new.localPath_(localPath)
 	}
-	clone { |url|
-		this.git([
+	clone { |url, additionalFlags(["--recurse-submodules"])|
+        var command = [
 			"clone",
 			url,
 			thisProcess.platform.formatPathForCmdLine(localPath)
-		], false);
+		] ++ additionalFlags;
+
+		this.git(command, false);
 		this.url = url;
 	}
 	pull {


madskjeldgaard:
I guess this should also pertain to Git.pull so that Quark updating updates the submodules too

dyfer:
Hello @madskjeldgaard, did you have a chance to give some thought to the points I've raised?

cdbzb:
Hi all - I don't think in our case that there is any situation in which we don't want to check out submodules.

The downside is generally storage or performance, that might make a difference in large libraries but I can't see an issue in any quarks as the whole universe of quarks is so small.

I think it would be reasonable to have it always on.

as for the idea of implementing it in Quark.install instead of in Git.clone, I think we can instead ask the question of whether there is anywhere in the codebase now where we would prefer to set the `additionalFlags` arg to nil. Otherwise we can accept this PR and simply leave `--recurse-submodules` as the default

cdbzb (approved):

elgiano:
I also don't see any case for not cloning submodules, but at the same time I don't think it's a good idea to have it as a default like this, because it would be hard to switch off should one ever desire to do so. How about we add a "recurseSubmodules" field to the quark file, which defaults to true unless the Quark author sets it to false?

> should this be just on, or configurable in a quark file? (in the latter case it is getting more complicated, we'd need to clone first, parse the quark file, and then run git submodule update --init --recursive)

The system is already checking out quark files before installing, because it needs to check for eventual dependencies. So we could also check `data["recurseSubmodules"] != false` and run git submodule update, at both install and update.

PR #5810, opened 2022-07-11
Topic/help minor fixes   mainly patterns

redFrik:
just some minor fixes and a helpfile for Pgtpar.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Pproto.schelp
+++ b/HelpSource/Classes/Pproto.schelp
@@ -32,7 +32,7 @@ When Pproto ends, these eventTypes will respond to the strong::cleanup:: call by
 The function yields each event. That event is then performed with possible modifications by enclosing patterns and the player (either an link::Classes/EventStreamPlayer:: or a link::Classes/ScoreStreamPlayer::). The resultant event is returned to the function where it can be assigned to a key within the protoEvent.
 
 The patternarray is played using Pfpar, a variant of Ppar that ends when any of its subpatterns end. In this way,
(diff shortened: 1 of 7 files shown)


HelpSource/Classes/Pproto.schelp
```diff
@@ -48,14 +48,14 @@ code::
 (
 SynthDef(\help_playbuf, { | out=0, bufnum = 0, rate = 1, startPos = 0, amp = 0.1, sustain = 1, pan = 0, loop = 1|
 	var audio;
-	rate = rate * BufRateScale.kr(bufnum);
-	startPos = startPos * BufFrames.kr(bufnum);
+	rate = rate * BufRateScale.ir(bufnum);
+	startPos = startPos * BufFrames.ir(bufnum);
```
dyfer, line 52: Why change `.kr` to `.ir` here? 
My thinking is that an expert user will know to make this optimization, however for a less experienced one it may lead to a surprising result ("why didn't the rate update after replacing the buffer?"). I _is_ quite unlikely in this example, but as users copy-paste code from help, I'd have a slight preference to leave these as `.kr`.

dyfer (changes requested):
Thank you for this!
I only have one request/suggestion, see below.

HelpSource/Classes/Pproto.schelp
dyfer: All right, I see. As this is for the granular stuff and indeed the .ir is used in the same SynthDef, let's keep the change. Thanks!

dyfer (approved):
Thanks!

PR #4050, opened 2018-09-09
[classlib] FFT: fix for using regular Buffer

dyfer:
Purpose and Motivation
----------------------

This is a fix for using regular Buffer with FFT UGens. I take no credit for this - @jamshark70 provided the solution in the linked issue; I wanted to get the ball rolling though, as his solution has solved the problem I had. I hope I'm not breaking etiquette by doing this...

Fixes #3399.

Types of changes
----------------

- Bug fix (non-breaking change which fixes an issue)

Checklist
---------

- [ ] All previous tests are passing
- [ ] Tests were updated or created to address changes in this PR, and tests are passing
- [ ] Updated documentation, if necessary
- [ ] This PR is ready for review

Remaining Work
--------------

I have no thorough understanding on the details of FFT processing in SC, so I can't comment on the validity of the solution, other than that it did fix an issue I had. It would be good to test this more to see if it breaks anything else with FFT.

See the reproducer below. Without this fix, if you swap ```LocalBuf``` for ```Buffer```, the synth will not post values, nor create the "trigger" signal from the FFT chain:

```SuperCollider
(
x.free;
b.free;
fork{
	b = Buffer.alloc(s, 32768);
	s.sync;
	x = {
		var in, chain;
		var trig, ff;
		in = WhiteNoise.ar(0.1);
		chain = FFT(LocalBuf(32768), in); //polling works
		// chain = FFT(b, in); //polling doesn't work
		chain.poll;
		trig = ( chain >=0 );
		trig.poll;
		IFFT(chain); // inverse FFT
	}.play;
};
)
```
Again, thanks to @jamshark70 for the solution!

PR #5811, opened 2022-07-12
Classlib: Invert the SynthDef topological sort (to fix a wide-graph case)

jamshark70:
## Purpose and Motivation

Per Binary SynthDef Format help, "Unit generators should be listed in an order that permits efficient reuse of connection buffers, which means that a depth first topological sort of the graph is preferable to breadth first."

But note the difference between these cases:

```
// OK case: an entire branch is calculated before being muladd-ed in
(
SynthDef(\addTest, {
	var sig = Array.fill(5, { |i| SinOsc.ar(220 * (i+1), mul: Rand(0.5, 1.0)) });
	Out.ar(0, sig.sum)
}).dumpUGens.add;
)

[ 0_SinOsc, audio, [ 440, 0.0 ] ]
[ 1_Rand, scalar, [ 0.5, 1.0 ] ]
[ 2_*, audio, [ 0_SinOsc, 1_Rand ] ]
[ 3_SinOsc, audio, [ 220, 0.0 ] ]
[ 4_Rand, scalar, [ 0.5, 1.0 ] ]
[ 5_MulAdd, audio, [ 3_SinOsc, 4_Rand, 2_* ] ]
[ 6_SinOsc, audio, [ 660, 0.0 ] ]
[ 7_Rand, scalar, [ 0.5, 1.0 ] ]
[ 8_MulAdd, audio, [ 6_SinOsc, 7_Rand, 5_MulAdd ] ]
[ 9_SinOsc, audio, [ 880, 0.0 ] ]
[ 10_Rand, scalar, [ 0.5, 1.0 ] ]
[ 11_MulAdd, audio, [ 9_SinOsc, 10_Rand, 8_MulAdd ] ]
[ 12_SinOsc, audio, [ 1100, 0.0 ] ]
[ 13_Rand, scalar, [ 0.5, 1.0 ] ]
[ 14_MulAdd, audio, [ 12_SinOsc, 13_Rand, 11_MulAdd ] ]
[ 15_Out, audio, [ 0, 14_MulAdd ] ]

// failed case: all SinOscs clump at the top (width-first)!
(
SynthDef(\addTest, {
	var n = 5;
	var sig = SinOsc.ar(220 * (1..n));
	var amps = Array.fill(n, { Rand(0.5, 1.0) });
	Out.ar(0, (sig * amps).sum)
}).dumpUGens.add;
)

[ 0_SinOsc, audio, [ 220, 0.0 ] ]
[ 1_SinOsc, audio, [ 440, 0.0 ] ]
[ 2_SinOsc, audio, [ 660, 0.0 ] ]
[ 3_SinOsc, audio, [ 880, 0.0 ] ]
[ 4_SinOsc, audio, [ 1100, 0.0 ] ]
[ 5_Rand, scalar, [ 0.5, 1.0 ] ]
[ 6_Rand, scalar, [ 0.5, 1.0 ] ]
[ 7_*, audio, [ 1_SinOsc, 6_Rand ] ]
[ 8_MulAdd, audio, [ 0_SinOsc, 5_Rand, 7_* ] ]
[ 9_Rand, scalar, [ 0.5, 1.0 ] ]
[ 10_MulAdd, audio, [ 2_SinOsc, 9_Rand, 8_MulAdd ] ]
[ 11_Rand, scalar, [ 0.5, 1.0 ] ]
[ 12_MulAdd, audio, [ 3_SinOsc, 11_Rand, 10_MulAdd ] ]
[ 13_Rand, scalar, [ 0.5, 1.0 ] ]
[ 14_MulAdd, audio, [ 4_SinOsc, 13_Rand, 12_MulAdd ] ]
[ 15_Out, audio, [ 0, 14_MulAdd ] ]
```

To narrow this graph, the topo sort needs to know that SinOsc index 0 joins with Rand index 0. But there is no way, looking down from the top, to know this.

But if you work in reverse -- sort from the tail to the head -- then the MulAdd can pull in all of its ancestors right away, resulting in a narrower graph.

After the changes in this PR, the same failed case produces a properly interleaved UGen order:

```
[ 0_Rand, scalar, [ 0.5, 1.0 ] ]
[ 1_SinOsc, audio, [ 440, 0.0 ] ]
[ 2_*, audio, [ 1_SinOsc, 0_Rand ] ]
[ 3_Rand, scalar, [ 0.5, 1.0 ] ]
[ 4_SinOsc, audio, [ 220, 0.0 ] ]
[ 5_MulAdd, audio, [ 4_SinOsc, 3_Rand, 2_* ] ]
[ 6_Rand, scalar, [ 0.5, 1.0 ] ]
[ 7_SinOsc, audio, [ 660, 0.0 ] ]
[ 8_MulAdd, audio, [ 7_SinOsc, 6_Rand, 5_MulAdd ] ]
[ 9_Rand, scalar, [ 0.5, 1.0 ] ]
[ 10_SinOsc, audio, [ 880, 0.0 ] ]
[ 11_MulAdd, audio, [ 10_SinOsc, 9_Rand, 8_MulAdd ] ]
[ 12_Rand, scalar, [ 0.5, 1.0 ] ]
[ 13_SinOsc, audio, [ 1100, 0.0 ] ]
[ 14_MulAdd, audio, [ 13_SinOsc, 12_Rand, 11_MulAdd ] ]
[ 15_Out, audio, [ 0, 14_MulAdd ] ]
```

Note that this fixes a case from the forum: https://scsynth.org/t/maximum-value-for-numwirebufs/5887/7?u=jamshark70

```
SynthDef('wide!', {
	var n = 100;
	var a = SinOsc.ar(440 * (1..n));
	var b = LFSaw.kr({ ExpRand(1, 4) } ! n).range(0, 1);
	Out.ar(0, (a * b).sum)
}).add;

exception in GraphDef_Load: exceeded number of interconnect buffers.
```

No error with the patch!

## Types of changes

- Bug fix

I'm not sure if it should be called a breaking change or not. Users should notice nothing different because both topo sorts place all of a unit's descendants later than the ancestor unit(s). This change should never produce a wrong UGen order, though it may produce a different one from the unpatched SC version.

A risk here is that we don't have a comprehensive SynthDef test suite. That is too big a job for me to take on single-handed. I hope this PR will not be held up waiting for me to imagine every SynthDef permutation that requires an as-yet unwritten test.

## To-do list

- [x] Code is tested -- I checked a few of my synthdefs, no noticeable problems
- [x] All tests are passing
- [ ] Updated documentation -- needed? Maybe not. Topo sort logic isn't documented AFAIK.

### DIFF
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -508,7 +508,7 @@ SynthDef {
 			ugen.antecedents = ugen.antecedents.asArray.sort(
 								{ arg a, b; a.synthIndex < b.synthIndex }
 							);
-			ugen.makeAvailable; // all ugens with no antecedents are made available
+			ugen.makeAvailable; // all ugens with no descendants are made available
 		};
 	}
 	cleanupTopoSort {
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -505,7 +505,7 @@ SynthDef {
 			ugen.initTopoSort;
 		};
 		children.reverseDo { arg ugen;
-			ugen.descendants = ugen.descendants.asArray.sort(
+			ugen.antecedents = ugen.antecedents.asArray.sort(
 								{ arg a, b; a.synthIndex < b.synthIndex }
 							);
 			ugen.makeAvailable; // all ugens with no antecedents are made available
@@ -525,7 +525,7 @@ SynthDef {
 		{
 			outStack = available.pop.schedule(outStack);
 		};
-		children = outStack;
+		children = outStack.reverse;
 		this.cleanupTopoSort;
 	}
 	indexUGens {
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -504,7 +504,7 @@ SynthDef {
 			// this populates the descendants and antecedents
 			ugen.initTopoSort;
 		};
-		children.reverseDo { arg ugen;
+		children.do { arg ugen;
 			ugen.antecedents = ugen.antecedents.asArray.sort(
 								{ arg a, b; a.synthIndex < b.synthIndex }
 							);
--- a/SCClassLibrary/Common/Audio/UGen.sc
+++ b/SCClassLibrary/Common/Audio/UGen.sc
@@ -520,19 +520,19 @@ UGen : AbstractFunction {
 	}
 
 	makeAvailable {
-		if (antecedents.size == 0, {
+		if (descendants.size == 0, {
 			synthDef.available = synthDef.available.add(this);
 		});
 	}
 
-	removeAntecedent { arg ugen;
-		antecedents.remove(ugen);
+	removeDescendant { arg ugen;
+		descendants.remove(ugen);
 		this.makeAvailable;
 	}
 
 	schedule { arg outStack;
-		descendants.reverseDo({ arg ugen;
-			ugen.removeAntecedent(this);
+		antecedents.reverseDo({ arg ugen;
+			ugen.removeDescendant(this);
 		});
 		^outStack.add(this);
 	}


jamshark70:
Rohan, on the forum, pointed out that it's possible to use buffers or buses to make connections between parts of a single graph. Changes in the sorting algorithm could result in different behavior here.

In/Out -- at least in a simple case, no change.

```
(
SynthDef(\outSideEffect, {
	Out.ar(0, PinkNoise.ar() * 0.1);
	ReplaceOut.ar(0, LPF.ar(In.ar(0, 1), 440));
}).dumpUGens;
)

// dev branch
outSideEffect
[ 0_PinkNoise, audio, [  ] ]
[ 1_*, audio, [ 0_PinkNoise, 0.1 ] ]
[ 2_Out, audio, [ 0, 1_* ] ]
[ 3_In, audio, [ 0 ] ]
[ 4_LPF, audio, [ 3_In[0], 440 ] ]
[ 5_ReplaceOut, audio, [ 0, 4_LPF ] ]

// my branch = same
```

However, I'm pretty sure I've never seen anyone do this, because the more obvious way to feed a signal forward into another part of the graph is just to put it into a UGen input. It strikes me as perverse to introduce two additional UGens (with CPU overhead) only for this purpose.

Buffers: This actually was never safe -- the current algorithm may easily push BufWr far down, after BufRd, introducing an additional block delay. (The currently available solution is to force the writer into the reader's inputs, using `<!`.)

However this PR seems to fix that! So this case, at least, would be an argument in favor of the change.

```
(
SynthDef(\notOK, { |out, bufnum|
	var ph = Phasor.ar(0, 1, 0, BufFrames.kr(bufnum));
	var sig = SinOsc.ar(440);
	var writer = BufWr.ar(sig, bufnum, ph);
	var reader = BufRd.ar(1, bufnum, ph);
	Out.ar(out, reader);
}).dumpUGens;
)

current develop branch:
[ 0_Control, control, nil ]
[ 1_BufFrames, control, [ 0_Control[1] ] ]
[ 2_Phasor, audio, [ 0, 1, 0, 1_BufFrames, 0.0 ] ]
[ 3_BufRd, audio, [ 0_Control[1], 2_Phasor, 1.0, 2 ] ]
[ 4_Out, audio, [ 0_Control[0], 3_BufRd[0] ] ]
[ 5_SinOsc, audio, [ 440, 0.0 ] ]
[ 6_BufWr, audio, [ 0_Control[1], 2_Phasor, 1.0, 5_SinOsc ] ]

my PR branch, same SynthDef code:
[ 0_SinOsc, audio, [ 440, 0.0 ] ]
[ 1_Control, control, nil ]
[ 2_BufFrames, control, [ 1_Control[1] ] ]
[ 3_Phasor, audio, [ 0, 1, 0, 2_BufFrames, 0.0 ] ]
[ 4_BufWr, audio, [ 1_Control[1], 3_Phasor, 1.0, 0_SinOsc ] ]
[ 5_BufRd, audio, [ 1_Control[1], 3_Phasor, 1.0, 2 ] ]
[ 6_Out, audio, [ 1_Control[0], 5_BufRd[0] ] ]
```

jamshark70:
BTW I've been using this branch for a couple of weeks now (including all of the development of ddwWavetableSynth, where the SynthDefs are not exactly simple), and encountered no weird behavior. That's not an absolute guarantee but it suggests at least that any negative impact would be limited to obscure corner cases.

I hadn't checked PV units. The ordering is different in this case -- both are valid (all antecedents precede descendants), but I don't know how to evaluate which one is better.

```
(
d = SynthDef(\width, {
	var sig = PinkNoise.ar;
	var fft = FFT(LocalBuf(1024, 1), sig);
	var even = PV_EvenBin(fft);
	var odd = PV_OddBin(fft);
	Out.ar(0, IFFT([even, odd]))
});

d.dumpUGens;
)

// current develop

width
[ 0_PinkNoise, audio, [  ] ]
[ 1_MaxLocalBufs, scalar, [ 2 ] ]
[ 2_LocalBuf, scalar, [ 1, 1024, 1_MaxLocalBufs ] ]
[ 3_FFT, control, [ 2_LocalBuf, 0_PinkNoise, 0.5, 0, 1, 0 ] ]
[ 4_BufFrames, scalar, [ 2_LocalBuf ] ]
[ 5_LocalBuf, scalar, [ 1, 4_BufFrames, 1_MaxLocalBufs ] ]
[ 6_PV_Copy, control, [ 3_FFT, 5_LocalBuf ] ]
[ 7_PV_EvenBin, control, [ 6_PV_Copy ] ]
[ 8_PV_OddBin, control, [ 3_FFT ] ]
[ 9_IFFT, audio, [ 7_PV_EvenBin, 0, 0 ] ]
[ 10_IFFT, audio, [ 8_PV_OddBin, 0, 0 ] ]
[ 11_Out, audio, [ 0, 9_IFFT, 10_IFFT ] ]

// my branch

width
[ 0_MaxLocalBufs, scalar, [ 2 ] ]
[ 1_LocalBuf, scalar, [ 1, 1024, 0_MaxLocalBufs ] ]
[ 2_BufFrames, scalar, [ 1_LocalBuf ] ]
[ 3_LocalBuf, scalar, [ 1, 2_BufFrames, 0_MaxLocalBufs ] ]
[ 4_PinkNoise, audio, [  ] ]
[ 5_FFT, control, [ 1_LocalBuf, 4_PinkNoise, 0.5, 0, 1, 0 ] ]
[ 6_PV_Copy, control, [ 5_FFT, 3_LocalBuf ] ]
[ 7_PV_EvenBin, control, [ 6_PV_Copy ] ]
[ 8_PV_OddBin, control, [ 5_FFT ] ]
[ 9_IFFT, audio, [ 7_PV_EvenBin, 0, 0 ] ]
[ 10_IFFT, audio, [ 8_PV_OddBin, 0, 0 ] ]
[ 11_Out, audio, [ 0, 9_IFFT, 10_IFFT ] ]
```

jamshark70:
Oh btw it just occurred to me that, if we want to merge this, I will have to update `TestUGenMathOptimization` because it's sensitive to the specific order of units. Not hard to do, just noting it.

muellmusik:
> A risk here is that we don't have a comprehensive SynthDef test suite

Sorry to be a pain, but I really think we shouldn't merge this without a test suite in place that takes account of edge cases and side effects, with tests for PureUGens, WidthFirstUGens, and code elimination (anything else?) to ensure any regressions don't creep in. This is long overdue tbh.

I know from personal experience that this area can be super tricky and in general is poorly understood. IIRC the last time this area came up, a fairly cogent argument was put forth by some that with modern system specs the sorting optimisations are less crucial and may be not worth the complexity and risk.

Sorry to be a downer!

jamshark70:
> I really think we shouldn't merge this without a test suite in place that takes account of edge cases and side effects

The first step toward this would be a centralized location to collect "interesting" SynthDefs.

Forum thread? RFC? ... sccode.org, and ask contributors to add a specific tag?

What I'm driving at is that there is no single person who's going to sit down and try to make a comprehensive suite. Also, by definition, edge cases are exactly the ones that you *won't* think of when you're sitting down and trying to systematically cover all the territory. Making it a community effort might be one way out of "somebody oughta" limbo.

> WidthFirstUGens

I will admit here that I don't know what `widthFirstAntecedents` is doing in the topo sort. AFAICS, after a WidthFirstUGen gets added, every new UGen after that keeps a reference to it as its widthFirstAntecedents, and I guess this will cause WidthFirstUGens to be made available earlier than others...?

It would be really helpful to explain this somewhere.

I did post an example comparing the resulting order with FFT and PV_Copy -- was there any problem with that?

> code elimination

Sure, needs to be tested, but this occurs prior to the topological sort.

--

FWIW I do agree that changes to SynthDef building should not be merged in without understanding the impact. At the same time, the practical outcome *de facto* of that position is that the SynthDef builder will be off-limits, forever, and I'm not sure this is wonderful either.

One thing I can say for certain about the proposed alternate is that it categorically will not put an ancestor after a descendent. For input-output chains, it cannot produce a wrong ordering. Side-effect UGens (esp. BufWr / RecordBuf) are essentially indeterminate from this perspective -- one thing to note is that the current sort handles BufWr-then-BufRd incorrectly (noted above)!

One of my reasons for posting it as a PR is that interested parties could check out the branch and use it for a while, and see if any problems emerge. I had been doing that and drifted away from it (and I think I'll start up with that again). In that time, I never encountered any wrong behavior, in some moderately large defs. I fully see your point about hasty and naïve optimism about changes in this area, but of pessimism about it, I might suggest, try it for a month and prove that it's broken. We'll never know concretely if nobody ever tries it (which then, unfairly, reinforces the idea that "this is dangerous and nobody should ever touch it").

I would be very happy if someone would follow up on this PR with "hey, here's a weird case where the result is just totally wrong" because then that would be something concrete to look into.

> a fairly cogent argument was put forth by some that with modern system specs the sorting optimisations are less crucial

Sure. The particular concern here, though, is the complexity of SynthDef permitted for a given number of wirebufs, e.g., with the current sort:

- SynthDef function args for freqs and amps --> additive `osc * amp` --> sum: essentially unlimited.
- Define freqs, then amps as separate UGens --> same additive structure: wirebufs are a hard limit on the number of oscillators.

But with the alternate, both are essentially unlimited. The alternate is slightly less optimal if you Pan2 each oscillator before summing, but the difference in behavior between the two sorts in that case is very small (whereas the difference for the above two cases is large).

That + the BufWr case are concrete benefits. Those benefits must be measured against the risk (which we currently don't have a good way to measure), and that balance may very well go against the change, and I'm fine with that. But I'd like it at least to have a fair hearing.

muellmusik:
Hi James, Agree completely that a test suite should be a group effort! It's been a long time since I've waded into this code, and I think it's been modified repeatedly since. I'm also away from home with a repair backup computer and a bit of post-COVID fog, so I'm not sure I'm at my best for this sort of thing, but as a start we might try a specification for what should be achieved in SynthDef construction. AFAIK, this has never been explicitly documented (!) but here's a rough go:

1. Any UGens that have a side-effect (e.g. modifying a buffer, sending a message) need to occur in the chain _before_ any UGens that come later in the user-specified order, to ensure that side-effects occur in the right order.
2. Where possible, reorder depth first to allow wireBufs to be reused.
3. Where appropriate, use code-elimination strategies to remove unused UGens, unnecessary MulAdds etc.

NB 1. is _crucial_ and _top priority,_ while 2. and 3. are nice to haves. If unsure, we should implement with that in mind.

Is that spec complete and accurate? Is it worded _correctly and precisely_?

> I will admit here that I don't know what `widthFirstAntecedents` is doing in the topo sort. AFAICS, after a WidthFirstUGen gets added, every new UGen after that keeps a reference to it as its widthFirstAntecedents, and I guess this will cause WidthFirstUGens to be made available earlier than others...?

That sounds about right to me, though as I said it's been a long time for me and others have touched the code. I might more precisely say that `widthFirstAntecedents` is a list of antecedents that _must_ come before this UGen, but I'd need to check if that's how it really works these days.

Your PV_Copy example looks okay to me in terms of sort. A good approach (and Unit test) might be to check the output, e.g. by using a processed and unprocessed branch and comparing to input.

> categorically will not...

Famous last words! ;-) I think the approach should be not 'prove that it's broken', but 'prove that it works and that it still does after any proposed changes.'

So let's drop the 'we can't think of all edge cases' idea. It feels defeatist, and the number of categories of issue that could arise cannot be infinite. We can start by collecting problem SynthDefs from the archives. Focussing initially on discussions of this issue would be a sensible start.

muellmusik:
To get the ball rolling: Here are a few things to test that came up in past discussions over the years. Could be cleaned up and simplified.

Swap LocalBuf and Buffer (from PR #4050)
```
(
x.free;
b.free;
fork{
	b = Buffer.alloc(s, 32768);
	s.sync;
	x = {
		var in, chain;
		var trig, ff;
		in = WhiteNoise.ar(0.1);
		chain = FFT(LocalBuf(32768), in); //polling works
		// chain = FFT(b, in); //polling doesn't work
		chain.poll;
		trig = ( chain >=0 );
		trig.poll;
		IFFT(chain); // inverse FFT
	}.play;
};
)
```
____
Does RandSeed come before anything that uses it?
```
b = Buffer.new;
{|pitch=2,rnd=0.02,gain=0.7,rate=0.15,rvb=0.3,hif=11000,hig=4|
    var pointer, warp, fades, rates;
    RandSeed.ir(1,777);
    pointer = LFDNoise3.kr([0.02,0.03,0.04]*rate.lag(1)).range(0,1).poll;
    fades = Line.kr(0,1,[15,30,60]);
    rnd = rnd.lag(1);
warp = (0.4*fades[0]*Warp1.ar(2, b, pointer[0], pitch, 0.2, 
windowRandRatio: rnd, interp: 2)) + (1.0*fades[1]*Warp1.ar(2, b, 
pointer[1], pitch*0.5, 0.1, windowRandRatio: rnd, interp: 2)).reverse + 
(0.6*fades[2]*Warp1.ar(2, b, pointer[2], pitch*0.25, 0.3, 
windowRandRatio: rnd, interp: 2))
         ;
    warp = MidEQ.ar(warp,hif,0.6,hig);
    tanh(gain*HPF.ar(FreeVerb.ar(warp,rvb,0.999,0.1),30));
}.asSynthDef.dumpUGens
```
____
Do the FFTs and IFFTs occur in correct order? Could also test output.
```
b = Buffer.alloc(s, 2048);
(
SynthDef("shareBuf", { arg out=0;
	var in, chain;
	in = WhiteNoise.ar(0.1);
	chain = FFT(b, in);
	Out.ar(out,
		IFFT(chain) // inverse FFT
	);



	in = Saw.ar(440, 0.1);
	chain = FFT(b, in);
	Out.ar(out + 1,
		IFFT(chain) // inverse FFT
	);
}).dumpUGens
)
```
___

I can find some other ones once I'm back home and on my repaired machine.

jamshark70:
> > categorically will not...
> 
> Famous last words! ;-)

- Current sort: UGens are added top to bottom (first to execute, first to be added). A UGen is available to be "scheduled" into the graph once all of its inputs have been scheduled. "Available" = UGens are added into the sort order only ever from an array named `available`, never from any other source: `outStack = available.pop.schedule(outStack);`, and UGens are put into this array based on the rule of inputs having been scheduled already. So the algorithm makes it literally impossible for any UGen to be added before its ancestor.
- Alternate sort: UGens are added bottom to top (first to execute, last to be added). A UGen is available to be scheduled into the graph once all of its output targets have been scheduled. By the same logic, it's impossible for any UGen to be sorted later than its descendant.

If there is a loophole here, then the same loophole would exist in JMc's original algorithm.

> - Any UGens that have a side-effect (e.g. modifying a buffer, sending a message) need to occur in the chain before any UGens that come later in the user-specified order, to ensure that side-effects occur in the right order.
> - Where possible, reorder depth first to allow wireBufs to be reused.
> - Where appropriate, use code-elimination strategies to remove unused UGens, unnecessary MulAdds etc.

These sound right to me, although the third takes place prior to sorting -- the sort is not responsible for optimization (and it would not make sense to sort, optimize, and then re-sort). Optimization needs to be part of the test suite, but this PR is about sorting.

"need to occur in the chain before any UGens that come later in the user-specified order" -- the current sort already violates this in at least one known case (which JMc decided to work around by introducing the `<!` operator).

"Where possible, reorder depth first" -- this also does not happen in at least one known case (not even an exotic case), where it is possible to go depth first, but, working from the top down, hard to determine that it's possible.

The 3 test cases are good ones -- I was thinking about RandSeed before reading this.

> Swap LocalBuf and Buffer (from PR #4050)

I tried it with both the original sort and this PR -- both produce the same sort order.

> Does RandSeed come before anything that uses it?

Yes, in both: in the original, it's UGen index 2; in mine, index 0! (The way it works in the inverted case is that the Out at the bottom greedily pulls in the entire tree above it, before it can move on to the next available thing, which is RandSeed.)

In both versions, you can mess up the order by writing RandSeed at the end of the synthdef (but this is intuitive).

> Do the FFTs and IFFTs occur in correct order?

Almost identical sort order, but the FFT/IFFT pairs are not broken. My alternate still works because the lower Out pulls in its whole chain, before the upper Out can process at all.

jamshark70:
> Does RandSeed come before anything that uses it?

Here, I did find a breaking case:

```
(
SynthDef(\randseedX2, { |out|
	var sig = Array.fill(2, { |i|
		RandSeed.ir(1, (i + 1) * 500);
		LFDNoise3.ar(500);
	}).sum;
	Out.ar(out, sig);
}).dumpUGens;
)```

Original way:

[ 0_Control, control, nil ]
[ 1_RandSeed, scalar, [ 1, 500 ] ]
[ 2_LFDNoise3, audio, [ 500 ] ]
[ 3_RandSeed, scalar, [ 1, 1000 ] ]  -- YES!
[ 4_LFDNoise3, audio, [ 500 ] ]
[ 5_+, audio, [ 2_LFDNoise3, 4_LFDNoise3 ] ]
[ 6_Out, audio, [ 0_Control[0], 5_+ ] ]

Inverted:

[ 0_RandSeed, scalar, [ 1, 500 ] ]
[ 1_RandSeed, scalar, [ 1, 1000 ] ]   -- NO!
[ 2_LFDNoise3, audio, [ 500 ] ]
[ 3_LFDNoise3, audio, [ 500 ] ]
[ 4_+, audio, [ 3_LFDNoise3, 2_LFDNoise3 ] ]
[ 5_Control, control, nil ]
[ 6_Out, audio, [ 5_Control[0], 4_+ ] ]

AFAICS, though, the correct result that we have currently is largely an accident -- widthFirstAntecedents are not used at all in the sorting logic. "I might more precisely say that widthFirstAntecedents is a list of antecedents that must come before this UGen" but I see only that this array is populated, not that it is ever used to guarantee precedence.

E.g., using the original sort, not mine! If you decide to be clever and try to force pre-created RandSeeds into the graph using `<!`, then it's wrong. It *only* works if RandSeeds are interleaved at creation time. I think inverting the sort could fix that, actually.

```
(
SynthDef(\randseedX2, { |out|
	var seeds = Array.fill(2, { |i|
		RandSeed.ir(1, (i + 1) * 500);
	});
	var sig = Array.fill(2, { |i|
		LFDNoise3.ar(DC.ar(500) <! seeds[i]);
	}).sum;
	Out.ar(out, sig);
}).dumpUGens;
)

[ 0_Control, control, nil ]
[ 1_RandSeed, scalar, [ 1, 500 ] ]
[ 2_RandSeed, scalar, [ 1, 1000 ] ]
[ 3_DC, audio, [ 500 ] ]
[ 4_firstArg, audio, [ 3_DC[0], 0.0 ] ]
[ 5_LFDNoise3, audio, [ 4_firstArg ] ]
[ 6_DC, audio, [ 500 ] ]
[ 7_firstArg, audio, [ 6_DC[0], 0.0 ] ]
[ 8_LFDNoise3, audio, [ 7_firstArg ] ]
[ 9_+, audio, [ 5_LFDNoise3, 8_LFDNoise3 ] ]
[ 10_Out, audio, [ 0_Control[0], 9_+ ] ]
```

I suspect widthFirstAntecedents would be the solution to the initial problem, though. Using the "inverted" dump, if 3_LFDNoise3 had 1_RandSeed as a widthFirstAntecedent, then adding 3_ could pull in 1_ immediately, and similarly for 2_ and 0_. This would mean that the RandSeeds would have to be canceled from the initial `available` pass, which I won't have time to work out this morning, but it's certainly a solvable problem.

jamshark70:
Sorry, initially I missed a closing triple-backtick and that bodged the formatting. Fixed online now, but not in your email inbox...

muellmusik:
James, I think we're getting immersed in implementation details here, when what we really need to do is go back to the start and:

1. Come up with a clear specification of what the result of SynthDef construction (including sort, elimination, ...?) _should_ be, irrespective of implementation, i.e. a requirements list
2. Come up with a comprehensive set of tests to capture that.

We've only just started on 1. and can't really complete 2. until we have that confirmed. Once we have that we can more easily assess this PR or any other changes. 

Would it make sense to move this discussion to an RFC and have comments? Given how sensitive this is, I think this should be be publicised as widely as possible.

muellmusik:
> 2\. Come up with a comprehensive set of tests to capture that.

BTW, I think – again given the centrality of this – a sensible precaution would be to include a test for every issue in SynthDef construction that's ever been raised. That way we ensure no regression.

jamshark70:
Just reporting that I'm beginning to prepare an RFC. It will take some time to do it right. No rush on any of this, in any case.

capital-G:
@JordanHendersonMusic do you have some insights how this PR relates to https://github.com/supercollider/supercollider/pull/6405

JordanHendersonMusic:
@capital-G #6405 makes this issue (and all other regarding sorting/optimisations) irrelevant/outdated. There will be some edge cases where the number of wirebufs could be reduced, but they are far fewer in number that the current system — they only arise when you have unusual orderings of stateful UGens inside multiple long chains. 

IMO, changing the sorting mechanism to fix a incorrectly constructed graph is always going to fail and get messy quickly.

Also, regarding @muellmusik's comment
> Come up with a clear specification of what the result of SynthDef construction (including sort, elimination, ...?) should be, irrespective of implementation, i.e. a requirements list
Come up with a comprehensive set of tests to capture that.

While I haven't written a formal document on what a valid synthdef compiler does, my approach in the testing has been to ensure a null test with the unoptimised version. I'm not really in favour of specifying (in writing or in a test) *what* the compiler does, only *that* it works by passing the null test.

jamshark70:
This PR is out of date, no further discussion needed.

PR #5814, opened 2022-07-15
class library: use makeFlopFunc for function flop

telephon:
## Purpose and Motivation

A function that has been returned from function flop does not work with keyword and ellipsis arguments. This change uses the existing `makeFlopFunction` and makes this work. Downside is that we have to call `interpret` for this.

For discussion, see: #5499

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


telephon:
wrong duplicate of #5499.

PR #5787, opened 2022-05-25
Fix UGen initialization: OscUGens

mtmccrea:
## Purpose and Motivation

A majority of UGens do not initialize with a correct initialization sample, and subsequently output an incorrect first sample. A lack of clarity and consistency around how UGens are initialized has led to many UGens being improperly initialized. The affects include: UGen graphs aren't properly "primed" (users see unexpected results), samples are delayed, filters can pop from incorrect feedback coefficients, accessing unassigned memory, etc. Users regularly report Issues that can be traced back to UGen initialization problems. 

This PR also serves as a concrete reference for discussion found in the [RFC: Fix UGen Initializations](https://github.com/supercollider/rfcs/pull/19), more details can be found there.

(I will update this post with Issues this fixes.)

## Types of changes

- Documentation (`FSinOsc`, `FoldIndex`)
- Bug fixes
- Breaking changes

Changes in behavior of UGens would all technically be breaking. However, the changes are largely (entirely?) confined to the output of samples within the first block.

I chose to use `OscUGens` as the starting point because I expect these changes are minor (likely won't be heard), and are hopefully not overly controversial. 

Changes to UGens which are more elaborate or might create more substantial change in behavior will be left out of these kind of "batch" fixes, and submitted as separate PRs (`TWindex`(#5815), `Klang` and `Klank` (#5817), in this case).

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #5815, opened 2022-07-16
TWindex: Ensure the random first sample matches init sample

mtmccrea:
## Purpose and Motivation

Previously, `TWindex` would _always_ output index `0` if no trigger was received on the first sample, regardless of the probability assigned to that index by the user. 

Specifically, the line:
```
unit->m_trig = -1.f; // make it trigger the first time
```
doesn’t actually ensure it triggers the first time; the trigger input may not fire, in which case it defaults to `unit->m_prevIndex = 0;` (which might violate the assigned probability).

This PR fixes that, as well as ensuring that the initialization sample matches the first sample (by preventing the generation of a new random value on a first-sample trigger).

Another change was made so that the `maxSum` variable is only updated internally on a received trigger , it's not needed otherwise. It's also checked so that it's updated no more than once per control period because the probability array input is only updated at control rate.

Part of https://github.com/supercollider/supercollider/pull/5787, and the [init sample RFC](https://github.com/supercollider/rfcs/pull/19).

## Types of changes

[x] Bug fix
[x] Optimization

## To-do list

[x] Code is tested
[x] All tests are passing
[-] Updated documentation
[x] This PR is ready for review

### DIFF
--- a/server/plugins/OscUGens.cpp
+++ b/server/plugins/OscUGens.cpp
@@ -47,6 +47,7 @@ struct Select : public Unit {};
 struct TWindex : public Unit {
     int32 m_prevIndex;
     float m_trig;
+    float m_maxSum;
 };
 
 struct Index : public BufUnit {};
@@ -527,26 +528,26 @@ void Select_next_a(Select* unit, int inNumSamples) {
 }
 ////////////////////////////////////////////////////////////////////////////////////
 
-void TWindex_Ctor(TWindex* unit) {
-    if (INRATE(0) == calc_FullRate) {
-        SETCALC(TWindex_next_a);
-    } else {
-        SETCALC(TWindex_next_k);
-    }
-    // Generate a random value here for the initialization & first sample
+static int32 TWindex_chooseNewIndex(TWindex* unit) {
     int maxindex = unit->mNumInputs;
     int32 index = maxindex;
-    float sum = 0.f;
-    float maxSum = 0.f;
     float normalize = ZIN0(1);
-    if (normalize == 1) {
-        for (int32 k = 2; k < maxindex; ++k) {
-            maxSum += ZIN0(k);
-        }
-    } else
-        maxSum = 1.f;
+    float maxSum = unit->m_maxSum;
+
+    if (maxSum < 0.f) {
+        maxSum = 0.f;
+        if (normalize == 1) {
+            for (int32 k = 2; k < maxindex; ++k) {
+                maxSum += ZIN0(k);
+            }
+        } else
+            maxSum = 1.f;
+        unit->m_maxSum = maxSum;
+    }
     RGen& rgen = *unit->mParent->mRGen;
+
     float max = maxSum * rgen.frand();
+    float sum = 0.f;
     for (int32 k = 2; k < maxindex; ++k) {
         sum += ZIN0(k);
         if (sum >= max) {
@@ -554,6 +555,17 @@ void TWindex_Ctor(TWindex* unit) {
             break;
         }
     }
+    return index;
+}
+
+void TWindex_Ctor(TWindex* unit) {
+    if (INRATE(0) == calc_FullRate) {
+        SETCALC(TWindex_next_a);
+    } else {
+        SETCALC(TWindex_next_k);
+    }
+    unit->m_maxSum = -1.f; // trigger update of maxSum
+    int32 index = TWindex_chooseNewIndex(unit);
     OUT0(0) = index;
     unit->m_prevIndex = index;
     // Ensure a first-sample trigger doesn't cause another rand val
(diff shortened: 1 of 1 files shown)


server/plugins/OscUGens.cpp
```diff
@@ -533,12 +533,34 @@ void TWindex_Ctor(TWindex* unit) {
     } else {
         SETCALC(TWindex_next_k);
     }
-    unit->m_prevIndex = 0;
-    unit->m_trig = -1.f; // make it trigger the first time
-    TWindex_next_k(unit, 1);
+    // Generate a random value here for the initialization & first sample
```
telephon, line 536: would it perhaps be easier to make sure that the condition `trig > 0.f && unit->m_trig <= 0.f)` is fulfilled, then call `TWindex_next_k` and then set `trig = 0` again? Then you don't have to duplicate all the code. Usually it is done like this in other cases.

mtmccrea: > would it perhaps be easier to make sure that the condition trig > 0.f && unit->m_trig <= 0.f) is fulfilled

We can control the condition `unit->m_trig <= 0.f`, but not ` trig > 0.f` because `trig` is read from the input. The solution I have here is required, from what I can tell, by the case where there is not a trigger on the first input frame.

So I get around this by at least making this test _false_ on the first frame by use of `unit->m_trig = 1.f`, so I can control the value of both the initialization and first sample (so a new random number isn't generated). I agree duplicating code isn't pretty :-/

> Usually it is done like this in other cases.
The unique thing here is that randomness is used in the calc function, whereas many other UGens don't involve randomness.

I'll put a bit more thought into it just in case...

telephon: yes, it would be great to have a solution that works without so much code duplication. If there is none, well then that is also a good result – but we may have to check other UGens that need randomness in the calc function. How is it with `TRand`?

mtmccrea: `TRand` also sets the init sample directly in the Ctor without calling a `next_` function:
https://github.com/supercollider/supercollider/blob/0fe26b63d3c10add37fbc270789535aab3737b1a/server/plugins/NoiseUGens.cpp#L512-L514
In that case there isn't much overhead in terms of using that randomly generated number (just scaling it between `hi` and `lo`)—it _is_ the output value. (At a quick glance this UGen will also need to be updated to fix the init sample bug.)

Whereas most of the duplicated code is how the generated number is used: normalizing and scaling it then using it in the loop that chooses the index to output.

An alternative would be to make a helper function that does all of this, so it could be factored out of the Ctor and both calc functions. But this would be a departure from UGen style generally, and I'm not sure what kind of overhead that function might add to the calc function.

Having a look though _NoiseUGens.cpp_, many of the UGens that use `mRGen` do so in the Ctor and set the output sample directly.

mtmccrea: Here's a version using a helper function `TWindex_chooseNewIndex`:
```
static int32 TWindex_chooseNewIndex(TWindex* unit) {
    int maxindex = unit->mNumInputs;
    int32 index = maxindex;
    float normalize = ZIN0(1);
    float maxSum = unit->m_maxSum;
    
    if (maxSum < 0.f) {
        maxSum = 0.f;
        if (normalize == 1) {
            for (int32 k = 2; k < maxindex; ++k) {
                maxSum += ZIN0(k);
            }
        } else
            maxSum = 1.f;
        unit->m_maxSum = maxSum;
    }
    RGen& rgen = *unit->mParent->mRGen;
    
    float max = maxSum * rgen.frand();
    float sum = 0.f;
    for (int32 k = 2; k < maxindex; ++k) {
        sum += ZIN0(k);
        if (sum >= max) {
            index = k - 2;
            break;
        }
    }
    return index;
}

void TWindex_Ctor(TWindex* unit) {
    if (INRATE(0) == calc_FullRate) {
        SETCALC(TWindex_next_a);
    } else {
        SETCALC(TWindex_next_k);
    }
    unit->m_maxSum = -1.f; // trigger update of maxSum
    int32 index = TWindex_chooseNewIndex(unit);
    OUT0(0) = index;
    unit->m_prevIndex = index;
    // Ensure a first-sample trigger doesn't cause another rand val
    // so initialization sample will equal first output sample
    unit->m_trig = 1.f;
}

void TWindex_next_k(TWindex* unit, int inNumSamples) {
    float trig = ZIN0(0);
    float* out = ZOUT(0);
    int32 index;
    unit->m_maxSum = -1.f;
    if (trig > 0.f && unit->m_trig <= 0.f) {
        index = TWindex_chooseNewIndex(unit);
        unit->m_prevIndex = index;
    } else {
        index = unit->m_prevIndex;
    }
    LOOP1(inNumSamples, ZXP(out) = index;)
    unit->m_trig = trig;
}

void TWindex_next_a(TWindex* unit, int inNumSamples) {
    float* trig = ZIN(0);
    float* out = ZOUT(0);
    int32 index;
    float curtrig;
    unit->m_maxSum = -1.f;

    LOOP1(
        inNumSamples, curtrig = ZXP(trig); if (curtrig > 0.f && unit->m_trig <= 0.f) {
            index = TWindex_chooseNewIndex(unit);
            unit->m_prevIndex = index;
        } else index = unit->m_prevIndex;

        ZXP(out) = index; unit->m_trig = curtrig;)
}
```
It's a more substantial overhaul. I could push this change for discussion in the Review, though if we revert, it may involve a force-push, and I'm not sure what that does to the Review history. I could also revert without force-pushing, and the commits could be squashed on merge.

If you could let me know the etiquette for force-push in this situation and that's the preferred way, I'll remove this post and we could continue the review that way.

Thanks for following up with this, @telephon!

joshpar (approved):

PR #5817, opened 2022-07-17
Klang, Klank: Fix initialization

mtmccrea:
## Purpose and Motivation

Part fo the [initialization sample fix project](https://github.com/supercollider/rfcs/pull/19), specifically in the [OscUGens subset](https://github.com/supercollider/supercollider/pull/5787).

## Types of changes

- Bug fix

Changes to `Klang`:
- correct the initial phase (it was off by 1 sample).

Change to `Klank`:
- Don’t assume init sample is zero—reorganize coeff calc func and ctor to get correct init sample.
- The memory allocation for `m_coefs` was moved to the Ctor, otherwise, the `return` statement escapes the `Klank_SetCoefs`, but not the Ctor, so `m_coefs` might still be accessed after allocation fails.
- Calculate the correct init sample by calling `Klank_next(unit, 1)`, temporarily setting `mFilterLoops` and  `mFilterRemain` for the single cycle of initialization. 
  - Usually the pattern of calling `_next(unit, 1)` works, but `Klank_next` doesn’t actually use the `inNumSamples` argument, because it uses the looping scheme designed to optimize second-order filter calculations. As a consequence, a full block of calculations are run over which memory assigned for input data is accessed before it’s been written, so it’s running unnecessary calculations with garbage memory.
  - Here’s a small dump from that internal loop:

```
[Klank_next 133] 0.052083
[Klank_next 144] 0.103113
[Klank_next 155] 5818479576238205845020707127296.000000
[Klank_next 133] 11667158119450385126825574531072.000000
[Klank_next 144] 17284522609726230684285924278272.000000
...
```
- The last commit removes commented Print statements. I’m not sure what the etiquette is for touching old code for trivial changes, so I can revert this commit if needed.

## To-do list

- [x] Code is tested
- [x] All tests are passing (`TestUGen_RTAlloc`)
- [-] Updated documentation

### DIFF
--- a/server/plugins/OscUGens.cpp
+++ b/server/plugins/OscUGens.cpp
@@ -2895,21 +2895,13 @@ void Klang_next(Klang* unit, int inNumSamples) {
 ////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 static void Klank_SetCoefs(Klank* unit) {
-    int numpartials = (unit->mNumInputs - 4) / 3;
-    unit->m_numpartials = numpartials;
-
-    int numcoefs = ((unit->m_numpartials + 3) & ~3) * 5;
-    unit->m_coefs = (float*)RTAlloc(unit->mWorld, (numcoefs + unit->mWorld->mBufLength) * sizeof(float));
-    ClearUnitIfMemFailed(unit->m_coefs);
-
-    unit->m_buf = unit->m_coefs + numcoefs;
-
     float freqscale = ZIN0(1) * unit->mRate->mRadiansPerSample;
     float freqoffset = ZIN0(2) * unit->mRate->mRadiansPerSample;
     float decayscale = ZIN0(3);
 
     float* coefs = unit->m_coefs;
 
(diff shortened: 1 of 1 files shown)


telephon:
>Don’t assume init sample is zero—reorganize coeff calc func and ctor to get correct init sample.

just to make sure: if you write
```supercollider
{ Klank.ar(`[[1000], [1], [1]], Impulse.ar(0)) }.play
```
The Klank does interpret the input as a trigger, right?

jamshark70:
@telephon --

> if you write `{ Klank.ar(`[[1000], [1], [1]], Impulse.ar(0)) }.play`, the Klank does interpret the input as a trigger, right?

klank-impulse

mtmccrea:
Yes, just to show the triggered behavior a bit more clearly, here is
```{ Klank.ar(`[[1000], [1], [0.05]], Impulse.ar(0)) }.plot(0.1)```
Screen Shot 2022-08-12 at 10 03 01

joshpar (approved):

PR #5820, opened 2022-07-22
supernova: use the `/error` messages to turn on / off the console printing

vitreo12:
## Purpose and Motivation

Allow supernova to use the `/error` message just like scsynth. This was previously being ignored. As discussed on the [forum](https://scsynth.org/t/error-message-not-working-for-supernova/6275/4), it is a feature that is needed to match scsynth's capabilities. 

This patch does not take into account the local bundles (`/error, -1` and `/error, -2`), as supernova appears to treat the OSC bundles a little differently than scsynth, and I am still studying that part of the code base. This will come as a secondary patch in the future

## Types of changes

- Bug fix

## To-do list

- [X] Code is tested
- [ ] All tests are passing
- [X] Updated documentation

### DIFF
--- a/server/supernova/sc/sc_osc_handler.hpp
+++ b/server/supernova/sc/sc_osc_handler.hpp
@@ -277,6 +277,8 @@ public:
 
     void set_error_posting(int val) { error_posting = val; }
 
+    int get_error_posting() { return error_posting; }
+
 private:
     int error_posting = 1;
     /* @} */
--- a/server/supernova/server/server.hpp
+++ b/server/supernova/server/server.hpp
@@ -246,8 +246,7 @@ inline void run_scheduler_tick(void) {
     }
 }
 
-inline bool log_printf(const char* fmt, ...) 
-{
+inline bool log_printf(const char* fmt, ...) {
     if (instance->get_error_posting()) {
         va_list vargs;
         va_start(vargs, fmt);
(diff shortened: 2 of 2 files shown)


dyfer:
Thanks so much for the fix, @vitreo12 
What would you think about leaving a TODO or FIXME comment somewhere in the code indicating the need to implement the local error suppression (`-1`/`-2`) in the future? It's just something to consider, not a strong request.

vitreo12:
I thought about it, but then I saw that there already is a comment about it a couple of lines above:

https://github.com/supercollider/supercollider/blob/0aa48adbddc0ad50420f4ff99b9429858d342609/server/supernova/sc/sc_osc_handler.hpp#L276-L279

dyfer:
> Perhaps it could be more specific, I could surely modify it to make it more clear

Ah, I see. We can leave it as is - unless you feel like making the message more specific.

vitreo12:
I agree, we can just leave it as is.

dyfer (approved):
Thanks @vitreo12 
I tested it and it works as expected.

Spacechild1:
@vitreo12 @dyfer I don't think this is right. The `/error` message is supposed to disable *error* messages, but now it disables *all* console printout.

I think `error_posting` should be checked in the OSC method handlers instead, as it is done in `scsynth`.

As a side noe, the `/error` command also takes negative values to disable error messages locally. In Supernova, this is currently not implemented at all. See https://github.com/supercollider/supercollider/blob/b9a57003b34ebdbb45070abfc10d90a2acf1588f/server/scsynth/SC_MiscCmds.cpp#L1837-L1861.

PR #5826, opened 2022-07-29
[skip ci] Help: Clarify full sysex format for all OSes

jamshark70:
## Purpose and Motivation

MIDIOut sysex help is confusing: "On Windows, the function call must contain a full sysex message..."

That implies that it may be different on other OSes.

I tested in Linux, and it's the same. And a [helpful person from the forum](https://scsynth.org/t/midi-sysex-out-platform-compatibility/6312/4?u=jamshark70) repeated the test in May, and also the same.

So it should really just advise the user to include the delimiters.

## Types of changes

- Documentation

## To-do list

- [n/a] Code is tested
- [n/a] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/MIDIOut.schelp
+++ b/HelpSource/Classes/MIDIOut.schelp
@@ -52,12 +52,20 @@ method::sysex
 Sends a sysex command represented as an link::Classes/Int8Array:: to the device.
 
 note::
-On Windows, the function call must contain a full sysex message. In other words,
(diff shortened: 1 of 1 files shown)


dyfer (approved):

dyfer:
Thanks @jamshark70 !
FYI changes to help files only [don't trigger the CI](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/actions.yml#L12) so there's no need to manually add `skip ci` in that particular case. This was not documented on the wiki (sorry!), but I've added a note about it.

PR #5161, opened 2020-09-14
GridLines improvements: fix exponential grids and add spacing control

dyfer:
## Purpose and Motivation

This PR fixes exponential GridLines, which previously were not drawn cleanly. Fixes #5156

In detail this PR:
- fixes drawing exponential gridlines, based on solution by @sternsc 
  - both positive and negative ranges are supported
- adds control over grid density through ~`DrawGrid -pixDensity`~ `DrawGridX -tickSpacing`
- ~~adds option to not draw individual labels if they'd overlap, through `DrawGrid -dropLabelIfTooNarrow `~~
- labels are now dropped within the exponential gridlines generation for the X axis, based on line spacing and label width
- ~~removes `.0` from integer labels~~ this was addressed in #5827
- I realized that the original design [intended to create separate GridLines objects for various Warps](https://github.com/supercollider/supercollider/blob/develop/HelpSource/Classes/GridLines.schelp#L13). This PR now implements exponential grids as a `ExponentialGridLines` class, instead of [using switch statements when generating grids](https://github.com/dyfer/supercollider/commit/4b2ff1d4a0af6e35a9ad109b2573728c32bff730#diff-1d5b6087b26a5651c9a746f81acf4e906a45ed083d3e95b12322ab9442979426R367).

Other considerations
- ~~the mechanism to control grid spacing and density is implemented mostly in `GridLines -getParam`, but some of it is in `DrawGrid`; I have not changed this but this design could be reconsidered~~
- There are some minor drive-by formatting changes to the code; they were in the vicinity of the code I was changing but I'd be happy to remove those if they're not desired
- There is quite a bit of magic numbers that I came up with when tuning the grids
- ~~The documentation still needs updating~~ Basic documentation updates are now included in this PR

Examples:

```supercollider
(
c = ControlSpec.new(20,20000, \exp, units:"Hz");
g = c.grid;
d = DrawGrid.test(g, nil, 500@100);
)
```
Screen Shot 2022-09-07 at 11 43 59

```supercollider
(
c = ControlSpec.new(20,20000, \exp, units:"Hz");
g = c.grid;
d = DrawGrid.test(g, nil, 1000@100);
)
```
Screen Shot 2022-09-07 at 11 44 13

```supercollider
(
c = ControlSpec.new(0.1, 100, \exp, units:"Hz");
g = c.asSpec.grid;
d = DrawGrid.test(g, nil, 500@100);
)
```
Screen Shot 2022-09-07 at 12 00 40

```supercollider
[...]

PR #4511, opened 2019-07-29
Plotter: update default colors to valid colors.

mtmccrea:
## Purpose and Motivation

Revealed in fa67b746fa6cf9cd29eb42d4fba3f63133e3bf8d, `Plotter` never really handled default colors correctly. When accessing `QtGUI.palettle` colors, it would return unset colors, which would "gracefully" fail by defaulting to black. It seems that correcting the behavior of `-asHSV` made these calls break...ungracefully.

The problem is seen by running 
```supercollider
7.collect({ 15.collect({ rrand(0,1.0) }) }).plot.superpose_(true)
```
The proposed defaults access `QtGUI.palette.highlight`, which returns a valid color, whereas `QtGUI.palette.windowText`, used before, didn't.

I've also reduced the default `plotColor` down to one color. Four colors was a rather random number of colors to default to, and in any case the user can set them explicitly in `Plot` (more easily through `Plotter` in a pending PR merge — #4459).

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
~~- [ ] Updated documentation~~
- [x] This PR is ready for review

PR #5827, opened 2022-07-29
Plotter: fix grid and axis labels

mtmccrea:
## Purpose and Motivation

This adds full support for grid labels and axis labels for `Plotter`.

### Changes
- build label margins for each axis independently
- refresh when updating plot bounds
- account for x-AND y-labels when sizing left and right margins
- always draw the outer edge grid lines
- xlabels: handle edge case of very small width
- distinguish between x and y grid edges when getting label size
- rename x/yLabels -> x/ytkLabels - they're tick labels not axis labels
- fully support axis labels: add -labelX, -labelY methods, support multichannel plot labels
- docs: cleanup, fix examples, reorganize by subcategory
- don't display `.0` if all grid labels are whole numbers
- Env:plot: don't assume units are seconds.
- Buffer:plot: display units in x-axis label, not grid labels.
- Buffer:plot: match argument order to default/auto-complete
- don't call refresh again after `setProperties`
- add padding to x tick label inset, properly center y label
- ensure there's edge lines when only one grid line
- update axis label if axis specs contain units

  [x] Documentation
  [x] Bug fix
  [x] New feature

Some examples of the changes:
Note axis labels, grid labels, labels disappearing under height/width size limits, units listed as axis labels instead of grid labels, grid lines appearing at max/min plot bounds, etc.
aggregate imgs

## To-do list

- [x] Code is tested
- [x] Tests are passing (`TestPlotter`)
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Plotter.schelp
+++ b/HelpSource/Classes/Plotter.schelp
@@ -87,6 +87,8 @@ subsection:: Embedding in another view
 method:: makeWindow
 Open given plotter in a new window or within a given composite view.
 
+See the link::#Embedding in another View or GUI:: example.
+
 argument:: argParent
 Either a link::Classes/Window:: or link::Classes/View:: may be passed in - then the plot is embedded. Otherwise a new link::Classes/Window:: is created.
 
@@ -159,7 +161,7 @@ color specified or if all colors of a multi-channel plot are the same, in which
 case a link::Classes/Color:: is returned.
 
 
-method::superpose
+method:: superpose
 If code::true::, plotter displays channels overlaid on a single plot. (keyboard shortcut: s)
 code::
 a = { (0..30).scramble }.dup(2).plot;
@@ -167,9 +169,24 @@ a.superpose = true; a.refresh;
 ::
 
 
+method:: labelX
+Get/set the label for the x-axes. Can be a link::Classes/String:: or an link::Classes/Array:: of
+link::Classes/String::s, or code::nil:: to remove the label.
+See link::#Axis labels:: example below.
+
+
+method:: labelY
+Get/set the label for the y-axes. Can be a link::Classes/String:: or an link::Classes/Array:: of
+link::Classes/String::s, or code::nil:: to remove the label.
+See link::#Axis labels:: example below.
+
 
 method:: setProperties
 Set properties of all plot views. Defaults are taken from code::GUI.skins.at(\plot);::
+
+warning:: It's preferrable to use Plotter's instance methods when possible
+to set Plot properties to ensure proper behavior when using link::#-superpose::.
+::
 
 argument:: ... pairs
 A list of symbol,value pairs. Supported properties:
@@ -217,26 +234,29 @@ subsection:: Data display properties
 
 method:: specs
 Set or get the link::Classes/ControlSpec##spec:: for the y-axis (codomain).
-note::
-The y-axis label will update with the strong::units:: assigned
-to your spec, if it has not been previously set.
-::
+
 code::
 a = { (40..3000).scramble }.dup(2).plot;
 a.specs = \freq.asSpec; a.refresh;
 ::
-See also link::#examples::.
+See also the link::#Explicit domain and axis specs:: example below.
+note::
+The y-axis label will update with the strong::units:: assigned
+to your spec, if it has not been previously set.
+::
 
 
 method:: domainSpecs
 Set or get the link::Classes/ControlSpec##spec:: for the x-axis (domain).
-note::
-The x-axis label will update with the strong::units:: assigned
-to your spec, if it has not been previously set.
-::
+
 code::
 a = { (40..300).scramble }.dup(2).plot;
 a.domainSpecs = \freq.asSpec; a.refresh;
+::
+See also the link::#Explicit domain and axis specs:: example below.
+note::
+The x-axis label will update with the strong::units:: assigned
+to your spec, if it has not been previously set.
 ::
 
 discussion::
@@ -247,13 +267,24 @@ code::maxval:: of the domainSpecs.
 
 If a new link::#-value:: is set, you will need to update your domainSpecs.
 
-For examples, see link::#-domain:: and link::#examples::.
-
 
 method:: domain
 Set or get the x-axis positions of your data points. The size of the
 code::domainArray:: must equal the size of your value array, i.e. a domain
 value specified for each data point.
+
+code::
+(
+var data, domain, plot;
+domain = 25.collect({rrand(0, 50)}).sort;
+data = domain.linlin(0, 50, 0, 1);
+
+plot = data.plot.plotMode_(\points);
+plot.domainSpecs_([0, 50].asSpec);
+plot.domain_(domain);
+)
+::
+See also the link::#Explicit domain and axis specs:: example below.
 
 discussion::
 Domain values are mapped into the range of the link::#-domainSpecs::, so need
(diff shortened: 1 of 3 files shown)


mtmccrea:
I've rebased this branch onto #4511 to avoid conflicts with updates in that PR, so it should be merged AFTER that one. 
In hindsight, I could have waited for that to be merged then rebased onto `develop` which I may do instead.

telephon (approved):
thank you!

mtmccrea:
>In hindsight, I could have waited for that to be merged then rebased onto develop which I may do instead.

OK, I've rebased onto `develop` and that cleaned up the history. Thanks!

dyfer:
> there are 2 failed tests in `TestServer_boot`, which are unrelated to this PR. I tried re-running, no luck... but I'm guessing it's a problem on the CI machines?

Yes, some tests that use the server occasionally and inconsistently fail, unfortunately.

dyfer:
While I was resuming work on #5161, I noticed that this PR dropped ControlSpec units from the Grid labels. I understand that having a single label in the plotter is better, but it also made `DrawGrid.test` a bit worse. Moreover, there's weird cropping of the numbers happening here. Compare:

```supercollider
(
c = ControlSpec.new(20,20000, \lin, units:"Hz");
g = GridLines(c);
d = DrawGrid.test(g, g, 300@300);
)
```
before this PR:
Screen Shot 2022-08-30 at 09 26 41
after this PR:
Screen Shot 2022-08-30 at 09 27 18

I have a workaround for the cropping that I'm using while I work on the other PR, but I'm not sure what's the best/easiest way to add units to `DrawGrid.test`. Or maybe it's not a big deal there?

What do you think @mtmccrea @telephon ?

For reference, here's my current workaround for `DrawGrit.test`:
<details>

```supercollider
DrawGrid {
(...)
	*test { arg horzGrid,vertGrid,bounds;
		var w, grid, insetW = 0, insetH = 0;
		bounds = bounds ?? {Rect(0,0,500,400)};
		grid = DrawGrid(bounds,horzGrid,vertGrid);
		grid.x.labelOffset = 20 @ 18;
		grid.y.labelOffset = 20 @ 20;
		horzGrid !? {insetH = 20};
		vertGrid !? {insetW = 26};
		w = Window("Grid",bounds).front;
		UserView(w,bounds ?? {w.bounds.moveTo(0,0)})
		.resize_(5)
		.drawFunc_({ arg v;
			grid.bounds = v.bounds.insetAll(insetW, 0, 0, insetH);
			grid.draw
		})
		.background_(Color.white)
		^grid
	}
}
```
</details>
Screen Shot 2022-08-30 at 09 35 20
The numbers are still a little cropped since the margins are hardcoded and previously I was testing with smaller numbers. I don't imagine this is used in production, but I think it should be still functional... (maybe?) Also, I'm not sure why the first label is not there, haven't checked that yet.

mtmccrea:
Thanks for catching this, @dyfer!

This gave me the opportunity to look more closely at `GridLines` and I've put together a patch that should adapt `DrawGrid:*test` to the latest version:
```supercollider
*test { arg horzGrid, vertGrid, bounds;
		var w, grid;
		var insetH = 40, insetV = 20; 	// left, bottom margins for labels
		var pad = 15; 					// right, top margin

		bounds = bounds ?? { Rect(0, 0, 500, 400) };
		bounds = bounds.asRect; 		// in case bounds are a Point
		insetH = insetH + pad;
		insetV = insetV + pad;

		grid = DrawGrid(
			bounds.insetBy(insetH.half, insetV.half).moveTo(insetH-pad, pad),
			horzGrid, vertGrid
		);
		grid.x.labelOffset = insetH @ insetV.half;
		grid.y.labelOffset = insetH @ insetV.half;

		w = Window("Grid", bounds).front;

		UserView(w, bounds ?? { w.bounds.moveTo(0,0) })
		.drawFunc_({ |v| grid.draw })
		.onResize_({ |v|
			grid.bounds = v.bounds.insetBy(insetH.half, insetV.half).moveTo(insetH - pad, pad);
		})
		.resize_(5)
		.background_(Color.white);

		^grid
	}
```

The full fix would also need this setter to be swapped into `DrawGridY`:
```
lblRect = Rect.aboutPoint(
	Point(0, y), labelOffset.x/2, labelOffset.y/2
).right_(bounds.left - txtPad);
```
But may not be wouldn't be necessary for your testing...

In any case, is the best way to just create a new PR or somehow amend this one?

I'd like to revisit this tomorrow, re-test and update docs before pushing.

dyfer:
> This gave me the opportunity to look more closely at `GridLines` and I've put together a patch that should adapt `DrawGrid:*test` to the latest version:

Thanks!

> > The numbers are still a little cropped since the margins are hardcoded and previously I was testing with smaller numbers.
> 
> My example also hard-codes margins, but wide enough for testing default specs, I think.

Sounds good.
 
> > Also, I'm not sure why the first label is not there, haven't checked that yet.
> 
> That's because the spec goes from 20 -> 20k, so GridLines doesn't return a label for 20 Hz (steps are 5k). changing the spec from 0 -> 20k would show a label for 0. You can see the labels generated by the grid: `d.x.commands.do(_.postln)` (at the tail of the message list).

> This is also in part because the recent PR also makes sure grid lines are drawn at the boundaries of the grid, whether or not a labeled gridline coincides with the boundaries. Without this, you would have "uncapped" edges of the plot, which looks strange (see my screenshots in the first post).


Oh... so no label for the first line when not part of the step is consistent with the old behavior, right? I'm not sure I like it that way (would prefer the first line to always have a label), but that's probably something for a separate issue/PR.

> 
> I think it's still reasonable to have made these design decisions based on the `Plotter` use case, which is almost certainly the predominant one. (Another design decision was to not vertically center the top- and botttom-most y-labels on the top/bottom grid lines, otherwise they could collide with top of the view or the leftmost xlabel, respectively).

Yes, I agree the Plotter should be the template. And I like not centering the top/bottom y labels.


> In any case, is the best way to patch this by creating a new PR or somehow amend this one?

I'm thinking - a new PR, on a new branch (well, you could push to this branch and open a new PR with it? but I'm not sure that would work / merge cleanly)

> I'd like to revisit this tomorrow, re-test and update docs before pushing.

Great, thanks!

Aside from the inset, how do you feel about dropping spec units from the `*test` plot? I think that as long as they're included in the Plotter use case it's fine, but I was at first surprised to not see them in the `*test` case.

Finally, one more details: in the new version one of your plots misses vertical units. Is this intentional or some kind of a random occurence?
Screen Shot 2022-08-30 at 20 40 42

mtmccrea:
>> This is also in part because the recent PR also makes sure grid lines are drawn at the boundaries of the grid, whether or not a labeled gridline coincides with the boundaries. Without this, you would have "uncapped" edges of the plot, which looks strange (see my screenshots in the first post).

> Oh... so no label for the first line when not part of the step is consistent with the old behavior, right? 

Almost... in the case at hand, previously, there would have been _no first line_, just blank space (the 'quantized' GridLines would have started at 5k). This PR added lines at the extents of the grid. You can see this behavior by dynamically resizing the window and watching how the grid behaves at the extents.

> I'm not sure I like it that way (would prefer the first line to always have a label), but that's probably something for a separate issue/PR.

Yes, this would actually be a change in `GridLines` (likely `-getParams`) so that it produces a label/line for the values at the max and min of the spec (or range), not just the in-range 'quantized' steps.

>>In any case, is the best way to patch this by creating a new PR or somehow amend this one?

>I'm thinking - a new PR, on a new branch (well, you could push to this branch and open a new PR with it? but I'm not sure that would work / merge cleanly)

Ok, I'll put together a new PR 👍

> Aside from the inset, how do you feel about dropping spec units from the *test plot? I think that as long as they're included in the Plotter use case it's fine, but I was at first surprised to not see them in the *test case.

I'll look into this...  `Plotter` now puts the units in the _axis_ `labelX/Y`, not the _grid_ labels. It does this when setting the `spec` and `domainSpec`, which isn't happening in this `*test`, so it may just be a matter of explicitly setting the label.

> Finally, one more details: in the new version one of your plots misses vertical units. Is this intentional or some kind of a random occurrence?

The 'before' and 'after' images above aren't exact comparisons, because the plot sizes were just adjusted by hand to show a variety of behavior, including when labels automatically disappear below window size limits. So the case you highlighted likely isn't omitting units in particular, it's omitting all of the y-labels (axis and grid labels) altogether because of the size of the window, but I'll confirm this too.

mtmccrea:
See #5858 for the updated behavior:
Screen Shot 2022-08-31 at 15 06 34

PR #5828, opened 2022-07-30
Classlib: QuarksGUI should check schelp as well as helpdoc

jamshark70:
## Purpose and Motivation

Quark:help first tries to display a SCDoc help file, if one is listed under `schelp` in the quark file.

Failing this, it falls back to `helpdoc` (which is handled by `openOS`).

But QuarksGUI will display the "Open help" button only if `helpdoc` is populated.

So, in current development, if a quark author wants the help doc to be a SCDoc help file, they have to specify both schelp and helpdoc, but helpdoc will never be used.

Expected behavior: Check both keys, and display the button if either one is populated with a string.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
+++ b/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
@@ -322,7 +322,9 @@ QuarkDetailView {
 				this.pushRow("Dependencies", dependencies.collect(_.asString).join(Char.nl));
 			};
 			if(isInstalled or: {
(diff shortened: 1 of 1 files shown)


dyfer (approved):
Thanks!

This works as expected. 
For future reference - the fix only applies to the situation when the quark is downloaded but _not_ installed. "Open help" button was already properly shown for the installed quarks.

PR #5836, opened 2022-07-30
docs: fixing various typos and add general improvements

elifieldsteel:
## Purpose and Motivation
moved s.boot so it appears before a SynthDef instead of after.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/EZRanger.schelp
+++ b/HelpSource/Classes/EZRanger.schelp
@@ -1,13 +1,13 @@
 class:: EZRanger
 summary:: A wrapper class for a label, a rangeslider, and numberboxes
 categories:: GUI>EZ-GUI
-related:: Classes/EZGui, Classes/StaticText, Classes/Slider, Classes/NumberBox
+related:: Classes/EZGui, Classes/StaticText, Classes/RangeSlider, Classes/NumberBox
 
 description::
-EZRanger is wrapper class which creates an (optional) link::Classes/StaticText::, and a link::Classes/Slider:: plus a link::Classes/NumberBox::. If the parent is code::nil::, then EZRanger will create its own window. See link::Classes/EZGui:: more options.
+EZRanger is wrapper class which creates an (optional) link::Classes/StaticText::, and a link::Classes/RangeSlider:: plus a link::Classes/NumberBox::. If the parent is code::nil::, then EZRanger will create its own window. See link::Classes/EZGui:: more options.
 
 subsection:: Scrolling and Arrow Keys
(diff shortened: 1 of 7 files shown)


jamshark70:
FWIW, the change is fine, but also unnecessary: for a long time now, you can `.add` a SynthDef before booting the server, and it will be sent just after boot.

BTW the changes in several PRs accumulated (and I guess this will be the case for the other PRs), so merging is all-or-nothing at this point.

elifieldsteel:
Thanks for the info about adding before booting — was not aware.

I'm not super familiar with git, apologies for the accumulation. I was under the impression it would be somehow "cleaner" to create separate PRs for changes to separate files, and that they could be approved individually. If it's easiest to just reject them, please do. Is it better to cram them all into one PR?

I keep track of minor issues in the docs and am just now getting around to contributing the changes. If there's something I should be doing differently, let me know. That said, I don't think any of my proposed edits are wrong or break anything.

jamshark70:
> Thanks for the info about adding before booting — was not aware.

However -- that's relevant to your recent ServerBoot question on the forum (heading there next).

> I'm not super familiar with git, apologies for the accumulation. I was under the impression it would be somehow "cleaner" to create separate PRs for changes to separate files, and that they could be approved individually.

It is cleaner to have separate PRs, yes.

*Probably* the accumulation comes from a process like this:

1. Check out, and update, the develop branch.
2. `git checkout -b topic/MyNewTopic`
3. Make changes, commit, push, make PR.
4. `git checkout -b topic/MyOtherNewTopic`

... now, at 4, you have a second PR branch which is based on MyNewTopic, and contains all of the changes made in MyNewTopic *plus* the new changes.

This can be avoided by doing `git checkout develop` before making each new topic branch -- so that they are all based on the same source.

elifieldsteel:
> This can be avoided by doing `git checkout develop` before making each new topic branch -- so that they are all based on the same source.

I see — this makes perfect sense, and in hindsight, I'm annoyed I didn't realize this is what I was doing. Apologies again for making a bit of a mess. As I said, if it's easiest for you to just reject my changes, I don't might starting over with separate PRs.

elifieldsteel:
Or — would it make more sense for me to withdraw these PRs myself, and start over with better technique? If so, how would I withdraw them?

telephon (approved):
I've changed the title of the PR, this is the easiest! Sometimes it's fine to just add a bunch of improvements to documentation, if they are unproblematic. Next time, you can make them a little more fine grained, but this seems good to me.

dyfer:
Thanks @elifieldsteel @telephon @jamshark70, I'll merge this.

dyfer:
So... the slight twist (for me) here is that all the other PRs that also had these commits, automatically were marked as merged when I merged this. This is all fine, I was about to go through them and mark them as closed in favor of this PR.

dyfer:
@elifieldsteel I'll comment on resolving a situation that you encountered (creating a branch from a branch that has other changes instead of `develop`) in #5838 - now that this PR is in, we'll need to "fix" that other PR.

PR #5845, opened 2022-08-16
Update macOS build matrix

### LINKED ISSUE #5843: GHA macOS: 10.15 image will get deprecated and should be updated
## Motivation

We currently use macOS-10.15 image for our macOS builds. This image is deprecated and will stop working on 8/30/2022 (see https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/). 

Update our build matrix to use the newer macOS images.

## Plan for Implementation

I'll provide a PR with updated build matrix.

I'm thinking to jump ahead and use macOS-12 and the newest non-beta Xcode, but I'm open to suggestions on that. Let's discuss this in the PR once I open it.

dyfer:
## Purpose and Motivation

Fixes #5843

Github Actions runner image of macOS 10.15 is getting deprecated. This PR updates our build matrix to use macOS 11 and macOS 12 images.

- the legacy build still works down to macOS 10.10. I confirmed that by testing on that system
  -  in order for this to work, I switched to build the "vendored" portaudio library (just for the legacy build); this applies to supernova only anyway
  - for reference, the build only seemed to work on 10.10 when built using Xcode 12.4; it did not work when built using 11.7 or 13.1 (I haven't investigated that further, just chose the configuration that worked)
- the main build **doesn't work on macOS 10.13** (High Sierra) anymore. This is due to the use of homebrew build of Qt, which only supports macOS 10.14 (for homebrew installed on macOS 11). I've updated the documentation accordingly
- I've set other macOS builds to use macOS 12 and Xcode 13.4.1 for testing purposes
  - I've originally tried to set the main build to use this config as well, but it further limited backwards compatibility of the release build, so I stuck to using macOS 11 and Xcode 12 there

I have tested (*) these builds on macOS 10.10, 10.13 (legacy), 10.14 and 12.4 (main)

(*) I've only checked whether the app opens, interpreter starts and whether the servers boot

## Types of changes

- Bug fix (?)
- Breaking change (platform support for the main SC build)

## To-do list

- [x] Code is tested
- [x] All tests are passing (the application runs on various platforms as described above)
- [x] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -191,9 +191,9 @@ jobs:
         include:
 
           - job-name: ''
-            os-version: '10.15'
-            xcode-version: '11.7'
-            deployment-target: '10.13' # Qt 5.15 runs on macOS 10.13 and newer
+            os-version: '11'
+            xcode-version: '12.4'
+            deployment-target: '10.14' # Qt 5.15 runs on macOS 10.13 and newer; however homebrew build of Qt on macOS 11 only supports 10.14+
             use-syslibs: false
             shared-libscsynth: false
             build-libsndfile: true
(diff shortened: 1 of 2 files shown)


joshpar (approved):

joshpar:
I say go for it!

PR #5850, opened 2022-08-21
sclang/scsynth/supernova: Two fixes for UdpInPort error reporting

jamshark70:
## Purpose and Motivation

This fixes a couple of discrepancies in UdpInPort error reporting.

1. sclang prints newline after the error text; scsynth inexplicably does not. Obviously a typo.

2. In the current release, the above typo is the *only* way to tell whether sclang or scsynth encountered a UdpInPort error. This is rather less than informative from the user's perspective. So, this PR adds prefixes "(sclang)" "(scsynth)" "(supernova)" to the error posting.

3. sclang suppresses "connection refused" errors (apparently because this error happens a lot in Windows, spuriously). scsynth does not. Windows users encounter this error, particularly when quitting the server, and then worry that something has gone terribly wrong. So this PR copies the sclang `if` block into scsynth.

For the third -- I did not port the block into supernova -- I can do so if others feel that I should. Supernova tends to be more "strict" (favoring "correctness" over user-friendliness) so I think it's in keeping with the feeling of supernova that it would report unnecessary errors. But, if we would prefer supernova to behave closer to scsynth in this regard, then I'm happy to add another commit.

I suppose it could be called a bug, but as it's basically cosmetic, I left off the "bug" label.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing (code builds successfully; both scsynth and supernova start, and receive messages OK)
- [ ] Updated documentation (n/a, I think)

### DIFF
--- a/lang/LangPrimSource/SC_ComPort.cpp
+++ b/lang/LangPrimSource/SC_ComPort.cpp
@@ -111,7 +111,7 @@ void SC_UdpInPort::handleReceivedUDP(const boost::system::error_code& error, std
     }
 
     if (error) {
-        printf("SC_UdpInPort: received error - %s\n", error.message().c_str());
+        printf("(sclang) SC_UdpInPort: received error - %s\n", error.message().c_str());
         startReceiveUDP();
         return;
     }
--- a/server/scsynth/SC_ComPort.cpp
+++ b/server/scsynth/SC_ComPort.cpp
@@ -206,8 +206,14 @@ class SC_UdpInPort {
         if (error == boost::asio::error::operation_aborted)
             return; /* we're done */
 
+        if (error == boost::asio::error::connection_refused) {
+            // avoid windows error message
+            startReceiveUDP();
+            return;
+        }
+
         if (error) {
-            printf("SC_UdpInPort: received error - %s", error.message().c_str());
+            printf("(scsynth) SC_UdpInPort: received error - %s\n", error.message().c_str());
             startReceiveUDP();
             return;
         }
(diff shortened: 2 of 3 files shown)


jamshark70:
CI failures:

2 macOS jobs

DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support

1 macOS job

Process completed with exit code 1.

Searching the log for this job, for "FAIL:", turns up:

FAIL: a TestServer_boot: test_notifyAndServerTreeActions - Toggling Server.notify should not cause ServerTree actions to run.
Is:
	 3 
Should be:
	 1

... which is unrelated to this PR.

1 Linux job

The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

dyfer:
CI failures are on my radar, but I don't have a fix yet. 
<details>

If anybody is interested - I tried moving on from Ubuntu 18.04, but I had problems with QPM (our test runner) on 20.04. 
I'm tentatively thinking about not using QPM in the CI, once that is ready then I'll remove 18.04 jobs.

As for `TestServer_boot` errors - these show up inconsistently. I've tried poking around but couldn't find a solution. It seems that in the CI we end up either with hanging/multiple server process, or some other boot issue (e.g. I often see "couldn't create shm interface" errors) which derail tests that would otherwise pass.

</details>

joshpar (approved):

PR #5851, opened 2022-08-24
AppClock schedAbs

### LINKED ISSUE #5700: Task-play fails when using AppClock
## Environment
* SuperCollider version: current develop branch
* Operating system: macOS

## Steps to reproduce

```supercollider
// Routine works fine with AppClock:
Routine { "rout_appclock".postln }.play(AppClock); 

// Task fails: 
Task { "task_appclock".postln }.play(AppClock);
	// -> fails with schedAbs not understood, Receiver Meta_AppClock

```

Tasks should be fully playable with AppClock. 
#5626 introduced clock.schedAbs in PauseStream:play, 
which assumes that any clock understands schedAbs. 
So maybe we need to add an AppClock:schedAbs method? 

```
// full error dump
ERROR: Message 'schedAbs' not understood.
Perhaps you misspelled 'sched', or meant to call 'schedAbs' on another receiver?
RECEIVER:
class AppClock (0x11a993bc0) {
  instance variables [19]
    name : Symbol 'AppClock'
    nextclass : instance of Meta_ApplicationStart (0x11a3cff40, size=19, set=5)
    superclass : Symbol 'Clock'
    subclasses : nil
    methods : nil
    instVarNames : nil
    classVarNames : instance of SymbolArray (0x11a993d40, size=1, set=2)
    iprototype : nil
    cprototype : instance of Array (0x11a993e00, size=1, set=2)
    constNames : nil
    constValues : nil
    instanceFormat : Integer 0
    instanceFlags : Integer 0
    classIndex : Integer 42
    classFlags : Integer 0
    maxSubclassIndex : Integer 42
    filenameSymbol : Symbol '/Users/adc/src/_SC/supercollider/SCClassLibrary/Common/Core/Clock.sc'
    charPos : Integer 935
    classVarIndex : Integer 399
}
ARGS:
   Float 77.216567   3DE4CE32 40534DDC
Instance of Function {    (0x1229e5018, gc=94, fmt=00, flg=00, set=02)
  instance variables [2]
    def : instance of FunctionDef in Method PauseStream:play
    context : Frame (0x122d20378) of PauseStream:play
}
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg selector = 'schedAbs'
		arg args = [*2]
	PauseStream:play
		arg this = <instance of Task>
		arg argClock = <instance of Meta_AppClock>
		arg doReset = false
		arg quant = nil
		var tempClock = <instance of Meta_AppClock>
		var time = 77.216567491
		var res = nil
		var code = "Task { "task_appclock".postl..."
		var doc = nil
	Process:interpretPrintCmdLine
		arg this = <instance of Main>
^^ ERROR: Message 'schedAbs' not understood.
Perhaps you misspelled 'sched', or meant to call 'schedAbs' on another receiver?
RECEIVER: AppClock

### PR DESCRIPTION
**Pull request by telephon**
## Purpose and Motivation

Fixes #5700.

## Types of changes

- Documentation
- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestTask.sc
+++ b/testsuite/classlibrary/TestTask.sc
@@ -32,7 +32,7 @@ TestTask : UnitTest {
 	}
 
 	test_play_with_TempoClock {
-		this.makePlayTestWithClock(TempoClock.default, "TempoClock.default");
+		this.makePlayTestWithClock(TempoClock.new, "TempoClock.new");
 	}
 
 	test_play_with_LinkClock {
--- a/testsuite/classlibrary/TestTask.sc
+++ b/testsuite/classlibrary/TestTask.sc
@@ -23,44 +23,31 @@ TestTask : UnitTest {
 		this.assert(task.isPlaying.not, "Task be stoppable even if not running");
 	}
 
-
 	test_play_with_SystemClock {
-		var ok = false;
-		var task = Task { 0.001.wait; ok = true; };
-		task.play(SystemClock);
-		0.002.wait;
-		task.stop;
-		this.assert(ok, "Task plays with SystemClock");
-	}
-
-	test_play_with_TempoClock {
-		var ok = false;
-		var task = Task { 0.001.wait; ok = true; };
-		task.play(TempoClock.default);
-		0.002.wait;
-		task.stop;
-		this.assert(ok, "Task plays with TempoClock.default");
+		this.makeTestWithClock(SystemClock, "SystemClock");
 	}
 
 	test_play_with_AppClock {
-		var ok = false;
-		var task = Task { 0.001.wait; ok = true; };
-		task.play(AppClock);
-		0.002.wait;
-		task.stop;
-		this.assert(ok, "Task plays with AppClock");
+		this.makeTestWithClock(AppClock, "AppClock");
+	}
+
+	test_play_with_TempoClock {
+		this.makeTestWithClock(TempoClock.default, "TempoClock.default");
 	}
 
 	test_play_with_LinkClock {
+		this.makeTestWithClock(LinkClock.new, "LinkClock.new");
+	}
+
+	makeTestWithClock { |clock, nameString|
 		var ok = false;
 		var task = Task { 0.001.wait; ok = true; cond.signalAll };
-		var clock = LinkClock.new;
 		var cond = CondVar.new;
-		task.play(LinkClock);
+		task.play(clock);
 		cond.waitFor(1, { ok });
 		task.stop;
 		clock.stop;
-		this.assert(ok, "Task plays with LinkClock");
+		this.assert(ok, "Task plays with " ++ nameString);
 	}
 
 }
--- a/testsuite/classlibrary/TestTask.sc
+++ b/testsuite/classlibrary/TestTask.sc
@@ -53,10 +53,14 @@ TestTask : UnitTest {
 
 	test_play_with_LinkClock {
 		var ok = false;
-		var task = Task { 0.001.wait; ok = true; };
+		var task = Task { 0.001.wait; ok = true; cond.signalAll };
+		var clock = LinkClock.new;
+		var cond = CondVar.new;
 		task.play(LinkClock);
-		0.002.wait;
+		cond.waitFor(1, { ok });
 		task.stop;
+		clock.stop;
 		this.assert(ok, "Task plays with LinkClock");
 	}
+
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: testsuite/classlibrary/TestTask.sc [hunk 1]
```diff
@@ -23,4 +23,40 @@ TestTask : UnitTest {
...
+		task.play(AppClock);
+		0.002.wait;
+		task.stop;
+		this.assert(ok, "Task plays with AppClock");
+	}
+
+	test_play_with_LinkClock {
+		var ok = false;
+		var task = Task { 0.001.wait; ok = true; };
+		task.play(LinkClock);
```
jamshark70, line 57: This is not actually testing on a LinkClock.

```
LinkClock.default
-> a TempoClock

Task {
	thisThread.clock.postln;
}.play(LinkClock);

a TempoClock  // printed
```

LinkClock depends on the Ableton Link backend, which is dormant until an instance is created (see `prStart`, calling `_LinkClock_New`). This means, if the code hasn't created an instance, then it isn't using LinkClock.

It may also be necessary to wait a bit longer. *If* there are no other Link peers on the network at the time of running the test, then the Task will awaken *now*, with beats = 0.0 as we would expect from TempoClock. If there are other peers, then the clock time may be adjusted forward or backward to match the Link quantum. This shouldn't be the case in most testing, but you never know.

To cover all the bases, I'd suggest like this:

```
	test_play_with_LinkClock {
		var ok = false;
		var task = Task { 0.001.wait; ok = true; cond.signalAll };
		var clock = LinkClock.new;
		var cond = CondVar.new;
		task.play(LinkClock);
		cond.waitFor(1, { ok });
		task.stop;
		clock.stop;
		this.assert(ok, "Task plays with LinkClock");
	}
```

jamshark70 (changes requested):
In addition to the LinkClock issue, would you mind adding documentation for schedAbs? New methods should be documented unless they are private.

telephon:
> In addition to the LinkClock issue, would you mind adding documentation for schedAbs? New methods should be documented unless they are private.

I just had a quick coffee. :)

jamshark70:
> I just had a quick coffee. :)

Ha ha! OK... what happened was, I was just sitting down to add schedAbs myself, and then saw an e-mail that you had done it, and went to review it quickly.

I think that a prior SC maintainer would have preferred CondVar over `0.002.wait` (out of a general preference for avoiding hardcoded wait times). For that matter, even the `0.001.wait` within the Tasks are unnecessary; for the test, we only require the Task to awaken once, not twice. I guess I personally won't be that much of a stickler, but I do see their prior point that habitually consuming extra milliseconds in testing could cumulatively slow down the test suite more than you'd think.

So perhaps:

```
	test_play_with_AppClock {
		var ok = false;
		var cond = CondVar.new;
		var task = Task { ok = true; cond.signalAll };
		task.play(AppClock);
		cond.waitFor(0.1, { ok });
		task.stop;
		this.assert(ok, "Task plays with AppClock");
	}
```

telephon:
Yes, I did see your suggestion further up. I had decided for simplicity – while I do understand conditions,  I have to admit that I have to read three times the code needed to use them …

telephon:
OK, so here is a test series that factors out the "complicated" bits.

telephon:
@jamshark70 if you have the time, could you review it and if ok mark your change request as fulfilled?

testsuite/classlibrary/TestTask.sc
```diff
@@ -23,4 +23,31 @@ TestTask : UnitTest {
...
+	test_play_with_SystemClock {
+		this.makePlayTestWithClock(SystemClock, "SystemClock");
+	}
+
+	test_play_with_AppClock {
+		this.makePlayTestWithClock(AppClock, "AppClock");
+	}
+
+	test_play_with_TempoClock {
+		this.makePlayTestWithClock(TempoClock.default, "TempoClock.default");
```
jamshark70, line 35: I'd have to suggest `TempoClock.new` here, because it will be `stop`-ped later. (As a rule of thumb, if this block of code didn't create the clock, then it shouldn't be stopping the clock.)

dyfer:
@jamshark70 do you have any more requests for this PR?

jamshark70:
LGTM :+1:

jamshark70 (approved):

PR #5854, opened 2022-08-28
Plugins: Add 'ramp' flag to K2A

### LINKED ISSUE #5803: Add an argument to K2A to switch linear interpolation on or off
## Motivation

There are cases where it may be useful to do K2A as a zero order hold, rather than a linear ramp.

Currently there is no way at all, anywhere in SC, to convert control to audio rate with a zero order hold! That strikes me as a rather arbitrary limitation.

K2A has no mul/add arguments, so it would be safe to add a second input:

```
	*ar { arg in = 0.0, interpolation = 1.0;
		^this.multiNew('audio', in, interpolation)
	}
```

So the default behavior would be unchanged.

## Plan for Implementation

Doesn't seem difficult. If others like the idea, I think I could look at it during the summer holiday.

jamshark70:
## Purpose and Motivation

Fixes #5803.

Currently, SC UGens assume that a control rate signal should always linearly interpolate, when being promoted to audio rate.

In cases where a user needs a zero-order hold, as far as I know, at present the only way is with Duty.ar. But that seems inefficient, clunky and un-intuitive.

One such use case: In my ddwWavetableSynth quark, the wavetable-position input is translated into an index into a buffer. This index must be aligned with the beginning of one of the wavetable variants (using some rounding math that I've used for a long time, for flip-flop crossfading between a pair of chains). When the index is interpolated, alignment is not guaranteed -- and in fact, if I modulate wtPos rapidly, the artifacts are unacceptable.

Note: #5803 proposed making 'ramp' a floating point value, such that 0.5 would ramp over half the control block, and 0.75 over 3/4 of it. I'm not skilled enough with advanced C++ to see how to do that with the SIMD functions. I'm not opposed to the idea but would basically need someone else to explain "for dummies" how to do it (or write it, and I'll check it in).

## Types of changes

- Documentation
- New feature
- Breaking change -- well... technically it's "breaking" in that there is a new input. But, K2A has never had a second input (not even `mul, add`, which don't make sense in this context anyway), so it doesn't invalidate any prior usage. (WRT non-SC clients, 2nd, 3rd inputs to K2A have always been ignored in the server.)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/K2A.schelp
+++ b/HelpSource/Classes/K2A.schelp
@@ -7,8 +7,8 @@ categories::  UGens>Conversion
 Description::
 
 To be able to play a control rate UGen into an audio rate UGen,
-sometimes the rate must be converted. K2A converts via linear
-interpolation.
+sometimes the rate must be converted. By default, K2A converts via
+linear interpolation, though the interpolation can be turned off.
 
 
 classmethods::
@@ -17,6 +17,9 @@ method::ar
 
 argument::in
 The input signal.
+
+argument::ramp
+A flag to enable or disable linear interpolation. If nonzero, K2A will interpolate; if zero, there is no interpolation (the output is a zero-order hold, or stairstep). The default is 1.0 (enabled).
 
 Examples::
 
@@ -48,3 +51,22 @@ code::
 )
 ::
 
+subsection:: Ramp behavior
+
+The following plot contrasts a non-interpolated K2A against a normally interpolated K2A.
+
+If some degree of interpolation is desired, but not necessarily occupying a full control block, one option is to apply link::Classes/Slew:: to a non-interpolated K2A. (Note that the ramp time in this case will depend on the difference between the successive control rate values. Here, the difference between LFPulse values is always +1 or -1, so a slope of code::ControlRate.ir * 2:: will ramp over half of a control block. A smaller or larger differential would produce a shorter or longer ramp.)
+
+code::
+(
+{
+	var freq = ControlRate.ir * 0.25;
+	var krpulse = LFPulse.kr(freq);
+	var noRamp = K2A.ar(krpulse, 0);
+	var ramp = K2A.ar(krpulse, 1);
+	var slope = ControlRate.ir * 2;
+	var partialRamp = Slew.ar(noRamp, slope, slope);
+	[noRamp, ramp, partialRamp]
+}.plot;
+)
+::
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -1199,7 +1199,7 @@ struct K2A : SIMD_Unit {
 
     K2A(void) {
         mLevel.init(this);
-	mRamp.init(this);
+        mRamp.init(this);
         if (inRate(0) == calc_ScalarRate)
             set_unrolled_calc_function<K2A, &K2A::next_i<unrolled_64>, &K2A::next_i<unrolled>, &K2A::next_i<scalar>>();
         else
@@ -1212,7 +1212,7 @@ struct K2A : SIMD_Unit {
         else {
             mLevel.init(this);
             next_i<type>(inNumSamples);
-	}
+        }
     }
 
     template <int type> void next_i(int inNumSamples) { set_vec<type>(out(0), mLevel, inNumSamples); }
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -1195,9 +1195,11 @@ void SyncSaw_Ctor(SyncSaw* unit) {
 
 struct K2A : SIMD_Unit {
     ControlRateInput<0> mLevel;
+    ControlRateInput<1> mRamp;
 
     K2A(void) {
         mLevel.init(this);
+	mRamp.init(this);
         if (inRate(0) == calc_ScalarRate)
             set_unrolled_calc_function<K2A, &K2A::next_i<unrolled_64>, &K2A::next_i<unrolled>, &K2A::next_i<scalar>>();
         else
@@ -1205,10 +1207,12 @@ struct K2A : SIMD_Unit {
     }
 
     template <int type> void next_k(int inNumSamples) {
-        if (mLevel.changed(this))
+        if (mRamp && mLevel.changed(this))
             slope_vec<type>(out(0), mLevel.slope(this), inNumSamples);
-        else
+        else {
+            mLevel.init(this);
             next_i<type>(inNumSamples);
+	}
     }
 
     template <int type> void next_i(int inNumSamples) { set_vec<type>(out(0), mLevel, inNumSamples); }
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -619,5 +619,43 @@ TestCoreUGens : UnitTest {
 
 	}
 
+	test_K2A_ramp_input {
+		var buf, node, flag = false;
+		var data;
+		var cond = CondVar.new;
 
+		server.bootSync;
+
+		buf = Buffer.alloc(server, server.options.blockSize * 16, 2);
+		server.sync;
+
+		node = {
+			var pulse = LFPulse.kr(ControlRate.ir * 0.25);
+			RecordBuf.ar(K2A.ar(pulse, [0, 1]), buf, loop: 0, doneAction: 2);
+			Silent.ar(1)
+		}.play(server.defaultGroup);
+		node.onFree {
+			flag = true;
+			cond.signalAll
+		};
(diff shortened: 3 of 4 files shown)


testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -619,5 +619,43 @@ TestCoreUGens : UnitTest {
...
+		var cond = CondVar.new;
 
+		server.bootSync;
+
+		buf = Buffer.alloc(server, server.options.blockSize * 16, 2);
+		server.sync;
+
+		node = {
+			var pulse = LFPulse.kr(ControlRate.ir * 0.25);
+			RecordBuf.ar(K2A.ar(pulse, [0, 1]), buf, loop: 0, doneAction: 2);
```
telephon, line 634: you can also call `.asBuffer` directly on the function. That might be shorter.

mtmccrea: or even more directly, `.loadToFloatArray` can be called on a function.

HelpSource/Classes/K2A.schelp
```diff
@@ -7,8 +7,8 @@ categories::  UGens>Conversion
 Description::
 
 To be able to play a control rate UGen into an audio rate UGen,
-sometimes the rate must be converted. K2A converts via linear
-interpolation.
+sometimes the rate must be converted. By default, K2A converts via
```
mtmccrea, line 10: Here's a suggestion for a more verbose description of what K2A does, and what to expect if `ramp=0.0`:

`K2A` will generate audio-rate (code::ar::) samples from a control-rate (code::kr::) input sample. By default, this is done by linearly interpolating from the previous code::kr:: sample to the new sample over the period of the control block (64 samples by default). Therefore, the input value will be reached one block after it is received. This can be disabled by setting code::ramp = 0::, in which case the input value is immediately output and held over the control block (sample-and-hold).

mtmccrea:
This is a useful feature, in principle.

>Note: https://github.com/supercollider/supercollider/issues/5803 proposed making 'ramp' a floating point value, such that 0.5 would ramp over half the control block, and 0.75 over 3/4 of it.

I would also be in favor of this solution if possible. Making **ramp** a boolean state now means adding this feature later would be a breaking change. Offering `Slew` as a workaround in the docs is useful, but I wonder if we couldn't solve the issue in `K2A` directly.

If it's decided that a boolean flag is sufficient, should the default be `1` instead of `1.0`?

server/plugins/LFUGens.cpp
```diff
@@ -1195,20 +1195,24 @@ void SyncSaw_Ctor(SyncSaw* unit) {
...
         if (inRate(0) == calc_ScalarRate)
             set_unrolled_calc_function<K2A, &K2A::next_i<unrolled_64>, &K2A::next_i<unrolled>, &K2A::next_i<scalar>>();
         else
             set_unrolled_calc_function<K2A, &K2A::next_k<unrolled_64>, &K2A::next_k<unrolled>, &K2A::next_k<scalar>>();
     }
 
     template <int type> void next_k(int inNumSamples) {
-        if (mLevel.changed(this))
+        if (mRamp && mLevel.changed(this))
             slope_vec<type>(out(0), mLevel.slope(this), inNumSamples);
```
mtmccrea, line 1211: Achieving the float-valued ramp behavior referenced in the OP might be something like (best guess without testing it!):
```
int nRampSamps = (int) mRamp * inNumSamples;
slope_vec<type>(out(0), mLevel.slope(this), nRampSamps);
slope_vec<type>(out(nRampSamps), 0.f, inNumSamples - nRampSamps);
```
bounds of `nRampSamps` would of course need to be checked

redFrik:
there are a couple of other places where similar flags are used.  for example Pitch's `clar` argument and it's an Integer.

mtmccrea:
It would really just be for the user, a cue that the parameter is a boolean state, in both docs and auto-complete prompts.

server/plugins/LFUGens.cpp
```diff
@@ -1195,20 +1195,24 @@ void SyncSaw_Ctor(SyncSaw* unit) {
...
+        mRamp.init(this);
         if (inRate(0) == calc_ScalarRate)
             set_unrolled_calc_function<K2A, &K2A::next_i<unrolled_64>, &K2A::next_i<unrolled>, &K2A::next_i<scalar>>();
         else
             set_unrolled_calc_function<K2A, &K2A::next_k<unrolled_64>, &K2A::next_k<unrolled>, &K2A::next_k<scalar>>();
     }
 
     template <int type> void next_k(int inNumSamples) {
-        if (mLevel.changed(this))
+        if (mRamp && mLevel.changed(this))
```
mtmccrea, line 1210: Should `mRamp` be protected from being negative? (a negative value [would answer true](https://stackoverflow.com/a/18840449/11357554) here, I believe)

jamshark70:
I think I'll make an executive decision here.

It would be better for `ramp` to be a float 0.0 - 1.0, where 1.0 would ramp over the full control block, 0.0 would not ramp at all, and 0.5 would ramp over half of the control block.

However, SIMD_Unit's ControlRateInput... I just can't figure it out. We had a nice easy way to get ZIN/ZOUT... I was just trying to test @mtmccrea's suggestion (which appeared in an e-mail that appears to have been deleted in github), but I ran into static trying to get the latest value of mRamp (to `std::clamp` it) because the latest value is abstracted out into an undocumented interface. `std::clamp(mRamp, 0.f, 1.f)` choked in the compiler.

Yeah, that kind of thing, I'm out. (I appreciate Tim's reasons for preferring cooler newer C++ devices but for me, that will be a barrier to working on UGens.) I also think @mtmccrea is right that if we merge in the Boolean version, and then later change it for partial-block interpolation, then that would be a breaking change = bad. So I'll withdraw this PR and let someone else implement the better idea.

Of Boolean UGen inputs: As it happens, Pitch's `clar` argument does define Boolean as `input > 0.f`, so to follow precedent, indeed, it should have been `mRamp > 1.f`. (Is this written down anywhere, as a general principle?)

telephon:
> Yeah, that kind of thing, I'm out. (I appreciate Tim's reasons for preferring cooler newer C++ devices but for me, that will be a barrier to working on UGens.)

I have the same thing. For occasional authors and contributors, the added "complications" (which of course for c++ programmers are simplifications) form a social barrier. Perhaps we need a discussion about how this can be avoided in the future – we should avoid "professionalisation".




> I also think @mtmccrea is right that if we merge in the Boolean version, and then later change it for partial-block interpolation, then that would be a breaking change = bad. So I'll withdraw this PR and let someone else implement the better idea.

I don't think it needs to be a breaking change. Either …

- make the value non-modulatable in this version and require a number as an input. This is of course a disadvantage, but still would work.
- alternatively, wrapping it and in the case of a modulatable value (`isNumber.not`), the UGen could come wrapped in a `Ramp` or similar UGen.

mtmccrea:
> I was just trying to test @mtmccrea's suggestion (which appeared in an e-mail that appears to have been deleted in github)

Sorry I'm trying to break the habit of editing too much after posting 😬, and in this case I was going to toss a suggestion out, but I wasn't in a place to even cursorily verify it, so I removed it.

>(Is this written down anywhere, as a general principle?)

Good call, we should have some of these common patterns documented!

>I don't think it needs to be a breaking change. Two possibilities for the first implementation would avoid that, either …
> - make the value non-modulatable in this version and require a number as an input. This is of course a disadvantage, but still would work.

IIUC, modulation isn't the issue, but rather if someone provides `ramp = 0.5`, currently they'd get a ramp over the full control period. In the suggested future update that argument would produce a ramp over half of the control period, making it 'breaking'.

> - alternatively, making it a pseudo-UGen and in the case of a modulatable value (isNumber.not), the UGen could come wrapped in a Ramp or similar UGen. A later version could cast this into a single UGen.

Curious to know if a pseudo-UGen is possible, given that a sample-and-hold version of `K2A` seems to be a necessary upstream component of any modified interpolation. I did some playing around, and couldn't avoid the delay of `ControlDur`.
```
(
{
	var ugen = SinOsc;
	var freq = 225;
	var krSig = ugen.kr(freq);
	var arRef = ugen.ar(freq);
	var k2a = K2A.ar(krSig);
	var latchTrig = Impulse.ar(ControlRate.ir);
	[
		arRef, // red
		k2a, // blue
		latchTrig, // cyan
		VarLag.ar( // green
			Latch.ar(k2a, Impulse.ar(ControlRate.ir)), 
			time: ControlDur.ir, 
			curvature: 0,
			warp: \step
		)
	]
}.plot(0.005)
.plotMode_(\plines)
.plotColor_([\red,\blue,\cyan,\green].collect(Color.perform(_)))
// .superpose_(true)
)
```

Issue #2914, opened 2017-06-02
style: Semicolon after last statement of block


brianlheim:
# Issue

SuperCollider gives the option of leaving off the semicolon for the last statement of a block. In the class library, which should we prefer?

### With semicolon
```supercollider
func {
    var value;
    value = this.multiplier * 3;
    ^value;
}
```

### Without semicolon
```supercollider
func {
    var value;
    value = this.multiplier * 3;
    ^value
}
```

snappizz:
with. that way statements are reorderable.

brianlheim:
I also prefer with-semicolon for that reason. I guess I always thought of it as a convenience to be able to omit a semicolon while prototyping or in one-line statements like `collection.do { |x| x.postln }`. But when it comes to more intentional development, I've been bitten several times by forgetting to add a semicolon to the former final statement.

I'd also like to add while we're having these discussions that another dimension to consider is strictness/stringency. For example, if we had categories of guidelines like recommend vs require, I'd say this one is a recommendation—it's worth knowing the preference is there, but it isn't a big deal if it's not consistently applied. #2913 is more of a requirement IMO because the syntax possibilities look so very different.

telephon:
Unlike in other languages, in sclang, the semicolon is not part of a statement, but it is a delimiter *between* statements. In such a way, it is very similar to white spaces in a sentence. One would not write `I am so happy .` but `I am so happy.`  This is because `happy` is the word and not `happy `.

Expressions that are followed by a semicolon suggest that another expression follows. A function return value isn't followed by any other statement.

I'd recommend the following convention (in continuity with a tacit practice):

1. for functions that are not methods never add a trailing semicolon: this communicates that the last expression is intended as the function's return value
2. for methods, follow the same principle wherever the last expression is intended not to be followed by another one (e.g. if there is only one `return` statement). If there are several `return`s *and* the last one is not a fallback (but could just as much be amended by further ones), then and only then add a trailing semicolon.

But I agree that this is just a recommendation of clarity.

brianlheim:
This reminds me somewhat of the discussion about trailing commas in C/C++:

https://stackoverflow.com/questions/7043372/int-a-1-2-weird-comma-allowed-any-particular-reason
https://stackoverflow.com/questions/29099517/why-is-an-extra-comma-not-allowed-in-a-parameter-list-when-it-is-allowed-in-a-br
https://stackoverflow.com/questions/6372650/trailing-commas-and-c

[...]

PR #5855, opened 2022-08-28
[classlib] Function:plot fix sample alignment with grid lines

mtmccrea:
## Purpose and Motivation
When plotting functions with a short duration, individual samples are visibly misaligned with the grid. The x-axis position of each sample doesn't account for the duration of the last sample period (a "fencepost problem"). This PR sets the sample position of the plotted points explicitly in the case that this time offset would be visible (i.e. a maximally wide plot would contain samples separated by ~2.5 pixels).

### Example
Below, both `SinOsc` and `Impulse` are plotted for a duration of 2 times their cycle period. So we expect the cycles to start at the leftmost part the plot (first cycle) and the center of the plot (second cycle). 

Currently:
Screen Shot 2022-08-28 at 00 44 20Screen Shot 2022-08-28 at 00 45 28
After the fix:
Screen Shot 2022-08-28 at 00 46 58Screen Shot 2022-08-28 at 00 47 24

Note the differences: 1. the second cycle begins at the center of the plot, 2. the _duration_ of the last sample is properly accounted for on the  time grid. The second point is further clarified when using the `plotMode = \steps` to visualize the sample-and-hold nature of discrete-time sampling:
Screen Shot 2022-08-28 at 00 47 06

Reproduce with ([more examples](https://gist.github.com/mtmccrea/dedb79a48312cd197b042d5c884ab44d)):
```
// SinOsc: the second cycle should begin exactly at
// time t = period (centered in the plot). fs = 48kHz.
(
var period = 0.0005, dur = period * 2;
f = { SinOsc.ar(period.reciprocal) };
f.plot(dur, bounds: Window.screenBounds.insetBy(400, 200)).plotMode_(\plines);
)
// Impulse - Control rate
( 
s.waitForBoot{
	var persamps = 6;
	var period = persamps * s.options.blockSize / s.sampleRate; // n samples per cycle
	var nper = 2;
	var dur = period * nper;
	postf("Expecting an impulse at:\n\ttime %\n\tsample %\n", period, nper.collect(_*persamps+1));

	f = { Impulse.kr(period.reciprocal) };
	f.loadToFloatArray(dur, action: {|arr| a = arr });
	f.plot(dur, bounds: Window.screenBounds.insetBy(400, 200)).plotMode_(\plines);
}
)
```

This PR also sneaks in a slight padding of x-grid tick labels, which was needed in some cases when the bounding rect of a `String` underestimates the width of the string (second commit).

## Types of changes

- Bug fix.

## To-do list

- [x] Code is tested
- [x] All tests are passing (`TestPlotter`)
~~[ ] Updated documentation~~

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -1128,31 +1128,52 @@ Plotter {
 
 + Function {
 	plot { |duration = 0.01, target, bounds, minval, maxval, separately = false|
+		var server, plotter, action;
+		var name = this.asCompileString;
 
-		var name = this.asCompileString, plotter, action;
-		if(name.size > 50 or: { name.includes(Char.nl) }) { name = "Function plot" };
+		if(name.size > 50 or: { name.includes(Char.nl) }) { name = "function plot" };
+
 		plotter = Plotter(name, bounds);
+		// init data in case function data is delayed (e.g. server booting)
 		plotter.value = [0.0];
+
 		target = target.asTarget;
+		server = target.server;
 		action = { |array, buf|
 			var numChan = buf.numChannels;
-			{
+			var numFrames = buf.numFrames;
+			var frameDur;
+
+			defer {
 				plotter.setValue(
-					array.unlace(numChan).collect(_.drop(-1)),
+					array.unlace(numChan),
 					findSpecs: true,
 					refresh: false,
 					separately: separately,
 					minval: minval,
 					maxval: maxval
 				);
+
 				plotter.domainSpecs = ControlSpec(0, duration, units: "Time (s)");
-			}.defer
+
+				// If individual values might be separated by at least 2.5 px
+				// (based on a plot at full screen width), set the x values (domain)
+				// explicitly for accurate time alignment with grid lines.
+				if(numFrames < (Window.screenBounds.width / 2.5)) {
+					frameDur = (this.value.rate == \control).if({
+						server.options.blockSize / server.sampleRate;
+					}, {
+						1 / server.sampleRate;
+					});
+					plotter.domain = numFrames.collect(_ * frameDur);
+				};
+			}
 		};
 
-		if(target.server.isLocal) {
-			this.loadToFloatArray(duration, target, action:action)
+		if(server.isLocal) {
+			this.loadToFloatArray(duration, target, action: action)
 		} {
-			this.getToFloatArray(duration, target, action:action)
+			this.getToFloatArray(duration, target, action: action)
 		};
 
 		^plotter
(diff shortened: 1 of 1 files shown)


SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -1128,31 +1128,52 @@ Plotter {
...
 		action = { |array, buf|
 			var numChan = buf.numChannels;
-			{
+			var numFrames = buf.numFrames;
+			var frameDur;
+
+			defer {
 				plotter.setValue(
-					array.unlace(numChan).collect(_.drop(-1)),
+					array.unlace(numChan),
```
mtmccrea, line 1149: Note: previously, the last data frame was dropped from in the plot. It's not apparent to me why, and seems to work fine plotting _all_ data frames.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -1128,31 +1128,52 @@ Plotter {
...
 				);
+
 				plotter.domainSpecs = ControlSpec(0, duration, units: "Time (s)");
-			}.defer
+
+				// If individual values might be separated by at least 2.5 px
+				// (based on a plot at full screen width), set the x values (domain)
+				// explicitly for accurate time alignment with grid lines.
+				if(numFrames < (Window.screenBounds.width / 2.5)) {
+					frameDur = (this.value.rate == \control).if({
```
telephon, line 1163: `frameDur = if(this.value.rate == \control) {`  would keep one style
Also no need for semicolons for the return of the functions (e.g. for `1 / server.sampleRate;`)

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
telephon: yes, this seems much better.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
telephon: (but this is just a suggestion)

mtmccrea: Thanks for suggesting this. I thought at the time that setting a variable with receiver syntax read more naturally, but I agree consistency is better.

After reviewing the [discussion](https://github.com/supercollider/supercollider/issues/2914) that led to return statement [guidelines](https://github.com/supercollider/supercollider/wiki/Code-style-guidelines#return-statements), I removed `;` from both return statements in the `if`, and added one after `defer` (that statement doesn't imply return and could be followed by more code).

telephon (approved):

PR #5856, opened 2022-08-29
class library: remove inline warnings in common

telephon:
## Purpose and Motivation

There should be no uninlinable variables in common, it creates inline warnings

## Types of changes

- Bug fix

## To-do list


dyfer (approved):
Thanks!

PR #5861, opened 2022-09-01
HistoryGui: improve display readability (fontsize and tabWidth)

adcxyz:
## Purpose and Motivation

HistoryGui has not been updated for Qt Guis, and thus had visual problems: 
Too small font size and too large tabWidth. 
Also, the window name was a bit confusingly "HistoryGui_History", 
which this PR clarifies to just say "History".
 
- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Core/HistoryGui.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Core/HistoryGui.sc
@@ -32,12 +32,14 @@ HistoryGui : JITGui {
 		);
 	}
 
+	winName { ^"History" }
+
 	makeViews { |options|
 		var font, flow;
 		var textViewHeight = numItems * skin.buttonHeight;
 		var listViewHeight = textViewHeight * 1.6;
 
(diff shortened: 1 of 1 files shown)


SCClassLibrary/Common/GUI/PlusGUI/Core/HistoryGui.sc
```diff
@@ -66,6 +68,8 @@ HistoryGui : JITGui {
 				};
 			};
 		})
+		.font_(font)
+		.tabWidth_(font.size * 0.6 * 4) // ca. 4 chars
```
telephon, line 72: Maybe better: 
`"1234".bounds(font).width`?

telephon: `tabWidth_` seems to be a pixel size, according to documentation

adcxyz: yes, that is better, done.

telephon (approved):

Issue #5879, opened 2022-09-28
Pattern.record appears to be broken


jamiehodge:
The inlining of `pattern.record` (https://github.com/supercollider/supercollider/pull/5275) is resulting in header-only files. I copied the previous, server recorder-based method into the class definition and it created a valid audio recording.

## Environment

* SuperCollider version: 3.12.2
* Operating system: Mac OS
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
pattern.record(dur: 10)
```

## Expected vs. actual behavior

AIFF file with 100 or so bytes of headers and no recording.

jamshark70:
```
(
Pbind(
	\degree, Pn(Pseries(0, 1, 8), 2),
	\dur, 0.25
).record("~/p-rec-test.wav".standardizePath, "wav", "int16", 2);
)
```

![sc-pat-rec-01](https://user-images.githubusercontent.com/318301/192912501-7a6db1e3-ca02-4e8f-a341-307b6d200b0c.png)

So the inlining is not broken. The recording logic is fine. (Actually the inlining was done for a reason -- you already pointed to the issue; it would be worth reading the history to understand why it was changed.)

What *does* appear to be a problem is that there is no default header or sample format.

So the fix would be:

```
	record { |path, headerFormat, sampleFormat, numChannels = 2, dur = nil, fadeTime = 0.2, clock(TempoClock.default), protoEvent(Event.default), server(Server.default), out = 0, outNumChannels|

		var buffer, nodeID, defname;
		var pattern = if(dur.notNil) { Pfindur(dur, this) } { this };

		if(headerFormat.isNil) {
			headerFormat = server.recHeaderFormat;
		};
		if(sampleFormat.isNil) {
			sampleFormat = server.recSampleFormat;
		};

		server.waitForBoot {
... etc etc...
		}
	}
```

telephon:
Btw. was there a reason that it wasn't possible to use the built in `Recorder` class for this?

jamshark70:
> was there a reason that it wasn't possible to use the built in Recorder class for this?

This information is not hard to find in the publicly available history. The link is right there at the top of the issue text...

https://github.com/supercollider/supercollider/pull/5275: "https://scsynth.org/t/calling-supercollider-script-from-windows-batch-file/2964/8 reports that an audio file produced by pattern.record begins with some silence, where a reasonable expectation would be that the recording begins exactly at the onset of the pattern's audio."

[...]

PR #5883, opened 2022-10-10
fix: Pattern.record

jamiehodge:
## Purpose and Motivation

Fixes https://github.com/supercollider/supercollider/issues/5879

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested (locally)
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Streams/Patterns.sc
+++ b/SCClassLibrary/Common/Streams/Patterns.sc
@@ -101,6 +101,14 @@ Pattern : AbstractFunction {
 		var buffer, nodeID, defname;
 		var pattern = if(dur.notNil) { Pfindur(dur, this) } { this };
 
+		if(headerFormat.isNil) {
+			headerFormat = server.recHeaderFormat;
+		};
+
+		if(sampleFormat.isNil) {
+			sampleFormat = server.recSampleFormat;
+		};
+
 		server.waitForBoot {
 			var group, bus, free, monitor;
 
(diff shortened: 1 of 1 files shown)


telephon (commented):
Looks good!

You should leave out the semicolon within the function.


You could also write it in short:
```
recHeaderFormat = recHeaderFormat ?? { server.recHeaderFormat };
```

but that is a matter of style.

jamiehodge:
Hi Julian! Long time no see.

Updated, hopefully correctly.

SCClassLibrary/Common/Streams/Patterns.sc
```diff
@@ -101,6 +101,9 @@ Pattern : AbstractFunction {
 		var buffer, nodeID, defname;
 		var pattern = if(dur.notNil) { Pfindur(dur, this) } { this };
 
+        headerFormat = headerFormat ?? { server.recHeaderFormat };
+		sampleFormat = sampleFormat ?? { server.recSampleFormat };
```
jrsurge, line 105: is there an advantage here of `??` over `?`? these operators always confuse me, trying to learn how/when is best to pick one over the other - any advice greatly appreciated!

dyfer: I think ?? evaluates the function on the right, while ? doesn't, just using that value directly. I might be wrong though.

telephon: In `?`, the call `server.recSampleFormat` is always made, even if it isn't needed. With `??`, it is only made if `headerFormat` is nil.

dyfer:
@telephon do you think this good to go?

SCClassLibrary/Common/Streams/Patterns.sc
jrsurge: Thanks, thought it might be something like that!

telephon (approved):

PR #5888, opened 2022-10-19
Fix NamedControl example code

chkworks:
Refactors the examples in the help file to working code. Before an exception was thrown because the NamedControl was instanced with multiple different default values.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/NamedControl.schelp
+++ b/HelpSource/Classes/NamedControl.schelp
@@ -5,7 +5,7 @@ related:: Classes/ControlName, Classes/Control
 
 description::
 
-A NamedControl directly combines a ControlName and a Control UGen conveniently. Also this makes it safe even if several identical controls exist (see example below).
+A NamedControl directly combines a ControlName and a Control UGen conveniently. It can be saved to a var to send the control value to different parts of the function or SynthDef.
 
 There are syntax shortcuts that generate NamedControls from the name:
 code::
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/NamedControl.schelp
```diff
@@ -5,7 +5,7 @@ related:: Classes/ControlName, Classes/Control
 
 description::
 
-A NamedControl directly combines a ControlName and a Control UGen conveniently. Also this makes it safe even if several identical controls exist (see example below).
+A NamedControl directly combines a ControlName and a Control UGen conveniently. It can be saved to a var to send the control value to different parts of the function or SynthDef.
```
telephon, line 8: Unfortunately, this is a misunderstanding. The sentence "Also this makes it safe even if several identical controls exist (see example below)." meant that you can use the same symbol several times.

Your attempt to correct it indeed indicates that it isn't clear enough, so thank you.
Would you prefer to improve or start a new one?

HelpSource/Classes/NamedControl.schelp
```diff
@@ -58,8 +58,12 @@ a.setn(\freq, [0, 2, 5].midiratio * 400);
 a = { SinOsc.ar(\freq.kr([300, 330, 370], [1, 0.3, 0.02])).sum * 0.1 }.play;
 
 // multiple usage of the same name:
-a = { SinOsc.ar(\freq.kr(440, 3.5)) + Saw.ar(\freq.kr(440, 0.05) * 0.5) * 0.1 }.play;
-
+(
```
telephon, line 61: The example is indeed better code, but doesn't fulfil its purpose to demonstrate the feature.

chkworks:
All valid points. I just noticed that most of the example code did not work, because of different default values being used in the same control. I added another subsection that shows how to deal with situations, where you want to apply different lag values to the same control and how you would store it into a varible. Is this clearer now?

telephon (approved):

PR #3733, opened 2018-05-17
exclude default paths: change from command line parameter to language file flag

miguel-negrao:
Before this PR:
 - default paths were excluded from the list of directories to search for class files if sclang was started with -a command line option.
- The "Exclude default search paths" option on the interpreter options in scide would toggle the use of `sclang -a`.

with this PR:
 - An additional method is added to exclude the default paths excluded from the list of directories to search for. To turn on or off excluding the default paths the 'excludeDefaultPaths' key can be used in the YAML language file.
 - The reasoning behind this PR is that some language config files require the default include paths to be active while others require the default include paths not to be active, in order for the language to compile successfully, therefore it makes sense that that choice is made in the yaml file itself, so that it will work automatically. This way switching to a different language config file will also switch whether default paths are excluded or not.
- The behaviour of `sclang -a` is kept the same.
- The "Exclude default search paths" option on the interpreter options in scide, doesn't use `sclang -a` anymore, instead just mirrors the value of 'excludeDefaultPaths' key  in the YAML language file. Before, with this PR the default paths were excluded for ALL language config files, now the choice is done in each YAML language file.
- A getter and setter were added for excludeDefaultPaths in the LanguageConfig class library class.
- In PyrLexer.cpp changes were made such that when recompiling the library the in-memory version of the LanguageConfig will be used instead of the version in the YAML file. This allows testing language settings by manipulation through the language (via LanguageConfig.sc) before writing to the file.

An example config file that uses a Class Library other than the system one:

```
includePaths:
    -   /home/miguel/Development/SuperCollider/supercollider/SCClassLibrary
excludePaths:
    []
postInlineWarnings: false
excludeDefaultPaths: true
```

PR #3278, opened 2017-11-05
SuperCollider projects V2

miguel-negrao:
This is a revised version of PR #1978. I've updated the code taking into account the changes added in #2861. The code has also been substantially changed with some features added.

## Overview
A supercollider project is an sclang config file which has an extra field "project: true". When this field is present sclang behaves differently:

- Relative paths in the config file are resolved relative to the directory containing the config file.
- Quarks download to the directory containing the config file.
- If present in the same directory as the config file a startup.scd is run at startup. The global startup.scd is not run.

This allows:
- Having self-contained projects which can be shared with other users by just sharing the directory containing the project file.
- Creating sclang standalones, that is, applications based on on supercollider and with a sclang Qt gui which can be installed and shared. (Currently the part of doing a system install or having a MacOS bundle is not implemented).

### general

- It is recommended that when using a project to also activate the exclude default paths option, but this was not hardcoded for flexibility.
- The extension for project config files is by default .scproj

### sclang

- excluding default paths was removed from sclang command line options and moved to the config file. This introduces a change in the command line options of sclang, removing the -a option. The rationale is that with -a a lang file has to be used anyhow otherwise the class library will not compile as it doesn't find an SCClassLibrary, therefore might as well make that an option which is settable in the language config file which also makes it easier to manage. This also changes the header API to the state it was in before pull-request #1021, it is unfortunate going back and forth, but I do think it is better as it is in this pull-request.
- readLibraryConfigYAML: code was simplified abstracting repeated code using lambdas. In the future more simplification could be achieved by sharing this code between the ide and sclang, although there are issues when templated code interacts with lambdas in C++11.
- PyrLexer: gResourceDirectory, gSystemExtensionDirectory, gUserExtensionDirectory are initialized once in initPassOne(). In passOne_ProcessDir the corresponding strings are substituted by the actual paths and relative paths are converted to absolute (using the path of the language config file).
- Platform.sc: in project mode, if a start.scd file is present in the same directory as the config file it is used at startup instead of global start.scd file (which in any case is not used if exclude default paths is activated).

### IDE

- An option was added to store the language config in the sessions (disabled by default). This can be changed in the settings dialog. It allows having a different project or config file associated with each session.
- Changing from storing lang in session or settings will update dialog and guis while still in the dialog (without doing apply). The method void Dialog::switchSclangPage(bool useLanguageConfigFromSession) is used in the preference panes because that choice is still not commited to the QSettings object.
[...]

PR #5898, opened 2022-11-02
New standalone approach

muellmusik:
## Purpose and Motivation

This PR demonstrates a new light approach to standalones. Goals:

- lightweight and easy to use
- potentially cross-platform
- will (when distributed as a signed app) solve issues with code-signing and modifying app bundles on recent MacOS

Currently MacOS only, but could be cross-platform fairly easily I think. (I might need some help with that if anyone is willing!) When configuring Cmake add -DSTANDALONE=ON to build in Standalone mode.

Folders ```Standalone``` and ```Standalone/Extensions``` are created adjacent to the app bundle and the latter added to lang and server paths. Users can simply add all code, resources and extensions needed there. (Thus no need to modify bundle or place resources elsewhere.) Normal extension directories are excluded.

I'm offering this for comment at this point, but I would like to get something relatively similar to this merged soon. I think this is a good solution, and avoids the ongoing problems with other standalone approaches. We can add bells and whistles later.

N.B. This slightly modifies the -a flag to sclang ('standalone mode'), which previously just excluded _all_ default folders including the default lib. I've asked the author about this with no response, but it seems likely this never worked well as a standalone strategy, since it would depend on a config file, which cannot include relative paths. See discussion here: https://scsynth.org/t/lighter-approach-to-standalones/5171/23

## Types of changes

- Documentation
- Bug fix (well, new approach to something broken)
- New (version of a) feature

## To-do list

- cross-platform support
- supernova support
- building still creates an Install/SuperCollider folder and an empty SuperCollider bundle. I can't quite find where this happens, so any advice would be gratefully received!

- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation


telephon:
A very good approach, I'm very much in favour of this. So one would still compile a version for standalone? Or can this be configured from within sclang?

muellmusik:
My idea is that SC Standalone could just be downloaded from GH and then customised by users. That way it’s code signed. Of course advanced users may still want to build their own, but the features added here can serve as the basis for that. (Anyone with a full dev account can of course  code sign standalones with modified bundles themselves, but that excludes most users.)

telephon:
I see, ok that makes sense.

muellmusik:
MacOS binary for people to try here: https://github.com/muellmusik/supercollider/releases/tag/v0.1-test-binary

telephon:
I get this (also with cmd-click):

Screenshot 2022-11-03 at 13 35 49

muellmusik:
Crap. I will investigate...

muellmusik:
Oh looks like a build issue. Not copying in the Frameworks.

dathinaios:
Thanks for working on this! Just to let you know that I am getting the same issue as @telephon on Mac Studio with MacOS Monterey.

muellmusik:
Yeah it looks like something I need to update since STANDALONE=OFF builds seem fine in terms of this.

muellmusik:
The build does work on my machine...

muellmusik:
Okay try it now, same link: https://github.com/muellmusik/supercollider/releases/tag/v0.1-test-binary

telephon:
Hm, still the same here.

muellmusik:
> Hm, still the same here.

Tested it on another machine and I get different error, that it's using a too new SDK. Hmm...

telephon:
I am on 12.6 (21G115)

muellmusik:
I am working on it. In the meantime found another little Cmake issue. Will force push again shortly.

muellmusik:
Okay actually think I'm done for tonight. There're two things: One is a build problem caused by the change I made, another is a codesigning issue. Bleh. Try again tomorrow...

telephon:
Thank you, Scott – looking forward, it'll solve a long standing issue!

dyfer:
Thank you @muellmusik for working on this! I wanted to revisit building with the standalone flag for quite some time.

> My idea is that SC Standalone could just be downloaded from GH and then customised by users. That way it’s code signed.

Well, we'd have to sign it the way we sign the release (signing does not happen automatically). But I do agree this is the best approach - to offer it with the release as a separate download.

> This slightly modifies the -a flag to sclang ('standalone mode'), which previously just excluded all default folders including the default lib

This needs to be carefully considered.

On one hand - "standalone mode" really should mean that the bundled class library is still included.
OTOH it is practical to be able to exclude all the paths (e.g. for testing the whole class library in the development folder). Changing the behavior of -a would be a breaking change, but maybe that's okay? Again, needs to be considered carefully. 
There was some previous discussion on these options here: https://github.com/supercollider/supercollider/pull/3733

muellmusik:
> Okay try it now, same link: https://github.com/muellmusik/supercollider/releases/tag/v0.1-test-binary

Okay try again now. I had to manually dequarantine it. It ran but was translocating so didn't find the standalone folder.

```xattr -d com.apple.quarantine /path/to/SuperCollider-Standalone```

Or just build yourself!

**Comment by muellmusik** [comment]
> Well, we'd have to sign it the way we sign the release (signing does not happen automatically). But I do agree this is the best approach - to offer it with the release as a separate download.

Yes, sorry. That's exactly what I had in mind!

> This needs to be carefully considered.
> 
> On one hand - "standalone mode" really should mean that the bundled class library is still included. OTOH it is practical to be able to exclude all the paths (e.g. for testing the whole class library in the development folder). 
> Changing the behavior of -a would be a breaking change, but maybe that's okay?

So in my version as now, it includes the bundled class library, but excludes system and user extension directories. I have left the existing 'exclude default search paths' option via LangConfig and GUI as is, because I think changing that could break some people's setups.

In a sense, changing -a is a 'breaking' change, but Miguel confirmed that the implementation was never really complete, so I don't think that should cause any problems at all.

**Comment by muellmusik** [comment]
> Thank you @muellmusik for working on this! I wanted to revisit building with the standalone flag for quite some time.

IAC I'm very glad you're interested, @dyfer! I could use a little help on some things if you have the space. (No rush.)

**Comment by telephon** [comment]
Great, this works now!

File: CMakeLists.txt [hunk 1]
```diff
@@ -8,7 +8,12 @@ endif()
 
 include(SCVersion.txt)
 # CMake versions can only consist of digits and periods, so don't use _TWEAK
-project(SuperCollider VERSION ${SC_VERSION_MAJOR}.${SC_VERSION_MINOR}.${SC_VERSION_PATCH})
+if(STANDALONE)
+	project("SuperCollider-Standalone" VERSION ${SC_VERSION_MAJOR}.${SC_VERSION_MINOR}.${SC_VERSION_PATCH})
```
telephon, line 12: could there be a variable that can be set that will allow us to set the name of the application? This could be "SuperCollider-Standalone" by default, but could be overridden.

dathinaios:
It is working for me now 🙂 But I do get an error:

```supercollider
'/quit' message sent to server 'localhost'.
compiling class library (debug build)...
	Found 857 primitives.
	Compiling directory '/private/var/folders/8m/ws36w8lj59n55wslmf9mfqcw0000gn/T/AppTranslocation/4208ECF0-41DC-4F5C-A52D-8EDD39E0FC34/d/SuperCollider-Standalone.app/Contents/Resources/SCClassLibrary'
	numentries = 839301 / 12499812 = 0.067
	5526 method selectors, 2262 classes
	method table size 13516568 bytes, big table size 99998496
	Number of Symbols 12409
	Byte Code Size 375171
	compiled 328 files in 0.23 seconds
compile done
Class tree inited in 0.02 seconds


*** Welcome to SuperCollider 3.13.0-dev. *** For help press Cmd-D.
ERROR: Message 'startStandalone' not understood.
RECEIVER:
Instance of Main {    (0x7fa3b8050058, gc=8C, fmt=00, flg=00, set=04)
  instance variables [11]
    classVars : instance of Array (0x7fa3d841ca40, size=471, set=9)
    interpreter : instance of Interpreter (0x7fa3b8050428, size=30, set=5)
    curThread : instance of Thread (0x7fa3b8050188, size=28, set=5)
    mainThread : instance of Thread (0x7fa3b8050188, size=28, set=5)
    schedulerQueue : instance of Array (0x7fa3b8070080, size=1, set=12)
    nowExecutingPath : nil
    platform : instance of OSXPlatform (0x7fa3b8050868, size=9, set=4)
    argv : nil
    recvOSCfunc : nil
    prRecvOSCFunc : nil
    openPorts : instance of Set (0x7fa3c0120fb8, size=2, set=2)
}
ARGS:
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg this = <instance of Main>
		arg selector = 'startStandalone'
		arg args = [*0]
	Main:startup
		arg this = <instance of Main>
		var didWarnOverwrite = false
^^ ERROR: Message 'startStandalone' not understood.
RECEIVER: a Main
```
Also, is there an option to skip the IDE completely?

muellmusik:
@dathinaios did you manually de quarantine as described above?

dathinaios:
I hadn't no. But I just checked and tried to run:

```
xattr -d com.apple.quarantine /Applications/SuperCollider-Standalone.app
```
But got
```
xattr: /Applications/SuperCollider-Standalone.app: No such xattr: com.apple.quarantine
```

muellmusik:
Hmm. startStandAlone issue went away after de quarantine for me, so I assumed it was a translocation problem

muellmusik:
See if it has any other attributes

dathinaios:
Ah ok, I erased everything and downloaded again and the de quarantine worked fine 👍 Thanks!

dyfer:
> In a sense, changing -a is a 'breaking' change, but Miguel confirmed that the implementation was never really complete, so I don't think that should cause any problems at all.

Ah, sorry, I want to retract what I wrote before. My approach for testing default class library from another location only uses the configuration file settings (accessed through `LanguageConfig` class), and not the `-a` flag. So IIUC this PR does not break the setup that I described.

I would personally be tentatively okay changing the behavior of `-a`, but we could also consider another letter, as suggested by @telephon.
> It ran but was translocating so didn't find the standalone folder.

FWIW SC in general used to run fine when translocated, so I think we should fix this so that it works whether the app is translocated or not. I presume that's possible, but I might be wrong.

miguel-negrao:
> . I've asked the author about this with no response, but it seems likely this never worked well as a standalone strategy, since it would depend on a config file, which cannot include relative paths

I can confirm this is the case (sorry for the long delay in answering !! :-( ).  I gave it a go at adding support for relative paths (and other things) in this PR: https://github.com/supercollider/supercollider/pull/3278 . My approach there was to remove the "-a" option and move that flag into the language config file. The reasoning was that with "-a" a language config file is always needed anyway.  You can find more info about my approach in the PR.

I'm glad that Scott has picked this up again.

miguel-negrao:
@muellmusik , how will this approach work on linux ? On Linux the class library is installed system-wide. Before this change the "-a" option excluded the system-wide class library, which allow on Linux at least creating non-distributable standalones. They are not distributable because they have absolute paths in the language config file, but otherwise they work. I believe this change will break that setup for linux users (like myself) that rely on that feature. Is there a way to avoid that ? Perhaps an additional flag not to load the default class library ? That flag could go into the language config file, it doesn't have be a command line option.

muellmusik:
> The reasoning was that with "-a" a language config file is always needed anyway. You can find more info about my approach in the PR.

That makes sense to me, tbh.

muellmusik:
> how will this approach work on linux ?

I guess on Linux, you would have to put a local class library adjacent to the binary, as well as the extensions. I've not written the code for Linux as I don't have a machine to test on, so could use some help with this. 

In the current form, this retains the existing flag in the LangConfig to ignore the default class library, so I don't think it breaks anything. For your 'non-distributable standalones' usage, you would probably use vanilla SC builds rather than the 'standalone' version though.

muellmusik:
> FWIW SC in general used to run fine when translocated, so I think we should fix this so that it works whether the app is translocated or not. I presume that's possible, but I might be wrong

Yes, I'm hoping this will work fine when code signed, but I'm a bit new to all that!

CMakeLists.txt
muellmusik: There was a stub for this (may not have really worked) previously.

miguel-negrao:
Currently this doesn't build on Linux, there's a dependency on code that only exists for macos:

```
[ 27%] Linking CXX executable scsynth
/usr/bin/ld: libscsynth.a(SC_Lib_Cintf.cpp.o): in function `SC_Filesystem::defaultStandaloneExtensionDirectory()':
/home/miguel/Development/SuperCollider/supercollider/common/SC_Filesystem.hpp:294: undefined reference to `SC_Filesystem::defaultStandaloneDirectory()'
/usr/bin/ld: libscsynth.a(SC_Lib_Cintf.cpp.o): in function `SC_Filesystem::defaultDirectory(SC_Filesystem::DirName const&)':
/home/miguel/Development/SuperCollider/supercollider/common/SC_Filesystem.hpp:256: undefined reference to `SC_Filesystem::defaultStandaloneDirectory()'
collect2: error: ld returned 1 exit status
make[2]: *** [server/scsynth/CMakeFiles/scsynth.dir/build.make:129: server/scsynth/scsynth] Erro 1
make[1]: *** [CMakeFiles/Makefile2:3369: server/scsynth/CMakeFiles/scsynth.dir/all] Erro 2
make: *** [Makefile:182: all] Erro 2
```

miguel-negrao:
I've had a look at the code. So in indeed this changes the "-a" command line option, but that is not really a big issue as that functionality had already been implemented as a lang file option (excludeDefaultPaths), so no functionality is lost I think in Linux. 

On the other hand to make this new standalone mode work in a similar way in Linux, a different approach is needed to determine the location of the "Standalone" folder. It could be the directory where the language file is placed (SC_LanguageConfig::getConfigFileDirectory()).

CMakeLists.txt
telephon: OK, that is a stumbling block, I suppose.

telephon:
@miguel-negrao so you think this can be a two step process, first just a macOS (and windows?) solution, and then separately, a linux one? Then it must be sure that the current interface will be compatible with the future linux version.

miguel-negrao:
> @miguel-negrao so you think this can be a two step process, first just a macOS (and windows?) solution, and then separately, a linux one? Then it must be sure that the current interface will be compatible with the future linux version.

It's probably a good a idea to think of the solution to both cases at the same time, indeed, but it is not strictly required.

For Linux I would start with a script such as this: 

```BASH
#!/bin/bash
SCRIPT_DIR=`dirname "$0"`
#let scide pick the right sclang binary
export PATH="$SCRIPT_DIR":$PATH
#don't let scide write in the user config directory
export XDG_CONFIG_HOME="$SCRIPT_DIR/config"
export XDG_DATA_HOME="$SCRIPT_DIR/share"
"$SCRIPT_DIR/scide" -style=gtk
```

This would redirect the .local/share and .config directories into the the standalone folder, and the PATH would pick up the local sclang and scsynth. Then a bit of C++ code is needed. The problem is that the language config file uses absolute paths, so we cannot use it, it must stay empty. One approach could be to disable compiling the /usr/local/share/SuperCollider library but still compile "	~/.local/share/SuperCollider/Extensions/" which through the redirection via the variable XDG_DATA_HOME would be the one in the standalone and not the global one, when using "-a" on the command line.  We also need to make sure scsynth doesn't pick anything in the /usr/local/share/SuperCollider directory. I would have to check but I think this can be overridden with something like `s.options.ugenPluginsPath = "%/plugins".format(LanguageConfig.currentDirectory).postln;`

I could have a go at this more towards new year's eve.

dyfer:
IMO it would be worth considering adding support for some sort of relative paths in the sclang config file as part of the work on the standalone. The details would need to be carefully considered, but _maybe_ one option would be to allow wildcards like e.g. `%SCLANG_DIR%` or similar in sclang config which would resolve to the location of the sclang binary, or we could define relative paths to be always relative to sclang binary.

miguel-negrao:
> IMO it would be worth considering adding support for some sort of relative paths in the sclang config file as part of the work on the standalone. The details would need to be carefully considered, but _maybe_ one option would be to allow wildcards like e.g. `%SCLANG_DIR%` or similar in sclang config which would resolve to the location of the sclang binary, or we could define relative paths to be always relative to sclang binary.

I agree, that would be a more complete solution, but also more work.

It appears the path of the executable can be obtained with boost in a cross-platform way https://www.boost.org/doc/libs/1_65_1/doc/html/boost/dll/program_location.html, if the paths were to be resolved relative to sclangs's binary path.

miguel-negrao:
For Linux, as further goal could be for the app to work on any Linux distribution. This can be done for instance with [appimage](https://docs.appimage.org/packaging-guide/manual.html), which looks quite simple. Importantly appimage applications allow "portable mode" which does [the same thing](https://docs.appimage.org/user-guide/portable-mode.html) as what I'm doing in the script above with the XDG directories, to keep the user files local to the portable application.

muellmusik:
> Currently this doesn't build on Linux, there's a dependency on code that only exists for macos:

Yes, I haven't written the folder lookup code for Linux or windows, again as I don't machines to test on.

muellmusik:
> On the other hand to make this new standalone mode work in a similar way in Linux, a different approach is needed to determine the location of the "Standalone" folder. It could be the directory where the language file is placed (SC_LanguageConfig::getConfigFileDirectory()). In any case, that doesn't need to block the effort on the macos side.

Since we want to support coexisting standalones, I think it would be a mistake to rely on anything stored in some sort of global preferences, like a particular language config. The whole idea is it's independent of global preferences, I think.

muellmusik:
> It's probably a good a idea to think of the solution to both cases at the same time, indeed, but it is not strictly required.

I think it would be good to have a plan for everything, but nice if one platform doesn't have to hold back any other. (Too many things wait forever in sc dev 😉)

> For Linux I would start with a script such as this:

I can see that a startup script might be nice, though I'd prefer something SC based then shell + env vars, so that it's not awkward to use on platforms where it's rare to start from cl. Again my preference is to keep this as simple as possible in the first instance. Does a standalone need on the fly reconfiguration? My instinct is no, so I'd want to make sure any additional complexity is worth it. My 2 cents, anyway.

muellmusik:
> IMO it would be worth considering adding support for some sort of relative paths in the sclang config file as part of the work on the standalone.

I think it's maybe best as a separate issue. Again if we have a generic signed standalone app, the lang config would be global, so I don't think it can or should use that. That's perhaps only an issue on Mac (Windows?) but I think any proliferation of approaches varying across platforms should clearly justify the technical debt and inconsistency.

muellmusik:
> It appears the path of the executable can be obtained with boost in a cross-platform way https://www.boost.org/doc/libs/1_65_1/doc/html/boost/dll/program_location.html, if the paths were to be resolved relative to sclangs's binary path.

Oh thanks, @miguel-negrao, that's really nice. That should hopefully reduce the need for platform specific folder location code, or at least generalise it considerably!

muellmusik:
> I think any proliferation of approaches varying across platforms should clearly justify the technical debt and inconsistency.

Just to be clear, I don't mean to be dismissive of anything or ignore important aspects of platform specific workflows, just that we might tread carefully. If we get a simple approach working and out there, more complex things can always be added if they're needed or prove desirable.

telephon:
OK, so let's do the macOS first and then the rest can follow later. Indeed it is a platform specific thing, after all.

Is there anything that still needs to be done?

miguel-negrao:
I gave it a go at appimage on linux and it looks very promising. Some of its tools can even rewrite hardcoded paths on binaries, so that no c++ code changes would be needed, if one wants to avoid that. I still don't have it working 100%, because of hard coded paths (to /usr ) and because of picking the local sclang and scsynth (one needs to manipulate the path), but none of these issues seem like a show stopper.  The executable stands at ~200MB and inside has all the libraries that it needs, as well as the documentation, plugins, etc.  I still haven't tried appimage-builder which can possibly take care of these issues mostly automatically. If this approach works, then possibly it's not even needed to change any C++ code, this would just be part of the CI which would build an appimage version of SuperCollider available for download. Then if there is a folder with the same name of the executable but with ".home" at the end, the app becomes portable and doesn't look at anything in $HOME, becoming a "standalone" in the SuperCollider sense.

dyfer:
> Is there anything that still needs to be done?

Making regular non-macOS builds work would be a good start ;)

miguel-negrao:
I think appimage is defintely doable. I've managed in the meantime to get the PATH working so that it picks the sclang and scsynth inside the appimage. My plan is to next use the standalone CMAKE/define flag to alter the paths in sclang and scsynth which point to /usr and change them to $APPDIR/usr which will pick the resources inside the appimage file (the equivalent of the macos bundle). Altering the paths of scsynth seems to me better so that people don't have to mess with the -U option and can just use the exact same code they had on a non-standlone version of SC. When I manage to get something working I will make a pull-request. Btw, i've noticed that scsynth is opening qt5 files in /usr (via /proc/pid/fd) which I've found surprising.

miguel-negrao:
So I would like to sketch how the standalone approach to linux would work and get feedback from you all.

Using strace I've compiled a list of paths accessed by scide, sclang and scsynth:

scsynth:

$HOME/.local/share/SuperCollider
/usr/share/SuperCollider
/usr/lib/SuperCollider

sclang:

$HOME/.local/share/SuperCollilder
$HOME/.sclang.sc
$HOME/.config/SuperCollider
/usr/share/SuperCollider

SCIDE

$HOME/.config/SuperCollider
$HOME/.local/share/scide

My proposal is:

- Use the linux app image system to compile a standalone version of SuperCollider which works on any Linux distro.
- The appimage would contain all the stuff in /usr inside the image file (the equivalente of the macos bundle). 
- The places in C++ code which refer to the paths in /usr must be changed when compiled with a specific cmake flag and C++ #define. This will require changes in C++ in sclang and scsynth. The path would be changed to $APPDIR/usr. The appimage fills the APPDIR env variable with the path where these files get decompressed to on runtime.
- The app image automatically remaps the $HOME env variable when a folder named the same as the appimage executable is present with name ending in ".home". Therefore, nothing needs to be done regarding paths in home for a linux standalone, in particular there is no need to use a command line standalone mode, or do any changes in sclang, etc. Users can just drop their code in "SuperCollider.appimage.home/.local/share/SuperCollilder".

Since for Linux we will always need to use the appimage portable mode via the "SuperCollider.appimage.home", I don't see much sense in implementing the code to look for a folder named "standalone" next to the appimage as is done for macos. If we did that then we would require two folders to be present ("SuperCollider.appimage.home" and "Standalone"). What do you all think ? But then the Linux implementation would be different enough, that perhaps it should use a different cmake/define flag such as APPIMAGE and not use STANDLONE ?

dyfer:
@muellmusik @miguel-negrao thank you both for your work on this (and #5938).
While this is still not a full-fledged review, I think I'd like us to unify the path handling between the two approaches, before proceeding further and hashing out platform specific details (platform specific cmake changes, bundle creation, github actions etc):
- while it's a technicality, I think we could get away with just the "STANDALONE" CMake for all platforms, that would do slightly different things on each platform; on Linux we'd assume this is for the AppImage approach for the time being
- I think the main structural difference is that in this PR there's a concept of "standaloneExtDir" etc, with additional methods in sclang, while #5938 modifies search paths to look inside the AppImage in a way that's transparent/invisible to sclang. I can't decide which one I'd prefer, but this is the point that I'd like to see unified between the two approaches first.

> Isn't it a problem for the macos standalone bundle that scide will look at stuff in $HOME ? Or it doesn't ?

IMO it probably shouldn't look at stuff in `$HOME` in the standalone mode.

I'll do my best to loop back to this in the next few weeks.

miguel-negrao:
@muellmusik please correct me if I'm wrong but this is my understanding of how SC in mac and linux work, and therefore the differences in approach for the standalone version.

1) Regarding SCClassLibrary, Docs and scsynth plugins:
- Macos: these are already inside the normal macos bundle.
- Linux: in the normal build these are installed system-wide. In AppImage version these are put inside an appimage executable, similarly to a macos bundle, and the c++ code changes in my PR make sclang and scide look inside of the appbundle instead of in /usr.

2) Regarding scide preferences and sessions, and sclang extensions folder, synthdefs, help, etc:
- Macos: these are in the home folder of the user. This PR changes the build such that it doesn't look in these locations, at least for class files. 
- Linux: the appimage version just changes the $HOME variable and makes it point to a directory standing next to the executable. I didn't implement anything to get this working, none of the c++ code changes are for this, this is standard functionality in appimage applications.

I hope this clarifies a bit the differences.

I believe that the approach for Linux and Macos being different for 1) is expected, since any macos app is a bundle, while this is not usually the case in Linux, therefore extra work is needed.  For 2) possibly the same approach could be used in both linux and mac, either the one of appimage or the one of this PR. I do wonder if with this PR (macos) it is true that scide, sclang and scsynth don't touch anything in the home SuperCollider folder, including stuff such as help files, synthdefs, etc. Some type of file open call tracker (like strace in linux) could be used to confirm that. I did that on the linux side and confirmed neither of the 3 programs (scide, sclang, scsynth) touch anything in ~/.local or ~/.config.

dyfer:
@miguel-negrao thank you for clarifying. This is my understanding as well, sorry if it wasn't clear.

My points regarding 1):
- Yes, macOS (and Windows) versions already look for bundled resources in the application bundle/folder. I would like to try for the Linux version to mimic this in a way that's not necessarily specific to AppImage - just define relative paths for these resources. This only applies to the C++ part, of course bundling etc needs to be done specifically for the AppImage approach.

and regarding 2):
- there's one more resource in addition to what you already listed that needs to be handled here: 3rd party server plugins. 
- I think on all platforms we additionally need to prevent SC from looking for these resources in _system_ locations (in addition to not looking in user locations, which you already mentioned)
- while I understand that AppImage doesn't require C++ changes since it manipulates $HOME, why don't we again use the approach from macOS where we'd look for these resources in relation to the sclang/scide/servers' executables? And/or use `$APPDIR` (IIRC) variable if defined?

My thinking is that while I don't have a specific plan for non-AppImage Linux approach to standalones, I'd like to start from an approach that gets SC there as far as possible without using specific AppImage mechanisms. 

Another possible use case: I find it difficult currently to test new builds on Linux locally without installing them in the proper location. I'd imagine using the more agnostic standalone functionality to e.g. run builds installed in arbitrary locations (e.g. inside the build folder) such that they're separated from the existing SC install. Does that make sense? Or maybe my point is unnecessarily complicated or problematic?

dyfer:
One more comment specifically about this PR: I think I'm leaning towards not specifying standalone resource folder as a separate entry. I'm thinking that in the standalone build `userAppSupportDir` (and probably `systemAppSupportDir`) should simply answer the bundled location, inside the standalone folder inside the app bundle, eventually on every platform. Feel free to disagree of course :)

In any case, I think it would be still useful to define `Platform -isStandalone`, so that one can programmatically check the type of build.

muellmusik:
Hi @miguel-negrao! Thank you for picking this up! In answer to your questions (I'll answer on the Mac side):

> 1. Regarding SCClassLibrary, Docs and scsynth plugins:
> 
> * Macos: these are already inside the normal macos bundle.

Yes. Help Source is in the bundle. Rendered docs go in user app support. ~/Library/Application Support/SuperCollider/Extensions/Help. (I think system wide ones may render there as well, but that's not so important for us.



> 2. Regarding scide preferences and sessions, and sclang extensions folder, synthdefs, help, etc:
> 
> * Macos: these are in the home folder of the user. This PR changes the build such that it doesn't look in these locations, at least for class files.

Yes, it should only look in the adjacent standalone folder. I've not disabled the langconfig IIRC, as I thought it might be useful when developing and testing standalones, but possibly should.

> I do wonder if with this PR (macos) it is true that scide, sclang and scsynth don't touch anything in the home SuperCollider folder, including stuff such as help files, synthdefs, etc.

It's possible I've missed something, but IIRC this is true.

muellmusik:
I can't really comment on the linux approach, and don't know much about app image. However I don't think using the app image approach on Mac is a good idea. IIUC, that would mean putting all the custom standalone resources in the app bundle. This would defeat one of the main goals of this PR, since it would break the code-signed standalone. As well, most Mac users don't really know how to look inside app bundles (or maybe that they even exist). So even if that approach could be made to work, it would add an additional learning burden.

I agree with @dyfer that it would be nice if Linux could share as much as possible code with MacOS and Windows for this. A unified approach means that users or standalone devs don't need to learn something new when switching between platforms (perhaps casually). For example, it would make it easier for a Windows-based standalone dev to troubleshoot a Linux version of their standalone if it worked in the same way. It's also less maintenance and technical debt going forward.

So is there any reason why the adjacent folder approach wouldn't work well on Linux? By the sounds of it, it does seem like using an app image to package up the default stuff and keep it separate from the system-wide install would be very sensible.

@dyfer your suggestion about the app support directories seems like it might be a good one. I guess it may avoid some problems in 3rd party libraries? Of course we should think about whether it might cause some too! :-)

I'll be busy with gigs, deadlines and family stuff for the next couple of weeks, but can get back to this in mid-January.

miguel-negrao:
@dyfer some more remarks:

> * Yes, macOS (and Windows) versions already look for bundled resources in the application bundle/folder. I would like to try for the Linux version to mimic this in a way that's not necessarily specific to AppImage - just define relative paths for these resources. This only applies to the C++ part, of course bundling etc needs to be done specifically for the AppImage approach.

First let me point out that the macos supercollider bundle uses a lot of functionality and c++ code specific to the macos operating system. For instance `SC_Filesystem::defaultResourceDirectory` uses macos functions (`CFBundleCopyResourceURL`, etc) to get the path of the resource directory.  An AppImage is an executable with an embedded filesystem image. The binary mounts the filesystem to `/tmp`, for instance `/tmp/appimage_sd9734d2e3`, then sets the environment variable `APPDIR` to  `/tmp/appimage_sd9734d2e3` and then calls a binary in the temporary folder (`/tmp/appimage_sd9734d2e/usr/bin/scide`). So the current approach to build the path of the resource directory (which sclang and scsynth use) is to use the environment variable `APPDIR`. So you would prefer if the paths are set by constructing the path relative to the path of the executable (scsynth, sclang, and scide) ? 

> Another possible use case: I find it difficult currently to test new builds on Linux locally without installing them in the proper location. I'd imagine using the more agnostic standalone functionality to e.g. run builds installed in arbitrary locations (e.g. inside the build folder) such that they're separated from the existing SC install. Does that make sense? Or maybe my point is unnecessarily complicated or problematic?

Yes, this is already the way an AppImage works. You can put the executable wherever you want and then you just `chmod u+x` it then run it. So this functionality is already in my PR (it is the way an appimage works).

@muellmusik :

> > 3\. Regarding scide preferences and sessions, and sclang extensions folder, synthdefs, help, etc:
> > 
> > 
> > * Macos: these are in the home folder of the user. This PR changes the build such that it doesn't look in these locations, at least for class files.
> 
> Yes, it should only look in the adjacent standalone folder. I've not disabled the langconfig IIRC, as I thought it might be useful when developing and testing standalones, but possibly should.

Ok, then I think we need to specify more carefully what do we want a standalone to be. For me, my main requirement for a standalone is that it lets me "package" a finished piece and then it will run without ever being disrupted by any changes that I do in the "normal" supercollider installation. That requires complete sandboxing including not loading synthdefs from the user app support directory, or interacting with anything in that directory. For instance, in my system because of WFS a default supercollider installation, without any changes, will not even start scsynth because I have a large number of synthdefs in the synthdef folder and the maximum number of synthdefs needs to be increased. So if someone gives me a supercollider standalone probably it would not be able to start scsynth if it looks at synthdefs there. I really don't see a downside to having full sandboxing. Perhaps it would be good to compile a list of features that we want from a SuperCollider standalone.

> I can't really comment on the linux approach, and don't know much about app image. However I don't think using the app image approach on Mac is a good idea. IIUC, that would mean putting all the custom standalone resources in the app bundle.

Currently with my PR that is not what is happening, the only thing that goes into the filesystem image inside the AppImage is the stuff in `/usr`. See below the directory structure of the final AppImage. Currently in my PR the custom standalone resources need to be placed in the `SuperCollider.app.home` folder. The PR could be modified to allow copying the custom standalone resources to the filesystem image, for instance to build an app that we want to distribute to users which will not do any development or even type in any code. The custom supercollider resources could be copied via cmake to `usr/share/SuperCollider/Extensions`.  Again, if we want this capability we need to be more precise in defining what a standalone needs to do.

```
AppDir$ tree -d
.
├── apprun-hooks
└── usr
    ├── bin
    ├── include
    │   └── SuperCollider
    │       ├── common
    │       ├── lang
    │       ├── plugin_interface
    │       └── server
    ├── lib
    │   ├── gedit
    │   │   └── plugins
    │   └── SuperCollider
    │       └── plugins
    ├── libexec
    ├── plugins
    │   ├── bearer
    │   ├── iconengines
    │   ├── imageformats
    │   ├── platforminputcontexts
    │   ├── platforms
    │   ├── position
    │   ├── printsupport
    │   └── xcbglintegrations
    ├── resources
    ├── share
    │   ├── applications
    │   ├── doc
    │   │   ├── libaom0
    │   │   ├── libavahi-client3
    │   │   ├── libavahi-common3
    │   │   ├── libavcodec58
    │   │   ├── libavformat58
    │   │   ├── libavutil56
    │   │   ├── libblkid1
    │   │   ├── libbluray2
    │   │   ├── libbrotli1
    │   │   ├── libbsd0
    │   │   ├── libbz2-1.0
    │   │   ├── libcairo2
    │   │   ├── libcairo-gobject2
    │   │   ├── libchromaprint1
    │   │   ├── libcodec2-0.9
    │   │   ├── libcups2
    │   │   ├── libdatrie1
    │   │   ├── libdav1d4
    │   │   ├── libdbus-1-3
    │   │   ├── libdouble-conversion3
    │   │   ├── libdw1
    │   │   ├── libelf1
    │   │   ├── libevent-2.1-7
    │   │   ├── libffi7
    │   │   ├── libfftw3-single3
    │   │   ├── libflac8
    │   │   ├── libgcrypt20
    │   │   ├── libgdk-pixbuf-2.0-0
    │   │   ├── libglib2.0-0
    │   │   ├── libgme0
    │   │   ├── libgnutls30
    │   │   ├── libgomp1
    │   │   ├── libgraphite2-3
    │   │   ├── libgsm1
    │   │   ├── libgssapi-krb5-2
    │   │   ├── libgstreamer1.0-0
    │   │   ├── libgstreamer-plugins-base1.0-0
    │   │   ├── libhogweed6
    │   │   ├── libhyphen0
    │   │   ├── libicu67
    │   │   ├── libidn2-0
    │   │   ├── libjpeg62-turbo
    │   │   ├── libk5crypto3
    │   │   ├── libkeyutils1
    │   │   ├── libkrb5-3
    │   │   ├── libkrb5support0
    │   │   ├── liblcms2-2
    │   │   ├── liblz4-1
    │   │   ├── liblzma5
    │   │   ├── libmd0
    │   │   ├── libmd4c0
    │   │   ├── libmfx1
    │   │   ├── libminizip1
    │   │   ├── libmount1
    │   │   ├── libmp3lame0
    │   │   ├── libmpg123-0
    │   │   ├── libnettle8
    │   │   ├── libnorm1
    │   │   ├── libnspr4
    │   │   ├── libnss3
    │   │   ├── libnuma1
    │   │   ├── libogg0
    │   │   ├── libopenjp2-7
    │   │   ├── libopenmpt0
    │   │   ├── libopus0
    │   │   ├── liborc-0.4-0
    │   │   ├── libpango-1.0-0
    │   │   ├── libpangocairo-1.0-0
    │   │   ├── libpangoft2-1.0-0
    │   │   ├── libpcre2-16-0
    │   │   ├── libpcre2-8-0
    │   │   ├── libpcre3
    │   │   ├── libpgm-5.3-0
    │   │   ├── libpixman-1-0
    │   │   ├── libpng16-16
    │   │   ├── libqt5core5a
    │   │   ├── libqt5dbus5
    │   │   ├── libqt5gui5
    │   │   ├── libqt5network5
    │   │   ├── libqt5positioning5
    │   │   ├── libqt5positioning5-plugins
    │   │   ├── libqt5printsupport5
    │   │   ├── libqt5qml5
    │   │   ├── libqt5qmlmodels5
    │   │   ├── libqt5quick5
    │   │   ├── libqt5quickwidgets5
    │   │   ├── libqt5sensors5
    │   │   ├── libqt5svg5
    │   │   ├── libqt5webchannel5
    │   │   ├── libqt5webenginecore5
    │   │   ├── libqt5webenginewidgets5
    │   │   ├── libqt5webkit5
    │   │   ├── libqt5websockets5
    │   │   ├── libqt5widgets5
    │   │   ├── librabbitmq4
    │   │   ├── libre2-9
    │   │   ├── libreadline8
    │   │   ├── librsvg2-2
    │   │   ├── libselinux1
    │   │   ├── libshine3
    │   │   ├── libsnappy1v5
    │   │   ├── libsndfile1
    │   │   ├── libsodium23
    │   │   ├── libsoxr0
    │   │   ├── libspeex1
    │   │   ├── libsqlite3-0
    │   │   ├── libsrt1.4-gnutls
    │   │   ├── libssh-gcrypt-4
    │   │   ├── libssl1.1
    │   │   ├── libswresample3
    │   │   ├── libsystemd0
    │   │   ├── libtasn1-6
    │   │   ├── libtheora0
    │   │   ├── libtinfo6
    │   │   ├── libtwolame0
    │   │   ├── libudev1
    │   │   ├── libudfread0
    │   │   ├── libunistring2
    │   │   ├── libunwind8
    │   │   ├── libva2
    │   │   ├── libva-drm2
    │   │   ├── libva-x11-2
    │   │   ├── libvdpau1
    │   │   ├── libvorbis0a
    │   │   ├── libvorbisenc2
    │   │   ├── libvorbisfile3
    │   │   ├── libvpx6
    │   │   ├── libwavpack1
    │   │   ├── libwebp6
    │   │   ├── libwebpdemux2
    │   │   ├── libwebpmux3
    │   │   ├── libwoff1
    │   │   ├── libx11-xcb1
    │   │   ├── libx264-160
    │   │   ├── libx265-192
    │   │   ├── libxau6
    │   │   ├── libxcb-glx0
    │   │   ├── libxcb-icccm4
    │   │   ├── libxcb-image0
    │   │   ├── libxcb-keysyms1
    │   │   ├── libxcb-randr0
    │   │   ├── libxcb-render0
    │   │   ├── libxcb-render-util0
    │   │   ├── libxcb-shape0
    │   │   ├── libxcb-shm0
    │   │   ├── libxcb-sync1
    │   │   ├── libxcb-util1
    │   │   ├── libxcb-xfixes0
    │   │   ├── libxcb-xinerama0
    │   │   ├── libxcb-xinput0
    │   │   ├── libxcb-xkb1
    │   │   ├── libxcomposite1
    │   │   ├── libxcursor1
    │   │   ├── libxdamage1
    │   │   ├── libxdmcp6
    │   │   ├── libxext6
    │   │   ├── libxfixes3
    │   │   ├── libxi6
    │   │   ├── libxkbcommon0
    │   │   ├── libxkbcommon-x11-0
    │   │   ├── libxml2
    │   │   ├── libxrandr2
    │   │   ├── libxrender1
    │   │   ├── libxslt1.1
    │   │   ├── libxss1
    │   │   ├── libxvidcore4
    │   │   ├── libzmq5
    │   │   ├── libzstd1
    │   │   ├── libzvbi0
    │   │   └── ocl-icd-libopencl1
    │   ├── emacs
    │   │   └── site-lisp
    │   │       └── SuperCollider
    │   ├── gtksourceview-3.0
    │   │   └── language-specs
    │   ├── icons
    │   │   └── hicolor
    │   │       ├── 128x128
    │   │       │   └── apps
    │   │       ├── 16x16
    │   │       │   └── apps
    │   │       ├── 256x256
    │   │       │   └── apps
    │   │       ├── 32x32
    │   │       │   └── apps
    │   │       ├── 48x48
    │   │       │   └── apps
    │   │       ├── 64x64
    │   │       │   └── apps
    │   │       └── scalable
    │   │           └── apps
    │   ├── mime
    │   │   └── packages
    │   └── SuperCollider
    │       ├── examples
    │       │   ├── Bela
    │       │   ├── between_languages
    │       │   ├── demonstrations
    │       │   ├── GUI examples
    │       │   ├── other
    │       │   ├── pieces
    │       │   └── research_and_tools
    │       ├── Extensions
    │       │   ├── scide_scel
    │       │   │   └── HelpSource
    │       │   │       └── Classes
    │       │   └── scide_scvim
    │       ├── HelpSource
    │       │   ├── Classes
    │       │   ├── Guides
    │       │   ├── images
    │       │   ├── lib
    │       │   ├── Other
    │       │   ├── Overviews
    │       │   ├── QC
    │       │   ├── Reference
    │       │   └── Tutorials
    │       │       ├── A-Practical-Guide
    │       │       ├── Getting-Started
    │       │       ├── JITLib
    │       │       └── Mark_Polishook_tutorial
    │       │           └── Japanese_version
    │       ├── HID_Support
    │       │   └── hut
    │       ├── SCClassLibrary
    │       │   ├── backwards_compatibility
    │       │   ├── Common
    │       │   │   ├── Audio
    │       │   │   │   ├── Bela
    │       │   │   │   └── iphone
    │       │   │   ├── Collections
    │       │   │   │   ├── linux
    │       │   │   │   ├── osx
    │       │   │   │   └── windows
    │       │   │   ├── Control
    │       │   │   │   └── asScore
    │       │   │   ├── Core
    │       │   │   ├── Files
    │       │   │   ├── Geometry
    │       │   │   ├── GUI
    │       │   │   │   ├── backwardsCompatibility
    │       │   │   │   ├── Base
    │       │   │   │   ├── iphone
    │       │   │   │   │   └── Base
    │       │   │   │   ├── PlusGUI
    │       │   │   │   │   ├── Collections
    │       │   │   │   │   ├── Control
    │       │   │   │   │   ├── Core
    │       │   │   │   │   ├── Files
    │       │   │   │   │   └── Math
    │       │   │   │   └── tools
    │       │   │   │       └── guicrucial
    │       │   │   ├── Math
    │       │   │   ├── Quarks
    │       │   │   │   └── GUI
    │       │   │   ├── Streams
    │       │   │   ├── UnitTesting
    │       │   │   │   └── GUI
    │       │   │   └── Unix
    │       │   ├── DefaultLibrary
    │       │   ├── deprecated
    │       │   │   ├── 3.10
    │       │   │   │   ├── GUI
    │       │   │   │   └── osc
    │       │   │   ├── 3.11
    │       │   │   ├── 3.12
    │       │   │   └── 3.13
    │       │   │       └── GUI
    │       │   ├── External
    │       │   │   └── Ableton
    │       │   ├── JITLib
    │       │   │   ├── GUI
    │       │   │   ├── Patterns
    │       │   │   └── ProxySpace
    │       │   ├── Platform
    │       │   │   ├── iphone
    │       │   │   │   └── SystemOverwrites
    │       │   │   ├── linux
    │       │   │   │   ├── GUI
    │       │   │   │   └── SystemOverwrites
    │       │   │   ├── osx
    │       │   │   └── windows
    │       │   │       └── SystemOverwrites
    │       │   ├── SCDoc
    │       │   └── scide_scqt
    │       ├── sounds
    │       └── translations
    └── translations
        └── qtwebengine_locales
```

> So is there any reason why the adjacent folder approach wouldn't work well on Linux? By the sounds of it, it does seem like using an app image to package up the default stuff and keep it separate from the system-wide install would be very sensible.

The location of the original appimage file is placed in the env var `APPIMAGE` therefore it would be easy to get to a folder next to the appimage file. So in principle it could be done. But to reiterate, I'm skeptical of an approach that is not a complete sandbox solution.

Ok, so given the feedback so far my thinking is as follows:
- There are two completely different functionalities being discussed here: 1) application bundles (which already exist on macos, but not on linux) and 2) standalones or sandboxing.
- My PR actually only implements directly the functionality 1). It also implements 2) indirectly because AppImages already have a builtin sandboxing functionality via a `.home` directory. I think that the two functionalities should be kept separete, meaning, one should be able to have a supercollider AppImage which is not a standlone, for instance, so that users of linux distros other than debian based can install supercollider easilly, and also to allow quickly testing a build from the CI without having to install it on our machine. Therefore I propose that my AppImage PR should be seen as proposal for an easy to install version of supercollider for any Linux distribution. So I think there should be two cmake variables: LINUX_APPIMAGE (to create a Linux "bundle) and STANDALONE (to implement sandboxing). On Linux activating STANDALONE would also activate  LINUX_APPIMAGE, but it would be possible to build with just LINUX_APPIMAGE activated.
- I think a SuperCollider standalone should be a complete sandboxing solution, but this needs further discussion. 
- The consensus seems to be that whatever solution is implemented for sandboxing it should be the same for macos and linux. Further discussion is needed on how to achieve this, which also depends on what level of sandboxing we want for a standalone.

So I propose continuing the discussion on how a standalone should work here, and on my AppImage PR implementing for now only the "Linux bundle" functionality. Let me know how you feel about this.

PR #5900, opened 2022-11-04
Topic refactor function flop

telephon:
## Purpose and Motivation

- The old implementation of function flop was hard to understand and not useful for other puroses.
- inEnvir had no support for keyword arguments

This fixes both.
Up for discussion: the benchmarks are all very low. Maybe just replace the old inEnvir with the new one.

```
f = { |a, b| [a, b, ~c, ~d] };
x = f.inEnvir((c: 1, d: 2));
y = f.inEnvirWithArgs((c: 1, d: 2));

bench { 1000.do { x.value(1, 2) } }
bench { 1000.do { y.value(1, 2) } }

bench { 100.do { f.inEnvir((c: 1, d: 2)) } }
bench { 100.do { f.inEnvirWithArgs((c: 1, d: 2)) } }
```

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Core/Kernel.sc
+++ b/SCClassLibrary/Common/Core/Kernel.sc
@@ -474,7 +474,7 @@ FunctionDef {
 	archiveAsCompileString { ^true }
 
 	hasPartialApplication {
-		^this.numArgs > 0 and: { argNames[0] == \_ }
+		^argNames.size > 0 and: { argNames[0] == \_ }
 	}
 
 	argumentString { arg withDefaultValues=true, withEllipsis=false;
--- a/SCClassLibrary/Common/Core/Kernel.sc
+++ b/SCClassLibrary/Common/Core/Kernel.sc
@@ -477,9 +477,11 @@ FunctionDef {
 		^argNames.size > 0 and: { argNames[0] == \_ }
 	}
 
-	argumentString { arg withDefaultValues=true, withEllipsis=false;
+	argumentString { arg withDefaultValues=true, withEllipsis=false, asArray=false;
 		var res = "", pairs;
 		var lastIndex, noVarArgs, varArgName;
+		if(asArray && withDefaultValues) { Error("argumentString as array should not have default values").throw };
+		if(asArray && withEllipsis) { Error("argumentString as array should not have ellipsis").throw };
 		if(this.hasPartialApplication) {
 			^if(withEllipsis) { " ... args" } { "args" }
 		};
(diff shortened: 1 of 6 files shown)


dyfer:
Wasn't `makeFlopFuncString` introduced in #5499? If so, it could be removed altogether, since we didn't have it in the release. Which would be a good reason to expedite it and include it in 3.13...

telephon:
@dyfer ok, so this is much better.

I would like to add one new method, which is just like flop but unbubbles the result. So when you call the function without args, it will return a value rather than that value in an array. Could be called `flopu` (flop unbubbled) or `flap` (old variant of flop).

e.g.

```
f = { |a, b| a + b }.flopu;
f.([1, 2], 3); // => [4, 5]
f.(1, 2); // 3
```

Any thoughts?

telephon:
Good that you ask, let's discuss it.

For the use case. Let's say you want to implement multichannel expansion for a given functionality, for simplicity, let's say for `if`.:

```supercollider
~if = { |condition, trueFunc, falseFunc| if(condition, trueFunc, falseFunc) }.flop;
~if.([1, 2, 3, 4, 5] > 3, { \yes }, { \no }); // ->  [ no, no, no, yes, yes ]
~if.(5 > 3, { \yes }, { \no }; // -> [ yes ]
```

The function will always return an array, even if your arguments are not arrays.
If you want the function to be like a multichannel expanding thing, e.g. a UGen, then you need flop1

```supercollider
~if = { |condition, trueFunc, falseFunc| if(condition, trueFunc, falseFunc) }.flop1;
~if.([1, 2, 3, 4, 5] > 3, { \yes }, { \no }); // ->  [ no, no, no, yes, yes ]
~if.(5 > 3, { \yes }, { \no }); // -> yes
```

In a sense, one could call `flop1` also `multichannelExpand`.

telephon:
Btw.  noticed a problem I need to fix, with `flop` and ellipsis args: `f = _ + _; f.flop;`

telephon:
OK, this is solved.

dyfer:
Thanks for the explanation.
> If you want the function to be like a multichannel expanding thing, e.g. a UGen, then you need flop1

Still, in a synth using a regular value versus a single-member array is almost interchangeable... that was the reason I asked for practical need for this.

I understand the argument of consistency when passing a non-array argument, so I guess it's fine. I just don't have an idea where I'd use it.

I'm re-running the tests to make sure they pass.

dyfer:
`TestMethod -test_method_argumentString` fails in this PR (I double checked in the downloaded build). The test is in `TestMethod` (you can check this using `TestMethod.run` after including the testsuite in the language config). 

It seems that the change to `FunctionDev -argumentString` triggered this. Do you have any idea what is going on there?

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -473,9 +473,17 @@ FunctionDef {
 	checkCanArchive { "cannot archive FunctionDefs".warn }
 	archiveAsCompileString { ^true }
 
+	hasPartialApplication {
+		^this.numArgs > 0 and: { argNames[0] == \_ }
+	}
```
dyfer, line 478: I think this is where the problem is - argNames might be `nil`. Possible solution:

```supercollider
	hasPartialApplication {
		^this.numArgs > 0 and: { argNames.notNil } and: { argNames[0] == \_ }
	}
```

dyfer: also, shouldn't the whole expression be in parentheses? (e.g. `^(this.numArgs > 0 and: {...} )`)

dyfer:
Thanks for fixing this, @telephon !

For some mysterious reason, the CI did not run after your last commit. Could you rebase the branch and force push it so that CI gets triggered? Sorry for the hassle, I don't seem to have the option to run it manually... (I guess it's something to add)

telephon:
hm, on what would you rebase to make it more than a noop?

telephon:
I've updated the helpfile, that helped.

dyfer (approved):
Looks good to me, thanks!

Anything else we should consider before merging?

telephon:
I'm quite happy with it. I have an idea for making the code clearer, but I hesitate, because I don't want to strain time and patience.

dyfer:
@telephon it's your call. I feel like we have a test suite that should catch most issues with this, so if you feel like giving it another go, I'd be fine with it. But if you think this is good as is, I'm happy to merge it.

telephon:
thanks! I've decided to move `makeFuncModifierString` and all the string operation stuff into `FunctionDef` because these are really definition related. This simplifies the code in `Function`. Now all is documented.

I won't touch it anymore unless there are suggestions from you (ready to merge).

dyfer:
Thanks @telephon !

This looks good to me, and since there are no failures in the testsuite related to this functionality, I'll merge this now.

Issue #5893, opened 2022-10-25
audio rate .sqrt of 0 (zero) produces NaN


woutersnoei:
## Environment

* SuperCollider version:
3.13dev (arm build)

* Operating system:
macOS Monterey 12.6

## Steps to reproduce

```supercollider
{ DC.ar(0).sqrt }.plot; // produces NaN
{ DC.kr(0).sqrt }.plot; // produces 0 (as it should)
```

## Expected vs. actual behavior

In two recent dev builds (for arm64) calling 'sqrt' on an audio signal that has value 0 produces a NaN. This isn't the case in the official build (3.12.2) so the issue is either caused by a change after that or by the arm64 architecture.

cheers & thanks,
Wouter

woutersnoei:
Addition: this only happens in the arm64 build. I tried in both x64 and arm64 builds from this action: https://github.com/supercollider/supercollider/actions/runs/3317669484
The x64 build (in Rosetta mode on my machine) returns 0 as it should, only the arm64 build returns NaN

dyfer:
I can confirm this issue. This is not good :/
I can also see a difference eg in `2.sqrt` on the server:
```supercollider
{ DC.ar(2).sqrt.poll }.play; 
// UGen(UnaryOpUGen): 1.41421 - Intel x86_64
// UGen(UnaryOpUGen): 1.41016 - Apple arm64
```
I'm guessing this might have to do with compiler settings but I'm not sure where it's coming from... we don't seem to set anything differently depending on the platform.

dyfer:
It seems that the problem occurs in the `nova-simd` class here: https://github.com/timblechmann/nova-simd/blob/master/detail/vec_math.hpp#L550-L558 I can see NaNs here when I set a breakpoint there. Also, I'm not sure if `VecType::gen_sign_mask();` is returning correct values.

I can see online that there are some differences in floating point representations between M1 and Intel, but I'm not sure if that's the cause of the issue here...

dyfer:
@timblechmann do you have any ideas how to solve this?

timblechmann:
@dyfer, you could double-check the nova-simd testsuite.

the sign mask generation in https://github.com/timblechmann/nova-simd/blob/master/vec/vec_neon.hpp#L64-L69 should probably go through a union, though, as it may cause strict aliasing issues

dyfer:
[...]

PR #5901, opened 2022-11-06
Fix .sqrt and .reciprocal on the server on M1 CPUs

dyfer:
## Purpose and Motivation

This is a workaround for #5893 and fixes the behavior of `.sqrt` and `.reciprocal` UGens on Apple M1.

The PR disables nova-simd (vectorized) implementations of `recip` and `sqrt` functions on ARM CPUs due to an issue upstream: https://github.com/timblechmann/nova-simd/issues/11

Once these functions are fixed upstream, the changes introduced in this PR should be reverted for best performance.

I'm adding the `macOS` label since performance on Apple M1 CPU is the primary target of this "fix". However, this might apply to all ARM64 CPUs, so I'm guessing it may also affect RaspberryPi 4 (not tested).

To test:
```supercollider
s.boot;
x = {DC.ar(2).reciprocal.poll; Silent.ar()}.play; // UGen(UnaryOpUGen): 0.5 (was: 0.499998)
x = {DC.ar(0).sqrt.poll; Silent.ar()}.play; // UGen(UnaryOpUGen): 0 (was: nan)
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/plugins/UnaryOpUGens.cpp
+++ b/server/plugins/UnaryOpUGens.cpp
@@ -1015,13 +1015,21 @@ static UnaryOpFunc ChooseNovaSimdFunc(UnaryOpUGen* unit) {
         case opCubed:
             return &cubed_nova_64;
         case opSqrt:
+#    ifdef __ARM_NEON // remove this once nova-simd implementation is fixed
+            func = &sqrt_a;
+#    else
             func = &sqrt_nova_64;
+#    endif
             break;
         case opExp:
             func = &exp_nova;
             break;
         case opRecip:
+#    ifdef __ARM_NEON // remove this once nova-simd implementation is fixed
+            return &recip_a;
+#    else
             return &recip_nova_64;
+#    endif
         case opMIDICPS:
             func = &midicps_nova;
             break;
@@ -1171,13 +1179,21 @@ static UnaryOpFunc ChooseNovaSimdFunc(UnaryOpUGen* unit) {
         func = &cubed_nova;
         break;
     case opSqrt:
+#    ifdef __ARM_NEON // remove this once nova-simd implementation is fixed
+        func = &sqrt_a;
+#    else
         func = &sqrt_nova;
+#    endif
         break;
     case opExp:
         func = &exp_nova;
         break;
     case opRecip:
+#    ifdef __ARM_NEON // remove this once nova-simd implementation is fixed
+        func = &recip_a;
+#    else
         func = &recip_nova;
+#    endif
         break;
     case opMIDICPS:
         func = &midicps_nova;


joshpar (approved):

timblechmann:
> x = {DC.ar(2).reciprocal.poll; Silent.ar()}.play; // UGen(UnaryOpUGen): 0.5 (was: 0.499998)

i guess this is more a feature than a bug: the division is done by an approximation plus a newton-raphson iteration, which should give a precision of 22bit mantissa. one can do a second iteration to get full 24bit precision.

on current simd hardware (sse/avx/neon) one can basically have 3 levels of accuracy for reciprocal: 12bit, 22bit and 24bit precision (raw, raw+1NR, raw+2NR)

dyfer:
> i guess this is more a feature than a bug:

On x86_64 (with SSE I presume) `DC.ar(2).reciprocal` reports `0.5`. I understand that this might not be reported accurately and I understand the basic principle of what you're saying about the algorithm, but here I am interested in consistent behavior between platforms. If `DC.ar(2).reciprocal` reports `0.5` on x86_64, it should report the same on aarch64 (and it does after this PR). 

BTW the test suite in nova-simd also reports issues with the reciprocal function (that's how I knew to double check it), while al tests pass on x86_64.

timblechmann:
> BTW the test suite in nova-simd also reports issues with the reciprocal function on aarch64

possibly because i wrote the test suite for x86_64, so take it with a grain of salt.

> If DC.ar(2).reciprocal reports 0.5 on x86_64, it should report the same on aarch64

floating point math is a beast, as it's typically not ieee754 complaint. it's an **incredibly** hard problem to ensure accurate floating point operations and compiler flags/compiler updates  (`-ffast-math` and friends) are going to break it. i've seen it many times that different platforms result in different floating point code or that numerical unit tests break when porting to a different platform (or simply an updated compiler)

dyfer:
> possibly because i wrote the test suite for x86_64, so take it with a grain of salt.

Noted, thanks.

> floating point math is a beast, as it's typically not ieee754 complaint. it's an **incredibly** hard problem to ensure accurate floating point operations

Thanks for emphasizing this. 

Still, do you have suggestions how to approach the reciprocal function on neon? Should that be updated to be done with improved accuracy, or is it a tradeoff of faster/vectorized vs more accurate non-vectorized calculations?

I assume that signed sqrt returning nan is still a valid issue to be addressed in nova-simd, correct?

timblechmann:
> Still, do you have suggestions how to approach the reciprocal function on neon? Should that be updated to be done with improved accuracy, or is it a tradeoff of faster/vectorized vs more accurate non-vectorized calculations?

it's a trade-off: it's using 1 NR step, while 2 NR steps could give a better result

> Also, I assume that signed sqrt returning nan is still a valid issue to be addressed in nova-simd, correct?

yes. that's a bug, which should be fixed. i'm not having an m1 at hand, but i'm happy to guide you through, if you want to debug it. the idea is to store the sign bit, perform the operation on the absolute and restore the old sign bit. it may help to print some of the intermediate steps

dyfer:
> it's a trade-off: it's using 1 NR step, while 2 NR steps could give a better result

Gotcha. But is this a difference in nova-simd's implementation between SSE and NEON? As I wrote before, SSE implementation seems more accurate (unless I'm missing something).

> i'm not having an m1 at hand, but i'm happy to guide you through, if you want to debug it. the idea is to store the sign bit, perform the operation on the absolute and restore the old sign bit.

Thank you this is helpful information, and I'd like to take you up on the offer to help (with no particular timeline in mind). I'll comment in https://github.com/timblechmann/nova-simd/issues/11 once I resume working on this.

Issue #5899, opened 2022-11-03
Supernova seems only compatible with VS 2017 on Windows


balintlaczko:
## Environment

* SuperCollider version: both 3.12.2 and local build of 3.13.0-dev
* Operating system: Windows 10
* Other details (Qt version, audio driver, etc.): Qt 5.15.2, ASIO

## Steps to reproduce

```supercollider
Server.supernova;
s.boot;
```

## Expected vs. actual behavior

I get the infamous exit code -1073741676, which is [apparently a divide-by-zero error](https://stackoverflow.com/questions/47422982/c-status-1073741676).

I have seen some other people facing this issue too, like [here](https://www.reddit.com/r/supercollider/comments/wnylee/im_trying_to_boot_supernova_now_with_312/) and [here](https://scsynth.org/t/win11-supernova/6377).

I compiled SC from source (with Supernova), and it seems that strangely enough if I am building in Debug mode, Supernova manages to boot with some quirks:
- a) I get the [thread affinity warnings](https://github.com/supercollider/supercollider/issues/5540) that indeed go away if I build from [this PR](https://github.com/supercollider/supercollider/pull/5618).
- b) when I try to use the default Synth, I get an error: `OffsetOut_Ctor: alloc failed, increase server's RT memory (e.g. via ServerOptions)` which does not seem affected by the size of `Server.default.options.memSize`. However if I define a SynthDef, that seems to work okay.

SC 3.13.0-dev in Debug mode (this works):
```supercollider
Server.supernova;
s.boot;

(
SynthDef("sine", { arg freq=800, chan=0;
    var osc;
    osc = SinOsc.ar(freq, 0, mul:1/600);
    Out.ar(chan, osc);
}).writeDefFile;
)

(
var group = ParGroup.new();
600.do{ Synth(\sine, target: group) };
)
```

BUT if I try this (still in SC 3.13.0-dev in Debug):
```supercollider
Server.supernova;
s.boot;

(
var group = ParGroup.new();
n = 10
n.do{ Synth(\default, target: group) };
)
```
I get a `OffsetOut_Ctor: alloc failed, increase server's RT memory (e.g. via ServerOptions)` n times.

If I build in Release mode, Supernova refuses to boot with the familiar exit code -1073741676, again.
SC 3.12.2 (public release) OR 3.13.0-dev built in Release mode:
```supercollider
Server.supernova;
s.boot;
```

I get:
```
Booting server 'localhost' on address 127.0.0.1:57110.
[...]

PR #5902, opened 2022-11-09
fix OffsetOut_Ctor

Spacechild1:
## Purpose and Motivation

Fixes the `OffsetOut_Ctor` is with supernova mentioned in #5899

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/plugins/IOUGens.cpp
+++ b/server/plugins/IOUGens.cpp
@@ -1327,15 +1327,20 @@ void OffsetOut_Ctor(OffsetOut* unit) {
     unit->m_busTouched = world->mAudioBusTouched;
     int32 offset = unit->mParent->mSampleOffset;
     int numChannels = unit->mNumInputs - 1;
-
-    unit->m_saved = (float*)RTAlloc(unit->mWorld, offset * numChannels * sizeof(float));
-    ClearUnitIfMemFailed(unit->m_saved);
+    // NB: if mSampleOffset is 0, RTAlloc() might return a nullptr and
+    // trigger ClearUnitIfMemFailed(), so we have to handle it specially.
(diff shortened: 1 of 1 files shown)


server/plugins/IOUGens.cpp
```diff
@@ -1327,15 +1327,20 @@ void OffsetOut_Ctor(OffsetOut* unit) {
...
+        ClearUnitIfMemFailed(unit->m_saved);
+    } else {
+        unit->m_saved = nullptr;
+    }
     unit->m_empty = true;
     // Print("<-Out_Ctor\n");
 }
 
 void OffsetOut_Dtor(OffsetOut* unit) {
-    // Ctor may not have run, if unit was immediately created->paused->freed!
```
dyfer, line 1338: Was this comment inaccurate? With so few people understand what's going on in SN, I wonder if we should remove breadcrumbs like this.

Spacechild1: Yes, the comment is wrong/misleading because a unit destructor only gets called if the constructor has been called.

Scsynth:
https://github.com/supercollider/supercollider/blob/02c37d28badc390f7ee320883b955052294d3cd6/server/scsynth/SC_Graph.cpp#L56-L65

Supernova:
https://github.com/supercollider/supercollider/blob/02c37d28badc390f7ee320883b955052294d3cd6/server/supernova/sc/sc_synth.cpp#L171-L176

dyfer (approved):
Thanks!

It seems to fix one issue on Windows/MSVC and doesn't seem to break the functionality on other platforms.

PR #5908, opened 2022-11-15
class library: only one server meter by default

### LINKED ISSUE #5545: Many meters we have.

telephon:
## Purpose and Motivation

This fixes #5545. ServerMeter now behaves like Scope. By result of the discussions in #5549, it may make sense to rethink the design of all server windows, and potentially an easy way to handle unique windows.

A refactoring of the server views is recommended, this is just the minimal step to fix the issue at hand.

Multiple ServerMeters can be created by directly instantiating the object (like Stethoscopse for multiple scopes).

Thanks to @capital-G and everyone else for discussion.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested

### DIFF
--- a/SCClassLibrary/Common/Control/Server.sc
+++ b/SCClassLibrary/Common/Control/Server.sc
@@ -357,7 +357,7 @@ Server {
 	var <defaultGroup, <defaultGroups;
 
 	var <syncThread, <syncTasks;
-	var <window, <>scopeWindow, <emacsbuf;
+	var <window, <>scopeWindow, <serverMeter, <emacsbuf;
 	var <volume, <recorder, <statusWatcher;
 	var <pid, serverInterface;
 	var pidReleaseCondition;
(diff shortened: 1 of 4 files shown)


telephon:
@dyfer it might be a candidate for the 3.13.0-rc2, just an annoying thing that is finally fixed.

It does change behaviour, this is why I have stated it is a `Breaking change`. But it fixes also something that wasn't good.

Suggestions welcome.

muellmusik:
Strongly support this!

dyfer:
Thanks @telephon 
Could you rebase this on the `3.13` branch? I'd like to get it in the next RC.

dyfer:
It seems you did not rebase onto 3.13... (or something else was wrong, there shouldn't be a commit with the 3.14 version change after the rebase). However, with the reverting commit I think I can just squash merge to clean it up. Should I do that or would you try rebasing again?

telephon:
Yes, it's been a mess, sorry – for convenience I tried to do it with GitHub Desktop, which seems to be broken in this respect.
With 
```
git rebase --onto 3.13 topic-fix-multiple-meters

```
it says `Everything up-to-date`. So I suppose the best is a squash.

dyfer (approved):
Thanks!

Issue #3567, opened 2018-03-10
falling back from inlining


telephon:
Here is a general design issue about the trade off between pure object oriented semantics and performance optimizations.

Strictly speaking, it should be **always** the receiver of a message that decides what to do with it. Because we generate optimized bytecode for commonly used operations, this rule is broken in some cases:

- `if`
- `switch`
- `while`

(any more?)

These cannot be extended by objects, which is problematic when you try to extend the language, because they are really core operations.

So the issue is: is there any middle ground between OOP and byte code optimized methods?

telephon:
Discussion inspired by #3564 and moved from there, because it is a separate issue.

jamshark70:
Worst case is that the compiler could write both inlined and non-inlined bytecodes into the functiondef, reserving the non-inlined version as a fallback. It would use more memory but may be worth it for correctness.

telephon:
I am actually not sure whether it is inlining only. 

An example is when you write an object that returns true or false with a certain probability:

```supercollider
Probably {
   *if { |trueFunc, falseFunc|
       if(0.8.coin, trueFunc, falseFunc)
    }
}
```

```supercollider
if(Probably) { 1 } { 0 }
```
returns `ERROR: Non Boolean in test.`

brianlheim:
Are there any strong examples of problems caused by this issue? I'm not sure I understand why anyone would want to overload what are de facto keywords.

telephon:
All  exceptions to rules are an issue of consistency and uniformity of a system, even more so if they concern central elements like control flow.

- orthogonality: you can't expect it to behave the same in all cases.  
- purity in OOP principles: the meaning of a message in an OOP system is determined by the receiver. Everything else is a leak, strictly speaking.
- extendibility: in this case it has an impact on the ability to cleanly extend the language. 


These operators are de jure messages as every other thing in sclang. The fact that they are de facto keywords is just a consequence of compromises in implementation. An example is the implementation of `UGen:if` (which is something like "the one exception allowed", which perhaps never should have happened). I would like to ask if the compromises are really necessary?

Examples:

- I've encountered the issue when implementing [Neutral](https://github.com/telephon/Neutral), which is intended as a prototype for our future `AbstractObject`. If you want a delegator object to pass on all messages, no message should be excluded.

- You also may want to implement values of an alternative logic, or a debugger object that stops on every call. The fact that nobody has done it might just be the consequence of the current restriction.

brianlheim:
[...]

PR #5919, opened 2022-11-25
class library: complete bool interface for AbstractFunction

telephon:
## Purpose and Motivation

At first, it seems not logical that we have `nand` and `&&`, and `and`  for Boolean, but `and` and `or` are missing from `AbstractFunction`. This implements it.

But we get a (minor?) problem, namely that `and` and `or` try to inline the function, so that we get an inline warning for a term like this:
```
a = and({ |x| x.not }, { |x| x }); // WARNING: FunctionDef contains variable declarations and so will not be inlined.
a.(true)
``` 

So let's not yet merge this, but decide first if we need to document better why there is no `and` and `or` for Functions, or make sure that inline warnings are thrown for the right reason (the fact that the function has an an argument seems to to be the right reason).

Note: implementing it like this:
```supercollider
	|| { arg function, adverb; ^this.composeBinaryOp('||', function, adverb) }
	&& { arg function, adverb; ^this.composeBinaryOp('&&', function, adverb) }
	or { arg function, adverb; ^this.composeBinaryOp('||', function, adverb) }
	and { arg function, adverb; ^this.composeBinaryOp('&&', function, adverb) }
```

won't help, it's an interpreter thing.

## Types of changes

- New feature

### DIFF
--- a/SCClassLibrary/Common/Core/AbstractFunction.sc
+++ b/SCClassLibrary/Common/Core/AbstractFunction.sc
@@ -160,6 +160,8 @@ AbstractFunction {
 
 	|| { arg function, adverb; ^this.composeBinaryOp('||', function, adverb) }
 	&& { arg function, adverb; ^this.composeBinaryOp('&&', function, adverb) }
+	or { arg function, adverb; ^this.composeBinaryOp('or', function, adverb) }
+	and { arg function, adverb; ^this.composeBinaryOp('and', function, adverb) }
 	xor { arg function, adverb; ^this.composeBinaryOp('xor', function, adverb) }
 	nand { arg function, adverb; ^this.composeBinaryOp('nand', function, adverb) }
 	not { ^this.composeUnaryOp('not') }


JordanHendersonMusic:
@telephon is this possible given what was discussed over here #3567?

JordanHendersonMusic:
@telephon I don't think this is possible given what was dicussed in #3567 

```supercollider
f = { |x| x.not };

r = f.and({ |x| x }) // fine

r = f.and({1}) // ERROR: Non Boolean in test.
```

telephon:
I need to read up on this again. But it seems hard to explain why only two operators are unsupported, while the rest is.

JordanHendersonMusic:
Just revisiting this as it is an old PR. 

I think the fact the interpreter can't inline functions with arguments is what makes this work.

```
a = { |x| x.not } or: { |x| x };  // this works because they aren't inlineable.

a = { false } or: { true } // ERROR: Non Boolean in test.
```

When there is an inlinable function, the interpreter turns the bytecode into a jump–if. This requires the lhs and rhs to be a boolean (not a function). 

One solution would be to then dispatch the `or` if the arguments aren't booleans (I forget who suggested this idea), however, the `or` has been lost as all that is present in the bytecode is the jump–if, and this is used for many different operations. Fixing this would require adding the selector to the bytecode as an operand. This is potentially do–able, but the opcodes need refactoring first because they are fragile. 

Also, consider `!?` and `??`, they are confusing here!
```
f = { nil }
g = { 2 }

(f !? g)
```
If we are composing operators, then this should return a function that returns nil. What you actually get is the literal `2` being returned.

IMO, as it stands, this PR cannot proceed until the jump bytecodes have their parent selector added as an operand, and even then, the wider goal of using all methods on function to allow composition will always have some edge cases. One case that could be insurmountable would be `while` as it is implemented in terms of itself.

PR #5923, opened 2022-12-01
class library: call dispatch on removeAt

### LINKED ISSUE #5922: Should EnvironmentRedirect call dispatch for removeAt?
EnvironmentRedirect help claims 'dispatch' is "A function or object that is called when the environment is modified."

However the behavior is inconsistent:

```
e = EnvironmentRedirect.new;

e.dispatch = { |... args| args.debug("e dispatch") };

e[\a] = 1;  // e dispatch: [ a, 1 ]

e.removeAt(\a)  // no post
```

At `removeAt`, the environment has been modified but the function was not called.

This could be resolved as follows:

```
	removeAt { arg key;
		var result = envir.removeAt(key);
		dispatch.value(key, nil);
		^result
	}
```

(One question is whether we would then need a `localRemoveAt` method mirroring `localPut`.)

telephon:
## Purpose and Motivation

This fixes #5922

- call dispatch on removeAt (because the object changed)
- return object on removeAt (this is standard in collections)
- add localRemoveAt for removing without dispatch

## Types of changes

- Bug fix
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
+++ b/SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
@@ -31,13 +31,20 @@ EnvironmentRedirect {
 		dispatch.value(key, obj);
 	}
 
+	removeAt { arg key;
+		var result = envir.removeAt(key);
+		dispatch.value(key, nil);
+		^result
+	}
+
 	localPut { arg key, obj;
 		envir.put(key, obj)
 	}
 
-	removeAt { arg key;
+	localRemoveAt { arg key;
 		^envir.removeAt(key)
 	}
+
 
 	// behave like environment
 
@@ -213,13 +220,22 @@ LazyEnvir : EnvironmentRedirect {
 
 	put { arg key, obj;
 		this.at(key).source_(obj);
-		dispatch.value(key, obj); // forward to dispatch for networking
+		dispatch.value(key, obj);
 	}
 
 	removeAt { arg key;
 		var proxy;
 		proxy = envir.removeAt(key);
 		if(proxy.notNil) { proxy.clear };
+		dispatch.value(key, nil);
+		^proxy
+	}
+
+	localRemoveAt { arg key;
+		var proxy;
+		proxy = envir.removeAt(key);
+		if(proxy.notNil) { proxy.clear };
+		^proxy
 	}
 
 	localPut { arg key, obj;


capital-G (changes requested):
Thanks for the change.

The code looks is consistent within the class, which I think is most important here (although the code-as-is has some quirks, e.g. dispatch is not an argument of the constructor? but out-of-scope of this PR).

I personally think this PR is lacking some documentation (see https://dev.docs.supercollider.online/Classes/EnvironmentRedirect.html and https://dev.docs.supercollider.online/Classes/LazyEnvir.html for current state), as this will also introduce a change in the API (dispatch will now also be called on deletion, but with an empty value to be aware of it).

Your PR already state some changes, at least these should be included in the docs.

> * return object on removeAt (this is standard in collections)
> * add localRemoveAt for removing without dispatch

~~It would be beyond the scope of this PR, but it would be nice to have an interesting example how this can be used in an interesting way as it seems a powerful tool for live coding.~~ => LayzeEnvir docs give a good hint how this can be used.

It would be good to have this rebased from dev to get a test run.

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
```diff
@@ -31,14 +31,21 @@ EnvironmentRedirect {
...
 	localPut { arg key, obj;
 		envir.put(key, obj)
 	}
 
-	removeAt { arg key;
+	localRemoveAt { arg key;
 		^envir.removeAt(key)
 	}
 
+
```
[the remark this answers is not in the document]
telephon: Unintended, thanks!

telephon:
> should EnvironmentRedirect run dispatch also when trying to remove a value that was not there? I think it shouldn't, since the Env didn't actually change. 

Yes, you are right, it shouldn't do that..

> However, I think it should run in the `put` case, even if the user is putting the same value that was already stored.

This is a little less clear, but I would also say that it should – this is also how it happens now, and I don't want to change that.


Btw. off topic: for LazyEnvir, I often had thought that it would be much better to have a configurable function that returns the new proxy instaead of the class. Maybe there is an ok way to do this, but that is for another issue.

elgiano:
> Yes, you are right, it shouldn't do that..
Maybe something like this?
```supercollider
removeAt { arg key;
    var hadKey = true;
    var result = envir.removeAtFail(key) { hadKey = false; nil };
    if (hadKey) { dispatch.value(key, nil) };
    ^result
}
```
Would you like me to push these contribs to your PR?

telephon:
yes, please go ahead. Thanks for the code – just cek again if everything that is needed in `removeAt` is there (I think you may need to add a clear).

elgiano:
>  (I think you may need to add a clear).

do you mean to class dispatch on clear? I don't see why we should call clear in removeAt (sorry!)

elgiano:
failed test are unrelated: `TestServer_boot:test_notifyAndServerBootActions`
(it doesn't fail on my machine when tested individually)

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
```diff
@@ -31,11 +31,18 @@ EnvironmentRedirect {
 		dispatch.value(key, obj);
 	}
 
+	removeAt { arg key;
+		var hadKey = true;
+		var result = envir.removeAtFail(key) { hadKey = false; nil };
```
capital-G, line 36: I personally find the postfix notation here not as clear as it could/should be? Or are there some performance gains here in this case?

At least it took me some minutes in order to grasp what's going on here, which is essentially

```supercollider
removeAt {|key|
	var result = envir.removeAtFail(
		key: key,
		// function will only be evaluated
		// if the key is not present
		function: {
			// bail out early with nil
			^nil;
		},
	);
	dispatch.value(key, nil);
	^result;
}
```

The comments seem to be necessary b/c of the lacking docs of `removeAtFail` and due to its generic arg name.

testsuite/classlibrary/TestEnvironmentRedirect.sc
```diff
@@ -0,0 +1,28 @@
...
+
+		notified = false;
+		e[\test] = nil;
+		this.assert(notified, "should run dispatch when an existing value is set to nil");
+		e[\test] = 2;
+
+		notified = false;
+		e.removeAt(\test);
+		this.assert(notified, "should run dispatch when an existing value is removed");
+
```
capital-G, line 23: maybe add an additional test that `\test` does indeed not trigger a callback after its deletion? I am sure it won't but that is what we are testing, right?

We should also check the arguments passed to the dispatcher?

capital-G (approved):

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
telephon: yes, that is good. I would still avoid keyword aruments in the class library if not necessary, because they are slower. Also here, they are redundant anyway. I'll add a fix.

telephon:
I've simplified, could you please take another look? (it was complicated because the branch was stale and I had to rebase, sorry!)

capital-G:
https://github.com/supercollider/supercollider/pull/5923/commits/471acba83b526e2282189cda1d58dead4b1b8a13 looks a bit off b/c you are merging your own branch into your own branch? At least the git history of this PR seems broken? :/

telephon:
sorry, rebased onto a not updated develop branch.

telephon:
now it should be ok (I think).

testsuite/classlibrary/TestEnvironmentRedirect.sc
telephon: ok, done in https://github.com/supercollider/supercollider/pull/5923/commits/730df88f55b1a55e8707d34afcbe9f5bb46219ba

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
```diff
@@ -31,11 +31,20 @@ EnvironmentRedirect {
...
 	}
 
+	removeAt { arg key;
+		var result = envir.removeAt(key);
+		if (result.notNil) { dispatch.value(key, nil) };
+		^result
+	}
+
+	// if used in a distributed system, these local methods act without
+	// calling the dispatch and thereby redistributing the data recursively
```
capital-G, line 41: Are distributed systems where an envir like an Ndef/Proxyspace of one client is in fact controlled by multiple clients? Is this a thing?

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
```diff
@@ -213,13 +222,27 @@ LazyEnvir : EnvironmentRedirect {
...
 	put { arg key, obj;
 		this.at(key).source_(obj);
-		dispatch.value(key, obj); // forward to dispatch for networking
+		dispatch.value(key, obj);
 	}
 
 	removeAt { arg key;
+		var proxy;
+		proxy = envir.removeAt(key);
+		if (proxy.notNil) {
```
capital-G, line 231: same question as above?

capital-G (approved):
As mentioned I'd appreciate if someone else could review this as I don't have much understanding about environments - maybe @elgiano or someone from [the class library team](https://github.com/supercollider/supercollider/wiki/Council-structure#class-library) can step in?

If nobody reviews it I think it is fine to merge as-is as it is written by a member of the class lib team which are free to merge non-controversial changes by themselves.

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
```diff
@@ -31,11 +31,20 @@ EnvironmentRedirect {
 		dispatch.value(key, obj);
 	}
 
+	removeAt { arg key;
+		var result = envir.removeAt(key);
+		if (result.notNil) { dispatch.value(key, nil) };
```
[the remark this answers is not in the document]
telephon: yes, this would be the correct way. We will need to require that the proxy class responds to `source`. 

We also need to decide one (new) thing: do we want `removeAt(\key)` be exactly equivalent to `put(\key, nil)` with regard to dispatch?

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
telephon: this was the original use. The implementation is in `JITLibExtensions` (`Public`).

SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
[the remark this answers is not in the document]
telephon: > I think the former implementation which used a callback function was able to differentiate between deleting a `nil` value and deleting a non-existing key.

I now believe that there should not be a difference between the two, with respect to callback.

adcxyz (approved):
just one unused variable to remove, all looks good to go!

adcxyz:
Oh sorry, did not see the squash on merge label!

PR #5920, opened 2022-11-27
Topic/upgrade boost 1.80.0

shimpe:
## Purpose and Motivation
Fixes #5865

## Types of changes
- Bug fix
- Upgrade of boost version to 1.80.0

## To-do list

- [X ] Code is tested (on a few of my own projects)
- [X ] All tests are passing
- [ ] Updated documentation
- [ X] This PR is ready for review

PR #5924, opened 2022-12-01
Topic/upgrade boost 1.80.0 unpatched

### LINKED ISSUE #5865: ERROR: Primitive '_FileCopy' failed. when starting scide - boost needs upgrade to higher version to fix
My /home is mounted on a different physical disk than /usr. 

When starting scide, I get the following error:
```
*** Welcome to SuperCollider 3.13.0-dev. *** For help press Ctrl-D.
Couldn't set realtime scheduling priority 1: Operation not permitted
SCDoc: Indexing help-files...
ERROR: Primitive '_FileCopy' failed.
caught exception 'Invalid cross-device link: copy from "/usr/local/share/SuperCollider/HelpSource/browse.css" to "/home/shimpe/.local/share/SuperCollider/Help/browse.css"' in primitive in method Meta_File:copy
RECEIVER:
class File (0x563d1fd95400) {
  instance variables [19]
    name : Symbol 'File'
    nextclass : instance of Meta_FileDialog (0x563d1f9419c0, size=19, set=5)
    superclass : Symbol 'UnixFILE'
    subclasses : nil
    methods : instance of Array (0x563d1fd95580, size=12, set=4)
    instVarNames : instance of SymbolArray (0x563d1fd95700, size=1, set=2)
    classVarNames : instance of SymbolArray (0x563d1fd95880, size=1, set=2)
    iprototype : instance of Array (0x563d1fd957c0, size=1, set=2)
    cprototype : instance of Array (0x563d1fd95940, size=1, set=2)
    constNames : nil
    constValues : nil
    instanceFormat : Integer 0
    instanceFlags : Integer 0
    classIndex : Integer 1077
    classFlags : Integer 0
    maxSubclassIndex : Integer 1077
    filenameSymbol : Symbol '/usr/local/share/SuperCollider/SCClassLibrary/Common/Files/File.sc'
    charPos : Integer 0
    classVarIndex : Integer 212
}

PROTECTED CALL STACK:
	Meta_MethodError:new	0x563d1f733700
		arg this = PrimitiveFailedError
		arg what = caught exception 'Invalid cross-device link: copy from "/usr/local/share/SuperCollider/HelpSource/browse.css" to "/home/shimpe/.local/share/SuperCollider/Help/browse.css"' in primitive in method Meta_File:copy
		arg receiver = File
	Meta_PrimitiveFailedError:new	0x563d1f73b540
		arg this = PrimitiveFailedError
		arg receiver = File
	Object:primitiveFailed	0x563d1f158d00
		arg this = File
	ArrayedCollection:do	0x563d20200ac0
		arg this = [ [ /usr/local/share/SuperCollider/HelpSource/BrokenLink.html, BrokenLink.html ], [ /usr/local/share/SuperCollider/HelpSource/browse.css, browse.css ], [ /usr/local/share/SuperCollider/HelpSource/Browse.html, Browse.html ], [ /usr/local/share/SuperCollider/HelpSource/browse.js, browse.js ], [ /usr/local/share/SuperCollider/HelpSource/codemirror.css, codemirror.css ], [ /usr/local/share/SuperCollider/HelpSource/editor.css, editor.css ], [ /usr/local/share/SuperCollider/HelpSource/editor.js, editor.js ], [ /u...etc...
		arg function = a Function
		var i = 1
	Meta_SCDoc:indexAllDocuments	0x563d1f65afc0
		arg this = SCDoc
		arg clearCache = false
		var now = 0.549377424
		var key = Classes/String
		var doc = SCDocEntry("Classes/QKeyModifiers", "QKeyModifiers", "(Undocumented class)")
		var nonHelpFiles = [ [ /usr/local/share/SuperCollider/HelpSource/BrokenLink.html, BrokenLink.html ], [ /usr/local/share/SuperCollider/HelpSource/browse.css, browse.css ], [ /usr/local/share/SuperCollider/HelpSource/Browse.html, Browse.html ], [ /usr/local/share/SuperCollider/HelpSource/browse.js, browse.js ], [ /usr/local/share/SuperCollider/HelpSource/codemirror.css, codemirror.css ], [ /usr/local/share/SuperCollider/HelpSource/editor.css, editor.css ], [ /usr/local/share/SuperCollider/HelpSource/editor.js, editor.js ], [ /u...etc...
		var undocClasses = IdentitySet[ BelaServerOptions, Range, PureMultiOutUGen, QDragSink, BinaryOpXStream, NodeID, DrawGridX, AbstractGroup, MIDIEvent, MethodQuote, QCurve, OutOfContextReturnError, QTextViewBase, QDialog, ScopeBuffer, FuncStreamAsRoutine, Penv, PatternControl, HIDFuncDeviceProtoMatcher, QCallback, QFont, QRenderProcessTerminationStatus, PlusFreqScopeWindow, HIDElementProtoDispatcher, MIDISMPTEAssembler, QScope, QScope2, CurveWarp, QColorGroup, BlankGridLines, QNumberBox, CleanupStream, TextArchiveMDPlugin, HIDRa...etc...
		var additions = Dictionary[ (Classes/String -> [ /usr/local/share/SuperCollider/HelpSource/Classes/String.ext.schelp ]) ]
	Meta_SCDoc:documents	0x563d1f65e980
		arg this = SCDoc
	Meta_SCDoc:prepareHelpForURL	0x563d1f662c00
		arg this = SCDoc
		arg url = file:///home/shimpe/.local/share/SuperCollider/Help/Help.html
		var path = /home/shimpe/.local/share/SuperCollider/Help/Help.html
		var targetBasePath = /home/shimpe/.local/share/SuperCollider/Help
		var pathIsCaseInsensitive = false
		var subtarget = Help
		var src = nil
		var c = nil
		var cmd = nil
		var doc = nil
		var destExist = false
		var destMtime = nil
		var verpath = /home/shimpe/.local/share/SuperCollider/Help/version
	a FunctionDef	0x563d1efa4400
	Function:prTry	0x563d1f9c1100
		arg this = a Function
		var result = nil
		var thread = a Routine
		var next = nil
		var wasInProtectedFunc = true
	
CALL STACK:
	MethodError:reportError
		arg this = <instance of PrimitiveFailedError>
	< closed FunctionDef >
		arg error = <instance of PrimitiveFailedError>
	Integer:forBy
		arg this = 0
		arg endval = 0
		arg stepval = 2
		var i = 0
		var j = 0
		arg this = [*2]
	Scheduler:seconds_
		arg newSeconds = 0.742669093
	Meta_AppClock:tick
	Process:tick
		arg this = <instance of Main>
^^ The preceding error dump is for ERROR: Primitive '_FileCopy' failed.
caught exception 'Invalid cross-device link: copy from "/usr/local/share/SuperCollider/HelpSource/browse.css" to "/home/shimpe/.local/share/SuperCollider/Help/browse.css"' in primitive in method Meta_File:copy
RECEIVER: File
```

shimpe:
## Purpose and Motivation

fixes #5865

## Types of changes
- bugfix by upgrade to boost 1.80.0

## To-do list

- [ X] Code is tested (very quick "first" test only - I will be using this version for a while)
- [X ] All tests are passing
- [ ] Updated documentation
- [ X] This PR is ready for review


dyfer:
hm, I guess that answers it... seems that MinGW still needs that patch. Thank you for checking!

shimpe:
ouch... internal compiler error in CI / Windows 64-bit MingW 

D:/a/supercollider/supercollider/external_libraries/boost/boost/thread/detail/invoke.hpp:101:43: internal compiler error: in gimplify_expr, at gimplify.c:12188
         return (*boost::forward<A0>(a0)).*f;

shimpe:
I've applied only the part of the patch that changes something in boost thread (which is what triggered the internal compiler error). Not sure why the MacOS shared libscsynth now failed. I don't see any clear compiler error in the logs.

dyfer:
> Not sure why the MacOS shared libscsynth now failed. I don't see any clear compiler error in the logs.

Yeah, it's weird, it might be a random error. I'm re-running the CI to check.

dyfer:
Note to self: my main concern was whether this doesn't impede compatibility with older platforms. I tested the macOS legacy build from this PR on macOS 10.10 and it still runs (server boots, `File` operations appear to work).

I'll let others chime in before merging this, but I think this should be good to go. Also, I think we should use this "minimally patched" version as opposed to #5920.

CMakeLists.txt
```diff
@@ -145,8 +145,11 @@ if(WIN32)
     add_definitions(-DBOOST_INTERPROCESS_BOOTSTAMP_IS_SESSION_MANAGER_BASED)
 endif()
 
-add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE)
+add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE -DBOOST_FILESYSTEM_NO_CXX20_ATOMIC_REF)
 
+if(APPLE)
+    add_definitions("-fno-aligned-allocation")
+endif()
```
dyfer, line 152: I think this should be applied more selectively - only on affected platforms (macOS <10.14) and probably only for affected code. 
This seems to work: https://github.com/dyfer/supercollider/commit/f458244f72302a0c045b7ee7ba171db60a56a121

CMakeLists.txt
```diff
@@ -145,8 +145,11 @@ if(WIN32)
     add_definitions(-DBOOST_INTERPROCESS_BOOTSTAMP_IS_SESSION_MANAGER_BASED)
 endif()
 
-add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE)
+add_definitions(-DBOOST_CHRONO_HEADER_ONLY -DBOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE -DBOOST_FILESYSTEM_NO_CXX20_ATOMIC_REF)
```
dyfer, line 148: Do you have any idea why we have to set this manually? It seems that it should be [detected automatically](https://github.com/boostorg/filesystem/blob/boost-1.80.0/CMakeLists.txt#L35), but then I don't see `has_cxx20_atomic_ref.cpp` in the extracted boost source...

dyfer (changes requested):
Upon a closer look I have a couple of requests here.

CMakeLists.txt
shimpe: I think you are right. As you can see from the history in the "patched" version I was kind of poking in the dark since I have zero experience with MacOS based systems.

CMakeLists.txt
shimpe: I suspect bcp is not smart enough to include has_cxx20_atomic_ref.cpp because it is only referenced in the CMakeLists.txt and not in the actual source code?

dyfer: Not sure if that's relevant, but a shot in the dark - did you use bcp from boost 1.80?

shimpe: I did indeed. I didn't find any package for bcp so had to build it myself.

shimpe: @dyfer Are any more actions expected from me?

dyfer: @shimpe thanks for checking - nothing at this time, thanks. 
We might want to pull it into 3.13.1 (not sure yet) so it's a matter of timing this right.

dyfer:
I think this is technically good to go, but I would like to wait before merging this to see how we deal with FluCoMa and possibly extending the plugin interface. I'd prefer to merge this once we at least have a plan to upgrade the plugin interface to accommodate FluCoMa needs (see https://github.com/flucoma/flucoma-sc/issues/149)

In any case, I think this should be (eventually) merged into `develop` to be included in the 3.14 release, not earlier.

shimpe:
I'm ok with waiting since I can anyway use my own build until it gets merged.

dvzrv:
Since we are building against a 1.80.0 boost (system) version on Arch Linux already, I guess flucoma-sc is broken there already. :thinking:

dyfer:
If Flucoma is built against the same boost version, then all is fine...
However, I hope that we'll fix the need to link to boost by plugins by extending plugin API https://github.com/supercollider/supercollider/issues/5939

dvzrv:
> If Flucoma is built against the same boost version, then all is fine... However, I hope that we'll fix the need to link to boost by plugins by extending plugin API #5939

Ah, alright. I'm not yet providing it in the repos. Maybe some day though!

shimpe:
@dyfer  Since several people have hit the same filesystem bug, and since no one seems to be working on the plugin API (correct me if I'm wrong), does it still make sense to keep waiting?

Spacechild1:
> does it still make sense to keep waiting?

IMO there's no reason to wait. Frankly speaking, if plugin authors are depending on *private* classes/functions, it's really their problem. The very point of private APIs is that we can change them whenever we want.

(To be clear, I'm all for extending the public plugin API, because I'm suffering from its limitations myself, but that's a completely different issue.)

dyfer (dismissed):
Yes, I think we should proceed, enough time has passed. Maybe once the plugins break, it'll encourage work on the plugin api...

capital-G:
Is it ok if we close this in favor of #6334 ?

shimpe:
Sure go ahead, as long is doesn't take another year. My only concern is getting the bug finally fixed :D

Issue #5931, opened 2022-12-10
Broken bytecode emission


LookMumNoBackup:
## Environment

* SuperCollider version: 3.13.0-rc1
* Operating system: Ubuntu 22.04.1 LTS (jammy)
* Other details (Qt version, audio driver, etc.): all dependecies fetched from official operation system repository. Build locally with CMake.

## Steps to reproduce

- Open SuperCollider IDE
  - Interpreter starts automatically
- Type in following code

```supercollider
(
g = { arg a, b; ^(a + b) };
g.def.dumpByteCodes;
g.value(5,10);
)
```

Receive following output in post window

```supercollider
BYTECODES: (6)
  0   30		 PushTempZeroVar 'a'
  1   31		 PushTempZeroVar 'b'
  2   FF       TailCallReturnFromMethod
  3   E0       SendSpecialBinaryArithMsg '+'
  4   F3       Return
  5   F2       BlockReturn
```

### Expected
Option (1): ```15``` is printed out in the post window after byte-code dump.
Option (2): Syntax error as ```^``` seems to be not allowed in this context.
Option (3): Exception thrown at runtime, that something is wrong with the function.

### Actual
Neither the result is printed, nor any kind of error thrown. Any further code in the same region seems to be silently ommited. E.g. following code do not change the output in the post window after clearing:

```supercollider
(
g = { arg a, b; ^(a + b) };
g.def.dumpByteCodes;
g.value(5,10);
10.value;
)
```

**Note**: I am just starting to getting warm with SC and though do not speak this language fluently. Somehow I though after looking at different class definitions that the ```^``` symbol syntactically means return. Now it seems to me that it is only valid in class definitions. Please correct me if I am wrong.

dyfer:
> Now it seems to me that it is only valid in class definitions. 

Yes, that is correct.

LookMumNoBackup:
@dyfer Thanks for the fast feedback. I think this behaviour is at least worth being documented, if fixing is too cumbersome. If I get a hint where to place this information and which branch to use, I can create a corresponding pull request.

dyfer:
Yes, it is certainly worth being documented. I was sure it was, but indeed I can't find it.

I think it would be worth mentioning it in one or all of:
- https://github.com/supercollider/supercollider/blob/develop/HelpSource/Reference/Functions.schelp
- https://github.com/supercollider/supercollider/blob/develop/HelpSource/Classes/Function.schelp
[...]

Issue #2674, opened 2017-01-21
"^" is legal when executing code interactively


snappizz:
The method return token is legal in functions in the interactive interpreter. It behaves strangely. For example, this code doesn't post anything at all:

    (
    { { ^4 } }.value.value;
    "hey".postln;
    )

brianlheim:
Even simpler reproducer:
```supercollider
^4
```
vs
```supercollider
^4; "test"; // parse error
```

`^x` (with an optional `;`) is allowed in any executed code so long as it's the last bit of the executed line.

```supercollider
(
var arr = Array.rand(10, 1, 8);
arr.sum.postln;
arr.product.postln;
arr.collect(_.bitNot).sum.postln;
^4
)
```

Pretty sure this is another trivial parser fix, provided we want to change it.

snappizz:
Is there any good reason to use `^` in interactive code?

jamshark70:
Gut feeling:

1. `^` in an interactively-defined function doesn't behave like a true C-style `return`, but the lack of a parser error for it leads the user to expect it. So, we should forbid it.

2. I don't think there's a good reason to write it in interactive code, because it exits from the last method called rather than from the containing block.

3. Currently the only way to exit an interactively-defined function early is to fake it with `block`, which is inconvenient. Would be nice to address that (but also perhaps risky).

snappizz:
The presence of `^` in interactive code probably means that the user copy-pasted something from the class library and forgot to remove `^`. That's how I found this in the first place.

PR #5932, opened 2022-12-11
Add doc for issue #2674

LookMumNoBackup:
## Purpose and Motivation

Document strange behavior when using caret symbol ```^``` in functions which are not part of a class definition.
Reported in issue #5931 and than had been identified as duplicate of issue #2674

As discussed with @dyfer at least documentation of the behavior is needed. This pull request put corresponding notes within three help pages suggested by @dyfer.

## Types of changes

- Documentation

## To-do list

- [x] Local build had been performed
- [x] SuperCollider IDE shows the notes within the internal browser

### DIFF
--- a/HelpSource/Classes/Function.schelp
+++ b/HelpSource/Classes/Function.schelp
@@ -33,8 +33,10 @@ code::
 NOTE:: 
 In case you already know how to define your own classes, you came accross the usage
 of teletype::^:: within methods to return values back to the caller. Please be aware
-that usage of teletype::^:: in functions not part of classes is not implemented yet
+that usage of teletype::^:: in functions not part of classes is not implemented.
 and will most likely behave in a non-intuitive way!
+
+Consider using the teletype::block:: method in case you really need the semantic of teletype::^:: in this context.
 ::
 
 subsection:: Related Keywords
--- a/HelpSource/Classes/Function.schelp
+++ b/HelpSource/Classes/Function.schelp
@@ -30,6 +30,13 @@ code::
 ((lambda (x) x) 1) // Lisp
 ::
 
+NOTE:: 
+In case you already know how to define your own classes, you came accross the usage
+of teletype::^:: within methods to return values back to the caller. Please be aware
+that usage of teletype::^:: in functions not part of classes is not implemented yet
+and will most likely behave in a non-intuitive way!
+::
+
 subsection:: Related Keywords
 
 method:: thisFunction
--- a/HelpSource/Reference/Functions.schelp
+++ b/HelpSource/Reference/Functions.schelp
@@ -27,6 +27,13 @@ code::
 {}.value.postln;
 ::
 
+NOTE:: 
+In case you already know how to define your own classes, you came accross the usage
+of teletype::^:: within methods to return values back to the caller. Please be aware
+that usage of teletype::^:: in functions not part of classes is not implemented yet
+and will most likely behave in a non-intuitive way!
+::
+
 A function can be thought as a machine able to perform a task on demand, e.g. a calculator. The calculator can receive input (args) and can output a value, the result of the performed operations. The function definition can then be thought as the building of the calculator: once built, the calculator does nothing until it is requested to work (by passing the value method to a function).
 The following figure depicts an empty function, input without output, output without input, and the general case with input and output.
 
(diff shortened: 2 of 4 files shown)


telephon (commented):
Thanks for adding the note!
In a sense, "not allowed" is misleading. It is allowed, it just behaves a little weird if you don't think it through. Effectively, it causes a return from the current method, namely `Interpreter:functionCompileContext`:

```

g = { ^thisMethod.postln };
g.value;
)
```

telephon:
But indeed it behaves not as one may expect:
```

(
f = {
	var g = { ^thisMethod };
	g.value;
}
)

f.value;
```

LookMumNoBackup:
> But indeed it behaves not as one may expect:
> 
> ```
> 
> (
> f = {
> 	var g = { ^thisMethod };
> 	g.value;
> }
> )
> 
> f.value; // returns nothing. Why? well, because before the (implicit) postln it has exited already
> ```

@telephon Many thanks for this clarification! Looking forward to learn more about the SC language!

HelpSource/Classes/Function.schelp
```diff
@@ -30,6 +30,13 @@ code::
 ((lambda (x) x) 1) // Lisp
 ::
 
+NOTE:: 
+In case you already know how to define your own classes, you came accross the usage
+of teletype::^:: within methods to return values back to the caller. Please be aware
+that usage of teletype::^:: in functions not part of classes is not implemented yet
```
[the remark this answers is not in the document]
LookMumNoBackup: @telephon Thank you for the feedback. I hope than my new commit is now complete for a merge. In case further corrections are needed, feel free to add them. Unfortunately I will have only a limited access to my system in the next weeks and that`s why can not estimate when I can add further changes.

HelpSource/Classes/Function.schelp
```diff
@@ -30,6 +30,15 @@ code::
 ((lambda (x) x) 1) // Lisp
 ::
 
+NOTE:: 
+In case you already know how to define your own classes, you came accross the usage
+of teletype::^:: within methods to return values back to the caller. Please be aware
+that usage of teletype::^:: in functions not part of classes is not implemented.
+and will most likely behave in a non-intuitive way!
+
+Consider using the teletype::block:: method in case you really need the semantic of teletype::^:: in this context.
```
dyfer, line 39: I don't think we need the word "really" here. Here's a suggestion (feel free to adjust it):

"Consider using link::Classes/Function#-block:: if you need to return early from the function."

(please check if the link works correctly, I haven't tested it)

Also, I'm not sure if that sentence would work better inside or outside of the `NOTE` block.

HelpSource/Reference/Functions.schelp
```diff
@@ -27,6 +27,15 @@ code::
...
 ::
 
+NOTE:: 
+In case you already know how to define your own classes, you came accross the usage
+of teletype::^:: within methods to return values back to the caller. Please be aware
+that usage of teletype::^:: in functions not part of classes is not implemented.
+and will most likely behave in a non-intuitive way!
+
+Consider using the teletype::block:: method in case you really need the semantic of teletype::^:: in this context.
+::
```
dyfer, line 37: I have the same suggestions as in the Function class helpfile.

dyfer (changes requested):
Thanks for working on this!

I have a few suggestions that I think would make notes more clear or accurate. They're not the "final versions" so feel free to modify them.

dyfer:
@LookMumNoBackup I wanted to check in regarding the suggested changes, thanks!

dyfer:
FYI if this PR won't be updated soon, it won't make it to 3.13.0. That's fine with me, but I wanted to let everyone know.

capital-G:
@LookMumNoBackup still interested in getting this merged? Would be appreciated!

LookMumNoBackup:
Hi Dennis,

sorry but I am completely occupied by my own project where I write a
programmable sequencer in Go in my free time. Feel free to contact me if
that topic is interessting to you.

Am Fr., 30. Aug. 2024 um 11:41 Uhr schrieb Dennis Scheiba <
***@***.***>:

capital-G:
@LookMumNoBackup thanks for the quick reply! 

@dyfer @telephon should we close this in this case?

dyfer:
I can go either way on this - we can close this, or if you think the additional clarification is desirable, I'd apply my changes and merge it.

PR #5938, opened 2022-12-19
Add support for Linux AppImage (Linux standalone and distribution of binaries for any Linux distribution)

miguel-negrao:
## Purpose and Motivation

This PR adds the possibility of building SuperCollider as a [Linux appimage](https://appimage.org/) (via LINUX_APPIMAGE cmake option), a selfcontained file with all the needed libraries and resources, similar to a macos bundle. This enables:

- Building SuperCollider as an appimage with the CI and make it available for download on S3 or github releases. Any linux distribution with a recent enough kernel, libc and fuse installed should be able to run the executable.
- Run SuperCollider in "portable" mode by having a SuperCollider.AppImage.home folder next to the executable. This implements the historical "standalone" feature of SC. In this mode the binary will not look at anything in the home directory of the user, allowing for self-contained and distributable files.

While the CI is building this PR the appimage can already be tested by downloading from the artifacts section [here](https://github.com/miguel-negrao/supercollider/actions/runs/3730334680).

Some notes:
- Because we just remap the HOME folder completely to SuperCollider.AppImage.home there is no need to use or implement "standalone" modes in scide or sclang. Users should put their code inside SuperCollider.AppImage.home
- I have not used the work done in the [Standalone PR for macos bundles](https://github.com/supercollider/supercollider/pull/5898) since the approach here is a bit different. Discussion is needed in terms of how these two approaches should be implemented and made to co-exist. My approach so far is to make it two completely different things.
- For proof of concept currently I'm building in the CI and uploading as an artifact but I'm not uploading to S3. Can someone remind me, why do we need S3. Isn't github artifacts enough ?
- For a distributable standalone supercollider project the language config file should not be used as it has hard-coded paths. Instead all code should go to the SuperCollider.AppImage.home/.local/share/SuperCollilder/Extensions folder. Development can be done using the lang config file, but for the final version "Extensions" should be used.

Before I polish this further I would like to request feedback on this approach. Does this look good to you all ?

## Types of changes

- New feature

## To-do list

- [ ] Implement functionality for supernova
- [ ] Use S3 instead of github artifacts ?
- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
Thanks @miguel-negrao for working on this!
I'm really glad that there is interest in working on the standalone version of SC. Ultimately, I would like SC to provide standalone versions on all supported platforms (Linux/macOS/Windows).

Here is my initial feedback:

> * I have not used the work done in the [Standalone PR for macos bundles](https://github.com/supercollider/supercollider/pull/5898) since the approach here is a bit different. Discussion is needed in terms of how these two approaches should be implemented and made to co-exist. My approach so far is to make it two completely different things.

I have been lurking at #5898 but have not digested that fully. In principle, I would like the approaches to be "synchronized" between platforms, at least as much as it's feasible, but I can't fully evaluate this before digging deeper into both approaches.


> * Can someone remind me, why do we need S3. Isn't github artifacts enough ?

Artifacts were uploaded to S3 when we were building in Travis and AppVeyor (I believe S3 was originally set up by @scztt) and were later ported to GitHub actions in the interest of continuity. In particular, downloading directly from the artifacts page requires one to be logged into GitHub, AFAIR (this does not apply to downloads from releases, these are available without logging in).  IMO it's totally fine to have these on GHA, but it would need to be also uploaded to the release page (right now it's only uploaded to the GHA artifacts page)

> * For a distributable standalone supercollider project the language config file should not be used as it has hard-coded paths.

Do you mean that we can only use absolute paths in language config? Or some other way of hardcoding paths? IMO we should start supporting relative paths in the language config file, but I'm not sure if it should be included in this PR, as opposed to e.g. #5898.

I also have some feedback on the actual github actions implementation:
- I would prefer if the AppImage build script was actually implemented as a step in the actions yml file, as opposed to be a separate script
  - I'd be fine with not providing the actual `.sh` script, but I can also see that it might be useful for users trying to build their own standalone
  - I understand that having both the script and have the steps spelled out in the yml file causes code duplication, but I try to avoid having 3rd party scripts called for build-related jobs, in the interest of clarity
- I have a slight preference for the job to be part of the Linux build matrix, as opposed to be a separate entry
  - I think that implementation as part of the matrix would be cleaner, removing some code duplication to set up the build environment
  - On the other hand, I feel like failure in AppImage, which could happen e.g. due to unavailable upstream resources, shouldn't block running tests or uploading to the release, which could happen if it's part of the build matrix; keeping it as a separate job would allow the granularity to not depend on AppImage build for release uploads etc
- Whether it's a separate job or a matrix entry, let's have it right after the other Linux builds
- I assume it's impossible to avoid pulling the AppImage template from `linuxdeploy`, correct? Asking in case it'd be possible  to include it e.g. as a submodule
- IMO the artifact name should indicate SC version (see `{{ needs.lint.outputs.sc-version }}`)
- as indicated earlier, the AppImage should be set up to be uploaded to the release, see [here](https://github.com/supercollider/supercollider/pull/5938/files#diff-2702f6f7d10a1bffbc8596d1a9925b1234bc4c7a6ff4c21ca6b34c13aa75780eL987-L995)

Again, thanks so much for working on this!

miguel-negrao:
> Do you mean that we can only use absolute paths in language config? Or some other way of hardcoding paths? IMO we should start supporting relative paths in the language config file, but I'm not sure if it should be included in this PR, as opposed to e.g. #5898.

@dyfer  thanks for the feedback ! Let me answer this one first: I didn't change anything relating with the language config file, therefore all previous limitations still apply. One such limitation is that it doesn't supporte relative paths (or possibly those get resolved to the CWD? don't remember anymore...). If someone wants to add the relative paths functionality I would prefer if that was a separate PR, as that is not really needed for the appimage functionality, given that the users can use the extensions folder in the SuperCollider.AppImage.home folder.

PR #5942, opened 2022-12-25
classlib: Plotter: extend tick labels

mtmccrea:
## Purpose and Motivation

This follows  up #5161, in response to [feedback](https://github.com/supercollider/supercollider/pull/5161#issuecomment-1321457870) about retaining previous functionality in how the tick labels are handled in `DrawGrid`.

In particular this adds functionality so that you can achieve the same look as previous version of SC:
- Move and align tick axis labels
- Append a string to axis labels
- Constrain labels to the edges of the grids
- en/disable labels independently
- set the Font and Color of tick and axis labels independently

~~This PR is unfinished but as it's blocking the next RC, I'm posting it early so those involved can comment on what's been done so far. @dyfer @sternsc.~~

This required basically a complete overhaul of how the new tick/axis label spacing is calculated so that everything resizes properly. ~~As such, I have a good amount of testing to do.~~

This PR is now ready for review (I'll clean up the commit history at the final stage).

Depending on the timeline of the RC, documentation won't be the priority, but I'll submit that as a separate PR later if the RC makes it out before I get to it.

Here's a few examples of the new interface to set `Plot` and `DrawDrid` properties from `Plotter`.
```supercollider
(     // default behavior plus axis labels
p = [10, 34, 1672].collect({ |f|
	101.collect{ |i| sin((i/1000) * 2pi * f) }
}).plot;
p.axisLabelY_("Amp")
.axisLabelX_((1..3).collect{ |elem| "X label " ++ elem })
)

(    // vertically compact
p = [10, 34, 1672].collect({ |f|
	101.collect{ |i| sin((i/1000) * 2pi * f) }
}).plot;
p.setGridProperties(\y,
	\labelAnchor, \right, \labelAlign, \center,
	\constrainLabelExtents, true, \labelOffset, -3 @ 0
).setGridProperties(\x,
	\labelAnchor, \leftBottom, \labelAlign, \left,
	\constrainLabelExtents, true, \labelOffset, 3 @ -1,
	\fontColor, Color.red, \labelAppendString, " sec"
).specs_([-1.5, 1.5].asSpec) // give data headroom to see inlaid y tick labels
)

(    // minimal
p = [10, 34, 1672].collect({ |f|
	101.collect{ |i| sin((i/1000) * 2pi * f) }
}).plot;
p.setProperties( \showTickLabelsX, false, \showTickLabelsY, false)
)
```
The three above examples produce these plots:
Screen Shot 2023-01-03 at 21 56 20

Note that because I opted not to add convenience methods to `Plotter` that would allow direct access to setting grid properties at this time, that is accomplished with the `Plotter:-setGridProperties`, which mimics the behavior of `Plotter:-setGridProperties:-setProperties`, in that it accepts a list of key-value pairs. 

The advantage is that setting multiple parameters (as is likely when customizing your grid) is more compact than multiple method calls to individual parameters for each axis. The downside is the user doesn't get the benefit of auto-complete for hints about to the available parameters. So I'll add a section to the docs in the future that will list the grid and plot parameters.

## Types of changes

- Documentation
- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation (More to follow this PR)

### DIFF
--- a/SCClassLibrary/Common/GUI/Base/Grid.sc
+++ b/SCClassLibrary/Common/GUI/Base/Grid.sc
@@ -16,9 +16,9 @@ DrawGrid {
 		this.gridColors = [Color.grey(0.7), Color.grey(0.7)];
 	}
 	bounds_ { arg b;
-		bounds = b;
-		x.bounds = b;
-		y.bounds = b;
+		bounds = b.asRect;
+		x.bounds = bounds;
+		y.bounds = bounds;
 	}
 	draw {
 		Pen.push;
--- a/SCClassLibrary/Common/GUI/Base/Grid.sc
+++ b/SCClassLibrary/Common/GUI/Base/Grid.sc
@@ -66,9 +66,8 @@ DrawGrid {
 
 	// make a Window with a UserView that draws this DrawGrid
 	preview {
-		var insetH = 45, insetV = 35; // left, bottom margins for labels
+		var insetH = 30, insetV = 15; // left, bottom margins for labels
 		var gridPad = 15;             // right, top margin
-		var axLabelPad = 3;           // label offset from window's edge
 		var win, winBounds, font, fcolor;
 
 		// refresh and return the view if it already exists
@@ -85,7 +84,7 @@ DrawGrid {
 
 		// bounds of the grid lines, without its labels
 		this.bounds = this.bounds ?? { Rect(0, 0, 500, 400) };
-		// translate the grid to make room for axis & tick labels
+		// translate the grid to make room for tick labels
 		this.bounds = this.bounds.moveTo(insetH-gridPad, gridPad);
 
 		winBounds = this.bounds + Size(insetH, insetV);
@@ -98,33 +97,7 @@ DrawGrid {
 			win, winBounds.size.asRect
 		)
 		.drawFunc_({ |uv|
-			var unitsStr;
-
-			// draw the drid
-			this.draw;
-
-			// draw x-axis label
-			unitsStr = this.x.grid.spec.units;
-			if(unitsStr.size > 0) {
-				Pen.push;
-				Pen.translate(this.bounds.center.x, uv.bounds.bottom);
-				Pen.stringCenteredIn(unitsStr,
-					unitsStr.bounds.center_(0@0).bottom_(axLabelPad.neg),
-					font, fcolor);
-				Pen.pop;
-			};
-
-			// draw y-axis label
-			unitsStr = this.y.grid.spec.units;
-			if(unitsStr.size > 0) {
-				Pen.push;
-				Pen.translate(0, this.bounds.center.y);
-				Pen.rotateDeg(-90);
-				Pen.stringCenteredIn(unitsStr,
-					unitsStr.bounds.center_(0@0).top_(axLabelPad),
-					font, fcolor);
-				Pen.pop;
-			};
+			this.draw; // draw the drid
 		})
 		.onResize_({ |uv|
 			this.bounds = uv.bounds
--- a/SCClassLibrary/Common/GUI/Base/Grid.sc
+++ b/SCClassLibrary/Common/GUI/Base/Grid.sc
@@ -97,7 +97,7 @@ DrawGrid {
 			win, winBounds.size.asRect
 		)
 		.drawFunc_({ |uv|
-			this.draw; // draw the drid
+			this.draw;
 		})
 		.onResize_({ |uv|
 			this.bounds = uv.bounds
@@ -124,6 +124,7 @@ DrawGridX {
 	var <tickSpacing = 64;
 	var <numTicks = nil; // nil for dynamic numTicks with view size
 	var <>showLabels = true; // tick labels
+	var <>labelsHiddenBySize = false;
 	var <labelAppendString = nil;
 	var <drawBoundingRect = true, <drawBaseLine = false, <drawBoundingLines = false;
 
@@ -295,7 +296,7 @@ DrawGridX {
 			};
 
 			// Tick labels
-			if(showLabels and: { p['labels'].notNil }, {
+			if(showLabels and: { labelsHiddenBySize.not and: { p['labels'].notNil } }) {
 				commands = commands.add(['font_',font ] );
 				commands = commands.add(['color_',fontColor ] );
 
@@ -339,7 +340,7 @@ DrawGridX {
 						commands = commands.add([alignSym, val[1], labelRect]);
 					}
 				}
-			});
+			};
 
 			commands // return
 		}
@@ -349,13 +350,18 @@ DrawGridX {
 	// grid, given its size, anchoring, and offset. Userful when laying out
 	// this grid in a UserView to allowing room for labels.
 	labelOverhang { |labelStr|
+		var strRect, rect;
 		var overhang = [0, 0, 0, 0]; // left, top, right, bottom
-		var strRect = if(labelStr.notNil) {
+
+		if (showLabels.not) {
+			^overhang
+		};
+		strRect = if(labelStr.notNil) {
 			labelStr.bounds(font);
 		} {
 			this.labelSize.asRect
 		};
-		var rect = strRect.anchorTo(labelOffset, labelAnchor);
+		rect = strRect.anchorTo(labelOffset, labelAnchor);
 
 		// top side: assumes label doesn't extend past top bound
 		overhang[3] = abs(max(rect.bottom, 0)); // bottom
@@ -493,7 +499,7 @@ DrawGridY : DrawGridX {
 				};
 			};
 
-			if(showLabels and: { p['labels'].notNil }, {
+			if(showLabels and: { labelsHiddenBySize.not and: { p['labels'].notNil } }) {
 				commands = commands.add(['font_',font ] );
 				commands = commands.add(['color_',fontColor ] );
 
@@ -536,7 +542,7 @@ DrawGridY : DrawGridX {
 					);
 					commands = commands.add([alignSym, labelStr, labelRect]);
 				};
-			});
+			};
 
 			commands // return
 		}
@@ -544,9 +550,15 @@ DrawGridY : DrawGridX {
 
 	// See description in DrawGridX:-labelOverhang
 	labelOverhang { |labelStr|
+		var strRect, rect;
 		var overhang = [0, 0, 0, 0]; // left, top, right, bottom
-		var strRect = if(labelStr.notNil) { labelStr.bounds(font) } { this.labelSize.asRect };
-		var rect = strRect.anchorTo(labelOffset, labelAnchor);
+
+		if (showLabels.not) {
+			^overhang
+		};
+
+		strRect = if(labelStr.notNil) { labelStr.bounds(font) } { this.labelSize.asRect };
+		rect = strRect.anchorTo(labelOffset, labelAnchor);
 
 		// right side: assume label doesn't extend past right bound
 		overhang[0] = abs(min(rect.left, 0)); // left
--- a/SCClassLibrary/Common/GUI/Base/Grid.sc
+++ b/SCClassLibrary/Common/GUI/Base/Grid.sc
@@ -125,7 +125,7 @@ DrawGridX {
 	var <numTicks = nil; // nil for dynamic numTicks with view size
 	var <>showLabels = true; // tick labels
 	var <>labelsHiddenBySize = false;
-	var <labelAppendString = nil;
+	var <labelAppendString = nil, <labelsShowUnits = false;
 	var <drawBoundingRect = true, <drawBaseLine = false, <drawBoundingLines = false;
 
 
@@ -214,7 +214,26 @@ DrawGridX {
 	}
 	labelAppendString_ { |str|
 		this.grid.appendLabel_(str);
+		labelAppendString = str;
+		labelsShowUnits = false;
 		this.clearCache;
+	}
+	labelsShowUnits_ { |bool|
+		var specUnits;
+		if (bool) {
+			specUnits = grid.spec.units;
+			if(specUnits.isEmpty.not and: {
+				labelsShowUnits or: { this.labelAppendString.isNil }
+			}) {
+				this.labelAppendString = " " ++ specUnits;
+				labelsShowUnits = true;
+			};
+		} {
+			if(labelsShowUnits) {
+				this.labelAppendString = nil;
+				labelsShowUnits = false;
+			}
+		};
 	}
 	constrainLabelExtents_ { |bool|
 		constrainLabelExtents = bool;
@@ -322,19 +341,19 @@ DrawGridX {
 									labelRect = labelRect.right_(
 										min(labelRect.right, rightBound - constrainedPad)
 									);
-									alignSym = this.prGetJustString(\right);
+									alignSym = this.prGetJustifySymbol(\right);
 								},
 								leftBound, {
 									labelRect = labelRect.left_(
 										max(labelRect.left, leftBound + constrainedPad)
 									);
-									alignSym = this.prGetJustString(\left);
+									alignSym = this.prGetJustifySymbol(\left);
 								}, {
-									alignSym = this.prGetJustString(labelAlign)
+									alignSym = this.prGetJustifySymbol(labelAlign)
 								}
 							);
 						}, {
-							alignSym = this.prGetJustString(labelAlign);
+							alignSym = this.prGetJustifySymbol(labelAlign);
 						});
 
 						commands = commands.add([alignSym, val[1], labelRect]);
@@ -394,9 +413,6 @@ DrawGridX {
 			protoLabel = "0".dup(
 				try { labels.flatten.collect(_.size).maxItem } { 4 }
 			).join;
-			labelAppendString !? {
-				protoLabel = protoLabel ++ labelAppendString
-			};
 
 			labelSz = try {
 				protoLabel.bounds(font).asSize
@@ -423,7 +439,7 @@ DrawGridX {
 		^cmds
 	}
 
-	prGetJustString { |alignSym|
+	prGetJustifySymbol { |alignSym|
 		^switch(alignSym,
 			\center, { 'stringCenteredIn' },
 			\left,   { 'stringLeftJustIn' },
@@ -452,7 +468,7 @@ DrawGridY : DrawGridX {
 
 	commands {
 		var p;
-		var labelSz, labelRect, labelStr, anchorPoint, alignSym;
+		var labelSz, labelRect, anchorPoint, alignSym;
 		var y, valNorm, lineColor;
 		var bottomBound = bounds.bottom.asInteger;
 		var topBound = bounds.top.asInteger;
@@ -535,12 +551,8 @@ DrawGridY : DrawGridX {
 						);
 					};
 
-					alignSym = this.prGetJustString(labelAlign);
-					labelStr = if(labelAppendString.isNil,
-						{ val[1].asString },
-						{ val[1].asString ++ labelAppendString }
-					);
-					commands = commands.add([alignSym, labelStr, labelRect]);
+					alignSym = this.prGetJustifySymbol(labelAlign);
+					commands = commands.add([alignSym, val[1], labelRect]);
 				};
 			};
 
@@ -654,7 +666,7 @@ AbstractGridLines {
 		};
 		^if(appendLabel.notNil) { str = str ++ appendLabel } { str }
 	}
-	appendLabel_ { |str| appendLabel = str.asString }
+	appendLabel_ { |str| appendLabel = str !? { str.asString } }
 }
 
 LinearGridLines : AbstractGridLines {
--- a/SCClassLibrary/Common/GUI/Base/Grid.sc
+++ b/SCClassLibrary/Common/GUI/Base/Grid.sc
@@ -8,8 +8,8 @@ DrawGrid {
 		^super.new.init(bounds, horzGrid, vertGrid)
 	}
 	init { arg bounds, h, v;
-		x = DrawGridX(h);
-		y = DrawGridY(v);
+		x = DrawGridX(h, bounds.asRect);
+		y = DrawGridY(v, bounds.asRect);
 		this.bounds = bounds;
 		this.font = Font(Font.defaultSansFace, 9);
 		this.fontColor = Color.grey(0.3);
@@ -22,11 +22,11 @@ DrawGrid {
 	}
 	draw {
 		Pen.push;
-			Pen.alpha = opacity;
-			Pen.smoothing = smoothing;
-			if(linePattern.notNil) {Pen.lineDash_(linePattern)};
-			x.commands.do({ arg cmd; Pen.perform(cmd) });
-			y.commands.do({ arg cmd; Pen.perform(cmd) });
+		Pen.alpha = opacity;
+		Pen.smoothing = smoothing;
+		if(linePattern.notNil) {Pen.lineDash_(linePattern)};
+		x.commands.do({ arg cmd; Pen.perform(cmd) });
+		y.commands.do({ arg cmd; Pen.perform(cmd) });
 		Pen.pop;
 	}
 	font_ { arg f;
@@ -68,7 +68,7 @@ DrawGrid {
 	preview {
 		var insetH = 45, insetV = 35; // left, bottom margins for labels
 		var gridPad = 15;             // right, top margin
-		var txtPad = 2;               // label offset from window's edge
+		var axLabelPad = 3;           // label offset from window's edge
 		var win, winBounds, font, fcolor;
 
 		// refresh and return the view if it already exists
@@ -109,7 +109,7 @@ DrawGrid {
 				Pen.push;
 				Pen.translate(this.bounds.center.x, uv.bounds.bottom);
 				Pen.stringCenteredIn(unitsStr,
-					unitsStr.bounds.center_(0@0).bottom_(txtPad.neg),
+					unitsStr.bounds.center_(0@0).bottom_(axLabelPad.neg),
 					font, fcolor);
 				Pen.pop;
 			};
@@ -121,7 +121,7 @@ DrawGrid {
 				Pen.translate(0, this.bounds.center.y);
 				Pen.rotateDeg(-90);
 				Pen.stringCenteredIn(unitsStr,
-					unitsStr.bounds.center_(0@0).top_(txtPad),
+					unitsStr.bounds.center_(0@0).top_(axLabelPad),
 					font, fcolor);
 				Pen.pop;
 			};
@@ -143,20 +143,32 @@ DrawGrid {
 
 DrawGridX {
 
-	var <grid,<>range,<>bounds;
-	var <font,<fontColor,<gridColor,<labelOffset;
-	var commands,cacheKey;
-	var txtPad = 2; // match with Plot:txtPad
+	var <grid,<>bounds;
+	var <>range;
+	var <font,<fontColor,<gridColor;
+	var <labelOffset, <labelAlign, <labelAnchor, <constrainLabelExtents;
+	var commands, cacheKey, lastLabelSize;
 	var <tickSpacing = 64;
-	var <numTicks = nil; // nil for dynamic with view size
+	var <numTicks = nil; // nil for dynamic numTicks with view size
+	var <>showLabels = true; // tick labels
+	var <labelAppendString = nil;
+	var <drawBoundingRect = true, <drawBaseLine = false, <drawBoundingLines = false;
 
-	*new { arg grid;
-		^super.newCopyArgs(grid.asGrid).init
+
+	*new { arg grid, bounds;
+		^super.newCopyArgs(grid.asGrid, bounds).init
 	}
 
 	init {
 		range = [grid.spec.minval, grid.spec.maxval];
-		labelOffset = "20000".bounds.size.asPoint;
+		labelAnchor = \top;
+		labelOffset = 0 @ 2;
+		labelAlign = \center;
+		constrainLabelExtents = false;
+		font = Font(Font.defaultSansFace, 9);
+		fontColor = Color.grey(0.3);
+		gridColor = Color.grey(0.7);
+		lastLabelSize = "000".bounds(font).asSize;
 	}
 	grid_ { arg g;
 		grid = g.asGrid;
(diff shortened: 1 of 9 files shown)


SCVersion.txt
```diff
@@ -2,7 +2,7 @@
 # Note that you need to "make install" for this information to be copied into all places.
 
 set(SC_VERSION_MAJOR 3)
-set(SC_VERSION_MINOR 13)
+set(SC_VERSION_MINOR 14)
```
mtmccrea, line 5: Note sure what this is about

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -977,6 +975,30 @@ Plotter {
 		};
 	}
 
+	// Convenience methods to dispatch settings shared between all plot grids
```
mtmccrea, line 978: Would like feedback on this implementation:
These are convenience methods to set tick label properties of all of the subplots. 
In order to make the "setters", they also had to be defined as instance variables of `Plotter`. 
Because there are 2 axes with ticks (`x` and `y`), rather than making twice as many methods, I made the axis a variable. As a result the "getters" take an `ax` argument. Getters that take arguments seems odd, but perhaps ok in this scenario?
The getters

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -977,6 +975,30 @@ Plotter {
...
+		plots.do { |plot|
+			plot.drawGrid.perform(axis.asSymbol).perform(method, val)
+		};
+		this.updatePlotBounds;
+	}
+	prGetInSubplotGrid { |axis, method, val|
+		// assume all plot grids have the same properties
+		// (this is an convenience anyway)
+		^plots[0].drawGrid.perform(axis).perform(method);
+	}
```
mtmccrea, line 1000: Would like feedback on this implementation:
These are convenience methods to set tick label properties of all of the subplots. 
In order to make the "setters", they also had to be defined as instance variables of `Plotter`. 
Because there are 2 axes with ticks (`x` and `y`), rather than making twice as many methods, I made the axis a variable. As a result the "getters" take an `ax` argument. Getters that take arguments seems odd, but perhaps ok in this scenario?

SCClassLibrary/Common/Geometry/Rect.sc
```diff
@@ -144,6 +144,7 @@ Rect {
 	}
 
 	asRect { ^this }
+	asSize { ^Size(width, height) }
```
mtmccrea, line 147: Note: `Rect` has a `-size` method, so this is a synonym to that method. But this brings `Rect` into line with a number of other related classes which respond to `-asSize`: `Number`, `Point`, and `Size`. So methods can accept all of these types when converting to a `Size`, and it's more explicit (`size` in SC is also the length of an array).

SCVersion.txt
dyfer: Yeah, sorry about that. When you get a chance, please remove these two commits.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
dyfer: I don't want to be the final adjudicator on these methods, but I'd lean towards either duplicating methods ( `.tickLabelOffsetX`/`.tickLabelOffsetY`) or using array of arguments, OR two separate args ( `.tickLabelOffsets_(pntX, pntY)` or `.tickLabelOffsets_([pntX, pntY])`), with the ability to provide nil as either argument when we want to set only one of them.

telephon: - setters can/should only have one single argument (otherwise you can't write `something.label = "hello"`).
- is there a reason why to call `labelAlign` instead `tickLabelAlign` when it is called on the subplots? 
- setters like `labelOffsetX`, as Marcin suggests
- instead of wrapping everything in private methods, it might be better to have a method that gives you: `plot[0].drawGrid.perform(axis)`

e.g.
```supercollider
representativeGrid { |axis|
^plot[0].drawGrid.perform(axis)
}
```

then you can write:

```supercollider
labelOffsetX_ { |point| this.representativeGrid(\x).labelOffset_(point) }
```

A note:
Writing `asSymbol`, like in  `plot.drawGrid.perform(axis.asSymbol)` over time has proven to be a bad idea in some cases. You make code work that shouldn't really.

mtmccrea: Thanks @dyfer and @telephon for your insightful comments as usual :) I've reworked things based on your feedback and it's getting closer.

>is there a reason why to call `labelAlign` instead `tickLabelAlign` when it is called on the subplots?

(Thanks for asking about this, it made me look more closely at anatomy of `Plotter`'s components again...)
Yes, `Plot` had a concept of axis labels, just called `labelX/Y` but not tick labels. 
`DrawGridX/Y` had a concept of tick labels, which it calls "label" (e.g. `labelOffset`), but no concept of axis labels. 
At the level of `Plotter` we distinguish between the two. So when Plotter acts on it's axis labels, it now sets `axisLabelX/Y` (used to just be `labelX/Y`), and when it acts on the grid labels, it calls them `tickLabelParamX/Y` (e.g. `tickLabelAlignX`). Make sense?

A long term goal is to do 90% of what you might need through `Plotter`, so you rarely have to go into individual plots/grids, where conventions and responsibility become murky. 

>instead of wrapping everything in private methods, it might be better to have a method that gives you: `plot[0].drawGrid.perform(axis)`

Curious what you think of the new approach, which ditches private methods for more useful ones: b17c7b429a4cba7394e313de639ba7c92a7a6d5e

Thanks for being patient with this... there are quite a lot of details to manage, but approaching a milestone that can be suitable for the RC.

SCClassLibrary/Common/Geometry/Rect.sc
telephon: yes, that makes sense.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -65,75 +64,127 @@ Plot {
...
 	}
 
 	bounds_ { |viewRect|
-		var gridRect;
-		var xtkLabels, ytkLabels, ytkLabelSize, xtkLabelSize, xLabelGridOffset, yLabelGridOffset;
-		var xLabelSize, yLabelSize, maxWidthWithLabel, maxHeightWithLabel;
-		var hshift, vshift, hinset, vinset, lmargin, rmargin, totalVSpace;
-		var minGridMargin = 4;
+		var gridRect, boundsSize;
+		var xTkLabelWd, xAxLabelHt, yAxLabelWd, yTkLabelHt, bottomLabelHt;
```
telephon, line 76: Given the size of this method, perhaps you have an idea of how to factor the functionality into either a helper method or a dedicated class that calculates such things?

dyfer (commented):
~~Thanks @mtmccrea ! I have some cosmetic comments/questions.~~ I'm dismissing my comments for now :)

mtmccrea:
Hi @dyfer and @telephon, sorry for the commit noise, this PR isn't in a review-ready state just yet, hence var name changes, and force-pushes after refactoring. I wouldn't normally have submitted a PR that wasn't ready for review, but wanted those waiting on the RC release to see that there was activity here.

I'll notify when it's ready for review, which won't be much longer.

SCClassLibrary/Common/Geometry/Rect.sc
```diff
@@ -49,9 +49,13 @@ Rect {
 	right_ { |r| left = left - (this.right - r) }
 
 	leftTop { ^Point.new(this.left, this.top) }
```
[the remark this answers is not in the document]
mtmccrea: Sure, I see your point, and I was going to ask about this... 
I was refactoring some of the plotter convenience methods into more generally useful ones, and was going to try dispatching some methods with keys I'd already implemented as `topLeft`, etc. But I agree it's not worth muddling the methods in Rect further.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -1,14 +1,15 @@
 Plot {
 	classvar <initGuiSkin;
 
-	var <>plotter, <value;
-	var <bounds, <plotBounds,<>drawGrid;
-
+	var <>plotter; // copyArgs
```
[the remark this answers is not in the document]
mtmccrea: This refers to the fact that `*new` uses `super.newCopyArgs`, which copies it's args, in order, to the first variables declared in the class. I find it useful to isolate these variable declarations from other vars that aren't automatically assigned, and make a note so they aren't later mistakenly reordered. 
I can remove this is it's unnecessary.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -65,75 +64,127 @@ Plot {
...
-			xLabelGridOffset = Point(xtkLabelSize.width, xtkLabelSize.height);
-		};
+		// Turn tick and axis labels on/off depending on available space
+
+		// determine if hidden labels should now be shown (or vice versa)
+		if(prevBoundsSize != boundsSize) {
+
+			// Calculate grid offset to accomodate tick and axis labels.
+			// Currently only supports tick labels below and to the left of the grid.
+			xTkLabelWd = drawGrid.x.labelSize.width;
```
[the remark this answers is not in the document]
mtmccrea: This particular variable isn't used anymore, but the latest commit is closing in on the final form, so I'd be happy to have feedback on `Plot:-prCalcLabelSpace` which is where this layout work is now done.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
mtmccrea: See comment above about this now being factored out to `Plot:-prCalcLabelSpace`.

SCClassLibrary/Common/Geometry/Rect.sc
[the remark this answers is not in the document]
mtmccrea: More specifically, I've added `Rect:-anchorTo`.

https://github.com/mtmccrea/supercollider/blob/30d062651c4ca4f133e03e8992dd115e2fa3f440/SCClassLibrary/Common/Geometry/Rect.sc#L72

```supercollider
anchorTo { arg aPoint, position;
		var p = aPoint;
		^switch(position,
			// center_ returns new Rect, side setters modify this
			\center,      { this.center_(p) },
			\top,         { this.center_(p).top_(p.y) },
			\bottom,      { this.center_(p).bottom_(p.y) },
			\left,        { this.center_(p).left_(p.x) },
			\right,       { this.center_(p).right_(p.x) },
			\topLeft,     { this.copy.top_(p.y).left_(p.x) },
			\topRight,    { this.copy.top_(p.y).right_(p.x) },
			\bottomLeft,  { this.copy.bottom_(p.y).left_(p.x) },
			\bottomRight, { this.copy.bottom_(p.y).right_(p.x) }
		)
	}
```
Would you prefer that these keys match similar (getter) methods in `Rect`? E.g. `\bottomRight` would be changed to `\rightBottom`. This would then mean the keys change in `DrawGridX:-labelAnchor_`:

https://github.com/mtmccrea/supercollider/blob/30d062651c4ca4f133e03e8992dd115e2fa3f440/SCClassLibrary/Common/GUI/Base/Grid.sc#L225

```supercollider
labelAnchor_ { |anchor|
		var anchorSym = anchor.asSymbol;
		var anchorList = #[
			\center, \top, \bottom, \left, \right, \topLeft, \topRight, \bottomLeft, \bottomRight
		];
...
```

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -733,7 +789,13 @@ Plotter {
 				);
 				parent.refresh;
 			};
-		});
+		}) // end keyDownAction_
+		;  // end interactionView setup
```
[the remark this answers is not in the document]
mtmccrea: The chained method calls here extend beyond a single screen, making it a bit confusing where you are in the steps of `makeWindow`, where this lives. So I was planning to leave these comments, but also OK removing them if you'd like.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -918,10 +980,10 @@ Plotter {
 
 	plotColor_ { |colors|
 		plotColor = colors.as(Array);
-		plots.do { |plt, i|
+		plots.do { |plot, i|
```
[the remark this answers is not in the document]
mtmccrea: This was in response to @telephon's feedback that dropping just a couple letters may not be worth the brevity. And looking through this file, other methods like `makePlots`, `updatePlots`, `updatePlotSpecs`, etc. use the `plot` arg.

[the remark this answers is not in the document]
dyfer: Thanks! I understand the comment about dropping letters, but I'd presume this to apply to new argument names. Changing old ones is always tricky - will it break others' code? It might... but for single args maybe that's unlikely enough?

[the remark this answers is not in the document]
mtmccrea: in this case the arg is internal to the loop over `plots`, so users' code won't touch it.

SCClassLibrary/Common/GUI/Base/Grid.sc
```diff
@@ -166,176 +178,418 @@ DrawGridX {
...
 	}
-	gridColor_{ |color|
+	gridColor_ { |color|
 		gridColor = color;
 		this.clearCache;
 	}
-	// labelOffset is effectively a point describing the size of a grid label
-	labelOffset_{ |pt|
-		labelOffset = pt;
+	labelOffset_ { |pnt|
```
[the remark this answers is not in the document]
mtmccrea: This setter isn't currently defined  with an arg name in `main`:

https://github.com/supercollider/supercollider/blob/608a4fdcb0c002a14c6d021062c254c2957205a0/SCClassLibrary/Common/GUI/Base/Grid.sc#L80

It was added for RC1, so I think it's still fair game ;)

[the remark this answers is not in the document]
dyfer: > It was added for RC1, so I think it's still fair game

Oh, in that case for sure! Thanks for pointing this out.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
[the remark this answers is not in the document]
dyfer: Ugh... my bad, sorry!

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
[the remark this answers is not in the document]
dyfer: Okay, we can leave them in if you like them :)

SCClassLibrary/Common/GUI/Base/Grid.sc
```diff
@@ -478,7 +737,6 @@ ExponentialGridLines : AbstractGridLines {
 			numTicks ?? {numTicks = (pixRange / (tickSpacing * nDecades))};
 			tick = first;
 			while ({tick <= (valueMax + step)}) {
-				var drawLabel = true, maxNumTicks;
```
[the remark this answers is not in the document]
mtmccrea: They were causing an non-inlined variable warning (on account of being in `while`). Is there a different reason to leave them as before?

SCClassLibrary/Common/Geometry/Rect.sc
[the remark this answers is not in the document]
telephon: yes, I think that `anchorTo` is a good idea.
 I'd keep the whole convention uniform and match the methods in `Rect`.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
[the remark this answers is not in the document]
telephon: I see, yes. Better remove it because it isn't understandable out of context here.

telephon:
I think it is completely fine to have a discussion like this before a final version.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
telephon: Ah yes, that makes it cleaner. I suppose one can't factor out the three return values into separate methods …

One general thing: if you want the compiler to optimise boolean expressions, don't forget the parentheses.
Instead of `wdAllowsYLabels and: htAllowsYLabels` it should be `wdAllowsYLabels and: { htAllowsYLabels }`.

Then, one very general question (sorry!): why do we need the complicated calculation? Couldn't we just resize the view according to what is needed?

mtmccrea: > why do we need the complicated calculation? Couldn't we just resize the view according to what is needed?

It stems from the fact that the _axis_ labels are independent from _tick_ labels, and further independent between _axes_. The labels on each axis are hidden if the corresponding dimension of the view is small enough that the plot:label ratio goes below a threshold (otherwise the plot becomes useless because it's all labels and tiny data). 

So the calculation gets really tricky depending which labels are active (subject to user settings and plot size)...

> I suppose one can't factor out the three return values into separate methods …

... and these interdependencies make factoring them out quite a challenge (though future work might be done here).

mtmccrea: > One general thing: if you want the compiler to optimise boolean expressions, don't forget the parentheses.
Instead of wdAllowsYLabels and: htAllowsYLabels it should be wdAllowsYLabels and: { htAllowsYLabels }

Thanks, I didn't realize this optimization also applied to evaluation of single boolean values 👍

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -65,75 +66,152 @@ Plot {
...
-			totalVSpace = txtPad + xtkLabelSize.height + txtPad + xLabelSize.height + txtPad;
-			hinset = xtkLabelSize.width/2 + txtPad;
-			vinset = totalVSpace * (2/3);     // inset more on sides with labels
-			vshift = totalVSpace * (1/3).neg;
-			xLabelGridOffset = Point(xtkLabelSize.width, xtkLabelSize.height);
+	// Calculate grid offset to accomodate tick and axis labels,
+	// return modified gridRect and bools allowing X and Y labels.
+	// Currently only supports tick labels that don't extend past
+	// the side opposite their respective axis.
+	// TODO: this method could be refactored to use the showTickLabelsX/Y
```
telephon, line 93: isn't that done? Line 111 does the skip.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -241,27 +319,31 @@ Plot {
...
+				tklabelPad = if (tklabelPad > 0) { tklabelPad + labelMargin } { 0 };
 				Pen.push;
-				Pen.translate(
-					plotBounds.left - drawGrid.y.labelOffset.x + txtPad + (lbounds.height/2),
-					plotBounds.center.y);
+				Pen.translate(plotBounds.left - tklabelPad - (lbounds.height/2), plotBounds.center.y);
 				Pen.rotateDeg(-90);
 				Pen.stringCenteredIn(labelY, lbounds.center_(0@0), font, fontColor);
 				Pen.pop;
+				// "Drawing Y Label".postln;
```
telephon, line 346: something to clean out before merge

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -553,73 +634,46 @@ Plot {
...
 	var <value, <data, <domain;
-	var <plots, <specs, <domainSpecs, plotColor, labelX, labelY;
+	var <plots, <specs, <domainSpecs, plotColor;
 	var <cursorPos, plotMode, <>editMode = false, <>normalized = false;
 	var <>resolution = 1, <>findSpecs = true, <superpose = false;
 	var modes, <interactionView;
 	var <editPlotIndex, <editPos;
-
 	var <>drawFunc, <>editFunc;
+	var axisLabelX, axisLabelY;
```
telephon, line 651: maybe add getters?

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -65,75 +66,152 @@ Plot {
...
+			totalHorizPad   = leftPad + rightPad;
+			wdAllowsYLabels = viewRect.width >= (totalHorizPad * (hideLabelsWidthRatio+1));
+
+			sizeAllowsYLabels = wdAllowsYLabels and: { htAllowsYLabels };
+
+			if(sizeAllowsYLabels) {
+				gridRect.width = viewRect.width - (leftPad + rightPad);
+				gridRect.left  = leftPad;
+			}
+			// { "width hides Y axis labels".postln }
```
telephon, line 208: something to clean out before merge

telephon: … like the other commented out posts.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -553,73 +634,46 @@ Plot {
...
 		} {
 			bounds = bounds ?? { parent.bounds.moveTo(0, 0) };
 			interactionView = UserView.new(parent, bounds);
 			interactionView.drawFunc = { this.draw };
 		};
-		modes = [\points, \levels, \linear, \plines, \steps, \bars].iter.loop;
-		this.plotMode = \linear;
-		this.plotColor = GUI.skins.plot.plotColor;
 
+		// set up interactionView
```
telephon, line 676: this comment is redundant, I think?

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
mtmccrea: Line 111 is a shortcut in case _both_ `showTickTickLabels` vars are false. This comment points to refactoring further when only one is disabled, then there can likely be some calculation skipped.

mtmccrea:
OK, I  would consider this "review-ready" at this point 😆.

I've updated the topic post with some details and examples. As mentioned there, I won't get to complete docs before RC2, but I think the interface is stable at this point (pending any feedback), so could go into the RC.

Thanks!

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -50,90 +55,157 @@ Plot {
...
+		tkBHang = max(xTkBHang, yTkBHang);
+
+		// if there's label overhang, give it its margin
+		tkBMargin = if(tkBHang  > 0) { labelMargin } { 0 };
+		tkTMargin = if(yTkTHang > 0) { labelMargin } { 0 };
+
+		// only the largest margin of all elements is needed
+		bottomMargin = maxItem([xAxMargin, borderMargin, tkBMargin]);
+
+		// add up all elements extending below the grid
```
mtmccrea, line 143: Note the nonconventional alignment of `=` in this method.
It breaks from the style guidelines, but I find it easier to follow when there are so many variables with subtly different names and varying length.
If this is a problem I can remove the extra white space.

dyfer: My preference is to not include any manual non-standard formatting. I acknowledge that it looks better, but I believe even more strongly that (almost?) everything should be formatted automatically.
OTOH, the [automatic liner](https://github.com/supercollider/supercollider/actions/runs/3832847520/jobs/6523627143#step:3:59) for the class library files (which is set to not cause CI to fail since some existing .sc files don't conform) does not seem to catch this as an issue. So... maybe this is acceptable?

sternsc:
Thanks everyone for the efforts you are putting in on this! I’m just a bit worried that you may be overdoing it. As far as I am concerned, the .labelOffset functionality in SC 3.12 was good enough, it just wasn’t documented - and then it broke in 3.13 RC 1.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
mtmccrea: BTW, it makes sense to me for this kind of "helper" method to be private, but if you have thoughts on that, just let me know. Also whether these private methods should live close to the method they're helping, or rather gathered at the bottom of the class definition.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
mtmccrea: This is interesting... I didn't know we were linting sc files. Ah ok i see it's just [checking](https://github.com/supercollider/supercollider/blob/develop/.editorconfig) tab character choice and length. In this case all of these `=`s align somewhere mid-line so the head of the line still conforms.

I'll defer to whoever feels strongest about this ;)

mtmccrea:
@sternsc 
This PR can be considered as an extension of the previous one which added much more functionality than a modification of `labelOffset`. Because you noticed there was this breaking change, I wanted to make sure the old functionality was still replicable. But again, that's only a small part of the additions.

So just as a heads up, your code will still likely need to be updated. Here's an example that recreates some of the things you missed about the previous version (in particular labelOffset, and strings appended to tick labels):

```supercollider
// ~ Old Plotter replicated functionality: x labels at top, with strings appended
(
var gridHeight, freq = 22;
p = 101.collect({ |i| sin((i/1000)*2*pi* freq)}).plot;
gridHeight = p.plots[0].plotBounds.height;

p.setGridProperties(\x,
	\labelAnchor, \topLeft,
	\labelAlign, \left,
	\labelOffset, 8 @ gridHeight.neg,
	\labelAppendString, " xxx",
	\constrainLabelExtents, true,
).setGridProperties(\y,
	\labelAnchor, \bottomLeft,
	\labelAlign, \left,
	\labelOffset, 3 @ -3
);
)
```
If you plan for your views to dynamically resize, you can make sure extreme tick label offsets stay in place this way:
```supercollider
(
// Because use the view height to set the x tick label offset, update label offset on resize.
p.interactionView.onResize_({ |v|
	var gridHeight = p.plots[0].plotBounds.height;
	p.setGridProperties(\x, \labelOffset, 8 @ (gridHeight.neg+10))
});
)
```

mtmccrea:
I've cleaned up the commit history.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
mtmccrea: OK I've done the style cleanup you requested 👍 https://github.com/supercollider/supercollider/pull/5942/commits/58e7e9594307a87381fe7592022ee71902ae4c67

telephon:
ok, let us know when it is ready for a final review. Thanks a lot!

mtmccrea:
I've updated the docs now so this is ready to go!

SCClassLibrary/Common/GUI/Base/Grid.sc
```diff
@@ -98,33 +97,7 @@ DrawGrid {
...
-			if(unitsStr.size > 0) {
-				Pen.push;
-				Pen.translate(0, this.bounds.center.y);
-				Pen.rotateDeg(-90);
-				Pen.stringCenteredIn(unitsStr,
-					unitsStr.bounds.center_(0@0).top_(txtPad),
-					font, fcolor);
-				Pen.pop;
-			};
+			this.draw; // draw the drid
```
telephon, line 100: drid -> grid

telephon (approved):
I've only found one typo. 

All the rest is good, I think. You made a TODO-note in the code, it would be good to resolve further issues, but I think these are minor things.  Thanks a lot!

SCClassLibrary/Common/GUI/Base/Grid.sc
telephon: (you can leave out the comment, it is redundant)

mtmccrea: Thanks for catching this. Removed.

mtmccrea:
Thanks @telephon and @dyfer for your helpful reviews!

telephon (approved):

dyfer:
Thank you @mtmccrea for your work on this and @telephon for the review.
@sternsc could you confirm that the changes here fix the issues you've encountered?

sternsc:
Thanks to all!

Unfortunately I won’t have time soon to try the new code – I don’t have a build environment for all of SC set up and ready to go. Would it work to get only the relevant .sc files (Plotter, GridLines etc) from the latest version, and replace the ones in 3.12.2, just to try? If so, which files exactly?

dyfer:
> Unfortunately I won’t have time soon to try the new code – I don’t have a build environment for all of SC set up and ready to go. 

For macOS and Windows you can download the builds from GitHub actions (see "details" next to the CI checks, or the "Checks" tab on top, then click "Summary" and find artifacts on the bottom right, after scrolling down). 

> Would it work to get only the relevant .sc files (Plotter, GridLines etc) from the latest version, and replace the ones in 3.12.2, just to try? If so, which files exactly?

If might. I'd try all the .sc files changed in this PR (see "Files changed" on top_ and additionally the ones which were changed in #5161 (but of course taken from the the branch in _this_ PR).

telephon:
the first thing you can try to just copy the whole class library over. If there weren't any changes to kernel classes and primitives, it might just work. If it doesn't, you could try copy the GUI classes.

dyfer:
> the first thing you can try to just copy the whole class library over. 

I didn't suggest this since AFAIR there were some changes to the interpreter and when I tried using class library from a recent `develop` on the SC 3.12 build it didn't work. But yes, this would typically be the first thing to try.

sternsc:
OK, thanks. I won’t have time to do this soon, I have confidence in the changes made and in the process, and I don’t want to hold up the next RC release, so don’t wait for me.

mtmccrea:
@sternsc here is code that reproduces the grids you showed in your [previous post](https://github.com/supercollider/supercollider/pull/5161#issuecomment-1321457870).

<details>
	<summary>Example Code</summary>

	```
	( // Example 1
		var win = Window("DrawGrid test",
			Size(350, 300).asRect.center_(Window.screenBounds.center)
		).front;
		var uview = UserView(win, win.bounds.size.asRect);
		var xspec = [30, 95].asSpec;  // MIDI
		var yspec = [45, 120].asSpec; // dB
		var inset = 2; // border around the grid
		
		d = DrawGrid(uview.bounds.insetBy(inset),
			xspec.grid, yspec.grid
		);
		
		uview.drawFunc_({ |uv| d.draw; })
		.onResize_({ |uv|
			d.bounds = uv.bounds
			.insetBy(inset)
			.moveTo(inset, inset);
		})
		.resize_(5)
		.background_(Color.white)
		;
		
		d.x
		.labelAnchor_(\bottomLeft)
		.labelAlign_(\left)
		.labelOffset_(3 @ -3)
		.labelAppendString_(" MIDI")
		.gridColor_(Color.green)
		.fontColor_(Color.green)
		.font_(Font(Font.defaultSansFace, 11))
		;
		
		d.y
		.labelAnchor_(\topLeft)
		.labelAlign_(\left)
		.labelOffset_(3 @ 3)
		.labelAppendString_(" dB")
		.gridColor_(Color.green)
		.fontColor_(Color.green)
		.font_(Font(Font.defaultSansFace, 11))
		;
	)
	
	( // Example 2
		var win = Window("DrawGrid test",
			Size(350, 200).asRect.center_(Window.screenBounds.center)
		).front;
		var uview = UserView(win, win.bounds.size.asRect);
		var xspec = [-4.8, 0].asSpec; // s
		var yspec = [0, 28].asSpec;
		var inset = 2; // border aroudn the grid
		var xtickoffset = 10;
		d = DrawGrid(uview.bounds.insetBy(inset),
			xspec.grid, yspec.grid
		);
		
		uview.drawFunc_({ |uv| d.draw; })
		.onResize_({ |uv|
			d.bounds = uv.bounds
			.insetBy(inset)
			.moveTo(inset, inset);
			d.x.labelOffset_(0@d.bounds.height.neg + xtickoffset)
		})
		.resize_(5)
		.background_(Color.black)
		;
		
		d.x
		.labelAnchor_(\topLeft)
		.labelAlign_(\left)
		.labelOffset_(0@d.bounds.height.neg + xtickoffset)
		.labelAppendString_(" s")
		.gridColor_(Color.gray)
		.fontColor_(Color.gray)
		.font_(Font(Font.defaultSansFace, 11))
		;
		
		d.y
		.labelAnchor_(\topLeft)
		.labelAlign_(\left)
		.labelOffset_(3 @ 3)
		.gridColor_(Color.gray)
		.fontColor_(Color.gray)
		.constrainLabelExtents_(false)
		.font_(Font(Font.defaultSansFace, 11))
		.numTicks_(5)
		;
	)
	```
</details>

Which produces:
Screen Shot 2023-01-13 at 16 39 31

(Including behavior for update on resize). 

So I think the functionality you've described is covered here.

sternsc:
Thanks for your kind example @mtmccrea , it looks great! Looking forward to it!

dyfer:
Thanks so much @mtmccrea !
Sorry for looking into the details late in the conversation here.
Do I read correctly that in the process we lost automatic display of `ControlSpec` units? If so, I'd consider that as a regression from SC 3.12, where units were always displayed.
IMO they should still be displayed by default, but the question is whether that default should be next to the value, or along the axis. 
I think it would be best to have a method to switch between these two options easily, e.g.;

`DrawGridX/Y -unitDisplay(arg)` where arg is `\tick` (appended to tick labels, 3.12 behavior), `\axis` (on axis label, "new" behavior), `nil` (not displayed). Either `\tick` or \axis` could be the default, depending what we all feel like.

What do you think?

Sorry for dragging this out!

mtmccrea:
Yes spec unit auto label functionality was [removed](https://github.com/supercollider/supercollider/pull/5942/commits/1d7928977ff10cc48eb290aa87c39c326f0cfbf0) (IIRC, in response to the last PR).

Some background:
<details>
  <summary>`3.12.2` default behavior:</summary>
  
Just plotting a stereo buffer:
Screen Shot 2023-01-14 at 12 40 21
```
(
p = (4 ** (-5..0)).plot;
p.specs = \delay;
p.domainSpecs = [0, 10, \lin, 0, 0, " xxx"].asSpec; p.refresh;
)
```
Screen Shot 2023-01-14 at 12 21 48
</details>


<details>
  <summary>`3.13-rc` default behavior:</summary>
 
Screen Shot 2023-01-14 at 12 24 03

Screen Shot 2023-01-14 at 12 23 43
</details>


<details>
  <summary>`3.13-rc` to achieve the old behavior (with new space padding):</summary>

```
// ... after plotting the buffer
p.axisLabelX = nil;
p.setGridProperties(\x, \labelAppendString, " " ++ p.domainSpecs.first.units)
```
Screen Shot 2023-01-14 at 12 28 27

```
(
p = (4 ** (-5..0)).plot;
p.specs = \delay;
p.domainSpecs = [0, 10, \lin, 0, 0, " xxx"].asSpec; p.refresh;

p.setGridProperties(\y, \labelAppendString, p.specs.first.units);
p.setGridProperties(\x, \labelAppendString, p.domainSpecs.first.units);
)
```
Screen Shot 2023-01-14 at 12 26 19
</details>

Notice `Buffer:-plot` still shows units by default ([though not inferred from the spec](https://github.com/mtmccrea/supercollider/blob/0de631b5983b08bfc837bacd6bf17942e6852b38/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc#L1307)), to align  with previous behavior, but shown in the axis label. 
`Function:-plot` does as well, because in both cases the units can be inferred confidently, unlike, e.g., `Array:-plot`.
I opted to put the units in the axis label because I thought having units on every tick label was redundant and crowded, and axis labels can replace that functionality. The extra vertical space taken is a consideration, so it will [omit the unit label](https://github.com/mtmccrea/supercollider/blob/0de631b5983b08bfc837bacd6bf17942e6852b38/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc#L1318) when  the channel count is > 4.

Moving forward:
So you can still achieve the same behavior in 3.13. Now the questions are 
- whether spec units should be shown by _default_  (in all cases where a spec has units, rather than just `Buffer` and `Function` plots)
- if so, whether  the units should display with tick labels or as the axis label
If I'm not mistaken, @telephon authored the original approach, so may have thoughts on this.

dyfer:
> Moving forward: So you can still achieve the same behavior in 3.13. Now the questions are

Yes, but it requires extra code.

> * whether spec units should be shown by _default_  (in all cases where a spec has units, rather than just `Buffer` and `Function` plots)

I do think it should.

> * if so, whether  the units should display with tick labels or as the axis label

With this one I could go either way. I like the visual aspect of the "new" way, but it does take a bit more space around the plot.

telephon:
I'm fine either way, there is much improvement.

mtmccrea:
OK thanks, I've made the units show by default (in the axis label—easy to change this default if we hear feedback from users in the RC).

I've also added these methods to the docs.

telephon:
good! let us know when it is ready for merge.

mtmccrea:
It's ready on my end :)

dyfer:
Thank you both!

sternsc:
Super, thanks for 3.13.0-rc2. I've tested the new Grid-related stuff and found the following: 
1. in ExponentialGridLines.getParams there is a small bug (Grid.sc line 751 and following: 
`						) {
							drawLabel = false // drop labels for tightly spaced upper area of the decade
						};
`
But drawLabel must be made true again when increasing to the next decade: 
`						) {
							drawLabel = false // drop labels for tightly spaced upper area of the decade
						} { drawLabel = true } ;  // sternsc: bug fix: restore labels for the next decade
`
2. The warp property of a ControlSpec is set using a symbol (\lin or \exp), but the getter function now returns instead an object instance of a subclass of AbstractGridLines. This seems asymmetric and makes it more awkward to inquire for the type of warp. I suggests a getter method called .warper that returns the object, and keeping .warp to return the symbol. Maybe. 
3. It would be great if .formatLabel could have a drawUnits argument in addition to the drawLabel argument. The reason is that it helps to plot the unit only on whole decades, as shown in this figure.

Untitled

I've been able to work my way around to this result by overriding .getParams and twiddling with calls to .appendLabel, as shown by this code excerpt from the end of ExponentialGridLines.getParams: 
`

			// p['labels'] = lines.collect({ arg arr, inc;
			// 	var val, drawLabel, thisLabel;
			// 	#val, drawLabel = arr;
			// 	[val, this.formatLabel(val, nfrac ? nfracarr[inc] ? 1), nil, nil, drawLabel.not];
			// });

			//// sternsc's exponential version: draw the label units only at whole decades
			//// It could be neater if .formatLabel had also a drawUnits argument
			p['labels'] = lines.collect({ arg arr, inc;
				var val, drawLabel, thisLabel;
				var saveUnits, retVal;
				#val, drawLabel = arr;
				saveUnits = this.appendLabel;
				if (val.abs.log10.frac > 0.01, { this.appendLabel_(nil) });
				retVal = [val, this.formatLabel(val, nfrac ? nfracarr[inc] ? 1), nil, nil, drawLabel.not];
				this.appendLabel_(saveUnits);
				retVal
			});
		};
		^p
`
4. The usage of the numTicks argument with exponential axes needs to be clarified. So far as I can tell, it means approximately the number of ticks per decade: 1 or 3 or 10; where three gives 1,2,5 (fine!) and anything more than 4 gives up to 10. The default behavior is a bit clunky, so I found myself changing numTicks in the view's .onResize method. This probably can't be automated because it depends on font sizes and units strings as well. 

Hope this is of some help. And I can live with overriding - you can't cater to everyone's preferences...

PR #5953, opened 2023-01-22
GHA: create macOS universal binary with homebrew libraries

dyfer:
## Purpose and Motivation

This PR replaces a partial script for manual creation of universal libraries for homebrew packages with a [3rd-party script which I wrote](https://gist.github.com/dyfer/6c83905d4593750105897e51e87ec345), that achieves this more reliably.

A little history how we got here:
GHA does not offer arm64 macOS runners, so in order to provide an arm64 build (or a universal one), we need to cross-compile. This requires all dependencies to be available as universal binaries, so at first we used vcpkg to create arm64 binaries of libsndfile and fftw, and created a script that manually created universal binaries of readline, qt5 and its dependencies. This script was fairly specific, didn't handle some packages well, was able to use caching in a limited way etc. Also, libsndfile in homebrew was not supporting mpeg at that time, so libsndfile was installed from vcpkg. Unfortunately, it wasn't possible to create a universal binary of libsndfile (because of the build issue with LAME library, I believe), only a single architecture arm64.

Libsndfile in homebrew now offers mpeg support, so we're coming full circle, again using homebrew dependencies in the main supercollider build, but this time as a universal binary.

I opted to download the script from my gist, as opposed to including it in this repo, as I believe it doesn't really belong here - it's a medium-term solution and rather circumstantial to SC itself. Eventually, I think we can stop using this method and use 1) official qt installer to get the universal binary of Qt6 (once we move to Qt6) and 2) use vcpkg to build universal binary of libsndfile (once the build issue is fixed). For now, this solution offers a universal binary and **appears to support the same range of OS versions** as the regular build (I tried running the universal build on macOS 10.14, which is currently the oldest supported OS for the regular build, and it seems to work fine).

One outstanding question is whether we pull it into `develop` for the next release, or do we want to release 3.13.1 with this, or maybe even 3.13.0-rcX so that it lands in 3.13.0?

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation (N/A)

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -258,10 +258,9 @@ jobs:
             use-syslibs: false
             shared-libscsynth: false
             system-portaudio: false
-            build-libsndfile: true
-            build-readline: true
-            build-fftw: true
-            vcpkg-triplet: arm64-osx-release-supercollider # required for build-libsndfile
+            build-libsndfile: false
+            build-readline: false
+            build-fftw: false
             artifact-suffix: 'macOS-arm64' # set if needed - will trigger artifact upload
             verify-app: true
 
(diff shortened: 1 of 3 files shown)


dyfer:
After conversation with @joshpar we will include this in 3.13.0-rc2. This PR now includes update to the changelog regarding macos universal binary build.

joshpar (approved):

Issue #4127, opened 2018-10-17
Initial state of many UGens is not sample accurate.


mtmccrea:
Environment
-----------

* Your SuperCollider version: 3.10 beta-2
* Your operating system and version: OS X 10.12.6

Steps to reproduce
-----------------------------

Once illustrative example using `Osc` with a sine table:
```supercollider
(
var p1, p2, buf1, buf2, numSampsToPlot = 10;
fork ({
	s = Server.local;
	buf1 = Buffer.alloc(s, 8192, 1);
	buf2 = Buffer.alloc(s, 8192, 1); // just for plotting

	0.3.wait;
	buf1.sine1(1.0, true, true, true);

	// plot the source wavetable, not in wavetable format
	buf2.sine1(1.0, true, false, true);
	buf2.loadToFloatArray(action: {|arr|
		var win = arr[0..numSampsToPlot-1].round(1e-7);
		postf("Wavetable source\n>> %\n", win);
		p1 = win.plot("Wavetable source, starts at t=0").plotMode_(\points)
	});

	p2 = {
		Osc.ar(buf1.bufnum, (8192/s.sampleRate).reciprocal, 0)
	}.plot(
		numSampsToPlot+1/s.sampleRate,
		bounds: p1.parent.bounds.left_(p1.parent.bounds.right)
	).plotMode_(\points);
	
	// p.name_("Doesn't work"); // Plotter:name_ bug
	p2.parent.name_("Osc UGen, starts at t=1");
	
	0.2.wait;
	postf("Osc UGen\n>> %\n", p2.data[0][0..numSampsToPlot-1].round(1e-7));
	x = p2
}, AppClock)
)

/*
Wavetable source
>> [ 0.0, 0.000767, 0.001534, 0.002301, 0.003068, 0.0038349, 0.0046019, 0.0053689, 0.0061359, 0.0069029 ]
Osc UGen
>> [ 0.000767, 0.001534, 0.002301, 0.003068, 0.0038349, 0.0046019, 0.0053689, 0.0061359, 0.0069029, 0.0076698 ]
*/
```

Or simply, with `SinOsc`:
```supercollider
p = {SinOsc.kr(10, 0)}.plot(duration:0.02).plotMode_(\points);
p.data;

// >> [ [ 0.091058231890202, 0.18135985732079, 0.27015459537506, ...
````

Expected Behavior
-----------------
The first sample in both of the above examples should be `0`.

Current Behavior
----------------
[...]

PR #5965, opened 2023-02-09
Poll: fix initialization to fire on first-sample trigger

mtmccrea:
## Purpose and Motivation

`Poll` misses a trigger on the first sample... but not anymore!

This bug was revealed only with the updated behavior of `Impulse` (as of 3.13). There was no issue filed, but addresses a [discussion](https://github.com/supercollider/supercollider/issues/4127#issuecomment-1424294256) that uncovered the bug.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [N/A] Updated documentation

### DIFF
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -765,7 +765,6 @@ void Poll_Ctor(Poll* unit) {
         SETCALC(Poll_next_kk);
     }
 
-    unit->m_trig = IN0(0);
     const int idSize = (int)IN0(3); // number of chars in the id string
     unit->m_id_string = (char*)RTAlloc(unit->mWorld, (idSize + 1) * sizeof(char));
     ClearUnitIfMemFailed(unit->m_id_string);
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks! This seems to work as expected:
```supercollider
(
{
	var trig = Impulse.kr(1);
	Sweep.kr(1).poll(trig);
	FreeSelf.kr(PulseCount.kr(trig) >= 3);
	Silent.ar(1)
}.play;
)
```
results in
```
-> Synth('temp__1' : 1001)
UGen(Sweep): 0.00133333
UGen(Sweep): 1.00267
UGen(Sweep): 2.00267
```
While this looks fine to me, it does _not_ produce the double posting at the beginning of the synth that you showed [here](https://github.com/supercollider/supercollider/issues/4127#issuecomment-1424294256). I just wanted to double check whether the behavior in this PR is exactly what you'd expect.

mtmccrea:
> it does not produce the double posting

Yes, that's the correct behavior. The double posting from before was just the quick "patch" for the previous thread, to make sure the issue wasn't with Impulse.

Thanks for your attention to detail :)

dyfer (approved):
Thanks!

PR #5968, opened 2023-02-12
class library: keep SystemSynthDefs in dictionary

### LINKED ISSUE #5967: can't send system synth defs on demand
## Motivation
The SytemSynthDefs are sent to the server at `StartUp`, which is not called when using an NRT server.
This should be slightly refactored so that we can send them on demand.

add a new method of SystemSynthDefs

## Plan for Implementation

I'll do this now.

telephon:
This will allow to send them to an NRT server

## Purpose and Motivation

Fixes #5967 

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Audio/SystemSynthDefs.sc
+++ b/SCClassLibrary/Common/Audio/SystemSynthDefs.sc
@@ -5,6 +5,7 @@ SystemSynthDefs {
 	classvar <>tempNamePrefix = "temp__";
 	classvar tempDefCount = 0;
 	classvar <>maxTempDefNames = 512;
+	classvar <>synthDefs;
 
 	*generateTempName {
 		var name = tempNamePrefix ++ tempDefCount;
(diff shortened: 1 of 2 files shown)


adcxyz (approved):

mtmccrea:
@telephon, was this ready to be merged? I see you noted a possible alternative approach using `SynthDescLib` in the linked issue, and this wasn't checked as "ready for review".

telephon:
It is ok – it is clear and solves the problem (I can't find a linked issue, though and don't quite remember). 

Thanks for asking! Of course, if you see a problem, let me know...

mtmccrea:
Ok sounds good 👍 

The only minor comment I might have suggested was to consider making the `addSystemSynthDefs` "private" if the user won't ever explicitly call it. But nbd...

(btw the linked issue I referred to is the one in your description above)

telephon:
So far, we have to call `addSystemSynthDefs` in score explicitly, and I think it is good to keep it that way.

telephon:
(something that needs documentation though!)

PR #5969, opened 2023-02-12
removes inline warnings

telephon:
## Purpose and Motivation

this removes inline warnings

Note: as a developer, please set:
```supercollider
LanguageConfig.postInlineWarnings = true;
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested

### DIFF
--- a/SCClassLibrary/Common/Streams/Stream.sc
+++ b/SCClassLibrary/Common/Streams/Stream.sc
@@ -538,14 +538,13 @@ EventStreamPlayer : PauseStream {
 	prNext { arg inTime;
 		var nextTime;
 		var outEvent = stream.next(event.copy);
+		var roundedBeat, deltaFromRounded;
 		if (outEvent.isNil) {
 			streamHasEnded = stream.notNil;
 			cleanup.clear;
 			this.removedFromScheduler;
 			^nil
-		}{
-			var roundedBeat;
-			var deltaFromRounded;
+		} {
 			nextTime = outEvent.playAndDelta(cleanup, muteCount > 0);
 			if (nextTime.isNil) { this.removedFromScheduler; ^nil };
 			nextBeat = inTime + nextTime;	// inval is current logical beat


dyfer:
We are about to release 3.13.0-rc3, which I hope will be the final RC before the final release.
I think this could go into rc3, but could you rebase it on the `3.13` branch so that it can be pulled in directly?

telephon:
I couldn't find the branch `3.13.0-rc3` so I changed the target to `3.13` above.

dyfer:
The branch `3.13` is correct, we don't make separate branches for RC.

However, my request was to [rebase](https://github.com/supercollider/supercollider/wiki/Creating-pull-requests#rebasing-to-change-the-base-branch-of-a-pr-branch) this PR on `3.13`, (changing the PR target is trivial and I could've done this myself). I know this is a bit more cumbersome, would you be able to do that? If not, I think it would be easier to just keep this on `develop`, I don't see much harm if this doesn't get into 3.13. Would that be okay?

dyfer:
As we're about to release rc3, I think it's fine if this stays on `develop`. But if you can fix the rebase in the next hour or so, we can pull this in.

dyfer:
I think we can leave this to be merged into `develop`.

telephon:
OK, I'm a little confused how we best proceed. I wasn't aware of the release schedule (I searched on scsynth, but there was nothing, maybe I searched for the wrong thing).

This should definitely go into a final release, if we want to keep inline warnings as a feature.

server/plugins/TriggerUGens.cpp
```diff
@@ -775,8 +774,7 @@ void Poll_Ctor(Poll* unit) {
 
     unit->m_id_string[idSize] = '\0';
     unit->m_mayprint = unit->mWorld->mVerbosity >= -1;
-
```
telephon, line 778: I have no clue where this suddenly comes from. It is not visible in my own branch …

telephon:
It's just a tiny change, that can be made directly in any branch. So just let me know what is needed and we trash this pull request.

dyfer:
@telephon I understand this is a small change, but I couldn't pull it in into 3.13 directly in the current state.

Here's the [announcement for 3.13.0-rc2](https://scsynth.org/t/3-13-0-rc2-announcement/7115) that says that we'll do another release in "about a week". I'm sorry this wasn't more clear. 

What is the rebase procedure that you're following? I don't know why this ends up in an incorrect state.

3.13.0-rc3 has been tagged (I'm waiting with the announcement until we update the official website). If you really feel we need to pull this into 3.13, then I'll make another rc with this after we pull it in.

telephon:
@dyfer thank you, and sorry that this has escaped me. 

Also I wonder why nobody else noticed this pretty obvious glitch …


> What is the rebase procedure that you're following? I don't know why this ends up in an incorrect state.
I am not sure what happened and why the thing turned stale.

In any case, my proposal would be to deviate from protocol and just directly fix it before release without adding another candidate officially. But I'd go with whatever you prefer!

server/plugins/TriggerUGens.cpp
dyfer: This seems to come from commit 3a7af1564eed65ec8ffbc2c3a271710ce73e2ecd

dyfer:
I understand this is a simple change, but does this actually affect users? I'm not even sure how to reproduce the problem, I don't see warnings when I play a pattern.

Also, can you please explain how you did the rebase? Maybe our instructions on the wiki need to be updated to avoid issues with this moving forward.

telephon:
Sure! It will cause an inline warning each time someone recompiles their class library. I always search for the source of those warnings and try to fix them.

Of course, many people nowadays have their inline warnings switched off, because this is the default setting. Anyone interested in contributing to the class library or publishing code should keep it on, however.

This is why I wrote that there is an implicit decision here. If we think that non-inlined variables are part of the accepted standard now, then we should explicitly communicate this. Otherwise, we probably should set the default to "on".

But we shouldn't publish a class library that posts warnings on recompile.

telephon:
> Also, can you please explain how you did the rebase? Maybe our instructions on the wiki need to be updated to avoid issues with this moving forward.

Good! I used GithubDesktop and rebased on the `3.13` branch. Then I force pushed, went to the website and changed the target by editing the pull request.

dyfer:
> I used GithubDesktop and rebased on the `3.13` branch.

Ah, I see.
I tried this, and this doesn't work well in the current state of branches. What you'd need to do is to do an _interactive_ rebase on top of `3.13` branch (instructions on the wiki) and then just leave the single commit that does what you want ("remove inline warnings"). After that we should see a single commit here (in the GH web interface) with the target set as `3.13`.

dyfer:
> Sure! It will cause an inline warning each time someone recompiles their class library. I always search for the source of those warnings and try to fix them.

Sorry about the confusion, I didn't save the language config when I tried this. I see it now (and indeed this is a regression from previous behavior).

It would be great to add that to the testsuite somehow, but I'm not sure how.

telephon:
> In the future it would be great to add that to the testsuite somehow, but I'm not sure how.

One waas would be to read the last line in the post window. But, it may have been a mistake to make the post window a completely separate thing. Like this, we have no access to its output (`Document.allDocuments.collect(_.name)`).

Alternatively, change the default in the preferences, so it is on by default? This was the case in older versions.

dyfer:
I think we could change the default (on `develop`).
In the meantime, could you attempt the rebase as I indicated [above](https://github.com/supercollider/supercollider/pull/5969#issuecomment-1434302558)?

telephon:
> In the meantime, could you attempt the rebase as I indicated [above](https://github.com/supercollider/supercollider/pull/5969#issuecomment-1434302558)?

ok, done.

telephon:
> I think we could change the default (on `develop`).

yes, that would be a low-impact change.

dyfer (approved):
Thanks!
Let's merge this in for the upcoming release.

dyfer:
FYI the CI failure is unrelated to this fix. IIUC, it is because of an issue upstream with GHA runner images and pkg-config not being preinstalled, or something like that (see https://github.com/actions/runner-images/pull/7125). I'll wait with the final release until that is fixed.

PR #5970, opened 2023-02-12
test: proxy as source should be initialised by receiver

telephon:
## Purpose and Motivation

Just a test for an ambiguous situation.

## To-do list

- [x] Code is tested
- [x] All tests are passing


dyfer (approved):
Thanks!

PR #5974, opened 2023-02-13
scsynth: /n_fill fix behaviour

### LINKED ISSUE #5973: scsynth: /n_fill command needs an extra parameter
## Environment

* SuperCollider version: "develop" branch, commit b751c733a67c6a51233f33a2bcbd02b9a3843989
* Operating system: macOS 12.6
* Other details: In the project where I've encountered the error, I'm not using scide / sclang, but instead controlling scsynth directly via osc with [ofxSuperCollider](https://github.com/PlaymodesStudio/ofxSuperCollider).

## Steps to reproduce

I've reproduced the issue within sclang to make it easier to track.
Please go step by step and check manually the output (sorry in advance if there is a more helpful way of debugging this kind of behaviour)

```supercollider
//Boot server
s.boot

//Create Synth definition with an array argument (levels)
(
SynthDef(\test, {
	var sig = SinOsc.ar([72, 75, 79, 82].midicps, 0, \levels.kr(0!4));
	Out.ar(0, Splay.ar(sig));
}).add;
)

//Create synth
x = Synth(\test);

//Turn on all oscillators
x.set(\levels, [0.1, 0.1, 0.1, 0.1])

//Turn off all oscillators
x.set(\levels, [0, 0, 0, 0])

//Turn on directly via osc
s.sendMsg("/n_setn", x.nodeID, "levels", 4, 0.1, 0.1, 0.1, 0.1)

//Turn off directly via osc
s.sendMsg("/n_setn", x.nodeID, "levels", 4, 0, 0, 0, 0)

//Turn on directly via osc via n_fill (No sound, shoud sound)
s.sendMsg("/n_fill", x.nodeID, "levels", 4, 0.1)

//Fill command sends the same osc message (No sound)
x.fill("levels", 4, 0.1)

//Add an additional parameter to n_fill (we get sound)
s.sendMsg("/n_fill", x.nodeID, "levels", 4, 0.1, 0.1)

//Changes in first parameter don't affect the value processed.
s.sendMsg("/n_fill", x.nodeID, "levels", 4, 0, 0.1)

//Setting all levels to 0, without the need of additional parameters work.
s.sendMsg("/n_fill", x.nodeID, "levels", 4, 0)
```

When looking the n_fill command at the Server Command Reference https://github.com/supercollider/supercollider/blob/b751c733a67c6a51233f33a2bcbd02b9a3843989/HelpSource/Reference/Server-Command-Reference.schelp#L234-L245

It looks to me that giving 4 parameters (node, control name, number of values and value) to n_fill command should modify correctly all the values in the control. 
The **actual behaviour** is that it does not.
It appears that an additional value is needed, and this last value is the only that matters.

The **expected  behaviour** should be that sending only the required parameters as stated in the command reference should set the values correctly.

eduardfrigola:
Removes getting and setting the value prior to checking the type. This caused the function to try to get and set the value two times, with the second time getting and setting a value of 0, as no more osc messages were provided.

## Purpose and Motivation

Fixes #5973
As described in the issue linked above, _n_fill_ command seems to be broken (or at least not behaving the same way is documented.
Looking at the code and debugging the code execution, I've found that the first (and correctly documented) value argument passed via osc, is wrongly overwritten by another value that is always 0 (as no new value is provided) or is a non meaningful value of the next "block" of control - n values - value.

What I've detected (and comparing it to other similar commands such as _n_set_, or _n_setn_. Is that before checking the next osc message type, the value is requested and set, and after that, the "correct" (at least at my eyes) way of dealing with the message is applied, and thus, the first (and documented) value passed is not significant, as it's being overwritten by wrongly parsed, next argument.

## Types of changes

- Bug fix
- Breaking change

Removing the unnecessary getter of the value and setting the node control to that value before the correct way of dealing with it, I think achieves the correct behaviour of the n_fill command, but breaks code [where an extra parameter was used as a workaround to deal with this issue](https://github.com/PlaymodesStudio/ofxSuperCollider/blob/f633f1733bfbe3051f244f2880ec0b97ee2b89ce/src/ofxSCSynth.cpp#L142-L143).

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/server/scsynth/SC_MiscCmds.cpp
+++ b/server/scsynth/SC_MiscCmds.cpp
@@ -609,11 +609,6 @@ SCErr meth_n_fill(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRe
             int32* name = msg.gets4();
             int32 hash = Hash(name);
             int32 n = msg.geti();
-            float32 value = msg.getf();
-
-            for (int i = 0; i < n; ++i) {
-                Node_SetControl(node, hash, name, i, value);
-            }
 
             if (msg.nextTag('f') == 's') {
                 const char* string = msg.gets();
@@ -632,11 +627,6 @@ SCErr meth_n_fill(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRe
         } else {
             int32 index = msg.geti();
             int32 n = msg.geti();
-            float32 value = msg.getf();
-
-            for (int i = 0; i < n; ++i) {
-                Node_SetControl(node, index + i, value);
-            }
             if (msg.nextTag('f') == 's') {
                 const char* string = msg.gets();
                 if (*string == 'c') {


capital-G:
@eduardfrigola is this ready to be reviewed? It is still marked as draft/WIP

eduardfrigola:
Hi @capital-G, thanks for taking a look at this.
This is still a draft, as issue #5973 is still not discussed. And therefore this may try to fix something that is not really a bug.
I did this pull request to give an issue way to reproduce the differences in that issue between what I understand that is a bad behaviour, and what is correct behaviour.
As discussed in a meeting in the past, this pull request may not be the way to go to move forward with the issue, as the issue has not been discussed yet.

If is preferable, we can close this pull request now, and I can keep the branch that fixes the issue in our fork. Later on I can redo the pull request when the issue moves forward.

Eduard

PR #5984, opened 2023-02-24
GHA: use Xcode 10.3 for macOS legacy build

dyfer:
## Purpose and Motivation
This is a temporary change to make it work down to macOS 10.10. macOS 10.15 images will be unavailable at the end of March 2023.

This should allow for making one last release that supports macOS 10.10. After that we'll have to bump to... 10.11 :) (this seems to work fine on macOS 11 runner with Xcode 12)

This is a fix for an issue reported [on the forum](https://scsynth.org/t/problem-starting-sc-13-3/7188).

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -252,8 +252,8 @@ jobs:
             verify-app: true
 
           - job-name: 'x64 legacy'
-            os-version: '11'
-            xcode-version: '12.4'
+            os-version: '10.15'
+            xcode-version: '10.3'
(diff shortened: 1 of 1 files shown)


dyfer:
I have run the legacy builds on both macOS 10.10 and 10.13, copying it to the local filesystem and starting the server. It seems to work fine.

joshpar (approved):

PR #6001, opened 2023-03-26
fix Dictionary trueAt

### LINKED ISSUE #6000: Dictionary trueAt returns entry value not boolean if it exists
## Environment

* Operating system: MacOS (though presumable all affected)

## Steps to reproduce

```supercollider
c = Dictionary.newFrom([\foo, 3])

if (c.trueAt(\foo), {"true".postln})
```

This throws an error.

```
ERROR: Non Boolean in test.
RECEIVER:
   Integer 3
```

Looking at the code for `trueAt` this makes sense, but it's explicitly not what the docs say.

https://github.com/rumblesan/supercollider/blob/develop/SCClassLibrary/Common/Collections/Dictionary.sc#L25-L27
https://doc.sccode.org/Classes/Dictionary.html#-trueAt

Hapy to raise a PR to change either the dictionary code or the docs, but just want to make sure which it should be.

telephon:
This fixes #6000.

## Types of changes

- Bug fix
- Breaking Change

Considering the possible breaking changes:

|      trueAt            | obj | nil | true | false |
|-|--|-|-|-|
| previous: | obj | false | true |false |
| new:        | false | false | true |false |

So if you were assuming that `trueAt` and `falseAt` would return a boolean, your code would previously fail.
The change only extends the range of possible correct answers and does not change those that were already correct.

I however can imagine boundary cases where you'd use a `try { }` around the statement. Or, if someone were trying to fix the problem by calling `.trueAt(key).asBoolean`.

*Edit* There is an alternative possibility, that I favour (not implemented here yet, please discuss): The original idea of `trueAt` – I think – was to allow an empty (unspecified) value to count as `false`. So maybe better throw an error if it is neither:

```supercollider
trueAt { |key|
   var value = this.at(key);
   if(value.notNil and: { value.isKindOf(Boolean).not }) { 
       Error("trueAt requires either a boolean or nil entry").throw
   });
   ^value == true
}
```

Alternatively, one could also call `asBoolean` to allow numbers. But I don't like this so much.

=> please discuss

P.S. by the way, `Object.trueAt` returns `false` for any key. This may be an argument for *not* throwing an error in the case of a dictionary that has an object at the respective key?

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Collections/Dictionary.sc
+++ b/SCClassLibrary/Common/Collections/Dictionary.sc
@@ -23,7 +23,7 @@ Dictionary : Set {
 		^nil
 	}
 	trueAt { arg key;
-		^this.at(key) ? false
+		^this.at(key) == true
 	}
 	add { arg anAssociation;
 		this.put(anAssociation.key, anAssociation.value);
--- a/SCClassLibrary/Common/Collections/Dictionary.sc
+++ b/SCClassLibrary/Common/Collections/Dictionary.sc
@@ -23,7 +23,7 @@ Dictionary : Set {
 		^nil
 	}
 	trueAt { arg key;
-		^this.at(key) == true
+		^this.at(key).booleanValue
 	}
 	add { arg anAssociation;
 		this.put(anAssociation.key, anAssociation.value);
--- a/testsuite/classlibrary/TestDictionary.sc
+++ b/testsuite/classlibrary/TestDictionary.sc
@@ -27,4 +27,9 @@ TestDictionary : UnitTest {
 
 	}
 
+	test_trueAt {
+		var dict = (x:7, y:false, z:true);
+		this.assert([\x, \y].every(dict.falseAt(_)) and: dict.trueAt(\z), "dictionary trueAt and falseAt should respond only true if the value really equals true.");
+	}
+
 }


adcxyz (approved):

adcxyz:
For consistency with ```Object.trueAt``` being always false, the logic that anything except ```true``` is ```false``` makes sense. 
Also pragmatically, flags are set after/when conditions are fulfilled; before that, the key will usually be ```nil```, 
so assuming ```nil``` means ```false``` seems reasonable. 

Regarding numbers as Booleans, maybe better yes? 
People who are used to see 0/1 interchangeably with Booleans would expect that.

adcxyz (approved):
in discussion with telephon, agreed on 
.booleanValue to enable SimpleNumbers as flags.

telephon (commented):
looks good.

capital-G:
@telephon

I think this causes a failure in our test framework, see https://github.com/supercollider/supercollider/actions/runs/9197688519/job/25301259138?pr=6320#step:9:5954

grafik


https://github.com/supercollider/supercollider/blob/766791ca951f39a9479de2663fe51ab76dd06841/testsuite/classlibrary/TestDictionary.sc#L30-L33

vs the implementation

https://github.com/supercollider/supercollider/blob/766791ca951f39a9479de2663fe51ab76dd06841/SCClassLibrary/Common/Collections/Dictionary.sc#L25-L27

but

```supercollider
7.booleanValue
-> true
```

I think the implicit type conversion should be documented.

telephon:
yes, definitely. It was a change of behaviour. Seems we should have actuall run the tests :/

telephon:
here you are: https://github.com/supercollider/supercollider/pull/6321

mtmccrea:
It seems this has introduced inconsistent behavior:
https://github.com/supercollider/supercollider/pull/6321#issuecomment-2131289648

Issue #232, opened 2012-05-06
'try { loop ... }' doesn't unwind the stack properly if an error occurs within the loop


jleben:
[Issue migrated from SourceForge | ID: 3472743 | Submitted by 'jamshark70']
[http://sourceforge.net/support/tracker.php?aid=3472743]

```
x = [nil];
"%\n".postf(x.do({ |item|
try { max(1, x) == 2 }
}));
```

"try" should swallow the error, and the postf call should print the value of x (the return from 'do'). Instead:

```
ERROR: Message 'postf' not understood.
RECEIVER:
Instance of Array {    (0x48a1e78, gc=E0, fmt=01, flg=00, set=02)
  indexed slots [0]
}
ARGS:
Instance of Array {    (0x4831508, gc=E4, fmt=01, flg=00, set=02)
  indexed slots [1]
      0 : nil
}
```

The empty array that is the receiver of postf is postf's default argument -- see String.findMethod(\postf).prototypeFrame.

The most likely explanation is that some function or method arguments get pushed onto the stack, and they are supposed to be removed before postf, but the combination of 'try' and the loop causes this not to happen.
- It seems to occur only if 'try' is inside a loop.

```
"%\n".postf(try { max(1, nil) == 1 })  // postf succeeds
```
- The bug does NOT occur if the method applied to the loop result has no arguments (receiver only). I discovered the issue when trying to do ["First item"] ++ anArray.select { try { ... } { false } }.

```
// 'size' succeeds
x.select({ |item|
try { max(1, x) == 2 } { false }
}).size
```
- The issue does not occur in a five month old revision. I bet the bug was there ever since try/protect were introduced. Since it wasn't discovered in several years (!) it's not likely to cause major headaches in 3.5, so don't waste time trying to fix this for 3.5.

Marked for 3.6.

telephon:
Simpler reproducer: 

`x = [nil]; "%\n".postf(try { max(x, 1) });`

olafklingt:
I don't use try anymore because of i had problems because of this bug long ago. Now I am would like to know, is it possible to know under which conditions it is save to use try?

jamshark70:
```
x = [nil]; "hello".postf(try { max(x, 1) });
```

Based on Julian's simpler case, I'll update the issue title: there is no loop *outside* `try` here, but there is a loop hidden within `try` -- `max(array, something)` does a collectAs on the array. When the error occurs within a loop within try, then we see the bug.

[...]

PR #6003, opened 2023-04-05
Run testsuite directly

dyfer:
## Purpose and Motivation

This PR changes the procedure of running the test suite by replacing QPM with direct calls to sclang. 

The need to set up QPM makes it harder to recreate testsuite environment locally. IMO by running tests directly with sclang it's easier to debug tests as they might behave within the whole testsuite. Additionally QPM uses Python 2 which is deprecated.

### test runner

The [new test runner](https://github.com/supercollider/supercollider/blob/37923d1485d816a676176f113cfcaf938a48a173/testsuite/scripts/sclang_test_runner.scd) is very similar to the [test runner used previously in QPM](https://github.com/supercollider/qpm/blob/topic/try-catch-workaround/qpm/scscripts/test_runner.scd). The changes include removing some `try {}` statements (trying to avoid #232), as well as adjusting data format (see below).

### QPM vs direct calls to sclang

QPM had the following roles for the purpose of running tests IIUC:
1. adding `testsuite/classlibrary` to the compile paths
  1a.  adding `API` quark to the compile paths
2. running tests
3. printing the results
4. capturing output of sclang and parsing for error messages

(1), (2), and (3) is now done as 3 separate calls to sclang. (1a) is not needed. (4) is not implemented.

### data format for storing test results

Previously json format was used to store test results. This required using the `API` quark when running testsuite. This has been replaced by native SC objects. Additionally, the "proto" file that indicates which tests to run is now not overwritten with the results; instead a separate file is created to store the results.

### testing procedure

The testing procedure is largely the same as before, and its reliability hasn't improved that much, but I believe this paves the way for further improvements. One outcome of the proposed changes is that **the same script used in the CI can be run directly in the IDE** with virtually no changes.

### reliability

The reliability of the tests themselves is similar to the previous state. However, I believe the test environment is less likely to do "empty passes", like [here](https://github.com/supercollider/supercollider/actions/runs/4401489752/jobs/7774057478#step:9:341), which I _think_ was related to running array operations within `try {}` (i.e. the error reported in #232, IIUC). Still, occasional failures do happen: [some tests occasionally fail](https://github.com/dyfer/supercollider/actions/runs/4603959195/jobs/8136716497#step:9:128), [sclang segfaults](https://github.com/dyfer/supercollider/actions/runs/4621221849/jobs/8172706068#step:8:758), and [mysterious crashes occur when trying to run the whole testsuite on Windows](https://github.com/dyfer/supercollider/actions/runs/4603959195/jobs/8136716637#step:9:18).

### sclang scripts

This PR adopts the main test runner script from qpm and adds two other scripts: one to add `testsuite/classlibrary` to the compiled paths, and one to post the results. 

The scripts, if run outside of the IDE, uses ANSI color escape codes. This might be problematic if the scripts are run in a terminal that doesn't support these, but I'd argue this is an issue to be addressed if it actually proves problematic to anyone.

Posting results is done using a separate script and is triggered in a separate "step" in GHA. This makes it easier to see the summary of the results. However, in certain cases, if sclang (erroneously) doesn't compile the class library, the posting fails (I've only seen that on Windows). IMO this is something to investigate later.

### updated build matrix

As a part of this update, I've moved the Linux tests to run on a build done under Ubuntu 22.04 and GCC 12. **NOTE:** This also addresses the problem of the `ubuntu-18.04` images being deprecated.

### EDIT: fixed tests

I initially didn't intend to include any actual fixes to the UnitTests in this PR, but I've found both `TestThreadReschedule` and `TestSerialPort` failing/hanging on multiple runs; the included commits updating these two seem to make test suite more reliable.

###  test runner script vs `UnitTest.runAll`

One could argue that we shouldn't need a separate test runner script and that it should be enough to run `UnitTest.runAll`. The test runner provides the following functionality, not provided by `UnitTest.runAll`:
- store test results incrementally in a file
- catch unhandled errors thrown in the testsuite
- print the results after running the tests

We could consider migrating/integrating that functionality into UnitTest.runAll, but I believe that would be something for another PR. In any case, it would be useful to have everything directly in the main repository, as opposed to running with QPM.

### documentation

There's only minimal documentation on qpm and running testsuite in the CI ([1](https://github.com/supercollider/supercollider/wiki/Miscellaneous-project-information-(CI,-maintenance-scripts,-etc.)#qpm-test-runner), [2](https://github.com/supercollider/supercollider/wiki/Continuous-Integration---GitHub-Actions#test-suite)); no documentation in the helpfiles AFAICT. This PR does not add any documentation... partially because I think this is still a work in progress - we might move the runner script into the UnitTest class, or change it considerably. 

_If you feel that some documentation should be added to this PR, please let me know_. Otherwise I'd just update the two linked wiki entries indicating that we are now running testsuite directly in sclang after this PR is merged.

### tests on Windows

Testing should now be possible on all platforms using essentially the same code.

Testing environment on Windows was added and the pipeline seems to work, but it's currently disabled. There seem to be additional issues on Windows that result in sclang's unexpected behavior, which prevents the testuite from ever finishing. I made some progress with that, but I feel it should be saved for another PR. 

[You can see a small subset of the testsuite running on Windows here](https://github.com/dyfer/supercollider/actions/runs/4621249778/jobs/8173230596).

Since GHA runners offer no audio hardware, the Windows runner installs Jack2 with JackRouter (virtual ASIO soundcard), which is then used by scsynth during tests.

### running time of the whole suite

I believe that the running time of the whole suite is similar to previous implementation. However, the overall running time is **not being reported** next to the number of passed tests, as it has been with qpm implementation. This is something I could add if desired, but IMO it's enough that the time of the testing step is reported in GitHub Actions UI.

## Types of changes

- ~~Documentation~~
- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing (as much as they had before)
- [ ] Updated documentation (n/a?)


joshpar (approved):

dyfer:
@jrsurge @joshpar just to confirm: I've run these scripts in both Windows Command Prompt and in PowerShell. In both cases they print output properly _and_ in color (i.e. properly interpreting ANSI escape codes). Moreover, the shells are fully operational afterwards.

Issue #4117, opened 2018-10-07
`Pipe.callSync` interface needs review


jamshark70:
Environment
-----------

* Your SuperCollider version: 90f968b23e
* Your operating system and version: Ubuntu Studio 16.04 (not relevant though)

Issue summary
-------------

`Pipe.callSync` has a number of weaknesses.

It discards the process's exit code (claiming, in a comment, that there is no way to get it). But the exit code is the method return from `Pipe:close`. So it is actually possible to get, and use, the exit code. https://github.com/jamshark70/supercollider/commit/8f8dc0f2525710048a93244929fc13fd8df7dc26#diff-e9e453d1b56d13140f47280d5d902486R211

Then... There are `onSuccess` and `onError` functions. But it decides which one to call based not on the process's error code, but on the presence or absence of text output. As Brian pointed out elsewhere, this means:

- `echo blah && exit 1` calls `onSuccess`, despite there being a nonzero exit code.

- The following calls `onError`, despite exitcode == 0.
  ```supercollider
  f = File(PathName.tmp +/+ "test.txt", "w");
  f.close;
  
  Pipe.callSync("cat " + (PathName.tmp +/+ "test.txt"), { "success".postln }, { "failure".postln });

  File.delete(PathName.tmp +/+ "test.txt");
  ```

There is no way to fix this without a breaking change. (`callSync` is used throughout Quarks.)

Probably the best course is to deprecate callSync and replace it with something that uses the exit code to make more logical decisions.

Opening an issue for discussion of the proper interface, before coding.

Also: `Pipe.call` wraps `callSync` in a Routine, but this is largely pointless because the pipe will block the interpreter anyway. Suggest to deprecate that method too.

brianlheim:
As I said on your now-closed PR, I think the original design is more or less irredeemable. The way I would expect this call to work is:

```supercollider
Pipe.execute(cmdLine, onSuccess, onFailure)
```

where `onSuccess` and `onFailure` both have parameters `output: String, exitCode: Integer`, and rough pseudocode implementation is:

```
execute { |cmdLine, onSuccess, onFailure|
    p = Pipe.open(cmdLine)
    o = p.getOutput()
    c = p.getExitCode()
    return c == 0 ? onSuccess(o, c) : onFailure(o, c)
}
```

[...]

PR #6006, opened 2023-04-07
Topic/two way pipe

scztt:
## Purpose and Motivation

sclang currently has no ability to communicate bidirectionally with a subprocess called by `argv`. This introduces a primitive which opens both a read and write pipe (stdin / stdout) for a subprocess and returns both.

Note that this is NOT intended as a "high-level" interface to bidirectional pipes, but only as an initial primitive that a user-facing representation could be built on top of. Designing proper bidirectional subprocess communication is a larger pure-sclang task that can be tackled separately.

This change is critical for the sclang code formatter (https://github.com/scztt/sclang-format).

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

## Unresolved

- **WINDOWS SUPPORT**: The Windows `argv` implementation is completely separate from the OSX. My knowledge of pipes is limited on unix systems, and basically zero on Windows, so some help would be appreciated in implementing the Windows half of this PR.
- **PID MANAGEMENT**: Currently, Pipes have an associated pid if they correspond to a subprocess - the subprocess is closed when the pipe is closed. Since we have two pipes here, we have a two choices... We can bind the pid to ONE of the two Pipes - closing this pipe would result in the underlying process being closed. Or (the implementation in this PR), we can tie the pid to BOTH pipes, and close it when the first pipe is closed. This can result in an error message if code attempts to close the second pipe, since the PID is already closed. It's important to note again that this is a *primitive* API - ideally, this would be wrapped in a high-level user facing class that handles the details so users don't have to care about closing what Pipe when.
- **BLOCKING**: As is always the case with pipes, reads are blocking and it's easy to hang sclang if the pipe is not ready for reading. As I'm not THAT familiar with pipes, I'm not sure if there are things we can do to avoid some of these language hangs, either in the primitive itself or via sclang code. Help / advice here would be appreciated.

### DIFF
--- a/SCClassLibrary/Common/Files/File.sc
+++ b/SCClassLibrary/Common/Files/File.sc
@@ -184,6 +184,17 @@ Pipe : UnixFILE {
 		^super.new.openArgv(args, mode);
 	}
 
+	*argvReadWrite { arg args;
+		var pid, readPipe, writePipe;
+
+		#pid, readPipe, writePipe = this.prOpenArgvReadWrite(args);
+
+		^[
+			super.newCopyArgs(readPipe, pid).addOpenFile(), 
+			super.newCopyArgs(writePipe, pid).addOpenFile()
+		]
+	}
+
 	*call { arg command, onSuccess, onError, maxLineLength=4096;
 		var r, cancel, closePipe;
 		r = Routine.run({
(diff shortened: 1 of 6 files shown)


dyfer:
Thanks! Just FYI I noticed there was some prior discussion on SC's `Pipe` interface: #4117

telephon:
sounds good to me. Boost is a dependency anyhow, so it should be ok to use it also for this part.

Spacechild1:
> BLOCKING: As is always the case with pipes, reads are blocking

On macOS/Linux, pipes can be made non-blocking and you can use them with `select` and `poll`.

On Windows you can use wait modes (https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipe-type-read-and-wait-modes) or overlapped I/O. For non-blocking reads you can also use `PeekNamedPipe`.

I have used pipes quite a bit on Windows, macOS and Linux. If you have any questions, feel free to ask!

dyfer:
@scztt are you planning to keep working on this?

PR #6009, opened 2023-04-14
docs: Add live signal example for GrainBuf

tdug:
## Purpose and Motivation

I wanted to add a GrainBuf example using a live signal, as this was somewhat tricky to figure out.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/GrainBuf.schelp
+++ b/HelpSource/Classes/GrainBuf.schelp
@@ -59,9 +59,10 @@ instancemethods::
 private:: init, argNamesInputsOffset
 
 examples::
+
+From a link::Classes/Buffer::
 code::
 s.boot;
-
 (
 var winenv;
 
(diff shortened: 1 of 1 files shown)


nhthn (changes requested):
Thanks! I see two major issues here:

- too much decoration, recommend against using Ndef, Pbind, or FreeVerb2. Minimal examples preferred over musical examples.
- BufWr and GrainBuf write block by block so it is possible for the GrainBuf to read over the discontinuity, creating clicks. I think ensuring that grainPos never crosses over phasor - BlockSize.ir, minus a few extra samples to deal with cubic interpolation, might be enough to prevent clicks

It sucks that SC doesn't have a good grain delay UGen, I might get on that eventually.

joshpar (changes requested):
Thank you for the examples! I am find pulling this in, but I also think we need to have a more basic example showing how to interact with the live signal in a straightforward manner, and using SynthDef / Synth. Reverbs and anything extra should be left out of that example. Then, I feel like that AND this would be fine to include in the docs. Thanks!

mtmccrea:
@tdug, do you plan to update this PR in response to the feedback above, or should we close this for now?

capital-G:
I'll close this for now as this seems to be stale? Feel free to re-open if some of the concerns mentioned here are addressed.

PR #6014, opened 2023-04-23
(supecollider/supercollider - improving documentation) Update docs with the help of ChatGPT

tiagomoraismorgado:
(supecollider/supercollider - improving documentation) Update docs with the help of ChatGPT
please feel free to merg this to my previous pull requests

### DIFF
--- a/HelpSource/Classes/AudioIn.schelp
+++ b/HelpSource/Classes/AudioIn.schelp
@@ -9,7 +9,7 @@ Description::
 Reads audio from the sound input hardware.
 
 note::
-This is provided for backwards compatibility with SC2 code. For normal use link::Classes/SoundIn::, which has bus numbers beginning at 0, as AudioIn may be deprecated and removed at some point in the future.
+The backward compatibility with SC2 code is the reason for providing this, but for regular use, it is recommended to use link::Classes/SoundIn::, which has bus numbers starting from 0. The AudioIn class may become obsolete and eliminated in the future.
 ::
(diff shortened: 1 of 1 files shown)


telephon (commented):
This is not an improvement.

tiagomoraismorgado:
I initially felt it could be an improvement. text is grammarly correct. but the way of efficiency in terms of natural language processing of chatgpt is high. so, I thought, why not contributing at polishing supercollider documentation with the aid of chatgpt. anyway I could do my own version of the whole supercollider documentation, put it available to the entire community. but then it could look like overwhelming to integrate in the supercollider project, specially if there was no aiming at that. and provide better language, and better examples within code. because we could end up creating a pi mode documentation for supercollider, allowing people to develop supercollider programming faster. even though there is a reason why we maintain things intact and only turn them more solid, which is to develop things on the longer run more carefully, and provide assitance. so tell me is a reviewed/anyway version of supercollider documentation desirable, or should this be like a separate project? please drop my comments on that. and please delete the post if this commit is not desirable

cheers
T.

telephon:
I think you would have to get consent from each author of the supercollider documentation to feed it into such an AI-database. This is probably not going anywhere, because some will certainly disagree. Also, due to extremely subtle mistakes the whole processed documentation would have to be reviewed carefully, which is a lot of work, more than any reviewer is ready to do.

nhthn:
Don't do this.

PR #6026, opened 2023-04-28
clarify documentation for randomness methods

mxw:
updates SimpleNumber and Operators docs. the randomness guide seems a bit haphazard so i didn't touch it.

## Purpose and Motivation

documentation is good

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/SimpleNumber.schelp
+++ b/HelpSource/Classes/SimpleNumber.schelp
@@ -974,51 +974,84 @@ stores this on the given stream
 method:: printOn
 prints this on the given stream
 
+
 subsection:: random
+Unless noted otherwise, when the receiver and arguments to these methods are all link::Classes/Integer::s, the return will be as well.
 
 method:: coin
 Let emphasis::x:: be the receiver clipped to the range [0, 1]. With probability emphasis::x::, return true. With probability 1 - emphasis::x::, return false.
+code::
+0.4.coin;
+
+{ SinOsc.ar(110).coin }.plot;
+::
 
 method:: rand
-returns:: Random number from zero up to the receiver, exclusive.
+Random number uniformly distributed in the range [0, this).
+code::
+10.rand;
+
+{ SinOsc.ar(110).rand }.plot;
+::
 
 method:: rand2
-returns:: a random number from -this to +this.
+Random number uniformly distributed in the range [-this, +this].
+code::
+10.rand2;
+
+{ SinOsc.ar(110).rand2 }.plot;
+::
 
 method:: rrand
-argument::aNumber
-the upper limit
+Random number uniformly distributed in the range [this, hi]. Note that unlike link::Classes/SimpleNumber#-rand::, the range is inclusive of the upper bound.
+argument::hi
+The upper limit of the range.
 argument::adverb
-returns:: a random number in the interval ]a, b[.
-discussion::
-If both a and b are link::Classes/Integer:: then the result will be an link::Classes/Integer::.
+discussion:: Equivalent to code::(hi - lo) * 1.0.rand + lo::, except for the inclusive upper bound.
 
 method:: linrand
-returns:: a linearly distributed random number from zero to this.
+Random number linearly distributed in the range [0, this], i.e., sampled from a distribution over [0, 1] with PDF 2(1 - x) and scaled by this.
+code::
+10.linrand;
+
+{ SinOsc.ar(110).linrand }.plot;
+::
 
 method:: bilinrand
-returns:: Bilateral linearly distributed random number from -this to +this.
+Bilateral linearly distributed random number in the range [-this, +this]. Like linrand, but the PDF is also supported in the range [-1, 0] by 2(x + 1).
+code::
+10.bilinrand;
+
+{ SinOsc.ar(110).bilinrand }.plot;
+::
 
 method:: sum3rand
-This was suggested by Larry Polansky as a loose approximation of gaussian.
-returns:: A random number from -this to +this that is the result of summing three uniform random generators to yield a bell-like distribution.
+Random number from -this to +this that is the result of summing three uniform random generators to yield a bell-like distribution.
+code::
+10.sum3rand;
+
+{ SinOsc.ar(110).sum3rand }.plot;
+::
+
+discussion::
+This was suggested by Larry Polansky as a loose approximation of Gaussian. It is equivalent to:
+code::
+{ 1.0.rand }.dup(3).sum - 1.5 * (2/3)
+::
 
 method:: exprand
-an exponentially distributed random number in the interval ]a, b[. This is always a link::Classes/Float::.
-(Note that the distribution of numbers is not exactly an EMPHASIS::exponential distribution::, since that would be unbounded: we might call it a EMPHASIS::logarithmic uniform distribution::.)
-argument::aNumber
-the upper limit
+Random number in the range [this, hi] drawn from the log-uniform distribution. Note that the value is emphasis::not:: drawn from an exponential distribution.
+argument::hi
+The upper limit of the range.
 argument::adverb
+returns:: Always a link::Classes/Float::.
+discussion:: Equivalent to code::(hi / lo) ** 1.0.rand * lo::, except for the inclusive upper bound.
 
 method:: gauss
-a gaussian distributed random number.
-argument::standardDeviation
-the upper limit
-discussion::
-Always returns a link::Classes/Float::.
-code::
-(0..1000).collect { |num| gauss(0.0, num) }.plot;
-::
+Random number sampled from a Gaussian distribution with mean and standard deviation (this, stddev).
+argument::stddev
+The standard deviation.
+returns:: Always a link::Classes/Float::.
 
 method:: partition
 randomly partition a number into parts of at least min size.
--- a/HelpSource/Classes/SimpleNumber.schelp
+++ b/HelpSource/Classes/SimpleNumber.schelp
@@ -71,7 +71,7 @@ method:: round
 Round to multiple of aNumber
 
 method:: roundUp
-Round up to a multiple of aNumber. For roundDown use Link::Classes/SimpleNumber#-trunc#trunc::.
+Round up to a multiple of aNumber. For roundDown use link::Classes/SimpleNumber#-trunc#trunc::.
 
 method:: smallButNotZero
 Check if the value is closer to zero than a threshold, but not zero.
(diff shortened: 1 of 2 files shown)


HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -974,52 +974,85 @@ stores this on the given stream
 method:: printOn
 prints this on the given stream
 
+
 subsection:: random
+Unless noted otherwise, when the receiver and arguments to these methods are all link::Classes/Integer::s, the return will be as well.
```
telephon, line 979: maybe better: 
Unless noted otherwise, the methods return numbers of the same class as the receiver and arguments.

Currently, SimpleNumber has only Floats and Interger as subclasses, but this may be extended.

HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -974,52 +974,85 @@ stores this on the given stream
...
+{ SinOsc.ar(110).rand }.plot;
+::
 
 method:: rand2
-returns:: a random number from -this to +this.
+Random number uniformly distributed in the range [-this, +this].
+code::
+10.rand2;
+
+{ SinOsc.ar(110).rand2 }.plot;
```
telephon, line 1002: This is an example of `rand2` for UGen. So this is potentially confusing, especially because the UGen behaves like a sequence of floats, not ints. 

Maybe better:

`{ 10.rand2 }.dup(30).plot.plotMode_(\steps)`

`{ 10.0.rand2 }.dup(30).plot.plotMode_(\steps)`

There are more such cases, maybe you can reconsider them?

Edit: I just saw that you copy the methods into operators. This is good, but on the other hand, there we may want to collect a more varied set of examples. It is a trade-off in documentation.

I'd say, keep this one really short and make a more extended one for Operators.

HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -974,52 +974,85 @@ stores this on the given stream
...
+The upper limit of the range.
 argument::adverb
-returns:: a random number in the interval ]a, b[.
-discussion::
-If both a and b are link::Classes/Integer:: then the result will be an link::Classes/Integer::.
+discussion:: Equivalent to code::(hi - lo) * 1.0.rand + lo::, except for the inclusive upper bound.
 
 method:: linrand
-returns:: a linearly distributed random number from zero to this.
+Random number linearly distributed in the range [0, this], i.e., sampled from a distribution over [0, 1] with PDF 2(1 - x) and scaled by this.
```
telephon, line 1013: "PDF" will be hard for people to search on the internet :) Maybe expand?

telephon (commented):
Thank you, this is very good. I've added some suggestions.

HelpSource/Classes/SimpleNumber.schelp
mxw: your wording is ambiguous—this line is intended to indicate that you get an int back if and only if all participants are ints. "of the same class" is unspecific when the receiver and arguments are of different types.

HelpSource/Classes/SimpleNumber.schelp
mxw: can do.

HelpSource/Classes/SimpleNumber.schelp
mxw: i'm not terribly interested in authoring a collection of good examples; i believe my change is an improvement to the documentation even if i deleted most of the existing ones (which are... mediocre).

i'll add a code block before the copymethods in the Operators helpfile to indicate these operators work with a diversity of receivers (SimpleNumber, UGen, Array), along with your plot example, and i'll delete the copypasta examples from SimpleNumber.

mxw: well, i guess the mere inclusion of Array as a receiver is an improvement in its own right.

HelpSource/Classes/SimpleNumber.schelp
telephon: I see. Then you should probably write "only when" or add "only" to the end.

telephon:
> i'm not terribly interested in authoring a collection of good examples;

you don't need to be. If you want to improve upon your own improvements, you could keep operators and SimpleNumber separate, so someone can fix them separately.

telephon:
Just to clarify the intention: the operators help file was written to collect examples across different classes, for direct comparison. So copying them over from one single class may look the easiest thing to do right now, but doesn't really help that much. It can sometiumes be a drag to have duplicate information, but there I don't see a better way here.

HelpSource/Classes/SimpleNumber.schelp
mxw: done. also added a note that sum3rand is always a Float.

mxw:
@telephon that makes sense, though i think the changes to the Operator docs are also an improvement as-is. i don't think the N copies of 

```
10.foo;

{ SinOsc.ar(110).foo }.plot;
```

were very enlightening beyond perhaps the first copy. i've kept one copy and added your plot comparison and a line for these methods on Array. i'm going for "useful", not "ideal".

that said, i don't feel strongly at all; feel free to commandeer and revert the Operator help changes if you like.

mxw:
also, apologies if i'm coming across as curt. i just wanted to do some drive-by documenting after i spent time figuring out what these methods actually did. i'm sure this isn't the clearest it could be, but it wasn't ideal before and i'm just trying to avoid spending too much time on it. hope you understand!

telephon:
yes, all fine – I think it should be much easier to improve documentation on the fly. Also it is sometimes a little counter-intuitive to learn the intent of a document.

> i don't think the N copies of
> 
> ```
> 10.foo;
> 
> { SinOsc.ar(110).foo }.plot;
> ```
> 
> were very enlightening beyond perhaps the first copy.

The almost identical examples in the operators helpfiles are meant to be run by the reader. If they are not there, nobody will try them out. So they are not meant to reveal anything special through reading.

Much of the documentation of sclang is like this, and in an interactive environment that makes sense.

nhthn (approved):
Thanks!

telephon:
@nhthn I still had requests for changes (didn't want to block, in order not to be unwelcoming). 

Do you disagree? 

> The almost identical examples in the operators helpfiles are meant to be run by the reader. If they are not there, nobody will try them out. So they are not meant to reveal anything special through reading.
> 
> Much of the documentation of sclang is like this, and in an interactive environment that makes sense.

HelpSource/Overviews/Operators.schelp
```diff
@@ -122,62 +122,27 @@ Convert cycles per second to decimal octaves.
...
 ::
 
-method::coin
-Returns one or zero with the probability given by the argument.
-code::
-0.4.coin;
-
-{ SinOsc.ar(110).coin }.plot;
-::
+copymethod:: SimpleNumber -coin
```
telephon, line 138: I'd prefer if these weren't referenced. There was a good reason why they were here separately (see dicussion).

telephon (changes requested):

cdbzb:
Hi @telephon  - @mxw wrote "that said, i don't feel strongly at all; feel free to commandeer and revert the Operator help changes if you like." --perhaps that might be the simplest way to get this PR finished? I do find the proposed changes to SimpleNumber clarifying so it would be good to get at least that done.

cdbzb:
@mxw wrote: 
>that said, i don't feel strongly at all; feel free to commandeer and revert the Operator help changes if you like.

might it be posssible to merge the changes to SimpleNumber.schelp only? I actually found these illuminating!

telephon:
It is fine to copy some useful examples over to `SimpleNumber` help. But the `Randomness` file should better be kept and edited separately, (not using `copymethod`). It is not always better to avoid duplicates – in this case, the two documents serve slightly different purposes: the main object helpfile should give short examples, the overview should give a comparative story.

cdbzb:
Thanks for the speedy reply! Yes I understand - my suggestion is simply to accept the changes to the SimpleNumber.schelp file (which illuminate the issues of bounds, return values and distribution function), and to reject the changes to the Operators.schelp file. Since all the parts that you object to are in the Operators file that ought to work for you I hope!

telephon:
Yes, that is fine!

cdbzb:
Hi @dyfer it looks like everyone is ok with the plan of merging the changes in HelpSource/Classes/SimpleNumber.schelp
and rejecting the changes in
HelpSource/Overviews/Operators.schelp

Is that straightforward? could get one more PR off the list FWIW ;)

dyfer:
> Hi @dyfer it looks like everyone is ok with the plan of merging the changes in HelpSource/Classes/SimpleNumber.schelp
> and rejecting the changes in HelpSource/Overviews/Operators.schelp

@mxw would you be able and willing to keep only the changes in SimpleNumber.schelp in this PR and remove the changes to the Operators.schelp

dyfer:
At the dev meeting we agreed that @HotwheelsSisyphus will try to finish this PR

HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -974,51 +974,55 @@ stores this on the given stream
...
 method:: rrand
-argument::aNumber
-the upper limit
+Random number uniformly distributed in the range [this, hi]. Note that unlike link::Classes/SimpleNumber#-rand::, the range is inclusive of the upper bound.
+argument::hi
+The upper limit of the range.
 argument::adverb
-returns:: a random number in the interval ]a, b[.
-discussion::
-If both a and b are link::Classes/Integer:: then the result will be an link::Classes/Integer::.
```
mtmccrea, line 994: I think this is an important note to retain in the discussion.

cdbzb: this commit moves the note on integer from each individual method to the subsection...

mtmccrea: Ah, thanks for pointing that out.  I guess it's easy to miss if one looks up documentation for a method directly, expecting the relevant info to be there. Not sure what the best approach is 🤔

HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -974,51 +974,55 @@ stores this on the given stream
 method:: printOn
 prints this on the given stream
 
+
 subsection:: random
+Unless noted otherwise, the return value of all these methods will be an link::Classes/Integer:: only when the receiver and all arguments are Integers as well.
```
mtmccrea, line 979: I wonder if it would be useful to add a note describing the notation used below for in/exclusive bounds [], [), etc. otherwise to make it explicit in each method description (which I find useful)

HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -974,51 +974,55 @@ stores this on the given stream
...
+Random number linearly distributed in the range [0, this], i.e., sampled from a distribution over [0, 1] with a probability density function (PDF) of 2(1 - x) and scaled by this.
 
 method:: bilinrand
-returns:: Bilateral linearly distributed random number from -this to +this.
+Bilateral linearly distributed random number in the range [-this, +this]. Like linrand, but the PDF is also supported in the range [-1, 0] by 2(x + 1).
 
 method:: sum3rand
-This was suggested by Larry Polansky as a loose approximation of gaussian.
-returns:: A random number from -this to +this that is the result of summing three uniform random generators to yield a bell-like distribution.
+Random number from -this to +this that is the result of summing three uniform random generators to yield a bell-like distribution.
```
mtmccrea, line 1004: you updated other method descriptions to use "in the range [-this, +this]" instead of "from -this to +this", so good to use it here for consistency.

HelpSource/Classes/SimpleNumber.schelp
HotwheelsSisyphus: for the style guide, we settled (for accessibility reasons) on the following: 

> When an argument expects a certain range, specify the minimum and maximum valid inputs. Where appropriate, specify invalid inputs: For example 0 is an invalid argument for the exponential range in .explin. Mathematical interval notation (such as [0,1)) is welcome, but should be accompanied by prose: “Between zero and one (including zero, not including one).”

HotwheelsSisyphus: I wonder, is the inverted square brackets ]0, 1[  notation for open sets common in computer science/programming contexts? In more mathy sources I've only seen (0, 1), but I think the docs uses the ][ notation at multiple places. (Question is ultimately: A. acknowledge the ][ option in style guide, or B. change all instances of ][ to () ?)

HotwheelsSisyphus: Fwiw, I did a quick grep on the HelpSource folder and it seems like the two files in here may be the only ones to use that notation... so this is easy to replace then.

HelpSource/Classes/SimpleNumber.schelp
tedmoore: I know it would mean putting the note about integers in *every* one of these methods, but I'd be in favor of that. When searching docs and coming to look at just one method, I think it's good to have the information there for the reader.

HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -974,51 +974,55 @@ stores this on the given stream
...
+returns:: Always a link::Classes/Float::.
+
+discussion::
+This was suggested by Larry Polansky as a loose approximation of Gaussian. It is equivalent to:
```
tedmoore, line 1008: I don't personally know Larry, but this seems to be unnecessary information. Many people contribute in many ways and citing all would be onerous. Perhaps there's a context or history that I'm unaware of?

HotwheelsSisyphus: I agree that this is unnecessary here, especially if there is no specific publication being cited (and avg of n uniform distributions approaching gaussian as n goes to inf... isn't that just the central limit theorem? So this may not be a super remarkable suggestion anyway, requiescat in pace).

Otoh, I think where a note like this is appreciated is in the implementation code... in this case, I checked and it is there (plugin_interface/SC_RGen.h), so we're not erasing any historical details anyway...

mxw:
can't believe these commits are still pending and under active discussion !

@HotwheelsSisyphus if you look at the original commits, the then-contemporary version of the docs _seemed_ to suggest inclusive bounds to me, based on the use of square brackets. of course, there was quite a… diverse array of notations and phrasings used to describe bounds, so maybe i shouldn't have read into it.

i also figured it's better to err on the side of inclusive, since the likelihood of actually producing a boundary value is theoretically infinitesimal, so a user should thus neither expect a boundary value will definitely be returned, nor should they expect a boundary value to _never_ be returned.

i agree with your reading of the source code. i am not sure what you are testing for—no amount of simulation can prove the exclusion of the bound in question.

anyway, see y'all in 2027

HotwheelsSisyphus:
> no amount of simulation can prove the exclusion of the bound in question.
haha I know, BUT on a stupidly pragmatic note, the lower bound did show up reliably every ten million tries or so, and the upper bound did _not_ show up.

telephon (approved):

dyfer:
At the SC release meeting we applied changes suggested by @telephon and decided to merge this.

HotwheelsSisyphus:
Ok, though I have an almost-finished update... that can wait until the next release though. Sorry for being behind on this, just had a job interview today and was a bit busy with prep for that...

capital-G:
> Ok, though I have an almost-finished update... that can wait until the next release though. Sorry for being behind on this, just had a job interview today and was a bit busy with prep for that...

We pushed the release of RC1 back to monday, so there is still a chance it may become part of 3.14

Hope the job interview went well :)

HotwheelsSisyphus:
Ok, here's that update. I'll reword the commit name when I squash that... I'll look at it the whole thing again tomorrow for typos etc., I should be able to finalize it before the meeting; unless @tedmoore or @telephon have time to give it a quick once-over.

--changes to .coin are only line breaks.
-- this might be too much info in some cases? idk.

PR #6035, opened 2023-05-17
Fix UGen initialization: LFUGens pt. 1/3, Trig, Trig1

### LINKED ISSUE #4279: Line, XLine: Incorrect starting value, early arrival, etc.
## Environment

* SuperCollider version: 3.10.1
* Operating system: MacOS

## Steps to reproduce

```supercollider
/* Case 1 */
(
s.waitForBoot({ // ar - BUG: doesn't start at 0, arrives 1 sample early
	{
		Line.ar(0, 7, (5 / s.sampleRate))
	}.loadToFloatArray((6 / s.sampleRate), action: {|arr| arr.do(_.postln)});
})
)
/* Case 2 */
(
s.waitForBoot({ // kr - BUG: same as AR
	{
		Line.kr(0, 7, (5 * s.options.blockSize / s.sampleRate))
	}.loadToFloatArray((6 * s.options.blockSize / s.sampleRate), action: {|arr| arr.do(_.postln)});
})
)
/* Case 3 */
(
// kr variant - BUG: despite non-zero duration, Line "starts" at the end
// Update: this isn't a bug, see conversation below
s.waitForBoot({ 
	{
		Line.kr(0, 7, (5 / s.sampleRate)) // NOTE: dur shorter than control period
	}.loadToFloatArray((6 * s.options.blockSize / s.sampleRate), action: {|arr| arr.do(_.postln)});
})
)
```

I expect `Line` to start at ... the starting value, and end after the duration has elapsed (the first sample marking time `0`).
Also, I would expect that if the `duration` is non-zero, the output will always begin with the starting value. If the `duration` is less than the sample duration (at either `ar` or `kr`), the end value would occur at the sample immediately following the end of 'duration` (see Case 3).

There are identical bugs in `XLine`.

### The solution: underway
I have a fix for Case 1, and 2 as part of a larger overhaul of UGen initialization, but found Case 3 today and wanted to document it so I remember to come back around to it.

This bug is related to the many initialization sample bugs I've noted in a number of other places, but this particular problem looks to be related to the slope being incremented before writing out the initialization sample, not after.

mtmccrea:
## Purpose and Motivation

This fixes the initialization of a batch of `LFUGens`:
- `Line`, `XLine`
- `Vibrato`
- `LFPulse`, `LFSaw`, `LFPar`, `LFCub`, `LFTri`
- `Trig` and `Trig1` were also fixed (they're in `TriggerUGens`) - this was needed in order to pass related unit tests

`Line` was also made to call `Line_next_nova_64`, which was omitted previously.

(fixed Issues will be updated here)
- Fixes #4279

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
  - initial phase tests were added for LFPulse, LFSaw, LFPar, LFCub, LFtri, SinOsc
  - start/end value test was added for Line, XLine
  -  The following Unit tests were made to pass with these changes:
```supercollider
TestCoreUGens: test_ugen_generator_equivalences - "SinOsc.kr can match Line.kr.sin"
TestCoreUGens: test_ugen_generator_equivalences - "SinOsc.kr can match Line.kr.cos"
TestCoreUGens: test_ugen_generator_equivalences - "Trig1.ar(_,0) has same effect as (_>0) on variable-amplitude impulses"
TestCoreUGens: test_ugen_generator_equivalences - "Trig.ar(_,0) is no-op when applied to Impulse.ar, whatever the amplitude of the impulses"
TestCoreUGens: test_ugen_generator_equivalences - "Trig1.ar(_,0) is no-op when applied to Impulse.ar"
```
  - Note there are failing tests for `Duty` in `TestCoreUGens`, and `TestEnvGen_server: test_zero_gate_n_off_not_sent`, but they were failing before so are unrelated.
- [ ] Updated documentation - **N/A**

### DIFF
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -588,11 +588,12 @@ void LFPar_Ctor(LFPar* unit) {
         SETCALC(LFPar_next_k);
 
     unit->mFreqMul = 4.0 * unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1);
+    double initPhase = unit->mPhase = ZIN0(1);
 
     LFPar_next_k(unit, 1);
-}
 
+    unit->mPhase = initPhase;
+}
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
 
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -676,9 +676,12 @@ void LFTri_Ctor(LFTri* unit) {
     }
 
     unit->mFreqMul = 4.0 * unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1);
+
+    double initPhase = unit->mPhase = sc_wrap(static_cast<double>(ZIN0(1)), 0.0, 4.0);
 
     LFTri_next_k(unit, 1);
+
+    unit->mPhase = initPhase;
 }
 
 
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -1401,7 +1401,7 @@ FLATTEN void Line_next_nova_64(Line* unit, int inNumSamples) {
 void Line_Ctor(Line* unit) {
 #ifdef NOVA_SIMD
     if (BUFLENGTH == 64)
-        SETCALC(Line_next_nova);
+        SETCALC(Line_next_nova_64);
     else if (boost::alignment::is_aligned(BUFLENGTH, 16))
         SETCALC(Line_next_nova);
     else
@@ -1419,7 +1419,6 @@ void Line_Ctor(Line* unit) {
     } else {
         unit->mLevel = start;
         unit->mSlope = (end - start) / unit->mCounter;
-        unit->mLevel += unit->mSlope;
     }
     unit->mEndLevel = end;
     ZOUT0(0) = unit->mLevel;
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -522,9 +522,11 @@ void LFSaw_Ctor(LFSaw* unit) {
         SETCALC(LFSaw_next_k);
 
     unit->mFreqMul = 2.0 * unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1);
+    double initPhase = unit->mPhase = ZIN0(1);
 
     LFSaw_next_k(unit, 1);
+
+    unit->mPhase = initPhase;
 }
 
 
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -475,10 +475,13 @@ void LFPulse_Ctor(LFPulse* unit) {
     }
 
     unit->mFreqMul = unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1);
-    unit->mDuty = ZIN0(2);
+    double initPhase = unit->mPhase = ZIN0(1);
+    float initDuty = unit->mDuty = ZIN0(2);
 
     LFPulse_next_k(unit, 1);
+
+    unit->mPhase = initPhase;
+    unit->mDuty = initDuty;
 }
 
 
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -289,9 +289,6 @@ void Linen_Ctor(Linen* unit);
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
 
-// in, rate, depth, rateVariation, depthVariation
-// 0   1     2      3              4
-
 void Vibrato_next(Vibrato* unit, int inNumSamples) {
     float* out = ZOUT(0);
     float* in = ZIN(0);
@@ -402,23 +399,30 @@ void Vibrato_next(Vibrato* unit, int inNumSamples) {
 
 void Vibrato_Ctor(Vibrato* unit) {
     unit->mFreqMul = 4.0 * SAMPLEDUR;
-    unit->mPhase = 4.0 * sc_wrap(ZIN0(7), 0.f, 1.f) - 1.0;
+    const double initPhase = unit->mPhase = 4.0 * sc_wrap(ZIN0(7), 0.f, 1.f) - 1.0;
 
     RGen& rgen = *unit->mParent->mRGen;
     float rate = ZIN0(1) * unit->mFreqMul;
     float depth = ZIN0(2);
     float rateVariation = ZIN0(5);
     float depthVariation = ZIN0(6);
-    unit->mFreq = rate * (1.f + rateVariation * rgen.frand2());
-    unit->m_scaleA = depth * (1.f + depthVariation * rgen.frand2());
-    unit->m_scaleB = depth * (1.f + depthVariation * rgen.frand2());
+    float initFreq = unit->mFreq = rate * (1.f + rateVariation * rgen.frand2());
+    float initScaleA = unit->m_scaleA = depth * (1.f + depthVariation * rgen.frand2());
+    float initScaleB = unit->m_scaleB = depth * (1.f + depthVariation * rgen.frand2());
     unit->m_delay = (int)(ZIN0(3) * SAMPLERATE);
     unit->m_attack = (int)(ZIN0(4) * SAMPLERATE);
     unit->m_attackSlope = 1. / (double)(1 + unit->m_attack);
     unit->m_attackLevel = unit->m_attackSlope;
     unit->trig = 0.0f;
     SETCALC(Vibrato_next);
+
     Vibrato_next(unit, 1);
+
+    unit->mPhase = initPhase;
+    unit->mFreq = initFreq;
+    unit->m_scaleA = initScaleA;
+    unit->m_scaleB = initScaleB;
+    unit->trig = 0.0f;
 }
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -1527,7 +1527,7 @@ void XLine_Ctor(XLine* unit) {
         ZOUT0(0) = start;
         unit->mCounter = counter;
         unit->mGrowth = pow(end / start, 1.0 / counter);
-        unit->mLevel = start * unit->mGrowth;
+        unit->mLevel = start;
     }
 }
 
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -635,9 +635,11 @@ void LFCub_Ctor(LFCub* unit) {
         SETCALC(LFCub_next_k);
 
     unit->mFreqMul = 2.0 * unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1) + 0.5;
+    double initPhase = unit->mPhase = ZIN0(1) + 0.5;
 
     LFCub_next_k(unit, 1);
+
+    unit->mPhase = initPhase;
 }
 
 
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -645,6 +645,139 @@ TestCoreUGens : UnitTest {
 
 	}
 
+	test_line_startAndEndSampleAreOuputOverDur {
+		var durFrames = 5; // duration for the line to complete its trajectory, in frames
+		var renderFrames = durFrames + 1; // render one extra frame to check that the end value is reached
+		var startVal = 2;
+		var endVal = 20;
+		var tolerance = -100.dbamp;
+		var condvar = CondVar();
+
+		server.bootSync;
+
+		[\ar, \kr].do{ |rate|
+			[Line, XLine].do{ |ugen|
+				var fs = server.sampleRate;
+				var rateScale = switch(rate, \ar, { 1.0 }, \kr, { server.options.blockSize });
+				var lineDur = durFrames * rateScale / fs;
+				var testDur = renderFrames * rateScale / fs;
+
+				{
+					ugen.perform(rate, startVal, endVal, lineDur)
+				}.loadToFloatArray(testDur,
+					action: { |data|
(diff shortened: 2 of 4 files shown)


dyfer:
Thanks!

> * The following Unit tests were made to pass with these changes

In that case, could you re-enable these in the test suite by [removing this](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/test_run_proto_gha.scd?plain=1#L18-L23)?

mtmccrea:
> Note there are failing tests for Duty in 'TestCoreUGens'

Unless I'm mistaken because there are still other (unrelated) failing tests in 'TestCoreUgens', we wouldn't want to remove this from CI yet.

dyfer:
Ah I see, thanks for pointing it out. In that case let's not include back `TestCoreUGens` at this time.

mossheim:
changes lgtm thanks! is there any plan for further testing and how should we integrate these? is it ok to merge piecemeal or do you want to wait until all have been reviewed?

mtmccrea:
Thanks, @mossheim.

> is there any plan for further testing

I try to do tests with a variety of argument rates, etc. for each UGen:
https://github.com/mtmccrea/supercollider/wiki/UGen-Fix-List#notes-on-testing-technique

So I'd consider my testing done, and my hope has been that these spend some time in circulation in the dev branch for more exposure.

>how should we integrate these? is it ok to merge piecemeal or do you want to wait until all have been reviewed?

I try to batch these into sensible clusters for each PR so they can be be reviewed and merged individually, otherwise it would run on for years, literally 😆

For UGens that are particularly complex, if the changes are beyond fixing initialization problems, or if changes may be contentious for other reasons, I try to make them separate PRs (E.g. Delay1/2, EnvGen, etc.)

mossheim:
> I try to do tests with a variety of argument rates, etc. for each UGen:

Are these in the test suite or purely manual testing?

> I try to batch these into sensible clusters for each PR so they can be be reviewed and merged individually, otherwise it would run on for years, literally

OK, so it is not a problem if we release a version with fixes to some UGens but not all?

mtmccrea:
The tests are manual. I have a template, but unfortunately each one requires manual setup specific to each UGen.

>OK, so it is not a problem if we release a version with fixes to some UGens but not all?

Correct, each PR is standalone. I try to avoid cross-dependencies in general, but I'll note them in the description if there are (this one doesn't have any).

mtmccrea:
> What do you mean? Are you instrumenting the C++ code somehow or using additional tooling?

My template is in SC, it generates multiple function calls for the UGen under test, with varying i/o and arg rates, then plots the output over some useful number of cycles. Then I inspect plots. By "manual" I mean I have to tune each UGen test to set argument ranges and input signals that are appropriate to the UGen.

I also have to insert print statements in the C++ code in order to inspect the init sample, since this can't be inspected from SC (it's never written out), and compare it to the first output sample.

mtmccrea:
Pinging on this batch of initialization fixes. 
It would be great to have this review completed— these changes fix a number of unit tests that have been broken for a loooong time. And further work on fixing other UGen initializations depends on these UGens behaving properly.

telephon (commented):
I'm not sure if it would help to add a few more tests? Since this changes subtle behaviour it would be good if one could understand (and fix) these changes through the boundary cases.

mtmccrea:
> I'm not sure if it would help to add a few more tests? Since this changes subtle behaviour it would be good if one could understand (and fix) these changes through the boundary cases.

Sure, currently working on adding unit tests now... I suppose there's a balance to be struck for the larger project of fixing UGen initialization bugs. If we require unit tests for each fix, it adds quite a bit of overhead to a project that's already a bit of a slog. I don't disagree it's useful of course...

JordanHendersonMusic:
@mtmccrea I *think* its the existing unit tests that check for initialisation that are often flaky when ran on through CI. Might want to be extra careful how they are written.

mtmccrea:
> @mtmccrea I _think_ its the existing unit tests that check for initialisation that are often flaky when ran on through CI. Might want to be extra careful how they are written.

I'm not sure what kind of initialization you're referring to, but this pertains to the "initialization sample" problem, which can't actually be checked directly because that sample is never actually written out—instead the approach is to check the proper phase of the signal, which will be off if the UGen state isn't reset in the course of graph initialization.

In any case the tests are working so far, just struggling to try reconfiguring `clang-format` after updating my OS and XCode recently. after running `./format.py develop` I'm getting the error:
```shell
*** ERROR:
clang-format found, but incorrect version at /opt/homebrew/Cellar/clang-format@8/8.0.1_1/bin/clang-format-8 with version: clang-format version 8.0.1 (tags/RELEASE_801/final)

Accepted versions: 14.y.z
```
So it seems to be expecting version 14 now :-/

And btw earlier when trying to reinstall  `clang-format@8` with homebrew, I get
```shell
Error: clang-format@8 has been disabled because it is a versioned formula!
```

I love our formatting setup!

mtmccrea:
OK I managed to get formatting working (and made a [janky update](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions#requirements) to the wiki... which should be more properly updated).

mtmccrea:
Since this PR fixed a handful of unit tests in `TestCoreUGens: test_ugen_generator_equivalences`, I've re-enabled that suite. Theres now just 2 unit tests in that suite that fail—`Duty` tests—so I commented those 2 out so the other 110 tests can run.

So it looks like this CI should pass, so I think this is finally good to go.

mtmccrea:
@JordanHendersonMusic could you confirm I address your concern with the unit tests? Otherwise I think I've addressed everything in the comments/reviews and this should be ready to go.

JordanHendersonMusic:
Oh yea, I was wrong and thought it was failing for a different reason!

mtmccrea:
would anyone like to give this an "official" review?

server/plugins/LFUGens.cpp
```diff
@@ -402,23 +399,30 @@ void Vibrato_next(Vibrato* unit, int inNumSamples) {
 
 void Vibrato_Ctor(Vibrato* unit) {
     unit->mFreqMul = 4.0 * SAMPLEDUR;
-    unit->mPhase = 4.0 * sc_wrap(ZIN0(7), 0.f, 1.f) - 1.0;
+    double initPhase = unit->mPhase = 4.0 * sc_wrap(ZIN0(7), 0.f, 1.f) - 1.0;
```
JordanHendersonMusic, line 402: Since these aren't mutated, its good practice to declare these variables `const double` https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable

This happens throughout.

I mean, I've never seen actual code guidelines for supercollider, so I guess you could ignore this?

server/plugins/LFUGens.cpp
```diff
@@ -664,9 +676,12 @@ void LFTri_Ctor(LFTri* unit) {
     }
 
     unit->mFreqMul = 4.0 * unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1);
+
+    double initPhase = unit->mPhase = sc_wrap((double)ZIN0(1), 0.0, 4.0);
```
JordanHendersonMusic, line 680: Can you make this a static_cast<double> please rather than a c cast.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts

testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -645,6 +646,155 @@ TestCoreUGens : UnitTest {
...
+		var durFrames = 5; // duration for the line to complete its trajectory, in frames
+		var renderFrames = durFrames + 1; // render one extra frame to check that the end value is reached
+		var startVal = 2;
+		var endVal = 20;
+		var tolerance = -100.dbamp;
+		var condvar = CondVar();
+
+		server.bootSync;
+
+		[\ar, \kr].do{ |rate|
```
JordanHendersonMusic, line 659: This might be easier to write as `[[\ar, \kr], [Line, XLine]].flop.do` to remove a level of indent. Just an idea, feel free to ignore.

testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -645,6 +646,155 @@ TestCoreUGens : UnitTest {
...
+					action: { |data|
+
+						this.assertFloatEquals(data.first, startVal,
+							"%.%'s first output sample should be its given start value.".format(ugen, rate), within: tolerance);
+						this.assert(data.last == endVal and: { data[data.size-2] != endVal },
+							"%.% should not reach its end value before the given duration".format(ugen, rate));
+						// Note an exception to the previouse test would be if the duration of the Line is less than a frame duration.
+						// In that (odd) case the "correct" first value is the end value. See GH issue #4279.
+						condvar.signalOne;
+					}
```
JordanHendersonMusic, line 676: This whole bit is hard to read. Perhaps define the action function in a variable and format it ...
```
{ ugen.perform(rate, startVal, endVal, (durFrames * rateScale / fs)) }
   .loadToFloatArray(renderFrames * rateScale / fs, server, action: actionFunc);
```

testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -645,6 +646,155 @@ TestCoreUGens : UnitTest {
...
+
+			// LFPar: test initial phase
+			// [initPhase, value] pairs, LFPar phase is 0->4
+			phs_val = [
+				0, 1.0,
+				1, 0.0,
+				2, -1.0,
+				3, 0.0,
+				4, 1.0
+			].clump(2).do{ |pvPair|
```
JordanHendersonMusic, line 735: Since you've already formatted it in pairs, its probably easier to just put the braces in per line rather than call clump.

JordanHendersonMusic (changes requested):
This looks really good fixes many failing tests on my system, thanks! 

I've got no issue with the approach.

The comments here are about C++ guidelines and sc formatting and code clarity (except the static cast which should definitely be used instead). If you don't agree with them, I'm quite happy to just approve as is however!

server/plugins/LFUGens.cpp
mtmccrea: I agree in principle, though I think the prevailing approach in the cpp codebase, when making small local changes, is to stick with the surrounding convention, making changes that are only pertinent to the functional changes (e.g. not adding `const` to all the variables in this local function).

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: > [[\ar, \kr], [Line, XLine]].flop.do

This would actually interleave the vars so both rates aren't accessible from each UGen. And I think this separates the logic better (for each rate, iterate through each ugen...), and theoretically adding UGens to the list would be trivial, but wouldn't work in the flopping scheme.

server/plugins/LFUGens.cpp
JordanHendersonMusic: Sure, but in this case, it's a new variable that wasn't there before. Adding const would only change your code.

I don't think following the style of the surrounding code is good practice when it's harder to read, uses unsafe or discouraged features. Following the surrounding code over the recommendations of the language's creator seems problematic when applied in general.

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: I've refactored a bunch of the tests so hopefully it's more compact while still being easy to read.

In this particular case I opted to keep the `action` function explicit an in-place as with other tests. I think it flows more logically for the DSP to be defined first, then the test follows. But hopefully the refactor addresses some of the readability issue you mentioned.

server/plugins/LFUGens.cpp
mtmccrea: I don't disagree, again, it's really a matter of convention. Is this code meaningfully more safe?
```
void Vibrato_Ctor(Vibrato* unit) {
    unit->mFreqMul = 4.0 * SAMPLEDUR;
    double initPhase = unit->mPhase = 4.0 * sc_wrap(ZIN0(7), 0.f, 1.f) - 1.0;

    RGen& rgen = *unit->mParent->mRGen;
    float rate = ZIN0(1) * unit->mFreqMul;
    float depth = ZIN0(2);
    float rateVariation = ZIN0(5);
    float depthVariation = ZIN0(6);
    float initFreq = unit->mFreq = rate * (1.f + rateVariation * rgen.frand2());
    float initScaleA = unit->m_scaleA = depth * (1.f + depthVariation * rgen.frand2());
    float initScaleB = unit->m_scaleB = depth * (1.f + depthVariation * rgen.frand2());
    const float someNewVar = 3.f;

    unit->m_delay = (int)(ZIN0(3) * SAMPLERATE);
    unit->m_attack = (int)(ZIN0(4) * SAMPLERATE);
    unit->m_attackSlope = 1. / (double)(1 + unit->m_attack);
    unit->m_attackLevel = unit->m_attackSlope;
    unit->trig = 0.0f;
    SETCALC(Vibrato_next);
    Vibrato_next(unit, 1);

    unit->mPhase = initPhase;
    unit->mFreq = initFreq;
    unit->m_scaleA = initScaleA;
    unit->m_scaleB = initScaleB;
    unit->trig = 0.0f;
}
```

testsuite/classlibrary/TestCoreUGens.sc
JordanHendersonMusic: Yea this looks much better and depends on less methods and classes than before, testing only what's intended.

server/plugins/LFUGens.cpp
JordanHendersonMusic: Yea I am, as I mentioned at the beginning it's only really the c cast that must be changed. It's fine as is.

server/plugins/LFUGens.cpp
mtmccrea: done!

server/plugins/LFUGens.cpp
mtmccrea: I've pushed the updated version with the const declaration, maybe we start a trend going forward :)

JordanHendersonMusic (approved):

Issue #5948, opened 2023-01-16
Scheduling problems on Apple M1


Josemiguelfernandez:
## Environment

* SuperCollider version: 3.12.2, 3.13.0-rc-1 arm64 and x64
* Operating system: macOS Monterey 12.6.1 and Ventura

## Steps to reproduce

```supercollider
SynthDef(\imp, { OffsetOut.ar(0, Impulse.ar(0)); FreeSelf.kr(Impulse.kr(0)); }).add;

(
t = Task({
	10000.do({ arg i;
		Synth(\imp);
		0.05.wait;
		});
	});
)

t.start;
t.stop;
```

## Expected vs. actual behavior

I have done different tests sending messages from scland and from another OSC client (Antescofo) to scsynth and I find scheduling problems (rhythmic irregularity or interruptions). 
If I launch the Task process (example above Steps to reproduce) on a Mac M1 the rhythm generated by scland in scsynth has irregularities and is not completely stable. For example if MacOS starts to make an indexation (Spotligth), if I connect to the internet or switch from one application to another (command+tab), there are interruptions. It seems that every time the system does a background operation, there are interruptions sending or receiving OSC messages.
I have done several tests on 2 MacBook Pro M1 with Monterey (12.6.1) and a MacBook Air M1 with Ventura.
This problem does not happen on a MacBook Pro with Intel processor running SC 3.12.2.
I've tried with the x64 version on Rosetta and native ARM64 and the problem is the same...

Apparently there is a priority problem, but using the 'renice' command in the terminal and increasing the priority of scsynth and sclang process there seems to be a very small improvement, but it doesn't solve the problem.

Any idea how to solve this problem? To make it work like on an Intel processor without interruptions ?

Thanks,

José

Josemiguelfernandez:
I just launched the test using supernova and it works as expected without interruptions! It seems to be a problem with scsynth...

smrg-lm:
If you replace `Synth(\imp)` with `(instrument: \imp).play()` does it play fine in the server? It could be an OS performance difference but since `Synth(\imp)` doesn't guaranties timing, because no bundles are created, the code can sound very different depending on the load.

Josemiguelfernandez:
Hola Lucas, 
Thanks for the suggestion! I have done different tests with sclang and antescofo sending messages in different ways in udp and tcp, the example I have sent is minimal and can be done in several ways. I have also tried with patterns:
```
[...]

Issue #5624, opened 2021-11-26
Use audio workgroups for supernova DSP thread pool.


Spacechild1:
macOS 11 has a new API for managing audio threads that cooperate with each other. I think this would be appropriate for supernova's DSP thread pool.

Here's an introduction: https://developer.apple.com/documentation/audiotoolbox/workgroup_management/understanding_audio_workgroups

And here's a code example which might be relevant for our use case: https://developer.apple.com/documentation/audiotoolbox/workgroup_management/adding_parallel_real-time_threads_to_audio_workgroups

Should be rather simple to implement. I just don't have access to a macOS 11 machine.

scztt:
I looked at this last summer. Unfortunately, the new API's require access to specific CoreAudio objects in order to set the workgroup properties. This is a problem for two reasons:
1. `supernova` is built on top of PortAudio, and we have very little access to platform-specific objects and data via the PortAudio API (by design: it's an audio hardware abstraction). While we MIGHT be able to root around and find the right CoreAudio object we need on the SuperCollider side (though I couldn't immediately see how), this still probably belongs in PortAudio in the end anyway, since it IS supposed to abstract platform details from the client code.
2. It was not obvious how to get the right CoreAudio objects required for the workspace API's from the objects that PortAudio creates in it's MacOS implementation (see https://developer.apple.com/documentation/apple-silicon/porting-your-audio-code-to-apple-silicon#Prioritize-Realtime-Threads-Using-Workgroups). Probably there IS a way to do this, but it was not obvious and would likely require some substantial code change.

Sadly there is no mention of this on the PortAudio github, and indeed there are many years-old MacOS issues open in that repo, and it doesn't seem like the platform is very actively maintained.

Spacechild1:
Hi Scott, thanks for looking into this!

> supernova is built on top of PortAudio, and we have very little access to platform-specific objects and data via the PortAudio API (by design: it's an audio hardware abstraction).

FWIW, portaudio offers platform specific extension APIs, e.g. `pa_asio.h`, `pa_jack.h`, `pa_linux_alsa.h`, `pa_mac_core.h`, etc. (https://github.com/PortAudio/portaudio/tree/master/include)
In fact, `pa_win_wasapi.h` even has a method for accessing the underlying native objects:
https://github.com/PortAudio/portaudio/blob/5b5feb3a331b674cfc71009414514000685bf547/include/pa_win_wasapi.h#L354-L362
[...]

PR #6043, opened 2023-06-21
scsynth/supernova: fix thread priorities

Spacechild1:
On macOS, `nova::thread_priority_interval()` just returns dummy values (0), thus it would accidentally *lower* the thread priority. This can lead to particularly poor performance on recent Apple Silicon machines because the thread might get scheduled on efficiency cores. I removed any calls to `nova::thread_priority_interval`; instead we simply keep the default thread priority (for now).

On Windows, thread priorities are not a continuous range, so you cannot take the mean of the range return by `nova::thread_priority_interval_rt`. In fact, this would pass an invalid value to `SetThreadPriority` and essentially do nothing. Instead I simply use the lower value which effectively raises the thread priority a bit (`THREAD_PRIORITY_ABOVE_NORMAL`). This also matches the behavior of Supernova, see `io_thread_init_functor`: https://github.com/supercollider/supercollider/blob/ef627ce2c564fe323125234e4374c9c4b0fc7f1d/server/supernova/server/server.cpp#L294-L297

## Purpose and Motivation

Fixes https://github.com/supercollider/supercollider/issues/5948

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/scsynth/SC_ComPort.cpp
+++ b/server/scsynth/SC_ComPort.cpp
@@ -443,12 +443,10 @@ SC_TcpConnection::~SC_TcpConnection() { mParent->connectionDestroyed(); }
 //////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 static void asioFunction() {
+    /* NB: on macOS we just keep the default thread priority */
 #ifdef NOVA_TT_PRIORITY_RT
     std::pair<int, int> priorities = nova::thread_priority_interval_rt();
     nova::thread_set_priority_rt((priorities.first + priorities.second) / 2);
-#else
-    std::pair<int, int> priorities = nova::thread_priority_interval();
-    nova::thread_set_priority(priorities.second);
 #endif
 
     boost::asio::io_service::work work(ioService);
--- a/server/scsynth/SC_ComPort.cpp
+++ b/server/scsynth/SC_ComPort.cpp
@@ -445,8 +445,8 @@ SC_TcpConnection::~SC_TcpConnection() { mParent->connectionDestroyed(); }
 static void asioFunction() {
     /* NB: on macOS we just keep the default thread priority */
 #ifdef NOVA_TT_PRIORITY_RT
-    std::pair<int, int> priorities = nova::thread_priority_interval_rt();
-    nova::thread_set_priority_rt((priorities.first + priorities.second) / 2);
+    int priority = nova::thread_priority_interval_rt().first;
+    nova::thread_set_priority_rt(priority);
 #endif
 
     boost::asio::io_service::work work(ioService);
--- a/server/scsynth/SC_CoreAudio.cpp
+++ b/server/scsynth/SC_CoreAudio.cpp
@@ -107,13 +107,7 @@ static void syncOSCOffsetWithTimeOfDay() {
 }
 
 static void resyncThreadFunc() {
-#    ifdef NOVA_TT_PRIORITY_RT
-    std::pair<int, int> priorities = nova::thread_priority_interval_rt();
-    nova::thread_set_priority_rt((priorities.first + priorities.second) / 2);
-#    else
-    std::pair<int, int> priorities = nova::thread_priority_interval();
-    nova::thread_set_priority(priorities.second);
-#    endif
+    /* keep the default thread priority */
 
     while (true) {
         sleep(20);
@@ -345,12 +339,10 @@ SC_AudioDriver::~SC_AudioDriver() {
 }
 
 void SC_AudioDriver::RunThread() {
+    /* NB: on macOS we just keep the default thread priority */
 #ifdef NOVA_TT_PRIORITY_RT
     std::pair<int, int> priorities = nova::thread_priority_interval_rt();
     nova::thread_set_priority_rt((priorities.first + priorities.second) / 2);
-#else
-    std::pair<int, int> priorities = nova::thread_priority_interval();
-    nova::thread_set_priority(priorities.second);
 #endif
 
     TriggersFifo* trigfifo = &mWorld->hw->mTriggers;
--- a/server/scsynth/SC_CoreAudio.cpp
+++ b/server/scsynth/SC_CoreAudio.cpp
@@ -341,8 +341,8 @@ SC_AudioDriver::~SC_AudioDriver() {
 void SC_AudioDriver::RunThread() {
     /* NB: on macOS we just keep the default thread priority */
 #ifdef NOVA_TT_PRIORITY_RT
-    std::pair<int, int> priorities = nova::thread_priority_interval_rt();
-    nova::thread_set_priority_rt((priorities.first + priorities.second) / 2);
+    int priority = nova::thread_priority_interval_rt().first;
+    nova::thread_set_priority_rt(priority);
 #endif
 
     TriggersFifo* trigfifo = &mWorld->hw->mTriggers;
--- a/server/supernova/server/server.cpp
+++ b/server/supernova/server/server.cpp
@@ -292,12 +292,10 @@ void thread_init_functor::operator()(int thread_index) {
 }
 
 void io_thread_init_functor::operator()() const {
+    /* NB: on macOS we just keep the default thread priority */
 #ifdef NOVA_TT_PRIORITY_RT
     int priority = thread_priority_interval_rt().first;
     thread_set_priority_rt(priority);
-#else
-    int priority = thread_priority_interval().second;
-    thread_set_priority(priority);
 #endif
 
     name_thread("Network Send");
--- a/server/supernova/utilities/osc_server.hpp
+++ b/server/supernova/utilities/osc_server.hpp
@@ -49,6 +49,7 @@ class network_thread {
 public:
     void start_receive(void) {
         thread_ = std::thread([this] {
+        /* NB: on macOS we just keep the default thread priority */
 #ifdef NOVA_TT_PRIORITY_RT
             thread_set_priority_rt(thread_priority_interval_rt().first);
 #endif


Spacechild1:
@Josemiguelfernandez Can you give it try?

Spacechild1:
@dyfer For some reason lots of tests seem to fail on Linux, but this PR shouldn't affect any of these... Is this a known issue?

dyfer:
> Is this a known issue?

Yes, sorry it's not well documented. I added this to the list in #5360.

The `TestThreadReschedule:test_Threads_rescheduleToOtherClock` _sometimes_ fails, but when it does, it derails the interpreter and thus the rest of the testsuite. I'll just restart that job.

Spacechild1:
@dyfer Thanks for the heads-up! I just I discovered another thread priority bug, this time on Windows.

dyfer:
Thanks so much for working on this!
For the macOS changes, wouldn't it be better to address this in the `nova-tt` module? Or is that too cumbersome/not feasible at this time?

server/scsynth/SC_ComPort.cpp
```diff
@@ -443,12 +443,10 @@ SC_TcpConnection::~SC_TcpConnection() { mParent->connectionDestroyed(); }
...
 static void asioFunction() {
+    /* NB: on macOS we just keep the default thread priority */
 #ifdef NOVA_TT_PRIORITY_RT
-    std::pair<int, int> priorities = nova::thread_priority_interval_rt();
-    nova::thread_set_priority_rt((priorities.first + priorities.second) / 2);
-#else
-    std::pair<int, int> priorities = nova::thread_priority_interval();
-    nova::thread_set_priority(priorities.second);
+    int priority = nova::thread_priority_interval_rt().first;
+    nova::thread_set_priority_rt(priority);
```
dyfer, line 449: Should this apply to all platforms?

Spacechild1: `NOVA_TT_PRIORITY_RT` is defined for Windows and Linux.

dyfer: So this change also changes behavior on Linux, is that correct?

Spacechild1:
> For the macOS changes, wouldn't it be better to address this in the nova-tt module? Or is that too cumbersome/not feasible at this time?

I need to make a PR to `nova-tt` anyway because of https://github.com/supercollider/supercollider/pull/5618. (I finally have time to do this.) In the course of this I could do some research about thread priorities on macOS and fix `nova::thread_priority_interval`. Actually, I'm not even sure if the implementation of `nova::thread_set_priority` is really correct for macOS...

On the other hand, the default thread priority is probably fine and it would at least be an improvement. But I want to wait for @Josemiguelfernandez to test, as he is the one who raised the issue. We can always try to fix `nova-tt` later. This PR is more intended as a quick fix.

---

Generally, I have the feeling that we could add helper functions for raising/lowering thread priorities, using some descriptive enums, and then let these functions do the dirty platform-specific work with `nova-tt`.

dyfer:
> Generally, I have the feeling that we could add helper functions for raising/lowering thread priorities, using some descriptive enums, and then let these functions do the dirty platform-specific work

I don't disagree, but maybe this was the idea behind having nova-tt in the first place? I can only guess, as I'm neither the author, nor was I active in the development when these changes happened.

Spacechild1:
> I don't disagree, but maybe this was the idea behind having nova-tt in the first place?

In theory, yes. However, it is quite low-level and it doesn't really map nicely to Windows. Eventually, we need some `#ifdefs` to properly handle every OS. Instead of repeating these all the time, it might be better to express the intent (e.g. `sc::setThreadPriority(sc::ThreadPriority::High)`) and let that function handle the gritty details.

dyfer:
I agree, my (limited) intuition is that it'd be a good idea to abstract this in SC codebase.

server/scsynth/SC_ComPort.cpp
Spacechild1: Yes, it slightly changes the value passed to `nova::thread_priority_interval_rt`, but as I mentioned in the PR description, this is in line with Supernova.

dyfer:
BTW I can confirm that functionally this PR improves scheduling's susceptibility to system load _tremendously_ on macOS 12 with M1 CPU.

Spacechild1:
Thanks for testing! (I don't have a M1 machine.) Maybe I will be able to attend the next dev meeting where we could discuss how to proceed with this.

server/scsynth/SC_ComPort.cpp
dyfer: Ah, so "in line with Supernova" also applies to Linux, thanks for clarifying.

For a good measure, I'd like to test the performance on Windows and Linux before pulling it in. (IMO it's all good on macOS)

Josemiguelfernandez:
> Thanks for testing! (I don't have a M1 machine.) Maybe I will be able to attend the next dev meeting where we could discuss how to proceed with this.



> > For the macOS changes, wouldn't it be better to address this in the nova-tt module? Or is that too cumbersome/not feasible at this time?
> 
> I need to make a PR to `nova-tt` anyway because of #5618. (I finally have time to do this.) In the course of this I could do some research about thread priorities on macOS and fix `nova::thread_priority_interval`. Actually, I'm not even sure if the macOS implementation of `nova::thread_set_priority` is really correct...
> 
> On the other hand, the default thread priority is probably fine and it would at least be an improvement. But I want to wait for @Josemiguelfernandez to test, as he is the one who raised the issue. (The code should be the same as in [#5948 (comment)](https://github.com/supercollider/supercollider/issues/5948#issuecomment-1413989094), but I just want to make sure that I didn't screw up.)
> 
> Generally, I have the feeling that we could add helper functions for raising/lowering thread priorities, using some descriptive enums, and then let these functions do the dirty platform-specific work with `nova-tt`. This would remove quite a bit of code duplication. It would also allow us to have more elaborate comments.

Hi @Spacechild1 !

Thanks for fixing this problem!
Right now I don't have access to my computer, I'm on vacation until the end of the month :-)
It would be interesting to experiment with a higher priority on MacOs and see if scsynth is more reactive than with the default priority. I've tried but I couldn't, I don't quite understand how priorities work in MacOs...

thank you so much !!

timblechmann:
on macos one should probably not rely on mach's time constraint threads these days, but use audio workgroups instead.

https://developer.apple.com/documentation/audiotoolbox/workgroup_management/adding_parallel_real-time_threads_to_audio_workgroups?language=objc

Spacechild1:
> on macos one should probably not rely on mach's time constraint threads these days, but use audio workgroups instead.
> 
> https://developer.apple.com/documentation/audiotoolbox/workgroup_management/adding_parallel_real-time_threads_to_audio_workgroups?language=objc

FWIW, I have already opened an issue for that :) https://github.com/supercollider/supercollider/issues/5624

dyfer:
> on macos one should probably not rely on mach's time constraint threads these days, but use audio workgroups instead.

Just to clarify, are you suggesting for audio workgroups to be also considered for the network thread? IIUC the main point of this PR is to address the timing inaccuracies that are caused by the lower priority, not the audio performance itself.

Spacechild1:
I think @timblechmann only meant the audio threads (which currently use time constraint based scheduling), just as another thing that could be improved. However, we'd need a separate PR for this and further discussion should go to https://github.com/supercollider/supercollider/issues/5624.

Josemiguelfernandez:
> > Thanks for testing! (I don't have a M1 machine.) Maybe I will be able to attend the next dev meeting where we could discuss how to proceed with this.
> 
> > > For the macOS changes, wouldn't it be better to address this in the nova-tt module? Or is that too cumbersome/not feasible at this time?
> > 
> > 
> > I need to make a PR to `nova-tt` anyway because of #5618. (I finally have time to do this.) In the course of this I could do some research about thread priorities on macOS and fix `nova::thread_priority_interval`. Actually, I'm not even sure if the macOS implementation of `nova::thread_set_priority` is really correct...
> > On the other hand, the default thread priority is probably fine and it would at least be an improvement. But I want to wait for @Josemiguelfernandez to test, as he is the one who raised the issue. (The code should be the same as in [#5948 (comment)](https://github.com/supercollider/supercollider/issues/5948#issuecomment-1413989094), but I just want to make sure that I didn't screw up.)
> > Generally, I have the feeling that we could add helper functions for raising/lowering thread priorities, using some descriptive enums, and then let these functions do the dirty platform-specific work with `nova-tt`. This would remove quite a bit of code duplication. It would also allow us to have more elaborate comments.
> 
> Hi @Spacechild1 !
> 
> Thanks for fixing this problem! Right now I don't have access to my computer, I'm on vacation until the end of the month :-) It would be interesting to experiment with a higher priority on MacOs and see if scsynth is more reactive than with the default priority. I've tried but I couldn't, I don't quite understand how priorities work in MacOs...
> 
> thank you so much !!

Hi @Spacechild1, @dyfer,

I just tried this PR and it works as expected. I will still try if it is possible to increase the priority on macOs and see if it can be done better! waiting also for improvements in the audio part in scsynth and supernova #5624 !

Thanks !

Spacechild1:
@dyfer I think this is ready to merge, right? It has been successfully tested on Windows and macOS. (It doesn't really affect Linux.)

dyfer:
I thought it would be better to address this in nova-tt, tbh. Otherwise this is a WIP change (though an improvement nonetheless).
But still, once this is fixed in nova-tt, the code for macos would need to be re-added... Would you consider making a PR to nova-tt? Or is that not possible atm?

Spacechild1:
> I thought it would be better to address this in nova-tt, tbh

I think the consensus was rather to add high-level functions and only use nova-tt under the hood where applicable. See https://github.com/supercollider/supercollider/pull/6043#issuecomment-1601691048

I agree that the whole thing should be refactored, but this is out of scope for this PR (at least for me) and can always be done later. If you want to wait for the "proper" fix, that's fine for me (I am not a Mac user :), but then we might have a situation like in https://github.com/supercollider/supercollider/pull/5618.

dyfer (approved):
All right, let's pull this in. Thanks!

PR #6044, opened 2023-06-21
GHA: update ccache on Windows

dyfer:
## Purpose and Motivation

`ccache` was updated in chocolatey past version 4.8, which did not work properly for MSVC, so we can un-pin the version now.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- ~~Updated documentation~~ n/a

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -487,7 +487,7 @@ jobs:
       - name: install ccache
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


dyfer:
At the time of submitting the PR the MSVC build time is about 10 minutes, so _I think_ the cache works... (?)

PR #6045, opened 2023-06-22
Topic/integer xrand xrand2 overflow

mjsyts:
## Purpose and Motivation

Fixed bug in xrand/xrand2 methods to avoid integer overflow when using large integers. Example: 2147483647.xrand2 currently returns either 2147483647 or -2147483648 due to overflow. Adding 1 when calling .rand on an integer will result in the same integer overflow bug.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Math/Integer.sc
+++ b/SCClassLibrary/Common/Math/Integer.sc
@@ -19,11 +19,13 @@ Integer : SimpleNumber {
 		if ( res == exclude ) { res = this };
 		^res;
 	}
+
 	xrand2 { arg exclude=0;
 		var res = this.rand2;
 		if ( res == exclude ) { res = [this, this.neg].choose };
 		^res;
 	}
+
 	degreeToKey { arg scale, stepsPerOctave = 12;
 		^scale.performDegreeToKey(this, stepsPerOctave)
 	}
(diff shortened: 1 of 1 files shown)


SCClassLibrary/Common/Math/Integer.sc
```diff
@@ -15,13 +15,17 @@ Integer : SimpleNumber {
...
+		^res;
 	}
+
 	xrand2 { arg exclude=0;
-		var res;
-		res = (2 * this).rand - this;
-		if (res == exclude, { ^this },{ ^res });
+		var res = this.rand2;
+		if ( res == exclude ) { res = [this, this.neg].choose };
+		^res;
```
mossheim, line 26: this algorithm unfortunately doesn't produce an even distribution; you can use `Collection:-plotHisto` to quickly check if your output makes sense. Note that `rand2`'s range is `[-this, this]` (inclusive) while `rand` is `[0, this)` (exclusive).

elgiano: how about this version? it doesn't overflow and seems to produce an uniform distribution

```supercollider
        var res = this.rand2;
	// avoid infinite loop: -1.rand2 and 0.rand2 only produce 0s
	if ([0,-1].includes(x)) { ^0 };
	while { res  == exclude } { res = x.rand2 };
	res
```

Issue #6048, opened 2023-06-26
Discrepancy with Numbers: Different Outputs on Linux (correct) and MacOS (bug)


smoge:
Hello guys,

I have come across an intriguing issue while using my `Rational` Quark (https://github.com/smoge/Rational) type for fraction arithmetic. 

For example, trying to multiply two `Rational` numbers, I observed that the output differs between Linux and MacOS. 

Here is the snippet of the code:

```
Rational(3,7) * Rational(9,5) 
// or alternatively
(3 %/ 7) * (9 %/ 5)
```

On my Linux machine (recent development build), this produces the expected result:

```
--> 27 %/ 35
```

However, on MacOS  (recent development build and stable official release), the output has changed. It used to be the same as on Linux when I wrote it, but now it displays something different:

```
--> 3e+01 %/ 4e+01
```

I am wondering if anyone else has encountered this discrepancy and if there is any insight into why this might be happening.

smoge:
The discrepancy arises because the .asStringPrec(precision) function is being used to remove decimal numbers when posting, given that the denominator and numerator are represented as 64-bit floats. I'm intrigued as to why this change was implemented and why it seems to have no impact on Linux. Can someone provide some insights or explanations?

smoge:
https://github.com/supercollider/supercollider/pull/4425

dyfer:
Thanks for the report @smoge 
Could you change the original example to use just the core class library?
Also, are you sure that the change in behavior occurred in the PR you posted?

smoge:
Just changed the example. 

 I suspect this change because Git tells us it happened between the time I wrote the code (when I remember there was no problem) and now. It was probably not really tested at the time and might have gone unnoticed since it's not really a common method.

smoge:
I think I found the problem. `%.<precision>g` is used for formatting floating-point numbers, but it does not work (as I understand, at least)  if the goal is to control the number of decimal places.  It prints the number in either fixed-point or exponential notation (if I'm not mistaken). The precision part isn’t just about decimals. It's like "How many digits should I squeeze in here?" (Right?) So, 17.1 with precision 1 gets all sci-fi and turns into 2e+01.

I tried to swap the ‘g’ for an ‘f’, and it seems to work.

```
diff --git a/lang/LangPrimSource/PyrPrimitive.cpp b/lang/LangPrimSource/PyrPrimitive.cpp
index 28ff8a78d..14cb65f46 100644
[...]

PR #4425, opened 2019-06-01
Clang format develop

brianlheim:
**Please read carefully.**

This is the PR to format develop. I have triple-checked everything here, including the .clang-format file itself, but please if you have time, check it again.

**This PR should be merged via rebase, not merge commit.**

The next steps should be:

(step zero, review this PR and 3.10 PR)

1. Merge this PR **via rebase**.
2. Merge the 3.10 PR **via rebase**.
3. Merge 3.10 into develop.
4. Create two new tags, `tag-clang-format-3.10` on the clang-format commit for 3.10, and `tag-clang-format-develop` on the clang-format commit for develop. Push them here. Delete any releases from CI.

Questions/comments welcome. I would like to merge this by the end of the day.

PR #6050, opened 2023-06-29
fix asStringPrec method

smoge:
Related to issue #6048

@dyfer

This PR addresses a formatting issue with floating-point numbers. The existing code uses the` %.g format` specifier, which isn't quite what we need if we want to have fine control over the number of decimal places. The thing with %.g is that it switches between fixed-point or exponential notation based on which is shorter, and the precision value specifies the maximum number of significant digits, rather than just decimal places.

Let’s take an example. Before the change, formatting 17.1 with a precision of 1 turned it into scientific notation, like this:

```
17.1.asStringPrec(1) // -> 2e+01
11.0.asStringPrec(0) // -> 1e+01

```
With this PR's changes, the output is more intuitive (and is closer to the behavior prior to #4425):

```
17.1.asStringPrec(1) // -> 17.1
11.0.asStringPrec(0) // -> 11.0
```

Now, there's a catch. I noticed that, ideally, `11.0.asStringPrec(0) `should return 11 (without the .0), as was before and is presently on Linux. Although it's not a mirror of C++, we might want to discuss if mirroring C++ is essential or if this alternative makes more sense.

--

Also, I’d like to propose considering an alternative approach. Instead of:

```
sprintf(fmt, "%%.%df", precision);
sprintf(str, fmt, slotRawFloat(a));
```

We could use the safer `snprintf`:

`snprintf(str, sizeof(str), "%.*f", precision, slotRawFloat(a));`

Looking forward to your feedback and thoughts on this!

Cheers!

### DIFF
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -455,7 +455,7 @@ int prFloat_AsStringPrec(struct VMGlobals* g, int numArgsPushed) {
     if (precision >= 200)
         precision = 200; // Nothing is that big anyway. And we know we will be smaller than our 256 char string
 
-    sprintf(fmt, "%%.%dg", precision);
+    sprintf(fmt, "%%.%df", precision);
     sprintf(str, fmt, slotRawFloat(a));
 
     PyrString* string = newPyrString(g->gc, str, 0, true);


smoge:
The checks seem to be performing more favorably on my fork, for some odd reason:

image

smoge:
No problem with checks on the fork (same code and tests): https://github.com/smoge/supercollider/actions/runs/5408876860/jobs/9832796423

PR #6051, opened 2023-06-30
added checkInputs method to ToggleFF

### LINKED ISSUE #5917: ToggleFF.ar(0) returns erratic output
## Steps to reproduce

```supercollider
{ ToggleFF.ar(0).poll; 0.0 }.play // incorrct, toggles all the time
{ ToggleFF.kr(0).poll; 0.0 }.play // correct
{ ToggleFF.ar(Impulse.ar(0)).poll; 0.0 }.play // correct
{ ToggleFF.kr(Impulse.kr(0)).poll; 0.0 }.play // correct

{ SetResetFF.ar(0, 0).poll; 0.0 }.play // ERROR: SetResetFF first input is not audio rate:  0
{ SetResetFF.kr(0, 0).poll; 0.0 }.play // // ERROR: SetResetFF first input is not control rate:  0
{ SetResetFF.ar(DC.ar(0), DC.ar(0)).poll; 0.0 }.play // 0, correct
{ SetResetFF.kr(DC.kr(0), DC.kr(0)).poll; 0.0 }.play  // 0, correct
{ SetResetFF.ar(Impulse.ar(0)).poll; 0.0 }.play // correct
{ SetResetFF.kr(Impulse.kr(0)).poll; 0.0 }.play // correct
{ SetResetFF.ar(Impulse.ar(0), Impulse.ar(0)).poll; 0.0 }.play // race condition. correct
{ SetResetFF.kr(Impulse.kr(0), Impulse.kr(0)).poll; 0.0 }.play // race condition.  correct
```

`ToggleFF` should throw an error like `SetResetFF`

cdbzb:
## Purpose and Motivation

Fixes #5917

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Audio/Trig.sc
+++ b/SCClassLibrary/Common/Audio/Trig.sc
@@ -131,17 +131,15 @@ SetResetFF : PulseCount {
 }
 
 ToggleFF : UGen {
-
 	*ar { arg trig = 0.0;
 		^this.multiNew('audio', trig)
 	}
 	*kr { arg trig = 0.0;
 		^this.multiNew('control', trig)
 	}
-
+	checkInputs { ^this.checkSameRateAsFirstInput }
 	signalRange { ^\unipolar }
 }
-
 
 ZeroCrossing : UGen {
 	*ar { arg in = 0.0;


dyfer:
Thanks!
@telephon @cdbzb just to double check:
Before this PR, the following would run (copied from the linked issue):
```supercollider
{ ToggleFF.ar(0).poll; 0.0 }.play // incorrct, toggles all the time
{ ToggleFF.kr(0).poll; 0.0 }.play // correct
```
After this PR both examples return `ERROR: ToggleFF first input is not [audio/control] rate`. I'm a little hesitant this will break users code, are we certain about this change?

JordanHendersonMusic:
Another option is to use the optimize ugen method to convert any scalar values into DC.

dyfer:
> Another option is to use the optimize ugen method to convert any scalar values into DC.

This sounds like a good idea to me.

elgiano:
I have been looking at the cpp implementation. It looks like ToggleFF doesn't handle well ar/kr/ir combinations and that's why it works as expected only when its input is at the same rate as the UGen.

I propose we fix it on scsynth side, instead of only on sclang, mainly because the UGen overflows in adjacent buffers, and we should fix this also for non-sclang clients. Here is a more in-depth explanation.

tl;dr: let's save current implementation as ToggleFFGlitch and add robust ar/kr logic to ToggleFF. We won't enforce same rates, won't need to wrap inputs in DC, and older synths won't stop working, but they'll loose a potential (fun) glitch.

----------------------------------

https://github.com/supercollider/supercollider/blob/f70e312dfcd9e078a8c07681e0cd3b866c5b0bf1/server/plugins/TriggerUGens.cpp#L897C1-L918C1
```cpp
void ToggleFF_next(ToggleFF* unit, int inNumSamples) {
    float* out = ZOUT(0);
    float* trig = ZIN(0);
    float level = unit->mLevel;


    LOOP1(inNumSamples, float curtrig = ZXP(trig); if (prevtrig <= 0.f && curtrig > 0.f) level = 1.f - level;
          ZXP(out) = level; prevtrig = curtrig;);
    unit->m_prevtrig = prevtrig;
    unit->mLevel = level;
}
```
This UGen has only one this logic for its calc function, so the only difference between audio and control rate is that in kr `inNumSamples=1`. That's why LOOP1 there works as expected only when both ToggleFF and trig are at the same rate:
- audio rate: reads `bufferSize` samples from trig, outputs `bufferSize` samples to out
- control rate (inNumSamples=1): reads 1 sample from trig and outputs 1 sample to out

When ToggleFF is audio rate and trig is kr/scalar, the UGen reads `bufferSize` samples from a 1-sample buffer, effectively overflowing into adjacent buffers, hence the "garbage" output.
Also, ToggleFF.kr is stable for kr/scalar inputs, but likely misses trigs in audio rate ones.

I think that the best fix going forward would be to save the current implementation as ToggleFFGlitch, and modify ToggleFF_next to handle ar/kr inputs correctly. The only combination that would be left out is ToggleFF.kr reading an ar input, I can't look into it right now but I guess there should be a way to make that work too using `fullBufferSize`.

cdbzb:
closed in favor of #6760

PR #6053, opened 2023-07-02
Topic/add check inputs to inRange

### LINKED ISSUE #5547: InRange/Clip/Wrap/Fold need to check audio rate
## Environment

* SuperCollider version: 3.12 (relevant code is stable for a long time)
* Operating system: doesn't matter

## Steps to reproduce

```supercollider
a = { Clip.ar(0, 0, 1) }.play;  // crash

a = { InRange.ar(0, 0, 1) * 0.05 }.play;  // buzzing at kr frequency
```

Crashes, by definition, should be fixed or prevented. (The crash for Clip, incidentally, is because Clip is the only one of these UGens with SIMD functions.)

Also, running an `ar` loop over a `kr` or scalar input, without interpolation, should be disallowed (because then the control block processes one valid sample and n-1 garbage samples).

The above cases can be fixed by adding the following to InRange in Trig.sc:

```
	checkInputs {
		if(rate == \audio) { ^this.checkSameRateAsFirstInput }
	}
```

... which would be an easy enough PR, except that there may be other similar cases. So... code review or not? I'll leave that for others to decide.

I'll label this as a class library issue, because I think it isn't the job of SIMD functions to check validity -- rather, it's the job of the language to ensure validity before generating a SIMD-optimized ar unit. Also, if we decide not to do a comprehensive code review, this would be an easy fix for a new contributor.

(Edit: Added one comment below, then deleted it as I had missed something... sorry for the extra email.)

cdbzb:
## Purpose and Motivation

fixes #5547

## Types of changes

- Bug fix

## To-do list

- [ x] Code is tested
- [ x] All tests are passing
- [ ] Updated documentation
- [ x] This PR is ready for review

### DIFF
--- a/HelpSource/Classes/UnitTest.schelp
+++ b/HelpSource/Classes/UnitTest.schelp
@@ -396,28 +396,30 @@ TestYourClass : UnitTest {
 
 	test_yourMethod {
 		// every method whose name begins with "test_" will be run
+		var synth;
 
 		this.assert( 6 == 6, "6 should equal 6");
 
 		this.assertEquals( 9, 9, "9 should equal 9");
 
-		this.assertFloatEquals( 4.0 , 1.0 * 4.0 / 4.0  4.0,
+		this.assertFloatEquals( 4.0, 1.0 * 4.0 / 4.0 * 4.0,
 			"floating point math should be close to equal");
 
 		// we are inside a Routine, you may wait
 		1.0.wait;
 
 		// this will wait until the server is booted
-		this.bootServer;
-
 		// if the server is already booted it will free all nodes
 		// and create new allocators, giving you a clean slate
-		p = { SinOsc.ar };
-		p.play;
-		p.register;
+		this.bootServer;
+
+		synth = { SinOsc.ar / 10 }.play;
+		synth.register; 
+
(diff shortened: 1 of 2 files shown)


dyfer:
Hello @cdbzb 
I'm afraid the proposed fix causes some side effects, please see the linked issue for details.

cdbzb:
Hi @dyfer can you please recheck with the code in the PR? Thanks!

dyfer:
@cdbzb sorry, this was my bad. I tested code with a manually copied change with missing `^`.

However, there are two issues about this PR:
- it includes an unrelated commit/change in `UnitTest.schelp`
- I have another question about this change and I posted it in the original issue. Once we agree on what the proper change is (i.e. what to do if the rate is not `\audio`), this PR should include a single commit with only that change. The "proper" way to do this is with interactive rebasing (see wiki), but for a simple change like this it might be easier to start a new branch/PR - just make sure to base it off of the `develop` branch, and not an existing PR branch. Does that make sense?

cdbzb:
Sorry for that I’ll fix tonight (and thanks for the hand holding - much appreciated)

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #6066, opened 2023-08-12
Pbind multiple instruments check and error

JordanHendersonMusic:
## Purpose and Motivation

Pbind does not work with multiple instruments, this checks for that, and tells the user how to fix it.

## Types of changes

- Documentation
- Usability and Errors

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [ ] there might be some disagreement whether this is the correct solution or if it should be dealt with automatically. https://scsynth.org/t/pbind-multichannel-expansion/8057/25

### DIFF
--- a/HelpSource/Classes/Pbind.schelp
+++ b/HelpSource/Classes/Pbind.schelp
@@ -10,9 +10,6 @@ The patterns bound to keys are referred to as emphasis::value patterns:: and the
 
 The keys used in a Pbind are usually determined by link::Classes/Event::'s default mechanism and the controls defined for the link::Classes/SynthDef:: to be played. (See link::#SynthDef and Event:: below for a brief discussion of both in relation to Pbind.)
 
-
-
-
 ClassMethods::
 
 method::new
@@ -554,6 +551,3 @@ f.stream = Pbind(
 ).asStream;
 )
 ::
-
-
-
--- a/HelpSource/Classes/Pbind.schelp
+++ b/HelpSource/Classes/Pbind.schelp
@@ -10,6 +10,9 @@ The patterns bound to keys are referred to as emphasis::value patterns:: and the
 
 The keys used in a Pbind are usually determined by link::Classes/Event::'s default mechanism and the controls defined for the link::Classes/SynthDef:: to be played. (See link::#SynthDef and Event:: below for a brief discussion of both in relation to Pbind.)
 
+
+
+
 ClassMethods::
 
 method::new
@@ -88,6 +91,42 @@ Pbind(*[
 ]).play;
 )
 ::
+
+
+subsection::Multiple Instruments
+
+Pbind cannot be used to play multiple instruments. The following will not work.
+
+code::
+Pbind(
+     \instrument, [\beep, \boop], // this line is wrong
+     \dur, 1,
+     \freq, 220
+ )
+::
+
+This is a limit of the implmentation, but also encourages the use of smaller more managable patterns.
+
+To help solve this the following function is provided, but it is just one way to solve the problem - feel free to copy it into your projects as needed!
+
+code::
+~putInstrumentsInPpar = {|...instruments|
+	Ppar(instruments.collect({|i| Pbind(\instruments, i) }), inf)
+};
+
+Pdef(\beepAndBoop, 
+	~putInstrumentsInPpar.(\beep, \boop) <> Pbind(
+		\dur, 1,
+		\freq, 220
+	)
+);
+
+Pdef(\beepAndBoop).trace.play
+::
+
+This uses a function to build up a Ppar. Then the Ppar and the Pbind are combined with `<>` (see PChain for more info). 
+
+
 
 subsection::SynthDef and Event
 
@@ -515,3 +554,6 @@ f.stream = Pbind(
 ).asStream;
 )
 ::
+
+
+
--- a/SCClassLibrary/Common/Streams/Patterns.sc
+++ b/SCClassLibrary/Common/Streams/Patterns.sc
@@ -344,7 +344,7 @@ Pbind : Pattern {
 	*new { arg ... pairs;
 		if (pairs.size.odd, { Error("Pbind should have even number of args.\n").throw; });
         if (pairs.asEvent[\instrument].rank != 0, 
-            { Error("Pbind does not work on multiple instruments, please see section in Pbind help file for to achieve the same result with Ppar.").throw });
+            { Error("Pbind does not work on multiple instruments, please see section in Pbind help file for how to achieve the same result with Ppar.").throw });
 		^super.newCopyArgs(pairs)
 	}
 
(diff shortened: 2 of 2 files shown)


JordanHendersonMusic:
Does any one know if it is possible to create link to the documentation in the post window?

jamshark70:
I'm not sure if it's entirely ready for a PR but I have started on multiple instrument support here: https://github.com/jamshark70/supercollider/commit/3aa6de8e36004cdc90077352868380e8289e3a8b

HelpSource/Classes/Pbind.schelp
```diff
@@ -89,6 +89,42 @@ Pbind(*[
...
+ )
+::
+
+This is a limit of the implmentation, but also encourages the use of smaller more managable patterns.
+
+To help solve this the following function is provided, but it is just one way to solve the problem - feel free to copy it into your projects as needed!
+
+code::
+~putInstrumentsInPpar = {|...instruments|
+	Ppar(instruments.collect({|i| Pbind(\instruments, i) }), inf)
```
telephon, line 111: There is a secret convention that `i` is the counting argument. Better use `item`, `each`, or `x`.

HelpSource/Classes/Pbind.schelp
```diff
@@ -89,6 +89,42 @@ Pbind(*[
...
+Pbind(
+     \instrument, [\beep, \boop], // this line is wrong
+     \dur, 1,
+     \freq, 220
+ )
+::
+
+This is a limit of the implmentation, but also encourages the use of smaller more managable patterns.
+
+To help solve this the following function is provided, but it is just one way to solve the problem - feel free to copy it into your projects as needed!
```
telephon, line 107: This si a little wordy for a helpfile. Maybe: "the following is one of many ways to handle multiple instruments in parallel"

Also, you can't just copy this into projects, because the environment variables may not work, eg. when you are inside patterns.

HelpSource/Classes/Pbind.schelp
```diff
@@ -89,6 +89,42 @@ Pbind(*[
...
+	~putInstrumentsInPpar.(\beep, \boop) <> Pbind(
+		\dur, 1,
+		\freq, 220
+	)
+);
+
+Pdef(\beepAndBoop).trace.play
+::
+
+This uses a function to build up a Ppar. Then the Ppar and the Pbind are combined with `<>` (see PChain for more info).
```
telephon, line 124: `Pchain` instead of `PChain`. You may want to write `link::Classes/Pchain::`

SCClassLibrary/Common/Streams/Patterns.sc
```diff
@@ -343,6 +343,8 @@ Pbind : Pattern {
 	var <>patternpairs;
 	*new { arg ... pairs;
 		if (pairs.size.odd, { Error("Pbind should have even number of args.\n").throw; });
+        if (pairs.asEvent[\instrument].rank != 0,
```
telephon, line 346: This creates unnecessary objects for the garbage collector to deal with, just for a test. It also won't work for syntactic cases like: `Pbind([\instrument\freq], [[\be, \bo], [400, 600]]`.

telephon (changes requested):

cdbzb:
> I'm not sure if it's entirely ready for a PR but I have started on multiple instrument support here: [jamshark70@3aa6de8](https://github.com/jamshark70/supercollider/commit/3aa6de8e36004cdc90077352868380e8289e3a8b)

I've been using this for a few days without issues. being able to write 

`(instrument: [ \default, \sawSynth ],freq: [300,500]).play`

is nice!

I'd encourage you to do a PR

JordanHendersonMusic:
> I'm not sure if it's entirely ready for a PR but I have started on multiple instrument support here: [jamshark70@3aa6de8](https://github.com/jamshark70/supercollider/commit/3aa6de8e36004cdc90077352868380e8289e3a8b)

I'm just hanging off from making any changes until @jamshark70 gets back as that would be a better solution. I only proposed this as it would stop the question being asked multiple times on scsynth.

jamshark70:
Apologies... I've been tidying up a paper and this one drifted to the back burner. I can look at it in the next day or two, I think.

telephon:
An alternative may the a method that flops the whole event and then iterates over the list? This may be easier to maintain than the very complicated procedure.

telephon:
So roughly like this:

```

(
f = { |event|
	var msgFuncs = event[\instrument].asArray.collect { |x| var def = SynthDescLib.global[x]; def !? { def.msgFunc } };
	var msgArgs = event.use { msgFuncs.collect { |x| x.valueEnvir[0,2..] } };
	
	event.use {
		msgArgs = msgArgs.flat.as(IdentitySet).do { |key|
			key.envirPut(key.envirGet.value)
		};
	};
	
	event.asPairs.flop.collect(_.asEvent).do(_.play)
};
)


f.((instrument: \default, freq: [56, 87], amp: 0.2));
```

removing all the flop stuff from the `note` envir.

Just an idea …

jamshark70:
I just updated my branch for Pmono(Artic) support, passes a basic test:

```
(
p = PmonoArtic(#[default, fmEp],
	\degree, Pwhite(-7, 7, inf),
	\dur, 0.2,
	\legato, Pwrand(#[0.8, 1.01], #[1, 4].normalizeSum, inf)
).play;
)

p.stop;
```

For fun (and in the interest of testing with wrong input), I tried it both with and without fmEp being loaded. If the SynthDef is loaded, it generates pairs of messages and they play correctly. If the second one is not loaded, it still generates pairs of messages (which matches the single-instrument behavior when you provide a missing defname), and half of those messages fail in the server.

Julian's approach is probably fine for `\note` and `\set`.

When I added PmonoArtic, there really wasn't any way to do it other than make custom stream classes that are tightly coupled to the Event, because stream needs to remember and manipulate specific node IDs according to properties of the instrument. I don't believe it will be enough to split the multi-instrument event into separate events; the PmonoStream or PmonoArticStream will likely not get all of the node IDs that it's supposed to manipulate.

The approach in my branch is to create an infrastructure in the default event prototype to collect information about instruments into one place, and then use that to streamline the event type functions. Fixing up Pmono just now, I did run into some snags, but it would have been *a lot* more painful without the infrastructure. Having these support functions would also simplify custom event types with multiple instruments.

telephon:
> Should the conversation move over there now, and this pr be closed?

Over where?

telephon:
> don't believe it will be enough to split the multi-instrument event into separate events; the PmonoStream or PmonoArticStream will likely not get all of the node IDs that it's supposed to manipulate.
> 
> The approach in my branch is to create an infrastructure in the default event prototype to collect information about instruments into one place, and then use that to streamline the event type functions.

Yes, that is understandable. I hope we find a way to keep the complexity of Event as low as possible.

jamshark70:
OK, I just opened PR #6083. I did \note and the Pmono support event types. Just noticed `\set` isn't updated, but can't do it just this second. Anything else missing, I guess raise it as a comment on the PR.

PR #6067, opened 2023-08-12
Pfunc doc should reference Plazy for returning patterns

JordanHendersonMusic:
## Purpose and Motivation
Pfunc only works with values but the documentation does mention how to return a pattern, that is Plazy

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Pfunc.schelp
+++ b/HelpSource/Classes/Pfunc.schelp
@@ -4,7 +4,7 @@ categories:: Streams-Patterns-Events>Patterns>Function
 related:: Classes/Pfuncn, Classes/FuncStream
 
 description::
-Returns a link::Classes/Stream:: that returns values from the code::nextFunc::. To return a pattern, see Plazy.
+Returns a link::Classes/Stream:: that returns values from the code::nextFunc::. To return a pattern, see link::Classes/Plazy::.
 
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/Pfunc.schelp
```diff
@@ -4,7 +4,7 @@ categories:: Streams-Patterns-Events>Patterns>Function
 related:: Classes/Pfuncn, Classes/FuncStream
 
 description::
-Returns a link::Classes/Stream:: that returns values from the code::nextFunc::.
+Returns a link::Classes/Stream:: that returns values from the code::nextFunc::. To return a pattern, see Plazy.
```
cdbzb: should `Plazy` not be a link? link::Classes/Plazy

JordanHendersonMusic: Not too sure on how GitHub works, but I added another commit correcting this shortly after. It should be visible in the second commit. Is there something I have to do?

cdbzb: I see it - all good!

telephon (approved):
thanks!

PR #6049, opened 2023-06-28
sclang/scide: Pass and Merge multiple sclang_confs

mraethel:
## Purpose and Motivation

This PR introduces the environment variable `SCLANG_CONF_PATH` to pass multiple sclang configuration files to both `sclang` and `scide` terminal commands by merging those files in a non-breaking way:
 + Files are ordered:
    1\. File passed by `sclang -l /path/to/sclang_conf.yaml` command option
    2\. Files passed by PATH-like environment variable `SCLANG_CONF_PATH` (left to right)
    3\. If not 1, then user configuration file
    4\. If not 1 and 3, then global configuration file
 + Include paths depending on their position in the order:
   - If a path is listed in `excludePaths` of a configuration file without being listed in `includePaths` of a _preceding_ file, then the path is excluded. 
   - If a path is listed in `inlcudePaths` of a configuration file without being listed in `excludePaths` of a _preceding or the same_ file, then the path is included.
 +  Boolean options (`postInlineWarnings`, `excludeDefaultPaths`) are defined by the first file in the order to validly set the option.

See https://scsynth.org/t/additively-pass-quarks-to-supercollider-both-sclang-and-scide-from-the-command-line/6089/3 for motivation and further details.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] Updated documentation
- [x] This PR is ready for review

PR #6068, opened 2023-08-15
lang: pass multiple sclang configurations

mraethel:
## Purpose and Motivation

This PR introduces the environment variable `SCLANG_CONF_PATH` to pass multiple sclang configuration files to both `sclang` and `scide` terminal commands by merging those files in a non-breaking way:
 + Files are ordered:
    1\. File passed by `sclang -l /path/to/sclang_conf.yaml` command option
    2\. Files passed by PATH-like environment variable `SCLANG_CONF_PATH` (left to right)
    3\. If not 1, then user configuration file
    4\. If not 1 and 3, then global configuration file
 + Include paths depending on their position in the order:
   - If a path is listed in `excludePaths` of a configuration file without being listed in `includePaths` of a _preceding_ file, then the path is excluded. 
   - If a path is listed in `inlcudePaths` of a configuration file without being listed in `excludePaths` of a _preceding or the same_ file, then the path is included.
 +  Boolean options (`postInlineWarnings`, `excludeDefaultPaths`) are defined by the first file in the order to validly set the option.

See https://scsynth.org/t/additively-pass-quarks-to-supercollider-both-sclang-and-scide-from-the-command-line/6089/3 for motivation and further details.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] Updated documentation


dyfer:
Hello @mraethel 
Thanks so much for the PR.
I'm not able to provide a full review right now (sorry!), just wanted to mention two things:
- in previous discussions about SC standalone operation (see #5898, #5938), one issue that was coming up was that the language config paths could not be resolved relatively. Does this PR change that behavior? I'm not stating (yet) that it should or should not, but I think this is something to discuss if the language config is being reworked
- I'm not sure if renaming `testsuite/sclang` to `testsuite/lang` is needed, was there a specific goal you had in mind for this?

mraethel:
Hey @dyfer, thanks for the feedback and take your time with reviewing.
Concerning the points you made:
 - If I understand correctly you mean relative paths used in `sclang_conf.yaml` files for `includePaths` and `excludePaths` options. This PR does not change the previous behaviour of that. However I am happy to work on this in a seperate PR since it seems to be a much needed feature to make standalones work.
 - I thought that `testsuite/lang`'s structure should somehow reflect the project's root structure (like it's done for `testsuite/server`). We currently have `root/lang` and `root/testsuite/sclang`, which is slightly confusing. However I have no problem with reverting.

One suggestion: If we can get relative paths to work I could add a cmake variable (and build option) 'SCLANG_CONF_PATH'. Then we could have language config files for each system (i.e. `linux_sclang_conf.yaml`, `linux_standalone_sclang_conf.yaml`, `macos_sclang_conf.yaml`, ..., `testing_sclang_conf.yaml`) to specify exactly which (default) paths to include and exclude by setting the `SCLANG_CONF_PATH` variable to the location of the system's language config file. By that we probably could get rid of some system dependent code in lang. In addition the build option would support customization of paths to be included (Quarks and Extensions) by package managers.

mraethel:
Contrary to what I said above I have added a commit to resolve relative paths like this:
 - Relative paths to configuration files passed by `SCLANG_CONF_PATH` resolve to the program location or can be prefixed with `USERHOME` to resolve to the current user's home directory.
 - Relative paths inside a configuration file resolve to the location of that file or can be prefixed with `USERHOME` or `PROGRAM_LOCATION` to resolve to the current user's home directory or the program location respectively.

Further I have added a CMake build option `-DCMAKE_SCLANG_CONF_PATH` and adjusted documentation and tests. The `boost/dll` library is used to resolve the program location.

---

Concerning my suggestion from above I noticed four issues:
 - How do we resolve user language configurations and plugin paths at build time? _This is solved by relative paths and prefixes._
 - It seems that the only reason we have "default" paths is the automated creation of those paths if nonexistent (e.g. /home/user/.local/share/SuperCollider/Extensions). By removing the default paths category this would fall flat.
**Solution:** We could add a `defaultPaths` configuration option to either append all included paths from the the same configuration file to the default paths list or just try to create the directories as we process those paths.
 - In the PR description I have mentioned the order of passing configuration files introduced by this PR. The suggestion would make 1 non-exclusive to 3 and 4 anymore and therefore be a breaking change, because users would have to exclude 3 and 4 by hand in thier custom configuration files if needed.
 - The same goes for 3 and 4 being exclusive to each other. _However this can be solved by a preference chain using the " > " character I introduced: `SCLANG_CONF_PATH = "user_conf.yaml>system_conf.yaml"`. It passes the first existing configuration file of that chain._

cdbzb:
I built this on Mac and tested the relative paths in sclang_conf.yaml which worked for me! - I'll try to check out the additive aspect of this when I get a chance

cdbzb:
(On Mac) I tried passing in a directory for a second sclang_conf.yaml using SCLANG_CONF_PATH but did not have success.

I set up in the /tmp directory:
sclang_conf.yaml
filesAndConf +
   |  sclang_conf.yaml
   |  some other files

then 
`export SCLANG_CONF_YAML=/tmp/filesAndConf`

and finally
`sclang -l /tmp/sclang.conf.yaml`

only the paths in /tmp/sclang_conf.yaml were searched - the ones in filesAndConf/sclang_conf.yaml were not

mraethel:
Hey @cdbzb, thanks for testing! Could you first of all try to use `SCLANG_CONF_PATH` instead of `SCLANG_CONF_YAML`. My initial PR was using `SCLANG_CONF_YAML` like proposed in this [forum thread](https://scsynth.org/t/additively-pass-quarks-to-supercollider-both-sclang-and-scide-from-the-command-line/6089/3) but I think `SCLANG_CONF_PATH` is a [better fit](https://github.com/supercollider/supercollider/pull/6049#issuecomment-1644867172) for this. Sorry for the confusion!
If that doesn't work you could try to put quotes like `SCLANG_CONF_PATH="/path/to/file.conf"`. Notice that config paths include the actual config file in them.

cdbzb:
Sorry!! the error was mine I'm afraid!

Using `SCLANG_CONF_PATH ` everything worked as expected. 

I tested both with and without `-l` switch, and with and without multiple paths separated by `:` 

This seems quite useful, thank you.

cdbzb:
friendly bump if you have a minute to review @jrsurge - sorry to pester!

mraethel:
@jrsurge let me know if I can help with something or explain further details.

capital-G:
@mraethel this has now merge conflicts b/c of #6640 - sorry

Are we sure that stacking multiple YAML files is a good approach? It seems to me that the YAML approach is inherently sub-optimal IMO b/c of its implicitness and opaqness due to YAML and one probably wants to state the startup options of a process either explicit via CLI options (see #6640) or use ENV variables: both are easy to verify, modify and stack during startup procedure, but the YAML file requires parsing a YAML file, which is essentially not feasible with default unix tools, so the program which spawns sclang does not know the state of sclang.

With #6640 it is possible to start with a (default) YAML file and overwrite it explicitly for startup procedure (although I'd like to phase out the YAML file b/c of the reasons above). It would be really nice if this could also take ENV variables into account for advanced users?

What do you think @mraethel ?

mraethel:
Hey @capital-G!
Your suggestion seems reasonable. I remember the main goal of this was to pass multiple quarks additively, which certainly is also possible with CLI options (as in your PR) and ENV variables. I'll take a look at how this can be achieved for ENV variables.

I guess we can close this PR then.

PR #6072, opened 2023-08-18
added defaultBranchName method to Git.sc

### LINKED ISSUE #5096: How should SC support quarks with other default branch names than "master"?
## Motivation 
In https://github.com/supercollider/rfcs/pull/10, a related side aspect turned up: 
Authors of supercollider-quarks will likely decide to follow the github recommendation 
to rename default branches from "master" to "main"; the current implementations 
of the Quark and Git classes in SC have "master" branch names hardcoded in 3 places. 

SuperCollider should checkout default branches of quarks regardless of their names. 

More specifically, four compatibility use cases should work transparently: 
1 new-style SC versions (using default branch) should find new-style quarks (using default branch) 
2 new-style SC versions (using default branch) should find old-style quarks  (with "master" branch) 
3 old-style SC versions ("master" branch) should find new-style quarks (using default branch) 
4 old-style SC versions ("master" branch) should find old-style quarks  (with "master" branch) 

1 and 4 are trivial, 2 is easy, 3 will likely be clumsy, and require a transition period. 

## Plan for Implementation
- The relevant methods could have branch name arguments for full flexibility
-  New-style quarks could list their default branch name in the description file
- checkout branch name given in description file
- if none given, Strategy for checkout could be: 
-- ask for default branch , then for main, then master

case 3 would require that quark authors keep a "master" branch around for a transition period;
if there is a way to redirect a call to "master" to another branch, this is easy. 

Questions for the git experts: 
- how to checkout the default branch without calling its explicit name? 
- what is the simplest way to set up an alias branch name in a repo, such that 
  asking for hardcoded "master" name returns default branch?

### LINKED ISSUE #6071: Quark.update fails if quark repo uses "main" rather than "master"

Calling Quark("scparco").update leads to an error ( @shimpe )

"fatal: couldn't find remote ref master"

This is because the branch "master" is hard coded into the .update method here

https://github.com/supercollider/supercollider/blob/9d54ee82fbffc8ea4fb1e07368de8dd4b81bf44f/SCClassLibrary/Common/Quarks/Quark.sc#L107

This could arguably be called an enhancement but A) the docs do not insist that Quarks use "master" and B) more and more repos are using "main"

cdbzb:
## Purpose and Motivation
currently "master" is hardcoded as the branch to checkout in Quark.update causing update to fail for Quarks using "main" (which has become a norm in recent years).

This adds a method which uses "git symbolic-ref refs/remotes/origin/HEAD --short
" to find out the appropriate name

Fixes #6071  
Fixes #5096

## Types of changes

- Bug fix

## To-do list

If someone could test further it would be great! Its working for me so far.

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
+++ b/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
@@ -475,7 +475,7 @@ QuarkDetailView {
 		if(tags.size == 0 or: {
 			remoteLatest != model.git.shaForTag(tags.first);
 		}, {
-			versions.add([remoteLatest, "Latest master"]);
+			versions.add([remoteLatest, "Latest default"]);
 		});
 
 		// working copy
--- a/SCClassLibrary/Common/Quarks/Git.sc
+++ b/SCClassLibrary/Common/Quarks/Git.sc
@@ -1,5 +1,5 @@
 Git {
-	var <>localPath, >url, tag, sha, remoteLatest, tags;
+	var <>localPath, >url, tag, sha, remoteLatest, defaultBranchName, tags;
 	classvar gitIsInstalled;
 
 	*isGit { |localPath|
@@ -17,7 +17,7 @@ Git {
 		this.url = url;
 	}
 	pull {
-		this.git(["pull", "origin", "master"])
+		this.git(["pull", "origin", this.defaultBranchName])
 	}
 	checkout { |refspec|
 		this.git(["checkout", refspec])
@@ -88,10 +88,16 @@ Git {
 			sha = this.git(["rev-parse HEAD"]);
 		}
 	}
+	defaultBranchName {
+		//find out what the remote default branch name is (ie master or main)
+		^defaultBranchName ?? {
+			defaultBranchName = this.git(["symbolic-ref refs/remotes/origin/HEAD --short"]).split($/)[1]
+		}
+	}
 	remoteLatest {
 		// find what the latest commit on the remote is
 		^remoteLatest ?? {
-			remoteLatest = this.git(["rev-parse origin/master"]);
+			remoteLatest = this.git(["rev-parse origin/" ++ this.defaultBranchName]);
 		}
 	}
 	tags {
--- a/SCClassLibrary/Common/Quarks/Git.sc
+++ b/SCClassLibrary/Common/Quarks/Git.sc
@@ -97,7 +97,7 @@ Git {
 	remoteLatest {
 		// find what the latest commit on the remote is
 		^remoteLatest ?? {
-			remoteLatest = this.git(["rev-parse origin/" ++ this.defaultBranchName]);
+			remoteLatest = this.git(["rev-parse origin" +/+ ( this.defaultBranchName ? "" )]);
 		}
 	}
 	tags {
--- a/SCClassLibrary/Common/Quarks/Git.sc
+++ b/SCClassLibrary/Common/Quarks/Git.sc
@@ -17,7 +17,7 @@ Git {
 		this.url = url;
 	}
 	pull {
-		this.git(["pull", "origin", this.defaultBranchName])
+		this.git(["pull", "origin", this.defaultBranchName ? ""])
 	}
 	checkout { |refspec|
 		this.git(["checkout", refspec])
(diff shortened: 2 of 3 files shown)


dyfer:
Thanks for this! It does seem to work for a variety of repos I tested (using both default names).

I'd like to make sure that this is as robust as possible (despite our generally brittle git interfacing, which doesn't really provide a way to deal with errors).

After looking around (e.g. [here](https://stackoverflow.com/questions/47978081/why-is-remotes-origin-head-sometimes-mapped-to-origin-master-and-sometimes-n)) I see that it is possible (if maybe rare?) for `refs/remotes/origin/HEAD` to not exist (IIUC). I tried unsetting this in an existing repo, and the results were that `this.git(["symbolic-ref refs/remotes/origin/HEAD --short"])` returned an empty string, while `this.git(["symbolic-ref refs/remotes/origin/HEAD --short"]).split($/)[1]` returned `nil`. Since we don't really have any real error handling here, I personally think that `nil` returned from this method is okay here.
However, `this.git(["pull", "origin", this.defaultBranchName])` would end up trying to check out the `nil` branch, IIUC (again, not 100% sure since there's no error returned and I didn't dig super deep). In this case, I wonder whether in the `git -pull` method we shouldn't replace nil with an empty string at least? (e.g. `this.git(["pull", "origin", this.defaultBranchName ? ""])` I think this part could/should be investigated a bit.

Although cosmetic, I think additionally [this](https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc#L478) could be changed to `"Latest default branch"` or similar

cdbzb:
Agree that it should be `this.git(["pull", "origin", this.defaultBranchName ? ""])` 

git behavior if a branch is not specified is here: https://git-scm.com/docs/git-pull#_default_behaviour 

>If branch.<name>.merge configuration for the current branch <name> exists, that is the name of the branch at the remote site that is merged.

I can't immediately see a situation where this doesn't just work? (unless the user is working in their own branch?)

>Although cosmetic, I think additionally [this](https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc#L478) could be changed to "Latest default branch" or similar.

agree

>it would be great to update the help [here](https://github.com/supercollider/supercollider/blob/develop/HelpSource/Guides/UsingQuarks.schelp#L207),

This I think we could leave it as it is  - The graf is just referring to repos migrated from SVN and is more a historical document - since Felix did use "master" it is strictly true for for the referenced repos so perhaps fine as it is (for now! the whole section could probably be replaced) 

I'll make the above two changes...

dyfer:
> Agree that it should be `this.git(["pull", "origin", this.defaultBranchName ? ""])`

Shouldn't `? ""` or something else also be used [here](https://github.com/supercollider/supercollider/pull/6072/files#diff-81924613c731be6908759777925e75da67a9d874e5faebebaa29e17c34817e54R100) and [here](https://github.com/supercollider/supercollider/pull/6072/files#diff-78a02100c666befb3163f197184c8ae0a29d983785d6fce390e971fcc7b4eb71R107)? I'm not sure, just checking.

cdbzb:
good point. But no. `git checkout ""` will not work.

I think here we maybe have to fallback to "main" then "master" - I'll test in the AM

dyfer:
> Shouldn't `? ""` or something else also be used [here](https://github.com/supercollider/supercollider/pull/6072/files#diff-81924613c731be6908759777925e75da67a9d874e5faebebaa29e17c34817e54R100) and [here](https://github.com/supercollider/supercollider/pull/6072/files#diff-78a02100c666befb3163f197184c8ae0a29d983785d6fce390e971fcc7b4eb71R107)? I'm not sure, just checking.

@cdbzb what do you think?

cdbzb:
hi @dyfer ...sorry for the slow reply!

I added the check for nil to both your suggestions above out of an abundance of caution! 

looking at it though I'm not totally sure I understand why the git.checkout is being performed after git pull but at least it will not fail if `defaultBranchName` is nil!

dyfer:
> looking at it though I'm not totally sure I understand why the git.checkout is being performed after git pull

Yeah, when I looked at this, it also jumped out at me. Seems like it's a bit of a wonky process... but fixing that is of course something for another PR, if at all.

Thanks for your work on this!

dyfer (approved):

PR #6073, opened 2023-08-19
making Synthdef work in many threads

JordanHendersonMusic:
## Purpose and Motivation

Go off the conversation around threads on scsynth....
If many things are promises or otherwise capable of waiting, it becomes necessary for synthdefs to work in multiple threads. 
Since synthdef is the most 'basic' and only way for a user or even Quark library writer to make a 'synth' (as it is tied into UGen.buildSynthDef) it ought to do the write thing.

Further, in my own work I often want to make owned resource as a part of the synthdef and synth creation process.
Since I have rewritten most of my code to make use of promises, not being able to define multiple synthdefs at a time is a big issue. This fixed that, and as far as I can tell, is the only way to fix it.

## Types of changes

- A couple of new classes to deal with thread local and thread hierarchy  singletons 
- Bug fix - synthdef doesn't work in threads
- Breaking change (small) - `buildSynthDef` -> `UGen.buildSynthDef`

There is one breaking change.
Code that once called the classvar `buildSynthDef` must now call `UGen.buildSynthDef`. Many classes already do this so there was only a few changes to make. 
This will require Quarks to change their code, but it is minimal and once done, works as before.
I think this is the smallest change that is needed to achieve the result. 

## To-do list

I have not done the documentation for this as I'm just curious to know if it might be accepted.

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

Examples

This code works, clearly alternating between each synthdef whilst still producing the correct controls.
```supercollider
(
s.waitForBoot {
	fork {
		SynthDef(\a, {
			'doingA'.postln;
			\arg1.kr(1);
			1.0.wait;
			'doingA'.postln;
			\arg2.kr(2);
			SinOsc.ar(223, mul: 0.1) |> Out.ar(0, _)
		}).add;
	};

	fork {
		SynthDef(\b, {
			0.5.wait;
			'doingB'.postln;
			\argm1.kr(-1);
			1.0.wait;
			'doingB'.postln;
			\argm2.kr(-2);
			SinOsc.ar(432, mul: 0.1) |> Out.ar(1, _)
		}).add;
	};
}
)


SynthDescLib.at(\a)
SynthDescLib.at(\b)
```

### DIFF
--- a/SCClassLibrary/Common/Audio/UGen.sc
+++ b/SCClassLibrary/Common/Audio/UGen.sc
@@ -1,5 +1,4 @@
 UGen : AbstractFunction {
-	classvar <>buildSynthDef; // the synth currently under construction
 	var <>synthDef;
 	var <>inputs;
 	var <>rate = 'audio';
@@ -7,6 +6,8 @@ UGen : AbstractFunction {
 	var <>synthIndex = -1, <>specialIndex=0;
 
 	var <>antecedents, <>descendants, <>widthFirstAntecedents; // topo sorting
+
+    *buildSynthDef { ^SynthDef.getActiveSingleton }
 
 	// instance creation
 	*new1 { arg rate ... args;
@@ -305,7 +306,7 @@ UGen : AbstractFunction {
 	@ { arg y; ^Point.new(this, y) } // dynamic geometry support
 
 	addToSynth {
-		synthDef = buildSynthDef;
+		synthDef = UGen.buildSynthDef;
 		if (synthDef.notNil, { synthDef.addUGen(this) });
 	}
 
@@ -551,7 +552,7 @@ UGen : AbstractFunction {
 					a.optimizeGraph
 				}
 			};
-			buildSynthDef.removeUGen(this);
+			UGen.buildSynthDef.removeUGen(this);
 			^true;
 		};
 		^false
@@ -611,7 +612,7 @@ OutputProxy : UGen {
 		^super.new1(rate, itsSourceUGen, index)
 	}
 	addToSynth {
-		synthDef = buildSynthDef;
+		synthDef = UGen.buildSynthDef;
 	}
 	init { arg argSource, argIndex;
 		source = argSource;
--- a/SCClassLibrary/Common/Audio/BasicOpsUGen.sc
+++ b/SCClassLibrary/Common/Audio/BasicOpsUGen.sc
@@ -172,7 +172,7 @@ BinaryOpUGen : BasicOpUGen {
 
 		if (b.isKindOf(UnaryOpUGen) and: { b.operator == 'neg' and: { b.descendants.size == 1 } }) {
 			// a + b.neg -> a - b
-			buildSynthDef.removeUGen(b);
+			UGen.buildSynthDef.removeUGen(b);
 			replacement = a - b.inputs[0];
 			// this is the first time the dependants logic appears. It's repeated below.
 			// We will remove 'this' from the synthdef, and replace it with 'replacement'.
@@ -186,7 +186,7 @@ BinaryOpUGen : BasicOpUGen {
 
 		if (a.isKindOf(UnaryOpUGen) and: { a.operator == 'neg' and: { a.descendants.size == 1 } }) {
 			// a.neg + b -> b - a
-			buildSynthDef.removeUGen(a);
+			UGen.buildSynthDef.removeUGen(a);
 			replacement = b - a.inputs[0];
 			replacement.descendants = descendants;
 			this.optimizeUpdateDescendants(replacement, a);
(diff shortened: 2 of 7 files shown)


telephon:
Interesting. In terms of implementation, don't you think this may be done with composition rather than inheritance? It seems a strong constraing that a class must inherit from the "manager" to use the manager. Imagine a pattern that needs it – you'd have to make every pattern a manager.

jamshark70:
I think it's worth considering very carefully how much more sophisticated SynthDef needs to be.

SynthDef is fundamentally the language-side representation of a GraphDef in the server. Making a GraphDef is an extremely complicated process, and I think because of this we sometimes want SynthDef to "do it all" -- but SC needs to have a low-level server-object abstraction for GraphDef. If SynthDef gloms on more features, then it would be necessary to move at least some of what is now SynthDef into a lower-level class.

With that in mind, what are the resources which *both* 1/ require blocking and 2/ properly belong in the middle of a SynthDef function?

`Buffer.read` should... well, "never say never," but I feel pretty confident in saying this should *never* be done inside a SynthDef. The user *should* get an error if they try to do this (perhaps even especially if they're new users: encouraging bad habits at the beginning seems to me like an anti-pattern).

The only other thing I can think of is to get something from a file or UI, or OSC resource... but a strong argument could be made that the user should collect all the data first, and only when everything is ready, then build the SynthDef.

So my general suggestion for extensions to SynthDef is not to change SynthDef, but to build on top of SynthDef. There could be a SynthDef-builder-helper that manages the traffic flow, and finishes up with the SynthDef construction.

jamshark70:
With all of that said, if the interpreter ever supports pre-emptive multitasking, then SynthDef will have to be made thread safe anyway... so it may be worth fixing that up in advance.

JordanHendersonMusic:
@telephon composition > inheritance
That was the wrong choice, changed.

@jamshark70 
> we sometimes want SynthDef to "do it all" -- but SC needs to have a low-level server-object abstraction for GraphDef.
> [...] 
So my general suggestion for extensions to SynthDef is not to change SynthDef, but to build on top of SynthDef. There could be a SynthDef-builder-helper that manages the traffic flow, and finishes up with the SynthDef construction.

Completely agree, unfortunately Ugen and similar classes use many of SynthDef's methods  to reorder the graph and do other stuff - and Quarks too may do this. In other words, the interface of SynthDef is integral to the interface of UGen and all unit generators. This is this the only way I could think of doing this without breaking compatibility.

> Buffer.read should... well, "never say never," but I feel pretty confident in saying this should never be done inside a SynthDef.

Completely disagree. I don't want to mess around with creating and managing resources separate from where they are used, particularly if they are used once and exist from beginning, to the end of the piece.

@smrg-lm 
> My point is that if it was just for that reason these changes are too complicated while the problem is elsewhere.

I don't actually understand the point.
The change here is very simple, it is just that a bunch of other classes needed to be updated to call a method rather than access a classvar.

SCClassLibrary/Common/Audio/UGen.sc
```diff
@@ -8,6 +7,8 @@ UGen : AbstractFunction {
 
 	var <>antecedents, <>descendants, <>widthFirstAntecedents; // topo sorting
 
+    *buildSynthDef { ^SynthDef.getActiveSingleton }
```
JordanHendersonMusic, line 10: This is the actually change to the public interface.

smrg-lm:
Just checked, I don't have the brain to follow right now, but the ugen graph is actually built from ugen classes, it's just an AST based on the AbstractFunction interface, for some reason it depends on the state of the SynthDef class (it's the graph processor/compiler), the thing that is it uses a global (library wise) state. My hypothesis was it could be simpler to pass the AST to an instance object for processing, with some simple and known compilers techniques, instead of using a global state in the library and the ping-pong methods between SynthDef and UGen. Well, that said, it's out of scope. It was on my todo list.

I don't have a position about this change, for me the problem is elsewhere, maybe there are too much diff changes in SynthDesc.

jamshark70:
> > Buffer.read should... well, "never say never," but I feel pretty confident in saying this should never be done inside a SynthDef.
> 
> Completely disagree. I don't want to mess around with creating and managing resources separate from where they are used, particularly if they are used once and exist from beginning, to the end of the piece.

Will take this question to the forum... not quite directly related to the PR itself.

JordanHendersonMusic:
My apologise for leaving this so long — life.

In the mean time I've thought about this some more and there are many quarks that this would break, here is a quick list of quarks that touch the member `Ugen.buildSynthDef` based off a quick search on https://github.com/supercollider-quarks/downloaded-quarks.

1. Feedback
2. Bending
3. autogui
4. cruciallib Instr
5. Sequencer
6. wslib
7. Steno
8. cruciallib Players
9. UgenStructure

About half of these can be easily fixed by turning the member access into a method call (`Ugen.buildSynthDef()`), but the other half would a large amount of work — most would require making thread safe, while Bending would be the biggest challenge.

While Synthdef should probably be made thread safe, without a way to tie quark versions to sclang versions I don't think it is currently worth the cost of breaking other's code.

Should I close this pr?

PR #6085, opened 2023-09-10
docs: Add overview of pattern midi events

madskjeldgaard:
## Purpose and Motivation

This is a small PR that adds an overview file in the help system for midi commands. I generally found it very hard to figure out without having to dig into the Event.sc source code, so I figure someone else could use this :) 

## Types of changes

- Documentation

### DIFF
--- a/HelpSource/Overviews/MidiPatterns.schelp
+++ b/HelpSource/Overviews/MidiPatterns.schelp
@@ -1,0 +1,184 @@
+title:: How to control midi events using patterns
+categories:: Midi, Streams-Events-Patterns
+summary:: An overview of midi commands and how to use them with Pbind, etc.
+
+This document is a brief overview of some commonly used midi commands
+available in the midievent part of LINK::Classes/Event:: and which arguments
+they offer.
+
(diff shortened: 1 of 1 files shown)


madskjeldgaard:
I didn't include the `\smpte` nor `\songPtr` events since I couldn't imagine a situation where I would use them, but if anyone has some examples to add for these, feel free.

HelpSource/Overviews/MidiPatterns.schelp
```diff
@@ -0,0 +1,184 @@
...
+)
+
+p.stop;
+::
+
+SECTION:: \polyTouch
+
+CODE::
+p = Pbind(
+	// These keys are common to all midi commands
```
cdbzb, line 101: maybe instead of `// These keys are common to all midi commands` it should be
` //  the \type, \midiout, and \midicmd keys are common to all midi commands`

I stared at these examples for a minute before understanding (though I might be dense!)

Either way this is super helpful...

cdbzb (approved):
this is helpful - thanks!

mossheim (approved):
thanks :)

Issue #4261, opened 2019-01-20
Wrap, Fold: mismatched input/output rates cause incorrect output and NaNs


mtmccrea:
Environment
-----------

* Your SuperCollider version: 3.10.0

Current Behavior
----------------
`kr` input signals to the `ar` method of `Fold` and `Wrap` cause incorrect values and `NaN`s.

Expected Behavior
-----------------
You shouldn't of course be sending a control-rate signal into the `ar` method, but it still shouldn't break.

Steps to reproduce (for bugs)
-----------------------------

```supercollider
// Needs to limit first input to output rate
// ar/kr
{Fold.ar(Line.kr(1, 10, 0.2), 0.5, 500)}.loadToFloatArray(0.0025, action: {|arr| a = arr})
a.any(_.isNaN) // >> true: nan in output
a.do(_.postln) // and values are just wrong

// ar/ar
{Fold.ar(Line.ar(1, 10, 0.2), 0.5, 500)}.loadToFloatArray(0.0025, action: {|arr| a = arr})
a.any(_.isNaN) // >> false
a.do(_.postln) // and values are OK

// ar/kr
{Wrap.ar(Line.kr(1, 10, 0.2), 0.5, 500)}.loadToFloatArray(0.0025, action: {|arr| a = arr})
a.any(_.isNaN) // >> true: nan in output
a.do(_.postln) // and values are just wrong

// ar/ar
{Wrap.ar(Line.ar(1, 10, 0.2), 0.5, 500)}.loadToFloatArray(0.0025, action: {|arr| a = arr})
a.any(_.isNaN) // >> false
a.do(_.postln) // and values are OK
```

Suggested solution
-------------------
This should either be taken care of in the class definition (catching the `kr` rate input), or the underlying UGen should correctly accommodate the `kr` input.

madskjeldgaard:
Has anyone started working on this? If not I think the rate checking solution in the sc class is a good idea

telephon:
yes, that will be the easiest for the time being. The methods `. fold` etc. will choose the correct version anyway.

mtmccrea:
Will be fixed by #6088

PR #6088, opened 2023-09-15
Add checkSameRateAsFirstInput to InRange and checkValidInputs to AmpComp

### LINKED ISSUE #5547: InRange/Clip/Wrap/Fold need to check audio rate

cdbzb:
## Purpose and Motivation

Fixes #5547, #4261

Added checkSameRateAsFirstInput to InRange and added checkValidInputs to AmpComp

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Audio/Line.sc
+++ b/SCClassLibrary/Common/Audio/Line.sc
@@ -45,7 +45,7 @@ AmpComp : PureUGen {
 	*ir { arg freq = 60.midicps, root = 60.midicps, exp = 0.3333;
 		^this.multiNew('scalar', freq, root, exp)
 	}
-	*ar { arg freq = 60.midicps, root = 60.midicps, exp = 0.3333;
+	*ar { arg freq, root = 60.midicps, exp = 0.3333;
 		^this.multiNew('audio', freq, root, exp)
 	}
 	*kr { arg freq = 60.midicps, root = 60.midicps, exp = 0.3333;
--- a/SCClassLibrary/Common/Audio/Line.sc
+++ b/SCClassLibrary/Common/Audio/Line.sc
@@ -51,7 +51,10 @@ AmpComp : PureUGen {
 	*kr { arg freq = 60.midicps, root = 60.midicps, exp = 0.3333;
 		^this.multiNew('control', freq, root, exp)
 	}
-	checkInputs { ^if(rate === \audio) { this.checkSameRateAsFirstInput } }
+	checkInputs {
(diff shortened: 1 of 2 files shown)


mtmccrea (approved):
I think this looks good. I did some basic testing and the checks seem to work properly.

Note in the case of `AmpComp` this change replaces an _identity_ check with an _equality_ check, but this seems to be OK, as there is not a consistent approach to this (many other cases check equality), and `UGen::new1` checks that `rate` is a Symbol:

https://github.com/supercollider/supercollider/blob/2db872ad2a42ff85726566149855ecdb60d65b77/SCClassLibrary/Common/Audio/UGen.sc#L13

mtmccrea:
@cdbzb, I've added 2 more commits removing the default values for the first argument of the *ar methods, since scalar rate values aren't supported.

telephon (approved):

mtmccrea:
Because I was working nearby on another PR I added some documentation for these argument rate specs.
https://github.com/supercollider/supercollider/pull/6172/commits/08f2821df58804e5a6ef52154bbac64f50ce9343

dyfer:
Hello @telephon @mtmccrea, is there a reason this hasn't been merged yet?

mtmccrea:
Thanks for the ping @dyfer

PR #6094, opened 2023-09-25
Fixed typos in Main.sc

remnantblackbox:
## Types of changes

- Fix typos

## To-do list

### DIFF
--- a/SCClassLibrary/DefaultLibrary/Main.sc
+++ b/SCClassLibrary/DefaultLibrary/Main.sc
@@ -1,5 +1,5 @@
 Main : Process {
-	var <platform, argv;
+	var platform, argv;
 	var recvOSCfunc, prRecvOSCFunc;
 	var recvRawFunc, prRecvRawFunc;
 	var openPorts;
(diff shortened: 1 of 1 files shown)


muellmusik:
Thanks for this, a couple of comments in my review...

dyfer:
@muellmusik have you submitted your review?

In any case removing a getter in `Process` seems like a potentially intrusive change, definitely not a typo fix.

nuss:
> In any case removing a getter in `Process` seems like a potentially intrusive change, definitely not a typo fix.

`Main` is responsible for creating a singleton instance that's held in `thisProcess`. So, calling `thisProcess.platform` returns the OS platform you're currently on. Hence, `platform` **must** be kept as a  getter if that functionality shall not get lost  (see https://doc.sccode.org/Classes/Main.html#.thisProcess)

remnantblackbox:
i'm sorry for my mistake, thank you all for the explanation. 
I'll correct it right away, if it's ok

dyfer:
@remnantblackbox thanks for correcting the getter. 
I appreciate your time to make the PR. At the same time, our code style actually [recommends to omit the semicolon](https://github.com/supercollider/supercollider/wiki/Code-style-guidelines#recommendation-dont-place-a-semicolon-after-the-final-statement-of-a-method-or-function) after the last statement of a function to indicate a return value. This might not be exactly applicable in this case as we're not using the return value of the function, but still, this is the last statement of the function and I don't see how adding the semicolon helps with readability or anything else here.

dyfer:
With no response, I'm closing this PR. Feel free to reopen if you feel this is needed.

Issue #558, opened 2012-10-02
Option for lazy auto-completion (trigger by esc)


telephon:
Currently, auto-completion is always on, even when you know what you are writing anyhow. It would be useful to make it lazy so that you can trigger it when you need it at a specific point. 

The key binding could be esc - which makes it effectively toggle on/off with esc and which is very intuitive. TeXShop e.g. does it like this, too.

brianlheim:
@telephon I can work on this. For anyone else interested @jleben @muellmusik would Alt-Space or Ctrl-Space make more sense? Currently it seems we have a plurality of non-Cmd shortcuts using Alt.

telephon:
if it is easy enough, yes, would be very nice. It's not a very high priority of course.

Issue #555, opened 2012-10-01
find next / previous key binding


telephon:
It would be very good to have a key binding for find next and find previous. On OS X the standard ones are cmd-g [ctrl-g] and cmd-shift-g [ctrl-shift-g], used throughout the system. I'd suggest to replace cmd-g [ctrl-g] which currently means "go to line number X" by cmd-# [ctrl-#].

Issue #333, opened 2012-05-18
[SC-IDE] Document implementation


miguel-negrao:
I know this is a bit particular, but at least it's a use case to think about in terms of how to allow some degree of scripting of the editor.

For instance I would like to run a function on the result value of anything that I evaluate, with a special keyboard shortcut. I think I could have done this in the old Document class, right ?

scztt:
@miguel-negrao You okay closing this, since the Document class was revived? Is there still missing functionality?

telephon:
For 3.7 it is fine. We'll need a more particular issue for individual features.

PR #6095, opened 2023-09-26
CSS Themes for docs

### LINKED ISSUE #5477: Apply color theme to docs as well

capital-G:
Closes #5477
Closes #5487
Closes #6308 
Supersedes #6309

Top post last updated on 2024-05-25 - many thanks to @elgiano for working together on this one!

## Purpose and Motivation

This PR makes the docs theme-able via CSS and introduces the IDE themes for the docs as well by introducing CSS variables. 
This is a re-implementation of #5487 which used SASS to build the themed CSS files - this PR only uses plain CSS files to reduce build dependencies.

To store the active theme [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) of the Browser is used which acts like a cookie but is also supported in a `file://` environment instead of being limited to `http` - it defaults to `default` which is the current look of the docs.
Switching of the themes is managed via JavaScript, so it is independent of the IDE (and theme of IDE).

In order to make theming easier, the CSS colors have been consolidated (from around 100 different colors to around 20) and given names inspired by [tailwind css](https://tailwindcss.com).
The color for the themes were extracted from https://github.com/supercollider/supercollider/blob/develop/editors/sc-ide/core/settings/default_themes.cpp 

I also document the development workflow of adjusting the CSS through this setup within `HelpSource/themes/README.md`.
It is also documented via SCDocs how the used CSS variables can be used in `custom.css` to create custom themes.

A way to switch the theme from within the IDE has been once implemented within this PR (see https://github.com/supercollider/supercollider/pull/6095#issuecomment-2097114101) but has been scraped in order to *keep it simple*.

## Screenshots

The screenshot may not represent the latest state as the colors are still tweaked.

**Default**

image

image

**Classic**

image

**Dark**

image

**Dracula**

image

The theme can be switched via a *Theme* button in the navbar. This setting should be stored in the *localStorage* (see above) until the Browser-cache is cleared.

image

## Known problems

As the theme is stored in `localStorage` there seems to be a bug when multiple instances of sclang open the HelpBrowser.
This causes a conflict within the Chromium Engine which slows down the opening of the docs and may change the loaded theme on the 2nd instance of the docs.
For further information see https://github.com/supercollider/supercollider/pull/6095#issuecomment-2118897244 ff.
Although not nice, I think it is an edge case and the benefits outweigh this problem which could be deactivated by modifying the JS code manually.

## Types of changes

- [x] Documentation
- [x] New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

## Merge notes

In order to reduce noise and "un-touch" reverted code segments, the PR should be squashed and merged.


HelpSource/scdoc.js
```diff
@@ -190,6 +190,39 @@ function fixTOC() {
         });
     });
 
+    // if served via web, show theme switcher
+    // within IDE use the theme settings of the IDE to control the docs theme
+    if(window.location.protocol.startsWith("http")) {
```
dyfer, line 195: That means that helpfile opened locally in a help browser doesn't have this selector, and it won't follow the IDE theme (unless it would be possible to generated the css file dynamically when the IDE changes theme). That's not the end of the world, but maybe there would be a way to hide the selector only in the HelpBrowser opened by IDE, but show elsewhere, even when serving a local file?

dyfer (commented):
Thanks so much!

I have a few questions/suggestions that I'd like us to consider before moving forward with this. Overall, I think the approach with css variables is good!

Unfortunately I couldn't test this due to #6104, but I was wondering about updating the theme without restarting the IDE. Does the theme update on `SCDoc.indexAllDocuments` or `SCDoc.renderAll`? If so, maybe such sclang command could be triggered by the IDE when changing the theme (I don't know how, but I believe it should be possible ;) )

HelpSource/scdoc.js
```diff
@@ -190,6 +190,39 @@ function fixTOC() {
...
+        create_menubar_item("Theme \u25bc", "#", function (a, li) {
+            var indexes_menu = $("<div>", {class: "submenu"}).hide()
+                .appendTo(li);
+
+            var nav_items = ["classic", "dark", "default", "dracula", "solarizedDark", "solarizedLight"];
```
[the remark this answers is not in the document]
capital-G: Sadly not because accessing files on the local drive is rather limited through JavaScript (but for good reasons).

HelpSource/themes/README.md
```diff
@@ -0,0 +1,24 @@
+# CSS Themes
```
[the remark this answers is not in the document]
capital-G: I decided to put the documentation here because one would most likely not edit this within the SC-IDE environment and referring to relative paths seems easier to me with a `.md` file than a `.schelp` file.
If you think its better within a `.schelp` file I could adjust it to a `Themes.schelp`.

HelpSource/themes/classic.css
```diff
@@ -0,0 +1,36 @@
...
+    --sc-text-warning: var(--cm-keyword-color);
+    --sc-text-color-lighter: #333;
+
+    --sc-tertiary: var(--cm-built-in-color);
+
+    --sc-neutral: #aaa;
+    --sc-neutral-darker: #555;
+    --sc-neutral-lighter: #ccc;
+    --sc-link-color: var(--cm-string-color);
+}
```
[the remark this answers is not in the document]
capital-G: Would be possible but this would introduce the IDE as an dependency to the docs iirc - which would make headless-builds with docs more difficult.
Also I found it "saner" for me to keep C++ code with file operations to a minimum so it does not crash on some platforms.

And I found that most themes need some slight adjustments, for example `lighter` needs to be either a *lighter* or a *darker* color, based on the brightness of the general theme.
Currently the CSS of the Qt Web Browser does not support color operations which could automate this.

At the same time - setting a custom theme for the docs can still be achieved by over-writing the vars in a `custom.css`.

capital-G:
Thanks for the feedback @dyfer - I think we are on a good track here :)

I still trying to get the IDE to reload the Docs browser w/o restarting the IDE, and I also applied some little tweaks that I observed after a week of usage.

capital-G:
Hey @dyfer - I am really struggling with the C++ part of this - sometimes the theme gets updated when changed via the settings dialog, sometimes not - at least it always gets picked up when the IDE is restarted. But maybe this is also caused by the strange caching mechanism of the docs?
I think the 2nd time I switch the theme it doesn't get picked up by the browser anymore.

But there are also IDE actions I really don't understand, for example
https://github.com/supercollider/supercollider/blob/4da85d106c0e4cddb8af92d4d46778435db1b43a/editors/sc-ide/widgets/help_browser.cpp#L134
is supposed to reload the web browser document according to the [QT docs](https://doc.qt.io/qt-6/qwebengineview.html#reload) - but it doesn't, and there's not even a runtime error message.

And while I can attach a debugger to the IDE process via vscode, the breakpoints are not hit, which makes the development process really slow :/

Is there anyone who could provide a review for this and suggest how to do this properly or maybe even a shared screen session?

Also, this is blocked b/c there is currently no way to lint the C++ code on MacOS, see #6096 for a PR that could mitigate this.

smoge:
I created a patch of this PR because it's too cool to not be used

https://gist.github.com/smoge/57b1fd5375e5cd922757fa716c54c0e1

capital-G:
> I created a patch of this PR because it's too cool to not be used.
> 
> https://gist.github.com/smoge/57b1fd5375e5cd922757fa716c54c0e1
> 
> Any issue I should be aware?

It is not in a tested/finished state b/c I have some struggles regarding the WebEngine of Qt.
But sadly progress towards merging seems to be stuck somehow.

In the meantime I added a way to select a custom CSS file via a file dialog - it also has the default option to simply use the css file attached to the IDE theme.
This has IMO the best of both worlds: follow the IDE theme for the docs or select a custom or predefined CSS file for the docs.

image

image


If we could get the docs to include more `divs`, then we can also create an visual overhaul of the docs. But this is time for another PR.

smoge:
> It is not in a tested/finished state b/c I have some struggles regarding the WebEngine of Qt.
> But sadly progress towards merging seems to be stuck somehow, but I would love to get this merged.

I thought if changing SCDocRenderer.sc, SCDoc.renderAll would be enough. A simple configuration line in startup.scd would be enough for most people.

How WebEngine is creating problems? I will see and try to make sense.

smoge:
> > It is not in a tested/finished state b/c I have some struggles regarding the WebEngine of Qt.
> > But sadly progress towards merging seems to be stuck somehow, but I would love to get this merged.
> 
> I thought if changing SCDocRenderer.sc, SCDoc.renderAll would make it. Interesting that it goes through other places.
> 
> That would be good already:
> 
> > "default option to simply use the css file attached to the IDE theme."

Although, many people use SCDoc with the WebEngine without scide (emacs, vscode etc etc). It would mean the user would need to open scide just for that.

smoge:
> WebEngine of Qt.

What is the problem with WebEngine? I tried to find something in the code, but didn't find a conflict. 

Here, a simple test using a custom.css file simply works so far (not polished, of course)

Screenshot_2024-05-07_01-05-24

smoge:
> In the meantime I added a way to select a custom CSS file via a file dialog - it also has the default option to simply use the css file attached to the IDE theme.
> This has IMO the best of both worlds: follow the IDE theme for the docs or select a custom or predefined CSS file for the docs.

That's cool.

I've been just writing a custom.css file, and so far there have been no conflicts with javascript or webwidget. Do you know exactly how qt and js code interfere with scdoc theme?

s3

elgiano:
Hello everyone and thanks @capital-G for all the big work.
I'm thinking, maybe we could keep this simpler:
- SuperCollider provides a number of default themes, as you are doing here
- default themes are scoped as body classes, so theme selector just switches class on body and caches to localStorage
- display theme selector everywhere, I don't think it's a problem for IDE users to see it
- ScIDE can use mWebView->runJavaScript to call the same function as the Theme Selector
- users wanting to add custom themes can do so by writing their `custom.css` 

If needed, I can help with the cpp part!

capital-G:
Hey @elgiano 

> Hello everyone and thanks @capital-G for all the big work. I'm thinking, maybe we could keep this simpler: 
>     * display theme selector everywhere, I don't think it's a problem for IDE users to see it
>     * users wanting to add custom themes can do so by writing their `custom.css`

Thanks for this idea - this makes the theme switching the same on every platform and IDE and results in less and easier code. Just use the KISS principle :)
I reverted the adjustments on the IDE, it now looks like this

image

The theme menu is hidden if there is not enough width to display it.

Can you check out the build and give some feedback or a review?
I'd prefer this PR over your #6309 because here we already have a longer discussion and also implemented all necessary themes already.

If you are fine with this I'd add some documentation to this feature and ask some others for approval and then we hopefully can get this merged soon :)

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -230,11 +230,12 @@ SCDocHTMLRenderer {
 		// (Search.html, Browse.html, etc.) if necessary.
 		stream
 		<< "</title>\n"
+		<< "<link rel='stylesheet' href='" << baseDir << "/custom.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/scdoc.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/codemirror.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/editor.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/frontend.css' type='text/css' />\n"
-		<< "<link rel='stylesheet' href='" << baseDir << "/custom.css' type='text/css' />\n"
+    	<< "<link rel='stylesheet' id='scdoc-theme' href='" << baseDir << "/themes/default.css' type='text/css' />\n"
```
elgiano, line 238: Shouldn't custom.css be the last stylesheet loaded, after themes? I think it's supposed to be the last to be able to override everything

capital-G: Yes, thanks for the hint - changed it.

elgiano:
If we are going for localStorage to switch themes, we need to enable it for sclang's HelpBrowser too:
```diff
diff --git a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
index d7b77275e..09e119edf 100644
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -222,6 +222,7 @@ HelpBrowser {
                h = winRect.height - y - marg;
                webView = WebView.new( window, Rect(x,y,w,h) ).resize_(5);
                webView.setHtml("Please wait while initializing Help... (This might take several seconds the first time)");
+               webView.setAttribute(\localStorageEnabled, true);

                webView.setFontFamily(\fixed, Platform.case(
                        \osx, { "Monaco" },

```

elgiano (commented):
found a couple extra things

HelpSource/scdoc.js
```diff
@@ -190,11 +190,78 @@ function fixTOC() {
...
+        $(document).on("click", function (e) {
+            if (!$(e.target).closest(li).length) {
+                indexes_menu.hide();
+            }
+        });
+    });
+}
+
+function setTheme(themeName) {
+    console.log(`Set theme to ${themeName}`);
```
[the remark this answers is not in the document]
capital-G: Yes, I'll remove it.

capital-G:
> If we are going for localStorage to switch themes, we need to enable it for sclang's HelpBrowser too:
> 
> ```diff
> diff --git a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
> index d7b77275e..09e119edf 100644
> --- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
> +++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
> @@ -222,6 +222,7 @@ HelpBrowser {
>                 h = winRect.height - y - marg;
>                 webView = WebView.new( window, Rect(x,y,w,h) ).resize_(5);
>                 webView.setHtml("Please wait while initializing Help... (This might take several seconds the first time)");
> +               webView.setAttribute(\localStorageEnabled, true);
> 
>                 webView.setFontFamily(\fixed, Platform.case(
>                         \osx, { "Monaco" },
> ```

Interesting - for me it works without setting this flag, maybe it is on by default? Can anyone else verify? If it is on by default I'd consider keeping it implicit.

I think this PR could be very close to completion :)

HelpSource/themes/classic.css
```diff
@@ -0,0 +1,36 @@
...
+    https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix
+    */
+    --sc-bg-color: var(--cm-background-color);
+    --sc-bg-warning: var(--cm-selection-color);
+
+    --sc-text-color: var(--cm-text-color);
+    --sc-text-warning: var(--cm-keyword-color);
+    --sc-text-color-lighter: #333;
+
+    --sc-tertiary: var(--cm-built-in-color);
```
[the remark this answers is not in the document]
elgiano: Good idea! I agree. It makes sense to me that if one wants to change the "extension indicator", one could do it directly, without us needing to define a variable for it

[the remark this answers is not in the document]
capital-G: OK, I also think it is a good way. I deleted the variable.

HelpSource/scdoc.css
```diff
@@ -238,10 +238,10 @@ a.menu-link:hover {
     float: right;
     font-size: 12pt;
     padding: 0.4em 0.4em 0.2em 0.4em;
-    background-color: var(--sc-tertiary);
-    color: white;
+    background-color: var(--sc-bg-warning);
```
elgiano, line 241: Sorry for the misunderstanding, I didn't mean to remove the default extensions color, but rather to put back `#601c8b`

elgiano:
> Interesting - for me it works without setting this flag, maybe it is on by default? Can anyone else verify? If it is on by default I'd consider keeping it implicit.

You're right, it should be on by default, so it's not necessary to set it explicitly. However, I found out I can reproduce the error regardless when I:
- first open `HelpBrowser()` from the IDE
- then open another HelpBrowser() from a separate instance of `sclang`, outside of the IDE.
(it also works in reverse order: first sclang and then IDE)
The error:
```
[119803:119803:0518/162809.030883:ERROR:service_worker_storage.cc(1753)] Failed to delete the database: Database IO error
```
It's not fatal, it just makes HelpBrowser loading some (3-4) seconds slower, and not load the correct theme, but then theme switching works. And theme is shared between all sclang HelpBrowser instances (inside and outside the IDE), when they are not open at the same time. My intuition is that since both instances use the same localStorage, different processes are not allowed to write to it at the same time.

Now, I know this is a kind of an edge case, but I wouldn't feel good ignoring it. Sorry to prolong an already long discussion, but my suggestion is that we either try to find a way to not cause the error, or we simplify this PR further. Here is an idea for the latter:
- support only two themes, default and dark, and switch automatically according to system preferences. This would results also in what I would consider a better behavior for web docs.
- keep the other themes you made for a future PR where we also figure out a way for the IDE to switch them in. It could be by injecting javascript that directly changes css vars. After all it would make sense to me if only the IDE could switch in IDE themes.

What do you think about it? Again, this is my point of view, and I'm open to other people perhaps preferring to ignore this edge-case for now.

capital-G:
> * first open `HelpBrowser()` from the IDE
> 
> * then open another HelpBrowser() from a separate instance of `sclang`, outside of the IDE.
>   (it also works in reverse order: first sclang and then IDE)
>   The error:

TIL you can have multiple instances of the help browser :)

I replicated the steps, but for me, the "outside browser" can have a different theme and does not share the theme of the embedded help browser.
For me it is possible to create multiple external help browsers and they will share the same theme.
I didn't see any error in the console or any slowdown in the browser, even when I changed the theme on the "outside" browser.

image

Does this only happen when you switch themes? Or did you set any additional flags like "private mode" because it seems strange to me that a database should be deleted at this point?

I used Qt version `5.15.10`.


> * support only two themes, default and dark, and switch automatically according to system preferences. This would results also in what I would consider a better behavior for web docs.

Personally, I don't like automatic theme switching, especially since SC is also used as a visual component during livecoding we should be wary with any implicit or automatic changes here. Also, it should always be possible to switch the theme. We only have a semi-default light theme (status quo), but no "official" dark theme.

elgiano:
Just to be clear, what I meant to reproduce the error is that one needs to run an instance of `sclang` outside of the IDE process. For example it happens if one uses sclang in another editor (e.g. scnvim), or if it's invoked directly from command line or scripts. On macos you should be able to find the exact path of the sclang executable somewhere in `/Applications/SuperCollider.app/` (sorry I don't remember the exact path, I'm on linux).

When one runs multiple instances of `sclang` (for example one in the IDE and one from the command line) and open HelpBrowser in each of them, the error occurs. I don't know how many users are likely to be affected, personally I don't think many people would open multiple HelpBrowsers in multiple sclang processes, so for me it won't be a reason to block this PR, but maybe just something we should be aware of.

> it should always be possible to switch the theme and for this we would still need some way to store the state of the theme

I agree, and I don't see better alternatives than localStorage. The only one that comes to my mind is writing `frontend.css`, but perhaps it would be even worse, in terms of conflicts between multiple frontends.

HelpSource/scdoc.css
```diff
@@ -55,49 +60,14 @@ input {
...
-    color: #558;
-    font-weight: normal;
-/*    margin-left: 2em;*/
-}
-
-a.inheritedmets_toggle:hover {
-    text-decoration: none;
-    color: #000;
-}
-
```
elgiano, line 92: I understand you wanted to simplify the colors, but why did you remove all these rules? As a results inherited methods look less readable to me, as they get closer together since they lose some spacing.

HelpSource/themes/default.css
```diff
@@ -0,0 +1,36 @@
...
+
+    /* derived colors
+    in future use `color-mix` to calculate them which allows for fewer variables
+    but currently the qt browser does not support it, see
+    https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix
+    */
+    --sc-bg-color: var(--cm-background-color);
+    /* normally each theme uses --cm-keyword-color, but this generates grey text
+    on blue background, which is bad to read, therefore it gets manually adjusted here */
+    --sc-bg-warning: #aaa;
```
elgiano, line 26: Could the warning background for the default theme stay yellow as it was?

capital-G: Can do :)

HelpSource/scdoc.css
capital-G: Good catch - I reintroduced these lines back again.

HelpSource/scdoc.css
```diff
@@ -56,6 +56,41 @@ a:link:hover {
...
+    color: #558;
+    font-weight: normal;
+/*    margin-left: 2em;*/
+}
+
+a.inheritedmets_toggle:hover {
+    text-decoration: none;
+    color: #000;
+}
+
```
elgiano, line 93: Ok, layout got better, but now the colors are hardcoded again. Use your variables:
#000 is --sc-text-color
#333 is --sc-text-color-lighter
#555 and #558 are --sc-neutral-darker

HelpSource/scdoc.css
```diff
@@ -492,14 +502,14 @@ td p {
...
+    color: var(--sc-neutral-darker);
     text-align: right;
     margin-top: 2em;
     margin-bottom: 2em;
-    border-top: 1px solid #ddd;
+    border-top: 1px solid var(--sc-neutral-lighter);
 }
 .doclink a {
-    color: #444;
+    color: var(--sc-neutral-darker);
```
elgiano, line 512: should we give two different shades to `.doclink` and `.docklink a`? would be great to keep the differentiation

HelpSource/scdoc.css
```diff
@@ -13,8 +13,13 @@ html, body {
...
     font-family: Arial, Helvetica, sans-serif;
     font-size: 10pt;
-    color: black;
-    background: white;
+    color: var(--sc-text-color);
+    background: var(--sc-bg-color);
+}
+
+input, select, textarea {
+    color: var(--sc-text-color);
```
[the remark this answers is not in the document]
elgiano: To be clearer, i mean:
```css
input, select, textarea {
    color: var(--color-input-fg);
    background: var(--color-input-bg)
}

:root {
   --color-input-fg: var(--sc-text-color);
   --color-input-bg: var(--sc-neutral-darker);
}
```

HelpSource/scdoc.css
```diff
@@ -611,17 +621,18 @@ ul.tree li {
...
 :target {
-    border-bottom: 6px solid rgba(255,255,0,0.4);
+    border-bottom: 6px solid var(--sc-bg-warning);
     padding-top:20px;
     margin-top:-20px;
 }
 
 a.undoc {
-    color: #88b;
+    color: var(--sc-text-color);
```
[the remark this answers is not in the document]
capital-G: This would be indeed a good idea but CSS does not allow for this (at least currently, see https://caniuse.com/mdn-css_types_color_color-mix )

I switched to `--sc-text-color-lighter` instead.

HelpSource/scdoc.css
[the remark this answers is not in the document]
capital-G: This is indeed a good idea, added it.

I think we should generally re-structure the DOM and CSS in another PR in order to allow for a wide variety of documentation styling which go beyond mere coloring.

HelpSource/scdoc.css
capital-G: yes, do you have a preference which color to pick?

HelpSource/scdoc.css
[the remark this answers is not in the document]
elgiano: How about:
```css
a.undoc {
   color: var(--sc-link-color);
   opacity: 50%;
}
```

HelpSource/scdoc.css
elgiano: just two different shades of the main text color, for example (didn't test it though):
```css
.doclink { color: var(--sc-neutral-darker) }
.doclinka { color: var(--sc-text-color-lighter) }
```

HelpSource/scdoc.css
```diff
@@ -633,17 +649,17 @@ div#search_box {
 }
 
 #search_input {
-    border: 1px solid #aaa;
-    background: #eee;
+    border: 1px solid var(--sc-neutral);
+    background: var(--sc-bg-color);
```
elgiano, line 653: Now this background should either be `var(--color-input-bg)`, or removed (to inherit the same value)

HelpSource/scdoc.css
```diff
@@ -611,18 +626,20 @@ ul.tree li {
...
 :target {
-    border-bottom: 6px solid rgba(255,255,0,0.4);
+    border-bottom: 6px solid var(--sc-bg-warning);
     padding-top:20px;
     margin-top:-20px;
 }
 
 a.undoc {
-    color: #88b;
+    color: var(--sc-text-color-lighter);
```
elgiano, line 640: now that we set opacity to 50%, could this be the same color as the other links? removing the color definition would be enough to inherit their color.

capital-G:
Hey, thanks for all your input and feedback! I worked on this some years ago so I am a bit rusty on all the CSS classes within the docs, sorry.

I like the idea of providing a good interface for colors. I personally find tailwind too verbose, but it can serve as a tested and accepted reference here.
But I am not sure if we should do this without a namespace (so `--color-fg`) or with a namespace (`--sc-color-fg`). I always prefer namespaces, but I also don't know if the lack of a namespace could interfere with something outside our scope, e.g. through an iframe?

elgiano:
Thank you for bearing with me through all this :D

There is one more unresolved conversation about the extension badge, that maybe you have lost in this long series of back and forth, you'll find it if you scroll up, it's the last one left open from me.

About namespaces: css variables are scoped to the document they are defined in, and they don't pierce through iframes. For this reason, I don't think we need to namespace them with anything else than their type.

capital-G:
@elgiano something like this? I Want to make sure before committing. Be aware, I also changed some positions so it makes more sense with the new syntax.

image

HelpSource/scdoc.css
```diff
@@ -1,3 +1,8 @@
+:root {
+    --color-input-fg: var(--sc-text-color);
+    --color-input-bg: var(--sc-neutral-darker);
+ }
```
elgiano, line 4: Shouldn't these definitions be in theme files instead of here? I understand that it's to avoid repetitions, but it would also be nice to have all definitions in one place.

HelpSource/scdoc.css
```diff
@@ -514,12 +524,12 @@ td p {
...
     background: #fff;
-    border-right: 1px solid #bbb;
+    border-right: 1px solid var(--color-fg-100);
 }
 
 #sidetoc a:hover {
     text-decoration: none;
-    color: black;
+    color: var(--color-bg);
 }
```
elgiano, line 533: I think we can safely remove `#sidetoc` and `#sidetoc a:hover` rules, I can't find any reference to these ids anymore, either in HelpSource or in my compiled helpTargetDir. 
I would delete them all together, but if you want to be more conservative you could just comment them out with a note saying something like "unused?".

The reason I looked into this is because #sidetoc has an hardcoded background color (#fff)

HelpSource/scdoc.css
```diff
@@ -738,17 +748,17 @@ div.met_subclasses a.seemore {
...
 #method_note {
     font-size: 9pt;
-    color: #777;
+    color: var(--color-fg-200);
     border-top: 1px solid #ddd;
```
elgiano, line 752: shall we replace this border-top #ddd color with var(--color-fg-100)?

HelpSource/scdoc.css
capital-G: If it is not used anymore then we should remove it.

capital-G:
found another hardcoded border which I replaced with a color from the theme

https://github.com/supercollider/supercollider/pull/6095/commits/d29e7a7599c75c40af7f49c3ead168158685edd8#diff-df72372f1e55055310e9d066f51b110bbbbed7974ae1b0f7c7d403d1410888beR279

elgiano:
Great, thanks so much for all the work, I don't have any other structural changes to propose. Now we need to double check all themes for readability issues and eventually change either the fg shades, or their assignment. I saw already some problems with inputs and lighter text shades in almost all of them, but didn't have time yet to go through all details. Could you have a look?

Then the only thing left is documentation: we could add a paragraph in `HelpSource/Classes/SCDocHTMLRenderer.schelp`, after `CSS styling`. But I think there is an issue there, because it's hard to find (e.g. if you search for css it doesn't show up). So maybe we could add a new file at `Reference/SCDocStyling.html`, a bit like SCDocSyntax, and put the word css in its summary so that it would show up? I could also write a draft of this file if that's good for you!

capital-G:
@elgiano I fixed some styling issues, mainly color of input and selection colors - I couldn't spot any difficult areas anymore.

I would be hesitant to include the CSS variables or documentation in SCDocs as anyone familiar with CSS styling will easily find the `theme/*.css` files and adding custom themes beyond `custom.css` is currently not possible due to hardcoded JS variables (which have to be hardcoded due to lack of disk access) as they would be deleted during an update.
I think this would be more interesting in case we rework the DOM and CSS structure to make themes possible which go beyond color-swapping.

elgiano:
I went through all themes and made some little changes, which I think improve readability. Try them out and let me know:
```diff
themes/classic.css
-    --color-fg-100: #ccc;
+    --color-fg-100: #eee;

themes/dark.css
-    --color-fg-200: #aaa;
-    --color-fg-100: #555;
+    --color-fg-200: #777;
+    --color-fg-100: #333;
 
themes/default.css
-    --color-fg-200: #aaa;
-    --color-fg-100: #ccc;
+    --color-fg-200: #999;
+    --color-fg-100: #f0f0f0;

themes/dracula.css
-    --color-fg-100: #555;
+    --color-fg-100: #333;

themes/solarizedDark.css
-    --color-fg-200: #555;
-    --color-fg-100: #555;
+    --color-fg-200: #4f6e76;
+    --color-fg-100: #054252;

themes/solarizedLight.css
-    --color-cm-string: #93a1a1;
+    --color-cm-string: #7b8d8d;

-    --color-warning-bg: var(--color-cm-evaluated);
+    --color-warning-bg: #dbecf1;

-    --color-fg-100: #ccc;
+    --color-fg-100: #eee;
 ```

File: HelpSource/themes/solarizedLight.css [hunk 20]
```diff
@@ -0,0 +1,42 @@
...
+
+    /* derived colors
+    in future use `color-mix` to calculate them which allows for fewer variables
+    but currently the qt browser does not support it, see
+    https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix
+    */
+    --color-bg: var(--color-cm-background);
+
+    --color-warning-bg: var(--color-cm-evaluated);
+    --color-warning-fg: #dbecf1;
```
elgiano, line 27: Here there is a mistake, I meant:
    --color-warning-bg: #dbecf1;
    --color-warning-fg: var(--color-cm-keyword);

capital-G:
> ok, but then maybe put #7b8d8d for links? They would become a little bit more visible and I think it's nice
> 
> Also, do you experience flashes (especially with dark themes) when navigate from a page to another? On my system looks like the theme takes a while to load. I'm looking into it

Thats to be expected because the default theme is loaded within the DOM and JS replaces the CSS file with the one stored in localStorage.
We could call this twice - as soon as possible and once after the DOM has been loaded fully (is there a case where the 1st call would not be sufficient?).

```diff
diff --git a/HelpSource/scdoc.js b/HelpSource/scdoc.js
index 13681b623..9fe05a993 100644
--- a/HelpSource/scdoc.js
+++ b/HelpSource/scdoc.js
@@ -260,6 +260,7 @@ window.addEventListener('DOMContentLoaded', () => {
     applyTheme();
 });
 
+applyTheme();
 
 // Set up a QWebChannel for communicating with C++ IDE objects. The main app publishes a handle
 // to IDE functionality at "IDE" which is made globally available here after the page and

```

More importantly: Does your `localStorage` of the IDE survive a restart of the IDE? It looks like mine doesn't - this would make the current setup not really sufficient IMO.

elgiano:
> Thats to be expected because the default theme is loaded within the DOM and JS replaces the CSS file with the one stored in localStorage.
We could call this twice - as soon as possible and once after the DOM has been loaded fully (is there a case where the 1st call would not be sufficient?).

I think once as soon as possible will be sufficient. Also, I think we need to load scdoc.js before we try to load docmap.js, it's a big file, can make things slower.

I made a patch with all the required changes (also the link color above), if you want to try it: https://gist.github.com/elgiano/c4e56a98a28889cd7b2140c97260c800

elgiano:
Thanks for the fast reaction, I updated the patch after your last commits, because we still need to move some <links> around (and add a missing </head> since we are here):
```diff
diff --git a/HelpSource/BrokenLink.html b/HelpSource/BrokenLink.html
index 2169e6d3e..ceda7f063 100644
--- a/HelpSource/BrokenLink.html
+++ b/HelpSource/BrokenLink.html
@@ -4,8 +4,8 @@
 <head>
     <title>Broken link</title>
     <link rel='stylesheet' href='./scdoc.css' type='text/css' />
-    <link rel='stylesheet' href='./frontend.css' type='text/css' />
     <link rel='stylesheet' id='scdoc-theme' href='./themes/default.css' type='text/css' />
+    <link rel='stylesheet' href='./frontend.css' type='text/css' />
     <link rel='stylesheet' href='./custom.css' type='text/css' />
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta http-equiv='Content-Type' content='text/html; charset=UTF-8' />
diff --git a/HelpSource/Browse.html b/HelpSource/Browse.html
index 13bdeaa7e..7bb609262 100644
--- a/HelpSource/Browse.html
+++ b/HelpSource/Browse.html
@@ -6,8 +6,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta http-equiv='Content-Type' content='text/html; charset=UTF-8' />
     <link rel='stylesheet' href='./scdoc.css' type='text/css' />
-    <link rel='stylesheet' href='./frontend.css' type='text/css' />
     <link rel='stylesheet' id='scdoc-theme' href='./themes/default.css' type='text/css' />
+    <link rel='stylesheet' href='./frontend.css' type='text/css' />
     <link rel='stylesheet' href='./custom.css' type='text/css' />
     <link rel='stylesheet' href='./browse.css' type='text/css' />
     <script src="lib/jquery.min.js"></script>
@@ -15,11 +15,11 @@
         var scdoc_title = "Document Browser";
         var scdoc_sc_version = "";
     </script>
-    <script src="docmap.js" type="text/javascript"></script>
     <script src="scdoc.js" type="text/javascript"></script>
+    <script src="docmap.js" type="text/javascript"></script>
     <script src="browse.js"></script>
     <script src="frontend.js"></script>
-
+</head>
 <body>
     <noscript>
         <!--<meta http-equiv="refresh" content="3; URL=Overviews/Categories.html">
diff --git a/HelpSource/Overviews/Classes.html b/HelpSource/Overviews/Classes.html
index 8960932cf..1137d2ca1 100644
--- a/HelpSource/Overviews/Classes.html
+++ b/HelpSource/Overviews/Classes.html
@@ -5,16 +5,16 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta http-equiv='Content-Type' content='text/html; charset=UTF-8' />
     <link rel='stylesheet' href='../scdoc.css' type='text/css' />
-    <link rel='stylesheet' href='../frontend.css' type='text/css' />
     <link rel='stylesheet' id='scdoc-theme' href='../themes/default.css' type='text/css' />
+    <link rel='stylesheet' href='../frontend.css' type='text/css' />
     <link rel='stylesheet' href='../custom.css' type='text/css' />
     <script src="../lib/jquery.min.js"></script>
     <script>
         var scdoc_title = "Classes";
         var scdoc_sc_version = "";
     </script>
-    <script src="../docmap.js" type="text/javascript"></script>
     <script src="../scdoc.js" type="text/javascript"></script>
+    <script src="../docmap.js" type="text/javascript"></script>
     <script src="../frontend.js" type="text/javascript"></script>
 <noscript>
 <p>The class index needs JavaScript.
diff --git a/HelpSource/Overviews/Documents.html b/HelpSource/Overviews/Documents.html
index 2a8744eec..f4d2c3e24 100644
--- a/HelpSource/Overviews/Documents.html
+++ b/HelpSource/Overviews/Documents.html
@@ -5,16 +5,16 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta http-equiv='Content-Type' content='text/html; charset=UTF-8' />
     <link rel='stylesheet' href='../scdoc.css' type='text/css' />
-    <link rel='stylesheet' href='../frontend.css' type='text/css' />
     <link rel='stylesheet' id='scdoc-theme' href='../themes/default.css' type='text/css' />
+    <link rel='stylesheet' href='../frontend.css' type='text/css' />
     <link rel='stylesheet' href='../custom.css' type='text/css' />
     <script src="../lib/jquery.min.js"></script>
     <script>
         var scdoc_title = "Documents";
         var scdoc_sc_version = "";
     </script>
-    <script src="../docmap.js" type="text/javascript"></script>
     <script src="../scdoc.js" type="text/javascript"></script>
+    <script src="../docmap.js" type="text/javascript"></script>
     <script src="../frontend.js" type="text/javascript"></script>
 <noscript>
 <p>The document index needs JavaScript.
diff --git a/HelpSource/Overviews/Methods.html b/HelpSource/Overviews/Methods.html
index 88c2b9252..8ee2309a4 100644
--- a/HelpSource/Overviews/Methods.html
+++ b/HelpSource/Overviews/Methods.html
@@ -5,16 +5,16 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta http-equiv='Content-Type' content='text/html; charset=UTF-8' />
     <link rel='stylesheet' href='../scdoc.css' type='text/css' />
-    <link rel='stylesheet' href='../frontend.css' type='text/css' />
     <link rel='stylesheet' id='scdoc-theme' href='../themes/default.css' type='text/css' />
+    <link rel='stylesheet' href='../frontend.css' type='text/css' />
     <link rel='stylesheet' href='../custom.css' type='text/css' />
     <script src="../lib/jquery.min.js"></script>
     <script>
         var scdoc_title = "Methods";
         var scdoc_sc_version = "";
     </script>
-    <script src="../docmap.js" type="text/javascript"></script>
     <script src="../scdoc.js" type="text/javascript"></script>
+    <script src="../docmap.js" type="text/javascript"></script>
     <script src="../frontend.js" type="text/javascript"></script>
 <noscript>
 <p>The method index needs JavaScript.
diff --git a/HelpSource/Search.html b/HelpSource/Search.html
index 0c0c5ed95..4a5d6d70e 100644
--- a/HelpSource/Search.html
+++ b/HelpSource/Search.html
@@ -14,8 +14,8 @@
         var scdoc_title = "Search";
         var scdoc_sc_version = "";
     </script>
-    <script src="docmap.js" type="text/javascript"></script>
     <script src="scdoc.js" type="text/javascript"></script>
+    <script src="docmap.js" type="text/javascript"></script>
     <script src="search.js"></script>
     <script src="frontend.js"></script>
 </head>
diff --git a/HelpSource/scdoc.js b/HelpSource/scdoc.js
index ee65c45ab..2175576a7 100644
--- a/HelpSource/scdoc.js
+++ b/HelpSource/scdoc.js
@@ -256,8 +256,8 @@ function applyTheme(theme=null) {
     cssThemeTag.href = `${scriptLocation.split('/').slice(0, -1).join("/")}/themes/${themeName}.css`;
 }
 
-// scdoc.js is placed after the `theme` CSS to be changed
-// in the DOM, so it is safe to call it here.
+// scdoc.js is loaded after css themes so it's safe to call it here:
+// run immediately, don't wait for document load, to prevent "unstyled flash"
 applyTheme();
 
 // Set up a QWebChannel for communicating with C++ IDE objects. The main app publishes a handle
diff --git a/SCClassLibrary/SCDoc/SCDocRenderer.sc b/SCClassLibrary/SCDoc/SCDocRenderer.sc
index 8dd495253..bfdf4ba96 100644
--- a/SCClassLibrary/SCDoc/SCDocRenderer.sc
+++ b/SCClassLibrary/SCDoc/SCDocRenderer.sc
@@ -233,8 +233,8 @@ SCDocHTMLRenderer {
 		<< "<link rel='stylesheet' href='" << baseDir << "/scdoc.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/codemirror.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/editor.css' type='text/css' />\n"
-		<< "<link rel='stylesheet' href='" << baseDir << "/frontend.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' id='scdoc-theme' href='" << baseDir << "/themes/default.css' type='text/css' />\n"
+		<< "<link rel='stylesheet' href='" << baseDir << "/frontend.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/custom.css' type='text/css' />\n"
 		<< "<meta name='viewport' content='width=device-width, initial-scale=1'>\n"
 		<< "<meta http-equiv='Content-Type' content='text/html; charset=UTF-8' />\n"
```

elgiano:
I've written some documentation, here is a patch: https://gist.github.com/elgiano/c4e56a98a28889cd7b2140c97260c800

Basically I've removed the CSS Styling paragraph from SCDocHTMLRenderer, and added a link to a new document in Reference/SCDocStyling:

```
title:: SCDoc Styling
summary:: Help system CSS styles customization and themes
categories:: HelpSystem
related:: Guides/WritingHelp, Classes/SCDoc, Classes/SCDocHTMLRenderer

Documentation rendered HTML reads the global style from teletype::scdoc.css::,
but also reads teletype::frontend.css:: and teletype::custom.css:: (in that
order) if available, to enable specific frontends and users to override the CSS.

So to customise the CSS, the user can create a teletype::custom.css:: in their
link::Classes/SCDoc#*helpTargetDir:: or at the root of any HelpSource directory
(for example in teletype::YourExtension/HelpSource/custom.css::).

code::
// open HelpTargetDir to see these css files
SCDoc.helpTargetDir.openOS
::

section::Colors

Colors are defined as CSS variables, so that to change any color, it's
sufficient for a user to override the relevant variable in their
teletype::custom.css::. These variables are then used to set any color in all
other css files of the documentation.

Specific variables are not reported here, please refer to the default CSS theme
for all their names and default values.

code::
// open default theme
(SCDoc.helpTargetDir +/+ "themes" +/+ "default.css").openDocument

// copy default theme to 'custom.css' as a starting point
File.copy(
	SCDoc.helpTargetDir +/+ "themes" +/+ "default.css",
	SCDoc.helpTargetDir +/+ "custom.css"
)
::

subsection::Themes

CSS themes distributed in teletype::HelpTargetDir/themes:: are
provided only to match ScIDE's UI themes. There is currently no
implementation for modifying CSS themes dynamically or to add custom ones, so
for any customization it's recommended to write a teletype::custom.css:: file as
explained above
```

What do you think?

capital-G:
thanks for writing the text @elgiano - i think it gives a good introduction on the topic and added it via a commit.

I'll update the top post one more time and I think we are good to go then? :)

elgiano (approved):
This PR is good for me:
- it defines all colors as variables so that they can be customized
- it adds css themes matching scide UI and a simple mechanism for the user to switch them
- themes are readable
- it relies on localStorage, which although wasn't really used before, it appears to be the simplest and most widely used solution for persisting user choice in terms of css themes.
- if nothing works, or javascript is disabled (unrealistic, as without js many other docs features don't work anyways), it defaults to the previous default theme

my only perplexity in terms of maintainance is whether ide themes should be isolated, managed and offered only by the ide. However, these css themes will probably require the same maintainance as the ide themes, which were never changed since 2018/2019.

So, all in all, I think this PR should be merged.

HelpSource/themes/classic.css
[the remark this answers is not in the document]
elgiano: I think we got a bit confused on this one? Or at least I got confused:
didn't we decide to remove --color-extension-bg and --color-extension-fg, and put the default values directly in scdoc.css? like this
```css
.extension-indicator-ctr {
    background-color: #601c8b;
    color: white;
 }
```
Now I've opened an extension page in a non-default theme and I have seen a strange color, so I checked and we still have the variables and different values for every theme. I would say let's remove them all from themes, and put back default values directly in scdoc.css. Then as we said above, if someone wants to customize this they can write a simple rule for `.extension-indicator-ctr` in their `custom.css`.

elgiano:
@capital-G  I've just found a minor thing regarding `.extension-indicator-ctr` that we might have lost in the middle of things, could you fix it please? I reopened the conversation above

HelpSource/editor.css
```diff
@@ -6,7 +6,7 @@ textarea.editor {
     height: auto;
     padding: 0 1em;
     margin-top: 1em;
-    border-left: 2px solid #e1e1e1;
+    border-left: 2px solid var(--color-fg-100);
     font-size: 1em;
     line-height: normal;
```
elgiano, line 11: and one more thing: could you add 
```css
background: var(--color-cm-background);
```
here please? I know in your themes code background is the same as all the other backgrounds, but it would make sense to see it applied here in case somebody customizes it with two different colors

HelpSource/editor.css
```diff
@@ -6,7 +6,7 @@ textarea.editor {
     height: auto;
     padding: 0 1em;
     margin-top: 1em;
-    border-left: 2px solid #e1e1e1;
+    border-left: 2px solid var(--color-fg-100);
```
elgiano, line 9: also, either this border-left is `var(--color-cm-border)`, or we remove the variable cm-border, it's not used anywhere else

HelpSource/scdoc.js
```diff
@@ -190,11 +190,76 @@ function fixTOC() {
...
+    "dark",
+    "default",
+    "dracula",
+    "solarizedDark",
+    "solarizedLight"
+];
+
+function buildThemeSwitcher() {
+    create_menubar_item("Theme \u25bc", "#", function (a, li) {
+        var indexes_menu = $("<div>", { class: "submenu" }).hide()
```
elgiano, line 214: minor thing, but could you rename this to themes_menu instead of indexes_menu?

capital-G: good catch - I decided to name it `themesMenu` which is not in style of `scdoc.js` but is more consistent with JS. I hope we can soon refactor the whole JS code nonetheless.

HelpSource/themes/classic.css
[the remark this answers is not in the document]
capital-G: I deliberately put both colors as var now because I think it would be strange if every color is variable but the extension badge banner is static, so I decided to keep it variable and use some colors from the theme to customize it accordingly.

Let me know what you think about it or if we should really remove it.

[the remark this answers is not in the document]
elgiano: For me it would be ok either to ship the same "extension badge" style for all themes (aka delete the variables), or to touch up the current colors with this patch: https://gist.github.com/elgiano/c4e56a98a28889cd7b2140c97260c800

In the patch I use --color-link for extension background and white for text everywhere, to keep it in line with the original design, except in dracula.css, where white on light green is not readable, and I left your choice of --color-cm-char (with white text)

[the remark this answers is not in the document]
capital-G: Thx for your suggestions, I applied them in 0737d5e3f09f54bd3c5f580853402464031da77a

capital-G:
@JordanHendersonMusic can you review it?

HelpSource/Overviews/Documents.html
```diff
@@ -5,15 +5,16 @@
...
     <link rel='stylesheet' href='../frontend.css' type='text/css' />
     <link rel='stylesheet' href='../custom.css' type='text/css' />
     <script src="../lib/jquery.min.js"></script>
     <script>
         var scdoc_title = "Documents";
         var scdoc_sc_version = "";
     </script>
-    <script src="../docmap.js" type="text/javascript"></script>
     <script src="../scdoc.js" type="text/javascript"></script>
+    <script src="../docmap.js" type="text/javascript"></script>
```
JordanHendersonMusic, line 17: Why is it necessary to load these files in this order now? Should they be dependent on each other?

HelpSource/Reference/SCDocStyling.schelp
```diff
@@ -0,0 +1,46 @@
...
+::
+
+section::Colors
+
+Colors are defined as CSS variables, so that to change any color, it's
+sufficient for a user to override the relevant variable in their
+teletype::custom.css::. These variables are then used to set any color in all
+other css files of the documentation.
+
+Specific variables are not reported here, please refer to the default CSS theme
```
JordanHendersonMusic, line 26: Would it make sense to refer to the readme.md file around here?

HelpSource/Reference/SCDocStyling.schelp
```diff
@@ -0,0 +1,46 @@
+title:: SCDoc Styling
+summary:: Help system CSS styles customization and themes
+categories:: HelpSystem
+related:: Guides/WritingHelp, Classes/SCDoc, Classes/SCDocHTMLRenderer
+
+Documentation rendered HTML reads the global style from teletype::scdoc.css::,
+but also reads teletype::frontend.css:: and teletype::custom.css:: (in that
+order) if available, to enable specific frontends and users to override the CSS.
```
JordanHendersonMusic, line 8: This sentence is a bit complicated. New sentence for 'To enabled ...' maybe? Then you can remove the brackets.

JordanHendersonMusic (commented):
This looks great thanks! How the colours are extracted into variables seems reasonable. 

If the names of these need to change in future, or more variables added, would it be easy to keep backwards compatibly with older themes?

Just a couple of things in the docs, and one question about dependencies across JavaScript files.

HelpSource/Reference/SCDocStyling.schelp
capital-G: I think it is sufficient if the readme becomes apparent when navigating to the folder where the default CSS theme is located? Otherwise there could be a question why the `README` is not included here.

If you think it should be mentioned I can include it.

HelpSource/Reference/SCDocStyling.schelp
capital-G: So something like this?

> Documentation rendered HTML reads the global style from teletype::scdoc.css::,
but also reads teletype::frontend.css:: and teletype::custom.css:: (in that
order) if available.
> This is to enable specific frontends and allow users to override the CSS.

HelpSource/Reference/SCDocStyling.schelp
JordanHendersonMusic: Since you can make them in extensions, there ought to be some reference to it in the help files as the readme won't be next to the file. If you wouldn't mind?

capital-G: Added in https://github.com/supercollider/supercollider/pull/6095/commits/35f287b4badcab6ff9e00be3ecfb7d8a51230858

capital-G:
> If the names of these need to change in future, or more variables added, would it be easy to keep backwards compatibly with older themes?

In such a case we could add a comparability layer which uses the current variables as a default, if declared, for future variables, see https://developer.mozilla.org/en-US/docs/Web/CSS/var#using_a_custom_property_as_a_fallback

```css
:root {
  --color-future: var(var(--color-old), "bright");
}
```

dantepfer:
@capital-G just curious about this — I see this was merged, but the help in the latest dev version still doesn't adapt to the current theme. Do I need to do something to enable it?

PR #6096, opened 2023-09-26
Update to clang 14 via pre-commit

capital-G:
## Purpose and Motivation

Clang8 has been [deprecated(?) and is currently not possible to build under the most recent macOS version Ventura](https://github.com/Homebrew/homebrew-core/commit/58b0ce1e17c7cbde7fccfcd81867c485d24866e5) which makes it harder and harder to install on up-to-date machines, yet it is a mandatory tool when contributing C++ code (see e.g. #6095). Also it is necessary to install specific versions of clang-format as the linting/styling guidelines change within clang versions. 

Instead just updating clang and using [custom python code](https://github.com/supercollider/supercollider/blob/develop/tools/clang-format.py) to control/verify the formatting I would suggest to instead switch to [pre-commit](https://pre-commit.com/) which is a management tool for git hooks and provides the ability to

* pin linting-tools to specific versions via a file and also manage their installation
* add multiple automatic checks before commiting

[`pre-commit`](https://pre-commit.com/) is an established tool from the python eco-system, but extended already to a variety of other languages and could be seen as a package manager for [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).
Pre-commit allows to pin the clang dependency via a [compiled python "wheel"](https://github.com/ssciwr/clang-format-wheel) so that one becomes independent of the package manager of the OS and instead can rely on the Python package manager.
As the current formatting script also relies on Python, using `pre-commit` would not introduce a new dependency but instead would allow to provide a fixed clang version via code.

With `pre-commit` active (running `pre-commit install`), every commit-delta gets checked by the script and if the formatter script does not pass it will apply proper formatting and one need to stage the changes in order to successfully make a commit.
Installing `pre-commit` is optional, but the CI would run the same script and verify that the `pre-commit` pipeline passes, which verifies that that all files are linted as it should be.
The dependencies/scripts are tracked by a `.pre-commit-config.yaml` which allows to have branches with different clang-format version co-exist.

If ever the `pre-commit` workflow should fail it would still be possible to revert to the old-school formatting, enforcing the user to install the proper clang version.
But pre-commit is used by 145k projects according to github.

Why clang14, although there is clang16 already?

* Debian stable (bookworm) has it https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=clang
* Ubuntu LTS has it https://packages.ubuntu.com/search?suite=jammy&searchon=names&keywords=clang
* Windows builds are availbale via github https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.0
* But - it is not available via brew: https://formulae.brew.sh/formula/clang-format#default

But in the end this should not be a big concern as installing clang-format would only be necessary if the python-wheel is not existing for the plattform, but since v10.0 there are wheels for macOS, windows and linux, see https://pypi.org/simple/clang-format/

This is more of a pro-active proposal, initiated by its urgent necessity due to the current missing macOS support.
I already proposed something similar via #5804 but found the pre-commit-wheel way easier for ocassional usage.

If this gets positive feedback or any kind of approval I would continue to modify the existing documentation regarding formatting, adjust the CI pipeline and maybe try to figure out the clang-format setting to suppress the noise like

image

Happy for feedback :)
Sorry to touch this delicate topic, but I think it needs some resolvement and maybe an improvement :)

## Types of changes

- Documentation
- Bug fix
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
Thanks @capital-G !

Thanks for picking up the formatting issue. We definitely need to update our formatting tools. 
However, I think there are 3 separate issues that I believe should be addressed in separate threads (Issues or PRs):
1. which version of clang-format to use (I don't disagree with your choice, but I think this is something that should be agreed upon with other devs)
    - this might need to be a separate issue opened first, or part of the formatting PR (see `2.`)
2. format existing codebase with agreed on clang-format version
    - we should consider the process when the previous (original) formatting happened, as there was extensive discussion on the procedure (see [old dev list post](https://listarc.cal.bham.ac.uk/lists/sc-dev-2019/msg58957.html), also see formatting PRs #4425, #4426)
    - we should consider whether we create a tag for this reformat to help with git history searching
    - I _think_ this PR should also include revision of the formatting [tools](https://github.com/supercollider/supercollider/tree/develop/tools) we have in the repository, as well as docs (including the wiki, particularly [here](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions#requirements))
       - though this might overlap with `3.`, so maybe this should be moved to a separate PR? (wiki changes are of course not done through PRs, but I'm mentioning them since they should happen around the same time)
     - in this PR som `.js` files are also formatted, is this intentional? I thought that we only cormatted Cpp files before. If we want to format js files and we have not before, it should be discussed in this PR as well
     - Finally, doing the reformat will cause pains for all open PRs that touch Cpp files (not that many currently, but still!); we should look into this and merge/close Cpp PRs when possible before running the reformat
3. consider using git hooks
    - again, I'm not strictly opposed to it, but as it is adding a (soft) external dependency, I'd like this to be discussed separately

capital-G:
Thanks for your feedback @dyfer 

> Thanks for picking up the formatting issue. We definitely need to update our formatting tools. However, I think there are 3 separate issues that I believe should be addressed in separate threads (Issues or PRs):

I agree with this, as this was mainly a proposition on a "how it could be" state if I would do it on my own.
But at the same time this makes the transition really easy (as clang-format gets dropped as a system dependency) and if all 3 steps are considered worth doing we could just do them at once because it makes it much easier for everyone I believe.

>   1. which version of clang-format to use (I don't disagree with your choice, but I think this is something that should be agreed upon with other devs)

I must admit that I don't have any experience with clang-format, therefore the version was chosen by an uneducated guess - but I hope we have some experienced clang people around here which can give some advice?

> 2. format existing codebase with agreed on clang-format version
>        
>  * we should consider the process when the previous (original) formatting happened, as there was extensive discussion on the procedure (see [old dev list post](https://listarc.cal.bham.ac.uk/lists/sc-dev-2019/msg58957.html), also see formatting PRs [Clang format develop #4425](https://github.com/supercollider/supercollider/pull/4425), [Clang format 3.10 #4426](https://github.com/supercollider/supercollider/pull/4426))
>        * we should consider whether we create a tag for this reformat to help with git history searching
>        * I _think_ that PR should also include revision of the formatting [tools](https://github.com/supercollider/supercollider/tree/develop/tools) we have in the repository, as well as GHA checks, and the docs (including the wiki, particularly [here](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions#requirements))
>          
>  * though this might overlap with `3.`, so maybe this should be moved to a separate PR? (wiki changes are of course not done through PRs, but I'm mentioning them since they should happen around the same time)
>  * in this present PR some `.js` files are also formatted, is this intentional? I thought that we only formatted Cpp files before. If we want to format js files and we have not before, it should be discussed in this PR as well
>  * Finally, doing the reformat will cause pains for all open PRs that touch Cpp files (not that many currently, but still!); we should look into this and merge/close Cpp PRs when possible before running the reformat

If a clang version has been decided I would go through the changelog of clang-format to figure out how to turn "off" the new formatting rules so the formatting noise is kept at a minimum (and of course throw away the current "intermediate" commits). I hope clang-format can be tweaked in such a way that essentially no file gets modified via its config file - this gives benefits for git history as well as any pending PRs.

At the same time there are changes which improve readability, for example

https://github.com/supercollider/supercollider/blob/ebaed8bd07ed1209ec51c7bee23682c580ab564c/common/SC_AllocPool.h#L166-L178

vs 

https://github.com/supercollider/supercollider/blob/6bd1ed55bae1c00dc0a46ecd3a28e254622253b1/common/SC_AllocPool.h#L166-L177

 but I would not consider worth tainting the git history for it.

I would also turn off the JS formatting for sure and adjust the CI/CD and any docs after a version has been decided.


>  3. consider using git hooks
>        
>  * again, I'm not strictly opposed to it, but as it is adding a (soft) external dependency, I'd like this to be discussed separately

Understandable - it is always really worth to discuss adding any kind of dependency to a project. Yet I found git hooks/pre-commit a real time saver when working on multiple projects as it will simply pull the necessary dependencies to run all code-checks and also apply them, so I don't have to care about any "repository" rules and there is no discussion about "how to apply formatting" and no "fix linting" commits.

Also it moves the current `clang-format + python` dependency to just `python` as python can take care of the `clang-format` dependency.

There are also some other interesting pre-commit hooks https://pre-commit.com/hooks.html - yet it is always optional to install them - if one knows how to format according to clang the `pre-commit` hook will not be necessary, but the CI would check if the commit passes the `pre-commit` checks.
One cool thing about `pre-commit` is that it only runs on the staged files - but enough of my `pre-commit` pitch.

What would you consider some next steps @dyfer ? Maybe we can ping some people and find a common ground for the next steps?

I would propose as next steps

* [ ] Decide upon a clang-format version
* [ ] Decide if `pre-commit` is a good way to manage formatting/linting jobs
* [ ] Tweak https://github.com/supercollider/supercollider/blob/develop/.clang-format so the formatting noise gets reduced to essentially 0 / as few as possible which would be the highest priority

Ideally this should be done in a timely fashion (although without rushing!) as the necessary C++ formatting tool is not available anymore on the most recent macOS version.

capital-G:
Hey @dyfer - I adjusted the formatter and `.pre-commit` config and I managed to get the affected lines down from 17k to around 200 (mainly by not formatting the minified .js files^^ and also skipping the lexer files). Ideally it would be 0 lines but I believe there are also some inconsistencies within the code-base that only newer versions of clang-format pick-up and need to be decided in order to be deterministic.

If some C++ expert can provide some hints how to lower the noise even more I will gladly try them out. The config documentation for the formatter is available at https://clang.llvm.org/docs/ClangFormatStyleOptions.html

---

I know formatting has been/is a delicate topic but the current status of formatting is really bad - for example @telephon told me that due the formatting/linter rules he is incapable of touching C++ code and the current macOS version does not support clang-format-8 anymore, which makes complying with it really hard - not to say impossible w/o  Docker or a linux machine.

I know that introducing new development tools into world of SuperCollider should be well thought through, but I am sure that `.pre-commit` provides an easy and reliable setup not just for clang-format tasks but also other "cleaning" tasks to come. I use it since many years in multiple projects and is a big part of the Python community and beyond.

As mentioned prior, introducing pre-commit would allow to replace the dev-dependencies python + clang-format with just python as pre-commit can manage the necessary clang-format binary through its own package management system.

---

Regarding the current changes: The most common change are for example

https://github.com/supercollider/supercollider/blob/2db872ad2a42ff85726566149855ecdb60d65b77/server/scsynth/SC_Jack.cpp#L125-L129

becoming [I must admit I don't even understand the cpp syntax here?]

grafik

Some suggestions seem useful, for example the following

grafik

which was "massacred" by the clang-8 format commit and looked before like this

https://github.com/supercollider/supercollider/blob/9b270e05bd9c007dc74f7508be42ae1ada5b5011/common/SC_AllocPool.h#L183-L195

---

Are there some people to ask for opinion on this topic in order to get some progress on this issue? :)

capital-G:
Reduced it down to 20 files with 120 lines affected

dyfer:
We discussed this at the dev meeting on 2023-12-17.
Tentatively the devs thought that with the recent change the # of lines affected is acceptable to be pulled in as a single commit. Additionally we discussed whether doing this with a pre-commit hook is the change that should be part of this PR or not.
@mossheim since you worked on the "great reformat", do you have any input how to proceed with this one? Should we squash and merge, or would you suggest changes to this PR?
If we merge this, what should we do with the old python formatting code?
Finally, after merging this, we should update the documentation (wiki) about formatting.

mossheim:
> @mossheim since you worked on the "great reformat", do you have any input how to proceed with this one? Should we squash and merge, or would you suggest changes to this PR?
> If we merge this, what should we do with the old python formatting code?

I will try to review this in the coming weeks. Thank you both for working on this.

mossheim:
Some initial thoughts:

- Can you test whether there any difference between 14 and 16 on our codebase using your new .clang-format? Looking at https://clang.llvm.org/docs/ClangFormatStyleOptions.html the new features seem to mostly touch formatting of concepts and other things we don't use. If not we could say 14+ is ok and that the CI check uses 14.
- You can download packages for clang from the LLVM website so there is no need to be super limited about what is in various package managers.
- Any new formatting commits should use the same committee/author name as the previous clang format commit
- Don't add new tooling if the old way works, do that separately in a new PR
- SCDoc/SCDoc.tab.hpp is auto generated and should not be formatted: https://github.com/supercollider/supercollider/blob/db7eed2a17c361503dbc7f70a557874b6001e3cd/tools/clang-format.py#L128

capital-G:
Thanks for your feedback moss!

Let's follow your suggestion and separate the `clang-format` bump and the possible switch to `pre-commit` into two separate PRs.

The `clang-format` bump is now located in #6196 - I'll close this PR therefore.

PR #6097, opened 2023-09-28
Add new class method "localIPs" to NetAddr

prko:
Added a new method .localIPs to get all local IP addresses of the machine except "127.0.0.1".

## Purpose and Motivation

Please see the following and subsequent posts:
https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/5?u=prko

## Types of changes

- New feature

## To-do list
Needs testing on more Linux systems.
- [x] Code is tested
- [x] Some tests are passing
  only tested with Windows 11 Pro, macOS 12.7 and Ubuntu 22.04 ARM64
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Control/NetAddr.sc
+++ b/SCClassLibrary/Common/Control/NetAddr.sc
@@ -61,6 +61,43 @@ NetAddr {
 		^connections.copy;
 	}
 
+	*localIPs {
+		var winCMD, macCMD, linCMD, ips;
+
+		winCMD = {
+			"ipconfig | findstr /C:IPv4"
+			.unixCmdGetStdOut
+			.replace("   IPv4 Address. . . . . . . . . . . : ", "")
+		};
+
+		macCMD = {
+			(
+				"ifconfig | grep -Fv" + "inet 10."
+				.quote + "| grep" + "inet "
+				.quote + "| grep -Fv 127.0.0.1 | awk '{print $2}'"
+			).unixCmdGetStdOut
+		};
+
+		linCMD = {
+			(
+				"ip address | grep" + "inet ".quote +
+				"| grep -Fv 127.0.0.1 | awk '{print $2}' | cut -d '/' -f1"
+			)
+			// or: // ("ifconfig | grep" + "inet ".quote + "| grep -Fv 127.0.0.1 | awk '{print $2}'")
+			.unixCmdGetStdOut
+		};
+
+		ips = Platform.case(
+			\osx,       { macCMD.() },
+			\linux,     { linCMD.() },
+			\windows,   { winCMD.() }
+		);
+
+		ips = ips.split(Char.nl);
+		ips.removeAt(ips.size-1);
+		^ips
+	}
+
 	hostname_ { arg inHostname;
 		hostname = inHostname;
 		addr = inHostname.gethostbyname;


Spacechild1:
It seems that this implementation of `localIPs` just collects all local IP addresses, but what is the purpose? Typically, you would be only interested in the local IP address of the actual network interface that is used for outbound UDP connections. This can be done with some socket programming, as described here: https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/12?u=spacechild1

Also, keep in mind that `127.0.0.1` is not the only loopback address - it just happens to be the most popular one. See https://en.wikipedia.org/wiki/Localhost#Loopback.

prko:
> It seems that this implementation of localIPs just collects all local IP addresses, but what is the purpose? Typically, you would be only interested in the local IP address of the actual network interface that is used for outbound UDP connections. This can be done with some socket programming, as described here: https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/12?u=spacechild1

A few years ago I installed SC on 12 Raspberry Pi and controlled it from a MacBookPro. I made the code in the thread https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791?u=prko to simplify and reduce the time to set up the IP addresses used for UDP connections. The reason it collects all local IP addresses is due to advice from @dyfer:
https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/2?u=prko:

> >IMO the command to check for one’s IP should always return an array (excluding 127.0.0.1) - it’s not uncommon to have more than one interface (e.g. wireless and ethernet) and it’s not always clear which one you’d want to know about. Also, they should probably be listed in the order they are set in the OS
> > - in this spirit I believe that there should be a separate command to get all addresses, and a separate one to get it for a specific interface

Unfortunately, implementing what you describe in the post https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/12?u=spacechild1 is beyond my capabilities. Sorry... The information you have written is all Greek to me.

> Also, keep in mind that `127.0.0.1` is not the only loopback address - it just happens to be the most popular one. See https://en.wikipedia.org/wiki/Localhost#Loopback.

I see, but sclang does not accept `"::1"`:
```
n = NetAddr("localhost", 57110); // -> a NetAddr(127.0.0.1, 57110)
n.ip;                            // -> 127.0.0.1

m = NetAddr("127.0.0.1", 57110); // -> a NetAddr(127.0.0.1, 57110)
m.ip;                            // -> 127.0.0.1

l = NetAddr("::1", 57110);       // ERROR: Primitive '_GetHostByName' failed.
l.ip; 
```

I had not thought that this PR could be accepted, as described elsewhere in the same thread above. Another user thought this functionality was missing, and I made this PR inspired by his request.
I would be happy if you could implement such functionality in the better way you suggest, and my amateurish PR could be closed! 

I just wanted to bring it up for discussion because someone would suggest a better way and implement it professionally.

Spacechild1:
> I see, but sclang does not accept "::1":

`::1` is the (single) IPv6 loopback address, but sclang only supports IPv4. 

> I would be happy if you could implement such functionality in the better way you suggest, and my amateurish PR could be closed!

Your PR is not amateurish! I just converted it to a draft PR to make it clear that it is not to be merged (yet). I would prefer to keep it open as a point of reference.

It would be good to collect some concrete use cases to figure out what `localIPs` should actually do. After all, a new feature typically tries to solve some real word problems :)

Spacechild1:
Hey, check this out: https://github.com/supercollider/supercollider/pull/6153. This allows you get the local IP address for a specific network. If you are only connected to a single network (the typical case), the network address can be omitted, so you can just do `NetAddr.localIP`.

I am wondering if there would still be a need for `NetAddr.localIPs`. Do you have any use cases?

Funnily enough, I discovered that the `_MatchLangIP` primitive, used for `NetAddr.matchLangIP`, already contains the necessary code to list all network interface addresses! We would just need to extract the code, so it can be reused for an (imaginary) `_LocalIPs` primitive.

Either way, I do not think that we should filter out "127.0.0.1". If needed, this can be done with a single line of code: `NetAddr.localIPs.reject(_ == "127.0.0.1")`.

prko:
> I am wondering if there would still be a need for `NetAddr.localIPs`. Do you have any use cases?


Detecting multiple local IPs was suggested by @dyfer. I agree with him because I have experienced some cases where I needed to organise multiple IPs. 

1. When using multiple computers for a project, an IP connected to the Internet via WIFI and an IP connected to project computers via Ethernet should be managed separately. I have often experienced this when using multiple computers.

2. I used a product from Astro spatial audio: https://www.astroaudio.eu/. If I remember correctly, it needed two IPs: one for the sound device (audio driver) and one to control the spatial audio server. I had to use two Ethernet connections via two Thunderbolt-to-Ethernet adapters.

When doing things like this, I can of course get the IP addresses from the system preferences (or system settings or control panel) of the operating system. However, it would be helpful, because it can save time, if I could get these IPs in sclang to automatically assign the required IP from the array of detected IPs.

> I do not think that we should filter out.
I also think it would be better to return "127.0.0.1" in the array as well.

Thanks for your PR! If you implement ".localIPs" in your PR, I would like to close this PR. Otherwise, I would keep this PR alive until other developers close it.

Spacechild1:
> When using multiple computers for a project, an IP connected to the Internet via WIFI and an IP connected to project computers via Ethernet should be managed separately. I have often experienced this when using multiple computers.

But what is the issue exactly? When you are connected to multiple networks, the OS automatically sends the network packets to the appropriate interface based on the destination address. For example, if interface A has the address "192.168.0.X" and interface B has the router address "192.168.1.X", then a packet sent to "192.168.0.Y" will automatically sent through A and "192.168.1.Y" will be sent through B.

What would you want to *do* with the local IP addresses, once you obtained them? It would be great to see an actual real-life use case with (pseudo) code.

prko:
> What would you want to _do_ with the local IP addresses, once you obtained them? It would be great to see an actual real-life use case with (pseudo) code.
> 
> (To be honest, I cannot even think of a real-life use case for #6153 ...)

Two real-life cases of this PR and of #6153

1. Multi-directionally Synchronisable Digital Sheet Music:
The Construction of a WiFi-based Musical Score System

From 2013 to 2018, I made an interconnected score for musicians and me as a sound operator, and I developed it up to 4 players (a total of 5 performers including me). The following describes the core idea and the problems:

_This paper deals with the automatic synchronisation of electronic sheet music based on wireless networks in which the displayed section of sheet music, changed on any device, is automatically reflected on the screen of all other devices in the system when all or some of the members of a group use tablets as musical scores. In this study, we implemented a method using several tablets to control the shared screen of a server that displays sheet music and a method of autonomously updating the sheet music by reading the corresponding data in cloud-based libraries caused by section ID transmission; we then compared their characteristics with those in a related study. As a result, it is ascertained that the method using section ID allows the establishment of a fast and stable sheet music synchronisation system, even with inexpensive apparatuses. Using such a system in various classes, playing diverse compositions, improvising, etc., allows a reduction in the noise and inconvenience of turning paper, so that students, teachers, musicians, and audiences can be more immersed in the music. Moreover, this method can also be applied to various environments in which screen synchronisation is required._

I cannot paste the full codes, but I used a structure similar to the following to use an SCD file on each computer:
```
(
var win, mainComputerIP; 
win = Window();
mainComputerIP = "192.168.219.102";
StaticText(win, win.bounds.width@win.bounds.height)
.string_(
	if (mainComputerIP.matchRegexp(~getLocalIPs.().join))
	{ 
		"This is the main computer. You can control all SuperCollider instances running on the computers connected to the local netork." 
	} {
		"This is not main computer. You can only change the slides of scores on the screen."
	}
);
win.front;
)
```

2.  A loudspeaker ensemble with 15 percussion instruments 
https://youtu.be/FVat3NMPWYk?si=zp_mqrjEJyJc-6UC
For this project I used a MacBook Pro and 15 Raspberry Pi machines. 
- On each Raspberry Pi, SuperCollider is running and its audio output is connected to the transducer attached to a membrane (or two membranes) or the metal surface of each percussion instrument. 
- The MacBook Pro runs SuperCollider, Max and VNC clients to control each SuperCollider running on each Raspberry Pi.
- Visitors to the installation can control each membrane using their own smartphones (via a web browser) or the sensor connected to each Raspberry Pi.
To control each Raspberry Pi, I also used a similar structure as above:
```
(
{
	var ip_perc1, ip_perc2, detectInst, sig; 
	ip_perc1 = "192.168.219.103";
	ip_perc2 = "192.168.219.102";
	detectInst = { |which| which.matchRegexp(~getLocalIPs.().join) }; 
	sig = case
	{ detectInst.(ip_perc1) } {
		("This is ip_perc1: " + ip_perc1.postln;).postln;
		Decay.ar(Impulse.ar(1), 3, Impulse.kr(1))}
	{ detectInst.(ip_perc2) } {
		("This is ip_perc2: " + ip_perc2.postln;).postln;
		WhiteNoise.ar * Env.sine(2, 0.2).ar(Done.freeSelf)  
	};
	Pan2.ar(sig, 0, 0.1)
}.play
)
```
Would you accept these examples as real use cases for my PR?

Spacechild1:
Side note: actually, once we have `localIPs`, `matchLangIP` could be implemented purely in sclang:
```
matchLangIP { arg ipstring;
    if (ipstring == "127.0.0.1") { ^true };
    ^this.localIPs.includesEqual(ipstring);
}
```
I would just need to replace the existing `_MatchLangIP` primitive with `_LocalIPs`, keeping the existing code for retrieving the interfaces.

prko:
> Hmm... so you seem to use the local IP address as an identifier for a particular role. Usually, you would just _tell_ the client which role it has, e.g. by setting a variable in the startup script, reading a config file, with a command line argument, etc.

The usual way, if I understand what you mean, is more complex and time-consuming to code. 
It does not matter if you use Max, SuperCollider or something else.
The way I have found reduces time, typing and the number of files when working with more than two computers. Please think about what I need to do as a composer, engineer and performer. If one or more computers have problems, they should work by rebooting. Otherwise, I will not be able to rehearse and perform.

Again, the array of IPs was not my idea, and there is a user who strongly needs this functionality in the forum posts:
- https://listarc.cal.bham.ac.uk/lists/sc-users-2015/msg46783.html
- https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/2?u=prko
- https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/9?u=prko

muellmusik:
> Funnily enough, I discovered that the `_MatchLangIP` primitive, used for `NetAddr.matchLangIP`, already contains the necessary code to list all network interface addresses! We would just need to extract the code, so it can be reused for an (imaginary) `_LocalIPs` primitive.

Yes, that would be the best approach I think. Sorry, I always meant to do that after adding matchLangIP, and never got around to it.



> However, your last link contained an actual use case: a quick way to know your local IP, so that other users in a LAN can send messages to you. Of course, you can do that with the right command in your terminal or by visiting the right settings menu, but it's different on every OS and much more awkward than just calling `NetAddr.localIP`. I think you've sold me :)

I have found it comes up in network situations that it may make sense for a machine to have multiple network adapters (not necessarily for SC). This can lead to cases where you might need to test if an address matches your machine. ```matchLangP``` was added partly for that, but getting a list would definitely be useful. It's nice to be able to query the system without having to rely on cl tools (which can have drawbacks), at least for core functionality, and networking is core for SC in my opinion.



> The idea is the following:
> 
> 1. User `A` joins the LAN: broadcast `/ping <name> ...` message to all peers in the LAN
> 2. Existing peers: receive `/ping` from `A` and store its name + reply address. Reply with `/ping` + own name.
> 3. `A`: receive `/ping` messages from peers and store their names and reply addresses
> 
> Now everybody nows each other IP addresses and names.

Perhaps everyone here is aware, but this has very nice functionality to do exactly that: https://github.com/muellmusik/Utopia

Look at the ```Hail``` class.

Spacechild1:
@muellmusik I just added `NetAddr.localIPs` to https://github.com/supercollider/supercollider/pull/6153. `NetAddr.machLangIP` can now be implemented purely in sclang, using the output of `NetAddr.localIPs`.

prko:
@Spacechild1 
> Oh, there is a misunderstanding! "ping" can have several meanings:
> 
> In the narrow sense, it means a special network message that is used to test the reachability of a peer; this is what the `ping` command in the terminal does.
> 
> In a broad sense, it is any kind a message that basically says "Hi!" to another party, often expecting a reply. So what I actually meant was to send a broadcast message with `NetAddr("255.255.255.255", ~port)`. (NB: you have to enable broadcasting with `NetAddr.broadcastFlag = true`.)

Thank you for the kind explanation!

@muellmusik Thanks for your comments!

@Spacechild1 
> @muellmusik I just added NetAddr.localIPs to https://github.com/supercollider/supercollider/pull/6153. NetAddr.matchLangIP can now be implemented purely in sclang, using the output of NetAddr.localIPs.

Since you have implemented `.localIPs` in your PR https://github.com/supercollider/supercollider/pull/6153, I think it is better to close this PR. Thank you for your time and effort!

I can now reduce some pages of my own tutorial written in Korean, which is dedicated to explaining how to connect and control multiple sclang over networks, by removing the functions to get localIPs. This part will be more compact and the reader can concentrate more on the topic of this part!

Thanks again!

PR #6099, opened 2023-09-30
classlib: Fix Event's handling of 'lag' for MIDI events

jamshark70:
## Purpose and Motivation

Event's `\lag` key is documented as being expressed in seconds, and it's defined in `schedBundleArrayOnClock` as seconds for server-messaging events.

The default event prototype's `midi` event type, however, treats it as beats.

This PR mimics the schedBundleArrayOnClock logic for MIDI sending.

Technically this is a breaking change, so we don't know how much existing user code will be affected. We might guess "not much," however, because the MIDI event type was [implemented in 2005](https://github.com/supercollider/supercollider/blame/a36edf848fe3fb7f38a41253528482730839637f/SCClassLibrary/Common/Collections/Event.sc#L756), not substantially changed since then, and nobody noticed the bug until 2023. I take this to suggest that `\lag` vs MIDI has not been widely used (plausibly, perhaps never used). I think it's worth changing.

I'd like some consensus to be reached on that before I spend time writing a unit test, though. (If the PR won't be merged then it's not a productive use of my time to write a formal test.)

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [ ] Code is tested (not written a formal unit test yet; informal test passes)
- [ ] All tests are passing
- [ ] ~Updated documentation~ The change brings behavior in line with documentation, so no doc update needed

### DIFF
--- a/SCClassLibrary/Common/Collections/Event.sc
+++ b/SCClassLibrary/Common/Collections/Event.sc
@@ -811,13 +811,21 @@ Event : Environment {
 						bndl.do {|msgArgs, i|
 							var latency;
 
-							latency = i * strum + lag;
+							latency = i * strum;
 
-							if(latency == 0.0) {
-								midiout.performList(midicmd, msgArgs)
+							if(lag == 0.0) {
+								if(latency == 0.0) {
+									midiout.performList(midicmd, msgArgs)
+								} {
+									thisThread.clock.sched(latency, {
+										midiout.performList(midicmd, msgArgs);
+									})
+								}
 							} {
-								thisThread.clock.sched(latency, {
-									midiout.performList(midicmd, msgArgs);
+								SystemClock.sched(lag, {
+									thisThread.clock.sched(latency, {
+										midiout.performList(midicmd, msgArgs);
+									})
 								})
 							};
 							if(hasGate and: { midicmd === \noteOn }) {


telephon:
I think `timingOffset`, which is beat relative, should then also exist in midi event, like in note event.
Then there is an easy fix if someone has used `lag`, simply replace it by `timingOffset`.

jamshark70:
> I think timingOffset, which is beat relative, should then also exist in midi event, like in note event.

OK, this is done now. That led me to spot a bug: `SystemClock.sched(lag, { thisThread.clock.sched(...) })` is not valid because `thisThread` is no longer the thread playing the events. Fixed that now, and also renamed `latency` to `beatOffset` because it isn't at all related to server latency.

jamshark70:
Forgot to mention: I was concerned about unit testing on macOS and Windows because testing directly through MIDIOut would require a loopback, which (IIUC) requires user intervention to configure in these OSes. (It's easy in Linux by `.connect`-ing an SC output port to an SC input port.)

But an Event-style object prototype can stand in for a MIDIOut, and measure SystemClock.seconds directly.

So I know *how* to write the unit test, just would rather not dot all the i's until we know that it's needed.

telephon (commented):
I wonder whether it might be good to use the `schedBundleArrayOnClock` like in the `note` event type. This would make it easier to compare, perhaps?

As a separate issue, perhaps it would be better to implement the midi event completely as a variant of the note event, which, instead of sending to the server, sends to the midi device.

This would make it more consistent and more maintainable when there are future changes in the note event.

PR #6105, opened 2023-10-03
making the post window outputs of some subclasses of Collection follow the recommended code style

prko:
## Purpose and Motivation

The post window output of the collection's subclasses does not follow the recommended code style.
This PR makes these post-window outputs follow the recommended code style, so that users can use the post-window output in new code without removing whitespace.

Please see the following scsynth forum thread:

https://scsynth.org/t/array-recommended-style-of-code-and-how-to-return-to-the-post-window/8310

## Types of changes

- minor but practical change
 
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation <- not required
- [x] This PR is ready for review

PR #6106, opened 2023-10-04
Removed white space in brackets printed on the post window.

prko:
This PR is a correct PR (based on my branch) of the following closed wrong PR (based on the develop branch):

https://github.com/supercollider/supercollider/pull/6105

## Purpose and Motivation

The post window output of the collection's subclasses does not follow the recommended code style.
This PR makes these post-window outputs follow the recommended code style, so that users can use the post-window output in new code without removing whitespace.

Please see the following scsynth forum thread:

https://scsynth.org/t/array-recommended-style-of-code-and-how-to-return-to-the-post-window/8310

## Types of changes

- minor but practical change
 
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation <- not required

### DIFF
--- a/SCClassLibrary/Common/Collections/Array.sc
+++ b/SCClassLibrary/Common/Collections/Array.sc
@@ -290,15 +290,15 @@ Array[slot] : ArrayedCollection {
 
 	printOn { arg stream;
 		if (stream.atLimit, { ^this });
-		stream << "[ " ;
+		stream << "[" ;
 		this.printItemsOn(stream);
-		stream << " ]" ;
+		stream << "]" ;
 	}
 	storeOn { arg stream;
 		if (stream.atLimit, { ^this });
-		stream << "[ " ;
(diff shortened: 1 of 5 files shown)


prko:
@dyfer @cdbzb I think I did what you pointed out correctly this time, but I would ask you to check if I really did. I apologise for any inconvenience.

dyfer:
Thanks @prko! The PR (branches etc) looks fine now.

At the dev meeting on 2023-10-08 we discussed that this PR is 1) technically a breaking change, since the generated code with `.storeOn` would change, but 2) since this is a whitespace change, it should not actually break any reasonably written code. 

I'll pull this in now.

However, we should review other uses of storeOn/printOn to potentially provide a similar fix, for consistency. I'm planning to do that.

dyfer (approved):
Thanks!

Issue #2322, opened 2016-09-06
Inconsistent initial state of Delay2


Sciss:
Compare:
- `Delay1:` first two samples output are identical, i.e. initial state
      is same as first input sample
- `Delay2:` first sample is zero. second sample is same as first input
      sample.
- `DelayN(x, SampleDur.ir, SampleDur.ir)`: first sample is zero
- `DelayN(x, 2 * SampleDur.ir, 2 * SampleDur.ir)`: first two samples are zero

While there are good arguments to let `Delay1` repeat the first input sample, clearly `Delay2` does not do anything coherent by outputting zero followed by a duplicate of the first input sample. It should either output the first input sample three times, or it should output two frames of zero, followed by the first input sample.

snappizz:
Ack. Yet another case of sloppy ugen writing. Fixing it is yet another API change.

Zeros makes the most sense to me. A fairly standard definition in DSP theory is that an audio signal `x[t]` is defined for `t >= 0`, and for all negative times we set `x[t] = 0`. A delayed version of the signal `x'` would be defined as `x'[t] = x[t - delay]`, which would put zeros at the beginning of the delayed signal. I would propose making both `Delay1` and `Delay2` work this way.

snappizz:
Actually -- using zeros for Delay1 and Delay2 may be the wrong idea, because one of the most common use cases for these ugens is to recreate low-order FIR filters like LPZ/HPZ. dang.

snappizz:
My recent tribulations with Ctor functions revealed something important that is relevant to this bug -- Delay1 is not repeating the first sample.

A single-channel audio signal can be characterized as a discrete function `x[t]` of time `t >= 0`. But SuperCollider signals inside SynthDefs have an extra sample that gets propagated through the Ctor functions, which you can consider to be `x[-1]`. Most ugens will have `x[-1] = x[0]`, which is how it should be. Some do not.

danstowell:
You asked for "mathematical" comment. Mathematically behaviour at t<0 is undefined so we have no guidance, and I don't think we have any original spec to guide us. I'd suggest that the most sensible consistent interpretation for these delays is to assume zeros in the past (since under the other alternative the initial sample is putting an unbounded amount of energy into the system). As you say, there are consequences to making this choice. I don't know.

mtmccrea:
> since under the other alternative the initial sample is putting an unbounded amount of energy into the system

[...]

Issue #1313, opened 2015-02-14
Delay1's first output sample is its input, rather than 0


jamshark70:
Are we absolutely sure this is correct, in Delay1?

```
void Delay1_Ctor(Delay1* unit)
{
    //printf("Delay1_Reset\n");
    SETCALC(Delay1_next);
    unit->m_x1 = ZIN0(0);
    Delay1_next(unit, 1);
}
```

Specifically `unit->m_x1 = ZIN0(0);`

By definition, a delay line should be silent (i.e., 0) until the delay time has passed. The delay time here is 1 sample (ar) or 1 control block (kr). But instead of silence, this code replicates the first input value.

One user impact is that you can't use Delay1 on a trigger that occurs in the first sample or control block:

```
a = {
    var trig = Impulse.kr(0);
    Poll.kr(Delay1.kr(trig), trig);
    FreeSelf.kr(trig <= 0);
    Silent.ar(1)
}.play;
```

... prints nothing, where I'd have expected Poll's trigger input to be 0 in the first kr block, then 1 in the second. As a result, to delay an initial trigger (assuming kr) by 1 block, you have to use the clumsier construction `DelayN.kr(trig, 0.01, ControlDur.ir)`.

telephon:
The ````Delay1_Ctor```` method calls ````Delay1_next(unit, 1);````, which should *normally* then do the right thing. But: 

````
float x1 = unit->m_x1;
<...>
LOOP(inNumSamples & 3,
		x0 = ZXP(in);
		ZXP(out) = x1;
		x1 = x0;
	);

````
This will output ````unit->m_x1````, which has been set by  ````Delay1_Ctor```` to the input. So yes, this is wrong.

telephon:
Trying your example (and a variant) I do get the expected result.

````
(
a = {
    var trig = Impulse.ar(0);
	Poll.ar(Delay1.ar(trig), trig, "hello");
	FreeSelf.kr(T2K.kr(trig) <= 0);
    Silent.ar(1)
}.play;
)

(
a = {
    var trig = Impulse.kr(0);
	Poll.kr(Delay1.kr(trig), trig, "hello");
	FreeSelf.kr(trig <= 0);
    Silent.ar(1)
}.play;
)
`````
Returns: ````hello: 0````

jamshark70:
Probably close this one. I must have been seeing this only in my faulty (now-retired) local branch. Using a new local branch, forked off from very recent master, I don't see the problem anymore.

[...]

PR #6110, opened 2023-10-13
Fix UGen initialization: Delay1, Delay2

### LINKED ISSUE #4229: Delay2: first output sample is garbage
Environment
-----------

* Your SuperCollider version: 3.10.0
* Your operating system and version: macOS 10.12.6

Related to #2322, and could be fixed with it.

`Delay2_Ctor` doesn't initialize `m_x2`, which is x(-2):
https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/FilterUGens.cpp#L2032-L2037

So it outputs garbage for it's first sample 😕

Steps to reproduce (for bugs)
-----------------------------
```supercollider
// run each of these in order, observe the output
{LFSaw.ar(500, 0.5, 0, mul: 0.8)}.loadToFloatArray((10 / s.sampleRate), action: {|arr| arr.postln});
{Delay1.ar(LFSaw.ar(500, 0.5, mul: 0.8))}.loadToFloatArray((10 / s.sampleRate), action: {|arr| arr.postln});
{Delay2.ar(LFSaw.ar(500, 0.5, mul: 0.8))}.loadToFloatArray((10 / s.sampleRate), action: {|arr| arr.postln});
// garbage (random?) y(0)!

// the following line is identical to the previous
{Delay2.ar(LFSaw.ar(500, 0.5, mul: 0.8))}.loadToFloatArray((10 / s.sampleRate), action: {|arr| arr.postln});
// DIFFERENT (random?) y(0)!

// though that value is consistent after this point (unless you start from the top)
{Delay2.ar(LFSaw.ar(500, 0.5, mul: 0.8))}.loadToFloatArray((10 / s.sampleRate), action: {|arr| arr.postln});
```

Expected Behavior
-----------------
I personally expect 2 zeros, but that's another discussion (#2322, #1313)

Current Behavior
----------------
As shown in the above example, the first output is a seemingly garbage value, and changes on the second execution of the same line, then remains consistent.

I'd submit a fix for this by itself, but I'm undertaking a larger overhaul of UGen initialization fixes, so this is in line for that. And that process is pending community discussion and ... a lot of work.

mtmccrea:
## Purpose and Motivation

A majority of UGens do not initialize with a correct initialization sample, and subsequently output an incorrect first sample. This PR is part of an ongoing effort to https://github.com/supercollider/rfcs/pull/19, the progress of which is [tracked here](https://github.com/mtmccrea/supercollider/wiki/UGen-Fix-List).

This fixes the initialization of `Delay1` and `Delay2`.

Fixes #4229, #1313, #2322

## Types of changes

- Bug fix
- "Breaking" change, in that the first output sample is changed.

## To-do list

- [x] Code is tested
- [x] All tests are passing
  - added unit tests for `Delay1` and `Delay2` at `ar` and `kr` rates in: `TestCoreUGens:test_ugen_generator_equivalences`
- [ ] Updated documentation


mossheim:
changes lgtm, thanks :) will approve once question on first PR is answered

timblechmann:
there's a good reason to initialize delay ugens to the input sample rather than initializing it to zero: zero makes sense for "audio" signal, but not necessarily for "control" signals e.g. frequency/phase input for oscillators.

mtmccrea:
Good point, @timblechmann. I think the audio-rate case is more common and has lead to more bugs/unexpected behavior (I've listed a few bugs this fixes in the description), but the control signal case is valid and should be addressed.

I propose we add an arg to the tail, e.g. `initVal`.
Yes, adding an arg is awkward, esp. after mul/add. It's a precedent that may need to be set, because there are other UGens which would benefit from an `initVal` arg (esp. filters) and should be added in the future—it's "standard" DSP construction and it's unfortunate this wasn't built in.

I propose a non-conventional default: `initVal: 0.0` when the UGen runs at audio rate, `initVal: (in)` when run at control rate. I think it actually fits "common" use cases and minimizes breaking code. This would be clearly documented in the Help file.

Alternatives?

mossheim:
Yeah that is a good point. I have no comment on this, whatever people decide I'll review.

timblechmann:
> I propose a non-conventional default: initVal: 0.0 when the UGen runs at audio rate, initVal: (in) when run at control rate.

it's not necessarily about "audio rate" vs "control rate", though, but it's more the semantics of a signal: e.g. an audio-rate signal could very well be a frequency for fm/pm or a vcf. having some policy argument to "feed zero" or "feed initial value" could make sense, though

mtmccrea:
>  but it's more the semantics of a signal: e.g. an audio-rate signal could very well be a frequency for fm/pm or a vcf.

I agree. I'm hoping the defaults I'm proposing just minimize backwards breaking. I'll update this PR with those changes so the details can be reviewed.

mtmccrea:
To be more clear on my thinking for the default values, just based on my experience/assumptions:

1. When running at audio rate, we can more safely assume "audio" signals, so we base our default on the "analog" model of delay, i.e. preceded by silence, `0.0`.
2. When running at control rate, presumably a less common case, we can at least assume the signal is in the category of "control", so we are less confident about what the signal is. So default to the input.

Open to other ideas though!

mtmccrea:
I've updated the implementation so that predelay samples are now arguments. A bonus is that this will make these UGens much more useful now for filter design.

As described above, default predelay samples are `0.0` for `ar`, and equal to the first input sample at `kr`.

`nil` is handled as indicating the presample should be equal to the input. That makes the code concise, but I'm open to suggestions on alternatives, such as using a more descriptive Symbol.

Still TODO:
- update docs
- unit tests

mtmccrea:
OK, now unit tests have been added, docs updated, and the description above has been updated to include the implementation notes.

SCClassLibrary/Common/Audio/Delays.sc
```diff
@@ -1,14 +1,33 @@
...
-		^this.multiNew('audio', in).madd(mul, add)
-	}
-	*kr { arg in = 0.0, mul = 1.0, add = 0.0;
-		^this.multiNew('control', in).madd(mul, add)
-	}
+Delay1 : Filter {
+
+    *ar { arg in = 0.0, mul = 1.0, add = 0.0, x1 = 0.0;
+        // nil initializes the delay buffer with the input value
+        x1 = x1 ?? in;
```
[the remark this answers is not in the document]
mtmccrea: > should either be x1 = x1 ? in; or x1 = x1 ?? { in };

thanks, updated!

> Well, if you wanted it to match in, you'd have to write nil explicitly. This is quite implicit, I'd actually leave this line 5 out.

I'm not quite sure what you mean. Because the default is `0.0`, `nil` would need to be set explicitly to make `x1` the same as `in` by design—in which case line 5 would be necessary.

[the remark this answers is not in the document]
telephon: sorry for my late reply, I had not seen a notification of the update you made.
```supercollider
// so for initialisation with the input
// the intention is that you write:
DelayL.ar(in, 0.5, 0.5, nil)  
// why not write:
DelayL.ar(in, 0.5, 0.5, in)  
```

that would make it more explicit and you don't need the hidden logic in the ugen interface.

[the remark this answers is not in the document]
mtmccrea: I see what you mean that intention can be inferred from the default values being explicitly made `nil` in the class method definition. But in fact both are valid as is:
```supercollider
Delay1.ar(in, x1: in) 
Delay1.ar(in, x1: nil) // equivalent but obscure 
```
and I agree your suggestion is preferred for clarity.
The default behavior is more obvious at control rate:
```supercollider
Delay1.kr(in) // default behavior: predelay == in
Delay1.kr(in, x1: 0.0) // explicit predelay value
```
I've removed the explicit `nil` argument values from the method definitions.

I also revisited the help files with fresh eyes, and made changes that hopefully also make the intention more clear.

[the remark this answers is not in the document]
telephon: Thank you, I understand better now. If I understand this correctly, unlike at control rate, using `in` as `x1` in an audio rate UGen is not more standard than anything else, right? I read that from your reconstruction example.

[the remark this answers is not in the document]
mtmccrea: IIU your question correctly, yes, based on the prior conversation we arrived at the "standard" (default) behavior of **x1** being `0.0`, so specifying `x1: in` would be more exceptional, explicit behavior. I.e. the default follows the "analog" model of delay where silence precedes the delayed signal.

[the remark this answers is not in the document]
telephon: OK, good, thank you for the explantion. I would then suggest to remove the `?? { in }` phrases in the audio rate version, because it is unnecessarily implicit to define that via `nil`. In control rate, it is good as it is!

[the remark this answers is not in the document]
mtmccrea: > I would then suggest to remove the `?? { in }` phrases in the audio rate version, because it is unnecessarily implicit to define that via `nil`.

In that case, `nil` would no longer be valid for `x1` at audio rate—I can see your rationale for that. I should then update the description in the help docs and perhaps add some clarifying language about this additional difference between ar and kr.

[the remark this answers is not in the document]
telephon: yes, that would be good. I think it will be much better …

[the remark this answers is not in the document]
mtmccrea: After making the change, I tried to update the description of the behavior in the help files, and it actually came across as more convoluted.

> it is unnecessarily implicit to define that via `nil`
I do see your point with this. However, that logic not visible to the user unless they're inspecting the class file, and in effect supports more input types rather than less.

To try to illustrate this, I updated the help description and added clarifying comments in the class file. Have a look at the latests two commits to see those changes, and let me know if you still think it would be better to remove support for `nil` at audio rate.

HelpSource/Classes/Delay1.schelp
```diff
@@ -24,17 +24,27 @@ This value will be added to the output.
 
 argument::x1
 The sample value preceding the delayed input, which is also the value of the first output sample.
-This is a numeric value, a code::UGen::, or code::nil::.
-A value of code::nil:: is equivalent to setting strong::x1:: to strong::in::, effectively repeating the first input sample once.
+
+This can be a numeric value or a code::UGen::.
+code::nil:: is also supported, which is equivalent to setting code::x1: in::, effectively repeating the first sample of the input signal.
```
telephon, line 29: If you like, you could write more simply:

If code::x1:: is unspecified (code::nil::), it is internally set to code::x1 = in::, effectively repeating the first sample of the input signal.

HelpSource/Classes/Delay2.schelp
```diff
@@ -24,23 +24,31 @@ This value will be added to the output.
...
-strong::x1:: can be a numeric value, a code::UGen::, or code::nil::.
-A value of code::nil:: is equivalent to setting strong::x1:: to strong::in::, effectively repeating the first input sample.
 
 argument::x2
 The sample preceding the delayed input signal by two samples, which is also the first output sample.
-strong::x2:: can be a numeric value, a code::UGen::, or code::nil::.
-A value of code::nil:: is equivalent to setting strong::x2:: to strong::in::, effectively repeating the first input sample.
+
+strong::x1:: and strong::x2:: can be a numeric value or a code::UGen::.
+code::nil:: is also supported, which is equivalent to setting them to strong::in::, effectively repeating the first sample of the input signal.
```
telephon, line 32: … analogously here.

SCClassLibrary/Common/Audio/Delays.sc
```diff
@@ -1,14 +1,34 @@
...
-		^this.multiNew('audio', in).madd(mul, add)
-	}
-	*kr { arg in = 0.0, mul = 1.0, add = 0.0;
-		^this.multiNew('control', in).madd(mul, add)
-	}
+Delay1 : Filter {
+
+    *ar { arg in = 0.0, mul = 1.0, add = 0.0, x1 = 0.0;
+        // nil is supported for consistency with *kr, indicating x1 is set to in[0]
+        x1 = x1 ?? { in };
```
telephon, line 5: so this line (5) can be left out, this is better.

SCClassLibrary/Common/Audio/Delays.sc
```diff
@@ -1,14 +1,34 @@
...
+        ^this.multiNew('control', in, x1).madd(mul, add)
+    }
 }
 
-Delay2 : Delay1 { }
+Delay2 : Filter {
+
+    *ar { arg in = 0.0, mul = 1.0, add = 0.0, x1 = 0.0, x2 = 0.0;
+        // nil is supported for consistency with *kr, indicating x1 and x2 are set in[0]
+        x1 = x1 ?? { in };
```
telephon, line 20: same here, better leave out these two lines.

HelpSource/Classes/Delay1.schelp
mtmccrea: Yes, this is better wording.

However, with your suggested change to remove `?? { in }`, then `nil` would no longer be supported at audio rate (the UGen would fail to compile if `nil` were passed in), so this description wouldn't be accurate. 

So there would have to be two separate descriptions, one for audio rate and one for control rate. This was the reason for suggesting that `?? { in }` remain in, since it has no practical impact users (it actually gives them more options, even if it's seldom used), and, most importantly IMO, simplifies communication about the behavior.

I don't outright object to removing `?? { in }`, I just want to make sure my intention was clear before trying to rework the docs again.

telephon: I see your point, it is more uniform from the perspective of argument values. I would look at it from the perspective of default values. Currently, in the kr case, the default is `in`, and in the ar case, the default is `0`, _unless_ you actively replace it with `nil`. I think this makes things hard to understand. It would be clearer if this "unless" could be dropped.

Btw it could be even made clearer if you'd write
```
*kr { arg in = 0.0, mul = 1.0, add = 0.0, x1 = (in);
```

mtmccrea: > Btw it could be even made clearer if you'd write
*kr { arg in = 0.0, mul = 1.0, add = 0.0, x1 = (in);

Ah, this is great! I didn't know it was possible to use argument names when specifying programatic defaults.

This actually solves the issue completely, as `nil` was only being used to handle this default case. So now we can just remove support for `nil` as an argument (unless there's some other reason to keep it?). I'll update the class, docs, and tests in the next day or two.

telephon:
yes, so it would read clearly:
```
*ar { arg in = 0.0, mul = 1.0, add = 0.0, x1 = 0.0, x2 = 0.0;
*kr { arg in = 0.0, mul = 1.0, add = 0.0, x1 = (in), x2 = (in);
```

and all the `x1 = x1 ?? { in }` can go away.

HelpSource/Classes/Delay1.schelp
telephon: yes, so it would read clearly:
```
*ar { arg in = 0.0, mul = 1.0, add = 0.0, x1 = 0.0, x2 = 0.0;
*kr { arg in = 0.0, mul = 1.0, add = 0.0, x1 = (in), x2 = (in);
```

and all the `x1 = x1 ?? { in }` can go away.

mtmccrea: I've now removed all the logic related to supporting `nil`, updated the docs, and the unit tests.

SCClassLibrary/Common/Audio/Delays.sc
```diff
@@ -1,14 +1,27 @@
...
-	*kr { arg in = 0.0, mul = 1.0, add = 0.0;
-		^this.multiNew('control', in).madd(mul, add)
-	}
+    *ar { arg in = 0.0, mul = 1.0, add = 0.0, x1 = 0.0;
+        ^this.multiNew('audio', in, x1).madd(mul, add)
+    }
+
+    // Unlike *ar, x1 defaults to the first sample of the input value
+    *kr { arg in = 0.0, mul = 1.0, add = 0.0, x1 = (in);
+        x1 = x1 ?? { in };
```
telephon, line 9: line 9 can now be left out

telephon (changes requested):
just a typo, I think.

mtmccrea:
> just a typo, I think.

Ah, thanks for catching that, sorry to have missed it! Updated...

telephon (approved):

mtmccrea:
Would anyone like to make the merge on this? There have been a number of eyes on it and I think it's been thoroughly fleshed out at this point.

The 3 commits are grouped by `comp`, they could either be squashed or left as-is.

mtmccrea:
Thanks, all!

PR #6111, opened 2023-10-15
Fixes for windows dev

jrsurge:
## Purpose and Motivation

Visual Studio has supported CMake projects for a while now, without the need to generate a VS project.

We fail to compile (at least in this mode) with VS2022 (possibly others, didn't check) due to the way we explicitly invoke the resource compiler (for setting things like the app icon etc) - which was now being called with invalid/deprecated args.

Adding the resource files (.rc) directly to the targets in question (sclang, scide), CMake will detect and invoke the resource compiler automatically for us, so we can simplify this quite a bit.

## Implementation

- Add extra ignores for Visual Studio when running in cmake mode
- Fix resource compilation by adding directly to target (CMake sorts it out for us, no need to explicitly call it)

## Types of changes
- Bug fix
## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,14 @@ platform/mac/build
 CMakeLists.txt.user
 external_libraries/portaudio/asio*
 
+# Visual Studio
+.vs/**
+out/**
+CMakeSettings.json
+
+# CMake User Presets
+CMakeUserPresets.json
+
 # working files made by packager script
 common/Packager/LinuxExclusions.txt
 common/Packager/SuperCollider*.tar.gz
--- a/editors/sc-ide/CMakeLists.txt
+++ b/editors/sc-ide/CMakeLists.txt
@@ -277,16 +277,6 @@ if(APPLE)
     endforeach()
 endif()
 
-# This sets up the exe icon for windows.
-if(WIN32)
- set(RES_FILES ${CMAKE_SOURCE_DIR}/platform/windows/Resources/scide.rc)
- set(CMAKE_RC_COMPILER_INIT windres)
- ENABLE_LANGUAGE(RC)
- SET(CMAKE_RC_COMPILE_OBJECT
- "<CMAKE_RC_COMPILER> -O coff <DEFINES> -i <SOURCE> -o <OBJECT>")
-endif(WIN32)
-
-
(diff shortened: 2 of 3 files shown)


mossheim (dismissed):
Thanks!

mossheim:
The build failed with:

> C:\mingw64\bin\windres.exe: invalid option -f

I wonder if CMake is treating stuff like `add_definitions( -fno-math-errno)` as applying to .rc files as well as .cpp with MINGW. Could run the build with verbose output turned on to see what it's actually passing to windres.exe to get a better idea.

sadguitarius:
Just realizing this has been around for a while. I put the below in my CMakeLists.txt and it does fix the build on Win32. I assume it also should be fine for MSYS but I don't have that set up to try at the moment.

```
if(WIN32)
	set(RES_FILES ${CMAKE_SOURCE_DIR}/platform/windows/Resources/scide.rc)
	ENABLE_LANGUAGE(RC)
	if(MSYS)
		set(CMAKE_RC_COMPILER_INIT windres)
		SET(CMAKE_RC_COMPILE_OBJECT
		"<CMAKE_RC_COMPILER> -O coff <DEFINES> -i <SOURCE> -o <OBJECT>")
	endif(MSYS)
endif(WIN32)
```

dyfer:
@jrsurge could we merge https://github.com/supercollider/supercollider/pull/6653 instead, as it preserves code for mingw? 

Also, should the change in this PR, [moving the rc file from `add_executable` to `targer_sources`](https://github.com/supercollider/supercollider/pull/6111/files#diff-3b5ebf986988ae0060dfa14c9cfd28f31a34e26c31cfc9261f761bd1a03cf575L318-R315) still be applied?

jrsurge:
I'm not quite sure - it might need a mixture of the two, would need to check. Feel free to merge the other one and I can rebase this one. I'll either get it working again or close it if it's all ok

dyfer:
@jrsurge the other PR is merged, please check if we still need the changes here. Thanks!

dyfer:
I believe a similar fix for resource file has already been merged. 
Closing due to inactivity - feel free to reopen if desired!

PR #6112, opened 2023-10-15
GHA: checkout older readline from vcpkg

dyfer:
## Purpose and Motivation

It seems that readline shared library got broken in vcpkg - https://github.com/microsoft/vcpkg/issues/34501

I propose to roll back the readline port version until this is fixed upstream and makes its way to GHA images.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -523,6 +523,9 @@ jobs:
         if: matrix.vcpkg-triplet
         shell: bash
         run: |
+          cd $VCPKG_INSTALLATION_ROOT
+          git fetch origin 80f8da23a88a982cf2d81d1a01623c55bb58df15
+          git checkout 80f8da23a88a982cf2d81d1a01623c55bb58df15 -- ports/readline*
           vcpkg install readline --triplet="${{ matrix.vcpkg-triplet }}" --overlay-triplets="$GITHUB_WORKSPACE/vcpkg/triplets"
       - name: configure
         shell: bash


.github/workflows/actions.yml
```diff
@@ -523,6 +523,9 @@ jobs:
         if: matrix.vcpkg-triplet
         shell: bash
         run: |
+          cd $VCPKG_INSTALLATION_ROOT
+          git fetch origin 80f8da23a88a982cf2d81d1a01623c55bb58df15
+          git checkout 80f8da23a88a982cf2d81d1a01623c55bb58df15 -- ports/readline*
```
jrsurge, line 528: Just to check, where did this commit come from? Should we be reverting to a release tag instead?

dyfer: This commit is from the previous update to the readline-win32 port: https://github.com/microsoft/vcpkg/commits/master/ports/readline-win32/portfile.cmake (PR https://github.com/microsoft/vcpkg/pull/33247)
I guess we could revert to a previous release if you'd like, though vcpkg is updated in gha runners more frequently than releases. Either way for me.
One thing to consider: the [fix for the readline issue](https://github.com/microsoft/vcpkg/pull/34502) has been _just_ pulled in upstream. So maybe we just wait until it makes its way to gha? (a week or two?)

dyfer:
The updated vcpkg made has just made its way into the gha runner image. I think these date take 3-4 days to deploy. I think we can just wait for this instead of applying (and later reverting) this fix.

Issue #6104, opened 2023-10-03
macOS build fails due to libsndfile linking


dyfer:
## Environment

* SuperCollider version: develop
* Operating system: macOS 12.6
* Other details (Qt version, audio driver, etc.): libsndfile 1.2.2 installed from homebrew; architecture arm64 (Apple Silicon)

```brew info libsndfile```:
<details>

```
==> libsndfile: stable 1.2.2 (bottled)
C library for files containing sampled sound
https://libsndfile.github.io/libsndfile/
/opt/homebrew/Cellar/libsndfile/1.2.2 (53 files, 1MB) *
  Poured from bottle using the formulae.brew.sh API on 2023-09-11 at 14:22:06
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/lib/libsndfile.rb
License: LGPL-2.1-or-later
==> Dependencies
Build: cmake ✘
Required: flac ✔, lame ✔, libogg ✔, libvorbis ✔, mpg123 ✔, opus ✔
==> Analytics
install: 34,099 (30 days), 92,184 (90 days), 201,941 (365 days)
install-on-request: 1,590 (30 days), 4,377 (90 days), 8,859 (365 days)
build-error: 11 (30 days)
```
</details>

## Steps to reproduce

```sh
cmake -G Xcode -D CMAKE_PREFIX_PATH=`brew --prefix qt5` -D SC_VERIFY_APP=ON ..
cmake --build . --target install --config RelWithDebInfo
```

## Expected vs. actual behavior

**It seems that there's no way currently to create a working build of SC on macOS** with the current version of libsndfile.

Actual behavior: the verification phase indicates the issue:
<details>

```
-- Verifying app
-- ===========================================================================
-- Analyzing app='/Volumes/data/Dokumenty/src/supercollider/build/Install/SuperCollider/SuperCollider.app'
-- bundle='/Volumes/data/Dokumenty/src/supercollider/build/Install/SuperCollider/SuperCollider.app'
-- executable='/Volumes/data/Dokumenty/src/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/SuperCollider'
-- valid='1'
-- executable file 1: /Volumes/data/Dokumenty/src/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/SuperCollider
[...]

PR #6115, opened 2023-10-17
GHA: update the brew-install-universal script

dyfer:
## Purpose and Motivation

We use my [script](https://gist.github.com/dyfer/6c83905d4593750105897e51e87ec345#file-brew-install-universal-sh) to create "universal" (x86_64 + arm64) homebrew packages to facilitate macOS universal build.

A recent change in homebrew caused an issue with deploying some libraries with `macdeployqt` (see https://github.com/supercollider/supercollider/issues/6104). While we wait (I think?) for the updated libsndfile bottle in homebrew, I realized that I can "manually" change the non-fully qualified paths to full paths in the manually prepared bottles, since I am manipulating the paths already: https://gist.github.com/dyfer/6c83905d4593750105897e51e87ec345#file-brew-install-universal-sh-L226-L251

In this update, the script converts the non-fully qualified paths (@loader_path/.. etc) to fully qualified paths, to avoid problems with macdeployqt.

I also added the `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK`, since some "universal" packages were being overwritten by the automatic update (i.e. the single architecture package would overwrite the one installed with a script).

With this update, we are getting back functional macOS builds in the CI.

## Types of changes

- Bug fix (kind of)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -247,6 +247,7 @@ jobs:
       HOMEBREW_NO_ANALYTICS: 1
       HOMEBREW_NO_AUTO_UPDATE: 1
       HOMEBREW_NO_INSTALL_CLEANUP: 1
+      HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
       USE_SYSLIBS: ${{ matrix.use-syslibs }}
       SHARED_LIBSCSYNTH: ${{ matrix.shared-libscsynth }}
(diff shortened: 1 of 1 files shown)


dyfer:
Libsndfile bottle was update upstream. In that case we don't need this update currently.

dyfer:
Actually, let's pull this in, I think. The updated script will cover a case if the loader_paths are used again.
However, more importantly, the addition of `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK` will fix #6117

joshpar (approved):

PR #6116, opened 2023-10-18
CMakeLists: only enable SSE on x86 variants

Xeonacid:
Make it build on RISC-V, MIPS etc.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,10 +188,10 @@ option(FINAL_BUILD "Build as single source file." OFF)
 
 option(FFT_GREEN "Use internal 'Green' FFT lib rather than FFTW. (Not recommended.)" OFF)
 
-if(NOT ${CMAKE_SYSTEM_PROCESSOR} MATCHES "^arm|aarch64")
+if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^AMD64|EM64T|x86_64|X86|i386|i686")
     option(SSE "Compile with support for SSE instructions." ON)
     option(SSE2 "Compile with support for SSE2 instructions." ON)
-else() # ARM platforms do not have SSE
+else() # non-x86 platforms do not have SSE
     set(SSE OFF)
     set(SSE2 OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -188,10 +188,10 @@ option(FINAL_BUILD "Build as single source file." OFF)
 
 option(FFT_GREEN "Use internal 'Green' FFT lib rather than FFTW. (Not recommended.)" OFF)
 
-if(NOT ${CMAKE_SYSTEM_PROCESSOR} MATCHES "^arm|aarch64")
+if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^AMD64|EM64T|x86_64|X86|i386|i686")
```
dyfer, line 191: Thanks for the PR!
Are you sure i386 should be a part of this? I'm not sure if SC would build on such an old architecture, but SSE was not available until Pentium II/III I think (which is i686 IIUC).

Xeonacid: > At the same time, I'm not sure if i386 wouldn't show up as an architecture on some systems even for current CPUs... do you know if this information can be checked somewhere?

I take this from https://stackoverflow.com/questions/70475665/what-are-the-possible-values-of-cmake-system-processor and https://stackoverflow.com/questions/45125516/possible-values-for-uname-m .
> Are you sure i386 should be a part of this? I'm not sure if SC would build on such an old architecture, but SSE was not available until Pentium II/III I think (which is i686 IIUC).

This value can be taken from `uname -m` (from the answer above). I think it does not mean SSE is unsupported, instead just a kernel baseline. There should not be a real i386 CPU running these days. Of course it depends on your opinion. :P

dyfer: :)
Got it, thanks for the reference. Let's keep it as is.

dyfer (approved):
Thanks!

PR #6118, opened 2023-10-20
GHA macOS legacy: build fftw

dyfer:
## Purpose and Motivation

While I acknowledge that it's unsure how much longer we'll be supporting the macOS legacy build, I'd like to make its impact on our GHA build times minimal.
Due to homebrew dropping support for macOS 11, some dependencies of fftw started to [build from source](https://github.com/supercollider/supercollider/actions/runs/6579631472/job/17881931695#step:9:115), increasing build times significantly. I'd like to switch fftw to build from vcpkg, which is faster, and is also cached.

EDIT: I decided to also change the runner image to macOS 12, since there seem to be other issues with using macOS 11.

## Types of changes

- Bug fix (kind of)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -219,6 +219,7 @@ jobs:
             shared-libscsynth: false
             build-libsndfile: true
             system-portaudio: false
+            build-fftw: true
             vcpkg-triplet: x64-osx-release-supercollider # required for build-libsndfile
             artifact-suffix: 'macOS-x64-legacy' # set if needed - will trigger artifact upload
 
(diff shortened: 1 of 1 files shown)


dyfer:
As an alternative, we could move to macOS 12 for building the legacy build, though I'd need to test if it still works on the older platforms.

dyfer:
I suggest that we pull this in to get all the builds working again, which will allow running the testsuite.

joshpar (approved):

PR #6124, opened 2023-11-06
[Windows] switch system calls to UTF-16

dyfer:
## Purpose and Motivation

On Windows it seems that for the Qt6 build we need to be building with the `UNICODE` flag set, and in turn make system calls using UTF-16 (WideChar) functions and convert the strings to/from UTF-8 as needed. This PR includes only the UTF-16 changes so that it can be reviewed separately from the rest of the Qt6 update (which is still WIP). Another improvement is that it's now possible to use any allowed unicode characters for environment variables on Windows.

EDIT:
This is updated after env variable interface was redone with `std::string`: #6216.

This PR is now using the [codecvt](https://github.com/supercollider/supercollider/blob/develop/common/SC_Codecvt.hpp#L60-L69) functions that already existed in the codebase, instead of adding more helper functions.

I removed unused function `win32_GetKnownFolderPath` along the way.

Outdated comments hidden below:
<details>

Please review carefully, in particular in terms of memory management, as I don't have much experience in this area.

EDIT: 
After consulting with @jrsurge , I've set up wrapper functions to handle the conversion and used `unique_ptr` to help managing the memory.

Please note: I've snuck in one semi-related change into this PR - I increased the hardcoded maximum size for the value of an environment variable. The previous max was 1024 and on my Windows system `PATH` was already ~1500 characters long, so I couldn't fully set it from SC if I wanted to. I increased the max size to 32767, which seems to be the [maximum size on Windows](https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables?redirectedfrom=MSDN#:~:text=The%20maximum%20size%20of%20a%20user%2Ddefined%20environment%20variable%20is%2032%2C767%20characters.). On POSIX the maximum size appears to be [larger](https://man7.org/linux/man-pages/man2/execve.2.html#:~:text=Limits%20on%20size%20of%20arguments%20and%20environment), so I think the Windows's max is a reasonable choice.

</details>

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/common/SC_Win32Utils.cpp
+++ b/common/SC_Win32Utils.cpp
@@ -52,14 +52,6 @@ void win32_ExtractContainingFolder(char* folder, const char* pattern, int maxCha
         folder[0] = 0;
 }
 
-void win32_GetKnownFolderPath(int folderId, char* dest, int size) {
-    // Use a temporary buffer, as SHGetFolderLocation() requires it
-    // to be at least MAX_PATH size, but destination size may be less
-    char buf[MAX_PATH];
-    SHGetFolderPath(NULL, folderId, NULL, 0, buf);
-    strncpy(dest, buf, size);
-}
-
 char* win32_basename(char* path) {
     int pathLen = strlen(path);
     int lastPathSepFoundPos = -1;
--- a/lang/LangPrimSource/PyrStringPrim.cpp
+++ b/lang/LangPrimSource/PyrStringPrim.cpp
@@ -622,10 +622,13 @@ int prString_Getenv(struct VMGlobals* g, int /* numArgsPushed */) {
         return this_err;
 
 #ifdef _WIN32
-    const auto count = GetEnvironmentVariable(this_str.c_str(), nullptr, 0);
-    std::string buf(count, 0);
+    const auto this_str_w = SC_Codecvt::utf8_cstr_to_utf16_wstring(this_str.c_str());
+    const auto count = GetEnvironmentVariableW(this_str_w.c_str(), nullptr, 0);
+    std::string buf;
     if (count != 0) {
-        GetEnvironmentVariable(this_str.c_str(), buf.data(), buf.size());
+        std::wstring wbuf(count, 0);
+        GetEnvironmentVariableW(this_str_w.c_str(), wbuf.data(), count);
+        buf = SC_Codecvt::utf16_wcstr_to_utf8_string(wbuf.c_str());
     }
     char* value = count != 0 ? buf.data() : nullptr;
 #else
@@ -658,7 +661,7 @@ int prString_Setenv(struct VMGlobals* g, int numArgsPushed) {
 
     if (IsNil(slot_value)) {
 #ifdef _WIN32
-        SetEnvironmentVariable(this_name_str.c_str(), nullptr);
+        SetEnvironmentVariableW(SC_Codecvt::utf8_cstr_to_utf16_wstring(this_name_str.c_str()).c_str(), nullptr);
 #else
         unsetenv(this_name_str.c_str());
 #endif
@@ -667,7 +670,8 @@ int prString_Setenv(struct VMGlobals* g, int numArgsPushed) {
         if (value_err != errNone)
             return value_err;
 #ifdef _WIN32
-        SetEnvironmentVariable(this_name_str.c_str(), value_str.c_str());
+        SetEnvironmentVariableW(SC_Codecvt::utf8_cstr_to_utf16_wstring(this_name_str.c_str()).c_str(),
+                                SC_Codecvt::utf8_cstr_to_utf16_wstring(value_str.c_str()).c_str());
 #else
         setenv(this_name_str.c_str(), value_str.c_str(), 1);
 #endif
--- a/server/scsynth/SC_Lib_Cintf.cpp
+++ b/server/scsynth/SC_Lib_Cintf.cpp
@@ -38,6 +38,7 @@
 
 #ifdef _WIN32
 #    include "SC_Win32Utils.h"
+#    include "SC_Codecvt.hpp"
 #else
 #    include <dlfcn.h>
 #    include <libgen.h>
@@ -288,11 +289,12 @@ static bool PlugIn_Load(const bfs::path& filename) {
     // because the native encoding on Windows is utf-16.
     const std::string filename_utf8_str = SC_Codecvt::path_to_utf8_str(filename);
     if (!hinstance) {
-        char* s;
+        wchar_t* s;
         DWORD lastErr = GetLastError();
-        FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL,
-                      lastErr, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (char*)&s, 0, NULL);
-        scprintf("*** ERROR: LoadLibrary '%s' err '%s'\n", filename_utf8_str.c_str(), s);
+        FormatMessageW(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                       NULL, lastErr, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (wchar_t*)&s, 0, NULL);
+        scprintf("*** ERROR: LoadLibrary '%s' err '%s'\n", filename_utf8_str.c_str(),
+                 SC_Codecvt::utf16_wcstr_to_utf8_string(s).c_str());
         LocalFree(s);
         return false;
     }
@@ -311,10 +313,10 @@ static bool PlugIn_Load(const bfs::path& filename) {
 
     void* ptr = (void*)GetProcAddress(hinstance, "load");
     if (!ptr) {
-        char* s;
-        FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL,
-                      GetLastError(), MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (char*)&s, 0, NULL);
-        scprintf("*** ERROR: GetProcAddress err '%s'\n", s);
+        wchar_t* s;
+        FormatMessageW(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                       NULL, GetLastError(), MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (wchar_t*)&s, 0, NULL);
+        scprintf("*** ERROR: GetProcAddress err '%s'\n", SC_Codecvt::utf16_wcstr_to_utf8_string(s).c_str());
         LocalFree(s);
 
         FreeLibrary(hinstance);
--- a/common/SC_Win32Utils.h
+++ b/common/SC_Win32Utils.h
@@ -56,7 +56,6 @@ typedef int pid_t;
 void win32_ReplaceCharInString(char* string, int len, char src, char dst);
 // Finds the parent folder of a specified path pattern (including trailing slash)
 void win32_ExtractContainingFolder(char* folder, const char* pattern, int maxChars);
-void win32_GetKnownFolderPath(int folderId, char* dest, int size);
 void win32_synctimes();
 char* win32_basename(char* path);
 char* win32_dirname(char* path);
--- a/editors/sc-ide/CMakeLists.txt
+++ b/editors/sc-ide/CMakeLists.txt
@@ -348,6 +348,7 @@ if(APPLE)
     target_link_libraries( libscide PUBLIC "-framework Cocoa" )
 elseif(WIN32)
     target_link_libraries( libscide PUBLIC wsock32 )
+    target_compile_definitions( libscide PUBLIC UNICODE _UNICODE)
     # The following prevents a Windows console from showing up
     # when the executable is started:
     set_target_properties( SuperCollider PROPERTIES WIN32_EXECUTABLE TRUE )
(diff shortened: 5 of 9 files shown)


server/scsynth/SC_Lib_Cintf.cpp
```diff
@@ -283,17 +283,21 @@ bool checkServerVersion(void* f, const char* filename) {
 
 static bool PlugIn_Load(const bfs::path& filename) {
 #ifdef _WIN32
-    HINSTANCE hinstance = LoadLibraryW(filename.wstring().c_str());
+    HINSTANCE hinstance = LoadLibrary(filename.wstring().c_str());
```
timblechmann, line 286: btw, i'd recommend to always use the `A` or `W` suffixed functions explicitly instead of relying on the preprocessor defines. this also allows you to mix ansi and wide apis in the same codebase

dyfer: Thanks for the feedback!
In that case I'd change all the calls to the Wide versions for consistency, including e.g. [FormatMessage](https://github.com/supercollider/supercollider/pull/6124/files#diff-f36a627946e1f45158f677041908350129be5deb4f0b82537a03c8751b143f67L293). Should I do this?

timblechmann: not sure if it's worth the effort. when it's about obtaining error codes, the ansi variant of the functions is probably good enough. main rationale is not to rely on the `UNICODE` preprocessor define (e.g. never use `LoadLibrary`, but either `LoadLibraryW` or `LoadLibraryA`)

details: https://learn.microsoft.com/en-us/windows/win32/intl/conventions-for-function-prototypes

dyfer: The page you linked says:
> They [applications] should be written with generic functions, and should define UNICODE to compile the functions into Unicode functions.

So it seems that the MS's advice is to use generic functions and the preprocessor definition, right?

timblechmann: i'd not follow microsoft's recommendation here, but be explicit (in my experience relying on the preprocessor define calls for trouble the more complex the codebase becomes)

dyfer: Updated. Thoughts?

lang/LangPrimSource/PyrStringPrim.cpp
```diff
@@ -611,7 +624,13 @@ int prString_Setenv(struct VMGlobals* g, int /* numArgsPushed */) {
...
             return err;
 #ifdef _WIN32
-        SetEnvironmentVariable(key, value);
+        wchar_t key_w[256 * sizeof(wchar_t)];
+        wchar_t value_w[1024 * sizeof(wchar_t)];
+        auto size_k = MultiByteToWideChar(CP_UTF8, 0, key, -1, nullptr, 0);
+        auto size_v = MultiByteToWideChar(CP_UTF8, 0, value, -1, nullptr, 0);
+        if ((MultiByteToWideChar(CP_UTF8, 0, key, -1, key_w, size_k) != 0)
+            && (MultiByteToWideChar(CP_UTF8, 0, value, -1, value_w, size_v) != 0))
+            SetEnvironmentVariableW(key_w, value_w);
```
dyfer, line 633: Looking again at this... if the `MultiByteToWideChar` conversion fails, maybe we should call `SetEnvironmentVariableA(key, value)`?

lang/LangPrimSource/PyrStringPrim.cpp
```diff
@@ -568,12 +568,18 @@ int prString_Getenv(struct VMGlobals* g, int /* numArgsPushed */) {
...
-    if (size == 0 || size > 1024)
-        value = 0;
-    else
-        value = buf;
+    auto key_w = win32_CharToWchar(key);
+    DWORD size = GetEnvironmentVariableW(key_w.get(), nullptr, 0);
+    auto value_w = std::make_unique<wchar_t[]>(size);
+    size = GetEnvironmentVariableW(key_w.get(), value_w.get(), size);
+
+    std::unique_ptr<char[]> value_ptr;
```
dyfer, line 576: I've defined `value_ptr` here so that it stays in scope until `PyrString* pyrString = newPyrString(g->gc, value, 0, true);`. Without this, if I did `value = win32_WcharToChar(value_w.get()).get()` (line 580), I'd get garbage data in `value`....

lang/LangPrimSource/PyrStringPrim.cpp
dyfer: nvm, this has changed.

lang/LangPrimSource/PyrStringPrim.cpp
xunil-cloud: Hi, Since you commented on this section, I’m assuming you're not certain about this code?

common/SC_Win32Utils.cpp
```diff
@@ -52,12 +52,18 @@ void win32_ExtractContainingFolder(char* folder, const char* pattern, int maxCha
...
 }
 
-void win32_GetKnownFolderPath(int folderId, char* dest, int size) {
-    // Use a temporary buffer, as SHGetFolderLocation() requires it
-    // to be at least MAX_PATH size, but destination size may be less
-    char buf[MAX_PATH];
-    SHGetFolderPath(NULL, folderId, NULL, 0, buf);
-    strncpy(dest, buf, size);
+std::wstring win32_StringToWstring(const std::string& string) {
+    DWORD size = MultiByteToWideChar(CP_UTF8, 0, string.c_str(), -1, nullptr, 0);
```
JordanHendersonMusic, line 56: What happens if the string is empty? 

Cpp ref says this
> If empty() returns true, the pointer is a non-null pointer that should not be dereferenced.

dyfer: Arguably, both these functions are not doing any checking, neither whether the input string is valid, nor whether the conversion succeeded. 

Here, do you suggest I'd do something like 

```cpp
DWORD size = MultiByteToWideChar(CP_UTF8, 0, string.empty() ? nullptr : string.c_str(), -1, nullptr, 0);
```
?

common/SC_Win32Utils.cpp
```diff
@@ -52,12 +52,18 @@ void win32_ExtractContainingFolder(char* folder, const char* pattern, int maxCha
...
+    std::wstring outputString(size, 0);
```
dyfer, line 57: Similarly, I'm checking whether size is > 0. However, that is done elsewhere in the Env var code, typically.

Should I do the rest like
```cpp
if(size > 0) {
    std::wstring outputString(size, 0);
    MultiByteToWideChar(CP_UTF8, 0, string.c_str(), -1, outputString.data(), string.length());
    return outputString;
} else {
   return nullptr; // or an empty string?
}
```

dyfer:
@JordanHendersonMusic I might go another route with this, and use exisgint [codecvt](https://github.com/supercollider/supercollider/blob/develop/common/SC_Codecvt.hpp#L60-L69) functions for these conversions...

lang/LangPrimSource/PyrStringPrim.cpp
```diff
@@ -622,10 +622,13 @@ int prString_Getenv(struct VMGlobals* g, int /* numArgsPushed */) {
...
+    const auto this_str_w = SC_Codecvt::utf8_cstr_to_utf16_wstring(this_str.c_str());
+    const auto count = GetEnvironmentVariableW(this_str_w.c_str(), nullptr, 0);
+    std::string buf;
     if (count != 0) {
-        GetEnvironmentVariable(this_str.c_str(), buf.data(), buf.size());
+        std::wstring wbuf(count, 0);
+        GetEnvironmentVariableW(this_str_w.c_str(), &wbuf[0], count);
+        buf = SC_Codecvt::utf16_wcstr_to_utf8_string(wbuf.c_str());
     }
     char* value = count != 0 ? buf.data() : nullptr;
```
dyfer, line 633: @JordanHendersonMusic I was wondering, now that I'm getting `buf` string from the `utf16_wcstr_to_utf8_string` function, should I use `buf.c_str()` here instead of `.data()`?

dyfer: Upon further reflection, I think this should indeed be `buf.c_str()` so I'm changing this.

dyfer: Well... it seems that doesn't work, after all. I got 
`D:\a\supercollider\supercollider\lang\LangPrimSource\PyrStringPrim.cpp(633,53): error C2440: 'initializing': cannot convert from 'const _Elem *' to 'char *' `
when compiling.
Reverting back to `.data()`.

JordanHendersonMusic:
I didn't know that header existed (thanks!), sounds good.

Also, and I'm not too sure about this... How big is `wchar`? I don't actually know what a wide string is, variable sized types always confuse me. There is a dedicated `std::u16string` would that work?

dyfer:
> Also, and I'm not too sure about this... How big is `wchar`?

It seems to be [implementation dependent](https://learn.microsoft.com/en-us/cpp/standard-library/string-typedefs?view=msvc-170#wstring), so it could be 2 or 4 bytes. It's 2 bytes in MSVC. 

Generally I see examples of `SetEnvironmentVariableW` and other `...W` functions interfacing with `wchar`s, so I thought that would be fine...?

JordanHendersonMusic:
I deleted that question about `std::u16string` as I realised I was wrong (so ignore that!), in addition to the size, it seems it's backed by a signed `char16_t` whereas `wchar` is unsigned.

Text is annoyingly complex.

dyfer:
@JordanHendersonMusic do you have any more feedback on this PR? 
I tested this on a Windows machine and it seems to work fine - setting and getting env vars works.

lang/LangPrimSource/PyrStringPrim.cpp
```diff
@@ -622,10 +622,13 @@ int prString_Getenv(struct VMGlobals* g, int /* numArgsPushed */) {
...
 #ifdef _WIN32
-    const auto count = GetEnvironmentVariable(this_str.c_str(), nullptr, 0);
-    std::string buf(count, 0);
+    const auto this_str_w = SC_Codecvt::utf8_cstr_to_utf16_wstring(this_str.c_str());
+    const auto count = GetEnvironmentVariableW(this_str_w.c_str(), nullptr, 0);
+    std::string buf;
     if (count != 0) {
-        GetEnvironmentVariable(this_str.c_str(), buf.data(), buf.size());
+        std::wstring wbuf(count, 0);
+        GetEnvironmentVariableW(this_str_w.c_str(), &wbuf[0], count);
```
JordanHendersonMusic, line 630: `&wbuf[0]` could be `wbuf.data()`.

JordanHendersonMusic (approved):
I can't test as I don't have a windows machine, but I've read through this a few times and it all looks good. Using the features we already had was definitely the way to go 😀 

could probably use `.data` rather than `&b[0]`.

lang/LangPrimSource/PyrStringPrim.cpp
dyfer: I was a bit confused whether string.data() should be allowed to be manipulated/modified or not... I'm happy to change, just trying to understand the nuances.

JordanHendersonMusic: The non-const overload of `data` was added in c++17 (after strings became contiguous in 11).

Cpp ref mentions that it is undefined behaviour to change the char at `data() + size()` to anything other than the null terminator. 

I'd say, the presence of a non-const overload (that isn't there due to legacy reasons) and cpp ref's warning means mutations are fine.

dyfer: Thanks for the explanation!

dyfer:
Funny enough, I came across the information that [codecvt is deprecated in c++17 and will be removed in c++26](https://en.cppreference.com/w/cpp/header/codecvt). I guess these conversions will have to be rewritten eventually...

JordanHendersonMusic:
Nice find!

This could be a problem for Chinese, Hebrew, many Cyrillic languages, and a few more maybe. I'm unsure where this would actually be an issue, Interacting with the system is probably one. 

I have no idea what happens if you name a file or folder with unicode characters and try to print in it SC.

dyfer:
> This could be a problem for Chinese, Hebrew, many Cyrillic languages, and a few more maybe.

If you mean that lack of Unicode support would be a problem for these - actually, it's a problem for many many more, including e.g. Central European languages, which use Latin alphabet, but with a different set of extra characters than Western European ones. That's why I created this PR in the first place, to be able to use non-Western characters in system calls. 

BTW this does not limit the UTF-8 support in any way - Unicode characters are [supported](https://doc.sccode.org/Classes/String.html#-ascii) throughout SC (with some exceptions probably). *nixes use UTF-8, so all is good here. This is just about the conversions from UTF-16 (or, [UCS-2](https://unicode.org/faq/utf_bom.html#:~:text=UCS%2D2%20does%20not%20describe,to%20conformantly%20represent%20supplementary%20characters.)? I'm getting confused here), which is the unicode encoding used on Windows.

> I have no idea what happens if you name a file or folder with unicode characters and try to print in it SC.

Currently it works! Feel free to name a folder `åßçąśćół` and print that in SC. 

--

We will just need to swap codecvt implementation with something else, sooner or later. IIUC, we could even swap it with the native system calls, which is what I was trying to use in the earlier version of this PR, but that wouldn't solve our other uses of codecvt. I think the current implementation is in the right place, but will need to be changed. Supposedly one of the issues with codecvt is that it doesn't support conversion errors - what happens when the conversion fails? We'd need to consider that, I believe. The Windows native system calls do report whether the conversion was successful or not.

JordanHendersonMusic:
> UTF-16 or, UCS-2

I was reading about this today. So take this with a pinch of salt.
This is a nice resource https://utf8everywhere.org/

From what I can gather, UTF16 is variable encoding and can store any sized codepoint. UCS2 was made in 1996 and is 16bit but fixed width, meaning 65k characters max. UCS2 doesn't support right to left text and many Chinese characters. 

Windows uses UTF16 with the modern character set. It's bad and they should use UTF8 because it uses more space unnecessarily, but because it's variable width, it can represent whatever characters needed.

`åßçąśćół` I believe these characters are all in UC2 though. I wonder what happens if you give it something outside this character set?

dyfer:
> `åßçąśćół` I believe these characters are all in UC2 though. I wonder what happens if you give it something outside this character set?

Good point about the UCS2 character set. I don't know what happens when you give it something outside of that... Maybe this is when the conversion should fail? :)

In any case, on Windows we can't support anything beyond what the OS does. On other platforms UTF-8 is already well supported.

Issue #6126, opened 2023-11-06
SC fails linking when built with Clang and Qt6 on Linux


dyfer:
## Environment

* SuperCollider version: https://github.com/supercollider/supercollider/tree/topic/qt6-12 
* Operating system: Ubuntu 22.04
* Other details (Qt version, audio driver, etc.): Qt 6.2.4

## Steps to reproduce

See [here](https://github.com/supercollider/supercollider/blob/topic/qt6-12/.github/workflows/actions.yml#L132) and [here](https://github.com/supercollider/supercollider/blob/topic/qt6-12/.github/workflows/actions.yml#L145) for a list of packages to install in order to compile SC with Qt6 on Ubuntu 22.04
```sh
# install qt6 and clang

# configure and build SC"
export CC="clang-14"
export CXX="clang++-14"
cmake -DCMAKE_BUILD_TYPE=Release ..
make install
```

## Expected vs. actual behavior

SC builds and links fine with GCC. However, it fails to link when built with Clang.

Here's the [error from the CI](https://github.com/supercollider/supercollider/actions/runs/6777708616/job/18421854883#step:10:11129)
<details>

```
[ 97%] Linking CXX executable scide
/usr/bin/ld: libscide.a(sc_process.cpp.o): in function `QtConcurrent::RunFunctionTaskBase<ScIDE::ScLanguage::Introspection>::run()':
sc_process.cpp:(.text._ZN12QtConcurrent19RunFunctionTaskBaseIN5ScIDE10ScLanguage13IntrospectionEE3runEv[_ZN12QtConcurrent19RunFunctionTaskBaseIN5ScIDE10ScLanguage13IntrospectionEE3runEv]+0x8e): undefined reference to `QUnhandledException::QUnhandledException(std::exception_ptr)'
/usr/bin/ld: libscide.a(help_browser.cpp.o): in function `ScIDE::HelpBrowser::evaluateSelection(bool)':
help_browser.cpp:(.text+0x26a4): undefined reference to `QWebEnginePage::runJavaScript(QString const&, std::__1::function<void (QVariant const&)> const&)'
/usr/bin/ld: libscide.a(QcWebView.cpp.o): in function `QtCollider::WebView::toHtml(QtCollider::QcCallback*) const':
QcWebView.cpp:(.text+0x7dc): undefined reference to `QWebEnginePage::toHtml(std::__1::function<void (QString const&)> const&) const'
[...]

PR #6133, opened 2023-11-11
Switch Linux Clang to use libstdc++

dyfer:
## Purpose and Motivation

It turns out that when building SC against Qt6 on Linux with Clang (on a WIP qt6 branch), we need to set the flag to use libstdc++. 
I changed CMakeLists.txt to set this option automatically on Linux+Clang, and also added the needed linker options.
Note, this PR is against the current SC build using Qt5, for easier review before submitting the Qt6 update.
Fixes https://github.com/supercollider/supercollider/issues/6126

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- ~~[ ] All tests are passing~~
- [x] Updated documentation


dyfer:
I changed the CMake implementation so that the `SC_CLANG_USES_LIBSTDCPP ` option defaults to `ON` on Linux and updated the docs.

joshpar (approved):

joshpar (approved):

PR #6137, opened 2023-11-17
classlib: IDE Document should set path before autorun

jamshark70:
## Purpose and Motivation

When opening a `/*RUN*/` auto-run file, `thisProcess.nowExecutingPath` is not set so the code in the document can't resolve paths relative to its own location.

E.g. if I have `~/test.scd`:

```
/*RUN*/
thisProcess.nowExecutingPath.debug("path");
```

... and open it, I should get `path: /home/xxx/test.scd` in the post window. Without the fix, it's `path: nil`.

Easy fix.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- ~[ ] Updated documentation~

### DIFF
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -347,7 +347,7 @@ ScIDE {
 	*close {|quuid|
 		this.send(\closeDocument, [quuid]);
 	}
-	
+
 	*save {|quuid, docPath|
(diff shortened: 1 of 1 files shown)


telephon (commented):
Good! I would find it even better if this functionality had an explicit method that is equivalent to `Interpreter::executeFile` and `String::load`, but for strings in the current document. Maybe `Document::interpret` ?

dyfer (approved):
Thanks!

PR #6151, opened 2023-12-04
Add parseOSC method to Int8Array

Spacechild1:
## Purpose and Motivation

This PR adds a new method `asOSC` to `Int8Array` which tries to interpret the bytes as an OSC message/bundle, throwing an exception on failure. This is basically the inverse of `Array.asRawOSC`.

Internally, this is achieved with the new primitive `_OSCBytes_Array`.

The following tests all yield the expected result:
```
// OSC message
[ "/foo", 1, 2, 3 ].asRawOSC.asOSC; // -> [/foo, 1, 2, 3]

// OSC bundle (immediate)
[ nil, [ "/foo", 1, 2, 3 ] ].asRawOSC.asOSC; // -> [nil, [/foo, 1, 2, 3]]
// OSC bundle with timetag
[ 10.5, [ "/foo", 1, 2, 3 ] ].asRawOSC.asOSC; // -> [10.5, [/foo, 1, 2, 3]]
[ 0.0, [ "/foo", 1, 2, 3 ] ].asRawOSC.asOSC; // -> [0.0, [/foo, 1, 2, 3]]
// OSC bundle with multiple elements
[ nil, [ "/foo", 1, 2, 3 ], [ "/bar", "baz" ] ].asRawOSC.asOSC; // -> [nil, [/foo, 1, 2, 3], [/bar, baz]]

// not an OSC message or bundle -> throws exception
Int8Array[ 98, 111, 103, 117, 115 ].asOSC
// -> ERROR: Primitive '_OSCBytes_Array' failed.
// caught exception 'Not an OSC message' in primitive in method Int8Array:asOSC
// ...
```

Personally, I would need this for a new SC extension I have been working on, so I would be very happy if it could be included in the next SC release :)

---

Also fixes the documentation of `Array.asRawOSC`, which incorrectly claimed that the receiver must always be an OSC bundle.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


Spacechild1:
@dyfer, @telephon, @mtmccrea Anyone wants to review this? It's just a few lines of code. The new functions are `prOSCBytes_Array` and `ConvertOSCBundle`. They both leverage the already existing `ConvertOSCMessage` function.

I'm currently developing an extension where I need to parse raw OSC messages. For the moment, I'm doing this purely in sclang, but I'd rather use a primitive. That's why I made this PR.

lang/LangPrimSource/OSCData.cpp
```diff
@@ -652,6 +681,55 @@ static PyrObject* ConvertOSCMessage(int inSize, char* inData) {
...
+            throw std::runtime_error("Bad OSC bundle element size");
+        }
+    }
+
+    VMGlobals* g = gMainVMGlobals;
+    PyrObject* result = newPyrArray(g->gc, numElements + 1, 0, false);
+    PyrSlot* slots = result->slots;
+
+    if (oscTime != 1) {
+        double seconds = (double)oscTime * kOSCtoSecs;
```
JordanHendersonMusic, line 707: Why not a static cast?

lang/LangPrimSource/OSCData.cpp
```diff
@@ -652,6 +681,55 @@ static PyrObject* ConvertOSCMessage(int inSize, char* inData) {
...
+        data += sizeof(int32);
+        PyrObject* bundleElement;
+        if (IsBundle(data)) {
+            bundleElement = ConvertOSCBundle(size, data);
+        } else if (IsMessage(data)) {
+            bundleElement = ConvertOSCMessage(size, data);
+        } else {
+            throw std::runtime_error("Malformed OSC bundle element");
+        }
+        SetObject(slots + i + 1, (PyrObjectHdr*)bundleElement);
```
JordanHendersonMusic, line 724: Is this cast needed? Can't you implicitly convert to a base class pointer? If not, why not a static cast?

JordanHendersonMusic (commented):
This seems useful! I have just a few small queries.

lang/LangPrimSource/OSCData.cpp
```diff
@@ -558,6 +560,32 @@ static int prArray_OSCBytes(VMGlobals* g, int numArgsPushed) {
...
+
+// Try to interpret an Int8Array it as an OSC message/bundle,
+// throwing an exception on failure.
+static int prOSCBytes_Array(VMGlobals* g, int numArgsPushed) {
+    PyrSlot* a = g->sp;
+    const PyrInt8Array* array = slotRawInt8Array(a);
+    int size = array->size;
+    const char* data = (const char*)array->b;
+    if (size < 1)
+        return errFailed;
```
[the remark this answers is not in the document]
Spacechild1: I just did the same as in all the other methods, see `prArray_OSCBytes`, `prNetAddr_MsgSize`, etc. They all just return `errFailed` if the size is 0.

lang/LangPrimSource/OSCData.cpp
Spacechild1: > Can't you implicitly convert to a base class pointer?

Good point! I probably didn't realize that `PyrObject` *inherits* from `PyrObjHdr`... Will remove the cast.

lang/LangPrimSource/OSCData.cpp
Spacechild1: Why not a C-style cast? :)

Spacechild1:
@JordanHendersonMusic Thanks for reviewing!

lang/LangPrimSource/OSCData.cpp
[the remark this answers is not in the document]
JordanHendersonMusic: Ah I see, yeah this is fine then.

JordanHendersonMusic (approved):

capital-G (changes requested):
@Spacechild1 can you add 1-2 unit tests, just to make sure we never break it for some reason.

Everything else looks good to me. Once the tests are included, it is ready to be merged.

Spacechild1:
Thanks! I've never written a unit test for SC. Is there some guide? Or can you quickly show me a (good) example?

capital-G:
I think a minimal example can be found e.g. @ https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestEnv.sc

There is also an extensive guide in the wiki https://github.com/supercollider/supercollider/wiki/Unit-Testing-Guide 

The tests are also run in CI if you don't want to setup the testing environment locally.
As we don't have a test file for `Int8Array`, I think these should go into a new file, so `TestInt8Array.sc`. IIRC you don't need to add the file anywhere, it will be picked up automatically from within the directory.

HMU if you need further assistance :)

telephon:
How does this integrate with the `asOSCArgEmbeddedArray` and `asOSCArgArray` interface?

Spacechild1:
First off, `.asOSC` just uses the existing underlying `ConvertOSCMessage` C++ function which is used to convert incoming OSC message to Array-style messages. I did not write any conversion functions myself!

To answer your question: `asOSCArgEmbeddedArray` wraps the array in `[` and `]` typetag, so
```
['/foo'] ++ [1, 2, 3].asOSCArgEmbeddedArray
```
is really the same as
```
[ '/foo', $[, 1, 2, 3, $] ]
```

`asOSC` simply returns the original message:

```
var msg = ['/foo'] ++ [1, 2, 3].asOSCArgEmbeddedArray;
var raw = msg.asRawOSC;
var msg2 = raw.asOSC;
msg2 == msg; // -> true
```

Spacechild1:
Side note: one aspect where `asRawOSC` and `asOSC` are not symmetric is the handling of boolean arguments. `asRawOSC` converts `true` and `false` to Integers 1 and 0, whereas `asOSC` returns `T` and `F` typetags as `true` and `false`. There could be a `NetAddr.useBooleans` method - similar to `NetAddr.useDoubles` - to tell `asRawOSC` that it should use true booleans, but that would be another PR.

Spacechild1:
@capital-G I've added a bunch of unit tests and they are all passing.

capital-G (approved):
Thanks - the quality of the PR is excellent: readable CPP and sclang code, documented and tested.

Are there any open questions @telephon ? If not, I'd merge it.

telephon:
Just to think this through in context: we have then three messages: 
- `asRawOSC`
- `asOSC`
- `asOSCArgEmbeddedArray`

It would be good if:
- the semantics would work together well
- the simplest would be the most commonly usable

If `Int8Array.asOSC` is the inverse of `Array.asRawOSC`, this seems not be clear from the name, they seem to point to the same direction?

Spacechild1:
> Just to think this through in context: we have then three messages:
> - asRawOSC
> - asOSC
> - asOSCArgEmbeddedArray

I don't think `asOSCArgEmbeddedArray` belongs to this list because it is about OSC arguments, not whole messages/bundles.

> If Int8Array.asOSC is the inverse of Array.asRawOSC, this seems not be clear from the name, they seem to point to the same direction?

Yes, `IntArray.asOSC` is the inverse of `Array.asRawOSC`. But I see what you mean. Initially I thought about `asOSCMsg`, but it also handles OSC bundles, so I just went with `asOSC`. I also thought about `asOSCArray`, but this would be very similar to the existing `asOSCArgArray` method - which does something completely different! - and I thought this could be confusing. If anyone has a better suggestion, please go ahead!

JordanHendersonMusic:
Could you do something like `fromRawOSC`?

telephon:
Ah yes, that makes sense. What do you use it for? Maybe the use helps to find the right name.

capital-G:
> Could you do something like `fromRawOSC`?

Rust uses From and Into to convert *types*, where Into is the canonical approach.

In sclang we don't have `OSC` as a type in sclang (why not actually?! it is what sclang/scsynth is build upon), which makes naming even more ambiguous :/

Maybe something like `~myByteArray.intoOSCMessage`, `~myByteArray.asOSCMessage` or `~myByteArray.parseAsOSCMesage` - but in order to make sense, it would require a `OSCMessage` type, because what else should it return?
IMO the clearest approach would be a `OSCMessage` class with a constructor `fromByteArray`, with a method `asByteArray`/`intoByteArray`. Maybe even 2 classes - `OSCMessage` and `OSCBundleMessage`. But I found this is for some reason not SC canonical.

`asRawOSC` is also very confusing in the first place - it is assuming raw = byte array representation - something like `asOSCBytesArray` could have been clearer.

Naming things in SC is a lost cause :/ As long as it is documented I am fine with any naming, preferring verbosity and explicitness over conciseness and implicitness.

Spacechild1:
> Could you do something like fromRawOSC?

Thanks! I really like this. So this would be:
```
var msg = [ '/foo', 1, 2, 3 ];
var raw = msg.asRawOSC;
var msg2 = Array.fromRawOSC(raw);
```
To me this would be clearest naming so far. The user knows that
a) it has some relation to `asRawOSC`
b) the input is a raw OSC message
c) the resulting type is an Array

I'd need to move the method and unit tests from `Int8Array` to `Array`, but that's not a big deal.

Any objections?

JordanHendersonMusic:
If it's a class method, I think the practice so far has been `newFromT` rather than just `fromT`.

Which raises @capital-G's point, `Array.newFromRawOsc` looks like it accepts the type `RawOsc`.

Could we create a subclass class of `Int8Array` called `RawOSC`? This seems like a  more complete solution and *I think* should be backwards compatible.

If we can't, then I think `Array.fromRawOsc` is a good middle ground.

telephon:
Also, the `*newFrom` interface is connected to the `as` method.
E.g. because there is `Dictionary.newFrom(collecton)`, we can write `[1, 2, 3, 4].as(Dictionary)`.

> Could we create a subclass class of `Int8Array` called `RawOSC`? This seems like a more complete solution and _I think_ should be backwards compatible.

In principle, this should be easy, maybe the quesstion is where you check for correctness (possibly in `*new`).

Again, it would help me a lot if I had a few examples where the functionality is used in code.

> Personally, I would need this for a new SC extension I have been working on, so I would be very happy if it could be included in the next SC release :)

@Spacechild1, do you have something short?

telephon:
> Maybe I could rename it to `parseOSC`? There is at least a precedent with `String.parseJSON` and `String.parseYAML`.

Yes, that is now much clearer, thanks!

> I'm relaying OSC messages as OSC blobs inside other messages; the sender calls `asRawOSC`, the receiver unpacks it with `asOSC`.

So for the helpfile of the method, could you add such a case as a very short example? 

The closest I got was this:

```supercollider

n = NetAddr("127.0.0.1", NetAddr.langPort);
OSCFunc({ |msg| msg.asOSC.postln }, "/test_raw"); // will be parseOSC
a = ["/test_raw", 1, 2, "three"].asRawOSC;
n.sendRaw(a);
```

But it is not a good example because it doesn't show why you'd need it.
Do you have one?

Spacechild1:
> So for the helpfile of the method, could you add such a case as a very short example?

I don't have a simple example. (My personal use case is rather complex.) However, I think the documentation already states very clearly what the method does:

> Tries to interpret the bytes as an OSC message or OSC bundle. Throws an exception on failure. See also link::Classes/Array#-asRawOSC::.
```
// OSC message:
Int8Array[ 47, 115, 95, 110, 101, 119, 0, 0, 44, 115, 105, 105, 105, 115, 105, 0, 100, 101, 102, 97, 117, 108, 116, 0, -1, -1, -1, -1, 0, 0, 0, 1, 0, 0, 0, 1, 102, 114, 101, 113, 0, 0, 0, 0, 0, 0, 7, -87 ].parseOSC;
// OSC bundle:
Int8Array[ 35, 98, 117, 110, 100, 108, 101, 0, 0, 0, 0, 0, 25, -103, -103, -103, 0, 0, 0, 48, 47, 115, 95, 110, 101, 119, 0, 0, 44, 115, 105, 105, 105, 115, 105, 0, 100, 101, 102, 97, 117, 108, 116, 0, -1, -1, -1, -1, 0, 0, 0, 1, 0, 0, 0, 1, 102, 114, 101, 113, 0, 0, 0, 0, 0, 0, 7, -87 ].parseOSC;
// not an OSC message or bundle:
Int8Array[ 98, 111, 103, 117, 115 ].parseOSC;
```

For comparison, here's the complete (original) documention of `asRawOSC`:

> Returns the OSC message as an Int8Array. Receiver must be a bundle.
> ```[0.1, [\s_new, \default, -1, 1, 1, \freq, 1961]].asRawOSC;```

Resp. here's the fixed documentation:

> Returns the OSC message as an Int8Array. Receiver can be a message or a bundle. See also link::Classes/Int8Array#-parseOSC::.
```
[0.1, [\s_new, \default, -1, 1, 1, \freq, 1961]].asRawOSC;
[\s_new, \default, -1, 1, 1, \freq, 1961].asRawOSC; // message
[0.1, [\s_new, \default, -1, 1, 1, \freq, 1961]].asRawOSC; // bundle
```

Spacechild1:
If nobody objects, I will go ahead and rename `asOSC` to `parseOSC` and call it a day.

Spacechild1:
The test-linux CI job got stuck so I canceled it. Does anyone know how I can restart the pipeline?

JordanHendersonMusic:
Hit the `checks' tab on the top of this page, on the right is `re-run all jobs'. I don't think you can restart just the tests.

capital-G:
> Hit the 'checks' tab on the top of this page, on the right is 're-run all jobs'. I don't think you can restart just the tests.

Use re-run failed jobs

image

or use the left side tab

image

Spacechild1:
I've just renamed `asOSC` to `parseOSC` and all unit tests for `Int8Array` pass:
```
UNIT TESTS FOR 'TestInt8Array' COMPLETED
There were no failures
    Suite: TestIOUGens
```
https://github.com/supercollider/supercollider/actions/runs/10702297513/job/29673076383?pr=6151#step:6:2107

So I think this is ready to merge.

---

There are *lots* of test failures, but they are not related to my PR. Ping @JordanHendersonMusic

dyfer:
> Any ideas what's going on?

I tried restarting it again, to no avail.
It seems that the server hangs and prevents new servers from running after `TestCoreUGens: testbufugens`.

<details>
```
FAIL: a TestCoreUGens: test_muladd - "kr2.madd(kr5, kr5)"
11997 of 12000 items in array failed to match. Displaying arrays from index of first failure (0) onwards:
FloatArray[0.35521376132965, 0.23515141010284, 0.21471333503723, 0.0989990234375, 0.20335841178894, 0.31524109840393, 0.30331325531006, 0.65699172019958, 0.29867827892303, 0.62982797622681, 0.43703436851501, 0.55366230010986, 0.76575756072998, 0.20926928520203, 0.44768989086151, 0.071670889854431, 0.9549458026886, 0.94583368301392, 0.66032600402832, 0.37623989582062, 0.28266441822052, 0.065970301628113, 0.65257728099823, 0.1915910243988, 0.76737821102142, 0.34374928474426, 0.20621287822723, 0.1599543094635,...etc...
! = 
0


PASS: a TestCoreUGens: test_bufugens - data->loadCollection->PlayBuf->RecordBuf->loadToFloatArray->data (1 channels)

FAIL: a TestCoreUGens: test_muladd - "kr0.madd(ar0, kr5)"
23999 of 24000 items in array failed to match. Displaying arrays from index of first failure (0) onwards:
FloatArray[0.2691615819931, 0.22649741172791, 0.51876485347748, 0.032893538475037, 0.22922086715698, 0.32748711109161, 0.14557492733002, 0.11669838428497, 0.044102191925049, 0.95873892307281, 0.062782049179077, 0.97345650196075, 0.24437856674194, 0.93069016933441, 0.36867296695709, 0.59128546714783, 0.98839592933655, 0.99936056137085, 0.38911628723145, 0.47886669635773, 0.18049764633179, 0.40171933174133, 0.97072958946228, 0.4852169752121, 0.23992097377777, 0.47950732707977, 0.96206617355347, 0.274365186691...etc...
! = 
0


File '/tmp/1149184133' could not be opened: Format not recognised.

FAIL: a TestCoreUGens: test_bufugens - ERROR: Message '-' not understood.


	Test: test_impulse
TestCoreUGens : setting clientID to 0.
WARNING: Server name already exists: 'TestCoreUGens'. Please use a unique name
Booting server 'TestCoreUGens' on address 127.0.0.1:57110.
JackDriver: client name is 'SuperCollider-01'
JackDriver: connected  system:capture_1 to SuperCollider-01:in_1
JackDriver: connected  system:capture_2 to SuperCollider-01:in_2
JackDriver: connected  SuperCollider-01:out_1 to system:playback_1
JackDriver: connected  SuperCollider-01:out_2 to system:playback_2

*** ERROR: failed to open UDP socket: address in use.
You can use SuperCollider (sclang) to kill all running servers by running `Server.killAll`.
You can also kill scsynth using a terminal or your operating system's task manager.
Server 'TestCoreUGens' exited with exit code 1.
```
</details>

I haven't investigate what the might be the source of this.

capital-G:
> There are _lots_ of test failures, but they are not related to my PR. @JordanHendersonMusic @capital-G Any ideas what's going on?

Yeah, I tried to speed up some unit testing but turns out this makes them ocassionally fail which creates some havok, see #6441 - sorry.
I just merged #6374 which implements the same but in a more stable manner.

IMO we can still merge this branch even w/o passing CI b/c the added test succeed.

Spacechild1:
Thanks for all your input!

PR #6153, opened 2023-12-06
Add `localIP`, `localEndPoint` and `localIPs` class methods to NetAddr

Spacechild1:
- `localIP` returns the local IP address as a String; the optional `network` argument allows to specify a particular network, in case the machine is connected to multiple networks.

- `localEndPoint` returns a `NetAddr` with the local IP and a user provided port number (default: sclang port)

Interestingly, the `localEndPoint` class method already existed, but it was not documented and did not really work. Strangely, it tried to call a class method `langIP`, which does not really exist, thus it would always fail.

I have successfully tested this with
- a single interface with Internet
- a single interface without Internet
- two interfaces (one with Internet, the other without)

Also, `NetAddr.localIP("127.0.0.1")` correctly yields "127.0.0.1".

I implemented this as a result of the following discussion: https://scsynth.org/t/code-to-get-local-ip-on-all-platforms/5791/12?u=spacechild1

For example, this is useful if you want to quickly find out your own (local) IP address, so that other users in a LAN can send messages to you.

---

- `localIPs` returns an Array containing the IP addresses of all (available) network interfaces.
  With the optional `family` argument you can restrict the output to IPv4 addresses, IPv6 addresses, or both.
  Tested successfully on Windows and Linux.

- `matchLangIP` is now implemented in sclang, using the output of `localIPs`

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/NetAddr.schelp
+++ b/HelpSource/Classes/NetAddr.schelp
@@ -47,8 +47,21 @@ close all TCP connections.
 method::broadcastFlag
 Get or set the broadcast flag (whether or not broadcast messages can be sent).
 
+method::localIPs
+Get all local IP addresses (NICs) on this computer.
+
+argument::family
+A Symbol that specifies the desired address family.
+table::
+## code::all:: || all addresses
+## code::ipv4:: || only IPv4 addresses
+## code::ipv6:: || only IPv6 addresses
+::
+
+returns::An link::Classes/Array:: containing all local IP addresses of the desired family.
+
 method::matchLangIP
-Test an IP address to see if it matches that of one of the NICs on this computer.
+Test an IP address to see if it matches any local IP address. See also link::#*localIPs::.
 
 argument::ipstring
 A link::Classes/String:: to test containing an IP number in dot decimal notation (e.g. "192.168.34.56").
--- a/HelpSource/Classes/NetAddr.schelp
+++ b/HelpSource/Classes/NetAddr.schelp
@@ -23,6 +23,21 @@ create new net address using an integer IP number.
 method::langPort
 Get the port sclang is currently listening on (may change after a recompile).
 
+method::localIP
+Get the local IP address as a String.
+
+argument::network
+An optional IP string to select a particular network. This might be necessary if your machine is connected to multiple networks. Typically, you would use the IP address of the router, but it can really be any address within the network.
+
+method::localEndPoint
+Get a NetAddr with the local IP address and a user provided port number.
+
+argument::port
+The port number. If omitted, the sclang port is used.
+
+argument::network
+An optional IP string to select a particular network. See also link::#*localIP::.
+
 method::localAddr
 Get a NetAddr which corresponds to localhost and the port sclang is listening on.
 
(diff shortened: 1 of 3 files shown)


prko:
@Spacechild1 
> Oh, there is a misunderstanding! "ping" can have several meanings:
> 
> In the narrow sense, it means a special network message that is used to test the reachability of a peer; this is what the `ping` command in the terminal does.
> 
> In a broad sense, it is any kind a message that basically says "Hi!" to another party, often expecting a reply. So what I actually meant was to send a broadcast message with `NetAddr("255.255.255.255", ~port)`. (NB: you have to enable broadcasting with `NetAddr.broadcastFlag = true`.)

Thank you for the kind explanation!

@muellmusik Thanks for your comments!

@Spacechild1 
> @muellmusik I just added NetAddr.localIPs to https://github.com/supercollider/supercollider/pull/6153. NetAddr.matchLangIP can now be implemented purely in sclang, using the output of NetAddr.localIPs.

Since you have implemented `.localIPs` in your PR https://github.com/supercollider/supercollider/pull/6153, I think it is better to close this PR. Thank you for your time and effort!

I can now reduce some pages of my own tutorial written in Korean, which is dedicated to explaining how to connect and control multiple sclang over networks, by removing the functions to get localIPs. This part will be more compact and the reader can concentrate more on the topic of this part!

Thanks again!

prko:
I hope these wonderful features will be reviewed and merged before SC3.14 is released.  This is not directly related to audio synthesis and algorithmic composition, but it helps a lot of automated processes that can be useful in a broad sense in algorithmic composition environments.

telephon:
These make a lot of sense. I wonder if there could be a possibility of unit testing – what do you think?

muellmusik:
Hi! Sorry slow on this and thanks for the PR. I’ve assigned myself and will try to have a look. I’ve got gigs next week so will likely be after that.

Spacechild1:
> I wonder if there could be a possibility of unit testing – what do you think?

that would be a bit tricky because you would need to the know the local IP address(es) of the testing machine.

JordanHendersonMusic:
I think this looks good and we can move forward with this, would you mind rebasing so the tests can be rerun?

Spacechild1:
Just rebased onto current develop! For some reason, the `TestNodeProxy_Server:test_synthDef_isReleased_afterFree` test fails on Linux, but I'm pretty sure that's a false positive. @capital-G Seems like our test suite is still a bit fragile...

adcxyz:
very nice! would be great if this made it into 3.14.

dyfer:
Oh this is great! I've used `NetAddr.myIP` from netlib for this, but it's much better if it's built in. Also great that it won't interfere with netlib's existing method name (or the other way around, I guess).

I tested this on macOS with multiple interfaces and on Windows, works as expected.

One consideration: in netlib's `NetAddr.myIP` it's also possible to get the network address for a specified device name (e.g. `en0`). @Spacechild1 do you think it would be worthwhile to also add this here, as a second argument to `localIP`?

JordanHendersonMusic (approved):

prko:
Thanks for merging!!!!!!!!

PR #6156, opened 2023-12-06
fix `slotStrVal` implementation

Spacechild1:
## Purpose and Motivation

In `slotStrVal`, use `memcpy` instead of `strncpy`, since the latter does not terminate the string if the length exceeds the given buffer size! See https://en.cppreference.com/w/cpp/string/byte/strncpy. In general, `strncpy` should always be avoided.

There are a few other occurances of `strncpy` in the SC codebase, but this one has the largest impact, as `slotStrVal` is widely used.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -108,13 +108,15 @@ int slotStrLen(PyrSlot* slot) {
     return -1;
 }
 
-int slotStrVal(PyrSlot* slot, char* str, int maxlen) {
+int slotStrVal(PyrSlot* slot, char* str, size_t maxSize) {
+    assert(maxSize > 0);
     if (IsSym(slot)) {
-        strncpy(str, slotRawSymbol(slot)->name, maxlen);
+        size_t len = sc_min(maxSize - 1, slotRawSymbol(slot)->length);
+        memcpy(str, slotRawSymbol(slot)->name, len);
+        str[len] = 0;
         return errNone;
     } else if (isKindOfSlot(slot, class_string)) {
-        int len;
-        len = sc_min(maxlen - 1, slotRawObject(slot)->size);
+        size_t len = sc_min(maxSize - 1, slotRawObject(slot)->size);
         memcpy(str, slotRawString(slot)->s, len);
         str[len] = 0;
         return errNone;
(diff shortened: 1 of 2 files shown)


lang/LangPrimSource/PyrPrimitive.cpp
```diff
@@ -108,13 +108,14 @@ int slotStrLen(PyrSlot* slot) {
     return -1;
 }
 
-int slotStrVal(PyrSlot* slot, char* str, int maxlen) {
+int slotStrVal(PyrSlot* slot, char* str, int maxSize) {
```
JordanHendersonMusic, line 111: Should we be checking for negative values here as passing a negative to std::memcpy will result in converting it to `size_t`, resulting in a large amount of memory being copied? Better yet, this could be a size_t in the signature?

I know this is incorrect usage of the api, and supercollider doesn't do these checks anywhere else, but making the signature a size_t seems like an easy win because you also get the compiler warnings which can be fixed later.

JordanHendersonMusic (commented):
Thanks! Just a little question about what happens if the int is negative and whether that should be a part of this PR.

lang/LangPrimSource/PyrPrimitive.cpp
Spacechild1: > Better yet, this could be a size_t in the signature?

Yeah, this should really be a `size_t`. Though this wouldn't make the code any safer because if the upstream value is negative, the modulo will just happen here instead of instead `memcpy`. I think we need to trust the caller to use the function correctly. Range checks should only be necessary for user input.

That being said, I can change this to `size_t`, just to set a good example :)

JordanHendersonMusic: I think that is a good idea, because it means you get implicit conversion warnings higher up where you can actually do something about them, or at least reminding users of this function that the conversion will take place.

Another option would be something like `std::span<char>` providing debug range checks ... but that is too much of a change for this...

Spacechild1: Changed to `size_t`!

JordanHendersonMusic (approved):

capital-G:
Thanks for this fix!
I don't know much C++ yet, but explicit null termination of string seems like always a good idea! Merged.

capital-G (approved):

Issue #2363, opened 2016-09-28
LocalIn can have garbled initial input


Sciss:
I'm opening this issue, although I don't have a simple self-contained example to reproduce (I have a ScalaCollider patch now that when instantiating the same synth for long enough, shows the problem).

The symptoms are this:
- at some point, running the same synth, using `play {}`, i.e. each time with a freshly sent `SynthDef`, the initial contents of `LocalIn.kr(8)` are not zero as expected.
- once this happens, you can stop and play over and over again, and the contents remains exactly the same (e.g. a number of values around ` -6.51138e-27`). 
- but then, if you start _a second_ synth, with the identical synth-def code, it has again the correctly zero'ed content.

This leads me to believe that the problem is that some portion of memory is used again and again in the first case, and somewhere there is a missing initial clear of the allocated memory.

Here is the link to the Ctor of LocalIn, although I cannot see any mistake: https://github.com/supercollider/supercollider/blob/f7464ac463b02add1fe490ac3785e8dc222502e7/server/plugins/IOUGens.cpp#L1562

Sciss:
Ok, next step -- a reproducible code:

```scala
def mkSynths() = Seq.fill(8)(play {
  val numVoices   = 4

  val stateIn     = LocalIn.kr(numVoices * 2)
  var voiceFreq   = Vector.tabulate(numVoices)(i => stateIn \ i): GE
  var voiceOnOff  = Vector.tabulate(numVoices)(i => stateIn \ (i + numVoices)): GE
  val voiceNos    = (0 until numVoices): GE
  
  val trPoll      = "poll".tr
  
  var activated   = Vector.fill(numVoices)(0: GE): GE
  
  for (vIdx <- 0 until numVoices) {
    (voiceFreq \ vIdx).poll(trPoll, s"fr-in [$vIdx]")
  }
  
  voiceOnOff = activated  // release unused voices
  val stateOut = Flatten(voiceFreq ++ voiceOnOff)
  LocalOut.kr(stateOut)
})

// execute the following line-by-line
val xs = mkSynths()
xs.foreach(_.set("poll" -> 1))
xs.foreach(_.free())
val ys = mkSynths()
ys.foreach(_.set("poll" -> 1))  // some garbled contents
```

I will try to translate this into sclang.

Sciss:
Well. I found the origin of this.

https://github.com/supercollider/supercollider/commit/58279623b3045debfc16197d3c51435e5a1f56e6

[...]

PR #6159, opened 2023-12-09
Plugins/classlib: Support multiple RandIDs in one SynthDef

jamshark70:
## Purpose and Motivation

See https://scsynth.org/t/different-randseed-for-the-same-synthdef/7248/27, where it was desired to use more than one RNG ID in the same SynthDef. Prior to this change, this was impossible because RandID would set the ID only when the ID input changed. This use case requires setting the ID on every control block, once for one part of the graph, and again for the other part.

However, losing the old behavior would also not be desirable.

So this change adds a new input to RandID, `force`.

- If `force <= 0` (false), RandID behaves the old way, looking for changes in the input.
- If `force > 0` (true), RandID will set the ID in every control block, whether it changed or not.

The default is the old way.

As with a couple of other recent PRs, I would prefer not to have to write a unit test until we are reasonably sure this will be merged. So a unit test isn't yet provided. The approach would be to generate two separate, identically-seeded random number streams and make sure they are the same.

## Types of changes

- Documentation
- New feature
- Breaking change -- minimal impact I think(?)

## To-do list

- [x] Code is tested (informally -- unit test to be written later)
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/plugins/NoiseUGens.cpp
+++ b/server/plugins/NoiseUGens.cpp
@@ -790,9 +790,13 @@ void RandID_Ctor(RandID* unit) {
 
 void RandID_next(RandID* unit, int inNumSamples) {
     float id = ZIN0(0);
+    bool fire = ZIN0(1) > 0;
 
     if (id != unit->m_id) {
         unit->m_id = id;
+        fire = true;
+    }
+    if (fire) {
         uint32 iid = (uint32)id;
         if (iid < unit->mWorld->mNumRGens) {
             unit->mParent->mRGen = unit->mWorld->mRGen + iid;
--- a/HelpSource/Classes/RandID.schelp
+++ b/HelpSource/Classes/RandID.schelp
@@ -18,6 +18,14 @@ method::kr, ir
 argument::id
 The random number generator ID.
 
+argument::force
+A Boolean flag. A positive value is 'true', and indicates that RandID should set the synth's random number generator ID on every evaluation, even if the input ID has not changed. Negative or zero is 'false', meaning that RandID should set the synth's random number generator ID only when the input ID changes. Default is 0 (false), for compatibility with versions prior to the addition of this feature.
+
+"Forcing" is useful in the special case where different parts of the same SynthDef graph need to use different random number generators. See examples below.
+
+note:: code::RandID.ir:: will always evaluate only once, at the start of the synth, so the code::force:: argument has no effect in that case. ::
+
+
 Examples::
 
 code::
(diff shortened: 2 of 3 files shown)


telephon (approved):
Thank you!

jamshark70:
Thanks @telephon for approving -- I just added two unit tests:

- Seems we didn't have one for RandSeed before.
- And a second test for the specific new feature in the PR.

Sciss:
Urgent reminder of https://github.com/supercollider/supercollider/issues/2363 - adding UGen arguments on the server side breaks the binary API and potentially crashes clients and earlier written SynthDefs (accessing unallocated memory). I'd appreciate if this issue was always strongly on the mind when considering editing UGens. We currently don't have a safe way to handle this.

Also, an API change marker should be added to these commits should they be merged.

server/plugins/NoiseUGens.cpp
```diff
@@ -790,9 +790,13 @@ void RandID_Ctor(RandID* unit) {
 
 void RandID_next(RandID* unit, int inNumSamples) {
     float id = ZIN0(0);
+    bool fire = ZIN0(1) > 0;
```
Sciss, line 793: is there a way we can know the number of allocated UGen inputs, so we could write

```c
bool fire = num_ugen_args >= 2 && ZIN0(1) > 0
```
Then we would avoid crashing against clients and synth-defs using the previous ABI.

jamshark70:
Agreed with sciss that these changes should be un-approved until the problem is resolved.

jamshark70:
> adding UGen arguments on the server side breaks the binary API and potentially crashes clients and earlier written SynthDefs (accessing unallocated memory)

In this case, I could not reproduce a crash.

It's correct that we should keep this in mind (and my fault that I didn't think of it). It would also be good to understand the boundary conditions better (e.g., maybe adding a kr input is fine but adding an ar is not fine).

server/plugins/NoiseUGens.cpp
Sciss: Christoph noted that there _is_ already a `numInputs` that could/should be used here.

jamshark70: How about this?

```
    bool fire = (unit->mNumInputs >= 2 ? ZIN0(1) : 0.f) > 0.f;
```

Sciss: Should it be
```c
bool fire = unit->mNumInputs >= 2 ? (ZIN0(1) > 0.f) : false;
```
?

jamshark70: > `bool fire = unit->mNumInputs >= 2 && (ZIN0(1) > 0.f);`

OK, done, passes an inputs-mismatch test.

telephon (changes requested):
James requested an un-approve until the issue is solved. Thank you – done here.

Spacechild1:
I think the concerns have already been adressed with https://github.com/supercollider/supercollider/pull/6159#discussion_r1449720259. To me the code looks fine.

@jamshark70 Can you resolve the conflicts?

PR #6161, opened 2023-12-12
DOC: Topic/char documentation

paum3:
Documenting some minor methods in class `Char`
- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Char.schelp
+++ b/HelpSource/Classes/Char.schelp
@@ -135,7 +135,7 @@ returns:: the object itself.
 
 method::asUnicode
 
-returns:: a single Unicode value. Ascii is a subset uf Unicode.
+returns:: a single Unicode value. Ascii is a subset of Unicode.
 
 method::hash
 
--- a/HelpSource/Classes/Char.schelp
+++ b/HelpSource/Classes/Char.schelp
@@ -108,3 +108,47 @@ code::
 	[i,a = i.asAscii,a.isAlphaNum,a.isPrint,a.isPunct,a.isControl].postln;
 });
 ::
+
+method::isLower
+
+returns:: true if the char is a lowercase letter.
+
+method::isUpper
+
+returns:: true if the char is a uppercase letter.
+
+method::isPathSeparator
+
+returns:: true if the char is a platform specific path separator.
+
+method::isVowel
+
+returns:: true if the char is a vowel.
+
+method::isControl
+
+returns:: true if the char is a control character.
+
+method::asAscii
+
+returns:: the object itself.
+
+method::asUnicode
+
+returns:: a single Unicode value. Ascii is a subset uf Unicode.
+
+method::hash
+
+returns:: a code used to index into a hash table.
+
+method::++
+
+A concatenation of the two chars. The result is link::Classes/String::
+
+method::<
+
+True if the receiver is alphabetically sooner. It is comparing the Unicode values. Uppercase characters have lower Unicode value than lowercase characters.
+
+method::==
+
+True if the receiver equals the argument.
--- a/HelpSource/Classes/Char.schelp
+++ b/HelpSource/Classes/Char.schelp
@@ -119,7 +119,7 @@ returns:: true if the char is a uppercase letter.
 
 method::isPathSeparator
 
-returns:: true if the char is a platform specific path separator.
+returns:: true if the char is the platform specific path separator.
 
 method::isVowel
 
@@ -139,7 +139,7 @@ returns:: a single Unicode value. Ascii is a subset of Unicode.
 
 method::hash
 
-returns:: a code used to index into a hash table.
+returns:: the hash value of the receiver.
 
 method::++
 
@@ -147,7 +147,8 @@ A concatenation of the two chars. The result is link::Classes/String::
 
 method::<
 
-True if the receiver is alphabetically sooner. It is comparing the Unicode values. Uppercase characters have lower Unicode value than lowercase characters.
+returns:: true if the Unicode value of the receiver is less that the Unicode value of the argument. While this is equivalent to alphabetical order when comparing for example two lowercase characters, uppercase characters have lower Unicode values that lowercase characters, and digits have lower values than either.
+
 
 method::==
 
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/Char.schelp
```diff
@@ -108,3 +108,47 @@ code::
...
+
+returns:: the object itself.
+
+method::asUnicode
+
+returns:: a single Unicode value. Ascii is a subset of Unicode.
+
+method::hash
+
+returns:: a code used to index into a hash table.
```
cdbzb, line 142: I think I'd just say: `the hash value of the receiver`

HelpSource/Classes/Char.schelp
```diff
@@ -108,3 +108,47 @@ code::
...
+
+method::++
+
+A concatenation of the two chars. The result is link::Classes/String::
+
+method::<
+
+True if the receiver is alphabetically sooner. It is comparing the Unicode values. Uppercase characters have lower Unicode value than lowercase characters.
```
cdbzb, line 150: how about something like: `returns True if the Unicode value of the receiver is less that the Unicode value of the argument. While this is equivalent to alphabetical order when comparing for example two lowercase chars, note that uppercase characters have lower Unicode values that lowercase characters, and digits have lower than either.`

HelpSource/Classes/Char.schelp
```diff
@@ -108,3 +108,47 @@ code::
...
+
+returns:: true if the char is a lowercase letter.
+
+method::isUpper
+
+returns:: true if the char is a uppercase letter.
+
+method::isPathSeparator
+
+returns:: true if the char is a platform specific path separator.
```
cdbzb, line 122: should probably be
`returns:: true if the char is the platform specific path separator.`
(since it returns false for platform-specific path seperators for other platforms ;) )

cdbzb (commented):
Thanks for this! I learned some things I didn't know which is great. I made a couple of small suggestions - lmk what you think

cdbzb (approved):
Thanks!

HelpSource/Classes/Char.schelp
```diff
@@ -108,3 +108,48 @@ code::
...
+
+returns:: true if the char is a control character.
+
+method::asAscii
+
+returns:: the object itself.
+
+method::asUnicode
+
+returns:: a single Unicode value. Ascii is a subset of Unicode.
```
mossheim, line 138: the function name doesn't make sense because Unicode is not a concrete encoding (encodings of Unicode include UTF-8, UTF-16, etc.). SC strings (and therefore characters) are implicitly UTF-8, but Char objects with values between 128 and 255 (or -128 and -1) will return values that aren't UTF-8 codepoints (they print as the <?> unknown character glyph).

I would just document that this returns "the object itself" like you did for asAscii and optionally document this discrepancy.

HelpSource/Classes/Char.schelp
```diff
@@ -108,3 +108,48 @@ code::
...
+
+method::hash
+
+returns:: the hash value of the receiver.
+
+method::++
+
+A concatenation of the two chars. The result is link::Classes/String::
```
mossheim, line 146: The right hand side doesn't need to be a char. I would document this as exactly what its implementation is: `this.asString ++ that`.

HelpSource/Classes/Char.schelp
```diff
@@ -108,3 +108,48 @@ code::
...
+
+returns:: true if the char is the platform specific path separator.
+
+method::isVowel
+
+returns:: true if the char is a vowel.
+
+method::isControl
+
+returns:: true if the char is a control character.
```
mossheim, line 130: I suggest placing these methods in the order they are defined in https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Common/Core/Char.sc

HelpSource/Classes/Char.schelp
```diff
@@ -108,3 +108,48 @@ code::
...
+
+returns:: the hash value of the receiver.
+
+method::++
+
+A concatenation of the two chars. The result is link::Classes/String::
+
+method::<
+
+returns:: true if the Unicode value of the receiver is less that the Unicode value of the argument. While this is equivalent to alphabetical order when comparing for example two lowercase characters, uppercase characters have lower Unicode values that lowercase characters, and digits have lower values than either.
```
mossheim, line 150: Unfortunately doesn't make a lot of sense because of Char not really having a proper encoding. I would document this as simply `this.ascii < aChar.ascii`. Uppercase characters comparing less than lowercase doesn't apply in most encodings once you add in diacritics (ĕ is codepoint U+0115, Ã is U+00C3 for example), you may want to specify that you're talking about comparing ASCII values specifically.

HelpSource/Classes/Char.schelp
mossheim, line 138: "ASCII" should be written in all-caps as it is elsewhere in the article

mossheim (changes requested):
Thanks! There are a couple of issues, let me know if you have any questions. The Char class is unfortunately a bit broken in SC so it's hard to document accurately; I take the view that it is meant to be a representation of C/C++ 8-bit "char", which means it can represent values which aren't correctly encoded.

HelpSource/Classes/Char.schelp
paum3: ok. thanks

HelpSource/Classes/Char.schelp
paum3: ok. thank

HelpSource/Classes/Char.schelp
paum3: ok. thanks.

HelpSource/Classes/Char.schelp
paum3: ok. thanks.

paum3: I changed what you've proposed earlier: the object itself. (so no ASCII mention)

mossheim (approved):
thanks!

PR #6163, opened 2023-12-18
Add check before installing Quark dependencies

### LINKED ISSUE #6155: Quarks that are dependent on eachother cause infinite installation loop
## Environment

* SuperCollider version: 3.12
* Operating system: MacOS and Ubuntu

## Steps to reproduce

My quarks use full url's ala `https://github.com/madskjeldgaard/Monolithic` and `https://github.com/madskjeldgaard/PolyBuf` so the below example is simplified to make it clearer:

```supercollider
// Quark file in repo A
(
name: "QuarkA",
dependencies: ["QuarkB"]
);

// Quark file in repo B
(
name: "QuarkB",
dependencies: ["QuarkA"]
)
```

I would expect dependencies to be resolved and checked off once they are installed by one of the quarks. 

Instead it goes into an infinite loop where one is causing the other to install and vice versa.

cdbzb:
## Purpose and Motivation

Fixes #6155  - installing Quarks with circular dependencies would hang - I added an "installing" classvar that keeps track of any Quarks in the process of installing and prevents re-installing them as dependencies

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation
- [x ] This PR is ready for review

### DIFF
--- a/SCClassLibrary/Common/Quarks/Quarks.sc
+++ b/SCClassLibrary/Common/Quarks/Quarks.sc
@@ -10,7 +10,7 @@ Quarks {
 		cache,
 		regex,
 		installedPaths,
-		<>installing;
+		installing;
 
 	*install { |name, refspec|
 		var path, quark;
@@ -198,29 +198,27 @@ Quarks {
 			},
 			prev = this.installed.detect({ |q| q.name == quark.name });
 
-		if(prev.notNil and: {prev.localPath != quark.localPath}, {
+		if(prev.notNil and: {prev.localPath != quark.localPath}) {
 			("A version of % is already installed at %".format(quark, prev.localPath)).error;
 			^false
-		});
+		};
 
 		"Installing %".format(quark.name).postln;
 		// add currently installing quark to 'installing' Array
-		installing = (installing.size == 0).if{ quark.bubble }{ installing ++ quark.bubble };
+		installing = if(installing.size == 0) { quark.bubble } { installing ++ quark };
 		quark.checkout();
-		quark.isCompatible().not.if{
+		if(quark.isCompatible().not) {
 			^incompatible.value(quark.name);
 		};
 		quark.dependencies.do{ |dep|
 			var ok;
 			// check if dependency is already installing
-			installing.detect({ |q| q.name == dep.name }).isNil.if{				
+			if(installing.detect({ |q| q.name == dep.name }).isNil) {				
 				ok = dep.install();
-				ok.not.if{
+				if(ok.not) {
 					("Failed to install" + quark.name).error;
 					^false
 				}
-			}{
-				(dep.name + "already installing").postln
 			}
 		};
 		quark.runHook(\preInstall);
--- a/SCClassLibrary/Common/Quarks/Quarks.sc
+++ b/SCClassLibrary/Common/Quarks/Quarks.sc
@@ -210,8 +210,8 @@ Quarks {
 		});
 		quark.dependencies.do { |dep|
 			var ok, alreadyInstalled;
-			alreadyInstalled = this.installed.detect({ |q| q.name == quark.name }).notNil;
-			alreadyInstalled.if{
+			alreadyInstalled = this.installed.detect({ |q| q.name == quark.name}).notNil;
+			alreadyInstalled.not.if{				
 				ok = dep.install();
 				if(ok.not, {
 					("Failed to install" + quark.name).error;
--- a/SCClassLibrary/Common/Quarks/Quarks.sc
+++ b/SCClassLibrary/Common/Quarks/Quarks.sc
@@ -189,7 +189,7 @@ Quarks {
 	*installQuark { |quark|
 		var
 			deps,
-			incompatible = { arg name;
+			incompatible = { |name|
 				(quark.name
 					+ "reports an incompatibility with this SuperCollider version"
 					+ "or with other already installed quarks."
@@ -198,14 +198,14 @@ Quarks {
 			},
 			prev = this.installed.detect({ |q| q.name == quark.name });
 
-		if(prev.notNil and: {prev.localPath != quark.localPath}) {
+		if(prev.notNil and: { prev.localPath != quark.localPath }) {
 			("A version of % is already installed at %".format(quark, prev.localPath)).error;
 			^false
 		};
 
 		"Installing %".format(quark.name).postln;
 		// add currently installing quark to 'installing' Array
-		installing = if(installing.size == 0) { quark.bubble } { installing ++ quark };
+		installing = if(installing.size == 0) { [quark] } { installing ++ quark };
 		quark.checkout();
 		if(quark.isCompatible().not) {
 			^incompatible.value(quark.name);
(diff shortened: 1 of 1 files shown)


cdbzb:
@madskjeldgaard can you please test this when you have a minute? (I don't have any Quarks with circular dependencies handy!

one possible issue: I am only checking for installed Quarks with the same *name* what should if a Quark has a dependency that has a different version than one currently installed?

cdbzb:
...needs more work - I'll re-open when repaired

cdbzb:
OK working - added an `installing` classvar to keep track

madskjeldgaard:
This looks great, thanks so much for taking up this problem. I can't test it myself these days, I'm very sorry, but this quark has a circular dependency (with Monolithic):

https://github.com/madskjeldgaard/PolyBuf

Maybe you can try installing that with your changes?

cdbzb:
thanks @madskjeldgaard yep it works!

SCClassLibrary/Common/Quarks/Quarks.sc
```diff
@@ -9,7 +9,8 @@ Quarks {
 		fetched=false,
 		cache,
 		regex,
-		installedPaths;
+		installedPaths,
+		<>installing;
```
mtmccrea, line 13: Were these getters/setters just for debugging? do they need to be exposed?

SCClassLibrary/Common/Quarks/Quarks.sc
```diff
@@ -203,22 +204,31 @@ Quarks {
 		});
 
 		"Installing %".format(quark.name).postln;
-
+		// add currently installing quark to 'installing' Array
+		installing = (installing.size == 0).if{ quark.bubble }{ installing ++ quark.bubble };
```
mtmccrea, line 208: does appending the quark to `installing` require `.bubble`?

SCClassLibrary/Common/Quarks/Quarks.sc
```diff
@@ -203,22 +204,31 @@ Quarks {
...
+			var ok;
+			// check if dependency is already installing
+			installing.detect({ |q| q.name == dep.name }).isNil.if{				
+				ok = dep.install();
+				ok.not.if{
+					("Failed to install" + quark.name).error;
+					^false
+				}
+			}{
+				(dep.name + "already installing").postln
```
mtmccrea, line 223: Is this posted message useful in practice, or was it for debugging?

SCClassLibrary/Common/Quarks/Quarks.sc
```diff
@@ -203,22 +204,31 @@ Quarks {
 		});
 
 		"Installing %".format(quark.name).postln;
-
+		// add currently installing quark to 'installing' Array
+		installing = (installing.size == 0).if{ quark.bubble }{ installing ++ quark.bubble };
 		quark.checkout();
-		if(quark.isCompatible().not, {
+		quark.isCompatible().not.if{
```
mtmccrea, line 210: I notice this method mixes control flow syntaxes, did you go with this way because it was the "nearest" style?
Better to just convert them to the [recommended](https://github.com/supercollider/supercollider/wiki/Style-Guidelines%3A-SuperCollider#recommendation-use-trailing-closure-syntax-whenever-possible) trailing closure style?

SCClassLibrary/Common/Quarks/Quarks.sc
cdbzb: Thanks for the review @mtmccrea ! If `installing` is nil then yes you do need `.bubble`: 
```
nil ++Quark("Ctk") //  Quark: Ctk
nil ++ Quark("Ctk").bubble // [Quark: Ctk]
```

SCClassLibrary/Common/Quarks/Quarks.sc
cdbzb: I went with `(0==1).if{ true } { false}` throughout - I didn't take the style guideline as insisting on `if(0==1){ true }{ false }` since both forms are trailing closures but I'm happy to change all if you think its best.

SCClassLibrary/Common/Quarks/Quarks.sc
cdbzb: good catch - yes I'll remove

SCClassLibrary/Common/Quarks/Quarks.sc
cdbzb: no you're right - I'll remove

SCClassLibrary/Common/Quarks/Quarks.sc
cdbzb: OK I went ahead and changed all

SCClassLibrary/Common/Quarks/Quarks.sc
```diff
@@ -197,28 +198,35 @@ Quarks {
 			},
 			prev = this.installed.detect({ |q| q.name == quark.name });
 
-		if(prev.notNil and: {prev.localPath != quark.localPath}, {
+		if(prev.notNil and: {prev.localPath != quark.localPath}) {
```
mtmccrea, line 201: I know this wasn't your change, but since you're editing here, spaces pad enclosed curly brackets:
```supercollider
if(prev.notNil and: { prev.localPath != quark.localPath }) { ...
```

SCClassLibrary/Common/Quarks/Quarks.sc
mtmccrea: Ah ok, that makes sense, I was thinking from the perspective of but `installing` always being an array, but it's initial condition is `nil`. But in that case, given the check
```
(installing.size == 0)
``` 
because `nil.size == 0`, `installing` won't ever be `nil` when it hits the false case of `{ installing ++ quark.bubble }`, right?

wrt the use of `bubble`, as I understand `installing` will only ever be a 1D array, and because `bubble` is used for adjusting the shape of a (potentially) multidimensional arrays, perhaps `asArray` would be more direct?

mtmccrea:
Thanks for fixing this, @cdbzb!

There's just some minor comments above, otherwise looks good!

SCClassLibrary/Common/Quarks/Quarks.sc
cdbzb: changed - thanks! so long as we are making style changes I also changed the lone `arg` to pipes

SCClassLibrary/Common/Quarks/Quarks.sc
cdbzb: I decided to go with `[quark]` - my thinking was if I were reading the code I might imagine that `Quark` has an `asArray` method so might not be immediately sure what `Quark.asArray` returns.

SCClassLibrary/Common/Quarks/Quarks.sc
mtmccrea: Great!

SCClassLibrary/Common/Quarks/Quarks.sc
mtmccrea: Sounds good! Love saving those characters, keep it up and we'll be able to use 'em on something real nice one of these days.

mtmccrea (approved):

adcxyz:
Just tested here on macOS, works!

PR #6164, opened 2023-12-23
Topic/slash asterisk pairs in quotes are rewritten differently.

prko:
## Purpose and Motivation
`"/*"` is sometimes confused by sclang, so `/*` is sometimes treated as part of a comment. This leads to errors.
`" +/+ "*"` or `"/" ++ "*"` is better to avoid this problem.

## Types of changes
- Documentation

## To-do list
- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/OSCFunc.schelp
+++ b/HelpSource/Classes/OSCFunc.schelp
@@ -107,7 +107,7 @@ m = NetAddr("127.0.0.1", NetAddr.langPort); // loopback
 m.sendMsg("/chat", "Hello App 1");
 m.sendMsg("/chat", "Hello App 1"); // oneshot gone
 m.sendMsg("/ch?t", "Hello App 1");
-m.sendMsg("/*", "Hello App 1");
+m.sendMsg("/" ++ "*", "Hello App 1");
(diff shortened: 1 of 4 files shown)


prko:
An example of the problem:
```
/*
SoundFile.collect("sounds/*").do { |f| f.path.postln };
*/
1 + 1 // This line is incorrectly parsed as a comment.
*/
1 + 1

/*
SoundFile.collect("sounds" +/+ "*").do { |f| f.path.postln };
*/
1 + 1
*/  //This line is correctly parsed as not a comment.
1 + 1
```
Screenshot 2023-12-28 at 1 32 45

cdbzb:
but the real problem still persists - even with your changes this block:
```
(
/*
SoundFile.collect("sounds/*").do { |f| f.path.postln };
*/
)
```
throws `Open ended comment started on line 1 of interpreted text`

which is a bug.

prko:
OK! Is it better to close this PR and write an issue?

cdbzb:
Yes I think so! Thank you

PR #6165, opened 2023-12-25
Help: Correct wording in Amplitude help (it's not the peak)

jamshark70:
## Purpose and Motivation

Facebook conversation:

> Why does Amplitude return an unexpected value?:
> (
> {
>  Amplitude.ar(SinOsc.ar).poll;
>  Silent.ar
> }.play
> )
> Polled value is around 0.63, but I would expect the value to be 1.

Me: "You're expecting the peak amplitude. Many ways of measuring amplitude (e.g. RMS, which is well standard in audio) reflect the average level rather than the peak. ... Amplitude will return a larger value for louder signals but if you really need the peak, see Peak.ar ."

Them: "Makes sense, although my assumption was based on the description of the UGen: 'Tracks the peak amplitude of a signal.'"

Misleading wording.

## Types of changes

- Documentation

## To-do list
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Amplitude.schelp
+++ b/HelpSource/Classes/Amplitude.schelp
@@ -3,7 +3,7 @@ summary:: Amplitude follower
 categories:: UGens>Analysis>Amplitude
 
 description::
-Tracks the relative amplitude of a signal.
+Tracks the relative amplitude of a signal, using an envelope follower algorithm. An envelope follower converges toward rising amplitude according to code::attackTime::, and toward falling amplitude according to code::releaseTime::. See the plot under Examples.
 
 classmethods::
 
@@ -24,7 +24,24 @@ argument::add
 
 examples::
 
+The first example produces a plot, comparing a fast attack/release against slower settings. The green line (fast settings) approximates the amplitude more quickly, but also oscillates according to the wave shape. The blue line (slow settings) rises more slowly, but oscillates less (more stable measurement).
+
 code::
+(
+{
+	// LFSaw begins at 0, illustrates attack time better
+	var sig = LFSaw.ar(440);
+	var amps = Amplitude.ar(
+		sig,
+		attackTime: [0.001, 0.01],
+		releaseTime: [0.01, 0.2]
+	);
+	[sig] ++ amps
+}.plot
+.superpose_(true)
+.plotColor_([Color.black, Color.green, Color.blue]);
+)
+
 (
 // use input amplitude to control SinOsc frequency
 {
--- a/HelpSource/Classes/Amplitude.schelp
+++ b/HelpSource/Classes/Amplitude.schelp
@@ -3,7 +3,7 @@ summary:: Amplitude follower
 categories:: UGens>Analysis>Amplitude
 
 description::
-Tracks the peak amplitude of a signal.
+Tracks the relative amplitude of a signal.
 
 classmethods::
 


cdbzb:
0.63 is close to the average value of SinOsc.ar.abs (2/pi) - quick test with Pulse.ar(750) gave 0.5... So is Amplitude not actually approximating a moving average of the absolute value of the signal?

jamshark70:
> So is Amplitude not actually calculating a moving average of the absolute value of the signal?

No, it's a lowpass filter, similar to slop~ in Pd or slide~ in Max.

cdbzb:
can you think of a more complete description which is accurate? something like  "low-pass filter used for tracking relative amplitude" ?
(slop~ is described as "slew limiting low pass filter" and ~slide as "filters an input logarithmically [...] particularly useful for envelope following")

mtmccrea:
How about something like:

_Returns the running envelope of the signal, smoothed according to the attack and release times._

Then with a simple example included:
```supercollider
(
{
	var att = 0.001;
	var rel = 0.01;
	var src = LFSaw.ar(440) * -1;
	var env = Amplitude.ar(src, att, rel);
	[src, env]
}.plot.superpose_(true)
)
```
Screenshot 2024-05-02 at 22 55 33

smoge:
> How about something like:
> 
> _Returns the running envelope of the signal, smoothed according to the attack and release times._

I think it could be an opportunity to explain some basic concepts. 

RMS is just a fold (on an arbitrarily long list) using basic mathematical operation.

etc etc

mtmccrea:
> Returns the running envelope of the signal, smoothed according to the attack and release times.

The algorithm that is [used](https://github.com/supercollider/supercollider/blob/7bf46ead06c8c62b5c1ee17963e36e7c228fda63/server/plugins/FilterUGens.cpp#L3514) is an [envelope follower](https://www.cs.princeton.edu/courses/archive/spr05/cos579/DSP/DSP.html), so I think it's an accurate description.

Mentioning it's architecture as using a lowpass filter is informative and could be later in the description if @jamshark70 wants to go into that kind of detail, but it's not necessary. 

> Maybe a link to a dedicated page on amplitude, loudness, and their measurement methods.

I agree this would be a good topic for a Tutorial or Guide. But in lieu of that, I suggest we add `related::` classes to the help doc, and maybe even reference the "See also" list in the description.

```supercollider
Peak
RunningSum#*rms
MovingAverage#*rms
Loudness
Onsets
AmpComp
```

smoge:
> The algorithm that is [used](https://github.com/supercollider/supercollider/blob/7bf46ead06c8c62b5c1ee17963e36e7c228fda63/server/plugins/FilterUGens.cpp#L3514) is an [envelope follower](https://www.cs.princeton.edu/courses/archive/spr05/cos579/DSP/DSP.html), so I think it's an accurate description.

yes, it's accurate. I think I didn't make the point clear. 

"See also" would be a nice enhancement.

jamshark70:
> Mentioning it's architecture as using a lowpass filter is informative and could be later in the description if @jamshark70 wants to go into that kind of detail, but it's not necessary.

If others feel it's helpful to go into that much detail, they're welcome to send a PR to my fork, which I'd merge and then it would be updated here.

mtmccrea:
At the risk of delaying the PR, I'd say we just go for the current clarifying change (with the addition of `related::` links), and anyone is welcome to elaborate more in a separate PR.

mtmccrea:
Feel free also to include the example I mentioned (or any other that's useful), as I think the plot clarifies much more than a mini exposition would. The confusion of the original poster, for example, wrt it returning the peak would have been immediately clarified by the plot.

smoge:
> At the risk of delaying the PR, I'd say we just go for the current clarifying change (with the addition of `related::` links), and anyone is welcome to elaborate more in a separate PR.

I apologize for over-interpreting the goal of the PR. 

@mtmccrea it would be interesting to have a guide "How Do You Measure Loudness?" in more detail, maybe? I can try to collaborate on something in the future.  

However, I noticed we use alternative implementations of, for instance, RMS, which calculates the root mean square filtered input for each sample. Maybe the help files do not describe the real algorithm. 

Usually, RMS is integrated, it measures a certain range of signal levels over some time. Is it by design?

smoge:
Sorry, MovingAverage.rms seems to do it. Although there is an RMS UGen as I described.

jamshark70:
> it would be interesting to have a guide "How Do You Measure Loudness?" in more detail, maybe?

In general, I think SC documentation could provide better translations for DSP and audio engineering concepts -- how to FM/PM, how to calculate RMS etc. -- but that DSP / audio-engineering 101 concepts are not really SC help's responsibility. We already lack certain types of SC-specific documentation, without adding the burden of repeating information that is available online or in textbooks.

smoge:
There are two very different RMS calculations, that's sc-related or dsp 101? But you're right, documentation is fine.

smoge:
I don't remember when this was done, but would be nice to define the priorities for improving documentation.

image

cdbzb (approved):

mtmccrea:
@jamshark70 its not clear whether you intended to update the PR based on the discussion, or whether you'd like to keep it as is.

mtmccrea:
I see I may have confused the thread with this:
>  I'd say we just go for the current clarifying change (with the addition of related:: links), and anyone is welcome to elaborate more in a separate PR.

By this I meant staying focused on the Amplitude help file, and avoid expanding the scope into tutorials, etc. So I'm hoping my suggestion for _this_ help file would be considered. https://github.com/supercollider/supercollider/pull/6165#issuecomment-2094109781

jamshark70:
I went ahead and added a little bit. I think it's fair to give readers who are unfamiliar with envelope following some way to see what the attack and release parameters do.

At this point, I consider my work on this PR finished (unless I made a typo, but I don't think I did).

cdbzb (approved):
thanks @jamshark70!  this looks great

mtmccrea:
Thanks all!

PR #6172, opened 2023-12-30
Fix UGen initialization: LFUGens pt. 2/3

### LINKED ISSUE #4228: InRange, InRect: incorrect output for various input rates
Environment
-----------

* Your SuperCollider version: 3.10.0 and develop [0685eb1]
* Your operating system and version: MacOS 10.12.6

I've included the bugs for both UGens in this issue because the bugs are similar and they're neighbors in the source, so could be fixed together.

TLDR; arguments following `in` are only updated at control or scalar rate, not audio rate.

Expected Behavior [InRange]
-----------------------------
Half way through the block, the InRange should output `1`s.

Current Behavior
----------------
```supercollider
( // InRange
s.waitForBoot{
	{ 
		InRange.ar(
			DC.ar(1500), // input at 1500 (ar)
			200,         // lower bound 200 (ir)
			Line.ar(     // upper bound 1400 > 1600 (ar)
				1400, 
				1600, 
				(s.options.blockSize / s.sampleRate)
			)
		)
	}.loadToFloatArray(
		(s.options.blockSize +1 / s.sampleRate), // store 1 block + 1 sample
		action: { |arr| a = arr }                // save result to a
	);
}
)

a // >> all zeros
```
The upper bound of the range changes over one block, in which time the input signal should cross into range, but the whole first block outputs zeros:
```supercollider
a[0..s.options.blockSize-1].every(_==0) // first block's output is all 0's
```
The correct output only occurs at the beginning of the next block:
```supercollider
a[s.options.blockSize] // next block outputs 1
```

There is only one calculation function for `InRange`, and the bounds only update at `.kr`, :
https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/LFUGens.cpp#L2443-L2448

Expected Behavior [InRect]
-----------------------------
Output should be all `1`s at any input rate.

Current Behavior
-----------------
Only `ar` input rates are correct:
```supercollider
// .ar inputs works:
(
{ InRect.ar(
	DC.ar(1000), // .ar
	DC.ar(250),  // .ar
	Rect(200,200,1000, 300)
)}.loadToFloatArray((5 / s.sampleRate), action: {|arr| arr.postln});
)
// .kr inputs fails:
(
{ InRect.ar(
	DC.kr(1000), // .kr
	DC.kr(250),  // .kr
	Rect(200,200,1000, 300)
)}.loadToFloatArray((5 / s.sampleRate), action: {|arr| arr.postln});
)

// .ir input fails:
(
{ InRect.ar(
	1000,  // .ir
	250,   // .ir
	Rect(200,200,1000, 300)
)}.loadToFloatArray((5 / s.sampleRate), action: {|arr| arr.postln});
)
```
The calc function treats both `x` and `y` inputs as `ar` rates, so there needs to be more calc functions added.
https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/LFUGens.cpp#L2464-L2468
https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/LFUGens.cpp#L2474-L2476

Note also that because the third arg to `InRect` is a `Rect`, the last four inputs could just be read in in the `InRect_Ctor`and stored as member variables. (Does that save any overhead in the calc function?)
https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/LFUGens.cpp#L2469-L2472

mtmccrea:
## Purpose and Motivation

#### This fixes the initialization of a batch of `LFUGens`:
- `VarSaw`: properly reset state on initialization
- `T2A`: fix initialization, clip offset to buflength
- `InRect`: check input rates on x and y, init is OK
- `AmpComp`: call correct calc func in Ctor
#### Tests added for:
- `VarSaw`, `T2A`
#### Cleanup
- `LinExp`: remove repeated code from `LinExp_SetCalc`
- `Wrap`, `Trapezoid`: remove commented-out code
- `Unwrap`:  add note re: missing class definition
- `Cutoff`: remove struct to non-existant UGen
#### Documentation was added to:
- `VarSaw`, `SyncSaw`, `T2K`, `T2A`, `InRect`, `InRange`, `AmpCompA`
#### UGens reviewed and tested, no fix needed:
- `A2K`, `DC`, `Clip`, `Wrap`, `Fold`, `ModDif`

(fixed Issues will be updated here)
- Fixes #4228

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/T2K.schelp
+++ b/HelpSource/Classes/T2K.schelp
@@ -4,13 +4,18 @@ categories:: UGens>Conversion, UGens>Triggers
 related:: Classes/T2A, Classes/K2A, Classes/A2K
 
 description::
-Converts audio rate trigger into control rate trigger, using the maximum trigger in the input during each control period.
+Converts audio rate trigger into control rate trigger, using the maximum trigger level in the input during each control period.
 
 classmethods::
 method:: kr
 
 argument:: in
-input signal.
+An audio rate input signal.
+
+discussion::
+note::
+During initialization of the UGen graph, teletype::T2K:: will only produce a trigger if one is received on the first audio-rate input frame. Any trigger received after the first frame, but within the first block, will not be seen at initialization time. Therefore, downstream UGens will also not receive that missed trigger, which may affect their initialization state. Once the Synth runs, however, teletype::T2K:: will trigger as expected.
+::
 
 examples::
 code::
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -2114,7 +2114,7 @@ void AmpComp_Ctor(AmpComp* unit) {
         unit->m_exponent = -1.f * exp;
         SETCALC(AmpComp_next);
     }
-    AmpComp_next(unit, 1);
+    AmpComp_next_kk(unit, 1);
 }
 
 
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -114,6 +114,7 @@ struct Fold : public Unit {
     float m_lo, m_hi, m_range;
 };
 
+// Note: no classlib definition for Unwrap!
 struct Unwrap : public Unit {
     float m_range, m_half, m_offset, m_prev;
 };
@@ -1963,6 +1964,7 @@ void Clip_Ctor(Clip* unit) {
 }
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
+// Note: no classlib definition for Unwrap!
 
 void Unwrap_next(Unwrap* unit, int inNumSamples) {
     float* out = ZOUT(0);
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -93,11 +93,6 @@ struct XLine : public Unit {
     int mCounter;
 };
 
-struct Cutoff : public Unit {
-    double mLevel, mSlope;
-    int mWaitCounter;
-};
-
 struct LinExp : public Unit {
     float m_dstratio, m_rsrcrange, m_rrminuslo, m_dstlo;
 };
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -1096,13 +1096,16 @@ void VarSaw_Ctor(VarSaw* unit) {
     }
 
     unit->mFreqMul = unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1);
-    float duty = ZIN0(2);
-    duty = unit->mDuty = sc_clip(duty, 0.001, 0.999);
+    double phase = unit->mPhase = sc_wrap(static_cast<double>(ZIN0(1)), 0.0, 1.0);
+    float duty = unit->mDuty = sc_clip(ZIN0(2), 0.001f, 0.999f);
     unit->mInvDuty = 2.f / duty;
     unit->mInv1Duty = 2.f / (1.f - duty);
 
-    ZOUT0(0) = 0.f;
+    VarSaw_next_k(unit, 1);
+
+    unit->mPhase = phase;
+    // other members need not be reset, duty is unchanged because phase is
+    // guaranteed to be in range
 }
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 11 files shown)


mtmccrea:
@JordanHendersonMusic Seeing the squashed diff from that first LFUGens init sample fix PR made it more clear what you were seeing wrt adding `const` to changed code...

So, because this PR is still in the family/source of that companion PR, I've added `const` in to those instances and the ones local to this batch of changes in the [most recent commit](https://github.com/supercollider/supercollider/pull/6172/commits/8aab6836e815791ed05ca23bc3c05d805ffc66c1) to see if it would make sense in context. It still seems a bit "random" in the larger context, unless the scope of that particular type of change is expanded— only local changes -> all Ctors? -> the whole file? Then we have a single file specifying const vars among the whole library of UGens that don't. 

Thoughts?

JordanHendersonMusic:
I wouldn't touch something just to add const, unless it was in a commit by itself so it can be added to the blame ignore list.

I don't think having a single file or function with consts is a problem, it only changes how you read and think about the function by making things easier. It's not a fundamental difference in style or anything.

mtmccrea:
I added tests for the UGens to which there are functional changes, `VarSaw` and `T2A`, anticipating the request ;)
So this is ready for review!

mtmccrea:
Pinging @joshpar for potential review,

and @dyfer requesting this be added to the 3.14 milestone.

It's been waiting in the wings for a while, and generally the init sample fixes build on each other, so would be great to get this in :)

dyfer:
Thanks @mtmccrea 
Could you please resolve conflicts in this PR?

mtmccrea:
> Could you please resolve conflicts in this PR?

Done, thanks! Re-running failed jobs, though that may be a bigger CI issue...

server/plugins/LFUGens.cpp
```diff
@@ -1096,13 +1087,16 @@ void VarSaw_Ctor(VarSaw* unit) {
     }
 
     unit->mFreqMul = unit->mRate->mSampleDur;
-    unit->mPhase = ZIN0(1);
-    float duty = ZIN0(2);
-    duty = unit->mDuty = sc_clip(duty, 0.001, 0.999);
+    double phase = unit->mPhase = sc_clip(static_cast<double>(ZIN0(1)), 0.0, 1.0);
```
elgiano, line 1090: I wonder if we should wrap it instead, so that an initial phase of 1.5 becomes 0.5. It would produce a more similar result to what the unit does now (minus the init value glitch you fixed).

Alternatively, we should write more explicitly in schelp that the value is clipped.

HelpSource/Classes/T2K.schelp
```diff
@@ -10,7 +10,13 @@ classmethods::
 method:: kr
 
 argument:: in
-input signal.
+An audio rate input signal.
+
+discussion::
+note::
+During initialization of the UGen graph, teletype::T2K:: will not forward a trigger to downstream UGens if a trigger is received after the first sample in the first block.
+Once running, however, its first output sample will be a trigger if one is received within the first block.
```
elgiano, line 18: I'm having a hard time understanding these two sentences. Let's perhaps try to rephrase it and add a couple of examples?

elgiano (changes requested):
I had a good read of all of this and tested it, it looks all right to me, I only had a couple of questions.

Thanks for the great work!

HelpSource/Classes/T2K.schelp
mtmccrea: Thanks! I've updated the description, hopefully this reads more clearly.

Regarding examples, I don't think it's necessary, as it's quite a fringe case, and in the rare case that anyone encounters it, this Discussion would be sufficient to tell them what's going on IMO. Let me know though if you disagree!

server/plugins/LFUGens.cpp
mtmccrea: Good call, updated!

elgiano (approved):

PR #6175, opened 2024-01-02
[plugins] EnvGen: fix initialization

mtmccrea:
## Purpose and Motivation
This fixes the initialization of `EnvGen` as well as improves (but doesn't fully fix) a segment duration bug, #6180.

Additionally:
- unit tests added to check initialization by the first output sample.
- fix an `Env.circle` startup behavior

This PR is part of an ongoing effort to [fix UGen initializations](https://github.com/supercollider/rfcs/pull/19), the progress of which is [tracked here](https://github.com/mtmccrea/supercollider/wiki/UGen-Fix-List).

**Note:** I did some cleanup of extraneous print statements and stray comments, as a standalone commit. If this isn't preferred, I can remove it. I think it's helpful to clean up as we go.

## Types of changes

- Bug fixes

## To-do list

- [x] Code is tested
- [x] All tests are passing
  - ~~the state of the CoreUGens unit test is the same as reported in https://github.com/supercollider/supercollider/pull/6035 (all relevant tests passing)~~
  - ~~`TestEnvGen_server:test_shapeHold_setEndValue` fails but is fixed by fixed by #6185~~
- [ ] Documentation updated: n/a

### DIFF
--- a/HelpSource/Classes/Env.schelp
+++ b/HelpSource/Classes/Env.schelp
@@ -455,24 +455,45 @@ Env.cutoff(1, 1, \sine).test(2).plot;
 ::
 
 method::circle
-Creates a new envelope specification which cycles through its values. Unlike the link::#*new:: method, the strong::levels::, strong::times:: and strong::curve:: arrays are all the same size. The loopback time and curve are the last element of those arrays.
+Creates a new envelope which, when used by an link::Classes/EnvGen::, cycles through its values.
 
-For making a instantiated envelope cyclic, you can use the instance method link::#-circle::
+For making an already-created envelope cyclic, you can use the link::#-circle:: instance method, which takes the looparound time and curve as arguments.
 
 argument::levels
 The levels through which the envelope passes.
 
 argument::times
-The time between subsequent points in the envelope, which may be a single value (number), or an array of them. If too short, the array is extended.
+The time between subsequent strong::levels:: in the envelope. The last value is the looparound time.
+The array size of strong::times:: should be the same size of the strong::levels:: array.
 
 argument::curve
-The curvature of the envelope, which may be a single value (number or symbol), or an array of them.  If too short, the array is extended.
+The curvature of the envelope segments. The last value is the looparound curve. See link::#*new:: for valid values.
+The array size of strong::curve:: should be the same size of the strong::levels:: array.
 
+note::
+If a single value is given for strong::times:: or strong::curve::, or if the array is too short, it will be extended by wrapping around the given values
+::
 
 discussion::
+note::
+
+The circular code::Env:: must be instantiated inside the function used to construct the link::Classes/SynthDef::. It will not work not work if it is instantiated outside the function, e.g. passed in as a variable.
+
+Due to the internal initializing the cyclic behavior, there is a one-sample delay of the envelope.
+::
+
 code::
-// Plot the envelope
-{ EnvGen.kr(Env.circle([0.1, 1, 0.3], [0.01, 0.03, 0.06])) }.plot(0.4)
+( // Plot the envelope using *circle and -circle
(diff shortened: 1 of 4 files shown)


mossheim:
@mtmccrea what's the intention with these PRs since the corresponding RFC hasn't been merged yet?

mtmccrea:
The RFC was used as a space for corralling ideas from a number of issues and side threads that had gone on for a long time before leading up to that. So each PR is intended to be reviewed like any other, with the RFC serving as a contextual backdrop so I don't have to explain it with every PR.

The intention of the RFC was also to motivate people to fix this pervasive problem, less a question of _whether_ UGen initialization needed to be fixed (I think there's wide agreement on that). But so far I've been the only one working on it...

mtmccrea:
Note for the Reviewer:
In `EnvGen`, I found that resetting the state of `unit->m_counter` back to `0` could cause the synth not to release (for `doneAction: 2`), so when I'm resetting the UGen state at the tail of the Ctor, I don't reset `m_counter`. I suspect it may have something to do with the fact that `next_k` (which is called for the init sample calculation) decrements the counter:
https://github.com/supercollider/supercollider/blob/8338b0c55cc5d1be73341b09266c45abbb2ab8cd/server/plugins/LFUGens.cpp#L2822

I'd prefer to have a better grasp on why that is the case, and I'll continue looking into it. In any case it would be good if the Reviewer would also take a look at this aspect.

mossheim:
> @mossheim The RFC was used as a space for corralling ideas from a number of issues and side threads that had gone on for a long time before leading up to that. So each PR is intended to be reviewed like any other, with the RFC serving as a contextual backdrop so I don't have to explain it with every PR.
> 
> The intention of the RFC was also to motivate people to fix this pervasive problem, less a question of _whether_ UGen initialization needed to be fixed (I think there's wide agreement on that). But so far I've been the only one working on it.

Got it thanks! I will take a look at them sometime in the next weeks then if someone hasn't already.

mtmccrea:
In the course of fixing another EnvGen unit test, I decided to roll that, as well as the doc changes, into separate PRs. They don't need to wait on this PR to resolve.

mtmccrea:
Failing test will pass once #6185 is merged (why didn't this fail on Windows? 🤔)

mtmccrea:
Rabased now tests are all passing

telephon:
Should we have some more tests for envelope?

dyfer:
@mtmccrea thanks for this! Are you considering adding tests for this or would should we pull this in as is?

mtmccrea:
Thanks for following up with this @dyfer. I'm not actively contributing at the moment, so I don't think I'd be able to add tests soon. So it would be great to see this pulled in, and I can address tests as a separate issue in the future, assuming the current work is acceptable.

mtmccrea:
And FWIW, there were [2 tests updated](https://github.com/supercollider/supercollider/pull/6185) in the course of making this fix, which, while not testing specifically this change, hopefully offers some assurance that EnvGen is implicitly tested throughout many unit tests.

mtmccrea:
I've now added a few tests for the initialization of `EnvGen`. 
I also tested the `circle` methods, which I found had broken when the interface to `Delay1.kr` had changed, so that's fixed and docs touched up as well.

wrt the env duration bug, #6180, this PR _mostly_ fixes it, though there's an edge case regarding the last sample of the last segment, which should be handled in a separate PR, since this one is concerned with proper initialization.

mtmccrea:
There was a spurious test timeout failure in the pitch tracker (macOS), which I can't recreate on my machine and doesn't fail after re-running, so I think it's ok...

Otherwise I think I've addressed all the issues raised by commenters so far and it's ready for review.

telephon (approved):
Thanks, this is a very important fix.

PR #6187, opened 2024-01-07
Refactor IEnvGen initialization

mtmccrea:
## Purpose and Motivation

This fixes the initialization of `IEnvGen`, and removes duplicate code from the constructor by calling the calculation function instead and properly resetting the state.

**Note:** I did some cleanup of extraneous print statements and stray comments, as a standalone commit. If this isn't preferred, I can remove it. I think it's helpful to clean up as we go.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation: n/a

### DIFF
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -27,6 +27,9 @@
 
 #include <boost/align/is_aligned.hpp>
 
+// NaNs are not equal to any floating point number
+static const float uninitializedControl = std::numeric_limits<float>::quiet_NaN();
+
 static InterfaceTable* ft;
 
 struct Vibrato : public Unit {
@@ -3202,8 +3205,6 @@ void IEnvGen_Ctor(IEnvGen* unit) {
     // [dur, shape, curve, level] * numvals
     int numStages = (int)IN0(3);
     int numvals = numStages * 4; // initlevel + (levels, dur, shape, curves) * stages
-    float offset = unit->m_offset = IN0(1);
-    float point = unit->m_pointin = IN0(0) - offset;
     unit->m_envvals = (float*)RTAlloc(unit->mWorld, (int)(numvals + 1.) * sizeof(float));
     ClearUnitIfMemFailed(unit->m_envvals);
 
@@ -3215,38 +3216,13 @@ void IEnvGen_Ctor(IEnvGen* unit) {
         //      Print("val for: %d, %3.3f\n", i, unit->m_envvals[i]);
     }
 
-    //  float out = OUT0(0);
-    float totalDur = IN0(4);
-    float level = 0.f;
-    float newtime = 0.f;
-    int stage = 0;
-    float seglen = 0.f;
-    if (point >= totalDur) {
-        unit->m_level = level = unit->m_envvals[numStages * 4]; // grab the last value
-    } else {
-        if (point <= 0.0) {
-            unit->m_level = level = unit->m_envvals[0];
-        } else {
-            float segpos = point;
-            // determine which segment the current time pointer needs calculated
-            for (int j = 0; point >= newtime; j++) {
-                seglen = unit->m_envvals[(j * 4) + 1];
-                newtime += seglen;
-                segpos -= seglen;
-                stage = j;
-            }
+    unit->m_offset = IN0(1);
+    unit->m_pointin = uninitializedControl; // trigger calculation in next_k
+    unit->m_level = uninitializedControl; // will be set in next_k
 
-            segpos = segpos + seglen;
-            float begLevel = unit->m_envvals[(stage * 4)];
-            int shape = (int)unit->m_envvals[(stage * 4) + 2];
-            int curve = (int)unit->m_envvals[(stage * 4) + 3];
-            float endLevel = unit->m_envvals[(stage * 4) + 4];
-            float pos = (segpos / seglen);
+    IEnvGen_next_k(unit, 1);
 
-            GET_ENV_VAL
-        }
-    }
-    OUT0(0) = level;
+    unit->m_pointin = uninitializedControl; // trigger calculation in next_x
 }
 
 void IEnvGen_Dtor(IEnvGen* unit) { RTFree(unit->mWorld, unit->m_envvals); }
--- a/server/plugins/LFUGens.cpp
+++ b/server/plugins/LFUGens.cpp
@@ -3200,20 +3200,15 @@ void IEnvGen_Ctor(IEnvGen* unit) {
         SETCALC(IEnvGen_next_k);
     }
 
-    // pointer, offset
-    // initlevel, numstages, totaldur,
-    // [dur, shape, curve, level] * numvals
     int numStages = (int)IN0(3);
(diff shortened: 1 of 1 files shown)


elgiano:
Hello! I read through this, just one question: is this PR actually changing any behavior? To me it looks like it's a great refactoring but I don't see any change in how the init sample is produced. Sorry if I missed something, already this confusion is a good reason to refactor it the way you did :D

server/plugins/LFUGens.cpp
```diff
@@ -3228,56 +3231,24 @@ void IEnvGen_Ctor(IEnvGen* unit) {
...
+    unit->m_pointin = uninitializedControl; // trigger calculation in next_k
+    unit->m_level = uninitializedControl; // will be set in next_k
 
-            segpos = segpos + seglen;
-            float begLevel = unit->m_envvals[(stage * 4)];
-            int shape = (int)unit->m_envvals[(stage * 4) + 2];
-            int curve = (int)unit->m_envvals[(stage * 4) + 3];
-            float endLevel = unit->m_envvals[(stage * 4) + 4];
-            float pos = (segpos / seglen);
+    IEnvGen_next_k(unit, 1);
```
mtmccrea, line 3249: `next_k` is used to generate the initial sample, which also advances the state of the UGen.

server/plugins/LFUGens.cpp
```diff
@@ -3228,56 +3231,24 @@ void IEnvGen_Ctor(IEnvGen* unit) {
...
 
-            GET_ENV_VAL
-        }
-    }
-    OUT0(0) = level;
+    unit->m_pointin = uninitializedControl; // trigger calculation in next_x
```
mtmccrea, line 3251: Then this state variable is set to `uninitializedControl` so that once the `next` function is called, the state will be re-calculated.

mtmccrea:
It's been a year since doing this, but a quick look back over it, the change will be in the initial state of the UGen (likely affecting just the first sample). I've commented above in the additions to the code which create this change, in addition to the refactoring.

elgiano:
I'm sorry, I had missed the double assignment here (in code before this PR):
```cpp
float point = unit->m_pointin = IN0(0) - offset;
```
so i thought also the previous code didn't store changes to `m_pointin` during init.

I'm trying to understand if this PR actually changes IEnvGen's output in any way we should document. Please correct me if I'm wrong:

Since IEnvGen caches the env values at Ctor time, it doesn't make a difference if it caches also m_pointin. At first calc, it will get the new m_pointin from input, and if it didn't change, it will output the m_level it also cached during init. But even if it would calculate m_level again (like with this PR), it would get the same m_level value, because envelope vars were not updated in between init and first sample.

So I can't think of any case where this UGen output would change after this PR. @mtmccrea would you have any other case?

As a side note, this is the first UGen I find that is outputting other UGens' init samples (since it caches the envelope at Ctor time). Maybe we could use it to debug init samples in general?

mtmccrea:
I was just updating my response as your wrote...

I also looked back over the `_Ctor` and `next_k` functions and I think your reading of it is correct (it's been over a year since looking at this...). 

After rebasing and running some tests comparing to `develop`, I also didn't catch any behavior change, so in the end it appears that this is indeed just a refactor.

mtmccrea:
In addition to the refactoring, the rationale for calling `next_k`, then setting 
```
unit->m_pointin = uninitializedControl;
```
was to ensure that the output for both the init sample and the first output sample would be calculated by the same code, which also interacts with other state variables.

server/plugins/LFUGens.cpp
```diff
@@ -3197,56 +3200,24 @@ void IEnvGen_Ctor(IEnvGen* unit) {
...
```
elgiano, line 3220: Maybe we shouldn't reset m_pointin at all? I'm thinking that since the env segments are not read again in _next, we can avoid re-calculating (if input "index" didn't change, we can output m_level without reading again). This way this PR would be purely a refactoring, without even introducing the extra read. I don't see any other interaction between m_pointin and internal state, apart from m_level.

If we decide to not reset m_pointin, I think we should add a comment saying that resetting internal state after init is not necessary for IEnvGen.

What do you think?

mtmccrea: Sure that's reasonable. I was following the pattern that this collection of init sample fixes follows:
Init state -> calculate an init sample -> reset state so first sample == init sample.

There are exceptions already and I think leaving a comment as you suggested still reinforces the suggested pattern.

Updated!

server/plugins/LFUGens.cpp
```diff
@@ -3197,56 +3200,23 @@ void IEnvGen_Ctor(IEnvGen* unit) {
...
+    unit->m_offset = IN0(1);
+    unit->m_pointin = uninitializedControl; // trigger calculation in next_k
+    unit->m_level = uninitializedControl; // will be set in next_k
 
-            GET_ENV_VAL
-        }
-    }
-    OUT0(0) = level;
+    IEnvGen_next_k(unit, 1);
+    // state is unchanged by next_k, so no need to reset state after init
```
elgiano, line 3219: great, thank you! however it's still a bit incorrect: next_k updates state, it's the output that doesn't need to be calculated again.... I'm trying to find a more precise wording... maybe something like
```cpp
// we can avoid resetting internal state: 
// since env is not updated in _next, output depends only on pointin. 
// if pointin doesn't change, m_level cached by _next is valid.
```

mtmccrea: how about:

The internal state doesn't need to be reset here: `m_pointin` determines `m_level` and is unchanged between initialization an the first 'next_' call.

mtmccrea: I just thought we could be more direct and concise.

> since env is not updated in _next

I'm not sure what this refers to. env isn't a variable, and the envelope itself of course isn't updated between init and next.

> if pointin doesn't change

we know it doesn't change, so we can just say that because it doesn't change, m_level doesn't either.

How about
```
// No need to reset internal state: 
// The output depends only on pointin, which doesn't change between
// initialization and the `next_` call. So the cached m_level is valid.
```

Thanks for working through these details.

elgiano: Thanks for bearing with me :D 
I think your last version is fine, and anyways what we want to signal is that we are skipping the reset on purpose. For more detailed understanding, one has to read the rest of the code anyways.

For the records, pointin could change between init and _next, in case it's from a UGen which `y(0) != presample` (but if I'm not wrong, we consider this behavior a bug)

mtmccrea: Ok thanks!

> in case it's from a UGen which y(0) != presample

Yes this is considered a bug, and in fact it's the goal of this whole initialization fix effort to resolve it :)

mtmccrea: Updated!

elgiano (approved):
ready to merge! this is effectively only a refactor that doesn't change the unit's output or init sample, but makes code more readable, consistent with init sample conventions, and removes duplication.

thanks @mtmccrea!

Issue #5347, opened 2021-02-10
[plugin API] Make plugin API extendable


Spacechild1:
Currently, every change to the plugin API leads to an ABI breakage, so all existing plugins have to be recompiled.

Of course, ABI breakage is unavoidable when a public struct changes (members added, removed or changed in size).

However, there should be no need to bump the plugin API version everytime we add a new method.

One simple solution is to add a dummy `void *` array` to the end of the interface table. The array is zeroed, so a plugin can simply  check for `NULL` to see if the method is supported. This is necessary so that plugins compiled against a recent sclang version can still work (or at least not crash :-) with previous versions.

```
struct InterfaceTable {
    ... // existing methods
    void *reserved[256]; // set to zero
};
```
When we add an API method, we add it *before* the padding array and reduce the size of the array by one:
```
struct InterfaceTable {
    ... // existing methods
    void (someNewCoolMethod*)(void);
    void *reserved[255];
};
```

There are, of course, more elaborate techniques. For example, the plugin could get the interface table through a function that takes a version argument, so the host can return different interface tables. However, I have the feeling that it only makes things unnecessarily complicated. Most of time we should only need to add functions. *) If we have to *change* things, we can still bump the plugin API version and force a recompilation.

*) plugin API methods can also be used to extend the scsynth host context! Instead of adding new members to `World` and breaking ABI, one could bundle the members in a new struct and return a pointer through a (new) API method. Note that the techniques mentioned in https://github.com/supercollider/supercollider/issues/5346 could also apply to *internal* extensions!

Spacechild1:
There was discussion already on the SC forum: https://scsynth.org/t/scsynth-plugincmd-and-sending-responses/2638/5?u=spacechild1

@scztt feel free to join in!

scztt:
> There are, of course, more elaborate techniques. For example, the plugin could get the interface table through a function that takes a version argument, so the host can return different interface tables. However, I have the feeling that it only makes things unnecessarily complicated. Most of time we should only need to add functions. *) If we have to change things, we can still bump the plugin API version and force a recompilation.

I think this is the right approach. I think the better architecture MIGHT be a pull model, where the plugin requests the `InterfaceTable` with a version number - but this diverges from the mechanism we have now, and we'd need to maintain both indefinitely, so probably not worth it.

----

[...]

Issue #5224, opened 2020-10-28
[servers] porting audio servers for Emscripten/WebAssembly


brianlheim:
## Motivation

Prior discussion on mailing list here: https://www.listarc.bham.ac.uk/lists/sc-dev/msg00227.html

> I know main devs were not particularly keen on spending time on an
> scsynth wasm target when I asked last time, but perhaps there are other
> people out there who would be interested in exploring that avenue?
> 
> My personal interest got renewed after coming back to Scala.js for my
> client, and seeing that the JS back-end compiler these days is rock
> solid. I think it would be fantastic to make sound installations or
> sketches work in the client-side browser. I even hacked up a WebAudio
> UGen for my FScape non-realtime DSP system to test the thing -- see
> https://www.sciss.de/temp/soundprocesses.js/ -- but of course this was
> never designed for this scenario, so first of all, that's all JavaScript
> not Assembly, and it has a rather complicated streaming framework behind
> (Akka Stream) instead of the relatively simple scsynth graph processing;
> so no wonder it is not very efficient in CPU terms.
> 
> I believe the sweet spot is to run the client through JS (that could be
> ScalaCollider/SoundProcesses, supercollider.js, or other languages that
> probably can run in the browser, like Python) and the server through
> wasm. Csound people tell me wasm support was quite a good addition.
> 
> Now, as I've stated repeatedly, my C skills are very small, same for
> LLVM and emscripten, and I don't know how entangled the build targets
> for scsynth vs sclang are. But if anyone sees this as an opportunity to
> experiment, I'm happy to try to learn the necessary stuff to start
> building a branch of scsynth against wasm (if it's doable at all -- but
> why not?).
> 
> Feel free to reply to the thread or ping me directly.

- @Sciss 

## Description of Proposed Feature

this thread is mostly for discussing requirements and resources for the implementation.

brianlheim:
some resources i found:

It looks like Emscripten uses OpenAL as their audio backend: https://emscripten.org/docs/porting/Audio.html

Here's an OpenAL audio API programmer's guide: https://www.openal.org/documentation/OpenAL_Programmers_Guide.pdf

here's info on porting: https://emscripten.org/docs/porting/index.html. some interesting limitations i noted:
- only non-blocking networking calls are allowed
- can't use x86 inline assembly
- virtual filesystem with some limitations due to sandboxing
- function pointers must be cast to the exact type (i think sclang may actually make use of this)
[...]

PR #6191, opened 2024-01-11
improve/fix plugin unloading

Spacechild1:
## Purpose and Motivation

scsynth and supernova already try to unload plugins by searching for an "unload" function. However, this is not documented in the plugin API yet.

Changes:

* add `PluginUnload` macro to `SC_InterfaceTable.h`, so it can be easily used by third-party plugins

* use `PluginUnload` macro for the DiskIO and UI UGens (which actually had a wrong function signature!)

* unload static plugins

* don't forget to call `LoadLibrary()` resp. `dlclose()` in scsynth

## Types of changes

- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/scsynth/SC_Lib_Cintf.cpp
+++ b/server/scsynth/SC_Lib_Cintf.cpp
@@ -85,6 +85,7 @@ void read_section(const struct mach_header* mhp, unsigned long slide, const char
 }
 #endif
 
+#ifdef STATIC_PLUGINS
 extern void IO_Load(InterfaceTable* table);
 extern void Osc_Load(InterfaceTable* table);
 extern void Delay_Load(InterfaceTable* table);
@@ -107,8 +108,16 @@ extern void DynNoise_Load(InterfaceTable* table);
 extern void FFT_UGens_Load(InterfaceTable* table);
 extern void iPhone_Load(InterfaceTable* table);
 
+extern void DiskIO_Unload(void);
+extern void UIUGens_Unload(void);
+#endif // STATIC_PLUGINS
 
 void deinitialize_library() {
+#ifdef STATIC_PLUGINS
+    DiskIO_Unload();
+    UIUGens_Unload();
+#endif // STATIC_PLUGINS
+
 #ifdef _WIN32
     for (void* ptrhinstance : open_handles) {
         HINSTANCE hinstance = (HINSTANCE)ptrhinstance;
@@ -117,8 +126,8 @@ void deinitialize_library() {
             UnLoadPlugInFunc unloadFunc = (UnLoadPlugInFunc)ptr;
             (*unloadFunc)();
         }
+        FreeLibrary(hinstance);
     }
-    // FreeLibrary dlclose(handle);
 #else
     for (void* handle : open_handles) {
         void* ptr = dlsym(handle, "unload");
@@ -126,10 +135,12 @@ void deinitialize_library() {
             UnLoadPlugInFunc unloadFunc = (UnLoadPlugInFunc)ptr;
             (*unloadFunc)();
         }
+        dlclose(handle);
     }
 #endif
     open_handles.clear();
 }
+
 void initialize_library(const char* uGensPluginPath) {
     gCmdLib = new HashTable<SC_LibCmd, Malloc>(&gMalloc, 64, true);
     gUnitDefLib = new HashTable<UnitDef, Malloc>(&gMalloc, 512, true);
--- a/HelpSource/Guides/WritingUGens.schelp
+++ b/HelpSource/Guides/WritingUGens.schelp
@@ -133,6 +133,13 @@ list::
 ## The unit generator destructor is named code::PluginName_Dtor::
 ::
 
+subsection:: The Exit Point
+
+The optional code::PluginUnload():: macro exports a function that is called when the library is deinitialized,
+right before it is unloaded. This is only required in rare cases, typically for resource cleanup that cannot be safely done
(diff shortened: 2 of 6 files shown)


Spacechild1:
@dyfer I've followed the existing formatting, but the linter doesn't like it. What should I do?

mtmccrea:
I linted your changes and there's just some minor formatting changes needed. You'll just need to [run clang-format](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions).

mtmccrea:
If you don't yet have it set up and want to just get tests to pass, here's what the formatted changes look like:

https://github.com/supercollider/supercollider/commit/243e51f06c994a1ec036c7c809b1e03a27f688f0

But for future contributions you'll definitely want to have it set up.

Spacechild1:
> I linted your changes and there's just some minor formatting changes needed. You'll just need to [run clang-format](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions).

I am aware of that. Anyway, I just shut up and do what the linter tells me :)

Spacechild1:
This is only adds a macro, so there is no API/ABI break.

Plugin API/ABI breaks only happen if you modify one of the public structures (e.g. `World`)  or add functions to the interface table (see also https://github.com/supercollider/supercollider/issues/5347).

JordanHendersonMusic:
> modify one or more public structures (e.g. World) by adding/removing members or changing their types [ABI/API break]

I think reordering the members also counts as an ABI break?

Spacechild1:
> I think reordering the members also counts as an ABI break?

Yes! That's very important! I forgot to mention that. Thanks for pointing this out.

Spacechild1:
@capital-G @JordanHendersonMusic Can we merge this?

capital-G:
I don't have any objections if this does not touch the ABI - but I don't have enough expertise (regarding C interfaces) to give this a review.

server/scsynth/SC_Lib_Cintf.cpp
```diff
@@ -107,8 +108,16 @@ extern void DynNoise_Load(InterfaceTable* table);
...
+extern void DiskIO_Unload(void);
+extern void UIUGens_Unload(void);
+#endif // STATIC_PLUGINS
 
 void deinitialize_library() {
+#ifdef STATIC_PLUGINS
+    DiskIO_Unload();
+    UIUGens_Unload();
+#endif // STATIC_PLUGINS
+
```
JordanHendersonMusic, line 120: Can you think of any way to avoid manually specifying all the ugens to unload?

For a non-static library you don't have to do anything, but you have to do this extra step for static ones, which seems likely to be forgotten, leading to odd bugs.

I can't think of a nice way (even with context magic), so perhaps some comments warning users to add it here?

JordanHendersonMusic (changes requested):
This looks good and everything makes sense. Thanks! 

I've got a little comment/suggestion, but it's just a small thing.

Otherwise, were you going to add some documentation for this feature?

server/scsynth/SC_Lib_Cintf.cpp
Spacechild1: > Can you think of any way to avoid manually specifying all the ugens to unload?

Unfortunately, no. Note that you also have to explicitly call all the `*_Load()` functions, so this issue is not specific to this PR.

Spacechild1:
> Otherwise, were you going to add some documentation for this feature?

Well, the whole plugin API is completely uncommented. IMO there should be doxygen comments in all the public API headers, but this is out of scope for this PR :)

JordanHendersonMusic:
Unless there is a place where load is mentioned and unload isn't, personally, I wouldn't bother for now.

Spacechild1:
I've just added a small subsection to https://doc.sccode.org/Guides/WritingUGens.html.

server/scsynth/SC_Lib_Cintf.cpp
JordanHendersonMusic: Hm, I'm actually quite confused by this STATIC_PLUGINS flag.

Seems it doesn't work and needs plugin sources added to the cmake file to work?
https://github.com/supercollider/supercollider/issues/5224#issuecomment-752701112

Also, it seems it is only needed for wasm, which we don't support any way.

So perhaps this flag could just be removed?

...but that's beyond this PR.

JordanHendersonMusic: What do you think @Spacechild1, have I understood this correctly? If so, I'll make an issue and approve this.

Spacechild1: > Seems it doesn't work and needs plugin sources added to the cmake file to work?

Ah yes. In fact, we would need a CMake option that if set to `TRUE` adds a `STATIC_PLUGINS` compile definition and adds all plugin sources to the main executable. FWIW I have opened a feature request: https://github.com/supercollider/supercollider/issues/6505

> Also, it seems it is only needed for wasm, which we don't support any way.

Well, there is already a working proof-of-concept for running scsynth in the browser, so it's definitely something we should keep. In fact, it is useful for all platforms that don't support dynamic plugin loading or don't allow to ship shared libraries (e.g. iOS).

JordanHendersonMusic: Thanks!

JordanHendersonMusic (approved):

Spacechild1:
Anyone wants to merge this? Maybe @telephon, @dyfer or @muellmusik?

PR #6194, opened 2024-01-13
Decay, Decay2: fix UGen initialization

### LINKED ISSUE #6143: Decay and Decay2 initial sample is double what's expected
## Environment

* SuperCollider version: current dev
* Operating system: n/a

## Steps to reproduce

```supercollider
{ Decay.ar(Impulse.ar(200), 1.5/200) }.plot
```

decay-bug

The second impulse performs as expected: there is a small amount of energy that has not decayed away, and the second impulse peak adds exactly 1.0.

The first impulse, however, causes Decay to peak at about 2.0.

I'm fairly sure this is an initial-sample issue. Recently there were some changes about UGen initialization, but blame view on FilterUGens.cpp shows clang reformatting four years ago, and before that, 11, 14 and 21 years ago (and nothing about initialization).

I'm puzzled because I used Decay and Decay2 before, and I don't remember this behavior... but the code has "always been like this" lol

It looks like I can work around it like this:

```
{ Decay.ar(Impulse.ar(200) - (0.5 * Impulse.ar(0)), 1.5/200) }.plot
```

... but this doesn't seem right. If, at every impulse, you're adding in exactly 1 sample of 1.0 (full-scale) energy, then the amount of increase at each impulse should be consistent.

mtmccrea:
## Purpose and Motivation
A majority of UGens do not initialize with a correct initialization sample, and subsequently output an incorrect first sample. This addressed that issue for `Decay` and `Decay2`. Unit test also added.

Fixes #6143

This PR is part of an ongoing effort to [fix UGen initializations](https://github.com/supercollider/rfcs/pull/19), the progress of which is [tracked here](https://github.com/mtmccrea/supercollider/wiki/UGen-Fix-List).

## Types of changes

- Bug fix
- Breaking change - initial sample changes are technically breaking.

## To-do list

- [x] Code is tested
- [x] All tests are passing 
  - added unit test for `Decay2`
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Decay.schelp
+++ b/HelpSource/Classes/Decay.schelp
@@ -1,15 +1,12 @@
 class:: Decay
 summary:: Exponential decay
-related:: Classes/Decay2
+related:: Classes/Decay2, Classes/Integrator
 categories:: UGens>Filters>Linear, UGens>Envelopes
 
 Description::
 
-This is essentially the same as  link::Classes/Integrator::  except that
-instead of supplying the coefficient directly, it is calculated from a 60
-dB decay time. This is the time required for the integrator to lose 99.9%
-of its value or -60dB. This is useful for exponential decaying envelopes
-triggered by impulses.
+An integrating filter which, if given an impulse, will produce an exponentially decaying envelope.
+code::Decay:: is essentially the same as link::Classes/Integrator::, but instead of specifying the integration coefficient directly, it is calculated from the strong::decayTime::—the time required for input signal to attentuate 60 dB, or decay to 99.9% of its value.
 
 classmethods::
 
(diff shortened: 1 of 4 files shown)


joslloand:
Adding some of my own notes here on how to normalize the envelope generated by `Decay2`:

```
(
~t60 = 0.01;
~dur = ~t60;
~gain = 0.0;
// ~riseScale = 0.1;  // short rise
~riseScale = 0.5;  // medium rise
// ~riseScale = 0.9;  // long rise

// normalization factors
~riseFac = ~riseScale / (2 - ~riseScale);
~normFac = 0.5 * ~riseFac.pow(~riseFac / (~riseFac - 1)) / (1 - ~riseFac);

{ Decay2.ar(Impulse.ar(0, mul: ~normFac), ~t60 * ~riseFac, ~t60, ~gain.dbamp) }.plot(duration: ~dur);  // plot linear scale
// { Decay2.ar(Impulse.ar(0, mul: ~normFac), ~t60 * ~riseFac, ~t60, ~gain.dbamp).ampdb }.plot(duration: ~dur);  // plot in dB
)
```

mtmccrea:
Thanks @joslloand, your notes give a very helpful technique. I've incorporated them into the updated help docs for `Decay` and `Decay2`, with a bit of elaboration. Feel free to have a look at the latest changed files.

Note also that with the init sample bug fixed in this PR, the normalization factor won't need to be scaled by `0.5` anymore (see https://github.com/supercollider/supercollider/issues/6143, which will now be fixed).

joslloand:
@mtmccrea, just looking at your todo list above, documentation is now updated. Yes?

mtmccrea:
@joslloand indeed!

HelpSource/Classes/Decay.schelp
```diff
@@ -19,18 +16,78 @@ argument::in
...
+    Decay.ar(
+        in: Impulse.ar(XLine.kr(5,50,10), 0.25),
+        decayTime: 0.2,
+        mul: FSinOsc.ar(300)
+    )
+})
+)
+::
+
+As you might hear in the previous example, when code::Decay:: is used as a modulator, its immediate onset can lead to discontinuities and audible clicks. You can see this in the waveform:
```
telephon, line 50: In the helpfiles (at least in the reference, not in the tutorial), there is a tendency not to address the reader directly. It is a matter of choisce of course! I always avoid it, because I find it a little distracting here. (I just wanted to mention this)

telephon (commented):
so good that these classes finally have a good explanation!

HelpSource/Classes/Decay.schelp
joslloand: @telephon, are you suggesting the use of "we" rather "you" here?

If yes, this is a small detail, and does meet my preferences... but brings up a larger question regarding documentation style.

ATM, the documentation [Style Guidelines](https://github.com/supercollider/supercollider/wiki/Style-Guidelines%3A-SCDocs) don't include a discussion around language and tone.

I'd be up for pressing go here.... but if @mtmccrea is up for changing the voice, I'd be fine with that, too.

(@mtmccrea thanks for all your work on this!!)

mtmccrea: I've updated the language to the passive voice now.

I try to keep track of ideas for schelp style guidelines, so I'll add a note about the voice used as a point for discussion.  One day it may turn into a proposal, but there's a lot to discuss around docs, in some ways more than code style guidelines.

joslloand (approved):
Thanks @mtmccrea!! Looks good!

HelpSource/Classes/Decay.schelp
telephon: > @telephon, are you suggesting the use of "we" rather "you" here?
> 
> If yes, this is a small detail, and does meet my preferences... but brings up a larger question regarding documentation style.

No actually, I wasn't thinking of "we" instead of "you", but sometimes an aspect of passive voice, sometimes giving the voice to the entities we talk about:

Instead of writing:
You can modulate the frequency at audio rate

I would write
The frequency can be modulated at audio rate

Instead of writing:
You should never have a loop without a wait, it will crash

I would write
A loop without a wait will crash

Instead of writing:
Be careful not to forget this argument, otherwise ...

I would write
This argument should always be given, otherwise ...

Instead of writing:
You can equally do A, B, and C

I would write
A, B, and C are equally possible

joslloand:
@mtmccrea, hmm... we seem to be getting a CI job failure on the [macOS universal job](https://github.com/supercollider/supercollider/actions/runs/9508322543/job/26209443169#step:18:63462).

mtmccrea:
@joslloand yes, this is a known issue likely on account of some upstream build dependencies not being updated. This build test was [recently disabled](https://github.com/supercollider/supercollider/pull/6358) until it's brought back up to speed.

joslloand:
> @joslloand yes, this is a known issue likely on account of some upstream build dependencies not being updated. This build test was [recently disabled](https://github.com/supercollider/supercollider/pull/6358) until it's brought back up to speed.

Great, good to know!!

PR #6205, opened 2024-01-29
Add SoundFile:*write that works with multichannel files

JordanHendersonMusic:
There is no mention in the documentation of how a multichannel sound file should be written. This makes it work by default, and provides several error messages.

## Purpose and Motivation

It is not clear how a multichannel sound file should be written from the documentation, nor from SoundFile's interface.
Often, a sound file just wants to be written, this wraps all the logic of handling the file and interlacing the samples into a single method call. 

## Example of use
Creates 20 channels of 1 second of white noise. 
```supercollider
SoundFile.write("/tmp/test3.aiff", { {-1.0.rrand(1.0)} ! 44100} ! 20)
```
Without this changes the user would have to do...
```supercollider
a = { {-1.0.rrand(1.0)} ! 44100} ! 20;
a = a.lace(20*44100).asFloat.as(FloatArray);
f = SoundFile.new.numChannels_(20);
f.openWrite("/tmp/test3.aiff");
f.writeData(a);
f.close;
```
Where the conversion to FloatArray can be tricky if a nested array is passed in. While it might be better to make the change there instead, it is implemented as a primitive. 

## Types of changes
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/SoundFile.schelp
+++ b/HelpSource/Classes/SoundFile.schelp
@@ -34,6 +34,25 @@ Full path to the sound file. Use link::Classes/String#-standardizePath:: to reso
 
 returns:: A new SoundFile instance if successful, or code::nil:: if file open failed. User code should check for code::nil:: before doing anything with the SoundFile object.
 
+method:: write
+Writes an array (or nested array) to a path as a sound file.
+
+argument:: pathName
+The path as a link::Classes/String::.
+
+argument:: array
+The data to write. Can be an array of number, or if multichannel, an array of arrays of multichannel.
+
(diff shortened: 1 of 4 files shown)


telephon (commented):
it would be good to move the `replaceRegexp` implementation in a separate PR. Also I saw a "TODO" in the code, I suppose this will have to be fixed?

JordanHendersonMusic:
Ah sorry, I messed up and forgot to branch before making other changes and two prs got mushed together but only noticed a few days later.

PR #6209, opened 2024-01-31
Make ccache detection an option

a17r:
Arguably it should not even be enabled by default, but at least it should be optional.

## Purpose and Motivation

- Distribution packaging without automagic
- Let ccache be handled by source distribution facilities instead, if requested by user

## Types of changes

- Build system

## To-do list

- [x] Code is tested

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,32 +46,6 @@ CONFIGURE_FILE(
   "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
   IMMEDIATE @ONLY)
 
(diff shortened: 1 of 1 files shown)


dyfer:
@a17r sorry I missed this before.

Would you mind rebasing it on current `develop` for the CI to run with it? Thanks!

dyfer:
@a17r sorry we didn't get around to merging this, but seeing the changes in this PR made me realize we could clean up ccache integration. I made another PR to address this, including adding an option to disable ccache. Thanks again for submitting the PR in the first place.

Closing in favor of https://github.com/supercollider/supercollider/pull/6781.

PR #6212, opened 2024-02-04
ScopeView: allow lissajou for more than 2 chs as overlaid pairs

elgiano:
## Purpose and Motivation

Currently ScopeView's lissajou style limits itself to the first two channels being monitored. It is however useful to be able to plot lissajou curves for more than one channel pair, the simplest way being overlaying them. ScopeView is already able to monitor more than two channels, and it has an option to choose different `waveColors`, which make overlaid XY plots discernible, like in the current `overlay` style.

Before making this PR, I've tried composing two stethoscopes in a `StackLayout`, with transparent backgrounds, but it failed at overlaying plots, perhaps due to the QtCollider implementation drawing a bitmap.

This PR doesn't compromise previous functionality: when scoping a 2 channel bus, XY plots are still 2 channels, and default GUIs have options to change number of channels monitored and their offset, so that switching between adjacent channel pairs is still possible the same way as it was before. However, if anyone is relying on ScopeView hiding every other channel than the first two from XY plots, e.g. expecting to see XY plots only for the first 2 channels when monitoring a 6 channels bus, they would now get more confusing plots, especially if they use default colors.

*Possible use cases*:
- inspecting the stereo image of multiple sub-mixes in the same plot
- my real use case: monitoring time-delay phase-space plots of multiple channels in the same plot

*Documentation*:
I've changed the spots where it says that only the first two channels would be plotted

*Tests*:
I couldn't come up with ideas for tests of this new functionality, as we can't easily check how many channels are being plotted in lissajou style.

## Types of changes

- New feature

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/ScopeView.schelp
+++ b/HelpSource/Classes/ScopeView.schelp
@@ -44,7 +44,7 @@ METHOD:: style
     list::
     ## 0 = the channels are vertically spaced
     ## 1 = the channels are overlaid
-    ## 2 = lissajou; the first two channels are used for 2D plotting (as streams of x and y coordinates).
+    ## 2 = lissajou; each pair of channels is used for 2D plotting (as streams of x and y coordinates).
     ::
 
     argument::
--- a/HelpSource/Classes/Stethoscope.schelp
+++ b/HelpSource/Classes/Stethoscope.schelp
@@ -231,6 +231,21 @@ c.scope;
 // fact that ScopeOut.kr doesn't work yet.)
 ::
 
+SUBSECTION:: Multi-channel Lissajou plots
+code::
+(
+{ 
+	var f = { |l,h| SinOsc.kr(0.1, {pi.rand}).range(l, h) } ;
+	Pulse.ar({ f.(20, 1000) }!6, { f.(0.0, 1) }!6, { f.(0.0, 1.0) }!6) 
+}.play;
+)
+
+c = s.scope;
+c.numChannels = 6;
+c.scopeView.waveColors = [Color.red, Color.red, Color.cyan, Color.cyan, Color.white, Color.white];
+c.style = 2;
+::
+
 SUBSECTION:: Embedded use
 You can pass your own view in to add a stethoscope to it:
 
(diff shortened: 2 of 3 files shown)


elgiano:
here are some pictures of three overlaid phase plots from my current project. Note that these are three different moments in time, i.e. only two plots are overlaid at a given moment.
phaseplots

JordanHendersonMusic (commented):
This looks great! Code just extends what is there safety. 



I do think the existing documentation examples are quite verbose making it a little hard to discover this feature. Would you mind adding another really short example demonstrating this? Just something like:
```supercollider
{ 
	var f = { |l,h| SinOsc.kr(0.1, {pi.rand}).range(l, h) } ;
	Pulse.ar({ f.(20, 1000) }!6, { f.(0.0, 1) }!6, { f.(0.0, 1.0) }!6) 
}.play;

c = s.scope;
c.style = 2;
c.numChannels = 6;
c.scopeView.waveColors = [Color.red, Color.red, Color.cyan, Color.cyan, Color.white, Color.white];
```

elgiano:
thanks @JordanHendersonMusic, I've added your example to Stethoscope's HelpSource, is it where you envisioned it to be? 
I've just changed the order of operations because I think it results in a clearer progression if they are executed one by one.

JordanHendersonMusic:
Did you test that, because I found that setting the colour then changing the style causes the colours to be reset?

Other than that it looks fine, it's just so this feature is a bit more discoverable.

elgiano:
Yeah, I had tested it, and on my linux machine colors don't reset. What OS are you on?

JordanHendersonMusic:
> Yeah, I had tested it, and on my linux machine colors don't reset. What OS are you on?

Ah my mistakes, its changing the number of channels that causes the colour information to be lost.

elgiano:
Oh yeah, that I have noticed as well. Great, we are good to go then! thanks a lot

JordanHendersonMusic (approved):
Looks great! This PR adds a fully backwards compatible feature by just extending on the c++.

PR #6213, opened 2024-02-13
Implement replace regex using boost

JordanHendersonMusic:
## Purpose and Motivation
https://scsynth.org/t/testing-a-draft-of-replaceregexp

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Collections/String.sc
+++ b/SCClassLibrary/Common/Collections/String.sc
@@ -130,6 +130,7 @@ String[char] : RawArray {
 	format { arg ... items; ^this.prFormat( items.collect(_.asString) ) }
 	prFormat { arg items; _String_Format ^this.primitiveFailed }
 	matchRegexp { arg string, start = 0, end; _String_Regexp ^this.primitiveFailed }
+	replaceRegexp { |regex, with| _String_ReplaceRegex ^this.primitiveFailed }
 
 	fformat { arg ... args;
 		var str, resArgs, val, func;
--- a/testsuite/classlibrary/TestStringReplaceRegex.sc
+++ b/testsuite/classlibrary/TestStringReplaceRegex.sc
@@ -1,0 +1,30 @@
+TestStringReplaceRegex : UnitTest {
+	test_replace_simple {
+		this.assertEquals(
+			"foo, foo, bar, foo".replaceRegexp("foo", "car"),
+			"car, car, bar, car"
+		);
+	}
+	test_replace_empty {
+		this.assertEquals(
+			"".replaceRegexp("foo", "car"),
+			""
+		);
+	}
+	test_replace_more_complex {
+		this.assertEquals(
+			"texttextte9xtte823x234t".replaceRegexp("[0-9]", ""),
+			"texttexttexttext"
+		);
+		this.assertEquals(
+			"hello--Remove--rem0ve".replaceRegexp("(r|R)em(o|0)ve", ""),
+			"hello----"
+		);
+	}
+	test_replace_captures {
+		this.assertEquals(
+			"HelLO WoRlD".replaceRegexp("(\\\w)", "\\\L$1"),
+			"hello world"
+		)
+	}
+}
--- a/testsuite/classlibrary/TestStringReplaceRegex.sc
+++ b/testsuite/classlibrary/TestStringReplaceRegex.sc
@@ -27,4 +27,13 @@ TestStringReplaceRegex : UnitTest {
 			"hello world"
 		)
 	}
+	test_no_replace {
+        this.assertEquals("hello".replaceRegexp("w", ""), "hello");
+    }
+    test_wrong_args {
+        this.assertException({"hello".replaceRegexp(nil, "foo")}, PrimitiveFailedError);
+    }
+    test_invalid_regex {
+        this.assertException({"hello".replaceRegexp("\\a[\o", "foo").postln}, PrimitiveFailedError);
+    }
 }
--- a/HelpSource/Classes/String.schelp
+++ b/HelpSource/Classes/String.schelp
@@ -383,6 +383,31 @@ code::
 "[xtz]+nd".matchRegexp("xnd"); // true: any combination of x, t, z
 ::
 
+method::replaceRegexp
+This method is used to replace parts of text.
+
+argument::regex
+A perl regular expression (see link::Classes/String#Regular expressions::) with which to match the caller.
+
+argument::with
+The link::Classes/String:: to replace the found regex with.
+
+returns:: A link::Classes/String::.
+
+code::
+// remove numbers
+"g8et t8ho9se 3num5b89ers ou06t o8f h12er56e!".replaceRegexp("[0-9]", "")
+-> get those numbers out of here!
+
+// remove capital letters
+"HelLO WoRlD".replaceRegexp("(\\\w)", "\\\L$1")
+-> hello world
+
+// remove all capital letter unless at the start of a word
+"HelLO worLD! I weNT tO Paris yeSTErDay.".replaceRegexp("(\\\S)(\\\S*)", "$1\\\L$2")
+-> Hello world! I went to Paris yesterday.
+::
+
 method::findRegexp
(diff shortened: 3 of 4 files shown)


JordanHendersonMusic:
Oops, forgot to remove debug string posts....

JordanHendersonMusic:
Does anyone know if the object previously at the return slot needs to be free'd by the gc?

smoge:
I would guess not because of GC. But I have a question. When the code runs into a problem, like receiving the wrong type of data or encountering an issue while processing a regular expression, is SetNil(slot_this) enough?  if some data to return but then hits a snag, does the cleanup system still works without manually deallocating output_string? In other words, was it assigned to a slot for GC to manage?

(This is not a review, but just a question)

JordanHendersonMusic:
> I would guess not because of GC. But I have a question. When the code runs into a problem, like receiving the wrong type of data or encountering an issue while processing a regular expression, is SetNil(slot_this) enough? if some data to return but then hits a snag, does the cleanup system still works without manually deallocating output_string? In other words, was it assigned to a slot for GC to manage?
> 
> (This is not a review, but just a question)

I have no idea... just copied to conventions in the other primitives :slightly_smiling_face:

JordanHendersonMusic:
Ah you're right, this is a bug and I fill fix soon. I should have read the guide on primitives writing in the help docs a little closer --- it is quite thorough!

JordanHendersonMusic:
I've updated this now with tests and documentation.

prko:
Thank you for this! 
I found the following warnings when opening Help document `String`!

>WARNING: SCDoc: In /Users/prko/SuperCollider/build/Install/SuperCollider/SuperCollider.app/Contents/Resources/HelpSource/Classes/String.schelp
  Method -replaceRegexp has arg named 'findRegexp', but doc has 'argument:: regex'.
WARNING: SCDoc: In /Users/prko/SuperCollider/build/Install/SuperCollider/SuperCollider.app/Contents/Resources/HelpSource/Classes/String.schelp
  Method -replaceRegexp has arg named 'replace', but doc has 'argument:: with'.

prko:
Thank you for this!

prko:
I hope these wonderful features will be reviewed and merged before SC3.14 is released.  This is not directly related to audio synthesis and algorithmic composition, but it helps a lot of automated processes that can be useful in a broad sense in algorithmic composition environments.

JordanHendersonMusic:
@capital-G this commit is mess for some reason... give me a moment.

JordanHendersonMusic:
@capital-G alright it should be good now.

testsuite/classlibrary/TestStringReplaceRegex.sc
```diff
@@ -0,0 +1,30 @@
...
+			"hello--Remove--rem0ve".replaceRegexp("(r|R)em(o|0)ve", ""),
+			"hello----"
+		);
+	}
+	test_replace_captures {
+		this.assertEquals(
+			"HelLO WoRlD".replaceRegexp("(\\\w)", "\\\L$1"),
+			"hello world"
+		)
+	}
```
capital-G, line 29: can you add 

* a test which does "nothing"?
```supercollider
test_no_replace {
    this.assertEquals("hello".replaceRegexp("w", ""), "hello");
}
```

* a test which fails due to wrong args (pseudo code)
```supercollider
test_wrong_args {
    this.assertRaises(SomeExcepition, {"hello".replaceRegexp(nil, "foo")});
}
```

* a test with invalid regex, such as `"hello".replaceRegexp("\\o", "foo");`

SCClassLibrary/Common/Collections/String.sc
```diff
@@ -130,6 +130,7 @@ String[char] : RawArray {
 	format { arg ... items; ^this.prFormat( items.collect(_.asString) ) }
 	prFormat { arg items; _String_Format ^this.primitiveFailed }
 	matchRegexp { arg string, start = 0, end; _String_Regexp ^this.primitiveFailed }
+	replaceRegexp { |regex, with| _String_ReplaceRegex ^this.primitiveFailed }
```
capital-G, line 133: Shoud these args maybe default to an empty string instead of a `nil` value?

capital-G (changes requested):
Nice work - I am no expert in the primitives, but I could easily understand what happens and the test valitade that they are working as expected. I added some remarks though.

SCClassLibrary/Common/Collections/String.sc
JordanHendersonMusic: I'm not sure about that.
 
The `regex` argument needs to be supplied as it doesn't make sense to call this without some pattern.

The `with` argument could default to an empty string but then the method would best be described as `removeRegex` rather than `replaceRegex`. I don't think there is a clean default value here.

testsuite/classlibrary/TestStringReplaceRegex.sc
JordanHendersonMusic: Done!

SCClassLibrary/Common/Collections/String.sc
capital-G: I think a common trait in sclang is to fail silently instead of throwing exceptions where possible. But I also agree that default empty strings don't make much sense here. Maybe @telephon has a good idea here which is idiomatic?

capital-G (approved):

SCClassLibrary/Common/Collections/String.sc
telephon: I would say that keeping all default values nil is the most idiomatic one in this case.

JordanHendersonMusic:
Not a clue. From my experience it seems like some builds always fail **and** there are some tests that are more fragile. I wonder if there is anything in the build process that changes? Perhaps an update of a dependancy?

I've rebased to see if that changes things.

It is often the same UGen tests and occasionally one of the server methods.

capital-G:
The unreliable tests are nuisance and I don't understand why they *always* fail on this specific PR. Maybe we should prioritize making the testing pipeline more reliable? I suggested something in #6362 as a quick hack, but maybe there are other ways?

As long as this fails repeatedly I am hesitant in merging this into develop as this could introduce unstableness in other PRs.

telephon:
All of the tests that failed have to do with: `TestUGen_RTAlloc:test_allUGens`, same like in #6365.

capital-G:
> #6387 does this ease some concern over this PR?

I am almost sure that this PR does not introduce testing problems, but I re-triggered the CI 6 times in a row without a successful run - either we are really unlucky here or the PR has bad 气. I'd simply like to avoid introducing further instability into our fragile CI. 
If this PR is urgent and/or blocks another feature, we can merge it into dev - but until a successful test-run I am a bit hesitant r/n.

I can take a look at #6387 next week if you need some assistance :)

capital-G:
@JordanHendersonMusic do you want to rebase this PR against dev? I think if the pipeline succeeds we can merge this :)

capital-G:
Thanks! Let's merge!

capital-G:
Somehow dev CI is failing b/c of this merge

```
 	[0/1]	TestNodeProxy_Server:test_synthDef_isReleased_afterFree (2.19s)
```

https://github.com/supercollider/supercollider/actions/runs/10631002054/job/29471183472#step:7:3469

Will retrigger the test run as the PR seems to be unrelated to this change?

JordanHendersonMusic:
I've seen that one fail occasionally, rerunning seems to solve it.

capital-G:
Giving it a 4th spin now - maybe we should go for https://github.com/supercollider/supercollider/issues/6362?

JordanHendersonMusic:
Wonder if this could be fixed by adding another server.sync at the top of the function? 

```        
test_synthDef_isReleased_afterFree {
                var numBefore, numAfter;
\\ right here...
server.sync;

                // here we have to make sure that the statusWatcher.numSynthDefs is correct
                // before the experiment, so we force the otherwise regular update ...
                server.sendStatusMsg;
                server.sync;
                numBefore = server.statusWatcher.numSynthDefs;

         
```

Issue #6206, opened 2024-01-30
1. Plotter seems give different result depending OS. 2. .addKr not understood error


prko:
## Environment

* SuperCollider version: 3.12.x and above
* Operating system: all

## Steps to reproduce

Evaluate the code in the first post in the following forum thread:
https://scsynth.org/t/initial-trigger-with-pulsedivider/8847?u=prko

## Expected vs. actual behavior

- Expected:
  SC should plot the same result regardless of the OS where SC is running.
- Actual: 
   - Only SC3.12.0 on Raspberry Pi shows a part of the first impulse:
     - on Raspberry Pi: https://scsynth.org/t/initial-trigger-with-pulsedivider/8847/9?u=prko
     - on Windows and macOS: https://scsynth.org/t/initial-trigger-with-pulsedivider/8847/11?u=prko
  - With SC3.13.0 and above, the code has the following error: `ERROR: Message 'addKr' not understood.`
     https://scsynth.org/t/initial-trigger-with-pulsedivider/8847/9?u=prko

mtmccrea:
Interesting...this relates to the use of ControlNames for variables...
```
{ Impulse.ar(\tFreq.kr(400)) }.plot(0.01)
```

mtmccrea:
Actually, this isn't related to Plotter, it's the routine that's throwing the error:
```
{ Impulse.ar(\tFreq.ar(400)) }.fork //plot(0.01)
```
(same error as above)

prko:
The error occurs in the following line, if I am not mistaken:

https://github.com/supercollider/supercollider/blob/db7eed2a17c361503dbc7f70a557874b6001e3cd/SCClassLibrary/Common/Control/GraphBuilder.sc#L168

> ```
> { Impulse.ar(\tFreq.ar(400)) }.fork //plot(0.01)
> ```
In this case `name` is `tFreq` and its class is `Symbol`. I got this result by inserting the following code before the line168:
```
("name:" + name ++"; name.class:" + name.class).postln;
```
The problem seems to occur with `buildSynthDef.addKr(name, values.unbubble);`.
`buildSynthDef` does not understand `.addKr`...

prko:
> The problem seems to occur with `buildSynthDef.addKr(name, values.unbubble);`.
> `buildSynthDef` does not understand `.addKr`...

Hm...
`buildSynthDef` is `tFreq` and its class is `Symbol`.

mtmccrea:
[...]

PR #6217, opened 2024-02-17
classlib: Function:-asBuffer, -plot: utilize sampleRate. cleanup, update related docs

mtmccrea:
## Purpose and Motivation

These changes stemmed from a bug in which a plotting a function with a short duration could fail if it used `SynthDef.wrap`. This is because the rate of the function was [accessed](https://github.com/supercollider/supercollider/blob/db7eed2a17c361503dbc7f70a557874b6001e3cd/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc#L1319) by `function.value.rate`, which fails when evaluating the wrapped function outside the context of a `SynthDef`. So the rate is now inferred from the `sampleRate` of the buffer, which has to be set explicitly.

This led to a bit of cleanup wrt order of operations and server syncing when creating a Buffer from the Function, disambiguating variables by renaming, and correcting/expanding documentation.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation


mtmccrea:
style update per https://github.com/supercollider/supercollider/issues/6206#issuecomment-2009406491

telephon (approved):

PR #6219, opened 2024-02-18
SCClassLib make SoundFile easier write multichannel files.

JordanHendersonMusic:
## Purpose and Motivation

Quality of life for writing multichannel sound files.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/SoundFile.schelp
+++ b/HelpSource/Classes/SoundFile.schelp
@@ -26,6 +26,31 @@ ClassMethods::
 method::new
 Creates a new SoundFile instance.
 
+method:: writeArray
+Writes an array (or nested array) to a path as a sound file.
+
+argument:: array
+The data to write.
+Can be an array of numbers for a mono audio file, or an array of arrays-of-numbers for a multichannel audio file.
+Does not accept further nested arrays.
+Please note, that supercollider does not currently support very large arrays,
+roughly speaking,
+mono sound files longer than 45 mins at a sampling rate of 44100 will cause supercollider to fail silently.
+
+argument:: pathName
+The path as a link::Classes/String::.
+
+argument:: headerFormat
+See link::Classes/SoundFile#-headerFormat::.
+
+argument:: sampleFormat
+See link::Classes/SoundFile#-sampleFormat::.
+
+argument:: sampleRate
+A number, defaults to 44100.
+
+returns:: The path if successful, or throws if unsuccessful.
+
 method:: openRead
 Try to open the audio file at the given path.
 
--- a/SCClassLibrary/Common/Files/SoundFile.sc
+++ b/SCClassLibrary/Common/Files/SoundFile.sc
@@ -70,6 +70,39 @@ SoundFile {
 		^res
 	}
 
+    *writeArray {|array, pathName, headerFormat, sampleFormat, sampleRate|
+        var file = SoundFile.new;
+
+        if (array.rank > 2) {
+            Error(
+                "Cannot write an array with nested channels."
+                "The array should be formatted as: "
+                "[  [channel 1 samples...], [channel 2 samples...], ...]"
+            ).throw
+        };
+
+        // Assign if not nil.
+        headerFormat !? file.headerFormat_(_);
+        sampleFormat !? file.sampleFormat_(_);
+        sampleRate !? file.sampleRate_(_);
+
+        file.numChannels_(if(array.rank <= 1, { 1 }, { array.shape[0] }));
+
+        if(file.openWrite(pathName).not) {
+            Error("Could not open file at path:" + pathName).throw
+        };
+
+        // Interlace the samples and write, always close the file.
+        // Format will be: L1, R1, C1, L2, R2, C2 ...
+        protect {
+            var size = array.shape.product;
+            var flattenedArray = array.lace(size); // lace is faster than .flop.flat.
+            var data = FloatArray.newFrom(flattenedArray.asFloat);
+            file.writeData(data)
+        } { file.close };
+        ^pathName
+    }
+
 	openRead{ arg pathName;
 		path = pathName ? path;
 		^this.prOpenRead(path);


SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,36 @@ SoundFile {
 		^res
 	}
 
+	*write {|pathName, array, headerFormat, sampleRate, sampleFormat|
+        var file = SoundFile.new;
+
+        if(array.isKindOf(SequenceableCollection).not,
```
telephon, line 77: I would suggest the following strategy: convert the array to a `FloatArray` at the beginning of the method. Do not check for type, but any object that won't do it, will throw an error.

Also nested arrays will be caught by this: `Array[1, [2, 2], 3].as(FloatArray)`. If errors are not informative enough, I think they should be fixed in the repective conversion (here, in FloatArray).

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,36 @@ SoundFile {
...
+        sampleFormat !? file.sampleFormat_(_);
+        sampleRate !? file.sampleRate_(_);
+
+        file.numChannels_(if(array.shape.size == 1, 1, {array.shape[0]}));
+
+        if(file.openWrite(pathName).not,
+            {Error("Could not open file at path:" + pathName).throw});
+
+        // interlace the samples and write
+        file.writeData(FloatArray.with(*array.lace(array.shape.reduce('*')).asFloat));
```
telephon, line 98: I'm not sure if this conversion with `lace` will turn out to be a curse later on. At least it is very implicit.

telephon (changes requested):
Thank you, the method is very useful!

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,33 @@ SoundFile {
 		^res
 	}
 
+	*write {|pathName, array, headerFormat, sampleRate, sampleFormat|
+        var file = SoundFile.new;
+
+        if(array.shape.size > 2,
+            {Error(
+                "Cannot write an array with nested channels."
+                + "The array should be formatted as: "
```
telephon, line 80: you can leave out the `+`, sclang will combine subsequent lines

SCClassLibrary/Common/Files/SoundFile.sc
JordanHendersonMusic: After a quick benchmark, flop.flat is about 50x slower.
```supercollider
d = [];

[10, 100, 1000, 10000, 100000].do{|n|
	a = {{0.0.rrand(1.0)} ! n} ! 2;
	10.do{
		var dur_lace = {
			a.lace(a.shape.reduce('*'));
			//FloatArray.with(*a)
		}.bench();
		
		var dur_flop = {
			a.flop.flat;
			//a.as(FloatArray)
		}.bench();
		
		d = d.add(dur_flop / dur_lace ); 
	}
};

d.mean
```

telephon: Ah good to know! I had thought that flat is a primitive like flop. The advantage of `flop` is that you are guaranteed to have the right shape on the outermost level.

Also I've added a comment, because I mistook the documentation. 

> Also, this doesn't check that the shape of the array is regular, meaning all nested arrays are the same size. Do you think it is worth adding a check for that?

If the result would be a corrupted sound file, yes.

HelpSource/Classes/SoundFile.schelp
```diff
@@ -26,6 +26,24 @@ ClassMethods::
...
 Creates a new SoundFile instance.
 
+method:: write
+Writes an array (or nested array) to a path as a sound file.
+
+argument:: pathName
+The path as a link::Classes/String::.
+
+argument:: array
+The data to write. Can be an array of number, or if multichannel, an array of arrays of multichannel.
```
telephon, line 36: It is a little unclear what is expected. Maybe:
"Can be an array of numbers for a mono audio file, or an array of arrays of numbers for a multichannel audio file.

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,39 @@ SoundFile {
...
+            array = array.lace(array.shape.reduce('*')); // interlace the channels
+        };
+
+        headerFormat !? file.headerFormat_(_);
+        sampleFormat !? file.sampleFormat_(_);
+        sampleRate !? file.sampleRate_(_);
+
+        file.numChannels_(if(array.rank <= 1, 1, {array.shape[0]}));
+
+        if(file.openWrite(pathName).not) {
```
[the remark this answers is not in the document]
telephon: if not, make sure the conversion to `FloatArray` happens before the write. Otherwise an error will leave an open file hanging around.

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,39 @@ SoundFile {
...
+        // multichannel audio
+        if(array.rank == 2) {
+            array = array.lace(array.shape.reduce('*')); // interlace the channels
+        };
+
+        headerFormat !? file.headerFormat_(_);
+        sampleFormat !? file.sampleFormat_(_);
+        sampleRate !? file.sampleRate_(_);
+
+        file.numChannels_(if(array.rank <= 1, 1, {array.shape[0]}));
```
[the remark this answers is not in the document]
JordanHendersonMusic: Did not know this, assumed the raw int would be quicker! Fixed.

SCClassLibrary/Common/Files/SoundFile.sc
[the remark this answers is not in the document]
JordanHendersonMusic: There is one, but it only works with openRead. Might have a look at that separately from this.

I've wrapped the close call in `protect` for now.

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,39 @@ SoundFile {
...
+                "[  [channel 1 samples...], [channel 2 samples...], ...]"
+            ).throw
+        };
+
+        // multichannel audio
+        if(array.rank == 2) {
+            array = array.lace(array.shape.reduce('*')); // interlace the channels
+        };
+
+        headerFormat !? file.headerFormat_(_);
```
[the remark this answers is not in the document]
JordanHendersonMusic: I've added a comment to this.

HelpSource/Classes/SoundFile.schelp
JordanHendersonMusic: God that's awful English!
Fixed.

SCClassLibrary/Common/Files/SoundFile.sc
JordanHendersonMusic: By using lace the values will wrap around to the beginning, meaning the sound file is always valid and the shape of the array is guaranteed, but the user probably didn't mean for this to happen. Should this throw an error?

Here would be the check, unless there is some form of 'isRegularShape' method on array that I haven't heard of....
```supercollider
if(array.rank == 2) {
	var first_size = array[0].size();
	array.do{|channel, index|
		if (channel.size() != first_size) {
			Error("All channels should have the same size, first channels's size was" + first_size + "but channel number" + index + "had a size of" + channel.size()).throw
		}
	}
}
```

JordanHendersonMusic: I've put the lace back at the end as well because mutating it changed the number of channels making the rest of the logic stateful and confusing.

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,38 @@ SoundFile {
 		^res
 	}
 
+    *write {|pathName, array, headerFormat, sampleRate, sampleFormat|
```
telephon, line 74: Just one last thing: one can imagine that one may want to call `write` on an instance of SoundFile, to write back its data to a new path, or after some changes. Then this new `*write` method on the class would be a little skewed, because it writes an array.

Maybe call the method `*writeArray`? 

Even better it would be if one could make a new instance of SoundFile from the array and then write that SoundFile instance. But I am not sure right now if this is too complicated to implement.

JordanHendersonMusic: All done.

Thanks for you help ironing this out, there were far more issues than I expected, but I think the code look *much* better for it! Thanks again, J.

SCClassLibrary/Common/Files/SoundFile.sc
JordanHendersonMusic: Is this okay as it is now?

HelpSource/Classes/SoundFile.schelp
```diff
@@ -35,6 +35,7 @@ The path as a link::Classes/String::.
 argument:: array
 The data to write. Can be an array of numbers for a mono audio file, or an array of arrays-of-numbers for a multichannel audio file.
 Does not accept further nested arrays.
+Please note, that supercollider will break with large arrays, mono sound files longer that about 45 mins will not work.
```
telephon, line 38: "supercollider will break"? Maybe better: "Please note that  arrays may not work if they are too large, for example mono sound files longer than about 45 mins." 

(at what sample rate?) (maybe number of items?)

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,38 @@ SoundFile {
 		^res
 	}
 
+    *writeArray {|pathName, array, headerFormat, sampleRate, sampleFormat|
```
telephon, line 74: actually, you could put the array first and then the pathName. Then one could still add an automatic pathName later (like in Recorder.makePath), and you could write `SoundFile.writeArray(array);` Also `pathName, headerFormat, sampleFormat` always come in a row.

SCClassLibrary/Common/Files/SoundFile.sc
```diff
@@ -70,6 +71,38 @@ SoundFile {
...
+
+        if(file.openWrite(pathName).not) {
+            Error("Could not open file at path:" + pathName).throw
+        };
+
+        // Interlace the samples and write, always close the file.
+        // Format will be: L1, R1, C1, L2, R2, C2 ...
+        protect {
+            file.writeData(
+                FloatArray.newFrom(array.lace(array.shape.reduce('*')).asFloat)
```
telephon, line 100: stylistic (using variable name to convey intent):

```
var size = array.shape.reduce('*');
var flattenedArray = array.lace(size); // .lace is faster than .flop.flat here
var data = FloatArray.newFrom(flattenedArray.asFloat);
file.writeData(data)
```
(if you like, you could use `product` instead of `reduce('*')`, but this is a matter of taste.

telephon (changes requested):

telephon (approved):

capital-G (approved):

PR #6225, opened 2024-02-27
Add primitive `isRectangular` to array.

JordanHendersonMusic:
## Purpose and Motivation
Checks that array is regular is shape (not [jagged](https://en.wikipedia.org/wiki/Jagged_array#:~:text=In%20computer%20science%2C%20a%20jagged,edges%20when%20visualized%20as%20output.)).

I needed this in SoundFile:write to make sure all the channels were the same size, while this is overkill for that solution, functions like Array:size and others make this assumption. 

Implementing this is sclang would be significantly slower.

However, by doing it as a primitive means it only works on Array, and not List. This is because List overloads what a 'size' is. 

Examples:
```supercollider
[].isRegular == true;

(0..10).isRegular == true;

[[1,2], [3,4]].isRegular == true;

[[1,2], [3]].isRegular == false;
// yet...
[[1,2], [3,4]].shape == [[1,2], [3]].shape;

[[1], []].isRegular == false;
```

## Types of changes

- New primitive 

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/ArrayedCollection.schelp
+++ b/HelpSource/Classes/ArrayedCollection.schelp
@@ -52,6 +52,20 @@ code::
 [4, 5, 6].maxSize; // gosh
 ::
 
+method::isRectangular
+Returns true if all nested sub arrays are the same size and all elements have the same depth.
+This is a requirement of several nested array algorithms and formats, notably multichannel audio files.
+
+Example:
+code::
+[1, 2, 3].isRectangular // true
+[[1, 2], [3, 4]].isRectangular // true
+
+[1, 2, [3]].isRectangular // false
+[[1, 2], [3]].isRectangular // false
+[[1, 2], [3, [4, 5]]].isRectangular // false
+::
+
 method::at
 Return the item at strong::index::.
 
@@ -439,25 +453,36 @@ a.performInPlace(\normalizeSum, 3, 6);
 ::
 
 method::rank
-Rank is the number of dimensions in a multidimensional array.
+Returns the number of dimensions of the collection.
+code::rank:: inspects the size of the left-most elements of sub-arrays only,
+i.e. it's assumed that the collection link::#-isRectangular::,
+so subarrays of mismatched sizes may not return an expected result.
+A single value has a rank of code::0::.
+An empty array has a rank of code::1::.
+
 code::
-a = [4,7,6,8];
-a.rank;
-a = [[4,7],[6,8]];
-a.rank;
+0.rank // 0
+[].rank // 1
+[4, 7, 6, 8].rank // 1
+[[4, 7], [6, 8]].rank // 2
+[1 ,2, [3, 4]].rank // 1, this array is not rectangular and returns a meaningless value.
 ::
 
 method::shape
-For a multidimensional array, returns the number of elements along each dimension.
+Returns an array describing the dimension of each nested array.
+Requires link::#-isRectangular:: as a precondition.
+
 code::
-a = [4,7,6,8];
-a.shape;
-a = [[4,7],[6,8]];
-a.shape;
+[4, 7, 6, 8].shape // [4]
+[[4, 7], [6, 8]].shape // [2, 2]
+[[[4, 7]], [[6, 8]]].shape // [2, 1, 2]
+[1, 2, [3, 4]].shape // [3], this array is not rectangular and returns a meaningless value.
 ::
 
 method::reshape
-For a multidimensional array, rearranges the data using the desired number of elements along each dimension. The data may be extended using wrapExtend if needed.
+For a multidimensional array, rearranges the data using the desired number of elements along each dimension.
+The data may be extended using link::Classes/Array#-wrapExtend:: if needed.
+This will always return a rectangular array, see link::#-isRectangular::.
 code::
 a = [4,7,6,8];
 a.reshape(2,2);
--- a/SCClassLibrary/Common/Collections/ArrayedCollection.sc
+++ b/SCClassLibrary/Common/Collections/ArrayedCollection.sc
@@ -321,15 +321,17 @@ ArrayedCollection : SequenceableCollection {
 	}
 
 	// concepts borrowed from J programming language
+	isRectangular {
+	    _ArrayIsRectangular
+		^this.primitiveFailed
+	}
 	rank {
-		// rank is the number of dimensions in a multidimensional array.
-		// see also Object-rank
-		// this assumes every element has the same rank
-		^1 + this.first.rank
+        // precondition: isRectangular
+        ^1 + this.first.rank
 	}
 	shape {
-		// this assumes every element has the same shape
-		^[this.size] ++ this[0].shape
+        // precondition: isRectangular
+        ^[this.size] ++ this[0].shape
 	}
 	reshape { arg ... shape;
 		var size = shape.product;
@@ -452,11 +454,13 @@ ArrayedCollection : SequenceableCollection {
 }
 
 RawArray : ArrayedCollection {
-	// species { ^this.class }
 	archiveAsCompileString { ^true }
 	archiveAsObject { ^true }
 
 	rate { ^\scalar }
+	isRectangular { ^true }
+	shape { ^[this.size] }
+	rank { ^1 }
 
 	readFromStream { |stream, method|
 		if(method.notNil) {
--- a/HelpSource/Classes/Object.schelp
+++ b/HelpSource/Classes/Object.schelp
@@ -112,6 +112,17 @@ method::size
 
 Different classes respond to this message differently. Object always returns 0.
 
+method::rank
+Returns the number of dimensions the value has.
+A single value (scalar) has rank of zero.
+See link::Classes/ArrayedCollection#-rank:: for cases where this is useful.
+
+method::shape
+For a multidimensional array,
+returns an array of the number of elements along each consecutive dimension,
+see link::Classes/ArrayedCollection#-shape::.
+For a scalar value (most objects) returns nil.
+
 subsection::Copying
 
 method::copy
--- a/lang/LangPrimSource/PyrArrayPrimitives.cpp
+++ b/lang/LangPrimSource/PyrArrayPrimitives.cpp
@@ -26,12 +26,13 @@ Primitives for Arrays.
 #include "GC.h"
 #include "PyrKernel.h"
 #include "PyrPrimitive.h"
+#include "SCBase.h"
 #include "SC_InlineBinaryOp.h"
 #include "SC_Constants.h"
 #include "SC_Levenshtein.h"
 
 #include <cstring>
-#include <algorithm>
+#include <vector>
 
 // Primitives that work with Arrays. Most of these are used in ArrayedCollection and Array.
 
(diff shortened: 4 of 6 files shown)


mtmccrea:
This is a useful addition.

How does it behave with depths greater than 2? 
```supercollider
[[1,2], [3, [5,6]]]; // still regular?
```

>Implementing this is sclang would be significantly slower.

Did you happen to try it?

Regarding the naming, `Matrix` in the MathLib quark has a definition of `isRegular` related to whether the determinant is zero, whereas `isSquare` is used to ask about the shape. The wikipedia page you linked to mentions "rectangular" as specifying sub-arrays of equal length. Though not sure how well `isRectangular` translates to higher dimensional shapes. Just thinking out loud...

JordanHendersonMusic:
> How does it behave with depths greater than 2?
> 
> ```supercollider
> [[1,2], [3, [5,6]]]; // still regular?
> ```

This returns false, as expected. 

The only issues i have found is if you do this...
```
[[1,2], [3, List()]].isRegular; 
```

Which I have made throw an wrong type error and postl `Subarray was not an instance of ArrayedCollection`.
The issue is that List (and other non-primitive collections) don't respond to `obj->size` as expected since this refers to slots and items, and List has an array as its first slot.

I could add an implementation for SequenceableCollection in sc, rather than as a primitive, but it would be significantly slower.

I agree with the naming, but haven't been able to think of a better one.

cdbzb:
> not sure how well `isRectangular` translates to higher dimensional shapes

well `rectangular cuboid` and `hyperrectangle` still have "rectangle" in them - I think 'isRectangular' would be fine here.

smoge:
I think rectangular is a good definition, that's how I saw this regarding APL arrays. 

> . What is an Array?
> An array is a function from a set of indices to numbers, characters, … A rank-n array is one whose function f applies to n-tuples of non-negative integers. A rank-n array is rectangular if there exist non-negative integer maxima  s = (s0, s1, …, sn-1) such that  f(i0, i1, …, in-1)  is defined (has a value) for all integer ij such that (0≤ij)^(ij<sj).  s is called the shape of the array. Etc.
> This definition accommodates: APL/J rectangular arrays, J sparse arrays
> 

Alternativ: "sparse" (from J language)

https://www.jsoftware.com/papers/array.htm

mtmccrea:
I think we still need a precise, complete definition. So far this is the closest is

> If you can calculate the number of elements in an array by multiplying its length in each dimension then it is regular

Is that the full definition? The help doc says
> Returns true if all nested sub arrays are the same size.

But in the example I gave, `[[1,2], [3, [5,6]]]`, all sub-arrays are the same size.

And the other example I gave, 10 x 2 x numSamples, is rectangular by the first definition, but nested sub-arrays are not the same size. 

I could see a definition where, for example, all sub-arrays of the last dimension are the same size. But I don't know if that's how it's actually implemented. The definition should come from the implementation.

This isn't fully clear
> or if you can write the dimensions in a 1 dimensional list.

How we write the dimensions is a bit ambiguous, so this isn't necessarily a helpful description. For example, how do you write the dimensions of this? `[[1,2], [3, [5,6]]]`

More generally, Arrays hold any objects, not just numbers, so it's worth considering the broader use cases when writing definitions and examples.

mtmccrea:
It would be great also if the docs for `rank` and `shape` mentioned the assumptions about regularity, with examples that show its behavior in degenerate cases. Every time I use these I have to test them to remind myself how they behave 😆

JordanHendersonMusic:
Ah your right, that isn't a good definition in the help docs.

> The definition should come from the implementation.

It does a left-first depth-first search recording the size of each nested array, and a 'terminal' when it finds an element that isn't an array, at which point it stops recording the sizes. It then visits every element, ensuring it matches this specification. Additionally, if it encounters an element that is a sequenceable collection, but not an array (e.g. List), it will post a warning, and return nil. It could return false instead, or could consider List to be a terminal, and not considered it's size - I think that would be odd.

I don't really know how to turn that nonsense into a definition.

I think the number elements equaling the product of the shape is quite a useful description because it aligns with the use case of audio files.

I'll edit rank and shape too as the precondition is just in a comment right now.

smoge:
> > or if you can write the dimensions in a 1 dimensional list.
> 
> How we write the dimensions is a bit ambiguous, so this isn't necessarily a helpful description. For example, how do you write the dimensions of this? `[[1,2], [3, [5,6]]]`

For completeness and reference:

This video shows how it is done in APL: https://www.youtube.com/watch?v=hzPd3umu78g
image

JordanHendersonMusic:
How about...

>Returns true if all nested sub arrays are the same size and all elements have the same depth. This is a requirement of for many nested array, notably multichannel audio files.

...?


FYI, tensorflow says:
> The base tf.Tensor class requires tensors to be "rectangular"---that is, along each axis, every element is the same size. 

Which I think is what we are talking about here.

JordanHendersonMusic:
Another point, Object.shape returns nil, but Object.rank returns 0. 

Object.shape should really return an empty array to be consistent. That could potentially be breaking though.

mtmccrea:
Thanks for the doc updates, improvements all around :)

Interesting idea to test for rectangular-ity in shape/rank. I see the default is no test, so the change isn't breaking, which is good. Others may have thoughts as well...

Thanks!

JordanHendersonMusic:
I've just made two new commits which are breaking changes, which definitely need input.

The first makes SequenceableCollection report its shape and rank correctly. This is a breaking change, but it returned `nil` and `0` before, so it was useless. While not all SequenceableCollections can be reshaped, they do all have a shape.

The other breaking change is to make Object#-shape return `[]` rather than `nil`, rank should always be the size of the shape array. 

Both changes are breaking, but I *believe* they will be minimally breaking, but I'd appreciate any opinions on that!

JordanHendersonMusic:
So I think this should be it now.

A little update. I tried to implement this for SequenceableCollection, adding definitions for shape and rank (breaking changes, but quite small), while isRectangle (using a recursive function in sclang), rank and shape will all work okay allowing you to mix all types of SequenceableCollection, it gets very confusing when you call reshape as many objects in SequenceableCollection cannot be reshaped. It could be possible to convert the whole thing to an array first, then just return an array but that would add conversions.

Sidenote, I think SequenceableCollection and ArrayedCollection might have been incorrect naming choices and a distinction could have been made between MultidimensionCollections and FlatCollections instead.

Anyway, as it stands, this commit now works for array quite nicely (and isn't a breaking change), but because of the existing class hierarchy, I think it best to leave it as is and not fix the SequenceCollection shape (which always returns `nil`) and rank (always returns `0`).

JordanHendersonMusic:
Thanks for you help @mtmccrea, let me know if there is anything else needed.

SCClassLibrary/Common/Collections/ArrayedCollection.sc
```diff
@@ -321,14 +321,21 @@ ArrayedCollection : SequenceableCollection {
...
-		// rank is the number of dimensions in a multidimensional array.
-		// see also Object-rank
-		// this assumes every element has the same rank
+	isRectangular {
+	    // all nested arrays are the same size and elements are at the same depth.
+	    _ArrayIsRectangular
+		^this.primitiveFailed
+	}
+	rank {|test=false|
+	    if(test and: this.isRectangular.not){
```
telephon, line 330: You need braces `if(test and: { this.isRectangular.not }){` for this to optimize.

telephon (changes requested):
generally I would say: if the test (being a primitive) is really fast, perhaps the `test` argument can be dropped. I am a bit weary about adding this argument. It takes away possible (but completely hypothetical) ways to augment the method with a first argument. 

Maybe you can run a few statistics and we see how fast it is?

SCClassLibrary/Common/Collections/ArrayedCollection.sc
JordanHendersonMusic: The issues is that it has to visit every element O(n). 

```supercollider

// 1
a = [];
bench { a.isRectangular }; // time to run: 1.432999994222e-06 seconds.

// 2
a = (0..1000);
bench { a.isRectangular }; // time to run: 5.7009999920865e-06 seconds.

// 3
a = (0..(1000*1000));
bench { a.isRectangular }; // time to run: 0.0024116329999941 seconds.

// 4
a = 0!1000!1000;
bench { a.isRectangular }; // time to run: 0.0026010189999965 seconds.

// 5
a = 0!1000!1000!1000; // time to run: 2.506840667 seconds.
bench { a.isRectangular };
```
You can see that 3 and 4 are close in performance as they have the same number of elements.

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -439,25 +463,40 @@ a.performInPlace(\normalizeSum, 3, 6);
...
 
+
+
 method::shape
-For a multidimensional array, returns the number of elements along each dimension.
+Requires link::#-isRectangular:: as a precondition.
+For a multidimensional array, returns an array of the number of elements along each consecutive dimension.
+If the array is not rectangular the wrong result will be returned, as this only considers nested arrays at index zero.
+
+Optionally, shape takes an boolean as an argument indicating if it should test the array is rectangular.
```
mtmccrea, line 488: See my suggested structuring of `rank` above

mtmccrea:
> I am a bit weary about adding this argument. It takes away possible (but completely hypothetical) ways to augment the method with a first argument.

Thanks @telephon for your thoughts on this, it also drew my attention. Because the assumption of rectangularity is noted in the docs, it could just as well be up to the user to test the collection themselves before calling `isRectangular`. Judging from the nuance of this method and it's relation to other types of collections, it's not unreasonable to expect that an argument may be added in the future that changes/extends functionality, and which `test` doesn't really do and would stand 'in the way'.

mtmccrea:
FYI, I'm not the best person to review the cpp, so it'd be great if someone could look over that.

SCClassLibrary/Common/Collections/ArrayedCollection.sc
JordanHendersonMusic: I've remove the test

JordanHendersonMusic (commented):
Remove the test, improved documentation.

HelpSource/Classes/ArrayedCollection.schelp
JordanHendersonMusic: Okay I think that is better

smoge:
> Maybe you can run a few statistics and we see how fast it is?

Jordan's work, since it reminds me a bit of Array programming, and will be used for large arrays (sound files) made me wonder if the current implementation of sclang would allow using C++11 (and later) features to process Array types in parallel (possibly in a style reminiscent of APL or not, that is, as pairs of one-=dimentional vectors representing vales and shape). 

C++ std:thread allows this quite straightforwardly:


` ` ` 

#include <cmath> 
#include <iostream>
#include <thread>
#include <vector>

// Structure to represent an APL-style array
struct APLArray {
  std::vector<int> values; // Flat storage of array elements
  std::vector<int> shape;  // Dimensions of the array

  // Constructor
  APLArray(const std::vector<int> &values, const std::vector<int> &shape)
      : values(values), shape(shape) {}

  // Function to print the array in 3D format
  void print3D() const {
    if (shape.size() != 3) {
      std::cout << "Array is not 3D." << std::endl;
      return;
    }
    for (int i = 0; i < shape[0]; ++i) {
      std::cout << "Slice " << i + 1 << ":" << std::endl;
      for (int j = 0; j < shape[1]; ++j) {
        for (int k = 0; k < shape[2]; ++k) {
          std::cout << values[i * shape[1] * shape[2] + j * shape[2] + k]
                    << " ";
        }
        std::cout << std::endl;
      }
      std::cout << std::endl;
    }
  }
};

// Function to process a segment of the APLArray
void processSegment(APLArray &array, size_t start, size_t end) {
  for (size_t i = start; i < end; ++i) {
    array.values[i] = std::pow(array.values[i],
                               2); // Example operation: squaring each element
  }
}

// Function to parallel process the APLArray
void parallelProcessAPLArray(APLArray &array) {
  size_t numThreads = std::thread::hardware_concurrency();
  std::vector<std::thread> threads;
  size_t totalElements = array.values.size();
  size_t elementsPerThread = totalElements / numThreads;

  for (size_t i = 0; i < numThreads; ++i) {
    size_t start = i * elementsPerThread;
    size_t end = (i + 1) * elementsPerThread;
    if (i == numThreads - 1) {
      end = totalElements; // Make sure the last segment includes the end of the
                           // array
    }
    threads.emplace_back(processSegment, std::ref(array), start, end);
  }

  // Wait for all threads to complete
  for (auto &t : threads) {
    t.join();
  }
}

int main() {
  // Example: a 2x2x3 array
  APLArray myAPLArray({1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12}, {2, 2, 3});

  std::cout << "Original array in 3D format:" << std::endl;
  myAPLArray.print3D();

  parallelProcessAPLArray(myAPLArray);

  std::cout << "Processed array in 3D format:" << std::endl;
  myAPLArray.print3D();

  return 0;
}

// OUTPUT
// 
// Original array in 3D format:
// Slice 1:
// 1 2 3 
// 4 5 6 

// Slice 2:
// 7 8 9 
// 10 11 12 

// Processed array in 3D format:
// Slice 1:
// 1 4 9 
// 16 25 36 

// Slice 2:
// 49 64 81 
// 100 121 144 
```
`

**Comment by mtmccrea** [comment]
This would be an interesting topic, @smoge, and could draw a large discussion. So probably best not to flood this PR. Maybe over on discourse?

**Comment by smoge** [comment]
> This would be an interesting topic, @smoge, and could draw a large discussion. So probably best not to flood this PR. Maybe over on discourse?

Sure, just to register the connection to this idea. Bring the discussion to the best place.

**Comment by mtmccrea** [comment]
But this does make me think that, wrt this PR, there are subclasses of `ArrayedCollection` which wouldn't need to have their full depth searched because the element types are constrained, like `RawArray`s. I think (?) `RawArray`s are constrained to 1 dimension, but could themselves be stored in an array. 

It's a bit of an edge case (though writing soundfiles was a motivation for this, and `FloatArray`s are a smart way to deal explicitly with large audio-like `Signal` and `Wavetable`), so I wouldn't ask for a change in this PR unless @JordanHendersonMusic was interested in handling a more granular implementation.

**Comment by smoge** [comment]
> But this does make me think that, wrt this PR, there are subclasses of `ArrayedCollection` which wouldn't need to have their full depth searched because the element types are constrained, like `RawArray`s. I think (?) `RawArray`s are constrained to 1 dimension, but could themselves be stored in an array.
> 
> It's a bit of an edge case (though writing soundfiles was a motivation for this, and `FloatArray`s are a smart way to deal explicitly with large audio-like `Signal` and `Wavetable`), so I wouldn't ask for a change in this PR unless @JordanHendersonMusic was interested in handling a more granular implementation.

One of the reasons APL restricts some Arrays to rectangular forms is because it represents them as two one-dimensional vectors. While more complex structures, like trees, need three vectors to be internally represented. 

But in sclang, it won't improve performance unless we change the language a little bit. I'm not sure about the exact use cases you are considering here.

HSU

**Comment by JordanHendersonMusic** [comment]
> full depth searched because the element types are constrained

Actually, nothing needs to be searched these cases, as they cannot be nested as the shape is just `[this.size]`, the rank `1`, and `isRectangular` is always true. I've added an optimisation for this.

**Comment by JordanHendersonMusic** [comment]
Just added a c++ optimisation for raw arrays, literally two lines.
```
[
   FloatArray[1, 2],
   FloatArray[3, 4]
].isRectangular
// -> true
[
   FloatArray[1, 2],
   FloatArray[3, 4, 5]
].isRectangular
// -> false
```

**Review by telephon — CHANGES_REQUESTED** [review]
As far as I can see, one last thing to do for this is unit tests. Can you do them? This would be great.

**Comment by JordanHendersonMusic** [comment]
I've added the tests now.

**Comment by jamshark70** [comment]
> This sort of raises the question of why list doesn't support shape and rank? If it is a collection, then it probably should for completeness since it is sequenceable. I wonder if there is any history here? @jamshark70, do you happen to know anything about this (apologies for the summons)?

I'm late to the party... I don't recall specific history, and I suspect that it's just a casual omission.

For another example of an inconsistency wrt List (this isn't relevant to the current PR, just to illustrate that the interfaces aren't consistent everywhere):

```
(
f = { |coll|
	var result = coll.collect(_ * 2);
	"%: %\n".postf(result.class, result);
};

f.([1, 2, 3]);

f.(List[1, 2, 3]);
)

Array: [2, 4, 6]
Array: [2, 4, 6]  // but this one was a List :-\
```

`List.new.species` is `Array` -- where, bizarrely, `species` inherits from Collection(?) -- since Collection is basically an abstract class, I'd have expected `species { ^this.subclassResponsibility }` but... huh... anyway. There are definitely some aspects of the Collection hierarchy that appear not to have been fully checked.

(Some omissions may date back to SC2. Other omissions might be where somebody added something to Array but forgot that Array is not the only type of SequenceableCollection.)

telephon:
> `List.new.species` is `Array` -- where, bizarrely, `species` inherits from Collection(?) -- since Collection is basically an abstract class, I'd have expected `species { ^this.subclassResponsibility }` but... huh... anyway. There are definitely some aspects of the Collection hierarchy that appear not to have been fully checked.

hey james, I think I remember how this was (long ago). Originally, the attempt was that for efficiency reasons many primitives returned their closest species. For example, a list would return an array. This would be more efficient when you have a long chain of primitives, I think. This wasn't really understood well by us who inhereited the project and we started finding that it is annoying that operations change the base class, so `species` was added to a few more classes to return the same one.

So List is still the way it was originally, I think.

telephon (approved):
looks good to me (only reading, currently I can't test)

JordanHendersonMusic:
All CI is now passing, anyone want to review and merge?

mtmccrea:
I've built and tested, it appears to be working as expected. Thanks @JordanHendersonMusic and all!

Issue #6224, opened 2024-02-26
build error on macOS 14 (Sonoma) – possibly a boost problem? (BOOST_MPL_AUX_STATIC_CAST)


telephon:
For a while, I have been trying to build supercollider on macOS 14 (Sonoma), but I consistently get a boost error.

```
path/supercollider/external_libraries/boost/boost/mpl/aux_/integral_wrapper.hpp:73:31: warning: integer value -1 is outside the valid range of values [0, 3] for this enumeration type [-Wenum-constexpr-conversion]
path/supercollider/external_libraries/boost/boost/mpl/aux_/static_cast.hpp:24:47: note: expanded from macro 'BOOST_MPL_AUX_STATIC_CAST'
#   define BOOST_MPL_AUX_STATIC_CAST(T, expr) static_cast<T>(expr)
                                              ^
15 warnings and 1 error generated.
make[2]: *** [external_libraries/CMakeFiles/boost_thread_lib.dir/boost/libs/thread/src/pthread/thread.cpp.o] Error 1
make[1]: *** [external_libraries/CMakeFiles/boost_thread_lib.dir/all] Error 2
make: *** [all] Error 2
```

This was done in a fresh folder with the latest develop branch, using:

```
cmake -DCMAKE_PREFIX_PATH=`brew --prefix qt5`  ..
cmake --build . --target install --config RelWithDebInfo
```
Maybe the bundled boost needs to be updated? (see e.g. https://github.com/Kitware/KWStyle/pull/111)

smoge:
Julian, you can try to set cmake to build SC with the c++14 flag. Or you can try to build with " system boost" , which will be more recent and will not have this deprecated function (since C++17). 

May be time to the project adopt C++17, I think.

telephon:
So I now did my build (in a new folder) with 
```
cmake --std=c++14 -DCMAKE_PREFIX_PATH=`brew --prefix qt5`  ..
```
it worked, but didn't change the error on `cmake --build . --target install --config RelWithDebInfo`.

> Or you can try to build with " system boost", which will be more recent and will not have this deprecated function (since C++17).

How do I do this?

smoge:
> How do I do this?

-DSYSTEM_BOOST=ON

You can also see all options with `cmake-gui ..` if you have this tool installed.

telephon:
Thanks, I had completely forgotten.

This is now how I tried:

```
[...]

PR #6227, opened 2024-03-02
Implement `doesNotUnderstandWithKeys`

### LINKED ISSUE #6064: Pass nonmatching keyword args to the method call - doesNotUnderstand - Edit in PyrMessage.cpp (help)
Looking for help solving an issue as it involves the lang code and I am struggling to understand it.

## Motivation

For some class
```
Test {
	doesNotUnderstand{|...a|
		a.postln
	}
}
```
```
Test().foo(1, 2, bar: 4)
```
prints
```
WARNING: keyword arg 'bar' not found in call to ErTest:doesNotUnderstand
[ foo, 1, 2 ]
```

It would allow the creation of transparent wrapper classes if this would instead return
```
[foo, 1, 2, bar, 4]
```
meaning you could call code before and after all methods.

I've written a little `Buffer` wrapper that tests to see if it has loaded on the server (and optional waits) before every method call, meaning the user never has to call s.sync again, it is completely transparent and works every where I have tested... except with named args method calls. This would be a huge simplification for the user.

## Plan for Implementation

I think the problem lies here: 

https://github.com/supercollider/supercollider/blob/ef627ce2c564fe323125234e4374c9c4b0fc7f1d/lang/LangSource/PyrMessage.cpp#L840

and that the issue is simply a matter of copying all the arguments over. 
This section...
https://github.com/supercollider/supercollider/blob/ef627ce2c564fe323125234e4374c9c4b0fc7f1d/lang/LangSource/PyrMessage.cpp#L980
...copies only the args that match the current argument names. 
Is it possible to copy the rest that do not match over, but placed at the end of the list so it doesn't interfere with existing code? 
I cannot seem to quite figure it out, is there any one who still understands this part of supercollider around?

There is one place where this will not work though, which is when the method arg name matches in doesNotUnderstand, e.g.,
```
Test {
	doesNotUnderstand{|...a|
		a.postln
	}
}
```
```
Test().foo(1, 2, a: 4)
```
prints
```
4
```

The other option would be to create a doesNotUnderstandWithKeys in the class library. There is a cpp function `doesNotUnderstandWithKeys` but no corresponding method in `Object` - this is probably a bigger change though.

I'm also gonna post this over on scsynth.

Thanks,
Jordan

JordanHendersonMusic:
## Purpose and Motivation

Information is needlessly lost when calling doesNotUnderstand when there are keywords args.

Now, when a selector isn't understood and there are keysword args in the call, the following method is called, here is Objects implementation.
```supercollider
doesNotUnderstandWithKeys { |selector, withoutKeys, withKeys|
    DoesNotUnderstandWithKeysError(this, selector, [withoutKeys, withKeys]).throw;
}
```

Fixes #6064 

Here is some behaviour as it relates to identity dictionary.

Old:
```supercollider
e = (
   \speak : {|self, words, reason| postf("I said % because %\n", words, reason) }
);

e.speak("meow", reason: "a big dog barked")
// WARNING: keyword arg 'reason' not found in call to IdentityDictionary:doesNotUnderstand
// I said meow because nil
```

New.

```supercollider
e = (
   \speak : {|self, words, reason| postf("I said % because %\n", words, reason) }
);

e.speak("meow", reason: "a big dog barked")
// -> I said meow because a big dog barked
```
I admit, `nil` is probably a better reason to meow.

### More examples
```supercollider
e = (
   \speak : {|self, words, reason| postf("I said % because %\n", words, reason) }
);

e.speak("meow", "ahahah", reason: "a big dog barked")
// ERROR: Key reason already added
```

```supercollider
e = (
   \speak : {|self, words, link, reason| postf("I said % % %\n", words, link, reason) }
);

e.speak("meow", link: "because", reason: "a big dog barked")
// -> I said meow because a big dog barked

e.speak("meow", "because", reason: "a big dog barked")
// -> I said meow because a big dog barked

e.speak("meow", reason: "a big dog barked")
// -> I said meow nil a big dog barked

e.speak("meow", asdf: 234, reason: "a big dog barked"); 
// throws
//  Key 'asdf' not found in function call

e.speak(reason: "a big dog barked");
// -> I said nil because a big dog barked
```

Also provides some helper methods in `Object` to make redirecting easier. 
These are used to call functions, class methods and methods respectively.

```supercollider
// private impl prDoesNotUnderstandWithKeysHelper { |callingFunc, expectedArgNames, withoutArray, withEvent| ...}
doesNotUnderstandWithKeysFunctionHelper { |func, withoutArray, withEvent| ...}
doesNotUnderstandWithKeysClassMethodHelper { |class, methodName, withoutArray, withEvent| ...}
doesNotUnderstandWithKeysMethodHelper { |class, this_object, methodName, withoutArray, withEvent| ...}
doesNotUnderstandWithKeys { |selector, withoutKeys, withKeys|...}
```

## Types of changes

- Documentation
- Bug fix 
- New feature
- Breaking change ... think this unlikely, but if the user depended on an error being thrown it could break something.

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


JordanHendersonMusic:
I've added implementation for `IdentityDictionary`, `Scale` and `Tuning`. Most classes should be able to support this, `Maybe` looks a little more complex though.

JordanHendersonMusic:
All documentation is now updated, and I have not been able to find any issues, this is now ready for a review.

telephon:
This is a very very useful addition, thanks. Currently, for me, the build is broken (due to a #6224), so I can't test it. But I am in favour of adding this feature.

telephon (commented):
In terms of naming and documentaton, this could be improved a little.

`doesNotUnderstandWithKeysClassMethodHelper`
- What does this method actually do?
- Maybe there is a simpler name?

`withoutArray` argument: in the method call of `doesNotUnderstandWithKeys` this is called `withOutKeys`, isn't it?

And `withEvent` is called `withKeys`, right? This makes things harder to understand than necessary.

Also, I would write `without` and not `withOut`, becaus this is just one word.

This method will be very useful for object prototyping and calling pseudo-object code (envirs) with keyword arguments.

JordanHendersonMusic:
Once you have the two types of arguments  (without and with keywords) it is quite hard to do anything with them as there isn't any obvious way of combining them and calling a function. Perhaps keyword arguments were added later leading to an odd interface?

In this recent commit I've added some more methods to FunctionDef to centralise the error checking, ensuring all keywords are accepted, and none of the arguments collide with each other. Having each class the uses `doesNotUnderstandWithKeys` deploy their own error checking would lead to inconsistent behaviour.

There are then new versions of `perform...` and `value...` accepting the mixture of args and kw-args,
and a version of `functionPerformList` that accepts both types of arguments, which is needed for object prototyping.

I admit that having so many ways to call a function is dumb, but this approach seems the most consistent with supercollider's code base.

Here are some examples of how this all works together taken from the commit, which I think, provides a decent interface for anyone wanting to use this method.
```supercollider
// From Scale
*doesNotUnderstandWithKeys {|selector, argsArray, keywordArgsAsPairs|
   ^Scale.class
   .findRespondingMethodFor(\newFromKey)
   .evaluateWithArgsAndKwArgs([selector] ++ argsArray, keywordArgsAsPairs.asEvent, Scale)
}
```

```supercollider
// From IdentityDictionary
doesNotUnderstandWithKeys {|selector, argsArray, keywordArgsAsPairs|
   if(know.not){
      ^this.superPerformList(\doesNotUnderstandWithKeys, selector, argsArray, keywordArgsAsPairs)
   };

   this[selector] !? {|f|
      ^f.functionPerformWithArgsAndKwArgs(\value, [this] ++ argsArray, keywordArgsAsPairs.asEvent)
   };

   this[\forward] !? {|f|
      ^f.functionPerformWithArgsAndKwArgs(\value, [this, selector] ++ argsArray, keywordArgsAsPairs.asEvent)
   };
   ^nil
}
```

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -596,6 +628,16 @@ Method : FunctionDef {
 		^[names, values].flop.flatten
 	}
 
+	isClassMethod {	^ownerClass.isMetaClass	}
+
+	evaluateWithArgsAndKwArgs{|argsArray, keywordArgsEvent, thisObject|
```
telephon, line 633: why `evaluate` and not `perform`?

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,6 +555,32 @@ FunctionDef {
 		^().putPairs(this.keyValuePairsFromArgs)
 	}
 
+	makeNonOverlappingKeywordArgsEvent {|argsArray, keywordArgsEvent, omitArgNameIfFirst|
```
telephon, line 558: This method is a little opaque still. I'll see if I can give you some suggestions later.

telephon (changes requested):
great! Needs a few tweaks, but looks good already.

telephon:
Maybe it helps to take a look at `performWithEnvir`, this seems similar.

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -137,6 +137,33 @@ Object  {
...
+			argsArray: argsArray,
+			keywordArgsEvent: keywordArgsEvent,
+			thisObject: this
+		)
+	}
+	valueWithArgsAndKwArgs{|argsArray, keywordArgsEvent|
+		^this.performWithArgsAndKwArgs(\value, argsArray, keywordArgsEvent)
+	}
+	functionPerformWithArgsAndKwArgs{|selector, argsArray, keywordArgsEvent|
+		if(keywordArgsEvent.isNil or: {keywordArgsEvent.size == 0}){
```
telephon, line 154: `.size == 0` -> `isEmpty`

SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -559,26 +559,33 @@ IdentityDictionary : Dictionary {
...
-    // The Quant class assumes the quant/phase/offset scheduling model.
-    // If you want a different model, you can write a dictionary like so:
-    // (nextTimeOnGrid: { |self, clock| ... calculate absolute beat number here ... },
-    //	parameter: value, parameter: value, etc.)
-    // If you leave out the nextTimeOnGrid function, fallback to quant/phase/offset.
+		this[selector] !? {|f|
+			^f.functionPerformWithArgsAndKwArgs(\value, [this] ++ argsArray, keywordArgsAsPairs.asEvent)
+		};
+
+		// You cannot call a setter with a keyword arg, it does not make sense.
```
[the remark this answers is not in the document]
JordanHendersonMusic: Oh, `doesNotUnderstand` does a look up for setters here, this was just to show the symmetry between the two.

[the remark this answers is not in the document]
telephon: yes maybe better:

"// Unlike in doesNotUnderstand, we don't convert to a setter, this isn't possible with keyword args."

SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -560,12 +560,32 @@ IdentityDictionary : Dictionary {
...
+
+		// If the keyword arg names don't match this will fail.
+		// Note how the selector is passed in here.
+		this[\forward] !? {|f|
+			^f.functionPerformWithArgsAndKwArgs(\value, [this, selector] ++ argsArray, keywordArgsAsPairs.asEvent)
```
telephon, line 577: one more suggestion:

let's add a last `args` argument to `performWithEnvir` in both Object and Function, then the method `functionPerformWithArgsAndKwArgs` should nor be necessary.

What do you think?

JordanHendersonMusic: If I've understood you correctly — as far as I can see, classes have to opt into being called through object prototyping by implementing  `functionPerformList`, otherwise they just return themselves, meaning replacing this with `performWithEnvir` would be a breaking change. 

It would not be possible to construct an argument array from args and kwargs from inside identity dictionary as you'd have to know what the held object did inside `functionPerformList`,
and `functionPerformList` uses an array of args, meaning you can't get the argnames either.

Without breaking object prototyping, I can't see another way (unless I've misunderstood you).

If the interface for prototyping was to change though, one option would be for `doesNotUnderstand*` to check the first argument (second argument if it is 'this') of the object's `value` method (or FunctionDef if function) and if it is called 'self' then pass the dictionary in. This way, `functionPerformList` could be remove entirely, and any class or function can opt into this behaviour. While switch on argument names is perhaps odd, I don't think there is any other use for the word 'self' in supercollider.

SCClassLibrary/Common/Core/Kernel.sc
JordanHendersonMusic: Since this doesn't take a selector, `valueWithArgsAndKeywordArgs` would be the option,
however, this would require that I change all `perform*` methods to actually evaluate the method rather than looking up other methods inside of `Method`, while `Method` isn't the most common class, I was trying to break as little code as possible. 

I don't know that it makes sense to have `Method` be a true 'callable' object as it requires error checking the type of 'this' against the class — otherwise there would be very hard to find bugs.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -26,6 +26,12 @@ Class {
 	}
 	asClass { ^this }
 	isMetaClass { ^this.class === Class }
+	removeMeta {
```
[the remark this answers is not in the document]
JordanHendersonMusic: Just realised I never used this, so I will remove it.

JordanHendersonMusic:
Thanks for your feedback @telephon, I've addressed the smaller issues in this last commit (and resolved the comments), the larger ones I will try and get to tomorrow.

telephon:
>Perhaps keyword arguments were added later leading to an odd interface?
Yes, they were, but I think the interface is optimised for speed. Keyword arguments are slower than normal ones.

> Once you have the two types of arguments (without and with keywords) it is quite hard to do anything with them as there isn't any obvious way of combining them and calling a function. 

It would be easy to do one, this is how I would go about it. Ellipsis arguments would need to be taken care of still.

```

(
f = { |func, args, keywordArgDict|
	var callArgs = func.defaultArgs.copy.extend(func.argNames.size, nil);
	callArgs.overWrite(args);
	func.argNames.do { |name, i|
		var val = keywordArgDict[name];
		if(val.notNil) { callArgs[i] = val }
	};
	func.valueArray(callArgs)
}
)


f.({ |x, y = 0, freq, amp = 0.2| [x, y, freq, amp] }, [9], (freq:500, amp: 0.2))
	
```

SCClassLibrary/Common/Core/Kernel.sc
telephon: You can query the argNames and defaultArgs (prototypeFrame) of any method like you can for a function definition. I don't think you need all this complication.

SCClassLibrary/Common/Core/Kernel.sc
JordanHendersonMusic: I've renamed these in the latest to `makePerformableEnvir` and `makePerformableArray`.

HelpSource/Classes/Object.schelp
```diff
@@ -652,6 +662,41 @@ foo {
 }
 ::
 
+
+method::doesNotUnderstand
+Not usually called in user code.
```
telephon, line 667: also better leave out "Not usually called in user code."

HelpSource/Classes/Object.schelp
```diff
@@ -652,6 +662,41 @@ foo {
...
+
+argument::... args
+The arguments passed.
+
+code::
+4.someThing(2) // ERROR: Message 'someThing' not understood.
+::
+
+method::doesNotUnderstandWithKeys
+Not usually called in user code.
```
telephon, line 682: also better leave out "Not usually called in user code."

HelpSource/Classes/Object.schelp
```diff
@@ -652,6 +662,41 @@ foo {
...
+This is called when an object receives a message it does not understand, but there are keyword arguments.
```
telephon, line 683: `, and there are arguments passed via keywords`

HelpSource/Classes/Object.schelp
```diff
@@ -652,6 +662,41 @@ foo {
...
+Object's implementation will throw an error, but other classes (link::Classes/IdentityDictionary::) might override this.
+
+argument::selector
+The name of the message as a link::Classes/Symbol::.
+
+argument::argsArray
+The arguments passed that did not have keywords attached passed as an link::Classes/Array::
```
telephon, line 690: suggestion:

An link::Classes/Array:: of arguments that have been passed directly and not via keywords.

HelpSource/Classes/Object.schelp
```diff
@@ -652,6 +662,41 @@ foo {
...
+
+argument::keywordArrayPair
+The arguments passed that did have keywords attached.
+These are formatted as an link::Classes/Array:: of key-value pairs link::Reference/Key-Value-Pairs::.
+
```
telephon, line 695: the arguments that have been passed via keywords.

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -185,8 +193,17 @@ Tuning {
...
+		!? { |v| v } //return if valid
+		?? { super.doesNotUnderstand(selector, args) }
+	}
+
+	*doesNotUnderstandWithKeys {|selector, argsArray, keywordArgsAsPairs|
+		^Tuning.performWithEnvir(
+			\newFromKey,
+			Tuning.class.findRespondingMethodFor(\newFromKey)
+			.makePerformableEnvir([selector] ++ argsArray, keywordArgsAsPairs)
+		)
```
telephon, line 206: If `makePerformableEnvir` were implemented as method of Object, this code would look a bit simpler:

```
^Tuning.performWithEnvir(
			\newFromKey,
			Tuning.class.makePerformableEnvir([selector] ++ argsArray, keywordArgsAsPairs)
		)
```

Actually I am not sure I understand why the Method `newFromKey` is used to build a performableEnvir for a different `selector`. Couldn't this be the Method found via the `selector`?

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,6 +549,56 @@ FunctionDef {
...
+
+			out[k] = v;
+			setOfAddedNames.add(k);
+		};
+		^out
+	}
+
+	prMakePerformableArray {|argsArray=([]), keywordArgsEvent=(()), dropFirstN=0|
+		var ev = this.makePerformableEnvir(argsArray, keywordArgsEvent, dropFirstN);
+		var protoDrop = prototypeFrame.drop(dropFirstN);
```
telephon, line 581: you are not using `protoDrop` afterwards.

But if, you could use the suggestion above and get rid of `dropFirstN`

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,6 +549,56 @@ FunctionDef {
 		^().putPairs(this.keyValuePairsFromArgs)
 	}
 
+
+	prMakePerformableEnvir {|argsArray=([]), keywordArgsEvent=(()), dropFirstN=0|
+		var argNamesNoThis = argNames.drop(dropFirstN);
+		var setOfAddedNames = Set();
+		var out = ();
```
telephon, line 556: better `envir = ()` (this helps reading because the method is called `prMakePerformableEnvir`).

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,6 +549,56 @@ FunctionDef {
 		^().putPairs(this.keyValuePairsFromArgs)
 	}
 
+
+	prMakePerformableEnvir {|argsArray=([]), keywordArgsEvent=(()), dropFirstN=0|
```
telephon, line 553: maybe the extra argument `dropFirstN` could be made unnecessary if we would not write 
```
var argNamesNoThis = argNames.drop(dropFirstN);
```
but implement a method for both Method and FunctionDef
```
+ Method {
    argNamesForCall { ^argNames.drop(1) }
}

+ FunctionDef {
    argNamesForCall { ^argNames }
}

Then you just write:

```
var argNamesNoThis = this.argNamesForCall;
```

Comment (telephon) [inline SCClassLibrary/Common/Core/Kernel.sc:553]: maybe the extra argument `dropFirstN` could be made unnecessary if we would not write 
```
var argNamesNoThis = argNames.drop(dropFirstN);
```
but implement a method for both Method and FunctionDef
```
+ Method {
    argNamesForCall { ^argNames.drop(1) }
    defaultArgs { ^prototypeFrame.drop(1) }
}

+ FunctionDef {
    argNamesForCall { ^argNames }
// defaultArgs is already defined for FunctionDef
}
```

Then we just write:

```
var argNamesNoThis = this.argNamesForCall;
```

File: SCClassLibrary/Common/Core/Kernel.sc [hunk 16]
```diff
@@ -596,6 +646,19 @@ Method : FunctionDef {
 		^[names, values].flop.flatten
 	}
 
+	isClassMethod {	^ownerClass.isMetaClass	}
+
+
+	makePerformableEnvir {|argsArray=([]), keywordArgsEvent=(())|
```
telephon, line 652: I'd shift the default arguments to the *last* possible moment (lazy style). They are defined already in `prMakePerformableEnvir`.

telephon (commented):
Here a few more suggestions. It is fun, I hope it is ok that there are so many requests. It is a deeper change, so we need to make it really as good as possible.

SCClassLibrary/Common/Collections/Scale.sc
JordanHendersonMusic: It's `Scale.major(tuning: Tuning.choose)` so `\major` is the selector and you still need to build the arguments needed by `\newFromKey`.... I used scale here because I can't think of the name of a tuning, but it's the same.

I don't think it is necessary to add this to object because it would still have to go through Method/Functiondef to actually get the arguments. Also object is huge.

telephon: I see what you mean. 

Here is another suggestion. When I wrote `performWithEnvir`, I didn't think of passing non-keyword arguments. We could now extend this method to take an array of arguments.

Then the functionality of `makePerformableEnvir` could go in there. Some of the functionality is already there, and duplicating it in an extra method is perhaps not optimal anyway.

Then, in the above case, you could simply call 
```
Tuning.performWithEnvir(\newFromKey, keywordArgsAsPairs.asDict, argsArray)
```

For Function, this would be `valueWithEnvir`, which would take a number of arguments as a second argument.

**BUT/AND**: Of course, then, it may also be a simplification to keep the method `makePerformableArray` that converts the mixed data from array and envir to one single array that can be used to call `valueArray` and `performList` within the `valueWithEnvir` and `performWithEnvir` methods.

**SO:** `makePerformableEnvir`  may be dropped, `valueWithEnvir` and `performWithEnvir` improved by using `makePerformableArray`, and adding an extra array argument. All this is without too much extra code.

JordanHendersonMusic:
I've just pushed a commit that addressed your comments on the code, I think it look much better with your suggestions @telephon. I haven't finished looking at the docs yet, as I'll wait to see what happens with the other conversation before making progress there.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -452,11 +462,28 @@ FunctionDef {
 		_FunctionDefDumpContexts
 		^this.primitiveFailed
 	}
+
+	// These methods are used to indicate that only a subsection of the argName
+	//     and prototypeFrame are used to call the FunctionDef.
```
telephon, line 467: Would it be OK to go back a little bit and minimise the footprint? The commenting of the variable names is OK, but the rest feels a little bit too extensive for the current PR. I'll propose some "reductions", sorry – I know this is a lot of work.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -452,11 +462,28 @@ FunctionDef {
...
+		.drop(this.argNamesForCallStartIndex)
+		.keep(this.argNamesForCallSize)
+	}
+
 	inspectorClass { ^FunctionDefInspector }
 
-	findReferences { arg aSymbol, references;
-		var lits;
-		lits = selectors.asArray;
+	findReferences { |aSymbol, references|
```
telephon, line 485: It would be good to minimise reformatting please for now, not to muddy the water. It is a critical PR and in the end, we will have to be able to know what was changed. 

Otherwise, I also would like to have these cosmetics, but right now it feels like noise.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -467,17 +494,17 @@ FunctionDef {
...
+
 	checkCanArchive { "cannot archive FunctionDefs".warn }
 	archiveAsCompileString { ^true }
 
 	hasPartialApplication {
 		^argNames.size > 0 and: { argNames[0] == \_ }
 	}
 
-	argumentString { arg withDefaultValues=true, withEllipsis=false, asArray=false;
+	argumentString { |withDefaultValues=true, withEllipsis=false, asArray=false|
```
telephon, line 507: that is ok, I think.

telephon (commented):
Many improvments thank you so much, I am sorry to have a few "backward" suggestions …

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -452,11 +462,28 @@ FunctionDef {
 		_FunctionDefDumpContexts
 		^this.primitiveFailed
 	}
+
+	// These methods are used to indicate that only a subsection of the argName
+	//     and prototypeFrame are used to call the FunctionDef.
+	argNamesForCallStartIndex { ^0 }
+	argNamesForCallSize { ^argNames.size }
+
+	argNamesForCall {
```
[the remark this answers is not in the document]
JordanHendersonMusic: Actually.... I think I just found a bug in `performWithEnvir` directly related to why the keep is needed....

```supercollider

f = {|a...b| 
	var i = 0; 
	i.debug('i should be 0');
	a.debug('a should be 1');
	b.debug('b should be []');
};

f.(1);
// i should be 0: 0
// a should be 1: 1
// b should be []: []

f.performWithEnvir(\value, (a: 1));
// i should be 0: 0
// a should be 1: 1
// b should be []: [[], 0]    <<<<<<<<< 

```
In side of prototypeframe is also the default values of the variables, and these should not be passed to the performWithList (or any other perform).

This is another reason why I think `makePerformableEnvir` and `makePerformableArray` should live inside `FunctionDef`, because it is complicated and prone to error. I am also going to add a var args argument to these.


For reference.
```supercollider
performWithEnvir { |selector, envir|
		var argNames, args;
		var method = this.class.findRespondingMethodFor(selector);

		if(method.isNil) { ^this.doesNotUnderstand(selector) };

		argNames = method.argNames.drop(1);
		args = method.prototypeFrame.drop(1);
		argNames.do { |name, i|
			var val = envir[name];
			val !? { args[i] = val };
		};

		^this.performList(selector, args)
	}
```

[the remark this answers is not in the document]
telephon: > There may be other places in the class library that need to be taken care of once the interface with `defaultArgs` is settled.

I've looked around, in most places there are measures to drop the excess elements, but without comment. So having an explicit `defaultArgs` method will definitely be better.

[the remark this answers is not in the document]
JordanHendersonMusic: Ah that's a good idea. Alright I'll do that tomorrow.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -539,92 +566,79 @@ FunctionDef {
 	}
 
 	keyValuePairsFromArgs {
-		var values;
 		if(argNames.isNil) { ^[] };
-		values = this.prototypeFrame.keep(argNames.size);
-		^[argNames, values].flop.flatten
+		^[this.argNamesForCall, this.defaultArgsForCall].lace(this.argNamesForCallSize * 2);
```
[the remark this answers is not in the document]
telephon: Btw. there is a method that returns the number of args of a Function / FunctionDef: `{ |a| }.numArgs` it just isn't implemented on `Method`.

JordanHendersonMusic:
I just found another serious bug related to the way variable arguments are expanded while making `perform/value - With`, this time in `valueArray`.

For some testing function...
```supercollider
f = { |a, b, c, d ...e| 
   var i = 0;
   var j = 1;
	
   a.debug(\a);
   b.debug(\b);
   c.debug(\c);
   d.debug(\d);
   e.debug(\e);
};

```
This works as expected.

```supercollider
f.valueArray(\a, \b, \c, \d, [\e, \f, \g])

a: a
b: b
c: c
d: d
e: [e, f, g]
```

This does not.
```supercollider
a = [\a, \b, \c, \d, [\e, \f, \g]]
f.valueArray(a)

a: a
b: b
c: c
d: d
e: [[e, f, g]] <<<< double wrapped.
```

valueArray is supposed to work by taking the last item on the stack and expanding it, which it does!
However, when called with an actually array, the total number of items on the stack is not the array's SIZE, but its CAPACITY.
The solution is to alway call `valueArray` like this...

```supercollider
a = [\a, \b, \c, \d, [\e, \f, \g]]
f.valueArray(*a)  <<< teeny little star fixes problem

a: a
b: b
c: c
d: d
e: [e, f, g] <<<< correct
```
... making it quite brittle.

This is quite a big change as many (but not all) use cases get this wrong, a quick grep shows about 80 calls to this method.

This will also be the case for `valueArrayEnvir`.

I will make a separate issue on this.

telephon:
I don't think it is an issue, but see my comment there.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,12 +563,73 @@ FunctionDef {
 		^().putPairs(this.keyValuePairsFromArgs)
 	}
 
+
+	makePerformableArray {|argumentsArray=([]), keywordArgumentEnvir=(()), variableArgumentsArray=([])|
```
JordanHendersonMusic, line 567: I've put it here in FunctionDef because otherwise you have to duplicate it about in both object and function. Since it is somewhat fiddly, and I might try and turn it into a primitive, it is better to have it in one place.

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -72,12 +72,79 @@ Function : AbstractFunction {
 		// evaluate a function, using arguments from the supplied environment
 		// slightly faster than valueEnvir and does not replace the currentEnvironment
 		^this.valueArray(prototypeFrame)
+
+	valueWith {|argumentsArray=([]), keywordArgumentEnvir=(()), variableArgumentsArray=([])|
```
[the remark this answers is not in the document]
JordanHendersonMusic: > I figure that doesNotUnderstandWithKeys may just produce an array of (un-keyworded) argument values that is longer than the number of argument names. In this case, the keywordArgumentEnvir will be empty anyhow.

It isn't possible to have doesNotUnderstandWithKeys be called without any keyword args... unless the user does so themselves....

I will try and remove it and see what happens.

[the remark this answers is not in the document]
telephon: I suppose it may happen that someone calls a message with keyword arguments but also supplies more positional arguments than argument names exist. Then, still `doesNotUnderstandWithKeys` would be called?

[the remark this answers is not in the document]
JordanHendersonMusic: I've removed it and it all seems to work.
Yes it would be called then, but so long as performWith or makePerformableArray is called, it will do the right thing.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -564,24 +564,27 @@ FunctionDef {
...
+		// Return no argument in the varArgs position if empty.
+		// This means the default provided by the prototypeFrame ( '[]' ) is incorrect, so drop it.
+		if (this.varArgs){ arguments = arguments.drop(-1) };
 
 		// put non-keyword-arguments in first
 		argumentsArray.do{|a, i|
-			addedNames.add(argNamesCall[i]); // cannot collide here.
-			arguments[i] = a;
+			argNamesCall[i] !? {|n| addedNames.add(n) }; // cannot collide here. Nil if a varArg.
+			if (i >= arguments.size) {
```
telephon, line 583: simpler would be: 
```
[1, 2, 3].overWrite([5, 6, 7, 8])
[1, 2, 3].overWrite([5, 6])
```

SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -560,12 +560,39 @@ IdentityDictionary : Dictionary {
...
+			^this.superPerformList(\doesNotUnderstandWithKeys, selector, argsArray, keywordArgsAsPairs )
+		};
+
+		this[selector] !? {|f|
+			^f.functionPerformList(
+                \value,
+                 f.def.makePerformableArray([this] ++ argsArray, keywordArgsAsPairs.asEvent)
+             )
+		};
+
```
telephon, line 574: ... and this:

```
^f.valueWith([this] ++ argsArray, keywordArgsAsPairs.asEvent)
```

SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -560,12 +560,39 @@ IdentityDictionary : Dictionary {
...
+		// Unlike in doesNotUnderstand, we don't convert to a setter as this isn't
+		// possible with keyword args.
+
+		// If the keyword arg names don't match this will fail.
+		// Note how the selector is passed in here.
+		this[\forward] !? {|f|
+			^f.functionPerformList(
```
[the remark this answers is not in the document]
JordanHendersonMusic: No, that would change how object prototyping calls methods and potentially break stuff.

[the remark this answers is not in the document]
telephon: ah sorry, sure. I meant `^func.performWith(\value, [this] ++ argsArray, keywordArgsAsPairs.asEvent)`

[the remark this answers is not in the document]
JordanHendersonMusic: No that still won't work because it needs to go through `functionPerformList` which is how classes currently opt into object prototyping.

JordanHendersonMusic:
Alright, I think I've gotten everything you mentioned, just the documentation to review now.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -613,7 +613,7 @@ Method : FunctionDef {
 	var <ownerClass, <name, <primitiveName;
 	var <filenameSymbol, <charPos;
 
-	defaultArguments { ^this.prototypeFrame.drop(1).keep(argNames.size - 1).copy }
+	defaultArgs { ^this.prototypeFrame.drop(1).keep(argNames.size - 1).copy }
```
telephon, line 602: no need for an extra copy, because `drop(1)` is guaranteed to return a different array.

SCClassLibrary/Common/Collections/Dictionary.sc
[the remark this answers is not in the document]
telephon: I thought that it may be inefficient to calculate the array when none is needed. So a method like `functionPerformWith` would be logical, which would simply return `this` for objects and call `valueWith` for functions.

But, I suppose it is a rare case anyhow that someone passes arguments when simply accessing an object in a prototype.

Like: 
```
p = (finch: [1, 2, 3]);
p.finch; // yes
p.finch(freq: 900); // rather not
```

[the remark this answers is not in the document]
nuss: I think there's a nicely idiomatic way to handle this. It's not unlikely someone wants to `p.finch` to sometjing else, or?
```
p.finch_((freq: 900));
```
equivant to
```
p.finch = (freq: 900);
```
But that syntax doesn't cause any conflict, does it?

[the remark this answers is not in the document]
JordanHendersonMusic: That's not a keyword argument, and should already work? Away from computer, but I'll test later.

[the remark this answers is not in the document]
JordanHendersonMusic: > p.finch(freq: 900); 

Is this to demonstrate a case where the user just wants to access the value?

What you've written throws an error because the keyword does not exist as a selector on Array (I think).

I don't think it ever makes sense to use a keyword when you want to access something.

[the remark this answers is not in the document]
telephon: I imagine the case where the message `finch` would be sent to different prototype objects (dictionaries), and some of them have a function in finch to generate a new object, taking `freq` as an argument, and others have a fixed object in them, ignoring the argument.

In the case of normal objects, this just goes through without complaint:

```supercollider
p = Point(0, 3);
p.x(7)
p.x(where:7)
```

But my comment was mainly about optimization: if you call `(x:600).x(where:7)` and there is an object (here 600) is fixed, there is no need to go through the whole `makePerformableArray`.

But this is just something to keep in mind, I suppose.

SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -438,9 +438,9 @@ IdentityDictionary : Dictionary {
 		index = this.scanFor(key);
 		array.put(index+1, value);
 		if ( array.at(index).isNil, {
-			array.put(index, key);
-			size = size + 1;
-			if (array.size < (size * 4), { this.grow });
+		array.put(index, key);
```
telephon, line 441: maybe you can undo this reformatting (it is so annoying that autoindent doesn't leave comments alone)

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -19,8 +19,12 @@ Scale {
 	}
 
 	*doesNotUnderstand { |selector, args|
-		var scale = this.newFromKey(selector, args).deepCopy;
-		^scale ?? { super.doesNotUnderstand(selector, args) };
+		^this.newFromKey(selector, args).deepCopy
```
telephon, line 22: ok, this is cosmetic, but since you changed it, here a suggestion for readability:

```supercollider
    var scale = this.newFromKey(selector, args);
    if(scale.notNil) { ^scale.deepCopy };
    ^super.doesNotUnderstand(selector, args)
```

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -185,8 +189,16 @@ Tuning {
 	}
 
 	*doesNotUnderstand { |selector, args|
-		var tuning = this.newFromKey(selector, args).deepCopy;
-		^tuning ?? { super.doesNotUnderstand(selector, args) }
+		^this.newFromKey(selector, args).deepCopy
```
telephon, line 192:
```supercollider
    var tuning = this.newFromKey(selector, args);
    if(tuning.notNil) { ^tuning.deepCopy };
    ^super.doesNotUnderstand(selector, args)
```

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -185,8 +189,16 @@ Tuning {
...
+		!? { |v| v } //return if valid
+		?? { super.doesNotUnderstand(selector, args) }
+	}
+
+	*doesNotUnderstandWithKeys {|selector, argsArray, keywordArgsAsPairs|
```
telephon, line 197:
```
^Tuning.performWith(\newFromKey, [selector] ++ argsArray, keywordArgsAsPairs)
```

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -700,8 +746,8 @@ Interpreter {
 		protect {
```
telephon, line 746: could you undo the formatting here? I think it helps for future readability to know – at least – which files were actually touched an which weren't.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -1375,6 +1375,9 @@ void PyrMethodNode::compile(PyrSlot* result) {
         */
     } else if (slotRawSymbol(&gCompilingMethod->name) == s_doesNotUnderstand) {
         methType = methNormal;
+    } else if (slotRawSymbol(&gCompilingMethod->name) == s_doesNotUnderstandWithKeys) {
+        // this branch could be incorrect as I don't know what it does.
```
telephon, line 1379: is this something to do before we finish up?

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -74,10 +74,18 @@ Function : AbstractFunction {
...
 
+	valueWith {|argumentsArray=([]), keywordArgumentEnvir=(())|
+		^this.valueArray(def.makePerformableArray(argumentsArray, keywordArgumentEnvir));
+	}
+
 	performWithEnvir { |selector, envir|
 		if(selector === \value) { ^this.valueWithEnvir(envir) };
 		^super.performWithEnvir(selector, envir)
 	}
+	performWith {|selector, argumentsArray=([]), keywordArgumentEnvir=(())|
```
telephon, line 85: default arguments are given in `makePerformableArray` so `argumentsArray=([])` can simply be `argumentsArray`. This will be more efficient. Same for `keywordArgumentEnvir`.

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -137,6 +137,16 @@ Object  {
 		^this.performWithEnvir(selector, ().putPairs(pairs))
 	}
 
+	performWith {|selector, argumentsArray=([]), keywordArgumentEnvir=(())|
```
telephon, line 140: default arguments are given in `makePerformableArray` so `argumentsArray=([])` can simply be `argumentsArray`. This will be more efficient. Same for `keywordArgumentEnvir`.

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -169,6 +179,13 @@ Object  {
 	valueEnvir { ^this }
 	valueArrayEnvir { ^this }
 
+	valueWith {|argumentsArray=([]), keywordArgumentEnvir=(())|
```
telephon, line 182: default arguments are given in `makePerformableArray` so `argumentsArray=([])` can simply be `argumentsArray`. This will be more efficient. Same for `keywordArgumentEnvir`.

HelpSource/Classes/FunctionDef.schelp
```diff
@@ -45,18 +45,30 @@ return a list of all references to a given symbol.
...
 
-Get the Array of Symbols of the argument names.
+Get all of the arguments' names as a link::Classes/SymbolArray::.
 
 code::
 { |a = 9, b = 10, c| a + b }.def.argNames;
 ::
+
+method::argumentNamesForCall
+Same as link::#-argName::, but guarantees only the argument names the user sees are present.
```
telephon, line 55: Same as link::#-argName::, but guarantees only the argument names the user sees are present.
–>
Returns the argument names in code (link::#-argNames:: may have `this` as first item).

HelpSource/Classes/FunctionDef.schelp
```diff
@@ -110,6 +122,50 @@ code::
 { |a = 9, b = 10, c| a + b }.def.makeEnvirFromArgs;
 ::
 
+method::makePerformableArray
+Returns an array that can be passed to link::Classes/Object#-perform:: or link::Classes/Function#-value::.
+
+Both keyword and non-keyword arguments may be supplied at the same time.
+If a value for an argument is provided in both the keyword and non-keyword arguments this will throw an error.
+If a keyword argument is provide, but the name can't be found an error will be thrown.
```
telephon, line 130: provided

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -413,10 +413,17 @@ Process {
...
 
 FunctionDef {
-	var raw1, raw2, <code, <selectors, <constants, <prototypeFrame, <context, <argNames, <varNames;
-	var <sourceCode;
-
-	// a FunctionDef is defined by a code within curly braces {}
+	var raw1, raw2; // Internal - don't touch
+	var <code; // Compiled bytecode
+	var <selectors; // All the explicit selectors (messages) used
+	var <constants; // Constants used in function
```
telephon, line 419: "// All the constants used"

(for consistency)

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -452,6 +462,10 @@ FunctionDef {
 		_FunctionDefDumpContexts
 		^this.primitiveFailed
 	}
+
+	defaultArgs { ^this.prototypeFrame.keep(argNames.size).copy }
```
telephon, line 466: Since this is a basic method, let's avoid double copying, `keep` returns a new array also if n = a.size:

```

a = [1, 2, 3];
b = a.keep(3);
a === b // false
```

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,12 +563,45 @@ FunctionDef {
 		^().putPairs(this.keyValuePairsFromArgs)
 	}
 
+	makePerformableArray {|argumentsArray=([]), keywordArgumentEnvir=(())|
+		// This will return an array with all the arguments put in place.
+		// Variable arguments are place at the end (not in a seperate array).
+		// Consider moving this whole method to a primitive.
+		var arguments = this.defaultArgs
+		.drop(if(this.varArgs, -1, 0)) // the default provided by the prototypeFrame for varArgs '[]' is incorrect, so drop it.
```
telephon, line 571: It is not incorrect but unsuitable, let's write:

```
// here we don't want the default provided by the prototypeFrame for varArgs '[]', so drop it.
```

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,12 +563,45 @@ FunctionDef {
...
+		.overWrite(argumentsArray);
+		var argNamesCall = this.argumentNamesForCall;
+		var addedNames = argNamesCall.keep(argumentsArray.size);
+
+		// check keywordArgumentEnvir do not collide with a	argumentsArray,
+		//    and that they are present in the method.
+		keywordArgumentEnvir.keysValuesDo{|name, a|
+			if (addedNames.includes(name)) {
+				Error(
+					"Arguments cannot be duplicated, "
```
telephon, line 581: "The same argument cannot be passed twice, "

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,12 +563,45 @@ FunctionDef {
...
+					"got two values for argument '%'".format(name)
+				).throw
+			};
+
+			argNamesCall.detectIndex({|n| n == name }) ?? {
+				Error("Arguments '%' not understood by function '%'".format(name, this)).throw
```
telephon, line 587: Arguments –> Argument

telephon (commented):
I've added some more comments, these are just for finishing up. This looks really good now!

Optimisations can be done in a separate issue later?

SCClassLibrary/Common/Collections/Dictionary.sc
JordanHendersonMusic: Sorry I thought I got these. I've un-done all of them in the next.

SCClassLibrary/Common/Collections/Dictionary.sc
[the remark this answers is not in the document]
JordanHendersonMusic: Actually I think this bit of the code is wrong because right now it only works with functions, and won't work with other classes as it just looks up the `def`. I'll change this now.

JordanHendersonMusic:
> This looks really good now!

I found a little bug and have now made a bigger change to fix it !

I've removed all calls to functionPerformList and replace them with functionPerformValue. This does the optimization when it is a single value, and makes the code much neater. Let me know if you think this is a good idea, without it though, there would have to be a switch inside Identity dictionary to figure out if the function def should be looked up, or the method for 'value'.

lang/LangSource/PyrParseNode.cpp
JordanHendersonMusic: Well it hasn't caused any bugs? 

... but I have no idea what that line does because I couldn't figured out what an 'abnormal' method would be.
I just kept the symmetry.

telephon: I see. The comment is misleading because you added the code yourself :)

Maybe you could explicitly say that this code was copied from `s_doesNotUnderstand` without knowing what it does?

JordanHendersonMusic: Actually, I've removed the comment now as the options are
```
enum {
    methNormal = 0,
    methReturnSelf,
    methReturnLiteral,
    methReturnArg,
    methReturnInstVar,
    methAssignInstVar,
    methReturnClassVar,
    methAssignClassVar,
    methRedirect,
    methRedirectSuper,
    methForwardInstVar,
    methForwardClassVar,
    methPrimitive,
    methBlock
};
```
It is clearly a methNormal.

telephon (changes requested):
It is good that you have added `functionPerformWith`, things look clearer. 

But please don't remove `functionPerformList`, because:
- it is optimized in a primitive for the non-keyword case. Passing arguments is the slower part of supercollider and object prototyping should be as fast as possible
- code by others may depend on the method, we can never just remove a method. Deprecating methods is also not a very smooth process.

Deprecating would mean that we go through a deprecation "ceremony" (first amending the method with a deprecation post, then removing it after a long waiting period).

But it is not worth deprecating it. Since we have the pairing `performList` and `performWith`, it it completely fine to have `functionPerformList` and `functionPerformWith`. I'd even expect it to exist.

Sorry to be a roadblock here!

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -74,10 +74,18 @@ Function : AbstractFunction {
 		^this.valueArray(prototypeFrame)
 	}
 
+	valueWith {|argumentsArray=([]), keywordArgumentEnvir=(())|
```
telephon, line 77: also no need for the default args here (sorry, forgot to mention these)

JordanHendersonMusic:
> let's not deprecate functionPerformList
Alright I've reverted that.

It does mean that these two function must be kept in sync or behaviour will change depending on whether or not there are keyword arguments.

telephon (approved):
thank you!

telephon:
> It does mean that these two function must be kept in sync or behaviour will change depending on whether or not there are keyword arguments.

ok, I hdn't thought of this. 

Maybe, in the future, we have a discussion about deprecation separately? Maybe together with a primitive which would speed it up.

Or, one way to solve this compatibility issue would be to give `functionPerformList` one more argument, with the optional keyword envir. Then the two methods would be the same.

JordanHendersonMusic:
> Or, one way to solve this compatibility issue would be to give functionPerformList one more argument, with the optional keyword envir. Then the two methods would be the same.

I don't think that is possible because it isn't called via an array (nor list) in the code, everywhere calls it like this, although the arguments suggest otherwise?!
```supecollider
func.functionPerformList(\value, this, args);
```
I don't know how this works because Fdef has this...
```supercollider
functionPerformList { arg selector, arglist;
   ^this.performList(selector, arglist)
}
```
Meaning all places would not pass the arguments (expect the first) through Fdef correctly. I don't really know how Fdef works so can't test it properly.

---

I was kinda thinking something different.

How about we expand `performWith` and `valueWith` to take a third argument, an array of varArgs, and a fourth a bool, indicating whether arguments should be looked up in the currentEnvironment? 

Then we could have:

* `value` — for when the user calls a function directly (this would remain untouched)
* `valueWith` — for when the 'code' calls a function
* `performWith` — for when the 'code' calls a method
* `funtionPerformWith` — for when object prototyping calls a function

All four could be implemented as primitives.

This, I think, covers all cases, meaning we can remove all other `value*` or `perform*` methods.

I know we got rid of the varArgs, but to make it more easily compatible with `valueArray` this might be a compromise — although I wonder if `valueArray` call syntax is really needed?

Compare this to what we have now; there are many possible combinations of postfixes to value and perform, `valueArrayEnvir`, `valueArray`, `valueEnvir`, and not all the possible combinations have been made yet —  `performArrayEnvir`?

This would be slightly slower when calling the `*With` methods, as you'd need to put the arguments in an array, but vastly reduce the number of ways something can be called.

JordanHendersonMusic:
Oh, and thanks a bunch for helping with this! It looks so much better that when I started.

telephon:
> I don't know how this works because Fdef has this...

it seems wrong and broken, I just didn't notice.
But it works correctly for some reason:
```

Fdef(\x, { |...args| args });
q = (finch:Fdef(\x), func:{ |...args| args });
Fdef(\x).value(8, 3);
q.finch(8, 3)
q.func(8, 3)

```

telephon:
> How about we expand `performWith` and `valueWith` to take a third argument, an array of varArgs, and a fourth a bool, indicating whether arguments should be looked up in the currentEnvironment?

So the idea would be to get rid of `valueEnvir` and `valueArrayEnvir`? I fear that this is blocked by the fact that it takes variable args, so you can't add extra arguments.

Perhaps we can rethink this whole interface in a separate issue, maybe even a dreaded RFC ...

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -89,7 +89,13 @@ Object  {
 		^this.primitiveFailed
 	}
 	functionPerformList {
-		// perform only if Function. see Function-functionPerformList
+		// Used to opt into object prototyping. see Function-functionPerformList
```
telephon, line 92: can you reformat like this:

```
// Used in object prototyping. See Function-functionPerformList
// Ensure both functionPerformList and functionPerformWith do the same thing.
	functionPerformList { ^this }
	functionPerformWith { ^this }
```

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -137,6 +143,16 @@ Object  {
 		^this.performWithEnvir(selector, ().putPairs(pairs))
 	}
 
+	performWith {|selector, argumentsArray, keywordArgumentEnvir|
```
telephon, line 146: As long as we don't have a primitive, this could be optimised. I think that performList is faster than the star, but I am not sure.

```
performWith {|selector, argumentsArray, keywordArgumentEnvir|
		var method, argList;
		if(keywordArgumentEnvir.isNil) { 
				^this.performList(selector, argumentsArray) 
		};
		method  = this.class.findRespondingMethodFor(selector) ?? {
			^this.doesNotUnderstand(selector)
		};
		argList = method.makePerformableArray(argumentsArray, keywordArgumentEnvir;
		^this.performList(selector, argList)
		)
	}
```

telephon:
> don't think that is possible because it isn't called via an array (nor list) in the code, everywhere calls it like this, although the arguments suggest otherwise?!
> 
> ```
> func.functionPerformList(\value, this, args);
> ```

Hm: https://scsynth.org/t/object-prototyping-is-not-broken-for-higher-oder-functions-but-why/9171

telephon:
> I was kinda thinking something different.
> 
> How about we expand `performWith` and `valueWith` to take a third argument, an array of varArgs, and a fourth a bool, indicating whether arguments should be looked up in the currentEnvironment?
> 
> Then we could have:
> 
>     * `value` — for when the user calls a function directly (this would remain untouched)
> 
>     * `valueWith` — for when the 'code' calls a function
> 
>     * `performWith` — for when the 'code' calls a method
> 
>     * `funtionPerformWith` — for when object prototyping calls a function
> 
> 
> All four could be implemented as primitives.
> 
> This, I think, covers all cases, meaning we can remove all other `value*` or `perform*` methods.
> 
> I know we got rid of the varArgs, but to make it more easily compatible with `valueArray` this might be a compromise — although I wonder if `valueArray` call syntax is really needed?
> 
> Compare this to what we have now; there are many possible combinations of postfixes to value and perform, `valueArrayEnvir`, `valueArray`, `valueEnvir`, and not all the possible combinations have been made yet — `performArrayEnvir`?
> 
> This would be slightly slower when calling the `*With` methods, as you'd need to put the arguments in an array, but vastly reduce the number of ways something can be called.

@JordanHendersonMusic should we discuss this now or shift it to a future "reform"? We can always add arguments to the methods `somethingWith`, so going forward with the current PR won't do harm as far as I can see.

Are there any open points we should discuss before we go through it once more for cosmetics?

JordanHendersonMusic:
Is rather discuss that in the future as it's a larger change that isn't necessarily connected to this. I'd also need some time to play around with the idea a bit.

I think everything is settled, there hadn't really been a review of the C++ but it is okay.

telephon:
Yes, I agree. So I'll go through everything and make copy editing level suggestions.

HelpSource/Classes/FunctionDef.schelp
```diff
@@ -110,6 +122,50 @@ code::
...
+f = { |a b ...e|
+   var i = 0;
+   a.debug(\a);
+   b.debug(\b);
+   e.debug(\e);
+};
+
+f.value(
+	*f.def.makePerformableArray([\a], (e: [1,2,3]))
+);
```
telephon, line 161: (parentheses round regions, and explicit names for readability)

```
code::

(
f = { |a b ...e|
   var i = 0;
   a.debug(\a);
   b.debug(\b);
   e.debug(\e);
}
);


(
var args = f.def.makePerformableArray([\a], (e: [1,2,3]));
f.value(*args)
);
```

HelpSource/Classes/Method.schelp
```diff
@@ -40,14 +43,23 @@ instanceMethods::
 method::ownerClass
 returns::
 The link::Classes/Class:: for which the method is part of the implementation.
+If this is a class method, then the owner class will be a 'Meta_Class'.
```
telephon, line 46: If the receiver is an instance method, the owner class is the class for this instance. For a class method, it is the class of the class, whose name is prefixed with "Meta_".

code::
m = Array.findRespondingMethodFor('reverse');
m.ownerClass; // Array
m = Array.class.findRespondingMethodFor(\new);
m.ownerClass; // Meta_Array
::

HelpSource/Classes/Method.schelp
```diff
@@ -40,14 +43,23 @@ instanceMethods::
...
 A link::Classes/Symbol:: which is the name of the Method.
 
 method::primitiveName
 returns::
-A link::Classes/Symbol:: which contains the name of the primitive function that implements the Method, if there is one.
+A link::Classes/Symbol:: which contains the name of the primitive function that implements the Method,
+if there is one, otherwise returns link::Classes/Nil::.
+
+method::defaultArgs
+Returns the default arguments needed to call the function.
```
telephon, line 58: Returns the default arguments supplied in place of those not specified in a method call.

HelpSource/Classes/Object.schelp
```diff
@@ -516,6 +516,31 @@ a = { |a, b, c| postf("% plus % plus % is %\n", a, b, c, a + b + c); "" };
...
+argument::argumentsArray
+Arguments as an link::Classes/Array::.
+
+argument::keywordArgumentEnvir
+Keyword arguments as an link::Classes/IdentityDictionary::.
+
+discussion::
+code::
+f = { |a, b ...e|
```
telephon, line 531:
```

7819.performWith(\asTimeString, [0.1, 365], (decimalPlaces:4));
{ SinOsc.ar }.performWith(\plot, [0.05], (minval:0, maxval:1));

// calling value on function shows how arguments are assigned:
(
f = { |a, b ...e|
   a.debug(\a);
   b.debug(\b);
   e.debug(\e);
};

f.performWith(\value, [\a], (e: [1,2,3]));
)
```

SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -560,12 +560,32 @@ IdentityDictionary : Dictionary {
...
+		this[selector] !? {|f|
+			^f.functionPerformWith(\value, [this] ++ argsArray, keywordArgsAsPairs.asEvent);
+		};
+
+		// Unlike in doesNotUnderstand, we don't convert to a setter as this isn't
+		// possible with keyword args.
+
+		// Note how the selector is passed in here.
+		this[\forward] !? {|f|
+			^f.functionPerformWith(\value, [this, selector] ++ argsArray, keywordArgsAsPairs.asEvent)
```
telephon, line 577: better remove `asEvent` here and call it in `functionPerformWith`, which is the latest point of conversion.

There, it should be `asDict` because it is faster.

SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -560,12 +560,32 @@ IdentityDictionary : Dictionary {
...
-		// (nextTimeOnGrid: { |self, clock| ... calculate absolute beat number here ... },
-		//	parameter: value, parameter: value, etc.)
-		// If you leave out the nextTimeOnGrid function, fallback to quant/phase/offset.
+	doesNotUnderstandWithKeys {|selector, argsArray, keywordArgsAsPairs|
+		if(know.not){
+			^this.superPerformList(\doesNotUnderstandWithKeys, selector, argsArray, keywordArgsAsPairs )
+		};
+
+		this[selector] !? {|f|
+			^f.functionPerformWith(\value, [this] ++ argsArray, keywordArgsAsPairs.asEvent);
```
telephon, line 569: better remove `asEvent` here. In`functionPerformWith`, which is the latest point of conversion, it can also be a key value pair, because there, `keysValuesDo` is called, which works for both anyway. This also saves some memory for conversions.

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -185,8 +190,13 @@ Tuning {
 	}
 
 	*doesNotUnderstand { |selector, args|
-		var tuning = this.newFromKey(selector, args).deepCopy;
-		^tuning ?? { super.doesNotUnderstand(selector, args) }
+	    var tuning = this.newFromKey(selector, args);
+        if(tuning.notNil) { ^tuning.deepCopy };
```
telephon, line 194: `newFromKey` already copies, so this can be:
```
if(tuning.notNil) { ^tuning};
```

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -185,8 +190,13 @@ Tuning {
...
+        ^super.doesNotUnderstand(selector, args)
+	}
+
+	*doesNotUnderstandWithKeys {|selector, argsArray, keywordArgsAsPairs|
+        ^Tuning.performWith(\newFromKey, [selector] ++ argsArray, keywordArgsAsPairs.asEvent);
```
telephon, line 199: better remove `asEvent` here. In`functionPerformWith`, which is the latest point of conversion, it can also be a key value pair, because there, `keysValuesDo` is called, which works for both anyway. This also saves some memory for conversions.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -549,12 +560,46 @@ FunctionDef {
...
+
+		^arguments
+	}
 }
 
 Method : FunctionDef {
 	var <ownerClass, <name, <primitiveName;
 	var <filenameSymbol, <charPos;
 
+	defaultArgs { ^this.prototypeFrame.drop(1).keep(argNames.size - 1) }
```
telephon, line 600: clearer:
```
^this.prototypeFrame.keep(argNames.size).drop(1) }
```

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -88,10 +88,12 @@ Object  {
...
-	functionPerformList {
-		// perform only if Function. see Function-functionPerformList
-		^this
-	}
+	// Used in object prototyping. See Function-functionPerformList
+	// Ensure both functionPerformList and functionPerformWith do the same thing.
+	functionPerformList { ^this }
+	// Used in object prototyping. See Function-functionPerformList
+	// Ensure both functionPerformList and functionPerformWith do the same thing.
+	functionPerformWith {^this }
```
telephon, line 96: {^this }
–>
{ ^this }

telephon (commented):
So this is a last copy edit, I just found one structural thing (we can drop the conversions with `asDict`).

The keywordArgs argument names should still be unified across the code. Currently we have: `keywordArgumentEnvir`, `argsWithKeys`, and `keywordArgsAsPairs`. Since no conversion needs to be done, these can now all be the same. E.g. `keywordArgs`.

Maybe you can add a brief comment in the help files that anything that responds correctly to `keysValuesDo` will do.

JordanHendersonMusic:
Keeping the docs all synchronised is harder than fixing the behaviour....


I've renamed all the references to keywordArgumentEnvir to keywordArgumentPairs.
Also fixed all the docs and formatting.

HelpSource/Classes/Method.schelp
```diff
@@ -54,14 +61,55 @@ returns::
 A link::Classes/Symbol:: which contains the name of the primitive function that implements the Method,
 if there is one, otherwise returns link::Classes/Nil::.
 
-method::defaultArgs
-Returns the default arguments needed to call the function.
-This will correctly omit 'this'.
-
 method::argumentNamesForCall
-Same as link::#-argName:: but will omit 'this', useful when calling the method.
+Returns the argument names in as needed for calling the function as an link::Classes/Array::.
```
telephon, line 65: argument names in as
–>
argument names as

SCClassLibrary/Common/Core/Error.sc
```diff
@@ -192,12 +192,12 @@ DoesNotUnderstandError : MethodError {
 }
 
 DoesNotUnderstandWithKeysError : MethodError {
-	var <>selector, <>argsWithOutKeys, <>argsWithKeys, <suggestedCorrection, suggestion = "";
-	*new { arg receiver, selector, argsWithOutKeys, argsWithKeys;
+	var <>selector, <>argumentsArray, <>keywordArgumentPairs, <suggestedCorrection, suggestion = "";
+	*new { arg receiver, selector, argumentsArray, keywordArgumentPairs;
 		^super.new(nil, receiver)
```
telephon, line 197: if you want, you could write: `^super.newCopyArgs(selector, receiver, argumentsArray, keywordArgumentPairs).init`

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -53,13 +53,18 @@ Function : AbstractFunction {
...
 		^this.primitiveFailed
 	}
-	functionPerformWith {|selector, argumentsArray, keywordArgumentEnvir|
-		keywordArgumentEnvir ?? { this.functionPerformList(selector, argumentsArray) };
-		^this.valueWith(argumentsArray, keywordArgumentEnvir);
+	functionPerformWith {|selector, argumentsArray, keywordArgumentPairs|
+		// Used in object prototyping.
+		// Ensure both functionPerformList and functionPerformWith do the same thing.
+		keywordArgumentPairs ?? { ^this.functionPerformList(selector, argumentsArray) };
+		^this.valueWith(argumentsArray, keywordArgumentPairs);
```
telephon, line 67: drop closing semicolon suggests return

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -78,17 +83,17 @@ Function : AbstractFunction {
 		^this.valueArray(prototypeFrame)
 	}
 
-	valueWith {|argumentsArray, keywordArgumentEnvir|
-		^this.valueArray(def.makePerformableArray(argumentsArray, keywordArgumentEnvir));
+	valueWith {|argumentsArray, keywordArgumentPairs|
+		^this.valueArray(def.makePerformableArray(argumentsArray, keywordArgumentPairs));
```
telephon, line 87: drop closing semicolon suggests return

telephon (commented):
just final cleanup.

telephon (approved):

JordanHendersonMusic:
Would it be useful if I squashed the commits? Or is that something that gets done automatically?

telephon:
Ideally, there should be a thidr person looking over it and would then squash to commit.

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -19,8 +19,13 @@ Scale {
 	}
 
 	*doesNotUnderstand { |selector, args|
-		var scale = this.newFromKey(selector, args).deepCopy;
-		^scale ?? { super.doesNotUnderstand(selector, args) };
+	    var scale = this.newFromKey(selector, args);
+        if(scale.notNil) { ^scale };
```
lennart, line 23: just indentation

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -185,8 +190,13 @@ Tuning {
 	}
 
 	*doesNotUnderstand { |selector, args|
-		var tuning = this.newFromKey(selector, args).deepCopy;
-		^tuning ?? { super.doesNotUnderstand(selector, args) }
+	    var tuning = this.newFromKey(selector, args);
+        if(tuning.notNil) { ^tuning };
```
lennart, line 194: again just indentation

SCClassLibrary/Common/Core/Error.sc
```diff
@@ -191,6 +191,56 @@ DoesNotUnderstandError : MethodError {
...
+		"RECEIVER:\n".post;
+		receiver.dump;
+		"ARGS WITHOUT KEYS:\n".post;
+		argumentsArray.dumpAll;
+		"ARGS WITH KEYS:\n".post;
+		keywordArgumentPairs.dumpAll;
+		this.errorPathString.post;
+		if(protectedBacktrace.notNil, { this.postProtectedBacktrace });
+		this.dumpBackTrace;
+		// this.adviceLink.postln;
```
lennart, line 237: is this going to be used? otherwise remove this line

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -169,6 +183,14 @@ Object  {
 	valueEnvir { ^this }
 	valueArrayEnvir { ^this }
 
+	valueWith {|argumentsArray, keywordArgumentPairs|
+		^this.valueArray(
+			this.class
+                .findRespondingMethodFor(\value)
```
lennart, line 189: just indentation

lennart (commented):
I read through the whole PR (not yet into the depths of the codebase) so I only noted potential stylistic corrections. Apart from that, as far as I can see, this PR does what it claims to do.

SCClassLibrary/Common/Collections/Scale.sc
JordanHendersonMusic: Seems I've used spaces instead of tabs somehow... I'll fix shortly.

SCClassLibrary/Common/Core/Error.sc
JordanHendersonMusic: Its used in Object.doesNotUnderstandWithKeys

SCClassLibrary/Common/Core/Object.sc
JordanHendersonMusic: SCIDE does not let me insert the tab on the new line. I've had to use another program to follow the style guide.

SCClassLibrary/Common/Core/Error.sc
lennart: mh, maybe I misunderstood, but this `adviceLink` ist currently commented out, I was wondering if this line can be removed (following the idea of: no working code in comments)

JordanHendersonMusic: Ah, no I thought you meant the whole class. 

I just copied what DoesNotUnderstandError does, but looking at it a few weeks later, I'm going to remove this class.

adcxyz:
This will be brilliantly useful, for error messages in keyword calls, 
and especially for object modeling with dictionaries (which I use a lot).
ATM, I cannot test because compiling on macOS has hurdles, 
but will do ASAP. 
Thanks!

JordanHendersonMusic:
So because this is actually quite a large change. I'm going to make a new PR with just a couple of small commits in to keep history clean (sounds like something a fascist would say...) and to make this easier to review.

@adcxyz thanks, its something I've ran into in many places because it's needed to make proper wrapper classes!

JordanHendersonMusic:
@telephon @adcxyz @nuss @lennart 

I'm closing this PR because its a mess. 
#6297 is the neat version, with only 4 commits.

PR #6229, opened 2024-03-05
Add atomic bool const to lockLanguageOrQuit in PyrSched.h

smoge:
## Purpose and Motivation

Change `lockLanguageOrQuit` to use `std::atomic<bool> const&`

Modified the `lockLanguageOrQuit` function in PyrSched.h to take a constant reference to std::atomic<bool>. 

The actual problem was that the function took the atomic flag by value instead of by reference. This made the repeated checks entirely redundant and as a consequence it would not notice when the flag has changed in between, undermining the very purpose of this function.

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/lang/LangSource/PyrSched.h
+++ b/lang/LangSource/PyrSched.h
@@ -24,6 +24,8 @@
 #include "SC_Export.h"
 #include "SC_Lock.h"
 #include <cerrno>
+#include <atomic>
+#include <chrono>
 
 extern timed_mutex gLangMutex;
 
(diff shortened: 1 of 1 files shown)


Spacechild1 (approved):

telephon:
Just for reference, what is the purpose?

smoge:
> Just for reference, what is the purpose?

 <atomic> and <chrono>  are already part of the codebase. They were included to enhance the thread safety and precision of the timing and scheduling 
 
The purpose of this patch is that this file was not updated.

Now reads and writes to shouldBeRunning are "atomic", they happen as a single operation. In a multi-threaded environment it matters to prevent race conditions and variable's state is consistent across threads.

Spacechild1:
The original code was perfectly threadsafe, it just was not correct :)

The actual problem was that the function took the atomic flag by value instead of by reference. This made the repeated checks entirely redundant and as a consequence it would not notice when the flag has changes in between, undermining the very purpose of this function.

Spacechild1:
> It would be good if we had more of such explanations in the commits themselves for later …

Certainly! @smoge Can you ammend the commit message and force push?

smoge:
> > It would be good if we had more of such explanations in the commits themselves for later …
> 
> Certainly! @smoge Can you ammend the commit message and force push?

Done.

smoge:
Git added, "Co-authored-by: Bernardo Barros <bernardo.barros@protonmail.ch>." I have no idea why!

If I included credits at all, that would go for @Spacechild1  for very generously tutoring me on C++ hacking. Just want to register that.

Thank you.

dyfer:
Thanks for reviewing and merging @Spacechild1 @telephon 
And thanks for the PR @smoge

PR #6231, opened 2024-03-08
macOS 14 support / bump CI to macOS 14

capital-G:
## Purpose and Motivation

The current CI runs on macOS 12 which is the last supported version of macOS.
We should always include the current supported macOS versions in our CI and this PR tries to achieve this.
Ideally this would be implemented with a matrix test, but maybe only for one specific kind of build instead of all types macOS builds? This could be discussed here.

There is the potential that macOS 14 introduced some breaking (see https://scsynth.org/t/boost-fail-to-compile-on-macos-14-1-1/8612) - with a more up-to-date pipeline, this could have been mitigated at a prior stage.

It seems that the current build pipeline is broken on macOS 14 (according to CI run @ https://github.com/capital-G/supercollider/actions/runs/8208341214), so while we are at it, we can try to fix the issues here.

I am still on macOS 13, but I think we can try to get this done in a collaborative manner.

This is still WIP, but anyone is welcome to contribute - hopefully it is nothing too big.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -197,7 +197,7 @@ jobs:
 
           - job-name: 'universal'
             os-version: '14'
-            xcode-version: '15.3'
+            xcode-version: '15.2'
             deployment-target: '10.14'
             cmake-architectures: 'x86_64;arm64'
             use-syslibs: false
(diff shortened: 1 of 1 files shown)


capital-G:
As it is likely that boost needs an update here - should we remove the existing boost code in the repo and link it via a git submodule - `externals` has a mixture of both, is this due to legacy or is there a reason behind it?
This seems to me a better way to manage the code and does not introduce so much noise during an update procedure and makes updating much more easy.

PR #6235, opened 2024-03-15
Topic/fix plot build synth def

telephon:
## Purpose and Motivation

Fixes #6206.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing (plotter tests)
- [x] This PR is ready for review

PR #6236, opened 2024-03-15
clean up named control conditions

telephon:
## Purpose and Motivation

Just a small performance improvement and cleanup, follow up on the bugfix in #6235.

## Types of changes

minor edits.

## To-do list

- [ ] Code is tested
- [ ] All tests are passing

### DIFF
--- a/HelpSource/Classes/Function.schelp
+++ b/HelpSource/Classes/Function.schelp
@@ -582,7 +582,7 @@ argument::target
 A server, synth, or group where to place the synth that plays the function
 
 argument::action
-A function to call when writing is finished.
+A function to call when writing is finished. It is passed two arguments: the array and the buffer used for plotting.
 
 code::
 // get some data from the synth audio
@@ -606,7 +606,7 @@ argument::target
 A server, synth, or group where to place the synth that plays the function
 
 argument::action
-A function to call when writing is finished.
+A function to call when writing is finished.  It is passed two arguments: the array and the buffer used for plotting.
 
 argument:: wait
 The amount of time in seconds to wait between sending getn messages. Longer times are safer. The default is 0.01 seconds which seems reliable under normal circumstances. A setting of 0 is not recommended.
--- a/SCClassLibrary/Common/Control/GraphBuilder.sc
+++ b/SCClassLibrary/Common/Control/GraphBuilder.sc
@@ -78,10 +78,10 @@ NamedControl {
 
 		name = name.asSymbol;
 
-		if (spec.notNil) {
+		if(spec.notNil) {
 			spec = spec.asSpec;
 
-			if (values.isNil) {
+			if(values.isNil) {
 				values = spec.default;
 			};
 		};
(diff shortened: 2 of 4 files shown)


dyfer:
@telephon what's the status of this PR?

telephon:
I am still unsure if we should set the sample rate of the buffer produced in this way. The change would be that a buffer recorded at control rate and then played back at audio rate would come out correctly. 

As it is now, you just have to know if you have recorded something at control rate and then play it back at control rate again.
A change, however, may require a change also in the way `PlayBuf.kr` interprets the rate scaler. I am not sure if this is a good idea.

```supercollider
b = Buffer.alloc(s, 16, 1);
{ RecordBuf.kr(inputArray: LFNoise1.kr(200), bufnum: b, loop: 0, doneAction: 2) }.play;
{ PlayBuf.kr(1, b, BufRateScale.kr(b)) }.plot(s.options.blockSize / s.sampleRate * b.numFrames);
```

dyfer:
IIRC this part of code was recently brought up by @mtmccrea in context of #6710

telephon:
Yes, so let's leave the present one unmerged until we have a clean way that dosn't change any behavior.

> IIRC this part of code was recently brought up by @mtmccrea in context of #6710.

If the sample rate is set explicitly, we don't hae to worry, I suppose …

telephon:
Just for the future, here are the examples:

```supercollider
b = Buffer.alloc(s, 16, 1);
{ RecordBuf.kr(inputArray: LFNoise1.kr(200), bufnum: b, loop: 0, doneAction: 2) }.play;
{ PlayBuf.kr(1, b, BufRateScale.kr(b)) }.plot(s.options.blockSize / s.sampleRate * b.numFrames);
{ PlayBuf.ar(1, b, BufRateScale.kr(b)) }.plot(s.options.blockSize / s.sampleRate * b.numFrames);
{ PlayBuf.ar(1, b, BufRateScale.kr(b)) }.plot(1 / s.sampleRate * b.numFrames);

```

mtmccrea:
> IIRC this part of code was recently brought up by @mtmccrea in context of https://github.com/supercollider/supercollider/pull/6710.

In part... @telephon and I already implemented the changes from _Function.sc_ and _PlotView.sc_ in #6217. But changes to _GraphBuilder.sc_ in this PR are unique.

>BufRateScale is tricky, as it doesn't know (IIUC?) whether Playbuf is .kr or .ar, right?

If the server-side sample rate can be set (#6710), `BufRateScale` should work for playing back kr buffers as long as `PlayBuf`, _always_ runs at ar, since `BufRateScale` provides a rate scaler relative to the server's sampling rate (according to the help file at least).

dyfer:
> `BufRateScale` should work for playing back kr buffers as long as `PlayBuf`, _always_ runs at ar

Yes, naturally. But the [examples](https://github.com/supercollider/supercollider/pull/6236#issuecomment-2817306016) use `PlayBuf.kr`. At least it needs to be pointed out that `BufRateScale` only works with `PlayBuf.ar` (unless we can actually infer the rate of the ugen and calculate accordingly?)

PR #6242, opened 2024-03-20
Make ScLang's integer 64bits

JordanHendersonMusic:
## Purpose and Motivation
More numbers good.

There *should* be no breaking changes.

This is currently more of a test to see what the CI makes of it.

## Types of changes
The slot already had a int64 in, this just updates everything else (mostly function signatures) and tries to keep the old behaviour where possible.

There are many many places where integers are just cast around willy nilly, no attempt is made to fix this here, but should happen in the future — I have tried to make some of these casts explicit.

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/QtCollider/primitives/prim_misc.cpp
+++ b/QtCollider/primitives/prim_misc.cpp
@@ -169,7 +169,7 @@ QC_LANG_PRIMITIVE(Qt_SetGlobalPalette, 1, PyrSlot* r, PyrSlot* a, VMGlobals* g) 
 // The line below is a workaround. The non-win term causes Error C2440 in VS
 // https://msdn.microsoft.com/en-us/library/sy5tsf8z.aspx
 #if defined(_MSC_VER)
-    QPalette p = (QPalette&&)QtCollider::get(a);
+    QPalette p = (QPalette &&) QtCollider::get(a);
 #else
     QPalette p = QtCollider::get(a);
 #endif
--- a/QtCollider/primitives/prim_misc.cpp
+++ b/QtCollider/primitives/prim_misc.cpp
@@ -169,7 +169,7 @@ QC_LANG_PRIMITIVE(Qt_SetGlobalPalette, 1, PyrSlot* r, PyrSlot* a, VMGlobals* g) 
 // The line below is a workaround. The non-win term causes Error C2440 in VS
 // https://msdn.microsoft.com/en-us/library/sy5tsf8z.aspx
 #if defined(_MSC_VER)
-    QPalette p = (QPalette &&) QtCollider::get(a);
+    QPalette p = (QPalette&&)QtCollider::get(a);
 #else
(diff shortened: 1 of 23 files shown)


JordanHendersonMusic:
Does anyone know why the CI failing to pull in the git submodules?

JordanHendersonMusic:
Just closing this because I no longer have time to play with it.

telephon:
It is not good to make integer 64 bit because it prevents a possible future optimisation. We should just use integer floats if we need higher numbers.

smoge:
> It is not good to make integer 64 bit because it prevents a possible future optimisation. We should just use integer floats if we need higher numbers.

At least on x86_64, there is no performance difference. It won't get faster nor worse. Although there are 32bit computers around. This change would have to allow different compilations. I think it hasn't been done because it's a massive and repetitive work.

JordanHendersonMusic:
Just for reference, here are some odd things I found...

1. Int already is 64bit on 64bit machines
https://github.com/supercollider/supercollider/blob/28a0b12216fb97e4487373b2cf00a5ec28850f1f/lang/LangSource/PyrSlot64.h#L57
2. So is char?
https://github.com/supercollider/supercollider/blob/28a0b12216fb97e4487373b2cf00a5ec28850f1f/lang/LangSource/PyrSlot64.h#L56
3. 32bit is enforced by casting through the getters and setter for the slot, but only in some cases on some architectures
32bit setter
https://github.com/supercollider/supercollider/blob/28a0b12216fb97e4487373b2cf00a5ec28850f1f/lang/LangSource/PyrSlot64.h#L165
This setter seems problematic because long is 64 on unix and 32 on windows (https://en.cppreference.com/w/cpp/language/types)
https://github.com/supercollider/supercollider/blob/28a0b12216fb97e4487373b2cf00a5ec28850f1f/lang/LangSource/PyrSlot64.h#L169

4. Bit shifting is weird
```
1 << 30 == 1073741824 // correct
1 << 31 == -2147483648 // correct
1 << 31 - 1 == 2147483647 // correct
1 << 36 == 0 // okay just sets higher bits to zero
1 << 60 == 0 // okay just sets higher bits to zero
1 << 64  == 1 /// hmmm....
1 << 65  == 2
```

smoge:
> 1 << 30 == 1073741824 // correct
> 1 << 31 == -2147483648 // correct but (I think) undefined behaviour according to the C++ spec
> 1 << 31 - 1 == 2147483647 // correct
> 1 << 36 == 0 // okay just sets higher bits to zero, standard C rules
> 1 << 60 == 0

This is the expected behavior. Tthe leftmost bit in a signed integer represents its sign (0 for positive, 1 for negative), and shifting a 1-bit left by 31 positions sets that leftmost bit to 1. A major developement because it facilitates operations between positive and negatives, and avois having " +0"  and "-0"

PR #6232, opened 2024-03-09
HID updates

beatboxchad:
## Purpose and Motivation
This PR addresses supercollider/hidapi#18, fixes #6016, and provides general updates to the HID subsystem in sclang. No major API changes are introduced.

## Types of changes

- Documentation
- Bug fix
- Tests
- Refactoring

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation
- [ ] This PR is ready for review

PR #6248, opened 2024-04-09
Add unit tests for SCLang HIDAPI client code

beatboxchad:
## Purpose and Motivation
I dug into the HID code and saw lots of beautiful work that inspired me to spend weeks writing notes about all the different things I could update/improve in addition to all the reasons that could explain the segfaults and memory leaks I was facing as I tried to use the HID client in my performances.

I realized it would not be plausible to accomplish any of it alone, nor without good unit test coverage. Thus, this PR represents the first step.

## Types of changes
- Unit Tests

## To-do list
- [ ] Code is tested
- [ ] All tests are passing


beatboxchad:
SuperCollider can't "see" my test UHID device unless I use the build I made in #6232. And that build still has bugs, like it makes a duplicate HIDInfo for each usage+page combo (or something -- I haven't looked closely yet). 

So my conclusion is that I need unit tests as part of that PR anyway, and this one is redundant. Sorry for the noise.

PR #6253, opened 2024-04-14
Help: Correct error in Linux env vars discussion

jamshark70:
## Purpose and Motivation

Help stated: "If these are not set, SuperCollider will default to connecting SuperCollider's inputs and outputs to your system's inputs and outputs." This is not true. PR matches current behavior.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/AudioDeviceSelection.schelp
+++ b/HelpSource/Reference/AudioDeviceSelection.schelp
@@ -92,7 +92,9 @@ teletype::SC_JACK_DEFAULT_OUTPUTS::. These allow SuperCollider to detect system 
 
 These variables are written as a string that specifies another jack client or a comma-separated list of jack ports formatted as a string.
 
-If these are not set, SuperCollider will default to connecting SuperCollider's inputs and outputs to your system's inputs and outputs. This is the recommended way of changing the Jack environment variables for SuperCollider from within a SuperCollider script:
+If these are not set, the server will not connect to any JACK ports automatically. You may connect them manually, or by issuing teletype::jack_connect:: / teletype::jack_disconnect:: shell commands (link::Classes/String#-unixCmd::). Note that the SuperCollider language client sets these environment variables to "system" by default, so that a server booted from SC language emphasis::will:: connect by default. However, a server booted from the command line will not auto-connect, unless you set the variables in the same terminal session or shell profile (below).
+
+This is the recommended way of changing the Jack environment variables for SuperCollider from within a SuperCollider script:
 
 code::
 // connect first to input channels with system
@@ -107,6 +109,13 @@ As an alternative, these may be also be changed by setting the following environ
 code::
 export SC_JACK_DEFAULT_INPUTS="system"
 export SC_JACK_DEFAULT_OUTPUTS="system"
+::
+
+To disable autoconnect from the language:
+
+code::
+"SC_JACK_DEFAULT_INPUTS".unsetenv;
+"SC_JACK_DEFAULT_OUTPUTS".unsetenv;
 ::
 
 section:: Windows
--- a/HelpSource/Reference/AudioDeviceSelection.schelp
+++ b/HelpSource/Reference/AudioDeviceSelection.schelp
@@ -92,7 +92,9 @@ teletype::SC_JACK_DEFAULT_OUTPUTS::. These allow SuperCollider to detect system 
 
 These variables are written as a string that specifies another jack client or a comma-separated list of jack ports formatted as a string.
 
-If these are not set, the server will not connect to any JACK ports automatically. You may connect them manually, or by issuing teletype::jack_connect:: / teletype::jack_disconnect:: shell commands (link::Classes/String#-unixCmd::). Note that the SuperCollider language client sets these environment variables to "system" by default, so that a server booted from SC language emphasis::will:: connect by default. However, a server booted from the command line will not auto-connect, unless you set the variables in the same terminal session or shell profile (below).
+The SuperCollider language client sets these environment variables to "system" by default, so that a server booted from SC language will connect by default.
+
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic (commented):
> Note that the SuperCollider language client sets these environment variables to "system" by default

It's probably better to mention this first, as it appears the user *has* to set these, because that's the first sentence.

jamshark70:
> It's probably better to mention this first, as it appears the user has to set these, because that's the first sentence.

OK, perhaps better now.

JordanHendersonMusic:
Looks good!

JordanHendersonMusic:
I suppose it's possible the user does not have jack_connect but does have jack, on arch these are separate packages... but I don't know if that needs to be mentioned here.

JordanHendersonMusic (changes requested):
Perhaps this is my short coming, but I don't know what the 'Supercollider language client' is? Is it possible to explain this behaviour without this term, if not, perhaps a link is needed?

JordanHendersonMusic (commented):
> You may connect them manually, or by issuing 

Jack_connect is an example of manual connection, not a separate option. Perhaps, 'for example with the command jack-connect'?

jamshark70:
In various places, we refer to "the language" and "the client." Actually there are multiple clients, in different languages; I was just being specific about the SC language.

By "manual connection" I meant graphically in qjackctl or carla.

JordanHendersonMusic:
> Actually there are multiple clients, in different languages; I was just being specific about the SC language.

When running from the terminal, which is still sclang, the variable isn't set, so it isn't (just) the language that controls this but something else? Is it the IDE?

jamshark70:
> When running from the terminal, which is still sclang, the variable isn't set

I did not find that to be the case in my testing.

```
$ echo $SC_JACK_DEFAULT_OUTPUTS

```

^^ Empty line = not set.

```
$ sclang
compiling class library...
... snip...
*** Welcome to SuperCollider 3.14.0-dev. ...

sc3> "SC_JACK_DEFAULT_OUTPUTS".getenv
-> system
```

^^ It's set inside sclang, run from the terminal.

And if I `0.exit` at this point, and `echo $SC_JACK_DEFAULT_OUTPUTS` again, it's empty again.

So what I wrote accurately reflects the results of my testing (in Ubuntu). If it behaves differently in different Linux distributions, then I'd need testing transcripts from other OSes.

JordanHendersonMusic:
Ah! Just and tested and I get the same results. 

I though this line...

> A server booted from the command line will thus not auto-connect, unless you set the variables in the same terminal session or shell profile (below).

... meant it wouldn't set the variable, and wouldn't auto-connect, but this refers to the audio server, not the language client booted from the command line.

This is my fault as your writing does say this, but perhaps this paragraph...

>If these are not set, the server will not connect to any JACK ports automatically.

...could make this even clearer by starting with something like...
>When booting the audio server from the command line  (scsynth, not sclang)...

jamshark70:
> When booting the audio server from the command line (scsynth, not sclang)...

I don't mind to make this change (but tomorrow).

One thing that is puzzling me here -- I can't find where sclang is setting the variables.

There is a bit in LinuxPlatform, but if I insert debugging statements:

```
		// default jack port hookup
		// use "system" as default when env vars haven't been set by user
		if("SC_JACK_DEFAULT_INPUTS".getenv.debug("INPUTS").isNil.debug("isNil"), {
			"SC_JACK_DEFAULT_INPUTS".debug("setting").setenv("system")
		});
```

I get:

```
Class tree inited in 0.01 seconds
INPUTS: system
isNil: false
```

... and no "setting" line -- because it's already set before we get here. I've doublechecked my startup.scd and I'm never touching SC_JACK_DEFAULT_INPUTS.

Grepping the entire repository doesn't show anything relevant: https://github.com/search?q=repo%3Asupercollider%2Fsupercollider+%2F%28%3F-i%29INPUTS%2F&type=code

So it is happening somewhere but apparently well hidden in our codebase.

JordanHendersonMusic:
Strange, because, on rebooting the IDE, I do see that it is being set. Did you reboot the IDE or just recompile? as the latter will keep the environment variables (as your help file says!).

```supercollider
'In Linux Platform'.postln;

// default jack port hookup
// use "system" as default when env vars haven't been set by user
if("SC_JACK_DEFAULT_INPUTS".getenv.isNil, {
	'SETTING SC_JACK_DEFAULT_INPUTS'.postln;
	"SC_JACK_DEFAULT_INPUTS".setenv("system")
});

if("SC_JACK_DEFAULT_OUTPUTS".getenv.isNil, {
	'SETTING SC_JACK_DEFAULT_OUTPUTS'.postln;
	"SC_JACK_DEFAULT_OUTPUTS".setenv("system")
});
```
```
compile done
Class tree inited in 0.01 seconds
In Linux Platform
SETTING SC_JACK_DEFAULT_INPUTS
SETTING SC_JACK_DEFAULT_OUTPUTS
```

mtmccrea (dismissed):
Thanks for these clarifying changes!

dyfer:
@jamshark70 any updates on this?

jamshark70:
@JordanHendersonMusic is correct -- the LinuxPlatform class sets the environment variables the first time sclang inits. Recompiling the class library does not unset the variables -- that's the behavior I observed.

So I think this PR is ready to go.

jamshark70:
Actually... it shows "1 requested change" but as I scan through the conversation, I don't see what the change is, and github doesn't provide a link directly to the change comment. I can re-request a review or dismiss, but not jump to the prior discussion.

So... is there anything pending? If not, then I think merge it. But, if Jordan's requested change has gotten lost in the shuffle... at least I don't see what's left to do.

dyfer:
Thanks for looking back at this @jamshark70 

@JordanHendersonMusic indeed it seems that your original change request is still here - are you satisfied with the state of this PR? The conversation suggests that your questions have been addressed, AFAICT.

JordanHendersonMusic (approved):
I had a question about the wording, but I've looked at it again and think it looks fine.

PR #6257, opened 2024-04-17
Fix a segfault in the GC

### LINKED ISSUE #6234: Segmentation Fault Crash with Specific Large Array Sizes
Ok, so I think the property-based tests found something interesting.

It appears to be triggered by specific array sizes (in the context of the tests, that is, a few arrays are produced, 10 or so). Large Arrays can be used to represent sound files, so I think it's not an extreme case in principle.

The interpreter crashes with a segmentation fault when handling arrays of certain sizes. Specifically, an array size of around 1,850,426 elements consistently triggers this crash. Interestingly, the issue does not occur with all large arrays, as some larger arrays do not cause the interpreter to crash.

Initialize a DoubleArray with approximately 1,850,426 elements. Example data includes [0.0062277317047119, 0.76690030097961, -0.56259202957153, ...] (Array size: 1,850,426).
      
Observe that the interpreter crashes with a segmentation fault (Exit code: 11).

```
Test sucessful with case: DoubleArray[0.0062277317047119, 0.76690030097961, -0.56259202957153, -0.25090932846069, -0.074282646179199, 0.41903877258301, 0.38275265693665, 0.5122504234314, -0.21230411529541, 0.42823481559753, 0.72746348381042, 0.86688160896301, 0.88132953643799, 0.51663398742676, -0.44769358634949, -0.89134740829468, -0.6724100112915, 0.79992604255676, -0.49154472351074, -0.38824009895325, 0.61342024803162, 0.73421216011047, -0.56524085998535, 0.60288572311401, -0.30279064178467, 0.0027897357940674, -0.023514747619629...etc...
====> Array size
1850426
Interpreter has crashed or stopped forcefully. [Exit code: 11]
    
    

Test sucessful with case: [0, 1, 0, 0, 0, 1, 1, 0, 0, 2, 0, 0, 1, 1, 1, 0, 1, 0, 0, 2, 0, 2, 2, 0, 0, 1, 1, 2, 2, 1, 1, 2, 0, 0, 2, 2, 1, 1, 1, 2, 2, 0, 0, 0, 1, 0, 0, 1, 1, 1, 1, 0, 2, 0, 1, 1, 0, 1, 0, 1, 1, 2, 0, 0, 0, 2, 0, 0, 2, 0, 0, 0, 1, 2, 2, 0, 1, 0, 0, 1, 1, 1, 2, 2, 1, 2, 2, 0, 2, 1, 1, 1, 1, 2, 0, 0, 2, 1, 1, 1, 1, 2, 2, 0, 0, 1, 0, 0, 0, 1, 2, 0, 0, 1, 0, 1, 2, 0, 1, 0, 1, 2, 1, 2, 1, 1, 1, 1, 0, 0, 2, 1, 0, 2, 1, 0, 2, 2, 0, 1, 1, 2, 2, 0, 1, 1, 2, 0, 1, 2, 1, 0, 1, 2, 0, 0, 2, 1, 0, 0, 2, 1, 0, 1, 0, 2, 0, 1, 2, 0, 0...etc...
1824713
Interpreter has crashed or stopped forcefully. [Exit code: 11]

 But even larger Arrays don't have this problem. THIS IS OK:        

Test sucessful with case: [0, 2, 0, 0, 2, 2, 1, 0, 0, 2, 1, 1, 0, 2, 2, 2, 0, 1, 0, 0, 2, 1, 2, 0, 0, 2, 1, 0, 0, 2, 0, 0, 1,  2, 2, 1, 0, 0, 2, 0, 0, 1, 2, 2, 1, 0, 1, 1, 1, 0, 0, 0, 2, 2, 0, 1, 2, 2, 2, 2, 0, 2, 2, 1, 2, 2, 1, 0, 0, 2, 1, 0, 2, 0, 0, 0, 2, 2, 0, 1, 2, 1, 0, 1, 0, 1, 1, 0, 1, 2, 0, 2, 2, 1, 2, 1, 1, 0, 0, 1, 2, 1, 0, 0, 1, 1, 1, 1, 0, 1, 2, 1, 0, 1, 2, 2, 1, 1, 0, 1, 0, 0, 0, 2, 2, 1, 1, 2, 2, 2, 2, 2, 0, 2, 0, 0, 2, 1, 0, 2, 1, 0, 2, 1, 2, 1, 0, 0, 1, 1, 0, 1, 0, 1, 0, 0, 2, 1, 0, 0, 1, 1, 0, 0, 1, 1, 1, 0, 2, 2, 1...etc...
2279156
[1, 0, 2, 0, 1, 1, 1, 2, 0, 1, 2, 0, 0, 2, 1, 2, 2, 2, 2, 0, 0, 2, 1, 0, 2, 2, 0, 0, 0, 2, 2, 2, 1, 1, 1, 0, 2, 2, 1, 1, 0, 2, 2, 0, 0, 2, 2, 1, 1, 2, 0, 1, 2, 2, 1, 0, 0, 2, 1, 2, 0, 1, 0, 1, 0, 2, 1, 1, 0, 2, 2, 2, 1, 1, 1, 1, 2, 1, 0, 1, 2, 0, 2, 1, 0, 1, 2, 2, 2, 2, 2, 2, 1, 2, 1, 2, 2, 1, 2, 2, 2, 2, 1, 1, 1, 2, 2, 2, 0, 0, 0, 2, 1, 2, 0, 2, 1, 1, 1, 2, 0, 2, 0, 1, 2, 1, 0, 2, 1, 2, 2, 2, 2, 0, 1, 2, 0, 0, 2, 0, 1, 0, 1, 0, 0, 1, 0, 1, 2, 2, 1, 1, 0, 1, 1, 0, 0, 0, 0, 0, 0, 2, 1, 2, 1, 0, 2, 0, 2, 1, 2...etc...
Test 9 with case: [1, 0, 2, 0, 1, 1, 1, 2, 0, 1, 2, 0, 0, 2, 1, 2, 2, 2, 2, 0, 0, 2, 1, 0, 2, 2, 0, 0, 0, 2, 2, 2, 1, 1, 1, 0, 2, 2, 1, 1, 0, 2, 2, 0, 0, 2, 2, 1, 1, 2, 0, 1, 2, 2, 1, 0, 0, 2, 1, 2, 0, 1, 0, 1, 0, 2, 1, 1, 0, 2, 2, 2, 1, 1, 1, 1, 2, 1, 0, 1, 2, 0, 2, 1, 0, 1, 2, 2, 2, 2, 2, 2, 1, 2, 1, 2, 2, 1, 2, 2, 2, 2, 1, 1, 1, 2, 2, 2, 0, 0, 0, 2, 1, 2, 0, 2, 1, 1, 1, 2, 0, 2, 0, 1, 2, 1, 0, 2, 1, 2, 2, 2, 2, 0, 1, 2, 0, 0, 2, 0, 1, 0, 1, 0, 0, 1, 0, 1, 2, 2, 1, 1, 0, 1, 1, 0, 0, 0, 0, 0, 0, 2, 1, 2, 1, 0, 2, 0, 2, 1, 2...etc...
```

    

 Notably, the crash occurs with numerical data: floats (64-bit "doubles" ) and arrays filled with simple numbers (Integers), indicating that the content of the array does not affect the crash's occurrence, just the size.

Affected Array Sizes: Crash: Array sizes around 1,850,426 and 1,824,713 elements.
No Crash: Larger arrays, such as one with 2,279,156 elements, do not trigger the crash.

Anything less than those values is consistently correct. Even larger Arrays have not been tested in the same way yet. The pattern is not clear yet.

As I understand, large objects like this have a special treatment in the language (LargeObjSizeClass).

SYSTEM DETAILS:
------------------

GNU/Linux - Fedora 39 6.7.6-200 Intel x86_64
Custom compilation latest develop branch with native flags and -DNATIVE=ON

JordanHendersonMusic:
## Purpose and Motivation

Fixes #6234

On my machine, the following code consistently crashes in release mode, but not debug.

```supercollider
b = Pwhite.new(0, 10, inf).asStream;
Array.fill(1810000, {b.next});
```
I was able to trace this down to the member PyrGC::mNumToScan being negative by using print statements at the start and end of all the methods of PyrGC. 
 
For some reason that I cannot understand, this static_cast fixes the problem and avoids  setting the variable to a negative value, despite the fact that I cannot see why it would have an effect. I've even tested this with a print statement that clearly shows that the signed and unsigned right shifted values are the same?? Originally, I thought it might be right shifting a negative number, which is implementation defined, but the if statement already checks for this. I have no clue why it works, yet I have ran it multiple times, and as far as I can tell, completely fixes the issue. 

My machine: 
linux 64bit, manjaro, latest dev branch, compling with GCC13.2.1, -DNATIVE=(ON and OFF) -DINSTALL_HELP=ON -DSC_EL=OFF -DSYSTEM_BOOST=ON 
## Types of changes

Added a single static_cast that should effect nothing.

## To-do list

- [x] Code is tested
- [x] All tests are passing

---

More info:

The following also crash.
```c++
if(mNumToScan < 0) {
    // never gets printed.
    std::cout << "Before  " << mNumToScan << std::endl;
}

mNumToScan += mNumToScan >> 3;

if(mNumToScan < 0) {
    // never gets printed.
    std::cout << "After  " << mNumToScan << std::endl;
}
```
```c++
 mNumToScan += std::abs(mNumToScan) >> 3;
```
```c++
mNumToScan += std::abs(mNumToScan >> 3);
```
The casted value is always equal to the non-casted
```c++
const int32 copy = mNumToScan;
const int32 cast = static_cast<uint32>(mNumToScan);
if(copy != cast) {
    // never gets printed.
    std::cout << "ERRR  " << mNumToScan << std::endl;
}
```

### DIFF
--- a/lang/LangSource/GC.cpp
+++ b/lang/LangSource/GC.cpp
@@ -29,7 +29,6 @@
 
 #define PAUSETIMES 0
 
-
 double pauseBeginTime = 0.;
 double totalPauseTime = 0.;
 double maxPauseTime = 0.;
@@ -44,7 +43,7 @@ int maxPauseNumToScan = 0;
 int maxPauseSlotsScanned = 0;
 int checkStackScans = 0;
 int checkFlips = 0;
-int checkNumToScan = 0;
+uint64 checkNumToScan = 0;
 int checkScans = 0;
 int checkPartialScans = 0;
 int checkSlotsScanned = 0;
@@ -500,9 +499,7 @@ HOT void PyrGC::DoPartialScan(int32 inObjSize) {
     mNumPartialScans++;
     if (remain <= 0) {
         mPartialScanObj = nullptr;
-        mNumToScan -= 4;
-        if (mNumToScan < 0)
-            mNumToScan = 0;
+        mNumToScan -= (mNumToScan <= 4) ? mNumToScan : 4;
         return;
     }
     int32 numtoscan = sc_min(remain, mNumToScan);
@@ -510,13 +507,11 @@ HOT void PyrGC::DoPartialScan(int32 inObjSize) {
 
     if (numtoscan == remain) {
         mPartialScanObj = nullptr;
-        mNumToScan -= numtoscan + 4;
+        mNumToScan -= (mNumToScan <= numtoscan + 4) ? mNumToScan : numtoscan + 4;
     } else {
         mPartialScanSlot += numtoscan;
-        mNumToScan -= numtoscan;
+        mNumToScan -= (mNumToScan <= numtoscan) ? mNumToScan : numtoscan;
     }
-    if (mNumToScan < 0)
-        mNumToScan = 0;
     // post("partial %5d xx %4d %2d %s\n", mScans, mNumToScan, mNumGrey);
     // post("partial %5d %2d %4d %2d %s\n", mScans, i, mNumToScan, mNumGrey, slotRawSymbol(&obj->classptr->name)->name);
 }
@@ -552,13 +547,11 @@ HOT bool PyrGC::ScanOneObj() {
         DoPartialScan(size);
     } else if (size > 0) {
         ScanSlots(obj->slots, size);
-        mNumToScan -= 1L << obj->obj_sizeclass;
-        if (mNumToScan < 0)
-            mNumToScan = 0;
+        const auto sub = 1L << obj->obj_sizeclass;
+        mNumToScan -= (mNumToScan <= sub) ? mNumToScan : sub;
     } else {
-        mNumToScan -= 1L << obj->obj_sizeclass;
-        if (mNumToScan < 0)
-            mNumToScan = 0;
+        const auto sub = 1L << obj->obj_sizeclass;
+        mNumToScan -= (mNumToScan <= sub) ? mNumToScan : sub;
     }
     return true;
 }
@@ -647,7 +640,7 @@ void PyrGC::FullCollection() {
     SweepBigObjects();
 }
 
-void PyrGC::Collect(int32 inNumToScan) {
+void PyrGC::Collect(uint64 inNumToScan) {
     mNumToScan = sc_max(mNumToScan, inNumToScan);
     Collect(); // collect space
 }
@@ -660,43 +653,45 @@ HOT void PyrGC::Collect() {
 #ifdef GC_SANITYCHECK
     SanityCheck();
 #endif
+    if (mNumToScan == 0) {
+        mUncollectedAllocations = 0;
+        return;
+    }
 
-    if (mNumToScan > 0) {
-        // post("->Collect  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
-        // DumpInfo();
-        mNumToScan += static_cast<uint32>(mNumToScan) >> 3;
+    // post("->Collect  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
+    // DumpInfo();
+    mNumToScan += mNumToScan >> 3;
 
-        // post("->Collect2  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
-        // mCurSet = 0;
-        while (mNumToScan > 0) {
-            while (mNumToScan > 0 && (mNumGrey > 0 || mPartialScanObj)) {
-                if (mPartialScanObj) {
-                    DoPartialScan(ScanSize(mPartialScanObj));
-                } else {
-                    if (!ScanOneObj())
-                        break;
-                }
-            }
-            if (mNumGrey == 0 && mPartialScanObj == nullptr) {
-                if (!stackScanned) {
-                    stackScanned = true;
-                    mStackScans++;
-                    ScanStack();
-                    ScanFrames();
-                }
-                if (mNumGrey == 0 && mPartialScanObj == nullptr && stackScanned) {
-                    Flip();
+    // post("->Collect2  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
+    // mCurSet = 0;
+    while (mNumToScan > 0) {
+        while (mNumToScan > 0 && (mNumGrey > 0 || mPartialScanObj)) {
+            if (mPartialScanObj) {
+                DoPartialScan(ScanSize(mPartialScanObj));
+            } else {
+                if (!ScanOneObj())
                     break;
-                }
             }
         }
-        // post("<-Collect  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
-        // DumpInfo();
-        // post("size9:\n");
-        // TraceAnyPathToObjsOfSize(9);
-        // post("greys:\n");
-        // TraceAnyPathToAllGrey();
+        if (mNumGrey == 0 && mPartialScanObj == nullptr) {
+            if (!stackScanned) {
+                stackScanned = true;
+                mStackScans++;
+                ScanStack();
+                ScanFrames();
+            }
+            if (mNumGrey == 0 && mPartialScanObj == nullptr && stackScanned) {
+                Flip();
+                break;
+            }
+        }
     }
+    // post("<-Collect  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
+    // DumpInfo();
+    // post("size9:\n");
+    // TraceAnyPathToObjsOfSize(9);
+    // post("greys:\n");
+    // TraceAnyPathToAllGrey();
     // post("mNumToScan %d\n", mNumToScan);
 
     mUncollectedAllocations = 0;
--- a/lang/LangSource/GC.cpp
+++ b/lang/LangSource/GC.cpp
@@ -664,7 +664,7 @@ HOT void PyrGC::Collect() {
     if (mNumToScan > 0) {
         // post("->Collect  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
         // DumpInfo();
-        mNumToScan += mNumToScan >> 3;
+        mNumToScan += static_cast<uint32>(mNumToScan) >> 3;
 
         // post("->Collect2  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
         // mCurSet = 0;
--- a/lang/LangSource/GC.h
+++ b/lang/LangSource/GC.h
@@ -126,7 +126,7 @@ public:
     // users should not call anything below.
 
     void Collect();
-    void Collect(int32 inNumToScan);
+    void Collect(uint64 inNumToScan);
     void LazyCollect() {
         if (mUncollectedAllocations > kLazyCollectThreshold)
             Collect();
@@ -197,7 +197,7 @@ private:
     PyrObjectHdr mGrey;
 
     int32 mPartialScanSlot;
-    int32 mNumToScan;
+    uint64 mNumToScan;
     int32 mNumGrey;
 
     int32 mFlips, mCollects, mAllocTotal, mScans, mNumAllocs, mStackScans, mNumPartialScans, mSlotsScanned,


JordanHendersonMusic:
I was wondering if, @smoge, you might mind testing this as you also had segfaults?

muellmusik:
I think we need to figure out why mNumToScan is negative here in the first place?

JordanHendersonMusic:
If I do this...
```c++
if(mNumToScan < 0) {
  std::cout << "Before  " << mNumToScan << std::endl;
}

mNumToScan += mNumToScan >> 3;

if(mNumToScan < 0) {
  std::cout << "After  " << mNumToScan << std::endl;
}
```
It never prints anything.

This is very confusing.

I considered that maybe some other thread was touching it at the same time, but that doesn't explain why the cast fixes it.

mtmccrea:
Are we sure it every goes negative? 
Based on the flow control, it shouldn't ever reach this bit shift increment if it's negative...
```    
if (mNumToScan > 0) {
        mNumToScan += mNumToScan >> 3;
```
and if I'm understanding the bit shift with addition, it's incrementing mNumToScan by 1/8th of it's value, so this would also not take it negative.

casting it may prevent some kind of overflow that happens without the cast?

mtmccrea:
You might try testing the result of the increment with and without the cast 
```
const int32 orig = mNumToScan;
orig += mNumToScan >> 3;
mNumToScan += static_cast<uint32>(mNumToScan) >> 3;
if(orig != mNumToScan) {
    // never gets printed.
    std::cout << "ERRR  " << mNumToScan << std::endl;
}
```
(I'd test it myself but I just updated XCode and it derailed the build process)

JordanHendersonMusic:
@mtmccrea I've tried that and they are the same, hence why this is very odd.

However, I've just been comparing the assembly generated for with and without the cast in Clion (its like having godbolt in an ide, super cool!), and they are *very* different. 

Here is with the cast.

It is as expected, the assembly for that line is together.



... without the cast however, the assembly is interleaved with code from across the function and with an inlined function





This line corresponds to some code from the ScanSlots method.


There are also a few lines of assembly that clion can't figure out where they come from.


---


Now, I *think* the issue is in these two lines...

```
mNumToScan += mNumToScan >> 3;
```
```
while (mNumToScan > 0) {
```
... and the fact the compile can assume that mNumToScan is positive (because it is undefined behaviour to right shift a negative) and that the condition of the while loop on first occurrence is always true (because the increment cannot overflow as signed overflow is undefined). This is a little speculative though.

muellmusik:
> casting it may prevent some kind of overflow that happens without the cast?

It's possible that as we cast a negative number to an unsigned type, we get a positive value which by luck does not make the GC overflow. https://timsong-cpp.github.io/cppwp/n4659/conv.integral

My understanding (please correct if needed) is that on Debug builds memory allocations my be larger than needed, which may explain why that doesn't crash. Again just by luck.

> It never prints anything (and crashes).

I've tried similar tests to figure out where it goes negative. Similar issue.

muellmusik:
> (I'd test it myself but I just updated XCode and it derailed the build process)

Sympathies. I've also been dealing with this. There's a couple of issues and a PR about this. This works for me:

```
cmake -G Xcode -DCMAKE_CXX_FLAGS="--std=c++17 -D_LIBCPP_ENABLE_CXX17_REMOVED_FEATURES" -DCMAKE_PREFIX_PATH=`brew --prefix qt5` ..
```

JordanHendersonMusic:
Should say, the way I found this was by placing code like this...
```
if(mNumToScan < 0) { std::cout << "A" << std::endl; std::raise(SIGABRT); }
```
At the top and bottom of everything function, then gradually move them towards the middle to find the issue. This was the only line that touched the variable.

muellmusik:
Okay, in a debug build I uncommented the posts:

```
        post("->Collect  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
        // DumpInfo();
        mNumToScan += mNumToScan >> 3;
        post("->Collect2  ns %d  ng %d  s %d\n", mNumToScan, mNumGrey, mScans);
```

Here's the output.

```
->Collect  ns 300  ng 78  s 1433563
->Collect2  ns 337  ng 78  s 1433563
->Collect  ns 636  ng 85  s 1433602
->Collect2  ns 715  ng 85  s 1433602
-> a Routine
->Collect  ns 332  ng 88  s 1433672
->Collect2  ns 373  ng 88  s 1433672
->Collect  ns 2097168  ng 87  s 1433675
->Collect2  ns 2359314  ng 87  s 1433675
->Collect  ns 2097156  ng 2  s 1436420
->Collect2  ns 2359300  ng 2  s 1436420
->Collect  ns 576  ng 3  s 1436422
->Collect2  ns 648  ng 3  s 1436422
->Collect  ns 1925839492  ng 3  s 1436429
->Collect2  ns -2128397868  ng 3  s 1436429
-> [4, 10, 2, 4, 8, 4, 5, 3, 4, 8, ...etc...

```

1925839492 + (1925839492 >> 3) is -2128397868

muellmusik:
So the question is if ```1925839492``` is a correct value. If so we've just exceeded int32 range. I can try and look through that but would like a fresh brain. It's been a while since I've dug into GC stuff. If someone else wants to have a go that would be okay too.
 
Switching ```mNumToScan``` to int64 prevents the crash, but we need to make sure it's not hiding a problem, and makes sense generally.

muellmusik:
Okay added this post in PyrGC:NewFrame
```
    int32 credit = 1L << sizeclass;
    mAllocTotal += credit;
    mNumAllocs++;
    post("->NewFrame ns %d  c %d\n", mNumToScan, credit);
    mNumToScan += credit;
```
It builds steadily towards 1925839492 and then borks. So I'm guessing it's just the massive number of new frames this entails. 

**Please do** check my work here, but I'm guessing this is _not_ a bug, but just hitting the limits of the system.

mtmccrea:
Thanks for the sleuthing here @muellmusik and @JordanHendersonMusic!

To clarify, since it keeps coming up, 
>Are we sure it ever goes negative?

the issue seems _not_ to be that `mNumToScan` ever goes negative, but for the overflow once in the loop.

If so, and after confirming this as Scott says, the question becomes what kind of side effects might show up when bumping up the resolution to `size_t`? @Spacechild1's [related post](https://github.com/supercollider/supercollider/issues/6234#issuecomment-1996353002).

Is this all it takes to support the original breaking example or will other vars in the neighborhood have to also have their resolution increased?
https://github.com/supercollider/supercollider/blob/28a0b12216fb97e4487373b2cf00a5ec28850f1f/lang/LangSource/GC.h#L199-L204

These are the kinds of changes that, once merged, should spend a good bit of time circulating in the dev build...

It's also valid to consider whether or not to support operating on "37-minute long" arrays in this way :)

JordanHendersonMusic:
I've updated the pr. It now has `mNumToSet` as a uint64, which fixes the bug.

muellmusik:
> the issue seems _not_ to be that `mNumToScan` ever goes negative, but for the overflow once in the loop.

Correct. Wouldn't hurt for someone to confirm.

> These are the kinds of changes that, once merged, should spend a good bit of time circulating in the dev build...

Absolutely. And many pairs of eyes reviewing. This will touch every use.

> It's also valid to consider whether or not to support operating on "37-minute long" arrays in this way :)

Yes. I personally don't think switching to a 64 bit scan count would do any harm, and might be a reasonable improvement at this stage. But if the analysis is correct, this is not a bug, but just hitting the limits of the design. It's important to note that SC is not designed for massive number crunching, but real-time safety. Performance is usually optimised for normal use cases.

muellmusik:
> seems to be the result of _**not freeing the temporary function objects**_ each time they are invoked, preferring, instead, to wait until the line has been executed, this explains why filling with a float is fine. Interestingly, this means this bug has nothing to do with arrays and could occur elsewhere.

That matches my reading as well. So again not a bug per se, but hitting the limits of what the design can handle? (Unless I'm misunderstanding you.)





> I think the ideal solution would be if supercollider had a way to free shortlived objects that can be called during the loop that fills the array. Does anyone know if this is possible?

I tried to look through what's happening, although came up against the ```// special byte codes inserted by compiler for this method``` in do. Not sure where that happens?

But again SC is usually optimised for real-time safety, and good performance in common usage. If it's not scanning the temp objects until the end, it's possible that that was found to be more efficient in most cases. Some of those choices were made a long time ago, and may be worth checking again on modern systems. But let's tread carefully as this is really core stuff. :-)

muellmusik:
> I've updated the pr.

@JordanHendersonMusic as this has taken a different direction, would you mind updating the title and description to match the current content, or perhaps just close and make a clean PR? If you could include as much detail as possible about what's been done would be fab! Sorry to be a pain, but if we go ahead with this we need to be very careful and clear, and leave an easy to follow discussion trail for future devs if they need to revisit.

smoge:
> It's not the length that causes an issue, but not freeing the temporary objects needed in initialisation.

If I understood correctly, the GC is not good with large iterative operations with many short-lived objects, rapid allocations and short lifespans. (It would fit another case described in the forum with a similar kind of behavior)

I'm still studying the GC mode, (maybe that' s a good chance to read the original paper, other texts, and write comments when it makes sense?). 

It seems simply a limitation of the model with certain situation. And not related to just large arrays (there are many cases that use large arrays to represent sound-files, it' s not an exotic usage in SC code). 

Maybe would be possible try to do what you said, adapting it to those situations.  But changes like that would be experimental for a while until considered safe?

JordanHendersonMusic:
> perhaps just close and make a clean PR @muellmusik 

Ah, I had an idea last night and I think it warrants some discussion. The change to this pr was a little hasty and I should have slept on my idea.

Making `mNumToSet` a u64 works by increasing the about of memory the system can hold, it doesn't make the system use less memory. This is obviously a perfectly fine trade-off if it results in better performance. So I thought I'd test this.

So I inserted a `g->gc->Collect` at the bottom of `executeMethod` in PyrMessage (once everything created was on the stack and visible to the collector). This will: free all of the temporary objects after they are created and no longer used; but also scan after every method call.

Here are the benchmarks:

## with u64
```
b = Pwhite.new(0, 10, inf).asStream; bench { Array.fill(10810000, {b.next}); } 
time to run: 2.912465206 seconds. 
time to run: 2.472034116 seconds.
time to run: 2.46097923 seconds.
```
The first call is greater the rest, perhaps because of the malloc calls to increase the owned memory.

## calling Collect more frequently 
```
b = Pwhite.new(0, 10, inf).asStream; bench { Array.fill(10810000, {b.next}); } 
time to run: 2.259202985 seconds.
time to run: 2.147686644 seconds.
time to run: 2.232509631 seconds.
time to run: 2.121764891 seconds.
```

However, this means the collection phase is always run, and after running benchmarks that don't allocate, shows a performance regression.

The compromise I came up with is to run this only when the `mNumToSet` value is large. 
The following method is called at the end of `executeMethod`.
```
void PyrGC::CollectIfApproachingIntLimit() {
    if (mNumToScan >= 1'000'000)
        Collect();
}
```
The benchmarks for the large allocation show the same speed up as calling collect every time.

Here are some benchmarks that don't allocate — TLDR, they are almost identical.

## existing implementation
```
bench { var a = (\v: 0); while { a[\v] < 1000000} { a[\v] = a[\v].pow(1) + 0.1 } }
time to run: 0.897474206 seconds.
time to run: 0.957676498 seconds.
time to run: 0.830193156 seconds.
time to run: 0.860305335 seconds.
time to run: 1.02162904 seconds.
```


## calling colllect if large `mNumToScan`
```
bench { var a = (\v: 0); while { a[\v] < 1000000} { a[\v] = a[\v].pow(1) + 0.1 } }
time to run: 0.847071381 seconds.
time to run: 0.840482856 seconds.
time to run: 0.848130279 seconds.
time to run: 0.872771463 seconds.
```

As you can see, this doesn't occur any performance regression (actually shows some improvements but I put this down to a small sample size).

The `1'000'000` cut off I chose because it is big, and won't be hit unless actually making a large number of objects.

Overall, I think this is a better garbage collection strategy and actually addresses the underlying issue, reduces memory usage, and doesn't incur a loss of performance.   

If there aren't any immediate objections to this approach, I'll close this PR and open a new one with all this information in?

Spacechild1:
> void PyrGC::CollectIfApproachingIntLimit() {

this looks like a good idea to me.

>  Instead, 1925839492 seems to be the result of not freeing the temporary function objects each time they are invoked, preferring, instead, to wait until the line has been executed, this explains why filling with a float is fine

As you noted, sclang runs the GC every time it allocates, which *should* already be what we need. `PyrGC::CollectIfApproachingIntLimit()` is a (valid) escape hatch, but I does not address the underlying issue: who is retaining all these function objects?

smoge:
@Spacechild1 @JordanHendersonMusic 

Why the crash happens only with a certain range, but not with any larger objects? Why 1,000,000 exactly?

Do you see any relation to those GUI crashes? Unrelated bug? https://scsynth.org/t/help-with-crash-report/9287/8

JordanHendersonMusic:
> sclang runs the GC every time it allocates ... 
> If you allocate lots of temporary objects in a short timespan, it can easily build up. @Spacechild1 


When making a new frame in `exectueMethod` collect is only called if `needsHeapContext` is true.
```
frame = (PyrFrame*)g->gc->NewFrame(methraw->frameSize, 0, obj_slot, methraw->needsHeapContext);
```
`needsHeapContext` is set over in PyrParseNode.cpp, and is quite complex code to follow (with no comments), but I think the general reasoning is that if the function (`{b.next}` in this case) doesn't need to allocate anything to run (I don't know how closures work in SC and I'm only familiar with upvalues from craftinginterpreters [which is an amazing book]) then a GC scan can be avoided. 

This seems to overlook the fact that the function itself requires an allocation, no matter how small it is, leading to the bug in question.

`NewFrame` is only ever called by passing `methraw->needsHeapContext` to control the scan. 

While there might be a better solution to make the GC collect more frequently, I don't feel confident enough in my understanding of `PyrParseNode` to attempt this. 

Besides, it would likely have the performance characteristics of running `Collect` after executing every method, meaning setting a threshold would still be required to preserve performance.

---

> within a certain range, not with larger objects? @smoge 

I didn't test that because I didn't observe that behaviour, it just reached a limit then stopped. If it still exists after this fix, I'll have another look (time-money permitting). 


> Why 1,000,000 exactly?

Its a large number. This value doesn't refer to the amount of memory allocated in the system, but the amount that needs to be scanned (flagged as white/grey/black) before the gc sweep (might) begin.  

The value could be larger (you could analytically figure out the value where the overflow happens), but, as can be seen from the benchmarks, this means lots of supercollider's memory will be taken up with temporary objects. Freeing before that threshold is reached means the system can avoid calling malloc and recycle its memory. 

If there is an objection to this, perhaps an alternative heuristic to compute the value could be proposed, but I doubt an analytical value could be found?

> Do you see any relation to those GUI crashes?

No I didn't test this. I think its really important to make GC bugs as minimal as possible because tracing all this was quite time consuming, having to deal with more code would make things harder. However, I don't ***think*** they are related because it would require a truly large number of temporary objects with no allocations, and therefore scans, in between to trigger this behaviour. 

---

I'm going to close this and open another PR containing this information and the history of the bug.

Spacechild1:
> This seems to overlook the fact that the function itself requires an allocation, no matter how small it is, leading to the bug in question.

Ah, that would make a lot of sense! Thanks for looking into this!

> craftinginterpreters [which is an amazing book]

Yes, that's really a great book! Tip: the Lua source code is actually pretty accessible, for anyone interested in the implementation of dynamic scripting languages.

PR #3384, opened 2018-01-04
Refactor new object allocation

muellmusik:
Not for merging, just a quick and dirty PR for review and discussion.

This starts to look at the improvements mooted in #1634. Approach is as follows:

- The GC keeps a count of unreachable objects.
  - Each new allocation increments the count
  - Calls to Allocate() with ```inRunCollection = true``` will print an error and terminate if the count is not 0.
  - use a count as it provides the flexibility to still run with ```inRunCollection = false``` if desired.
  - Addresses problem 1. in #1634 
- A new smart pointer type ```NewPyrObjectPtr``` handles newly created objects
  - It is consumed by using it to make its object reachable, which decrements the GC count
  - Its destructor prints an error and terminates if its object has not been made reachable.
  - Addresses problem 5. in #1634 
- Two new functions, ```SetNewObjectOnStack``` and ```SetNewObjectInObject``` work with the smart pointers
  - the latter calls GCWriteNew() for you. A NewPyrObjectPtr either points at a white object or has been consumed, so this should be safe. (The GC should terminate if we do anything that's could make it not white)
  - Addresses problems 2. and 4. in #1634 

I wanted to make a SetNewObjectInVarSizedObject func as well, which would correctly update the size (address problem 3.), but this needs a look through to see what cases can be generalised.

```basicNew``` and ```SCLanguageClient::setCmdLine``` are altered to use the new approach, so it can get a *lot* of testing. A dummy primitive can be used to test the mistake checking with ```Main. allocateWhileUnreachable```.

I'm sure this can be improved in many ways, but at least provides an approach for discussion. (e.g. it seems like it should be possible and preferable to have versions of ```instantiateObject``` and ```newPyr...``` that return ```NewPyrObjectPtrs``` by value, but I can't get the assignment operator to work... :-/  Stupid C++...)

Once we've agreed an approach we can convert existing primitives bit by bit. Once that's done remove the old code and hopefully these classes of error can be entirely prevented.

PR #6261, opened 2024-04-21
Sclang: PyrGC: Fix GC segfault caused by uncollected temporary function objects by placing an upper limit on un-scanned objects. 

### LINKED ISSUE #6234: Segmentation Fault Crash with Specific Large Array Sizes

JordanHendersonMusic:
## Purpose and Motivation

When there are a large number of temporary objects mNumToScan can overflow leading to undefined behaviour (from the overflow of a signed int) and the implementation defined behaviour of right shifting a negative value, creating a GC bug that only appears on some systems, and incorrect, potentially leaking, but non-crashing behaviour on others.

This code cause a crash on some systems.
```supercollider
b = Pwhite.new(0, 10, inf).asStream;
Array.fill(1810000, {b.next});`
```

This commit places an upper limit on the number of objects to scan (5'000'000). Note, SC uses an incremental allocator, meaning this isn't a limit on the amount of memory, but amount of unchecked memory.

While all allocations should cause a collection, when PyrParseNode.cpp creates a PyrMethod, sometimes it sets  'needsHeapContext' to false which disables collection. This is (presumably) done for optimization. What this fails to consider is that frames in supercollider are themselves heap allocated, meaning, no matter how small, they increase mNumToScan, leading to the above bug.

While a change to PyrParseNode.cpp could be possible, the code is without comments and very complex.

Setting a limit on mNumToScan has a potential minuscule impact on performance, as the additional scan happens only rarely, and in some cases, actually improves performance as the GC is allowed to recycle the memory, reducing the number of calls to malloc — this is only noticeable in tight loops with many iterations that create functions.

---

Fixes #6234, also see #6257 for history.

## Benchmarks

Test involving numerous temporary functions (that doesn't crash with the old behaviour).
```supercollider
b = Pwhite.new(0, 10, inf).asStream; 
100.collect{ 
   { 
      Array.fill(100000, {b.next})
   }.bench(false) 
}.mean
```
Old: 0.02174366042
New: 0.01934624595
Delta: +12%

A test without temporary functions that should be unaffected by this PR.
```supercollider
100.collect{ 
   {   
      var a = (\v: 0);   
      while { a[\v] < 100000 } {   
         a[\v] = a[\v].pow(1) + 0.1   
      }     
   }.bench(false)  
}.mean
```
Old: 0.08627185895
New: 0.0867351175
Delta: -0.5% 

## Types of changes

- Bug fix
- Change in GC strategy, observed performance gains, but possible performance regressions although I haven't observed any significant losses.

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/GC.h
+++ b/lang/LangSource/GC.h
@@ -30,6 +30,7 @@ Based on Wilson and Johnstone's real time collector and the Baker treadmill.
 #include "VMGlobals.h"
 #include "AdvancingAllocPool.h"
 #include "function_attributes.h"
+#include <cstdint>
 
 void DumpSimpleBackTrace(VMGlobals* g);
 
@@ -37,7 +38,7 @@ const int kMaxPoolSet = 7;
 const int kNumGCSizeClasses = 28;
 const int kFinalizerSet = kNumGCSizeClasses;
 const int kNumGCSets = kNumGCSizeClasses + 1;
-const int kScanThreshold = 256;
+const std::int64_t kScanThreshold = 256LL;
 
 
 class GCSet {
@@ -197,7 +198,7 @@ private:
     PyrObjectHdr mGrey;
 
     int32 mPartialScanSlot;
-    int32 mNumToScan;
+    std::int64_t mNumToScan;
     int32 mNumGrey;
 
     int32 mFlips, mCollects, mAllocTotal, mScans, mNumAllocs, mStackScans, mNumPartialScans, mSlotsScanned,
@@ -286,7 +287,7 @@ inline void PyrGC::ToGrey(PyrObjectHdr* obj) {
     /* set grey list pointer to obj */
     obj->gc_color = mGreyColor;
     mNumGrey++;
-    mNumToScan += 1L << obj->obj_sizeclass;
+    mNumToScan += 1LL << obj->obj_sizeclass;
 }
 
 inline void PyrGC::ToGrey2(PyrObjectHdr* obj) {
@@ -303,9 +304,9 @@ inline void PyrGC::ToGrey2(PyrObjectHdr* obj) {
 }
 
 inline PyrObject* PyrGC::Allocate(size_t inNumBytes, int32 sizeclass, bool inRunCollection) {
-    if (inRunCollection && mNumToScan >= kScanThreshold)
+    if (inRunCollection && mNumToScan >= kScanThreshold) {
         Collect();
-    else {
+    } else {
         if (inRunCollection)
             mUncollectedAllocations = 0;
         else
--- a/testsuite/sclang/sclang_gc_crash_1.scd
+++ b/testsuite/sclang/sclang_gc_crash_1.scd
@@ -1,0 +1,6 @@
+// This first value caused a crash on some linux systems. 
+[ 1810000, 3620000, 7240000].do { |n|
+    var b = Pwhite().asStream;
+    Array.fill(n, { b.next });
+};
+0.exit;


JordanHendersonMusic:
The two force pushes were for the linter (which I can't run on my computer) and replacing 'or' with '||' because windows.

smoge:
Have you run the tests available in the project? I'll give it a try.

JordanHendersonMusic:
While benchmarking this, I tried setting the threshold very low (500). This causes a crash because PyrGC's constructor calls its own member functions before it is initialised.

JordanHendersonMusic:
> Have you run the tests available in the project? I'll give it a try.

Which tests are you referring to?

smoge:
> > Have you run the tests available in the project? I'll give it a try.
> 
> Which tests are you referring to?

In general, just to double check

JordanHendersonMusic:
> In general, just to double check

Yes, no crashes in the unit testing, nor with your original code
```
a = Pwhite.new(1802903, 1924888, inf).asStream;
b = Pwhite.new(0, 10, inf).asStream;

10 do: {
  Array.fill(a.next.round.asInteger.postln, {b.next})
}
```

smoge:
well done, my man. that was sweet

smoge:
In any case, the Unit tests we have are very weak inference, not good tests. Ideally some people should just use it for a while. I can do that btw.

muellmusik:
I think we need to test across a variety of use cases for performance and stability. Large sclang data arrays are not typical usage, and mostly we're prioritising realtime safety. Not that I see any obvious issues with this, but as it's at the core of everything...

Would be good to add more GC specific stuff to the test suites. Should think about how best to do that...

smoge:
> I think we need to test across a variety of use cases for performance and stability. Large sclang data arrays are not typical usage, and mostly we're prioritising realtime safety. Not that I see any obvious issues with this, but as it's at the core of everything...
> 
> Would be good to add more GC specific stuff to the test suites. Should think about how best to do that...

I was thinking about that, the unit tests in the project are not adequate at all. They use weak inference and hand-made tests.

The best options, without talking about working on another testing framework, would just push to the develop branch as ask people to test it wildely.

smoge:
> Wonder if it is possible to add some bookkeeping to the internal allocator (s) so they can check their memory is being addressed properly?

That's a good idea, but I believe it's pretty involved. You would need to introduce some "metadata", and some pattern of bytes before and after each allocation. 2) keep track of pointers to each allocation. 3) record each allocation and deallocation.

In a way, there are already tools to help with some of that: real-time tracing isn't already done by toolings like Valgrind. 

Property-based (and even "fuzz testing") would continue to be a tool to reveal problems, just like the test that triggered this fix. (Unfortunately, it is not even discussed right now in the community, maybe some CI trauma? Or something else?)

muellmusik:
> Wonder if it is possible to add some bookkeeping to the internal allocater(s) so they can check their memory is being addressed properly?

Some discussion and unnerved work here: #3384

smoge:
> > Wonder if it is possible to add some bookkeeping to the internal allocater(s) so they can check their memory is being addressed properly?
> 
> Some discussion and unnerved work here: #3384

There is a history of ideas and documented thought behind this that I don't have the time to study now. Sorry if I said uninformed noise. Eventually I will catch up.

muellmusik:
> @muellmusik There is a history of ideas and documented thought behind this that I don't have the time to study now. IF THAT'S THE CASE HERE, sorry if I made an uninformed noise. Eventually, I will catch up and check that.

What I linked to was an attempt to improve GC safety through an enhanced implementation. Not a test, so something a bit different.

muellmusik:
There is also the GC_SANITYCHECK flag. But it makes things *very* slow in tests, so probably not great for testing large things like this. 

You can trigger the sanity check ad hoc with the ```gcSanity``` message.

smoge:
> > @muellmusik There is a history of ideas and documented thought behind this that I don't have the time to study now. IF THAT'S THE CASE HERE, sorry if I made an uninformed noise. Eventually, I will catch up and check that.
> 
> What I linked to was an attempt to improve GC safety through an enhanced implementation. Not a test, so something a bit different.

Has anyone dared to suggest a major reform of GC taking into account modern C++ and the computers and procedures unthinkable at the time JMC was doing its first granular real-time syntheses? Maybe acting precisely where it seems to perform the epic failures?  Maybe only JMC or Tim would have the expertise to such a thing, I don't know.

JordanHendersonMusic:
> What I linked to was an attempt to improve GC safety through an enhanced implementation. Not a test, so something a bit different.

That's pretty cool! I was thinking about the same thing recently and was going to try having a small stack inside the allocator class that all allocations are added to and cleared from periodically. This way, at least within a primitive scope which doesn't call other sc code, you guarantee the object exists.

It wouldn't have caught this bug though. 

I might try making the allocator (when a flag is enabled) keep a record of all the alive objects and their size, this way you can ask the allocator if the object ptr is valid, how many objects are alive, etc.

That would have been useful for this bug.

smoge:
I thought about if RAII could supplement the more inadequate parts of the GC. I don't remember is I never wrote or deleted. 

Tim Blechmann mentioned "smart pointers", which is a similar feature of modern c++.

Maybe there is one way to fix but also introduce C++ modern features that will have more support and knowledge around to fine-tune it.

All those talks about refactoring never happened?

Thoughts?

JordanHendersonMusic:
> the more inadequate parts of the GC

I think the GC in principle is pretty awesome, incremental tricolour — I can't quite see what flip does, is it a to-and-from space? — but it pretty much does what you'd do today, maybe there'd be a generation system so new objects are scanned more often than old ones?  

Its main issue is that lack of comments, pretty odd naming conventions, and some bad practices that only trigger under odd conditions — the bug I found with PyrGC calling its own methods in the constructor is an example of that. There are some interfaces that could be improved, e.g., passing the whole stack to each primitive is quite dangerous and would be better to wrap this in something like `std::span<Slot>`, that way it is harder for the primitive to destroy the stack. The temporary object problem is another obvious area for improvement, @muellmusik's pr used RAII to address this. 

Smart pointers seem problematic because of reference cycles, you'd also probably want to roll your own because they are atomic counted in C++ and sc is single threaded. 

Anyway, this conversation is drifting away from the PR, might be better moved to a github discussion? It would also be nice to gather all previous discussions on the topic in one place.

smoge:
> Anyway, this conversation is drifting away from the PR and might be better moved to a github discussion? This PR thread should either focus on the code itself, or on how best to test it — there is a lot to unpack just there. It would also be nice to gather all previous discussions on gc in one place.

I'm behind on everything GC, I will need a few days to catch up to be able to engage with details. You're right that this last message opened up a door that steps back from using GC as it is. So, I'm off-topic here, sorry. But eventually, let's discuss all this when I get more time.

smoge:
Reading some discussions about whether to choose RAII or GC in an interpreter like SC, just taking note of the technical advantages, it would make more sense to implement mostly RAII, including for soft real-time applications, predictability.

Also, if in the future SCLang receives an update regarding concurrency and parallelism, RAII would also be a better choice, according to what I researched.

GC still is more idiomatic for OOP, in any case.

Some references:

1)    Hans-J. Boehm and Mike Spertus. 2009. Garbage collection in the next C++ standard. In Proceedings of the 2009 international symposium on Memory management (ISMM '09). Association for Computing Machinery, New York, NY, USA, 30–38. https://doi.org/10.1145/1542431.1542437

2) [just to get an idea for future improvements]    "C++ Concurrency in Action" by Anthony Williams
    re: advanced C++ usage, including RAII in multithreaded environments
https://dl.acm.org/doi/abs/10.1145/1542431.1542437

muellmusik:
> I might try making the allocator (when a flag is enabled) keep a record of all the alive objects and their size, this way you can ask the allocator if the object ptr is valid, how many objects are alive, etc.

Just to be clear, what do you mean by 'alive' here?

> the bug I found with PyrGC calling its own methods in the constructor is an example of that.

Did I miss that?

> Smart pointers seem problematic because of reference cycles, you'd also probably want to roll your own because they are atomic counted in C++ and sc is single threaded.

Yeah I did that in the PR. We need something very light weight and simple. The normal C++ smart pointers seemed like overkill.

> passing the whole stack to each primitive is quite dangerous and would be better to wrap this in something like `std::span<Slot>`, that way it is harder for the primitive to destroy the stack.

In general the way the receiver and args are passed to the primitives makes it easy to get wrong. ```QC_LANG_PRIMITIVE```does a better job of this. It would be nice to see it hidden behind a safer interface.

> Its main issue is that lack of comments
...
> I'm behind on everything GC

No worries. I've probably spent as much time mucking around in the GC as most people currently active, but not much in recent years so need to spend some more time myself. Would be good to add some comments and documentation and look at this more generally.

JordanHendersonMusic:
> > the bug I found with PyrGC calling its own methods in the constructor is an example of that.
> 
> Did I miss that?

I mentioned it in a comment, just submitted an issue: #6268 

> > I might try making the allocator (when a flag is enabled) keep a record of all the alive objects and their size, this way you can ask the allocator if the object ptr is valid, how many objects are alive, etc.
> 
> Just to be clear, what do you mean by 'alive' here?

Post allocation, but before the GC marks it as unreachable. It would be a way to check the GC has done its job. I don't know if this is covered in any of the sanity check — to be honest I haven't looked *that* closely at them.

muellmusik:
> Post allocation, but before the GC marks it as unreachable. It would be a way to check the GC has done its job. I don't know if this is covered in any of the sanity check — to be honest I haven't looked _that_ closely at them.

So you mean reachable or unexamined? Forgive me but I'm still a little unclear. What mechanism can we use to determine if an object is unreachable besides the GC itself?

If you mean making sure that newly created objects aren't unused, then the pointer thing I did in #3384 does that. If you mean checking that created objects are made reachable between calls to object creation functions the count in that covers that.

Or is this addressing something else? The issues we identified earlier were:


1. Make sure that created objects are made reachable between calls to object creation functions.
2. Make sure that objects made reachable by storing in another object get the correct corresponding calls to GCWrite.
3. Ensure that parent object sizes are correctly updated.
4. Make use of GCWriteNew rather than GCWrite where appropriate.
5. Catch any objects created which are unused. (This seems like it must be an error.)
6. Make common errors in things like stack access less likely.

smoge:
> Would be good to add some comments and documentation and look at this more generally.

I think this could even become a new guideline. SC codebase is an extreme case of avoiding comments.

Let's take a look at the other end of this spectrum, see how many informative comments one can find in the GHC compiler sourcecode: https://github.com/ghc/ghc/blob/master/compiler/GHC/Core.hs

JordanHendersonMusic:
This conversation seems to have dwindled. Was there any thing I should be doing? 

As far as I can see, the only real way to test this is just to put it into debug branch and see if anyone has any issues. Ideally, we'd have a proper bank of test pieces to run this on? Has there ever been a conversation about community members 'donating' piece to be tests or benchmarks?

smoge:
> This conversation seems to have dwindled. Was there any thing I should be doing?
> 
> As far as I can see, the only real way to test this is just to put it into debug branch and see if anyone has any issues. Ideally, we'd have a proper bank of test pieces to run this on — has there ever been a conversation about community members 'donating' piece to be tests or benchmarks?

I understand that's the only test available. I just wonder why so many users and developers are not even able to build their systems for so long, which will affect this method.

I can build and run a few more hairy codes here on Linux, but many more people would be needed.

JordanHendersonMusic:
All CI tests are now passing.

JordanHendersonMusic:
@muellmusik don't suppose you'd fancy reviewing this and moving it towards merging.

JordanHendersonMusic:
@capital-G @dyfer  don't suppose either of you could have a look at this? Its been sat around a while and is very simple change that fixes a segfault in the GC.

muellmusik:
Sorry I’ve been super busy but will try to look over it in the next few days.

capital-G:
I don't have enough knowledge how the GC works, so I can't wrap my head around if the change could lead to problems elsewhere, yet I really appreciate the comments in the code.

One note I have though: Why not add a test which verifies the new fix is working aka not creating a segfault? I think if we fix a bug we should add a simple test for this behavior.

JordanHendersonMusic:
@muellmusik, no worries, this isn't in a rush, just wanted to check it wasn't going stale.

@capital-G re tests. This bug was caused by two things. 
First, allocating certain objects in the parser does **not** call the garbage collector. 
Second, this causes a signed integer to overflow, which is then right shifted. Both of these are implementation defined, which meant this bug is only reproducible on certain hardware. Would a test be worth it in this case?

capital-G:
> is only reproducible on certain hardware

Which hardware are we talking about? Specific memory available, processor architecture ...?
If we can make a deterministic segfault (like in your initial comment) I think we should add a test for this, even if the bug/segfault in question does not occur in our current CI hardware - more tests are always better than no tests, we never know what future hardware for SC will look like.

smoge:
> Which hardware are we talking about? Specific memory available, processor architecture ...?
> If we can make a deterministic segfault (like in your initial comment) I think we should add a test for this, even if the bug/segfault in question does not occur in our current CI hardware - more tests are always better than no tests, we never know what future hardware for SC will look like.

I caught this bug with a simple sketch of a property-based test. The original issue report partially documents it. Before I reported it, it was possible to shrink it to a range.

JordanHendersonMusic:
@capital-G I've added a test. It segfaults the unit testing if it fails — couldn't think of a nice way around that.

I know it happens on linux x86-64, and that it doesn't seems to happen on mac arm. 

I believe it depend on if shifting a negative number preserves the sign ­— although the fact the number overflowed is also a bug and undefined behaviour because its a signed int.

https://en.cppreference.com/w/cpp/language/operator_arithmetic This suggests it is defined behaviour after c++20, so this might actually be compiler version dependent to.

Another option here would be to make the counter a `u64`, as it should never be negative. This wouldn't outright solve the issue, but push it back significantly.

JordanHendersonMusic:
I'm getting a lot of test failures where they don't complete. Its somewhere around...
```
	[1/1]	TestNodeProxy_Server:test_send_after_quit (1.51s)
		 * After server quit, sending should not set node as loaded
	[!]	TestNodeProxy_Server:test_rebuild_after_quit did not complete
... everything after this doesn't complete
```
I don't *think* it has anything to do with this PR, but it does seem particularly bad right now.

dyfer:
@JordanHendersonMusic I definitely lack the in-depth knowledge to review this, but I'd like to push this forward towards merging.

@scztt would you be able to review this by any chance?

lang/LangSource/GC.h
```diff
@@ -303,9 +303,20 @@ inline void PyrGC::ToGrey2(PyrObjectHdr* obj) {
 }
 
 inline PyrObject* PyrGC::Allocate(size_t inNumBytes, int32 sizeclass, bool inRunCollection) {
-    if (inRunCollection && mNumToScan >= kScanThreshold)
+    // When allocating a large number of objects, mNumToScan (int32) can overflow.
+    // This usually happens when creating temporary frames, which might not be collected until later.
+    // This is designed to force collection when there might be a large number of temporary objects.
+    // The threshold is somewhat arbitrary and could be made larger, but it is better to be safe than fast,
+    // and sclang is not designed nor optimized to process large amounts of data.
+    const bool aboveTemporaryObjectLimit = mNumToScan > 5'000'000;
```
dyfer, line 311: Sorry if this was explicitly discussed already (I only skimmed through the comments) - why not set it to the actual limit of int32? Like `INT32_MAX`? Or maybe half of that "to be safe" (not sure why)... `INT32_MAX / 2`?

JordanHendersonMusic: This line

>          mNumToScan += mNumToScan >> 3;

Means it will over shoot intmax - 1.

It could be set to intmax / 2, but feels pretty arbitrary either way.

The other option is to make mnumtoscan an `int64_t` or even a `uint64_t`. I believe this limit would be practically impossible to exceed.

Honestly, I don't think it matters which one. What matters is that right shifting negative signed numbers is implementation defined and causes a crash on some systems. So either a check is needed or the limit of the number should be so large you run out of ram before it overflows to negative.

capital-G (approved):
Should be rebased. If the pipeline is fine after rebasing this can be merged.

capital-G (changes requested):
This seems to introduce many failures in our testing pipeline, so this fix has probably unintended side-effects?

Marking this therefore as _request changes_ until the reason for this has been resolved.

JordanHendersonMusic:
I think it is because I forgot this value can sometimes go negative (it has been a while). I've made it a `std::int64_t` instead which should solve the problem.

JordanHendersonMusic:
@capital-G should be good now.

capital-G (approved):
Thx - feel free to merge @JordanHendersonMusic

PR #6260, opened 2024-04-21
SCDoc: add MathJax for rendering math

prko:
ADDED :

1. As this PR has been modified several times, along with many discussions and some reviews, please read the following post on this PR:
   https://github.com/supercollider/supercollider/pull/6260#issuecomment-2087751459

2. A summary of the main objection to the initial commit of this PR and the author's reaction: 
    https://github.com/supercollider/supercollider/pull/6260#issuecomment-2112753406

3. The state as of around 17 May 2024:
    https://github.com/supercollider/supercollider/pull/6260#issuecomment-2112248966
    - This PR 
       - SCDoc modal tag `math::` is restored
       - MathJax package minimised and optimised
       - Due to Debian policy, MathJax is not included in Debian when building from source.
          (done by adding code to CMakeLists.txt and tested successfully)
       - To do: 
          A MathJax package that does not conflict with Debian policy should be automatically installed when building. 
          Currently, using `npm` is the easiest way, at least from the point of view of the author of this PR.
    - @capital-G's suggestion
       - MathML via temml: not so successful
       - KaTeX is being tested
    - @scztt's suggestion
      - use a cached version on Debian for Debian policy
    - @smoge 
      - mentioned diagram feature of LaTex
---
INITIAL POST:

With this PR, we can represent mathematical expressions using LaTeX and MathML without an internet connection.
- MathJax supports LaTeX and MathML, so I also combined them into a PR.
- I think I have included the minimal files needed to use MathJax offline.

## Purpose and Motivation

I have spent a lot of time using SC-Doc to write course material for my Digital Sound Production lecture, as SuperCollider is one of the key tools in the course. My students have suffered from the lack of a reader friendly layout in SCdoc and I have also found that there are some editing features missing in SCdoc. One of these is the lack of mathematical expressions. Currently, to include reader-friendly mathematical expressions, we have to use graphics from other software such as LaTeX, but this is time consuming and tedious to edit. With this PR, we can save a lot of time and energy, and productivity will increase.

In the following thread are some reactions to the idea of implementing LaTeX and MathML in SCdoc: 
https://scsynth.org/t/scdoc-schelp-style-add-on-latex-mathml-subsubsection-multimedia-file-iframe-sub-sup-small-line-height-hr/9242

This PR is a revised PR of the relevant part of the following PR #6254
I think I have incorporated all the opinions of @nhthn and @jrsurge in this new PR.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
[...]

PR #6263, opened 2024-04-21
SCDoc: add various punctuation marks for non-code text and comments in code

prko:
This PR allows you to insert various quotation marks, ellipses and brackets using predefined character combinations.

Screenshot 2024-04-22 at 03 09 55

Screenshot 2024-04-22 at 03 29 45

## Purpose and motivation

When writing educational material using SCdoc, I needed to use 
- various quotation marks for non-code, text other than ' and ", 
- an apostrophe for non-code text other than ', 
- various brackets for non-code text other than () and []: brackets to indicate the year, brackets for additional explanation, and brackets to indicate the origin of terms (most Korean computer music terms are translations of European terms, and often an English expression is translated into a different expression when used for a different situation).

Inserting such punctuation from the character map is somewhat time-consuming and reduces productivity.
To solve these problems and improve productivity, I created this PR.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


prko:
Converted PDF from the corresponding reference in SCDoc included in this PR:

- [Various punctuation marks for non-code text _ SuperCollider 3.14.0-dev Help.pdf](https://github.com/supercollider/supercollider/files/15053083/Various.punctuation.marks.for.non-code.text._.SuperCollider.3.14.0-dev.Help.pdf)

- [Various punctuation marks for non-code text _ SuperCollider 3.14.0-dev Help with custom css.pdf](https://github.com/supercollider/supercollider/files/15053082/Various.punctuation.marks.for.non-code.text._.SuperCollider.3.14.0-dev.Help.with.custom.css.pdf)

JordanHendersonMusic:
A part of #6260

prko:
@JordanHendersonMusic 
The bugs described in #6278 can be avoided with this PR.

prko:
In this PR, `_` is used to enable various punctuation marks, but a word starting with `_` is related to primitives by its colour. This can be a bit confusing for developers or other users who know these relationships, it may be better to use `:` instead of `_`.

prko:
Converted PDF from the corresponding reference in SCDoc included in this PR:
[Non-code punctuation marks and symbols _ SuperCollider 3.14.0-dev Help.pdf](https://github.com/supercollider/supercollider/files/15344415/Non-code.punctuation.marks.and.symbols._.SuperCollider.3.14.0-dev.Help.pdf)

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -329,7 +368,7 @@ SCDocHTMLRenderer {
 		};
 		stream
 		<< "</h1>\n"
-		<< "<div id='summary'>" << this.escapeSpecialChars(doc.summary) << "</div>\n"
+		<< "\n<div id='summary'>" << SCDocHTMLRenderer.renderNonCodePunctuation(this.escapeSpecialChars(doc.summary)) << "</div>\n"
```
telephon, line 371: why not instead of `SCDocHTMLRenderer.renderNonCodePunctuation(this.escapeSpecialChars(string))`

`this.renderNonCodePunctuation(this.escapeSpecialChars(string))`?

But also I find it hard to unsderstand what kind of transfromations are applied to what kind of section. Why has `\CODE` no `renderNonCodePunctuation`?

It would be good if you could try and give intention revealing names so that all in all we improve the clarity and not make it even harder.

prko: >why not instead of `SCDocHTMLRenderer.renderNonCodePunctuation(this.escapeSpecialChars(string))`
>
>`this.renderNonCodePunctuation(this.escapeSpecialChars(string))`?

I followed your instructions and updated all the relevant parts. When I started creating the PRs, I used the format `this.newMethodInMyPR`. However, I didn't use `this.` when I was calling more than two methods. I cannot remember the reason exactly, but the form you recommended is better! Thanks!

>But also I find it hard to unsderstand what kind of transfromations are applied to what kind of section. Why has \CODE no renderNonCodePunctuation?

I did not add `renderNonCodePunctuation` to `\code`, `\teletype` etc. because I thought that special characters could be used in comments, but that they were not so useful. I have now added `renderNonCodePunctuation` to `\code`, `\teletype` etc.

>It would be good if you could try and give intention revealing names so that all in all we improve the clarity and not make it even harder.

I will do it! Thanks for pointing it out.

telephon: Thanks! Yes, if you think about the intent of each of the string manipulations (also those that you haven't touched), maybe you can find a clear reasoning. Something like:

prepareStringForCode
prepareStringForNonCode 

these could then include the escapeSpecialChars and renderNonCodePunctuation.

By the way "render" implies that something is done to convert to graphics. This is not really the case here, so the above would be better maybe.

# But:
I am now a little wary about all the special syntax this introduces in the helfile. Already now it is really hard if you don't know it, and making it more complicated won't help much in people being able to read the source of help docs and learning from them how to write them.

Maybe the new LaTeX support might make a clearer and more powerful way of doing this?

prko: >Thanks! Yes, if you think about the intent of each of the string manipulations (also those that you haven't touched), maybe you can find a clear reasoning. Something like:
>
>prepareStringForCode
>prepareStringForNonCode
>
>these could then include the escapeSpecialChars and >renderNonCodePunctuation.

**I changed the name of the method from `.renderNonCodePunctuation` to `.prepareStringForNonCode` (I need time to think about how to combine it with `escapeSpecialChars`.):**

Instead of using `.renderNonCodePunctuation`, I think `.prepareStringForNonCode` seems appropriate. Um, and the current method, `escapeSpecialChars`, only replaces three characters, `&`, `<` and `>`, which prevent corruption when converting SCDoc contents to HTML... So you can add them to `prepareStringForNonCode`. 

>By the way "render" implies that something is done to convert to graphics. This is not really the case here, so the above would be better maybe.

In other related PRs I have used the method names `pareHTML` and `parseMathML`. In this PR I have used the term `.renderNonCodePunctuation` because the font size changes when some brackets are used: 
https://github.com/prko/supercollider/blob/014a425e28343822cf43460be9800256a417fa63/SCClassLibrary/SCDoc/SCDocRenderer.sc#L57-L62
```
		.replace(/*"%E2%9D%AA"*/"/f(/", "<span style='font-size: 0.85em'>❪") // MEDIUM FLATTENED
		.replace(/*"%E2%9D%AB"*/"/f)/", "❫</span>")
		.replace(/*"%E2%9D%A8"*/"/m(/", "<span style='font-size: 0.85em'>❨") // MEDIUM
		.replace(/*"%E2%9D%A9"*/"/m)/", "❩</span>")
		.replace(/*"%E2%9D%B2"*/"/s(/", "<span style='font-size: 0.71em'>❲") // TORTOISE SHELL
		.replace(/*"%E2%9D%B3"*/"/s)/", "❳</span>")
```
However, I was not sure if the method name was appropriate. Thanks for pointing this out!

---
>Maybe the new LaTeX support might make a clearer and more powerful way of doing this?

I am intending to include the following characters:
```

		.replace("&amp;hsp;",    "&#x3164;")
		.replace("&amp;#x20;",   "&#x20;"  ) // space.
		.replace("&amp;#32;",    "&#32;"   ) // space.
		.replace("&amp;hairsp;", "&hairsp;") // hair space. &#8202;
		.replace("&amp;#8239;",  "&#8239;" ) // narrow no-break space.
		.replace("&amp;ensp;",   "&ensp;"  ) // en space. &#8194;
		.replace("&amp;emsp;",   "&emsp;"  ) // em space. &#8195;
		.replace("&amp;emsp16;", "&#8198;" ) // 6-per-em space.
		.replace("&amp;emsp14;", "&emsp14;") // 4-per-em space. &#8197;
		.replace("&amp;emsp13;", "&emsp13;") // 3-per-em space. &#8196;
		.replace("&amp;thinsp;", "&thinsp;") // thin space. &#8201;
		.replace("&amp;nbsp;",   "&nbsp;"  ) // non breaking space. &#160;
		.replace("&amp;puncsp;", "&puncsp;") // punctuation space. &#8200;
		.replace("&amp;numsp;",  "&numsp;" ) // figure space. &#8199;
		.replace("&amp;#x3164;", "&#x3164;") // Hangul Filler.
```
The various spaces above and some of the characters included in the PR whether they are linked with the size of characters or not, can not be used via KaTes if I correctly understand KaTeX....

>I am now a little wary about all the special syntax this introduces in the helfile. Already now it is really hard if you don't know it, and making it more complicated won't help much in people being able to read the source of help docs and learning from them how to write them.

Even though I made this PR, I am not sure which way is better:
- Copy special characters from the character map
   (since SC-IDE does not have its own character map, the system's built-in character map should be used, but the workflow is not as different as using typesetting tools)
- Place the symbols using the method described in this PR.

What I can say is,
- In LaTeX and Mathematica, the special symbols are entered via code, not by selecting characters from the character map.
- I switched from Finale to Dorico in 2018. I was a long-time Finale user. I find Dorico's workflow much more intuitive and faster. One reason for this is Dorico's popover, which allows users to place all musical symbols by typing in words, similar to what I did in this PR. 

If someone wanted to write the following paragraph in SCD right now
```
Sie können jeden einzeiligen Ausdruck ausführen, indem Sie auf eine beliebige Stelle in der Zeile klicken und die „Enter“-Taste drücken. Bitte beachten Sie, dass die „Enter“-Taste nicht mit der „Return“-Taste identisch ist.
```
`"` and `"` should be copied from the character map or the text should be written using e.g. Microsoft Word.

However, if this PR is approved, `"` and `"` can easily be expressed in SCDoc.
So I think it would be a good idea to discuss making it easier to abbreviate each symbol. However, I cannot push this idea if none of the main developers agree. I also understand what you are thinking, as I was also very hesitant before making this PR.

capital-G:
@prko I still don't fully understand what this PR implements - not that this PR is not valid, but what is the current way vs. what is the way this PR suggests?

I'd expect that any UTF-8 character which is not interfering with scdoc syntax should be passed from schelp to the rendered HTML file - and these should not need any additional treatment except for `<`, `>` and `&` which need to be escaped within HTML. Or is sccdoc not UTF-8 compatible? 

Does this PR implement "shortcuts"/macros for specific chars? If so, is it also possible to escape those?

Also, the schelp file could be formatted with a bit fewer blanklines

image

prko:
> Also, the schelp file could be formatted with a bit fewer blanklines

I have reduced the number of blank lines.

> is sccdoc not UTF-8 compatible?
> Does this PR implement "shortcuts"/macros for specific chars?

SCdoc is UTF-8 compatible, and this PR is a set of shortcuts (or macros) for entering some non-ASCII characters.

>If so, is it also possible to escape those?

I am not sure I understand your question.... I will do my best to answer correctly:

1. If your question means moving the things in `*prepareStringForNonCode' to `*escapeSpecialChars', their roles are different. So my answer is "no".
2. If your question means moving the things in `*prepareStringForNonCode` to `*escapeSpacesInAnchor`, their names seem incompatible, but their functionality is almost the same. It makes sense to do this because it simplifies the structure of the whole code. So my answer is “yes”.

capital-G:
> I am not sure I understand your question....

My concern is that in the end we introduce a couple of new "primitives" to the SCDoc syntax IIUC.
All these primitives would need to be transferred to a (new) implementation of SCDoc, and I'd prefer to have a minimalist syntax or at least a generic/expressive syntax.

Have you measured if this impacts performance? IIUC every string of our docs would get passed through these replacement methods, which are actually somehow expensive. Also I think this would need some tests that it works as expected and does not break other stuff.

IMO UTF-8 allows to shift this problem into OS UX - e.g. with macOS you can hit `contral command space` and brings up this menu

image

or, in the small layout

image

In which you can even select which kind of symbols you want to see

image

And for the more common stuff, there are keyboard modifiers such as `option u` (think you need to activate it) or `option` and `option shift`. 

Therefore I'd personally am not too fond of including it, but if you find others who think this should be integrated then I won't block it.

prko:
@capital-G 

Thanks for your opinion!

I am familiar with what you mentioned. Why I wanted to implement it was for the following reasons:
- On Windows and Linux, using character maps (or equivalent) is less convenient than on MacOS.
- I thought these shortcuts would be more productive than using these OS built-in tools.

Now, after several months of using my implementation, I do not know if my implementation is always more productive than these OS built-in tools. Sometimes it is very productive, but sometimes not so much. So at the moment, I think it might be better to close this if there is no positive feedback. Any of the developers can silently close this PR. 

I will also close it as soon as I am convinced that this implementation is not useful. Currently, my judgement is that this implementation is somewhat useful. (a bit weaker than a few months ago)

prko:
I am closing this thread for the following reasons.
1. This feature is sometimes handy when typing, sometimes not. Sometimes it is better to use the character map or symbol input functionality provided by the OS, as @capital-G suggested. 
2. Especially when editing text in the code editor of the SC-IDE, the typed text reduces the readability. 

But it has its positive side.
If someone finds this feature useful in the future, he or I can reopen this PR.

PR #6254, opened 2024-04-15
scdoc: adding subsubsection, audio, video, text decoration, some HTML tags, LaTex syntax and MathML syntax 

prko:
## Purpose and Motivation
To make the SCDoc (SuperCollider help documents) more user friendly, I added subsubsections (soft-coded), some HTML tags, LaTex and MathML.

- https://github.com/supercollider/supercollider/issues/6131
- https://scsynth.org/t/scdoc-schelp-style-add-on-latex-mathml-subsubsection-multimedia-file-iframe-sub-sup-small-line-height-hr/9242

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review, but this PR may change dynamically as other users provide feedback.

PR #6265, opened 2024-04-22
SCDoc improvements: 3. Some HTML tags are enabled.

prko:
This pull request enables you to incorporate HTML tags in order to enhance the appearance of the rendered HTML in both the web browser and SC Help Browser. Additionally, "\n" has been added to certain tags to make the rendered HTML more readable when viewed in a text editor.

## Purpose and Motivation
While working on writing educational material using SCdoc, I encountered the need to use HTML tags in order to improve the readability of the printed SCDoc. I used these tags to add empty lines and control line height, and so on. 

Some opinions have already been expressed in PRs #6260 and #6263. I am aware that some believe I may be doing things incorrectly. However, I am doing this in order to complete the task of dividing PR #6254.

This is a revised PR that addresses the relevant part of PR #6254. I plan to keep this PR open for several weeks to gather more feedback on my work. 

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/HTML-tags.schelp
+++ b/HelpSource/Reference/HTML-tags.schelp
@@ -1,0 +1,668 @@
+TITLE::HTML tags in SCDocs
+summary::How to use HTML tags in SCDocs
+categories:: HelpSystem
+related:: Guides/WritingHelp, Classes/SCDoc, Classes/SCDocHTMLRenderer
+
+
+
+
+
+
+
+
+
+
+section:: Formatting Text
+
+
+
+
+
+subsection:: Paragraph spacing
+
+
+The following code prints a blank line:
+
+table::
+##
+code::
+Line 1 begins.
+It is still Line 1.
+
+Line 2
+
+<br>Line 4 due to '<br>'
+
+<br> Line 6 due to '<br>'
+
+<br>
+Line 8
+::
+
+||
+Line 1 begins.
+It is still Line 1.
+
+Line 2
+
+<br>Line 4 due to '<br>'
+
+<br> Line 6 due to '<br>'
+
+<br>
+Line 8
+::
+
+<br>
+
+The height of a line can be adjusted with the units like teletype::em::, teletype::rem:: and teletype::%:::
+
+table::
+##code::
+Line 1
+<p style='line-height:33%'><br>
+Line 4
+<p style='line-height:0.33em'><br>
+Line 6
+<p style='line-height:0.33rem'><br>
+Line 8
+<p style='line-height:0.66%'><br>
+Line 10
+<p style='line-height:0.66em'><br>
+Line 12
+<p style='line-height:0.66rem'><br>
+::
+||
+Line 1: default line-height
+
+Line 2: 'line-height:33%'
+<p style='line-height:33%'><br>
+Line 4: 'line-height:0.33em
+<p style='line-height:0.33em'><br>
+Line 6: 'line-height:0.33em'
+<p style='line-height:0.33rem'><br>
+Line 8: 'line-height:66%'
+<p style='line-height:66%'><br>
+Line 10: 'line-height:0.66em'
+<p style='line-height:0.66em'><br>
+Line 12: 'line-height:0.66rem'
+<p style='line-height:0.66rem'><br>
+::
+
+<br>
+
+table::
+##strong::An Example:::
+table::
+##The standard Lorem Ipsum passage, used since the 1500s:
+
+emphasis::Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.::
+
+<p style='line-height:0.3em'><br><p>
+
+Section 1.10.32 of “de Finibus Bonorum et Malorum”, written by Cicero in 45 BC:
+
+"emphasis::Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?::"
+::
+::
+
+<br>
+
+
+
+
+
+subsection:: Table
+
+
+All table tags are supported as follow:
+
+code::
+  <table border='1px solid black' style='text-align:center'>
+    <caption>Order Details</caption>
+    <thead>
+      <tr>
+        <th scope='col' style='text-align:center'>Order ID</th>
+        <th scope='col'>Customer</th>
+        <th scope='col'>Items</th>
+		<th scope='col' colspan='2'>Shipping</th>
+      </tr>
+    </thead>
+
+    <tbody>
+      <tr>
+        <th scope='row' rowspan='2'>12345</th>
+        <td rowspan='2'>John Doe</td>
+        <td>Headphones</td>
+        <td rowspan='2'>Address</td>
+        <td>Standard</td>
+      </tr>
+      <tr>
+        <td>Speakers</td>
+        <td>Express</td>
+      </tr>
+      <tr>
+        <th scope='row'>67890</th>
+        <td>Jane Smith</td>
+        <td>Software Bundle</td>
+        <td colspan='2'>Downloadable</td>
+      </tr>
+    </tbody>
+  </table>
+::
+
+<br>
+
+table::
+##
+  <table border='1px solid black' style='text-align:center'>
+    <caption>Order Details</caption>
+    <thead>
+      <tr>
+        <th scope='col' style='text-align:center'>Order ID</th>
+        <th scope='col'>Customer</th>
+        <th scope='col'>Items</th>
+		<th scope='col' colspan='2'>Shipping</th>
+      </tr>
+    </thead>
+
+    <tbody>
+      <tr>
+        <th scope='row' rowspan='2'>12345</th>
+        <td rowspan='2'>John Doe</td>
+        <td>Headphones</td>
+        <td rowspan='2'>Address</td>
+        <td>Standard</td>
+      </tr>
+      <tr>
+        <td>Speakers</td>
+        <td>Express</td>
+      </tr>
+      <tr>
+        <th scope='row'>67890</th>
+        <td>Jane Smith</td>
+        <td>Software Bundle</td>
+        <td colspan='2'>Downloadable</td>
+      </tr>
+    </tbody>
+  </table>
+::
+
+
+
+
+subsection:: Horizontal Line
+
+
+The following code prints a horizontal line:
+
+code::
+<hr>
+::
+table::
+##<hr>
+::
+
+A horizontal line in teletype:
+
+table::
+##
+teletype::
+text
+<hr>text
+::
+::
+
+
+
+
+
+subsection:: Small Text
+
+
+The following code displays small text:
+code::
+<small>This is small text.</small>
+::
+
+table::
+##<small>This is small text.</small>
+::
+
+Normal and small text in teletype:
+
+table::
+##
+teletype::
+This is normal text.
+<small>This is small text.</small>
+::
+::
+
+
+
+
+
+subsection:: Strong and Emphasis
+
+
+list::
+##The following code displays bold text:
+code::
+strong::This is bold text.::
+::
+
+table::
+##strong::This is bold text.::
+::
+
+To use bold text in teletype, you can use a pair of the following tags:
+code::
+<strong></strong>
+::
+
+table::
+##
+teletype::
+<strong>This is bold text.</strong>
+::
+::
+
+##The following code
+code::
+emphasis::This is italic text.::
+::
+and the following code display italic text:
+code::
+<em>This is italic text.<em>
+::
+table::
+##
+<em>This is italic text.<em>
+::
+
+To use bold text in teletype, you can use a pair of the following tags:
+
+code::
+<em></em>
+::
+
+table::
+##
+teletype::
+<em>This is italic text.</em>
+::
+::
+
+##Bold italic text is not available with the built-in hardcoded link::Reference/SCDocSyntax::.
+The following code displays bold italic text:
+
+code::
+<em>strong::This is bold italic text.::</em>
+::
+
+table::
+##
+<em>strong::This is bold italic text.::</em>
+::
+
+To use bold text in teletype, you can use one of a pair of the following tags:
+
+code::
+<em><strong></strong></em>
+::
+table::
+##teletype::
+<em><strong>This is bold italic text.</strong></em>
+::
+::
+
+or
+
+code::
+<strong><em></em></strong>
+::
+
+table::
+##teletype::
+<strong><em>This is bold italic text.</em></strong>
+::
+::
+::
+
+<br>
+
+Small text can also be italic, bold and bold italic.
+table::
+##code||text||teletype
+##code::
+<small>small normal text</small>
+::
+||
+<small>small normal text</small>
+||
+teletype::
+<small>small normal text</small>
+::
+##
+code::
+<small><strong>small bold text</strong></small>
+::
+||
+<small><strong>small bold text</strong></small>
+||
+teletype::
+<small><strong>small bold text</strong></small>
+::
+##
+code::
+<small><em>small italic text</em></small>
+::
+||
+<small><em>small italic text</em></small>
+||
+teletype::
+<small><em>small italic text</em></small>
+::
+##
+code::
+<small><em><strong>small bold italic text</strong></em></small>
+::
+||
+<small><em><strong>small bold italic text</strong></em></small>
+||
+teletype::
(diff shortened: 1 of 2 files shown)


prko:
Converted PDF from the corresponding reference in SCDoc included in this PR:

- [HTML tags in SCDocs _ with custom.css.pdf](https://github.com/supercollider/supercollider/files/15066717/HTML.tags.in.SCDocs._.with.custom.css.pdf)

-  [HTML tags in SCDocs _ with default.css.pdf](https://github.com/supercollider/supercollider/files/15066722/HTML.tags.in.SCDocs._.with.default.css.pdf)

prko:
Converted PDF from the corresponding reference in SCDoc included in this PR:

- [HTML tags in SCDocs _ with custom.css.pdf](https://github.com/supercollider/supercollider/files/15088198/HTML.tags.in.SCDocs._.with.custom.css.pdf)

- [HTML tags in SCDocs _ with default.css.pdf](https://github.com/supercollider/supercollider/files/15088200/HTML.tags.in.SCDocs._.with.default.css.pdf)

prko:
updated again (deleted previous ones)

Converted PDF from the corresponding reference in SCDoc included in this PR:

- [HTML tags in SCDocs _ with custom css.pdf](https://github.com/supercollider/supercollider/files/15096658/HTML.tags.in.SCDocs._.with.custom.css.pdf)
- [HTML tags in SCDocs _ with default css.pdf](https://github.com/supercollider/supercollider/files/15096660/HTML.tags.in.SCDocs._.with.default.css.pdf)

capital-G:
I really appreciate the effort here and I see similarities to your proposed https://github.com/supercollider/supercollider/issues/6304 , but using replace to replace generated SCDoc HTML rendered code seems a hacky way of doing it as in this case we work against the output of SCDoc instead of telling it to "ignore" it.

For example, `<p class="foo">` and `< p class="foo">` are both valid HTML p tags, but would not be the same in the current output, and parsing HTML [even with the power of regex is not possible](https://stackoverflow.com/a/1732454), although a limited set is possible.

Maybe there should be a focus on how to extend the SCDoc syntax to include a `raw` or `html` tag which simply (by-) passes any HTML to the generated file.

So, instead of working against the SCdoc generator, we can just work with it :)

The definition of the SCdoc syntax is stated here
Lexical definition: https://github.com/supercollider/supercollider/blob/develop/SCDoc/SCDoc.l
Grammar definition: https://github.com/supercollider/supercollider/blob/develop/SCDoc/SCDoc.y

which are used by [Bison](https://www.gnu.org/software/bison/) to generate a parser, see  https://www.oreilly.com/library/view/flex-bison/9780596805418/ch01.html

I think it should be feasible to include a new tag here, if you are interested in this we can take a look at this together, but currently the `build_parser.sh` is broken (at least on my machine), maybe we should add this to the CI in general?

smoge:
> Maybe there should be a focus on how to extend the SCDoc syntax to include a `raw` or `html` tag which simply (by-) passes any HTML to the generated file.

I thought about it today about an HTML:: tag that bypasses the renderer. It's an idea.

prko:
@smoge 
>I thought about it today about an HTML:: tag that bypasses the renderer. It's an idea.

Wow, I think it can simplify a lot of things! I really like it!

---

@capital-G 
>The definition of the SCdoc syntax is stated here
Lexical definition: https://github.com/supercollider/supercollider/blob/develop/SCDoc/SCDoc.l
Grammar definition: https://github.com/supercollider/supercollider/blob/develop/SCDoc/SCDoc.y
I think it should be feasible to include a new tag here, if you are interested in this we can take a look at this together
 
I have modified the files you refer to several times over the past year to implement `subsubsection`, but to no avail.

Recently @smoge also recommended learning Bison and Flex...
I replied that I can not learn them because I do not know C++. Now @capital-G is recommending the same thing.... When two people on the forum recommend the same thing in the same week... Hard to say no again...

In between
- @smoge uploaded a draft on modal tags.
- I found the story of how to remove the `math::` tag at ......

OK! Modifying these files without exact knowledge is the reason for non-success. After I finish some urgent personal stuff, I will have a look at it.

>currently the build_parser.sh is broken (at least on my machine), maybe we should add this to the CI in general?

I do not know what `build_parser.sh` and what `CI` is... However, if you do I will try to learn it!

---
@smoge  @capital-G 
Thank you for helping me and giving me a great idea!

smoge:
It was a minor deprecation and easily fixable with `bison SCDoc.y --update` 

Also, the commented line that compiles with gcc, also works with a minor change.

I had heard about this "breakage" before. But it does not seem to be the case.

prko:
@smoge @capital-G 
I have success in implementing `html::` modal tag in the same way of `teletype::`.

Here I have a new question. The followings for example seems not so fun to typing:
- `html::<br>::`
- `html::<br><br>`
- `html::<hr>::` 
- `html::<p>::`
  
etc...

I think it would be better to make a poll for the new modal tags. 
For example 
- `html::<br>::` -> `lin::1::`
- `html::<br><br>::` -> `lin::2::` 
- `html::&nbsp;::` -> `sp::1::`` 
 
I also would like to have a tag for table with `<thead>` and `<th>`.

I think @capital-G would wish some modal tags related to this....
How do you think about it?

prko:
>- `html::<br>::` -> `lin::1::`
>- `html::<br><br>::` -> `lin::2::`

The followings seem to be better:
- `html::<br>::` -> `br::lin1::`
- `html::<br><br>::` -> `br::lin2::`

Also:
- `html::<p style='page-break-before: always;'>::` -> `br::pg::`

capital-G:
I think we should limit this more towards a `<div>` or context-based approach here instead of introducing all possible features of HTML through it. SCDoc should be independent of HTML features as otherwise we limit SCDoc for all time to come to be viewed only within a browser environment.

We could nonetheless include something like [admonitions in myst](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html) or [rst](https://docutils.sourceforge.io/docs/ref/rst/directives.html#toc-entry-3) with an either limited or open set of possible classes.

Something like

```
context:: math
    \int_0^\inf = \sin(x)~dx
::
```

would then be converted to

```html
<div class="math">
    \int_0^\inf = \sin(x)~dx
</div>
```

This makes it possible and very easy to edit such divs via JS or CSS and adjust/modify their style or content.

smoge:
> I think we should limit this more towards a `<div>` or context-based approach here instead of introducing all possible features of HTML through it. SCDoc should be independent of HTML features as otherwise we limit SCDoc for all time to come to be viewed only within a browser environment.
> 
> We could nonetheless include something like [admonitions in myst](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html) or [rst](https://docutils.sourceforge.io/docs/ref/rst/directives.html#toc-entry-3) with an either limited or open set of possible classes.
> 
> Something like
> 
> ```
> context:: math
>     \int_0^\inf = \sin(x)~dx
> ::
> ```
> 
> would then be converted to
> 
> ```
> <div class="math">
>     \int_0^\inf = \sin(x)~dx
> </div>
> ```
> 
> This makes it possible and very easy to edit such divs via JS or CSS and adjust/modify their style or content.



> I think we should limit this more towards a `<div>` or context-based approach here instead of introducing all possible features of HTML through it. SCDoc should be independent of HTML features as otherwise we limit SCDoc for all time to come to be viewed only within a browser environment.
> 
> We could nonetheless include something like [admonitions in myst](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html) or [rst](https://docutils.sourceforge.io/docs/ref/rst/directives.html#toc-entry-3) with an either limited or open set of possible classes.
> 
> Something like
> 
> ```
> context:: math
>     \int_0^\inf = \sin(x)~dx
> ::
> ```
> 
> would then be converted to
> 
> ```
> <div class="math">
>     \int_0^\inf = \sin(x)~dx
> </div>
> ```
> 
> This makes it possible and very easy to edit such divs via JS or CSS and adjust/modify their style or content.

Yes, let's follow the KISS principle. Let's add one or two tokens to the grammar, which will allow much more expressivity on the scaling side. It will be better for future ideas and adaptations. Bison works just fine; I don't think it's 'archaic,' it's a good tool. But grammars are a hard part of programing. Changes will always be more complicated.

prko:
>We could nonetheless include something like [admonitions in myst](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html) or [rst](https://docutils.sourceforge.io/docs/ref/rst/directives.html#toc-entry-3) with an either limited or open set of possible classes.
>
>context:: math
>    \int_0^\inf = \sin(x)~dx
>::
>
>html::
><iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/dQw4w9WgXcQ?si=WEomThZlWcYABnhC" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
>::
Yes, these are good!

Yes, these are good!

I already use HTML in SCDoc using the SCHTMLRenderer in this PR. 

There is a problem: by using HTML I should type in too much code to use `<style>` (often also `<span>`). You can see examples in the attached PDF at the end of this post. For the user's coding experience, row HTML is flexible but not so amusing. To solve this, we should add more CSS definitions in `SCDoc.css', linked to the modal tags. Of course, users can add `custom.css` if they want to customise some CSS definitions. 
Moreover, I think now this PR, #6263 and #6266 should be merged by adding the following modal tags.

Thus, we need the following modal tags:

- For readability, managing style for code parts and non-code parts and reducing bugs:
   - `char` or `punc` for various punctuation marks (quotation marks, brackets etc) listed in
       #6263 
   
   [note] 
   For errors caused by using the same punctuation for non-code text and code:
   - #6278 
   - #6263
- For for printing support:
   - `height` for `<p style='line-height:xx%'>` 
   - `br` or `break` to add various spaces, empty lines and page breaks 
      introduced in https://github.com/supercollider/supercollider/pull/6265#issuecomment-2109288990
   This is important for printing support.
   - `image` 
      An argument for width and height should be added. 
      I think including just the height is good because the width can be calculated automatically.
- For a more readable layout:
   - `context`
   - some from `attention`, `caution`, `danger`, `error`, `hint`, `important`, `seealso` and `tip`
   - `div`
   - `iframe`
   - `audio`
   - `video`
   - `post` 
      requested in #6304
   - subsubsection
   - `sub` and `sup`
   - `small`
   - `strongemphasis`
   - `pre` 

The following PDF is a guide to all these features. 
It is written with the wrong implementation because I did not know about adding new modal tags.
But it helps to understand what I mean.

[Advanced SCDoc features _ SuperCollider 3.14.0-dev Help.pdf](https://github.com/supercollider/supercollider/files/15315176/Advanced.SCDoc.features._.SuperCollider.3.14.0-dev.Help.pdf)


Is there anything I have missed?
Do we need to make a poll on the scsynth forum about this?

capital-G:
As mentioned above, I personally would strongly avoid *any* HTML beyond an `iframe` - though there are always exceptions to the rule.

By implementing a context (represented by wrapping the content inside a div that has a class as an attribute), it is possible to implement most of what you ask for.
Additional functionality or styles can then be implemented via `js' and `css' for that context, and can be reviewed/discussed on an atomic basis.
If these features don't get accepted, but you feel the need to share them with a group of people, you could simply distribute a `frontend.js` and `custom.css` file for the group.

In general, I don't think that development efforts within SuperCollider are best spent on making SCDoc equal to other documentation or typesetting tools. If the community agrees that there is an urgency to include such features, then we should replace SCDoc with a tool dedicated to documentation like doxygen.

prko:
>In general, I don't think that development efforts within SuperCollider are best spent on making SCDoc equal to other documentation or typesetting tools. If the community agrees that there is an urgency to include such features, then we should replace SCDoc with a tool dedicated to documentation like oxygen.

I would like to mention that the community here are all experts in using SC and many of them are professional level programmers. They are not so dependent on the SuperCollider help document as they often understand just by looking at the implementation of the methods.

However, my recent PRs are for my students who are outside the community. They are studying various subjects, are not musicians (I used to teach musicians, mainly composers, but not anymore) and are learning Audacity, Reaper and programming with SC for the first time in their lives.

I do not use PDF or other tools to present code because some characters can be converted, so errors can occur when evaluating pasted code. I mainly use SCD files and my own tutorials in Korean written in SCDoc. SCD files are good for code, but not good for writing explanations as comments. SCDoc is good for linking the help documents and other help documents. The advantage of using SCDoc is that you can embed images. However, it is not very eye-friendly and its readability needs to be improved by adjusting the lead (spacing between lines), table width, line width between items, etc. The current fonts are also not very good for Korean.

The use of the online documentation system is also very limited. There are computers with wired internet in the lecture room, but many students use their own laptops with Wi-Fi, but the Wi-Fi connection in the room is very unstable.

I do not think there is an urgent need to implement such features for the community.
I think implementing the features I recently suggested will be much faster than replacing Oxygen.

However, I think it would not be a bad idea to start a poll to replace SCDoc with Oxygen. What do you think? If I make a poll, it will be ridiculous because I am not competent to replace SCDoc with Oxygeon.

Or would it be better to close all my recent PRs related to this?

prko:
I have updated #6266. The subsubsection now seems to be implemented correctly.
Regarding this PR, I will do as @capital-G recommends. Could you give a specific list of what you think should be added? If you could be generous enough to include some of the above, I would appreciate it. But if not, it is not a problem as I understand what you have written.

capital-G:
@prko

> However, I think it would not be a bad idea to start a poll to replace SCDoc with Oxygen. What do you think? If I make a poll, it will be ridiculous because I am not competent to replace SCDoc with Oxygeon.

I suppose you mean [Doxygen](https://www.doxygen.nl/)?
I don't think we need to actually replace SCDoc as it still has it charms and history but also it limitations, but therefore also some perks that would most likely get lost on a transition.

I think the recent PRs showed that there is still interest in SCDoc and we will also hopefully find a way to have create a way to improve it over time, but please be aware that SCDoc will never become a perfect typesetting tool as this is out of scope. (BTW, doxygen is also not a typesetting tool^^)

Maybe a bit OT but if you are interested in writing course material I can maybe recommend a tool I wrote https://github.com/capital-G/sc_kernel which allows you to write Jupyter Notebooks with sclang.
Combined with [jupyter book](https://jupyterbook.org/en/stable/intro.html) you can combine these notebooks into a book which can be exported to a HTML webpage or a PDF via tex.

An example of this can be found 
https://capital-g.github.io/musikinformatik-sose2021/00_basics/sc_dimensions.html
where the source code is located at 
https://github.com/capital-G/musikinformatik-sose2021

## The Problem in general

I think the DOM structure of the SCDocs HTML could be improved, as it is currently almost impossible to get hold of specific sections of a document via JS and CSS and adjust their style, because the DOM is essentially flat (and not deterministic) rather than nested within divs which create namespaces.

image

Normally every kind of section and element (e.g. a paragraph) should be nested with a div.
But introducing this would also break the existing CSS.
I think this *could* be justified as this could lay a foundation for really new looks on the documentation which is currently really limited due to this lack.

But instead of going for such a big step I think it is best to implement something less drastic and would be useful even even if we have a proper nested DOM by allowing to create a context wrapper explicitly.

## Proposition

### Context

I think https://github.com/supercollider/supercollider/pull/6265#issuecomment-2110077273 already states how admonitions work within MyST/RST and this could be mimiced to create a context syntax within SCDoc.

Something like

```
context::{foo bar}
this is a context
::
```

should then result in

```
<div class="foo bar">
  this is a context
</div>
```

Via `scdoc.css` we could then address this contexts via

```css
.foo {
  background-color: red;
}

.bar {
  font-weight: bold;
}
```

The set of global classes could then be discussed in an atomic basis.
It is basically a generalization of the `NOTE` and `WARNING` boxes which *could* be replaced via

```
context::{warning}
this is a warning
::
```

If you could extend the SCDoc parser to include the parsing of the context including classes like above it would be really great!!

Maybe this could also allow to enter raw HTML for elements such as iframe

```
context::{raw-html}
<iframe>...</iframe>
::
```

via js

```js
    // pseudo-code
    for (let element of document.getElementsByClassName("raw-html")) {
        const text = element.textContent;
        element.innerHTML = element.textContent;
    }
```


Maybe there could be a syntax to create an inline-context via span classes, but I think this should be something for a 2nd PR.

prko:
@capital-G 
Thanks, I got it!

Do I need to add the modal tag `context` in the same way as `teletype`? Or would you like to do it, since it is your idea? I do not care who does it. What you suggest is much better than what I did in this PR!

---

Why have I tried to improve SCDoc?

- Copying and pasting code from external sources can change the characters in the code, including whitespace.
- Using an external application with SC is a bit clumsy. 
- The SCD file cannot contain images, tables, etc. Before Qt, this was fine as it has rtfd format!
- SCDoc can contain images, tables, lists and many other things, but there are some missing features.
- Allowing students to customise SCDoc using my `/SCClassLibrary/SCDoc/SCDocRenderer.sc` file and `/HelpSources/custom.css` takes too long.

Adding the modal tag `context` in SCDoc seems to resolve most of these things! 
In any case, #6263 seems to be closed...

---

>I suppose you mean [Doxygen](https://www.doxygen.nl/)?
(BTW, doxygen is also not a typesetting tool^^)

Yes, it was a typo. Sorry!

From my experience, 
- Using non-standard characters, as suggested in #6263, is not convenient in typesetting tools either.
- Adjusting the layout page by page is a lot of work, with slight changes in lead, font width, paragraph spacing, and spacing between the list (including images, tables, etc.) and the paragraphs above and below. 
- For me, there is not much 

---

>Maybe a bit OT but if you are interested in writing course material I can maybe recommend a tool I wrote https://github.com/capital-G/sc_kernel which allows you to write Jupyter Notebooks with sclang.
Combined with [jupyter book](https://jupyterbook.org/en/stable/intro.html) you can combine these notebooks into a book which can be exported to a HTML webpage or a PDF via tex.

I am already familiar with your great tool and I like to use it. However, more than 90% of my students use Windows and very few of them use MacOS. In 10 years of teaching experience, I only meet 1 or 2 students who use Linux. So I cannot use your wonderful tool with my students. I think that a teacher and his students should use the same tool to avoid many technical problems.

capital-G:
@prko I think most of your points concerning styling of the docs can be achieved by using nested layouts, switching to flex layouts and introducing the "custom-context" syntax .

I am most certain that the CSS wil break - but the code is already a mess so it could already require a rewrite.
While we are at it we could ditch our outdated jQuery for something more modern such as alpine.js in order to reduce LOC and complexity.

I'll try to think about it and start a PR in the end of june? In the meantime I'll try to formulate my thoughts more precisely in a new issue.

prko:
@capital-G
Thank you very much!

prko:
@capital-G 
We should also think about quotations. I think it can also be included in the context tag:
https://github.com/supercollider/supercollider/issues/6381#issuecomment-2229613470

prko:
@capital-G @tedmoore 
I would suggest that enabling a video frame may not be the most suitable approach, for the following reason:  
When printing rendered HTML, the video frame often fails to display the intended screenshot and instead appears as a blank space.  
In such cases, using a static image accompanied by a hyperlink tends to be more reliable and effective.

Issue #6138, opened 2023-11-17
At the top left of the help browser, the image hides the S of SuperCollider.


prko:
## Environment

* SuperCollider version: 3.14.0_dev
* Operating system: macOS 13.6.2, Windows 11, Ubuntu 22.04

## Steps to reproduce

1. launch SC
2. look at the top left of Help Browser

### expected
<img width="880" alt="Screenshot 2023-11-17 at 19 59 23" src="https://github.com/supercollider/supercollider/assets/416281/673d000c-bdc1-46f1-be57-66649aca8964">

### actual
<img width="844" alt="Screenshot 2023-11-17 at 20 00 59" src="https://github.com/supercollider/supercollider/assets/416281/4260ce88-ef7b-42db-9055-d0bf0208aaa5">
<img width="996" alt="Screenshot 2023-11-17 at 20 00 24" src="https://github.com/supercollider/supercollider/assets/416281/2d056690-173f-4d43-a321-b176ce87278a">
<img width="580" alt="Screenshot 2023-11-17 at 18 58 32" src="https://github.com/supercollider/supercollider/assets/416281/f4029042-1133-4258-8dc6-69722e0a53bc">

prko:
I think this is due to the following in `scdoc.css' in SC 3.13.0 and above:
```
#menubar .menuitem .home
```
Previously the corresponding parts were:
```
#menubar .menuitem:first-child
```
@scztt You seem to have changed it. Could you have a look at it?

prko:
https://github.com/supercollider/supercollider/pull/6348 fixes this problem.

PR #6275, opened 2024-04-28
Resolve the conflict between SC-cube and text in the SC help Home with a typo correction in scdoc.css

prko:
## Purpose and Motivation
1. This PR fixes the problem reported in #6138.
   - The reported problem:
   Screenshot 2024-04-28 at 14 49 11
   - Fixed with this PR:
   Screenshot 2024-04-28 at 14 50 37

I do not think my fix is the only way. I can remove it from this PR if @scztt fixes it, because I think this problem is definitely due to @scztt's fix as described in https://github.com/supercollider/supercollider/issues/6138#issuecomment-1930754168.

2. This PR also removes an unnecessary */ on line 691 in scdoc.css:
   https://github.com/supercollider/supercollider/blob/3b3eca75a1b42ad4a6c1146ad1893568d083d5c5/HelpSource/scdoc.css#L691

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/scdoc.css
+++ b/HelpSource/scdoc.css
@@ -131,7 +131,7 @@ a.subclass_toggle:hover {
     background-image: url("images/SC_icon.png");
     background-repeat: no-repeat;
     background-position: 0.5em center;
-    background-size: 1.5em;
+    background-size: 1.1em;
 }
 
 a.menu-link {
@@ -688,7 +688,7 @@ table#search_settings {
     color: #555;
     width: 100%;
     margin-left: 0px;
-    margin-right: 0px;*/
+    margin-right: 0px;
     margin: 0px;
     padding: 0px;
     border-collapse: collapse;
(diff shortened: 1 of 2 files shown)


capital-G (changes requested):
Thanks for this fix!

Interestingly enough, this seems to have been introduced only recently, as 3.13 docs do not have it, but dev docs have it.

The only change on the docs I found which could have introduced this is https://github.com/supercollider/supercollider/commit/39702ec3a9cdc9a4e2ee7385db747e9ba135c3d1

So maybe it would be better to adjust the selectors so they work properly again? This would be closest to "just keep things as they are", which I think is the intention here.

Also using whitespaces to adjust placement is not considered the best kind of CSS styling :)

edit:

I think adding the following code would fix it - can you verify it @prko ?

```css
li.menuitem a.menu-link.home {
    padding-left: 2.5em;
}
```

prko:
@capital-G 
>
```li.menuitem a.menu-link.home {
    padding-left: 2.5em;
}
```
Adding this block of code will fix the problem, so there is no need to change `scdoc.js`.
However, the sccube image should still be resized. 

Your suggestion is better than my PR! 
If you want to fix it, I can close it. It is not a problem!
Let me know what to do!

prko:
As #6348, which is better than this PR, is published, I close this PR.
Thanks, @capital-G

PR #6230, opened 2024-03-06
Add the argument bounds to the .scope and .plotTree methods to define their bounds; add the argument position to the .meter method to define its position

prko:
## Purpose and Motivation
`.scope` has no argument to define its window location. Often it is good to see this window on the centre of the screen, but sometimes it is not so optical. 
This modification adds one more argument for this feature. Thus, the following code is available (I intentionally paste all code to present the necessity of this feature):
```
(
Window.closeAll;
s.waitForBoot {
	var sigs = {
		[
			SinOsc.ar(440),
			(SinOsc.ar(440 * (1..48)) * (1..48).reciprocal / 16).sum,
			Resonz.ar(Resonz.ar(WhiteNoise.ar, 440, 0.1, 5), 440, 0.2, 5),
			Resonz.ar(
				Resonz.ar(WhiteNoise.ar, 440 * (1..48), 0.2, 5),
				440 * (1..48), 0.2, (1..48).reciprocal
			).sum,
			WhiteNoise.ar
		] * Env.perc(0.01, 0.79, 0.2).ar(Done.freeSelf)
	};
	FreqScope();
	sigs.().size.do { |i|
		{ sigs.()[i] ! 2 }.scope(position: 605@0); // <- look at this!
		{ sigs.()[i] }.plot(0.04, bounds:Rect(230 * i, 355 + 175, 230, 150));
		{ sigs.()[i] }.plot(1, bounds:Rect(230 * i, 355, 230, 150));
		2.wait
	}
}
)
``` 
demo video:
https://www.dropbox.com/scl/fi/qj0337im08mqyzcj2k4fx/Screen-Recording-2024-03-06-at-14.03.44.mov?rlkey=k82z5kwoi2w0nj23sy56nfxht&dl=0

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #6281, opened 2024-05-01
add .bounds_ method to Stethoscope

prko:
## Purpose and Motivation

This is a reopened PR: #6230
This PR
- allows you to set the size and position of `Stethoscope`,
  - e.g. `s.scope.bounds_(Rect(100, 200, 300, 100))`
-  and change the window resizing behaviour.

demo video:
https://www.dropbox.com/scl/fi/26952qepzu3iczj9bi0cq/s.bounds_-behaviour.mov?rlkey=suub8zyf6ujpksc7ugieggp02&dl=0

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Stethoscope.schelp
+++ b/HelpSource/Classes/Stethoscope.schelp
@@ -25,6 +25,8 @@ table::
 ## space || run, if not running already
 ## . (period) || stop
 ## M || toggle screen size
+
+(The Scope window adjusts its position to not exceed screen boundaries when resized.)
 ## + / - || zoom horizontally
 ## * / _ || zoom vertically
 ## S || change style between parallel and overlay
--- a/HelpSource/Classes/Stethoscope.schelp
+++ b/HelpSource/Classes/Stethoscope.schelp
@@ -177,14 +177,16 @@ SUBSECTION:: Size and position
 METHOD:: bounds
     The position and size of the window. The position is relative to the bottom-left corner of the screen.
     argument::
-        A Rect or a Point interpreted link::Classes/Point#-asRect#as Rect::.
+        A Rect, a Point interpreted link::Classes/Point#-asRect#.asRect:: or anything that responds to it.
     returns::
         A Rect.
 code::
 s.scope.bounds_(Rect(100, 200, 500, 600));
+s.scope.bounds_(500@600);
 s.scope.bounds_([100, 200, 500, 600]);
 s.scope.bounds_(Size(500, 600));
 ::
+
         
         
 SUBSECTION:: Operation
--- a/HelpSource/Classes/Stethoscope.schelp
+++ b/HelpSource/Classes/Stethoscope.schelp
@@ -172,6 +172,16 @@ METHOD:: style
         One of the above Integers.
 
 
+SUBSECTION:: Size and position
+        
+METHOD:: bounds
+    The position and size of the window. The position is relative to the bottom-left corner of the screen.
+    argument::
+        A Rect or a Point interpreted link::Classes/Point#-asRect#as Rect::.
+    returns::
+        A Rect.
+        
+        
 SUBSECTION:: Operation
 
 METHOD:: run
--- a/HelpSource/Classes/Stethoscope.schelp
+++ b/HelpSource/Classes/Stethoscope.schelp
@@ -134,6 +134,22 @@ SUBSECTION:: Display
 METHOD:: window
     The (parent) Window of the scope.
 
+METHOD:: bounds
+    The position and size of the window. The position is relative to the bottom-left corner of the screen.
+    argument::
+        A link::Classes/Rect::, or any object responding to the link::Search#asRect:: method. The width less than 264 will be changed to 264, the minimum width.
+    returns::
+        A Rect.
+code::
+s.scope.bounds_(Rect(100, 200, 500, 600));
+s.scope.bounds_([100, 200, 500, 600]);
+s.scope.bounds_(Size(500, 600));
+::
+code::
+s.scope.bounds_(Rect(0, 0, 500, 600));
+s.scope.bounds_(500@600);
+::
+
 METHOD:: size
     Sets the width and the height of the scope window.
     argument::
@@ -170,23 +186,6 @@ METHOD:: style
 
     argument::
         One of the above Integers.
-
-
-SUBSECTION:: Size and position
-        
-METHOD:: bounds
-    The position and size of the window. The position is relative to the bottom-left corner of the screen.
-    argument::
-        A Rect, a Point interpreted link::Classes/Point#-asRect#.asRect:: or anything that responds to it.
-    returns::
-        A Rect.
-code::
-s.scope.bounds_(Rect(100, 200, 500, 600));
-s.scope.bounds_(500@600);
-s.scope.bounds_([100, 200, 500, 600]);
-s.scope.bounds_(Size(500, 600));
-::
-
         
         
 SUBSECTION:: Operation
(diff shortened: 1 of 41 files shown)


prko:
Please let me know if this PR should be split again as follows:
- add .bounds_ method to s.scope
- Change resizing behaviour

SCClassLibrary/Common/GUI/tools/Stethoscope.sc
```diff
@@ -437,6 +460,10 @@ Stethoscope {
 		);
 		^true;
 	}
+	
+	bounds_ { arg rect;
```
mtmccrea, line 464: Please [use pipes](https://github.com/supercollider/supercollider/wiki/Style-Guidelines%3A-SuperCollider#rule-use-pipes-instead-of-the-arg-keyword-to-express-parameter-lists).

mtmccrea:
I'm in favor of adding the `bounds_` method to Stethoscope, but not making modifications of the requested size based on screen bounds. 

We found previously that setting geometry based on `Window.screenBounds` can have unpredictable results when using multiple monitors (a bug you helped uncover ;)). And in general the user may have a reason for setting bounds in unusual ways, and can handle limiting to screen bounds themselves if they'd like. Open to other opinions on this though.

SCClassLibrary/Common/GUI/tools/Stethoscope.sc
```diff
@@ -75,8 +75,17 @@ Stethoscope {
...
+			[
+				Size(250,250),
+				Size(500,500)
+			]
+		} {
+			[
+				Size(parent.bounds.width, parent.bounds.height),
+				Size(parent.bounds.width * 2, parent.bounds.height * 2)
+			]
+		};
```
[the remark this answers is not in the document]
prko: I did this to avoid typing the repeated words, but I corrected it according to your advice:
https://github.com/supercollider/supercollider/pull/6281/commits/a6bbbcf942dfe979d81b40d4ffbb6d7e267888c2

prko:
Done!

Now I see the benefit of following your and @jrsurge's advice! This PR is also the best for me, as I know at a glance what I have done here!

SCClassLibrary/Common/GUI/tools/Stethoscope.sc
```diff
@@ -535,4 +538,4 @@ BusScopeSynth {
 		buffer = nil;
 		synth = nil;
 	}
-}
+}
```
mtmccrea, line 541: Please try not to modify the EOF newline, [as described here](https://github.com/supercollider/supercollider/wiki/Style-Guidelines:-SuperCollider#rule-use-exactly-one-newline-at-the-end-of-a-file).

mtmccrea:
Please add documentation for the added method.

Note for the description of the argument that:
`rect` can be a `Rect` or anything that responds to `.asRect`.

For example the scope bounds can be set with:
```supercollider
s.scope.bounds_(Rect(100, 200, 500, 600));
s.scope.bounds_([100, 200, 500, 600]);
s.scope.bounds_(Size(500, 600));
```

SCClassLibrary/Common/GUI/tools/Stethoscope.sc
prko: Done! Sorry, I knew that, but I did not know how to fix it. Anyway, I solved the problem by replacing the previous file. Then I added the `bounds_` method again.

prko:
@mtmccrea 
>Please add documentation for the added method.

Done!

Thanks for the instructions, and sorry for my clumsy process. A few days ago, I merged by clicking update on each of the outdated branches... So the histories are very dirty. In the future, I will do what you wrote on the wiki.

telephon (commented):
an object with a setter should also have a getter.

HelpSource/Classes/Stethoscope.schelp
```diff
@@ -172,6 +172,23 @@ METHOD:: style
         One of the above Integers.
 
 
+SUBSECTION:: Size and position
```
mtmccrea, line 175: This probably doesn't need it's own subsection, but can be included in the previous subsection `SUBSECTION:: Display`, above or below the documentation for the `size` method.

HelpSource/Classes/Stethoscope.schelp
```diff
@@ -172,6 +172,23 @@ METHOD:: style
...
+        
+METHOD:: bounds
+    The position and size of the window. The position is relative to the bottom-left corner of the screen.
+    argument::
+        A Rect, a Point interpreted link::Classes/Point#-asRect#.asRect:: or anything that responds to it.
```
mtmccrea, line 180: Using`Point` perhaps isn't the best example since, although that's valid, specifying a Point implicitly sets the position to `[0,0]`. So a different example would illustrate the the position _and_ size are set by `bounds_`.

Something like:
>A link::Classes/Rect::, or any object responding to the link::Search#-asRect:: method, such as an link::Classes/Array:: or A link::Classes/Size::.

mtmccrea:
> an object with a setter should also have a getter.

Good point! I noticed the method documentation listed a return value, though there isn't yet actually a getter method.

prko:
@telephon @mtmccrea 
>an object with a setter should also have a getter.
>Good point! I noticed the method documentation listed a return value, though there isn't yet actually a getter method.

I thought about it too! Thank you very much. I have done that.
I also added a message if the width of a `Rect` is less than the minimum value of 264:
https://github.com/supercollider/supercollider/pull/6281/commits/7ea3085e16c5a64f4708597fdaea1a59979d0d67

HelpSource/Classes/Stethoscope.schelp
prko: have done!
https://github.com/supercollider/supercollider/pull/6281/commits/39802c61144acfa132c8c72b41da4d118490cddb

HelpSource/Classes/Stethoscope.schelp
prko: OK, first I did what you wrote, and then I thought maybe it needed a new subsection. However, it is a bit too complicated. I changed it as you described.
https://github.com/supercollider/supercollider/pull/6281/commits/39802c61144acfa132c8c72b41da4d118490cddb

mtmccrea:
Thanks, @prko, for all your work on getting this sorted, it's looking good!

I'll leave final approval to @telephon, who started the original review.

SCClassLibrary/Common/GUI/tools/Stethoscope.sc
```diff
@@ -437,6 +437,16 @@ Stethoscope {
 		);
 		^true;
 	}
+	
+	bounds_ { arg rect;
+		if( window.notNil ) {
+			if (rect.width < 264) {
+				"The width value you set will be changed to 264, the minimum width.".postln;
```
mtmccrea, line 444: Not a request for change, but I'm wondering if we have guidance on when we do/don't post warnings or informative notes (and considering this clipped minimum is documented in the Help file). @telephon

mtmccrea (approved):

prko:
@mtmccrea thanks!!!

SCClassLibrary/Common/GUI/tools/Stethoscope.sc
telephon: Hm, there is no general guidance. Things to keep in mind: 
- if things may happen repeatedly, it is not good to repeatedly post warnings
- sometimes it is ok just not to warn and rely on an educated user

PR #6283, opened 2024-05-01
Add position arg to .meter to set s.meter position (width and height are automatically calculated)

prko:
## Purpose and Motivation

It is not currently possible to specify the position of the meter on a server instance.
For example, `s.meter` always makes its window a left bottom corner of the main screen.
With this PR it is possible to set the position of the `.meter' for a server instance.

(I use here position instead of bounds because the width and height are automatically calculated.)

**It is easier and more effective than using `meterView` in a `window' instance.**

- currently:
   ```
   (
       var numIns, numOuts;
       numIns = s.options.numInputBusChannels;
       numOuts = s.options.numOutputBusChannels; 
       w = Window("test", Rect(400, 400, ServerMeterView.getWidth(numIns, numOuts),    ServerMeterView.height));
       ServerMeterView(s, w, 0@0, numIns, numOuts);
       w.front
   )
   ```
- with this PR:
   ```
   s.meter(position: 400@400)
   ```

This PR was a part of https://github.com/supercollider/supercollider/pull/6230

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


prko:
updated `Purpose and Motivation` section

prko:
Could someone look at the practical aspect of this PR?
This PR makes code compact, as shown in the original post of this PR.

cdbzb:
agree this is a good idea!

on my system (mac  / SCNvim) the window gets put at Rect(5.0, 305.0, 248.0, 230.0) for some reason!

are there other windows/tools that lack position or bounds arguments? It would be nice if there is consistency everywhere.

cdbzb:
agree it is annoying - the default position 5@305 makes no sense on my system (SCNvim)

A simpler approach is to add a leftUp argument to ServerMeter so its signature is |server numIns numOuts leftUp| and then likewise s.meter(numIns, numOuts, leftUp)

the reason to use leftUp is that that is the argument in ServerMeterView but `position` could also be ok.

But perhaps a better alternative is to do as @mtmccrea suggested in your PR re: `plot` and add a `position_` method to ServerMeter - the rationale would be the same: what if we want to add RMS size or fall time or peak settings later, the `leftUp` arg would be in the wrong position.

mtmccrea:
> A simpler approach is to add a leftUp argument to ServerMeter so its signature is |server numIns numOuts leftUp| and then likewise s.meter(numIns, numOuts, leftUp)

I was just looking into this :) This would avoid repeated logic in both `ServerMeter:*new` and `Server:-meter` related to numIns, numOuts and width.

I'm not a big fan of `leftUp`, because, while I see the intention, my initial reading of this was that the origin of the screen coordinates is left-up, when it's actually left-down/bottom-left.

So the change would look something like:
```supercollider
ServerMeter {

	var <window, <meterView;

	*new { |server, numIns, numOuts, position = (5@305)|

		var window, meterView;

		numIns = numIns ?? { server.options.numInputBusChannels };
		numOuts = numOuts ?? { server.options.numOutputBusChannels };

			Rect(position.x, position.y, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
			false);
...
```

mtmccrea:
> But perhaps a better alternative is to do as @mtmccrea suggested in your PR re: plot and add a position_ method to ServerMeter

This was nuanced and admittedly difficult to follow in the other thread. 

The reason `position` would have to be a method argument instead of an instance method is that `s.meter` returns a `Server` instead of the `ServerMeter` instance. So one cannot call `s.meter.position_(50@50)` :-/ So we'd just have to live with this side effect:

> if we want to add RMS size or fall time or peak settings later, the leftUp arg would be in the wrong position.

And in any case, I see `s.meter` as a convenience method, so if someone is interested in modifying properties of the `ServerMeter`, like peak or smoothing settings, they may as well create a `ServerMeter` directly and interact with its (hypothetical) methods for meter settings.

telephon:
> The reason `position` would have to be a method argument instead of an instance method is that `s.meter` returns a `Server` instead of the `ServerMeter` instance.

This was a mistake, I think, all these methods should return the instance. But it is stored in `serverMeter`, so you can write:

`s.meter; s.serverMeter.position_(50@50);`

cdbzb:
"s.meter returns a Server" ...not here! 

for me `s.meter` returns `a ServerMeter` 

(mac/ARM, SCNvim)

mtmccrea:
@telephon do you think we should fix the mistake and have .meter return the ServerMeter instance? (technically a breaking change but likely low impact) This would allow setting the position to be a one-liner, then no need for the additional arg to .meter.

If you'd prefer no breaking change, do you recommend using the -serverMeter var to set position, as you showed, rather than adding a new arg to .meter?

In either case the ServerMeter:-position_ setter method would need to be added.

cdbzb:
one thought: the additional arg for `s.meter` would have tne advantage of being discoverable via the signature - if we need to write `s.meter.serverMeter.position_` we would need to find that in the docs somewhere...

telephon:
I would actually say, do both: return the meter and add an argument for bounds.

prko:
Thank you for sharing your thoughts and insights on this matter.  

I would like to raise a point that I am not entirely convinced about:  

```  
s.meter.serverMeter.position_  
```  

In my opinion, this functionality might be more appropriately achieved using:  

```  
s.meter.position_  
```  

To explore this, I conducted the following checks:  

1. Whether `<serverMeter` could be replaced with `<meter` in the `Server.sc` file.  
2. Whether there might be any issues arising from `s.meter` returning `ServerMeter`.  

Based on my testing, which involved recompiling and running the following code snippets, I did not encounter any problems:  

```  
s.meter  
```  
```  
m = s.meter(position: 600@600)  
m.position_(300@300)  
```  
```  
s.meter.position_(500@300)  
```  

If this proposed change aligns with your thoughts, I would be happy to proceed by updating the relevant documentation accordingly.

telephon:
Yes, of course – `s.meter` should return the ServerMeter instance!

telephon:
> I would actually say, do both: return the meter and add an argument for bounds.
(as I said)

prko:
@telephon 

>I would actually say, do both: return the meter and add an argument for bounds.
(as I said)

Could you please provide the code snippet that you would like to work as intended? 
Currently, the following three versions are available:

#### code 1
```supercollider
s.meter // -> a ServerMeter
```
#### code 2
``` supercollider
s.meter(position: 400@400) // -> a ServerMeter
```
#### code 3
```supercollider
s.meter.position_(500@500) // -> a ServerMeter
```
At present, `s.meter` appears as a singleton window, and evaluating Code 2 or Code 3 changes its position.

Additionally, I have implemented `bounds` as @mtmccrea suggested:
https://github.com/supercollider/supercollider/blob/1e7e0495d3bb3bfc1f358f28c1cd5aea51582ebe/SCClassLibrary/Common/GUI/tools/ServerMeter.sc#L270

I am not entirely sure what you mean by "add an argument for bounds." I apologise if I have misunderstood. If you could provide an example of the code you would like to see, I will update this PR to include the requested functionality.

Thank you!

HelpSource/Classes/Server.schelp
```diff
@@ -631,6 +631,13 @@ Show frequency analyzer window.
 method:: meter
 Show input/output level meters.
 
+argument:: numIns
+the number of input channels. The default is nil and the number of input channels defined by code::Server.local.options.numInputBusChannels:: is used when using a local server.
+argument:: numOuts
+the number of output channels. The default is nil and the number of output channels defined by code::Server.local.options.numOutputBusChannels:: is used when using a local server.
+argument:: position
+the position of the scope window. The value should be a point: x@y. The default is nil and the scope window will be in the centre of the display.
```
mtmccrea, line 639: I'd suggest capitalizing "The" at the beginning of each sentence.

> The value should be a point: x@y

As implemented the more accurate description would be:  

The value should be a point, teletype::x@y::, or respond to teletype::asPoint::.

Also consider using teletype:: or code:: when writing `nil`, as it's code.

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
```diff
@@ -263,4 +263,11 @@ ServerMeter {
 	isClosed {
 		^window.isClosed
 	}
+
+		position_ { |point|
```
mtmccrea, line 267: check the indentation here

SCClassLibrary/Common/Control/Server.sc
```diff
@@ -364,7 +364,7 @@ Server {
 	var <defaultGroup, <defaultGroups;
 
 	var <syncThread, <syncTasks;
-	var <window, <>scopeWindow, <serverMeter, <emacsbuf;
+	var <window, <>scopeWindow, <meter, <emacsbuf;
```
mtmccrea, line 367: Renaming this instance variable to `meter` and making it a getter method now conflicts with the `meter` method (which creates the meter). I'd assume the latter method takes precedence, so the getter wouldn't every actually be called, but rather a new meter would be created every time `.meter` is accessed.

prko: Yes, I have considered that as well. However, I haven’t encountered any issues after renaming `<serverMeter` to `meter` in `Server.sc` and `server-meter.sc`.

The following is the behaviour when using the original `<serverMeter` in `Server.sc` and `server-meter.sc`:

```supercollider
x = s.meter
x.serverMeter.meterView // -> a ServerMeterView
x.serverMeter.window.bounds // -> Rect(5.0, 274.0, 134.0, 230.0)
x.serverMeter.window.bounds_(Rect(0, 0, 134, 230))
x.serverMeter.close  
```

After renaming `<serverMeter` to `meter` in `Server.sc` and `server-meter.sc`, the observed behaviour is as follows:

```supercollider
x = s.meter
x.meterView // -> a ServerMeterView
x.window.bounds // -> Rect(5.0, 274.0, 134.0, 230.0)
x.window.bounds_(Rect(0, 0, 134, 230))
x.position_(200@200)
x.close
```

Although this renaming introduces a potential conflict, it is more convenient to use, In my humble opinion.

Still, if this point is crucial to my design considerations, I am inclined to revert to the original naming. Please let me know your thoughts on this matter.

HelpSource/Classes/Server.schelp
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6283/commits/0b0936e8cb1a4c201f4cd91d2876c6202619f953

I did not capitalised the `the` in the following:
>the position of the scope window. 

If you think I should capalise it as well, please let me know.

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
prko: Thanks! Done in https://github.com/supercollider/supercollider/commit/0b0936e8cb1a4c201f4cd91d2876c6202619f953

SCClassLibrary/Common/Control/Server.sc
mtmccrea: Great, this is what @telephon had originally suggested :)

prko:
Thank you, everyone, for reviewing and providing advice on the final code. While updating the documentation, I came across an issue: #6736. I believe it would be wise to review the responses on this issue before concluding the documentation. I appreciate your understanding.

SimonDeplat:
We've been discussing this PR during dev meeting and agreed that it would be nice to merge it before 3.14 release :) .

if this PR is ready, I would suggest merging it even if #6736 would require to update the documentation afterwards, so we keep going forward, and treat problems separately.

I believe checks are currently failing because `server-meter.sc` is missing an EOF newline ?

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
```diff
@@ -263,4 +263,11 @@ ServerMeter {
 	isClosed {
 		^window.isClosed
 	}
+
+	position_ { |point|
+		point !? {
```
telephon, line 268: why do you not want to catch an error if someone tries to set the point to nil?

SCClassLibrary/Common/GUI/PlusGUI/Control/server-meter.sc
```diff
@@ -1,10 +1,11 @@
...
-	meter { |numIns, numOuts|
-		if(numIns.notNil or: numOuts.notNil and: { serverMeter.notNil }) { serverMeter.close };
+	meter { |numIns, numOuts, position|
 		if(serverMeter.isNil or: { serverMeter.isClosed }) {
 			serverMeter = ServerMeter(this, numIns, numOuts)
 		} {
 			serverMeter.window.front
-		}
+		};
+		position !? {serverMeter.position_(position) };
```
telephon, line 8: add a blank space after { 
(sorry!)

prko: No problem, thanks!
Done in https://github.com/supercollider/supercollider/pull/6283/commits/f4b583e070b7af46a4346390abae65532bc7ea40

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
prko: I initially believed it would be more user-friendly for the default position, which has been in use for a long time, to be applied automatically. However, my reasoning was not entirely logical. The default position is already invoked within `.meter`, meaning that the `point` argument in the `.position` method does not need to set the default position again. Instead, the user should explicitly define a point when using `.position_`.  

I truly appreciate your guidance in helping me gain a clearer understanding. Thank you! 😊

Done in https://github.com/supercollider/supercollider/pull/6283/commits/f4b583e070b7af46a4346390abae65532bc7ea40

prko:
@SimonDeplat 
> I believe checks are currently failing because server-meter.sc is missing an EOF newline ?

Ah, thanks! I added an empty line at the end of the file.

SimonDeplat:
I don't know if your IDE could be responsible for this ?

There's an awesome tool that checks if SC formatting is respected during commit : https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit . This might save you some edits :) !

prko:
Thanks for letting me know it.

I test code in SC-IDE, then copy them into vscodium. I will try to use pre-commit.
Thanks!

prko:
My final commit is satisfactory. I have reviewed all files online, and the last line is hidden in view mode. However, in editing mode, I can see the final empty line.

HelpSource/Classes/Server.schelp
```diff
@@ -631,6 +631,38 @@ Show frequency analyzer window.
 method:: meter
 Show input/output level meters.
 
+code::
+// Create a singleton ServerMeter window at the default position if no ServerMeter window is open:
+s.meter // -> a ServerMeter
+
+// Bring the ServerMeter window to the front if a ServerMeter window is already open:
+s.meter // -> a ServerMeter
```
mtmccrea, line 639: Just one example of s.meter is enough. The comment could clarify the behavior:
```
// Create a singleton ServerMeter, or bring the window to the front if it has already been created.
```

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
```diff
@@ -263,4 +263,13 @@ ServerMeter {
...
 		^window.isClosed
 	}
+
+	position_ { |point|
+		point = point.asPoint;
+		window.bounds_(window.bounds.left_(point.x).top_(point.y)).front
+	}
+
+	position {
+		^("Position(left:" ++ window.bounds.left ++ ", bottom:" ++ (window.bounds.bottom - window.bounds.height) ++ ")")
```
mtmccrea, line 273: If you'd like to create a getter method, best to return the actual values, as a Point.

Returning formatted text isn't a typical practice for getters, and the way this is formatted creates the illusion of a Position class, which doesn't exist.

HelpSource/Classes/ServerMeter.schelp
```diff
@@ -42,13 +42,33 @@ note::
...
+m.position //-> Position(left:5.0, bottom:277.0)
+
+// Query the current window bounds
+m.window.bounds // -> Rect(5.0, 277.0, 172.0, 230.0)
+
+// Move to new position
+m.position_(0@100) //-> a ServerMeter
+
+// Retrieve the moved new window position
+m.position // -> Position(left:0.0, bottom:100.0)
```
mtmccrea, line 70: Could compress this into:
```
// Set/Get the ServerMeter's screen position by the bottom left corner of its window
m.position_(0@100);
m.position // -> Point(0.0, 100.0)
```

Also, see my comment below w.r.t. a position getter method.

HelpSource/Classes/ServerMeter.schelp
```diff
@@ -42,13 +42,33 @@ note::
...
 code::
-
 s = Server.internal; // use the internal server
 s.boot;
+
 // display 4 input channels & main stereo output
 m = ServerMeter.new(s, 4, 2);
+
+// Query the current window position
+m.position //-> Position(left:5.0, bottom:277.0)
```
mtmccrea, line 61: see my comment below w.r.t. a position getter method.

HelpSource/Classes/ServerMeter.schelp
```diff
@@ -42,13 +42,33 @@ note::
...
+
+// Query the current window bounds
+m.window.bounds // -> Rect(5.0, 277.0, 172.0, 230.0)
```
mtmccrea, line 64: I don't think we need to make the return value a comment here. It will appear in the Post window, and what you've written here will not match all cases if, e.g., someone has a different number of inputs and outputs causing the size to be different.

HelpSource/Classes/ServerMeter.schelp
```diff
@@ -42,13 +42,33 @@ note::
...
+
+// Move to new position
+m.position_(0@100) //-> a ServerMeter
+
+// Retrieve the moved new window position
+m.position // -> Position(left:0.0, bottom:100.0)
+
+// Retrieve the moved new window bounds
+m.window.bounds // -> Rect(0.0, 100.0, 172.0, 230.0)
```
mtmccrea, line 73: As above, don't include output post

HelpSource/Classes/Server.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6283/commits/4ad70d94f4ff8b4ee959c3dd6c9c60061800dc7c!

HelpSource/Classes/ServerMeter.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6283/commits/4ad70d94f4ff8b4ee959c3dd6c9c60061800dc7c!

HelpSource/Classes/ServerMeter.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6283/commits/4ad70d94f4ff8b4ee959c3dd6c9c60061800dc7c!

prko:
@mtmccrea 
Thanks!
I think I reflected your reviews. Please let me know if I misunderstood something!

telephon (approved):

HelpSource/Classes/Server.schelp
```diff
@@ -631,6 +631,35 @@ Show frequency analyzer window.
...
+s.meter.window.bounds // -> Rect(5.0, 277.0, 134.0, 230.0)
+
+// Move the open ServerMeter to the specified position:
+s.meter.position_(200@200) // -> a ServerMeter
+
+// Retrieve the new position of the ServerMeter window:
+s.meter.position // -> Position(left:200.0, bottom:200.0)
+
+// Retrieve the current bounds of the ServerMeter window:
+s.meter.window.bounds // -> Rect(200.0, 200.0, 134.0, 230.0)
```
mtmccrea, line 651: These examples should be updated the same way they were updated in `ServerMeter.schelp`.

Or, you could just link to the `ServerMeter::-position` method rather than repeat the identical examples here.

HelpSource/Classes/Server.schelp
```diff
@@ -631,6 +631,35 @@ Show frequency analyzer window.
...
+::
+
+argument:: numIns
+the number of input channels. The default is nil and the number of input channels defined by code::Server.local.options.numInputBusChannels:: is used when using a local server.
+argument:: numOuts
+the number of output channels. The default is nil and the number of output channels defined by code::Server.local.options.numOutputBusChannels:: is used when using a local server.
+argument:: position
+the position of the scope window.
```
mtmccrea, line 659: Note this is the meter window, not the scope (Stethoscope) window.

It would be helpful to explain the point of reference for the position point (bottom left of the meter window I believe?) and the origin of the screen coordinate system.

E.g. setting the position to 150@50 places the bottom left corner of the meter window 150 pixels from the left and 50 pixels up from the bottom of the screen. (I'm assuming...this may not be correct)

SCClassLibrary/Common/GUI/PlusGUI/Control/server-meter.sc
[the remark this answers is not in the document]
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6283/commits/2ff16caf6099db10f60b2d5e54dca08465badcd1!

HelpSource/Classes/Server.schelp
prko: Thank you! (no need to apologise!)

In my humble opinion, it is not so easy for beginners to think of these examples, so I have left these examples, but corrected what you mentioned.

Done in https://github.com/supercollider/supercollider/pull/6283/commits/2ff16caf6099db10f60b2d5e54dca08465badcd1!

telephon:
thanks! when you are ready, we can merge.

PR #6285, opened 2024-05-01
remove bogus argument keyword in Pfindur help file

elifieldsteel:
## Purpose and Motivation
Removed an invalid keyword from a code example in the Pfindur help file. This example included the keyword 'filling.' Evaluating this example returns a warning, and I have absolutely no idea why this keyword is included in the example.

I believe this example, in combination with the example immediately above it, are simply meant to show how Pfindur behaves when constraining an infinite-length pattern compared to a finite-length pattern.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Pfindur.schelp
+++ b/HelpSource/Classes/Pfindur.schelp
@@ -33,7 +33,7 @@ x = a.asStream;
 
 (
 var a, b;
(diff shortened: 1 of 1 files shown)


mtmccrea (approved):
Yea strange... I don't see `filling` as an arg anywhere in the project. Thanks!

PR #6289, opened 2024-05-03
Stethoscope resize behaviour change

prko:
## Purpose and Motivation
split from #6281

demo video:
https://www.dropbox.com/scl/fi/26952qepzu3iczj9bi0cq/s.bounds_-behaviour.mov?rlkey=suub8zyf6ujpksc7ugieggp02&dl=0

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/GUI/tools/Stethoscope.sc
+++ b/SCClassLibrary/Common/GUI/tools/Stethoscope.sc
@@ -75,8 +75,17 @@ Stethoscope {
 		cycle = cycleSpec.constrain(cycle_);
 		yZoom = 1.0;
 
-		smallSize = Size(250,250);
-		largeSize = Size(500,500);
+		# smallSize, largeSize = if(parent.isNil) {
+			[
+				Size(250,250),
+				Size(500,500)
+			]
+		} {
+			[
+				Size(parent.bounds.width, parent.bounds.height),
+				Size(parent.bounds.width * 2, parent.bounds.height * 2)
+			]
+		};
 
 		makeGui = { arg parent;
 			var gizmo;
@@ -85,14 +94,13 @@ Stethoscope {
 
 			if( window.notNil ) {window.close};
 
-			if( parent.isNil ) {
-				view = window = Window(
+			view = window = if( parent.isNil ) {
+				Window(
 					bounds: (smallSize).asRect.center_(Window.availableBounds.center)
-				).name_("Stethoscope");
-			}{
-				view = View( parent, Rect(0,0,250,250) );
-				window = nil;
-			};
+				)
+			} {
+				parent;
+			}.name_("Stethoscope");
 
 			// WIDGETS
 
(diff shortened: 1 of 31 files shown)


mtmccrea:
Please add a description of this PR, rather than linking to another, as this is now separate from that one.

mtmccrea:
I'll reiterate my comment from the previous PR:

IMO modifications of the requested size based on screen bounds is error prone and may surprise the user. We found previously that setting geometry based on Window.screenBounds can have unpredictable results when using multiple monitors (a bug you helped uncover, thanks ;)). And in general the user may have a reason for setting bounds in unusual ways, and can handle limiting to screen bounds themselves if they'd like. Open to other opinions on this though.

prko:
@mtmccrea 
>Please add a description of this PR, rather than linking to another, as this is now separate from that one.

Done.

>We found previously that setting geometry based on Window.screenBounds can have unpredictable results when using multiple monitors (a bug you helped uncover, thanks ;)). 

As far as I can tell, the norm for width and height is taken from the main display selected in the system settings. This means that if you press m when the stethoscope shows an expanded display, it may be in a different position to what the user expects.
I only noticed this now because the main display is the built-in display of my laptop. After reading @mtmccrea's comment, I tested it. Oh, sorry! If you think this is worse than the current behaviour, I will close this PR.

prko:
I intend to close this PR if there is no response by the end of July.

capital-G:
I'm not really into the GUI, but isn't the advantage of having the top left corner fixed that it is not possible to loose grip on a window, so closing or minimizing is always possible? That is at least on macOS, on windows this would need to be the top right corner.
If the top bar gets "out of screen" it will be hard to move or close it in any way, and this could become possible if the window becomes really large or if the screen resolution is really low.

On the other hand, I was not aware of the `m` feature at all - if someone who uses it thinks it improves the workflow, I trust their opinion and I am in favor of approving it. Or at least we could introduce a flag where this can be set?

SCClassLibrary/Common/GUI/tools/Stethoscope.sc
```diff
@@ -393,7 +397,22 @@ Stethoscope {
...
-		if( window.notNil ) { window.setInnerExtent(sz.width,sz.height) };
+		if(window.notNil) {
+			var widthIsWithinBounds = sz.width + window.bounds.left < Window.screenBounds.width;
+			var heightIsWithinBounds = window.bounds.bottom - sz.height > 0;
+			if(widthIsWithinBounds && heightIsWithinBounds) {
+				window.setInnerExtent(sz.width,sz.height);
+			} {
+				window.bounds_(
+					Rect(
+						if(widthIsWithinBounds) { window.bounds.left } { Window.screenBounds.width - sz.width },
```
telephon, line 408: Maybe check if this may become negative?
- `Window.screenBounds.width - sz.width`
- `window.bounds.bottom - sz.height`
They probably shouldn't, right?

What you do here could be a method in view, which sets the view size in a particular way. If you want to spend the time thinking this through in general, it may be used in other places as well.

Generally speaking, a lot of the GUI bounds suggestions may point toward a more general solution?

prko: >`Window.screenBounds.width - sz.width`
>`window.bounds.bottom - sz.height`

They do not become negative.

The following three screenshots summarise what happens when you press m:

- The main display is on top. 
  A and B appear as C 
 Screenshot 2024-07-07 at 19 42 47

-  The main display is in the centre.
  A -> B, C -> D, E -> F:
Screenshot 2024-07-07 at 21 38 20

-  Main display is at the bottom
  A -> B, C -> D, E -> F:
  Screenshot 2024-07-07 at 19 46 04

JordanHendersonMusic:
> IMO modifications of the requested size based on screen bounds is error prone and may surprise the user. We found previously that setting geometry based on Window.screenBounds can have unpredictable results when using multiple monitors (a bug you helped uncover, thanks ;)). And in general the user may have a reason for setting bounds in unusual ways, and can handle limiting to screen bounds themselves if they'd like. Open to other opinions on this though.

I agree with @mtmccrea here. You'd have to test multiple monitors across multiple operating systems. Managing windows in a consistent way across operating systems is a nightmare problem.

As long as we aren't moving all of the window out of bounds I don't think this is an issue. On Linuxs you can press (something like) super+mouse1 and drag the window around from anywhere inside it. I'm sure mac has a similar option for moving windows around or otherwise recovering the window when the title bar is off screen, and if it doesn't, that not *our* problem.

Issue #4607, opened 2019-10-03
thisProcess.newExecutingPath is not updated immediately from the IDE


dyfer:
## Environment

* SuperCollider version: 3.10.3
* Operating system: macOS 10.13
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce
This shows the cause of the problem:
```supercollider
// nowExecutingPath test
// 1. save this as a .scd document
// 2. run these lines separately in short succession
thisProcess.recompile;
thisProcess.nowExecutingPath.postln; //evaluate this line before recompile finishes
// -> nil

// 3. evaluate this again after the recompile is done
thisProcess.nowExecutingPath.postln; //evaluate this line before recompile finishes
// -> proper path
```
The problem can be clearly seen if the code that's run right after recompiling depends on `thisProcess.nowExecutingPath`:

```supercollider
// resolveRelative test
// 1. save this as a .scd document
// 2. run these lines separately in short succession
thisProcess.recompile;
"".resolveRelative; //evaluate this line before recompile finishes
// -> ERROR
```

## Expected vs. actual behavior

`thisProcess.nowExecutingPath` and thus `"a/path".resolveRelative` should be available to the user right after recompiling the library (or before any other code is executed from the IDE). Currently it takes a short time before that value is available. 

The case where this is problematic is when the code includes `"".resolveRelative` and the user wants to re-start the system, hitting the keyboard shortcut to recompile the library and right away attempting to run the code. This result is `ERROR: can't resolve relative to an unsaved file` and the execution stops.

brianlheim:
i don't believe this has anything to do with SCIDE, does it?

dyfer:
> i don't believe this has anything to do with SCIDE, does it?

My understanding is that `thisProcess.nowExecutingPath` is updated from the IDE - the interpreter is not running a file but rather it is used interactively, isn't it?

jamshark70:
https://github.com/supercollider/supercollider/blob/9f6bc367ba7d42002e7b3c5261e824ec3031ed07/SCClassLibrary/Common/Core/Kernel.sc#L613

```
thisProcess.nowExecutingPath = ideClass.currentPath
```

I believe the IDE issues `ScIDE.currentPath = "..."` commands directly.

In ScIDE *handshake, there's 
`this.send(\requestCurrentPath);` -- this is asynchronous so of course there is no way to make it absolutely immediate.

[...]

PR #6291, opened 2024-05-04
Process.schelp: fix outdated incorrect explanation for `thisProcess.nowExecutingPath` with examples

prko:
## Purpose and Motivation

It fixes an outdated incorrect explanation for `thisProcess.nowExecutingPath`:
>This method is supported in the SuperCollider IDE, the macOS-only SuperCollider.app, and the scel (SuperCollider-Emacs-Lisp) environment. In other editor environments, it will return code::nil::. 

`thisProcess.nowExecutingPath` also works correctly on Windows (tested with Windows 11) and Linux (tested with Ubuntu 22.04).
Also added some examples with other classes.

Some examples are based on https://scsynth.org/t/serverboot-add-and-thisprocess-nowexecutingpath/8662/7?u=prko
Thanks @jamshark70, @cdbzb and @smoge 

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Process.schelp
+++ b/HelpSource/Classes/Process.schelp
@@ -68,45 +68,33 @@ table::
 
 If a code ("fileMain.scd" in the example code below) executes another file on disk ("fileForLoad.scd" in the example code below) using link::Classes/String#-load:: or link::Classes/String#-loadPaths::, teletype::thisProcess.nowExecutingPath:: will be the location of the executed file ("fileForLoad.scd" in the example code below). If a function containing thisProcess.nowExecutingPath is called (the function defined in "fileForLoad.scd" in the example code below), teletype::thisProcess.nowExecutingPath:: will return the path of the document containing the function call ("fileMain.scd" in the example code below).
 
-strong::Steps:::
+table::
+## strong::Steps:::
 
 numberedlist::
 
-## Preparation code:
+## Preparation:
 
+numberedlist::
+## Create an SCD file called fileMain.scd in your home folder using the following block of code:
 code::
 (
-var test, fileMain, fileForLoad;
-
-test = "thisProcess.nowExecutingPath";
-~fileMainPath = "~/fileMain.scd".standardizePath;
-fileMain = File(~fileMainPath, "w");
-fileMain.write(
-	"(\n" ++
-	"(" ++ (test ++ ":").quote ++ ").postln;\n" ++
-	"(" ++ ("\tin fileMain.scd:                       ").quote + "+" + test ++ ").postln;\n" ++
-	"~test = (" ++ test ++ ".dirname +/+" + "fileForLoad.scd".quote ++");\n" ++
-	"~test.load.testFunction;\n" ++
-	"~test.openOS; // n.b.: .openDocument only works in SC-IDE\n" ++
-	"'fileMain.scd tasks finished'.postln;\n" ++
-	")"
-);
-fileMain.close;
-
-fileForLoad = File(~fileMainPath.dirname +/+ "fileForLoad.scd", "w");
-fileForLoad << (
-	"(" ++ "\tin fileForLoad.scd:                    ".quote + "+" + test ++ ").postln;\n" ++
-	"(\n" ++
-	"testFunction: {\n" ++
-	"\t(" ++ "\tin the testFunction in fileForLoad.scd:".quote + "+" + test ++ ").postln
-}\n)");
-fileForLoad.close;
+("thisProcess.nowExecutingPath:").postln;
+("	in fileMain.scd:                       " + thisProcess.nowExecutingPath).postln;
+~test = (thisProcess.nowExecutingPath.dirname +/+ "fileForLoad.scd");
+~test.load.testFunction;
+'fileMain.scd tasks finished'.postln;
 )
 ::
-
-Post window returns:
-teletype::
--> a File
+## Create an SCD file called fileForLoad.scd in your home folder using the following block of code:
+code::
+("	in fileForLoad.scd:                    " + thisProcess.nowExecutingPath).postln;
+(
+testFunction: {
+	("	in the testFunction in fileForLoad.scd:" + thisProcess.nowExecutingPath).postln
+}
+)
+::
 ::
 
 ## Execute SCD file:
@@ -114,29 +102,35 @@ teletype::
 list::
 ## In SC-IDE or other editors:
 
+numberedlist::
+## Create a new SCD file (you do not need to save it for this step), then evaluate the following code:
+
 code::
-~fileMainPath.openOS;
-// Note: .openDocument only works in SC-IDE.
-// .openOS will work in other editors and the system default application will open the SCD file.
+"~/fileMain.scd".standardizePath.openOS;
+// Note:
+// - .openDocument only works in SC-IDE.
+// - .openOS will work in other editors and
+//   the system default application will open the SCD file.
 ::
 
-Post window returns:
+## Post window returns:
 
 teletype::
--> /Users/prko/fileMain.scd
+-> /Users/_Your_Account_Folder_/fileMain.scd
 thisProcess.nowExecutingPath:
-	in fileMain.scd:                        /Users/prko/fileMain.scd
-	in fileForLoad.scd:                     /Users/prko/fileForLoad.scd
-	in the testFunction in fileForLoad.scd: /Users/prko/fileMain.scd
+	in fileMain.scd:                        /Users/_Your_Account_Folder_/fileMain.scd
+	in fileForLoad.scd:                     /Users/_Your_Account_Folder_/fileForLoad.scd
+	in the testFunction in fileForLoad.scd: /Users/_Your_Account_Folder_/fileMain.scd
 fileMain.scd tasks finished
 -> fileMain.scd tasks finished
+::
 ::
 
 ## In command line interface (CLI):
 
 code::
 // Linux:
-sclang '/home/parallels/fileMain.scd'
+sclang '/home/_Your_Account_Folder_/fileMain.scd'
 
 // macOS (/Applications/SuperCollider.app might be changed to properly):
 /Applications/SuperCollider.app/Contents/MacOS/sclang ~/fileMain.scd
@@ -151,10 +145,11 @@ teletype::
 ... (messages are truncated) ...
 *** Welcome to SuperCollider 3.13.0. *** For help press F1.
 thisProcess.nowExecutingPath:
-        in fileMain.scd:                        C:\Users\YourAccountFolder\fileMain.scd
-        in fileForLoad.scd:                     C:\Users\YourAccountFolder\fileForLoad.scd
-        in the testFunction in fileForLoad.scd: C:\Users\YourAccountFolder\fileMain.scd
+        in fileMain.scd:                        C:\Users\_Your_Account_Folder_\fileMain.scd
+        in fileForLoad.scd:                     C:\Users\_Your_Account_Folder_\fileForLoad.scd
+        in the testFunction in fileForLoad.scd: C:\Users\_Your_Account_Folder_\fileMain.scd
 fileMain.scd tasks finished
+::
 ::
 ::
 ::
@@ -194,16 +189,16 @@ list::
 teletype::
 ... (messages are truncated) ...
 localhost: keeping clientID (0) as confirmed by server process.
-- ServerBoot’s thisProcess.nowExecutingPath: /Users/prko/Library/Application Support/SuperCollider/startup.scd
-- ServerTree’s thisProcess.nowExecutingPath: /Users/prko/Library/Application Support/SuperCollider/startup.scd
+- ServerBoot’s thisProcess.nowExecutingPath: /Users/_Your_Account_Folder_/Library/Application Support/SuperCollider/startup.scd
+- ServerTree’s thisProcess.nowExecutingPath: /Users/_Your_Account_Folder_/Library/Application Support/SuperCollider/startup.scd
 Shared memory server interface initialized
 ::
 
 ## When pressing CMD/control + . after evaluating the code above:
 
 teletype::
-- CmdPeriod’s thisProcess.nowExecutingPath: /Users/prko/Library/Application Support/SuperCollider/startup.scd
-- ServerTree’s thisProcess.nowExecutingPath: /Users/prko/Library/Application Support/SuperCollider/startup.scd
+- CmdPeriod’s thisProcess.nowExecutingPath: /Users/_Your_Account_Folder_/Library/Application Support/SuperCollider/startup.scd
+- ServerTree’s thisProcess.nowExecutingPath: /Users/_Your_Account_Folder_/Library/Application Support/SuperCollider/startup.scd
 ::
 ::
 ## If there is no startup.scd:
--- a/HelpSource/Classes/Process.schelp
+++ b/HelpSource/Classes/Process.schelp
@@ -20,98 +20,15 @@ method::nowExecutingPath
 
 Usage: code::thisProcess.nowExecutingPath::
 
+Returns the full path to the file containing the code that is currently executing emphasis::interactively:: in the interpreter. Usually this is the current document. If the code block executes another file on disk, using link::Classes/String#-load:: or link::Classes/String#-loadPaths::, teletype::nowExecutingPath:: will be the location of the executed file.
 
 teletype::nowExecutingPath:: is valid only for interactive code, i.e., code files with a teletype::.scd:: extension. It does not apply to class definitions (teletype::.sc::). For that, use code::thisMethod.filenameSymbol:: or code::this.class.filenameSymbol::.
 
+This method is supported in the SuperCollider IDE, the scel (SuperCollider-Emacs-Lisp) environment and the command-line interface (CLI). In other editor environments, it will return code::nil::.
+
+See link::#examples#Examples:: for various uses of code::thisProcess.nowExecutingPath:: with link::Classes/CmdPeriod::, link::Classes/ServerBoot:: and link::Classes/ServerTree::, link::Classes/Routine:: and link::Classes/Task::.
+
 WARNING:: teletype::nowExecutingPath:: has a corresponding setter method, teletype::nowExecutingPath_::, for internal use only by the interpreter. Do not call the setter method!::
-
-Returns:: The full path to the file containing the code that is currently executing emphasis::interactively:: in the interpreter when used in the command-line interface (CLI), and IDE such as SuperCollider IDE and the scel (SuperCollider-Emacs-Lisp) environment. In some environments, it can return code::nil::
-Usually this is the current document. If a code block ("fileMain.scd" in the example code below) executes another file on disk ("fileForLoad.scd" in the example code below), using link::Classes/String#-load:: or link::Classes/String#-loadPaths::, teletype::nowExecutingPath:: will be the location of the executed file ("fileForLoad.scd" in the example code below). However, if the function is called in the executed file ("fileForLoad.scd" in the example code below), teletype::nowExecutingPath:: will be the code block ("fileMain.scd" in the example code below).
-
-code::
-(
-var fileMainPath, fileMain, fileForLoad;
-
-fileMainPath = "~/Downloads/fileMain.scd".standardizePath;
-fileMain = File(fileMainPath, "w");
-fileMain.write("(
-(" ++ "thisProcess.nowExecutingPath in fileMain.scd:".quote + "+ thisProcess.nowExecutingPath).postln;\n
-~test = (thisProcess.nowExecutingPath.dirname +/+" + "fileForLoad.scd".quote ++");\n
-~test.load.function1;\n
-~test.openOS; // because .openDocument only works in SC-IDE\n
-'fileMain.scd tasks finished'.postln;
-)");
-fileMain.close;
-
-fileForLoad = File(fileMainPath.dirname +/+ "fileForLoad.scd", "w");
-fileForLoad << ("(
-" ++ "thisProcess.nowExecutingPath in fileForLoad.scd:".quote + "+ thisProcess.nowExecutingPath).postln;\n(
-function1: {
-(" ++ "thisProcess.nowExecutingPath in the function in fileForFunctions.scd:".quote + "+ thisProcess.nowExecutingPath).postln
-}\n)");
-fileForLoad.close;
-
-fileMainPath.openOS; // because .openDocument only works in SC-IDE\n
-)
-
-/* Usage in command line interfece */
-// Linux:
-sclang '/home/parallels/Downloads/fileMain.scd'
-
-// macOS (/Applications/SuperCollider.app might be changed to properly):
-/Applications/SuperCollider.app/Contents/MacOS/sclang ~/Downloads/fileMain.scd
-
-// Windows (C:\Program Files\SuperCollider_dev\ should be changed to properly):
-"C:\Program Files\SuperCollider_dev\sclang.exe" %userprofile%\Downloads\fileMain.scd
-::
-
-teletype::
--> /Users/prko/Downloads/fileMain.scd
-thisProcess.nowExecutingPath in fileMain.scd: /Users/prko/Downloads/fileMain.scd
-thisProcess.nowExecutingPath in fileForLoad.scd: /Users/prko/Downloads/fileForLoad.scd
-thisProcess.nowExecutingPath in the function in fileForFunctions.scd: /Users/prko/Downloads/fileMain.scd
--> thisProcess.nowExecutingPath in the function in fileForFunctions.scd: /Users/prko/Downloads/fileMain.scd
-::
-
-When used in link::Classes/CmdPeriod::, link::Classes/ServerBoot:: or link::Classes/ServerTree::, it will return teletype::nil:::
-
-code::
-(
-// CmdPeriod.add {
-// ServerTree.add {
-ServerBoot.add {
-	var path = thisProcess.nowExecutingPath;
-	("\tServerBoot's thisProcess.nowExecutingPath:" + path).postln
-};
-s.reboot
-)
-
-(
-// r {
-Routine {
-	var path = thisProcess.nowExecutingPath;
-	("\tthisProcess.nowExecutingPath in path fork:" + path).postln
-}.()
-)
-
-(
-// s.waitForBoot {
-// s.doWhenBooted {
-fork {
-	var path = thisProcess.nowExecutingPath;
-	("\ts.waitForBoot's thisProcess.nowExecutingPath:" + path).postln
-};
-s.reboot;
-)
-
-(
-Task {
-	var path = thisProcess.nowExecutingPath;
-	("\tTask's thisProcess.nowExecutingPath:" + path).postln
-}.start
-// }.play
-)
-::
 
 method::startup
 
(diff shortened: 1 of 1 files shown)


prko:
The PDF of Process help document in this PR

- [Process _ SuperCollider 3.14.0-dev Help - default css.pdf](https://github.com/supercollider/supercollider/files/15206478/Process._.SuperCollider.3.14.0-dev.Help.-.default.css.pdf)

- [Process _ SuperCollider 3.14.0-dev Help _ custom.css.pdf](https://github.com/supercollider/supercollider/files/15206477/Process._.SuperCollider.3.14.0-dev.Help._.custom.css.pdf)

HelpSource/Classes/Process.schelp
```diff
@@ -24,7 +24,9 @@ Returns the full path to the file containing the code that is currently executin
 
 teletype::nowExecutingPath:: is valid only for interactive code, i.e., code files with a teletype::.scd:: extension. It does not apply to class definitions (teletype::.sc::). For that, use code::thisMethod.filenameSymbol:: or code::this.class.filenameSymbol::.
 
-This method is supported in the SuperCollider IDE, the macOS-only SuperCollider.app, and the scel (SuperCollider-Emacs-Lisp) environment. In other editor environments, it will return code::nil::.
+This method is supported in the SuperCollider IDE, the scel (SuperCollider-Emacs-Lisp) environment and the command-line interface (CLI). In other editor environments, it will return code::nil::.
```
prko, line 27: > In other editor environments, it will return code::nil::.

I have only tested it with
- SC-IDE on MacOS, Windows and Ubuntu
- Terminal on MacOS and Ubuntu
- Command Prompt on Windows

So I do not know if it is true or not.

smoge: This edition is already an enhancement since the "macOS-only SuperCollider.app" doesn't exist anymore.

cdbzb: nowExecutingPath **does** work on SCNvim on Mac! 

...how about if we write: "This method is supported in the SuperCollider IDE, SCNvim, scel, and the command-line interface (CLI). In unsupported editor environments, it will return code::nil::."

prko: @cdbzb Thanks for the suggestion!
After the following thread,

https://scsynth.org/t/what-does-thisprocess-nowexecutingpath-return-in-editor-environments-other-than-sc-ide-scel-and-terminal-command-prompt/9474/4

The following IDE's support this feature: Visual Studio Code, VSCodium and Neovim.

I am not sure, but Neovim seems to be the same as SCNvim.... Am I right?

Combining these with your idea results in:

"This method is supported in the SuperCollider IDE, SCNvim, scel, Neovim, Visual Studio Code, VSCodium, and the command line interface (CLI). In unsupported editor environments it will return code::nil::".

How about that? It is just a suggestion to reflect my little poll on the scsynth forum.

smoge: Suggestion: 

The method is supported in various environments, including the SuperCollider IDE, SCNvim, Neovim, scel (Emacs), VSCode, VSCodium, and the command line interface (CLI). In unsupported editors, it will return code::nil::.

smoge: @cdbzb You can come up with a better way to refer to the vim side. SCNvim is a "mode" or an editor?

smoge (commented):
Just a suggestion:

When a code block (such as "fileMain.scd" in the example below) executes another file on disk (e.g., "fileForLoad.scd"), by using link::Classes/String#-load:: or link::Classes/String#-loadPaths::, teletype::thisProcess.nowExecutingPath:: will point to the location of the executed file ("fileMain.scd"). However, if this function is invoked within the executed file itself ("fileForLoad.scd"), teletype::thisProcess.nowExecutingPath:: will instead reference the code block's document ("fileForLoad.scd").

HelpSource/Classes/Process.schelp
```diff
@@ -56,3 +58,235 @@ list::
...
 becomes the current Thread, with the main Thread as its parent.
+
+examples::
+
+strong::Example 1.:: Comparison of the path of the evaluated code block in a saved SCD, the loaded SCD and the function in the loaded SCD:
+
+table::
+##
+
+If a code block ("fileMain.scd" in the example code below) executes another file on disk ("fileForLoad.scd" in the example code below), using link::Classes/String#-load:: or link::Classes/String#-loadPaths::, teletype::thisProcess.nowExecutingPath:: will be the location of the executed file ("fileForLoad.scd" in the example code below). However, if the function is called in the executed file ("fileForLoad.scd" in the example code below), teletype::thisProcess.nowExecutingPath:: will be the document of the code block ("fileMain.scd" in the example code below).
```
cdbzb, line 69: I'm confused - if a function containing `thisProcess.nowExecutingPath` is called, `thisProcess.nowExecutingPath` will return the path of the document containing the function call, right? You are defining the function in fileForLoad but calling it in fileMain...

smoge: Ok, I got confused now too.

HelpSource/Classes/Process.schelp
smoge: Visual Studio Code and vscode and two different editors. Be careful not to let them confuse you.

HelpSource/Classes/Process.schelp
smoge: Maybe deserves a less confusing rewrite? Writing like your comment "path of the document containing the function call" is already more clear.

prko: Oh, sorry! You both are correct. My writing is wrong.

What I wanted to write was:

>However, if the function stored in the executed file ("fileForLoad.scd" in the example code below) is called  in another file ("fileMain.scd" in the example code below), `thisProcess.nowExecutingPath` will be the
document of the code block ("fileMain.scd" in the example code below).

How about the following?

>If a code ("fileMain.scd" in the example code below) executes another file on disk ("fileForLoad.scd" in the example code below) using `String: -load` or `String: -loadPaths`, `thisProcess.nowExecutingPath` will be the location of the executed file ("fileForLoad.scd" in the example code below). If a function containing `thisProcess.nowExecutingPath` is called (the function defined in "fileForLoad.scd" in the example code below), `thisProcess.nowExecutingPath` will return the path of the document containing the function call ("fileMain.scd" in the example code below).

prko: @cdbzb @smoge 
Thanks and sorry! You both are correct. My writing is wrong.

What I wanted to write was:

>However, if the function stored in the executed file ("fileForLoad.scd" in the example code below) is called in another file ("fileMain.scd" in the example code below), `thisProcess.nowExecutingPath` will be the document of the code block ("fileMain.scd" in the example code below).

(I seem to have left some words out while writing)


@smoge. Yes, @cdbzb's sentence is very clear! Thanks!
How about the following?

>If a code ("fileMain.scd" in the example code below) executes another file on disk ("fileForLoad.scd" in the example code below) using `String: -load` or `String: -loadPaths`, `thisProcess.nowExecutingPath` will be the location of the executed file ("fileForLoad.scd" in the example code below). If a function containing `thisProcess.nowExecutingPath` is called (the function defined in "fileForLoad.scd" in the example code below), `thisProcess.nowExecutingPath` will return the path of the document containing the function call ("fileMain.scd" in the example code below).

HelpSource/Classes/Process.schelp
prko: @smoge 
Thanks, I will use your suggestion!
>The method is supported in various environments, including the SuperCollider IDE, Neovim (SCNvim), Emacs (scel), VSCode/VSCodium, and the command line interface (CLI). In unsupported editors, it will return code::nil::.

HelpSource/Classes/Process.schelp
smoge: It works different with this example:

```
// Main.scd
thisProcess.nowExecutingPath.postln ;
~func = {thisProcess.nowExecutingPath.postln};
"~/test/secondary.scd".standardizePath.load;

// secondary.scd
thisProcess.nowExecutingPath.postln;
~func.value.postln;

// OUTPUT
/Users/smoge/test/main.scd
/Users/smoge/test/secondary.scd
/Users/smoge/test/secondary.scd
/Users/smoge/test/secondary.scd
```

smoge: It's always the path containing the function call, right?

smoge: Is it necessary to create other files in the help file to demonstrate this? I'm not sure it will just take the focus away from the main point. What do you think?

prko: @smoge 

>// Main.scd
>
>thisProcess.nowExecutingPath.postln ;
>~ func = {thisProcess.nowExecutingPath.postln};
"~/test/secondary.scd".standardizePath.load;
~func2.value;
>
>// secondary.scd
thisProcess.nowExecutingPath.postln;
{~func}.value.value.postln;
~func2 = {thisProcess.nowExecutingPath.postln};
>
>// OUTPUT
/Users/smoge/test/main.scd
/Users/smoge/test/secondary.scd
/Users/smoge/test/secondary.scd
/Users/smoge/test/secondary.scd
/Users/smoge/test/main.scd


Yes!
By evaluating the example in this PR:
- fileMain.scd:
   ```
   (
   ("thisProcess.nowExecutingPath:").postln;
   ("	in fileMain.scd:                       " + thisProcess.nowExecutingPath).postln;
   ~test = (thisProcess.nowExecutingPath.dirname +/+ "fileForLoad.scd");
   ~test.load.testFunction;
   ~test.openOS; // n.b.: .openDocument only works in SC-IDE
   'fileMain.scd tasks finished'.postln;
   )
   ```
- fileForLoad.scd:
   ```
   ("	in fileForLoad.scd:                    " + thisProcess.nowExecutingPath).postln;
   (
   testFunction: {
   	("	in the testFunction in fileForLoad.scd:" + thisProcess.nowExecutingPath).postln
   }
   )
   ```
- Post window:
   ```
   -> a File
   -> /Users/prko/fileMain.scd
   thisProcess.nowExecutingPath:
   	in fileMain.scd:                        /Users/prko/fileMain.scd
   	in fileForLoad.scd:                     /Users/prko/fileForLoad.scd
   	in the testFunction in fileForLoad.scd: /Users/prko/fileMain.scd
   fileMain.scd tasks finished
   -> fileMain.scd tasks finished
   ```

>Is it necessary to create other files in the help file to demonstrate this? I'm not sure it will just take the focus away from the main point. What do you think?

Evaluating a path in HelpBrowser returns a wrong path. To make it easy for readers to understand how it works, users should learn it by doing or by watching a video. I think including this example helps. For me, http://doc.sccode.org/Classes/File.html is a well-written example in this respect.

HelpSource/Classes/Process.schelp
```diff
@@ -56,3 +58,235 @@ list::
...
+test = "thisProcess.nowExecutingPath";
+~fileMainPath = "~/fileMain.scd".standardizePath;
+fileMain = File(~fileMainPath, "w");
+fileMain.write(
+	"(\n" ++
+	"(" ++ (test ++ ":").quote ++ ").postln;\n" ++
+	"(" ++ ("\tin fileMain.scd:                       ").quote + "+" + test ++ ").postln;\n" ++
+	"~test = (" ++ test ++ ".dirname +/+" + "fileForLoad.scd".quote ++");\n" ++
+	"~test.load.testFunction;\n" ++
+	"~test.openOS; // n.b.: .openDocument only works in SC-IDE\n" ++
```
smoge, line 90: Maybe just avoid code like .openOS that only works in scide?

HelpSource/Classes/Process.schelp
smoge: I'm not sure what this line demonstrates:

`~test = (thisProcess.nowExecutingPath.dirname +/+ "fileForLoad.scd");`

HelpSource/Classes/Process.schelp
prko: `openOS` works in the terminal and the SC-IDE. Does it not work on Emacs?

smoge: > `openOS` works in the terminal and the SC-IDE. Does it not work on Emacs?

It does something strange. If I evaluate this in Emacs, it opens SCIDE ...

smoge: I would use this method for files like PDF, but it can result in strange behaviors with scd files.

prko: This is because SC-IDE is the default application for scd document on your OS.

smoge: I would just describe the code in each file. Maybe you don't need to create two files to show.

prko:
@cdbzb  @smoge 

I revised PR reflecting what you pointed out:
[Process _ SuperCollider 3.14.0-dev Help.pdf](https://github.com/supercollider/supercollider/files/15242888/Process._.SuperCollider.3.14.0-dev.Help.pdf)

Sorry and thank you for your efforts to work with a person without sufficient English skills...

HelpSource/Classes/Process.schelp
smoge: Yes, I usually have emacs running first and open files from it. So I don't change xdg-open default.

prko: I also do not close SC-IDE. Anyway I changed that part as follows:
```
code::
 ~fileMainPath.openOS;
 // Note: .openDocument only works in SC-IDE.
 // .openOS will work in other editors and the system default application will open the SCD file.
 ::
```

smoge:
> @cdbzb @smoge
> 
> I revised PR reflecting what you pointed out: [Process _ SuperCollider 3.14.0-dev Help.pdf](https://github.com/supercollider/supercollider/files/15244165/Process._.SuperCollider.3.14.0-dev.Help.pdf)
> 
> Sorry and thank you for your efforts to work with a person without sufficient English skills...

I'm not sure about this part. Do we want to generate the source files like this? It can be confusing. 

image

prko:
@smoge 
Thanks for your opinion. Yes, it is indeed confusing. There are three reasons:

1. The use of the variable `test`:
    If I do not use `test = "thisProcess.nowExecutingPath`, the code block is as follows:
   Screenshot 2024-05-10 at 01 05 12

3. The use of `\n` with `"`:
    If I do not use  `\n` with `"`, the code block is as follows:
   Screenshot 2024-05-10 at 01 03 22

4. The use of `quote`:
   There is no way to remove `quote` as there is no `'''` and `"""` in sclang.

I can only imagine these two alternatives. If you want, I can choose one of these examples.
If not, could you suggest another example? I can also use your example.

smoge:
> I can only imagine these two alternatives. If you want, I can choose one of these examples.
> If not, could you suggest another example? I can also use your example.

I thought just writing the actual files. Not generating them. What do you think?

prko:
@smoge like this?
image

prko:
@smoge @cdbzb 
I saw @cdbzb's like emoticon, so I updated the help file!
- [Process _ SuperCollider 3.14.0-dev Help with custom css.pdf](https://github.com/supercollider/supercollider/files/15265244/Process._.SuperCollider.3.14.0-dev.Help.with.custom.css.pdf)
- [Process _ SuperCollider 3.14.0-dev Help with default css.pdf](https://github.com/supercollider/supercollider/files/15265245/Process._.SuperCollider.3.14.0-dev.Help.with.default.css.pdf)

prko:
I also feel that the last update, which reflects @smoge's suggestion, is now easier to understand.
I was obsessed with automation...

prko:
Sorry!
Closed due to incorrect touch.

HelpSource/Classes/Process.schelp
prko: Can I mark it as solved?
 
Screenshot 2024-05-10 at 08 54 23

cdbzb:
there's an interesting subtlety (at least on Mac):

if fileOne.scd contains `thisProcess.nowExecutingPath.postln` and if we create a fileTwo.scd in the same directory and execute `"fileOne.scd".load` we get `fileOne.scd` in the Post window... If we write "./fileOne.scd" we get "./fileOne.scd". In other word we get the string we called load on!

cdbzb:
I still feel this could be clearer! how about something like:

Save a file on disk called `fileOne.scd` with the following content:
```
thisProcess.nowExecutingPath
```

in the same directory save a file called `fileTwo.scd` with the following contents and evaluate each line
```
thisProcess.nowExecutingPath // full path of fileTwo.scd
"./fileOne.scd".load // ./fileOne.scd - the string we called load on
"fileOne.scd".load // fileOne.scd 
~test = { thisProcess.nowExecutingPath } //a Function
~test.() // full path of fileTwo.scd
```

now save a third file `fileThree.scd` and evaluate the following:

```
~test.() // full path of fileThree.scd 
```

in the case of functions `thisProcess.nowExecutingPath` is set when the function is evaluated, not when it is defined.

cdbzb:
OK I think we should change this behavior - I'm going to make a PR to set `.thisProcess.nowExecutingPath` to  `PathName(pathname).fullPath` instead of `pathname` in `Interpreter: executeFile`

prko:
@cdbzb 

>OK I think we should change this behavior - I'm going to make a PR to set .thisProcess.nowExecutingPath to PathName(pathname).fullPath instead of pathname in Interpreter: executeFile

Ok, then this PR is reserved until your PR is reviewed, I believe.

However, I would like to ask one question and to mention one thing:

1. >I still feel this could be clearer! if we're just documenting the (suboptimal) current behavior how about something like:
   
   This is a very good idea! Should the example added or should the example replace the old one with `FileMain` and `FileForLoad`?

2. >if fileOne.scd contains thisProcess.nowExecutingPath.postln and if we create a fileTwo.scd in the same directory and execute "fileOne.scd".load we get fileOne.scd in the Post window... If we execute "./fileOne.scd" we get ./fileOne.scd. In other word we get the string we called load on! This seems like a bug, no?
    
   Thanks for this case!
   On my machine:
   
   1. The following code
      ```
      "fileOne.scd".load
      ```
      returns:
      >file "fileOne.scd" does not exist.
      -> nil
   2. The following code
        ```
        "./fileOne.scd".load
        ```
        returns
        >file "./fileOne.scd" does not exist.
        -> nil
   3. The following code
      ```
      "fileOne.scd".loadRelative
      ```
      returns:
      >/Users/prko/fileOne.scd
      -> [/Users/prko/fileOne.scd]
   4. The following code
      ```
      "./fileOne.scd".loadRelative.class
      ```
      returns:
      >/Users/prko/./fileOne.scd
      -> Array
   5. The following code
      ```
      "fileOne.scd".resolveRelative.load
      ```
      returns:
      >/Users/prko/fileOne.scd
      -> /Users/prko/fileOne.scd
   6. The following code
      ```
      "./fileOne.scd".resolveRelative.load
      ```
      returns:
      >/Users/prko/./fileOne.scd
      -> /Users/prko/./fileOne.scd
   
   **The fourth and sixth cases are weird! As @cdbzb mentioned, this seems to be a bug in my opinion.
   The third and fourth cases are also odd. Why does `loadRelative` return an array?**
   
   Based on the test result of my laptop, the 2nd and 4th lines in `fileTwo.scd`:
   >thisProcess.nowExecutingPath // full path of fileTwo.scd
   "./fileOne.scd".load // ./fileOne.scd - the string we called load on
   "fileOne.scd".load // fileOne.scd 
   ~test = { thisProcess.nowExecutingPath } // a Function
   ~test.() // full path of fileTwo.scd
   should be changed. Also `.postln` should be added to each line as follows:   
   ```
   thisProcess.nowExecutingPath.postln; // full path of fileTwo.scd
   "./fileOne.scd".loadRelative.postln; // ./fileOne.scd - the string we called load on
   "fileOne.scd". loadRelative.postln; // fileOne.scd 
   ~test = { thisProcess.nowExecutingPath.postln; } // a Function
   ~test.() // full path of fileTwo.scd
   ```
   Also, the `fileOne.scd` should be
   ```
   thisProcess.nowExecutingPath.postln
   ```

@cdbzb 
According to your response, I will update the relevant part of this PR, although we should wait for the review of your PR related to the behaviour change of `thisProcess.nowExecutingPath`.

prko:
I think the part this PR touches should be improved, but there seems to be little interest. 
Should this PR stay open or should I close it?

@cdbzb 
Is there any update from your side?

telephon:
> 2\. Why does `loadRelative` return an array?

Because it may interpret several files, when the path is a wildcard.  `loadRelative` calls `loadPaths` which returns an array. This is more reliable, and it is hard to check, since your code may return anything, including arrays ...

telephon:
In my memory, there is indeed a bug in loadRelative, and resolveRelative (If I understand, you also found it?) This should be done in a separate issue, and probably fixed before we merge this explanation.

prko:
> there's an interesting subtlety (at least on Mac):
> 
> if fileOne.scd contains `thisProcess.nowExecutingPath.postln` and if we create a fileTwo.scd in the same directory and execute `"fileOne.scd".load` we get `fileOne.scd` in the Post window... If we execute "./fileOne.scd" we get `./fileOne.scd`. In other word we get the string we called load on! This seems like a bug, no?

@cdbzb 

I was not sure what you meant by this quote. Do you mean what I did according to your description? If I did it correctly, `.load' needs the full path of a file: [demo video](https://www.dropbox.com/scl/fi/yftarrmdcnvi9fan1k1r3/loadRelative-is-broken.mov?rlkey=ew2xmod66ilq7m07mk2jf397w&st=reahgzdw&dl=0)

However, `.loadRelative' does not need the full path of a file.

On 1 March 2025, `.loadRelative' appears to be broken.

HelpSource/Classes/Process.schelp
```diff
@@ -56,3 +58,232 @@ list::
...
+	var path = thisProcess.nowExecutingPath;
+	("- ServerBoot’s thisProcess.nowExecutingPath:" + path).postln
+};
+ServerTree.add {
+	var path = thisProcess.nowExecutingPath;
+	("- ServerTree’s thisProcess.nowExecutingPath:" + path).postln
+};
+s.reboot
+)
+::
```
telephon, line 182: maybe add a note how to remove them again?

telephon (commented):
It might also be good to point to `loadRelative` – which covers many use cases in one concise and clear expression.

HelpSource/Classes/Process.schelp
prko: ok I will do!

HelpSource/Classes/Process.schelp
prko: I thnk I can mark this as resolved.

prko:
The rendered result of https://github.com/supercollider/supercollider/pull/6291/commits/f5b2e86db146404bb263e2898e86df9e80908967:
[Integer | SuperCollider 3.14.0-dev Help.pdf](https://github.com/user-attachments/files/19082505/Integer.SuperCollider.3.14.0-dev.Help.pdf)
[Process | SuperCollider 3.14.0-dev Help.pdf](https://github.com/user-attachments/files/19082566/Process.SuperCollider.3.14.0-dev.Help.pdf)

One question:
I have included the following examples:
- `loadRelative[0]` and
- `loadPaths[0]`,

but I could not find the use case of 
- `loadRelative[1]` and
- `loadPaths[1]`.

What are their use cases?

telephon:
You could have a wildcard in the path – then loadPaths returns several results. But I would either do `[0]` or `collect`(or `do`) – because you can't rely on how many files will match.

Btw. for me, `testFunction` doesn't work.

prko:
>You could have a wildcard in the path – then loadPaths returns several results. But I would either do [0] or collect(or do) – because you can't rely on how many files will match.

Then, I think it is better not to change the code.

>Btw. for me, testFunction doesn't work.

For me, `testFunction` works on all OSes. Please watch the video:
[demo video](https://www.dropbox.com/scl/fi/ek4cjgfhyrupb715hp3vh/Screen-Recording-2025-03-07-at-05.58.04.mov?rlkey=mpno0qo7ycuzbuil2mt8ozwtr&st=db1jus6g&dl=0)

The version of SC in each OS in the above demo video:
- On Linux: [124b43e84]
- On macOS:  [91eacc8]
- On Windows: [91eacc8]

All three operating systems are running through a Parallels virtual machine.

Can anyone reproduce `testFunction` not working?

telephon:
Where is the method `testFunction` defined? In my distro, I don't seem to have it.

prko:
.testFunction` is not a method provided by sclang.
It is in the event of the `test_function.scd` file: 

https://github.com/supercollider/supercollider/blob/f5b2e86db146404bb263e2898e86df9e80908967/HelpSource/Classes/Process.schelp#L350-L364

It could be written as follows:

- at method - `.at(\testFunction)`:
    `(~path.dirname +/+ "test_function.scd").load.at(\testFunction);`
    `"test_function.scd".loadRelative[0].at(\testFunction);`
    `"~/test_function.scd".standardizePath.loadPaths[0].at(\testFunction)`

- bracket notation - `[\testFunction]`:
    `(~path.dirname +/+ "test_function.scd").load[\testFunction];`
    `"test_function.scd".loadRelative[0][\testFunction];`
    `"~/test_function.scd".standardizePath.loadPaths[0][\testFunction]`

- syntactic sugar (dot notation? I am not sure): `. testFunction`
    `(~path.dirname +/+ "test_function.scd").load.testFunction;`
    `"test_function.scd".loadRelative[0].testFunction;`
    `"~/test_function.scd".standardizePath.loadPaths[0].testFunction`

Please let me know whether I falsely used it.

telephon:
ah sure, now I remember. Thanks!

prko:
If a future PR related to #4607 and #6675 is approved, this PR should/could perhaps be updated.

JordanHendersonMusic:
Assuming [#6678](https://github.com/supercollider/supercollider/pull/6678) goes ahead, I'd not be in favour of this PR as it currently is, simply because in most cases, you actually want the path the code was written in, not where it is currently being executed. In this case, we should just recommend `thisFunctionDef.filenameSymbol`.

prko:
> Assuming https://github.com/supercollider/supercollider/pull/6678 goes ahead, I'd not be in favour of this PR as it currently is, simply because in most cases, you actually want the path the code was written in, not where it is currently being executed. In this case, we should just recommend thisFunctionDef.filenameSymbol.

If accomplishing https://github.com/supercollider/supercollider/pull/6678 does not change the current behaviour of `thisProcess.nowExecutingPath`, could I add just the following sentence in the related sections?

```
note::
It is recommended to use code::thisFunctionDef.filenameSymbol:: when the path the code was written in is required. See:  XXX
::
```

JordanHendersonMusic:
In the description, sure! But I don't think the examples will be necessary.

`nowExecutingPath` returns the path of execution, not the file path. I think most of your examples are an attempt to use it to get the file path, if there is a better way to do this, then the examples should not be included.

I don't actually know of a practical use case for `nowExecutingPath`, why would you want to know where a function was evaluated? 

Also, why is everything repeated per each operating system? If they return the same thing then there is no need to repeat it. It would be clearer if only the differences were listed. This might also help reduce the size of this pr.

cdbzb:
> I don't actually know of a practical use case for `nowExecutingPath`, why would you want to know where a function was evaluated?

I have a hotkey that brings up the file a SynthDef was defined in when hovering over its' name - I altered SynthDef.new to store that path in metadata. In one piece I display the music's source as it plays - a Song object stores the path which it was loaded from using `nowExecutingPath` and forwards as needed. There may be better ways to do these but anyhow

prko:
>I don't actually know of a practical use case for nowExecutingPath, why would you want to know where a function was evaluated?

If I recall correctly, I used `nowExecutingPath` within a function to respond to a question on a forum. Beyond that instance, and aside from creating this PR, I haven’t utilised it. I also haven’t encountered any guidance suggesting that `nowExecutingPath` shouldn’t be used within a function. Therefore, I think it would be beneficial to include examples showcasing its use.

>Also, why is everything repeated per each operating system? If they return the same thing then there is no need to repeat it. It would be clearer if only the differences were listed. This might also help reduce the size of this pr.

In my previous classes, SuperCollider and Audacity were among the leading software tools, with the majority of students primarily using Windows. Over the past decade, only two or three students have used Linux, and macOS users have been quite rare. As a result, most students are unfamiliar with paths specific to macOS or Linux environments. Based on my experience:  
- Linux users tend to possess strong technical expertise, such as building software from source, using the terminal, and managing paths.  
- macOS users exhibit varying levels of technical knowledge; some are highly proficient, while others struggle with technical concepts.  
- Windows users also display a wide range of technical abilities, from very skilled to less proficient, with the majority—except for computer science students—generally less familiar with technical tasks.  

While clear and concise documentation is undoubtedly important, it is also worth considering whether it is sufficiently user-friendly. Experienced developers often bypass documentation by directly examining method implementations. However, beginners and those encountering new concepts for the first time rely heavily on documentation to identify applicable use cases. Unfortunately, the current SuperCollider documentation lacks practical examples, leaving many users, myself included, asking similar or repetitive questions in forums. These questions are often specific to individual use cases. I believe this PR has the potential to address and reduce such recurring issues.

JordanHendersonMusic:
> I have a hotkey that brings up the file a SynthDef...

@cdbzb 

You'll also be able to jump to the line number of the line of the final `}`.



> I also haven’t encountered any guidance suggesting that nowExecutingPath 

@prko 

Assuming that PR passes - there are still kinks - the guidance will probably be not to use it... Unless you really do want the path of execution, which I don't think any of your examples show.


> unfamiliar with paths specific

I might be wrong, but isn't the only difference which way the slash goes? If so, you can just say that in one sentence and only show one path. 

Students' general tech skills, like being able to find a directory, are definitely pretty poor - frustratingly so! I remember reading an article putting it down to how they grew up with tablets and cloud storage, which shy away from directories. However, Supercollider cannot be a complete education in technological literacy, keeping track of and maintaining all the help files, tutorials, examples, etc., is already too much work. At some point, a student will have to Google a term and do their own research. Understanding paths is, in my opinion, one of those places.

telephon:
> > I don't actually know of a practical use case for nowExecutingPath, why would you want to know where a function was evaluated?
> 
> If I recall correctly, I used `nowExecutingPath` within a function to respond to a question on a forum. Beyond that instance, and aside from creating this PR, I haven’t utilised it. I also haven’t encountered any guidance suggesting that `nowExecutingPath` shouldn’t be used within a function. Therefore, I think it would be beneficial to include examples showcasing its use.

Of course! Everybody can just use whatever is suitable. I just meant that sometimes it is difficult to communicate a new way to do things because there are all the (good) examples around that would have to be updated to do so ...

prko:
@JordanHendersonMusic 

>I might be wrong, but isn't the only difference which way the slash goes? 

There is an issue regarding the formatting of file paths in Windows. Please see the examples below:

```
- in main.scd:            C:/Users/AccountFolder/main.scd
- in loaded test.scd:
  - using .load:          C:/Users/AccountFolder\test.scd
  - using .loadRelative:  C:\Users\AccountFolder\test.scd
  - using .loadPaths:     C:\Users\AccountFolder\test.scd
```

The slashes are inconsistent in these paths. While this may not directly affect functionality, it can be visually jarring and inconvenient for developers working across different platforms.
Another difficulty stems from the **Windows OS's use of drive letters**. This differs from the UNIX-style paths and can be particularly challenging for users transitioning between the two systems.
   
>Unless you really do want the path of execution, which I don't think any of your examples show.

In the process of writing this PR, I've identified `nowExecutingPath`, `load` and `loadRelative` as the only methods for obtaining the file's path within the code. I plan to adjust this PR once the new method is merged to ensure accuracy.

>Assuming that PR passes - there are still kinks - the guidance will probably be not to use it... Unless you really do want the path of execution, which I don't think any of your examples show.

My primary goal with this PR is to update outdated information and address inaccuracies in the current documentation. While I acknowledge there may still be room for improvement due to my limited technical expertise, I believe this PR, despite its kinks, represents meaningful progress compared to the existing content. I would greatly appreciate any specific feedback or suggestions to help refine and enhance it further.

JordanHendersonMusic:
Ah thanks! Now I understand.

> The slashes are inconsistent in these paths.

This is a bug and should be fixed not documented.

I don't have a windows build system setup right now, but I ***think*** this is caused by this line...

https://github.com/supercollider/supercollider/blob/3f5f5e2fb871c41b6e019e1f90c0a9feea6b3c00/editors/sc-ide/core/doc_manager.cpp#L1188

One fix might be something like...

```
ScIDE {
   currentPath_ { |p| 
      currentPath = p.replaceBackslashesIfOnWindows (method doesn't exist but you get the point)
   }
}
```
Basically, when you evaluate `thisProcess.nowExecutingPath` in the **open** file, you are getting the path straight from Qt, but when you do `load`, sclang is altering the path, something fixing it to one form of slash, sometimes mixing them both up.

telephon:
doesn't `standardizePath` do this conversion?

JordanHendersonMusic:
```
ScIDE {
   currentPath_ { |p| 
      currentPath = p.standardizePath;
   }
}
```
I won't be able to test this for a few days (if anyone else wants to, go ahead!), but if this fixes the weird path issue, it seems like a simple PR.

The cases where there are both slashes might just fix itself after this.

prko:
like this?

https://github.com/prko/supercollider/blob/87115988726480cfd6d3b053c41cb8ceb5920e0a/SCClassLibrary/scide_scqt/ScIDE.sc#L397-L399

With this, I have the following results on Windows:

----

`"~/".standardizePath.load;` 
returns
   ```
   thisProcess.nowExecutingPath:
   - in main.scd:            C:\Users\prko/main.scd
   - in loaded test.scd:
     - using .load:          C:\Users\prko\test.scd
     - using .loadRelative:  C:\Users\prko\test.scd
     - using .loadPaths:     C:\Users\prko\test.scd
   Test completed; main.scd tasks completed.
   -> Test completed; main.scd tasks completed.
   ```

----

`"~\\main.scd".standardizePath.load;`
returns
   ```
   thisProcess.nowExecutingPath:
   - in main.scd:            C:\Users\prko\main.scd
   - in loaded test.scd:
     - using .load:          C:\Users\prko\test.scd
     - using .loadRelative:  C:\Users\prko\test.scd
     - using .loadPaths:     C:\Users\prko\test.scd
   Test completed; main.scd tasks completed.
   -> Test completed; main.scd tasks completed.
   ```

----


Can I open a PR with this? Or should I change more for the first case?

JordanHendersonMusic:
> Can I open a PR with this?

Thanks so much for this testing this!!! Yes I think this looks good, it seems to fix the problem.

`C:\Users\prko/main.scd` this one line with the wrong slashes seems to be caused by `~/`. I think `standardizePath` should be replacing this but isn't. This should be a separate PR as there could easily be issues with compatibility here.

prko:
I accidentally pressed a button incorrectly, which caused this PR to become temporarily corrupt. However, @capital-G has now resolved the issue.  

We can therefore proceed once again with this PR, as well as #6678 and the other related PRs.

JordanHendersonMusic:
@prko I want to thank you for starting this because it has led to #6678 which I think actually solves the issues with `nowExecutingPath`.

However, I think this PR documents behaviour that the user should never do.

If the user wants to refer to a relative file, they should use `thisFunctionDef.filename`, as this is intended to get the real path of the file where that line of code was written. `nowExecutingPath` doesn't do this, it returns the path that is currently being executed, a subtle but important distinction. 

I don't think we should document what happens when the user uses the wrong method, instead we should swiftly redirect them towards the correct one.

Ultimately, I think this PR should be closed.
However, where this PR touches on issues regarding `standardizePath` and the oddities of windows path formatting, there could be other PRs or issues that address those specific points. 

What do you think, does this sound fair, or have I missed something?

prko:
Thank you for #6678.

Besides the detailed examples of thisProcess.nowExecutingPath, this PR also includes:
- Corrections to outdated or inaccurate information in Process.schelp
- A useful note regarding 7.exit in Integer.schelp

I believe that the valuable aspects of this PR should be retained. However, a brief warning should be included to advise users against using thisProcess.nowExecutingPath.

You are welcome to suggest which parts ought to be kept or removed. Alternatively, if it would be more efficient, you may simply remove any unnecessary sections from the PR—I do not mind who carries this out.

Of course, if, after reviewing the details, you still believe this PR should be closed, I will proceed with closing it.

Would this be suitable?

JordanHendersonMusic:
Hi @prko, I'm not quite sure what this PR is any more. Have you added examples in Process, because I can't see any.

PR #6282, opened 2024-05-01
Add bounds arg to {}.scope

prko:
## Purpose and Motivation
This PR is a part of #6230

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #6294, opened 2024-05-05
add 'bounds' arg to `.plotTree`; make `.plotTree` a single instance window; add `ServerTreeView` class

prko:
## Purpose and Motivation
Currently `s.plotTree` creates new windows whenever the shortcut is pressed or `s.plotTree` is evaluated. This behaviour is strange  because the following server-related windows are singleton windows
- `s.meter`
- `s.freqscope`
- `s.scope`.

With this PR, `s.plotTree` makes only one singleton window.
I thought this was a bug, but someone may see it as a new feature.

## Merge Note
When this PR is approved or merged, #6282 should be checked to see if it has already been approved or merged.
## Types of changes

- Bug fix or 
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


telephon:
Thank you!

The important thing in all the changes was that we still have an option to have several of them if needed (e.g. for multi-screen setups).

s.meter –> `ServerMeter`
s.freqscope -> `FreqScope`
s.scope –> `Stethoscope`

But we have no `ServerTreeView`. This would be good.
Otherwise, we have no way to have multiple `plotTree` for one server if we need it.

telephon (changes requested):
We should have a possibility to have several views if needed.

prko:
@telephon 
Thank you for your comments. This is a good idea!
I have added the feature you mentioned and the corresponding help documentation.
Will this be enough? Please let me know if I should revise the added methods and documentation or add more methods.

- PDFs of rendered help file:
  - [ServerTreeView _ SuperCollider 3.14.0-dev Help with default css.pdf](https://github.com/supercollider/supercollider/files/15213702/ServerTreeView._.SuperCollider.3.14.0-dev.Help.with.default.css.pdf)
  - [ServerTreeView _ SuperCollider 3.14.0-dev Help with custom css.pdf](https://github.com/supercollider/supercollider/files/15213704/ServerTreeView._.SuperCollider.3.14.0-dev.Help.with.custom.css.pdf)

I think what you want is the behaviour of making a new window like `ServerMeter` and `Stethoscope`. See the **Comment** below.
```
ServerTreeView.new(s)
ServerTreeView(s)
ServerTreeView(s, Rect(300, 600, 300, 600)) // // arguments: server, bounds
ServerTreeView(s, Rect(300, 600, 300, 600), 1) // arguments: server, bounds, interval
```

#### Comment
- `ServerMeter` and `Stethoscope` make always a new window:
  ```
  ServerMeter.new(s, 4, 2)
  // Evaluating this code twice will display two separate ServerMeter windows.
  ```
  ```
  Stethoscope.new 
  // Evaluating this code twice will display two separate Stethoscope windows.
  ```
- `FreqScope` makes only one window:
  ```
  FreqScope.new(400, 200, 0, server: s)
  // Evaluating this code twice will display only one FreqScope window.
  ```

prko:
PDFs of rendered help file:

- [ServerTreeView _ SuperCollider 3.14.0-dev Help _ with default css.pdf](https://github.com/supercollider/supercollider/files/15217450/ServerTreeView._.SuperCollider.3.14.0-dev.Help._.with.default.css.pdf)

- [ServerTreeView _ SuperCollider 3.14.0-dev Help _ custom css.pdf](https://github.com/supercollider/supercollider/files/15217451/ServerTreeView._.SuperCollider.3.14.0-dev.Help._.custom.css.pdf)

HelpSource/Classes/ServerTreeView.schelp
```diff
@@ -0,0 +1,55 @@
...
+Close the receiver.
+
+examples::
+
+code::
+// Start server
+s.boot;
+
+// Create a ServerTreeView:
+a = ServerTreeView.new(s)
```
telephon, line 38: either both with `.new` or none.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,23 @@
...
+		var onClose, window;
+		window = Window.new(name.asString, bounds, scroll:true).front;
+		onClose = aServer.plotTreeView(interval, window.view, { defer {window.close}; });
+		window.onClose = {
+			onClose.value;
+		};
+		^window.view.hasHorizontalScroller_(false).background_(Color.grey(0.9));
+	}
+
+	*closeAll {
```
telephon, line 12: better keep a class variable called `all` that keeps track of the existing views. Otherwise a window accidentally called "ServerTreeView" (e.g. by some existing large system) will be closed, too. And it makes docuemntation complicated (would need to go into `Window` help).

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,23 @@
+ServerTreeView {
+	*new { |aServer, bounds=(Rect(128, 64, 400, 400)), interval=0.5|
+		var onClose, window;
+		window = Window.new(name.asString, bounds, scroll:true).front;
+		onClose = aServer.plotTreeView(interval, window.view, { defer {window.close}; });
```
telephon, line 5: could you move the code from `plotTreeView` here? This would simplify the code there, you can just check if a window exists already and if not, make it an instance of `ServerTreeView`.

telephon: Also, the obscure `onClose` logic could finally go away.
To be backward compatible, `plotTreeView` could still return a function that closes the instance (although I'd prefer if it returned the instance… hm.)

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,23 @@
...
+		window.onClose = {
+			onClose.value;
+		};
+		^window.view.hasHorizontalScroller_(false).background_(Color.grey(0.9));
```
telephon, line 9: `*new` should return an instance of `ServerTreeView`, not its window.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,23 @@
...
+	}
+
+	*closeAll {
+		Window.allWindows.do{ |win|
+		if (win.name.debug == "ServerTreeView") { win.close }
+		}
+	}
+
+	close { this.close }
```
telephon, line 18: hm – sure ? this will go into a loop.

telephon (changes requested):

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
telephon: Actually, a second thought about this: it is a little bit ambivalent whether "all" is supposed to mean all views that concern a given server or all of them. So maybe just leave out `closeAll`. We don't need it that much and it makes things overly complicated. Other views like `Stethoscope`, also have no closeAll.

HelpSource/Classes/ServerTreeView.schelp
prko: OK, Done! `.new' for both.
https://github.com/supercollider/supercollider/pull/6294/commits/1bc7706c859daafa68f26caf704e501f8ad8c5d4

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: ok, Done!
https://github.com/supercollider/supercollider/pull/6294/commits/1bc7706c859daafa68f26caf704e501f8ad8c5d4

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: The first PR did not loop because it returned `Window`. However, I did it wrong, as you pointed out above. While preparing the second commit, I ran into a loop once and fixed it by rewriting all the code to move the `plotTreeView'.
https://github.com/supercollider/supercollider/pull/6294/commits/1bc7706c859daafa68f26caf704e501f8ad8c5d4

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: @telephon 
I think I can do it, could you confirm that I have understood your request correctly? Which of the two things should I do?

1. Move the `plotTreeView` method in the `Server` class to the `ServerTreeView` class. For backward compatibility, create a simplified `plotTreeView` method in the `Server` class.

2. Implement the `plotTreeView' method in the `Server' class as a function in the `.new' method of the `ServerTreeView' class. Leave the `plotTreeView` method in `Server` untouched for backward compatibility.

telephon: I meant (1), but make it backward compatible, by making the `server.plotTreeView` return a function that works like the old one ("on close").

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,17 +1,165 @@
...
 	var <window;
 	*new { |aServer, bounds=(Rect(128, 64, 400, 400)), interval=0.5|
-		var onClose, window;
+		var aServerTreeView, window;
 		window = Window.new(name.asString, bounds, scroll:true).front;
-		onClose = aServer.plotTreeView(interval, window.view, { defer {window.close}; });
-		window.onClose = {
-			onClose.value;
-		};
+		aServerTreeView = ServerTreeView.plotTreeView(aServer, interval, window, { defer { window.close } });
```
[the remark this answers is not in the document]
telephon: I'd suggest the following interface:

```supercollider
*new { |aServer, bounds=(Rect(128, 64, 400, 400)), parent|
      ^super.newCopyArgs(aServer).makeWindow(bounds, parent)
}

makeWindow { |bounds, parent|
   ... here all the window building goes ...
}

start { |interval=0.5, actionIfFail|
 .... here all the actions go, like resp building ....
}

stop {
 .... actions ar ended ...
like:
			updater.stop;
			CmdPeriod.remove(updateFunc);
			resp.free;
}
```

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -377,171 +525,17 @@ ServerTreeView {
...
 
 	plotTree { |interval=0.5|
-		var onClose;
 		if(plotTreeWindow.isNil) {
-			plotTreeWindow = Window.new(name.asString + "Node Tree",
-			Rect(128, 64, 400, 400),
-			scroll:true
-			);
-			plotTreeWindow.view.hasHorizontalScroller_(false).background_(Color.grey(0.9)).front
+			plotTreeWindow = ServerTreeView.new(this)
```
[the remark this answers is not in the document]
prko: ah, sorry, that was my mistake of carelessness

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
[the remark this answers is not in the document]
prko: To understand your suggestion:
Which of the following use cases do you have in mind?
- Use case 1
   ```
   x = ServerTreeView(s)
   x.start
   x.stop 
   // x.close // <- not provided.
   ```
- Use case 2
   ```
   x = ServerTreeView(s)
   x.start
   x.stop
   x.close
   ```
- Use case 3
   ```
   x = ServerTreeView(s)
   x.close // x.start and x.stop is automatically done
   ```

I ask this question because there is no `interval` argument in the `*new` method in your suggestion:
>*new { |aServer, bounds=(Rect(128, 64, 400, 400)), parent|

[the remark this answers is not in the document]
telephon: I would say that it is (2), and `close` will do the `stop`.
```
x = ServerTreeView(s).start;
x.close
```

So in principle one could make a user interface with a ServerTreeView without a booted server.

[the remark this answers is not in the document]
prko: It is a great idea!

Here is my implementation: https://github.com/supercollider/supercollider/pull/6294/commits/0325b39df812cfd73dac3e87738ffc67e16bf335

and here a demo video of test: https://www.dropbox.com/scl/fi/1cns3wlno0j53gg74t2cq/ServerTreeView-.plotTree-.plotTreeView.mov?rlkey=ceii91a8xnae6vwnv24ppxjkt&dl=0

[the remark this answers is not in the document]
prko: I have changed the behaviour of `s.plotTree` when it is evaluated while there is already a `s.plotTree`: https://github.com/supercollider/supercollider/pull/6294/commits/6193918ff634205a18a23ed69b95fc7b327f102e

telephon (changes requested):
looks already quite good, thank you! There are still some little things to do, but it is getting there...

telephon:
One more idea, to make multiple views even more useful: instead of passing the argument `server`, you could call it `target`. Then, the view could monitor any group in the node tree.

Within the code, you can then call `asTarget` to get the group, and `asTarget.server` to get the server.

E.g. 
```supercollider
g = Group.new(s);

g.asTarget // the group itself
g.asTarget.server // the server

```

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,197 @@
...
+		};
+		^super.newCopyArgs(server).makeWindow(bounds, parent)
+	}
+
+	makeWindow { |bounds, parent|
+		if(parent == nil) {
+			window = Window.new(this.asString, bounds, scroll:true);
+			viewParent = window;
+		} {
+			if (parent.name.contains(" Node Tree")) {
```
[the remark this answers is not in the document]
prko: OK, removed! I found another way to make `s.plotTree` a single window.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,197 @@
...
+		if(bounds.asRect.width < 400) {
+			var bnds = bounds.asRect;
+			"The width value you set will be changed to 400, the minimum width.".postln;
+			bounds = Rect(bnds.left, bnds.top, 400, bnds.height);
+		};
+		^super.newCopyArgs(server).makeWindow(bounds, parent)
+	}
+
+	makeWindow { |bounds, parent|
+		if(parent == nil) {
```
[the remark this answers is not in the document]
prko: Thanks! corrected!

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,197 @@
+ServerTreeView {
+	var <server, <bounds, viewParent, <window, resp, updateFunc, updater, tabSize = 25, treeViewStatus;
+
+	*new { |server, bounds=(Rect(128, 64, 400, 400)), parent|
+		if(bounds.asRect.width < 400) {
```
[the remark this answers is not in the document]
prko: Done! 
I also moved the bounds argument's initial value to `.makeWindow`

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,197 @@
...
+			} {
+				parent.close;
+				window = Window.new(parent.name, parent.bounds, scroll:true);
+			};
+			viewParent = window;
+			window.view.hasHorizontalScroller_(false).background_(Color.grey(0.9))
+		};
+		treeViewStatus = StaticText(window, Rect(tabSize, 0, window.bounds.width - (tabSize * 2) + 1, tabSize))
+		.string_(" STOPPED")
+		//.background_(Color.black)
```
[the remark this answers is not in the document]
prko: removed! Thanks!

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,33 +1,42 @@
...
+				"The width value you set will be changed to 400, the minimum width.".postln;
+				Rect(bnds.left, bnds.top, 400, bnds.height);
 			} {
+				bounds
+			}
+		};
+		if(parent.isNil) {
+			window = Window(this.asString, bounds, scroll:true);
+		} {
+			if(parent.view.asString == "a TopView") {
```
telephon, line 24: can you explain what this does and why this is necessary?

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,197 @@
...
+	start { |interval = 0.5, actionIfFail|
+		var done = false;
+		var collectChildren, levels, countSize;
+		var view, bounds;
+		var pen, font;
+
+		treeViewStatus.string_(" STARTED: current server tree");
+
+		pen = GUI.current.pen;
+
```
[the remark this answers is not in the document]
prko: Done!

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -37,62 +46,60 @@ ServerTreeView {
...
-						pen.fillRect(rect);
-						pen.strokeRect(rect);
-						pen.color = Color.black;
-						pen.stringInRect(
-							" " ++ node.key.asString + node.value.asString,
-							rect
-						);
-						ytabs = ytabs + 1;
-					});
+		drawFunc = {|group, xtabs = 0, ytabs = 0|
```
telephon, line 58: `xtabs = 0, ytabs = 0` are not used in the call, maybe they can be variables?

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -160,14 +167,24 @@ ServerTreeView {
...
+						} {
+							treeViewStatus.string_(" STOPPED: last updated server tree")
+						}
+					};
 					this.server.sendMsg("/g_queryTree", 0, 0);
 					interval.wait;
 				}
 			}
 		};
+		updateFunc = {
```
telephon, line 185: you can write `updater = updateFunc`

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -160,14 +167,24 @@ ServerTreeView {
...
+			updater.()
```
telephon, line 186: `updater.value;` to align styistically with line 188.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,3 +1,197 @@
...
+		pen = GUI.current.pen;
+
+		font = Font.sansSerif(10);
+
+		view = UserView.new(viewParent, Rect(0,0,400,400));
+
+		view.drawFunc = {
+			var xtabs = 0, ytabs = 0, drawFunc;
+
+			drawFunc = {|group|
```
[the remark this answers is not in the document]
prko: Oh, thank you! I did not look at the code in detail. Of course, the definition of the same function should be one shot!

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,33 +1,42 @@
...
 	var <server, <bounds, viewParent, <window, resp, updateFunc, updater, tabSize = 25, treeViewStatus;
 
-	*new { |server, bounds=(Rect(128, 64, 400, 400)), parent|
-		if(bounds.asRect.width < 400) {
-			var bnds = bounds.asRect;
-			"The width value you set will be changed to 400, the minimum width.".postln;
-			bounds = Rect(bnds.left, bnds.top, 400, bnds.height);
-		};
+	*new { |server, bounds, parent|
+		server = server ? Server.default;
```
telephon, line 5: `server  = server ?? { Server.default };` would be optimised.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -1,33 +1,42 @@
...
-		if(parent == nil) {
-			window = Window.new(this.asString, bounds, scroll:true);
-			viewParent = window;
+		bounds = if(bounds.isNil) {
+			Rect(128, 64, 400, 400)
 		} {
-			if (parent.name.contains(" Node Tree")) {
-				window = parent
+			if(bounds.asRect.width < 400) {
+				var bnds = bounds.asRect;
```
telephon, line 14: if you accept things that can be converted with `asRect`, you should do the conversion befoere the if statement in line 13.


And: `Rect` has `.width` as a method, so `.bounds` is unnecessary.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -360,166 +554,33 @@
...
+			plotTreeWindow.onClose = { plotTreeWindow = nil };
+			this.plotTreeView(interval, plotTreeWindow, { defer { plotTreeWindow.close } });
+		} {
+			bounds = bounds ?? this.plotTreeWindow.bounds;
+			bounds = detectWidth.(bounds);
+			plotTreeWindow.bounds_(bounds).front;
+		};
+	}
+	plotTreeView { |interval=0.5, parent, actionIfFail|
+		if(parent == nil) {
```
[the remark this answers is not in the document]
prko: Thanks! Done!

telephon (changes requested):
thank you very much, it is already very good! I made a few more comments.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: The part 
```
if(parent.view.asString == "a TopView") {
				var name = parent.name;
				(
					"The Window should be created with scroll argumrnt value true. e.g.:" ++
					"Window.(" ++ "title".quote ++ ", scroll: true)"
				).postln;
				window = Window(parent.name, bounds, scroll:true);
				parent.close;
			} {
				window = parent;
			}
```
is used to prevent the following errors:

```
-> a Window
ERROR: Message 'hasHorizontalScroller_' not understood.
RECEIVER:
Instance of TopView {    (0x12b7f3918, gc=34, fmt=00, flg=00, set=06)
  instance variables [35]
    qObject : RawPointer 0x6000051913e0
    finalizer : instance of Finalizer (0x12cbbd6a8, size=2, set=1)
    virtualSlots : instance of Array (0x12b94cc78, size=1, set=2)
    wasRemoved : false
    font : nil
    resize : Integer 1
    alpha : Float 1.000000   00000000 3FF00000
    decorator : nil
    layout : nil
    userCanClose : true
    deleteOnClose : true
    action : nil
    mouseDownAction : nil
    mouseUpAction : nil
    mouseEnterAction : nil
    mouseLeaveAction : nil
    mouseMoveAction : nil
    mouseOverAction : nil
    mouseWheelAction : nil
    keyDownAction : nil
    keyUpAction : nil
    keyModifiersChangedAction : nil
    keyTyped : nil
    focusGainedAction : nil
    focusLostAction : nil
    dragLabel : nil
    beginDragAction : nil
    canReceiveDragHandler : nil
    receiveDragHandler : nil
    toFrontAction : nil
    endFrontAction : nil
    onClose : nil
    onResize : nil
    onMove : nil
    window : instance of Window (0x122b0c588, size=5, set=3)
}
ARGS:
   false
PATH: /Users/prko/Dropbox/prko/__myDocs/Writings/Making Sound using Open Sources/mixed/dev - Bleeding edge/SuperCollider.app/Contents/Resources/HelpSource/Tutorials/aKoreanGuide/test.scd
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg this = <instance of TopView>
		arg selector = 'hasHorizontalScroller_'
		arg args = [*1]
	ServerTreeView:makeWindow
		arg this = <instance of ServerTreeView>
		arg bounds = <instance of Rect>
		arg parent = <instance of Window>
	Server:plotTreeView
		arg this = <instance of Server>
		arg interval = 0.5
		arg parent = <instance of Window>
		arg actionIfFail = nil
	< closed FunctionDef >  (no arguments or variables)
		var res = nil
		var code = "~v = s.plotTreeView(parent: ~u)"
		var doc = nil
	Process:interpretPrintCmdLine
		arg this = <instance of Main>
^^ ERROR: Message 'hasHorizontalScroller_' not understood.
RECEIVER: a TopView
```
However, this solution has a significant issue: the connection between the variable and its window is disconnected.
```
~u = Window();
~v = s.plotTreeView(parent: ~u)
~w = s.plotTreeView(parent: ~u)
~u.alwaysOnTop_(true)
```
Would it be better to change this part as follows (post an informative text to solve the error).?
```
if(parent.isNil) {
			window = Window(this.asString, bounds, scroll:true);
		} {
			if(parent.view.asString == "a TopView") {
				(
					"The Window should be created with scroll argumrnt value true. e.g.:" ++
					"Window.(" ++ "title".quote ++ ", scroll: true)"
				).postln;
			};
			window = parent
		};
```

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: You are correct. My mistake was due to lack of caution.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: I'm sorry, but I didn't understand it. The two options aren't working.
```
updater = updateFuncl
```
and
```
updater = updateFunc {
updater.value
}:

File: SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc (hunk 12 above)
Comment (telephon) [reply SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc:24]: Now I understand.

This is not the right way to do it (in general), because it makes the behaviour depend on the title of a window which can be anything.

To solve it, you just look what classes implement `hasHorizontalScroller_`. It is only `AbstractScroll`. 

Using `AbstractScroll.dumpClassSubtree` you can see all classes that implement it.

I think in this case it probably is `ScrollView`.

A better way is below, where the method is called, to do a check:

```supercollider
if(window.view.respondsTo(\hasHorizontalScroller_)) { window.view.hasHorizontalScroller_(false) };
```

But actually, I think this specification should not happen here at all, but where the new window is created: in line 22.

telephon: … then you can drop the whole extra branch.

prko: Thanks!

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: ok! I have done!

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: done!

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: >if you accept things that can be converted with asRect, you should do the conversion befoere the if statement in line 13.

I have moved `var bnds = bounds.asRect;` before `if(bounds.asRect.width < 400) {`.

>And: Rect has .width as a method, so .bounds is unnecessary.

Could you point out the related line?

prko:
@telephon 
Could you see my comments to the following two comments of yours? 
- https://github.com/supercollider/supercollider/pull/6294#discussion_r1605954759
- https://github.com/supercollider/supercollider/pull/6294#discussion_r1605955280

---

>One more idea, to make multiple views even more useful: instead of passing the argument server, you could call it target. Then, the view could monitor any group in the node tree.
>
>Within the code, you can then call asTarget to get the group, and asTarget.server to get the server.
>```E.g.
>
>g = Group.new(s);
>
>g.asTarget // the group itself
>g.asTarget.server // the server
>
>// server asTarget returns the default group 
>s.asTarget
>// maybe, for passing a server, we want to get the parent
>group = if(target.isKindOf(Server)) { s.defaultGroup.parent } { target.asTarget };
>
>// then you could also write
>ServerNodeTree.new; // monitors the default server and the full tree
>ServerNodeTree.new(42); // monitors group 42
>
>Ndef(\x, { SinOsc.ar(Rand(200, 800)) * 0.1 });
>Ndef(\x).fadeTime = 3;
>Tdef(\x, { loop { 1.wait; Ndef(\x).send } }).play;
>ServerNodeTree.new(Ndef(\x));
>```

It is also a great idea. But what do you mean by this?
- Create a new class `ServerNodeTree`? (I think this may be what you want)
- Implement this idea in this PR?

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
telephon: two functions `updateFunc = { someFunction.value }` is equivalent to `updateFunc = someFunction` (if the function doesn't take arguments)

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
telephon: thank you, that was the line I meant.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -21,20 +21,11 @@ ServerTreeView {
...
-				window = Window(parent.name, bounds, scroll:true);
-				parent.close;
-			} {
-				window = parent;
-			}
+			window = parent
 		};
 		viewParent = window;
-		window.view.hasHorizontalScroller_(false).background_(Color.grey(0.9));
+		if(window.view.respondsTo(\hasHorizontalScroller_)) { window.view.hasHorizontalScroller_(false) };
```
telephon, line 27: thinking about it – perhaps we don't need the test. Doesn't the view of a window always respond to `hasHorizontalScroller_` ?

telephon (changes requested):
Re-reading the code, I found that there are a few things to do, I hope this is fine for you.

- move the whole class definition to a separate file. "ServerPlusGui" implies that these are extensions only
- for now, maybe let's drop the feature of passing a parent? This can be added later. Let's just make a new window always and just pass the bounds. The current way of handling the parent is brittle. Or do you see a reason that we need that right now?
- the updater thing is still not correct, this may have been wrong already in the code you copied from the server, I am not sure. The updater is supposed to be a  Routine, resulting from the fork, but currently it is just a function. So `updateFunc.stop` never works! Line 162 should be `updater = fork { ...` without the extra function brackets. Then, still the stopping on cmdPeriod needs to be fixed. If you need support on that I'll help, but this is getting long so I'll stop here.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: Ah, ok, you mean
`updateFunc = updater` It works.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: I think we need the test.
Please watch the following video:
https://www.dropbox.com/scl/fi/mqns39chx08rbobhbyiew/hasHorizontalScroller_.mov?rlkey=7ymra5gy3nk5kae6tdo5z3kvz&dl=0

prko:
Thank you for your review!

>move the whole class definition to a separate file. "ServerPlusGui" implies that these are extensions only

Please let me know the name of the new file and where it is located. What I think is `ServerTreeView.sc` in the same folder as `ServerPlusGui.sc`. Is my thinking correct? 

>for now, maybe let's drop the feature of passing a parent? This can be added later. Let's just make a new window always and just pass the bounds. The current way of handling the parent is brittle. Or do you see a reason that we need that right now?

Do you mean the parent argument in `ServerTreeView.new` and `ServerTreeView.makeWindow`? They come from .plotTreeView:
https://github.com/supercollider/supercollider/blob/2027bac572540fc82f24306ea4f91798c795e50b/SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc#L375

Following your recommendation, I have moved some parts of `.plotTreeView` to `ServerTreeView.new` and `ServerTreeView.makeWindow`, and the `parent` argument has also been copied. Can you explain the brittleness of the parent handling? I will make it more robust.


>the updater thing is still not correct, this may have been wrong already in the code you copied from the server, I am not sure. The updater is supposed to be a Routine, resulting from the fork, but currently it is just a function. So updateFunc.stop never works! Line 162 should be updater = fork { ... without the extra function brackets. Then, still the stopping on cmdPeriod needs to be fixed. If you need support on that I'll help, but this is getting long so I'll stop here.

Um... anyway, `.stop' has worked (and sorry for not taking a closer look at the part nor writing it by myself!):

- The following video is the test with the [commit](https://github.com/supercollider/supercollider/pull/6294/commits/2342bb268e8811c285ce4ca1188a04d9d5bfeac4) you wrote [your last comment](https://github.com/supercollider/supercollider/pull/6294#pullrequestreview-2065625394):
   https://www.dropbox.com/scl/fi/rmcueze5fyyu7h4v53em8/plotTree-and-ServerTreeView-20240519.mov?rlkey=29rqlocz534jwjh2kyajxdiwf&dl=0 
   watch from 0:27

- >So updateFunc.stop never works! Line 162 should be updater = fork { ... without the extra function brackets. Then, still the stopping on cmdPeriod needs to be fixed. 
  
   Done! The following video is the test with the [commit](https://github.com/supercollider/supercollider/pull/6294/commits/44941da37083cf88205f0bbcee837d344e8fb39c) I uploaded after [your last comment](https://github.com/supercollider/supercollider/pull/6294#pullrequestreview-2065625394):
   https://www.dropbox.com/scl/fi/3jt2da0ifo76co9ogw3by/plotTree-and-ServerTreeView-2024-05-20-at-23.06.37.mov?rlkey=y75lra5wgvr3rr3onxf3wusxi&dl=0

telephon:
Sorry for not coming back to this, I was busy with outher things. Will continue later, if this is ok.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
telephon: ah good, I see. You could have told me directly :) – only subclasses of `AbstractScroll` have it …

telephon: Thinking about it: if someone hands in a parent that has horizontalScorll on, why not? I'd try not to change something from the inside.

prko: OK, I removed the line:
```
if(window.view.respondsTo(\hasHorizontalScroller_)) { window.view.hasHorizontalScroller_(false) };
```
https://github.com/supercollider/supercollider/pull/6294/commits/dd7d58932f56d008acd43ad8a3d0615593f2b486

>Thinking about it: if someone hands in a parent that has horizontalScorll on, why not? 

To avoid confusion:
I did not add the following line:
```
window.view.hasHorizontalScroller_(false).background_(Color.grey(0.9));
```
It was already there when I saw this file to edit.

prko:
@telephon 
>Thank you for your review!
>
>>move the whole class definition to a separate file. "ServerPlusGui" implies that these are extensions only
>Please let me know the name of the new file and where it is located. What I think is ServerTreeView.sc in the same folder as >ServerPlusGui.sc. Is my thinking correct?
>
>for now, maybe let's drop the feature of passing a parent? This can be added later. Let's just make a new window always and just pass the bounds. The current way of handling the parent is brittle. Or do you see a reason that we need that right now?

I have split `ServerTreeView` from `ServerPlusGui.sc` and created `ServerTreeView.sc` in the same folder.

prko:
I have moved _adding bound argument to `plotTree`_ here. It was originally in #6282. I hope this will speed up the review process for both PRs.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerTreeView.sc
```diff
@@ -0,0 +1,204 @@
...
+			}
+		};
+		if(parent.isNil) {
+			window = Window(this.asString, bounds, scroll:true);
+		} {
+			window = parent
+		};
+		viewParent = window;
+		//if(window.view.respondsTo(\hasHorizontalScroller_)) { window.view.hasHorizontalScroller_(false) };
+		window.view.background_(Color.grey(0.9));
```
telephon, line 28: the background should probably be in the view, not in the window. Maybe the UserView needs the background color?

All calls to the window that change its state should only happen if the window as locally created here. Otherwise, we want the responibility elsewhere. So `front` should be in line 22, I think.

Imagine you want to build a GUI and insert this view into it. Then suddenly your window comes to the front, which you may not have intended.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -360,166 +360,29 @@
...
-		updateFunc = {
-			updater = fork {
-				loop {
-					this.sendMsg("/g_queryTree", 0, 0);
-					interval.wait;
-				}
+	plotTree { |interval=0.5, bounds|
+		var detectWidth = { |thisRecBounds|
+			if(thisRecBounds.asRect.width < 400) {
+				var bnds = thisRecBounds.asRect;
```
telephon, line 366: here, there are similar calculations as needed in the other merge request. Maybe you can unify them?

telephon (changes requested):
Just a little change. The original code was a bit convoluted already, so in the future maybe we can improve that. For now, fine.

The size correction stuff should perhaps be done once and correctly, so we don't have to rethink it each time. You could remove it here and add it later.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: I believe you are referring to the following line that I introduced:  
https://github.com/supercollider/supercollider/blob/ea98edddb71d2baafad0adee6d0a6df6b45690cb/SCClassLibrary/Common/GUI/tools/Stethoscope.sc#L441-L451

However, in `bounds_` for `Stethoscope.sc`, I did not set any values, and the size remains unspecified.  
The size of `Stethoscope` is complex and appears to be significantly influenced by several factors. eg:
https://github.com/supercollider/supercollider/blob/ea98edddb71d2baafad0adee6d0a6df6b45690cb/SCClassLibrary/Common/GUI/tools/Stethoscope.sc#L394-L397

There are seven cases that determine the size of the bounds, primarily affecting the `width`:
```
SCClassLibrary/Common/Geometry/Rect.sc:
  104  		var pos, center;
  105: 		if (width > height, {
  106  			pos = (width - height) * 0.5 + left;

SCClassLibrary/Common/GUI/Base/QRangeSlider.sc:
  11  			r = bounds.asRect;
  12: 			if( r.width > r.height ) {
  13  				this.orientation_( \horizontal );

SCClassLibrary/Common/GUI/Base/QSlider.sc:
  38  			r = bounds.asRect;
  39: 			if( r.width > r.height ) {
  40  				this.orientation_( \horizontal );

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc:
  146  			totalHorizPad = xLeftPad.max(borderMargin) + xRightPad.max(borderMargin);
  147: 			wdAllowsXLabels = viewRect.width >= (totalHorizPad * (hideLabelsWidthRatio+1));
  148  			sizeAllowsXLabels = wdAllowsXLabels;

  188  			totalHorizPad = leftPad + rightPad;
  189: 			wdAllowsYLabels = viewRect.width >= (totalHorizPad * (hideLabelsWidthRatio+1));
  190  

SCClassLibrary/JITLib/GUI/NdefGui.sc:
  102  		var numLines = 1;
  103: 		if (width > 500) { // two lines, break them up near the middle
  104  			numLines = 2;

SCClassLibrary/Common/GUI/tools/Stethoscope.sc:
  441	bounds_ { arg rect;
  442  		if( window.notNil ) {
  443: 			if (rect.width < 264) {
  444  				"The width value you set will be changed to 264, the minimum width.".postln;
  445			};
  446			window.bounds_(rect.asRect)
  447		};
  448	}
  449	bounds { ^this.window.bounds }
  450	}
```
I believe I could implement an instance method for `Rect` that would accommodate many of the cases within these seven lines of code. If you deem this necessary, I am happy to do so after this PR has been merged. However, I think it would be best to leave this part unchanged, as modifying it at this stage would likely delay the completion of this PR.

telephon: After 9 month, I just don't remember to what other merge request of yours I was referring, I should have been more explicit in my comment. Certainly, the other meothods you are referring to are not relevant to what I meant.

So yes, keep this part for now, I suppose.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerTreeView.sc
prko: I sincerely appreciate your patience and guidance. I have endeavoured to implement the changes you reviewed, as per the following commit:

https://github.com/supercollider/supercollider/pull/6294/commits/271df8bbe2a1e4ebd7e7be7d5ad2516682b4db63

I would be grateful if you could confirm whether I have correctly understood your suggestions or if I have made any mistakes in the process. 

Apologies for the delayed response, and thank you again for your time and support.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
prko: I believe I have understood what you meant at the time.  

There are three identical code blocks in this PR:  
- [ServerTreeView.sc (Line 14)](https://github.com/prko/supercollider/blob/503e541399fcab8c688d67b1ecd2a029d7f6c5fe/SCClassLibrary/Common/GUI/PlusGUI/Control/ServerTreeView.sc#L14)
- [ServerPlusGUI.sc (Line 366)](https://github.com/prko/supercollider/blob/503e541399fcab8c688d67b1ecd2a029d7f6c5fe/SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc#L366)
- [ServerPlusGUI.sc (Line 372)](https://github.com/prko/supercollider/blob/503e541399fcab8c688d67b1ecd2a029d7f6c5fe/SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc#L372)]

To streamline their usage, I have introduced a new method for `Rect`:  
[Rect Method (Lines 207–219)](https://github.com/prko/supercollider/blob/503e541399fcab8c688d67b1ecd2a029d7f6c5fe/SCClassLibrary/Common/Geometry/Rect.sc#L207-L219)  

Please let me know if this aligns with your expectations or if further adjustments are needed.

prko:
This is the current status of this PR:
- [Demo video](https://www.dropbox.com/scl/fi/j056iorouc6123y8e0vjv/demo-s.plotTree-ServerTreeView.mov?rlkey=ab6nkca8dm4tjs6es5f6p551e&st=h38y3do9&dl=0)
- [Server _ SuperCollider 3.14.0-dev Help.pdf](https://github.com/user-attachments/files/19931016/Server._.SuperCollider.3.14.0-dev.Help.pdf)
- [ServerTreeView _ SuperCollider 3.14.0-dev Help.pdf](https://github.com/user-attachments/files/19931109/ServerTreeView._.SuperCollider.3.14.0-dev.Help.pdf)

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
telephon: Ah nice, now I see. 

- What about dropping the warnings? They may cause troubles and are not so very necessary, maybe.
- I would find it good if the method were not a setter and if it returned a new Rect.
- call `asSize` on `minSize`, then you can also write something like 500 @ 700
- instead of `bounds ? Rect(128, 64, 400, 400);` always `bounds ?? { Rect(128, 64, 400, 400) }`. This saves a little electricity.
- The else branch in https://github.com/prko/supercollider/blob/503e541399fcab8c688d67b1ecd2a029d7f6c5fe/SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc#L371C4-L373C41 should be left out, perhaps? We don't want to enforce the limit if someone has consciously resized the view?

prko: Done with https://github.com/supercollider/supercollider/pull/6294/commits/a542bdf3e5d6357a88e7441ceef16982ceee6029!
Thanks!

prko:
Now, `bounds` argument works again, but it does not change the size even the size is smaller than `395@386`.

telephon:
sorry, it's been a while. I think this would be the right way to do it:

```supercollider
	plotTree { |interval=0.5, bounds|
	
	if(plotTreeWindow.isNil) {
		bounds = bounds ?? { Rect(128, 64, 400, 400) };
		bounds = bounds.minSize(395@386);
		plotTreeWindow = Window(name.asString ++ " Node Tree", bounds, scroll:true);
		plotTreeWindow.onClose = { plotTreeWindow = nil };
		this.plotTreeView(interval, plotTreeWindow, { defer { plotTreeWindow.close } });
	} {
		if(bounds.notNil) { 
			bounds = bounds.minSize(395@386);
			plotTreeWindow.bounds_(bounds) 
		};
	};
	
	plotTreeWindow.front;
}
```

What do you think?

prko:
Yes, it is good, Thanks!

telephon (approved):

telephon:
The new structure is good. Trying this, @prko I found that we have lost a few things in this new system which should be added in. 

Screenshot 2025-05-09 at 23 32 35

The node ids for groups are not visible anymore, and also the nesting structure. Would you like to help by adding it back in?

prko:
Thank you for bringing this to my attention, and I sincerely apologise for any inconvenience I caused.  

I believe I have identified the issue and resolved it in #6834.  

I have also attached a demonstration video for your reference.  

Please do let me know if there is anything else that requires further refinement or updates.

PR #6296, opened 2024-05-05
ServerPlugins: PanAZ: fix bug when width is small and negative

### LINKED ISSUE #6012: PanAz.ar crashes server if width is negative
PanAz.ar crashes server if width is modulated with a negative number, with the following error:
```
```
Minimal example to reproduce:
```
(
{
	10.do({ arg i;
		var snd = SinOsc.ar(1+(i*5));
		PanAz.ar(
			4,
			snd,
			0,
			1,
			LFNoise2.kr(0.1 + (i*2)).range(-2,2)
		).poll(2);
	})
}.play
)
```

JordanHendersonMusic:
## Purpose and Motivation

When width is negative in `PanAz` the server crashes #6012 
```supercollider
{PanAz.ar(2, DC.ar, 0, 1, -0.001)}.play
```

Since width is a magnitude, it makes no sense for it to be negative. Taking the abs is an easy fix for #6012

Fixes #6012

## Types of changes

- Bug fix
- Docs

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/HelpSource/Classes/PanAz.schelp
+++ b/HelpSource/Classes/PanAz.schelp
@@ -38,9 +38,11 @@ A control rate level input.
 
 argument::width
 
-The width of the panning envelope. Nominally this is code::2.0:: which pans between pairs of adjacent speakers.
+The width of the panning envelope.
+Nominally this is code::2.0:: which pans between pairs of adjacent speakers.
 Values greater than code::2:: will spread the pan over greater numbers of speakers.
 Values less than code::1:: will leave silent gaps between speakers.
+The absolute of this value is taken as negative values have no meaning (a width of code::-1:: is the same as a width of code::1::).
 
 
 argument::orientation
--- a/server/plugins/PanUGens.cpp
+++ b/server/plugins/PanUGens.cpp
@@ -1302,7 +1302,7 @@ void PanAz_Dtor(PanAz* unit) {
 void PanAz_next_ak(PanAz* unit, int inNumSamples) {
     float pos = ZIN0(1);
     float level = ZIN0(2);
-    float width = ZIN0(3);
+    float width = std::abs(ZIN0(3));
     float orientation = ZIN0(4);
 
     int numOutputs = unit->mNumOutputs;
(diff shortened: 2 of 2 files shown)


smoge:
Maybe a warning since the input makes no sense?

JordanHendersonMusic:
> Maybe a warning since the input makes no sense?

Actually I think it would be better with a note in the docs, --- the warning would have to come from the server

adcxyz (approved):
Brilliant that you fixed a server-crashing bug! 
Please add a note to PanAz help that width is taken as absolute value, 
so negative values are mirrored into positive. 
Thanks!
Also,triggering a warning seems too much: 
Since width can be a UGen, that can create lots of messages;
and as the implementation now has well-defined behavior, 
a note in the help seems sufficient.

telephon:
>Since width is a magnitude, it makes no sense for it to be negative. 

I think "width" is a metaphor, and may be interpreted in different ways. It may be taken as the vector that describes the relation between the left and the right pan direction.

Just an idea: wouldn't it make sense to say that a negative width turns around pan direction? This seems to be better than just taking the absolute value.

JordanHendersonMusic:
I should note that I never bothered to investigate why small negative values caused a crash.

@adcxyz just updated the docs. 

@telephon I don't really understand what you mean by turn around pan position?

For a 16 channel ring with the first channel directly centre ahead:

this is position 7 and a width of 1.
pos_7_width1

this is position 7.5 and a width of 1.
pos_7 5_width1

this is position 7.5 and a width of 1.00001.
pos_7 5_width1 00001

A width of one specifies an infinitesimal (ignoring rounding errors and computer limitations)  dead zone between each speaker. Widths less than one increase the size of that dead zone. Widths more than one spread the sound across multiple speakers.

This is a pos of 8 and a width of 0.00001
pos_8_width0 000001

if you offset the position even slightly, to say 8.00001, it falls to silence.

If you set the width to zero and position to a whole number, it falls silent. This one might actually be a bug/rounding error.

A width of zero seems like it ought to mean the sound must be perfectly aligned with the speaker or there will be silence, in practice, due to rounding errors I imagine, this always outputs silence.


A negative value could mean mirror the channels about the pan position, but since `PanAz` only works on mono sources, this is equivalent to taking the absolute value.

telephon:
> @telephon I don't really understand what you mean by turn around pan position?

ah yes, sorry, this would only work for `SplayAz`. Ignore :)

adcxyz:
thanks @JordanHendersonMusic !

PR #6305, opened 2024-05-10
SCDocHTMLRenderer: Remove script referring to non-existing files from SCDoc HTML

prko:
## Purpose and Motivation

I found unnecessary parts in SCDocRenderer and marked them as comments for future development. The commented lines create scripts that refer to the non-existing files. It only makes errors when loading the SC help documentation HTML file.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/SCDoc/SCDocRenderer.sc
+++ b/SCClassLibrary/SCDoc/SCDocRenderer.sc
@@ -246,7 +246,7 @@ SCDocHTMLRenderer {
 		<< "var scdoc_sc_version = '" << Main.version << "';\n"
 		<< "</script>\n"
 		<< "<script src='" << baseDir << "/scdoc.js' type='text/javascript'></script>\n"
-		// << "<script src='" << baseDir << "/docmap.js' type='text/javascript'></script>\n" // FIXME: remove?  // does not exist
+		<< "<script src='" << baseDir << "/docmap.js' type='text/javascript'></script>\n" // FIXME: remove?
 		// << "<script src='" << baseDir << "/frontend.js' type='text/javascript'></script>\n" // does not exist
 		// QWebChannel access
 		// << "<script src='qrc:///qtwebchannel/qwebchannel.js' type='text/javascript'></script>\n" // does not exist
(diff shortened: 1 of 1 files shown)


smoge:
docmap.js is supposed to be generated

https://github.com/supercollider/supercollider/blob/8dc8241d21d37c40424f32ddec3f879127aa6c8c/SCClassLibrary/SCDoc/SCDoc.sc#L456C2-L525C1

(I'm not so sure about the other two)

prko:
>docmap.js is supposed to be generated
@smoge 
Yes, you are correct!
I have uncommented it!
Thanks!

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -247,9 +247,9 @@ SCDocHTMLRenderer {
 		<< "</script>\n"
 		<< "<script src='" << baseDir << "/scdoc.js' type='text/javascript'></script>\n"
 		<< "<script src='" << baseDir << "/docmap.js' type='text/javascript'></script>\n" // FIXME: remove?
-		<< "<script src='" << baseDir << "/frontend.js' type='text/javascript'></script>\n"
+		// << "<script src='" << baseDir << "/frontend.js' type='text/javascript'></script>\n" // does not exist
 		// QWebChannel access
-		<< "<script src='qrc:///qtwebchannel/qwebchannel.js' type='text/javascript'></script>\n"
+		// << "<script src='qrc:///qtwebchannel/qwebchannel.js' type='text/javascript'></script>\n" // does not exist
```
capital-G, line 252: This exists within the Qt IDE as far as I am aware but I don't know for what, so its best to keep it in place.

capital-G (changes requested):
Thanks for the PR :) There are some minor things which should be taken care of to make it the intention of the PR more consistent.

capital-G:
After using git blame on the LOC in question, I think it is better to object to the deletion, see

https://github.com/supercollider/supercollider/commit/227b1b61f7ee0bde44268d911c3439303a0ebf0c

> This corresponds to the existing frontend.css file, which is used for IDE or platform specific overrides. This commit also explicitly loads frontend.css in some html files where it was not being done before.

This may be useful for the vscode integration of SC? Only a guess.

Would you still insist on the deletion in this case @prko ?

SCClassLibrary/SCDoc/SCDocRenderer.sc
prko: Thanks for letting me know it!

prko:
@capital-G @smoge 
Thanks for letting me know that none of these three scripts are useless.
Since I made this PR from my insufficient knowledge, it is better to close.

PR #6310, opened 2024-05-13
Ndef gui help overflow

### LINKED ISSUE #5517: NdefParamGui not detecting overflow (and thus not showing scrollbar in Ndef's gui)
## Environment

* Operating system: Arch Linux OS
* Other details (Qt version, audio driver, etc.): scnvim

## Steps to reproduce

```supercollider
// Create a synth with 150 frequency params
Ndef(\1, {SinOsc.ar(freq: 150.collect{|i| "freq%".format(i).asSymbol.kr(100, spec: Spec.specs[\freq]) })}).gui
```
results in all 150 sliders being showcased at the same time, no overflow detected by NdefParamGui:

2021-07-21-17:19:41-screenshot

This has been tested on BSPWM and I3, two popular tiling window managers running under X11 as well as MacOS (running 3.12.0 thanks @redFrik).

I am not sure but I think the problem is somewhere in NdefParamGui and related to the overflow variable never being able to trigger the EZController to set visibility to true.

This is where overflow is set:
https://github.com/supercollider/supercollider/blob/de60e3405b417779bd35c19add1c6b2cca3bd194/SCClassLibrary/JITLib/GUI/NdefParamGui.sc#L80

And this is where it should be triggered to set `.visible` to true to show the scroll bar instead of showing all 150 sliders:
https://github.com/supercollider/supercollider/blob/de60e3405b417779bd35c19add1c6b2cca3bd194/SCClassLibrary/JITLib/GUI/NdefParamGui.sc#L108-L115

adcxyz:
This PR fixes #5517 by explaining NdefGui usage for a border case:
When creating an NdefGui for an Ndef with very many parameters,
it may create more paramViews than can fit on the window or screen, 
which makes the parameters seem inaccessible by scrolling. 
In such cases, it is best to set numItems so they fit into the available 
screen space. 

## Types of changes
- Documentation

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/NdefGui.schelp
+++ b/HelpSource/Classes/NdefGui.schelp
@@ -30,7 +30,7 @@ In such cases, it is best to set a reasonable maximum for numItems:
 code::
 // an Ndef with 100 controls
 Ndef(\1, {
-	SinOsc.ar(freq: 150.collect{|i|
+	SinOsc.ar(freq: 100.collect{|i|
 		"freq%".format(i).asSymbol.kr(100, spec: Spec.specs[\freq])
 	})
 });
--- a/HelpSource/Classes/NdefGui.schelp
+++ b/HelpSource/Classes/NdefGui.schelp
@@ -24,6 +24,26 @@ Ndef(\a, { |freq=300, dens=20, amp=0.1, pan|
 )
 ::
 
+Note that when creating an NdefGui for an Ndef with very many parameters,
+it may create more sliders than can fit on the window or screen.
+In such cases, it is best to set a reasonable maximum for numItems:
+code::
+// an Ndef with 100 controls
+Ndef(\1, {
+	SinOsc.ar(freq: 150.collect{|i|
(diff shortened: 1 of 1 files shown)


madskjeldgaard:
This looks good, thanks @adcxyz !

mtmccrea:
It's good that this was noted in the `NdefGui` help file, but it should probably also be added to the documentation of both `NodeProxy:-gui` and `NDef:-gui`, which are both unfortunately undocumented, and are more likely to be how the user encounters this problem.

Docs aside, this does not fix #5517, it just explains why it's a problem. IMO fixing that issue requires catching the overflow and making the scroll view visible, as @madskjeldgaard helpfully described.

madskjeldgaard:
https://github.com/madskjeldgaard/nodeproxygui2

For reference, we fixxed this and other issues in this repo. @redFrik @salkin-mada

PR #6313, opened 2024-05-17
Give volume permanent group and nodeID 

### LINKED ISSUE #4157: give server volume a permanent nodeID
Proposed Feature: 
Server volume should be transparently visible and (if desired) settable from multiple clients. 
This is useful for all network music setups and installations which use shared servers, e.g. in our current case,  a group of raspis all running SC. 

To make this possible, the nodeID for ```server.volume.ampSynth``` must be known to all clients;
then they can request the control value from the server with ```s.volume.ampSynth.getMsg``` etc. 

IMO this nodeID should be known by design - simplest option: always reserve the first permanent nodeID for it, and keep it forever, e.g. when remaking ampSynth after CmdPeriod etc. 
This would have a side benefit of make  ampSynth maintenance simpler and more robust.  

Does anyone see any potential problems with that? 
The only issue I can think of is that users may have been reserving the first permanent nodeID for other things already, and that this would potentially conflict. 

I'll make a test branch for it to see if any problems arise when implementing it.

adcxyz:
## Purpose and Motivation

Setting volume on the server can be made more robust and easier to reason about:
By giving it a known permanent nodeID, 2, and creating a permanent synth object for it, 
there are much fewer possibilities for the volume state to get out of sync. 

A second advantage is with remote servers as used in network music: 
All remote clients will be able to use the same single volume synth to get and set main volume.

Currently, multiple clients form a chain, where each one has its own volume, 
which also sets volume of earlier clients' groups: 
```
// node tree
defaultgroup client2
 ... client 2 synths ...
 client2 volume
defaultgroup client1
 ... client 1 synths ...
 client1 volume
1 - defaultgroup client0 (usually local client)
  ... client 0 synths ...
 client 0 volume
```
With this PR, the structure is: 
```
// node tree
defaultgroup client2
 ... client 2 synths ...
defaultgroup client1
 ... client 1 synths ...
1 - defaultgroup client0 (usually local client)
  ... client 0 synths ...
 client 0 volume, get- and settable by all clients eventually.
```
This PR is the first infrastructural step toward this goal. 

This PR takes up #4985, see discussion there, 
and closes #4157.

## Types of changes
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -4,14 +4,22 @@ Volume {
...
 	var <>window;
+	var <group, baseAmpSynth;
 
 	*new { | server, startBus = 0, numChannels, min = -90, max = 6, persist = false |
 		^super.newCopyArgs(server ?? { Server.default }, startBus, numChannels, min, max, persist).init;
 	}
 
 	init {
+		// group nodeID should always be 2, so multiple clients can use it
+		group = Group.basicNew(server, server.nextPermNodeID);
```
telephon, line 17: It may be that `server.nextPermNodeID` doesn't return the nodeID of 2. It depends on something rather brittle, namely on the call order for the server tree.

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -4,14 +4,22 @@ Volume {
...
+		defName = (\volumeAmpControl ++ this.numChannels).asSymbol;
+		// synth nodeID should always be 3, so multiple clients can use it
+		baseAmpSynth = Synth.basicNew(defName, server, server.nextPermNodeID);
```
telephon, line 20: same here, not necessarily 3.

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -4,14 +4,22 @@ Volume {
 
 	var <volume = 0.0, <lag = 0.1, <isMuted = false;
 
-	var <ampSynth, <numOutputChannels, defName, updateFunc, initFunc;
+	var <ampSynth, <numOutputChannels, <defName, updateFunc, initFunc;
 	var <>window;
+	var <group, baseAmpSynth;
```
telephon, line 9: "base amp" sounds like the amplitude of the base. 

It is a little unclear what the difference between `ampSynth` and `baseAmpSynth` is...

SCClassLibrary/Common/Control/Volume.sc
adcxyz: I think the numbers are fixed and safe: 
I am making them correctly initially, and then keeping them forever:
Server:new creates its volume object in the init method, 
and Volume:init calls s.nextPermID twice, getting 2 and 3. 

But when the server boots, it resets nodeAllocator perm to 2, 
so the next two calls to it get 2 and 3 when they should be reserved. 

Mazbe the simplest fix: 
hardcode numbers 2 and 3 for volume group and synth,
and reset nodeAllocator perm to 4. 
yes?

SCClassLibrary/Common/Control/Volume.sc
adcxyz: baseAmpSynth should maybe be called basicAmpSynth or permAmpSynth:
It is only made once, and sends its newMsg to the server when needed. 
ampSynth is either nil when baseAmpSynth is not running, 
or is assigned to baseAmpSynth when running. 

could also be a flag ampSynthRunning, would that be clearer?

SCClassLibrary/Common/Control/Volume.sc
telephon: yes!

SCClassLibrary/Common/Control/Volume.sc
telephon: @adcxyz Ah now I understand, `ampSynth` becomes `baseAmpSynth` when the volume is running. So I suppose it would be possible to have only `ampSynth` and keep it permanent? Then you'd need the flag `ampSynthRunning`, right? But you could register the permanent amp synth and then you'd know it it isn't running, no?

adcxyz: yes, that is the idea. 
The next steps are:
- the ampSynth gets a SendReply to send its current settings back to all clients; 
- the clients update their volume settings with the incoming info
- because all clients use the same volume group/nodeID, 
any client can start a volume synth which is accessible equally from all clients.

adcxyz:
@telephon 
synced volume between clients is working now, see tests and explanations in intro.

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -64,17 +117,33 @@ Volume {
 	updateSynth {
 		var amp = if(isMuted.not) { volume.dbamp } { 0.0 };
 		var active = amp != 1.0;
+
+		thisMethod.postln;
```
telephon, line 121: this probably needs to be removed?

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -4,37 +4,96 @@ Volume {
...
+
 		// execute immediately if we're already past server tree functions
 		if(server.serverRunning) {
 			this.sendSynthDef;
 			this.updateSynth;
 		};
 
 		initFunc = {
-			ampSynth = nil;
+			ampSynthRunning = false;
```
telephon, line 30: couldn't this be done with `Group(...).register`?

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -4,37 +4,96 @@ Volume {
...
+			// server.post; msg.round(0.001).postcs;
+
+			invol = invol.ampdb;
+			if (invol.equalWithPrecision(volume).not) {
+				"sync vol: ".postln;
+				volume = invol;
+				this.changed(\amp, volume);
+			};
+			if (inlag.equalWithPrecision(lag).not) {
+				"sync lag:".postln;
```
telephon, line 62: we probably don't want these postlns?

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -64,17 +117,33 @@ Volume {
...
 		if(active) {
 			if(server.hasBooted) {
-				if(ampSynth.isNil) {
-					ampSynth = Synth.after(server.defaultGroup, defName,
-						[\volumeAmp, amp, \volumeLag, lag, \bus, startBus])
+				if(ampSynthRunning.not) {
+					// Synth.after(server.defaultGroup, defName,
+					// [\volumeAmp, amp, \volumeLag, lag, \bus, startBus])
+
+					server.sendMsg(*group.newMsg(RootNode(server), \addToTail));
```
telephon, line 129: same here?

SCClassLibrary/Common/Control/Volume.sc
adcxyz: yes, is removed now

SCClassLibrary/Common/Control/Volume.sc
adcxyz: yes, removed

SCClassLibrary/Common/Control/Volume.sc
adcxyz: I guess yes. What would the advantages of registering be? 

As it is, we already track whether the node is alive or not by its volume status messages. 
The only case I see where that misses a change is when ```s.volume.ampSynth.free``` is called,
because then the ampSynth does not send a last triggered update with gate 0.

telephon: I though that it may be easier to read and clearer expression of intent.

adcxyz: yes, you are right. with onFree, this catches irregular ends to ampSynth as well. thanks!

SCClassLibrary/Common/Control/Volume.sc
adcxyz: done I think

SCClassLibrary/Common/Control/Volume.sc
adcxyz: done

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -64,17 +119,37 @@ Volume {
...
+						*[\volumeAmp, amp, \volumeLag, lag, \bus, startBus]
+					);
+
+					// also catches stray endings of ampSynth
+					ampSynth.onFree {
+						// "ampSynth was freed".postln;
+						this.volume = 0;
+						ampSynth.isPlaying = false;
+					};
+					ampSynth.isPlaying = true;
```
telephon, line 143: if the ampSynth uses `onFree`, then it is registered anyhow. But you can register it repeatedly without harm.

```


g = Group.basicNew(s, 67);

s.sendMsg(*g.register.newMsg(RootNode(s), \addToTail));

g.isPlaying

// send again
s.sendMsg(*g.register.newMsg(RootNode(s), \addToTail));

g.isPlaying // true

s.freeAll;

g.isPlaying; // false

// send again
s.sendMsg(*g.register.newMsg(RootNode(s), \addToTail));

g.isPlaying // true


// you can register repeatedly:
3.do { g.register };
NodeWatcher.all[\localhost].nodes

```

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -95,10 +170,8 @@ Volume {
 	}
 
 	freeSynth {
-		ServerTree.remove(updateFunc, server);
-		updateFunc = nil;
 		ampSynth.release;
-		ampSynth = nil
+		ampSynth.isPlaying = false;
```
telephon, line 174: this one could go.

SCClassLibrary/Common/Control/Volume.sc
telephon, line 143: this one could go.

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -64,17 +119,37 @@ Volume {
...
+						group.nodeID,
+						*[\volumeAmp, amp, \volumeLag, lag, \bus, startBus]
+					);
+
+					// also catches stray endings of ampSynth
+					ampSynth.onFree {
+						// "ampSynth was freed".postln;
+						this.volume = 0;
+						ampSynth.isPlaying = false;
+					};
```
telephon, line 142: this one could go.

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -64,17 +119,37 @@ Volume {
...
+						Node.addActions[\addToTail],
+						group.nodeID,
+						*[\volumeAmp, amp, \volumeLag, lag, \bus, startBus]
+					);
+
+					// also catches stray endings of ampSynth
+					ampSynth.onFree {
+						// "ampSynth was freed".postln;
+						this.volume = 0;
+						ampSynth.isPlaying = false;
```
telephon, line 141: this one could go.

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -4,37 +4,98 @@ Volume {
...
+				if (invol.equalWithPrecision(volume).not) {
+					volume = invol;
+					this.changed(\amp, volume);
+				};
+				if (inlag.equalWithPrecision(lag).not) {
+					lag = inlag;
+				};
+				// triggers when gate closed, so we know ampSynth ends
+				if (ingate <= 0.0) {
+					ampSynth.isPlaying = false;
```
telephon, line 69: this one could probably go.

SCClassLibrary/Common/Control/Volume.sc
```diff
@@ -4,37 +4,98 @@ Volume {
...
+
+			// only start listener to volume settings if multiple clients expected
+			if (server.options.maxLogins > 1) {
+				this.startListener
+			}
 		};
+		ServerBoot.add(initFunc, server);
+
+		updateFunc = {
+			ampSynth.isPlaying = false;
```
telephon, line 45: this one could go.

SCClassLibrary/Common/Control/Volume.sc
adcxyz: ok

SCClassLibrary/Common/Control/Volume.sc
adcxyz: ok

SCClassLibrary/Common/Control/Volume.sc
adcxyz: done.

SCClassLibrary/Common/Control/Volume.sc
adcxyz: ok

telephon (commented):
just one suggestion for the test

telephon:
Hey, @adcxyz if you are out of patience with this, it is ok, we can live with it! Just let me know …

adcxyz:
ok, I found the examples for CondVar, will try to get that right.

adcxyz:
@telephon: 
I replaced all the waits with condition, and all tests pass, 
so I think this is good to go.
Thanks for the pointers!

telephon (approved):
This is good now., thanks a lot! Note that this increases the permanent allocated node ids by two. We need to put a warning in the release notes that:
- all sclang versions in the network need to be up to date
- anyone using the node ids 2 and 3 hardcoded will have to change this (it is generally not recommended to do this anyway, but just in case)
- networked situations with non-sclang clients may need to adjust their code, minimally by calling allocPerm for two ids at server boot.

```supercollider
s.reboot;
(s.nodeAllocator.allocPerm == 4).postln
```

telephon:
@muellmusik could you take a final look and merge if good?

dyfer:
FYI we are trying to get all the PRs merged for the 3.14 release before 2025-05-31.

adcxyz:
> FYI we are trying to get all the PRs merged for the 3.14 release before 2025-05-31.

great!

Issue #909, opened 2013-07-15
Disable mathjax - an unused feature that can cause problems


danstowell:
The MathJax feature isn't really used by anyone, but it can cause problems.

See this discussion here where Dan, Tim, Jonatan agreed it should go:
http://comments.gmane.org/gmane.comp.audio.supercollider.devel/56552

Then months later, it caused problems for Yvan too:
http://permalink.gmane.org/gmane.comp.audio.supercollider.devel/59212

(It also complicates debian packaging since debian policy insists on using system rather than bundled mathjax.)

We should remove the mathjax feature.

PR #6317, opened 2024-05-18
Add formula support for SCDoc via kaTeX

capital-G:
**DISCLAIMER**: The intent of this PR is to create a more compact PR in terms of implementation and discussion than the already existing #6260. Please keep the discussion on this PR close to the topic.

Big thanks to @prko and @smoge on the initial work and impulse to get this in - most of the code is inspired or copied from their PR.

## Purpose and Motivation

Programming, and DSP in particular, is tied to formulas. This PR introduces the ability to write formulas within the documentation using LaTeX, using [kaTeX](katex.org) as the rendering engine within the rendered HTML files.

The following screenshots of the new documentation explain the feature via the new SCDoc syntax

image

and

image

---

There is also an existing PR #6260 which tries to add LaTeX support using [MathJaX](https://www.mathjax.org/), but when comparing MathJaX to kaTeX there are certain differences when looking at how this could be integrated into SC

* kaTeX consists of only 2 files plus font files - mathJaX consists of at least 8 files which have to be selected manually from the build and are loaded on demand.
* kaTeX is much lighter - 550kB vs. 7.4 MB for mathJaX, both including fonts
* mathJaX has a [debian package](https://packages.debian.org/sid/libjs-mathjax). Due to Debian's shared library policy, this would add additional steps to the packaging or building process. Although kaTeX also has a [debian package](https://packages.debian.org/sid/katex), it targets the CLI capabilities of kaTeX using a NodeJS runtime, which we don't use, so the shared lib policy would not cause any conflicts..

Basically, due to its smaller footprint, it becomes another static js file next to the already included jquery and codemirror js libraries.
The same arguments were used in a German math forum to replace mathJaX with kaTeX, see https://www.mathelounge.de/763103/neu-latex-rendern-nun-mittels-katex-statt-mathjax.

---

Regarding (future) maintenance: Since JS rarely deprecates features, it is not really necessary to update kaTeX. If an update should be necessary, a `README.md` is included that describes how the current set of kaTeX files was generated. 

---

SuperCollider once had the ability to include LaTeX code in the docs via mathJaX, but it was removed because it caused crashes, see https://github.com/supercollider/supercollider/issues/909 (the dead links lead to https://listarc.cal.bham.ac.uk/lists/sc-dev-2013/msg49910.html ).
These crashes are no longer reproducible and kaTeX is extensively tested [on all 3 major browser engines](https://github.com/KaTeX/KaTeX/blob/bedce93260cc413abb7730a5693dc27222297178/.github/workflows/screenshotter.yml#L15-L27), where `sc-ide` relies on the Chromium engine via Qt web engine.

---

There was also some discussion about whether such a feature was even necessary for SCDoc, considering it bloat and discussing different ways to implement such a feature without touching the SCDoc code. But I think the recent initiative by @prko and @smoge shows that there is a demand for such a feature and with the move to kaTeX, I think the footprint in terms of file size and maintenance becomes tolerable.

## Types of changes

- [x] Documentation
- [x] New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


telephon (commented):
The changes to bison seem to be partly unrelated?

telephon:
Thank you – this looks really good already.

capital-G:
> The changes to bison seem to be partly unrelated?

The changes within the SCDoc parser are necessary because this PR extends the syntax of `.schelp` with a `math` keyword - the parser is managed by Bison, therefore it is necessary to modify these files.

smoge:
> The changes to bison seem to be partly unrelated?

the bison file will need a change anyway. Right now it uses a deprecated function. The fix is very straightforward, if you run `bison --update` it will replace the deprecated function.

Maybe, for the sake of git history organization, this should have its own commit?

Further changes would be necessary only if we prefer to use the `math:: ::` syntax. It would need to have two variations, to discern inline and display options. 

Example inline and display syntaxes:
```

math::a^2 + b^2 = c^2::

math::
e^{i\pi} + 1 = 0
::

```


It was also proposed to use the `\( )\` and `\[ \]` syntax (inline and display).

smoge:
Also, just to remember another difference between the two libraries. MathJax can also create block diagrams, which are used all the time in DSP at the sample level. If SuperCollider and Faust become more integrated in the future, for example, this would be even more useful. So the SC help would provide information that would inform Faust UGen development, and vice versa. 

Maybe a topic for future discussions?

image

telephon:
> Maybe, for the sake of git history organization, this should have its own commit?

yes, this would be nice for clarity, if it is not too much of a hassle.

> It was also proposed to use the \( )\ and \[ \] syntax (inline and display). But, I think math:: is more consistent.

yes, it alignes better with the general syntax of the schelp.

smoge:
One question I have is if Katex has the equivalent of MathJax `async` option, which makes it run without interfering or waiting with any other script. 

Katex has a smaller footprint, but the way it is loaded and run has an equal impact on performance since other js scripts are running on SCDoc.

prko:
The mathjax mode, which the mathjax lead developer recommends, is `defer`, if I remember correctly. (um, I do not know which one is better)... 

Does Katex all of the three options: `sync`, `sync` and `defer`?

smoge:
> The mathjax mode, which the mathjax lead developer recommends, is `defer`, if I remember correctly. (um, I do not know which one is better)...
> 
> Does Katex all of the three options: `sync`, `sync` and `defer`?

Ah, ok. From the docs: "the [defer attribute](https://developer.mozilla.org/en/HTML/Element/script#Attributes) indicates that the script doesn't need to execute until the page has loaded, speeding up page rendering;"

`<script defer src="/dist/katex.min.js" crossorigin="anonymous"></script>`

prko:
@smoge 
><script defer src="/dist/katex.min.js" crossorigin="anonymous"></script>

@capital-G  already did it.

---

@capital-G  Thanks, I tested it! Your implementation also renders tables: Wow!
Screenshot 2024-05-24 at 19 40 25

capital-G:
> yes, this would be nice for clarity, if it is not too much of a hassle.

It essentially affects one LOC - https://github.com/supercollider/supercollider/pull/6317/files#diff-6e1dd573c3a1ee898126f224ea1664f9e713c232b7a1ec8272827006c617689aR54

If people insist on it I could rebase it, but I think it is not too necessary in this case?

Is there anything else that could block this? What could be ways to get this reviewed and merged?

telephon (approved):
I've compiled it and it runs well. The code changes look reasonable t me, although I have to say that I don't know about the bison part. It seems not to break anything, so I'll approve.

capital-G:
> I've compiled it and it runs well. The code changes look reasonable t me, although I have to say that I don't know about the bison part. It seems not to break anything, so I'll approve.

Maybe I can explain the changes to the Bison part, just to be sure?

## Lexer part

image

Just copy paste from teletype, as this is the behavior we want to have as well for a math block

## Parser part

image

This was necessary because of a deprecation within newer Bison versions which made the file not valid anymore. It was fixed/adjusted with the help of update command of Bison.
This could have been its own commit :/

image

Copy-paste from teletype, but with different name.

The other files are auto-generated code via the flex and bison commands - often these auto-generated files are only deterministic within a version, so it generates some noise when re-generating them.
But if it compiles and the test succeed, it should be valid.

Do you think it needs some additional changes?

---

Is there a rule when a PR can be merged? Is one approval sufficient or should there be at least 2?

telephon:
> Is there a rule when a PR can be merged? Is one approval sufficient or should there be at least 2?

For slightly larger things like this, it is good practice that a third person does the merge.

capital-G:
Just checked if this is compatible with #6095 - it is :)

image

image

capital-G:
@JordanHendersonMusic can you review this? :)

CMakeLists.txt
```diff
@@ -462,7 +462,7 @@ configure_file(common/SC_Version.hpp.in common/SC_Version.hpp)
 include_directories(${CMAKE_CURRENT_BINARY_DIR}/common)
 
 file(GLOB_RECURSE ClassLibrary SCClassLibrary/*sc)
-file(GLOB_RECURSE HelpSource   HelpSource/*)
+file(GLOB_RECURSE HelpSource   HelpSource/**)
```
JordanHendersonMusic, line 465: Could just explain what this does and why it's needed?

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -235,11 +235,13 @@ SCDocHTMLRenderer {
 		<< "<link rel='stylesheet' href='" << baseDir << "/editor.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/frontend.css' type='text/css' />\n"
 		<< "<link rel='stylesheet' href='" << baseDir << "/custom.css' type='text/css' />\n"
+    	<< "<link rel='stylesheet' href='" << baseDir << "/lib/katex/katex.min.css' />\n"
```
JordanHendersonMusic, line 238: The indentation is off here, is it a tabs Vs spaces thing?

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -591,6 +593,16 @@ SCDocHTMLRenderer {
...
 			},
+            \MATHBLOCK, { // uses temml to convert tex to MathML
+                stream << "<span class='math block'>\n"
+                << this.escapeSpecialChars(node.text)
+                << "\n</span>";
+            },
+            \MATH, {
+                stream << "<span class='math'>"
+                << this.escapeSpecialChars(node.text)
+                << "</span>";
```
JordanHendersonMusic, line 604: Indentation sim.

JordanHendersonMusic (changes requested):
Looks good! 

There's something I don't understand in the cmake, and I think you are using spaces over tabs in the SC code?

This seems to address all the previous concerns regarding dependencies as this is a relatively small static js file that we can upgrade (or not) as we like, and the changes to the scdoc system are minimal.

JordanHendersonMusic:
Does the CI test the docs in any way?

CMakeLists.txt
capital-G: I remember having problems that some files were not copied, but this seems this was caused when doing some experiments with MathJax. Normally in glob, `*` matches everything except `/` and `**` anything, so including all sub directories.

I just ran a build with the old CMakeLists from a clean build dir and it still works, so I will rebase/edit the commit.

SCClassLibrary/SCDoc/SCDocRenderer.sc
capital-G: well spotted! vscode converts indentations form tab to 4 spaces which happened here - have changed my config and will try to reduce noise by rebasing the commit.

capital-G:
Thanks for the fast response and pointing out some spots :) I'll try to fix those by rebasing from develop.

> Does the CI test the docs in any way?

IIRC the CI does build SCDoc as part of SCLang, but we don't perform a build of the docs/help files.
I suggested it once but it was rejected back then - we have https://github.com/supercollider/sc-docs and https://github.com/capital-g/sc-docs-dev (see https://github.com/supercollider/sc-docs/issues/3) though, but this only provides weekly builds from the dev branch.

capital-G:
@JordanHendersonMusic I hope I fixed everything, can you give it another go? :)

HelpSource/scdoc.js
```diff
@@ -282,3 +282,17 @@ function setUpWebChannel(port) {
...
+    for (let aSpan of document.getElementsByClassName("math")) {
+        const tex = aSpan.textContent;
+        katex.render(tex, aSpan, {
+            throwOnError: false,
+            displayMode: aSpan.classList.contains("block"),
+        });
+    }
+}
+
+window.addEventListener('DOMContentLoaded',function () {
```
JordanHendersonMusic, line 296: I'm not super familiar with web tech, but [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event) says 

'DOMContentLoaded may fire before your script has a chance to run, so it is wise to check before adding a listener.'

...and gives this solution...
```js
function doSomething() {
  console.info("DOM loaded");
}

if (document.readyState === "loading") {
  // Loading hasn't finished yet
  document.addEventListener("DOMContentLoaded", doSomething);
} else {
  // `DOMContentLoaded` has already fired
  doSomething();
}
```

Is that something that should be done here?

capital-G: MDN says

> The DOMContentLoaded event fires when the HTML document has been completely parsed, and all deferred scripts ([<script defer src="…">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer) and [<script type="module">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#module)) have downloaded and executed. It doesn't wait for other things like images, subframes, and async scripts to finish loading.

which is pertty much what we want here as we work with a defered script and need to wait for the DOM to be available in the browser. I think I also copied this code mostly from the kaTeX docs https://katex.org/docs/autorender

I think everything is as it should be regarding the execution of JS code.

JordanHendersonMusic (approved):

PR #6319, opened 2024-05-22
Add pre commit

capital-G:
Closes #6096 

## Purpose and Motivation
SuperCollider has a formatter/linter policy for its C++ files, which is checked via clang-format.
This check is also done in CI, and any commit that does not pass the clang-format stage results in a failed build that is not checked further.

As the formatting rules of clang-format are not consistent between major versions, it is necessary to enforce a specific version of clang-format to create deterministic formatting - the correct version is checked by our formatting tool `clang-format.py`.

https://github.com/supercollider/supercollider/blob/766791ca951f39a9479de2663fe51ab76dd06841/tools/clang-format.py#L121-L122

To have access to a specific clang-format version on a system is not trivial and made e.g. #6196 necessary.

In order to reduce dev dependency management for users who occasionally want to work on CPP files, this PR suggests using https://pre-commit.com/ to manage any necessary formatting and its dependencies to apply that formatting.
Not only does pre-commit check that the files are properly formatted before committing, but it also allows the correct clang-format version to be downloaded as a Python dependency (see https://github.com/ssciwr/clang-format-wheel which supports all major clang format versions for linux/mac/windows).
As Python is already a dependency for formatting code, this does not introduce a new dependency beyond a Python package, but drops the dev dependency of clang-format as a system installation!

It is still possible to work without pre-commit, but this can still lead to rejection by CI if the code is not formatted properly.
The formatter will also only run if any C++ files have been touched.

If we update the clang format version, or want to add additional checks, we can simply edit/add them in the `.pre-commit-config.yaml` file - this makes working in branches easier, as formatting rules are specified via files.
Pre-commit will automatically pick up any necessary changes (e.g. to dependencies).

Pre-commit is a trusted tool within the Python community and beyond, and has been around for over 10 years, see https://github.com/pre-commit/pre-commit
The (optional) installation is as simple as `pip install pre-commit` and `pre-commit install`, see https://pre-commit.com/ for more details.

## Example

Lets say I have the following C++ file changes staged which I want now to commit

grafik

These clash with how clang-format wants to format them, and when I want to commit this staged file, `pre-commit run` will automatically run in the background and check if the formatting is ok - as clang-format says it is not ok, the `git commit` process will fail.

grafik

The Python formatting script `tools/clang-format.py formatall` has already applied all formatting fixes to the staged files, and can be compared from the current (fixed) state to the staged state.

grafik

When we stage these fixes, we can run the `commit` command again, the files will be checked for formatting, and if the formatter does not complain, the commit will be performed successfully.
This allows to create properly formatted commits all the time without thinking about how to apply the formatting, it happens automatically in the background.

The CI will also check that the codebase passes the formatter, as before.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### Merge notes

Should also be documented in the wiki or dev setup docs.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -27,12 +27,11 @@ jobs:
         run: |
           sudo npm install -g lintspaces-cli
           lintspaces -e .editorconfig SCClassLibrary/**/*.sc || true # ignore failure
-      - name: lint cpp files
-        run: |
-          sudo apt-get install -y clang-format-14
-          echo "Running tools/clang-format.py lintall"
-          tools/clang-format.py -c clang-format-14 -d clang-format-diff-14 lintall || exit 1
-          echo "Lint successful"
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+      - uses: pre-commit/action@v3.0.0
+        name: 'Run pre-commit'
       - name: set version string for artifacts
         id: set-version
         run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     hooks:
     -   id: clang_format
         name: Format C++ code
-        entry: python tools/clang-format.py formatall
+        entry: python tools/clang-format.py lintall
         pass_filenames: false
         types_or:
           - "c++"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    -   id: check-merge-conflict
+    -   id: check-case-conflict
+-   repo: local
+    hooks:
+    -   id: clang_format
+        name: Format C++ code
+        entry: python tools/clang-format.py formatall
+        pass_filenames: false
+        types_or:
+          - "c++"
+        language: python
+        additional_dependencies: ['clang-format==14.0.6']
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,15 @@ repos:
     hooks:
     -   id: clang_format
         name: Format C++ code
-        entry: python tools/clang-format.py lintall
+        entry: python tools/clang-format.py formatall
+        pass_filenames: false
+        types_or:
+          - "c++"
+        language: python
+        additional_dependencies: ['clang-format==14.0.6']
+    -   id: clang_lint
+        name: Lint C++ code
+        entry: python tools/clang-format.py lintall 
         pass_filenames: false
         types_or:
           - "c++"


JordanHendersonMusic:
Thank you, this is a huge quality of life upgrade, particularly for not having to keep track of which clang-format version is currently required! This seems like an all round better solution from what we currently have.

capital-G:
I allowed myself to add `pre-commit/action@*` to this projects allowed actions as otherwise the CI of this PR would not start.

I also rebased it from develop, to trigger the CI and benefit from the recent develop additions which fixed the CI.

@JordanHendersonMusic what could be a way to get this reviewed and potentially merged?

JordanHendersonMusic:
@capital-G I've asked over on the slack developer page. If I'm understand correctly, reverting this would be entirely trivial, and since it only affects developers, I don't see any reason why it can't just be merged. 

I wonder if the linter is mentioned in the wiki anywhere, I had a quick look and couldn't find it, but this could be followed with an update over in the developing section (if this goes ahead).



---
From the slack:

> If I'm reading it correctly, it uses CI to run the formatter, rather than as a local precommit hook? Getting CI to make code changes is usually not recommended, I think it'll get messy at some point, but I won't get in the way if it's what people want

@jrsurge 

Is this true? My understanding was that invoking `git commit` locally would trigger the formatter, but otherwise nothing would change?

--- 

... there are definitely too many modes of communication, and I now agree with you about the GitHub discussions!

capital-G:
> @capital-G I've asked over on the slack developer page. If I'm understand correctly, reverting this would be entirely trivial, and since it only affects developers, I don't see any reason why it can't just be merged.

Reverting this would be to simply delete the `.pre-commit.yaml` file and delete/change the GitHub CI accordingly.

> I wonder if the linter is mentioned in the wiki anywhere, I had a quick look and couldn't find it, but this could be followed with an update over in the developing section (if this goes ahead).

I found something via the search, but this is outdated as still relies on travis.
https://capital-g.github.io/supercollider-wiki/contributing/style/cpp-formatting/Cpp-formatting-instructions.html

> 
> From the slack:
> 
> > If I'm reading it correctly, it uses CI to run the formatter, rather than as a local precommit hook? Getting CI to make code changes is usually not recommended, I think it'll get messy at some point, but I won't get in the way if it's what people want
> 
> @jrsurge
> 
> Is this true? My understanding was that invoking `git commit` locally would trigger the formatter, but otherwise nothing would change?

`pre-commit` is essentially a package manager/configuration system for git hooks. You can activate/install the hooks on your own/local machine and each commit will be checked prior by this framework according to the rules we set (e.g. C++ code must be linted according to clang-format 14.0.*).
This installation via `pre-commit install` is totally optional, but lets you forget about any formatting workflows.

But we also install `pre-commit` in the CI to check if the commits are passing the checks we defined - it is therefore possible to run the same tests in CI and locally.
 
> ... there are definitely too many modes of communication, and I now agree with you about the GitHub discussions!

TIL there is a SC slack channel :D So discourse, discord, github, slack, telegram ...? :D I enjoy the variety, but it is difficult to get feedback from as much people as possible this way.

mtmccrea:
> Getting CI to make code changes is usually not recommended,

Will it make code changes? Wouldn't it just run the linter and fail if it doesn't pass?

mtmccrea:
> Getting CI to make code changes is usually not recommended

Can it be set in CI to just run the linter and fail if it doesn't pass? I see how it could get messy if you push changes that are subsequently changed by CI, so your local repo is then out of sync (you'd need to pull the changes made by CI to sync).

> This installation via pre-commit install is totally optional, but lets you forget about any formatting workflows.

^This then seems like the main advantage—cutting out the step of manually running formatting. (I often forget this and find it to be a hassle when setting up on a new machine). If CI doesn't run the formatter (only lints) the local pre-commit is "optional" in the same way that clang-format is currently "optional".

capital-G:
> ^This then seems like the main advantage

This, and also the setup/installation becomes really easy as the proper `clang-format` version can be actually be setup by `pre-commit`.

> Can it be set in CI to just run the linter and fail if it doesn't pass?

That is how it works :)

## Locally
* Stage files for commit and perform actual commit
* If C++ files are touched, these files will be formatted and linted - if any step has modified the files and/or fails (via exit code 1), the git-hook will return exit code 1 and the commit will be rejected/unsuccessful - it will also tell you what the problem was
* Verify the diff of your staged files and the modifications done by the linter - stage the changes try to perform the commit again. If all C++ files are valid, the git hook will return exit code 0 and the commit will be performed

## CI
* The CI will simply perform all checks on all files of the repository - if any check fails it will return exit code 1 and the CI linting step will fail. The CI does not generate commits.

## An Example:

Lets say I have the following change staged

```diff
diff --git a/lang/LangSource/SC_LanguageConfig.cpp b/lang/LangSource/SC_LanguageConfig.cpp
index f4b22c293..dab935ec1 100644
--- a/lang/LangSource/SC_LanguageConfig.cpp
+++ b/lang/LangSource/SC_LanguageConfig.cpp
@@ -95,7 +95,9 @@ bool SC_LanguageConfig::forEachIncludedDirectory(bool (*func)(const Path&)) cons
 
 bool SC_LanguageConfig::pathIsExcluded(const Path& path) const { return findPath(mExcludedDirectories, path); }
 
-bool SC_LanguageConfig::addIncludedDirectory(const Path& path) { return addPath(mIncludedDirectories, path); }
+bool SC_LanguageConfig::addIncludedDirectory(const
+Path& 
+iWantThisRenamed) { return addPath(mIncludedDirectories, path); }
 
 bool SC_LanguageConfig::addExcludedDirectory(const Path& path) { return addPath(mExcludedDirectories, path); }
```

```shell
❯ git status
On branch pre-commit
Your branch is up to date with 'fork/pre-commit'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   lang/LangSource/SC_LanguageConfig.cpp
```

Now I want to commit this change

```shell
❯ git commit -m "this will fail"
Check for merge conflicts................................................Passed
Check for case conflicts.................................................Passed
Format C++ code..........................................................Failed
- hook id: clang_format
- files were modified by this hook
Lint C++ code............................................................Passed
```

The formatter changed the file and therefore it failed, a `git diff` shows the changes between staged files and current state

```diff
diff --git a/lang/LangSource/SC_LanguageConfig.cpp b/lang/LangSource/SC_LanguageConfig.cpp
index dab935ec1..849287b2d 100644
--- a/lang/LangSource/SC_LanguageConfig.cpp
+++ b/lang/LangSource/SC_LanguageConfig.cpp
@@ -95,9 +95,9 @@ bool SC_LanguageConfig::forEachIncludedDirectory(bool (*func)(const Path&)) cons
 
 bool SC_LanguageConfig::pathIsExcluded(const Path& path) const { return findPath(mExcludedDirectories, path); }
 
-bool SC_LanguageConfig::addIncludedDirectory(const
-Path& 
-iWantThisRenamed) { return addPath(mIncludedDirectories, path); }
+bool SC_LanguageConfig::addIncludedDirectory(const Path& iWantThisRenamed) {
+    return addPath(mIncludedDirectories, path);
+}
 
 bool SC_LanguageConfig::addExcludedDirectory(const Path& path) { return addPath(mExcludedDirectories, pat
h); }
```

After we verify that the changes by the formatter are good, we can add the changes to our staged files via `git add lang/LangSource/SC_LanguageConfig.cpp` and now, `git commit -m "this will work"` shall succeed.

JordanHendersonMusic:
Another comment from the slack

> That’s a bad idea.
It’s quite counter intuitive to have a pre-commit hook running in CI. Especially in this case since the job is not even committing anything but only linting.
I would recommend running explicitly the same commands that are run in the current hook in CI.
Also there’s another problem, right now the there are just two hooks, what if a new one is added? Will the job continue working? Will it break misteriously? Will more dependencies be required? It creates too many failure points.

mtmccrea:
>CI
The CI will simply perform all checks on all files of the repository - if any check fails or if any files have been modified by a hook it will return exit code 1 and the CI linting step will fail. The CI does not generate commits.

OK great, so I believe this addresses @jsurge's concern.

capital-G:
> Another comment from slack

Although I appreciate your messenger efforts, I think it would be easier if the discussion of this PR would take place solely on GitHub to keep things simpler. Maybe you can invite the people over here?

> It’s quite counter intuitive to have a pre-commit hook running in CI. Especially in this case since the job is not even committing anything but only linting.
> I would recommend running explicitly the same commands that are run in the current hook in CI.
> Also there’s another problem, right now the there are just two hooks, what if a new one is added? Will the job continue working? Will it break misteriously? Will more dependencies be required? It creates too many failure points.

It seems there is a misunderstanding what this PR is about or how [pre-commit](https://pre-commit.com/) works, as these points do not really apply here and this PR essentially solves the voiced concerns. I don't really know what to respond in this case.

silvanocerza:
> It seems there is a misunderstanding what this PR is about or how [pre-commit](https://pre-commit.com/) works, as these points do not really apply here and this PR essentially solves the voiced concerns. I don't really know what to respond in this case.

The comment above is mine.

I'm not against pre commit hooks by the way, and I do understand their purpose. But I believe that using it as CI step hides quite some pitfalls. Just the first ones that come to mind.
* Makes it harder to debug failures
* Adds indirection
* Makes it harder to understand what the CI is actually doing
* Coupling of pre commit hooks and CI checks
* Assumes all pre commit hooks will only lint code

The only benefit I see it that if local hooks run this will pass. Though from my experience most people ignore them with either `--no-verify` or just don't install them.

If this PR is supposed to be adding pre commit hooks it should be enough to remove the CI changes. If instead it's supposed to add new linting and formatting checks I would suggest changing the CI to run those checks explicitly.

If you want CI and local hooks in sync I would add a pre commit hook that runs a script for linting and use that same script in the CI step. I wouldn't run the hook though.

Also considering that the action you're using is in maintanance mode as [stated in its `README.md`](https://github.com/pre-commit/action) adds yet another concern.

capital-G:
Thanks for your comment, I still have some questions concerning your comment.

> I'm not against pre commit hooks by the way, and I do understand their purpose. But I believe that using it as CI step hides quite some pitfalls. Just the first ones that come to mind.
> 
>  * Makes it harder to debug failures

What kind of failures are you referring to? Linting failures or software failures of our formatting tool?

>  * Adds indirection

What kind of indirection? I'd argue it removes indirection.

> * Makes it harder to understand what the CI is actually doing

It runs the pre-commit pipeline? I think the steps are actually quite clear based on the logs and the `pre-commit` file.

> * Coupling of pre commit hooks and CI checks

I think it is a common pattern? Many projects do the same, e.g. requests, flask, pandas, ... - see https://github.com/psf/requests/blob/main/.github/workflows/lint.yml

>  * Assumes all pre commit hooks will only lint code

Why does this PR assume this? Currently it would be limited to this, but there could be discussion if we use additional functionality/hooks of pre-commit.

> The only benefit I see it that if local hooks run this will pass. Though from my experience most people ignore them with either `--no-verify` or just don't install them.

We can't force people to install anything - if they have their own way to generate linted code it is fine, it is mainly build who want to contribute but don't know how to lint properly or have problems installing the required clang-format version, see e.g. #6196 
Some people stopped to contribute C++ code - this would make their formatting-C++-code-live much easier.
 
> If this PR is supposed to be adding pre commit hooks it should be enough to remove the CI changes. If instead it's supposed to add new linting and formatting checks I would suggest changing the CI to run those checks explicitly.

I don't understand why we should  maintain two ways of formatting/linting code (one locally and one for CI) if this is a tool where we can run the same code on both ends?

> If you want CI and local hooks in sync I would add a pre commit hook that runs a script for linting and use that same script in the CI step. I wouldn't run the hook though.

But this is the setup of this PR? We call the Python script for formatting via pre-commit.

> Also considering that the action you're using is in maintanance mode as [stated in its `README.md`](https://github.com/pre-commit/action) adds yet another concern.

This action was deprecated (in favor of pre-commit.ci) but due to outcry they put it in maintenance mode. I think it wouldn't be too hard to replace this with

```yaml
      - name: 'Run pre-commit'
        runs:
           pip install pre-commit 
           pre-commit run --all-files
```

telephon:
> Some people stopped to contribute C++ code - this would make their formatting-C++-code-live much easier.

Yes, I haven't been able to do any work on the C++ side since the linter was added, because I simply can't get the complicated workflow to work. I know of others for whom this is also a significant entry barrier. Unlike in other projects, there are people in the community who need to do things in the engine-room but aren't ful time engineers.

But it is important of course to carefully check the proposal.

silvanocerza:
@capital-G 
> What kind of failures are you referring to? Linting failures or software failures of our formatting tool?

Any kind of failure really.

> What kind of indirection? I'd argue it removes indirection.
>
> It runs the pre-commit pipeline? I think the steps are actually quite clear based on the logs and the `pre-commit` file.

These two go hand in hand in my opinion. Ideally each step should be easy to understand without external context, calling a script as part of a step forces you to understand what the script does. This creates indirection I think.

> I think it is a common pattern? Many projects do the same, e.g. requests, flask, pandas, ... - see https://github.com/psf/requests/blob/main/.github/workflows/lint.yml

That's a choice, in my opinion not the best one as it precludes the use of pre-commit hooks for anything else other than linting code.

> Why does this PR assume this? Currently it would be limited to this, but there could be discussion if we use additional functionality/hooks of pre-commit.

This is the main point of the discussion I think. If you make this assumption now you add an extra obstacle to use pre-commit hooks for anything else other than linting code. It might seem a small thing but will require extra work in the future. 

If we instead decide not use them as part of the CI now you can easily add new ones with no overhead in the future.

Right now you also risk people adding new hooks that are not related to linting. If those new hooks don't fail in CI for any reason you just added something unrelated in the linting CI job. Not saying it will happen for sure, it's an additional risk though.

> We can't force people to install anything - if they have their own way to generate linted code it is fine, it is mainly build who want to contribute but don't know how to lint properly or have problems installing the required clang-format version, see e.g. #6196 Some people stopped to contribute C++ code - this would make their formatting-C++-code-live much easier.

Enforcing use of some tools is wrong in my opinion, though enforcing formatting is not. If people don't know how to follow formatting convention of a project we should ideally teach them how to do that. If they don't want to respect convention cause of any reason that's just bad and those contributions should not be welcome in my opinion.

> I don't understand why we should maintain two ways of formatting/linting code (one locally and one for CI) if this is a tool where we can run the same code on both ends?
> 
> But this is the setup of this PR? We call the Python script for formatting via pre-commit.

My bad on this, didn't notice the hook was already running a script, I should have checked better.

Though I think this is the way to go. Since we already have the scripts we should use those.

Also I noticed that the formatting is pretty slow, it could actually be beneficial to run formatting and linting in a matrix job. It's faster and you see if both fails so you can fix both at the same time. Running them sequentially would prevent noticing a linting failure if formatting fails too.

While writing the above I remembered that you can actually run pre-commit hooks individually. That would be the best idea I think. 

We keep the scripts for local and CI checks centralized in the same place, but instead of running all the hooks in CI we specify the ones to run. So instead of steps that run the pre-commit action you'd have two different ones that run `pre-commit run clang_format` and `pre-commit run clang_lint`.

What do you think of this approach? This will give the flexibility of adding new hooks without disrupting CI in any way. If it's ok for you I can open a PR that does this.

> This action was deprecated (in favor of pre-commit.ci) but due to outcry they put it in maintenance mode. I think it wouldn't be too hard to replace this with
> 
> ```yaml
>       - name: 'Run pre-commit'
>         runs:
>            pip install pre-commit 
>            pre-commit run --all-files
> ```
> 
> in case they remove this action.

I would go without the action right away to be fair, given they already tried to kill it it's just a matter of time before they do that.


By the way I sent you a mail, would love to discuss about this kind of stuff if you want.

silvanocerza:
Had a chat with @capital-G, we decided to got forward with the current solution.

I'll open an issue, or multiple, to outline some proposals on how to enhance CI and tooling following this. 🚀

JordanHendersonMusic:
Any one with any experience of github CI mind reviewing this? I know nothing about this, but each change seems reasonable?

capital-G:
> Any one with any experience of github CI mind reviewing this? I know nothing about this, but each change seems reasonable?

Maybe we could just merge it? The worst that can happen is that it creates trouble and we revert the change - as it only affect devs, and they seem to be rather annoyed by the current formatting setup, and pre-commit is opt-in, I think the benefits outweigh the danger. It only adds an easy and implicit way of formatting.

JordanHendersonMusic (approved):

capital-G:
I wrote a wiki post on how it works and how to set it up, see https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit

This is also linked this in 
* https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions
* https://github.com/supercollider/supercollider/wiki/Contributing-directory
* https://github.com/supercollider/supercollider/wiki/Development-resources-directory

Will merge this now.

mtmccrea:
Thanks @capital-G for following up with the docs!

mtmccrea:
I got around to testing pre-commit and it seems that linting takes a very long time, about 40 seconds or so for both passes (lint1, fail and format, lint2). Is this expected? Is linting and formatting happening across the full code base or just changes related to the commit?

capital-G:
> I got around to testing pre-commit and it seems that linting takes a very long time, about 40 seconds or so for both passes (lint1, fail and format, lint2). Is this expected? Is linting and formatting happening across the full code base or just changes related to the commit?

It currently always scans the whole code base :/
This is because we use the formatting script and not the pre-commit native clang-format hook, which would allow us to natively only take a look at the touched files.
At least the hooks `formatall` and `lintall` are only executed when touching C++ code.

The whole scan is currently necessary as we want to verify the full repo in the CI (so it is not possible to sneak an ill-formatted commit when pushing two commits), therefore `formatall` and `lintall` is used to obtain a stateless result of our formatting/linting state of the repo.

I see two solutions:
1. Extend the formatting script for a CI workflow to work on a per-file basis - it currently only allows to work on a per-commit and per-branch basis which is somehow contrary to the pre-commit workflow which assumes the formatting tools are not git-aware. Pre-commit sadly does not allow to pass the `--all-files` flag into a script, but we could sneak this into the formatting script via a environment variable, so adapt to something like `FULL_CHECK=1 pre-commit --all-files` in the CI.
2. ditch our formatting script over a pre-commit native hook. pro: less code to maintain, con: makes pre-commit a hard dev dependency, and there are some advanced features in the formatting script, though I don't know if they are still in use. This would run the formatting and linting only on the touched files

I think for now I prefer solution 1 as it is easier done and does not involve a discussion if we want to deprecate our current formatting tools.

capital-G:
@mtmccrea I created a PR which implements the first solution, see https://github.com/supercollider/supercollider/pull/6344
You can give it a try and if you think it improves the situation, you can approve it and we can merge it.

PR #6320, opened 2024-05-22
Fix boost version for macOS use system libraries build

capital-G:
## Purpose and Motivation

The current macOS build `macOS x64 use system libraries` is broken, see https://github.com/supercollider/supercollider/actions/runs/9187974816/job/25266805454

This is caused due to an upgrade to a new boost version 1.85 via brew which is not compatible with our codebase (as this build configuration is using the boost lib from the system instead of the provided via our codebase). By using https://formulae.brew.sh/formula/boost@1.76 instead of https://formulae.brew.sh/formula/boost we can still use an older version of boost for the build which will make the build pass. Please note that this is already deprecated.
As these boost files are not symlinked by brew, we need to link them explicitly for cmake.

We should still make the codebase compatible with the new boost version as this causes additional problems currently, I am currently working on that. But for now, this will at least fix the build issues.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -331,7 +331,7 @@ jobs:
           vcpkg install fftw3 --triplet="${{ matrix.vcpkg-triplet }}" --overlay-triplets="$GITHUB_WORKSPACE/vcpkg/triplets"          
       - name: install system libraries
         if: env.USE_SYSLIBS == 'true'
-        run: brew install yaml-cpp boost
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic (approved):

JordanHendersonMusic:
Looks like a good temporary fix until we can support the newer boost! Thanks!

PR #6321, opened 2024-05-23
adjust trueAt/falseAt documentation and tests to change of behaviour

telephon:
## Purpose and Motivation
This adjusts the tests and documentation to the change in trueAt / falseAt, see #6001 
## Types of changes

- Documentation
- Tests

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Dictionary.schelp
+++ b/HelpSource/Classes/Dictionary.schelp
@@ -193,10 +193,22 @@ d.matchAt(\abc)
 ::
 
 method::trueAt
-Returns link::Classes/True:: if the item at the key is true, otherwise false. This method is also valid in link::Classes/Object::.
+Returns link::Classes/True:: if the item's booleanValue at the key is true, otherwise false. This method is also valid in link::Classes/Object::.
+code::
+d = (x:true, y:false);
+d.trueAt(\x) // true
+d.trueAt(\y) // false
+d.trueAt(\foo) // false
+
+// using binary value equivalents
+d = (x:1, y:0);
+d.trueAt(\x) // true
+d.trueAt(\y) // false
+d.trueAt(\foo) // false
+::
 
 method::falseAt
-Returns link::Classes/False:: if the item at the key is not true, otherwise true. This method is inherited from link::Classes/Object::.
+Returns link::Classes/False:: if the item's booleanValue at the key is not true, otherwise true. This method is inherited from link::Classes/Object::.
 
 subsection::Testing
 method::includes
--- a/testsuite/classlibrary/TestDictionary.sc
+++ b/testsuite/classlibrary/TestDictionary.sc
@@ -28,8 +28,14 @@ TestDictionary : UnitTest {
 	}
 
 	test_trueAt {
-		var dict = (x:7, y:false, z:true);
-		this.assert([\x, \y].every(dict.falseAt(_)) and: dict.trueAt(\z), "dictionary trueAt and falseAt should respond only true if the value really equals true.");
+		var dict = (k: 1, x:nil, y:false, z:true);
+		this.assert(1.falseAt(\y), "falseAt should respond with true for any object by default.");
+		this.assert(1.trueAt(\y).not, "trueAt should respond with false for any object by default.");
+		this.assert(dict.falseAt(\y), "dictionary falseAt should respond with false if its booleanValue is equivalent to false.");
+		this.assert(dict.falseAt(\x), "dictionary falseAt should respond with false if its value is unspecified.");
+		this.assert(dict.trueAt(\z), "dictionary trueAt should respond true if its booleanValue is equivalent to true.");
+		this.assert(dict.trueAt(\k), "dictionary trueAt should respond true if its booleanValue is equivalent to true.");
+
 	}
 
 }
--- a/testsuite/classlibrary/TestDictionary.sc
+++ b/testsuite/classlibrary/TestDictionary.sc
@@ -28,14 +28,14 @@ TestDictionary : UnitTest {
 	}
 
 	test_trueAt {
-		var dict = (k: 1, x:nil, y:false, z:true);
-		this.assert(1.falseAt(\y), "falseAt should respond with true for any object by default.");
-		this.assert(1.trueAt(\y).not, "trueAt should respond with false for any object by default.");
-		this.assert(dict.falseAt(\y), "dictionary falseAt should respond with false if its booleanValue is equivalent to false.");
-		this.assert(dict.falseAt(\x), "dictionary falseAt should respond with false if its value is unspecified.");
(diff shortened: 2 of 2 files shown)


capital-G (approved):
Thanks for this change!

The tests are currently not running b/c of the broken macOS build (see #6320) but looking over this manually it seems to fix the test.

Also thanks for adding the documentation! :)

testsuite/classlibrary/TestDictionary.sc
```diff
@@ -28,8 +28,14 @@ TestDictionary : UnitTest {
 	}
 
 	test_trueAt {
-		var dict = (x:7, y:false, z:true);
-		this.assert([\x, \y].every(dict.falseAt(_)) and: dict.trueAt(\z), "dictionary trueAt and falseAt should respond only true if the value really equals true.");
+		var dict = (k: 1, x:nil, y:false, z:true);
+		this.assert(1.falseAt(\y), "falseAt should respond with true for any object by default.");
+		this.assert(1.trueAt(\y).not, "trueAt should respond with false for any object by default.");
+		this.assert(dict.falseAt(\y), "dictionary falseAt should respond with false if its booleanValue is equivalent to false.");
```
capital-G, line 34: shouldn't it respond with false in this case? The text seems wrong, the test looks wright

capital-G (changes requested):
Found something on closer inspection

capital-G (approved):
Thanks for adjusting.

I found the tests a bit hard to read now with using `assertEquals` for an `assert` operation but if works, it works.

telephon:
yes, I did this because the error messages, if they happen, are easier to understand.

mtmccrea:
#6001 seems to have introduced inconsistent behavior:

```supercollider
~testVal = 440;
d = Dictionary();
d.put(\a, ~testVal);

d.trueAt(\a);  // -> 440 (current)
d.trueAt(\a);  // -> true (new behavior #6001)
440.trueAt     // -> false (now inconsistent)
440.falseAt    // -> true (now inconsistent)
d.falseAt(\a); // -> breaks (not supported!)
```

This behavior also contradicts what's said in the `Dictionary:-trueAt` help description:
"This method is also valid in link::Classes/Object::."

Unless I'm mistaken with this, I suggest this approval be reverted for now until this is worked out.

HelpSource/Classes/Dictionary.schelp
```diff
@@ -193,10 +193,22 @@ d.matchAt(\abc)
 ::
 
 method::trueAt
-Returns link::Classes/True:: if the item at the key is true, otherwise false. This method is also valid in link::Classes/Object::.
+Returns link::Classes/True:: if the item's booleanValue at the key is true, otherwise false. This method is also valid in link::Classes/Object::.
```
mtmccrea, line 196: See my comment in the Conversation thread regarding the inconsistent behavior with `Object:-trueAt`

mtmccrea (changes requested):

mtmccrea:
[Removing my previous post after looking at this more closely...]

There's an inconsistency between `Dictionary:-trueAt` and `Object:-trueAt`. I'm not sure whether this should be addressed in this PR, the recently merged #6001, and/or a new Issue/PR.

At least, this PR should probably remove the part of the `Dictionary:-trueAt` help description:
>"This method is also valid in Object" 

Which isn't true:

```supercollider
~testVal = 440;
d = Dictionary();
d.put(\a, ~testVal);

d.trueAt(\a);      // -> 440    (current)
d.trueAt(\a);      // -> true   (new behavior #6001)
d[\a].booleanValue // -> true   (new implementation for trueAt)
~testVal.trueAt    // -> false  (inconsistent with booleanValue implementation)
~testVal.falseAt   // -> true   (inconsistent with booleanValue implementation)
d.falseAt(\a);     // -> breaks (not supported!)
```

What's strange is that `Object:-falseAt` takes a `key:`  arg and `Object:-trueAt`doesn't, but it's called by `Object:-falseAt` and passed the key arg.

```
	trueAt { ^false }
	falseAt { arg key;
		^this.trueAt(key).not
	}
```

Not sure whether be making changing to the behavior of `Dictionary:-true/falseAt` without addressing `Object:-true/falseAt`, or if they should just be considered separate behaviors and remove the suggested connection in the docs.

telephon:
OK, there needs to be done more. I propose 

```

// object
falseAt { ^true }

// Dictionary
falseAt { arg key;
		^this.at(key).booleanValue.not
	}
```

This is the behaviour it had before, just clarified.

capital-G:
> OK, there needs to be done more.

Would it be possible to create a small patch in the meantime so our CI/test suite is stable again?

HelpSource/Classes/Dictionary.schelp
mtmccrea: Given the divergence in behavior with `Object:-falseAt`, I think we should just remove this reference to implied parallel behavior with `Object`:
`This method is also valid in link::Classes/Object::.`

mtmccrea:
For the sake of the patch, I think @telephon's suggested change makes sense. 

The inconsistency between Dictionary and Object implementations could then be just considered a separate issue?

I'm not sure of the use of `Object:-true/falseAt`. A quick search through the project, I don't see it called in the classlib. I'm not sure why all objects would be considered false:
```
// object
falseAt { ^true }
```
Fore example 
```
true.falseAt // -> true :-/
```
Anyway ... separate issue?

capital-G:
Maybe we can just make the test "pass" for now, and adjust it again after the discussion about how to properly handle this case is settled?

Something like this would make the test suite pass again which would be beneficial for some other PRs.

```diff
diff --git a/testsuite/classlibrary/TestDictionary.sc b/testsuite/classlibrary/TestDictionary.sc
--- a/testsuite/classlibrary/TestDictionary.sc
+++ b/testsuite/classlibrary/TestDictionary.sc
@@ -29,7 +29,7 @@ TestDictionary : UnitTest {
 
        test_trueAt {
                var dict = (x:7, y:false, z:true);
-               this.assert([\x, \y].every(dict.falseAt(_)) and: dict.trueAt(\z), "dictionary trueAt and falseAt should respond only true if the value really equals true.");
+               this.assert([\y].every(dict.falseAt(_)) and: [\x, \z].every(dict.trueAt(_)), "dictionary trueAt and falseAt should respond only true if the value really equals true.");
        }
 
 }
```

telephon:
yes it is ok to ignore the test for now.

Btw. personally, I would prefer if object.trueAt / falseAt would throw an error. It is a breaking change but that usage was never intended, I think.

mtmccrea:
Should we then just go ahead and merge this current PR as is? I think it should make the tests pasts with the recent changes of https://github.com/supercollider/supercollider/pull/6001.

 I can follow up with another PR with @telephon's suggested changes,

telephon:
yes, very good – thanks!

PR #6322, opened 2024-05-23
add choose relative weights

### LINKED ISSUE #6311: Make wchoose and windex use relative weights (normalize by default)
## Motivation

`wchoose` and `windex` don't normalize; that makes it easy to get unexpected results.

```
(
var list, pick;
list= [\red, \green, \blue];
pick = list.wchoose([100,200,300]);
)
```

Looking at the above code, one would expect to get `\blue` thrice as often and `\green` twice as often as `\red`. In reality,`\red` is the only element that is picked.

The correct option would have been to use `list.wchoose([100,100,100].normalizeSum)`. In practice, all lists with weights that are not normalized (and I assume that they are the majority) will need to have `.normalizeSum` applied.

The proposal is to make those functions normalize by default, as other languages as [python](https://docs.python.org/3/library/random.html#random.choices). If efficiency is a concern, a flag can be added for the arguably rare cases where the sum of the list is already equal to one.

Asmatzaile:
## Purpose and Motivation

`wchoose` and `windex` don't normalize; that makes it easy to get unexpected results. Instead of modifying them and potentially slowing down existing code, an optional argument is added to `.choose`.

Fixes #6311.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/SequenceableCollection.schelp
+++ b/HelpSource/Classes/SequenceableCollection.schelp
@@ -507,6 +507,10 @@ Choose an element from the collection at random.
 code::
 [1, 2, 3, 4].choose;
 ::
+There's an optional argument for specifying relative weights with a list. The length of both lists must be equal.
+code::
+[110, 220, 330, 440].choose([1, 3, 1, 2.5]);
+::
 
 method::wchoose
 Choose an element from the collection at random using a list of probabilities or weights. The weights must sum to 1.0.
--- a/SCClassLibrary/Common/Collections/SequenceableCollection.sc
+++ b/SCClassLibrary/Common/Collections/SequenceableCollection.sc
@@ -111,9 +111,16 @@ SequenceableCollection : Collection {
 	asSequenceableCollection { ^this }
 
 	// select an element at random
-	choose {
-		^this.at(this.size.rand)
+	choose { arg weights;
+		^if (weights.isNil) {
+			// with equal probabilities
+			this.at(this.size.rand)
+		} {
+			// with relative weights
+			this.wchoose(weights.normalizeSum)
+		}
 	}
+
 	// select an element at random using an array of weights
 	wchoose { arg weights;
 		^this.at(weights.windex)


HelpSource/Classes/SequenceableCollection.schelp
```diff
@@ -507,6 +507,10 @@ Choose an element from the collection at random.
 code::
 [1, 2, 3, 4].choose;
 ::
+There's an optional argument for specifying relative weights with a list. The length of both lists must be equal.
+code::
+[1, 2, 3, 4].choose([1, 2, 3, 4]);
```
telephon, line 512: this example may be a bit confusing for beginners.
Perhaps something like this?
```
[110, 220, 330, 440].choose([1, 3, 1, 2]);
```

SCClassLibrary/Common/Collections/SequenceableCollection.sc
```diff
@@ -111,8 +111,10 @@ SequenceableCollection : Collection {
 	asSequenceableCollection { ^this }
 
 	// select an element at random
-	choose {
-		^this.at(this.size.rand)
+	choose { arg weights=nil;
+		if (weights.isNil)
+			{^this.at(this.size.rand)}
+			{this.wchoose(weights.normalizeSum)};
```
telephon, line 117: You are proposing to change a method that many people use all the time. You should at least check if it works! This code doesn't.

Also it would be good if you keep in line with the standard formatting (look around in the same file).

telephon (changes requested):
Thanks! … I'll also make a forum thread, since I'd like to hear some opinions from others boiut this change.

SCClassLibrary/Common/Collections/SequenceableCollection.sc
Asmatzaile: Certainly!

HelpSource/Classes/SequenceableCollection.schelp
Asmatzaile: Certainly!

SCClassLibrary/Common/Collections/SequenceableCollection.sc
```diff
@@ -111,9 +111,16 @@ SequenceableCollection : Collection {
 	asSequenceableCollection { ^this }
 
 	// select an element at random
-	choose {
-		^this.at(this.size.rand)
+	choose { arg weights=nil;
```
telephon, line 114: you don't need to specify `nil`, it is the default. Just leave it out...

SCClassLibrary/Common/Collections/SequenceableCollection.sc
```diff
@@ -111,9 +111,16 @@ SequenceableCollection : Collection {
...
+		if (weights.isNil) {
+			// with equal probabilities
+			^this.at(this.size.rand)
```
telephon, line 117: you could factor out the `^` before the `if`. Then it is clear that we return something out of the method in any case.

SCClassLibrary/Common/Collections/SequenceableCollection.sc
```diff
@@ -111,9 +111,16 @@ SequenceableCollection : Collection {
...
+		} {
+			// with relative weights
+			^this.wchoose(weights.normalizeSum)
+		};
```
telephon, line 121: no need for the trailing `;`

telephon (changes requested):
Thanks! now it works, so just cosmetics. Let's see what the discussion on the forum results in.

telephon:
ok, good – thanks! Now let's wait and see what comes out of https://scsynth.org/t/choosing-how-to-choose-with-self-normalising-weights/9557/9

capital-G:
@telephon @Asmatzaile what is the status of this? Should we continue or close it?

dyfer:
Thanks @Asmatzaile for the PR!
It seems that there was a lengthy discussion on adding arguments vs a new method in https://github.com/supercollider/supercollider/pull/6601
Would you consider #6601 to supersede your PR?

Asmatzaile:
> Thanks @Asmatzaile for the PR! It seems that there was a lengthy discussion on adding arguments vs a new method in #6601 Would you consider #6601 to supersede your PR?

Yes!

telephon:
@Asmatzaile thank you for suggesting it!

PR #6327, opened 2024-05-27
GitHubIssue/PR Templates: update link and make wording more explicit.

JordanHendersonMusic:
## Purpose and Motivation

The contributing.md file hasn't existed for a while, it is now on the wiki.

Makes some of the wording a little more explicit to hopefully improve the quality of the bug reports.

## Types of changes

- gh issues/pr templates

## To-do list

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,8 @@ assignees: ''
 
 ---
 
-<!-- For informations about contributing see: https://github.com/supercollider/supercollider/wiki/Contributing-directory -->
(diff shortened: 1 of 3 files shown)


cdbzb:
this looks great @JordanHendersonMusic just wanted to be sure you had seen this RFC https://github.com/supercollider/rfcs/pull/21 which could possibly change the location of the wiki

JordanHendersonMusic:
Ah no I hadn't actually! It looks like a good idea to me.

Updating all the links after that would have to happen anyway, so doesn't really interfere with this, actually this makes it easier because you can grep for the SC wiki URL.

cdbzb (approved):

PR #6328, opened 2024-05-28
Bump GH actions to node 20 versions

capital-G:
## Purpose and Motivation

The current build process generates some warnings by using outdated GH actions - this PR upgrades the actions from `v3` to `v4`. The  currently used`v3` version [will be deprecated in a few months](https://github.com/actions/upload-artifact/tree/v4/?tab=readme-ov-file#actionsupload-artifact).

See also https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

grafik

Maybe it would be a good idea to let a dependabot watch over this - I could set this up.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -20,7 +20,7 @@ jobs:
     outputs:
       sc-version: ${{ steps.set-version.outputs.version }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


mtmccrea (approved):
Thanks for taking care of this!

PR #6329, opened 2024-05-29
replace "Monospace" with system monospace font

capital-G:
## Purpose and Motivation

During startup of the IDE, it yields a log message

```
qt.qpa.fonts: Populating font family aliases took 242 ms. Replace uses of missing font family "Monospace" with one that exists to avoid this cost. 
```

This can also take longer (500ms), depending on your system, slowing down every startup of the IDE.

This PR replaces the font with the default monospace font of the system and therefore removes the error message.
I increased the font size a bit in order to match the current size (only checked this on MacOS).

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/util/status_box.cpp
+++ b/editors/sc-ide/widgets/util/status_box.cpp
@@ -31,7 +31,7 @@ StatusLabel::StatusLabel(QWidget* parent): QLabel(parent) {
     setTextColor(Qt::white);
 
     QFont font = QFontDatabase::systemFont(QFontDatabase::FixedFont);
-    font.setPointSize(font.pointSize() + 3);
+    font.setPointSize(font.pointSize());
     font.setStyleHint(QFont::Monospace);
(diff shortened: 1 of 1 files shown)


telephon (approved):

mtmccrea:
Tested (also on mac) with different font sizes and it looks like the enclosing text box resizes with the font size, so I wouldn't expect any surprises like text overshooting on different platforms. If you agree, @capital-G, I can go ahead and merge this.

capital-G:
I switched back to the default font size which actually matches the current way. I rebased the branch, if it passes it can be merged.

mtmccrea:
Thanks!

mtmccrea:
Sorry I added the wrong `comp` tag to the squashed commit. I was testing your other doc icon PR at the same time and got my wires crossed.

PR #6331, opened 2024-05-30
Replace `boost::filesystem` with `std::filesystem`

capital-G:
## Purpose and Motivation

With C++17, the API of [`boost::filesystem`](https://www.boost.org/doc/libs/master/libs/filesystem/doc/index.htm) has been adapted into [`std::filesystem`](https://en.cppreference.com/w/cpp/filesystem).
As SuperCollider currently already [targets C++17 officially](https://github.com/supercollider/supercollider/blob/24c716421b9fa9e3679df4e961b9433c2e1dbbd8/README.md?plain=1#L13), this PR proposes to replace `boost::filesystem` with `std::filesystem`

**Pro:**
* reduces denpendencies
* uses C++ standards which may help us in the future
* reduces compilation time in theory
* makes boost upgrades more easy and already fixes https://github.com/supercollider/supercollider/issues/5865
* As we also support system boost libraries we would need to adapt the code for boost 1.85. Recently brew updated to [boost 1.85](https://www.boost.org/users/history/version_1_85_0.html) which [failed](https://github.com/supercollider/supercollider/actions/runs/9187974816/job/25266805454#step:18:2557) our *macOS with boost system libraries* pipeline, therefore we would need to adjust the filesystem code already
  <details>
    <summary>More info</summary>
  
  The build chain *macOS with boost system libraries*  recently because of an boost-upgrade of brew which changed the  `boost::filesystem` API.
  
  ```
   /Users/runner/work/supercollider/supercollider/server/supernova/sc/sc_synth_definition.cpp:67:37: error: use of undeclared identifier 'recursive_directory_iterator'
         boost::make_iterator_range(recursive_directory_iterator(dir), recursive_directory_iterator())) {
  ```

  This is due [a change](https://www.boost.org/users/history/version_1_85_0.html) in the `boost::filesystem` API v.1.85?
  
  > On POSIX systems that support [fdopendir](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fdopendir.html), openat and related APIs defined in POSIX.1-2008, as well as on Windows, `recursive_directory_iterator` now uses file descriptors/handles instead of paths to query the file statuses and open nested directories during iteration. This makes directory iteration more resilient to concurrent modifications of the filesystem.
  </details>

**Con:**
* makes C++17 support mandatory for the compiler - this drops some older compilers such as
  * g++7 - which I think is OK b/c the compiler is from 2017 and the oldest supported Debian version already uses g++8.3 [which supports](https://www.reddit.com/r/cpp/comments/7o9kg6/gcc_80_support_stdfilesystem_include_filesystem/) `std::filesystem` 
  * ~~mingw: uses an older version of g++8 which only supports C++17 in an experimental state. We can't update mingw that easily because Qt5 only supports mingw 8.1, see https://doc.qt.io/qt-5/windows.html - with a potential Qt6 upgrade it would allow us to use mingw11, see https://doc.qt.io/qt-6/windows.html. Maybe it is possible to make this available via Cmake flags but I could only test this via CI as I don't have a windows machine - any help or expertise for this platform would be great! :)~~ -> **UPDATE**: With a bump in the windows OS version from 2019->2022 this now works again.
  * We would need to drop macOS Mojave (10.14, got deprecated 2021) and below, so the next version of SuperCollider would only run on the 64bit-only macOS Catalina (10.15, released 2019) and beyond.

This is not ready for review yet and more an explicit kickstarter for a discussion.
I would be interested in still supporting mingw, maybe by switching to msys2?

~~There is also a file related test which fails `TestQuark:test_runQuarkInstallHooks`, see https://github.com/capital-G/supercollider/actions/runs/9300444791/job/25597559238#step:9:6936~~ -> Should be fixed via https://github.com/capital-G/supercollider/commit/f48ff264d17457578eb48073d7219f8db1b78f0b

A CI run based on https://github.com/supercollider/supercollider/pull/6331/commits/f48ff264d17457578eb48073d7219f8db1b78f0b with deactivated mingw can be found here https://github.com/capital-G/supercollider/actions/runs/9302363265

We could also replace `boost::container` with `std::container` in a similar manner - but this did not produce any problems so far.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/common/SC_Filesystem_win.cpp
+++ b/common/SC_Filesystem_win.cpp
@@ -67,7 +67,7 @@ SC_Filesystem::Glob* SC_Filesystem::makeGlob(const char* pattern) {
     // remove a trailing backslash. Even if searching with 'foo/.', this will
     // change to 'foo' harmlessly. Use has_parent_path() because otherwise '.'
     // (referring to the CWD) won't be recognized as a valid query.
-    if (path.filename_is_dot() && path.has_parent_path())
+    if (path.filename() == "." && path.has_parent_path())
         path = path.parent_path();
 
     // expand to home directory, if path starts with tilde
@@ -103,7 +103,7 @@ Path SC_Filesystem::globNext(Glob* glob) {
 
         if (!::FindNextFileW(glob->mHandle, &glob->mEntry))
             glob->mAtEnd = true;
-    } while (glob->mFilename.filename_is_dot() || glob->mFilename.filename_is_dot_dot());
+    } while (glob->mFilename.filename() == "." || glob->mFilename.filename() == "..");
 
     // add preferred separator (L'\\') for directories on Windows, to match
     // POSIX globbing. boost::filesystem::is_directory won't work for this because
--- a/common/SC_Filesystem_win.cpp
+++ b/common/SC_Filesystem_win.cpp
@@ -30,8 +30,7 @@
 #    include "SC_Filesystem.hpp"
 #    include "SC_Codecvt.hpp"
 
-// boost
-#    include <boost/filesystem/operations.hpp> // is_directory
+#    include <filesystem>
 
 // system
 #    include <Shlobj.h> // SHGetKnownFolderPath
@@ -63,7 +62,7 @@ SC_Filesystem::Glob* SC_Filesystem::makeGlob(const char* pattern) {
     Glob* glob = new Glob;
 
     // use make_preferred() to change / -> \ on Windows
-    boost::filesystem::path path = SC_Codecvt::utf8_str_to_path(pattern).make_preferred();
+    std::filesystem::path path = SC_Codecvt::utf8_str_to_path(pattern).make_preferred();
 
     // remove a trailing backslash. Even if searching with 'foo/.', this will
     // change to 'foo' harmlessly. Use has_parent_path() because otherwise '.'
@@ -111,7 +110,7 @@ Path SC_Filesystem::globNext(Glob* glob) {
     // in the case of input '.' and '..', the filename here is just a single folder,
     // not a relative path, and so can't be correctly identified. Plus, it's faster
     // to check the attributes than to make another system call.
-    return isDirectory ? glob->mFilename += boost::filesystem::path::preferred_separator : glob->mFilename;
+    return isDirectory ? glob->mFilename += std::filesystem::path::preferred_separator : glob->mFilename;
 }
 
 //============= PRIVATE METHODS ==============//
--- a/lang/LangPrimSource/PyrFilePrim.cpp
+++ b/lang/LangPrimSource/PyrFilePrim.cpp
@@ -110,6 +110,8 @@ int prFileDeleteAll(struct VMGlobals* g, int numArgsPushed) {
 
 std::time_t to_time_t(std::filesystem::file_time_type file_time) {
     using namespace std::chrono;
+    // in case of C++20 update, consider
+    // https://stackoverflow.com/questions/61030383/how-to-convert-stdfilesystemfile-time-type-to-time-t
     auto sctp = time_point_cast<system_clock::duration>(file_time - std::filesystem::file_time_type::clock::now()
                                                         + system_clock::now());
     return system_clock::to_time_t(sctp);
--- a/lang/LangPrimSource/PyrFilePrim.cpp
+++ b/lang/LangPrimSource/PyrFilePrim.cpp
@@ -49,9 +49,7 @@ Primitives for File i/o.
 
 /* C++ stdlib headers */
 #include <tuple>
-
-/* boost headers */
-#include <boost/filesystem.hpp>
+#include <filesystem>
 
 /* system headers */
 #ifndef _WIN32
@@ -70,7 +68,7 @@ Primitives for File i/o.
 
 #define DELIMITOR ':'
 
-namespace bfs = boost::filesystem;
+namespace fs = std::filesystem;
 
 int prFileDelete(struct VMGlobals* g, int numArgsPushed) {
     PyrSlot *a = g->sp - 1, *b = g->sp;
@@ -80,9 +78,9 @@ int prFileDelete(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p = SC_Codecvt::utf8_str_to_path(filename);
-    boost::system::error_code error_code;
-    bfs::remove(p, error_code);
+    const fs::path& p = SC_Codecvt::utf8_str_to_path(filename);
+    std::error_code error_code;
+    fs::remove(p, error_code);
 
     if (error_code)
         SetFalse(a);
@@ -100,8 +98,8 @@ int prFileDeleteAll(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p = SC_Codecvt::utf8_str_to_path(filename);
-    if (bfs::remove_all(p) > 0) {
+    const fs::path& p = SC_Codecvt::utf8_str_to_path(filename);
+    if (fs::remove_all(p) > 0) {
         SetTrue(a);
     } else {
         SetFalse(a);
@@ -118,9 +116,10 @@ int prFileMTime(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p = SC_Codecvt::utf8_str_to_path(filename);
-    time_t mtime = bfs::last_write_time(p);
-    SetInt(a, mtime);
+    const fs::path& p = SC_Codecvt::utf8_str_to_path(filename);
+    auto mtime = fs::last_write_time(p);
+    const time_t time = decltype(mtime)::clock::to_time_t(mtime);
+    SetInt(a, time);
     return errNone;
 }
 
@@ -132,8 +131,8 @@ int prFileExists(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p = SC_Codecvt::utf8_str_to_path(filename);
-    bool res = bfs::exists(p);
+    const fs::path& p = SC_Codecvt::utf8_str_to_path(filename);
+    bool res = fs::exists(p);
     SetBool(a, res);
     return errNone;
 }
@@ -149,13 +148,13 @@ int prFileRealPath(struct VMGlobals* g, int numArgsPushed) {
         return err;
 
     bool isAlias = false;
-    bfs::path p = SC_Codecvt::utf8_str_to_path(ipath);
+    fs::path p = SC_Codecvt::utf8_str_to_path(ipath);
     p = SC_Filesystem::resolveIfAlias(p, isAlias);
     if (p.empty())
         return errFailed;
 
-    boost::system::error_code error_code;
-    p = bfs::canonical(p, error_code).make_preferred();
+    std::error_code error_code;
+    p = fs::canonical(p, error_code).make_preferred();
     if (error_code) {
         SetNil(a);
         return errNone;
@@ -178,8 +177,8 @@ int prFileMkDir(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p = SC_Codecvt::utf8_str_to_path(filename);
-    bool result = bfs::create_directories(p);
+    const fs::path& p = SC_Codecvt::utf8_str_to_path(filename);
+    bool result = fs::create_directories(p);
 
     SetBool(a, result);
     return errNone;
@@ -197,10 +196,10 @@ int prFileCopy(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p1 = SC_Codecvt::utf8_str_to_path(filename1);
-    const bfs::path& p2 = SC_Codecvt::utf8_str_to_path(filename2);
-    boost::system::error_code error_code;
-    bfs::copy(p1, p2, error_code);
+    const fs::path& p1 = SC_Codecvt::utf8_str_to_path(filename1);
+    const fs::path& p2 = SC_Codecvt::utf8_str_to_path(filename2);
+    std::error_code error_code;
+    fs::copy(p1, p2, error_code);
     if (error_code) {
         std::ostringstream s;
         s << error_code.message() << ": copy from \"" << filename1 << "\" to \"" << filename2 << "\"";
@@ -218,9 +217,9 @@ int prFileType(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p = SC_Codecvt::utf8_str_to_path(filename);
-    bfs::file_status s(bfs::symlink_status(p));
-    SetInt(a, s.type());
+    const fs::path& p = SC_Codecvt::utf8_str_to_path(filename);
+    fs::file_status s(fs::symlink_status(p));
+    SetInt(a, static_cast<int>(s.type()));
     return errNone;
 }
 
@@ -232,8 +231,8 @@ int prFileSize(struct VMGlobals* g, int numArgsPushed) {
     if (error != errNone)
         return error;
 
-    const bfs::path& p = SC_Codecvt::utf8_str_to_path(filename);
-    uintmax_t sz = bfs::file_size(p);
+    const fs::path& p = SC_Codecvt::utf8_str_to_path(filename);
+    uintmax_t sz = fs::file_size(p);
     SetInt(a, sz);
     return errNone;
 }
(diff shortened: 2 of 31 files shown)


capital-G:
One question: Do we actually support mingw? The CI has a build step but I also found

https://github.com/supercollider/supercollider/blob/24c716421b9fa9e3679df4e961b9433c2e1dbbd8/README_WINDOWS.md?plain=1#L229-L233

r900:
not sure if that helps, but i got this to compile and run on a gentoo ~amd64 system with boost-1.85.0

capital-G:
That is good information! Can you verify that something like File loading or saving actually works via sclang? 

I found the document https://github.com/supercollider/supercollider/wiki/Platform-Support,  which states

> As much as possible, SuperCollider should support building with the latest release of Boost, and the version packaged in the source tree should be updated reasonably soon after a release.

Therefore this PR complies with these goals (as seen in #6334 ), it even improves them - so I think this could be ready for review.

The main question is therefore: Should we drop mingw support in order to be fully compatible with C++17 which will help us down the road?

telephon:
Maybe we can go ahead for now and support mingw again once it has support for Boost 1.85 ? Currently it goes up to 1.84 (https://nuwen.net/mingw.html).

r900:
> That is good information! Can you verify that something like File loading or saving actually works via sclang?

gave it a try just now.
loading a sample into a buffer works, and recording output via s.record works as well

are those enough, or is there something else i can test this with?

telephon:
You could try sclang, too. Maybe load the file in `SoundFile`. Also `resolveRelative` should return the right path.

capital-G:
> Maybe we can go ahead for now and support mingw again once it has support for Boost 1.85 ? Currently it goes up to 1.84 (https://nuwen.net/mingw.html).

The issue is in this case not mingw but qt, as qt only supports and provides qt5 binaries for Mingw81 - building qt from source is difficult, especially if we use a compiler version which is not officially supported.

But we could use [msys2](https://www.msys2.org/), which includes a package manager for the mingw environment which could allow to combine qt5 with a more recent version of mingw? But it would be good to have a Windows machine to test this on and not use the CI to set this up :)

Additionally, I don't really know what the benefit of a mingw build is (beyond using unix tools to build?) and if it is really relevant for someone? I am not really familiar with the Windows environment anymore.

> loading a sample into a buffer works, and recording output via s.record works as well

This already look promising 👍 
Can you test if sclang can also open/read/write files, e.g. through https://docs.supercollider.online/Classes/File.html
Also if it fixes this bug https://github.com/supercollider/supercollider/issues/5865 ?

r900:
> Can you test if sclang can also open/read/write files, e.g. through https://docs.supercollider.online/Classes/File.html

i ran all the examples at the bottom of that page, no problems.

then copied a file, a directory and symlink: no errors
copying from one partition to another also worked

just some weird behaviour:
when copying a directory it ignored the subdirectories within the source directory. files directly in the source dir were copied, directories ignored.

also tried File.delete, File.exists, File.realpath, File.mkdir , worked as expected

i never ran into bug https://github.com/supercollider/supercollider/issues/5865 , so that's hard to recreate for me.
"Invalid cross-device link:" sounds a lot like it tried to create a hard link to a different partition/filesystem. none of my copying tests triggered this error

r900:
> Also `resolveRelative` should return the right path.

tested with a file in /tmp:
`"../usr/bin".resolveRelative;` -> `/tmp/../usr/bin`
`File.realpath("../../usr/./../../../usr/bin".resolveRelative);` -> `/usr/bin`

capital-G:
> just some weird behaviour:
> when copying a directory it ignored the subdirectories within the source directory. files directly in the source dir were copied, directories ignored.

Interesting, but this is also the behavior I can replicate in 3.13, so this seems not to have changed.

JordanHendersonMusic:
Are there UnitTests for Path and other file system stuff that would touch these functions?

dyfer:
> One question: Do we actually support mingw? The CI has a build step but I also found

Not officially :) I was the one who added MinGW to the build matrix. 
I believe @sonoro1234 uses MinGW to build supernova, so while we have no obligation to support that toolchain based on the agreed on policies, the job is there so that we to know if we break MinGW build, as a courtesy :)

Historically (from memory so this might be inaccurate), SC on Windows was first supported by MinGW only, since it easier to migrate to from Linux. However, Qt for MinGW does not provide the WebEngine, so once the IDE with help system running in the QtWebEngine came around, the supported build system was changed to MSVC.

bagong:
Just for the record: back in the day there were two competing trends: stay close to nix ways of doing things, basically using gcc wherever possible vs using "native" libraries/compilers/SDKs. What flipped the coin towards VS was that qtwebengine didn't make it into MinGW-*64*, so we were stuck on 32-bit on Windows for a while, until mainly Lucas Cornelisse, Luis Lloret and later Brian Heim got Win64 compilable with VS. Supernova and sc3-plugins were lagging behind in that shift, so that at that point in time we even had hybrid experimental releases.

smoge:
> > Are there UnitTests for Path and other file system stuff that would touch these functions?
> 
> I know that at least the Quarks test access the filesystem, see https://github.com/capital-G/supercollider/actions/runs/9302363265/job/25605094051#step:9:6901 (from a CI run where I disabled mingw so the tests were run)
> 
> It would be great to get people from windows testing this as the automatic testing on windows is currently not available.
> 

Have you managed to use Msys2 (the package manager for MinGW) on the CI? It was straightforward to resolve all issues in question here, as I understood from a recent conversation. 

Maybe a call for testers?

capital-G:
Rebased w/ bumping the windows version for the mingw build - let's wait for the build pipeline.

dyfer:
Thanks for this.
Without reviewing the C++ portion of this on a technical level, I'm cautiously in favor for this.

I know this is in the description of the PR but I want to spell it out - **this will raise the oldest supported macOS to 10.15**. I think this is the biggest drawback to consider. However, I think we supported 10.11 long enough, but I want to make sure that everybody is okay with this change. There was a user on macOS 10.13 who inquired after the SC 3.13 release where the legacy build indeed did not work and needed to be fixed. 

At the same time, there's no easy path forward without this - the only way around would be to keep using `boost::filesystem` but provide `#ifdef`s to support multiple boost versions in order to introduce compatibility with the current boost versions. That would be messy and probably not worth the effort...

ahihi:
just my perspective as someone still running a few 10.14 systems — the support so far is greatly appreciated, but this OS is in many ways becoming impractical to run, with even Firefox ESR support ending next month. so i think dropping support is fair and personally i am fine with keeping these systems on SC 3.13 for their remaining lifetime.

dyfer:
Thanks for your input @ahihi ! Much appreciated.

dyfer:
I tested the Windows build on a Windows 10 machine.

I performed the file operations below and they all seem have worked.

*I found one discrepancy*, which seems to be a bugfix, compared to 3.13 release:

```supercollider
File.copy("~/Desktop".standardizePath, "~/Desktop/desktopCopy".standardizePath); 
```

The build from this PR copied all Desktop files into `Desktop/desktopCopy` (except the subdirectories), which is the expected behavior AFAIU; I could recreate the same on macOS running the 3.13 release. However, on Windows 3.13 release the same command copied all the files into `~/Desktop/Desktop` (!). This might be an issue of handling forward/backward slashes (`.standardizePath` results in a mix of forward and backward slashes), or in handling trailing directories; in any case, this PR seems to introduce the correct behavior. 

IMO this is good to be merged, from the functionality perspective. @telephon @JordanHendersonMusic what do you think?

For reference, here are the commands I ran to test this on Windows:
<details>

```supercollider
(
var f, g;
f = File("~/test.txt".standardizePath,"w");
f.write("Does this work?\n is this thing on ?\n");
f.close;
)

// read it again:
(
g = File("~/test.txt".standardizePath,"r");
g.readAllString.postln;
g.close;
)

// try the above with File.use:

File.use("~/test.txt".standardizePath, "w", { |f| f.write("Doesn't this work?\n is this thing really on ?\n"); });
File.use("~/test.txt".standardizePath, "r", { |f| f.readAllString.postln });


// more file writing/reading examples:
(
var h, k;
h = File("~/test.dat".standardizePath.postln, "wb");
h.inspect;
h.write( FloatArray[1.1, 2.2, 3.3, pi, 3.sqrt] );
h.close;

k = File("~/test.dat".standardizePath, "rb");
(k.length div: 4).do({ k.getFloat.postln; });
k.close;
)


(
var f, g;
f = File("~/test.txt".standardizePath.postln,"w");
100.do({ f.putChar([$a, $b, $c, $d, $e, $\n].choose); });
f.close;

g = File("~/test.txt".standardizePath,"r");
g.readAllString.postln;
g.close;

g = File("~/test.txt".standardizePath,"r");
g.getLine(1024).postln;
"*".postln;
g.getLine(1024).postln;
"**".postln;
g.getLine(1024).postln;
"***".postln;
g.getLine(1024).postln;
"****".postln;
g.close;
)

(
//var f, g;
f = File("~/test.dat".standardizePath,"wb");
f.inspect;
100.do({ f.putFloat(1.0.rand); });

f.inspect;
f.close;

g = File("~/test.dat".standardizePath,"rb");
100.do({
    g.getFloat.postln;
});
g.inspect;
g.close;
)

(
//var f, g;
f = File("~/test.dat".standardizePath,"r");
f.inspect;
f.close;
)

// ----SoundFile

(
f = SoundFile.new;
f.openRead(Platform.resourceDir +/+ "sounds/a11wlk01.wav");
f.inspect;
f.close;
)

// ---- Buffer
// read a soundfile
s.boot;
p = Platform.resourceDir +/+ "sounds/a11wlk01.wav";
b = Buffer.read(s, p);
b.plot;

// now play it
(
x = SynthDef(\help_Buffer, { arg out = 0, bufnum;
    Out.ar( out,
        PlayBuf.ar(1, bufnum, BufRateScale.kr(bufnum))
    )
}).play(s,[\bufnum, b]);
)
x.free; b.free;
s.quit;

// other
"hello".resolveRelative; //-> C:/Users/username/Desktop\hello


File.realpath("c:/Windows/System32/../System32/kernel32.dll"); //-> C:\Windows\System32\kernel32.dll


File.exists("C:\\Windows\\System32\\kernel32.dll");
File.exists("C:/Windows/System32/kernel32.dll");

File.copy("~/Desktop".standardizePath, "~/Desktop/desktopCopy".standardizePath); //with std::filesystem copied all Desktop files into Desktop/desktopCopy (except the subdirectories); on 3.13 it copied all the files into "~/Desktop/Desktop", which seems to be a bug
```
</details>

sonoro1234:
Tried to build with gcc 13

```
[ 28%] Building CXX object server/scsynth/CMakeFiles/libscsynth.dir/__/__/common/SC_PaUtils.cpp.obj
C:\supercolliderrepos\supercollider\common\SC_PaUtils.cpp: In function 'PaError TryGetDefaultPaDevices(PaDeviceIndex*, PaDeviceIndex*, int, int, double)':
C:\supercolliderrepos\supercollider\common\SC_PaUtils.cpp:130:21: error: 'uint32_t' was not declared in this scope
  130 |                 if (uint32_t(inSR) != uint32_t(outSR)) {
      |                     ^~~~~~~~
C:\supercolliderrepos\supercollider\common\SC_PaUtils.cpp:6:1: note: 'uint32_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
    5 | #include <cstdio>
  +++ |+#include <cstdint>
    6 |
mingw32-make[2]: *** [server\scsynth\CMakeFiles\libscsynth.dir\build.make:511: server/scsynth/CMakeFiles/libscsynth.dir/__/__/common/SC_PaUtils.cpp.obj] Error 1
```

capital-G:
> Tried to build with gcc 13 from mingw-w64

Thanks for reporting - the error seems unrelated to this PR though. Have you tried the suggested fix by adding `#include <cstdint>`?
Link to offending line: https://github.com/capital-G/supercollider/blob/03aa08ba50e2a26f84785ab065e0fee958a0d9d4/common/SC_PaUtils.cpp#L130

I think if someone is interested in providing mingw/gcc13 support, it should be a PR on its own after this one has been merged.

dyfer:
@sonoro1234 let us know if you had a chance to test this
@capital-G do you think it's worth rebasing this to see that green checkmark next to the test CI?

capital-G:
Rebased against develop.

If everything looks good, maybe we should just merge it? Develop/Nightly is "ok" to fail IMO, caution is justified but we can't be 100% sure about these things until some people download the nightly build on their machines and report problems, because so far no problems have been observed with this branch.

dyfer (dismissed):
I'm in favor of merging this.

dyfer:
> I have pulled again an could build but I am not sure about std:filesystem being used because there were boost/libs/filesystem files present in the build

Did file operations work OK in your build? You can see my [comment](https://github.com/supercollider/supercollider/pull/6331#issuecomment-2325302483) for the possible test code (in the "details" section).

capital-G:
@JordanHendersonMusic can you give this a review? :)

JordanHendersonMusic:
I probably won't have time this week - maybe at the weekend.

sonoro1234:
> > I have pulled again an could build but I am not sure about std:filesystem being used because there were boost/libs/filesystem files present in the build
> 
> @sonoro1234 did file operations work OK in your build? You can see my [comment](https://github.com/supercollider/supercollider/pull/6331#issuecomment-2325302483) for the possible test code (in the "details" section).

It copies to desktopCopy only files not subfolders

dyfer:
> > > I have pulled again an could build but I am not sure about std:filesystem being used because there were boost/libs/filesystem files present in the build
> > 
> > 
> > @sonoro1234 did file operations work OK in your build? You can see my [comment](https://github.com/supercollider/supercollider/pull/6331#issuecomment-2325302483) for the possible test code (in the "details" section).
> 
> It copies to desktopCopy only files not subfolders

@sonoro1234 thanks for testing! This is the expected behavior (well, at least consistent with the other platforms).

dyfer:
@capital-G would would you think about pulling in https://github.com/supercollider/supercollider/pull/6457, then rebasing this PR? That way I could proceed with the Qt updates without introducing conflicts with this PR.

lang/LangPrimSource/PyrFilePrim.cpp
```diff
@@ -110,6 +108,13 @@ int prFileDeleteAll(struct VMGlobals* g, int numArgsPushed) {
     return errNone;
 }
 
+std::time_t to_time_t(std::filesystem::file_time_type file_time) {
+    using namespace std::chrono;
+    auto sctp = time_point_cast<system_clock::duration>(file_time - std::filesystem::file_time_type::clock::now()
```
JordanHendersonMusic, line 113: Seems this could fail if files are edited really fast, but this is unlikely.

Here is a whole stack overflow post on exactly this issue.

https://stackoverflow.com/questions/61030383/how-to-convert-stdfilesystemfile-time-type-to-time-t

Perhaps it might be worth adding a comment here in case we ever do update to C++ 20?

JordanHendersonMusic (changes requested):

lang/LangPrimSource/PyrFilePrim.cpp
capital-G, line 113: So something along
```suggestion
    // in case of C++20 update, consider
    // https://stackoverflow.com/questions/61030383/how-to-convert-stdfilesystemfile-time-type-to-time-t
    auto sctp = time_point_cast<system_clock::duration>(file_time - std::filesystem::file_time_type::clock::now()
```
?

JordanHendersonMusic: sure!

capital-G:
@JordanHendersonMusic added the comment via https://github.com/supercollider/supercollider/pull/6331/commits/867a5d6d4d8f2f7cca06e191e7715f7190b1bb26

Is there any other obstacle left from your POV?

JordanHendersonMusic (approved):

common/SC_Filesystem_win.cpp
```diff
@@ -104,14 +103,14 @@ Path SC_Filesystem::globNext(Glob* glob) {
...
-    } while (glob->mFilename.filename_is_dot() || glob->mFilename.filename_is_dot_dot());
+    } while (glob->mFilename.filename() == "." || glob->mFilename.filename() == "..");
 
     // add preferred separator (L'\\') for directories on Windows, to match
     // POSIX globbing. boost::filesystem::is_directory won't work for this because
     // in the case of input '.' and '..', the filename here is just a single folder,
     // not a relative path, and so can't be correctly identified. Plus, it's faster
     // to check the attributes than to make another system call.
-    return isDirectory ? glob->mFilename += boost::filesystem::path::preferred_separator : glob->mFilename;
+    return isDirectory ? glob->mFilename += std::filesystem::path::preferred_separator : glob->mFilename;
```
capital-G, line 113: Maybe this is the origin of #6501 ?

Issue #6333, opened 2024-05-30
Revert temporary fix on using system boost version macOS build


capital-G:
Currently we support to use the system boost version of the OS instead of the one provided via `external_libraries/boost`.
We support this also on macOS, but as brew upgraded to boost 1.85, this introduced problems on the filesystem lib.
In order to mitigate this issue, we decided to fixate the boost version on macOS via [`brew install boost@1.76`](https://formulae.brew.sh/formula/boost@1.76#default) in https://github.com/supercollider/supercollider/pull/6320   - but this is already deprecated so can only be a temporary fix until we actually upgrade boost or make the source code compatible with the newer boost version.

This was done in order to get the CI running/passing again as it blocked running the tests.

## Environment

* SuperCollider version: dev
* Operating system: macOS >11
* Other details (Qt version, audio driver, etc.): use system boost build flag

capital-G:
Fixed via https://github.com/supercollider/supercollider/pull/6476

Issue #6244, opened 2024-03-31
build failure


prko:
## Environment

* SuperCollider version: 3.14.0_dev
* Operating system: macOS 14.4.1

## Steps to reproduce
in terminal `(base) prko@prkoMBP2021 ~ % `:
```
cd SuperCollider
mkdir -p build
cd build
cmake -G Xcode -DCMAKE_PREFIX_PATH=`brew --prefix qt5`  ..
```
then
```
cmake --build . --target install --config RelWithDebInfo
```
## Expected vs. actual behavior
- Expected:  a sc build.
- actual: failure.

The followings are error messages:
```
In file included from /Users/prko/GitHub/SuperCollider_related/toBuild/SuperCollider/external_libraries/boost/boost/type_index/stl_type_index.hpp:47:
/Users/prko/GitHub/SuperCollider_related/toBuild/SuperCollider/external_libraries/boost/boost/container_hash/hash.hpp:131:33: error: 
      no template named 'unary_function' in namespace 'std'; did you mean '__unary_function'?
        struct hash_base : std::unary_function<T, std::size_t> {};
                           ~~~~~^~~~~~~~~~~~~~
                                __unary_function
```
```
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.4.sdk/usr/include/c++/v1/__functional/operations.h:15:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.4.sdk/usr/include/c++/v1/__functional/unary_function.h:46:1: note: 
      '__unary_function' declared here
using __unary_function = __unary_function_keep_layout_base<_Arg, _Result>;
^
```
```
In file included from /Users/prko/GitHub/SuperCollider_related/toBuild/SuperCollider/external_libraries/boost/boost/function/function_base.hpp:22:
/Users/prko/GitHub/SuperCollider_related/toBuild/SuperCollider/external_libraries/boost/boost/type_traits/has_trivial_copy.hpp:34:4: warning: 
      builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead [-Wdeprecated-builtins]
   BOOST_HAS_TRIVIAL_COPY(T) BOOST_TT_TRIVIAL_CONSTRUCT_FIX
   ^
```
```
[...]

PR #6246, opened 2024-04-03
build-system: cmake: added compiler flags for c++17

doubtfulpalace:
## Purpose and Motivation

Fixes #6244 (on my Mac at least).

## Types of changes

- Bug fix

## To-do list

- [X ] Code is tested // i.e., builds
- [X ] All tests are passing // ctest only. Crash tests are supposed to segfault, right?
- [n/a ] Updated documentation
- [ X] This PR is ready for review

PR #6334, opened 2024-05-30
Upgrade to boost 1.83

capital-G:
* This is built on top of #6331 which makes the upgrade not too complicated

---

* Fixes https://github.com/supercollider/supercollider/issues/6224
* Fixes https://github.com/supercollider/supercollider/issues/6244
* Fixes https://github.com/supercollider/supercollider/issues/5865
* Fixes https://github.com/supercollider/supercollider/issues/6333
* 

---

* Potentially supersedes https://github.com/supercollider/supercollider/pull/6246
* Potentially supersedes https://github.com/supercollider/supercollider/pull/6231
* Potentially supersedes https://github.com/supercollider/supercollider/pull/5924

## Purpose and Motivation

As macOS are starting to push forward regarding C++ standards and brew has a rolling release structure, it seems that compiling older boost parts within SuperCollider becomes a bit harder and is only possible by setting flags such as `_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION)` and `no-enum-constexpr-conversion` (as seen in https://github.com/supercollider/supercollider/pull/6246/files), but this does not fix issues such as #6333.
Instead this PR suggests to update boost to a version which is fully compatible with C++17, see https://www.boost.org/doc/libs/1_83_0/doc/html/container/cpp_conformance.html

This PR bumps the boost version to 1.83, which is the current version on Debian testing, see https://packages.debian.org/en/sid/libboost-dev

I am not a cmake/c++/boost expert and is my first time working with them, so please be kind :) I am open to any suggestions and/or edits.

A CI which comments out mingW (see #6331 ) and builds and tests successfully can be found @ https://github.com/capital-G/supercollider/actions/runs/9302263387

This PR is not ready yet, but I think it would be an interesting kickstarter for a conversation on how we could get a new boost version running.

**The inline patches are not applied yet!**

**Known Problem**: This will break (at least) the flucoma plugin as it uses a hack to access `SendReply` which is not exposed via the Plugin API - this requires the plugin to access some internal functionality which requires boost, therefore breaking the binary compatibility due to version mismatch which will lead to a crash of the server. This **can not** be fixed if we want to upgrade boost. See https://github.com/flucoma/flucoma-sc/issues/149 and https://github.com/supercollider/supercollider/pull/5924#issuecomment-1356873575

Potential fixes are
* Re-compile flucoma-sc for the new SC version - the new extension version would only be compatible with the new SC version and vice versa
* Fix the problem for further updates by switching to a C API https://github.com/supercollider/supercollider/issues/6145 and/or adding SendReply functionality to the plugin API via https://github.com/supercollider/supercollider/issues/5939 - this circumstances stated above would still apply

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation


smoge:
Does this change fix the deprecation error for MinGW?

Compiler error:

```
In file included from D:\a\supercollider\supercollider\editors\sc-ide\core\doc_manager.hpp:23,
                 from D:\a\supercollider\supercollider\editors\sc-ide\core\session_manager.cpp:21:
D:/a/supercollider/supercollider/editors/sc-ide/widgets/code_editor/editor.hpp:59:71: warning: 'constexpr QFlags<T>::QFlags(QFlags<T>::Zero) [with Enum = QTextDocument::FindFlag; QFlags<T>::Zero = int QFlags<QTextDocument::FindFlag>::Private::*]' is deprecated: Use default constructor instead [-Wdeprecated-declarations]
     bool find(const QRegExp& expr, QTextDocument::FindFlags options = 0);
                                                                       ^
In file included from D:/a/supercollider/Qt/5.15.2/mingw81_64/include/QtCore/qglobal.h:1304,
                 from D:/a/supercollider/Qt/5.15.2/mingw81_64/include/QtGui/qtguiglobal.h:43,
                 from D:/a/supercollider/Qt/5.15.2/mingw81_64/include/QtWidgets/qtwidgetsglobal.h:43,
                 from D:/a/supercollider/Qt/5.15.2/mingw81_64/include/QtWidgets/qplaintextedit.h:43,
                 from D:/a/supercollider/Qt/5.15.2/mingw81_64/include/QtWidgets/QPlainTextEdit:1,
                 from D:/a/supercollider/supercollider/editors/sc-ide/widgets/code_editor/editor.hpp:23,
                 from D:\a\supercollider\supercollider\editors\sc-ide\core\doc_manager.hpp:23,
                 from D:\a\supercollider\supercollider\editors\sc-ide\core\session_manager.cpp:21:
D:/a/supercollider/Qt/5.15.2/mingw81_64/include/QtCore/qflags.h:123:80: note: declared here
     QT_DEPRECATED_X("Use default constructor instead") Q_DECL_CONSTEXPR inline QFlags(Zero) noexcept : i(0) {}
```

Patch:

```patch
--- a/editors/sc-ide/widgets/code_editor/editor.hpp
+++ b/editors/sc-ide/widgets/code_editor/editor.hpp
@@ -56,10 +56,10 @@ public:
     bool showWhitespace();
     bool showLinenumber();
     bool find(const QRegExp& expr, QTextDocument::FindFlags options = QTextDocument::FindFlags());
-    bool replace(const QRegExp& expr, const QString& replacement, QTextDocument::FindFlags options = 0);
-    int findAll(const QRegExp& expr, QTextDocument::FindFlags options = 0);
-    int replaceAll(const QRegExp& expr, const QString& replacement, QTextDocument::FindFlags options = 0);
+    bool replace(const QRegExp& expr, const QString& replacement, QTextDocument::FindFlags options = QTextDocument::FindFlags());
+    int findAll(const QRegExp& expr, QTextDocument::FindFlags options = QTextDocument::FindFlags());
+    int replaceAll(const QRegExp& expr, const QString& replacement, QTextDocument::FindFlags options = QTextDocument::FindFlags());

     void showPosition(int charPosition, int selectionLength = 0);
     QString symbolUnderCursor();

```

dyfer:
Hi @capital-G 
Thanks for this!
I think there are multiple aspects here beyond boost. At the very least it all should be pointed out...

This PR raises the deployment target for macOS legacy build - I'm not sure it we absolutely need to keep it at 10.11, but putting it at 10.15 might cut some people off from the new version of SC. That might be the "price of progress" but it should be at least discussed.

My preference would be to split this into 3 PRs:
- boost update (1.80 or 1.83)
- CI updates
- std::filesystem, utf fixes, and possibly old OS deprecations

What do you think?

telephon:
What would be interesting if there is a formal and practical way to mark which commits require a new binary. A lot of improvements in sclang are just class library, and they are (often) fully backwards compatible.

dyfer:
> As mentioned in the top post, the `std::filesystem` bump is a PR on its own, see #6331 - I simply bulid this PR on top of it 

Ah I see, sorry I missed that.


> This PR therefore bumped the CI because `std::filesystem` requires C++17 and was only introduced via macOS Catalina, which sadly is the first 64bit only macOS :/ 
> 
> > This PR raises the deployment target for macOS legacy build 
>
> I agree, but I also think the recent major versions of SC rarely introduced a "killer feature" where everyone wants to upgrade to the most recent version immediately - and if there is such a feature it may be advised to upgrade your OS nonetheless. In the end those legacy platforms create much pressure and probably holds dev effort on newer, probably more relevant features such as ARM or WASM support, often back. People can still use older binaries/versions, and it allows to have better support for recent platforms. If something really serious occurs or we say we fix bugs on older versions we could also introduce cherry-pick workflow if there are people interested in maintaining older versions.

TL;DR I think at this point in time, setting 10.15 as the oldest platform that SC runs on is fine with me.

I am generally a fan of maintaining usability for old systems - as a matter of accessibility (people might be using older platform for economic reasons). However, I think that at this point we can jump to 10.15. We've supported 10.11 long enough :)

And yes, if there is a demand, we can figure out a separate branch for the truly legacy build, but I don't think this would be likely the case.

> 
> > What do you think?

> In general I think we should embrace C++17, so we could just make the clean cut here, but I am fine either way as long as compiling on macOS 14 works w/o trouble.

Do you mean - building against system boost, right? Or is there an issue building with the included boost on the newer platforms?

> 
> I think it mainly comes down if we want to go with #5924 or #6331 combined with this PR - I think I am biased here but also lack enough xp to have a proper opinion on it as this is my first C++ commit. 

I think I'd lean towards boost 1.83 upgrade.
However, looking at #5924, that should also include small CMAKE change and probably removal of the boost patching script. Would you be amenable to creating a standalone boost 1.83 upgrade PR?

Afterwards, we could run the rest of this PR (OR, if we want to be granular, have another update that does c++17 updates + CI, and then run the std::filesystem and related changes in a subsequent PR). I just think that having a standalone boost 1.83 upgrade PR would be great.

telephon:
> > What would be interesting if there is a formal and practical way to mark which commits require a new binary. A lot of improvements in sclang are just class library, and they are (often) fully backwards compatible.
> 
> @telephon what do you mean? an sclang binary? or plugins binary?

I meant changes to the class library that work across binaries (of either sclang or plugins, in fact). And of course, the corresponding tests and help.

dyfer:
> I meant changes to the class library that work across binaries (of either sclang or plugins, in fact). And of course, the corresponding tests and help.

Got it; it just isn't clear to me how this relates to this PR.

The discrepancy between sclang and class library did catch me a few times by surprise, but I don't think there's a way to systemically tell which ones are compatible and which ones aren't.

telephon:
> Got it; it just isn't clear to me how this relates to this PR.

Sorry, indeed, I had just read your comment and forgot about the topic. We can discuss that another time in a separate issue.

dyfer:
@capital-G would you consider making a separate boost 1.83-only PR separately as I asked about [earlier](https://github.com/supercollider/supercollider/pull/6334#issuecomment-2182064968), then coming back to the rest of this PR?

dyfer:
@capital-G are you planning to follow up with a boost-only PR?

Also, looking at your [question in the other PR](https://github.com/supercollider/supercollider/pull/6231#issuecomment-1986366993) - the reason to have boost as a copy instead of a submodule was that it needed to be patched in order to work in SC. ~~As of boost 1.80 patching seems to be unnecessary~~ There's less patching needed these days, but there's still some in order to compile under MinGW (see https://github.com/supercollider/supercollider/pull/5924)). While MinGW is not an officially supported toolchain, I believe some are still using it so let's keep having boost as a copy with a patch for now. https://github.com/supercollider/supercollider/blob/2188ef6b9746b979766436c372bff1c29f93c070/external_libraries/boost_sc_changes.patch#L15-L21

dyfer:
@capital-G I wanted to test something with a newer version of boost and I'm stuck at the update process - can you describe how you updated boost? did you use the release zip to do that (which file exactly?), then run the `extract_boost.sh` script? When I do that, the script seems to fail, removing one SC source file and then stopping with an error...

<details>

I got the following after extracting the contents of `boost-1.86.0-b2-nodocs.7z` to `external_libraries/boost`, then running the script:

```sh
supercollider % ./external_libraries/extract_boost.sh $PWD/external_libraries/boost/
INFO: tracking source dependencies of library filesystem due to presence of "  //                                    class path                                      //
  //                                                                                    //
  //------------------------------------------------------------------------------------//

  class path :
    public filesystem::path_detail::path_constants<
#ifdef BOOST_WINDOWS_API
      wchar_t, L'/', L'\\', L'.'
#else
      char, '/', '/', '.'
#endif
    >
  {" in file "boost/filesystem/path.hpp"
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/filesystem/path.hpp
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/filesystem/path.hpp
INFO: tracking source dependencies of library thread due to presence of "        BOOST_THREAD_DECL void set_tss_data(void const* key,detail::thread::cleanup_caller_t caller,detail::thread::cleanup_func_t func,void* tss_data,bool cleanup_existing);" in file "boost/thread/tss.hpp"
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/thread/tss.hpp
INFO: tracking source dependencies of library regex due to presence of "BOOST_REGEX_DECL void BOOST_REGEX_CALL reset_stack_guard_page();" in file "boost/regex/config.hpp"
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/regex/config.hpp
INFO: tracking source dependencies of library program_options due to presence of "    class BOOST_PROGRAM_OPTIONS_DECL option_description {" in file "boost/program_options/options_description.hpp"
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/program_options/options_description.hpp
CAUTION: don't know how to trace dependencies through macro: " <memory>" in file: boost/exception/exception.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ASIO_CUSTOM_HANDLER_TRACKING" in file: boost/asio/detail/handler_tracking.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_MATH_LOGGER_INCLUDE" in file: boost/math/tools/toms748_solve.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_CORE_BACKEND_HEADER(boost/atomic/detail/caps_)" in file: boost/atomic/detail/capabilities.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_CORE_ARCH_BACKEND_HEADER(boost/atomic/detail/caps_arch_)" in file: boost/atomic/detail/capabilities.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_WAIT_BACKEND_HEADER(boost/atomic/detail/wait_caps_)" in file: boost/atomic/detail/wait_capabilities.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_CORE_BACKEND_HEADER(boost/atomic/detail/fence_ops_)" in file: boost/atomic/detail/fence_operations.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_REGEX_H1" in file: boost/regex/pending/static_mutex.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_REGEX_H2" in file: boost/regex/pending/static_mutex.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_REGEX_H3" in file: boost/regex/pending/static_mutex.hpp
INFO: tracking source dependencies of library exception due to presence of "        int clone_current_exception_non_intrusive( clone_base const * & cloned );" in file "boost/exception/detail/clone_current_exception.hpp"
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/exception/detail/clone_current_exception.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_TYPE_INDEX_USER_TYPEINDEX" in file: boost/type_index.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_EXTRA_BACKEND_HEADER(boost/atomic/detail/extra_ops_)" in file: boost/atomic/detail/extra_operations.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_CORE_ARCH_BACKEND_HEADER(boost/atomic/detail/caps_arch_)" in file: boost/atomic/detail/caps_gcc_atomic.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_CORE_ARCH_BACKEND_HEADER(boost/atomic/detail/fence_arch_ops_)" in file: boost/atomic/detail/fence_arch_operations.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_FP_BACKEND_HEADER(boost/atomic/detail/fp_ops_)" in file: boost/atomic/detail/fp_operations.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_EXTRA_FP_BACKEND_HEADER(boost/atomic/detail/extra_fp_ops_)" in file: boost/atomic/detail/extra_fp_operations.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_CORE_BACKEND_HEADER(boost/atomic/detail/core_ops_)" in file: boost/atomic/detail/core_operations.hpp
INFO: tracking source dependencies of library atomic due to presence of "BOOST_ATOMIC_DECL void* short_lock(atomics::detail::uintptr_t h) BOOST_NOEXCEPT;" in file "boost/atomic/detail/lock_pool.hpp"
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/atomic/detail/lock_pool.hpp
CAUTION: dependent file Jamroot does not exist.
   Found while scanning file boost/atomic/detail/lock_pool.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_CORE_ARCH_BACKEND_HEADER(boost/atomic/detail/core_arch_ops_)" in file: boost/atomic/detail/core_arch_operations.hpp
CAUTION: don't know how to trace dependencies through macro: "BOOST_ATOMIC_DETAIL_WAIT_BACKEND_HEADER(boost/atomic/detail/wait_ops_)" in file: boost/atomic/detail/wait_operations.hpp
CAUTION: don't know how to trace dependencies through macro: " <memory>" in file: boost/exception/detail/shared_ptr.hpp
CAUTION: don't know how to trace dependencies through macro: "PP1" in file: boost/type_traits/detail/is_function_ptr_helper.hpp
CAUTION: don't know how to trace dependencies through macro: "PP1" in file: boost/type_traits/detail/is_function_ptr_helper.hpp
CAUTION: don't know how to trace dependencies through macro: "PP1" in file: boost/type_traits/detail/is_function_ptr_helper.hpp
CAUTION: don't know how to trace dependencies through macro: "PP1" in file: boost/type_traits/detail/is_function_ptr_tester.hpp
CAUTION: don't know how to trace dependencies through macro: "PP2" in file: boost/type_traits/detail/is_function_ptr_tester.hpp
CAUTION: don't know how to trace dependencies through macro: "PP3" in file: boost/type_traits/detail/is_function_ptr_tester.hpp
CAUTION: don't know how to trace dependencies through macro: "PPI" in file: boost/type_traits/detail/is_mem_fun_pointer_impl.hpp
CAUTION: don't know how to trace dependencies through macro: "PPI" in file: boost/type_traits/detail/is_mem_fun_pointer_impl.hpp
CAUTION: don't know how to trace dependencies through macro: "PPI" in file: boost/type_traits/detail/is_mem_fun_pointer_impl.hpp
CAUTION: don't know how to trace dependencies through macro: "PPI" in file: boost/type_traits/detail/is_mem_fun_pointer_tester.hpp
CAUTION: don't know how to trace dependencies through macro: "PPI" in file: boost/type_traits/detail/is_mem_fun_pointer_tester.hpp
CAUTION: don't know how to trace dependencies through macro: "PPI" in file: boost/type_traits/detail/is_mem_fun_pointer_tester.hpp
Copying (and overwriting) file: ../../common/Samp.hpp

ERROR  ERROR  ERROR  ERROR  ERROR  ERROR  ERROR  ERROR  ERROR  ERROR  ERROR

****************************** std::exception *****************************
boost::filesystem::copy_file: No such file or directory [system:2]: "/path/to/supercollider/external_libraries/boost/../../common/Samp.hpp", "external_libraries/boost/../../common/Samp.hpp"
***************************************************************************
```

</details>

dyfer:
@capital-G I'd like to walk back my previous request. After testing possible further boost upgrade, I now understand that updating boost also requires dealing with `boost::filesystem` issues. Sorry I wasn't clear on that earlier.

In that case I think this PR is on the right track, combining `boost::filesystem` changes and boost upgrade.

Here are my observations and requests:
- it seems that boost 1.83 and 1.86 have similar compatibility requirements - how about we upgrade to boost 1.86 right away to get the latest version?
- the patch for boost and mingw seems not needed anymore in newer mingw/gcc; in that case `README_BOOST.md` should be updated to remove references to it and the `boost_sc_changes.patch` file should be removed. In order for the CI to work, it's enough to [change the Windows MinGW runner to use the 2022 OS version](https://github.com/dyfer/supercollider/commit/9e80be1c937d19e029bb45ac1abe09d45afb1eb4)
- the structure of our CI workflow files has changed since this PR was created, so now the changes to the CI need to be re-done. I've made working CI changes [in my branch, based on this PR](https://github.com/dyfer/supercollider/commits/upgrade-boost-mod/)

I personally don't feel I can review the `std::filesystem` part of this PR well. @telephon how you feel about this particular change?

If we agree that the filesystem part is good, then I think this PR is the path forward, after the points above are addressed.

capital-G:
Thanks for your feedback @dyfer - I'll try to re-polish this PR soon and get it in merge-able state according to your feedback, so thanks for that!

As mentioned in the first post I settled for boost 1.83 b/c it is the version used by the Debian package manager for sid and trixie (see https://packages.debian.org/de/sid/libboost-dev ) - and if I am unsure which version to use I always go for the latest debian version. I think it shouldn't be too hard to bump this 1.86 if you think we should go for it, but in my experience 1.83 would make more "sense" b/c of this debian backing? If we decide to go for 1.86 we should definitely not include the 1.83 upgrade in our history in order to avoid an unnecessary massive commit.

Regarding the boost::filesystem deprecating/PR - although it is my first C++ PR I think it is still valid - 95% of the changes are replacing `boost::` with `std::` and change the imports - but would be great if someone with a windows machine could check out this branch and test some I/O operations.

Oh, and just to re-iterate - this PR/boost upgrade would break the sc-flucoma plugin, but there is no way around this because they used non-public methods as a hack. I think they are relying on SendReply API which isn't exposed through the plugin API, but they could just include a basic C++ OSC library to send OSC messages w/o SendReply like in https://github.com/capital-G/OSCMessenger/

dyfer:
> As mentioned in the first post I settled for boost 1.83 b/c it is the version used by the Debian package manager for sid and trixie

1.86 seens to have built even in on ubuntu 20.04 - I guess we should test whether it works :) but if it does, I'd vote for the newest version, since we probably don't want to update it too often.

> Regarding the boost::filesystem deprecating/PR - although it is my first C++ PR I think it is still valid

Sounds good, let's test this.


> Oh, and just to re-iterate - this PR/boost upgrade would break the sc-flucoma plugin, but there is no way around this 

I'm well aware! And I made efforts to keep everyone in the loop about that previously. Quite a bit of time has passed and since we don't have a solution by now (e.g. extending plugin API), then I think we should proceed with the update, since there is a number of issues that the boost update will fix.

dyfer:
A note on the boost version to upgrade to, once `std::filesystem` PR is merged:
Our wiki states that we should [bundle the latest boost available](https://github.com/supercollider/supercollider/wiki/Platform-Support#:~:text=As%20much%20as%20possible%2C%20SuperCollider%20should%20support%20building%20with%20the%20latest%20release%20of%20Boost%2C%20and%20the%20version%20packaged%20in%20the%20source%20tree%20should%20be%20updated%20reasonably%20soon%20after%20a%20release.), when possible. 
I'd suggest that we try to update to do that. If it still compiles for the oldest supported platforms (e.g. Ubuntu 22.04), I think having the latest version would be preferable.

capital-G:
The only thing I want to take into consideration here is that there are plugins such as flucoma, which also rely on boost. I think it would be in "debian building tradition" to prefer the common system version here just to avoid any friction.

But in the end I don't have enough XP to evaluate this properly, I just want to take this into consideration.
If you think we should go for the latest boost, then I'll rebase it with the newest boost version after #6331 has been merged.

dyfer:
> plugins such as flucoma, which also rely on boost.

It's _just_ flucoma :) And eventually I hope this dependency would be solved by properly addressing this in the plugin API.

> even ubuntu 24.04 is still sticking to the latest debian version

Do you know if debian/ubuntu SC packagers use bundled or system boost?

In any case, as long as it compiles under the last two LTS Ubuntu versions, and also can be linked against whichever versions of the system libraries are in there, I think the newest boost would be desirable in our source.

capital-G:
> Do you know if debian/ubuntu SC packagers use bundled or system boost?

I couldn't find it in a quick search, but the scsynth package is relying upon a [`libboost-filesystem1.74.0`](https://packages.debian.org/bookworm/libboost-filesystem1.74.0) package, so I think they are using this as a patch for the boost version of debian (1.74 for bookworm)?

Nonetheless, let's simply use the latest one :)

dyfer:
Superseded by https://github.com/supercollider/supercollider/pull/6477

PR #6335, opened 2024-05-31
Fix Dictionary:-trueAt to respond to all objects

mtmccrea:
## Purpose and Motivation

#6001 introduced a breaking change that answers `-trueAt` based on the receiver's `-booleanValue`. 

But `-booleanValue` is only defined for `Boolean`, `Nil`, and `SimpleNumber`, so any other object will fail.

## Steps to reproduce (actual behavior)

```supercollider

d = (num: 23.7, yes:true, no:false, symb:\bar);
d.trueAt(\num) // true
d.trueAt(\yes) // true
d.trueAt(\no) // false
d.trueAt(\foo) // false (d[\foo] is nil)
d.trueAt(\symb) // breaks

```

## Proposed behavior

The proposed solution is that `Dictionary:-trueAt` answers `false` for any object that isn't `true` or a positive `SimpleNumber` (this aligns with the initial proposal in #6001).
```supercollider
trueAt { arg key;
	^try { this.at(key).booleanValue } { false }
}
```

This lets individual kinds of objects change behavior in the future by implementing `-booleanValue`, rather than implementing `Object:-booleanValue`, which could be trickier considering a default behavior for all kinds of objects.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

Also,  `Object:-falseAt` previously took a `key` argument and passed it to `trueAt`, which has no arguments defined. This updated method has identical behavior to the previous, but is more explicit in the implementation (the proposal to change behavior is a separate issue)

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Dictionary.schelp
+++ b/HelpSource/Classes/Dictionary.schelp
@@ -193,24 +193,31 @@ d.matchAt(\abc)
 ::
 
 method::trueAt
-Returns link::Classes/True:: if the item's booleanValue at the key is true, otherwise false. This method is also valid in link::Classes/Object::.
-code::
-d = (x:true, y:false);
-d.trueAt(\x) // true
-d.trueAt(\y) // false
-d.trueAt(\foo) // false
+Returns link::Classes/True:: if the item's teletype::booleanValue:: at the key is teletype::true::, otherwise teletype::false::.
 
+code::
 // using binary value equivalents
 d = (x:1, y:0);
-d.trueAt(\x) // true
-d.trueAt(\y) // false
+d.trueAt(\x)   // true
+d.trueAt(\y)   // false
 d.trueAt(\foo) // false
+
+// other kinds of objects
+d = (num: 23.7, yes:true, no:false, sym:\foo);
+d.trueAt(\num) // true
+d.trueAt(\yes) // true
+d.trueAt(\no)  // false
+d.trueAt(\sym) // false
+d.trueAt(\bar) // false (d[\bar] is nil)
 ::
 
 method::falseAt
-Returns link::Classes/False:: if the item's booleanValue at the key is not true, otherwise true. This method is inherited from link::Classes/Object::.
+Returns link::Classes/True:: if the item's teletype::booleanValue:: at the key is teletype::false::, otherwise teletype::true::.
+See link::#-trueAt:: for examples.
+
 
 subsection::Testing
+
 method::includes
 Returns true if the specified item is stored in the Dictionary as a value. Element is checked for equality (not for identity). For identity matching see subclasses: link::Classes/IdentityDictionary:: or link::Classes/Event::.
 code::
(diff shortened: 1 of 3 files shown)


SCClassLibrary/Common/Collections/Dictionary.sc
```diff
@@ -23,7 +23,10 @@ Dictionary : Set {
 		^nil
 	}
 	trueAt { arg key;
-		^this.at(key).booleanValue
+		^try { this.at(key).booleanValue } { false }
```
telephon, line 26: better, in this case:
```
var val  = this.at(key); 
if(val.respondsTo(\booleanValue).not) { ^false };
^val.booleanValue
```

HelpSource/Classes/Dictionary.schelp
```diff
@@ -193,24 +193,31 @@ d.matchAt(\abc)
...
 // using binary value equivalents
 d = (x:1, y:0);
-d.trueAt(\x) // true
-d.trueAt(\y) // false
+d.trueAt(\x)   // true
+d.trueAt(\y)   // false
 d.trueAt(\foo) // false
+
+// other kinds of objects
+d = (num: 23.7, yes:true, no:false, sym:\foo);
```
telephon, line 206: 23.7 is not really the intended use. It should be said that this will be true, but `1` and `0` are the interface for which it is supposed to work. Could you change 23.7 to 1 and add an extra comment that any number will do? Thanks!

telephon (changes requested):

HelpSource/Classes/Dictionary.schelp
mtmccrea: The example directly preceding this illustrates the binary integer use case. I moved that example first, expecting that it's the intended case.

Given that a dictionary can hold anything, this second set of examples is meant to show how it will respond to a range of objects (i also included a Symbol). 

Should I still modify it?

SCClassLibrary/Common/Collections/Dictionary.sc
mtmccrea: Yes, that's better, updated!

HelpSource/Classes/Dictionary.schelp
telephon: no, it's actually ok. I just have the uneasy feeling that we promote non-standard use if it is not explictly mentioned.

mtmccrea: ah ok, I see also that the diff obscured the example reordering, so that may have been misleading

mtmccrea:
The Linux test failed for an unrelated reason, I've re-run it now all tests are passing.

telephon (approved):

capital-G (approved):
The changes make sense to me. The question is though - why does this method exist in the first place - but maybe this is out of scope for this PR :)

Feel free to merge @mtmccrea

mtmccrea:
> why does this method exist in the first place - but maybe this is out of scope for this PR :) 

Indeed ;) this came up in another related thread but we decided to stick with just fixing the bug. Deprecations and pruning of the class lib would likely be a long conversation and complex project to build consensus on let alone carry out.

Thanks!

telephon:
yes, It was a bad idea of mine years ago, just for having a transparent way for deal with SynthDef presets.

PR #6337, opened 2024-06-05
Remove gedit plugin sced

capital-G:
## Motivation

The vim and emacs plugins are linked via git-submodules in the `editors` folder, but there is also a gedit plugin (https://github.com/supercollider/supercollider/tree/develop/editors/sced) which is, like the sc-ide, actually a part of this repo.
Yet it had no real changes since 2012 (last actual change was https://github.com/supercollider/supercollider/commit/516574a405bb7b3ab1a1d2af4170ddd1b925e795).

IIUC the plugin is also not working since 2014/2015.

<details>
<summary>More info</summary>

The plugin could only be loaded in Ubuntu 22.04 after I changed the plugin type from `python` to `python3`.
Upon starting the plugin, an exception is thrown, complaining about

https://github.com/supercollider/supercollider/blob/85bf9b078ddf78342a7e3fde2a745be8f3cb0c77/editors/sced/sced3/supercollider.py#L345

saying that `Window` object has no attribute `get_ui_manager` - if you google this error you will find issues concerning an API break of gedit which happend around 2014/2015.
So I think in general it is fair to say that this does not work in its current state and has not been used by people because otherwise it would have been fixed in the meantime.

</details>

If people decide that a gedit plugin should be a thing (again) and want to build upon the existing code base, it is possible to transfer its old history to a new repo via `git-filter-branch`, even if we decide to delete the files now.

IMO everything in the main repo should be somehow supported, maintained and somehow working, but it seems like there is no interest for many years in gedit maintenance, so I think its time to go.
The unclear state could potentially create future maintenance which is not necessary.

Release Notes: remove gedit plugin

* Delete `editors/sced` and all its references

## Types of changes

- Breaking change

## To-do list

There is a mention concerning sced in the windows build

https://github.com/supercollider/supercollider/blob/85bf9b078ddf78342a7e3fde2a745be8f3cb0c77/platform/windows/CMakeLists.txt#L15-L17

I don't know anything about `NSIS` files, does the deletion make anything easier in the future? I would just keep it like this.

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


smoge:
I think it makes sense, but keep the code in a repository of the same project. Even if with a note or a "archive" label. 

FWIW, there is another plugin for Kate (KDE text editor) written by Jakob Leben when he was implementing QtCollider. I don't know if it still works, maybe it does.

telephon (approved):

dyfer:
I understand that we'll be able to recover this later, but maybe there should be a trace left in that directory? 
I vaguely remember that some distros offered the gedit plugin as a separate package, thought I can't remember when was the last time I saw it working.
I agree that what we have in the codebase should be maintained and working though.

dyfer:
I wonder if a more transparent change would be to extract that directory into a separate repo and indicate that it's not supported, or even archive it. But is it worth the hassle?

capital-G:
> I wonder if a more transparent change would be to extract that directory into a separate repo and indicate that it's not supported, or even archive it. But is it worth the hassle?

IMO not - last change is 12 years ago, it seems broken since 10 years ago.

> EDIT: also references to SC_ED should probably be removed from CMakeLists?

I think I deleted them all.

dyfer:
All right, let's remove this - we can bring it back (maybe as a submodule) in case anybody actually uses it somehow.

PR #6344, opened 2024-06-11
add `FULL_CHECK` flag for clang-format

capital-G:
## Purpose and Motivation

Currently pre-commit performs a full scan of the C++ files on every commit that touches them - this is necessary because we need to verify the state of the full repo in the CI and not just of the latest commit so it is not possible to "sneak in" unformatted code when pushing multiple commits.

We can assume though that for local setups it is sufficient to check only the diff staged for the current commit.
This drastically speeds up local pre-commit times (from 50 seconds down to 2 seconds).
On CI we still perform a full check of the repo. As it is not possible to pass the `--all-files` flag of the pre-commit invocation to the actual script, we use an environment variable.

Also see https://github.com/supercollider/supercollider/pull/6319#issuecomment-2159966408 and reply/below

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

=> Will update wiki with a note on how to perform a full scan when this gets merged

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -32,6 +32,8 @@ jobs:
           python-version: '3.10'
       - uses: pre-commit/action@v3.0.0
         name: 'Run pre-commit'
+        env:
+            FULL_CHECK: 1
       - name: set version string for artifacts
         id: set-version
         run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     hooks:
     -   id: clang_format
         name: Format C++ code
-        entry: python tools/clang-format.py formatall
+        entry: python tools/clang-format.py format
         pass_filenames: false
         types_or:
           - "c++"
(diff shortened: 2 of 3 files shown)


mtmccrea:
Thanks for this quick fix @capital-G! 
I think the solution is suitable, given that your [second proposed solution](https://github.com/supercollider/supercollider/pull/6319#issuecomment-2160237445) would rely fully on pre-commit to handle formatting introducing a hard developer dependency. This could still be considered in the future if it turns out that all contributors just end up using pre-commit anyway (seems likely).

I've tested this PR and the speedup is huge, making pre-commit very usable 🚀

I'll defer to your expertise on specifics of the implementation, but it looks clean and well commented.

mtmccrea (approved):

capital-G:
Updated the wiki on how to run a check across all files https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit#run-pre-commit-on-all-files

Will merge now :)

PR #6345, opened 2024-06-12
SClang: typed argument checker for primitives

JordanHendersonMusic:
## Purpose and Motivation

Managing the stack pointer is annoying. 

Making sure primitives check the types and produce good consistent errors is annoying.

Ensuring you don't overwrite a value on the stack while you are writing to it is annoying.

Here is a proposal for a solution I came up with — would love opinions!

```c++
int typedArgumentTest(struct VMGlobals* g, int numArgsPushed) {
    auto maybeArgs =
        TypedArgsFromStack<
                           FloatChecker,
                           IntChecker,
                           SymbolChecker,
                           OrNil<ArrayChecker>,
                           OrNil<FloatChecker>,
                           OrChecker<FloatChecker, IntChecker>
       >::get(g, "typedArgumentTest");

    if (!maybeArgs)
        return errWrongType;

 
    auto [arg0, arg1, arg2, arg3, arg4, arg5] = *maybeArgs;
    ...
}
```

See how IDEs can even insert type hints and giving you proper lsp support.

They can then be used as expected.
```c++
    post("arg0: %f\n", arg0); // this is a double
    post("arg1: %d\n", arg1); // this is a int
    post("arg2: %s\n", arg2->name); // this is a PyrSymbol*

    if (arg3)   // an optional<PyrObject*>
        post("arg3: is an array\n");
    else
        post("arg3: is nil\n");

    if (arg4)
        post("arg4: %d\n", *arg4);  
    else
        post("arg4: is nil\n");

    // the OrChecker produces a variant.
   arg5.visit(Overloaded {
        [](FloatChecker, double f) { post("arg5 is a float: %f\n", f); },
        [](IntChecker, int i) { post("arg5 is a int: %d\n", i); },
        [](ArrayChecker, PyrObject* a) { post("arg5 is an array with size %d\n", a->size); },
    });

   if (arg5.holds<FloatChecker>())
        post("arg5: got a float: %f\n", arg5.force<FloatChecker>());
    else if (arg5.holds<IntChecker>())
        post("arg5: got a int %d\n", arg5.force<IntChecker>());
    else if (arg5.holds<ArrayChecker>())
        post("arg5: got an array with size %d\n", arg5.force<ArrayChecker>()->size);
    else
        post("arg5: falied\n");


    if (auto f = arg5.tryHolds<FloatChecker>(); f)
        post("arg5: got a float: %f\n", *f);
    else if (auto i = arg5.tryHolds<IntChecker>(); i)
        post("arg5: got an int: %d\n", *i);
    else if (auto a = arg5.tryHolds<ArrayChecker>(); a)
        post("arg5: got an array of size: %d\n", (*a)->size);
    else
        post("arg5: failed\n");


```

If the user makes a mistake only the arguments that are invalid have error messages printed.

Here is what happens when you call the above primitive with only a receiver.
```
Wrong argument type in primitive call to typedArgumentTest.
	Argument 1 expected a Int but received a Nil with value nil.
	Argument 2 expected a Symbol but received a Nil with value nil.
	Argument 5 expected a any of( Float Int ) but received a Nil with value nil.
```

Here, argument 3 and 4 are not printed because nil is a valid state for them.

This commit also contains some demo code (that would obviously be removed). 

If any one wants to try it out, you can do...
```
something.typedTest
```

Turns out the old mac doesn't support `std::visit` so the demo code I've written won't work. It might be worth writing a custom variant for this. Since we know what types it will have, its not too hard (also, I like this kind of code).

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Core/Object.sc
+++ b/SCClassLibrary/Common/Core/Object.sc
@@ -954,4 +954,7 @@ Object  {
 	help {
 		this.class.asString.help
 	}
+	typedTest {|arg1, arg2, arg3, arg4, arg5|
+	    _TypedArgumentTest
+    }
 }
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -51,6 +51,7 @@
 #include "SC_Version.hpp"
 #include <map>
 
+#include "TypedArguments.hpp"
 #ifdef _WIN32
 #    include <direct.h>
 #else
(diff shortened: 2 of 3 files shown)


smoge:
As you said, it's a great idea and will improve error handling. I'm just curious about the custom variant class template, I imagine std::variant and std::visit must be optimized for storing and accessing types like this. The only way to be sure is testing and benchmarking seriously, right?

smoge:
https://github.com/martinmoene/variant-lite

JordanHendersonMusic:
I've made a variant. If you used the c++ one, you'd have to make type wrappers which would be more complicated. This has a much easier interface.

JordanHendersonMusic:
I'm closing this because I don't have time to work on it and I don't think it is that useful.

PR #6323, opened 2024-05-24
help: document the usage of addSystemSynthDefs

telephon:
## Purpose and Motivation

Documentation for addSystemSynthDefs

Issue #6340, opened 2024-06-06
Tooling, testing and CI enhancement proposal


silvanocerza:
Hello hello everyone! After the discussion that I had with @capital-G on #6319 regarding pre-commit hooks we had a really nice chat together about the project CI and other related stuff.

I decided I'd make a proposal on how to enhance it so here it is. This is really high level, if there needs be I can outline in more details what I want and can do.

## Tooling & Linting

The project is using a pretty old version of `clang-format`, it should be updated.
@capital-G told there's been some friction on this cause of huge formatting changes that caused some merge conflicts.

I think we can work around that by making the change gradual and having formatting checks only on files that are actually modified in future PRs.

I'd advocate to remove the version requirement set in `tools/clang-format.py` and deprecate it right after, as it makes it hard to use the actual `clang-format`. 

Change CI and pre-commit hooks so they check only files that have actually been modified in the branch or PR that is.

This will enable us to update the rules in `.clang-format` to support the latest ones and gradually enhance the project formatting without causing to much conflicts.

## Testing

I see that testing is done using [qpm](https://github.com/supercollider/qpm), a pretty outdated internal library that still runs on Python 2.7.

We should find a solution to that, either find a new way to run tests or rewrite the test runner for Python 3.

@capital-G also mentioned to me that some tests are flaky, we would need to find a way to mark those tests and rerun them some time before considered them completely failed.
## CI

The CI needs some improvement. 

Main problem I see right now is that there's a single `actions.yml` workflow that is quite convoluted and complex. 

First thing I would do is split that into multiple reusable workflows that are called by the existing one. This would already give us some breathing room.

This way we can rework the workflows so that we don't need to wait for all the builds to end to run tests. I'd probably go and split the building and testing process into 3 workflows, one for each platform. I'll figure it out as I go probably. In any case I won't make any change in the logic for the time being.

Splitting the `actions.yml` workflow we can also enhance the release process obviously.

Having said that I would like to hear any concerns from the usual contributors and go from there.
If there are no issues with the approach I propose I can take this into my hand and work on it while documenting it obviously.

silvanocerza:
> v14.0 is the latest stable version distributed via debian and we recently updated from 8 to 14 - I think we can stay on 14 for a few years? The main goal of the formatter should be that it does not produce problems.
> 
> Because formatting rules differ between clang-format versions we need to stick to one. I think by switching to pre-commit we solved this in a nice manner.

[...]

PR #6346, opened 2024-06-13
Use single blank line as file ending in sc files

capital-G:
## Purpose and Motivation

Inspired by https://github.com/supercollider/supercollider/pull/6323#issuecomment-2165076689

*Controversial* proposition which enforces a single blank line as file ending (`\n` aka LF) on `.sc`, `.schelp` and `.scd`.
As this touches many files (and therefore git history), this is open for discussion - if we want to go this way, the code is here. But, other than clang-formatting, it does not taint `git blame`.

The formatting rules are checked via pre-commit, so if people have set it up, their files will be adjusted automatically.

Most interesting files are
* https://github.com/supercollider/supercollider/blob/d3d999270a8be8ec47c05ba8ed49ebc3282a8784/tools/remove_trailing_blank_lines.py
* https://github.com/supercollider/supercollider/blob/d3d999270a8be8ec47c05ba8ed49ebc3282a8784/.pre-commit-config.yaml

## Types of changes

- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation


mtmccrea:
I'm actually surprised how _few_ files this touches. My immediate hesitation was about the classlib, but I see only _four_ are modified there.

The majority seem to be _.schelp_, examples, and some testsuite files, which are somewhat "low stakes" in terms of revision history, IMO. So it could be worth it if we have the benefit of acquiring this pre-commit rule going forward? It's indeed a cumbersome thing to check for and ask of PR authors.

(Aside: wishing I would have done this in the recent style reformat of the help docs so many of these touched files would have been grouped into that commit :-/)

JordanHendersonMusic:
Perhaps this would be a good commit to test how blame might ignore certain commits?

capital-G:
> Perhaps this would be a good commit to test how blame might ignore certain commits?

Thanks for bringing this up. Bundled this comment with the discussion https://github.com/supercollider/supercollider/issues/6340 this resulted in https://github.com/supercollider/supercollider/pull/6350 which I think is pretty neat :)

mtmccrea:
Tested and working as expected using git CLI. Nice!

Though I use GitHub Desktop, and there's one snag: when committing a file that is subsequently modified by the hook to remove EOF whitespace, GH Desktop doesn't see that file as being "touched" so the diff view still shows the EOF whitespace. Switching to XCode, I see the whitespace removed, switching back to GH Desktop the updated diff shows properly. 
Seems like a GHD issue... (quick web search didn't provide any immediate answers) Or could it be that the files are modified in such a way that they aren't broadcasting as being "touched"? Could be related to the fact the changes are whitespace only?

capital-G:
Rebased against develop to resolve merge confilcts.

Would like to get a definitive yay/nay on it.

mtmccrea:
I'm in favor of this. If merged it would be good to add to `.git-blame-ignore-revs` and add a note in docs about this behavior (maybe in formatting instructions and/or docs about git/CI behavior)

capital-G:
> I'm in favor of this. If merged it would be good to add to `.git-blame-ignore-revs` and add a note in docs about this behavior (maybe in formatting instructions and/or docs about git/hook behavior)

We could add this commit to `.git-blame-ignore-revs', but this would only be helpful during git blame, not while traversing the git tree - since no actual code is touched here, and git blame on a file ending is not too interesting, I decided not to include it in the file, but I can add it if you think it should be.

mtmccrea:
>git blame on a file ending is not too interesting

Good point. So it is really just a judgement call, I could see adding it just for the sake of if thematically being a large whitespace modification, but I don't feel strongly about it either way :)

capital-G:
Rebased and added the offending commit to `.git-blame-ignore-revs`

JordanHendersonMusic (approved):
Looks good, I don't think its controversial because it is just the end of the line. 

One note to merger, this should **not** be squashed as the middle commit has been added to .git-blame-ignore-revs.

dyfer:
There seems to be a consensus that we want this - merging. Thanks!

PR #6348, opened 2024-06-13
Fix doc icon position

capital-G:
## Purpose and Motivation

Current state:
grafik

Fixed state:
grafik

Fix was made necessary by https://github.com/supercollider/supercollider/commit/39702ec3a9cdc9a4e2ee7385db747e9ba135c3d1

Supersedes https://github.com/supercollider/supercollider/pull/6275

Thanks @prko for bringing this up and providing the impulse.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/scdoc.css
+++ b/HelpSource/scdoc.css
@@ -137,6 +137,7 @@ a.subclass_toggle:hover {
     background-repeat: no-repeat;
     background-position: 0.5em center;
     background-size: 1.5em;
+    display: inline-block;
 }
 
 a.menu-link {
@@ -150,6 +151,10 @@ a.menu-link {
 a.menu-link:hover {
     text-decoration: none;
     color: var(--color-fg-200);
+}
+
+li.menuitem a.menu-link.home {
+    padding-left: 2.5em;
 }
 
 #menubar .menuitem .home a {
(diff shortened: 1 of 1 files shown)


mtmccrea:
On my system (macOS) it seems to cause the other menu items to drop down to the bottom of the menu:
develop:
Screenshot 2024-06-18 at 14 04 01
this fix:
Screenshot 2024-06-18 at 14 05 36

To see this, you have to collapse the browser to a point where "SuperCollider" is hidden.

capital-G:
thanks for the feedback, will take a look at it

capital-G:
I believe this is fixed - can you verify @mtmccrea ?

capital-G:
> Although I see also now there is a position when shrinking the browser you can get it into a state where "indexes" wraps down out of the menu:

I am aware of this but this has been an issue for a longer time, see the current latest stable docs
image

The DOM of the docs should be rewritten at some point nonetheless as the code base is not in a good shape. We could then switch to flex containers which would allow us to reduce our CSS code base while also getting better results - but this would be indeed another PR.

mtmccrea (approved):

Issue #5169, opened 2020-09-20
General support ticket for Qt 6


mossheim:
This is a ticket for general discussion on Qt 6.

Qt 6 will require c++17, which we are tracking on another ticket.

-----
Edited by @dyfer to help with keeping track of the progress; last update 2024-11-04.
Original TODO from @mossheim 
- [x] fix register in SCDoc/lex.scdoc.cpp #5396
- [x] upgrade to C++17 #5396
- [x] fix Qt deprecation warnings, ~~keeping compatibility with Qt 5.9 if possible~~ #6464
- [x] determine whether to deprecate QuartzComposer code - deprecated in #5710
- [x] migrate any remaining code incompatible with Qt6 #6475
- [x] determine what code relies on Core5Compat and migrate as much as possible #6475 (except regex code)
- [x] port search/replace regex code https://github.com/supercollider/supercollider/pull/6512
- [x] remove Core5Compat from cmake and build system (after all code relying on it is migrated) ##6524

Use the ~~[Qt6 project](https://github.com/supercollider/supercollider/projects/21)~~ [Qt6 milestone](https://github.com/supercollider/supercollider/milestone/26) to keep track of open PRs.

brianlheim:
here is a branch which at least compiles with the Qt 6.0.0-beta2 release: https://github.com/brianlheim/supercollider/tree/topic/qt6

some notes i have:
- Qt6 provides a Core5Compat module which saves us from having to migrate too many things at once
- seems like the MacExtras module is gone?
- many compile errors / commented out code from my branch would likely be fixed simply by addressing deprecation warnings under Qt 5.15. that should be the first priority.
- it doesn't seem like QuartzComposer/cocoa stuff is present in qt6, at least i couldn't find it. if it's truly gone, i think we should plan to deprecate it in 3.12 to give people a heads up
- not sure the SafePtr replacement is correct but worth looking at in more detail. no reason for us to reimplement that now.
- since this requires C++17, the `register`s in SCDoc/lex.scdoc.cpp have to go. there should probably be an option in lex to not generate them, i just removed them without much concern for that in order to compile.
- need to find what the correct replacement for the cmake command qt6_add_translation is
- the only thing which seems to require some real thought is moving from QRegExp to QRegularExpression in autocompleter.cpp, since we must interop with a Qt object that requires it
- supporting both qt5 and qt6 is probably doable for awhile

overall this is a *lot* less painful than i was expecting, kudos to Qt for that!

[...]

PR #6351, opened 2024-06-13
Add VS Code IDE settings

capital-G:
This top post will be kept up to date to reflect the current state of the PR.

## Purpose and Motivation

Although embracing a [semi-open](https://vscodium.com/) editor can lead to unwanted dependencies and reduce editor-diversity, VS Code is a free (as in free beer) and adds a really low barrier to start contributing to SuperCollider across multiple platforms by providing excellent auto-complete support and debugging capabilities for the C++ codebase.

This PR does not force anybody to use VS Code, but for people who are accustomed to VSCode, this PR provides the necessary config files so it is possible to start coding, building and debugging without any additional setup or CLI maneuvering.

For now, I've ~~only~~ added the MacOS and Linux settings, but it's possible to add Windows settings as well, but since I don't currently have a Windows machine someone else has to provide the necessary config files.
I think it is beneficial to share such a config among developers, as it took me at least some time to get everything set up correctly.

~~I also initiated a CMakePresets.json to make it easier to configure the build from within an IDE - no CLI needed anymore.  I think it would be interesting to include some default configurations here, such as Debug, Release, Headless ... - see https://martin-fieber.de/blog/cmake-presets/~~ => Since we switched to Qt6 the configuration step can run w/o additional information.

By using [vscode-cmake-tools](https://vector-of-bool.github.io/docs/vscode-cmake-tools/debugging.html) it is not necessary anymore to explicitly provide a launch config for every target.

Targets available for debugger
grafik

Debugging sclang

grafik

~~Once the current state of the wiki is cleared (see https://github.com/supercollider/rfcs/pull/21)~~ I can write a small tutorial how to get things running, but it should be as easy as

* Open git repo with vscode - accept to install suggested extensions
* Click on run-debug (via lldb/gdb) which will trigger a build of `all`. Afterwards select the binary you want to debug (e.g. sclang, scsynth, ...)

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

Merge note: Should be squashed


JordanHendersonMusic:
Its a shame vscode doesn't have better cmake support, CLion finds all of these targets (and more) automatically. 
Anyway, I think this is a good idea. 
What happens if someone already has a `.vscode` directory?

capital-G:
> Its a shame vscode doesn't have better cmake support, CLion finds all of these targets (and more) automatically. Anyway, I think this is a good idea. What happens if someone already has a `.vscode` directory?

I think this would result in a merge conflict.
Actually I found the cmake support quite decent - I still have a CLion license but actually don't use it anymore.

JordanHendersonMusic:
> I think this would result in a merge conflict.

I am unsure about this. If my ide setup was invalidated by updating I'd be a bit annoyed. 

On the other hand, the setup of all the targets is really helpful. 

Is there a middle ground here?

capital-G:
> Is there a middle ground here?

Thanks for your input. I took another look at this and found a better way by excluding `settings.json` from tracking via git. This allows to have local user preferences, but things such as debug targets, recommended plugins (which then automatically pick up settings such as proper tab/spaces config etc.) and CPP adjustments can be shared, so everyone benefits from changes on those things but not beyond.

I rebased and modified the files accordingly.

smoge:
> > Its a shame vscode doesn't have better cmake support, CLion finds all of these targets (and more) automatically. Anyway, I think this is a good idea. What happens if someone already has a `.vscode` directory?
> 
> I think this would result in a merge conflict. Actually I found the cmake support quite decent - I still have a CLion license but actually don't use it anymore.

I think vscode can " export settings" , maybe it would be distributed using this way?

capital-G:
@JordanHendersonMusic bump :)

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,32 @@
...
+                "${workspaceFolder}/SCDoc/**",
+                "${workspaceFolder}/server/**"
+            ],
+            "customConfigurationVariables": {
+                "-DCMAKE_PREFIX_PATH": "/usr/local/opt/qt@5",
+                "-DCMAKE_BUILD_TYPE": "Debug"
+            },
+            "defines": [],
+            "macFrameworkPath": [
+                "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks"
```
JordanHendersonMusic, line 22: Is it possible to install Xcode elsewhere? Perhaps if you had different versions? I'm not really a mac user.

.vscode/launch.json
```diff
@@ -0,0 +1,134 @@
+{
+    // see https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/debug-launch.md
+    "version": "0.2.0",
+    "configurations": [
+        // sclang target
+        {
+            "type": "lldb",
+            "request": "attach",
+            "name": "macOS: sclang attach (lldb)",
```
JordanHendersonMusic, line 9: Anyway to make this cross platform? It seems just the name is mac specific.

CMakePresets.json
```diff
@@ -0,0 +1,28 @@
+{
```
JordanHendersonMusic, line 1: This file is one generated by CMake and not a part of vscode's settings. I'm just concerned that other IDE will pick this up and attempt to use it, particularly if they aren't on a mac.

.gitignore
```diff
@@ -37,4 +37,4 @@ __pycache__
 .*.sw?
 
 # vs code files
-.vscode
+.vscode/settings.json
```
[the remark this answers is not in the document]
JordanHendersonMusic: Also, this is gitignore.io's recommendation:
```
# Created by https://www.toptal.com/developers/gitignore/api/visualstudiocode
# Edit at https://www.toptal.com/developers/gitignore?templates=visualstudiocode

### VisualStudioCode ###
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
!.vscode/*.code-snippets

# Local History for Visual Studio Code
.history/

# Built Visual Studio Code Extensions
*.vsix

### VisualStudioCode Patch ###
# Ignore all local history of files
.history
.ionide

# End of https://www.toptal.com/developers/gitignore/api/visualstudiocode
```

CMakePresets.json
capital-G: One reason why I am so eager about this is that such a file allows us to make the building in CI much easier and also provide "suggested" configurations for e.g. headless build which are currently stored in a README - and it would also help me in setting up the wasm build chain.

An IDE may pick this file up, but as long as one doesn't select any preset, Cmake will not start configuring. And even if one applies this by incident, it is always possible to manually configure via the CLI. In general this allows to ditch the CLI completely for 95% of the dev work.

The idea is to support more configurations over time.

.vscode/launch.json
capital-G: macOS uses lldb and I think linux mostly relies on gcc/gdb(?) and then there is windows which is completely different - I think this could develop over time with contributions from people with different platforms, and maybe there is a better/more generic way of doing things, but I am not aware of one currently. this is one intention of the PR, to benefit of knowledge of other devs which will provide an improving setup for everyone over time.

.vscode/c_cpp_properties.json
capital-G: This was set via vscode itself and I think most devs have only one XCode version on their Mac as it is managed via the AppStore. I think it can be considered a sensible default, but maybe there is a better way of doing it, but I think the advanced devs with multiple XCode versions can figure out to modify this file in case they use vscode?

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,32 @@
+{
+    "configurations": [
+        {
+            "name": "Mac",
```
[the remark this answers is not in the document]
capital-G: Yes - I can currently only provide one for mac, but maybe other people can contribute one for their platform.

.gitignore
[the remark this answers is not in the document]
capital-G: So something along

```gitignore
# vs code files
.vscode/*
!.vscode/tasks.json
!.vscode/extensions.json
!.vscode/launch.json
!.vscode/c_cpp_properties.json

```

CMakePresets.json
JordanHendersonMusic: Ah, actually, this is really cool, changes the way people build, and something I didn't know about!
CLion picks these up and gives them as options.

Some thought...

Could this be its own PR, seems like vscode is actually secondary?

Could we use conditions to disable things...
```
{
    "inherits": "macOS base",
    "name": "macOS_supernova",
    "displayName": "macOS with supernova",
    "cacheVariables": {
        "SUPERNOVA": "ON"
    },
    "condition": {
        "type": "equals",
        "lhs": "${hostSystemName}",
        "rhs": "Darwin"
    }
},
```

Why is this disabling supernova? Having a config for every combination of settings will be too much. Perhaps just two per OS, recommended and testing? 

Should we also use build presets?

Can we use test presets with the sclang unit tests?

.gitignore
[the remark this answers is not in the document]
JordanHendersonMusic: Seems good to me.

CMakePresets.json
JordanHendersonMusic: Actually, scrap that... I think this should be its own PR, and would be a mistake to hide this quite important useful feature in a PR about IDE settings. I'm also happy to contribute and add linux stuff — I know nothing about windows.

scztt:
I would suggest renaming the files in `.vscode` to something like `launch.json.template` or `launch.json.example`. While this means one extra step to start working in VSCode (and east step: rename them), having these files committed in git makes it basically impossible to have your OWN workspace or launch settings, something that makes it somewhat unusable for a local development environment.

.vscode/launch.json
```diff
@@ -0,0 +1,134 @@
...
+            "program": "${workspaceFolder}/build/lang/Debug/sclang",
```
scztt, line 10: This should be `"program": "${command:cmake.launchTargetPath}",`?

.vscode/launch.json
```diff
@@ -0,0 +1,134 @@
...
+            "waitFor": true,
+        },
+        {
+            "name": "macOS: sclang launch (lldb)",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/build/lang/Debug/sclang",
+            "args": [],
+            "stopAtEntry": true,
+            "cwd": "${fileDirname}",
```
scztt, line 20: I see some benefits but also some hazards to having the cwd be based on the current file - this could create differences in behavior depending on which filed you're viewing, which could be incredibly confusing esp for someone that doesn't know about vscode launch settings. Maybe a more stable option is `${command:cmake.getLaunchTargetDirectory}`?

.vscode/launch.json
```diff
@@ -0,0 +1,134 @@
...
+        // sclang target
+        {
+            "type": "lldb",
+            "request": "attach",
+            "name": "macOS: sclang attach (lldb)",
+            "program": "${workspaceFolder}/build/lang/Debug/sclang",
+            "waitFor": true,
+        },
+        {
+            "name": "macOS: sclang launch (lldb)",
```
scztt, line 14: Is this sclang binary actually launchable from this location? I seem to remember that the class library and some other things mean that it doesn't really work?

.vscode/launch.json
```diff
@@ -0,0 +1,134 @@
+{
+    // see https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/debug-launch.md
```
scztt, line 2: I think we could streamline these and increase reliability if we created an install task in `tasks.json`, and then added: `preLaunchTask: "install"` - this way, you'd get a full build  + when you run a launch command? I can try to add an example of this.

CMakePresets.json
```diff
@@ -0,0 +1,28 @@
...
+    "version": 3,
+    "configurePresets": [
+        {
+            "name": "macOS base",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/build/",
+            "generator": "Xcode",
+            "cacheVariables": {
+                "CMAKE_PREFIX_PATH": "/usr/local/opt/qt@5"
```
scztt, line 10: This path isn't standard on mac, I don't think? Definitely different on my system. I wonder if we can insert the output of `brew --prefix qt@5` here?

CMakePresets.json
```diff
@@ -0,0 +1,28 @@
+{
+    "version": 3,
+    "configurePresets": [
+        {
+            "name": "macOS base",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/build/",
```
scztt, line 7: I would suggest removing this? AFAIK this has a reasonable default in the global vscode cmake settings, and I believe this will override a users own settings, which could cause other problems in their build environment.

CMakePresets.json
```diff
@@ -0,0 +1,28 @@
...
+            "generator": "Xcode",
```
scztt, line 8: This should probably be labelled "xcode", otherwise it's too easy to pick it in e.g. vscode and then have a non-functioning build setup.

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,32 @@
...
+            "includePath": [
+                "${workspaceFolder}/common/**",
```
scztt, line 6: Do we need these? Aren't these fed to us by cmake?

.vscode/c_cpp_properties.json
scztt: This will probably cause problems if you don't have the Xcode app installed (and, for a vscode user at least, theres no reason to).

.gitignore
```diff
@@ -37,4 +37,4 @@ __pycache__
 .*.sw?
```
scztt, line 37: Maybe a comment to explain this one, I don't think I know what this file is.

CMakePresets.json
dyfer: Hm, the output of `brew --prefix qt@5` would be better, but it won't always be correct.
Is it still possible to configure/overwrite this on the user end? The prefix path would be different when e.g. building with an official Qt install.

.vscode/c_cpp_properties.json
capital-G: But can you actually build w/o xcode? I thought you need XCode to obtain the clang compiler, so I think we can assume that anyone who wants to contribute to the C++ part of SuperCollider will also have to install XCode?

.vscode/launch.json
capital-G: Well it depends on what you want to do with it - I think the class library is not available at this point, but you can actually start the CLI which may be sufficient.

.vscode/launch.json
capital-G: While it sounds like a good idea, the `install` target takes aronud 30+ secs even w/o code changes due to code signing. Therefore I think it may be better to make this explicit instead of slowing down each debug spawning.

CMakePresets.json
capital-G: Presets are fully optional to use - they are not intended to cover every kind of configuration as this would be a duplication of the manual configuration via cli. But I agree, it would be the best way to cover as much variants as possible as long as it does not make the setup harder - sensible defaults which can be overwritten explicitly.

On the topic of if the output of a shell command could be used, I think this is not possible. At least I haven't found it in the [docs](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).
But I think it could be possible via a cmake script which gets included via a build which can be set via a preset. The script could be something like 

```cmake
execute_process(
  COMMAND brew --prefix qt@5
  OUTPUT_VARIABLE QT5_BREW_PREFIX
  OUTPUT_STRIP_TRAILING_WHITESPACE
)
```

Maybe this would be a good script as most likely 95% of the macOS user base use brew to obtain qt5 and this would allow those users to not have to state the qt path explicitly anymore.

CMakePresets.json
capital-G: I personally think the `binaryDir` is one of the best aspects of CmakePresets - no need to execute `mkdir -p build && cd build` anymore.
Again, presets are fully optional to use, they are opinionated and provide sensible defaults which should cover most setups.

But I don't understand the remark regarding vscode cmake settings? This file is actually independent of vscode.

.gitignore
capital-G: I also don't know this file ending, but I didn't put it there, it is not part of this PR.

.vscode/c_cpp_properties.json
capital-G: I actually don't know - I am a real newbie in C++ and just get started with CMake etc - I just added the distinct paths which had C++ files in it.

CMakePresets.json
capital-G: I think there are multiple ways of writing said file, but other people have to write them for their respective OSes - I think one there are shared functionality there can be talks about how this could be implemented in the best possible manner.
But one thing that I learned from the CI though: don't rely too much on conditions, they create difficult to maintain implicit states, where I would favor code duplication in favor of readability.

I don't understand how you could use this for testing, but I want to actually use this to make the CI setup a bit more manageable.
I think we don't need to limit it to 2 per platform because it is possible to list all presets and add a description for them.

I agree this would have been better to have its own PR, I would have thought that this would be less controversial actually :D I included it b/c it is the preferred way on how to configure C++ projects in vscode.

> Why is this disabling supernova?

I think the default way of building SC is w/o supernova? I always consider it optional, but it would be good to promote it further.

capital-G:
> I would suggest renaming the files in `.vscode` to something like `launch.json.template` or `launch.json.example`.

I also thought about this but I think this creates more problems - if ones `launch.json` diverges from `launch.json.template` one has to maintain it manually - it will just grow stale over time b/c people forget to sync - and there is no thing that tells me "oh, there is this new feature now that you may want to try/activate/include...".
The idea is to share knowledge on how to setup the IDE in a good manner so new people can easily try out some coding instead of configuring x hours their IDE and build chain. At least it took me multiple days to figure this stuff out.

And it is still possible to create custom tasks in `launch.json` - just don't check them into the repository if they are not interesting for other people.

CMakePresets.json
dyfer: I also think that CMakePresets should be a separate PR, maybe preceding the one specifically for VS Code.

> I think the default way of building SC is w/o supernova?

Yes and no :)
Technically the default in CMake is without, but we have included it in the release builds for a while.
One reason for having it as off by default was that it needs additional libraries on macOS (portaudio and fftw). There were also some issues with it running on Windows, so I'm not sure if it's used much on that platform.
I'd personally be in favor of tuning it on by default.

.vscode/launch.json
capital-G: Maybe `"cwd": "${workspaceFolder}",`?

capital-G:
I decided to ditch `CmakePresets.json` in this PR to keep things more atomic - a new PR will try to introduce this.

I also excluded any user specific paths (such as qt5 path) - this makes it currently not possible to configure the build via VSCode.

capital-G:
#6384 tried to use CMake to find the current qt installation which made configuring the IDE much easier/completely automatic, but @dyfer says that this step will be obsolete once Qt6 support is through why is due in the following weeks.

So until Qt6 support is implemented, this PR is blocked, see #5169

dyfer:
Qt6 is in, so I removed the "blocked" label.

.vscode/launch.json
capital-G: CMake tools will trigger a build before launching.

.vscode/c_cpp_properties.json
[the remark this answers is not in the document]
capital-G: This seems most likely optional as on Linux I did not had to provide such a config.

capital-G:
I re-wrote this to use [`cmake-tools` plugin](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) , which provides some implicit magic and makes the setup really easy.

Tested on macOS and linux (fedora).

Can you give this another review @JordanHendersonMusic ?

capital-G:
@dyfer do you think this could be merged? I think this would be a good addition to provide an easy way for debugging with free (as in free beer) software and lowers the barrier to contribute to SC. At least I would have looked forward to this when I started :)

Hopefully someone can provide a working windows config at some point.

capital-G:
@scztt @JordanHendersonMusic can you give this a review? Would really help my development, and I think I would not be the only one.
If this is merged I'd write a guide for the wiki.

dyfer:
I think this is a good addition but can't test right now, I'd appreciate a review from @scztt or @JordanHendersonMusic , thanks!

JordanHendersonMusic:
I'm going to try and get to this tomorrow!

JordanHendersonMusic:
I've tried running  this today, but I'm not really sure how to test it...

I followed your steps in the description, but I just get the following warning.

```
Cannot configure: No configure preset is active for this CMake project
```

capital-G:
> I've tried running this today, but I'm not really sure how to test it...
> 
> I followed your steps in the description, but I just get the following warning.
> 
> ```
> Cannot configure: No configure preset is active for this CMake project
> ```

Just tried this out on a fresh fedora/vscode installation and worked for me.

* Remove build directory (just to be sure)
* switch to this branch `vscode-ide`
* open vscode via `code .`
* install recommended extensions (pop-ups in bottom right corner)
* use ctrl shift p and search for `Cmake configure` and run it
* Build via ctrl shift p and search `Cmake: build target` - choose e.g. sclang executable
* After build, click on the lower bar with a bug on it saying *Launch the debugger for selected target* and the debugger will start, allowing you to debug sclang now :)

JordanHendersonMusic:
I am supposed to be generating my own `CMakePresets.json` after running `Cmake configure`?

.gitignore
```diff
@@ -37,4 +37,8 @@ __pycache__
...
 
 # vs code files
-.vscode
+.vscode/*
+!.vscode/tasks.json
+!.vscode/extensions.json
+!.vscode/launch.json
+!.vscode/c_cpp_properties.json
```
JordanHendersonMusic, line 44: should we add `out/*` since this is going to be putting all the build file there?

capital-G: For me it is putting everything into `build` which is already ignored

capital-G:
> I am supposed to be generating my own `CMakePresets.json` after running `Cmake configure`?

No, you don't need a `CmakePresets.json` - when configuring (or during starting) you will be asked to select a compiler - afterwards everything should just work automatically.

capital-G:
> Okay when I run `Cmake configure` it asks to select a preset and gives a little warning in the corner `Cannot configure: No configure preset is active for this CMake project`. Is that right? I have all the extensions installed.
> 
> If I just select toolchain and just hit enter, it makes a CMakePresets.json.
> 
> Just to be clear, this isn't a problem for me, I just make a cmake preset and it works great! But I'm struggling to get your instructions to work. I'm starting to think we should ship presets with this.

Thats strange b/c I tried it on two "vanilla" linux laptops since and didn't get this dialog.

Can you verify that you don't have `cmake.useCMakePresets` not set in your `.vscode/settings.json`, see https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-presets.md#enable-cmakepresetsjson-in-the-cmake-tools-extension for further information why and when vscode tries to use a `CMakePresets.json`.

JordanHendersonMusic:
Okay so I think that was a problem, but now it won't build because I need to set `SC_EL=no`. I've tried setting all of the following but it still doesn't seem to like it.

```
 "cmake.buildArgs": [
        "SC_EL=no"
    ],
  "cmake.buildEnvironment": {
      "SC_EL": "no"
  },
  "cmake.configureEnvironment": {
      "SC_EL": "no"
  }
```

This is one of the reasons why I think presets are a good idea, because we can set the default one to include as few dependencies as possible so new users perhaps will less dev experience can get right into it.

capital-G:
@JordanHendersonMusic  I'd also appreciate presets as this would allow e.g. NoX11, RaspberryPi, ... presets, but I didn't want to include it in this PR and keep the atomic.

Have you tried [`cmake.configureSettings`](https://vector-of-bool.github.io/docs/vscode-cmake-tools/settings.html#cmake-configuresettings) (or see <https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-settings.md>?
These are client dependent configurations which should go into `settings.json` which is not checked in.

If this gets merged I'd write a Wiki entry where stuff like this gets explained.

JordanHendersonMusic:
I'm just thinking that if this is supposed to be a helpful way for new users to fire up a debugger, and presets help cut down configuration, a few basic ones ought to be included, this way the documentation could just be 'here is how you select the preset, pick the minimal one', if you need custom presets, here is how you add your own.

The other issue is that its not uncommon to have diverging debug and release flags, documenting this with the current approach feels awkward.

Anyway, if you think this PR needs to come first and presets should be separate then I'm fine with that.

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,10 @@
+{
+    "configurations": [
+        {
+            "name": "Mac",
```
dyfer, line 4: could this be "macOS" instead?

.vscode/extensions.json
```diff
@@ -0,0 +1,11 @@
+{
+	"recommendations": [
+		"ms-vscode.cpptools",
+		"ms-vscode.cpptools-extension-pack",
+    "ms-vscode.cmake-tools",
```
dyfer, line 5: is there a reason this is indented differently?

.vscode/c_cpp_properties.json
capital-G: This is somehow used as an identifier by vscode, so i would stick to it, see

image

.vscode/extensions.json
capital-G: ah thx - tab vs spaces... switched to spaces.

capital-G:
> BTW I had to rebase the branch on current develop because of cmake 4.

Rebased

> Which part of this tells CMake to use `build` as the build directory?

We use the default from the cmake plugin which is `${workspaceFolder}/build` - see `cmake.buildDirectory` at https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-settings.md

> BTW on macOS ideally we'd like to build with the Xcode generator.

Is there a benefit using this? I switched to Ninja b/c I otherwise can't build the Install target in CLion.
In the end I think it is more useful to specify this via a `CMakePresets.json` file rather then enforcing this only on the vscode configuration, but this should be another PR b/c this would probably also take multiple iterations.

> And I think it worked, sort of? But I don't know how to run sclang alone so that it finds the class library:

I also wouldn't know currently, I only debugged sclang w/o class library in vscode.

In the end it seems one would have to modify the `launch.json` - we could probably add an explicit debug target for sclang since we now have the ability to specify the class library target explicitly via #6640 (so something like `${workspaceFolder}/ClassLibrary`?).

As you see there are multiple things involved to make this really useful.

The intention of this PR is not to provide a final/definitive configuration but already provides people proper auto-complete, navigating Class implementations and debugging capabilities across all platforms - if you currently open the SuperCollider repository in vscode you don't have anything of these.

My intention here is to share knowledge (aka "my config") here and let other people also improve upon it over time so SC development becomes easier for people who don't own CLion or have access to XCode.


> I don't mind merging this PR, but we definitely need to store some documentation about this - probably on the wiki pages, developer's section. 

Agreed and I will write a small entry - but this PR would already provide a big benefit for people who are used to vscode.

BTW - this is somehow in tandem to my other guide @ https://github.com/capital-G/ExampleUGen which demonstrates how to debug custom UGens via VSCode and CLion.

JordanHendersonMusic (approved):

.vscode/extensions.json
```diff
@@ -0,0 +1,15 @@
...
+    "ms-vscode.cpptools-extension-pack",
+
+    // uses .editorconfig file to apply e.g. 
+    // tab indentation on sc files and space indentation elsewhere
+    "EditorConfig.EditorConfig",
+
+    // syntax highlighting in sc/schelp files for vscode
+    "jatinchowdhury18.vscode-supercollider",
+  ],
+}
```
dyfer, line 15: Purely cosmetic: this file now uses 2-space indentation, while other json files use 4 spaces. Could go in either as is or be changed for consistency.

dyfer:
Thanks for all the explanation!

> We use the default from the cmake plugin which is `${workspaceFolder}/build` - see `cmake.buildDirectory` at https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-settings.md

Ah, good to know.
> 
> > BTW on macOS ideally we'd like to build with the Xcode generator.
>
> Is there a benefit using this? 

Not sure. Xcode is the officially supported generator on macOS and also allows building SC using Xcode app. Since this is the "native" generator, we want to make sure that it works first. But aside from that, I'm not aware of a practical benefit.

> > But I don't know how to run sclang alone so that it finds the class library:
> 
> I also wouldn't know currently, I only debugged sclang w/o class library in vscode.

Haha okay, good to know that what I got was expected at this point :) 

> The intention of this PR is not to provide a final/definitive configuration but already provides people proper auto-complete, navigating Class implementations and debugging capabilities across all platforms

Yeah, sounds good.

> Agreed and I will write a small entry - but this PR would already provide a big benefit for people who are used to vscode.

Perfect, thanks.

> > The documentation should also note what VS plugins are being installed to use this functionality
>
> Yes - I started to annotate the `.vscode/extensions.json` file which then can be linked from the wiki so we don't diverge there.

Great, thanks!

.vscode/extensions.json
capital-G: thx for the hint - i did not format those properly as our `.editorconfig` suggests using 2 spaces for json indentation

https://github.com/supercollider/supercollider/blob/916b10bb59a5020155cc7fa45758bbd9b307645e/.editorconfig#L29-L31

I now applied the 2 space formatting to the other .json files of this PR.

dyfer:
Let's (squash-)merge!

PR #6353, opened 2024-06-14
SCLang: make slot tag fixed width integer

JordanHendersonMusic:
## Purpose and Motivation

On unix `long` is 64 bits, but on windows `long` is 32. Throughout the code assumptions are made that assume the size of the `PyrSlot` to be 128bits. This makes that explicit rather than relying on alignment, which I believe was the original intention as otherwise this could easily have been a uint8.

## Types of changes

- Bug fix

This fix should change nothing, just makes the code more explicit.

## To-do list

Can't test because of broken CI, but should work.

- [ ] Code is tested
- [ ] All tests are passing

### DIFF
--- a/lang/LangSource/PyrSlot64.h
+++ b/lang/LangSource/PyrSlot64.h
@@ -50,7 +50,7 @@ enum {
 };
 
 typedef struct pyrslot {
-    long tag;
+    uint64 tag;
 
     union {
         int64 c; /* char */


telephon:
Just as a reference – recently, I had a conversation with James McCartney and he told me that the intended slot size in sclang would be 64 bit. Everything is done with nan tagging on the basis of this.

He had the impression this had been accidentally changed at some later stage.

JordanHendersonMusic:
The 64 bit build has never used nan tagging. It has always been 128 bits.

https://github.com/supercollider/supercollider/commit/60bb2887fb2c66303594cbbf5d48a0c5e2327d84

The 32 bit build is using nan-tagging and is 64 bits.

JordanHendersonMusic:
The long in PyrMerhodRaw should also be a uint64.

telephon:
> The 64 bit build has never used nan tagging. It has always been 128 bits.
> 
> [60bb288](https://github.com/supercollider/supercollider/commit/60bb2887fb2c66303594cbbf5d48a0c5e2327d84)
> 
> The 32 bit build is using nan-tagging and is 64 bits.
> 
> I could try and add it back in?

If I understood James correctly, he suggested that the 64 bit build could also use nan tagging, yes. But I don't know how it works.

JordanHendersonMusic:
It's not hard to do - assuming all the accessing functions are used (IsNil, slotRawObject and the likes). I'll have a look later.

JordanHendersonMusic:
Closing because https://github.com/supercollider/supercollider/pull/6365 is better.

PR #6354, opened 2024-06-14
Move build jobs in separate reusable workflows

silvanocerza:
## Purpose and Motivation

This PR relates to #6340.

This is the first of a series of PRs that will simplify and enhance the project CI.

In this PR I remove the step to get the version string from the `lint` job as it's not semantically correct and moved it into its own job, I updated `needs` dependencies accordingly.

The `Linux`, `macOS` and `Windows` job instead has been completely moved into reusable workflows called `build_linux.yml`, `build_macos.yml` and `build_windows.yml` respectively.

## Types of changes

- CI

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


silvanocerza:
Rebased to include Mac OS CI fix from #6358.

capital-G (approved):
Looks good to me, but would it be possible to "outsource" every platform into its own file in one go?
this would allow easier navigation in git history/blame in case we need to find "why was this flag turned on" as we can simply jump to the commit before this split.

silvanocerza:
@capital-G 
Good for me. I was planning to make a separate PR per platform to ease review, but if you prefer a single one I can go ahead with that.

capital-G:
> @capital-G Good for me. I was planning to make a separate PR per platform to ease review, but if you prefer a single one I can go ahead with that.

Yes, normally approaching things in an atomic manner is preferable, but here I think it would be good to have a single commit (via squashing) in order to simply get to a state where one can figure out why line `x` exists.

silvanocerza:
Sorry, I had to fix a small mistake I made that I didn't notice. 🙏

capital-G (approved):
Looks good to me, but maybe we wait for another pair of eyes to merge it.
Thanks for this work :)
It should be squashed to make it easier to "go before" via git blame.

dyfer (changes requested):
Thanks for this!

I think that splitting this up is generally good. I wonder if the build matrices should be separated as well, but maybe not in this PR.

However, take a look at the Windows build artifacts. The version number doesn't seem to be passed properly, as it's missing from the filename.

dyfer (approved):
Thanks! I think it's good now.

We'll find out for sure after the next SC release, I guess (but it looks like it should work fine).

PR #6356, opened 2024-06-15
lang: LinkClock beats/secs conversion should trap 'inf'

### LINKED ISSUE #6355: LinkClock returns garbage floats for beats2secs on 'inf'
## Environment

* SuperCollider version: 3.14
* Operating system: Linux

## Steps to reproduce

```supercollider
TempoClock.default.beats2secs(inf);
-> inf

l = LinkClock.new;
l.beats2secs(inf);
-> -9223372035485.3
```

At minimum, it's an inconsistency: LinkClock is in this case *not* a drop-in replacement for TempoClock (but should be).

I'm fairly certain it wasn't like this in the past (or I would have seen bugs in my live coding system long ago -- just spotted this yesterday). If that's true, then it's a regression.

I suspect we could just trap 'inf' in our primitive. Let me see...

EDIT: Ohhhhh... https://en.cppreference.com/w/cpp/numeric/math/isinf

> [GCC](https://gcc.gnu.org/wiki/FloatingPointMath) and [Clang](https://clang.llvm.org/docs/UsersManual.html#controlling-floating-point-behavior) support a -ffinite-math option (additionally implied by -ffast-math), which allows the respective compiler to assume the nonexistence of special IEEE-754 floating point values such as NaN, infinity, or negative zero. In other words, std::isinf is assumed to always return false under this option.

I bet the Link library is being compiled with -ffinite-math.

jamshark70:
## Purpose and Motivation

Fixes #6355.

For compatibility with TempoClock, beats2secs(inf) and secs2beats(inf) should return inf (and, -inf --> -inf).

## Types of changes

- Bug fix (I believe it's a regression -- critical bug for my live coding system actually, it breaks slurring notes)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- ~~ [ ] Updated documentation~~

### DIFF
--- a/lang/LangPrimSource/SC_LinkClock.cpp
+++ b/lang/LangPrimSource/SC_LinkClock.cpp
@@ -45,12 +45,14 @@ void LinkClock::SetAll(double tempo, double inBeats, double inSeconds) {
 }
 
 double LinkClock::BeatsToSecs(double beats) const {
+    if(std::isinf(beats)) return beats;
     auto sessionState = mLink.captureAppSessionState();
     double secs = linkToHrTime(sessionState.timeAtBeat(beats, mQuantum)) - mLatency;
     return secs;
 }
 
 double LinkClock::SecsToBeats(double secs) const {
+    if(std::isinf(secs)) return secs;
     auto sessionState = mLink.captureAppSessionState();
     double beats = sessionState.beatAtTime(hrToLinkTime(secs + mLatency), mQuantum);
     return beats;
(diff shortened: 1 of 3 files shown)


telephon (approved):

capital-G:
@jamshark70 thanks for this fix - could you rebase this PR from the current develop branch? The macOS build was broken and therefore the test pipeline did not run, but I'd like to see if your tests pass in CI.

If it succeeds it can be merged IMO.

jamshark70:
> could you rebase this PR from the current develop branch?

I see -- btw the new stuff in develop post-dates these commits.

PR #6357, opened 2024-06-15
Deactivate macOS universal build

capital-G:
## Purpose and Motivation

A quick fix to get our pipeline working again, see https://github.com/supercollider/supercollider/issues/6352. This must be a temporary solution as we currently do not have a native build for Apple ARM computers, but it helps us to get our build and testing pipeline up and running again.

* Disable universal build
* Switching to x64-legacy binary for testing on macOS

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation
- [ ] This PR is ready for review

Issue #6352, opened 2024-06-13
macOS Universal build verification is broken


capital-G:
Once again, (most likely) a brew update breaks our CI :/

See https://github.com/supercollider/supercollider/actions/runs/9501252970/job/26192773490

```log
2024-06-13T16:22:25.7103150Z -- Verifying app
2024-06-13T16:22:25.7165060Z -- ===========================================================================
2024-06-13T16:22:25.7166950Z -- Analyzing app='/Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app'
2024-06-13T16:22:25.7169180Z -- bundle='/Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app'
2024-06-13T16:22:25.7171520Z -- executable='/Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/SuperCollider'
2024-06-13T16:22:25.7173060Z -- valid='1'
2024-06-13T16:22:26.2895860Z -- executable file 1: /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/SuperCollider
2024-06-13T16:22:27.1989680Z -- 
2024-06-13T16:22:27.1996390Z warning: cannot resolve item '/usr/local/Cellar/glib/2.80.3/lib/libglib-2.0.0.dylib'
2024-06-13T16:22:27.1997310Z 
2024-06-13T16:22:27.1997540Z   possible problems:
2024-06-13T16:22:27.1998170Z     need more directories?
2024-06-13T16:22:27.2108150Z     need to use InstallRequiredSystemLibraries?
2024-06-13T16:22:27.2109000Z     run in install tree instead of build tree?
2024-06-13T16:22:27.2111680Z 
2024-06-13T16:22:27.7322430Z -- executable file 2: /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/sclang
2024-06-13T16:22:27.9222680Z -- 
2024-06-13T16:22:27.9224360Z warning: cannot resolve item '/usr/local/Cellar/glib/2.80.3/lib/libglib-2.0.0.dylib'
2024-06-13T16:22:27.9227740Z 
2024-06-13T16:22:27.9230180Z   possible problems:
2024-06-13T16:22:27.9231320Z     need more directories?
[...]

PR #6358, opened 2024-06-15
Skip verify on macOS universal CI

capital-G:
## Purpose and Motivation

A **temporary** fix for building the macOS universal binary, see #6352 
Supersedes https://github.com/supercollider/supercollider/pull/6357 which deactivated the build completely, here we just de-activate the verifying step during build macOS - I am surprised that the binary can be picked up by a different runner for testing though.

## Types of changes

- Bug fix
- 
## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -208,7 +208,7 @@ jobs:
             build-readline: false
             build-fftw: false
             artifact-suffix: 'macOS-universal'
-            verify-app: true
+            verify-app: false
 
           - job-name: 'x64 legacy'
             os-version: '12'


silvanocerza:
Seems like the build was successful. 👀

JordanHendersonMusic:
What does `verify-app` actually do?

silvanocerza:
> What does `verify-app` actually do?

If it's `true` the `verify_app(<app>)` function is called. From the [docs](https://cmake.org/cmake/help/latest/module/BundleUtilities.html): 

> Verifies that an application `<app>` appears valid based on running analysis tools on it

JordanHendersonMusic:
Ah! Thanks for the quick reply.

That might explain why the test still works because verify_app is trying to verify scide, whereas the test only needs sclang?

JordanHendersonMusic (approved):
A temporary fix for the CI. Thanks @capital-G!

Issue #6245, opened 2024-04-02
Build fails with strict-aliasing violations


eli-schwartz:
I tried to build with the following *FLAGS to optimize the build: `-flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing`

The -Werror=* flags are important to detect cases where the compiler can try to optimize based on assuming UB cannot happen, and miscompile code that has UB in it. strict-aliasing issues are always bad but LTO can make them even worse.

I got this error:
```
[...]

PR #6365, opened 2024-06-20
SCLang: Make 64-bit slot use nan boxing.

JordanHendersonMusic:
This commit also unites the implementations of the 32 and 64 bit slots.

## Purpose and Motivation

This will half the amount of ram used in the 64 bit build. While meaning we don't have to support two implementations for `PyrSlot`.

## Types of changes

- Internal change, should functionally change nothing.

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


JordanHendersonMusic:
@telephon Here is the 64-bit slot implementation I mentioned, took a day or so.

The CI doesn't test on 32-bit builds and I don't know if big endian gets tested (I usually mess that up the first time),
but I think the basic approach is here.

scztt:
This is a great change, nice work. I'll try to find some time do give it a thorough review, but in general it would be nice to land this!

lang/CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ set(sclang_sources
 	LangSource/SimpleStack.cpp
 	LangSource/VMGlobals.cpp
 	LangSource/dumpByteCodes.cpp
+        LangSource/pyslottest.cpp
```
scztt, line 74: Indent incorrect?

JordanHendersonMusic: This file needs removing anyway. I haven't looked at how to add cpp tests yet.

lang/LangSource/PyrMathOps.cpp
```diff
@@ -227,7 +227,7 @@ int doSpecialUnaryArithMsg(VMGlobals* g, int numArgsPushed) {
             SetTrue(a);
             break;
         case opAsInteger:
-            SetTagRaw(a, tagInt);
```
scztt, line 230: Can we remove the `SetTagRaw` function entirely? it seems like your change eliminates nearly all of them (sorry if you already did this and I missed it)

lang/LangSource/PyrMathOps.cpp
```diff
@@ -648,10 +648,10 @@ int doSpecialBinaryArithMsg(VMGlobals* g, int numArgsPushed, bool isPrimitive) {
                 SetRaw(a, slotRawInt(a) ^ slotRawInt(b));
                 break;
             case opLCM:
-                SetRaw(a, sc_lcm((long)slotRawInt(a), (long)slotRawInt(b)));
+                SetRaw(a, static_cast<int32_t>(sc_lcm((long)slotRawInt(a), (long)slotRawInt(b))));
```
scztt, line 651: Instead, would it make sense to overload `sc_lcm` so that these combinations of types behave correctly (e.g. move casts to one place)?

lang/LangSource/PyrMathOps.cpp
```diff
@@ -675,31 +675,31 @@ int doSpecialBinaryArithMsg(VMGlobals* g, int numArgsPushed, bool isPrimitive) {
                 SetFloat(a, pow((double)slotRawInt(a), (double)slotRawInt(b)));
                 break;
             case opShiftLeft: {
-                long ia = slotRawInt(a);
-                long ib = slotRawInt(b);
+                int64_t ia = slotRawInt(a);
```
scztt, line 678: `auto`?

scztt: I think maybe the intention is to move to explicit types but ensure generated code doesn't change - in that case, I think this is correct.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -4207,7 +4207,9 @@ bool findVarName(PyrBlock* func, PyrClass** classobj, PyrSymbol* name, int* varT
     }
     if (name->name[0] >= 'A' && name->name[0] <= 'Z')
         return false;
-    for (j = 0; func; func = slotRawBlock(&func->contextDef), ++j) {
+
+    j = 0;
+    while (func) {
```
scztt, line 4212: More explicit with `func != nullptr`?

lang/LangSource/PyrSlot.h
```diff
@@ -27,16 +27,398 @@ A PyrSlot is an 8-byte value which is either a double precision float or a
...
+    uint64_t u_raw;
+    double u_double;
+    PaddedNumber<char> u_char;
+    PaddedNumber<int32_t> u_int;
+    MaybePaddedPointer<void*> u_ptr;
+    MaybePaddedPointer<struct PyrObjectHdr*> u_objectHeader;
+    MaybePaddedPointer<struct PyrSymbol*> u_symbol;
+
+    template <uint64_t T> [[nodiscard]] bool tagChecker() const noexcept {
+        return isBoxed() && ((u_raw >> 48) == (T >> 48));
```
scztt, line 203: Can we extract either the `48` literal or this entire `==` operation to something descriptive?

lang/LangSource/PyrSlot.h
```diff
@@ -27,16 +27,398 @@ A PyrSlot is an 8-byte value which is either a double precision float or a
...
+//                *
+
+
+static constexpr uint64_t boxedBit = 0x0001000000000000;
+static constexpr uint64_t nanExponent = 0x7FF8000000000000;
+static constexpr uint64_t boxed = boxedBit | nanExponent;
+
+static constexpr uint64_t pointerMask = 0x0000FFFFFFFFFFFF;
+
+static constexpr uint64_t objHdrTag = 0xFFFF000000000000; // boxed | 0x0001000000000000;
```
scztt, line 111: Group these as an enum? e.g.
```
enum class Tag : int {
    objHeader = 0xFFFF000000000000
    // ...etc
};
```

lang/LangSource/PyrSlot.h
```diff
@@ -27,16 +27,398 @@ A PyrSlot is an 8-byte value which is either a double precision float or a
...
+    tagNil,
+    tagFalse,
+    tagTrue,
+    tagPtr,
+    tagFloat,
+    tagUnused,
+};
+
+inline int GetTag(const PyrSlot* slot) noexcept {
+    if (slot->isChar())
```
scztt, line 311: I wonder if we should move nil/bool/int cases to the top here, since they are more likely to be performance critical and common? Actually, an empirical answer to "which of these is called the most" would be nice to have.

lang/LangSource/PyrSlot.h
```diff
@@ -27,16 +27,398 @@ A PyrSlot is an 8-byte value which is either a double precision float or a
...
+#endif
+};
+
+static_assert(sizeof(MaybePaddedPointer<void*>) == sizeof(double));
+
+// cpp reference
+template <class To, class From>
+std::enable_if_t<sizeof(To) == sizeof(From) && std::is_trivially_copyable_v<From> && std::is_trivially_copyable_v<To>,
+                 To>
+bit_cast(const From& src) noexcept {
```
scztt, line 177: You're a better judge than I, but it seems like there are many things in this file that are more or less implementation details for `PyrSlot`. Could we move those to a `namespace detail`?

lang/LangSource/PyrSlot.h
JordanHendersonMusic: I'll take some measurements! (Probably be next week)

lang/LangSource/PyrSlot.h
```diff
@@ -27,16 +27,398 @@ A PyrSlot is an 8-byte value which is either a double precision float or a
...
+            return false;
+        }
+        if (rhs.isDouble())
+            return false;
+        return lhs.u_raw == rhs.u_raw;
+    }
+
+    [[nodiscard]] constexpr static PyrSlot make(double d) noexcept { return { PrivateTag(), d }; }
+    [[nodiscard]] static PyrSlot make(char c) noexcept {
+        return { PrivateTag(), charTag | static_cast<uint64_t>(bit_cast<uint8_t>(c)) };
```
scztt, line 229: Minor but, I wonder if it would be better to switch to a three arguement private constructor, and then consolidate the `|` bit operation to one place?

scztt:
I did a quick pass review, mostly minor changes. I'll try to look more carefully at the critical things this weekend.

lang/LangSource/PyrMathOps.cpp
JordanHendersonMusic: Yes it removed it completely The approach doesn't really make sense with nan boxing on 64-bit systems. Besides, it only saved a couple of shifts and bit ops.

lang/LangSource/PyrSlot.h
JordanHendersonMusic: I went with 
```
struct Masks {
 static constexpr uint64_t boxed = ...;
};
```
Otherwise you have to do something like `static_cast<std::underlying_type<Masks>>(Masks::boxed)` to get the value out correctly...

JordanHendersonMusic:
@scztt I've  just added another commit which could be its own PR, but requires this change,
which removes PyrMethodRaw and turns it into a proxy so that strict aliasing isn't violated https://github.com/supercollider/supercollider/issues/6245.

Let me know if you think this would be better in a separate PR rather than a separate commit, and if so, how that is normally dealt with (do I make a new branch from this one?).  Its really just a bunch of find and replaces.

lang/LangSource/PyrSlot.h
JordanHendersonMusic: I've changed this now to better reflect the structure of the tags, using `if`s for the values (nil, true, and false) and a switch on the remaining tags.

timblechmann:
> Performance should be drastically improved on 64 bits builds

out of curiosity: do you have numbers? slot-type dispatching will be much simpler with the tag enum than extracting the type via a bitmask (the switch can generate a jump-table, which may be friendlier to the branch predictor). it could be that reduced memory footprint outweighs the faster slot type access /type dispatching, though it would be good to validate this with numbers (in my gut feeling, packing the type into nans may actually reduce interpreter performance rather than increasing, but i could very well be wrong)

JordanHendersonMusic:
> do you have numbers?

Not yet but I will this week when I get a chance.

It should be as simple as doing one mask, then using a switch statement, so I can't imagine it would be significantly slower. The main benefit should come in array loops. But because GC works in size bands this could have strange effects, as it could increase the amount of allocations initially, but decrease them over time as more memory can be recycled.

JordanHendersonMusic:
@timblechmann 

# Benchmarks

*Let me know if anyone wants to see something specific and I'll add it.*

---

### Compiling class library 
*(This one is the most impressive but has the least samples...)* 
| dev | nan | delta |
| --- | --- | --- |
|1.86 |  1.27 | -46% |

---

```
bench {
	10000.do{
		(0..1000).collect( _ * 2)
	}
}
```
| dev | nan | delta |
| --- | --- | --- |
|1.011892081 | 0.916178409 | -10% |

---

```
bench {
	1000.do {
		(0..1000)
		.collect(Ref(_))
		.collect{ |r| r.().pow(2.0) }
		.collect{ |x| x.asString }
		.reduce('++')
	}
}
```
| dev | nan | delta |
| --- | --- | --- |
| 1.588166514 | 1.694247884 | **+6%**|

** Regression***

I haven't look at why yet.

---

```
bench {
	var r = Pwhite().asStream;
	r.next;
	1000.do {
		Array.fill(100000, {r.next})
	}
}
```

| dev | nan | delta |
| --- | --- | --- |
| 22.074586908 | 20.814369008 | -6% |

JordanHendersonMusic:
@timblechmann 

I've spent a bit of time today doing some benchmarks. They are really inconsistent, if you have a better suggest let me know.


```
~b = { |func|
	var n = 1000;
	var f = n.collect {
		func.bench(false)
	};

	var m = (f.sum / n);
	var d = f.collect({|i| (m - i).pow(2)}).sum / n;

	"MEAN = %, STD = %\n".format(m, d).post;
};

// TEST 1
~b.({
	var e = ();
	10000.do{ |i|
		var n = 10000.rand;
		if(e[n.asSymbol].isNil) {
			e[n.asSymbol] = ();
		} {
			e[n.asSymbol][i.asSymbol] = n
		}
	};
});
// TEST 2
~b.({
	10000.collect( _ * 2 );
});
// TEST 3
~b.({
	10000.collect({|n| n.asFloat * 2.0 });
});
// TEST 4
~b.({
	10000
	.collect(Ref(_))
	.collect(_.value)
	.collect(Ref(_))
	.collect(_.value)
	.collect(Ref(_))
	.collect(_.value)
	.collect(_.asString)
	.collect(_.asArray)
	.collect(_.asInteger)
	.sum
	;
});
// TEST 5
~b.({
	var p = Pwhite().asStream;
	Array.fill(10000, {p.next})
});
```

Results are the mean of 3 runs of the above code.

| Develop Branch| Nan Boxing | percent diff |
|-|-|-|
|0.00867315753499993|0.00816365047666667|%+6.24116698515622|
|0.000800361019666723|0.000805761256000033|%-0.670203027645899|
|0.000879039181333413|0.000887520166000057|%-0.955582193119753|
|0.0133162953333333|0.0127610672803337|%+4.35095310448945|
|0.00203714231100007|0.00208340391000003|%-2.22048152918969|


Average percent change : %+1.34917066793807

I really though it would be faster...

timblechmann:
GC is always very slow. i was always wondering if it makes sense to increase the number of size classes to reduce the size of the lists for the individual size classes.

though i guess there may be difference between stressing the GC as opposed to stressing the VM (iirc some matrix inversions used to be rather slow)

> It should be as simple as doing one mask, then using a switch statement, so I can't imagine it would be significantly slower.

i guess a simple index into a jump table can help codegen and speculative execution (branch target prediction), but it's just a gut feeling. 15-ish years ago, the goal for optimized interpreter design was to reduce branch mis-predictions (compare e.g. the threaded code (computed gotos) in the bytecode interpreter). back then when i introduced it to sclang, it gave a bit of a a speedup (maybe 5-10%ish, i don't recall the numbers).
i'd also expect different CPUs to perform slightly differently (depending how smart the branch predictor is)

JordanHendersonMusic:
> back then when i introduced it to sclang, it gave a bit of a a speedup (maybe 5-10%ish, i don't recall the numbers). @timblechmann 

Are you talking about the 128bit slot? Where I see performance hits, they are small 1-2% and probably within the margin of error, but I'm on a ryzen 5600X, so reasonably new.

timblechmann:
> > back then when i introduced it to sclang, it gave a bit of a a speedup (maybe 5-10%ish, i don't recall the numbers). @timblechmann
> 
> Are you talking about the 128bit slot? Where I see performance hits, they are small 1-2% and probably within the margin of error, but I'm on a ryzen 5600X, so reasonably new.

ii was referring to computed gotos (601ff4d25a8c226bf12dba8fdf2f76b23ef99250), where the goal is to reduce branch mispredictions in the interpreter.

JordanHendersonMusic:
In some micro-benchmarks I was able to get better performance (10%) using switch statements; seems if the compiler knows it can't fall out of the switch it can do some more optimisation, I did it by adding `__unreachable()` to the default case.

I've also updating the PR with a much cleaner switch on the types.

JordanHendersonMusic:
Just so its clear, the CI is passing... its just doing an extra run on state of the code as was initially committed (this might be my fault my force pushing...).

JordanHendersonMusic:
## Update

@capital-G this might interest you too.

There was an inconsistent segfault... Turns out, it came from the same bit of code that had been causing the CI tests to fail. I guess the old implementation caused the test to fail, perhaps the slot to return an incorrect value, whereas this one causes a segfault. It's `TestUGen_RTAlloc`. 

I've commented it out,  as the other tests call into the slot millions of times, meaning the error is most likely in the test, not this pr's code. 

I **think** the issue is with FloatArray, either in the sum method, iteration, or in initialisation. I've also deleted a line in a signal primitive that was checking for the rate, which might have been remnants of an old feature that needed to be deleted.

I think this is ready for a thorough review, with the caveat that the test is commented out, and I'm still yet to move the test over to a better location.

JordanHendersonMusic:
I have reenabled the test after finding the cause of (at least some) of the CI's issues #6387. Since then, they pass, every time.

The segfault was being caused by convolution3 sometimes produce nans, but not normal nans, weird nans that collided with the object tag, causing a double to be read as a pointer.

To solve this I have added a test for nan in the double section of the slot's constructor. This is not an optimal solution as it will have a performance cost. However, this could be an issue anytime supercollider receives raw data external to the VM without checking it.

lang/LangSource/PyrMathOps.cpp
JordanHendersonMusic:
```
inline long sc_lcm(long a, long b) 
inline long sc_lcm(int a, int b) 
```
On windows long is the same size as int, that is way the cast must go here.
I think sc should be moved to fixed width types throughout, but that is a different PR.

lang/LangSource/PyrMathOps.cpp
JordanHendersonMusic: Again, this is the long as int32 vs long as int64 issue, I just assumed since we are about to do some shifting, that the larger type was needed.

lang/LangSource/PyrSlot.h
JordanHendersonMusic: done

telephon:
I've had a look through with James McCartney today over this as well, and he thinks that it will be an overall gain because more objects fit into a smaller memory for operation.

telephon:
It should be added to the information about this merge request that nan-boxing will reduce not only RAM usage, but also allow to use the processor cache more efficiently. An example is random access to a larger array, such as a large dictionary like Event.

JordanHendersonMusic:
While this is theoretically true, I am yet to actually measure a performance increase. I think we have to be really careful with this one and claims about performance - the reduction in memory is undoubtedly though.

capital-G:
@JordanHendersonMusic what is the state of this? Can there be something done to help you to get this merged, e.g. organizing an additional review or having a vote if this should be merged?

BTW - IIUC "nan-boxing" correctly: sclang is using 64bit floats to bundle type and data storage into one object such that sclang does not have to operate so much on the heap but can use the stack instead, which creates some performance boost.
An sclang-object is thereby considered a float if the `nan` bit of float is false, and else the "upper" 32 bits(?) of the float will be used to store information about the object.
But as we are nowadays most often on 64 bit platforms, we can't store a 64 bit pointer on the lower 32 bits of the 64 bit float anymore - is there some concatenation happening here or what is nan-boxing on 64 bit systems? Obviously sclang has found a way b/c it is running on 64bit systems so maybe it is out of scope for the PR, but you seem currently deep within the topic and would be great if you could "dump your knowledge about it" in some way so the next person coming along can pick this up w/o so much detective work :)

Would be great to provide some insights and/or documentation here.

JordanHendersonMusic:
> what is the state of this?

It's functional, but it needs extra logic where SC receives float from io as we need to guarantee they haven't given us odd nans. I think I've put this in the constructor, which is slow.

I could also improve my comments.

Other than that, it needs someone to review the core approach here as I think that's solid.


> sclang is using 64bit floats to bundle type and data storage into one object such that sclang does not have to operate so much on the heap but can use the stack in.

This is not quite right as it doesn't affect heap usage. SC stores primitive types (floats, int, bool, ptrs) in the slot. You need a way to know which of these types is held. It's currently a tagged union. For alignment reason this means the slot would be 128 bits. Wasting 60ish bits. Nan boxing lets you store data in  the top part of the float as there are many possible nan states. SC supports i32s, so this is just stored in the lower bits. Your right that pointers are oddities, but actually user space programs only use the lower 48 (or was it 42?) bits of the pointer, you just have to put them back to zero before dereferencing. You can look up 'canonical pointer form' for more info on that. Technically, you can also store information in the lower bits of the pointer too because nothing is bit aligned.

This means all slots get their size cut from 128bits to 64. Which definitely means less memory usage, and theoretically means better performance - but I haven't been able to measure that - as sc bottleneck is the GC by a large margin. It may in future be worth editing the GC size bands, particularly since this reduces the memory usage, which could lead to a performance boost (speculation).

JordanHendersonMusic:
I've updated this PR to : 
* tidy parts of the code
*  move the nan checker into the rest of the code base, this means it is the responsibility of primitives to know whether they might receive bad nan values (only possible when loading a double/float from outside sc)
* update signal as it was previously being accessed as a PyrObject when in fact it is a PyrFloatArray, this is because 20 years ago it was going to have a rate symbol prep-pended to it. This was an issue for this PR because it is a little strictly on casting.


If anyone has time I'd appreciate either a review or some testing because this should be good to go now. @scztt I don't suppose you'd like to have another peek at this?

dyfer:
@JordanHendersonMusic re: testing on Windows
I've combined this branch with one that has a running Windows 32-bit testsuite:
[topic/nan-tagged-slot-testonwindows](https://github.com/supercollider/supercollider/tree/topic/nan-tagged-slot-testonwindows)
Test run is here: https://github.com/supercollider/supercollider/actions/runs/14611666939
The only failures I saw were with the tests that somtimes fail also on Linux and macOS. 
Feel free to use this branch as you see fit :)

JordanHendersonMusic:
@dyfer Ah thank you! In that case this should be good. I'd still like a good review of the C++, the general approach I've taken, along with the way I've dealt with nan values; other than that, this should be good.

lang/LangSource/cmdLineFuncs.cpp
```diff
@@ -20,7 +20,12 @@
 
 #include "SC_LanguageClient.h"
 
+#include "PyrSlot.h"
+
 int main(int argc, char** argv) {
+    // TODO jordan: temporary move to tests
```
dyfer, line 26: todo?

JordanHendersonMusic:
If this is intended for 3.15, it should be merged sooner rather than later in case there are any problem. Leaving this in develop for a while is really the only way to test it.

lang/LangSource/PyrSignal.cpp
```diff
@@ -18,6 +18,9 @@
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 */
 
+
+// Note: For some reason Signals are PyrFloatArrays, this seems incorrect and they should be PyrFloatArrays instead.
```
dyfer: Is this comment accurate?

JordanHendersonMusic: Ah no I had to change it! They are float arrays now.

capital-G:
@JordanHendersonMusic I'd say lets merge! :) Can you rebase / resolve the merge conflicts?

telephon:
Let's wait till the end of the week.

JordanHendersonMusic:
Alright, once the tests pass this should be good to go

lfnoise:
Two notes.

You check for colliding NaNs at nearly every call to SetFloat. I don't think any other language implementation (LuaJIT, in particular) checks for that, but relies on the behavior that CPUs don't produce certain bit patterns of NaNs.

You define a lot of "make" calls to create values instead of providing constructors. Why? To me, one of the strengths of C++ is that you can create domain specific languages because of operator and function overloading and implicit conversions via constructors.

JordanHendersonMusic:
@lfnoise thank you very much for having a look at this - really appreciate it!!

We only check for nan when set float is called like so... `SetFloat<CouldBeBadNan>(...)`. If it is called without the template, it doesn't do a check. I have worked on the assumption SC doesn't itself produce nans, but that they might come from external sources - I might have been over (or under) zealous with these checks, but this definitely doesn't check every math op. I did that initially, it was very slow, but it might be something worth doing in debug.


I see your point about the `make` calls, and remember there was a reason why... But I can't recall it because I wrote this a while ago now. I think it was something to do with msvc and the other compilers disagreeing on what was valid c++, maybe something to do with union initialisation in constructors ...(?).

Do you think this pr should be merged as is, or do you think I should revisit the `make` stuff?

lfnoise:
ship it.

PR #6372, opened 2024-06-25
Split test step in multiple reusable workflows

silvanocerza:
## Purpose and Motivation

In a similar way to #6354 this PR splits the `test` step in separate reusable workflows for each platform.

I also renamed some existing steps for clarity.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -2,18 +2,18 @@ name: CI
 on:
   push:
     paths-ignore:
-      - "examples/**"
-      - "HelpSource/**"
-      - "sounds/**"
-      - "*.md"
+      - 'examples/**'
+      - 'HelpSource/**'
+      - 'sounds/**'
+      - '*.md'
   pull_request:
     paths-ignore:
-      - "examples/**"
-      - "HelpSource/**"
-      - "sounds/**"
-      - "*.md"
+      - 'examples/**'
+      - 'HelpSource/**'
+      - 'sounds/**'
+      - '*.md'
   schedule:
-    - cron: "0 0 * * 0" # run weekly to refresh cache
+    - cron: '0 0 * * 0' # run weekly to refresh cache
 
 jobs:
   lint:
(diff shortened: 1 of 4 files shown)


silvanocerza:
@dyfer I think you'll appreciate this one. :)

dyfer:
Thanks @silvanocerza!
As far as testing goes, this looks good to me.

I just wanted to ask - why all the changes to the quotes? While this seems to work fine, sometimes such changes have unintended consequences, as IIUC the single vs double quotes are handled slightly differently in yaml. I'm mostly concerned about the release and S3 deployment code, which won't be tested until we do the release.

If there's a reason for the change (GH guidelines, consistent formatting etc), please let me know. If there isn't, I'd prefer for this PR to not change the quotes. Thanks again!

silvanocerza:
@dyfer my bad, I have auto formatting on save turned on so and it converts single quotes to double in YAML. It follows the `.editorconfig` but there's no standard and globally supported way to define the quotes to use.

The only difference between quotes style in YAML is how escape sequences are handled, it doesn't affect any string in our case. I'll revert it anyway, those changes are unrelated now.

Ideally I'd like to introduce a better formatter that can be configured with something more than `.editorconfig` as it's quite lacking. Otherwise we risk having a mix of different styles that just cause confusion. I'd lean on Prettier as it supports basically any format.

Formatting style doesn't matter to me, the important thing is that it's well defined.

capital-G (approved):
Thanks - making the test require only their platform is a very nice feature!

Merge note: Should be squashed.

PS: @silvanocerza do you have a suggestion for a yaml linter/formatter workflow? But maybe this is also overkill? I hope we don't need to change those strings all the time :)

silvanocerza:
@capital-G I'd love to bring in a formatter for YAML and other config file like Prettier, even with a pre-commit hook. Though I realise that might bother some people.

So I'd go with [`check-yaml` pre-commit hook](https://github.com/pre-commit/pre-commit-hooks/tree/main?tab=readme-ov-file#check-yaml) to start off. Or even better [`actionlint` pre-commit hook](https://github.com/rhysd/actionlint) to lint just the workflow files since those are the only YAML files we have as of now. I was actually planning to bring it in right after this PR.

capital-G:
> or even better [`actionlint` pre-commit hook](https://github.com/rhysd/actionlint) to lint just the workflow files since those are the only YAML files we have as of now.

That's a nice tool, although it would add Go or a binary to our dev dependencies, therefore I think it would be better to go with something like https://github.com/adrienverge/yamllint ? But that's for another PR/Issue :)

dyfer (approved):
Thanks @silvanocerza !

PR #6373, opened 2024-06-26
Add workflow to lint workflows

silvanocerza:
## Purpose and Motivation

This PR adds a new workflow that runs [`actionlint`](https://github.com/rhysd/actionlint), a workflow files static checker.

This workflow will run only when a workflow YAML file is created or modified.

### DIFF
--- a/.github/workflows/workflows_linting.yml
+++ b/.github/workflows/workflows_linting.yml
@@ -1,0 +1,21 @@
+name: Github workflows linter
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/**"
+
+jobs:
+  lint-workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+
+      - name: Install actionlint
+        run: go install github.com/rhysd/actionlint/cmd/actionlint@latest
+
+      - name: Run actionlint
+        run: actionlint


silvanocerza:
@capital-G I'd like to add a pre-commit hook for this check too. Ideally the [`actionlint-docker`](https://github.com/rhysd/actionlint/blob/9fe5134f105bcb3687cd5c38466b15458b9bb65e/.pre-commit-hooks.yaml#L10-L16) one to avoid the burden on contributors. 

In any case it will be skipped if Github workflows are not modified in any way when commiting, so the burden will be none for most developers.

Though my main concern is [this step](https://github.com/supercollider/supercollider/blob/74bc5959bc2b5e2673a77906c3ddc7734e934f8d/.github/workflows/actions.yml#L35-L38). That hook must not run there as it's unrelated to code changes. One solution would be to run only certain checks explicitly. Another solution could be skipping the `actions.yml` workflow if there are no code changes, though there could be a PR with both code and workflow changes. Albeit that's rare we might be running the workflows linting multiple times, especially annoying if it fails too.

capital-G:
> Ideally the [`actionlint-docker`](https://github.com/rhysd/actionlint/blob/9fe5134f105bcb3687cd5c38466b15458b9bb65e/.pre-commit-hooks.yaml#L10-L16) one to avoid the burden on contributors.

IIRC pre-commit will install the hooks once they are declared and not when they are actually used - I think this could introduce docker as a dev dependency which I would like to omit.
Another approach could be to "abstract it away", basically using a shell script which is only executed when the workflow files are touched which then calls the formatter (but a shell script is not cross-platform compatible :/). 

> Though my main concern is [this step](https://github.com/supercollider/supercollider/blob/74bc5959bc2b5e2673a77906c3ddc7734e934f8d/.github/workflows/actions.yml#L35-L38). That hook must not run there as it's unrelated to code changes.

I really don't understand this part - the `pre-commit` step are linked to code changes. And this step needs to be stateless, so we always need to verify every file in this case.

silvanocerza:
> IIRC pre-commit will install the hooks once they are declared and not when they are actually used - I think this could introduce docker as a dev dependency which I would like to omit. Another approach could be to "abstract it away", basically using a shell script which is only executed when the workflow files are touched which then calls the formatter (but a shell script is not cross-platform compatible :/).

Good point, though I would expect a developer to have Docker installed. Using a custom script could be a valid solution, though that's extra stuff to maintain.

For the time being we could go without the hook I guess.

> I really don't understand this part - the `pre-commit` step are linked to code changes. And this step needs to be stateless, so we always need to verify every file in this case.

`pre-commit` in this project is only related to code changes so it's fine using it in that step for the time being. Though that step ideally must only check code files, and no other file like config YAMLs, JSON or whatnot. 

Adding a pre-commit hook that lints any file that is non code related is not a problem if every checks passes on that step. Though it might be confusing in case of errors since you could get errors to files that are unrelated to code.

Also it might also push you to avoid running checks for different files in different workflows since every check is already running there. Problem is that you're always building and running tests in CI even for changes that don't influence code in any way.

This goes back to my argument of running pre-commit hooks explicitly with `pre-commit <hook>` in CI. This way we have a single source of truth for all checks, but we only run the checks required.

capital-G:
> Adding a pre-commit hook that lints any file that is non code related is not a problem if every checks passes on that step. Though it might be confusing in case of errors since you could get errors to files that are unrelated to code.

Just to be clear, you consider config code in a json/yaml not code in this case, right?
But as we run a stateless `pre-commit run --all-files` in CI we guarantee that every file in the repo always passes every pre-commit check - so if one forked from a good state these checks would only fail if one modifies such files.

> Also it might also push you to avoid running checks for different files in different workflows since every check is already running there.

I don't understand this statement, can you give more explanation on it?

> Problem is that you're always building and running tests in CI even for changes that don't influence code in any way.

Yes, but I'd argue that this is exactly what we want as this guarantees and checks that everything is always in a good shape.
Therefore if something fails it must be due to the changes within the PR.

silvanocerza:
@capital-G 

I consider config files in json/yaml to be code. My point is that you don't always need to run the whole CI and test suite if you edit a random config file if it doesn't affect neither code nor build process. Same thing for documentation. 

What I'm arguing for is to have separate workflow to lint different types or categories of files. So we can also have a clear separation of concerns and run only what we really need for this or that PR.

The current approach prevents splitting this kind of checks and forces us to run everything every time, that's why I'm pushing for splitting this stuff also in #6375. I believe this to pretty important to simplify stuff and making it less error prone.

Issue #5993, opened 2023-03-13
Increase UDP socket receive buffer size


Spacechild1:
Some platforms, such as Windows, use a very small UDP socket receive buffer size by default. If you send many messages in batches this can easily cause packet loss - even over localhost!

Solution: set a more appropriate buffer size, e.g. 1 MB. This needs to be done both in sclang and scsynth.

dyfer:
Just to clarify, is this about just the buffer size in c++ implementation? (not the size of individual UDP packets being sent?)

Spacechild1:
It is about the UDP socket receive buffer. It can be set with the `SO_RCVBUF` socket option, see https://man7.org/linux/man-pages/man7/socket.7.html resp. https://learn.microsoft.com/en-us/windows-hardware/drivers/network/so-rcvbuf.

scztt:
+1 This would be very useful for the sclang language server, as it often has to break up large messages into a bunch of smaller chunks, which adds performance overhead.

Spacechild1:
> This would be very useful for the sclang language server

I don't know anything about LSP, but I would have assumed that it would communicate over TCP.

mtmccrea:
+1 for this enhancement! This limit has complicated unit tests repeatedly, so would be great to have some more overhead.

JordanHendersonMusic:
This could eliminate many unnecessary calls to `s.sync` in things like `loadToFloatArray` offering a sizable performance boost. `s.sync` is often used to throttle, although there is nothing sync there. Particularly useful if you're doing flucoma style many small buffer processes.

JohnC56:
I am not sure if it is the same issue, but when I was trying to analyses big datasets with Flucoma tools very often that process stopped randomly at the middle. The language was not crashing neither the server, but the debugging variables on the post window simply stopped to be posted. Sometimes `s.sync` was solving the issues, but not always... 

I tested this on various machines and OS, I got better results on high performances desktop but still having problem to complete the analysis. 

A more detailed description here:
https://discourse.flucoma.org/t/exploratory-analysis-plotxycolor-example/2410

JordanHendersonMusic:
>  Sometimes s.sync was solving the issues, but not always...

Increasing UDP buffer would push back this issue, but you could try TCP? This is probably off topic and you might get for detailed responses over at scsynth.

xunil-cloud:
[...]

Issue #5870, opened 2022-09-13
sclang: udp messages dropped when they are too fast/too many


elgiano:
## Environment

* SuperCollider version: 3.13dev
* Operating system: Arch Linux
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
s.boot;
b = { Buffer.read(s, Platform.resourceDir +/+ "/sounds/a11wlk01.wav") } ! 200

// looks like we are not receiving some b_query replies
b.collect(_.numFrames).occurrencesOf(nil).postln
```

## Expected vs. actual behavior
I would expect to get 200 buffer object with their fields updated. However, fields are not updated for some of those buffers, because some "\b_info" replies from server get lost. This is an issue, e.g. if code is relying on completion actions to proceed, because if some server replies are lost, those are not triggered either.
The issue doesn't occur when `server.protocol = \tcp`. 

I've checked with Wireshark (sniffing network traffic) and the server correctly sends all replies. Playing around with the code sclang uses to receive UDP packets, it turns out that some are not received. The reason seems to be that, under stress, processing an incoming OSC message prevents the receiving of concurrent ones.

https://github.com/supercollider/supercollider/blob/b9a57003b34ebdbb45070abfc10d90a2acf1588f/lang/LangPrimSource/SC_ComPort.cpp#L131-L132

Here `ProcessOSCPacket` can take too long (it also requires locking a mutex to call `runInterpreter`), and inverting the order of these two calls doesn't help. I know nothing about it, but I guess that until this function returns, we are not receiving new messages?

I sketched a FIFO, like scsynth has, and it solves the issue, at the cost of having one more thread to go through it: 
https://github.com/supercollider/supercollider/compare/develop...elgiano:supercollider:topic/udpqueue
Looking forward to get feedback :)

jamshark70:
Ubuntu Linux here, seemingly unreproducible.

```
-> [ Buffer(0, nil, nil, nil, /usr/local/share/SuperCollider/sounds/a11wlk01.wav), Buffer(1, nil, nil, nil, /usr/local/share/SuperCollider/sounds/a11wlk01.wav), ...etc...
0
-> 0
```

Also tried:

```
(
fork {
	var sent = IdentitySet.new;
	var recd = IdentitySet.new;
	var resp = OSCFunc({ |msg|
		recd.add(msg[2]);
	}, '/done', s.addr, argTemplate: ['/b_allocRead']);
	var n = 200;
[...]

PR #6374, opened 2024-06-27
Testsuite: Speed up CoreUGens tests

mtmccrea:
## Purpose and Motivation

In many unit tests, wait times are ad hoc, which accumulates quite a bit of time when lots of tests are run. This also points to uncertainties about how the tests are scheduled and run.

This PR tries to make the logic more strict and rational, and effectively reduces the run time from ~115 sec to ~45 sec. The “test on x” CI jobs take between 13-15 minutes for each of Mac and Linux tests. So if this approach is robust, it may be able to be applied to tests more broadly.

Improvements so far:
- add a `classvar` to have a default number of blocks to run for tests. Defaults to 10 blocks. (It’s still up to the test author to use this variable.)
- The largest improvement comes from revising the `wait` scheme
    - Removed random wait times.
    - Rely on `server.sync` to throttle looping during asynchronous testing
        - In these cases, the approach of counting the `completed` tests for the predicate of the waiting condition, and in those cases an additional check is made to ensure all tests completed.
        - An edge case appears to be when only one (or some small number of tests?) is run in a loop, the wait condition times out. The timeout time is calculated based on `testDuration x numTests x 1.5`, so when only one test runs with a very short duration (say, 10 blocks, the default) it times out because the time for resource allocation is comparable to the test run time. So, there is a minimum timeout of 2 seconds imposed. This is just a failsafe, as none of these test loops should ever time out, and an error is thrown if it does.
- `Pitch` and `ZeroCrossing` tests were rewritten to be easier to reason without magic numbers, and in turn now pass with higher accuracy and shorter run times. (were 10 sec each)
- `Out.ar` was removed from these tests so there is no sound out during tests
        - For tests that do use output (e.g. `RecordBuf` uses `.play` to write to a buffer), the `outbus` is set to `server.options.numInputBusChannels + server.options.numOutputBusChannels`, i.e. the first virtual audio bus, so no sound is written to hardware.
- `report` was set to `false` throughout. This would otherwise create trivial noise in the output of passing tests. Perhaps someone can provide some rationale for setting this to `true`.

## Types of changes

- Tests

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -5,6 +5,15 @@ TestCoreUGens.run
 */
 TestCoreUGens : UnitTest {
 
+	classvar <>defaultNumBlocksToTest = 10;
+	// The wait condition following the tests is set, by loose convension,
+	// to timeout at (1.5 * testDur * numTests), which should be more than
+	// enough to never time out when running asynchronous tests.
+	// However, this can be problematic if, e.g., there is a single test
+	// with a very short duration causing a very short timeout. So
+	// minTimeOut ensures a minimum timeout time after the last test.
+	classvar <>minTimeOut = 2;
+
 	var server;
 
 	setUp {
@@ -17,13 +26,13 @@ TestCoreUGens : UnitTest {
 	}
 
 	test_ugen_generator_equivalences {
-		var n, v;
-		var condvar = CondVar();
+		var cond = CondVar();
 		var completed = 0;
+		var n, v, tests, testDur, testsFinished;
 
 		// These pairs should generate the same shapes, so subtracting should give zero.
 		// Of course there's some rounding error due to floating-point accuracy.
-		var tests = Dictionary[
+		tests = Dictionary[
 			//////////////////////////////////////////
 			// Ramp generators:
 			"Line.ar can match LFSaw.ar" -> {Line.ar(0,1,1) - LFSaw.ar(0.5)},
@@ -287,7 +296,7 @@ TestCoreUGens : UnitTest {
 		[
 			[DelayN, BufDelayN],
 			[DelayL, BufDelayL],
-			//	[DelayC, BufDelayC] // not equivalent, fixme
+			//	[DelayC, BufDelayC] // not equivalent, FIXME
 		].do { |classes|
 			tests = tests.add(
 				"% == % [control rate]".format(classes[0], classes[1]) -> {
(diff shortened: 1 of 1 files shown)


testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -309,125 +316,143 @@ TestCoreUGens : UnitTest {
...
-				completed = completed + 1;
-				condvar.signalOne;
-			});
-			0.32.wait;
+				RecordBuf.ar(PlayBuf.ar(numchans, dataBuf, BufRateScale.ir(dataBuf), doneAction: 2), recBuf, loop:0) * 0.1;
+			}.play(server, outbus: server.options.numInputBusChannels + server.options.numOutputBusChannels); // avoid hw output: write to first virutal bus
+
 			server.sync;
+
+			fork {
```
mtmccrea, line 424: Would be good to have feedback on this: a routine is embedded within the test. I _think_ this is ok because the timing condition is waiting for the `completed` variable update, which is shared between the main routine and this forked routine.

telephon: yes, it should be ok. A `forkIfNeeded` would embed the test if it happens in  a routine already.

telephon: actually, I was wrong, better *not* use a forkIfNeeded.

telephon:
```

// works as expected
(
fork {

	var con = CondVar.new;
	fork {
		2.wait;
		con.signalOne;
	};
	con.wait;
	"hello".postln
	
}
)


// doesn't work correctly
(
fork {

	var con = CondVar.new;
	forkIfNeeded {
		2.wait;
		con.signalOne;
	};
	con.wait;
	"hello".postln
	
}
)
```

testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -309,125 +316,143 @@ TestCoreUGens : UnitTest {
...
+				(bufDur * 1.05).wait; // wait for the buffer to fill up
+				recBuf.loadToFloatArray(action: { |renderedData|
```
telephon, line 426: Now there is a method `asBuffer` in function. This may simplify the logic a little bit here.

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: Yes, you can think of the routine being "flattened" to remove the inner fork:
```supercollider
// doesn't work correctly
(
fork {

	var con = CondVar.new;
	forkIfNeeded {
		"waiting".postln;
		2.wait;
		"waited, signalling".postln;
		con.signalOne;
	};
	"waiting outside".postln;
	con.wait;
	"hello".postln
}
)
```
So it signals the condition before the thread is made to wait on the condition.

telephon: yes, we could mention this in `forkIfNeeded` sometime.

telephon (approved):

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: In this case we're testing the data written to a preallocated buffer via `RecordBuf`,  as opposed to testing the output of the "synth" (function). Unless you're referring to a different function?

mtmccrea:
Just did some minor cleanup of comments and variables that were replaced.

testsuite/classlibrary/TestCoreUGens.sc
telephon: yes, just in that case, of the output of a synth.

telephon:
Just a guess: maybe this needs to be higher for the TestUGen_RTAlloc ugens?

```
server.options.memSize = 1024;
```

dyfer:
> Just a guess: maybe this needs to be higher for the TestUGen_RTAlloc ugens?

Hm, that might be it: https://github.com/supercollider/supercollider/actions/runs/9902226869/job/27358145553?pr=6374#step:6:14560

telephon:
Exactly, that's what I mean.

mtmccrea:
> > Just a guess: maybe this needs to be higher for the TestUGen_RTAlloc ugens?
> 
> Hm, that might be it: [`LocalBuf_allocBuffer: alloc failed, increase server's RT memory (e.g. via ServerOptions)`](https://github.com/supercollider/supercollider/actions/runs/9902226869/job/27358145553?pr=6374#step:6:14560)

https://scsynth.org/chat/c/-/45/3280

capital-G:
@mtmccrea can you rebase this against develop so we can see if we get a successful CI? RTAlloc tests are now temporarily disabled in develop b/c they create problems.

capital-G (approved):
Thanks for rebasing! The PR is really appreciated, anything that makes the tests more deterministic and stable is a great help!

The changes look good from my side, as @telephon has some unresolved comments in this PR, he should merge it.

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: @telephon, to clarify whether this comment is resolved, I don't see that `asBuffer` is useful in this case, since there is no synth output being generated. The synth under test only uses `RecordBuf`, not `Out` and it's the contents of the buffer being written to that is tested against `data`.

I see how 
```
// random data for test
data = { 1.0.rand }.dup((server.sampleRate * bufDur).round * numchans);
// load data to server
dataBuf = Buffer.loadCollection(server, data, numchans);
```
could make use of `asBuffer` to be more concise, but the `data` array is used later
```
this.assertArrayFloatEquals(renderedData - data, ///
```
so if a buffer were loaded directly with `asBuffer` it would just need to be unpacked later for the assert test.

mtmccrea:
Rebased again after conflict with #6441, which was actually redundant with this PR.

mtmccrea:
No problem, thanks for reviewing!

capital-G:
I'll merge it - hopefully it improves the state of the tests that #6441 caused 😶‍🌫️

testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -309,124 +318,143 @@ TestCoreUGens : UnitTest {
...
+
 		tests.keysValuesDo{|name, func|
-			func.loadToFloatArray(0.01, server, { |data|
+			func.loadToFloatArray(testDur, server, { |data|
 				this.assertArrayFloatEquals(data, 0, name.quote, report: true);
 				completed = completed + 1;
-				condvar.signalOne;
+				cond.signalOne;
 			});
+			server.sync;
```
JordanHendersonMusic, line 391: @mtmccrea I was just wondering (after #6441) if  you could explain why this `server.sync` is needed? I can't seem to why this passes, but the other approach fails.

JordanHendersonMusic: Ah I think I've figured this out. If you switch to TCP and remove the syncs, you don't need this and performance increases by a factor of 2. I suppose this means the udp messages are getting dropped.

I've also found you get a significant performance increase (circa 5x) in a tight loop by only syncing every 6 iterations (UDP)
```
if (n % 6 == 0) { s.sync };
```

What do you think about switching all these tests over to use TCP and removing the s.sync? It passes on my machine.

---

Sorry for not commenting on this when it was still open!

testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -309,124 +318,143 @@ TestCoreUGens : UnitTest {
...
 		};
 
-		condvar.waitFor(1, { completed == tests.size });
+		testsFinished = cond.waitFor((1.5 * tests.size * testDur).max(minTimeOut), { completed == tests.size });
```
JordanHendersonMusic, line 394: I just had a question about your intent here. The time here is the time between each `condvar.signalOne`, not the total time of waiting (documentation seems lacking here), is it correct to multiply by the number of tests in this case  — I mean it works so it doesn't really matter.

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: Seems like a great thing to propose in a new PR, so others can weigh in if they know more useful info about UDP vs. TCP behavior. I'm surprised UDP gets dropped on an internal messaging. I'm not sure where the point of failure is vs. TCP... does TCP  just have a larger message queue?

One of the goals of this PR was to model a fairly principled approach to timing tests, which might then find its way into more areas of the test suite. So if we can pin down in some degree of detail why TCP allows higher throughput, we could document it as best practices.

Looks promising!

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: Thanks for looking at this in detail. The way to think about this is the timeout is set to a duration beyond the completion time of _all_ tests. It's a bit confusing because this condition is signaled after each test, but it's actually waiting for the predicate `completed == tests.size` to be `true`. From the [docs](https://docs.supercollider.online/Classes/CondVar.html#-waitFor):

>If predicate is not `nil`, this method returns immediately if `predicate.value` is `true`, and otherwise blocks either until the timeout expires, or until the thread is woken by `signalOne` or `signalAll` and `predicate.value` is `true`.

The mechanics of `CondVar` take some getting used to (I have to re-read docs every time I use it).

JordanHendersonMusic: Thanks for this! I actually wrote some tests before I wrote that to try and make sure what I was saying was correct... I guess I messed up somewhere as I can't repeat it now! Thanks again!

testsuite/classlibrary/TestCoreUGens.sc
mtmccrea: @dyfer linked to @Spacechild1's helpful post [here](https://github.com/supercollider/supercollider/issues/5870#issuecomment-1310965786).

So, if we're running into UDP throughput limits, one solution would be to increase the the socket option's receive buffer size. (not sure if this is compile-time or runtime config... could we just increase this by default in SC??)

In the meantime, `s.sync` seems like the standard option for throttling server-side message backlog.

Spacechild1: > So, if we're running into UDP throughput limits, one solution would be to increase the the socket option's receive buffer size. 

See https://github.com/supercollider/supercollider/issues/5993 :)

However, I agree with @JordanHendersonMusic that if you want to have reliable communication, TCP is the way to go!

>  So if we can pin down in some degree of detail why TCP allows higher throughput, we could document it as best practices.

With UDP you need to artificially throttle the traffice with `s.sync` or `wait` to prevent socket receive buffer overruns. With TCP this is not necessary, which leads to the curious situation that in some situations you get higher throughput, even though TCP is the slower protocol.

PR #6375, opened 2024-06-27
Isolate Github release process in a separate workflow

silvanocerza:
## Purpose and Motivation

This PR moves the release process in a separate workflow from `actions.yml`.

This enhances the process so it's clear that only pushing a `Version-` prefixed tag will trigger the release process. With the current solution any tag would have a triggered a release as the check `startsWith(github.ref, 'refs/tags/')` doesn't verify the tag format.

Also since `actions.yml` is trigger on push, PRs and on a schedule it could have been possible to trigger a release multiple times by mistake. That's a really slim chance but it could have still have happened. 

Another problem could have happened if a tag is pushed long after the push that triggers `actions.yml`, at that point the release would have been skipped since the version from the `version-getter` would be simply the commit hash.

This change also made it possible to simplify a bit the logic in `actions.yml` as only the commit hash is now necessary to create the test binaries.

I decided not to run tests in the `release.yml` as I would expect whoever handles the release to tag only commits with passing checks.

--- 

I have not moved the release on S3 as I'm unsure what are the expectations of that job. I left a comment about this in [#6340](https://github.com/supercollider/supercollider/issues/6340#issuecomment-2194386306), let's discuss there.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,67 @@
+name: Release on Github
+on:
+  push:
+    tags:
+      - Version-[0-9]+.[0-9]+.[0-9]+*
+
+jobs:
+  version-getter:
+    runs-on: ubuntu-22.04
+    outputs:
+      sc-version: ${{ steps.set-version.outputs.version }}
+
+    steps:
+      - name: set version string for artifacts
+        id: set-version
+        run: |
+          SC_VERSION="${${{ github.ref_name }}##Version-}"
+          echo "version=$SC_VERSION" >> "$GITHUB_OUTPUT"
+
+  build-linux:
+    needs: [version-getter]
+    uses: ./.github/workflows/build_linux.yml
+    with:
+      sc-version: ${{ needs.version-getter.outputs.sc-version }}
+    secrets: inherit
+
+  build-macos:
+    needs: [version-getter]
+    uses: ./.github/workflows/build_macos.yml
+    with:
+      sc-version: ${{ needs.version-getter.outputs.sc-version }}
+    secrets: inherit
+
+  build-windows:
+    needs: [version-getter]
+    uses: ./.github/workflows/build_windows.yml
+    with:
+      sc-version: ${{ needs.version-getter.outputs.sc-version }}
+    secrets: inherit
+
+  github-release:
+    name: "Create GitHub release"
+    needs: [version-getter, build-linux, build-macos, build-windows]
+    runs-on: ubuntu-latest
+    env:
+      INSTALL_PATH: ${{ github.workspace }}/Install
+      ARTIFACT_FILE_PREFIX: SuperCollider-${{ needs.version-getter.outputs.sc-version }}
+    steps:
+      - name: Download release artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: ${{ env.INSTALL_PATH }} # no "name" paramter - download all artifacts
+
+      - name: Create and upload release files
+        uses: softprops/action-gh-release@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          files: |
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-macOS-x64.dmg/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-macOS-universal.dmg/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-macOS-arm64.dmg/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-macOS-x64-legacy.dmg/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win32-installer/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64-installer/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win32/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64/*
+          draft: true
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -38,63 +38,46 @@ jobs:
         env:
           FULL_CHECK: 1
 
-  version-getter:
-    runs-on: ubuntu-22.04
-    outputs:
-      sc-version: ${{ steps.set-version.outputs.version }}
-
-    steps:
-      - name: set version string for artifacts
-        id: set-version
-        run: |
-          if [[ "$GITHUB_REF" == refs/tags/* ]]; then
-            FULL_TAG=${GITHUB_REF#refs/tags/}
-            SC_VERSION=${FULL_TAG##Version-}
-          else
-            SC_VERSION=$GITHUB_SHA
-          fi
-          echo "version=$SC_VERSION" >> $GITHUB_OUTPUT
-
   build-linux:
-    needs: [lint, version-getter]
+    needs: [lint]
     uses: ./.github/workflows/build_linux.yml
     with:
-      sc-version: ${{ needs.version-getter.outputs.sc-version }}
+      sc-version: ${{ github.sha }}
     secrets: inherit
 
   build-macos:
-    needs: [lint, version-getter]
+    needs: [lint]
     uses: ./.github/workflows/build_macos.yml
     with:
-      sc-version: ${{ needs.version-getter.outputs.sc-version }}
+      sc-version: ${{ github.sha }}
     secrets: inherit
 
   build-windows:
-    needs: [lint, version-getter]
+    needs: [lint]
     uses: ./.github/workflows/build_windows.yml
     with:
-      sc-version: ${{ needs.version-getter.outputs.sc-version }}
+      sc-version: ${{ github.sha }}
     secrets: inherit
 
   test-linux:
-    needs: [version-getter, build-linux]
+    needs: [build-linux]
     uses: ./.github/workflows/test_linux.yml
     with:
-      artifact-file: "SuperCollider-${{ needs.version-getter.outputs.sc-version }}-linux-jammy-gcc12"
+      artifact-file: SuperCollider-${{ github.sha }}-linux-jammy-gcc12
     secrets: inherit
 
   test-macos:
-    needs: [version-getter, build-macos]
+    needs: [build-macos]
     uses: ./.github/workflows/test_macos.yml
     with:
-      artifact-file: "SuperCollider-${{ needs.version-getter.outputs.sc-version }}-macOS-universal.dmg"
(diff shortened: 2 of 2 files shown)


dyfer:
Thanks for this! 

Things can certainly be changed, but I want to make sure we know why things are the way they are. Also, the current release workflow has been working for 3+ (?) releases of SC (and each release means there are a few pre-releases), so it seems to have been tested well. "If it ain't broke..." :)
 
> This enhances the process so it's clear that only pushing a `Version-` prefixed tag will trigger the release process. With the current solution any tag would have a triggered a release as the check `startsWith(github.ref, 'refs/tags/')` doesn't verify the tag format.

I _think_ the ability to createa release from non `Version-` tags is intentional, in order to create test releases.

> Also since `actions.yml` is trigger on push, PRs and on a schedule it could have been possible to trigger a release multiple times by mistake. That's a really slim chance but it could have still have happened.

But the release workflow is only triggered when pushing the tag, how would multiple triggers be possible?

> Another problem could have happened if a tag is pushed long after the push that triggers `actions.yml`, at that point the release would have been skipped since the version from the `version-getter` would be simply the commit hash.

I don't see how that would be possible. If the tag is pushed, the version will be based on the tag. What am I missing?
 
> I decided not to run tests in the `release.yml` as I would expect whoever handles the release to tag only commits with passing checks.

OTOH, if for some reason an issue has slipped past the previous stages, running tests on release could catch that.

Logically, I think the release should _not_ be triggered unless _all the tests pass_... However, our testing suite is not 100% reliable (yet?) so this is not the right approach.

> I have not moved the release on S3 as I'm unsure what are the expectations of that job. I left a comment about this in [#6340](https://github.com/supercollider/supercollider/issues/6340#issuecomment-2194386306), let's discuss there.

I left comments there as well.

silvanocerza:
First of all I realised while writing this that I didn't commit the new release workflow. 🤦
I pushed it now. 

> Things can certainly be changed, but I want to make sure we know why things are the way they are. Also, the current release workflow has been working for 3+ (?) releases of SC (and each release means there are a few pre-releases), so it seems to have been tested well. "If it ain't broke..." :)

I would argue that if the release process is ambiguous, unclear and part of the testing process is broken. :)

> I _think_ the ability to createa release from non `Version-` tags is intentional, in order to create test releases.

If with test releases you mean release candidates this PR handles that. The below snippet means that the `release.yml` workflow will be triggered only by tag pushes that match that glob. e.g `Version-4.0.0-rc1` will match.

```
on:
  push:
    tags:
      - Version-[0-9]+.[0-9]+.[0-9]+*
```

If you mean test builds from PRs I'd say it's out of scope and should not be part of the release process. Probably something we should handle in the S3 job, or whatever that job will become.

> But the release workflow is only triggered when pushing the tag, how would multiple triggers be possible?
>
> I don't see how that would be possible. If the tag is pushed, the version will be based on the tag. What am I missing?

Nope. Since the release job is part of the `actions.yml` workflow it will be triggered when a commit or a PR that doesn't edit any of the ignore paths, or on a weekly schedule. You can check the `on` field of the `actions.yml` to verify that.

This bring some risks as pushing a tag together with a commit or the HEAD of a PR will trigger a release. Same thing if the latest commit on `develop` is tagged whenever the schedule runs. 

Another risk is that you push a commit that should be release but forget to tag it. Pushing the tag at a later time won't trigger the release in any way. You can see this behaviour in action [here](https://github.com/silvanocerza/nodenode/actions/runs/9711069466). It's a workflow run that I trigger a repository of mine that I use to experiment with Github Actions. I mocked the current flow of the `actions.yml` workflow. As you can see the `deploy release` job didn't run, though the commit that triggers that run [is tagged](https://github.com/silvanocerza/nodenode/commit/b44bcd987a414eb4fc7b034b4064003ba0d688ab). Since I tagged it after pushing the commit the release didn't work.

The chances of all the above happening are certainly slim, but they exist. If you know how to tackle those issue it's not a huge issue, it's just annoying. The new workflow I introduce prevents all the above and correctly creates a release only when a version commit is pushed, and no other times. 

> OTOH, if for some reason an issue has slipped past the previous stages, running tests on release could catch that.
> 
> Logically, I think the release should _not_ be triggered unless _all the tests pass_... However, our testing suite is not 100% reliable (yet?) so this is not the right approach.

That's currently not true though, the release doesn't depend on successful tests but only on successfull build. Not even all the builds actually. [See the `needs` clause](https://github.com/supercollider/supercollider/blob/9409b8fb4588a28730a0c10b576d9c2ea3524c32/.github/workflows/actions.yml#L198). 

I acted on the assumption that this was the expected behaviour. If we want to run tests right before the actual release that's fine for me.

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+    steps:
+      - name: Download release artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: ${{ env.INSTALL_PATH }} # no "name" paramter - download all artifacts
+
+      - name: Create and upload release files
+        uses: softprops/action-gh-release@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
silvanocerza, line 57: This is the new of setting the token as using the `GITHUB_TOKEN` env var is deprecated, see [the official docs](https://github.com/softprops/action-gh-release?tab=readme-ov-file#environment-variables).

dyfer:
> > "If it ain't broke..." :)
> 
> I would argue that if the release process is ambiguous, unclear and part of the testing process is broken. :)

I get the ambiguous part. The mechanics of it are not ideal, maybe. 
HOWEVER what I'm saying is that it worked _correctly_, as far as I remember, for a number of releases _in the current form_ (well, before the most recent splitting of the workflow files, but it seems that the release part was not touched).

Please note that testing these changes is cumbersome - a test release needs to be created. Also see below about "test releases".

> > I _think_ the ability to createa release from non `Version-` tags is intentional, in order to create test releases.
> 
> If with test releases you mean release candidates this PR handles that. 
> If you mean test builds from PRs I'd say it's out of scope and should not be part of the release process. 

Neither. I mean - testing the exact type of changes you are proposing, changes to the GHA release process itself.
If we push a `Version-` tag to test this, then we commit to creating a release, for the downstream clients. 
I understand we could make limited testing on another repo, but it's not the same and wouldn't test the S3 part. That might be acceptable though? I'm not sure.

> > But the release workflow is only triggered when pushing the tag, how would multiple triggers be possible?
> > I don't see how that would be possible. If the tag is pushed, the version will be based on the tag. What am I missing?
> 
> Nope. Since the release job is part of the `actions.yml` workflow it will be triggered when a commit or a PR that doesn't edit any of the ignore paths, or on a weekly schedule. You can check the `on` field of the `actions.yml` to verify that.

But the release only runs if it's triggered from the [tag](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/actions.yml#L197) isn't it?

> This bring some risks as pushing a tag together with a commit or the HEAD of a PR will trigger a release. 

How would that happen outside of tagging a commit for a release? I don't think I follow.

> Same thing if the latest commit on `develop` is tagged whenever the schedule runs.

The schedule runs only on the default (`develop`) branch AFAIU, I don't think the ref would ever be the tag in that case. And the release only runs if the ref is a tag.

> Another risk is that you push a commit that should be release but forget to tag it. Pushing the tag at a later time won't trigger the release in any way. You can see this behaviour in action [here](https://github.com/silvanocerza/nodenode/actions/runs/9711069466). 

Interesting. In the past I would create a release (using the GH web gui) _after_ pushing the commit and it seemed to have worked. How did you create the tag? How does one tag a commit without pushing that tag?

> > OTOH, if for some reason an issue has slipped past the previous stages, running tests on release could catch that.
> > Logically, I think the release should _not_ be triggered unless _all the tests pass_... However, our testing suite is not 100% reliable (yet?) so this is not the right approach.
> 
> That's currently not true though, the release doesn't depend on successful tests but only on successfull build. 

I didn't say that it is true currently, only that maybe it _should_ be. The reason it's not is mainly because our testing is not reliable, IIRC.

> Not even all the builds actually. [See the `needs` clause](https://github.com/supercollider/supercollider/blob/9409b8fb4588a28730a0c10b576d9c2ea3524c32/.github/workflows/actions.yml#L198).

Well, we only trigger deployment on the distributable release builds (macOS and Windows). If we made e.g. an AppImage for Linux, then it would make sense to deploy Linux build as well.

silvanocerza:
@dyfer 

> I get the ambiguous part. The mechanics of it are not ideal, maybe. HOWEVER what I'm saying is that it worked _correctly_, as far as I remember, for a number of releases _in the current form_ (well, before the most recent splitting of the workflow files, but it seems that the release part was not touched).
>
> Please note that testing these changes is cumbersome - a test release needs to be created. Also see below about "test releases".

It worked correctly but the logic is convoluted and hidden. There's also no separation between testing and release processes, separation of concerns is important in CI as in code.

As much as you can't reason about code without splitting its context into smaller pieces you can't do the same with CI stuff.

It must be treated as code.

> Neither. I mean - testing the exact type of changes you are proposing, changes to the GHA release process itself. If we push a `Version-` tag to test this, then we commit to creating a release, for the downstream clients. I understand we could make limited testing on another repo, but it's not the same and wouldn't test the S3 part. That might be acceptable though? I'm not sure.

Testing this is trivial and can be done either by triggering a prerelease or no release at all by commenting out the actual release step.

> But the release only runs if it's triggered from the [tag](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/actions.yml#L197) isn't it?

Yes, but any tag will do.

Also this is quite hidden, you can't know that pushing a tag will trigger a release just by looking at the `on` field of the `actions.yml` workflow.

You need to know the full context and not miss the single `if` that decides whether to release or not.

> How would that happen outside of tagging a commit for a release? I don't think I follow.

If anyone that works by creating and pushing branches directly to this repo executes the commands below it will trigger a release.

```
git checkout develop
git checkout -b my-branch
git tag my-tag
git push -u origin my-branch && git push origin my-tag
```

This happens cause the Github release job runs if `github.ref` is a tag. Whenever a push triggers a workflow `github.ref` is either the branch or the tag that was pushed.

> The schedule runs only on the default (`develop`) branch AFAIU, I don't think the ref would ever be the tag in that case. And the release only runs if the ref is a tag.

Yeah, you're right. My bad. [The documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) even says so.

> Interesting. In the past I would create a release (using the GH web gui) _after_ pushing the commit and it seemed to have worked. How did you create the tag? How does one tag a commit without pushing that tag?

I'm wrong on this one too. I forgot that pushing a tag actually counts as `push` trigger so the `actions.yml` workflow is run anyway if there are no filters as said above.

It's not really a good practice though. `actions.yml` should be run for pushes to `develop` and `main` only, other than PRs and on a schedule.

Hiding the release step behind a single `if` is bad too. Again separation must be clear.

> I didn't say that it is true currently, only that maybe it _should_ be. The reason it's not is mainly because our testing is not reliable, IIRC.

I was just pointing out the current situation, no worries. If tests are not reliable that's another problem in any case. I still would argue that we don't run tests in the release process after a build, ideally we should tag only commits with tests that already run sucessfully.

> Well, we only trigger deployment on the distributable release builds (macOS and Windows). If we made e.g. an AppImage for Linux, then it would make sense to deploy Linux build as well.

Ah, makes sense. We should completely omit the Linux build from the process then.

Issue #6341, opened 2024-06-07
Git: *checkForGit returns true when xcrun: error is returned in terminal


joslloand:
## Environment

* SuperCollider version: [unknown - reporting for a 3rd party]
* Operating system: macOS Sonoma 14.5 (23F79)
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
Git.checkForGit
```
returns:
```supercollider
-> true
```

Where
```supercollider
(
Pipe.callSync("git", { |res|
    "True: git".postln;
    res.postln
}, {
    "False: git".postln;
    Git.checkForGit.postln
})
)
```
returns:
```supercollider
False: git
true
-> nil
```

## Expected vs. actual behavior

I would expect that if `Pipe: *callSync` returns `false`, `Git: *checkForGit` would also return `false` to indicate a missing or non-functional git installation.

From the terminal, `/usr/bin/git` returns:

```
xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun
```

So, for the user's installation, an error has been returned in the terminal.

However, within SuperCollider, we're _not receiving an indication of an error_, and are instead lead to believe that a functioning git installation is available on the system.

## User working solution

Update the xcode installed version of git, as [documented here](https://apple.stackexchange.com/questions/254380/why-am-i-getting-an-invalid-active-developer-path-when-attempting-to-use-git-a).

NOTE: if git is installed in another location, e.g., `/usr/local/bin/git`, `String: -setenv` may be used to update the `PATH` environment variable. E.g.:

```supercollider
"PATH".setenv("/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin/")
```

## Possibly related issues

- #3550
- #2196

## Related forum discussions 

- https://scsynth.org/t/unixcmd-shellcmd/5435
- https://scsynth.org/t/solved-git-checkforgit-windows/5644
- https://scsynth.org/t/no-output-with-unixcmdgetstdoutlines-on-bigsur/6128
- https://scsynth.org/t/three-problems-or-characteristics-of-unix-related-methods-and-class/7644

capital-G:
I just had a look at this and the Pipe interface also seems at least counter-intuitive to me - and it is also not documented.
Pipe is not using the default shell (`/bin/zsh` for macOS) but uses `/bin/sh -c`, so lets compare.

```shell
[...]

Issue #4751, opened 2020-01-28
Quarks.prFetchDirectory should do a shallow checkout


jamshark70:
## Environment

* SuperCollider version: any
* Operating system: any

## Expected vs. actual behavior

Expected: `Quarks.fetchDirectory` leaves the user with three not-very-large text files. This should take a couple of seconds to download.

Actual: If the quarks directory has not been downloaded before, it takes longer than you expect, up to a few minutes. The total size of the text files is less than 30k. There is absolutely no way it's reasonable to expect the user to sit there waiting for a measly 30k of text.

The reason is that all of the quarks code used to be stored under supercollider-quarks/quarks.

When `prFetchDirectory` does a full clone, it's downloading every previous version of every quark, and then also downloading a very large commit where these were deleted. 13.11 MB -- about 437 times as much data as will be used (!). Here in China, that's coming down at about 44 K/sec so, 13.11 MB * 1024 KB/MB / 44 KB/sec = 305 seconds = 5 minutes. 5... minutes :open_mouth: :open_mouth: :open_mouth: ... for 30 K? :man_facepalming:

Users don't need this.

I suggest adding `--depth 50` to the git command. With this, git downloads about 60K, in just a few seconds.

brianlheim:
could also be resolved by making a different branch/history on that repo and cloning from that instead. shallow checkouts seem like risky business and i personally don't want to know what can of worms that opens. would be pleased to know that it is safe if someone has experience with it.

jamshark70:
> could also be resolved by making a different branch/history on that repo and cloning from that instead

I'd be ok with that.

> shallow checkouts seem like risky business and i personally don't want to know what can of worms that opens. would be pleased to know that it is safe if someone has experience with it.

For this particular repository, I think the risk is basically 0. We're doing two things routinely with that repository:

- Users check it out so they can install quarks by name. That's read-only. A full history is irrelevant to the vast majority of checkouts.

- Quark authors submit a PR based on the tip of the history, to edit the directory. There should be no need to revert back to an earlier state for this -- basically, all PRs on this repository should be fast-forward-able (except if multiple authors have submitted PRs simultaneously, but that should be going back only a couple of commits).

Anything beyond that, we can assume that you have to know what you're doing (and you'll know how to check out a full history manually).

But yes, cloning another branch is probably easier/better.

jamshark70:
> could also be resolved by making a different branch/history on that repo and cloning from that instead

[...]

Issue #4359, opened 2019-03-15
[win] Quarks.update a/ does not update anything and b/ hangs the interpreter


jamshark70:
## Environment

* SuperCollider version: 3.10.2
* Operating system: Win10

## Steps to reproduce

```supercollider
Quarks.update("someQuarkThatIAlreadyCheckedOut");
```

## Expected vs. actual behavior

Expected: SC should issue `git pull` and `git checkout` commands against the repository, and allow you to recompile the class lib.

Actual: SC reports "Quark 'ddwOSCSyncClocks' updated to version: 0.1 tag: nil refspec: 857a0c8272a97e8610f3b1c4bda73db2bc38b0a1" but this will be an old refspec, not the most current. Then, if you try to recompile the class library, nothing happens. The only way to get control back is to go to Task Manager and force-kill sclang. (Also, at this time, trying to run something like `2+2` also posts nothing -- so the interpreter is totally locked up.)

Diagnostics:

1. `"dir".unixCmd` works fine (interpreter doesn't lock up).

2. `Pipe.callSync("dir", { |... args| args.debug("ok") }, { |... args| args.debug("failed") });` also OK.

3. `["cd", thisProcess.platform.formatPathForCmdLine("C:\\Program Files"), "&&", "dir"].join(" ").unixCmd` (cribbing from `Git:git`) is OK.

4. `Pipe.callSync(["cd", thisProcess.platform.formatPathForCmdLine("C:\\Program Files"), "&&", "dir"].join(" "), { |... args| args.debug("ok") }, { |... args| args.debug("failed") });` is OK.

5. `Pipe.callSync(["cd", thisProcess.platform.formatPathForCmdLine(Quarks.at("ddwOSCSyncClocks").localPath), "&&", "git pull"].join(" "), { |... args| args.debug("ok") }, { |... args| args.debug("failed") });` "says" OK but `Updating 857a0c8..695293d` (old refspec). Interpreter is locked up now.

So something about the git command messes up the interpreter's state.

jamshark70:
Upon further investigation... a local change was blocking `git pull`. We report this to the user as success (??!!) and of course locking up the interpreter is not quite what we want.

Quarks error reporting is... let's say... a problem.

dyfer:
Maybe not related to the this problem, but possibly helpful when trying to find a solution - getting a status from `git` is also a [problem when trying to install an incorrect url](https://github.com/supercollider/supercollider/issues/4283).

[...]

PR #6380, opened 2024-06-30
Project, SCLang: Use libgit2 to implement sc langs quarks

JordanHendersonMusic:
# This is a draft.

## Purpose and Motivation

You can't always depend on having git install on a system. For users of which supercollider is their first programming language, using the terminal and other steps needed to install git can be a large barrier  to using quarks.

This approach uses [libgit2](https://github.com/libgit2/libgit2), a well respected implementation of git — github uses it, amongst many others. Their [API is stable and well documented](https://libgit2.org/libgit2/#HEAD). It is GPLv2 with a special Linking Exception.

## Types of changes

- New dependency, git sub modul
- Will replace the existing Git class, this PR is a draft, but this probably should be a breaking change unless we really want two 'git' classes.

Right now, only clone is implemented, see the test file. 
## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


JordanHendersonMusic:
I can't figure out how to get cmake to link libgit2 for windows and mac. I'm sure there is some simple way to do this.

dyfer:
Thanks for working on this.
I personally think that we should have a way of using quarks without OS-installed git, particularly on macOS and Windows.

Just so you know, concerns that were brought up in the past about adding git to sclang were:
- security - maintaining the updated library and handling any security concerns
- sclang structure - some devs feel sclang is too bloated already and were weary of adding more primitives to sclang
  - a plugin system for sclang could address this, but that hasn't been worked on
- I think it's an open question whether supporting the OS-provided git is supported. I think Linux installs might favor that; however, we do have a number of issues with our interfacing with git (list below), so I'm not sure if maintaining/fixing this is viable. Hopefully using a built-in gitlib could address some of these(?)

<details> 

https://github.com/supercollider/supercollider/issues/6341
https://github.com/supercollider/supercollider/issues/5754
https://github.com/supercollider/supercollider/issues/4751
https://github.com/supercollider/supercollider/issues/4359
https://github.com/supercollider/supercollider/issues/3554
https://github.com/supercollider/supercollider/issues/3459
https://github.com/supercollider/supercollider/issues/2196
https://github.com/supercollider/supercollider/issues/1462
</details>

JordanHendersonMusic:
> security - maintaining the updated library and handling any security concerns

Updating it should be easy because its a git submodule. I guess its only as secure as libgit, but its used by *many* big companies.


> sclang structure - some devs feel sclang is too bloated already and were weary of adding more primitives to sclang
a plugin system for sclang could address this, but that hasn't been worked on

I've actually found most of the primitives have a lot of duplication, while working on keyword args I removed over 1000 lines with just a couple of abstractions. Most of these primitives will be very simple, as the git clone example in this project demonstrates.

> I think it's an open question whether to still support the OS-provided git. Linux installs might favor that; however, we do have a number of issues with our interfacing with git (list below), so I'm not sure if maintaining/fixing this is viable. Hopefully using a built-in gitlib could address some of these issues(?)

It would mean changing the `Git` classes public API as right now it has you build arrays of arguments that are passed to the command line, this approach would be more like `Git().clone("https...", "/home/jordan/.scQuarks")`. Supporting the old and the new seems unnecessary and just bound to cause issues.

---

That being said... I've given up trying to set up CMake correctly.

I can get it to see the headers I've included rather than the system ones.
I can get it to build a shared library (cannot get it to build a static one), but the linker refuses to link against it, instead using my system libraries, and I suspect the CI's, if not, then I don't understand why linux can find it but mac and windows can't because I've gotten the CI to work on linux on one point but not the rest.

Don't suppose there are any CMake magicians out there who wouldn't mind taking a look? It ***should*** be easy...

smoge:
> * security - maintaining the updated library and handling any security concerns

Using Git as a package manager (git or libgit) does come with several security concerns. The commit hashes can give a false impression they have something with security.

- git repositories can be cloned and modified without any form of verification
- Unlike some package managers that require cryptographic signatures, Git does not inherently enforce signed commits or tags,

Besides, git will not help with most package managing we need to improve:

- Without proper version control (and signatures in a way), updates from remote repositories can introduce breaking changes/vulnerabilities/go-knows-what


Quick ideas for improvements/migration:
- Cryptographic Signatures
- Dependency Pinning: Explicitly pin dependencies to specific commits or tags (already implemented, but buggy in some respects, and not embraced by community)

Extra goodies:
- Automated Scanning: Use tools to scan for vulnerabilities and sensitive information

dyfer:
@smoge yes, these are all valid points, and something to consider when we discuss the future of package management in SC.

My point with security was more about the possible (but not likely) security vulnerabilities with everything that accesses the network, e.g. in the context of attacks like arbitrary code execution through unpatched web mechanisms. BTW we ship a whole web browser in SC (WebView) and that also is a security concern. I'm fine to treat adding libgit as a negligible risk, especially that - as you pointed out - the quarks system does not provide much protection against malicious use atm.

smoge:
> security vulnerabilities with everything that accesses the network, e.g. in the context of attacks like arbitrary code execution through unpatched web mechanisms. BTW we ship a whole web browser in SC (WebView) and that also is a security concern. I'm fine to treat adding libgit as a negligible risk,

Yes, I think the risk is already there. The quark system already has access to git.

The help browser is a huge risk if it access the internet. Wouldn't be better to just block internet access in the help browser?

dyfer:
> Wouldn't it be better to block internet access in the help browser?

SC provides the `WebView` class, running using the same engine as the help browser - web browsing is (unfortunately) an existing feature of SC and as such it probably has been used in the wild. I think limiting network access _could_ be considered, but probably as a runtime switch, and not a hard block; some 3d party help documents include links to web resources and these open in the help browser currently, though I imagine that they could be just redirected to the OS browser instead.

telephon:
> SC provides the `WebView` class, running using the same engine as the help browser - web browsing is (unfortunately) an existing feature of SC and as such it probably has been used in the wild.

I think we can live without this web browsing. It may be used, but I don't think it is a good idea to carry around a web browser in the ide. Better open the browser from links.

If there are good reasons, I'd reconsider, but so far I find it annoying that I open web pages accidentally with a non-fully-featured web browser (scide).

JordanHendersonMusic:
One nice thing about web browsing is that you can display the wiki information from here and other places - including the GitHub page of quark files. This means we can update information independent of the release cycle.

I'd be sad to see this feature go. I don't think it's been used enough, but offers many possibilities and since we are all in on Qt, there's little extra burden.

smoge:
> One nice thing about web browsing is that you can display the wiki information from here and other places - including the GitHub page of quark files. This means we can update information independent of the release cycle.
> 
> I'd be sad to see this feature go.


Even browsers under development, they take care of security. they know the risks.

The bare minimum would use a DNF block list, with known malicious addresses.

smoge:
Maybe there is a way to run it in a sandboxed environment with limited permissions, much more restricted than the rest of the IDE itself. Not sure it that would imply another process/application.

lang/CMakeLists.txt
```diff
@@ -244,6 +248,23 @@ endif()
 target_link_libraries(libsclang boost_thread_lib boost_system_lib boost_regex_lib boost_filesystem_lib)
 target_include_directories(libsclang PUBLIC ${boost_include_dirs})
 
+
+#target_include_directories(libsclang PUBLIC ${LIBGIT2_INCLUDES} ${LIBGIT2_DEPENDENCY_INCLUDES})
+#target_include_directories(libsclang SYSTEM PRIVATE ${LIBGIT2_SYSTEM_INCLUDES})
+
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(LIBGIT2 REQUIRED IMPORTED_TARGET libgit2)
```
dyfer, line 256: I don't think we should be using PkgConfig when building the library ourselves. This might be the reasons of your issues, but I haven't looked closely.

We already have a similar situation in SC codebase - with `portaudio` and `yamlcpp` libraries. They can be built from source or linked from system. You may want to check how portaudio is linked when `SYSTEM_PORTAUDIO` option is off. This configuration is used on the Windows builds by default.

JordanHendersonMusic:
Closing because I don't have the time to figure out what CMake is doing. If someone else wants to do that part, I'll happily help write some of the primtiives.

dyfer:
I am hoping to take a look at the cmake part of this! Just FYI.

capital-G:
Just FYI - I am working on a package manager which does **not** rely on git anymore and is an external binary - so maybe the priority isn't too high anymore to get git primitives into sclang?

dyfer:
Hm, I know adding primitives is debatable. I guess - if this works, it could be a nice addition for 3.15... but if we will introduce a package manager by 3.16, then there's no point, probably. What do you think @capital-G and @JordanHendersonMusic ?

capital-G:
Maybe this could also be part of a FFI endeavor? (see https://github.com/supercollider/supercollider/pull/6615#issuecomment-2863407870)

Hopefully I'll have the package manager ready for 3.15, so all the plugin changes/updates due to breaking server API can be handled by the package manager :)

dyfer:
I'd like to share my progress here for reference:

TL;DR here's the branch with libgit building on all platforms: https://github.com/dyfer/supercollider/tree/topic/lib-git-cmake-mp-rebased-2 

(sorry it's squashed into a [single commit](https://github.com/dyfer/supercollider/commit/38e7cf7c6a6e3f79ddc7e6bd7b9f8d9602a216c5), I had problems with submodules when switching branches... although rebasing didn't end up helping, I kept updates on this branch, since it is also rebased on recent develop to get the updated build tools etc)

In short, one way to use a library in the build tree is to do add_subdirectory and use the targets exported by the library. 
It turned out, that libgit2 does not provide a static target that generates .a file under xcode, so I had to manually create a static target. That included some arguable practices, like adding an empty .c file so that xcode generator actually creates the static file... (I'm not sure if this is the only way to do it). Additionally, in order to override cmake flags of the library that's being imported/built, I've put it in a [sub-subfolder](https://github.com/dyfer/supercollider/commit/38e7cf7c6a6e3f79ddc7e6bd7b9f8d9602a216c5) (like we do with portaudio).

PkgConfig should be only used with system libraries. In my branch I have a [global variable](https://github.com/dyfer/supercollider/commit/38e7cf7c6a6e3f79ddc7e6bd7b9f8d9602a216c5#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR218) that switches between system on bundled libgit.

Once we get or create a proper target, or get it from pkgconfig, [linking to it is trivial](https://github.com/dyfer/supercollider/commit/38e7cf7c6a6e3f79ddc7e6bd7b9f8d9602a216c5#diff-6b82f8796ad633271adf03a85d3510784c692ce23a000df4dbdd4d1a4879d755R250-R258) and does not require manually specifying include libraries etc.

I've spent more time on this than I care to admin ;) I'll leave it here, in case it's useful for anybody in the future. Feel free to use the code that I came up with!

dyfer:
Okay, to clean this up...
It turns out that much of the messy cmake code inclusion was due to libgit2 missing a workaround for xcode and static lib creation. Here's another branch/commit, where I link my fork of libgit2 with a fix for that. Of course this is not meant for production, but rather to show the cleanest implementation. I have also submitted a PR to libgit2 with a proposed fix.

https://github.com/dyfer/supercollider/commit/6198957af756aed8539977c33c9c83cf8fdf1395

Importing a submodule really only requires [setting flags/variables and running `add_subdirectory`](https://github.com/dyfer/supercollider/commit/6198957af756aed8539977c33c9c83cf8fdf1395#diff-874dfd40805046a9377853ca4cdabb37b1cfdad78901831cdb66cd39557b0d0aR8-R17).

This creates target(s) from the subdirectory. Knowing which one to link to was tricky for me. It turns out that it's the [libgit2package](https://github.com/dyfer/supercollider/commit/6198957af756aed8539977c33c9c83cf8fdf1395#diff-6b82f8796ad633271adf03a85d3510784c692ce23a000df4dbdd4d1a4879d755R257) target that actually creates the library to link to.

JordanHendersonMusic:
@dyfer thanks for having a look at this! 

... although I fear for you that you might become the go to cmake wizard from now on...

PR #6383, opened 2024-07-02
SCLang: use likely and unlikely attributes for a performance boost

JordanHendersonMusic:
Measured at anywhere between a 5 and 19 percent boost!! No regressions encountered.

<details><summary>Using this benchmark</summary>
<p>

```
~b = { |func|
	var n = 1000;
	var f = n.collect {
		func.bench(false)
	};

	(f.sum / n).postln;
};

~b.({
	var e = ();
	10000.do{ |i|
		var n = 10000.rand;
		if(e[n.asSymbol].isNil) {
			e[n.asSymbol] = ();
		} {
			e[n.asSymbol][i.asSymbol] = n
		}
	};
});

~b.({
	10000.collect( _ * 2 );
});

~b.({
	10000.collect({|n| n.asFloat * 2.0 });
});

~b.({
	10000
	.collect(Ref(_))
	.collect(_.value)
	.collect(Ref(_))
	.collect(_.value)
	.collect(Ref(_))
	.collect(_.value)
	.collect(_.asString)
	.collect(_.asArray)
	.collect(_.asInteger)
	.sum
	;
});

~b.({
	var p = Pwhite().asStream;
	Array.fill(10000, {p.next})
});


"done";

```

</p>
</details> 

I saw the following performance increases:
```
17%
8%
5%
15%
19%
```
Linux GCC.
I think these are too good and I'm currently doing some more benchmarks to get a better value, but it is some improvement.

Pretty compelling for such a simple change. Would be great if others could test this!

## Types of changes

- C++ attributes, no functional changes.

I've added unlikely on the error path, and likely on the message lookup case because other op codes jumps to it with a `goto`. Also added this to the GC preferring when memory is recycled.

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/lang/LangSource/GC.h
+++ b/lang/LangSource/GC.h
@@ -303,7 +303,7 @@ inline void PyrGC::ToGrey2(PyrObjectHdr* obj) {
 }
 
 inline PyrObject* PyrGC::Allocate(size_t inNumBytes, int32 sizeclass, bool inRunCollection) {
-    if (inRunCollection && mNumToScan >= kScanThreshold)
+    if (inRunCollection && mNumToScan >= kScanThreshold) [[unlikely]]
         Collect();
     else {
         if (inRunCollection)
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
@timblechmann you mentioned you wrote this code, would love your opinion.

JordanHendersonMusic:
I'm closing this because I've messed up my benchmarks... its still a boost, but much smaller and I'll need more time to unpick what is actually going on...

PR #6384, opened 2024-07-03
Search for homebrew qt5 via cmake

capital-G:
## Purpose and Motivation

Currently it is necessary to manually specify the qt5 directory on macOS during configuration of the build.
This PR introduces a Cmake plugin/script which looks if `QT5_PREFIX_PATH` is specified, and if not, it will look for qt5 via homebrew and use it if it exists.
Therefore the configure command becomes

```
cmake -G Xcode  ..
```

instead of
```
cmake -G Xcode -DCMAKE_PREFIX_PATH=`brew --prefix qt5`  ..
```

for default use-cases. This mimics the behavior of other libraries such as `libsndfile`.

The PR therefore makes the configure process easier (as one has not to look up the `-DCMAKE_PREFIX_PATH=brew --prefix qt@5` command), which also making the IDE setup easier (see a potential #6351).

This is my first time working on Cmake files, so please review and give advice! One problem I found that if I invoke the command `cmake -G Xcode -DQT5_PREFIX_PATH=/usr/  ..` with a path that does not exist I can't "unset" it via `cmake -G Xcode ..` - only the deletion of my build dir makes it working again. Is this how cmake works?

I also don't know if it is the proper way to set the `CMAKE_PREFIX_PATH` variable to give Cmake a hint on the qt path as this could potentially overwrite the variable?

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer:
@capital-G thanks for this.

I was myself interested in not having to type the qt5 location. However, I found it cumbersome to implement it so that the non-homebrew options work as well. The tricky part I believe is that Qt might be available in the `PATH`, in which case the prefix path would not be set, and we shouldn't call `brew` in that case. This is what we're using for the legacy build right now (with the [install-qt-action](https://github.com/jurplel/install-qt-action?tab=readme-ov-file#more-info)) and more importantly this might be the way to install qt6 for releases moving forward, since the official Qt6 is a universal arm64/x64 binary. 

Currently, if Qt is in the `PATH` and `Qt5_DIR` is specified, the installed Qt (not from homebrew) is found automatically without specifying the prefix path. I believe that this PR might be breaking this, though the legacy build seemed to have worked, surprisingly. Still, in the configure step both [homebrew Qt](https://github.com/supercollider/supercollider/actions/runs/9781298339/job/27018849313?pr=6384#step:17:105) and Qt in the `PATH` are listed and I believe that shouldn't be the case.

dyfer:
One more piece of information: IIRC, the reason we need this is because qt5 from brew is not linked into system for some valid reasons. If qt were linked, we wouldn't need to specify the prefix.
I _think_ qt6 is by default linked so specifying the prefix path won't be needed once we migrate I believe. 

I can't double check this right now, but if that's the case, then I don't think we should merge this, even if it works well, as we will need to move to qt6 eventually.

capital-G:
> I don't think we should merge this, even if it works well

I can kind of understand the reasoning, but at the same time I'd like to make it easier for potential new developers/code lurkers to get started. The IDE setup took me several days/weeks and I think it would be good to share this so people can spend their time contributing instead of setting up their IDE/build chain.

Do you see a way to get both concerns under one hood? Calling a script to get env variables does not seem to be allowed in the VSCode IDE setup, but I think CLion, XCode etc do not allow this either?
We could also provide an additional preset in `CMakePresets.json` where the qt path is hardcoded (e.g. `macOS brew-qt`), as I am sure that 95% of MacOS developers simply use the Qt version provided by brew. Those 5% would most likely find a way to manually configure their compiler via flags (or maintain it via `CMakeUserPresets.json`) - see https://github.com/supercollider/supercollider/pull/6351

telephon:
> This mimics the behavior of other libraries such as `libsndfile`.

This is a good argument for this change, I think.

dyfer:
Again, we need to move to qt6. I'm 95% sure this won't be needed with qt6, since it is by default linked to $PATH by homebrew. Similarly, if you run `brew link qt5` (or whatever the correct command is), you don't need to provide the prefix path... 
If we do this, it would need to be removed once we move to Qt6. That update is on my list, but I have limited time and there are a few larger issues to solve so I don't have an ETA.
If we want to merge it temporarily I don't object, but we'd be removing it once we move to Qt6.

As for the logic here, IMO the script should _first_ check if Qt is available with `find_package(Qt5...` (without the `REQUIRED` flag) and only run the `brew` command if Qt is not found. But I'm not sure if that's a correct CMake approach...

dyfer:
FYI I _think_ we should be able to migrate to qt6 in a couple of weeks max (if nothing major comes up), at which point this won't be needed anymore.

capital-G:
In this case let's close it.

Issue #3604, opened 2018-03-25
Create flatpak distribution for Linux


nhthn:
[flatpak](https://flatpak.org/) seems to be a good way to go for packaging Linux binaries in a way that works across multiple distributions.

snappizz:
putting this on my personal backlog -- reopen if you plan on working on this before i do.

zapateo:
Any progress on this?

snappizz:
@brianlheim is working on CPack which will address many SC packaging
issues, but nobody's specifically working on this or planning to.

can someone close this issue? (im not logged in on mobile)

joshpar:
looks closed already

Newbytee:
Are there any plans to revitalise these efforts? Would you be open to outside help from the community to make this happen?

dyfer:
I am certainly open to help from the community :)

Newbytee:
Glad to hear. Could we reopen this in that case? I may have a go at it soon.

djiamnot:
So, I was looking for SC on flathub and for any efforts towards that, and stumbled upon [this repo](https://github.com/x37v/SuperCollider-flatpak) and since it didn't include sc3 plugins, I contributed to it and only after found this issue here. It makes sense, that SuperCollider org claims parenthood on flatpak. So, I've communicated with @x37v, and we agreed that I propose this solution here. 

I've tested the build, and it works on my computer and @x37v reported success installing it from a USB device, so this is, potentially, almost ready to ship (will have to update to pull the latest version). So, if there's anyone with more experience building flatpaks, please let us know if the manifest can be improved. Additionally, if someone has already an idea of how this could be integrated into this repo, I'm willing to tackle the process.

Can't wait to run SC on platforms that make flatpak a default (and much more friction-less than other methods)!

Cheers!

x37v:
> @x37v reported success installing it from a USB device

I haven't actually installed the SC flatpak from USB yet

quimnuss:
What's missing then? Installing from usb?

djiamnot:
[...]

PR #6393, opened 2024-07-18
build: install: add an AppStream metainfo file

djiamnot:
This change introduces the metainfo.xml file, that is required by the flatpak build and distribution system.

AppStream is an initiative that aims to create a unified software metadata database.

Add installation directive to install metainfo file

Fix issues with metainfo related to AppStream linter

## Purpose and Motivation

This MR is a follow-up to #3604. A [PR](https://github.com/flathub/flathub/pull/5350) into Flathub databse was opened and the only thing missing, at this point, is AppStream metainfo file. 

## Types of changes

- build/install

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [] Updated documentation

I think the only thing missing from this PR is integrating updates to the `metainfo` file via CI at release-time. I'm willing to look into that (I have not yet looked at the SC CI pipeline). It could be a separate PR.

In any case, the Flathub merge will likely not happen until the next CS release.

Regarding passing tests. On my computer, `develop` hangs at:

```
14/18 Test #14: timetag_test_run ...................   Passed    0.00 sec
      Start 15: sclang_crash_1_run
```
So does this branch, of course.

### DIFF
--- a/platform/linux/CMakeLists.txt
+++ b/platform/linux/CMakeLists.txt
@@ -10,5 +10,5 @@ install(FILES ../../icons/sc_ide.svg
 install(FILES supercollider.xml
     DESTINATION share/mime/packages )
 
-install(FILES io.github.supercollider.SuperCollider.metainfo.xml
+install(FILES online.supercollider.SuperCollider.metainfo.xml
     DESTINATION share/metainfo )
--- a/platform/linux/CMakeLists.txt
+++ b/platform/linux/CMakeLists.txt
@@ -9,3 +9,6 @@ install(FILES ../../icons/sc_ide.svg
 
 install(FILES supercollider.xml
     DESTINATION share/mime/packages )
+
+install(FILES io.github.supercollider.SuperCollider.metainfo.xml
+    DESTINATION share/metainfo )
--- a/platform/linux/io.github.supercollider.SuperCollider.metainfo.xml
+++ b/platform/linux/io.github.supercollider.SuperCollider.metainfo.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>io.github.supercollider.SuperCollider</id>
+
+  <name>SuperCollider</name>
+  <summary>SuperCollider is a platform for audio synthesis and algorithmic composition</summary>
+
+  <metadata_license>CC-BY-SA-4.0</metadata_license>
+  <project_license>GPL-3.0-or-later</project_license>
+  <developer_name>SuperCollider Community</developer_name>
+  <url type="homepage">https://supercollider.github.io/</url>
+  <url type="bugtracker">https://github.com/supercollider/supercollider/issues/</url>
+  <url type="help">https://scsynth.org/</url>
+  <url type="contribute">https://github.com/supercollider/supercollider/blob/develop/CONTRIBUTING.md</url>
+  <launchable type="desktop-id">SuperColliderIDE.desktop</launchable>
+
+  <description>
+    <p>
+      SuperCollider is a platform for audio synthesis and algorithmic composition, used by musicians, artists, and researchers working with sound. It consists of:
+    </p>
+    <p>
+      - scsynth, a real-time audio server with hundreds of unit generators (&quot;UGens&quot;) for audio analysis, synthesis, and processing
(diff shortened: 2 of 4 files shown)


capital-G:
Hey, thanks for this and sorry for the long wait. I strongly support the approach towards flatpack!

I still have some questions regarding the PR: Shouldn't the flatpack build script not be part of this repo so we can control and maintain it? I thought that the real benefit of flatpack is that there is no up/downstream anymore as "the devs" can maintain a unified binary/build or a build script for linux.

In the end I'd like to endorse this as the "supported" way of using SC under Linux, but this would make it also mandatory to add it to our build chain in our CI.
Our CI is currently broken :/ But if you need support with this, I can take a look so we can add this as an additional build step.

@JordanHendersonMusic as you are using Linux (IIRC), what is your take on this?

One thing that I'd like to omit would be the `github.io` namespace in front of the meta file? I think we don't want to be under the umbrella of a proprietary and not-owned namespace.

JordanHendersonMusic:
> @JordanHendersonMusic as you are using Linux (IIRC), what is your take on this?

I know nothing about flatpack, sorry!

capital-G:
Regarding `github.io` namespace - I just took a look at https://docs.flathub.org/docs/for-app-authors/requirements/#application-id  and https://docs.flathub.org/docs/for-app-authors/verification/#what-is-verification and IIUC, if we have our own URL we could use this instead of the GitHub.

I am still in process in establishing the `supercollider.online` URL (as seen e.g. in https://docs.supercollider.online/) as a somehow official alternative, but this would require a RFC to pull through with it.
But in the end I think it would be worth the effort, and would demonstrate why such a self-owned URL would be necessary.

djiamnot:
@capital-G : thank you for your feedback and questions!

> I still have some questions regarding the PR: Shouldn't the flatpack build script not be part of this repo so we can control and maintain it? 

I thought so, too, initially. I admit, I'm cutting my flatpak teeth on this project, so still a rookie. It turns out, that the flatpak manifest must be merged into the [Flathub](https://github.com/flathub/) repo. I'm still unsure if the approach I took is the most efficient. I do think, that at this time it's mostly frictionless.

Note, however, that this PR is for [AppStream](https://www.freedesktop.org/software/appstream/docs/) `meta-info` file. While required by Flatpak, it's not exclusive to it. It serves the Freedesktop project in a wider sense.

> I thought that the real benefit of flatpack is that there is no up/downstream anymore as "the devs" can maintain a unified binary/build or a build script for linux.

Flatpak is one possible solution to package and distribute software that aims to be distribution agnostic. It also offers sandboxing functionalities. It does not eliminate the need for upstream, though.

For the record, I piggybacked the issue https://github.com/supercollider/supercollider/issues/3604 that was addressing Flatpak, perhaps this discussion could continue over there.

> One thing that I'd like to omit would be the `github.io` namespace in front of the meta file? I think we don't want to be under the umbrella of a proprietary and not-owned namespace.

I agree. As I understand it, the `app-id` parameter must be the reverse-DNS of the application.

> I am still in process in establishing the supercollider.online URL  […]

Oh, cool! The Flathub PR https://github.com/flathub/flathub/pull/5350 will probably only pull through with the next SC release, I would say, so there's still time to update the rDNS, once it's official.

Thank you!

capital-G:
Just as an update - this is currently blocked by https://github.com/supercollider/supercollider.github.io/pull/278 to which I'll hopefully write a RFC soon in order to get things going.

djiamnot:
So I was wondering, do we really need to wait for the consensus on the SuperCollider domain to approve this PR?  Both the AppStream metainfo and Flatpak will have to be updated on new releases, anyway. I'm thinking this, because I see a new SC release _and_ a bot on the Flatpak repo is reminding me of an old PR, that is blocked by this one.

djiamnot:
Thank you @capital-G for digging into this! Much appreciated. Of course, I can give write access to the repo, and can keep an eye on keeping up the maintenance. Obviously, I didn't read that part of the Flathub documentation. Let's wait til it's all settled, then.

And yes, I have been installing the Flatpak build manually, when I needed. 
There are instructions for [hosting own flatpak repo on GitLab/HitHub](https://docs.flatpak.org/en/latest/hosting-a-repository.html#hosting-a-repository-on-gitlab-github-pages), I didn't try this way.

dyfer (approved):

capital-G (approved):

PR #6394, opened 2024-07-21
add clipInsert method to ArrayedCollection and List

prko:
- add clipInsert method to ArrayedCollection and List
- update documentation for ArrayedCollection, Array and List

## Purpose and Motivation
After seeing the following thread on the scsynth forum,
https://scsynth.org/t/list-array-expansion-question/9808/8
I think the necessity of the `.clipInsert` method for `ArrayedCollection` and `List`.

## Types of changes

- Documentation
- New feature
- 
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/ArrayedCollection.schelp
+++ b/HelpSource/Classes/ArrayedCollection.schelp
@@ -247,6 +247,30 @@ y.postln;
 )
 ::
 
+method::clipInsert
+Inserts the item into the contents of the receiver, maintaining the size of the receiver. This method may return a new ArrayedCollection. For this reason, you should always assign the result of code::insert:: to a variable - never depend on add changing the receiver.
+code::
+~array = [10, 20, 30]
+~array_modified = ~array.clipInsert(0, 1)
+~array
+~array_modified
+
+~array = [10, 20, 30]
+~array_modified = ~array.clipInsert(1, 1)
+~array
+~array_modified
+
+~array = [10, 20, 30]
+~array_modified = ~array.clipInsert(2, 1)
+~array
+~array_modified
+
+~array = [10, 20, 30]
+~array_modified = ~array.clipInsert(3, 1)
+~array
+~array_modified
+::
+
 method::move
 Moves an item from one position to another.
 
--- a/HelpSource/Classes/List.schelp
+++ b/HelpSource/Classes/List.schelp
@@ -100,6 +100,31 @@ Inserts the strong::item:: at the beginning of the List.
 
 method::insert
 Inserts the strong::item:: into the contents of the List at the indicated strong::index::.
+code::
+~list = List[10, 20, 30, 40]
+~list.insert(0, 1)
+~list
+::
+
+method::clipInsert
+Inserts the strong::item:: into the contents of the List at the indicated strong::index::, maintaining the size of the List.
+code::
+~list = List[10, 20, 30, 40]
+~list.clipInsert(0, 1)
+~list
+
+~list = List[10, 20, 30, 40]
+~list.clipInsert(2, 1)
+~list
+
+~list = List[10, 20, 30, 40]
+~list.clipInsert(3, 1)
+~list
+
+~list = List[10, 20, 30, 40]
+~list.clipInsert(10, 1)
+~list
+::
 
 method::pop
 Remove and return the last element of the List.
--- a/SCClassLibrary/Common/Collections/ArrayedCollection.sc
+++ b/SCClassLibrary/Common/Collections/ArrayedCollection.sc
@@ -183,6 +183,11 @@ ArrayedCollection : SequenceableCollection {
 		_ArrayInsert
 		^this.primitiveFailed;
 	}
+	clipInsert { arg index, item;
+		var lastIndex = this.size - 1;
+		index = if(index > lastIndex) { lastIndex } { index };
+		^this.insert(index, item)[0..lastIndex]
+	}
 	move { arg fromIndex, toIndex;
 		^this.insert(toIndex, this.removeAt(fromIndex))
 	}
--- a/SCClassLibrary/Common/Collections/List.sc
+++ b/SCClassLibrary/Common/Collections/List.sc
@@ -48,6 +48,10 @@ List : SequenceableCollection {
 	add { arg item; array = array.add(item); }
 	addFirst { arg item; array = array.addFirst(item); }
 	insert { arg index, item; array = array.insert(index, item); }
+	clipInsert { arg index, item;
+		var lastIndex = this.size - 1;
+		index = if(index > lastIndex) { lastIndex } { index };
+		array = array.insert(index, item)[0..lastIndex] }
 	removeAt { arg index; ^array.removeAt(index); }
 	pop { ^array.pop }
 	first { if (this.size > 0, { ^array.at(0) }, { ^nil }) }
--- a/HelpSource/Classes/Array.schelp
+++ b/HelpSource/Classes/Array.schelp
@@ -81,6 +81,8 @@ copymethod:: ArrayedCollection -put
 
 copymethod:: ArrayedCollection -insert
 
+copymethod:: ArrayedCollection -clipInsert
+
 copymethod:: ArrayedCollection -clipAt
 
 copymethod:: ArrayedCollection -wrapAt


JordanHendersonMusic:
Could you explain why this is necessary here rather than linking to a post? 

Why can't someone just write `a.insert(i.clip(0, a.size), x)` is this supposed to make it harder to get wrong, in that case, I'd rather see better error messages for insert added.

prko:
@JordanHendersonMusic 
>Could you explain why this is necessary here rather than linking to a post?

I added the explanation in the PR.

---

>Why can't someone just write `a.insert(i.clip(0, a.size), x)` is this supposed to make it harder to get wrong?

Sorry. I do not understand the code `a.insert(i.clip(0, a.size), x)`. What is `i` here?

---

>Also, insert does change the size of the container
a = []
a.insert(0, 1) // [1]
a // [1]

um..  Not always:

```
a = []
a.insert(0, 1) // [1]
a // [1]
a.insert(0, 2) // [2, 1]
a // [1]
a.insert(0, 3) // [3, 1]
a // [1]
```

JordanHendersonMusic:
@prko Thanks for the update!

I think the name is wrong. 

The clipped prefix, as in `clippedPut` clips the index, not the container size.

prko:
> I think the name is wrong.
> 
> The clipped prefix, as in `clippedPut` clips the index, not the container size.

Do you mean the `.clipPut` method?
https://doc.sccode.org/Overviews/Methods.html#clipPut

In fact, I took the name for the functionality in this PR from `clipPut`. If the `clipInsert` is a wrong name, could you suggest one? How about `insertNonIncremental`? It is too long, I think.

JordanHendersonMusic:
> I took the name for the functionality in this PR from clipPut

And that is why I think the name is wrong because this does something else on top of clipping the index.

The name should be long because this is complex behaviour that combines clipping the index and trimming the size.
I don't have a good  idea for the name, but how about `clipInsertMaintainSize`.

---

Actually.... I've just had a change of thought. 

I don't think this should be a method on `Array`, but on `RingBuffer`. Then the method is just called `clipInsert` as maintaining the size is what `RingBuffer` does.

prko:
@JordanHendersonMusic 
Adding this method to `RingBuffer`would not be a problem.
Changing the method name for `ArrayedCollection` and `List` to `clipInsertMaintainSize` is OK for me.
I do not want this long name, but I accept your idea as your thought is more logical than mine.

However, `RingBuffer` is not used very often, I think. It only appears 7 times in sccode.org:
http://sccode.org/search?q=RingBuffer

What do you think if I did the following?
- Change the method name to `clipInsertMaintainSize` for `List` and `ArrayedCollection`.
- add `clipInsert` to `RingBuffer

JordanHendersonMusic:
Actually, I had a closer look at RingBuffer and don't think it would work, it's not designed to silently drop items.

Perhaps `clipInsertKeepSize` is better, follows on from `keep`.

prko:
OK, then I do not change `RingBuffer`.

How about `insertKeepSize` instead of `clipInsertKeepSize `?

JordanHendersonMusic:
If you drop the clipping behaviour and so what insert does when you try to insert in an invalid index, then you should drop the clip from the name.

prko:
> Perhaps `clipInsertKeepSize` is better, follows on from `keep`.

Done!
But `keep` is a method of `SequenceableCollection` whiliest`clipInsertKeepSize` is a method for `ArrayedCollection` and `List`. So it could not follow on from `keep`.

prko:
Would it be OK if I implemented the `insertKeepSize` method on this PR? I think I would use `insertKeepSize` more than `clipInsertKeppSize`.

prko:
added `insertKeepSize`.

prko:
Is there anything that needs correcting?
Could this PR be approved and merged before the 3.14 release?

dyfer:
@JordanHendersonMusic what's the status of your review?

telephon (commented):
My suggestion would be to just add one of the two methods (the clipInsertKeepSize). I don't see why we need the first one separately. Then, the name `clipInsert` would cover everything just fine (roughly speaking, insert in a clipped way).

What do you think?

prko:
I am in favour of removing `insertKeepSize` and renaming `clipInsertKeepSize` to `clipInsert`.  

This pull request was created ten months ago, and I can no longer recall the exact reasoning behind my later addition of `insertKeepSize`. The only thing I remember is that I felt the name `clipInsertKeepSize` was excessively long. Therefore, I am uncertain whether implementing this functionality is truly beneficial for structuring code, considering the length of the method name.  

```  
~newItem = 1;  
~array1 = [20, 30, 40, 90];  

~array2 = ([1] ++ ~array1).keep(~array1.size);  
~array2 = ~array1.insert(0, 1).keep(~array1.size);  
~array2 = ~array1.clipInsertKeepSize(0, 1);  
~array2 = ~array1.clipInsert(0, 1);  
```  

That said, the name `clipInsert` is concise, and I believe it is a good choice.  

If @JordanHendersonMusic also agrees with this, I will update this PR immediately.

SCClassLibrary/Common/Collections/ArrayedCollection.sc
```diff
@@ -183,6 +183,15 @@ ArrayedCollection : SequenceableCollection {
...
 		^this.primitiveFailed;
 	}
+	insertKeepSize { arg index, item;
+		var lastIndex = this.size - 1;
+		^this.insert(index, item)[0..lastIndex]
+	}
+	clipInsertKeepSize { arg index, item;
+		var lastIndex = this.size - 1;
+		index = if(index > lastIndex) { lastIndex } { index };
+		^this.insert(index, item)[0..lastIndex]
```
JordanHendersonMusic, line 193: I'm just thinking about when this makes a copy of the array and when it just returns the old array. Because the size doesn't change, I think it should always return the old array, but the `[0..n]` forces it to make a copy. Is that bit necessary?

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -247,6 +247,39 @@ y.postln;
 )
 ::
 
+method::insertKeepSize
+Same as link::#-insert::, but it maintains the size of the receiver.
+code::
+~array = [10, 20, 30]
+~array_modified = ~array.insertKeepSize(0, 1)
+~array
+~array_modified
```
JordanHendersonMusic, line 256: Could you add in comments what the results of these variables are please? Having to run the documentation to find out what the method does is sometimes annoying.

JordanHendersonMusic (changes requested):
I like the shorter name. There are just two points here, one's about copies and the other about the docs.

HelpSource/Classes/ArrayedCollection.schelp
prko: ok, I will do it when updating the next commit after we finish our discussion in https://github.com/supercollider/supercollider/pull/6394#discussion_r2055604390

SCClassLibrary/Common/Collections/ArrayedCollection.sc
prko: Thanks.

I changed `clipInsert`method for `ArrayedCollection : SequenceableCollection` as follows:
```
ArrayedCollection : SequenceableCollection {
...
	clipInsert { arg index, item;
		index = index.clip(0, this.lastIndex);
		^this.insert(index, item)[0..this.lastIndex]
	}
...
}
```
Then, the following is the result:
```
~arrayOld = [1, 2, 3, 4] // -> [1, 2, 3, 4]
~arrayOld.clipInsert(20, \a) // -> [1, 2, 3, a]
~arrayOld // -> [1, 2, 3, 4]

~arrayNew = ~arrayOld.clipInsert(20, \a) // -> [1, 2, 3, a]
~arrayOld // -> [1, 2, 3, 4]
~arrayNew // -> [1, 2, 3, a]
```

The `clipInsert` method for `List` is as follows:
```
List : SequenceableCollection {
...
	clipInsert { arg index, item;
		var lastIndex = array.size - 1;
		index = if(index > lastIndex) { lastIndex } { index };
		this.insert(index, item);
		array.removeAt(lastIndex + 1);
		array
	}
...
}
```
The test result is as follows:
```
~listOld = List [1, 2, 3, 4] // -> List[1, 2, 3, 4]
~listOld.clipInsert(20, \a) // -> List[1, 2, 3, a]
~listOld // -> List[1, 2, 3, a]

~listOld = List [1, 2, 3, 4] // -> List[1, 2, 3, 4]
~listNew = ~listOld.clipInsert(20, \a) // -> List[1, 2, 3, a]
~listNew // -> List[1, 2, 3, a]
```
If this is correct, I shall proceed with the documentation and upload a new commit!

JordanHendersonMusic: Okay I see why that is necessary now. Yes this looks good.

JordanHendersonMusic:
> 	var arrayTemp = this[0..lastIndex];

No don't do this. I misunderstood why the copy was needed. This implementation https://github.com/supercollider/supercollider/pull/6394#discussion_r2055828349 is good.

HelpSource/Classes/ArrayedCollection.schelp
prko: Done in https://github.com/supercollider/supercollider/pull/6394/commits/7228e2967c774bf81401f1e055896e8873afde89

SCClassLibrary/Common/Collections/ArrayedCollection.sc
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6394/commits/7228e2967c774bf81401f1e055896e8873afde89

SCClassLibrary/Common/Collections/List.sc
```diff
@@ -48,6 +48,19 @@ List : SequenceableCollection {
...
+		array.removeAt(lastIndex + 1);
+		array
+	}
+	clipInsertKeepSize { arg index, item;
+		var lastIndex = array.size - 1;
+		index = if(index > lastIndex) { lastIndex } { index };
+		this.insert(index, item);
+		array.removeAt(lastIndex + 1);
+		array
+	}
```
elgiano, line 63: shouldn't `insertKeepSize` and `clipInsertKeepSize` be replaced with clipInsert?

elgiano (changes requested):
thanks! I think List has to be updated too, to have only `clipInsert` rather than the two older implementations

SCClassLibrary/Common/Collections/List.sc
prko: Thank you very much! 
I apologise for forgetting to upload this part. 

I should interpret the sccode in VScodium, but I find SC-IDE more convenient to use, so I always manually copy the code from SC-IDE to VScodium. 

Unfortunately, I missed it this time
—
my apologies, and thank you

elgiano: shouldn't List.clipInsert return a new List? the current implementation returns an array and modifies the old list

elgiano (changes requested):

prko:
@elgiano 
>shouldn't List.clipInsert return a new List? the current implementation returns an array and modifies the old list

I would like to ensure I have correctly understood your review.  

**Step. 1:**  
> The current implementation returns an array ...  

Based on the evaluation results, the current implementation behaves as follows:  

```supercollider
l = List[1, 2, 3, 4, 5]; // -> List[1, 2, 3, 4, 5]
l.clipInsert(0, \a);     // -> List[a, 1, 2, 3, 4]
l                        // -> List[a, 1, 2, 3, 4]
l.class                  // -> List
```  

As illustrated above, `List.clipInsert` returns an instance of `List`.  


**Sep. 2:**  
However, I believe you may be referring to a different array—the array within `List` that is revealed using the `.dump` method, as demonstrated below:  

```supercollider
l = List[1, 2, 3, 4, 5]; // -> List[1, 2, 3, 4, 5]
l.dump                   
/*
List's array:
Instance of Array {    (0x162424a98, gc=B0, fmt=01, flg=00, set=03)
  indexed slots [5]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
      3 : Integer 4
      4 : Integer 5
}
-> List[1, 2, 3, 4, 5]
*/
l.clipInsert(0, \a);     // -> List[a, 1, 2, 3, 4]
l                        // -> List[a, 1, 2, 3, 4]
l.class                  // -> List
l.clipInsert(6, \b);     // -> List[a, 1, 2, 3, b]
l                        // -> List[a, 1, 2, 3, b]
l.class                  // -> List
l.dump
/*
List's array:
Instance of Array {    (0x162424a98, gc=B0, fmt=01, flg=00, set=03)
  indexed slots [5]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
      4 : Symbol 'b'
}
-> List[a, 1, 2, 3, b]
*/
```  

Since the virtual memory address (the hexadecimal number prefixed with `0x`) remains unchanged, I presume this is what you meant when stating:  
> "The current implementation ... modifies the old list."  

Regarding the implementation, I reused the logic of `List.insert` to create the `.clipInsert` method for `List`. The behaviour of `List.insert` appears to be consistent in this regard (**the virtual memory address remains unchanged!**), as demonstrated below:  

```supercollider
m = List[1, 2, 3, 4, 5] // -> List[1, 2, 3, 4, 5]
m.dump
/*
List's array:
Instance of Array {    (0x325938518, gc=9C, fmt=01, flg=00, set=03)
  indexed slots [5]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
      3 : Integer 4
      4 : Integer 5
}
-> List[1, 2, 3, 4, 5]
*/  
m.insert(0, \a) // -> List[a, 1, 2, 3, 4, 5]
m.class         // -> List
m.dump
/*
List's array:
Instance of Array {    (0x325938518, gc=9C, fmt=01, flg=00, set=03)
  indexed slots [6]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
      4 : Integer 4
      5 : Integer 5
}
-> List[a, 1, 2, 3, 4, 5]
*/
```  

**Sep. 3:**  
Would you mind confirming whether I have understood your point correctly?  

If I misunderstood you, could you give the correct implementation? In many cases, I am a but unsure with code structure and underlying concepts, so I often misunderstand many thing. The thing illustrated above is just sho what I think.








I would like to ensure that I have correctly understood your review.  

### Step 1
> The current implementation returns an array...

Based on the evaluation results, the current implementation behaves as follows:  

```supercollider
l = List[1, 2, 3, 4, 5]; // -> List[1, 2, 3, 4, 5]
l.clipInsert(0, \a);     // -> List[a, 1, 2, 3, 4]
l                        // -> List[a, 1, 2, 3, 4]
l.class                  // -> List
```  
As illustrated above, `List.clipInsert` returns an instance of `List`.  

### Step 2
However, I believe you may be referring to a different array—the array within `List` that is revealed using the `.dump` method, as demonstrated below:  

```supercollider
l = List[1, 2, 3, 4, 5]; // -> List[1, 2, 3, 4, 5]
l.dump                   
/*
List's array:
Instance of Array {    (0x162424a98, gc=B0, fmt=01, flg=00, set=03)
  indexed slots [5]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
      3 : Integer 4
      4 : Integer 5
}
-> List[1, 2, 3, 4, 5]
*/
l.clipInsert(0, \a);     // -> List[a, 1, 2, 3, 4]
l                        // -> List[a, 1, 2, 3, 4]
l.class                  // -> List
l.clipInsert(6, \b);     // -> List[a, 1, 2, 3, b]
l                        // -> List[a, 1, 2, 3, b]
l.class                  // -> List
l.dump
/*
List's array:
Instance of Array {    (0x162424a98, gc=B0, fmt=01, flg=00, set=03)
  indexed slots [5]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
      4 : Symbol 'b'
}
-> List[a, 1, 2, 3, b]
*/
```  

Since the virtual memory address (the hexadecimal number prefixed with `0x`) remains unchanged, I presume this is what you meant when stating:  
> "The current implementation ... modifies the old list."  

However, regarding the implementation, I believe that I reused the logic of `List.insert` to create the `.clipInsert` method for `List`. The behaviour of `List.insert` appears to be consistent in this regard (**the virtual memory address remains unchanged!**), as demonstrated below:  

```supercollider
m = List[1, 2, 3, 4, 5] // -> List[1, 2, 3, 4, 5]
m.dump
/*
List's array:
Instance of Array {    (0x325938518, gc=9C, fmt=01, flg=00, set=03)
  indexed slots [5]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
      3 : Integer 4
      4 : Integer 5
}
-> List[1, 2, 3, 4, 5]
*/  
m.insert(0, \a) // -> List[a, 1, 2, 3, 4, 5]
m.class         // -> List
m.dump
/*
List's array:
Instance of Array {    (0x325938518, gc=9C, fmt=01, flg=00, set=03)
  indexed slots [6]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
      4 : Integer 4
      5 : Integer 5
}
-> List[a, 1, 2, 3, 4, 5]
*/
```  

### **Step 3:**  
Would you mind confirming whether I have correctly understood your point?  

If I have misunderstood your comment, I would appreciate any guidance on the correct implementation. I sometimes struggle with code structure and underlying concepts, so I may misinterpret certain aspects. The explanation above simply reflects my current understanding.  (I did not use `.dump` while implementing these methods...)

Thank you for your time and feedback—I genuinely appreciate it!

SCClassLibrary/Common/Collections/List.sc
prko: No need to say sorry! I answered below:
1. https://github.com/supercollider/supercollider/pull/6394#issuecomment-2835867659
2. https://github.com/supercollider/supercollider/pull/6394#issuecomment-2835910611

After writing my initial response, I realised that it might not directly address your review. To ensure clarity and relevance, I have briefly written a new answer.

elgiano:
You're right that your method returns a List. I got confused because the last line of List.clipInsert is `array` and I thought that was the return value. However, since it doesn't have a caret (`^array`) and it's a class method, it's not returning that value, but it implicitly returns `this` (the List). 

You're also right about List not "needing" a reassignment. None of their methods does, as they manage the array re-assignment internally. So, it's idiomatic for List to have modifications in-place, without reassignment. 

So, it's all good, I just got confused, but why don't we copy the implementation for List.insert:
```supercollider
	insert { arg index, item; array = array.insert(index, item); }
```
and we make clipInsert with the same structure?
```supercollider
	clipInsert { arg index, item; array = array.clipInsert(index, item); }
```
This is also how all other List methods are implemented, and IMO this would be less code to maintain and think about, and less prone to confusion. What do you think?

elgiano:
You're right, we have a problem now. Actually it turns out we also have it with SequenceableCollection. Try:
```supercollider
a = [1,2,3]
a.clipInsert(0,0) // -> [0,1,2,3]
```
This doesn't work, the array is growing in size (exactly like the list).
To debug this, try to add some postln to the two calls to this.lastIndex:
```supercollider
	clipInsert { arg index, item;
		index = index.clip(0, this.lastIndex.postln);
		^this.insert(index, item)[0..this.lastIndex.postln];
	}
```
now this code:
```supercollider
[1,2,3].clipInsert(0,0)
```
prints:
```supercollider
2
3
-> [0,1,2,3]
```
So, the first number (2) is this.lastIndex at line 187. The second (3) is this.lastIndex at line 188.
What happens is the following:
1. we clip index between 0 and 2
2. we insert an item into the array, now the array is size 4, and this.lastIndex is 3
3. we apply the final clipping [0..this.lastIndex], but now this.lastIndex is 3, so the array is not resized at all, and stays size 4.

This happens because Array.insert didn't return a new Array. I know it's confusing, but actually Array.insert will return a new array only if a new allocation is needed. And a new allocation is needed when the array grows more than its "internal" size, which you can check by using `maxSize`:

```supercollider
a = (0..2)
a.size // -> 3
a.maxSize // -> 4

a = (0..9)
a.size // -> 10
a.maxSize // -> 16

a = (0..199)
a.size // -> 200
a.maxSize // -> 256
```

Max size is always the nearest bigger power of 2. I believe this is done for instance to reduce frequent reallocation. Let me know if you would like more guidance to find where in sclang's cpp this happens.

My suggestion for implementation is: store the old size first, then insert and clip.

```supercollider
	clipInsert { arg index, item;
		var oldSize = this.size;
		^this.insert(index, item).keep(oldSize);
	}
```

note that we also don't need to clip the index, it's done automatically by [_ArrayInsert](https://github.com/supercollider/supercollider/blob/916b10bb59a5020155cc7fa45758bbd9b307645e/lang/LangPrimSource/PyrArrayPrimitives.cpp#L934C1-L937C37) (which is called by this.insert)

I hope it's all clear enough! Please let me know if you have any question!

elgiano:
While I was checking this, it turned out we already have `clipPut`. Have you looked into it? It looks like it does exactly the same thing. I'm sorry this didn't come up before, but I'm happy if it was an occasion for practicing and learning something (in case you missed it, check out the previous comments for why the current implementation didn't work)!

```supercollider
[0,1,2].clipPut(-1,\a) // [\a, 1, 2]
[0,1,2].clipPut(1,\a)  // [0 \a, 2]
[0,1,2].clipPut(99,\a) // [0, 1, \a]

List[0,1,2].clipPut(-1,\a) // [\a, 1, 2]
List[0,1,2].clipPut(1,\a)  // [0 \a, 2]
List[0,1,2].clipPut(99,\a) // [0, 1, \a]
```

elgiano:
I'm super sorry for the unnecessary chaos, @cdbzb is right that clipPut is not the same as clipInsert. I should have triple checked, thanks a lot! 

About the implementation with `drop` I like it more than the one with `keep` because it's more coincise

prko:
Thank you for your valuable and kind opinions.

I have conducted further testing and would like to provide a detailed report along with the reasoning behind my latest commit.

### Comparison of Suggested Implementations
#### 1. The implementation and results of this PR before April

**Implementation:**
<details><summary>ArrayedCollection</summary>
<pre>
ArrayedCollection {
	...
	clipInsertKeepSize { arg index, item;
		var lastIndex = this.size - 1;
		index = if(index > lastIndex) { lastIndex } { index };
		^this.insert(index, item)[0..lastIndex]
	}
	...
}
</pre>
</details> 

<details><summary>List</summary>
<pre>
List {
	...
	clipInsertKeepSize { arg index, item;
		var lastIndex = array.size - 1;
		index = if(index > lastIndex) { lastIndex } { index };
		this.insert(index, item);
		array.removeAt(lastIndex + 1);
		array
	}
	...
}
</pre>
</details>

**Method renaming:**
<details><summary>ArrayedCollection</summary>
<pre>
ArrayedCollection {
	...
	clipInsert { arg index, item;
		var lastIndex = this.size - 1;
		index = if(index > lastIndex) { lastIndex } { index };
		^this.insert(index, item)[0..lastIndex]
	}
	...
}
</pre>
</details>

<details><summary>List</summary>
<pre>
List {
	...
	clipInsert { arg index, item;
		var lastIndex = array.size - 1;
		index = if(index > lastIndex) { lastIndex } { index };
		this.insert(index, item);
		array.removeAt(lastIndex + 1);
		array
	}
	...
}
</pre>
</details>

**Test results:**
<details><summary>ArrayedCollection</summary>
<pre>
a = [1, 2, 3]; // -> [1, 2, 3]
a.maxSize      // -> 4
a.dump
/*
Instance of Array {    (0x153aca938, gc=C4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/

b = a.clipInsert(0, \a); // -> [a, 1, 2]
b                        // -> [a, 1, 2]
b.class                  // -> Array
b.dump
/*
Instance of Array {    (0x153b69d58, gc=C4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
}
-> [a, 1, 2]
*/

a                        // -> [a, 1, 2, 3]
a.dump
/*
Instance of Array {    (0x153aca938, gc=C4, fmt=01, flg=00, set=02)
  indexed slots [4]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
}
-> [a, 1, 2, 3]
*/

c = b.clipInsert(-2, '-a');  // -> [-a, a, 1]
c                            // -> [-a, a, 1]

d = c.clipInsert(2, \b);     // -> [-a, a, b]
d                            // -> [-a, a, b]

e = d.clipInsert(9, \c);     // -> [-a, a, c]
e                            // -> [-a, a, c]
e.dump
/*
Instance of Array {    (0x15b14ee88, gc=C4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol '-a'
      1 : Symbol 'a'
      2 : Symbol 'c'
}
-> [-a, a, c]
*/

a.dump
/*
Instance of Array {    (0x153aca938, gc=C8, fmt=01, flg=00, set=02)
  indexed slots [4]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
}
-> [a, 1, 2, 3]
*/
</pre>
</details>

<details><summary>List</summary>
<pre>
l = List[1, 2, 3];      // -> List[1, 2, 3]
l.dump
/*
List's array:
Instance of Array {    (0x15ecfa548, gc=C8, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> List[1, 2, 3]
*/

l.clipInsert(0, \a);    // -> List[a, 1, 2]
l                       // -> List[a, 1, 2]
l.class                 // -> List

l.clipInsert(-2, '-a'); // -> List[-a, a, 1]
l                       // -> List[-a, a, 1]

l.clipInsert(2, \b);    // -> List[-a, a, b]
l                       // -> List[-a, a, b]

l.clipInsert(9, \c);    // -> List[-a, a, c]
l                       // -> List[-a, a, c]

l.dump
/*
List's array:
Instance of Array {    (0x15ecfa548, gc=CC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol '-a'
      1 : Symbol 'a'
      2 : Symbol 'c'
}
-> List[-a, a, c]
*/
</pre>
</details> 

#### 2. The implementation based on @elgiano’s latest idea

**Implementation:**
<details><summary>ArrayedCollection</summary>
<pre>
ArrayedCollection {
	...
	clipInsert  { arg index, item;
		var oldSize = this.size;
		index = index.clip(0, this.lastIndex);
		^this.insert(index, item).keep(oldSize);
	}
	...
}
</pre>
</details>

<details><summary>List</summary>
<pre>
List {
	...
	clipInsert { arg index, item; array = array.clipInsert(index, item); }
	...
}
</pre>
</details>

**Test results:**
<details><summary>ArrayedCollection</summary>
<pre>
a = [1, 2, 3]; // -> [1, 2, 3]
a.maxSize      // -> 4
a.dump
/*
Instance of Array {    (0x15ef61c58, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/

b = a.clipInsert(0, \a); // -> [a, 1, 2]
b                        // -> [a, 1, 2]
b.class                  // -> Array
b.dump
/*
Instance of Array {    (0x3018ae6e8, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
}
-> [a, 1, 2]
*/

a                        // -> [a, 1, 2, 3]
a.dump
/*
Instance of Array {    (0x15ef61c58, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [4]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
}
-> [a, 1, 2, 3]
*/

c = b.clipInsert(-2, '-a');  // -> [-a, a, 1]
c                            // -> [-a, a, 1]

d = c.clipInsert(2, \b);     // -> [-a, a, b]
d                            // -> [-a, a, b]

e = d.clipInsert(9, \c);     // -> [-a, a, c]
e                            // -> [-a, a, c]
e.dump
/*
Instance of Array {    (0x1393b4118, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol '-a'
      1 : Symbol 'a'
      2 : Symbol 'c'
}
-> [-a, a, c]
*/

a.dump
/*
Instance of Array {    (0x15ef61c58, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [4]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
}
-> [a, 1, 2, 3]
*/
</pre>
</details>

<details><summary>List</summary>
<pre>
l = List[1, 2, 3];      // -> List[1, 2, 3]
l.dump
/*
List's array:
Instance of Array {    (0x15ef5b0f8, gc=B4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> List[1, 2, 3]
*/

l.clipInsert(0, \a);    // -> List[a, 1, 2]
l                       // -> List[a, 1, 2]
l.class                 // -> List

l.clipInsert(-2, '-a'); // -> List[-a, a, 1]
l                       // -> List[-a, a, 1]

l.clipInsert(2, \b);    // -> List[-a, a, b]
l                       // -> List[-a, a, b]

l.clipInsert(9, \c);    // -> List[-a, a, c]
l                       // -> List[-a, a, c]
l.dump
/*
List's array:
Instance of Array {    (0x138d8f628, gc=B4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol '-a'
      1 : Symbol 'a'
      2 : Symbol 'c'
}
-> List[-a, a, c]
*/
</pre>
</details>

#### 3.  The implementation combining @cdbzb’s approach for arrays and @elgiano’s approach for lists

**Implementation:**

<details><summary>ArrayedCollection</summary>
<pre>
ArrayedCollection {
	...	
	boundedInsert { arg index, item;
		^this.drop(-1).insert(index, item);
	}
	...
}
</pre>
</details>

<details><summary>List</summary>
<pre>
List {
	...	
	boundedInsert { arg index, item; array = array.boundedInsert(index, item); }
	...
}
</pre>
</details>

**Test results:**
<details><summary>ArrayedCollection</summary>
<pre>
a = [1, 2, 3]; // -> [1, 2, 3]
a.maxSize      // -> 4
a.dump
/*
Instance of Array {    (0x15d878ee8, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/

b = a.boundedInsert(0, \a); // -> [a, 1, 2]
b                        // -> [a, 1, 2]
b.class                  // -> Array
b.dump
/*
Instance of Array {    (0x15b8e99b8, gc=B0, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
}
-> [a, 1, 2]
*/

a                        // -> [1, 2, 3]
a.dump
/*
Instance of Array {    (0x15d878ee8, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/

c = b.boundedInsert(-2, '-a');  // -> [-a, a, 1]
c                            // -> [-a, a, 1]

d = c.boundedInsert(2, \b);     // -> [-a, a, b]
d                            // -> [-a, a, b]

e = d.boundedInsert(9, \c);     // -> [-a, a, c]
e                            // -> [-a, a, c]
e.dump
/*
Instance of Array {    (0x15fcdadd8, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol '-a'
      1 : Symbol 'a'
      2 : Symbol 'c'
}
-> [-a, a, c]
*/

a.dump
/*
Instance of Array {    (0x15d878ee8, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/
</pre>
</details>

<details><summary>List</summary>
<pre>
l = List[1, 2, 3];      // -> List[1, 2, 3]
l.dump
/*
List's array:
Instance of Array {    (0x13951a1b8, gc=A8, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> List[1, 2, 3]
*/

l.boundedInsert(0, \a);    // -> List[a, 1, 2]
l                       // -> List[a, 1, 2]
l.class                 // -> List

l.boundedInsert(-2, '-a'); // -> List[-a, a, 1]
l                       // -> List[-a, a, 1]

l.boundedInsert(2, \b);    // -> List[-a, a, b]
l                       // -> List[-a, a, b]

l.boundedInsert(9, \c);    // -> List[-a, a, c]
l                       // -> List[-a, a, c]
l.dump
/*
List's array:
Instance of Array {    (0x13930c4c8, gc=A8, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol '-a'
      1 : Symbol 'a'
      2 : Symbol 'c'
}
-> List[-a, a, c]
*/
</pre>
</details>

### Differences between implementations
There are notable differences in how these three implementations affect the code evaluation results:
- My implementation before this April:
  - ArrayedCollection 
     - An array instance retains its memory address. 
     - The receiver (the array itself) is also modified within `maxSize`.
  - List 
     - A list instance retains its memory address.
     - The receiver (the list itself) is also affected.
- @elgiano's implementation suggested:
  - ArrayedCollection 
     - Behaviour remains the same as described above.
  - List 
     - **A list instance changes its memory address**.
     - The receiver (the list itself) is also modified.
- @cdbzb's implementation suggested:
  - ArrayedCollection 
     - An array instance **changes its memory address**.
     - **The receiver (an array instance itself) is not affected**.
  - List 
     - A list instance **changes its memory address**.
     - The receiver (a list instance itself) is also affected.

### Conclusion
Considering the evaluation of `.insert` for `ArrayedCollection` and `List`:
```supercollider
a = [1,2,3] // -> [1, 2, 3]
a.dump
/*
Instance of Array {    (0x15825e488, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/
a.insert(1, 200) // -> [1, 200, 2, 3]
a.dump
/*
Instance of Array {    (0x15825e488, gc=AC, fmt=01, flg=00, set=02)
  indexed slots [4]
      0 : Integer 1
      1 : Integer 200
      2 : Integer 2
      3 : Integer 3
}
-> [1, 200, 2, 3]
*/
```
```supercollider
l = List[1, 2, 3];      // -> List[1, 2, 3]
l.dump
/*
List's array:
Instance of Array {    (0x13951a1b8, gc=A8, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> List[1, 2, 3]
*/
```
the following conclusions can be drawn:
- Both `ArrayedCollection` and `List` instances should preserve their memory addresses while still being affected by the method implemented in this PR.
- Among the three implementations examined, only the first implementation meets these criteria.

Thus, I believe the first implementation is the correct choice.
While I personally favour the brevity of `.clipInsert`, I find @cdbzb’s reasoning quite convincing. Considering all of these aspects, I have uploaded the latest commit accordingly: https://github.com/supercollider/supercollider/pull/6394/commits/5c27af7978a9cf520e098a626cbfb0276fc38e6b

I would greatly appreciate your feedback once again!

**One apology**:
When updating this PR in April, I overlooked the implementation for `lastIndex`, which resulted in unintended behaviour in that commit. I will be more cautious in the future.

SCClassLibrary/Common/Collections/List.sc
prko: See: https://github.com/supercollider/supercollider/pull/6394#issuecomment-2840806443

elgiano:
From your report above, I understand you don't want `clipInsert` to re-allocate an Array. I think it's reasonable, because `insert` reallocates only if we exceed maxSize, and in this case we want to keep size constant, so we should never exceed maxSize and thus we should never reallocate. Even if this behavior is sometimes confusing, I think it's beneficial for `clipInsert` and `insert` to share the same behavior, and not forcing a copy if `insert` wouldn't also do it.

(About renaming to boundedInsert, I have no strong opinion, but I tend to prefer `clipInsert` until we make an explicit decision)

I have two points: 
1. **List implementation**: whatever we choose to do, List implementation should be done like most other List methods, calling `clipInsert` on the internal array. I don't see any reason to have a special implementation for List, so we can keep code in one place and directly call the Array one:
```supercollider
	clipInsert { arg index, item; array = array.clipInsert(index, item); }
```

2. **Array implementation**: we could combine @cdbzb method with the requirement of not reallocating, by first removing last item and then inserting the new one:
```supercollider
	clipInsert { arg index, item;
		if (this.size == 0) { ^this };
		this.pop;
		this.insert(index, item);
	}
```

note that I needed to addres the "array is empty" case explicitly, otherwise `[].clipInsert` would not keep its size to 0.

prko:
Thank you very much for your comment and implementation suggestion!  
**Your implementation is indeed correct!**  

Here is how I have demonstrated that my approach was incorrect, and yours is correct:

#### Result of the method in `ArrayedCollection`:
The following is the implementation of your last idea for `ArrayedCollection`:
```
a = [1, 2, 3]; // -> [1, 2, 3]

a.dump
/*
Instance of Array {    (0x12cee2de8, gc=D4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/

b = a.clipInsert(0, \a); // -> [a, 1, 2]
b                        // -> [a, 1, 2]
b.class                  // -> Array
b.dump
/*
Instance of Array {    (0x12cee2de8, gc=D4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
}
-> [a, 1, 2]
*/

a                        // -> [a, 1, 2]
a.dump
/*
Instance of Array {    (0x12cee2de8, gc=D4, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
}
*/

b == a  // -> true
b === a // -> true
```

There are two different things in comparison to my implementation idea:
```
a = [1, 2, 3]; // -> [1, 2, 3]
a.dump
/*
Instance of Array {    (0x17ac152c8, gc=D0, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Integer 1
      1 : Integer 2
      2 : Integer 3
}
-> [1, 2, 3]
*/

b = a.boundedInsert(0, \a); // -> [a, 1, 2]
b                        // -> [a, 1, 2]
b.class                  // -> Array
b.dump
/*
Instance of Array {    (0x3080f72e8, gc=D0, fmt=01, flg=00, set=02)
  indexed slots [3]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
}
-> [a, 1, 2]
*/

a                        // -> [a, 1, 2, 3]
a.dump
/*
Instance of Array {    (0x17ac152c8, gc=D0, fmt=01, flg=00, set=02)
  indexed slots [4]
      0 : Symbol 'a'
      1 : Integer 1
      2 : Integer 2
      3 : Integer 3
}
-> [a, 1, 2, 3]
*/

b == a  // -> false
b === a // -> false
```
I am also attaching this difference with an image:
cccc

**I inspected the `.insert` method behaviour for `ArrayedCollection` as well. Your implementation is much closer to `.insert` method behaviour for `ArrayedCollection`.** 

**Additionally, my failure to consider a size 0 instance of `ArrayedCollection` was incorrect.**

#### Result of the method in `List`:
The code evaluation of the implementation of your last idea for `List` is the same as the `insert` implementation for `List`:

#### Conclusion
I have changed the method in this PR as you suggested: https://github.com/supercollider/supercollider/pull/6394/commits/9987dcf5893db862d271e79e10cc964bb49b7d08

Thanks again!

(For the name, I want to see the opinion of other reviewers again.)

HelpSource/Classes/List.schelp
```diff
@@ -100,6 +100,35 @@ Inserts the strong::item:: at the beginning of the List.
...
+~alist                        // -> List[10, 20, 1, 30]
+
+~alist = List[10, 20, 30, 40] // -> List[10, 20, 30, 40]
+~alist.boundedInsert(3, 1)    // -> List[10, 20, 30, 1]
+~alist                        // -> List[10, 20, 30, 1]
+
+~alist = List[10, 20, 30, 40] // -> List[10, 20, 30, 40]
+~alist.boundedInsert(10, 1)   // -> List[10, 20, 30, 1]
+~alist                        // -> List[10, 20, 30, 1]
+::
```
elgiano, line 131: Do we need to repeat examples here? Or could we just link to Classes/ArrayedCollection#-boundedInsert ?

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -247,6 +247,32 @@ y.postln;
...
+~arrayA                                   // -> [a, 1, 2]
+~arrayB                                   // -> [a, 1, 2]
+~arrayC = ~arrayA.boundedInsert(20, \a)   // -> [a, 1, a]
+~arrayA                                   // -> [a, 1, a]
+~arrayC                                   // -> [a, 1, a]
+~arrayD = ~arrayA.boundedInsert(-2, '-1') // -> [-1, a, 1]
+~arrayA                                   // -> [-1, a, 1]
+~arrayD                                   // -> [-1, a, 1]
+::
+
```
elgiano, line 275: Could the examples be shorter and focus on a few core points we want to communicate about this method? It could be useful to take the help for Array#-insert as an inspiration

prko: Um... there as a problem:
```
(
// in this case a new object is returned
var y, z;
z = [1, 2, 3, 4];
y = z.boundedInsert(1, 999);
z.postln;
y.postln;
)
```
returns
```
[1, 999, 2, 3]
[1, 999, 2, 3]
-> [1, 999, 2, 3]
```
The problem seems to be the inconsistent behaviour of the array:
```
a = [1, 2, 3, 4] // -> [1, 2, 3, 4]
b = a.insert(1, 99) // -> [1, 99, 2, 3, 4]
a                   // -> [1, 2, 3, 4]
b                   // -> [1, 99, 2, 3, 4]

a = [1, 2, 3] // -> [1, 2, 3]
b = a.insert(1, 99) // -> [1, 99, 2, 3]
a                   // -> [1, 99, 2, 3]
b                   // -> [1, 99, 2, 3]
```

elgiano:
Yes, now boundedInsert never returns a new object anymore, which I think it's fine. Insert also is not supposed to do it, only sometimes. I think we can keep the example structure and use it to show that the original array is always affected

HelpSource/Classes/ArrayedCollection.schelp
prko: See https://github.com/supercollider/supercollider/pull/6394/commits/5e18589a240aa319f02e7744454cb8b2daf754e3.

prko:
How about this change in https://github.com/supercollider/supercollider/pull/6394/commits/5e18589a240aa319f02e7744454cb8b2daf754e3?

elgiano:
Could somebody else help reviewing the documentation for this PR, please? I'm afraid I don't have enough time and I can't make up my mind on the balance between verbosity and clarity.... Pinging @tedmoore because for now there is no one else on the docs team

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -235,18 +235,66 @@ z.fill([1, 2, 3, 4]).postln;
...
+y.postln;         // [1, 999, 2, 3]
+(z == y).postln;  // true
+(z === y).postln; // true
+)
+::
+
+method::boundedInsert
+Same as link::#-insert::, but it maintains the size of the receiver, and the value for the strong::index:: greater than link::Classes/ArrayedCollection:: instance size minus one will be clipped to code::size - 1::, which is the last index.
+
+ccode::
```
HotwheelsSisyphus, line 270: typo: 
code::

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -235,18 +235,66 @@ z.fill([1, 2, 3, 4]).postln;
...
+// .boundedInsert never returns a new object  
+// because inserting a new item never exceeds the maxSize.  
+var y, z;
+z = [1, 2, 3, 4];
+y = z.boundedInsert(1, 999);
+z.postln;         // [1, 999, 2, 3]
+y.postln;         // [1, 999, 2, 3]
+(z == y).postln;  // true
+(z === y).postln; // true
 )
```
HotwheelsSisyphus, line 281: Suggest removing the ~arrayA etc. example (284-296); 
but demonstrate behavior when index out of bounds with the y,z example, like in 293-94.
(Please check that this works!)

y = y.boundedInsert(20, \a);            // [1, 999, 2, a]
z = z.boundedInsert(-2, \b);             // [b, 1, 999, 2]

HelpSource/Classes/List.schelp
```diff
@@ -100,6 +100,32 @@ Inserts the strong::item:: at the beginning of the List.
 
 method::insert
 Inserts the strong::item:: into the contents of the List at the indicated strong::index::.
+code::
```
HotwheelsSisyphus, line 103: Suggestion: x, not ~alist; semicolon at end of every line; also changed exmaples slightly

code::
x = List[1, 2, 3, 4]; // -> List[1, 2, 3, 4]
x.insert(0, 999);           // -> List[999, 1, 2, 3, 4]
x;                        // -> List[999, 1, 2, 3, 4]
::

method::boundedInsert
Same as link::#-insert::, but removes the receiver's last element before inserting strong::item::.
This changes the receiver but maintains its size.


code::
(
x = List[1, 2, 3, 4]; // -> List[1, 2, 3, 4]
x.boundedInsert(1, 999);    // -> List[1, 999, 2, 3]
x;                            // -> List[1, 999, 2, 3]
x.boundedInsert(20, \a);   // -> List[1, 999, 2, a]
x.boundedInsert(-2, \b);   // ->  List[b, 1, 999, 2]
)
 ::

HotwheelsSisyphus (commented):
Thanks! Alright, I quickly went over the docs (not the sclang code). Please verify that the changes I proposed work as I think they do. 

There are some general doc code style questions that haven't been decided yet, for now let's not worry about that too much here.

dyfer:
I updated the PR title to reflect the current method name, since that goes into the commit message. In general let's try to keep the title (and, if feasible, the opening comment) in sync with the state of the PR. Feel free to change/update as needed.

HelpSource/Classes/List.schelp
prko: Thank you, done in https://github.com/supercollider/supercollider/pull/6394/commits/f9624bb4509700ff51b39a4fe48a32d3cd301abe.

I believe that these code lines should be individual; hence, I have excluded the parentheses in the second code example block. 

I also feel that a one-line code snippet does not require a terminating semicolon delimiter. However, this is purely my personal preference, so I have left it as is. I have also commented on this matter in comment #6806.

HelpSource/Classes/ArrayedCollection.schelp
prko: Thank you, done in https://github.com/supercollider/supercollider/commit/f9624bb4509700ff51b39a4fe48a32d3cd301abe.

HelpSource/Classes/ArrayedCollection.schelp
prko: Thank you, done in https://github.com/supercollider/supercollider/commit/f9624bb4509700ff51b39a4fe48a32d3cd301abe.

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -235,18 +235,66 @@ z.fill([1, 2, 3, 4]).postln;
...
+y = z.insert(1, 999);
+z.postln;         // [1, 999, 2, 3]
+y.postln;         // [1, 999, 2, 3]
+(z == y).postln;  // true
+(z === y).postln; // true
+)
+::
+
+method::boundedInsert
+Same as link::#-insert::, but it maintains the size of the receiver, and the value for the strong::index:: greater than link::Classes/ArrayedCollection:: instance size minus one will be clipped to code::size - 1::, which is the last index.
```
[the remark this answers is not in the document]
prko: Thank you, done in https://github.com/supercollider/supercollider/commit/f9624bb4509700ff51b39a4fe48a32d3cd301abe.

HelpSource/Classes/ArrayedCollection.schelp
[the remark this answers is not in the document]
prko: Thank you, done in https://github.com/supercollider/supercollider/commit/f9624bb4509700ff51b39a4fe48a32d3cd301abe.

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -235,16 +235,46 @@ z.fill([1, 2, 3, 4]).postln;
...
-y = z.insert(1, 999);
-z.postln;
-y.postln;
-)
+Inserts the item into the contents of the receiver.
+This method may return a new ArrayedCollection.
+For this reason, always assign the result of code::.insert:: to a variable - never depend on code::.insert:: changing the receiver.
+code::
+// In this case, a new object is returned  
+// because the maxSize of `z` is 4,
```
tedmoore, line 243: Is there a place where the question of `Why is the max size 4?` answered. It is confusing to a reader that this magic number appears here so if more context is given, that would be excellent.

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -235,16 +235,46 @@ z.fill([1, 2, 3, 4]).postln;
...
+y = z.insert(1, 999); // -> [1, 999, 2, 3]
+z;                    // -> [1, 999, 2, 3]
+y;                    // -> [1, 999, 2, 3]
+(z == y);             // -> true
+(z === y);            // -> true
+::
+
+method::boundedInsert
+Same as link::#-insert::, but removes the receiver's last element before inserting strong::item::.
+This changes the receiver but maintains its size.
```
tedmoore, line 265: Because `method::insert` explains what the method returns in the method description, I think `.boundedInsert` should here as well.  I see it says it in the code block. It should also say it in the description.

tedmoore: something like "This changes the receiver but maintains its size, therefore unlike code::.insert:: it is not necessary to reassign the result of code::.boundedInsert::."

tedmoore (approved):

HelpSource/Classes/ArrayedCollection.schelp
HotwheelsSisyphus: that was my formulation, my thinking was that "same as .insert, but" would clarify what is being returned, and that "This changes the receiver" implies that the reassignment is not needed. Your suggestion sounds good too, though.

HelpSource/Classes/ArrayedCollection.schelp
HotwheelsSisyphus: suggestion: (@tedmoore please check): 

because the maximum size of z is set at instantiation and cannot be changed at a later stage.

also below, 252ff.:
because the new number of elements does not exceed the number of elements at instantiation.

HelpSource/Classes/ArrayedCollection.schelp
```diff
@@ -235,16 +235,46 @@ z.fill([1, 2, 3, 4]).postln;
...
+y = z.insert(1, 999); // -> [1, 999, 2, 3, 4]
+z                     // -> [1, 2, 3, 4]
+y                     // -> [1, 999, 2, 3, 4]
+(z == y)              // -> false
+(z === y)             // -> false
+
+// In this case, the same object is returned  
+// because the maxSize of `z` is 4,  
+// and inserting a new item does not exceed this limit. 
+z = [1, 2, 3];        // -> [1, 999, 2, 3]
```
HotwheelsSisyphus, line 255: 255: change to

z = [1, 2, 3]; // -> [1, 2, 3]

I hope this isn't something from my suggestions last week, if so, apologies!

prko: No need to apologise!
You suggested a space after the code, and applying this to the whole code example is as follows
```
code::
// In this case, a new object is returned  
// because the maxSize of `z` is 4,  
// and inserting a new item exceeds this limit. 
z = [1, 2, 3, 4]; // -> [1, 2, 3, 4]
y = z.insert(1, 999); // -> [1, 999, 2, 3, 4]
z // -> [1, 2, 3, 4]
y // -> [1, 999, 2, 3, 4]
(z == y) // -> false
(z === y) // -> false

// In this case, the same object is returned  
// because the maxSize of `z` is 4,  
// and inserting a new item does not exceed this limit. 
z = [1, 2, 3]; // -> [1, 999, 2, 3]
y = z.insert(1, 999); // -> [1, 999, 2, 3]
z; // -> [1, 999, 2, 3]
y; // -> [1, 999, 2, 3]
(z == y); // -> true
(z === y); // -> true
::
```
I thought it was a bit confusing because the vertical alignment of the comments is different.

But I guess that is my personal preference. (Aligning comments vertically is time-consuming.) 

I will include the code example above in my next commit. If I now commit this change, the two previous discussions can be considered outdated.

---

Maybe we need a code style:
- A white space before a single-line comment
- Do not align single-line comments within a block of code.

HotwheelsSisyphus: Oh, to be clearer: My last comment wasn't about the spacing or alignment but about how z = [1,2,3]  will return [1,2,3] and not [1,999,2,3]!

HelpSource/Classes/ArrayedCollection.schelp
prko: To confirm if I understand correctly:

lines 240-244:
```
// In this case, a new object is returned  
// because adding a new item would exceed the fixed maximum size of 4, 
// which is set at the instantiation of the array z and 
// cannot be changed at a later stage.
// (By default, an array initialized with a size between 1 and 4 
// retains a maximum size of 4): 
```

lines 252-254:
```
// In this case, the same object is returned 
// because the new number of elements does not exceed 
// the number of elements at instantiation:
```

HelpSource/Classes/ArrayedCollection.schelp
prko: To confirm if I correctly understand:

```
method::boundedInsert
Change the receiver but maintains its size, therefore unlike code::.insert:: it is not necessary to reassign the result of code::.boundedInsert::.

code::
....
```
like this?

As far as I understand, the recommended approach for the first sentence in a method is to omit the subject when it refers to the method’s name and to use the infinitive to create an instructive, declarative tone.

HelpSource/Classes/ArrayedCollection.schelp
prko: Ah, thank you! I misread... sorry! corrected in my local repository.

HelpSource/Classes/ArrayedCollection.schelp
tedmoore: Correct. The verb would be this:

`Changes the receiver but maintains its size, therefore unlike code::.insert:: it is not necessary to reassign the result of code::.boundedInsert::.`

As in "This method"...`Changes the receiver...`

Thanks!

HelpSource/Classes/ArrayedCollection.schelp
tedmoore: Thanks @prko !

> // (By default, an array initialized with a size between 1 and 4 
> // retains a maximum size of 4):

This is the part that I am thinking will be valuable, and your longer explanation is great. Thank you!

HelpSource/Classes/ArrayedCollection.schelp
prko: @tedmoore 
I completely agree with you. However, would you be able to take a look at the following posts?
- https://github.com/supercollider/supercollider/pull/6534#issuecomment-2613427923
- https://scsynth.org/t/time-to-have-standard-style-to-write-help-documentation/11088/12?u=prko

Previously, I changed all the first verbs to the singular third-person present form, but later revised them to the infinitive form.

Could you kindly review this and confirm once more?”

tedmoore: jamshark70 (I don't know James' GH handle?) said it well:

> Here, the method itself is the only possible subject of the sentence, and the verb form can’t be imperative, so the meaning must be that the method is what’s doing it. For English speakers, it might be a bit pedantic to write “keep(n) – this method keeps the first n items…”

> Without the s, then yes, it feels like an imperative – “keep(n) – keep the first n items” although this could be taken as instructing the computer to do the thing.

> For English speakers, I don’t think either one is wrong. But you raise a good point – non-native English speakers may get tripped up over “is that a command or not?”

In some recent schelp styleguide meetings we've opted to "use the s". In the example above it would mean using "Changes..." (with the s)

HelpSource/Classes/ArrayedCollection.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6394/commits/93b4c0074a1ff0584a37a6a7c2187e296a92c28f!

tedmoore (approved):

HelpSource/Classes/ArrayedCollection.schelp
prko: >In some recent schelp styleguide meetings we've opted to "use the s".

Sorry, I did not know it! Thank you for letting me know it.

Done in https://github.com/supercollider/supercollider/pull/6394/commits/93b4c0074a1ff0584a37a6a7c2187e296a92c28f!

dyfer:
@JordanHendersonMusic there were changes to this PR after your approval, is this still good to go?

JordanHendersonMusic:
This has changed a lot since then - for the better! The implementation is good but I don't have time to look at the documentation.

PR #6398, opened 2024-07-23
SCLang: add keywords to newCopyArgs as constructors are messy with many instance variables

JordanHendersonMusic:
## Purpose and Motivation
If you have a class with many instance variables and want to only initialise some of them, you currently need to call setters, which gets messy and can be overloaded.

Now you can call them like so...
```
super.newCopyArgs(1, 2, 3, foo: 10, bar: 20)
```

You can also directly expose the constructor like this...
```
*new { |...args, kwargs|
   ^super.performArgs(\newCopyArgs, args, kwargs)
}
```
In future this could become the default implementation of `Object*new`.

This PR applied this to `SynthDef` as an example.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/Classes/Object.schelp
+++ b/HelpSource/Classes/Object.schelp
@@ -50,29 +50,36 @@ method::newCopyArgs
 
 Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined.
 
-Importantly, super class constructors and NOT called, but the first values will be copied.
-Think of instance variables as entries in an array,
-the super class's instance variables are at the beginning and the child class's at the end,
-this method only copies the arguments into that array.
+Importantly, superclass constructors are NOT called, but the values are still placed in the superclass's instance variables.
 
 code::
 MyClass {
 	var a, b, c;
     // Will copy arg1, arg2, arg3 to variables a, b, c.
 	*new { |arg1, arg2, arg3| ^super.newCopyArgs(arg1, arg2, arg3) }
+	*newKw { |arg1| ^super.newCopyArgs(arg1, c: 10) }
 	value { ^[a, b, c] }
 }
-MyClass(1, 2, 3).() == [1, 2, 3]
+MyClass.new(1, 2, 3).value() == [1, 2, 3];
+MyClass.newKw(1).value() == [1, nil, 10];
 ::
 
+Example of inheritance.
 code::
-MyClass {
+Base {
 	var a, b, c;
-    // Also supports keyword arguments, unassigned variables default to nil.
-	*new { |arg1| ^super.newCopyArgs(arg1, c: 10) }
-	value { ^[a, b, c] }
+	*new { ^\didNotCallBase } // This is NEVER called
 }
-MyClass(1).() == [1, nil, 10]
+
+Derived : Base {
+    var d, e, f;
+    *new { |d, e, f| ^super.newCopyArgs(1, 2, 3, d, e, f) }
+    *newKw { |d, e, f| ^super.newCopyArgs(c: 3, d: d, e: e, f: f) }
+	value { ^[a, b, c, d, e, f] }
+}
+
+Derived.new(10, 11, 12).value() == [1, 2, 3, 10, 11, 12];
+Derived.newKw(10, 11, 12).value() == [nil, nil, 3, 10, 11, 12];
 ::
 
 
--- a/HelpSource/Classes/Object.schelp
+++ b/HelpSource/Classes/Object.schelp
@@ -50,7 +50,7 @@ method::newCopyArgs
 
 Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined.
 
-Importantly, superclass constructors are NOT called, but the values are still placed in the superclass's instance variables.
+Importantly, superclass constructors are NOT called, but the values are directly placed in the superclass's instance variables.
 
 code::
 MyClass {
--- a/HelpSource/Classes/Object.schelp
+++ b/HelpSource/Classes/Object.schelp
@@ -48,19 +48,33 @@ Create a new instance. The creation of new instances of any Object actually happ
 
 method::newCopyArgs
 
-Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined. If the superclass's code::new:: method requires arguments, the first arguments passed to code::newCopyArgs:: will be passed on to that method, and the following arguments will be copied to this class's instance variables. Class variables are ignored.
+Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined.
+
+Importantly, super class constructors and NOT called, but the first values will be copied.
+Think of instance variables as entries in an array,
+the super class's instance variables are at the beginning and the child class's at the end,
+this method only copies the arguments into that array.
 
 code::
 MyClass {
 	var a, b, c;
-
-	*new {
-		arg arg1, arg2, arg3;
-		// will copy arg1, arg2, arg3 to variables a, b, c
-		^super.newCopyArgs(arg1, arg2, arg3);
-	}
+    // Will copy arg1, arg2, arg3 to variables a, b, c.
+	*new { |arg1, arg2, arg3| ^super.newCopyArgs(arg1, arg2, arg3) }
+	value { ^[a, b, c] }
 }
+MyClass(1, 2, 3).() == [1, 2, 3]
 ::
+
+code::
+MyClass {
+	var a, b, c;
+    // Also supports keyword arguments, unassigned variables default to nil.
+	*new { |arg1| ^super.newCopyArgs(arg1, c: 10) }
+	value { ^[a, b, c] }
+}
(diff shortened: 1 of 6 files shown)


HelpSource/Classes/Object.schelp
```diff
@@ -48,20 +48,34 @@ Create a new instance. The creation of new instances of any Object actually happ
 
 method::newCopyArgs
 
-Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined. If the superclass's code::new:: method requires arguments, the first arguments passed to code::newCopyArgs:: will be passed on to that method, and the following arguments will be copied to this class's instance variables. Class variables are ignored.
+Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined.
+
+Importantly, super class constructors and NOT called, but the first values will be copied.
```
mtmccrea, line 53: ... superclass constructors are NOT ...

JordanHendersonMusic: This was never the case. Perhaps it's a bug?

HelpSource/Classes/Object.schelp
```diff
@@ -48,20 +48,34 @@ Create a new instance. The creation of new instances of any Object actually happ
...
+Think of instance variables as entries in an array,
+the super class's instance variables are at the beginning and the child class's at the end,
+this method only copies the arguments into that array.
```
mtmccrea, line 56: > this method only copies the arguments into that array.

Could you clarify what "that array" is? Do you mean the method only copies the args to the child's instance variables?

and spelling: superclass

HelpSource/Classes/Object.schelp
```diff
@@ -48,20 +48,34 @@ Create a new instance. The creation of new instances of any Object actually happ
...
 
 code::
 MyClass {
 	var a, b, c;
+    // Will copy arg1, arg2, arg3 to variables a, b, c.
+	*new { |arg1, arg2, arg3| ^super.newCopyArgs(arg1, arg2, arg3) }
+	value { ^[a, b, c] }
+}
+MyClass(1, 2, 3).() == [1, 2, 3]
```
mtmccrea, line 65: In this case perhaps it's best to just call `.value` instead of `.()`, to be explicit.

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -11,13 +11,9 @@ Object {
 		// to actually put things in the object you need to
 		// add them.
 	}
-	*newCopyArgs { arg ... args;
+	*newCopyArgs { | ... args, kwargs |
 		_BasicNewCopyArgsToInstVars
 		^this.primitiveFailed
-		// creates a new instance that can hold up to maxSize
```
mtmccrea, line 17: is all the info in this comment no longer applicable? if some of the comment is valid, perhaps updating/rewording it to fit the new functionality (or clarify it... offhand I'm not sure what it's referring to, or if it's mistakenly referring to slotted objects like `Array`?)

mtmccrea:
Interesting feature! I'm not starting a full review, just some quick comments after a glance at the updated documentation.

HelpSource/Classes/Object.schelp
mtmccrea: I was just referring to spelling of _superclass_ and _are_.

But yes, if there's a doubt about whether args are supposed to propagate to the superclass, would be good to confirm that (I suspect not, this would obscure behavior).

JordanHendersonMusic: .... Hahaha! I proofed this too 😂

SCClassLibrary/Common/Core/Object.sc
JordanHendersonMusic: It's a repeat of `new`, it's directly above it and this is a copy and paste. It just doesn't apply here, the size is fixed and the items are inserted.

JordanHendersonMusic:
@mtmccrea thanks for the review, I've fixed the typos now.

I also discovered a bug, I think it was a long standing bug that was only trigger by the recent keyword argument improvements. 

```
A {
   var a, b;
   *new { ^super.newCopyArgs(10, b: 1) }
}
```
The above method would not set `b` to `1` as the keyword arguments where not being passed to the underlying primitive. This was a compiler issue in call node and I just removed the optimisation when there are keywords. There might be a better fix, but because the byte-codes are complicated, numerous, and poorly documented in the code, I can't find it. This wasn't found before because there were very few places where primitives are called with keyword arguments, and even fewer where the method's type is changed from `methNormal` to `methPrimitive`.

JordanHendersonMusic:
One benifit to this is that we don't need to create public getters and setters to initialise only some of the member variables in a constructor.

HelpSource/Classes/Object.schelp
mtmccrea: The example you added is illustrative, thanks!

JordanHendersonMusic:
@mtmccrea did you have any more thoughts about this one? IMO, I think it simplifies constructors quite a lot and helps keep encapsulation as you don't need to access setters in the *new method, so I can't see any downside.

mtmccrea:
I agree this is a good addition. There's a meta discussion that could be had about proper/encouraged construction and setting of members, but that's on the class authoring side. The feature only adds more flexibility, and fixes a bug, so I'm in favor of it. 

Unfortunately I'm not able to properly review the cpp code, so it would be good if someone more familiar with this could have a look.

Thanks for your work on this!

HelpSource/Classes/Object.schelp
```diff
@@ -48,20 +48,41 @@ Create a new instance. The creation of new instances of any Object actually happ
 
 method::newCopyArgs
 
-Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined. If the superclass's code::new:: method requires arguments, the first arguments passed to code::newCopyArgs:: will be passed on to that method, and the following arguments will be copied to this class's instance variables. Class variables are ignored.
+Creates a new instance and copies the arguments to the instance variables in the order that the variables were defined.
+
+Importantly, superclass constructors are NOT called, but the values are still placed in the superclass's instance variables.
```
telephon, line 53: "still" can be left out, could add "directly"

telephon (commented):
THis is a good idea. Currently, I can't compile, so I can't test. But from reading, it looks good.

HelpSource/Classes/Object.schelp
JordanHendersonMusic: Okay all done :)

telephon (changes requested):
Could you please add a test where instance variables are set that do not exist?

```
*newNonExistentKeywords { |a, b| ^super.newCopyArgs(1, 2, 3, y:c, d:a, x:b) }
```

JordanHendersonMusic:
@telephon I actually already had this hidden in there. I've made it more explicit now.

TLDR:

It throws a warning.
```
WARNING: Could not find instance variable with name 'y' in class 'TestConstructorA'
```

However, if the instance exists in a derived class the warning isn't shown.
```
TestConstructorBase {
    var a, b, c;
    // see 'd' here is set.
    *newNonExistentKeywords {  ^super.newCopyArgs(1, 2, 3, y: 10, d: 20, x: 30) }
    value { ^[a, b, c] }
}

TestConstructorDerived : TestConstructorBase {
    var d, e, f;
    value { ^[a, b, c, d, e, f] }
}
```
```
TestConstructorBase.newNonExistentKeywords().() == [1, 2, 3]; // shows warning for y, d and x.

TestConstructorDerived.newNonExistentKeywords().() == [1, 2, 3, 20, nil, nil]; // d was set, shows warning for y and x.
```

This could be a little surprising in some circumstances, but I don't *think* it will cause any confusion?

telephon (approved):

telephon:
> This could be a little surprising in some circumstances, but I don't _think_ it will cause any confusion?

So this means there are cases in which you call *newCopyArgs with keywords that do not exist, there won't be a warning, but it will just silently (correctly) fail to set them?

telephon:
Yes, that is unchanged – thanks fo rthe clear example, I think this should go somewhere in the help files. What I meant was the `Could not find instance variable` warning: is it omitted in the case of subclasses, and may that be a problem?

JordanHendersonMusic:
Sorry I don't really understand what you are asking. The warning triggers if an instance variable of that name cannot be found in the class being constructed.

telephon:
Sorry! Yes, we want this warning. If this warning is always there when we want it to be there, all is well.

JordanHendersonMusic:
This seems like it is good to merge to me, any one else have any concerns? 

The main benefit of this is to simply construction of objects when you want to initial members in an arbitrary order without exposing their public setter.

telephon:
yes, ready to merge, maybe someone else can take a brief look and merge.

JordanHendersonMusic:
Would any one like to merge?

PR #6402, opened 2024-07-27
Supernova fix

Spacechild1:
## Purpose and Motivation

- Update nova-tt to get bug fix for `nova::detail::pause`.
- Improve `rw_spinlock::pause` function in `SC_SndBuf.h` (add support for MSVC and ARM platforms)
- Supernova: avoid division by zero in back off calibration

Fixes https://github.com/supercollider/supercollider/issues/5899.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/include/plugin_interface/SC_SndBuf.h
+++ b/include/plugin_interface/SC_SndBuf.h
@@ -29,8 +29,14 @@
 #    include <atomic>
 #    include <cassert>
 
-#    ifdef __SSE2__
-#        include <emmintrin.h>
+/* for rw_spinlock::pause() */
+#    if defined(_MSC_VER)
+#        ifndef NOMINMAX
+#            define NOMINMAX
+#        endif
+#        include <Windows.h>
+#    elif defined(__SSE2__)
+#        include <immintrin.h>
 #    endif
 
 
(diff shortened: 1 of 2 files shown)


include/plugin_interface/SC_SndBuf.h
```diff
@@ -39,9 +45,17 @@ class rw_spinlock {
...
+    static inline void pause() { YieldProcessor(); }
+#    elif defined(__SSE2__) /* all modern Intel processors */
     static inline void pause() { _mm_pause(); }
+#    elif defined(__aarch64__) /* 64-bit ARM */
+    static inline void pause() { __asm__ __volatile__("isb"); }
+#    elif defined(__arm__) /* 32-bit ARM */
+    static inline void pause() { __asm__ __volatile__("yield"); }
 #    else
+#        warning "unknown architecture: fall back to busy-waiting"
     static inline void pause() {}
```
JordanHendersonMusic, line 59: Why not use the compiler GCC/clang builtin  `__builtin_ia32_pause`?

Spacechild1: Why should I? `_mm_pause` does the same thing and is the more portable of the two.

dyfer:
This is amazing, thank you so much!
I tried the CI build on a Windows 11 machine and Supernova boots and makes sound. This is a big improvement :)

Does this mean that supernova plugins that use the `SC_SndBuf.h` stuff would need to be recompiled in order to work correctly? That's fine IMO, but _if that's the case_ we might want to be explicit about this, e.g. put this info in the commit message so that we remember to include it in the release notes.

Spacechild1:
> Does this mean that supernova plugins that use the SC_SndBuf.h stuff would need to be recompiled in order to work correctly? 

No worries, existing Supernova plugins keep working as expected!

I just discovered that our implementation of `rw_spinlock` had the same problem as `nova::detail::pause` where the `lock` function would accidentally busy-wait on MSVC and ARM platforms. The only problem is that it wastes power during the spin loop. You can recompile your plugins for these platforms to get the new, more efficient implementation, but you don't have to. (Of course, this only applies to plugins who make use of sound buffers in the first place.)

dyfer:
This sounds great, thank you for the explanation @Spacechild1 !

dyfer (approved):
This looks good to me!

include/plugin_interface/SC_SndBuf.h
JordanHendersonMusic: Ah, okay, only asking because I'd seen people write spin-locks with `__builtin_ia32_pause` and not `_mm_pause`.

Looking again, Rust does exactly what your code does too.
https://doc.rust-lang.org/1.80.0/src/core/hint.rs.html#217

JordanHendersonMusic:
All good :smile:

JordanHendersonMusic (approved):

PR #6403, opened 2024-07-27
add version and build to Platform.userAppSupportDir

prko:
## Purpose and Motivation

Currently, Platform.userAppSupportDir is fixed as follows:

- Window:
%USERPROFILE%\AppData\Local\SuperCollider
- Linux:
~/.local/share/SuperCollider
- macOS:
~/Library/Application Support/SuperCollider

This leads to developers and users running multiple SCs sometimes removing some subfolders of Platform.userAppSupportDir (at least the Help folder).

This PR adds version and build to Platform.userAppSupportDir:
- Window:
%USERPROFILE%\AppData\Local\SuperCollider\<version>\<build>
- Linux:
~/.local/share/SuperCollider/<version>/<build>
- macOS:
~/Library/Application Support/SuperCollider/<version>/<build>

## Types of changes

- Documentation
- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation
- [x] This PR is a draft PR, but comments or reviews are needed to finalise it.

### DIFF
--- a/README_WINDOWS.md
+++ b/README_WINDOWS.md
@@ -172,9 +172,9 @@ You will likely see (the grandpa referred to earlier):
 
     C:\Users\<USERNAME>\AppData
 
-The userAppSupportDir is in the subfolder `Local\SuperCollider`:
+The userAppSupportDir is in the subfolder `Local\SuperCollider\<version>\<build>`:
 
-    C:\Users\<USERNAME>\AppData\Local\SuperCollider
+    C:\Users\<USERNAME>\AppData\Local\SuperCollider\<version>\<build>
 
 This folder is user writable and *not* deleted by uninstalling. Therefore your
 settings, quarks, and plugins persist through updates. This is usually good,
--- a/common/SC_Filesystem_iphone.cpp
+++ b/common/SC_Filesystem_iphone.cpp
@@ -102,7 +102,7 @@ Path SC_Filesystem::defaultUserHomeDirectory() {
 Path SC_Filesystem::defaultUserAppSupportDirectory() {
     // Note: I have not added XDG support here because that seems highly unlikely on iPhone. -BH
     const Path& p = defaultUserHomeDirectory();
-    return p.empty() ? p : p / DOCUMENTS_DIR_NAME;
+    return p.empty() ? p : p / DOCUMENTS_DIR_NAME / SC_VersionString() / SC_BuildString();
 }
 
 Path SC_Filesystem::defaultUserConfigDirectory() {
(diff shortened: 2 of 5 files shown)


JordanHendersonMusic:
What happens when a user update there SC I install?

prko:
> What happens when a user update there SC I install?

The folders from the old SC version should be removed manually by the user.

JordanHendersonMusic:
So their existing stuff won't work on a newly downloaded version - even minor build change. I'm not a fan of that. I wonder what others think, is copying the previous version a good idea of the folder doesn't already exist?

capital-G:
I personally also feel hesitant about this as this would make the update procedures for all users more difficult - and as sclang rarely breaks anything there isn't such a motivation to versioning the whole quark environment.

I think most of his can be handled via a proper `sclang_conf.yaml` setup which will be passed to the CLI of sclang. Maybe we can introduce such a CLI flag for the IDE as well, so people can create shortcuts for different SC-IDE environments?

prko:
>So their existing stuff won't work on a newly downloaded version - even minor build change. I'm not a fan of that.

For each build, Quarks, Extension and all YALM files should be manually copied.
I can remove build info from the path!

--

> I think most of this can be handled via a proper sclang_conf.yaml setup which will be passed to the CLI of sclang. Maybe we can introduce such a CLI flag for the IDE as well, so people can create shortcuts for different SC-IDE environments?

It is a good idea.

--

When conceiving this idea, I only thought of the independent maintenance of SC. I did not think of the already configured setting.

How about moving some of the following items from `Platform.userAppSupportDir` to `Platform.sharedAppSupportDir`
- `downloaded-quarks` folder 
- `Extensions` folder
- `sclang_conf.yalm`
- `sc_ide_conf.yalm`

Since `Platform.sharedAppSupportDir` does not exist, it should be created in the corresponding cpp file.

capital-G:
> How about moving some of the following items from `Platform.userAppSupportDir` to `Platform.sharedAppSupportDir`

I think this could introduce problems during a version upgrade as it would break backwards-compability (say 3.14 would introduce such a directory and use it, a parallel installation of 3.13 and before would not pick this up - but should 3.14 still respect the files in  `userAppSupportDir`? This can lead to clashes :/)

I think the cleanest way would be to embrace more passing the YAML config file path via CLI approach - this would need to be checked if it actually works and also somehow documented for multiple platforms, so people could actually use this feature.

What is your opinion on it @prko ?

prko:
>I think the cleanest way would be to embrace more passing the YAML config file path via CLI approach

I also think this is a better idea!

PR #6406, opened 2024-07-30
update switch and case in /HelpSource/Reference/Control-Structures.schelp

prko:
## Purpose and Motivation
There are some wrong indentations in code snippets and missing explanations on `.case` and `.switch` `/HelpSource/Reference/Control-Structures.schelp`. 
This PR corrects the indentation and adds some explanations to help users (especially newbies) understand `.case' and `.switch'.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


JordanHendersonMusic:
I'm not sure that the indentation is wrong, you need to check against the style guide and ignore the built-in indenter. 

The guide doesn't actually say anything about this case, but it does talk of sequential method calls
```
Button()
    .states_([["blorp", nil, nil]])
    .action_({
        "hey hey hey".postln
    });
```
Based on this, I think the original was correct. Curious about what others think!

capital-G:
I am also unsure about the indentation here - they look like tabs to me which [should be switched to 4 spaces according to the style guide](https://capital-g.github.io/supercollider-wiki/contributing/style/Style-Guidelines%3A-SuperCollider.html#id1), but most of our code is indented with tabs?

I don't care too much about code styles, but I am interested in consistency, but regarding e.g. the `case` changes there seems no consistency in the code base :/

https://github.com/supercollider/supercollider/blob/d52b0522b210def4fe47829ca332f0bc411b3509/SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc#L108-L111

but there is also

https://github.com/supercollider/supercollider/blob/d52b0522b210def4fe47829ca332f0bc411b3509/SCClassLibrary/Common/GUI/iphone/Base/SCView.sc#L983-L987

and this version, which could be seen as a violation of the [*Use trailing closure syntax whenever possible*](https://capital-g.github.io/supercollider-wiki/contributing/style/Style-Guidelines%3A-SuperCollider.html#recommendation-use-trailing-closure-syntax-whenever-possible) recommendation.

https://github.com/supercollider/supercollider/blob/d52b0522b210def4fe47829ca332f0bc411b3509/SCClassLibrary/Common/Control/HIDFunc.sc#L54-L57

If @prko thinks this improves the consistency of the source code layout I am fine with this, but then

* it should be written down as a rule or suggestion in our styling guidelines and 
* be checked if our naive "auto-indentation" workflow from the IDE doesn't interfere with this suggestion.

In the end, having non-exhausting formatting rules without an automatic formatter is just really frustrating :/

JordanHendersonMusic:
> but most of our code is indented with tabs

The editor defaults to tabs. I thought I read that c++ was spaces and sc was tabs, but cant find the text now.

Perhaps we should avoid touching contentious formatting issues until a proper automatic formatter can be made? Then reformat everything in one commit?
 
I was going to work on this, but have decided to spend time on synthdef compilation as it seems more pressing.

capital-G:
> > but most of our code is indented with tabs
> 
> The editor defaults to tabs. I thought I read that c++ was spaces and sc was tabs, but cant find the text now.

Sorry, you are right - the cited block refers to files other than sc(d) files - maybe we should not put it under the sc formatting guidelines in this case?

> Perhaps we should avoid touching contentious formatting issues until a proper automatic formatter can be made? Then reformat everything in one commit?
> 
> I was going to work on this, but have decided to spend time on synthdef compilation as it seems more pressing.

IMO this would be a really great addition!

On the other hand, I think code within the docs should not need to obey such formatting rules as they could be formatted in a particular manner to make a point? So the changes of this PR would still be relevant?

Maybe we can create a new issue in which we collect all style changes which are currently ambiguous but should be decided through a style guideline in the future once such a formatter is in the works, so we can integrate the new proposed rules from a collected resource?

prko:
I restored the code style as it was.

I am not sure about the indentation of the first part of the added code for the `.case' syntax.

JordanHendersonMusic:
```
switch(value,
        testvalue1, trueFunction1,
        testvalue2, trueFunction2,
        ...
        testvalueN, trueFunctionN,
        defaultFunction // defaultFunction can be omitted
);

switch(value,
        {testvalue1}, {trueFunction1},
        {testvalue2}, {trueFunction2},
        ...
        {testvalueN}, {trueFunctionN},
        {defaultFunction} // defaultFunction can be omitted
);
```
Okay here is a really oddity to consider...

In #3567 I showed that these two ways of writing the function will actually call different **methods** on the value object. The former will actually go and call value's `switch` method, but the later will do a compiler optimisation that won't call the method. This means the behaviour ***could*** actually differ (it doesn't in the class library ...*so far*... ).

I do not know what should be recommended here, it is just a mess and impossible to fix without a large performance hit. Perhaps, it is fine and we just ignore it, bury our heads in the sand, deny its existence...
Or perhaps we remove all reference to the first syntax in the docs and only show the second, while stating that some of these 'core' methods cannot be overridden.

prko:
The inline-related explanation is what I do not understand. 

I think the current explanation and structure could be simpler and easier:
- `switch` has a long explanation.
- `case` has only one sentence.
- `if` and `while` have no explanation.
- There is a section at the end of the help file: Inline optimisation

I am not sure if I understand this correctly, but it would be better if `if`, `while`, `switch` and `case` had a note like this in the discussion:
```
note:: .... will be inline if .... See link::# Inline optimisation for .::

JordanHendersonMusic:
Yea, after sleeping on it, I think its best to ignore that issue, at least here. This should be a simple introduction to control structures.

capital-G (approved):
I think this adds value to the documentation - lets wait for another pair of eyes to merge it though.

prko:
Is there anything that needs correcting?
Could this PR be approved and merged before the 3.14 release?

dyfer:
This looks straightforward to me. Merging. Thanks!

PR #6409, opened 2024-08-01
ClassLib: fix Set.remove not returning the item

### LINKED ISSUE #6408: ClassLib: Set.remove returns the collection when the docs say it should return the item.
## Environment

* SuperCollider version: dev

## Steps to reproduce

```supercollider

Set[1, 2, 3].remove(3); // should return 3.
[1, 2, 3].remove(3) // returns 3
Set[1, 2].remove(3) // returns Set[1, 2]
[1, 2].remove(3) // returns nil
```

This is clearly mentioned in the docs of `Collection`.

JordanHendersonMusic:
## Purpose and Motivation

Fixes #6408 

```supercollider
[1, 2, 3].remove(3) //  3
Set[1, 2, 3].remove(3) // Set[1,2]  << incorrect


[1, 2].remove(3) // nil
Set[1, 2].remove(3) // Set[1, 2] << incorrect
```

From `Collection` help file
> Remove anObject from the receiver. Answers the removed object.

## Types of changes

- Bug fix

Could be a breaking bug fix, but documentation is clear

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Collections/Bag.sc
+++ b/SCClassLibrary/Common/Collections/Bag.sc
@@ -17,7 +17,7 @@ Bag : Collection {
 		});
 	}
 	remove { arg item, count=1;
-		var newCount;
+		var newCount, out;
 		if ( this.includes(item), {
 			newCount = contents.at(item) - count;
 			if (newCount <= 0, {
@@ -25,7 +25,9 @@ Bag : Collection {
 			},{
 				contents.put(item, newCount);
 			});
+			^item
 		});
+		^nil
 	}
 
 	// accessing
--- a/SCClassLibrary/Common/Collections/Bag.sc
+++ b/SCClassLibrary/Common/Collections/Bag.sc
@@ -17,7 +17,7 @@ Bag : Collection {
 		});
 	}
 	remove { arg item, count=1;
-		var newCount, out;
+		var newCount;
 		if ( this.includes(item), {
 			newCount = contents.at(item) - count;
 			if (newCount <= 0, {
--- a/SCClassLibrary/Common/Collections/SequenceableCollection.sc
+++ b/SCClassLibrary/Common/Collections/SequenceableCollection.sc
@@ -350,10 +350,12 @@ SequenceableCollection : Collection {
 		^this.species.fill(newSize, { |i| this.blendAt(i * factor) })
 	}
 
-	remove { arg item;
+	removeAt { |index| ^this.subclassResponsibility(thisMethod) }
+	remove { |item|
 		var index = this.indexOf(item);
 		^if ( index.notNil, {
 			this.removeAt(index);
+			^item
 		},{
 			nil
 		});
--- a/SCClassLibrary/Common/Collections/SequenceableCollection.sc
+++ b/SCClassLibrary/Common/Collections/SequenceableCollection.sc
@@ -355,7 +355,7 @@ SequenceableCollection : Collection {
 		var index = this.indexOf(item);
 		^if ( index.notNil, {
 			this.removeAt(index);
-			^item
+			item
 		},{
 			nil
 		});
--- a/SCClassLibrary/Common/Collections/SequenceableCollection.sc
+++ b/SCClassLibrary/Common/Collections/SequenceableCollection.sc
@@ -350,7 +350,7 @@ SequenceableCollection : Collection {
 		^this.species.fill(newSize, { |i| this.blendAt(i * factor) })
 	}
 
-	removeAt { |index| ^this.subclassResponsibility(thisMethod) }
+	removeAt { |index| this.subclassResponsibility(thisMethod) }
 	remove { |item|
 		var index = this.indexOf(item);
 		^if ( index.notNil, {
--- a/SCClassLibrary/Common/Collections/Array2D.sc
+++ b/SCClassLibrary/Common/Collections/Array2D.sc
@@ -43,8 +43,9 @@ Array2D : Collection {
 		^array.copyRange(ri * cols, ri * cols + cols - 1)
 	}
 
-	// overide Array
-	// add { ^thisMethod.shouldNotImplement }
+	add { ^this.shouldNotImplement(thisMethod)  }
+	remove { ^this.shouldNotImplement(thisMethod) }
+
 	printOn { arg stream;
 		// not a compileable string
 		stream << this.class.name << "[" ;
--- a/SCClassLibrary/Common/Collections/Interval.sc
+++ b/SCClassLibrary/Common/Collections/Interval.sc
@@ -20,6 +20,7 @@ Interval : Collection {
 		forBy(start, end, step, function);
 	}
 
+	remove { ^this.shouldNotImplement(thisMethod) }
 	add { ^this.shouldNotImplement(thisMethod) }
 	put { ^this.shouldNotImplement(thisMethod) }
 	storeArgs { ^[start, end, step] }
@@ -50,6 +51,7 @@ Range : Collection {
 		^(val >= start) and: { (val < this.end)  and: { val.frac == 0 }}
 	}
 
+	remove { ^this.shouldNotImplement(thisMethod) }
 	add { ^this.shouldNotImplement(thisMethod) }
 	put { ^this.shouldNotImplement(thisMethod) }
 
--- a/SCClassLibrary/Common/Collections/LinkedList.sc
+++ b/SCClassLibrary/Common/Collections/LinkedList.sc
@@ -64,14 +64,16 @@ LinkedList : SequenceableCollection {
 		});
 		size = size + 1;
 	}
-	remove { arg obj;
+	remove { |obj|
 		var node = this.findNodeOfObj(obj);
 		if ( node.notNil, {
 			if (head == node, { head = node.next; });
 			if (tail == node, { tail = node.prev; });
 			node.remove;
 			size = size - 1;
+			^obj
 		});
+		^nil
 	}
 	pop {
 		var node;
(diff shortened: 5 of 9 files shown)


capital-G (commented):
Makes sense and matches [the docs](https://docs.supercollider.online/Classes/Collection.html#-remove), but maybe add a unit test for this as well?

JordanHendersonMusic:
Okay I found a bunch of other classes that are broken. I've added some tests that attempt to check all of Collection's subclasses. It isn't exhaustive, but the ones that it checks now pass.

SCClassLibrary/Common/Collections/Bag.sc
```diff
@@ -17,15 +17,17 @@ Bag : Collection {
 		});
 	}
 	remove { arg item, count=1;
-		var newCount;
+		var newCount, out;
```
capital-G, line 20: why is the variable `out` here? It dosen't seem to be used.

capital-G (approved):

SCClassLibrary/Common/Collections/Bag.sc
JordanHendersonMusic: My mistake

telephon (approved):

JordanHendersonMusic:
@capital-G is this good to merge? 

The tests might fail if someone adds a class that inherits from Collection that doesn't implement these methods correctly, but that shouldn't be an issue.

telephon:
It is good to enforec a bit of semantic consistency here.

JordanHendersonMusic:
Additionally — in another PR — we could actually test that **all** classes that have a 'subclass responsibility' method in their parent (and no child classes) actually implement it.

telephon:
> Additionally — in another PR — we could actually test that **all** classes that have a 'subclass responsibility' method in their parent (and no child classes) actually implement it.

It would be good if there was an easy way to test it. I am not so sure if we should really enforce a complete implementation, it is more a way to throw the right error whenever someone happens to try to call the method. Incomplete interfaces are fine (I think). But I am open to discussion.

JordanHendersonMusic:
> It would be good if there was an easy way to test it. I am not so sure if we should really enforce a complete implementation, it is more a way to throw the right error whenever someone happens to try to call the method. Incomplete interfaces are fine (I think). But I am open to discussion.


I kind disagree with this because it causes confusion.

Getting the list of methods that throw a subclass responsibility method that don't have child classes can be done like this.
```
(
var subclassMethodCode = Collection.findMethod(\do).code;
Class.allClasses.do { |c| 
	if (c.subclasses.size == 0) {
		var methods = ();
		c.collectMethods(methods);
		methods.do { |m|
			if (m.isNil.not){
				if (m.code == subclassMethodCode) {
					"%:% has not been implemented".format(c.name, m.name).postln
				}
			}
		}
	}
}
)
``` 

It returns the following list...
```
Archive:isAssociationArray has not been implemented
Array2D:remove has not been implemented
Array2D:add has not been implemented
Array2D:isAssociationArray has not been implemented
BinaryOpStream:put has not been implemented
BinaryOpXStream:put has not been implemented
CSVFileReader:put has not been implemented
CleanupStream:put has not been implemented
Complex:mod has not been implemented
Complex:div has not been implemented
ControlSpec:defaultControl has not been implemented
EmbedOnce:put has not been implemented
EventStreamPlayer:put has not been implemented
FuncStream:put has not been implemented
FuncStreamAsRoutine:put has not been implemented
IdentityBag:isAssociationArray has not been implemented
Interval:remove has not been implemented
Interval:isAssociationArray has not been implemented
Library:isAssociationArray has not been implemented
NAryOpStream:put has not been implemented
ObjectTable:isAssociationArray has not been implemented
OneShotStream:put has not been implemented
OrderedIdentitySet:isAssociationArray has not been implemented
Pair:remove has not been implemented
Pair:add has not been implemented
Pair:isAssociationArray has not been implemented
Pipe:pos_ has not been implemented
PmonoArticStream:put has not been implemented
PmonoArticStream:next has not been implemented
Polar:mod has not been implemented
Polar:< has not been implemented
Polar:div has not been implemented
Polar:pow has not been implemented
Pretty:next has not been implemented
Pretty:pos_ has not been implemented
Range:remove has not been implemented
Range:isAssociationArray has not been implemented
SemiColonFileReader:put has not been implemented
StreamClutch:put has not been implemented
TabFileReader:put has not been implemented
Task:put has not been implemented
UnaryOpStream:put has not been implemented
```

One thing to note immediately  is that isAssociationArray isn't define on many classes — this is probably a mistake because its a boolean, but it doesn't have any documentation.

The issue this PR addresses can also be seen in the list above where `remove` hasn't been implemented.

IMO, if a class doesn't implement a method, it should explicitly use `this.shouldNotImplement`. This way, we can search for it.

One example of where we can apply this is in `MultiLevelIdentityDictionary` where the method `remove` is implemented as `remove { ^this.shouldNotImplement(thisMethod) }` yet the help documentation entry is still there and says...
```
.[removeAt]
Remove only the item located by the **path.
**
```

telephon:
> One thing to note immediately is that isAssociationArray isn't define on many classes — this is probably a mistake because its a boolean, but it doesn't have any documentation.

Depending on how we interpret "responsibility" in the term "subclassResponsibility", we get two different ways to implement stuff for an unknown argument. 

1. responsibility means duty: if possible, cover all subclasses with a default behaviour in a class higher up. `subclassResponsibility` is only needed where this is not possible.
2.  responsibility means freedom: this is something the superclass can't decide, so if needed, implement.


## isAssociationArray

The case of `isAssociationArray` is a little opaque. It belongs to the `asAssociations`:

````
Set[1, 2, 3].asAssociations;
// ERROR: 'isAssociationArray' should have been implemented by Set.
```

The alternative could be to implement `isAssociationArray` to return `false` on `Object` or `Collection`, and then "whitelist" it wherever possible. Then we would get an error if any class doesn't implement `pairsDo`.

```supercollider
asAssociations { |class|
		var res;
		class = class ? Array;
		if(this.isAssociationArray) {
			^if(class == this.class) { this } { this.as(class) }
		};
		res = class.new(this.size div: 2);
		this.pairsDo { |key, val| res = res.add(key -> val) }
		^res
	}
```

We would then get an error `Set: pairsDo not understood`.

JordanHendersonMusic:
> responsibility means freedom: this is something the superclass can't decide, so if needed, implement.

In this case the child class could implement it with `doesNotImplement` this way the freedom is still there, but it is explicit. The child is opting out, rather than opting for the default (which is unknown).

dyfer:
@telephon @capital-G is this good to be merged?

capital-G:
For me, it is up to @telephon to decide - feel free to merge if you think it is in a good state.

SCClassLibrary/Common/Collections/SequenceableCollection.sc
```diff
@@ -350,10 +350,12 @@ SequenceableCollection : Collection {
 		^this.species.fill(newSize, { |i| this.blendAt(i * factor) })
 	}
 
-	remove { arg item;
+	removeAt { |index| ^this.subclassResponsibility(thisMethod) }
+	remove { |item|
 		var index = this.indexOf(item);
```
telephon, line 355: the inner caret (^) should be removed

telephon (changes requested):

SCClassLibrary/Common/Collections/SequenceableCollection.sc
JordanHendersonMusic: I've done this, but there is loads of code that already does this.

SCClassLibrary/Common/Collections/SequenceableCollection.sc
```diff
@@ -350,10 +350,12 @@ SequenceableCollection : Collection {
 		^this.species.fill(newSize, { |i| this.blendAt(i * factor) })
 	}
 
-	remove { arg item;
+	removeAt { |index| this.subclassResponsibility(thisMethod) }
+	remove { |item|
 		var index = this.indexOf(item);
 		^if ( index.notNil, {
 			this.removeAt(index);
+			item
```
JordanHendersonMusic, line 358: @telephon did you mean this line? I've removed it, thanks!

dyfer:
@telephon is this good to be merged now?

telephon (approved):

capital-G:
@telephon As you are one of the council members of the class library it is somehow unclear (at least to me) why this hasn't been merged after your approval or what would be the missing steps towards merging from your POV.

For further formalism see https://github.com/supercollider/supercollider/wiki/Council-structure#who-is-merging-prs

> If you approve a PR but do not merge it please leave a comment what are the missing steps towards a merge (e.g. "looks fine to me, but I'd like to have an additional review by X who can then also merge this PR.") so it is always clear what the next step is.

Is this okay to be merged as is or would you prefer an additional review?

telephon:
I haven't requested another review. There ad been just one thing pending which was resolved, and I approved.

So far, the idea was that the person who approves does not merge (only in trivial cases). Has this changed?

telephon:
> > If you approve a PR but do not merge it please leave a comment what are the missing steps towards a merge (e.g. "looks fine to me, but I'd like to have an additional review by X who can then also merge this PR.") so it is always clear what the next step is.
> 
> Is this okay to be merged as is or would you prefer an additional review?

So I assume that if there are no additional comments, it is ready to be merged.

JordanHendersonMusic:
I think this is good. Just worth acknowledging this is potentially a breaking bug fix.

The tests are maybe a bit overkill for this PR, but that isn't necessary a bad thing.

dyfer:
I marked this as "API change" so that we note this in the changelog.

@telephon please merge at your leisure, thanks!

PR #6415, opened 2024-08-09
String literal size limit: Documentation and bugfix

jamshark70:
## Purpose and Motivation

1. String literals have a size limit, enforced by the compiler. To my knowledge, this has never been documented before. So, here it is.

2. The lack of documentation has come up as an issue particularly for large strings in text archives (see https://scsynth.org/t/large-strings-truncated-by-object-readarchive/9894). I.e., if aString.size > 8188, then `aString.asCompileString` produces a string that does not compile back to the original string -- hence asCompileString is broken in that case. This PR fixes the bug by splitting long strings into this type of expression: `["substr0", "substr1", "substr2"...].join`.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/Literals.schelp
+++ b/HelpSource/Reference/Literals.schelp
@@ -217,6 +217,22 @@ a.size // 9
 The SuperCollider IDE uses UTF-8 to decode and display strings.
 See link::Classes/String#Character encodings##String:: for more information.
 
+A string literal may comprise no more than 8188 characters (bytes). A
+string object can be larger as a result of string operations.
+
+code::
+// artificially generate a long string
+a = String.fill(9000, { "abcdefghijklmnopqrstuvwxyz".choose });
+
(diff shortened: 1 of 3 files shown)


jamshark70:
PS I also checked escape characters:

```
a = Array.fill(6000, "\\c").join;
a.size  // 12000

("\"" ++ a ++ "\"").interpret.size
-> 6000
```

So escape chars added by asCompileString do not count toward the 8188 limit -- so `clump` is valid in the bugfix.

dyfer:
At the dev meeting we decide to merge this, but also increase the literal size in a new PR. The new PR should increase the literal size, as well as the documentation, classlib check, and tests.

dyfer (approved):

Issue #6410, opened 2024-08-01
TestUGen_RTAlloc:test_allUGens relies on being able to fill the server's realtime memory (probably source of CI bugs)


JordanHendersonMusic:
## Description

This test assumes the server's real time memory is exactly as stated and does include any padding.

## Environment
* SuperCollider version: current

## Explanation

Once you unpack the code it boils down to the following (removing synchronisation to make things clearer)...
```supercollider
{ AllpassC.ar(DC.ar(1), 1, 0) }.loadToFloatArray(0.001, s) { |out|
	out.sum != 0  << first test
};

{ LocalBuf(1024 * 1000 / 4) }.play;

{ AllpassC.ar(DC.ar(1), 1, 0)  }.loadToFloatArray(0.001, s) { |out|
	out.sum == 0  << second test
};
```

I believe the point here is to fill the server's real time memory with the buffer so the allpass allocationfails.

There are several things wrong with this.
1. The server's memory does not exactly have to be as reported, in can be more.
2. The `LocalBuf` UGen doesn't have any side effects and would be completely reasonable for the synthdef compiler to remove it #6405.

JordanHendersonMusic:
@capital-G do you think my reading of the code is correct?

capital-G:
The tests are indeed a bit opaque, e.g. what is this syntax? A method call followed by a function?

https://github.com/supercollider/supercollider/blob/d52b0522b210def4fe47829ca332f0bc411b3509/testsuite/classlibrary/TestUGen_RTAlloc.sc#L166-L170

Maybe I should go back to SC Syntax school :D

Nonetheless, I'll try to put it into my words from what I understand by repeating.

IIUC the idea is to spawn stateful UGens which allocate memory - first time with enough memory available, the 2nd time w/o enough memory.

By using delaytime values such as `0.0` they should simply bypass the signal the original signal in this case, so if we have `DC.ar(1.0)` as input, the sum of the rendered/returned float array should be > 0.

In the 2nd case we allocate some memory on the server, with the means of an `LocalBuf` which consumes the whole memory of the server.
If we now spawn such a stateful UGen, it should fail (because of mem alloc error) and the UGen should return 0.0.

I tried to reproduce this locally, but I seem to have some wrong calculations concerning RT memory?

```supercollider

// the setup
{DC.ar(0.0)}.loadToFloatArray(duration: 0.01, action: {|a| a.sum.postln;})
[...]

PR #6429, opened 2024-08-27
Disable RTAlloc test in CI

capital-G:
## Purpose and Motivation

Our CI has been broken for a few weeks due to an RTAlloc test. Since no one has come up with a fix yet (see #6410), and since this breaks the CI and thus any further PRs, I'd suggest disabling this test for now. 

## Types of changes

- Bug fix (well, not really...)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/scripts/test_run_proto_gha.scd
+++ b/testsuite/scripts/test_run_proto_gha.scd
@@ -123,4 +123,10 @@
 		'skip': "true",
 		'skipReason': "Requires ALSA kernel module, not allowed in GHA",
 	),
+	(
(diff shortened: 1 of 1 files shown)


dyfer (approved):
I'm okay disabling them for now, unless @elgiano feels we should leave them in (or better yet - has an idea how to fix them).

capital-G:
I will merge this now, as it seems that other tests are already starting to fail because they are being drowned in the review process by these failing tests.

PR #6433, opened 2024-08-30
Add additional sync for `test_synthDef_isReleased_afterFree`

capital-G:
## Purpose and Motivation

See https://github.com/supercollider/supercollider/pull/6213#issuecomment-2320933394 - but I must admit I am not too sure how the test works.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestNodeProxy_Server.sc
+++ b/testsuite/classlibrary/TestNodeProxy_Server.sc
@@ -57,6 +57,10 @@ TestNodeProxy_Server : UnitTest {
 	test_synthDef_isReleased_afterFree {
 		var numBefore, numAfter;
 
+		// make sure state is synced
+		// see https://github.com/supercollider/supercollider/pull/6213#issuecomment-2320933394
+		server.sync;
+
 		// here we have to make sure that the statusWatcher.numSynthDefs is correct
 		// before the experiment, so we force the otherwise regular update ...
 		server.sendStatusMsg;


capital-G:
@JordanHendersonMusic this implements your suggested fix and pipeline seems to pass first try.

The problem popped up elsewhere again after rebase on another PR, see https://github.com/supercollider/supercollider/actions/runs/10632378565/job/29475401061?pr=5618#step:7:3469

JordanHendersonMusic:
I suppose this is probabilistic so we don't *know* this is the solution. 

My suspicion was that, since this test checks the number of nodes before and after the test, nodes like volume which are created through `ServerBoot.add` might not have been created by the time this function is entered. This actually points to a bug/issue with the interplay between `Server.bootSync` and `ServerBoot`.

capital-G:
@telephon can you give some educated guess if this could solve the experienced issue?
What @JordanHendersonMusic writes makes sense to me.

capital-G:
Mh, seems this is not the solution as the CI in my repo on this PR failed, see  https://github.com/capital-G/supercollider/actions/runs/10632244106/job/29477345070#step:7:3435 and https://github.com/capital-G/supercollider/actions/runs/10632244106

JordanHendersonMusic:
Although adding `server.sync` is wrong, I think my diagnosis might still be correct. 

What we want to do in this case (the test is using `Server.bootSync`) is wait until the server is booted  *and* `ServerBoot` has finished, that way, we get a stable node count.

Here is some code from `Volume` that is called inside of `ServerBoot`.
```supercollider
sendSynthDef {
	fork {
		SynthDef(...).send(server);
		...
		server.sync;
		...
	}
}
```
This boils down to the following example...
```supercollider
fork {
	// boot server
	// start ServerBoot
	// Volume's func
	fork {
		s.sync;
		\FinishedServerBoot.postln;
	};
	
	// the test
	s.sync;
	\DoingTest.postln;
}
```

What we want to see is `\FinishedServerBoot` then `\DoingTest`, but the actual order depends on how the threads are woken up. On my machine, I always get `\DoingTest` then `\FinishedServerBoot`. I don't know if there is a defined or predictable ordering, but this isn't something to be relied upon.

One way to check this would be replacing the `server.sync in test_synthDef_isReleased_afterFree with `2.wait` (or something), on my machine, this seems to work, but because it's infrequent, I can't say for certain.

capital-G:
Closing as the small fix seems not sufficient. I extracted the relevant information into issue #6438

Issue #6435, opened 2024-09-03
fix `incompatible-pointer-types` warnings


Spacechild1:
## Environment

* SuperCollider version: latest develop (31a4bc07a16311a5ae28fd1f233167a822b048b3)
* Operating system: Windows, Msys2
* Other details (Qt version, audio driver, etc.): compiler GCC 14.1.0 (Msys2)

GCC 14 has upgraded the `-Wincompatible-pointer-types` warning to an error. As a consequence, some (external) C code in SC fails to compile, namely `portmidi` and `hidapi`:

```
C:/Repos/supercollider/external_libraries/portmidi/pm_win/pmwinmm.c:1052:31: error: assignment to 'uint32_
t *' {aka 'unsigned int *'} from incompatible pointer type 'DWORD *' {aka 'long unsigned int *'} [-Wincomp
atible-pointer-types]
 1052 |         midi->fill_offset_ptr = &(hdr->dwBytesRecorded);
      |                               ^
C:/Repos/supercollider/external_libraries/hidapi/hidapi_parser/hidapi_parser.c: In function 'hid_parse_ele
ment_info':
C:/Repos/supercollider/external_libraries/hidapi/hidapi_parser/hidapi_parser.c:1630:64: error: passing arg
ument 2 of 'HidP_GetLinkCollectionNodes' from incompatible pointer type [-Wincompatible-pointer-types]
 1630 |     nt_res = HidP_GetLinkCollectionNodes( linkCollectionNodes, &numColls, pp_data );
      |                                                                ^~~~~~~~~
      |                                                                |
      |                                                                int *
```

As a temporary fix we might pass `-Wno-error=incompatible-pointer-types` so that we still get the warning, but it's not treated as an error. Ideally, the issue should be fixed upstream.

---

Also, there is another compilation error in `SC_PAUtiles` because of a missing include of `<cstdint>`:

```
C:/Repos/supercollider/common/SC_PaUtils.cpp: In function 'PaError TryGetDefaultPaDevices(PaDeviceIndex*,
PaDeviceIndex*, int, int, double)':
C:/Repos/supercollider/common/SC_PaUtils.cpp:130:21: error: 'uint32_t' was not declared in this scope
  130 |                 if (uint32_t(inSR) != uint32_t(outSR)) {
      |                     ^~~~~~~~
C:/Repos/supercollider/common/SC_PaUtils.cpp:6:1: note: 'uint32_t' is defined in header '<cstdint>'; this
[...]

Issue #6430, opened 2024-08-29
Using `-latest` Github Runner images in CI


dyfer:
## Motivation

@capital-G recently inquired about using the `ubuntu-latest-lts` runner image for sc3-plugins. 
I had considered this before and I wanted to share my thoughts here, with the idea that's applicable to both for SC and sc3-plugins.

## Description of Proposed Feature

I'd like to consider whether we want to switch some of our build jobs to use the `-latest` runner images, as opposed to fixed versions, which we use currently.

Pros of fixed version:
- build environment does not change out of our control (well, mostly, e.g. on macOS homebrew libraries still do get updated...)
- releases are built using consistent tools

Cons of fixed version:
- we need to periodically manually update the build matrix to check compatibility with newer systems
- if we wait too long and the runner image gets deprecated/removed, build jobs stop working (as was the case with sc3-plugins recently)

Pros of `-latest` version:
- **easier testing against new/current OSes and toolchains** - build environment automatically switches to the newest available OS (newest in the sense of their availability in GHA)
- theoretically there's no need to manually update the build environment once OS runner versions get updated
  - however, sometimes some build details still need to be adjusted for newer OS; i.e. build matrix might break when the new image is released
  - but that also forces us to update compatibility not only in our GHA setup, but also documentation and possibly other areas of the project(s)

Cons of `-latest` version:
- we might be forced to fix our builds when the runner images update (i.e. timing of that would be controlled by GHA updates, not developers availability)
- if used for release builds, it would possibly cause switching the build environment in the middle of the release process

Implied above, but to spell it out: I think there are two slightly different goals with the CI: 
1. testing build environment
2. creating release and testing binaries

I think it would be reasonable for `1.` to target the newest OSes and `2.` to use fixed image versions.

Finally, the changes in GHA have different timing for different OS environments: Windows seems to be updated every 3 years, Linux every 2 years (coinciding with Ubuntu LTS releases) and macOS every year (for every new major macOS version).

## Plan for Implementation

My proposition is to keep fixed OS versions for the main release builds for macOS and Windows, but to add `-latest` images for testing build envifornment.

I think the relevance of this change is different for each platform, due to different timing of their updates.

macOS seems the most affected, as it changes most frequently. I think we could switch the "use system libraries" build to run on the `latest` macOS in order to check compatibility with the newest provided toolchains (this also broke most recently). This can only happen after we upgrade boost (see #5865 and linked/related PRs)

[...]

PR #6436, opened 2024-09-03
Fix for GCC 14

Spacechild1:
fixes https://github.com/supercollider/supercollider/issues/6435

Please merge ASAP :)

### DIFF
--- a/server/scsynth/CMakeLists.txt
+++ b/server/scsynth/CMakeLists.txt
@@ -54,7 +54,7 @@ elseif(AUDIOAPI STREQUAL portaudio AND SYSTEM_PORTAUDIO)
       message(FATAL_ERROR "Portaudio selected as audio API, but development files not found")
     endif()
 elseif(AUDIOAPI STREQUAL bela)
-    # The Bela lib will have its own set of libraries and includes
+    # The Bela lib will have its own set of libraries and includes
     # However, the Bela audio backend for SC has some direct calls to Xenomai
     # Therefore we need to get flags for those as well
     find_package(Xenomai REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,9 +103,13 @@ if (${CMAKE_COMPILER_IS_GNUCXX})
     )
 
     add_compile_options(
+        -Werror=return-type      # Error out if function definitions lack a return type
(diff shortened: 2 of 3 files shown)


JordanHendersonMusic (approved):
Thanks for the speedy work :)

capital-G (changes requested):
Thanks for the quick fix!

In general, I'd prefer to do it "properly" by fixing our dependencies if it's not too much work for now.

* It seems our portmidi dependency hasn't been updated since [9 years](https://github.com/supercollider/supercollider/commits/develop/external_libraries/portmidi) - maybe we should update it as it's still maintained, see https://github.com/PortMidi/portmidi ? while we're at it, maybe we can make it a git submodule instead of copying files around to keep our history/repo a bit cleaner?
* Can you estimate how much effort it would take to fix hidapi? IIUC it is a fork maintained by SC, so we could quickly in-cooperate any PRs.

Additionally, if we decide to go with the fix, we should add gcc14 to our linux build pipeline (maybe replacing gcc7/gcc9?). If you need help with this, I can help.
Maybe @dyfer can also provide some input here, as we recently talked about upgrading the versions used in our CI, see https://github.com/supercollider/supercollider/issues/6430

If there is a consensus that this is urgent and should be merged without fixing upstream dependencies - the proposed changes look good to me.

Spacechild1:
> In general, I'd prefer to do it "properly" by fixing our dependencies if it's not too much work for now.

Sure, that's why I said it's a temporary fix. The issue is urgent because developers who have upgraded to GCC 14, like myself, are currently not able to build SC.

capital-G (commented):
OK, in this case let's merge it once the pipeline passes.

Spacechild1:
Thanks :)

server/scsynth/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ elseif(AUDIOAPI STREQUAL portaudio AND SYSTEM_PORTAUDIO)
       message(FATAL_ERROR "Portaudio selected as audio API, but development files not found")
     endif()
 elseif(AUDIOAPI STREQUAL bela)
-    # The Bela lib will have its own set of libraries and includes
+    # The Bela lib will have its own set of libraries and includes
```
dyfer, line 57: If we consider this PR, this change probably shouldn't be here :)

Issue #6439, opened 2024-09-03
Speed up unit tests


capital-G:
Unit tests are best when they are fast, so they can be run often. But a current run of the unit tests take over 10 minutes, maybe it is possible to improve the runtime of the test frameworks with some clever tricks?
If you have some idea how to reduce their runtime, please feel free to create a PR - it is greatly appreciated!

<details><summary>How to analyze the runtimes</summary>
There is a python script which automates the reports, it is located @ <a href="https://gist.github.com/capital-G/253bdb368345a63f5e1180c13a00234e">https://gist.github.com/capital-G/253bdb368345a63f5e1180c13a00234e</a>.
Select a run and obtain its "View raw logs URL" (by actually clicking on it, as the browser will then generate a URL which is only valid for a limited amount of time). Pass it as a parameter or set the URL as a variable - the script will generate two html files which can be pasted directly into GH.
<img src="https://github.com/user-attachments/assets/4648a568-6b0d-4a8c-ac7c-e44bbdf7da11"/>
</details>

This post will be updated occasionally to get an overview - @ me me if you want a "re-run" of the calculations.

Logs from 2024-09-15 @ https://github.com/supercollider/supercollider/commit/381513e6dc504e6ddef60d05598b7c6fa84d52c1, based on https://github.com/supercollider/supercollider/actions/runs/10865580837/job/30152892774

## Top 50 tests

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>file_name</th>
      <th>test_name</th>
      <th>time</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>21759</th>
      <td>TestBus</td>
      <td>test_controlFree</td>
      <td>13.76</td>
    </tr>
    <tr>
      <th>22780</th>
      <td>TestCoreUGens</td>
      <td>test_muladd</td>
      <td>10.93</td>
    </tr>
    <tr>
      <th>19074</th>
      <td>TestSerialPort</td>
      <td>test_rxErrors_bufferOverflow</td>
      <td>8.87</td>
    </tr>
    <tr>
      <th>17948</th>
      <td>TestPV_ChainUGen</td>
      <td>test_whiteframe_noop</td>
      <td>7.93</td>
    </tr>
    <tr>
      <th>22660</th>
      <td>TestCoreUGens</td>
      <td>test_ugen_generator_equivalences</td>
      <td>6.61</td>
    </tr>
    <tr>
[...]

PR #6440, opened 2024-09-03
Speed up PV_UGens tests by sharing server

capital-G:
## Purpose and Motivation

Currently, the PV tests take about 100 seconds in CI (see #6439). This PR reduces the PV tests to ≅ 10 seconds by sharing the server for all PV tests instead of starting and killing the server for each test. I think this approach is valid as the worst case scenario would be a false report for PV_UGens that run after a failed test due to a corrupted state somehow, but I think this would then be inspected further and the runtime benefits outweigh this problem.

I also had problems with the CondVar, so I added a predicate to them, which seems to have helped them run "as fast as possible" instead of waiting for timeout.

### New
https://github.com/supercollider/supercollider/actions/runs/10687410343/job/29625594482#step:7:6196

```
TestPV_ChainUGen:test_whiteframe_noop (8.26s)
TestPV_ChainUGen:test_whiteframe_zeroing (3.89s)
```

### Old
https://github.com/supercollider/supercollider/actions/runs/10681804483/job/29606801735#step:7:6201

```
TestPV_ChainUGen:test_whiteframe_noop (80.07s)
TestPV_ChainUGen:test_whiteframe_zeroing (24.77s)
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestPV_ChainUGen.sc
+++ b/testsuite/classlibrary/TestPV_ChainUGen.sc
@@ -4,6 +4,8 @@ UnitTest.gui
 */
 TestPV_ChainUGen : UnitTest {
 
+classvar <>server;
+
 // This test manually creates an FFT frame equivalent to pure white noise, 
 // on which certain ugens' action should be a no-op
 test_whiteframe_noop {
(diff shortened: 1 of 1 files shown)


telephon:
Is there a reason why these tests need to restart the server for each one? It may be enough to free all synths after each test, right?

capital-G:
Closed in favor of https://github.com/supercollider/supercollider/pull/6460

PR #6441, opened 2024-09-03
Speed up CoreUGens muladd test

capital-G:
## Purpose and Motivation

Reduces time from around 25 seconds to around 2 seconds by removed random wait time and lower the rendered time from 0.1 seconds to 0.01 seconds.

Please review if the random time or the longer length is actually necessary here. Tests still seem to run.

### New
https://github.com/supercollider/supercollider/actions/runs/10687826107/job/29626830696?pr=6441#step:7:5528
```
TestCoreUGens:test_muladd (2.73s)
```

### Old
https://github.com/supercollider/supercollider/actions/runs/10681804483/job/29606801735#step:7:5529

```
TestCoreUGens:test_muladd (25.4s)
```

## Types of changes

- Bug fix
- Performance fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -368,7 +368,6 @@ TestCoreUGens : UnitTest {
 				completed = completed + 1;
 				condvar.signalOne;
 			});
-			rrand(0.06, 0.15).wait;
 		};
 
 		condvar.waitFor(1, { completed == tests.size });
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -363,7 +363,7 @@ TestCoreUGens : UnitTest {
 
 		server.bootSync;
 		tests.keysValuesDo{|name, func|
-			func.loadToFloatArray(0.1, server, { |data|
+			func.loadToFloatArray(0.01, server, { |data|
 				this.assertArrayFloatEquals(data, 0, name.quote, report: true);
 				completed = completed + 1;
 				condvar.signalOne;


telephon (commented):
one block is 64 / 44100, and these are DC tests, so `0.002` would be enough. I don't know what the random wait is for, maybe to avoid a peak out in the server?

mtmccrea:
In this same source file I've tried to make the number of samples or blocks explicit by something like

`numBlocksToTest * server.options.blockSize / server.sampleRate`

so it isn't a magic number and is sample rate independent.

mtmccrea:
Since this has already been merged, I can update the test dur in #6374 if you'd like.

JordanHendersonMusic:
@capital-G 

Just had this fail 

https://github.com/supercollider/supercollider/actions/runs/10698105192/job/29657016686?pr=6446#step:6:12956

capital-G:
@JordanHendersonMusic I also observe this now ocassionally 😞  I think the whole test framework needs some inspection, tests clearly are not atomic currently as the state of one test can affect another test. I really don't know what to do about it :/

Maybe we can merge #6374 which may fix this more properly as its rebased CI runs smoothly?

mtmccrea:
A key difference is that #6374 is also using server syncing to throttle tests, removing the need for magic waittimes.

This PR removed the wait time entirely (line 371), which isn't stable.

JordanHendersonMusic:
@mtmccrea do you know why it isn't stable without a wait time?

JordanHendersonMusic:
@capital-G seems that this only happens on Linux, just happened to @Spacechild1 parseOsc pr, Mac test was also fine when I had this occur.

capital-G:
No, I also observed fails on macOS - it is nondeterministic when this fails. sorry for the screw up :/

capital-G:
From https://github.com/supercollider/supercollider/actions/runs/10702297513/job/29682290517?pr=6151#step:6:12832
```
 FAIL: a TestCoreUGens: test_muladd - "kr0.madd(ar0, kr5)"
23999 of 24000 items in array failed to match. Displaying arrays from index of first failure (0) onwards:
FloatArray[0.2691615819931, 0.22649741172791, 0.51876485347748, 0.032893538475037, 0.22922086715698, 0.32748711109161, 0.14557492733002, 0.11669838428497, 0.044102191925049, 0.95873892307281, 0.062782049179077, 0.97345650196075, 0.24437856674194, 0.93069016933441, 0.36867296695709, 0.59128546714783, 0.98839592933655, 0.99936056137085, 0.38911628723145, 0.47886669635773, 0.18049764633179, 0.40171933174133, 0.97072958946228, 0.4852169752121, 0.23992097377777, 0.47950732707977, 0.96206617355347, 0.274365186691...etc...
! = 
0
File '/tmp/1149184133' could not be opened: Format not recognised.
FAIL: a TestCoreUGens: test_bufugens - ERROR: Message '-' not understood.
	Test: test_impulse
TestCoreUGens : setting clientID to 0.
WARNING: Server name already exists: 'TestCoreUGens'. Please use a unique name
Booting server 'TestCoreUGens' on address 127.0.0.1:57110.
JackDriver: client name is 'SuperCollider-01'
JackDriver: connected  system:capture_1 to SuperCollider-01:in_1
JackDriver: connected  system:capture_2 to SuperCollider-01:in_2
JackDriver: connected  SuperCollider-01:out_1 to system:playback_1
JackDriver: connected  SuperCollider-01:out_2 to system:playback_2
*** ERROR: failed to open UDP socket: address in use.
You can use SuperCollider (sclang) to kill all running servers by running `Server.killAll`.
You can also kill scsynth using a terminal or your operating system's task manager.
Server 'TestCoreUGens' exited with exit code 1.
```

There seems to be multiple failures here:
* Why is the array != 0? It seems many tests are writing directly to `Out.ar(0)`, maybe we should rewrite them to a bus-based approach so they can run more isolated? But if other tests are still running in parallel or leave a tainted state, there is something else already wrong in the way we run tests.
* Why is there a `tmp` file, and why can't it be opened?
* What is the message `-`?
* Why is there a server dangling around?

I think the way we run tests needs some more investigation, I have the feeling that much of our time is spend on it :/

dyfer:
> I think the way we run tests needs some more investigation, I have the feeling that much of our time is spend on it :/

Yes...

Just FYI in case that's useful - some of the unpredictable failures can be recreated locally when running test locally on a system under load, see https://github.com/supercollider/supercollider/issues/5360#issuecomment-1601615272

JordanHendersonMusic:
If you look above that in the log, is says that the TestCoreUGens has completed, then reports these specific sub test failures. I wonder if this a bug with condvar when the system is underload? Perhaps the condition is triggering incorrectly somehow?

capital-G:
> If you look above that in the log, is says that the TestCoreUGens has completed, then reports these specific sub test failures. I wonder if this a bug with condvar when the system is underload? Perhaps the condition is triggering incorrectly somehow?

I also experienced problems with CondVar in https://github.com/supercollider/supercollider/pull/6440 - but I am not sure if I simply don't understand the details of CondVar yet fully. But I found adding a predicate more reliable, often `signalOne` does not seem to release the `.wait`, with an added predicate this seems to work, but I don't know the price of adding the predicate - maybe it is firing to soon now?

Spacechild1:
> But I found adding a predicate more reliable

You absolutely *must* use a predicate when waiting on a `CondVar`! (The help file really should be more clear about this! Ideally, with a big fat warning.) The predicate can either be "internal" (passed as a function to `wait`) or "external" (part of your own while loop)

capital-G:
Ah, this is somehow counter-intuitive if one expects a `await` like syntax! Thanks for clarification.

So IIUC e.g. 

https://github.com/supercollider/supercollider/blob/09a212b5b010442854dcd30e232e5361010c43b9/testsuite/classlibrary/TestPV_ChainUGen.sc#L73-L85

would not work "properly" and always wait for 3 seconds?
I found when adding a predicate (in #6440 through a boolean variable `finished`) we don't always wait for 3 seconds.

Why is there even an option to use CondVar without a predicate then?

telephon:
No, in both cases it will not wait for the predicate if you don't pass one:

```supercollider


(
fork {
	var c;
	c = CondVar.new;
	fork { 1.wait; c.signalOne };
	c.wait;
	"it has happened".postln;
}
)


(
fork {
	var c;
	c = CondVar.new;
	fork { 1.wait; c.signalOne };
	c.waitFor(10);
	"it has happened".postln;
}
)

```

Both happen after 1s.

telephon:
> You absolutely _must_ use a predicate when waiting on a `CondVar`! Otherwise, calls to `signal` can get lost.

I don't quite understand why.

JordanHendersonMusic:
I believe it is in case the signal happens before you get to the wait - in this case it hangs forever.

Spacechild1:
> would not work "properly" and always wait for 3 seconds?

Yes. This is because `signalOne` is called before `waitFor`, so the latter will block until the timeout. That's exactly why you need the predicate.

Now, you might ask yourself: isn't `loadToFloatArray` an asynchronous method? How can it execute *before* `c.waitFor`?

The reason is that `loadToFloatArray` actually blocks when it's called inside a Routine! (It uses `forkAsNeeded` under the hood.) *Of course* the documentation does not mention this...

The solution is simple:
```
fork {
    b.loadToFloatArray(action: { |data| 
        this.assertArrayFloatEquals(data, equaltothis, message.format(name), within: 0.001, report: true); 
        c.signalOne; 
     });
};
c.waitFor(3); 
```

> Ah, this is somehow counter-intuitive if one expects a await like syntax! Thanks for clarification.

I see. `CondVar` is modelled after condition variables, which is a thread synchronization primitive. It has nothing to do with async/await.

This might be a bit confusing for people coming from `Condition`, which acts more like a semaphore in the sense that signals don't get lost.

> Why is there even an option to use CondVar without a predicate then?

Because the predicate might be external to the `wait` function. You can check the documentation for examples.

telephon:
> The reason is that `loadToFloatArray` actually blocks when it's called inside a Routine! (It uses `forkAsNeeded` under the hood.) _Of course_ the documentation does not mention this...

It is only recently that I have realised that `forkAsNeeded` may cause subtle issues. Unfortunately, changing it to `fork` will also have side effects. So yes, something that needs documentation!

Spacechild1:
> https://github.com/supercollider/supercollider/issues/6448

FWIW I have opened a ticket: https://github.com/supercollider/supercollider/issues/6448

PR #6443, opened 2024-09-03
Add additional waiting time in `TestNodeProxy_Server:test_synthDef_isReleased_afterFree`

### LINKED ISSUE #6438: Improve success consistency of `test_synthDef_isReleased_afterFree`
Currently many test runs fail in CI due to `test_synthDef_isReleased_afterFree` , e.g. in https://github.com/supercollider/supercollider/actions/runs/10649343960/job/29543125921#step:7:3469

```
 	[0/1]	TestNodeProxy_Server:test_synthDef_isReleased_afterFree (2.19s)
```

Jordan has an idea how to fix this problem

---

              Although adding `server.sync` is wrong, I think my diagnosis might still be correct. 

What we want to do in this case (the test is using `Server.bootSync`) is wait until the server is booted  *and* `ServerBoot` has finished, that way, we get a stable node count.

Here is some code from `Volume` that is called inside of `ServerBoot`.
```supercollider
sendSynthDef {
	fork {
		SynthDef(...).send(server);
		...
		server.sync;
		...
	}
}
```
This boils down to the following example...
```supercollider
fork {
	// boot server
	// start ServerBoot
	// Volume's func
	fork {
		s.sync;
		\FinishedServerBoot.postln;
	};
	
	// the test
	s.sync;
	\DoingTest.postln;
}
```

What we want to see is `\FinishedServerBoot` then `\DoingTest`, but the actual order depends on how the threads are woken up. On my machine, I always get `\DoingTest` then `\FinishedServerBoot`. I don't know if there is a defined or predictable ordering, but this isn't something to be relied upon.

One way to check this would be replacing the `server.sync` in test_synthDef_isReleased_afterFree with `2.wait` (or something), on my machine, this seems to work, but because it's infrequent, I can't say for certain.

_Originally posted by @JordanHendersonMusic in https://github.com/supercollider/supercollider/issues/6433#issuecomment-2321768779_

capital-G:
## Purpose and Motivation

This is rather unstable test and it is unclear to me why it occasionally fails (see https://github.com/supercollider/supercollider/actions/runs/10649343960/job/29543125921#step:7:3469).
I now added a waiting time in case the server needs some time for garbage collection or else? If this still fails, there should be a waiting time added in the beginning of the test.

I am unsure if this PR actually fixes the underlying issue, but at least it provides some additional information about the state of the test through printing.

If everything goes well, this fixes #6438 

Generally these tests could also run on the same server, instead of booting and killing it for every test - see #6440 (max 50 seconds to gain here).

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestNodeProxy_Server.sc
+++ b/testsuite/classlibrary/TestNodeProxy_Server.sc
@@ -68,13 +68,17 @@ TestNodeProxy_Server : UnitTest {
 		server.sync;
 		proxy.source = nil;
 
+		// ... we give the server some time to catch up? ...
+		// see https://github.com/supercollider/supercollider/issues/6438
+		0.1.wait;
+
 		// ... and also afterwards, we force an update
 		server.sync;
 		server.sendStatusMsg;
 		numAfter = server.statusWatcher.numSynthDefs;
 
 		this.assertEquals(numBefore, numAfter,
-			"Removing the  NodeProxy source function should remove SynthDef from server"
+			"Removing the NodeProxy source function should remove SynthDef from server (was %, is %)".format(numBefore, numAfter)
 		);
 	}
 


capital-G:
Has someone a clue why this seems to break now another test? :/

```
 	[1/2]	TestServer_clientID_booted:test_remoteLogin (1.84s)
		 * after first login, remote client should have its requested clientID.
		 ! after first login, remote client should be known to be fully running.
```

Fails on

* this PR: https://github.com/supercollider/supercollider/actions/runs/10688273732/job/29628490240#step:7:6071 
* CI from my fork: https://github.com/capital-G/supercollider/actions/runs/10687772986/job/29627798778#step:7:6071

telephon:
I see, the problem behind this seems to be that there is no `done` OSC message sent when a SynthDef is freed from the server. So there is nothing to wait for in `s.sync`, and the sync is immediately:

```supercollider


OSCFunc({ |...args| "--done message: --".postln; args.postln }, "/done");
Ndef(\x, { Silent.ar });

(
fork {
	Ndef(\x).source = nil;
	s.sync;
	"done".postln;
}
)

```

capital-G:
So the proposed waiting time in between both states to compare seems like a valid idea here? If this is the case we should add it as a comment for our future self's and others.

But I am still trying to understand why https://github.com/supercollider/supercollider/blob/31bf288b777f70a64e6267f04dbe9e29da4fd86a/testsuite/classlibrary/TestServer_clientID_booted.sc#L131-L162 seems to break now because of this test :/

One thing that seems strange on this clientID test: doesn't `self.assert` "panic" on a false value (because continuing in a "false/unintended" state could be dangerous), therefore the killing of the PID is never performed when this test fails which is not good. IIUC sclang doesn't have a `finally` block on `try catch` that could be normally used here.

Should we discuss this new failing test here or should this be another PR/Issue?

capital-G:
Haven't seen this in a while, so probably not.

Issue #2823, opened 2017-03-30
scide -> recent files doesn't work


ycollet:
I use supercollider compiled from sources on fedora 25 64 bits.
When I click on a file in the "recent file" list in scide, the file is not opened.

snappizz:
Does this happen even when the file has not been moved or deleted?

ycollet:
The file I try to open has not been deleted.

But supercollider is installed in a non usual directory: 
/opt/supersollider-devel

I also have supercollider-3.7.2 installed as a standard fedora package.

snappizz:
Is the problem consistently reproducible if you save a new file?

ycollet:
Yes, the same problem happens if I create a new file via scide saves it and then try to reopen it via recent files menu.

patrickdupuis:
I don't have a Fedora machine to test on at the moment. I'm pretty certain that I've used 'recent files' without problem in the past. Could be related to your particular system.

ycollet:
I just noticed something today (after total git clone + submodule update + recompilation): when I try to open /home/me/test.sc via recent file menu, a message is printed in the status bar:
" Cannot open file /&home/me/test.sc (file doesn't exist)" ...

ycollet:
Informations related to qt5 installed on Fedora 24:

Nom         : qt5-qtbase
Architectur : x86_64
Époque      : 0
Version     : 5.6.2
Révision    : 3.fc24

patrickdupuis:
Tested on Fedora 25. Scide could create, save, and open file from 'Recent Files' menu when installed into /usr/local and /opt/supercollider. I could not get the interpreter to run when built into /opt/supercollider, however (that's another story).

Try removing the sc3.7.2 package from your system to see if that helps.

ycollet:
I removed all the supercollider packages from my linux box.
I opened scide, enter some text, saves it, restart scide and try to reopen the saved file via recent files and it doesn't work.
I found a message in the console:
```
QSortFilterProxyModel: index from wrong model passed to mapFromSource
```

PR #6450, opened 2024-09-05
scide: fix failing to open recent files if mnemonic is enabled

sonata-chen:
## Purpose and Motivation

I encountered the similar error message described in https://github.com/supercollider/supercollider/issues/2823#issuecomment-294336058

This pull request fixes https://github.com/supercollider/supercollider/issues/2823 by storing file path in `QAction`'s data field.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/main_window.cpp
+++ b/editors/sc-ide/widgets/main_window.cpp
@@ -849,8 +849,10 @@ void MainWindow::updateRecentDocsMenu() {
 
     const QStringList& recent = mMain->documentManager()->recents();
 
-    foreach (const QString& path, recent)
-        mRecentDocsMenu->addAction(path);
+    foreach (const QString& path, recent) {
+        QAction* action = mRecentDocsMenu->addAction(path);
+        action->setData(QVariant(path));
+    }
 
     if (!recent.isEmpty()) {
         mRecentDocsMenu->addSeparator();
@@ -858,7 +860,9 @@ void MainWindow::updateRecentDocsMenu() {
     }
 }
 
-void MainWindow::onOpenRecentDocument(QAction* action) { mMain->documentManager()->open(action->text()); }
+void MainWindow::onOpenRecentDocument(QAction* action) {
+    mMain->documentManager()->open(action->data().value<QString>());
+}
 
 void MainWindow::onInterpreterStateChanged(QProcess::ProcessState state) {
     switch (state) {


sonata-chen:
image


From Qt [Document](https://doc.qt.io/qt-6/qaction.html#text-prop):
> Certain UI elements, such as menus or buttons, can use '&' in front of a character to automatically create a mnemonic (a shortcut) for that character. 

So `/data/Text File.txt` is actually stored as `/&data/Text File.txt`. This causes an error when opening that file, since the actual filename does not contain a `&`.

This pull request fixes it by storing the file path in `QAction`'s **data** field. So we don't need to worry about an `&`   being inserted into `QAction`'s **text** field.

capital-G (dismissed):
Thanks for this fix, well spotted!

All changes look fine to me, but let's wait for another pair of eyes to merge it.

dyfer:
Thanks! 
The implementation looks fine to me, however...

I have a question: do mnemonics need to be somehow enabled?

As this seems to work on Linux (judging from @sonata-chen's screenshot), I wanted to check other platforms.
I don't see mnemonics on macOS, so I checked the Windows build. There, the mnemonics show up, but only for some actions; there doesn't seem to be a mnemonic for "Open Recent":
Screenshot 2024-09-12 at 8 00 45 PM
Is this behavior Linux-specific, or is it somehow settable somewhere for other OSes?
FWIW, opening recent files works fine on macOS and Windows, with or without this change.

capital-G:
On Fedora I also can't see these underlines

image

I agree with @dyfer that a consistent behavior of this across all platforms is desirable, but I think such an implementation is out of scope of this PR as this is about a bug fix in case those shortcuts exist.
If you think the implementation is fine @dyfer , should we merge it?

sonata-chen:
To be honest, I am also not sure how to control the behaviour of mnemonic. I am on Arch linux. I **can't** reproduce this error using Qt5 either. However, it happens when Qt6 is used.

capital-G:
> To be honest, I am also not sure how to control the behaviour of mnemonic. I am on Arch linux. I **can't** reproduce this error using Qt5 either. However, it happens when Qt6 is used.

How do you use Qt6? I thought we always bundle Qt5 statically in the build? IIUC SuperCollider is not (yet) compatible with Qt6? But maybe I don't understand the Qt5 build process too well.

As mentioned, I think we should focus within this PR on a fix of the stated issue, not on the compatibility of other OSes.

sonata-chen:
>  I think we should focus within this PR on a fix of the stated issue, not on the compatibility of other OSes.

totally agree with you. For Qt6, I am using dyfer's branch.

dyfer:
Just so it's clear, I wasn't suggesting to make this behavior cross-platform as part of the PR, but rather I didn't know what I was missing when trying to test this.

It's a minor fix and I can't see it being in any way wrong, so we could merge this now. 
However, I personally feel that if this indeed can be observed only in qt6, it would be marginally better to wait for our qt6 migration, then rebase this PR and see how it looks "in the wild" in the qt6 builds from GHA.

capital-G:
Then let's block it and wait for the qt6 branch :)
I'll add a qt6 milestone because there are quite some PRs/Issues attached to it and the original project is not available anymore?

dyfer:
I dismissed the review so that we don't accidentally merge this.

sonata-chen:
It seems that it also happens on Qt 5.6.2 ( https://github.com/supercollider/supercollider/issues/2823#issuecomment-294336358) But I didn't find a way to reproduce it on Qt 5.15.

I tried to use this [function](https://doc.qt.io/qt-5/qkeysequence.html#qt_set_sequence_auto_mnemonic) to enable mnemonic, but there is a linking error on Arch linux. Maybe we can try this function on other platforms.

BTW, my desktop environment is KDE plasma.

sonata-chen:
Screenshot_20240916_150751

I'm now pretty sure we won't encounter these errors on other platforms. The behaviour of  automatically inserting an `&` into `text` is a **KDE** "feature".

One can disable this feature by adding these two lines in `~/.config/kdeglobals`.

```
[Development]
AutoCheckAccelerators=false
```

Ref: https://api.kde.org/frameworks/kxmlgui/html/classKCheckAccelerators.html

dyfer:
All right, so this indeed seem like a KDE "feature": https://bugs.kde.org/show_bug.cgi?id=337491 It does not seem like it has been addressed yet though. IIUC correctly some of the comments suggest that the feature that inserts `&` should be enabled only for KDE apps, not for all Qt apps.

I have mixed feelings whether we want to be providing a workaround for it. OTOH the change is fairly small and seems non-invasive... so I'm split on this one.

dyfer:
@sonata-chen how to you think we should proceed with this? I have a feeling that providing this workaround is not a desirable addition to SC codebase... but if you feel that this is valid addition, we could pull this in.

sonata-chen:
@dyfer I understand this is actually a KDE bug, still hoping we can include this. It makes sense to do so given that `QText` has a `data` field.

dyfer:
All right @sonata-chen, would you mind rebasing it on `develop` so that CI builds it with Qt6?

dyfer (approved):
Let's pull this in as it fixes the behavior under KDE and doesn't seem to disturb the functionality on other platforms.

dyfer:
Thanks @sonata-chen !

PR #6453, opened 2024-09-07
testsuite: RTAlloc refactor and fix

### LINKED ISSUE #6410: TestUGen_RTAlloc:test_allUGens relies on being able to fill the server's realtime memory (probably source of CI bugs)

elgiano:
## Purpose and Motivation

Fixes #6410 

These tests are meant as an exhaustive (as much as possible) coverage of UGens using RTAlloc. They have been failing more or less unpredictably and got recently disabled (https://github.com/supercollider/supercollider/pull/6429).
I'm working on it to hopefully get these tests back, and I'm mainly making this PR to test on CI (if anyone knows a better method, please let me know!).

Refactoring and trying to make CI pass for these tests, especially the test_allUGens ones:
- refactored away async logic, hopefully it gets a bit more readable now
- changed strategy: instead of filling RT memory with a LocalBuf, we try to make each UGen fail by itself by allocating too much memory

And if anyone has any better idea to test RTAlloc pass/fail behavior, that would be welcome as well!
(see also: https://github.com/supercollider/supercollider/pull/5713#issuecomment-1056171574)
## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestUGen_RTAlloc.sc
+++ b/testsuite/classlibrary/TestUGen_RTAlloc.sc
@@ -9,8 +9,11 @@ TestUGen_RTAlloc : UnitTest {
 	*setUpClass {
 		server = Server(this.name);
 		server.options.sampleRate = 48000;
-		server.options.memSize = 1024;
 		server.options.blockSize = 64;
+		server.options.memSize = 2 ** 13; // scsynth default
+		// - tests fail with memSize < 256 (GVerb fails)
+		// - testing with memSize >= 2 ** 20 would require multiple allocations
+		//   (allocation size itself can overflow size_t for some UGens)
 		server.bootSync;
 	}
 
@@ -22,7 +25,12 @@ TestUGen_RTAlloc : UnitTest {
 
 	awaitSynthOutput { |synthFn, runtimeSamples=1, timeout=1|
 		var cond = CondVar(), synthOutput = nil;
-		synthFn.loadToFloatArray(runtimeSamples/server.sampleRate, server) { |out|
+		var runtimeSeconds = runtimeSamples / server.sampleRate;
+		// allow extra timeout for longer tests (e.g. testFFT with big memSize)
+		if (runtimeSeconds > timeout) { 
+			timeout = runtimeSeconds + timeout
+		};
+		synthFn.loadToFloatArray(runtimeSeconds, server) { |out|
 			synthOutput = out;
 			cond.signalOne();
 		};
@@ -30,7 +38,7 @@ TestUGen_RTAlloc : UnitTest {
 		^synthOutput;
 	}
 
-	assertAllocPass { |name, func, runtimeSamples=(server.options.blockSize)|
+	assertAllocPass { |name, func, runtimeSamples|
 		var out = this.awaitSynthOutput(func, runtimeSamples);
 		if (out.isNil) {
 			this.assert(false, "% allocPass test should complete".format(name))
@@ -41,7 +49,7 @@ TestUGen_RTAlloc : UnitTest {
 		};
 	}
 
-	assertAllocFail { |name, func, runtimeSamples=(server.options.blockSize)|
+	assertAllocFail { |name, func, runtimeSamples|
 		var out = this.awaitSynthOutput(func, runtimeSamples);
 		if (out.isNil) {
 			this.assert(false, "% allocFail test should complete".format(name))
@@ -52,11 +60,17 @@ TestUGen_RTAlloc : UnitTest {
 		};
 	}
 
+	// mem calc helpers
+
+	memSizeKb { ^server.options.memSize }
+	memSizeFloats { ^this.memSizeKb * (1024 / 4) }
+	memSizeSeconds { ^this.memSizeFloats / server.sampleRate }
+
 	// Tests
 
 	test_allocFail_clearUnit {
 		var output = this.awaitSynthOutput {
-			DelayN.ar(DC.ar(1), 100)
+			DelayN.ar(DC.ar(1), this.memSizeSeconds * 2)
 		};
 		this.assertEquals(output[0], 0.0,
 			"a failed RTAlloc should clear unit outputs")
@@ -64,7 +78,7 @@ TestUGen_RTAlloc : UnitTest {
 
 	test_allocFail_continueProcessing_sameNode {
 		var output = this.awaitSynthOutput {
-			DelayN.ar(DC.ar(1), 100); DC.ar(1)
+			DelayN.ar(DC.ar(1), this.memSizeSeconds * 2); DC.ar(1)
 		};
 		this.assertEquals(output[0], 1.0,
 			"a failed RTAlloc should not block other units in the same node")
@@ -72,7 +86,7 @@ TestUGen_RTAlloc : UnitTest {
 
 	test_allocFail_continueProcessing_otherNodes {
 		var output = nil;
-		var memFill = { DelayN.ar(Silent.ar, 100) }.play(server);
+		var memFill = { DelayN.ar(Silent.ar, this.memSizeSeconds * 2) }.play(server);
 		server.sync;
 		output = this.awaitSynthOutput { DC.ar(1) };
 		memFill.free;
@@ -82,7 +96,7 @@ TestUGen_RTAlloc : UnitTest {
 
 	test_allocFail_setDoneFlag {
 		var output = this.awaitSynthOutput {
-			K2A.ar(Done.kr(DelayN.ar(Silent.ar, 100)))
+			K2A.ar(Done.kr(DelayN.ar(Silent.ar, this.memSizeSeconds * 2)))
 		};
 		this.assert(output[0] > 0,
 			"a failed RTAlloc should set the Done flag")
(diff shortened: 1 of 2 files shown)


capital-G:
> I'm working on it to hopefully get these tests back, and I'm mainly making this PR to test on CI (if anyone knows a better method, please let me know!).

Are you aware of this insight? https://github.com/supercollider/supercollider/issues/5360#issuecomment-1601615272 Maybe it helps you during investigation, best of luck! And thanks for taking a look at this, much appreciated!

elgiano:
@capital-G thanks, I didn't know that! I'm going to try!

So far it seems like all tests pass, even the FFT/PV ones, except one because of a minor bug with a UGen (opened a specific PR: https://github.com/supercollider/supercollider/pull/6454)

JordanHendersonMusic:
I wasn't able to recreate that - but it might be worth switching over to TCP for testing?

elgiano:
What would it be the advantage of going \tcp?

I usually fancy it when I send lots of udp osc messages and some can get lost... for example when allocating lots of small buffers... but in these tests I'm not seeing network problems, so why have you thought of tcp?

Sent from Proton Mail Android

JordanHendersonMusic:
Wouldn't that aid in situations where the CPU is under load - if that is one cause of test failures that occurs when running on ci. Unless I've misunderstood, which is possible!

capital-G:
One thing I don't understand with the TCP/UDP - we are **not** sending a ton of messages during testing, and scsynth is normally capable of accepting 100s of thousands messages per second.

Though we use lots of `server.sync`, so maybe this creates some internal locking issues?

Also: Maybe GitHub runners are not configured like "normal" machines regarding its network stack (because it could get abused?). Maybe we should check for/verify/inspect things like [`sysctl net.core.rmem_max`](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/5/html/tuning_and_optimizing_red_hat_enterprise_linux_for_oracle_9i_and_10g_databases/sect-oracle_9i_and_10g_tuning_guide-adjusting_network_settings-changing_network_kernel_settings). This could explain why we sometimes observe which looks like UDP message when the tests run too fast?

elgiano:
Are your networking considerations related to other (perhaps more general) CI issues I'm not aware of? I don't have an overview of all our UnitTests and CI history, but:
- I don't see evidence of networking problems here
- I've tested RTAlloc tests many times under heavy load (100% CPU) and never had a failure
- CI for this PR has only one failure, the SpecPcile bug I address in the other PR

So, unless anyone has better ideas on how to test these RTAlloc, for me it's good enough for now!

Note: I've reduced passVerbosity to \brief, to output less details when tests pass, and moved to using a single server for all tests, thus avoiding wasting time in rebooting it between different tests.

capital-G:
I recently managed to bring down unrelated tests by removing some wait time https://github.com/supercollider/supercollider/pull/6441 - it was fixed by a more sophisticated approach (#6374). It is unclear why these unrelated tests failed, but I think that's the reason we try to reason about potential issues with the testing suite.

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -70,210 +96,154 @@ TestUGen_RTAlloc : UnitTest {
...
-
-		ugenTests.keysValuesDo {|name, func|
-			testAllocPass.value(name.asString, func, 0.001);
-		};
+	test_allUGens_allocPass {
+
+		var memSizeKb = server.options.memSize;
+		var memSizeFloats = memSizeKb * 1024 / 4;
+
+		this.assertAllocPass("AllpassC", { AllpassC.ar(DC.ar(1), 1, 0) });
```
capital-G, line 105: although you have calculated `memSizeFloats` you don't use it and still rely on the magic $1$. Maybe we can make sure these tests pass with a variety of memSize?

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -70,210 +96,154 @@ TestUGen_RTAlloc : UnitTest {
...
+			b.free;
+		}.value;
+	}
 
+	test_allUGens_allocFail {
+
+		var memSizeKb = server.options.memSize;
+		var memSizeFloats = memSizeKb * 1024 / 4;
+
+		this.assertAllocFail("AllpassC", { AllpassC.ar(DC.ar(1), 100) });
```
capital-G, line 168: same as above, make the failing of the test dependent of memSizeFloat, so we could run the tests with different memSizes at some point.

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -70,210 +96,154 @@ TestUGen_RTAlloc : UnitTest {
...
-		var completed = nil, cond = CondVar();
-		var runtime = 2**9 / server.sampleRate / 4;
-		{ FFT(LocalBuf(2**9), DC.ar(1), 0.25) }.loadToFloatArray(runtime, server) { |out|
-			this.assert(out.any(_ > 0), "should return a valid bufnum");
-			completed = true;
-			cond.signalOne;
-		};
-		cond.waitFor(runtime + 1) { completed.notNil };
-		if (completed.isNil) {
+		var out = this.awaitSynthOutput({ FFT(LocalBuf(2**8), DC.ar(1), 0.25) }, 2**6);
```
capital-G, line 234: maybe add a small comment that 2**6 is due the hop size being `0.25` (IIUC?)

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -70,210 +96,154 @@ TestUGen_RTAlloc : UnitTest {
...
+		if (out.isNil) {
 			this.assert(false, "successful alloc test should complete");
+		} {
+			this.assert(out.any(_ > 0), "should return a valid bufnum (bufnum: %)".format(out));
```
capital-G, line 238: why is `out` related to the bufnum here? isn't it testing if any of the output samples are `>0`?

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -285,42 +255,34 @@ TestUGen_RTAlloc : UnitTest {
 		var bigBuf = Buffer.alloc(server, 2**19);
```
capital-G, line 255: some more magic numbers, but already present in the old code

capital-G (commented):
Thanks, the test code looks so much cleaner now! Maybe it would be good to make the tests dependent on the memSize instead of using magic numbers?

Otherwise the tests looks fine to me.

JordanHendersonMusic:
@capital-G Ithink I figured that one out. It seems that UDP messages were getting dropped in the loop. There were a few hundred made very quickly (there is another issue about this on GitHub) - if you switch to TCP it worked (on machine at least). Hence my previous comment about just blindly switching all tests to TCP.

There isn't a tight loop here in this PR, everything is serial. So that doesn't apply. But performance might be an issue, how long does this take to run?

capital-G:
> @elgiano do you know what the performance difference is between the old and this approach? There's a parallel effort to make the tests faster.

Using:https://github.com/supercollider/supercollider/issues/6439#issuecomment-2336476375

Branch | Aggregated runtime of `TestUGen_RTAlloc` in seconds
--- | ---
This PR |  9.63s
[Latest dev run](https://github.com/supercollider/supercollider/actions/runs/10755499258/job/29827337134) | 11.63s

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,65 +1,91 @@
...
+		this.assertEquals(output[0], 1.0,
+			"a failed RTAlloc should not block other units in the same node")
+	}
 
-		cond.waitFor(1, { output != -1 });
-		this.assertFloatEquals(output, 1.0, "a failed RTAlloc should not block other nodes in the processing chain")
+	test_allocFail_continueProcessing_otherNodes {
+		var output = nil;
+		var memFill = { DelayN.ar(Silent.ar, 100) }.play(server);
+		server.sync;
```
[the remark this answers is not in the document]
elgiano: isn't it what the previous test (`test_allocFail_continueProcessing_sameNode`) does?

testsuite/classlibrary/TestUGen_RTAlloc.sc
elgiano: Yeah, maybe a bit obscure, but FFT normally outputs  a stream of -1. When FFT has processed a frame, which typically takes more than a blockSize, on that block it will output the bufnum instead of -1. It's a way for PV_UGens to know that a new spectral frame is ready and at the same time get the bufnum.
So in this case, out is a stream of -1 and a bufnum > 0. That's why `out.any(_>0)`

testsuite/classlibrary/TestUGen_RTAlloc.sc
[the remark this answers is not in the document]
capital-G: Well, that's the same node, but what about a node running in parallel? But most likely you are right, if the same node does not fail, another node will not fail as well.

testsuite/classlibrary/TestUGen_RTAlloc.sc
capital-G: Ah, thank you! There is always stuff to learn and explore in SC.
Maybe we can add your comment as a comment in a code?

```suggestion
			// FFT normally outputs a stream of -1. When FFT has processed a frame, that block
			// will output the bufnum instead of -1.
			this.assert(out.any(_ > 0), "should return a valid bufnum (bufnum: %)".format(out));
```

testsuite/classlibrary/TestUGen_RTAlloc.sc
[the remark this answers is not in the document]
elgiano: I think we are not understanding each other here :D I mean:
- test_allocFail_continueProcessing_sameNode: there are two UGens in the same Node, one fails allocation, and the other should continue unaffected
- test_allocFail_continueProcessing_otherNodes: two separate Node, one fails allocation, and the other should continue unaffected

are these the cases you would like to cover, or are there some others cases I'm missing? 
love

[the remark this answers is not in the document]
capital-G: > test_allocFail_continueProcessing_otherNodes: two separate Node, one fails allocation, and the other should continue unaffected

Yes, but isn't the `s.sync` "delaying" the spawning of the 2nd node, so we don't actually test that on the same block but on the successive block? It is really nit-picky, so I am fine with the current way.

elgiano:
Updates:
- rebased on latest develop
- removed magic numbers as much as possible, commented the ones remaining
- directed memFill output to a dummy bus

So, I think I've addressed all of the above discussions.

About testing with different memSizes, I've added some comments:
we are now testing with scsynth default `2**13`. GVerb crashes below 256. Around 2**20 and up, we start to have other crashes, sometimes due to the size of a single memory block allocation overflowing size_t (long int).

Some UGens crash the server when trying to allocate enormous amounts of memory... fixing this would require a separate issue, and separate tests that for now would fail.
```
// crashes the server no matter what s.options.memSize
{ DelayL.ar(Impulse.ar(10), 1e6) }.play
// delay doesn't work, but server doesn't crash?
{ DelayL.ar(Impulse.ar(10), 1e7) }.play
```

capital-G (approved):
Thanks, everything looks fine to me! Maybe @JordanHendersonMusic can give it another overview?

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,326 +1,325 @@
...
-	// for each UGen in /server/plugins which calls RTAlloc
-	// test successful alloc: should output non-zero values
-	// test failed alloc: should only output zeros
+	// mem calc helpers
 
-	// FFT UGens needs separate testing (see below)
-	// Some UGens were left out (see end of file)
+	memSizeKb { ^server.options.memSize }
+	memSizeFloats { ^this.memSizeKb * (1024 / 4) }
+	memSizeSeconds { ^this.memSizeFloats / server.sampleRate }
```
mtmccrea, line 67: I wonder if these properties could be promoted to the `UnitTest` class for broad use across the test suite (and documented)?

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,326 +1,325 @@
...
-		};
+	// FFT UGens needs separate testing (see below)
+	// Some UGens were left out (see end of file)
 
-		memFill.free;
-		// now test PartConv
-		partConvTest.value;
+	test_allUGens_allocPass {
+		// allocate as little memory as possible for each UGen
+		var allocSeconds = 1 / server.sampleRate; // 1 sample
```
mtmccrea, line 120: This seems like a rather tight duration default. Are there any potential downsides, esp. wrt FFT ugens which are using some minimum number of samples for their FFT buffer?

In general, if the tests are running in parallel, having a minimum test dur wth a bit more overhead (even, say, 1 block) shouldn't be a big hit for overall test time?

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,326 +1,325 @@
...
 
-	test_allocFail_continueProcessing_otherNodes {
-		var cond = CondVar(), output = -1;
-		server.makeBundle(nil) {
-			{ DelayN.ar(Silent.ar, 100) }.play(server);
-			{ DC.ar(1) }.loadToFloatArray(1 / server.sampleRate, server, {|val|
-				output = val[0];
-				cond.signalOne();
-			});
+	assertAllocPass { |name, func, runtimeSamples|
```
mtmccrea, line 41: To be more explicit about timing, this could perhaps be called `awaitAllocPass`.

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,326 +1,325 @@
...
-		{
-			K2A.ar(Done.kr(DelayN.ar(Silent.ar, 100)))
-		}.loadToFloatArray(1 / server.sampleRate, server, {|val|
-			output = val[0];
-			cond.signalOne();
-		});
-
-		cond.waitFor(1, { output != -1 });
-		this.assertFloatEquals(output, 1.0, "a failed RTAlloc should set the Done flag")
+	assertAllocFail { |name, func, runtimeSamples|
```
mtmccrea, line 52: To be more explicit about timing, this could perhaps be called `awaitAllocFail`.

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,326 +1,325 @@
...
+		} {
+			// FFT returns a stream of -1, except when a spectral frame is ready, 
+			this.assert(out.any(_ > 0), "should return a valid bufnum (bufnum: %)".format(out));
 		};
+		// FFT should return always -1 when alloc fails
+		// - alloc half of memSize, so that LocalBuf can succeed, but FFT can fail
+		fftSize = nextPowerOfTwo(this.memSizeFloats * 0.5);
+		// - use hop = blockSize / fftSize to make FFT output frames every blockSize
+		out = this.awaitSynthOutput({ FFT(LocalBuf(fftSize), DC.ar(1), hop: blockSize / fftSize) }, blockSize);
+		out.postln;
```
mtmccrea, line 277: .postln necessary here?

mtmccrea:
I'm getting a consistent failure on my machine:
```
UNIT TESTS FOR 'TestUGen_RTAlloc' COMPLETED
There were failures:
a TestUGen_RTAlloc: test_FFT - should only return -1 on alloc fail
```

mtmccrea:
I've created a [draft of potential follow-on work](https://github.com/mtmccrea/supercollider/commit/30b914b0838eb7fa97dd3f210fac032949431e69) which parallelizes some tests to make them pass in < 1sec. It would need some more work, so I won't hold up this PR with that idea, especially if this is already an improvement in speed and stability.

Hopefully I'll get around to completing and proposing it...

testsuite/classlibrary/TestUGen_RTAlloc.sc
elgiano: I think for now it makes sense to have them just in this test, since it's especially related to filling real-time memory. But if we have other cases where it could be useful we could think of moving these methods to UnitTest or even Server or ServerOptions.

testsuite/classlibrary/TestUGen_RTAlloc.sc
elgiano: I see your point, however I would also keep the assert in there....`awaitAssertAllocPass`?

elgiano:
@mtmccrea 
I got two more comments from you on my email which I can't find here: one about consistent failure of `test_FFT - should only return -1 on alloc fail` and the other about increasing runtime to a minimum of like a blockSize.
Did you remove them or is there something strange happening with GitHub?

About parallelization: I'm not sure it's a good idea in this case, because it changes the meaning of these tests from testing each node in isolation to testing a more populated server tree all at the same time. Which sounds good to me, and it should work in practice, but I'm afraid that if we add more tests we would get into more complications. This is not meant to discard the idea, just to voice this perplexity of mine. What do you think?

Also, if it can be of any help, it looks like your parallelization mechanism replicates `Array.fork`, which we could use instead of modifying the assertAllocPass/Fail functions

```
var tests = [
    { this.assertAllocPass(...) },
    { this.assertAllocPass(...) },
    { this.assertAllocPass(...) }
];
// run in parallel
tests.fork // finishCondVar mechanism for free
// run in sequence
test.do(_.value)
```

dyfer:
> One thing I don't understand with the TCP/UDP - we are **not** sending a ton of messages during testing, and scsynth is normally capable of accepting 100s of thousands messages per second.
> 
> Though we use lots of `server.sync`, so maybe this creates some internal locking issues?

I was re-reading this thread and I remembered something: `sclang` certainly can discard UDP OSC messages when there are too many! https://github.com/supercollider/supercollider/issues/5870#issuecomment-1310965786

All I'm saying is that indeed with many messages and sync responses we might be getting unpredictable behavior as a result of UDP message queue overflow.

testsuite/classlibrary/TestUGen_RTAlloc.sc
mtmccrea: I would assume that the assertion is implicit in passing allocation test. What wasn't clear when first seeing the method was that it blocks the thread, so I thought `awaitAllocPass` would explain the function and side effect of the method.
But in any case it's just a suggestion...

testsuite/classlibrary/TestUGen_RTAlloc.sc
mtmccrea: Sure, thanks for considering it!

mtmccrea:
> Also, if it can be of any help, it looks like your parallelization mechanism replicates Array.fork, which we could use instead of modifying the assertAllocPass/Fail functions

It's a bit tricky because we need to keep track of the number of successful tests, and in my version that's what assertAllocPass is now also doing (incrementing a classvar counter). It was a rushed solution to gauge viability, there could be other approaches, and Array.fork may very well be useful.

> So, I have the parallel code ready, but probably we need more testing/thought. What do you think? Should we go for parallelization now or merge this first and parallelize later?

I think it's a promising direction, but agree it will require more testing. And the work I linked to was just one test method and should be extended to the whole class of tests if it's valid.

So because the goal of this PR is to get a stable set of RTAlloc tests, lets stick with the work you've done here and others have reviewed. Parallel tests can be built out and tested further for a future PR imo. 

Thanks for your work on this!

mtmccrea:
Re-ran a sporadic, unrelated test failure on linux, all passing again.

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,326 +1,322 @@
...
-		cond.waitFor(1, { output != -1 });
-		this.assertFloatEquals(output, 0.0, "a failed RTAlloc should clear unit outputs")
-	}
+	// Async helpers
 
-	test_allocFail_continueProcessing_sameNode {
-		var cond = CondVar(), output = -1;
-		{ DelayN.ar(Silent.ar, 100); DC.ar(1)}.loadToFloatArray(1 / server.sampleRate, server, {|val|
-			output = val[0];
+	awaitSynthOutput { |synthFn, runtimeSamples=(server.options.blockSize), timeout=1|
```
JordanHendersonMusic, line 26: Could you just remind me when this value for the runtime samples is calculated as I actually don't know!? If you change the server options at some arbitrary moment, how does this behave?

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -1,326 +1,322 @@
...
+		};
+		synthFn.loadToFloatArray(runtimeSeconds, server) { |out|
+			synthOutput = out;
 			cond.signalOne();
-		});
-
-		cond.waitFor(1, { output != -1 });
-		this.assertFloatEquals(output, 1.0, "a failed RTAlloc should not block other units in the same node")
+		};
+		cond.waitFor(timeout) { synthOutput.notNil };
```
JordanHendersonMusic, line 37: Might be worth putting an assert here (cond.waitforn returns a bool) checking we didn't timeout as not all tests check the output is nil?

JordanHendersonMusic (commented):
This looks much better thank you for your work on this! Just a question and a small comment.

testsuite/classlibrary/TestUGen_RTAlloc.sc
mtmccrea: Note that the server has to be rebooted for the server.options.blockSize to take effect.

JordanHendersonMusic: So it does `a = a ?? { server.options...}`.
Thanks!

testsuite/classlibrary/TestUGen_RTAlloc.sc
elgiano: I've added the missing timeout checks. I tried to integrate them in awaitSynthOutput, but in the end I thought it was more readable if done explicitly in every test. Otherwise tests would have to check `synthOutput.notNil` anyways, or they would run the other (non timed-out) assert.

elgiano:
While committing some missing timeout checks I have also committed the parallelized code. I admit it was by mistake, but maybe since the code is there already, shall we keep it? It would need another review at least. Or I can keep the current code organization and just run all tests in sequence instead of in parallel (it's easy to change this behavior now).
@mtmccrea what do you think?

testsuite/classlibrary/TestUGen_RTAlloc.sc
JordanHendersonMusic: Thanks!

One interesting corner case could be if the timeout expires, but before the waiting thread is resumed, the thread that sets the output completes. Your approach would not count this as an error.... but this seems inconsequential because the data would be there.

elgiano: Yes, in that case the test fails with a "test should complete" message, signaling that anyways the test took too long to complete, and that's because something wrong happened. This happens for example if some test crashes the server, causing all subsequent tests to fail... maybe we should have this covered too? For example by waiting for server to boot if it's not booted already (mostly applies only to test_allUGens test)

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -72,16 +72,24 @@ TestUGen_RTAlloc : UnitTest {
...
 			DelayN.ar(DC.ar(1), this.memSizeSeconds * 2)
 		};
-		this.assertEquals(output[0], 0.0,
-			"a failed RTAlloc should clear unit outputs")
+		if (output.isNil) {
+			this.assert(false, "test should complete")
+		} {
+			this.assertEquals(output[0], 0.0,
+				"a failed RTAlloc should clear unit outputs")
+		}
```
mtmccrea, line 80: Suggestion: an idea for this pattern of if/else that occurs multiple times—it can be made more concise throughout (untested):
```supercollider
this.assert(output.notNil, "test should complete (timeout)", onFailure: {});
this.assertEquals(output[0], 0.0, "a failed RTAlloc should clear unit outputs");
```
The docs indicate that providing a non nil value to `onFailure` will halt the test. So the method can return immediately after timing out.

mtmccrea:
>While committing some missing timeout checks I have also committed the parallelized code. I admit it was by mistake, but maybe since the code is there already, shall we keep it? It would need another review at least. Or I can keep the current code organization and just run all tests in sequence instead of in parallel (it's easy to change this behavior now).
@mtmccrea what do you think?

I think this PR can represent a dependable state of the RTAlloc tests, which can be reverted to if future additions aren't reliable. Also, the parallelization discussion seems to be ongoing and unsettled, and I think if we come up with a solid solution, that can have wider coverage in the tests. 

So my feeling is that the parallel stuff should be left out. Open to others' thought on this.

capital-G:
Maybe we should structure the discussion a bit more b/c I have the feeling there are now multiple places were we are talking and trying to coordinate the same things?
Maybe we can use #6439 as a meta issue to discuss parallelization, UDP/TCP etc.?

JordanHendersonMusic (approved):
Thanks, this has been a thorn in development!

For some reason the mac universal build isn't working — but that is unrelated.

JordanHendersonMusic:
Woops I just saw the mac tests haven't been done. Can we get the mac build working before merging this? I have no idea why it has failed.

mtmccrea:
I tried re-running it and it still fails, the logs indicate something wrong with verifying the app:
(abbreviated)
>2024-09-11T06:54:47.5975410Z -- Verifying app
2024-09-11T06:54:48.1602070Z -- executable file 1: /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/SuperCollider
2024-09-11T06:54:48.5155360Z warning: cannot resolve item '/usr/local/Cellar/glib/2.82.0/lib/libglib-2.0.0.dylib'
2024-09-11T06:54:48.5156100Z 
2024-09-11T06:54:48.5156240Z   possible problems:
2024-09-11T06:54:48.5157260Z     need more directories?
2024-09-11T06:54:48.5158010Z     need to use InstallRequiredSystemLibraries?
2024-09-11T06:54:48.5158960Z     run in install tree instead of build tree?
2024-09-11T06:54:48.5159410Z 
2024-09-11T06:54:49.0288970Z -- executable file 2: /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/sclang
2024-09-11T06:54:49.2200350Z -- 
2024-09-11T06:54:49.2203820Z warning: cannot resolve item '/usr/local/Cellar/glib/2.82.0/lib/libglib-2.0.0.dylib'
2024-09-11T06:54:49.2204680Z 
2024-09-11T06:54:49.8265010Z -- executable file 3: /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/Resources/supernova
2024-09-11T06:54:49.9927300Z -- executable file 4: /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/Resources/scsynth
2024-09-11T06:54:50.1225400Z -- executable file 5: /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/Frameworks/QtWebEngineCore.framework/Versions/5/Helpers/QtWebEngineProcess.app/Contents/MacOS/QtWebEngineProcess
2024-09-11T06:54:50.2560840Z -- 
2024-09-11T06:54:50.2563540Z warning: cannot resolve item '/usr/local/Cellar/glib/2.82.0/lib/libglib-2.0.0.dylib'
2024-09-11T06:54:50.2564330Z 
2024-09-11T06:54:50.5734970Z CMake Error at /usr/local/Cellar/cmake/3.30.3/share/cmake/Modules/BundleUtilities.cmake:1127 (message):
2024-09-11T06:54:50.5736070Z   error: verify_app failed
2024-09-11T06:54:50.5736860Z Call Stack (most recent call first):
2024-09-11T06:54:50.5737570Z   editors/sc-ide/cmake_install.cmake:110 (verify_app)
2024-09-11T06:54:50.5738230Z   editors/cmake_install.cmake:44 (include)
2024-09-11T06:54:50.5738790Z   cmake_install.cmake:63 (include)
2024-09-11T06:54:50.5739240Z 
2024-09-11T06:54:50.6102670Z ** BUILD FAILED **
2024-09-11T06:54:50.6103240Z 
2024-09-11T06:54:50.6105580Z The following build commands failed:
2024-09-11T06:54:50.6108710Z 	PhaseScriptExecution CMake\ PostBuild\ Rules /Users/runner/work/supercollider/supercollider/build/build/SuperCollider.build/Release/install.build/Script-93426DA6A4AC2C2AC9CE883C.sh (in target 'install' from project 'SuperCollider')

capital-G:
this is most likely because of https://github.com/supercollider/supercollider/pull/6458 - what do you think @dyfer ?

But just to be sure, please rebase against dev - could be an outdated cache.

dyfer:
> this is most likely because of #6458 - what do you think @dyfer ?

I wouldn't say it's _because of #6458_...

TL;DR these failures should be temporary; they should go away after some update happens in the homebrew ecosystem (which might take hours or days, I'm not sure). **It should be enough to re-run the job after some time**. 

The occasional failure is an issue that stems from us using a hacky workaround to create universal homebrew libraries (I know it's hacky - I made it). I _think_ it happens when there's a discrepancy between the version of the packages that are installed and the ones that are downloaded from homebrew repository, however IIRC I was doing everything I could to make sure that the downloaded package is the same version that would be installed by homebrew. Unfortunately, I wasn't able to pinpoint the issue. Also, I _think_ the caching itself should not be an issue, the script checks for the SHA of the package that's in cache and the one that would be downloaded. However, this whole process is delicate so I might be missing something.

> But just to be sure, please rebase against dev - could be an outdated cache.

Rebasing doesn't get rid of cache, at least not automatically. Cache (as in cache in GHA, which we use for homebrew downloads, ccache, vcpkg builds etc) is stored per branch, but also the ones from the default branch are available to the other branches, IIUC. The cache is not used if the [cache key](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/build_macos.yml#L112) changes, which might happen when you rebase, in case `develop` has updates to cache keys.

capital-G:
Thanks for your reply! Good that you are around to maintain that *strange* shell script :D I thought about switching to MacPorts which supports native universal builds, but that would be for another issue/discussion/PR with pros/cons.

> Rebasing doesn't get rid of cache, at least not automatically. Cache (as in cache in GHA, which we use for homebrew downloads, ccache, vcpkg builds etc) is stored per branch, plus the cache from the default branch is available to the other branches, IIRC. The cache is not used if the [cache key](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/build_macos.yml#L112) changes, which might happen when you rebase, in case `develop` has updates to cache keys.

I had https://github.com/supercollider/supercollider/pull/6465 in mind which recently modified those keys.

dyfer:
> Thanks for your reply! Good that you are around to maintain that _strange_ shell script :D I thought about switching to MacPorts which supports native universal builds, but that would be for another issue/discussion/PR with pros/cons.

No problem :)
<details>
The script is there because at the time there were no ARM-based GH runners and thus no way to get Qt and other libraries in either ARM or ARM/X86_64 "universal" flavor.  Currently we could theoretically switch to the macOS 14 runner to build Arm-only build. I personally like one universal build, but it did get big. It's even bigger when built with Qt6, so we can certainly consider separate ones (with all its logistical consequences). Blah blah blah - more for another issue.
</details>

> 

> I had #6465 in mind which recently modified those keys.

Ah, I see. The ccache key is unlikely to be an issue here, IMO. I was thinking that you were referring to homebrew cache, which did not change.

> 
> Edit: @dyfer can you maybe write some of the information concerning #6458 into the PR or the issue? Hopefully it will not be needed, but in case it pops up again it would be great to have some lead on what to expect etc.

Most of my verify_app knowledge stems from the thread with [this comment](https://github.com/supercollider/supercollider/issues/5294#issuecomment-791244428). What would be a good way to make sure people can find this? It shows up when you look for `verify_app`, but it's burried in many less informative issues.

> What is your advice for this PR - wait a few days and re-run the CI? Because I'd personally would only like to merge this when I see a valid CI run.

I'm checking that issue right now, will update once the builds finish...

dyfer:
> > Thanks for your reply! Good that you are around to maintain that _strange_ shell script :D I thought about switching to MacPorts which supports native universal builds, but that would be for another issue/discussion/PR with pros/cons.
> 
> No problem :)
> 
> > 
> 
> > I had #6465 in mind which recently modified those keys.
> 
> Ah, I see. The ccache key is unlikely to be an issue here, IMO. I was thinking that you were referring to homebrew cache, which did not change.
> 
> > Edit: @dyfer can you maybe write some of the information concerning #6458 into the PR or the issue? Hopefully it will not be needed, but in case it pops up again it would be great to have some lead on what to expect etc.
> 
> Most of my verify_app knowledge stems from the thread with [this comment](https://github.com/supercollider/supercollider/issues/5294#issuecomment-791988719). What would be a good way to make sure people can find this? It shows up when you look for `verify_app`, but it's burried in many less informative issues.

EDIT: I could probably describe the specifics where that relates to the possible issues with universal binaries made with the script.

> 
> > What is your advice for this PR - wait a few days and re-run the CI? Because I'd personally would only like to merge this when I see a valid CI run.
> 
> I'm checking that issue right now, will update once the builds finish...

dyfer:
> What is your advice for this PR - wait a few days and re-run the CI?

I manually deleted the brew_universal cache (in the Actions webpage interface) and am re-running the job. I think the discrepancy between the brew_universal cache and homebrew were the source of the issue. We'll see...

dyfer:
I think something is temporarily wrong with either gha macos images, or with homebrew. I tried to f[orce updating homebrew on my branch and that failed](https://github.com/dyfer/supercollider/actions/runs/10824265712/job/30031225520) in itself. I think we need to wait for the fix.

Otherwise, if the tests pass on Linux, I think merging this would be fine as is.

dyfer:
The builds are now fixed on the develop branch, rebasing this PR should fix the build issue.
I also wrote more some [information about verify_app here](https://github.com/supercollider/supercollider/issues/6480).

dyfer:
I assume this is good to be merged now?

elgiano:
Yes I think so to!

Sent from Proton Mail Android

telephon:
I get the a few errors:

```
<path>external_libraries/boost/boost/move/detail/type_traits.hpp:134:46: note: expanded from macro 'BOOST_MOVE_HAS_NOTHROW_COPY'
#     define BOOST_MOVE_HAS_NOTHROW_COPY(T) (__has_nothrow_copy(T))
                                             ^
In file included from <path>external_libraries/boost/libs/thread/src/pthread/thread.cpp:11:
In file included from <path>external_libraries/boost/boost/thread/thread_only.hpp:22:
In file included from <path>external_libraries/boost/boost/thread/detail/thread.hpp:41:
In file included from <path>external_libraries/boost/boost/functional/hash.hpp:6:
<path>external_libraries/boost/boost/container_hash/hash.hpp:131:33: error: no template named 'unary_function' in namespace 'std'; did you mean '__unary_function'?
        struct hash_base : std::unary_function<T, std::size_t> {};
                           ~~~~~^~~~~~~~~~~~~~
                                __unary_function




<path>external_libraries/boost/boost/mpl/aux_/integral_wrapper.hpp:73:31: error: integer value -1 is outside the valid range of values [0, 3] for this enumeration type [-Wenum-constexpr-conversion]
    typedef AUX_WRAPPER_INST( BOOST_MPL_AUX_STATIC_CAST(AUX_WRAPPER_VALUE_TYPE, (value - 1)) ) prior;
                              ^
<path>external_libraries/boost/boost/mpl/aux_/static_cast.hpp:24:47: note: expanded from macro 'BOOST_MPL_AUX_STATIC_CAST'
#   define BOOST_MPL_AUX_STATIC_CAST(T, expr) static_cast<T>(expr)
                                              ^
			
<path>external_libraries/boost/boost/mpl/aux_/integral_wrapper.hpp:73:31: error: integer value -1 is outside the valid range of values [0, 3] for this enumeration type [-Wenum-constexpr-conversion]
<path>external_libraries/boost/boost/mpl/aux_/static_cast.hpp:24:47: note: expanded from macro 'BOOST_MPL_AUX_STATIC_CAST'
#   define BOOST_MPL_AUX_STATIC_CAST(T, expr) static_cast<T>(expr)
                                              ^
13 warnings and 3 errors generated.
make[2]: *** [external_libraries/CMakeFiles/boost_thread_lib.dir/boost/libs/thread/src/pthread/thread.cpp.o] Error 1
make[1]: *** [external_libraries/CMakeFiles/boost_thread_lib.dir/all] Error 2
make: *** [all] Error 2
```

capital-G:
> I get the a few errors:

Should be unrelated. See #6477 which should fix those issues.

mtmccrea:
Is it preferable to squash merge PRs like this?

elgiano:
@mtmccrea I agree, I would also have expected a squash merge with so many (even contradictory) commits. For the future: @dyfer was it just an oversight or was it on purpose, or did you expect me to squash it first?

mtmccrea:
I think it's ultimately up to the discretion of the merging reviewer unless there's been some explicit discussion about it. It's easy to overlook though, so when I prepare PRs I try to have the commit history cleaned up in a way that if it's merged without squashing it's still concise and divided into separate concerns/components.

dyfer:
> @dyfer was it just an oversight or was it on purpose, or did you expect me to squash it first?

An oversight on my part, sorry! I probably would've squashed it if I looked at the commit list, and I didn't.

Sometimes PR authors would indicate in a comment they'd want a squash on merge, which is a helpful hint for the person merging.

PR #6454, opened 2024-09-07
plugins: SpecPcile output zeros on RTAlloc fail

elgiano:
## Purpose and Motivation

SpecPcile calls RTAlloc in its _next function (thus after _Ctor). If RTAlloc fails, the UGen still outputs some non-zero values for one block. This is not the expected behavior, especially by RTAlloc UnitTests, which were failing.

This PR ensures clear output when RTAlloc fails.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/server/plugins/ML_SpecStats.cpp
+++ b/server/plugins/ML_SpecStats.cpp
@@ -209,7 +209,10 @@ void SpecPcile_next(SpecPcile* unit, int inNumSamples) {
         // Used to be MAKE_TEMP_BUF but we can handle it more cleanly in this specific case:
         if (!unit->m_tempbuf) {
         unit->m_tempbuf = (float*)RTAlloc(unit->mWorld, numbins * sizeof(float));
-        ClearUnitIfMemFailed(unit->m_tempbuf);
+        if (!unit->m_tempbuf) {
+            ClearUnitOutputs(unit, inNumSamples);
+            ClearUnitOnMemFailed;
+        }
         unit->m_numbins = numbins;
         unit->m_halfnyq_over_numbinsp2 = ((float)unit->mWorld->mSampleRate) * 0.5f / (float)(numbins + 2);
     }


server/plugins/ML_SpecStats.cpp
```diff
@@ -209,7 +209,10 @@ void SpecPcile_next(SpecPcile* unit, int inNumSamples) {
         // Used to be MAKE_TEMP_BUF but we can handle it more cleanly in this specific case:
         if (!unit->m_tempbuf) {
         unit->m_tempbuf = (float*)RTAlloc(unit->mWorld, numbins * sizeof(float));
-        ClearUnitIfMemFailed(unit->m_tempbuf);
+        if (!unit->m_tempbuf) {
+            ClearUnitOutputs(unit, inNumSamples);
+            ClearUnitOnMemFailed;
```
capital-G, line 214: The example plugins also show that they "map" the next_calc function to `ft->fClearUnitOutputs`, see https://github.com/supercollider/example-plugins/blob/d0622ac6a4919bc7db134561d292d71337a9823d/03-AnalogEcho/AnalogEcho.cpp#L47-L53

Is this relevant here? Don't have too much experience with rt malloc (yet).

capital-G (commented):
The formatting looks funky to me (due to `FFTAnalyser_GET_BUF` macro(?)), but if clang-format approves its fine.

I can't really evaluate if the changes are properly done, but if it makes the CI pass it sounds reasonable?

server/plugins/ML_SpecStats.cpp
elgiano: Yes, it's done by `ClearUnitOnMemFailed`:
```
#define ClearUnitOnMemFailed                                                                                           \
    Print("%s: alloc failed, increase server's RT memory (e.g. via ServerOptions)\n", __func__);                       \
    SETCALC(*ClearUnitOutputs);                                                                                        \
    unit->mDone = true;                                                                                                \
    return;
```

capital-G (approved):

Issue #6456, opened 2024-09-08
SerialPort tests break sclang on Ubuntu 24.04


dyfer:
## Environment

* SuperCollider version: develop
* Operating system: Ubuntu 24.04 - GHA runner
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider

// I think that would be:
TestSerialPort.run;

```

## Expected vs. actual behavior

Actual behavior: sclang _freezes_ and tests never finish 
Expected: all tests pass :)

As part of the Github Actions maintenance I wanted to update the OS version of our tests runner. It turned out that the SerialPort tests make sclang hang when run in Ubuntu 24.04. For reference, currently passing tests run fine in Ubuntu 22.04.

I have made a little test matrix to check whether this might be an OS/library or a compiler issue: https://github.com/dyfer/supercollider/actions/runs/10755770617/job/29827953623.  It looks like it's caused by the OS/library version, since the build made with gcc12 on Ubuntu 22.04 passes, but the one made with gcc12 on Ubuntu 24.04 fails.

I haven't investigated further, as I don't have a 24.04 system available currently. I'm filing this in so that we are aware this will likely come up when we update build/test matrix.

<details>

```
Test: test_rxErrors_bufferOverflow
/usr/bin/socat
FAIL: a TestSerialPort: test_rxErrors_bufferOverflow
UNIT TESTS FOR 'TestSerialPort' COMPLETED
There were failures:
a TestSerialPort: test_rxErrors_bufferOverflow
ERROR: Message 'split' not understood.
RECEIVER:
   nil
ARGS:
   Character 10 '
'
KEYWORD ARGUMENTS:
PATH: /home/runner/work/supercollider/supercollider/testsuite/scripts/sclang_test_runner.scd
PROTECTED CALL STACK:
	Meta_MethodError:new	0x55ee11c0ce00
		arg this = DoesNotUnderstandError
		arg what = nil
		arg receiver = nil
	Meta_DoesNotUnderstandError:new	0x55ee11c0f540
		arg this = DoesNotUnderstandError
		arg receiver = nil
		arg selector = split
		arg args = [
]
		arg keywordArgumentPairs = []
	Object:doesNotUnderstand	0x55ee1132be00
		arg this = nil
		arg selector = split
		arg args = nil
		arg kwargs = nil
	a FunctionDef	0x55ee133ca6e8
		sourceCode = "<an open Function>"
		arg fail = an UnitTestResult
[...]

Issue #6455, opened 2024-09-07
GHA: using version numbers for Linux OS identifiers?


dyfer:
## Motivation

We currently use the Ubuntu distribution names to indicate runner image OS version for our Linux builds:
```
...
- Linux focal gcc9
- Linux jammy gcc10
...
```
As I was updating the build matrix, I found myself looking up (again) the names of the distributions, corresponding to the version numbers (22.04, 24.04). The version numbers refer to the release year and month, which IMO makes it easier to identify the OS, as opposed to the name, which is more obscure for those not using Ubuntu on regular basis.

## Description of Proposed Feature

This is cosmetic, but I propose that we use **version numbers** to describe Linux builds. These numbers refer explicitly to _Ubuntu_ versions, so I wonder whether the jobs should be called `Linux` or `Ubuntu`. In either case, I'd find them easier to understand.

I suggest to go with either
```
- Linux 20.04 gcc9
- Linux 22.04 gcc10
```
or
```
- Ubuntu 20.04 gcc9
- Ubuntu 22.04 gcc10
```

What do you think?

## Plan for Implementation

I'd implement whatever we agree on.

capital-G:
I'd vote for `Ubuntu 22.04 gcc10`, seems to be the most explicit and matching to me.

PR #6457, opened 2024-09-08
GHA: update build matrix

dyfer:
## Purpose and Motivation

We need to update the build matrix for the upcoming Qt updates that I'm planning to submit.

I loosely followed our [toolchain support documentation](https://github.com/supercollider/supercollider/wiki/Platform-Support#toolchains) and included the last 4 versions of GCC and Clang, plus additionally the oldest versions that still work without issues.

I wanted to submit GHA changes separately, also including GHA commits from #6331, so that Qt updates don't need to wait for the other PR to be merged.

Note: this update includes entries for clang 16, 17, 18, but these builds currently fail due to outdated boost. I left them in, but they are commented out and should be enabled once we update boost (I believe that's the same issue as #6224 on macOS).

Also note: I wanted to update the test runner image to 24.04, but encountered issues - see https://github.com/supercollider/supercollider/issues/6456 - so the tests are kept using the same configuration as before (Ubuntu 22.04, gcc 12).

## Types of changes

- Maintenance

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation - n/a: this is a transitory state - let's discuss documentation in the upcoming Qt update PRs

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -13,25 +13,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - job-name: "focal gcc7"
-            os-version: "20.04"
-            c-compiler: "gcc-7"
-            cxx-compiler: "g++-7"
-            use-syslibs: false
-            shared-libscsynth: false
 
-          - job-name: "focal gcc9 shared libscsynth"
-            os-version: "20.04"
+          - job-name: "jammy gcc9"
+            os-version: "22.04"
             c-compiler: "gcc-9"
             cxx-compiler: "g++-9"
             use-syslibs: false
-            shared-libscsynth: true
-
-          - job-name: "jammy gcc10 use system libraries"
-            os-version: "22.04"
-            c-compiler: "gcc-10"
-            cxx-compiler: "g++-10"
-            use-syslibs: true
             shared-libscsynth: false
 
           - job-name: "jammy gcc11"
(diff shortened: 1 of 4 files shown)


capital-G (commented):
Can we incorporate https://github.com/supercollider/supercollider/issues/6455 ? I am already mixing up the Debian names often - I think for Ubuntu it is valid to stick to their date name?

Otherwise looks fine to me - is this still a draft/WIP though?

dyfer:
> Otherwise looks fine to me - is this still a draft/WIP though?

I kept it as draft only because I wanted to see if you're OK merging this first, then rebasing #6331 

> Can we incorporate https://github.com/supercollider/supercollider/issues/6455 ? 

Sure, I'll add these changes.

.github/workflows/build_linux.yml
```diff
@@ -13,78 +13,80 @@ jobs:
...
-          - job-name: "focal gcc7"
-            os-version: "20.04"
-            c-compiler: "gcc-7"
-            cxx-compiler: "g++-7"
-            use-syslibs: false
-            shared-libscsynth: false
 
-          - job-name: "focal gcc9 shared libscsynth"
-            os-version: "20.04"
+          - job-name: ""
```
capital-G, line 17: why is the jobname now empty? I thought it would be `22.04 gcc9 shared libscsynth` in this case.

dyfer: I made a change so that `job-name` really becomes "job name suffix": the name is now constructed as [`Ubuntu ${{ matrix.os-version }} ${{ matrix.c-compiler }} ${{ matrix.job-name }}`](https://github.com/supercollider/supercollider/pull/6457/files#diff-857070821eb9baccd26d4c48bc577d65473e8e4f0dfe3a69473fe0f26bcfcbc2R89) - less duplicated typing.

It's hard to see in the diff, but the `use system libraries` and `share libscsynth` were [moved to the `gcc-13` and `gcc-14` builds](https://github.com/supercollider/supercollider/blob/cb15f22eb62dbd137441ce7e7974ad89714c06f0/.github/workflows/build_linux.yml#L39-L51), respectively.

capital-G: Oh right, should have looked closer, sorry!

capital-G:
I think this is ready to be merged once the pipeline succeeds - should this be squashed?

dyfer:
> should this be squashed?

Either way I think - IMO having the name change as a separate commit is slightly clearer, but I really could go either way, it's a small PR.

dyfer:
@capital-G do you think we can merge this?

capital-G (approved):

dyfer:
I discovered that this update broke the Windows MinGW build. 
While it did build in the CI, when trying to run scide/sclang/scsynth, I get the following message:
```
The procedure entry point 
_ZNKSt25_codecut_utf&_utf16_baselwE10d_unshiftERg_MbstatetPoss_R33- 
could not be located in the dynamic link library C:\Users\<user>\Desktop\SuperCollider\scide.exe.
```
This happens in the build created with the toolchain provided by github actions runner image, `gcc 12.2.0`.

We don't officially support MinGW, but we were checking whether it still builds. It seems that checking only whether it builds didn't catch  this issue.
@sonoro1234 do you have any insights as to what might be the reason for this?

sonoro1234:
I dont usually build scide but only scserver and supernova.
The issue seems to be that `_ZNKSt25_codecut_utf&_utf16_baselwE10d_unshiftERg_MbstatetPoss_R33` which is the decorated name of a C++ function cant be found in some dll library required by scide

dyfer:
> I dont usually build scide but only scserver and supernova. 

I got these error when trying to run all of the executables, including `scsynth` and `supernova`. Could you please try building from the latest develop to see if you can reproduce this problem?

> The issue seems to be that `_ZNKSt25_codecut_utf&_utf16_baselwE10d_unshiftERg_MbstatetPoss_R33` which is the decorated name of a C++ function cant be found in some dll library required by scide

Thanks. I have a feeling this is related to #6124, that PR adds dependence on codecvt in all the executables to support Unicode in system calls on Windows.

sonoro1234:
I could build and boot without problems.
It seem a problem related to a missing dll in your built directory, althought I have no new dll in my built directory.

I use mingw-w64 but not using MSYS at all.

dyfer:
@sonoro1234 sorry to bother you, just do double check: when you said your build worked, did you try running the build from outside of the build directory?

sonoro1234:
No

dyfer:
Thanks!

Issue #6449, opened 2024-09-04
Document that you must never wait on a CondVar without a predicate


Spacechild1:
`CondVar` is a classic condition variable and therefore *always* requires a predicate.

The documention does mention two ways of using `CondVar.wait`:

1. Manually checking the predicate in a loop
2. Passing a predicate function

However, the documentation should also make it clear that you absolutely cannot omit the predicate! I think this should be in a prominent `note::` or even `warning::` block.

Ideally, we should also explain *why* you must not omit the predicate (= lost wake-ups). This is not obvious to people who are not familiar classic condition variables (such as `pthread_cond_t` or `std::condition_variable`). Also, the similarly named `Condition` class does not have this issue*), so we could maybe also point out the differences.

*)  `Condition` has its own set of issues, which is why `CondVar` has been introduced in the first place.

Issue #6448, opened 2024-09-04
Document that certain asynchronous methods block inside Routines


Spacechild1:
Asynchronous methods like `Buffer.loadToFloatArray` use `forkIfNeeded` behind the scene, which means that they actually block when executed inside a Routine! Unfortunately, this behavior is currently not documented and many users are not aware of it, leading to very subtle issues such as https://github.com/supercollider/supercollider/pull/6441#issuecomment-2329977621

(I find it a bit strange that an asynchronous method would block depending on its execution context, but we cannot really change it.)

Anyway, this behavior is currently not documented at all!

This applies to several `Buffer` buffer methods and maybe also some other methods. A grep for `forkIfNeeded` may help find all affected methods.

muellmusik:
Can we not fix it? I don’t remember what the intention was but I agree it seems odd.

Spacechild1:
I think we cannot change it without breaking backwards compatibility...

telephon:
Seemed to be a good idea at the time, 15 years ago. The idea was that you could call several of them in order and they would embed in the outer routine.

```supercollider
var buffers;
fork {
   var result;
    buffers.do { |b| b.loadToFloatArray(action: { |x| result = result.add(x) }) }
}
```

this would not be possible as simply with a fork.

Now the problem appears when we write

```supercollider
var buffers;
fork {
   var result, cond;
   cond = CondVar.new;
    buffers.do { |b| 
         var a;
         b.loadToFloatArray(action: { |x|  a = x; cond.signalOne; });
         cond.wait { a.notNil };
         result = result.add(x) 
    }
}
```
It doesn't reach `cond.wait` until it is ready anyway and has called `signalOne`.
I am not sure if there is a general way to resolve this, maybe you have an idea?

Spacechild1:
> The idea was that you could call several of them in order and they would embed in the outer routine.

Yes, that's what I figured. It makes sense given that SC didn't really have a modern async programming model.

(
If the SC Class Library had been written 15 years later, this would probably look as follows:
```
var buffers;
fork {
    // loadToFloatArray would returns a Future
    var result = (buffers.collect { |b| b.loadToFloatArray }).awaitAll;
}
```
)

> I am not sure if there is a general way to resolve this, maybe you have an idea?

As I've written in the other thread, you can just wrap the call to the async function in a Routine:

```
fork {
    var a;
    fork {         
        b.loadToFloatArray(action: { |x|  a = x; cond.signalOne; });
[...]

PR #6460, opened 2024-09-08
UnitTest, TestPV_Chain: bug fix, CondVar should have a predicate.

JordanHendersonMusic:
## Purpose and Motivation

This test was waiting until the timeout occurred due to the confusion around #6449  and #6448. 

Should now take circa 5 seconds to run, rather than 100 ish.

There are a large number of line changes here because the code was indented wrong.
See method `pv_equivalencetests_common` for actual changes.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestPV_ChainUGen.sc
+++ b/testsuite/classlibrary/TestPV_ChainUGen.sc
@@ -4,95 +4,96 @@ UnitTest.gui
 */
 TestPV_ChainUGen : UnitTest {
 
-// This test manually creates an FFT frame equivalent to pure white noise, 
-// on which certain ugens' action should be a no-op
-test_whiteframe_noop {
-	var fakeframe;
-	
-	// Every real is 1.0, every imag is 0.0. same as every mag is 1.0, every phase is 0.0.
-	fakeframe = ([1,1] ++ {[1,0]}.dup(511)).flat;
-	
-	// All of these should be no-ops when applied to a faked frame with reals all 1 and imags all 0:
-	this.pv_equivalencetests_common(fakeframe,
-		Dictionary[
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
A part of this effort #6460

capital-G (approved):
I think this can be speed up even more when both tests share the same server instead of doing a cold server start for each test. This could intorduce some "test-leakage"  between both tests if the server state is poisoned from the other test, but most likely the test results would show this as well, but could save another 1-2 secs.
And thanks for the much improved formatting!

Maybe we should introduce a `UGenUnitTest` class at some point which takes care of providing a running server environment?

Everything looks fine to me. Should we merge this and close https://github.com/supercollider/supercollider/pull/6440?

JordanHendersonMusic:
>  but could save another 1-2 secs.

Lets see what we can do without sharing state first, as that might introduce other bugs or subtly change some of the tests?
We could actually allocate all the buffers first, then run all the synths, then do all the tests. This could be even faster as you don't have to wait for each synth, but all at once. I'm current rewriting TestCoreUGens to do this better.

>Should we merge this and close https://github.com/supercollider/supercollider/pull/6440?

Ah sorry, didn't see that you had already tackled this! I don't mind which one, but this also checks the timeout doesn't fail.

capital-G:
> We could actually allocate all the buffers first, then run all the synths, then do all the tests. This could be even faster as you don't have to wait for each synth, but all at once.

I think most time is spend on booting, or waiting for a CondVar, which we can tackle easily. But the tests should be dumb and not go too much into abstraction which needs to be tested again as well.

Should we wait for another review or can this be merged?

JordanHendersonMusic:
I think its good to go, everything is green.

PR #6462, opened 2024-09-08
UnitTest, TestCoreUGens: better parallelization.

JordanHendersonMusic:
## Purpose and Motivation

Shaves about 20% of the duration from `TestCoreUGens` by writing better parallelisation in supercollider and using TCP, which means you don't have to `server.sync`.

A part of the continued effort to speed up unit test #6439
## Types of changes

- Unit test performance

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -18,21 +18,49 @@ TestCoreUGens : UnitTest {
 
 	setUp {
 		server = Server(this.class.name);
+		server.options.protocol = \tcp
+
 	}
 
 	tearDown {
-		server.quit;
+		var cond = CondVar();
+		var done = false;
+		server.quit(onComplete: {
+			done = true;
+			cond.signalAll;
+		});
+		cond.wait { done };
 		server.remove;
 	}
 
-	test_ugen_generator_equivalences {
+	run_buffer_tests_in_parallel { |name, argNames, args, funcProducer, testDur, actionWithArray|
 		var cond = CondVar();
 		var completed = 0;
-		var n, v, tests, testDur, testsFinished;
+		args.do { |currentArgs|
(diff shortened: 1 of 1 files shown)


capital-G:
Are we sure we want to run tests on the server in parallel? IMO we should avoid this as it can lead to unknown and not-reproducible states and contamination, and adds complexity (especially since async/thread spawning and parallelism is a bit *wonky* in sclang).

I am fine with removing the booting/shutdown sequences of the server because these take multiple seconds each time and I trust the CmdPeriod command enough to provide a clean state of the server.

Also using TCP feels like a hack to me as TCP is not really the common way of communicating with scsynth. IMO tests should run on a provided server and not spawn their own servers.

Let's see what other people think about this.

JordanHendersonMusic:
> it can lead to unknown and not-reproducible states and contamination

Since this is only using `loadToFloatArray` it should be fine as nothing should be reused. Though in general, I agree with this.

> Also using TCP feels like a hack to me as TCP is not really the common way of communicating with scsynth.
> especially since async/thread spawning and parallelism is a bit wonky in sclang.

I kinda disagree with this, the sclang thread stuff is annoying because it lacks promises, but once you accept callback  hell, it seems to work well, particularly when combined with `CondVar` and a cout. The only wonky part I've ever found is when you try to send multiple messages to the server quickly, hence TCP.
Technically I could disable TCP for some of these tests as they don't , create that many messages, but UDP will fail as you add more.

I started uses TCP when using Flucoma as modifying lots of small buffers is a common pattern and doing them sequentially can be significantly slower. It took forever to find out the UDP was the problem and I was dealing with run times circa 30 mins, when it could have been more like 3. TCP should be more common.

JordanHendersonMusic:
I'm going to close this, not because I don't think its a good idea, but TCP seems to be broken on the CI and I don't have time/known enough about CI to fix it.

Issue #5337, opened 2021-02-06
Scsynth doesn't clean up TCP connections


dyfer:
## Environment

* SuperCollider version: 3.11.2
* Operating system: macOS 10.14
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
// tcp (dis)connection test - can performed locally
// start the server
~pid = (Server.program + s.options.protocol_(\tcp).maxLogins_(10).asOptionsString).unixCmd;

// then connect as if it was a remote server
(
~server = Server.remote(\localRemote, s.addr, ServerOptions().protocol_(\tcp));
~server.addr.tryConnectTCP; 
)

~server.makeGui; // check if it runs

(
// disconnect
~server.stopAliveThread;
~server.addr.tryDisconnectTCP;
~server.remove;
)

// so far, so good

// try connecting again
(
~server = Server.remote(\localRemote, s.addr, ServerOptions().protocol_(\tcp));
~server.addr.tryConnectTCP; // needed before 3.12
)

// result (posted by the server)
/*
Bad file descriptor
Bad file descriptor
Bad file descriptor
Bad file descriptor
Bad file descriptor
Bad file descriptor
...
*/

// cleanup - stop the server
thisProcess.platform.killProcessByID(~pid);
// and disconnect
(
~server.stopAliveThread;
~server.addr.tryDisconnectTCP;
~server.remove;
)

```

## Expected vs. actual behavior

Scsynth doesn't seem to properly close/clean up after TCP connections. After disconnecting from the server using TCP connection, subsequent attempts to connect fail and the server prints `Bad file descriptor` multiple times. I looked through the existing issues and it seems like #932 might be related.
For local servers this is not an issue, since we typically quit the server when we're done with it, but for remote servers it's a perfectly valid scenario - wanting to disconnect from it, but not actually quit it.
Supernova does not have this issue.

PR #6463, opened 2024-09-08
UnitTest, TestFilterUgens: remove s.sync, and udp -> tcp

JordanHendersonMusic:
## Purpose and Motivation

Reduces execution time by circa half.

A part of the continued effort to speed up unit test https://github.com/supercollider/supercollider/issues/6439

## Types of changes

- Unit test performance

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestFilterUGens.sc
+++ b/testsuite/classlibrary/TestFilterUGens.sc
@@ -3,6 +3,7 @@ TestFilterUGens : UnitTest {
 
 	setUp {
 		server = Server(this.class.name);
+		server.options.protocol = \tcp;
 		server.bootSync;
 	}
 
@@ -114,7 +115,6 @@ TestFilterUGens : UnitTest {
 					condvar.signalOne;
 				});
 
-				rrand(0.012,0.035).wait;
 			}
 		};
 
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
Neither this nor #6462 seems to work on the CI. It doesn't seem able to make the TCP connection, does any one have any idea why?

dyfer:
Ah, this might be the reason: #5337 
Which would mean that we probably can't use TCP in the test suite unless we fix that.

JordanHendersonMusic:
It's just odd it seems to work locally, although sometimes it prints non-fatal errors: `bad pipe`, and failing to connect (but this retries and succeeds).

dyfer:
@JordanHendersonMusic I think switching to TCP is not viable currently and we can consider this only after addressing existing issues. I wrote more about an attempt with TCP (inspired by this PR) in https://github.com/supercollider/supercollider/issues/6439#issuecomment-2819083252

For this PR, could you replace the removed wait times with `server.sync` and remove the `\tcp` switch? I think this would still be an improvement.

JordanHendersonMusic:
I'm closing this because I don't have time to work on it and it seems tcp isn't working in the CI.

PR #6465, opened 2024-09-08
GHA: macOS: fix caching when switching deployment targets

dyfer:
## Purpose and Motivation

I observed that despite the presence of cache, [macOS builds take a long time after changing the deployment target](https://github.com/supercollider/supercollider/actions/runs/10761956706/job/29841728669). I think that the old cache is not valid when changing the deployment target, so the cache key should also include deployment target.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested - seems to work: [the build on the second run in this branch was fast](https://github.com/supercollider/supercollider/actions/runs/10762655520/job/29843269817#step:18:1)
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -90,8 +90,8 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/Library/Caches/ccache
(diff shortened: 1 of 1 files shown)


capital-G:
sounds reasonable, lets merge :)

capital-G (approved):

capital-G:
BTW - why don't we store the build directory in the cache? Wouldn't this allow incremental builds? It could introduce other problems though...

dyfer:
AFAIU a non-clean build directory is a recipe for troubles indeed. `ccache` generally is a robust solution.

PR #6467, opened 2024-09-10
server: add /rtMemoryStatus cmd for debugging

elgiano:
## Purpose and Motivation

scsynth provides a call for querying the amount of currently free real-time memory, but this call is never used and inaccessible. This PR adds a server /cmd (and related sclang interface) to get this information:
```
s.boot;
s.rtMemoryStatus
// Used RT memory: 16.5625 kb
// Free RT memory: 8175.4375 kb
// Largest free chunk: 8175.4375 kb

x = {DelayN.ar(DC.ar(0.0), maxdelaytime: 4.0)}.play;
s.rtMemoryStatus
// Used RT memory: 1043.0625 kb
// Free RT memory: 7148.9375 kb
// Largest free chunk: 7148.0625 kb

x.free;
s.rtMemoryStatus
// Used RT memory: 16.4375 kb
// Free RT memory: 8175.5625 kb
// Largest free chunk: 8175.5625 kb
```

I have no idea how to implement this for supernova.
Perhaps another option would be to include mem info in /status? Sounds nicer to me, but perhaps also too much of a breaking change?

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/Classes/Server.schelp
+++ b/HelpSource/Classes/Server.schelp
@@ -394,6 +394,13 @@ s.queryAllNodes; // note the root node (ID 0) and the default group (ID 1)
 s.quit;
 ::
 
+method:: rtMemoryStatus
+Queries the server for the amount of currently free real-time memory
+argument:: action
+a Function that will be evaluated when the server responds, with the current
+amount of free RT memory and the size of largest continuous block (both in bytes) passed as
+arguments. The default action prints a message.
+
 method:: peakCPU, avgCPU
 Get peak and average CPU usage.
 
--- a/HelpSource/Reference/Server-Command-Reference.schelp
+++ b/HelpSource/Reference/Server-Command-Reference.schelp
@@ -134,6 +134,18 @@ The standard human-readable version string can be constructed by concatenating c
 
 The git branch/tag name and commit hash could be anything if the user has forked SC, so they should only be used for display and user interface purposes.
 
+
+subsection:: /rtMemoryStatus
+
+Queries the amount of currently free real-time memory (in bytes).
+
+definitionlist::
+## /rtMemoryStatus.reply || table::
+## int || Amount of free real-time memory (in bytes).
+## int || Size of largest continuous block (in bytes).
+::
+::
+
 section:: Synth Definition Commands
 
 subsection:: /d_recv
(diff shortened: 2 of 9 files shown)


elgiano:
I've sketched a supernova implementation, seems to work. Other people, please have a look, it's definitely not the part of the code where I feel most at home :D

capital-G (approved):

capital-G:
@telephon can you revert this merge as the linting was not good in this one, see https://github.com/supercollider/supercollider/actions/runs/10837644650/job/30074037345?pr=6467

capital-G:
Created a PR to fix the issue https://github.com/supercollider/supercollider/pull/6474
@telephon please merge before people create PRs from current dev

dyfer:
To avoid accidental merge in the future, should we add a rule to branch protection that requires a passing at least the lint CI? I think that can still be overridden by checking a box, and we could revert that change if we don't like it.
We could also require the full CI to pass, but with occasional issues with the build pipeline, maybe that's not a good option. Another option would be to require Linux builds to pass, they seem to be the most solid.

capital-G:
IMO the current workflow is in a good state - there will be errors and if these happens, the CI will make them aware to us and a fix such as #6474, while not nice, is not much work and can be merged fast. Essentially we got dev into a "bad state" for one commit and a couple of minutes, but I think it is fine to occasionally break develop.

I'd say we should add this additional checks and protections if we repeatedly run into such problems, this seems like a one off.

dyfer:
Sounds good.
FWIW we have a protection like that in sc3-plugins. I can't see I like it better or not - but the red button "merge without checks" or something like that is noticeable.

elgiano:
Hello, I'm really sorry I forgot the linting, I've set up pre-commit now. Thanks for fixing this and sorry for the trouble!

However, I wasn't expecting to see this merged so fast, and as I wrote in the comment above, I would have liked somebody else to review especially the supernova part. I don't want to sound like we shouldn't have merged it, but I would still like an extra review.

@joshpar and @Spacechild1 are the first that come to my mind, I remember having discussed API changes with them in the past, so if you have the time, I would be happy if you could look at these changes.

Spacechild1:
@elgiano Looks good to me! No complaints from my side :)

PR #6469, opened 2024-09-13
GHA: macOS: update universal script

dyfer:
## Purpose and Motivation

I updated the script for creating "universal" homebrew binaries, fixing some bugs...

I removed the HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK variable, as things should now work without it. This will remove many of the warnings posted underneath the build actions.

Note: this may or may not help with recent build failures on macOS.

EDIT: I'm not certain how solid this fix is, but it does seem that installing all the libraries in homebrew first and enabling dependents checking resulted in a properly finished and linked build.

We might be also hitting against stopping support in homebrew for macOS 12 - when running `brew update` (which this PR does NOT currently add), we get messages that it's not supported anymore, see the [build job here](https://github.com/supercollider/supercollider/actions/runs/10843097310). More changes might be needed, but this script update should probably be merged for now.

Also, I will eventually make a little write up about the script and the issues we've been seeing with `verify_app`.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- n/a Updated documentation - there's NO documentation as this script is a temporary workaround...

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -64,9 +64,9 @@ jobs:
       BUILD_PATH: ${{ github.workspace }}/build
       INSTALL_PATH: ${{ github.workspace }}/build/Install
       HOMEBREW_NO_ANALYTICS: 1
+      HOMEBREW_NO_ENV_HINTS: 1
       HOMEBREW_NO_AUTO_UPDATE: 1
       HOMEBREW_NO_INSTALL_CLEANUP: 1
-      HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
       USE_SYSLIBS: ${{ matrix.use-syslibs }}
       SHARED_LIBSCSYNTH: ${{ matrix.shared-libscsynth }}
       ARTIFACT_FILE: "SuperCollider-${{ inputs.sc-version }}-${{ matrix.artifact-suffix }}.dmg"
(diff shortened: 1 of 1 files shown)


dyfer:
Alternatively, we could - at least temporarily - separate x64 and arm64 builds and get rid of the script altogether (or stop using it for now, then decide later what the best course of action is)

capital-G:
I just took a look at this and it seems like Qt is really hesitant for universal builds - I don't understand why they are working so much against the FOSS community, not providing universal builds for a CI environment is a really annoying move.. /rant

Although I would really like to provide universal builds, they are really fragile and create lots of maintenance effort - and in the end they don't provide much value other than "just download that one file" instead of "please choose based on your architecture".

Also, I am afraid macOS will soon phase out universal binaries to further push Apple Silicone, e.g. GitHub actions is also already phasing out x64 runners which is not a good sign? In an effort to provide as long as possible support for Intel Macs, maybe we should now settle for a dedicated x64 build (via cross compilation), labeled as our new "legacy-macOS" build? 😢 [myself is still using an intel mac, so I am somehow interested to keep this build^^].

dyfer:
All valid points @capital-G - I have many thoughts on this but let's save this for another issue - I'm planning to lay this out and discuss provided buids and supported platforms etc. Specifically the official Qt6 binary is provided as a universal build and our problem would be "only" with universal fftw, portaudio, and readline (which take 10 minutes without cache to prepare now - waaay to long). Again, I'll write about this in a separate issue.

For this PR, in the _short_ term: 
- universal build takes a long time, however it seems to work again with this fix
- I think that we could either merge this as is, or we can split the builds for x68_64 and arm64 just so that we still have develop builds for both platforms
  - I would treat this as a potentially temporary change - this is something to discuss in the issue I am about to write, as mentioned above.

So... merge this or should I add separating the builds?

capital-G (approved):
I think we can merge and see how it develops over time.
Feel free to merge this any time :)

If you want some help or discussion about the supporting platforms we can create a shared scratchpad or have a small Jitsi meeting - just as an offer, if you don't want to do it alone.

OT: I tried out to setup MacPorts today and it isn't as slow as I thought it would be https://github.com/supercollider/supercollider/compare/develop...capital-G:supercollider:mac-ports-ci , see https://github.com/capital-G/supercollider/actions/runs/10852676884/job/30119297693
But I found the caching is hard - feasible but I think it will be really slow without it.

dyfer:
> If you want some help or discussion about the supporting platforms we can create a shared scratchpad or have a small Jitsi meeting - just as an offer, if you don't want to do it alone.

Thanks! As a first step I'll write up an issue - I have thought about it a lot on and off and need to put it out in the written format. That would be a good starting point to disccuss further I think - with all the suggested formats being a possibility for sure.

PR #6470, opened 2024-09-13
GHA: update pre-commit action

dyfer:
## Purpose and Motivation

This should remove the warning in GHA about the action `cache@v3` and deprecated node.js version.

BTW, the [readme of the pre-commit action](https://github.com/pre-commit/action) says it's in maintenance mode... Does it mean it might need to be replaced at some point?

## Types of changes

- Maintenance

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           python-version: '3.10'
 
-      - uses: pre-commit/action@v3.0.0
+      - uses: pre-commit/action@v3.0.1
         name: 'Run pre-commit'
         env:
           FULL_CHECK: 1


capital-G:
> BTW, the [readme of the pre-commit action](https://github.com/pre-commit/action) says it's in maintenance mode... Does it mean it might need to be replaced at some point?

Their intention of 3.0 release was to push their own github actions alternative pre-commit.ci, so they deprecated it - but there was so much backlash that they said it would be put into maintenance mode in 3.0.1. Notice that this is only affecting the pre-commit-CI stuff, not about pre-commit in general.

In the end, we could just copy/past the few steps here directly and get rid of the action

https://github.com/pre-commit/action/blob/f7acafac0271bdd064cdfa1b13f17b4350e565ed/action.yml#L11-L20 could be translated to

```shell
# assuming python is setup
pip install pre-commit
pre-commit run --all-files --show-diff-on-failure --color=always
```

We could also add caching via the `setup-python` action if we want to.

capital-G (approved):

capital-G:
See https://github.com/supercollider/supercollider/pull/6472 which would get rid of this action altogether

dyfer:
Thanks! Merging this for now to get rid of the warning.

> ```shell
> # assuming python is setup
> pip install pre-commit
> pre-commit run --all-files --show-diff-on-failure --color=always
> ```

> We could also add caching via the setup-python action if we want to.


I think I'd like that! That would be more transparent.

PR #6472, opened 2024-09-13
Dedicated lint workflow to verify all files in CI

capital-G:
## Purpose and Motivation

Currently the CI only runs when

https://github.com/supercollider/supercollider/blob/e49e42cbf42be6dc9ff4806cf14ff9b9d14e0bb5/.github/workflows/actions.yml#L1-L14

Though we may want to also run linting checks on these ignored files, see e.g. #6346 which would introduce such checks and #6451 which could benefit from such checks. If we enforce checks via pre-commit which are not checked in CI, this can cause a failing pre-commit to propagate once we actually run the check on these files. As pre-commit is quite lightweight, it should be run every time.

__UPDATE__: Instead of running an additional lint workflow every time I simply added a new workflow which runs on the files we normally ignore, see https://github.com/capital-G/supercollider/actions/runs/12206666964/job/34056591335 of a test commit https://github.com/capital-G/supercollider/commit/782982c17e46b3c6784e28884d6c5cfaf49df47c

~~In order to verify all files at every push we need a dedicated workflow as GitHub Actions triggers can only be induced on a "per-workflow" level and not on a "per-job" level (so it is not possible to "only run this job when these files changed", see [here](https://github.com/orgs/community/discussions/25669)).~~

~~But due to this split of workflows, we loose the ability to "depend" on the linting step for other jobs - yet I think this is negligible as the CI would still fail and it may be interesting to see if the build/tests are OK even if the formatting is off. If it is still desirable that build/tests only run when the formatting/linting is valid, we could run pre-commit two times (one time dedicated, one time as job within the CI workflow), which would result in some code duplication within CI.~~

## Types of changes

- New feature

## To-do list

### DIFF
--- a/.github/workflows/lint_non_code.yml
+++ b/.github/workflows/lint_non_code.yml
@@ -1,0 +1,37 @@
+name: Lint non-code files
+on:
+  push:
+    paths:
+      - 'examples/**'
+      - 'HelpSource/**'
+      - 'sounds/**'
+      - '*.md'
+  pull_request:
+    paths:
+      - 'examples/**'
+      - 'HelpSource/**'
+      - 'sounds/**'
+      - '*.md'
+
+jobs:
+  lint-non-code-files:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: lint class library
+        run: |
+          sudo npm install -g lintspaces-cli
+          lintspaces -e .editorconfig SCClassLibrary/**/*.sc || true # ignore failure
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - uses: pre-commit/action@v3.0.1
+        name: 'Run pre-commit'
+        env:
+          FULL_CHECK: 1


dyfer:
I think we _should_ keep depending on the linting for the other build jobs. That was the idea when the clang-format was added to GHA workflow.

I think that I would rather implement this so that we run whole CI on more files, including `examples` and `HelpSource`, but add a check in the lint step if these are the only files changed (like in the example you linked), and if so, skip the latter part of the CI.

capital-G:
> I think we _should_ keep depending on the linting for the other build jobs. That was the idea when the clang-format was added to GHA workflow.
> 
> I think that I would rather implement this so that we run whole CI on more files, including `examples` and `HelpSource`, but add a check in the lint step if these are the only files changed (like in the example you linked), and if so, skip the latter part of the CI.

While this would be preferable, it is not really feasible via GitHub w/o some hacking. With this PR the CI would still fail, only difference is that it would not stop but also continue with building and testing.
Or we just decide to run the CI every time - we don't have too many PRs which are targeted at these files.

dyfer:
> While this would be preferable, it is not really feasible via GitHub w/o some hacking.

I thought we just run check on which files changed, set the output of the action based on that, and depend on that output in other CI jobs?

Otherwise I'd be fine including the folders in question to trigger CI, they are seldom changed indeed.

dyfer:
@capital-G what do you think about my suggestion to just run our regular CI for `HelpSource/**` folders as well?

capital-G:
> @capital-G what do you think about my suggestion to just run our regular CI for `HelpSource/**` folders as well?

I think we could be a bit broader and just run every time a PR is made? There would still be the possibility to skip the CI via commit names, see  https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs
Or do you have objections to this?

dyfer:
I don't have strong objections, but I don't see the point on running CI on changes in `sounds/**` and on `*.md`, so I'd keep these exclusions.

dyfer:
OTOH... thinking about this again, maybe the helpfile changes are not worth running the full CI? maybe a better solution would be to have two linters, one _only_ for PRs and HelpSource, the other as is...

capital-G:
@dyfer can you give this another review? I simply added an "inversed" lint workflow which only runs if any of the ignored files of our default-ci are changed.

dyfer:
Looks good.
One thing I'd consider - do we really want this on every push? I think this could be PR only + on demand.

Maybe that's another PR/Issue to discuss, but I'd be leaning towards generally running CI on push requests + main branches (not this PR though?) + on demand.

What do you think?

capital-G:
> One thing I'd consider - do we really want this on every push? I think this could be PR only + on demand.

Linting only takes 1 minute, so I think it is fine? I think it is best to move people towards pre-commit as it allows to get rid of "clang-format" commits in between which taint the history. Having running this linter on every push is an additional reviewer reminding you of this and narrows down what went wrong when.

Testing takes longer, but on the other hand: If I want to review a PR I am glad that the result of the CI is already there instead of waiting 20+ minutes for it in which I most likely had moved on to something else.

IMO the status quo is fine - or what would be your motivation?
E.g. I am currently working on a branch and having a CI w/o a PR is really comfortable for me.

The only thing I found a bit distracting were the weekly CI runs, which also happened in forks and due to a failing pipeline resulted in a weekly mail. But these have been turned off already.

dyfer (approved):
Thanks!

dyfer:
> IMO the status quo is fine - or what would be your motivation?

Limiting unnecessary CI runs. E.g. when you make a PR from the fork, the CI runs both on the fork and in the PR in the main repo.

But that's probably not a big concern for now.

I want to stress that if we were to limit CI runs, we'd add the option to also run it on demand, so that you could still run it on any branch, just not automatically; just you'd need to trigger the run from the web UI.

PR #6473, opened 2024-09-13
UnitTest, TestBus: improve performance, reduce number of control busses.

JordanHendersonMusic:
## Purpose and Motivation

`TestBus.test_controlFree` is slow. This reduces the number of control busses on the server to make this test faster. 
Reason being that since the logic does not depend on size, it can be small.
 
```
Old : time to run: 8.9728592389997 seconds.
New : time to run: 0.0006905000000188 seconds.
```

A part of #6439.
## Types of changes

- Performance

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestBus.sc
+++ b/testsuite/classlibrary/TestBus.sc
@@ -28,6 +28,7 @@ TestBus : UnitTest {
 	test_controlFree {
 		var s,busses;
 		s = Server(thisMethod.name);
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Well spotted, thanks!

If we decide at some point to make the tests independent of the server we should maybe revisit this configuration, but this would be up for another PR.

PR #6476, opened 2024-09-13
Revert pinning boost version for macOS system libs

dyfer:
## Purpose and Motivation

IIUC we don't need this after #6331 was merged.
This gets rid of the warning about the versioned boost formula in the CI.

## Types of changes

- Bug fix (?)

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


capital-G (approved):
Thanks for thinking on reverting this hack.

Boost 1.86 PR is also incoming.

PR #6477, opened 2024-09-14
Upgrade to boost 1.86

capital-G:
## Purpose and Motivation

This boost update helps us to get native macOS 14 compilation support without any hacking through compiler flags and resolves many deprecation warnings and should provide us a good basis for the next years.

Most of the update work was done by replacing `boost::filesystem` with the C++17 `std::filesystem` implementation in #6331 .

This PR supersedes and closes
* https://github.com/supercollider/supercollider/pull/6246
* https://github.com/supercollider/supercollider/pull/6334
* https://github.com/supercollider/supercollider/issues/6224

I upgraded by deleting the existing `boost` folder and using the provided `extract_boost.sh` to generate a new/fresh boost extraction.
The prior deletion was done in order to get rid of `boost::filesystem`.

> This will break the flucoma plugin, as they relied on some boost internals which are not exposed through the plugin API, so the ABI will break which requires them to rebuild their SuperCollider plugin, there is nothing we can do about this.
They are aware of the situation and will be contacted once more if this gets pushed through.

## How to review

I don't think one needs to review boost code, therefore the interesting bits and questions would be

* `CMakeFiles`
* Should we still apply the  `boost::chrono` performance patch from https://github.com/supercollider/supercollider/blob/develop/external_libraries/boost_sc_changes.patch? Or is the compiler "smart enough" on its own to inline it? I have a PR with CI where these patches are applied @ https://github.com/capital-G/supercollider/actions/runs/10860248830
* The same for mingw patch - not applied here, as it seems to build without? Can someone test it though?
* I defined a new minimum version of boost of 1.80 in https://github.com/capital-G/supercollider/blob/89e255a55dff69d220da4ef4253d708b27415168/CMakeLists.txt#L325 - this is somehow a guess as it was 1.50 and we actually need less of boost then before, but maybe it is good to occasionally raise the bar? Especially since using the system version of boost is an edge case.

Beyond that I'd like to put any additional edits and requests into their dedicated PR after this has been merged as this PR is really large and therefore hard to review.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer:
Thanks!

> * `CMakeFiles`

We should probably look on the feedback I had for https://github.com/supercollider/supercollider/pull/5924 - I'll try to review later.

> * Should we still apply the  `boost::chrono` performance patch from https://github.com/supercollider/supercollider/blob/develop/external_libraries/boost_sc_changes.patch? Or is the compiler "smart enough" on its own to inline it? I have a PR with CI where these patches are applied @ https://github.com/capital-G/supercollider/actions/runs/10860248830
> * The same for mingw patch - not applied here, as it seems to build without? Can someone test it though?

AFAIR we don't need any of the patches anymore. Newer mingw seems to work without the patch (as proven in the CI), IIRC; I don't think `boost::chrono` was needed either. IMO the patch file should be deleted, either in this PR or in a separate one.

> * I defined a new minimum version of boost of 1.80 in https://github.com/capital-G/supercollider/blob/89e255a55dff69d220da4ef4253d708b27415168/CMakeLists.txt#L325 - this is somehow a guess as it was 1.50 and we actually need less of boost then before, but maybe it is good to occasionally raise the bar? Especially since using the system version of boost is an edge case.

I'm not sure if we need to raise this here. 1.7x would likely still work. IMO there's no harm in keeping this low (as is), even if it proves inaccurate eventually - but let me know if you can think of a reason where keeping it as is might pose a problem.

server/scsynth/CMakeLists.txt
```diff
@@ -231,7 +231,6 @@ elseif(AUDIOAPI STREQUAL coreaudio)
 	target_link_libraries(libscsynth "-framework CoreAudio")
 endif()
 
-target_link_libraries(libscsynth boost_system_lib boost_filesystem_lib)
```
dyfer, line 234: sclang and supernova still link with `boost_system_lib`, is that not needed here? (just noticing the difference, I don't know about the functionality etc)

dyfer:
Boost 1.80.0 PR also had [this change](https://github.com/supercollider/supercollider/pull/5924/files/fb3c4d70f3b419d803a38e897b3832eac4f1ec26..fed94aefa5b8f6527dae4f09aa8997d67ad84d9b#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR148). Don't we need it in https://github.com/supercollider/supercollider/pull/6477/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR164 ?

capital-G:
Thanks for taking a look at it!

> Boost 1.80.0 PR also had [this change](https://github.com/supercollider/supercollider/pull/5924/files/fb3c4d70f3b419d803a38e897b3832eac4f1ec26..fed94aefa5b8f6527dae4f09aa8997d67ad84d9b#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR148). Don't we need it in https://github.com/supercollider/supercollider/pull/6477/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR164 ?

Most likely we don't need it as we removed `boost::filesystem` altogether?

> IMO the patch file should be deleted, either in this PR or in a separate one.

Agree it should be deleted if not needed anymore. Maybe it is good to attach it to this PR so it can be found easily if it ever comes up again?

> I'm not sure if we need to raise this here. 1.7x would likely still work.

So should we put it to `1.7.0`?

> sclang and supernova still link with `boost_system_lib`, is that not needed here? (just noticing the difference, I don't know about the functionality etc)

Interesting that it still compiles - but yeah, I'll include it again.

dyfer:
> > Boost 1.80.0 PR also had [this change]
> 
> Most likely we don't need it as we removed `boost::filesystem` altogether?

Aaah, of course you're right. Completely missed that connection.

> > IMO the patch file should be deleted, either in this PR or in a separate one.
> 
> Agree it should be deleted if not needed anymore. Maybe it is good to attach it to this PR so it can be found easily if it ever comes up again?

I think it can always be retrieved from the git history... And it will still be in previous releases' source packages. I don't think we need to save it elsewhere.

> 
> > I'm not sure if we need to raise this here. 1.7x would likely still work.
> 
> So should we put it to `1.7.0`?

I'd keep it unchanged until it becomes an issue.

> 
> > sclang and supernova still link with `boost_system_lib`, is that not needed here? (just noticing the difference, I don't know about the functionality etc)
> 
> Interesting that it still compiles - but yeah, I'll include it again.

Maybe none of `boost_system_lib` includes are needed then? :) who knows.

dyfer:
One more observation: I think the removal of `boost_filesystem` links from CMakeLists belongs more to the changes from the `std::filesystem` PR than boost upgrade. Could these changes be in a separate commit here?

capital-G:
> One more observation: I think the removal of `boost_filesystem` links from CMakeLists belongs more to the changes from the `std::filesystem` PR than boost upgrade. Could these changes be in a separate commit here?

I am somehow unsure how CMake handles these files? I think while #6331 removed all references of `boost::filesystem`, there were (possibly) still references within `boost` itself to filesystem as `boost::filesystem` was still compiled (but maybe CMake is not "smart enough" to detect an unused import as we still referenced it?)

I could rebase the PR, but in the end I think anyone who will investigate this PR will also soon pick up the std filesystem change?

New changes:

* Deleted the patch file: c01bc6ab1f6ebbccddffa47383c92ae612f84085
* Reverted to boost minimum v 1.5.0 (untested): 73c2478b7f22d92ba9af96cf621d902f00f077c6

Also I noticed a really unstable macOS CI which fails at different spots during compilation - really strange but also no consistent failures.

dyfer:
Thanks for the changes.

> (but maybe CMake is not "smart enough" to detect an unused import as we still referenced it?)

I'm guessing that we were still linking to these libs, since they were still listed in CMake, but I might be wrong.

I think this PR now has everything I'd like it to, but IMO it would be great to re-organize the commits, if that's not too much hassle. I think the following should be separate commits:
- the boost update alone (`external_libraries/boost` and `external_libraries/CMakeLists.txt`)
- Other CMake changes removing linking to `boost::filesystem`
- Removing the patch, along with the readme entry about it


> Also I noticed a really unstable macOS CI which fails at different spots during compilation - really strange but also no consistent failures.

Yeah, I saw that too, probably something on the GHA side.

capital-G:
> I think this PR now has everything I'd like it to, but IMO it would be great to re-organize the commits, if that's not too much hassle. I think the following should be separate commits:
> 
>     * the boost update alone (`external_libraries/boost` and `external_libraries/CMakeLists.txt`)
> 
>     * Other CMake changes removing linking to `boost::filesystem`
> 
>     * Removing the patch, along with the readme entry about it

Done

dyfer (approved):
Thanks @capital-G!

mtmccrea:
This is great, thanks @capital-G and @dyfer!

mtmccrea:
Are any build flags or other updates needed to accomadate this change?
I'm getting a build error:
```
SC_Filesystem_macos.cpp:92:13: error: use of undeclared identifier
      'assert'
            assert(false && "Resolved path could not be converted to C string");
            ^
1 error generated.
```

capital-G:
> Are any build flags or other updates needed to accomadate this change? I'm getting a build error:
> 
> ```
> SC_Filesystem_macos.cpp:92:13: error: use of undeclared identifier
>       'assert'
>             assert(false && "Resolved path could not be converted to C string");
>             ^
> 1 error generated.
> ```

Which macOS are you using? On macOS 14 it should work w/o any additional flags.

Normally you don't need any additional flags. Just to be sure, clean/remove your build and cache directory and please try again.

mtmccrea:
I was working on a clean build, and just running that standard build sequence
```bash
cmake -G Xcode -DCMAKE_PREFIX_PATH=`brew --prefix qt@5`  ..
cmake --build . --target install --config Release
```
So I went ahead and made the obvious fix in _SC_Filesystem_macos.cpp_, and added
```c++
#include <cassert>
```
And that fixed it. Surprising this wasn't caught in CI, and that cassert wasn't included somewhere upstream in the include chain. Could this be the fix?

capital-G:
> And that fixed it. Surprising this wasn't caught in CI, and that cassert wasn't included somewhere upstream in the include chain. Could this be the fix?

I just re-ran a clean build with the same commands on 14.6.1 without any problems (`** BUILD SUCCEEDED **`).
In addition that the CI also passes I think there is actually an issue on your end?

What gets returned by

```
❯ clang --version
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin23.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

on your machine?

Also can you share your full configure log?

```
❯ cmake -G Xcode -DCMAKE_PREFIX_PATH=`brew --prefix qt@5`  ..
-- The C compiler identification is AppleClang 14.0.0.14000029
-- The CXX compiler identification is AppleClang 14.0.0.14000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- SuperCollider Version: 3.14.0-dev
-- Building from branch develop, commit hash is 0cec549c01
-- Please specify the build configuration in the next step
-- Compiling with Qt GUI
-- Found Pthreads: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/lib/libpthread.tbd
-- Using bundled boost
-- Using bundled yaml-cpp
CMake Deprecation Warning at external_libraries/hidapi/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- HIDAPI components:
--     mac
--     hidapi_parser
-- Building with HID support
-- Found install of libsndfile in /usr/local/opt/libsndfile/lib/libsndfile.dylib
-- Performing Test SNDFILE_HAS_VORBIS
-- Performing Test SNDFILE_HAS_VORBIS - Success
-- Performing Test SNDFILE_HAS_OPUS
-- Performing Test SNDFILE_HAS_OPUS - Success
-- Performing Test SNDFILE_HAS_MPEG
-- Performing Test SNDFILE_HAS_MPEG - Success
-- FFT library (scsynth): vDSP
-- Audio API (scsynth): coreaudio
-- Found a homebrew install of readline /usr/local/opt/readline
-- Found GNU Readline version 8.2: /usr/local/opt/readline/lib/libreadline.dylib
-- Found Qt: /usr/local/opt/qt@5/lib/QtCore.framework/QtCore
-- sclang: Building with QtWebEngine
-- Compiling with Ableton Link support
-- Compiling with Readline support
-- Building the Qt IDE
-- IDE: Building with QtWebEngine
-- Found Qt deploy utility: /usr/local/opt/qt@5/bin/macdeployqt
-- Configuring done (22.9s)
-- Generating done (0.1s)
-- Build files have been written to: /Users/scheiba/github/supercollider/build
```

and, just to be sure

```
❯ git rev-parse HEAD
0cec549c015a96b088d85510110149f6c5969dd8
```

capital-G:
Ah, updated to Clang15 and now I also run into a build error. @mtmccrea do you want to create a PR to fix this? :)

mtmccrea:
Looks like you beat me to it, thanks for your help!

shimpe:
Kudos to everyone who worked on this until it was finalized!

PR #6478, opened 2024-09-14
QtCollider: sunken inset shadows respect alpha

### LINKED ISSUE #2574: EnvelopeView does not properly render background color or transparency
```
(
 // make a simple SoundFileView
y = Window.screenBounds.height - 120;
w = Window.new("soundfile test", Rect(200, y, 740, 100));
w.front;
t = ScrollView(w, w.view.bounds);
a = SoundFileView.new(t, Rect(20,20, 700, 60));

f = SoundFile.new;
f.openRead(Platform.resourceDir +/+ "sounds/a11wlk01.wav");
// f.inspect;

a.soundfile = f;            // set soundfile
a.read(0, f.numFrames);     // read in the entire file.
a.refresh;                  // refresh to display the file.
EnvelopeView(t, a.bounds).background_(Color.clear); // should be transparent but isn't
)
```

envview

### LINKED ISSUE #3704: Slider2D background transparency is not as expected
If view A = `View()` and view B = `View(a, ...)`, and I set B's background color to anything where alpha = 0, I expect the background to be fully transparent -- that is, you should see A's colors coming through the background unmodified.

This works correctly for two View instances, or View and UserView, but it does not work as expected if B is, for instance, a Slider2D.

```
z = Image(10, 10);
10.do { |x| 10.do { |y| z.setColor(Color.gray(1.0.rand), x, y) } };

v = View(nil, Rect(800, 200, 100, 100)).front;
v.setBackgroundImage(z, 10);

u = View(v, v.bounds.moveTo(0, 0)).background_(Color.clear);
u.background = Color(1, 1, 1, 0);
u.background = Color(0.5, 0.5, 0.5, 0);
```

When adding `u` on top of `v`, the display does not change -- `u` is completely transparent.

Continuing:

```
u.remove;

u = Slider2D(v, v.bounds.moveTo(0, 0)).background_(Color.clear);  // darker
u.background = Color(1, 1, 1, 0);  // lighter
u.background = Color(0.5, 0.5, 0.5, 0);  // lower contrast
```

It looks to me like, in this case, alpha = 0 is producing the display I would expect for alpha = 0.5.

I don't have time right now to test all views.

It's funny -- I was trying this approach to work around a bug with #2575 (you can't apply a background image to a Slider2D directly), when... surprise, surprise... I find a bug with the workaround. So close to something kind of cool and then... the only way to get what I really want is to re-implement Slider2D's logic as a UserView...? Fun. (I'll accept the faded colors for now.)

elgiano:
## Purpose and Motivation
Fixes #3704 
Fixes #2574

Some widgets (e.g. EnvelopeView and Slider2D) have a "sunken" effect, obtained by drawing a semi-transparent rectangle on top of their background. Since the alpha value for such shadows is hardcoded, a user can't make such widgets have a transparent background.

This PR makes the shadow's alpha dependant on the widget's background alpha, thus making it possible to achieve full transparency.

Example code for comparison:
```
// a green window with a transparent red widget should look green
var win = View(bounds:200@200).background_(Color.green);
var ev = EnvelopeView().background_(Color.red(1, 0));
win.layout = VLayout(ev);
win.front
```
Here the square on the left is before PR, and the one on the right is after PR:
image

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/QtCollider/style/routines.hpp
+++ b/QtCollider/style/routines.hpp
@@ -160,7 +160,7 @@ void drawSunken(QPainter* p, const QPalette& plt, const ShapeT& shape, const QCo
     // shadow
 
     QColor c2 = color;
-    c2.setAlpha(150);
+    c2.setAlpha(color.alpha() * 0.5);
     p->setPen(Qt::NoPen);
     p->setBrush(c2);
     shape.draw(p, shape._rect.adjusted(2, 2, -2, -2));


QtCollider/style/routines.hpp
```diff
@@ -160,7 +160,7 @@ void drawSunken(QPainter* p, const QPalette& plt, const ShapeT& shape, const QCo
     // shadow
 
     QColor c2 = color;
-    c2.setAlpha(150);
+    c2.setAlpha(color.alpha() / 2);
```
telephon, line 163: better: `c2.setAlpha(color.alpha * 0.5);`

elgiano: right, fixed! thanks

capital-G: Why though is `color.alpha * 0.5` *better* than `color.alpha() / 2`? From a computing perspective `/2` is a cheap operation and therefore commonly used from low- to high-level, and also the common mathematical notation would be `/2`.

In the end I am fine with either way, but I think it is hard to argue and reason about this.

capital-G (approved):
Thanks for spotting the error and fixing it!

capital-G:
@dyfer I see you linked this PR to #5169 but I actually can't find it how exactly this is mentioned there.
As you are doing major rework on the Qt part I think you can estimate best if this PR is best pushed after the Qt6 PR or before (in the latter case, please feel free to merge) - in the other case, please add it to the Qt6 milestone and add a `blocked` tag to the PR.

dyfer:
> @dyfer I see you linked this PR to #5169 but I actually can't find it how exactly this is mentioned there. 

Sorry! That was a typo and I already fixed that, but the reference seemed to stick around. This PR is not related to Qt6 in any way AFAICT.

Merging!

Issue #5296, opened 2020-12-21
plotMode=\bars does not draw the last element


tremblap:
Hello!

`(0..9).plot.plotMode=\bars;`

is unexpectedly drawing only 9 bars. My expectations might be wrong, but a 10 item array, I expected 10 bars. Boundary issues I reckon.

prko:
\levels and \steps have the same behaviour.
https://scsynth.org/t/plot-bars-mode-problem/8683/7?u=prko

mtmccrea:
>\levels and \steps have the same behaviour.

Yes, and my initial impression is that they should have different behavior. I think both `\levels` and `\bars` would be better suited to be centered on their domain values, which would require some special handling of the `domainSpecs` in those plots to allow for some padding at either end of the plot to accommodate.

`\steps` is ok because it follows the sample-and-hold paradigm, which seems appropriate. There could also be some consideration of padding on the ends of this plot mode, but it would have to conform to proper display of duration in the case of, e.g. plotting audio sample values.

also discussed in [#6479 ](https://github.com/supercollider/supercollider/pull/6479#issuecomment-2355188852)

tremblap:
I think i disagree. in an oversampling case (like here, 10 values for 200 pixels) a good method might be to centre the 10 values, like a histogram. In any case, it should have 10 values equally represented in pixel.

as for SC's approach to undersampling, it is not ideal either for soundwaves. A good practice in DAWs is to take the absolute maximum distance to zero in all the samples that are 'represented' by that pixel, which is not what SC is doing, which leads so surprising drawings and commissions. other cool modes are maximum difference from the past pixel (good for disrupting waves) but hey, that would need a complete rewrite - I had that discussion many times with many SC users and I think this report is the fruit of the last one - pointing at the lowest hanging fruit.

prko:
The following three modes display the correct number of stems:
```
(0..9).plot.plotMode=\dstems 
(0..9).plot.plotMode=\pstems
(0..9).plot.plotMode=\stems
```
It would be better to leave the bars as they are for backward compatibility.

How about creating a new mode like this?
 `\bstems`
Where b stands for 'bar'.

I think bar mode is logically correct, but it is not practical. Sometimes users need thick stems.

mtmccrea:
Thanks for drafting an implementation of this :)

> I think bar mode is logically correct, but it is not practical. 

[...]

PR #6479, opened 2024-09-14
ClassLib: PlotView stepwise modes fix missing last step

elgiano:
## Purpose and Motivation

When plotting with plotModes `steps`, `bars` and `levels`, the last value becomes invisible, because its left edge is collapsed with the right margin of the plot.

This PR fixes it, the only question I have is whether plotMode `steps` has to be included or not, since its last value is actually visible as the end of a vertical line.

image
> [1,2,3] plotted as `\plines`, `\steps`, `\levels` and `\bars`, before (left) and after (right) PR.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -373,14 +373,17 @@ Plot {
 		};
 	}
 
-	domainCoordinates { |size|
+	domainCoordinates { |size, mode|
 		var range, vals;
+		// step-like displays need one more x coordinate
+		var isSteplike = this.hasSteplikeDisplay(mode);
+		if (isSteplike) { size = size + 1 };
 
 		vals = if (plotter.domain.notNil) {
 			domainSpec.unmap(plotter.domain);
 		} {
 			range = domainSpec.range;
-			if (range == 0.0 or: { size == 1 }) {
+			if (range == 0.0 or: { size == 1 } and: {isSteplike.not}) {
 				0.5.dup(size) // put the values in the middle of the plot
 			} {
 				domainSpec.unmap(
@@ -399,7 +402,6 @@ Plot {
 
 	drawData {
 		var ycoord = this.dataCoordinates;
-		var xcoord = this.domainCoordinates(ycoord.size);
 
 		Pen.use {
 			Pen.width = 1.0;
@@ -410,16 +412,24 @@ Plot {
 			if(ycoord.at(0).isSequenceableCollection) { // multi channel expansion when superposed
 				ycoord.flop.do { |y, i|
 					var mode = plotMode.wrapAt(i);
+					var xcoord = this.domainCoordinates(ycoord.size, mode);
 					Pen.beginPath;
 					Pen.fillColor = plotColor.wrapAt(i);
 					Pen.strokeColor = plotColor.wrapAt(i);
 					this.perform(mode, xcoord, y);
 				}
 			} {
+				var mode = plotMode.at(0);
+				var xcoord = this.domainCoordinates(ycoord.size, mode);
 				Pen.beginPath;
 				Pen.strokeColor = plotColor.at(0);
 				Pen.fillColor = plotColor.at(0);
 				this.perform(plotMode.at(0), xcoord, ycoord);
+				if (this.needsPenFill(mode)) {
+					Pen.fill
+				} {
+					Pen.stroke
+				}
 			};
 			Pen.joinStyle = 0;
 		};
@@ -565,7 +575,7 @@ Plot {
 	}
 
 	editData { |x, y, plotIndex|
-		var index = this.getIndex(x);
+		var index = this.getIndex(x, plotIndex);
 		this.editDataIndex( index, x, y, plotIndex );
 	}
 
@@ -576,8 +586,8 @@ Plot {
 		var val;
 
 		// get indexes related to ends of the line
-		i1 = this.getIndex(pt1.x);
-		i2 = this.getIndex(pt2.x);
+		i1 = this.getIndex(pt1.x, plotIndex);
+		i2 = this.getIndex(pt2.x, plotIndex);
 
 		// if both ends at same index, simplify
 		if( i1 == i2 ) {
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks @elgiano 
I'm not sure about this. In your change the `steps`/`levels`/`bars` don't line up with the points, i.e. the point falls somewhere "in the middle" of the step.

I think a better solution would be to extend the horizontal axis in these modes, so that they are of length `size + 1`, in which case the currently drawn lines would be visible, but the beginnings of the lines/bars would properly line up with the data grid.

mtmccrea:
To add to @dyfer's comment, it's important that the data aligns properly with x axis, in the context of discrete sampling of data (like audio). So the current behavior of `step` would be correct in terms of sample-and-hold behavior. 

When there is no duration specification, the x-axis shows sample indices. In your proposed change implies there are 4 data points instead of 3.

I agree there could be room for improvement here, though no obvious general solution comes to mind atm. An exceptional case could be `bars`, which could be centered on the data index. In that case the x axis (and grid) would have to be extended a half-step to the left _and_ right of the plot in order to accommodate a bar centered on its corresponding data index.

elgiano:
I agree with both of you and I'm thinking of a better solution... but as I don't have much time in the coming days, I'll leave some quick notes.

I think that for starters, \steps doesn't need to be changed, as it can behave exactly like \lines. This leaves us with \bars and \levels.

Without changing any ClassLib code, the simplest work-around would be:
```
x = [0,1,2].plot.plotMode_(\bars)
x.domain = (0..x.value.size-1)
x.domainSpecs = ControlSpec(0, x.value.size)
```
for superimposed graphs:
```
x = ((0..2)!2).plot.plotMode_([\plines, \bars]).superpose_(true)

x.domain = (0..x.value.shape.last - 1);
x.domainSpecs = ControlSpec(0, if([\bars, \levels].includesAny(x.plotMode)) { 
	x.value.shape.last 
} { 
	x.value.shape.last - 1
});
```
This has to be done again whenever plot value is updated with `findSpecs: true`.
But maybe the most consistent approach would be not to change anything (after all it's a consistent behavior, although confusing) and just document it with a note and an example?

mtmccrea:
I think the solution would be closer to this:
```supercollider
(
x = ((1..3)!2).plot.plotMode_([\plines, \bars]).superpose_(true);

x.domain_((0..x.data[0].size-1)); // explicitly set to data indices for introspection below

x.plotMode.includesAny([\bars, \levels]).if {
	x.domainSpecs_([x.domain.first-0.5, x.domain.last+0.5].asSpec) // assumes data domain is just the index value
};
x.specs_([x.value.flat.minItem-0.5, x.value.flat.maxItem+0.5].asSpec); // for clarity so first data point isn't hidden by grid bounds
)
```
where the bars would be properly centered on their domain value (imagine them centered on the `\plines` points)

>But maybe the most consistent approach would be not to change anything (after all it's a consistent behavior, although confusing) and just document it with a note and an example?

Yes, at least documentation could be improved!

mtmccrea:
And in any case \bars and \levels don't appear to behave correctly:
```supercollider
x = (1..3).plot.plotMode_(\bars)
x = (1..3).plot.plotMode_(\levels)
```

elgiano:
@mtmccrea 
do you mean with this PR? I see the three bars (like in the screenshot above), but the first has zero height because the default range for y-axis goes from 1 to 3. In this case, set `x.specs = ControlSpec(0,3)` to see the first as well

mtmccrea:
Hmmmm, on both `develop` and `3.13`, I get:
```supercollider
x = (1..3).plot.plotMode_(\bars)
```
Screenshot 2024-09-17 at 14 50 10

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -410,16 +412,24 @@ Plot {
...
 					this.perform(mode, xcoord, y);
 				}
 			} {
+				var mode = plotMode.at(0);
+				var xcoord = this.domainCoordinates(ycoord.size, mode);
 				Pen.beginPath;
 				Pen.strokeColor = plotColor.at(0);
 				Pen.fillColor = plotColor.at(0);
 				this.perform(plotMode.at(0), xcoord, ycoord);
+				if (this.needsPenFill(mode)) {
```
telephon, line 428: In the current design, the fill / stroke differece is handles in the plotMode itself. So `fill` or `stroke` should have been called already in `this.perform(plotMode.at(0), xcoord, ycoord);`.

Is there a reason this is needed?

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -373,14 +373,17 @@ Plot {
...
+		// step-like displays need one more x coordinate
+		var isSteplike = this.hasSteplikeDisplay(mode);
+		if (isSteplike) { size = size + 1 };
 
 		vals = if (plotter.domain.notNil) {
 			domainSpec.unmap(plotter.domain);
 		} {
 			range = domainSpec.range;
-			if (range == 0.0 or: { size == 1 }) {
+			if (range == 0.0 or: { size == 1 } and: {isSteplike.not}) {
```
telephon, line 386: I think you need parentheses: 
```
if (range == 0.0 or: { { size == 1 } and: { isSteplike.not } })
```

otherwise it will always require stepwise.

prko:
I think this could be fixed by adding a few padding spaces before the first value and after the last value on the x-axis:
https://github.com/supercollider/supercollider/issues/5296#issuecomment-2752837332

If this PR is not pursued, how about redesigning the coordinate grid and the size of its padding? Currently, the first and last values are partially hidden, and in some modes, this is not so obvious, but in some modes, these hidden parts cause a problem.

mtmccrea:
Hi @elgiano, would you like to keep working on this PR? If so, it would be to reconsider the current/latest behavior of the last horizontal "step/level" line, which isn't shown. 

The primary issue of this PR—displaying the last value—was partially addressed in #7089.  That PR fixed the last value issue for the `\bars` plot mode. It also and added plot modes `psteps`, `dsteps` , `plevels`, and `dlevels` to draw a marker (dot or point) for the values. 

```supercollider
x = (1..3).plot.plotMode_(\steps)
```
Screenshot 2026-05-08 at 12 42 36



So the last value can currently be visible as a marker, but the horizontal line is still truncated by the plot's x (domain) range. To see the last step/level line, the `domain` and `domainSpecs` would need to be set explicitly:

```supercollider
x = (1..3).plot.plotMode_(\steps).domain_((0..2)).domainSpecs_([0,5].asSpec)
```
Screenshot 2026-05-08 at 12 35 56

if we wanted to show the last step line, there would have to be logic added to extend the domain range under the hood (when `domainSpec` is unspecified) in the way that `\bars` now does.

PR #6485, opened 2024-09-16
correct .editorconfig syntax

xunil-cloud:
## Purpose and Motivation

The curly braces are mismatched. I believe this was a typo? I found it because my editor complained about it. 

NOTE: I'm not sure if this change will affect our linting behavior.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,6 +26,6 @@ indent_style = tab
 indent_size = 4
 
 # Matches the exact files either .travis.yml
-[*.{yml,yaml,json,js}}]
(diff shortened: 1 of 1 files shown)


capital-G:
Thanks for discovering this typo. This may change our linting on some files, but only if they are touched. Currently we don't have a way to automatically verify these files, so the config is treated more like a suggestion nonetheless.
But better if these suggestions are working.

capital-G (approved):

PR #6491, opened 2024-09-20
[tests] TestCoreUGen updates

mtmccrea:
## Purpose and Motivation

This update does some minor fixes and updates to the CoreUGen tests, which came up in the course of other work.

- the tolerance of the UGen equivalence tests was too permissive — `0.001` corresponds to a -60 dB tolerance, so that's been lowered to -90 dB. It could be a bit lower if not for the reversible test for `\midicps`<-> `\cpsmidi` test, which seems to have higher error because of the exponential operations involved. 
  - this change caught an erroneously passing `Integrator`/`Line` equivalence test.

- the delays/bufdelays tests and the reversible ops tests were refactored with a rate switch for a smaller footprint
  - Note that for the delays tests, previously the `kr` test was still running the Delay UGens at `ar`, and only switching the delaytime argument to `kr`. I assume this was an oversight so now those UGens under test, and those used for its inputs, are made to run at `kr`.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation: **N/A**

### DIFF
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -275,18 +275,14 @@ TestCoreUGens : UnitTest {
 			[\tan, \atan]
 		].do { |selectors|
 			[selectors, selectors.reverse].do { |pair|
-				tests = tests.add(
-					"x == %(%(x)) [control rate]".format(*pair) -> {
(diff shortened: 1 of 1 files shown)


dyfer (approved):
Looks good, thanks!

PR #6492, opened 2024-09-20
GHA macOS: don't stop if clearing homebrew downloads fails

dyfer:
## Purpose and Motivation

Recently macOS jobs started to fail on the step clearing homebrew downloads. I'm not sure what the problem is, but I propose to - temporarily - keep going even if that step fails. That should re-enable the build matrix.

This issue does not seem to occur on macOS 13/14 images (like in qt6 PR), so this might be needed only temporarily.

Still, from the logical standpoint, I think that failing to clear the downloads shouldn't probably stop the whole build.

## Types of changes

- Bug fix (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer:
If we pull in #6475, we probably won't need it; OTOH I think this is a valid addition, as it may prevent such failure in the future if something changes.

dyfer:
Superseded by https://github.com/supercollider/supercollider/pull/6493

PR #6493, opened 2024-09-22
GHA macOS: clear homebrew downloads only

dyfer:
## Purpose and Motivation

Currently CI on macOS fails.

I realized that we were previously deleting all homebrew cache, instead of just the downloads. This PR fixes that. It also prevents the recent failures in macOS GHA runners.

With this change we won't need #6492 anymore.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing


mtmccrea (approved):

capital-G:
Thanks for spotting this, lets merge.

PR #6497, opened 2024-09-25
Add <https://git-scm.com> to the Quark error message

prko:
Add a kind instruction to the error message for git not installed.

## Purpose and Motivation

Using Quarks is a barrier for newcomers. The current error message is not helpful in this regard:

 >ERROR: Quarks requires git to be installed

By adding the following sentence, newcomers can reduce the time to resolve the error:

>Follow the installation instructions at <https://git-scm.com>.

## Types of changes

- New feature: not exactly a new feature. Just an error message is a detailed explanation. 
 
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Quarks/Git.sc
+++ b/SCClassLibrary/Common/Quarks/Git.sc
@@ -148,7 +148,7 @@ Git {
 			Pipe.callSync(gitFind, {
 				gitIsInstalled = true;
 			}, { arg error;
-				"Quarks requires git to be installed".error;
+				"ERROR: Quarks requires git to be installed. Follow the installation instructions at <https://git-scm.com>.".error;
 				gitIsInstalled = false;
 			});
 		};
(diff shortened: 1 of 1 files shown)


mtmccrea (commented):
Thanks for this helpful message.

Note you don't need to add "ERROR: " because the `.error` method will do this for you. You can run
```supercollider
"This is an error message".error;
```
to see the result

SCClassLibrary/Common/Quarks/Git.sc
```diff
@@ -148,7 +148,7 @@ Git {
 			Pipe.callSync(gitFind, {
 				gitIsInstalled = true;
 			}, { arg error;
-				"Quarks requires git to be installed".error;
+				"ERROR: Quarks requires git to be installed. Follow the installation instructions at <https://git-scm.com>.".error;
```
capital-G, line 151:
```suggestion
				"Quarks requires git to be installed. Follow the installation instructions at <https://git-scm.com>.".error;
```

capital-G (changes requested):
Thanks, this is indeed a good hint to solve the error. I added the suggestion as mentioned by mtmccrea - once this has been fixed/applied, this can be merged.

capital-G (approved):

PR #6499, opened 2024-09-25
scsynth: add cmdName to error when /cmd not found

elgiano:
## Purpose and Motivation

When triggering a Plug-in defined command (with a `/cmd` osc message), and that command is not registered, the current error message is:

    FAILURE IN SERVER /cmd failed

Which could be clearer, if the requested command name and actual failure reason were specified. This PR makes it to:

    FAILURE IN SERVER /cmd /nn_load not found

Where /nn_load is the supposed plug-in command name, thus giving users a more precise hint of what has gone wrong. While I was there I have also applied the same change to Unit Commands.

Note that in order to print a dynamic error message, I had to resort to throwing an exception, which although didn't have precedents in any other server command, is already expected (and caught) by the function which runs these commands. An alternative could be to directly print to stdout.

## Types of changes

- New feature

## To-do list

### DIFF
--- a/server/scsynth/SC_UnitDef.cpp
+++ b/server/scsynth/SC_UnitDef.cpp
@@ -132,7 +132,7 @@ int Unit_DoCmd(World* inWorld, int inSize, char* inData) {
         return kSCErr_Failed;
     UnitCmd* cmd = unitDef->mCmds->Get(cmdName);
     if (!cmd)
-        return kSCErr_Failed;
+        throw std::runtime_error(std::string((char*)cmdName) + " not found");
 
     // only run unit command if the ctor has been called!
     if (graph->mNode.mCalcFunc == (NodeCalcFunc)&Graph_FirstCalc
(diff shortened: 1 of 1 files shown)


capital-G:
Seems like an useful addition. Do you know how supernova reacts to this and if this behavior can be matched in it as well?

elgiano:
supernova's behavior is matching already:
```cpp
std::cout << "unable to find cmd plugin: " << name << std::endl;
```

capital-G (approved):

server/scsynth/SC_UnitDef.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "SC_Endian.h" // first to avoid win32 IN clash
 
+#include "SC_Errors.h"
```
JordanHendersonMusic, line 25: What was this header needed for?

JordanHendersonMusic (commented):
Looks good! Just a little query.

server/scsynth/SC_UnitDef.cpp
elgiano: good catch, thanks a lot! it was just some LSP byproduct of previous work. I've removed it now

JordanHendersonMusic (approved):

PR #6511, opened 2024-10-21
CMake: fix qt6 detection

dyfer:
## Purpose and Motivation

This ~~should fix~~ fixes Qt6 detection on some systems that have both Qt6 and Qt5 installed.

Since we run `find_package(QT (...) )` in CMake, then the resulting variable should indeed be `QT_FOUND` and not `Qt6_FOUND` (thought the latter seemed to work on some systems), since in the call to the `find_package` function we set the package name to be `QT`. This behavior is described in the [cmake help](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature).

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- n/a Updated documentation

### DIFF
--- a/QtCollider/CMakeLists.txt
+++ b/QtCollider/CMakeLists.txt
@@ -16,7 +16,7 @@ set(
 
 # find Qt6 or Qt5
 find_package(QT NAMES Qt6 ${REQUIRED_QT_VERSION} QUIET COMPONENTS Core)
-if (NOT Qt6_FOUND)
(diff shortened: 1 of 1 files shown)


dyfer:
@capital-G I think this is a rather straightforward change, could we get it in?

capital-G (approved):
CI looks fine, so I think it is safe to merge

PR #6512, opened 2024-10-22
Qt: replace QRegExp with QRegularExpression

sonata-chen:
## Purpose and Motivation

This pull request is a follow-up for the Qt6 [migration](https://github.com/supercollider/supercollider/issues/5169),  replacing `QRegExp` with `QRegularExpression` in the scide codebase.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/core/sc_lexer.cpp
+++ b/editors/sc-ide/core/sc_lexer.cpp
@@ -28,51 +28,48 @@ QVector<ScLexer::LexicalRule> ScLexer::mLexicalRules;
 void ScLexer::initLexicalRules() {
     /* NOTE:
 
-    The lexical analysis algorithm demands that all regexps
-    start with a caret "^", to only match at beginning of string.
-
     Order is important:
     -- floatRegexp is subset of radixFloatRegex -> must come later
     -- classRegexp and primitiveRegexp are subsets of symbolArgRegexp -> must come later
 
     */
 
-    mLexicalRules << LexicalRule(Token::WhiteSpace, "^\\s+");
+    mLexicalRules << LexicalRule(Token::WhiteSpace, "\\s+");
 
     initKeywordsRules();
     initBuiltinsRules();
 
-    mLexicalRules << LexicalRule(Token::RadixFloat, "^\\b\\d+r[0-9a-zA-Z]*(\\.[0-9A-Z]*)?");
+    mLexicalRules << LexicalRule(Token::RadixFloat, "\\b\\d+r[0-9a-zA-Z]*(\\.[0-9A-Z]*)?");
 
     // Never heard of this one? Check the "Literals" help file :)
-    mLexicalRules << LexicalRule(Token::ScaleDegreeFloat, "^\\b\\d+(s+|b+|[sb]\\d+)\\b");
+    mLexicalRules << LexicalRule(Token::ScaleDegreeFloat, "\\b\\d+(s+|b+|[sb]\\d+)\\b");
 
     // do not include leading "-" in Float, as left-to-right algorithm does
     // not know whether it is not rather a binary operator
-    mLexicalRules << LexicalRule(Token::Float, "^\\b((\\d+(\\.\\d+)?([eE][-+]?\\d+)?(pi)?)|pi)\\b");
+    mLexicalRules << LexicalRule(Token::Float, "\\b((\\d+(\\.\\d+)?([eE][-+]?\\d+)?(pi)?)|pi)\\b");
 
-    mLexicalRules << LexicalRule(Token::HexInt, "^\\b0x(\\d|[a-f]|[A-F])+");
+    mLexicalRules << LexicalRule(Token::HexInt, "\\b0x(\\d|[a-f]|[A-F])+");
 
-    mLexicalRules << LexicalRule(Token::SymbolArg, "^\\b[A-Za-z_]\\w*\\:");
+    mLexicalRules << LexicalRule(Token::SymbolArg, "\\b[A-Za-z_]\\w*\\:");
 
-    mLexicalRules << LexicalRule(Token::Name, "^[a-z]\\w*");
+    mLexicalRules << LexicalRule(Token::Name, "[a-z]\\w*");
 
-    mLexicalRules << LexicalRule(Token::Class, "^\\b[A-Z]\\w*");
+    mLexicalRules << LexicalRule(Token::Class, "\\b[A-Z]\\w*");
 
-    mLexicalRules << LexicalRule(Token::Primitive, "^\\b_\\w+");
+    mLexicalRules << LexicalRule(Token::Primitive, "\\b_\\w+");
 
-    mLexicalRules << LexicalRule(Token::Symbol, "^\\\\\\w*");
+    mLexicalRules << LexicalRule(Token::Symbol, "\\\\\\w*");
 
-    mLexicalRules << LexicalRule(Token::Char, "^\\$\\\\?.");
+    mLexicalRules << LexicalRule(Token::Char, "\\$\\\\?.");
 
-    mLexicalRules << LexicalRule(Token::EnvVar, "^~\\w+");
+    mLexicalRules << LexicalRule(Token::EnvVar, "~\\w+");
 
 
-    mLexicalRules << LexicalRule(Token::SingleLineComment, "^//[^\r\n]*");
+    mLexicalRules << LexicalRule(Token::SingleLineComment, "//[^\r\n]*");
 
-    mLexicalRules << LexicalRule(Token::MultiLineCommentStart, "^/\\*");
+    mLexicalRules << LexicalRule(Token::MultiLineCommentStart, "/\\*");
 
-    mLexicalRules << LexicalRule(Token::Operator, "^[\\+-\\*/&\\|\\^%<>=]+");
+    mLexicalRules << LexicalRule(Token::Operator, "[+\\-*/&|\\^%<>=]+");
 }
 
 void ScLexer::initKeywordsRules() {
@@ -84,7 +81,7 @@ void ScLexer::initKeywordsRules() {
              << "this"
              << "var";
 
-    QString keywordPattern = QStringLiteral("^\\b(%1)\\b").arg(keywords.join("|"));
+    QString keywordPattern = QStringLiteral("\\b(%1)\\b").arg(keywords.join("|"));
     mLexicalRules << LexicalRule(Token::Keyword, keywordPattern);
 }
 
@@ -102,7 +99,7 @@ void ScLexer::initBuiltinsRules() {
              << "currentEnvironment"
              << "topEnvironment";
 
-    QString builtinsPattern = QStringLiteral("^\\b(%1)\\b").arg(builtins.join("|"));
+    QString builtinsPattern = QStringLiteral("\\b(%1)\\b").arg(builtins.join("|"));
     mLexicalRules << LexicalRule(Token::Builtin, builtinsPattern);
 }
 
@@ -153,12 +150,14 @@ Token::Type ScLexer::nextTokenInCode(int& lengthResult) {
 
         for (; it != end; ++it) {
             LexicalRule const& rule = *it;
-            int matchIndex = rule.expr.indexIn(mText, mOffset, QRegExp::CaretAtOffset);
-            // a guard to ensure all regexps match only at beginning of string:
-            Q_ASSERT(matchIndex <= mOffset);
-            if (matchIndex != -1) {
+            QRegularExpressionMatch match = rule.expr.match(mText, mOffset, QRegularExpression::NormalMatch,
+                                                            QRegularExpression::AnchoredMatchOption);
+            if (match.hasMatch()) {
+                // a guard to ensure all regexps match only at the beginning of the string:
+                Q_ASSERT(match.capturedStart() == mOffset);
+
                 type = rule.type;
-                length = rule.expr.matchedLength();
+                length = match.capturedLength();
                 break;
             }
         }
--- a/editors/sc-ide/widgets/code_editor/editor.cpp
+++ b/editors/sc-ide/widgets/code_editor/editor.cpp
@@ -177,22 +177,23 @@ void GenericCodeEditor::setShowWhitespace(bool show) {
 
 void GenericCodeEditor::setShowLinenumber(bool show) { mLineIndicator->setHideLineIndicator(!show); }
 
-static bool findInBlock(QTextDocument* doc, const QTextBlock& block, const QRegExp& expr, int offset,
-                        QTextDocument::FindFlags options, QTextCursor& cursor) {
+static bool findInBlock(QTextDocument* doc, const QTextBlock& block, const QRegularExpression& expr, int offset,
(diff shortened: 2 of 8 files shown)


dyfer:
Thanks for the PR!

I'd like to move this forward. Unfortunately I don't have deep expertise in the implementation details of this, nor am I fluent in regex. But I attempted to test this so that we can maybe collectively push it towards merging.

I downloaded the build and tested the Regex Searching in a text document in the SC IDE. I followed examples from http://www.regular-expressions.info/quickstart.html, copied the text from that webpage into a document in the IDE and ran a search with the Regex option checked. 

The following regex searches provide the same result in SC 3.13 and in this build:
- `a`
- ` gr[ae]y`
- `^b` (though it didn't match "b" in "bbb" in either build, is that correct?)
- `cat|dog`
- `<[A-Za-z][A-Za-z0-9]*>`
- `<.+>`
- `Set(Value)?`
-`\n` or `\r` don't seem to match newlines in either build, though I'm not sure if I'm using them correctly
- `([abc])=\1`
- `q(?=u)`
- `q(?!u)` (in both cases this matches not only "q" at the end of the word, but also "q" in "q(?=u)" and single "q", is that correct?)

Differences:
- `\p{L}` which, according to the linked website, should match all `L`etters (L being a Unicode category), works in this PR, matching all letters, but does not match anything in SC 3.13. I _think_ this PR is correct, but let me know if you think otherwise.
- `\P{L}` similarly as above works only in this PR, matching all non-letters. Again, is this correct?

I'd like to discuss/confirm if matching Unicode categories in this PR is correct, and whether any of the behaviors consistent between this build and SC 3.13 need to be reviewed/adjusted.

@capital-G @JordanHendersonMusic would you be able to look over the implementation details of this PR? 
Anybody else has experience with regex to test this?

sonata-chen:
@dyfer Thanks for the reviewing. I also tested some regex patterns you listed above using this build:

- `^b`, the first b in "bbb" is matched.( I put `bbb`  at the beginning of a line)
- For `q(?=u)` and `q(?!u)`, it seems correct for me, though i am not familiar with regex Lookaround.  `q(?=u)` matches any 'q' which is followed by a 'u'; `q(?!u)`matches any 'q' which is NOT followed by a 'u'.

Some migration notes are listed below:

1. Autocompleter and Syntax Highlighting

   The regex patterns in `sc_lexer.cpp` have been modified, primarily by removing the `^` character. Additionally, the `QRegularExpression::AnchoredMatchOption` option is set for regex matching.

2. Find and Replace

   - For plaintext searching, patterns are created by `QRegularExpression::escape` static function.
     See `editors/sc-ide/widgets/find_replace_tool.cpp` for details.
   - For pattern matching, `QRegExp::indexIn()` and `QRegExp::lastIndexIn()` are replaced with `QString::indexOf()` and `QString::lastIndexOf()`. See `editors/sc-ide/widgets/code_editor/editor.cpp:findInBlock()` for details.

dyfer:
> * `^b`, the first b in "bbb" is matched.( I put `bbb`  at the beginning of a line)

Thanks. Is it supposed to match only at the beginning of the line then?

capital-G:
I can take a look at this on tuesday, but the testing @dyfer validated seem to already match most cases one would use regex - I think 90% of the time regex is used for search is either `.*`, `[A-z]`, `^` or `$` (as we don't do any capturing in a search).

One thing that could be documented (where?) which kind of regex-dialect Qt is using under the hood - IIUC it is aimed to emulate the perl dialect, see https://doc.qt.io/qt-6/qregularexpression.html#introduction

> QRegularExpression implements Perl-compatible regular expressions.

For anything regex related-testing, be sure to check out https://regex101.com/ :)

editors/sc-ide/core/sc_lexer.cpp
```diff
@@ -153,12 +154,13 @@ Token::Type ScLexer::nextTokenInCode(int& lengthResult) {
...
         for (; it != end; ++it) {
             LexicalRule const& rule = *it;
-            int matchIndex = rule.expr.indexIn(mText, mOffset, QRegExp::CaretAtOffset);
-            // a guard to ensure all regexps match only at beginning of string:
-            Q_ASSERT(matchIndex <= mOffset);
-            if (matchIndex != -1) {
+            QRegularExpressionMatch match = rule.expr.match(mText, mOffset, QRegularExpression::NormalMatch,
+                                                            QRegularExpression::AnchoredMatchOption);
+            // a guard to ensure all regexps match only at the beginning of the string:
+            Q_ASSERT(match.hasMatch() && match.capturedStart() == mOffset);
```
JordanHendersonMusic, line 160: This assert fails if it doesn't match the first lexical rule. Should it go inside the following if statement?

editors/sc-ide/widgets/code_editor/editor.cpp
```diff
@@ -321,81 +328,93 @@ int GenericCodeEditor::findAll(const QRegExp& expr, QTextDocument::FindFlags opt
...
+    bool caps = (expr.patternOptions() & QRegularExpression::PatternOption::DontCaptureOption) == 0;
+    if (cursor.hasSelection()) {
+        QRegularExpressionMatch match = expr.match(cursor.selectedText());
+        if (match.hasMatch()) {
+            QString rstr = replacement;
+            if (caps) {
+                rstr = resolvedReplacement(rstr, match);
+            }
+            cursor.insertText(rstr);
+        }
```
JordanHendersonMusic, line 387: Does this block of code do more than what it is replacing? What does `caps` stand for? Is this doing something different with case sensitivity?

editors/sc-ide/widgets/code_editor/editor.cpp
```diff
@@ -321,81 +328,93 @@ int GenericCodeEditor::findAll(const QRegExp& expr, QTextDocument::FindFlags opt
...
     mSearchSelections.clear();
     updateExtraSelections();
 
-    if (expr.isEmpty())
+    if (expr.pattern().isEmpty())
         return 0;
 
     int replacements = 0;
-    bool caps = expr.patternSyntax() != QRegExp::FixedString;
+    bool caps = (expr.patternOptions() & QRegularExpression::PatternOption::DontCaptureOption) == 0;
```
JordanHendersonMusic, line 402: Same here, could this be made a little more explicit?

editors/sc-ide/widgets/code_editor/editor.cpp
JordanHendersonMusic: Edit: just realised it stands for 'captures'. Perhaps write that out in full?

editors/sc-ide/core/sc_lexer.cpp
```diff
@@ -28,68 +28,68 @@ QVector<ScLexer::LexicalRule> ScLexer::mLexicalRules;
...
+    // clang-format on
 
-    QString keywordPattern = QStringLiteral("^\\b(%1)\\b").arg(keywords.join("|"));
+    QString keywordPattern = QStringLiteral("\\b(%1)\\b").arg(keywords.join("|"));
     mLexicalRules << LexicalRule(Token::Keyword, keywordPattern);
 }
 
 void ScLexer::initBuiltinsRules() {
     QStringList builtins;
+    // clang-format off
```
JordanHendersonMusic, line 92: I'd remove these, just let the formatter do its thing.

JordanHendersonMusic (dismissed):
@sonata-chen thanks for this! 

Just had a quick look through the C++, I haven't tested it yet  — just a few minor things.

I haven't looked at the regex as I'm not super familiar with the details.

editors/sc-ide/core/sc_lexer.cpp
sonata-chen:
```
if (match.hasMatch()) {
    // a guard to ensure all regexps match only at the beginning of the string:
    Q_ASSERT(match.capturedStart() == mOffset);
    type = rule.type;
    length = match.capturedLength();
    break;
}
```

Good catch! Is this what you mean?

editors/sc-ide/widgets/code_editor/editor.cpp
sonata-chen: It will perform regex backreference if `caps` is `true`. If `rstr` contains `\1`, `\2`, `\3` etc., `resolveReplacement()` function will return a new string with these backreferences(`\1`, `\2`...) replaced by their corresponding captured groups.

editors/sc-ide/widgets/code_editor/editor.cpp
sonata-chen: Agree, the variable name also confuses me when i try to reason the original code.

editors/sc-ide/core/sc_lexer.cpp
JordanHendersonMusic: sure, that is what I thought you meant in the first place.

editors/sc-ide/widgets/code_editor/editor.cpp
JordanHendersonMusic: Perhaps call it something like `replace_back_references_with_captures` ?

dyfer:
@capital-G thanks for the feedback and confirming the behavior of `^b`.

> One thing that could be documented (where?) which kind of regex-dialect Qt is using under the hood - IIUC it is aimed to emulate the perl dialect, see https://doc.qt.io/qt-6/qregularexpression.html#introduction

Yeah, it's a bit buried in the docs.
The Perl dialect is mentioned in the [String -*Regexp help](https://doc.sccode.org/Classes/String.html#Regular%20expressions), but it's not explicitly indicated that the same applies in the IDE.
SuperCollider IDE has a one page entry in the help and [searching with regular expressions is briefly mentioned there](https://doc.sccode.org/Guides/SCIde.html#Find%20and%20Replace). That would be possibly another place do indicate the dialect (in a separate PR IMO).

> For anything regex related-testing, be sure to check out https://regex101.com/ :)

Ah, a great resource, thanks!

capital-G (changes requested):
The PR crashes on my machine (macOS 15.0 x86_64) when I click on the cursor after startup, the console prints

```shell
Path override failed for key base::DIR_APP_DICTIONARIES and path '/Users/scheiba/github/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/Frameworks/QtWebEngineCore.framework/Versions/A/Helpers/QtWebEngineProcess.app/Contents/MacOS/qtwebengine_dictionaries'
ASSERT: "match.hasMatch() && match.capturedStart() == mOffset" in file /Users/scheiba/github/supercollider/editors/sc-ide/core/sc_lexer.cpp, line 160
[1]    68522 abort      
```

editors/sc-ide/core/sc_lexer.cpp
JordanHendersonMusic: @capital-G this should fix that.

sonata-chen: Ah! I forgot to test it in debug build.

sonata-chen:
@capital-G
The issue has been fixed. Please try again!

editors/sc-ide/core/sc_lexer.cpp
dyfer: @sonata-chen it looks like clang-format wants to leave the following lines as they were originally: https://github.com/supercollider/supercollider/actions/runs/11573295631/job/32227021915#step:5:133

sonata-chen: It seems that different version of clang-formats do not have the same behaviour. I'm using a latest one. There is a [issue](https://github.com/llvm/llvm-project/issues/88483) opened on llvm repo.

dyfer: FWIW [we use clang-format 14](https://github.com/supercollider/supercollider/wiki/Cpp-formatting-instructions#clang-format). I don't think we can expect exactly the same behavior from different versions of clang-format - it took some effort to migrate from version 8 to 14...

sonata-chen: I have set up pre-commit successfully, thanks!

capital-G:
Thanks, now works on my machine.

> * `q(?=u)`
> 
> * `q(?!u)` (in both cases this matches not only "q" at the end of the word, but also "q" in "q(?=u)" and a single "q", is that correct?)

I don't understand the comment, but I found that both expressions, negative and positive lookahahead, capture as expected.

Some things that I found are not working as expected

* Positive/Negative lookbehind: `(?<=foo)bar` should match `foobar` but not `barbar`, yet it fails to match `foobar`.
* Turn a capture from greedy to lazy: while `<.+>` is greedy and therefore matches the whole group `<div>test</div>` (which works), turning it lazy via `<.+?>` should match `<div>` and `</div>`, yet it does not capture anything.

But both captures also didn't worked in 3.13/Qt5, and I also think if someone needs such advanced regex capabilities they should rely on the sclang implementation or some other tool and not a simple IDE search.

So from my side, the regex work fine!

---

I noticed poor performance of the IDE - e.g. I can't alt-tab into it and it is often frozen for some seconds when selecting the window, especially when closing the IDE which takes around 15 sec - but I think this is not tied to this PR as this is also happening on dev branch, so this is most likely a Qt6 issue - just want to write it down here as I noticed it here during testing.

dyfer:
> > * `q(?!u)` (in both cases this matches not only "q" at the end of the word, but also "q" in "q(?=u)" and a single "q", is that correct?)
> 
> I don't understand the comment, but I found that both expressions, negative and positive lookahahead, capture as expected.

I meant that it did capture "q" in the string "q(?=u)" which was part of the ide document; in any case, thank you for confirming that this works a expected.

> I noticed poor performance of the IDE 

Yes, I unfortunately also noticed that with the Qt6 build after the migration. I hope that this is an issue that can eventually be addressed and not just the overall performance of Qt6, haha. In any case, I can confirm that the performance is not related to this PR.

dyfer:
@capital-G @JordanHendersonMusic do you think this PR is good to go?

> Some things that I found are not working as expected
@capital-G is it worth it filing them as a new issue?

JordanHendersonMusic:
Looks good to me, but I haven't looked at the regex details.

capital-G:
> @capital-G is it worth it filing them as new issue?

I don't think so. These are advanced features in regex I think nobody is using/relying on such features in a basic IDE search. I think there are other more interesting features to work on in the IDE.

editors/sc-ide/core/sc_lexer.cpp
```diff
@@ -1,3 +1,4 @@
+
```
dyfer, line 1: @sonata-chen do you know whether this whitespace change added by clang-format? It's not that important, but we try to avoid accidental whitespace changes.

sonata-chen: I believe it's me accidentally added that empty line. I will forced-push again.

dyfer:
Between my tests, the review by @capital-G and code review by @JordanHendersonMusic I'd be comfortable pulling this is. Are you both okay with that?

JordanHendersonMusic:
Sounds good!

capital-G (approved):

Issue #6036, opened 2023-05-22
Reduce CI on forks


capital-G:
## Motivation

Everyone who has a recent fork of SuperCollider also clones its CI pipeline, which is convenient for development but also a waste of resources, as there are weekly builds that consume over 1 day of CI time (see https://github.com/capital-G/supercollider/actions/runs/5034978486). I also get an email every week with a failed Windows build.

## Description of Proposed Feature

Github CI allows to access the current repository via [`github.repository`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) and each job or step can have an additional if condition where this job/step is run if the if condition holds.

I'd recommend to skip all matrix builds within a fork and only run e.g. linting and a test pipeline and running the full build pipeline only when creating a PR against `supercollider/supercollider`.
I like the idea of making life *harder* for Microsoft, but even though Azure uses green energy according to their marketing material, we should not waste resources.

## Plan for Implementation

I could take care of this once it is agreed upon which CI steps to skip in forks and which not to skip.

dyfer:
Thanks @capital-G 

I have certainly thought about limiting CI in general, but when this was discussed in the dev meetings, the feedback I got was typically to keep it running so that we can catch problems early.

Also, the typical run is not 1 day, but rather 30-45 mintues. It seems that you haven't updated your `develop` branch and the outdated linux build holds up the suite for one day. Also, it's not Windows builds that typically fail, but one of the tests.

The weekly builds are there to refresh cache (particularly the one for homebrew universal binaries) so that builds on non-default branch can use that; without the automated build, the cache for the repository grows really quickly beyond 10GB.

Also, it's not feasible with the current GHA and our matrix structure to only build one linux build and run tests, unfortunately there's no conditional (AFAIK) to run selected jobs from the matrix. The only solution IIUC is to created matrix dynamically, and makes the whole workflow file even less readable.

----
In any case, I'm open to making the build matrix more configurable, but I think each repository should be able to make a decision when to run the builds (scheduled, every push, every push on protected branches, PRs only etc). I _think_ one way to implement this would be to use secrets in conditions for each scenario, which then can be set individually in different repos.

dyfer:
@capital-G I forgot to mention that you can also disable all CI on your fork, if you don't need it (e.g. if you only need it for PRs to the main repo). However, all my previous points still stand, particularly the one about updating the `develop` branch on your fork.

capital-G:
Lets do some maths :)
[...]

PR #6518, opened 2024-10-26
GHA: don't run weekly

dyfer:
## Purpose and Motivation

We started running GHA weekly after I added the custom script that was creating "universal" x86_64 and arm64 versions of homebrew binaries on macOS. It seems this was in late 2022.
When a cached version of these "universal" binaries was not available, the main macOS build would take 40-50 minutes to complete. I then added the scheduled running of the CI so that the cache was updated weekly and thus was (theoretically) always available to other CI jobs (cache from the default branch is available to jobs from other branches, but not vice versa).

After #6489 we don't make a universal macOS build anymore and  don't use the script. Now the uncached build takes 15-20 minutes, and cached builds take 4-10 minutes. I don't think it's as crucial to keep the cache updated, especially that sometimes cache is not available anyway (e.g. if we go over the storage limit and the oldest cache that's evicted happens to be from the default branch).

I propose that we stop triggering the builds weekly.

This does not change the availability of the develop builds on the SC website (hosted on the s3 server) - these are still created anytime there's a push to `develop`.

## Types of changes

- Maintenance

## To-do list

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -12,8 +12,6 @@ on:
       - 'HelpSource/**'
       - 'sounds/**'
       - '*.md'
-  schedule:
-    - cron: '0 0 * * 0' # run weekly to refresh cache
 
 jobs:
   lint:


capital-G (approved):
Good catch.
I think this also closes #6036

PR #6519, opened 2024-10-29
scide style.cpp: fix icon sizes on tab bar and help toolbar

sadguitarius:
## Purpose and Motivation

When building with the latest Qt6 changes, the icons in the tab bar and help toolbar are showing up with the wrong size in the tab bar case and not centered properly in the help toolbar case. This commit should hopefully fix the issue, although more testing may be needed for other platforms and display resolutions.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/style/style.cpp
+++ b/editors/sc-ide/widgets/style/style.cpp
@@ -140,9 +140,10 @@ void Style::drawComplexControl(ComplexControl control, const QStyleOptionComplex
             QIcon::State iconState = option->state & QStyle::State_Selected ? QIcon::On : QIcon::Off;
 
             QPixmap pixmap = icon.pixmap(toolOption->iconSize, iconMode, iconState);
-            QRect pixRect = pixmap.rect();
+            QRect pixRect;
+            pixRect.setSize(QSize(16, 16));
             pixRect.moveCenter(option->rect.center());
-            painter->drawPixmap(pixRect.topLeft(), pixmap);
+            painter->drawPixmap(pixRect, pixmap);
         } else {
             QStyle::PrimitiveElement elem = Style::PE_CustomBase;
             switch (toolBtn->arrowType()) {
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Thanks, I can confirm this fixes the issue on macOS

image

I am not too familiar with Qt - are hardcoded integers normal or should something like `em` (relative to font-size) from CSS should be used? But for me this is still good as-is.

On which platform did you test it? It would be great if someone could take a look how it looks under windows, I can also take a look how it looks under Linux (Fedora).

capital-G:
Ah, this was not ready to be merged yet :/ This required some additional feedback from other OSes :/

telephon:
Sorry!  Let's not approve things that need further work, ok?

capital-G:
I don't think this required more work but only an additional review from a person with a windows machine - as stated in the approval, therefore I approved it for macOS (and linux).

But in the end I think it is fine if `dev` is not *perfect* / bug free, that what its for - if someone on windows experiences troubles they now need to do a PR.

telephon:
Yes, that is what I thought too in this case. It is aminor fix.

capital-G:
I've added the `needs-testing` tags for future PRs, one for general and one for each major platform - so I'd suggest that in the future, as long as a PR still has such a tag, it shouldn't be merged while it still can be approved? The tags can be used as a hint for people to test it out and provide feedback.

telephon:
We have the tag `do not merge yet`, which should be added in these case as well.

sadguitarius:
I’ve only tested this on Windows, so we’re good! Sorry, I forgot to mention. I think that at least for Qt6, scaling is handled automatically, so hard-coding is the proper way to do it. Someone correct me if I’m wrong though!

dyfer:
re: approvals
I personally prefer only approving the PRs if I think they are ready to merge. If I want to submit a partial approval, I just leave a comment.
I also agree that there's no harm done either way - `develop` doesn't need to be perfect all the time.

Issue #1022, opened 2014-01-21
Create server with random port


miguel-negrao:
It's sometimes desirable to create a server with a random udp port that's not used by anyone else. That's ease to do in the terminal:

```
scsynth -u 0
```

```
miguel@miguel-MacBookPro:~$ lsof -i -n -P
COMMAND    PID   USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
...
scsynth   8624 miguel    9u  IPv4 720446      0t0  UDP *:40164
```

If I try the same in the language I get a primitive failed error, although the scsynth is still created with the random port:

```
scsynth   8724 miguel    9u  IPv4 763570      0t0  UDP *:50694 
scsynth   8724 miguel   12u  IPv4  24557      0t0  UDP *:57120 

caught exception in primitive NetAddr:sendMsg
ERROR: send_to: Argumento inválidoERROR: Primitive '_NetAddr_SendMsg' failed.
Failed.
RECEIVER:
Instance of NetAddr {    (0x101abfd8, gc=20, fmt=00, flg=00, set=02)
  instance variables [4]
    addr : Integer 2130706433
    port : Integer 0
    hostname : "localhost"
    socket : nil
}
PATH: /home/miguel/bin/fileplayer.scd

PROTECTED CALL STACK:
    Meta_MethodError:new    0x6ddd140
        arg this = PrimitiveFailedError
        arg what = Failed.
        arg receiver = a NetAddr(127.0.0.1, 0)
    Meta_PrimitiveFailedError:new   0x6de3700
        arg this = PrimitiveFailedError
        arg receiver = a NetAddr(127.0.0.1, 0)
    Object:primitiveFailed  0x1250280
        arg this = a NetAddr(127.0.0.1, 0)
    NetAddr:sendStatusMsg   0x7ca2dc0
        arg this = a NetAddr(127.0.0.1, 0)
    Server:status   0x85f42c0
        arg this = test
    a FunctionDef   0x85f1900
        sourceCode = "<an open Function>"
    Routine:prStart 0x779f380
        arg this = a Routine
        arg inval = 1390309410.2242

CALL STACK:
    MethodError:reportError   0xf2255f8
        arg this = <instance of PrimitiveFailedError>
    < closed FunctionDef >   0xf229cc8
        arg error = <instance of PrimitiveFailedError>
    Integer:forBy   0xd06ff68
        arg this = 0
        arg endval = 0
        arg stepval = 2
        arg function = <instance of Function>
[...]

PR #6528, opened 2024-11-07
scsynth: correct Rendezvous port byte order, send assigned if 0

frenchy64:
## Purpose and Motivation

For user convenience, [Overtone](https://github.com/overtone/overtone) boots `scsynth` on their behalf most of the time. It chooses a random UDP port, [which doesn't always work](https://github.com/overtone/overtone/issues/573) if it's already bound. It's a bit more reliable to [request a port from the OS](https://github.com/overtone/overtone/commit/0ad8dfea0cd4cf8c9dd0ce16af8d3d67a66fe7c7) to determine which port to use, then close it before passing it to `scsynth`. It would be more reliable for `scsynth` itself to request a port via Boost.

As noted in https://github.com/supercollider/supercollider/issues/1022, `scsynth -u 0 -t 0` is assigned free ports from the OS, but the port is not communicated back to the client. There is a long discussion about how to do this but seemingly no solution was implemented.

I started studying how port 0 is handled and noticed on macOS an unusual log is printed by `PublishPortToRendezvous`. The provided ports are byte-flipped before sending to Rendezvous. I confirmed with the [Discovery browser](https://www.tildesoft.com/).

```
$ scsynth -u 44445 -t 55556
...
PublishPortToRendezvous 0 40365
PublishPortToRendezvous 1 1241
```

Discovery shows udp=40365 and tcp=1241.

When providing zero, there's two issues: the _provided_ port is byte-flipped, not the _assigned_ one:

```
$ scsynth -u 0 -t 0
...
PublishPortToRendezvous 0 0
PublishPortToRendezvous 1 0
```

These entries don't seem to appear in Discovery at all.

With this fix, the port is correct in Discovery and the log for both cases.

```
$ ./scsynth -u 0 -t 0
PublishPortToRendezvous 0 49956
PublishPortToRendezvous 1 52023

$ ./scsynth -u 44445 -t 55556
PublishPortToRendezvous 0 44445
PublishPortToRendezvous 1 55556
```

Tested on macOS 14.7.1 Intel MBP 2020.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

## Merge notes

I considered updating the `scsynth` documentation to note that 0 will allocate a free port. This might be premature to advertise since there's no official cross-platform way to retrieve the port.


frenchy64:
Just occurred to me I never tried to connect to the printed port. It doesn't seem to work, so needs another pass.

frenchy64:
Seemed like an endianness issue, fixed.

frenchy64:
Put this back into draft because there might be a second bug here. The endianness of the port is swapped (50014 becomes 24259).

```
$ scsynth -u 50014
Found 0 LADSPA plugins
Number of Devices: 7
   0 : "BlackHole 2ch"
   1 : "External Headphones"
   2 : "MacBook Pro Microphone"
   3 : "MacBook Pro Speakers"
   4 : "Camo Micro"
   5 : "Aggregate Device"
   6 : "Multi-Output Device"

   Streams: 1
      0  channels 1

"External Headphones" Output Device
   Streams: 1
      0  channels 2

SC_AudioDriver: sample rate = 44100.000000, driver's block size = 512
PublishPortToRendezvous 0 24259
SuperCollider 3 server ready.
```

frenchy64:
Combined fixes for both issues.

server/scsynth/SC_ComPort.cpp
```diff
@@ -258,13 +258,15 @@ class SC_UdpInPort {
...
+        if (inPortNum == 0)
+            mPortNum = udpSocket.local_endpoint().port();
 
         boost::asio::socket_base::send_buffer_size option(65536);
         udpSocket.set_option(option);
 
 #ifdef USE_RENDEZVOUS
         if (world->mRendezvous) {
-            SC_Thread thread(boost::bind(PublishPortToRendezvous, kSCRendezvous_UDP, sc_htons(mPortNum)));
+            SC_Thread thread(boost::bind(PublishPortToRendezvous, kSCRendezvous_UDP, mPortNum));
```
capital-G, line 269: Are we sure the byte swap is not necessary here? I "blamed" back and this seems to be like this since 2002 with its introduction via https://github.com/supercollider/supercollider/commit/531758a60742d3aea9f817b6c2c76621c5ce0c17 so I would be surprised if this has been not working properly for 20 years?

capital-G (changes requested):
Hello and thanks for contribution!

Currently the [CI is failing](https://github.com/supercollider/supercollider/actions/runs/11736888646/job/32706405952?pr=6528) because of formatting errors. You can read https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit how to set up our formatting pipeline - you probably need to run `FULL_CHECK=1 pre-commit run --all-files` as `pre-commit` will normally only check the staged files and here we need to fix already commited files.

I think the behavior of auto-assigning a port could be documented in the `-h` print of `scsynth`, located at https://github.com/supercollider/supercollider/blob/917faf1a8d2d6dc7c7a753590307518085a7419a/server/scsynth/scsynth_main.cpp#L58

At least I didn't know that port `0` is a convention to let the OS pick a port.
We could ask Windows and Linux users to verify that port `0` works as intended.

Other than these formalities and the question regarding the byte order (could this be an ARM/x86 thing?) I think the change is good, but maybe someone with some who is familiar with rendezvous/bonjour protocol is happy to give this also an review.

Maybe we should rename this to bonjour, since it has been renamed 2005 - but this should be another PR.

server/scsynth/SC_ComPort.cpp
frenchy64: I don't know, I was just following my nose. I will dig into it. Perhaps the version of Bonjour is relevant, but I tried on Intel and M1 and the published port is incorrectly byte-swapped for both.

frenchy64: I think there's evidence that this worked at some point for at least someone in the past

https://github.com/supercollider/supercollider/blob/917faf1a8d2d6dc7c7a753590307518085a7419a/HelpSource/Tutorials/Mark_Polishook_tutorial/01_Startup.schelp#L38C27-L38C32

frenchy64:
The build fails but I'm not sure it's related. But I think this deserves a unit test that checks the reported Bonjour port. I will work on that.

frenchy64:
Added a unit test written in Clojure which should work cross-platform. Verified locally against this PR on macOS 14.7.1 Intel.

capital-G:
> Added a unit test written in Clojure which should work cross-platform. Verified locally against this PR on macOS 14.7.1 Intel.

Thanks, but I'd actually prefer it w/o the clojure test b/c introducing a new programming language (with a new paradigm to this codebase) and the whole JVM stack for one test seems overkill and introduce additional maintenance work in the long run.

I'd be fine to merge this w/o the integration test, but I'd let this sit for some days in order some additional people may also have a look at this.

frenchy64:
No problem, removed. If someone finds it valuable they might be able to transliterate it into sclang or bash.

capital-G (approved):
Thanks, looks fine to me - will merge :)

PR #6530, opened 2024-11-10
ClassLib: Make Pbind work with kwargs

JordanHendersonMusic:
## Purpose and Motivation

Simply append the kwargs (as keyword-value pairs) to the list of pairs.

Explain different in the help file, along with the nuance with unpacking multiple keys from PTuple.

Add test to check keywords are identical with key-value pairs.

```supercollider
Pbind(a: Pseq([1, 2, 3]), dur: 1).trace.play
```

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Pbind.schelp
+++ b/HelpSource/Classes/Pbind.schelp
@@ -13,13 +13,19 @@ The keys used in a Pbind are usually determined by link::Classes/Event::'s defau
 ClassMethods::
 
 method::new
-The arguments to Pbind are an alternating sequence of keys and patterns. A pattern can also be bound to an array of keys. In this case, the pattern must specify a sequence whose elements are arrays with at least as many elements as there are keys.
+The arguments to Pbind are an alternating sequence of keys and patterns, this can alternatively be specified using keyword arguments.
+A pattern can also be bound to an array of keys.
+In this case, the pattern must specify a sequence whose elements are arrays with at least as many elements as there are keys.
 
 Examples::
 
 code::
 (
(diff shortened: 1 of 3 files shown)


telephon (approved):
Yes, this was one of the convincing points about kwargs.

telephon (changes requested):
only merge when tested …

JordanHendersonMusic:
One question is whether I should change all examples in the help docs?? Opinions? The down side (as mentioned) is that it can't do unpacking* but this doesn't occur often.



*Technically, it might be possible, to make the following a language feature...
```
f.(#[a, b]: [1, 2]) == f.(a: 1, b: 2)
```

I kinda like this as it's similar to 
```
#a, b = g.();
```

telephon:
yes, I would first make the syntax complete before changing all examples. And I think the syntax should be implemented, because it avoids people have to think through special cases. If you have time, that would be excellent of course.

JordanHendersonMusic:
This one might take a moment because it means dealing with the parsenodes (which are quite messy). I've messed with them before with trying to get variable declarations in any order, I got quite close, but didn't have time to investigate.

JordanHendersonMusic:
So adding that additional syntax might actually be quite challenging so should be a separate PR. 

This PR just makes kwargs possible with `Pbind` and alters the documentation to mention this.

telephon (approved):

capital-G (approved):
Thanks Jordan, this is such a great addition! Looking forward to use it soon in performances :)

PR #6532, opened 2024-11-13
Build: divide each test run into 5 deterministic separate jobs

frenchy64:
## Purpose and Motivation

Related https://github.com/supercollider/supercollider/issues/6439

This PR speeds up CI runs of `test-macos` and `test-linux` by about 3-4x (12m => 4m).

The approach is to deterministically partition the tests into separate concurrent jobs. Test failures will be locally reproducible by passing the same arguments to the test script as whichever job fails.

Demonstration: https://github.com/frenchy64/supercollider/actions/runs/11807687690

## Types of changes

- CI Build performance improvement

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

## Merge notes

Reviewers please help me verify that the test partitioning logic is deterministic (per commit). Each test should run exactly once and in the same order.

`.clumps()` partitions the tests but it assumes the array it is clumping is itself deterministic. Please double check this.

Suggestions welcome on how to improve `testsuite/scripts/setup_test_clumps_output.sh`. Generating JSON from bash is ugly, but it at least seems lean and fast.

### DIFF
--- a/testsuite/scripts/sclang_test_runner.scd
+++ b/testsuite/scripts/sclang_test_runner.scd
@@ -1,15 +1,28 @@
-// this script is meant to be run with two arguments: path to test prototype indicating tests to be run, and path to test result
+// this script is meant to be run with four arguments:
+// 1. path to test prototype indicating tests to be run
+// 2. path to test result
+// 3. which partition of the tests should be run (zero based)
+// 4. the number of test partitions
 (
 ~exitWhenDone = false;
 if(thisProcess.argv.size > 0) {
 	"passed args: %".format(thisProcess.argv).postln;
 	~testProto = thisProcess.argv[0];
 	~testResult = thisProcess.argv[1];
+  if(thisProcess.argv.size > 2) {
+    ~thisClump = thisProcess.argv[2].asInteger();
+    ~nClumps = thisProcess.argv[3].asInteger();
+  } {
+    ~thisClump = 0;
+    ~nClumps = 1;
+  };
 	~exitWhenDone = true;
 } {
 	~testProto = PathName(thisProcess.nowExecutingPath).pathOnly +/+ "test_run_proto_gha.scd";
 	// ~testProto = PathName(thisProcess.nowExecutingPath).pathOnly +/+ "test_run_proto_all.scd"; // for testing
 	~testResult = PathName(thisProcess.nowExecutingPath).pathOnly +/+ "run" +/+ "test_result.scxtar";
+	~thisClump = 0;
+	~nClumps = 1;
 };
 
 if(File.exists(~testProto).not) {
@@ -18,6 +31,7 @@ if(File.exists(~testProto).not) {
 
 "~testProto path: %".format(~testProto).postln;
 "~testResult path: %".format(~testResult).postln;
+"clump % of %".format(~thisClump, ~nClumps).postln;
 
 ~verboseSetup = false; // true or false
 
@@ -62,6 +76,11 @@ if(File.exists(~testProto).not) {
 
 	testsDict;
 };
+
+~selectThisTestClump = {|tests,thisClump,nClumps|
+  tests.clump(tests.size()/nClumps).at(thisClump) ?? #[];
+};
+
 
 ~testrun = {|protoFile, resultFile|
 	var tests = ~getTestRecord.(protoFile);
@@ -143,6 +162,10 @@ if(File.exists(~testProto).not) {
 	// test_record["tests"] = tests;
 	// ~writeResult.(tests, file);
 	~verboseSetup.if({"After exclude: ".post; tests.do(_.postln)});
+
+  tests = ~selectThisTestClump.(tests, ~thisClump, ~nClumps);
+
+	~verboseSetup.if({"After clump: ".post; tests.do(_.postln)});
 
 	"\n\n\t*** Running the tests ***\n\n".post;
 
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -13,6 +13,10 @@ on:
       - 'sounds/**'
       - '*.md'
 
+env:
+  # number of jobs to split each test suite into
+  NCLUMPS: 5
+
 jobs:
   lint:
     runs-on: ubuntu-22.04
@@ -40,8 +44,15 @@ jobs:
     runs-on: ubuntu-22.04
     outputs:
       sc-version: ${{ steps.set-version.outputs.version }}
+      n-clumps: ${{ steps.set-test-clumps.outputs.n-clumps }}
+      test-clumps-matrix: ${{ steps.set-test-clumps.outputs.test-clumps-matrix }}
 
     steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: testsuite/scripts/setup_test_clumps_output.sh
+          sparse-checkout-cone-mode: false
+
       - name: set version string for artifacts
         id: set-version
         run: |
@@ -52,7 +63,9 @@ jobs:
             SC_VERSION=$GITHUB_SHA
           fi
           echo "version=$SC_VERSION" >> $GITHUB_OUTPUT
-
+      - name: set test clumps
+        id: set-test-clumps
+        run: ./testsuite/scripts/setup_test_clumps_output.sh
   build-linux:
     needs: [lint, version-getter]
     uses: ./.github/workflows/build_linux.yml
@@ -79,6 +92,8 @@ jobs:
     uses: ./.github/workflows/test_linux.yml
     with:
       artifact-file: "SuperCollider-${{ needs.version-getter.outputs.sc-version }}-ubuntu-22.04-gcc12"
+      test-clumps-matrix: ${{ needs.version-getter.outputs.test-clumps-matrix }}
+      n-clumps: ${{ needs.version-getter.outputs.n-clumps }}
     secrets: inherit
 
   test-macos:
@@ -86,6 +101,8 @@ jobs:
     uses: ./.github/workflows/test_macos.yml
     with:
       artifact-file: "SuperCollider-${{ needs.version-getter.outputs.sc-version }}-macOS-x64.dmg"
+      test-clumps-matrix: ${{ needs.version-getter.outputs.test-clumps-matrix }}
+      n-clumps: ${{ needs.version-getter.outputs.n-clumps }}
     secrets: inherit
 
   # test-windows:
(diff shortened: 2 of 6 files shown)


testsuite/scripts/sclang_test_runner.scd
```diff
@@ -144,6 +167,10 @@ if(File.exists(~testProto).not) {
 	// ~writeResult.(tests, file);
 	~verboseSetup.if({"After exclude: ".post; tests.do(_.postln)});
 
+  tests = ~selectThisTestClump.(tests, ~thisClump, ~nClumps);
```
capital-G, line 170: Is the indentation off here? Looks strange on GH.

capital-G (commented):
Hey, thanks for this PR, this is really great!

I opposed parallelizing the tests in the pasts, but running them in parallel on dedicated machines looks like a good solution to me.
I found everything well implemented and clearly documented and it really benefits the pipeline - so from my side this is good to merge except the one small comment.

But let's wait what @dyfer thinks about it.

dyfer:
Hi @frenchy64 and thanks for the PR!
I generally think splitting the tests to speed them up is great.

> `.clumps()` partitions the tests but it assumes the array it is clumping is itself deterministic. Please double check this.

Good point. I'm not 100% sure about this. I think technically there's no guarantee that the class tree is deterministic, as it uses an `IdentitySet`... maybe we should sort the test classes before clumping them?

Regarding the Bash script: I appreciate the automation, but maybe for transparency it would be easier to just use the build matrix for this, something like
```
matrix:
        this-clump: [0, 1, 2, 3, 4] # consecutive numbers starting from 0
        n-clumps: 5 # needs to be one larger than the highest `this-clump`
```
This of course is error-prone, but it's also very transparent and avoids passinge JSON and hidden automation.

I think we need to 
1. test/ensure the deterministic list
2. consider using Bash vs manual matrix

testsuite/scripts/sclang_test_runner.scd
dyfer: Yes, please fix the indentation @frenchy64

dyfer (changes requested):
see previous comment

frenchy64:
> EDIT: maybe the number of clumps could be calculated with ${{ strategy.job-total }} instead of specifying it manually, but I haven't tested that.

Seems to work great! Halved the lines of changes needed.

dyfer:
I like it!

Before merging this I'd like to consider some aesthetic aspects:

- should the test-clump parameter had a short comment saying that it divides the test suite into a number of segments equal to the number of items in the array, and also that it doesn't matter when what these numbers are?
- I wonder if in the test running and/or test result steps we should add posting like "test suite segment X out of X"?



I wonder if it's worth it to add some

frenchy64:
Yes, good ideas. I will add that. I also need to investigate the sorting/determinism before this is merged.

capital-G:
One thing that should be verified before merging - do indeed all tests run as expected in CI?

I recently thought about writing a JUnit XML wrapper/generator for our test suite, it seems like such a thing would come handy now when we split the run across multiple machines.

testsuite/scripts/sclang_test_runner.scd
```diff
@@ -1,15 +1,32 @@
...
+		~nClumps = 1;
+	};
 	~exitWhenDone = true;
 } {
 	~testProto = PathName(thisProcess.nowExecutingPath).pathOnly +/+ "test_run_proto_gha.scd";
 	// ~testProto = PathName(thisProcess.nowExecutingPath).pathOnly +/+ "test_run_proto_all.scd"; // for testing
 	~testResult = PathName(thisProcess.nowExecutingPath).pathOnly +/+ "run" +/+ "test_result.scxtar";
+	~thisClump = 0;
+	~nClumps = 1;
 };
```
dyfer, line 30: Quick suggestion: instead of two definitions of 
```supercollider
~thisClump = 0;
~nClumps = 1;
```
you could add below this block:
```supercollider
~thisClump ?? {~thisClump = 0};
~nClumps ?? {~nClumps = 1};
```
or
```supercollider
~thisClump = ~thisClump ? 0;
~nClumps = ~nClumps ? 1;
```

dyfer:
> One thing that should be verified before merging - do indeed all tests run as expected in CI?

This is a good point.

> I recently thought about writing a JUnit XML wrapper/generator for our test suite, it seems like such a thing would come handy now when we split the run across multiple machines.

However, I'm weary of using a wrapper for a testsuite, as it provides hurdles when running tests locally (point in case: qpm). 

I think we could consider adding the script which generates the list of all tests after the tests run and compares the combined results between the jobs (exchanging that information would be cumbersome, but definitely doable).

capital-G:
> However, I'm weary of using a wrapper for a testsuite, as it provides hurdles when running tests locally (point in case: qpm).

My idea is to output the results of the test suit (also) as JUnit XML format so it can be parsed by tools such as https://github.com/marketplace/actions/publish-test-results  - it wouldn't affect the running of the tests, it is simply a standard to serialize unit test results. But lets keep this in its separate Issue/PR (but if you have already objections to this dyfer plz write me)

frenchy64:
> I think we could consider adding a script which generates the list of all tests after the tests run and compares it with the results combined from all the jobs (exchanging that information would be cumbersome, but definitely doable).

I've done this successfully several times, I can outline how to achieve this using GitHub Actions.

Here's an example of a job that downloads prior test results, then fans-out tests and collects results after fanning-in, then prints and uploads the test results to load-balance future runs.

1. [Download previous timings via a cache and upload an artifact to share with splits](https://github.com/threatgrid/ctia/blob/3478e16c2f19ccdbd9d2cd2e3957ec67cd15b867/.github/workflows/pr.yml#L82-L114)
2. [Download the timings artifact on each split](https://github.com/threatgrid/ctia/blob/3478e16c2f19ccdbd9d2cd2e3957ec67cd15b867/.github/workflows/pr.yml#L240-L244)
3. [Upload results from each split](https://github.com/threatgrid/ctia/blob/3478e16c2f19ccdbd9d2cd2e3957ec67cd15b867/.github/workflows/pr.yml#L302-L307)
4. [Compile all results and upload to cache for future builds](https://github.com/threatgrid/ctia/blob/3478e16c2f19ccdbd9d2cd2e3957ec67cd15b867/.github/workflows/pr.yml#L340-L365)

The logic on how to determine which tests to run is [here](https://github.com/threatgrid/ctia/blob/3478e16c2f19ccdbd9d2cd2e3957ec67cd15b867/dev/ctia/dev/split_tests.clj#L102-L157).

smoge:
Thanks for the excellent contribution.

Your PR introduces deterministic partitioning to predictably distribute tests across parallel CI jobs, which helps with the reproducibility of test runs locally. 

However, I noticed there might be old confusion in this community regarding terminology—this kind of determinism (predictable test distribution) is different from property-based testing determinism (where test case generation follows well-defined procedures that produce consistent results). Property-based tests are inherently deterministic by design.

 Making this distinction clear could help avoid confusion. These are two separate aspects of test reproducibility, and there is still a lot of confusion in the community regarding the word "determinism" in the context of tests.
 
 Thanks again

frenchy64:
I'm going to continue work [here](https://github.com/frenchy64/supercollider/pull/7) where I have permission to cancel broken builds and I'll push my work here once it works.

smoge:
> I'm going to continue work [here](https://github.com/frenchy64/supercollider/pull/7)

Maybe this can be developed elsewhere, but to mention something pertinent to supercollider testing enhancement perspectives.

SuperCollider's scheduler has to handle concurrent operations (even though it's single-core, it still needs to handle multiple "concurrent" tasks). That is, sclang is a single-core, non-deterministic scheduler. It needs to handle timing and order of operations, race conditions, and try the best possible correct sequencing.

There is a compelling development of using a deterministic parallel scheduler to allow you to consistently reproduce timing-related bugs using the same seed. Also, that's the only safe way to shrink: when finding bugs, you could reliably shrink test cases to minimal examples since the scheduling would remain consistent. Also, one way to fix tricky bugs regarding edge-case with race conditions (an increasingly common bug when using large GUI systems or doing something non-trivial)

https://stevana.github.io/parallel_property-based_testing_with_a_deterministic_thread_scheduler.html

capital-G:
Thanks for your input regarding property based testing @smoge, but I think this PR has nothing to do with property-based testing as it only changes the way we separate and distribute tests on our test runners in CI.
The determinism discussed here is about the partition of said tests across the runners and not about any determinism within tests of the language or server.

If you want to discuss the introduction of property testing please do so in a separate issue/PR/RFC to keep things focused, atomic and isolated.

smoge:
> I think this PR has nothing to do with property-based testing as this PR only changes the way we separate and distribute tests on our test runners in CI.

I know that.No problems, @capital-G. If this issue needs some more attention elsewhere, I'm available to contribute. Take care.

frenchy64:
I found a bug in my partitioning algorithm where some tests were being dropped (as speculated by @capital-G). The problem was I was making the clump size too small when certain non-divisible numbers of tests were provided, so an extra clump was being made for the leftovers that no job was picking up.

I'm hardening the code and adding assertions to double and triple check the partitions are correct.

smoge:
> I found a bug in my partitioning algorithm where some tests were being dropped (as speculated by @capital-G). The problem was I was making the clump size too small when certain non-divisible numbers of tests were provided, so an extra clump was being made for the leftovers that no job was picking up.
> 
> I'm hardening the code and adding assertions to double and triple check the partitions are correct.

I quickly looked at ~selectThisTestClump and drafted some code using ceiling division just to check it.

```
~selectThisTestClump = {|tests,thisClump,nClumps|
    tests.sort.clump(tests.size()/nClumps).at(thisClump) ?? #[];
};
```

```supercollder
~selectThisTestClump = {|tests, thisClump, nClumps|
    var sorted, clumpSize, clumps, result;
    
    sorted = tests.sort { |a, b| 
        if(a[\suite] == b[\suite]) {
            a[\test] < b[\test]
        } {
            a[\suite] < b[\suite]
        }
    };
    
    // ceiling division / no dropping of tests
    clumpSize = (sorted.size / nClumps).ceil.asInteger;
    
    clumps = sorted.clumps(clumpSize);
    
    // double-check partitioning
    protect {
        var totalTests = sorted.size;
        var testInClumps = clumps.flatten.size;
        
        "Test partitioning:".postln;
        "  Total tests: %".format(totalTests).postln;
        "  Number of clumps: %".format(clumps.size).postln;
        "  This clump index: %".format(thisClump).postln;
        "  Tests in this clump: %".format(clumps[thisClump].size).postln;
        
        if(testInClumps != totalTests) {
            "Test partitioning error: % tests in clumps, expected %".format(testInClumps, totalTests).error;
        };
        
        if(clumps.size > nClumps) {
            "Created % clumps, expected <= %".format(clumps.size, nClumps).error;
        };
    };
    
    // Returns clump, or empty array 
    clumps[thisClump] ?? #[]
};
```

PR #6534, opened 2024-11-15
Various help documentation updates

prko:
This PR has been updated! It contains various changes to the help documentation:
- Current status:
[https://github.com/supercollider/supercollider/pull/6534#issuecomment-2612920068](https://github.com/supercollider/supercollider/pull/6534#issuecomment-2612920068)
- Previous status:
[https://github.com/supercollider/supercollider/pull/6534#issuecomment-2600598633](https://github.com/supercollider/supercollider/pull/6534#issuecomment-2600598633)

The following is the Initial post of this thread:
 
----

## Purpose and Motivation
While discussing other things, the term `Answer` was pointed out:
https://github.com/supercollider/supercollider/issues/6533#issuecomment-2477642569
I have tried to replace the word `Answer|s` with `Return|s` throughout the documentation.
- The first word `Answer(|s)` after the method name is changed to `Return`.
- The first word `Answer(|s)` in the following sentences is changed to `Returns`.

Please proofread my change!

## Types of changes

- Documentation
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

Issue #6533, opened 2024-11-14
adding `.average`?


prko:
## Motivation
`.sum` and `.product` exist, but `.average` does not. I think `.average` would also be useful.

## Description of Proposed Feature
It should work similar to `.sum` or `.product`

## Plan for Implementation
Implementing only for `collection` seems OK to me, but implementing for `function` and `matrix` should be discussed if it is necessary.

capital-G:
Hey prko - I allowed myself to edit your post so the backticks are matching.

Are you aware of [`.mean`](https://docs.supercollider.online/Classes/Collection.html#-mean)? If yes, what would be the difference between `.average` and `.mean`? Mean has a somehow precise mathematical definition (https://en.wikipedia.org/wiki/Arithmetic_mean) but average is a bit more vague, see https://en.wikipedia.org/wiki/Average#Summary_of_types

prko:
Sorry for the typo.

Yes, `.mean` is what I want. Although I know the term, when I searched for it yesterday, I only found `.average` - my fault. But even if `.mean` had come to mind and I had found it, I would still have started a thread because there is no description of `.mean` in the help. The `function` argument without explanation seems to be an obstacle to selecting `.mean` to use as follows: 
```
(1..10).mean
```
`.sum` also have `function` argument. Hm.

How about modifying / adding the following to the help documentation?
```
method::sum
Answer the sum of all items or the results of the optional strong::function:: argument evaluated for each item in the receiver. The function is passed two arguments, the item and an integer index.
code::
List[1, 2, 3, 4].sum;
(0..8).sum { |i| 1 / (2 ** i) }; // only item is used as index is not used in the function
::

method::mean
Answer the arithmetic mean of the results of the sum with an optional function argument. See link::#-sum:: for optional function argument. 
code::
[1, 2, 3, 4].mean;
::
```

capital-G:
I always appreciate improvements and additions in the documentation! I think this could be a nice PR.

I enjoy your choose of words regarding answering, I think the more established term seems to be "returning" in the domain of programming.
And maybe we can make use of the new math syntax within schelp? :)

prko:
OK, I will do that, but 'answering' is not my choice. 
I made firstly a PR changing 'Answer' and 'Answer' to 'Return' or 'Returns'.
If that PR is approved, I will modify some method explanations in the Collection documentation with KaTex!

prko:
I am closing this as it has been split into #6534 and #6537.

PR #6537, opened 2024-11-16
Add/Modify Explanation for .sum, .product and .mean in Collection Documentation with KaTex

prko:
- Change explanation for sum with KaTex
- Add explanation for mean with KaTex

## Purpose and Motivation
While finding average method for Array, I feel the documentation is missing and not so friendly to find the method with use case:
https://github.com/supercollider/supercollider/issues/6533#issuecomment-2477595224

## Types of changes

- Documentation
   Screenshot 2024-11-16 at 12 45 00

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [] This PR is ready for review 

## Merge notes
#6534 should be approved.

### DIFF
--- a/HelpSource/Classes/Collection.schelp
+++ b/HelpSource/Classes/Collection.schelp
@@ -374,7 +374,7 @@ code::
 
 The above code is represented in mathematical notation as:
 math::
-\frac{\sum_{i=0}^{n} \frac{1}{2^{x_i}}}{n-x_0+1} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\} \quad (n=8)
+\frac{\sum_{i=0}^{n} \frac{1}{2^{x_i}}}{n+1} ~ \mid ~ x = \lbrace 0, \dots, n \rbrace, n=8
 ::
 where math::x_i:: is the function argument teletype::item::.
 
--- a/HelpSource/Classes/Collection.schelp
+++ b/HelpSource/Classes/Collection.schelp
@@ -307,13 +307,13 @@ method::sum
 Answer the sum of the items in the collection, optionally applying a link::Classes/Function:: to each item before summing. When provided, the link::Classes/Function:: is passed two arguments: the item and an integer index.
 
 code::
-(0..8).sum { |item| 1 / (2 ** item) }
 // only the item is used; index is unused.
+(0..8).sum { |item| 1 / (2 ** item) }
 ::
 
 The above code is represented in mathematical notation as:
 math::
-\sum_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\} \quad (n=8)
+\sum_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\}, n=8
 ::
 where math::x_i:: is the function argument teletype::item::.
 
@@ -326,7 +326,7 @@ table::
 ::
 equivalent to:
 math::
-\text{Sum} \{ x_0, x_1, \ldots, x_n \} := \prod_{i=0}^{n} x_i
+\text{Sum} \{ x_0, x_1, \ldots, x_n \} := \sum_{i=0}^{n} x_i
 ::
 ::
 
@@ -345,7 +345,7 @@ code::
 
 The above code is represented in mathematical notation as:
 math::
-\prod_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\} \quad (n=8)
+\prod_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\}, n=8
 ::
 where math::x_i:: is the function argument teletype::item::.
 
@@ -368,8 +368,8 @@ code::
 method::mean
 Answer the arithmetic mean of the items in the collection, optionally applying a link::Classes/Function:: to each item before averaging. When provided, the link::Classes/Function:: receives two arguments: the item and its integer index.
 code::
-(0..8).mean { |item| 1 / (2 ** item) }
 // only the item is used; index is unused.
+(0..8).mean { |item| 1 / (2 ** item) }
 ::
 
 The above code is represented in mathematical notation as:
--- a/HelpSource/Classes/Collection.schelp
+++ b/HelpSource/Classes/Collection.schelp
@@ -357,7 +357,7 @@ table::
 ::
 equivalent to:
 math::
-\text{Product} \{ x_0, x_1, \ldots, x_n \} = \prod_{i=0}^{n} x_i
+\text{Product} \{ x_0, x_1, \ldots, x_n \} := \prod_{i=0}^{n} x_i
 ::
 ::
 Example:
--- a/HelpSource/Classes/Collection.schelp
+++ b/HelpSource/Classes/Collection.schelp
@@ -304,63 +304,83 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
 ::
 
 method::sum
-Answer the sum of the items in the collection, optionally applying a link::Classes/Function:: to each item before summing. When provided, the link::Classes/Function:: is passed two arguments: the item and an integer index.
-
+Answer the sum of the items in the collection.
 code::
-// only the item is used; index is unused.
-(0..8).sum { |item| 1 / (2 ** item) }
+[3, 6, 12, 24].sum // -> 45
 ::
 
-The above code is represented in mathematical notation as:
+This can be represented as:
 math::
-\sum_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\}, n=8
+\sum_{i=0}^{n} x_i \quad | \quad \{x_0, x_1, \ldots, x_n\}
+::
+
+Optionally, a link::Classes/Function:: can be provided to be applied to each item in the collection before summing.
+
+argument:: function
+If provided, the link::Classes/Function:: is passed two arguments: the item and the iteration index.
+
+code::
+[3, 6, 12, 24].sum { |item, i| i / (2 ** item) } // -> 0.016113460063934
+::
+
+The above code can be represented as:
+math::
+\sum_{i=0}^{n} f(x_i,i) = \sum_{i=0}^{n} \frac{i}{2^{x_i}} \quad | \quad \{x_0, x_1, \ldots, x_n\}
 ::
 where math::x_i:: is the function argument teletype::item::.
 
-discussion::
-The sum of all items in a collection can be expressed as:
-code::
-List[1, 2, 3, 4].sum
-::
 
 method::product
-Answer the product of the items in the collection, optionally applying a link::Classes/Function:: to each item before multiplication. When provided, the link::Classes/Function:: receives two arguments: the item and its integer index.
-
+Answer the product of the items in the collection.
 code::
-// only the item is used; index is unused.
-(0..8).product { |item| 1 / (2 ** item) }
+[3, 6, 12, 24].product // -> 5184
 ::
 
-The above code is represented in mathematical notation as:
+This can be represented as:
 math::
-\prod_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\}, n=8
+\prod_{i=0}^{n} x_i \quad | \quad \{x_0, x_1, \ldots, x_n\}
+::
+
+Optionally, a link::Classes/Function:: can be provided to be applied to each item in the collection before multiplication.
+
+argument:: function
+If provided, the link::Classes/Function:: is passed two arguments: the item and the iteration index.
+
+code::
+[3, 6, 12, 24].product { |item, i| item.sqrt ** i } // -> 3456.0
+::
+
+The above code can be represented as:
+math::
+\prod_{i=0}^{n} f(x_i,i) = \prod_{i=0}^{n} \left( \sqrt{x_i} \right) ^ i \quad | \quad \{x_0, x_1, \ldots, x_n\}
 ::
 where math::x_i:: is the function argument teletype::item::.
-
-discussion::
-The product of all items in a collection can be expressed as:
-code::
-[1, 2, 3, 4].product
-::
 
 method::mean
-Answer the arithmetic mean of the items in the collection, optionally applying a link::Classes/Function:: to each item before averaging. When provided, the link::Classes/Function:: receives two arguments: the item and its integer index.
+Answer the arithmetic mean of the items in the collection.
 code::
-// only the item is used; index is unused.
-(0..8).mean { |item| 1 / (2 ** item) }
+[3, 6, 12, 24].mean // -> 11.25
 ::
 
-The above code is represented in mathematical notation as:
+This can be represented as:
 math::
-\frac{\sum_{i=0}^{n} \frac{1}{2^{x_i}}}{n+1} ~ \mid ~ x = \lbrace 0, \dots, n \rbrace, n=8
+\frac{1}{n+1} \sum_{i=0}^{n} x_i \quad | \quad \{x_0, x_1, \ldots, x_n\}
+::
+
+Optionally, a link::Classes/Function:: can be provided to be applied to each item in the collection before multiplication.
+
+argument:: function
+If provided, the link::Classes/Function:: is passed two arguments: the item and the iteration index.
+
+code::
+[3, 6, 12, 24].mean { |item, i| i / (item ** 2) } // -> 0.01171875
+::
+
+The above code can be represented as:
+math::
+\frac{1}{n+1} \sum_{i=0}^{n} f(x_i,i) = \frac{1}{n+1} \sum_{i=0}^{n} \frac{i}{{x_i}^2} \quad | \quad \{x_0, x_1, \ldots, x_n\}
 ::
 where math::x_i:: is the function argument teletype::item::.
-
-discussion::
-The mean of all items in a collection can be expressed as:
-code::
-[1, 2, 3, 4].mean
-::
 
 method::maxItem
 Answer the maximum of the results of strong::function:: evaluated for each item in the receiver. The function is passed two arguments, the item and an integer index.
--- a/HelpSource/Classes/Collection.schelp
+++ b/HelpSource/Classes/Collection.schelp
@@ -311,11 +311,11 @@ code::
 // only the item is used; index is unused.
 ::
 
-The code above is mathematically equivalent to:
+The above code is represented in mathematical notation as:
 math::
 \sum_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\} \quad (n=8)
 ::
-(emphasis::In the mathematic formula, the argument name:: teletype::item:: emphasis::in sclang code is represented as:: math::x_i::.)
+where math::x_i:: is the function argument teletype::item::.
 
 discussion::
 The sum of all items in a collection can be expressed mathematically as:
@@ -343,11 +343,11 @@ code::
 // only the item is used; index is unused.
 ::
 
-The code above is mathematically equivalent to:
+The above code is represented in mathematical notation as:
 math::
 \prod_{i=0}^{n} \frac{1}{2^{x_i}} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\} \quad (n=8)
 ::
-(emphasis::In the mathematic formula, the argument name:: teletype::item:: emphasis::in sclang code is represented as:: math::x_i::.)
+where math::x_i:: is the function argument teletype::item::.
 
 discussion::
 The product of all items in a collection can be expressed mathematically as:
@@ -372,11 +372,11 @@ code::
 // only the item is used; index is unused.
 ::
 
-The code above is mathematically equivalent to:
+The above code is represented in mathematical notation as:
 math::
 \frac{\sum_{i=0}^{n} \frac{1}{2^{x_i}}}{n-x_0+1} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\} \quad (n=8)
 ::
-(emphasis::In the mathematic formula, the argument name:: teletype::item:: emphasis::in sclang code is represented as:: math::x_i::.)
+where math::x_i:: is the function argument teletype::item::.
 
 discussion::
 The mean of all items in a collection can be expressed mathematically as:
--- a/HelpSource/Classes/Collection.schelp
+++ b/HelpSource/Classes/Collection.schelp
@@ -304,10 +304,33 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
 ::
 
 method::sum
-Answer the sum of the results of strong::function:: evaluated for each item in the receiver. The function is passed two arguments, the item and an integer index.
+Answer the sum of all items or the results of the optional strong::function:: argument evaluated for each item in the receiver. The function is passed two arguments, the item and an integer index.
 code::
 List[1, 2, 3, 4].sum;
+
 (0..8).sum { |i| 1 / (2 ** i) };
+// Only item is used as index is not used in the function.
+::
+discussion::
+If there is no strong::function:: argument, it can be expressed mathematically as follows:
+code::
+[\x_0, \x_1, ..., \x_n].sum // This is conceptual. Do not evaluate.
+// _0, _1 and _n indicate the index of the element.
+::
+math::
+\text{Sum} \{ x_0, x_1, \ldots, x_n \} = \sum_{i=0}^{n} x_i
+::
+
+method::mean
+Answer the arithmetic mean of the results of the sum with an optional function argument. See link::#-sum:: for optional function argument.
+discussion::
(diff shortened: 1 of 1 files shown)


prko:
I used AI chatbots to check the sentences in this PR and noticed that AI chatbots tend to prefer the term "elements" over "items" when referring to entities within arrays, collections, and similar classes. However, I retained the term "item" as it appeared in the previous documentation, disregarding the AI chatbots' suggestion.

This led me to compare the terminology used for these entities across different programming languages, including C++, C, Python, Smalltalk, Lisp, SuperCollider, Csound, Max, and Pure Data (Pd). According to some AI chatbots, it seems that most programming languages generally use the term "elements," while SuperCollider and other audio-related languages favour the use of "items."

If I recall correctly, SuperCollider-related books and help documentation often appear to use these terms interchangeably. I mention this solely to ensure clear documentation for newcomers to computer-aided sound generation and composition with SuperCollider. (I also have a vague recollection of researching the difference between "item" and "element" when I first noticed this distinction.) For experienced users, this difference in terminology will not cause any confusion, though I believe in the importance of using unambiguous terms whenever possible.

smoge:
It looks nice. I wonder why the original documentation prefers "answer" and not "return." Is there a reason for that?

Does something like this sound good when using "return"?

> Returns the sum of all items in the receiver or, optionally, the sum of values returned by evaluating a function for each item. When a function is provided, it receives two arguments: the item and its index.

telephon:
I would use "element" when speaking of a set in mathematics, and "item" when speaking of a collection in sclang.

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,53 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
-Answer the sum of the results of strong::function:: evaluated for each item in the receiver. The function is passed two arguments, the item and an integer index.
+Answer the sum of all items or the results of the optional strong::function:: argument evaluated for each item in the receiver. The function is passed two arguments, the item and an integer index.
 code::
 List[1, 2, 3, 4].sum;
+
 (0..8).sum { |i| 1 / (2 ** i) };
+// Only item is used as index is not used in the function.
+::
+discussion::
+If there is no strong::function:: argument, it can be expressed mathematically as follows:
```
telephon, line 315: The standard case in mathematics is that there *is* a function argument.
So you can write yours as \sum_{i = 0}^{8} \frac{1}{2^i}

You can add sclang examples with functions, for mean and product as well.

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,53 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+code::
+[\x_0, \x_1, ..., \x_n].sum // This is conceptual. Do not evaluate.
```
telephon, line 317: instead of "// This is conceptual. Do not evaluate."
// pseudo-code

will be enough.
Commanding the reader is not necessary here, better just inform.

telephon (changes requested):

telephon:
It is nice to have a clear documentation of the math, so people can see how easy it is to express it in sclang.

HelpSource/Classes/Collection.schelp
mtmccrea: And as it's pseudo code, the items probably don't need to be Symbols, but rather just
[x_0, x_1, ..., x_n]
And the comment about the underscored numbers can be removed. I think it's clear what they represent from the latex equation that follows.

prko:
SC-IDE window appeared between 2 minutes 7 seconds and 2 minutes 10 seconds after double click (sclang boot time not measured) before finding the solution mentioned.

prko:
The rendered result of https://github.com/supercollider/supercollider/pull/6537/commits/86e51bdf3bd636b6283b5e73b1355fbe886cf2dd:

Screenshot 2025-02-21 at 01 18 15

HelpSource/Classes/Collection.schelp
prko: Done, thanks! See https://github.com/supercollider/supercollider/pull/6537#issuecomment-2671991562

HelpSource/Classes/Collection.schelp
prko: Done, thanks! See https://github.com/supercollider/supercollider/pull/6537#issuecomment-2671991562

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,53 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+// _0, _1 and _n indicate the index of the element.
+::
+math::
+\text{Sum} \{ x_0, x_1, \ldots, x_n \} = \sum_{i=0}^{n} x_i
+::
+
+method::product
+Answer the product of the items in the collection, optionally applying a function to each item before multiplication. See link::#-sum:: for optional strong::function:: argument.
+discussion::
+If there is no strong::function:: argument, it can be expressed mathematically as follows:
```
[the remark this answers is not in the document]
prko: Done, thanks! See https://github.com/supercollider/supercollider/pull/6537#issuecomment-2671991562

prko:
#6599 is also included in this PR.

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,53 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+code::
+[\x_0, \x_1, ..., \x_n].sum // This is conceptual. Do not evaluate.
+// _0, _1 and _n indicate the index of the element.
+::
+math::
+\text{Sum} \{ x_0, x_1, \ldots, x_n \} = \sum_{i=0}^{n} x_i
+::
+
+method::product
+Answer the product of the items in the collection, optionally applying a function to each item before multiplication. See link::#-sum:: for optional strong::function:: argument.
```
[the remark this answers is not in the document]
prko: I think I did what you think is better! Please look at my change!

mtmccrea:
> The rendered result of [86e51bd](https://github.com/supercollider/supercollider/commit/86e51bdf3bd636b6283b5e73b1355fbe886cf2dd):
> 
> Screenshot 2025-02-21 at 01 18 15

Thanks for the screenshot, that's helpful. A few observations:

- IMO it's a bit overkill to use the math notation in both the method description and the discussion. I think one or the other is fine (perhaps just the discussion?).

- The item and index is still being conflated in the math notation. For example, your example for `sum` could instead be:

$$
\sum_{i=0}^n \frac{1}{2^{x_i}} \quad \text{for} \quad \\{ x_0, x_1, ... x_n \\}.
$$

This kind of notation would also make it more clear when the total number of elements is used (like `mean`), in which case the total number of elements is $n+1$

prko:
@mtmccrea 
Thank you for your opinion and kind feedback!

I agree that it might be overkill to use math notation in both the method description and the discussion. However, I believe it can be helpful for novices. 
If more reviewers think that having the math notation in both places is redundant, I can replace the current pseudo-code and its math notation with the math notation and its explanation from the method description for `sum`,  `product` and `mean`.

Below is the rendered result of the corrected code after incorporating your suggestions. What are your thoughts on it?

Screenshot 2025-02-28 at 15 48 29

mtmccrea:
Thanks for updating the formula!

Note that this statement is not correct:
> (In the mathematic formula, the argument name item in sclang code is represented as i .)

In fact `item` is represented by x_i, where the subscript i is the index.

This is the kind of confusion we'd avoid with an array of numbers that don't happen to also correspond to the index.

prko:
>In fact item is represented by x_i, where the subscript i is the index.

Thanks for pointing this out. It was my fault for not reading the text carefully about the changed mathematical notation. 

The updated result of the rendering is as follows:
Screenshot 2025-02-28 at 16 42 51

mtmccrea:
> The updated result of the rendering is as follows: 

You might consider a more brief wording, such as:

The above code is represented in mathematical notation as:

$$
\sum_{i=0}^n \frac{1}{2^{x_i}} \quad \text{for} \quad \\{ x_0, x_1, \ldots, x_n \\},
$$

where $x_i$ is the function argument `index`.

prko:
Done! 
However, @mtmccrea, I think it is better not to remove (n=8) for beginners, so I did not get rid of it, but I can also remove it if more reviewers want it to be removed.
Screenshot 2025-02-28 at 18 22 44
Screenshot 2025-02-28 at 18 22 44

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,94 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
 ::
 
 method::sum
-Answer the sum of the results of strong::function:: evaluated for each item in the receiver. The function is passed two arguments, the item and an integer index.
+Answer the sum of the items in the collection, optionally applying a link::Classes/Function:: to each item before summing. When provided, the link::Classes/Function:: is passed two arguments: the item and an integer index.
+
+code::
+(0..8).sum { |item| 1 / (2 ** item) }
+// only the item is used; index is unused.
+::
```
capital-G, line 312: What do you think of splitting this up and provide a more step by step explanation, sprinkled which some mathematical notation?
> This is only my suggestion to approach such a documentation, you don't need to change it like this - it may help as an inspiration though :)

```scdoc
seealso:: Classes/Collection#-do
seealso:: Classes/Collection#-collect

Returns the sum of the collection it was called on.
In mathematical notation this is the equivalent to

math::
\sum_{x_i \in x} x_i 
::

for a given collection math::x::.

Example:

code::
(0..4).sum;  // -> 10

(0..4).sum == (0+1+2+3+4)  // -> true
::

argument::function
An optional function which is invoked for every element with two parameters (the element and the index of the element) whose return value will be summed.

In mathematical notation, this would translate to a function math::f: \mathbb{S}, \mathbb{N}_{\geq0} \rightarrow \mathbb{S}:: where ::\mathbb{S}:: is the scalar space of our collection, such that the values of this function will be summed for a given vector math::x \in \mathhbb{S}^n:: via

math::
\sum_{i=0}^{n} f(x_i, i)
::

If no function argument is set it will default to ::nil:: and the values will not be transformed (or - in mathematical notation - math::f:: would be the identity function of math::\mathbb{S}::).

Example:

code::
// only the item is used; index is unused.
(0..8).sum({|item| 1 / (2 ** item)});  // -> 1.99609375
::

is the equivalent to

math::
\sum_{i=0}^{8} f(x_i, i) = \frac{1}{2^{x_i}} ~ \mid ~ x = \lbrace 0, \dots 8 \rbrace
::
```

It seems we don't have a generic `example` block and also missing a `seealso` block for a function, or do I miss something?

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,94 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+discussion::
+The sum of all items in a collection can be expressed mathematically as:
+
+table::
+##teletype::
+[x_0, x_1, ..., x_n].sum // pseudo-code
+::
+equivalent to:
+math::
+\text{Sum} \{ x_0, x_1, \ldots, x_n \} = \prod_{i=0}^{n} x_i
```
capital-G, line 329: I think you mixed something up here?

```suggestion
\text{Sum} \{ x_0, x_1, \ldots, x_n \} = \sum_{i=0}^{n} x_i
```

Also, probably use the "walrus operator" here as it is a definition

```suggestion
\text{Sum} \{ x_0, x_1, \ldots, x_n \} := \prod_{i=0}^{n} x_i
```

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,94 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+code::
+List[1, 2, 3, 4].sum
+::
+
+method::product
+Answer the product of the items in the collection, optionally applying a link::Classes/Function:: to each item before multiplication. When provided, the link::Classes/Function:: receives two arguments: the item and its integer index.
+
+code::
+(0..8).product { |item| 1 / (2 ** item) }
+// only the item is used; index is unused.
```
capital-G, line 343: IMO it is always preferable to add comments _before_ a statement - otherwise the comment will not be picked up by the user if the user struggles with the given statement.

```suggestion
// only the item is used; index is unused.
(0..8).product { |item| 1 / (2 ** item) }
```

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,94 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+
+discussion::
+The product of all items in a collection can be expressed mathematically as:
+table::
+## teletype::
+[x_0, x_1, ..., x_n].product // pseudo-code
+::
+equivalent to:
+math::
+\text{Product} \{ x_0, x_1, \ldots, x_n \} = \prod_{i=0}^{n} x_i
```
capital-G, line 360: Definitions should use the "walrus operator"

```suggestion
\text{Product} \{ x_0, x_1, \ldots, x_n \} := \prod_{i=0}^{n} x_i
```

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,94 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+::
+::
+
+Example:
+code::
+List[1, 2, 3, 4].sum
+::
+
+method::product
+Answer the product of the items in the collection, optionally applying a link::Classes/Function:: to each item before multiplication. When provided, the link::Classes/Function:: receives two arguments: the item and its integer index.
```
capital-G, line 339: I think this could also be structured like the `sum` documentation above.

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,94 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+method::mean
+Answer the arithmetic mean of the items in the collection, optionally applying a link::Classes/Function:: to each item before averaging. When provided, the link::Classes/Function:: receives two arguments: the item and its integer index.
+code::
+(0..8).mean { |item| 1 / (2 ** item) }
+// only the item is used; index is unused.
+::
+
+The code above is mathematically equivalent to:
+math::
+\frac{\sum_{i=0}^{n} \frac{1}{2^{x_i}}}{n-x_0+1} \quad \text{for} \quad \{x_0, x_1, \ldots, x_n\} \quad (n=8)
```
capital-G, line 377: I think this is wrong? Should probably something along 
```suggestion
\frac{\sum_{i=0}^{n} \frac{1}{2^{x_i}}}{n+1} ~ \mid ~ x = \lbrace 0, \dots, n \rbrace, n=8
```

HelpSource/Classes/Collection.schelp
```diff
@@ -304,10 +304,94 @@ List[1, 2, 3, 3, 4, 3, 4, 3].occurrencesOf(3);
...
+
+discussion::
+The mean of all items in a collection can be expressed mathematically as:
+table::
+## teletype::
+[\x_0, \x_1, ..., \x_n].mean // pseudo-code
+::
+equivalent to:
+math::
+\text{Mean}\{ x_0, x_1, \ldots, x_n \} = \frac{\sum_{i=0}^{n} x_i}{n-x_0+1}
```
capital-G, line 389: Walrus
```suggestion
\text{Mean}\{ x_0, x_1, \ldots, x_n \} := \frac{\sum_{i=0}^{n} x_i}{n-x_0+1}
```

```suggestion
\text{Mean}\{ x_0, x_1, \ldots, x_n \} = \frac{\sum_{i=0}^{n} x_i}{n+1}
```

HelpSource/Classes/Collection.schelp
```diff
@@ -547,7 +631,11 @@ symmetricDifference(a, b);
...
-a = Set[1, 2, 3, 4];
-Set[1, 2].isSubsetOf(a); // true
-Set[1, 5].isSubsetOf(a); // false
+a = [1, 2, 3, 4];
+// a = List[1, 2, 3, 4];
+// a = Set[1, 2, 3, 4];
+// a = Interval(1, 4, 1);
+// a = Bag[1, 2, 3, 4];
+[1, 3].isSubsetOf(a); // true
+[1, 5].isSubsetOf(a); // false
```
capital-G, line 640: Could also benefit some nice katex notation, but also okay to do this in another PR.

capital-G (commented):
Sorry for proposing another approach and also requesting some further changes... :)

I think the math notation really suits the quirkiness that one occasionally finds in sclang docs :) I personally really appreciate it, but I also hope these equations are not considered daunting to other users.

HelpSource/Classes/Collection.schelp
prko: Thanks for your suggestions and examples! 
(I love this kind of explanation, but we should wait for others' opinions)

In `SCDoc`, the `example::` tag cannot be used inside a method because the `example::` tag is equivalent to a section. Also, there is no `seealso::` tag, but `link::document_address:::`, so we could use it inside `note::` or `table::`. 
I changed your code to render (For `seealso::`, I used both `note::` and `table::`.):  
```
method::sum
Answer the sum of the items in the collection, optionally applying a link::Classes/Function:: to each item before summing. When provided, the link::Classes/Function:: is passed two arguments: the item and an integer index.

code::
(0..8).sum { |item| 1 / (2 ** item) }
// only the item is used; index is unused.
::
note::
See also: link::Classes/Collection#-do::.

See also: link::Classes/Collection#-collect::.
::

table::
##See also: link::Classes/Collection#-do::.

See also: link::Classes/Collection#-collect::.
::
Returns the sum of the collection it was called on.
In mathematical notation this is the equivalent to

math::
\sum_{x_i \in x} x_i
::

for a given collection math::x::.

Example:

code::
(0..4).sum;  // -> 10

(0..4).sum == (0+1+2+3+4)  // -> true
::

argument::function
An optional function which is invoked for every element with two parameters (the element and the index of the element) whose return value will be summed.

In mathematical notation, this would translate to a function math::f: \mathbb{S}, \mathbb{N}_{\geq0} \rightarrow \mathbb{S}:: where math::\mathbb{S}:: is the scalar space of our collection, such that the values of this function will be summed for a given vector math::x \in \mathbb{S}^n:: via

math::
\sum_{i=0}^{n} f(x_i, i)
::

If no function argument is set it will default to teletype::nil:: and the values will not be transformed (or - in mathematical notation - math::f:: would be the identity function of math::\mathbb{S}::).

Example:

code::
// only the item is used; index is unused.
(0..8).sum({|item| 1 / (2 ** item)});  // -> 1.99609375
::

is the equivalent to

math::
\sum_{i=0}^{8} f(x_i, i) = \frac{1}{2^{x_i}} ~ \mid ~ x = \lbrace 0, \dots 8 \rbrace
::

discussion::
The sum of all items in a collection can be expressed mathematically as:

table::
##teletype::
[x_0, x_1, ..., x_n].sum // pseudo-code
::
equivalent to:
math::
\text{Sum} \{ x_0, x_1, \ldots, x_n \} = \prod_{i=0}^{n} x_i
::
::

Example:
code::
List[1, 2, 3, 4].sum
::
```
The rendered result is as follows:
Screenshot 2025-02-28 at 20 18 53

HelpSource/Classes/Collection.schelp
prko: Thanks! I committed the second one.

HelpSource/Classes/Collection.schelp
prko: Thanks!

HelpSource/Classes/Collection.schelp
prko: like as follows?
### example 1:
```supercollider
code::
a = [1, 2, 3, 4];
// a = List[1, 2, 3, 4];
// a = Set[1, 2, 3, 4];
// a = Interval(1, 4, 1);
// a = Bag[1, 2, 3, 4];
[1, 3].isSubsetOf(a); // true
[1, 5].isSubsetOf(a); // false
::
In mathematical notation, the last two lines of the above code  are the equivalent to  
math::
[1, 3] \subseteq [1, 2, 3, 4]  \Rightarrow  \text{True}
::
math::
[1, 5] \subseteq [1, 2, 3, 4]  \Rightarrow  \text{False} 
::
```
Screenshot 2025-02-28 at 21 09 02

### example 2
```supercollider
code::
a = [1, 2, 3, 4];
// a = List[1, 2, 3, 4];
// a = Set[1, 2, 3, 4];
// a = Interval(1, 4, 1);
// a = Bag[1, 2, 3, 4];
::
table::
##sclang||equivalent mathmatical notation
##code::
[1, 3].isSubsetOf(a); // true
::
||math::
[1, 3] \subseteq [1, 2, 3, 4]  \Rightarrow  \text{True}
::
##code::
[1, 5].isSubsetOf(a); // false
::
||math::[1, 5] \subseteq [1, 2, 3, 4]  \Rightarrow  \text{False}
::
::
```
Screenshot 2025-02-28 at 21 20 58

p.s. I do not know why the first mathematical notation is indented in the table. As soon as I find out, I will open a PR about it.

prko:
@capital-G Thanks for correcting typos and making better suggestions.

1. @capital-G Please confirm which is better in 
- https://github.com/supercollider/supercollider/pull/6537#discussion_r1975323874.
- https://github.com/supercollider/supercollider/pull/6537#discussion_r1975261917.

  I will update according to your suggestion.

3. Regarding https://github.com/supercollider/supercollider/pull/6537#discussion_r1975261917:
I think the perspectives on this part seem to be different. I think @mtmccrea wants a simple explanation and @capital-G wants a step-by-step explanation. I want a detailed step-by-step explanation, although my explanation will not be as professional as @capital-G's. We need more opinions or consensus.

4. The following is the rendered result of https://github.com/supercollider/supercollider/pull/6537/commits/133f4519c4290ffab8232d66358a17c8f00b6c51: Screenshot 2025-02-28 at 22 34 30.

HelpSource/Classes/Collection.schelp
prko: Partially done, except for the detailed step-by-step explanation. Let's see how the others react. If the detailed step-by-step explanation is preferred, I will create text for `.product` and `.mean` based on your text for `.sum`. See https://github.com/supercollider/supercollider/pull/6537#issuecomment-2690728941.

mtmccrea:
Yes, in general I prefer a more simple explanation. It's nice to ideate on ways to use the math notation, but IMO adding notation related to scalar space, and mentioning things like identity functions is a out of scope for SC docs (at least in method documentation). 

Of course this really depends on readers' comfort level with math concepts and notation, but as it is now, I think it's distracting from simply describing the methods. It could be that reducing the equations to one per description could help, as I mentioned before, rather than having one in the description and discussion.

telephon:
https://bufferbuffer.com/why-writing-small-docs-is-a-game-changer/

Yes, let's please keep it as lean as possible. One formula per method is enough! Readers can infer the rest.

HelpSource/Classes/Collection.schelp
prko: @capital-G The first example is implemented: https://github.com/supercollider/supercollider/pull/6537/commits/8f27ef6d9e0b8f759af0853f2ea8a55572dfeb14

I can change this with the second one if you want.

prko:
Thank you for your feedback on simplifying explanations and keeping the documentation lean.

Comparing SuperCollider's documentation with Max's, I see opportunities to improve beginner friendliness. For example, I feel the need to follow @capital-G's .sum example approach, which makes concepts more accessible to newcomers without a programming background. While a complete overhaul would require significant resources...

Although I think that leaving only one mathematical notation for `.sum' without a function helps the newcomer, I also agree that leaving only one clear formula per method helps the reader to grasp essential concepts without confusion.

Following this principle, I've tried to simplify the mathematical notations and explanations in my latest commit https://github.com/supercollider/supercollider/pull/6537/commits/8f27ef6d9e0b8f759af0853f2ea8a55572dfeb14, focusing on clarity while avoiding overwhelming detail.

Could you please review these changes? I'm open to any additional adjustments needed to get this PR approved.

I would like to strike a balance between simplicity and accessibility, while adhering to the project's documentation standards.

Thank you for your consideration.

mtmccrea:
What do you think about this approach?

It attempts to:
- keep the description concise
- clarifies the use of the optional function arg by moving the description into a dedicated `argument::` tag
  - this does introduce 2 equations into the description, but I found @capital-G's math notation of the function argument helpful
- makes use of the iterator `i` to disambiguate it from `item`
- use a collection whose items don't equal the index, to avoid confusion
- remove the redundancy of the `discussion` (List) example

Screenshot 2025-03-10 at 11 38 48

cdbzb:
hi tiny editorial suggestion: text reads `is expressed in mathematical notation` then `is represented in mathematical notation` - suggest both should be the same, perhaps `can be represented`

prko:
@mtmccrea 
@cdbzb 
I think it is good. I will wait a few days for others to respond and then update the PR based on your suggestion and others' feedback. Thank you very much!

prko:
Thank you for your feedback. I apologise for my shortcomings and truly appreciate your patience as I am learning rather than fully contributing at this stage. Thank you for your understanding.

Rendered result of https://github.com/supercollider/supercollider/pull/6537/commits/e982977e953cab803c90816ea9d0188c88a6df42:
Screenshot 2025-03-22 at 10 17 51

Screenshot 2025-03-22 at 10 22 28

telephon (approved):
Good, thank you!

mtmccrea:
Thank you for all your hard work on this @prko! It's been a helpful discussion for future additions to the docs as well.

Issue #6523, opened 2024-11-01
QT6 build IDE window switching lags, makes IDE almost unusable


telephon:
## Environment

* SuperCollider version: d52b0522b (develop)
* Operating system: macOS Sonoma 14.7 **occurs on an intel chipset, pre M1**
* Other details: [QT6](qt: stable 6.7.2 (bottled), HEAD) 

## Problem

- Switching between windows of the IDE, or from another app to the IDE is very laggy. The cpu load also shoots up. 
- Other GUI tasks, like drawing lines or GUI elements are not slowed. Also switching between windows in sclang is **not slowed**.

telephon:
We tested a compile on an M1 macOS computer and there this issue did not occur. Maybe it is related to intel chipsets.

Could someone else try to run the binary on an intel? That would help to find out if it is relevant.

PR #6540, opened 2024-11-20
Disable IDE menu bar icons on macOS

capital-G:
Fixes https://github.com/supercollider/supercollider/issues/6523

## Purpose and Motivation

Only affects macOS.

Currently the IDE takes around 1-2 seconds after "tabbing into" before it displays, while also consuming lots of CPU.
I profiled the application and it turns out that the scaling of icons consumes lots of CPU cycles.

grafik

I remembered that the icons appeared with Qt6 (see https://github.com/supercollider/supercollider/pull/6475#issuecomment-2351937292) so, I decided to disable the icons and the "tabbing in" performance is as in Qt5 now.

Old:
grafik

With PR:
grafik

For some reason this problem only seems to affect Intel Macs (verification needed), so maybe there is a better way than simply disabling the Icons? I looked through https://doc.qt.io/qt-6/qiconengine.html (which is present in the flamegraph), but I couldn't find a setting.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/main_window.cpp
+++ b/editors/sc-ide/widgets/main_window.cpp
@@ -508,6 +508,9 @@ void MainWindow::createMenus() {
     // On Mac, create a parent-less menu bar to be shared by all windows:
 #ifdef Q_OS_MAC
     menuBar = new QMenuBar(0);
+    // icons in menu bars impacts the performance on (intel) macs considerably
+    // while switching windows, therefore we de-activate those. (#6523)
+    QApplication::instance()->setAttribute(Qt::AA_DontShowIconsInMenus, true);
 #else
     menuBar = this->menuBar();
 #endif


telephon:
Fixes the issue for me!

dyfer:
Just so I understand - what do you mean by "tabbing in"? Switching to SuperCollider from another app?

I think this is good.
FWIW I thought that I did observe sluggish performance of the IDE with a local build on Arm64 macOS, but I can't recreate it now (both this PR and an old build seem to work fast). In any case, I think we should merge this if it fixes the issue for others.

dyfer (approved):

capital-G:
> Just so I understand - what do you mean by "tabbing in"? Switching to SuperCollider from another app?

Sorry - yes, using cmd+tab to switch programs.
Thx for the review, in this case I'll merge it.

PR #6542, opened 2024-11-22
Update hidapi

xunil-cloud:
## Purpose and Motivation

Update hidapi to fix `incompatible-pointer-types` error. See https://github.com/supercollider/supercollider/issues/6435.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [x] All tests are passing
- [x] Updated documentation


capital-G:
Thanks for the fix! Looks good from my side.

Do you think the "reversion" of the temporary fix https://github.com/supercollider/supercollider/pull/6436 should be made in this PR or in a separate PR? Maybe it is good to combine this to see if the CI works?

xunil-cloud:
>Maybe it is good to combine this to see if the CI works?

I think this is good, but PortMidi is also impacted by this. Fortunately, the fix is actually [simple](https://github.com/PortMidi/portmidi/blob/7a5de5b7597c46f963d72a83defe7592f901e5f1/pm_win/pmwinmm.c#L989C36-L989C45).

I can make a seperate PR for portmidi and also remove the temporary fix in https://github.com/supercollider/supercollider/pull/6436 or I just make the changes all together is this PR.

capital-G:
I think this depends if we want to fix our version of portmidi or switch to upstream via git subrepo - do you have a perference on this?

If we fix it within our repo I think it is fine to combine this fix and the CI change in one go, if we should switch over to upstream I think this should be a dedicated PR.

xunil-cloud:
I tried to [update](https://github.com/xunil-cloud/supercollider/tree/update-portmidi) portmidi to v2.0.4. But it failed on every Windows CI builds. I suspect that we still need this [patch](https://github.com/supercollider/supercollider/commit/6b265703cc83c93708f9279f15a7db65fb06c5f3), which seems strange. I guess we will handle portmidi in a separate PR.

capital-G:
Ok, thanks for trying it out. Then I'd say we merge this as-is?

xunil-cloud:
OK.

capital-G (approved):

xunil-cloud:
@capital-G I’ve figured out why it failed. I might create a PR to update PortMidi later, then. :smiley:

PR #6546, opened 2024-11-24
Do not change indentation within multiline comments

capital-G:
## Purpose and Motivation

Currently the IDE resets the indentation within multi-line comments - this PR changes the behavior.

E.g. given the code

image

Currently the IDE aligns the indentation within the multi-line comment with its surrounding, resulting in

image

This PR disables the indentation within multi-line comments, so the code would look like in image 1 after indentation. 
It also does not align/indent a line if it is equal to `/*` (the beginning of a multi-line comment).

I am not sure if this is the proper way to implement, therefore I would greatly appreciate a review.

edit: it seems win32 does not allow for `and` within if cases? Am I right if this should be switched for `&&`?
Also sorry for pushing directly as a sc branch, I forgot to set the proper origin.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

## Merge note

Should be squashed.

### DIFF
--- a/editors/sc-ide/widgets/code_editor/highlighter.cpp
+++ b/editors/sc-ide/widgets/code_editor/highlighter.cpp
@@ -245,8 +245,10 @@ void SyntaxHighlighter::highlightBlock(const QString& text) {
             break;
 
         default:
-            if (lexer.state() >= ScLexer::InComment)
+            if (lexer.state() >= ScLexer::InComment) {
                 highlightBlockInComment(lexer);
+                blockData->isInMultilineComment = true;
+            }
         }
     }
 
--- a/editors/sc-ide/widgets/code_editor/sc_editor.cpp
+++ b/editors/sc-ide/widgets/code_editor/sc_editor.cpp
@@ -595,7 +595,6 @@ void ScCodeEditor::indentCurrentRegion() { indent(currentRegion()); }
 void ScCodeEditor::indent(EditBlockMode editBlockMode) { indent(textCursor(), editBlockMode); }
 
 void ScCodeEditor::indent(const QTextCursor& selection, EditBlockMode editBlockMode) {
-    static QString plainMultilineCommentBegin = QString("*/");
     if (selection.isNull())
         return;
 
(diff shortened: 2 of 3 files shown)


JordanHendersonMusic:
> edit: it seems win32 does not allow for and within if cases? Am I right if this should be switched for &&?

Yes it is annoying, I think I made an issue for it a while ago. The fix is just setting a compiler flag.

telephon (approved):
Behaves correctly and fixes the bug.

capital-G:
Thanks for verifying the function @telephon !
Please do not merge before we had a review on the C++ part as this currently also breaks the windows build.

> Yes it is annoying, I think I made an issue for it a while ago. The fix is just setting a compiler flag.

A quick SA search showed this https://stackoverflow.com/a/74832138 - maybe it is better to stick to the good ol' `&&` in this case?

capital-G:
I switched from `and` to `&&` which resolves the failed build for windows.

I also removed the WIP label as this may scare off potential reviewers? Please do not merge before such a review :)

capital-G:
> @capital-G – what do we need in terms of review?

Someone who greenlights my C++ *Stückwerk*.

To me it seems really hacky to add member variables at a very far away part of the code from its actual usage (i.e why do we add indentation related information while working on analyzing code for highlighting?) and also the string comparison of `/*` could be a rather expensive operation, but I am not sure if there is another way around this.
To me the PR somehow violates the separation of concerns, yet this is not a library but a Qt gui one could say and it also happens at other places.

Maybe someone has a better idea / can have a look if this is considered valid?

JordanHendersonMusic:
Can you do `some_q_string == "*/"` with qstrings?

Having a separate pass to create indentation data seems a bit overkill for this feature. 

I think this pr is fine.

I don't have time to test, but what happens if you do...

```
1 + 1;
               /* Indented start of comment
Something 
         Indented something 
*/
```
Does it still put the '/*' symbol on the far left?

capital-G:
> Does it still put the '/*' symbol on the far left?

In this case yes, because the start of the comment will be aligned to its current indetation except the line is exactly `*/`.
This is chosen because the multi line comment could start in a line which contains code which most likely should be aligned, e.g. the tool would not change the format of this code

```supercollider
f = {
  1 + 1; /* Indented start of comment
something
*/
}
```



> Can you do `some_q_string == "*/"` with qstrings?

Don't know - I could try if this is preferable as it should also be a static.

JordanHendersonMusic:
>  I could try if this is preferable as it should also be a static.

Literals don't allocate, they are a part of the function definition. Making a QString like that might do an allocation (I think they do COW and I don't know if they have optimisations for when it's backed by a literal), but it probably has unnecessary overhead.

capital-G:
Switched to your proposed solution.

I also found a bug - we want to compare for `/*`, not for `*/`

JordanHendersonMusic (approved):
Looks good to me!

PR #6549, opened 2024-11-25
Add file tree widget to IDE

capital-G:
## Purpose and Motivation

Adds a file tree widget to the IDE. This widget can be positioned/opened/closed like a post window etc.

image

Double-clicking on a file will open the file in the IDE.
Right clicking on a file will show the menu

image

I think the main advantage is for bigger projects and to navigate within sample databases.

I coded this some weeks ago but before I would get the final touches on it I think it would be good to know what the general feedback on such a widget would be.

## Questions

* What should be the default directory to be opened? Currently it is `~/`
* Currently any file will be openend in the IDE, so even binary files. Maybe this should be limited to non-binary files (how?)
* Should the file tree follow
* Should there be an sclang API for this (e.g. `IDEFileTree.navigateTo("~/my/folder")`) - if this is possible
* Allow to display file attributes such as filesize and mtime in a dedicated column

## ToDo

* Test Windows and Linux
* Navigating to the parent of `~/` is possible but does not order the items accordingly anymore

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/CMakeLists.txt
+++ b/editors/sc-ide/CMakeLists.txt
@@ -88,6 +88,7 @@ set ( ide_moc_hdr
     widgets/tool_box.hpp
     widgets/audio_status_box.hpp
     widgets/lang_status_box.hpp
+    widgets/file_tree.hpp
     widgets/code_editor/editor.hpp
     widgets/code_editor/sc_editor.hpp
     widgets/code_editor/highlighter.hpp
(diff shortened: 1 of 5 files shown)


telephon:
One thing: if such a tree view exists, each session should reconstruct its previous tree.

smoge:
My opinion is that this feature should get inspiration from the idea of a 'project', like treemacs in emacs, or any other IDE.

So, when you open a folder, that folder is considered your 'project', and the tree browser will stick to this root folder for this session.

Issue #6488, opened 2024-09-16
sclang crash/segfault on exit when built with newer Xcode


dyfer:
## Environment

* SuperCollider version: `develop` (both local build and GHA build)
* Operating system: macOS 14
* Other details (Qt version, audio driver, etc.): platform: arm64, local build using Xcode 15.4

~~Update: also can be observed on x86_64 and Xcode 15 - it seems to be associated with Xcode/clang version.~~

After testing various GHA environments I observed that the issue seems to occur with Xcode 14 (Apple Clang 14) and higher.

This will eventually prevent us from running tests after upgrading build environment in GHA, see #6489 

EDIT:
The error does _not_ seem to occur if SC is [run on macOS 12](https://github.com/supercollider/supercollider/actions/runs/11301406336/job/31436226860#step:1:4) (i.e. lower than 13), even if it was built using [Xcode 15](https://github.com/supercollider/supercollider/actions/runs/11301406336/job/31435787765#step:12:25). I confirmed the same build crashes on exit on macOS 14. In practical terms it probably doesn't change much, we still need to figure out why it crashes on newer Xcode + newer OS.

## Steps to reproduce

```sh
# run sclang in terminal
./Install/SuperCollider/SuperCollider.app/Contents/MacOS/sclang
... (sclang starts)
sc3> 0.exit
-> 0
main: waiting for input thread to join...
sc3> main: quitting...
cleaning up OSC
zsh: segmentation fault  Install/SuperCollider/SuperCollider.app/Contents/MacOS/sclang
```
System report:

<details>

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               sclang [30768]
Path:                  /Volumes/VOLUME/*/SuperCollider.app/Contents/MacOS/sclang
Identifier:            sclang
Version:               ???
Code Type:             ARM-64 (Native)
Parent Process:        zsh [45056]
Responsible:           iTerm2 [50972]
User ID:               501

Date/Time:             2024-09-16 12:07:59.9674 -0700
OS Version:            macOS 14.5 (23F79)
Report Version:        12
Anonymous UUID:        B401E5DC-9974-9C9F-9DE4-F9F4274CD77C

Sleep/Wake UUID:       711A23C0-2503-4055-843D-119B1B5BEF02

Time Awake Since Boot: 2600000 seconds
Time Since Wake:       6108 seconds

System Integrity Protection: enabled

[...]

Issue #6531, opened 2024-11-12
macOS 12 Actions Runner will be removed December 3rd


frenchy64:
Related https://github.com/actions/runner-images/issues/10721

We use this to run the [macOS tests](https://github.com/supercollider/supercollider/blob/90dbcbddaddc072f615fabd22214388311748f64/.github/workflows/test_macos.yml#L11). I'm not sure which version we want to move to.

frenchy64:
I [tried](https://github.com/frenchy64/supercollider/actions/runs/11789496665/job/32838978143) macos-13 and there's a segfault during the tests:

```
compiling class library...
Path override failed for key base::DIR_APP_DICTIONARIES and path '/Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/MacOS/qtwebengine_dictionaries'
	Found 865 primitives.
	Compiling directory '/Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/Resources/SCClassLibrary'
	numentries = 841006 / 1[25](https://github.com/frenchy64/supercollider/actions/runs/11789496665/job/32838978143#step:9:26)09320 = 0.067
	5540 method selectors, 2258 classes
	method table size 13371768 bytes, big table size 100074560
	Number of Symbols 12436
	Byte Code Size 380723
	compiled 328 files in 0.75 seconds
compile done
localhost : setting clientID to 0.
internal : setting clientID to 0.
Class tree inited in 0.01 seconds


*** Welcome to SuperCollider 3.14.0-dev. *** For help type cmd-d.

Adding include path /Users/runner/work/supercollider/supercollider/testsuite/classlibrary

cleaning up OSC
/Users/runner/work/_temp/f8841b6a-4788-4a93-a62f-09715d76b400.sh: line 2:  1562 Segmentation fault: 11  $SCLANG $SCRIPT_ADD_PATH $TESTS_PATH
```

capital-G:
We are working on fixing this, see https://github.com/supercollider/supercollider/issues/6488

Once this is fixed, we will bump the runner image.

PR #6551, opened 2024-11-26
Destroy UDP socket explicitly

capital-G:
Big thanks to @Spacechild1 who came up with the solution and helped me a lot!

## Purpose and Motivation

Fixes https://github.com/supercollider/supercollider/issues/6488 and enables https://github.com/supercollider/supercollider/issues/6531

## Types of changes

Releasing the socket explicitly to avoid "double-destruction" of the global socket variable.

- Bug fix
 
## To-do list

Would be great if all major platforms could verify it works - simply exit `sclang` and if it exits without a crash/segfault, everything is how it should be.

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/lang/LangPrimSource/OSCData.cpp
+++ b/lang/LangPrimSource/OSCData.cpp
@@ -862,9 +862,8 @@ void init_OSC(int port) {
 #endif
 
     startAsioThread();
-
     try {
-        gUDPport.reset(new InPort::UDP(port, HandlerType::OSC));
+        gUDPport = std::make_unique<InPort::UDP>(port, HandlerType::OSC);
     } catch (std::exception const& e) { postfl("No networking: %s", e.what()); }
 }
 
@@ -923,6 +922,10 @@ void closeAllCustomPorts() {
 
 void cleanup_OSC() {
     postfl("cleaning up OSC\n");
+
+    // NOTE: the socket must be destroyed *before* the IO service.
+    // We cannot rely on the global object destructor because the order would be undefined.
+    gUDPport = nullptr;
 
     stopAsioThread();
 


lang/LangPrimSource/OSCData.cpp
```diff
@@ -925,6 +926,9 @@ void cleanup_OSC() {
     postfl("cleaning up OSC\n");
 
     stopAsioThread();
+    // destroy socket after ASIO thread has been joined!
```
telephon, line 929: Just regarding your comment: does `stopAsioThread();` join the thread? I would have thought it just ends it.

capital-G: The existing function joins the thread

https://github.com/capital-G/supercollider/blob/c87ec0782b351a47b24f23ff3a8b454b7873715c/lang/LangPrimSource/SC_ComPort.cpp#L62-L65

I'll maybe try to refactor the asio thread into its own class so we don't mangle with global variables from two files which makes it really scattered and hard to follow, but this would be another PR.

JordanHendersonMusic:
> // NOTE: don't rely on the global object destructor!!

Did you find out why this isn't working? This should be well defined C++ behaviour and perhaps points to a bug elsewhere.

capital-G:
I still try to wrap my head around this which part manages which resource, but most likely @Spacechild1 has to chime in here.

IIUC, the UDP socket attaches to a service, and when we close this service the UDP port also tries to cancel any outstanding tasks, see https://live.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/basic_datagram_socket/_basic_datagram_socket.html

> This function destroys the socket, cancelling any outstanding asynchronous operations associated with the socket as if by calling cancel. 

As the ioService has already been released when we exit the main method (where all global variables such as `gUdpPort` will be freed), this will create a crash.

Although I still don't fully understand why we don't change the order into 

```cpp
    gUDPport = nullptr;
    stopAsioThread();
```

b/c it seems udpPort has a reference to the `ioService` which will be released when we close the thread.

Both orders seem to fix the bug.

JordanHendersonMusic:
Are you saying there is a circular reference between `ioService` and `gUDPport`? If so, using shared and weak ptrs might be cleaner? My reservation here is that you are doing manual memory management when smart pointers are specifically designed to avoid having to do this.

capital-G:
First off sorry if not everything that I am referring to is correct, I am still a novice in C++ resource management.

The problem seems to be global variables which are wrapped in smart pointers so they only get released after the program exits (I don't understand the real use of them in this case).

I'll try to unwrap it and please correct me where I am wrong.

The language client calls `init_osc`

https://github.com/supercollider/supercollider/blob/b9984860a1c566e45ab33e6f8875908f0bb924d6/lang/LangSource/SC_LanguageClient.cpp#L108

which spawns an asio thread and allocates a new UDP socket in the global variable `gUDPPort`, located in `OSCData`.

https://github.com/supercollider/supercollider/blob/b9984860a1c566e45ab33e6f8875908f0bb924d6/lang/LangPrimSource/OSCData.cpp#L864-L868

The global variables ASIO thread and io service of `SC_ComPort`  are initiated by the call  `startAsioThread`

https://github.com/supercollider/supercollider/blob/b9984860a1c566e45ab33e6f8875908f0bb924d6/lang/LangPrimSource/SC_ComPort.cpp#L47-L60

In the same file, the constuctor of the UDP socket claims (?) this ioService as a member variable

https://github.com/supercollider/supercollider/blob/b9984860a1c566e45ab33e6f8875908f0bb924d6/lang/LangPrimSource/SC_ComPort.cpp#L189-L194

As the UDP socket therefore requires the ioService it is necessary to shut them down together (in a proper order(?), see my comment above).

Maybe we could wrap the`ioService` into a `shared_ptr` so it only gets destructed after there is no reference left?

But IMO it would be much clearer and cleaner to wrap everything that is dependent on each other into an object/class which can then be attached to another object instead of relying on multiple global variables across files that all need to be created in a proper order - e.g. wrap every resource necessary for the OSC thread into its own `OSCAsioThread` class and use constructors and destructors for resource management and try to keep everything as stateless and encapsulated as possible and attach this object to the `LanguageClient` object.

lang/LangPrimSource/OSCData.cpp
```diff
@@ -925,6 +926,9 @@ void cleanup_OSC() {
...
+    // NOTE: don't rely on the global object destructor!!
```
Spacechild1, line 930: Maybe we can expand this a bit:

```
// NOTE: the socket must be destroyed *before* the IO service, so we cannot rely on the global object destructor because the order would be undefined.
```

Spacechild1:
TL;DR: the PR is fine as it is but the comments and the commit message could be expanded.

> Releasing the socket explicitly to avoid "double-destruction" of the global socket variable.

First off, the PR description is a bit misleading because this is not about double free, but about undefined destruction order.

To clear things up:

An ASIO socket holds a reference to the IO service object and the socket destructor automatically unregisters the socket from the service. This means that the service must be the last object to be destroyed. In our case the problem was that both the socket and the service are global variables and the order of construction/destruction of global variables is undefined! This is the famous https://en.cppreference.com/w/cpp/language/siof. To solve this issue, I explicitly release the socket in the `cleanup_OSC` method. After all, we also have to *create* the socket at the right place, i.e. in `init_OSC`.

Now, you could put the ASIO thread and the socket into some class, but you would still need to make sure that the class is created after resp. destroyed before the IO service object. But the `init_OSC` and `cleanup_OSC` functions already do this! Why use classes when you can use simple functions?

To be honest, I think that `init_OSC` and `cleanup_OSC` are pretty clear in what they do:

1. create a UDP socket
2. start the network thread

resp.

1. stop/join the network thread
2. destroy the UDP socket

With the addition of https://github.com/supercollider/supercollider/pull/6551/files#r1860973685 things should be pretty clear IMO.

However, I would suggest to expand the commit message, e.g.:
```
destroy upd socket explicitly 

The UDP socket internally holds a reference to the IO service and therefore must be destroyed *before* the latter. In particular, we can't rely on the global object destructors because the order would be undefined.
```

capital-G:
Thanks for your summary, I think I finally got it and finally explains this https://github.com/supercollider/supercollider/issues/6488#issuecomment-2492349592 - the problem is not a double destruction but trying to de-register from the `ioService` which gets deleted "at the same time" (aka in an undefined order) as the `socket` due both being a global variable, but the order of releasing is actually important here.

> Why use classes when you can use simple functions?

I personally find classes very good at encapsulating resources and functionality - to me it acts like namespace and makes code much easier to understand, separate and refactor.

I don't know much C++ yet but https://en.cppreference.com/w/cpp/language/raii seems like a good pattern to me and IIUC, the current implementation is not following this pattern. Ofc the ressource order would still be made to be explicit here, but instead of remembering to calling the "close asio thread" function one could rely on the constructor and destructor of such objects.

E.g. Couldn't be the RAII pattern be used to implicitly enforce the order of con- and destruction? Then we have guarded this order on a static type level which seems always preferable?
Make it mandatory to provide an io_service for the construction of the UDP port instead of taking the global io_service (which would have to be initiated before) - if explicitly passing the service you get around the problem of having set up the environment before.
Having side-effect free code seems preferable in such complex situations.
Also during destruction of the thread object it would be clear that the socket has to be released prior to releasing the service.

JordanHendersonMusic:
@Spacechild1 thanks! This also explains why it was compiler dependent.

Spacechild1:
I have the feeling you're overcomplicating things.

> I don't know much C++ yet but https://en.cppreference.com/w/cpp/language/raii seems like a good pattern to me and IIUC, the current implementation is not following this pattern. Instead of remembering to calling the "close asio thread" function one could rely on the constructor and destructor of such objects which are also more safer in regards to exceptions.

RAII is great for resource management, but as you've acknowledged yourself, it doesn't save you from ordering issues. By moving the initialization and shutdown code to a constructor resp. destructor you're just adding another layer of indirection. You still have to manually create and destroy that class instance, though, to guarantee the correct ordering, so you would still have `init_OSC` and `cleanup_OSC` functions. We are basically talking about two objects ( `UDPport` and `asioThread`). `UDPport` would still have to be globally accessible since it's used in quite a few primitive functions. The only thing this would buy us is hiding the `asioThread` object, but noone would need to access it anyway (apart from `startAsioThread` and `stopAsioThread`).

Actually, both `gUDPport` and `ASIOport` aren't accessed outside their respective source files and could therefore be `static` resp. in an anonymous namespace.

Also note that there are other UDP sockets (`gCustomUdpPorts`) that run on the same ASIO thread but follow a different life cycle (they are destroyed whenever the class library is recompiled).

> Make it mandatory to provide an io_service for the construction of the UDP port instead of taking the global io_service (which would have to be initiated before)

`ioService` is a global variable and as such initialized before resp. destroyed after `main`. As long as it's only accessed from within `main`, there are no issues.

Generally, we should avoid global variables, but some things are truely global. For example, if you have a logger object, you wouldn't pass that down to every function just so you can log a message, you would have it in a global variable or global getter function. In our case, `ioService` is essentially a singleton and needs to be accessible from many places; it totally makes sense to have it as a global extern variable.

That being said, adding `io_service` as an argument to the `InPort::UDP` constructor wouldn't be a bad thing.

> edit: regarding the commit message: force-push it or how should we handle this?

I would say force-push.

Spacechild1:
Anyway, I won't try to stop you from refactoring, but I think there are better areas to spend your time and energy on...

capital-G:
Force pushed with more information. This should be ready for merging, please review.

JordanHendersonMusic:
This looks good to me now. Intent is much clearer. I don't have a machine that this crashed on so I can't test, but assuming it fixes the problem, I think this is good to be merged.

FYI, if you made the `io_service` a raw pointer, could also do this with the custom destructor of the unique_ptr. 
Not saying you should, the function calls are fine, but it is a way to use the type system to get this behaviour without too much refactoring.
```c++
#include <memory>
#include <iostream>

struct Foo {
    Foo() { std::cout << "constructor\n"; }
    Foo(Foo&&) = default;
    Foo& operator =(Foo&&) = default;
    ~Foo() { std::cout << "deleter\n"; }
};

int main() {
    const auto mk_foo = []() -> Foo* {
        auto f = new Foo();
        std::cout << "starting other thing\n";
        return f;
    };
    const auto destroy_foo = [](Foo* f) -> void {
        std::cout << "stopping other thing\n";
        f->~Foo();
    };
    auto ptr = std::unique_ptr<Foo, decltype(destroy_foo)>(mk_foo(), destroy_foo);
}


// constructor
// starting other thing
// stopping other thing
// deleter
```

Spacechild1:
> FYI, if you made the io_service a raw pointer, could also do this with the custom destructor of the unique_ptr.
Not saying you should, the function calls are fine, but it is a way to use the type system to get this behaviour without too much refactoring.

That's not a good idea because you can't join threads in a global object destructor. You would still need to manually create and destroy the whole thing `init_OSC` and `cleanup_OSC` so it doesn't really buy you anything except for adding another layer of indirection.

Spacechild1:
I have a little comment, but I need to go to a concert. Will write later in the evening.

JordanHendersonMusic:
> you can't join threads in a global object destructor

I did not know that! Do you have any source for this, or any terms to Google? I'm trying to learn more info but not really finding anything other than a few stack overflow posts about DLLs.

Spacechild1:
> > you can't join threads in a global object destructor
> 
> I did not know that! Do you have any source for this, or any terms to Google? I'm trying to learn more info but not really finding anything other than a few stack overflow posts about DLLs.

Turns out I misremembered and this is only true for DLLs. I just tried and it seems to work in EXEs both with mingw and MSVC. In dynamically loaded DLLs you will definitely get deadlocks, I can tell from experience. Generally, you should avoid running code before/after `main()` when possible.

Instead of abusing `std::unique_ptr`, you can make a simple class that starts and stops the asio thread and create an instance at the top of the `main()` function. `ioService` can still be a global variable since it needs to be globally accessible anyway.

If you want to use RAII for automatic resource management *in a well defined order* while still providing global access, here's a pattern I've found quite useful:
```
class Test* gTest;

class Test {
public:
    Test()
    {
        // initialize...
        // then set global variable
        gTest = this;
    }

    ~Test
    {
        // shutdown
    }
};

int main()
{
    Test test;

    // remaining code
    return 0;
}
```

Turns out, sclang already uses this pattern for the language client instance:
https://github.com/supercollider/supercollider/blob/b9984860a1c566e45ab33e6f8875908f0bb924d6/lang/LangSource/SC_LanguageClient.cpp#L74-L89

And so does Supernova for the server instance:
https://github.com/supercollider/supercollider/blob/b9984860a1c566e45ab33e6f8875908f0bb924d6/server/supernova/server/server.cpp#L45-L56


So instead of creating a dedicated class just for the ASIO thread management, you can simply call `startAsioThread` and `stopAsioThread` in the language client constructor/destructor! RAII will ensure proper cleanup.

Generally, I think it is a bit of misconception that every resource initialization/cleanup needs to be wrapped in a dedicated class. It's totally fine to use a pair of functions and call them in the constructor/destructor of an existing class.

FWIW, I do think that `init_OSC()` resp. `cleanup_OSC()` shouldn't be responsible for asio thread management. However, instead of wrapping this in a class, I would simply manage the asio thread in the constructor/destructor of `SC_LanguageClient`, or better, `HiddenLanguageClient`. The UDP socket management should stay in `init_OSC()` and `cleanup_OSC()`, though.

Spacechild1:
@capital-G I was wrong about the ordering between thread and socket creation/shutdown!

In traditional socket programming you must not close a socket concurrently while there are pending read/write operations, that's why I suggested to first join the asio thread and then destroy the socket. However, asio sockets are always closed asynchronously, so there is no race condition. (After all, custom UDP ports would still be added/removed while the ASIO thread is running.) So the natural order is to create the asio thread before any sockets and only join it after all sockets have been destroyed. In practice, both methods work, but my comment is pretty misleading.

Therefore I would suggest to the following changes:

In `init_OSC`, replace
```
    // first create the UDP socket
    try {
        gUDPport = std::make_unique<InPort::UDP>(port, HandlerType::OSC);
    } catch (std::exception const& e) { postfl("No networking: %s", e.what()); }
    // then start the network thread
    startAsioThread();
```
with
```
    startAsioThread();

    try {
        gUDPport = std::make_unique<InPort::UDP>(port, HandlerType::OSC);
    } catch (std::exception const& e) { postfl("No networking: %s", e.what()); }
```

And in `cleanup_OSC` replace
```
    stopAsioThread();
    // destroy socket after ASIO thread has been joined!
    // NOTE: the socket must be destroyed *before* the IO service,
    // so we cannot rely on the global object destructor because the order would be undefined.
    gUDPport = nullptr;
```
with
```
    // NOTE: the socket must be destroyed *before* the IO service.
    // We cannot rely on the global object destructor because the order would be undefined.
    gUDPport = nullptr;

    stopAsioThread();
```

capital-G:
> Therefore I would suggest the following changes:

Thanks - I changed it and forced-pushed again. I think this answers my prior question @ https://github.com/supercollider/supercollider/pull/6551#issuecomment-2503558676 :)

I know that it is a bit useless to refactor this part of the code, but it is the only part of sclang that I am familiar with and I still find relying on global variables really confusing and leads easily to errors like seen here.
Also from what I have learned in my programming experience: functions should not have any side-effects, this is something what methods are for  - but actually: methods also should also not have any side-effects beyond con- and destructor, instead use types to ensure the necessary state of an object via the type system.

I also think the naming could be a bit more precise - e.g. one is not just creating an asio thread but also implicitly an asio runtime which gets created as a side-effect of the function.
I have an idea how to refactor this, even if it doesn't get merged maybe you can give a review on this so I can improve my C++ skills, but this is another PR.

JordanHendersonMusic:
> Why global variables at all though?

It sounds like a lot of effort to refactor everything to pass the state around explicitly. The VM uses globals everywhere too, so this isn't an isolated issue.

Spacechild1 (approved):

Spacechild1:
I think this PR should be ready to merge now. Maybe @telephon?

---

> I know that it is a bit useless to refactor this part of the code, but it is the only part of sclang that I am familiar with and I still find relying on global variables really confusing and leads easily to errors like seen here.

The error we've seen here is not unique to global variables, though. You can get the same bug with RAII just by declaring class members in the wrong order. For example:

```
class Foo {
    std::unique_ptr<InPort::UDPCustom> mUdpPort;
    asio::io_service mService;
};
```

With the default destructor this would lead to the same crash, as `mService` would be destructed before `mUdpPort`. RAII only makes sure that resources are automatically releases when they go out of scope. With interdependent resources you are still responsible for ensuring the correct order!

> Also from what I have learned in my programming experience: functions should not have any side-effects, this is something what methods are for

It really depends. A `clamp()` function certainly shouldn't have side effects, but a `printf` function certainly will. There are no hard rules!

> but actually: methods also should also not have any side-effects beyond con- and destructor (which are actually also not methods)

That's nice in theory, but it doesn't hold up in practice. It's very normal for methods to modify state outside its own class. For example, a SC ugen can write samples into a sound buffer or audio bus.

>> Side note: If you want to use RAII for automatic resource management in a well defined order while still providing global access, here's a pattern I've found quite useful:
> Why global variables at all though? Java/C# don't have it and as we are one level below Java we should be even stricter in regards to using the type system to avoid clashes due to implicit behavior whenever possible.

TL;DR: it's for convenience.

First off, Java has been designed during the OOP craze of the 90s and probably shouldn't serve as our role model. Heck, it doesn't even have free functions! Java is the prime example of what's wrong with jamming everything into a class.

Also, Java certainly does have global variables, they are just disguised as static class members :) Here's a simple example:
```
System.out.println("hello world");
```
Of course, `System.out` is really a global variable!

Certain resources are naturally singletons, i.e. there will only ever be single instances. Some examples:
- logging
- rendering
- event loop
- audio I/O
- the app itself!

Typically, these resources are used throughout the code base and therefore have to be easily accessible. 

For example, you certainly don't want to pass a `Logger` instance to every function just in case it - or any downstream function! - want to log something. Neither would you store a reference in every object that might require logging.

Sometimes, a global reference is just the natural solution. It might be disguised as static class members, static class methods or global functions, but at the end of the day they are all the same.

The pattern I've showed has the big advantage that you can control the initialization and shutdown sequence - which is often not possible with traditional "Singleton" patterns! - while still providing convenient global access.

Global objects are often combined with global functions. For example, you'd rather want to write
```
LogError("ouch!");
``` 
and not
```
mApp->GetLogger()->logError("ouch!");
```

Here's another example, this time from the SC plugin interface. The interface table can be accessed through the `World* mWorld` member, but plugins still store a reference to a global variable to enable various macros. For example, this allows us to write
```
RTAlloc(mWorld, size);
```
instead of
```
mWorld->ft->fRTAlloc(mWorld, size);
```

Another thing: if a *plugin* wants to share state across instances, there's often no other choice than using global variables. Where else would you put it?

To sum things up: yes, you shouldn't scatter your application state across global variables and instead try to keep it in a class, but it can be totally fine to provide certain *unique* resources through global object pointers or global getter functions.

> Also Rust does not rely on global variables and Rust seems for me a good guidance towards safe and clean code.

Well, C++ is not Rust. Rust's borrow checker makes it hard to use global variables. On the one hand, this is a good thing, on the other hand, global variables can be really convenient. Again, different languages have different paradigms.

dyfer:
I skimmed through the discussion - it seems there has been a consensus and this is ready to be merged, but since there were many implementation details considered, I thought I'd double check: is everyone feeling this is good to go?

capital-G:
@dyfer if you think this is fine I think this is ready to be merged.

---

Thanks for sharing your insights @Spacechild1 , I really appreciate them :)

> The error we've seen here is not unique to global variables, though. You can get the same bug with RAII just by declaring class members in the wrong order. For example:
> 
> ```
> class Foo {
>     std::unique_ptr<InPort::UDPCustom> mUdpPort;
>     asio::io_service mService;
> };
> ```

I thought that by wrapping `mUpdPort` into its own class/object the order of destruction could be made implicit, but you are right, there seems no way to ensure this via e.g. the type system.

> Different languages have different paradigms. I would suggest to first get familiar with the language and its patterns before attempting any code refactoring.

Yeah, I think I am guilty of preferring the ways I am used to instead of exposing myself of what is the common way in other programming languages, great to get some feedback/adjustment to this!

telephon:
This still persists for me (macOS 14): https://github.com/supercollider/supercollider/issues/6488

Issue #977, opened 2013-10-12
Shift-Ctrl-Tab should work to cycle back through tabs


danstowell:
If you press Ctrl-Tab quickly and release it, it cycles "forward" through the tabbed documents.

If you press Shift-Ctrl-Tab, it inserts a tab character into the current document. This is unexpected. Shift-Ctrl-Tab should be the "inverse" of Ctrl-Tab, as it is in analogous tabbed-browsing GUIs like firefox.

(Shift-Tab is the key to insert a tab character. I don't believe that Shift-Ctrl-Tab is deliberately designated for that.)

**Note:** if you press Ctrl-Tab and keep Ctrl held down, then the tab selection floaty menu is visible, and **then** you can use Shift-Ctrl-Tab. But you can't use it to initiate.

**Note**: On MacOS this is `ALT+TAB` and `ALT+SHIFT+TAB`

vdonnefort:
We can trigger "switchDocument" when hitting Ctrl-Shift-Tab. But I tested it and I don't think this is a good idea. 

The behavior is not next/prev tab like with web browsers but it uses a history. If you hit once ctrl+tab, it goes to the next tab. if you hit it once again, it will go back to the previous one and not the next. This way, a Ctrl-Shift-Tab shortcut would be strange.

alln4tural:
Hi Geoffrey, 

that ( ctrl+tab and shift+ctrl+tab to cycle through open docs in the IDE) been working for me for years, on windows and macos. Does it not work for you? 

(there still is the quirk that Dan reported originally) 

Cheers, 
eddi
-- 
https://soundcloud.com/all-n4tural
https://alln4tural.bandcamp.com


On 21/01/2019, 15:28 Geoffroy Montel <notifications@github.com> wrote:
I agree that it could be nice that ctrl+tab would go to next document, and ctrl+shift+tab would go to previous document, like on a Web browser.
The change is easy to do, but as it may break old habits, do we all agree to do it ?

—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or mute the thread.
[...]

PR #6554, opened 2024-11-30
Use tab to switch between documents

capital-G:
This post will be kept up-to-date to reflect the state of the PR.

## Purpose and Motivation

Closes https://github.com/supercollider/supercollider/issues/977 by using ctrl+(shift)+tab to cycle through documents like tabs in a browser. (edit: this is actually control on macOS and windows).
This replaces the alt + arrow keys shortcut.

Works on macOS, needs testing on Windows and Linux.

This could be added to https://dev.docs.supercollider.online/Reference/KeyboardShortcuts.html but I somehow think this document should include not just macOS shortcuts, so I think this should be done in another PR?

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/multi_editor.cpp
+++ b/editors/sc-ide/widgets/multi_editor.cpp
@@ -560,18 +560,18 @@ void MultiEditor::createActions() {
 
     mActions[NextDocument] = action = new QAction(tr("Next Document"), this);
 #ifndef Q_OS_MAC
-    action->setShortcut(tr("Alt+Right", "Next Document"));
+    action->setShortcut(tr("Ctrl+Tab", "Next Document"));
 #else
-    action->setShortcut(tr("Ctrl+Alt+Right", "Next Document"));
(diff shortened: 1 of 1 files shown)


telephon:
Thank you!

xunil-cloud:
I test this on Linux and Qt says something like "Ctrl-Tab" shortcut is conflicted. 🫠 "Ctrl-Tab" also binds to "Switch Document". If I use "Meta-Tab" for "Switch Document" (right below your changes), then, it works well. "Ctrl-Shift-Tab" works just fine.

---

And how about we go back to the first document if we reach the last one?

telephon (approved):

dyfer:
@capital-G I assume this can be merged?

capital-G:
> @capital-G I assume this can be merged?

Yes, feel free to merge if you think this is fine :)

PR #6555, opened 2024-12-01
GHA: move macOS tests to the latest OS

### LINKED ISSUE #6531: macOS 12 Actions Runner will be removed December 3rd

dyfer:
## Purpose and Motivation

macOS 12 runners will be disabled soon; let's run the tests on macOS 15.

Fixes #6531 

EDIT: it turned out that on newer macOS runners we need to grant microphone access to `/usr/local/opt/runner/provisioner/provisioner`. This PR adds it as well.

## Types of changes

- Maintenance
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -48,6 +48,9 @@ jobs:
           HOMEBREW_NO_INSTALL_CLEANUP: 1
         run: brew install coreutils # to get 'timeout' utility
 
+      - name: Grant microphone access # solution from https://github.com/actions/runner-images/issues/9330#issuecomment-1938616628
+        run: sqlite3 $HOME/Library/Application\ Support/com.apple.TCC/TCC.db "INSERT OR IGNORE INTO access VALUES ('kTCCServiceMicrophone','/usr/local/opt/runner/provisioner/provisioner',1,2,4,1,NULL,NULL,0,'UNUSED',NULL,0,1687786159,NULL,NULL,'UNUSED',1687786159);"
+
       - name: Run tests
         shell: bash
         continue-on-error: true
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test-macos:
-    runs-on: macos-12
+    runs-on: macos-15
     env:
       INSTALL_PATH: ${{ github.workspace }}/build/Install
       ARTIFACT_FILE: ${{ inputs.artifact-file }}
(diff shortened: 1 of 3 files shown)


dyfer:
Oh no, there seems to be some incompatibility between our server and the audio subsystem on the arm64 macOS 15 runners: https://github.com/supercollider/supercollider/actions/runs/12101205465/job/33741278060?pr=6555#step:6:539

I'll revert this PR to an x64 runner...

dyfer:
It turned out to be "microphone access" permission issue. I found the solution [here](https://github.com/actions/runner-images/issues/9330#issuecomment-1938616628).

dyfer:
When debugging this (by capturing a screenshot in the gha runner...), I also noticed that the provisioner asks for permission to access local networks:
debugscreen

This didn't seem to prevent the tests from running, so I didn't add setting that permission.

.github/workflows/test_macos.yml
```diff
@@ -48,6 +48,9 @@ jobs:
           HOMEBREW_NO_INSTALL_CLEANUP: 1
         run: brew install coreutils # to get 'timeout' utility
 
+      - name: Grant microphone access
+        run: sqlite3 $HOME/Library/Application\ Support/com.apple.TCC/TCC.db "INSERT OR IGNORE INTO access VALUES ('kTCCServiceMicrophone','/usr/local/opt/runner/provisioner/provisioner',1,2,4,1,NULL,NULL,0,'UNUSED',NULL,0,1687786159,NULL,NULL,'UNUSED',1687786159);"
```
capital-G, line 52: Maybe add a comment to the URL where this originates in case this needs an update/fix?

capital-G (changes requested):
Great detective work, much appreciated!

One small change request and afterwards we can merge this.

dyfer:
> This looks like some github runner service requesting network access, I don't think this has something to do with sc.

This is the service that seems to run the jobs - that is the one that also needed microphone access, so I assume this gets triggered when sclang or scsynth start.

.github/workflows/test_macos.yml
dyfer: This URL? https://github.com/actions/runner-images/issues/9330#issuecomment-1938616628

capital-G: Yes :)

dyfer: done!

capital-G (approved):

PR #6556, opened 2024-12-01
Fix `incompatible-pointer-types` warning/error for portmidi

### LINKED ISSUE #6435: fix `incompatible-pointer-types` warnings

xunil-cloud:
## Purpose and Motivation

Fix `incompatible-pointer-types` warning/error for portmidi. Fixes #6435.

I originally wanted to fix this by updating protmidi to a newer version, but then realized the upstream fix has not been released yet.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,10 +104,6 @@ if (${CMAKE_COMPILER_IS_GNUCXX})
 
     add_compile_options(
         -Werror=return-type      # Error out if function definitions lack a return type
-        # GCC 14 has made -Wincompatible-pointer-types an error by default and as a
-        # consequence, some external C would fail to compile. This is a temporary fix!
-        # Ideally, the affected dependencies should be fixed upstream.
-        -Wno-error=incompatible-pointer-types
         -fschedule-insns2        # Optimize instruction scheduling
         -fomit-frame-pointer     # Optimize by omitting the frame pointer where possible
         -fno-math-errno          # Optimize math functions by not setting errno
--- a/external_libraries/portmidi/pm_win/pmwinmm.c
+++ b/external_libraries/portmidi/pm_win/pmwinmm.c
@@ -1049,7 +1049,7 @@ static PmError winmm_write_byte(PmInternal *midi, unsigned char byte,
         m->hdr = hdr = get_free_output_buffer(midi);
         assert(hdr);
         midi->fill_base = (unsigned char *) m->hdr->lpData;
-        midi->fill_offset_ptr = &(hdr->dwBytesRecorded);
+        midi->fill_offset_ptr = (uint32_t *) &(hdr->dwBytesRecorded);
         /* when buffer fills, Pm_WriteSysEx will revert to calling
          * pmwin_write_byte, which expect to have space, so leave
          * one byte free for pmwin_write_byte. Leave another byte


CMakeLists.txt
```diff
@@ -104,10 +104,6 @@ if (${CMAKE_COMPILER_IS_GNUCXX})
 
     add_compile_options(
         -Werror=return-type      # Error out if function definitions lack a return type
```
capital-G, line 106: The temporary fix https://github.com/supercollider/supercollider/pull/6436/files also turned `-Wreturn-type` into `-Werror=return-type`.
Should this be reverted as well?

Otherwise this seems fine to me, GCC14 pipeline runs.

xunil-cloud: >Should this be reverted as well?

It seems that this is unrelated to the temporary fix, so we might keep this. (maybe we can ask Spacechild1?)

[quote](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wreturn-type) from GCC doc:
>-Wreturn-type

>    Warn whenever a function is defined with a return type that defaults to int (unless -Wimplicit-int is active, which takes precedence). Also warn if execution may reach the end of the function body, or if the function does not contain any return statement at all. 

and `Werror=return-type` will make the `return-type` warning into an error.

capital-G (approved):

CMakeLists.txt
Spacechild1: Yes, we should definitely keep this! I once spend half a day trying to fix a bug before I realized I forgot the return statement in a method. From this day, the first thing I do in every new project is `-Werror=return-type`. (I don't understand why this still isn't the default...)

Spacechild1 (approved):

PR #6557, opened 2024-12-01
GHA: fix caching on macOS

dyfer:
## Purpose and Motivation

I noticed that "macOS x64 use system libraries" GHA job typically takes longer to build. Upon closer look, I realized that it uses the same cache string as the "shared libscsynth" job.

I know that we've discussed whether we really need both of these... I'm not sure, but for now I proposed to fix caching so that it's differentiated between these builds.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- n/a Updated documentation

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -106,8 +106,8 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/Library/Caches/ccache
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Well spotted

dyfer:
I confirmed the caching works. Merging.

PR #6558, opened 2024-12-01
GHA workflow: move the definition of artifacts

dyfer:
## Purpose and Motivation

When I was updating test jobs, I found it cumbersome to have to modify the artifact name in a different file than the test job itself. I think the proposed change is cleaner to maintain.

If others don't like it I won't push for it, but this would be my preference.

## Types of changes

- Maintenance

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- n/a Updated documentation


capital-G (approved):
Looks valid to me - this makes `actions.yml` a dispatcher and each test action needs to resolve the provided version information accordingly.

PR #6563, opened 2024-12-03
Add basic zoom via touchpad gesture

capital-G:
## Purpose and Motivation

Allows to zoom via touchpad. Only tested on macOS. Is not the "native macOS" zoom as it only increases the font size and does not increase the focus of a viewport - still useful.

https://github.com/user-attachments/assets/b11ff198-d7e6-435b-8f02-285d8f458b9d

Also added a boundary for font size.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/code_editor/editor.cpp
+++ b/editors/sc-ide/widgets/code_editor/editor.cpp
@@ -873,7 +873,7 @@ void GenericCodeEditor::resetFontSize() { mDoc->resetDefaultFont(); }
 
 void GenericCodeEditor::zoomFont(int steps) {
     QFont currentFont = mDoc->defaultFont();
-    const int newSize = clampFontSize(currentFont.pointSize() + steps);
+    const float newSize = clampFontSize(currentFont.pointSizeF() + steps);
     currentFont.setPointSizeF(newSize);
     mDoc->setDefaultFont(currentFont);
 }
--- a/editors/sc-ide/widgets/code_editor/editor.cpp
+++ b/editors/sc-ide/widgets/code_editor/editor.cpp
@@ -30,6 +30,7 @@
 
 #include <QApplication>
 #include <QDebug>
+#include <QGestureEvent>
 #include <QGraphicsView>
 #include <QKeyEvent>
 #include <QPainter>
@@ -100,6 +101,9 @@ GenericCodeEditor::GenericCodeEditor(Document* doc, QWidget* parent):
     doc->setLastActiveEditor(this);
 
     applySettings(Main::settings());
+
+    grabGesture(Qt::PinchGesture);
+    setAttribute(Qt::WA_AcceptTouchEvents);
 }
 
 GenericCodeEditor::~GenericCodeEditor() {
@@ -457,6 +461,23 @@ QString GenericCodeEditor::symbolUnderCursor() {
     return tokenInStringAt(position, blockString);
 }
 
+bool GenericCodeEditor::gestureEvent(QGestureEvent* event) {
+    if (QGesture* pinch = event->gesture(Qt::PinchGesture)) {
+        auto* pinchGesture = static_cast<QPinchGesture*>(pinch);
+        if (pinchGesture->state() == Qt::GestureUpdated) {
+            qreal scaleFactor = pinchGesture->scaleFactor();
+            if (scaleFactor >= 1.0) {
+                // some magic number - 2 seems to "feel good"
+                zoomFont(2);
+            } else {
+                zoomFont(-2);
+            }
+        }
+        return true;
+    }
+    return false;
+}
+
 bool GenericCodeEditor::event(QEvent* event) {
     switch (event->type()) {
     case QEvent::ShortcutOverride: {
@@ -470,6 +491,9 @@ bool GenericCodeEditor::event(QEvent* event) {
             return true;
         }
         break;
+    }
+    case QEvent::Gesture: {
+        return gestureEvent(static_cast<QGestureEvent*>(event));
     }
     default:
         break;
@@ -855,7 +879,7 @@ void GenericCodeEditor::resetFontSize() { mDoc->resetDefaultFont(); }
 void GenericCodeEditor::zoomFont(int steps) {
     QFont currentFont = mDoc->defaultFont();
     const int newSize = currentFont.pointSize() + steps;
-    if (newSize <= 0)
+    if (newSize <= 7 || newSize >= 50)
         return;
     currentFont.setPointSize(newSize);
     mDoc->setDefaultFont(currentFont);
--- a/editors/sc-ide/widgets/code_editor/editor.cpp
+++ b/editors/sc-ide/widgets/code_editor/editor.cpp
@@ -465,8 +465,8 @@ bool GenericCodeEditor::gestureEvent(QGestureEvent* event) {
     if (QGesture* pinch = event->gesture(Qt::PinchGesture)) {
         auto* pinchGesture = static_cast<QPinchGesture*>(pinch);
         if (pinchGesture->state() == Qt::GestureUpdated) {
-            qreal scaleFactor = pinchGesture->scaleFactor();
-            zoomFontScaler(scaleFactor);
+            float scaleFactor = pinchGesture->scaleFactor();
+            zoomFont(scaleFactor);
         }
         return true;
     }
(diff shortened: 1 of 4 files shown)


dyfer:
Thanks!
Do you think we should also expose an option in preferences to turn it on/off?

capital-G:
I don't think it will be necessary to turn it off? Browsers or vscode also don't allow to turn this off either.

dyfer (changes requested):
Thanks! I think this is a very useful addition.

I have two requests here:
1. Scale Factor - the zoom seems too large / sensitive to me. Please compare it with the pinch zooming of the help browser window (native) - I think they should feel the same and currently they don't.
BTW pinch zooming of the help browser window seems to be centered in the corner instead of from the cursor position, but that's for a separate issue/PR.
2. When trying this in, I realized that I'd expect the post window to also zoom this way, would you be able to add that in?

capital-G:
> 1. Scale Factor - the zoom seems too large / sensitive to me. Please compare it with the pinch zooming of the help browser window (native) - I think they should feel the same and currently they don't IMO.

I adjusted the "zoom speed" from 3 to 2 - I think it is closer to the Help Browser now. Hope it is better now.

> 2. When trying this in, I realized that I'd expect the post window to also zoom this way, would you be able to add that in?

Thanks for the suggestions. I now added the same "pinch functionality" to the post window.

As the zoom level is not globally but for each document, I also made this zoom level of the post window independent of the document.

dyfer:
I still wasn't convinced by the "feel" of the zoom, so I tried to use floats for scaling. Here's my approach: https://github.com/dyfer/supercollider/commit/ec57db8d4a6b3959d5676602e78add1e09ba4fd2 
It uses the pinch gesture scale factor directly to scale the font size (float). What do you think? 

I just implemented the main doc pinch, but I think it should be the same for the post window

dyfer:
Additionally, for the min/max, I think it would be great to set these in relation to the currently set font. IDE allows setting font size between 1 and 99 - way beyond what a standard setup would need, BUT there might be edge cases....

If possible, I'd suggest setting min to something like 0.25 or maybe 0.3 or even 0.5 of the current size in the settings, and max to 4x or so?

telephon:
I also notice that on very long help documents, on macOS, the scrolling by two-finger-touch is not as you would expect it. It sometimes maps much slower than expected, making it tedious in places.

dyfer:
> I also notice that on very long help documents, on macOS, the scrolling by two-finger-touch is not as you would expect it. It sometimes maps much slower than expected, making it tedious in places.

Thanks for reporting @telephon. Two-finger scrolling is unrelated to this PR, IIUC.
Could you please report this as a separate issue, including which hardware it pertains to and how long the document needs to be for the issue to transpire? I myself didn't notice scrolling issues on M1 MBP, but I also don't use touchpad all that often.

telephon:
Ok, good, here it is: #6603

capital-G:
> If possible, I'd suggest setting min to something like 0.25 or maybe 0.3 or even 0.5 of the current size in the settings, and max to 4x or so?

This is a good suggestion - implemented it. And thanks for your code, I applied it here and modified it a bit.

Do you think this is good to go now, @dyfer?

dyfer:
Thanks @capital-G!

Could you implement the same scaling approach for the post window as well, instead of integers/magic numbers?

capital-G:
> Could you implement the same scaling approach for the post window as well, instead of integers/magic numbers?

Oh, thanks for the hint, forgot about that. Updated the implementation for the post window as well.

editors/sc-ide/widgets/code_editor/editor.cpp
```diff
@@ -855,12 +874,21 @@ void GenericCodeEditor::resetFontSize() { mDoc->resetDefaultFont(); }
 void GenericCodeEditor::zoomFont(int steps) {
     QFont currentFont = mDoc->defaultFont();
     const int newSize = currentFont.pointSize() + steps;
-    if (newSize <= 0)
+    if (newSize <= 7 || newSize >= 50)
```
dyfer, line 877: The pinch zoom works great now - if "feels" good to me and the range is good as well.

However, I'm unsure about this particular change. Right now with the default settings on macOS I can get a slightly smaller font using pinch zoom out than with `cmd -`.

I'd either remove this change, keeping the old behavior in place (no limit), or clamp it to the same value as pinch zoom - `std::clamp(newSize, defaultFontSize * 0.5, defaultFontSize * 8.0);`.

dyfer (changes requested):

editors/sc-ide/widgets/code_editor/editor.cpp
capital-G: Oh, I somehow oversaw that, thanks for pointing this out. I added a `GenericCodeEditor::clampFontSize` function via https://github.com/supercollider/supercollider/pull/6563/commits/caca3c6f7ca54f3ae737cc91ad585fcc3ce89b88 which is also used by the Post Window and also by the keyboard zoom.

editors/sc-ide/widgets/code_editor/editor.cpp
```diff
@@ -854,13 +873,23 @@ void GenericCodeEditor::resetFontSize() { mDoc->resetDefaultFont(); }
 
 void GenericCodeEditor::zoomFont(int steps) {
     QFont currentFont = mDoc->defaultFont();
-    const int newSize = currentFont.pointSize() + steps;
-    if (newSize <= 0)
-        return;
-    currentFont.setPointSize(newSize);
+    const int newSize = clampFontSize(currentFont.pointSize() + steps);
+    currentFont.setPointSizeF(newSize);
```
dyfer, line 877: Does it matter whether we use `setPoinstSizeF` instead of `setPointsSize` here?

dyfer (approved):

dyfer:
@capital-G thanks so much for this! Works great now.

I have one minor outstanding question - see above. Otherwise, this is good to go.

Should be squash merged, of course.

dyfer:
@capital-G I think this is really close, could we wrap this up?

editors/sc-ide/widgets/code_editor/editor.cpp
capital-G: Quoting https://doc.qt.io/qt-6/qfont.html#setPointSizeF

> Sets the point size to pointSize. The point size must be greater than zero. The requested precision may not be achieved on all platforms.

IMO this seems preferable in our case

capital-G:
> @capital-G I think this is really close, could we wrap this up?

If you think the question is sufficiently answered its up to you to merge it b/c this looks fine to me :)

editors/sc-ide/widgets/code_editor/editor.cpp
dyfer: I guess my point is that `newSize` is `int` - so why do we use a float function?

capital-G: ah thx, missed to look one line above... changed this to be float throughout

editors/sc-ide/widgets/post_window.cpp
```diff
@@ -248,13 +252,18 @@ void PostWindow::zoomOut(int steps) { zoomFont(-steps); }
 
 void PostWindow::zoomFont(int steps) {
     QFont currentFont = font();
-    const int newSize = currentFont.pointSize() + steps;
-    if (newSize <= 0)
-        return;
+    const int newSize = GenericCodeEditor::clampFontSize(currentFont.pointSize() + steps);
     currentFont.setPointSize(newSize);
```
dyfer, line 256: This function now uses ints (consistently), but 
[void GenericCodeEditor::zoomFont(int steps)](https://github.com/supercollider/supercollider/pull/6563/files#diff-305813f383824a6e3e1172ebee7b331ad6017ad5c7d3e7ca3ac07f4d9ae11186L855)
uses floats. Not a big deal, but do we want more consistency here? Sorry for being pedantic!

capital-G: np for me being pedantic, always enjoy to go for a good solution. I hope I got it more consistent by using floats

dyfer:
Thanks!

HotwheelsSisyphus:
@dyfer Commenting here so as to not spam the other thread:
On my ubuntu studio, with a current 3.15 dev build, the pinch-to-zoom only works for the help browser, not for editor or post window. Tested this both with wayland and x11.

Operating System: Ubuntu Studio 24.04
KDE Plasma Version: 5.27.12
KDE Frameworks Version: 5.115.0
Qt Version: 5.15.13
Kernel Version: 6.17.0-19-generic (64-bit)
Graphics Platform: Wayland

dyfer:
Thanks @HotwheelsSisyphus !

I don't believe you'd be able to test with Qt6, correct? Where this might work possibly is qt6 + wayland, but not likely.

HotwheelsSisyphus:
@dyfer correct, I'm not able to test with qt6 on the OS side right now. The scide in question was built with qt 6.4.2, btw.

dyfer:
Ah, so SC is running on Qt6? In that case it confirms we'd need a different solution for Linux.

capital-G:
Just tested this on my Linux laptop using qt6 and this indeed doesn't work - at least with my trackpad.

@HotwheelsSisyphus do you want to create an issue for this? I think otherwise this gets lost.

---

* SuperCollider version: 3.15.0-dev (Built from branch 'develop' [b365366bc6])
* Operating system: linux Fedora Linux 42 (Workstation Edition) (x86_64)
* Other details
	* Qt version: 6.10.2
	* argv: 
	* server: scsynth
	* includes:

PR #6564, opened 2024-12-07
Increase UDP socket buffer size

### LINKED ISSUE #5993: Increase UDP socket receive buffer size

xunil-cloud:
## Purpose and Motivation
Set `SOL_SOCKET/SO_RCVBUF` using [boost API](https://live.boost.org/doc/libs/1_86_0/doc/html/boost_asio/reference/basic_datagram_socket/receive_buffer_size.html) to increase default UDP receive socket buffer size. Close #5993 and probably #5870. For now, I increase the buffer size to 8MB. But the choose is somewhat arbitrary and I'm happy to change it if we come up with a better default.

## Types of changes

- New feature

## To-do list

- [x] Code is tested (only on Linux)
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/lang/LangPrimSource/SC_ComPort.cpp
+++ b/lang/LangPrimSource/SC_ComPort.cpp
@@ -205,8 +205,8 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
         }
     }
 
-    boost::asio::socket_base::send_buffer_size sendBufferSize;
     try {
+        boost::asio::socket_base::send_buffer_size sendBufferSize;
         mUdpSocket.get_option(sendBufferSize);
         if (sendBufferSize.value() < UDP::sendBufferSize) {
             sendBufferSize = UDP::sendBufferSize;
@@ -216,8 +216,8 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
         printf("(sclang) SC_UdpInPort: WARNING: failed to set send buffer size\n");
     }
 
-    boost::asio::socket_base::receive_buffer_size receiveBufferSize;
     try {
+        boost::asio::socket_base::receive_buffer_size receiveBufferSize;
         mUdpSocket.get_option(receiveBufferSize);
         if (receiveBufferSize.value() < UDP::receiveBufferSize) {
             receiveBufferSize = UDP::receiveBufferSize;
--- a/lang/LangPrimSource/SC_ComPort.cpp
+++ b/lang/LangPrimSource/SC_ComPort.cpp
@@ -212,7 +212,6 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
         try {
             mUdpSocket.set_option(sendBufferSize);
         } catch (boost::system::system_error& e) {}
-        mUdpSocket.get_option(sendBufferSize);
     }
 
     boost::asio::socket_base::receive_buffer_size receiveBufferSize;
@@ -222,7 +221,6 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
         try {
             mUdpSocket.set_option(receiveBufferSize);
         } catch (boost::system::system_error& e) {}
-        mUdpSocket.get_option(receiveBufferSize);
     }
 
     initHandler(handlerType);
--- a/lang/LangPrimSource/SC_ComPort.cpp
+++ b/lang/LangPrimSource/SC_ComPort.cpp
@@ -207,8 +207,8 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
 
     boost::asio::socket_base::send_buffer_size sendBufferSize;
     mUdpSocket.get_option(sendBufferSize);
-    if (sendBufferSize.value() < 8 * 1024 * 1024) {
-        sendBufferSize = 8 * 1024 * 1024;
+    if (sendBufferSize.value() < UDP::sendBufferSize) {
+        sendBufferSize = UDP::sendBufferSize;
         try {
             mUdpSocket.set_option(sendBufferSize);
         } catch (boost::system::system_error& e) {}
@@ -217,8 +217,8 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
 
     boost::asio::socket_base::receive_buffer_size receiveBufferSize;
     mUdpSocket.get_option(receiveBufferSize);
-    if (receiveBufferSize.value() < 8 * 1024 * 1024) {
-        receiveBufferSize = 8 * 1024 * 1024;
+    if (receiveBufferSize.value() < UDP::receiveBufferSize) {
+        receiveBufferSize = UDP::receiveBufferSize;
         try {
             mUdpSocket.set_option(receiveBufferSize);
         } catch (boost::system::system_error& e) {}
--- a/lang/LangPrimSource/SC_ComPort.cpp
+++ b/lang/LangPrimSource/SC_ComPort.cpp
@@ -206,22 +206,22 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
     }
 
     boost::asio::socket_base::send_buffer_size sendBufferSize;
-    mUdpSocket.get_option(sendBufferSize);
-    if (sendBufferSize.value() < UDP::sendBufferSize) {
-        sendBufferSize = UDP::sendBufferSize;
-        try {
+    try {
+        mUdpSocket.get_option(sendBufferSize);
+        if (sendBufferSize.value() < UDP::sendBufferSize) {
+            sendBufferSize = UDP::sendBufferSize;
             mUdpSocket.set_option(sendBufferSize);
-        } catch (boost::system::system_error& e) {}
-    }
+        }
+    } catch (boost::system::system_error& e) { printf("(sclang) SC_UdpInPort: WARNING: failed to set send buffer size\n"); }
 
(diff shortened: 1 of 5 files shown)


capital-G:
Haven't reviewed it yet but sound like a good idea. Maybe this would be also a good addition for supernova?

capital-G (changes requested):
Thanks for implementing the supernova side as well.

There are some smaller comments, but generally I think it is nice and useful addition.

It could be useful to set this value via a CLI flag, but this could be implemented in a later PR if someone thinks it becomes necessary.

server/scsynth/SC_ComPort.cpp
```diff
@@ -261,8 +261,21 @@ class SC_UdpInPort {
...
 
-        boost::asio::socket_base::send_buffer_size option(65536);
-        udpSocket.set_option(option);
+        boost::asio::socket_base::send_buffer_size send_buffer_size(65536);
+        udpSocket.set_option(send_buffer_size);
+
+        boost::asio::socket_base::receive_buffer_size recv_buffer_size;
+
+        udpSocket.get_option(recv_buffer_size);
+        scprintf("SC_UDP_PORT: default socket receive buffer size: %d\n", recv_buffer_size.value());
```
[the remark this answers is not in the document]
xunil-cloud: Ah, that is for debugging purpose. I can remove that later.

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -611,6 +611,11 @@ void sc_osc_handler::open_udp_socket(ip::address address, unsigned int port) {
     else
         sc_notify_observers::udp_socket.open(udp::v4());
 
+    boost::asio::socket_base::receive_buffer_size recv_buffer_size(8 * 1024 * 1024);
```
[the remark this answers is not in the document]
xunil-cloud: >Or is there a reason to not set the value for supernova?

I noticed that we already set send buffer size to 65536 in sclang and scsynth. I'm not sure why we chose 65536, probably because the field for packet size is 32 bits long in UDP header? However, doing this will likely reduce the send buffer size compared to the default on Linux. (On Linux, typical value is 212992. When you set to 65536, kernel will double it, so it is 131072.) Maybe the behavior is different on other OS?

dyfer:
Thanks @xunil-cloud for the PR. What do we need to wrap it up?

xunil-cloud:
The behavior is now consistent across scsynth, supernova, and sclang. We will configure both the UDP send and receive buffer sizes to 8MB. I also implemented Spacechild1's suggestion: we won't make any changes if the default buffer size is already greater than 8MB.

dyfer:
@capital-G @Spacechild1 do you think this now good to be merged?

capital-G (approved):
Looks fine to me, but @Spacechild1 should have a look and should merge if he thinks everything is ok.

Thanks for the PR! :)

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -205,8 +205,25 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
         }
     }
 
-    boost::asio::socket_base::send_buffer_size option(65536);
-    mUdpSocket.set_option(option);
+    boost::asio::socket_base::send_buffer_size send_buffer_size;
```
Spacechild1, line 208: nitpick, but we generally use camelCase, not snake case, i.e. `sendBufferSize` instead of `send_buffer_size`. Same for other variables.

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -205,8 +205,25 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
...
+    mUdpSocket.get_option(send_buffer_size);
+    if (send_buffer_size.value() < 8 * 1024 * 1024) {
```
Spacechild1, line 210: It's not good to repeatedly hardcode numbers. Instead I would suggest to add 
```
static constexpr int sendBufferSize = 8 * 1024 * 1024;
```
to the `UDP::UDP` class and use that. If we want to change the send buffer size, we can then do it in one place.

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -205,8 +205,25 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
...
+        send_buffer_size = 8 * 1024 * 1024;
+        try {
+            mUdpSocket.set_option(send_buffer_size);
+        } catch (boost::system::system_error& e) {}
+        mUdpSocket.get_option(send_buffer_size);
+    }
+
+    boost::asio::socket_base::receive_buffer_size recv_buffer_size;
+    mUdpSocket.get_option(recv_buffer_size);
+    if (recv_buffer_size.value() < 8 * 1024 * 1024) {
```
Spacechild1, line 220: Similarly, let's add
```
static constexpr int receiveBufferSize = 8 * 1024 * 1024;
```
to `UDP::UDP`.

(In the future, someone may want to use different buffer sizes for sending and receiving.)

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -205,8 +205,25 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
         }
     }
 
-    boost::asio::socket_base::send_buffer_size option(65536);
-    mUdpSocket.set_option(option);
+    boost::asio::socket_base::send_buffer_size send_buffer_size;
+    mUdpSocket.get_option(send_buffer_size);
```
Spacechild1, line 209: Strictly speaking, `get_option` can also throw. Just wrap the whole code for each buffer in a try-catch statement.

Also, I would post a warning message in the catch-clause.

server/scsynth/SC_ComPort.cpp
```diff
@@ -261,8 +261,25 @@ class SC_UdpInPort {
         if (inPortNum == 0)
             mPortNum = udpSocket.local_endpoint().port();
 
-        boost::asio::socket_base::send_buffer_size option(65536);
-        udpSocket.set_option(option);
```
Spacechild1, line 265: See my comments in the other file.

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -611,6 +611,26 @@ void sc_osc_handler::open_udp_socket(ip::address address, unsigned int port) {
     else
         sc_notify_observers::udp_socket.open(udp::v4());
```
Spacechild1, line 613: See my comments in the other file. The only exception: supernova uses snake case, so you can keep it here!

Spacechild1 (changes requested):

Spacechild1:
Thanks for the PR! Generally, it looks good to me. I just added some remark regarding code style and error handling.

dyfer:
@xunil-cloud would you be able to respond to @Spacechild1 's suggestions?

xunil-cloud:
I'm happy to make the changes. I think I can do that this week.

dyfer:
We are planning to make the release in about 12 hours, FYI. 
Not mean to rush, it just won't make it to the release.

dyfer:
@xunil-cloud just FYI we weren't able to finish the release today. We plan to release 3.14.0-rc1 on Monday. If this is merged by then, it'll make the release. Thanks!

xunil-cloud:
OK, I think it's ready for review again.

lang/LangPrimSource/SC_ComPort.cpp
```diff
@@ -205,8 +205,27 @@ UDP::UDP(int inPortNum, HandlerType handlerType, int portsToCheck): mPortNum(inP
         }
     }
 
-    boost::asio::socket_base::send_buffer_size option(65536);
-    mUdpSocket.set_option(option);
+    boost::asio::socket_base::send_buffer_size sendBufferSize;
```
Spacechild1: Just a little comment (feel free to ignore it): there is no need to declare the buffer size variables outside the try-block.

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -611,6 +611,24 @@ void sc_osc_handler::open_udp_socket(ip::address address, unsigned int port) {
...
 
+    boost::asio::socket_base::send_buffer_size send_buffer_size;
+    try {
+        udp_socket.get_option(send_buffer_size);
+        if (send_buffer_size.value() < sc_osc_handler::udp_send_buffer_size) {
+            send_buffer_size = sc_osc_handler::udp_send_buffer_size;
+            udp_socket.set_option(send_buffer_size);
+        }
+    } catch (boost::system::system_error& e) { printf("WARNING: failed to set send buffer size\n"); }
```
Spacechild1: Nitpick, but supernova generally uses `std::cout` instead of `printf`

Spacechild1:
Looks good! If you have time, you can address my two comments. Otherwise we can just merge it as is on Monday.

lang/LangPrimSource/SC_ComPort.cpp
xunil-cloud: That makes sense. I changed it.

server/supernova/sc/sc_osc_handler.cpp
xunil-cloud: Done.

Spacechild1 (approved):

Spacechild1:
Thanks!

PR #6566, opened 2024-12-08
classlib: more tolerant plot

### LINKED ISSUE #6565: plot array that has subarrays and one number

Expected:
```supercollider
// if you happen to have a number in one of the dimensions:
[[1, 2, 3], [6, 2, 1], 0].plot  // plot something reasonable
```

Actual
```supercollider
// if you happen to have a number in one of the dimensions:
[[1, 2, 3], [6, 2, 1], 0].plot  // => some spec error
```

telephon:
## Purpose and Motivation

Fixes #6565

It easily happens that a subarray is a number.

## Types of changes

- more tolerant plot behaviour

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation
- Release Notes: plot doesn't fail with opaque error when subarray turns out to be a number in a multichannel array

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -1290,8 +1290,8 @@ Plotter {
 		depth = array.maxDepth;
 		hasSubArrays = depth > 1;
 
-		if(depth > 2) {
-			"Cannot currently plot an array with more than 2 dimensions".warn;
+		if(depth > 3) {
+			"Cannot currently plot an array with more than 3 dimensions".warn;
 			^nil
 		};
 
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -1284,13 +1284,17 @@ Plotter {
 
 + ArrayedCollection {
 	plot { |name, bounds, discrete = false, numChannels, minval, maxval, separately = true, parent|
-		var array, plotter;
+		var array, plotter, depth, hasSubArrays;
 		array = this.as(Array);
 
-		if(array.maxDepth > 3) {
-			"Cannot currently plot an array with more than 3 dimensions".warn;
+		depth = array.maxDepth;
+		hasSubArrays = depth > 1;
+
+		if(depth > 2) {
+			"Cannot currently plot an array with more than 2 dimensions".warn;
 			^nil
 		};
+
 		plotter = Plotter(name, bounds, parent);
 		if(discrete) {
 			plotter.plotMode = \points
@@ -1309,7 +1313,11 @@ Plotter {
 				if(elem.isNil) {
 					Error("Cannot plot array: non-numeric value at index %".format(i)).throw
 				};
-				elem
+				if(hasSubArrays) {
+					elem.asArray
+				} {
+					elem
+				}
 			}
 		};
 
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -1290,8 +1290,8 @@ Plotter {
 		depth = array.maxDepth;
 		hasSubArrays = depth > 1;
 
-		if(depth > 3) {
-			"Cannot currently plot an array with more than 3 dimensions".warn;
+		if(depth > 2) {
+			"Cannot currently plot an array with more than 2 dimensions".warn;
 			^nil
 		};
 
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -932,11 +932,16 @@ Plotter {
 
 	superpose_ { |flag|
 		var dom, domSpecs;
+		if(flag and: { value.isRectangular.not }) {
+			"Plotter can't superpose unequally sized arrays.".warn;
+			^this
+		};
+
 		dom = domain.copy;
 		domSpecs = domainSpecs.copy;
 
 		superpose = flag;
-		if ( value.notNil ){
+		if(value.notNil) {
 			this.setValue(value, false, false);
 		};
 


telephon:
> I agree this is a reasonable addition. It may have side effects though...

yes, let's check.
 
> How does this behave with setting/getting `domainSpec`s?
>
> https://github.com/supercollider/supercollider/blob/a4ffd92d88e4f89d426318ac35b21ad12bd8a11f/HelpSource/Classes/Plotter.schelp#L471


 [[0, 1, 2, 3, 4, 5, 6, 7], [8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]].plot;
a = [[0, 1, 2, 3, 4, 5, 6, 7], 4,  [8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16], 8].plot; // this works now, instead of throwing an error

a.domainSpecs = [\freq, [0, 100], [0, 30], \freq, \freq];


 
> And is it robust as to where in the multichannel of data the single-value lies? E.g.
> 
> ```
> [[1, 2, 3], [6, 2, 1], 0].plot 
> [[1, 2, 3], 0, [6, 2, 1]].plot 
> [0, [1, 2, 3], [6, 2, 1]].plot 
> ```

These all work now – didn't work before. It doesn't matter where the single value is.

> And multiple single-value channels?
> 
> ```
> [0, [1, 2, 3], 0, [6, 2, 1]].plot 
> ```

Also works now, didn't work before.

There are cases that still will throw an error, as before.

```
[[1, [2, 3]], [6, 2, 1]].plot 
```

This PR doesn't attempt to make sense of all of that …

What do you think?

mtmccrea:
Great to know, thanks for testing!

One more request... would you mind also confirming that it works when `superpose` is toggled on/off?

The code looks good to me.

telephon:
Screenshot 2025-02-06 at 21 32 42

```supercollider
(
a = [[0, 1, 2, 3, 4, 5, 6, 7], 4,  [8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16], 8].plot; // this works now, instead of throwing an error
a.superpose_(true);
)
```

Yes, works.

mtmccrea:
Thanks for posting the image, that's helpful. 
The behavior of extending subarrays to the length of the longest subarray is odd to me. After running your change locally I see the odd  is a result of `Plotter:prReshape`, not anything in your changes. 

I'd consider this case of superposing a dataset with non-uniform sizes an unsupported case because it appears to generate new data points that aren't there. Or in any case could use further thought on how to display this case.

I don't think it should block your change because the case of "discrete" plots is fixed now. So the above case could either not be allowed (errors out), or could post a warning that it's unsupported. Both cases would allow it to be supported "properly" in the future. Note that `Plotter:domain_` does something similar — 
```
ERROR: [Plotter:-domain_] Setting the domain values isn't supported for multichannel data of different sizes...
```

What do you think?

mtmccrea:
Another option would be to just put a comment in the code acknowledging that the superpose case doesn't properly display non-rectangular data shapes, so that if someone wanted to address this in the future, they wouldn't be prevented by it being a "breaking" change.

dyfer:
@mtmccrea could you open a new issue to describe the concern regarding the "case of superposing a dataset with non-uniform sizes"?

Otherwise, if you think this PR is good to go, feel free to approve, thanks!

mtmccrea:
Thanks for having a look at this Marcin! @dyfer

I'd just like to hear from @telephon on the issue that superposing a dataset with non-uniform sizes appears to generate new data points that aren't there. 

I proposed ways to address it without blocking the current fix: 

In response to superposing a dataset with different sub-array sizes:
- could either not be allowed (errors out), or
- post a warning that it's not unsupported, while still displaying the current behavior or 
- just put a comment in the code mentioning that the superpose case doesn't properly display non-rectangular data shapes
 
That way, if someone wanted to address this in the future, they wouldn't be prevented by it being a "breaking" change.

telephon:
> I'd just like to hear from @telephon on the issue that superposing a dataset with non-uniform sizes appears to generate new data points that aren't there.

Ah yes, that is not very nice. I would either want to make it work or just not allow superpose on unequal subarrays. Then it can still be supported later. 

In principle, we clould allow `\lines` which is the same as `\linear`.  On the other hand, `\filled` does not make too much sense. So one way could be is to force the display on `\lines` when the subarrays are unequal and document that.

Would you prefer a warning?

mtmccrea:
Forcing `line` mode for just this case would add complexity, e.g. when un-superposing and having to keep track of the original plot mode.

>I would either want to make it work or just not allow superpose on unequal subarrays. Then it can still be supported later.

I agree. In hindsight, a warning may just annoy the user, so might as well just add support for superpose later in a fully functioning state.

telephon:
Hm, I just noticed that in the case of 3-dim arrays, the problem already occurs:

```
[(0..10)%4, (0..50)%32, [(0..71)%16, (0..71)%23].flop].plot
```

And there, superpose throws an error even.
So I think the current PR is not ready, unfortunately, we should close it. 

The better way to do it would be a rewrite of the `drawData` function so that it works for different sizes.

mtmccrea:
I think that adding your change is still useful. 

Checking that all features are supported for 3D arrays could be a separate project IMO, especially if it's a much less common use case.

telephon:
Ok, yes, that is true. I've simply reverted the one commit to keep the check within 2 dimensions.

mtmccrea:
Great, thanks!

So I think the last outstanding issue is just to catch the case of trying to superpose non-rectangular data (i.e. it's not supported currently)

JordanHendersonMusic:
This PR seems to break env plotting.
```
Env().plot
```
```
WARNING: Cannot currently plot an array with more than 2 dimensions
ERROR: Message 'domainSpecs_' not understood.
RECEIVER:
   nil
ARGS:
Instance of Array {    (0x561ad7075d48, gc=EC, fmt=01, flg=00, set=02)
  indexed slots [1]
      0 : instance of ControlSpec (0x561ad5f1f418, size=9, set=4)
}
KEYWORD ARGUMENTS:
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg this = nil
		arg selector = 'domainSpecs_'
		arg args = [*1]
		arg kwargs = [*0]
	Env:plot
		arg this = <instance of Env>
		arg size = 400
		arg bounds = nil
		arg minval = nil
		arg maxval = nil
		arg name = nil
		arg parent = nil
		var plotLabel = "Envelope"
		var plotter = nil
		var duration = [*1]
		var channelCount = 1
		var totalDuration = [*1]
		var res = nil
		var code = "Env().plot"
		var doc = nil
	Process:interpretPrintCmdLine
		arg this = <instance of Main>
^^ ERROR: Message 'domainSpecs_' not understood.
RECEIVER: nil


```

mtmccrea:
Thanks for catching this @JordanHendersonMusic 

The data Env is trying to plot is`[this.asMultichannelSignal(size).flop]`, see:
```supercollider
+ Env {
	plot { |size = 400, bounds, minval, maxval, name|
		var plotLabel = if (name.isNil) { "Envelope" } { name };
		var plotter = [this.asMultichannelSignal(size).flop]
		.plot(name, bounds, minval: minval, maxval: maxval);
```
@telephon,
Is the fix readily apparent to you or would it be best to revert these changes until investigated further?

telephon:
I wonder why this was necessary. Is it possible to plot Envelopes that are more than 2D ?

It should be fine to write: `var plotter = this.asMultichannelSignal(size)`

PR #6568, opened 2024-12-11
SCLang: implement superPerformArgs fixing inf. loop

### LINKED ISSUE #6567: Scale.asdf crashes interpreter (infinite loop)
Running `Scale.asdf`, or anything that triggers `Scale.doesNotUnderstand` causes an infinite loop.
To me it looks like super.performArgs doesn't actually call the superclass. I fixed it on my system by using this.superPerformList.
I don't know if it's even a bug with `super`.
Could it be related to #6339?

## Environment
* SuperCollider version: 3.14dev
* Operating system: linux (should crash others too)

## Steps to reproduce
```supercollider
Scale.asdf
```

Expected: `DoesNotUnderstandError`
Actual: `Scale.doesNotUnderstand` calls itself in an infinite loop, making the interpreter unusable

JordanHendersonMusic:
## Purpose and Motivation
`super.performArgs` doesn't work as expected, so this is needed.

Fixes #6567 

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Collections/Dictionary.sc
+++ b/SCClassLibrary/Common/Collections/Dictionary.sc
@@ -541,7 +541,7 @@ IdentityDictionary : Dictionary {
 
 	doesNotUnderstand { |selector... args, kwargs|
 		if (know.not) {
-            ^super.performArgs(\doesNotUnderstand, args, kwargs)
+            ^this.superPerformArgs(\doesNotUnderstand, args, kwargs)
 		};
         this[selector] !? { |func|
 			^func.performArgs(\functionPerformList, [\value, this, args], kwargs);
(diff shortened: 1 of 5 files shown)


telephon (approved):

telephon:
I would propose to merge now.

But please let's make a separate PR for documentation – also we need to check if there are other places that need amendment. For example, `Function` help should have an example to show how to use the `|...args, kwars| ` syntax.

JordanHendersonMusic:
I agree with the need for documentation. I might make a new document, something like, 'on writing wrapper classes/methods', as having a place to just detail keywords seems like it would be hard to access and link together with other documents.

On a different note, I've noticed the documentation layout of the kwargs in signatures odd, along with the popup in scide.

telephon:
> On a different note, I've noticed the documentation layout of the kwargs in signatures odd, along with the popup in scide.

I don't find it, do you have a place where this occurs?

> I might make a new document, something like, 'on writing wrapper classes/methods',

Yes, that may be good, under Tutorials. I believe we should also have examples of ellipsis and kwargs in `Function`, right under `Description`.

JordanHendersonMusic:
In my opinion, this is real to be merged as it's an important bug fix.

Issue #6508, opened 2024-10-21
can't add items to TreeView


sadguitarius:
## Environment

* SuperCollider version: 3.14.0 dev
* Operating system: Windows 11
* Other details (Qt version, audio driver, etc.): Qt 6.8

## Steps to reproduce

These commands both fail. I realized this by trying to open Quarks.gui in 3.14

```supercollider

(
t = TreeView().front;
t.addItem(["hi"]);
)

(
t = TreeView().front;
t.insertItem(0, ["hi"]);
)

```

Here's the error:

```
ERROR: Primitive '_QObject_InvokeMethod' failed.
Failed.
RECEIVER:
Instance of TreeView {    (0000026FC8558EF8, gc=64, fmt=00, flg=00, set=06)
  instance variables [36]
    qObject : RawPointer 0000026F83004870
    finalizer : instance of Finalizer (0000026FC86202C8, size=2, set=1)
    virtualSlots : nil
    wasRemoved : false
    font : nil
    resize : Integer 1
    alpha : Float 1.000000   00000000 3FF00000
    decorator : nil
    layout : nil
    userCanClose : true
    deleteOnClose : true
    action : nil
    mouseDownAction : nil
    mouseUpAction : nil
    mouseEnterAction : nil
    mouseLeaveAction : nil
    mouseMoveAction : nil
    mouseOverAction : nil
    mouseWheelAction : nil
    keyDownAction : nil
    keyUpAction : nil
    keyModifiersChangedAction : nil
    keyTyped : nil
    focusGainedAction : nil
    focusLostAction : nil
    dragLabel : nil
    beginDragAction : nil
    canReceiveDragHandler : nil
    receiveDragHandler : nil
    toFrontAction : nil
    endFrontAction : nil
    onClose : nil
    onResize : nil
    onMove : nil
    itemPressedAction : nil
    onItemChanged : nil
}
CALL STACK:
	MethodError:reportError
		arg this = <instance of PrimitiveFailedError>
	Nil:handleError
		arg this = nil
		arg error = <instance of PrimitiveFailedError>
	Thread:handleError
		arg this = <instance of Thread>
		arg error = <instance of PrimitiveFailedError>
	Object:throw
		arg this = <instance of PrimitiveFailedError>
	Object:primitiveFailed
		arg this = <instance of TreeView>
	TreeView:addItem
		arg this = <instance of TreeView>
		arg strings = [*1]
	Interpreter:interpretPrintCmdLine
[...]

PR #6521, opened 2024-10-30
QtCollider: QcTreeWidget: fix instance methods

sadguitarius:
## Purpose and Motivation

Fixes #6508. My current solution is to hard-code the types that are passed to Qt macros. The ItemPtr typedef broke things because Qt was relying on string comparisons to determine type equivalence. Maybe someone has a better idea for how to handle this though.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation
- [x] This PR is ready for review

PR #6570, opened 2024-12-12
[qt5] QMetaType fix for Qt5

dyfer:
## Purpose and Motivation

In the process of reviewing https://github.com/supercollider/supercollider/pull/6521 it turned out that the issue in https://github.com/supercollider/supercollider/issues/6508 was already present with Qt 5.15 after adding https://github.com/supercollider/supercollider/commit/f70364c704ea18f00e5aa10aa390aa1ef1b8fdb6. This PR brings back old behavior at the cost of four extra ifdefs in the codebase, reverting that change for Qt5. See discussion in https://github.com/supercollider/supercollider/pull/6521#issuecomment-2539351954.

Thanks @xunil-cloud for the tip.

There might be a better way to address this issue, but I have no expertise to pursue that. I think the ifdef is a quick and easy fix, and will be removed when we remove Qt5 support altogether.

Fixes https://github.com/supercollider/supercollider/issues/6508 for Qt 5.15 (macOS legacy build).

Code that was previously failing for reference:
```supercollider
(
t = TreeView().front;
t.addItem(["hi"]);
)
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- n/a Updated documentation

### DIFF
--- a/QtCollider/QObjectProxy.cpp
+++ b/QtCollider/QObjectProxy.cpp
@@ -87,7 +87,11 @@ static bool serializeSignature(QVarLengthArray<char, 512>& dst, const char* meth
     int i;
     for (i = 0; i < argc; ++i) {
         int typeId = argv[i].type()->id();
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
+        const char* typeName = QMetaType::typeName(typeId);
+#else
         const char* typeName = QMetaType(typeId).name();
+#endif
         int len = qstrlen(typeName);
         if (len <= 0) {
             qcErrorMsg("Could not get argument type name.");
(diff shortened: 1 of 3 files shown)


capital-G (approved):
I think this is fine.

If this would be more permanent we could declare ~~a new type alias via `using` for this~~ the fix in a `compat.cpp`.
But as all introduced changes are retrievable and revertable by searching for `if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))` it seems fine as is.

Thanks for taking care of this! Maybe wait a week for an additional review and if no one else gives this a review/remark we can merge this?

PR #6571, opened 2024-12-13
classlib: JITLib: add possitiblity of partially mapping node proxies

bgola:
## Purpose and Motivation

Introduces `NodeProxyPartMap`, an object to allow mapping a subset of the outputs of a proxy. It also adds a convenience method `partMap` to `NodeProxy`.

This was implemented a few years ago in conversation with @telephon but for some reason we never proposed a PR upstream. 

## Types of changes

- New feature

## To-do list

- [X] Code is tested
- [ ] All tests are passing
- [X] Updated documentation


SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
```diff
@@ -173,12 +174,58 @@ ProxyNodeMap : NodeMap {
...
 + BusPlug {
 	asOSCArgEmbeddedArray { |array|
 		^this.asControlInput.asOSCArgEmbeddedArray(array)
 	}
 }
 
+
 + Object {
 	wakeUpToBundle {}
+	proxyNeedsWakeUp { ^false }
```
capital-G, line 230: Could there be another way of doing this instead of introducing a new member to Object?

capital-G (commented):
Thanks for this, looks and feels like a really useful addition.

I asked myself why you didn't consider adding this as argument to `Ndef.ar/kr`, e.g. something like

```supercollider
Ndef {
	*ar { | key, numChannels, offset = 0, channelArray = nil |
                // nil defaults to all channels
		^this.new(key).ar(numChannels, offset, channelArray)
	}
}
```

I only have minor comments, once they have been resolved this looks fine to me. I haven't looked through all tests, but good that they exist! :)

SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
```diff
@@ -173,12 +174,58 @@ ProxyNodeMap : NodeMap {
...
 }
 
+
+// represents a subset of the output bus of a proxy
+// that can be mapped to a control.
+
+NodeProxyPartMap {
+	var <proxy, <indices;
+
+	*new { |proxy, indices=0|
```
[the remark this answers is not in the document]
bgola: mmmh i am not sure, maybe it can be one index or an array of indices? do you think it is too confusing to allow for both options?

SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
bgola: maybe we can ask if the object respondsTo(\proxyNeedsWakeUp) and then only implement this method in NodeProxy and NodeProxyPartMap?

bgola: i mean, the previous code was 

`if (item.isKindOf(NodeProxy))`

now it would be 

`if (item.respondsTo(\proxyNeedsWakeUp) and: {item.proxyNeedsWakeUp})`

SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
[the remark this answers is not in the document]
capital-G: I don't think it's confusing to allow both, indices and index, as it is quite common in sclang. But the question is what a default would be. Currently it seems to be the first channel, but maybe all channels are a better default (via `null` which gets replaced with a full array)? But first channel could also be a good default.

SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
```diff
@@ -93,10 +93,10 @@ ProxyNodeMap : NodeMap {
 
 	put { |key, item|
 		if(proxy.notNil) {
-			if(this.at(key).isKindOf(NodeProxy)) {
+			if(this.at(key).proxyNeedsWakeUp) {
```
telephon, line 96: The method name `proxyNeedsWakeUp` is too specific. We add children for at least two different reasons: waking up parents, but also rebuilding when the channels have changed. And also `proxyNeedsWakeUp` reads a bit as a question "does this proxy need wake up?" – which is misleading, because it is true for every proxy and only false for other objects.

Let's see what is the best solution, below.

SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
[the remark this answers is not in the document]
telephon: I think it would be better without a default. The case of the first channel is just as special as any other case for this class.

Indeed, all channels are a better default.

SCClassLibrary/JITLib/ProxySpace/NodeMap.sc
telephon: that'a also not the best way (see above).

Btw. I wonder if `wakeUpToBundle` needs to be implemented in `Object`. But certainly, one could ask `respondsTo(\wakeUpToBundle)` – I just find that not such a good way of conveying intent yet.

telephon:
I suggest another naming because it conveys intent better: children are not only there for parents to wake them up.
One way to do this would be to add an interface to `NodeProxy` and `NodeProxyPartMap`, or whoever is concerned, which is:

```supercollider

addChildNodeProxy {  |proxy|
     children = children.add(proxy);
}

removeChildNodeProxy {  |proxy|
     children.remove(proxy);
}
```

Then, we can either ask in `ProxyNodeMap` 
```
item.respondsTo(\addChildNodeProxy)
```

Or we can add this method to Object to do nothing.

What do you think?

PR #6575, opened 2024-12-19
QtCollider: Fix touchpad mousewheel event scaling

jpburstrom:
## Purpose and Motivation

Delta values in View.mouseWheelEvent are truncated, mostly producing zeroes, on my macbook trackpad. This is due to the delta integer being multiplied with 0.25 (apparently for historical reasons), truncating all small changes to zero. By casting the delta value to QPointF _before_ scaling, delta should no longer be truncated.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested (manually)
- [ ] All tests are passing


dyfer (approved):
Thanks!

Issue #6386, opened 2024-07-06
Docs: Document `(a: 10)` vs. `(a : 10)`


JordanHendersonMusic:
## Motivation
The following behaviour with the colon placement is not documented.
```
a = [];
m1 = (a : 10); <--- space
m2 = (a: 10); <--- no space

m1[a] // 10
m1.a // nil
m1[\a] // nil

m2[a] // nil
m2.a // 10
m2[\a] // 10

```

## Description of Proposed Feature

Add docs.

## Plan for Implementation

A good first issue.

nuss:
I'll add a note about the syntax in the `Event` help file when I find time. Just a bit busy with other things right now...

But in general users must be careful when providing keys with colons (as I just found out):
```supercollider
f = { |a = 2, b = 3, c = 4| a + b + c };

f.(b: 6, c: 5) // --> 13
f.(b : 6, c: 5) //  syntax error

ERROR: syntax error, unexpected ':', expecting ')'
  in interpreted text
  line 1 char 6:

  f.(b : 7, c: 1) 
       ^
```
... at least it's a useful error message but should this additionally be mentioned somewhere?  At least in SCIDE code-colorizing (as well as here in markdown) gives you a hint. However, I'm not sure where in the help files some info on this could go...

JordanHendersonMusic:
> at least it's a useful error message

It could be better though. The parser error messages are quite hard to unpick, so off the top of my head I don't know if it's easy to change. At least it gets the location correct.

> However, I'm not sure where in the help files some info on this could go...

This is a good point. 
Perhaps it would be a good idea to have a dedicated working with keyword arguments page as a part of https://github.com/supercollider/supercollider/pull/6339?

nuss:
> Perhaps it would be a good idea to have a dedicated working with keyword arguments page as a part of https://github.com/supercollider/supercollider/pull/6339?

Ah, yes, sounds like a good idea!

JordanHendersonMusic:
This was completed.

PR #6581, opened 2024-12-24
Add section 'Syntax subtleties' to Event helpfile

nuss:
## Purpose and Motivation

This is a pull request in response to this discussion: #6386.

## Types of changes

I've added a subsection "Syntax subtleties" right below "Event as a name space for keeping objects" which in turn is the first paragraph under "Basic Usage" - seemed like the best place. Let me know if it should go somewhere else.

- Documentation

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Event.schelp
+++ b/HelpSource/Classes/Event.schelp
@@ -240,6 +240,38 @@ q.y = q.x * q.n;	// shorter way to do the same (pseudo-methods)
 q.y;			// [100, 200, 300]
 ::
 
+subsection::Syntax subtleties
+
+Events have a very special meaning in a musical context within SuperCollider, especially when it comes to sequencing with Patterns. Nevertheless, as the class inherits from link::Classes/IdentityDictionary::, it is often used as a convenient-to-type replacement for (Identity)Dictionary-like objects:
+
+code::
+e = (a: 5, b: 7, c: 9);
+::
+
+As one can see the syntax very much reminds of the declaration of keyword arguments in function or method calls. However, there is a subtle difference:
+
+code::
+e = (a: 10); // -> ('a': 10)
+f = (a : 10); // -> nil
+              // ERROR: Primitive '_IdentDict_Put' failed.
+::
+
+Though the above declartion of code::f:: failed it's strong::not:: a syntax error:
+
+code::
+a = "abracadabra";
+f = (a : 10); // -> ("abracadabra": 10)
+::
+
+A declaration like code::(a: 5, c: 7):: constitutes an Event whose link::Classes/Dictionary#-keys#keys:: can be queried as a link::Classes/Set:: of link::Classes/Symbol::s and its link::Classes/Dictionary#-values#values:: as a link::Classes/List:: of values. Adding a space after a key and before the colon, however, breaks that rule in a way that sclang now regards the key as a emphasis::variable name::. Nevertheless, the same rules apply here as for all other Events or IdentityDictionaries: Values emphasis::must:: must be queried under a key emphasis::identical:: to the one it has been declared with:
+
+code::
+f["abracadabra"]; // -> nil - "abracadabra" is equal to the string stored in a, not identical
+f[a];             // -> 10
+::
+
+This may not be common practice but possibly worth mentioning and in some cases even come handy.
+
 subsection::Event for specifying different things to happen
 
 Event provides a link::#defaultParentEvent:: that defines a variety of different event types and provides a complete set of default key/value pairs for each type. The type is determined by the value of the key strong::\type:: which defaults to strong::\note::. Note events create synths on the server.
--- a/HelpSource/Classes/Event.schelp
+++ b/HelpSource/Classes/Event.schelp
@@ -246,6 +246,7 @@ Events have a very special meaning in a musical context within SuperCollider, es
 
 code::
 e = (a: 10);  // -> ('a': 10)
+a = nil;
 f = (a : 10); // -> nil
               // ERROR: Primitive '_IdentDict_Put' failed.
               // a is treated as a variable!
--- a/HelpSource/Classes/Event.schelp
+++ b/HelpSource/Classes/Event.schelp
@@ -254,8 +254,6 @@ f = (a : 10); // -> ("abc": 10)
 f["abc"];     // -> nil
               // "abc" is equal to the string stored in a but not identical
 f[a]:         // -> 10
-f.a;          // -> nil
-              // a is a variable, not a Symbol
 ::
 
 subsection::Event for specifying different things to happen
--- a/HelpSource/Classes/Event.schelp
+++ b/HelpSource/Classes/Event.schelp
@@ -4,7 +4,7 @@ related::Classes/Pattern, Classes/Environment, Classes/IdentityDictionary
 categories::Collections>Unordered, Streams-Patterns-Events>Events
 
 DESCRIPTION::
-An Event specifies an action to be taken in response to a link::#-play:: message. Its key/value pairs specify the parameters of that action. Event inherits most of its methods from its superclasses, especially from link::Classes/Dictionary::. For the usage and meaning of the code::parent:: and code::proto:: events, see link::Classes/IdentityDictionary::.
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Thanks for pointing out this detail, I was not aware of it. To me this feels more like a syntax which confuses more than it helps, but it is good that it is finally documented.

Another person should review this and then it is good for merging :)

HelpSource/Classes/Event.schelp
```diff
@@ -240,6 +240,38 @@ q.y = q.x * q.n;	// shorter way to do the same (pseudo-methods)
...
+a = "abracadabra";
+f = (a : 10); // -> ("abracadabra": 10)
+::
+
+A declaration like code::(a: 5, c: 7):: constitutes an Event whose link::Classes/Dictionary#-keys#keys:: can be queried as a link::Classes/Set:: of link::Classes/Symbol::s and its link::Classes/Dictionary#-values#values:: as a link::Classes/List:: of values. Adding a space after a key and before the colon, however, breaks that rule in a way that sclang now regards the key as a emphasis::variable name::. Nevertheless, the same rules apply here as for all other Events or IdentityDictionaries: Values emphasis::must:: must be queried under a key emphasis::identical:: to the one it has been declared with:
+
+code::
+f["abracadabra"]; // -> nil - "abracadabra" is equal to the string stored in a, not identical
+f[a];             // -> 10
+::
```
dyfer, line 271: Thanks for the PR! The space before the colon behavior is a details I wasn't aware of...

I have one point here: is it worth using a string here, instead of a symbol? This section now talks about two unrelated matters: `a` being a key vs a variable name, and then keys' identity, which is covered e.g. in the IdentityDictionary help, but apparently not in Event help. In any case, I think the point about the identity would be better mentioned separately from the "Syntax subtleties" section.

In particular, using a string assigned to a variable seems such an edge case... I think it'd be better to point out the identity requirement (separately from the "Syntax subtleties" and instruct users to use a Symbol instead of a String.

nuss: > is it worth using a string here, instead of a symbol? This section now talks about two unrelated matters: a being a key vs a variable name, and then keys' identity, which is covered e.g. in the IdentityDictionary help, but apparently not in Event help. 

I think using a string here makes sense though I didn't have in mind the difference between strings and symbols in particular - I wanted to line out that keys can be _any_ object. However, there there's this special result being posted in the post window when you create an Event (different from what you get when creating an IdentityDictionary):
```supercollider
a = "abracadabra";
e = (a : 10); // -> ("abracadabra": 10)
e["abracadabra"]; // -> nil

// and of course not...
e.abracadabra

b = \abracadabra;
f = (b : 10); // -> ('abracadabra': 10)
f[\abracadabra]; // -> 10

// and also, dot notation...
f.abracadabra // -> 10
```
... so, using a Symbol in the example suggests to the user that everything's fine. It's just a bit of a quirky way of  creating an Event. Yet what I'm really missing in the Event helpfile is an explicit mention that e.g. `(a: 20)` really creates a key `a` which **is** a Symbol (will everyone who comes along the Event helpfile have read the IdentityDictionary help first? Will everyone who's reading the Event helpfile understand that really an Event is a special kind of IdentityDictionary?). 

So, long story short: A result `("abracadabra": 10)` being posted might obfuscate the real nature of Events. One might even be tempted to create an Event like `e = ("abracadabra": 10)` (which is possible) and be surprised when trying to call the value under the given key. I meant this code more like a practical "heads up" example, omitting a lot of explanation, yet pointing out a problematic aspect. A sentence like "Event inherits most of its methods from its superclasses, especially from Dictionary." (third sentence in the leading description) doesn't really make clear why `e = ("abracadabra": 10)` won't yield the desired result ("I can have strings as keys in a Dictionary. Why the hell doesn't that work in an Event?") .  Maybe it would be good to explicitly mention the inheritance of Event from IdentityDictionary and that keys in Events are (or should be) Symbols.

Dunno...

dyfer: Thinking again about this, the point about the keys can be added separately from this PR.
One suggestion I'd have is to include something that succinctly describes the syntax subtlety inline, like:
```supercollider
e = (a: 10); // a is interpreted as symbol 'a'
f = (a : 10); // a is interpreted as variable name; a is nil in the example above and that's it fails
```

dyfer: @nuss what do you think about the above suggestion?\

nuss: Sorry, I've had so many other things going on... your suggestion is concise and clear (at least for me). Nevertheless I would like to emphasize a bit more that `Event` isn't simply another kind of `Dictionary` but an `IdentityDictionary`. For that reason I'd maybe amend your examples a little bit:
```
e = (\a : 5) // valid, equivalent to e = (a: 5)
f = ("abcd": 5) // valid but not really useful - value can only be queried under a key *identical* to the one that it has been stored with

... I don't really know.

Comment (dyfer) [reply HelpSource/Classes/Event.schelp:271]: I like this suggestion! But I guess the previous example highlights the nuance of a symbol vs a variable name, which I wanted to clarify with the comment.
Separately, I absolutely agree that the comment about a string as a key is helpful... but orthogonal to the issue of symbol vs varname.

Comment (nuss) [reply HelpSource/Classes/Event.schelp]: Ok, I've updated my original pull request:

- added a sentence about Event's inheritance  from IdentityDictionary to the helpfile's introduction. Though the introduction links to IdentityDictionary for more explanations on proto and parent Events it's not exactly obvious that Event inherits from IdentityDictionary.
- condensed the subsection 'Syntax subtleties' a bit but extended the example of an Event holding a value under a variable key a bit.

File: HelpSource/Classes/Event.schelp [hunk 2]
```diff
@@ -240,6 +240,24 @@ q.y = q.x * q.n;	// shorter way to do the same (pseudo-methods)
...
+f = (a : 10); // -> nil
+              // ERROR: Primitive '_IdentDict_Put' failed.
+              // a is treated as a variable!
+a = "abc";
+f = (a : 10); // -> ("abc": 10)
+f["abc"];     // -> nil
+              // "abc" is equal to the string stored in a but not identical
+f[a]:         // -> 10
+f.a;          // -> nil
+              // a is a variable, not a Symbol
```
JordanHendersonMusic, line 258: I think this line is a little confusing. `a` is a variable - so won't work as a fake method -, but if it was also a `Symbol` (`a=\b`) this still wouldn't work. I think you mean something closer to a symbol--literal, but that probably over complicates things.

nuss: Probably you're right. I wouldn't bother omitting the last two lines.

dyfer:
> Don't know how to revert that :\

Rebase! First update your `develop` from upstream, then rebase onto that branch, possibly using the `-i` switch to review the list of commits and adjust as needed. [Our wiki has some information on rebasing](https://github.com/supercollider/supercollider/wiki/Rebasing-a-PR).

HelpSource/Classes/Event.schelp
```diff
@@ -240,6 +240,22 @@ q.y = q.x * q.n;	// shorter way to do the same (pseudo-methods)
...
 
+subsection::Syntax subtleties
+
+Events have a very special meaning in a musical context within SuperCollider, especially when it comes to sequencing with Patterns. Nevertheless, as the class inherits from link::Classes/IdentityDictionary::, it is often used as a convenient-to-type replacement for (Identity)Dictionary-like objects. The syntax very much reminds of the declaration of keyword arguments in function or method calls. However, there is a subtle difference:
+
+code::
+e = (a: 10);  // -> ('a': 10)
+f = (a : 10); // -> nil
+              // ERROR: Primitive '_IdentDict_Put' failed.
+              // a is treated as a variable!
```
dyfer, line 251: Can you add `a = nil; // make sure a is nil` beforehand just to make sure it actually is not assigned to anything? There's a= assignment earlier in this help doc.

nuss: Yes, will do tomorrow. It's already quite late and I need a little sleep.

PR #6583, opened 2025-01-03
JITLib: TaskProxyGui: improve src/env editing

unthingable:
- always focus the new editor
- include proxy key in doc name instead of just "edit me"
- do not prompt to save on close
- fix doc reference, was broken

## Purpose and Motivation

It is nice to edit src/env contents directly from TaskProxyGui but was annoying to work with it. The newly opened editor was not in focus and closing it triggered a save dialog needlessly.

Also, upon a closer look the code was expecting to receive a doc reference from `String.newTextWindow` which actually returns the string; so the call to set the new document bounds would have never worked. Note that there are other instances  of `newTextWindow` usage with this exact problem, this PR does not address them.

## Types of changes

- Bug fix/improvement

A fairly minimal change to 12 year old code, opening a document directly instead of via `String.newTextWindow` so the latter remains unchanged.

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/JITLib/GUI/TaskProxyGui.sc
+++ b/SCClassLibrary/JITLib/GUI/TaskProxyGui.sc
@@ -240,8 +240,12 @@ TaskProxyGui : JITGui {
 	}
 
 	openDoc { |strings, bounds|
-		var doc = strings.join.newTextWindow("edit me");
-		try { doc.bounds_(bounds ? Rect(0, 400, 400, 200)) };
+        var doc = Document.new("edit" + super.getName, strings.join, false);
+		try {
+            doc.front;
+            doc.promptToSave = false;
+            doc.bounds_(bounds ? Rect(0, 400, 400, 200))
+        };
 	}
 
 }
--- a/SCClassLibrary/JITLib/GUI/TaskProxyGui.sc
+++ b/SCClassLibrary/JITLib/GUI/TaskProxyGui.sc
@@ -244,7 +244,7 @@ TaskProxyGui : JITGui {
 		try {
             doc.front;
             doc.promptToSave = false;
-            doc.bounds_(bounds ? Rect(0, 400, 400, 200))
+            doc.bounds_(bounds ?? { Rect(0, 400, 400, 200) })
         };
 	}
 


SCClassLibrary/JITLib/GUI/TaskProxyGui.sc
```diff
@@ -240,8 +240,12 @@ TaskProxyGui : JITGui {
 	}
 
 	openDoc { |strings, bounds|
-		var doc = strings.join.newTextWindow("edit me");
-		try { doc.bounds_(bounds ? Rect(0, 400, 400, 200)) };
+        var doc = Document.new("edit" + super.getName, strings.join, false);
+		try {
+            doc.front;
+            doc.promptToSave = false;
+            doc.bounds_(bounds ? Rect(0, 400, 400, 200))
```
telephon, line 247: While you are at it, could you please write it in a more optimized way:
```
doc.bounds_(bounds ?? { Rect(0, 400, 400, 200) })
```

telephon: Otherwise, great, thank you!

unthingable: Good call, thank you!

telephon (approved):
thanks!

capital-G:
Just FYI - I think this PR is not formatted properly. Strangely this looks OK in the PR diff view

grafik

But looking at the merged commit https://github.com/supercollider/supercollider/commit/4c09d626c48f7da300709e1574b759ec499fba1d, it looks off

grafik

And vim tells me that there are whitespaces instead of tabs

grafik

I think sclang formatting is not that strict, but as we settled for tabs we should stick to it - just for a future PR :)

Maybe we could add something to the CI to detect things like this - but as we don't have an easy way of parsing sclang this seems not really reliable :/

unthingable:
Yeah, I kind of noticed the same thing after the fact. Strange on github's part indeed.

> but as we don't have an easy way of parsing sclang this seems not really reliable :/

Nothing about indenting with tabs is easy or reliable :)

I did miss the [styleguide](https://github.com/supercollider/supercollider/wiki/Style-Guidelines:-SuperCollider#rule-use-tabs-for-indentation), that's on me — though it was a little nontrivial to find, not linked in main README or Contributing Guide.

Should I make the followup PR with the fix?

It would be pretty easy to detect and disallow mixed indentation during CI: after squashing multiline comments, all first characters that are whitespace must be either tab or space (or use `tabnanny` python module). Could also be a handy feature to detect this in ScIDE and show a warning, though not as easy as a CI step.

telephon:
@unthingable can you just select that region in the code and hit `tab`- it should remove the whitespaces. Then you just male another PR.  Thank you!

PR #6584, opened 2025-01-04
class library: keyword args for composed and higher order functions

telephon:
## Purpose and Motivation

```
// We would like to be able to write things like this:
f = { |x, y = 1| x + y } * { |x, y = 2| x ** y };
f.(x:2, y:6) 

// or this
f = { |x, y = 1| x + y };
g = f.linexp(0, 255, 300, 1000);
g.(x: 8);
```

This has a little bit of performance impact, so let's discuss.
@JordanHendersonMusic what do you think?

## Types of changes

- New feature

To be done:
- [x] UnaryOpFunction
- [x] BinaryOpFunction
- [x] NaryOpFunction
- [x] FunctionList
- [x] UnaryOpFunctionProxy
- [x] BinaryOpFunctionProxy
- [x] NAryOpFunctionProxy
- [x] Maybe
- [x] CallOnce

Not yet, because it is really core:
- Message
- MethodQuote

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/FunctionList.schelp
+++ b/HelpSource/Classes/FunctionList.schelp
@@ -61,7 +61,7 @@ code::
 // example
 
 a = nil;
-a = a.addFunc { |x = "", y = ""| "this % is an % example\n".postf(x, y); 1 };
+a = a.addFunc { |x = "", y = ""| "this % is an % example\n".postf(x, y); 1 };
 a.postln;
 a = a.addFunc { |x = "", y = ""| "there is no % that is %\n".postf(x, y); 2 };
 a.value;
@@ -110,6 +110,7 @@ a.value
 a = a.flop;
 a.value
 a.value([-1, 1])
+a.value(x:[-1, 1]); // kwargs
 ::
 
 code::
--- a/SCClassLibrary/JITLib/Patterns/Fdef.sc
+++ b/SCClassLibrary/JITLib/Patterns/Fdef.sc
@@ -78,8 +78,8 @@ Maybe : Ref {
 		this.source.do(function) // problem: on the fly change is not picked up in this case.
 	}
 
-	doesNotUnderstand { |selector ... args|
-		^this.composeNAryOp(selector, args)
+	doesNotUnderstand { |selector ... args, kwargs|
+		^this.composeNAryOp(selector, args, kwargs)
 	}
 
 	// streams and patterns
@@ -108,8 +108,8 @@ Maybe : Ref {
 	reverseComposeBinaryOp { |aSelector, something, adverb|
 		^BinaryOpFunctionProxy.new(aSelector, something, this, adverb);
 	}
-	composeNAryOp { |aSelector, anArgList|
-		^NAryOpFunctionProxy.new(aSelector, this, anArgList)
+	composeNAryOp { |aSelector, anArgList, aKeywordArgList|
+		^NAryOpFunctionProxy.new(aSelector, this, anArgList, aKeywordArgList)
 	}
 
 
--- a/SCClassLibrary/JITLib/Patterns/Fdef.sc
+++ b/SCClassLibrary/JITLib/Patterns/Fdef.sc
@@ -7,66 +7,66 @@ Maybe : Ref {
 	classvar <>protected = false, <>verbose = false;
 
 	source { ^value }
-	source_ { arg obj;
+	source_ { |obj|
 		this.value = obj;
 		this.changed(\source, obj)
 	}
 
 	clear { value = nil }
 
-	value { arg ... args;
+	value { |... args|
 		^this.reduceFuncProxy(args)
 	}
-	valueArray { arg args;
+	valueArray { |args|
 		^this.reduceFuncProxy(args)
 	}
-	valueEnvir { arg ... args;
+	valueEnvir { |... args|
 		^this.notYetImplemented(thisMethod)
 		//^value.valueEnvir(*args) ? defaultValue
 	}
-	valueArrayEnvir {  arg ... args;
+	valueArrayEnvir { |... args|
 		^this.notYetImplemented(thisMethod)
 		//^value.valueArrayEnvir(args) ? defaultValue
 	}
-	functionPerformList { arg selector, arglist;
+	functionPerformList { |selector, arglist|
 		^this.performList(selector, arglist)
 	}
 
 	// this allows recursion
-	apply { arg ... args;
+	apply { |... args|
 		^this.reduceFuncProxy(args, false)
 	}
 	// function composition
-	o { arg ... args;
+	o { |... args|
 		^NAryValueProxy(this, args)
 	}
-	<> { arg that;
+	<> { |that|
 		^o (this, that)
 	}
 
 	// override some collection methods
 
-	at { arg ... args;
+	at { |... args|
 		^this.composeNAryOp(\at, args)
 	}
 
-	atAll { arg ... args;
+	atAll { |... args|
 		^this.composeNAryOp(\atAll, args)
 	}
 
-	put { arg ... args;
+	put { |... args|
 		^this.composeNAryOp(\put, args)
 	}
 
-	putAll { arg ... args;
+	putAll { |... args|
 		^this.composeNAryOp(\putAll, args)
 	}
 
-	add { arg ... args;
+	add { |... args|
 		^this.composeNAryOp(\add, args)
 	}
 
-	addAll { arg ... args;
+	addAll { |... args|
 		^this.composeNAryOp(\addAll, args)
 	}
 
@@ -74,18 +74,18 @@ Maybe : Ref {
 	all {
 		^this.source.all
 	}
-	do { arg function;
+	do { |function|
 		this.source.do(function) // problem: on the fly change is not picked up in this case.
 	}
 
-	doesNotUnderstand { arg selector ... args;
+	doesNotUnderstand { |selector ... args|
 		^this.composeNAryOp(selector, args)
 	}
 
 	// streams and patterns
 
-	embedInStream { arg inval;
-		^Prout { arg inval;
+	embedInStream { |inval|
+		^Prout { |inval|
 			var curVal, str;
 			var outval;
 			while {
@@ -99,23 +99,23 @@ Maybe : Ref {
 	}
 
 	// math
-	composeUnaryOp { arg aSelector;
+	composeUnaryOp { |aSelector|
 		^UnaryOpFunctionProxy.new(aSelector, this)
 	}
-	composeBinaryOp { arg aSelector, something, adverb;
+	composeBinaryOp { |aSelector, something, adverb|
 		^BinaryOpFunctionProxy.new(aSelector, this, something, adverb);
 	}
-	reverseComposeBinaryOp { arg aSelector, something, adverb;
+	reverseComposeBinaryOp { |aSelector, something, adverb|
 		^BinaryOpFunctionProxy.new(aSelector, something, this, adverb);
 	}
-	composeNAryOp { arg aSelector, anArgList;
+	composeNAryOp { |aSelector, anArgList|
 		^NAryOpFunctionProxy.new(aSelector, this, anArgList)
 	}
 
 
 	// used by AbstractFunction:reduceFuncProxy
 	// to prevent reduction of enclosed functions
-	valueFuncProxy { arg args;
+	valueFuncProxy { |args|
 		if(verbose and: { value.isNil }) {
 			("Maybe: incomplete definition: %\n").postf(this.infoString(args))
 		};
@@ -124,7 +124,7 @@ Maybe : Ref {
 		}
 	}
 
-	reduceFuncProxy { arg args, protect=true;
+	reduceFuncProxy { |args, protect=true|
 		if(verbose and: { value.isNil }) {
 			("Maybe: incomplete definition: %\n").postf(this.infoString(args))
 		};
@@ -138,7 +138,7 @@ Maybe : Ref {
 		}
 	}
 
-	catchRecursion { arg func;
+	catchRecursion { |func|
 		var val, previous;
 		try {
 			protect {
@@ -184,7 +184,7 @@ Maybe : Ref {
 		^if(res.notNil) { "~" ++ res } { this.asString }
 	}
 
-	infoString { arg args;
+	infoString { |args|
 		var who, str="", src;
 		who = this.findKey;
 		if(who.isNil) { ^this.asString };
@@ -201,7 +201,7 @@ Maybe : Ref {
 		^currentEnvironment.findKeyForValue(this)
 	}
 
-	storeOn { arg stream;
+	storeOn { |stream|
 		// maybe should try to catch a recursion here:
 		stream << this.class.name << "(" <<< value << ")" }
 
@@ -214,7 +214,7 @@ Fdef : Maybe {
 		all = IdentityDictionary.new
 	}
 
-	*new { arg key, val;
+	*new { |key, val|
 		var res;
 		res = all[key];
 		if(res.isNil) {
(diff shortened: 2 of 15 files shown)


JordanHendersonMusic:
A benchmark is probably needed to compare.... But I've found when recompiling the class library performance can differ by up to 10%, so be careful there.

I really don't want to suggest this, because there could be a huge explosion of methods... But we could implement `valueArgs` which might speed things up a little if really needed. 

Otherwise, I think this is a good addition, as keywords should work everywhere normal arguments work.
Can you think of any other places where they don't work?

telephon:
Yes, there are others that should also take keword arguments.

Here is a list, I tried to make it complete, but I may have missed something:

Message
MethodQuote
UnaryOpFunctionProxy
BinaryOpFunctionProxy
NAryOpFunctionProxy
Maybe
CallOnce

I do think that a message `valueArgs` would be good (also for readability, but...), because there may be quite a bit of overhead in the message dispatch. Roughly a 1:5 relation:

```supercollider
f = { |x| x };
bench { 10000.do { f.performArgs(\value, [5]) } }; // time to run: 0.011808750001364 seconds.
bench { 10000.do { f.valueArray(5) } }; // time to run: 0.0019109839995508 seconds.
```

Roughly half of it comes from the array construction:

```supercollider
bench { 10000.do { f.performArgs(\value, #[5]) } }; // time to run: 0.0028367470004014 seconds.
bench { 10000.do { f.valueArray([5]) } }; // time to run: 0.0057900880019588 seconds.
```
We should keep in mind of course that these are low numbers on an absolute scale, and in the methods we talk about, the arrays are already constructed (currently they call `valueArray`). But they are also core functionality.

JordanHendersonMusic:
So since the array exists the difference is between these two right?

```
bench { 10000.do { f.valueArray(5) } }; 

// time to run: 0.0019109839995508

bench { 10000.do { f.performArgs(\value, #[5]) } }; 

// time to run: 0.0028367470004014 seconds
```

So more like 1:1.5 when the array exists.

`valueArgs` would only cut out the lookup of the method, the array unpacking is still needed, which is going to be the slow bit, so I'm not sure how much impact this would have.

JordanHendersonMusic:
So since the array exists the difference is between these two right?

```
bench { 10000.do { f.valueArray(5) } }; 

// time to run: 0.0019109839995508

bench { 10000.do { f.performArgs(\value, #[5]) } }; 

// time to run: 0.0028367470004014 seconds
```

So more like 1:1.5 when the array exists.

`valueArgs` would only cut out the lookup of the method, the array unpacking is still needed, which is going to be the slow bit, so I'm not sure how much impact this would have.

telephon:
> So more like 1:1.5 when the array exists.

yes.

> `valueArgs` would only cut out the lookup of the method, the array unpacking is still needed, which is going to be the slow bit, so I'm not sure how much impact this would have.

Yes, but in the cases at hand, we have the array anyhow.

Compare:

```supercollider
value { | ... args, kwargs |
		^a.performArgs(\value, args, kwargs).perform(selector)
	}
	value1 { | ... args |
		^a.valueArray(args).perform(selector)
	}
```

telephon:
> Yes, there are others that should also take keword arguments.
> 
> Here is a list, I tried to make it complete, but I may have missed something:
> 
> Message MethodQuote UnaryOpFunctionProxy BinaryOpFunctionProxy NAryOpFunctionProxy Maybe CallOnce
> 
> I do think that a message `valueArgs` would be good (also for readability, but...), because there may be quite a bit of overhead in the message dispatch. Roughly a 1:5 relation:
> 
> ```supercollider
> f = { |x| x };
> bench { 10000.do { f.performArgs(\value, [5]) } }; // time to run: 0.011808750001364 seconds.
> bench { 10000.do { f.valueArray(5) } }; // time to run: 0.0019109839995508 seconds.
> ```
> 
> Roughly half of it comes from the array construction:
> 
> ```supercollider
> bench { 10000.do { f.performArgs(\value, #[5]) } }; // time to run: 0.0028367470004014 seconds.
> bench { 10000.do { f.valueArray([5]) } }; // time to run: 0.0057900880019588 seconds.
> ```
> 
> We should keep in mind of course that these are low numbers on an absolute scale, and in the methods we talk about, the arrays are already constructed (currently they call `valueArray`). But they are also core functionality.
> 
> P.S. when I add `value1` as the original version with `valueArray`, I get e.g. this:
> 
> ```supercollider
> // performance varies quite a bit
> f = { |x| x };
> g = f.neg;
> bench { 10000.do { g.value(5) } }; // time to run: 0.0036773150022782 seconds.
> bench { 10000.do { g.value1(5) } }; // time to run: 0.0043830569993588 seconds.
> ```

... so this difference is the difference in the dispatch, I would say.

capital-G:
I'll always favor elegance and expressiveness over performance in high-level languages. This seems like a really useful and powerful feature as the performance impact seems reasonable, so I'd vote to include it.

telephon:
Yes, we want this to work! @JordanHendersonMusic do you have time for a `valueArgs` primitive implementation? For now, I would simply add a method in sclang that forwards to `this.performArgs(\value, args, kwargs)`, which could be replaced by a primitive.

JordanHendersonMusic:
I think it is basically a case of sticking blockValueWithKeys on the successful return paths of objectPerformArgs.

Unfortunately, I have my viva to prep for and some work that needs finishing in between, so I don't know if I will be able to get to this until mid February.

smoge:
> // We would like to be able to write things like this:
> f = { |x, y = 1| x + y } * { |x, y = 2| x ** y };
> f.(x:2, y:6) 
> 
> // or this
> f = { |x, y = 1| x + y };
> g = f.linexp(0, 255, 300, 1000);
> g.(x: 8);

The essential idea suggests a similar hack for currying, although I think it would not be possible to make this a "native" language feature. (Probably someone tried this somewhere already)

```
+Function {
    curry { | pinned = #[] |
        var needed = this.def.argNames.size;
        if(needed == 0) {
            ^this.value; 
        };
        if (pinned.size > needed) {
            Error("Too many arguments").throw;
        };
        if (pinned.size == needed) {
            ^this.valueArray(pinned);
        } {
            ^{ |x|
                if(x.isNil) {
                    Error("Argument cannot be nil").throw;
                };
                this.curry(pinned ++ [x]);
            }
        }
    }
}

/* tests
f = { |x, y, z| x + y + z };
g = f.curry;
g.(1).(2).(3)
f.curry.(1).(2).(3)
f.value(1, 2, 3)
~addTen = f.curry.(10);
~addTen.(20).(30)        
*/

```

telephon:
> I think it is basically a case of sticking blockValueWithKeys on the successful return paths of objectPerformArgs.

> But this would only be for function objects, there is no point in this for things that don't call blockValue*, as you need to go through sendMessage and dispatch anyway.

sounds good.

> Unfortunately, I have my viva to prep for and some work that needs finishing in between, so I don't know if I will be able to get to this until mid February.

Good luck with this! We can add the optimisation then later.

telephon:
Hi @JordanHendersonMusic – would now be a good moment to continue?

SCClassLibrary/JITLib/Patterns/OpFunctionProxy.sc
```diff
@@ -1,119 +1,121 @@
...
 	// behave like a pattern
-	embedInStream { arg inval;
+	embedInStream { | inval |
 		^this.value.embedInStream(inval)
 	}
 
 }
 
-// maybe make it an abstract function object.
+// this proxy is built on function proxiy composition
```
JordanHendersonMusic, line 106: Typo.

SCClassLibrary/JITLib/Patterns/OpFunctionProxy.sc
```diff
@@ -1,119 +1,121 @@
...
 NAryValueProxy : NAryOpFunctionProxy {
 
-	*new { arg receiver, args;
+	*new { | receiver, args |
 		^super.new(nil, receiver, args ? [])
 	}
-	reduceFuncProxy { arg args;
-		^a.reduceFuncProxy(arglist.collect(_.reduceFuncProxy(args)))
+	reduceFuncProxy { | args, protect = true, kwargs |
+		// kwargs are only passed to each proxy: keyword compositon isn't supported
```
JordanHendersonMusic, line 113: What is keyword composition in this context?

SCClassLibrary/JITLib/Patterns/Fdef.sc
```diff
@@ -7,85 +7,85 @@ Maybe : Ref {
...
 	all {
 		^this.source.all
 	}
-	do { arg function;
+	do { |function|
 		this.source.do(function) // problem: on the fly change is not picked up in this case.
 	}
 
-	doesNotUnderstand { arg selector ... args;
+	doesNotUnderstand { |selector ... args|
```
JordanHendersonMusic, line 81: Should there be kwargs here?

JordanHendersonMusic (commented):
This all looks good and makes sense. Do the tests cover all the touched methods? I don't know what the valueProxys do so can't check them.

Re. implementing valueArgs as a primitive: I'm going to be busy until the end of march (at least) with a project, but am happy to make an implementation after. I'd also be more than  happy to have a look at anyone else's implementation (earlier than after march). I know I've written all the newly refactored stack calling and kwargs stuff, so it might be beneficial to have someone else try and use it - check it is clear for whomever needs to use it next?

SCClassLibrary/JITLib/Patterns/OpFunctionProxy.sc
telephon: I've explained it better now:

```
		// kwargs are only passed to each proxy individually:
		// in function compositon we have no way to specify a keyword
		// to which argument of the next function the output value
		// of the previous function in the chain is passed
```

SCClassLibrary/JITLib/Patterns/Fdef.sc
telephon: Ok, yes. Needs touching a bit more, but it will be more complete this way.

telephon:
> This all looks good and makes sense. Do the tests cover all the touched methods? 

I'll go through once more, there may be still omissions.

>I don't know what the valueProxys do so can't check them.

These proxies are placeholders on which you can call methods and later redefine their receiver. Now the nary-ops on these placeholders can also use keywords, which wasn't possible before.

telephon:
> > This all looks good and makes sense. Do the tests cover all the touched methods?
> 
> I'll go through once more, there may be still omissions.

Should be done with now.

telephon:
One question i am left with: it would in principle be possible to introduce a class `DoesNotUnderstandOpProxy`, which handles the case of the forwarding of `doesNotUnderstand` in `Maybe`. Then, the class `NAryOpFunction` could stay as it was. Usually, `composeNAryOp` is called with the argument positions known, so it isn't necessary to pass and keep kwargs. But, perhaps we want this to be the case? One may want to be able to specify keywords without knowing their position for a method call? It is certainly a feature, and is currently implemented. But if we say this is too hypothetical, I'd throw this out again.

thoughts? @JordanHendersonMusic @capital-G

capital-G (commented):
Coming back to the posted examples

```supercollider
// We would like to be able to write things like this:
f = { |x, y = 1| x + y } * { |x, y = 2| x ** y };
f.(x:2, y:6) 
```

I found the implicit wrapping of functions not the most straight forward to grasp yet - what would be e.g. `f.x(2)` - is `y` `1` or `2` in this case?
I think it is fine doing this, but maybe there should be an example or documentation what is happening here.

```supercollider
f = { |x, y = 1| x + y };
g = f.linexp(0, 255, 300, 1000);
g.(x: 8);
```

While this is a cool feature, does it also works when operating with unmatched `kwargs`? I know that we already have a discussion about the importance/usage of `kwargs`, but this is a really powerful pattern within Python which helps with inheritance and API stabilization, which may also be relevant for sclang.

E.g. is this working?

```supercollider
f = {|...args, kwargs| kwargs.asDict[\something];} * {|...args, kwargs| kwargs.asDict[\else];};
f.(something: 20, else: 40};
```

I also think the addition could be mentioned in something like https://dev.docs.supercollider.online/Reference/Functions.html or https://dev.docs.supercollider.online/Classes/AbstractFunction.html#Function%20Composition so people can pick it up :)

Also, there is a test failing - I don't know how this could be linked to this PR but it is failing rather reliable in this PR, see https://github.com/supercollider/supercollider/actions/runs/13366576705/job/37325836829#step:7:7895

```
 	[0/1]	TestNodeProxy_Server:test_synthDef_isReleased_afterFree (2.21s)
```

I haven't grasped the implications on `Thunk` and `Maybe` yet, but I am sure they are correct if the tests are passing

HelpSource/Classes/FunctionList.schelp
```diff
@@ -61,7 +61,7 @@ code::
 // example
 
 a = nil;
-a = a.addFunc { |x = "", y = ""| "this % is an % example\n".postf(x, y); 1 };
+a = a.addFunc { |x = "", y = ""| "this % is an % example\n".postf(x, y); 1 };
```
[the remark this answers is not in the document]
telephon: probably this was an automatic formatting, removing an invisible special whitespace character. After the pipe, sometimes this happens, e.g. staying on `alt` a little too long when hitting space, so you get `alt-space`.

capital-G:
Thanks for the additional info, much appreciated! Looking forward to the feature!

> I've double checked, calling `UnitTest.runTest("TestNodeProxy_Server:test_synthDef_isReleased_afterFree")` succeeds and all tests in that class do as well.

It is only failing on Linux and not on macOS, so it makes sense that the test is passing on your local machine.
Another recent PR is passing this test though, see https://github.com/supercollider/supercollider/pull/6620

Just to be sure it is not an older bug - can you rebase against develop and push-force? `develop` should be ok currently.
If it still fails we have to figure out what is wrong.

dyfer:
I think the test failure was random, but this is too old to manually retrigger the test - I'd suggest rebasing and current develop and force pushing.

A lot of discussion here - are there any outstanding issues to address?

telephon:
It is good as is, but there are two open questions that could perhaps be addressed in a later PR?

## primitive implementation
> Yes, we want this to work! @JordanHendersonMusic do you have time for a `valueArgs` primitive implementation? For now, I would simply add a method in sclang that forwards to `this.performArgs(\value, args, kwargs)`, which could be replaced by a primitive.

## DoesNotUnderstandOpProxy
> One question i am left with: it would in principle be possible to introduce a class `DoesNotUnderstandOpProxy`, which handles the case of the forwarding of `doesNotUnderstand` in `Maybe`. Then, the class `NAryOpFunction` could stay as it was. Usually, `composeNAryOp` is called with the argument positions known, so it isn't necessary to pass and keep kwargs. But, perhaps we want this to be the case? One may want to be able to specify keywords without knowing their position for a method call? It is certainly a feature, and is currently implemented. But if we say this is too hypothetical, I'd throw this out again.

telephon:
Hm, I think if we just squash the whole thing it should be fine. I don't see anything problematic in the diffs: https://github.com/supercollider/supercollider/pull/6584/files

telephon:
But of course the tests fail (https://github.com/supercollider/supercollider/actions/runs/14454834427) I don't know why they do.

telephon:
OK, done. I hope I have chosen the right code, please double-check.

JordanHendersonMusic:
They fail because I pushed a broken version! That is why it might be easier if you remove what I've done and I'll send a fresh commit?

telephon:
ah ok. Well, let's see, I've reverted and pushed.

telephon:
hm, still fail, but apparently for unrelated reasons.

dyfer:
> hm, still fail, but apparently for unrelated reasons.

Yes, I've also seen both `TestServer_clientID_booted:test_repeatedRemoteLogin` and `TestNetAddr:test_receiveOSC` fail randomly.

telephon:
Superseded by: #6786

PR #6587, opened 2025-01-08
Fix Pwalk boundary behavior when using directionPattern for folding

jamshark70:
## Purpose and Motivation

See discussion at https://scsynth.org/t/pwalk-how-do-you-always-fold-at-both-boundaries/10401/10?u=jamshark70

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] ~Updated documentation~

### DIFF
--- a/SCClassLibrary/Common/Streams/ListPatterns.sc
+++ b/SCClassLibrary/Common/Streams/ListPatterns.sc
@@ -391,7 +391,7 @@ Pwalk : ListPattern {
 
 
 	embedInStream { arg inval;
-		var	step;
+		var	step, stepDir;
 		var index = startPos.value(inval);
 		var stepStream = stepPattern.asStream;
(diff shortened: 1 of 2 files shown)


SCClassLibrary/Common/Streams/ListPatterns.sc
```diff
@@ -391,7 +391,7 @@ Pwalk : ListPattern {
 
 
 	embedInStream { arg inval;
-		var	step;
+		var	step, stepDir;
```
telephon, line 394: "stepDir" sounds like "step direction", which it isn't. Maybe "dirStep" or "directionalStep"?

jamshark70: OK, done.

capital-G (approved):

Issue #6189, opened 2024-01-09
`NdefMixer` does not survive `Ndef.clear`


capital-G:
## Environment

* SuperCollider version: 3.13.0
* Operating system: macOS 13
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
NdefMixer.new(s);
Ndef(\foo, {SinOsc.ar});
// NdefMixer gets updated
```

<img width="796" alt="image" src="https://github.com/supercollider/supercollider/assets/8267062/c79f357b-0403-41be-81d3-804959587ee3">

```supercollider
Ndef.clear;

// \foo disapears from NdefMixer
```

<img width="796" alt="image" src="https://github.com/supercollider/supercollider/assets/8267062/b4537b9a-e975-4684-8707-acebf85f987c">

```supercollider
Ndef(\foo, {SinOsc.ar});
// \foo does not appear anymore
```

<img width="796" alt="image" src="https://github.com/supercollider/supercollider/assets/8267062/0d2e8dcb-888b-40ee-98d8-777d8df63a5a">

When passing a Server to `NdefMixer.new`, this Server gets transfered to a ProxySpace

https://github.com/supercollider/supercollider/blob/23ec13a5a2610b0af3755f053611142c7faea875/SCClassLibrary/JITLib/GUI/ProxyMixer.sc#L382-L389

which calls

https://github.com/supercollider/supercollider/blob/23ec13a5a2610b0af3755f053611142c7faea875/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc#L1113-L1121

Most likely this ProxySpace will get cleared via `Ndef.clear`? Would it be a bad idea to keep a "resilient meta ProxySpace" for such use cases?

## Expected vs. actual behavior

I think `NdefMixer` should survive a `Ndef.clear`, such that it represents the Ndefs of a server at all time.

JordanHendersonMusic:
As this is a part of the design, I'm going to close the issue. Feel free to open a discussion if anyone thinks the behaviour should be changed.

telephon:
It is a valid point, so I am happy to rethink if it is reopened at a later stage.

capital-G:
> As this is a part of the design, I'm going to close the issue.

[...]

PR #6588, opened 2025-01-10
Ndef: make clear methods more flexible

adcxyz:
## Purpose and Motivation

Current Ndef.clear seems too drastic for many uses, so this PR proposes two separate methods: 
```Ndef.hardClear;```, which clears all existing Ndefs, and removes the proxyspaces in ```Ndef.all```, 
which is how Ndef.clear behaves so far. 
and 
```Ndef.clear;```, which clears all existing Ndefs, but now keeps the (empty) proxyspaces in Ndef.all.

This changes the default behavior of clear: all objects that track spaces in Ndef.all will keep seeing them, 
which fixes the surprise/arguable bug noted in #6189 with minimal intervention.

The current harder behavior is still available with a clearer name ```hardClear```, 
in analogy to ```Server:freeAll, Server:hardFreeAll```, ```Main:stop, Main:hardStop```, etc. 

## Types of changes

- Small change of default behavior
- Bug fix
- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Ndef.schelp
+++ b/HelpSource/Classes/Ndef.schelp
@@ -65,7 +65,10 @@ method::kr
 equivalent to code::*new(key).kr(numChannels, offset):: (see link::Classes/BusPlug#-kr::)
 
 method::clear
-clear all proxies
+clear all proxies (in all proxyspaces)
+
+method::hardClear
+clear all proxies and remove all existing proxyspaces
 
 method::defaultServer
 set the default server (default: code::Server.default::)
--- a/HelpSource/Classes/Ndef.schelp
+++ b/HelpSource/Classes/Ndef.schelp
@@ -67,7 +67,7 @@ equivalent to code::*new(key).kr(numChannels, offset):: (see link::Classes/BusPl
 method::clear
 clear all proxies (in all proxyspaces)
 
-method::hardClear
+method::clearDict
 clear all proxies and remove all existing proxyspaces
 
 method::defaultServer
--- a/HelpSource/Classes/Ndef.schelp
+++ b/HelpSource/Classes/Ndef.schelp
@@ -65,10 +65,10 @@ method::kr
 equivalent to code::*new(key).kr(numChannels, offset):: (see link::Classes/BusPlug#-kr::)
 
 method::clear
-clear all proxies (in all proxyspaces)
+clear all Ndefs in all proxyspaces of the Ndef class
 
 method::clearDict
-clear all proxies and remove all existing proxyspaces
+clear all Ndefs and remove all proxyspaces of the Ndef class
 
 method::defaultServer
 set the default server (default: code::Server.default::)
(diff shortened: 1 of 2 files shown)


telephon (commented):
The functionality is good – it is a breaking change, so we should make that explicit. The name should start with `clear` so it is found in autocompletion. `clearAll` seems best to me.

We should also think again why this is necessary. I keep missing the reason why NdefMixer can't follow a ProxySpace that has been emptied.

capital-G:
I don't know enough about the internals of JITlib in order to oversee any implications, but I think it is a good addition.
And although I appreciate a good autocomplete, I think the consistency of `hardFree` is actually favorable here.

 In other programming languages I'd add a default argument to `clear` (e.g. `hard=false`), but I think these things are not too common in sclang?

capital-G (approved):

adcxyz:
> The functionality is good – it is a breaking change, so we should make that explicit. The name should start with `clear` so it is found in autocompletion. `clearAll` seems best to me.
> 
> We should also think again why this is necessary. I keep missing the reason why NdefMixer can't follow a ProxySpace that has been emptied.

The current implementation of Ndef:clear empties all proxyspaces in Ndef.all, which is fine. 
The problem is that it also empties the dictionary Ndef.all as well, so now no server has a proxyspace in Ndef.all. 
Any observers, such as NdefMixers, will usually have a proxyspace as its object (as NdefMixer does), and thus will hold on to the previous proxyspace they were watching. 
The server variable idea explored in the previous PR seems really complicated to explain, and creates more 
Since they should work with any proxyspace, it makes more sense to
When a new Ndef is made on that same server, a new proxyspace is implicitly created for that server, 
and put a in Ndef.all. So now there is the old proxyspace that bel

telephon:
Yes, that makes sense to me.

telephon:
I would prefer it if you could replace `hardClear` by `clearAll` or `clearUp`.

adcxyz:
> I don't know enough about the internals of JITlib in order to oversee any implications, but I think it is a good addition. And although I appreciate a good autocomplete, I think the consistency of `hardFree` is actually favorable here.
I agree.
> In other programming languages I'd add a default argument to `clear` (e.g. `hard=false`), but I think these things are not too common in sclang?
could be done - so ```Ndef.clear(hard: true)``` does ```Ndef.hardClear```

telephon:
Well, my point was that "hard" sounds a bit like without a fade time, or something, well, hard. This is why I suggested another word.

telephon:
@adcxyz what do you think?

capital-G (approved):

adcxyz:
As discussed in person with @telephon, I will rename the proposed hardClear method to Ndef.clearDict, 
because this clearly says that the dictionary which Ndef class keeps for all servers 
will be cleared, and then it is good to go.

adcxyz:
ready when you are :-)

telephon:
@adcxyz just two corrections in the documentation and we are done …

adcxyz:
> @adcxyz just two corrections in the documentation and we are done …

thanks @telephon !
One more tweak if I may : 
> clear all proxies (in all proxyspaces of the Ndef class)
clear all Ndefs in all proxyspaces of the Ndef class 

> clear all proxies and remove all existing proxyspaces of the Ndef class
clear all Ndefs and remove all proxyspaces of the Ndef class

OK?

telephon (approved):

PR #6593, opened 2025-01-18
Add macOS arm nightly build upload to s3

capital-G:
## Purpose and Motivation

See https://github.com/supercollider/supercollider.github.io/issues/285
PR for website which should be merged after this PR has been merged: https://github.com/supercollider/supercollider.github.io/pull/286

Please double-check the CI changes - they only run on main/develop and we should not use those branches for CI debugging :)

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -101,11 +101,18 @@ jobs:
       matrix:
         include:
           - artifact-suffix: macOS-x64
-            s3-os-name: osx
+            s3-os-name: osx-x64
             s3-artifact-suffx: ''
             s3-artifact-extension: 'dmg'
             artifact-extension: '.dmg' # for download-artifacts action, for non-zip only, should include `.`
             s3-create-latest-link: true # create link to pointing to the "latest" build; activate only one per branch per s3-os-name  
+        
+          - artifact-suffix: macOS-arm64
+            s3-os-name: osx-arm
+            s3-artifact-suffx: ''
+            s3-artifact-extension: 'dmg'
+            artifact-extension: '.dmg'
+            s3-create-latest-link: true
 
           - artifact-suffix: win64
             s3-os-name: win64
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -102,21 +102,24 @@ jobs:
         include:
           - artifact-suffix: macOS-x64
             s3-os-name: macos-x64
-            s3-artifact-suffx: 'macos-x64'
+            s3-artifact-prefix: '-macOS-x64'
+            s3-artifact-suffix: ''
             s3-artifact-extension: 'dmg'
             artifact-extension: '.dmg' # for download-artifacts action, for non-zip only, should include `.`
             s3-create-latest-link: true # create link to pointing to the "latest" build; activate only one per branch per s3-os-name  
         
           - artifact-suffix: macOS-arm64
             s3-os-name: macos-arm64
-            s3-artifact-suffx: 'macos-arm64'
+            s3-artifact-prefix: '-macOS-arm64'
+            s3-artifact-suffix: ''
             s3-artifact-extension: 'dmg'
             artifact-extension: '.dmg'
             s3-create-latest-link: true
 
           - artifact-suffix: win64
             s3-os-name: win64
-            s3-artifact-suffx: 'win64'
+            s3-artifact-prefix: '-win64'
+            s3-artifact-suffix: ''
             s3-artifact-extension: 'zip'
             s3-create-latest-link: true # create link to pointing to the "latest" build
 
@@ -130,7 +133,7 @@ jobs:
       UPLOAD_TO_S3: ${{ (secrets.S3_ACCESS_KEY_ID != 0) && !startsWith(github.ref, 'refs/tags/') }}
       S3_CREATE_LATEST_LINK: ${{ matrix.s3-create-latest-link && (secrets.S3_ACCESS_KEY_ID != 0) && startsWith(github.ref, 'refs/heads/') }}
       S3_ARTIFACT_PATH: ${{ github.workspace }}/build/s3-upload
-      S3_ARTIFACT_NAME: SC-${{ github.sha }}${{ matrix.s3-artifact-suffx }}.${{ matrix.s3-artifact-extension }}
+      S3_ARTIFACT_NAME: SC${{ matrix.s3-artifact-prefix }}-${{ github.sha }}${{ matrix.s3-artifact-suffix }}.${{ matrix.s3-artifact-extension }}
       S3_BUILD_LOCATION: builds/supercollider/supercollider/${{ matrix.s3-os-name }}
       S3_ROOT_URL: 'https://supercollider.s3.amazonaws.com'
     steps:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -101,22 +101,22 @@ jobs:
       matrix:
         include:
           - artifact-suffix: macOS-x64
-            s3-os-name: osx-x64
-            s3-artifact-suffx: ''
+            s3-os-name: macos-x64
+            s3-artifact-suffx: 'macos-x64'
             s3-artifact-extension: 'dmg'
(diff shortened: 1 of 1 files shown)


dyfer (changes requested):
Thanks!

I looked this over and AFAICT this should work.

FYI:
With the proposed implementation we're gonna get the following paths on S3:
`builds/supercollider/supercollider/osx-x64/SC-<SHA>.dmg` and `builds/supercollider/supercollider/osx-arm/SC-<SHA>.dmg`
which is fine, and maybe even consistent with the win32 / win64 builds (back when we did win32). However, this creates two DMGs with the same name but for different architecture.
As an alternative, we could use `s3-artifact-suffix` in the matrix, which could be used to create something like 
`builds/supercollider/supercollider/osx/SC-<SHA>-x64.dmg` and `builds/supercollider/supercollider/osx/SC-<SHA>-arm.dmg`

I don't have a strong preference, the former is consistent with the win64 upload, but I like the latter because it adds the OS name to the artifact. In fact, I wouldn't mind changing it so that we use `SC-<SHA>-win64` for the Windows build, if we are touching this code.

Here's the "requested change" part:

Whichever route we go, I'd like to replace `osx` with `macos`. I'd also prefer `arm64` over `arm`. Beyond that, feel free to choose between whichever approach.

Note to self: after we merge this, we'll need to update the links on the website, as the "latest" link will look differently, even for the macOS Intel build.

capital-G:
Thanks for you review @dyfer ! So maybe something like this should include everything - should we settle for this?
While we are at it - lets replace osx with macOS.

```diff
diff --git a/.github/workflows/actions.yml b/.github/workflows/actions.yml
index cd691f1242..09f52c9e0f 100644
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -101,15 +101,15 @@ jobs:
       matrix:
         include:
           - artifact-suffix: macOS-x64
-            s3-os-name: osx-x64
-            s3-artifact-suffx: ''
+            s3-os-name: macos-x64
+            s3-artifact-suffx: 'macos-x64'
             s3-artifact-extension: 'dmg'
             artifact-extension: '.dmg' # for download-artifacts action, for non-zip only, should include `.`
             s3-create-latest-link: true # create link to pointing to the "latest" build; activate only one per branch per s3-os-name  
         
           - artifact-suffix: macOS-arm64
-            s3-os-name: osx-arm
-            s3-artifact-suffx: ''
+            s3-os-name: macos-arm64
+            s3-artifact-suffx: 'macos-arm64'
             s3-artifact-extension: 'dmg'
             artifact-extension: '.dmg'
             s3-create-latest-link: true

```

BTW - who owns/pays for this s3 bucket? :D

capital-G:
> Should we _also_ add `win64` to the Windows artifact name while we're at it? It would be an unrelated change, but this code is seldom touched and testing is cumbersome...

Sounds like a good idea, will include it.



> If you push your branch to the main SC repository, the artifacts will be uploaded to S3, even if they're not on a protected branch.

Oh, okay? Looking at

https://github.com/supercollider/supercollider/blob/a4ffd92d88e4f89d426318ac35b21ad12bd8a11f/.github/workflows/actions.yml#L116

it seems we publish the "latest download" on every direct push of any branch in the SuperCollider repo? I don't think this is actually good? Shouldn't this include like a `if branch_name==develop`?

dyfer:
> it seems we publish the "latest download" on every direct push of any branch in the SuperCollider repo? I don't think this is actually good? Shouldn't this include like a `if branch_name==develop`?

I had to check, but keep in mind this has been like that for years now, so if it indeed was incorrect, somebody would likely notice... OR nobody uses that feature... in any case: that latest link includes the branch name (see https://github.com/supercollider/supercollider/pull/6593/files#diff-2702f6f7d10a1bffbc8596d1a9925b1234bc4c7a6ff4c21ca6b34c13aa75780eL150) so this is functionally fine, creating a unique "latest" link for each branch on the main repo. With GHA artifacts this might not be unnecessary... but if we are to discuss the change to that behavior, let's do that in a separate Issue/PR.

dyfer:
One more note - I am guessing that the reason to use a name like`.../<platform>/SC-<SHA>.zip` was to easily download a build for any commit without the need to hunt for a URL. BUT I don't think this is as crucial now that we also have GHA artifacts and I'm guessing this feature is mostly used for the latest build anyway.

capital-G:
I adjusted the code and also pushed the branch directly to this repo so we can test the CI, see https://github.com/supercollider/supercollider/actions/runs/12910600924, which results in

[`deploy macOS-x64 to s3`](https://github.com/supercollider/supercollider/actions/runs/12910600924/job/36003156778#step:7:21)

* https://supercollider.s3.amazonaws.com/builds/supercollider/supercollider/macos-x64/SC-8454530c5f24f54f3843cc5c6921bbcd5f5d203bmacos-x64.dmg
* https://supercollider.s3.amazonaws.com/builds/supercollider/supercollider/macos-x64/macOS-arm-s3-download-latest.html

and [`deploy macOS-arm64 to s3`](https://github.com/supercollider/supercollider/actions/runs/12910600924/job/36003158089#step:7:21)

* https://supercollider.s3.amazonaws.com/builds/supercollider/supercollider/macos-arm64/SC-8454530c5f24f54f3843cc5c6921bbcd5f5d203bmacos-arm64.dmg
* https://supercollider.s3.amazonaws.com/builds/supercollider/supercollider/macos-arm64/macOS-arm-s3-download-latest.html

and [`deploy win64 to s3`](https://github.com/supercollider/supercollider/actions/runs/12910600924/job/36003158611#step:7:21)

* https://supercollider.s3.amazonaws.com/builds/supercollider/supercollider/win64/SC-8454530c5f24f54f3843cc5c6921bbcd5f5d203bwin64.zip
* https://supercollider.s3.amazonaws.com/builds/supercollider/supercollider/win64/macOS-arm-s3-download-latest.html

> I had to check, but keep in mind this has been like that for years now, so if it indeed was incorrect, somebody would likely notice... OR nobody used that feature... in any case: that latest link includes a [branch name](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/actions.yml#L150):
> 
> ```
>         echo ${{ env.S3_ARTIFACT_PATH }}/${GITHUB_REF#refs/heads/}-latest.html >> $GITHUB_ENV
> ```

Maybe we should move towards less use of GitHub actions-specific code and use dedicated shell or python scripts for this as it allows to be debugged and tested locally - this would also make a transition away from GHA easier.
I still don't understand what the LOC actually does :D But this is something for another PR.

And there seems to be a typo, s3-artifact-__suffx__ is missing an `i` - but I have kept it b/c it is a local variable.

capital-G:
Decided to introduce a prefix for the S3 files so the downloaded file becomes

`SC-macOS-arm64-9c8ecb934c78300de7718924ca9450c618dae455.dmg`

instead of

`SC-9c8ecb934c78300de7718924ca9450c618dae455-macos-arm64.dmg`

dyfer (approved):
Thanks! 
And thanks for fixing the typo :)

dyfer:
> I still don't understand what the LOC actually does :D

I missed this before. There are a few things happening in the section [here](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/actions.yml#L153-L155):
1. we set an environment variable to be available in subsequent steps in GHA. Because each step is somewhat separated, it's a GHA-specific way of setting a variable - instead of `export VARNAME=value` it's `
echo "{VARNAME}={value}" >> "$GITHUB_ENV"` (see [GHA docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable))
2. we remove `refs/heads` from the git ref to get the branch name: `${GITHUB_REF#refs/heads/}`. This is just bash.
3. in the [linked code](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/actions.yml#L153-L155) this is separated into multiple lines (that's why there's `EOF` at the end)

PR #6595, opened 2025-01-18
classlib: support ugens for more math ops

telephon:
- complex on ugens work for square root
- sign checks work for UGens

## Purpose and Motivation

- It is generally possible to use complex numbers whose components are ugens. Only the square root function has a tacit type assumption in it. This PR fixes this.
- there are a number of methods that check the sign of numbers – they should also be lifted to UGens.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/UGen.schelp
+++ b/HelpSource/Classes/UGen.schelp
@@ -344,8 +344,34 @@ code::
 ::
 
 method::binaryValue
-Outputs 1, representing true for positive, 0 for negative or zero input values.
+Outputs 1, representing true for positive, 0 representing false for negative or zero input values.
 
+
+method::isPositive
+Outputs 1, representing true for positive or zero, 0 representing false for negative input values.
+
+method::isStrictlyPOsitive
+Outputs 1, representing true for positive, 0 representing false for negative or zero input values.
+
+method::isNegative
+Outputs 1, representing true for negative, 0 representing false for positive or zero input values.
+
+
+method::|==|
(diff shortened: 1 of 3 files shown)


telephon:
I'll add documentation if we agree on the change.

dyfer:
This looks fine to me, but I don't think I can see the larger scale repercussions of the change.

@joslloand do you have any opinion on this change?

joslloand:
> This looks fine to me, but I don't think I can see possible larger scale repercussions of this change.
> 
> @joslloand do you have any opinion on this?

I don't anticipate any issues, so looks good to me.

telephon:
Great! I've added documentation (also for the |==| operator, which was missing).

dyfer (approved):

prko:
This is already merged, but the `isPositive` method seems to be mistyped in the schelp file:

https://github.com/supercollider/supercollider/blob/59963f48f75558f471f2e20df53b4c478eb3d93f/HelpSource/Classes/UGen.schelp#L353

https://github.com/supercollider/supercollider/blob/59963f48f75558f471f2e20df53b4c478eb3d93f/SCClassLibrary/Common/Audio/UGen.sc#L290

Could you check this? I get a warning when doing `SCDoc.renderAll`.

telephon:
> Could you check this? I get a warning when doing `SCDoc.renderAll`.

I get hundreds of warnings and errors doing this, so it has escaped me …

prko:
Thank you very much!

PR #6608, opened 2025-01-31
docs: BufRd, DelTapRd: interpolation argument: explanatory note

HotwheelsSisyphus:
## Purpose and Motivation

Primarily to get a hang of the contribution workflow ("my first PR"). 

(EDIT: now I notice I accidentally added empty lines after the explanatory note, so I guess now I gotta learn how to modify PRs or ... whatever it is that I need to do. I thought I could just add another commit on top of the existing ones and then add it to the PR (or even not), or amend the previous commits. But both of these things don't seem to work, for reasons I don't fully understand; can't seem to push this new commit to my fork. I'll call it quits today but would appreciate guidance...).

As concerns the documentation: 
Clarify that the choice of 1,2,4 for the interpolation arg is intentional, and why.  
This confused me at one point, I thought it was a typo, as 1,2,3 would have seemed more intuitive; 
but then I checked the source and of course there is indeed a mathematical reason behind it. At any rate, for people
who mostly want to make noises it might help to point out that the choice of 1,2, or 4 for the interp 
arg is intentional and not a typo.

## Types of changes

- Documentation
1. Clarification re interpolation argument
2. Cosmetics: removed some double spaces in the text of those schelp files.

### DIFF
--- a/HelpSource/Classes/BufRd.schelp
+++ b/HelpSource/Classes/BufRd.schelp
@@ -23,25 +23,23 @@ is compiled.
 note::
 If you supply a code::bufnum:: of a buffer that has a
 different code::numChannels:: than you have specified to
-the BufRd, it will post a warning and output the channels it can.
-::
+the BufRd, it will post a warning and output the channels it can.::
 
 argument::bufnum
 The index of the buffer to use.
 
 argument::phase
 Audio rate modulateable index into the buffer.
-Warning:: The phase argument only offers precision for addressing 2**24 samples (about 6.3 minutes at 44100Hz). ::
+Warning:: The phase argument only offers precision for addressing 2**24 samples (about 6.3 minutes at 44100Hz).::
 
 argument::loop
 1 means true, 0 means false. This is modulateable.
 
-
 argument::interpolation
-1 means no interpolation, 2 is linear, 4 is cubic interpolation. (These
-numbers correspond to the number of consecutive samples needed to compute each type of
-interpolation.) Any other values for this argument will silently default to no interpolation.
-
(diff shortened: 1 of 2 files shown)


telephon (approved):
This is a good change, thank you!

capital-G (approved):

capital-G:
Thanks for your contribution!

> I thought I could just add another commit on top of the existing ones and then add it to the PR (or even not), or amend the previous commits. But both of these things don't seem to work, for reasons I don't fully understand; can't seem to push this new commit to my fork. I'll call it quits today but would appreciate guidance

You should be able to do commits locally as these are local changes and can't be rejected e.g. by GitHub.
Try to verify the status of your local repo with `git status` and `git log` and check if you are in a dangling state (e.g. your local repo could be in an intermediate state of rebasing or merge conflict which could result in rejected commits).
If everything else fails - switch to your fork branch `HotwheelsSisyphus:develop` and perform a `git --hard reset` (this will delete any local modifications in the repo!).
Afterwards you should be in a good state to do a commit and push those to your fork.

If you still run into problems, do not hesitate to reply :)

HotwheelsSisyphus:
Ok, I think I mostly figured it out.  (In essence I had to pull from remote, resolve merge conflicts, then push to remote again.) Just wanted to do it properly, and also get a better understanding of what I need to do on the git side of things. I figure this kind of situation (updating a pull request) arises frequently enough so it's worth doing that legwork now.

dyfer:
@HotwheelsSisyphus thanks for the PR! And welcome!

PR #6609, opened 2025-02-02
Fix double invocation of unsaved documents dialog upon quitting

### LINKED ISSUE #6562: Unsaved document dialog appears twice
## Environment

* SuperCollider version: dev
* Operating system: macOS 15.0.1

## Steps to reproduce

* Open IDE
* write something into an unsaved document
* close the IDE

After clicking *discard* on the unsaved documents dialog, it will appear a second time asking the same question

grafik

capital-G:
## Purpose and Motivation

Fixes #6562 

## Types of changes

- Bug fix

I checked the call stacks of both dialog instances

```cpp
// call stack of first window

ScIDE::MainWindow::promptSaveDocs() main_window.cpp:1224
ScIDE::MainWindow::quit() main_window.cpp:807
ScIDE::MainWindow::onQuit() main_window.cpp:819

// call stack of second window

ScIDE::MainWindow::promptSaveDocs() main_window.cpp:1224
ScIDE::MainWindow::quit() main_window.cpp:807
ScIDE::MainWindow::closeEvent(QCloseEvent *) main_window.cpp:885
```

~~I decided to make the second call stack, invoked by a `closeEvent`, to do nothing. Please review if this could have any unintended side-effects - but I haven't figured out any so far.~~
=> I applied the fix suggested in https://github.com/supercollider/supercollider/pull/6609#issuecomment-2646122928 which seems to fix the issue. Please verify on Windows and Linux.

I think the behavior was changed by the Qt6 switch because the life-cycle of the main-event-loop may have been adjusted by Qt?

## To-do list

As I haven't heard any other people complain about this I think this could be a macOS only problem? Therefore it may be good if people with Linux and Windows could try out this branch and see if it has some unintentional side-effects (e.g. no unsaved documents dialog popping up)

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/main_window.cpp
+++ b/editors/sc-ide/widgets/main_window.cpp
@@ -244,7 +244,7 @@ void MainWindow::createActions() {
     // (such as cmd+q or window closing)
     action->setMenuRole(QAction::QuitRole);
 
-    QObject::connect(action, SIGNAL(triggered()), this, SLOT(onQuit()));
+    QObject::connect(action, &QAction::triggered, this, &QWidget::close, Qt::QueuedConnection);
     settings->addAction(action, "ide-quit", ideCategory);
 
     mActions[DocNew] = action = new QAction(QIcon::fromTheme("document-new"), tr("&New"), this);


xunil-cloud:
>I think this could be a macOS only problem?

On my Arch Linux system, there is no issue If I close the IDE using the window's close button. I discovered that I can reproduce the issue by either pressing Ctrl+q or closing the window through the File -> Quit menu. (both `onQuit` and `closeEvent` are called.)

But only `closeEvent` is called If I close the IDE using the window's close button, so the dialog does not show up with this change.

I come up with this change (without the change in this PR), seems to works on my side.
```diff
diff --git a/editors/sc-ide/widgets/main_window.cpp b/editors/sc-ide/widgets/main_window.cpp
index 7e99449681..9e318dc647 100644
--- a/editors/sc-ide/widgets/main_window.cpp
+++ b/editors/sc-ide/widgets/main_window.cpp
@@ -244,7 +244,7 @@ void MainWindow::createActions() {
     // (such as cmd+q or window closing)
     action->setMenuRole(QAction::QuitRole);

-    QObject::connect(action, SIGNAL(triggered()), this, SLOT(onQuit()));
+    QObject::connect(action, &QAction::triggered, this, &QApplication::quit, Qt::QueuedConnection);
     settings->addAction(action, "ide-quit", ideCategory);

     mActions[DocNew] = action = new QAction(QIcon::fromTheme("document-new"), tr("&New"), this);
```

prko:
> I think this could be a macOS only problem?

On Windows 11, there is no double invocation of unsaved documents dialog upon quitting.

capital-G:
> I come up with this change (without the change in this PR), seems to works on my side.

Thanks 👍  This also seems to do the fix the issue on macOS. I changed the commit to include your change and force-pushed it.

Please review :)

xunil-cloud:
Thanks for giving this a try. I find this doesn't work in Qt5 (dialog won't show up). Actually, I think it's more reasonable to use `QWidget::close` instead of `QApplication::quit` because what I really want is for the `closeEvent` to be called. Also, we don't want to quit the application if the user click the cancel button in dialog.

Quote from Qt [doc](https://doc.qt.io/qt-6/qcloseevent.html):
>Close events are sent to widgets that the user wants to close, usually by choosing "Close" from the window menu, or by clicking the X title bar button. They are also sent when you call [QWidget::close](https://doc.qt.io/qt-6/qwidget.html#close)() to close a widget programmatically.


[QGuiApplication::lastWindowClosed()](https://doc.qt.io/qt-6/qguiapplication.html#lastWindowClosed):
>By default, [QGuiApplication](https://doc.qt.io/qt-6/qguiapplication.html) quits after this signal is emitted.

dyfer:
@capital-G what do you think about @xunil-cloud's suggestion?

capital-G:
Sounds like the proper way to do it, I hope I can fix the PR accordingly next week :) Will ping you for a review @dyfer :)

capital-G:
@dyfer @xunil-cloud can you give this a test? Rebased to develop and applied the suggested changes by xunil-cloud - on Qt6 and macOS everything seems to work as expected for me.

dyfer (approved):
I tested both macOS arm64 (Qt6) and legacy (Qt5) builds and both seem to behave correctly, thanks!

xunil-cloud:
I tested this on Arch Linux using both Qt6 and Qt5, and the behavior is correct, thanks.

dyfer:
Thanks, merging!

Issue #6611, opened 2025-02-08
Error in boot with last develop


sonoro1234:
After updating to last develop I get this errors in the boot scsynth process:

```
while reading file: 'C:\Lua2SC\SuperCollider\synthdefs\bowsoundboard.scsyndef'exception in GraphDef_Load: UGen '' not installed.
while reading file: 'C:\Lua2SC\SuperCollider\synthdefs\cascabeles.scsyndef'exception in GraphDef_Load: name too long (> 31 chars): E
```

The synths were from sc 3.13
If I delete the synthdefs and generate them again with last develop, the same issue happens when rebooting.

Any clue of what could be happening?

sonoro1234:
Trying to see when the problem begun to appear: In commit SHA-1 b8d6d1d40064d42f8178713ae74f755a8d27b421 (* Merge pull request #6464 from dyfer/topic/qt-remove-deprecations-2

QT: remove deprecations for QT 5.15

) it was already present in supernova but not in scsynth.

Quite exhausted by long compilation times. Tomorrow I will check 3.13.1-rc1

sonoro1234:
3.13.0: works

3.13.1-rc1: works

578eb17aac26f304c2e727f8f0f8f1b35d3c83a3 which should be pre-std::filesystem: scsynth works, supernova fails

last develop: fails

-------------------------
last two compilations mean that something was failing with supernova previously to std::filesystem (If I founded the correct commit)
Otherwise: filesystem only affects to the path of files not to the content readed which is what seems to fail.
The problem I am facing now is that I need to do two different bisects: one for scsynth and the other for supernova ;-(

sonoro1234:
bisecting for scsynth:

4969a0ff0dc4d7d669bac6a060ca9215eb91e8c1 fails to compile

capital-G:
The line which fails here seems to be

https://github.com/supercollider/supercollider/blob/863953326a6775a965c75f3436e3ee0692daf730/server/scsynth/SC_GraphDef.cpp#L63-L72

We have experienced some problems like #6501 with the change from `boost::filesystem` to `std::filesystem` (which happened in #6331).
The introduced bugs only seem to affect windows, so it seems worth if someone with a windows machine could debug these issues in question - having the PR changes available it should be possible to narrow it down where something fails.

[...]

PR #6612, opened 2025-02-09
synthdef load needs binary on windows

sonoro1234:
solves issue #6611 

There are other ifstream and ofstream in the codebase that COULD need that but they are in components not used by me, so I am not sure.
Basically: when a file is opened in binary mode in windows there is not alteration of content, when it is opened in text mode ends of line are converted.

https://stackoverflow.com/questions/73570406/what-are-the-consequences-of-opening-a-file-in-c-in-text-or-binary-mode

### DIFF
--- a/server/scsynth/SC_GraphDef.cpp
+++ b/server/scsynth/SC_GraphDef.cpp
@@ -615,7 +615,7 @@ GraphDef* GraphDef_LoadGlob(World* inWorld, const char* pattern, GraphDef* inLis
 }
 
 std::string load_file(const std::filesystem::path& file_path) {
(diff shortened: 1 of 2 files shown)


Spacechild1 (approved):

capital-G (approved):

capital-G:
Thanks! The fix look fine to me and makes somehow sense and their impact _should_ be limited.
If any other problems with file reading in windows comes up we should consider applying these fixes elsewhere as well, but currently I am not aware of any so let's merge it as-is.

See also https://en.cppreference.com/w/cpp/io/c/FILE#Binary_and_text_modes

PR #6569, opened 2024-12-12
Add wasm build of scsynth

capital-G:
# wasm build of scsynth

__ALL CREDIT GOES TO @sciss / Hans-Holger Rutz who implemented this already 4 years ago (see https://github.com/supercollider/supercollider/pull/5309) - a BIG BIG thanks for this!__

Try it out @ <https://supercollider.dennis-scheiba.com/> or download the files locally via the CI, e.g. https://github.com/capital-G/supercollider/actions/runs/12298127473/artifacts/2311994694

I always felt that this would be an awesome addition to the SuperCollider project and I therefore decided to pick up the PR again (see https://github.com/supercollider/supercollider/pull/5309 and https://github.com/supercollider/supercollider/pull/5571).
I hopefully adressed most of the mentioned issues.

I am willing to take care of this build and already have plans for some additions to it.
With CI we can be sure that we don't break it accidentially ourselves.
If the browser manufactures decide to remove/change this functionallity this would need some care or we could remove it, but for now it works.

Wasm is using 32bit, so it is a good test for our code base!

## Changes

* Rebased https://github.com/supercollider/supercollider/pull/5571 on dev (ignored submodule updates)
* Translated shell build scripts into CMake
* Updated emscript SDK to 3.1.53
  * Made export of `_malloc` explicit
* Added CI
* Adjusted `README_WASM.md`
* Removed commit 924c5e58c524bb8f5a70407370079ec5d88073de which touched delay UGens
* Modularized scysnth wasm build (use `await createScsynth()` instead of a vague `Module` global variable - this also allows for multiple scsynth instances on one website)
* Replaced port simulation within `SC_WebInPort` with a dedicated `SC_WasmBinding` class
* Revamped JS code to make more use of ES6 features
* Added native OSC message builder and parser for wasm using the already included `oscpack`, and therefore removed the formerly included [osc.js library](https://github.com/colinbdclark/osc.js)

## Questions

* Should `SC_OscWasmBuilder.cpp` etc. be wrapped with `#ifdef __EMSCRIPTEN__` from top to bottom or is it sufficient that these files are only build for emscripten via CMake? Generally: Should these be `cpp` files or `hpp`? I decided to not separate declaration and implementation b/c these files is not referenced elsewhere.
* How much OSC functionality do we actually need from the JS exposed OSC constructor/parser?

## Next steps

* wasm build of sclang......
[...]

PR #6615, opened 2025-02-10
Add WebSocket support for sclang

capital-G:
## Purpose and Motivation

This PR proposes a WebSocket server and client implementation via [boost beast](https://github.com/boostorg/beast) by using language primitives.
WebSockets allow for direct communication with e.g. a Browser and is also more reliable when working with NAT-routed networks because it is using a TCP connection. This started as a side-project of #6569 in order to debug it more properly, but I think this feature could be of general interest.

The C++ implementation should be quite compact (ignore the new boost files while reviewing^^) and is split into a boost beast section (`SC_WebSocket`) and a primitve section (`SC_WebSocketPrim`). 

I would be happy if someone
* could give this some testing and
* review the proper use of passing around things in C++. I also don't fully understand how to manage resources within sclang, therefore I'd appreciate if someone with some experience could chime in here.

I also tried to implement a `WebSocketNetAddr` such that one could connect to a remote server using the WebSocket connection, but I failed to decipher the `ServerStatusWatcher` implementation. Therefore I propose to implement this in another PR.

## Types of changes

- Documentation
- New feature

## To-do list

- [ ] Code is tested
- [x] All tests are passing
- [x] Updated documentation

Merge notes: Can be squashed


lang/LangPrimSource/SC_WebSocket.h
```diff
@@ -0,0 +1,156 @@
...
+    WebSocketClient* m_self;
+
+    WebSocketClient(boost::asio::io_context& ioContext);
+
+    // why is port a string?!
+    void run(const std::string& host_, std::string& port);
+
+    beast::error_code closeConnection();
+
+    void enqueueMessage(WebSocketData message);
```
JordanHendersonMusic, line 137: This makes a copy. 

Did you mean `const WebSocketData&`?

If not, you can avoid the copy by calling it like...
```c++
web_socket_client.enqueneMessage(std::move(web_socket_data));
```

... but if that is the intended use case, it is often better to declare the function 
```c++
void enqueueMessage(WebSocketData&& message);
```

lang/LangPrimSource/SC_WebSocketPrim.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include "PyrInterpreter.h"
+#include "SC_WorldOptions.h"
+#include "PyrSymbolTable.h"
+
+#include "SC_WebSocketPrim.h"
+#include "SC_WebSocket.h"
+
+using namespace SC_Websocket_Lang;
+
+std::vector<uint8_t> getByteVector(VMGlobals* g, int& error) {
```
JordanHendersonMusic, line 16: Drop error code and return an optional.

I think the signature should look like this...
```c++
std::optoinal<std::vector<uint8_t>> getByteVector(PyrSlot* slot);
```
this way you pass the slot directly rather than getting it from the stack.

lang/LangPrimSource/SC_WebSocketPrim.cpp
```diff
@@ -0,0 +1,336 @@
...
+        scprintf("Raw message must be an Int8Array\n");
+        error = errBadPrimitive;
+        return std::vector<uint8_t>();
+    };
+    auto sclangByteArray = slotRawInt8Array(g->sp);
+    auto data = std::vector<uint8_t>(sclangByteArray->b, sclangByteArray->b + sclangByteArray->size);
+    return data;
+}
+
+std::string getStringMessage(VMGlobals* g, int& error) {
```
JordanHendersonMusic, line 27: Why is this needed? Could you not Just use `slotStrStdStrVal` directly?

lang/LangPrimSource/SC_WebSocketPrim.h
```diff
@@ -0,0 +1,103 @@
...
+PyrObject* convertMessage(std::vector<uint8_t>& message);
+
+class WebSocketServer {
+public:
+    static int start(VMGlobals* g, int numArgsPushed);
+
+    static int stop(VMGlobals* g, int numArgsPushed);
+
+private:
+    enum SLOT_OFFSET {
```
JordanHendersonMusic, line 27: I think it would be better to do what we already do elsewhere and write a struct like this..

```c++
struct SC_WebSocketServer: PyrObjectHdr {
   static constexpr auto className = "WebSocketServer";
   PyrSlot* host;
   PyrSlot* port:
   PyrSlot* pointer:
};
```
Then reinterpret cast the slot to `SC_WebSocketServer`, once you have checked the class name symbol.

Then all the methods can live inside that.


Similar for the other classes.

lang/LangPrimSource/SC_WebSocketPrim.h
```diff
@@ -0,0 +1,103 @@
...
+class WebSocketSession;
+class WebSocketClient;
+typedef std::variant<std::vector<uint8_t>, std::string> WebSocketData;
+}
+
+// namespace to connect the boost beast websocket implementation
+// from `SC_WebSocket` to sclang primitives.
+namespace SC_Websocket_Lang {
+
+PyrObject* convertMessage(std::string& message);
```
JordanHendersonMusic, line 17: These should take `const` string and vector references

lang/LangPrimSource/SC_WebSocketPrim.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+void WebSocketConnection::receiveLangMessage(SC_Websocket::WebSocketSession* session,
+                                             SC_Websocket::WebSocketData& message) {
+    VMGlobals* g = gMainVMGlobals;
+    gLangMutex.lock();
+    SetObject(++g->sp, getsym("WebSocketConnection")->u.classobj);
+    SetPtr(++g->sp, session);
+
+    if (std::holds_alternative<std::string>(message)) {
```
JordanHendersonMusic, line 179: I think you should use `std::visit` here

lang/LangPrimSource/SC_WebSocketPrim.cpp
```diff
@@ -0,0 +1,336 @@
...
+    VMGlobals* g = gMainVMGlobals;
+    gLangMutex.lock();
+    SetObject(++g->sp, getsym("WebSocketClient")->u.classobj);
+    SetPtr(++g->sp, client);
+    SetBool(++g->sp, connected);
+    runInterpreter(g, getsym("prSetConnectionStatus"), 3);
+    gLangMutex.unlock();
+}
+
+PyrObject* WebSocketClient::getObject(VMGlobals* g) { return slotRawObject(g->sp); }
```
JordanHendersonMusic, line 303: Could you delete this and just call it directly?

JordanHendersonMusic (commented):
I like this addition, I assume it is used to communicate with a server running in a browser? Could you just talk about how that will work?


I've left a few comments here. Generally I think the primitives, casting to and from slots, and how the stack is handle could be clearer. I've given some suggestion of how I'd do this.

capital-G:
Thanks for the review, much appreciated - will get through it the following days.

> I like this addition, I assume it is used to communicate with a server running in a browser? Could you just talk about how that will work?

Indeed - this turned from a "i need this to debug my other PR #6569" to a full-fledged feature which would be interesting beyond that.
Basically this feature would allow to control local or remote scsynth wasm instances, which makes testing the functionality much easier.
Currently this is not possible b/c there is some obscure behavior happening in `ServerStatusWatcher` that I can't get my head around currently - and as this seems more delicate I thought I'll tackle this in another PR after this gets merged.

I don't know if we instead should go for a "rawer" approach by exposing POSIX sockets, but this would be even more complicated?

dyfer:
FWIW I would like to see this addition getting merged. I can't comment on the cpp part, my only concern being that we are "bloating" sclang. But there's no other solution to get new functionality in, as we don't have a plugin system for the language.

I would love to try to make a version of https://github.com/dyfer/WsGUI working with this implementation instead of the currently used node.js websockets bridge. However, WsGUI also needs a simple HTTP server, so I guess I'd still need to use node or python for that. TBC...

capital-G:
I am somehow unsure if we should go for such functionality b/c the backwards compatibility promise in "userspace" creates some friction for me and is the reason why I essentially don't contribute to sclang-space. This is especially difficult for this implementation as I want to be able to modify or even throw this out anytime b/c this can introduce security issues due to the possibility to exposing sclang to the web - therefore it doesn't seem to fit for sclang and is contrary to my believe that we should keep core-sclang as minimal as possible in order to keep maintenance viable (I think this is one of the best things in the mindset of Hadron). Also WebSocket are somehow legacy already as well, and we would therefore need to remove it at some point.

A more proper/generic implementation would be to introduce posix sockets to sclang instead of implementing websockets as a primitive. Although the windows support makes this implementation more work, but we could use boost though -> https://original.boost.org/doc/libs/1_85_0/doc/html/boost_asio/overview/networking/bsd_sockets.html and https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers

Or an even more generic solution: Implement a FFI for sclang and create a C-Plugin interface ;) Then we could implement this outside, independent of SC dev cycles and outside of any API promises :)

sclang is lacking any kind of network capabilities beyond OSC, and OSC via UDP is not really feasible for todays network infrastructure. The network stack of sclang really needs some care and I think sclang's lack of internet-communication capabilities is one of the reasons why I see some users leave sc for other projects.

Maybe I'll just clean this up in order to learn something here, but maybe we could discuss this PR at a dev meeting?

JordanHendersonMusic:
> Or an even more generic solution: Implement a FFI for sclang and create a C-Plugin interface ;) Then we could implement this outside, independent of SC dev cycles and outside of any API promises :)

It would be really useful to know exactly what features the FFI needs to make this work —how minimal could it be?

JordanHendersonMusic:
@capital-G what is the status of this? It seems like a good idea?

capital-G:
> @capital-G what is the status of this? It seems like a good idea?

I think the FFI approach would be rather interesting - though I know that @lnihlen has also some ideas on how to extend the runtime w/ external libraries - maybe it is worth to join efforts here b/c I don't want to write the beast stuff two times :D

capital-G:
This PR became obsolete b/c w/ #7189 the code of this PR has been possible to be outsourced to a dedicated repo https://github.com/capital-G/sclang_websocket.
This is preferable as it allows to keep "core sclang" smaller which is preferable in regards to maintenance.

Issue #6610, opened 2025-02-07
update portaudio to last


sonoro1234:
Hello,

portaudio_submodule is 4 years old. It would be good to update it. Specifically for ASIO it now downloads ASIO SDK and makes a small patch nedded.

The process would be to update portaudio in portaudio_submodule and add PA_USE_ASIO in CMakeLists.txt in portaudio folder.

I could do it if you wish.

capital-G:
Sounds like a good idea to keep our dependencies up-to-date.

> I could do it if you wish.

Would be really appreciated! :) I can hopefully do a review of the PR.

PR #6616, opened 2025-02-11
update portaudio_submodule to last master branch

sonoro1234:
## Purpose and Motivation

Update to last portaudio

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

Due to changes in last portaudio `PA_USE_ASIO=ON` must be explicitly set in case ASIO is wanted. (which is a breaking change for ASIO building) see: https://github.com/PortAudio/portaudio/blob/master/CMakeLists.txt#L184

The advantages are: get new additions to portaudio and if CMake version is not less than 3.18 asiosdk is downloaded by cmake and patched as explained in https://github.com/PortAudio/portaudio/pull/997

A way to avoid any surprise caused by the breaking change could be adding `PA_USE_ASIO=ON` in CMakelists.txt when we are on windows, so that portaudio is always built with ASIO without any requirement for the user.

Portaudio is used on windows but can also be used in macos for supernova (without ASIO). I could not test that and I dont even know if it worked previously.

### DIFF
--- a/README_WINDOWS.md
+++ b/README_WINDOWS.md
@@ -304,7 +304,10 @@ following folder structure *exactly*:
             lib
         fftw
 
-In order to get support for ASIO drivers, follow this directory structure:
+In order to get support for ASIO drivers:
+PA_USE_ASIO must be set in cmake command `-DPA_USE_ASIO=ON`
+
+If cmake version is less than 3.18, you must also follow this directory structure:
 
     supercollider
         external_libraries
@@ -315,6 +318,7 @@ In order to get support for ASIO drivers, follow this directory structure:
                     common
                     ...
             ...
+Otherwise (cmake not less than 3.18) this is not needed.
 
 FFTW does not provide build files for Visual Studio. In the **Developer Command
 Prompt for VS2019** (or VS2017; note that this is not `cmd.exe`), `cd` to the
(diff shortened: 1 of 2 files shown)


Spacechild1:
I don't think it's a good idea to upgrade portaudio to some experimental state (= the current master branch), instead we should wait for the next release. (To be honest, I don't really understand why they haven't made a new release since 2021 although development is quite active...)

sonoro1234:
I have been confused by #6610

capital-G:
> I have been confused by #6610 ASIO building improvement made me rush. And I am used to rolling releases.

Sorry for green-lighting this earlier but I have to agree with @Spacechild1 here - SC has an emphasis on stability and we should always rely on stable and tested dependencies. I didn't know there hasn't been any release so far.

Do you know if there is a reason why they haven't released any new version? Maybe they can release a new major version? Even [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/portaudio/-/commits/main) does not use a newer version than the latest stable.

Is it OK to close this PR and re-open this once a new major version of portaudio has been released?

dyfer:
> Do you know if there is a reason why they haven't released any new version?

Historically [PortAudio releases happened quite infrequently](https://files.portaudio.com/download.html). PA team moved development to github a few years back and there has been only one stable release since then. It seems that a new release is "right around the corner", but that may as well mean another 1-2 years (sooner hopefully!), considering the past record...

PR #6620, opened 2025-02-14
class library: MIDIClient init cleans up old connections

### LINKED ISSUE #4495: MIDIClient.init() continually accumulates new ports
## Environment

* SuperCollider version: 3.10.2
* Operating system: openSUSE Leap 15.0

## Steps to reproduce

Call `MIDIClient.init` repeatedly, and you'll see the list of inputs and outputs continually increasing; at least on Linux it happens in a rather odd fashion as described below.

IIUC this is at least partially due to the following code which ignores how many input and output ports have already been created:

```supercollider
		if(inports.isNil,{inports = sources.size});
		if(outports.isNil,{outports = destinations.size});
		// this.disposeClient;

		this.prInit(inports,outports);
```

A contributory factor is that on Linux, [`sources` and `destinations` include SuperCollider's own ports](http://doc.sccode.org/Classes/MIDIClient.html#*externalSources), thereby setting up a kind of positive feedback loop.

## Actual behavior

For example, if I first do `MIDIClient.init(0, 0)` then internally it calls `this.prInit(0, 0)` resulting in this:

```
MIDI Sources:
	MIDIEndPoint("System", "Timer")
	MIDIEndPoint("System", "Announce")
	MIDIEndPoint("Midi Through", "Midi Through Port-0")
	MIDIEndPoint("iRig KEYS 37", "iRig KEYS 37 MIDI 1")
	MIDIEndPoint("SuperCollider", "out0")
MIDI Destinations:
	MIDIEndPoint("Midi Through", "Midi Through Port-0")
	MIDIEndPoint("iRig KEYS 37", "iRig KEYS 37 MIDI 1")
	MIDIEndPoint("SuperCollider", "in0")
```

Now (because I'm on Linux) it considers there to be 5 sources and 3 destinations, so if I do `MIDIClient.init()` then internally it calls `this.prInit(5, 3)` resulting in 5 `inports` and 3 `outports`:

```
port created: SuperCollider:midi/capture_1
port created: SuperCollider:midi/playback_2
MIDI Sources:
	MIDIEndPoint("System", "Timer")
	MIDIEndPoint("System", "Announce")
	MIDIEndPoint("Midi Through", "Midi Through Port-0")
	MIDIEndPoint("iRig KEYS 37", "iRig KEYS 37 MIDI 1")
	MIDIEndPoint("SuperCollider", "out0")
	MIDIEndPoint("SuperCollider", "out1")
	MIDIEndPoint("SuperCollider", "out2")
MIDI Destinations:
	MIDIEndPoint("Midi Through", "Midi Through Port-0")
	MIDIEndPoint("iRig KEYS 37", "iRig KEYS 37 MIDI 1")
	MIDIEndPoint("SuperCollider", "in0")
	MIDIEndPoint("SuperCollider", "in1")
	MIDIEndPoint("SuperCollider", "in2")
	MIDIEndPoint("SuperCollider", "in3")
	MIDIEndPoint("SuperCollider", "in4")
```

Now there are 7 sources and 7 destinations, so if I call it again then it will internally call `this.prInit(7, 7)`, then  `this.prInit(11, 9)`, then  `this.prInit(13, 13)`, and so on.

Technically it could be argued that this *is* the documented behaviour, since [the docs](http://doc.sccode.org/Classes/MIDIClient.html#*init) say:

> **`MIDIClient.init(inports, outports, verbose: true)`**
>
> Initializes the MIDIClient, checks which available MIDI sources and destinations there are, and opens as many connections as desired.
>
> Arguments:
> - `inports` - the number of MIDI input connections to open; if `nil` then opens as many `inports` as there are MIDI sources.
> - `outports` - the number of MIDI output connections to open; if `nil` then opens as many `outports` as there are MIDI destinations.
> - `verbose` - A flag whether or not to post the MIDI sources and destinations that were found. Default is `true`.

and the pedants amongst us may note that nowhere in these docs does it promise to *delete* any input / output ports, only to open new ones.  Nor does it state whether the sources and destinations include pre-existing internal SuperCollider ports.  So the net effect is to continually accumulate new ports in this weird fashion.

## Workaround

This can be avoided by specifying a non-`nil` number of `inports` and `outports`, e.g. `MIDIClient.init(2, 2)`; however I suggest that the current behaviour violates the Principle of Least Surprise.

## Expected behaviour

After `MIDIClient.init` has already been called once, if `inports` and/or `outports` are not specified in subsequent calls, the resulting number of ports only increases if required in order to match the number of *external* MIDI sources / destinations (e.g. if new MIDI devices were plugged in since it was called the first time).

## A related documentation nitpick

Being pedantic again, it seems to me that the terminology in the existing docs for this method is slightly confused due to treating the concepts of "port" and "connection" as synonymous when in fact a connection involves a pair of ports.  As a corollary, I think that the corresponding verb should not be "open" but rather "create".  When combined with the expected behaviour described above, I would suggest that this should eventually be changed to read:

> - `inports` - the number of MIDI input ports SuperCollider should provide; if `nil` then ensures as many input ports as there are external MIDI sources, so that each external source can be consumed as an input into SuperCollider.
> - `outports` - the number of MIDI output ports SuperCollider should provide; if `nil` then ensures as many output ports as there are external MIDI destinations, so that each external destination can be sent an output from SuperCollider.

Does this make sense?

If so, it's worth noting that one side-effect of this behaviour would be to ensure that after calling `MIDIClient.init` with no arguments, the number of input ports (both internal and external combined) and output ports (ditto) would be equal, since:

- `inports` == `MIDIClient.externalSources.size`
- `outports` == `MIDIClient.externalDestinations.size`
- combined number of ports which can receive input == `inports` + number of external destinations
- combined number of ports which can generate output == `outports` + number of external sources

telephon:
## Purpose and Motivation

For those who forgot to clean up MIDIClient before calling init several times, this caused an issue. This does it automatically. Fixes #4495.

## To-do list

- [ ] Code is tested
- [ ] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Control/MIDIOut.sc
+++ b/SCClassLibrary/Common/Control/MIDIOut.sc
@@ -15,7 +15,7 @@ MIDIClient {
 	classvar <initialized=false;
 	*init { arg inports, outports, verbose=true; // by default initialize all available ports
 								// you still must connect to them using MIDIIn.connect
-
+		this.disposeClient; // remove old sources and destinations, in case init is called repeatedly
 		this.prInitClient;
 		this.list;
 		if(inports.isNil,{inports = sources.size});


dyfer (approved):
I can't test on Linux ATM, where this issues supposedly occurs.

Doesn't seem to break anything on macOS.

I think this can be merged as is. Thanks!

Issue #6496, opened 2024-09-24
Caching of docs lead to CSS errors when downgrading SuperCollider


capital-G:
## Environment

* SuperCollider version: 3.13.0
* Operating system: macOS

## Steps to reproduce

* Have SuperCollider 3.13 (or non-dev) installed in `/Applications`
* Build dev (or in the future: install 3.14 side-by-side to 3.13) and launch the IDE
* Close dev/3.14 and open the existing 3.13 again, which now looks like

<img width="1532" alt="image" src="https://github.com/user-attachments/assets/3f79231a-4942-42e1-af14-401addfc1cda">

## Expected vs. actual behavior

There should be no displaying errors of the docs when multiple SuperCollider versions are installed side-by-side.

## Plans to fix this bug

The resulting displaying errors are due to a change in our CSS structure introduced in https://github.com/supercollider/supercollider/pull/6095 

Although the `SCDoc.version` has been bumped

* dev: https://github.com/supercollider/supercollider/blob/57615d160a876dd7b78e6e8f259b471ca6487de9/SCClassLibrary/SCDoc/SCDoc.sc#L384-L387 
* 3.13 https://github.com/supercollider/supercollider/blob/c64c5244c5e7bbcb26a829e9cf72d249a529c213/SCClassLibrary/SCDoc/SCDoc.sc#L384-L387

the non-help files in the `Help` dir only gets replaced if the mtime (modification time) of the file is older than in `HelpSource`, therefore ignoring any kind of version mismatch.  

https://github.com/supercollider/supercollider/blob/57615d160a876dd7b78e6e8f259b471ca6487de9/SCClassLibrary/SCDoc/SCDoc.sc#L506-L515

3.13 file
```shell
❯ ls -alh /Applications/SuperCollider.app/Contents/Resources/HelpSource/scdoc.js

-rw-r--r--@ 1 scheiba  admin   6.8K Feb 20  2023 scdoc.js
```

vs in `Help` dir which has been updated by the dev version

```shell
❯ ls -alh ~/Library/Application\ Support/SuperCollider/Help/scdoc.js
-rw-r--r--@ 1 scheiba  staff   6.8K Sep 24 13:29 scdoc.js
```

As the `scdoc.js` mtime of `3.14` is newer than `3.13`, this will not be replaced when booting into 3.13.

[...]

PR #6207, opened 2024-01-30
added line numbers to code mirror of the help document

prko:
## Purpose and Motivation

Reading long code examples in help documents is not as straightforward as in SC-IDE's code editor because there are no line numbers on the left side of the code mirror in help documents.
This PR adds line numbers to the code mirror of the help document.
 
https://scsynth.org/t/implementing-line-numbers-on-the-left-side-of-the-code-editor-in-schelp-files/5986

Here are some examples:
Screenshot 2024-01-31 at 3 47 12
Screenshot 2024-01-31 at 3 46 45
Screenshot 2024-01-31 at 3 46 19
Screenshot 2024-01-31 at 18 09 13

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] This PR is ready for review

PR #6627, opened 2025-02-18
Fix SCDocs assets caching

capital-G:
Merge note: should be squashed

Fixes https://github.com/supercollider/supercollider/issues/6496

## Purpose and Motivation

Currently the caching of static files does not react to a mismatch of `scdoc_version` file but is invalidated by comparing the `mtime` of the static files.
As https://github.com/supercollider/supercollider/pull/6095 changed some static files this will now create problems if a `develop` build is launched (which will replace the static files in the HelpTargetDir as their mtime is newer) and afterwards an existing SuperCollider installation of `3.13` is launched: because the static files are considered "up-to-date" by `3.13` (the mtime of the `HelpSource` file is lower than the one in the `HelpTargetDir`) they will not be copied/replaced, although they are not matching the actual `SCDoc.version` - this create some displaying problems (see #6496).

This PR ties the static files to `SCDoc.version`, so in case we want to update any static files we can do so by incrementing `SCDoc.version` which will invalidate now all files on a version-mismatch (and not just `schelp` files).

Because we can not change the caching-code in older SuperCollider instances (we want SC `<=3.13` to not consider any static files from a newer version valid) it was also necessary to put all static files into a new directory so that any new static files changed via #6095 will not be considered because they are at a different location.
Doing this also helps to clean up the directory though :)

> This will still yield bugs because files like `Browse.html` are still considered "up-to-date" by `<=3.13`, yet the displaying of the help file is not immensly broken (it will be displayed with the newer css theme applied).
> This could only be fixed by changing the locations of `Browse.html` etc. which I consider too drastic, but I am open to input from other users here!

Anyone who had run an build between #6095 and this PR should consider deleting `SCDoc.helpTargetDir` after merging of this PR if switching between `develop` and `stable` is wanted in order to let `<=3.13` copy/use their own static files .

edit: this PR also fixes future modifications of static files

---
I am not too happy with the implementation b/c this introduces side effects to `*checkVersion`, yet I found it the most clean way b/c it is a bit hard to reason about which entrypoints exist to build the docs.
This should catch any changes in `SCDoc.version` at its core.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


prko:
Thanks for this, I am satisfied with the theme switcher implemented by you.
Is there a possibility of including the idea or implementation in https://github.com/supercollider/supercollider/pull/6207#issuecomment-2148536975?

capital-G:
@tedmoore can you take a look at this? it is not the most fun PR, but it holds me back to using a dev build more often :)

tedmoore:
Sorry I haven't had a chance to look closely at this yet (I'm a bit short on time currently). Ahead of the meeting tomorrow I figured I would at least respond to you (unfortunately I can't be a the meeting though)! What I think would be useful is to chat with you sometime so you can walk me through the issue more closely. It's been a couple years since I was intimate with the SCDocs rendering system and I don't think I had trouble with versioning docs. It sounds to me like a more drastic solution might be a better path forward.

capital-G:
> What I think would be useful is to chat with you sometime so you can walk me through the issue more closely.

Sure - hmu in the forum chat (dscheiba) and we can figure out a meeting where I can walk you through the PR.

HelpSource/Overviews/Documents.html
```diff
@@ -4,18 +4,18 @@
...
-    <link rel='stylesheet' href='../scdoc.css' type='text/css' />
-    <link rel='stylesheet' id='scdoc-theme' href='../themes/default.css' type='text/css' />
-    <link rel='stylesheet' href='../frontend.css' type='text/css' />
-    <link rel='stylesheet' href='../custom.css' type='text/css' />
-    <script src="../lib/jquery.min.js"></script>
+    <link rel='stylesheet' href='./../static/scdoc.css' type='text/css' />
+    <link rel='stylesheet' id='scdoc-theme' href='./../static/themes/default.css' type='text/css' />
+    <link rel='stylesheet' href='./../frontend.css' type='text/css' />
+    <link rel='stylesheet' href='./../custom.css' type='text/css' />
+    <script src="./../staticc/lib/jquery.min.js"></script>
```
capital-G, line 11: probably not call this `staticc`?

capital-G:
I'll rebase for the review

HelpSource/Overviews/Documents.html
tedmoore, line 11: This `static` has two `c`s? I believe it's a typo.

tedmoore: Oh I see you've found it too!

tedmoore (changes requested):
Just the one type in `Documents.html` where there is an extra c in "static"

capital-G:
> Just the one type in `Documents.html` where there is an extra c in "static"

Fixed that and rebased.

If everything is fine by you, feel free to squash-merge it :)

tedmoore (approved):

prko:
Today, I found #6859 and #6860 while testing #6854.
Are they related to this?

On macOS, I have been using a build from last summer and have been manually updating the necessary parts from some commits. However, I could not do it more with some recent commits, and I think this could be related to this PR. Am I correct?

Issue #6304, opened 2024-05-09
[Request] SCDoc modal tag: post


prko:
## Motivation
There is currently no built-in way to display coloured text as in the Post window when including the Post window return in Help documents. The only way is to use images, which is unproductive. 

## Description of Proposed Feature
The block code shows the colour of the code:
```
code::
SinOsc.ar
::
```
Like this, it would be good to have a `post` tag.

## Plan for Implementation

Unfortunately I do not know C++. If someone gives me instructions I could try it, otherwise we should wait for a developer who feels the same need.

telephon:
The coloring in the post window is currently quite inconsistent. So we need to fix that first. When you post something that has several lines, it will highlight the first line only, because it starts with a "->". I also find it inconvenient that the "->" is appended, because when you copy the code, you have to get rid of it in an extra step.

capital-G:
I really like this idea, and IMO it would be a good way to add an even more general tag `div`, which the current documentation lacks. Due to this lack it makes it difficult/impossible to create truly new designs for the docs as the placement and look of individual blocks can not be accessed or changed in a consistent manner.

If we manage to introduce a kind of div wrapper with a `class` variable, any kind of visual or interactive could be handled in traditional css or js ways as it seems less a concern of code generation from within scdocs but an enhancement of already generated output.

Myst has something like this as well: https://myst-parser.readthedocs.io/en/v0.17.1/syntax/optional.html#syntax-admonitions

prko:
@telephon 
> ... So we need to fix that first. ....

Yes, even though I am used to it, the inconsistency sometimes confuses me. If the order of development is as you say, there is no other way but to wait...

---

@capital-G 
>IMO it would be a good way to add an even more general div tag, which the current documentation lacks.

The easiest way to do this is to enable the necessary HTML tags directly in SCDoc, without having to create SCDoc modal tags, which is done in cooperation between C++ (`SCDoc.y`, `SCDoc.l`, etc) and SCClasslibrary (`SCDoc.sc` and `SCDocRenderer.sc`) as in my recent PR:

[SCDoc improvements: 3. Some HTML tags are enabled](#6265)

However, I think none of the developers would see this as positive, as the SCDoc syntax will be ruined....

My PR #6256 may be based on my amateur experience of programming skills, but it is flexible. Also, considering that sclang is not the only syntax style, the direct use of HTML in SCDoc does not seem so bad.... Also, considering the speed of maintenance and development of SC..., hm...

[...]

PR #6630, opened 2025-02-20
Fix table width to improve readability

prko:
## Purpose and Motivation

Currently, the width of the table depends on the length of the text in the table. This is fine if there is only one table.
However, multiple tables with different table widths reduce readability.
Compare the following two examples:
Screenshot 2025-02-21 at 01 18 15.
Screenshot 2025-02-21 at 01 37 36.
To me, the inconsistent table width seems to be a bug.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/scdoc.css
+++ b/HelpSource/scdoc.css
@@ -30,8 +30,9 @@ div.contents {
 table {
     border-collapse: collapse;
     font-size: 10pt;
-    margin-top: 1em;
-    margin-left: 2em;
+    margin: 0.7em 0;
+    width: 100%;
+    table-layout: auto;
 }
 table td {
     border: 1px solid var(--color-fg-200);


capital-G (commented):
Code and aesthetic looks fine to me, thanks.
Have you checked if this affects other tables in the docs, e.g.

https://dev.docs.supercollider.online/Classes/Server.html
image

or https://dev.docs.supercollider.online/Classes/SendTrig.html

image

Can you provide some screenshots of these instances with the fix applied?

In these cases the compact layout actually seems beneficial as it follows the "left-bound" reading.

prko:
Thanks for your feedback. Here are the screenshots:
- SendTrig: .dumpOSC
Screenshot 2025-02-23 at 19 21 38
- Server: Description
Screenshot 2025-02-23 at 19 22 54

**I agree that the current way is good for some cases, especially the ones you mention. The current way is only not ideal when many tables are on the screen or a page.**

My fix automatically resizes the table width to match the window width.

To get the best result, we could do one of the following:
- Let the user decide how to fix the table. See the previous discussion: 
   - #6265 
   - #6304 
- Let the table width be automatically resized in two ways:
   - resize the table width to AA% (e.g.: 65%.) of the 'window width' (or page width) if the table width does not exceed AA% of the window width, or
   - resize the table width to 95% of the 'window width' (or page width) if the table width exceeds AA% of the window width.

capital-G (approved):
Thanks for the screenshots - it looks much better than expected.

To me this is fine as-is, but let's wait for another pair of eyes on it.

Your mentioned extensions are also interesting, but I think this could be another PR.

PR #6640, opened 2025-02-24
Use boost program options for sclang CLI options

capital-G:
## Purpose and Motivation

This replaces the current CLI options parsing in sclang with boost program options.

This allows to
* use multi-character options (e.g. `--standalone` is now as valid as `-a` but is probably easier to remember?)
* automatic generation of help command
* overwrite language configuration settings configured via YAML. Allows to set e.g. included directories or else w/o the need to create a YAML file to set such parameters. This probably makes it easier to provide a "run sclang testsuite" and "build docs" command within cmake or via a shell script.
* replaces custom written code with boost library code which hopefully reduces maintenance

While at it, I also transformed some `char*` to `std::string` and `char**` to `std::vector<std::string>>` where they are not within the realm of the sclang managed memory.

`--help` now returns this

```shell
❯ ./sclang --help
Start a REPL interpreter:           sclang [ -options ]
Execute a .scd file:                sclang [ -options ] path_to_sc_file [ args passed to sclang... ]

Generic options:
  -h [ --help ]                     Print this message and exit
  -v [ --version ]                  Print sclang version and exit

sclang options:
  -d [ --runtime-directory ] arg    Set runtime directory
  -D [ --daemon ]                   Enter daemon mode
  -g [ --heap-growth ] arg (=256k)  Set heap growth size (allows k, m and g as 
                                    suffix multipliers)
  -l [ --yaml-config ] arg          Set yaml config
  -m [ --heap-size ] arg (=2m)      Set initial heap size (allows k, m and g as
                                    suffix multipliers)
  -r [ --call-run ]                 Call Main.run on startup
  -s [ --call-stop ]                Call Main.stop on shutdown
  -u [ --port ] arg (=57120)        Set UDP listening port
  -i [ --ide ] arg                  Set IDE name
  -a [ --standalone ]               Standalone mode (exclude SCClassLibrary and
                                    user and system Extensions folders from 
                                    search path)

Language config options (overwrites provided yaml entries):
  --include-path arg                class library path to be included for 
                                    searching (allows multiple mentions)
  --exclude-path arg                class library path to be to excluded from 
                                    searching (overrides includePaths, allows 
                                    multiple mentions)
  --post-inline-warnings            Boolean flag to turn on warnings about 
                                    missing inline opportunities.
  
```

This is more an exercise to get familiar with the sclang code than a necessary feature, yet I think it is still an improvement over the current code basis.
Maybe this could be a start to transform some "older" sclang parts into more modern C++ where there is no need for manual memory management so they become more robust and easier to maintain?

I think the `notifier` syntax of program options allows for really compact code.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

Although I tested it on my machine, it wouldn't hurt if someone else give this some testing.


capital-G:
I introduced and fixed a bug regarding the config file creation, but from my perspective everything is now working fine.

This PR would allow us to make testing and doc creation easier as it introduces a way to explicitly set language config parameters instead of relying on setting them through a yaml file.

lang/LangSource/SC_CLIOptions.cpp
```diff
@@ -0,0 +1,277 @@
...
+    // for the language config we don't use the notifier of program options because
+    // we want to fill the config with a provided yaml file or their default values.
+    // as po does not allow to specify the "notify order", we parse the config manually
+    parse_language_config(vm);
+
+
+    return false;
+}
+
+po::options_description CLIOptions::build_generic_description() {
```
Spacechild1, line 97: `CLIOptions::build_*` methods are really pure functions and as such they should either be `static` methods or free functions! Personally, I tend to prefer the latter.

lang/LangSource/SC_CLIOptions.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+po::options_description CLIOptions::build_language_config_description() {
+    po::options_description description("Language config options (overwrites provided yaml entries)");
+
+    // clang-format off
+    description.add_options()
+        (
+            "include-path",
+            po::value<vector<string>>(),
+            "class library path to be included for searching (allows multiple mentions)"
```
Spacechild1, line 187: "class library path" should start with a capital letter for consistency :monocle_face:

lang/LangSource/SC_CLIOptions.hpp
```diff
@@ -0,0 +1,31 @@
...
+using namespace std;
+
+namespace SC_CLI {
+class CLIOptions {
+public:
+    CLIOptions() {};
+    bool parse(int argc, char* argv[], SC_TerminalClient::Options& terminal_options, int& exit_code);
+
+    string mInputFile = "";
+    vector<string> mScArgs = {};
```
Spacechild1, line 15: Redundant initializers! Should be:
```
string mInputFile;
vector<string> mScArgs;
```

lang/LangSource/SC_LanguageClient.cpp
```diff
@@ -99,8 +99,9 @@ void SC_LanguageClient::initRuntime(const Options& opt) {
     // start virtual machine
     if (!mHiddenClient->mRunning) {
         mHiddenClient->mRunning = true;
-        if (opt.mRuntimeDir) {
-            int err = chdir(opt.mRuntimeDir);
+        if (opt.mRuntimeDir.data() != nullptr) {
```
Spacechild1, line 102: If the string is empty, the return value of `data()` is undefined! I guess you meant to write
`if (!opt.mRuntimeDir.empty()) {`

lang/LangSource/SC_LanguageClient.cpp
```diff
@@ -99,8 +99,9 @@ void SC_LanguageClient::initRuntime(const Options& opt) {
...
+            // int err = chdir(opt.mRuntimeDir);
+            int err = 0;
```
Spacechild1, line 104: This needs a comment. Why did you remove the `chdir` call? If you remove it, then you should also remove the subsequent error check.

If there is a good reason not to switch the current working directory, then you should put the whole `if` clause in an `#if 0 ` block (+ add a comment why the code is disabled)

lang/LangSource/SC_LanguageConfig.cpp
```diff
@@ -38,12 +38,12 @@ bool SC_LanguageConfig::gPostInlineWarnings = false;
...
-static const char* EXCLUDE_PATHS = "excludePaths";
-static const char* POST_INLINE_WARNINGS = "postInlineWarnings";
-static const char* CLASS_LIB_DIR_NAME = "SCClassLibrary";
-const char* SCLANG_YAML_CONFIG_FILENAME = "sclang_conf.yaml";
-static const char* EXCLUDE_DEFAULT_PATHS = "excludeDefaultPaths";
+static const std::string INCLUDE_PATHS = "includePaths";
+static const std::string EXCLUDE_PATHS = "excludePaths";
+static const std::string POST_INLINE_WARNINGS = "postInlineWarnings";
+static const std::string CLASS_LIB_DIR_NAME = "SCClassLibrary";
+const std::string SCLANG_YAML_CONFIG_FILENAME = "sclang_conf.yaml";
```
Spacechild1, line 45: missed a `static`.

Actually, you can just put the whole block in an anonymous namespace instead.

include/lang/SC_LanguageClient.h
```diff
@@ -40,17 +41,17 @@ SCLANG_DLLEXPORT void destroyLanguageClient(class SC_LanguageClient*);
...
         int mMemGrow; // memory growth in bytes
         int mPort; // network port number
-        char* mRuntimeDir; // runtime directory
+        std::string mRuntimeDir; // runtime directory
     };
 
 protected:
     // create singleton instance
-    SC_LanguageClient(const char* name);
+    SC_LanguageClient(const std::string name);
```
Spacechild1, line 54: Should be `const std::string&`

include/lang/SC_LanguageClient.h
```diff
@@ -92,7 +93,7 @@ class SCLANG_DLLEXPORT SC_LanguageClient {
     void runLibrary(const char* methodName);
     void interpretCmdLine();
     void interpretPrintCmdLine();
-    void executeFile(const char* fileName);
+    void executeFile(const std::string fileName);
```
Spacechild1, line 96: should be `const std::string&`

lang/LangSource/SC_LanguageClient.cpp
```diff
@@ -167,7 +168,7 @@ void SC_LanguageClient::runLibrary(const char* methodName) {
     unlock();
 }
 
-void SC_LanguageClient::executeFile(const char* fileName) {
+void SC_LanguageClient::executeFile(const std::string fileName) {
```
Spacechild1, line 171: `const std::string&`

lang/LangSource/SC_LanguageConfig.cpp
```diff
@@ -103,7 +103,7 @@ bool SC_LanguageConfig::removeIncludedDirectory(const Path& path) { return remov
 
 bool SC_LanguageConfig::removeExcludedDirectory(const Path& path) { return removePath(mExcludedDirectories, path); }
 
-static void processPathList(const char* nodeName, YAML::Node& doc,
+static void processPathList(const std::string nodeName, YAML::Node& doc,
```
Spacechild1, line 106: `const std::string&`

lang/LangSource/SC_LanguageConfig.cpp
```diff
@@ -117,14 +117,14 @@ static void processPathList(const char* nodeName, YAML::Node& doc,
     }
 }
 
-static void processBool(const char* nodeName, YAML::Node& doc, const std::function<void(bool)>& successFunc,
+static void processBool(const std::string nodeName, YAML::Node& doc, const std::function<void(bool)>& successFunc,
```
Spacechild1, line 120: `const std::string&`

lang/LangSource/SC_TerminalClient.cpp
```diff
@@ -71,7 +74,7 @@ static FILE* gPostDest = stdout;
 static UINT gOldCodePage; // for remembering the old codepage when we switch to UTF-8
 #endif
 
-SC_TerminalClient::SC_TerminalClient(const char* name):
+SC_TerminalClient::SC_TerminalClient(const std::string name):
```
Spacechild1, line 77: `const std::string&`

lang/LangSource/SC_TerminalClient.h
```diff
@@ -56,24 +56,22 @@ class SCLANG_DLLEXPORT SC_TerminalClient : public SC_LanguageClient {
...
         bool mCallRun;
         bool mCallStop;
         bool mStandalone;
-        int mArgc;
-        char** mArgv;
+        std::vector<std::string> mArgs;
     };
 
-    SC_TerminalClient(const char* name);
+    SC_TerminalClient(const std::string name);
```
Spacechild1, line 74: `const std::string&`

Spacechild1 (changes requested):

lang/LangSource/SC_CLIOptions.cpp
capital-G: Thanks for the comprehensive review, helps me a lot to learn these things (somehow forgot to pass the strings by reference...).

Concerning this comment tho: these methods need access to the CLIOptions, see e.g.
https://github.com/supercollider/supercollider/blob/6585c12efd2750853f91e709aaf472df743e8ca9/lang/LangSource/SC_CLIOptions.cpp#L97-L102
where `mPrintVersion` and `mPrintHelp` are member variables.

Where it was possible to keep them pure I used static, but I failed to come up with a way to make them static here. So, is it fine to use non-static methods here?

capital-G:
Hope I fixed most things except the two open questions?

I had to change the signature of

https://github.com/capital-G/supercollider/blame/4c11b5e76d00f76f97c72cf97d566922a78b5fab/include/lang/SC_LanguageClient.h#L54

Although this is in the include folder (and therefore exposed to other libs?) this shouldn't be a big problem b/c we don't have a stability contract regarding the LanguageClient interface, right?

If everything is done properly I'll try to give this a rebase so we have a proper git history for this.

lang/LangSource/SC_LanguageClient.cpp
capital-G: This was actually a comment during development - reintroduced the functionality, thx for spotting it!

lang/LangSource/SC_CLIOptions.cpp
Spacechild1: > where mPrintVersion and mPrintHelp are member variables.

A sorry, totally missed that! Nevermind then.

lang/LangSource/SC_LanguageConfig.cpp
Spacechild1: Ah, I see!

include/lang/SC_LanguageClient.h
```diff
@@ -40,17 +41,17 @@ SCLANG_DLLEXPORT void destroyLanguageClient(class SC_LanguageClient*);
 class SCLANG_DLLEXPORT SC_LanguageClient {
 public:
     struct Options {
-        Options(): mMemSpace(2 * 1024 * 1024), mMemGrow(256 * 1024), mPort(57120), mRuntimeDir(0) {}
+        Options(): mMemSpace(2 * 1024 * 1024), mMemGrow(256 * 1024), mPort(57120), mRuntimeDir("") {}
```
Spacechild1, line 44: No need to explicitly initialize `mRuntimeDir` because `std::string` has a default constructor.

Spacechild1:
> this shouldn't be a big problem b/c we don't have a stability contract regarding the LanguageClient interface, right?

Yes, the interface is only used by internal clients (the terminal client and QtCollider) so we are free to make changes.

lang/LangSource/SC_CLIOptions.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+namespace po = boost::program_options;
+using namespace std;
+
+namespace SC_CLI {
+
+/// parses the CLI - if the program should terminate after parsing it will return true
+/// an exit code is passed by reference
+bool CLIOptions::parse(int argc, char* argv[], SC_TerminalClient::Options& terminalOptions, int& exit_code) {
+    exit_code = 0;
```
Spacechild1, line 19: Nitpick: camelCase 🤓

lang/LangSource/SC_CLIOptions.cpp
```diff
@@ -0,0 +1,284 @@
...
+    description.add_options()
+        (
+            "include-path",
+            po::value<vector<string>>(),
+            "Class library path to be included for searching (allows multiple mentions)"
+        )
+        (
+            "exclude-path",
+            po::value<vector<string>>(),
+            "Class library path to be to excluded from searching (overrides includePaths, allows multiple mentions)"
```
Spacechild1, line 199: -> "Class library path to be excluded ..."

lang/LangSource/SC_CLIOptions.hpp
```diff
@@ -0,0 +1,31 @@
...
+    bool mPrintHelp = false;
+    bool mPrintVersion = false;
+
+    po::options_description buildGenericDescription();
+    po::options_description buildTerminalDescription(SC_TerminalClient::Options& terminalOptions);
+    static po::options_description buildLanguageConfigDescription();
+
+    static void parseLanguageConfig(po::variables_map& vm);
+
+    static int parseMemoryString(std::string input);
```
Spacechild1, line 28: Should be `const std::string&`?

lang/LangSource/SC_LanguageClient.cpp
```diff
@@ -99,9 +99,8 @@ void SC_LanguageClient::initRuntime(const Options& opt) {
     // start virtual machine
     if (!mHiddenClient->mRunning) {
         mHiddenClient->mRunning = true;
-        if (opt.mRuntimeDir) {
-            int err = chdir(opt.mRuntimeDir);
-            if (err)
+        if (!opt.mRuntimeDir.empty()) {
+            if (int err = chdir(opt.mRuntimeDir.c_str()))
```
Spacechild1, line 103: Unused variable `err`. Better:
```
if (chdir(opt.mRuntimeDir.c_str()) != 0)
```

lang/LangSource/SC_CLIOptions.cpp
```diff
@@ -0,0 +1,284 @@
...
+        multiplier = 1024 * 1024;
+    } else if (suffix == 'g' || suffix == 'G') {
+        multiplier = 1024 * 1024 * 1024;
+    } else if (!isdigit(suffix)) {
+        cerr << "Unknown modifier '" << suffix << "'" << endl;
+        return 0;
+    }
+
+    std::string numberPart = std::isdigit(suffix) ? input : input.substr(0, input.size() - 1);
+    int value = boost::lexical_cast<int>(numberPart);
```
Spacechild1, line 251: Note: `boost::lexical_cast` throws an `boost::bad_lexical_cast` exception if the conversion is not successfully. I guess we should catch it here, print an error and return 0?

lang/LangSource/SC_TerminalClient.cpp
```diff
@@ -101,168 +104,29 @@ void SC_TerminalClient::postError(const char* str, size_t len) {
...
-
-optArgInvalid:
-    ::post("%s: invalid argument for option -%c -- %s\n", getName(), optopt, optarg);
-    quit(1);
-    return false;
-}
-
 int SC_TerminalClient::run(int argc, char** argv) {
     Options& opt = mOptions;
+    int error_code;
```
Spacechild1, line 109: camelCase 🤓

lang/LangSource/SC_TerminalClient.cpp
```diff
@@ -101,168 +104,29 @@ void SC_TerminalClient::postError(const char* str, size_t len) {
...
 
-    if (!parseOptions(argc, argv, opt)) {
-        return mReturnCode;
-    }
-
-    // finish argv processing
-    const char* codeFile = nullptr;
+    SC_CLI::CLIOptions cli_options;
```
Spacechild1, line 111: camelCase 🤓

lang/LangSource/SC_TerminalClient.h
```diff
@@ -56,24 +56,22 @@ class SCLANG_DLLEXPORT SC_TerminalClient : public SC_LanguageClient {
 
     struct Options : public SC_LanguageClient::Options {
         Options():
-            mLibraryConfigFile(0),
+            mLibraryConfigFile(""),
```
Spacechild1, line 59: Redundant because `std::string` has a default constructor

lang/LangSource/SC_TerminalClient.h
```diff
@@ -56,24 +56,22 @@ class SCLANG_DLLEXPORT SC_TerminalClient : public SC_LanguageClient {
...
             mDaemon(false),
             mCallRun(false),
             mCallStop(false),
             mStandalone(false),
-            mArgc(0),
-            mArgv(0) {}
+            mArgs({}) {}
```
Spacechild1, line 64: Redundant because `std::vector` has a default constructor

Spacechild1 (changes requested):

Spacechild1:
Thanks! Caught a few more things.

lang/LangSource/SC_CLIOptions.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <string>
+#include "SC_TerminalClient.h"
+
+namespace po = boost::program_options;
+using namespace std;
+
+namespace SC_CLI {
+class CLIOptions {
+public:
+    CLIOptions() {};
```
Spacechild1, line 11: Please no semi-colon after inline function definitions :) Also, this constructor does nothing, so you can just omit it altogether.

lang/LangSource/SC_TerminalClient.h
```diff
@@ -55,25 +55,17 @@ class SCLANG_DLLEXPORT SC_TerminalClient : public SC_LanguageClient {
...
-            mLibraryConfigFile(0),
-            mDaemon(false),
-            mCallRun(false),
-            mCallStop(false),
-            mStandalone(false),
-            mArgc(0),
-            mArgv(0) {}
-
-        const char* mLibraryConfigFile;
+        Options(): mDaemon(false), mCallRun(false), mCallStop(false), mStandalone(false) {}
```
Spacechild1, line 58: Just a comment: since C++11 you can initialize member variables inline. In this case, `Options` wouldn't even need a user-defined constructor. In fact, that's what you are already doing in `CLIOptions`. Generally, in modern C++ we only tend to write constructors when they are really necessary.

include/lang/SC_LanguageClient.h
```diff
@@ -40,17 +41,17 @@ SCLANG_DLLEXPORT void destroyLanguageClient(class SC_LanguageClient*);
 class SCLANG_DLLEXPORT SC_LanguageClient {
 public:
     struct Options {
-        Options(): mMemSpace(2 * 1024 * 1024), mMemGrow(256 * 1024), mPort(57120), mRuntimeDir(0) {}
+        Options(): mMemSpace(2 * 1024 * 1024), mMemGrow(256 * 1024), mPort(57120) {}
```
Spacechild1, line 44: Just as a side note, this constructor can be omitted as well:
```
struct Options {
    int mMemSpace = 2 * 1024 * 1024; // memory space in bytes
    int mMemGrow = 256 * 1024; // memory growth in bytes     
    int mPort = 57120; // network port number
    std::string mRuntimeDir; // runtime directory
};
```
A little bit more verbose, but maybe clearer:
```
struct Options {
    static constexpr int defaultMemSpace = 2 * 1024 * 1024;
    static constexpr int defaultMemGrow = 256 * 1024;
    static constexpr int defaultPort = 57120;

    int mMemSpace = defaultMemSpace; // memory space in bytes
    int mMemGrow = defaultMemGrow; // memory growth in bytes     
    int mPort = defaultPort; // network port number
    std::string mRuntimeDir; // runtime directory
};
```

lang/LangSource/SC_CLIOptions.cpp
```diff
@@ -0,0 +1,288 @@
...
+    } else if (!isdigit(suffix)) {
+        cerr << "Couldn't parse memory value '" << input << "': '" << suffix << "' is an unknown modifier" << endl;
+        exit(1);
+    }
+
+    std::string numberPart = std::isdigit(suffix) ? input : input.substr(0, input.size() - 1);
+    try {
+        int value = boost::lexical_cast<int>(numberPart);
+        return value * multiplier;
+    } catch (boost::bad_lexical_cast& e) {
```
Spacechild1, line 254: Nitpick: should be `const boost::bad_lexical_cast&`. You can also omit the name of the variable if you don't actually use it.

Spacechild1:
A few more little things. Sorry I didn't catch these immediately...

capital-G:
> A few more little things. Sorry I didn't catch these immediately...

No problem at all, I really enjoy learning some things here, hope you don't get annoyed by it :)

lang/LangSource/SC_CLIOptions.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+    if (mPrintVersion) {
+        cout << "sclang " << SC_VersionString() << " (" << SC_BuildString() << ")" << endl;
+        return true;
+    }
+
+    // this is handled here and not in the description declaration b/c we need access to the declarations
+    if (mPrintHelp) {
+        // construct a options_description w/o hidden positional options
+        po::options_description visibleDescription("");
```
Spacechild1, line 77: Since you said you are ok with nitpicks, here's another one: no need to explicitly pass an empty string, just use the default constructor, as you do with the other instances of `po::options_description`.

Spacechild1 (changes requested):

Spacechild1 (approved):

Spacechild1:
I think this is a great feature! I have do admit that I didn't actually test the command line options myself, but I trust you when you say that they all work as expected. I don't think that any of your changes could possibly behave differently on Windows.

lang/LangSource/SC_CLIOptions.hpp
```diff
@@ -0,0 +1,29 @@
+#include "boost/program_options.hpp"
+#include <string>
+#include "SC_TerminalClient.h"
+
+namespace po = boost::program_options;
+
+namespace SC_CLI {
+class CLIOptions {
+public:
+    bool parse(int argc, char* argv[], SC_TerminalClient::Options& terminalOptions, int& exitCode);
```
JordanHendersonMusic, line 10: I think this signature is a bit confusing with the bool and the exit code. IIUC, this function returns three possible states? Perhaps just use an enum? Or maybe an enum and an optional?

JordanHendersonMusic (commented):
I think this is a good idea, I've just got one suggestion where I think the code could be easier to understand.

lang/LangSource/SC_CLIOptions.hpp
Spacechild1: I think you're right. There are basically three possible states:
1. return `true` + exit code 0 (e.g. `--version`)
2. return `true` + exit code 1 (parse error)
3. return `false`: ok and continue

> Or maybe an enum and an optional?

A simpler solution would be to just return the error code as an `std::optional`:
```
std::optional<int> parse(int argc, char* argv[], SC_TerminalClient::Options& terminalOptions);
```
If the optional is valid, parsing has stopped and the exit code can be retrieved from the optional.

But that's a matter of taste I guess.

capital-G: Went for the `std::optional` approach which seems like a good way to do things.

IIUC there is no way to make this actually type safe? So something like
```cpp
    if (const auto errorCode = cliOptions.parse(argc, argv, mOptions)) {
        return *errorCode;
    }
```
could introduce runtime problems b/c the type system doesn't guarantee that error code is actually non-null - only the runtime will assure me this, right?

E.g. in Rust you would wrap the return value into an [`Option`](https://doc.rust-lang.org/std/option/) and pattern match it, so e.g.

```rust
    let parse_error: Option<int> = Some(42);

    match parse_error {
        Some(error_code) => {
            // error_code is guaranteed to exist as int by the type checker
            !exit(error_code)
        },
        None => {},
    }
```

JordanHendersonMusic:
Looks good! I also appreciate the '\\\' so it appears in the editor 😃

JordanHendersonMusic (approved):

lang/LangSource/SC_CLIOptions.hpp
```diff
@@ -0,0 +1,29 @@
+#include "boost/program_options.hpp"
+#include <string>
+#include "SC_TerminalClient.h"
+
```
JordanHendersonMusic, line 4: `#include <optional>`

dyfer:
I've run this Windows build from this PR in a VM, tested a few of the long args and they seem to work.

Adjacent to this PR:
In the past, when the "standalone"/"portable" build of SC was discussed, there were 2 different concepts of the "standalone" parameter proposed:
- excluding all search paths (and needing to include the scclasslibrary path in order to run SC) - useful e.g. for running SC with a different class library for testing; current `-a` option; running a "portable" build of SC requires manually including the path to the included class library
- excluding all search paths _external to the app bundle/folder_ (i.e. user/system extensions) - this is actually desirable for a "portable"/"standalone" app; not currently implemented

While changing/extending this is probably beyond the scope of this PR, I also wonder if the new `--standalone` flag could actually mean the _latter_ (new) behavior; (the long option for `-a` could be `--exclude-all-search-paths` or something like that).

Even if we merge this as is, we can still change the behavior of the flag before the 3.14 release, if we think it's a desirable change. Please let me know what you think!

capital-G:
> While changing/extending this is probably beyond the scope of this PR, I also wonder if the new `--standalone` flag could actually mean the _latter_ (new) behavior; (the long option for `-a` could be `--exclude-all-search-paths` or something like that).

Yeah, I also found this a bit misleading and I also find your flag more sensible in this case.
Maybe we should use `--exclude-default-search-paths` instead of `--standalone` and also explain this a bit better it in the help of the command - should I adapt this @dyfer ?

This PR is actually one step towards making isolated sclang environments easier :)

capital-G:
@dyfer I now added [switch from --standalone to --exclude-default-paths CLI flag](https://github.com/supercollider/supercollider/pull/6640/commits/d692730e6def44e726025934d2a70c26fedae4c1)

Help looks now like

```
❯ ./sclang --help
Start a REPL interpreter:           sclang [ -options ]
Execute a .scd file:                sclang [ -options ] path_to_sc_file [ args passed to sclang... ]

Generic options:
  -h [ --help ]                     Print this message and exit
  -v [ --version ]                  Print sclang version and exit

sclang options:
  -d [ --runtime-directory ] arg    Set runtime directory
  -D [ --daemon ]                   Enter daemon mode
  -g [ --heap-growth ] arg (=256k)  Set heap growth size (allows k, m and g as 
                                    suffix multipliers)
  -l [ --yaml-config ] arg          Set yaml config
  -m [ --heap-size ] arg (=2m)      Set initial heap size (allows k, m and g as
                                    suffix multipliers)
  -r [ --call-run ]                 Call Main.run on startup
  -s [ --call-stop ]                Call Main.stop on shutdown
  -u [ --port ] arg (=57120)        Set UDP listening port
  -i [ --ide ] arg                  Set IDE name
  -a [ --exclude-default-paths ]    Exclude searching in default SCClassLibrary
                                    and user and system extensions paths (aka 
                                    'standalone' mode).

Language config options (overwrites provided yaml entries):
  --include-path arg                Class library path to be included for 
                                    searching (allows multiple mentions)
  --exclude-path arg                Class library path to be excluded from 
                                    searching (overrides includePaths, allows 
                                    multiple mentions)
  --post-inline-warnings            Boolean flag to turn on warnings about 
                                    missing inline opportunities.

```

dyfer:
Looks great! I like the changed flag.

capital-G:
@JordanHendersonMusic is there anything missing for a merge?

dyfer:
Sorry @capital-G, #6727 created a conflict here... needs to be resolved.

dyfer:
I see the individual commits from this PR on the develop branch... This PR was supposed to be squashed on merge, but I don't think it was. 

I certainly made that mistake in the past! Just a note for the future.

JordanHendersonMusic:
Ah my fault! Can it be undone easily?

PR #4221, opened 2018-12-21
fix Score examples and Platform.defaultTempDir on OSX

geoffroymontel:
Purpose and Motivation
----------------------

Fixes #1271

Types of changes
----------------

- documentation of Score
- changed `Platform.defaultTempDir` because `/tmp` is not writeable on OSX and is used in Score rendering

Checklist
---------

- [X] All previous tests are passing
- [ ] Tests were updated or created to address changes in this PR, and tests are passing
- [X] Updated documentation, if necessary
- [X] This PR is ready for review

Remaining Work
--------------

PR #6642, opened 2025-02-25
Update defaultTempDir to use OS Environment Variables

### LINKED ISSUE #4653: Platform.defaultTempDir is wrong on Windows
Hi there, on Windows 10 (SC 3.10.3), the temporary directory is not set as indicated by the documentation (or what would be expected).

## Environment

* SuperCollider version: 3.10.3
* Operating system: Windows 10

## Steps to reproduce

Evaluate any of the following:

```supercollider
PathName.tmp
thisProcess.platform.defaultTempDir
Platform.defaultTempDir
```

They all return the same as `Platform.userAppSupportDir` (in my case, `C:\Users\username\AppData\Local\SuperCollider\`). They would be expected to return a *temporary* directory, such as `C:\Users\username\AppData\Local\Temp` (which is the value of the TMP and TEMP environment variables in my Windows setup).

The documentation (e.g. in `PathName.tmp` doc) also incorrectly suggests it will use `/WINDOWS/TEMP`.

### LINKED ISSUE #5239: OSX uses an SC managed defaultTempDir but never cleans it up
## Environment

* SuperCollider version: 3.11
* Operating system: OSX 10.14.16

#4221 changed to an SC created temp dir on OSX, because of apparent problems writing to /tmp/ in El Capitan, though details were not confirmed. This seems a bit weird (what's the point of a temp dir you can't use), and I can't find this documented elsewhere. IAC, this means that on OSX 'temp' files are never cleaned up.

Assuming this is still a problem, the obvious solution would be to add a cleanup stage on quit. I suggest a little more investigation and testing might be good though, as most Google results suggest this is a permissions problem. If that's the case, would make sense to revert to /tmp/.

prko:
closes #5239

---

To understand the actual states of this PR, please read from: https://github.com/supercollider/supercollider/pull/6642#issuecomment-2687065872

---

## Purpose and Motivation
The defaultTempDir on Windows does not seem to be fully defined:
```
Platform.defaultTempDir.standardizePath
```
returns
>-> C:\Users\\[_User_Account_]\AppData\Local\SuperCollider\

```
Platform.userAppSupportDir
```
returns
>-> C:\Users\\[_User_Account_]\AppData\Local\SuperCollider

This PR is based on the defaultTempDir path implementation for MacOS:
https://github.com/supercollider/supercollider/blob/4f6ba9951a611f319f14a34690cadf7f9fa73804/SCClassLibrary/Platform/osx/OSXPlatform.sc#L80-L87

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Platform/windows/WindowsPlatform.sc
+++ b/SCClassLibrary/Platform/windows/WindowsPlatform.sc
@@ -43,7 +43,7 @@ WindowsPlatform : Platform {
 
 	defaultTempDir {
 		// +/+ "" looks funny but ensures trailing slash
-		var tmp = this.userAppSupportDir +/+ "tmp\\";
+		var tmp = "TEMP".getenv +/+ "supercollider".withTrailingSlash;
 		if(File.exists(tmp).not) { tmp.mkdir };
 		^tmp;
 	}
--- a/SCClassLibrary/Platform/windows/WindowsPlatform.sc
+++ b/SCClassLibrary/Platform/windows/WindowsPlatform.sc
@@ -43,8 +43,9 @@ WindowsPlatform : Platform {
 
 	defaultTempDir {
 		// +/+ "" looks funny but ensures trailing slash
-		var tmp = this.userAppSupportDir +/+ "";
-		^if(File.exists(tmp)) { tmp }
+		var tmp = this.userAppSupportDir +/+ "tmp\\";
+		if(File.exists(tmp).not) { tmp.mkdir };
+		^tmp;
 	}
 
 	myDocumentsDir {
--- a/SCClassLibrary/Platform/windows/WindowsPlatform.sc
+++ b/SCClassLibrary/Platform/windows/WindowsPlatform.sc
@@ -42,7 +42,7 @@ WindowsPlatform : Platform {
 	}
 
 	defaultTempDir {
-		// +/+ "" looks funny but ensures trailing slash
+		// ensure trailing slash due to backwards compatibility
 		var tmp = "TEMP".getenv +/+ "supercollider".withTrailingSlash;
 		if(File.exists(tmp).not) { tmp.mkdir };
 		^tmp;
--- a/SCClassLibrary/Platform/osx/OSXPlatform.sc
+++ b/SCClassLibrary/Platform/osx/OSXPlatform.sc
@@ -78,11 +78,10 @@ OSXPlatform : UnixPlatform {
 	}
 
 	defaultTempDir {
-		var dir;
 		// ensure trailing slash due to backwards compatibility
 		// see discussion at https://github.com/supercollider/supercollider/pull/4221
-		dir = Platform.userAppSupportDir +/+ "tmp/";
-		if(File.exists(dir).not) { dir.mkdir };
-		^dir;
+		var tmp = "TMPDIR".getenv +/+ "supercollider".withTrailingSlash;
+		if(File.exists(tmp).not) { tmp.mkdir };
+		^tmp;
 	}
 }
--- a/SCClassLibrary/Platform/osx/OSXPlatform.sc
+++ b/SCClassLibrary/Platform/osx/OSXPlatform.sc
@@ -79,7 +79,6 @@ OSXPlatform : UnixPlatform {
 
 	defaultTempDir {
 		// ensure trailing slash due to backwards compatibility
-		// see discussion at https://github.com/supercollider/supercollider/pull/4221
 		var tmp = "TMPDIR".getenv +/+ "supercollider".withTrailingSlash;
 		if(File.exists(tmp).not) { tmp.mkdir };
 		^tmp;


capital-G:
Wouldn't `C:\Users\userName\AppData\Local\Temp` be the proper temp dir for windows? See https://learn.microsoft.com/en-us/dotnet/api/system.io.path.gettemppath?view=net-9.0&redirectedfrom=MSDN&tabs=windows#System_IO_Path_GetTempPath

I'd be surprised if ` C:\Users\[User_Account]\AppData\Local\SuperCollider\tmp` would get cleaned by the OS, but a temp folder should be cleaned by the OS.

prko:
>I'd be surprised if  C:\Users\[User_Account]\AppData\Local\SuperCollider\tmp would get cleaned by the OS, but a temp folder should be cleaned by the OS.

I do not believe that Windows OS automatically cleans `C:\Users\[User_Account]\AppData\Local\SuperCollider\tmp`. 
In fact, I'm not sure if it cleans `C:\Users\[User_Account]\AppData\Local\Tempeither`. 
Similarly, I do not think that macOS automatically cleans `~/Application Support/SuperCollider/tmp`.

If my memory serves me correctly, I have manually removed the temporary files generated by sclang on macOS, Windows, and Linux. However, I am not entirely certain about this.

The reason I chose `C:\Users\[User_Account]\AppData\Local\SuperCollider\tmp` is to keep it consistent with macOS. On macOS, `Platform.defaultTempDir` is a subfolder of `Platform.userAppSupportDir`. This is due to my need to check across multiple operating systems. Typically, users would use SuperCollider on a single platform, so synchronising the location of `Platform.defaultTempDir` might not be necessary.

Thank you for your understanding, and I look forward to your thoughts on this matter.

capital-G:
> I do not believe that Windows OS automatically cleans `C:\Users\[User_Account]\AppData\Local\SuperCollider\tmp`.
> In fact, I'm not sure if it cleans `C:\Users\[User_Account]\AppData\Local\Temp` either.

`C:\Users\[User_Account]\AppData\Local\SuperCollider\tmp` will not be cleaned, but `C:\Users\[User_Account]\AppData\Local\Temp` could be cleaned and is the advised as the temp directory for an application, so I think we should use that so it is possible to clean up any temp data across applications easily.

> Similarly, I do not think that macOS automatically cleans `~/Application Support/SuperCollider/tmp`.

This is correct - this should also point to `/tmp` which would be cleared by the OS. In Linux it is set to `/tmp` IIUC.
Anything within `~/Application Support/SuperCollider/` is considered user data - i.e. it will be backed up by TimeMachine and this contradicts the idea of a temporary folder/file. See https://stackoverflow.com/questions/43250073/standard-temporary-directory-in-osx and https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW2 

> Put temporary data in the tmp/ directory. Temporary data comprises any data that you do not need to persist for an extended period of time. Remember to delete those files when you are done with them so that they do not continue to consume space on the user’s device. The system will periodically purge these files when your app is not running; therefore, you cannot rely on these files persisting after your app terminates.

prko:
Then I would suggest changing `Platform.defaultTempDir` as follows
- MacOS: `/tmp`.
- Windows: `C:\Users\UserName\AppData\Local\Temp`.

Would it be acceptable to make these changes?
(I see more consistency across all platforms now, and that is a good thing. One odd thing is that only Windows seems to provide an individual temp folder per user account, but this does not seem to matter as `Platform.defaultTempDir` is outside of `Platform.userAppSupportDir` on all platforms).

capital-G:
Sounds like a reasonable change to me. Lets hope that nobody uses the temp folder to store some persistent data :D

prko:
I think so, but there is one thing to check first. I found a comment: 
https://github.com/supercollider/supercollider/blob/4f6ba9951a611f319f14a34690cadf7f9fa73804/SCClassLibrary/Platform/osx/OSXPlatform.sc#L80-L87

In #4221, The author wrote:
>documentation of Score
>changed Platform.defaultTempDir because /tmp is not writeable on OSX and is used in Score rendering

What do you think about this?
(wow, changing one line requires so many things to be considered... amazing...)

capital-G:
> changed Platform.defaultTempDir because /tmp is not writeable on OSX and is used in Score rendering

This is interesting - I don't understand why `tmp` wouldn't be writable? This works for me

```shell
❯ echo "hello" > /tmp/foobar

❯ cat /tmp/foobar
hello
```

IIRC you also have a mac? Can you test if you can write a file to `/tmp` using sclang? E.g.

```supercollider
f = File.new("/tmp/mysc", "w")
f.write("Does it work?");
f.close;
```

works on my machine as well

```shell
❯ cat /tmp/mysc
Does it work?
```

prko:
> ```shell
> ❯ cat /tmp/mysc
> Does it work?
> ```

Yes, it works.

Also the following works on macOS and Windows:
```
(
~path = (Platform.case(
	\osx,       { "/tmp" },
	\linux,     { "/tmp" },
	\windows,   { "~/AppData/Local/Temp".standardizePath }
) +/+ "mysc"
);

f = File.new(~path, "w");
f.write("It works");
f.close;

File.use(~path, "r", { |text| 
	(text.readAllString).postln;
	"\ndone!".postln
})
)
```

capital-G:
Thanks - do you want to make a PR which changes the defaultTempDir for Windows and macOS then? :)

prko:
>Thanks - do you want to make a PR which changes the defaultTempDir for Windows and macOS then? :)

I would be happy to proceed with making a PR to change the `defaultTempDir` for Windows and macOS.

However, I believe it might be prudent to wait for feedback from other reviewers who were involved or have shared their opinions on #4221. There could still be some issues that we haven't identified yet.

What do you think about it?

dyfer:
Here's another related issue: https://github.com/supercollider/supercollider/issues/5239

How about we try using environment variables for these on Windows and macOS? (this PR can stick to doing this on Windows only I think; macOS can be a separate PR).

IIUC on Windows it could be either `TEMP` or `TMP`: https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables 

On macOS that probably should be `TMPDIR`, as mentioned in the other PR.

We probably need a fallback directory if the variable is not set for some reason.

prko:
Like this?

- SCClassLibrary/Platform/windows/WindowsPlatform.sc:
   ```
   defaultTempDir {
    		^"%USERPROFILE%/AppData/Local/Temp"
 	}
   ```

- SCClassLibrary/Platform/osx/OSXPlatform.sc:
   ```
   defaultTempDir {
    		^"echo $TMPDIR".unixCmdGetStdOut
 	}
   ```

dyfer:
Not really. Env variables can be obtained in sclang directly:
```supercollider
   "TMPDIR".getenv
```
Then I'd check if this is a valid path (if the directory exists) and have some kind of a fallback (possibly to the currently used path?) if it doesn't

dyfer:
Example:

```supercollider
	defaultTempDir {
		var tmp = ("TEMP".getenv +/+ "supercollider").withTrailingSlash;
		if(File.exists(tmp).not) { tmp.mkdir };
		^tmp;
	}
```

prko:
OK, then for macOS:
```supercollider
defaultTempDir {
		var tmp = "TMPDIR".getenv +/+ "supercollider".withTrailingSlash;
		if(File.exists(tmp).not) { tmp.mkdir };
		^tmp;
	}
```
I do have a few queries:

1. I have modified the file `SCClassLibrary/Platform/windows/WindowsPlatform.sc` in my local repository. Would you (@dyfer) prefer that I upload it now, or should I wait for feedback from @capital-G and the other reviewers?

2. Could the changes in `SCClassLibrary/Platform/osx/OSXPlatform.sc` be included in the same pull request, or would it be more appropriate to submit them separately?

dyfer:
> OK, then for macOS:
> 
> ```supercollider
> defaultTempDir {
> 		var tmp = "TMPDIR".getenv +/+ "supercollider".withTrailingSlash;
> 		if(File.exists(tmp).not) { tmp.mkdir };
> 		^tmp;
> 	}
> ```
> 
> I do have a few queries:
> 
> 1. I have modified the file `SCClassLibrary/Platform/windows/WindowsPlatform.sc` in my local repository. Would you (@dyfer) prefer that I upload it now, or should I wait for feedback from @capital-G and the other reviewers?

I'd suggest to go ahead with the change and get feedback from @capital-G afterwards - I think that using env vars is the way to go here. We just need to test it a bit in different environments to make sure it works fine.

> 2. Could the changes in `SCClassLibrary/Platform/osx/OSXPlatform.sc` be included in the same pull request, or would it be more appropriate to submit them separately?

I have a slight preference for a separate PR, @capital-G seemed to be fine with including it in this one - I think maybe we can do it in this one for a quicker review.

prko:
@dyfer Done, Thanks!

dyfer:
I can't seem to link an issue to this PR - this will fix #5239.

SCClassLibrary/Platform/windows/WindowsPlatform.sc
```diff
@@ -43,8 +43,9 @@ WindowsPlatform : Platform {
 
 	defaultTempDir {
 		// +/+ "" looks funny but ensures trailing slash
```
dyfer, line 45: This is now outdated. We can either remove it, or replace with "// ensure trailing slash due to backwards compatibility" similarly to macOS implementation?

prko: done, thanks!

prko:
@dyfer 

Additionally, I have updated the opening post of this PR to include a reference to #5239.

dyfer (approved):
I tested this by running examples from the Score helpfile and checking `Platform.defaultTempDir.openOS` on macOS 14 and Windows 11 and the temp directory worked as expected.

IMO this can be merged, unless @capital-G or @jrsurge want to test on their systems first.

dyfer:
@capital-G or @jrsurge do you have any input on this PR?

capital-G (changes requested):
> @capital-G or @jrsurge do you have any input on this PR?

Sorry to go another round here, but it seems worth it to discuss such a change more thoroughly b/c IMO this could lead to some ugly problems for users.

The documentation of `Platform.defaultTempDir` says

https://github.com/supercollider/supercollider/blob/4bec645f56a8587f2ff8f7c3cd76d626b856d205/HelpSource/Classes/Platform.schelp#L74-L75

but on 3.13 w/ macOS this is not really a temp directory b/c it resides in the userConfigDir which doesn't get deleted automatically like a normal temporary directory.
In case people used this directory (wrongly) to store presets, files, ... - those would get deleted going on from this PR w/o any kind of notice.

We could argue "it was implemented wrong and we are just fixing things", but deleting or loosing files is getting into the territory which can lead to some serious frustrations which I'd like to avoid for the users.

To me it would be cleaner to deprecate the method (or issue a warning?) and create a new method called `createTempDirectory` with an optional name parameter (pre- and postfix?). It seems such a method should be placed within Platform, but I personally wouldn't search in Platform for such a method but under `FilePath`? Maybe people from the sclang team can provide some guidance here as I am not an expert on the language design of sclang.

edit: sorry that I was the person on proposing to change the behavior :/

dyfer:
> In case people used this directory (wrongly) to store presets, files, ... - those would get deleted going on from this PR w/o any kind of notice.

@capital-G this is a valid concern. Still, the current behavior was filed as an issue. So... idk.

capital-G:
@jamshark70 @telephon @muellmusik what would you advise here?

I think the difficulty is described @ https://github.com/supercollider/supercollider/pull/6642#pullrequestreview-2829089931

jamshark70:
> I personally wouldn't search in Platform for such a method but under FilePath?

Currently Platform is the place for platform-specific file locations: Platform.systemAppSupportDir, Platform.userAppSupportDir, Platform.userExtensionDir (and systemExt...), Platform.recordingsDir, Platform.resourceDir -- I don't think it's a stretch for Platform to provide the OS-specific temp directory as well.

> In case people used this directory (wrongly) to store presets, files, ... - those would get deleted going on from this PR w/o any kind of notice.

Hm. You might be right about deprecation. If we need to call users' attention to misuse of the temp directory, I'm not sure there's a better way.

dyfer:
I don't oppose deprecating the method if we want to "play it safe".

I wouldn't necessarily give the option to create a specific name - it should "just work" to provide a single place for temporary files. I'd suggest a simple name like `Platform.tempDir`.

dyfer:
Also, we have `PathName.tmp` which defaults to `Platform.defaultTempDir`. We'd need to consider whether we'd depracate that method as well.

@capital-G should we push this change before 3.14?

dyfer:
@prko sorry for the delay. I think we can move this forward, with following guidelines:
- create a new method `tempDir` for all platforms (method name is still up for discussion I think)
  - for macOS and Windows use the implementation proposed in this PR
  - for Linux, use the current implementation
- _suggestion:_ for the new implementation, I suggest dropping trailing slash, as other Platform methods don't use it
- deprecate `defaultTempDir` - see [here for examples how to deprecate](https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/deprecated/3.11/deprecated-3.11.sc). Of course here we'd use version 3.15.
  - note: _I think_ the deprecation should not redirect to the new method, but only indicate that the old one is deprecated. @capital-G @jamshark70 what do you think?
  - the deprecated method needs to include trailing slash, as the old method did

prko:
Thanks!

Should it be part of this PR to move the changed method definition in this PR to the following files?
- /SCClassLibrary/deprecated/3.15/OSXPlatform.sc
- /SCClassLibrary/deprecated/3.15/WindowsPlatform.sc

dyfer:
> Thanks!
> 
> Should it be part of this PR to move the changed method definition in this PR to the following files?
> 
> * /SCClassLibrary/deprecated/3.15/OSXPlatform.sc
> * /SCClassLibrary/deprecated/3.15/WindowsPlatform.sc

I think so... And also deprecate the implementation in the UnixPlatform class (which is in Platform.sc).

I also noticed that we may want to change the Linux (UnixPlatform) implementation. But before we move forward, let's see what others think. This PR is growing in scope...

Re: Linux implementation:

Currently we choose between `/tmp` and `userAppSupportDir`:
```supercollider
		^["/tmp/", this.userAppSupportDir +/+ ""].detect({ |path|
			File.exists(path);
		});
```
Maybe we should first check for `TMPDIR` and only revert to the current implementation if TMPDIR is not set?

dyfer:
@capital-G @jamshark70 how should we move forward here? The last 3 comments seem to represent the current state (we decided to deprecate old methods, but the question remains on whether we first check env variables when choosing directories on Linux).

PR #6645, opened 2025-02-25
Windows build guild: Update SET command to persist Qt PATH

soundlake:
## Purpose and Motivation

from the commit message:
> The environment variables by `set` command only valid during the `cmd.exe` process. To make the change persistent, there are a few different approaches. And using `setx` might be the most efficient, especially, because the doc is already using `cmd.exe`.
>
> Further down in the same doc, it is stressed to put the Qt's executable path in the PATH environment variable. But it does not mention how.

Personally, as a newbie to VS and SC build, I was puzzled when I firstly got "Qt6WebEngine" error. Further down, I found the note that the env var has to be persistent, then I realized `set` might not persist the change. Then, I found `setx`.

> I moved the command a bit earlier on the doc. Because this particular command is not required to be executed under the `supercollider/build` directory.
>
> [skip ci]

## Types of changes

- Documentation

## To-do list

- [x] Code is tested (manual test with set env var with `set` and `setx`)
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/README_WINDOWS.md
+++ b/README_WINDOWS.md
@@ -339,12 +339,21 @@ Configuration and Building
 
 *Note:* You should always use forward slashes for paths passed to CMake.
 
+Set the environment variables PATH first.
+
+    setx PATH "C:\Qt\6.8.2\msvc2022_64\bin;%PATH%"
+
+*Note:* `setx` is usable from Windows 10. Unlike `SET`, which affects
+only current `cmd.exe`, `setx` makes the change persistent.
+
+*Note:* Do not execute this command multiple times. It will populate
+`PATH` environment variable with the same path.
+
 **Confirm the location of your Qt install before executing these commands.** You
 may need to modify them if you installed Qt somewhere else. The following
 commands should be executed starting from the root directory of the
 SuperCollider repository.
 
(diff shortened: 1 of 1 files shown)


capital-G (commented):
Not sure we should advise `setx` as this will probably introduce problems later  but this should be evaluated/decided by some windows user.

dyfer:
@soundlake @capital-G how do we best to proceed with this PR, improving the instructions but avoiding a potentially problematic approach?

capital-G:
I still think we should not advise people to set global env variables as it is not considered best practice which will probably create problems later.

soundlake:
Excuse me for veeery late response and looong delayed updates. There are
some personal things are going on, and my laptop runs out of storage due to
installing different Qt versions.

Using setx isn't the way to go. I see that at the first response from
@capital-G.

I think closing this particular PR is okay. I will open a new one later
with more carefulness.

2025년 4월 8일 (화) 10:12, Dennis Scheiba ***@***.***>님이 작성:

dyfer:
@soundlake sounds good, thanks! Closing this for now.

PR #6646, opened 2025-02-28
Add GitHub Actions reminders

capital-G:
## Purpose and Motivation

In order to counteract a staleness of PRs and issues here I think it would be nice to have a reminder tool - e.g. I'd like to set a reminder in 2 weeks that this can be merged on https://github.com/supercollider/supercollider/pull/6643#pullrequestreview-2650404002

Sadly, this is not available natively via github, so we have to rely on an GitHub action to inform us.

For usage see

https://github.com/agrc/kitchen-sink/issues/285#issuecomment-2316411983
https://github.com/agrc/reminder-action?tab=readme-ov-file

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/org_post_reminder.yml
+++ b/.github/workflows/org_post_reminder.yml
@@ -1,0 +1,13 @@
+name: Issue Comment Events
+
+on:
+  issue_comment:
+    types: [created, edited]
+
+jobs:
+  reminder:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: check for reminder
+        uses: agrc/create-reminder-action@v1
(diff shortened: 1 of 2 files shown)


capital-G:
I am unsure if this works on a per-branch basis or on a per-repo basis - so it may be that the following command only works after a merge.

/remind me to merge this PR in 1 week

dyfer:
Thanks for this!
I think this is generally desirable, though I thought that we'd rather need an automatic ping/close issue action after a certain time, like [`stale` action in vcpkg](https://github.com/microsoft/vcpkg/blob/master/.github/workflows/stale.yml). Do you think a manual reminder is more useful in our case?

capital-G:
I think we should maybe not go for this anymore as it is somehow hacky. Hopefully this can be introduced natively by GH or we find a way to run a GH bot ourselves or use a calendar for something like this.

Closing for now.

Issue #6647, opened 2025-02-28
notify scsynth tcp not working


sonoro1234:
Hello,

This sequence of /notify messages and reponses happens in scsynth with tcp only:

* sending /notify,1 returns /done
* sending /notify,0 returns /fail not registered
* sending /notify,1 returns /fail already registered

I modified `bool NotifyCmd::Stage2()` to print `mReplyAddress` just to check if they were different, but they were the same in the three calls.
The three calls were done leaving some time between them so that it does not seem a threading issue (but who knows)

I cant find an explanation for this failing only in tcp.
Any clues to keep on searching?

Can anyone confirm or deny this behaviour?
Any results would be helpful.

sonoro1234:
First clues:

All is related to a comparision in https://github.com/supercollider/supercollider/blob/develop/server/scsynth/SC_SequencedCommand.cpp#L1241

where the comparision is not between pointers but all the members of Class ReplyAddress: https://github.com/supercollider/supercollider/blob/develop/common/SC_Reply.cpp#L27 

So I plugged a scprintf statement in where `notify address` the addres in the call and `mUsers address` are the addresses stored previously:

See how the calling address is changing althought it should not.

```
notify address 0000004db41ff9e0, mAddress 0000004db41ff9a0, mProtocol 1, mPort 1342187887, mSocket 2715506
NODE TREE Group 0
notify address 0000004db41ff9e0, mAddress 0000004db41ff9a0, mProtocol 1, mPort 0, mSocket 0
mUsers address 0000004db41ff9e0, mAddress 0000004db41ff980, mProtocol 1, mPort 1342187887, mSocket 2715506 equal 0
notify address 0000004db41ff9e0, mAddress 0000004db41ff9a0, mProtocol 1, mPort 0, mSocket 0
mUsers address 0000004db41ff9e0, mAddress 0000004db41ff980, mProtocol 1, mPort 0, mSocket 0 equal 1
/notify : already registered
notify address 0000004db41ff9e0, mAddress 0000004db41ff9a0, mProtocol 1, mPort 0, mSocket 0
mUsers address 0000004db41ff9e0, mAddress 0000004db41ff980, mProtocol 1, mPort 0, mSocket 0 equal 1
/notify : already registered
notify address 0000004db41ff9e0, mAddress 0000004db41ff9a0, mProtocol 1, mPort 7929971, mSocket 7602286
[...]

PR #6652, opened 2025-03-02
scsynth tcp: fixes uninitialized ReplyAddress (issue #6647)

sonoro1234:
Fixes issue #6647

Also fixed for internal server.

Just in case: supernova does not have this issue.

- [X] Code is tested
- [x] All tests are passing

(There are two failing CI tests completely unrelated and they also fail in other PRs)
`CMake Error at QtCollider/CMakeLists.txt:38 (find_package):`

### DIFF
--- a/server/scsynth/SC_ComPort.cpp
+++ b/server/scsynth/SC_ComPort.cpp
@@ -376,6 +376,9 @@ private:
         packet->mReplyAddr.mProtocol = kTCP;
         packet->mReplyAddr.mReplyFunc = tcp_reply_func;
         packet->mReplyAddr.mReplyData = (void*)&socket;
+        packet->mReplyAddr.mPort = socket.remote_endpoint().port();
+        packet->mReplyAddr.mSocket = 0;
+        packet->mReplyAddr.mAddress = socket.remote_endpoint().address();
 
         packet->mSize = OSCMsgLength;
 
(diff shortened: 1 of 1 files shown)


capital-G:
Thanks for this fix, it was really interesting to follow the investigation :)

IIUC - another approach could be to fix the identity check for TCP or is there something that would create problems in this case?
Maybe the approach to fix the identity is cleaner b/c we don't have to populate members that aren't really necessary for a TCP connection?

If we settle for the current path (which is also perfectly fine to me) I think there should be a comment added to the source code why we populate those fields :)

sonoro1234:
> IIUC - another approach could be to fix the identity check for TCP or is there something that would create problems in this case?
> Maybe the approach to fix the identity is cleaner b/c we don't have to populate members that aren't really necessary for a TCP connection?

It really would mean not to fix identity but also making some unused fields to disappear, but I could see some references in sclang that I am not able to say if are unused (they could be used in .sc files) and the same happens with the identity (Which existed  before Tim Blechmann refactored the code). So, given that those fields are there it is a must to initialize them.

> I think there should be a comment added to the source code why we populate those fields :)

It is difficult to give a short explanation of what is found in issue #6647( It remains as a reference in the commit text). A short explanation would be `dont let uninitialize fields` but is too obvious and leaves the important details out.

Issue #6526, opened 2024-11-06
integer exponent pow


telephon:
## Motivation

The behaviour of `pow` / `**` with UGens differs from the behaviour with numbers. The reason is that UGens should be understood as streams of reals: they incude both negative and positive numbers, as well as mostly non-integers. An exponential function with a non-integer exponent has no unambiguous definition for negative basis (e.g. `-2^2 = 4` and `2^1.3 = 2.4622888266898`, but `-2^1.3 = nan`.

For integer exponents, however, the behaviour is well defined. But we have no operation in sclang that makes this assumption for UGens. So we can't write `SinOsc.ar ** 2` and make it behave like `SinOsc.ar.squared`.

## Description of Proposed Feature

A binary operator `powi` that assumes and ensures that the exponent is integer, which works like `pow` for numbers, but has a separate implementation for UGens so that it behaves the same as in numbers.

## Plan for Implementation

- a binary op ugen implementation
- an interface like `pow`

If this is agreed to be useful, I (@telephon) can implement it.

JordanHendersonMusic:
I think we could do something more complete and implement a `powAbs` function that never produces nans.

$-2^{1.5}$ does have a value, it is $\sqrt{2} i$. Taking the absolute of this gives us $\sqrt{2}$. 
I think (!?) this produces the same result as your `powi` function, but works across the whole $R$ domain. 

This can simply be implement as`x.abs.pow(n)`.

JordanHendersonMusic:
Here is another option....

$-2^{1.3}$ does have a value, it is $-1.44730... - i 1.99203... $.

We could instead take the magnitude of this result which is $2.46229...$ and apply the sign of the real part, giving $-2.46229..$

Lets call this operation `powMag`.
It creates a nice continuity when moving the exponent around, consider... 

`-2.powMag(3) == -8`
`-2.powMag(3.0001) == -8.00055...`

I think this is complete and works the same as your `powi` (!?) for integer inputs but does not produce nans.

telephon:
Ok, it seems we have two additional features here:
1. extend the ugen support in a new way considering complex arithmetic. There is this old issue: https://github.com/supercollider/supercollider/issues/5221 – maybe this coud include something along these lines?
2. have a predictable pow function that works the same between numbers and ugens (I need this for differential equations)

[...]

PR #6659, opened 2025-03-06
sclang: PyrSignal, PyrMathOps: implement pow function for Signals (#6558)

### LINKED ISSUE #6658: Signal.pow(2) fails
## Environment

* SuperCollider version: 3.14.0-dev built from branch develop (c99a9eff4)
* Operating system: Linux (Ubuntu)

## Steps to reproduce

I cannot call the pow function on signal

```supercollider
Signal[1.0, 3.0, 1.0].pow(2) // This is not working
```

It is working for normal arrays and also for FloatArray (which Signal extends from):

```supercollider
[1.0, 3.0, 1.0].pow(2) // This is working
FloatArray[1.0, 3.0, 1.0].pow(2) // This is also working
```

Interestingly `squared` is working

```supercollider
Signal[1.0, 3.0, 1.0].squared // This is working fine!
```

I would expect a Signal with the values `[1.0, 9.0, 1.0]` but I get the following error:

```
ERROR: binary operator 'pow' failed.
RECEIVER:
   Float 0.200000   9999999A 3FC99999
ARGS:
Instance of Signal {    (0x7c5680482608, gc=28, fmt=03, flg=00, set=02)
  indexed slots [3]
      0 : 1
      1 : 3
      2 : 1
}
   nil
PATH: /home/...

PROTECTED CALL STACK:
	Meta_MethodError:new	0x64cfb466e100
		arg this = BinaryOpFailureError
		arg what = nil
		arg receiver = 0.2
	Meta_DoesNotUnderstandError:new	0x64cfb46708c0
		arg this = BinaryOpFailureError
		arg receiver = 0.2
		arg selector = pow
		arg args = [Signal[1.0, 3.0, 1.0], nil]
		arg keywordArgumentPairs = nil
	SimpleNumber:performBinaryOpOnSignal	0x64cfb53b1380
		arg this = 0.2
		arg aSelector = pow
		arg aSignal = Signal[1.0, 3.0, 1.0]
		arg adverb = nil
	EvaluateProvider:doEvaluate	0x64cfb4749d00
		arg this = an EvaluateProvider
		arg func = a Function
		var result = nil
	a FunctionDef	0x64cfb474ad40
	Function:prTry	0x64cfb49a4800
		arg this = a Function
		var result = nil
		var thread = a Routine
		var next = a Function
		var wasInProtectedFunc = true
	Function:try	0x64cfb49a42c0
		arg this = a Function
		arg handler = a Function
		var result = nil
	EvaluateProvider:onReceived	0x64cfb474a180
		arg this = an EvaluateProvider
		arg method = textDocument/evaluateSelection
		arg params = Dictionary[(sourceCode -> Signal[1.0, 3.0, 1.0].pow(0.2)), (textDocument -> Dictionary[(uri -> file:///...)])]
		var source = Signal[1.0, 3.0, 1.0].pow(0.2)
		var document = a LSPDocument
		var function = a Function
		var result = nil
		var deferredResult = Deferred(1339402738)
	a FunctionDef	0x64cfb5806080
	Function:prTry	0x64cfb49a4800
		arg this = a Function
		var result = nil
		var thread = a Routine
		var next = nil
		var wasInProtectedFunc = true
	
CALL STACK:
		arg this = <instance of BinaryOpFailureError>
	< FunctionDef in Method EvaluateProvider:onReceived >
		arg error = <instance of BinaryOpFailureError>
	Function:try
		var result = <instance of BinaryOpFailureError>
	EvaluateProvider:onReceived
		arg this = <instance of EvaluateProvider>
		arg method = 'textDocument/evaluateSelection'
		arg params = <instance of Dictionary>
		var source = "Signal[1.0, 3.0, 1.0].pow(0.2)"
		var document = <instance of LSPDocument>
		var function = <instance of Function>
		var result = nil
		var deferredResult = <instance of Deferred>
	< FunctionDef in Method Deferred:using >  (no arguments or variables)
	Function:prTry
		var result = nil
		var thread = <instance of Routine>
		var next = nil
		var wasInProtectedFunc = true
	Function:try
		var result = nil
	< FunctionDef in Method Deferred:using >
		var funcResult = 'noValue'
	Routine:prStart
		arg this = <instance of Routine>
		arg inval = 10549.406846273
^^ ERROR: binary operator 'pow' failed.
RECEIVER: 0.2
```

passyur:
## Purpose and Motivation

Fixes #6658. The issue is that `pow` operations on `Signal` Objects error out mysteriously—the cause being that there are no `pow` functions implemented for `Signal`s.

The fix adds three new functions, mirroring the structure for existing `Signal` operations. (One lets you take a number to the power of a `Signal`, the other a `Signal` to the power of a number, and another a `Signal` to the power of another `Signal`.) They are included appropriately in the opcode switch cases for use when being interpreted.

## Types of changes

- Bug fix
- New feature (see above)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

I do notice that attempting to take a float/double to the power of a `Signal` will crash the interpreter. I'm a bit lost as to why—though it's certainly specific to `Signals`, as it does not crash for plain arrays. Taking an integer to the power of a `Signal` of floats/doubles works as expected. Might open an issue for that. On a practical note, I'm not sure at all why you'd want to do this, since the output is just an Array type.


capital-G:
Thanks for this fix!

The code looks good to me, but maybe it is better if someone who is more proficient with sclang internals to review this. I'll ping someone.

Would you mind adding a small unit test here https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestSignal.sc so we can increase our test coverage? :)

> I do notice that attempting to take a float/double to the power of a `Signal` will crash the interpreter. I'm a bit lost as to why—though it's certainly specific to `Signals`, as it does not crash for plain arrays. Taking an integer to the power of a `Signal` of floats/doubles works as expected. Might open an issue for that. On a practical note, I'm not sure at all why you'd want to do this, since the output is just an Array type.

Thanks for this observation - can you make a bug report of this so we don't loose this information? The interpreter should never crash.

JordanHendersonMusic:
Lets just take a second here...

Mod doesn't work either. This is mentioned in the help file of `Signal`. It is confusingly placed under the `<!` operator. I'd like to see mod fixed with this PR.

Is there a reason these haven't been implemented yet?
Could it be because there is ambiguity with negative numbers, i.e, `SinOsc.ar.pow(2) != SinOsc.ar.squared`? Modular with negative numbers is similarly undefined (or rather there are multiple definitions).

Your implementation here would remove the negative numbers, which makes sense for `FloatArray`, but given a `Signal` is usually audio, I'd expected it to behave the same as the `UGen`. Creating a disparity here seems wrong.

Given we already have this behaviour with `UGen`s I'd be in favour of preserving the negative sign and noting in the documentation that if you want to do the usual maths thing, you should do this.
```supercollider
Signal[1.0, 3.0, 1.0].as(FloatArray).pow(2).as(Signal)
```
I admit this is ugly, but can't see another way. This issue also came up over in #6526.

---

Re: the implementation in this PR. It all looks good. We shouldn't be using the `UNROLL_CODE ` and `BINOP_LOOP2` macros, but given all the code around them uses them in an identical pattern, it is okay for now.

passyur:
> Would you mind adding a small unit test here https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestSignal.sc so we can increase our test coverage? :)

Not at all. :) 

> Thanks for this observation - can you make a bug report of this so we don't loose this information? The interpreter should never crash.

Err, I would, but I seem to have fixed it. In 53b0cbe, there's no `return` after taking a `Float` to the power of a `Signal`, hence the crash. Fixed in 6c572a7.

passyur:
I've read over that issue you linked, Jordan, and yeah, that `SinOsc.ar.pow(2) != SinOsc.ar.squared` is weird but to fix it so that they're equal seems breaking. I'm having a bit of trouble getting for myself your intuition about `Signal` powers from `UGen` powers—could you provide an example of how you'd expect a `Signal` to work, specifically in the context of a negative number? 

That the modulo is multiply-defined seems worth mentioning in the documentation. It seems like the dominant modulo function in the lang, `sc_mod(int,int)` (`include/plugin_interface/SC_InlineBinaryOp.h:338`), has some casework before defaulting to C++'s truncated modulo `%` and, well, I'll just paste it:
```c++
inline int sc_mod(int in, int hi) {
    // avoid the divide if possible
    const int lo = 0;
    if (in >= hi) {
        in -= hi;
        if (in < hi)
            return in;
    } else if (in < lo) {
        in += hi;
        if (in >= lo)
            return in;
    } else
        return in;

    if (hi == lo)
        return lo;

    int c;
    c = in % hi;
    if (c < 0)
        c += hi;
    return c;
}
```
I take it that this should return a valid remainder `c` s.t. `c ≡ in mod hi`. There is something pretty odd about this, since modular arithmetic over negative dividends and divisors will give a shifted (by `hi`) or negated residue system. E.g., the complete residue system modulo -5 "should be" $\langle0, -1, -2, -3, -4\rangle$, but 
```supercollider
[-5,-4,-3,-2,-1,0].collect({ |a, i| a % -5 });
-> [0, 1, 2, 3, 4, 0]
``` 
and perhaps more shockingly,
```supercollider
[-10,-9,-8,-7,-6,-5].collect({ |a, i| a % -5 });
-> [0, -9, -8, -7, -6, 0]
// all x < -10 will wrap around to the same values as these
```
I can't think of another language off the top of my head that'll do something like this with `%` out-of-the-box, so I think this is worth mentioning wherever modulo is mentioned.

Anyway, implementation-wise, it seems possible to have a modulo for Signals consistent with the modulo for the rest of the applicable types, hence 6c572a7.

JordanHendersonMusic:
> I'm having a bit of trouble getting for myself your intuition about Signal powers from UGen powers—could you provide an example of how you'd expect a Signal to work, specifically in the context of a negative number?

Well ugens preserve the sign of the value with `pow` so you can use it as a kind of wave shaping, otherwise you'd get a DC offset.
```
some_sig.pow(\shape.kr(1))
```
If Signal, which is used almost exclusively for audio, behaved differently from ugen it would be unexpected. As both are used for audio. By letting them diverge we'd have two 'types' of audio.

This is the way I'm looking at it, but I admit things would be easier if pow behaved normally!

capital-G:
> Well ugens preserve the sign of the value with `pow` so you can use it as a kind of wave shaping, otherwise you'd get a DC offset.

TIL - this should be documented somewhere. It is definitely very useful, but not something I'd expect.

Maybe we should simply introduce an additional `pow` method which does not preserve the sign for signals.
@telephon is good at naming such things, maybe he has an idea :)

So a plan could be to

* implement `pow` for Signal which mimics `UGen.pow`, so preserving the sign of the signal
* implement `UGen.newPow` and `Signal.newPow` (WT) - `UGen.pow` / `Signal.pow` could use this implementation via `UGen.pow(sig) = UGen.sign(sig) * UGen.newPow(sig)` (but don't know if this is not as efficient as it could be?)

telephon:
Over time, there were often some methods missing in one or the other class. A while ago, e.g. I added `rand` for `UGen`, it is implemented for `Signal`. So this case is probably just an omission, and good that you take it up.

Ideally, all methods should work the same for UGen, Array, Number, Signal.

The special behaviour of `pow` is well known, there has been work towards adding a second method, `powReal` that behaves like the real part of the complex solution (https://github.com/supercollider/supercollider/issues/6526).

So add pow and mod for Signal now, behaving like in float or ugen, and then add powReal later also for signal. Does this sound good for you?

Thanks for taking it up!

passyur:
Ok, should be good now. In my testing, `mod` works the same for both `UGen`s and `Signal`s, and 06c6c05 ensures that the sign of the base for `pow` is preserved with `Signal`s like with `UGen`s. Please let me know if the way in which I wrote the unit tests is proper.

capital-G:
Haven't reviewed the code yet, but the formatting is a bit off, see https://github.com/supercollider/supercollider/pull/6659/checks

You can take a look at https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit how to format your contribution according to our standards automatically - you'd actually need to perform this step https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit#run-pre-commit-on-all-files

telephon:
(After reformatting should be fine!)

capital-G (changes requested):
A newly introduced test fails, see

https://github.com/supercollider/supercollider/actions/runs/13768499502/job/38501745389?pr=6659#step:7:110

This would need to be fixed before merging

telephon:
A possibility: Ah I was already wondering – you have entered floats explicitly, but their precision may not be sufficient. Perhaps it is better to calculate them, also using `pow` but by collect over the original array.

It also shows that the test message is not very informative ;-) : `Signal mod Float should work`.

passyur:
> Perhaps it is better to calculate them, also using pow but by collect over the original array.

Great idea. Also, it should be a bit more lenient, I think, i.e., it should compare two `FloatArray`s and ensure that each element is within 0.0001 of the other (with `assertArrayFloatEquals`). I say this because I get very, very small differences of size around 10^-8 for some ops (e.g., `3.1.pow(Signal[-2.0]` is slightly (about 5 x 10^-8 ) larger than `3.1.pow(-2.0)`).

> It also shows that the test message is not very informative ;-)

Ha, I'm not very fond of it. Now I want to say something like, "X OP Y should be within 0.0001 of [taking OP of] each element," which is somewhat verbose, but may be more informative.

telephon:
There is a method `assertArrayFloatEquals` that does what you want, I suppose?

telephon:
Yes, you should use `assertArrayFloatEquals`, also comes with a message.

testsuite/classlibrary/TestSignal.sc
```diff
@@ -58,4 +58,34 @@ TestSignal : UnitTest {
...
+		var test = Signal[1.5, 3.0, -2.0, 0.0].pow(Signal[1.8, 30.1, -2.0, 0.0]).as(FloatArray);
+		this.assertEquals(test, FloatArray[2.0747427940369, 2.2979996024832e+14, -0.25, 1.0], "Signal to power of Signal should work");
+	}
+
+	test_mod_xf {
+		var test = Signal[1.0, 3.1, 5.0, -8.0, 2.2].mod(2.2).as(FloatArray);
+		this.assertEquals(test, FloatArray[1.0, 0.89999985694885, 0.59999990463257, 0.79999995231628, 0.0], "Signal mod Float should work");
+	}
+
+	test_mod_fx {
```
telephon, line 82: I'd write it as:

```

var signal = Signal[1.0, 3.1, 5.0, -8.0, 2.2];
var array = signal.as(FloatArray);

var arrayResult = 2.2.mod(array);
var signalResult = 2.2.mod(signal);

this.assertArrayFloatEquals(signalResult.as(FloatArray), arrayResult);
```

telephon: … and similar for the other examples.

JordanHendersonMusic: Just in case this wasn't obvious, `FloatArray` and `Signal` are f32 arrays, but supercollider's inbuilt floating point type is f64. So you may  need to ease of the precision of the comparison significantly.

telephon: > any particular reason to have `arrayResult` and `signalResult` separate from `signal` and `array`?

I thought it would be less prone to typos if the definition is made only once, and more general for other cases.

passyur: My current plan is to avoid these particular precision-related edge cases in the test cases. (That isn't to say that I'm avoiding edge cases related to precision entirely.) I think it may be a separate issue that `2.2.mod(FloatArray[2.2])` gives `[2.2]`, but maybe this behavior is just par for the course with 64- vs. 32-bit floats.

telephon: Perhaps you can tune down the precision a litte in `assertArrayFloatEquals` – it is by default within `0.0001`, which may be too small for the case.

passyur: I might write a separate test case for large `Float`s that does that—the tolerance should still be pretty small for smaller `Float`s, I think.

testsuite/classlibrary/TestSignal.sc
```diff
@@ -60,32 +60,96 @@ TestSignal : UnitTest {
...
+		this.assertEquals(signalresult, arrayresult, "For large values, Float to power of Signal should give similar result to Float to power of Array while preserving sign of base");
 	}
 
 	test_pow_xx {
-		var test = Signal[1.5, 3.0, -2.0, 0.0].pow(Signal[1.8, 30.1, -2.0, 0.0]).as(FloatArray);
-		this.assertEquals(test, FloatArray[2.0747427940369, 2.2979996024832e+14, -0.25, 1.0], "Signal to power of Signal should work");
+		var signal = Signal[1.5, 3.0, -2.0, 0.0];
+		var array = [1.5, 3.0, -2.0, 0.0];
+		var expo = [1.8, 3.1, -2.0, 0.0];
+		var arraysign = array.collect { |base| if (base.isPositive, {1}, {-1}) };
```
telephon, line 109: In the present case, you could just write `var arraysign = array.sign`. That this will be zero for zero doesn't matter, because you are multiplying only with zero later.

testsuite/classlibrary/TestSignal.sc
```diff
@@ -60,32 +60,96 @@ TestSignal : UnitTest {
...
 	test_mod_xf {
-		var test = Signal[1.0, 3.1, 5.0, -8.0, 2.2].mod(2.2).as(FloatArray);
-		this.assertEquals(test, FloatArray[1.0, 0.89999985694885, 0.59999990463257, 0.79999995231628, 0.0], "Signal mod Float should work");
+		var signal = Signal[1.0, 3.1, 5.0, -8.0, 2.1];
+		var array = [1.0, 3.1, 5.0, -8.0, 2.1];
+		var modulus = 2.1;
+
+		var signalresult = signal.mod(modulus);
+		var arrayresult = array.mod(modulus);
+		this.assertArrayFloatEquals(signalresult.as(FloatArray), arrayresult.as(FloatArray), "Signal mod Float should give similar result to Array mod Float", 0.001);
```
telephon, line 133: give a result "for", not "to", I think.

testsuite/classlibrary/TestSignal.sc
telephon: But actually, I don't understand why you multiply the sign – all the versions should be exactly the same.

passyur: > In the present case, you could just write `var arraysign = array.sign`. That this will be zero for zero doesn't matter, because you are multiplying only with zero later.

The edge case I'm looking out for with zero is `0.pow(0)` = 1. Of course, `0.pow(x)` for x ≠ 0 will give 0, so then it doesn't matter. But it does matter for the former case.

> But actually, I don't understand why you multiply the sign – all the versions should be exactly the same?

Somewhere earlier in this thread it was mentioned that `pow` for `Signal`s should preserve sign like pow for `UGen`s—`Array`s do not preserve sign; that's why I multiply by base sign in the test cases. I take the absolute value of the array so that we can ignore sign changes based on the parity of the exponent, so multiplying by the base sign will "restore" the value to the base sign. 

I have multiplying by the sign in there in case someone wants to tool around with the test case and not have to change numerous other features to make the test work as it should. Maybe these details should be commented for clarity.

testsuite/classlibrary/TestSignal.sc
passyur: I'm partial to "to"—reads like "X is similar to Y," which seems more standard than "X is similar for Y," which would seem to imply that some feature of X somehow extends to Y. (E.g., "for" _would_ be more appropriate if the function for `Signal` mod `Float` were the same as what is happening with `Array` mod `Float`, but it isn't the same.)

testsuite/classlibrary/TestSignal.sc
telephon: No I think this is a misunderstanding. Something to the power of a negative number is a fraction of the corresponding positive number, e.g. $3^{-2} = 1 / (3^2) = 1/9$. What the difference to the standard behaviour is concerns the negative basis: e.g. $ -2^2 = 4$ normally, but when we are dealing with the message `pow`, it is different, then we keep the sign: $-2 ^2 = -4$. See discussion here: #6526.

So all the three should be the same, and are the same currently:
```supercollider

a = [1, 3, -2, 0];
b = 4.1;
{ (b pow: DC.kr(a)).poll }.play;
b pow: a;
b pow: a.as(Signal);


a = [1, 3, -2, 0];
b = -4.1;
{ (b pow: DC.kr(a)).poll }.play;
b pow: a;
b pow: a.as(Signal);
```

passyur: At least for the lines you commented on—under `test_pow_xx`—`var array` is the base and `var expo` is the exponent. I'm not entirely sure which array you're pointing to in your second sentence, but I'll assume the variable named `array`. The test only tracks the signs of values in `array`—`array.abs.pow(expo)` should just return `array.pow(expo)` with all entries as though `array` were all positive (the sign of `expo` is not touched). Then multiplying that pairwise with the signs of the bases (`array`) will give the desired result. So I think the behavior is as you describe it should be; i.e., we keep the sign of the bases, and perform `pow` otherwise as it usually is.

telephon: Ah sorry, indeed I was confused, because the variable `array` was used differently in different tests. If the behaviour is as I've described it then it's good!

testsuite/classlibrary/TestSignal.sc
telephon: Now I understand, yes, keep it. Thanks!

JordanHendersonMusic (changes requested):
Sorry for jumping in so late, but I only just noticed this.

I have a suggestion regarding precision. As you've noticed in your tests there is often difference of precision between doing `[..].pow(2.0)` and `Signal[...].pow(2.0)`, where possible, I think this should be minimised.

Right now you have this
```cpp
SetObject(a, signal_pow_fx(g, slotRawFloat(a), slotRawObject(b)));
```
`slotRawFloat` actually returns a `double`, as the vm only works on doubles. Precision is lost here through the implicit (dam c++) conversion through the function interface.

I think it would be better if the function `signal_mod|pow_XX` where changed to be either:

* signal_Y_xx(float*, float*)
* signal_Y_xf(float*, double)
* signal_Y_fx(double, float*)

Then this part `sc_mod(ina, *b)` should be rewritten to convert things to doubles in *all* cases.
`static_cast<float>(sc_mod(static_cast<double>(ina), static_cast<double>(*b)))`.

***What a mouthful***

This way, the exponent and mod is always performed on doubles, and converted to floats only at the last moment. This should allow for the precision in the tests to refined further, as a precision of 3 significant figures is rather poor.

JordanHendersonMusic:
Hmm, there is a bit of a conflict there, do we align with `FloatArray`'s results or attempt to align with `Array`'s. 

Honestly, I think `Signal` is a bit of a problematic part of the code, the fact the server works with single and the language with double precision creates this tension.

After sitting on this, I'd be quite happy if the precision was made to either match `FloatArray`'s or get closer to the language. Whatever you think @passyur .

Using a float array with one element is *I think* the only way to work with single precision is SC - might make comparisons easier.

passyur:
A thought: to have it get closer to the language feels like a bigger change—maybe even a different issue. I do think it should be closer to the language, or that the language should act as though it is closer to Signal when doing these precision-sensitive operations, in the end. The latter is what it's currently (aa8020d) doing, actually, via the cast from double to float. So the suggestions are sort of like (same for `pow`):

`double % [float] => double % [double] => [double] => [float]` versus `double % [float] => float % [float] => [float]`. 

For now, I feel most comfortable having Signal behave more closely like FloatArray since Signal is defined as a FloatArray but am not opposed to opening an issue about what to do about this weird behavior. So, for now, I’ll just do as written in my last comment and change the tests to compare against FloatArrays.

dyfer:
@JordanHendersonMusic @telephon @capital-G what's missing to approve this PR?

passyur:
Since ´Signal´ is a ´FloatArray´, I believe that all comparisons between the two should be _exact_, not just close (as that function you mention will look within a certain margin of error).

JordanHendersonMusic (approved):

JordanHendersonMusic:
@telephon I see you've approved this, just wanted to check you don't have any remaining concerns because it was a while ago. If not, then I think this can be merged.

capital-G (approved):
Thanks for the implementation!

PR #6660, opened 2025-03-07
Use aqt for macOS x64 in CI and switch macOS system lib build to arm

### LINKED ISSUE #6648: macOS nightly build fails due to qt update from homebrew
Currently we have some problems regarding Qt6 - this could be due to Qt updates from homebrew?
My local build still somehow succeeds though? (maybe there is a difference in silicon vs intel mac?)

See e.g. a currently failing build https://github.com/supercollider/supercollider/actions/runs/13587575374/job/37985931506#step:13:89 using

```
 ==> Installing qt
🍺  /usr/local/Cellar/qt/6.8.2: 16,372 files, 348.2MB
```

which fails

```
 -- Could NOT find Qt6WebEngineCore (missing: Qt6WebEngineCore_DIR)
CMake Error at QtCollider/CMakeLists.txt:38 (find_package):
  Found package configuration file:

    /usr/local/lib/cmake/Qt6/Qt6Config.cmake

  but it set Qt6_FOUND to FALSE so package "Qt6" is considered to be NOT
  FOUND.  Reason given by package:

  Failed to find required Qt component "WebEngineCore".

  Expected Config file at
  "/usr/local/lib/cmake/Qt6WebEngineCore/Qt6WebEngineCoreConfig.cmake" does
  NOT exist



  Configuring with --debug-find-pkg=Qt6WebEngineCore might reveal details why
  the package was not found.

  Configuring with -DQT_DEBUG_FIND_PACKAGE=ON will print the values of some
  of the path variables that find_package uses to try and find the package.

  lang/CMakeLists.txt:184 (include)
```

The last build from dev which still worked was https://github.com/supercollider/supercollider/actions/runs/13252984681/job/36994687033 which was using

```
 ==> Installing qt
==> Caveats
You can add Homebrew's Qt to QtCreator's "Qt Versions" in:
  Preferences > Qt Versions > Link with Qt...
pressing "Choose..." and selecting as the Qt installation path:
  /usr/local
==> Summary
🍺  /usr/local/Cellar/qt/6.7.3: 15,091 files, 679.6MB
```

See https://github.com/Homebrew/homebrew-core/commits/555cc061aa19066b0e0428e8858c231de0c4f70a/Formula/q/qt.rb for diff in the homebrew formula.

There is also 300 mbs in diff between both versions which seems suspicious.

capital-G:
## Purpose and Motivation

Fixes #6648 - See https://github.com/supercollider/supercollider/issues/6648#issuecomment-2706222529 for more details.

Essentially it gets harder to build SuperCollider on older macOS x64 platforms due to lacking support from Hombrew (new qt version 6.8.2 ships without webengine on macOS 13 on x64 and it is somehow hard to install older versions using homebrew. Although the most recent macOS 15 still works w/ x64 macs, this is not supported by GitHub anymore...).

This PR bumps Qt to 6.8.2 in our macOS CI and switches completely to aqt for installing Qt on macOS (instead of using homebrew).
This provides a more stable CI, but may create a divergence on how users build SuperCollider locally and how we build in CI, but that's just the way it is now it seems :/
Maybe we can introduce an macOS arm build which is using qt from homebrew, but this is something for another PR.

I thought about adding a disclaimer for macOS 13 users but decided against this for now.

Edit: This PR has a _high priority_ as our CI is currently failing b/c of this and makes reviewing a bit more difficult.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -55,11 +55,11 @@ jobs:
             extra-cmake-flags: ""
             artifact-suffix: "macOS-x64-legacy" # set if needed - will trigger artifact upload
 
-          - job-name: "x64 use system libraries"
-            os-version: "13"
-            xcode-version: "15.2"
+          - job-name: "arm64 use system libraries"
+            os-version: "15"
+            xcode-version: "16.0"
             deployment-target: ""
-            cmake-architectures: x86_64
+            cmake-architectures: "arm64"
             homebrew-packages: "qt@6 libsndfile readline fftw portaudio yaml-cpp boost"
             vcpkg-packages: ""
             vcpkg-triplet: ""
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -68,9 +68,11 @@ jobs:
           - job-name: "x64 shared libscsynth"
             os-version: "13"
             xcode-version: "15.2"
+            qt-version: "6.7.3" # will use qt from aqtinstall
+            qt-modules: 'qtwebengine qtwebchannel qtwebsockets qtpositioning'
             deployment-target: ""
             cmake-architectures: x86_64
-            homebrew-packages: "qt@6 libsndfile readline fftw portaudio"
+            homebrew-packages: "libsndfile readline fftw portaudio"
             vcpkg-packages: ""
             vcpkg-triplet: ""
             extra-cmake-flags: "-D LIBSCSYNTH=ON"


dyfer:
Thanks @capital-G for looking into this! I
IIUC Qt 6.8 raises the minimum macOS version to macOS 12. My slight preference would be to stay at Qt 6.7.x for the releases for the time being.
I'm in favor of switching all builds to aqt for the time being. However, it is a larger and important discussion how to proceed to support building SC locally.

dyfer:
Actually - I think I'd prefer to fix this by switching the "system libraries" build to a newer arm64 image. The purpose of that job is to test building locally using homebrew etc. IIUC qt in homebrew still ships with the webengine on macos 14+, correct?

capital-G:
> Actually - I think I'd prefer to fix this by switching the "system libraries" build to a newer arm64 image. The purpose of that job is to test building locally using homebrew etc. IIUC qt in homebrew still ships with the webengine on macos 14+, correct?

IIUC yes and sounds like a good idea!
So basically the `system libraries` would become the "build like @ home on an arm mac" build? This would give us hints if any implicit brew upgrades breaks the build by being our _"bleeding edge"_ build

> IIUC Qt 6.8 raises the minimum macOS version to macOS 12. My slight preference would be to stay at Qt 6.7.x for the main release build for the time being.

Thx for pointing this out, well spotted.

We should probably indeed stick to Qt 6.7 via aqt on all builds (until it creates some other problems) except `system_libraries` which would use qt (6.8) via brew?
Shall I adjust the PR for this?

dyfer:
> IIUC yes and sounds like a good idea! So basically the `system libraries` would become the "build like @ home on an arm mac" build? 

Yes!
To clarify, previously "system libraries" was mainly meant to test against boost from homebrew (at that time the release SC builds used Qt from homebrew). As we moved away from Qt from homebrew for the releases (to increase backwards compatibility), I consider the "system libraries" job to test for both homebrew boost, as well as homebrew qt.

BTW It can certainly be discussed (separately) whether testing compatibility with homebrew boost is reasonable/needed or not.

> We should probably indeed stick to Qt 6.7 via aqt on all builds (until it creates some other problems) except `system_libraries` which would use qt (6.8) via brew? Shall I adjust the PR for this?

Yes, please! That would be my preference.

capital-G:
Updated the PR.

@dyfer can you double-check the caching of aqt? Seems correct to me b/c we either have macOS 13 (x64) or macOS 15 (arm), so we shouldn't need to add an architecture flag to it.

dyfer:
@capital-G thanks for the changes.

> @dyfer can you double-check the caching of aqt? Seems correct to me

Yes, it's correct, AFAICT. 
In any case, downloaded qt libraries are "universal binaries" on macOS and there are no specific builds for different platforms, IIUC. Caching is handled automatically by the aqtinstall action, so I trust it does the right thing, caching based on qt version and chosen modules.

dyfer (approved):

tedmoore:
I just tried building and got the error below. For building "at home" is the plan also to use aqtinstall to get the WebEngine?

Mac M2 Max
Ventura 13.5.2

```
CMake Error at QtCollider/CMakeLists.txt:38 (find_package):
  Found package configuration file:

    /opt/homebrew/lib/cmake/Qt6/Qt6Config.cmake

  but it set Qt6_FOUND to FALSE so package "Qt6" is considered to be NOT
  FOUND.  Reason given by package:

  Failed to find required Qt component "WebEngineCore".

  Expected Config file at
  "/opt/homebrew/lib/cmake/Qt6WebEngineCore/Qt6WebEngineCoreConfig.cmake"
  does NOT exist



  Configuring with --debug-find-pkg=Qt6WebEngineCore might reveal details why
  the package was not found.

  Configuring with -DQT_DEBUG_FIND_PACKAGE=ON will print the values of some
  of the path variables that find_package uses to try and find the package.

  lang/CMakeLists.txt:186 (include)


```

dyfer:
> For building "at home" is the plan also to use aqtinstall to get the WebEngine?

Typically no... Qt from homebrew ships with the WebEngine, EXCEPT on "older" OSs, as of recently. I'm not sure what the easiest solution is... possible solutions include:
- installing older Qt in homebrew (I'm not sure if that's even possible these days)
- updating the OS to at least macOS 14
- using aqt or official Qt installer, and adding qt to path

tedmoore:
Ah, ok, I can update my OS.

tedmoore:
For what it's worth. I'm now on MacOS 15 and everything built successfully. Thanks @dyfer !

PR #6664, opened 2025-03-11
plugins: EnvGen: fix endtime

### LINKED ISSUE #6180: EnvGen has incorrect duration
## Environment

* SuperCollider version: 3.14.0-dev
* Operating system: macOS

## Steps to reproduce

UPDATE: the second channel uses the standard `Env.new`, and still shows the duration problem, so it looks like the problem is in `EnvGen`, specifically running at `kr`.
```supercollider
(
{[
	EnvGen.kr(Env.circle([0, 1, 0.3], [0.02, 0.03, 0.05])),
	EnvGen.kr(Env(
		[0, 1, 0.3].dup(10).flat ++ 0,
		[0.02, 0.03, 0.05].dup(10).flat // duration sums to 1.0
	))
]}.plot(1)
)
```
Screenshot 2024-01-05 at 11 54 47

Note the hold behavior of the first segment of the first channel (using `Env.circle`) looks to be a separate issue.

In the second channel, the `times` of `Env` sum to `1.0` but the generated signal completes the envelope before then.

#### Original `Env.circle` example:
```supercollider
{EnvGen.kr(Env.circle([0, 1, 0.3], [0.02, 0.03, 0.05]))}.plot(0.4)
```

####  Expected vs. actual behavior
In the example above, the `times` array sums to `0.1`, but as seen in the plot, the cycle period is less than 0.1.
Screenshot 2024-01-04 at 09 47 38

mtmccrea:
## Purpose and Motivation

EnvGen accumulates rounding error over the durations of its segments. This can be especially noticeable if the Env has many breakpoints, especially at control rate. 

Another example is if you're expecting equivalence between, e.g., Line and linear segment of an Envelope, because of the incorrect Env duration, the step size is incorrect so sample values aren't identical to a Line with the same nominal duration.

For example, this envelope's duration is 1.0 sec but ends early:
```supercollider
(
{	EnvGen.kr(Env(
		[0, 1, 0.3].dup(10).flat ++ 0,
		[0.02, 0.03, 0.05].dup(10).flat // duration sums to 1.0
	))
}.plot(1)
)
```
Screenshot 2025-03-18 at 11 11 14

Fixes #6180. 

This follows on #6175, where the duration bug was found and partially addressed.

## Types of changes

- Documentation
- Bug fix
- Tests added
  - equivalence test for Line vs. linear Env, XLine vs. exp Env
  - test correct env times for all curve types) and modified (suppress output of successful tests)
  - replaces previous start/end sample value test with this more comprehensive test that tests sample values and timings
- Breaking change (technically breaking - the output changes)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Env.schelp
+++ b/HelpSource/Classes/Env.schelp
@@ -471,7 +471,7 @@ The curvature of the envelope segments. The last value is the looparound curve. 
 The array size of strong::curve:: should be the same size of the strong::levels:: array.
 
 note::
-If a single value is given for strong::times:: or strong::curve::, or if the array is too short, it will be extended by wrapping around the given values.
+If the array of strong::times:: or strong::curve:: is too short, it will be extended by wrapping around the given values. If a single value is given, it applies to all envelope stages.
 ::
 
 discussion::
--- a/HelpSource/Classes/Env.schelp
+++ b/HelpSource/Classes/Env.schelp
@@ -471,7 +471,7 @@ The curvature of the envelope segments. The last value is the looparound curve. 
 The array size of strong::curve:: should be the same size of the strong::levels:: array.
 
 note::
-If a single value is given for strong::times:: or strong::curve::, or if the array is too short, it will be extended by wrapping around the given values
+If a single value is given for strong::times:: or strong::curve::, or if the array is too short, it will be extended by wrapping around the given values.
 ::
 
 discussion::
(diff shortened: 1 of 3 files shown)


telephon (approved):
I've compiled, run the tests you have written. It reads correct, although I didn't try to understand the logic behind the envelopes.

joshpar (approved):

joshpar:
Looks good here, and great tests!

dyfer:
Should this be squashed when merging?

mtmccrea:
> Should this be squashed when merging?

yes definitely, thanks!

dyfer:
> yes definitely, thanks! (I included the `squash on merge` tag)

Ah, completely missed that! Thanks!

PR #6656, opened 2025-03-05
DetectIndex [Issue #6641]

HotwheelsSisyphus:
## Purpose and Motivation

I was drawn in by the "good first issue" tag...
https://github.com/supercollider/supercollider/issues/6641

Obviously I'm still struggling with git. I thought this would be just 2 commits from today but now this includes some older business as well... apologies. I should also test if that UGen still works...

## Types of changes

- Documentation: as requested; plus found a typo in the ServerPluginsAPI file.
- Bug fix (?): 
in the source for the audio rate calc, it seems like the `prev = inval` assignment should go inside the body of the {statement} to be evaluated:
``` 
LOOP1(
        inNumSamples, inval = ZXP(in);
        if (inval != prev) { prevIndex = DetectIndex_FindIndex(table, inval, maxindex); } prev = inval; //this one here
        ZXP(out) = (float)prevIndex;);
```
because if `prev == inval` already, not much sense in assigning it over again. I guess it doesn't do any harm in its present form, but it is a lot of unnecessary assignments... I haven't tested the changes though, and am also new to cpp stuff. Maybe there is a good reason for this after all?

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [X] Updated documentation
- [ ] This PR is ready for review

Issue #6641, opened 2025-02-24
Improve `DetectIndex` documentation 


JordanHendersonMusic:
## Motivation

The following things are not addressed in the documentation for `DetectIndex`.

What happens if the value can't be found?
What happens if there are duplicates, which index does it return?
What kind of search does it do - performance concerns?

## Plan for Implementation

Good first issue. Noncritical.

JordanHendersonMusic:
Yes I thought this was closed automatically.

PR #6666, opened 2025-03-13
Docs: DetectIndex: add doc

HotwheelsSisyphus:
As requested in Issue https://github.com/supercollider/supercollider/issues/6641.
First PR attempt https://github.com/supercollider/supercollider/pull/6656 had messy commit history, hopefully this one is better. 

Documents return values, search method, behavior in case of multiple possible matches.
Change to code comment for clarity.

## Purpose and Motivation
@JordanHendersonMusic's issue

## Types of changes

- Documentation

### DIFF
--- a/HelpSource/Classes/DetectIndex.schelp
+++ b/HelpSource/Classes/DetectIndex.schelp
@@ -7,7 +7,8 @@ Linearly searches a buffer for a value (i.e., loops over the frames in the
 buffer from the beginning, checking for float equality), and returns index of first occurrence of said value.
 Returns -1 if value is not found.
 The object performs a search whenever a change in the strong::in:: argument is
-detected.
+detected; 
+changes in the strong::bufnum:: argument do not trigger searches.
 Checks for changes occur at the selected rate (ar or kr).
 
 classmethods::
--- a/HelpSource/Classes/DetectIndex.schelp
+++ b/HelpSource/Classes/DetectIndex.schelp
@@ -3,7 +3,12 @@ summary:: Search a buffer for a value
 categories:: UGens>Buffer
 
 description::
-Search a buffer for a value.
+Linearly searches a buffer for a value (i.e., loops over the frames in the
+buffer from the beginning, checking for float equality), and returns index of first occurrence of said value.
+Returns -1 if value is not found.
+The object performs a search whenever a change in the strong::in:: argument is
+detected.
+Checks for changes occur at the selected rate (ar or kr).
 
 classmethods::
 method:: ar, kr
@@ -13,7 +18,7 @@ index of the buffer
 argument:: in
 the input signal.
 returns::
-index
+index of first occurrence of value, or -1 if value is not found.
 
 examples::
 code::
@@ -30,7 +35,7 @@ s.listSendMsg(b.allocMsg(b.setnMsg(0, t)));
 	var index, in, out, f0, fdiff;
 	var bufnum = b;
 	var input;
-	input = MouseX.kr(0, max).round(1); // round to precision
+	input = MouseX.kr(0, max).round(1); // round to integer
 	index = DetectIndex.kr(bufnum, input);
 	index.poll;
 	SinOsc.ar(index.linexp(0, max, 200, 700)) * 0.1


HelpSource/Classes/DetectIndex.schelp
```diff
@@ -3,7 +3,12 @@ summary:: Search a buffer for a value
 categories:: UGens>Buffer
 
 description::
-Search a buffer for a value.
+Linearly searches a buffer for a value (i.e., loops over the frames in the
+buffer from the beginning, checking for float equality), and returns index of first occurrence of said value.
+Returns -1 if value is not found.
+The object performs a search whenever a change in the strong::in:: argument is
```
JordanHendersonMusic, line 9: Does it also do a new search when the buffer changes?

JordanHendersonMusic (commented):
Thanks for this! This PR looks good, clear and concise. Did you find this information by reading the code, or through experimentation?

HotwheelsSisyphus:
Yup, the good first issue tag, but also I'm generally focusing on contributing to doc improvements. I've been more active on the forum, though confusingly under a different username (e.g. https://scsynth.org/t/schelp-style-guidelines-discussion-coordination-thread/11285).
And while there's tons of sc documentation that is lacking, I figured if somebody is asking for something specific, that at least gives me somewhere to start that doesn't just reflect my own preoccupations of the day... also, I wanted to get a better sense of how the scsynth insides work, so I guess that's an opportunity to do something productive while doing that.
So yes, I mostly looked at the code. I didn't do that much experimenting.

HelpSource/Classes/DetectIndex.schelp
HotwheelsSisyphus: No, buffer updates don't trigger a search. I should probably add a note to that effect?

HotwheelsSisyphus: That's possible... but I don't currently know how/if buffers keep track of their changed/unchanged state?

HotwheelsSisyphus: did some haphazard digging, there is a SndBufUpdates struct that seems meant to keep track of this, but it seems that BufWr or RecordBuf don't use it, at least not directly...

JordanHendersonMusic: Ah sorry I meant if buffnum argument to the plugin changes.

HotwheelsSisyphus: seems doable, but I don't have a huge amount of experience writing cpp code (reading it is easier, obviously). I'd be happy to give it a try, but you'd want it to be reviewed thoroughly...

JordanHendersonMusic: Sorry, I'm not being clear!

This ugen has two arguments, the buffnum and the value. When the value changes the ugen does a search, as your documentation states. What happens when the bufnum changes? Does a search happen? I think this PR should add that answer to the documentation (whatever it may be, I don't know) to be complete.

However, I would expect the ugen to do a search when the buffer changes, if it doesn't then perhaps this is a bug/ommision and we (sc developers) should make a new issue and try to fix it in the future. 

I'm only suggesting that it should be **documented** here, not requiring you fix it (if it is indeed a bug) before this PR can progress - sorry for the confusion!

HotwheelsSisyphus: no worries. I'll add a sentence to that effect... git/gh newbie question: should I add another commit or amend the present one? Squash? Gently crush with a mortar and pestle?

JordanHendersonMusic: I'd add a new commit. When this is merged into the main `develop` branch all the commits here will be squashed (technical term) down into a single commit (unless there is a reason not to). Having seperate commits at this stage just makes things easier for us to review.

HotwheelsSisyphus: alright, I'll do that then.
yeah I am aware that these are technical terms, I'm just still new enough to this that I find them funny.

JordanHendersonMusic (approved):

JordanHendersonMusic:
@telephon @capital-G I saw you both commented on the earlier draft of this. Do you think changing the buffnum should retrigger a search? It would be a different issue.

telephon:
Yes, I quickly looked into it an hour ago, it would be easy to do, but it would be another issue. If we say we want to change it, the explicit comment in the docs could be left out for now.

JordanHendersonMusic:
@HotwheelsSisyphus Sorry for the back on forth on this. Lets revert the previous commit and just not mention the bufnum argument for now. This way, this PR can be merged, and once #6682 is done, the documentation can be adjusted.

PR #6671, opened 2025-03-16
sclang: optionally exit when a fatal error is encountered

scztt:
## Purpose and Motivation

There is currently only one case of fatalerror - grey count errors. When these are encountered, sclang is no longer operable - the reasonable thing here is to quit rather than leave sclang in a half-constructed and unresponsive state.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


scztt:
I kept the old behavior and exposed it with an env variable, in case for some reason this is useful for debugging? I don't see a reason to not exit by default in this case: sclang is unresponsive and there's no possibility of recovery.

For context, this is important in cases where SuperCollider is launched by another process, in case of e.g. an installation or when using the language server. If a grey count error is encountered on launch, the sclang process is unresponsive but there's no way for the launching process to detect the error state.

capital-G (commented):
Thanks for detecting and fixing this scott!

I am somehow unsure if we should introduce to control such behavior using ENV variables, it seems more sensible to use CLI options for this (see e.g. #6640 which allows to introduce expressive multi-char flags) as this (obscure) behavior would be clearly exposed to the user by implicit documention.

Also, is `-1` a valid exit code? https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html suggests to use the 0-255 range (does it underflow to 255?)

scztt:
I personally would opt to not use an env flag at all, and just always quit, unless anyone can think of a reason why we would want to halt sclang in the broken state (note that even if someone is debugging a GC issue, it's actually better to exit here anyway bc a throw could lose the stack context).

I'm open to suggestions re the error code. -1 was there already (in the comment out exit). I did a quick search and see that `exit(-1)` is used here and there, including in boost, so I don't think it's improper.

JordanHendersonMusic:
I'm not at my pc right now, but I couldn't find where `mNumGreys` was written to? It seems it is a debug feature? Since SC uses linked lists here and moves all the objects at once it shouldn't be iterating over things it recycles, so doesn't need a count. This would mean this method is never called... But I might have missed it as I can't search on mobile.

I don't see the need for an environment variable here, if there is no usable information then exit.


CppRef says to use `EXIT_FAILURE`.

capital-G (approved):
If we ever need to change this we can change it again - but as this is not affecting user code in a negative way I think it is fine to just quit.

I am fine either way, but would prefer an implementation w/o if b/c it generates question why this exists.

dyfer:
I think this can be merged, but ideally we would add this env variable to the documentation _somewhere_. This PR or another?

dyfer:
> I'm open to suggestions re the error code. -1 was there already (in the commented out `exit`). I did a quick search and see that `exit(-1)` is used here and there, including in boost, so I don't think it's improper.

Is an exit code a unit8, i.e. wouldn't `-1` overflow? I'm not sure if that makes it improper though.

dyfer:
@scztt what do you think about updating this PR to use `EXIT_FAILURE` and removing the environment variable?

Spacechild1:
Hi, we have just been discussing this on the dev meeting. What exactly is the problem with the current implementation that just throws an exception? If the exception isn't caught, the process should terminate anyway.

In theory, sclang may be embedded in applications, in which case `exit()` would also kill the host program.

Spacechild1:
Aha, I see! `runLibrary()` just swallows all exceptions. I think we just need some custom exception type that we can throw in the case of critical errors. We would then make sure that this exception is not swallowed, so it can eventually be caught by the host application (who would then likely terminate the program).

Side note: I'm not sure about the `catch(...)` after `catch(std::exception&)`. Do we really want to swallow *any* exception?

JordanHendersonMusic:
Does #6951 address this instead?

JordanHendersonMusic:
@scztt I'm going to close this PR because I think the other one addresses your use case and has already been merged. Do reopen if you think I'm wrong! Just trying to clear up the approved but not yet merged PRs as there are a few.

PR #6673, opened 2025-03-17
Classlib:Plot:remove interpolation for resampling of integer data

HotwheelsSisyphus:
Not sure if this was the best use of my time, but here goes:

Adds a check in Plot.value_ whether codomain is integer (tested with `any(_.isInteger.not)`); if so, turns off interpolation in prResampValues.

Only has an effect when `Plot.value.size` is larger than approx. `Plotter.bounds.width - 50` or so(depending on labels etc.).

## Purpose and Motivation

When the aren't enough pixels available, Plotter will internally resample the data from a linearly interpolated version of `.value`. This is suitable for most pseudo-continuous data such as audio signals.
Where the data is discrete, however, this behavior may occasionally be unwanted. 
Here is an example adapted from Ppoisson.schelp. Width and data size are unmodified, I just removed the histogram and added the missing Window.front:
 ```
(
var a, c, w;
a = Ppoisson(10.0, inf);
c = a.asStream.nextN(500);
w = Window.new("Ppoisson", Rect(10, 10, 540, 400));
// plot the values
c.plot(bounds: Rect(10, 10, 520, 380), discrete: true, parent: w);
w.front
)
```
Because the `Plotter.bounds.width` is too small for the data (`value.size`), you have to enlarge the width to see that the distribution is discrete (as it should be). 
 
I guess in this case, it may be preferable to simply skip values rather than interpolate. With this PR, Plotter does this when value is an array of integers. That is, if the `interp` flag is set to false, instead of getting the values to be drawn from `value.blendAt(scaledIndex)`, plot takes them from `value.at(scaledIndex.asInteger)`. 

For pseudo-continuous signals that happen to be all in integers this slightly changes the behavior (from linear interpolation to no interpolation), but obviously only when there aren't enough pixels. I think this side effect is even more niche than the PR itself? 
But users can still effectively choose the behavior by passing in arrays of the respective datatype; i.e., if they want interpolation on some array of integers, just convert that to float before you pass it to plot.

## Types of changes

- Bug fix 

## To-do list
Tests: Tested on my installation, not sure if there are more serious tests I should be doing?

Doc: I don't think this fix itself needs to be documented, but maybe there should be some more explicit documentation of Plot's resampling behavior and how many pixels are needed to prevent resampling. Resampling is currently only briefly discussed in the examples, and there with respect to a DomainSpec; but (from what I could tell) the spec is implemented with continuous data in mind (there's a `spec.warp.unmap` somewhere, so the step argument to the spec won't do anything), so I don't think it can be used to solve the above issue.

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -209,7 +209,7 @@ Plot {
 	value_ { |array|
 		value = array;
 		valueCache = nil;
-		interp = value.any { |v| v%1 > 0};
+		interp = value.any(_.isInteger.not);
 	}
 	spec_ { |sp|
 		spec = sp;


dyfer:
Thanks @HotwheelsSisyphus 

> Not sure if this was the best use of my time

I know the feeling...

Looks like a good addition on the first glance, but I'd need to think more about it.

@mtmccrea as you've worked with plotting quite a bit, do you have any thoughts on this?

mtmccrea:
Thanks @HotwheelsSisyphus for the proposal.

I agree the interpolated data is problematic for a few reasons (a big one also being speed).

But I don't think the interpolation behavior should result from whether `data` are integers or floats. That's an single case and we probably want a more general solution.

I've thought about this in the past and thought a good solution would be to do as the `ScopeView` does: when the view width is less than the data size, any two adjacent pixels/data show the max and min values for that span of data.
https://github.com/supercollider/supercollider/blob/9b0515a8de92aeee7d8f280d9bde17df57cff395/QtCollider/widgets/QcScopeShm.cpp#L249-L279

I could also see your solution of just dropping data to be an optional behavior (e.g. for speed). It could be that the default behavior would be specific to the `plotMode`, but would take further consideration.

HotwheelsSisyphus:
Thanks for the response! Agree that the min/max approach is a better idea (for integers...); also that there should be some way for the user to control/override the behavior. 

But the motivation for me is that one doesn't want to confuse people... And I'd think that getting real output from a discrete distribution is more confusing than some values going missing. (E.g., I initially thought Ppoisson was broken before I realized that .plot was interpolating.) 

All of this is probably really not that important, but just once again to clarify the motivation (it boils down to the question of why/when one plots Arrays of integers):

My thinking is that if I want to plot something with more datapoints than I have pixels for, that will be something where individual values won't play a large role. In the float case, one might still want relatively "smooth" waveform displays (one looks at the pseudo-continuous curve, not at individual values); hence linear interpolation is a better bet than other approaches because it minimizes kinks in the curve (and cubic etc. is unnecessary, on that I agree).  Otoh, if I'm using integers, I'm again probably looking at global features of the data (such as: discrete or not?), but not so much in terms of curves/slopes etc., because integers aren't terribly well suited to express these things.  

Re speed:
I didn't think of speed here, but I'm not a computer scientist... I just never had issues with plot in  that respect, as I don't call it that often... Would it make it more efficient to move the interpolation (or whatever other procedure we settle on) inside the qc code rather than the class lib?

But as I said, I don't think this is a terribly important PR, feel free to close.

mtmccrea:
@HotwheelsSisyphus this may be the beginning of a discussion on interpolation behavior in Plotter, which would be better suited for an Issue or a forum thread. So perhaps this could be closed but still referenced for context elsewhere if needed.

PR #6676, opened 2025-03-20
Update Server.schelp

carltesta:
## Purpose and Motivation

Simple Help file fix: I couldn't get the Remote Server option working with the Help documentation as is. I had to add a line about making the remote server the default server and think adding this to the Help documentation would help other users going forward.

## Types of changes

- Documentation

## To-do list

- [X ] Code is tested
- [X] All tests are passing
- [X] Updated documentation

### DIFF
--- a/HelpSource/Classes/Server.schelp
+++ b/HelpSource/Classes/Server.schelp
@@ -89,6 +89,7 @@ o = ServerOptions.new;
 o.maxLogins = 2;
 t = Server.remote(\remote, NetAddr("192.168.0.130", 57110), o); // set to correct address and port
 // info about returned client ID should be posted in the post window
+Server.default = t; //set your new remote server to be default
 t.makeWindow; // make a window for monitoring
 )
 ::


telephon:
> I couldn't get the Remote Server option working with the Help documentation as is.

Often, you'd also want to keep both local and remote servers. To play on the remote server, one has to pass it as an argument to patterns or synths.

What functionality did you find that didn't work? Making a remote server default is one simple way to use it of course.

carltesta:
Ah yes, I understand now. I pretty much expected it to work with normal Function.play type coding and when it didn’t decided to make it the default server to make that workflow easier.

What would be an example that we could add to the help file to make it clearer how to get sound out of the remote server?

telephon:
Yes, definitely. Maybe you could add an example below the one, where you show how to play some nice little sound on the remote server? A note may have to be added that the server needs to be booted on the remote computer.

dyfer:
My preference also would be to not set it as default, but maybe adding _below_ the code block:

```supercollider
// if you want to use the remote server without the need of explicitly passing it to e.g. Synth or Function -play, set it to be the default server:
Server.default = t;
```

I also like @telephon 's idea of adding some sound playing on the remote server (nicer than this one, but here's an example): `{PinkNoise.ar(-12.dbamp)}.play(target: t);`

dyfer:
@carltesta would you consider making the requested changes?

carltesta:
Will do!

dyfer:
@carltesta are you still planning on getting back to this?

dyfer (approved):
Great, thanks!

PR #6677, opened 2025-03-21
tests: add test in lcm/gcd

telephon:
Add test for fundamental theorem of arithmetic.

## Purpose and Motivation

This is an important property of gcd and lcm, make sure it passes.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing


JordanHendersonMusic (approved):
Seems like a simple merge, I can't think of a case where this shouldn't hold.

PR #6678, opened 2025-03-22
Move filenameSym from Method to FunctionDef

### LINKED ISSUE #6675: scheduled functions should know the path in which they have been defined
## Motivation
Within a scheduled function, we cannot access the path from which it was created. This is an issue when trying to load another file with a relative path.

In particular, it should be possible to schedule any function to be called later. This functionality is broken by the lack of access to `thisProcess.nowExecutingPath`. For example, `resolveRelative` is recommended for file loading, but cannot be scheduled.

```supercollider
(
"directly called, path is: %".format(thisProcess.nowExecutingPath).postln;
Routine { 1.wait; "scheduled in routine, path is: %".format(thisProcess.nowExecutingPath).postln }.play(AppClock);
AppClock.sched(0.5, { "scheduled, path is: %".format(thisProcess.nowExecutingPath).postln });
)
```
The first two return a path, the third doesn't.

For discussion see: #4607.

As @jamshark70 has proposed, functions could save their paths, like Routines do.

## Plan for Implementation

The instance variable `filenameSymbol` could be moved up from `Method` to `FunctionDef`. It could then be populated in a similar way as `Routine` (`Thread`) populates its `executingPath`.

Routine does it in the primitive `prRoutineResume`:
```c++
slotCopy(&thread->oldExecutingPath, &g->process->nowExecutingPath);
slotCopy(&g->process->nowExecutingPath, &thread->executingPath);
```

Special care must be taken with exceptions, but it seems that this is done already in Thread any may not concern functions.

Open who will do it.

JordanHendersonMusic:
## Purpose and Motivation

Closes #6675

`thisProcess.nowExecutingPath` doesn't return the file path where the code was defined. 

This PR moves `filenameSym` from `Method` to `FunctionDef` (@jamshark70 's suggestion).

Save this code in a file. Either run it directly from the file, or use `String.load`.

```supercollider
thisFunctionDef.filenameSymbol.debug(\a);

fork { thisFunctionDef.filenameSymbol.debug(\b) };
```

## Types of changes

- New feature

## To-do list

I haven't run the full test suits, and a test is yet to be added, but the implementation should be there. 

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Core/Kernel.sc
+++ b/SCClassLibrary/Common/Core/Kernel.sc
@@ -415,6 +415,7 @@ Process {
 FunctionDef {
 	var raw1, raw2, <code, <selectors, <constants, <prototypeFrame, <context, <argNames, <varNames;
 	var <sourceCode;
+	var <filenameSymbol;
 
 	// a FunctionDef is defined by a code within curly braces {}
 	// When you use a FunctionDef in your code it gets pushed on the stack
@@ -563,7 +564,7 @@ FunctionDef {
 
 Method : FunctionDef {
 	var <ownerClass, <name, <primitiveName;
-	var <filenameSymbol, <charPos;
+	var <charPos;
 
 	openCodeFile {
 		this.filenameSymbol.asString.openDocument(this.charPos, -1);
@@ -662,7 +663,6 @@ Interpreter {
 	interpretPrintCmdLine {
 		var res, func, code = cmdLine, doc, ideClass = \ScIDE.asClass;
 		preProcessor !? { cmdLine = preProcessor.value(cmdLine, this) };
-		func = this.compile(cmdLine);
 		if (ideClass.notNil) {
 			thisProcess.nowExecutingPath = ideClass.currentPath
 		} {
@@ -670,6 +670,7 @@ Interpreter {
 				thisProcess.nowExecutingPath = doc.tryPerform(\path);
 			}
 		};
+		func = this.compile(cmdLine);
 		res = func.value;
 		thisProcess.nowExecutingPath = nil;
 		codeDump.value(code, res, func, this);
--- a/lang/LangSource/PyrObject.cpp
+++ b/lang/LangSource/PyrObject.cpp
@@ -1636,12 +1636,12 @@ void initClasses() {
     addIntrinsicVar(class_fundef, "argNames", &o_nil);
     addIntrinsicVar(class_fundef, "varNames", &o_nil);
     addIntrinsicVar(class_fundef, "sourceCode", &o_nil);
+    addIntrinsicVar(class_fundef, "filenameSymbol", &o_nil);
 
     class_method = makeIntrinsicClass(s_method, s_fundef, 5, 0);
     addIntrinsicVar(class_method, "ownerClass", &o_nil);
     addIntrinsicVar(class_method, "name", &o_nil);
     addIntrinsicVar(class_method, "primitiveName", &o_nil);
-    addIntrinsicVar(class_method, "filenameSymbol", &o_nil);
     addIntrinsicVar(class_method, "charPos", &o_zero);
     // addIntrinsicVar(class_method, "byteMeter", &o_zero);
     // addIntrinsicVar(class_method, "callMeter", &o_zero);
@@ -2478,6 +2478,8 @@ PyrBlock* newPyrBlock(int flags) {
         block = (PyrBlock*)gMainVMGlobals->gc->New(numbytes, flags, obj_notindexed, false);
     block->classptr = class_fundef;
     block->size = numSlots;
+    SetNil(&block->filenameSym);
+
 
     // clear out raw area
     methraw = METHRAW(block);
@@ -2492,6 +2494,22 @@ PyrBlock* newPyrBlock(int flags) {
     methraw->popSize = 0;
 
     nilSlots(&block->rawData1, numSlots);
+
+    if (gCompilingVMGlobals && &gCompilingVMGlobals->process) {
+        PyrSlot* path = &gCompilingVMGlobals->process->nowExecutingPath;
+        if (path && IsObj(path)) {
+            const auto path_length = slotRawObject(path)->size;
+
+            static constexpr auto SZ = 126;
+            char small_string_buf[126];
+            char* string_copy = path_length > 126 ? (char*)malloc(path_length + 1) : small_string_buf;
+            memcpy(string_copy, slotRawString(path)->s, path_length);
+            string_copy[path_length] = 0; // symbols must be null terminated.
+            SetSymbol(&block->filenameSym, getsym(string_copy));
+            if (path_length > 126)
+                delete string_copy;
+        }
+    }
     return block;
 }
 
--- a/lang/LangSource/PyrKernel.h
+++ b/lang/LangSource/PyrKernel.h
@@ -154,13 +154,13 @@ struct PyrBlock : public PyrObjectHdr {
     PyrSlot argNames; // ***arguments to block
     PyrSlot varNames; // ***variables in block
     PyrSlot sourceCode; // source code if it is a closed function.
+    PyrSlot filenameSym;
 };
 
 struct PyrMethod : public PyrBlock {
     PyrSlot ownerclass;
     PyrSlot name;
     PyrSlot primitiveName;
-    PyrSlot filenameSym;
     PyrSlot charPos;
     // PyrSlot byteMeter;
     // PyrSlot callMeter;


JordanHendersonMusic:
Regarding @telephon's example of failing code posted in #6675.
```supercollider
(
"directly called, path is: %".format(thisFunctionDef.filenameSymbol).postln;
Routine { 1.wait; "scheduled in routine, path is: %".format(thisFunctionDef.filenameSymbol).postln }.play(AppClock);
AppClock.sched(0.5, { "scheduled, path is: %".format(thisFunctionDef.filenameSymbol).postln });
)

directly called, path is: /home/jordan/tmp/b.scd
-> AppClock
scheduled, path is: /home/jordan/tmp/b.scd
scheduled in routine, path is: /home/jordan/tmp/b.scd

```

lang/LangSource/PyrObject.cpp
```diff
@@ -2492,6 +2493,22 @@ PyrBlock* newPyrBlock(int flags) {
     methraw->popSize = 0;
 
     nilSlots(&block->rawData1, numSlots);
+
+    if (gCompilingVMGlobals && &gCompilingVMGlobals->process) {
+        PyrSlot* path = &gCompilingVMGlobals->process->nowExecutingPath;
+        if (path && IsObj(path)) {
+            const auto path_length = slotRawObject(path)->size;
+
+            static constexpr auto SZ = 126;
```
telephon, line 2502: did you want to use `SZ` in the expressions below?

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -563,7 +564,7 @@ FunctionDef {
 
 Method : FunctionDef {
 	var <ownerClass, <name, <primitiveName;
-	var <filenameSymbol, <charPos;
+	var <charPos;
```
telephon, line 567: Seeing `charPos`, I wonder if this value is available also in the `gCompilingVMGlobals`? Then `FunctionDef` could also know the position in the file – this would make a nice debugging feature.

JordanHendersonMusic: Good idea. I'm gonna rejig the code quick to make this more elegant.

JordanHendersonMusic:
> Seeing charPos, I wonder if this value is available also in the gCompilingVMGlobals? Then FunctionDef could also know the position in the file – this would make a nice debugging feature.


So I have a version of this working — I also added lineNumber —, real quick to do, but it doesn't behave quite how you'd expect. 

TLDR: it returns the last line number or character count of the scope.




### line number

```
1 | thisFunctionDef.lineNumber
```
Above returns 1.

```
1 | (
2 |
3 | thisFunctionDef.lineNumber.debug(\top);
4 |
5 |
6 |
7 |
8 | { 
9 |	
10|	
11|	
12|	
13| 	thisFunctionDef.lineNumber
14| }.().debug(\bot)
15| )
```
Prints 
```
top: 15
bot: 14
```

### char pos
```
thisFunctionDef.charPos  
```
Returns 24

```
(


thisFunctionDef.charPos;
)
```
Returns 31

```
(

    thisFunctionDef.charPos.debug(\top);




{




	thisFunctionDef.charPos
}.().debug(\bot)
)
```
Prints
```
top: 99
bot: 80
```

JordanHendersonMusic:
This raises another question....

String.loadRelative (and friends) do odd things when run from a routine.


I would suggest we deprecate this, I know that is not popular, but most (if not all) of the time you want to be relative to the file it was defined in, not the executing file.

Here is a little implementation that builds off this PR and solves the issue.


```supercollider
String {
   loadRelativeTo { |filename, warn = true, action|
		^(this.dirname ++ thisProcess.platform.pathSeparator ++ filename).loadPaths(warn, action)
   }
}
FunctionDef {
   filename { ^filenameSymbol.asString }
}

thisFunctionDef.filename.loadRelativeTo("a.scd")
```

If we decided we wanted new keywords in the language, which would make this much neater, it could just look like...

```supercollider
thisFilename.loadRealtiveTo("a.scd")
```

telephon:
Yes, I also think that the default should be the file a function or routine was defined in. But we could keep this in a separate pull request? The current one builds the basic capability. 

There could be another merge request for better debugging, but that is for later.

**I've tested it and it works well.**

One case where it doesn't do what is expected: when you evaluate the source code for a function in an existing file, the `def.lineNumber` is not the line in the file, but in the current selected part. I am not sure if this is avoidable, but it seems like a possible problem.

JordanHendersonMusic:
> One case where it doesn't do what is expected:

That's exactly what I expected to happen - probably because I've been digging in that part of the code -, but yes it isn't all that useful.

> Yes, I also think that the default

Sorry, the default of what?

telephon:
> > Yes, I also think that the default should be the file a function or routine was defined in.
> 
> Sorry, the default of what?

The default for what we mean when we want to resolve a path (resolveRelative).

JordanHendersonMusic:
Gotcha. The slight issue is that you can't write `thisFunctiomDef.filenameSymbol` inside `String.load`, as it's defined in String.sc. 

I haven't been able to come up with a better alternative than `thisFunctionDef.loadRelativeTo(...)`.

telephon:
Well, I think that `resolveRelative` should always resolve to the current `filenameSymbol` instead of to `nowExecutingPath`. 

We have two possibilities:

1. either we call `resolveRelative` in some text file directly (not enclosed in a function explicitly), then the function is the one instantiated from `interpret` (see there), and we want the text file's path where the function was built (which is the one which we are in), so `filenameSymbol`.
2. or we call `resolveRelative` from a function. Then it should resolve to the file path in which this function was instantiated, so `filenameSymbol`. 

To test (1.) run the following in a saved file:
```supercollider
thisFunctionDef.filenameSymbol.postln;
```

Only if we want to build a library function that, wherever it is called, resolves relative to that file, we'd need `nowExecutingPath`. But I would tend to argue that this is not a good design :-/ and would recommend passing the path to the function instead. Or, as a workaround, this one can actually really use `nowExecutingPath`.

JordanHendersonMusic:
> that resolveRelative should always resolve to the current filenameSymbol instead of to nowExecutingPath.

Agreed!

> then the function is the one instantiated from interpret (see there),

I'm not quite following what you're suggesting.. but I think I get the gist.

So you're suggesting the use of `getBackTrace`? I tried this but couldn't get it to work. I'll have another look though. Both one and two can use this mechanism if it works.

>Only if we want to build a library function that, wherever it is called, resolves relative to that file,

To which file? The library file, or the user's file? Having helper methods to resolve paths near the users file seems useful.

Isn't this what `loadRelative` does anyway? 

This is why I suggested `thisFunctionDef.loadRelative(nameOfFile)`, and similarity, `thisFunctionDef.resolveRelative(fileName)`. As within the functiondef object is the only place (bare the actual file) you have access to the parser's currently active file.

If we added keywords to the languages, you could even write it like this

```
currentFile.loadRelative("myfile")
```

Otherwise you'd have to flip the order and write, `"file.scd".resolveTo(thisFunctionDef.filename)`, as you need to write the `thisFunctionDef` bit in the file you wish to resolve relative to.

telephon:
I think we don't have to change the interface. When you call `"my/path/to.scd".resolveRelative`, you are inside a function def, and that can be used in the implementation of `resolveRelative`. If that file has code that in turn resolves a path, it will do so relative that file's path, and so on.

> To which file? The library file, or the user's file?

I meant a user's file, but it should not make a difference.

> Isn't this what `loadRelative` does anyway?

Currently, it resolves to `nowExecutngPath`, which gives us a problem if this happens in a function that is scheduled (like a `defer`).

telephon:
> So you're suggesting the use of `getBackTrace`? I tried this but couldn't get it to work. I'll have another look though. Both one and two can use this mechanism if it works.

I haven't ever used this for resolving paths … will try tomorrow.

JordanHendersonMusic:
> "my/path/to.scd".resolveRelative, you are inside a function def

Yes, but that functiondef (or rather `Method`) is `String.resolveRelative` located in the class library.  You'd end up with `/path/to/classlib/Collections/my/path/to.scd`, not `/path/to/current/file/my/path/to.scd`.


I don't think it will be possible to fix this. Even with `getBackTrace` as it won't work when you nest these calls across multiple files in the same call stack. Also walking the call stack should really be avoided as it makes the behaviour vary depending on the call sight.

JordanHendersonMusic:
> resolveRelative {
		var path, caller;
		caller = thisMethod.getBackTrace.caller.functionDef;
		if(caller

I'm not a fan of this implementation. It will fall back to the executing path if used in a quark or class lib. That's going to be (remain) confusing. Less so in class lib as we probably won't be storing resources there, but this seems reasonable in a quark where you might store irs (for example).

Also if you were to wrap this in another method it will behave different as the call sight has changed.

This is why I like moving these methods (load and whatnot) into functiondef. It isn't relative to the current open file in the IDE, nor the current function in the scheduler, nor does it depend on call sight, it is relative to **this** file, the file where `thisFunctionDef` is literally written. I don't think any of the other options are quite this clear.


> Currently, when I put the following line in a file: 

Weird I tested this. Need to add it to the test suit. Will look at later.

JordanHendersonMusic:
I can't get it to work properly with nested files
Adding your above tweak to `resolveRelative`.


```
// path: dir/nested/b.scd
"in file nested/b.scd".postln;
"c.scd".resolveRelative;
```

```
// path: dir/a.scd
"nested/b.scd".loadRelative 
```

This final line should print `"..../dir/nested/c.scd"`, but I'm getting `"/home/jordan/.repo/SuperCollider/SCClassLibrary/Common/Core/d.scd"`.

JordanHendersonMusic:
> But there is one issue still:
Currently, when I put the following line in a file:


You've missed the `%` in the format string :wink:

telephon:
> I can't get it to work properly with nested files Adding your above tweak to `resolveRelative`.
> This final line should print `"..../dir/nested/c.scd"`, but I'm getting `"/home/jordan/.repo/SuperCollider/SCClassLibrary/Common/Core/d.scd"`.

Well, I haven't yet added a tweak to `loadRelative`, you currently need to load with `"path".resolveRelative.load;`

telephon:
Hm strange. Just for precision.

1. save a file named "sc-test.scd" on the desktop with the following as content:

```supercollider
"outside of a function, the full path is: '%'".format("".resolveRelative).postln;
AppClock.sched(0.3, { "Inside a scheduled function, the full path is: '%'".format("".resolveRelative).postln });
```

Over here, I get the correct paths posted.

2. open an empty file and call 
 
```supercollider
"~/Desktop/sc-test.scd".standardizePath.load
```

telephon:
My String.sc has this:

```
resolveRelative {
		var path, caller;
		caller = thisMethod.getBackTrace.caller.functionDef;
		caller.filename.postcs;
		if(caller.isKindOf(FunctionDef) && (caller != Interpreter.findMethod(\interpretPrintCmdLine))) {
			path = caller.filenameSymbol.asString;
		} {
			path = thisProcess.nowExecutingPath;
		};
		if(this[0] == thisProcess.platform.pathSeparator) { ^this };
		if(path.isNil) { Error("can't resolve relative to an unsaved file.\nPath to resolve: \"%\"\n".format(this)).throw };
		^(path.dirname ++ thisProcess.platform.pathSeparator ++ this)
	}
```

telephon:
yes, good counter examples. So it seems that we need two methods of `resolveRelative`, one that resolves relative to the call site and the other one that resolves relative to the definition path …

JordanHendersonMusic:
Okay so let's keep this as a single pr as it is functional and useful now, and I'll put together another pr to update string load/resolve relative and implement another way of doing the same, but that lets you pass around the thing that you are relative to. 

Do you think the documentation is sufficient here? 

Now executing path should probably be updated so there is a clear difference between the path of execution and the path of the definition, but I don't want to step on the toes of the other pr and the work done there.

@prko which would you prefer? That you update your pr with the information here, or that I, borrowing from your pr, update now executing? The order everything gets merged in is also important.

I'm quite happy either way 😊

HelpSource/Classes/FunctionDef.schelp
```diff
@@ -100,6 +102,29 @@ code::
 { |a = 9, b = 10 ... c| a + b }.def.argumentString(asArray: true);
 ::
 
+method::filenameSymbol
+
+Return the file path (as a link::Classes/Symbol::) where the code was defined, if the file has not been saved (or you are using a terminal interface) returns nil.
```
telephon, line 107: –> Return the file path (as a link::Classes/Symbol::) where the function was defined,

lang/LangSource/PyrObject.cpp
```diff
@@ -2492,6 +2494,12 @@ PyrBlock* newPyrBlock(int flags) {
     methraw->popSize = 0;
 
     nilSlots(&block->rawData1, numSlots);
+
+    if (optional_filename)
+        SetSymbol(&block->filenameSym, optional_filename);
```
telephon, line 2499: format?

lang/LangSource/PyrParseNode.cpp
```diff
@@ -112,6 +112,22 @@ void compileTail() {
 }
 
 
+/// Returns the filename currently being parsed.
+/// If code is passed from an unsaved file, or from the command line, return nullptr.
```
telephon, line 116: assuming that we generate lots of functions, do you see any performance impact?

lang/LangSource/PyrSymbol.cpp
```diff
@@ -0,0 +1,23 @@
+#include "PyrSymbol.h"
+#include "PyrObject.h"
+#include <cstring>
+#include "PyrSymbolTable.h"
+
+PyrSymbol* stringToSymbol(PyrString* str) {
```
telephon, line 6: There is a primitive `prStringAsSymbol`. 

```
int prStringAsSymbol(struct VMGlobals* g, int numArgsPushed) {
    PyrSlot* a = g->sp;
    SetSymbol(a, stringToSymbol(slotRawString(a)));
    return errNone;
}
```

Should they be considered together?

telephon (changes requested):
just a few questions and minor things before finishing up.

lang/LangSource/PyrParseNode.cpp
JordanHendersonMusic: A small impact mostly from the string to symbol conversion. 

I doubt this would be measurable. 

The only way around this would be a new keyword that only gets the path when invoked. Or perhaps a fast string to symbols conversion that moves the memory, avoiding an allocation when there is a new symbol.

lang/LangSource/PyrSymbol.cpp
JordanHendersonMusic: No because the primitive isn't defined in a header.

lang/LangSource/PyrObject.cpp
JordanHendersonMusic: What's wrong? Clang-format has no problem.

lang/LangSource/PyrParseNode.cpp
telephon: > Or perhaps a fast string to symbols conversion that moves the memory, avoiding an allocation when there is a new symbol.

Would be good also for other purposes perhaps, but I suppose this is ok here.

lang/LangSource/PyrObject.cpp
telephon: the indentation looks different from the lines below it. The clang-formatter only checks the code that it changed. Not a big issue.

prko:
>@prko 
which would you prefer? That you update your pr with the information here, or that I, borrowing from your pr, update now executing? The order everything gets merged in is also important.

@JordanHendersonMusic 
Thank you for your consideration! Here's what I propose:

1. You could incorporate the most useful parts from my PR into yours, focusing on the practical implementation that makes your PR so effective. 
2. Once you've done that, I'll update my PR by removing those duplicated sections and adding proper references to your documentation instead.

This approach seems quite sensible as:
- It respects the time and effort we've both invested
- It creates clearer, more streamlined documentation for users
- It plays to our respective strengths (your practical approach and my detailed explanations)
- It helps me avoid having to rewrite everything in English, which I'm not entirely confident with

I’m looking forward to collaborating on this, and I appreciate your consideration of my work, even if it might be a bit overly detailed in places.

What do you think about it?

JordanHendersonMusic:
@prko  I think that's a great idea. One just been a bit busy but will try to get to this soon.

lang/LangSource/PyrObject.cpp
JordanHendersonMusic: The formatting looks fine to me.

lang/LangSource/PyrParseNode.cpp
JordanHendersonMusic: I was wrong, this can't be done without a big rewrite of the symbol table which I don't want to do.

dyfer:
The change in the docs is good, but I'd put the warning right under `WARNING:: This should NOT be used to return the path of the current file`...

JordanHendersonMusic:
Did we ever decide if this was going into 3.14?

HelpSource/Classes/FunctionDef.schelp
```diff
@@ -1,10 +1,23 @@
 class::FunctionDef
-summary:: FunctionDefs contain code which can be executed from a Function.
+summary:: FunctionDefs contain information about functions.
 categories::Core>Kernel
 related::Classes/Function
 
 description::
 
+code::FunctionDef:: is primarily used by the interpreter and most users shouldn't use it,
```
telephon: "and most users shouldn't use it," 
Better leave this out, because anyone may want to use it, e.g. by looking up the byteCodes. The helpfiles should be written without having in mind anyone particular / anyone with a particular skill. So maybe just say what it is useful for?

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -415,11 +415,16 @@ Process {
...
 	var <sourceCode;
+	var <filenameSymbol;
+	var <charPos;
+	var <lineNumber;
 
 	// a FunctionDef is defined by a code within curly braces {}
 	// When you use a FunctionDef in your code it gets pushed on the stack
 	// as an instance of Function
 
+	filename { ^filenameSymbol !? (_.asString) }
```
telephon: should better be:

filename { ^filenameSymbol !? { filenameSymbol.asString } }

telephon (changes requested):
Just minor things.

This will be very useful, e.g. we can add more information to errors, making them easier to find the source code.

JordanHendersonMusic:
@telephon I think I've addressed all your concerns with this.

JordanHendersonMusic (commented):
@telephon do you think this is good?

JordanHendersonMusic:
Note to self: rebase!

JordanHendersonMusic:
Note to self: Look again at the current character index, right now it refers to the final '}', it should refer to the first. Have a look again at this.

telephon:
> @telephon do you think this is good?

yes :) (sorry I had forgotten about this)

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -458,11 +458,16 @@ Process {
 FunctionDef {
 	var raw1, raw2, <code, <selectors, <constants, <prototypeFrame, <context, <argNames, <varNames;
 	var <sourceCode;
+	var <filenameSymbol;
+	var <charPos;
```
scztt, line 462: charPos was an absolute offset - does this change the meaning to a line offset? Or keep the old meaning (in which case, the lineNumber isn't useful because we dont know where in the line). I think absolute is preferable anyway, as we can jump directly to a code location (useful for code lens type things in the IDE and VSCode) without counting lines in the file.

If we're thinking about a more forward-looking interface, I think the idea would be `startChar` and `endChar`, since just getting the end char is pretty useless. You can get the startChar by taking the end of the previous neighbor AST node - iirc this will still include whitespace for e.g. methods, but otherwise I remember this being accurate. Then `charPos` would forward to `endChar` and be deprecated. But also no need to blow the scope of this PR if its a can of worms!

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -705,14 +709,14 @@ Interpreter {
...
 		if (ideClass.notNil) {
 			thisProcess.nowExecutingPath = ideClass.currentPath
 		} {
 			if(\Document.asClass.notNil and: {(doc = Document.current).tryPerform(\dataptr).notNil}) {
 				thisProcess.nowExecutingPath = doc.tryPerform(\path);
 			}
 		};
+		func = this.compile(cmdLine);
 		res = func.value;
 		thisProcess.nowExecutingPath = nil;
```
scztt, line 721: IIRC there are bugs when func.value throws, leaving a dangling `nowExecutingPath` - this should probably be wrapped in `protect` (if this is the case...), as the other `nowExecuting` logic is. Unrelated to the PR, but probably a good time to fix it.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -3783,7 +3800,7 @@ void PyrBlockNode::compile(PyrSlot* slotResult) {
     // create a new block object
 
     flags = compilingCmdLine ? obj_immutable : obj_permanent | obj_immutable;
-    block = newPyrBlock(flags);
+    block = newPyrBlock(flags, getActivelyParsedFile(), mLineno, linestarts[mLineno] + mCharno);
```
scztt, line 3803: I THINK a better implementation might be to pass the active file in as an argument to `PyrBlockNode::compile`, and in turn to Interpreter:compile. Rationale is, we probably need to be able to attach an arbitrary path to a compiled function in cases where we can't programmatically associate it with e.g. a `Document` (e.g. i load compilable strings from a JSON file) - passing in as an argument is more direct and reliable than global scope solutions. This moves the logic about how to determine the current path to the language side, which is more flexible and keep the C++ simple.

JordanHendersonMusic:
Thanks @scztt I'll have a look at this when I get the chance!

SCClassLibrary/Common/Core/Kernel.sc
jamshark70: I tried to reproduce this but failed to do so:

```
thisProcess.nowExecutingPath.debug("unscheduled");
unscheduled: /...redacted.../misc-1.scd

SystemClock.sched(2.0, { thisProcess.nowExecutingPath.debug("scheduled") });
scheduled: nil

// first run this
SystemClock.sched(2.0, { thisProcess.nowExecutingPath.debug("scheduled") });

// and before it fires, run this
1 + nil

-->
scheduled: nil
```

Do you have a concrete scenario where a dangling nowExecutingPath is not cleared before the next thing happens in the interpreter?

SCClassLibrary/Common/Core/Kernel.sc
JordanHendersonMusic: Just coming back to this comment a few months later.

> charPos was an absolute offset 

charPos in the lexer was the offset into the current line. The offset into the source code is called textPos. (very confusing)!

You can't go to the previous ast node from a given node, its is a single linked list.

... but there is a bigger problem, the entire way the parser/lexer communicate relied on global state, both for line position and semantic value, calling yylex changed this state, however the parser was free to call this as a part of its look ahead. This meant you some times it would report the incorrect location data! Also, because we do bottom up parsing, it was impossible to get the location of the first rule matched if it was a literal (in the case of functions, this is usually the open curly). Luckily I couldn't find any places where the semantic value was returned to the parser incorrectly else that would have been a major bug!

I've refactor the entire thing in #7366 making use of the bison approved `yyloc` and `yylval`. This makes getting the source code location trivial.  

I'm probably going to absorb this PR into that one, its already huge...

jamshark70:
[Discussed](https://scsynth.org/t/function-throws-error-when-ran-from-oscdef-but-runs-normally-when-not/13124/10?u=jamshark70) with Jordan. This PR will be obsolete after #7394 goes in, so I'm closing this.

PR #6405, opened 2024-07-29
ClassLib: fix and refactor SynthDef compiler

JordanHendersonMusic:
## Purpose and Motivation

This is a large refactor/rewrite of the SynthDef compiler because it is broken and complicated to use. See https://scsynth.org/t/proposal-refactor-synthdef-ugen-optimizations/9834/86 for a wider discussion on the issue.

This is currently a draft.

Feedback about implementation is welcome here, for a more wider conversation please use scsynth.

## Description of proposed changes

Weak edges will be placed between server side resource access to ensure correct ordering.

Optimisations (replacing ugens or chains thereof) can be done by dealing with just the connections, no need to delete anything from inside a UGen or alter the whole graph state.

The graph will topologically sort itself, optimising for long chains of strong connections to reduce wirebuf usage.

Dead code elimination happens by default because the server resources are tracked and something meaningful only happens when some resource (bus, buffer, etc...) is touched.

## So far complete 
* Initial UGenResourceManagers creating additional edges between resources (there are still some UGens that need attention).

## Demo

Right now, when ever a weak edge is made, it logs it to the post window. 
```
SynthDef(\t, {
	var sig = 5.collect({|i| In.ar(i) }).sum;
	BufWr.ar(sig, 0, LFNoise2.ar(2).range(0, 1), 1);
	3.collect({|i| Out.ar(i, sig) });
	In.ar(2)
})

post window:
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an In, an Out]
[createWeakConnectionTo, an Out, an In]
[createWeakConnectionTo, an Out, an In]
[createWeakConnectionTo, an Out, an In]
```
That is each Input (x5) connecting to all the Outputs (x3), then all the Outputs (x3) connecting to the final Input.

## Types of changes
 
- Documentation (TODO)
- Bug fix
- New feature
[...]

PR #6681, opened 2025-03-24
Plugin code refactoring and small fixes

Spacechild1:
## Purpose and Motivation

This PR contains some small bug fixes and code refactoring of the Server plugins, as a preparation for my upcoming reblocking/resampling PR.

Most importantly, 23fce872c2b1be1a9ad83a9fbdaca29680c5734f consequently replaces all raw accesses of `mWorld->mFullRate.*` members with the appropriate macros in `plugin_interface/SC_Unit.h`. The intention is that in my later reblock/resample PR these macros will be changed to point to `Rate*` members *in the Graph* instead of the global `Rate*` members. Theses changes have been done semi-automatically and should not change the actual code, but you never know...

The remaining commits mostly concern the IO UGens. I have tested all of them and haven't found an issue so far, but I would really appreciate if other people can test as well!

**Please test! Also try with Server block sizes lower than 64 and 16 to run the non-SIMD version of certain process functions.**

(For anyone curious, I already have a working and complete implementation of per-Synth reblocking and upsampling: https://github.com/Spacechild1/supercollider/tree/reblock. However, I only want to make an official PR once this PR has been sufficiently tested.)

## Types of changes

- Bug fix
- Code cleanup

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/server/plugins/IOUGens.cpp
+++ b/server/plugins/IOUGens.cpp
@@ -560,7 +560,7 @@ void In_next_a(IOUnit* unit, int inNumSamples) {
         if (guard.isValid && (touched[i] == bufCounter))
             Copy(inNumSamples, out, in);
         else
-            Fill(inNumSamples, out, 0.f);
+            Clear(inNumSamples, out);
     }
 }
 
@@ -715,12 +715,12 @@ void InFeedback_next_a(InFeedback* unit, int inNumSamples) {
             unit->m_busUsedInPrevCycle = true;
         } else if (guard.isValid && diff == 1) {
             if (unit->m_busUsedInPrevCycle) {
-                Fill(inNumSamples, out, 0.f);
+                Clear(inNumSamples, out);
                 unit->m_busUsedInPrevCycle = false;
             } else
                 Copy(inNumSamples, out, in);
         } else {
-            Fill(inNumSamples, out, 0.f);
+            Clear(inNumSamples, out);
             unit->m_busUsedInPrevCycle = false;
         }
     }
--- a/server/plugins/IOUGens.cpp
+++ b/server/plugins/IOUGens.cpp
@@ -410,7 +410,7 @@ void LagControl_Ctor(LagControl* unit) {
     for (int i = 0; i < numChannels; ++i, mapin++) {
         unit->m_y1[i] = **mapin;
         float lag = ZIN0(i);
-        unit->m_b1[i] = lag == 0.f ? 0.f : (float)exp(log001 / (lag * unit->mRate->mSampleRate));
+        unit->m_b1[i] = lag == 0.f ? 0.f : (float)exp(log001 / (lag * SAMPLERATE));
     }
 
     if (unit->mNumOutputs == 1) {
@@ -680,7 +680,7 @@ void LagIn_Ctor(LagIn* unit) {
     unit->m_fbusChannel = -1.;
 
     float lag = ZIN0(1);
-    unit->m_b1 = lag == 0.f ? 0.f : (float)exp(log001 / (lag * unit->mRate->mSampleRate));
+    unit->m_b1 = lag == 0.f ? 0.f : (float)exp(log001 / (lag * SAMPLERATE));
 
     SETCALC(LagIn_next_k);
     unit->m_bus = world->mControlBus;
--- a/server/plugins/IOUGens.cpp
+++ b/server/plugins/IOUGens.cpp
@@ -42,46 +42,59 @@ using nova::slope_argument;
 
 #include <boost/align/is_aligned.hpp>
 
-
 static InterfaceTable* ft;
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
 
+struct Control : Unit {};
+
+struct AudioControl : Unit {
+    float* prevVal;
+    float* m_prevBus;
+    bool m_busUsedInPrevCycle;
+};
+
+struct TrigControl : Unit {};
+
+struct LagControl : Unit {
+    float* m_b1;
+    float* m_y1;
+};
+
 struct IOUnit : public Unit {
     int32* m_busTouched;
-    float m_fbusChannel;
     float* m_bus;
+    float m_fbusChannel;
 };
+
+struct InTrig : IOUnit {};
+
+struct In : IOUnit {};
+
+struct InFeedback : public In {
+    bool m_busUsedInPrevCycle;
+};
+
+const int kMaxLags = 16;
+
+struct LagIn : public IOUnit {
+    float m_b1;
+    float m_y1[kMaxLags];
+};
+
+struct Out : IOUnit {};
 
 struct XOut : public IOUnit {
     float m_xfade;
 };
 
+struct ReplaceOut : IOUnit {};
+
 struct OffsetOut : public IOUnit {
     float* m_saved;
+    // QUESTION: why don't we just zero the save buffer in
+    // OffsetOut_Ctor instead of checking 'm_empty'?
     bool m_empty;
-};
-
-struct InFeedback : public IOUnit {
-    bool m_busUsedInPrevCycle;
-};
-
-struct AudioControl : public InFeedback {
-    float* prevVal; // this will have to be a pointer later!
-    float* m_prevBus;
-};
-
-const int kMaxLags = 16;
-
-struct LagControl : public IOUnit {
-    float* m_b1;
-    float* m_y1;
-};
-
-
-struct LagIn : public IOUnit {
-    float m_b1;
-    float m_y1[kMaxLags];
 };
 
 struct LocalIn : public Unit {
@@ -90,30 +103,46 @@ struct LocalIn : public Unit {
     float* m_realData;
 };
 
+struct LocalOut : Unit {
+    float* m_bus;
+    int32* m_busTouched;
+};
+
+struct SharedOut : Unit {
+    float* m_bus;
+    float m_fbusChannel;
+};
+
+struct SharedIn : Unit {
+    float* m_bus;
+    float m_fbusChannel;
+};
+
 extern "C" {
-void Control_Ctor(Unit* inUnit);
-void Control_next_k(Unit* unit, int inNumSamples);
-void Control_next_1(Unit* unit, int inNumSamples);
+void Control_Ctor(Control* inUnit);
+void Control_next_k(Control* unit, int inNumSamples);
+void Control_next_1(Control* unit, int inNumSamples);
 
 void AudioControl_Ctor(AudioControl* inUnit);
-//	void AudioControl_Dtor(AudioControl *inUnit);
+void AudioControl_Dtor(AudioControl* inUnit);
 void AudioControl_next_k(AudioControl* unit, int inNumSamples);
 void AudioControl_next_1(AudioControl* unit, int inNumSamples);
 
-void TrigControl_Ctor(Unit* inUnit);
-void TrigControl_next_k(Unit* unit, int inNumSamples);
-void TrigControl_next_1(Unit* unit, int inNumSamples);
+void TrigControl_Ctor(TrigControl* inUnit);
+void TrigControl_next_k(TrigControl* unit, int inNumSamples);
+void TrigControl_next_1(TrigControl* unit, int inNumSamples);
 
 void LagControl_Ctor(LagControl* inUnit);
+void LagControl_Dtor(LagControl* inUnit);
 void LagControl_next_k(LagControl* unit, int inNumSamples);
 void LagControl_next_1(LagControl* unit, int inNumSamples);
 
-void InTrig_Ctor(IOUnit* unit);
-void InTrig_next_k(IOUnit* unit, int inNumSamples);
+void InTrig_Ctor(InTrig* unit);
+void InTrig_next_k(InTrig* unit, int inNumSamples);
 
-void In_Ctor(IOUnit* unit);
-void In_next_a(IOUnit* unit, int inNumSamples);
-void In_next_k(IOUnit* unit, int inNumSamples);
+void In_Ctor(In* unit);
+void In_next_a(In* unit, int inNumSamples);
+void In_next_k(In* unit, int inNumSamples);
 
 void LagIn_Ctor(LagIn* unit);
 void LagIn_next_0(LagIn* unit, int inNumSamples);
@@ -122,30 +151,37 @@ void LagIn_next_k(LagIn* unit, int inNumSamples);
 void InFeedback_Ctor(InFeedback* unit);
 void InFeedback_next_a(InFeedback* unit, int inNumSamples);
 
-void LocalIn_Ctor(LocalIn* unit);
-void LocalIn_Dtor(LocalIn* unit);
-void LocalIn_next_a(LocalIn* unit, int inNumSamples);
-void LocalIn_next_k(LocalIn* unit, int inNumSamples);
-
-void Out_Ctor(IOUnit* unit);
-void Out_next_a(IOUnit* unit, int inNumSamples);
-void Out_next_k(IOUnit* unit, int inNumSamples);
+void Out_Ctor(Out* unit);
+void Out_next_a(Out* unit, int inNumSamples);
+void Out_next_k(Out* unit, int inNumSamples);
 
 void XOut_Ctor(XOut* unit);
 void XOut_next_a(XOut* unit, int inNumSamples);
 void XOut_next_k(XOut* unit, int inNumSamples);
 
-void ReplaceOut_Ctor(IOUnit* unit);
-void ReplaceOut_next_a(IOUnit* unit, int inNumSamples);
-void ReplaceOut_next_k(IOUnit* unit, int inNumSamples);
+void ReplaceOut_Ctor(ReplaceOut* unit);
+void ReplaceOut_next_a(ReplaceOut* unit, int inNumSamples);
+void ReplaceOut_next_k(ReplaceOut* unit, int inNumSamples);
 
 void OffsetOut_Ctor(OffsetOut* unit);
 void OffsetOut_Dtor(OffsetOut* unit);
 void OffsetOut_next_a(OffsetOut* unit, int inNumSamples);
 
-void LocalOut_Ctor(IOUnit* unit);
-void LocalOut_next_a(IOUnit* unit, int inNumSamples);
-void LocalOut_next_k(IOUnit* unit, int inNumSamples);
+void LocalIn_Ctor(LocalIn* unit);
+void LocalIn_next_a(LocalIn* unit, int inNumSamples);
+void LocalIn_next_k(LocalIn* unit, int inNumSamples);
+
+void LocalOut_Ctor(LocalOut* unit);
+void LocalOut_next_a(LocalOut* unit, int inNumSamples);
+void LocalOut_next_k(LocalOut* unit, int inNumSamples);
+
+void SharedIn_Ctor(SharedIn* unit);
+void SharedIn_next_a(SharedIn* unit, int inNumSamples);
+void SharedIn_next_k(SharedIn* unit, int inNumSamples);
+
+void SharedOut_Ctor(SharedOut* unit);
+void SharedOut_next_a(SharedOut* unit, int inNumSamples);
+void SharedOut_next_k(SharedOut* unit, int inNumSamples);
 }
 
 //////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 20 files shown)


capital-G (commented):
Only have skimmed through this and needs further testing/reviewing, but some things I already noticed/have questions about:

One thing I'd like to know - why introduce an indirection through a macro which simply maps to `unit->mWorld->mFullRate.mSampleDur`? The latter, although more verbose, seems more clearer to me what happens and reduces the vocabulary I have to keep in mind while coding/reviewing. Or maybe I am just an anti-macro guy :D Has this advantages for your future PR?

Also: Can you say if you simply moved `AudioBusGuard` to another location (which introduces a somehow bigger diff) or have you modified it? Currently it is difficult to say based on the diff.
Also - is `AudioControl_next_channel` a simple refactoring from the `next` method or has something changed here?

Maybe you could add some documentation/comments about it, this would make reviewing and future maintenance easier and would be really valuable b/c it seems you are currently in the topic and if some person at some point in time needs to re-visit this it would be great to build on top of your knowledge :)

Have you tested the I/O changes also with supernova?

Spacechild1:
> One thing I'd like to know - why introduce an indirection through a macro which simply maps to unit->mWorld->mFullRate.mSampleDur? The latter, although more verbose, seems more clearer to me what happens and reduces the vocabulary I have to keep in mind while coding/reviewing. Or maybe I am just an anti-macro guy :D Has this advantages for your future PR?

Ah yes, that is not entirely obvious. I just added the missing crucial info the PR description:

> The intention is that in my later reblock/resample PR these macros will be changed to point to Rate* members in the Graph instead of the global Rate* members.

The idea is that with these macros plugins only need to be recompiled against a new SC version and they would just work correctly in a reblocked/resampled Synth. In theory, the *built-in* plugins would not need these macros because they always ship with the according SC version. However, I believe that the built-in plugins should be written in the same style as external plugins to set a good example. (That would also include to get rid of all these anachronoustic LOOP macros, btw, but I disgress.) 

I also do think that `FULLRATE` is more concise than `unit->mWorld->mFullRate.mSampleDur`. Note that these macros are only necessary for old C-style macros. The C++ wrapper in `SC_PlugIn.hpp` exposes these variables as proper methods (e.g. `SCUnit::fullRate()`)

Spacechild1:
> Also: Can you say if you simply moved AudioBusGuard to another location (which introduces a somehow bigger diff) or have you modified it? Currently it is difficult to say based on the diff.
Also - is AudioControl_next_channel a simple refactoring from the next method or has something changed here?

The idea of splitting a PR into multiple commits is exactly that people can understand individual changes and not have to look at a giant diff. So don't just read the whole diff, read through the individual commits!

Your questions above should we answered by commits 083317717836fb5a3630e134279a445eb2026182 and ed82664ede379edb562fffeabfb27bbb057ec53d

capital-G:
> Your questions above should we answered by commits [0833177](https://github.com/supercollider/supercollider/commit/083317717836fb5a3630e134279a445eb2026182) and [ed82664](https://github.com/supercollider/supercollider/commit/ed82664ede379edb562fffeabfb27bbb057ec53d)

Ah thanks - I just found out that on the `Files changed/Commits` tab you can step through commits - this seems really helpful as the comments of the commits indeed answer most questions :) See the next button on the right

image

> What would you want me to comment/document exactly? The implementation of the individual IO UGens?

If you want to do that I'd appreciate it, but I was thinking about a small comment about e.g.
 
* `AudioBusGuard` (what does it guard? does it act as some kind of a mutex on a bus?)
* `AudioControl_next_channel` (e.g. it iterates over a list of channels and does what? Plumps values?)

We don't have a documentation about this functionality yet, but I think it would be a great opportunity to add some comments to this essential functionality. It doesn't need to be a big text, but 1-2 lines would allow to understand things easier in scysnth/plugin land w/o looking into the implementation.

> Yes! But again, I would appreciate if other people would test as well!

Will test it for sure, also with different block sizes :)

server/plugins/IOUGens.cpp
```diff
@@ -177,143 +287,111 @@ void Control_Ctor(Unit* unit) {
...
+
+    size_t memSize = numChannels * (sizeof(float) + sizeof(bool));
+    char* mem = (char*)RTAlloc(world, memSize);
+    ClearUnitIfMemFailed(mem);
+
+    unit->m_prevVal = (float*)mem; // see AudioControl_Dtor()!
+    mem += numChannels * sizeof(float);
+    std::fill_n(unit->m_prevVal, numChannels, 0.f);
+
+    unit->m_busUsedInPrevCycle = (bool*)mem;
```
capital-G, line 392: are you sure this value gets freed in the destructor?

IIUC: You allocate mem via RTAlloct, use the first bits for `m_prevVal` and the 2nd section for `m_busUsedInPrevCycle`.

But then in the destructor you only seem to free `m_prevVal` which would not include the appended `numChannels * sizeof(bool)` section?
Or maybe I am not seeing something obvious here.

Spacechild1:
>  Ah thanks - I just found out that on the Files changed/Commits tab you can step through commits - this seems really helpful as the comments of the commits indeed answer most questions :)

Yes, that's one of the purposes of commit messages 🙂 

> if you want to do that I'd appreciate it, but I was thinking about a small comment about e.g.
> AudioBusGuard (what does it guard? does it act as some kind of a mutex on a bus?)
> AudioControl_next_channel (e.g. it iterates over a list of channels and does what? Plumps values?)

Ah, I see. Yes, I can add some comments to these helper functions/classes.

> AudioControl_next_channel

It just processes a single audio control channel and it is called by `AudioControl_next_k` resp. `AudioControl_next_1` to avoid code duplication. I will add a comment, though.

> Will test it for sure, also with different block sizes :)

👍

server/plugins/IOUGens.cpp
Spacechild1: We allocate a single chunk, so we must only free that chunk! `m_busUsedInPrevCycle` points into the memory, but it does not *own* it. `m_prevVal` actual points to the beginning of the memory chunk, that's why we pass it to `RT_Free` in the destructor.

`AudioControl_Dtor` actually has a comment (`// see AudioControl_Ctor()!`) but it puts too much exercise on the reader 😆 Will add a better comment!

Spacechild1: Side note: this is a common pattern in the Server code to reduce the number of (RT) memory allocations. It's essentially a crude linear allocator. In fact, Supernova has a dedicated helper class (`linear_allocator`) for this purpose.

capital-G: Ah sorry ... still try to wrap my head around C++ mem management. So`m_prevVal` takes the ownership of the memory address with its (implied) size and we simply cast a section of it to a float array pointer and cast the latter section to the bool array.

The allocator keeps track of memory address and its size and once we call the freeing of `m_prevVal` (whose address is equal to the initiated `mem`) it will free `m_prevVal` with the init range of `mem`, so including `m_prevVal` and `m_busUsedInPrevCycle`.

Makes sense, b/c you don't specify the range you want to free starting from a given pointer address...

Thanks for explaining this :)

Spacechild1: Side note: it's also important to first "allocate" the `float` array and only then the `bool` array, otherwise the `float` array might be misaligned.

Spacechild1:
@capital-G Suddenly there are lots of test failures, although the last commit only added some comments. I had similar test failures in another PR (https://github.com/supercollider/supercollider/pull/6153#issuecomment-2746563139). Seems like our test suite is quite fragile... Do you have any ideas what's going on? Has this happened in the past?

capital-G:
> Seems like our test suite is quite fragile... Do you have any ideas what's going on? Has this happened in the past?

Yeah, basically we observe some unreliable results in CI - it seems this happens when CPU usage on the host is high or we also experienced some dropping of UDP packages. This is an ongoing investigation.

Current hack is to re-run the failed job until they turn green. :/

Spacechild1:
> or we also experienced some dropping of UDP packages.

Don't the tests use TCP? If not, they absolutely should!

capital-G:
> > or we also experienced some dropping of UDP packages.
> 
> Don't the tests use TCP? If not, they absolutely should!

We have a PR about this, see https://github.com/supercollider/supercollider/pull/6463 and also the meta-issue https://github.com/supercollider/supercollider/issues/6439#issuecomment-2342955863

Though I'd be interested why we absolutely should do this? Normal way to interact with SC is through UDP so I think I'd prefer to test it this way (although it would allow to introduce some nondeterministic behavior through the host machine). It would be great to figure out why it occasionally fails though so we could tweak the test pipeline accordingly.

From my own experience I know that running in a virtualized environment that RT priorities and anything regarding RT-clocks behave wonky - but IIUC macOS isn't virtualized on github runners but those still fail as well?

Spacechild1:
Thanks for the links!

>  Though I'd be interested why we absolutely should do this? 

The problem, as also mentioned in the links, is that UDP is fundamentally unreliable.

> Normal way to interact with SC is through UDP

If we are testing the UDP communication itself, then yes, we should UDP. Otherwise I don't see how switching to a *more reliable* protocol would cause any issues.

(IMO, scsynth should never have used UDP in the first place because its unreliable nature leads to very leaky abstractions. There are so many parts of the code base that are essentially UDP workarounds. It also affects user code, see for example https://scsynth.org/t/how-best-to-slow-down-queue-outgoing-osc-messages/11350.)

capital-G:
Although UDP is an unreliable protocol it seems to me rather reliable when working on the same host - I don't have problems pushing millions of messages / sec around on my laptop w/o dropping - so I'd be surprised if the CI would have problems with some thousands/sec? Sclang/scsynth doesn't seem to move around too much network traffic after all.

> scsynth should never have used UDP in the first place

+1 :)

> Otherwise I don't see how switching to a _more reliable_ protocol would cause any issues.

Yeah, but b/c UDP is rather unreliable I think it is fine to follow a more integration based approach b/c thats how 99% of the users interact with it, so it would be great to be aware if UDP creates trouble at some point.
In the end it would simply be nice to understand why the CI is so unreliable, but I must admit I don't even know how to run the test suite locally, would love to have a simply CMake target for this. Although I would somehow still be hesitant to run the test-suite in a non-isolated environment, I have way too much fear that it will break something on my machine/sclang setup. But I am working on making sclang enivronments more isolated, stay tuned...

Spacechild1:
> Although UDP is an unreliable protocol it seems to me rather reliable when working on the same host - I don't have problems pushing millions of messages / sec around on my laptop w/o dropping

You are on macOS, right? IIRC macOS grows the socket receive buffer on demand. Linux and Windows, however, just drop incoming messages that don't fit into the buffer. UDP packet loss over localhost is a real thing! See again https://scsynth.org/t/how-best-to-slow-down-queue-outgoing-osc-messages/11350 for a real-life example.

> so it would be great to be aware if UDP creates trouble at some point.

It already does and always has.

> In the end it would simply be nice to understand why the CI is so unreliable, but I must admit I don't even know how to run the test suite locally, would love to have a simply CMake target for this.

That would be great! For example, I have some failing tests in my reblocking/upsampling branch, which I cannot explain, and I would like to find out if these are false positives or not.

server/scsynth/SC_Lib_Cintf.cpp
```diff
@@ -286,7 +286,7 @@ bool checkAPIVersion(void* f, const char* filename) {
 bool checkServerVersion(void* f, const char* filename) {
     if (f) {
         InfoFunction fn = (InfoFunction)f;
-        if ((*fn)() == 1)
+        if ((*fn)() != sc_server_scsynth)
```
JordanHendersonMusic, line 289: Could you just explain what is happening here please? Does this function only work on scsynth, not supernova?

Spacechild1:
> *** ERROR: failed to open UDP socket: address in use.
> This could be because another instance of scsynth is already using it.

Aha! Looks like scsynth does not use the `SO_REUSEADDR` socket flag (https://man7.org/linux/man-pages/man7/socket.7.html). 

As we know, normally you cannot bind multiple socket (of the same type) to the same port. However, this creates issues when you need to immediately restart a program while using the same port number. This is particularly a problem with TCP sockets as they can "linger" for quite a while (up to several minutes!), effectively blocking the port from being used by subsequent programs. The reason is that TCP sockets need to stay in the `TIME-WAIT` after being actively closed. See https://en.wikipedia.org/wiki/Transmission_Control_Protocol#/media/File:Tcp_state_diagram_fixed_new.svg. In theory, this doesn't apply to UDP sockets and we *should* be able to quickly recycle them. However, some systems apparently do not release UDP sockets fast enough.

This is particular problematic for CI testing because the Server might need to shutdown and restart in quick succession. (One workaround would be to use different Server ports everytime, e.g. one could simply increment the port number.)

When you set `SO_REUSEADDR` you *can* bind multiple sockets the same port. This avoids the problem of "lingering" sockets mentioned above. One downside is that you don't get a clear warning if you *accidentally* use an existing port number.

(Side note: in the case where multiple programs actually receive from the same socket, the behavior depends on the OS. IIRC on Linux incoming packets are delivered in a round robin fashion, so you can actually use this to increase networking throughput. On Windows only one application actually receives the packets.)

In Pd we decided to just set `SO_REUSEADDR` on all sockets.

Spacechild1:
> I'm a bit busy right now and can't test this, but do you know to what extent the unit tests cover these ugens?

Good question. I have to say I'm not really familiar with the test suite. @capital-G might be able to answer this question :)

server/scsynth/SC_Lib_Cintf.cpp
Spacechild1: Every plugin exports the following symbols:
- "load": the plugin entry point
- "api_version": a function that returns the plugin API version
- "server_version": a function that returns the server type:
  ```
  typedef enum { sc_server_scsynth = 0, sc_server_supernova = 1 } SC_ServerType;
  ```

See also the `PluginLoad` macro in `plugin_interface/SC_InterfaceTable.h`

`checkServerVersion` is responsible for the server type check. I just noticed that it used a magic number (1 = supernova) instead of the proper enum value. I also reversed the logic to make the intent more clear.

JordanHendersonMusic: Ah thank you!

dyfer:
> *** ERROR: failed to open UDP socket: address in use.

Re: testsuite:
The issue is that we don't wait for scsynth to close e.g. if it takes longer, there's higher overall CPU load etc, and in that case the whole remaining testsuite falls apart because we can't start another scsynth while the port is "taken". Our Server class in the language is also not that great about following the exact state of the server - there have been attempts to rewrite it, but haven't come to fruition yet.

Let's keep discussing TCP vs UDP for the test suite separately in the issues linker previously.

dyfer:
>  Also try with Server block sizes lower than 64 and 16 to run the non-SIMD version of certain process functions.

It seems like a good way to test this would be to write some new unit tests... I don't want to throw this at @Spacechild1 but I'm not sure when I'll be able to contribute the tests myself.

Spacechild1:
> It seems like a good way to test this would be to write some new unit tests... I don't want to throw this at @Spacechild1

JordanHendersonMusic (approved):
I've had a good read of this and tested with different block sizes using `TestIOUGens.run`. No issues. However, these tests are not the most extensive. @Spacechild1 can you think of any behaviour not covered by those tests that this code might touch? 

I've also tested this with #6405 which tests reordering many of these ugens and they compare the same.

Otherwise, I think this is good! Looking forward to the next PR towards reblocking :smiley:

dyfer:
> Should be no problem! The Server block size is independent from the hardware buffer size.

Ah duh! Of course. It would be great to make some tests for different block sizes. If I ever get to it...

Spacechild1:
> I've had a good read of this and tested with different block sizes using `TestIOUGens.run`. No issues.

Thanks a lot! That's very reassuring :)

> can you think of any behaviour not covered by those tests that this code might touch?

In theory, this PR shouldn't cause any regressions because it doesn't really change any functionality. I was just afraid I might have made a typo somewhere in the course and refactoring.

> Looking forward to the next PR towards reblocking 😃

The good thing is, if this PR proves to be ok, the reblocking/upsampling PR is pretty much guaranteed not to cause any regressions :)

dyfer (changes requested):
I'm marking "request changes" as I'm investigating this...

I've run the whole testsuite locally at `blockSize` **4** (I just changed the default value in the `Server.sc` file) and I got a number of failures:
<details>

```
// build 8e965b0efc8c9064a734a14792483650f14bdaa5 from the PR
// blockSize: 4

Failed tests:

	[4/8]	TestSanitize:test_sanitize (1.79s)
		 ! Sanitize.kr(8888)
		 ! Sanitize.kr(nan)
		 ! Sanitize.kr(-inf)
		 ! Sanitize.kr(inf)
	[7/10]	TestCoreUGens:test_delay1_2_predelayValues (2.24s)
		 ! Delay2.kr: predelay samples should defaults to the input value (0.4).
		 ! Delay2.kr: when x1 or x2 ([0.3, 0.0]) is a number, the corresponding predelay sample equals that number.
		 ! Delay2.kr: when x1 or x2 ([0.3, 0.6]) is a number, the corresponding predelay sample equals that number.
	[6/8]	TestCoreUGens:test_line_startAndEndSampleAreOuputOverDur (1.93s)
		 ! XLine.kr should not reach its end value before the given duration
		 ! Line.kr should not reach its end value before the given duration
	[49/50]	TestCoreUGens:test_oscillators_startAtCorrectPhase (4.38s)
		 ! LFSaw.kr should start at the correct initial phase
	[31/39]	TestUGen_RTAlloc:test_allUGens_allocPass (9.22s)
		 ! Convolution3 allocPass test should complete
		 ! StereoConvolution2L allocPass test should complete
		 ! Convolution2 allocPass test should complete
		 ! Onsets allocPass test should produce non-zero output
		 ! Convolution2L allocPass test should produce non-zero output
		 ! PartConv allocPass test should produce non-zero output
		 ! Convolution allocPass test should complete
		 ! IFFT allocPass test should produce non-zero output
	[3/4]	TestUGen_RTAlloc:test_FFT (1.85s)
		 ! should return a valid bufnum (bufnum: FloatArray[-1.0])
	[5/10]	TestUGen_RTAlloc:test_PVUGens (2.21s)
		 ! PV_RandComb should only return -1 on alloc fail
		 ! PV_BinScramble should only return -1 on alloc fail
		 ! PV_RandWipe should only return -1 on alloc fail
		 ! PV_Diffuser should only return -1 on alloc fail
		 ! PV_MagFreeze should only return -1 on alloc fail
```
</details

The testsuite run at blocksize 64 passed though.

dyfer (approved):

dyfer:
Tagging @joshpar @scztt @elgiano in accordance with the ["council structure" proposal](https://scsynth.org/t/move-supercollider-development-into-a-council-structure/11365#p-51974-a-first-draft-4)
I'd like to merge this, is it okay with everyone?

server/plugins/IOUGens.cpp
```diff
@@ -1165,7 +1209,7 @@ FLATTEN void XOut_next_a_nova(XOut* unit, int inNumSamples) {
 
     if (xfade0 != next_xfade) {
         float slope = CALCSLOPE(next_xfade, xfade0);
-        for (int i = 0; i < numChannels; ++i) {
+        for (int i = 0; i < numChannels; ++i, out += bufLength) {
```
mtmccrea, line 1212: Is there a reason `out` wasn't incremented before, or was this a bug?

Spacechild1: As the commit message indicates ("XOut: fix broken NOVA_SIMD version") this was a bug.

Spacechild1:
I just squashed all the IOPlugins commits into one.

elgiano:
Hi, I gave it a test too, it looks good to me! As a personal preference, I think it would be nicer for git history if we squashed the last part that only changes IOUGens in one single commit, keeping commit messages as summary, and then I would be happy to merge this. 

To be clearer, I mean everything from f4356da9ca46ffb3141a4aa110d7a50b3249fb8a (LocalIn: fix wrong output when Bus hasn't been written to) to the last commit.

Spacechild1:
> Could you explain if any of these changes to IOUGens affect users and should be communicated?

There are no changes that affect SC users.

I have added a new macro resp. SCUnit method, though:
https://github.com/supercollider/supercollider/blob/f159280af8fe938ea9ef5a53abb07861ba323aee/include/plugin_interface/SC_Unit.h#L115
https://github.com/supercollider/supercollider/blob/f159280af8fe938ea9ef5a53abb07861ba323aee/include/plugin_interface/SC_PlugIn.hpp#L196-L200

Spacechild1:
I don't, but you only have to check the history of the `include/plugin_interface` folder :) https://github.com/supercollider/supercollider/commits/develop/include/plugin_interface

PR #6683, opened 2025-03-24
Change the style of path in Windows by updating ScIDE.sc

prko:
## Purpose and Motivation

The issue of Windows paths being displayed in the UNIX manner has been partially addressed in this PR.
I would like to extend my gratitude to @JordanHendersonMusic and @telephon for their insights and contributions, which made this improvement possible:
https://github.com/supercollider/supercollider/pull/6291#issuecomment-2747742875

Please note, however, that this PR does not address instances where a user manually types "/" in the code. Addressing such cases will require a separate PR in the future.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -394,6 +394,10 @@ ScIDE {
 		}
 	}
 
+	currentPath_ { |p|
+		currentPath = p.standardizePath;
+	}
+
 
 	// PRIVATE ///////////////////////////////////////////////////////////
 


JordanHendersonMusic (approved):

JordanHendersonMusic:
> Unfortunately, standardizePath does not replace / with \:

@Spacechild1 

Okay now I'm confused because it seems to do just that, see this PR and the linked comment.

There is nothing in the primitive to suggest that this happens. 

What version of SC has `standardizePath` been tested on, I wonder if the migration from boost's filesystem to c++'s might play a role?

PR #6685, opened 2025-03-26
Add BufferExamplesPath class

### LINKED ISSUE #5859: Add BufferExamplePath class
## Motivation

*Forked* from https://github.com/supercollider/supercollider/issues/5604 which discusses adding more included samples to SC.

```supercollider
b = Buffer.read(s, Platform.resourceDir +/+ "sounds/a11wlk01-44_1.aiff");
```

is often used for examples or in teaching environments as it is the only one of two included samples in SC .
Both samples do not have a trivial filenames, making it necessary to copy these lines all the time (which are hidden in examples or in the `Buffer` documentation).
It also lacks any kind of documentation what those samples are.

Adding a new Class which abstracts those file paths makes accessing those examples more easy via auto-completion and dedicated documentation, also allowing for an interface to add new examples (see #5604 ).

Python libraries also handle this via namespaces/classes, see https://github.com/supercollider/supercollider/issues/5604#issuecomment-1232926043

## Plan for Implementation

Implementation is rather trivial, it could look something like

```supercollider
BufferExamplePaths {
  *nasa { ^Platform.resourceDir +/+ "sounds/a11wlk01-44_1.aiff" }

  *baby { ^Platform.resourceDir +/+ "sounds/a11wlk01.wav" }
}
```

replacing

```supercollider
b = Buffer.read(s, Platform.resourceDir +/+ "sounds/a11wlk01-44_1.aiff");
```

with

```supercollider
b = Buffer.read(s, BufferExamplePaths.nasa);
```

capital-G:
## Purpose and Motivation

Closes #5859

Hopefully I am not the only one who would consider such a class helpful - still can't remember `a11wlk01-44_1.aiff` after some years...

Docs:

Screenshot 2025-03-26 at 20-28-47 BufferExamplesPath SuperCollider 3 14 0-dev Help

In a future this class could be used to add additional sound examples.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

Questions:

* [x] Should we add tests? (Would allow additional assurance that we don't move files by accident?)
* [ ] Should we add some information about license? (can be added later?)
* [ ] Should such a class return Strings or PathName? Strings seem to be the more common way when operating with Buffers though?

### DIFF
--- a/HelpSource/Classes/BufferExamplesPath.schelp
+++ b/HelpSource/Classes/BufferExamplesPath.schelp
@@ -1,0 +1,61 @@
+TITLE:: BufferExamplesPath
+summary:: A shortcut to example sounds bundled with SuperCollider
+categories:: Files
+related:: Classes/Buffer, Classes/Soundfile
+
+DESCRIPTION::
+Provides a shortcut to the paths of example sounds which are bundled with SuperCollider and are located in LINK::Classes/Platform#*resourceDir::.
+
+CODE::
+// instead of writing
+Platform.resourceDir +/+ "sounds" +/+ "a11wlk01-44_1.aiff";
+// one can write
+BufferExamplesPath.apollo11;
+::
+
+CLASSMETHODS::
+
+METHOD:: apollo11 
+A radio recording from the Apollo 11 moon landing program where Bruce McCandless says EMPHASIS::"Columbia, this is Houston. Over."::.
+For more information see LINK::https://www.nasa.gov/history/alsj/a11/a11.mobility.html:: at the mark TELETYPE::110:25:41:: (it seems to be the third repetition, though the pitch does not match).
+
+TABLE::
+## Filename || TELETYPE::a11wlk01-44_1.aiff::
+## Channels || 1
+## Samplerate || 44100 Hz
+## Format || pcm_s16be
+## Duration || 00:00:02.44
+::
+
+RETURNS:: The full path of TELETYPE::a11wlk01-44_1.aiff:: as a LINK::Classes/String::
+
+DISCUSSION::
+Trivia time: Although the file in its current form was introduced on 2004-07-11 by Scott Wilson via commit TELETYPE::bc9a4d4fd8cdccd2b1c787010c776176c7993d2c::, its original version as a 11025 Hz mono file dates back to at least 2002-09-28 via the commit TELETYPE::f30d769ab6b8eb1f36027ed3a400efd3b829f43a:: by James McCartney.
+
+This can be considered the "Hello world" audio sample of SuperCollider.
+
+METHOD:: sinedPink
+A very short stereo sample which has 10 cycles of a 440Hz LINK::Classes/SinOsc:: on its first channel and a LINK::Classes/PinkNoise:: on its second channel. 
+
+TABLE::
+## Filename || TELETYPE::SinedPink.aiff::
+## Channels || 2
+## Samplerate || 44100 Hz
+## Format || pcm_f32be
+## Duration || 00:00:00.02
+::
+
+returns:: The full path of TELETYPE::SinedPink.aiff:: as a LINK::Classes/String:: 
+
+METHOD:: child
+Allegedly a re-recording of Link::Classes/BufferExamplesPath#*apollo11:: performed by a child of a developer.
+
+TABLE::
+## Filename || TELETYPE::a11wlk01.wav::
+## Channels || 1
+## Samplerate || 44100 Hz
+## Format || pcm_s16le
+## Duration || 00:00:04.28
+::
+
+returns:: The full path of TELETYPE::a11wlk01.wav:: as a LINK::Classes/String::
--- a/HelpSource/Classes/Buffer.schelp
+++ b/HelpSource/Classes/Buffer.schelp
@@ -1,6 +1,7 @@
 class:: Buffer
 summary:: Client-side representation of a buffer on a server
 categories:: Server>Abstractions
+related:: Classes/BufferExamplesPath
 
 description::
 
--- a/SCClassLibrary/Common/Files/BufferExamplesPath.sc
+++ b/SCClassLibrary/Common/Files/BufferExamplesPath.sc
@@ -1,0 +1,13 @@
+BufferExamplesPath {
+	*apollo11 {
+		^Platform.resourceDir +/+ "sounds" +/+ "a11wlk01-44_1.aiff";
+	}
+
+	*sinedPink {
+		^Platform.resourceDir +/+ "sounds" +/+ "SinedPink.aiff";
+	}
+
+	*child {
+		^Platform.resourceDir +/+ "sounds" +/+ "a11wlk01.wav";
+	}
+}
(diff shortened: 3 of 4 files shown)


telephon (dismissed):
I think you should also add the classic `floating_1` sound file …

capital-G:
> I think you should also add the classic `floating_1` sound file …

But `floating_1` is not included in the SC distribution anymore :/

cdbzb:
Love the details in the docs here! ... and pardon the noise but at the Symposium there was discussion of asking the child who performed a11wlk01 to rerecord it for the next update (she is perhaps in her late 20s)...

HelpSource/Classes/BufferExamplesPath.schelp
```diff
@@ -0,0 +1,61 @@
...
+related:: Classes/Buffer, Classes/Soundfile
+
+DESCRIPTION::
+Provides a shortcut to the paths of example sounds which are bundled with SuperCollider and are located in LINK::Classes/Platform#*resourceDir::.
+
+CODE::
+// instead of writing
+Platform.resourceDir +/+ "sounds" +/+ "a11wlk01-44_1.aiff";
+// one can write
+Platform.apollo11;
```
JordanHendersonMusic, line 13: I think you mean `BufferExamplesPath.apollo11`

JordanHendersonMusic (commented):
Great idea! 

Yes to tests.

No opinion/knowledge on the licensing.

HelpSource/Classes/BufferExamplesPath.schelp
capital-G: Thanks - fixed!

capital-G:
> Yes to tests.

Done

> I'd return a string because it is far more common, although I wish pathname was used more.

Same (but even Python struggles with phasing out str in favor of pathlib - and I found the sclang pathlib not as reliable as I'd like)

HelpSource/Classes/BufferExamplesPath.schelp
```diff
@@ -0,0 +1,61 @@
...
+
+RETURNS:: The full path of TELETYPE::a11wlk01-44_1.aiff:: as a LINK::Classes/String::
+
+DISCUSSION::
+Trivia time: Although the file in its current form was introduced on 2004-07-11 by Scott Wilson via commit TELETYPE::bc9a4d4fd8cdccd2b1c787010c776176c7993d2c::, its original version as a 11025 Hz mono file dates back to at least 2002-09-28 via the commit TELETYPE::f30d769ab6b8eb1f36027ed3a400efd3b829f43a:: by James McCartney.
+
+This can be considered the "Hello world" audio sample of SuperCollider.
+
+METHOD:: sinedPink
+A very short sample stereo sample which has a 10 cycles of a 440Hz LINK::Classes/SinOsc:: on its first channel and a LINK::Classes/PinkNoise:: on its second channel.
```
JordanHendersonMusic, line 38: Double 'sample'. 
Also, 'which has 10 cycles' (no 'a').

capital-G: thx - fixed

JordanHendersonMusic:
Do we think that some examples in the rest of the documentation could be updated to use this new classes. 

I'm just thinking about how users will discover this class?

capital-G:
@telephon anything missing to get this merged?

prko:
Is there no need to change all the sample paths with these?
I could do it if you think changing all paths to this new class and methods!
I ask first before I do it.

capital-G:
> Great!
> 
> Is there no need to change all the sample paths with these? I could do it if you think changing all paths to this new class and methods! I ask first before I do it.

I think you are referring to the paths of `a11wk...` in the docs? I just counted that there are 117 mentions of `a11wlk` in the docs, so there are some spots where this could be applied.

IMO it should be replaced for most parts: e.g. if you want to demonstrate some buffer functionality, replace it - if you want to talk about how to actually load audio files using a buffer - maybe don't replace it and use some raw paths in order to be more transparent?

But maybe this is a process which @tedmoore can guide? :)

prko:
Furthermore, I noticed that `SinedPink` appears nine times (excluding the instance in `BufferExamplesPath.schelp` in the SCDoc).

>But maybe this is a process which @tedmoore can guide? :)

Yes, the process is best guided by him.

dyfer:
Sorry coming late to this - was an alternative to the word "Buffer" considered? To me SoundFile seems like a more accurate reference to... well... sound files (SoundFileExamplesPath or SoundFilePaths would be my preference).
But I also understand if we want to keep this as is, since most of them are used in the Buffer examples.

tedmoore:
I was also late to this. It could just be `ExampleFiles` in case the folder ends up containing files that aren't even sound files (json? sdif?).

Also, I'm not crazy about accessing the paths through methods. My heart would like closer to how [FluCoMa did their path constructing class](https://github.com/flucoma/flucoma-sc/blob/main/release-packaging/Classes/FluidFilesPath.sc) in which one has to pass in the file name, so basically the class is only handling constructing the beginning of the path to the folder.

tedmoore:
> has the advantage of code completion hints

I like code completion hints. And I agree not needing to know the extension is clean.

I am inclined towards `ExampleFiles`. What does @capital-G think of this?

capital-G:
> I am inclined towards changing the name of this class to like `ExampleFiles`, `ExamplesPath`, or `FilesPath`. What does @capital-G think of this?

One intention of this rather clunky class name was to make users who type in `Buffer` aware of `BufferExamplesPath` - the same for making this a static method so one can quickly access common example files w/o looking up any filenames. If we would have (better) LSP support we could probably add adjunct classes in the autocomplete which don't share the beginning of the name through docs - but this is another discussion.

Currently I don't think we have too many other static files bundled within SC (I think there are 1/2 images?), so is there actually a need to reference other example files?
OOTH - we could also create additional classes for other file types - I don't think we necessarily bundle everything in one class and I found names which explain directly what they represent/doing mostly a good way to organize code.

But if we want to go with an unified approach, I'd favor `ExampleFiles` :)

I am fine with any way, as long as it serves usability, clarity and consistency :)

dyfer:
> One intention of this rather clunky class name was to make users who type in `Buffer` aware of `BufferExamplesPath`

Ah, okay. This explains the practical usability of the current name.

I still think it's unfortunate, as it's the examples are more about soundfiles than buffers...

> Currently I don't think we have too many other static files

My objective was not to refer to other (non-sound) files, but rather to not refer to sound files as BufferSomething. But I understand that having a hint when typing would be useful... except, again, these are not buffers! :) 

I'd have a slight preference to change this to `ExampleFiles`, but if there's otherwise a consensus to keep this as `BufferExamplesPath`, then won't oppose sticking to that.

capital-G:
I think while this isn't part of a major release we can still rename it, but afterwards we should make its name stable. If it wouldn't be for the auto-complete I would have named it differently - so maybe not the best reason/argument to name something? :)

dyfer:
We certainly can still rename it before the 3.14 release, compatibility-wise.

tedmoore:
I think there's a good argument for `ExampleFiles` so I vote we go with that for 3.14. 

That will have to happen before https://github.com/supercollider/supercollider/pull/6705 https://github.com/supercollider/supercollider/pull/6706 https://github.com/supercollider/supercollider/pull/6707 So maybe I'll try to make a PR renaming the class (my next few weeks are pretty full with end-of-semester things so we'll see).

tedmoore:
> maybe I'll try to make a PR renaming the class

This now exists: https://github.com/supercollider/supercollider/pull/6720

PR #6690, opened 2025-03-30
plugin interface: fix ClearUnitOnMemFailed macro

Spacechild1:
## Purpose and Motivation

The `ClearUnitOnMemFailed` macro must zero the first sample of all outputs before returning from the current function; otherwise the output will contain garbage.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


elgiano (commented):
looks good to me, thanks! (also thanks for adding the extra comment!)

elgiano (approved):

dyfer:
Sorry for a stupid question... this doesn't break the plugin interface, does it?

PR #6693, opened 2025-03-31
TestUGen_RTAlloc: don't set samplerate or blocksize

dyfer:
## Purpose and Motivation

When I was running the TestUGen_RTAlloc test locally, it failed to boot the server, as my hardware was not able to run at the requested samplerate (there might be a separate bug here...).
I don't believe tests should request a specific samplerate, as this is environment dependent. AFAICT this is the only test that actually requests a particular sample rate.

I understand that there might be tests that are samplerate-specific, but I think handling such cases should be discussed separately.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestUGen_RTAlloc.sc
+++ b/testsuite/classlibrary/TestUGen_RTAlloc.sc
@@ -8,8 +8,6 @@ TestUGen_RTAlloc : UnitTest {
 
 	setUp {
 		server = Server(this.class.name);
-		server.options.sampleRate = 48000;
-		server.options.blockSize = 64;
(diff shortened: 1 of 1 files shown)


telephon (approved):

dyfer:
Thanks for the context @elgiano. Currently the testsuite runs at 48k so setting it was redundant for the CI, while it caused issues when running the test locally.
After running the tests a few times at 44100 it I got a number of failures, but upon subsequent runs - no failures... so I guess the tests themselves are as reliable at 44.1k as they are at 48k.

PR #6695, opened 2025-04-02
docs: ways of getting current doc's path [skip ci]

### LINKED ISSUE #5757: File.getcwd results differently on Windows
## Environment

* SuperCollider version: 3.12.2
* Operating system: Windows

## Steps to reproduce

```supercollider
File.getcwd
```

On Linux, `File.getcwd` gives the user's home directory. On Windows it gives the script's working directory, like `thisProcess.nowExecutingPath.dirname`. This difference can be confusing and create portability problems, for example when writing SC code on windows and then deploying it on a Raspberry Pi, and I couldn't find references to it in the docs.

I faced this problem while working with FileReader.

elgiano:
## Purpose and Motivation

Fixes #5757

Attempting to put all the knowledge emerged from #5757 in the docs.
`File.getcwd` now lists common defaults for each OS, mentions how to affect cwd via CLI and IDE, and specifies that this is not necessarily the current document's path.
`File.cwd`, `thisProcess.nowExecutingPath`, `String.resolveRelative` and `Document.current.dir` all refer to each other as "see also".
I also added some examples to thisProcess.nowExecutingPath, trying to clarify the differences with String.resolveRelative and filenameSymbol.

Maybe it's too much information? As often, in SC we have different method to do almost the same thing. In this case, they all do something slightly different and have different reasons to exist, so let's document them all. Perhaps we should a dedicated "Guide" section somewhere, instead of having most info in `thisProcess.nowExecutingPath` and cross references everywhere?

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Document.schelp
+++ b/HelpSource/Classes/Document.schelp
@@ -77,6 +77,7 @@ A Document.
 discussion::
 code::
 Document.current.name.postln; // Prints "Document.html"
+Document.current.dir; // Prints the current document's path
 ::
 
 method:: allDocuments
@@ -121,6 +122,8 @@ code::
 Document.dir = "~/Documents/SuperCollider";
 ::
 
+To get the current document's directory, see link::Classes/Document#-dir::.
+
 method:: standardizePath
 argument:: p
 The file system path to the directory. An instance of link::Classes/String::.
@@ -162,12 +165,18 @@ code::
 Document.current.path.postln;
 ::
 
+See also link::Classes/Process#-nowExecutingPath:: and
+link::Classes/String#-resolveRelative::
+
 method:: dir
 Returns the directory of a Document.
 discussion::
 code::
 Document.current.dir.postln;
 ::
+
+See also link::Classes/Process#-nowExecutingPath:: and
+link::Classes/String#-resolveRelative::
 
 method:: ==
 A binary operator.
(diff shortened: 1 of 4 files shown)


JordanHendersonMusic:
Did you also see #6678? It moves filename symbol into functiondef so works with all functions.

dyfer:
@elgiano FYI there's no need to use [skip ci] for documentation - our full CI does not run for docs-only changes anymore.

prko:
I wanted to bring PR #6291 to your attention.

@elgiano, feel free to incorporate any relevant text from PR #6291 into your work on this PR if it would be helpful. I have already suggested the same to @JordanHendersonMusic regarding their PR #6678.

This matter requires immediate attention. My proposed document has been pending approval for 11 months and has undergone several revisions during this time. If portions of my text can be integrated into your respective PRs and subsequently approved in a timely manner, I would consider that a satisfactory resolution.

Given the circumstances, I believe it would be prudent to close PR #6291 once its relevant content has been successfully incorporated elsewhere.

Thank you for your cooperation and understanding.

elgiano:
Hi everyone! I hadn't seen these PRs, I see that's a lot of work going on about this confusing topic :D
I'll close this PR, so we keep focus on the others. When #6678 is done, we'll talk about docs again

@prko I hear you, sometimes things can get very lengthy and it can get frustrating, especially when things are delicate and/or confusing. Thanks for your efforts! Also, sometimes even if a PR is not merged, that work contributes anyways to discussing the subject, bringing attention to it, and develop some ideas for later. Or even if it shows a direction not to be  pursued, it helps letting better ones emerge. Let's see how #6678 goes and take it from there!

PR #6699, opened 2025-04-03
sclang: fix `mod` behavior for negative integer modulus

### LINKED ISSUE #6698: `mod` with negative modulo gives unexpected result over negative integers
## Environment

* Operating system: macOS 15.3.2

## Steps to reproduce

```supercollider
[-5, -4, -3, -2, -1, 0] % -5
// -> [0, 1, 2, 3, 4, 0]
[-10, -9, -8, -7, -6, 0] % -5
// -> [0, -9, -8, -7, -6, 0]

/* and you might find this interesting: (correct behavior for non-integers) */
[-5.1, -4.1, -3.1, -2.1, -1.1, -0.1] % -5
// -> [-0.1, -4.1, -3.1, -2.1, -1.1, -0.1]
[-10.1, -9.1, -8.1, -7.1, -6.1, -0.1] % -5
// -> [-0.1, -4.1, -3.1, -2.1, -1.1, -0.1]
```

Expect, for the first block, `[0, -4, -3, -2, -1, 0]` for both.

Actually gives non-minimal remainders. These results are all technically congruent `mod -5`, but `mod` ideally should give a result like in C/++ (as it does for positive integers!), where for `a % b`, it returns the difference of `a` and the product of `b` and the truncated result of `a/b`, in the range of $[0$, `b`$]$.

passyur:
## Purpose and Motivation

Fixes #6698. Currently, `mod` for negative integers gives unexpected results (particularly, when that negative integer is the modulus). This fix puts `mod`'s behavior for `int`s closer to `mod`'s behavior for `double`s and `float`s and so applies a consistent definition of modulo.

## Types of changes

- Bug fix
- Breaking change (output changes)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Documentation updated

### DIFF
--- a/HelpSource/Classes/Integer.schelp
+++ b/HelpSource/Classes/Integer.schelp
@@ -227,12 +227,32 @@ method:: -
 Subtraction.
 
 method:: modSeaside
-Pre-3.14 modulo with unexpected behavior for negative integer modulus. E.g., 
+
+Pre-3.14 modulo with unexpected behavior for negative integer modulus.
+
+For dividends smaller than a negative divisor, modSeaside pulls the resulting remainder below the divisor.
+It thereby leaves zero as peaks on the zero line. Like in the usual mod, all larger dividends result in remainders above zero.
+
 code::
-(..-9).mod(-3) 
+(..-9).mod(-3);
 // returns [0, -1, -2, 0, -1, -2, 0, -1, -2, 0]
-(..-9).modSeaside(-3)
+(..-9).modSeaside(-3);
 // returns [0, 2, 1, 0, -4, -5, 0, -4, -5, 0]
+
+// plot
+(-25..25).modSeaside(-5).plot.plotMode_(\steps);
+
+// modSeaside
+Pbind(\note, modSeaside(Pseries(-15, 1, 25), -5), \dur, 0.2).play;
+
+// modSeaside with different divisors
+Pbind(\note, modSeaside(Pseries(-15, 1, 25), [3, -4, -5]), \dur, 0.2).play;
+
+// mod
+Pbind(\note, mod(Pseries(-15, 1, 25), -5), \dur, 0.2).play;
+
+// mod with different divisors
+Pbind(\note, mod(Pseries(-15, 1, 25), [3, -4, -5]), \dur, 0.2).play;
 ::
 
 method:: clip
--- a/HelpSource/Classes/Integer.schelp
+++ b/HelpSource/Classes/Integer.schelp
@@ -226,6 +226,15 @@ Addition.
 method:: -
 Subtraction.
 
+method:: modSeaside
+Pre-3.14 modulo with unexpected behavior for negative integer modulus. E.g., 
+code::
+(..-9).mod(-3) 
+// returns [0, -1, -2, 0, -1, -2, 0, -1, -2, 0]
+(..-9).modSeaside(-3)
+// returns [0, 2, 1, 0, -4, -5, 0, -4, -5, 0]
+::
+
 method:: clip
 returns::
 LIST::
--- a/HelpSource/Overviews/Operators.schelp
+++ b/HelpSource/Overviews/Operators.schelp
@@ -303,7 +303,8 @@ method:: /
 Division.
 
 method:: %
-Floating point modulo.
+Modulo.
+NOTE:: Uses floored division. Prior to SC 3.14, integer modulo would exhibit unexpected behavior for negative modulus; see link::Classes/Integer#-modSeaside::.::
 
 method:: **
 Exponentiation. Same as pow.
--- a/HelpSource/Classes/Float.schelp
+++ b/HelpSource/Classes/Float.schelp
@@ -76,6 +76,9 @@ See also: link::Guides/Randomness::
 method::xrand2
 returns::a random float from this.neg to this, excluding the value exclude.
 
+method::modSeaside
+returns::the same result as code::this.mod(aNumber, adverb)::; provided for compatibility with link::Classes/Integer#-modSeaside::.
+
 method:: isFloat
 returns:: code::true:: since this is a Float.
 
--- a/HelpSource/Classes/Number.schelp
+++ b/HelpSource/Classes/Number.schelp
@@ -27,6 +27,9 @@ Integer division.
 method:: %
 Modulo.
 
+method:: modSeaside
+Pre-3.14 modulo with unexpected behavior for negative integer modulus; see link::Classes/Integer#-modSeaside::. Calling code::modSeaside:: on a floating-point operand will fall back to code::mod:: behavior.
+
 method:: **
 Exponentiation.
 
--- a/HelpSource/Classes/SequenceableCollection.schelp
+++ b/HelpSource/Classes/SequenceableCollection.schelp
@@ -732,6 +732,9 @@ method:: +, -, *, /, div, min, max, <, <=, >, >=, bitXor, lcm, gcd, round, trunc
 
 method:: %, **, &, |, >>, +>>, <!
 
+method:: modSeaside
+Pre-3.14 modulo with unexpected behavior for negative integer modulus; see link::Classes/Integer#-modSeaside::. Calling code::modSeaside:: on a non-integer operand will fall back to code::mod:: behavior; see link::Classes/Float#-modSeaside::.
+
 method::performBinaryOp
 Creates a new collection of the results of applying the selector with the operand to all elements in the receiver. If the operand is a collection then elements of that collection are paired with elements of the receiver.
 code::
--- a/SCClassLibrary/Common/Collections/SequenceableCollection.sc
+++ b/SCClassLibrary/Common/Collections/SequenceableCollection.sc
@@ -850,6 +850,7 @@ SequenceableCollection : Collection {
 	/ { arg aNumber, adverb; ^this.performBinaryOp('/', aNumber, adverb) }
 	div { arg aNumber, adverb; ^this.performBinaryOp('div', aNumber, adverb) }
 	mod { arg aNumber, adverb; ^this.performBinaryOp('mod', aNumber, adverb) }
+	modSeaside { arg aNumber, adverb; ^this.performBinaryOp('modSeaside', aNumber, adverb) }
 	pow { arg aNumber, adverb; ^this.performBinaryOp('pow', aNumber, adverb) }
 	min { arg aNumber, adverb; ^this.performBinaryOp('min', aNumber, adverb) }
 	max { arg aNumber=0, adverb; ^this.performBinaryOp('max', aNumber, adverb) }
--- a/SCClassLibrary/Common/Core/AbstractFunction.sc
+++ b/SCClassLibrary/Common/Core/AbstractFunction.sc
@@ -100,6 +100,7 @@ AbstractFunction {
 	/ { arg function, adverb; ^this.composeBinaryOp('/', function, adverb) }
 	div { arg function, adverb; ^this.composeBinaryOp('div', function, adverb) }
 	mod { arg function, adverb; ^this.composeBinaryOp('mod', function, adverb) }
+	modSeaside { arg function, adverb; ^this.composeBinaryOp('modSeaside', function, adverb) }
 	pow { arg function, adverb; ^this.composeBinaryOp('pow', function, adverb) }
 	min { arg function, adverb; ^this.composeBinaryOp('min', function, adverb) }
 	max { arg function=0, adverb; ^this.composeBinaryOp('max', function, adverb) }
--- a/SCClassLibrary/Common/Core/Symbol.sc
+++ b/SCClassLibrary/Common/Core/Symbol.sc
@@ -166,6 +166,7 @@ Symbol {
 	* { ^this }
 	/ { ^this }
 	mod { ^this }
+	modSeaside { ^this }
 	min { ^this }
 	max { ^this }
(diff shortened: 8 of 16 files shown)


dyfer:
@JordanHendersonMusic @scztt what do you think about this change?

JordanHendersonMusic:
The change here is in a plugin interface header, this means it affects the class library and the server plugins, including any other plugins not written by us. I agree the behaviour is weird, but it might have been here so long that people's work depends on it.

dyfer:
I agree that there is almost certainly code depending on this (arguably not desirable) behavior.

One theoretical possibility to address this would be to
1. deprecate the old method
2. include a new method with the correct behavior
but that's not practical with a universal math symbol.

I'm not sure what's the best path forward... but I'd love to get the proposed behavior in, somehow. Something to discuss at the upcoming [dev meeting](https://scsynth.org/t/supercollider-dev-meeting-01-2025-2025-04-06/11448)!

elgiano:
A few points to contribute to the discussion:

- unexpected results are given only when modulo is negative (e.g. `x % -3`).
- plugins would most probably use the float version IMO: they are server plugins, and most probably deal with signals. A case for ints is samples or sizes, but those are usually positive mods.
-  `-7 % -3 = -4` (out of range, should be -1). This is not only weird, but definitely a bug.

For these reasons, I lean towards fixing it, with a breaking change note in changelogs.

---------------------------------------------
now, the confusing part:
there are in general two conventions for negative modulo, adopted by different programming langs:

**Same sign as dividend**: `a % b = a - b * trunc(a/b)` (e.g. cpp): 
-7 % 3 = -1
7 % -3 = 1

**Same sign as divisor** `a % b = a - b * floor(a/b)` (e.g. python): 
-7 % 3 = 2
7 % -3 = -2

_(in both cases: -7 % -3 = -1)_

For floats, sc uses the same as [python](https://docs.python.org/3/reference/expressions.html#binary-arithmetic-operations). Also, sc relies on the same behavior for wrapping list indices (e.g. -1 % 10 = 9). I think we should stick to this, make it consistent and document it.

```supercollider
(..6) % 3   // -> [0, 1, 2, 0, 1, 2, 0] like python
(..-6) % 3  // -> [0, 2, 1, 0, 2, 1, 0] like python
(..6) % -3  // -> [0, 1, 2, 0, 1, 2, 0] like cpp
(..-6) % -3 // -> [0, 2, 1, 0, -4, -5, 0] like no one else
```

python has an `fmod` function, to implement the opposite behavior. Maybe in the future we could implement an `fmod` function in sc too (although perhaps with a less confusing name)?

dyfer:
@elgiano thanks for your thoughts!

> (although perhaps with a less confusing name)?

yes! In a true SC fashion: `mod1` :)

> plugins would most probably use the float version IMO:

I also thought about this aspect - I agree that this is likely an issue to consider mostly in terms of sclang.

JordanHendersonMusic:
```
(..-6) % -3 // -> [0, 2, 1, 0, -4, -5, 0] like no one else
```
I see, yes that is unusable.

Okay in that case I think this implementation is good, but that there should be some sclang unit tests.

telephon:
Normally, yes, but – I don't think it is such a good idea changing the current implementation, which has been in there since the [initial commit](https://github.com/supercollider/supercollider/blame/5079d1445bc9717b4b23a0afe0b944d7ad9c5eb6/Headers/plugin_interface/SC_InlineBinaryOp.h#L372C2-L375C11) 23 years ago. There will be code that subtly depends on it.

I'd say document the current way it is and suggest an alternative if you need another way (either by adding a method as suggested `mod1`, or by suggesting using doubles).

I would be in favour of the change of course if someone would give a convincing argument that nobody will have ever depended on sclang with negative integers in modulo arithmetic in any important way.

passyur:
Since this change only affects about a quarter of all purely integer modulo combinations (those that are negative integers modulo negative integers), I'm gonna suggest that a sufficient argument for the change is that it won't _irreparably_  break anyone's code. So, add a function like `oldMod` that implements the old behavior and document the change. If `oldMod` doesn't receive any use, then it can be removed. If someone notices the breaking change, then they can switch to `oldMod`. 

It also seems like better language practice to have `%` behave like `%` should. A sociological argument is that the affected cases of the modulo aren't too significant since no one has hitherto reported on this weird behavior. (But someone with a better way around the places where people might've mentioned that can correct me.)

elgiano:
@passyur just to be correct, also positive integers with negative modulo (e.g 3 % -2) will be affected. But I also can't find examples of negative modulo divisor in use, so just out of curiosity, how have you encountered this behavior?

One point about `oldMod`: it might not be possible to switch back to oldMod for scsynth plugins, since they would have to do it at compile time. However, no built-in or sc3-plugin uses integer sc_mod, so it would be up to eventual external plugin developer to switch to `oldMod` if they prefer.

dyfer:
> One point about `oldMod`: it might not be possible to switch back to oldMod for scsynth plugins,

I wouldn't consider this to be an issue with plugins, since floats are used for signals and that's the only place I'd really be concerned about it. IMO we should mostly consider the class library use... It's almost certain that somebody somewhere used this (incorrect) behavior, but I also think that it's so unreasonable that it is worth correcting.

I'm _tentatively_ supportive of adding `oldMod` or similar.

At the very least, once there's a consensus how to move forward, this PR should add updated documentation indicating this change.

passyur:
For your curiosity @elgiano: I encountered this behavior while considering how to implement `Signal`'s `mod` function in #6659. So, not while actually coding anything where it was systemically important.

telephon:
Given the fact that the behaviour is so erratic, the existing uses for it will probably be musical. I do think that it is reasonable to change it and if someone wants the old version, use `modOld` which keeps the old version.

elgiano:
I agree with @telephon, it's reasonable and although its saves a mathematically "wrong" primitive, it also saves a possible musical application of a "wrong" mathematical function, which I think fits with SuperCollider's spirit.

Can we agree on what this PR should do then? AFAIC:

1. the new sc_mod primitive for integers is fine
2. add another primitive `sc_mod_old`? I think it in this case it could be done just in sclang, even though we would be ignoring server plugins eventually wanting to use the old impl
3. add a note in schelp: perhaps just in `Overviews/Operators.schelp`?
4. a test for mod behavior

and that would be it for me. If we ever wanted to implement the "other" convention for mod (like python's fmod), we could do it in another PR.

HelpSource/Classes/Integer.schelp
```diff
@@ -226,6 +226,15 @@ Addition.
 method:: -
 Subtraction.
 
+method:: modOld
+The original SC modulo for integers, with unexpected (but constant) output for negative modulus. E.g., 
+code::
+(..-6).mod(-3) 
+// returns [0, -1, -2, 0, -1, -2, 0]
+(..-6).modOld(-3)
+// returns [0, 2, 1, 0, -4, -5, 0]
```
dyfer, line 235: > with unexpected (but constant) output

Just a minor comment: I'm not sure what "constant" means in this context. I think "unexpected" alone is fine; other alternatives to "unexpected" I'd consider would be "arguably unexpected" or "peculiar".

As for the examples, just to reiterate (hm...) the behavior for less technical users, I'd use an array until -9; also what do you think about writing the code and comment in a single line?

```supercollider
(..-9).mod(-3); // [0, -1, -2, 0, -1, -2, 0, -1, -2, 0]
(..-9).modOld(-3); // [0, 2, 1, 0, -4, -5, 0, -4, -5, 0]
```

passyur: About "constant": I was trying to (maybe prematurely) disabuse a reader from thinking that it was somehow non-deterministic. Don't think that "constant" is as clear, in hindsight; I like your suggestions for alternatives since they specify that it's a matter of interpretation.

About the code comments and line breaks: I'm often reading the docs in the help panel docklet, which isn't particularly wide. The wrap around makes it a bit harder to follow, to me at least. If it weren't such a large array, I'd probably feel differently.

I do support (..-9), since it shows that the first cycle of the modulo is actually unique.

elgiano:
@dyfer @capital-G @telephon we just discussed this on the dev meeting, was the consensus to renounce modOld?

capital-G:
> @dyfer @capital-G @telephon we just discussed this on the dev meeting, was the consensus to renounce modOld?

I think since the current implementation can be considered buggy it is fine to simply replace it w/ a consistent implementation. IMO we don't need to support buggy/inconsistent code, although it would be great if there are code snippets within sclang which will obtain the old result in e.g. the changelog.

I think it would be great if we have a document which specifies deprecation policies - it seems many contributions to sclang are striking on this in some way or another - maybe the class library team can come up with some document which provides guidelines and examples for (the not really existing) deprecation? :) At least I'd appreciate it.

dyfer:
I thought that in this case we were thinking about indeed keeping modOld. 

If we really don't want that, another option would be to push methods like this into a quark and make an sclang-only implementation of this, what do you think?

telephon:
Let's keep the current version under the name of `mod_old`, if someone needs it.

dyfer:
We have a dedicated [place](https://github.com/supercollider/supercollider/tree/develop/SCClassLibrary/deprecated) for deprecated methods already. 
However, historically we haven't been putting the "_old" methods there IIRC, since the idea of backwards compatibility was about maintaining existing SC pieces to work in the future, as much as possible. 
Ultimately, I think this one is small enough that keeping it around isn't too bad.
Eventually, if we ever get to a point of language plugins and an infrastructure for "compatibility mode", it could be moved out of the main library along with other "_old" methods. But for now my vote is to just keep it around...

HelpSource/Overviews/Operators.schelp
```diff
@@ -303,7 +303,8 @@ method:: /
 Division.
 
 method:: %
-Floating point modulo.
+Modulo.
+NOTE:: Uses floored division. Integer modulo used to exhibit unexpected behavior for negative modulus; see link::Classes/Integer#-mod_old::.::
```
dyfer, line 307: How about we explicitly add the version name here?

Suggested:
`Integer modulo used to exhibit unexpected behavior for negative modulus before SC 3.14; see link::Classes/Integer#-mod_old::.::

passyur:
I'm in support of `modSeaside` too, lol. I think I'll leave the sclang code to refer to `mod_old` and the primitive to be `_ModOldInt` just to be clear that it's from a prior implementation. All classlib and doc stuff will refer to `modSeaside`. It'll be some hours until I'm able to commit it—like 11 PM UTC.

dyfer:
At the dev meeting we agreed with renaming the classlib method to `modSeaside`.
However, we also thought it would be better to also use `mod_seaside` `_modSeasideInt` in the cpp to make the naming consistent. Then add a comment that it comes from an earlier implementation, as we'd need a comment about it either way.

dyfer:
@JordanHendersonMusic should we merge this?

JordanHendersonMusic (approved):

JordanHendersonMusic:
Squash and merge please :)

dyfer:
@JordanHendersonMusic in this case the commits are logically split between various areas of the project. Are you sure we want this squashed?

telephon:
I wonder: now that everything is fixed, shouldn't modSeaside have a uniform behaviour across all number types? When someone needs it for backward compatibility, it will be because of unexpected behaviour, and then the choice is clear.

Trying hard, I can imagine a case where we have mixes of ints and floats, and the different behaviour is exploited. But this seems to me less important than a clarity of which version does what. Also easier to explain that way.

Now that we have it also for float (we've never had it before), and AbstractFunction, we have to make sure it "works" also for UGens. But otherwise it would also be completely fine to just have modSeaside for integers only?

passyur:
> shouldn't modSeaside have a uniform behaviour across all number classes

I was still under the impression that `modSeaside` was to be a backwards compatibility function à la `hammingWindow_old` (so, built-in), and so would befit the quirks of the previous `mod`. One such quirk is that `mod` exhibits different behavior for `Integer`s than it does for `Float`s and associated floating-point classes. That's a bit confusing*, but that's also how it's been operating. And that the edge cases are slim shouldn't necessarily imply that they don't warrant support, right? (That's partly the motivation for the backwards compatibility, now that I think about it...) 

> do you think this is possible?

I do see a value in having a function that behaves like a different definition of modulo for all number classes—something discussed earlier in this thread. But I agree with where we left off on that—that that's a different feature. I see this as a good opportunity for a Quark in the meantime and, in the future, as a PR to add support for different definitions of modulo.

--- 

*Also: If there're hitherto any unclarities about the behavior as explained in the docs attached to this PR, please let me know so that they can be more precise & helpful. Thanks!

telephon:
There are some pending requests independent of the last issue we have been discussing. In the meantime, I'll think about what's best.

dyfer:
At the dev meeting we decided to accept this, adding suggestions by @telephon

dietcv:
I get a warning from sc_mod when building a plugin release because of the implicit conversion, can this be changed? 
```
[{
	"resource": "/c:/Users/jlkle/Documents/C++Projects/supercollider/include/plugin_interface/SC_InlineBinaryOp.h",
	"owner": "msCompile",
	"code": "C4244",
	"severity": 4,
	"message": "\"return\": Konvertierung von \"float64\" in \"int\", möglicher Datenverlust [C:\\Users\\jlkle\\D",
	"source": "cpp",
	"startLineNumber": 345,
	"startColumn": 15,
	"endLineNumber": 345,
	"endColumn": 15
}]
```

Issue #6696, opened 2025-04-02
Builds fail due to implicit cmake4 upgrade


capital-G:
## Environment

* SuperCollider version: dev
* Operating system: linux/windows
* Other details (Qt version, audio driver, etc.): see CI

## Steps to reproduce

Currently the CI fails due to cmake compatibility, see https://github.com/supercollider/supercollider/actions/runs/14215760688/job/39835816096#step:8:49

```shell
CMake Error at external_libraries/hidapi/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

I don't know why this happens though b/c ubuntu 22.04 should still use [`3.22.1-1ubuntu1`](https://packages.ubuntu.com/jammy/cmake)  and 24.04 should use [`3.28.3-1build7`](https://packages.ubuntu.com/noble/cmake) which we can verify in the CI - see https://github.com/supercollider/supercollider/actions/runs/14215760688/job/39835816096#step:5:230 - so I don't understand why we are affected by a CMake 4.0 removal, see https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features

> Compatibility with versions of CMake older than 3.5 has been removed. Calls to [cmake_minimum_required()](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required) or [cmake_policy()](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy) that set the policy version to an older value now issue an error. Note that calls to those commands can still support older versions of CMake by using their VERSION arguments' <min>...<max> syntax. This requires only the <min> version of CMake, but when running a newer version, sets policies up to the <max> version.

I don't really understand why the CI runs into problems here, but it is best to simply update

https://github.com/supercollider/hidapi/blob/976cb116102767dd72d58c191dd79236fca238d7/CMakeLists.txt#L1

to 3.5 and update the dependency upstream in the SC repo

edit: this also affects portaudio, see https://github.com/PortAudio/portaudio/issues/1015

dyfer:
[...]

PR #4925, opened 2020-05-12
Update PortAudio

dyfer:
## Purpose and Motivation

This PR removes our fork of PortAudio and adds official PortAudio repository from ~~Assembla~~ GitHub. Our CMakeLists is modified to accommodate the change.

## Types of changes

- Bug fix (?)
- New feature (?)
- ~~Breaking change~~ hopefully not

## To-do list

- [x] Code is tested
- [x] Documentation is updated
  - will update once we agree on other aspects of this PR
- [x] This PR is ready for review
- [ ] commit history is cleaned up

PR #6700, opened 2025-04-04
Patch yaml-cpp for cmake 4

dyfer:
## Purpose and Motivation

~~Update yaml-cpp to version 0.8.0 (tag `0.8.0`) _using the [upstream yaml-cpp repo](https://github.com/jbeder/yaml-cpp)_.~~

~~The underlying reason for this is not an update in functionality, but in CMake. This PR is a partial fix for #6696 (fixes building on Linux).~~

~~Note: **This PR changes the submodule URL to use the upstream version instead of the supercollider-hosted fork**. I don't see the reason for us to host [our own fork](https://github.com/supercollider/yaml-cpp), as it doesn't have any custom changes on it. I propose that we use the upstream repository as the submodule and that we archive our fork if this PR gets merged.~~

~~We will need to run **`git submodule sync`** in order to update our local repositories afterwards.~~

EDIT: This PR now only adds a cmake 4 workaround.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/external_libraries/CMakeLists.txt
+++ b/external_libraries/CMakeLists.txt
@@ -96,6 +96,7 @@ if(NOT SYSTEM_YAMLCPP)
   message(STATUS "Using bundled yaml-cpp")
 
   # yaml-cpp
+  set(CMAKE_POLICY_VERSION_MINIMUM 3.10) # workaround for cmake 4
   aux_source_directory(${CMAKE_SOURCE_DIR}/external_libraries/yaml-cpp/src yaml_src)
   CREATE_FINAL_FILE(${CMAKE_CURRENT_BINARY_DIR}/libyamlcpp.cpp ${yaml_src})
 


dyfer:
BTW macOS and Linux images were [reverted](https://github.com/actions/runner-images/pulls?q=is%3Apr+is%3Aclosed+cmake) to CMake 3.31.x. Still I think the yaml-cpp update is worth doing.

capital-G:
I think this is a good and can be merged - but before merging we should be clear if we should worry about windows not compiling? I can't figure out why it fails to compile though, it just seems to stop? Don't want to introduce additional errors in this currently broken state.

dyfer:
Originally I wrote:
Windows doesn't compile because of cmake4, with fix proposed in #6701. 

HOWEVER, after creating a test branch to include both changes, INDEED there is a problem building this on Windows. Investigating...

dyfer:
I think I found a solution, but it seems a little hacky (i.e. sounds like something that should be addressed / work oob upstream?)

It does seem to work properly though - here's a [run on a branch that includes this fix along with the cmake4 fix](https://github.com/dyfer/supercollider/actions/runs/14286361822) to fixes windows builds.

capital-G:
I don't know what is the best way to go on here - maybe @jrsurge can give some insights or advice here? :)

Also we should maybe re-consider switching the submodule URL - there was probably some (good?) reason to have it.
I don't have too much experience regarding best-practices for C++ projects. In case we need to patch/maintain yaml we can do this w/o any additional switch for the users (I can already see the issue regarding "my build fails b/c of yaml -> have you tried `git submodule sync`?"). We can't get rid of the repo anymore now anyway, so maybe just continue using it?

dyfer:
Arguably, maintaining yaml-cpp fork hasn't been much of an issue in the past.

I can't speak to what was the intention of the original fork, the last time we did this was for [PortAudio](https://github.com/supercollider/supercollider/pull/4925) - PortAudio fork was used previously I believe because upstream PA repository was on Assembla and/or some changes needed to be made to it.
We definitely have a mix of our own forks, upstream forks, and straight up copied files in [external_libraries](https://github.com/supercollider/supercollider/tree/develop/external_libraries). E.g. `nova-tt` and `nova-simd` point to upstream repos, despite forks being present in SC org (for whatever reason...).
Also, when we last updated boost, we removed any modifications of the upstream lib that were previously included, i.e. there's a previous effort to use external libraries in their original form. 

In a similar sense, I think using the upstream repo is the cleanest solution in general.

Ultimately, I don't feel strongly about this. We could update CMake file in our fork instead, or merge this (if it passes @jrsurge 's review). I'd think that ultimately using the usptream is a cleaner solution, but it doesn't seem to be a big issue either way.

external_libraries/CMakeLists.txt
```diff
@@ -105,8 +105,12 @@ if(NOT SYSTEM_YAMLCPP)
...
   set(YAMLCPP_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/external_libraries/yaml-cpp/include)
 
-  if(CMAKE_COMPILER_IS_GNUCXX)
-    target_compile_options(yaml PRIVATE -Wno-deprecated-declarations)
+  if(WIN32)
+    if(MSVC) # MSVC needs these exports even for the static library
+      target_compile_definitions(yaml PRIVATE yaml_cpp_EXPORTS YAML_CPP_DLL)
+    else() # for MinGW we need to set YAML_CPP_STATIC_DEFINE as public for some reason
+      target_compile_definitions(yaml PUBLIC YAML_CPP_STATIC_DEFINE)
+    endif()
```
jrsurge, line 113: I think we might be able to replace all of this stuff with `add_subdirectory(yaml-cpp)` - which will give us a library of `yaml-cpp::yaml-cpp` to link against where we need it.

The surrounding logic here appears to be trying to inject whatever boost library _we_ are using into yaml-cpp - but yaml-cpp have replaced their usage of boost, since 0.7.0 as of: https://github.com/jbeder/yaml-cpp/commit/24fa1b33805c9a91df0f32c46c28e314dd7ad96f - so we can probably simplify this quite a bit:

```cmake
if(NOT SYSTEM_YAMLCPP) # line 95 above!
    message(STATUS "Using bundled yaml-cpp")
    add_subdirectory(yaml-cpp)
endif()
```

and then updating everywhere that tries to link against the one we defined?

dyfer: Thanks for this feedback! I'll try this and will report back.

dyfer:
Thanks @jrsurge, `add_subdirectory()` seems much cleaner overall!

HOWEVER I now realized that the latest release of yaml-cpp (0.8.0) still lists [3.4 as the minimum CMake version](https://github.com/jbeder/yaml-cpp/blob/0.8.0/CMakeLists.txt#L2) and thus does not solve an issue with CMake 4. I'm sorry for missing this earlier.

One solution would be to use the tip of the master brach, which [does seem to fix this](https://github.com/jbeder/yaml-cpp/blob/master/CMakeLists.txt#L3).

Another solution would be to add `set(CMAKE_POLICY_VERSION_MINIMUM_ 3.10)` to override the version minimum before adding the subdirectory - this could be done with or without updating the yaml-cpp itself.

What do you think is the better path forward?

capital-G:
I think I'd favor to add `set(CMAKE_POLICY_VERSION_MINIMUM 3.10)` as we also allow to use system yaml which would always be a versioned release.

I think a good modus operand seems to be to stick to releases for dependencies where we also allow to link them via the system and fix the build process in our CMake file. This would also be consistent with our portaudio fix @https://github.com/supercollider/supercollider/pull/6701/files#diff-a6820d5aa6c0a83e347d2f23e993d8a1d23912d45feb0a326d6295546bc81671R18

dyfer:
I reverted this branch to just add `set(CMAKE_POLICY_VERSION_MINIMUM 3.10)` before adding yaml directory, keeping the current version/fork of yaml-cpp.

I think that once the next version of yaml-cpp is released (>0.8.0, which should include the fix for Cmake version), we could consider again using the upstream fork and cleaning up the CMake integration by using `add_subdirectory()`. For reference, here's the branch using 0.8.0 and `add_subdirectory()`: https://github.com/dyfer/supercollider/tree/update-yamlcpp-0.8.0-wip

capital-G (approved):

Issue #6102, opened 2023-10-03
Plugin interface changes


dyfer:
## Motivation

I'd like to create a meta-issue to keep track of the planned plugin interface changes. Since changing the plugin interface will break plugins compatibility, I'd like us to make as many changes as possible all at once, ideally before the next major version release.
Let's use this issue to only discuss the interaction between these changes and additions/omissions to the list, and keep feature-specific discussions in applicable Issues/PRs.

## List of plugin interface changes to consider
[ ] SCFFT_Allocator is not ABI compatible #4438
[ ] Improve `.midiratio` precision #5016 (I'm not sure if fixing this would break the plugin interface?)
[ ] Make plugin API extendable #5347 (also #5346 ?)
[ ] API extension: `SendReply` #5939
[ ] Set library paths for plugins on Windows #6040 (I'm not sure if fixing this would break the plugin interface?)
[ ] Extend Plugin Interface to allow FluCoMa on supernova https://github.com/flucoma/flucoma-sc/issues/90 (no SC Issue/PR atm)
[ ] Unify scsynth/supernova plugin interface #6101

Spacechild1:
I think this is an important topic! I would really like to participate and I already have some plans. Please ping me when you want to discuss this in a dev meeting!

Spacechild1:
Here's another one: https://github.com/supercollider/supercollider/issues/6145

Spacechild1:
Removed #5016 because it wouldn't break the plugin API or ABI. Added #6702 because it does break ABI.

dyfer:
I also added #6709

capital-G:
In the dev meeting https://github.com/supercollider/supercollider/wiki/Development-meetings:-minutes#2025-04-06 we decided to make 3.14 a release using the old plugin API - 3.15 will have a new version of the plugin API and therefore break the ABI, hopefully for one last time.

This means: 3.14 will work with all existing plugins which worked with 3.13 (except flucoma - see ). 3.15 will not be compatible with <=3.14 plugins, so they need to be re-compiled!

dyfer:
> Added [#6702](https://github.com/supercollider/supercollider/pull/6702) because it does break ABI.

So did I! I'll remove the duplicate.

PR #6702, opened 2025-04-06
Synth reblocking and upsampling

Spacechild1:
## Purpose and Motivation

This PR allows Synths to run at different block sizes or samplerates. This enables two important techniques that haven't been possible so far:
- tight feedback loops, down to a single sample
- upsampling + filtering to remove/reduce aliasing, e.g. for distortion, hard-syncing, osciallators and any other type of synthesis that is prone to aliasing

---

### For users

`Synth.new` (and similar methods) have two additional arguments:
- `blockSize`: the block size (`nil`: Server block size); can not be larger than the Server block size!
- `upsample`: the upsampling factor (`nil`: no upsampling); must be a power of two.
  (In the future we might also support downsampling, that's why the argument is a floating point number.)

Here are two practical examples:
```sc
// test 1 sample delay
(
SynthDef(\delay, { |out|
    var source, local;

    source = Decay.ar(Impulse.ar(2), 0.1) * WhiteNoise.ar(0.2);
    local = LocalIn.ar(2) + [source, 0]; // read feedback, add to source
	local = DelayC.ar(local, 0.5, MouseX.kr.pow(2) * 0.5); // delay sound

    // reverse channels to give ping pong effect, apply decay factor
	LocalOut.ar(local.reverse * MouseY.kr.pow(0.5).clip(0, 0.999));

	Out.ar(out, local);
}).add
)

Synth(\delay); // min. delay length: Server block size
Synth(\delay, blockSize: 16); // min. delay length: 16 samles
Synth(\delay, blockSize: 1); // min. delay length: 1 sample

// test upsampling + anti-aliasing
(
SynthDef(\hardsync, { |out|
	var cutoff = 18000;
	var sig = SyncSaw.ar(MouseY.kr(1, 1000, 2), MouseX.kr(100, 4000, 2));
	sig = LPF.ar(sig, cutoff);
	sig = LPF.ar(sig, cutoff);
	Out.ar(out, sig * 0.05 ! 2);
}).add;
)

// lots of aliasing
Synth(\hardsync); 
// watch aliasing disappear with increasing upsampling factors
Synth(\hardsync, upsample: 2.0);
Synth(\hardsync, upsample: 4.0);
Synth(\hardsync, upsample: 8.0);
```

---

### For plugin authors

On the implementation side, this adds new `mBufRate` and `mFullRate` members to `Graph`. For "normal" Synths, these just point to the global `Rate` members in `World`. If the Synth is reblocking and/or upsampling, it has its own `Rate` instances.

UGens "just work" as long as they consistently use the `SAMPLERATE`, `BUFLENGTH`, `FULLRATE`, `FULLBUFLENGTH` etc. macros in `SC_Unit.h` resp. the corresponding C++ methods in `SC_Plugin.hpp`, because they now refer to the Graph's rates instead of the global World rates.

The only UGens that need to aware of reblocking and upsampling are I/O UGens because they are responsible for reblocking/resampling when reading from / writing to Busses. (Actually, this applies to all UGens that use Busses, but there are very few third-party plugins that do, as far as I am aware.)

---

**NOTE**: this feature requires a break in the plugin API. I have bumped the plugin API version in this PR to make sure that existing third-party plugins are not loaded, as they would not behave correctly in reblocked/upsampled Synths. In the end, this has to be coordinated with other PRs that would also require a plugin API break. See also this meta-issue: https://github.com/supercollider/supercollider/issues/6102

---

### Testing

It would be great if you could test this PR by running your existing SynthDefs at different combinations of block sizes and upsampling factors and check if they still work/sound as expected.

Here are two important guide lines to avoid false positives:

- the Synth sounds different at lower block sizes -> please check that this isn't also the case when setting the global Server block size to the same value (ideally with the official SC build)
- the Synth sounds different when upsampled -> please check that this isn't also the case when running the Server at the corresponding global sample rate  (ideally with the official SC build)

UGens might behave weirdly in a reblocked/upsampled Synth, but they should never crash the Server.

## Types of changes

- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Control/Node.sc
+++ b/SCClassLibrary/Common/Control/Node.sc
@@ -457,7 +457,7 @@ Synth : Node {
 		if (blockSize.notNil or: { upsample.notNil }) {
 			server.sendMsg(67, //"s_newEx"
 				defName, nodeID, addActionID, targetNode.nodeID,
-				blockSize, upsample, *(args.asOSCArgArray)
+				blockSize ? 0, upsample ? 0.0, *(args.asOSCArgArray)
 			);
 		} {
 			server.sendMsg(9, //"s_new"
@@ -503,7 +503,7 @@ Synth : Node {
 		target = (target ? server.defaultGroup);
 		group = if(addActionID < 2) { target } { target.group };
 		if (blockSize.notNil or: { upsample.notNil }) {
-			^[67, defName, nodeID, addActionID, target.nodeID, blockSize, upsample ] ++
+			^[67, defName, nodeID, addActionID, target.nodeID, blockSize ? 0, upsample ? 0.0 ] ++
 			args.asOSCArgArray // "/s_newEx"
 		} {
 			^[9, defName, nodeID, addActionID, target.nodeID] ++
--- a/SCClassLibrary/Common/Control/Node.sc
+++ b/SCClassLibrary/Common/Control/Node.sc
@@ -425,46 +425,24 @@ Synth : Node {
 	var <>defName;
 
 	/** immediately sends **/
-	*new { arg defName, args, target, addAction=\addToHead;
-		var synth, server, addActionID;
+	*new { arg defName, args, target, addAction=\addToHead, blockSize, upsample;
+		var synth, server;
 		target = target.asTarget;
 		server = target.server;
-		addActionID = addActions[addAction];
 		synth = this.basicNew(defName, server);
-
-		synth.group = if(addActionID < 2) { target } { target.group };
-		server.sendMsg(9, //"s_new"
-			defName, synth.nodeID, addActionID, target.nodeID,
-			*(args.asOSCArgArray)
-		);
+		synth.prSend(target, addAction, blockSize, upsample, args);
 		^synth
 	}
 
-	*newPaused { arg defName, args, target, addAction=\addToHead;
-		var synth, server, addActionID;
+	*newPaused { arg defName, args, target, addAction=\addToHead, blockSize, upsample;
+		var synth, server;
 		target = target.asTarget;
 		server = target.server;
-		addActionID = addActions[addAction];
 		synth = this.basicNew(defName, server);
-		synth.group = if(addActionID < 2) { target } { target.group };
-		server.sendBundle(nil, [9, defName, synth.nodeID, addActionID, target.nodeID] ++
-			args.asOSCArgArray, [12, synth.nodeID, 0]); // "s_new" + "/n_run"
-		^synth
-	}
-
-	*replace { arg nodeToReplace, defName, args, sameID=false;
-		var synth, server, newNodeID;
-		if(sameID) { newNodeID = nodeToReplace.nodeID };
-		server = nodeToReplace.server;
-		synth = this.basicNew(defName, server, newNodeID);
-
-		server.sendMsg(9, //"s_new"
-			defName,
-			synth.nodeID,
-			4, // addReplace
-			nodeToReplace.nodeID,
-			*(args.asOSCArgArray)
-		);
+		server.sendBundle(nil,
+			synth.newMsg(target, args, addAction, blockSize, upsample),
+			[12, synth.nodeID, 0]
+		); // "s_new" + "/n_run"
 		^synth
 	}
 
@@ -473,68 +451,106 @@ Synth : Node {
 		^super.basicNew(server, nodeID).defName_(defName.asDefName)
 	}
 
-	newMsg { arg target, args, addAction = \addToHead;
+	prSend { arg targetNode, addAction, blockSize, upsample, args;
 		var addActionID = addActions[addAction];
-		target = target.asTarget;
-		group = if(addActionID < 2) { target } { target.group };
-		^[9, defName, nodeID, addActionID, target.nodeID] ++ args.asOSCArgArray //"/s_new"
+		group = if(addActionID < 2) { targetNode } { targetNode.group };
+		if (blockSize.notNil or: { upsample.notNil }) {
+			server.sendMsg(67, //"s_newEx"
+				defName, nodeID, addActionID, targetNode.nodeID,
+				blockSize, upsample, *(args.asOSCArgArray)
+			);
+		} {
+			server.sendMsg(9, //"s_new"
+				defName, nodeID, addActionID, targetNode.nodeID,
+				*(args.asOSCArgArray)
+			);
+		}
 	}
 
-	*after { arg aNode, defName, args;
-		^this.new(defName, args, aNode, \addAfter)
+	*after { arg aNode, defName, args, blockSize, upsample;
+		^this.new(defName, args, aNode, \addAfter, blockSize, upsample)
 	}
 
-	*before { arg aNode, defName, args;
-		^this.new(defName, args, aNode, \addBefore)
+	*before { arg aNode, defName, args, blockSize, upsample;
+		^this.new(defName, args, aNode, \addBefore, blockSize, upsample)
 	}
 
-	*head { arg aGroup, defName, args;
-		^this.new(defName, args, aGroup, \addToHead)
+	*head { arg aGroup, defName, args, blockSize, upsample;
+		^this.new(defName, args, aGroup, \addToHead, blockSize, upsample)
 	}
 
-	*tail { arg aGroup, defName, args;
-		^this.new(defName, args, aGroup, \addToTail)
+	*tail { arg aGroup, defName, args, blockSize, upsample;
+		^this.new(defName, args, aGroup, \addToTail, blockSize, upsample)
 	}
 
-	replace { arg defName, args, sameID;
-		^this.class.replace(this, defName, args, sameID)
+	*replace { arg nodeToReplace, defName, args, sameID=false, blockSize, upsample;
+		var synth, server, newNodeID;
+		if(sameID) { newNodeID = nodeToReplace.nodeID };
+		server = nodeToReplace.server;
+		synth = this.basicNew(defName, server, newNodeID);
+		synth.prSend(nodeToReplace, \addReplace, blockSize, upsample, args);
+		^synth
+	}
+
+	replace { arg defName, args, sameID, blockSize, upsample;
+		^this.class.replace(this, defName, args, sameID, blockSize, upsample)
 	}
 
 	// for bundling
-	addToHeadMsg { arg aGroup, args;
+	newMsg { arg target, args, addAction = \addToHead, blockSize, upsample;
+		var addActionID = addActions[addAction];
 		// if aGroup is nil set to default group of server specified when basicNew was called
-		group = (aGroup ? server.defaultGroup);
-		^[9, defName, nodeID, 0, group.nodeID] ++ args.asOSCArgArray	// "/s_new"
+		target = (target ? server.defaultGroup);
+		group = if(addActionID < 2) { target } { target.group };
+		if (blockSize.notNil or: { upsample.notNil }) {
+			^[67, defName, nodeID, addActionID, target.nodeID, blockSize, upsample ] ++
+			args.asOSCArgArray // "/s_newEx"
+		} {
+			^[9, defName, nodeID, addActionID, target.nodeID] ++
+			args.asOSCArgArray //"/s_new"
+		}
 	}
 
-	addToTailMsg { arg aGroup, args;
-		// if aGroup is nil set to default group of server specified when basicNew was called
-		group = (aGroup ? server.defaultGroup);
-		^[9, defName, nodeID, 1, group.nodeID] ++ args.asOSCArgArray // "/s_new"
+	addToHeadMsg { arg aGroup, args, blockSize, upsample;
+		^this.newMsg(aGroup, args, \addToHead, blockSize, upsample);
 	}
 
-	addAfterMsg { arg aNode, args;
-		group = aNode.group;
-		^[9, defName, nodeID, 3, aNode.nodeID] ++ args.asOSCArgArray // "/s_new"
+	addToTailMsg { arg aGroup, args, blockSize, upsample;
+		^this.newMsg(aGroup, args, \addToTail, blockSize, upsample);
 	}
 
-	addBeforeMsg { arg aNode, args;
-		group = aNode.group;
-		^[9, defName, nodeID, 2, aNode.nodeID] ++ args.asOSCArgArray // "/s_new"
+	addAfterMsg { arg aNode, args, blockSize, upsample;
+		^this.newMsg(aNode, args, \addAfter, blockSize, upsample);
 	}
 
-	addReplaceMsg { arg nodeToReplace, args;
-		group = nodeToReplace.group;
-		^[9, defName, nodeID, 4, nodeToReplace.nodeID] ++ args.asOSCArgArray // "/s_new"
+	addBeforeMsg { arg aNode, args, blockSize, upsample;
+		^this.newMsg(aNode, args, \addBefore, blockSize, upsample);
+	}
+
+	addReplaceMsg { arg nodeToReplace, args, sameID=false, blockSize, upsample;
+		var msg = this.newMsg(nodeToReplace, args, \addReplace,
+			blockSize, upsample);
+		if (sameID) {
+			// bash node ID
+			msg[2] = nodeToReplace.nodeID;
+		};
+		^msg;
 	}
 
 	// nodeID -1
-	*grain { arg defName, args, target, addAction=\addToHead;
-		var server;
+	*grain { arg defName, args, target, addAction=\addToHead, blockSize, upsample;
+		var server, addActionID;
 		target = target.asTarget;
 		server = target.server;
-		server.sendMsg(9, defName.asDefName, -1, addActions[addAction], target.nodeID,
-			*(args.asOSCArgArray)); //"/s_new"
+		addActionID = addActions[addAction];
+		if (blockSize.notNil or: { upsample.notNil }) {
+			server.sendMsg(67, defName.asDefName, -1, addActionID,
+				target.nodeID, blockSize ? 0, upsample ? 0.0,
+				*(args.asOSCArgArray)) //"/s_newEx"
+		} {
+			server.sendMsg(9, defName.asDefName, -1, addActionID,
+				target.nodeID, *(args.asOSCArgArray)); //"/s_new"
+		};
 		^nil
 	}
 
--- a/include/plugin_interface/SC_Graph.h
+++ b/include/plugin_interface/SC_Graph.h
@@ -24,6 +24,10 @@
 #include "SC_Rate.h"
 #include "SC_SndBuf.h"
 
+enum { kGraph_Reblock = 0x1, kGraph_Resample = 0x2 };
+
+#define kGraph_ReblockOrResample (kGraph_Reblock | kGraph_Resample)
+
 /*
  changes to this struct likely also mean that a change is needed for
     static const int sc_api_version = x;
@@ -31,6 +35,11 @@
  */
 struct Graph {
     Node mNode;
+
+    uint16 mNumTicks;
+    uint16 mTickCounter;
+
+    uint32 mFlags;
 
     uint32 mNumWires;
     struct Wire* mWire;
@@ -60,6 +69,9 @@ struct Graph {
     SndBuf* mLocalSndBufs;
     int32 localBufNum;
     int32 localMaxBufNum;
+
+    Rate* mFullRate;
+    Rate* mBufRate;
 
     void* mPrivate;
 };
--- a/include/plugin_interface/SC_Graph.h
+++ b/include/plugin_interface/SC_Graph.h
@@ -41,10 +41,13 @@ struct Graph {
 
     uint32 mFlags;
 
-    uint32 mNumWires;
-    struct Wire* mWire;
+    Rate* mFullRate;
+    Rate* mBufRate;
 
+    uint32 mNumWires;
     uint32 mNumControls;
+
+    struct Wire* mWire;
     float* mControls;
     float** mMapControls;
     int32* mAudioBusOffsets;
@@ -53,25 +56,22 @@ struct Graph {
     int32* mControlRates;
 
     uint32 mNumUnits;
-    struct Unit** mUnits;
-
     uint32 mNumCalcUnits;
+
+    struct Unit** mUnits;
     struct Unit** mCalcUnits; // excludes i-rate units.
 
     int32 mSampleOffset;
+    float mSubsampleOffset;
+
     struct RGen* mRGen;
 
     struct Unit* mLocalAudioBusUnit;
     struct Unit* mLocalControlBusUnit;
 
-    float mSubsampleOffset;
-
     SndBuf* mLocalSndBufs;
     int32 localBufNum;
     int32 localMaxBufNum;
-
-    Rate* mFullRate;
-    Rate* mBufRate;
 
     void* mPrivate;
 };
--- a/include/plugin_interface/SC_Unit.h
+++ b/include/plugin_interface/SC_Unit.h
@@ -120,9 +120,10 @@ template <typename ToType, typename Value>
 #define BUFLENGTH (unit->mBufLength)
 #define BUFRATE (unit->mRate->mBufRate)
 #define BUFDUR (unit->mRate->mBufDuration)
-#define FULLRATE (unit->mWorld->mFullRate.mSampleRate)
-#define FULLBUFLENGTH (unit->mWorld->mFullRate.mBufLength)
-#define FULLSAMPLEDUR (unit->mWorld->mFullRate.mSampleDur)
+#define FULLRATE (unit->mParent->mFullRate->mSampleRate)
+#define FULLSAMPLEDUR (unit->mParent->mFullRate->mSampleDur)
+#define FULLBUFLENGTH (unit->mParent->mFullRate->mBufLength)
+#define REBLOCK_OR_RESAMPLE (unit->mParent->mFlags & kGraph_ReblockOrResample)
 
 #ifdef SUPERNOVA
 
--- a/include/server/ErrorMessage.hpp
+++ b/include/server/ErrorMessage.hpp
@@ -71,7 +71,9 @@ std::string apiVersionMismatch(std::string const& utf8Filename, int const expect
     assert(expectedVersion != actualVersion);
 
     // both 1 and 2 were introduced in 3.6
-    static map<int, string> const scVersionForAPIVersion = { { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" } };
+    static map<int, string> const scVersionForAPIVersion = {
+        { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" }, { 4, "3.14" }
+    };
 
     stringstream message;
     message << "ERROR: API version mismatch: " << utf8Filename << "\n";
--- a/include/server/SC_OSC_Commands.h
+++ b/include/server/SC_OSC_Commands.h
@@ -119,5 +119,8 @@ enum {
 
     cmd_b_setSampleRate = 66,
 
-    NUMBER_OF_COMMANDS = 67
+    cmd_s_newEx = 67,
+    cmd_s_newargsEx = 68,
+
+    NUMBER_OF_COMMANDS = 69
 };
--- a/server/plugins/DelayUGens.cpp
+++ b/server/plugins/DelayUGens.cpp
@@ -439,7 +439,7 @@ void SampleDur_Ctor(Unit* unit, int inNumSamples) { ZOUT0(0) = FULLSAMPLEDUR; }
 
 void ControlDur_Ctor(Unit* unit, int inNumSamples) { ZOUT0(0) = BUFDUR; }
 
-void RadiansPerSample_Ctor(Unit* unit, int inNumSamples) { ZOUT0(0) = unit->mWorld->mFullRate.mRadiansPerSample; }
+void RadiansPerSample_Ctor(Unit* unit, int inNumSamples) { ZOUT0(0) = unit->mParent->mFullRate->mRadiansPerSample; }
 
 void BlockSize_Ctor(Unit* unit, int inNumSamples) { ZOUT0(0) = FULLBUFLENGTH; }
 
--- a/server/plugins/IOUGens.cpp
+++ b/server/plugins/IOUGens.cpp
@@ -52,6 +52,7 @@ struct AudioControl : Unit {
     float* m_prevVal;
     float* m_prevBus;
     bool* m_busUsedInPrevCycle;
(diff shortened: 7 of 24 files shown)


capital-G:
Thanks so much for this truly awesome feature, can't wait to give this some extensive testing! <3

As discussed in the dev meeting, 3.14 will be a release with the old plugin API still in place, so this is blocked until 3.14 has been released. Hopefully this PR will give us the drive to finish both releases soon :)

Spacechild1:
> If I let this run for a while, after 512 seconds (when sum is 512) the blocksize 1 synth will stop accumulating.. the other one will go forever. If I change the rate to anything above 1.5 it will continue.

That's a limitation of single-precision floating point numbers. There are only 23 bits for the fractional part, so the max. positive number that can be expressed with full precision is 2^23 (= 8388608). Once you exceed this number, the lower bits are simply cut off. 512 seconds @ 48 kHz correspond to 24,576,000 samples. At this stage, the number cannot be incremented further by a single sample. However, larger increments still work, that's why you found that the output does not stop with larger `rate` arguments. (It will stop eventually, though.) You have the same problem with `BufRead`, see the documentation of the `phase` argument.

bgola:
ah yes, makes sense! 
thanks for the explanation and of course it also happens in current SC if you are running the server at blockSize 1

another thing I noticed is that LocalIn.ar is ignoring the default value?

```
{ LocalIn.ar(1, 300).poll }.play
```

in SC 3.13 it polls 300 but in this current branch it polls 0.

But checking a bit further it seems like it is also an issue in the `develop` branch.

SCClassLibrary/Common/Control/Node.sc
```diff
@@ -473,68 +451,107 @@ Synth : Node {
...
 
-	*head { arg aGroup, defName, args;
+	*head { arg aGroup, defName, args, blockSize, upsample;
 		^this.new(defName, args, aGroup, \addToHead)
 	}
 
-	*tail { arg aGroup, defName, args;
+	*tail { arg aGroup, defName, args, blockSize, upsample;
 		^this.new(defName, args, aGroup, \addToTail)
 	}
```
mccrmck, line 484: These should pass `blockSize` and `upsample` to `^this.new`, no?

Spacechild1: Yes, they should 🤦‍♂️

mccrmck: Hehe, at least it's an easy fix! ;)

mccrmck:
Firstly, fantastic work @Spacechild1! I haven't had a chance to test it myself yet, but it's very impressive work! :)

I wonder if it's worth discussing the choice to pass the `upsample` argument as a float (I see your comment to support downsampling in the future). Would users then choose a ratio for the downsampling? Like `upsample: 0.653`, or `upsample: 12000/s.sampleRate`? I feel like the upsampling ratio makes more sense as an integer (especially if it must be a power of 2) and I feel like it would be more intuitive to choose an integer for the (lower) sampleRate than a float. What do others think about this?

Spacechild1:
> Would users then choose a ratio for the downsampling?

Yes, e.g. 0.5 for 2x downsampling or 0.25 for 4x downsampling. This is how you specificy up- and downsampling factors in Pd and I quite like it.

> upsample: 0.653
> upsample: 12000/s.sampleRate

Note that the downsampling factor also has to be a (negative) power of two. The idea of an upsampling/downsampling *factor* is that users don't need to care about the actual sample rate. 

> and I feel like it would be more intuitive to choose an integer for the (lower) sampleRate than a float.

Yes, the alternative would be to provide the absolute sample rate, but personally I prefer upsampling/downsampling factors because that's how this feature is used in practice. (E.g. "I want 4x upsampling for this distortion effect.")

mccrmck:
> Note that the downsampling factor also has to be a (negative) power of two

Ah, okay - if the resample factor has to be a power of two and not an arbitrary float, then that makes sense. Perhaps a more descriptive argument name for the `upsample` argument to `resampleFactor` (or something else)?

(nitpicks, I know...)

Spacechild1:
> Perhaps a more descriptive argument name for the upsample argument to resampleFactor (or something else)?

I also thought of naming the argument `resample`. It would definitely make more sense in case we also support downsampling. On the other hand, upsampling would certainly be the more common case - by far. `resampleFactor` would be very clear, but maybe a bit clunky. I don't have any strong preferences, though.

mccrmck:
> `resampleFactor` would be even clearer, but maybe a bit clunky.
Yeah, agreed - naming things is hard!

I also agree that upsampling will be the more common case - but if the downsampling implementation is likely, it might be worthwhile to consider an argument name that doesn't have to change.

Looking forward to testing this soon - EXCITING!

telephon (commented):
One possible simplification

SCClassLibrary/Common/Control/Node.sc
```diff
@@ -473,68 +451,107 @@ Synth : Node {
...
+	prSend { arg targetNode, addAction, blockSize, upsample, args;
 		var addActionID = addActions[addAction];
-		target = target.asTarget;
-		group = if(addActionID < 2) { target } { target.group };
-		^[9, defName, nodeID, addActionID, target.nodeID] ++ args.asOSCArgArray //"/s_new"
+		group = if(addActionID < 2) { targetNode } { targetNode.group };
+		if (blockSize.notNil or: { upsample.notNil }) {
+			server.sendMsg(66, //"s_newEx"
+				defName, nodeID, addActionID, targetNode.nodeID,
+				blockSize ? 0, upsample ? 0.0, *(args.asOSCArgArray)
```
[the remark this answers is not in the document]
Spacechild1: Thanks, done!

Spacechild1:
> This approach allows only for the entire synth to be resampled, it wouldn't ever be possible to do resample juts a section of the graph?

I actually thought about that. This would be very cool, but it would be pretty complicated to implement. You would not only have to manage resampling and reblocking in the I/O UGens but also between UGen inputs and outputs "at the border". I guess this could be done by secretely inserting special UGens that do the conversions. You would also have to extend the SynthDef format. Reblocking/resampling only a section of a graph could be very interesting because it would theoretically enable Pd-style FFT via reblocking and overlap. The question is whether it's worth the complexity. But someone can give it a try :) I certainly won't have the time and energy to do it.

> to be ran at a variety of hardware blocksizes (not a part of this PR, but connected) and then a variety of synth–blocksizes and synth–upsample factor.

Yes, we should definitely test our UGens at different blocksizes, mostly to test the different vectorized and unvectorized versions. However, I think it should be enough to do this either on the Server or Graph level.

I'm not sure why testing with different sample rates would be necessary. Can you imagine any case where a UGen would work at 48000 kHz but not at 96000 kHz?

The only UGens that do care about reblocking/resampling are the I/O UGens; these should indeed be tested with different blocksize/resample combinations.

JordanHendersonMusic:
> I'm not sure why testing with different sample rates would be necessary. Can you imagine any case where a UGen would work at 48000 kHz but not at 96000 kHz?

Incorrect timing calculations, e.g., slopes. I believe the unit tests only ever run at one sample rate. It would be good to test that nothing unusual changes. I was thinking it would be good to test this pr works by comparing against what we'd expect to happen, but actually if we've never tested these ugens at different sample rates (particularly unusual ones like 22100) then they might not be a reliable 'source of truth'.

> However, I think it should be enough to do this either on the Server or Graph level.

Agreed. Graph would be my preference.

> The only UGens that do care about reblocking/resampling are the I/O UGens; these should indeed be tested with different blocksize/resample combinations.

Do you think that should be a part of this PR then?

Spacechild1:
> Incorrect timing calculations, e.g., slopes. I believe the unit tests only ever run at one sample rate.

But how would this happen in practice? UGens just get the sample rate from the `World` struct (or in this case: from the `Graph` struct) and there is only ever a single process function for all sample rates. As long as UGens don't use *hardcoded* sample rate values, I don't see any potential pitfalls.*) 

*) I just grepped for "44100" and "48000" and I have indeed found a few UGens (`BeatTrack`, `KeyTrack` and `MFCC`) that apparently only work with a limited set of sample rates. It would be great if these UGens would detect unsupported sample rates and print a warning.

> Do you think that should be a part of this PR then?

If someone could contribute these tests I would be very grateful :)

cdbzb:
this def:
```
SynthDef(\NC_harpsichord2, {
	arg
	//Standard Values
	out = 0, amp = 0.1, freq = 440, pan = 0, rel = 5,
	//Pluck arguments (blend goes from 0 to 1)
	trig = 1, coef = 0.1, blend = 0.6;

	var exciter, root, octave, snd;

        // You can use white noise here, but Pink is more realistic
	exciter = PinkNoise.ar(amp);

	// Fundamental
        root = Pluck.ar(
	        in: exciter,
	        trig: trig,
	        maxdelaytime: 1/freq,
	        delaytime: 1/freq,
	        decaytime: rel,
	        coef: coef,
	        mul: blend
        );

	// Octave Above
	octave = Pluck.ar(
	        in: exciter,
	        trig: trig,
	        maxdelaytime: 1/(2 * freq),
	        delaytime: 1/(2 * freq),
	        decaytime: rel,
	        coef: coef,
	        mul: (1 - blend)
        );

	// Output Stuff
	snd = Mix.ar(root + octave);
	snd = Limiter.ar(snd);

    DetectSilence.ar(in: snd, doneAction: 2);

	Out.ar(out, Pan2.ar(snd, pan));
},
metadata: (
	credit: "Zé Craum",
	category: \keyboards,
	tags: [\pitched]
	)
).add;
```
sounds different (or won't sound) depending on setttings

Synth(\NC_harpsichord2, upsample:1) //base case
Synth(\NC_harpsichord2, upsample:4) //brighter
Synth(\NC_harpsichord2, upsample:32) // chatters
Synth(\NC_harpsichord2, blockSize:4, upsample:8) //silent

Spacechild1:
@cdbzb Thanks for testing! I definitely hear the difference. However, I just compared `upsample: 2` with running the Server at 96 kHz instead of 48 kHz (with `upsample: 1`) and I hear the same result! This would mean that the sound of your SynthDef generally depends on the sample rate. Must be something in `Pluck`. Can you confirm this?

(I will add a note to the PR description.)

muellmusik:
> Must be something in Pluck. Can you confirm this?

The One Pole filter? I can check the source tomorrow but as you just provide the coefficient it looks like it may not compensate?

Spacechild1:
> but as you just provide the coefficient it looks like it may not compensate?

Good catch! @cdbzb So your Synth is indeed sample rate dependent.

cdbzb:
here's one: `SynthDef(\foo, { var sig = Impulse.ar(4); sig = FreeVerb.ar(sig, 1, 1) ; Out.ar(0, sig) }).add`

reverb length seems to vary with `upsample` arg

cdbzb:
`	SynthDef(\foo, { Impulse.ar(4) => CombN.ar(_, 1, \del.kr(0.1), 5) => Out.ar(0, _)}).add;  `

with upsample > 1 I hear only the first impulse

Spacechild1:
> not hearing any difference between sampleRates when upsample is 1... 

Are you sure? I hear a clear difference between 48 kHz, 96 kHz and 196 kHz and it sounds the same as the difference between 1x, 2x and 4x upsampling (at 48 kHz).

I had a look at the code of `FreeVerb` and it doesn't seem to consider the sample rate. `GVerb` *does* take the sample rate into account, but it still sounds considerably different at various sample rates (Server or Graph).

AFAICT the problem lies in the UGen code, not in the Graph upsampling. I find it pretty shocking that they sound completely different at 48 kHz and 96 kHz!

Spacechild1:
> [EDIT] Checking with just Impulse that is one problem here SynthDef(\foo, { Out.ar(0,  Impulse.ar(4) }).add

The problem is that downsampling is done by dropping samples, so depending on the phase and frequency some or all impulses will be dropped:

```
SynthDef(\foo, { |rate=125| Out.ar(0, Impulse.ar(rate)) }).add

a = Synth(\foo, upsample: 2);

a.set(\rate, 124);
a.set(\rate, 123);
a.set(\rate, 126);
```

This means you cannot safely output an Impulse train from an upsampled Synth *without filtering*. (In practice, you should always add an anti-aliasing filter before writing to an audio bus in upsampled Synths.)

The following does produce the same sound for all upsampling factors:
```
(
SynthDef(\foo, {
	var cutoff = 18000;
	var sig = CombN.ar(Impulse.ar(\rate.kr(4)), 1, \del.kr(0.1), 5);
	sig = LPF.ar(sig, cutoff);
	sig = LPF.ar(sig, cutoff);
	Out.ar(0, sig);
}).add;
)
```

You will notice that the sound gets quieter with increasing upsampling factors. This is expected because a single impulse with the same amplitude loses energy at increasing sample rates.

---

Note that this problem with `Impulse` is only caused by downsampling. Inside the upsampled Graph the output of `Impulse` is perfectly fine. You can verify this by writing the output into a Buffer:

```
b = Buffer.alloc(s, 64);

SynthDef(\foo, { RecordBuf.ar(Impulse.ar(s.sampleRate / 4), b, loop: 0.0) }).add

// pick one
Synth(\foo); // -> 1 impulse every 4 samlpes
Synth(\foo, upsample: 2); // -> 1 impulse every 8 samples
Synth(\foo, upsample: 4); // -> 1 impulse every 16 samples

// then plot
b.loadToFloatArray(action: { |array| array.clump(4).do(_.postln) });
```

Spacechild1:
Thanks for testing BTW! These are all pitfalls that could be mentioned in a help document. Let's collect them in this PR and then write it up :)

cdbzb:
> Are you sure? I hear a clear difference between 48 kHz, 96 kHz and 196 kHz and it sounds the same as the difference between 1x, 2x and 4x upsampling (at 48 kHz).

no I don't hear that here on Mac using built-in audio - all samplerates sound the same, all upsample settings different...

Spacechild1:
> no I don't hear that here on Mac using built-in audio - all samplerates sound the same, all upsample settings different...

Are you 100% sure? Please check the console and see if the Server is actually running at the specified sample rate and not at the default sample rate. Look out for the following line:

> SC_AudioDriver: sample rate = 192000.000000, driver's block size = 64

I have a hard time believing that `FreeVerb` really sounds the same for you with all Server sample rates when it clearly doesn't on my system... And since the code doesn't really take the sample rate into account, I would *expect* it to sound different.

JordanHendersonMusic:
It might be useful to create an issue regarding ugens not respecting sample rate as this topic isn't strictly related to this pr?

cdbzb:
> Are you 100% sure? Please check the console

yes the console *does* show the requested samplerates, and I *do * hear differences in Pluck but not in FreeVerb...

Spacechild1:
> It might be useful to create an issue regarding ugens not respecting sample rate as this topic isn't strictly related to this pr?

Yes, UGens not respecting the Server sample rate is indeed not the topic of this PR and someone might want to open a dedicated issue about it. (We might not be able to change the behavior of these UGens, but the dependency on the sample rate should at least be documented.)

---

In this case, @cdbzb claimed that `FreeVerb` sounds the same with different Server sample rates and that it only sounds different in an upsampled Synth. That would indeed point to an issue with upsampling. However, I cannot reproduce this. Here's my SynthDef:
```
(
SynthDef(\foo, {
	var sig = Impulse.ar(1);
	sig = FreeVerb.ar(sig.dup, 1, 1);
	Out.ar(0, sig)
}).add
)
```
And here are three sound files: [FreeVerb test.zip](https://github.com/user-attachments/files/19721972/FreeVerb.test.zip)

Every file contains three short excerpts of the SynthDef above with increasing upsampling factors (1x, 2x and 4x). The recording has been done with three different Server sample rates (48k Hz, 96 kHz and 192 kHz).

First, contrary to @cdbzb's claim, all three files sound completely different.

Second, it also demonstrates that the upsampled examples match the equivalent global Server sample rate:

| SR | upsample | upsample  | upsample  |
| --- | --- |--- |  --- |
| 48 kHz | 1x  | 2x | 4x |
| 96 kHz | - | 1x  | 2x |
| 196 kHz | -  | - | 1x |

@cdbzb If you still cannot confirm these results on your system, can you do the same experiment and upload the files?

Spacechild1:
I should maybe stress that I am mostly worried about the I/O UGens because they are responsible for the upsampling and reblocking logic. I don't really expect "normal" UGens to misbehave (which, of course, doesn't mean that it can't happen :)

So it would be great if people could not only test individual Synths, but also combination of Synths that are connected through busses and or control mappings. Let's not forget about things like `OffsetOut`, `XOut`, `LagIn`, `TrigIn`, etc., and the various types of controls. I think I have tested all of these and didn't notice any issues so far, but I might have missed something.

Spacechild1:
@cdbzb Can you confirm my results in https://github.com/supercollider/supercollider/pull/6702#issuecomment-2799043902?

cdbzb:
> @cdbzb Can you confirm my results in [#6702 (comment)](https://github.com/supercollider/supercollider/pull/6702#issuecomment-2799043902)?

Yes I can confirm your reults - my previous impression was due to user error - sorry for the trouble!

Spacechild1:
@cdbzb No worries! And thanks for testing!

JordanHendersonMusic:
Do we think the Convolution tests are related to this change? It is odd they are both failing on both operating systems.

Spacechild1:
> Do we think the failing Convolution tests are related to this change?

Honestly, I don't know. With our test suite it's sometimes hard to tell which test failures are valid and which are false positives. This PR does not touch the Convolution UGens at all, so I have no idea why they would fail...

Spacechild1:
The Convolution2 test also failed in a completely unrelated PR: https://github.com/supercollider/supercollider/actions/runs/14552893526/job/40826393145

capital-G:
Since develop is now 3.15, what are the missing steps to merging this? (except the merge conflict)

Should we try to organize a review of some kind this in some way or should this be merged as-is?

mtmccrea:
Could the information "For plugin authors" in your description be added to the Writing Unit Generators help file?

Once merged, it would be nice to also have the various tutorials on writing ugens updated as well...
[For plugin authors](https://github.com/supercollider/supercollider/wiki/Plugins-development)

telephon:
> Finally, we need to bikeshed the name of the resampling argument :-D (The current name is `upsample`).

The two arguments (currently `blockSize` and `upsample`) should somehow be similar? Strictly speaking that would either be `blockSize` and `upsamplingFactor`, or `reblock` and `upsample`? (the latter pair sound like flags, but are nice and short). "Reblock" is a new terminology (?), but makes sense once you know it.


(Note: Internallly, it is sometimes called resample. )

mtmccrea:
I would be in favor of making it more accurate at the expense of longer arg names.

What about `reblockSize` and `resampleFactor`? (Or `upsampleFactor` if you're sure there wouldn't be a future use to downsample)

sadguitarius:
This is so cool @Spacechild1! Been wanting something like this for years but had no idea how to begin implementing it. Is your plan to have the resampling filters be implemented as UGens? I've been messing with some efficient polyphase IIR filters (mostly from fred harris' publications) and I'd love the ability to hack on these in the context of a SuperCollider plugin. Even better if the plugin interface could specify different sample rates for input and output! So I guess that would mean making a subclass of an I/O UGen and adding the filtering method to that, or am I misunderstanding the implementation?

Re: the naming scheme, I like @mtmccrea's `reblockSize` and `resampleFactor`. It matches up well with existing methods like `Server.blockSize` and would work well with the underscore setter method. (and BTW downsampling can be useful for metering expensive processes at a lower frame rate, for example)

Spacechild1:
> Is your plan to have the resampling filters be implemented as UGens?

That's the idea.

> So I guess that would mean making a subclass of an I/O UGen and adding the filtering method to that, or am I misunderstanding the implementation?

The filter would be a regular UGen that is added after the inputs resp. before the outputs. In my example in the PR description I just use two 2nd order Butterworth low-pass filters (`LPF`) as a crude anti-aliasing filter.

Note that I've picked zero-order hold as the upsampling method because it kind of works with most input (although it certainly distorts the input). Zero-padding (without filtering) can give really strange results with certain inputs, just think of a trigger signal.

This means that if you want to filter the input signal, you first have to zero pad it. We can already emulate this with `Impulse.ar`, but if someone wants to write a dedicated anti-aliasing filter, it would be nice to have *optional* zero-padding of the input signal.

Spacechild1:
> Re: the naming scheme, I like @mtmccrea's reblockSize and resampleFactor. It matches up well with existing methods like Server.blockSize and would work well with the underscore setter method. 

Sounds fine to me. If there is a broad concensus, I would go ahead and rename all the arguments.

> (and BTW downsampling can be useful for metering expensive processes at a lower frame rate, for example)

I think I know what you mean, but I'm not entirely sure. Can you give a practical example?

telephon:
> What about `reblockSize` and `resampleFactor`?

These are better, yes!

sadguitarius:
> I think I know what you mean, but I'm not entirely sure. Can you give a practical example?

I guess something like an expensive FFT process you're using for metering in a GUI and don't want to use extra CPU cycles for. Not something I do much myself but I've seen people doing things like that wrapped in a downsampled poly~ in Max.

> The filter would be a regular UGen that is added after the inputs resp. before the outputs. In my example in the PR description I just use two 2nd order Butterworth low-pass filters (`LPF`) as a crude anti-aliasing filter.

Got it! That's super easy. Although having a way to run the filters inside of the actual upsampling process could be cool for saving a few cycles.

Everything you're saying about zero-padding etc makes sense. I'll have a go at adapting some of my resampling filter code to making a dedicated UGen for this. I have a nice one that's more of less tuned to get the same response as the Max poly~ filters with a comparable CPU load.

Spacechild1:
> I'll have a go at adapting some of my resampling filter code to making a dedicated UGen for this. I have a nice one that's more of less tuned to get the same response as the Max poly~ filters with a comparable CPU load.

Cool! It would be great if you could make a PR! Once we have upsampling support, I think it would make a lot of sense to have dedicated AA filters in the core plugins. I imagine something like `AAIn.ar` and `AAOut.ar`. The filter implementation could be the same, but `AAIn.ar` would also do the zero padding. I think the usage would be pretty clear:
```sc
SynthDef(\distortion, { |in, out|
    var sig = In.ar(in, 2);
    sig = AAIn.ar(sig);
    sig = NastyDistortion(sig);
    sig = AAOut.ar(sig);
    Out.ar(out, sig);
});
```

Spacechild1:
Now that I have completed #7329, I want to move one with the remaining entries in https://github.com/supercollider/supercollider/issues/6102. This PR is one of them.

I have no idea how much this branch has been tested, but at least I haven't received any bug reports so far.

I still have to write some unit tests and update the documentation (https://doc.sccode.org/Classes/Synth.html, https://doc.sccode.org/Guides/WritingUGens.html and https://doc.sccode.org/Reference/ServerPluginAPI.html)

Before I can do that, we should decide on the argument names. So far, at least 3 people are in favor of `reblockSize` and `resampleFactor`.

Honestly, I find `resampleFactor` quite verbose and I would rather shorten it to `resample`. Also, what is the problem with `blockSize`? I'm not sure if `reblockSize` is really a word...

Personally, I prefer conciseness and terseness in the context of creative coding. (SCIDE has not auto-completion for keywords...) I think we can expect that users are already familiar with the meaning of the arguments, particularly with a widely used class like `Synth`. If not, they can easily look it up in the help browser.

---

Long story short, I think we have two options for each argument:

reblocking:
  a. `blockSize`
  b. `reblockSize`

resampling:
  a. `resample`
  b. `resampleFactor`

Should we make a poll? Any other ideas how to make a decision?

capital-G:
> I have no idea how much this branch has been tested, but at least I haven't received any bug reports so far.

Lets simply test it in develop then - thats the best way to get some feedback and discover bugs.

> (SCIDE has not auto-completion for keywords...)

Spoiler: press tab while writing the arguments in a function call ;)

> Any other ideas how to make a decision?

Lets call @supercollider/class-library - according to "the plan" they should come up with a decision in some way.

If it makes a difference - I personally prefer option a

Spacechild1:
> Spoiler: press tab while writing the arguments in a function call ;)

😮 🤦‍♂️ 

> Lets call @supercollider/class-library - according to "the plan" they should come up with a decision in some way.

Excellent idea! :)

> If it makes a difference - I personally prefer option a

So we're currently at 2:3 for a. vs. b. in case that matters.

telephon:
I prefer (a).

A good compromise may be `blockSize` and `resampleFactor`.

sadguitarius:
Not to complicate things too much, but how about `reblock` and `resample`? Concise and also easily recognizable as belonging to the same family of operations

Spacechild1:
`reblock` would be fine as well. I agree that this would be more consistent, at least naming-wise.

Note, however, that reblocking takes an *absolute value* (the block size) and resampling takes a *factor*. Some people may find it confusing that the names sound very similar, but accept different types of values.

Now we have the following options:

- reblocking:
  a. `blockSize`
  b. `reblockSize`
  c. `reblock`

- resampling:
  a. `resample`
  b. `resampleFactor`
  
  Other than a slight dislike for `resampleFactor`, I have no strong opinion. What do others think?
  
  `blockSize` and `resample` seemed to be slightly ahead, but now have a new contestant :)

JordanHendersonMusic:
Reblock and resample from me - but no strong opinions, just want a decision as this pr is exciting!

Spacechild1:
I tend towards `reblock` + `resample` as well.

telephon:
ok, looks good to me as well.

Spacechild1:
I think I'll go ahead and change the arguments to `reblock` and `resample` then. I can always change them again later in case there are strong objections.

Spacechild1:
Actually, I'm not really happy with the fact that the resample factor and block size is provided via Synth constructor arguments.

While this seems to be the most natural solution (and it is also the easiest to implement), it has a serious flaw:

As a user, you always have to remember to pass the appropriate arguments whenever you create an upsampled or downblocked Synth. However, the resample factor and block size are typically tied to a specific DSP algorithm, so ideally it should be set *in the SynthDef*.

This gets even worse when spawning Synths with `Pbind` because we don't call the Synth constructor. We would have to add new `resample` and `reblock` keys to the Event prototypes...

I can imagine two alternatives:

---

#### Alternative nr. 1

Set the resample factor and/or block size via SynthDef constructor arguments:
```sc
(
SynthDef(\test, { |out|
    var sig;
    // sig = ...
    Out.ar(out, sig);
}, resample: 2, reblock: 1).add
)
```

We could still keep the `resample` and `reblock` arguments for the Synth constructor so the user can override the values provided in the SynthDef:

```sc
Synth(\test); // default (2x upsampling)
Synth(\test, resample: 4); // override upsampling 
```

For this to work I would have to extend the SynthDef file format, which shouldn't be a big deal. (If it *was* hard to extend, we should change that :)

The downside is that this wouldn't work with the `Function-play` or `Ndef`.

---

#### Alternative nr. 2

Set the resample factor and/or block size via *UGens*:
```sc
(
SynthDef(\test, { |out|
    var sig;
    Resample(2);
    Reblock(1);
    // sig = ...
    Out.ar(out, sig);
}).add;
)
```

This also works nicely with the `Function-play` and with `Ndefs`:
```sc
(
{
    var sig;
    Resample(2);
    Reblock(1);
    // sig = ...
    Out.ar(0, sig);
}.play
)

(
Ndef(\test, {
    var sig;
    Resample(2);
    Reblock(1);
    // sig = ...
    Out.ar(0, sig);
}).play
)
```
The Server side implementation would be pretty much identical to the alternative nr.1. The only difference is how the values are set in the Client. The `Resample` and `Reblock` UGens would just set members in the current SynthDef.

The downside is that the resample refactor and block size are fixed at SynthDef compile time.

Ideally, we should be able to set the values with a Synth control:
```sc
(
SynthDef(\test, { |out, resample, reblock|
    var sig;
    Resample.ir(resample);
    Reblock.reblock(reblock);
    // sig = ...
    Out.ar(out, sig);
}).add;
)
```

I'm not sure if this would be even possible to implement. We only have access to the controls when the Synth has been constructed, but we need to know the sample rate and block size when we create the Synth...

As a compromise, we could use `Resample` and `Reblock` UGens in the UGen graph function and let the user override the values with `resamples` and `reblock` Synth constructor arguments (as implemented in this PR).

---

Any thoughts on this?

telephon:
> We only have access to the controls when the Synth has been constructed, but we need to know the sample rate and block size when we create the Synth...

There is an info ugen called `SampleRate`. So you can get the rate, but maybe you want something else?

I like the idea of using a UGen and it is vaguely similar to the RandSeed ugen. But the fact that you can't change the factors for each synth individually is something we need to consider: is this important? One could always just make different synthdefs, like in multichannel graphs.

JordanHendersonMusic:
How about having either a static Def time definition, or a dynamic synth time one? Would this simplify some things? Seems unlikely you'd need both.

> The downside is that this wouldn't work with the Function-play or Ndef.

How about a special type of function to define the static defaults.

```
SynthDef(\t, Re(reblock: 1, resample: 2) {
   ...
}).add;

\\ easy conversion from function.play
Re(1, 2) {
   ...
}.play;

\\ I think this one would work, not too knowledgeable about ndef.
Ndef(\t, Re(1, 2) {
   ...
}).play
```

muellmusik:
I quite like alternative 2 (UGens). It seems consistent with similar things as Julian points out. I think it's okay to have to make a separate SynthDef for different blocks/rates. This is consistent with UGens that have variable numbers of outputs, and I think for most users would not be onerous.

Spacechild1:
> There is an info ugen called SampleRate. So you can get the rate, but maybe you want something else?

I want to set it :) The problem is just that this needs to be done *before* the Unit constructors run. In the case of Supernova, I already need to know the block size before the memory is allocated. (In Supernova, every Synth allocates its own wirebuffers because Synth may run in parallel. Scsynth can just use a single buffer for all Synths.)

> I like the idea of using a UGen and it is vaguely similar to the RandSeed ugen. 

Yes. I was also thinking of `MaxLocalBufs` (which only sets the `localMaxBufNum` member`Graph`).

> I think it's okay to have to make a separate SynthDef for different blocks/rates. This is consistent with UGens that have variable numbers of outputs, and I think for most users would not be onerous.

That's what I was thinking as well. That would be easiest solution.

However, in scsynth it would actually be possible to use Synth controls for `Resample` and `Reblock`. In `Graph_New` all the Synth control arguments are set before I even check the `reblock` and `resample` arguments. This means that `Resample` and `Reblock` could just add a field to the SynthDef file that either points to a constant or to a control.

The only issue is that in Supernova all the Synth arguments are only set after the Synth has been created, at which point it is already too late.

I guess I'll just add a warning when the user attempts to pass a control to `Resample` or `Reblock` in Supernova. It would still use the control's default value, but it would ignore any Synth arguments. If we really want, we can later rewrite the Synth creation logic to match the behavior of scsynth, but that's not something I would want to do at the moment.

Spacechild1:
I'm finally closing this in favor of https://github.com/supercollider/supercollider/pull/7402. The new version with the `Resample` and `Reblock` objects is so much better IMO.

PR #6707, opened 2025-04-07
Opportunities to link to `File` and `PathName` in Docs

tedmoore:
## Purpose and Motivation

I noticed a few places where help information about a `path` argument just said "A String". I think it is good to include a link to `link::Classes/PathName::` and `link::Classes/File::` since those are useful classes for dealing with `path`s. There are likely more opportunities in the help for this to happen. Perhaps in a forthcoming help styleguide this can be included.

## Types of changes

- Documentation

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/Classes/Buffer.schelp
+++ b/HelpSource/Classes/Buffer.schelp
@@ -117,7 +117,7 @@ Allocate a buffer and immediately read a soundfile into it. This method sends a 
 argument:: server
 The server on which to allocate the buffer.
 argument:: path
-A String representing the path of the soundfile to be read.
+An instance of link::Classes/String:: representing the path of the soundfile to be read. For path handling and management, see link::Classes/File:: and link::Classes/PathName::
 argument:: startFrame
 The first frame of the soundfile to read. The default is 0, which is the beginning of the file.
 argument:: numFrames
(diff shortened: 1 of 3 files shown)


SimonDeplat:
Thanks for drafting this. Since we're here, I'd like to say that I find `PathName` confusing. Sometimes, methods only work with `String`s, and not `PathName`s, sometimes it is the other way around, sometimes you need to `.standardizePath`, sometimes you don't. Or maybe I'm wrong, but I feel like I always need to try 2 - 3 different approach to get things working (and I never understand why).

If this is indeed the case, unifying methods path arguments is another PR, but if some methods are expecting `String` and others are expecting `PathName`, this should be absolutely clear within the method documentation.

tedmoore:
Interesting. Do you have some examples that might demonstrate this. Perhaps it is an issue that could be improved through better documentation of these classes?

SimonDeplat:
I think my previous comment is wrong. I parsed every `.scd` file I wrote containing `PathName`, and I couldn't find any occurrence of the described problem. It would seem any argument path is indeed always expecting a `String`.

I believe that what got me confused is that you can do this :
`this.executeFile("path/to/a/file")`
but not this :
`this.executeFile(PathName("path/to/a/file"))`
(which leads to this: `this.executeFile(PathName("path/to/a/file").fullPath)`)

In other words, a `PathName` instance is not a pathname. I think this is what got me confused.

I think this PR helps clarifying this.

SimonDeplat:
I opened  #6752 to clarify the ambiguity between the `PathName` class and the pathname concept. Turns out it might be worth removing `PathName` so that `String` is responsible for paths manipulation. This needs time to be implemented, but this would affect the results of this PR.

adcxyz:
@tedmoore - Please take a look at #6753, which proposes to replace PathName fully with String, and keep PathName for backwards compatibility.

tedmoore:
Thanks @adcxyz . We'll hold off on this until a decision is made about that!

Issue #6215, opened 2024-02-13
Buffer:-alloc failing and query gives unexpected results


mtmccrea:
## Environment

* SuperCollider version: 3.12.2, 3.13
* Operating system: MacOS Sonoma 14.3

## Steps to reproduce
The behavior is observable using the code snippet in `Buffer:-alloc` help doc:
```supercollider
s.boot;
b = Buffer.new(s, 44100 * 8.0, 2);
b.query; // numFrames = 0
b.alloc;
b.query; // numFrames = 352800
b.free;
```

## Expected vs. actual behavior
Buffer should be allocated (fails in 3.13), the the state should be accurately returned when queried (not sure what the full query results should be for an nonallocated buffer).
It seems that either the state isn't properly set and/or `query` is returning erroneous results.

### SC 3.12.2 release
```supercollider
b = Buffer.new(s, 44100 * 8.0, 2);
-> Buffer(0, 352800.0, 2, 48000.0, nil)

b.query; // numFrames = 0
-> Buffer(0, 352800.0, 2, 48000.0, nil)
    bufnum: 0
    numFrames: 0
    numChannels: 0
    sampleRate: 0.0

b.alloc;
-> Buffer(0, 0, 0, 0.0, nil) // failed silently?

b.query; // numFrames = 352800
-> Buffer(0, 0, 0, 0.0, nil) // no change in state, not allocated
    bufnum: 0
    numFrames: 0
    numChannels: 0
    sampleRate: 0.0
```

### SC 3.13 release
```supercollider
b = Buffer.new(s, 44100 * 8.0, 2);
-> Buffer(0, 352800.0, 2, 48000.0, nil)

b.query; // numFrames = 0
-> Buffer(0, 352800.0, 2, 48000.0, nil)
    bufnum: 0
    numFrames: 0
    numChannels: 0
    sampleRate: 0.0 

b.alloc;
-> Buffer(0, 0, 0, 0.0, nil)
    /b_alloc: memory allocation failed
```

I don't have time to dig deeper into the issue atm, but the problem seems specific to the _instance_ method because `Buffer:*alloc` appears to be working fine:
```supercollider
b = Buffer.alloc(s, s.sampleRate * 8.0, 2);
-> Buffer(3, 352800.0, 2, 48000.0, nil)

b.query
-> Buffer(3, 352800.0, 2, 48000.0, nil)
    bufnum: 3
    numFrames: 352800
    numChannels: 2
    sampleRate: 48000.0
```

[...]

PR #6710, opened 2025-04-07
scsynth: Buffer: Implementation of setSampleRate (a sampleRate setter)

### LINKED ISSUE #6543: make Buffer.sampleRate actually change the sample rate attribute on the server
## Motivation

as discussed here: https://scsynth.org/t/how-to-change-a-buffers-samplerate-everywhere/

currently a Buffer instance accepts the setter sampleRate yet changes nothing on the server. For a few moments, there is a discrepancy between the language side and the server side representation of the said buffer.

adding the /b_setSampleRate message to the server, and action it when Buffer.sampleRate is called.

## Plan for Implementation

I can do a PR with the said call to the server - I have some experience with augmenting the buffer class [here](https://github.com/tremblap/OfflineBufferProcesses/blob/master/OfflineBufferProcesses.cpp) and this is not hard to do and shouldn't break anything, just make the call consistent with its implied feature description.

If people agree, I am happy to run with it.

tremblap:
## Purpose and Motivation

The sampleRate method in the language was not updating the server side, which was a lost opportunity, as extensively discussed in both the [forum](https://scsynth.org/t/how-to-change-a-buffers-samplerate-everywhere/10379/15) and  bug report #6543 . After much considerations, it seems that it was worth implementing, as it would afford new creative coding opportunities without any backward compatibility issues.

## Types of changes

- Documentation
- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/Server-Command-Reference.schelp
+++ b/HelpSource/Reference/Server-Command-Reference.schelp
@@ -720,6 +720,14 @@ table::
 ::
 Set contiguous ranges of sample indices to sets of values. For each range, the starting sample index is given followed by the number of samples to change, followed by the values.
 
+subsection:: /b_setSampleRate
+Set the sampling rate of the buffer.
+table::
+## strong::int:: || buffer number
+## strong::float:: || the desired sampling rate
+::
+Set the sampling rate of the buffer.
+
 subsection:: /b_fill
 Fill ranges of sample value(s).
 table::
@@ -1130,8 +1138,12 @@ enum {
 
     cmd_version = 64,
 
-    NUMBER_OF_COMMANDS = 65
-};
+    cmd_rtMemoryStatus = 65,
+    
+    cmd_b_setSampleRate = 66,
+    
+    NUMBER_OF_COMMANDS = 67
+    };
 ::
 
 copyright © 2002 James McCartney
--- a/SCClassLibrary/Common/Control/Buffer.sc
+++ b/SCClassLibrary/Common/Control/Buffer.sc
@@ -595,6 +595,17 @@ Buffer {
 		^[\b_query, bufnum]
 	}
 
+	setSampleRate { arg value;
+		// sends the message first to check sanity of sr before changing the langage-side instance variable
+		server.listSendMsg(this.setSampleRateMsg(value));
+		sampleRate = value !? _.asFloat;
+    }
+
+	setSampleRateMsg { arg value;
+		if(bufnum.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
+		^[\b_setSampleRate, bufnum, value !? _.asFloat ?? 0.0]
+	}
+
 	updateInfo { arg action;
 		// add to the array here. That way, update will be accurate even if this buf
 		// has been freed
--- a/server/supernova/sc/sc_osc_handler.cpp
+++ b/server/supernova/sc/sc_osc_handler.cpp
@@ -2493,6 +2493,22 @@ void handle_b_setn(ReceivedMessage const& msg) {
     }
 }
 
+void handle_b_setSampleRate(ReceivedMessage const& msg) {
+    osc::ReceivedMessageArgumentIterator it = msg.ArgumentsBegin();
+    osc::ReceivedMessageArgumentIterator end = msg.ArgumentsEnd();
+    verify_argument(it, end);
+    osc::int32 buffer_index = it->AsInt32();
+    ++it;
+
+    SndBuf* buf = sc_factory->get_buffer_struct(buffer_index);
+    if (!buf) {
+        log_printf("/b_setSampleRate called on unallocated buffer\n");
+        return;
+    }
+
+    buf->samplerate = it->AsFloat();
+    buf->sampledur = 1.0 / buf->samplerate;
+}
 void handle_b_fill(ReceivedMessage const& msg) {
     osc::ReceivedMessageArgumentIterator it = msg.ArgumentsBegin();
     osc::ReceivedMessageArgumentIterator end = msg.ArgumentsEnd();
@@ -3162,6 +3178,10 @@ void sc_osc_handler::handle_message_int_address(ReceivedMessage const& message, 
         handle_b_setn(message);
         break;
 
+    case cmd_b_setSampleRate:
+        handle_b_setSampleRate(message);
+        break;
+
     case cmd_b_fill:
         handle_b_fill(message);
         break;
@@ -3409,6 +3429,11 @@ void dispatch_buffer_commands(const char* address, ReceivedMessage const& messag
 
     if (strcmp(address + 3, "setn") == 0) {
         handle_b_setn(message);
+        return;
+    }
+
+    if (strcmp(address + 3, "setSampleRate") == 0) {
+        handle_b_setSampleRate(message);
         return;
     }
 
--- a/HelpSource/Classes/Buffer.schelp
+++ b/HelpSource/Classes/Buffer.schelp
@@ -466,14 +466,20 @@ Returns the number of channels in the corresponding server-side buffer.
 method:: sampleRate
 Returns the language-side instance variable code::sampleRate::, of the corresponding server-side buffer. By default, this will be set to the server's sample rate when the code::Buffer:: is instantiated.
 
-The code::sampleRate:: setter method only changes the language-side instance variable. It does not resample the buffer or change its sample rate on the server.
-However, it can be useful, for example, for later inferring whether audio-rate or control-rate data has been writen to the buffer.
+The code::sampleRate:: setter method only changes the language-side instance variable. It does not resample the buffer or change its sample rate on the server. To change the server-side buffer sample rate, use code::setSampleRate::.
 note::
 Setting the code::sampleRate:: instance variable affects the returned link::#-duration:: (which returns code::numFrames / sampleRate::).
 ::
 warning::
 You can use link::#-query:: to inspect the buffer's sample rate on the server, but this will also emphasis::update the:: code::sampleRate:: emphasis::instance variable::, overwriting its value which you may have previously set.
 ::
+
+
+method:: setSampleRate
+Sets the sample rate of the buffer on the server side, and updates its value on the language side. This does not resample the audio.
+argument:: value
+The new desired sample rate.
+
 
 method:: duration
 Returns code::this.numFrames / this.sampleRate:: (language-side instance variables).
--- a/include/server/SC_OSC_Commands.h
+++ b/include/server/SC_OSC_Commands.h
@@ -117,5 +117,7 @@ enum {
 
     cmd_rtMemoryStatus = 65,
 
-    NUMBER_OF_COMMANDS = 66
+    cmd_b_setSampleRate = 66,
+
+    NUMBER_OF_COMMANDS = 67
 };


tremblap:
Terse test code:

```
d = Buffer.read(s, Platform.resourceDir +/+ "sounds/a11wlk01.wav");

d.play(true);

d.query; // note the sampling rate

d.sampleRate_(24100); /hear the change

d.query; // and query for good measure

**Comment by dyfer** [comment]
Also pinging @Spacechild1 as we happened to talk about buffer sample rate recently.

**Comment by tremblap** [comment]
Hello Marcin

Sorry for the clang-format-mishap, I just forgot! I have run it and push, but I'll read your docs now (I never use python for it, just run it myself, but you guys might have some cool stuff there so I'll check. 

As for the saving, yes, it actually replaces everything everywhere (on language and server) so it should all behave (even in real time).

**Comment by tremblap** [comment]
ok now lint passes, on to the next failed tests, let me know what I did wrong (or could do better) thanks!

**Comment by dyfer** [comment]
See the output of the CI: https://github.com/supercollider/supercollider/actions/runs/14316949072/job/40125276767#step:9:1535

When you build locally, did you build supernova as well? This is where the issue seems to be.

**Comment by tremblap** [comment]
ok I have compiled now, I hope it behaves, I'm new to supernova so I'll do the cursory tests above

**Comment by tremblap** [comment]
ok the last version works, except for one point:

`d.sampleRate_(24100.0); //hear the change

is needed - the message handler on line 2509 (it->AsFloat();) does not cast an INT to a float like the scsynth one. i wonder what people do to accept arguments in message that are casted - should I add a (asFloat) in the class definition?

**Comment by dyfer** [comment]
> i wonder what people do to accept arguments in message that are casted

I don't know, sorry.

**Comment by tremblap** [comment]
in all cases, this last version should past the tests - it builds and work locally - I will search the supernova code to see tricks of casting int to float like scsynth does.

**Comment by tremblap** [comment]
I changed the class defintion to be cleaner and sturdier, and tested in scsynth and supernova with int and float as new sr. I hope it is sturdier but again, code review might propose slicker approaches. Thanks for your patience and next time I'll be careful not to forget clang-format and supernova compile

**Comment by capital-G** [comment]
Merge note: should be squashed

**Comment by tremblap** [comment]
> Merge note: should be squashed

Indeed sorry, it was my first PR and I made 2 mistakes (forgot clang-format-ing and to test on supernova) which led to fixes in the class def too... the first post was tidy as I was trying to be a good person 😜

**Comment by capital-G** [comment]
> > Merge note: should be squashed
> 
> Indeed sorry, it was my first PR and I made 2 mistakes (forgot clang-format-ing and to test on supernova) which led to fixes in the class def too... the first post was tidy as I was trying to be a good person 😜

this is np at all and there is absolutely no need to say sorry :) we often perform squashing or do `git push --force`. Larger commits are appreciated to have "a proper git history", but this PR would be also fine as a single commit in the end :)

BTW I hope you found the `pre-commit` setup which is linked on the wiki page: https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit
It is by far the easiest to set up and applies our formatting rules every time you do a commit from there on :)

**Comment by tremblap** [comment]
Wow I didn't see that. If I survive this first PR I will take this for the next one for sure :)

File: SCClassLibrary/Common/Control/Buffer.sc [hunk 1]
```diff
@@ -595,6 +596,13 @@ Buffer {
 		^[\b_query, bufnum]
 	}
 
+	sampleRate_ { | newValue |
```
Spacechild1, line 599: All the other methods use the `arg` style.

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -595,6 +596,13 @@ Buffer {
...
+		sampleRate = newValue !? {
+			sampleRate = newValue.asFloat;
```
Spacechild1, line 601: Why the double assignment? Also note that this would set `sampleRate` to `nil` if `newValue` is `nil`. Is this intentional?

I would also suggest to use `sendMsg` instead of `listSendMsg`.

What about:
```
sampleRate_ { arg newRate;
    if (newRate.notNil) {
        sampleRate = newRate.asFloat;    
        server.sendMsg(\b_setSampleRate, bufnum, sampleRate)
    }
}
```

Now, if passing `nil` is actually an error, then it would make sense to treat is as such:
```
if (newRate.isNil) { Error("Sample rate must not be nil").throw };
```
```

File: HelpSource/Reference/Server-Command-Reference.schelp [hunk 3]
```diff
@@ -720,6 +720,17 @@ table::
...
 
+subsection:: /b_setSampleRate
+Set the sampling rate of the buffer.
+table::
+## strong::int:: || buffer number
+## N * || table::
+    ## strong::int:: || sample starting index
+    ## strong::float:: || the required sampling rate
+    ::
+::
```
Spacechild1, line 731: Seems like you just copy-pasted this from `/b_setn` or `/b_fill` :)

Should be:
```
table::
## strong::int:: || buffer number
## strong::float:: || the desired sampling rate
::
```

SCClassLibrary/Common/Control/Buffer.sc
tremblap: Sometimes in the code (see Buffer.sc line 383) nil is assigned (or I misunderstand it) so I wanted to let it happen for those times, and not update the server when that happens. I've refactor to remove the erroneous assignation.

HelpSource/Reference/Server-Command-Reference.schelp
tremblap: I had a git problem so indeed I copied back the wrong document. corrected.

SCClassLibrary/Common/Control/Buffer.sc
tremblap: @Spacechild1 I don't know how reviews work here - do I close all corrected 'conversations'?

Spacechild1: There is a button "Resolve conversation". But I'm not entirely sure myself who is responsible for clicking it :) I just went ahead and resolved the conversation.

tremblap: ok I'll resolve the non-ambiguous ones and will wait for your re-review in silence :)

SCClassLibrary/Common/Control/Buffer.sc
Spacechild1: > Sometimes in the code (see Buffer.sc line 383) nil is assigned

Where exactly?

I noticed you are still using `listSendMsg` instead of `sendMsg`. Only use `listSendMsg` if you have a good reason to do so.

Seems like you are still allowing to set the sample rate to `nil`. Are you sure this is not an error? If not, what does it *mean* to set the sample rate to `nil`?

Spacechild1: > now IIUC that is an allocation to the local instance of that variable (on the language side) that we definitely do not want to send to the server. This is in the original code. So I didn't want to break it

Aha, I see what you mean. But this is a misunderstanding. `sampleRate = nil` assigns directly to the member variable, it does *not* call the setter!

In fact, the Buffer methods must now be careful *not* to call the setter because it would send a message to the Server! I have found at least three instances where this is happening! See `Buffer.new`, `Buffer.alloc` and  `Buffer.initServerCache`. Note that it is not possible to set member variables inside class methods without calling the setter.

Therefore I would advice not to use `sampleRate_` for updating the sample rate, but rather use a more explicit method!

> as for using listSendMsg, the whole class does it, so I did it too. 

Most places just forward messages obtained from the corresponding`*Msg` method. In this case it makes sense to call `listSendMsg` because you get the message from another method as an Array. But if you pass the arguments yourself, there is no point in constructing a temporary Array and passing it to `listSendMsg`, instead of just passing the arguments directly to `sendMsg`. Yes, there are a few places in the `Buffer` class that do this, but IMO it's not good style.

> I don't know what I gain nor I lose.

With `listSendMsg` you gain nothing and lose a little bit of performance. In practice, it won't matter much, but I thought I should point it out.

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -595,6 +596,13 @@ Buffer {
 		^[\b_query, bufnum]
 	}
 
+	sampleRate_ { arg newValue;
```
telephon, line 599: a good method name would be `setSampleRate`. This would make clear it is a bit more than just setting the instance variable. In general, this kind of thing somtimes done one way, sometimes another, but here we need to make sure it doesn't create unwanted network messages.

Spacechild1:
I don't think it's a good idea to use the `sampleRate_` getter for changing the Buffer sample rate, for the following reasons:

1. It is inconsistent with the other instance variables, like `numFrames` or `numChannels`
2. I am not aware of any other setter method in the Class Library that sends a Server command
3. Some class methods need to call the getter and this would accidentally send a message to the Server
4. Practically all methods that send a Server command have a corresponding `*Msg` method which returns the OSC message as an Array for bundeling. How would this look like with a setter? `sampleRate_` vs. `sampleRateMsg`? 🫤 

I would therefore suggest to use a dedicated method for changing the sample rate. This would be more explicit and fix the above mentioned issues. Maybe `setSampleRate` and `setSampleRateMsg`?

---

To take a step back: How often do we actually need to *change* the sample rate of an already allocated Buffer?

Don't we rather want to specify the sample rate when reading/allocating the buffer?

Don't get me wrong, a `setSampleRate` method could still be useful, but we might also think about adding a `sample rate` argument to the `/b_alloc`, `/b_read`, etc. commands and the corresponding `Buffer` methods.

SCClassLibrary/Common/Control/Buffer.sc
Spacechild1: I just suggested the same thing :) https://github.com/supercollider/supercollider/pull/6710#issuecomment-2790760340

dyfer:
> Don't we rather want to specify the sample rate when reading/allocating the buffer?

For me that'd be probably the most common use, even the recommended one. However, on occasion I imagine the need to change samplerate of an existing buffer... e.g. to correct an improperly set rate, or for experimental purposes. 

Maybe that's an edge case though and could be as well handled when reading the buffer...? Maybe it's actually safer to only allow this at creation/read time? With great power... etc etc

tremblap:
This is the same discussion we had on the forum, and faster on the bug report. In effect, there is a much deeper issue, but at the moment the `sampleRate` method (and its documentation) lead the user to believe they can change the sampleRate. This is in line with other languages behaviour. What it actually does is to change only the language side value, gives no error, and silently re-replace to the (real) server value at the next query. In the meantime it has 2 different values.

What I would do, short of making sclang have private setters (it isn't possible IIUC) would be to deliver the (almost transparent) parity between the buffer and its representation, and use new methods to set locally the instance variables when it is done by various housekeeping functions.

So public facing sampleRate would be as I proposed (and discussed in those threads(, create a 'private' setting method when only the sclang needs to update its values (when reading a file). I think this was the conclusion of the threads and what @dyfer is proposing too.

The other option, just to be the devil's advocate:
1. correct the buffer.schelp to mark in bold that that some setters (sampleRate, numFrames, numChannels) shouldn't be public, have not effect, and shouldn't be touched. Maybe link to an explainer on the dichotomy between buffers and their representation, and again how that differs from other languages.
2. make an explicit public setter for sampleRate.

now, if you ask me, this latter option is adding a problem on a problem, interface-wise. Is there any other class that has a parallel setter like this? Is there really no option to make a private setter.

--
As for the internal calls: I am trying to make sense of the code, but .new and .alloc call the .cache method, which in turn calls .query, which in turn updates the internal states of the sclang side... am I reading wrong? I will give it a try anyway.

tremblap:
So after reading again the 'Writing Classes' helper, it seems the `prPrivateMethod` convention would be good to use here. I managed to make it work for `*new` and `*alloc` (inventing a `prSampleRate`) but I cannot catch where `*read` and `read` call for the setter. Is it outside of Buffer.sc? Any pointer welcome.

Spacechild1:
> This is the same discussion we had on the forum, and faster on the bug report. 

Unfortunately, I haven't seen this discussions, otherwise I would have intervened earlier...

> In effect, there is a much deeper issue, but at the moment the sampleRate method (and its documentation) lead the user to believe they can change the sampleRate. 

That's actually a problem with the documentation.

> This is in line with other languages behaviour. 

As I said in my earlier post, I don't know of a single setter that affects the Server representation. In general, the client-side abstractions are *observers* and you only change the Server side object with explicit commands.

Just because properties have public setters does *not* necessarily mean that it makes sense for the users to call them. Often they are only there so that class methods can set the instance variable. Consider the `Node` class:
```
var <>nodeID, <>server, <>group;
var <>isPlaying = false, <>isRunning = false;
```
All of these have public setters, but none of these should be called by the user.

> So public facing sampleRate would be as I proposed (and discussed in those threads(, create a 'private' setting method when only the sclang needs to update its values (when reading a file). I think this was the conclusion of the threads and what @dyfer is proposing too.

I would not agree with that conclusion though.

---

> The other option, just to be the devil's advocate:
> - correct the buffer.schelp to mark in bold that that some setters (sampleRate, numFrames, numChannels) shouldn't be public, have not effect, and shouldn't be touched. Maybe link to an explainer on the dichotomy between buffers and their representation, and again how that differs from other languages.
> - make an explicit public setter for sampleRate.

Yes, that's what I would suggest!

---

And again, what are the use cases for *changing* the sample rate on an already allocated Buffer? *) Wouldn't it make more sense to add `sampleRate` arguments to the read/alloc methods/commands?

*) I'm not saying such a use case doesn't exist, but if the common use case is to allocate/read a buffer with a user-provided sample rate, the API should reflect that.

Spacechild1:
BTW, you haven't really addressed concerns 2. and 4. in https://github.com/supercollider/supercollider/pull/6710#issuecomment-2790760340

tremblap:
@Spacechild1 as discussed in the other thread, it is about allowing bending a parameter, or setting it, or whatever a creative coder might want to do.

@dyfer if you think an explicit setter is the way, and if you all agree this is ok, then I'll convert it to setSampleRate (with a setSampleRateMsg too, don't worry ;)

Spacechild1:
> (with a setSampleRateMsg too, don't worry ;)

:)

BTW, you are completely right that the documentation of the `sampleRate` method is very misleading: 

> Changing this value only changes what the buffer thinks its sample rate is. It does not resample the buffer's content.

This is obviously false, calling the `sampleRate` getter doesn't do anything. Feel free to remove this passage from the docs or replace it with a reference to the new `setSampleRate` method.

dyfer:
Sounds good!

And I think we absolutely need to add this to _at least_ `.alloc`. With `.read` etc maybe it's not as, since we are changing a value that existed previously in the soundfile, but I strongly believe we should be able to set the sample rate of a new buffer at creation time.

tremblap:
> Feel free to remove this passage from the docs or replace it with a reference to the new `setSampleRate` method.

now, a certain Michael McCrea fixed that already 2 weeks ago (commit d5552c2) so it seems it is in the air of the time. It reads really well... so I can focus on adding the new info as a side note.

more soon, I'm implementing...

tremblap:
ok dear panel, here we go. and here is the new test code:

```
s.options.sampleRate_(48000)
s.reboot.dumpOSC

c = Buffer.alloc(s,100)

b = Buffer.new

d = Buffer.read(s, Platform.resourceDir +/+ "sounds/a11wlk01.wav");

d.play(true);

d.query; // note the sampling rate

d.setSampleRate(24100); //hear the change

d.sampleRate // check the state before 'query' updates it

d.query; // and query for good measure

c.sampleRate
c.query

b.read(Platform.resourceDir +/+ "sounds/a11wlk01.wav");
```

dyfer:
Thanks! Are you planning to add the argument for `.alloc`?

tremblap:
just for you my dear :) .new and .alloc should definitely have it indeed. But to be disciplined, I'd do another PR branching from this one (or post merge) ? or do you propose that I bundle that in?

dyfer:
I think that the initial implementation should include .new and .alloc at the very least.

tremblap:
agreed. but on their own PR I reckon?

dyfer:
IMO it should be one PR (this one), so that we merge complete functionality. What do others think?

tremblap:
ok, i understand (i.e. the release cherry picking dance might be affected, right?) in the meantime I'll get coding here.

tremblap:
ok dear colleague. I cannot say I love supernova. but I managed to make it work there too.

here is the test code for all the SR in alloc and new:

```SuperCollider
s.options.sampleRate_(48000)
s.reboot

a = Buffer.new(s,10,2,22000) //impose without alloc
a.alloc
a.query

b = Buffer.new(s,11,5) //default
b.alloc
b.query

c = Buffer.alloc(s,12,3,33221)//impose on alloc
c.query

d = Buffer.allocConsecutive(3,s,24,5,12345) //test consecutive
d[0].query

e = Buffer.alloc(s,1) //minima call
e.query

Server.supernova
s.reboot
```

tremblap:
so supernova build here, but nowhere on the CI... any hint?

capital-G:
> so supernova build here, but nowhere on the CI... any hint?

The CI is building supernova as well, but its failing, see https://github.com/supercollider/supercollider/actions/runs/14432311490/job/40468950546?pr=6710#step:14:7592

```
 /Users/runner/work/supercollider/supercollider/testsuite/server/supernova/buffer_manager_test.cpp:12:35: error: too few arguments to function call, expected 4, have 3
   12 |     bm.allocate_buffer(0, 44100, 2);
```

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -2,6 +2,7 @@ Buffer {
 
 	// don't use the setter methods for the vars below
 	// they're private and have no effect on the server
+	// except for sampleRate
```
Spacechild1, line 5: stale comment?

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
 	*initClass { serverCaches = IdentityDictionary.new }
 
 	// doesn't send
-	*new { arg server, numFrames, numChannels, bufnum;
+	*new { arg server, numFrames, numChannels, sampleRate, bufnum;
```
Spacechild1, line 15: New arguments must always come last! Otherwise this can break existing code.

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
...
 			numFrames,
-			numChannels
-		).sampleRate_(server.sampleRate).cache
+			numChannels,
+			sampleRate.asFloat
+		).cache
 	}
 
-	*alloc { arg server, numFrames, numChannels = 1, completionMessage, bufnum;
+	*alloc { arg server, numFrames, numChannels = 1, sampleRate, completionMessage, bufnum;
```
Spacechild1, line 28: New arguments must always come last! Otherwise this can break existing code.

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
...
 			numFrames,
-			numChannels
-		).alloc(completionMessage).sampleRate_(server.sampleRate).cache
+			numChannels,
+			sampleRate.asFloat
+		).alloc(completionMessage).cache
 	}
 
-	*allocConsecutive { arg numBufs = 1, server, numFrames, numChannels = 1, completionMessage, bufnum;
+	*allocConsecutive { arg numBufs = 1, server, numFrames, numChannels = 1, sampleRate, completionMessage, bufnum;
```
Spacechild1, line 41: New arguments must always come last! Otherwise this can break existing code.

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
...
+			numChannels,
+			sampleRate.asFloat
+		).cache
 	}
 
-	*alloc { arg server, numFrames, numChannels = 1, completionMessage, bufnum;
+	*alloc { arg server, numFrames, numChannels = 1, sampleRate, completionMessage, bufnum;
 		server = server ? Server.default;
 		bufnum ?? { bufnum = server.nextBufferNumber(1) };
+		sampleRate ?? { sampleRate = server.sampleRate };
```
Spacechild1, line 31: See corresponding comment in `Buffer.new`

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
...
 		^super.newCopyArgs(
 			server,
 			bufnum,
 			numFrames,
-			numChannels
-		).alloc(completionMessage).sampleRate_(server.sampleRate).cache
+			numChannels,
+			sampleRate.asFloat
```
Spacechild1, line 37: forward as is

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
...
+		).alloc(completionMessage).cache
 	}
 
-	*allocConsecutive { arg numBufs = 1, server, numFrames, numChannels = 1, completionMessage, bufnum;
+	*allocConsecutive { arg numBufs = 1, server, numFrames, numChannels = 1, sampleRate, completionMessage, bufnum;
 		var	bufBase;
 		bufBase = bufnum ?? { server.nextBufferNumber(numBufs) };
 		numFrames = numFrames.asInteger;
 		numChannels = numChannels.asInteger;
+		sampleRate ?? { sampleRate = server.sampleRate };
```
Spacechild1, line 46: See corresponding comment in `Buffer.new`

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
...
 		^Array.fill(numBufs, { |i|
 			// note, cannot use alloc or allocMsg here
 			// because those methods don't pass a buffer index for completion message
-			var newBuf = Buffer.new(server, numFrames, numChannels, i + bufBase);
-			server.sendMsg(\b_alloc, i + bufBase, numFrames, numChannels,
+			var newBuf = Buffer.new(server, numFrames, numChannels, sampleRate, i + bufBase);
+			server.sendMsg(\b_alloc, i + bufBase, numFrames, numChannels, sampleRate.asFloat,
```
Spacechild1, line 51: forward as is (float or nil).

IMPORTANT: new message arguments must be added at the end to ensure backwards compatibility!

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -67,7 +73,7 @@ Buffer {
 
 	allocMsg { arg completionMessage;
 		this.cache;
-		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, completionMessage.value(this)]
+		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, sampleRate, completionMessage.value(this)]
```
Spacechild1, line 76: IMPORTANT: new message arguments must be added at the end to ensure backwards compatibility!

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -595,6 +601,17 @@ Buffer {
...
 	}
 
+	setSampleRate { arg value;
+		// sends the message first to check sanity of sr before changing the langage-side instance variable
+		server.listSendMsg(this.setSampleRateMsg(value));
+		sampleRate = value.asFloat;
+    }
+
+	setSampleRateMsg { arg value;
+		if(value.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
```
Spacechild1, line 611: I don't understand this error message...

server/scsynth/SC_SequencedCommand.cpp
```diff
@@ -276,6 +276,7 @@ int BufAllocCmd::Init(char* inData, int inSize) {
     mBufIndex = msg.geti();
     mNumFrames = msg.geti();
     mNumChannels = msg.geti(1);
+    mSampleRate = msg.getf(1);
```
Spacechild1, line 279: After completition message! (See corresponding comment in `Buffer.allocMsg`.

server/scsynth/SC_SequencedCommand.cpp
```diff
@@ -287,7 +288,7 @@ void BufAllocCmd::CallDestructor() { this->~BufAllocCmd(); }
 bool BufAllocCmd::Stage2() {
     SndBuf* buf = World_GetNRTBuf(mWorld, mBufIndex);
     mFreeData = buf->data;
-    SCErr err = bufAlloc(buf, mNumChannels, mNumFrames, mWorld->mSampleRate);
+    SCErr err = bufAlloc(buf, mNumChannels, mNumFrames, mSampleRate);
```
Spacechild1, line 291: Better: 
``` 
// use Server sample rate as default value
auto sampleRate = mSampleRate != 0.0 ? mSampleRate : mWorld->mSampleRate;
SCErr err = bufAlloc(buf, mNumChannels, mNumFrames, sampleRate);
```

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -1988,9 +1988,13 @@ template <bool realtime> void handle_b_alloc(ReceivedMessage const& msg, endpoin
 
     args >> bufferIndex >> frames;
 
-    if (!args.Eos())
+    float sampleRate = sc_factory->world.mSampleRate;
```
Spacechild1, line 1991: After `extract_completion_message`! See corresponding comment in `Buffer.allocMsg`

Also, default to `world.mSampleRate` if the sample rate argument is 0.0.

HelpSource/Reference/Server-Command-Reference.schelp
```diff
@@ -566,6 +566,7 @@ table::
 ## strong::int:: || buffer number
 ## strong::int:: || number of frames
 ## strong::int:: || number of channels (optional. default = 1 channel)
+## strong::float:: || the required sample rate (optional. default = the server's sample rate)
```
Spacechild1, line 569: After completion message! See comment in `Buffer.allocMsg`

Spacechild1 (changes requested):

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
 	*initClass { serverCaches = IdentityDictionary.new }
 
 	// doesn't send
-	*new { arg server, numFrames, numChannels, bufnum;
+	*new { arg server, numFrames, numChannels, sampleRate, bufnum;
 		server = server ? Server.default;
 		bufnum ?? { bufnum = server.nextBufferNumber(1) };
+		sampleRate ?? { sampleRate = server.sampleRate };
```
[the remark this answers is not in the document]
tremblap: ok will reimplement

SCClassLibrary/Common/Control/Buffer.sc
tremblap: that is an interesting bias. I will respect it even if I disagree - user interface logic first is my motto, but you're the boss :)

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +12,43 @@ Buffer {
...
 		^super.newCopyArgs(
 			server,
 			bufnum,
 			numFrames,
-			numChannels
-		).sampleRate_(server.sampleRate).cache
+			numChannels,
+			sampleRate.asFloat
```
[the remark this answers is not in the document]
tremblap: this fails supernova if the user sends an int... it is not a problem for scsynth but it is for supernova as it doesn't casts

SCClassLibrary/Common/Control/Buffer.sc
tremblap: I'll make it readable

SCClassLibrary/Common/Control/Buffer.sc
telephon: This is a misunderstanding: of course, for new classes, we'd always take extra care for the best order of arguments. But we almost never change an existing interface later on. We try to keep deprecation to an absolute minimum and by all means try not to break existing code in new versions.

tremblap: I understand your perspective. I will implement accordingly.

SCClassLibrary/Common/Control/Buffer.sc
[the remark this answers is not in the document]
Spacechild1: I see. In that case you can do `sampleRate !? _.asFloat ?? 0.0`.

This would only be necessary in the `allocMsg` method, though. I would also add a comment that explains why this is necessary.

server/scsynth/SC_SequencedCommand.cpp
Spacechild1: Nevermind, we can do that in `BufAllocCmd::Init`. See my comment above.

server/supernova/sc/sc_osc_handler.cpp
Spacechild1: Another note: please follow the style of the other arguments:
```
osc::int32 bufferIndex, frames, channels;
float sampleRate;
...
if (!args.Eos()) {
    args >> sampleRate;
    if (sampleRate == 0.0)
        sampleRate = sc_factory->world.mSampleRate; 
} else {
    sampleRate = sc_factory->world.mSampleRate;
```

HelpSource/Reference/Server-Command-Reference.schelp
Spacechild1: Side note: we would also need an explanation that `0.0 has the special meaning of "use the server's sample rate".

(Since `channels` and `completionMsg` are both optional, `sampleRate` would need to stay optional, too.)

server/supernova/server/buffer_manager.cpp
```diff
@@ -25,7 +25,7 @@
 
 namespace nova {
 
-void buffer_wrapper::allocate(size_t frames, uint channels) {
+void buffer_wrapper::allocate(size_t frames, uint channels, double samplerate) {
```
mtmccrea, line 28: `sampleRate` is not actually used in this function?

mtmccrea:
Sorry if this has already been  discussed, but w.r.t. the instance vars being updated by `.query`, it's actually a side effect of there being an OSC responder listening for `\b_info` messages which is registered in `Buffer:*initServerCache`:
https://github.com/supercollider/supercollider/blob/1c4ccdf3ceb0fc3c9d9c674f422bffadb3454f3c/SCClassLibrary/Common/Control/Buffer.sc#L627-L643

I'm not sure what the original intent was for this, but it could be revisited. I would not expect `.query` to update Buffer vars, but rather just return or post buffer state from the server side. Perhaps a new method would explicitly update the local buffer's vars to sync with the server representation.

As referenced before, this came up in #6217. So it would be important that the result of this PR retained [that functionality](https://github.com/supercollider/supercollider/blob/1c4ccdf3ceb0fc3c9d9c674f422bffadb3454f3c/SCClassLibrary/Common/Core/Function.sc#L326-L333).

jamshark70:
> I would not expect .query to update Buffer vars, but rather just return or post buffer state from the server side.

I do see your point. If I recall, the thinking was, what is the point of knowing that the Buffer's language-side member vars don't match the server's values and *not* correcting the language side values?

mtmccrea:
> what is the point of knowing that the Buffer's language-side member vars don't match the server's values and not correcting the language side values?

My thinking is that these are two separate concerns:
1. `query`: tell me the state of the buffer on the server
2. `Buffer:-syncState` (hypothetical): update the Buffer's state vars to be in sync with the server's representation.

`Buffer:-syncState` would essentially create the responder that's currently in `serverCaches[server][\responder]` as a `oneShot`, the way `query` currently does for the responder that posts the state.

mtmccrea:
@jamshark70 Looks like we've had this discussion before :) #6215

After rediscovering that, I see that the blame history shows that the [original intention](https://github.com/supercollider/supercollider/blob/db7eed2a17c361503dbc7f70a557874b6001e3cd/HelpSource/Classes/Buffer.schelp#L838) of `query` was actually _not_ to update the lang-side state.

tremblap:
Ok what was in the end a simple feature addition now is a multi fix pr which will break main people use if query and Sr... which I don't think is wise. Proposal: I revert to the working entry for setSampleRate and that is mergeable in this pr. Then we can continue in another pr to fix the various changes to Buffer (query, alloc, new) that will break people's code...

server/supernova/server/buffer_manager.cpp
Spacechild1: Good catch!

Spacechild1:
> Ok what was in the end a simple feature addition now is a multi fix pr which will break main people use if query and Sr...

I think you misunderstood. @mtmccrea only said that this PR should keep the current behavior of `b.query` (which AFAICT it does) and only said that we *might* revisit it in the future.

> Then we can continue in another pr to fix the various changes to Buffer (query, alloc, new) that will break people's code...

How would the additions to `Buffer.new`, `Buffer.alloc` and `Buffer.allocConsecutive` break people's code (if following my PR comments)?

Again, we don't really want to touch `Buffer.query` in this PR anyway.

muellmusik:
> After rediscovering that, I see that the blame history shows that the [original intention](https://github.com/supercollider/supercollider/blob/db7eed2a17c361503dbc7f70a557874b6001e3cd/HelpSource/Classes/Buffer.schelp#L838) of `query` was actually _not_ to update the lang-side state.

The original query from over 20 years ago _did_ update the instance vars. I'd have to trace through the code and commits for fine changes, but I believe the update was removed from query as this was latterly handled by a separate listener, currently in the server caches. The current help file is inaccurate I think as this listener will update the vars.

From a design perspective the method exists on the principle that the server abstraction classes should provide an OOP interface to all relevant server messages.

tremblap:
indeed @Spacechild1 I will split this in 3 PRs as this has grown beyond the original intent by a few orders of magnitude (which I don't have time to cater for at the moment)

tremblap:
ok this is now a single issue solving PR: it adds a setter without touching the codebase. Let's consider this done, it fixes #6543 and passes all tests. I am doing another PR draft to try to implement the new desired setting behaviour at creation time for .alloc and .new - I will merge where my code was at and when I have time, I will try to re-read all of this discussion to see what would be the least damageable implementation and the query dance fix if needs be.

Spacechild1:
At this point you might want to squash all the commits into one and then force push:
```
# reset to develop branch while keeping all your changes:
git reset --soft develop
# make a new *single* commit from your changes
git commit
# then force push
git push -f
```

Side note Nr 1:  instead of *merging* the develop branch into your branch, consider *rebasing* your branch on top of develop (`git rebase develop`)

Side note Nr 2: if you want to fix the most recent commit, you can *ammend* it (`git commit --ammend`) and then force push. No need to push many little fix up commits. Keep in mind that commits are meant to be read by the reviewer.

> and the query dance fix if needs be.

I don't think that you need to worry about `query`.

tremblap:
@Spacechild1 thanks so much for that git wisdom. I am always afraid of rebasing and amending... but doing so here makes so much more sense...

I've cleaned up and made a single commit for eventual reviewers. This is now only implementing setSampleRate method for scsynth

SCClassLibrary/Common/Control/Buffer.sc
Spacechild1: Don't forget about the error message :)

Spacechild1:
Great! Next time you ammend, you can also fix the commit message :)

tremblap:
now I thought I did! "clean implementation of setSampleRate method for buffers"

isn't that clear enough?

Spacechild1:
> isn't that clear enough?

In case you're not joking:

- why "clean implementation"? That's what we would hope for anyway :-D I guess this is just a leftover from your last commit.
- you could add a short explanation of what this new Buffer method does
- you could also mention the new `/b_setSampleRate` Server command

Typically, the format of a commit message is:
```
<compartment>: <short description>

<more details>
```

tremblap:
I was definitely not joking. I'll try again, but this feels that the commit message is redundant with the PR explanation. If you have a good example in the current codebase of a terse good practice, let me know the commit hash and I'll get inspired

Spacechild1:
> what do you think about keeping this PR as adding setting sample rate of any existing buffer only, vs adding sample rate parameters when creating a buffer?

IIUC, @tremblap wants to implement the `sampleRate` argument for `.alloc` and `.new` in a separate PR.

>  I'm worried that .alloc and .new methods won't get implemented and we'll end up with what I consider an incomplete implementation.

I agree.  I would say let's only merge this PR (`setSampleRate`) *after* the other (`.alloc` and `.new`).

dyfer:
> There is a separate (draft) PR for `.alloc` and `.new`: #6800

As, I didn't see it, thanks!

I think it's a great idea to merge #6800 first and then this one.

tremblap:
@Spacechild1 error message fix, and check logic is better too.

tremblap:
this is all green except one - which I cannot reproduce here. Let me know.

tremblap:
the tests failing here are the same as #6800 and yet they don't touch the buffer, so I am a little out of resources to troubleshoot them.

dyfer:
Don't worry about these tests, they often fail.

Spacechild1:
Superseded by https://github.com/supercollider/supercollider/pull/6800

PR #6711, opened 2025-04-07
InlineUnaryOp.h: increase precision of constants

### LINKED ISSUE #5016: midiratio could be more precise
Not a big issue but sometimes annoying, especially in obvious cases:
```supercollider
12.midiratio // -> 1.9999999999945
2.pow(12/12) // -> 2.0 
```
When I think about Henry Partch (roughly) saying that equal temperament is a lie compared to just intonation, this gets particularly annoying :)

Looking at the source code, it seems like midicps/cpsmidi use more precise constants than midiratio/cpsoct... anyone know if this is a bug or intended for some reason I can't understand?

https://github.com/supercollider/supercollider/blob/f806ace7bd8565dd174e7d47a1b32aaa4175a46e/include/plugin_interface/SC_InlineUnaryOp.h#L283

https://github.com/supercollider/supercollider/blob/f806ace7bd8565dd174e7d47a1b32aaa4175a46e/include/plugin_interface/SC_InlineUnaryOp.h#L290
## Environment

* SuperCollider version: 3.11.1
* Operating system: Arch Linux

## Steps to reproduce

```supercollider
12.midiratio // -> 1.9999999999945
2.pow(12/12) // -> 2.0 
```

I would expect 12.midiratio to be exactly 2.0

dyfer:
## Purpose and Motivation

Based on the discussion in #5016 I think this is an innocuous change, but please review carefully.

Fixes #5016. 

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- ~~[ ] Updated documentation - should this be noted in the docs? I think not, but I'm open to suggestions.~~

### DIFF
--- a/include/plugin_interface/SC_InlineUnaryOp.h
+++ b/include/plugin_interface/SC_InlineUnaryOp.h
@@ -75,17 +75,13 @@ inline float32 sc_log2(float32 x) { return std::log2(x); }
 inline float32 sc_log10(float32 x) { return std::log10(std::abs(x)); }
 
 /// Convert MIDI note to cycles per second
-inline float32 sc_midicps(float32 note) {
-    return (float32)440. * std::pow((float32)2., (note - (float32)69.) * (float32)0.083333333333);
-}
+inline float32 sc_midicps(float32 note) { return 440.f * std::pow(2.f, (note - 69.f) / 12.f); }
 
 /// Convert cycles per second to MIDI note.
-inline float32 sc_cpsmidi(float32 freq) {
-    return sc_log2(freq * (float32)0.0022727272727) * (float32)12. + (float32)69.;
-}
+inline float32 sc_cpsmidi(float32 freq) { return sc_log2(freq / 440.f) * 12.f + 69.f; }
 
 /// Convert an interval in MIDI notes into a frequency ratio.
-inline float32 sc_midiratio(float32 midi) { return std::pow((float32)2., midi * (float32)0.083333333333); }
+inline float32 sc_midiratio(float32 midi) { return std::pow(2.f, midi / 12.f); }
 
 /// Convert a frequency ratio to an interval in MIDI notes.
 inline float32 sc_ratiomidi(float32 ratio) { return (float32)12. * sc_log2(ratio); }
@@ -94,7 +90,7 @@ inline float32 sc_ratiomidi(float32 ratio) { return (float32)12. * sc_log2(ratio
 inline float32 sc_octcps(float32 note) { return (float32)440. * std::pow((float32)2., note - (float32)4.75); }
 
 /// Convert cycles per second to decimal octaves.
-inline float32 sc_cpsoct(float32 freq) { return sc_log2(freq * (float32)0.0022727272727) + (float32)4.75; }
+inline float32 sc_cpsoct(float32 freq) { return sc_log2(freq / 440.f) + 4.75f; }
 
 /// Convert linear amplitude to decibels.
 inline float32 sc_ampdb(float32 amp) { return std::log10(amp) * (float32)20.; }
@@ -265,7 +261,7 @@ inline float32 sc_lg3interp(float32 x1, float32 a, float32 b, float32 c, float32
     float32 x3 = x1 - 2.f;
 
     float32 x03 = x0 * x3 * 0.5f;
-    float32 x12 = x1 * x2 * 0.16666666666666667f;
+    float32 x12 = x1 * x2 / 6.f;
 
     return x12 * (d * x0 - a * x3) + x03 * (b * x2 - c * x1);
 }
@@ -319,17 +315,13 @@ inline float64 sc_log2(float64 x) { return std::log2(std::abs(x)); }
 inline float64 sc_log10(float64 x) { return std::log10(std::abs(x)); }
 
 /// Convert MIDI note to cycles per second
-inline float64 sc_midicps(float64 note) {
-    return (float64)440. * std::pow((float64)2., (note - (float64)69.) * (float64)0.08333333333333333333333333);
-}
+inline float64 sc_midicps(float64 note) { return 440. * std::pow(2., (note - 69.) / 12.); }
 
 /// Convert cycles per second to MIDI note.
-inline float64 sc_cpsmidi(float64 freq) {
-    return sc_log2(freq * (float64)0.002272727272727272727272727) * (float64)12. + (float64)69.;
-}
+inline float64 sc_cpsmidi(float64 freq) { return sc_log2(freq / 440.) * 12. + 69.; }
 
 /// Convert an interval in MIDI notes into a frequency ratio.
-inline float64 sc_midiratio(float64 midi) { return std::pow((float64)2., midi * (float64)0.083333333333); }
+inline float64 sc_midiratio(float64 midi) { return std::pow(2., midi / 12.); }
 
 /// Convert a frequency ratio to an interval in MIDI notes.
 inline float64 sc_ratiomidi(float64 ratio) { return (float64)12. * sc_log2(ratio); }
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
I think this whole section could be rewritten.

First, there is no need for the casts. `0.` is a double, `0.f` is a float, and `0.l` is a long double. Casting just makes it harder to read.

Second, perhaps the constants here could be calculated instead, static consteval variables can be used to ensure the compiler calculates them at compile time when they use math functions. 

That means, rather than writing `0.8333...` perhaps it could just be `1./12.`.

dyfer:
@JordanHendersonMusic thanks for the feedback!

should all the casts be removed?
E.g. from:

```cpp
return (float32)440. * std::pow((float32)2., (note - (float32)69.) * (float32)0.083333333333);
```

should I do

```cpp
return 440. * std::pow(2., (note - 69.) * (1./12.));
```
or is division okay?
```cpp
return 440. * std::pow(2., (note - 69.) / 12.);
```
and should there be a final cast to float32?
```cpp
return (float32)440. * std::pow(2., (note - 69.) * (1./12.));
```

JordanHendersonMusic:
@dyfer there shouldn't be a cast from a literal to another literal type, you can always just use the correct literal. So the second example you have is what I imagined.


Re division or multiplication, I don't think it matters, but if we are bothered about performance here, we'd need to check compiler explorer and see what the compiler does - I *believe* the compiler can turn division into multiplication in some cases.
Either way, I'd be in favour of division because it's clearer.

The final cast only applies to the literal there... but there could be one for the return type if the internals are working at a higher precision, but I don't think that is the case here (didn't look too closely though!).

capital-G:
> we'd need to check compiler explorer and see what the compiler do

I think the most explicit way of doing this would be to put the value into a `static contexpr`, so something like this?

```cpp
    static constexpr float fraction = (1./12.);
    return 440. * std::pow(2., (note - 69.) * fraction);
```

But looking at the compiler output of clang14 w/ -o3, it already optimizes it - see https://godbolt.org/z/7M9axv5ej

JordanHendersonMusic:
> I think the most explicit way of doing this would be to put the value into a static contexpr, so something like this?

I think it might actually be `consteval` here, constexpr means it *can* be evaluated at compile time, but doesn't have to be. But yes, the compiler should do this one automatically :)

dyfer:
Thanks for the input!

To clarify, is this the refomended form?

```cpp
inline float32 sc_midicps(float32 note) {
   return 440.f * std::pow(2.f, (note - 69.f) / 12.f);
}
```

dyfer:
> I don't see a definition for `sc_log2(double)`

[Yes, it's there](https://github.com/supercollider/supercollider/pull/6711/files#diff-3d14c68523ee9addc1dcf6e4711a1daec26e7e70f13d8a304cee984a60340f89L317).

I removed the comment `all constants explicitly cast to prevent PowerPC frsp instruction generation.` as I'm guessing this is not an issue anymore with modern compilers, even on PPC... right? (also building SC on PPC is probably quite theoretical, at this point)

I have a feeling it would be great to add tests for this... but I'd rather not spend time on it if others don't think it's necessary.

include/plugin_interface/SC_InlineUnaryOp.h
```diff
@@ -412,40 +401,38 @@ inline float64 sc_scurve(float64 x) {
...
+    if (x >= 1.)
+        return 1.;
     return x;
 }
 
 /// Answer -1 if negative, +1 if positive or 0 if zero
-inline float64 sc_sign(float64 x) {
-    return x < (float64)0. ? (float64)-1. : (x > (float64)0. ? (float64)1.f : (float64)0.f);
-}
+inline float64 sc_sign(float64 x) { return x < 0. ? -1. : (x > 0. ? 1.f : 0.f); }
```
JordanHendersonMusic, line 417: The floats are incorrect here.

JordanHendersonMusic (changes requested):

include/plugin_interface/SC_InlineUnaryOp.h
dyfer: Thanks for catching!

JordanHendersonMusic:
> as I'm guessing this is not an issue anymore with modern compilers, even on PPC... right?

To be honest, I'm not really sure what that's all about. `frsp` is double to single conversion but if the compiler emits a single... Then there shouldn't be a conversion? Maybe it was some oddity of the compiler/ill defined c++ behaviour that has since been fixed (it's would have been pre C++11). I've searched around a bit but haven't found any extra information on it either.

JordanHendersonMusic (approved):

dyfer:
To double check: do you think it's okay to merge this without writing tests?

dyfer:
I didn't see other places testing for these methods - I added a few tests testing the improved accuracy (i.e. these tests fail in SC 3.13).

dyfer:
I assume it's okay to merge this?

JordanHendersonMusic:
Go ahead, I think this is good and shouldn't negatively affect anything.

include/plugin_interface/SC_InlineUnaryOp.h
```diff
@@ -265,7 +258,7 @@ inline float32 sc_lg3interp(float32 x1, float32 a, float32 b, float32 c, float32
     float32 x3 = x1 - 2.f;
 
     float32 x03 = x0 * x3 * 0.5f;
-    float32 x12 = x1 * x2 * 0.16666666666666667f;
+    float32 x12 = x1 * x2 / 6.f;
```
timblechmann, line 261: note: divisions can be an order of magnitude slower than multiplications with the reciprocal (and compilers may not be allowed to do these conversions).
these added divisions are quite likely to cause performance regressions. as a rule of thumb: avoid divisions whenever possible

JordanHendersonMusic: Shoul we enable `-freciprocal-math`?

https://godbolt.org/z/d3PMzrdrP

dyfer: Yes, I'm aware of the performance difference with divisions - _in principle_.

> as a rule of thumb: avoid divisions whenever possible

I heard about this many many years ago. Is this still the case, for constants, at this day and age?

@JordanHendersonMusic this looks like a useful approach. Are there any tradeoffs for using this flag?

capital-G: > Shoul we enable `-freciprocal-math`?
> 
> https://godbolt.org/z/d3PMzrdrP

I don't see a difference in the assembly with this turned on or off.

~~I also tested this with the quick-bench and so no differences with o3 or non optimizations turned on~~
~~https://quick-bench.com/q/iNuYm53x7gglWHb49cXTTa4ytyI~~

edit: There is actually a difference - multiply is around 1.3 times as fast than division.

https://quick-bench.com/q/rxH578wIqFrcK06qW40B0Na2Ylg

JordanHendersonMusic: > @JordanHendersonMusic this looks like a useful approach. Are there any tradeoffs for using this flag?

I don't think so, at least, not if the operation was well defined... if the tests pass, I'd be inclined to say it is fine, at least in the Lang, on the audio server, I'm less sure - but we don't deal with infs and nans other there so perhaps it isn't an issue?

dyfer: > I don't see a difference in the assembly with this turned on or off.

@capital-G I can see the assembly command changing from `mulss` to `divss` when the reciprocal math flag is removed.

~~BTW I don't think we use global optimization (-O3) in SC. Is this somthing to consider (maybe a lower one)?~~ Sorry for the noise, I guess this is set implicitly by the build type in cmake.

JordanHendersonMusic: I **think** CMake sets -O3 when you ask for a release build?

dyfer: Yes! I didn't know - I just looked it up now.

PR #6712, opened 2025-04-08
Remove RFC section from README

capital-G:
## Purpose and Motivation
In the last dev meeting it was decided that we should phase out the RFC procedure.
A first step towards this would be to

* add a deprecation note on the RFC repo -> https://github.com/supercollider/rfcs/pull/23
* remove it from the main repo `README.md` (this PR)

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation


dyfer:
Don't we need an RFC to agree on this?
j/k

dyfer:
On a more serious note, should we add some wording on how the larger changes are discussed? Maybe that's not needed, I'm not sure.

capital-G:
> On a more serious note, should we add some wording on how the larger changes are discussed? Maybe that's not needed, I'm not sure.

I think it is implied on most projects that changes are proposed via issues and implemented via PRs. More drastic changes would require asking our community, but we don't have those too often and most probably someone of the reviewers would propose to ask a broader audience at some point.

Don't think we would need to codify these - if we get some feedback that this implicitness is unclear we can change it, but I think we can simply skip this now.

dyfer (approved):

PR #6714, opened 2025-04-09
docs: fix em dash, typo, update urls to https

Staxxator:
## Purpose and Motivation
doc fix:
fixed em-dashes, spaces, and updated URLs to https://... in "WritingUGens.schelp"

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Guides/WritingUGens.schelp
+++ b/HelpSource/Guides/WritingUGens.schelp
@@ -24,7 +24,8 @@ SuperCollider code that abbreviates a certain configuration of UGens that gets u
 that superficially resembles a UGen class, but it only returns a composition of existing UGens. It has no efficiency
 savings, but it does save typing.
 
-The below example has only a code::.ar:: method, but you can just as easily have both code::.ar:: and code::.kr:: methods.
+The below example has only a code::.ar:: method, but you can just as easily have both code::.ar:: and code::.kr:: 
+methods.
 
 code::
 BoringMixer {
@@ -106,7 +107,8 @@ architecture as well as a bunch of helper functions. These are documented at lin
 
 subsection:: The Entry Point
 
-When the library is loaded the server calls a function in the library, which is defined by the code::PluginLoad():: macro.  This entry point has two responsibilities:
+When the library is loaded the server calls a function in the library, which is defined by the code::PluginLoad():: 
+macro. This entry point has two responsibilities:
 
 list::
 ## It needs to store the passed in pointer to the InterfaceTable in a global variable.
--- a/HelpSource/Guides/WritingUGens.schelp
+++ b/HelpSource/Guides/WritingUGens.schelp
@@ -12,7 +12,7 @@ SuperCollider language. UGens exist more or less independently of the interprete
 with SC to write them.
 
 Writing UGens is not too difficult, but it's arguably far less convenient and intuitive than the high-level tools that
-SuperCollider provides. You'll need a build system and a good amount of boilerplate code -- even fairly basic signal
+SuperCollider provides. You'll need a build system and a good amount of boilerplate code — even fairly basic signal
 processing operations can require a lot of code. You don't have an instant live coding environment, and mistakes can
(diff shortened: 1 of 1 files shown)


HelpSource/Guides/WritingUGens.schelp
```diff
@@ -24,8 +24,7 @@ SuperCollider code that abbreviates a certain configuration of UGens that gets u
 that superficially resembles a UGen class, but it only returns a composition of existing UGens. It has no efficiency
 savings, but it does save typing.
 
-The below example has only a code::.ar:: method, but you can just as easily have both code::.ar:: and code::.kr::
-methods.
+The below example has only a code::.ar:: method, but you can just as easily have both code::.ar:: and code::.kr:: methods.
```
dyfer, line 27: I don't think we need unwrapping the line here, unless I'm missing something?

HelpSource/Guides/WritingUGens.schelp
```diff
@@ -107,8 +106,7 @@ architecture as well as a bunch of helper functions. These are documented at lin
 
 subsection:: The Entry Point
 
-When the library is loaded the server calls a function in the library, which is defined by the code::PluginLoad()::
-macro.  This entry point has two responsibilities:
+When the library is loaded the server calls a function in the library, which is defined by the code::PluginLoad():: macro.  This entry point has two responsibilities:
```
dyfer, line 109: Also I don't believe we should unwrap the lines - but the double space before "This" could be removed.

dyfer (changes requested):
Thanks for the PR!

I don't think we should unwrap the lines. Could you please revert that change?

Also, what is the typo you are referring to? Can't see it on the first glance.

Staxxator:
Hi, I'm sorry, but I can't push anymore, the authentication always fails.
The term 'typo' is maybe not the correct one, what I found was a missing space after the .kr, and found that the following word was in the next line, so I put it on the upper one. Sure, there's no need for that unwrapping.
I'm doing all that for the first time, and wondering why the authentication always fails, even via SSH key I can't push.

dyfer:
> missing space after the .kr,

Ah I see now! Sure, that's good to fix. But let's keep the line break. And if possible, please fix the double space further down.



> I can't push anymore, the authentication always fails.

What does your `git remote -v` say?

Staxxator:
yes, I've already fixed it, the only problem now is that I can't push it anymore.

git remote -v
says:
origin	https://github.com/Staxxator/supercollider (fetch)
origin	https://github.com/Staxxator/supercollider (push)
upstream	https://github.com/supercollider/supercollider (fetch)
upstream	https://github.com/supercollider/supercollider (push)

and when I type
git push --set-upstream origin topic/WritingUnitGenerators-typo

I'm asked for my user name and password. After entering those there's the message:

remote: Support for password authentication was removed on August 13, 2021.
remote: Please see https://docs.github.com/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.
fatal: Authentication failed for 'https://github.com/Staxxator/supercollider/'

I also tried with ssh key but that didn't work either; the access to github was denied.

I think I'm going to need to create a new ssh key, create a new fork, clone and do everything once again and do a new pull request.

Do you know if pushing via https is deprecated?

dyfer:
> Do you know if pushing via https is deprecated?

Yes, kind of. At least https with the regular username/password: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github#https

I believe you need to change your remotes to use ssh. See https://docs.github.com/en/get-started/git-basics/managing-remote-repositories#changing-a-remote-repositorys-url and the section `If you're updating to use SSH...`

Staxxator:
Hi, I just fixed the problem and pushed it. So, I don't need to make a new pull request, do I?

dyfer:
> So, I don't need to make a new pull request, do I?

Correct, no need for a new PR.

I'll squash the commits on merging.

dyfer (approved):
Thanks!

PR #6715, opened 2025-04-11
Classlib: PlotView.sc: ArrayedCollection.plot should convert wavetables to Signals before passing to Plotter

HotwheelsSisyphus:
Wavetable.plot internally calls Wavetable.asSignal.plot because Wavetables have a special format. The case of Arrays of Wavetables was overlooked, this commit fixes that.
Also replaces the nested ifs with case for clearer conditional structure. 

## Purpose and Motivation

Old behavior:
```
u = Wavetable.sineFill(1024, [1]);
u.plot; //single Wavetable works fine (picture on left)
u.dup(2).plot; //array of wavetables doesn't. (picture on right).
```
Capture d’écran, le 2025-04-11 à 12 30 10

New behavior: The expected one (i.e., sines not zigzags back and forth to 0).

## Types of changes

- Bug fix

Tested with all the examples from https://github.com/supercollider/supercollider/pull/6566 to ensure I didn't break anything by rearranging the conditional.

- [x] Code is tested
- [x] All tests are passing

Noticed along the way: superpose doesn't work with Arrays of envelopes. Is this known behavior?

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -1312,18 +1312,11 @@ Plotter {
 
 		numChannels !? { array = array.unlace(numChannels) };
 		array = array.collect {|elem, i|
-			if (elem.isKindOf(Env)) {
-				elem.asMultichannelSignal.flop
-			} {
-				if(elem.isNil) {
-					Error("Cannot plot array: non-numeric value at index %".format(i)).throw
-				};
-				if(hasSubArrays) {
-					elem.asArray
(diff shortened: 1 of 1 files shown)


dyfer (approved):
Tested, works. Thanks!

dyfer:
> Noticed along the way: superpose doesn't work with Arrays of envelopes. Is this known behavior?

I don't think so. Probably not a common use case - I don't see an issue filed for it.

telephon:
> Noticed along the way: superpose doesn't work with Arrays of envelopes. Is this known behavior?

It is currently not supported. You can have envelopes with arrayed levels and times only ...

Issue #4281, opened 2019-01-30
scsynth fails to boot on Mac without hardware inputs, supernova works


mtmccrea:
## Environment

* SuperCollider version: 3.10.1
* Operating system: macOS Mojave 10.14.2
* Other details: New Mac mini, No default audio device with input streams (output only)

## Steps to reproduce

```supercollider
// State of HW devices:
ServerOptions.devices
// >> [ DisplayPort, Mac mini Speakers ]
ServerOptions.inDevices
// >> [ ]
ServerOptions.outDevices
// >> [ DisplayPort, Mac mini Speakers ]

// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
// Simple boot out of the gate
s.boot;
/*
-> localhost
Booting server 'localhost' on address 127.0.0.1:57110.
Found 0 LADSPA plugins
Number of Devices: 2
   0 : "DisplayPort"
   1 : "Mac mini Speakers"

get kAudioDevicePropertyStreamFormat error on input ?ohw
could not initialize audio.
Server 'localhost' exited with exit code 0.
*/
s.options.asOptionsString
// ->  -u 57110 -a 1024 -i 2 -o 2 -R 0 -l 1

// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(
s.options.numInputBusChannels_(0);
s.reboot;
)
// >> same error
s.options.asOptionsString
// ->  -u 57110 -a 1022 -i 0 -o 2 -R 0 -l 1

// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(
s.options.numInputBusChannels_(0);
s.options.inDevice_(nil);
s.reboot;
)
// -> same error
s.options.asOptionsString
//->  -u 57110 -a 1022 -i 0 -o 2 -R 0 -l 1

// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
thisProcess.recompile;
// ...
(
s.options.numInputBusChannels_(0);
s.options.device_("Mac mini Speakers");
s.reboot;
)
// >> same error

[...]

PR #6716, opened 2025-04-11
CoreAudio backend: don't try to query input device when numInputBusChannels = 0

### LINKED ISSUE #6203: SuperCollider can't boot on Aggregate Device with No Input selected
## Environment

* SuperCollider version:  3.12.2
* Operating system: Mac OS 14.2.1 (23C71)
* Other details (Qt version, audio driver, etc.): MacBook M2

## Steps to reproduce

1. Create Aggregate Device in Audio Midi Setup (optional: called "Supercollider")
2. Select an output device, leave the input device unselected
3. Run SC server with startup file containing

```supercollider
Server.default.options.numInputBusChannels  = 0;
Server.default.options.outDevice = "Supercollider";
```

Supercollider's synth server should run with no input device available, expecting a silent input from In.ar or other input ugens.

Instead, the server wouldn't even start. 
Forcing the user to open an input audio channel, even if not needed in the specific use case. 

This happens even if

```supercollider
Server.default.options.numInputBusChannels  = 0;
Server.default.options.outDevice = "Supercollider";
```

This behavior differs from PortAudio's

https://github.com/supercollider/supercollider/issues/844

dyfer:
## Purpose and Motivation

I believe we shouldn't try to get properties of an input device when `numInputBusChannels = 0`, as some devices have no inputs.

Fixes #6203 and I think #4281

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/server/scsynth/SC_CoreAudio.cpp
+++ b/server/scsynth/SC_CoreAudio.cpp
@@ -870,7 +870,7 @@ bool SC_CoreAudioDriver::DriverSetup(int* outNumSamplesPerCallback, double* outS
         return false;
     }
 
-    if (mInputDevice != kAudioDeviceUnknown) {
+    if ((mInputDevice != kAudioDeviceUnknown) && (mWorld->mNumInputs > 0)) {
         // get a description of the data format used by the input device
         count = sizeof(AudioStreamBasicDescription);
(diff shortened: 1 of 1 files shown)


telephon (approved):

dyfer:
I consider this to be non-contentious. If any issues arise, we can revert.

PR #4871, opened 2020-04-13
Add example to mismatched sample rate error

pearcemerritt:
## Purpose and Motivation

As someone familiar to code, but not SC, it took me quite a while to figure out how set the number of input channels to 0 like the existing error message suggests. Figured this example might help someone else.

It assumes the `s` server variable, which I could imagine might not be applicable in a more advanced set-up, but I also imagine that someone working on a more advanced set-up probably would be familiar with `ServerOptions` to the point where they wouldn't need this example, or would know how to apply it to whatever `Server` scheme they had.

## Types of changes

- Documentation
- Bug fix
- New feature

PR #6717, opened 2025-04-11
CoreAudio: try setting sample rate of output device when input and output sample rates are mismatched

dyfer:
## Purpose and Motivation

A common setup on macOS is to use built-in speakers and built-in mic. These devices may be set to different sample rates.

Instead of bailing out on such configuration, let's try to actually set the input to match the output.

EDIT: now we additionally **check if the input device is already open**, and if it is, we post an error with a slightly different error message instead of changing its sample rate and then exit. This should prevent the accidental interruption of e.g. recording happening in the background, while still provide the desired functionality when no audio capture is running.

To test:
- (quit the server)
- in Audio MIDI Setup set built-in mic and built-in output as default audio devices
- in Audio MIDI Setup set built-in mic to a _different_ sample rate than the output
- start the server with default settings (no devices or sample rate set); `s.boot;`

This will still fail if the input device's sample rate cannot be set to match the output. EDIT: additionally, **we error out if the input device is in use**.

Surprisingly I haven't found the issue filed for this, but I have heard this issue mentioned on occasion by other users. I believe this will greatly improve the user experience on macOS!

Here's the output if the SR setting succeeded:
```
Found 0 LADSPA plugins
Number of Devices: 24
(...)

WARNING: Input sample rate is 96000, but output is 48000. Attempting to set input sample rate to match the output.
   Streams: 1
      0  channels 1

"Komplete Audio 6 MK2" Output Device
   Streams: 1
      0  channels 6

SC_AudioDriver: sample rate = 48000.000000, driver's block size = 512
SuperCollider 3 server ready.

```
If the input device did not support the SR of the output device:

```
Found 0 LADSPA plugins
Number of Devices: 24
(...)

WARNING: Input sample rate is 48000, but output is 192000. Attempting to set input sample rate to match the output.
ERROR: Setting sample rate failed. OSStatus tahw
Possible solutions:
- explicitly set the sample rate to one supported by both devices:
    s.options.sampleRate = <rate>;
- or, in your system's "Audio MIDI Setup", set sample rate to the same value on both the input and output devices
- or, disable input completely:
    s.options.numInputBusChannels = 0;
could not initialize audio.
```

If the input device is in use (**note a slightly different message**, aiming to prioritize not interrupting the already open stream on the input device):
```
Found 0 LADSPA plugins
Number of Devices: 24
(...)

ERROR: Input sample rate is 96000, but output is 48000. Input device is in use - we won't try changing its sample rate to match the output.
Possible solutions:
- in your system's "Audio MIDI Setup", set sample rate to the same value on both the input and output devices
WARNING: this may interrupt audio capture of another application currently using the input device!
- or, disable input completely:
    s.options.numInputBusChannels = 0;
could not initialize audio.
```

**Question:** is the warning posting reasonable here? Maybe that should only happen if verbosity is increased?

I'd appreciate any feedback on this!

EDIT: Added posting similar to the one suggested previously by @mossheim: https://github.com/supercollider/supercollider/pull/4871#issuecomment-632678267

~~EDIT 2: This PR now also waits for the sample rate to be applied, but only in the case of mismatched SRs. The new functions introduced to wait for the changes to be applied should be used elsewhere when we set SR, but I want to avoid scope creep and keep that for another PR.~~

EDIT 3: this PR does _not_ wait to apply the sample rate change, but it will open both devices at the sample rate supported by both.

## Types of changes

- New feature (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/server/scsynth/SC_CoreAudio.cpp
+++ b/server/scsynth/SC_CoreAudio.cpp
@@ -885,17 +885,58 @@ bool SC_CoreAudioDriver::DriverSetup(int* outNumSamplesPerCallback, double* outS
             return false;
         }
 
-        if (
-            // We do not support mismatched input and output sample rates, but there's no harm in skipping
-            // this check if numInputBusChannels == 0. This allows the user to disable input entirely as
-            // a workaround for a sample rate mismatch.
-            mWorld->mNumInputs > 0 && !mExplicitSampleRate
-            && inputStreamDesc.mSampleRate != outputStreamDesc.mSampleRate) {
-            scprintf("ERROR: Input sample rate is %g, but output is %g. "
-                     "Mismatched sample rates are not supported. "
-                     "To disable input, set the number of input channels to 0.\n",
+        if (!mExplicitSampleRate && inputStreamDesc.mSampleRate != outputStreamDesc.mSampleRate) {
+            UInt32 dataSize;
+
+            // Check if the input device is in use
+            UInt32 inUse = 0;
+            dataSize = sizeof(inUse);
+
+            propertyAddress.mSelector = kAudioDevicePropertyDeviceIsRunningSomewhere;
+            propertyAddress.mScope = kAudioDevicePropertyScopeInput;
+
+            OSStatus status = AudioObjectGetPropertyData(mInputDevice, &propertyAddress, 0, nullptr, &dataSize, &inUse);
+
+            if (status == noErr && inUse) {
+                scprintf("ERROR: Input sample rate is %g, but output is %g. "
+                         "Input device is in use - we won't try changing its sample rate to match the output.\n"
+                         "Possible solutions:\n"
+                         "- in your system's \"Audio MIDI Setup\", set sample rate to the same "
+                         "value on both the input and output devices\n"
+                         "WARNING: this may interrupt audio capture of another application "
+                         "currently using the input device!\n"
+                         "- or, disable input completely:\n"
+                         "    s.options.numInputBusChannels = 0;\n",
+                         inputStreamDesc.mSampleRate, outputStreamDesc.mSampleRate);
+                return false;
+            }
+
+            scprintf("WARNING: Input sample rate is %g, but output is %g. "
+                     "Attempting to set input sample rate to match the output.\n",
                      inputStreamDesc.mSampleRate, outputStreamDesc.mSampleRate);
-            return false;
+
+            auto sampleRate = outputStreamDesc.mSampleRate;
+            dataSize = sizeof(sampleRate);
+
+            // Set the same rate on the input device
+            propertyAddress.mSelector = kAudioDevicePropertyNominalSampleRate;
+            propertyAddress.mScope = kAudioDevicePropertyScopeInput;
+
+            err = AudioObjectSetPropertyData(mInputDevice, &propertyAddress, 0, NULL, dataSize, &sampleRate);
+            if (err != noErr) {
+                scprintf("ERROR: Setting sample rate failed. OSStatus %4.4s\n"
+                         "Possible solutions:\n"
+                         "- explicitly set the sample rate to one supported by both devices:\n"
+                         "    s.options.sampleRate = <rate>;\n"
+                         "- or, in your system's \"Audio MIDI Setup\", set sample rate to the same "
+                         "value on both the input and output devices\n"
+                         "- or, disable input completely:\n"
+                         "    s.options.numInputBusChannels = 0;\n",
+                         (char*)&err);
+                return false;
+            }
+            // set mExplicitSampleRate so that we open both input and output with the new sample rate further down
+            mExplicitSampleRate = sampleRate;
         }
     }
     ////////////////////////////////////////////////////////////////////////////////////////////////


server/scsynth/SC_CoreAudio.cpp
```diff
@@ -891,11 +891,27 @@ bool SC_CoreAudioDriver::DriverSetup(int* outNumSamplesPerCallback, double* outS
...
+            AudioObjectPropertyAddress propertyAddress;
+
+            propertyAddress.mSelector = kAudioDevicePropertyNominalSampleRate;
+            propertyAddress.mScope = kAudioDevicePropertyScopeInput;
+            propertyAddress.mElement = kAudioObjectPropertyElementMaster;
+
+            err = AudioObjectSetPropertyData(mInputDevice, &propertyAddress, 0, NULL, dataSize, &sampleRate);
+            if (err != noErr) {
+                sampleRateSetSuccess = false;
+                scprintf("set input sample rate error %4.4s\n", (char*)&err);
```
dyfer, line 913: On thing I wonder about, should we `return false;` here?

dyfer: The answer is yes :) otherwise we open a devices at different sample rates...

telephon (approved):

capital-G:
I am somehow hesitant of applications setting the sample rate of my devices b/c they are always introducing outliers to my "48khz everywhere reign" - even more so if this is implicitly done.

Couldn't we error out on sclang level and print the code for people on how to set this via sclang?
Something like `s.outDevice.sampleRate_(48000)` would be awesome (it seems it would have been beneficial to have an `AudioDevice` class within sclang instead of all these numbers/strings lying next to each other in `ServerOptions`?)
But the problem seems that the language does not have the necessary primitives, only the server has those IIRC? But specifying those explictily via CLI on boot would also be possible (once #6640 is merged I may want to do the same for scsynth so we can introduce CLI flags more easily for scsynth/supernova).

Or maybe we can add a (boolean) flag to the `boot` method so there is still a way that scsynth doesn't mess with the audio settings?

Sorry for the nagging^^

dyfer:
@capital-G I understand hesitation re: changing samplerate automatically. However, I think the default behavior is designed to "just work"... when it isn't now, when the devices are at different SRs.

> my "48khz everywhere reign"

If you want 48k always, you can put `Server.default.options.sampleRate_(48000)` in your startup file. Would that work for you?

> Something like `s.outDevice.sampleRate_(48000)` would be awesome

I don't understand. We have `s.options.sampleRate_(x)` and this will set the samplerate for both input and output device.

> Couldn't we error out on sclang level and print the code for people on how to set this via sclang?

That's what we do so far! But that's arguably annoying, when you just wanted to boot the server to hear some sound, but instead you need to either disable inputs or set sample rate.

It is true that `ServerOptionw` cannot check for sample rate. At the same time, the whole device selection process would ideally be reworked to provide a tighter integration with sclang, which would be a larger task.

Being able to disable this behavior via CLI (maybe na env var? that's a bit hidden) might be desirable... Similarly, maybe a flag should also be there to "fail if the device is not available". Currently, if specified device is not available, server boots with the default device on macOS. Maybe the same flag could control these behaviors.

Maybe this is something or another PR? What do you think?

Spacechild1:
I'm not really a Mac users so I can't say too much. Just one thing: does this issue also exist with the portaudio backend on Supernova?

dyfer:
> does this issue also exist with the portaudio backend on Supernova?

Good question! I remember tweaking PA behavior on macOS, but I couldn't remember this issue.

After testing, here's the behavior:
- when devices have different sample rates, PA backend tries to set the input device to match the output (behavior matching the behavior in this PR)
- is the input device does not support the SR of the output, then an interesting thing happen: an error is shown, but somehow resampling kicks in and the stream opens; audio passing from input to output seems clean, so I assume that there must be resampling kicking in somewhere (in PA probably).

Output:
<details>

```
Selecting default system input/output devices

Opening audio devices:
  In: MacBook Pro Microphone
  Out: Komplete Audio 6 MK2
||PaMacCore (AUHAL)|| Error on line 520: err=''what'', msg=Unspecified Audio Hardware Error
||PaMacCore (AUHAL)|| Error on line 520: err=''what'', msg=Unspecified Audio Hardware Error
  Sample rate: 192000.000
  Latency (in/out): 0.122 / 0.025 sec
Audio devices opened successfully
Supernova ready
localhost: no maxLogins info from server process.
```

In this example the Microphone device was set to 96k (highest supported SR). Audio passing in SC seemed clear (no audible buffer under/overruns).
</details>

BTW I was thinking about writing a CoreAudio backend for supernova, so that the behaviors are consistent (and we get rid of the extra 64 sample (?) latency which I think I measured at some point, comparing in-to-out latency of scsynth and supernova on macOS).

@capital-G another point about managing device choices - from the client/server point of view, all audio device management should be done through the server; it's nice that we have ServerOptions, but that's client-specific. We already have a slight mismatch because supernova/portaudio sometimes lists devices differently than coreaudio (this might've been an obsolete issue though). In any case, I imagine a cleaner solution to have a command line switch for the servers that queries audio device and returns a JSON or similar with all the parameters, allowing any client implementing a device chooser etc.

capital-G:
Can you check the implications of changing the sample rate of an device: Let's say I use my internal mic to record at 44.1 kHz in Reaper but now SC comes along and tells the mic to switch to 48 kHz - will the sample rate switch bail out or will Reaper bail out at this moment (latter would be non-acceptable IMO).

What would your opinion be on booting but omitting the input device if there is a sample rate mismatch? This is also implicit behavior, but at least we don't change system settings and affect other applications.

In the end if there is a consensus to implement the implicit switch I'd be fine, but I'd really appreciate it if this could be switched off with a flag on the boot method.

> If you want 48k always, you can put `Server.default.options.sampleRate_(48000)` in your startup file. Would that work for you?

I personally don't like startup files, I want to always keep my SC setup as explicit and vanilla as possible b/c I often scratch my head b/c of the implicit behavior everywhere (especially since we don't have a logging system or debugger which makes it really hard to reason about the state of the app).

> In any case, I imagine a cleaner solution would be to have a command line switch for the server(s) that queries audio devices and returns a JSON or similar with all the parameters (including support SRs etc), allowing any client to implement a device chooser or similar.

This would be indeed a great thing to have! Lets introduce this after we switched to program options for scsynth, I think supernova is already using program options.
Would probably be possible to implement this once and just share it between supernova and scsynth.

<details>
<summary>JSON</summary>
But we also need to have proper JSON support in sclang, there are still some strange concerning `null` and we need a proper JSON encoder built in (see e.g. https://github.com/musikinformatik/JSONlib)

```
j = "{\"a\": null, \"b\": 42}".parseJSON
// a is a key with a nil value - this is normally not possible in sclang!
j.asDict // now a is gone :p
```
</details>

dyfer:
> Can you check the implications of changing the sample rate of an device: Let's say I use my internal mic to record at 44.1 kHz in Reaper but now SC comes along and tells the mic to switch to 48 kHz - will the sample rate switch bail out or will Reaper bail out at this moment (latter would be non-acceptable IMO).

Interestingly, it's neither. The new sample rate is set and _Reaper follows that change_, adjusting its processing SR.
However, the same happens when we explicitly request a sample rate when booting the server - the device is switched to the new SR and Reaper doesn't complain, just adjusts its processing to use the new SR. This talks more to Reaper's permissiveness, than SC behavior though.

Also, here's Reaper's behavior when it's set to _not requesting a particular sample rate_ and useing two different devices (e.g. built in mic and speakers) that are set to different SRs: **Reaper will set the output device's SR to match the input**. Essentially that's the same behavior as proposed in this PR, but prioritizing the input (which makes sense for a DAW IMO).

> What would your opinion be on booting but omitting the input device if there is a sample rate mismatch? This is also implicit behavior, but at least we don't change system settings and affect other applications.

The problem is that we always request a fixed number of inputs/outputs, so booting to something else than requested is problematic. Back in the day when the default number of ins/outs in SC was being changed from 8 to 2, I was wondering if we could also have a "default" for this, which would be the number of I/Os of the audio interface. However, the server has no way of updating the language with the number of IOs it is using. Server side depends on a set number of I/Os for the bus allocator. This could possibly be changed so that the server communicates it, but would be likely cumbersome. Having an interface for the server to return device info as json could also enable having "default number of ins/outs" as an option, if we wanted.
Currently, when we do request a number of inputs, I don't think disabling inputs would be an acceptable behavior, BUT OTOH maybe from the practical standpoint that would be fine?


> In the end if there is a consensus to implement the implicit switch I'd be fine, but I'd really appreciate it if this could be switched off with a flag on the boot method.

I can see how adding this could be useful. But how would we call such a flag? I think this only makes sense for a long option though (so I'm not sure if we'd choose an arbitrary letter and change it later, or should that be done in a separate PR after long options are added, OR if this PR should wait for the addition for long options). The ideas I have include: `--no-automatic-samplerate-changes` or `--dont-match-input-samplerate-automatically` or '--dont-match-input-samplerate-to-default-output`....


@capital-G I appreciate your input on this. I'm curious though, wasn't the current behavior (the server can't boot because the mic is set to a different SR) ever a hindrance to you?

dyfer:
> Can you check the implications of changing the sample rate of an device: Let's say I use my internal mic to record at 44.1 kHz in Reaper but now SC comes along and tells the mic to switch to 48 kHz

BTW this would only come up, if Reaper used internal mic + some output device (both running at 44.1k - Reaper will switch both devices to run at the same SR), but you'd asked SC to boot using internal mic + _a different output device, which happens to be running at 48k_. This is totally a possibility, but at that point this is already an advanced use case IMO. 

If SC is using the same devices as Reaper, it would not ask it to switch SR, since they would both run at the same SR already.

dyfer:
After the discussion at the last dev meeting, this PR now changes the _output_ device SR to match the input, to avoid possible interference with any recording that might be happening at the time.

Additionally, I realized that the fact that we were neither waiting for SR to be applied, nor updating the `AudioStreamBasicDescription` caused this not to work properly. I now added the functions to handle this, but I have limited experience with this code so please review carefully! Thanks!

dyfer:
@capital-G @telephon this PR is now updated. Please review if possible, especially the added functionality to wait for the SR changes to be applied.

dyfer:
I'll update this PR with the suggestions after the dev meeting

dyfer:
At the last dev meeting we discussed reverting the behavior to set the input device to match the output (to address the situation when bluetooth headset is set as an input). 
Also @dyfer will reach out to Reaper to discuss whether the possible recording disruption could be avoided.

dyfer:
@capital-G I added a check whether a device is already running before checking its sample rate. I feel pretty confident about this functionality now: if SR is mismatched, we check if the input device is already open; if it is, we fail with a message (slightly different than the suggested one, aimed at not disturbing open stream); if it isn't we try to **set the input** device to match the output. If we succeed, we boot; if we don't, we post a message suggesting a solution.

What do you think?

capital-G (approved):
Thanks dyfer!

The new implementation looks good to me - feel free to either merge it or request additional reviews :)

dyfer:
Thanks! I feel quite confident about this between my tests and tje discussion at the last meeting so I'll merge.

PR #6720, opened 2025-04-12
Change class name `BufferExamplesPath` to `ExampleFiles`

tedmoore:
## Purpose and Motivation

This is a follow up from the conversation on https://github.com/supercollider/supercollider/pull/6685 about changing this class name. The conversation came the idea that `BufferExamplesPath` might be misleading in that the class points to soundfiles rather than "buffers", also that `ExampleFiles` was more generic and could be extended in the future to a variety of file types.

## Types of changes

- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Buffer.schelp
+++ b/HelpSource/Classes/Buffer.schelp
@@ -1,7 +1,7 @@
 class:: Buffer
 summary:: Client-side representation of a buffer on a server
 categories:: Server>Abstractions
-related:: Classes/BufferExamplesPath
+related:: Classes/ExampleFiles
 
 description::
 
--- a/HelpSource/Classes/ExampleFiles.schelp
+++ b/HelpSource/Classes/ExampleFiles.schelp
@@ -1,4 +1,4 @@
-TITLE:: BufferExamplesPath
+TITLE:: ExampleFiles
 summary:: A shortcut to example sounds bundled with SuperCollider
 categories:: Files
 related:: Classes/Buffer, Classes/Soundfile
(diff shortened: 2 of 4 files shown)


capital-G (approved):
Seems like a good change to me. Auto completion shouldn't be the reason to name things, `ExampleFiles` are much clearer to what they represent.

But I'd like another opinion on this, so another reviewer can merge :)

prko:
In my opinion `ExampleFiles` could be audio files, SCD files, SC files, SCHELP files etc.

So I would like to suggest the following alternatives:
- `SoundFiles`
- `SoundFileExamples` 
- `ExampleSoundFiles` 
- `AudioFiles`
- `AudioFileExamples` 
- `ExampleAudioFiles`.

tedmoore:
The thinking of the more generic name was in case there were files used in examples that were not just sound files. There could be a case for having some other files available such as `.json`, `.sdif`, `.csv`, `.png`, `.svg`. These could a live in the same folder and be accessible by the same class.

dyfer:
I'm also in favor of ExampleFiles.

I don't think having many more example files classes would be useful in our case. This is just a shortcut for examples...

dyfer (approved):

tedmoore:
Is the protocol to delete branches after merging?

tedmoore:
Ah, is the usual protocol to make PRs from a personal fork and not push branches here?

dyfer:
Yes, that's the usual protocol. The [wiki](https://github.com/supercollider/supercollider/wiki/Setting-up-your-development-environment) has more details.

PR #6725, opened 2025-04-13
Document keystate usage on macOS

capital-G:
## Purpose and Motivation

Compare 

```
Ndef(\k, {KeyState.kr((0..127), lag: 0.0).sum.poll})
```

with and without Security Settings applied - only if Input monitoring is allowed keys like `abc` will be received by SuperCollider, otherwise it will only capture the arrow keys etc.

This PR documents that keystate usage is tied to Security settings on macOS.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/KeyState.schelp
+++ b/HelpSource/Classes/KeyState.schelp
@@ -19,6 +19,11 @@ The UGen will return MATH::0.0:: as constant value.
 For more information see https://github.com/supercollider/supercollider/issues/4544
 ::
 
+WARNING::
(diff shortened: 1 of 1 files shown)


telephon (approved):

adcxyz:
Nice, thanks! 
Interestingly, on macOS the keyboard is not very polyphonic: 
the maximum number of simultaneously pressed keys is 6.

PR #6727, opened 2025-04-13
sclang: exit if the library has not been compiled

### LINKED ISSUE #5218: Non-interactive scripts hang rather than aborting on compilation failures
[Originally submitted as #3393, but that turned into a much wider discussion, so refiling a more specific issue as per [this request](https://github.com/supercollider/supercollider/issues/3393#issuecomment-716069937).]

If I have a file `hello.scd` containing:

    "Hello, world!".postln;
    0.exit;

and I invoke `sclang hello.scd` from the CLI, it will output a bunch of stuff ending in:

```
*** Welcome to SuperCollider 3.11.1. *** For help type ctrl-c ctrl-h (Emacs) or :SChelp (vim) or ctrl-U (sced/gedit).
Hello, world!
cleaning up OSC
```

and then it exits cleanly.  So far so good.

However if there is a compilation error whilst compiling the classes or the script, apparently there is no way to avoid it hanging other than wrapping `sclang` in a timeout.  For example:

```
*** Welcome to SuperCollider 3.11.1. *** For help type ctrl-c ctrl-h (Emacs) or :SChelp (vim) or ctrl-U (sced/gedit).
ERROR: syntax error, unexpected NAME, expecting $end
  in interpreted text
  line 1 char 14:

  "Hello, "world!".postln;
           ^^^^^
  0.exit;
```

and it hangs at this point.

Similarly, if I deliberately break `File.sc` in the library, I get:

```
compiling class library...
        Found 856 primitives.
        Compiling directory '/home/adam/.STOW/supercollider/share/SuperCollider/SCClassLibrary'
        Compiling directory '/home/adam/.STOW/supercollider/share/SuperCollider/Extensions'
        Compiling directory '/home/adam/.local/share/SuperCollider/Extensions'
ERROR: syntax error, unexpected CLASSNAME
  in file '/home/adam/.STOW/supercollider/share/SuperCollider/SCClassLibrary/Common/Files/File.sc'
  line 287 char 0:

  }
  ^^^^^^^^
ERROR: file '/home/adam/.STOW/supercollider/share/SuperCollider/SCClassLibrary/Common/Files/File.sc' parse failed
error parsing
Library has not been compiled successfully.
```

and again the process hangs forever.

This is particularly problematic for [`guard-sclang`](https://github.com/aspiers/guard-sclang) which needs to spot when the developer has introduced (hopefully transient) compilation errors into the classes - currently wrapping `sclang` in a timeout seems to be the only way to do that.  This issue is being tracked in https://github.com/aspiers/guard-sclang/issues/2.

Very preliminary analysis of `lang/LangSource/SC_TerminalClient.cpp` seems to suggest that it calls `commandLoop()` regardless of which options are chosen, e.g. `-D` for daemon mode has no effect on this, so that might explain why it's always hanging.

## Environment

* SuperCollider version: 3.11.1 compiled from latest `develop` (291dd1a3727cb23b6b06dd32276256300b5d0e39)
* Operating system: openSUSE Tumbleweed

## Expected behavior

Non-interactive scripts immediately quit with non-zero exit status if the library or script fail to compile.

dyfer:
## Purpose and Motivation

Currently, when the library compilation fails, we get stuck in a limbo; the interpreter doesn't quit, but also doesn't operate. This makes it impossible to react to such failure programmatically.

I believe this would be an improvement for testing etc., among possible future uses.

~~This behavior can be disabled by setting an environment variable `SCLANG_NO_EXIT_ON_FATAL_ERROR`, borrowed from #6671. I'm open to suggestions whether we really need an env var, or if it should be called something else. I thought it would be easiest if it's the same as in the other PR, but that can also change.~~

~~I've chosen `1` as the exit code. Another option would be `-1` I guess (IIUC that would overflow to `255`?).~~ EDIT: I changed the exit code to `EXIT_FAILURE` macro from `<cstdlib>`, but ideally we'd rework exit codes project-wide - see https://github.com/supercollider/supercollider/issues/6728.

I'm open to any feedback on this!

Fixes #5218

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation - TBD... I feel like the env vars should be documented somewhere.

### DIFF
--- a/lang/LangSource/SC_TerminalClient.cpp
+++ b/lang/LangSource/SC_TerminalClient.cpp
@@ -25,6 +25,7 @@
 */
 
 #include "SC_TerminalClient.h"
+#include <cstdlib>
 #ifdef SC_QT
 #    include "../../QtCollider/LanguageClient.h"
 #endif
@@ -259,6 +260,14 @@ int SC_TerminalClient::run(int argc, char** argv) {
 
     // startup library
     compileLibrary(opt.mStandalone);
+
+    if (!compiledOK) {
+        post("ERROR: Library has not been compiled successfully.\n");
+        shutdownLibrary();
+        flush();
+        shutdownRuntime();
+        return EXIT_FAILURE;
+    }
 
     // enter main loop
     if (codeFile)


telephon (approved):
What a good thing, after so many years! Looks good to me.

capital-G (changes requested):
Thats an awesome idea! This will make the use of sclang as an interpreter much more reliable in a remote scenario :)

> This behavior can be disabled by setting an environment variable `SCLANG_NO_EXIT_ON_FATAL_ERROR`, borrowed from #6671. I'm open to suggestions whether we really need an env var, or if it should be called something else. I thought it would be easiest if it's the same as in the other PR, but that can also change.

I am somehow hesitant to use ENV variables for such behavior as they are nowhere documented. And is there really a reason to not exit in such a case? IIRC the current implementation will throw you in a REPL but you can't do there anything at all - so I think it is fine to just bail out here as you are caught in a process which you have to exit to do anything useful. (edit: although this should be tested - IIRC the IDE and sclang REPL has a re-compile command which re-compiles the class library w/o restarting the interpreter. would something only based on primitives like `2+2` still work w/ a failed class library?)

In general, it is IMO either worth to be switched on & documented via the CLI options or it is a dev only feature which can be controlled by changing the source code.

> I've chosen `1` as the exit code. Another option would be `-1` I guess (IIUC that would overflow to `255`?).

Maybe we should use a rather distinctive error code for this so we can separate this from the traditional `1.exit`? `255` is also way too common, maybe we can use something really distinctive like `43` (4'33 would be a good error code, but is too big^^)?
In a unix shell you'd check for `$? != 0` nonetheless, but if people report "i get error code 43 when starting sclang" we immediately know it is a non-compiling class library.

dyfer:
> edit: although this should be tested - IIRC the IDE and sclang REPL has a re-compile command which re-compiles the class library w/o restarting the interpreter. would something only based on primitives like `2+2` still work w/ a failed class library?)

I think in theory maybe this is how this was imagined - that the language still worked. That's why I included an option to disable this - in case it's needed for something... But that's not the case in the current implementation - the interpreter just sits there.

I'll remove the env var; if we later decide we need to have this behavior switched conditionally, we can add a command line switch or something.

As for the return code, I think it would be great to review this project-wise. How about we create a list of constants for exit codes that would be used throughout? 

like 
```cpp
#define LIBRARY_NOT_COMPILED 43
// for the servers:
#define DEVICE_NOT_AVAILABLE 64
#define SAMPLERATE_NOT_AVAILABLE 65
```

What do you think?

capital-G (approved):
> What do you think?

Sound like a good idea to have them defined as const values in one place :)

Code looks also fine to me!

Edit: according to the [new way of reviewing](https://github.com/supercollider/supercollider/wiki/Council-structure#who-is-merging-prs) I think this should be given a review by one of https://github.com/supercollider/supercollider/wiki/Council-structure#sclang and they are free to merge it.

lang/LangSource/SC_TerminalClient.cpp
```diff
@@ -260,6 +260,14 @@ int SC_TerminalClient::run(int argc, char** argv) {
     // startup library
     compileLibrary(opt.mStandalone);
 
+    if (!compiledOK) {
+        post("ERROR: Library has not been compiled successfully.\n");
+        shutdownLibrary();
+        flush();
+        shutdownRuntime();
+        return 43;
```
JordanHendersonMusic, line 268: This should be `EXIT_FAILURE` from `<cstdlib>`, not 43  — just yet :wink:

I'd rather define custom exit codes all in one PR and use explicit names rather than magic numbers, see my comment here https://github.com/supercollider/supercollider/issues/6728#issuecomment-2802651512

JordanHendersonMusic (changes requested):
Good change, but lets not use magic numbers for exit codes. Better to have a generic exit code than something magic and undocumented. 

Custom exist codes are a good idea, but I'd rather do this all in one PR and use explicit names — see my comment here https://github.com/supercollider/supercollider/issues/6728#issuecomment-2802651512

lang/LangSource/SC_TerminalClient.cpp
dyfer: Sounds good to me, I'll make the change. I think using a well established generic code in this PR is good for now - we'd adjust exit codes throughout when implementing https://github.com/supercollider/supercollider/issues/6728

PR #6731, opened 2025-04-15
sclang: add helper for reallocstack and correct objectPerformArgsImpl

JordanHendersonMusic:
## Purpose and Motivation

Adds a little helper for reallocstack as most of the time you want to append something to the stack and calculating all the offsets is error prone.

Also applies this in objectPerformArgsImpl, where I had missed this causing a segfault.

```supercollider
2.performArgs(\pow, {1.0.rand}!2048, [])
```

## Types of changes

- Helper method
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -840,6 +840,17 @@ void reallocStack(struct VMGlobals* g, int stackNeeded, int stackDepth) {
     g->sp = array->slots + stackDepth - 1;
 }
 
+bool maybeReallocStack(struct VMGlobals* g, int toAppend) {
+    PyrObject* stack = g->gc->Stack();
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
@telephon what do you think of this? I've just realised there the same issue over here https://github.com/telephon/supercollider/pull/12 and would be good to use this over there.

JordanHendersonMusic:
This is a potential segfault fix and should be in 3.14

capital-G (approved):
Can't really reason about the implementation, but I am sure it is an improvement.

Feel free to merge @JordanHendersonMusic

JordanHendersonMusic:
When you add stuff to the stack you need to make sure it is big enough... But this invalidates any pointer you have. This is just a nicer way to do it.

PR #6733, opened 2025-04-15
Fix boost thread on clang17

capital-G:
## Purpose and Motivation

macOS 15.4 bumps clang to version 17 which introduces an error when compiling `boost::thread` - see https://github.com/boostorg/thread/issues/402

```cmake
[3/5] Building CXX object lang/CMakeFi...bsclang.dir/LangSource/PyrObject.cpp.o
FAILED: lang/CMakeFiles/libsclang.dir/LangSource/PyrObject.cpp.o 
[...]
In file included from /Users/scheiba/github/supercollider/lang/LangSource/PyrObject.cpp:45:
/Users/scheiba/github/supercollider/external_libraries/boost/boost/thread/future.hpp:4671:19: error: no member named 'that' in 'run_it<FutureExecutorContinuationSharedState>'; did you mean 'that_'?
 4671 |           that_=x.that;
      |                   ^~~~
      |                   that_
/Users/scheiba/github/supercollider/external_libraries/boost/boost/thread/future.hpp:4649:55: note: 'that_' declared here
 4649 |     shared_ptr<FutureExecutorContinuationSharedState> that_;
      |                                                       ^
1 error generated.
```

This is a small fix/hack/patch so it is possible to build sclang again with the most recent version of macOS so we don't have to bump all of boost again (it seems boost 1.88 adds a bugfix).

The proper way would probably be to replace `boost::thread` with `std::thread`, which would be another PR.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/external_libraries/boost/boost/thread/future.hpp
+++ b/external_libraries/boost/boost/thread/future.hpp
@@ -4668,7 +4668,7 @@ namespace detail
       }
       run_it& operator=(BOOST_THREAD_RV_REF(run_it) x) BOOST_NOEXCEPT {
         if (this != &x) {
-          that_=x.that;
+          that_=x.that_;
           x.that_.reset();
         }
         return *this;


dyfer (approved):
If it compiles and runs, that's fine for me, even though I don't understand the deeper implications.

~~EDIT: is it worth adding a comment that this was changed manually?~~

dyfer:
I wonder if we should have one macOS job that always uses the newest xcode, to catch issues like this. The "use system libraries" wouldn't catch this particular issue though, as it uses system boost. Maybe the "shared libscsynth"...?

capital-G:
> Relatedly (but for another PR) I wonder if we should have one macOS job that always uses the newest xcode (at least available in GHA), to catch issues like this. The "use system libraries" wouldn't catch this particular issue though, as it uses system boost. Maybe the "shared libscsynth"...?

We wouldn't have catched this either in this case as the GitHub runner uses a different clang version (15.0.0) than stock macOS 15.4 w/ latest xcode does (17.0.0) - see https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
Somehow the provisioning of GHA runners is often counter-intuitive.

dyfer:
> GitHub runner uses a different clang version (15.0.0) than stock macOS 15.4 w/ latest xcode does (17.0.0)

Ah, interesting. I thought that the clang version is tied to xcode version... In any case, the point about checking this in CI is moot for now.

dyfer:
I think this can be merged. Any objections?

PR #6735, opened 2025-04-15
Topic/schelp add gray noise explanation & example

### LINKED ISSUE #6718: Enhancing the Documentation for `GrayNoise.ar`
The current implementation of `GrayNoise.ar` in SuperCollider differs from the commonly understood concept of "gray noise" (or "grey noise") based on the A-weighting curve. To clarify this distinction, it would be beneficial to include the following note in the documentation:

> `GrayNoise` is derived from digital signal characteristics (e.g., bit-flipping errors) and should not be confused with "gray noise" or "grey noise," which typically refers to noise adjusted to human auditory sensitivity curves, such as the A-weighting curve.

Adding this note would improve the clarity of `GrayNoise.ar`'s purpose and context, helping users, especially beginners, to better understand its specific implementation.

Would it be worthwhile to propose this as a pull request to enhance the documentation? I look forward to hearing your thoughts.

prko:
## Purpose and Motivation

closes #6718

@HotwheelsSisyphus: Thank you for your kind and detailed instructions, including the mathematical explanation and code example.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/GrayNoise.schelp
+++ b/HelpSource/Classes/GrayNoise.schelp
@@ -21,8 +21,6 @@ See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/
 Notice that the bit-flipping version of gray noise is mentioned in a reply to said usenet message.
 ::
 
-See link::#Comparison with a version of inverse-A-weighted noise#Comparison with a version of inverse-A-weighted noise in Example::.
-
 classmethods::
 
 method::ar, kr
@@ -39,27 +37,8 @@ code::
 (
 SynthDef("help-GrayNoise", { |out = 0|
 	Out.ar(out,
-		GrayNoise.ar(0.1)
+		GrayNoise.ar(-20.dbamp)
 	)
 }).play;
-)
-::
-
-subsection:: Comparison with a version of inverse-A-weighted noise
-code::
-{ GrayNoise.ar * -20.dbamp }.freqscope
-
-(
-{
-	var fftSize = 2048;
-	var binsz = s.sampleRate / fftSize;
-	var in, chain;
-	in = WhiteNoise.ar(-20.dbamp);
-	chain = FFT(LocalBuf(fftSize), in);
-	chain = chain.pvcollect(fftSize / 2, { |m, p, b, i|
-		m * AmpCompA.ir(b * binsz, minAmp: 0.2)
-	});
-	IFFT(chain);
-}.freqscope
 )
 ::
--- a/HelpSource/Classes/GrayNoise.schelp
+++ b/HelpSource/Classes/GrayNoise.schelp
@@ -15,11 +15,17 @@ the position of the nonzero digit is a uniform random variable.
 This type of noise has a high RMS level relative to its peak to peak level.
 The spectrum is emphasized towards lower frequencies.
 
-note::
-link::Classes/GrayNoise:: is derived from digital signal characteristics (e.g., bit-flipping errors) and
-should not be confused with "gray noise" or "grey noise",
-which typically refers to noise adjusted to human auditory sensitivity curves, such as the A-weighting curve.
-See link::https://en.wikipedia.org/wiki/Grey_noise::, and link::#Comparison with conventional gray noise::
+table::
+## strong::Discussion:::
+
+Grey/gray noise does not have a standard definition as pink and white noise do.
+While link::Classes/GrayNoise:: generates bit-flipping noise, the term "grey/gray noise" may also refer to noise shaped to perceptual equal loudness contours.footnote::
+
+See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71#Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise#grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_Noise#colors of noise::.
+Notice that the bit-flipping version of gray noise is mentioned in a reply to said usenet message.
+::
+
+See link::#Comparison with a version of inverse-A-weighted noise#Comparison with a version of inverse-A-weighted noise in Example::.
 ::
 
 classmethods::
@@ -44,7 +50,7 @@ SynthDef("help-GrayNoise", { |out = 0|
 )
 ::
 
-subsection:: Comparison with conventional gray noise
+subsection:: Comparison with a version of inverse-A-weighted noise
 code::
 { GrayNoise.ar * 0.1 }.freqscope
 
--- a/HelpSource/Classes/GrayNoise.schelp
+++ b/HelpSource/Classes/GrayNoise.schelp
@@ -1,22 +1,26 @@
 class:: GrayNoise
-summary:: Gray Noise.
+summary:: Bit-flipping Noise
 related:: Classes/BrownNoise, Classes/ClipNoise, Classes/PinkNoise, Classes/WhiteNoise
 categories::  UGens>Generators>Stochastic
 
-
-
 Description::
 
-Generates noise which results from flipping random bits in a word.footnote::Generated through math::X_{0} = 1;\;X_{n+1} = X_{n} \oplus z::
-, where math::\oplus:: is bitwise exclusive or, and math::z:: is a 32-bit binary number with exactly one nonzero digit; the position of the nonzero digit is a uniform random variable.
+Generates noise which results from flipping random bits in a word.footnote::
+Generated through math::X_{0} = 1;\;X_{n+1} = X_{n} \oplus z::,
+where math::\oplus:: is bitwise exclusive or,
+and math::z:: is a 32-bit binary number with exactly one nonzero digit;
+the position of the nonzero digit is a uniform random variable.
 ::
 
 This type of noise has a high RMS level relative to its peak to peak level.
 The spectrum is emphasized towards lower frequencies.
 
-note::link::Classes/GrayNoise:: is derived from digital signal characteristics (e.g., bit-flipping errors) and should not be confused with "gray noise" or "grey noise," which typically refers to noise adjusted to human auditory sensitivity curves, such as the A-weighting curve. See link::https://en.wikipedia.org/wiki/Grey_noise::::
-
-
+note::
+link::Classes/GrayNoise:: is derived from digital signal characteristics (e.g., bit-flipping errors) and
+should not be confused with "gray noise" or "grey noise",
+which typically refers to noise adjusted to human auditory sensitivity curves, such as the A-weighting curve.
+See link::https://en.wikipedia.org/wiki/Grey_noise::, and link::#Comparison with conventional gray noise::
+::
 
 classmethods::
 
@@ -40,7 +44,7 @@ SynthDef("help-GrayNoise", { |out = 0|
 )
 ::
 
-Comparison with usual gray noise:
+subsection:: Comparison with conventional gray noise
 code::
 { GrayNoise.ar * 0.1 }.freqscope
 
--- a/HelpSource/Classes/GrayNoise.schelp
+++ b/HelpSource/Classes/GrayNoise.schelp
@@ -15,18 +15,14 @@ the position of the nonzero digit is a uniform random variable.
 This type of noise has a high RMS level relative to its peak to peak level.
 The spectrum is emphasized towards lower frequencies.
 
-table::
-## strong::Discussion:::
-
 Grey/gray noise does not have a standard definition as pink and white noise do.
 While link::Classes/GrayNoise:: generates bit-flipping noise, the term "grey/gray noise" may also refer to noise shaped to perceptual equal loudness contours.footnote::
 
-See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71#Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise#grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_Noise#colors of noise::.
+See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71#Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise#grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_noise::.
 Notice that the bit-flipping version of gray noise is mentioned in a reply to said usenet message.
 ::
 
 See link::#Comparison with a version of inverse-A-weighted noise#Comparison with a version of inverse-A-weighted noise in Example::.
-::
 
 classmethods::
 
--- a/HelpSource/Classes/GrayNoise.schelp
+++ b/HelpSource/Classes/GrayNoise.schelp
@@ -17,8 +17,7 @@ The spectrum is emphasized towards lower frequencies.
 
 Grey/gray noise does not have a standard definition as pink and white noise do.
 While link::Classes/GrayNoise:: generates bit-flipping noise, the term "grey/gray noise" may also refer to noise shaped to perceptual equal loudness contours.footnote::
-
-See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71#Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise#grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_noise::.
+See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71##Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise##Grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_noise##Colors of noise::.
 Notice that the bit-flipping version of gray noise is mentioned in a reply to said usenet message.
 ::
 
@@ -48,18 +47,19 @@ SynthDef("help-GrayNoise", { |out = 0|
 
 subsection:: Comparison with a version of inverse-A-weighted noise
 code::
-{ GrayNoise.ar * 0.1 }.freqscope
+{ GrayNoise.ar * -20.dbamp }.freqscope
 
 (
 {
-    var binsz = s.sampleRate / 2048;
-    var in, chain;
-    in = WhiteNoise.ar(0.1);
-    chain = FFT(LocalBuf(2048), in);
-    chain = chain.pvcollect(1024, { |m, p, b, i|
-            m * AmpCompA.ir(b * binsz, minAmp: 0.2)
-            });
-    IFFT(chain);
+	var fftSize = 2048;
+	var binsz = s.sampleRate / fftSize;
+	var in, chain;
+	in = WhiteNoise.ar(-20.dbamp);
+	chain = FFT(LocalBuf(fftSize), in);
+	chain = chain.pvcollect(fftSize / 2, { |m, p, b, i|
+		m * AmpCompA.ir(b * binsz, minAmp: 0.2)
+	});
+	IFFT(chain);
 }.freqscope
 )
 ::
--- a/HelpSource/Classes/GrayNoise.schelp
+++ b/HelpSource/Classes/GrayNoise.schelp
@@ -4,11 +4,18 @@ related:: Classes/BrownNoise, Classes/ClipNoise, Classes/PinkNoise, Classes/Whit
 categories::  UGens>Generators>Stochastic
 
 
+
 Description::
 
-Generates noise which results from flipping random bits in a word.
-This type of noise has a high RMS level relative to its peak to peak
-level. The spectrum is emphasized towards lower frequencies.
+Generates noise which results from flipping random bits in a word.footnote::Generated through math::X_{0} = 1;\;X_{n+1} = X_{n} \oplus z::
+, where math::\oplus:: is bitwise exclusive or, and math::z:: is a 32-bit binary number with exactly one nonzero digit; the position of the nonzero digit is a uniform random variable.
+::
+
+This type of noise has a high RMS level relative to its peak to peak level.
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/GrayNoise.schelp
```diff
@@ -1,15 +1,26 @@
...
 related:: Classes/BrownNoise, Classes/ClipNoise, Classes/PinkNoise, Classes/WhiteNoise
 categories::  UGens>Generators>Stochastic
 
-
 Description::
 
-Generates noise which results from flipping random bits in a word.
-This type of noise has a high RMS level relative to its peak to peak
-level. The spectrum is emphasized towards lower frequencies.
+Generates noise which results from flipping random bits in a word.footnote::
```
dyfer, line 8: Can `footnote::` go in the new line?

dyfer (changes requested):
Thanks! I think the change in the description is good.

The representation of the actual grey noise doesn't sound right to me. I need to circle back to double check though. Marking "changes requested" so that we don't merge before that's checked.

HelpSource/Classes/GrayNoise.schelp
HotwheelsSisyphus: In that case, the parser will insert a space between the sentence and the footnote marker in the rendered help file, which isn't great typographically. But arguably that's a parser problem, and I agree that this looks bad in the source.

dyfer:
After researching this a bit, I found it difficult to find an objective definition of grey noise, aside from it being derived from a perceptual loudness curve. The various examples of grey noise I found online all sound a bit different. 

In this case, I suggest that this PR does not include the sound example of "comparison with grey noise", as it makes the impression that this implementation of grey noise is accurate and I can't say whether it is or not. 

I also suggest to change the wording slightly, see below.

HelpSource/Classes/GrayNoise.schelp
```diff
@@ -1,15 +1,26 @@
...
+::
 
+This type of noise has a high RMS level relative to its peak to peak level.
+The spectrum is emphasized towards lower frequencies.
+
+note::
+link::Classes/GrayNoise:: is derived from digital signal characteristics (e.g., bit-flipping errors) and
+should not be confused with "gray noise" or "grey noise",
+which typically refers to noise adjusted to human auditory sensitivity curves, such as the A-weighting curve.
+See link::https://en.wikipedia.org/wiki/Grey_noise::, and link::#Comparison with conventional gray noise::
```
dyfer, line 22: The signal generated by link::Classes/GrayNoise:: differs from a noise adjusted to human auditory sensitivity curves, such as the A-weighting curve, also known as "gray noise" or "grey noise". See link::https://en.wikipedia.org/wiki/Grey_noise:: for more information.

HotwheelsSisyphus: good call! Now I also did some more digging and the ultimate source for the wiki article seems to be a [newsgroup posting](https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71) from someone working at Ford in the mid-90s who _also_ doesn't know of a source for the term grey noise in the -inverse A-weighted sense (but... someone in the replies _does_mention the bit-flipping kind of noise). In that light, maybe we should qualify the wiki link and/or possibly link the original newsgroup post? And maybe we should be even more careful and change "often referred to" to "occasionally referred to"?

(will send along suggestion in separate comment)

I guess this somewhat unclear spec is the reason why the various implementation all sound slightly different...

HotwheelsSisyphus: Suggestion: 
note::
The term "grey/gray noise" is not as consistently used as, e.g., the terms "white" and "pink" noise, nor is its usage standardized. Apart from the bit-flipping noise generated by link::Classes/GrayNoise::, the term "grey/gray noise" also appears to be occasionally used to refer to noise adjusted to human auditory sensitivity curves, such as the A-weighting curve.::footnote
See  link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71#Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise#grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_Noise#colors of noise::. Notice that the bit-flipping version of gray noise is mentioned in a reply to said usenet message.
:: 
::

prko: Um... In the short span of time, I have received two suggestions, so I cannot simply replace the text with the latest one.

BUT,

in my humble opinion, @dyfer's suggestion has a softer tone than the current text; @HotwheelsSisyphus's new suggestion retains the friendly tone and provides additional detailed information.

Therefore, I would say that @HotwheelsSisyphus's new suggestion encompasses all the important aspects of the previous proposals and even incorporates new information! ;-)

If @dyfer also finds @HotwheelsSisyphus's new suggestion acceptable, I shall proceed to amend the text accordingly.

HotwheelsSisyphus: Actually I'm a bit worried that it (my suggestion) might be unnecessarily confusing, but it seems like it's the situation itself that is confusing here... Marcin, what do you think? 

I added the extra info mostly because I don't want to strengthen the sort of feedback cycle where _because_ it's in wikipedia, we end up believing it's a thing (a "convention") even when wikipedia is the reason _that_ it's a thing. (Doesn't just happen in crowdsourced projects like wiki, also happens in peer-reviewed scholarship if one looks closely enough.)

HelpSource/Classes/GrayNoise.schelp
```diff
@@ -24,13 +35,29 @@ This value will be added to the output.
...
 (
 SynthDef("help-GrayNoise", { |out = 0|
 	Out.ar(out,
 		GrayNoise.ar(0.1)
 	)
 }).play;
 )
+::
+
+subsection:: Comparison with conventional gray noise
```
HotwheelsSisyphus, line 47: either leave that part of the example out altogether like Marcin suggested,
or change line 47 to something more neutral like:

subsection:: Comparison with a version of inverse-A-weighted noise

HelpSource/Classes/GrayNoise.schelp
dyfer: Hahaha, that usenet post is a deep dive of sorts :)

I like @HotwheelsSisyphus suggestion, I'd just fine tune it for wording (e.g.  drop "occasionally"):

```
note::
The usage of the term "grey/gray noise" is neither standardized, nor as consistent as that of the terms "white" and "pink" noise. In distinction from the bit-flipping noise generated by link::Classes/GrayNoise::, the term "grey/gray noise" also appears to be used to refer to noise adjusted to human auditory sensitivity curves, such as the A-weighting curve.::footnote
See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71#Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise#grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_Noise#colors of noise::. Notice that the bit-flipping version of gray noise is mentioned in a reply to said usenet message.
::
::
```
How does that sound?

HelpSource/Classes/GrayNoise.schelp
dyfer: Yes, if we want to keep it, "a version of inverse-A-weighted noise" is fine to me.

HelpSource/Classes/GrayNoise.schelp
mtmccrea: Another idea is to actually make this whole discussion a footnote instead, as it is more of a contextual commentary not a discussion related to the use or implementation of GreyNoise (which is already nicely covered in the description).

dyfer: I like @mtmccrea 's wording
I also like the idea of putting the discussion in the footnote.

HotwheelsSisyphus: I also like @mtmccrea's version. I don't have a strong opinion on whether what's now in `discussion:: ::` should go into a footnote; I think with that version, the amount of info is manageable either way.

HelpSource/Classes/GrayNoise.schelp
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6735/commits/ada4ca15192ca501fcf5dc8c22e8b1b30d2e9efe

HelpSource/Classes/GrayNoise.schelp
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6735/commits/ada4ca15192ca501fcf5dc8c22e8b1b30d2e9efe

However, `Discussion::` can not be used in the same place. I provisionally used `table::## strong::Discussion:::.....`

prko:
Thank you very much for your opinions and suggestions. 
I have attached the screenshots of my latest commit. 
Please do let me know if there are any further modifications I should make.
Screenshot 2025-04-20 at 00 02 43
Screenshot 2025-04-20 at 00 02 53

mtmccrea:
Thanks @prko for the updates.

> Discussion:: can not be used in the same place. I provisionally used table::## strong::Discussion:::.....

Let's just put the text in the Description without the ad-hoc use of table and discussion "tag". It's short enough that it's ok I think.

Last thing, it seems that the aliased hyperlinks in the footnote aren't rendering properly (it shows the alias and the url). Could you doublecheck the formatting? (It could be an error in the renderer?)

Looking in other docs where aliased hyperlinks links are used, this formatting renders correctly in `Date` (though it's not in a footnote):
`link::http://www.boost.org/doc/libs/1_63_0/doc/html/date_time/date_time_io.html##here::`

Otherwise I think this is good to go!

prko:
@mtmccrea Thanks, done!

HelpSource/Classes/GrayNoise.schelp
```diff
@@ -1,15 +1,28 @@
...
+the position of the nonzero digit is a uniform random variable.
+::
+
+This type of noise has a high RMS level relative to its peak to peak level.
+The spectrum is emphasized towards lower frequencies.
+
+Grey/gray noise does not have a standard definition as pink and white noise do.
+While link::Classes/GrayNoise:: generates bit-flipping noise, the term "grey/gray noise" may also refer to noise shaped to perceptual equal loudness contours.footnote::
+
+See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71#Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise#grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_noise::.
```
dyfer, line 21: This still doesn't render quite right. I think it should be: `See link::https://web.archive.org/web/20090427190825/http://www.ptpart.co.uk:80/show.php?contentid=71##Joseph Wisniewski's 1996 usenet message:: that seems to have informed the wikipedia articles on link::https://en.wikipedia.org/wiki/Grey_noise##Grey noise:: and on link::https://en.wikipedia.org/wiki/Colors_of_noise##Colors of noise::.`

Otherwise looks good!

HelpSource/Classes/GrayNoise.schelp
```diff
@@ -24,13 +37,29 @@ This value will be added to the output.
...
+    var binsz = s.sampleRate / 2048;
+    var in, chain;
+    in = WhiteNoise.ar(0.1);
+    chain = FFT(LocalBuf(2048), in);
+    chain = chain.pvcollect(1024, { |m, p, b, i|
+            m * AmpCompA.ir(b * binsz, minAmp: 0.2)
+            });
+    IFFT(chain);
+}.freqscope
+)
```
dyfer, line 64: I think if we are keeping the example, it could be improved slightly by using a variable for fft size and dbamp for amplitude, to promote db value use:

```supercollider
(
{
	var fftSize = 2048;
	var binsz = s.sampleRate / fftSize;
	var in, chain;
	in = WhiteNoise.ar(-20.dbamp);
	chain = FFT(LocalBuf(fftSize), in);
	chain = chain.pvcollect(fftSize / 2, { |m, p, b, i|
		m * AmpCompA.ir(b * binsz, minAmp: 0.2)
	});
	IFFT(chain);
}.freqscope
)
```

HelpSource/Classes/GrayNoise.schelp
```diff
@@ -24,13 +37,29 @@ This value will be added to the output.
...
 	Out.ar(out,
 		GrayNoise.ar(0.1)
 	)
 }).play;
 )
+::
+
+subsection:: Comparison with a version of inverse-A-weighted noise
+code::
+{ GrayNoise.ar * 0.1 }.freqscope
```
dyfer, line 51:
```supercollider
{ GrayNoise.ar * -20.dbamp }.freqscope
```

dyfer: And maybe use -20.dbamp in the main example?

HelpSource/Classes/GrayNoise.schelp
HotwheelsSisyphus: Oof, I think both these last things were actually oversights from my suggestions, sorry...

dyfer: All good!

HelpSource/Classes/GrayNoise.schelp
prko: Thanks! Now it reads very well!

HelpSource/Classes/GrayNoise.schelp
prko: Thanks, this is good! 
Could `minAmp: 0.2` be `minAmp: -14.dbamp`?

HelpSource/Classes/GrayNoise.schelp
prko: Thanks, Done!

HelpSource/Classes/GrayNoise.schelp
dyfer: Possibly, but I wasn't if that's the best way of representing it here. @HotwheelsSisyphus do you have an opinion on this?

HelpSource/Classes/GrayNoise.schelp
dyfer: I also suggested to change the main example above, if possible: `GrayNoise.ar(0.1)` -> `GrayNoise.ar(-20.dbamp)`.

HelpSource/Classes/GrayNoise.schelp
HotwheelsSisyphus: Hm, I thought about it but didn't want to be too fussy about it. Actually probably best to just leave this optional argument out; the reason I included it was to make the dip in the freqscope display more obvious. (Default is minAmp: 0.36; the dip is still visible but not that strong.)
so, I suggest to change this line to `m * AmpCompA.ir(b * binsz)`

I think this is one of the issues with (this kind of) grey noise in general, namely that the AmpComp should be level dependent as well. 

Alternative to leaving it out, we could try to figure out the "correct" minAmp for the -20 db white noise.
Another alternative: Send Noise into the fft at full blast (0 db), and attenuate after ifft. (i.e., `IFFT(chain) * -20.dbamp`

dyfer: I can't really understand why is `minAmp` related to the input level... shouldn't it only matter in relation to `rootAmt`?

HotwheelsSisyphus: Well now that I look more closely at how AmpCompA is calculated: yes, you're right. I thought it had to do with with the general level-dependence of the a-weighting but it's really simpler than that. 

However, now I do think that attenuating after the fft-ifft chain is clearer, because that way the "dip" db (changing this to -60) is more directly related to the input db (0). Else, you'd have to _also_ provide the rootAmp arg. 
Using -60 because that's roughly the difference between min and max of the ISO226 curves at 40 phon. Impossible to say what the actual level will be that the user might play this back at.

Suggestion: 
(
{
	var fftSize = 2048;
	var binsz = s.sampleRate / fftSize;
	var in, chain;
	in = WhiteNoise.ar(0.dbamp);
	chain = FFT(LocalBuf(fftSize), in);
	chain = chain.pvcollect(fftSize / 2, { |m, p, b, i|
		m * AmpCompA.ir(b * binsz, minAmp: -60.dbamp)
	});
	IFFT(chain) * -20.dbamp;
}.freqscope
)

Alternative suggestion: 
(
{
	var fftSize = 2048;
	var binsz = s.sampleRate / fftSize;
	var in, chain;
	in = WhiteNoise.ar(-20.dbamp);
	chain = FFT(LocalBuf(fftSize), in);
	chain = chain.pvcollect(fftSize / 2, { |m, p, b, i|
		m * AmpCompA.ir(b * binsz, minAmp: -80.dbamp, rootAmp: -20.dbamp)
	});
	IFFT(chain);
}.freqscope
)


This will be the same plot, but minAmp had to be changed to rootAmp - 60db.

dyfer: > However, now I do think that attenuating after the fft-ifft chain is clearer, because that way the "dip" db (changing this to -60) is more directly related to the input db (0). Else, you'd have to _also_ provide the rootAmp arg.

I don't think that's correct.

These two examples look and sound the same to me:

```supercollider
(
{
	var fftSize = 2048;
	var binsz = s.sampleRate / fftSize;
	var in, chain;
	in = WhiteNoise.ar(0.dbamp);
	chain = FFT(LocalBuf(fftSize), in);
	chain = chain.pvcollect(fftSize / 2, { |m, p, b, i|
		m * AmpCompA.ir(b * binsz, minAmp: -60.dbamp)
	});
	IFFT(chain) * -20.dbamp;
}.freqscope
)
(
{
	var fftSize = 2048;
	var binsz = s.sampleRate / fftSize;
	var in, chain;
	in = WhiteNoise.ar(-20.dbamp);
	chain = FFT(LocalBuf(fftSize), in);
	chain = chain.pvcollect(fftSize / 2, { |m, p, b, i|
		m * AmpCompA.ir(b * binsz, minAmp: -60.dbamp)
	});
	IFFT(chain);
}.freqscope
)
```

Additionally, your last example (with `m * AmpCompA.ir(b * binsz, minAmp: -80.dbamp, rootAmp: -20.dbamp)`) does sound and look different.

HotwheelsSisyphus: just noticed that the using the rootamp gives a strange freqscope result, probably because the AmpCompA function is not meant to be used for higher frequencies? It's mostly odd that  setting rootamp to -20db makes any difference.
 Capture d’écran, le 2025-04-20 à 14 00 38
Plotting the same function, it seems ok:
Capture d’écran, le 2025-04-20 à 14 00 58

Obviously that's not to do with the GrayNoise doc, but that's why for now I'd leave the rootAmp arg out.

dyfer: Also, using white noise and the -60dB dip makes this sound very different than other examples of gray noise online, which tends to be more towards the "equal loudness" perception. This goes back to lack of the exact definitiion... but if we are to provide an example, it should at least sound somewhat similar... One way would be to use PinkNoise as the source, but I don't know how to justify if theoretically. 
"By ear" the 60dB dip sounds like too much to me.

HotwheelsSisyphus: yes, my second example was broken. Although I don't quite understand what's happening, possibly an issue with AmpCompA.

< < However, now I do think that attenuating after the fft-ifft chain is clearer, because that way the "dip" db (changing this to -60) is more directly related to the input db (0). Else, you'd have to also provide the rootAmp arg.
< I don't think that's correct.

You're right of course, rootAmp doesn't work the way I thought it did.

HotwheelsSisyphus: I didn't actually listen to it I confess ... it's also futile, as it's meant to depend on what volume your playback system is set up. How prominent or not the highs will sound is very dependent on the SPL. That's somewhat the point. Anyway, if someone wants to finetune this, go ahead, but this shouldn't be me as I have neither the virgin ears nor the equipment/infrastructure to do so. (How would one even start to make SPL levels comparable on consumer-level audio equipment?)

HotwheelsSisyphus: To move this along: Should we just settle on minAmp: -30.dbamp (or -15, or...) and include a note  or comment like: 

//Note: plausibility of this as "equally loud across the spectrum" will depend on your playback volume!

HotwheelsSisyphus: or just leave the example out?

dyfer: I'm not sure about the whole "it depends on the actual SPL" claim. I believe that main contour would be perceived similarly in a range SPL levels, so getting the relative differences in dB should be close enough.

> I didn't actually listen to it I confess ...

Posting examples without listening to them should really be avoided.

> or just leave the example out?

These kinds of nuances are the reason I originally suggested to leave the example out. I think, given all the discussion, that would be my preference.

HotwheelsSisyphus: Ok, let's leave it out.

> I'm not sure about the whole "it depends on the actual SPL" claim. I believe that main contour would be perceived similarly in a range SPL levels, so getting the relative differences in dB should be close enough.

The relative differences in dB are precisely what is SPL dependent.

HotwheelsSisyphus: Yup, I think the high end is where there may be a problem with the AmpCompA implementation, I might look that over at some point. It seems that the A weighting is not defined over 11k and that makes the AmpCompA return more highs than reasonable... gotta check first if maybe I just didn't read the docs properly on that.

HelpSource/Classes/GrayNoise.schelp
prko: Sorry, done in https://github.com/supercollider/supercollider/pull/6735/commits/cc09abfaf478c28d33ca20ee20b1032630283980

HelpSource/Classes/GrayNoise.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6735/commits/cc09abfaf478c28d33ca20ee20b1032630283980

dyfer (approved):
Thanks! I appreciate all the responses in this relatively lengthy review :)

prko:
Thank you very much; I've thoroughly enjoyed this, and it has been a pleasure reading the responses and working on it.

PR #6738, opened 2025-04-16
Add Server -quitSync

dyfer:
## Purpose and Motivation

We didn't have an async quit method that actually waits for the server process to quit (i.e. for the callback from the system command to be triggered). This PR adds method `quitSync` and changes all unitTests to use this method instead of .quit. This is done to ensure that the server is fully shut down, before a new server instance is started, especially under heavy system load.

### Notes
This is a workaround-quality implementation: our watching of the server status is quite messy IMO and there have been attempts to rewrite it in the past. This PR does not attempts this, but instead leverages currently available methods. 

In particular, for local servers the final "onComplete" callback for shutdown should be called when the process exists, i.e. it should be triggered by the server command callback function. This is just a note for a possible future implementation.

### ~~Failed unit tests~~

EDIT: I found the reason of the failures (calling `s.sync` after `s.quitSync`) and fixed it.
<details>

**This is not an issue anymore**

After implementing `.quitSync` in the test suite, the following tests would hang the testing routine:

[`TestNodeProxy_Server -test_loaded_after_quit`](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestNodeProxy_Server.sc#L81)
[`TestNodeProxy_Server -test_send_after_quit `](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestNodeProxy_Server.sc#L87)
[`TestNodeProxy_Server -test_rebuild_after_quit `](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestNodeProxy_Server.sc#L94)

~~I didn't follow the logic enough to understand what the issue might me. These tests have been disabled in this PR.~~

~~I appreciate any feedback on this!~~
</details>

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Server.schelp
+++ b/HelpSource/Classes/Server.schelp
@@ -935,6 +935,11 @@ Routine.run {
 )
 ::
 
+method:: quitSync
+Quit the Server and wait until the server process actually shuts down.
+argument:: condition
+an optional instance of link::Classes/Condition:: to be used within the synchronisation process.
+
 subsection:: Bela
 
 method:: belaScope
--- a/SCClassLibrary/Common/Control/Server.sc
+++ b/SCClassLibrary/Common/Control/Server.sc
@@ -1056,6 +1056,41 @@ Server {
 		this.newAllocators;
 	}
 
+	quitSync { |condition|
+		condition ?? { condition = Condition.new };
+		condition.test = false;
+		this.quit;
+		this.prWaitForPidRelease({
+			condition.test = true;
+			condition.signal
+		}, {
+			// we may end up here
+			// if the server happened to become unresponsive recently
+			// in that case we try to quit again
+			this.quit;
+			this.prWaitForPidRelease({
+				condition.test = true;
+				condition.signal
+			}, {
+				// as a last resort...
+				"Server '%' failed to quit".format(this.name).postln;
+				this.pid !? {
+					"Forcing process with pid % to stop via system command.".format(this.pid).postln;
+					thisProcess.platform.killProcessByID(this.pid)
+				};
+				this.prWaitForPidRelease({
+					condition.test = true;
+					condition.signal
+				}, {
+					"The process did not stop after a timeout, which may result in dangling network resources.".warn;
+					condition.test = true;
+					condition.signal
+				});
+			});
+		}, this.statusWatcher.timeoutBeforeConsideredDeadOnExit * 1.5); // timeout is longer than ServerStatusWatcher waiting to declare the server unresponsive
+		condition.wait;
+	}
+
 	*quitAll { |watchShutDown = true|
 		all.do { |server|
 			if(server.sendQuit === true) {
--- a/SCClassLibrary/Common/Control/ServerStatus.sc
+++ b/SCClassLibrary/Common/Control/ServerStatus.sc
@@ -12,6 +12,7 @@ ServerStatusWatcher {
 	var <sampleRate, <actualSampleRate;
 
 	var reallyDeadCount = 0, bootNotifyFirst = true;
+	var <>timeoutBeforeConsideredDeadOnExit = 3.0;
 
 	*new { |server|
 		^super.newCopyArgs(server)
@@ -100,13 +101,13 @@ ServerStatusWatcher {
 					}
 				}, '/done', server.addr);
 
-				AppClock.sched(3.0, {
+				AppClock.sched(timeoutBeforeConsideredDeadOnExit, {
 					if(serverReallyQuit.not) {
 						if(unresponsive) {
-							"Server '%' remained unresponsive during quit."
+							"Server '%' remained unresponsive during quit.".format(server.name).warn;
 						} {
-							"Server '%' failed to quit after 3.0 seconds."
-						}.format(server.name).warn;
+							"Server '%' failed to quit after % seconds.".format(server.name, timeoutBeforeConsideredDeadOnExit).warn;
+						};
 						// don't accumulate quit-watchers if /done doesn't come back
 						serverReallyQuitWatcher.free;
 						statusWatcher !? { statusWatcher.disable };
--- a/testsuite/classlibrary/TestBuffer_Server.sc
+++ b/testsuite/classlibrary/TestBuffer_Server.sc
@@ -9,7 +9,7 @@ TestBuffer_Server : UnitTest {
 
 	tearDown {
 		Buffer.freeAll(server);
-		server.quit;
+		server.quitSync;
 		server.remove;
 	}
 
--- a/testsuite/classlibrary/TestBus.sc
+++ b/testsuite/classlibrary/TestBus.sc
@@ -76,7 +76,7 @@ TestBus : UnitTest {
 		this.assertFloatEquals(get_value, set_value, "Bus:get works", 0.001);
 
 		bus.free;
-		s.quit;
+		s.quitSync;
 		s.remove;
 	}
 
@@ -105,7 +105,7 @@ TestBus : UnitTest {
 		this.assertArrayFloatEquals(get_values, set_values[0..1], "Bus:getn works", 0.001);
 
 		bus.free;
-		s.quit;
+		s.quitSync;
 		s.remove;
 	}
 
--- a/testsuite/classlibrary/TestCoreUGens.sc
+++ b/testsuite/classlibrary/TestCoreUGens.sc
@@ -21,7 +21,7 @@ TestCoreUGens : UnitTest {
 	}
 
 	tearDown {
-		server.quit;
+		server.quitSync;
 		server.remove;
 	}
 
--- a/testsuite/classlibrary/TestDoneActions.sc
+++ b/testsuite/classlibrary/TestDoneActions.sc
@@ -16,7 +16,7 @@ TestDoneActions : UnitTest {
 
 	tearDown {
 
-		server.quit.remove;
+		server.quitSync.remove;
 
 	}
 
--- a/testsuite/classlibrary/TestEnvGen_server.sc
+++ b/testsuite/classlibrary/TestEnvGen_server.sc
@@ -7,7 +7,7 @@ TestEnvGen_server : UnitTest {
 	}
 
 	tearDown {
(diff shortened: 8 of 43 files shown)


dyfer:
Note: build failures are note related to this PR, there seems to be an issue with installing Qt.

telephon:
> After implementing `.quitSync` in the test suite, the following tests would hang the testing routine:
> 
> [`TestNodeProxy_Server -test_loaded_after_quit`](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestNodeProxy_Server.sc#L81) [`TestNodeProxy_Server -test_send_after_quit `](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestNodeProxy_Server.sc#L87) [`TestNodeProxy_Server -test_rebuild_after_quit `](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestNodeProxy_Server.sc#L94)
> 
> I didn't follow the logic enough to understand what the issue might me. These tests have been disabled in this PR.
> 
> I appreciate any feedback on this!

I am not sure why this is. Have you tried writing instead:

```supercollider
	test_loaded_after_quit {
		proxy.send;
		server.quitSync;
		this.assertEquals(proxy.loaded, false, "NodeProxy should not be loaded after server quit");
	}
```
?

dyfer:
> I am not sure why this is. Have you tried writing instead:

I guess I should've linked to a file in this PR... [what you proposed is how they are written in this PR](https://github.com/supercollider/supercollider/pull/6738/files#diff-0e1ed2b8e14ce0360507b4a5189a14f125f2686d5f26d5d5419a9666543e9251R81-R101).

`server.quiteSync` is what makes the testing routine freeze, it's really curious.. 
Here's what I get when running the test locally on this branch (with added comments):
```supercollider
TestNodeProxy_Server.run;
```
```supercollider
RUNNING UNIT TEST 'TestNodeProxy_Server'
-> TestNodeProxy_Server
(...) // other tests running
Booting server 'TestNodeProxy_Server' on address // test_loaded_after_quit starts
(...) // server boots
Shared memory server interface initialized // server finished booting
server 'TestNodeProxy_Server' disconnected shared memory interface
'/quit' message sent to server 'TestNodeProxy_Server'.
Server 'TestNodeProxy_Server' exited with exit code 0.
PASS: a TestNodeProxy_Server: test_loaded_after_quit - NodeProxy should not be loaded after server quit
// and... NOTHING, subsequent tests don't run
```
I added some posting when testing this and I see that we don't start the subsequent test's `setUp` method, but we do get past `this.assertEquals(proxy.loaded, false, "NodeProxy should not be loaded after server quit");` in `test_loaded_after_quit`. I have no idea what's going on.

dyfer:
I found the reason for the failures. There was `server.sync` called in the teardown function, after `server.quitSync` was called in the test itself and the server wasn't running anymore.

HelpSource/Classes/Server.schelp
```diff
@@ -922,6 +922,11 @@ Routine.run {
 )
 ::
 
+method:: quitSync
+Quit the Server and wait until the server process actually shuts down.
+argument:: condition
+an optional instance of link::Classes/Condition:: used for evaluating this.
```
telephon, line 928: "used for evaluating this" is a bit unclear. Maybe: "to be  used within the synchronisation process"?

SCClassLibrary/Common/Control/Server.sc
```diff
@@ -1151,6 +1151,35 @@ Server {
...
+				// as a last resort...
+				"Server '%' failed to quit".format(this.name).postln;
+				this.pid !? {
+					"Forcing process to stop via system command.".postln;
+					thisProcess.platform.killProcessByID(this.pid)
+				};
+				condition.test = true;
+				condition.signal
+			});
+		}, 4); // timeout is longer than ServerStatusWatcher waiting to declare the server unresponsive
```
telephon, line 1179: This is actually: `this.options.pingsBeforeConsideredDead`. Maybe use that?

telephon (changes requested):

SCClassLibrary/Common/Control/Server.sc
dyfer: Hm, I don't think so? `-pingsBeforeConsideredDead` is a counter and is used while server is running. I was referring to https://github.com/supercollider/supercollider/blob/e9922139f8dddf95cfba4e6315727b4b1978fca8/SCClassLibrary/Common/Control/ServerStatus.sc#L103

HelpSource/Classes/Server.schelp
dyfer: I agree, but I copied it verbatim from [-bootSync description](https://github.com/supercollider/supercollider/pull/6738/files/e9922139f8dddf95cfba4e6315727b4b1978fca8#diff-7aa43aba358522fcf2dc1f56d8e54263d1554a55f5eec8186367ca523e6701e2L869). Should I update it only here or also in -bootSync?

SCClassLibrary/Common/Control/Server.sc
telephon: You are right. Could you be so kind and make this an explicit return value in ServerStatus? Whe I wrote it, I simply copied it over. If someone changes this value in the code, the test would subtly fail.

dyfer: Do you mean something like
```supercollider
AppClock.sched(timeoutBeforeConsideredDeadOnExit, { 
```
?

HelpSource/Classes/Server.schelp
dyfer: I updated the description as suggested (but did not change it in -bootSync, should I?).

SCClassLibrary/Common/Control/Server.sc
dyfer: I have added `timeoutBeforeConsideredDeadOnExit` to server status watcher and used that value, multiplied by `1.5` for the timeout when force quitting a hanging server.

dyfer:
@telephon sorry for the delay. Can we revisit this PR? It does not clean up the boot/quit process drastically, but it does seem to cover most cases when needing to wait for the server to quit. 

When rebasing this PR to address merge conflicts, I updated all tests to use `quitSync` instead of `quit`, including new ones added since opening this PR.

After consulting an LLM it seems that one scenario that's not covered by the proposed implementation is calling `quitSync` from more than one routine at the same time. Do you think I should also account for this?

telephon:
I'm not sure about the routines situation. In the case that parallel routines break quitSync, couldn't you pass in a shared condition to all the cases in different ones? And if yes, the condition may better be shared in the ServerWatcher anyway.

dyfer:
> In the case that parallel routines break quitSync, couldn't you pass in a shared condition to all the cases in different ones? And if yes, the condition may better be shared in the ServerWatcher anyway.

Maybe, I don't know. The question I was asking is whether you think such edge case is important enough to be addressed at this point, or not. I'd prefer to move forward without addressing it, but let me know if you think otherwise.

Have your other concerns been addressed? Do you have any other feedback on this PR? 

Thanks!

JordanHendersonMusic:
If this can't be resolved, I think it would be valuable to mark this as private and merge (or perhaps make it an extension method that is a part of the test suite) so we can use it in the ci and sure up some of the dodgy tests.

telephon:
I would even say that it can be public when the limitations are documented.

dyfer:
I think the quitSync concurrency was a false alarm. I tried both simultaneous and staggered execution of quitSync in parallel routines and all of them finished successfully. The only downside is that sometimes we get "the server didn't shut down after X seconds" warning when making concurrent .quitSync calls.

```supercollider
(
{
	s.bootSync;
	"server booted".postln;
	{
		s.quitSync;
		"1 done".postln;
	}.fork;
	
	{
		s.quitSync;
		"2 done".postln;
	}.fork;
	
	{
		s.quitSync;
		"3 done".postln;
	}.fork;
}.fork;
)

(
{
	s.bootSync;
	"server booted".postln;
	s.statusWatcher.timeoutBeforeConsideredDeadOnExit = 0.05; // force failure branch
	Buffer.read(s, "path/to/a/large/file"); // load a large audio file to block the server for a moment
	{
		s.quitSync;
		"1 done".postln;
	}.fork;
	
	0.1.wait;
	
	{
		s.quitSync;
		"2 done".postln;
	}.fork;
	
	{
		s.quitSync;
		"3 done".postln;
	}.fork;
}.fork;
)
```

@telephon you still have "changes requested" in this PR, have I addressed your concerns?

JordanHendersonMusic (approved):
>  I tried both simultaneous and staggered execution of quitSync in parallel routines and all of them finished successfully.

That sounds good to me!

telephon:
Thanks!

One question, when rereading: currently, `quitSync` is not exactly a sycnched version of `quit`, right? Because `quit` has a number of args while `quitSync` only has the condition as an argument.

In case we wanted to exchange one for the other in some "normal" (i.e. non-test) situaions, the interface would have to be adjusted.

JordanHendersonMusic:
Oh good point!

You might consider
```
quitSync { |condition ...a, k|
   ...
   this.performArgs(\quit, a, k);
   ...
}
```
this way, if the interface even changes, you get the change here to.

dyfer:
Well, `bootSync` doesn't match `boot` either.

FWIW, `quitSync` matches `bootSync`.

dyfer:
> You might consider
> 
> ```
> quitSync { |condition ...a, k|
>    ...
>    this.performArgs(\quit, a, k);
>    ...
> }
> ```
> 
> this way, if the interface even changes, you get the change here to.

The implementation is not as simple as wrapping `.quit` in some other code. `quitSync` goes through multiple fallbacks for shutdown "failure", so it's hard to justify what's `onSuccess` and what's `onFailure`.

telephon:
Makes sense!

telephon (approved):

dyfer:
Thanks! I'll merge then.

PR #6740, opened 2025-04-17
Modify `{}.freqscope` to control synth like `{}.play`

prko:
## Purpose and Motivation
`{}.freqscope` currently lacks functionality to control a synth in the same way as `{}.play`.  
At present, the following code is required to view the spectrum:
 ```supercollider
{ SinOsc.ar * 0.1 }.play(outbus:1)
FreqScope(busNum:1)
```
If this pull request is merged, the following functionality will become available:
```
{ SinOsc.ar * 0.1 }.freqscope(outbus: 1)
```
More examples:
```
x = { |freq = 440, amp = 0.1| SinOsc.ar(freq) * 0.1 }

y = x.freqscope
y.free

y = x.freqscope(outbus:1, args:[\freq, 880])
y.set(\freq, 1760)
y.release(4)
```

## Types of changes

- Documentation
- New feature
- Breaking change: If this PR is merged, `{}.freqscope` returns a `Synth`.

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
@@ -49,9 +49,8 @@
 		^synth
 	}
 
-	freqscope {
-		var server = Server.default;
-		this.play(server);
-		^FreqScope.new(server: server)
+	freqscope { arg target, outbus = 0, fadeTime = 0.02, addAction=\addToHead, args;
+		FreqScope.new(server: target)
+		^this.play(target, outbus, fadeTime, addAction, args)
 	}
 }
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
@@ -50,7 +50,7 @@
 	}
 
 	freqscope { arg target, outbus = 0, fadeTime = 0.02, addAction=\addToHead, args;
-		FreqScope.new(server: target)
+		FreqScope.new(server: target, busNum: outbus);
 		^this.play(target, outbus, fadeTime, addAction, args)
 	}
 }
--- a/HelpSource/Classes/Function.schelp
+++ b/HelpSource/Classes/Function.schelp
@@ -512,6 +512,38 @@ argument::zoom
 A zoom value for the scope's X axis. Larger values show more. The default is 1.
 argument:: bounds
 The bounds of the code::Window:: in which the scope is embedded. A Rect specifying position and size of the window. The size does not include the border and title bar. Position is measured from the bottom-left corner of the screen (this is different than link::Classes/View#-bounds::). The default is nil and the scope window will be in the centre of the display.
+
+
+method::freqscope
+
+Play a Synth using the UGens returned by the function, similar to link::#-play:: as described above.
+Additionally, it opens a link::Classes/FreqScope:: window to visualise the output spectrum.
+For details about the arguments, refer to link::#-play::.
+For more information on link::Classes/FreqScope::, consult its documentation.
+
+code::
+x = { |freq = 440, amp = 0.1| SinOsc.ar(freq) * 0.1 }
+
+y = x.freqscope
+y.free
+
+y = x.freqscope(outbus:1, args:[\freq, 880])
+y.set(\freq, 1760)
+y.release(4)
+::
+
+argument::target
+Similar to the teletype::target:: argument in link::#-play::. The display bus of the link::Classes/FreqScope:: window will be synchronised with this target.
+argument::outbus
+Similar to the teletype::outbus:: argument in link::#-play::. The display bus of the link::Classes/FreqScope:: window will be synchronised with this outbus.
+argument::fadeTime
+Identical to the teletype::fadeTime:: argument in link::#-play::.
+argument::addAction
+Identical to the teletype::addAction:: argument in link::#-play::.
+argument::args
+Identical to the teletype::args:: argument in link::#-play::.
+
+
 method::plot
 
 note::See link::Reference/plot:: for a general reference on plotting::

### CODE REVIEW

File: SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc [hunk 1]
```diff
@@ -49,9 +49,8 @@
 		^synth
 	}
 
-	freqscope {
-		var server = Server.default;
-		this.play(server);
-		^FreqScope.new(server: server)
+	freqscope { arg target, outbus = 0, fadeTime = 0.02, addAction=\addToHead, args;
+		FreqScope.new(server: target, busNum: outbus);
+		^this.play(target, outbus, fadeTime, addAction, args)
```
dyfer, line 54: It's useful to have arguments settable!

But I'm not sure about changing the returning object here. `Function -freqscope` is only a "helper" method, not the primary way of creating synthdefs; however I don't think the expanded functionality warrants the change in the interface. 

Note that returning FreqScope (as original) this wouldn't work: `y = x.freqscope(outbus:1, args:[\freq, 880]); y.set(\freq, 1760)`. I think it's better to maintain the original interface though... I wonder if others have opinion on this?

JordanHendersonMusic:
Rather than doing it this way, you might consider using the new kwargs syntax and passing along all keywords arguments with `performArgs`. This means it will automatically work the @Spacechild1 's reblocking, or any other additions to synth.

SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
prko: >Note that returning FreqScope (as original) this wouldn't work: y = x.freqscope(outbus:1, args:[\freq, 880]); y.set(\freq, 1760). I think it's better to maintain the original interface though... 
YesI could not found a way to control synth with returning `FreqScope` (as original).

What I could do is as follows:
```
freqscope { arg target, outbus = 0, fadeTime = 0.02, addAction=\addToHead, args;
		FreqScope.new(server: target, busNum: outbus);
		^[
				FreqScope.new(server: target, busNum: outbus);
 				this.play(target, outbus, fadeTime, addAction, args)
		]
```
However, this will be not accepted, I think.

For me, the advantages of moving away from the old approach are significant. Please refer to my writing (apologies, it is in Korean), and consider how this change could save space on paper and time during lectures by eliminating the need to use `{}.play` and `FreqScope()`:
Screenshot 2025-04-18 at 08 04 03

prko:
@JordanHendersonMusic  
Thank you for your advice. Unfortunately, I am having difficulty understanding it and am unsure how to apply `kwargs` and `performArgs` to make the necessary changes. Could you kindly point me to some practical usage examples? The materials I have found so far are rather brief and abstract, and I am struggling to adapt them to this specific case. My apologies for any lack of clarity on my part.  

If you find the idea behind this PR appealing, rather than teaching me, you are welcome to create a new PR yourself. I would greatly appreciate it!

prko:
The initial step to improve `.freqscope` functionality might be ensuring that the `FreqScope` window comes to the front if there is already an open `FreqScope` window.

I will consider other opinions before deciding on the approach:  
- Implement it within this PR  
- Open a new PR for the enhancement  

Thanks!

SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
dyfer: Can you clarify why do you need to repeatedly call {}.freqscope? once freqscope is open, it keeps running.
If the idea is to bring it to front, that could be solved with `s.freqscope.window.alwaysOnTop_(true)`.

prko: >Can you clarify why do you need to repeatedly call Freqscope() repeatedly? 
>If the idea is to bring it to front after each invocation, then why not keep it on top instead?`s.freqscope.window.alwaysOnTop_(true)`.

Ah, thanks! I have fogott `.window.alwaysOnTop_(true)`

I think we can close this PR after reading @JordanHendersonMusic's opinion!

prko: After using `.window.alwaysOnTop_(true)`, I would describe this functionality as being equivalent to adjusting the SC-IDE window to allow other windows to remain visible.

In certain scenarios, invoking the window with `.front` proves to be practical.  
```supercollider
(
Window.closeAll;
s.waitForBoot {
	~spectrum = FreqScope();
	~bounds = ~spectrum.bounds;
	~demo = { |aCode|
		var width, height;
		# width, height = [~bounds.width, ~bounds.height];
		{ aCode.interpret.scope(bufsize: s.sampleRate / 220, bounds: Rect(width + 1, 0, height, height))
			* Env.pairs([[0, 0, -9], [1, 1, 9], [5, 1], [6, 0]]).kr(2) * 0.1 ! 2 }.play;
		~spectrum.front };
};
)
```

Then I can use:  
```supercollider
~demo.("SinOsc.ar");
~demo.("LFTri.ar");
~demo.("LFPulse.ar");
~demo.("LFPulse.ar(width: 0.8)");
~demo.("Impulse.ar");
~demo.("Impulse.ar(1)");
```

Initially, my aim was to achieve a one-liner, such as `~demo.("Impulse.ar(1)")`, without the need to define the function `~demo`. That was my original intention.

dyfer: All right. 
I don't think we should be changing the returning object from {}.freqscope to satisfy a specific use case (and potentially break other code, even for a helper function like this). 

For your own use you may as well create a specific method/extension that would provide the behavior you need.

For this PR, I think we should stick to only exposing more parameters, possibly using the performArgs method (I haven't looked into it myself).

JordanHendersonMusic:
I agree with @dyfer about making sure to return the same type. Although it does seem one is unable to free the synth object afterwards? That seems like a bug?

I do need to add some more documentation for kwargs: basically you'd do something like this...

```supercollider

freqscope{ |target, outbus... args, kwargs|
   this.performArgs(\play, args, [\target, server] ++ kwargs);
   ^Freqscope(server: target, outbus: outbus)
}
```

This way you can call it like this...

```supercolloder
{...}.freqscope(s, bus, args: [\freq, 440], blockSize: 4, overSample: 2)
```
It is good for passing along keyword arguments to another interface because it doesn't require naming all the arguments; meaning if the interface for `play` changes, this method stays the same.

dyfer:
> Although it does seem one is unable to free the synth object afterwards? That seems like a bug?

Ah yes, probably :)

After all I think this is very much a "quick a dirty" method to make sound + see it on the freqscope. I don't think it's really worth it for any "production" use. Maybe it's not really worth it to touch it, so that it stays very limited, reinforcing its rudimentary status?

JordanHendersonMusic:
> this is very much a "quick a dirty" method to make sound + see it on the freqscope

I agree, although a minor oversight, not returning the synth instance isn't a big deal in this case.

Also, I don't think freqscope quite works the way this PR is expecting it to. Consider this...
```supercollider
s.boot

{SinOsc.ar}.play

{WhiteNoise.ar * 0.01}.freqscope

``` 
That `freqscope` will have both white noise and the sine wave in. This is because freqscope isn't really attached to the synth's output, but to a busses' final output. This method is a bit confusing as by writing `{}.freqscope` it semantically suggests inspecting the function's output, or at a minimum, inspecting a busses' output at a certain point in the graph, not the final output of the bus.

---

I do think it would be useful to pass on arguments —particularly blockSize and upsample (when merged) —, which seems like the key point of your (@prko) PR.

dyfer:
> Also, I don't think freqscope quite works the way this PR is expecting it to.

That's true! 
As this is a helper function, I don't think it's worth spending much time on fixing this. A possible solution would wrap a function so that the output is put on a private bus which is then scoped, and sent to the output... but this becomes even more cumbersome in terms of resource manegement.

I agree that this PR can just add the arguments to the function.

SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
adcxyz: > ```supercollider
> ~demo.("SinOsc.ar");
> ~demo.("LFTri.ar");
> ~demo.("LFPulse.ar");
> ~demo.("LFPulse.ar(width: 0.8)");
> ~demo.("Impulse.ar");
> ~demo.("Impulse.ar(1)");
> ```

FWIW, your example could be written easily with JITLib:
```
// Here your use case with JITLib:
(
Window.closeAll;
s.waitForBoot {
	~freqscope = FreqScope();
	~scope = s.scope;
	~fsbounds = ~freqscope.window.bounds;
	~scbounds = Rect(~fsbounds.right + 5, 0, 300, ~fsbounds.height);
	~scope.window.bounds = ~scbounds;
	Ndef(\demo).play.scope;
};
)

Ndef(\demo, { SinOsc.ar });
Ndef(\demo, { LFTri.ar ! 2 });
Ndef(\demo, { LFPulse.ar(width: 0.8) ! 2 });
Ndef(\demo, { Impulse.ar ! 2 });
Ndef(\demo, { Impulse.ar(1) ! 2 });
```

prko: >Is that a correct understanding of what you want this for?

Yes, that is correct.  

>and is this material for introductory tutorials for beginners, yes?

Yes, this material is intended to provide introductory tutorials for beginners.  

The reason why I do not use `Ndef` is simply due to the number of characters per line. I aim to structure the waveform name, waveform, spectrum, explanation, and example code as follows:

Screenshot 2025-04-23 at 18 08 10

Currently I use the following code:
```
(Window.closeAll;
s.waitForBoot { var spectrum = FreqScope(), bounds = spectrum.bounds, width = bounds.width, height = bounds.height;
    ~demo = { |aCode|
        var function = { aCode.interpret * Env.pairs([[0, 0, -9], [1, 1, 9], [5, 1], [6, 0]]).kr(2) * 0.1 };
        Stethoscope(s, 1, 0, s.sampleRate / 440 * 2).bounds_(Rect(width + 1, 0, height, height)).yZoom_(4);
        spectrum.front; { function ! 2 }.play;
        function.plot(20 / 440, bounds: Rect(0, height + 30, width + height, height));
        function.plot(0.07, bounds: Rect(0, height + 30 * 2, Window.screenBounds.width, height)).plotMode_(\dstems);
    }
})
```
```
~demo.("SinOsc.ar")
~demo.("LFTri.ar")
~demo.("LFPulse.ar")
```
I would prefer not to place the code in quotation marks and would like to use `Ndef(\demo, { SinOsc.ar });`. However, doing so would require two columns per table, which would add one or two more pages. My intention is simply to minimise the number of pages and provide readers with a concise overview, presenting examples within just a few pages.

dyfer:
I'm fine with closing this PR.

> However, I am uncertain whether this would be broadly useful in general cases.

I'm not sure this is something to include in the main library, but I think this would be a good functionality to include in your own Quark. Writing extensions for specific use cases is what Quarks system is better for.

PR #6744, opened 2025-04-18
TestServer_clientID_booted: wait for the process to exit

dyfer:
## Purpose and Motivation

Like with .quitSync, we should wait for the scsynth to exit before continuing with the tests.

## Types of changes

- Bug fix (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/testsuite/classlibrary/TestServer_clientID_booted.sc
+++ b/testsuite/classlibrary/TestServer_clientID_booted.sc
@@ -129,13 +129,14 @@ TestServer_clientID_booted : UnitTest {
 	// first login initializes fully,
 	// second login from same address gets the same clientID, and sets running state.
 	test_remoteLogin {
-		var options, serverPid, remote1, cond, timer;
+		var options, serverPid, remote1, cond, timer, exitCond, testsFinished;
 
 		options = ServerOptions.new;
 		options.maxLogins = 4;
 
+		exitCond = CondVar.new;
 		// simulate a remote server process by starting a server process independently of SC
-		serverPid = unixCmd(Server.program + options.asOptionsString);
+		serverPid = unixCmd(Server.program + options.asOptionsString, { exitCond.signalOne });
 
 		// login with standard Server.remote method to test for
 		remote1 = Server.remote(\remoteLogin1, options: options, clientID: 3);
@@ -159,17 +160,20 @@ TestServer_clientID_booted : UnitTest {
 
 		remote1.remove;
 		thisProcess.platform.killProcessByID(serverPid);
+		testsFinished = exitCond.waitFor(5); // wait for the server process to exit
+		this.assert(testsFinished, "TIMEOUT: server process did not quit");
 	}
 
 	test_repeatedRemoteLogin {
 
-		var options, serverPid, remote1, remote2, cond, timer;
+		var options, serverPid, remote1, remote2, cond, timer, exitCond, testsFinished;
 
 		options = ServerOptions.new;
 		options.maxLogins = 4;
 
+		exitCond = CondVar.new;
 		// simulate a remote server process by starting a server process independently of SC
-		serverPid = unixCmd(Server.program + options.asOptionsString);
+		serverPid = unixCmd(Server.program + options.asOptionsString, { exitCond.signalOne });
 
 		// login with standard Server.remote method to test for
 		remote1 = Server.remote(\remoteLogin1, options: options, clientID: 3);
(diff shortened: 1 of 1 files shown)


testsuite/classlibrary/TestServer_clientID_booted.sc
```diff
@@ -159,17 +160,19 @@ TestServer_clientID_booted : UnitTest {
 
 		remote1.remove;
 		thisProcess.platform.killProcessByID(serverPid);
+		exitCond.hang; // wait for the server process to exit
```
mtmccrea, line 163: Would we ever want this to time out so it doesn't hang indefinitely?

If so, using a mechanism like `CondVar:-waitFor` may be useful.

dyfer: Good idea! Implemented.

testsuite/classlibrary/TestServer_clientID_booted.sc
```diff
@@ -159,17 +160,19 @@ TestServer_clientID_booted : UnitTest {
 
 		remote1.remove;
 		thisProcess.platform.killProcessByID(serverPid);
+		exitCond.waitFor(5); // wait for the server process to exit
```
mtmccrea, line 163: Sorry I could have been more thorough in my previous comment.

If this does time out, it could be good to catch it. I did this in `TestCoreUGens`:
https://github.com/supercollider/supercollider/blob/7b97f535befdd4899bb1f2f4b9c6fd92b84a7870/testsuite/classlibrary/TestCoreUGens.sc#L342-L343

But there could be a more proper pattern to use.

testsuite/classlibrary/TestServer_clientID_booted.sc
```diff
@@ -129,13 +129,14 @@ TestServer_clientID_booted : UnitTest {
...
-		var options, serverPid, remote1, cond, timer;
+		var options, serverPid, remote1, cond, timer, exitCond;
 
 		options = ServerOptions.new;
 		options.maxLogins = 4;
 
+		exitCond = CondVar.new;
 		// simulate a remote server process by starting a server process independently of SC
-		serverPid = unixCmd(Server.program + options.asOptionsString);
+		serverPid = unixCmd(Server.program + options.asOptionsString, {exitCond.signalOne});
```
mtmccrea, line 139: minor: spaces within curly braces: `{ exitCond.signalOne }`

testsuite/classlibrary/TestServer_clientID_booted.sc
```diff
@@ -159,17 +160,19 @@ TestServer_clientID_booted : UnitTest {
...
-		var options, serverPid, remote1, remote2, cond, timer;
+		var options, serverPid, remote1, remote2, cond, timer, exitCond;
 
 		options = ServerOptions.new;
 		options.maxLogins = 4;
 
+		exitCond = CondVar.new;
 		// simulate a remote server process by starting a server process independently of SC
-		serverPid = unixCmd(Server.program + options.asOptionsString);
+		serverPid = unixCmd(Server.program + options.asOptionsString, {exitCond.signalOne});
```
mtmccrea, line 175: { exitCond.signalOne }

testsuite/classlibrary/TestServer_clientID_booted.sc
```diff
@@ -215,5 +218,6 @@ TestServer_clientID_booted : UnitTest {
 		remote1.remove;
 		remote2.remove;
 		thisProcess.platform.killProcessByID(serverPid);
+		exitCond.waitFor(5); // wait for the server process to exit
```
mtmccrea, line 221: See comment above w.r.t. catching the timeout

testsuite/classlibrary/TestServer_clientID_booted.sc
dyfer: Got it! How's it now?

mtmccrea: Looks good. One detail I included in the timeout message in my example above was the name of the failing test. 

This may not actually be needed if `assert` also posts the method that called it when it throws... are you able to test that (like with a 0 or ver short timeout)?

dyfer: Yes, it does post, that's why I omitted it: `FAIL: a TestServer_clientID_booted: test_repeatedRemoteLogin - TIMEOUT: server process did not quit`

mtmccrea (approved):

Issue #6746, opened 2025-04-18
osc bytes padding inconsistency scsynth vs supernova


andiejs:
I'm reporting a tiny inconsistency with the way osc scores are handled between scsynth and supernova. Scsynth is more permissive and supernova fails on this tiny detail. This is not an urgent issue for me since I was able to change my osc score production to account for it, but I thought I should report it anyway. 

The function .asRawOSC is adding extra bytes to d_recv synthdefs, such that
the output len is always len mod 4 = 3.

For example, compare the output of 
```
SynthDef(\NRTsineBYTES, { |out, freq = 440|
            Out.ar(out, SinOsc.ar(freq, 0, 0.2))
        }).asBytes

```
and 

```
[0.0, ['/d_recv',
        SynthDef(\NRTsineBYTES, { |out, freq = 440|
            Out.ar(out, SinOsc.ar(freq, 0, 0.2))
        }).asBytes
	]].asRawOSC
```

The result is 36 bytes at the beginning (as it should be), plus two arbitrary bytes at the end. Now simply change the name of the synthdef so that it has a different number of bytes, to get a different number of arbitrary padding bytes.

Scsynth does not seem to care about them either way, while supernova dumps core without explanation if these bytes are missing. I noticed this because I use my own osc writer that was modeled on supercollider's osc writer ~4 years ago, and my generated osc continues to work with scsynth but was not working with supernova. So I suspect that supercollider didn't add these extra bytes a few years ago and older supercollider osc scores may also not work with supernova (just a hypothesis).

Spacechild1:
> The function .asRawOSC is adding extra bytes to d_recv synthdefs, such that
the output len is always len mod 4 = 3.

Yes. The size of an OSC message *must* be a multiple of 4 according to the OSC standard.

If the message size is not a multiple of 4, the message is malformed and Supernova correctly rejects it. In fact, it's the underlying `oscpack` library that throws an exception.

The *actual* bug is that Supernova does not protect `server.run_nonrt_synthesis(args);` with a try/catch block. As a consequence, any exception caused by malformed OSC messages just aborts the program, instead of printing a helpful error message.

---

scsynth does not use `oscpack` for parsing OSC messages. In particular, the `nextOSCPacket` function does not properly validate the OSC message size. IMO it would be better if it would.

andiejs:
[...]

Issue #6728, opened 2025-04-14
Unify exit codes throughout the project


dyfer:
## Motivation

We don't have a defined set of exit codes to describe particular failures in SuperCollider, like "class library not compiled", "audio device not  available" etc. It would be good to codify some exit codes so that we get a better understanding of what went wrong.

## Description of Proposed Feature

Add an include file with a list of constants and document it.

Example:
```cpp
#define LIBRARY_NOT_COMPILED 43
// for the servers:
#define DEVICE_NOT_AVAILABLE 64
#define SAMPLERATE_NOT_AVAILABLE 65
```

## Plan for Implementation

TBD

JordanHendersonMusic:
Some thoughts...

C ref states that the code `0` or `EXIT_SUCCESS` are used to represent successful exit, but that the only way to, *across platform* represent a failure is by using `EXIT_FAILURE`. If we don't do this, it is probably a good idea to convert every non-zero value to `EXIT_FAILURE`. https://en.cppreference.com/w/cpp/utility/program/EXIT_status

Linux uses specific exit codes to mean different things — 'exit codes 1 - 2, 126 - 165, and 255, have special meanings' https://tldp.org/LDP/abs/html/exitcodes.html so it is probably best to avoid these.


Rather than macros, which pollute global namespace, I'd recommend this pattern. 
```c++
struct ErrorCode {
    static constexpr std::uint8_t PortAlreadyInUse = 3;
    static constexpr std::uint8_t LibraryNotCompiled = 4;
};

exit(ErrorCode::PortAlreadyInUse);
```
Enums are another option, but they too pollute the global namespace unless they are marked as `enum class`, which would require a cast.

JordanHendersonMusic:
Just found Linux returning 11 for a segfault, so maybe the higher numbers should be used instead.

dyfer:
I think we can start around 32 (if we want to have that 43 in range...) or maybe 64.

Spacechild1:
> Rather than macros, which pollute global namespace, I'd recommend this pattern.
> ```
> struct ErrorCode {
>    static constexpr std::uint8_t PortAlreadyInUse = 3;
>    static constexpr std::uint8_t LibraryNotCompiled = 4;
> };
> ```

I would recommend to use a namespace instead:

```
namespace ErrorCode {
   constexpr int PortAlreadyInUse = 3;
   constexpr int LibraryNotCompiled = 4;
};
```

PR #6748, opened 2025-04-19
supernova: catch exceptions in NRT synthesis

Spacechild1:
1. Supernova: catch exceptions in NRT synthesis and post helpful error message, instead of silently aborting the program.
  Also make sure that the soundfile backend is automatically closed in the destructor so we can rely on RAII.
  Fixes https://github.com/supercollider/supercollider/issues/6746

2. Scsynth: call `World_CleanUp()` after `World_NonRealTimeSynthesis()` throws an exception to prevent hang

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/server/supernova/server/main.cpp
+++ b/server/supernova/server/main.cpp
@@ -377,8 +377,14 @@ int supernova_main(int argc, char* argv[]) {
                 exit(1);
             }
             EventLoop::run([&server]() { server.run(); });
-        } else
-            server.run_nonrt_synthesis(args);
+        } else {
+            try {
+                server.run_nonrt_synthesis(args);
+            } catch (exception const& e) {
+                cout << "Non-rt synthesis failed: " << e.what() << endl;
+                exit(1);
+            }
+        }
     } catch (const std::exception& exc) {
         cout << "\n*** ERROR: in main(): " << exc.what() << endl;
         exit(1);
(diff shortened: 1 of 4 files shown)


server/supernova/server/main.cpp
```diff
@@ -377,8 +377,14 @@ int supernova_main(int argc, char* argv[]) {
...
             }
             EventLoop::run([&server]() { server.run(); });
-        } else
-            server.run_nonrt_synthesis(args);
+        } else {
+            try {
+                server.run_nonrt_synthesis(args);
+            } catch (exception const& e) {
+                cout << "Non-rt synthesis failed: " << e.what() << endl;
+                exit(1);
```
dyfer, line 385: I'm not the one to review C++, but how about `exit(EXIT_FAILURE )`? We discussed unifying exit codes in #6728 so hopefully that'll happen at some point.

Spacechild1: > but how about exit(EXIT_FAILURE)

I chose `exit(1)` for consistency with the remaining code.

Spacechild1:
I just added a small NRT fix for scsynth.

dyfer (approved):
Functionally this PR seems not to break anything :) NRT still works, for both supernova and scsynth.

PR #6749, opened 2025-04-19
sclang: Refactor bytecodes

JordanHendersonMusic:
## Purpose and Motivation

The byte codes are hard to read in the code.

Rather than do all the bit shifting and magic numbers in the parse nodes classes, this moves the responsibility into dedicated opcode classes, making things easier to read.

This would not have been possible without Moss Heim's work documenting the byte codes. Their documentation is copied into the source code as comments above the respective opcodes so when you hover over the opcode in an IDE, their helpful descriptions appears in the pop-up.

## Types of changes

- Refactor

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


JordanHendersonMusic:
Hi @mossheim , just wanted to tag you directly since this is not only built of your work, but actually copied some of your text, which I've (hopefully correctly) attributed. I also sent you an email a few days ago if you'd rather reply there. 

---

@scztt I was wondering what you thought of this approach? The goal is to make it explicit what is being emitted. It is obviously not finished, but I also intend to rework some of the main interpreter code with some macros to unite the computer gotos and case statements with these classes. There is also the jump bytecodes to do. The tests failing don't appear to be relevant and pass locally.

capital-G:
Thanks so much for this! I must admit that I don't have any clue about the inner mechanics here at all, but I appreciate someone digging through it, making it more transparent and documenting it, this will help the longevity and maintainability :)

Maybe some feedback from @lfnoise would be interesting here if he would be up for it? :)

PS: Just saw that you checked in a `CMakePresets.json` - is this intentional? I think this should be part of another PR b/c this would need to be cross-platform etc. ;)

capital-G:
Just to link moss document here, which is licensed under CC-BY-SA-4.0: https://gist.github.com/mossheim/939297a6421a4ea6c5f96cf3f1f0069e 

CC-BY-SA-4.0 seems to be compatible with our GPLv3, see https://wiki.creativecommons.org/wiki/ShareAlike_compatibility:_GPLv3, and [quoting](https://creativecommons.org/share-your-work/licensing-considerations/compatible-licenses/)

> GPLv3: The [GNU General Public License version 3 ](https://www.gnu.org/copyleft/gpl.html)was declared a “BY-SA–Compatible License” for version 4.0 on 8 October 2015. Note that compatibility with the GPLv3 is one-way only, which means you may license your contributions to adaptations of BY-SA 4.0 materials under GPLv3, but you may not license your contributions to adaptations of GPLv3 projects under BY-SA 4.0. Other [special considerations](https://wiki.creativecommons.org/wiki/ShareAlike_compatibility:_GPLv3#Considerations_for_adapters_applying_the_GPLv3) apply. See the full [analysis](https://wiki.creativecommons.org/wiki/ShareAlike_compatibility:_GPLv3) and [comparison](https://wiki.creativecommons.org/wiki/ShareAlike_compatibility_analysis:_GPL) for more information.

Nonetheless, it would still be nice to have some kind of blessing from moss directly for this as this is indeed a very valuable and labor intensive contribution :)

One question though: Should this be part of the upcoming 3.14 release or is this something more towards 3.15?

JordanHendersonMusic:
Re the release: I don't think it matters because this doesn't change behaviour --- as far as I can tell that is; I've done a fair amount of testing and can't find any mistakes. Besides, mistakes in his part of the code are usually really easy to find because the interpreter just blows up or the unit tests do something weird. In an odd way, this is quite a safe place to make changes because we have very high code coverage and errors are catastrophic and obvious.

capital-G:
Regarding on how to get this merged: As this is a purely internal re-structuring which wouldn't change anything for the user, would it be okay with you to merge this after a 3.14 release?

IIUC it seems to change code on a type level and therefore shouldn't change behavior, but I'd like to reduce the currently moving parts and focussing on ironing out bugs for the 3.14 release - afterwards we can go wild again and we'll have some time to testing things before 3.15 releases - so in a similar manner like #6702.

JordanHendersonMusic:
@capital-G sure that sounds absolutely fine!

lang/LangSource/SpecialSelectorsOperatorsAndClasses.h
```diff
@@ -0,0 +1,342 @@
+#pragma once
+
+// Note: this is not an extensive list of all the special selectors, operators or classes, but represent the raw enum
+// used to represent them. In future, some refactoring should be done to unify these into separate documents so it is
+// easy to understand at a glance.
+
+/* special unary math operators */
+enum {
```
scztt, line 8: Minor, but - any reason not to use `enum class op` for these, so they become `op::neg` etc? It's a minor amount of safety / modernization, without much effect on code readability.

JordanHendersonMusic: So these are the original, I've just moved them here in one place. I'd like to do something like that though, but this pr is already quite large.

capital-G:
> @capital-G are we merging things for 15 into development yet?

Yes :)

JordanHendersonMusic:
Could I get a review on this so it can be merged please?

capital-G:
@JordanHendersonMusic can you rebase against develop? I have some problems in my IDE b/c some header files are not picked up, but they are picked up on develop.

JordanHendersonMusic:
@capital-G all done!

JordanHendersonMusic:
@capital-G @dyfer can we merge this now?

capital-G (approved):
Yes - lets merge! :)

PR #6750, opened 2025-04-20
Help: EnvGen: Add filter modulation example (re: new init behavior)

jamshark70:
## Purpose and Motivation

Document solutions for a problem with EnvGen modulation of filter cutoff, which was exposed by #6175.

Previously, a short attack would cause EnvGen's first output sample to be close to the target of the first segment. This was bad for most use cases but good for filter cutoff because it reduced the slope of the first segment, avoiding a too-sudden ramp.

There's a nice approach that isn't immediately obvious -- this PR adds that to EnvGen help..

## Types of changes

- Documentation

## To-do list

- [ ] ~~Code is tested~~
- [ ] ~~All tests are passing~~
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/EnvGen.schelp
+++ b/HelpSource/Classes/EnvGen.schelp
@@ -292,9 +292,15 @@ code::
 
 subsection::Filter cutoff modulation and initial envelope level
 
-When using an envelope to modulate a filter's cutoff/center frequency, at higher resonance settings, a short attack time (below about 0.05 sec) can produce a spike in volume. If this effect is not desired, and the filter's attack was always at the beginning of the synth (as is usually the case when playing a pattern), the artifact can be avoided by setting the envelope's initial value to a higher number for shorter attack times.
+When using an envelope to modulate a filter's cutoff/center frequency, at higher resonance settings, a short attack time can produce a spike in volume.
+If this effect is not desired, and the filter's attack is always at the beginning of the synth (as is usually the case when playing a pattern), the artifact can be avoided by setting the envelope's initial level closer to target level for shorter attack times (below about 0.05 sec).
 
-Here, attack times between 0 and 0.05 seconds will produce initial envelope values between 1.0 (attack time = 0 means no attack) and 0.0. If the attack time is greater than 0.05 seconds, the envelope will start at 0.0. (The envelope is between 0.0 and 1.0 and mapped onto a valid frequency range.)
+The following example shows how to do this in a SynthDef:
+Attack times between 0 and 0.05 seconds will produce initial envelope values between 1.0 (attack time = 0 means no attack) and 0.0.
+If the attack time is greater than 0.05 seconds, the envelope will start at 0.0.
+(The envelope is between 0.0 and 1.0 and mapped onto a valid frequency range.)
+
+To hear the (potentially strong::loud::) short-attack artifacts, edit the code::var envInit:: line to read code::var envInit = 0;::.
 
 code::
 (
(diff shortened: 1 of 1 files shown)


HotwheelsSisyphus:
I wonder if it's conceivable to remove the UI and Bus and instead have the \atk be driven by a Pseq or such? E.g. 
`\atk, Pseq((0..10).lincurve(0, 10, 0, 0.12, 4);` or write out Pseq([0, 0.01, 0.02, 0.03, 0.05, 0.08, 0.12])`;

Background is, there is this idea floating around (not super official yet) that the info in the docs should try to be less "distributed" in the sense that a reader's understanding of EnvGen behavior shouldn't depend e.g. on their familiarity with the Bus class or UI (in this instance).

dyfer:
@HotwheelsSisyphus Patterns are much more of an advanced concept than a `Bus` for me, but that's probably different for every user.

If the example uses a Pattern, then I'm certainly fine with a Bus; I like GUIs so I don't mind that either.

If we wanted a basic example, then it shouldn't use neither Pseq, nor a GUI or a Bus. 

Personally I think the examples are great and they deal with an edge case anyway, which I'd consider a more advanced concept.

HotwheelsSisyphus:
Oh I didn't mean to say that the examples aren't good! Apologies, I should have been clearer. ("consider doing x" sounds like "you should do x", but I meant it more as, "what do you think about x"?, to which James can reply "nah" if he wants to.)

This was less about "advanced" vs. "beginner" but more about keeping examples focused as much as possible on what they should be demonstrating. Since there was already a Pbind in there, I figured a Pseq is not a big step away from that, _but_ one can demonstrate that EnvGen bit without the GUI and the bus, that's all.

This is probably a wider design question about the docs that we haven't actually sufficiently discussed. I didn't mean to overreach here.

dyfer:
FWIW I like the interactivity of the example, over a preset set of values. Alternatively this could've been done with e.g. MouseX instead of the gui.

> This is probably a wider design question about the docs that we haven't actually sufficiently discussed.

Yes, something to discuss outside of this PR.

dyfer (approved):
IMO this can go in as is, but feel free to further tweak it if desired.

jamshark70:
> FWIW I like the interactivity of the example, over a preset set of values. Alternatively this could've been done with e.g. MouseX instead of the gui.

Right, I had thought it would be more useful for readers to have control over the position within the range.

I'd considered MouseX but that's problematic with Wayland, isn't it? (Which I'm not using [yet] but who knows when...?)

dyfer:
> I'd considered MouseX but that's problematic with Wayland, isn't it?

Yes it is, unfortunately. And AFAIU there's no straightforward solution in sight, with Wayland architecture.

HotwheelsSisyphus (commented):
Thanks for this addition!

Apologies, I'm new to this and thought these comments had sent yesterday. These are all suggestions, was trying to make myself useful but if you feel like the changes aren't worth the time, that's fine. The intention certainly wasn't to step on anyone's toes. They're more in the spirit of a msword track-changes prose editing workflow which is a bit at odds with git, I guess, where it seems overly fussy.

Anyway if all the changes seem reasonable to you, you should be able to just copypaste the bits under "result" in the comments.

One general thing: The upcoming schelp style guide will suggest a max of one sentence per line for diff reasons. That guideline technically not yet in place and there are of course tons of existing help files that don't comply with this, but there does seem to be general agreement on that particular point.

Thanks!

jamshark70:
OK, I've checked in the changes recommended by @HotwheelsSisyphus -- btw I very much appreciate the time taken to include the results of the edit -- saved me some time to just copy/paste it in.

Issue #6753, opened 2025-04-20
Make PathName and String fully interchangeable


adcxyz:
## Motivation

PathName is a wrapper for a String object that adds functionality for handling file paths. 
It is used in much existing code, so we have to be careful to keep backwards compatibility. 
As noted in #6752, it's name leads one to expect that pathNames will just work as file paths, 
but they don't. 
Ideally, PathName and String should be fully interchangeable everywhere. 

## Description of Proposed Feature
This can be achieved by 
1. making the interface identical:
- Implement the functionality of all PathName methods in String, and
- Redirect all PathName methods  to their corresponding String method. 
Then String objects can be used to replace PathName objects everywhere.

2. implementing a conversion step and using it widely:
- implement .asPath for PathName and String, as in asTarget, asGroup, .asUGenInput, etc. 
- use it in all methods that will directly use a file path.  

## Plan for Implementation

I volunteer to implement this, PR to follow soon. 

Currently making good progress with Step 1:
Wrote TestPathName for most methods, and they pass; 
many methods in PathName updated and simplified. 
PR 1 can be the tests;
PR 2 can be the method updates + conversion methods;
PR 3 can be the move to redirect, then TestPathName will 
test both String methods and redirections from PathName.

PR 4 will be adding conversion step to all lowest-level read/write methods, 
because there the primitives expect strings.

PR #6755, opened 2025-04-20
Unify path name and string to simplify filepath handling

adcxyz:
## Purpose and Motivation

This PR proposes to make file path handling in SC easier and more consistent. 
File paths in SC are Strings, and PathName is a wrapper for a string that adds functionality for handling file paths.
As noted in  #6752 and #6753, the name PathName leads one to expect that pathNames will just work as file paths, but ATM they don't.

Ideally, PathName and String should be fully interchangeable anywhere.
As PathName used in much existing code, we have to keep backwards compatibility.

- This PR moves all method functionality from PathName to String, 
and redirects  the PathName methods to their String versions, 
so String objects can be used in place of PathName objects everywhere.
- This PR adds tests for all PathName methods which test the redirections.

For full interchangeability, I see two options:
1. Have PathName.new always return a String immediately. 
Then these strings directly work as filepaths anywhere. 
This requires no change to any file access methods,
it even allows to remove the PathName redirection methods again, 
as any path generated with PathName will already be a standardized String, 
and thus work direcly in any and all file access operations.
In effect, this removes unneeded complexity introduced by PathName. 

2. Keep PathName objects as they are now, 
and add a conversion step PathName:asPath, analogous to asGroup, asUGenInput, etc 
to convert them to Strings in all lowest-level file access methods in the class library.

I prefer option 1 :-) 
Opinions welcome!

## Types of changes
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] add interchangeability option 1 or 2, asPath conversion or direct string option
- [ ] Update documentation
- [ ] Make PR ready for review 

comp: classlib

### DIFF
--- a/SCClassLibrary/Common/Collections/String.sc
+++ b/SCClassLibrary/Common/Collections/String.sc
@@ -447,7 +447,8 @@ String[char] : RawArray {
 		var sep = thisProcess.platform.pathSeparator;
 		var hasLeftSep, hasRightSep;
 
-		if (path.respondsTo(\fullPath)) {
+		// if second arg is PathName, make it a PathName
+		if (path.isKindOf(PathName)) {
 			^PathName(this +/+ path.fullPath)
 		};
 
@@ -468,7 +469,19 @@ String[char] : RawArray {
 	}
 
 	asRelativePath { |relativeTo|
-		^PathName(this).asRelativePath(relativeTo)
+		var r, a, b, i, mePath;
+		mePath = this.absolutePath;
+		relativeTo = (relativeTo ? PathName.scroot ).absolutePath;
+		r = Platform.pathSeparator;
+
+		a = mePath.split(r);
+		b = relativeTo.split(r);
+
+		i = 0;
+		while { a[i] == b[i] and: { i < a.size } } {
+			i = i + 1;
+		};
+		^(".." ++ r).dup(b.size - i).join ++ a[i..].join(r)
 	}
 	asAbsolutePath {
 			// changed because there is no need to create a separate object
--- a/SCClassLibrary/Common/Files/PathName.sc
+++ b/SCClassLibrary/Common/Files/PathName.sc
@@ -1,6 +1,6 @@
 PathName {
 
-	var <fullPath, colonIndices;
+	var <fullPath;
 
 	classvar <>scroot;
 	classvar <>tmp;
@@ -17,236 +17,68 @@ PathName {
 		})
 	}
 
-	colonIndices {
-		^colonIndices ?? {
-			colonIndices = List.new;
-			fullPath.do({ | eachChar, i |
-				if(eachChar.isPathSeparator, { colonIndices.add(i) })
-			});
-			colonIndices
-		}
+	asPath { ^fullPath }
+
+	// old PathName methods, colon was Mac OS9 path separator...
+	colonIndices { ^fullPath.separatorIndices }
+	lastColonIndex { ^fullPath.lastSeparatorIndex }
+	// new better names
+	separatorIndices { ^fullPath.separatorIndices }
+	lastSeparatorIndex { ^fullPath.lastSeparatorIndex }
+
+	fileName { ^fullPath.fileName }
+	fileNameWithoutExtension { ^fullPath.fileNameWithoutExtension }
+	fileNameWithoutDoubleExtension { ^fullPath.fileNameWithoutDoubleExtension }
+	extension { ^fullPath.extension }
+
+	dirname { ^fullPath.dirname }
+	pathOnly { ^fullPath.pathOnly }
+
+	diskName { ^fullPath.diskName }
+
+	isRelativePath { ^fullPath.isRelativePath }
+	isAbsolutePath { ^fullPath.isAbsolutePath }
+	absolutePath { ^fullPath.absolutePath }
+
+	asRelativePath { |relativeTo| ^fullPath.asRelativePath(relativeTo) }
+	asAbsolutePath { ^fullPath.asAbsolutePath }
+
+	allFolders { ^fullPath.allFolders }
+
+	folderName { ^fullPath.folderName }
+
+	nextName { ^fullPath.nextName }
+	noEndNumbers { ^fullPath.noEndNumbers }
+	endNumber { ^fullPath.endNumber }
+	endNumberIndex { ^fullPath.endNumberIndex }
+
+	== { |other| { ^fullPath.absolutePath }
+		^other.respondsTo(\fullPath) and: { other.fullPath == this.fullPath }
 	}
 
-	fileName {
-		^fullPath.copyRange((this.lastColonIndex) + 1, (fullPath.size -1).max(0))
-	}
-
-	fileNameWithoutExtension {
-		var fileName = this.fileName;
-		fileName.reverseDo({ | char, i |
-			if(char == $.,{
-				^fileName.copyRange(0,fileName.size - (i + 2))
-			})
-		});
-		^fileName
-	}
-
-	fileNameWithoutDoubleExtension {
-		var fileName, pathName;
-		fileName = this.fileNameWithoutExtension;
-		pathName = PathName( fileName );
-		^pathName.fileNameWithoutExtension
-	}
-
-	extension {
-		var fileName;
-		fileName = this.fileName;
-		fileName.reverseDo({ | char, i |
-			if(char == $., {
-				^fileName.copyRange(fileName.size - i,fileName.size - 1)
-			})
-		});
-		^""
-	}
-
-	pathOnly {
-		^fullPath.copyRange(0, this.lastColonIndex)
-	}
-
-	diskName {
-		^fullPath.copyRange(0, this.colonIndices.first - 1)
-	}
-
-	isRelativePath { ^this.isAbsolutePath.not }
-
-	isAbsolutePath {
-		^fullPath.at(0).isPathSeparator
-	}
-
-	absolutePath{
-		^fullPath.absolutePath
-	}
-
-	asRelativePath { |relativeTo|
-		var r, a, b, i, mePath;
-		mePath = this.fullPath.absolutePath;
-		relativeTo = (relativeTo ? scroot ).absolutePath;
-		r = thisProcess.platform.pathSeparator;
-
-		a = mePath.split(r);
-		b = relativeTo.split(r);
-
-		i = 0;
-		while { a[i] == b[i] and: { i < a.size } } {
-			i = i + 1;
-		};
-		^(".." ++ r).dup(b.size - i).join ++ a[i..].join(r)
-	}
-
-	asAbsolutePath {
-		if(this.isAbsolutePath,{
-			^fullPath
-		},{
-			// this assumes relative to the sc app
-			^fullPath.absolutePath
-		})
-	}
-
-	allFolders {
-		var folderNames, pathCopy;
-		folderNames = List.new;
-		pathCopy = fullPath;
-
-		this.colonIndices.doAdjacentPairs({ | startColon, endColon |
-			folderNames.add( fullPath.copyRange(startColon + 1, endColon - 1) )
-		});
-
-		^folderNames
-	}
-
-	folderName {
-		var indexBeforeFolder;
-		var ci = this.colonIndices;
-		if (ci.isEmpty, { ^"" });
-
-		indexBeforeFolder =
-		if (ci.size == 1, 0, { ci.at(ci.size - 2) + 1 });
-
-		^fullPath.copyRange(indexBeforeFolder, this.lastColonIndex - 1)
-	}
-
-	lastColonIndex {
-		var ci = this.colonIndices;
-		^if(ci.isEmpty, { -1 }, { ci.last }) ;
-	}
-
-	nextName {
-		^if(fullPath.last.isDecDigit, {
-			this.noEndNumbers ++ (this.endNumber + 1)
-		}, {
-			fullPath ++ "1"
-		})
-	}
-
-	noEndNumbers {
-		^fullPath[..this.endNumberIndex]
-	}
-
-	endNumber {	// turn consecutive digits at the end of fullPath into a number.
-		^fullPath[this.endNumberIndex + 1..].asInteger
-	}
-
-	endNumberIndex {
-		var index = fullPath.lastIndex;
-		while({
-			index > 0 and: { fullPath.at(index).isDecDigit }
-		}, {
-			index = index - 1
-		});
-		^index
-	}
-
-	/* concatenation */
+/* concatenation */
 	+/+ { | path |
 		var otherFullPath = path.respondsTo(\fullPath).if({ path.fullPath }, { path.asString });
 		^this.class.new(fullPath +/+ otherFullPath)
 	}
 
-	entries {
-		var path = fullPath;
-		if(path.isEmpty) { ^[] };
-		^pathMatch(path +/+ "*").collect({ | item | PathName(item) })
-	}
+	pathMatch { ^fullPath.pathMatch }
+	entries { ^fullPath.entries }
+	isFolder { ^fullPath.isFolder }
+	isFile { ^fullPath.isFile }
 
-	pathMatch {
-		^pathMatch(fullPath)
-	}
+	files { ^fullPath.files }
+	folders { ^fullPath.folders }
+	deepFiles { ^fullPath.deepFiles }
 
-	isFolder {
-		var path = this.pathMatch;
-		^if(path.notEmpty, {
-			path.at(0).last.isPathSeparator
-		}, { false })
-	}
+	parentPath { ^fullPath.parentPath }
 
-	isFile {
-		var path = this.pathMatch;
-		^if(path.notEmpty, {
-			path.at(0).last.isPathSeparator.not
-		}, { false })
-	}
+	filesDo { |func| ^fullPath.filesDo(func) }
 
-	files {
-		^this.entries.select({ | item | item.isFile })
-	}
-
-	folders {
-		^this.entries.select({ | item | item.isFolder })
-	}
-
-	deepFiles {
-		^this.entries.collect({ | item |
-			if(item.isFile, {
-				item
-			},{
-				item.deepFiles
-			})
-		}).flat
-	}
-
-	parentPath {
-		var ci = this.colonIndices;
-
-		^if((fullPath.last.isPathSeparator) and: { ci.size > 1 }, {
-			fullPath.copyRange(0, ci[ci.size - 2])
-		}, {
-			fullPath.copyRange(0, this.lastColonIndex)
-		})
-	}
-
-	filesDo { | func |
-		this.files.do(func);
-		this.folders.do { | pathname |
-			pathname.filesDo(func)
-		};
-	}
-
-	streamTree { | str, tabs = 0 |
-		str << this.fullPath << Char.nl;
-		this.files.do({ | item |
-			tabs.do({ str << Char.tab });
-			str << item.fileNameWithoutExtension  << Char.nl
-		});
-		this.folders.do({ | item |
-			item.streamTree(str, tabs + 1);
-		});
-	}
-
-	dumpTree {
-		this.streamTree(Post)
-	}
-
+	streamTree { | str, tabs = 0 | fullPath.streamTree(str, tabs) }
+	dumpTree { fullPath.dumpTree }
 	printOn { | stream |
-		stream << "PathName(" << fullPath << ")"
+		stream << "PathName(" << fullPath.cs << ")"
 	}
-
-	dumpToDoc { | title="Untitled" |
-		var str, doc;
-		doc = Document.new(title);
-		str = CollStream.new;
-		this.streamTree(str);
-		doc.string = str.collection;
-		^doc
-	}
-
+	dumpToDoc { | title="Untitled" | fullPath.dumpToDoc(title) }
 }
--- a/SCClassLibrary/Common/Files/String_pathName.sc
+++ b/SCClassLibrary/Common/Files/String_pathName.sc
@@ -4,8 +4,8 @@
 + String {
 
 	// // backwards compat to PathName:
-	// fullPath { ^this.asPath }
 	asPath { ^this.standardizePath }
+	fullPath { ^this.asPath }
 
 	// bad name, old macOS ...
 	colonIndices { ^this.separatorIndices }
@@ -65,9 +65,6 @@
 		^(".." ++ r).dup(b.size - i).join ++ a[i..].join(r)
 	}
 
-	// old PathName method
-	asAbsolutePath { ^this.absolutePath }
-
 	allFolders {
 		^this.dirname.split.selectAs (_ != "", List)
 	}
@@ -75,11 +72,15 @@
 	folderName { ^this.dirname.basename }
 
 	nextName {
-		^if(this.last.isDecDigit, {
-			this.noEndNumbers ++ (this.endNumber + 1)
-		}, {
-			this ++ "1"
-		})
+		var name, ext;
+		if (this.last.isDecDigit) {
+			^this.noEndNumbers ++ (this.endNumber + 1)
+		};
+		#name, ext = this.splitext;
+		if (ext.isNil) {
+			^name ++ "1"
+		};
+		^name.nextName ++ "." ++ ext
 	}
 
 	noEndNumbers {
--- a/SCClassLibrary/Common/Files/String_pathName.sc
+++ b/SCClassLibrary/Common/Files/String_pathName.sc
@@ -1,0 +1,172 @@
+// All methods from PathName moved to String class,
+// so PathName methods only redirect here.
+
++ String {
+
+	// // backwards compat to PathName:
+	// fullPath { ^this.asPath }
+	asPath { ^this.standardizePath }
+
+	// bad name, old macOS ...
+	colonIndices { ^this.separatorIndices }
+
+	lastSeparatorIndex {
+		this.reverseDo { |char, revIndex|
+			if (Platform.isPathSeparator(char)) {
+				^this.size - revIndex - 1
+			};
+		};
+		^-1
+	}
+	lastColonIndex { ^this.lastSeparatorIndex }
+
+	separatorIndices {
+		var res = List[];
+		this.do { |ch, i| if (ch.isPathSeparator) { res = res.add(i) } };
+		^res
+	}
+
+	//
+	fileName { ^this.basename }
+	fileNameWithoutExtension { ^this.basename.splitext[0] }
+	fileNameWithoutDoubleExtension {
+		^this.basename.splitext[0].splitext[0]
+	}
+	extension {
+		^this.splitext.last ? ""
+	}
+
+	// PathName:pathOnly ended with separator
+	pathOnly {
+		^this.dirname +/+ Platform.pathSeparator
+	}
+	// not sure this is useful in any way.
+	// but backwards compat.
+	diskName {
+		^this.copyRange(0, this.colonIndices.first - 1)
+	}
+
+	isRelativePath { ^this.isAbsolutePath.not }
+	isAbsolutePath { ^this.at(0).isPathSeparator }
+
+	asRelativePath { |relativeTo|
+		var r, a, b, i, mePath;
+		mePath = this.absolutePath;
+		relativeTo = (relativeTo ? PathName.scroot ).absolutePath;
+		r = Platform.pathSeparator;
+
+		a = mePath.split(r);
+		b = relativeTo.split(r);
+
+		i = 0;
+		while { a[i] == b[i] and: { i < a.size } } {
+			i = i + 1;
+		};
+		^(".." ++ r).dup(b.size - i).join ++ a[i..].join(r)
+	}
+
+	// old PathName method
+	asAbsolutePath { ^this.absolutePath }
+
+	allFolders {
+		^this.dirname.split.selectAs (_ != "", List)
+	}
+
+	folderName { ^this.dirname.basename }
+
+	nextName {
+		^if(this.last.isDecDigit, {
+			this.noEndNumbers ++ (this.endNumber + 1)
+		}, {
+			this ++ "1"
+		})
+	}
+
+	noEndNumbers {
+		^this[..this.endNumberIndex]
+	}
+
+	endNumber {	// turn consecutive digits at the end of this into a number.
+		^this[this.endNumberIndex + 1..].asInteger
+	}
+
+	endNumberIndex {
+		var index = this.lastIndex;
+		while({
+			index > 0 and: { this.at(index).isDecDigit }
+		}, {
+			index = index - 1
+		});
+		^index
+	}
+
+	entries { ^pathMatch(this +/+ "*") }
+	isFolder { ^this.last.isPathSeparator }
+
+	isFile {
+		var path = this.pathMatch;
+		^if(path.notEmpty, {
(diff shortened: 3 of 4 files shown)


dyfer:
@adcxyz the test `TestPathName:test_filePathMethods`  currently fails FYI.

dyfer:
I'm personally a little uneasy about returning a different object from `.new`. With all methods moved to `String` this _should_ work in many/most cases, but surely there's code out there checking if one passes a `PathName`... Still, if the code has handling for both `PathName` and a `String`, then this should still work; only code that explicitly expects PathName (`.isKindOf(PathName)`) would break, IIUC.

Pinging @telephon @muellmusik @jamshark70 for a possible review.

muellmusik:
> Pinging @telephon @muellmusik @jamshark70 for a possible review.

Firstly, I'd just like to say that I'd really like this to be cleaned up. It's quite confusing as is. So thanks @adcxyz for grasping the nettle!

Admittedly without having really traced it through, I wonder if a useful solution might be some version of making PathName a subclass of String. If this were done keeping most of the functionality in PathName, it could serve as a sort of String+ with extra path-related stuff. Not my preferred design choice if starting from scratch, but perhaps a fair compromise, and avoids making String even bigger?

This could also work if the functionality moves to String. In that case ```PathName : String``` is just a shell and should be deprecated. I like this better than Alberto's option 2 above. Tbh if everything's moving to String, I think PathName should be deprecated in any case.

dyfer:
I like the idea of moving all the functionality to String and deprecating `PathName`! `PathName` would still be a PathName class so it would provide full backwards compatibility (with deprecation warning).

muellmusik:
Yes that’s what I was thinking. Making it a String subclass would mean it could just be a stub.

adcxyz:
> @adcxyz the test `TestPathName:test_filePathMethods` currently fails FYI.

@dyfer - What was the error post ? 
On macOS 15.4.1 I get no errors.

I also find the idea of returning a String from PathName:new a bit scary ;-),
but it is much cleaner than dragging on PathName as a shell around strings.  
When all the methods behave the same (and TestPathName says they do for me), 
then really only explicit ```something.isKindOf(PathName)``` would fail...
which seems inconvenient, but not horrible. 

We can deprecate PathName, and leave it in for eternal backwards compatibility. 
All of this info would go into a PathName.help.

Subclassing String seems really tricky, and not at all sure it will even work:
I tried as below, and PathNames then fail to post themselves correctly; 
attempts to bypass/replace .asString turns loopy pretty quickly. 
```
PathName2 : String { }   // just a stub
// make a blank one
x = PathName2.newClear(12); // posting complains
x.isKindOf(String); // true
x.indexedSize; // 0 - has no indexed slots!

// bit more effort
```
PathName2 : String {
// avoid post failure, though not correct yet
	asString { ^"PathName2(%)".format(this.cs) } 
	printOn { | stream |
		stream << "PathName2(" << this.cs << ")"
	}
}

x = PathName2("asd"); 
-> PathName2("")
x.cs; 
-> ""
x.dump; // dumps OK now, but no instvars, and no indexed slots!
Instance of PathName2 {    (0x13060e788, gc=BC, fmt=00, flg=00, set=00)
  instance variables [0]
}
```
Once someone gets PathName2 to be a proper subclass with indexed slots ... 
We can test whether primitives even accept anything but Strings;
fails at the moment because PathName2 has no string content. 
```(
var f = File(PathName2("~/test.txt".standardizePath), "w");
f.write("Does this work?\n is this thing on ?\n");
f.close;
)```

Overall, subclassing seems to cause more difficulties, 
so probably not worth pursuing. Or am I missing something?

**Comment by muellmusik** [comment]
> Once someone gets PathName2 to be a proper subclass with indexed slots ...

> Overall, subclassing seems to cause more difficulties,
> so probably not worth pursuing. Or am I missing something?

Away from my machine so can't test, but is it just that it needs to be ```PathName2[slot] : String { }```?

**Comment by capital-G** [comment]
I agree that the status quo for paths could be better, but maybe we should lean towards a "full Path" instead of "full string" migration? Strings are IMO somehow flawed for representing paths due to the unix/windows mismatch of handling paths and their therefore different ways of handling escaping - and as sclang is multi-platform we should somehow take this into account?

We could instead use `String.toPath` as a fallback and use `Path` in all of our classlib? So e.g.

```supercollider
myMethod {|path|
  path = if(path.isKindOf(Path), {path}, {path.toPath});
}
```

At least Python tries to get rid of string for paths and replace it with a Path class - see https://docs.python.org/3/library/pathlib.html and to quote some of their reasoning behind this in https://peps.python.org/pep-0428/

> The rationale to represent filesystem paths using dedicated classes is the same as for other kinds of stateless objects, such as dates, times or IP addresses. Python has been slowly moving away from strictly replicating the C language’s APIs to providing better, more helpful abstractions around all kinds of common functionality.
> ...
> Indeed, many people will prefer handling dates and times using the high-level objects provided by the datetime module, rather than using numeric timestamps and the time module API. Moreover, using a dedicated class allows to enable desirable behaviours by default, for example the case insensitivity of Windows paths.

In the end `PathName` could also provide a namespace for methods which are only applicable to paths - and I think sclang would benefit from namespaces - but I am also not too familiar w/ the smalltalk-way-of-doing-things (message passing) and I am more familiar w/ the Java-esque OOP (method-invoking).

I currently don't have much capacity to dig further into this and just wanted to leave this as food for thought - either way, I think anything which makes this more consistent is worth doing and merging :)

dyfer:
> @dyfer - What was the error post ?

Check the "Some checks were not successful" section at the bottom of the PR or the "Checks" tab on top

JordanHendersonMusic:
@adcxyz 

As I briefly mentioned in the issue, inheriting from literals is often dangerous and should be avoided. This is because strings are stored with a unique tag in the slot, so once the language identifies how the data is stored, it is free to assume the class is a string. 

Interestingly, although perhaps not relevant here, making a literal inherit from another class is often fine. So `String : PathName` ***should*** work but `PathName : String` probably won't.

adcxyz:
> > Once someone gets PathName2 to be a proper subclass with indexed slots ...
> 
> > Overall, subclassing seems to cause more difficulties,
> > so probably not worth pursuing. Or am I missing something?
> 
> Away from my machine so can't test, but is it just that it needs to be `PathName2[slot] : String { }`?

Thanks, will try that!

adcxyz:
> @adcxyz
> 
> As I briefly mentioned in the issue, inheriting from literals is often dangerous and should be avoided. This is because strings are stored with a unique tag in the slot, so once the language identifies how the data is stored, it is free to assume the class is a string. This is also why you can't have instance variables, because it is trying to store it in a `PyrString` not a `PyrObject`.
> 
> Interestingly, although perhaps not relevant here, making a literal inherit from another class is often fine. So `String : PathName` _**should**_ work — so long as it has no instance variables — but `PathName : String` probably won't.

Yes, thank you, I was aware that instvars are impossible. 
I also considered inheriting the other way round, but I am not sure that would help here.

For me, keeping PathName and redirecting the work to String is also fine, 
so one can keep all path-related methods there for clarity.

muellmusik:
> As I briefly mentioned in the issue, inheriting from literals is often dangerous

Nice catch, @JordanHendersonMusic. Maybe we should add a note to the Writing Classes help file?

> Away from my machine so can't test, but is it just that it needs to be `PathName2[slot] : String { }`
> Thanks, will try that!

Can confirm that is the correct way to subclass a slotted class. I guess option 1 with a deprecation warning might be the cleanest solution?

adcxyz:
> We could instead use `String.toPath` as a fallback and use `Path` in all of our classlib? So e.g.
> 
> ```supercollider
> myMethod {|path|
>   path = if(path.isKindOf(Path), {path}, {path.toPath});
> }
> ```
The Smalltalk way would be an asPath method, like asGroup, asTarget, etc:
```supercollider
// in String:
   asPath { ^this }
// in PathName: 
   asPath { ^fullPath }
// then myMethod can convert everything that understands asPath to the right thing
   myMethod {|path|
      path = path.asPath; 
      // now 
> }
> ```

**Comment by adcxyz** [comment]
> Can confirm that is the correct way to subclass a slotted class. I guess option 1 with a deprecation warning might be the cleanest solution?

So, you mean like this?
```
PathName {
   *new { |string| 
        // deprecated warning, roughly "PathName is deprecated; PathName objects can be fully replaced with String objects." 
        ^string
   }
 ```
 and in PathName.help:
 - explain that PathName is kept for backwards compatibility, 
 - move PathName methods documentation to String? 
 - or better a separate Path Handling help page?

adcxyz:
@dyfer - thanks, found the failed message. 
Unfortunately, the CI does not keep detailed posts from the assert calls, so we don't see which of several asserts fails.
(In TestPathName I followed the testsuite pattern of grouping related tests into one test method.) 
Is there a recommended way to find out which one fails in the CI (fails only there, not on my macOS)? 
Converting them all to separate methods seems overkill.

capital-G:
> Is there a recommended way to find out which one fails in the CI (fails only there, not on my macOS)?

See https://github.com/supercollider/supercollider/actions/runs/14570045387/job/40866039814?pr=6755#step:7:454

Search for the test name not just in *Post test results*, but also open the tab *Run tests* and you get the full output of the test run.

adcxyz:
OK thanks, found it!

telephon (commented):
you could unify the syntax for if and while (omit parentheses everywhere). Also, some comments may be dropped for the final version. Otherwise, it looks good to me! (So you could make the docs to finish up).

adcxyz:
> I agree with @capital-G
> > but maybe we should lean towards a "full Path" instead of "full string" migration? Strings are IMO somehow flawed for representing paths due to the unix/windows mismatch of handling paths and their therefore different ways of handling escaping - and as sclang is multi-platform we should somehow take this into account?

- IIUC String handles differing multiplatform path separators in standardizePath;
- are there any other differences still open? 

>     1. a `string` is not necessarily a "path" so being able to/encouraged to call `.filesDo` on it (for example) feels squishy
>     2. `PathName` could become more like Python's `Path` from pathlib which handles paths OS agnostically but also has some overlap with SC's `PathName` such as `.fileName`, `.extension`, `.fileNameWithoutExtension`, etc. Python's `Path` has more features like this which are very useful.

Again, I am fine with keeping PathName objects alive; as all PathName methods are string ops underneath, 
I hope it is fine to keep the implementation there, and just reinstate the forwards to there. 
The downsides to keeping PathName are:
- subclassing String was strongly discouraged by @JordanHendersonMusic (and seems hairy to get to work). 
- to to allow PathNames in all file access methods (which was the starting point of this proposal),  
we need to put asPath in all those. fine with me. 

How should we best decide?

adcxyz:
For completeness, a related discussion from 4 years ago (which I missed then):
https://scsynth.org/t/pathname-string-and-file/3173

telephon:
Generally, objects are crossing points between interfaces (i.e. they implement several of them), and part of the creative power of programming is that the same object can travel between use cases. This is both a source of elegance and trouble, so I get your point, @tedmoore. The string class is naturally one of the most amphibian one.

If we could wrap methods around objects, this would be an easier way to go. As for now, we have the trade off of either converting fore and back (losing the combinatory flexibility) or accumulating interfaces.

Since String already has a few path related methods, it seems straightforward to extend it. I am not sure if the new set of methods is somewhat complete and minimal. That would be an argument for adding it to string as proposed here.

adcxyz:
Just so this does not get stuck, here is a proposal for a way forward: 

We add a Path class modeled on Python pathlib now, 
where a full path-related interface will go eventually.
We keep all implementations where they are for compatibility.
We deprecate PathName with a long timeline, and in its methods, 
point to the new recommended interface. 
We add one conversion method (e.g. asPathString) for all 
file access methods that use primitives and expect String objects. 

Does that make sense to everyone? 

```
Path {
	var <string;
	// only officially recommended method interface here, modeled on Python pathlib.
	// I.e. all general path-related methods from String, File etc,
	// and acceptable ones with OK names from PathName

	*new { |string=""| ^this.newCopyArgs(string.standardizePath) }

	// conversion to string to use in file access methods that have primitives
	*asPathString { ^string }

	// filename related
	dirname
	filename
	extension
	filenameWithoutExtension

	// tests
	exists { ^File.exists(string) }
	isRelative
	isAbsolute
	isFile
	isFolder

	// search
	match { ^string.pathMatch }
	// maybe add recursion depth param here?
	entries
	files
	folders
	entriesDo
	filesDo
	foldersDo

}

PathName : Path {
	// only keep backwards compatibility methods here,
	// with deprecation notices for *new, aliases,
	// methods w badly chosen names etc
	fullPath { ^string }
}

String {
	// keep all well-named string-related implementations here
	// deprecate others, pointing to new Path (or existing String) methods.
}
```

dyfer:
We discussed this PR at the dev meeting. The consensus was to move forward with the `Path` class. Feel free to take a look at https://en.cppreference.com/w/cpp/filesystem/path to decide which methods should be instance vs class methods.
Note: Path object should support concatenation with and without a separator, posting, and all the other goodies currently provided by String and PathName...

dyfer:
I was recently looking at `String -unixCmd` and `ArrayedCollection -unixCmd` and I humbly propose, that if we have a new Path object, then `Path -unixCmd` should probably use the ArrayedCollection method (which IMO is considered better? can't find the reference atm), which means it needs to pass arguments as an array.

adcxyz:
OK, I committed a first sketch of the Path class for discussion.

capital-G:
@adcxyz just for planing - do you think this should be included in the 3.14 release (deadline end of this month) or is this something which could also be part of 3.15?

SCClassLibrary/Common/Files/PathName.sc
```diff
@@ -65,28 +82,72 @@ Path {
...
+		^str[..this.endNumberIndex]
+	}
+
+	endNumber {	// turn consecutive digits at the end of this into a number.
+		^str[this.endNumberIndex + 1..].asInteger
+	}
+
+	endNumberIndex {
+		var index = str.lastIndex;
+		while({
```
telephon: We have a method for this, if you like:
```
"qwerty0612".lastIndexForWhich({ |char| char.isDecDigit.not })
```

capital-G:
Hey @adcxyz - there are some tests failing in the pipeline, see https://github.com/supercollider/supercollider/actions/runs/15399483330/job/43328692838?pr=6755#step:7:8273

```
	[2/4]	TestPathName:test_conversions (0.0s)
		 ! asAbsolutePath w relative filename
		 ! absolutePath w relative filename
	[0/1]	TestSCDocHTMLRenderer:test_htmlForLink_externalLink (0.01s)
		 ! ERROR: Message 'asRelativePath' not understood.
```


with the respective failures

```

	Test: test_conversions
FAIL: a TestPathName: test_conversions - asAbsolutePath w relative filename
Is:
	 /home/runner/work/supercollider/supercollider/relative.scd 
Should be:
	 //relative.scd


FAIL: a TestPathName: test_conversions - absolutePath w relative filename
Is:
	 /home/runner/work/supercollider/supercollider/relative.scd 
Should be:
	 //relative.scd
```

and

```

FAIL: a TestSCDocHTMLRenderer: test_htmlForLink_externalLink - ERROR: Message 'asRelativePath' not understood.
```

adcxyz:
thanks @capital-G, saw them - 
the first two are the same error; they both redirect to 
String.absolutePath, which uses File.getcwd. 
current working directory is different between macOS user account and the CI test setup.
I can adapt the test to compare with constructed results.

The second one I am trying to track down, but don't see how 
TestSCDocHTMLRenderer: test_htmlForLink_externalLink 
indirectly calls asRelativePath. 
TestSCDocHTMLRenderer.run is fine on macOS, and I can't find a single use 
of asRelativePath that htmlForLink in the class library would reach. puzzling.

Still missing: update of PathName.help

adcxyz:
> At the release candidate meeting we decided to push it back to 3.15 to make sure that the interface is _perfect_.

OK, then perfection it is :-)

Current State: 
- Path class as proposed is fully implemented
- Help files for Path and PathName done 
- Tests for Path (new interface) and PathName (old interface) all run

Help needed: Opinions on the new Path interface please!
- what is good as proposed, 
- what is still missing, 
- what should be different, 
- what could be improved ? 

TIA, adcxyz

HelpSource/Classes/Path.schelp
```diff
@@ -0,0 +1,305 @@
...
+code::
+Path("/Users/xyz/sc/myfile.scd").str;
+// backwards compatibility with PathName
+Path("/Users/xyz/sc/myfile.scd").fullPath;
+::
+
+METHOD:: name, fileName
+the filename of the path
+code::
+Path("/Users/xyz/sc/myfile.scd").name;
```
telephon, line 72: I find `name` doesn't speak very well ("path name" is not the file's name), `fileName` mightbe enough?

HelpSource/Classes/Path.schelp
```diff
@@ -0,0 +1,305 @@
...
+// backwards compatibility with PathName
+Path("/Users/xyz/sc/myfile.scd").fileName;
+::
+
+METHOD:: extension
+the extension of the filename or a blank string if none.
```
telephon, line 78: maybe better `nil` if none? Then it is easy to test for.

HelpSource/Classes/Path.schelp
```diff
@@ -0,0 +1,305 @@
...
+::
+
+METHOD:: fileNameWithoutDoubleExtension
+the filename of the path without up to two extensions
+code::
+Path("/Users/xyz/sc/myfile.preset.scd").fileNameWithoutDoubleExtension;
+::
+
+METHOD:: parent
+the parent folder as a Path
```
telephon: So far, sclang has often used the unix terminology of `directory` and `dir`. We need to decide what way to go.

HelpSource/Classes/Path.schelp
```diff
@@ -0,0 +1,305 @@
...
+::
+
+METHOD:: dirname
+the parent folder as a String with trailing slash
+code::
+// as String
+Path("/Users/xyz/sc/myfile.scd").dirname;
+::
+
+METHOD:: folderName
```
telephon, line 110: I find the distinction between folder and directory confusing. They are synonyms, but here they get different meanings?

HelpSource/Classes/Path.schelp
```diff
@@ -0,0 +1,305 @@
...
+code::
+Path("/Users/xyz/sc/myfile.scd").parts.postcs;
+::
+
+METHOD:: isFile, isFolder, isDir
+check whether a path is a file or a folder/dir
+code::
+Path("/Users/xyz/sc/myfile.scd").isFile;
+Path("/Users/xyz/sc/").isFile;
+Path("/Users/xyz/sc/myfile.scd").isFolder;
```
telephon, line 210: again, what is the difference between folder and dir?

SCClassLibrary/Common/Collections/String.sc
```diff
@@ -468,7 +469,19 @@ String[char] : RawArray {
 	}
 
 	asRelativePath { |relativeTo|
-		^PathName(this).asRelativePath(relativeTo)
+		var r, a, b, i, mePath;
+		mePath = this.absolutePath;
+		relativeTo = (relativeTo ? PathName.scroot ).absolutePath;
```
telephon, line 508: Why not use `Path`?

telephon: … and why is this in String and not in Path? It seems like a good candidate for moving over.

SCClassLibrary/Common/Collections/String.sc
```diff
@@ -468,7 +469,19 @@ String[char] : RawArray {
...
+		r = Platform.pathSeparator;
+
+		a = mePath.split(r);
+		b = relativeTo.split(r);
+
+		i = 0;
```
telephon, line 514: `a.count { |x, i| x == b[i] }` would be shorter

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+	// call for OSC communication
+	asOSCArgArray { ^str.standardizePath }
+
+	// match with String.asPath
+	asPath { ^this }
+
+	// tests
+	isAbsolute {
+		^Platform.case(\windows) {
+			"Windows".postln;
```
telephon, line 66: don't post "Windows", I'd say.

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+			str.drop(1).beginsWith(":\\") or: { "\\\\".beginsWith("\\\\") }
+		} {
+			// unix
+			str.at(0).isPathSeparator
+		}
+	}
+	isRelative { ^this.isAbsolute.not }
+	exists { ^File.exists(str) }
+	== { |path2| ^str == path2.str }
```
telephon, line 75: maybe `other`?

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+
+	//// we need trailing slash for these
+	isFolder { ^str.last == Platform.pathSeparator }
+	isDir { ^str.last == Platform.pathSeparator }
+	isFile { ^str.last != Platform.pathSeparator }
+
+	// replace PathName:colonIndices
+	separatorIndices {
+		var res = List[];
+		str.do { |ch, i| if (ch.isPathSeparator) { res = res.add(i) } };
```
telephon, line 85: characters usually are abbreviated as `char`.

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+	+/+ { | path |
+		var otherstr = path.respondsTo(\str).if({ path.str }, { path.asString });
+		^this.class.new(str +/+ otherstr)
+	}
+
+	// search
+	match { ^str.pathMatch }
+	pathMatch { ^str.pathMatch } // compatibility
+
+	entries { ^pathMatch(str +/+ "*").collect { |str| this.class.new(str) } }
```
telephon, line 127: perhaps make sure that it is not a file?

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+	files { ^this.entries.select({ | item |item.isFile }) }
+	folders { ^this.entries.select({ | item | item.isFolder }) }
+
+	deepFiles { ^this.entries.collect({ | item |
```
telephon: isn't this one:

```
deepFiles {
	var files;
	this.filesDo { | item | files = files.add(item) };
	^files
}
```

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+	fullPath { ^str } // compat
+
+	// from File
+	type { File.type(str) }
+
+	// conversions
+	absolutePath { ^this.class.new(str.absolutePath) }
+	relativeTo { |path2| ^str.asRelativePath(path2) }
+	withName { |name| ^Path(str.dirname +/+ name) }
+
```
telephon, line 54: `resolveRelative` would be good to have.

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+	// replace PathName:lastColonIndex
+	lastSeparatorIndex { ^this.separatorIndices.last ? -1 }
+
+	nextName { |indexOfNumber, numExtToRemove=2|
+		var parent = str.dirname, basefile = str.basename;
+		var filename, extensions, nameParts, digitIndices;
+		var partIndex, endNumber = 0, nextFileName, nextFullName;
+		filename = basefile;
+		numExtToRemove.do { filename = filename.splitext[0] };
```
telephon, line 97: Should this be `fileNameWithoutDoubleExtension` or maybe `fileNameWithoutExtension(2)`?

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+		extensions = basefile.drop(filename.size);
+
+		nameParts = filename.separate { |a, b| a.isDecDigit != b.isDecDigit };
+		digitIndices = nameParts.selectIndices { |part| part[0].isDecDigit };
+
```
telephon, line 102: Have you tested with a filename like "abc123cde456.wav"?

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+		};
+		nextFullName = parent +/+ nextFileName ++ extensions;
+		if (nextFullName.beginsWith("./")) { nextFullName = nextFullName.drop(2) };
+
+		^nextFullName
+	}
+
+	/* concatenation */
+	+/+ { | path |
+		var otherstr = path.respondsTo(\str).if({ path.str }, { path.asString });
```
telephon: better move if to front: `if(path.respondsTo(\str)) { ... }`

telephon: but don't we have now 
```
path  = path.asPath;
^this.class.new(str +/+ path.asPathString)
```

SCClassLibrary/Common/Files/Path.sc
```diff
@@ -0,0 +1,176 @@
...
+		this.folders.do { | path |
+			Path(path).filesDo(func)
+		};
+	}
+
+	// text output
+	streamTree { | str, tabs = 0 |
+		str << this << Char.nl;
+		this.files.do({ | item |
+			tabs.do({ str << Char.tab });
```
telephon: leave out `({` in favour of `{`

telephon (changes requested):

HelpSource/Classes/Path.schelp
tedmoore: I would prefer `""` so that the method always returns the same type. `nil` might seem like an error.

HelpSource/Classes/Path.schelp
tedmoore: I agree we should choose one. I prefer directory (even though casually I often say "folder").

HelpSource/Classes/Path.schelp
adcxyz: agree with Ted

telephon: Wouldn't you want to write something like

```
ext = file.extension ? ".wav"
```

rather than

```
ext = file.extension;
if(ext == "") { ext = ".wav" };
```

JordanHendersonMusic: I think nil should be used. An empty string is still a valid instance of the type, nil clearly communicates the absence of something. Nil is also far more performant.

tedmoore: > ext = file.extension ? ".wav"

it is nice. i also don't mind the second one. things like this often seem like "nice tricks" for someone who knows programming and sc well: returning nil here is convenient given other useful syntax patterns. 

when i teach sc, things like this seem surprising and clever (in a bad way) to students, violating the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), making it harder to learn because it feels like sc has a lot of clever idiosyncrasies one has to memorize (which of course it does).

ultimately i'm not worried about which is chosen. just thought it would be interesting to share my thinking behind it

HelpSource/Classes/Path.schelp
adcxyz: I added name as a suggestion because python Path has it, 
and because it is short. Fine to remove it if no one likes it.

HelpSource/Classes/Path.schelp
adcxyz: The earlier implementation of PathName:extension returned "", so I kept that for backwards compatibility.

HelpSource/Classes/Path.schelp
adcxyz: for the documentation, I can change these to prefer directory. 
classlib has 58 method names with dir/directory, and 15 with folder. 
I think we need to keep the PathName interface for backwards compatiblity, 
which includes e.g. folderName.

HelpSource/Classes/Path.schelp
adcxyz: see below

HelpSource/Classes/Path.schelp
adcxyz: here, none. 
the old usage for folderName was analog to fileName, i.e. not its full path. 
the common usage of dirname is actually a bit misleading, since that is the full path ...

SCClassLibrary/Common/Files/Path.sc
adcxyz: ok, will add that.

SCClassLibrary/Common/Files/Path.sc
adcxyz: fileNameWithoutDoubleExtension existed, so we need to keep it. 
I can redirect it to fileNameWithoutExtension if there is agreement this is better.

HelpSource/Classes/Path.schelp
telephon: ok, so we should keep it in this case.

HelpSource/Classes/Path.schelp
adcxyz: no one spoke for it, so I remove it.

SCClassLibrary/Common/Files/Path.sc
adcxyz: yes, works fine: 
```
Path("abc123cde456.wav").nextName;
-> abc123cde457.wav
```

SCClassLibrary/Common/Files/Path.sc
adcxyz: done, also in entries.

SCClassLibrary/Common/Files/Path.sc
adcxyz: no other opinions, so it stays as is for compatibility.

SCClassLibrary/Common/Files/PathName.sc
adcxyz: thanks, done!

HelpSource/Classes/Path.schelp
adcxyz: Would it make sense to add an isDir method in parallel to isFolder?

SCClassLibrary/Common/Files/Path.sc
adcxyz: actually a bit tricky: this simple approach
```
resolveRelative { ^this.class.new(str.resolveRelative) }

```
resolves to the path of the class method def, not the file in question. 
Copying the logic over from string seems bad, so how best to do it?

SCClassLibrary/Common/Files/Path.sc
adcxyz: yes, thanks, even simpler: 
```
	/* concatenation */
	+/+ { | path |
		^this.class.new(str +/+ path.asPathString)
	}
	// test 
	Path("cba") +/+ Path("cba");
	Path("abc") +/+ "cba";
// both do 
-> Path("abc/def")	
```

HelpSource/Classes/Path.schelp
jamshark70: > Would it make sense to add an isDir method in parallel to isFolder?

Not sure about that (wouldn't it be "just one more way"?) -- but -- currently isFolder's implementation assumes it's a folder based on pathMatch. Now we have `File.type` which would seem more direct to me -- revise this?

```
	isFolder {
		^File.type(this.fullPath) == \directory
	}
```

adcxyz:
Path and PathName tests fail on Windows, likely due to pathSeparator assumptions. 
will investigate.

Issue #2314, opened 2016-08-29
setter for Document text adds instead of setting it, getter returns nothing


telephon:
via Nathan Ho (thanks!):

The default range arguments for Document:string and Document:string_ are a little silly:

```
d = Document.new;
d.string = "Hello, world!"; // contents of document are "Hello, world!"
d.string = "Hello, world!"; // contents of document are "Hello, world!ello, world!"
d.string; // empty string
```

I spent a while hunting down a bug because of this. Could we have rangestart: 0 and rangesize: -1 instead so one doesn't have to keep typing myDocument.string(0, -1)?

(also the string getter should be correct)

snappizz:
Note that this would be a breaking change.

brianlheim:
#2718 will (hopefully) address this

PR #6756, opened 2025-04-21
Document.string_ : fix default args

adcxyz:
## Purpose

This PR fixes a bug in Document.string_, 
where default arguments for start and range are wrong.

```
// Demonstrate current bug:
d = Document("test-string", "string")
// ask its string:
d.string // correct: -> string 

// now set it to something else
d.string = "strong"

// now the open document shows "strongtring", which is wrong;
// and it thinks it still has "string", which is also wrong.
d.string 
// by comparison, setting via .text_ works as expected:
d.text = "strong"
d.text.cs // correct "strong"
```

This PR sets the default so ```d.string_ = "something" ``` works correctly. 

## Types of changes
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -810,7 +810,7 @@ Document {
 		^this.rangeText(rangestart, rangesize);
 	}
 
-	string_ { | string, rangestart = -1, rangesize = 1 |
+	string_ { | string, rangestart = 0, rangesize = -1 |
 		this.prSetText(string, nil, rangestart, rangesize);
 	}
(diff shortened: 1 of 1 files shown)


dyfer:
FYI I edited the opening comment so that the text after the code is formatted properly

dyfer (approved):
Works, thanks!

BTW test failures don't seem to be related to this PR. Feel free to merge.

dyfer:
Note: this seems to fix #2314

dyfer:
I added API change label based on the discussion in https://github.com/supercollider/supercollider/issues/2314

Issue #4368, opened 2019-03-26
Can't boot properly on Windows 10


ZeCraum:
## Environment

* SuperCollider version: 3.10.2
* Operating system: Windows 10 64bits

When I try to boot it, the server status goes yellow and freeze with the post window at:
`Booting server 'localhost' on address 127.0.0.1:57110.`

To unfreeze it I kill all the servers and then got this message:

```
ÊXITO: o processo "scsynth.exe" com PID 5776 foi finalizado.
Server 'localhost' exited with exit code 1.
```
Translating: Success: the process "scsynth.exe" with PID5776 was finished. 

I need to boot, kill, boot, kill, boot, kill... many times until a moment that it finally properly boots!

After booting properly it display the last part of this message many times:

```
Booting server 'localhost' on address 127.0.0.1:57110.

Device options:
  - MME : Mapeador de som da Microsoft - Input   (device #0 with 2 ins 0 outs)
  - MME : Microphone (High Definition Aud   (device #1 with 2 ins 0 outs)
  - MME : Microphone (High Definition Aud   (device #2 with 2 ins 0 outs)
  - MME : Mapeador de som da Microsoft - Output   (device #3 with 0 ins 2 outs)
  - MME : Speakers (High Definition Audio   (device #4 with 0 ins 2 outs)
  - Windows DirectSound : Driver de captura de som primário   (device #5 with 2 ins 0 outs)
  - Windows DirectSound : Microphone (High Definition Audio Device)   (device #6 with 2 ins 0 outs)
  - Windows DirectSound : Microphone (High Definition Audio Device)   (device #7 with 2 ins 0 outs)
  - Windows DirectSound : Driver de som primário   (device #8 with 0 ins 2 outs)
  - Windows DirectSound : Speakers (High Definition Audio Device)   (device #9 with 0 ins 2 outs)
  - Windows WASAPI : Speakers (High Definition Audio Device)   (device #10 with 0 ins 2 outs)
  - Windows WASAPI : Microphone (High Definition Audio Device)   (device #11 with 2 ins 0 outs)
  - Windows WASAPI : Microphone (High Definition Audio Device)   (device #12 with 2 ins 0 outs)
  - Windows WDM-KS : Microphone 1 (HD Audio Muxed capture)   (device #13 with 2 ins 0 outs)
  - Windows WDM-KS : Microphone 2 (HD Audio Muxed capture)   (device #14 with 2 ins 0 outs)
[...]

PR #6758, opened 2025-04-22
scsynth/supernova: refactor SC_ServerBootDelayWarning

Spacechild1:
## Purpose and Motivation

Currently, the Server boot delay timer thread is not joined when the main function returns early due to an error condition. This causes the program to terminate instead of exiting cleanly.

I have turned the free functions into a RAII class to make sure that the timer thread will be joined automatically when the main function returns.

## Types of changes

- Bug fix
- Refactoring

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/common/SC_ServerBootDelayWarning.cpp
+++ b/common/SC_ServerBootDelayWarning.cpp
@@ -1,44 +1,39 @@
 #include "SC_ServerBootDelayWarning.h"
 
 #ifdef _WIN32
+#    include <cassert>
 #    include <iostream>
-#    include <thread>
-#    include <mutex>
 #    include <chrono>
-#    include <condition_variable>
 
-static bool g_isServerBooted = false;
-static std::mutex g_serverBootWarningMutex;
-static std::condition_variable g_serverBootWarningConditionVariable;
-static std::thread g_bootWarningThread;
-
-static void displayWarningIfServerHasNotBooted() {
-    std::unique_lock<std::mutex> lock(g_serverBootWarningMutex);
-
-    if (!g_serverBootWarningConditionVariable.wait_for(lock, std::chrono::seconds(10),
-                                                       [] { return g_isServerBooted; })) {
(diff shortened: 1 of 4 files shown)


dyfer:
@Spacechild1 it seems this does not build on Linux and macOS

Spacechild1:
D'oh! Fixed!

dyfer (approved):
The code looks fine to me, but I have limited understanding of cpp.

I was able to run the Windows build in a VM but I couldn't trigger the boot delay on that system so I couldn't see the posting myself. 

I think this could go in as is, unless there's somebody else who could review this.

prko:
This PR seems to resolve the delay of server boot.
I do not see the following message (reported in #6774):
>On some Windows-based machines, Windows Defender sometimes delays server boot by one minute.
You can add scsynth.exe and/or supernova.exe processes to Windows Defender exclusion list to avoid this delay. It's safe.

prko:
>>I do not see the following message (reported in https://github.com/supercollider/supercollider/issues/6774):

>Do you expect to see that message?

I kept encountering that message last week, and as a result, the server failed to boot, preventing me from using SuperCollider on Windows. However, I came across this PR and decided to try its artifact, which allowed me to use it once again. While this PR does not fully resolve issue #6774, it does appear to address the server boot problem in some manner.  

Many thanks!

Spacechild1:
> Many thanks!

Your're welcome, but I'm pretty sure that this PR did not actually solve your problem because it is functionally equivalent to what was there before :)

dyfer:
IIRC discussion in #4368, the circumstances when the server boot is delayed are not fully deterministic, that's why @prko  might've observed a different behavior with this PR, while the behavior was not changed by the PR itself.

@Spacechild1 I think this can indeed be merged as you are on the "server team" in the council structure and the change is relatively straightforward.

PR #6760, opened 2025-04-22
plugins: ToggleFF: support input rate != unit rate + fix init sample

### LINKED ISSUE #5917: ToggleFF.ar(0) returns erratic output

elgiano:
## Purpose and Motivation
Superseded #6051
Fixes #5917

When running at audio rate, ToggleFF doesn't handle well control-rate or scalar inputs: it actually tries to read them as if they were audio-rate, i.e. reading `bufferSize()` samples from a 1-sample buffer, effectively reading data from adjacent wires. This creates glitches, for example when running:
```supercollider
ToggleFF.ar(0) // toggles all the time
```
Instead of restricting input rates or converting input UGens on sclang - as discussed in #6051 - this PR adds variations to ToggleFF calc function, so that inputs are read properly. This way existing code continues to work.

Breaking changes:

- `ToggleFF.ar(0)`: will now work correctly, instead of toggling all the time
- `ToggleFF.kr(Impulse.ar(1))`: will now work correctly, toggling every time a trigger is received, even  in-between a control period (which might have been missed before this PR)

Additionally, this PR changes the "presample", or "initialization sample". It was hardcoded to 0, and I've changed to reflect y(0), i.e. if there is a trig in the first input sample, init sample outputs 1, otherwise 0.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing

Docs didn't mention restrictions to ToggleFF input's rate, so we don't need to add docs.


server/plugins/TriggerUGens.cpp
```diff
@@ -893,14 +892,56 @@ void SetResetFF_next_k(SetResetFF* unit, int inNumSamples) {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////
```
telephon, line 894: It is a bit a matter of taste, but I would find it more readable if we just had three functions without a template, and choose the right one from Ctor?

elgiano: For me the main advantage is that it makes it more clear what code is in common and what differs, and it also keeps the different parts closer together. For UGens where the diverging code is small, I also find it less error prone.

On a second thought, it would be probably better to have this discussion in a separate place and focus on ToggleFF here. The template is also not so necessary for a UGen so small and without many alternative calc functions, so I reverted to the more traditional style.

dyfer:
Pinging @mtmccrea for a possible review as this touches the init sample behavior :)

mtmccrea:
Did some quick tests running different ar/kr combinations and it seems to work as advertised. I'll try to have a look in more detail in the next day or two.

@elgiano, since you may be workin on more of these trigger init sample fixes, this may be useful for your testing... Here's the pattern that I've used to actually see what init sample is produced (since it's not actually output):

```c++
void ToggleFF_Ctor(ToggleFF* unit) {
    if (unit->mCalcRate == calc_FullRate) {
        if (INRATE(0) == calc_FullRate) {
            SETCALC(ToggleFF_next);
        } else {
            SETCALC(ToggleFF_next_ak);
        }
    } else {
        if (INRATE(0) == calc_FullRate) {
            SETCALC(ToggleFF_next_ka);
        } else {
            SETCALC(ToggleFF_next);
        }
    }

    unit->m_prevtrig = 0.f;
    unit->mLevel = 0.f;

    Print("Init sample:\n"); // << print before init sample is produced
    ToggleFF_next(unit, 1);
    Print("First sample:\n"); // << then directly following in advance of first ouput calculation
    // no need to reset internal state
    // if trig is detected in Ctor, UGen doesn't need to toggle again in _next
}

void ToggleFF_next(ToggleFF* unit, int inNumSamples) {
    float* out = ZOUT(0);
    float* trig = ZIN(0);
    float level = unit->mLevel;

    LOOP1(inNumSamples, float curtrig = ZXP(trig); if (prevtrig <= 0.f && curtrig > 0.f) level = 1.f - level;
          prevtrig = curtrig; ZXP(out) = level;
          Print("\tout level: %.3f\n", level); // << print actual output here
          );

    unit->m_prevtrig = prevtrig;
    unit->mLevel = level;
}
```

server/plugins/TriggerUGens.cpp
```diff
@@ -895,23 +897,67 @@ void SetResetFF_next_k(SetResetFF* unit, int inNumSamples) {
...
-
 void ToggleFF_next(ToggleFF* unit, int inNumSamples) {
     float* out = ZOUT(0);
     float* trig = ZIN(0);
     float prevtrig = unit->m_prevtrig;
     float level = unit->mLevel;
 
     LOOP1(inNumSamples, float curtrig = ZXP(trig); if (prevtrig <= 0.f && curtrig > 0.f) level = 1.f - level;
-          ZXP(out) = level; prevtrig = curtrig;);
+          prevtrig = curtrig; ZXP(out) = level;);
```
mtmccrea, line 929: Any particular reason to rearrange these? In general we'd like to keep non-functional changes to a minimum. 

And I think generally, setting state variables after the output is a common pattern, as it's prepping that variable for the next round of calculations.

server/plugins/TriggerUGens.cpp
```diff
@@ -895,23 +897,67 @@ void SetResetFF_next_k(SetResetFF* unit, int inNumSamples) {
...
+            SETCALC(ToggleFF_next);
+        }
+    }
 
     unit->m_prevtrig = 0.f;
     unit->mLevel = 0.f;
 
-    ZOUT0(0) = 0.f;
+    ToggleFF_next(unit, 1);
+    // no need to reset internal state
```
mtmccrea, line 918: I see why it's not necessary to reset the state because of how the toggling behaves. But, I think it would still be best to stick with the pattern of resetting the state after the init sample is calculated. Strictly speaking, the _state_ is inaccurate at the time the first sample is calculated in this case, i.e. `m_prevtrig` remains `1.0`, when in principle it should be `0.0` (same with level, which "should" be `0.0`) at the start of the first calculation block.

It's really a matter of consistency and to reinforce the proper pattern. Open to arguments otherwise though!

elgiano:
Thanks for the review, I'll address all your points soon. Just a quick note for now: it came up recently in a pull request about IEnvGen, that we can actually use IEnvGen to test other units init sample, because it reads its envelope only in the Ctor and never updates it again.

server/plugins/TriggerUGens.cpp
elgiano: I'm a bit conflicted about this:
- repeating calculation is not a big deal, I think it would be worth it to reinforce the pattern. However it's also not necessary and not repeating it is one less change to how the UGen behaves (at least in theory, probably negligible in practice)
- we recently had another case of skipping reset in #6187 with IEnvGen, where in the end we decided to put a comment... and in TriggerUGen there are many more cases, where y(0) is computed at Ctor and "cached" for the first _next call. 

It would be nice to decide which way we wanna go. Maybe IEnvGen stays as an exception, and all TriggerUGens should reset properly even if they can easily "cache" y(0)?

mtmccrea: Given how pervasive this initialization sample problem is, my vote is to follow the pattern. 

I have a sense that the problem took root because UGens were added that basically copied the Ctor of others, which weren't explicit about handling initial state.

I also think that having state cached between init and the first calc block is harder to reason about and obscures the important distinction between and init sample and what is output at runtime.

elgiano: Ok, since I'm working on more plugins from TriggerUGens, I think it will be better to look at them all together. I'll remove the change to initialization from this PR and open another one with all of them

server/plugins/TriggerUGens.cpp
elgiano: thanks! It was a leftover from my previous implementation

server/plugins/TriggerUGens.cpp
```diff
@@ -912,6 +925,37 @@ void ToggleFF_next(ToggleFF* unit, int inNumSamples) {
...
+}
+
+void ToggleFF_next_ka(ToggleFF* unit, int inNumSamples) {
+    float* out = ZOUT(0);
+    float* trig = ZIN(0);
+    float prevtrig = unit->m_prevtrig;
+    float level = unit->mLevel;
+
+    LOOP1(FULLBUFLENGTH, float curtrig = ZXP(trig); if (prevtrig <= 0.f && curtrig > 0.f) level = 1.f - level;
+          prevtrig = curtrig);
```
mtmccrea, line 956: There's a question of how the trigger should be interpreted at audio rate. As it's currently implemented, the `level` will toggle multiple times within the control block, but the output will just be whatever happens to be the last level.

Another interpretation (perhaps more appropriate for ToggleFF running at kr?) is that if there's a trigger at _any_ point within the control block, the level flips from whatever it was in the _previous block_. This would behave more like how `T2K` works.

server/plugins/TriggerUGens.cpp
```diff
@@ -912,6 +925,37 @@ void ToggleFF_next(ToggleFF* unit, int inNumSamples) {
...
+    unit->mLevel = level;
+}
+
+void ToggleFF_next_ak(ToggleFF* unit, int inNumSamples) {
+    float* out = ZOUT(0);
+    float* trig = ZIN(0);
+    float prevtrig = unit->m_prevtrig;
+    float level = unit->mLevel;
+
+    float curtrig = ZXP(trig);
```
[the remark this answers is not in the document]
elgiano: Thanks! other leftover from template implementation

server/plugins/TriggerUGens.cpp
elgiano: Ok, this is actually new behavior. I originally thought the UGen could be sensitive to every trigger that happens in a control block... but since T2K works like you said, now I'm leaning towards your second interpretation.

I'll look for other cases in other UGens and get back :)

elgiano: Also, would it be fine that `ToggleFF.kr(DC.ar(1))` would toggle every control block? Or should it toggle only once and never again?
In other words, are we setting prevtrig :
1. to 0 at the beginning of every block (like T2K, would toggle every control block)
2. to 1 if the control block had a trigger (would toggle every two control blocks, can't toggle two blocks in a row)
3. to whatever the input value was at the end of the block (would toggle once and never again)

I think it should be either 2 or 3, and that I would prefer 3, but what do you think?

jamshark70: > "3. to whatever the input value was at the end of the block (would toggle once and never again)"

This is the canonical meaning of a trigger in scsynth -- so IMO this is the only valid way to do it. By definition, a stream of 1, 1, 1, 1, 1, 1 can represent a single trigger at the beginning, and no triggers in the middle. The input *must* reset to 0 before another trigger can happen.

It's never been possible to get ToggleFF to output 1, 0, 1, 0, 1, 0.

If ToggleFF receives 1, 0, 1, 0, 1, 0... (fastest possible triggers -- note that it's never been possible in any trigger UGen to have triggers faster than this), then it should output 1, 1, 0, 0, 1, 1, 0, 0...

```
a = {
	var fastestPossibleTrigger = Impulse.ar(SampleRate.ir * 0.5);
	[fastestPossibleTrigger, ToggleFF.ar(fastestPossibleTrigger)]
}.plot(duration: 0.0025);
```

If this PR is about rate-support, I think it would be questionable best to slip an entirely different sort of breaking change in through the back door -- this PR should maintain compatibility while only adding the mixed-rate option.

elgiano: I see your point, but do you mean that by analogy it should apply also to ToggleFF.kr? It shouldn't be able toggle once per control block? In this case we should go for 2. 
If we go for 3 we could still have 101010 (where each sample is a control period).

Also, with T2K, we would have a trigger only if the input was non-positive for a full control block. Should this be the case here too, or is it sufficient that the input is non-positive for a single sample?

jamshark70: > I see your point, but do you mean that by analogy it should apply also to ToggleFF.kr (with an audio-rate input)? It shouldn't be able toggle once per control block?

A trigger is defined as a transition from non-positive to positive (and, at initialization, x(-1) is assumed to be 0, so any x(0) > 0 represents a trigger).

Therefore you cannot retrigger without a preceding nonpositive value. I don't think there's any exception to this rule anywhere in scsynth.

`DC.ar(1)` in a trigger input always, and only, represents a single trigger at the beginning, and no subsequent triggers.

Now, you do raise an interesting question -- in ToggleFF.kr, with an audio rate input, what if the input drops to zero in the middle of the block? Should we trigger on the next kr block?

I think that would be justifiable -- *but* that input is not DC.ar(1). I'd be OK with "set prevtrig to 0 if any ar input sample in this control block is <= 0."

But [the specific question](https://github.com/supercollider/supercollider/pull/6760#discussion_r2065044699) was "Also, would it be fine that `ToggleFF.kr(DC.ar(1))` would toggle every control block?" -- and that must be no. DC.ar(1) cannot retrigger, by definition.

elgiano: I guess we are left with two alternatives:

1. ToggleFF.kr reads only the first sample of an audio rate input. This will miss triggers and resets happening in the middle of a control period, making it kind of non-functional, but consistent with other kr trig ugens (eg. SetResetFF). We could also post a warning in sclang 3.14 for ToggleFF.kr with audio-rate input, and deprecate it in 3.15 by adding same-rate checkInput rules like @cdbzb proposed originally.

2. ToggleFF.kr reads all the audio input's buffer and toggles only once per block if there is any transition from non-positive to positive. Prevtrig is set to the input's last sample, so it's also sensitive to transitions across a control period.

In both cases DC.ar(1) never retriggers. I perceive 2 as more "accurate" (debatable), but since I don't know of any other UGen doing this, 1 would be more consistent with the rest of scsynth.

elgiano: Note: I'm reading through all TrigUGens, and it looks like ka/ak behavior is a common problem. Most UGens seem to overflow kr buffers in either the ka or ak case. Some, but not all have checkInput guards on sclang side, but I don't think this is enough.
So far it seems like the most used convention for the ka case is to read only the input's first sample. 

I'll make a more detailed report

jamshark70: > For the sake of this PR, I think I would make ToggleFF align with the general ka behavior (read only first sample), and still pursue the fix for ak as a precedent.

Related to https://scsynth.org/t/how-we-frame-something-as-architecture-design-over-engineering-or-technical-debt/11611 --

The meaning of an audio rate trigger in a control rate UGen has never been discussed (at least, I don't remember it being discussed).

So this would be a chance to draw up an actual spec, before doing anything else.

Question 1: What concrete advantage do we get from `AUGen.kr(Impulse.ar())`?

If there is none, then (IMO) the spec should be not to support it, and require trigger-processing UGens to enforce checkSameRateXXX. This is option 1 (read only first sample). In this case, adding C++ to TriggerUGens.cpp to handle the ka case increases complexity of the code base, for no increase in functionality (because the behavior would be no different from `AUGen.kr(A2K.kr(Impulse.ar()))`) -- a net loss.

But, one such concrete advantage has been proposed -- if a kr UGen receives a faster-than-kr input, then it becomes possible to trigger faster than half-kr (which is currently not possible).

So for me, C++ changes would be justified by adding this new feature but not justified without the new feature.

Question 2: What's the meaning of faster-than-kr triggers at kr?

- Case 1: ar block begins positive and ends with 0 -- easy, if prevtrig <= 0.f then trigger now, in this block (and set prevtrig <= 0 for the next kr block).
- Case 2: ar block begins positive and stays positive throughout -- easy, if prevtrig <= 0.f then trigger now, in this block (and set prevtrig > 0 for the next kr block).
  - Case 2b: ar block begins nonpositive and stays nonpositive throughout -- easy, no trigger, set prevtrig <= 0 for the next block.
- Case 3: ar block begins with 0 and ends positive -- should trigger in this block, or the next? I tend to think this block, but I could see the other point of view.
- Case 4: ar block begins and ends positive, but drops to <= 0 somewhere in the middle.
  - A/ If prevtrig <= 0 we must trigger now (because the beginning of the block represents a transition from non-positive to positive).
  - B/ If prevtrig > 0, then the <=0 to >0 transition exists in this block, but not at the beginning. I can see two ways to handle that.
    - B1/ Trigger now (in this block) and set prevtrig to the ar input's last sample.
    - B2/ Consider this kr block to be nonpositive, set prevtrig = 0, and wait for the next block.

I tend to prefer case 3 to trigger now and case 4, use B1.

Pseudocode:

```
// if `in` is a pointer to the ar block
float min = 1e32, max = -1e32;

for(int i = 0; i < numSamples; i++, in++) {
	if(*in < min) min = *in;
	if(*in > max) max = *in;
};

if((prevtrig <= 0 || min <= 0) && (max > 0)) {
	do trigger stuff;
};

prevtrig = *(in-1);
```

That is, if the audio block contains any positive sample and if it also contains any nonpositive sample (or the last block ended nonpositive), then trigger.

elgiano:
So far this PR spawned the necessity of more complex discussions, about all trigger UGens:

1. **Trigger UGens init sample**
fixes and enforcing conventions. We are already working on this along the lines of @mtmccrea's work.

2. **Trigger UGens "ak" case**
When running at audio-rate with control-rate inputs, TriggerUGens unexpectedly read adjacent input wires. While we can block this behavior with `checkInputs()`, that's a sclang-only solution that leaves a "bug" in scsynth. Moreover, not all trig UGens implement `checkInputs` at the moment.

3. **Redesigning trigger undersampling ("ka" case)**
what should TrigUGen.kr(Input.ar()) do? So far the convention is to read only the first sample of input (since the UGen is working at control rate), but also to forbid this rate mismatch via checkInputs in many cases.

Since the discussion went already long and multi-threaded on this PR, I think a good way forward would be to open another one where we only attempt to handle the "ak" case without overflowing for at least all UGens that don't have checkInputs in place, and preferably for all of them. 

Let me know if I should proceed or we should still pursue this PR for ToggleFF only.

server/plugins/TriggerUGens.cpp
elgiano: Great, thanks for the interesting reflections.
I think your redesign point can be resumed by: a control-rate UGen triggers if there is any transition from nonpositive to positive in this control block. And I agree this would be a meaningful definition of undersampling triggers. But we need to spawn a separate discussion, because we'll need to change all trigger UGens.

Also, if two adjacent control blocks contain two triggers, one per block, their KR undersampling wouldn't be a trigger (since kr output wouldn't be non-positive in between). Just pointing this out here, but again we need a broader discussion.

mtmccrea: Thanks @jamshark70 for your nice description of pros/cons of adding the feature and the various ways of interpreting trigger downsampling.

>I tend to prefer case 3 to trigger now and case 4, use B1 -- simpler principle, if there's a nonpositive-to-positive transition at the start or anywhere within the block, do the trigger now.

I agree with this approach (ka case). To add to @elgiano's description, I think a complete description (which may be useful for documentation) is:

```
If there is at least one valid trigger at any time within the block, the kr UGen triggers. 
The trigger state at the end of the block is determined by the last value of the audio-rate trigger input.
```

"valid trigger" is already well defined in SC.

As James noted, this approach does make it possible for the output of `ToggleFF.kr` to be `1,0,1,0,...`, which I agree justifies supporting audio-rate triggering at control rate.

A more conceptual justification is that it's a net benefit to support patching/interconnecting signals regardless of what rate they run at. E.g. If I want my ar noise signal to be used as a trigger in a subcomponent running at kr, it should "just work" without the addition of another UGen. 

SC should have a spec for this!

mtmccrea:
Since there's not yet a new Issue, I'll just say here...

> 2\. **Trigger UGens "ak" case**
>     When running at audio-rate with control-rate inputs, TriggerUGens unexpectedly read adjacent input wires. While we can block this behavior with checkInputs(), that's a sclang-only solution that leaves a "bug" in scsynth. 

This is a bug in the UGen and should be fixed there. I think limiting input rates should be used only in stringent cases where the behavior is really unsupported, by nature of the DSP algorithm, or otherwise undefined.

mtmccrea:
> Let me know if I should proceed or we should still pursue this PR for ToggleFF only.

A more granular approach would be to just fix the init sample behavior with this PR. That doesn't need to wait on working out I/O rate issue.

dyfer:
Please note that the current version of this PR doesn't build on Windows: https://github.com/supercollider/supercollider/actions/runs/14754617449/job/41419803054?pr=6760#step:14:1013

dyfer:
@elgiano @mtmccrea 
We are trying to release 3.14 in the next few days, would it be possible to wrap up this PR?

server/plugins/TriggerUGens.cpp
```diff
@@ -895,15 +896,18 @@ void SetResetFF_next_k(SetResetFF* unit, int inNumSamples) {
...
+    if (unit->mCalcRate == calc_FullRate and INRATE(0) != calc_FullRate) {
+        SETCALC(ToggleFF_next_ak);
+    } else {
+        SETCALC(ToggleFF_next);
+    }
 
     unit->m_prevtrig = 0.f;
     unit->mLevel = 0.f;
 
     ZOUT0(0) = 0.f;
```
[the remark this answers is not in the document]
elgiano: Init sample fixed in https://github.com/supercollider/supercollider/pull/6816

elgiano:
Now this PR only addresses the "ak" case, since init sample was fixed in https://github.com/supercollider/supercollider/pull/6816

I think it's reasonable to merge it, even though more UGens will need this fix. I've already opened an issue about it (https://github.com/supercollider/supercollider/issues/6803) and I'll try to summarize the discussion we had at https://github.com/supercollider/supercollider/pull/6760#discussion_r2064328572 there

server/plugins/TriggerUGens.cpp
```diff
@@ -899,7 +900,11 @@ void SetResetFF_next_k(SetResetFF* unit, int inNumSamples) {
 
 
 void ToggleFF_Ctor(ToggleFF* unit) {
-    SETCALC(ToggleFF_next);
+    if (unit->mCalcRate == calc_FullRate and INRATE(0) != calc_FullRate) {
```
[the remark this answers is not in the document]
mtmccrea: Sorry I missed the tests, was busy doing my own 😬

dyfer:
@capital-G can we merge this?

capital-G (approved):

Issue #6136, opened 2023-11-14
Help Browser navigation shortcuts only work on MacOS


prko:
## Environment

* SuperCollider version: 3.13.0
* Operating system: Windows, Ubuntu

## Steps to reproduce

1. Navigate the help document by pressing the "Browse" button.
2. Navigate the help document by pressing the "Search" button.
3. Go to the previous page by pressing the "Previous Page" button on the top right of the screen.
4. Press "alt + left arrow" to go to the previous page.
5. Press "alt + right arrow" to go to the previous page.

### On mac, the shortcuts do not appear in the pop-up menu, but work as follows:
- Pressing "cmd + [" works to go to the previous page; 
- pressing "cmd + ]" works to go to the nexe page.

macos: 
<img width="303" alt="Screenshot 2023-11-15 at 2 17 30" src="https://github.com/supercollider/supercollider/assets/416281/5a8ca1d4-72d1-432d-a2d4-3e8c7c173162">

### On Windows and on Ubuntu, the shortcuts that appear in the popup menu, do not work.

ubuntu: 
<img width="398" alt="Screenshot 2023-11-15 at 2 18 04" src="https://github.com/supercollider/supercollider/assets/416281/6a180c4a-c0c5-4136-8038-bf85277ceedb">

windows:
<img width="373" alt="Screenshot 2023-11-15 at 2 19 26" src="https://github.com/supercollider/supercollider/assets/416281/4326e529-3aec-422f-bb33-d6568185956a">

elgiano:
I'm on Arch Linux, without desktop environment. Shortcut displayed are the same as in Windows, and they all work, except "Zoom In" and "Reload". It seems it could be a GNOME/Qt issue?

Probably related: #6315

prko:
I re-tested it on Fedora 40 (arm64, x11 display server), Ubuntu 22.04 (arm64, x11 display server) and Windows 11: these shortcuts still do not work.
Also, `CTRL + R` does not work in the help browser.

On Windows, I could not test this because the help system does not currently work. It only displays the help home page and does not produce any other pages.

Issue #5672, opened 2021-12-30
Using CTRL+R leads to "QAction::event: Ambiguous shortcut overload: Ctrl+R" after building on Unbuntu.


simon-labomedia:
## Environment

* SuperCollider version: 3.12.1
* Operating system: Ubuntu 20.04 LTS
* Other details (Qt version, audio driver, etc.): QT5, built with QTWebEngine, cmake flag : -DNATIVE=ON, french keyboard layout.

## Steps to reproduce

Building latest release following the most basic process in the documentation.
Opening scide either from terminal or icon launcher.
Typing text in the IDE.
Selecting text.
Pressing CTRL+R.

## Expected vs. actual behavior

Expectation : opens the 'replace' UI in the bottom of the IDE.

Actual Behavior : nothing apparent, but terminal yields "QAction::event: Ambiguous shortcut overload: Ctrl+R".

## Additional notes

Was working fine with 3.10 ( from apt ).
Doesn't seem to be related to #5639 , as Emojis are not connected to CTRL+R.
CTRL+R not a custom shortcut on my OS.
Might be related to CTRL+R being a special command in stty ( rprnt ) ?
Possibly an issue with 'wayland' ?

simon-labomedia:
Just noticed CTRL+R actually refreshes the documentation panel. Is it linked with it being a browser, as CTRL+R have the same function in Firefox for example ?

PR #6762, opened 2025-04-23
scide: HelpBrowser fix ambiguous shortcuts

### LINKED ISSUE #6135: The class 'Download' does not work under Windows
## Environment

* Operating system: Windows 11 Pro

## Steps to reproduce

```supercollider
Download("https://theremin.music.uiowa.edu/sound%20files/MIS/Piano_Other/piano/Piano.mf.C4.aiff")
```

### expected  and actual behaviours on macOS and Ubuntu```
```
-> a Download
Download of https://theremin.music.uiowa.edu/sound%20files/MIS/Piano_Other/piano/Piano.mf.C4.aiff finished
```

### actual behaviours on Windows 11 Pro
```
-> a Download
```
The expected file is not downloaded.

### LINKED ISSUE #6315: CTRL + R does not invoke Find-Replace bar on the bottom of SC-IDE on Linux
## Environment

* SuperCollider version: 3.14.dev
* Operating system: Linux (tested on Ubuntu, Debian and Fedora)

## Steps to reproduce

Press CTRL + r

- Expected behaviour
  A bar with find-and-replace input fields should appear.
- Actual behaviour
   Nothing happens

elgiano:
## Purpose and Motivation
Fixes #6315 , #6136, #5672

Keyboard shortcuts for "Back", "Forward" and "Reload" actions can be ambiguous depending on OS and Desktop Environment.
This PR makes sure these actions are called when the Help Browser is in focus, resolving the ambiguity.
It also makes the "Reload" shortcut configurable in IDE settings (using the same shortcut as ide-reload-document).

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/help_browser.cpp
+++ b/editors/sc-ide/widgets/help_browser.cpp
@@ -154,14 +154,26 @@ void HelpBrowser::createActions() {
     // For the sake of display:
     mWebView->pageAction(QWebEnginePage::Copy)->setShortcut(QKeySequence::Copy);
     mWebView->pageAction(QWebEnginePage::Paste)->setShortcut(QKeySequence::Paste);
+
+    // proxy page actions to avoid shortcuts conflicts with main window
+    // note that we assign shortcuts here as they don't depend on IDE settings
+    auto proxyPageAction = [this](QAction* pageAction) {
+        // OverridingAction limits shortcut context to this widget
+        auto ovrAction = new OverridingAction(pageAction->icon(), pageAction->text(), this);
+        connect(ovrAction, SIGNAL(triggered()), pageAction, SLOT(trigger()));
+        // disable pageAction shortcut and assign it to ovrAction instead
+        ovrAction->setShortcut(pageAction->shortcut());
+        pageAction->setShortcut(QKeySequence());
+        ovrAction->addToWidget(this);
+        return ovrAction;
+    };
+    mActions[Back] = proxyPageAction(mWebView->pageAction(QWebEnginePage::Back));
+    mActions[Forward] = proxyPageAction(mWebView->pageAction(QWebEnginePage::Forward));
+    mActions[Reload] = proxyPageAction(mWebView->pageAction(QWebEnginePage::Reload));
 }
 
 void HelpBrowser::applySettings(Settings::Manager* settings) {
     settings->beginGroup("IDE/shortcuts");
-
-    mWebView->pageAction(QWebEnginePage::Back)->setShortcut(QKeySequence::Back);
-
-    mWebView->pageAction(QWebEnginePage::Forward)->setShortcut(QKeySequence::Forward);
 
     mActions[DocClose]->setShortcut(settings->shortcut("ide-document-close"));
 
@@ -388,9 +400,9 @@ void HelpBrowser::onContextMenuRequest(const QPoint& pos) {
         menu.addSeparator();
     }
 
-    menu.addAction(mWebView->pageAction(QWebEnginePage::Back));
-    menu.addAction(mWebView->pageAction(QWebEnginePage::Forward));
-    menu.addAction(mWebView->pageAction(QWebEnginePage::Reload));
+    menu.addAction(mActions[Back]);
+    menu.addAction(mActions[Forward]);
+    menu.addAction(mActions[Reload]);
 
 #    if (QT_VERSION < QT_VERSION_CHECK(6, 2, 0))
     if (contextData.selectedText().isEmpty())
@@ -403,9 +415,9 @@ void HelpBrowser::onContextMenuRequest(const QPoint& pos) {
 
     menu.addSeparator();
 
-    menu.addAction(mWebView->pageAction(QWebEnginePage::Back));
-    menu.addAction(mWebView->pageAction(QWebEnginePage::Forward));
-    menu.addAction(mWebView->pageAction(QWebEnginePage::Reload));
+    menu.addAction(mActions[Back]);
+    menu.addAction(mActions[Forward]);
+    menu.addAction(mActions[Reload]);
 
     menu.addSeparator();
 
@@ -483,9 +495,9 @@ HelpBrowserDocklet::HelpBrowserDocklet(QWidget* parent): Docklet(tr("Help browse
 
     toolBar()->addWidget(mHelpBrowser->loadProgressIndicator(), 1);
     toolBar()->addAction(mHelpBrowser->mActions[HelpBrowser::GoHome]);
-    toolBar()->addAction(mHelpBrowser->mWebView->pageAction(QWebEnginePage::Back));
-    toolBar()->addAction(mHelpBrowser->mWebView->pageAction(QWebEnginePage::Forward));
-    toolBar()->addAction(mHelpBrowser->mWebView->pageAction(QWebEnginePage::Reload));
+    toolBar()->addAction(mHelpBrowser->mActions[HelpBrowser::Back]);
+    toolBar()->addAction(mHelpBrowser->mActions[HelpBrowser::Forward]);
+    toolBar()->addAction(mHelpBrowser->mActions[HelpBrowser::Reload]);
     toolBar()->addWidget(mFindBox);
 
     connect(mFindBox, SIGNAL(query(QString, bool)), mHelpBrowser, SLOT(findText(QString, bool)));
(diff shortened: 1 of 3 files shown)


elgiano:
Thanks! I'll look into it, but could you test it on macos and see if perhaps this PR fixes it already?

prko:
I tested with the artifact of the following build: 
[CI / build-macos / macOS arm64 (pull_request)](https://github.com/supercollider/supercollider/actions/runs/14609190997/job/40983983233?pr=6762)

`shift + enter`,`cmd + enter` and `cmd + .` work.
However, `cmd + shift + p` does not work.

elgiano:
Ok, thanks! We'll need a separate PR then, I guess. Since you're already at it, could you try if shortcuts for 'back', 'forward' and 'reload' work on macos?

prko:
- On Windows 11
  - `shift + return` works  
  - 'ctrl + return` works  
  - `alt + left arrow` works  
  - `alt + right arrow` works
  - `ctrl + w` works
  - `ctrl + -` works
  - `ctrl + 0` works
  - `ctrl + +` and `ctrl + =` does not work
- On Ubuntu
  - Sorry that I do not know how to use the artifact. scdie does not start by double-clicking. Also, `SuperColliderIDE.desktop` does not work.

prko:
I tested on Ubuntu after building from source.
The result is the same as in Windows: 
`ctrl + +` and `ctrl + =` do not work.

prko:
On Ubuntu and Window,
- Pressing `ctrl+r` on the SC-IDE editor calls `find-replace` correctly.
- Pressing `ctrl+r` on the help browser calls also `find-replace` in SC-IDE editor.

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -154,14 +154,33 @@ void HelpBrowser::createActions() {
     // For the sake of display:
     mWebView->pageAction(QWebEnginePage::Copy)->setShortcut(QKeySequence::Copy);
     mWebView->pageAction(QWebEnginePage::Paste)->setShortcut(QKeySequence::Paste);
+
+    // proxy page actions to avoid shortcuts conflicts with main window
+    action = mWebView->pageAction(QWebEnginePage::Back);
+    mActions[Back] = ovrAction = new OverridingAction(action->icon(), action->text(), this);
+    connect(ovrAction, SIGNAL(triggered()), action, SLOT(trigger()));
+    action->setShortcut(QKeySequence());
```
capital-G, line 162: maybe add a comment that the actual shortcut will be set via the `applySettings` method?

capital-G (commented):
I think this can be merged if it fixes bugs on non-macOS platforms, yet I must admit that I haven't fully understood `OverridingAction`.

One thing though: The reloading shortcut on macOS (defaults to F5 - also doesn't work when customized) doesn't seem to work: Forwards/Backwards still works (TIL about those shortcuts^^). 
But clicking reload document in the file dialog doesn't reload the browser either.

I don't know how to proceed in this case with the PR? Do people need a refresh browser shortuct in macOS? Does the reload shortcut work on Windows and Linux? If so, why doesn't it work on macOS?

dyfer:
I can confirm that on macOS the refresh shortcut (cmd-r) is broken in this PR. 

I definitely have used the "refresh" shortcut in the help browser when editing help files...

elgiano:
With this PR, "browser reload" gets the same shortcut as "document reload". I thought that would be the default Refresh shortcut from Qt, but actually we are hardcoding it to F5 by default in main_window.cpp.

So I'm unintentionally changing the shortcut for browser refresh. I'll change that, to make "browser refresh" default to its Qt default, and perhaps adding a new setting item to customize it.

In the meanwhile, can you check wether Browser Refresh works on mac, with the right shortcut? Note that you have to focus the Help Browser to make it work.

dyfer:
> can you check wether Browser Refresh works on mac, with the right shortcut?

It doesn't not work on macOS, AFAICT. I made sure to the browser panel so that its in focus, also undocked it - it still doesn't reload on cmd-r.

elgiano:
@dyfer it should be F5 now, not cmd-r

dyfer:
> @dyfer it should be F5 now, not cmd-r

Ah got it. However, F5 doesn't work either, AFAICT.

elgiano:
Great! Thanks :D as a last thing, could you check if your document reload shortcut is actually F5? I'm afraid you need to dig in Preferences>Shortcuts and filter for "Reload" to find it

dyfer:
Additional info: if I go into the IDE preferences and change the reload shortcut to cmd-r, it worked in the help browser as well.

elgiano:
@dyfer ok this is a bit weird indeed...
I made some changes to use the default shortcut from Qt. However, I think it should still be F5 on mac. I have a few questions:

- how do you know, on mac, what shortcut is assigned to these browser actions? IIUC they don't show up in the menu, do them?
- are we sure we ever had Cmd+R doing refresh on Mac? From Qt docs it seems their default has been F5 for ever

dyfer:
> However, I think it should still be F5 on mac

I don't think so. Cmd-r is a [convention for the "reload" action](https://support.apple.com/en-us/102650#:~:text=a%20selected%20file.-,Command%2DR,-%3A%20(1)%20When%20an) on macOS and followed by many/most apps, e.g. Chrome.

> * how do you know, on mac, what shortcut is assigned to these browser actions? IIUC they don't show up in the menu, do them?

I don't know. But I'd press cmd-r as this is muscle memory from other apps.

> * are we sure we ever had Cmd+R doing refresh on Mac? From Qt docs it seems their default has been F5 for ever

I just double checked - in 3.13 cmd-r _does_ trigger reload in the help browser.

elgiano:
Thanks, I totally get your point, I would also do ctrl+r to reload on linux btw...

> https://github.com/supercollider/supercollider/commit/240d5b4ac61f9d8383e2046f1e19265622057531 seems to work correctly on macOS - cmd-r refreshes the browser window; in the document window F5 still works as reload and cmd-r as "replace".

This puzzles me, because [Qt docs](https://doc.qt.io/qt-6/qkeysequence.html#StandardKey-enum) don't mention Cmd+R at all. Then, in https://github.com/supercollider/supercollider/commit/f969f938b8cb561da830b2275af599ada928cbd2 I took the approach of setting the shortcut to whatever the WebView had set. It shouldn't be different from https://github.com/supercollider/supercollider/commit/240d5b4ac61f9d8383e2046f1e19265622057531

However, these shortcuts are both not visible and not customizable on Mac, so I'm thinking it would be better to make custom actions, so that they would be visible/editable in the settings.
We could initialize them to whatever Qt chooses for them, so not to break the existing behavior, and then let the user decide.

elgiano:
It turns out it's an error in Qt docs. Here we can see in the source they map Cmd+R to refresh on macOS, not F5:
https://github.com/qt/qtbase/blob/464e9298e24f3769f0eb776eb443fc7eb98271c0/src/gui/kernel/qplatformtheme.cpp#L234

yay!
so, for now I will just let the actions have whatever shortcut they had before, without making them configurable. I'll then sleep on it, and if we think it's better I'll look into making them into IDE actions that show up in the settings.

elgiano:
tl;dr: I think this PR is ready as it is, making these shortcuts configurable is extra work that is better suited for another PR eventually. So if people would give it a last test

### What to test
test that HelpBrowser shortcuts for "Back", "Forward" and "Reload" work when the HelpBrowser is in focus.
On Windows and Linux: shortcuts are displayed in the context menu (right click anywhere in HelpBrowser)
On macOS: the shortcuts are "Cmd+[" (Back), "Cmd+]" (Forward), and "Cmd+R" (Reload)

### Why not configurable now
I looked into making HelpBrowser actions' shortcuts configurable, and it's not easily doable. The shortcut preference page would complain that there are duplicate sequences, either with Replace or with Reload (Document) depending on OS. Then the user could configure them, but we couldn't take advantage of the same sequence triggering a different action for different focus widgets, which was the point of this PR.

I think the best way forward to make those actions visible and configurable would be to add an extra TreeWidget to ShortcutsPage, for a new category of shortcuts "HelpBrowser/shortcuts". This requires quite some refactoring of ShortcutsPage.

capital-G (approved):
Code looks fine to me and works on macOS.

I think you tested it on linux, so we now only need a test of it on windows and then it can be merged.

dyfer:
@dyfer will test on Windows

dyfer (approved):
Tested on Windows!

Thanks @elgiano for the clear testing instructions! Super helpful!

Issue #6742, opened 2025-04-18
unixCmd doesn't return the proper PID on Windows


dyfer:
## Environment

* SuperCollider version: 3.14-dev
* Operating system: Windows 11
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
// basic:
s.boot;
s.pid; // compare with the PID of the scsynth.exe in the Task Manager

// similarly when  running commands by hand:
r = (Server.program + Server.default.options.asOptionsString).unixCmd;
r.postln; // compare with the scsynth.exe's PID in the Task Manager

// incorrect PID also happens with ArrayedCollection -unixCmd, which has a different implementation IIUC
r = [Server.program, Server.default.options.asOptionsString].unixCmd; // note that this doesn't start the server on macOS due to shell quoting; however it does seem to work on Windows
r.postln; // pid is still incorrect though

// this makes it impossible to keep track of the process, e.g. to terminate it:
thisProcess.platform.killProcessByID(r); // doesn't stop the server process
```

## Expected vs. actual behavior

Expected: the PID matches the PID of the command. Actual: some other number is returned.

I found notes on related functionality in #1738 and #3539, but I don't see this problem being explicitly reported. 

On macOS and Linux we are getting the PID of scsynth reliably. It would be great to fix this...

This issue prevents the `TestServer_clientID_booted` tests from completing on Windows.

Spacechild1:
The issue is that scsynth is started via `unixCmd` which in turn runs the command inside a console (`cmd.exe`). As a consequence, `Server.pid` returns the PID of the `cmd.exe` instance and not the `scsynth.exe` instance. Will have a closer look.

dyfer:
Upon further investigation I found the following:
- `String -unixCmd` is supposed to run commands through a shell/console on all platforms
- by default on *nix, closing the shell process also closes its children
  - I think we can get a similar behavior on Windows if we run commands with `cmd /k` instead of `cmd /c` - testing
  - also, on macOS, running commands through shell `/bin/sh -c` for some reason does not start a separate shell and we get a pid of the destination command; not related to this PR, but a point in researching this
- `SequenceableCollection -unixCmd` is supposed to run commands _without a shell_. AFAICT this is not the case on Windows, where cmd.exe is used in that case as well. This should probably be fixed, investigating.

dyfer:
[...]

PR #6763, opened 2025-04-23
Tests in Windows - WIP

### LINKED ISSUE #4527: [CI] Run Unittests in AppVeyor
We currently do not run our UnitTest suite in AppVeyor. We should be running all tests (including ones that boot the server) in our Windows CI.

dyfer:
## Purpose and Motivation

I wanted to open a _draft_ PR to increase visibility of progress and issues of running the test suite on Windows in GHA.

TL;DR test suite runs, but there is a number of tests issues to address. I intend to keep this PR open as a draft and rebase it whenever the issue is fixed.

Fixes #4527.

Current tests failing only on Windows:

| test | issue | PR | notes |
|--|--|--|--|
| `TestServer_clientID_booted:test_remoteLogin` and `TestServer_clientID_booted: test_repeatedRemoteLogin ` | #6742  | | this PR has a workaround added to these tests; otherwise the whole test suite would hang | 
|`TestServer_boot:test_notifyAndServerBootActions` | | | |
|`TestDate:test_creation`| | | |
|`TestSCDocHTMLRenderer:test_htmlForLink_internalLink`| | | |
|`TestSCDocHTMLRenderer:test_htmlForLink_internalLinkWithLabel`| | | |
|`TestSCDocHTMLRenderer:test_htmlForLink_internalLinkWithAnchor`| | | |
|`TestSCDocHTMLRenderer:test_htmlForLink_internalLinkWithAnchorAndLabel`| | | |
|`TestSCDoc:test_helpSourceDirs_excludedExtensions`| | | |
|`TestFloat:test_asString_nan`| | | |
|`TestFilterUGens:test_Integrator_add_presample_once_only`| | | fails only sometimes? |

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing - definitely not! (not yet)
- [ ] Updated documentation


JordanHendersonMusic:
My opinion is that once the ci runs stably, this should be merged *without* fixing the other tests. The fact they fail means our test suite has been failing for years, we just couldn't observe it. That isn't this pr's concern.

---

Would it also be possible to add a 32 bit Linux version (separate pr)? We officially support raspberry pie, which is 32 bit, so we should test that. I see we have a 32 bit windows build already, but my understanding is that this isn't officially supported so it is perhaps not worth testing.

JordanHendersonMusic:
> I wonder how many people run SC on 32-bit RPis (RPi 1, 2, and Zero) 

I just looked on their website, it's up to 5 now, and the newer zero 2 is 64 bit. I think it is perfectly acceptable to stop supporting 32 bit builds altogether then - it would still work with the older version.

How do you think we might go about building consensus around dropping 32 bit support then? Not having 2 slot implementations would help reduce the code to maintain.

> This is correct! However, I would like to prevent a situation when we have a permanently failing testsuite for PRs that are unrelated. But do you think that's not an issue?

We'd just ignore the windows build results if there are known issues that are yet to be addressed... which is what we do already by not running it. Aside from the satisfaction of a green tick, I don't see the difference, particularly since we ignore it most of the time due to unsolved server synchronization issues.

> IMO we need to at least fix #6742

If the test suite hangs rather than fails, then yes I agree!

capital-G:
> How do you think we might go about building consensus around dropping 32 bit support then?

Please not yet, my wasm port requires a 32bit implementation (although no wasm build for sclang exists yet, we should keep this possible?). I think it would be advantageous if we keep this around - while 32 bit may be dead for desktop PCs and servers, small computers like RPi etc. still often rely on 32bit.

I once created a pipeline for RPi, see https://github.com/capital-G/supercollider/blob/cbc57cd34f2c5c501f37a15519e71ab06053595e/.github/workflows/actions.yml#L566

But due to emulation (GHA linux runners are x64) these are really slow - like 60 minutes for each build.
So maybe we should only run this upon request?

dyfer:
Most RPis are 64bit... but yeah, there might still be use cases for 32-bit. In any case, we are not dropping anything before 3.14 at the very least.

> BTW - I once created a pipeline for RPi, see

That's cool, I'd love to investigate that at some point. But it would be great to cross compile as well...

capital-G:
This is awesome - looking at the tests which are failing it seems manageable

```
Failed tests:

	[10/11]	TestDate:test_creation (0.0s)
		 ! ERROR: Primitive '_Date_Resolve' failed.
	[0/1]	TestSCDocHTMLRenderer:test_htmlForLink_internalLink (0.04s)
		 ! Got <a href="./..\Classes/View.html">View</a>, expected <a href="./../Classes/View.html">View</a>
	[0/1]	TestSCDocHTMLRenderer:test_htmlForLink_internalLinkWithLabel (0.04s)
		 ! Got <a href="./..\Classes/View.html">view class</a>, expected <a href="./../Classes/View.html">view class</a>
	[0/1]	TestSCDocHTMLRenderer:test_htmlForLink_internalLinkWithAnchor (0.04s)
		 ! Got <a href="./..\Classes/View.html#Mouse%20actions">View: Mouse actions</a>, expected <a href="./../Classes/View.html#Mouse%20actions">View: Mouse actions</a>
	[0/1]	TestSCDocHTMLRenderer:test_htmlForLink_internalLinkWithAnchorAndLabel (0.04s)
		 ! Got <a href="./..\Classes/View.html#Mouse%20actions">Hobbledy gobbledy</a>, expected <a href="./../Classes/View.html#Mouse%20actions">Hobbledy gobbledy</a>
	[0/1]	TestSCDoc:test_helpSourceDirs_excludedExtensions (0.02s)
		 ! ERROR: Primitive '_String_Regexp' failed.
	[0/1]	TestFloat:test_asString_nan (0.0s)
		 ! 
	[0/1]	TestFilterUGens:test_Integrator_add_presample_once_only (3.97s)
		 ! Integral of a single 1 should be 1

	----------------------------------------
	8 TESTS FAILED, 44 skipped, out of 4588
	----------------------------------------
```

E.g. HTML should always use unix path for the file locations, so this is indeed a bug on the SCDoc implementation.

So the only tests which need some more digging into

* `TestDate:test_creation`: Don't know whats wrong here? Maybe some rounding issue - or maybe b/c windows is not using unix time?
  https://github.com/supercollider/supercollider/blob/3761ef18635315ae6e5f7afc283ba44ed05a1b15/testsuite/classlibrary/TestDate.sc#L40-L43
  ```
  PASS: a TestDate: test_creation - Creation from all properties except dayOfWeek and rawSeconds
  ERROR: no valid time
  
  FAIL: a TestDate: test_creation - ERROR: Primitive '_Date_Resolve' failed.
  Failed.
  ```
* `TestSCDoc:test_helpSourceDirs_excludedExtensions` pops up multiple times - maybe we use some regex for scanning a file on windows, but using the the windows path separator `\` in it which needs escaping in a regex (which the unix separator `/` does not)
* `TestFilterUGens:test_Integrator_add_presample_once_only`: I don't know whats the problem in this case
  ```
  
  FAIL: a TestFilterUGens: test_Integrator_add_presample_once_only - Integral of a single 1 should be 1
  Is:
	   nil 
  Should be:
	   1
  ```

There are some problems regarding the printing of test results: if an error pops up it will be printed multiple times again across the test run - e.g. scroll down from https://github.com/supercollider/supercollider/actions/runs/15060058638/job/42333902861?pr=6763#step:6:17719

IMO this could be merged as is if we manage to fix the open bugs for 3.14 b/c I'd like to have a green pipeline for a release :)

dyfer:
> * `TestDate:test_creation`: Don't know whats wrong here?

The opening comment has an issue linked for this with more info :)

Let's try filing issues or PRs for the other failing tests.

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -82,6 +82,8 @@ SCDocHTMLRenderer {
 			};
 		};
 
+		if(thisProcess.platform.name == \windows) {result = result.replace("\\", "/")}; // replace backward with forward slash on Windows
```
capital-G, line 85: Couldn't we make this somehow more universal via

```
// enforce unix path for html
result = result.replace(Platform.pathSeparator, "/");
```

Platform dependent if statements can lead to some trouble while testing

dyfer:
@capital-G I think I found the issue with the Integrator test! It seemed to be a race condition between signaling the `Condition` from the OSC func vs by synth onFree.

This PR includes the fix.

Please do NOT squash :)

PR #6769, opened 2025-04-25
Make SCDoc images resizable

prko:
## Purpose and Motivation
**This is part of #6265. Since allowing HTML code in SCDoc is unlikely to be approved, I have implemented the same functionality using SCDoc syntax instead.**

Currently, there is no way to manually adjust the size of SCDoc images, as their dimensions are automatically determined based on the width of the HTML page and table cells.

**As a result, when rendering help documentation as HTML or converting it to PDF, larger images may get pushed to the next page, negatively affecting the document's layout. This PR addresses that issue.**

With this PR, users can control the image size by adjusting the vertical pixel value of an image. If vertical length control is needed, a height in pixels can be specified by adding a number after the third `#`, while the width remains automatically adjusted.

For example:
```supercollider
table::
##No options have been applied:
image::functions.png::Figure 1. This is ...
||The caption option has been applied:
image::functions.png#Figure 2. Function::This is ...
::
table::
##The caption and page link have been applied:
image::functions.png#Figure 3. Function#Reference/Functions#120::Click to see more information.
||The caption, page link, and vertical size have been applied:
image::functions.png#Figure 4. Function#Reference/Functions#120::Click to see more information.
::
```
Screenshot 2025-04-25 at 09 40 35

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/SCDocSyntax.schelp
+++ b/HelpSource/Reference/SCDocSyntax.schelp
@@ -247,17 +247,55 @@ teletype:: IMAGE\:: ::
 teletype::
 image::foo.png::
 ::
-Optionally an image caption can be given:
+Optionally, an image caption can be provided:
 teletype::
 image::foo.png#Figure 1::
 ::
 
-And, the image can be clickable and link to another page:
+Additionally, the image can be clickable and linked to another page:
 teletype::
 image::foo.png#click this image#Classes/SinOsc::
 ::
 
-Just leave the caption empty if you want a link but no caption.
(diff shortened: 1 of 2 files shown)


prko:
The initial commit did not function correctly with linked text, as the third `#` was interpreted as part of the link address. To resolve this issue, the vertical size is positioned now between the file path and the first `#`, and the vertical size is prefixed by `@`.

If the variable name `imageScale` is not correct, please let me know. I think `imageHeight` could be better.

Issue #6644, opened 2025-02-25
Optional Keyword Argument Method Extention Documentation


JordanHendersonMusic:
## Motivation

Since the keyword expansion a new way of extending method arguments is avaible that doesn't break backwards compatibility, that is, to only let the argument match when provided as a keyword, *not* as a positional argument through the use of `... args, kwargs` --- as in done in python.

While not currently used in supercollider, establishing this as a good practice (and documentation support) should be done first to avoid suprises.

e.g.
```supercollider
// old
A {
   foo{ |a, b| 
      ... 
   }
}
// new
A {
   foo { |a, b ...args, kwargs| 
      kwargs.asDict[\c] !? { ... do someting new with 'c'... }
      ... do the old thing ...
   }
}

```

**However**, often these keyword-only arguments come in groups and cannot be assigned independantly.
For example, `A.foo` might work with keyword arguments (`c` **and** `d`) **or** (`e` **and** `d` **and** optionally `f`), but **not** all four.

The document needs a way to communicate this.

Currently it looks like this...
```supercollider
A.foo(a: 1, b: 0)
   Arguments:
      a: ...
      b: ...
```

## Description of Proposed Feature

We should agree on what this should look like. 

Here is a proposal to get us started.

```supercollider
A.foo(a: 1, b: 0)
   Arguments:
      a: ...
      b: ...
   Additional Keyword-Only Options:
      Option: Name of C and D Option
         **required** c: ...
         **required** d: ...
      Option:  Name of D and E Option
         **required** d: ...
         **required** e: ...
         **optional** f: ...      
```

## Plan for Implementation

Lets agree first.

JordanHendersonMusic:
Just directly tagging a few whom have either been involved with kwargs or documentation changes incase this interests anyone😄 

@capital-G @telephon @prko

prko:
I would be happy to do something for the development of SC as soon as my capacity allows. I always have a positive opinion about the improvement of sclang. It seems to provide flexibility and looks good!

capital-G:
Hey @JordanHendersonMusic - I must admit don't really understand the issue here, so please bare with me :D

You basically want to introduce a dispatching based on the given kwargs which mimics more an overloading of a function?
In python I mostly use `**kwargs` within inheritance (as it allows for a more stable API if - as mentioned - a new keyword argument gets introduced) or during injection and I rarely use it for method dispatching.

> Here is a proposal to get us started.
> 
> ```
[...]

Issue #6743, opened 2025-04-18
Sclang sometimes doesn't load default library paths [Windows]


dyfer:
## Environment

* SuperCollider version: 3.14-dev
* Operating system: Windows 11
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

This is hard to reproduce reliably... may happen upon numerous retries.

```shell
# start sclang

C:\Users\Marcin\Desktop\SuperCollider>sclang.exe
compiling class library...
        Found 868 primitives.
        Compiling directory 'C:\Users\Marcin\Desktop\SuperCollider\SCClassLibrary'
...
*** Welcome to SuperCollider 3.14.0-dev. *** For help press F1.
sc3>
# press ctrl-d

# start sclang again:
C:\Users\Marcin\Desktop\SuperCollider>sclang.exe

# press ctrl-d

# ...repeat 10-20 times

# result (sometimes):
compiling class library...
        Found 868 primitives.
ERROR: There is a discrepancy.
numClassDeps 0   gNumClasses 82
Library has not been compiled successfully.

```

## Expected vs. actual behavior

Previously I thought that this happens if sclang crashes or gets terminated... but I also saw the behavior after sclang exited cleanly.
First observed when running the test suite, sclang exited with a non-0 exit code, then failed to start again. This might be a red herring though.

On the local machine, after the issue occurs, subsequent starts seem to work. In CI, sclang keeps failing to find the included class library. This is the main issue preventing the testsuite from running on Windows currently.

PR #6773, opened 2025-04-26
sclang: fix cli options flags initialization

### LINKED ISSUE #6764: Library fails to compile fully after boost CLI changes
## Environment

* SuperCollider version: Commit ID 06aaa4da or later
* Operating system: n/a

## Steps to reproduce

```supercollider
$ sclang
compiling class library...
	Found 882 primitives.
ERROR: There is a discrepancy.
numClassDeps 0   gNumClasses 82
ERROR: Library has not been compiled successfully.
cleaning up OSC
```

Commit 06aaa4da was found by git bisect.

If I go to the commit immediately before 06aaa4da, then sclang starts and runs normally.

I *think* the issue is that some locations (perhaps all locations) under Extensions are not being compiled. My startup.scd file contains references to at least one class defined under Extensions, and in the broken situation, if I open the IDE, I see compilation errors in the startup file specifically for that class (but the class is found in the working commit, without changing anything in the configuration).

Actually, I can confirm that -- if I explicitly add `/home/xxx/.local/share/SuperCollider/Extensions` to includePaths in sclang_conf.yaml, then the bad commit starts correctly.

So apparently, something in the bad commit removes Extensions from the list of default locations.

elgiano:
## Purpose and Motivation

Fixes #6764

After #6644, uninitialized cli options caused problems at startup for sclang on linux (see issue above).
The issue was that bool members of SC_TerminalClient::Options need to be initialized to `false`, since they are not zeroed by the default constructor. Before #6644, it was done by the user-defined constructor now removed.
The effect is that sclang is forced in either daemon (no interactivity) or standalone (no lib) mode.

Initializing all bool members to false fixes it.

I don't know why this is a problem only on linux and not on macos and windows.

## Types of changes

Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing

### DIFF
--- a/lang/LangSource/SC_TerminalClient.h
+++ b/lang/LangSource/SC_TerminalClient.h
@@ -56,10 +56,10 @@ public:
 
     struct Options : public SC_LanguageClient::Options {
         std::string mLibraryConfigFile;
-        bool mDaemon;
-        bool mCallRun;
-        bool mCallStop;
-        bool mStandalone;
+        bool mDaemon = false;
+        bool mCallRun = false;
+        bool mCallStop = false;
+        bool mStandalone = false;
         std::vector<std::string> mArgs;
     };
 


JordanHendersonMusic (approved):
This is a good idea regardless of whether it fixes the bug. 

Was the issue then, that the old approach assigned to these members when it was true or false, but the new body approach only assigns when true? @capital-G does that sound feasible? If this is the case, are there any other initialised variables used in this part of the code?

jamshark70:
I confirm that the issue is fixed on my machine with the changes here.

JordanHendersonMusic:
@jamshark70 okay, in that case I will merge because this is a trivial change otherwise.

capital-G:
Thanks for the fix!

So,  just for me to learn stuff: Via https://github.com/supercollider/supercollider/pull/6640#discussion_r2018242666 I removed the custom constructor, but when I removed it I failed to set default values in the struct?

Somehow I would have expected the compiler would complain about this? But I can't find a warning in the logs of https://github.com/supercollider/supercollider/actions/runs/14662014643/job/41148462797

So the boolean values are actually undefined and do not default to a `false` value? I would expect that the compiler would at least yield a warning, but I think a compile error would be more valid at this point as it leads to unexpected results.

Maybe we should add `-Wmissing-field-initializers` for compilation at least within CI? See https://stackoverflow.com/a/37649654 - is there a way to turn specific warnings into errors as well?

JordanHendersonMusic:
> I removed the custom constructor, but when I removed it I failed to set default values in the struct?

Sort of, boolean's default constructor leaves the variable uninitialised, but it was called, just as the string and vector's were, which initialised them as you'd expect. The problem is that the default constructor of literal types doesn't *zero initialise* the value. You can explicitly call the zero intialiser on the boolean like this `bool b{};`, which would have been another solution. 

Another option would be to call the zero initialiser on the options struct `Options m_options{}` as this will zero initalise all the members. 

This is way people recommend creating struct/classes with `T t{}` and not having any constructors. 

Online, people moan about memory safety in C++, but I've always found initialisation and literal casting to be the hard parts of C++. In Zig, if you want to leave a variable in an undefined state, you write `var b:bool = undefined`, otherwise it is set to false.

>  is there a way to turn specific warnings into errors as well?

Ideally, we'd turn all warnings into errors

`-Werror -Wall -Wextra`

The problem is that we have a tone of warnings that need to be fixed before that can happen. I'm not aware of being able to turn a specific type of warning on. But you can turn warning off for specific parts of the code base using GCC pragma macros.

Another option is to use warnings in the IDE so you can see the warning for only the code you are writing and not the whole project. IMO, clangd does a far better job of this than vscode.

dyfer:
Thanks for the explanations @JordanHendersonMusic !

Re: removing the default constructor

So was the constructor supposed to _always_ initialize the values before the #6640? In that case we won't know what the original reason for #6743 was, right? It _seemed_ like the same issue like the one fixed by this PR, but occurring _before_ #6640 and only on some Windows systems... I just wanted to bring this up as this was a behavior that occurred inconsistently, so I guess it's possible that it was caused by some uninitialized memory and I want to see if there's anything else in the codebase that could've caused that.

elgiano:
boolean flags were initialized to false by the SC_TerminalClient::Options() constructor. But from the posted messages in #6743 it looks to me like standalone mode was on anyways? One way to investigate this would be to checkout sc before #6640, add some post messages to check all the flags, and all the steps of `SC_TerminalClient::run`, and take it from there...

PR #6778, opened 2025-04-26
git: ignore clangd cache

dyfer:
## Purpose and Motivation

I was experimenting with clangd and it creates cache files, which would be good to ignore.

I added this before the .vscode ignores in order to not create a conflict with #6351, but I'm not sure if it'll work that way. Maybe this should wait for #6351 to be merged first...

## Types of changes

- New feature (?)

## To-do list

- [x] Code is tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,5 +36,8 @@ __pycache__
 #vim swap files
 .*.sw?
 
+# clangd
+.cache/*
+
 # vs code files
 .vscode


JordanHendersonMusic (approved):
I don't think we need to wait for the vscode pr, it would be simple to resolve, mind if I merge?

dyfer:
Merging this would be fine with me!

capital-G (approved):

PR #6787, opened 2025-04-28
doc: add undefined overflow notes to affected Number, RawArray classes

### LINKED ISSUE #6687: docs: Document overflow behaviour of `IntNArray`s
## Environment

* Operating system: macOS 15.3.2

## Steps to reproduce

```supercollider
Int32Array[2 ** 31 - 1] == Int32Array[2 ** 31] /* overflow */ 
// -> true 
Int16Array[2 ** 15 - 1] == Int16Array[2 ** 15] /* overflow */ 
// -> false
Int8Array[2 ** 7 - 1] == Int8Array[2 ** 7] /* overflow */ 
// -> false
DoubleArray[2 ** 1023.99999999999994] == DoubleArray[2 ** 1024] /* overflow */
// -> false
FloatArray[2 ** 127.999999957] == FloatArray[2 ** 128] /* overflow */
// -> false
```

Expect the buffer overflow value behavior to be the same for all `IntNArray`s, so all truncated to the max value (as is the case only for `Int32Array`) or all equal to 0 (as is the case only for `Int8Array`, `Int16Array`).

In actuality, `Int32Array` returns its max value `2 ** 31 - 1` (on macOS and Linux; for Windows, actually returns the minimum value, `-2 ** 31`) for all overflow values, whereas `Int8Array`, `Int16Array` go to `0`.

(As a note, `FloatArray` and `DoubleArray` have `inf` as their overflow value, which makes sense as they're floating point.)

passyur:
## Purpose and Motivation

Closes #6687. Overflow is an interesting behavior whose definition or lack thereof should be apparent in the documentation for classes that exhibit the behavior. Overflow behavior for float and (signed) integer types in C++ is undefined, and those types are on the backend of several `Number` and `RawArray` classes in sclang. 

There's already a divergence in overflow behavior for the `Int32Array` dependent on the platform (see linked issue).

(If I missed any other affected classes, please let me know.)

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Integer.schelp
+++ b/HelpSource/Classes/Integer.schelp
@@ -1,12 +1,14 @@
 class:: Integer
-summary:: Integer number
+summary:: 32-bit Integer number
 categories:: Math
 
 description::
-A 32 bit integer. Integer inherits most of its behaviour from its superclass.
+A 32-bit integer. Integer inherits most of its behaviour from its superclass, link::Classes/SimpleNumber::.
 
 note::
-A 32 bit signed integer can represent values in the range -2147483648 to 2147483647. Adding up further returns in a wrapped result, so that and 2147483647+1 = -2147483648. For a larger range, one can use link::Classes/Float::, which is 64 bit and supports many (but not all) numerical methods that int does.
+Integer can represent values in the range -2147483648 (code::-2^31::) to 2147483647 (code::2^31 - 1::).
+
+The overflow behavior of an Integer is undefined. This occurs whenever the result of an operation does not fit in the range of values supported by the return type, in this case, a 32-bit signed integer. Consider using the 64-bit link::Classes/Float:: for larger numbers (up to code::±2^53::).
 ::
 
 
--- a/HelpSource/Classes/DoubleArray.schelp
+++ b/HelpSource/Classes/DoubleArray.schelp
@@ -27,6 +27,11 @@ list::
 ## link::Classes/SymbolArray:: - symbols
 ::
 
+note::
+The overflow/underflow behavior of an element in a DoubleArray is undefined but usually code::inf::/code::-inf::, respectively. This occurs whenever the result of an operation does not fit in the range of values supported by the return type, in this case, a 64-bit floating point number. 
+See the note in link::Classes/Float::.
+::
+
 INSTANCEMETHODS::
 
 method::readFromStream
--- a/HelpSource/Classes/Float.schelp
+++ b/HelpSource/Classes/Float.schelp
@@ -3,7 +3,7 @@ summary:: 64-bit Floating point number
 categories:: Math
 
 description::
-A 64-bit double precision floating point number. Float inherits most of its behaviour from its superclass.
+A 64-bit double precision floating point number. Float inherits most of its behaviour from its superclass, link::Classes/SimpleNumber::.
 
 code::
 // generate 10 random floats between 0 and 1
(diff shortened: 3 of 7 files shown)


JordanHendersonMusic:
Is the float (and double) case undefined? I think the IEE standard says it should be set to inf or -inf. This depends whether we have fastmath on, which I don't know.

passyur:
@JordanHendersonMusic There's some hedging in this C++ reference: https://en.cppreference.com/w/cpp/language/types#Standard_floating-point_types

"float — single precision floating-point type. **Usually** [IEEE-754 binary32 format](https://en.wikipedia.org/wiki/Single-precision_floating-point_format).
double — double precision floating-point type. **Usually** [IEEE-754 binary64 format](https://en.wikipedia.org/wiki/Double-precision_floating-point_format)." (emphasis mine)

"Undefined" may be too strong, but it is definitely not specified by the C++ standard itself. I'd consider it hardware definition per the FPU.

HelpSource/Classes/Integer.schelp
```diff
@@ -1,12 +1,12 @@
...
+summary:: 32-bit Integer number
 categories:: Math
 
 description::
-A 32 bit integer. Integer inherits most of its behaviour from its superclass.
+A 32-bit integer. Integer inherits most of its behaviour from its superclass.
 
 note::
-A 32 bit signed integer can represent values in the range -2147483648 to 2147483647. Adding up further returns in a wrapped result, so that and 2147483647+1 = -2147483648. For a larger range, one can use link::Classes/Float::, which is 64 bit and supports many (but not all) numerical methods that int does.
+The overflow behavior of an Integer is undefined. Consider using the 64-bit link::Classes/Float:: for larger numbers.
```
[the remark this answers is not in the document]
passyur: Agreed; I added it back in the latest commit.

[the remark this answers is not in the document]
JordanHendersonMusic: Perhaps here it is worth mentioning floats can go to +- 2^53.

HelpSource/Classes/Integer.schelp
```diff
@@ -1,12 +1,14 @@
...
 
 description::
-A 32 bit integer. Integer inherits most of its behaviour from its superclass.
+A 32-bit integer. Integer inherits most of its behaviour from its superclass.
 
 note::
-A 32 bit signed integer can represent values in the range -2147483648 to 2147483647. Adding up further returns in a wrapped result, so that and 2147483647+1 = -2147483648. For a larger range, one can use link::Classes/Float::, which is 64 bit and supports many (but not all) numerical methods that int does.
+Integer can represent values in the range -2147483648 to 2147483647.
+
+The overflow behavior of an Integer is undefined. Consider using the 64-bit link::Classes/Float:: for larger numbers (from code::±2^53::).
```
dyfer, line 11: Unless I'm missing something, "from 2^53" should be -> "up to 2^53" ?

If using exponent here, I'd also include it in the line about the Integer range:

> Integer can represent values in the range -2147483648 to 2147483647 (±2^31).

Though technically that's -2^31 to +2^31-1, so I'm not sure how to handle this both elegantly and accurately.

telephon (approved):

dyfer (approved):
Thanks! I believe all review comments have been addressed.

I'll leave this to be merged by the doc team - @HotwheelsSisyphus or @tedmoore

HelpSource/Classes/Integer.schelp
```diff
@@ -1,12 +1,14 @@
 class:: Integer
-summary:: Integer number
+summary:: 32-bit Integer number
 categories:: Math
 
 description::
-A 32 bit integer. Integer inherits most of its behaviour from its superclass.
+A 32-bit integer. Integer inherits most of its behaviour from its superclass.
```
tedmoore, line 6: Should we list what it's superclass is right here?

tedmoore (approved):
One thing that's worth considering is that many musician's coming to these documents won't know what "overflow behavior" means. It may be good to link to this idea for further understanding: https://en.wikipedia.org/wiki/Integer_overflow

HelpSource/Classes/Integer.schelp
passyur: I've also gone ahead and added it to Float.schelp since they have a parallel structure.

passyur:
> One thing that's worth considering is that many musician's coming to these documents won't know what "overflow behavior" means.

@tedmoore I see what you mean. My only concern is that it's a bit of information-overkill to link elsewhere, and we might just say something inline, like: "The overflow/underflow behavior of [Float/Integer/DoubleArray/FloatArray/WhatEver] is undefined. This occurs whenever the result of an operation does not fit in the range of values supported by the return type."

tedmoore:
Thanks @passyur !

PR #6800, opened 2025-05-02
Scsynth/buffer sample rate initializing

tremblap:
## Purpose and Motivation

As discussed extensively in #6710 implementing an instantiation time sampling rate attribute to the .new and .alloc buffer methods is needed.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


tremblap:
The history will be cleaned, but this branches from #6710

tremblap:
review code:
```
Server.scsynth
s.options.sampleRate_(48000)
s.reboot

a = Buffer.new(s,10,2,sampleRate: 22000) //impose without alloc
a.alloc
a.query

b = Buffer.new(s,11,5) //default
b.alloc
b.query

c = Buffer.alloc(s,12,3,sampleRate: 33221)//impose on alloc
c.query

d = Buffer.allocConsecutive(3,s,24,5,sampleRate: 12345) //test consecutive
d[0].query

e = Buffer.alloc(s,1) //minima call
e.query

// retest with supernova
Server.supernova
s.reboot
```

Spacechild1:
Cool! Will have a closer look over the weekend.

For now I have one question: why are you basing this PR on https://github.com/supercollider/supercollider/pull/6710? Or in other words: what's the point of having two separate PRs when one fully contains the other? I would just start from the current develop branch. Once *this* PR has been merged, you can always rebase #6710 if necessary.

capital-G:
> [bloody clang-format will I ever get it](https://github.com/supercollider/supercollider/pull/6800/commits/898eab7cde7b2359a73e0c4347850023918966c0)

You just have to setup pre-commit once (see https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit) afterwards the code will always be formatted properly during commit, so no more formatting issues :)

tremblap:
ok this should be right now. Here it works on MacOS in scsynth and in supernova

dyfer:
@tremblap this still seems to have compilation error, please take a look.

dyfer:
Yes, sc_error does obfuscate the real error; I searched for "error:" (with a semicolon) and found it right away:
https://github.com/supercollider/supercollider/actions/runs/14840364087/job/41662127812?pr=6800#step:14:7670

tremblap:
thanks that makes sense... but this is failing on a test of something that we are changing, so should I edit the test? That feels like cheating :D

tremblap:
ok I get it all to pass except TestServer_boot's test_notifyAndServerBootActions and test_notifyAndServerTreeActions, neither of which touching the buffers nor the new code. Help?

dyfer:
> ok I get it all to pass except TestServer_boot's test_notifyAndServerBootActions and test_notifyAndServerTreeActions

These tests fail semi-randomly, don't worry about them.

tremblap:
ok so the question of @Spacechild1 is semi relevant now - we can merge this then the other, since both are passing the tests, whenever you want.

I wanted to split because I didn't want to do this which seems stucked in the weeds of various query/sync/stuff (see the other thread) and anyway it was more disciplined that way. So if/when this one is merged, I'll rebase the other one on the updated tip of develop and we can merge it right after.

dyfer:
@tremblap there seems to be some misunderstanding. AFAICT, this PR includes commits from https://github.com/supercollider/supercollider/pull/6710, so it includes both `setSampleRate` and `new`/`alloc` implementations. If we merge this, there will be nothing to merge from 6710. 

In other words, this PR is a complete implementation, which is what I wanted 6710 to be.

Depending how @Spacechild1 feels about this, we could review this as a whole, or, as you seemed to have planned for it, only consider the `new`/`alloc` changes, in which case you'd need to remove `setSampleRate` functionality from this PR.

dyfer:
> query/sync/stuff (see the other thread)

There were multiple attempts to explain that these were not relevant concerns and you could've just continue work in that PR.

tremblap:
oh you're right - shit - ok I tried to clean my act and made it worse. Probably the least offensive way forward now is to review this one as a whole (since you decided it was conditional to the feature I actually need :) ) and if that is accepted we nuke 6710.

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +11,40 @@ Buffer {
...
 		server = server ? Server.default;
 		bufnum ?? { bufnum = server.nextBufferNumber(1) };
 		^super.newCopyArgs(
 			server,
 			bufnum,
 			numFrames,
-			numChannels
-		).sampleRate_(server.sampleRate).cache
+			numChannels,
+			sampleRate !? _.asFloat ?? 0.0;
```
telephon, line 22: drop semicolon (also below)

telephon (approved):
This looks good. Just one cosmetic change to be done.

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -595,6 +597,17 @@ Buffer {
...
 
+	setSampleRate { arg value;
+		// sends the message first to check sanity of sr before changing the langage-side instance variable
+		server.listSendMsg(this.setSampleRateMsg(value));
+		sampleRate = value !? _.asFloat;
+    }
+
+	setSampleRateMsg { arg value;
+		if(bufnum.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
+		^[\b_setSampleRate, bufnum, value !? _.asFloat ?? 0.0]
```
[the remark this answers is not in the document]
tremblap: this was discussed in the other thread and was suggested in a review by yourself 3 weeks ago... 
it has to be left nil on language to be compatible with old code
it has to be converted to float for supernova to be happy
so 0.0 is the nil sr code, which is handled both on language and both servers

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -1985,21 +1985,30 @@ template <bool realtime> void handle_b_alloc(ReceivedMessage const& msg, endpoin
...
+    float sampleRate;
 
     args >> bufferIndex >> frames;
 
-    if (!args.Eos())
+    if (!args.Eos()) {
         args >> channels;
-    else
+        if (!args.Eos()) {
+            args >> sampleRate;
```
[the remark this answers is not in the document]
tremblap: ok so this is how to stack it at the end. I'll change accordingly.

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -2493,6 +2502,22 @@ void handle_b_setn(ReceivedMessage const& msg) {
...
+    ++it;
+
+    SndBuf* buf = sc_factory->get_buffer_struct(buffer_index);
+    if (!buf) {
+        log_printf("/b_setSampleRate called on unallocated buffer\n");
+        return;
+    }
+
+    buf->samplerate = it->AsFloat();
+    buf->sampledur = 1.0 / buf->samplerate;
```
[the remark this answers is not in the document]
tremblap: done in the first pass as described above

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -67,7 +69,7 @@ Buffer {
 
 	allocMsg { arg completionMessage;
 		this.cache;
-		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, completionMessage.value(this)]
+		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, sampleRate, completionMessage.value(this)]
```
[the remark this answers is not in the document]
Spacechild1: > even for completion messages?

I don't understand. You cannot change the order of arguments in a public API. The Server commands are part of the public API. Just make sure that the new `sampleRate` argument comes last.

>  in all cases it is almost impossible to test

I don't understand what you mean by that.

SCClassLibrary/Common/Control/Buffer.sc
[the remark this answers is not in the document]
Spacechild1: > it has to be left nil on language to be compatible with old code

That's not what I meant. There is no reason why this method *must* support `nil` in the first place. You could also throw an error. However, it's indeed a nice feature that you can use `nil` to (re)set the Buffer to the Server sample rate. All I'm saying is that you need to properly handle it on the Server side, which you are currently *not* doing. Again, see the corresponding comments in the OSC handlers.

server/scsynth/SC_MiscCmds.cpp
```diff
@@ -1406,6 +1406,20 @@ SCErr meth_b_setn(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRe
...
+SCErr meth_b_setSampleRate(World* inWorld, int inSize, char* inData, ReplyAddress* inReply);
+SCErr meth_b_setSampleRate(World* inWorld, int inSize, char* inData, ReplyAddress* inReply) {
+    sc_msg_iter msg(inSize, inData);
+    int bufindex = msg.geti();
+    SndBuf* buf = World_GetBuf(inWorld, bufindex);
+    if (!buf)
+        return kSCErr_Failed;
+
+    buf->samplerate = msg.getf();
+    buf->sampledur = 1.0 / buf->samplerate;
```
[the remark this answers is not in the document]
Spacechild1: > by then it has been updated to a valid value (0 will have been converted to server samplerate)

Where? There is no place where this could have be done. I think you are confusing this with the `/b_alloc` command.

And again, make sure that the sample rate is not negative.

server/scsynth/SC_SequencedCommand.cpp
```diff
@@ -287,7 +288,9 @@ void BufAllocCmd::CallDestructor() { this->~BufAllocCmd(); }
 bool BufAllocCmd::Stage2() {
     SndBuf* buf = World_GetNRTBuf(mWorld, mBufIndex);
     mFreeData = buf->data;
-    SCErr err = bufAlloc(buf, mNumChannels, mNumFrames, mWorld->mSampleRate);
+    // use Server sample rate as default value
+    auto sampleRate = mSampleRate != 0.0 ? mSampleRate : mWorld->mSampleRate;
+    SCErr err = bufAlloc(buf, mNumChannels, mNumFrames, sampleRate);
```
Spacechild1, line 293: Make sure that `bufAlloc` returns `kSCErr_Failed` for negative sample rate values.

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -1985,21 +1985,30 @@ template <bool realtime> void handle_b_alloc(ReceivedMessage const& msg, endpoin
...
+    }
 
     completion_message message = extract_completion_message(args);
 
     cmd_dispatcher<realtime>::fire_system_callback([=, message = std::move(message)]() mutable {
         sc_ugen_factory::buffer_lock_t buffer_lock(sc_factory->buffer_guard(bufferIndex));
         try {
             sample* free_buf = sc_factory->get_nrt_mirror_buffer(bufferIndex);
-            sc_factory->allocate_buffer(bufferIndex, frames, channels);
+            sc_factory->allocate_buffer(bufferIndex, frames, channels, sampleRate);
```
Spacechild1, line 2011: Make sure that `sc_plugin_interface::allocate_buffer` throws an error if the sample rate is <= 0, similar to how it handles frames/channels.

tremblap:
ok for now SR <= 0 revert to server sample rate. elegant,  no?

tremblap:
so @Spacechild1 if that can be agreeable, this is now finished. if not, then please propose how you would check

HelpSource/Classes/Buffer.schelp
```diff
@@ -466,15 +472,21 @@ Returns the number of channels in the corresponding server-side buffer.
...
 ::
 warning::
 You can use link::#-query:: to inspect the buffer's sample rate on the server, but this will also emphasis::update the:: code::sampleRate:: emphasis::instance variable::, overwriting its value which you may have previously set.
 ::
 
+
+method:: setSampleRate
+Sets the sample rate of the buffer on the server side, and updates its value on the language side. This does not resample the audio.
+argument:: value
+The new desired sample rate.
```
Spacechild1, line 487: Please document that `nil` (re)sets to the Server sample rate.

HelpSource/Reference/Server-Command-Reference.schelp
```diff
@@ -720,6 +721,14 @@ table::
 ::
 Set contiguous ranges of sample indices to sets of values. For each range, the starting sample index is given followed by the number of samples to change, followed by the values.
 
+subsection:: /b_setSampleRate
+Set the sampling rate of the buffer.
+table::
+## strong::int:: || buffer number
+## strong::float:: || the desired sampling rate
```
Spacechild1, line 728: Please document that you can pass `0.0` to (re)set to the Server sample rate.

Spacechild1 (changes requested):

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -67,7 +69,7 @@ Buffer {
 
 	allocMsg { arg completionMessage;
 		this.cache;
-		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, completionMessage.value(this)]
+		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, completionMessage.value(this), sampleRate]
```
Spacechild1, line 72: Here we also need to do `sampleRate !? _.asFloat ?? 0.0`!

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +11,40 @@ Buffer {
...
 		server = server ? Server.default;
 		bufnum ?? { bufnum = server.nextBufferNumber(1) };
 		^super.newCopyArgs(
 			server,
 			bufnum,
 			numFrames,
-			numChannels
-		).alloc(completionMessage).sampleRate_(server.sampleRate).cache
+			numChannels,
+			sampleRate !? _.asFloat ?? 0.0
```
Spacechild1, line 34: forward `sampleRate` as is! `alloc` is responsible for sending the appropriate value to the Server, see my comment in `allocMsg`.

Spacechild1 (changes requested):

Spacechild1:
@tremblap Thanks! We are getting very close :) See my review comments.

SCClassLibrary/Common/Control/Buffer.sc
Spacechild1: We also need to update the documentation for `allocMsg`!

tremblap: all done, except the 'doc' of 'allocMsg' which is already quite terse and has none of the arguments (so no need to add anything)

Spacechild1: Sorry, I get confused for a moment :) You are absolutely right.

dyfer:
Pinging @elgiano in case he wants to take a look before merging.

elgiano:
Hey all, I can take a look and test it on linux before the next dev meeting, if you want an extra pair of eyes on it!

@tremblap thanks for this! It's a feature I also desired for a long while...

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -11,38 +11,40 @@ Buffer {
...
 		server = server ? Server.default;
 		bufnum ?? { bufnum = server.nextBufferNumber(1) };
 		^super.newCopyArgs(
 			server,
 			bufnum,
 			numFrames,
-			numChannels
-		).sampleRate_(server.sampleRate).cache
+			numChannels,
+			sampleRate !? _.asFloat ?? 0.0
```
elgiano, line 22: I don't understand why there has to be a difference between *new and *alloc in converting the supplied sampleRate. Since the conversion is done when sending messages (allocMsg, setSampleRateMsg), I think we can remove it from here.

server/scsynth/SC_SequencedCommand.cpp
```diff
@@ -279,6 +279,9 @@ int BufAllocCmd::Init(char* inData, int inSize) {
 
     GET_COMPLETION_MSG(msg);
 
+    auto sampleRate = msg.getf(0);
```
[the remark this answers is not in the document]
tremblap: thanks @elgiano for the thorough review. I will reply inline, but I didn't love that it is the only one and is quite style-awkward. It is a catch 22. If we make it style-consistent it breaks code. if we don't it adds friction to SClang. It is the community's decision to chose which is less painful. I vote for syntax-consistance all the time but who am I to know

SCClassLibrary/Common/Control/Buffer.sc
tremblap: there shouldn't be so I'll correct

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -595,6 +597,17 @@ Buffer {
 		^[\b_query, bufnum]
 	}
 
+	setSampleRate { arg value;
+		// sends the message first to check sanity of sr before changing the langage-side instance variable
+		server.listSendMsg(this.setSampleRateMsg(value));
+		sampleRate = value !? _.asFloat;
```
elgiano, line 603: If one calls `buf.setSampleRate(0)` or `buf.setSampleRate` this will set client-side sampleRate to 0 or nil. Then, when asking for `buf.duration` one would get either `inf` or `ERROR: binary operator '/' failed` respectively.

Perhaps instead of setting sampleRate, we should call this.updateInfo(action) and add action to args?

server/scsynth/SC_MiscCmds.cpp
```diff
@@ -1406,6 +1406,21 @@ SCErr meth_b_setn(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRe
     return kSCErr_None;
 }
 
+SCErr meth_b_setSampleRate(World* inWorld, int inSize, char* inData, ReplyAddress* inReply);
+SCErr meth_b_setSampleRate(World* inWorld, int inSize, char* inData, ReplyAddress* inReply) {
+    sc_msg_iter msg(inSize, inData);
+    int bufindex = msg.geti();
+    SndBuf* buf = World_GetBuf(inWorld, bufindex);
```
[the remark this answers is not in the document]
tremblap: I should know, FluCoMa does it a lot... yet it completely slipped my mind. I think your proposal is slick, if @Spacechild1 and @dyfer agree.

tremblap:
> Had a looked and tested it, there is only one major problem with alloc in scsynth, when no completion message is supplied.

This is strange, since I tested with the code above (2nd post in this thread) and it all behaved. I'll try again, but maybe I misunderstand what you are testing that I am not testing above?

tremblap:
> And then one can run it with `UnitTest.runTest("TestBuffer_Server:test_setSampleRate")`

feel free to push it on this branch. Or I can put it myself in testsuite/classlibrary/TestBuffer.sc ?

elgiano:
Pushed the tests and a few typo fixes. Remaining issues:
- GET_COMPLETION_MSG: handle next arg when completionMsg is 0: 
https://github.com/supercollider/supercollider/pull/6800#discussion_r2095863636
- RT/NRT: update both server buffers info in .setSampleRate: 
scsynth: https://github.com/supercollider/supercollider/pull/6800#discussion_r2096316911
supernova: https://github.com/supercollider/supercollider/pull/6800#discussion_r2096333441
- update sclang cached sampleRate on alloc/setSampleRate
https://github.com/supercollider/supercollider/pull/6800#discussion_r2096495246
https://github.com/supercollider/supercollider/pull/6800#discussion_r2096033463

While this is being done, I also would like someone else to have a look at the tests I wrote

tremblap:
ok the code pushed by @elgiano is not working anymore - line 47 of the class def has 2 commas, and the order of arguments is not working when I correct that. I'll investigate and try to plough through the todo list. Since it was working when I submitted it, I have to trace back when things got mixed up...

SCClassLibrary/Common/Control/Buffer.sc
tremblap: for this one, should we do that replacement there and then also in setSampleRateMsg, or just in the latter, or just in the former?

tremblap:
ok this is all done and good, except for https://github.com/supercollider/supercollider/pull/6800#discussion_r2096495246 which I left a question as I am not too sure how to best manage it.

tremblap:
Don't worry I will rebase once this last question is solved, and I have tested it. BTW, I don't know how to run the unit-tests on my local machine, but they look good.

dyfer:
> BTW, I don't know how to run the unit-tests on my local machine, but they look good.

https://github.com/supercollider/supercollider/wiki/Unit-Testing-Guide
In SC: 
1. add `<supercollider_source>/testsuite/classlibrary` folder to your SuperCollider compile paths (you can do that by running  https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/add_include_path.scd from your local repo)
2. reboot interpreter
3. run test, e.g. with `UnitTest.runTest("TestBuffer_Server:test_setSampleRate")`

SCClassLibrary/Common/Control/Buffer.sc
elgiano: It's true that new, alloc and allocRead used to cache info before getting a reply from the server. It's also true that all the `read` methods send a "/query" msg at the end to update sclang. I don't know what would be best: to send an additional message (as I suggested), or to cache directly the value we reasonably assume the server would set.
In any case, for the scope of this PR, I think we should cache the assumed value in *new and *alloc (like they were doing before with `.sampleRate_(server.sampleRate)`, only this time it would be `if(value.isNil or: {value <= 0}) { server.sampleRate } { sampleRate }`)

Then for .setSampleRate we could stick to the same pattern.

tremblap:
ok I've cleaned everything (except the last fix @elgiano proposed) so we have a working state now. I'll implement it in the next hours

SCClassLibrary/Common/Control/Buffer.sc
tremblap: ok I think I did what you wanted, and to make it clear, the last commit is that only dif:

1. added check and cache in *new
2. added check and cache in *alloc
3. added check and cache in *allocconsecutive
4. removed the check and casting in allocMsg
5. added the check and cache in setSampleRate
6. did it too in setSampleRateMsg in case it is called directly

let me know what I did wrong.

tremblap:
> This is just a quick note, I can implement this all later

thanks @elgiano I think I did the latter in the last commit, and will observe how you propose to do the former (or is it just sending 0 as I was doing?)

In all cases, what seemed a simple PR untangled quite something. thanks for your support.

p

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -40,13 +42,14 @@ Buffer {
...
 		numChannels = numChannels.asInteger;
+		if(sampleRate.isNil or: {sampleRate <= 0}) { sampleRate = server.sampleRate };
 		^Array.fill(numBufs, { |i|
 			// note, cannot use alloc or allocMsg here
 			// because those methods don't pass a buffer index for completion message
 			var newBuf = Buffer.new(server, numFrames, numChannels, i + bufBase, sampleRate);
 			server.sendMsg(\b_alloc, i + bufBase, numFrames, numChannels,
 				completionMessage.value(newBuf, i),
-				sampleRate !? _.asFloat ?? 0.0);
+				sampleRate);
```
elgiano, line 52: we need sampleRate.asFloat here (for supernova, since we are sending the message directly)

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -70,7 +73,7 @@ Buffer {
 
 	allocMsg { arg completionMessage;
 		this.cache;
-		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, completionMessage.value(this), sampleRate !? _.asFloat ?? 0.0]
+		^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, completionMessage.value(this), sampleRate]
```
elgiano, line 76: we want to keep `sampleRate !? _.asFloat ?? 0.0` here, since allocMsg can be used directly, and nil.asFloat wouldn't work

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -600,13 +603,14 @@ Buffer {
...
+	if(value.isNil or: {value <= 0}) { sampleRate = server.sampleRate } {sampleRate = value.asFloat};
 		server.listSendMsg(this.setSampleRateMsg(value));
-		sampleRate = value !? _.asFloat;
     }
 
 	setSampleRateMsg { arg value;
 		if(bufnum.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
-		^[\b_setSampleRate, bufnum, value !? _.asFloat ?? 0.0]
+		if(value.isNil or: {value <= 0}) { sampleRate = server.sampleRate } {sampleRate = value.asFloat};
+		^[\b_setSampleRate, bufnum, value]
```
elgiano, line 613: we need to keep the check and the float conversion here, since this method can be called independently of setSampleRate
```supercollider
setSampleRateMsg { arg value;
	if(bufnum.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
	^[\b_setSampleRate, bufnum, value !? _.asFloat ?? 0.0]
}
```
About setting client-side sampleRate here, I'm thinking about it.
It looks like Buffer.sc is quite inconsistent in changing client-side members in methods that construct a message vs. their counterparts that actually send the message. Examples of caching only when sending are `alloc`, `allocConsecutive`, `cueSoundFile`, `write`; examples of caching when constructing the message are `readMsg`, `readChannelMsg`, `freeMsg`; there are even examples of caching in both methods: `allocRead`, `allocReadChannel`, `readChannel`...
My personal preference would be to do it only on .setSampleRate, when the message is sent, but we could also just do it here. I just wouldn't do it in both, since .setSampleRate calls .setSampleRateMsg. If we do it here, we need to send the convert value (sampleRate, not value).

```supercollider
setSampleRateMsg { arg value;
	if(bufnum.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
	if(value.isNil or: {value <= 0}) { sampleRate = server.sampleRate } {sampleRate = value.asFloat};
	^[\b_setSampleRate, bufnum, sampleRate]
}
```

elgiano:
@tremblap yeah, it's a complicated situation, it's also the first time an osc message gets extended, it seems. We'll need a bit more discussion on how to deal with an optional completionMsg followed by sampleRate, but I need more time to formulate it clearly.

SCClassLibrary/Common/Control/Buffer.sc
tremblap: I reckon the advantage of doing it in setSampleRateMsg is that it covers all grounds. I'd be happy with this.

elgiano:
For the first time we are dealing with an extra OSC argument after completionMsg, so we have to establish a sensible precedent and document it for external clients. Until now completionMsg has always been the last OSC argument of any message, so we didn't need to formalize a default value for it.

At the last dev meeting, we agreed to completely omit an empty completionMsg, and eventually directly supply sampleRate, but I don't think this is a good choice, because it's not what we do for any other optional argument. OSC args are positional, and I don't think it's a good idea to make these positions variable. I'll use \b_alloc before this PR as an example:
```
/b_alloc bufnum:int numFrames:int numChannels:int? completionMsg:bytes?
```
numChannels and completionMsg are optional, but if one wants to send a completionMsg, numChannels can't be omitted, and it has to be set to a default value. The only omission possible is if one wants to omit all optional arguments and truncate the message at numFrames.

With this PR, b_alloc becomes:
```
/b_alloc bufnum:int numFrames:int numChannels:int? completionMsg:bytes? sampleRate:float?
```
So if we want to supply sampleRate, we can't omit completionMsg, and we need to agree on a default value for it.

sclang sends `nil`, which is converted to int(0). I propose we formalize this and make GET_COMPLETION_MSG skip an int if it doesn't find a blob. 

This wouldn't impact existing clients, that are not sending int(0) to signify a null completionMsg, until they will want to implement the sampleRate argument. In that case they should learn what to send instead of an empty completionMsg, and I think int(0) is a good candidate, also because it's what sclang has always sent.

Pinging @scztt because we happened to talk through this at the last dev meeting, and @Spacechild1 for an extra eye.

Spacechild1:
> At the last dev meeting, we agreed to completely omit an empty completionMsg, and eventually directly supply sampleRate, 

We absolutely cannot do this!

> OSC args are positional, and I don't think it's a good idea to make these positions variable.

You are completely right. As I tried to point out several times in this PR, we can only ever *add* new arguments to existing methods/commands.

> sclang sends nil, which is converted to int(0). I propose we formalize this and make GET_COMPLETION_MSG skip an int if it doesn't find a blob.

I agree 100%. AFAICT that is the only sensible solution.

> This wouldn't impact existing clients, that are not sending int(0) to signify a null completionMsg, until they will want to implement the sampleRate argument.

Exactly!

tremblap:
ok I've done the setSampleRate streamlining and tested. I found another discrepancy in the \read code, this time between scsynth and supernova. I think it predates this PR and might be a 'feature'

```
Server.scsynth
s.options.sampleRate_(48000)
s.reboot
b = Buffer.new
b.read(Platform.resourceDir +/+ "sounds/a11wlk01.wav")
```

this doesn't read but doesn't moan. b.query is all 0s

in supernova (after Server.supernova and s.reboot) I get 'sample rate mismatch'

should we fix or should I make another bug report or is it normal behaviour?

SCClassLibrary/Common/Control/Buffer.sc
elgiano: I think this one slipped your last pushes. It should be:
```supercollider
^["/b_alloc", bufnum, numFrames.asInteger, numChannels.asInteger, completionMessage.value(this), sampleRate !? _.asFloat ?? 0.0]
```
we need that check because allocMsg can be used on its own, and then we need to make sure sampleRate is valid and float

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -595,6 +601,16 @@ Buffer {
...
 	}
 
+	setSampleRate { arg value;
+		server.listSendMsg(this.setSampleRateMsg(value));
+    }
+
+	setSampleRateMsg { arg value;
+		if(bufnum.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
+		if(value.isNil or: {value <= 0}) { sampleRate = server.sampleRate } {sampleRate = value.asFloat};
+		^[\b_setSampleRate, bufnum, value.asFloat];
```
elgiano, line 611: since we did all the checks and conversion on sampleRate, it would make sense to use it in the message:
```supercollider
^[\b_setSampleRate, bufnum, sampleRate];
```

elgiano (changes requested):

elgiano:
> I found another discrepancy in the \read code, this time between scsynth and supernova. I think it predates this PR and might be a 'feature'

This happens also on develop, so I don't think we should worry about it in this PR, which doesn't touch b_read at all. Note that if you would use .allocRead instead of .read, it would work as expected. It also works if you pre-allocate the buffer. What you are doing in the code you posted is reading a soundfile in a non-allocated buffer... I agree there could be some error message, but it's not related to this PR. You could open a new issue though!

SCClassLibrary/Common/Control/Buffer.sc
tremblap: done!

SCClassLibrary/Common/Control/Buffer.sc
elgiano: this still needs to be done, otherwise supernova complains for "wrong argument type"
```supercollider
server.sendMsg(\b_alloc, i + bufBase, numFrames, numChannels,
				completionMessage.value(newBuf, i),
				sampleRate.asFloat);
```

elgiano:
added a note in "Server Command Reference" about null completion message being represented by int 0.
Unless there are any opinions opposing this solution, I think this PR is good to go!

Spacechild1:
I just noticed an issue in `setSampleRateMsg`. The rest looks fine!

SCClassLibrary/Common/Control/Buffer.sc
```diff
@@ -595,6 +601,16 @@ Buffer {
 		^[\b_query, bufnum]
 	}
 
+	setSampleRate { arg value;
+		server.listSendMsg(this.setSampleRateMsg(value));
+    }
+
+	setSampleRateMsg { arg value;
+		if(bufnum.isNil) { Error("Cannot change the sample rate of a % that has been freed".format(this.class.name)).throw };
+		if(value.isNil or: {value <= 0}) { sampleRate = server.sampleRate } {sampleRate = value.asFloat};
```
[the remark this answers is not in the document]
Spacechild1: > which will be zero'd in the c++ code.

I think that's fine. In either case, it gets converted to the Server sample rate.

Spacechild1 (approved):

elgiano:
Thanks to everyone who worked on this! @tremblap @Spacechild1 @dyfer @scztt 
I think we can merge this for 3.14, I don't see any problem, and I think the decision we took for int(0) as a default empty completionMsg is sensible.

Spacechild1:
Thanks to everybody for your patience, in particular @tremblap :)

PR #6802, opened 2025-05-03
Add sclang testsuite cmake target

capital-G:
## Purpose and Motivation

This PR provides an easy way to run the testsuite locally using Cmake commands

* `cmake --build build --target sclang_lang_testsuite` (tests w/o server)
* `cmake --build build --target sclang_full_testsuite` (all tests)

This does not need any modifications or additional files on the system and uses the current state of the repo automatically.
This has became possible thanks to #6640 :)

Having the ability to run the test suite locally in an easy manner may allow to debug the test suite easier. The test suite still takes a considerable amount of time (~~15 minutes sclang only til an error appears, 45 minutes on full test suite until an error appears on my i9 mbp - maybe this is also due to using a Debug config?~~ => this is due to using a debug build), so maybe there should be a way to run only certain sections of the test-suite?

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing => not really, see below
- [ ] Updated documentation => Once merged I will add this to our dev wiki
- [ ] This PR is ready for review => This is probably not finished yet but I want some input from some more experienced users

Open questions:

* [ ] I need the plugins as a build dependency for the test suite - it seems we don't have a "build all plugins" target, so I added it manually, but this seems errorprone?
* [ ] When running the test suite and an error occurs the interpreter will not exit but stay in an indefinite state, making it necessary to kill the interpreter via `killall sclang`. Maybe we should add a flag when running a `scd` file which simply exits w/ error code `1` when the interpreter gets into the REPL state, making it mandatory for the script to exit?
* [ ] I never run the test suite locally before, but I run into some problems, especially the LinkClock seems to be troublesome and currently the test suite fails reliably @ https://gist.github.com/capital-G/b8225d98c240075d5182e2e38fcb7d72


dyfer:
@capital-G are you aware that you can run test suite by simply running [sclang_test_runner.scd](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/sclang_test_runner.scd) in the IDE? 

What's the benefit of having this as a cmake target, as opposed to a simple script that's run after a regular build?

dyfer:
> * When running the test suite and an error occurs the interpreter will not exit but stay in an indefinite state

I was thinking about this recently! Let's discuss this in #6807.

dyfer:
> * I never run the test suite locally before, but I run into some problems, especially the LinkClock seems to be troublesome and currently the test suite fails reliably

I was keeping track of failed tests in #5359, #5360, but I'm not sure how up to date that is.

capital-G:
> @capital-G are you aware that you can run test suite by simply running [sclang_test_runner.scd](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/sclang_test_runner.scd) in the IDE?
> 
> What's the benefit of having this as a cmake target, as opposed to a simple script that's run after a regular build?

For me: I don't have to spin up the SC IDE and have to create/modify the language config file - this is somehow cumbersome as I stay on latest SC stable, so I have to compile and boot up a Dev IDE to run the latest test suite, and the required language config file is incompatible with my stable IDE due to sclang feature mismatch between 3.13 and dev.

With this PR I can simply trigger a test run from within my development IDE (vscode/clion/...) or via CLI and being completely independent of any SC system installation.
As the whole test suite takes really long it isn't really useful yet (IME if tests take longer than 20 seconds to run, people don't run them), but this can be changed by cutting the test suite into smaller chunks or run them in parallel (should be fine for tests w/o server?).

capital-G:
I'll close this since the testsuite takes too long to run anyway.

JordanHendersonMusic:
How about a vscode task instead?

https://github.com/supercollider/supercollider/pull/7572

Issue #6560, opened 2024-12-02
Icons in settings menu are off


capital-G:
## Environment

* SuperCollider version: dev
* Operating system: macOS 15.0.1
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

Open settings menu.

## Expected vs. actual behavior

<img width="268" alt="grafik" src="https://github.com/user-attachments/assets/73047e50-27d3-43d4-add8-db28a74a3ee6">

Either

* no icons on any item
* icons on every item without overlapping

I tried to "fix" it by removing the `QIcon` at

https://github.com/supercollider/supercollider/blob/213a43a802bfd8b7a005e539cb4f39f16b5a23f0/editors/sc-ide/widgets/settings/dialog.cpp#L66

but the icon would still appear?

PR #6809, opened 2025-05-04
Do not display setting menu icons on macOS

capital-G:
## Purpose and Motivation

Closes https://github.com/supercollider/supercollider/issues/6560

Fixes a bug in the settings menu on macOS which was introduced by Qt6 which introduced QtIcons in some places on macOS.
As these icons seem to be not consistently available (I don't know where they are located and which are available) I decided to use conditions to only display icons in the settings menu if we are not on macOS, restoring the 3.13/qt5 layout for macOS.

w/o PR

grafik

w/ PR

grafik

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/settings/dialog.cpp
+++ b/editors/sc-ide/widgets/settings/dialog.cpp
@@ -38,19 +38,6 @@
 
 namespace ScIDE { namespace Settings {
 
-// Since qt6 some icons are matched on macOS and some are not.
-// As there doesn't seem to be a consistent icon set available on macOS
-// the icons are omitted for macOS.
-// See https://github.com/supercollider/supercollider/issues/6560
-QListWidgetItem* makeListWidgetItem(const QIcon& icon, const QString& text) {
-#ifdef Q_OS_MAC
-    return new QListWidgetItem(text);
-#else
-    return new QListWidgetItem(QIcon::fromTheme(icon), text);
-#endif
-}
-
-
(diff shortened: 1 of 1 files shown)


dyfer:
This doesn't currently build on Linux and Windows. 

I think we could do away with the icons on all systems, as on Windows they are also only shown for "Interpreter" and "Shortcuts", and even though there's no clash, it isn't very readable:
Screenshot 2025-05-04 at 8 51 50 PM

capital-G:
> I think we could do away with the icons on all systems, as on Windows they are also only shown for "Interpreter" and "Shortcuts", and even though there's no clash, it isn't very readable:

Deleted them now for every platform - seems easier to maintain.


> EDIT: why is there an icon for "Editor" on macOS, but not on Windows? And vice versa for the "Interpreter"? So weird...

From what I understood: On Linux Qt uses https://specifications.freedesktop.org/icon-naming-spec/latest/ - but these are not available by default on macOS or Windows.
Instead Qt seems to searches through some folders on the macOS/Windows for icons and if it matches one it will use it? It is unclear to me which locations these are. This could probably also be different for each machine.
Qt advises to ship icons with the application, but seems a bit much work.

dyfer (approved):
I think this is good to go and can be merged.
Please squash on merge.

PR #6810, opened 2025-05-04
plugins: TriggerUGens standardize init sample

elgiano:
## Purpose and Motivation

Add internal state resets for the sake of enforcing conventions on UGens initialization.

This PR is the first part of a larger effort to fix init sample behavior for Trigger UGens. This part doesn't change any init sample or output value, so no breaking change. UGens that require a fix that would change either init sample or output values are left for subsequent PRs.

When initializing a UGen, the convention is to reset internal state after the first calc function call in Ctor. UGens changed by this PR were already calculating init sample and managing internal state correctly, so adding an explicit reset was not functionally necessary. However, it makes it easier to reason about initialization, by aligning with the convention, and provides a better reference to be imitated when writing new UGens.

I made an exception for 3 UGens (see exceptions below), for which I didn't enforce the convention.

Note that this is in principle not just a refactor: the side effect is it forces y(0) recalculation after Ctor, which was not the case before. However, due to the simplicity of these UGens, I argue that the performance cost is negligible (I haven't measured it though).

### affected UGens
These UGens could pre-calculate y(0) in Ctor and wouldn't need resetting their internal state to output a valid y(0). Internal state reset was added only for the sake of adhering to the convention.

- SetResetFF
- Gate
- Schmidt
- PulseCount
- Stepper
- LastValue
- Latch: I was unsure if making an exception for Latch, but I decided not to.
- LeastChange
- MostChange

Note for MostChange and LeastChange: although this PR doesn't affect their output or init sample, they behave differently from what the docs say (see #6814).

### exceptions:
These UGens don't call _next in Ctor, and I thought it was for a good reason, so I made an exception and didn't enforce the init convention of calling _next and resetting internal state.

- ZeroCrossing: needs at least 4 samples to output a value
- RunningMin: correctly initialize output to first input value, no calc needed
- RunningMax: correctly initialize output to first input value, no calc needed

### unchanged UGens
These UGens were doing all right already:

- Trig1: correct y(0), resets
- Trig: correct y(0), resets
- FreeSelf: no output, resets
- PauseSelf: no output, resets (ok to pause twice)
- FreeSelfWhenDone: no internal state
- PauseSelfWhenDone: no internal state
- Done: correct y(0), nothing to reset
- SendTrig: no output, shouldn't call calc in Ctor
- SendReply: no output, shouldn't call calc in Ctor
- Poll: no output, shouldn't call calc in Ctor
- Pause: no calcFunc call, shouldn't call pause in Ctor
- Free: no calcFunc call, shouldn't call NodeEnd in Ctor
- SendPeakRMS: no output, no calcFunc call in Ctor

### need fix
The remaining UGens need a fix that would change their init sample or output values. I'll fix them in a subsequent PR.

- PulseDivider: FIX reset internal state, changes output
- ToggleFF: FIX (changes init sample)
- Timer: FIX (changes init sample)
- PeakFollower: FIX (changes init sample)
- Peak: FIX (changes init sample)
- TDelay: FIX (changes output: trig at y(0) is not delayed correctly)
- Sweep: FIX: reset unit->m_previn (changes output)
- Phasor: FIX (changes output)

## Types of changes

- New feature (refactoring init sample logic)

## To-do list

### DIFF
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -849,6 +849,10 @@ void SetResetFF_Ctor(SetResetFF* unit) {
     unit->mLevel = 0.f;
 
     SetResetFF_next_k(unit, 1);
+
+    unit->m_prevtrig = 0.f;
+    unit->m_prevreset = 0.f;
+    unit->mLevel = 0.f;
 }
 
 
@@ -964,7 +968,10 @@ void Latch_Ctor(Latch* unit) {
     unit->m_prevtrig = 0.f;
     unit->mLevel = 0.f;
 
-    ZOUT0(0) = ZIN0(1) > 0.f ? ZIN0(0) : 0.f;
+    Latch_next_ak(unit, 1);
+
+    unit->m_prevtrig = 0.f;
+    unit->mLevel = 0.f;
 }
 
 
@@ -1043,6 +1050,8 @@ void Gate_Ctor(Gate* unit) {
     unit->mLevel = 0.f;
 
     Gate_next_ak(unit, 1);
+
+    unit->mLevel = 0.f;
 }
 
 
@@ -1082,6 +1091,8 @@ void Schmidt_Ctor(Schmidt* unit) {
     unit->mLevel = 0.f;
 
     Schmidt_next(unit, 1);
+
+    unit->mLevel = 0.f;
 }
 
 void Schmidt_next(Schmidt* unit, int inNumSamples) {
@@ -1152,6 +1163,10 @@ void PulseCount_Ctor(PulseCount* unit) {
     unit->mLevel = 0.f;
 
     PulseCount_next_k(unit, 1);
+
+    unit->m_prevtrig = 0.f;
+    unit->m_prevreset = 0.f;
+    unit->mLevel = 0.f;
 }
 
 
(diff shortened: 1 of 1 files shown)


mtmccrea (approved):
Thanks! This all looks correct.

mtmccrea:
Not for this PR, but an idea to consider:

Since we have a canonical definition of a trigger in SC, could it be worth it to make this a function available either in the scope of TriggerUGens or across the full UGen suite?
```
if(prevtrig <= 0.f && curtrig > 0.f)
```
becomes
```
hasTriggered(prevtrig, curtrig)
```
The advantage would be consistency, readability, and slightly lighter cognitive load especially when parsing blocks like:
```
LOOP1(
        inNumSamples, float curtrig = ZXP(trig); if (prevreset <= 0.f && curreset > 0.f) level = 0.f;
        else if (prevtrig <= 0.f && curtrig > 0.f) { level += 1.f; } ZXP(out) = level; prevtrig = curtrig;
        prevreset = curreset;);
```

mtmccrea:
@elgiano would you mind migrating your overview of the problem and your proposal in the Description over to an Issue? That way the plan lives in one place that PRs can reference, others can comment on it, and we'll know the project (with the scope of TriggerUGens) is complete when the issue closes.

elgiano:
Thanks! I made #6813 to document and discuss the process.

While writing the next PR I found out that LeastChange doesn't need a fix, only a small refactor that doesn't change output or init sample. I'm adding it to the "exception" list, please let me know if it looks all right. MostChange, on the other hand, needs a little discussion and eventually a fix.

elgiano:
Sorry for the mess, after further research it turned out both LeastChange and MostChange can be standardized without changing their init sample or y(0) values. The problem with them is that they behave differently from what the docs say. See issue https://github.com/supercollider/supercollider/issues/6814

mtmccrea (approved):

PR #6812, opened 2025-05-04
Make doc themes persistent

capital-G:
## Purpose and Motivation

It seems the Qt6 change introduced a bug which made the local storage of the HelpBrowser not persistent.
This PR re-introduces this behavior by making the storage path explicit by providing a profile.
It also seems that changing the defaultProfile is not really working as the values were not modified when looked at them with a debugger - I therefore also removed the existing changes to it and included them in the new profile.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/QtCollider/interface.cpp
+++ b/QtCollider/interface.cpp
@@ -90,11 +90,6 @@ void QtCollider::init() {
 
         gSystemPalette = qcApp->palette();
 
-#ifdef SC_USE_QTWEBENGINE
-        // Enable javascript localStorage for WebViews
-        QWebEngineProfile::defaultProfile()->settings()->setAttribute(QWebEngineSettings::LocalStorageEnabled, true);
-#endif
-
         // NOTE: Qt may tamper with the C language locale, affecting POSIX number-string conversions.
         // Revert the locale to default:
         setlocale(LC_NUMERIC, "C");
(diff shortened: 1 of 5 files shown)


dyfer:
Thanks so much!

> It seems the Qt6 change introduced a bug which made the local storage of the HelpBrowser not persistent.

Yes, thank you! I remember that I was really unsure about this change - I didn't know what exactly was the functionality of if.

So... what is local storage for? How do I test whether this works now?

dyfer (approved):
Thanks! I tested on macOS and this works, solving the issue with a theme not persisting between IDE restarts.

For my limited understanding of Cpp and Qt this looks okay and I think it could be merged.

capital-G:
> For my limited understanding of Cpp and Qt this looks okay and I think it could be merged.

Says the person who ported everything to Qt6 :D

Maybe @Spacechild1 can chime in for some feedback on the C++ part? :)

dyfer:
BTW this PR also removes a console warning when running scide from the terminal and closing the app that I recently noticed:

```
Release of profile requested but WebEnginePage still not deleted. Expect troubles !
```

capital-G:
@Spacechild1 can you give this a brief review? If everything is fine from your side you can merge it.

Issue #6813, opened 2025-05-05
plugins: TriggerUGens initialization sample fix plan


elgiano:
This issue is meant to document and discuss the ongoing effort to fix TriggerUGens init sample calculation.

## part 1: standardize init sample calculation (#6810)
This part doesn't change any init sample or output value, so no breaking change. UGens that require a fix that would change either init sample or output values are left for subsequent PRs.

When initializing a UGen, the convention is to reset internal state after the first calc function call in Ctor. UGens changed by #6810 were already calculating init sample and managing internal state correctly, so adding an explicit reset was not functionally necessary. However, it makes it easier to reason about initialization, by aligning with the convention, and provides a better reference to be imitated when writing new UGens.

I made an exception for 3 UGens (see exceptions below), for which I didn't enforce the convention.

Note that this is in principle not just a refactor: the side effect is it forces y(0) recalculation after Ctor, which was not the case before. However, due to the simplicity of these UGens, I argue that the performance cost is negligible (I haven't measured it though).

### Affected UGens:
- SetResetFF
- Gate
- Schmidt
- PulseCount
- Stepper
- LastValue
- Latch: I was unsure if making an exception for Latch, but I decided not to.

### Exceptions:

These UGens don't call _next in Ctor, and I thought it was for a good reason, so I made an exception and didn't enforce the init convention of calling _next and resetting internal state.

- ZeroCrossing: needs at least 4 samples to output a value
- RunningMin: correctly initialize output to first input value, no calc needed
- RunningMax: correctly initialize output to first input value, no calc needed
- LeastChange: first sample output is fixed to IN0(0), no calc needed. Ctor had a small refactor to make this behavior clearer.

### Unchanged UGens:
These UGens were doing all right already:

- Trig1: correct y(0), resets
- Trig: correct y(0), resets
- FreeSelf: no output, resets
- PauseSelf: no output, resets (ok to pause twice)
- FreeSelfWhenDone: no internal state
- PauseSelfWhenDone: no internal state
- Done: correct y(0), nothing to reset
- SendTrig: no output, shouldn't call calc in Ctor
- SendReply: no output, shouldn't call calc in Ctor
- Poll: no output, shouldn't call calc in Ctor
- Pause: no calcFunc call, shouldn't call pause in Ctor
- Free: no calcFunc call, shouldn't call NodeEnd in Ctor
- SendPeakRMS: no output, no calcFunc call in Ctor

## part 2: fixes

[...]

PR #6816, opened 2025-05-06
plugins: TriggerUGens: fix initialization sample

elgiano:
## Purpose and Motivation

This is the second part of the implementation plan in #6813.

This PR fixes a batch of UGens, affecting their init sample or output sample in a limited way. Although this is a breaking change, I would argue that it's of the same entity as the ones introduced by #6035, and so I propose we merge it in SC 3.14, if there is no objection of course.

### affected UGens

- ToggleFF: changes init sample to 1 if UGen has trig on first input sample (x[0] from now on)
- Peak, PeakFollower: changes init sample for negative input on x[0]
- PulseDivider: changes first output sample for x[0] > 0 and div[0] = 1
- Timer: changes output correcting the first measured duration by one sample

The following code illustrates the kind of breaking changes this PR introduces
```supercollider
(
~printOutput = { |fn, nSamples(s.options.blockSize), action(_.postln)|
	fn.loadToFloatArray(nSamples/s.sampleRate, s, action)
};
~printOutputTrig = { |fn, nSamples=32|
	~printOutput.(fn, nSamples, {|v| v.asInteger.join.postln })
};
~printInitSample = { |ugen|
	~printOutput.({IEnvGen.ar(Env([ugen.value]),0)}, 1)
};
~printInitAndY0 = { |ugen|
	{
		var out = ugen.value;
		var initSample = IEnvGen.ar(Env([out]),0);
		[initSample, out]
	}.loadToFloatArray(1/s.sampleRate, action: {|v|
		"init = %\ty(0) = %".format(v[0], v[1]).postln
	})
}
)

s.boot;

// ToggleFF: changes init sample
~printInitAndY0.({ToggleFF.ar(1)})
// before: init = 0.0	y(0) = 1.0
// after:  init = 1.0	y(0) = 1.0

// PeakFollower: changes init sample
~printInitAndY0.({PeakFollower.ar(DC.ar(-1))},1)
// before: init = -1.0	y(0) = 1.0
// after:  init = 1.0	y(0) = 1.0

// Peak: changes init sample
~printInitAndY0.({Peak.ar(DC.ar(-1))})
// before: init = -1.0	y(0) = 1.0
// after:  init = 1.0	y(0) = 1.0


// PulseDivider: changes y[0] (used to ignore trig on in(0)[0])
~printInitAndY0.({PulseDivider.ar(DC.ar(1),DC.ar(1))})
// before: init = 1.0	y(0) = 0.0
// after:  init = 1.0	y(0) = 1.0
~printOutputTrig.({PulseDivider.ar(DC.ar(1), 1)})
// before: 00000000000000000000000000000000
// after:  10000000000000000000000000000000
~printOutputTrig.({PulseDivider.ar(Impulse.ar(SampleRate.ir/2), 1)})
// before: 00101010101010101010101010101010
// after:  10101010101010101010101010101010
~printOutputTrig.({PulseDivider.ar(Impulse.ar(SampleRate.ir/2), 2)})
// before: 00100010001000100010001000100010
// after:  00100010001000100010001000100010

// Timer: changes output (for first measured duration)
~printInitAndY0.({Timer.ar(Impulse.ar(SampleRate.ir/2))})
// after:  init = 0.0	y(0) = 0.0
~printOutput.({Timer.ar(Impulse.ar(SampleRate.ir/2))})
// dividing output values by sampleDur for clarity
// before: 0, 0, 3, 3, 2, 2, 2, 2
// after:  0, 0, 2, 2, 2, 2, 2, 2
~printOutput.({
	var dt = SampleDur.ir * 4;
	Timer.ar(DelayN.ar(Impulse.ar(SampleRate.ir/2), dt, dt))
})
// before: 0, 0, 0, 0, 0.5, 0.5, 2, 2, 2, 2
// before: 0, 0, 0, 0, 4, 4, 2, 2, 2, 2
```

I wonder if we should turn these code snippets into tests. Let me know what you think!

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested (no new tests so far)
- [x] All tests are passing (already present tests involving affected UGens are passing)

### DIFF
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -899,8 +899,9 @@ void ToggleFF_Ctor(ToggleFF* unit) {
 
     unit->m_prevtrig = 0.f;
     unit->mLevel = 0.f;
-
-    ZOUT0(0) = 0.f;
+    ToggleFF_next(unit, 1);
+    unit->m_prevtrig = 0.f;
+    unit->mLevel = 0.f;
 }
 
 
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -1664,7 +1664,10 @@ void Peak_Ctor(Peak* unit) {
         }
     }
     unit->m_prevtrig = 0.f;
-    ZOUT0(0) = unit->mLevel = ZIN0(0);
+    float initLevel = unit->mLevel = std::abs(ZIN0(0));
+    Peak_next_ai(unit, 1);
+    unit->m_prevtrig = 0.f;
+    unit->mLevel = initLevel;
 }
 
 void Peak_next_ak(Peak* unit, int inNumSamples) {
@@ -1938,7 +1941,9 @@ void PeakFollower_Ctor(PeakFollower* unit) {
     }
 
     unit->mDecay = ZIN0(1);
-    ZOUT0(0) = unit->mLevel = ZIN0(0);
+    float initLevel = unit->mLevel = std::abs(ZIN0(0));
+    PeakFollower_next(unit, 1);
+    unit->mLevel = initLevel;
 }
 
 void PeakFollower_next(PeakFollower* unit, int inNumSamples) {
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -118,12 +118,12 @@ struct TDelay : public Unit {
 
 struct ZeroCrossing : public Unit {
     float mLevel, m_prevfrac, m_previn;
-    int32 mCounter;
+    long mCounter;
 };
 
 struct Timer : public Unit {
     float mLevel, m_prevfrac, m_previn;
-    int32 mCounter;
+    long mCounter;
 };
 
 struct Sweep : public Unit {
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -1109,9 +1109,13 @@ void PulseDivider_Ctor(PulseDivider* unit) {
 
     unit->m_prevtrig = 0.f;
     unit->mLevel = 0.f;
-    unit->mCounter = (long)floor(ZIN0(2) + 0.5);
+    long initCounter = unit->mCounter = static_cast<long>(floor(ZIN0(2) + 0.5));
 
     PulseDivider_next(unit, 1);
+
+    unit->m_prevtrig = 0.f;
+    unit->mLevel = 0.f;
+    unit->mCounter = initCounter;
 }
 
 
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -1377,9 +1377,14 @@ void Timer_Ctor(Timer* unit) {
     SETCALC(Timer_next_a);
 
     unit->m_prevfrac = 0.f;
-    unit->m_previn = ZIN0(0);
-    ZOUT0(0) = unit->mLevel = 0.f;
-    unit->mCounter = 0;
+    unit->m_previn = 0;
+    unit->mCounter = -1;
+    unit->mLevel = 0;
+    Timer_next_a(unit, 1);
+    unit->m_prevfrac = 0.f;
+    unit->m_previn = 0;
+    unit->mCounter = -1;
+    unit->mLevel = 0;
 }
 
 void Timer_next_a(Timer* unit, int inNumSamples) {


mtmccrea:
Thanks for the PR @elgiano, I'll try to have a look in the coming days.

Regarding tests, yes, I think we should have unit tests for all UGens whose functionality has been corrected. Ideally of course all UGens would have tests, so if it's straightforward to include other related UGens in any tests you make, even better 🙂. 

WRT the init sample testing method, I'm not sure if that should be formalized into a unit test. It cleverly uses IEnvGen, for the reason you noted that it passes the init sample through to the output, but that isn't super obvious when inspecting the test. Perhaps others have thoughts on that...

elgiano:
Thanks! No rush of course!

Regarding tests: i wouldn't want to rely on IEnvGen, it's a hack exploiting an implementation detail, who knows perhaps its implementation changes in the future and all these tests would possibly fail...

I see two options:
- we test IEnvGen first, as part of testing init samples
- better: we write a simple UGen just for this purpose

In the second case I would add it to implementation plan and do it in another PR

mtmccrea:
I think a UGen that outputs its initialization sample is a nice idea and could be useful for debugging and unit testing!

As it stands now, no unit tests cover initialization sample issues, so that would be a new suite/category of testing.

At a minimum, we should aim to have the first "startup samples" tested for UGens, 
so we know it outputs what is expected, which is often discernible from the first few samples/blocks. Historically this has meant testing the first x samples against a known output (e.g. a known series of impulses or starting value of an oscillator with a particular phase offset, etc.), and/or testing against UGens we know should be equivalent (like Line and a linear segment of an EnvGen). For triggers, comparing the output to an expected series of triggers sounds the most simple and direct.

Thanks for thinking of tests! I was going to bring this up in the general Issue discussion, but you're already on it 🙃. A lot of the work of the previous init sample project became writing unit tests—you can see how the tasks begin to sprawl once you look closely at UGens 😆. I think it's tedious but important work–you could imagine how many headaches and time would have been saved if UGens had test coverage two decades ago...

elgiano:
we discussed this at the dev meeting:
- bug fixes with small breaking changes like these can be merged in 3.14
- it would be nice to have a Changelog for these kind of changes as a .schelp

I don't think we need to make this Changelog now, and also note that each commit description of this PR has a description of what changed, so that it can be picked up eventually later.

mtmccrea:
This isn't directly related to the PR, but as I was reasoning through `Peak`, I was confused by the description in the Help file:
>Internally, the absolute value of the signal is used, to prevent underreporting the peak value if there is a negative DC offset.

I _think_ the intention is just to say that _The reported peak will always be positive, i.e. the absolute value of the signal at it's peak amplitude, even when that peak is negative._ (or similar)

If so, then that's not the same as what the current doc implies. Should we update that here?

mtmccrea:
Also not directly related, but important:
There's an implicit conversion of `long` to `int` in `next_` functions of both `ZeroCrossing` and `Timer`.

Can we change the type of `mCounter` in both structs to be `long`?

https://github.com/supercollider/supercollider/blob/6c04748ec3ab49aec0163310f07efbf29febf3e5/server/plugins/TriggerUGens.cpp#L119-L127

mtmccrea:
I looked through the changes and they all look good to me!

Thanks for the helpful examples with output in the Description. Do you think you would be able to turn these into unit tests?
I think in the cases where you're testing multiple samples of output, it looks like 8 or 10 samples of output is enough to know if it's behaving properly.

Also, did you test at kr?

Here's an example of how ar and kr tests can be made more compact by just looping over the rate:
https://github.com/supercollider/supercollider/blob/83591f4ad575fd9f986ad2224a3cc6a651d18834/testsuite/classlibrary/TestCoreUGens.sc#L994-L1018

mtmccrea (commented):
I looked through the changes and they all look good to me!

Thanks for the helpful examples with output in the Description. Do you think you would be able to turn these into unit tests?
I think in the cases where you're testing multiple samples of output, it looks like 8 or 10 samples of output is enough to know if it's behaving properly.

Also, did you test at kr?

Here's an example of how ar and kr tests can be made more compact by just looping over the rate:
https://github.com/supercollider/supercollider/blob/83591f4ad575fd9f986ad2224a3cc6a651d18834/testsuite/classlibrary/TestCoreUGens.sc#L994-L1018

elgiano:
Thanks a lot for your review!
About tests: yes, I'm working on a PR with all these tests already (and the InitializationSample UGen). My syntax is a little bit different, but similar to your suggestions (also testing kr), so I'll compare them and see what fits best <3 However, I think it's a longer discussion, in a way we might want to test all rate combinations... but let's keep that discussion for the "tests" PR.

I think it's no problem to remove implicit conversions for Timer and ZeroCrossing, I'll push an additional commit. For Peak documentation, would you mind opening an extra PR or issue?

Thanks again for all the work!

jamshark70:
> I think the intention is just to say that "The reported peak will always be positive, i.e. the absolute value of the signal at its peak amplitude, even when that peak is negative." (or similar)
> 
> If so, then that's not the same as what the current doc implies.

The original is probably my wording -- I'm totally OK with simplifying it.

It *is* true that `SinOsc.ar(440) - 0.25` has a negative peak that is larger than the positive peak -- but it's also true that `SinOsc.ar((SinOsc.ar(440) + 1) * 440)` has peaks at +-1 *and* a negative DC offset -- so it would be correct to take DC offset out of this description.

mtmccrea (approved):
@elgiano, I'm approving this now so it makes the 3.14 deadline and since the only outstanding issue is unit tests, which you have coming in another PR (let's be sure to include those in the near future if possible :)).

elgiano:
Sure, I'll push the tests PR now, but we'll need also to fix Phasor and Sweep for all the tests to pass.

Thanks a lot!!

PR #6817, opened 2025-05-06
IDE: fix sclang config file dialog behavior

### LINKED ISSUE #6692: IDE: The list of Include/Exclude paths doesn't update when switching language config files
## Environment

* Operating system: macOS
* Other details (Qt version, audio driver, etc.): recent develop build from the CI; Qt 6

## Steps to reproduce

1. Open IDE
2. Open Preferences -> Interpreter
3. Create multiple config files with different include/exclude paths
4. Switch from one config file to another using the dropdown menu

Expected: the list of Include and Exclude paths updates immediately upon choosing the new config file

Actual: neither list doesn't immediately update; after reopening Preferences, the correct/updated lists are displayed though.

**This seems to be a regression from SC 3.13**. In 3.13 the list of paths updates immediately after selecting the different config file.

----
## Note:

This has nothing to do with the requirement to reboot the language after selecting the new config file. This is just about the immediate update of the gui when switching between config files.

Also, when I was searching whether this was already reported, I found https://github.com/supercollider/supercollider/issues/3530. It's not related, but a fix to both issues may be touching the similar code, I think.

dyfer:
## Purpose and Motivation

Fixes #6692

Apparently `currentIndexChanged(const QString&)` signal has been deprecated / removed and instead we should use `currentTextChanged(const QString&)`.

This also fixes theme/font dialogs, I believe.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing (the failing tests are unrelated to this PR...)
- ~~[n/a] Updated documentation~~

### DIFF
--- a/editors/sc-ide/widgets/settings/editor_page.cpp
+++ b/editors/sc-ide/widgets/settings/editor_page.cpp
@@ -46,11 +46,11 @@ EditorPage::EditorPage(QWidget* parent):
     connect(ui->tabs, SIGNAL(currentChanged(int)), this, SLOT(onCurrentTabChanged(int)));
 
     connect(ui->onlyMonoFonts, SIGNAL(toggled(bool)), this, SLOT(onMonospaceToggle(bool)));
-    connect(ui->fontCombo, SIGNAL(currentTextChanged(QString)), this, SLOT(updateFontPreview()));
+    connect(ui->fontCombo, &QComboBox::currentTextChanged, this, &EditorPage::updateFontPreview);
     connect(ui->fontSize, SIGNAL(valueChanged(int)), this, SLOT(updateFontPreview()));
     connect(ui->fontAntialias, SIGNAL(stateChanged(int)), this, SLOT(updateFontPreview()));
 
-    connect(ui->themeCombo, SIGNAL(currentTextChanged(QString)), this, SLOT(updateTheme(QString)));
+    connect(ui->themeCombo, &QComboBox::currentTextChanged, this, &EditorPage::updateTheme);
     connect(ui->themeCopyBtn, SIGNAL(clicked()), this, SLOT(dialogCopyTheme()));
     connect(ui->themeDeleteBtn, SIGNAL(clicked()), this, SLOT(deleteTheme()));
     connect(ui->textFormats, SIGNAL(currentItemChanged(QTreeWidgetItem*, QTreeWidgetItem*)), this,
@@ -227,7 +227,7 @@ void EditorPage::populateFontList(bool onlyMonospaced) {
 
 void EditorPage::populateThemeList(const QString& sel) {
     /* managing the combo box send parasite signals */
-    disconnect(ui->themeCombo, SIGNAL(currentTextChanged(QString)), this, SLOT(updateTheme(QString)));
+    disconnect(ui->themeCombo, &QComboBox::currentTextChanged, this, &EditorPage::updateTheme);
 
     QMap<QString, Theme*>::const_iterator itr = mThemes.begin();
     QList<QString> list = itr.value()->availableThemes();
@@ -247,7 +247,7 @@ void EditorPage::populateThemeList(const QString& sel) {
             i++;
     }
 
-    connect(ui->themeCombo, SIGNAL(currentTextChanged(QString)), this, SLOT(updateTheme(QString)));
+    connect(ui->themeCombo, &QComboBox::currentTextChanged, this, &EditorPage::updateTheme);
 }
 
 void EditorPage::store(Manager* s) {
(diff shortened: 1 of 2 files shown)


capital-G (approved):
Thx - actually [`currentIndexChanged`](https://doc.qt.io/qt-6/qcombobox.html#currentIndexChanged) is not deprecated, the problem is that this signal passes an `int` to the callback while our callback expects a QString - instead of crashing it just seem to do nothing - maybe this behavior changed with Qt6?

I am a bit surprised [that we don't get a compiler warning](https://github.com/supercollider/supercollider/actions/runs/14841698308/job/41666034716#step:14:12193) for such behavior, but maybe I don't get something about Qt/C++, because it would be good to know which signals are broken during compile time.

Maybe @Spacechild1 has also some experience with Qt and give us a clue here? :)

Otherwise this looks fine to me, though @Spacechild1 should merge it - maybe he knows a way to make this into compilation warnings.

dyfer:
> actually [`currentIndexChanged`](https://doc.qt.io/qt-6/qcombobox.html#currentIndexChanged) is not deprecated, the problem is that this signal passes an `int` to the callback while our callback expects a QString

IIUC `currentIndexChanged(QString)` IS removed (has been deprecated), but `currentIndexChanged(int)` is not, that's why I listed it with the types. Is it not so?

I'm also surprised this wasn't picked up by the compiler

dyfer:
> it seems possible to do such checks at compile time w/ the "new" signal syntax

Wow, a new syntax for Qt**5**, we're a bit behind on adopting this...

JordanHendersonMusic:
> There is no automatic disconnection when the 'receiver' is destroyed because it's a functor with no QObject. However, since 5.2 there is an overload which adds a "context object". When that object is destroyed, the connection is broken (the context is also used for the thread affinity: the lambda will be called in the thread of the event loop of the object used as context).

From the qt doc, as far as I can see this is the only potential drawback and will need to be handled carefully as it can/will lead to a seg fault.

Spacechild1:
> Maybe we should adapt this syntax? Compile time checks are a nice thing and we should make use of them :)

Yes! Personally, I have actually never used the old syntax :)

dyfer:
How does this look now?

capital-G (approved):
Looks right to me, especially the lifetime object seems to be passed properly (as before).

In gereneral it seems we should replace all "old-style" connections with their type-safe equivalent https://doc.qt.io/qt-6/qobject.html#connect-4 - plus-points for using lambda functions which would probably make all our interfaces a lot smaller and clearer.

capital-G:
I just compiled this and it indeed fixes #6692 

Though maybe it would be nice to also display this dialog as well when switching config files while we are at it? Or should this be another PR?

grafik

https://github.com/supercollider/supercollider/blob/92ad1c94353e97acab079aa9a7957fb7ce07d814/editors/sc-ide/widgets/settings/sclang_page.cpp#L234-L236

It currently is only displayed when e.g. adding/deleting paths.

capital-G:
It seems something like this could implement this? At least it worked for me on a quick test.

```patch
diff --git a/editors/sc-ide/widgets/settings/sclang_page.cpp b/editors/sc-ide/widgets/settings/sclang_page.cpp
index 2aef242219..d50144d787 100644
--- a/editors/sc-ide/widgets/settings/sclang_page.cpp
+++ b/editors/sc-ide/widgets/settings/sclang_page.cpp
@@ -51,6 +51,7 @@ SclangPage::SclangPage(QWidget* parent): QWidget(parent), ui(new Ui::SclangConfi
 
     connect(ui->activeConfigFileComboBox, &QComboBox::currentTextChanged, this,
             &SclangPage::changeSelectedLanguageConfig);
+    connect(ui->activeConfigFileComboBox, &QComboBox::currentTextChanged, this, &SclangPage::markSclangConfigDirty);
 
     connect(ui->sclang_add_configfile, SIGNAL(clicked()), this, SLOT(dialogCreateNewConfigFile()));
     connect(ui->sclang_remove_configfile, SIGNAL(clicked()), this, SLOT(dialogDeleteCurrentConfigFile()));
@@ -68,6 +69,8 @@ SclangPage::SclangPage(QWidget* parent): QWidget(parent), ui(new Ui::SclangConfi
 SclangPage::~SclangPage() { delete ui; }
 
 void SclangPage::load(Manager* s) {
+    // disconnect while loading init settings
+    disconnect(ui->activeConfigFileComboBox, &QComboBox::currentTextChanged, this, &SclangPage::markSclangConfigDirty);
     s->beginGroup("IDE/interpreter");
 
     ui->autoStart->setChecked(s->value("autoStart").toBool());
@@ -87,6 +90,8 @@ void SclangPage::load(Manager* s) {
     s->endGroup();
 
     readLanguageConfig();
+    // reconnect again
+    connect(ui->activeConfigFileComboBox, &QComboBox::currentTextChanged, this, &SclangPage::markSclangConfigDirty);
 }
```

PR #6818, opened 2025-05-06
Add dialog to restart interpreter after switching config file

### LINKED ISSUE #3530: Preferences panel doesn't warn to reboot language when switching language config
1. Have at least 2 sclang_conf_**.yaml files.

2. Launch the IDE and go to Edit > Preferences > Interpreter.

3. Choose a different config file.

For all other changes to the language config, the preferences panel tells the user to reboot the interpreter. In this case, a reboot is required (I tested -- the changes do not take effect if you switch config without rebooting the language) but there is no warning.

dyfer:
## Purpose and Motivation

Fixes #3530

I moved the dialog to a separate function and added another flag to keep track of changed config file, without marking the file "dirty" (i.e. needing to save it). Please review the logic carefully.

~~Note: I'll rebase this after #6817 is in - I needed to include the commit from the other PR in order to test this functionality.~~ The PR is now rebased.

## Types of changes

- Bug fix
- New feature (?)

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/settings/sclang_page.cpp
+++ b/editors/sc-ide/widgets/settings/sclang_page.cpp
@@ -86,6 +86,8 @@ void SclangPage::load(Manager* s) {
 
     s->endGroup();
 
+    sclangConfigChanged = false;
+
     readLanguageConfig();
 }
 
@@ -131,6 +133,7 @@ void SclangPage::removeExcludePath() {
 
 void SclangPage::changeSelectedLanguageConfig(const QString& configPath) {
     selectedLanguageConfigFile = configPath;
+    sclangConfigChanged = true;
     readLanguageConfig();
 }
 
@@ -196,8 +199,13 @@ void SclangPage::readLanguageConfig() {
 }
 
 void SclangPage::writeLanguageConfig() {
-    if (!sclangConfigDirty)
+    if (!sclangConfigDirty) {
+        if (sclangConfigChanged) {
+            dialogConfigFileUpdated();
+            sclangConfigChanged = false; // avoid double invocation
+        }
         return;
+    }
 
     using namespace YAML;
     using std::ofstream;
@@ -232,9 +240,7 @@ void SclangPage::writeLanguageConfig() {
     ofstream fout(languageConfigFile().toStdString().c_str());
     fout << out.c_str();
 
-    QMessageBox::information(this, tr("Sclang configuration file updated"),
-                             tr("The SuperCollider language configuration has been updated. "
-                                "Reboot the interpreter to apply the changes."));
+    dialogConfigFileUpdated();
 
     sclangConfigDirty = false;
 }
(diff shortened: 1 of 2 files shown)


capital-G:
I tried something similar yesterday for the theme switcher, but I found that multiple signals are firing all over the place during setup procedure of the settings dialog (e.g. first setting to default settings, afterwards the currently applied settings, which one could confuse for that something changed while Qt/IDE itself fired the change signal) - this should be carefully inspected with a debugger, will do this after the merge/rebase.

capital-G:
See https://github.com/supercollider/supercollider/pull/6817#issuecomment-2857740070 for alternative implementation, but yours is probably less hacky.

Should adapt the new Qt connect syntax - maybe this could be rewritten as a lambda method?

dyfer:
> See [#6817 (comment)](https://github.com/supercollider/supercollider/pull/6817#issuecomment-2857740070) for alternative implementation, but yours is probably less hacky.

I introduced a separate flag to avoid marking the config file "dirty", which would trigger rewriting it's contents to disk on close when we did not change anything, but only selected a new config file. Please let me know if you think this approach can be improved.

editors/sc-ide/widgets/settings/sclang_page.cpp
```diff
@@ -196,8 +199,13 @@ void SclangPage::readLanguageConfig() {
 }
 
 void SclangPage::writeLanguageConfig() {
-    if (!sclangConfigDirty)
+    if (!sclangConfigDirty) {
+        if (sclangConfigChanged) {
+            dialogConfigFileUpdated();
+            sclangConfigChanged = false; // avoid double invocation
```
capital-G, line 205: how could this trigger a double invocation? Can `writeLanguageConfig` be invoked multiple times?

capital-G (commented):
Thanks for this feature! I have one question, but in general this seems fine to be merged.

I first thought this could be written somehow easier/clearer but due to the loading of the config values through a signal instead during initialization we need to separate those events from user events which makes the code a bit more convoluted IMO.

A third PR could be to add a custom button to the dialog which allows to send a reboot signal ;)

editors/sc-ide/widgets/settings/sclang_page.cpp
dyfer: I don't know. But I observed double invocation. And also the flag should be reset anyway, no?

dyfer:
> I first thought this could be written somehow easier/clearer but due to the loading of the config values through a signal instead during initialization we need to separate those events from user events which makes the code a bit more convoluted IMO.

Yeah, I haven't re-thought the whole logic of it. Do you think we should address it here?



> A third PR could be to add a custom button to the dialog which allows to send a reboot signal ;)

Oh yes, I like this idea! (for a separate PR)

editors/sc-ide/widgets/settings/sclang_page.cpp
capital-G: Oh right - we have an *Apply* button in which case it should be indeed retested as the widget would not be destroyed in this case.

capital-G:
> Yeah, I haven't re-thought the whole logic of it. Do you think we should address it here?

No, I don't think we should fix it in this particular place - I think this is a general theme of the IDE. I don't know enough Qt to understand if there is actually a good reason to it this way.

> Oh yes, I like this idea! (for a separate PR)

<3

capital-G (approved):

PR #6822, opened 2025-05-07
plugins: Sweep: fix reset on init sample and kk/ka calc

### LINKED ISSUE #6541: Sweep starts at a sample count of 1 when beeing triggered by Impulse.ar(0)
## Environment

## Steps to reproduce
```
(
{ |t_trig = 1, duration = 0.02|
	
	var initTrigA = Impulse.ar(0);
	var oneShotPhaseA = Sweep.ar(initTrigA, SampleRate.ir) * (1 / duration * SampleDur.ir) * (PulseCount.ar(initTrigA) > 0);

	var initTrigB = Trig1.ar(t_trig, SampleDur.ir);
	var oneShotPhaseB = Sweep.ar(initTrigB, SampleRate.ir) * (1 / duration * SampleDur.ir) * (PulseCount.ar(initTrigB) > 0);
	
	[oneShotPhaseA, oneShotPhaseB];

}.plot(0.00041);
)
```
grafik

Sweep should always start from a sample count of 0.

elgiano:
## Purpose and Motivation

Fixes #6541.

Sweep init sample calculation was ignoring triggers on first sample of input. This fixes it, making Sweep.ar correctly output 0 for its first sample if triggered.

While at it, I found that when using the `kk` or `ka` calcFunctions (i.e. when first input is control-rate), the UGen would reset to 0 on trigger, but it would also immediately apply the increment and thus it would never write that 0 to output. I have fixed it, note that I've switched from LOOP1 to LooP because now inNumSamples can be 0.

This PR is part of #6813, and introduces small bug-fixing breaking changes in Sweep output and init sample, which are in the order of a 1-sample delay or 1 sampleDur order of magnitude. Here is some code to show what changes:

```supercollider
// AA/AK cases
~printInitAndY0.({Sweep.ar(Impulse.ar(0)) / SampleDur.ir})
// before: init = 0.0	y(0) = 1.0
// after:  init = 0.0	y(0) = -0.0
// KA/KK cases
~printInitAndY0.({Sweep.kr(Impulse.kr(0)) / ControlDur.ir})
// after:  init = 0.0	y(0) = -0.0
```

Note that there is still one issue that this PR doesn't fix. When not triggered at init/first sample time, Sweep still starts at sample count of 1, instead of 0. This is because of how the calculation is implemented: if not triggered, it first increments the counter and then writes to output. This is reasonable and accurate for all cases except y[0], so I'd like to discuss how to fix this too.
Examples:

```supercollider

~printInitAndY0.({Sweep.ar(DC.ar(0)) / SampleDur.ir})
// before: init = 0.0	y(0) = 1.0
// after:  init = 1.0	y(0) = 1.0
~printInitAndY0.({Sweep.kr(DC.kr(0)) / ControlDur.ir}, 64)
// before: init = 0.0	y(0) = 1.0
// after:  init = 1.0	y(0) = 1.0
```

Note that although it doesn't fix it, at least this PR makes the behavior consistent across input rates, and makes the init sample equal to y[0].

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing

Note about tests: I haven't written new tests for this, we're discussing testing init and y[0] values in #6813


server/plugins/TriggerUGens.cpp
```diff
@@ -1459,9 +1462,14 @@ void Sweep_next_kk(Sweep* unit, int inNumSamples) {
     if (previn <= 0.f && curin > 0.f) {
         float frac = -previn / (curin - previn);
```
mtmccrea, line 1463: I think this fraction calculation may be wrong.

Assume
```
previn = -0.25;
curin = 0.5;
```
The trigger happens at
```
-(-0.25) / (0.5 - -0.25)  = 1/3
```
i.e 1/3 of the way through the frame. So the resulting Sweep value should be `2/3 * rate`.

So this currently gives the "wrong side" of the increment ratio.

server/plugins/TriggerUGens.cpp
```diff
@@ -1459,9 +1462,14 @@ void Sweep_next_kk(Sweep* unit, int inNumSamples) {
     if (previn <= 0.f && curin > 0.f) {
```
mtmccrea, line 1462: What do you think about renaming the variable `rate` to `step` or `stepsize` in cases where the `rate` arg is control rate. After the mul by `SAMPLEDUR`, that value is not longer a "rate". It may even read better if all instances of `rate` are called `step`.

mtmccrea (commented):
I think it would be good to address the pre-increment behavior, which I'd consider a bug and in scope of the init sample fix. So we'd increment `level` by `rate` only after writing out `level`, as is done in `Phasor`:
https://github.com/supercollider/supercollider/blob/7c6e99019d66f244ac871d75ddcaf900f6215c2e/server/plugins/TriggerUGens.cpp#L1594

server/plugins/TriggerUGens.cpp
mtmccrea: In fact this bug was already caught in `Phasor`:
https://github.com/supercollider/supercollider/blame/7c6e99019d66f244ac871d75ddcaf900f6215c2e/server/plugins/TriggerUGens.cpp#L1614

So I suppose this `frac` calculation should  be checked throughout the whole source file (`Timer`, `ZeroCrossing`), and in UGens outside this source file. Something for another PR I suppose...

mtmccrea:
What do you think about renaming the variable `rate` to `step` or `stepsize` in cases where the `rate` arg is control rate?

After the mul by `SAMPLEDUR`, that value is not longer a "rate". It may even read better if all instances of `rate` are called `step`.

server/plugins/TriggerUGens.cpp
mtmccrea: Filed an Issue, #6883,  so I'll leave it up to you whether you'd like to include this fix or save that for another PR.

elgiano: It was caught in Phasor, but the fix is also still wrong.. it should be `1 - (-previn/(curin-previn))`

dyfer:
@elgiano @mtmccrea 
We are trying to release 3.14 in the next few days, would it be possible to wrap up this PR?

elgiano:
It turns out that the init sample fix was mostly for adhering to convention. So there is no rush for merging this if we don't also fix the behavior on reset (and it wouldn't fix the linked issue either).

After all the discussion, I think a better fix for the reset behavior is to apply rate after writing to output. I added a commit for this. The concern I expressed above instead (what rate to apply when doing sub-sample interpolation) shouldn't be a concern here.

Note also that here subsample interpolation works with Impulse already, but only because it's using the wrong side of the offset. So, by mistake, we are actually exposing the expected behavior for impulses.

elgiano:
@mtmccrea 
> Tested and confirmed that with input and output rates matching, the init/first samples match, and the reset behavior is correct. And I agree with the approach to incrementing here.

I've tested that again, and actually it seems to work for all input/output rates combinations, except one:
```supercollider
// outputs all zeros
~printOutputTrig.({Sweep.ar(Impulse.kr(ControlRate.ir/2), DC.ar(1))/SampleDur.ir}, 129)
```
That is: Sweep output is audio-rate, trig is control-rate, but rate is audio-rate. It fails because of aliasing, I've fixed it now by storing `step` before writing to output.

Here are all the tests:
```supercollider
// init:
// all of these print init = -0.0	y(0) = -0.0
~printInitAndY0.({Sweep.ar(Impulse.ar(0), DC.ar(1)) / SampleDur.ir})
~printInitAndY0.({Sweep.ar(Impulse.ar(0), DC.kr(1)) / SampleDur.ir})
~printInitAndY0.({Sweep.ar(Impulse.kr(0), DC.ar(1)) / SampleDur.ir})
~printInitAndY0.({Sweep.ar(Impulse.kr(0), DC.kr(1)) / SampleDur.ir})
~printInitAndY0.({Sweep.kr(Impulse.ar(0), DC.ar(1)) / SampleDur.ir})
~printInitAndY0.({Sweep.kr(Impulse.ar(0), DC.kr(1)) / SampleDur.ir})
~printInitAndY0.({Sweep.kr(Impulse.kr(0), DC.ar(1)) / SampleDur.ir})
~printInitAndY0.({Sweep.kr(Impulse.kr(0), DC.kr(1)) / SampleDur.ir})

// reset:
// all of these start and reset to 0
// kr: KK
~printOutputTrig.({Sweep.kr(Impulse.kr(ControlRate.ir/4), DC.kr(1))/ControlDur.ir}, 64*16)
// kr: KA
~printOutputTrig.({Sweep.kr(Impulse.kr(ControlRate.ir/4), DC.ar(1))/ControlDur.ir}, 64*16)
// kr: AK
~printOutputTrig.({Sweep.kr(Impulse.ar(ControlRate.ir/4), DC.kr(1))/ControlDur.ir}, 64*16)
// kr: AA
~printOutputTrig.({Sweep.kr(Impulse.ar(ControlRate.ir/4), DC.ar(1))/ControlDur.ir}, 64*16)
// ar: AK
~printOutputTrig.({Sweep.ar(Impulse.ar(SampleRate.ir/4))/SampleDur.ir})
// ar: AA
~printOutputTrig.({Sweep.ar(Impulse.ar(SampleRate.ir/4), DC.ar(1))/SampleDur.ir})
// note: these last two reset after 128 samples
// ar: KK
~printOutputTrig.({Sweep.ar(Impulse.kr(ControlRate.ir/2))/SampleDur.ir}, 129)
// ar: KA
~printOutputTrig.({Sweep.ar(Impulse.kr(ControlRate.ir/2), DC.ar(1))/SampleDur.ir}, 129)
```

Issue #6823, opened 2025-05-07
Bison Grammar: backtick/grave does unexpected things


JordanHendersonMusic:
## Environment
Supercollider 3.14

## Steps to reproduce

```supercollider
a = [1, 2];
`a.[0] // (`a).[0] // returns 1
`a[0] // `(a[0]) // returns Ref(1)
```

## Expected vs. actual behavior

Expected:
They behave the same.

I don't think this is possible to fix, at least not with the bison file in the state it is currently in. So perhaps this should be a documentation improvement as the backtick syntax isn't really discussed anywhere, but I am unsure.

Here is another interesting use of the backtick I didn't know about, `` `_ `` returns ` {|x| x}`.

```supercollider
a = `_; 
a.class == Function // ehh?
a.(x)  // `(x)
a.(x).value == x
```

JordanHendersonMusic:
`a.[]` doesn't imply value in the code though.           

It's implementation is exactly the same as `a[]`

```
  | expr '.' '[' arglist1 ']'
                        {
                                PyrSlotNode *selectornode;
                                PyrSlot slot;
                                PyrParseNode* args;

Here>>                           SetSymbol(&slot, s_at);
                                selectornode = newPyrSlotNode(&slot);
                                args = linkNextNode(
                                        (PyrParseNode*)$1,
                                        (PyrParseNode*)$4);
                                $$ = (intptr_t)newPyrCallNode(selectornode, args, 0, 0);
                        }
```

telephon:
I would have expected `a.[]` to throw an error.

JordanHendersonMusic:
Another surprising one is this `` `a = b`` which expands to ```(a = b)``.

jamshark70:
> a.[] doesn't imply value in the code though.

That relates to smoge's comments about functional specs, and the kinds of trouble we get into by never writing down a functional spec. Half this discussion is because the meaning of `.` without a method selector hasn't been written down.

`.` indicates a dot-syntax method call. If the method selector is omitted, then the it's assumed to be `value` -- even (to my surprise) when the receiver is a literal class:

```
{
	Point(0, 1);  // selector is `new`
	Point.(0, 1)  // selector is `value`
}.def.dumpByteCodes
```

I think if the code disagrees with this principle, then the code is wrong. (Incidentally, this is why `a.[0]` should not be expected to be a syntax error. It has always been valid to apply `[]` for array indexing to the result of an expression.)

[...]

PR #6828, opened 2025-05-09
sclang: remove `expr.[]` syntax, it is redundant, use `expr[]`

JordanHendersonMusic:
## Purpose and Motivation

From discussion over #6823

No one seems knows what `a.[1]` is supposed to do. 

In practice, it is the same as `a[1]`. Meaning it is very easy to replace.

## Types of changes

- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer:
After reading the conversation in https://github.com/supercollider/supercollider/issues/6823 I think this can indeed be removed, as risky as it sounds ("any possible current syntax or functionality is likely being used somewhere out there"...). But yes, this is easily fixable if it actually breaks anything.

BUT would it be worthwhile to note this somewhere in the docs? (e.g. "The `a.[]` syntax worked before 3.14 but it was removed; use `a[]` instead). Not sure where the best place should be though.

dyfer:
I marked this as "API change" so that we remember (hopefully) to note the change in the changelog.

prko:
I also found this a year or two ago and wondered why it was there.

Is the following part of the help documentation relevant?
https://docs.supercollider.online/Overviews/SymbolicNotations.html#Where%20f%20is%20a%20function
Screenshot 2025-05-12 at 23 35 09

telephon:
No, this is really the reason – now I also remember when it was introduced. So, after all, let's better keep it.

JordanHendersonMusic:
Yes this should be closed, since it does have an intended use.

PR #6829, opened 2025-05-09
GHA: add an option to manually trigger GHA builds

dyfer:
## Purpose and Motivation

Sometimes we may want to [manually re-trigger a build](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually) without pushing a new commit and this PR will allow that _once it's merged_. I don't think this is testable before it's merged, unfortunately.

(We could also consider whether we really need to run CI on every push to non-protected branches if we can trigger them manually, but that's for a separate issue, PR).

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -12,6 +12,7 @@ on:
       - 'HelpSource/**'
       - 'sounds/**'
       - '*.md'
+  workflow_dispatch:
 
 jobs:
   lint:


capital-G (approved):

dyfer:
Nice! It works:

image

Issue #6830, opened 2025-05-09
RandSeed is set at each cycle


telephon:
I just found this old forgotten branch – maybe I still find out what problem this is the answer to :).

PR #6831, opened 2025-05-09
RandID sets the id at each cycle.

telephon:
See #6830.

### DIFF
--- a/server/plugins/NoiseUGens.cpp
+++ b/server/plugins/NoiseUGens.cpp
@@ -792,11 +792,11 @@ void RandID_next(RandID* unit, int inNumSamples) {
     float id = ZIN0(0);
 
     if (id != unit->m_id) {
-        unit->m_id = id;
-        uint32 iid = (uint32)id;
-        if (iid < unit->mWorld->mNumRGens) {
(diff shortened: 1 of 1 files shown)


Spacechild1:
What is this PR supposed to do? I don't really see why we would need to set `mParent->mRGen` on every cycle...

telephon:
There was a reason, but I can't remember. I'll close this if I don't find the reason.

telephon:
It may have been an incomplete attempt at resolving this situation:

```supercollider


s.options.blockSize = 1;s.reboot;
(
{
	var trig = Impulse.ar(500);
	RandID.kr(LFPulse.kr(500));
	RandSeed.ar(trig, 3881);
	LFNoise0.ar(2000)
}.plot(0.05)
)
```
Screenshot 2025-05-09 at 23 49 34

Spacechild1:
Can you quickly describe the expected behavior VS the actual behavior?

telephon:
It should be a repeating waveform, but starts differently.

telephon:
Now it is like this:

Screenshot 2025-11-30 at 15 44 13

It should be precisely repeating, there is an extra value at the 8th position.

Issue #4128, opened 2018-10-17
Phasor adds the current input sample when triggered, but the previous input sample when not


to7m:
Environment
-----------

* Your SuperCollider version: 3.8.0
* Your operating system and version: Arch Linux

Steps to reproduce (for bugs)
-----------------------------
Here is some code that shows the problem:
```supercollider
{
    var impulse = Impulse.kr(100);
    var in = SinOsc.kr.range(0, 1) + (impulse * 5);
    var trig = impulse;
    var out = Phasor.kr(trig, in, 0, inf);
    [in, trig, out]
}.plot(0.03, separately: true);

{
    var impulse = Impulse.kr(100);
    var in = SinOsc.kr.range(0, 1) + (impulse * 5);
    var trig = Delay1.kr(impulse);
    var out = Phasor.kr(trig, in, 0, inf);
    [in, trig, out]
}.plot(0.03, separately: true);
```

Expected Behavior
-----------------
Phasor should add its own previous sample to the current input sample, unless triggered in which case it should add its reset position to the current input sample. Failing that, it should add the previous input sample in both cases.

Current Behavior
----------------
Phasor adds it's own previous sample to the _previous_ input sample, unless triggered in which case it adds its reset position to the _current_ input sample.

In the plot example, the first graph's Phasor adds the spike sample twice, but the second graph's Phasor doesn't add the spike sample at all.

mtmccrea:
I see a couple of issues in the source:

First, it looks like the logic for setting the calculation function in Phasor's Ctor, checks if the rate of the ugen is `calc_FullRate` (audio rate):

https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/TriggerUGens.cpp#L1787-L1801

and if it's NOT audio rate, sets  the calc function to be `Phasor_next_ak`, which doesn't make sense because that first input is audio rate, and so it will be reading the trigger input as if it's audio rate regardless of it's actual rate:

https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/TriggerUGens.cpp#L1833-L1837

It looks like perhaps lines 1797 and 1800 need to be swapped.

It also looks like there isn't a condition to check rates and assign `Phasor_next_ka`, which actually doesn't exist.

Then there's the issue of the underlying algorithm, which I'll get to next...

mtmccrea:
[...]

PR #6833, opened 2025-05-09
plugins: Phasor: fix init sample: don't ignore trig on first sample

### LINKED ISSUE #6805: plugins: Phasor doesn't start from resetPos on initial trigger
## Environment

## Steps to reproduce

```supercollider
(
{ 
  Phasor.ar(Impulse.ar(50), SampleDur.ir*10, start: 0, end: 1, resetPos: 0.5); 
}.plot(0.1)
)
```

Since Impulse emits a trigger already on its first output sample, I would expect Phasor to start from `resetPos` right away. Instead, Phasor starts from `start` for its first cycle and then correctly resets to `resetPos` for all the others.

Image

I've checked Phasor usage in our docs and it seems to me that this is not intended, although I wonder if anyone could have been using this on purpose to, e.g. start a buffer player from the beginning and only reset to another position after its first cycle.

The bug is related to Phasor init sample calculation which I would like to fix together with other TriggerUGens.

elgiano:
## Purpose and Motivation

Phasor was not calling its calcFunc in its Ctor, ignoring a trigger on its first sample and assuming its initial value to be its startPos. This PR fixes it, causing Phasor to start from resetPos instead if triggered on its first input sample.

This PR is part of #6813, and introduces a bug-fixing breaking change in Phasor output.

This is the only PR in this process for which I think somebody might be using the wrong behavior on purpose (Phasor always starts from startPos, and can reset to resetPos only later). However, I still see it as marginal. Documentation is clear on Phasor resetting to resetPos every time it receives a trigger, and never mentions it ignoring a trig on x[0] or always starting from startPos regardless of trig. There is no instance in the docs or examples in which this behavior is used or even encountered. But I also didn't find any complaint against this unwanted behavior, perhaps because I hardly find examples of different startPos and resetPos.

## Types of changes

- Bug fix
- Breaking change

This code shows the breaking changes in Phasor behavior, if Phasor is triggered at x[0] and has startPos != endPos.
If Phasor is triggered at x[0], and then again, then the change is only occurring between the first two triggers.

```supercollider
(
~printOutput = { |fn, nSamples(s.options.blockSize), action(_.postln)|
	fn.loadToFloatArray(nSamples/s.sampleRate, s, action)
};
~printOutputTrig = { |fn, nSamples=32|
	~printOutput.(fn, nSamples, {|v| v.round.asInteger.join.postln })
};
~printInitAndY0 = { |ugen, nSamples(s.options.blockSize)|
	~printOutput.({
		var out = ugen.value;
		var initSample = if (out.rate == \audio) {
			IEnvGen.ar(Env([out]),0);
		} {
			IEnvGen.kr(Env([out]),0);
		};
		[initSample, out]
	}, nSamples, {|v|
		"init = %\ty(0) = %".format(v[0], v[1]).postln
	})
}
)

~printInitAndY0.({Phasor.ar(1, 1, 0, 10, 5)})
// after:  init = 5.0	y(0) = 5.0
~printOutputTrig.({Phasor.ar(1, 1, 0, 10, 5)})
// before: 012345678901234567890
// after:  567890123456789012345
```

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

All already existing tests involving Phasor are passing.
For more discussion about testing (needs a special UGen) and documentation (needs a UGen Changelog), see #6813.

### DIFF
--- a/HelpSource/Classes/Phasor.schelp
+++ b/HelpSource/Classes/Phasor.schelp
@@ -60,7 +60,7 @@ code::
 
 // basic: ramp between 'start' and 'end'
 // note that the ramp ends at 9, not 10
-{ Phasor.kr(0, 1, start: 0, end: 10) }.plot(0.03)
+{ Phasor.kr(0, 1, start: 0, end: 10) }.plot(0.03).plotMode_(\dlines)
 
 // on trigger: ramp jumps to 'resetPos', goes until 'end' and wraps back to 'start'
 // note that the ramp starts from 'resetPos', since it receives a trigger immediately on its first sample
@@ -69,7 +69,7 @@ code::
 	var trig = Impulse.kr(ControlRate.ir/5);
 	var ramp = Phasor.kr(trig, 1, start: 0, end: 10, resetPos: 7);
 	[ramp, trig * 10]
-}.plot(0.03)
+}.plot(0.03).plotMode_([\dlines, \dstems])
 )
 
 
--- a/HelpSource/Classes/Phasor.schelp
+++ b/HelpSource/Classes/Phasor.schelp
@@ -58,6 +58,21 @@ Examples::
 
 code::
 
+// basic: ramp between 'start' and 'end'
+// note that the ramp ends at 9, not 10
+{ Phasor.kr(0, 1, start: 0, end: 10) }.plot(0.03)
+
+// on trigger: ramp jumps to 'resetPos', goes until 'end' and wraps back to 'start'
+// note that the ramp starts from 'resetPos', since it receives a trigger immediately on its first sample
+(
+{
+	var trig = Impulse.kr(ControlRate.ir/5);
+	var ramp = Phasor.kr(trig, 1, start: 0, end: 10, resetPos: 7);
+	[ramp, trig * 10]
+}.plot(0.03)
+)
+
+
 // phasor controls sine frequency: end frequency matches a second sine wave.
 
 (
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -1574,7 +1574,9 @@ void Phasor_Ctor(Phasor* unit) {
 
     unit->m_previn = 0;
     unit->mLevel = IN0(2);
-    Phasor_next_kk(unit, 1);
+
+    unit->mCalcFunc(unit, 1);
+
     unit->m_previn = 0;
     unit->mLevel = IN0(2);
 }
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -1572,8 +1572,11 @@ void Phasor_Ctor(Phasor* unit) {
         SETCALC(Phasor_next_ak);
     }
 
-    unit->m_previn = ZIN0(0);
-    ZOUT0(0) = unit->mLevel = ZIN0(2);
+    unit->m_previn = 0;
+    unit->mLevel = IN0(2);
+    Phasor_next_kk(unit, 1);
+    unit->m_previn = 0;
+    unit->mLevel = IN0(2);
 }
 
 void Phasor_next_kk(Phasor* unit, int inNumSamples) {
--- a/server/plugins/TriggerUGens.cpp
+++ b/server/plugins/TriggerUGens.cpp
@@ -1573,12 +1573,12 @@ void Phasor_Ctor(Phasor* unit) {
     }
 
     unit->m_previn = 0;
-    unit->mLevel = IN0(2);
+    float initLevel = unit->mLevel = IN0(2);
 
     unit->mCalcFunc(unit, 1);
 
     unit->m_previn = 0;
-    unit->mLevel = IN0(2);
+    unit->mLevel = initLevel;
 }
 
 void Phasor_next_kk(Phasor* unit, int inNumSamples) {


fivemtebex:
Excellent contribution to open source! The community will benefit greatly from this.

elgiano:
I added an extra commit to call mCalcFunc at init, instead of hardcoding Phasor_next_kk.
For Phasor, it's because `kk` is different from other functions in that it doesn't do subsample interpolation (see #4128). This PR is not aiming to fix subsample interpolation, but it should ensure `init sample = first sample` for all variants of Phasor.

Moreover, calling mCalcFunc seems to me a more explicit implementation of "pre-calculating y[0]", while choosing a specific variant works only under the assumption that all calcFunctions would produce the same y[0]. So, it seems to me a better candidate for a general pattern in calculating initSample, and I wonder if we should adopt it in all cases.

mtmccrea:
Thanks for the thorough look at this UGen!

> Documentation is clear on Phasor resetting to resetPos every time it receives a trigger, and never mentions it ignoring a trig on x[0] or always starting from startPos regardless of trig.

Can this be made explicit in the docs now for this change?

elgiano:
I think the docs are explicit enough, but maybe you mean adding something like a note saying that it was broken before the fix?

elgiano:
Although I think docs are clear enough, I also agree we can make it more explicit. But how about adding a few basic examples so that we can make it extra clear?

```supercollider
// basic: ramp between 'start' and 'end'
// note that the ramp ends at 9, not 10
{Phasor.kr(0, 1, start: 0, end: 10)}.plot(0.03)

// on trigger: ramp jumps to 'resetPos', goes until 'end' and wraps back to 'start'
// note that the ramp starts from 'resetPos', since it receives a trigger immediately on its first sample
{
	var trig = Impulse.kr(ControlRate.ir/5);
	var triggered = Phasor.kr(trig, 1, start: 0, end: 10, resetPos: 7);
	[triggered, trig * 10]
}.plot(0.03)
```

HelpSource/Classes/Phasor.schelp
```diff
@@ -58,6 +58,21 @@ Examples::
 
 code::
 
+// basic: ramp between 'start' and 'end'
+// note that the ramp ends at 9, not 10
+{ Phasor.kr(0, 1, start: 0, end: 10) }.plot(0.03)
```
mtmccrea, line 63: With examples like this, it can be helpful to make use of `plotMode_`, in particular `\dlines`, to illustrate increment sizes, etc.

`{ Phasor.kr(0, 1, start: 0, end: 10) }.plot(0.03).plotMode_(\dlines)`

HelpSource/Classes/Phasor.schelp
```diff
@@ -58,6 +58,21 @@ Examples::
...
+
+// on trigger: ramp jumps to 'resetPos', goes until 'end' and wraps back to 'start'
+// note that the ramp starts from 'resetPos', since it receives a trigger immediately on its first sample
+(
+{
+	var trig = Impulse.kr(ControlRate.ir/5);
+	var ramp = Phasor.kr(trig, 1, start: 0, end: 10, resetPos: 7);
+	[ramp, trig * 10]
+}.plot(0.03)
+)
```
mtmccrea, line 73: or `\dstems` for illustrating timing of triggers that are impulses:
```supercollider
(
{
	var trig = Impulse.kr(ControlRate.ir/8);
	var ramp = Phasor.kr(trig, 1, start: 0, end: 10, resetPos: 7);
	[ramp, trig * 10]
}.plot(0.03).plotMode_([\dlines, \dstems])
)
```

mtmccrea: Actually, by playing with this it made it more obvious that actually it's reseting to `resetPos + 1`, not `reset` as it should...

mtmccrea:
After running the example above, I see that actually it's reseting to `resetPos + 1`, not `resetPos` as it should...

I'll have to look with fresh eyes later, feels like a 🕳️🐇

mtmccrea:
> For this PR, I would be happy to fix the init sample behavior

I think that's reasonable, given that more problems keep showing up with Phasor.

Note, though, that this doesn't fully fix #6805, in particular at these rates/inputs (there could be more combinations):
```supercollider
// only some combinations start at 5 on initial trigger
~printOutputTrig.({Phasor.ar(Impulse.ar(0), 1, 0, 10, 5)}, 10) // :(
~printOutputTrig.({Phasor.kr(Impulse.ar(0), 1, 0, 10, 5)}, 64*10) // :(
~printOutputTrig.({Phasor.ar(Impulse.kr(0), 1, 0, 10, 5)}, 10) // OK
~printOutputTrig.({Phasor.kr(Impulse.kr(0), 1, 0, 10, 5)}, 64*10) // :(
~printOutputTrig.({Phasor.ar(1, 1, 0, 10, 5)}, 10) // OK
~printOutputTrig.({Phasor.kr(1, 1, 0, 10, 5)}, 64*10) // :(
```
The broken ones in this case all call `Phasor_next_ak`, which does the inter-sample trigger calculation.

So I'd say just update the title to reflect that this fixes the init/first sample agreement, and remove "fixes #6805" so GitHub doesn't suggest closing it when this is merged.

elgiano:
I changed the title and addressed https://github.com/supercollider/supercollider/pull/6833#discussion_r2105081558

elgiano:
I don't remember if I had proposed this before, probably because it's not a great idea, but: how about making Phasor calculate sub-sample offset like Sweep (`-previn/(curin+previn)`)? It would still be wrong, but it would work as expected with Impulse. So it would at least exposed the expected behavior.

I don't like the idea of fixing something with another implementation that is known to be wrong... but I'm thinking it would minimize changes across 3.14 and 3.15 regardless of how we want to fix sub-sample offset, if that fix happens in 3.15

dyfer:
If this is good to be merged, I'll let @elgiano or @joshpar to do the merging.

PR #6613, opened 2025-02-09
fix Windows pathname resolution

sadguitarius:
## Purpose and Motivation

Fixes #6501

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation
- [x] This PR is ready for review 

The fix is related to the incompatibility of `FindFirstFileW` with paths that end in trailing path separators, documented [here](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstfilew). If anyone knows of a better way to do this without having to mess with strings, suggestions are welcome!

PR #6834, opened 2025-05-10
ServerTreeView: fix "group rect issue" and "node visibility issue" of #6294, fix "Rect collision issue with too many nodes"

prko:
---
## Summary
Fixes visibility of node ids for groups and related issues introduced in https://github.com/supercollider/supercollider/pull/6294, see https://github.com/supercollider/supercollider/pull/6294#issuecomment-2867917864.

However,

There are new perspectives by @dyfer.
@SimonDeplat will open a new PR after merging this PR.
@telephon pointed out a rect collision issue when many nodes are used. (This collision is not related to [#6294](https://github.com/supercollider/supercollider/pull/6834#6294)) and the purpose of this PR). I fixed it in the last commit.

----

## Purpose and Motivation

Fixes #6294
Please watch [the demo video](https://www.dropbox.com/scl/fi/mpn6k26wm17l7eljw4i9b/Fix-plotTree-group-boundaries.mov?rlkey=d6eeijnb3cxln6xlubkcz2k6e&st=skiea7ds&dl=0)

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


prko:
In the video attached to the initial post of this pull request, there is one more issue that requires attention. 

Prior to #6294, all nodes remained visible when adjusting the window height. However, in #6294, any nodes positioned outside the initial height of the window are not displayed. The commit https://github.com/supercollider/supercollider/pull/6834/commits/8237fa7ae802301e78b926dfabc439e83b4b2fab addresses that issue.  

Please watch [the demo video.](https://www.dropbox.com/scl/fi/o419rwhb4t6y19iobqd3l/Fix-plotTreeview-node-hidden.mov?rlkey=yg40ip2dh999soxhr0fnko8ah&st=s7iv6a3b&dl=0)

Therefore, I believe that both problems related to #6294 have now been resolved. 

NOTE: I commented on unnecessary parts for other developers, but I can remove those comments if necessary.

Please do let me know if I have overlooked anything.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerTreeView.sc
```diff
@@ -152,8 +153,10 @@ ServerTreeView {
 				size
 			};
 			defer {
-				view.bounds = Rect(0, 0, 400, max(400, tabSize * (countSize.value(levels) + 2)));
-				view.refresh;
+				//view.bounds = Rect(0, 0, 400, max(view.bounds.height, tabSize * (countSize.value(levels) + 2)));
+				//view.refresh;
```
dyfer, line 157: Is there a reason these commented out lines are kept?

prko: Because I am not sure about my fix.... Someone may find a better solution during the review process or in the future.

I can remove it if reviewers find that this fix is good.

prko: @telephone
Could you take a closer look at this PR? I think I have fixed the problem you reported and another one I found when checking the group node issue.
Sorry for not checking the group nodes in #6294 and the other issue I explained here.

telephon:
Thanks, this almost fixes it.

There are some changes that you should still do.

Firstly, there is an overlap at the end of the view (it may be considered nice, but might be unintended):
Screenshot 2025-05-13 at 09 47 41

Then, I think that the info text with "STARTED" in bold is not necessary, I think – the title of the window already says it.

Screenshot 2025-05-13 at 09 49 23


Finally, I get an error when I close the view and then reopen it with `s.plotTree`. Probably there is a cleanup missing in the window `onClose` action.

```
s.plotTree;

ERROR: Message 'height' not understood.
RECEIVER:
   nil
```

prko:
@telephon 
Thanks for your comments!

1. >Firstly, there is an overlap at the end of the view (it may be considered nice, but is a little confusing):

   I cannot reproduce it. Could you please give me the code snippet you tested? 
  The following screenshots are from the `s.plotTree` on my end (the second one is scrolled down, and there is no overlap at the end):
   Screenshot 2025-05-13 at 22 30 38
   Screenshot 2025-05-13 at 22 30 43

2. >Then, I think that the info text with "STARTED" in bold is not necessary, I think – the title of the window already says it.
 
   The info text is from the `ServerTreeView`:
   ```
   ~aServerTreeView = ServerTreeView(s) 
   // ~aServerTreeView label: `STOPPED`
   
   aServerTreeView.start 
   // aServerTreeView starts updating nodes on the server, 
   // and info text changes to ` STARTED: current server tree`
   // with colour change.

   aServerTreeView.stop 
   // aServerTreeView stops updating nodes on the server, 
   // and info text changes to ` STOPPED: last updated server tree`
   // with colour change.
   ```
   With commit https://github.com/supercollider/supercollider/pull/6834/commits/c180f8e50d0ee71f5d54c0b5ce7dc753b2dd4297, the following ways are available:
   ```
   s.plotTree
   s.serverTreeView.stop
   s.serverTreeView.start
   s.serverTreeWindow.close

   x = s.plotTree
   x.stop
   x.start
   x.close
   ```
   **To enable this functionality, `s.plotTree` now returns `a ServerTreeView`.**.

3. >Finally, I get an error when I close the view and then reopen it with s.plotTree. Probably there is a cleanup missing in the window

   It is fixed with commit https://github.com/supercollider/supercollider/pull/6834/commits/c180f8e50d0ee71f5d54c0b5ce7dc753b2dd4297.

dyfer:
Thanks for fixing this!

Btw how would we feel about changing "STARTED" to "RUNNING"?

prko:
>Btw how would we feel about changing "STARTED" to "RUNNING"?

OK! Thanks, done with https://github.com/supercollider/supercollider/pull/6834/commits/a527af88635d3d3e25c48851528bdd6cb86c2520!

dyfer:
I'm not sure if the issue was introduced in #6294 or here, but restarting the server does not update the "RUNNING"/"STOPPED" text properly. 

```supercollider
s.boot;
s.plotTree;
s.quit;
s.boot; // the header still says "STOPPED"
```

Also, when the plotTree is open, pressing cmd-, triggers an error:
<details>

```
ERROR: Message 'cmdPeriod' not understood.
RECEIVER:
Instance of Routine {    (0x129c4eb08, gc=F0, fmt=00, flg=00, set=05)
  instance variables [28]
    state : Integer 5
    func : instance of Function (0x1587b0418, size=2, set=2)
    stack : instance of Array (0x110368080, size=1, set=9)
    method : instance of Method ServerTreeView:start (0x110784880)
    block : instance of FunctionDef in Method ServerTreeView:start
    frame : Frame (0x1583d1c48) of Function
    ip : RawPointer 0x110788bf8
    sp : RawPointer 0x1103680a8
    numpop : Integer 0
    receiver : instance of ServerTreeView (0x1599b7118, size=9, set=4)
    numArgsPushed : Integer 1
    parent : nil
    terminalValue : nil
    primitiveError : Integer 0
    primitiveIndex : Integer 0
    randData : instance of Int32Array (0x129c383b8, size=4, set=2)
    beats : Float 50.867299   A42F1ED0 40496F03
    seconds : Float 815.475903   A6410BA8 40897BCE
    clock : instance of TempoClock (0x129c481c8, size=7, set=3)
    nextBeat : Float 51.367299   A42F1ED0 4049AF03
    endBeat : nil
    endValue : nil
    environment : instance of Environment (0x110234708, size=5, set=3)
    exceptionHandler : nil
    threadPlayer : nil
    executingPath : nil
    oldExecutingPath : nil
    rescheduledTime : nil
}
ARGS:
KEYWORD ARGUMENTS:
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg this = <instance of Routine>
		arg selector = 'cmdPeriod'
		arg args = [*0]
		arg kwargs = [*0]
	ArrayedCollection:do
		arg this = [*3]
		var i = 2
	List:do
		arg this = <instance of List>
	Meta_CmdPeriod:run
		arg this = <instance of Meta_CmdPeriod>
	Main:stop
		arg this = <instance of Main>
	Interpreter:interpretCmdLine
	Process:interpretCmdLine
		arg this = <instance of Main>
^^ ERROR: Message 'cmdPeriod' not understood.
RECEIVER: a Routine
```

</details>

prko:
>restarting the server does not update the "RUNNING"/"STOPPED" text properly

Thanks for pointing that out.

That feature was commented out in an earlier commit. It is restored with https://github.com/supercollider/supercollider/pull/6834/commits/bab9e4cbd48b0aab20d44dff9166da15bee86797.

dyfer:
Thanks.

Sorry if this distracts in this PR, but I feel it should've been addressed in #6294: In the current implementation [this block](https://github.com/supercollider/supercollider/pull/6834/files#diff-f3e6a5c8f848ee065d1987846138df3ed5f198136a50bbd0886247b6e64810f1L165-L172) keeps being called every second if the server is not running.

Could we consider updating the running/stopped text using the `doOnServerBoot`/`doOnServerQuit`? They could also start and stop the updater routine too (which would be an improvement over the original design).

prko:
To confirm my understanding. 
Like this (this currently displays only STOPPED, I just want to know if this direction is correct )?
```
ServerTreeView {
	...
	start {
	...
	updater = {
			fork {
				loop {
					defer {
						if(window.isClosed.not) {
							this.doOnServerBoot;
							this.doOnServerQuit;
						}
					};
					this.server.sendMsg("/g_queryTree", 0, 0);
					interval.wait;
				};
			}
		};
		...
	}
	doOnServerBoot {
			treeViewStatus.string_(" RUNNING: current server tree")
			.background_(Color.grey(0.9))
	}

	doOnServerQuit {
			treeViewStatus.string_(" STOPPED: last updated server tree")
			.background_(Color.grey(0.7))
	}
}
```

dyfer:
I _think_ the implementation would look like this:

```supercollider
ServerTreeView {
	var active = false;


	makeWindow { |bounds, parent|
		...
		.onClose_{ this.stop }; // run the full cleanup on close
		active = true;

		if(server.serverRunning) {
			this.start;
		} {
			this.stop; // or maybe just ServerBoot.add(this.start, server)?
		}
	}
	
	start { |interval = 0.5, actionIfFail|
		...
		
		updater = {
			fork {
				loop {
					this.server.sendMsg("/g_queryTree", 0, 0);
					interval.wait;
				};
			}
		};
		...
		ServerQuit.add(this.stop, server);
	}
	
	stop {
		...
		if(active) {ServerBoot.add(this.start, server)}; // only register the action of the window is open
	}
	
	close {
		active = false;
		this.stop;
		window.close; 
		ServerBoot.remove(this.start, server);
		ServerQuit.remove(this.stop, server);
	}
}
```
But I haven't worked with `ServerBoot` / `ServerQuit` classes  much so I may be missing something.
If you don't feel confident doing this rework, maybe let's not do it. It seems out of scope of _this_ PR. Such rework it needs to be tested for race conditions etc to make sure routine is stopped when needed etc... still, I think it would be cleaner than a the current solution.

Additionally, the ServerTreeView would start itself up, so `.start` would need to be removed from the `Server .plotTree` method).

What do you think?

prko:
With your implementation idea, I get a warning and an error:

```
s.plotTree 
```
>->WARNING: Server failed to respond to Group:queryTree!
```
s.serverTreeWindow.close 
```
>-> a Window
ERROR: Message 'hasColor' not understood.

I am trying to resolve the problems.

---

>Additionally, the ServerTreeView would start itself up, so `.start` would need to be removed from the Server `.plotTree` method).
What do you think?

From the user's experience, removing the `.start` from the server's `.plotTree' method has two consequential effects:
- Removing it is logical, but inconvenient for the user, as the user should always start it manually.
   (Using `.plotTree` offers the user several conveniences over using `ServerTreeView': single instance window, automatic start.)
- For consistency of `.plotTree`, removal of `.start` from the server should be accompanied by removal of auto start by server boot and removal of audo stop by server quit.

dyfer:
> * Removing it is logical, but inconvenient for the user, as the user should always start it manually.

Sorry I wasn't clear. 
Functionally I think it should still start by itself! But I thought that with my proposed implementation it would start automatically when `ServerTreeView` is created:

```supercollider
	makeWindow { |bounds, parent|
		...

		if(server.serverRunning) {
			this.start;
		};
```

I haven't tested this so I'm not sure why it does not.

prko:
Ah, you are correct!

>Additionally, the ServerTreeView would start itself up, so .start would need to be removed from the Server .plotTree method).

Sorry for my misreading.
In your implementation, it should be changed. I will rethink!

dyfer:
Sorry for derailing this conversation!
Yes, let's leave redesign for a separate PR, I'll write my thoughts in a new issue.
I'll let @telephon finish this review - I think the basic functionality is now working here.

prko:
Please let me know the proper new name for the class `ServerTreeView` if it should be changed. What I can imagine are:
- ServerTreeView
- PlotTreeView
- NodeTree 
- NodeTreeView
- ServerNodeTree
- ServerNodeTreeView

telephon:
If I understood correctly, the name chosen in the dev meeting was `NodeTreeView` – right? https://github.com/supercollider/supercollider/issues/6853#issuecomment-2898707494

telephon:
stop the server while the view is open. It closes by itself after a while, and reports:
```
WARNING: Server failed to respond to Group:queryTree!
```

Then, I still get this error:

```
ERROR: Message 'hasColor' not understood.
RECEIVER:
   nil
ARGS:
   Symbol 'window'
KEYWORD ARGUMENTS:
```

The view overlap is reproducible here:

```supercollider

(
s.bind {
	g = Group(s);
	{ 
		var h = Group(g);
		{ Synth(\default, [\amp, 0.0001], target: h, addAction: \addAfter) } ! 5
	} ! 15;
}
)

s.plotTree
```

prko:
Have a look at the demo video, which shows this problem for example with `s.plotTree` (the related class also has this problem, although I did not show it in the video):
[demo video](https://www.dropbox.com/scl/fi/cyyabxj4fseqbf9utfck3/CmdPeriod-bug.mov?rlkey=6k7bly3yuz0g8ehc3dx59yxhw&st=419siy0z&dl=0)

telephon:
Sorry, demo videos are difficult to understand, I didn't get what the problem and the suggested solution was.

prko:
I have uploaded a new commit. To read the changes I made, please refer to the text at https://github.com/supercollider/supercollider/pull/6834/commits/8346c0f9d7033371e982cd3d566f91e2427779a3.

As I mentioned in the commit message, there is interference between the start, stop, and interval functions due to the presence of only a single updater and OSCFunc. I do have an idea for resolving this issue, but I believe it would be more appropriate to address it in a separate PR. Additionally, as I personally do not use multiple instances of ServerTreeView, I feel that the approach to handling this should be discussed with developers who require support for multiple instances of ServerTreeView.

@telephon 
>The view overlap is reproducible here:

Yes, the view overlap is reproducible with your code.
I realise that it has been present since at least SC 3.9.3. While I am unsure why you brought it up when reviewing this PR, I have nonetheless attempted to address it.
You will find [a demo video](https://www.dropbox.com/scl/fi/m3gg8waa47rektl9pod1r/plotTree-demo.mov?rlkey=k6dafu7jdma4tpb189gvu48zj&st=pmt0kb7i&dl=0) along with the code below used in the demonstration:

```
s.boot

s.plotTree

(
{ Group(s) } ! 2;

{}.play;

x = {
	s.bind {
		g = Group(s);
		{
			var h = Group(g);
			{ Synth(\default, [\amp, 0.0001], target: h, addAction: \addAfter) } ! 4
		} ! 100; // ! 100 to see in a extreme case

		{}.play;
	};
};

x.();
)

s.meter;

s.freqscope;

s.scope

s.quit

s.boot // automatically starts plotting

s.serverTreeView.stop // not: a.plotTree.stop

x.()

s.serverTreeView.start // not: s.plotTree.start

s.serverTreeView.stop

/* cmd + .*/
// I think pressing cmd + . should not update plotting when the label is stopped.


s.serverTreeView.start

s.serverTreeWindow.close

////////////////////////

s.quit

s.plotTree

s.plotTree;s.plotTree

a = ServerTreeView()

s.boot

s.plotTree // this does not start when a window is already open.
s.plotTree(1)
s.plotTree.stop //short.but not recommended to use

a.start(2)
a.stop
a.close

s.plotTree
b = s.plotTree

b.stop
b.start(1)
b.start
b.close
b.start
```

telephon:
Hi @prko – thank you! I will have to defer this, maybe someone else can jump in as well.

dyfer:
@prko could you please update the first post to summarize what does this PR try to address?

We are planning to make a release in about a week, I feel like this should be addressed beforehand.

dyfer:
> > #6294 is a PR
> 
> Yes, it is a PR. After it was merged, @telephon found a problem and reported it as a comment on that PR. See the following comment and with the subsequent coment: [#6294 (comment)](https://github.com/supercollider/supercollider/pull/6294#issuecomment-2867917864).

Okay. Then at it would've been better to link to that comment, instead of the whole PR.

I'd still prefer if the opening text said something like "Fixes visibility of node ids for groups and related issues introduced in #6294, see https://github.com/supercollider/supercollider/pull/6294#issuecomment-2867917864"

> 
> Then, @dyfer pointed a new problem: [#6834 (comment)](https://github.com/supercollider/supercollider/pull/6834#issuecomment-2882680551), and [#6834 (comment)](https://github.com/supercollider/supercollider/pull/6834#issuecomment-2884487697) with subsequent comments.

Yes, I have! But I've withdrawn my suggestions later: https://github.com/supercollider/supercollider/pull/6834#issuecomment-2887144927. Sorry for derailing the conversation.
> 
> In the last PR I introduced the structure suggested by @dyfer: [#6834 (comment)](https://github.com/supercollider/supercollider/pull/6834#issuecomment-2885528384)

As I said in https://github.com/supercollider/supercollider/pull/6834#issuecomment-2887144927, I've withdrawn my suggestion to rework it at this time. I'm sorry for the noise and extra work this has caused.

Could you revert this PR so that it only fixes for the visibility of node ids and rect collision?

prko:
I removed `ServerBoot.add`, `ServerBoot.remove`, `ServerQuit.add` and `ServerQuit.remove`.
While doing that, I fixed all the minor bugs which I found.

The following is the test code which I used: 

```
s.plotTree // will be closed since server is not running

s.boot

s.plotTree

(
{ Group(s) } ! 2;

{}.play;

x = {
	s.bind {
		g = Group(s);
		{
			var h = Group(g);
			{ Synth(\default, [\amp, 0.0001], target: h, addAction: \addAfter) } ! 4
		} ! 100; // ! 100 to see in a extreme case

		{}.play;
	};
};

x.();
)

s.meter;

s.freqscope;

s.scope

s.quit // the last plot is still there

s.boot // automatically starts plotting

//stop plotting current node structure
s.serverTreeView.stop // not: a.plotTree.stop

x.() // produces sound, but plotTree does not update the current node structure

// plotting again
s.serverTreeView.start // not: s.plotTree.start

s.serverTreeView.stop

/* cmd + .*/
// I think pressing cmd + . should not update plotting when the label is stopped. So, it behaves as I intended

s.serverTreeView.start

s.serverTreeWindow.close

////////////////////////

s.quit

s.plotTree // will be closed since server is not running

s.plotTree;s.plotTree // same as above

a = ServerTreeView()
a.start // will not be closed in 3 seconds unlikely to .plotTree

s.boot // a starts automatically when server boots

s.plotTree

x = Pn((dur: 0.4, midinote: 69, amp: 0.1)).play;

s.serverTreeView.stop

s.plotTree // this does not start when a window is already open.

s.plotTree(1)
s.plotTree.stop //short, but not recommended to use. See below the recommended way

a.stop

// recommended way:
b = s.plotTree
b.stop
b.start(1)
b.start
b.front
b.close
b.start

a.close
```

dyfer:
Thanks @prko for cleaning up the PR!

Before making any changes, let's discuss the following.

To move forward I'd like to eventually:
1. clarify desired display
2. clarify interface
3. improve updating implementation
4. rename the class
5. other remarks


### 1: display
In this previous discussions in this and former PR, what was the idea behind posting the "Running" vs "Stopped" message: Is it to inform user that the _server_ is running, or that the _tree plot_ is running?
I think the former might be misleading: other server related views, like FreqScope and ServerMeter can't be stopped explicitly.

If we are informing the user about the server status, why not simply post a blank screen when the server stops? That would be very clear.
However, I understand that the user may want to see the node tree e.g. when the server stopped... but I'd treat that as a special case, and the blank window as default. **What do you think?**

### 2: interface

We shouldn't post multiple warnings for a single action, like we do when we open a plot and the server is not running. 

We also shouldn't post a warning when bringing the plot window to front with `s.plotTree`. `s.meter` and `s.freqscope` don't post a warning upon subsequent execution.

I think the current logic with warnings makes `Server .plotTree` unnecessarily extensive. I think we only need
```supercollider
	plotTree { |interval, bounds|
		if(serverTreeWindow.isNil or: { serverTreeWindow.isClosed }) {
			var actionIfFail = { defer { serverTreeWindow.close } };
			bounds = bounds ?? { Rect(128, 64, 400, 400) };
			bounds = bounds.minSize(395@386);
			serverTreeWindow = Window(name.asString ++ " Node Tree", bounds, scroll:true);
			this.plotTreeView(
				interval ?? { 0.5 },
				serverTreeWindow,
				actionIfFail);
			serverTreeWindow.front
		} {
			serverTreeWindow.alwaysOnTop_(true).front.alwaysOnTop_(false);
		};
		^serverTreeView
	}
```

Also, I don't see a reason to stop/start the node display itself... so I wouldn't instruct users to run .start / .stop themselves. What's the reason for it?

### 3. implementation

Keeping track of the server status is currently a bit convoluted. I have an idea how to streamline it, but it touches a lot of code.
**Would it be okay if I pushed to your branch** a suggested solution (as a new commit), or should we merge this as is and should I do this in a separate PR?

### 4. rename the class

I think I'll do that in a new PR, after this is merged, unless you prefer to do it right here.

### 5. other

- I was really surprised that the window closes after a few seconds if plotTree is open when the server is not running - I never realized that's a thing. IMO that's a peculiar behavior, as the view does not close when the server quits (and I don't think it should!). I'd leave addressing this for a separate issue/PR, but I'm curious whether you find this functionality desirable.
- I would not close an already open plot window, if the server is not running. Was this functionality requested in the previous review?

Thanks again for working on this!

HelpSource/Classes/ServerTreeView.schelp
```diff
@@ -7,7 +7,7 @@ description::
 link::Classes/ServerTreeView:: shows the node tree on a server.
 
 note::
-The link::Classes/ServerTreeView:: will disappear when the pointed server is not running.
+Unlikely to link::Classes/Server#-plotTree::, the link::Classes/ServerTreeView:: will not disappear when the pointed server is not running.
```
dyfer, line 10: `Unlikely to link::Classes/Server#-plotTree::,` -> `Unlike link::Classes/Server#-plotTree::,`

HelpSource/Classes/ServerTreeView.schelp
```diff
@@ -57,7 +57,7 @@ a.stop
 [a, c].do { |v| v.close }
 
 // Open a plotTree window:
-s.plotTree
+s.plotTree // This changes also the interval of the ServerTreeView instances assinge to variabled since there is currently only one updater for this class.
```
dyfer, line 60: minor change suggested:
`This also changes the interval of the ServerTreeView instance assigned to a variable, since`...

dyfer (commented):
When looking through the PR, I marked a few minor requests in the docs.

HelpSource/Classes/ServerTreeView.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6834/commits/bf4cc47da33fa088117afc433197068e265af4db

HelpSource/Classes/ServerTreeView.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/6834/commits/bf4cc47da33fa088117afc433197068e265af4db

HelpSource/Classes/Server.schelp
```diff
@@ -708,21 +728,23 @@ s.waitForBoot {
...
 	the first node will appear after a 2-second delay (once four notes have played),
 	meaning only every fourth synth node is visible.
 	*/
 
 	5.wait; // waits for 5 seconds.
 
-	plotTree.plotTreeWindow.close; // closes the plotTree window
 	x.stop; // stops the playback of the pattern
+
+	2.wait; // waits for 5 seconds.
```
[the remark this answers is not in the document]
prko: Thanks don in https://github.com/supercollider/supercollider/pull/6834/commits/47631829360c543e7d602117baf058d64c699a2e

Applied also other comments in the same code example.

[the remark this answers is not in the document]
prko: Regarding this:

Despite being discussed multiple times in an attempt to establish a common convention in SC documentation, it seems that it largely comes down to personal preference. Personally, I find it unclear and difficult to follow. Some reviewers prefer infinitives in sentences with contextual but not grammatical substance, while others prefer the third-person singular form. I have therefore decided to defer to the opinion of the individual reviewer.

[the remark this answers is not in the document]
dyfer: I'm sorry! This was a typo on my part. I don't have a strong preference over infinitive vs imperative... my comment was meant to be about the **number**! (5 vs 2)

dyfer:
> ### 1. Display (Consistency & unique window policy)
> The original purpose of this PR and its ancestor (#6294) is to improve consistency between `.meter`, `.freqscope`, `.scope`

Yes!

> Regarding the “Running” and “Stopped” labels: these refer to the state of the plot rather than the server. My intention was to make the interface more approachable, including all possible cases. 

Note that ServerMeterView can be stopped as well, but does not display anything when it's stopped. The extra verbosity may be unnecessary here (but see my proposition below).

Also, it's much more common to have a stopped server, than to have a stopped view IMO. I think the stopped server could be indicated as such (also see below).

> I believe that keeping the last-known node tree visible when the server has stopped can be helpful, for instance, when there have been rapid changes which the user might want to review. (I experienced some of such cases, and making it blank when the server is quitting or when stopping plotting does not help to inspect the node structure.) However, I do appreciate the clarity that a blank screen could provide, and remain open to changing to that approach if there is consensus.

My thoughts on this are:
- I appreciate both the last known state and the clarity of the blank state
- I came up with a mixed solution: making the display semi-transparent (faded) when the view or the server are not running. 

> 
> ### 2. Interface (Warning minimisation & start/stop)
> With regard to warnings, my current implementation includes alerts which I believe are especially useful to beginners. However, I agree that repeated warnings—particularly when simply bringing an existing window to the front—can become redundant or inconsistent with `.meter` and `.freqscope`. If preferred for greater consistency, I am happy to revise the logic.
> 
> As for the `.start` and `.stop` methods: these were introduced following reviewer suggestions (see [@telephon's comments and my subsequent responses](https://github.com/supercollider/supercollider/pull/6294#discussion_r1591988989)).
> 
> (On reflection, I wonder whether I should not have followed @telephone’s suggestion after all. However, I still believe it would not have been possible to achieve my goal without accepting the reviewer’s recommendation. @SimonDeplat wasn't a reviewer, but I noticed in [a comment on his PR](https://github.com/supercollider/supercollider/pull/6545#issuecomment-2595738603) that, like @telephone, he didn't want a unique 'plotTree' window. Other reviewers would probably have highlighted such a change as well.)

The .start/.stop methods are good in terms of the class behavior. But I personally don't see much user for them from the user perspective.


> 
> ### 3. Implementation
> I favour the simplest and cleanest solution. If you have suggestions for streamlining server status management—either by direct commits to this PR, or by a subsequent PR after merging—please proceed as you feel is best. I am happy to collaborate and leave the details to your discretion.

Sounds good! I'll push a commit to this branch with the proposed solution and we can refine this further.


> ### 4. Renaming the class
> With regard to renaming the class and related methods, I am happy to follow your suggestions. If you proceed with renaming in a subsequent PR, I would recommend updating related getters, such as `.serverTreeView` and `.serverTreeWindow` in `Server.sc` (see line 367). Please let me know if there are any preferences or conventions to observe.

Yes, I'd update the getters etc. But I'll do that in a separate PR, after this one is merged.

> 
> ### 5. Other (Auto-closing window behaviour)
> Regarding the automatic closing of the `.plotTree` window if opened while the server is not running: I originally found this behaviour unexpected, but have since become accustomed to it. 

Well, I also found it unexpected!

I'm quite convinced that now that the plot follows the server status reliably (in my PR) it should just stay on. **This is what I implemented in my commit** 

By contrast, `ServerTreeView` does not implement auto-close, in response to requests from developers who wished to be able to open and keep multiple windows even when the server is stopped. 
> I believe this approach provides a good balance between the original design intention (unique window with auto-close) and the needs of users requiring persistent windows.

I appreciate the desire to keep the original design, but I don't find it reasonable, when the view updates properly once the server starts. That's how s.meter `behaves` BTW - you can open it without the server running and it will start monitoring when the server starts.

---

I'll push a PR with my proposed changes.

dyfer:
I pushed _proposed_ changes.

The main improvement is with keeping track of the server status - it is now done using `SimpleController`, similarly how s.makeGUI does it.

Aside from some code cleanup (unused variables etc), I maintained the existing node plotting, but I made minor layout adjustments to avoid scroll bars by default and made the group rectangles resize with the window width. We can further tweak this as needed.

I changed posted text a bit. My main point was for the plot to look close to what it looked like before all these changes when the server is running. 
What we want to know is when something is _NOT_ running, so I made posting in those cases.
Additionally I added fading of the existing plot when the server quits (or when the updating is stopped).

I removed automatic closing of the window if the server is not running. I also commented out the associated warning - I'm not sure if it should be removed or not. Maybe someone else can make that call later on.

`a = s.plotTree;` - server not running:
Screenshot 2025-05-29 at 2 55 23 PM

Server booted:
Screenshot 2025-05-29 at 2 55 30 PM

Tree plotted as before:
Screenshot 2025-05-29 at 2 55 36 PM

View stopped (`a.stop`):
Screenshot 2025-05-29 at 2 55 48 PM

Server quit (`s.quit`):
Screenshot 2025-05-29 at 2 55 57 PM

----
Please let me know what you think! I feel pretty good about the code structure now, but I'm happy to adjust posting, formatting etc.

SCClassLibrary/Common/GUI/PlusGUI/Control/ServerPlusGUI.sc
```diff
@@ -359,23 +359,23 @@
 		this.startAliveThread;
 	}
 
-	plotTree { |interval=0.5, bounds|
-		if(plotTreeWindow.isNil) {
+	plotTree { |interval, bounds|
+		if(serverTreeWindow.isNil or: { serverTreeWindow.isClosed }) {
```
dyfer, line 363: Is there a reason the default value was removed from `interval`? IMO it's more readable with the default value there, so that e.g. it shows up in help.

I wanted to ask about the addition of bounds here (from #6294). Since now we have a separate class, which can be instantiated with desired bounds, do we still need to have the option of adding the bounds here?

FWIW, s.meter e.g. does not provide the `bounds` argument.

prko: > Is there a reason the default value was removed from interval? IMO it's more readable with the default value in the argument list, so that e.g. it shows up in help

The default value for `interval` was removed quite recently to distinguish between the following scenarios:
```supercollider
// 1. Boot the server:
s.boot

// 2. Open the unique s.plotTree window and start plotting since the server is running:
s.plotTree

// 3. Take any action that brings the open s.plotTree window back (for example, resizing the sc-ide to obscure part of the s.plotTree window).

// 4. Bring the existing .plotTree window (created in step 2) to the front explicitly.
//     Previously, there was no dedicated way to do this, so I structured the logic as follows:
//     If .plotTree is called when a plotTree window already exists, sclang checks whether an interval argument is provided:
//      - If there is no interval argument, it simply brings the existing window to the front.
//      - If there is an interval argument, it re-initialises the plot interval: the previous plot is stopped, and then a new one is started, applying the given interval.
s.plotTree       // brings the existing .plotTree window to the front.
s.plotTree(1)    // brings the existing window to the front and updates the plot interval, restarting the plotting accordingly.
```
**Summary**:

The `s.plotTree` function checks whether its unique window is already open. If not, it creates the window. If the window is open, whether it is simply brought to the front or the plotting interval is changed (stopping the current plotting and starting a new one with a different interval) depends on the existence of the interval argument.

This behaviour is not possible if a default value remains in the argument list, as invoking `s.plotTree` would always apply the default interval rather than simply bringing the window forward. I appreciate that this may have an impact on how the method is displayed in the help, but separating these cases provides clearer and more flexible behaviour for the user.

---

>I wanted to ask about the addition of the bounds argument here (from https://github.com/supercollider/supercollider/pull/6294). Since now we have a separate class, which can be instantiated with desired bounds, do we still need to have the option of adding the bounds here?
>
>FWIW, s.meter does not provide the bounds argument.

For `s.meter`, there is a `position` argument. This was recently added. I intended to add bounds, but as the size of s.meter is automatically calculated, position is needed.

For `s.scope`, see #6282, which implements various approaches based on @mtmccrea’s suggestion:
- For `s.scope`, `.bounds` and `.bounds_` can be used.
- For `{}.scope`, my initial suggested `bounds:` argument is still supported.

I had considered making a PR to add either a `.bounds` method or a bounds argument to `s.freqscope` as well. However, this appears somewhat redundant, since `s.freqscope` always opens its window at the lower-left corner of the display. On occasion, I think it would be useful to control its position, but as things stand, one can simply use:
`s.freqscope.window.bounds_(Rect(0.0, 48, 604.0, 328.0))` if needed.

Thus, with the exception of `s.freqscope`, other server GUI tools either have or should have arguments or methods to control their position on screen.

With the introduction of the new class (provisionally named `ServerTreeView`), it will be possible for users to create multiple instances of these plotting GUIs, each potentially with its own bounds. However, I suspect that most users will not require multiple instances. While testing `ServerTreeView`, I noticed:
- The server can produce sound late, with messages such as `late 0.08973143`.
- It is theoretically possible to allow each instance to control its own interval, but at present this is not implemented, and pursuing it could result in performance issues alongside the aforementioned latency problem.

Therefore, while the underlying class supports bounds, I believe it is appropriate to retain the bounds argument, as it provides the expected flexibility for users without introducing unnecessary complexity for the majority.
	
Please compare the following two examples to see which is more consistent and simpler.
```
FreqScope(server: s)
ServerMeter(s, 2, 2).position_(0@360)
Stethoscope(s).bounds_(Rect(140, 360, 264, 264))
ServerTreeView(s, Rect(0, 400, 400, 800))
```

```
s.freqscope
s.meter(position: 0@360)
s.scope.bounds_(Rect(140, 360, 264, 264))
s.plotTree(s, Rect(0, 624, 405, 630))
ServerTreeView(s, Rect(0, 400, 400, 800)) // To me, it does not seem appropriate!
```
	
Of course, we need the following code to automate:
```
(
~f = s.freqscope; ~fB = ~f.window.bounds;
s.meter;	// the default valuse should be change to placed above s.freqscope. I will open a PR.
~m = ServerMeterView; ~mW = ~m.getWidth(2, 2);
s.scope.bounds_(Rect(~mW, ~fB.bottom, ~m.height, ~m.height));
~sBb = s.scope.bounds.bottom + 60;
s.plotTree(1, Rect(0, ~sBb, ~m.height, Window.availableBounds.height - ~sBb))
)
```

prko:
Thank you for your rework and  agreement to what I explained,

I just test the current state of the last commit with the following code:
```
s.boot
s.plotTree
s.plotTree(1) // here it does not start. THIS IS OK!
s.serverTreeWindow.close

// THE FOLLOWING THING IS A PROBLEM:
b = s.plotTree // plotting command 1
b.start(1) // plotting command 2
b.close
```
When excuting plotting command 2, plotting command continues its task, and two more commands are added. So, when closing the window assigned to variable b, the plotting command still continues. This is a problem. (I spent an extremely long time resolving this problem...)

Please watch [the demo video](https://www.dropbox.com/scl/fi/elqnlizzg897gqa1qrl3y/plotting-does-not-stop.mov?rlkey=yjnfajo5exu4rjvdsglofb5qj&st=vwhjvze4&dl=0)

HelpSource/Classes/Server.schelp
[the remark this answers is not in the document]
prko: Ah, sorry for misreading! Done in https://github.com/supercollider/supercollider/pull/6834/commits/aedef58d3f5b9fb3940664abc7f33253baa7d146

dyfer:
> s.boot
> s.plotTree
> s.plotTree(1) // here it does not start. THIS IS OK IF YOU WANT THIS!
> s.serverTreeWindow.close
> 
> // THE FOLLOWING THING IS A PROBLEM:
> b = s.plotTree // plotting command 1
> b.start(1) // plotting command 2
> b.close

Both cases should be fixed in https://github.com/supercollider/supercollider/pull/6834/commits/e7ecad06959e7214fc79dc55c27e9b9c55a9a9ac

----

Also I realized that #6294 added `plotTreeWindow` to `Server` and this PR adds `plotTreeView`. I think we only need one - `plotTreeView`. I'll try to implement this.

dyfer:
I changed the implementation to only use `serverTreeView` in the Server class. I also fixed a possible error thrown if the closing would happen when the tree was being updated (I think).

prko:
Thanks for the correction and updating the implementation.
I  updated the schelp files based on your implementation.
(I had also found some typos in `ServerTreeView.schelp`, and this is corrected in https://github.com/supercollider/supercollider/pull/6834/commits/104b7de239450eea9dbc7ff91ab992e0724a415a)

Now, there are some problems:
1. `Group 2` appears by default:
     Please watch [this video](https://www.dropbox.com/scl/fi/ktl9cv2hjcdn843rsulp8/2-groups-by-default.mov?rlkey=5lkc09rpcdeoek4abknuzlqrv&st=9m6kta1m&dl=0)

2. The corresponding node positions for `s.meter`, `s.freqscope` and `s.scope` are wrong.
    Please watch [this video](https://www.dropbox.com/scl/fi/g8wq7ehom1s71f2ugxlu7/wrong-node-positions-of-.meter-.scope-and-.freqscope.mov?rlkey=pkxxew2xfd7kc1bcj2c0dimk5&st=dg3d5ngt&dl=0)

3.  One more thing to be corrected. The following code is from the `.plotTree` help documentation:
    ```
    // Create plotTree window at the default position with default size:
    s.plotTree; 
    
    // Move plotTree window to the defined position with defined size:
    s.plotTree(2, Rect(200, 200, 100, 200));
    // <- The size (width: 100; height: 200) is smaller than the minimum size, so the size will be ajusted.
    
    s.plotTree; // brings the plotTree window to the front without changing its position or size.
    ``` 
    The second code `s.plotTree(2, Rect(200, 200, 100, 200));` does not work now....

dyfer:
I don't think I touched the logic in node drawing in my commits.

> * `Group 2` appears by default:
>   Please watch [this video](https://www.dropbox.com/scl/fi/ktl9cv2hjcdn843rsulp8/2-groups-by-default.mov?rlkey=5lkc09rpcdeoek4abknuzlqrv&st=9m6kta1m&dl=0)

I don't see "Group 2" on my system. Are you sure you're not creating it somewhere else? Also compare with `s.queryAllNodes`.

> * The corresponding node positions for `s.meter`, `s.freqscope` and `s.scope` are wrong.
>   Please watch [this video](https://www.dropbox.com/scl/fi/g8wq7ehom1s71f2ugxlu7/wrong-node-positions-of-.meter-.scope-and-.freqscope.mov?rlkey=pkxxew2xfd7kc1bcj2c0dimk5&st=dg3d5ngt&dl=0)

I don't see this behavior on my system, the nodes meter/scope nodes appear right beneath the rest. Are you sure you are running the correct version of the class library, i.e. this branch with no other changes?

> * One more thing to be corrected. The following code is from the `.plotTree` help documentation:

This was easy fix: https://github.com/supercollider/supercollider/pull/6834/commits/de45a2342a15c08071edbd3352cfa5296055b246

prko:
Thank you for yor update. The test with two schelp files is ok!

---

>I don't see "Group 2" on my system. Are you sure you're not creating it somewhere else? Also compare with s.queryAllNodes.

I found the reason. The recent commit adds the Group 2: See line 42 in `Volume.sc` of #6613:
https://github.com/supercollider/supercollider/blob/c806b9bb6651dfe615e66254b8e62f1f7b8c1623/SCClassLibrary/Common/Control/Volume.sc#L42

I should rewrite the `drawFunc` part in `ServerTreeView.sc`. 
I am now very sleepy, I will do it later with other tests.

dyfer:
Ah, your description wasn't fully pointing to the reason, but now I found it.

#6313 added another group, but it only happens when `maxLogins > 1`, so the reproducer is:
```supercollider
s.options.maxLogins_(2);
s.boot;
s.plotTree;
```

> I should rewrite the `drawFunc` part in `ServerTreeView.sc`. I am now very sleepy, I will do it later with other tests.

No, why? We should list all the nodes on the server, the same as they are reported with `s.queryAllNodes`. I think this PR is functionally sound.

We really need to wrap this PR up! We are planning to do the first release candidate of SC 3.14 next week and there more PRs to finish before the release.

Aside from the second group behavior, which is correct IMO, is there anything left in this PR we need to wrap up?

prko:
>Aside from the second group behavior, which is correct IMO, is there anything left in this PR that we need to wrap up before merging?

No, if it's urgent, you should merge it.
If I resolve the behaviour of the second group after you merge this PR, I will make a new PR.

dyfer:
> If I resolve the behaviour of the second group after you merge this PR, I will make a new PR.

Can you please explain in 1-2 sentences what is the issue? 
We should plot all the groups that are on the server, and that's what we are doing currently, I believe.

prko:
>I think this PR can be merged as is. @prko is that okay?

Yes!

dyfer (approved):
Thanks @prko !

dyfer:
@prko I found some inaccuracies between the docs and the implementation, preparing a PR now.

I wanted to ask about the minimum size: 
1. what is the reason for choosing that size? A little smaller is still usable. I personally don't see a reason for limiting the size, but I wanted to ask what was the process of coming up with that functionality. Of course, beyond certain point the plot is unreadable, but I find it to still be somewhat useful at half the current width, if that's what the user needs/has space for.
2. `s.plotTree(bounds: <newSize>)` when called with the plotTree already open, does not conform to the minimum size.

Before changing it one way or another (either implementation or the docs) I wanted to learn about the reason behind applying the limit with the current values.

prko:
I'm not sure if I've understood your question correctly. I found the following related parts from this PR:

1. bounds.minSize(395@386):
   - `bounds = bounds.minSize(395@386);` in `.plotTree`
   - `bounds.asRect.minSize(395@386)` in `.makeWindow` of `ServerTreeView`
2. `view.bounds = Rect(0, 0, min(window.bounds.width - 4, 394), max(height, tabSize * countSize.value(levels) * 1.06));` in `.plotTree` in `resp` of `ServerTreeView`

I don't remember, as I made it almost a year ago.
Now, I can only assume that the reason is my usual attitude to redesigning.
- I preserve the previous design as much as possible.
- I do not want the text in the box to be clipped.
- I want the box with the widest width to have even margins in each direction.

Regarding point 1:
From these aspects, the minimum width should be the value
- which does not display the horizontal scroll bar.
- This makes the left and right margins even.

The minimum height should also be set to a value that does not display the vertical scroll bar.


Regarding No. 2:
I am currently redesigning this section. Please disregard this part for now.
In order to display every rectangle correctly, the number of nodes to be plotted must be counted before each group is drawn.
Using a min, max factor is not ideal.

prko:
>Due to Group 2, there are still collisions

However, I do not need to fix it if @SimonDeplat redesigns this soon.
Simon Deplat: When will you be able to complete the new design? Sorry for the delayed response. I'm having the same problem as you. I had not thought of doing so many things when opening my PR. However, this seems to be a normal process for an open-source project.

PR #6835, opened 2025-05-10
Convert emacs and vim to an opt-in build flag on linux

capital-G:
## Purpose and Motivation

This PR only affects linux users.

When building SC on linux in the default configuration, one needs emacs as a dependency.
This PR makes emacs an optional dependency and one explicitly needs to turn on the building of the plugin using CMAKE flags.

At the same time I also turned off to build the vim plugin in the default configuration (though I think this didn't need to have vim as a dependency, or maybe this is only due that vim is already present on most systems).

I think this can be "voted" in the dev meeting - I personally think it would be good to have this merged for for 3.14 release.

edit: We still build the plugins in CI, so we don't abandon the build process and get aware of issues early.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/README_LINUX.md
+++ b/README_LINUX.md
@@ -77,7 +77,6 @@ sudo apt-get install \
   libudev-dev \
   pkg-config \
   libncurses5-dev \
-  emacs \
   jackd2
 ```
 
@@ -157,7 +156,6 @@ sudo dnf install \
   fftw-devel \
   alsa-lib-devel \
   libatomic
-sudo dnf install emacs # if building with the sc-el backend (default)
 sudo dnf install libXt-devel 
 ```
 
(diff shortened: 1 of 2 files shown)


jamshark70:
I'd agree to this.

README_LINUX.md
```diff
@@ -249,11 +247,12 @@ cmake ..
...
-> So if you do not have `Emacs` installed, you should add `-DSC_EL=NO` as in the following command:
+> **Example:** It is possible to build an optional [Emacs](https://www.gnu.org/software/emacs/) plugin for SuperCollider.
+> As of 3.14, this plugin is not built in by default and must be enabled manually.
+> To tell CMake if this plugin should also be built as part of the build process, the flag `SC_EL` can be used - i.e. `-DSC_EL=YES`:
 >
 > ```shell
-> cmake -DSC_EL=NO ..
+> cmake -DSC_EL=YES ..
 > ```
 >
```
JordanHendersonMusic, line 257: Should we mention the vim plugin in the help files too?

capital-G: IMO emacs simply acts/acted as an example on build flags and not to describe what alternative editors are available. It used to be the example that most users probably want to use/set, maybe we should choose another one instead now? maybe `DSUPERNOVA`? But maybe this could be another PR? Or what is your opinion on it?

JordanHendersonMusic: Ah I see, that's a separate issue then.

JordanHendersonMusic (approved):

dyfer:
The build system doesn't seem have a separate "representative" in the new council structure. 
There seems to be an agreement this is good to go, who should merge?

capital-G:
> The build system doesn't seem have a separate "representative" in the new council structure. There seems to be an agreement this is good to go, who should merge?

I think in this particular case the "emacs person" of this repo could decide/merge this - but since there seems to be not such a person we can simply ask at the upcoming dev meeting if someone opposes making this from opt-out to opt-in and decide it this way?

I don't think we need every aspect of this codebase "council-fied" as there will always be unclear situations and we should also have/practice a way to move on in such areas.
If an area starts to get more busy someone can probably take charge of it (having a person who is good at CMake would be pretty nice though^^).

elgiano:
@davidgranstrom @madskjeldgaard 
Hello! As we are discussing making scvim an optional build target, we were wondering if scnvim is depending on any of it. It doesn't seem to be the case, but I thought we could ask you directly to double-check

davidgranstrom:
@elgiano Thanks for your consideration. There are no dependencies in the scnvim project that involves scvim.

PR #3539, opened 2018-02-26
Add SequenceableCollection#unixCmdGetStdOut

miguel-negrao:
This adds a method unixCmdGetStdOut to SequenceableCollection similar to the method in String. Internally it uses Pipe#*argv . A new primitive was added to Pipe for this. Also added documentation.

Issue #1738, opened 2015-11-25
unixCmd adds "sh -c" to the command


jamshark70:
`sc_popen()` prevents user code from killing a process launched by unixCmd. Among the user impacts, this means the MP3 quark is broken, with no easy workaround.

common/sc_popen.cpp, lines 53-54:

```
    argv[0] = (char *)"sh";
    argv[1] = (char *)"-c";
```

In POSIXy systems, this executes a command that looks like `sh -c user...command...`. Both OSX and Linux respond by creating a process, with one PID, for "sh." Subsequently, "sh" launches another process for the user's command, with a different PID.

In sclang, `unixCmd` returns the PID to the user. However, this is the PID of `sh`. If you then do `"kill -9 %".format(thePid).unixCmd`, the `sh` process will be killed, _but the user-command process keeps running!_

There is no guaranteed, consistent way to extrapolate from the "sh" PID to the user-command PID -- hence, no reliable way to kill the actual command process.
- If there is no concrete benefit to prepending `sh -c`, then... it's actively doing harm, so take it out.
- If there is a good reason to have it, then we need to find some way to make both PIDs available to the user.

jamshark70:
Example with the MP3 quark:

```
SynthDef(\di2, { |bufnum|
	Out.ar(0, DiskIn.ar(2, bufnum));
}).add;

Dialog.openPanel { |path|
	m = MP3(path);
};

m.start;
-> MP3.start completed (PID 21860)

// In terminal:
$ ps x | grep lame
21860 ?        S      0:00 sh -c /usr/bin/lame --decode --silent --resample 44.1  "/home/dlm/Music/ddw_itch_to_scratch.mp3" /tmp/sc3mp3--1907853434.fifo > /dev/null
21862 ?        S      0:00 /usr/bin/lame --decode --silent --resample 44.1 /home/dlm/Music/ddw_itch_to_scratch.mp3 /tmp/sc3mp3--1907853434.fifo
```

Note here that the real `lame` process is 21862, but unixCmd returned the PID 21860.

```
b = Buffer.cueSoundFile(s, m.fifo, 0, 2, 65536);
a = Synth(\di2, [bufnum: b]);

m.stop;

// In terminal:
$ ps x | grep lame
21862 ?        S      0:00 /usr/bin/lame --decode --silent --resample 44.1 /home/dlm/Music/ddw_itch_to_scratch.mp3 /tmp/sc3mp3--1907853434.fifo
```

[...]

PR #6837, opened 2025-05-10
Fix SequenceableCollection -unixCmd on Windows 

### LINKED ISSUE #6742: unixCmd doesn't return the proper PID on Windows

dyfer:
## Purpose and Motivation

In this PR:
- `ArrayedCollection -unixCmd` is fixed to _not_ run the command through `cmd.exe`, which matches behavior on *nix
  - as a result, a proper PID of the requested command is now returned from this method
- `sc_popen()` is renamed to `sc_popen_shell()` to reflect the behavior of the function.

The rest of the functionality previously present in this PR will be addressed in separate Issues/PRs.

How to test this:
```supercollider
// pid with ArrayedCollection
r = [Server.program, "-u", "57110"].unixCmd; // run scsynth using ArrayedCollection -unixCmd
r.postln; // this should match the scsynth pid in Task Manager
thisProcess.platform.killProcessByID(r); // scsynth should exit
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/common/sc_popen.cpp
+++ b/common/sc_popen.cpp
@@ -24,7 +24,7 @@
 #    include <array>
 #    include <string>
 
-std::tuple<pid_t, FILE*> sc_popen_shell(std::string&& command, const std::string& type) {
+std::tuple<pid_t, FILE*> sc_popen_shell(std::string command, const std::string& type) {
     std::vector<std::string> argv;
     argv.emplace_back("/bin/sh");
     argv.emplace_back("-c");
@@ -166,7 +166,7 @@ std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings, 
     return sc_popen_c(commandLine.data(), type.data());
 }
 
-std::tuple<pid_t, FILE*> sc_popen_shell(std::string&& command, const std::string& type) {
+std::tuple<pid_t, FILE*> sc_popen_shell(std::string command, const std::string& type) {
     command = "cmd /c " + command;
     return sc_popen_c(command.data(), type.data());
 }
--- a/common/sc_popen.cpp
+++ b/common/sc_popen.cpp
@@ -24,7 +24,7 @@
 #    include <array>
 #    include <string>
 
-std::tuple<pid_t, FILE*> sc_popen(std::string&& command, const std::string& type) {
+std::tuple<pid_t, FILE*> sc_popen_shell(std::string&& command, const std::string& type) {
     std::vector<std::string> argv;
     argv.emplace_back("/bin/sh");
     argv.emplace_back("-c");
@@ -166,7 +166,7 @@ std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings, 
     return sc_popen_c(commandLine.data(), type.data());
 }
 
-std::tuple<pid_t, FILE*> sc_popen(std::string&& command, const std::string& type) {
+std::tuple<pid_t, FILE*> sc_popen_shell(std::string&& command, const std::string& type) {
     command = "cmd /c " + command;
     return sc_popen_c(command.data(), type.data());
 }
--- a/common/sc_popen.cpp
+++ b/common/sc_popen.cpp
@@ -163,10 +163,11 @@ std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings, 
         strings.begin(), strings.end(), std::string(),
         [](const std::string& a, const std::string& b) -> std::string { return a + (a.length() > 0 ? " " : "") + b; });
 
-    return sc_popen(std::move(commandLine), type);
+    return sc_popen_c(commandLine.data(), type.data());
 }
 
 std::tuple<pid_t, FILE*> sc_popen(std::string&& command, const std::string& type) {
+    command = "cmd /c " + command;
     return sc_popen_c(command.data(), type.data());
 }
 
@@ -187,7 +188,7 @@ std::tuple<pid_t, FILE*> sc_popen_c(const char* utf8_cmd, const char* mode) {
         return error_result;
     }
 
-    std::wstring cmd = L"cmd /c " + SC_Codecvt::utf8_cstr_to_utf16_wstring(utf8_cmd);
+    std::wstring cmd = SC_Codecvt::utf8_cstr_to_utf16_wstring(utf8_cmd);
 
     current_pid = GetCurrentProcess();
 
--- a/common/sc_popen.h
+++ b/common/sc_popen.h
@@ -37,6 +37,6 @@ std::tuple<pid_t, FILE*> sc_popen_c_argv(const char* filename, char* const argv[
  *
  * This function assumes a UTF-8 encoded, narrow-char string.
  */
-std::tuple<pid_t, FILE*> sc_popen_shell(std::string&& command, const std::string& type);
+std::tuple<pid_t, FILE*> sc_popen_shell(std::string command, const std::string& type);
 std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings, const std::string& type);
 int sc_pclose(FILE* iop, pid_t mPid);
--- a/common/sc_popen.h
+++ b/common/sc_popen.h
@@ -37,6 +37,6 @@ std::tuple<pid_t, FILE*> sc_popen_c_argv(const char* filename, char* const argv[
  *
  * This function assumes a UTF-8 encoded, narrow-char string.
  */
-std::tuple<pid_t, FILE*> sc_popen(std::string&& command, const std::string& type);
+std::tuple<pid_t, FILE*> sc_popen_shell(std::string&& command, const std::string& type);
 std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings, const std::string& type);
 int sc_pclose(FILE* iop, pid_t mPid);
--- a/lang/LangPrimSource/PyrFilePrim.cpp
+++ b/lang/LangPrimSource/PyrFilePrim.cpp
@@ -1377,7 +1377,7 @@ int prPipeOpen(struct VMGlobals* g, int numArgsPushed) {
 
     pid_t pid;
     FILE* file;
-    std::tie(pid, file) = sc_popen(std::move(commandLine), mode);
+    std::tie(pid, file) = sc_popen_shell(std::move(commandLine), mode);
     if (file != nullptr) {
         SetPtr(&pfile->fileptr, file);
         SetInt(callerSlot, pid);
--- a/lang/LangPrimSource/PyrUnixPrim.cpp
+++ b/lang/LangPrimSource/PyrUnixPrim.cpp
@@ -158,7 +158,7 @@ int prString_POpen(struct VMGlobals* g, int numArgsPushed) {
 
     pid_t pid;
     FILE* stream;
-    std::tie(pid, stream) = sc_popen(std::move(cmdline), "r");
+    std::tie(pid, stream) = sc_popen_shell(std::move(cmdline), "r");
     if (stream != nullptr) {
         SC_Thread thread(std::bind(string_popen_thread_func, pid, stream, IsTrue(postOutputSlot)));
         thread.detach();


dyfer:
@Spacechild1 I went down this rabbit hole... Do you have any opinion on the proposed solution?

common/sc_popen.cpp
```diff
@@ -163,10 +163,11 @@ std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings,
         strings.begin(), strings.end(), std::string(),
         [](const std::string& a, const std::string& b) -> std::string { return a + (a.length() > 0 ? " " : "") + b; });
 
-    return sc_popen(std::move(commandLine), type);
+    return sc_popen_c(std::move(commandLine).data(), type.data());
 }
 
 std::tuple<pid_t, FILE*> sc_popen(std::string&& command, const std::string& type) {
+    command = "cmd /c " + command;
     return sc_popen_c(command.data(), type.data());
```
Spacechild1, line 171: `sc_open_c` is just the C-style version of `sc_open`, so they should have the same behavior (= spawn a process with the given arguments). If `cmd /c` is required, it should be explicitly added by the caller.

dyfer: Maybe I'm not understanding correctly? 
Running `String .unixCmd` in a shell is a [behavior defined in SC](https://github.com/supercollider/supercollider/pull/6837/files#diff-a15e8fe708dbb2bcd27256def5a360fbacc152340db2385f013b05af940fb8f8R675). On *nix we do [sh -c](https://github.com/supercollider/supercollider/pull/6837/files#diff-ddec1c4dde7311f60d467ba6918bc5abcfe4b4718638d305850f16d3263984e8R29-R30) so I think `cmd /c` matches the intended behavior on Windows?

Spacechild1: Of course, this would also apply to the Linux version. (`sc_popen` adds `/bin/sh -c` and forwards to `sc_popen_argv`.)

In either case, I would expect `sc_popen` to just run the given command as is.

If you are motivated, it would be cool to fix this on both sides. Otherwise just keep your solution, but maybe add a comment.

Spacechild1: Here's another idea: keep your fix, but rename `sc_popen` to something like `sc_popen_shell` in both implementations. Then it would be clear what the function does :)

common/sc_popen.cpp
```diff
@@ -163,10 +163,11 @@ std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings,
         strings.begin(), strings.end(), std::string(),
         [](const std::string& a, const std::string& b) -> std::string { return a + (a.length() > 0 ? " " : "") + b; });
 
-    return sc_popen(std::move(commandLine), type);
+    return sc_popen_c(std::move(commandLine).data(), type.data());
```
Spacechild1, line 166: Since you are just calling a member function, there's no point in moving the object!

Spacechild1:
> ArrayedCollection -unixCmd is fixed to not run the command through cmd.exe, which matches behavior on *nix
> as a result, a proper PID of the requested command is now returned from this method

Cool!

This might be out of scope to this PR, but maybe you could add cross links between `String-unixCmd` and `SequencableCollection-unixCmd` in the documentation, e.g. "If you want to run a program without a shell, please use link::Classes/SequencableCollection#-unixCmd::" and vice versa.

> String -unixCmd still returns the PID of cmd.exe and not of the specified command, but there doesn't seem to be a straightforward solution for this
> execvp used on *nix for running a command is not available on Windows

`execvp` is used together with `fork` to overlay the *child* process with the desired program. `fork + exec` is basically the Unix equivalent to `CreateProcess`.

However, `sh` in turn executes the given command(s) as child process(es), so I would totally expect that `String -unixCmd` returns the PID of the shell itself, both on Windows and Unix! Are your sure that this isn't the case?

---

Side note: I think it's a pretty strange design that `SequencableCollection-unixCmd` and `String-unixCmd` have different behavior. Personally, I would have given `String-unixCmd` a different name altogether, e.g. `String-shellCmd`.

dyfer:
> This might be out of scope to this PR, but maybe you could add cross links between `String-unixCmd` and `SequencableCollection-unixCmd` in the documentation, e.g. "If you want to run a program without a shell, please use link::Classes/SequencableCollection#-unixCmd::" and vice versa.

Good idea, I think it's a small addition and could go in here.

> > String -unixCmd still returns the PID of cmd.exe and not of the specified command, but there doesn't seem to be a straightforward solution for this
> > execvp used on *nix for running a command is not available on Windows
> 
> `execvp` is used together with `fork` to overlay the _child_ process with the desired program. `fork + exec` is basically the Unix equivalent to `CreateProcess`.
> 
> However, `sh` in turn executes the given command(s) as child process(es), so I would totally expect that `String -unixCmd` returns the PID of the shell itself, both on Windows and Unix! Are your sure that this isn't the case?

I checked on current macOS and I can't see a separate shell and I get the PID of the actual process I'm runnnig.

HOWEVER I remember previously seeing a separate `sh` shell being spun and getting a PID of that! Actually there's an existing issue: https://github.com/supercollider/supercollider/issues/1738

One difference I think was that killing that shell would also kill the command on *nix (I think?), which wasn't the case on Windows before this PR.

> 
> Side note: I think it's a pretty strange design that `SequencableCollection-unixCmd` and `String-unixCmd` have different behavior. Personally, I would have given `String-unixCmd` a different name altogether, e.g. `String-shellCmd`.

Yeah, very strange. Incidentally, the issue describing `sh -c` also links to the SeqCollection-unixCmd addition: https://github.com/supercollider/supercollider/pull/3539

Spacechild1:
> HOWEVER I remember previously seeing a separate sh shell being spun and getting a PID of that!

Yes, that's exactly what I would expect. In fact, I don't really see how `sh -c <command>` could possibly return the PID of `<command>`.

Thanks for the linked issue. I didn't know that `SequenceCollection -unixCmd` was added after `String -unixCmd`. They really should have picked a different name :)

Spacechild1:
> I checked on current macOS and I can't see a separate shell and I get the PID of the actual process I'm runnig.

Can you share the command you used for testing?

dyfer:
> Can you share the command you used for testing?

Sure! 
```supercollider
p = unixCmd("/Applications/SuperCollider_3.14-dev.app/Contents/Resources/scsynth" + s.options.asOptionsString, {"I'm done!".postln});
```

> Yes, that's exactly what I would expect. In fact, I don't really see how `sh -c <command>` could possibly return the PID of `<command>`.

I was also wondering if using `exec` in the command itself would've influenced it (sclang [does add `exec` in front of the scsynth executable](https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Platform/osx/OSXPlatform.sc#L21) on *nix) but that's why I tested the confirm with a command without exec.

Just to re-iterate, I don't see an extra shell after running the command above.

BTW, if I run this in terminal:
```sh
sh -c "/Applications/SuperCollider_3.14-dev.app/Contents/Resources/scsynth -u 57110"
```
I also _don't see a new shell process_. 
Something specific to macOS maybe?

common/sc_popen.cpp
dyfer: I renamed `sc_popen` to `sc_popen_shell`

common/sc_popen.cpp
dyfer: Thanks! I removed `move`.

Spacechild1:
> I was also wondering if using exec in the command itself would've influenced it (sclang [does add exec in front of the scsynth executable](https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Platform/osx/OSXPlatform.sc#L21) on *nix)

That should be obsolete as we are using `SequencableCollection -unixCmd`. Should we go ahead and remove `exec` as part of this PR?

> I also don't see a new shell process in the Activity Monitor.

Interesting! Maybe `sh -c` automatically overlays the process with the last command? I need to check myself.

However, I definitely see the shell process on Linux, as I would expect!

Also, killing the shell process on Linux (with `Platform.killProcessByID` in sclang or `kill` in a terminal) does *not* automatically kill the child process. Contrary to popular believe, Linux does *not* automatically kill child processes when the parent process dies.

I am mentioning this because you've changed the Windows implementation of `Platform.killProcessByID` to kill all child processes. I'm not really against this change, but it should be mentioned in the docs that this behaves slightly different on Windows and Linux. Or could we maybe add a Boolean argument, e.g. `children`, that lets the user control the behavior?

On Linux, killing a process with all its child processes can be done with `pkill <PID>`. Note that the `pkill` command may not be present on some (older) systems, so we would need to check for the existance of `pkill` in `Platform.initClass`...

> EDIT: I realized that I didn't look into systemCmd, I wonder how that behaves regarding PIDs...

`systemCmd` is blocking, so it just returns the error code.

dyfer:
> That should be obsolete as we are using `SequencableCollection -unixCmd`. Should we go ahead and remove `exec` as part of this PR?

Well, I haven't changed server boot to using `SequenceableCollection -unixCmd` in this PR. I think that should be a separate PR and there we could remove `exec`.


> However, I definitely see the shell process on Linux, as I would expect!
> 
> Also, killing the shell process on Linux (with `Platform.killProcessByID` in sclang or `kill` in a terminal) does **not** automatically kill the child process. Contrary to popular believe, Linux does _not_ automatically kill child processes when the parent process dies.

Ah! 
So does this work on Linux?
```supercollider
r = (Server.program + Server.default.options.asOptionsString).unixCmd; 
r.postln; // is this a pid of `sh -c` or of scsynth?
thisProcess.platform.killProcessByID(r); // did the server stop? or just the shell?
```
We have [code like this in the test suite](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestServer_clientID_booted.sc#L162) and it does seem to work in CI... That particular test was the reason I went down chasing this in the first place, as it was freezing the testsuite on Windows.

> 
> I am mentioning this because you've changed the Windows implementation of `Platform.killProcessByID` to kill all child processes. I'm not really against this change, but it should be mentioned in the docs that this behaves slightly different on Windows and Linux. Or could we maybe add a Boolean argument, e.g. `children`, that lets the user control the behavior?
> 
> On Linux, killing a process with all its child processes can be done with `pkill <PID>`. Note that the `pkill` command may not be present on some (older) systems, so we would need to check for the existance of `pkill` in `Platform.initClass`...

Yea, we should clean this up on all platforms.

Curiously on macOS it works as advertised (I get a pid of the command), but let's check on Linux. It seems that the claim of returning PID of the command is not true there?

> `systemCmd` is blocking, so it just returns the error code.

Duh, of course. Thanks!

Spacechild1:
> use ArrayedCollection -unixCmd in the class library, particularly in the Server class

Yes, please! Not only would it fix our particular issue with killing the Server on Windows, we could also get rid of that ugly `exec` command in `Server.program`.

dyfer:
> > The scsynth command does not create a new shell. However, sh -c "yes > /dev/null" does create a new instance of bash. And indeed, killing bash does not kill the yes command.
> 
> Here's a wild guess: if `sh -c <command>` consists of a single command, macOS's shell implementation internally uses `exec` to replace the shell with that command as an optimization. It does not do that if there are multiple commands involved, which kind of makes sense.

Wow! Your guess is absolutely correct. This actually does start a new shell:
```sh
sh -c "/Applications/SuperCollider_3.14-dev.app/Contents/Resources/scsynth -u 57110 > /dev/null"
```
Needless to say, stopping that shell does not stop scsynth.

I'll proceed with other changes.

dyfer:
Now this PR only fixes the `SequenceableCollection -unixCmd`.

common/sc_popen.cpp
```diff
@@ -163,10 +163,11 @@ std::tuple<pid_t, FILE*> sc_popen_argv(const std::vector<std::string>& strings,
         strings.begin(), strings.end(), std::string(),
         [](const std::string& a, const std::string& b) -> std::string { return a + (a.length() > 0 ? " " : "") + b; });
 
-    return sc_popen(std::move(commandLine), type);
+    return sc_popen_c(commandLine.data(), type.data());
 }
 
-std::tuple<pid_t, FILE*> sc_popen(std::string&& command, const std::string& type) {
+std::tuple<pid_t, FILE*> sc_popen_shell(std::string&& command, const std::string& type) {
+    command = "cmd /c " + command;
```
Spacechild1, line 170: Modifying and reassinging to an rvalue reference is a bit funky. I think `sc_popen_shell` should just take `std::string` by value.

Generally, functions should only take arguments by rvalue reference if it *must* be an r-value reference, e.g. to distinguish a copy constructor from a move constructor. Another use case are "unversial references":
```
template<typename T>
void foo(T&& x);
```
Otherwise, just pass by value. The caller can still move into the function and the function itself can move the argument.

Spacechild1:
Cool! Just a little code style comment from my side. Apart from that, this looks good and can be merged, at least from my side.

common/sc_popen.cpp
dyfer: I know this is basic stuff, but it goes beyond my knowledge of cpp. I tried applying the change, is this what you meant?

dyfer: Also, I noticed that `sc_popen_shell()` on *nix uses moves the command string when passing the data to the vector. Is that done "just in case" or is there an actual reason for it?

Spacechild1: > I tried applying the change, is this what you meant?

Yes, thanks!

> Also, I noticed that sc_popen_shell() on *nix moves the command string when passing the data to the vector. Is that done "just in case" or is there an actual reason for it?

This is just a small performance optimization. `std::move` turns a normal reference into a rvalue-reference, so that the destination - here `std::vector::push_back` - knows that it can move from its argument. In case of `std::string` it can steal the underlying buffer instead of making a copy. (Does this actually matter in the context of spawning a subprocess? Probably not :)

capital-G:
I haven't followed the whole thread, but I currently run into some problems using `.unixCmd` on macOS and switched to `.runInTerminal`, but this creates another set of problems...
E.g. `"wget -h".unixCmd(postOutput: true);` will not print anything to stdout, but `"ls".unixCmd(postOutput: true);` will do - seems inconsistent and in fact `wget` will not get executed at all - maybe this is a security feature of macOS as I could run some binaries but not all?

I haven't checked the source code of the popen implementation, but maybe we should switch to [`boost.Process`](https://www.boost.org/doc/libs/1_86_0/doc/html/process.html) in order to outsource the maintenance workload of multi-platform support?
It would be great to e.g. specify cwd or environment vars w/o patching the shell command - this also make multi-platform support easier.

<details>

BTW: TIL there is a difference on unix between using `"` and `'`

E.g.

```
HELLO=42 /bin/zsh -c "echo $HELLO"
```

will return nothing, but

```
HELLO=42 /bin/zsh -c 'echo $HELLO'
```

will return `42` - the difference is that `'` will not be evaluated in the current shell, but `"` will be evaluated and expanded by the surrounding shell.
</details>

dyfer:
> E.g. `"wget -h".unixCmd(postOutput: true);` will not print anything to stdout, but `"ls".unixCmd(postOutput: true);` will do

~~I've had issues with the past with commands not posting, but that seemed to be due to them either flushing or not flushing their output (which I think is inferred from the environment they run in?)~~

Actually, for wget on macOS that's probably an issue with `$PATH`. I'm not sure what's the mechanism of inheriting the environment vars in SC, but I've found that homebrew-installed commands were sometimes not found when trying to run them in SC.

```supercollider
"wget -h".unixCmd({|err| "exit code: ".post; err.postln;}); 
```
^^ this posts 127, which I believe is "command not found".
However, this works on my macOS:
```supercollider
"/opt/homebrew/bin/wget -h".unixCmd;
```

I can't remember how I circumvented this in the past...


In any case, the total rework is maybe desirable, but should be a separate PR.

What I think is REALLY desirable is cleaning up unixCmd/systemCmd String/SequenceableCollection interface, it's quite confusing currently.

dyfer:
@Spacechild1 if this looks good to you, could you approve the PR?

Spacechild1 (approved):

Issue #6824, opened 2025-05-07
Replace "old-style"


dyfer:
In gereneral it seems we should replace all "old-style" connections with their type-safe equivalent https://doc.qt.io/qt-6/qobject.html#connect-4 - plus-points for using lambda functions which would probably make all our interfaces a lot smaller and clearer.

_Originally posted by @capital-G in https://github.com/supercollider/supercollider/pull/6817#pullrequestreview-2820781417_

dyfer:
@capital-G should we keep this closed or reopen it so that we remember to replace the remaining old style connections, like the ones you did not change due to mismatched types?

capital-G:
I hope that once we remove qt5 support we will search for the necessary ifdefs - and all other places can be found easily via regex, so I think it is fine closing this.

I linked it in the drop qt5 support issue https://github.com/supercollider/supercollider/issues/7006

PR #6838, opened 2025-05-11
Make status bar themed

capital-G:
## Purpose and Motivation

A friend asking "why is the status bar not following the theme" resulted in a small hackathon this evening...

Old style
image

New idle
image

New boot
image

New booted
image

New on Monokai theme
image

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -104,7 +104,7 @@ void AudioStatusBox::applySettings(Settings::Manager* settings) {
     mRecordLabel->setBackground(backgroundColor);
 
     // used if e.g. server is not recording is not recording
-    noActionColor = Qt::black;
+    noActionColor = settings->getThemeVal("text").foreground().color();
 
     unresponsiveColor = settings->getThemeVal("postwindowwarning").foreground().color();
     runningColor = settings->getThemeVal("postwindowsuccess").foreground().color();
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -29,6 +29,8 @@
 namespace ScIDE {
 
 AudioStatusBox::AudioStatusBox(ScServer* server, QWidget* parent): StatusBox(parent) {
+    mServer = server;
+
     mStatisticsLabel = new StatusLabel;
     mVolumeLabel = new StatusLabel;
     mMuteLabel = new StatusLabel;
@@ -110,6 +112,12 @@ void AudioStatusBox::applySettings(Settings::Manager* settings) {
     runningColor = settings->getThemeVal("postwindowsuccess").foreground().color();
     notRunningColor = settings->getThemeVal("text").foreground().color();
     errorColor = settings->getThemeVal("postwindowerror").foreground().color();
+
+    if (mServer) {
+        updateVolumeLabel(mServer->volume());
+        updateMuteLabel(mServer->isMuted());
+        updateRecordLabel(mServer->isRecording());
+    }
 }
 
 void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, bool unresponsive) {
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -24,6 +24,8 @@
 #include <QHBoxLayout>
 #include <QWheelEvent>
 
+#include "main.hpp"
+
 namespace ScIDE {
 
 AudioStatusBox::AudioStatusBox(ScServer* server, QWidget* parent): StatusBox(parent) {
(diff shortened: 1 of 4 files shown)


dyfer:
Thanks for all the work! 
It does _look_ better IMO. I wonder if it's actually better for readability now that it blends in better or not so much?

But I tentatively agree that this is a desirable change.

I have 2 comments:
- "M" and "R" buttons seem always black. They are not really readable on darker themes (completely invisible in the "dark" theme)
- this might be about theme handling in general, but specifically for this PR, but did you notice that switching from light to darker themes, e.g. 1. switching to default theme, 2. recompile library, 3. switch to monokai, then some text stays dark? (e.g. "Welcome to SuperCollider" and the server status)
It does get fixed after recompiling again though. Again, this is for another PR, but if we're tweaking appearance maybe this is something to look into?

I think M/R buttons here could be adjusted, otherwise this looks good!

editors/sc-ide/widgets/audio_status_box.cpp
```diff
@@ -87,19 +89,39 @@ AudioStatusBox::AudioStatusBox(ScServer* server, QWidget* parent): StatusBox(par
...
+    mMuteLabel->setBackground(backgroundColor);
+    mRecordLabel->setBackground(backgroundColor);
+
+    // used if e.g. server is not recording is not recording
+    noActionColor = Qt::black;
+
+    unresponsiveColor = settings->getThemeVal("number").foreground().color();
+    runningColor = settings->getThemeVal("symbol").foreground().color();
+    notRunningColor = settings->getThemeVal("text").foreground().color();
+    warningColor = settings->getThemeVal("postwindowerror").foreground().color();
```
dyfer, line 112: totally cosmetic - if we're using the "error" color here, maybe the variable would be better called `errorColor`?

editors/sc-ide/widgets/audio_status_box.cpp
```diff
@@ -87,19 +89,39 @@ AudioStatusBox::AudioStatusBox(ScServer* server, QWidget* parent): StatusBox(par
...
+    mStatisticsLabel->setBackground(backgroundColor);
+    mVolumeLabel->setBackground(backgroundColor);
+    mMuteLabel->setBackground(backgroundColor);
+    mRecordLabel->setBackground(backgroundColor);
+
+    // used if e.g. server is not recording is not recording
+    noActionColor = Qt::black;
+
+    unresponsiveColor = settings->getThemeVal("number").foreground().color();
+    runningColor = settings->getThemeVal("symbol").foreground().color();
```
dyfer, line 110: This works I think, but we have "postwindowsuccess" that may may be more consistent?

editors/sc-ide/widgets/lang_status_box.cpp
```diff
@@ -39,7 +41,27 @@ LangStatusBox::LangStatusBox(ScProcess* lang, QWidget* parent): StatusBox(parent
...
+
+    mLang = lang;
+}
+
+void LangStatusBox::applySettings(Settings::Manager* settings) {
+    backgroundColor = settings->getThemeVal("text").background().color();
+
+    notRunningColor = settings->getThemeVal("text").foreground().color();
+    runningColor = settings->getThemeVal("symbol").foreground().color();
+    startingColor = settings->getThemeVal("number").foreground().color();
```
dyfer, line 59: Similarly to other comments, I'd suggest "postwindowsuccess" and "postwindowwarning" colors here.

capital-G:
> I think M/R buttons here should be adjusted. Additionally I have a few comments regarding the color choice, see below.

Agreed, I applied the theme color "text" to them if they are not in action and "warning" otherwise.

> I think this would work better if this was a "postwindowwarning" color, what do you think?

Yeah - but I am not too sure about the consistency/quality of the postwindow colors on themes (e.g. success prints green on white in the default theme - not great in regards to readability, maybe we should darken the green?) - but went with those colors for now, then it is at least as inconsistent as within the post window ;)

Maybe we should replace those theme key-strings with some enums in general, I always have a bad feeling using strings as constant key references. But this should be a dedicated PR.

> * but did you notice that when switching from light to darker themes (e.g. 1. switch to default theme, 2. recompile library, 3. switch to monokai), some text stays dark (e.g. "Welcome to SuperCollider" and the server status) until recompiling the library again?

Yes - I noticed that sometimes it is hard to re-render stuff after a theme change because the information/values are not available when calling the render function - I therefore made the widgets more stateful in order to be able to re-render them upon request - hopefully this is not a qt sin?

But I think the signaling is a bit convoluted in the IDE? I think using the new syntax (#6824) would help a bit to be a bit more sure about whats going on within the IDE as there are probably some broken or doubled signals that are hard to spot currently.

dyfer:
> > I think M/R buttons here should be adjusted. Additionally I have a few comments regarding the color choice, see below.
> 
> Agreed, I applied the theme color "text" to them if they are not in action and "warning" otherwise.

👍

> > I think this would work better if this was a "postwindowwarning" color, what do you think?
> 
> Yeah - but I am not too sure about the consistency/quality of the postwindow colors on themes (e.g. success prints green on white in the default theme - not great in regards to readability, maybe we should darken the green?) - but went with those colors for now, then it is at least as inconsistent as within the post window ;)

Yes, possibly it would be good to darken the green... but I think it's okay for now.

> I therefore made the widgets more stateful in order to be able to re-render them upon request - hopefully this is not a qt sin?

I wouldn't know, unfortunately...

dyfer (approved):
I think this is good! Feel free to merge.

PR #6842, opened 2025-05-12
help: clarify pid behavior in String -unixCmd

dyfer:
## Purpose and Motivation

The documentation of `String -unixCmd` erroneously stated that it returns the pid of the actual command, which is not always the case.

This came up when discussing #6837; also see #1738 for related previous discussion on using a shell in unixCmd.

I also plan to propose cleaning up the interface of unixCmd / systemCmd, but in a separate PR. For now, let's just make the documentation reflect the current state of things.

Additionally, this PR replaces a few instances of "Unix" with "system" as the commands in question are cross-platform, and cross-links String -unixCmd with SequenceableCollection -unixCmd for discoverability.

## Types of changes

- Documentation

## To-do list

- [ ] Updated documentation


jamshark70:
LGTM.

My only minor, minor (nonessential, ignore it if you like) comment is that more formal English avoids contractions. "Don't" is common enough that it sometimes appears in tech writing (we don't need to do a search-and-destroy for all contractions in the help!), but "might've" strikes me slightly odd in this context -- "For these reasons it might have appeared" may be a tiny bit better -- but not strictly necessary to change.

dyfer:
@tedmoore @HotwheelsSisyphus would you be okay merging this PR?

HelpSource/Classes/String.schelp
```diff
@@ -676,13 +676,29 @@ Despite the fact that the method is called 'unixCmd', strong::it does work in Wi
...
+To get the actual pid of the command, use link::Classes/SequenceableCollection#-unixCmd::, which does not use a shell and returns the actual pid of the command.
+
+Use link::Classes/Integer#-pidRunning:: to test if a process is alive.
+note::
+While the underlying system call will obtaing the pid of the shell, we may get an actual pid of the command in the following situations:
+list::
+## On *nix systems, if the command is run with code::exec::, e.g. code::"exec scsynth -u 57110".unixCmd;::, which causes the calling process (the shell) to be replaced with the new command
+## On macOS specifically, if the command consists of a single executable, where macOS seems the optimize the call by not spawining a new shell
+::
+For these reasons it might have appeared that we are always getting the pid of the actual command, as opposed to the shell, which is what the documentation erroneously stated in the past.
```
tedmoore, line 698: I find this sentence a bit confusing, is there a way to state it more directly? Perhaps:

"Past documentation incorrectly stated that this method always returned the pid of the actual command."

tedmoore (approved):

HelpSource/Classes/String.schelp
dyfer: Thanks! I used your wording, it's much better.
I also added interpunction at the end of the two list entries, is that okay?

PR #6847, opened 2025-05-14
Add `scroll-margin-top` to fix scrolling to anchor

capital-G:
Merge note: Squash it

## Purpose and Motivation

When navigating to a method one "overshoots" currently b/c the height of the menu bar is not respected - e.g.

image

This PR fixes this by providing a `scroll-margin-top` to every anchor

image

I hope we are not using something else than `a` for anchors...

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/static/scdoc.css
+++ b/HelpSource/static/scdoc.css
@@ -1,3 +1,7 @@
+:root {
+    --menu-bar-height: 33px;
+}
+
 *, *::before, *::after {
     box-sizing: inherit;
 }
@@ -46,6 +50,12 @@ table table {
 p {
     margin-top: 1em;
     margin-bottom: 0.3em;
+}
+
+a {
+    /* The menu bar takes some pixels which we need to respect
+    when jumping to an anchor. */
+    scroll-margin-top: var(--menu-bar-height);
 }
 
 a:link, a:visited, a:link:hover {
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks! This works well.

However, I needed to run `SCDoc.renderAll` to see the difference. I haven't looked closely into help rendering - do we have any way of marking it as "dirty" forcing a re-render?

capital-G:
> Thanks! This works well.
> 
> However, I needed to run `SCDoc.renderAll` to see the difference. I haven't looked closely into help rendering - do we have any way of marking it as "dirty" in this PR forcing a re-render on the user's end?

Yes, as described here

https://github.com/supercollider/supercollider/blob/b86371764eca48eafaccf6e4cbed7cd0f42560d8/SCClassLibrary/SCDoc/SCDoc.sc#L1-L23

I could bump it @

https://github.com/supercollider/supercollider/blob/b86371764eca48eafaccf6e4cbed7cd0f42560d8/SCClassLibrary/SCDoc/SCDoc.sc#L384-L387

but I thought as it only affects dev builds we don't need to bump it? But can do.

dyfer:
> Yes, as described here

Thanks, I remember seeing that, but I wasn't sure if that would be applicable in this case. 

> it only affects dev builds

Ah, I see. This is because we already bumped it in https://github.com/supercollider/supercollider/pull/6627. In that case yeah, no need to bump it again.

dyfer (approved):

dyfer:
Thanks!

Issue #6844, opened 2025-05-13
Cannot force close unresponsive server on Windows


dyfer:
## Environment

* SuperCollider version: 3.14-dev
* Operating system: Windows
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
s.boot;
s.pid; // check pid - it's not the pid of the server
thisProcess.platform.killProcessByID(s.pid); // doesn't work on Windows
```

## Expected vs. actual behavior

We should be able to close an unresponsive server with `thisProcess.platform.killProcessByID`. This currently doesn't work on Windows, see [discussion here](https://github.com/supercollider/supercollider/pull/6837#issuecomment-2873693969).

Once #6837 is merged, I'll provide a PR to fix this.

dyfer:
Another option to address this would be to close child processes by default when calling `killProcessByID`. This is what [#6849](https://github.com/supercollider/supercollider/pull/6849) implements and it would fix this issue as well.

PR #6849, opened 2025-05-15
Add option to close subprocesses using killProcessByID

dyfer:
## Purpose and Motivation

When originally working on 6837 - see https://github.com/supercollider/supercollider/pull/6837#issuecomment-2871776697 - I realized that we don't have an option of killing a process along with its subprocess, e.g. when running `"scsynth -u 57110".unixCmd`, which _would_ start a shell and that shell would start scsynth (note that this is _not_ what's happening by default on *nix when booting the server, because we use `exec scsynth` command instead).

In any case, I thought that it would be good to have the option to terminate subprocesses of a given process. Additionally, I added an option to send a `TERM` instead of a `KILL` signal.

This PR **changes the default behavior** by terminating both parent and children processes. I was going back and forth whether that should be the case... I'm leadning towards keeping this behavior, as it would ensure that an unresponsive server is terminated on Windows, even when it's started with `String -.unixCmd` (this would be an indirect fix to #6844), though that issue would be better addressed by moving server boot command to `SequenceableCollection -unixCmd`, which avoids a shell and returns a proper pid (I plan to submit that PR once #6837 is merged).

I'm open to opinions on this functionality and whether terminating subprocesses should be on by default.

I'll keep this PR as a draft until I (or someone else) tests it on all platforms.

## Types of changes

- Documentation
- Bug fix (indirectly)
- New feature

## To-do list

- [ ] Code is tested
  - [ ] tested on macOS
  - [ ] tested on Linux
  - [ ] tested on Windows
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
FWIW the test suite, which does use `killProcessByID` passes on macOS and Linux. Additionally, I rebased #6763 and Windows tests that use `killProcessByID` also seem to work. Some manual testing would still be desirable, but on the surface this PR seems to work.

elgiano:
Yes, sorry i didn't get around to formulate a report yet! I'll do it soon!

elgiano:
Sorry it took a while, it's quite a confusing subject, and I'm indeed still confused about it.
About the implementation, it looks good to me, as we can't be sure a Linux system ships pkill.

I couldn't get certain info about the behavior on linux: on my system, bash often optimizes -c by calling exec itself, so the shell gets substituted by the command. But since I don't know how to predict this behavior (e.g. `yes > /dev/null` forks instead), I think it's good to have the option of killing subprocesses as well.

If the intention of this PR is to make up for not being able to get the child process pid when running in a shell, I would consider making `subprocesses` default to false, so not to change existing behavior, and passing true when we need to (e.g. in tests, do we still need it though?). I can imagine in the future we might be able to distinguish between a process and its subprocesses without having to deal with the enclosing shell? In that case we wouldn't want to set `killProcessByID` to kill subprocesses by default, would we?

dyfer:
> I couldn't get certain info about the behavior on linux: on my system, bash often optimizes -c by calling exec itself, 

I believe if you add a redirection to a command, it won't optimize: e.g. add `> /dev/null` at the end of the server command. YMMV.

My intention behind setting this to true was that we can't be sure whether a command run through `.unixCmd` starts the subprocess or not (depending on system behavior etc) and I'd argue that killing subprocess would be the expected behavior. But I agree it does change the default...
Where this comes up practically is that on Windows there's no way to have the `exec` behavior so a process is always a subprocess. So the code suggested in https://github.com/supercollider/supercollider/blob/c806b9bb6651dfe615e66254b8e62f1f7b8c1623/HelpSource/Classes/Platform.schelp#L183-L186 doesn't work, and I believe it should by default.
If we really don't want this enabled by default on *unix, then I think we should at least have it enabled on Windows, because of the example linked (so there's for sure code out there using this), but I also don't like differentiating defaults for each platform and I don't see harm in terminating subprocesses, as I'd expect that by default.

There's some discussion on this in https://github.com/supercollider/supercollider/issues/1738

What are the scenarios when one would not expect the subprocess to be terminated by default BTW?

elgiano:
I think I was being over-cautious here. For sure it makes sense that if we want it on by default on Windows, we also make it on by default on unix.
I don't know if and how people call .killProcessByPid in their code (I tend to use runInTerminal for long lasting processes). But I imagine if one would spawn a process with unixCmd, and then wanted to terminate it by PID, it would be reasonable to terminate its eventual subprocesses as well.

So, sorry for holding this up!

elgiano (approved):
Tested it on linux, it works as expected:
```supercollider

// subprocess not terminated with subprocesses:false
p ="scsynth -u 57110 > /dev/null". // 96347
thisProcess.platform.killProcessByID(p, subprocesses:false)
"pgrep scsynth".unixCmdGetStdOut // 96353

Server.killAll; // clean-up

// subprocess terminated with subprocesses:true (default)
p ="scsynth -u 57110 > /dev/null". // 97243
thisProcess.platform.killProcessByID(p)
"pgrep scsynth".unixCmdGetStdOut // (empty)
```

dyfer:
Thanks for the review!

PR #6850, opened 2025-05-15
GHA: use --include-path in the testsuite

dyfer:
## Purpose and Motivation

This is a small script simplification enabled by #6640

## Types of changes

- Maintenance

## To-do list

- [x] Code is tested
- [ ] All tests are passing

### DIFF
--- a/.github/workflows/test_linux.yml
+++ b/.github/workflows/test_linux.yml
@@ -17,7 +17,6 @@ jobs:
       TEST_LIST_PROTO: ${{ github.workspace }}/testsuite/scripts/test_run_proto_gha.scd
       TEST_LIST_RESULT: ${{ github.workspace }}/testsuite/scripts/run/gha_result.scxtar
       SCRIPT_RUN_TESTS: ${{ github.workspace }}/testsuite/scripts/sclang_test_runner.scd
-      SCRIPT_ADD_PATH: ${{ github.workspace }}/testsuite/scripts/add_include_path.scd
       SCRIPT_PRINT_RESULTS: ${{ github.workspace }}/testsuite/scripts/print_results.scd
 
     steps:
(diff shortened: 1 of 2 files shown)


capital-G (changes requested):
good idea, but maybe we should also delete

https://github.com/supercollider/supercollider/blob/eb4a3a4e76b399fbce94e1b67e42e07abaebc3ac/.github/workflows/test_linux.yml#L20

(and its respective macOS file) and

https://github.com/supercollider/supercollider/blob/eb4a3a4e76b399fbce94e1b67e42e07abaebc3ac/testsuite/scripts/add_include_path.scd

in general. at least I don't think we need those around anymore after this?

dyfer:
Ah yes, good catch!

dyfer:
With the [include path](https://github.com/supercollider/supercollider/blob/eb4a3a4e76b399fbce94e1b67e42e07abaebc3ac/testsuite/scripts/add_include_path.scd), I'm unsure if I want to remove it; it is still useful if you want to add the testsuite path locally. What do you think?

capital-G:
> With the [add_include_path.scd](https://github.com/supercollider/supercollider/blob/eb4a3a4e76b399fbce94e1b67e42e07abaebc3ac/testsuite/scripts/add_include_path.scd), I'm unsure if I want to remove it; it is still useful if you want to add the testsuite path locally. What do you think?
 
I hope nobody
* runs the script in a sclang interpreter, passing the path of the class library as an argument, which then
* will create a sclang_conf, which then can be used by
* restarting the interpreter to then
* run the test suite

I think at this point people could simply create a shell script like in the CI which can be invoked to run the test suite :)

Or simply use the single command `cmake --target run-test-suite` of #6802 ;)

dyfer:
> I hope nobody

No no! You can run this file in scide and then it'll just add the testsuite (automatically the correct path) to the current config file. I think that's useful, maybe?

capital-G:
> > I hope nobody
> 
> No no! You can run this file in scide and then it'll just add the testsuite (automatically the correct path) to the current config file. I think that's useful, maybe?

Ah I didn't knew that - maybe we should add at some point a `README` to the testsuite library in order to give people some documentation on how to run/use the testsuite, e.g. like this.

In this case this PR is fine by me, feel free to merge @dyfer

capital-G (approved):

dyfer:
Yeah, the never-ending need for docs, I agree!

PR #6854, opened 2025-05-16
Add line number switch to SCDoc theme settings

capital-G:
merge note: squash on merge

Supersedes #6207 with a simpler implementation. Please review and test @prko :)

## Purpose and Motivation

w/o line numbers (default)
grafik

w/ line numbers
grafik

Not the most beautiful implementation, but the docs template will hopefully be rewritten soon though^^

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/static/codemirror.css
+++ b/HelpSource/static/codemirror.css
@@ -202,7 +202,6 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket {color: #a22;}
 .CodeMirror-gutters {
   position: absolute; left: 0; top: 0;
   min-height: 100%;
-  z-index: 3;
 }
 .CodeMirror-gutter {
   white-space: normal;
@@ -213,19 +212,16 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket {color: #a22;}
 }
 .CodeMirror-gutter-wrapper {
   position: absolute;
-  z-index: 4;
   background: none !important;
   border: none !important;
 }
 .CodeMirror-gutter-background {
   position: absolute;
   top: 0; bottom: 0;
-  z-index: 4;
 }
 .CodeMirror-gutter-elt {
   position: absolute;
   cursor: default;
-  z-index: 4;
 }
 .CodeMirror-gutter-wrapper ::selection { background-color: transparent }
 .CodeMirror-gutter-wrapper ::-moz-selection { background-color: transparent }
--- a/HelpSource/static/codemirror.css
+++ b/HelpSource/static/codemirror.css
@@ -24,8 +24,7 @@
 /* GUTTER */
 
 .CodeMirror-gutters {
-  border-right: 1px solid #ddd;
-  background-color: #f7f7f7;
+  background-color: transparent;
   white-space: nowrap;
 }
 .CodeMirror-linenumbers {}
--- a/HelpSource/static/editor.js
+++ b/HelpSource/static/editor.js
@@ -174,12 +174,11 @@ function getLineNumberStorageValue() {
 function setLineNumberStorageValue(v) {
     window.localStorage.setItem("showLineNumbers", v ? "true" : "false");
     toggleLineNumbers(v);
-
 }
 
 function toggleLineNumbers(v) {
-    Array.from(document.querySelectorAll('textarea')).filter((t) => t.hasOwnProperty("editor")).forEach((textarea) => {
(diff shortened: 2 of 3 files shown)


prko:
Thank you for submitting this PR.

This submission is notably more concise than PR #6207. However, there are a few differences that merit discussion:

1. **Appearance:** As demonstrated in [the demo video linked in PR #6207](https://github.com/supercollider/supercollider/pull/6207#issuecomment-2148536975), I have employed a radio button to indicate the current state. I believe that this radio button approach is particularly user-friendly.

2. **Separator Usage:** In this PR, there is no separator between the line number and the code portion. In contrast, [#6207](#6207) includes a separator, as shown in the aforementioned demo video.

3. **Left Padding:** The left-hand padding of the line numbers in this PR is wider than that in [#6207](#6207), as observed in the aforementioned demo video. While I feel that this might lead to a less efficient use of space, I recognise that it could be considered a more aesthetically pleasing choice depending on one’s personal preference.

4. **Navigation History:** The aforementioned demo video for [#6207](#6207)illustrates that the navigation history remains undisturbed when toggling the display of line numbers—the current view does not change. Regrettably, in this PR, activating or deactivating the line numbers resets the display to the top of the page, which I believe to be a bug. Although it is unlikely that many users would modify this setting repeatedly, the operational expectation would be for the previous state to be maintained.

While the differences in appearance (points one to three) might simply reflect personal preference, the issue with the navigation history (point four) directly affects usability.

In PR [#6207](#6207), I made a concerted effort to incorporate the feedback of all reviewers, including your valuable input and my thoughts. It appears that this PR does not include some of the detailed recommendations from my side. Would it be possible either to integrate the insights from PR #6207 into this submission or to consider reviewing and merging PR #6207 with modifications limited solely to the incorrect aspects?

capital-G:
@prko Thanks for the feedback, I hope I addressed all your concerns now (but I left the padding at the default settings of CodeMirror which seems reasonable to me and is in the end personal preference)

Can you give it another go? :)

grafik

prko:
I tried the last update on Windows 11 and macOS 14.7.6.
It works properly now.
Activating/deactivating the showing line number column does not move the page to the top.
I do not care about the design.
The only thing I do not completely agree with is the left-right shift of the code part by activating/deactivating the showing of the line number column. However, it will not disturb many users because a user would not change this. For example, I will always use the line number column. So I think it is ok.

However, there are other problems:

- The problem reported in [#6862](#6862) appears more frequently:
   It happens extremely often. I should 5-times try to launch scide.exe on Windows. 
- Clicking 'Browse' shows nothing and prevents the help browser from working properly (on macOS/Windows)
- Clicking 'Search' does not return a search result, but continues searching and prevents the help browser from working properly (on macOS/Windows)
- Clicking 'Indexes/Documents' shows nothing, and prevents the help browser from working properly (on macOS/Windows)
- Clicking 'Indexes/Classes' shows nothing, and prevents the help browser from working properly (on macOS/Windows)
- Clicking 'Indexes/Methods' shows nothing, and prevents the help browser from working properly (on macOS/Windows)

prko:
I tested this PR's change using [17abf35] on Windows. The following problems still exist:

>The followings show nothing and prevent the help browser from working properly (on macOS/Windows)
>- Clicking 'Browse'
>- Clicking 'Search'
>- Clicking 'Indexes/Documents'
>- Clicking 'Indexes/Classes'
>- Clicking 'Indexes/Methods'

I think there are collisions between the scripts.

prko:
Would the following video be helpful to debug?
[demo video](https://www.dropbox.com/scl/fi/pkrkw6hq97e742467rsdu/video-for-debugging.mov?rlkey=ulml9wuksrl6gzgf2hy1sk3tv&st=dhq1mhcf&dl=0)

capital-G:
> Would the following video be helpful to debug? [demo video](https://www.dropbox.com/scl/fi/pkrkw6hq97e742467rsdu/video-for-debugging.mov?rlkey=ulml9wuksrl6gzgf2hy1sk3tv&st=dhq1mhcf&dl=0)

Thanks, this was indeed very helpful. I think I fixed it, would you mind give it another try? :) I think we are getting closer ;)

The problem was that not all pages load both JS files... We really need a proper template solution soon :)

prko:
It is resolved! Thank you!!!

tedmoore (approved):

Issue #6839, opened 2025-05-11
[windows] 3.14.0-dev, scsynth and supernova don't start if any device is specified in ServerOptions


sternsc:
When testing 3.14.0-dev on Windows 11 I find that the server does not boot, 
```
Booting server 'localhost' on address 127.0.0.1:57110.
Server 'localhost' exited with exit code 1.
```
unless I remove my ASIO driver (for my RME interface) from the startup file.

`// Server.local.options.device = "ASIO Fireface USB";`

I can access the interface using the MME API's but not ASIO. 
3.13.0 still boots fine and works with ASIO. Any ideas? 

System: 
Windows 11 Pro, Version 24H2, OS build 26100.3775
Processor	Intel(R) Core(TM) i7-14700   2.10 GHz
Installed RAM	32,0 GB (31,6 GB usable)
System type	64-bit operating system, x64-based processor

dyfer:
Can you provide the output from the post window after running `s.boot` without any ServerOptions specified?
Also, do you specify any other ServerOptions, like samplerate or buffer size? Also, what sample rate is Fireface USB set to? Is to clocked externally by any chance?

sternsc:
The Fireface 802 clock is set to internal, 44100 Hz, With all references to Server and Server.options... commented out of the startup.scd, this is what I get on a recompile and s.boot: 

```
compiling class library...
	Found 869 primitives.
	Compiling directory 'C:\Users\Sten\Desktop\SuperCollider_develop\SCClassLibrary'
	Compiling directory 'C:\ProgramData\SuperCollider\Extensions'
	Compiling directory 'C:\Users\Sten\AppData\Local\SuperCollider\Extensions'
	numentries = 1366718 / 23959530 = 0.057
	6615 method selectors, 3622 classes
	method table size 22931624 bytes, big table size 191676240
	Number of Symbols 17987
	Byte Code Size 641693
	compiled 666 files in 0.25 seconds

Info: 5 methods are currently overwritten by extensions. To see which, execute:
MethodOverride.printAll

compile done
localhost : setting clientID to 0.
internal : setting clientID to 0.
Class tree inited in 0.01 seconds
Sonic alerts are activated.


*** Welcome to SuperCollider 3.14.0-dev. *** For help press Ctrl-D.
-> localhost
Booting server 'localhost' on address 127.0.0.1:57110.

Device options:
  - MME : Microsoft Sound Mapper - Input   (device #0 with 2 ins 0 outs)
[...]

PR #6855, opened 2025-05-16
classlib: use `SequenceableCollection -unixCmd`

dyfer:
## Purpose and Motivation

As discussed in https://github.com/supercollider/supercollider/pull/6837#issuecomment-2873704883, it would be preferred to use `SequenceableCollection -unixCmd` when running the server.

Aside from the changing the call from `String -unixCmd` to `SequenceableCollection -unixCmd`, I needed to rework `ServerOptions -asOptionsString` method, replacing it with `-asOptionsArray`. However, I kept `-asOptionsString` for backwards compatibility; as it now processes the array from asOptionsArray, the quoting mechanism is more fragile (it is not called explicitly for specified parameters, but is based on the contents of each cmd line parameter, i.e. whether it contains a path separator, or a space). This affects handling of ugen Plugins Path and device names.

Needs testing on Windows and Linux. Seems to work fine on macOS.

Fixes: https://github.com/supercollider/supercollider/issues/6844, https://github.com/supercollider/supercollider/issues/6839

## Types of changes

- Bug fix
- Breaking change (hopefully not! but using `ServerOptions -asOtionsString` needs to be tested...

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/Classes/Platform.schelp
+++ b/HelpSource/Classes/Platform.schelp
@@ -181,7 +181,9 @@ argument:: pid
 an Integer which is the pid of the process to kill.
 code::
 // start a server program from the cmdLine for testing
-~pid = unixCmd(Server.program + s.options.asOptionsString(57100));
+// note that we are using SequenceableCollection -unixCmd instead of String -unixCmd
+// because the latter may not return the pid of the process
+~pid = unixCmd([Server.program] ++ s.options.asOptionsArray(57100));
 // kill just that server program by its pid:
 thisProcess.platform.killProcessByID(~pid);
 ::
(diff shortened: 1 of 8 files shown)


dyfer:
While testing this on Windows, I found that https://github.com/supercollider/supercollider/pull/6837 is not fully correct; i.e. the Windows implementation would still split arguments using spaces. 
```supercollider
a = [Server.program, "-u 57110"].unixCmd; // currently works on Windows, while it should not;  [Server.program, "-u", "57110"] should be required
```

I need to fix that first...

dyfer:
https://github.com/supercollider/supercollider/issues/6856 is merged, but I need to re-thing handling arguments as array vs as string in the most optimal way...

PR #6857, opened 2025-05-16
Set background color of scope via sclang

capital-G:
## Purpose and Motivation

Allows to set the background color of a scope, see https://scsynth.org/t/accessing-parameters-from-qtscopeshm-in-sclang/11708

The C++ side already implemented the setting of background color

https://github.com/supercollider/supercollider/blob/9028422a145230c322cfc923e3c1a41fc4a1b998/QtCollider/widgets/QcScopeShm.h#L69-L72

https://github.com/supercollider/supercollider/blob/9028422a145230c322cfc923e3c1a41fc4a1b998/QtCollider/widgets/QcScopeShm.h#L50

but the sclang side lacked this implementation. This PR fixes this.

grafik

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/ScopeView.schelp
+++ b/HelpSource/Classes/ScopeView.schelp
@@ -83,6 +83,12 @@ METHOD:: waveColors
     argument::
         An Array of Colors, one per channel.
 
+METHOD:: background
+    The background color of the scope.
+
+    argument::
+        A link::Classes/Color::.
+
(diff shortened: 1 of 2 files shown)


capital-G:
This is actually not the way to do this as this is already handled by its superclass `View`...

PR #6858, opened 2025-05-17
Fix unixCmd quoting on Windows

### LINKED ISSUE #6839: [windows] 3.14.0-dev, scsynth and supernova don't start if any device is specified in ServerOptions

### LINKED ISSUE #6843: [Windows] `String -unixCmd` doesn't support multiple pairs of quotes
## Environment

* SuperCollider version: 3:14-dev

## Steps to reproduce

It seems that we don't handle more than one set of quotes in a single String -unixCmd command in Windows:

```supercollider
s.options.device = "anything"; // any string, a default device will be selected
(Server.program ++ s.options.asOptionsString(57110)).unixCmd; // the command doesn't start

(Server.program ++ s.options.asOptionsString(57110)).postln; 
// -> "C:\Users\Marcin\Desktop\SuperCollider\scsynth.exe" -u 57110 -a 1024 -i 2 -o 2 -H "anything" -R 0 -C 1 -l 1
```
Note two sets of quotes, one around the command, one around device name
Copy path from the post window into cmd.exe and run it - it works on my machine:
```cmd
C:\Users\Marcin>"C:\Users\Marcin\Desktop\SuperCollider\scsynth.exe" -u 57110 -a 1024 -i 2 -o 2 -H "anything" -R 0 -C 1 -l 1

Device options:

(...)

SC_AudioDriver: sample rate = 44100.000000, driver's block size = 64
SuperCollider 3 server ready.
```

The command should run, as it does inside cmd.exe.

I'm assuming this is an issue with the underlying sclang primitive, but I'm not sure.

### LINKED ISSUE #6856: SequenceableCollection -unixCmd on Windows
## Environment

* Operating system: Windows

## Steps to reproduce

```supercollider

a = [Server.program, "-u 57110"].unixCmd; // currently works on Windows, while it should not
a = [Server.program, "-u", "57110"].unixCmd; // this is the proper syntax
a = [Server.program, "-u", "57110", "-H", "device name with spaces", "device name with spaces"].unixCmd; // this is where it becomes an issue

```

After https://github.com/supercollider/supercollider/pull/6837 introduced a `SequenceableCollection -unixCmd` implementation of Windows without a shell, it turns out that the underlying primitive concatenates arguments using spaces. This results in incorrect parsing of arguments  with spaces and makes the implementation inconsistent with *nix.

dyfer:
## Purpose and Motivation

This PR:
- introduces proper quoting for `SequenceableCollection -unixCmd` so that the implementation matches the behavior on *nix
- adds quotes to `String -unixCmd` so that the command is properly parsed even if it contains multiple quotes inside

Fixes #6856
Fixes #6843 
Fixes #6839

Please review carefully as this is not my area of expertise. Thanks!

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
@Spacechild1 would you be able to take a look at this?

dyfer:
For reference, here's boost code that seems to handle escaping CreateProcess string: https://github.com/boostorg/process/blob/develop/src/windows/default_launcher.cpp#L41

dyfer:
TODO: add C++ unitTest for `quoteWindowsArg`

scztt:
Other than test, lgtm

scztt (approved):
At a glance, adding cpp unit tests would be a pain. I wonder if we can do something like call an simple "echo" command with some complex quoting and see if we get the right output?

dyfer:
Working on adding tests now.

dyfer:
I added tests for quoting with String -unixCmd and for argument parsing for SequenceableCollection -unixCmd. For the latter I couldn't figure out how to pass arguments to an "inline" script in cmd.exe so I'm writing a tiny .bat file for testing.

I added this PR to the https://github.com/supercollider/supercollider/pull/6763, you can see the result here.

I was also trying to write a test where I'd pass arguments with quotes to SequenceableCollection -unixCmd, but honestly I don't know what the desired result should be really. SequenceableCollection -unixCmd needs to escape quotes and such escaping is then shown when echoing arguments in the script. 

<details>

```
	test_unixCmd_basicQuotes {
		var str1 = "\"Hello World\"";
		var res1 = if(thisProcess.platform.name == \windows) {str1.quote} {str1};
		var out1;
		var cmd = ["echo"];
		if(thisProcess.platform.name == \windows) {cmd = ["cmd", "/c"] ++ cmd};
		out1 = (cmd ++  [str1]).unixCmdGetStdOut.replace("\n", "");
		this.assertEquals(out1, res1, "double quotes result asCompileString: %".format(out1.asCompileString));
	}
// result on Windows: "\"Hello World\"" 

	test_unixCmd_mixingQuotes {
		var str1 = "\"This is 'quoted' inside double quotes\"";
		var res1 = if(thisProcess.platform.name == \windows) {str1.quote} {str1};
		var out1;
		var cmd = ["echo"];
		if(thisProcess.platform.name == \windows) {cmd = ["cmd", "/c"] ++ cmd};
		out1 = (cmd ++  [str1]).unixCmdGetStdOut.replace("\n", "");
		this.assertEquals(out1, res1, "mixed 'quotes' result asCompileString: %".format(out1.asCompileString));
	}
// result on Windows: "\"This is 'quoted' inside double quotes\"" 

	test_unixCmd_multipleQuotes {
		var str = "\"Hello\" \"World\"";
		var res = str;
		var out;
		var cmd = ["echo"];
		if(thisProcess.platform.name == \windows) {cmd = ["cmd", "/c"] ++ cmd};
		out = (cmd ++  [str]).unixCmdGetStdOut.replace("\n", "");
		this.assertEquals(out, res, "multiple quotes result asCompileString: %".format(out.asCompileString));
	}
// result on Windows: "\"Hello\" \"World\"" 

```
</details>

The above tests are not included in this PR; I'm not sure what the expected behavior should be in these cases. Any advice would be appreciated!

dyfer:
I'm feeling pretty good about these tests now.

> **Tl;DR:** This PR adds quotes around arguments with spaces and internal quotes within commands/arguments. This quote and escape chars seem to be passed to the underlying commands. Is this how we should be handling arguments with spaces and quotes in Windows, when not running through cmd.exe?

@Spacechild1 do you have any thoughts on this by any chance?

common/sc_popen.cpp
```diff
@@ -156,18 +156,49 @@ static SC_Lock processlist_mutex;
...
 
+std::string quoteWindowsArg(const std::string& arg) {
+    if (arg.empty())
+        return "\"\"";
+
+    bool needQuotes = arg.find_first_of(" \t\"") != std::string::npos;
+    if (!needQuotes)
+        return arg;
+
+    std::string result = "\"";
```
Spacechild1, line 167: I think you have just reinvented  `std::quoted` :) https://en.cppreference.com/w/cpp/io/manip/quoted

Simple example: https://godbolt.org/z/W9G8xehEh

Spacechild1:
Just a general question: are you sure this is backwards compatible? What if the user already passes properly quoted and escaped strings? Shouldn't this be controlled with an argument instead?

dyfer:
> Just a general question: are you sure this is backwards compatible?

I'm certainly not! :)

But I don't think having more than one quote worked at all previously. 

If someone had `[].unixCmd` working before #6837, then it's already "broken" after this PR, but the previous behavior was not correct anyway, so I don't think there was a case where user had arguments properly quoted.

My assumption is that the interface of `[].unixCmd` should not require the user to quote the args (matching the behavior on *nix). But should we check whether the arg is already quoted and pass it through if that's the case?

common/sc_popen.cpp
dyfer: Thanks! This is exactly the kind of feedback I was hoping for. I'll try it and run against the tests I added here.

BTW the basic assumption in tests is that echo on Windows outputs the argument exactly as it was passed to the command, including escape chars. Is this what a Windows app expects, when it's supposed handle an argument with spaces or quotes?

Spacechild1:
> EDIT: now I see that this PR certainly breaks previous behavior, see the implemented fix to make openOS work properly

I guess that one should really use `String.unixCmd` instead of `Array.unixCommand`. After all, the former already spawns a console for us.

> But should we check whether the arg is already quoted and pass it through if that's the case? That would maintain backward compatibility, I guess.

It's not perfect, but that's also what I would have suggested. Just one caveat: you wouldn't be able to pass an argument that actually starts and ends with a quote, though.

Personally, I would also be fine to make an exception and just break backwards compatibility.

dyfer:
> "cmd /c echo <str>", would I get the escape backslashes back? Or would they be "swallowed" by `cmd` or `echo`?

Answering myself... it seems that `echo` echoes everything, including backslashes.

Spacechild1:
> Did you mean "\"Hello\" \"World\"" (with external quotes)? Just to double check I understand.

Yes! Sorry.

dyfer:
I reverted the implementation to the previously used function.

Tests can be seen running [here](https://github.com/dyfer/supercollider/actions/runs/15261051769/job/42919018252#step:6:11628). The tests are making different assumption on what's the desired result on *nix vs Windows... I hope that's correct?

dyfer:
@Spacechild1 as we had discussions about this implementation, I'd appreciate your approval if you think it works fine.

Spacechild1:
Thanks for the comment and references to the MS documentation! When I initially read this function, it just looked wrong to me. I didn't realized it's because of MS's bizarre quoting rules :)

Spacechild1 (approved):

PR #6865, opened 2025-05-18
scdoc renderer: use forward slash on Windows

dyfer:
## Purpose and Motivation

When creating links on Windows, use `/` instead of `\`. This caused some `TestSCDocHTMLRenderer` tests to fail on Windows.

I haven't follow the logic of SCDocRenderer fully... please review carefully. I'm looking forward to any feedback on this.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested (in #6763)
- [x] All (relevant) tests are passing

### DIFF
--- a/SCClassLibrary/SCDoc/SCDocRenderer.sc
+++ b/SCClassLibrary/SCDoc/SCDocRenderer.sc
@@ -82,6 +82,9 @@ SCDocHTMLRenderer {
 			};
 		};
 
+		// use only forward slash as path separator
+		result = result.replace(Platform.pathSeparator, "/");
+
 		if(linkAnchor.isEmpty) {
 			^result
 		} {


SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -82,6 +82,8 @@ SCDocHTMLRenderer {
 			};
 		};
 
+		if(thisProcess.platform.name == \windows) {result = result.replace("\\", "/")}; // replace backward with forward slash on Windows
+
```
capital-G, line 86: Couldn't we make this somehow more universal via

```
// enforce unix path for html
result = result.replace(Platform.pathSeparator, "/");
```

Platform dependent if statements can lead to some trouble while testing, e.g. invoking methods which have been deleted, but since I would not be on windows the testing suite wouldn't complain.

dyfer:
```
result = result.replace(Platform.pathSeparator, "/");
```
This looks cleaner indeed, but is redundant on Unix. Do you think it's fine to always run it?

> invoking methods which have been deleted

I'm not sure I understand?

dyfer: @capital-G I applied the suggest change.

capital-G (approved):

PR #6867, opened 2025-05-18
SCLang, UnitTests: standardise nan & inf printing across platforms

### LINKED ISSUE #6863: `(0/0).asString` on Windows x86_64 vs other platforms
## Environment

* Operating system: Windows
* Other details (Qt version, audio driver, etc.): observed on x64, but not on arm64

## Steps to reproduce

```supercollider
(0/0).asString; // -nan(ind)

```

`(0/0).asString;` seems to return different result, depending on both  the OS and the hardware platform (maybe the compiler as well? not sure).

Currently observed:

macOS x86_64: `nan`
macOS arm: `nan`
Windows x86_64: **`-nan(ind)`**
Windows arm64 VM (x86_64 binary through an emulator): `nan`
Linux: (haven't tested)

In our testsuite we check whether (0/0).asString is [either `nan` or `-nan`](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestFloat.sc#L47-L51), so the check fails on Windows.

I'm not sure whether `-nan(ind)` is a valid output we should accept, and/or if we should even test for a particular representation of `0/0`, as it seems to differ between platforms.

This issue currently causes one `TestFloat` test to fail on Windows.

JordanHendersonMusic:
## Purpose and Motivation

Closes #6863

Nan & inf should be printed the same across platforms and c library implementations.

This is particularly salient for `inf` as it must be convertible to a compile string and back.

## Types of changes

- Bug fix
- Breaking change ? Maybe if someone relied on this, but it didn't work across platform anyway.

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestFloat.sc
+++ b/testsuite/classlibrary/TestFloat.sc
@@ -44,11 +44,7 @@ TestFloat : UnitTest {
 
 	test_asString_nan {
 		var zeroDivString = (0/0).asString;
-		if(zeroDivString == "-nan") {
-			this.assertEquals(zeroDivString, "-nan")
-		} {
-			this.assertEquals(zeroDivString, "nan")
-		}
+		this.assertEquals(zeroDivString, "nan");
 	}
 
 	test_asString_largeNumberUsesExp {
--- a/testsuite/classlibrary/TestFloat.sc
+++ b/testsuite/classlibrary/TestFloat.sc
@@ -51,4 +51,13 @@ TestFloat : UnitTest {
 		this.assertEquals((1e26).asString, "1e+26")
 	}
 
+	test_asCompileString {
+		this.assertEquals(1.0.asCompileString.compile.(), 1.0, "1.0 as compile string");
+		this.assertEquals(inf.asCompileString.compile.(), inf, "inf as compile string");
+		this.assertEquals(-inf.asCompileString.compile.(), -inf, "-inf as compile string");
+
+		// Since 'nan' isn't a keyword in sc, there is no way to perform this check.
+		// This means the conversion from value to string to value isn't complete over the float type.
+	}
+
 }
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
I've also removed the check for `-nan` because it is unnecessary.

JordanHendersonMusic:
Hmm, according to cpp ref, this could also be an issue for `inf`

>The floating-point conversion functions convert infinity to `inf` or `infinity`. Which one is used is implementation defined.
Not-a-number is converted to `nan` or `nan(char_sequence)`. Which one is used is implementation defined.

testsuite/classlibrary/TestFloat.sc
```diff
@@ -44,15 +44,20 @@ TestFloat : UnitTest {
...
 	}
 
 	test_asString_largeNumberUsesExp {
 		this.assertEquals((1e26).asString, "1e+26")
 	}
 
+	test_asCompileString {
+		this.assertEquals(1.0.asCompileString.compile.(), 1.0, "1.0 as compile string");
+		this.assertEquals(inf.asCompileString.compile.(), inf, "inf as compile string");
+		this.assertEquals(-inf.asCompileString.compile.(), -inf, "inf as compile string");
```
dyfer, line 57: This is cosmetic, but I think the string should be `"-inf as compile string"` (i.e. "-inf" and not "inf")?

dyfer:
@JordanHendersonMusic I think this is good! Tests pass (including Windows x64). 
Could you change the test's string to "-inf(...)"?

testsuite/classlibrary/TestFloat.sc
JordanHendersonMusic: Alright, should be fixed.

PR #6871, opened 2025-05-19
Prevent {}.scope from Posting Width Change Messages

prko:
## Purpose and Motivation
In a recent developer version of SC, I noticed that `{}.scope` posts a message concerning width. This message was not present (or I did not observe?), at least until the merging of #6282. With this commit, `{}.scope` will no longer post a message regarding the width change.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation


dyfer (approved):
Thanks!

Issue #6862, opened 2025-05-18
Scide sometimes fails to start and should try again


prko:
## Environment

* SuperCollider version: 3.14.0-dev [563fc04]
* Operating system: Windows
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

1. Launch SC-IDE 
 
## Expected
SC-IDE should launch whenever scide.exe is double-clicked in File Explorer
## Actual
1. Double-click sc-ide
2. scide.exe is occasionally not launched (in task manager, scide.exe quits automatically)
3. Double-click sc-ide again until it launches
4. Quit SC-IDE
6. Repeat the processes from steps 1 to 4.

Watch [the demo video](https://www.dropbox.com/scl/fi/ilaeh7hipes7rz57al7h3/Scide-sometimes-fails-to-start-and-should-try-again.mov?rlkey=qf3vxgbtviiobbkzvzxuvt83v&st=q23c0ulh&dl=0)

dyfer:
I just saw this too! I thought it was a quirk with a VM or something... @prko is this something you haven't seen before https://github.com/supercollider/supercollider/commit/563fc04fb693ef369811b2b79842d44ad5bd0782 ?

prko:
>@prko is this something you haven't seen before https://github.com/supercollider/supercollider/commit/563fc04fb693ef369811b2b79842d44ad5bd0782 ?

I believe so... At least [17abf35] does not show this problem:
I tested it 20 times successively, and it launches 20 times sc-ide.exe without any problems.

capital-G:
Thanks for the report @prko, this is really valuable!

Can someone attach a debugger to this and figure out which line is creating the problem here?

Maybe on windows the passed settings could be a nullpointer which could lead to a crash. I don't see any other reason why #6838 could create problems otherwise.

Although the question is why this can sometimes be a nullpointer on windows and even more so, why this is non-deterministic behavior - somehow worrying... :/ Maybe the settings loading is deferred b/c it is a file-bound task?
If we don't know what the problem is we could simply wrap a nullpointer check around it, but I think this would be more of a hack and I'd like to understand and fix this properly?

prko:
Um, I have not tested every commit.

I think the problem may have existed before https://github.com/supercollider/supercollider/pull/6838.

I briefly thought that https://github.com/supercollider/supercollider/pull/6812 and https://github.com/supercollider/supercollider/pull/6627 might be related to this problem. So I tried the following:
[...]

PR #6873, opened 2025-05-19
Disable redrawing lang status box on theme change

capital-G:
## Purpose and Motivation

#6838 has introduced a serious bug on linux and windows, making the IDE crash during startup.
For more information see https://github.com/supercollider/supercollider/issues/6862#issuecomment-2889937573

This introduces a quick fix which simply comments out the line which creates a segfault (at least on Linux).
W/o this line the style of the language status box will not be updated upon a theme change - not nice but better than a crash of the IDE ;)

I tested this fix on linux, but @jamshark70 can you give this a try if this fixes your issue?

@prko @dyfer can you verify that this also fixes the issue on windows?

Once it is verified that it fixes the issue on windows I think this could be merged?

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/lang_status_box.cpp
+++ b/editors/sc-ide/widgets/lang_status_box.cpp
@@ -60,7 +60,9 @@ void LangStatusBox::applySettings(Settings::Manager* settings) {
 
     // re-render the box, otherwise we will not update after e.g. a theme switch
     if (mLang != nullptr) {
-        onInterpreterStateChanged(mLang->state());
+        // this crashes on linux and windows, so we will skip that currently,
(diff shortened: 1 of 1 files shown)


prko:
I have tested the scide.exe in the artifact of this PR 20 times by double-clicking it.
scide.exe is launched every time.
So I would say this PR fixes my issues illustrated in https://github.com/supercollider/supercollider/issues/6862.

dyfer:
Thanks @capital-G for a quick fix!

dyfer:
@capital-G I closed the issues about the IDE crashing, but I filed a new one to keep track of the status box updating: #6876

PR #6877, opened 2025-05-19
Float.schelp: add missing `code` tag.

dyfer:
## Purpose and Motivation

The helpfile for `Float` would not render correctly after #6787 was merged:

```supercollider
ERROR: In /(...)/src/supercollider/HelpSource/Classes/Float.schelp:
  At line 148: syntax error, unexpected ::, expecting end of file
```

This PR adds the missing tag.

Note to self: it would be great to have a CI job that would catch schelp errors...

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Float.schelp
+++ b/HelpSource/Classes/Float.schelp
@@ -145,7 +145,7 @@ subsection::Using Floats as replacement for Integers
 
 In SuperCollider, Floats are 64 bits wide. Because an link::Classes/Integer:: is 32-bit, it can only capture integers in the range -2147483648 (code::-2^31::) to 2147483647 (code::2^31 - 1::).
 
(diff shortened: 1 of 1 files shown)


capital-G (approved):

capital-G:
Interesting that this creates a syntax error, I would have expected that this is simply not regarded as any kind of token.

Well, thanks for spotting and fixing!

Having a CMake target wouldn't be too hard w/ the new language config CLI, but it would be probably great to have a proper reporting mechanism for this in order to get aware of errors, something like JUnit and https://github.com/marketplace/actions/publish-test-results

dyfer:
We also have [SC_DOC_RENDER](https://github.com/supercollider/supercollider/blob/develop/CMakeLists.txt#L623C4-L623C17) flag in CMake, but I never used it. And yes, we'd need to somehow parse the rendering result either way.

PR #6832, opened 2025-05-09
plugins: TDelay: fix initialization sample

elgiano:
## Purpose and Motivation

Minimal fix for TDelay initialization, just providing the necessary internal state reset.
Due to missing internal state reset after init, TDelay used to incorrectly advance its internal counter. This caused an incorrect delay time of a trigger occuring on its first input sample, which would be delayed one sample too early.

This PR is part of #6813, and introduces small bug-fixing breaking changes in TDelay output.

## Types of changes

- Bug fix
- Breaking change

Output changes only if TDelay has a trigger on its first sample of input. If delayTime > dt (where dt is the time between first and second trigger), the output re-aligns with the previous version at the second trigger.
```supercollider
(
~printOutput = { |fn, nSamples(s.options.blockSize), action(_.postln)|
	fn.loadToFloatArray(nSamples/s.sampleRate, s, action)
};
~printOutputTrig = { |fn, nSamples=32|
	~printOutput.(fn, nSamples, {|v| v.round.asInteger.join.postln })
};
~printInitSample = { |ugen|
	~printOutput.({IEnvGen.ar(Env([ugen.value]),0)}, 1)
};
~printInitAndY0 = { |ugen, nSamples(s.options.blockSize)|
	~printOutput.({
		var out = ugen.value;
		var initSample = if (out.rate == \audio) {
			IEnvGen.ar(Env([out]),0);
		} {
			IEnvGen.kr(Env([out]),0);
		};
		[initSample, out]
	}, nSamples, {|v|
		"init = %\ty(0) = %".format(v[0], v[1]).postln
	})
}
)

// init sample is unchanged, output changes if triggered on x[0]
~printInitAndY0.({TDelay.ar(Impulse.ar(s.sampleRate/2), 1/SampleRate.ir)})
// before: init = 0.0	y(0) = 1.0
// after:  init = 0.0	y(0) = 0.0

// examples of outputs re-aligning after second trigger
// note that before PR the initial trig wasn't delayed correctly
~printOutput.({ var sr = SampleRate.ir; TDelay.ar(Impulse.ar(sr/2), 1/sr)})
// before: 10010101....
// after:  01010101....
~printOutputTrig.({ var sr = SampleRate.ir; TDelay.ar(Impulse.ar(sr/3), 5/sr)})
// before: 00001000000010000010000010000010
// after:  00000100000010000010000010000010

// example of outputs non re-aligning
[...]

PR #6879, opened 2025-05-20
plugins: add InitializationSample to test init samples

elgiano:
## Purpose and Motivation

Part of #6813, adds tests for all related PRs (#6810, #6816, #6832, #6822, #6805).
In order to test init samples, I wrote a new UGen `InitializationSample` that just output its input init sample forever (caching it at Ctor time).

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [ ] All tests are passing (see merge notes)
- [ ] Updated documentation

## Merge notes
- Dependencies: tests for Phasor and Sweep are failing before #6805 and #6822 are merged
- Missing documentation for InitializationSample


testsuite/classlibrary/TestUGen_initSample.sc
```diff
@@ -0,0 +1,253 @@
...
+		this.assertInitEqualsFirstSample { InitializationSample.ar(0) }
+	}
+
+	test_TriggerUGens {
+		var basicTest = { |ugenClass|
+			var tests = [];
+			if(ugenClass.respondsTo(\ar)) {
+				tests = tests ++ [
+					"{%.ar(DC.ar(0))}",
+					"{%.ar(DC.ar(1))}",
```
mtmccrea: It would be good to use less common "default" values for the input signal. E.g. a problematic UGen might have it's init sample hard-coded to `0.0`, so the test signal `DC.ar(0)` wouldn't catch the bug.

testsuite/classlibrary/TestUGen_initSample.sc
```diff
@@ -0,0 +1,253 @@
...
+				]
+			};
+			if(ugenClass.respondsTo(\kr)) {
+				tests = tests ++ [
+					"{%.kr(DC.kr(0))}",
+					"{%.kr(DC.kr(1))}",
```
mtmccrea: same as above wrt default input values

testsuite/classlibrary/TestUGen_initSample.sc
```diff
@@ -0,0 +1,253 @@
...
+		this.assertSynthOutput(
+			{ InitializationSample.ar(0) }, [0.0, 0.0],
+			"InitializationSample"
+		);
+		this.assertInitEqualsFirstSample { InitializationSample.ar(1) };
+		this.assertInitEqualsFirstSample { InitializationSample.ar(0) }
+	}
+
+	test_TriggerUGens {
+		var basicTest = { |ugenClass|
```
mtmccrea: are these tests "basic" in the sense that they're only testing the init sample and with matched input/output rates?

maybe a more descriptive name to that effect, like `initSampleTest`

testsuite/classlibrary/TestUGen_initSample.sc
```diff
@@ -0,0 +1,253 @@
...
+			[1,0,1,0,1,0,1,0],
+			"PulseDivider.kr(Impulse.kr(ControlRate.ir/2), 1)"
+		);
+
+		basicTest.(PulseCount);
+		basicTest.(Stepper);
+		this.assertInitEqualsFirstSample({Stepper.ar(DC.ar(1), DC.ar(1))});
+		this.assertInitEqualsFirstSample({Stepper.ar(DC.ar(1), DC.ar(0))});
+		this.assertInitEqualsFirstSample({Stepper.ar(DC.ar(0), DC.ar(1))});
+		this.assertInitEqualsFirstSample({Stepper.ar(DC.ar(0), DC.ar(0))});
```
mtmccrea: minor formatting: spaces within { }

There are more occurrences below, maybe just do a pass over the full file for formatting

testsuite/classlibrary/TestUGen_initSample.sc
```diff
@@ -0,0 +1,253 @@
...
+		);
+		this.assertSynthOutput(
+			{ InitializationSample.ar(0) }, [0.0, 0.0],
+			"InitializationSample"
+		);
+		this.assertInitEqualsFirstSample { InitializationSample.ar(1) };
+		this.assertInitEqualsFirstSample { InitializationSample.ar(0) }
+	}
+
+	test_TriggerUGens {
```
mtmccrea: This is discretionary, but... this test method is feeling rather monolithic. 

I see why it can be useful to have tests grouped by UGen, but you group instead by the kind of test. One simple approach would be to have init==first sample tests in their own test method, and break out the `assertSynthOutput` tests to their own method.

There may be clues how to more clearly separate tests by [naming convention](https://github.com/supercollider/supercollider/wiki/Unit-Testing-Guide#naming).

testsuite/classlibrary/TestUGen_initSample.sc
```diff
@@ -0,0 +1,253 @@
...
+		server = Server(this.class.name);
+		server.bootSync;
+	}
+
+	tearDown {
+		server.quit;
+		server.remove;
+	}
+
+	// Async helpers
```
mtmccrea, line 19: Are these async helpers useful enough to promote to a superclass that all UGen unit tests might have access to?

If you like the idea but it requires more thought or a lot more work restructuring inheritance of other UnitTests, it could be just earmarked with an Issue for a future PR.

mtmccrea (commented):
Thanks! This is nice factoring of many tests!

testsuite/classlibrary/TestUGen_initSample.sc
elgiano: the intention was to test triggers, but I see your point and we could use -1234 and 1234 instead

testsuite/classlibrary/TestUGen_initSample.sc
elgiano: Yeah I always wonder about this, since splitting server tests involves extra overhead for server start and stop. But it seems reasonable to split initSample == y[0] from the others

testsuite/classlibrary/TestUGen_initSample.sc
elgiano: Actually it could be the case, and I think it should be another PR, where we evaluate other places it is or can be used and come up with a "final" version

testsuite/classlibrary/TestUGen_initSample.sc
elgiano: I've split tests that use `assertInitEqualsFirstSample` from the ones using `assertSynthOutput`, which went to a separate `test_TriggerUGens_initialOutput` method

elgiano:
@mtmccrea I've finally got around to addressing all your comments :) thank you!
- I've refactored the big monolithic test in two separate ones: I wouldn't split it more than this because of the overhead of starting/stopping a separate server for each test method.
- Replaced input values 0 and 1 with less common -1234 and 1234
- fixed formatting

All the tests are passing.

Now, for me there are only two things left:
- Reusing async helpers in other test methods, but I think this should be another PR.
- In this block:
```supercollider
		// test start on reset when triggered at init
		this.assertSynthOutput({ Phasor.ar(1, 1, 0, 10, 5) }, [5]);
		// AK and AA subsample interpolation causes unexpected out: starts at 6
		this.assertSynthOutput({ Phasor.kr(1, 1, 0, 10, 5) }, [6]);
		this.assertSynthOutput({ Phasor.ar(DC.ar(1), 1, 0, 10, 5) }, [6]);
		this.assertSynthOutput({ Phasor.ar(DC.ar(1), DC.ar(1), 0, 10, 5) }, [6]);
```
Phasor resets to `reset + 1`, i.e. 6 in this case, for all except the first rates combination. We know this is a controversial behavior... I'm not sure whether it's better to test for it as it is, or not to test it at all.

mtmccrea:
Thanks for your work on this!
>Reusing async helpers in other test methods, but I think this should be another PR.

Agreed.

>Phasor resets to reset + 1, i.e. 6 in this case, for all except the first rates combination. We know this is a controversial behavior... I'm not sure whether it's better to test for it as it is, or not to test it at all.

I think it's ok to leave the tests, because once the Phasor behavior is fixed or otherwise resolved, these tests should break and it never hurts to consider why they break and update them :)

That said, could the explanatory comment be more specific, something as simple as:
>AK and AA subsample interpolation of the trigger causes unexpected out:

and also reference the appropriate Issue/PR?

elgiano:
Thanks! So I've left those tests and added a link to #6883 
Since I was there I noticed I had left a note about missing RunningMin/RunningMax tests, and I added them!

I think this is good to go!

PR #6882, opened 2025-05-21
sclang: improve performArg kwargs forwarding syntax

JordanHendersonMusic:
## Purpose and Motivation

Writing `a.performArgs(\foo, args, kwargs)` is awkward. This new syntax (which was added in the main keyword args PR) is now extended to make calling this easier.

Here are some examples.

```supercollider

a.*foo(args, kwargs);  // a.performArgs(\foo, args, kwargs)

a.*(args, kwargs); // a.performArgs(\value, args, kwargs)

```

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


JordanHendersonMusic:
@scztt this was the syntax change I mentioned today

capital-G:
IIRC this is the equivalent to pythons `**` on dicts, right?

I don't know if I am completely sold on the `*` in front of the method name b/c it would interfere w/ auto-completion and it looks more like a class method than something which would expand the last argument?

As we can't deprecate syntax I think we should always thrive for the best possible solution in these cases, but I can't really tell how much work it is to make this "more consistent" with the already existing syntax of expanding positional arguments. Would it make it easier to use a different symbol than `**`, e.g. `##`?

I think I'd prefer a different solution, even if this means it would only make 3.15 - and therefore indeed would make 3.14 a bit of a strange in-between release regarding syntax?

What are others thinking about this? @scztt @telephon ?

ps: maybe the `**` syntax could be useful outside of just function calls? In python it is also used to unpack the content of a dictionary within a dictionary, i.e.

```python
In [1]: a = {"a": 42}
   ...: b = {"b": 44}

In [2]: c = {**a, **b}

In [3]: c
Out[3]: {'a': 42, 'b': 44}
```

but in sclang this would plainly be `c = (\a: 42) ++ (\b: 44)` - don't know if this unpacking syntax could be really useful elsewhere in scalng?

JordanHendersonMusic:
If this does go ahead, the versions for class names calls and functional messages should be added to.

PR #6884, opened 2025-05-23
Improve the readability of SCDoc TOC

prko:
## Purpose and Motivation
In some documentation, the current table of contents (TOC) may be confusing for users due to its layout. At times, this could even be perceived as a bug.
As demonstrated in the two comparison examples below, this pull request (PR) enhances the readability of the SCDoc TOC:
Screenshot 2025-05-23 at 09 44 41Screenshot 2025-05-23 at 09 44 06
Screenshot 2025-05-23 at 09 47 23Screenshot 2025-05-23 at 09 47 03

## Types of changes

- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/SCDoc/SCDocRenderer.sc
+++ b/SCClassLibrary/SCDoc/SCDocRenderer.sc
@@ -886,59 +886,59 @@ SCDocHTMLRenderer {
 			node.children.do {|n|
 				switch(n.id,
 					\DESCRIPTION, {
-						stream << "<li class='toc1'><a href='#description'>Description</a></li>\n";
+						stream << "<li class='toc1'><a href='#description'>- Description</a></li>\n";
 						this.renderTOC(stream, n);
 					},
 					\EXAMPLES, {
-						stream << "<li class='toc1'><a href='#examples'>Examples</a></li>\n";
+						stream << "<li class='toc1'><a href='#examples'>- Examples</a></li>\n";
 						this.renderTOC(stream, n);
 					},
 					\CLASSMETHODS, {
 						if(n.notPrivOnly) {
-							stream << "<li class='toc1'><a href='#classmethods'>Class methods</a></li>\n";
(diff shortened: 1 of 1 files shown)


capital-G:
IMO it looks clearer w/o the ticks? I think this is a rather subjective change, I don't know how others feel about it?

The technical implementation should be done differently via https://developer.mozilla.org/en-US/docs/Web/CSS/list-style though - don't inject bullets as text.
I had a look at this and it seems the styling is not as easy as it should be because the SCDoc HTML has invalid syntax regarding the TOC (e.g. we open a `ul` an place another `ul` into it, although this nested `ul` should be wrapped in a `li` - but there are many other places as well, e.g. `<p/>`) - the browser recovers from this syntax but we can't style this properly anymore.
I'd personally advise to abandon SCDoc in its current state and focus our efforts more on a complete rewrite of SCDoc for 3.15.

prko:
>I think this is a rather subjective change, 

It could be because this is my latest feeling.
It did not bother me before... Strange...

Let's see what others think!

prko:
@capital-G @tedmoore 
Thanks for your valuable opinions!

- I removed adding `- ` before each title. The reason is as follows:
I tested it with my tutorials for my students. In my tutorials, I use a numbered title as follows:
   ```
   1. An introductory overview
   2. Getting started
   2.1. Getting Sound
   2.1.1. Generating Sound from Scratch
   2.1.2. Defining and Instantiating Synths
   2.1.3. Live Coding Essentials
   2.1.4. Building Musical Patterns
   ```
   With this style, `- ` decreases the readability.

- Instead, I adjusted the line padding between toc items.

- I think adding `.` before method names looks good.

The following is an example of the rendering using the last commit:   

I would be happy if the development team considered this pull request when revising the approach to SCDoc. For now, I will leave this PR as open.
Screenshot 2025-06-11 at 11 20 39
Screenshot 2025-06-11 at 11 20 54

PR #6888, opened 2025-05-24
lang: systemInformation and automatic printing of bug report info

scztt:
## Purpose and Motivation

Provide an automated way of printing pertinent system information for use in bug reports and CI scenarios.

## Types of changes

- Documentation
- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/Classes/Main.schelp
+++ b/HelpSource/Classes/Main.schelp
@@ -200,3 +200,9 @@ Get the current link::Classes/Platform::
 
 method::argv
 Get the command-line arguments passed to sclang.
+
+method::systemInformation
+Return a dictionary with system/context information like SuperCollider, QT, and OS versions.
+
+method::postBugReportInfo
+Post system information that is useful to attach to bug reports.
--- a/QtCollider/primitives/prim_misc.cpp
+++ b/QtCollider/primitives/prim_misc.cpp
@@ -55,6 +55,12 @@ QC_LANG_PRIMITIVE(QtGUI_DebugLevel, 0, PyrSlot* r, PyrSlot* a, VMGlobals* g) {
     return errNone;
 }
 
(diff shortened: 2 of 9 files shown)


scztt:
I'll fix the whitespace issues in the .sc files.
Review appreciated on the implementation? I'm printing the info in a way thats human readable, but also yaml-compatible in case this ever needs to be used for something automated.

dyfer:
I've built this PR locally - I think the posting is good in general.

What was the rationale of including audio devices? If the bug pertains to the server, then a server boot output should be included anyway, and that one prints the devices anyway. IMO the devices could be omitted, or included optionally with a flag. I have about 20 devices on my system and it makes the `postBugReportInfo` quite long.

Also, what do you think about formatting the output of `.postBugReportInfo` in markdown, so that it could be pasted straight into our issue template? And make it use the same naming convention?
```
* SuperCollider version:
* Operating system:

```
----
It would be great have the PR at least build in CI when submitting for a review, thanks!

scztt:
Markdown seems reasonable, I'll switch it. I added audio devices because it's requested in our bug report template, and there's no straightforward way to print JUST the device thats current in-use by the server (as far as I could tell). What if we switched both the devices and Quarks list to be comma-separated instead of doing each on a separate line? This would make it a bit less dense.

dyfer:
> it's requested in our bug report template,

I don't think I see that being requested, where's that?

> What if we switched both the devices and Quarks list to be comma-separated

Yes, I think that would be better.

scztt:
> I don't think I see that being requested, where's that?

I was thinking of this:
https://github.com/supercollider/supercollider/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md?plain=1#L17

capital-G:
Superseded by https://github.com/supercollider/supercollider/pull/6928 - thanks @scztt :)

PR #6892, opened 2025-05-25
scide: audio_status_box: init colors before widgets

elgiano:
## Purpose and Motivation

audio_status_box (the server status gui in scide) needs to initialize colors before rendering widgets. Otherwise the numbers displayed don't respect the theme colors until the server is booted a first time. This is particularly annoying on dark themes, since they are rendered in black.

## Types of changes

- Bug fix

## To-do list


capital-G (approved):
Ah, well spotted - I thought this was due to some unclear update flow, but it was just wrong order of code in #6838

PR #6896, opened 2025-05-26
CI: run ctests with linux build

JordanHendersonMusic:
## Purpose and Motivation

We have shit C++ test coverage (with the exception of supernova), and the ones we have, we don't run. Going forward we should add more C++ tests,  which can easily be done using the already existing boost tests.

Enable ctests during build on CI.

You can't run these in the 'test' phase because you need the build directory and only the install directory is exported as an artefact.
This means the sclang unit tests won't run unless the ctests pass, but that seems fine.

This is only run on linux because that is the only one I know how to make a virtual display for. In future we could run this elsewhere.

## Types of changes

- new tests
- ci change

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -185,5 +185,6 @@ jobs:
           Xvfb $DISPLAY -screen 0 1024x768x16 &
 
       - name: Run CTests
+        continue-on-error: true
         if: matrix.ctest
         run: cmake --build $BUILD_PATH --config Release --target test
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -20,6 +20,7 @@ jobs:
             cxx-compiler: "g++-9"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
           - job-name: ""
             os-version: "22.04"
@@ -27,6 +28,7 @@ jobs:
             cxx-compiler: "g++-11"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
           - job-name: ""
             os-version: "22.04"
@@ -35,6 +37,7 @@ jobs:
             use-syslibs: false
             shared-libscsynth: false
             artifact-suffix: "ubuntu-22.04-gcc12" # will trigger artifact upload; after changing this update the test-linux artifact-file in actions.yml
+            ctest: true
 
           - job-name: "use system libraries"
             os-version: "24.04"
@@ -42,6 +45,7 @@ jobs:
             cxx-compiler: "g++-13"
             use-syslibs: true
             shared-libscsynth: false
+            ctest: true
 
           - job-name: "shared libscsynth"
             os-version: "24.04"
@@ -49,6 +53,7 @@ jobs:
             cxx-compiler: "g++-14"
             use-syslibs: false
             shared-libscsynth: true
+            ctest: false
             
           - job-name: ""
             os-version: "22.04"
@@ -56,6 +61,7 @@ jobs:
             cxx-compiler: "clang++-11"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
           - job-name: ""
             os-version: "24.04"
@@ -63,6 +69,7 @@ jobs:
             cxx-compiler: "clang++-15"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
           - job-name: ""
             os-version: "24.04"
@@ -70,6 +77,7 @@ jobs:
             cxx-compiler: "clang++-16"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
           - job-name: ""
             os-version: "24.04"
@@ -77,6 +85,7 @@ jobs:
             cxx-compiler: "clang++-17"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
           - job-name: ""
             os-version: "24.04"
@@ -84,6 +93,7 @@ jobs:
             cxx-compiler: "clang++-18"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
     name: Ubuntu ${{ matrix.os-version }} ${{ matrix.c-compiler }} ${{ matrix.job-name }}
     env:
@@ -149,6 +159,7 @@ jobs:
           cd $BUILD_PATH
           make install -j2
 
+
       - name: create archive
         if: matrix.artifact-suffix
         run: cd $INSTALL_PATH && zip --symlinks -r $ARTIFACT_FILE.zip .
@@ -159,3 +170,21 @@ jobs:
         with:
           name: ${{ env.ARTIFACT_FILE }}
           path: ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE }}.zip
+
+      - name: Configure CTests
+        if: matrix.ctest
+        run: |
+          cd $BUILD_PATH
+          sudo apt-get install --yes xvfb jackd1 socat
+          # start jack
+          jackd --no-realtime --silent -d dummy &
+
+          # start virtual display server
+          export DISPLAY=:99
+          echo "DISPLAY=$DISPLAY" >> $GITHUB_ENV
+          Xvfb $DISPLAY -screen 0 1024x768x16 &
+
+      - name: Run CTests
+        if: matrix.ctest
+        run: cmake --build $BUILD_PATH --config Release --target test
+
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -30,7 +30,7 @@ jobs:
             shared-libscsynth: false
             ctest: false
 
-          - job-name: ""
+          - job-name: "with ctest"
             os-version: "22.04"
             c-compiler: "gcc-12"
             cxx-compiler: "g++-12"
@@ -39,7 +39,7 @@ jobs:
             artifact-suffix: "ubuntu-22.04-gcc12" # will trigger artifact upload; after changing this update the test-linux artifact-file in actions.yml
             ctest: true
 
-          - job-name: "use system libraries"
+          - job-name: "use system libraries with ctest"
             os-version: "24.04"
             c-compiler: "gcc-13"
             cxx-compiler: "g++-13"
@@ -87,13 +87,13 @@ jobs:
             shared-libscsynth: false
             ctest: false
 
-          - job-name: ""
+          - job-name: "with ctest"
             os-version: "24.04"
             c-compiler: "clang-18"
             cxx-compiler: "clang++-18"
             use-syslibs: false
             shared-libscsynth: false
-            ctest: false
+            ctest: true
 
     name: Ubuntu ${{ matrix.os-version }} ${{ matrix.c-compiler }} ${{ matrix.job-name }}
     env:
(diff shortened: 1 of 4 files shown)


JordanHendersonMusic:
@capital-G what do you think of this?

In future, I will add an interpreter directory to the `testsuite` with a bunch of C++ tests for primitives and whatnot.

dyfer:
Hi @JordanHendersonMusic, yes, I can enable actions.

I think I'd rather run all tests in the test job. Can you change this so that we create an additional artefact with the whole build directory and use that in the test step? That way we won't be running C tests for every build configuration, which I think is unnecessary (?)

Re: virtual display - I believe display server is always active on macOS and Windows. At the very least, we didn't have to set it up for the testsuite. 

My questions - is running tests with this action the best way to do it? Is the action safe? I don't know anything about CTest.

capital-G:
IMO this should be run in its independent file b/c we probably want to make this a singleton? This should be run dependent on a successful linux build, e.g.

```yml
name: CTest
on:
  workflow_call:

jobs:
  test-linux:
    runs-on: ubuntu-24.04

    steps:
      - uses: actions/checkout@v4
        with:
          submodules: recursive

      - name: Setup environment
        run: |
          # @todo install dependencies ...
          
          # start jack
          jackd --no-realtime --silent -d dummy &

          # start virtual display server
          export DISPLAY=:99
          echo "DISPLAY=$DISPLAY" >> $GITHUB_ENV
          Xvfb $DISPLAY -screen 0 1024x768x16 &

      - name: Run ctests
        shell: bash
        continue-on-error: true
        run: |
          timeout 20m cmake --build build --target test

```

I'd also think using vanilla cmake is sufficient here over introducing an additional app?

At a later stage we could/should introduce junit reporting to this.

dyfer:
Oh yes, I think a separate build might be easier than storing a build directory artifact etc.
BTW it should also use ccache to speed up the process (i.e. it should be similar to our linux build process, but without the compiler configurability etc.).

capital-G:
> I don't really know what these words mean 😆

GHA offers a matrix configuration, where every possible configuration gets tested, see

https://github.com/supercollider/supercollider/blob/b06dc8b4f72b803eed792aeff91d75433d110024/.github/workflows/build_linux.yml#L9-L22

We could modify it to include it a ctest flag, which runs a specific command only if this flag is set

```patch
index 13b14ebd04..a62b7a039d 100644
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -20,6 +20,7 @@ jobs:
             cxx-compiler: "g++-9"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: true

           - job-name: ""
             os-version: "22.04"
@@ -159,3 +160,8 @@ jobs:
         with:
           name: ${{ env.ARTIFACT_FILE }}
           path: ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE }}.zip
+
+      - name: ctest
+        if: matrix.ctest
+        run: cmake --build build --target test
+
```

At the same time, keeping things atomic is always preferable, b/c certain actions are only triggered if actions before were successful (i.e. server-tests are only executed when the build was successful - nightlies are also only published when tests were successful(?)), but GHA does not allow to share the state across jobs.

Somehow difficult what is best in this case - duplicating all the dependencies seems also error-prone?

.github/workflows/build_linux.yml
```diff
@@ -20,13 +20,15 @@ jobs:
...
 
           - job-name: ""
             os-version: "22.04"
             c-compiler: "gcc-11"
             cxx-compiler: "g++-11"
             use-syslibs: false
             shared-libscsynth: false
+            ctest: false
 
           - job-name: ""
```
dyfer, line 33: I'd append a name to jobs with ctest, e.g. `"with ctest"`.

.github/workflows/build_windows.yml
```diff
@@ -197,3 +201,12 @@ jobs:
...
           path: ${{ env.INSTALL_PATH }}/*.exe
+
+      - name: Configure CTests
+        if: matrix.ctest
+        run: cd ${{ github.workspace }}
+
+      - name: Run CTests
+        if: matrix.ctest
+        shell: bash
+        run: cmake --build $BUILD_PATH --config Release --target test
```
dyfer, line 212: see comment on macOS regarding working directory

dyfer (commented):
Thanks @JordanHendersonMusic for working on this!

@capital-G I don't think there's really a better way to do this than to incorporate it in the build jobs; otherwise we'd need to copy the basic build setup for each platform, and then either build again, or copy the whole build directory between steps using a GH artifact, which would push our [cache usage](https://github.com/supercollider/supercollider/actions/caches) even more (we're constantly running over the limit anyway, mostly because of multiple Qt6 caches).

@JordanHendersonMusic re: MinGW: we don't officially support this platform, we build it as a courtesy (MinGW was _the_ way to build SC on Windows back in the day, but it was eventually dropped once the QtWebBrowser was added as that module is not available in Qt for MinGW). Unless the fix is straightforward, I'd suggest just disabling tests on MinGW. We could still file an issue just to note there's a problem with it...

EDIT: one more thought on ctest: do any of them depend on Qt? If we'd build without the Qt and the IDE, then the build would be a bit faster, and installing qt is one of the bigger dependency to install (esp on macOS and Windows). If we didn't need to build with Qt, then maybe duplicating code for the build would be okay?

dyfer:
All right, now that I understand this a little better, I'm leaning towards using separate workflow files for this, with setting up the minimal necessary dependencies.

Do we actually need to setup a virtual display to run these tests? At least the current tests don't seem to be using a display.

capital-G:
> I don't think there's really a better way to do this than to incorporate it in the build jobs; otherwise we'd need to copy the basic build setup for each platform, and then either build again

k, then lets continue w/ the current approach :)

Any feelings regarding if the exit code of ctest should affect the build step as a whole? We could use https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error to continue to the class library tests? But then we wouldn't get easily aware of failing ctest which is probably the purpose of this PR.

I'd propose to let the build fail if its ctest fail and at a later stage, with another PR, introduce junit reporting which would tell us about any failing tests by posting those in the PR.

capital-G:
> Do we actually need to setup a virtual display on Linux to run these tests? At least the current tests don't seem to be using a display.

IIRC when sclang gets compiled w/ QtCollider it needs access to the display server during startup or it will bail out.
But as these tests wouldn't require a sclang binary I don't think we would to emulate such a display server?

dyfer:
> IIRC when sclang gets compiled w/ QtCollider it needs access to the display server during startup or it will bail out.
> But as these tests wouldn't require a sclang binary I don't think we would to emulate such a display server?

On macOS I've run 
```sh
cmake -G Xcode -D SUPERNOVA=ON -D SC_QT=OFF ..
cmake --build . --config Release --target RUN_TESTS
```
so no Qt.
I guess if there were tests added that use Qt, then we'd need to build with Qt... but not for now?

Again, given that the test building is fast, and only dependency installation takes time, I'd now lean towards putting them in their own jobs.

JordanHendersonMusic:
I think we should compile with Qt because we should have tests for the IDE — right now there are none, sure, but part of the problem is that their is no good way to write C++ test, this is step one in that effort.

While there are some sclang tests here, in the future most of these tests will be checking C++ functionality, so every dependency, particularly something big, needs to be included, and we need to check this on every platform.

> Any feelings regarding if the exit code of ctest should affect the build step as a whole?

If the C++ code is failing, something is very wrong; I don't think it makes sense to do the sclang unit tests in this case.


> EDIT: yes, at least on Xcode I confirmed that RUN_TESTS works: cmake --build . --config Release --target RUN_TESTS (in the build directory)

Amazing thank you!

dyfer:
> I think we should compile with Qt because we should have tests for the IDE

Got it.

I still lean towards having these in a separate file, despite the need to set up the dependencies, since they don't need an extensive compile time (and don't need the result of the main compilation).

Do you need help setting up the GHA scaffolding for this?

dyfer:
> if you think you can do it quicker than reviewing my attempt

I think I can, I'll report back in a moment.

JordanHendersonMusic:
@dyfer thanks for taking a look at this, I've basically copied your approach and everything seems to be working correctly now...


Unfortunately, this has revealed more bugs in windows, which I know you spent a lot of effort fixing to enable the windows CI.

Here is a summary:

### windows 32
* 1 - scsynth_test_run (Exit code 0xc0000135
* 6 - server_dsp_thread_test_run (Exit code 0xc0000374

### windows 64 msvc 2022
* 1 - scsynth_test_run (Exit code 0xc0000135
* 6 - server_dsp_thread_test_run (Exit code 0xc0000374
* 15 - sclang_crash_1_run (SEGFAULT)

### windows 64 MinGW
* 1 - scsynth_test_run (Exit code 0xc0000135
* 2 - buffer_manager_test_run (Exit code 0xc0000135
* 12 - sndfile_backend_test_run (Exit code 0xc0000135
* 15 - sclang_crash_1_run (Exit code 0xc0000139
* 16 - sclang_crash_2_run (Exit code 0xc0000139
* 17 - sclang_crash_3_run (Exit code 0xc0000139
* 18 - sclang_crash_deprecated_run (Exit code 0xc0000139

*Just a note, the tests labelled 'server' or 'buffer_ma..' come from supernova — these should probably be renamed to make it clearer where they come from.*

I think tests 1 and 6 need fixing and should be turned into issues, these are problems with scsynth and supernova so I don't know if I will be much help fixing these.

Test 15 also fails for some reason, I'll have a look at this soon, I imagine it is uninitialised memory. @jamshark70 git says you wrote this test (immediately below), I know it was 15 years ago, but I don't suppose you have any recollection of it?
```
x = { 1+1 }.def;
x.instVarSize.do { |i|
	i.debug("index");
	x.instVarAt(i).postln
};
0.exit;
```

It would be good to fix all these issues (1, 6 & 15) before the next release as these are serious bugs — I don't know if that will be possible given the time constraint.

Regarding mingw, since we don't officially support this, I'm inclined to ignore this. Perhaps this build should just be removed because it is clearly not stable enough for use — all the sclang tests crash.

What do we think about merging this — do the tests need fixing first?

JordanHendersonMusic:
Well since these haven't been ran as a part of the CI it is possible they have never worked on those platforms.


### 1

https://github.com/supercollider/supercollider/blob/develop/testsuite/server/scsynth/scsynth_test.cpp

```cpp
BOOST_AUTO_TEST_CASE(create_and_destroy_world) {
    WorldOptions options;
    options.mRealTime = false;
    World* world = World_New(&options);
    World_Cleanup(world, true);
}
```

### 6

https://github.com/supercollider/supercollider/blob/bd34e16719ab2748a956a4cd3adb490de966f13e/testsuite/server/supernova/server_dsp_thread_test.cpp

Something to do with supernova. It might be possible to add a verbose flag to ctest so it prints exactly which test failed.

### 15

https://github.com/supercollider/supercollider/blob/bd34e16719ab2748a956a4cd3adb490de966f13e/testsuite/sclang/sclang_crash_1.scd
```supercollider
x = { 1+1 }.def;
x.instVarSize.do { |i|
	i.debug("index");
	x.instVarAt(i).postln
};
0.exit;
```

jamshark70:
> @jamshark70 git says you wrote this test (immediately below), I know it was 15 years ago, but I don't suppose you have any recollection of it?

Seems that it was originally written by Tim Blechmann; I only extended it.

I don't remember the scenario.

PR #6880, opened 2025-05-20
Correct openOS behaviour on Windows

prko:
Fixes #6312. Code designed by @Spacechild1. Thank you!

## Purpose and Motivation
Fixes https://github.com/supercollider/supercollider/issues/6312. Code designed by @Spacechild1. Thank you!

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] This PR is ready for review

PR #6898, opened 2025-05-26
classlib: update String .openOS on Windows

dyfer:
## Purpose and Motivation

`"start"` is a `cmd.exe` command so we can use String .unixCmd instead

Background: https://github.com/supercollider/supercollider/pull/6837 removed `cmd /c` from `SequenceableCollection .unixCmd`, which inadvertently broke `String .openOS`. The latter was then fixed in #6880, but probably a cleaner solution is to use `String .unixCmd` for `String .openOS`, as suggested by @Spacechild1 in https://github.com/supercollider/supercollider/pull/6858#issuecomment-2907722768.

## Types of changes

- Cleanup?

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Collections/windows/extString_windows.sc
+++ b/SCClassLibrary/Common/Collections/windows/extString_windows.sc
@@ -4,14 +4,14 @@
 	}
 
 	openOS {
- 		// start "title" "command"
-     	if (PathName(this).isFolder) {
- 			// workaround for folders, see https://github.com/supercollider/supercollider/issues/6312
- 			// NB: Explorer.exe does not accept path names with forward slashes!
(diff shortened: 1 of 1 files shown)


prko:
I tested it with the artefact [CI / build-windows / Windows 64-bit MSVC 2022 (pull_request)](https://github.com/supercollider/supercollider/actions/runs/15262333608/job/42922313699?pr=6898).

It works, thanks!

dyfer:
@Spacechild1 as this is about Windows, do you think this is okay?

Spacechild1:
Ok with me!

PR #6899, opened 2025-05-27
portaudio/WASAPI: enable automatic sample rate conversion

Spacechild1:
## Purpose and Motivation

This PR enables automatic sample rate conversion for the WASAPI backend.

Addresses https://github.com/supercollider/supercollider/issues/6506.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #6905, opened 2025-05-28
servers: fix PortAudio hostAPI stream info (Win32)

dyfer:
## Purpose and Motivation

#6899 broke booting the server on Windows with non-WASAPI devices (i.e. the default device...)

This PR fixes that, allowing the servers to boot again.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested - tested booting with the default device (MME) and with a WASAPI device
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/common/SC_PaUtils.cpp
+++ b/common/SC_PaUtils.cpp
@@ -174,20 +174,22 @@ PaStreamParameters MakePaStreamParameters(int device, int channelCount, double s
     streamParams.hostApiSpecificStreamInfo = &macInfo;
 #elif defined(_WIN32)
     // WASAPI options
-    static PaWasapiStreamInfo wasapiInfo = []() {
-        PaWasapiStreamInfo info;
-        memset(&info, 0, sizeof(info));
-        info.size = sizeof(PaWasapiStreamInfo);
-        info.hostApiType = paWASAPI;
-        info.version = 1;
-        info.flags = paWinWasapiAutoConvert; // automatic sample rate conversion
-        return info;
-    }();
-
     const PaDeviceInfo* info = Pa_GetDeviceInfo(device);
     const PaHostApiInfo* api = Pa_GetHostApiInfo(info->hostApi);
-    if (api->type == paWASAPI)
+    if (api->type == paWASAPI) {
+        static PaWasapiStreamInfo wasapiInfo = []() {
+            PaWasapiStreamInfo info;
+            memset(&info, 0, sizeof(info));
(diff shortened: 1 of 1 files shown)


dyfer:
@Spacechild1 this seems to work, can you take a look?

common/SC_PaUtils.cpp
```diff
@@ -174,20 +174,23 @@ PaStreamParameters MakePaStreamParameters(int device, int channelCount, double s
...
-        info.flags = paWinWasapiAutoConvert; // automatic sample rate conversion
-        return info;
-    }();
-
     const PaDeviceInfo* info = Pa_GetDeviceInfo(device);
-    const PaHostApiInfo* api = Pa_GetHostApiInfo(info->hostApi);
-    if (api->type == paWASAPI)
+    const PaHostApiInfo* api = info ? Pa_GetHostApiInfo(info->hostApi) : nullptr;
+
+    if (api && api->type == paWASAPI) {
```
Spacechild1, line 180: Small nitpick, but this could really be simplified to:
```
const PaDeviceInfo* info = Pa_GetDeviceInfo(device);
const PaHostApiInfo* api = Pa_GetHostApiInfo(info->hostApi);
if (api->type == paWASAPI) {
```

We got the device index and API index from portaudio, so these are guaranteed to be valid. We never do null-checks on those. If you feel like it, you can change this. Otherwise I will just merge it as is.

Spacechild1 (approved):

common/SC_PaUtils.cpp
dyfer: Sounds good, changing

Spacechild1: Cool! Will merge now.

dyfer:
> Thanks! You would have only needed to add the missing `else` branch at the end, but this version is also fine.

FWIW this was my first attempt, but it was still crashing: https://github.com/dyfer/supercollider/commit/3a5f6c32a3

Spacechild1:
> FWIW this was my first attempt, but it was still crashing: https://github.com/dyfer/supercollider/commit/3a5f6c32a3

Sure, because this still left `hostApiSpecificStreamInfo` uninitialized for APIs other than WASAPI. You only added a null-check for `info`, which didn't really do anything :)

Anyway, shame on me for making this stupid mistake in the first place.

dyfer:
Thanks! I realized my mistake right after posting this comment, that's why I removed it... But not before you replied :)

PR #6907, opened 2025-05-29
UserView help: remove refreshInRect

dyfer:
## Purpose and Motivation

Remove example with refreshInRect, as it is not available in the Qt gui system.

To clarify: Even after changing class names to use the Qt GUI, the code still errors out.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation


dyfer:
@tedmoore @HotwheelsSisyphus I think it would be good to merge this before 3.14 to remove outdated information, what do you think?

tedmoore (approved):

PR #6909, opened 2025-05-30
Adjust the bottom position of the FreqScope and the bottom position of the ServerMeter for all platforms

prko:
…m position of the ServerMeter for all platforms.

## Purpose and Motivation
1. On Windows, the default position of `s.freqscope` and `FreqScope()` is covered by the taskbar.
  image
 
2. On all platforms, the position of `s.meter` and `ServerMeter(s, x, x)` and `s.freqscope` (also `FreqScope()`) collides as follows:
   - on Linux:
      Screenshot 2025-05-30 at 14 24 21
   - on Windows
      Screenshot 2025-05-30 at 14 25 39
   - on macOS
      Screenshot 2025-05-30 at 14 37 55

This PR resolves these collisions as follows:
- on Linux
Screenshot 2025-05-30 at 14 45 06
- on Windows
Screenshot 2025-05-30 at 14 22 37
- on macOS
Screenshot 2025-05-30 at 14 48 53

This is not a bug, but it is very inconvenient, and to me, at least, it seems to be buggy.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -172,9 +172,16 @@ Window {
 	onClose { ^view.onClose }
 	onClose_ { arg func; view.onClose_(func) }
 
-	*availableBottom {
-		^this.screenBounds.height - this.availableBounds.top - this.availableBounds.height
-	}
+	*lowestPosition {
+ 		^this.screenBounds.bottom - this.availableBounds.bottom
+ 	}
+
+	moveToBottom { |margin = 0|
+		this.bounds = this.bounds.moveTo(
+			this.bounds.left,
+			Window.lowestPosition + margin
+		)
+    	}
 
 	// TODO
 	addToOnClose{ arg function; }
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -173,15 +173,15 @@ Window {
 	onClose_ { arg func; view.onClose_(func) }
 
 	*lowestPosition {
- 		^this.screenBounds.bottom - this.availableBounds.bottom
- 	}
-
+		^this.screenBounds.bottom - this.availableBounds.bottom
+	}
+	
 	moveToBottom { |margin = 0|
 		this.bounds = this.bounds.moveTo(
 			this.bounds.left,
 			Window.lowestPosition + margin
 		)
-    	}
+	}
 
 	// TODO
 	addToOnClose{ arg function; }
--- a/SCClassLibrary/Common/GUI/Base/QWindow.sc
+++ b/SCClassLibrary/Common/GUI/Base/QWindow.sc
@@ -172,6 +172,10 @@ Window {
 	onClose { ^view.onClose }
 	onClose_ { arg func; view.onClose_(func) }
 
+	*availableBottom {
+		^this.screenBounds.height - this.availableBounds.top - this.availableBounds.height
+	}
+
 	// TODO
 	addToOnClose{ arg function; }
 	removeFromOnClose{ arg function; }
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -373,12 +373,9 @@ FreqScope {
 				});
 			};
 
-			window = Window(
-				"Freq Analyzer",
-				rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
-				.moveTo(5, Window.availableBottom + 20)
-				, false
-			);
+			window = Window("Freq Analyzer", resizable: false).bounds_(
+ 				rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
+			).moveToBottom;
 
 			freqLabel.size.do({ arg i;
 				freqLabel[i] = StaticText(window, Rect(pad[0] - (freqLabelDist*0.5) + (i*freqLabelDist), pad[2] - 10, freqLabelDist, 10))
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -376,7 +376,7 @@ FreqScope {
 			window = Window(
 				"Freq Analyzer",
 				rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
-				.moveTo(5, Window.availableBottom)
+				.moveTo(5, Window.availableBottom + 20)
 				, false
 			);
 
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -373,7 +373,13 @@ FreqScope {
 				});
 			};
 
-			window = Window("Freq Analyzer", rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4), false);
+			window = Window(
+				"Freq Analyzer",
+				rect.resizeBy(
+					pad[0] + pad[1] + 4,
+					pad[2] + pad[3] + 4
+				).moveTo(0, Platform.case(\windows, { 78 }, { 0 })),
+				false);
 
 			freqLabel.size.do({ arg i;
 				freqLabel[i] = StaticText(window, Rect(pad[0] - (freqLabelDist*0.5) + (i*freqLabelDist), pad[2] - 10, freqLabelDist, 10))
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -374,7 +374,7 @@ FreqScope {
 			};
 
 			window = Window("Freq Analyzer", rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
-				.moveTo(5, Window.screenBounds.height - Window.availableBounds.height * 1.6),
+				.moveTo(5, Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height),
 				false
 			);
 
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -374,12 +374,17 @@ FreqScope {
 			};
 
 			window = Window(
-				"Freq Analyzer",
-				rect.resizeBy(
-					pad[0] + pad[1] + 4,
-					pad[2] + pad[3] + 4
-				).moveTo(0, Platform.case(\windows, { 78 }, { 0 })),
-				false);
+ 				"Freq Analyzer",
+ 				rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4
+ 				).moveTo(
+					5,
+					Platform.case(
+						\windows, { Window.screenBounds.height - Window.availableBounds.height * 1.6 },
+						{ 0 }
+						)
+					),
+				false
+			);
 
 			freqLabel.size.do({ arg i;
 				freqLabel[i] = StaticText(window, Rect(pad[0] - (freqLabelDist*0.5) + (i*freqLabelDist), pad[2] - 10, freqLabelDist, 10))
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -373,16 +373,8 @@ FreqScope {
 				});
 			};
 
-			window = Window(
- 				"Freq Analyzer",
- 				rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4
- 				).moveTo(
-					5,
-					Platform.case(
-						\windows, { Window.screenBounds.height - Window.availableBounds.height * 1.6 },
-						{ 0 }
-						)
-					),
+			window = Window("Freq Analyzer", rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
+				.moveTo(5, Window.screenBounds.height - Window.availableBounds.height * 1.6),
 				false
 			);
 
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -373,9 +373,11 @@ FreqScope {
 				});
 			};
 
-			window = Window("Freq Analyzer", rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
-				.moveTo(5, Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height),
-				false
+			window = Window(
+				"Freq Analyzer",
+				rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
+				.moveTo(5, Window.availableBottom)
+				, false
 			);
 
 			freqLabel.size.do({ arg i;
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -237,15 +237,9 @@ ServerMeter {
 
 		numIns = numIns ?? { server.options.numInputBusChannels };
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
-
-		window = Window.new(
-			server.name ++ " levels (dBFS)",
-			Rect(
-				5,
-				328 + 60 + Window.availableBottom,
-				ServerMeterView.getWidth(numIns, numOuts),
-				ServerMeterView.height
-			),
+		
+ 		window = Window.new(server.name ++ " levels (dBFS)",
+ 			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
 			false);
 
 		meterView = ServerMeterView(server, window, 0@0, numIns, numOuts);
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -238,9 +238,12 @@ ServerMeter {
 		numIns = numIns ?? { server.options.numInputBusChannels };
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
 
-		window = Window.new(server.name ++ " levels (dBFS)",
-			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
-			false);
+		window = Window.new(server.name ++ " levels (dBFS)", resizable: false).bounds_(Rect(
+			5,
+			0,
+			ServerMeterView.getWidth(numIns, numOuts),
+			ServerMeterView.height
+		)).moveToBottom(356); // 356 is the height of FreqScope
 
 		meterView = ServerMeterView(server, window, 0@0, numIns, numOuts);
 		meterView.view.keyDownAction_( { arg view, char, modifiers;
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -237,9 +237,9 @@ ServerMeter {
 
 		numIns = numIns ?? { server.options.numInputBusChannels };
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
-		
- 		window = Window.new(server.name ++ " levels (dBFS)",
- 			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
+
+		window = Window.new(server.name ++ " levels (dBFS)",
+			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
 			false);
 
 		meterView = ServerMeterView(server, window, 0@0, numIns, numOuts);
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -238,8 +238,13 @@ ServerMeter {
 		numIns = numIns ?? { server.options.numInputBusChannels };
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
 
-		window = Window.new(server.name ++ " levels (dBFS)",
-			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
+		window = Window.new(
+			server.name ++ " levels (dBFS)",
+			Rect(
+				5,
+				372 + Platform.case(\windows, { 62 }, \linux, { 20 }, \osx, { 0 }),
+				ServerMeterView.getWidth(numIns, numOuts),
+				ServerMeterView.height),
 			false);
 
 		meterView = ServerMeterView(server, window, 0@0, numIns, numOuts);
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -238,15 +238,9 @@ ServerMeter {
 		numIns = numIns ?? { server.options.numInputBusChannels };
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
 
-		window = Window.new(
- 			server.name ++ " levels (dBFS)",
- 			Rect(
- 				5,
- 				376 + Platform.case(\windows, { 58 }, \linux, { 26 }, \osx, { 0 }),
- 				ServerMeterView.getWidth(numIns, numOuts),
- 				ServerMeterView.height
-				),
- 			false);
+		window = Window.new(server.name ++ " levels (dBFS)",
+			Rect(5, Window.screenBounds.height - Window.availableBounds.height * 1.6 + 360, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
+			false);
 
 		meterView = ServerMeterView(server, window, 0@0, numIns, numOuts);
 		meterView.view.keyDownAction_( { arg view, char, modifiers;
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -238,8 +238,14 @@ ServerMeter {
 		numIns = numIns ?? { server.options.numInputBusChannels };
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
 
-		window = Window.new(server.name ++ " levels (dBFS)",
-			Rect(5, Window.screenBounds.height - Window.availableBounds.height * 1.6 + 360, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
+		window = Window.new(
+			server.name ++ " levels (dBFS)",
+			Rect(
+				5, 
+				Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height + 328 + 70, 
+				ServerMeterView.getWidth(numIns, numOuts), 
+				ServerMeterView.height
+			),
 			false);
 
 		meterView = ServerMeterView(server, window, 0@0, numIns, numOuts);
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -239,13 +239,14 @@ ServerMeter {
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
 
 		window = Window.new(
-			server.name ++ " levels (dBFS)",
-			Rect(
-				5,
-				376 + Platform.case(\windows, { 58 }, \linux, { 20 }, \osx, { 0 }),
-				ServerMeterView.getWidth(numIns, numOuts),
-				ServerMeterView.height),
-			false);
+ 			server.name ++ " levels (dBFS)",
+ 			Rect(
+ 				5,
+ 				376 + Platform.case(\windows, { 58 }, \linux, { 26 }, \osx, { 0 }),
+ 				ServerMeterView.getWidth(numIns, numOuts),
+ 				ServerMeterView.height
+				),
+ 			false);
 
 		meterView = ServerMeterView(server, window, 0@0, numIns, numOuts);
 		meterView.view.keyDownAction_( { arg view, char, modifiers;
--- a/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
+++ b/SCClassLibrary/Common/GUI/tools/ServerMeter.sc
@@ -241,9 +241,9 @@ ServerMeter {
 		window = Window.new(
 			server.name ++ " levels (dBFS)",
 			Rect(
-				5, 
-				Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height + 328 + 70, 
-				ServerMeterView.getWidth(numIns, numOuts), 
+				5,
+				328 + 60 + Window.availableBottom,
+				ServerMeterView.getWidth(numIns, numOuts),
 				ServerMeterView.height
 			),
 			false);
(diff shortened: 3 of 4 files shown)


dyfer:
Thanks for the PR.

With the amount of work before the release we may need to focus on functional (as opposed to "cosmetic") PRs.

Just on the quick look - used magic (hardcoded) numbers for each platform is not the best way to address this.

`Window.availableBounds` gives available display area and this could be used to accurately place the window on screen for all systems consistently.

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
```diff
@@ -373,7 +373,18 @@ FreqScope {
 				});
 			};
 
-			window = Window("Freq Analyzer", rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4), false);
```
dyfer, line 376: I thinking about something like (code NOT tested):
```supercollider
rect = rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4);
rect.moveTo(5, rect.height);
// or
// rect = rect.moveTo(5, Window.screenBounds.height - Window.availableBounds.height + rect.height);
Window("Freq Analyzer", rect, false)
```

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
```diff
@@ -238,9 +238,15 @@ ServerMeter {
 		numIns = numIns ?? { server.options.numInputBusChannels };
 		numOuts = numOuts ?? { server.options.numOutputBusChannels };
 
-		window = Window.new(server.name ++ " levels (dBFS)",
-			Rect(5, 305, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
```
dyfer, line 242:
```supercollider
Rect(5, ServerMeterView.height, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
```

dyfer (commented):
Again, I don't believe we should have any Platform switches here (but I have not tested all the possibilities).
Also, meter's height is available as a classvar, you don't need to create the meterview first.
See my (untested) suggestions below (sorry in advance for any typos, but this is to demonstrate the logic I'm suggesting).

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
dyfer: Of course you can offset it further to not collide with the freqscope... but I'm not sure if that's the best default? I tend to use servermeter only and then I'd prefer it closer to the bottom.
I wander what others think about the default here.

dyfer: Okay, I see now that you _are_ trying to offset the servermeter by the size of the freqscope. 
Maybe we can just change the hardcoded number for now:
```supercollider
Rect(5, 400, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
```
or something like this.

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
prko: Thank you for your suggestion.

---

With the first idea:
```
rect = rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4);
rect.moveTo(5, rect.height);
Window("Freq Analyzer", rect, false)
```
No window is created.

---

With the following code:
```
rect = rect.moveTo(5, Window.screenBounds.height - Window.availableBounds.height + rect.height);
Window("Freq Analyzer", rect, false)
```
The following error occurs:
```
ERROR: Message 'onClose_' not understood.
RECEIVER:
   nil
ARGS:
Instance of Function {    (0x15a35bfe8, gc=B4, fmt=00, flg=00, set=02)
  instance variables [2]
    def : instance of FunctionDef in Method Meta_FreqScope:new
    context : Frame (0x158c78918) of Meta_FreqScope:new
}
KEYWORD ARGUMENTS:
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg this = nil
		arg selector = 'onClose_'
		arg args = [*1]
		arg kwargs = [*0]
	Meta_FreqScope:new
		arg this = <instance of Meta_FreqScope>
		arg width = 522
		arg height = 300
		arg busNum = 0
		arg scopeColor = <instance of Color>
		arg bgColor = nil
		arg server = <instance of Server>
		var rect = <instance of Rect>
		var scope = <instance of FreqScopeView>
		var window = nil
		var pad = [*4]
		var font = <instance of Font>
		var freqLabel = [*12]
		var freqLabelDist = 47.454545454545
		var dbLabel = [*17]
		var dbLabelDist = 18.75
		var setFreqLabelVals = <instance of Function>
		var setDBLabelVals = <instance of Function>
		var nyquistKHz = 24.0
		var res = nil
		var code = "s.freqscope"
		var doc = nil
	Process:interpretPrintCmdLine
		arg this = <instance of Main>
^^ ERROR: Message 'onClose_' not understood.
RECEIVER: nil
```

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
prko: Thanks for the suggestion.

---

With the first code:
```
Rect(5, ServerMeterView.height, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
```
I get the following result:
스크린샷 2025-06-05 08 04 01

---

With the second code:
```
Rect(5, 400, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
```
I get the following result:
스크린샷 2025-06-05 08 06 24

스크린샷 2025-06-05 08 09 43

---

As you might see, 
- the first one makes the collision region bigger.
- the second one is ideal for macOS (and also Linux, not tested on Linux), but not for Windows.

dyfer: Okay. This is not a priority right now, but I still feel the implementation should not use a Platform switch.
We should be able to figure out the offset using some combination  of `Window.screenBounds` and `Window.availableBounds`. The first step would be to check how these bounds behave in relation to the Windows Taskbar.

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
dyfer: I said the code is untested. 

You likely have .onClose calling on the wrong object.

prko:
@dyfer 
>Again, I don't believe we should have any Platform switches here (but I have not tested all the possibilities).

This issue seems to arise because the taskbar in Windows is positioned at the bottom of the screen by default, whereas similar UI elements in other operating systems are typically located at the top. Isn't `Platform.case` designed precisely to address such OS-specific differences?  

I’m curious to understand the basis for the belief that an alternative solution exists. Have you tested other possibilities? If so, could you share the results?  

From my perspective, there are only two viable solutions:  
1. Use `Platform.case` to accommodate OS differences  
2. Add a note in `readme_Window.rm` or `FreqScope` advising Windows users to relocate the taskbar to the top of the screen  

Which option do you believe is more appropriate? Considering that Windows users generally do not reposition the taskbar unless absolutely necessary, I wonder if the second option is impractical. I experimented with making Windows resemble macOS in the past but found it inconvenient. Moreover, as far as I’m aware, Windows 11 does not allow users to change the taskbar’s position.  

I’d appreciate hearing your thoughts on this.

SCClassLibrary/Common/GUI/tools/ServerMeter.sc
prko: If you do not want to use Platform.case, the following is possible:
- `Window` for `FreqScope`
    from
    ```
    Platform.case(
    	\windows, { Window.screenBounds.height - Window.availableBounds.height * 1.6 },
    	{ 0 }
    	)
    )
    ```
    to
    ```
    Window.screenBounds.height - Window.availableBounds.height * 1.6
    ```
- `Window` for `ServerMeter`
    ```
    window = window = Window.new(
    	server.name ++ " levels (dBFS)",
     	Rect(
    		5,
    		376 + Platform.case(\windows, { 58 }, \linux, { 26 }, \osx, { 0 }),
    		ServerMeterView.getWidth(numIns, numOuts),
    		ServerMeterView.height
   		),
    	false);
    ```
    to 
    ```
    			Rect(5, Window.screenBounds.height - Window.availableBounds.height * 1.6 + 350, ServerMeterView.getWidth(numIns, numOuts), ServerMeterView.height),
			false);
    ```
Then `FreqScope` has a bottom margin on all platforms. On macOS, it appears as follows (I think this could be a viable option.):
스크린샷 2025-06-05 12 55 43

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
prko: >I said the code is untested.

Yes, I see! I just reported the test result.

>You likely have .onClose calling on the wrong object.

No, I didn’t modify anything; I only replaced the corresponding part.

prko:
@dyfer 
I removed platform switch in https://github.com/supercollider/supercollider/pull/6909/commits/dd9b748383b7bd790e2935c30f8b69a5de26c9fc

The results are as follows:
스크린샷 2025-06-05 13 04 59
스크린샷 2025-06-05 13 05 13
스크린샷 2025-06-05 13 07 27

dyfer:
TL;DR:
`Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height` seems to be a way to get the offset for the height parameter, taking taskbar into account. Please test to confirm and if it works okay, use it as an offset for both FreqScope and ServerMeter. AFAICT it works correctly on both Windows and MacOS.

---

> > Again, I don't believe we should have any Platform switches here (but I have not tested all the possibilities).
> 
> This issue seems to arise because the taskbar in Windows is positioned at the bottom of the screen by default, whereas similar UI elements in other operating systems are typically located at the top. Isn't `Platform.case` designed precisely to address such OS-specific differences?

Not really. `Platform` is designed to handle structural differences between OSs, like which system commands to call.

We should be able to get the available screen bounds using existing abstractions (here: `Window.availableBounds` and `Window.screenBounds`). If we can't, then it's an issue that should be fixed. It would be silly if we couldn't do that...

As for Windows, we also don't know where the taskbar is and we absolutely shouldn't request the user to move it.

For the same reason, "magic numbers" to offset a position is not a good idea, because it makes arbitrary assumptions (e.g. the size and position of the taskbar, or the fact that it could be hidden).

Does this make sense?

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
dyfer: > > You likely have .onClose calling on the wrong object.
> 
> No, I didn’t modify anything; I only replaced the corresponding part.

Well, I don't know what's going on, but I'm sure you can find the culprit. Make sure that the variable that is supposed to hold the Window/View actually holds it and that it is not nil.

prko:
@dyfer 
>Does this make sense?

Yes, thank you for letting me know it!

---

>Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height

- This works on macOS, Linux and Windows for `ServerMeter`
- This works on macOS and Linux for `FreqScope`, but not on Windows.

`Window.availableBounds.top` returns `0` on Windows. The results are as follows:
Screenshot 2025-06-06 at 05 31 39
Screenshot 2025-06-06 at 05 43 49
Screenshot 2025-06-06 at 05 43 52

dyfer:
> `Window.availableBounds.top` returns `0` on Windows. The results are as follows:

But you should use `Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height`. It returns 48` on my Windows VM.

dyfer:
> Yes, the following is the code used for `FreqScope` on the three platforms in [#6909 (comment)](https://github.com/supercollider/supercollider/pull/6909#issuecomment-2946168963):

I _believe_ you forgot to add the height of the Window to the moveTo command. The Window's anchor is still top left, but it's position on screen is calculated from _bottom_ left.

prko:
>I believe you forgot to add the height of the Window to the moveTo command. The Window's anchor is still top left, but it's position on screen is calculated from bottom left.

I am very confused. Could you please check the commit https://github.com/supercollider/supercollider/pull/6909/commits/55f60c2781fd3b9df52ca21a9a7b2f791c4e0684 which is used to produce the screenshots in https://github.com/supercollider/supercollider/pull/6909#issuecomment-2946168963?

Using this commit, `FreqScope` shows the labels `-84` and above on Windows. Before this PR, `FreqScope` shows only the labels `-72` and above on Windows. To let `FreqScope` show the whole screen on Windows, the corresponding part in `FreqScope`should be changed as follows (There is one more possibility to use `Platform.case`, but I do not write it here again since this is not appropriate to use here.):
```
window = Window("Freq Analyzer", rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
				.moveTo(
					5,
					if (Window.availableBounds.top > 0) {
						Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height
					}{
						Window.screenBounds.height - Window.availableBounds.height + 30
					}
				),
				false
			);
```
The screenshot on Windows:
Screenshot 2025-06-06 at 08 57 13

prko:
The result of https://github.com/supercollider/supercollider/pull/6909#issuecomment-2947139613 on Windows:
Screenshot 2025-06-06 at 09 20 59

The variable `rect` is used multiple times after those lines, I think the value of rect should not be changed here....

prko:
```
rect = rect.debug(0).resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4).debug(1);
			rect.moveTo(5, Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height + rect.height);
			window = Window("Freq Analyzer", rect.debug(3), false);
```
RESULT:
```
0: Rect(0, 0, 522, 300)
1: Rect(0, 0, 604, 328)
3: Rect(0, 0, 604, 328)
```

So 10 `rect`variables in the following part after these lines are Rect(0, 0, 604, 328):
```
			freqLabel.size.do({ arg i;
				freqLabel[i] = StaticText(window, Rect(pad[0] - (freqLabelDist*0.5) + (i*freqLabelDist), pad[2] - 10, freqLabelDist, 10))
					.font_(font)
					.align_(\center)
				;
				StaticText(window, Rect(pad[0] + (i*freqLabelDist), pad[2], 1, rect.height))
					.string_("")
				;
			});

			dbLabel.size.do({ arg i;
				dbLabel[i] = StaticText(window, Rect(0, pad[2] + (i*dbLabelDist), pad[0], 10))
					.font_(font)
					.align_(\left)
				;
				StaticText(window, Rect(pad[0], dbLabel[i].bounds.top, rect.width, 1))
					.string_("")
				;
			});

			scope = FreqScopeView(window, rect.moveBy(pad[0], pad[2]), server);
			scope.xZoom_((scope.bufSize*0.25) / width);

			setFreqLabelVals.value(scope.freqMode, 2048);
			setDBLabelVals.value(scope.dbRange);

			Button(window, Rect(pad[0] + rect.width, pad[2], pad[1], 16))
				.states_([["stop", Color.white, Color.green(0.5)], ["start", Color.white, Color.red(0.5)]])
				.action_({ arg view;
					if(view.value == 0, {
						scope.active_(true);
					},{
						scope.active_(false);
					});
				})
				.font_(font)
				.canFocus_(false)
			;

			StaticText(window, Rect(pad[0] + rect.width, pad[2]+20, pad[1], 10))
				.string_("BusIn")
				.font_(font)
			;

			NumberBox(window, Rect(pad[0] + rect.width, pad[2]+30, pad[1], 14))
				.action_({ arg view;
					view.value_(view.value.asInteger.clip(0, server.options.numAudioBusChannels));
					scope.inBus_(view.value);
				})
				.value_(busNum)
				.font_(font)
			;

			StaticText(window, Rect(pad[0] + rect.width, pad[2]+48, pad[1], 10))
				.string_("FrqScl")
				.font_(font)
			;
			PopUpMenu(window, Rect(pad[0] + rect.width, pad[2]+58, pad[1], 16))
				.items_(["lin", "log"])
				.action_({ arg view;
					scope.freqMode_(view.value);
					setFreqLabelVals.value(scope.freqMode, 2048);
				})
				.canFocus_(false)
				.font_(font)
				.valueAction_(1)
			;

			StaticText(window, Rect(pad[0] + rect.width, pad[2]+76, pad[1], 10))
				.string_("dbCut")
				.font_(font)
			;
			PopUpMenu(window, Rect(pad[0] + rect.width, pad[2]+86, pad[1], 16))
				.items_(Array.series(12, 12, 12).collect({ arg item; item.asString }))
				.action_({ arg view;
					scope.dbRange_((view.value + 1) * 12);
					setDBLabelVals.value(scope.dbRange);
				})
				.canFocus_(false)
				.font_(font)
				.value_(7)
			;
```

When not redefining `rect` in these lines, 
```
rect = rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4);
			rect.moveTo(5, Window.screenBounds.height - Window.availableBounds.top - Window.availableBounds.height + rect.height);
			window = Window("Freq Analyzer", rect, false);
```
the result is as follows (it is also not expected):
Screenshot 2025-06-06 at 11 00 15

telephon:
This doesn't read well yet. We should have a method 

```supercollider
Window {
*theNameOfWhichIDontKnow {
     ^this.screenBounds.height - this.availableBounds.top - this.availableBounds.height
}
}

**Comment by prko** [comment]
To display the complete window of `FreqScope` on Window 11:
```supercollider
Window {
	...
	*availableBottom {
		^if (Window.availableBounds.top > 0) {
			this.screenBounds.height - this.availableBounds.top - this.availableBounds.height
		}{
			this.screenBounds.height - this.availableBounds.height + 30
		}
	}
}
```

To hide `-90` and `-96` labels of `FreqScope` on Window 11:
```supercollider
Window {
	...
	*availableBottom {
		^this.screenBounds.height - this.availableBounds.top - this.availableBounds.height
	}
}
```

capital-G:
What happens if windows users put their taskbar on the left side of the screen? We shouldn't hardcode such values as they imply the desktop configuration of a user.

We should probably expose https://doc.qt.io/qt-6/qscreen.html#availableGeometry-prop via QtCollider for this, so it is still possible to create fullScreen GUIs (current way) or to stay within the bounds of the available geometry (new way).

prko:
On Windows 11, I am unable to find any options for adjusting the taskbar height and location.
https://support.microsoft.com/en-us/windows/customize-the-taskbar-in-windows-0657a50f-0cc7-dbfd-ae6b-05020b195b07
On this page, we can find `location` and `height` when selecting Windows 10, but we cannot find thoese terms when selecting Windows 11.

The following is the answer of Microsoft Copilot AI:
>In Windows 11, you can't freely move the taskbar to different positions, but you can adjust its size. Here's how:
>
>### Changing Taskbar Size (Using Registry Editor)
>1. Press **Win + R**, type `regedit`, and open the Registry Editor.
>2. Navigate to: `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced`
>3. Find `TaskbarSi` and set its value to **0 (small), 1 (default), or 2 (large)**.
>4. Restart your computer for the changes to take effect.
>
>### Changing Taskbar Position (Using Explorer Patcher)
>Windows 11 does not natively allow moving the taskbar, but **Explorer Patcher** can help.
>1. Download **Explorer Patcher** from [GitHub](https://m.blog.naver.com/pascalbed01/223557290833).
>2. Install it and adjust the settings to move the taskbar to your preferred position.

I've customised a lot of things in Windows from my 20s to my 40s, but now I don't feel like doing it anymore. I'm not sure how Windows users these days approach customisation.

prko:
>We shouldn't hardcode such values as they imply the desktop configuration of a user.

OK, I will not resist it.

---

I uploaded a new commit again reflecting all reviews.
For `ServerMeter`, there are still two fixed values:
`328` is the height of `FreqScope`, and `60` is the best value for all platforms not to collide `s.freqscope` and `s.meter`.  The corresponding value for `328 + 60` was already hardcoded before this PR. There is no way to calculate the height of `FreqScope` without creating it. 

The result of https://github.com/supercollider/supercollider/pull/6909/commits/48ea8256c31535867651bcc07493aa1f59213b84:

Screenshot 2025-06-07 at 19 28 51
Screenshot 2025-06-07 at 19 28 31
Screenshot 2025-06-07 at 19 28 27

prko:
A quick question.

How about the following?
```
Window {
	...
	*availableBottom {
		^if (Window.availableBounds.top > 0) {
			this.screenBounds.height - this.availableBounds.top - this.availableBounds.height
		}{
			this.screenBounds.height - this.availableBounds.height * 1.6
		}
	}
}
```
`* 1.6` is not a hardcoded number!

capital-G:
> `* 1.6` is not a hardcoded number!

What else is it? ;)

BTW - there is no way of knowing how much the header of a window will consume in pixels as this can also be tied to user preferences etc. (e.g. font size of window names)

If you want to have a non-overlapping server meter and a freqscope across all platforms it is the easiest to put both into a shared window, b/c the content of a window should be deterministic across platforms thanks to Qt. Everything outside its canvas is dependent on the OS.

capital-G:
> > What else is it? ;)
> 
> 1.6 is the factor needed to display the full FreqScope window on Windows 11. It's a pity for Windows users who would always have to manually adjust the FreqScope.

I totally agree with this and IMO it should be fixed - yet the solution should not be tied to user settings etc. - hardcoded constants like `1.6` are likely to break in future or older versions of windows and maybe ignore some user specific settings (at the same time, the whole QtCollider code is full of constants...)

In the end I don't have enough GUI and sclang experience to know what a good solution could be :/ I think one way could be to have an additional constructor for rect which can take the DE boundaries into account?

I think its currently best to wait until someone from the sclang team proposes how to go on here?

jamshark70:
> If you want to have a non-overlapping server meter and a freqscope across all platforms it is the easiest to put both into a shared window...

I'd support this. It's a weakness in the current FreqScope that the controls are put directly into a Window, not a View. If that were changed, then it would be easy to make a single window for both.

prko:
Previously, I placed the following code in `startup.scd`:
```
(
var b, wA, wAB, v, vAudio, vNodes, meter;
b = Rect(0, 0, 368, Window.screenBounds.height);
wA = Window("test", b);
wAB = wA.bounds;
v = View(wA, Rect(0, 0, wAB.width, wAB.height));
meter = ServerMeterView;
vAudio = View(v, wAB.width @(meter.height * 2));
vNodes = View(
    v,
    Rect(
        0,
        meter.height * 2,
        wAB.width,
        wAB.height - (meter.height * 2)
    )
);
ServerMeterView(s, vAudio, 0@0, 2, 2);
Stethoscope(s, 2,
    view: UserView(
        vAudio,
        Rect(
            meter.getWidth(2, 2), 0,
            v.bounds.width
            - meter.getWidth(2, 2),
            vAudio.bounds.height / 2 - 25
        )
    )
);
FreqScopeView(
    vAudio,
    Rect(
        3,
        4 + meter.height,
        v.bounds.width - 6,
        meter.height - 10
    )
)
    .active_(true).freqMode_(1);
s.plotTreeView(0.5, vNodes);
wA.front
)
```
K03_04_-_server_viewers

However, this is not always needed.....

dyfer:
Layout would respect the parameter .minWidth_() of the child view, I think.
However I'm not sure how to best implement it so that the individual view is collapsed when the given widget is not displayed...

dyfer:
Before more effort is spent on this I wanted to point out the following:

- it would be good to prevent the widgets to not be covered by the taskbar (*)
- at the same time, I don't think it's valid to simply assume that the meter should be positioned high enough not to interfere with the freqscope; it doesn't work well if one is using just the meter primarily
- one thing that could be considered is that the window positions could be remembered... see Scott's [WindowViewRecall quark](https://github.com/scztt/WindowViewRecall.quark/blob/master/WindowViewRecall.sc); that way users would need to move them to where they like, but only once
- users may or may not want to have all the views in one Window
- if we decided to add functionality to put various views in one window, IMO it would need to be configurable in many ways: whether it happens or not (i.e. old vs new behavior), what the layout is, etc etc.; these settings should persist between sessions (probably using `Archive` object)


(*) currently we can [calculate the available bottom left boundary](https://github.com/supercollider/supercollider/pull/6909#issuecomment-2952078717), but we don't have a way to get the size of the window's title bar which we'd need to get the accurate position of the _Window's top left corner;_ that value would need to be hardcoded for now, unless we add functionality in Qt to get it.

prko:
>Window.availableBottom method (is this the name we want?)

The name is based on the `Window.availableBounds`. I do not think `Window.availableBottom` is best for `^this.screenBounds.height - this.availableBounds.top - this.availableBounds.height`, but I cannot find a better one.

>Window.availableBottom + 20

I think `Window.availableBottom + 32` or `Window.availableBottom + 31` is better than `Window.availableBottom + 20`, but I followed your suggestion:

Screenshot 2025-07-13 at 08 05 59

dyfer:
> I think `Window.availableBottom + 32` or `Window.availableBottom + 31` is better than `Window.availableBottom + 20`,

The behavior observed on screenshots (and confirmed on a Windows VM) makes no sense - `Window.avaliableBottom` _alone_ should've cleared the taskbar.

I was trying to find out why and it turns out that when set to be non-resizable, Window's position is shifted on screen: https://github.com/supercollider/supercollider/issues/7044. We should fix that separately, unrelated to this PR.

@telephon do you think the `Window.avaliableBottom` should be an addition in this PR or separate?

If it's to be done in this PR, then the commit history would need to be cleaned up through [interactive rebase](https://github.com/supercollider/supercollider/wiki/Rebasing-a-PR#rearranging-combining-commits-or-editing-commit-messages-with-rebase), squashing related commits, and a force push, so that `Window` addition and `FreqScope` change are in separate commits.

prko:
@dyfer I think I resolved the issue:

Could I update this PR with the following three modifications?

```
Window {
	...
	
	*availableBottom {
		^this.screenBounds.bottom - this.availableBounds.bottom
	}
	
	...
}
```
```
FreqScope {
	...
			window = Window("Freq Analyzer", resizable: false).bounds_(
 				rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4)
 				.moveTo(5, Window.availableBottom)
			);
	...
```
```
ServerMeter {
 	...
		window = Window.new(server.name ++ " levels (dBFS)", resizable: false).bounds_(Rect(
			5 + Window.availableBounds.left,
			Window.availableBottom + 365,
			ServerMeterView.getWidth(numIns, numOuts),
			ServerMeterView.height
		));
 	...
```
Screenshot 2025-07-19 at 13 34 46
Screenshot 2025-07-19 at 13 35 32
Screenshot 2025-07-19 at 13 36 08
Screenshot 2025-07-19 at 13 46 40
Screenshot 2025-07-19 at 13 48 49
Screenshot 2025-07-19 at 13 49 26

telephon:
> @telephon do you think the `Window.avaliableBottom` should be an addition in this PR or separate?

I think it is ok to bundle new methods that come with a refactoring for solving a problem. It shows why a new feature is intriduced and also tests it right away.


> The name is based on the `Window.availableBounds`. I do not think `Window.availableBottom` is best for `^this.screenBounds.height - this.availableBounds.top - this.availableBounds.height`, but I cannot find a better one.

Yes, indeed, the name `Window.avaliableBottom` doesn't speak so well yet.

```
*availableBottom {
		^this.screenBounds.bottom - this.availableBounds.bottom
	}
```
shows that it is a difference between the bottom of the screen and the bottom of the available bounds.

I don't understand what this difference is, maybe you can explan and then the name might be more obvious?

prko:
>I don't understand what this difference is, maybe you can explan and then the name might be more obvious?

```
^this.screenBounds.bottom - this.availableBounds.bottom
```
is noticeably shorter than
```
^this.screenBounds.height - this.availableBounds.top - this.availableBounds.height
```

This calculation is used to determine the bottom position of a window that avoids overlapping with system elements such as the taskbar or dock.

 I believe we can simply use `this.screenBounds.bottom - this.availableBounds.bottom` without needing to introduce a dedicated method for `Window`.

However, if we opt for the longer expression `this.screenBounds.height - this.availableBounds.top - this.availableBounds.height`, then defining a method for `Window` might be worthwhile for the sake of readability and reuse.

Regarding the name of the method, I have no good idea, but I list some possible ones:
- getVisibleBoundsBottom
- visibleBoundsBottom
- visibleBottom

prko:
>Yes, I think so. What is the 356 ?

The height of the FreqScope window cannot currently be calculated without creating the window itself, so this hardcoded value is necessary. 356 seems to be a good value for all platforms (The resolution of the following screenshots are 1920*1080):

Screenshot 2025-07-20 at 06 14 15
Screenshot 2025-07-20 at 06 04 42
Screenshot 2025-07-20 at 06 05 11
Screenshot 2025-07-20 at 06 09 05
Screenshot 2025-07-20 at 06 09 35

prko:
@telephon 
>Maybe now, you could write a comment that this is the height of the FreqScope?

Done in https://github.com/supercollider/supercollider/pull/6909/commits/4f886bf0baa5f17e4af9a84f8fcc06d076e87158

---

@dyfer 
What do you think about these changes?

SCClassLibrary/Common/GUI/Base/QWindow.sc
```diff
@@ -172,9 +172,16 @@ Window {
...
+	*lowestPosition {
+ 		^this.screenBounds.bottom - this.availableBounds.bottom
+ 	}
+
+	moveToBottom { |margin = 0|
+		this.bounds = this.bounds.moveTo(
+			this.bounds.left,
+			Window.lowestPosition + margin
+		)
+    	}
```
telephon: indentation

telephon (changes requested):
just formatting

SCClassLibrary/Common/GUI/Base/QWindow.sc
prko: done in https://github.com/supercollider/supercollider/pull/6909/commits/92daf223de09e9a65112a913b24fc1ed8650d60c

Thanks!

telephon (approved):

telephon (changes requested):
documentation needed.

prko:
>Please rebase this PR so that there's one commit for Window addition and one for FreqScope changes. This can't be done automatically when merging.

Do you mean adding
- one commit for Window addition,
- one for ServerMeter changes and
- one for FreqScope changes
after deleting all previous commit history?

prko:
@telephon 
Documentation is added in https://github.com/supercollider/supercollider/pull/6909/commits/feece6c3c5d0b8acfdb74c68d73b092cc334b911

prko:
>Please rebase this PR so that there's one commit for Window addition and one for FreqScope changes. This can't be done automatically when merging.

Do you mean adding
- one commit for Window addition with documentation,
- one for ServerMeter changes and
- one for FreqScope changes
after deleting all previous commit history?

dyfer:
> Are you suggesting that we should:
> 
> * add one commit for the Window addition with documentation,
> * another for the ServerMeter changes, and
> * one more for the FreqScope modifications,

I think ServerMeter and FreqScope can be in the same commit, these are both changes to server-related guis.

Window addition + docs should be in a separate commit as this is a more general GUI change.

> after removing all previous commit history?

Typically you'd `squash` related commits into a single one, instead of removing them (removing them may make the rebase fail). I suggest to try it on a copy of the branch if you've never done it.

<details><summary>The general idea is:</summary>
Let's say you have 4 commits, two about feature A and two about feature B, when you enter interactive rebase you'd get:

```
pick <sha> feature A first commit
pick <sha> feature A second commit
pick <sha> feature B first commit
pick <sha> feature B second commit
```

What you want to do is to squash first two and second two commits, e.g.

```
pick <sha> feature A first commit
squash <sha> feature A second commit
pick <sha> feature B first commit
squash <sha> feature B second commit
```

After this you'll end up with only two commits. You may also need to edit their descriptions (also in interactive rebase, using the `reword` directive). 

Note: if you reorder the commits, you may have conflicts when squashing. Make sure that you only squash commits touching the same set of files.

Interactive rebasing uses `Vim` text editor by default. Using Vim is adventure on its own... Pressing `i` enters "insert mode" to edit text, double Esc exist from that mode; `shift-d` cuts the whole line, `p` pastes the line (when NOT in insert mode). Esc, then `:wq` writes the file and exits Vim.

After you're done with rebase, you need to **force push** the branch. This overwrites the contents of the branch on the `remove` (server).

Again, I suggest that you make a copy of your branch and try it there first.

See git instructions on interactive rebasing. Let me know if this makes sense...
</details>

telephon (commented):
Thank you! Btw. does this value change when you move the dock to the side?

prko:
>Btw. does this value change when you move the dock to the side?

Yes. Please refer to the screenshots in https://github.com/supercollider/supercollider/pull/6909#issuecomment-3092575879. Especially,
- 2 Linux screenshots
- 2 macOS screenshots

More screenshots on macOS:
- Dock is on the left side, and automatic hide is activated:
Screenshot 2025-07-21 at 07 02 18

- Dock is on the left side, and automatic hide is not activated:
Screenshot 2025-07-21 at 07 02 00

prko:
@dyfer 

Thanks for https://github.com/supercollider/supercollider/pull/6909#issuecomment-3094762046

I will do it later. It seems a bit complicated, and I need time and concentration to do it.

telephon:
Thank you @prko – screenshots were not necessary, I would have believed you :)

PR #6910, opened 2025-05-30
Enhance User-Wide Build Instructions in README_LINUX.md

prko:
## Purpose and Motivation

Today, I changed some sc files on Linux, but had difficulties using the changed files. I could not overwrite the files since I built SC as a system-wide installation. I tried to do a user-wide installation, but I could not use SC properly. 
- Double-clicking the scide file in `/home/parallels/usr/local/bin/scide`: 
   >Interpreter has quit. [Exit code: 0]
- Typing the following command in `/home/parallels/usr/local/bin/`in a terminal window:
   ```shell
   scide
   ``` 
   >Interpreter has quit. [Exit code: 0]

After a longer time with many tries (including reinstalling Linux), I realised that I should reboot after user-wide installation.
This PR add some explanation for other users who, like me, do not know such a convention.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing


JordanHendersonMusic:
I don't have this problem. Perhaps this is an issue with your linux setup?

capital-G (commented):
The content of the PR doesn't seem to match its description?

In general, you wouldn't need to restart your linux laptop, you simply need to refresh your shell using the `source` command - although a login and logout of your current user would also achieve the same.

prko:
The users who start SuperCollider and Linux at the same time might be not yet familiar with editing ~/.profile or handling hidden files through the terminal. In such cases, they will likely search the internet or ask an AI assistant for help. Adding this explanation reduces that extra effort and helps beginners follow the build instructions more smoothly.

JordanHendersonMusic:
At some point we are no longer documenting supercollider, but just documenting basic Linux behaviour. 

Think about this documentation in terms of dependencys. If any of those distros changes their file browser app this document goes out of date.

prko:
>Think about this documentation in terms of dependencys. If any of those distros changes their file browser app this document goes out of date.

I removed the specific names of GUI application to avoid distro‑dependent details.

---

>At some point we are no longer documenting supercollider, but just documenting basic Linux behaviour.

I understand the concern. My intention is not to document general Linux usage, but to reduce unnecessary friction for beginners or for users who run SuperCollider on Linux only occasionally.
From my own experience—using Ubuntu mainly for testing—I often had to spend extra time searching for very basic steps when launching SC from a graphical environment.
Including a minimal, distro‑agnostic clarification in the README can help such users without tying the documentation to any specific Linux setup.

capital-G:
I share Jordans view. IMO looking things up is part of the linux journey, which is simply more fragmented and involves more manual adjustment than windows/macOS - we can not document every edge case, such as how to edit files - especially since distros such as arch doesn't have a `~/profile` file IIRC.

I hope that #7569 will serve as an easier way for ocassional linux/sc users.

prko:
OK, I close this! Thanks. ^^

JordanHendersonMusic:
I think a linked forum post might be a better way to help users? That's far more flexible and easy to update when things change.

capital-G:
> I think a linked forum post might be a better way to help users? That's far more flexible and easy to update when things change.

But this would just start duplicating content at some point which will be annoying to keep in sync ;)

prko:
> a linked forum post 

Regarding the linked forum post:

This is the part I really do not understand. Some questions on the forum are repetitive, and in the past, the forum was very active. Nowadays it is less active, but several users — I assume from the moderator group — still respond with a lot of enthusiasm.
It seems more efficient if some of this energy and time could be directed toward improving the official documentation instead.

The forum appears to be losing momentum, and I think this may be related both to the overall decline in activity and to the fact that the help documentation is not very beginner‑friendly. Of course, it is possible that the number of SuperCollider users has increased thanks to various YouTube resources, while the number of forum participants has decreased.
However, since I often feel that the available documentation is insufficient, I keep coming back to the conclusion that the help system needs to be more accessible and welcoming.

Back around 2010, I believed that SuperCollider would eventually become more widely used than Max. I do not check the Max forum often these days, so I cannot say for certain, but my impression is that Max has become more widely adopted.
It feels unfortunate that SuperCollider seems to have drifted into a somewhat elitist culture. I wish the documentation and the community were more approachable, so that more people could feel comfortable engaging with it.

JordanHendersonMusic:
> It seems more efficient if some of this energy and time could be directed toward improving the official documentation instead.

Imo, SC doesn't need more documentation, it needs less dumb behaviour and better tooling. Most people don't read docs, mostly because they are hard to navigate when you don't know what to search.

---

If a forum post goes out of date a user can just comment on it, and I'm sure some one will pick it up. It far lower stakes.

---

> It feels unfortunate that SuperCollider seems to have drifted into a somewhat elitist culture

I'm literally in the midst of rewriting a large part of the compiler just so we can nice error message and eventually good tooling, and Dennis has just released the wasm version vastly expanding the user base.

You directed that comment at two people *actively* fighting against nasty legacy code to try and make SC more open and friendly. I haven't met a single developer who doesn't want this, the problem is that it's really hard when you have 25 years of technical debt and can't break backwards compatibility.

prko:
Sorry — I really didn’t mean to criticise either of you, and I apologise if my wording came across as hurtful. That wasn’t my intention at all.

I genuinely appreciate the work you’re doing. I love the recent improvements such as the WASM port, the rich text editor, the enhanced error messages, and even things like arbitrary keyword arguments in function definitions. These changes make SC feel much more modern, and I’m getting the sense that we may not even need an “SC4” — simply maintaining and enhancing SC as you’re doing already brings us much closer to what many users want.

Recently, I saw a forum post where a user mentioned that SuperCollider felt too complex for them. That made me think that easy and kind documentation still plays an important role, which is why I restarted the stuck PR to rework it. I just hoped this could also contribute to making SC more approachable.

But I completely agree with your point — from a normal user’s perspective, helpful and kind answers on the forum often feel more vivid and accessible. Thank you again for all your hard work!

capital-G:
> This is the part I really do not understand. Some questions on the forum are repetitive, and in the past, the forum was very active. Nowadays it is less active, but several users — I assume from the moderator group — still respond with a lot of enthusiasm.

I think this is also accelerated due to AI - see e.g. https://meta.stackoverflow.com/questions/437921/how-does-the-continued-decline-in-posts-since-may-25-influence-our-interpretati [^1]

[^1]: IMO our moderation could also be different and is also a reason why we lost some valuable contributors to the forum and is also the reason why I don't want to interact more than necessary w/ the forum. These concerns have been voiced multiple times, but nothing has changed and I gave up on this front - though r/n it is rather calm which I appreciate.

> However, since I often feel that the available documentation is insufficient, I keep coming back to the conclusion that the help system needs to be more accessible and welcoming.

+1

SC is a great project, but the docs are not really targeted at beginners [if they even exist and not locked away in an expensive book, looking at you, scsynth and sclang^^]. The docs lack consistency, bigger scopes and good introductions into broader techniques.
Though: SC will always be full of "inconsistencies", due to its backwards compatibility promise and multiple people working w/ different "dialects" on it over the years - this is just part of the fun/identity, though it is indeed confusing/frustrating at the beginning, but can also be really inspiring as it demonstrates a multiplicity of approaches, so people can find their own preference. I think communicating this to the beginners upfront can resolve some "frustrations".

I think all devs are vary of feature-creep (at least nowadays^^), so there is r/n a push-back against expanding the scope of the project b/c of limited dev resources - e.g. we don't want to maintain a document where linux file editing of hidden files gets explained b/c there are already plenty of resources which explain this and we can't explain every combination in linux and we simply have to defer to the abilities of the user at some point.

> Back around 2010, I believed that SuperCollider would eventually become more widely used than Max. I do not check the Max forum often these days, so I cannot say for certain, but my impression is that Max has become more widely adopted.

Based on my personal xp/surroundings I'd guess 90% max/msp w/ 5% for pd and sc each - I would be surprised if this was/is a recent development. But strudel is taking a lot of the cake r/n ;)

> It feels unfortunate that SuperCollider seems to have drifted into a somewhat elitist culture.

image

joke aside: where max/msp is driven by commercial interests [and therefore has a strong interest in new users pouring in], pd by the vision of their bdfl (plz correct me if I am wrong here!), SC is driven by a group of dedicated people who have been around for some time - this shifts the focus to advanced/niche features and maintenance instead of the smoothest onboarding procedure.

nonetheless, as jordan mentioned, there is a lot of effort put into making sc nicer for all people, and I think people w/ less experience will benefit most of them.

> Imo, SC doesn't need more documentation, it needs less dumb behaviour and better tooling. Most people don't read docs, mostly because they are hard to navigate when you don't know what to search.

I read docs all the time and I think it is pretty crucial for sc b/c of its lacking type system (as well as digging into the source code - but this is only possible in some areas of the classlib^^)
I think once we have a native LSP we can expose the existing docs much easier to the user and can provide better assistance.

PR #6913, opened 2025-05-31
Rename ServerTreeView to NodeTreeView

### LINKED ISSUE #6853: Rename ServerTreeView
## Motivation

As I was reviewing #6834, it realized that the recently added a class is called `ServerTreeView`, but which is not related to `ServerTree` in a way that e.g. `FreqScopeView` is related to `FreqScope`.

I wasn't following the discussion in #6294 and I'm not sure how it relates to the previously proposed `.plotTree` rework in #6545, but the latter PR proposed a pair of classes: `ServerNodeTree` which uses a `NodeTreeView`.

I believe that `ServerTreeView` should be changed, as it doesn't really relate to the existing class `ServerTree`.  One option would be to use `NodeTree` or `ServerNodeTree`, as that name is suggested in #6545. However #6545 proposes a design that seems to follow better (at least at a first glance) the established practices in e.g. `FreqScope`/`FreqScopeView`.

## Plan for Implementation

Let's update the structure of the current implementation (i.e. class names and relationships) so that it follows established practices. We need to do this before 3.14 is release, which is planned at the end of May 2025. Once we introduce a new class in a release it's difficult to change its interface later. 

## Current view of ServerTreeView

Image

dyfer:
## Purpose and Motivation

Fixes #6853 

## Types of changes

- Bug fix (not really...)

## To-do list

- [x] Code is tested


dyfer:
@capital-G can you please approve/merge ?

PR #6914, opened 2025-05-31
Update node tree view help / implementation

dyfer:
## Purpose and Motivation

After merging #6834 I found some discrepancies between the documentation and the implementation.

- `s.plotTree` returns a `NodeTreeView`, as [introduced](https://github.com/supercollider/supercollider/pull/6834/files#diff-a35b5be480643c127f781e8f7c8cc60f263ca1af432aa3c881ba3d043623e6c0R377) in #6834. I reverted the previous change in this PR as suggested by @telephon.

- I made the `s.plotTreeView` (the method that actually creates a view, but that's not typically called directly by the user) return the newly created instance of the NodeTreeView. In SC 3.13 this return a function that needed to be called to clean it up, or something like that. This is technically a _breaking change_, comparing with the state in SC 3.13. I don't think this function is meant to be used directly by the users anyway, and it's not sound-related so I'm okay with this "breaking change" (btw breaking change was introduced first in #6834)

- **note slight interface change**: `s.nodeTreeView` instance variable is now updated by the `s.plotTree` method, NOT by `s.plotTreeView` method. I think that's clearer? But i'm open to feedback on this. Also, both methods still return an instance of `NodeTreeView`.

- ~~examples in the docs assumed that `s.plotTree` returns the `NodeTreeView`, or its parent `Window`, contrary to the previous statement about `s.plotTree` returning an instance of the server.~~ I updated the examples to get an instance of the view directly from the server - `s.nodeTreeView` - instead of assigning it to a variable.

- I removed the note about `s.plotTree` window automatically closing when the server is not running (this behavior was changed in #6834)

~~This PR depends on #6913, let's merge that one first.~~ done!

## Types of changes

- Documentation
- Breaking change (kind of)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/NodeTreeView.schelp
+++ b/HelpSource/Classes/NodeTreeView.schelp
@@ -6,9 +6,6 @@ related:: Classes/Server, Classes/ServerTree
 description::
 link::Classes/NodeTreeView:: shows the node tree on a server.
 
-note::
-Unlike to link::Classes/Server#-plotTree::, the link::Classes/NodeTreeView:: will not disappear when the pointed server is not running.
-::
 
 classmethods::
 
--- a/HelpSource/Classes/Server.schelp
+++ b/HelpSource/Classes/Server.schelp
@@ -672,56 +672,41 @@ note::
 To use multiple instances of link::Classes/NodeTreeView:: per link::Classes/Server:: instance or workspace, refer to the link::Classes/NodeTreeView#examples#Examples in NodeTreeView Documentation::.
 ::
 
-The method returns a link::Classes/Server:: (the receiver of this call), and the created link::Classes/View:: instance can be accessed via the link::#-nodeTreeView::.
-
-note::
-If the server is not running, the window containing teletype::plotTree:: will briefly appear and then close, displaying the following message in the Post window:
-teletype::
--> localhost
-WARNING: Server failed to respond to Group:queryTree!
-::
-::
+This method returns an instance of link::Classes/NodeTreeView::, which can also be accessed via link::#-nodeTreeView:: method. Before version 3.14, this method returned an instance of the Server itself.
 
 code::
 // start the SuperCollider server:
-s.boot
+s.boot;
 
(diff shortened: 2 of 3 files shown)


dyfer:
@prko @telephon @capital-G this is a minor cleanup after the recent s.plotTree rework.

prko:
Thank you for the update. Everything looks good, but I do have one question.

Is there no need to inform users about the 'plotTree' function?

The server documentation provides information about NodeTreeView, but there is no mention of plotTree in the NodeTreeView documentation. Would it be beneficial to include a reference to it to help users determine whether they are using it properly?

dyfer:
> Is there no need to inform users about the 'plotTree' function?

Possibly yes. But this PR focuses on fixing the discrepancies between the documentation and implementation, not on improving the documentation in itself. 

I think my main concern to consider is whether `s.plotTree` should return a `Server`, or a `NodeTreeView`. 
- Argument for returning a `Server`: this is what `s.plotTree` was returning previously. If we change it, we'd break code like `s.plotTree.meter` (though this is a minor issue IMO);
- Argument for returning a `NodeTreeView`: `s.meter` and `s.freqscope` return a `ServerMeter` and a `FreqScope`, respectively, so returning a `NodeTreeView` would be more consistent with these functions.

Relatedly, I wonder whether `s.plotTreeView` (the method called internally by `s.plotTree`) should also return a `NodeTreeView`, or a `Server` (this is a change in the interface from 3.13 anyway, but that class is not typically used directly, I think)

dyfer:
> I think my main concern to consider is whether `s.plotTree` should return a `Server`, or a `NodeTreeView`.

@telephon do you have an opinion on this?

telephon:
In general, it should return the NodeTreeView. This is a breaking change, but probably not too severe?

dyfer:
@telephon thanks for the input! I updated the implementation to return NodeTreeView.

I also thought that it would be cleaner to update the class variable in `.plotTree` instead of `.plotTreeView`, what do you think?

PR #6915, opened 2025-05-31
testsuite: TestNetAddr: fix server msg interference

elgiano:
## Purpose and Motivation

Fixes some recent failures in testsuite.

- Server quit messages were interfering with other OSC message tests:
This PR filters out messages sent to an unexpected port. This effectively filters out interfering server messages.
Test would fail anyways if a message is not received on the expected port before timeout.
- Replaced Condition() with CondVar()
- Moved tests requiring a Server to their own file, since not all of them did and they were firing up a server for no reason.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing


testsuite/classlibrary/TestNetAddr.sc
```diff
@@ -53,29 +25,21 @@ TestNetAddr : UnitTest {
...
 
-			this.assert(messageValue != \timeout);
-			this.assertEquals(msg, testMsg, "Correct message received");
-			this.assertFloatEquals(time, AppClock.seconds, "Timestamp is valid", within: 0.1);
-			this.assertEquals(replyAddr.addr, addr.addr, "Hostname is valid");
-			this.assertEquals(replyAddr.port, 57120, "Port is valid");
-			this.assertEquals(recvPort, port, "recvPort is valid");
-		}
+		#msg, time, replyAddr, recvPort = messageValue;
+		this.assert(messageValue != \timeout);
```
mtmccrea: it looks like you've removed code where `messageValue` may be set to \timeout, so this won't ever fail.

IIUC, you can test for the timeout with 
```supercollider
var received;
...
received = messageReceived.waitFor(1) { messageValue.notNil }; 
...
if (messageValue.isNil) {
    this.assert(false, "addr.% was never received".format(sendMethod));
};
#msg, time, replyAddr, recvPort = messageValue;
this.assert(received, "Timeout.", report: false);
...
```
Though states of "received" and "timeout" are identical in this case so a bit redundant.

testsuite/classlibrary/TestNetAddr.sc
```diff
@@ -53,29 +25,20 @@ TestNetAddr : UnitTest {
...
+		messageReceived.waitFor(1) { messageValue.notNil };
 
 		thisProcess.perform(unregisterMethod, func);
 
-		if (messageValue == \timeout) {
-			this.assert(false, "addr.sendRaw was never received");
-		} {
-			#msg, time, replyAddr, recvPort = messageValue;
+		if (messageValue.isNil) {
+			this.assert(false, "addr.% was never received".format(sendMethod));
```
mtmccrea, line 33: optional, replace if statement with: 
```
this.assert(messageValue.notNil, ...
```

elgiano: thanks a lot! however I would prefer to keep the if, so that the assert doesn't run if messageValue.notNil. I also thought a failed assert would stop the test, but since it doesn't, I've put the other tests in the else block. This way, we check fields only if message was received.

mtmccrea: I believe passing an `onFailure` argument to `assert` (like an empty function or anything that responds to `.value`) will error out so any subsequent tests in that thread won't run. Should be tested though if you decide to do that...

elgiano: Ah yes, thanks for reminding! However, I don't like how `onFailure` works, I think it's more readable as it is with the `if/else` check

PR #6916, opened 2025-05-31
sclang, unittest: assert no uninlined functions in class lib.

### LINKED ISSUE #1926: `Post Inline Warnings` should be true by default
See  #1925.

### LINKED ISSUE #6889: class library: contains variable declarations and so will not be inlined.
In the common class library, all functions should be inligned. Currently, there are many warnings on startup:

```
WARNING: FunctionDef contains variable declarations and so will not be inlined.
```

This does not give us a choice but to tick this box off.

I had actually done this, because I used Quarks that triggered so many warnings, and forgot.
I'd suggest that the develop version should always have this ticked by default, so we avoid this.

These are the ones on the to-do list:

- [ ] SCDoc.sc
- [ ] Error.sc
- [ ] Dictionary.sc
- [ ] Server.sc
- [ ] TestCollection.sc
- [ ] TestUGen_RTAlloc.sc

```supercollider
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/SCClassLibrary/SCDoc/SCDoc.sc'
  line 706 char 49:

  					var relativePath = file.asRelativePath(dir);
                                                   
  					var destinationPath = SCDoc.helpTargetDir +/+ relativePath;
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/SCClassLibrary/Common/Core/Error.sc'
  line 187 char 0:

  		keywordArgumentPairs !? {|kw|
  
  			"KEYWORD ARGUMENTS:".postln;
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/SCClassLibrary/Common/Collections/Dictionary.sc'
  line 546 char 0:

          this[selector] !? { |func|
  
  			^func.performArgs(\functionPerformList, [\value, this, args], kwargs);
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/SCClassLibrary/Common/Collections/Dictionary.sc'
  line 559 char 0:

          this[\forward] !? { |func|
  
  			^func.performArgs(\functionPerformList, [\value, this, selector, args], kwargs);
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/SCClassLibrary/Common/Control/Server.sc'
  line 1355 char 31:

  				var freeKb = msg[1] / 1024;
                                 
  				"Used RT memory: % kb".format(options.memSize - freeKb).postln;
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/testsuite/classlibrary/TestCollection.sc'
  line 23 char 0:

  			obj !? {|o|
  
  			    var r;
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/testsuite/classlibrary/TestCollection.sc'
  line 45 char 0:

          			obj !? {|o|
  
          			    var r;
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/testsuite/classlibrary/TestUGen_RTAlloc.sc'
  line 293 char 55:

  				var noneIsInvalid = out.drop(1).any(_ == (-1)).not;
                                                         
  				this.assert(noneIsInvalid, "% successful alloc should return a valid bufnum".format(ugen));
WARNING: FunctionDef contains variable declarations and so will not be inlined.
  in file 'path/supercollider/testsuite/classlibrary/TestUGen_RTAlloc.sc'
  line 303 char 50:

  				var allInvalid = out.drop(1).every(_ == (-1));
                                                    
  				this.assert(allInvalid, "% should only return -1 on alloc fail".format(ugen));
	numentries = 934644 / 16517916 = 0.057
	6498 method selectors, 2542 classes
	method table size 15151896 bytes, big table size 132143328
	Number of Symbols 14416
	Byte Code Size 649389
	compiled 448 files in 1.10 seconds
compile done
Class tree inited in 0.01 seconds
```

JordanHendersonMusic:
## Purpose and Motivation

Uninlined functions are slow, don't use them in the class library. From now on, if you do, you will break this test.

Closes #1926
Closes #6889

## Types of changes

- new test

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/PyrParseNode.h
+++ b/lang/LangSource/PyrParseNode.h
@@ -77,6 +77,10 @@ enum {
 
 extern AdvancingAllocPool gParseNodePool;
 
+// This value count the un-inlined functions, these are not desirable in the class library because they are slow.
+// There is a primitive that returns this value so it can be checked in sclang's unit tests.
+extern int gNumUninlinedFunctions;
+
 #define ALLOCNODE(type) (new (gParseNodePool.Alloc(sizeof(type))) type())
 #define ALLOCSLOTNODE(type, classno) (new (gParseNodePool.Alloc(sizeof(type))) type(classno))
 #define COMPILENODE(node, result, onTailBranch) (compileNode((node), (result), (onTailBranch)))
--- a/testsuite/classlibrary/TestUGen_RTAlloc.sc
+++ b/testsuite/classlibrary/TestUGen_RTAlloc.sc
@@ -284,23 +284,25 @@ TestUGen_RTAlloc : UnitTest {
 		var blockSize = server.options.blockSize;
 		var smallBuf = Buffer.alloc(server, blockSize);
 		var bigBuf = Buffer.alloc(server, nextPowerOfTwo(this.memSizeFloats) * 2);
+		var noneIsInvalid, out;
 
 		var testAllocPass = { |ugen, testFunc|
 			var out = this.awaitSynthOutput(testFunc, blockSize * 4);
 			if (out.isNil) {
 				this.assert(false, "% successful alloc test should complete".format(ugen));
 			} {
-				var noneIsInvalid = out.drop(1).any(_ == (-1)).not;
+				noneIsInvalid = out.drop(1).any(_ == (-1)).not;
 				this.assert(noneIsInvalid, "% successful alloc should return a valid bufnum".format(ugen));
 			};
 		};
 
 		var testAllocFail = { |ugen, testFunc|
-			var out = this.awaitSynthOutput(testFunc, blockSize * 4);
+			var allInvalid;
+			out = this.awaitSynthOutput(testFunc, blockSize * 4);
 			if (out.isNil) {
 				this.assert(false, "% failed alloc test should complete".format(ugen));
 			} {
-				var allInvalid = out.drop(1).every(_ == (-1));
+				allInvalid = out.drop(1).every(_ == (-1));
 				this.assert(allInvalid, "% should only return -1 on alloc fail".format(ugen));
 			};
 		};
--- a/SCClassLibrary/Common/Collections/Dictionary.sc
+++ b/SCClassLibrary/Common/Collections/Dictionary.sc
@@ -540,11 +540,13 @@ IdentityDictionary : Dictionary {
 	}
 
 	doesNotUnderstand { |selector... args, kwargs|
+		var sel, forward;
 		if (know.not) {
             ^this.superPerformArgs(\doesNotUnderstand, args, kwargs)
 		};
-        this[selector] !? { |func|
-			^func.performArgs(\functionPerformList, [\value, this, args], kwargs);
+        sel = this[selector];
+		sel !? {
+			^sel.performArgs(\functionPerformList, [\value, this, args], kwargs);
         };
 
         if (selector.isSetter) {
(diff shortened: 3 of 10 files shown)


JordanHendersonMusic:
Will close #6889 once I fix the existing violations.

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -226,7 +226,7 @@ TestUGen_RTAlloc : UnitTest {
 			// memSizeSeconds * 340 produces an allocation of memSizeFloats floats
 			["GVerb", { GVerb.ar(DC.ar(1), 1, maxroomsize: this.memSizeSeconds * 340) }],
 
-			// Convolution UGens don't need a valid
```
dyfer: Unrelated whitespace change

testsuite/classlibrary/TestUGen_RTAlloc.sc
```diff
@@ -260,7 +260,7 @@ TestUGen_RTAlloc : UnitTest {
 		var blockSize = server.options.blockSize;
 		var fftSize = blockSize;
 		var out = this.awaitSynthOutput({ FFT(LocalBuf(fftSize), DC.ar(1), hop:1) });
-		this.assert(out.notNil,
```
dyfer: Unrelated whitespace change

lang/LangSource/PyrParseNode.h
```diff
@@ -77,6 +77,10 @@ enum {
 
 extern AdvancingAllocPool gParseNodePool;
 
+// This value count the un-inlined functions, these are not desirable in the class library because they are slow.
```
dyfer, line 80: This is minor, but when I read that these are not desirable "in the class library" the first thing that came to my mind was the "default class library", while they are not desirable in general in sclang. But here I guess this means all classes, built-in and extensions alike, right?

I'd consider 
```
// This value counts the un-inlined functions, these are not desirable because they are slow.
```
but maybe I think keeping "in the class library" is better, then let's keep it as is.

dyfer (commented):
Do we want to remove the whitespace change? 
Also I have one comment/suggestion. 

Otherwise this is great, thanks so much!

lang/LangSource/PyrParseNode.h
JordanHendersonMusic: Well if the user wants to do this and write their own classes then it doesn't matter. The only issue is that they are slow and in generic code you don't know if the user will be calling it in a tight loop.

testsuite/classlibrary/TestUGen_RTAlloc.sc
JordanHendersonMusic: I can't even see this as a change on my machine, git just doesn't recognise it?? (same with the other whitespace change). I wonder if it is some odd newline character?

lang/LangSource/PyrParseNode.h
dyfer: My point was to simply suggest not to ever use them. Not prevent it, but indicate that it's just not a good idea. I think that "in the class library" is just superfluous here, but if you don't see it that way I won't push for it, this is really minor/cosmetic.

testsuite/classlibrary/TestUGen_RTAlloc.sc
dyfer: Screenshot 2025-06-03 at 1 10 08 PM

I think this is just a space at the end of the line?

It's still here, AFAICT (see "Files changed" tab):

Screenshot 2025-06-03 at 1 12 11 PM

testsuite/classlibrary/TestUGen_RTAlloc.sc
dyfer: Screenshot 2025-06-03 at 1 13 46 PM

lang/LangSource/PyrParseNode.h
JordanHendersonMusic: To be honest, I doubt anyone is going to read this, it is hidden in a file one doesn't usually look through.

testsuite/classlibrary/TestUGen_RTAlloc.sc
JordanHendersonMusic: Ah, found it, I had trailing whitespace changes disabled in gitkraken...

PR #6920, opened 2025-06-02
sclang: kwarg expand syntax

JordanHendersonMusic:
## Purpose and Motivation

Closes #6882.

```supercollider
a = ()
a.meow = {|self... a, k| (self: self, a: a, k: k)}

b = [q: 10]

a.meow(**b) // ('self': ('meow': a Function), 'a': [], 'k': [q, 10])

f = {|... a, k| (a: a, k: k)}
f.(**b) // ('a': [], 'k': [q, 10])
```

Currently, I can't get `f.(*args, **kwargs)`. Pleasing the parser here is a pain, this is probably because I find bottom--up parsing very confusing.

This would be good to land in 3.14, but the deadline is close.

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/lang/LangSource/Bison/lang11d
+++ b/lang/LangSource/Bison/lang11d
@@ -1,13 +1,13 @@
 %token	NAME INTEGER SC_FLOAT ACCIDENTAL SYMBOL STRING ASCII PRIMITIVENAME CLASSNAME CURRYARG
 %token  VAR ARG CLASSVAR SC_CONST
 %token	NILOBJ TRUEOBJ FALSEOBJ
-%token	PSEUDOVAR
+%token	PSEUDOVAR KWEXPAND
 %token  ELLIPSIS DOTDOT PIE BEGINCLOSEDFUNC
 %token  BADTOKEN INTERPRET
 %token  BEGINGENERATOR LEFTARROW WHILE
 %left	':'
 %right  '='
-%left	BINOP KEYBINOP '-' '<' '>' '*' '+' '|' READWRITEVAR
+%left	BINOP KEYBINOP '-' '<' '>' '*' '+' '|' READWRITEVAR KWEXPAND
 %left	'.'
 %right  '`'
 %right  UMINUS
@@ -208,7 +208,7 @@ msgsend : name blocklist1
 				$$ = (intptr_t)newPyrCallNode((PyrSlotNode*)$2, (PyrParseNode*)$5,
 						(PyrParseNode*)$6, (PyrParseNode*)$8);
 			}
-		| name '(' arglistv1 optkeyarglist ')'
+		| name '(' argListWithPerformList optkeyarglist ')'
 			{
 				PyrSlotNode *selectornode;
 				PyrSlot slot;
@@ -226,7 +226,7 @@ msgsend : name blocklist1
 					newPyrPushLitNode((PyrSlotNode*)$1, NULL));
 				$$ = (intptr_t)newPyrCallNode(selectornode, args, (PyrParseNode*)$4, 0);
 			}
-		| '(' binop2 ')' '(' arglistv1 optkeyarglist ')'
+		| '(' binop2 ')' '(' argListWithPerformList optkeyarglist ')'
 			{
 				PyrSlotNode *selectornode;
 				PyrSlot slot;
@@ -287,7 +287,7 @@ msgsend : name blocklist1
 					(PyrParseNode*)$3);
 				$$ = (intptr_t)newPyrCallNode(selectornode, args, (PyrParseNode*)$4, (PyrParseNode*)$6);
 			}
-		| classname '(' arglistv1 optkeyarglist ')'
+		| classname '(' argListWithPerformList optkeyarglist ')'
 			{
 				PyrSlotNode *selectornode, *selectornode2;
 				PyrSlot slot, slot2;
@@ -308,6 +308,9 @@ msgsend : name blocklist1
 				args = linkNextNode(args, (PyrParseNode*)$3);
 				$$ = (intptr_t)newPyrCallNode(selectornode, args, (PyrParseNode*)$5, 0);
 			}
+
+
+
 		| expr '.' '(' ')' blocklist
 			{
 				PyrSlotNode *selectornode;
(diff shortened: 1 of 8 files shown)


JordanHendersonMusic:
> IIUC: this PR introduces to either have the ability to capture all uncaptured positional args or all uncaptured kwargs, but not both at the same time?

Yup

> From my experience of this feature in Python one rarely uses the one w/o the other b/c it allows to hijack/overload any method call and provides...

I agree, which is why I did the other syntax `cat.*meow(args, kwargs) and dog.*(a, k)` that really nicely fit into the grammar - no one seemed to like that though! 

Unfortunately, I get shift/reduce conflicts when I try `f.(*a, **k)` and I can't seem to get around them. Our grammar isn't the easiest beast to work with...

telephon:
> From my experience of this feature in Python one rarely uses the one w/o the other b/c it allows to hijack/overload any method call and provides...

I see. Well for now, as a first step, we have `performArgs` and `valueArgs` for forwarding. Maybe that is ok as a first step? 


> I agree, which is why I did the other syntax `cat.*meow(args, kwargs) and dog.*(a, k)` that really nicely fit into the grammar - no one seemed to like that though!

It is certainly something that would need to get used to, and I am not sure if the semantic of the syntax itself is consistent enough. Maybe you can explain what the "syntax means", e.g. why it is logical in the parser?

JordanHendersonMusic:
>  why it is logical in the parser?

It's invalid code to write `.*`, so this slots into that gap. If all variants of this syntax were previously invalid, there can be no conflict when it is made valid. 

Whereas because of how `*a` and all the message sends are implemented, there are conflicts with `f.a(*a, **k)` which I can't resolve  and suspect a larger refactoring would be needed to resolve them. Which runs the risk of breaking other rules and drastically changing the language by mistake.

JordanHendersonMusic:
You could look at it like that... Or you could think the of multi-character glyph '.*' as being 'access and expand'.

JordanHendersonMusic:
I hadn't thought of that! 

.... I haven't tried this, and predicting what the grammar thinks is dangerous so I'll give it a go soon and report back. However I doubt it will work.

I'm going to have another go at trying to get `f.(*a, *k)` to work, I've just had some new ideas.

JordanHendersonMusic:
...Okay it didn't work. I just cannot get the grammar to behave... Unless there is someone who happens to know bison really well (its an odd older form of parser generator) I don't think `f.(*a, **k)` will happen.

telephon:
Thanks for trying!

prko:
@JordanHendersonMusic 
Could you please provide a use case in the form of a SCHELP file? I'm finding it difficult to understand when it is needed and how to use it.
Sorry for this request.

JordanHendersonMusic:
Are you referring to this pr? Or keyword arguments in general?

telephon:
> From my experience of this feature in Python one rarely uses the one w/o the other b/c it allows to hijack/overload any method call and provides a way to increase API stability in case additional args or kwargs are introduced in future versions. This combined w/ keyword-exclusive usage improves stability across updates of code.

Do you have an example for this in python? Would be good to understand how this would work with performArgs right now and how it would benefit from a special syntax.

> btw - probably OT but another neat way on how to destructure a passed object is offered by JS, see the section _Unpacking properties from objects passed as a function parameter_ of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring - this becomes a really powerful tool and makes code much more readable and concise when passing objects.

Yes, this is something useful indeed. If the `=` were a proper message and a `var` were an object, it could be implemented in the class library.

If we added a `forward` instvar to `Maybe`, it could support destructuring:

```supercollider
LazyEnvir.push;
~arr.destructure = [~a, ~b, ~c];
~arr = [1, 2, 3]; // calls ~a = 1; ~b = 2; ~c = 3;
```

even recursively: 

```supercollider
~a.destructure = [~k, ~l];
~arr = [[10, 20], 2, 3];  // calls ~a = [10, 20], which in turn calls ~k = 10; ~l = 20;
```

But in the end, what one probably wants is a system where all values and args are passed as environments, so that methods take only one argument and return one. We have this in the Pattern system, but not complete (@rkuivila proposed an improvement long ago) …

capital-G:
> Do you have an example for this in python? Would be good to understand how this would work with performArgs right now and how it would benefit from a special syntax

I don't know how `performArgs` works but `*args, **kwargs` is essentially used to append inherited methods (see e.g. [this example](https://stackoverflow.com/questions/4269605/django-override-save-for-model-only-in-some-cases)) and  [monkey patching](https://stackoverflow.com/questions/5626193/what-is-monkey-patching) (which isn't a thing in sclang due to being static dispatch).
At the same time, sclang doesn't use excessive arguments like python (could be a chicken/egg problem though^^), see e.g. https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html#sklearn.svm.SVC 

I think it is best practice in python to capture and pass any arbitrary arguments via `*args, **kwargs` as this allows for better compatibility between library versions - see e.g. https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.html

> it could be implemented in the class library.

Oh, this is very cool! I wonder if there is a way to combine this with static efforts in order to leverage on this via LSP/autocomplete, as such features start to become really useful/powerful combined with an LSP.

telephon:
Yes, I also thought it is just there already, just in a different syntax.

```python
 def save(self, *args, **kwargs):
        if self.image:
            small=rescale_image(self.image,width=100,height=100)
            self.image_small=SimpleUploadedFile(name,small_pic)
        super(Model, self).save(*args, **kwargs)
```

```supercollider
// pseudocode
save = { |self ...args, kwargs|
	if(self.isKindOf(SimpleNumber)) {
		self.rescale; // ...
	};
        other_save.performArgs(args, kwargs)
}

```
Or in a class:
```supercollider
save { | ...args, kwargs|
	if(this.isKindOf(SimpleNumber)) {
		this.rescale; // ...
	};
	this.superPerformArgs(\save, args, kwargs)
}
```

telephon:
Well, we do have:

```
var a, b, c;
#a, b ... c = {1.rand}!10;
```

So it would be the closest to have:
```
var #a, b ... c = {1.rand}!10;
```

JordanHendersonMusic:
Okay I'm going to close these. I don't think its worth it.

PR #6924, opened 2025-06-02
ide: gui: restart sclang after config change

dyfer:
## Purpose and Motivation

This PR adds a button to allow restarting interpreter when the language config changes are applied. This was suggested in https://github.com/supercollider/supercollider/pull/6818#pullrequestreview-2827772629.

~~I wasn't sure what's the best way to connect the `SclangPage` with `ScProcess` - I ended up going through the settings manager. But I don't have a good birds eye view of scide, so I'm not sure if this was the best way to go...~~

Screenshot 2025-06-02 at 2 54 06 PM

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/editors/sc-ide/widgets/settings/sclang_page.cpp
+++ b/editors/sc-ide/widgets/settings/sclang_page.cpp
@@ -29,6 +29,7 @@
 #include "ui_settings_sclang.h"
 #include "../../core/settings/manager.hpp"
 #include "../../core/util/standard_dirs.hpp"
+#include "../../core/main.hpp"
 
 #include <yaml-cpp/yaml.h>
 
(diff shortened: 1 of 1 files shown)


editors/sc-ide/core/sc_process.cpp
```diff
@@ -53,6 +53,8 @@ ScProcess::ScProcess(Settings::Manager* settings, QObject* parent):
     connect(mIpcServer, SIGNAL(newConnection()), this, SLOT(onNewIpcConnection()));
     connect(this, SIGNAL(stateChanged(QProcess::ProcessState)), this,
             SLOT(onProcessStateChanged(QProcess::ProcessState)));
+
+    connect(settings, &Settings::Manager::requestRestartLanguage, this, &ScProcess::restartLanguage);
```
capital-G: Wouldn't it be "cleaner" to have the connection within the dialog?

The sc_process shouldn't need to be aware of any settings dialog, but the settings dialog of the lang could be aware of an process b/c it also displays the respective warning.

capital-G (commented):
Thanks, this is a really nice addition which helps to improve the UX <3

I only have a small remark/question, but in general the implementation looks solid to me.

editors/sc-ide/core/sc_process.cpp
dyfer: Yeah, but how to expose the `ScProcess` in the `SclangPage` part of the code? I couldn't figure it out, as I was trying to explain in the opening comment.

capital-G: What do you think about

```cpp
    if (reply == QMessageBox::Yes) {
        if (auto process = Main::instance()->scProcess()) {
            process->restartLanguage();
        } else {
            qWarning() << "SclangPage::dialogConfigFileUpdated(): Main->scProcess pointer is null!";
        }
    }
```

dyfer: Ah, that's much better! Changed, thanks.

Issue #6349, opened 2024-06-13
Add `postEnvironment` command


capital-G:
## Motivation

Currently each bug report should state

https://github.com/supercollider/supercollider/blob/8378bb4aa854b949001ef1f8c407426b47fb7709/.github/ISSUE_TEMPLATE/bug_report.md?plain=1#L13-L17

Maybe we can add a command such as

```
this.postEnvironment;
```

which could yield as much context as possible for SC class library bugs, which then could be simply pasted here.

This issue could be to decide if we want to have such a command and what this command should include exactly.
Adding information such as active extension folders etc would be helpful I think.

telephon:
It could be a class method of `Platform`.

capital-G:
Implemented via https://github.com/supercollider/supercollider/pull/6928

PR #6928, opened 2025-06-04
Add Main.postBugReportInfo

capital-G:
## Purpose and Motivation

Builds upon #6888 but adds

* reverts whitespace formatting issues
* makes printing of includes optional (enabled by default)
* makes printing of audio devices optional (disabled by default)
* switch to markdown formatting which mimics https://github.com/supercollider/supercollider/blob/742da519d27a9fb73b7a3e58d8d1e50b66b81049/.github/ISSUE_TEMPLATE/bug_report.md?plain=1#L13-L17
* fix some formatting issues on OS version
* add architecture to macOS version

The output of `Main.postBugReportInfo` becomes

```
Copy the content between the ### lines
########################################
* SuperCollider version: 3.14.0-dev (Built from branch 'topic/print-bug-report-info-2' [14647e70f1])
* Operating system: osx 15.4.1 (x86_64)
* Other details
	* Qt version: 6.9.0
	* argv: 
	* server: scsynth
	* includes: SCClassLibrary, SafetyNet
########################################
```

and, when pasted in markdown, becomes

---

* SuperCollider version: 3.14.0-dev (Built from branch 'topic/print-bug-report-info-2' [14647e70f1])
* Operating system: osx 15.4.1 (x86_64)
* Other details
	* Qt version: 6.9.0
	* argv: 
	* server: scsynth
	* includes: SCClassLibrary, SafetyNet

---

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


SCClassLibrary/DefaultLibrary/Main.sc
```diff
@@ -206,6 +207,55 @@ Main : Process {
 		}
 	}
 
+	*systemInformation {
+		^(
+			version: Main.version,
+			buildString: Main.scBuildString,
+			qtVersion:  \QtGUI.asClass !? _.version,
```
JordanHendersonMusic: This would count as an uninlined function #6916... although I do like this monadic(?) style

dyfer:
Thanks! This is great. Some suggestions:
- do we want this in `Main` as opposed to `Platform`? (https://github.com/supercollider/supercollider/issues/6349#issuecomment-2166119559)
- I'd keep architecture for every platform; Linux Arm64 and even Windows Arm64 is possible, when built locally
- Maybe we don't need a branch name? This would only show up for local builds anyway. I was thinking a hash alone would be both succinct and informative enough: `* SuperCollider version: 3.14.0-dev [14647e70f1]`
- What do you think about adding a field for audio device to be filled out?
```
* Other details
	* active audio device: <paste here>
```
or something like this?

dyfer (approved):
Good addition!

PR #6929, opened 2025-06-04
Update if.schelp

prko:
## Purpose and Motivation
There is missing information in `if.schelp`:
- information on various writing styles. 
- how `if` method controls sound with Ugen.

This PR aims to add:
- a link to explanations of various writing styles in link::Reference/Control-Structures::
- a basic explanation of how the `if` method controls sound with `UGen`

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/if.schelp
+++ b/HelpSource/Reference/if.schelp
@@ -5,7 +5,10 @@ summary:: conditional execution
 
 method:: if
 
-section::example
+section::Basic examples
+
+This section focuses solely on the basic writing style. For various writing styles of code::if::, refer to link::Reference/Control-Structures::.
+
 code::
 if ( [false, true].choose,				// Boolean expression (chooses one at random)
 	{ "expression was true".postln },	// true function
@@ -19,21 +22,34 @@ z.postln;
 )
 ::
 
-UGens can also use if
-
-the condition ugen is 0 / 1
+Any subclass of link::Classes/UGen:: can also utilize code::if::. An output value of code::0:: is interpreted as code::false::, while code::1:: is code::true::; values between code::0:: and code::1:: are interpolated accordingly. At code::0::, code::if:: outputs only the sound from the code::false:: UGen. Between code::0:: and code::0.5::, the sound from the code::true:: UGen gradually mixes in. At code::0.5::, code::if:: outputs the sound from both UGens in equal proportion. If both UGens are identical, their output levels match. Between code::0.5:: and code::1::, the sound from the code::false:: UGen diminishes, and at code::1::, code::if:: outputs only the sound from the code::true:: UGen.
 
 code::
+// Experience how if controls sound using MouseX:
+(
+{
+	if(
+		MouseX.kr(1, 0).poll, // left: true; right: false
+		SinOsc.ar * [0.1, 0], // or: SinOsc.ar * 0.1,
+		SinOsc.ar * [0, 0.1]  // or: SinOsc.ar * 0.1
+	)
+}.play
+)
+
+// Automatically mix two sound sources using the random generator LFNoise1:
 (
 	{
-		if( LFNoise1.kr(1.0,0.5,0.5) , SinOsc.ar, Saw.ar )
+		if(
+		LFNoise1.kr(1.0).range(0, 1).poll,
+		SinOsc.ar,
+		Saw.ar ) * 0.1
 	}.play
 )
 ::
 
-section:: optimization
+section:: Optimization
 
-the functions will be inlined, which plucks the code from the functions and uses a more efficient jump statement.
+The functions will be inlined, which plucks the code from the functions and uses a more efficient jump statement.
 
 code::
 {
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
I don't know if we want to advertise if with ugens, it's a bit of a hack...

prko:
>I don't know if we want to advertise if with ugens, it's a bit of a hack...

If I recall correctly, the use of if within SynthDefs and if with Ugens has been a recurring topic in both the current forum and the previous mailing list. This may also explain why the following sections are included in the [User FAQ](https://doc.sccode.org/Guides/UserFAQ.html#SynthDef%20Issues):
- "If" statements inside a SynthDef
- Why is x ＞ 0 "non-Boolean" in the "test"?
- If you can't branch, what good is a comparison in the server?
- Logical operators: And, Or, Not, Xor

Furthermore, for users who are unaware that these explanations are available at that document, finding that section can be quite challenging.

**Documentation should not only help experienced SuperCollider users recall forgotten details, but also be organized so that newcomers can quickly find solutions to their problems.**

prko:
Is it better to move the following sections from the [User FAQ](https://doc.sccode.org/Guides/UserFAQ.html#SynthDef%20Issues) 
- "If" statements inside a SynthDef
- Why is x ＞ 0 "non-Boolean" in the "test"?
- If you can't branch, what good is a comparison in the server?
- Logical operators: And, Or, Not, Xor

in place of the following text?
```
Further explanations can be found in the following parts of the link::Guides/UserFAQ:::
list::
## "If" statements inside a SynthDef
## Why is x ＞ 0 "non-Boolean" in the "test"?
## If you can't branch, what good is a comparison in the server?
## Logical operators: And, Or, Not, Xor
::
```

jamshark70:
> What I think would be excellent to add to this PR is to explain why if is kind of a hack and then show a different approach such as Select.*r that better reflects how to think about streams on the server.

I wrote that bit of the user FAQ ages ago -- it's migrated from a wiki page into help.

It does explain why one shouldn't expect branching to work in the server, and gives a template for the use of `Select`.

Perhaps you might wish to emphasize "if is a hack" a bit more? But the content you describe is already there.

prko:
>Perhaps you might wish to emphasize "if is a hack" a bit more? But the content you describe is already there.

This is my thought, so I linked that part in `if.schelp`.

HelpSource/Reference/if.schelp
```diff
@@ -5,7 +5,10 @@ summary:: conditional execution
 
 method:: if
 
-section::example
+section::Basic examples
+
+This section focuses solely on the basic writing style. For various writing styles of code::if::, refer to link::Reference/Control-Structures::.
```
HotwheelsSisyphus: I think this is a good thing to point out, and also good to link to Control Structures, but I'm not sure if this should be referred to as "_the_ basic writing style" (implying that other syntax variants are less basic). Instead I'd use the terms from the Syntax Shortcut helpfiles: The syntax used here is "function call syntax" (without trailing argument block). 

Suggestion:

This section uses the emphasis::function call syntax:: for code::if::. For various alternate syntaxes (e.g., the frequently used emphasis::function call with trailing arguments:: notation), see link::Reference/Control-Structures:: or link::Reference/Syntax-Shortcuts::.

prko: Thanks! It's been done in https://github.com/supercollider/supercollider/pull/6929/commits/bb38182dfa5d8aa6454f2a576a4368e79d34bc19.

Note:
I added sections because the linked documentation is long. Is it ok for you?

prko:
I think:
The text in this PR should be moved to PR #6967 once it has been merged.
Alternatively, #6967 could replace this PR.

telephon (approved):

HotwheelsSisyphus:
I just looked over the optimization section in here and I think I now understand it enough to say that it doesn't significantly differ from the optimization section in Control-Structures.schelp. 

But even if if.schelp doesn't end up getting deleted altogether, at least lines 93 through the end should be, because they are an exact duplicate of lines 36-59 that seems to have slipped in there by accident.

tedmoore:
I'm OK with deleting if.schelp and moving that information to a doc that is shared by other control structures. 

I think it may *help* learners by realizing that "if" has friends like "while", etc., for use in similar but different situations. Beginner coders often learn "if" early on, but "switch", "case", even "while", might fall through the cracks for a bit. Having it all in the same document makes these other control structures "discoverable" and therefore make the language more "learnable". It can be helpful when the taxonomy of the documentation reflects the taxonomy of the language.

prko:
noticed that if.schelp was removed in #6967.
If the changes proposed here are already covered by the corresponding content in #6967, I believe this PR can be closed.
Please feel free to close it if appropriate!

HotwheelsSisyphus:
Since I finally merged 6967, I'm now closing this. Thanks for all the feedback..

PR #6932, opened 2025-06-04
classlib: check for ScrollView's parent

### LINKED ISSUE #3435: ScrollView crashes interpreter on improper canvas setting
I'm not sure if this is a legitimate issue, stumbled across it by accident.
Proper use for ScrollView asks to provide a View as its "canvas":
```
~scroll = ScrollView();
~canvas = View();
~scroll.canvas = ~canvas;
```
However, if that View has (improperly) defined the ScrollView as its parent, setting the canvas crashes the interpreter:
```
~scroll = ScrollView();
~canvasView = View(~scroll);
~scroll.canvas = ~canvasView; //crash
```
Ideally we'd get an error message, but then again this happens when one is doing the _wrong_ thing. Anyone has thoughts on that?

dyfer:
## Purpose and Motivation

This PR attempts to fix the crash with the following code:

```
~scroll = ScrollView();
~canvasView = View(~scroll);
~scroll.canvas = ~canvasView; //crash before this PR
```

This is just a manual check in the class library; I'm not sure if this would be better fixed in the Qt primitive?

I decided to throw an error. Or should we fail silently?

Fixes #3435

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/GUI/Base/BasicViews.sc
+++ b/SCClassLibrary/Common/GUI/Base/BasicViews.sc
@@ -172,8 +172,12 @@ ScrollView : AbstractScroll {
 	}
 
 	canvas_ { arg view;
-		canvas = view;
-		this.invokeMethod( \setWidget, view, true );
+		if(view.notNil and: {view.parent == this}) {
+			Error("Can't set canvas to a view, whose parent is this ScrollView.").throw; // when that's the case, the interpreter crashes
+		} {
+			canvas = view;
+			this.invokeMethod( \setWidget, view, true );
+		}
 	}
 
 	action_ { arg handler;


mtmccrea:
An error seems appropriate to signal to a user their mistake or mistaken intention.

If you're unsure about erroring out in case halting disrupts some continuous process (unlikely?), a warning would serve the same purpose.

dyfer:
> If you're unsure about erroring out in case halting disrupts some continuous process (unlikely?), a warning would serve the same purpose.

Yeah, it does and it doesn't. If the execution doesn't stop, then the warning will likely get unnoticed. 

This feels like a clumsy workaround, instead of a proper fix, but I'm not sure what the best solution is? I feel like passing a wrong parent should cause a crash on the cpp side, but rather throw an error...

mtmccrea:
As an intermediate solution to prevent the crash, I think an Error is ok!

dyfer:
Not sure who to tag to push this forward... @telephon do you have any guidance here? Or is this okay as a crash-preventing solution?

JordanHendersonMusic (approved):
Seems like a simple straightforward way to prevent a crash. Might not be the perfect solution but it is better than what currently happens.

dyfer:
Thanks! I'll merge then.

PR #6942, opened 2025-06-06
GHA: use all cores on linux build

capital-G:
## Purpose and Motivation

This should speed up our linux pipeline by using every available core.
Currently these are 4 (see https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) but this may change - with this we should scale dynamically according to our environment.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -157,7 +157,7 @@ jobs:
       - name: build
         run: |
           cd $BUILD_PATH
-          make install -j2
+          make install -j$(nproc)
 
 
       - name: create archive


JordanHendersonMusic (approved):

capital-G:
IIRC CMake on macOS automatically scales according to the available cores - it does so at least on my intel mac.

BTW - is there a reason to use the `make` command instead of wrapping everything via `cmake` under linux? I think I also use `cmake` exclusively for building?

JordanHendersonMusic:
I've only ever use `make`... after using a modern build system (zig's) I try to stay away from cmake because it erodes my soul every time I think of it... but if you wanted to change it I wouldn't mind.

dyfer (approved):
I haven't checked what nproc returns on GHA... but I assume this is good :)

dyfer:
OK to merge this @capital-G @JordanHendersonMusic ? I think it's good as is.

PR #6943, opened 2025-06-06
Build help in CI

### LINKED ISSUE #6940: Render docs as a part of the CI
## Motivation

It's easy to leave out a `::`, making the docs invalid.

Add a unit test, should be easy.

capital-G:
## Purpose and Motivation

Closes #6940 

I first wanted to build on linux to make it more easily testable for everyone locally, but due to sclang requiring a window server to run I switched to macOS where the window server is already available.

Still unsure on how to best report any issues while building/parsing the docs here, hopefully @JordanHendersonMusic has a good idea on how to do this properly :)

Question for @dyfer: Is it bad for the cache if one job is additionally installing `coreutils` in order to obtain timeout, or should therefore all builds install `coreutils`?

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/tools/build_help.scd
+++ b/tools/build_help.scd
@@ -1,0 +1,7 @@
+SCDoc.helpSourceDir_("SC_HELP_SOURCE_DIR".getenv);
+
+SCDoc.helpTargetDir_("SC_HELP_TARGET_DIR".getenv.mkdir);
+
+SCDoc.renderAll(includeExtensions: false);
+
+0.exit;
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -172,6 +172,22 @@ jobs:
           name: ${{ env.ARTIFACT_FILE }}
           path: ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE }}.zip
 
+      - name: Build help
+        if: matrix.build_help
+        continue-on-error: false
+        run: timeout 5m cmake --build $BUILD_PATH --config Release --target Help
+    
+      - name: create help archive
+        if: matrix.build_help
+        run: cd $BUILD_PATH/Help && zip --symlinks -r Help.zip .
+
+      - name: upload help artifacts
+        uses: actions/upload-artifact@v4
+        if: matrix.build_help
+        with:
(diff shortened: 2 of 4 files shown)


JordanHendersonMusic:
Working on this right now. I'll send you a PR on your repo.

JordanHendersonMusic:
@mtmccrea suggesting also checking links. We can add additional passes and walk the 'ast' to check these, but maybe that is something for later.

tools/build_help.scd
```diff
@@ -0,0 +1,7 @@
+SCDoc.helpSourceDir_("SC_HELP_SOURCE_DIR".getenv);
```
dyfer, line 1: You could also use argv instead of env variables, but I'm not sure which one's better? 
See [test runner](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/sclang_test_runner.scd#L6).

capital-G: I know but I personally prefer options args instead of positional args as they make their intent more explicit - otherwise I always have to double check that I don't mix up pos0 and pos1 - and by prepending env variables to a command I still can achieve something like option args.

capital-G:
This depends on #6950 in order to exit properly (although `(SCDoc.parseErrors.size>0).if(1, 0).exit` could probably also work^^).

But lets wait for the more proper fix.

dyfer:
@capital-G with #6953 merged this can now proceed.

> due to sclang requiring a window server to run I switched to macOS where the window server is already available

With macOS runners being generally more "expensive" and already taking more time due to installing dependencies, maybe you can just copy the [virtual X setup from the test runner](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/test_linux.yml#L53-L55) (run conditionally in a separate step when building help) and do this on Linux?

dyfer:
@capital-G are you interested in moving this forward? 
I'd also recommend switching this to Linux, as I noted previously. But if you want to stick with macOS, that's fine too.

dyfer:
@capital-G what's the status of this PR?

PR #6944, opened 2025-06-06
Center Stepwise Plot Modes and Add Additional Values

prko:
## Purpose and Motivation
closes: #5296, #6479

With this PR, the following code block
```
(
a = [5, 0, 1, 2, 3, 4];
b = a.reverse;
[a, b, a, b, a, b, a, b, a, b, a, b, a, b].plot(bounds: Rect(0, 0, 400, Window.availableBounds.height))
.plotMode_([\points, \dots, \lines, \plines, \dlines, \stems, \pstems, \dstems, \filled, \pfilled, \dfilled, \levels, \steps, \bars])
)
```
returns the following graphic:

Screenshot 2025-06-06 at 20 46 41

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -496,20 +496,34 @@ Plot {
 	}
 
 	levels { |x, y|
+		var xOffset = x[1] - x[0] / 2;
 		Pen.smoothing_(false);
 		y.size.do { |i|
-			Pen.moveTo(x[i] @ y[i]);
-			Pen.lineTo(x[i + 1] ?? { plotBounds.right } @ y[i]);
+			var lastX = { if (y.size - 1 == i) {
+						plotBounds.right + xOffset
+					} {
+						plotBounds.right
+					}
+				};
+			Pen.moveTo(x[i] - xOffset @ y[i]);
+			Pen.lineTo(x[i + 1] ?? lastX - xOffset @ y[i]);
 		};
 		Pen.stroke
 	}
 
 	steps { |x, y|
+		var xOffset = x[1] - x[0] / 2;
 		Pen.smoothing_(false);
 		Pen.moveTo(x.first @ y.first);
 		y.size.do { |i|
-			Pen.lineTo(x[i] @ y[i]);
-			Pen.lineTo(x[i + 1] ?? { plotBounds.right } @ y[i]);
+			var lastX = { if (y.size - 1 == i) {
+						plotBounds.right + xOffset
+					} {
+						plotBounds.right
+					}
+				};
+			Pen.lineTo(x[i] - xOffset @ y[i]);
+			Pen.lineTo(x[i + 1] ?? lastX - xOffset @ y[i]);
 		};
 		Pen.stroke;
 	}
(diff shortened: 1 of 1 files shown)


prko:
@tremblap, @mtmccrea, @elgiano, @dyfer, @telephon 

Hello, could you look at this PR?

What I want to have is like this:
UntitledSSSS

However, the result of this PR is also usable, I believe.....

elgiano:
Hi @prko, thanks for this!
So far I only tested it, without reading the code (sorry). I have a few notes:

1. `(0,0.1..1).plot.plotMode_(\bars)` doesn't display correct x-axis labels. It seems to happen whenever the array items are not integers. Luckily it doesn't affect Function:-plot, because that function is setting domainSpecs explicitely.
plot

2. this adds padding by default for all plot modes. Are we sure we want this?
3. it doesn't work properly in edit mode (edit mode is still using the "old" x indices)
4. the width of the last bar doesn't adapt to domainSpecs other than linear.

You can test all of these with:

```supercollider
(
x = (0,0.1..1).plot
.plotMode_(\bars)
.editMode_(true)
)

x.domainSpecs = [amp.asSpec]
```

Personally, it seems a very complicated area of sclang code. I'm wondering if we should change implementation to use something like [Qt Graphs](https://doc.qt.io/qt-6/qtgraphs-index.html) instead

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -1328,6 +1349,15 @@ Plotter {
...
 		);
 
+		if (flatArraySorted.every { |item| item.isInteger }){
+			plotter.domain_((flatArraySorted.first..flatArraySorted.last))
+		};
+		// This is to prevent the following error from occurring when the domain value is a float:
+		// ERROR: [Plotter:-domain_] Domain array size [87] does not match data array size [9]
+		//[87] and [9] vary according to array items.
+
+		plotter.domainSpecs_([flatArraySorted.first - 0.5, flatArraySorted.last + 0.5].asSpec);
```
mtmccrea, line 1359: Thanks for getting this started @prko.

`domainSpecs` is a property that the user can later set, so the drawing of levels may be disrupted unless there is an internal bookkeeping of domainSpecs as set by the user vs. used for display (otherwise an internally-updated `domainSpec` will differ from what a user sets, which would be confusing).

Also, we'll want to make sure the various features of plotter are still supported with this change, e.g. `editMode`, `superpose`, and setting domain values explicitly.

One outstanding issue is that `step` plot mode should probably be handled differently than `levels` and `bars`. The latter are centered on the data points, while `step` is more like sample-and-hold, which holds the value from the current until the next (though perhaps this is debatable). This may require special handling of the last value.

Finally, I think the handling of the visibility of the max/min values should be a separate PR, because there could be differing opinions on it.

I started testing this and a number of these issues came up, so I begin drafting an alternative approach, which isn't complete yet, but perhaps you might be interested:
https://github.com/supercollider/supercollider/compare/develop...mtmccrea:supercollider:topic/plotter-center-steplike-data

I'm not sure if I'll have time to follow up with this in the near term, but I'll try to circle back. If you have any comments on the approach I linked to, I'm happy to discuss.

prko: Thank you for this. I find your draft to be far superior, and I appreciate that this PR accomplishes much more. I believe it would be best for me to await your work. I am truly grateful that you have taken the time to share your valuable contributions.

Issue #6689, opened 2025-03-29
Create a LinkClock UGen


capital-G:
## Motivation

Currently it is no native way of reliably sync sclang and scsynth/supernova with a common clock or phase. sclang [already implements a LinkClock](https://dev.docs.supercollider.online/Classes/LinkClock.html), but not the server.

Having a way to synchronize both would e.g. allow to make scsynth responsible of the phase and sharing it with sclang which allows e.g. phase modulations aka tempo modulations. Maybe something along `{LinkClockSetter.ar(phase: Integrator.ar(SinOsc.ar(0.2).range(0.2, 0.3)`) or vice versa by setting the tempo from the language and the server sequencing clock follows.

In general this would allow to combine the server sequencing approach with sequencing using the sclang pattern system.

This would not allow for sample accurate sequencing but this is not the intention here.

## Plan for Implementation

I could look into this, but I currently have already much stuff to work on currently - but I was having the thought that something like this would be nice for everyone.
I thought that writing down the idea would be a good start?

### Scheduling within sclang?

_This could/should probably be another PR but just want to mention this here?_

I am not sure how the sclang pattern system actually schedules events - in order to allow for fluctuating tempos/modulated phases it should never use `.wait` as such can not be cancelled anymore IIUC?
Maybe `CondVar` could help here, but afaiu it does not replicate the `async/await/promise` capabilities known from other languages. [tokios select](https://tokio.rs/tokio/tutorial/select) offers IMO a nice way to await an event in a dynamic manner - just as an inspiration.
Ideally it would be great to also allow to move backwards via the phase, but this is probably not possible with the current pattern system?

One additional problem is that the language needs to run ahead of the clock. Maybe there is some clever way to handle this, e.g. spawn a master clock on the server but only use its values with a delay of `Server.latency` so language and server are somehow in sync again?

jamshark70:
> I am not sure how the sclang pattern system actually schedules events - in order to allow for fluctuating tempos/modulated phases it should never use .wait as such can not be cancelled anymore IIUC?

(Yes, this probably should go somewhere else.)

[...]

PR #6947, opened 2025-06-07
Add Ableton Link UGen

capital-G:
## Purpose and Motivation

Closes https://github.com/supercollider/supercollider/issues/6689 which adds a way to synchronize language and server (and other software) through Ableton Link.

This is only a first draft until a good API has been established - please provide some feedback :)

## Demo

```supercollider
s.reboot;


// create a tempo clock through a ugen
(
Ndef(\clock, {
	var phase, tempo;
	var modTempo = LFDNoise3.kr(1.0).exprange(100, 380).poll(1.0, \tempo);
	#phase, tempo = LinkUGen.kr(
		setTempo: DC.kr(1.0),
		newTempo: modTempo,
		setPlaying: DC.kr(0.0),
		playingValue: DC.kr(1.0),
		quantum: 1.0,
	);
	// some annoying metronome
	SinOscFB.ar(1440.0, 1.0) * (phase < 0.1) * 0.2.dup;
}).play;
)


// this gets synced after a short time :) - works also accross network!
(
Ndef(\otherClock, {
	var phase, tempo;
	#phase, tempo = LinkUGen.kr(
		setTempo: DC.kr(0.0),
		newTempo: DC.kr(0.0),
		setPlaying: DC.kr(0.0),
		playingValue: DC.kr(1.0),
		quantum: 1.0,
	);
	// some other annoying metronome
	SinOscFB.ar(4440.0, 1.0) * (phase < 0.1) * 0.2.dup;
}).play;
)

// create a language clock
l = LinkClock(1.0).latency_(Server.default.latency);

// pattern is now in sync w/ UGen clock :)
(
Pdef(\p, Pbind(
	\dur, 0.5,
	\amp, 0.5,
	\scale, Scale.minor,
	\degree, Pwrap(Pseries(0.0, 1.1), lo: 0.0, hi: 8.9),
)).clock_(l).play;
)


// delete master clock - tempo is now static
Ndef(\clock).clear;

Pdef(\p).clear;

Ndef.clear;
```

Things to work out:

### Interface

* How should a "jump to beat" be established?
* Maybe do not expose tempo at all and instead rely solely on phase? Tempo seems to be somehow more for UI feedback than for usage according to docs? 

### Implementation

* Improve performance (how though?)
* On heavy tempo modulation, some events of a pattern will not be released - maybe due to UDP overflow?
* Improve init - use a pointer for ableton link which then can be initialized in a deferred thread? Yet this is something which won't be spawned 100s of times, probably one wants to keep this like a singleton
* How to obtain output latency of the current hardware? (Is this output latency also respected within sclang?)
* On high tempi it seems that `phase < 0.1` is not working correctly anymore, even when using audio rate. Maybe I calculate some offsets wrong?

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Audio/LinkUGen.sc
+++ b/SCClassLibrary/Common/Audio/LinkUGen.sc
@@ -1,0 +1,16 @@
+LinkUGen : MultiOutUGen {
+	*ar { |setTempo, newTempo, setPlaying, playingValue, quantum=1.0, hardwareBlockSize=512|
+	    var delay = hardwareBlockSize / SampleRate.ir;
+		^this.multiNew('audio', setTempo, newTempo, setPlaying, playingValue, quantum, delay);
+	}
+
+    *kr { |setTempo, newTempo, setPlaying, playingValue, quantum=1.0, hardwareBlockSize=512|
+        var delay = hardwareBlockSize / SampleRate.ir;
+        ^this.multiNew('control', setTempo, newTempo, setPlaying, playingValue, quantum, delay);
+    }
+
+	init {arg ... theInputs;
+		inputs = theInputs;
+		^this.initOutputs(2, rate);
+	}
+}
--- a/server/plugins/LinkUGen.cpp
+++ b/server/plugins/LinkUGen.cpp
@@ -1,0 +1,120 @@
+#include <ableton/Link.hpp>
+
+#include "SC_InterfaceTable.h"
+#include "SC_Unit.h"
+#include "SC_PlugIn.hpp"
+
+static InterfaceTable* ft;
+
+class LinkUGen : public SCUnit {
+public:
+    LinkUGen(): mLink(120.0) {
+        // this is not rt safe :p
+        mLink.enable(true);
+
+        mMSecPerTick = std::chrono::microseconds(static_cast<int>(10e6 / sampleRate()));
+
+        // how do we obtain the passed 'control' or 'audio' parameter in the plugin api?
+        if (isControlRateIn(0)) {
+            mCalcFunc = make_calc_function<LinkUGen, &LinkUGen::next_k>();
+            // prime the pipe
+            next_k(1);
+        } else {
+            mCalcFunc = make_calc_function<LinkUGen, &LinkUGen::next>();
+            next(1);
+        }
+    }
+    ~LinkUGen() {
+        mLink.enable(false);
+        // do we need to destroy mLink here ;?
+    }
+
+private:
+    // how to allocate this not during construction?
+    ableton::Link mLink;
+    std::chrono::microseconds mMSecPerTick = std::chrono::microseconds::zero();
+
+    void next_k(int numSamples) {
+        auto sessionState = mLink.captureAudioSessionState();
+
+        const float setTempo = in0(0);
+        const float newTempo = in0(1);
+        const float setIsPlaying = in0(2);
+        const float isPlayingVal = in0(3);
+        const float quantum = in0(4);
+        const float delay = in0(5);
+
+        bool tempoSet = false;
+        bool playSet = false;
+
+        const std::chrono::microseconds hardwareOffset = std::chrono::microseconds(static_cast<int>(10e6 * delay));
+
+        // should we calculate from the beginning or from middle of the block (size)?
+        const auto time = mLink.clock().micros() + hardwareOffset;
+
+        if (setTempo > 0.0f) {
+            tempoSet = true;
+            sessionState.setTempo(newTempo, time);
+        }
+        if (setIsPlaying > 0.0f) {
+            playSet = true;
+            sessionState.setIsPlaying(isPlayingVal, time);
+        }
+
+        if (playSet || tempoSet) {
+            mLink.commitAudioSessionState(sessionState);
+        }
+
+        // phase
+        out0(0) = sessionState.phaseAtTime(time, quantum);
+        // tempo
+        out0(1) = sessionState.tempo();
+    }
+
+    void next(int numSamples) {
+        auto sessionState = mLink.captureAudioSessionState();
+
+        float* phase = out(0);
+        float* tempo = out(1);
+
+        const float* setTempo = in(0);
+        const float* newTempo = in(1);
+        const float* setIsPlaying = in(2);
+        const float* isPlayingVal = in(3);
+        const float* quantum = in(4);
+        const float* delay = in(5);
+
+        bool tempoSet = false;
+        bool playSet = false;
+
+        const std::chrono::microseconds hardwareOffset = std::chrono::microseconds(static_cast<int>(10e6 * delay[0]));
+
+        for (int i = 0; i < numSamples; i++) {
+            // i * mSecPerTick could be pre-calculated once - this seems to be computational intensive for some reason?
+            const auto time = mLink.clock().micros() + (i * mMSecPerTick) + hardwareOffset;
+
+            if (setTempo[i] > 0.0f && !tempoSet) {
+                // only set once in a block? sample accurate though
+                tempoSet = true;
+                sessionState.setTempo(newTempo[i], time);
+            }
+            if (setIsPlaying[i] > 0.0f && !playSet) {
+                // only set once in a block? sample accurate though
+                playSet = true;
+                sessionState.setIsPlaying(isPlayingVal[i], time);
+            }
+
+            phase[i] = sessionState.phaseAtTime(time, quantum[i]);
+            tempo[i] = sessionState.tempo();
+        }
+
+        if (playSet || tempoSet) {
+            mLink.commitAudioSessionState(sessionState);
+        }
+    }
+};
+
+PluginLoad(LinkUGen) {
+    ft = inTable;
+    registerUnit<LinkUGen>(ft, "LinkUGen", false);
+}
--- a/server/plugins/CMakeLists.txt
+++ b/server/plugins/CMakeLists.txt
@@ -142,6 +142,15 @@ add_library(ML_UGens MODULE
 	${scplugin_common_sources}
 )
 
+if(SC_ABLETON_LINK)
+	add_library(Link_UGen MODULE
+		LinkUGen.cpp
+	)
+	include(../../external_libraries/link/AbletonLinkConfig.cmake)
+	target_link_libraries(Link_UGen PRIVATE Ableton::Link)
+	list(APPEND plugins Link_UGen)
+endif()
+
 if(NOT NO_LIBSNDFILE)
 	set(diskio_sources DiskIO_UGens.cpp)
 


telephon:
How nice.

Random notes: if you call `IN0[i]` for the parameters like `setTempo` there should be no need to pass in a `DC.kr(0)` – a `0` should always be equivalent. 

Maybe you want to use `-1` or `NaN` as a signal for "nothing"? One may want to pause the clock by setting the tempo to zero.

If two output values are awkward, it is also possible to have two UGens: `LinkPhase` and `LinkTempo`, which return or set the respective value as you need it.

jamshark70:
> One may want to pause the clock by setting the tempo to zero.

I don't think there's a good use case for pausing a LinkClock. What should other peers do in that case?

jamshark70:
> On heavy tempo modulation, some events of a pattern will not be released - maybe due to UDP overflow?

Heavy tempo modulation can't be made safe with LinkClock, because of messaging latency.

capital-G:
I re-designed the API and I think it is quite neat now? The fundamental change: The server has only one singleton link clock, which is somehow fitting for Link.

Maybe @telephon and @jamshark70 can provide some feedback regarding the sclang implementation?
Some questions:
* [ ] How should I handle the `LinkUGen.server` init? If I restart the server, the state of my `LinkUGen` class is wrong :/ - though it probably wouldn't hurt to "start" the clock two-times on the server
* [ ] I am also not sure if we should use UGens for e.g. enabling a clock or switch to Plugin cmds? IMO UGens are nicer b/c they allow to do stuff on the server, and plugin commands really shine when strings needs to be transferred to the server.
* [ ] How to obtain hardware buffer size? :?

server/plugins/LinkUGen.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+private:
+    bool mWarned = false;
+
+    void next_k(int numSamples) {
+        float* out = mOutBuf[0];
+        if (LINK_CLOCK->isEnabled()) {
+            if (in0(0) > 0.0f) {
+                auto state = LINK_CLOCK->captureAudioSessionState();
+                if (in0(3) >= 0.0f) {
```
capital-G:
```suggestion
                if (in0(3) > 0.0f) {
```

jamshark70:
>  How should I handle the LinkUGen.server init? If I restart the server, the state of my LinkUGen class is wrong :/ - though it probably wouldn't hurt to "start" the clock two times on the server

To restart the server, first the server must quit. So the ServerQuit class can detect that this has happened, and update the enabled flag. I think I would not automatically re-enable the server-side Link, since there are probably other things you would have to reinitialize in a newly booted server.

LinkUGen currently assumes it will only ever be used on one server, doesn't it? I guess if it were me, I would have a dictionary of associations from server to status. Some people do use multiple servers for parallelism, so it might be better to have something like this:

```
LinkUGen {
	classvar <>servers;
	
	*initClass {
		servers = IdentityDictionary.new;
		Class.initClassTree(Server);
		Server.all.do { |server|
			servers[server] = (enabled: false);
		};
		SimpleController(Server)
		.put(\serverAdded, { |obj, what, server|
			servers[server] = (enabled: false);
		});
	}
	
	*enable { |server(Server.default)|
		if(server.hasBooted.not, {
			"Server % is not running - can not enable LinkUGen".format(server).warn;
			^this;
		});
		// ... where the 'enabled' method looks up the flag in the dictionary
		if(this.enabled(server), {
			"LinkUGen is already enabled".warn;
			^this;
		});
		{FreeSelf.kr(LinkEnabler.kr())}.play(server);
		servers[server][\enabled] = true;
	}
	... etc
}
```

You can't be certain that the server which is the default after `Class.initClassTree(Server)` will still be the default by the time the user is ready to enable Link.

> I am also not sure if we should use UGens for e.g. enabling a clock or switch to Plugin cmds?

Hm.... I don't have an opinion about that right now.

> How to obtain hardware buffer size?

I don't think there is any way (but we should add one -- it's an unfortunate lapse).

capital-G:
Update 2026-01-07:

* Renamed `LinkUGen` to `LinkPhase`
  * Renamed `LinkUGen.enable` to `start` and `disable` to `stop`
  * Moved `server` into the argument
  * Removed state management of active clock from sclang side
  * Renamed `LinkUGen.setTempo` to `LinkPhase.setBPM`
  * Added `lag` Parameter to `LinkPhase.setBPM` which allows for a crossfade between existing tempo and target tempo over a specified amount of seconds
* Renamed `LinkTempo` to `LinkBPM` as the lingua-franca on Link is `bpm` which is rather unconventional for SuperCollider - to make this clearer I decided to make this transparent via the UGen
* Fixed init behavior of Link UGens by switching to `set_calc_function`
* Added docs
* Added tests
* Fixed Windows linking error (thx christof!)
* Make UGens multi-channel expandable (not `LinkEnabler` and `LinkDisabler` though)

capital-G:
@telephon @jamshark70 IMO this is now in a finished state. The only question I have is if we should stick to `bpm`, which is Ableton Link speak and makes the limits of 20-999 bpm a bit more obvious, or switch to `cps` which is SC speak (or tps, bps?), which would result in a range between `[0.33, 16.65]`?

capital-G:
Will be switched to cps to be consistent with existing `AbletonLink` and SC clocks in general.

telephon:
I just checked, nowhere else in sclang do we have a reference to BPM – we assume that "everybody knows" that a minute has 60 seconds, so that we can just set the tempo by `.tempo_(bpm / 60)`. 

Now there may by an argument that this is a waste of a division (since ableton uses bpm), but I lean towards just omitting all BPM related vocabulary for uniformity.

The alternative would be to add BPM stuff to all clocks, I find this not so attractive.

capital-G:
This is ready for review now

HelpSource/Classes/LinkCPS.schelp
```diff
@@ -0,0 +1,25 @@
+TITLE:: LinkCPS
+summary:: Gets and sets the cycles per second of the Link clock of the server
```
telephon: better make explicit that it is a UGen (true for all the other ones as well). E.g. "UGen that gets and sets ..."

HelpSource/Classes/LinkCPS.schelp
```diff
@@ -0,0 +1,25 @@
...
+categories:: UGens>Clock, Streams-Patterns-Events
+related:: Classes/LinkPhase, Classes/LinkClock, Classes/LinkJump
+
+DESCRIPTION::
+
+Allows to get and set the CPS (cycles per second) of the servers Link clock, such that the CPS can be modulated through a signal.
```
telephon: –> "A UGen that allows us to get and set the CPS" (in standard English, either "allow us to get" or "allow for getting".

HelpSource/Classes/LinkPhase.schelp
```diff
@@ -0,0 +1,104 @@
...
+argument:: server
+A LINK::Classes/Server::, defaults to LINK::Classes/Server#-default::.
+
+METHOD:: stop
+Exit a Link session within the network.
+All informations from the Link session will be set to MATH::-1.0::.
+argument:: server
+A LINK::Classes/Server::, defaults to LINK::Classes/Server#-default::.
+
+METHOD:: setCPS
```
telephon: just to be sure: shouldn't we just say `setTempo`? Usually it is tempo that is measured in cps.

SCClassLibrary/Common/Audio/LinkUGen.sc
```diff
@@ -0,0 +1,59 @@
+LinkPhase : UGen {
+	*start {|server=nil|
```
telephon: in the class library, this is always written without a default argument: `*start {|server|`. I see why you want it, but let's just keep it to the standard in this case.

SCClassLibrary/Common/Audio/LinkUGen.sc
```diff
@@ -0,0 +1,59 @@
...
+		{FreeSelf.kr(LinkEnabler.kr)}.play(server ? Server.default);
```
telephon: The method function.play will choose the default server, just let it do it and omit "? Server.default".

server/plugins/LinkUGen.cpp
```diff
@@ -0,0 +1,166 @@
+#include <ableton/Link.hpp>
+
+#include "SC_InterfaceTable.h"
+#include "SC_Unit.h"
+#include "SC_PlugIn.hpp"
+
+static InterfaceTable* ft;
+
+/*! we only have one global link instance per server.
```
telephon, line 9: if this stays, could you add a note to the helpfile? Also this may change in the future, so good to be explict about.

server/plugins/LinkUGen.cpp
```diff
@@ -0,0 +1,166 @@
...
+ *  we do not have access to this value through the Plugin API.
+ */
+static ableton::Link* LINK_CLOCK = nullptr;
+
+/*!
+ * @class LinkControl
+ * @brief Enables or disables a Link clock, which needs to be deferred to
+ * a non-RT thread.
+ */
+template <bool IsEnabler> class LinkControl : public SCUnit {
```
telephon: Wouldn't it be simpler to have a member variable `bool isEnabler` rather than adding this to the type interface?

telephon (commented):
Just a few comments for your consideration. Thanks!!

server/plugins/LinkUGen.cpp
capital-G: The benefit of using a template variable here is that two branchless implementations of LinkControl are generated, moving the runtime checks/branching to compile time.

This is probably optimized anyway when enabling aggressive compiler optimizations, but we already have much more awkward optimizations in the code base.

capital-G:
Thanks for the feedback - I hope I addressed most of it correctly

server/plugins/LinkUGen.cpp
```diff
@@ -0,0 +1,166 @@
...
+public:
+    LinkControl() {
+        mCalcFunc = make_calc_function<LinkControl, &LinkControl::next_k>();
+        if (IsEnabler ? !LINK_CLOCK->isEnabled() : LINK_CLOCK->isEnabled()) {
+            ft->fDoAsynchronousCommand(mWorld, nullptr, nullptr, nullptr, &LinkControl::set_clock_status, nullptr,
+                                       nullptr, &LinkControl::do_nothing, 0, nullptr);
```
Spacechild1: `do_nothing` could be replaced by a lambda.

server/plugins/LinkUGen.cpp
Spacechild1: > Wouldn't it be simpler to have a member variable bool isEnabler rather than adding this to the type interface?

We can't really use a member variable here because these are supposed to be two different classes `LinkEnabler` and `LinkDisabler`.

The alternative would be to make two different classes, so this is more about avoiding code duplication.

Personally, I would have just made two classes because I think it makes the code easier to read. But that's just my feeling.

SCClassLibrary/Common/Audio/LinkUGen.sc
```diff
@@ -0,0 +1,59 @@
+LinkPhase : UGen {
+	*start {|server|
+		{FreeSelf.kr(LinkEnabler.kr)}.play(server);
```
Spacechild1: It seems like the only thing `LinkEnabler` does is enable/disable the LinkClock. In this case this should really be a plugin command.

SCClassLibrary/Common/Audio/LinkUGen.sc
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	*stop {|server|
+		{FreeSelf.kr(LinkDisabler.kr)}.play(server);
```
Spacechild1: Same as above.

server/plugins/LinkUGen.cpp
Spacechild1: Now that I read the sclang code, it seems like this whole class can be replaced by a simple plugin command.

SCClassLibrary/Common/Audio/LinkUGen.sc
capital-G: Do we have use plugin commands in the standard library so far?

My motivation of avoiding a plugin command was that other frontends such as supriya can adapt the UGen more easily b/c it doesn't require custom logic as everything is a UGen.

Spacechild1: > Do we have use plugin commands in the standard library so far?

No, but we have plenty of asynchronous commands. Plugin commands allows us to add asynchronous commands without adding a new Server command.

> My motivation of avoiding a plugin command was that other frontends such as supriya can adapt the UGen more easily

A plugin command is just an OSC message, so I don't see how it would make it harder for other frontends.

capital-G: > A plugin command is just an OSC message, so I don't see how it would make it harder for other frontends.

Well, it requires custom logic since plugin commands are not used that commonly ;) ofc it is only sending OSC messages which every client needs to have implemented, but still.

capital-G: But maybe just do it as plugin command - I wasn't sure about which path to choose and since you have a preference it is settled ;)

Spacechild1: If a client wants to implement `Buffer -read`, it has to implement the corresponding Server command. So what's the issue with `LinkPhase -start` in particular? EDIT: I didn't see your last post.

HelpSource/Classes/LinkJump.schelp
```diff
@@ -0,0 +1,59 @@
...
+::
+
+For further information on Link clock see LINK::Classes/LinkClock:: and the official Link documentation at https://ableton.github.io/link/ .
+
+CLASSMETHODS::
+
+METHOD:: kr
+If the server clock has not been enabled via CODE::LinkPhase.start::, it will return MATH::-1.0::, otherwise it will return MATH::0.0::.
+argument:: trigger
+A value of MATH::>0.0:: will invoke a jump using the following arguments.
```
Spacechild1: That's not how triggers work in the Class Library. A trigger input only triggers when it goes from non-positive to positive. Check out `RecordBuf` as an example.

You have to update the C++ code accordingly. The general pattern goes like this:
```c++
float trig = IN0(<index>);
if (trig > 0.f && prevtrig <= 0.f) {
    // triggered
}
unit->m_prevtrig = trig;
```

server/plugins/LinkUGen.cpp
```diff
@@ -0,0 +1,166 @@
...
+public:
+    LinkJump() { set_calc_function<LinkJump, &LinkJump::next_k>(); }
+
+private:
+    bool mWarned = false;
+
+    void next_k(int numSamples) {
+        float* out = mOutBuf[0];
+        if (LINK_CLOCK->isEnabled()) {
+            if (in0(0) > 0.0f) {
```
Spacechild1, line 94: This is not a trigger. See https://github.com/supercollider/supercollider/pull/6947/changes#r2723180059

Spacechild1:
> How to obtain hardware buffer size? :?

I think we could add the hardware buffer size (and also the current OSC timestamp) to the `World` struct as part of the plugin API revamp. (I would also need it for my `aoo` external.)

HelpSource/Classes/LinkJump.schelp
capital-G: I know - I am not such a fan of triggers since they limit the speed in which things can be triggered to sr/2 - or in this case bs/2 which is already considerable.
Is there an advantage to using triggers over saying anything >0.0 will trigger? They also tend to induce additional memory and calculation^^

Spacechild1: > Is there an advantage to using triggers over saying anything >0.0 will trigger?

As always: consistency. "trigger" has a well-defined meaning within the ClassLibrary and it is used in plenty of UGens. You would need a *very* good reason to depart from that convention.

> or in this case bs/2 which is already considerable.

You could make the trigger audio rate. However, is `LinkJump` something you would trigger several times within a block?

capital-G: One reason why triggers are bad - a trigger acting upon a trigger now reduces the max resolution to 4 times the block size :/

```supercollider
	*setTempo {|cps, lag=0.0, curve=\exp, server=nil|
		lag = lag.max(0.0);
		if(lag > 0.0, {
			{
				var startCps = Latch.kr(in: LinkCPS.kr, trig: 1.0);
				LinkCPS.kr(
					// create alternating signals to trigger as fast as possible
					trigger: ToggleFF.kr(Impulse.kr(ControlDur.ir.reciprocal/2.0)),
					cps: Env([startCps, cps], [lag], curve).kr(
						doneAction: Done.freeSelf,
					);
				);
			}
		}, {
			{
				LinkCPS.kr(1.0, cps);
				FreeSelf.kr(1.0);
			}
		}).play(server);
	}
```

edit: we would need to add some additional time b/c we may end on a position where we do not trigger, resulting in an unreached final envelope value for the link clock :/

Spacechild1: I don't quite understand the code above.

> `var startCps = Latch.kr(in: LinkCPS.kr, trig: 1.0);`

Why do you need the `Latch` here? Why can't we pass `LinkCPS.kr` directly to the envelope?

> `ToggleFF.kr(Impulse.kr(ControlDur.ir.reciprocal/2.0)),`

First off, `ControlDur.ir.reciprocal` is the same as `ControlRate.ir` ;-)

`Impulse` is already a trigger signal; why do we need `ToggleFF`? 

And again, if you think the trigger is too slow, you can make it audio rate.

---

All that being said, maybe what we actually need here is not a trigger but a **gate**. In that case, the original `> 0` behavior would be correct. The documentation may read like this:

"While the gate is open (> 0.0), the CPS (second argument) will be used to set the Link clock and published to the other clients."

Terminology matters!

capital-G: Switched code to gate

server/plugins/LinkUGen.cpp
```diff
@@ -0,0 +1,162 @@
...
+                LINK_CLOCK->enable(true);
+                // do not continue to stage 3
+                return false;
+            };
+            auto disableClock = [](World*, void*) -> bool {
+                LINK_CLOCK->enable(false);
+                return false;
+            };
+
+            AsyncStageFn callback = start ? +activateClock : +disableClock;
```
Spacechild1: The plus signs first had me stumped because I had never seen this before. Then I learned that this is a trick to force a conversion to function pointer.

In this case, this seems to be a workaround for a long-standing bug in MSCV where it would try to find the common type for the result of the ternary operator *before* converting the lambdas to function pointers.

I would suggest to simply use `AsyncStageFn` instead of `auto` for the two lambda variables.

capital-G:
@Spacechild1 @telephon are there something missing from this getting merged? Just so I know if additional work is necessary, b/c from my side this PR is in a good state now.

Spacechild1:
Looks good from the C++ side! I can't really comment on the actual functionality because I have zero practical experience with Ableton Link.

HelpSource/Classes/LinkCPS.schelp
```diff
@@ -0,0 +1,25 @@
...
+Ableton Link internally uses BPM which gets converted by the server to CPS by dividing the BPM value by 60.
+See LINK::Classes/LinkPhase:: for examples.
+
+CLASSMETHODS::
+
+METHOD:: kr
+Returns the current CPS of the Link clock of the server.
+If the server clock has not been enabled via LINK::Classes/LinkPhase#*start::, it will return MATH::-1.0::.
+argument:: gate
+If >0, the CPS (second argument) will be used to set for the Link clock and published to the other clients.
```
telephon: If code::>0::, the CPS (second argument) will be used to set the Link clock and published to the other clients.

(used to set for the -> used to set the)

HelpSource/Classes/LinkCPS.schelp
```diff
@@ -0,0 +1,25 @@
...
+argument:: cps
+The CPS to be set.
```
telephon: The tempo to be set in cycles per second.

HelpSource/Classes/LinkPhase.schelp
```diff
@@ -0,0 +1,108 @@
...
+
+METHOD:: setTempo
+Helper method to set the CPS of a Link clock.
+NOTE::
+See also LINK::Classes/LinkCPS::.
+::
+argument:: cps
+New cps of the Link session
+argument:: lag
+If a value is set and MATH::>0.0::, the current cps will be cross-faded into the target cps over this time in seconds.
```
telephon: since the tempo is a value, it ic clearer to say that

"the tempo linearly changes from current to the target cps"

telephon: But from the comment below I notice that there can be overlapping processes. Wouldn't it be the best if the result would be the mean value of all parallel processes?

SCClassLibrary/Common/Audio/LinkUGen.sc
```diff
@@ -0,0 +1,55 @@
+LinkPhase : UGen {
+	*start {|server|
+		server = server ? Server.default;
+		server.sendRaw(LinkPhase.startMsg(true).asRawOSC);
+	}
+
+	*stop {|server|
+		server = server ? Server.default;
+		server.sendRaw(LinkPhase.startMsg(false).asRawOSC);
```
telephon: Better:
```
server.sendRaw(this.startMsg(false).asRawOSC);
```

SCClassLibrary/Common/Audio/LinkUGen.sc
```diff
@@ -0,0 +1,55 @@
...
+			}
+		}, {
+			{
+				LinkCPS.kr(1.0, cps);
+				FreeSelf.kr(1.0);
+			}
+		}).play(server);
+	}
+
+	*kr {|quantum=4|
```
telephon: could you move this up as a first method?

HelpSource/Classes/LinkPhase.schelp
capital-G: This would require some bus arithmetic since we can't do this on a UGen level - I would therefore say this is a convenience method for basic usage - everything more advanced should be coded by the user or by other libraries.

SCClassLibrary/Common/Audio/LinkUGen.sc
capital-G: I think I need to use `LinkPhase` instead of `this` since`startMsg` is a static method?

telephon: What is a static method in sclang? (you meant class method, maybe?)

telephon: in a class method, "this" refers to the class `LinkPhase` which is exactly what you want, no?

Spacechild1: Side note: Why `sendRaw`? Just do:
```
server.listSendMsg(this.startMsg(false));
```
or
```
server.sendMsg(*this.startMsg(false));
```
This is the pattern that is used in the Class Library.

capital-G:
> * if you spell out "cps" as a word, many sentences don't really make proper sense: we don't literally "set the cycles per second" but we "set the tempo".

I think tempo is an umbrella term - the link documentation defines

> Tempo is a well understood parameter that represents the velocity of a beat timeline with respect to real time, giving it a unit of beats/time.

and commonly uses bpm as the unit for tempo, as most other DAWs do.

> So I'd suggest that all argument names and class name components that are called "cps" should actually be called "tempo".

I think renaming to `LinkTempo` is a good change - it was actually named that way in my first implementation.
But I think naming the arguments `cps` makes it clear what kind of value is expected - e.g. `Car.setVelocity(tempo: 20);` would be too vague, but `Car.setVelocity(kmph: 20)` is clearer to the user.

> * Naive question: if a clock is set while the gate is > 0, rather than on a trigger, it is a little hard to make sure you set the clock exactly once, no?

I don't understand? Something like `Impulse.kr` is suited for this and allows to also set the value on two consecutive blocks (via `DC.kr(1.0)`) which is desirable for morphing values.

SCClassLibrary/Common/Audio/LinkUGen.sc
capital-G: I don't know - I started to use `sendRaw` when working with DynGen in order to obtain the size and I actually don't know how to send stuff to the server other way - I forgot about `sendMsg`^^
I think this is not really interesting b/c this is an implementation detail which is opaque to the user - but I can change it if you think it improves the situation.

capital-G: > in a class method, "this" refers to the class `LinkPhase` which is exactly what you want, no?

I remember why I switched to using the full class identifier - given the class

```supercollider
Foobar {
	*new {
		^super.newCopyArgs();
	}
	*foo {
		^"static";
	}
	foo {
		^"method";
	}
	baz {
		^this.foo;
	}
}
```

calling

```supercollider
~x = Foobar();
~x.baz;
```

now always calls the method - which is what it should do, but it created some problems for me once... burned fingers etc^^

So I think using the class name for static methods is more explicit in these cases - and I personally am always a fan of being more explicit in code ;)

But if you think `this` is better I can adapt, but I find the class name method easier to follow and less ambiguous.

---

> What is a static method in sclang? (you meant class method, maybe?)

What non-smalltalk languages commonly refer to as static methods ;)

Spacechild1: > I think this is not really interesting b/c this is an implementation detail which is opaque to the user

Please do! Such implementation details matter because we want consistency in the Class Library codebase. We should not break existing patterns unless there is a good reason to do so.

Spacechild1: > But if you think this is better I can adapt, but I find the class name method easier to follow and less ambiguous.

Actually, there is no ambiguity: in an instance method `this` *always* refers to the instance; in a class method `this` always refers to the Class. You cannot use `this` to call class methods from an instance method. In your example above, `Foobar -foo` does *not* shadow `Foobar *foo`!

@telephon is right that there is no reason for doing `LinkPhase.startMsg` here.

Spacechild1:
> So I'd suggest that all argument names and class name components that are called "cps" should actually be called "tempo".

I very much agree! The Class Library clearly defines "tempo" as the number of beats per *second* and it would be consistent with TempoClock.

"cps" is generally used for frequencies, like in filters and oscillators. I certainly wouldn't use it in this context.

> I think tempo is an umbrella term - the link documentation defines

It does not really matter how Link itself defines "tempo", this is just an implementation detail. What matters is the usage in the Class Library, because that's what users are familiar with and will expect.

SCClassLibrary/Common/Audio/LinkUGen.sc
telephon: The reason why not to use the class name is that potential future subclasses can't define their own `*startMsg`.

telephon:
> But I think naming the arguments `cps` makes it clear what kind of value is expected

Yes, that is ok. I just find it slightly irritating because nowhere else do we bother, so it seems as if it were something special all of a sudden.

So see @Spacechild1 's comment, I tend to see it the same way.

telephon:
> Something like `Impulse.kr` is suited for this

ok, if it is checked every block, this makes sense.

HelpSource/Classes/LinkPhase.schelp
capital-G: Settled for

```
If MATH::>0.0::, the current tempo will be faded into the target tempo over this value in seconds using an LINK::Classes/Envelope::.
If this value is CODE::0.0::, the clock will jump immediately to the new tempo.
```

since we can't be sure if it is linear or not (default is exponential since cps is logarithmic).

capital-G:
I think I now got it - I hope I addressed all of your concerns. I think we are getting a really good API here :)

Spacechild1:
Looks good to me now!

capital-G:
Hey - thanks for reviewing. I think we made an okay-ish implementation into something really neat. Looking forward to see its usage in the wild :)

PR #6948, opened 2025-06-08
NodePlotView: fix a bug &  add scroll to view parent

prko:
## Purpose and Motivation
#6294 introduces the following error
```
ERROR: Message 'view' not understood.
Perhaps you misspelled 'asView', or meant to call 'view' on another receiver?
RECEIVER:
Instance of View {    (00000298416839C8, gc=F4, fmt=00, flg=00, set=06)
  instance variables [34]
    qObject : RawPointer 000002983DF33270
    finalizer : instance of Finalizer (00000298420542F8, size=2, set=1)
    virtualSlots : nil
    wasRemoved : false
    font : nil
    resize : Integer 1
    alpha : Float 1.000000   00000000 3FF00000
    decorator : nil
    layout : nil
    userCanClose : true
    deleteOnClose : true
    action : nil
    mouseDownAction : nil
    mouseUpAction : nil
    mouseEnterAction : nil
    mouseLeaveAction : nil
    mouseMoveAction : nil
    mouseOverAction : nil
    mouseWheelAction : nil
    keyDownAction : nil
    keyUpAction : nil
    keyModifiersChangedAction : nil
    keyTyped : nil
    focusGainedAction : nil
    focusLostAction : nil
    dragLabel : nil
    beginDragAction : nil
    canReceiveDragHandler : nil
    receiveDragHandler : nil
    toFrontAction : nil
    endFrontAction : nil
    onClose : nil
    onResize : nil
    onMove : nil
}
ARGS:
KEYWORD ARGUMENTS:
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg this = <instance of View>
		arg selector = 'view'
		arg args = [*0]
		arg kwargs = [*0]
	NodeTreeView:makeWindow
		arg this = <instance of NodeTreeView>
		arg bounds = <instance of Rect>
		arg parent = <instance of View>
	Server:plotTreeView
		arg this = <instance of Server>
		arg interval = 0.5
		arg parent = <instance of View>
		arg actionIfFail = nil
		var res = nil
		var code = "(
w = Window();
w.front;

b ..."
		var doc = nil
	Process:interpretPrintCmdLine
		arg this = <instance of Main>
^^ ERROR: Message 'view' not understood.
Perhaps you misspelled 'asView', or meant to call 'view' on another receiver?
RECEIVER: a View
```
when evaluating the following code:
```supercollider
(
w = Window();

b = View(w, w.bounds.moveTo(0, 0));
n = View(b, b.bounds);
s.plotTreeView(0.5, n);

w.front;
)
```
This PR fixes this wrong behaviour. (SC3.13.0 does not return this error and creates the requested GUI.) 
(In #6834, #6853, #6914 and #6917, this bug is not addressed.)

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
@@ -21,7 +21,15 @@ NodeTreeView {
 		}
 		.front;
 
-		view = UserView(window, window.view.bounds.resizeBy(-4));
+		view = UserView(
+			window,
+			if(window.isView)
+			{
+				window.bounds.resizeBy(-4)
+			} {
+				window.view.bounds.resizeBy(-4)
+			}
+		);
 		view.onClose_{
 			view = nil;
 			this.stop;
@@ -164,7 +172,7 @@ NodeTreeView {
 						size = size + (countSize.value(elem.value) * sizeFactor) + 2.5
 					} {
 						size = size + 1
-				};
+					};
 				});
 				size
 			};
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
@@ -17,7 +17,7 @@ NodeTreeView {
 		if(parent.isNil) {
 			window = Window(this.asString, bounds, scroll:true);
 		} {
-			window = if(parent.isView) {
+			window = if((parent.class.asString == "View")) {
 				parent.scroll(bounds: parent.bounds.moveTo(0, 0), hasVerticalScroller: true)
 			} {
 				window = parent
@@ -27,7 +27,7 @@ NodeTreeView {
 
 		view = UserView(
 			window,
-			if(window.isView)
+			if((parent.class.asString == "View"))
 			{
 				window.bounds.resizeBy(-4)
 			} {
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
@@ -1,5 +1,6 @@
 NodeTreeView {
-	var <server, <bounds, <window, view, resp, updateFunc, updater, tabSize = 25, treeViewStatus;
+	var <server, <bounds, <window, view, scrollView;
+	var resp, updateFunc, updater, tabSize = 25, treeViewStatus;
 	var runningFunc, stoppedFunc, serverController;
 	var globalAlpha = 1;
 
@@ -8,8 +9,7 @@ NodeTreeView {
 		^super.newCopyArgs(server).makeWindow(bounds, parent)
 	}
 
-	makeWindow 	{ |bounds, parent|
-		var scrollView;
+	makeWindow { |bounds, parent|
 		bounds = if(bounds.isNil) {
 			Rect(128, 64, 400, 400)
 		} {
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
@@ -17,7 +17,7 @@ NodeTreeView {
 		if(parent.isNil) {
 			window = Window(this.asString, bounds, scroll:true);
 		} {
-			window = if((parent.class.asString == "View")) {
+			window = if((parent.isKindOf(View))) {
 				parent.scroll(bounds: parent.bounds.moveTo(0, 0), hasVerticalScroller: true)
 			} {
 				window = parent
@@ -27,7 +27,7 @@ NodeTreeView {
 
 		view = UserView(
 			window,
-			if((parent.class.asString == "View"))
+			if((parent.isKindOf(View)))
 			{
 				window.bounds.resizeBy(-4)
 			} {
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
@@ -17,7 +17,11 @@ NodeTreeView {
 		if(parent.isNil) {
 			window = Window(this.asString, bounds, scroll:true);
 		} {
-			window = parent
+			window = if(parent.isView) {
+				parent.scroll(bounds: parent.bounds.moveTo(0, 0), hasVerticalScroller: true)
+			} {
+				window = parent
+			}
 		}
 		.front;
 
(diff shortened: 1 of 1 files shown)


prko:
@capital-G, @dyfer 
Should I open a new PR using 3.14 branch as I made #6968 for #6965?

dyfer:
> Should I open a new PR using 3.14 branch as I made #6968 for #6965?

@prko no need, I changed the destination branch to 3.14 here and this is still a clean merge against 3.14, as there were no diverging commits between `develop` and `3.14` yet.

prko:
As `.isView` is not a built-in method, I changed the algorithm to check whether the instance's class is "View".

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
```diff
@@ -17,7 +17,7 @@ NodeTreeView {
 		if(parent.isNil) {
 			window = Window(this.asString, bounds, scroll:true);
 		} {
-			window = if(parent.isView) {
+			window = if((parent.class.asString == "View")) {
```
telephon: This is not the usual way this is done in sclang, and generally testing for a specific class is done only in very exceptional cases. There are different ways to do it, in this case I'd say that you probably want `if(parent.isKindOf(View)`?

prko: Yes! I forgot `isKindOf(...)`... sorry and thanks!
Done in https://github.com/supercollider/supercollider/pull/6948/commits/b285d55861df421c9ccc07b6734e3f126d7f6e94

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
```diff
@@ -17,11 +17,23 @@ NodeTreeView {
...
-		view = UserView(window, window.view.bounds.resizeBy(-4));
+		view = UserView(
+			window,
+			if((parent.isKindOf(View)))
+			{
+				window.bounds.resizeBy(-4)
+			} {
+				window.view.bounds.resizeBy(-4)
+			}
+		);
```
dyfer: I think a simpler solution is to write
```supercollider
view = UserView(window, window.asView.bounds.resizeBy(-4));
```
Can you confirm this addresses the issue you are trying to fix?

dyfer: Actually, I think this can be addressed differently - see the other comment.

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
```diff
@@ -17,11 +17,23 @@ NodeTreeView {
 		if(parent.isNil) {
 			window = Window(this.asString, bounds, scroll:true);
 		} {
-			window = parent
+			window = if((parent.isKindOf(View))) {
+				parent.scroll(bounds: parent.bounds.moveTo(0, 0), hasVerticalScroller: true)
+			} {
+				window = parent
+			}
```
dyfer: I think this can still be improved, as the logic is a bit convoluted.

In particular, the currently proposed implementation does not show scroll bars when the parent is a Window:
```supercollider
(
w = Window().front;
s.plotTreeView(0.5, w);
)
// create nodes to require scrolling
```

I think it would be cleaner to do:

(Window or Parent, no scrolling needed) > explicitly created ScrollView > UserView 

Sorry for rewriting this section, but I think this would be cleaner. What do you think?

```supercollider
	makeWindow { |bounds, parent|
		var scrollView;
		bounds = if(bounds.isNil) {
			Rect(128, 64, 400, 400)
		} {
			bounds.asRect.minSize(395@386)
		};
		if(parent.isNil) {
			window = Window(this.asString, bounds);
		} {
			window = parent
		}
		.front;

		scrollView = ScrollView(window, window.bounds.moveTo(0, 0));
		scrollView.hasVerticalScroller_(true);
		scrollView.hasHorizontalScroller_(false);
		scrollView.hasBorder_(false);

		window.asView.onResize_({scrollView.bounds_(window.bounds.moveTo(0, 0).resizeBy(-4))});

		view = UserView(scrollView, scrollView.bounds.resizeBy(-4));
		view.onClose_{
			view = nil;
			this.stop;
		};

		treeViewStatus = StaticText(
			view,
			Rect(tabSize, 0, view.bounds.width - (tabSize * 2) + 1, tabSize);
		).align_(\center).stringColor_(Color.grey(0.4));

	}
```

Note: **please make `scrollView` an instance variable**, like the `view` variable, instead of being local to this function. 

In this case we don't create a ScrollView implicitly when we create window, but we always create a ScrollView inside whatever parent is passed, and we resize it with the parent.

Thoughts?

prko: I think the short code is always better than the longer one.

Thanks!

Done in https://github.com/supercollider/supercollider/pull/6948/commits/3395c5363514e7c728ea38cdd18907386531ffe3!

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
prko: Thanks!
Reflected in https://github.com/supercollider/supercollider/pull/6948/commits/3395c5363514e7c728ea38cdd18907386531ffe3!

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
```diff
@@ -8,20 +8,28 @@ NodeTreeView {
 		^super.newCopyArgs(server).makeWindow(bounds, parent)
 	}
 
-	makeWindow { |bounds, parent|
+	makeWindow 	{ |bounds, parent|
```
dyfer: Can you remove the whitespace change here? Sorry if that was in the code I posted.

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
```diff
@@ -8,20 +8,28 @@ NodeTreeView {
...
+		var scrollView;
```
dyfer: Can you move it so that `scrollView ` is together with other instance variables? Technically it doesn't have to be, but we're not using local variables elsewhere in this method so I'd like to do that for consistency.

It can go after `view` or be in a separate line.

dyfer (changes requested):
Thanks for the changes! I have small outstanding requests.

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
prko: Done in https://github.com/supercollider/supercollider/pull/6948/commits/5a271ff11407fb63c63ecf89bab8c9a3e0901415!

SCClassLibrary/Common/GUI/PlusGUI/Control/NodeTreeView.sc
prko: Done in https://github.com/supercollider/supercollider/pull/6948/commits/5a271ff11407fb63c63ecf89bab8c9a3e0901415!

PR #6950, opened 2025-06-08
Topic/fix exit codes

JordanHendersonMusic:
## Purpose and Motivation

```supercollider
1.exit;
"still on".postln;
0.exit;
```
This returns an exit code of `0` because there is no good way to stop the interpreter.

This fix works by emitting a new bytecode and creating a new keyword in sc.

That keyword is `pr_force_exit_sc` ... hopefully no one uses that for anything!

Technically, all this bytecode does is `return`.

@capital-G @elgiano 

This PR is built on top of #6749 so it has all the commits from that. Once that is merged, this is a very small single commit, see https://github.com/supercollider/supercollider/pull/6950/commits/2c72d1f8a9aa28069a3304295fa9a5eeac958b41

## Types of changes

- Bug fix
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


prko:
Thanks for this!
Could you check #6291?
If you think the text for `Integer.schelp` in #6291 is inappropriate, could you please edit it in this pull request (PR)? I will then close #6291.

capital-G:
Thanks for this fix :) But shouldn't this PR also resolve the conflicting primitive name `_Exit`? It seems the primitive is still implemented two times?

* https://github.com/supercollider/supercollider/blob/13a31c2b33617cfdae674f4c9aeaea36f1eac9df/lang/LangPrimSource/OSCData.cpp#L1610
* https://github.com/supercollider/supercollider/blob/13a31c2b33617cfdae674f4c9aeaea36f1eac9df/lang/LangSource/SC_TerminalClient.cpp#L227

Although it is only specified/used once within the std lib

https://github.com/supercollider/supercollider/blob/13a31c2b33617cfdae674f4c9aeaea36f1eac9df/SCClassLibrary/Common/Math/Integer.sc#L185-L188

BTW - could you quickly explain the advantage of a dedicated byte code over plain [`exit(1)`](https://en.cppreference.com/w/cpp/utility/program/exit)? IIUC `std::exit` would also terminate all attached resources, and the OS would also guarantee us that no open resources are dangling afterwards?

JordanHendersonMusic:
Hmm, actually I've looked again at this and don't think this is the way the vm was designed.... I do however think it is the right design choice, having the compiler emit a code that says 'stop' would mean we don't need the long jump...

JordanHendersonMusic:
So just to expand, the way the main intrepreter loop is existed is through a long jump hidden away in `methodReturn`... it isn't very elegant, nor intuitive. However, the functionality to exit the interpreter loop has always existed, it just wasn't being used. I'll make a new PR.

PR #6951, opened 2025-06-09
sclang: throw fatal exception

JordanHendersonMusic:
## Purpose and Motivation

When sclang has a fatal error and throws, the exception is caught in `runLibrary`, meaning the vm just sits there in an invalid state, without informing the call sight. Now the exception is bubbled up the call stack, cause a proper error where ever it is need. 

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/lang/LangSource/PyrLexer.cpp
+++ b/lang/LangSource/PyrLexer.cpp
@@ -2262,7 +2262,7 @@ void signal_init_globs();
 
 void dumpByteCodes(PyrBlock* theBlock);
 
-SCLANG_DLLEXPORT_C void runLibrary(PyrSymbol* selector) {
+SCLANG_DLLEXPORT_C void runLibrary(PyrSymbol* selector) noexcept(false) {
     VMGlobals* g = gMainVMGlobals;
     g->canCallOS = true;
     try {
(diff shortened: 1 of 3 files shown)


JordanHendersonMusic:
@scztt does this implement the feature you needed in #6671?

lang/LangSource/PyrLexer.cpp
```diff
@@ -2273,7 +2273,10 @@ SCLANG_DLLEXPORT_C void runLibrary(PyrSymbol* selector) {
         } else {
             postfl("Library has not been compiled successfully.\n");
         }
-    } catch (std::exception& ex) {
+    } catch (const FatalInterpreterError& er) {
+        error("A fatal interpreter error has occured. Reason: %s\n", er.what());
+        std::rethrow_exception(std::current_exception());
```
Spacechild1: If you want to rethrow the current exception, you can simply write `throw;` (without any exception object!)

JordanHendersonMusic: Oh I didn't know that! I'll change it when I get the change. Thanks!

Spacechild1 (approved):

PR #6953, opened 2025-06-09
classlib, sclang: actually exit the interpret when calling `Integer.exit`

JordanHendersonMusic:
## Purpose and Motivation

```supercollider
32.exit;
1.exit;

// returns exit code 1.
```

The way supercollider exits the main interpreter loop is confusing. It does so through the `returnMethod` functions, which try to find the parent sclang method, if it can't, then it will exit.... however it does so through a long jump. Perhaps an alternative design would have been to make the compile add an `exit` byte code that leaves the main loop, this way the main `runLibrary` function could return an optional exitcode, however...

Exit codes live inside the `SCLanguageClient` as a member. The exit primitive sets this, but didn't call the long jump to exit. We can't just call `std::exit` because there is cleanup to do, when we create an exit code it isn't an error, but a polite request the sc stop.

An example of where stopping the interpreter is implement correctly in sc is with the `Object.halt` method.

This PR copies the body of halt into the body of exit. I don't want to call it because it is a primitive method and might in the future do some stack stuff that isn't compatible with exit.  

I've copied all of the code, including the comments, because I am not sure what they are referring to.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Core/Object.sc
+++ b/SCClassLibrary/Common/Core/Object.sc
@@ -331,7 +331,7 @@ Object {
 		OnError.run;
 		this.prHalt
 	}
-	// _Halt will exit the intrepreter, but isn't buy itself an error, see Integer.exit.
+	// _Halt will exit the interpreter, but isn't by itself an error, see Integer.exit.
 	// Object.halt does trigger OnError.
 	prHalt {
 		_Halt


prko:
Thanks for this!
Could you check #6291?
If you think the text for Integer.schelp in #6291 is inappropriate, could you please add the corresponding text in this pull request (PR)? I will then close #6291.

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -331,6 +331,8 @@ Object {
 		OnError.run;
 		this.prHalt
 	}
+	// _Halt will exit the intrepreter, but isn't buy itself an error, see Integer.exit.
```
dyfer: "buy itself" -> "by itself" ?

lang/LangSource/SC_TerminalClient.cpp
```diff
@@ -572,16 +572,28 @@ int SC_TerminalClient::prArgv(struct VMGlobals* g, int) {
...
+    // PyrSlot *bottom = g->gc->Stack()->slots;
+    // slotCopy(bottom,g->sp);
+    // g->sp = bottom; // ??!! pop everybody
+    g->method = nullptr;
+    g->block = nullptr;
+    g->frame = nullptr;
+    SetNil(g->sp);
+    longjmp(g->escapeInterpreter, 3);
+    // hmm need to fix this to work only on main thread. //!!!
+    // g->sp = g->gc->Stack()->slots - 1;
```
capital-G, line 594: should this and the comments above be part of this PR?

capital-G (commented):
Would still be interested in this, though I am unsure what the state of this PR is.

Is there a reason why a plain https://en.cppreference.com/w/cpp/utility/program/exit wouldn't be sufficient in this case? Is it necessary to "unwind" the interpreter?

JordanHendersonMusic:
Yes you need to unwind the interpreter to do proper cleanup. E.g., freeing finaliziers.

lang/LangSource/SC_TerminalClient.cpp
JordanHendersonMusic: I've copied these from the other place this occurs. I don't really understand the comments, they are from the original commit. So I don't really know if they are important. The code that is present looks correct and does what I'd expect. 

I could remove them?

capital-G: Ah I thought you added them during development as notes - better keep those around in this case :)

capital-G:
> Yes you need to unwind the interpreter to do proper cleanup. E.g., freeing finaliziers.

What would happen if you don't run/free the finalizers? Wouldn't the OS take care of this and any other open resources in the end? I am fine w/ the current approach, just want to learn something here :)

capital-G (approved):

lang/LangSource/SC_TerminalClient.cpp
dyfer: Do you think it would be worth making a note where these were copied from? Otherwise, looking at the source and not git history it will be hard to deduct the purpose.

JordanHendersonMusic: Good idea 👍

PR #6963, opened 2025-06-09
Fix typo in the 3.14 changelog

capital-G:
## Purpose and Motivation

Minor typo.

I think this should target the 3.14 branch which will then later be merged w/ dev again? If this would be merged onto develop there wouldn't be a way to get this into 3.14 again? (But this will also add the additional merge commits again which we don't want? - maybe we need to cherry pick here?)

Should be merged by @dyfer who is most familiar w/ the release process.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
Yes, we just merge it to 3.14 and then it'll make it to develop at the next merge 3.14->develop (at the next release at the latest, but could be sooner).

From https://github.com/supercollider/supercollider/wiki/git-workflow-and-guidelines#git-flow: 

> Periodically (roughly every week), the release branch is merged into develop so that develop always has both the latest bug fixes and features.

dyfer (approved):
Thanks for catching this!

PR #6966, opened 2025-06-10
Update News-3_14.schelp

telephon:
The ** syntax is not, or not yet, supported.


capital-G:
Oh right - thanks so much for spotting and fixing this! :)

Since this is a PR which targets 3.14, the PR branch needs to be based of the 3.14 release branch and the PR it needs to target this.
This should not go directly into develop (but via 3.14 into develop).

telephon:
Now this might work? Not sure, it wasn't easy. I shouldn't have tried to edit the document here directly :/

telephon:
Ah no, everything is messed up again, sorry. Need to do this later.

PR #6967, opened 2025-06-10
docs: Reference/Control-Structures update (WIP)

HotwheelsSisyphus:
## Purpose and Motivation

Noticed some things that could be better explained in the Control Structures help file; since this is a pretty central concept it might be a good thing to look over.

E.g., the introductory paragraph explains the Smalltalk/OOP structure of how if/while/etc work, but then doesn't set that in relation to the "possible but harder to read, so, not preferred" receiver notation `expr.if (trueFunc, falseFunc)`.

So I'm trying to add explanations/comments that clarify this, using the terminology in the Syntax Shortcuts file to distinguish the different notations. 

There is also the terminology question that I probably need some help with: 
Namely, currently, the file talks about "message sends" to an "expression".
I suppose that this is _not_ simply synonymous with "methods called on a receiver", but I don't understand the difference well enough currently... 
Even if it is not synonymous, is there a reason we shouldn't replace the message-send wording with method-calls? 

Anyway, here goes...

## Types of changes

Documentation

## To-do list
- [x] intro
- [x] if 
- [ ] while 
- [ ] etc.
- [ ] optimization: Move content from if.schelp in here if it isn't already in here?

### DIFF
--- a/HelpSource/Reference/Control-Structures.schelp
+++ b/HelpSource/Reference/Control-Structures.schelp
@@ -3,31 +3,41 @@ summary:: flow control
 categories:: Language
 related:: Classes/Boolean
 
-Control structures in SuperCollider are implemented via message sends. Here are a few of those available.
-See link::Reference/Syntax-Shortcuts:: for the various ways expressions can be written.
+Control structures in SuperCollider are implemented via message sends (written as methods called on a receiver). 
+Here are a few of those available. 
+Note that while the syntax can look C-like, code::if, while,:: etc. are emphasis::not:: keywords in SuperCollider but methods.
+See link::Reference/Syntax-Shortcuts:: for the various ways conditional expressions can be written.
 
 section:: Basic Control Structures
 
 method:: if
 
-Conditional execution is implemented via the code::if:: message. The code::if:: message is sent to an expression which must return a link::Classes/Boolean:: value.
+Conditional execution is implemented via the code::if:: message. 
+The code::if:: message is sent to an expression which must return a link::Classes/Boolean:: value.
 
-In addition it takes two arguments: a function to execute if the expression is true and another optional function to execute if the expression is false. The code::if:: message returns the value of the function which is executed. If the falseFunc is not present and the expression is false then the result of the if message is nil.
+In addition it takes two arguments: a function to execute if the expression is true, 
+and another optional function to execute if the expression is false. 
+The code::if:: message returns the value of the function which is executed. 
+If the falseFunc is not present and the expression is false, then the result of the if message is nil.
 
 discussion::
 Syntax
 code::
-if(expr) { true body } { false body };
+// receiver notation, reflects the above eplanation
+// but in practice is hard to read and rarely used
+expr.if (trueFunc, falseFunc);
 
-// alternate
+// function call notation
 if(expr, trueFunc, falseFunc);
 
-// possible but harder to read, so, not preferred
-expr.if (trueFunc, falseFunc);
+// function call notation with trailing argument block
+// looks like C. 
+if(expr) { trueBody } { falseBody };
 ::
 
 Examples
 code::
+// function call notation with trailing argument block
 (
 if([false, true].choose) {
 	"expression was true".postln  // true function
@@ -42,11 +52,20 @@ z = if(a < 5) { 100 } { 200 };
 z.postln;
 )
 
+// no falseBody, expr is true
 (
 var x;
 if(x.isNil) { x = 99 };
 x.postln;
 )
+
+// no falseBody, expr is false
+(
+var x = 1;
+if(x < 0) { x = 99 };
+x.postln;
+)
+
 ::
 
 


JordanHendersonMusic:
Great work!

I'll have a close look if I get time shortly. 

I don't think the three categories of if are that useful because only two are mutually exclusive. E.g., `a.if {a} {b}`. It might be better to just introduce trailing blocks separately from control structures, they work with any message. `a.foo {...} {...} ...` & `foo(a) {...} {...}...`

Something that isn't explicitly stated (though you do have an example) is that control structures (like everything) return a value.
`var a = if(b) {2} {3}`

Consider moving case closer to if, as this is how you should implement else if statements in SC.

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -0,0 +1,591 @@
...
+summary:: conditional expressions in SuperCollider
+categories:: Language
+related:: Classes/Boolean,  Classes/Function, Classes/Exception, Reference/Syntax-Shortcuts 
+
+This article gives an overview of the conditional expressions in SuperCollider and the many ways to write them.
+It starts with addressing the relevant syntactic generalities, then discusses the specific conditional expressions in question, 
+and finally touches on optimization issues, the (strongly discouraged) use of if-statements in Synths, and provides a table of classes implementing the methods discussed here.
+For discussions of specific expressions, skip ahead to their respective sections: 
+link::#if::, link::#while::, link::#switch::, link::#case::.
+
```
JordanHendersonMusic, line 7: It might be worth having a little table at the top that just summarises the recommended syntax for all these structures.

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -0,0 +1,591 @@
...
+If the number of arguments is even, code::switch:: will return teletype::nil:: in such cases, whereas if it is odd, the last argument should be a teletype::defaultFunction:: whose code::.value:: will be returned.
+
+The comparison function used is equality, i.e., code::==::, unless the switch
+statement is inlined (see also link::#Optimization through inline expansion#below::), 
+in which case switch compares by identity, i.e., code::===::. (See link::Guides/EqualityIdentity::)
+
+The switch statement will automatically be inlined if two conditions are met: 
+strong::First::, the test objects are all values with unique representations
+(Floats, Integers, Symbols, Chars, nil, false, or true) and strong::second::, functions used in the switch statement have no variable or argument declarations. 
+The inlined switch uses a hash lookup (which is faster than nested if statements), so it should be very fast and scale to any number of clauses.
```
JordanHendersonMusic, line 242: I don't quite remember if they do use hash lookup - I'll double check later unless you know for sure?

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -0,0 +1,591 @@
...
+While 1.0 and 1 represent the same numeric value, one is a Float and the other is an Integer, so they cannot be identical.
+On the other hand, if we prevent inlining by declaring a variable within one of the functions, the code will compare by equality:
+code::1 == 1.0:: returns true. 
+
+
+code::
+(
+// 'var x' prevents inlining
+switch(1)
+	{ 1.0 } { var x; "yes" }
```
JordanHendersonMusic, line 345: I did not know this...

I think if the case is in brackets, then we might consider making this a compiler error instead? It seems like this just adds to the confusion around inlining.

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -0,0 +1,591 @@
...
+(
+var i, test1, test2, true1, true2, default;
+i = [true, false].choose;
+test1 = {i};
+test2 = {i.not};
+true1 = { "i is true" };
+true2 = { "i is false" };
+default = { "just in case"};
+case(
+	test1, true1,
```
JordanHendersonMusic: So these aren't inlined. I think this is missing from your description throughout.

You need to literally write the character `{` for it to be inlined. This is because the 'function' isn't compiled as a function, but copied into the body of the parent scope. This is also why you can't have variables, because SC doesn't let you have variables in the middle of a block.

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -0,0 +1,591 @@
...
+		[SinOsc.ar(freq), Saw.ar(freq)]
+		) * 0.2;
+	}.scope(bufsize: s.sampleRate/freq);
+)
+::  
+
+section:: Table of conditional expressions and classes that implement them.
+
+The following table gives an overview of the control structures discussed here, and shows which classes implement them as methods. 
+
```
JordanHendersonMusic, line 561: I don't know if you want to list the classes because their implementations aren't used when things are inlined.

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -0,0 +1,591 @@
...
+table::
+	## strong::subsection:: || strong::Class::
+    ## if || Boolean
+	## while || Function
```
JordanHendersonMusic: Technically while is only implemented in the compiler... Take a look at the code for function.while and you will see why this whole inlining thing is confusing.

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -0,0 +1,591 @@
...
+This article gives an overview of the conditional expressions in SuperCollider and the many ways to write them.
+It starts with addressing the relevant syntactic generalities, then discusses the specific conditional expressions in question, 
+and finally touches on optimization issues, the (strongly discouraged) use of if-statements in Synths, and provides a table of classes implementing the methods discussed here.
+For discussions of specific expressions, skip ahead to their respective sections: 
+link::#if::, link::#while::, link::#switch::, link::#case::.
+
+section:: Syntax of conditional expressions
+
+Control structures in SuperCollider are implemented as methods called on a receiver; 
+thus, code::if, while,:: etc. are technically emphasis::not:: keywords in SuperCollider but methods,
```
JordanHendersonMusic: Not quite right, technically while is a keyword: it is its own token.

The other methods are short of keywords. SC has 'special' messages which are implemented with different bytecodes, if it finds one of these, it will check the name of the message, if it matches one of the known ones, it will compile it differently. 

So in the lexer, while is a keyword, but in the compiler, they all have special implementations (along with other things)... It is very confusing 😯 

I don't think this document should be a technical document - it shouldn't document compiler internals -, so maybe just avoid this kind of language?

HelpSource/Reference/Conditional-Execution.schelp
HotwheelsSisyphus: yes they do, at least according to the old version of the file ... it was a detail I cut. I can put that back in if you think it's a good idea!

HelpSource/Reference/Conditional-Execution.schelp
HotwheelsSisyphus: ah I see, will revise accordingly!

HelpSource/Reference/Conditional-Execution.schelp
HotwheelsSisyphus: Good idea. I think I was just trying to emphasize the difference between C syntax and sc syntax despite the superficial similarities, I probably don't need to talk about keywords to do that.

HelpSource/Reference/Conditional-Execution.schelp
HotwheelsSisyphus: yeah, it's probably also not super useful now I've removed the for/By, do, try and protect discussion. Might just remove the table.

HelpSource/Classes/Function.schelp
```diff
@@ -37,9 +37,7 @@ The global pseudo-variable code::thisFunction:: always evaluates to the current
 enclosing Function.
 
 note::
-Be aware of link::Reference/Control-Structures#Inline
-optimization#inline optimizations:: which will be reflected in the value of
-code::thisFunction::.
+Be aware of link::Reference/Conditional-Execution#Optimization through inline expansion:: which will be reflected in the value of code::thisFunction::.
```
telephon: maybe take the occasion to rewtite this sentence - it's really hard to understand.

telephon:
This needs to be explained, I think:

```supercollider
g = { thisFunction };
f = { if(true, g)  };
f.value === f; // false
f.value === g; // false
f.value === thisFunction; // also false
f.value.cs; // { thisFunction }
h = f.value;
g.identityHash // not the same as ...
h.identityHash // ... this one
```

HotwheelsSisyphus:
@telephon 

Hm, so I'm not sure if this has to do with the Conditional Expressions/Control Structures reference? If more explanation is needed than you already added, this should probably go into a separate update to Function.schelp. The reason I changed that file at all  was simply not to have dead links, so I did this with a blind find/replace... but thanks for correcting the link to the section!

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -185,63 +248,88 @@ if(x < 0) { 90 + x};
...
-it is a method of link::Classes/Function::.
-If the teletype::testFunc:: answers true when evaluated, 
-then the teletype::loopFunc:: is evaluated and the process is repeated. 
-Once the teletype::testFunc:: returns false, the loop terminates.
-
-Note the distinction to code::if::: 
-teletype::testFunc:: is a Function (which returns itself), e.g., code::{x < 0}::, 
-whereas teletype::expr:: above was an expression, e.g., code::x < 0::, which returned a Boolean.
+Function also implements a code::case:: method which allows for conditional evaluation with multiple cases. 
+The receiver here but can be thought of as simply the zeroth argument, representing the first of multiple "cases".
```
telephon, line 163: fix: "The receiver here but can be thought of"

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -137,15 +200,15 @@ expr.if(trueFunc, falseFunc);
 if(expr, trueFunc, falseFunc);
 
 // function call notation with trailing argument block
-// looks almost like C (without the else keyword).
+// looks almost like C (without the "else").
 if(expr) { trueFuncBody } { falseFuncBody };
 
 // example:
 if(0 < 3) {"aha".postln} {"brb".postln};
```
telephon: while you are at it, this would be better:

if(0 < 3) { "aha" } { "brb" };

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -108,10 +140,41 @@ while { testFuncBody } { loopFuncBody };
...
 // in receiver notation
 { testFuncBody }.while({ loopFuncBody })
-// distinct from C-style, keyword-based while, which will not work in SuperCollider:
+// distinct from C-style while, which will not work in SuperCollider:
 while (testFuncBody) { loopFuncBody }; // -> Syntax Error
 ::
 
+subsubsection:: Return values (.if, .case, .switch)
+
+Unlike other common programming languages, conditional expressions (code::.if::, code::.case::, and code::.switch::) in SuperCollider have return values by default;
```
telephon: This could be expressed more generally, in SuperCollider everything is an expression, i.e. returns a value. There are no pure statements as far as I know. Even `var x;` is an expression that returns `nil`.

So "have return values by default;" is a little misleading, because they always have a return value.

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -90,15 +116,21 @@ if(condition) {t} {f}; //permissible but it may not be what you want to do.
...
+if(condition, trueFuncBody, falseFuncBody); // the bodies are expressions, not Function objects;
 // whether this works depends on how those expressions respond to the .value method.
 ::
 
-subsection:: Receiver blocks (relevant to .while, .case, .try, .protect)
+subsubsection:: Receiver blocks (.case, .while)
+
+If the receiver is itself an expression surrounded by curly brackets, a special version of the trailing argument block syntax can be used:
+Here, not only the arguments can be written as trailing {}-enclosed blocks, but also the receiver itself.
+For instance, code::{ funcBody }.fork:: is often written as code::fork { funcBody }::.
```
telephon: This is slightly misleading, perhaps? Every method call can be expressed in two different ways: receiver notation and functional notation, recv.message(...) and message(recv ... ) respectively. 

```
f = { functionBody };
f.fork;
fork(f);


{ functionBody }.fork;
fork({ functionBody });
fork { functionBody }
```

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -90,15 +116,21 @@ if(condition) {t} {f}; //permissible but it may not be what you want to do.
 // equivalent to if(condition, {t}, {f}); 
 // i.e.will return t rather than t.value as usually intended.
 
-if(condition, trueFuncBody, falseFuncBody); // here the bodies are expressions, not Function objects;
+if(condition, trueFuncBody, falseFuncBody); // the bodies are expressions, not Function objects;
```
telephon: I know this was there before, I think this mixes up levels. 
The bodies are expressions that return function objects. They may be inlined, but this should not be observalble.

```
if(condition, a = trueFuncBody, b = falseFuncBody);
a.postln; // the trueFuncBody
```

HelpSource/Reference/Conditional-Execution.schelp
```diff
@@ -1,32 +1,57 @@
...
+For other control structures, see link::Reference/loop:: and link::Classes/Exception::.
 
 section:: Syntax of conditional expressions
 
-Control structures in SuperCollider are implemented as methods called on a receiver; 
-thus, code::if, while,:: etc. are technically emphasis::not:: keywords in SuperCollider but methods,
-even if the syntax can be made to emphasis::look:: as if they were keywords.
+Conditional Expressions in SuperCollider are implemented as strong::methods that are called on a receiver and which return a value::.
+This is distinct from many other languages where, e.g., code::if:: is a reserved identifier (keyword) independent of a particular receiver. 
+For instance, in C or in python, you cannot create a variable called code::if::, whereas in SuperCollider, you can (which does not mean that you should...).
```
telephon: In SuperCollider all methods are not reserved identifiers. The message `if` is also just a message to a receiver. If this is made clear in the beginning, a lot of other things follow. The whole thing with variable names is confusing, perhaps?

But it is good to clarify all this, anyway.

HelpSource/Reference/Conditional-Execution.schelp
JordanHendersonMusic: I agree with what is said, but I do think that `var x = 2;` and `var y;` are statements. It is just that it is impossible to write them in a situation where you'd depend on their return value so they appear like they return nil. Consider these three examples.

```
(
{
	var x;
	var y = 3;
	2
}.def.dumpByteCodes
)
BYTECODES: (2)
  0   65       PushSpecialValue 2
  1   F2       BlockReturn

(
{
	var x;
	var y = 3;
}.def.dumpByteCodes
)
BYTECODES: (2)
  0   6E       PushSpecialValue nil
  1   F2       BlockReturn
```
No bytecodes are emitted.

```
(
{
	var x = 2;
	var y = 3;
	x + y
}.def.dumpByteCodes
)

BYTECODES: (5)
  0   30		 PushTempZeroVar 'x'
  1   31		 PushTempZeroVar 'y'
  2   B0       TailCallReturnFromFunction
  3   E0       SendSpecialBinaryArithMsg '+'
  4   F2       BlockReturn


```
Only where they are used is the corresponding byte codes emitted. 

The bytecode `PushTempZeroVar` contains a 'drop' instruction. This means whatever is put on the stack is immediately removed from it, so it is like the return value doesn't exist, which I believe would be the lhs of `var x = n` or `nil` if no default is provided — `PushTempZeroVar` copies the default from the frame.

Anyway, it is probably best to leave out this detail from this document!

HelpSource/Reference/Conditional-Execution.schelp
HotwheelsSisyphus: Thanks for pointing that out, I do understand how the syntax works in this case but yes this could be clearer. I'll keep thinking about how to phrase this better. (I'm also concurrently drafting an overhaul of the syntax shortcuts file... )
It's also a bit a question of terminology. And this file may ultimately also not be the right place for a detailed explanation. 

The current syntax shortcuts file does _not_ include cases where the _receiver_ is a {}-enclosed block, it only talks about trailing arguments... and I've not found that notation explained elsewhere. We do need it for case and while here.

Of course the role of the receiver in the function notation `method(receiver)` is syntactically that of an argument... but the problem is that this doesn't mesh with the terminology implied by `receiver.method(arg1, arg2, ...)`; and the receiver notation _can_ combine with the trailing argument notation, as Jordan pointed out (hence the 2x2 table in the new version). By which I mean to say, I can't easily write them side by side e.g.
```
method { arg1_functionNotation } { arg2_functionNotation } ... etc 
// because that would become, reshuffled to receiver notation: 
arg1_functionNotation.method(arg2_functionNotation)
// which is at odds with what one would call the arguments
// on the basis of receiver notation:
recv.method { arg1_recvNotation } {arg2_recvNotation }
```
Hence that language about the "zeroth argument", and why I treat the "receiver block" as a special case.

The reason all this syntax stuff is in here at all is that the current Control Structures file lists multiple syntax variants for each method. I'm open to having less of that in here and putting it instead where it properly belongs, i.e. the syntax shortcuts file (or some other new syntax related help file).

HelpSource/Reference/Conditional-Execution.schelp
HotwheelsSisyphus: I think I should have made this clearer. This is meant to be a "don't do this" example... maybe there's the misunderstanding. 

In the current version, I put a disclaimer earlier on:

> NOTE: the distinguish already instantiated Function object from the expressions that define them (their "bodies"), the present file adopts the following convention: someFunc is equivalent to { someFuncBody } 

 This is probably an awkward solution on my end, but what I meant by that is that the bodies by themselves do _not_ return function objects. For instance, in 
 ```
if(true) { trueFuncBody } { falseFuncBody }
```
the bodies may be something like the Strings "aha" or "brb" ; so that expression will return one of those strings.
 whereas if the bodies were in fact returning Function objects, one would get something like this by replacement:
 
 ```
 if(true) { { "aha" } } { { "brb" } }
```
 which returns `a Function` .
 
 The distinction is necessary because of the trailing block syntax. Without trailing blocks, `if(true, { "aha"}, {"brb"})` returns a string, whereas `if(true) { {"aha"} } { { "brb } } )` returns a Function. So I can't just take `trueFunc` out of 
 ```
if(expr, trueFunc, falseFunc) 
// and rewrite as
if(expr) { trueFunc } { falseFunc }
// because the former outputs 
trueFunc.value
// whereas the latter outputs
trueFunc
```

Please do let me know if you can think of a less confusing solution. I also think my current explanation is suboptimal.

File: HelpSource/Reference/Conditional-Execution.schelp (hunk 12 above)
Comment (telephon) [reply HelpSource/Reference/Conditional-Execution.schelp]: yes, that is good – thank you for the details, very much appreciated! In the end, you can decide what should be said where for best understanding.

File: HelpSource/Reference/Conditional-Execution.schelp (hunk 13 above)
Comment (telephon) [reply HelpSource/Reference/Conditional-Execution.schelp]: Ah I see! It is much simpler, and has nothing to do with syntax. I misread your examples because "trueFunc" looks like a function body.

If the receiver is true, the first argument is evaluated (i.e. sent the message `.value`), if it is false, the second argument is evaluated.

Maybe a good example, because it is unusual:
```
r = Routine { 1.yield };
if(true, r); // -> 1, not "a Routine"
```

Because by default, objects return themselves when they get `value`, the string `"aha"` is returned.

Maybe this is best shown in the class definitions of booleans? THis is unusual for non-smalltalk programmers, so it could be shown (if it isn't already).
```
True {
    if { arg trueFunc, falseFunc; ^trueFunc.value }
}

False {
   if { arg trueFunc, falseFunc; ^falseFunc.value }
}
```

File: HelpSource/Reference/Conditional-Execution.schelp (hunk 14 above)
Comment (telephon) [reply HelpSource/Reference/Conditional-Execution.schelp]: P.S. maybe the best is to make all examples with code that can actually be executed, then there is less space for ambiguity.

File: HelpSource/Reference/Conditional-Execution.schelp (hunk 13 above)
Comment (telephon) [reply HelpSource/Reference/Conditional-Execution.schelp]: And you could just say: "To return an unevaluated function, e.g. { |x| x + 1 }; write it like this: if(true) { { |x| x + 1 } }"

File: HelpSource/Reference/Conditional-Execution.schelp (hunk 11 above)
Comment (jamshark70) [reply HelpSource/Reference/Conditional-Execution.schelp]: > I agree with what is said, but I do think that var x = 2; and var y; are statements.

Drifting off topic, but, to muddy those waters a bit:

```
(
f = {
	var x = 1;
	var y = 2 + 3;
};

f.def.dumpByteCodes;
f.def.prototypeFrame;
)

BYTECODES: (8)
  0   65       PushSpecialValue 2           <-- no '1'
  1   2C 03    PushInt 3
  3   E0       SendSpecialBinaryArithMsg '+'
  4   80 01    StoreTempVar 'y'
  6   6E       PushSpecialValue nil
  7   F2       BlockReturn
-> [1, nil]                                 <-- oh, *there* is the '1'
```

> The bytecode PushTempZeroVar contains a 'drop' instruction. 

Does it? If you do PushTempZeroVar x, and it drops, then PushTempZeroVar y and it drops, then `+` in your example would have nothing to operate on.

As noted, `var x = 1` goes into the prototypeFrame and doesn't emit byte codes. But the function has to have *something*, which the compiler enforces to be nil.

Comment (JordanHendersonMusic) [reply HelpSource/Reference/Conditional-Execution.schelp]: Ah yes! You are right about PushTempZeroVar not dropping, I think it is done in BlockReturn instead, I confused it with `StoreTempVar`.

I suppose you could think of `var ...` being an expression that returns nil, but the result of `var x = 1` on the stack is placing `1` there, so you'd expect it to return `1`. 

Interestingly in your example, `StoreTempVar` does drop as it consumes from the stack.

https://github.com/supercollider/supercollider/blob/9c55c8b3738ee3675ffcf493cd549b7da4d900bd/lang/LangSource/PyrInterpreter3.cpp#L1362

| code    | stack |
| -------- | ------- |
| PushSpecialValue 2  | [2]    |
| PushInt 3 | [2, 3]     |
| SendSpecialBinaryArithMsg '+'    | [5]    |
| StoreTempVar 'y' | [] |
| PushSpecialValue nil | [nil]


I would think of an expression as something the puts something on the stack that can be returned. I *think* `var` is the only thing that can't be returned as you'd expect it to return the value like the assignment operator.


```
(
f = {
	var x = 1;
	var y = 2 + 3;
	y
};

f.def.dumpByteCodes;
f.def.prototypeFrame;
)

BYTECODES: (8)
  0   65       PushSpecialValue 2
  1   2C 03    PushInt 3
  3   E0       SendSpecialBinaryArithMsg '+'
  4   80 01    StoreTempVar 'y'
  6   31		 PushTempZeroVar 'y'
  7   F2       BlockReturn
```
More proof I was wrong about PushTempZeroVar. 

If `var y= 2 + 3` was an expression, it wouldn't need to be pushed again.

**Comment by prko** [comment]
@HotwheelsSisyphus 
How about including the following?
https://scsynth.org/t/early-break-of-iterations/7928?u=prko

**Comment by HotwheelsSisyphus** [comment]
> @HotwheelsSisyphus How about including the following? https://scsynth.org/t/early-break-of-iterations/7928?u=prko

Yes, thanks for the suggestion! That will go in the iterations help file. (I don't quite understand how this works, but hopefully I'll figure it out)

**Comment by HotwheelsSisyphus** [comment]
Ok, so I let this lie for a few weeks to come back to it with fresh eyes. Seems mostly ok I think, even though probably just generally too wordy. I might remove the section about alternative syntaxes once I get done with the proper rewrite of the syntax shortcuts file. 

Pdf here, will push in a minute

[Conditional Execution Aug 20 | SuperCollider 3.14.0-dev Help.pdf](https://github.com/user-attachments/files/21927679/Conditional.Execution.Aug.20.SuperCollider.3.14.0-dev.Help.pdf)


Since this PR deletes a few help files (if.schep and control-structures.schelp), I wonder if we have a procedure for that? I can imagine someone wondering what happened to them, so I can imagine it would be good to keep a sort of dummy file around until the next minor version, some file that simply says where things have gone, for instance:

```
title::Control Structures
summary:: see conditional expressions, loop/repeat, Exception
categories:: Language
related:: Reference/Conditional-Execution, Reference/loop, Reference/Exception

As of September 2025, the Control Structures helpfile has been reorganized/merged into separate files:

list::
 ##link::Reference/Conditional-Execution::
 ##link::Reference/loop::
 ##link::Reference/Exception::
::

```

Similarly for if.schelp. Or is this excessively cautious?

**Comment by prko** [comment]
Thanks!
I think #6929 can be closed by this PR, and I think `if.schelp` can be safely removed.

**Comment by cdbzb** [comment]
> some file that simply says where things have gone

I think that would be considerate! I reach for the `control structures` help file quite often personally...

**Comment by prko** [comment]
>>some file that simply says where things have gone
>
>I think that would be considerate! 

I think this could be documented in `News in SuperCollider version 3.14.1`.

>I reach for the control structures help file quite often personally...

I do, too!

**Comment by HotwheelsSisyphus** [comment]
ok, not super familiar with pre-commit and whether that failing check can be ignored. The weird thing is also that the commit for which it fails (75b7b11) doesn't concern the file that pre-commit seems to complain about (that's in 7ac4edc).

Anyway: 
1.Added those courtesy files (if, loop/repeat, control sturctures)
2.Added the "Iteration" file, which right now is just the for/do parts of the control-structures file merged with the contents of the loop/repeat file. I have more edits ready for that, but I feel they might make the discussion here even more tedious, so I'm saving them for another PR. 
3. The more-or-less final version of conditional-execution.schelp which corresponds to the pdf I posted yesterday.

File: HelpSource/Reference/iteration.schelp [hunk 15]
```diff
@@ -0,0 +1,80 @@
...
+helpfiles.)
+
+There are a number of ways to implement iteration in SuperCollider.
+The most familiar ones will be the link::Classes/Number#-for:: and link::Classes/Collection#-do:: methods.
+The link::Clases/Function#-while:: method is discussed link::Reference/Conditional-Execution#while#here::.
+For Routines, Streams, and Patterns, special methods exist: link::Classes/Function#-loop:: and link::Classes/Object#-repeat::.
+
+table::
+	## strong::Thing you want to do:: ||  strong::method to use:: || strong:: returns::
+	## iterate over a collection  	|| link::#do:: 	|| code::loopFunc.value(x):: for each element code::x:: of collection ::
```
prko: The two colons at the end should be removed:

```
## iterate over a collection || link::#do:: || code::loopFunc.value(x):: for each element code::x:: of collection
```

prko: >`iterate over a collection`

`.do` is also applicable to integers, floats  etc. It is not simple to summarise shortly. The following is too long:

`iterate over a collection or a numeric value n, invoking the action with a 0-based index”`

HotwheelsSisyphus: Thanks! But this isn't meant to be a summary of all the things you _can_ do with `.do`, but rather a quick orientation as to how those methods (`.do` and `.for` in this case) are different from each other. The content from the control structures file further below explains the details.

HotwheelsSisyphus: Thanks, resolved!

HelpSource/Reference/iteration.schelp
```diff
@@ -0,0 +1,80 @@
+title:: Iteration
```
HotwheelsSisyphus, line 1: more drastically: I forgot to paste the actual `.for`, `.do` content!

HelpSource/Reference/iteration.schelp
HotwheelsSisyphus: But now I see that I forgot to paste that content! facepalm emoji

HelpSource/Reference/iteration.schelp
```diff
@@ -0,0 +1,211 @@
...
+summary:: for, do, loop, repeat
+categories:: Core,Common methods, Streams-Patterns-Events
+related::  Object, Function, Routine, Stream, Pattern
```
prko: Line 4 should be:
```
related::  Classes/Object, Classes/Function, Classes/Routine, Classes/Stream, Classes/Pattern
```

HelpSource/Reference/iteration.schelp
```diff
@@ -0,0 +1,211 @@
...
+
+(This page combines material formerly in the Control Structures and loop/repeat
+helpfiles.)
+
+There are a number of ways to implement iteration in SuperCollider.
+The most familiar ones will be the link::Classes/Number#-for:: and link::Classes/Collection#-do:: methods.
+The link::Clases/Function#-while:: method is discussed link::Reference/Conditional-Execution#while#here::.
```
prko: Line 11 should be
```
The link::Classes/Function#-while:: method is discussed link::Reference/Conditional-Execution#while#here::.
```

HotwheelsSisyphus:
Note: Apart from the header and the table at the beginning, everything in the iteration file is as it was before in the Control-Structures (for, do sections) and the loop / repeat file. Please do not review these sections here (no actual changes, just relocations), I have changes ready for them that I will put in another PR once this one is merged

prko:
> Seems mostly ok I think, even though probably just generally too wordy.

Thank you. I greatly appreciate the concise nature of the SC help documentation; however, I find that when brevity renders concepts less accessible to beginners, a more detailed explanation can be beneficial. Thank you again for your hard work.

HelpSource/Reference/iteration.schelp
HotwheelsSisyphus: done, thanks!

HotwheelsSisyphus:
thanks all for your patience with this! I boldly merged it for now... let's see how useful/annoying people find it.

PR #6970, opened 2025-06-11
QtCollider: fix Q_INVOKABLE SharedImage typename

elgiano:
## Purpose and Motivation

Fixes #6936 

We had this problem before with `TreeView`, see #6521. This PR applies the same fix. 

`SharedImage` is a typedef for `QSharedPointer<QtCollider::Image>`. Qt registers it as a MetaType with its full name `QSharedPointer<QtCollider::Image>`, so Q_INVOKABLE methods needs to use the same name, otherwise method calls fail with `QMetaMethod::invoke: cannot convert formal parameter 0 from QSharedPointer<QtCollider::Image> to SharedImage.`. Even if the type is the same, Qt gets confused about its name.

Here is a note from [Qt docs](https://doc.qt.io/qt-6/qtqml-cppintegration-exposecppattributes.html#exposing-properties):
> Note: Do not use typedef or using for [Q_PROPERTY](https://doc.qt.io/qt-6/qobject.html#Q_PROPERTY) types as these will confuse moc. This may make certain type comparisons fail.

For any other use than Q_INVOKABLE,  `SharedImage` should be fine, so we can keep the `typedef`.
I decided to patch also when used with Q_PROPERTY, since Qt mentions it explicitly, even though those methods didn't fail.

You can use this code to test affected Widgets:
```supercollider
var image = Image("PATH/TO/YOUR/IMAGE");
View(bounds:300@300).layout_(VLayout(
	View().setBackgroundImage(image),
	Button().icon_(image),
	Menu(MenuAction("test").icon_(image))
)).front
```

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [x] This PR is ready for review

PR #6971, opened 2025-06-11
Print stderr of sclang to IDE post window

capital-G:
## Purpose and Motivation

Currently `stderr` of sclang is not printed to the post window of the IDE.
But `stderr` can actually be very helpful in finding bugs in primitives (see https://github.com/supercollider/supercollider/pull/6970#issuecomment-2962593076).

This PR adds the printing of stderr to the post window

grafik

Before this PR the line `ERROR: QMetaMethod::invoke: cannot convert formal parameter 0 from QSharedPointer<QtCollider::Image> in call to QcCanvas::setBackgroundImage(QtCollider::SharedImage,QRectF,int,double)` was not displayed - it was displayed when sclang was started from within terminal, emacs, ... though.

`ERROR: ` is prefix in order to change the appearance of the line in the warning color of the theme.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/editors/sc-ide/core/sc_process.cpp
+++ b/editors/sc-ide/core/sc_process.cpp
@@ -49,10 +49,12 @@ ScProcess::ScProcess(Settings::Manager* settings, QObject* parent):
     mCompiled(false) {
     prepareActions(settings);
 
-    connect(this, SIGNAL(readyRead()), this, SLOT(onReadyRead()));
+    connect(this, SIGNAL(readyReadStandardOutput()), this, SLOT(onReadAllStandardOutput()));
+    connect(this, SIGNAL(readyReadStandardError()), this, SLOT(onReadAllStandardError()));
     connect(mIpcServer, SIGNAL(newConnection()), this, SLOT(onNewIpcConnection()));
     connect(this, SIGNAL(stateChanged(QProcess::ProcessState)), this,
             SLOT(onProcessStateChanged(QProcess::ProcessState)));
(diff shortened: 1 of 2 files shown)


elgiano:
Looks good to me, thanks! Shall we include this in 3.14, so to have more detailed bug reports before the next release?

dyfer:
I think this could be included in the 3.14.0-rc2. @capital-G can you rebase on 3.14?

Eventually we could restructure posting so that there's a way to e.g.differentiate sclang posting, sclang errors, child process (server) posting, and maybe even IDE posting, but for now this is a useful addition as is IMO.

capital-G:
After revisiting this I think I'd prefer to merge this into 3.15 - less git work and also less moving parts for rc2 as there is still enough work left for a proper 3.14 release where we probably want to reduce moving parts.

I'll have a quick vote in the dev-chat.

capital-G:
After a vote in the chat it seems we should go for 3.15

dyfer (approved):
I think this is 1) useful and 2) fairly straightforward in terms of implementation. 

I don't have great insights into the cpp side of this, but I think it's fine. Feel free to seek another review or merge :)

capital-G:
Thanks dyfer - I'll merge it in this case.

dyfer:
FYI I believe this causes git status messages to appear as "errors": https://github.com/supercollider/supercollider/issues/7602

Issue #6980, opened 2025-06-13
Examples Folder should be easy to reach


telephon:
## Motivation

In the first versions of supercollider, the examples were the central hub around which one would learn the language. I do see that nowadays we have repositories like sccode.org, but a well curated folder of examples is still worth more than a lot of documentation, in my view.

As a first step, the existing examples should be somehow part of the binary distribution, or easy to reach. Last time I wanted to show them to a student, I couldn't find them.

## Description of Proposed Feature

Have a menu item "open examples folder", perhaps? 

## Plan for Implementation

No plan yet.

prko:
I use the following example code in my tutorials:
```
(
[
    "aftergoeyvaerts.scd",
    "atheoryofharmony.scd",
    "autohausen.scd",
    "microhelix_study.scd",
    "some_constellations.scd",
    "Trains.scd"
].do { |filename|
    (
        Platform.resourceDir.replace("/SuperCollider.app/Contents/Resources", "")
            +/+ "examples/pieces" +/+ filename
    ).openDocument
}
)
```
From this, could the following method be implemented?
(Also, `Platform.case` could be used instead of `.replace`.)

```
Platform {
	...
	*exampleDir {
		^Platform.resourceDir.replace("/SuperCollider.app/Contents/Resources", "")
	}
	...
}
```
and
```
ExampleFiles {
	...
	*folder {
		^Platform.resourceDir.replace("/SuperCollider.app/Contents/Resources", "")
	}
	...
}
```
Then
```
ExampleFiles.folder.openOS

Platform.exampleDir.openOS
```

capital-G:
I think we could handle them like the HelpFiles in CMake and expose this directory via sclang.
It would be great if someone checked if the examples are working b/c I have a vague memory that some examples didn't worked which was not the nicest welcoming as a beginner :D If we recommend them as a resources we would need to test them for each release if they are working as expected.

We could also add a note that those examples are crafted in a way to show the diverse styles of people writing sclang b/c most examples from other frameworks use the examples to introduce a canonical style which arguable doesn't exist in SC (?).

On a more general note regarding the examples:
I think we should embed the examples into the documentation if they demonstrate some functionality in a "reusable" way (so they can be linked and searched) or they should reside on sccode.org if they tend to be more in the inspiring section.
[...]

PR #6981, opened 2025-06-13
Add examples folder path

prko:
## Purpose and Motivation
Implement the `examples` folder in `Platform` and `ExamplesFile` as suggested https://github.com/supercollider/supercollider/issues/6980#issuecomment-2971755922.
(The file menu item is not implemented. Sorry.)

With this PR, the following codes are supported:
```
ExampleFiles.folder.openOS
```
and
```
Platform.exampleDir.openOS
```

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


capital-G:
I think we would need CMake if we want to bundle the example files like help files. E.g. on my SC build I don't have any examples available - they are bundled within the dmg file on macOS though.

prko:
>I think we would need CMake if we want to bundle the example files like help files. E.g. on my SC build I don't have any examples available - they are bundled within the dmg file on macOS though.

Yes, on macOS, it is a bit different from other OS...
Please refer to https://github.com/supercollider/supercollider/issues/6980#issuecomment-2972616116

prko:
https://github.com/supercollider/supercollider/pull/6981/commits/d02bfe167bb90f3d82136021c00248e13f696e2d reflects #6986

SCClassLibrary/Common/Files/ExampleFiles.sc
```diff
@@ -10,4 +10,8 @@ ExampleFiles {
 	*child {
 		^Platform.resourceDir +/+ "sounds" +/+ "a11wlk01.wav";
 	}
+
+	*folder {
```
capital-G: This is  mixing unix (directory) and windows (folder) vocabulary - I think within sclang we tend more towards dir, see e.g. `Platform.resourceDir`).

SCClassLibrary/Platform/Platform.sc
```diff
@@ -262,6 +262,11 @@ Platform {
 
 		"#".dup(40).join.postln;
 	}
+
+	*exampleDir {
+		^(Platform.resourceDir +/+ "examples")
+	}
```
capital-G, line 268: IMO it would be sufficient to have one implementation via e.g. `Path.sclangExampleDir` (someone with more sclang experience should propose a good name or where to put this constant - e.g. I do not fully understand the motivation why `resourceDir` is located under `Platform` and not a static member of `Path`).

If we really want to have the same information in two places we should use the same implementation and not copy the implementation as this easily introduces bugs/inconsistencies.

HelpSource/Classes/ExampleFiles.schelp
```diff
@@ -59,3 +59,6 @@ TABLE::
 ::
 
 returns:: The full path of TELETYPE::a11wlk01.wav:: as a LINK::Classes/String::
+
+METHOD:: folder
+returns:: The full path of code::examples:: folder as a LINK::Classes/String::
```
capital-G: IMO it would be good to state that this dir is about `sclang`/`scd` examples as all other mentioned files here are sound files.

capital-G (commented):
I don't know if `ExampleFiles.*folder` is the best way to name this feature. Currently `ExampleFiles` is currently referring exclusively to sound files (which I think is good as it reduces ambiguity of its methods). I'd find it somehow counter-intuitive if the `directory` method returns a path which is not related to the other methods within this class.
IMO the sclang example files directory could be made available via something like `Platform.sclangExamplesDirectory`, but someone from the class lib team or help team should/needs to decide this.

This also breaks the test suite currently since the static methods of `ExmpleFiles` are expected to be files and need to be fixed before merging - but this is dependent on the decision above.

SCClassLibrary/Platform/Platform.sc
prko: >why resourceDir is located under Platform 

Perhaps... because the place of the resource directories is platform-dependent?

HelpSource/Classes/ExampleFiles.schelp
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6981/commits/4d8dd4f160dcf314364b324762f0fdefa6e50cc1.

SCClassLibrary/Common/Files/ExampleFiles.sc
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6981/commits/4d8dd4f160dcf314364b324762f0fdefa6e50cc1.

SCClassLibrary/Platform/Platform.sc
capital-G: Yes, but this is e.g. an implementation detail which should be of no concern to the user.

prko: I realised this in 2016. It confused and inconvenienced me.
- The location of the example folder differs between Windows and Mac OS X.
 
Also, it was a bit odd that the sounds folder is hidden on Mac OS X because it is inside the SuperCollider.app package folder, but I'm used to it now.

cdbzb: > Yes, but this is e.g. an implementation detail which should be of no concern to the user.

this is a good point! It might be nice to consolidate important paths in Path! *If* we do this generally maybe `Path.exampleFilesDir` would be good here so that typing "Path.example" will get you the correct autocomplete...

prko: Um... what about this?

Move the 'sounds' folder to the 'examples' folder and move all subfolders in the current 'examples' folder to the 'sclangCodes' folder in the `examples` folder.

- .../examples/sounds/
- .../examples/sclangCodes/

telephon: There is a misunderstanding in the name "ExampleFiles" here. The name, as I understand it, was given to mean "the files that the examples use". It doesn't mean "the files that contain examples".

We should keep that separate. If the supercollider examples are in a particular directory, they can be found via a path, and poossibly there can be a menu item that says "Open Examples Folder".

tedmoore:
> Currently `ExampleFiles` is currently referring exclusively to sound files

I think the decision was that we went with a filetype agnostic name to allow filetypes other than sound files to also exist in the example files folder.

SCClassLibrary/Platform/Platform.sc
tedmoore: To highlight the unfortunate naming situation we've found ourselves in, I agree that `ExampleFiles` should not point towards the "example files". Perhaps, we should start referring to the "files that contain examples" as "code examples". This way `ExampleFiles` can be any files used in examples and "code examples" can be the actual code examples.

> If the supercollider examples are in a particular directory, they can be found via a path, and poossibly there can be a menu item that says "Open Examples Folder".

I think this is an excellent suggestion.

capital-G: > and poossibly there can be a menu item that says "Open Examples Folder".

Should we put this under the `File` menu or the `Help` menu?

grafik

grafik

Either way, we could also add links to websites such as sccode.org and scsynth.org under Help.

If there is a consensus on where this should be put I am happy to provide the matching PR ;)

prko: I think under `File` menu, `beneath Open user support directory` would be better.

(If this folder could be included in the help home, it could also be an option, but I am not sure if it is a good idea.)

dyfer: I agree File menu is better than Help.

Additionally, it would be worthwhile I think to also add a link to the examples on the main Help page.

prko: >Additionally, it would be worthwhile I think to also add a link to the examples on the main Help page.

Thanks!
Done in #7082. Please leave your comments there!

PR #6985, opened 2025-06-14
Add sccode.org  class search results to the end of class documentation (before footnotes)

prko:
## Purpose and Motivation
To make it easier for users to see more examples of classes, I added the search results for class names on sccode.org to the end of the class documentation, before the footnotes. E.g:
Screenshot 2025-06-14 at 20 35 13

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


prko:
Ah, there is a small problem. Some help files do not have Example section at the end. Eg:
image

prko:
@HotwheelsSisyphus, @tedmoore 

What do you think about this PR?
Could you review this PR?

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -1040,6 +1040,15 @@ SCDocHTMLRenderer {
 
 		this.renderHeader(stream, doc, body);
 		this.renderChildren(stream, body);
+
+		if(doc.isClassDoc) {
+			var moreExamples = "https://sccode.org/search?q=" ++ doc.klass;
```
capital-G: You should rather search via

```suggestion
			var moreExamples = "https://sccode.org/tag/class/%".format(doc.klass);
```

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -1040,6 +1040,15 @@ SCDocHTMLRenderer {
...
+			stream << (
+				"</br>For community examples, please refer to &lt;<a href='"
+				+ moreExamples ++ "'>" ++ moreExamples ++ "</a>&gt;."
```
capital-G: what is the motivation to surround the link with `<` `>`?

capital-G (commented):
IMO text which is really similar across 1000+ of pages should be solved in a more generic manner - maybe a button or a plain link to sccode on the top of the class page? It would be interesting to check how other documentation tools solve this or what people w/ some UX XP could advise here?

edit: I really like the idea of linking to the community examples though - this is a great way to integrate sccode and make users more familiar with this site.

SCClassLibrary/SCDoc/SCDocRenderer.sc
prko: I thought this would make the web address stand out visually, like a small cosmetic change.
I'm not sure if it's really effective.

I can remove it if you don't like it.

SCClassLibrary/SCDoc/SCDocRenderer.sc
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6985/commits/5be310e958764dfda92ec52c33b16e232a0d43ad.

SCClassLibrary/SCDoc/SCDocRenderer.sc
prko: Done in https://github.com/supercollider/supercollider/pull/6985/commits/5be310e958764dfda92ec52c33b16e232a0d43ad!

capital-G: We don't do this elsewhere so I think it is rather confusing than helpful?

We could also make use of the HTML anchors to shorten it to something like

```
<a herf="https://sccode.org/search?q=SinOsc">Community examples</a>
```

prko: As follows?

>For community examples, please refer to [Community examples](https://sccode.org/search?q=SinOsc).

Perhaps the full text should be changed?

capital-G: I think a plain link called _Community examples_ would be sufficient ;)

prko: Done in https://github.com/supercollider/supercollider/pull/6985/commits/458c5b1fa09e7970552d2403ca11e1be1b228ff2.

Screenshot 2025-06-17 at 00 44 47

tedmoore:
> The examples folder in the distribution could get a careful overhaul and then represent well-curated examples. Then they could be linked in a closer way (they are also local to the machine).

I agree that rather than pointing to a collection of examples that may or may not be up-to-date or good we should focus on updating and curating the examples provided in the distribution.

SCClassLibrary/SCDoc/SCDocRenderer.sc
```diff
@@ -1043,10 +1043,7 @@ SCDocHTMLRenderer {
 
 		if(doc.isClassDoc) {
 			var moreExamples = "https://sccode.org/tag/class/%".format(doc.klass);
-			stream << (
-				"</br>For community examples, please refer to <a href='"
-				+ moreExamples ++ "'>" ++ moreExamples ++ "</a>."
-			)
+			stream << ("</br><a href='" + moreExamples ++ "'>" ++ moreExamples ++ "</a>")
```
capital-G: This is now lacking the context that these are community examples.
Also, a `br` can not be closed - I don't know if this is used elsewhere in the code base but this is invalid html syntax.

e.g.

```supercollider
var scCodeLink = "https://sccode.org/tag/class/%".format(doc.klass);
stream << ("<br><a href=\"%\">Community examples</a>".format(scCodeLink));
```

would look most clean to me. But please wait for the discussion to resolve before adjusting the implementation ;)

prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/6985/commits/817cf980c2af6684169419d78b8bf129d47d9570.

>But please wait for the discussion to resolve before adjusting the implementation ;)

YES!

prko:
There are differing views regarding this pull request. 

Based on my tests:
- Several classes do have associated examples on sccode.org; however, many of them appear to follow outdated coding styles.
- Additionally, clicking on the search results leads to an issue:
SC Help Browser cannot open these documents, as it does not support tabs, while the documents themselves are designed to launch in a new tab.

Although including search results from sccode.org initially seems like a worthwhile idea, it proves to be of limited usefulness due to these issues in practice.

On balance, I believe this pull request could be closed.
What do other developers think about it?

PR #6993, opened 2025-06-16
Help: update guides to reflect kwargs.

JordanHendersonMusic:
## Purpose and Motivation

Adds a guide explaining object prototyping using IdentityDictionary.

Expands the documentation for IdentityDictionary and Functions to include information about its usage in object prototyping and variable arguments.

Clarifies class writing documentation, including newCopyArgs and init methods.

## Types of changes

- Documentation

## To-do list

### DIFF
--- a/HelpSource/Classes/Function.schelp
+++ b/HelpSource/Classes/Function.schelp
@@ -3,7 +3,8 @@ summary::Implements a function
 categories::Core>Kernel
 related::Reference/Functions, Classes/AbstractFunction, Classes/FunctionDef
 
-This document deals with the usage of function objects, for a more general introduction and documentation of syntax, see link::Reference/Functions::.
+This document deals with the usage of Function objects. 
+For a more general introduction and documentation of syntax, see link::Reference/Functions::.
 
 description::
 A Function is a reference to a FunctionDef and its defining context Frame. When a FunctionDef is encountered in your code it is pushed on the stack as a Function. A Function can be evaluated by using the 'value' method. See the Functions help file for a basic introduction.
--- a/HelpSource/Classes/Function.schelp
+++ b/HelpSource/Classes/Function.schelp
@@ -3,6 +3,8 @@ summary::Implements a function
 categories::Core>Kernel
 related::Reference/Functions, Classes/AbstractFunction, Classes/FunctionDef
 
+This document deals with the usage of function objects, for a more general introduction and documentation of syntax, see link::Reference/Functions::.
+
 description::
 A Function is a reference to a FunctionDef and its defining context Frame. When a FunctionDef is encountered in your code it is pushed on the stack as a Function. A Function can be evaluated by using the 'value' method. See the Functions help file for a basic introduction.
 
--- a/HelpSource/Classes/IdentityDictionary.schelp
+++ b/HelpSource/Classes/IdentityDictionary.schelp
@@ -4,7 +4,8 @@ related::Classes/Environment, Classes/Event, Guides/ObjectPrototyping
 categories::Collections>Unordered
 
 DESCRIPTION::
-An IdentityDictionary is an associative collection mapping keys to values. Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
+An IdentityDictionary is an associative collection mapping keys to values. 
+Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
 
 An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::), which behave like objects whose methods can be programmed at runtime.
 
--- a/HelpSource/Classes/IdentityDictionary.schelp
+++ b/HelpSource/Classes/IdentityDictionary.schelp
@@ -8,8 +8,6 @@ An IdentityDictionary is an associative collection mapping keys to values. Keys 
 
 An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::), which behave like objects whose methods can be programmed at runtime.
 
-
-
 Often, the subclass link::Classes/Event:: is used as an IdentityDictionary, because there is a syntactical shortcut:
 code::
 a = (); // return a new Event.
@@ -22,7 +20,6 @@ Keys match only if they are strong::identical objects::. (i.e. === returns true.
 
 The contents of a Dictionary are strong::unordered::. You must not depend on the order of items in a Dictionary.
 ::
-
 
 IdentityDictionary is often used to assign names to instances of a particular class. For example, the proxy classes (link::Classes/Pdef::, link::Classes/Pdefn::, link::Classes/Tdef::, link::Classes/Ndef::), and link::Classes/NodeWatcher:: all have class variables named all implemented as IdentityDictionaries.
 
@@ -72,7 +69,6 @@ e[\b]
 -> 20
 ::
 
-
 CLASSMETHODS::
 
 method::new
@@ -84,7 +80,6 @@ a = IdentityDictionary(know: true);
 a.put(\foo, { |x, y| "--".postln; ("x:" ++ x).postln;  ("y:" ++ y).postln; y.squared });
 a.foo(-10.01);
 ::
-
 
 warning::
 Only keys that are not already instance methods of IdentityDictionary (or its superclasses) can be used in such a way. E.g. the key "free" will not work, because it is implemented in Object. This means that class extensions (see: link::Guides/WritingClasses::) can break code. It is advisable to use underscores in method names to make this improbable.
@@ -100,7 +95,6 @@ a.fuzzy_plus(7, 4);
 ::
 
 INSTANCEMETHODS::
-
 
 method::know
 If set to true, the dictionary interprets method calls as look ups. This allows you to implement object prototypes (see above).
@@ -159,7 +153,6 @@ a.cs;
 
 image::IdentityDictionary_01.png#Example schema: order of overriding in proto and parent.::
 
-
 method::insertParent
 Inserts a dictionary into the chain of parents of the receiver (rather than replacing the parent).
 argument::newParent
@@ -168,7 +161,6 @@ argument::insertionDepth
 Level at which the new parent is inserted. Zero (default) means directly above, Inf means at the top of the parent chain.
 argument::reverseInsertionDepth
 If the new parent dictionary has parents itself, this parameter specifies where the original parents are placed in the new parent chain. Zero means directly above, Inf (default) means at the top of the chain.
-
 
 image::IdentityDictionary_03.png#Compare a.insertParent(b, 0) and a.insertParent(b, 1)::
 
--- a/HelpSource/Classes/IdentityDictionary.schelp
+++ b/HelpSource/Classes/IdentityDictionary.schelp
@@ -6,7 +6,7 @@ categories::Collections>Unordered
 DESCRIPTION::
 An IdentityDictionary is an associative collection mapping keys to values. Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
 
-An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::). They behave like objects whose methods can be programmed at runtime.
+An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::), which behave like objects whose methods can be programmed at runtime.
 
 
 
--- a/HelpSource/Classes/IdentityDictionary.schelp
+++ b/HelpSource/Classes/IdentityDictionary.schelp
@@ -6,7 +6,8 @@ categories::Collections>Unordered
 DESCRIPTION::
 An IdentityDictionary is an associative collection mapping keys to values. Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
 
-IdentityDictionary's can be used to implement a sort-of 'runtime' editable class, see link::Guides/ObjectPrototyping::.
+An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::). They behave like objects whose methods can be programmed at runtime.
+
 
 
 Often, the subclass link::Classes/Event:: is used as an IdentityDictionary, because there is a syntactical shortcut:
--- a/HelpSource/Classes/IdentityDictionary.schelp
+++ b/HelpSource/Classes/IdentityDictionary.schelp
@@ -1,10 +1,13 @@
 CLASS::IdentityDictionary
 summary::associative collection mapping keys to values
-related::Classes/Environment, Classes/Event
+related::Classes/Environment, Classes/Event, Guides/ObjectPrototyping
 categories::Collections>Unordered
 
 DESCRIPTION::
 An IdentityDictionary is an associative collection mapping keys to values. Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
+
+IdentityDictionary's can be used to implement a sort-of 'runtime' editable class, see link::Guides/ObjectPrototyping::.
+
 
 Often, the subclass link::Classes/Event:: is used as an IdentityDictionary, because there is a syntactical shortcut:
 code::
--- a/HelpSource/Guides/ObjectPrototyping.schelp
+++ b/HelpSource/Guides/ObjectPrototyping.schelp
@@ -3,13 +3,11 @@ categories:: Guides
 summary:: Writing runtime 'classes'
 related:: Classes/IdentityDictionary
 
-
 You cannot create nor edit a class without recompiling the language... but what if you need that?
 
 Object prototyping is a middle ground that lets you build class–like structures at runtime.
 It works by taking advantage of the class link::Classes/IdentityDictionary:: (often the subclass link::Classes/Event:: is used) and the special method link::Classes/Object#-doesNotUnderstand::.
 While there are some oddities with this approach, it is extremely powerful.
-
 
 section:: Instance Variables (Members)
 
@@ -33,7 +31,7 @@ a.meow // 10
 
 section:: Methods
 
-By assigning a function to a key, you can call it just as a normal instance method would be called.
+By assigning a function to a key, you can call that function as if it were a normal instance method.
 
 However, the link::Classes/IdentityDictionary:: is passed into the function as the first argument.
 Usually, we call this argument code::self:: as it performs a similar role to the keyword code::this::.
@@ -73,7 +71,6 @@ dog.speak("quietly").postln; // I say "woof" quietly.
 )
 ::
 
-
 section:: Private Instance Variables (Members)
 
 These can be implemented inside the constructor function as normal variables, however, you don't add them to the dictionary.
@@ -95,7 +92,6 @@ p1.assign_secret("I'm a dog");
 p1.speak_secrets().postln; // My secret is "I'm a dog".
 )
 ::
-
 
 section:: Dangers
 
--- a/HelpSource/Guides/ObjectPrototyping.schelp
+++ b/HelpSource/Guides/ObjectPrototyping.schelp
@@ -6,8 +6,8 @@ related:: Classes/IdentityDictionary
 
 You cannot create nor edit a class without recompiling the language... but what if you need that?
 
-Object prototyping is a middle ground that lets you build class like structures at runtime.
-It works by taking advantage of the class link::Classes/IdentityDictionary:: (or more generally its child link::Classes/Event::) and the special method link::Classes/Object#-doesNotUnderstand::.
+Object prototyping is a middle ground that lets you build class–like structures at runtime.
+It works by taking advantage of the class link::Classes/IdentityDictionary:: (often the subclass link::Classes/Event:: is used) and the special method link::Classes/Object#-doesNotUnderstand::.
 While there are some oddities with this approach, it is extremely powerful.
 
 
@@ -69,7 +69,7 @@ var cat = makeSpeaker.("meow");
 var dog = makeSpeaker.("woof");
 
 cat.speak("loudly").postln; // I say "meow" loudly.
-dog.speak("quitely").postln; // I say "woof" quitely.
+dog.speak("quietly").postln; // I say "woof" quietly.
 )
 ::
 
@@ -106,7 +106,7 @@ Essentially, what makes link::Classes/IdentityDictionary:: powerful, also makes 
 The only way to avoid this mistake is by being vigilant when adding keys to an object prototype.
 
 The second drawback is that you cannot have a key with the same name as a method in link::Classes/Event:: or any parent class,
-otherwise the real class's method will be called, not the pseudo method you have written.
+otherwise the real class's method will be called, not the pseudo–method you have written.
 
 code::
 a = (numChannels: 10)
@@ -116,9 +116,9 @@ a.numChannels // 1
 As seen above, one expects the value code::10:: to be returned, but instead link::Classes/Object#-numChannels:: is called and the value code::1:: returned.
 
 There are two ways to mitigate this.
-One, use longer keys or names in snake case rather than pascal case: that is names_written_like_this, ratherThanNamesWrittenLikeThis, as SuperCollider only uses the latter.
+strong::One::, use longer keys or names in snake case rather than pascal case: that is names_written_like_this, ratherThanNamesWrittenLikeThis, as SuperCollider only uses the latter.
 
-Two, assign outside of the brackets, where a warning will be generated.
+strong::Two::, assign outside of the brackets, where a warning will be generated.
 
 code::
 a = ();
--- a/HelpSource/Guides/ObjectPrototyping.schelp
+++ b/HelpSource/Guides/ObjectPrototyping.schelp
@@ -11,9 +11,9 @@ It works by taking advantage of the class link::Classes/IdentityDictionary:: (or
 While there are some oddities with this approach, it is extremely powerful.
 
 
-section:: Members
+section:: Instance Variables (Members)
 
-Instance member–like things can be created by just adding keys to an link::Classes/IdentityDictionary::
+Instance variable–like things can be created by just adding keys to an link::Classes/IdentityDictionary::
 
 code::
 a = (meow: 10)
@@ -39,22 +39,22 @@ However, the link::Classes/IdentityDictionary:: is passed into the function as t
 Usually, we call this argument code::self:: as it performs a similar role to the keyword code::this::.
 
 code::
-a = ( speak: { |self, adverb| "I say \"%\" %.".format(self.word, adverb) } );
+a = ( speak: { |self, adverb| "I say '%' %.".format(self.word, adverb) } );
 a.word = "meow";
-a.speak(adverb: "loudly") // I say "meow" loudly.
+a.speak(adverb: "loudly") // I say 'meow' loudly.
 ::
 
 link::Reference/Functions#Variable Arguments:: also work.
 
 code::
-a = ( speak: {|self... words, kwargs| "I say \"%\" %.".format(words, kwargs.asDict[\adverb]) } );
+a = ( speak: {|self... words, kwargs| "I say '%' %.".format(words, kwargs.asDict[\adverb]) } );
 
 a.speak("meow", "woof", adverb: "loudly") // I say "[meow, woof]" loudly.
 ::
 
 section:: Constructors
 
-There is no way to mimic a constructor, but this isn't a problem, as we can simply make a function that returns the link::Classes/IdentityDictionary::.
+There is no direct way to mimic a constructor, but this isn't a problem, as we can simply make a function that returns the link::Classes/IdentityDictionary::.
 
 code::
 (
@@ -74,7 +74,7 @@ dog.speak("quitely").postln; // I say "woof" quitely.
 ::
 
 
-section:: Private Members
+section:: Private Instance Variables (Members)
 
 These can be implemented inside the constructor function as normal variables, however, you don't add them to the dictionary.
 
@@ -101,12 +101,12 @@ section:: Dangers
 
 There are two drawbacks over classes.
 
-The first, is that any key, including 'methods', can be overriden at any time.
+The first, is that any key, including 'methods', can be overridden at any time.
 Essentially, what makes link::Classes/IdentityDictionary:: powerful, also makes it fragile and easy to break.
 The only way to avoid this mistake is by being vigilant when adding keys to an object prototype.
 
 The second drawback is that you cannot have a key with the same name as a method in link::Classes/Event:: or any parent class,
-otherwise the real class's method will be called, not the psuedo method you have written.
+otherwise the real class's method will be called, not the pseudo method you have written.
 
 code::
 a = (numChannels: 10)
@@ -115,7 +115,15 @@ a.numChannels // 1
 
 As seen above, one expects the value code::10:: to be returned, but instead link::Classes/Object#-numChannels:: is called and the value code::1:: returned.
 
-The best way to mitigate this issue is by using longer keys, or using names in snake case rather than pascal case: that is names_written_like_this, ratherThanNamesWrittenLikeThis, as SuperCollider only uses the latter.
+There are two ways to mitigate this.
+One, use longer keys or names in snake case rather than pascal case: that is names_written_like_this, ratherThanNamesWrittenLikeThis, as SuperCollider only uses the latter.
+
+Two, assign outside of the brackets, where a warning will be generated.
+
+code::
+a = ();
+a.numChannels = 30; // WARNING: ...
+::
 
 section:: Inheritance
 
--- a/HelpSource/Guides/ObjectPrototyping.schelp
+++ b/HelpSource/Guides/ObjectPrototyping.schelp
@@ -1,0 +1,134 @@
+title:: Object Prototyping
+categories:: Guides
+summary:: Writing runtime 'classes'
+related:: Classes/IdentityDictionary
+
+
+You cannot create nor edit a class without recompiling the language... but what if you need that?
+
+Object prototyping is a middle ground that lets you build class like structures at runtime.
+It works by taking advantage of the class link::Classes/IdentityDictionary:: (or more generally its child link::Classes/Event::) and the special method link::Classes/Object#-doesNotUnderstand::.
+While there are some oddities with this approach, it is extremely powerful.
+
+
+section:: Members
+
+Instance member–like things can be created by just adding keys to an link::Classes/IdentityDictionary::
+
+code::
+a = (meow: 10)
+a.meow // 10
+a.meow = 20;
+a.meow // 20;
+::
+
+Unlike normal classes, you can create new instance variables at runtime.
+
+code::
+a = (meow: 10)
+a.woof = 4;
+a.woof // 4
+a.meow // 10
+::
+
+section:: Methods
+
+By assigning a function to a key, you can call it just as a normal instance method would be called.
+
+However, the link::Classes/IdentityDictionary:: is passed into the function as the first argument.
+Usually, we call this argument code::self:: as it performs a similar role to the keyword code::this::.
+
+code::
+a = ( speak: { |self, adverb| "I say \"%\" %.".format(self.word, adverb) } );
+a.word = "meow";
+a.speak(adverb: "loudly") // I say "meow" loudly.
+::
+
+link::Reference/Functions#Variable Arguments:: also work.
+
+code::
+a = ( speak: {|self... words, kwargs| "I say \"%\" %.".format(words, kwargs.asDict[\adverb]) } );
+
+a.speak("meow", "woof", adverb: "loudly") // I say "[meow, woof]" loudly.
+::
+
+section:: Constructors
+
+There is no way to mimic a constructor, but this isn't a problem, as we can simply make a function that returns the link::Classes/IdentityDictionary::.
+
+code::
+(
+var makeSpeaker = { |word|
+	(
+		word: word,
+		speak: { |self, adverb| "I say \"%\" %.".format(self.word, adverb) }
+	)
+};
+
+var cat = makeSpeaker.("meow");
+var dog = makeSpeaker.("woof");
+
+cat.speak("loudly").postln; // I say "meow" loudly.
+dog.speak("quitely").postln; // I say "woof" quitely.
+)
+::
+
+
+section:: Private Members
+
+These can be implemented inside the constructor function as normal variables, however, you don't add them to the dictionary.
+
+code::
+(
+var makeSpeaker = {
+	var secretWords = "I'm a cat";
+	(
+		assign_secret: { |self, secret| secretWords = secret },
+		speak_secrets: { |self| "My secret is \"%\".".format(secretWords.value) }
+	)
+};
+
+var p1 = makeSpeaker.();
+
+p1.speak_secrets().postln; // My secret is "I'm a cat".
+p1.assign_secret("I'm a dog");
+p1.speak_secrets().postln; // My secret is "I'm a dog".
+)
+::
+
+
+section:: Dangers
+
+There are two drawbacks over classes.
+
+The first, is that any key, including 'methods', can be overriden at any time.
+Essentially, what makes link::Classes/IdentityDictionary:: powerful, also makes it fragile and easy to break.
+The only way to avoid this mistake is by being vigilant when adding keys to an object prototype.
+
+The second drawback is that you cannot have a key with the same name as a method in link::Classes/Event:: or any parent class,
+otherwise the real class's method will be called, not the psuedo method you have written.
+
+code::
+a = (numChannels: 10)
+a.numChannels // 1
+::
+
+As seen above, one expects the value code::10:: to be returned, but instead link::Classes/Object#-numChannels:: is called and the value code::1:: returned.
+
+The best way to mitigate this issue is by using longer keys, or using names in snake case rather than pascal case: that is names_written_like_this, ratherThanNamesWrittenLikeThis, as SuperCollider only uses the latter.
+
+section:: Inheritance
+
+This is implemented with link::Classes/IdentityDictionary#-'parent' and 'proto' variables::.
+Technically, this can be used to do more than inheritance, but only inheritance is shown here.
+
+This is done by assigning the code::parent:: to another dictionary to lookup keys in.
+
+code::
+a = (word: "meow" );
+b = (speak: {|self| "I say \"%\".".format(self.word) } );
+b.parent = a; // assign 'a' as a parent.
+b.speak // I say "meow".
+::
+
+This is similar to how Javascript and other prototyping languages work.
--- a/HelpSource/Guides/WritingClasses.schelp
+++ b/HelpSource/Guides/WritingClasses.schelp
@@ -3,7 +3,6 @@ summary:: Writing SuperCollider Classes
 categories:: Language>OOP
 related:: Reference/Classes, Classes/Class, Classes/Object
 
-
 SuperCollider follows a pure object-oriented paradigm. It is not built on data types, but on objects, which respond to messages. A class is an object that holds information about how its objects (instances) respond to such messages. Writing a class gives a definition of this behavior.
 
 This is an overview of idioms used in writing classes. It is not a tutorial on writing a system of interrelated classes. It gives an overview of some typical expressions. See also: link::Guides/Intro-to-Objects::, link::Reference/Messages::, and link::Reference/Classes::.
@@ -22,7 +21,6 @@ Platform.systemExtensionDir; // Extensions available to all users on the machine
 
 It is not possible to enter a class definition into an interpreter window and execute it.
 ::
-
 
 section:: Inheriting
 
@@ -43,8 +41,6 @@ MyClass { // : Object is implied
 ::
 
 This is why in the above example, the message code::new:: can be called without being explicitly defined in the class.
-
-
 
 section:: Methods
 
@@ -72,7 +68,6 @@ a = MyClass.new // returns a new instance
 a.instanceMethod // posts "hello instance"
 ::
 
-
 To return from the method use teletype::^:: (caret). Multiple exit points also possible. If no teletype::^:: is
 specified, the method will return the instance (and in the case of Class methods, will return the class). There is no
 such thing as returning void in SuperCollider.
@@ -94,12 +89,9 @@ MyClass {
 }
 ::
 
-
 subsection:: Class Methods
 
-
 An object's  strong::class methods:: are defined alongside its instance methods. They are specified with an asterisk (teletype::*::) before the method name.
-
 
 A link::Classes/Class:: is itself an object. It is what all instances of it have in common. Class methods are the instance methods of the object's class. That's why within class methods, the keyword code::this:: refers to the class.
 
@@ -121,7 +113,6 @@ This could then be used as follows:
 code::
 MyClass.classMethod // posts "hello class"
 ::
-
 
 subsection::Overriding methods (overloading)
 
@@ -156,8 +147,6 @@ MyClass : SomeSuperclass {
 	}
 }
 ::
-
-
 
 section:: Instances
 
@@ -199,7 +188,6 @@ Foo {
 If initialisation logic that prepares the arguments before assignment is required,
 ideally, it should go before the call to code::super.newCopyArgs::.
 This way the final object never exists in an invalid state — an alternative is the use of an code::init:: method and shall be shown shortly.
-
 
 However, when there is a hierarchy of classes, code::newCopyArgs:: requires that all the base class arguments be passed in the order they are defined. 
 This makes changing the base class difficult.
@@ -251,7 +239,6 @@ Another benefit is that the instance is never in an incomplete state: either all
 The disadvantages come when you have complex initialisation logic where the arguments in code::Derived:: require knowledge of the arguments in code::Base::.
 Traditionally in SuperCollider this has been solved with an code::init:: method as seen below, although a 'builder' pattern might be considered for complex cases.
 
-
 code::
 Base {
 	var a, b;
@@ -298,10 +285,7 @@ While the code::init:: pattern is the most flexible, care must be taken as insid
 the instance is in an incomplete state and other methods might not work as expected;
 for example, this may become confusing when combined with threads and asynchronous execution, for example, when interacting with the server.
 
-
-
 section:: Class instances
-
 
 Class variables are accessible within class methods and in any instance methods.
 
@@ -327,9 +311,6 @@ MyClass {
 }
 ::
 
-
-
-
 section::Alternatives to inheritance
 
 Overreliance on inheritance is usually a design flaw. Inheritance is mainly a way to organise code, and shouldn't be mistaken for a categorisation of objects. Two objects may respond to a message in different ways (polymorphism), and objects delegate control to ther objects they hold in their instance variables (object composition).
@@ -351,7 +332,6 @@ MyClass {
 a = MyClass.new;
 a.value; // returns 1
 ::
-
 
 subsection::Object Composition
 Often, an object passes control to one of the objects it has in its instance variables. Because these objects can be of any kind, this is a very flexible way to achieve a wide range of functionalities. For example, a link::Classes/Button:: has an code::action:: instance variable, which may hold anything that responds to the message code::value::.
@@ -376,7 +356,6 @@ b.value(8);
 
 Often, variables like code::action:: above are filled with custom objects that belong to code::MyClass::. Thus, one will write many small classes that can be well combined in such a way. This is called "pluggable behavior".
 
-
 section::Variables
 
 subsection:: Initializing variables directly
@@ -390,7 +369,6 @@ MyClass {
 	var y = #[1, 2, 3];
 }
 ::
-
 
 subsection:: Variable Scope
 
@@ -426,7 +404,6 @@ MyClass : SomeSuperclass {
 // SomeSuperclass.returnXHere returns 0
 // MyClass.returnXHere returns 1
 // MyClass.returnX returns 0
-
 
 ::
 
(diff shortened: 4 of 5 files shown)


HelpSource/Classes/IdentityDictionary.schelp
```diff
@@ -1,11 +1,14 @@
 CLASS::IdentityDictionary
 summary::associative collection mapping keys to values
-related::Classes/Environment, Classes/Event
+related::Classes/Environment, Classes/Event, Guides/ObjectPrototyping
 categories::Collections>Unordered
 
 DESCRIPTION::
 An IdentityDictionary is an associative collection mapping keys to values. Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
 
+IdentityDictionary's can be used to implement a sort-of 'runtime' editable class, see link::Guides/ObjectPrototyping::.
```
telephon: suggestion:
An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::). They behave like objects whose methods can be programmed at runtime.

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,134 @@
...
+
+
+You cannot create nor edit a class without recompiling the language... but what if you need that?
+
+Object prototyping is a middle ground that lets you build class like structures at runtime.
+It works by taking advantage of the class link::Classes/IdentityDictionary:: (or more generally its child link::Classes/Event::) and the special method link::Classes/Object#-doesNotUnderstand::.
+While there are some oddities with this approach, it is extremely powerful.
+
+
+section:: Members
```
telephon: In sclang / smalltalk, members are called intance variables, so you could write: section:: Instance Variables (Members)
or the other way round.

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,134 @@
...
+::
+
+Unlike normal classes, you can create new instance variables at runtime.
+
+code::
+a = (meow: 10)
+a.woof = 4;
+a.woof // 4
+a.meow // 10
+::
```
telephon, line 30: you could show here already how to make a new instance, by (deep) copying.
`a.deepCopy.`

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,134 @@
...
+
+section:: Methods
+
+By assigning a function to a key, you can call it just as a normal instance method would be called.
+
+However, the link::Classes/IdentityDictionary:: is passed into the function as the first argument.
+Usually, we call this argument code::self:: as it performs a similar role to the keyword code::this::.
+
+code::
+a = ( speak: { |self, adverb| "I say \"%\" %.".format(self.word, adverb) } );
```
telephon: you could write: `"I say '%' %."`

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,134 @@
...
+link::Reference/Functions#Variable Arguments:: also work.
+
+code::
+a = ( speak: {|self... words, kwargs| "I say \"%\" %.".format(words, kwargs.asDict[\adverb]) } );
+
+a.speak("meow", "woof", adverb: "loudly") // I say "[meow, woof]" loudly.
+::
+
+section:: Constructors
+
```
telephon, line 54: there is no reason why this constructor shouldn't live in a prototype. So maybe
```
q = ();
q.makeSpeaker = { |self, word|
	(
		word: word,
		speak: { |self, adverb| "I say \"%\" %.".format(self.word, adverb) }
	)
};
a = q.makeSpeaker("meow"); // etc.
```

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,134 @@
...
+var cat = makeSpeaker.("meow");
+var dog = makeSpeaker.("woof");
+
+cat.speak("loudly").postln; // I say "meow" loudly.
+dog.speak("quitely").postln; // I say "woof" quitely.
+)
+::
+
+
+section:: Private Members
```
telephon: Private instance variables (members)

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,134 @@
...
+otherwise the real class's method will be called, not the psuedo method you have written.
+
+code::
+a = (numChannels: 10)
+a.numChannels // 1
+::
+
+As seen above, one expects the value code::10:: to be returned, but instead link::Classes/Object#-numChannels:: is called and the value code::1:: returned.
+
+The best way to mitigate this issue is by using longer keys, or using names in snake case rather than pascal case: that is names_written_like_this, ratherThanNamesWrittenLikeThis, as SuperCollider only uses the latter.
```
telephon: I think the best way to migitate this is to assign them from the outside:
```
a = ();
a.numChannels = 30; // warns you
```

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+superclass, which in turn calls its superclass, up until code::Object.new:: is called and an object is actually created and its memory allocated.
 
 code::
 MyClass {
+	// This is a normal constructor method.
+	*new { ^super.new }
+}
+::
+
+In this case note that code::super.new:: calls the method code::new:: on the superclass and returned a new object.
```
telephon: returns

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+
+Instance can have members (instance variables), where each instance of the class can have a different value for each member.
+code::
+Foo {
+	var a, b, c; // Instance variables.
+}
+::
+
+These can be assigned in two ways.
+Through code::newCopyArgs:: through an code::init:: method.
```
telephon: reformat this …

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+	*new { |myC| ^super.newCopyArgs(a: 10, b: 20, c: myC) }
+}
+::
+
+If initialisation logic that prepares the arguments before assignment is required,
+ideally, it should go before the call to code::super.newCopyArgs::.
+This way the final object never exists in an invalid state — an alternative is the use of an code::init:: method and shall be shown shortly.
+
+
+However, when there is a heirachy of classes code::newCopyArgs:: requires all the base class arguments be passed in the order they are defined in, this makes changing the bass class difficult.
```
telephon: However, when there is a hierachy of classes, code::newCopyArgs:: requires all the base class arguments be passed …

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+Here is one way to mitigate this issue using variables keyword arguments link::Reference/Functions#Variable Arguments::.
```
telephon: The message code::superPerformArgs:: is a way to mitigate

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
 
-Warning:: If the superclass also happened to call super.new.init it will have expected to call the .init method defined
-in that class (the superclass), but instead the message .init will find the implementation of the class that the object
-actually is, which is our new subclass. In such cases, a unique method name like myclassInit should be used.
 ::
 
-One easy way to copy the arguments passed to the instance variables when creating a class is to use link::Classes/Object#*newCopyArgs::.  This
-method will copy the arguments to the instance variables in the order that the variables were defined in the class,
-starting from the parent classes and working it's way down to the current class.
+Although this approach might seem complex, the benifit is that you can add new members to either class and it will always be clear which member you meant to assign to.
```
telephon: the benefit is that new instance variables can be added to either class and it will always be clear which one you meant to assign to.

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+If it is desirable to be able to remove instance members in the future,
+it is advised to require only keyword arguments be used by removing the named arguments and throwing an error if code::args:: is not empty.
+
```
telephon, line 222: Should it be possible to remove instance members in the future,
it is good to require only keyword arguments be used. This can be achieved by removing the named arguments and throwing an error if code::args:: is not empty.

telephon: an example may be good!

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+Another benifit is that the instance is never in an incomplete state, either all the instance members have been assigned to, or the object does not exist.
```
telephon: benefit

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+
+The disadvantage come when you have complex initialisation logic where the arguments in code::Derived:: require knowledge of the arguments in code::Base::.
```
telephon: The disadvantages come

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+	*new {|a, b| ^super.new.init(a, b) }
+	init {|newA, newB|
+		// Initialsation logic goes here.
+		a = newA;
+		b = newB;
 	}
 }
+::
+
+This work but becomes more complex in inheritance chains.
```
telephon: works

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+		// do initialisation with all the arguments
+		this.init(newA, newB) // call Base.init
+		c = newC; // assign to Derivied's c
 	}
 }
 ::
 
+Warning:: If the superclass also happened to call super.new.init it will have expected to call the .init method defined
+in that class (the superclass), but instead the message .init will find the implementation of the class that the object
+actually is, which is our new subclass.
```
telephon: This sentence is still a little unclear

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+In such cases, a unique method name like myclassInit should be used.
```
telephon: or better, a specialised name that indicates what kind of initialization it performs.

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,136 @@ MyClass : SomeSuperclass {
...
+::
+
+While the code::init:: pattern is the most flexible, care must be taken as inside the code::init:: functions,
+the instance is in an incomplete state and other methods might not work as expected;
+this becomes exceptionally confusing when combined with threads and asynchronous execution, for example, when interacting with the server.
```
telephon: for example, this may become confusing when

HelpSource/Reference/Functions.schelp
```diff
@@ -8,23 +8,42 @@ section:: Introduction
...
 Function definitions are enclosed in curly brackets code::{}::. Argument declarations, if any, follow the open bracket. Variable declarations follow argument declarations. An expression follows the declarations.
 code::
-{ arg a, b, c;  var d;   d = a * b; c + d }
+{
+	|a, b, c|
+	var d;
+	d = a * b;
+	c + d
+}
 ::
```
telephon, line 16: maybe keep with the "standard" layout?
```
{ |a, b, c|
	var d;
	d = a * b;
	c + d
}
```

HelpSource/Reference/Functions.schelp
```diff
@@ -8,23 +8,42 @@ section:: Introduction
...
 f.value(4, 5).postln;
-f.value(10, 200).postln;
+f.(4, 5).postln; // same as above
+f.(10, 200).postln;
 ::
 
 An empty function returns the value nil when evaluated.
 code::
-{}.value.postln;
+{}.().postln;
```
telephon: this is unnecessarily confusing, I think. {}.value.postln; was fine.

HelpSource/Reference/Functions.schelp
```diff
@@ -87,18 +89,57 @@ code::
 
 See link::Reference/Literals:: for more information.
 
+
+
+section:: Variable Arguments
+
+What happens if you don't know how many arguments a function should have?
```
telephon: For reference (unlike guides) it would be good not to address th ereader directly. At least this is the style that exists and I'd prefer not to mix too much.

HelpSource/Guides/ObjectPrototyping.schelp
JordanHendersonMusic: I don't really think that is the same though. `q` above is the equivalent of a metaclass, we don't really need that for a constructor as it is just a function. Of course, someone could do that if they wanted, but its probably overkill for most use cases.

HelpSource/Guides/WritingClasses.schelp
JordanHendersonMusic: I don't quite understand which bit this comment refers to?

HelpSource/Guides/ObjectPrototyping.schelp
telephon: Yes, sure, I rather wondered if people will ask themselves if there is a reason not to do that. Since in prototype objects, there is no distinction betwee class and metaclass as such, I thought it is not such a big thing to just continue to use the same principle everywhere.

But fine with me!

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
+title:: Object Prototyping
+categories:: Guides
+summary:: Writing runtime 'classes'
+related:: Classes/IdentityDictionary
+
+
+You cannot create nor edit a class without recompiling the language... but what if you need that?
+
+Object prototyping is a middle ground that lets you build class like structures at runtime.
```
HotwheelsSisyphus: class-like (hyphen)

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+It works by taking advantage of the class link::Classes/IdentityDictionary:: (or more generally its child link::Classes/Event::) and the special method link::Classes/Object#-doesNotUnderstand::.
```
HotwheelsSisyphus: Maybe use "subclass" instead of "child"? (And change that one mention of parent and MyChildClass in WritingClasses to superclass and subclass)

Reasoning: It seems that in sc, the child/parent language is mostly associated with UI Objects and Nodes (I verified this by searching the relevant docs, i.e. Classes, Class, Intro to Objects, ...), so maybe that would make things ever so slightly easier to understand. Otoh, it's also probably fine as is.

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+		word: word,
+		speak: { |self, adverb| "I say \"%\" %.".format(self.word, adverb) }
+	)
+};
+
+var cat = makeSpeaker.("meow");
+var dog = makeSpeaker.("woof");
+
+cat.speak("loudly").postln; // I say "meow" loudly.
+dog.speak("quitely").postln; // I say "woof" quitely.
```
HotwheelsSisyphus: quietly

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+p1.speak_secrets().postln; // My secret is "I'm a dog".
+)
+::
+
+
+section:: Dangers
+
+There are two drawbacks over classes.
+
+The first, is that any key, including 'methods', can be overridden at any time.
```
HotwheelsSisyphus, line 100: Bizarre American punctuation would have: 'methods,' ... i.e., comma before closing quotation mark. Feel free to ignore, I hate it anyway.

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+Essentially, what makes link::Classes/IdentityDictionary:: powerful, also makes it fragile and easy to break.
+The only way to avoid this mistake is by being vigilant when adding keys to an object prototype.
+
+The second drawback is that you cannot have a key with the same name as a method in link::Classes/Event:: or any parent class,
+otherwise the real class's method will be called, not the pseudo method you have written.
```
HotwheelsSisyphus: pseudo-method (hyphen)

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+
+code::
+a = (numChannels: 10)
+a.numChannels // 1
+::
+
+As seen above, one expects the value code::10:: to be returned, but instead link::Classes/Object#-numChannels:: is called and the value code::1:: returned.
+
+There are two ways to mitigate this.
+One, use longer keys or names in snake case rather than pascal case: that is names_written_like_this, ratherThanNamesWrittenLikeThis, as SuperCollider only uses the latter.
```
HotwheelsSisyphus: Maybe highlight enumerations somehow... I think a table would be overblown but you could strong::One:: and strong::Two::. Using 1. and 2. also works to visually highlight the enumeration.

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
 
 section:: Instances
 
-code::Object.new:: will return a new object. When overriding the class method code::.new:: you must call the
-superclass, which in turn calls its superclass, up until code::Object.new:: is called and an object is actually created
-and its memory allocated.
+code::Object.new:: will return a new object.
+When overriding the class method code::.new:: you must call the
```
HotwheelsSisyphus: linebreak

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
+	*new { |myC| ^super.newCopyArgs(a: 10, b: 20, c: myC) }
+}
+::
+
+If initialisation logic that prepares the arguments before assignment is required,
+ideally, it should go before the call to code::super.newCopyArgs::.
+This way the final object never exists in an invalid state — an alternative is the use of an code::init:: method and shall be shown shortly.
+
+
+However, when there is a hierarchy of classes, code::newCopyArgs:: requires all the base class arguments be passed in the order they are defined.
```
HotwheelsSisyphus: requires that ... be passed 
is probably clearer to non-native speakers?

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
+}
+::
+
+If initialisation logic that prepares the arguments before assignment is required,
+ideally, it should go before the call to code::super.newCopyArgs::.
+This way the final object never exists in an invalid state — an alternative is the use of an code::init:: method and shall be shown shortly.
+
+
+However, when there is a hierarchy of classes, code::newCopyArgs:: requires all the base class arguments be passed in the order they are defined. 
+This makes changing the bass class difficult.
```
HotwheelsSisyphus: it's all about the base

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
+Here is one way to mitigate this issue using variables keyword arguments link::Reference/Functions#Variable Arguments:: and link::Classes/Object#-superPerformArgs::.
```
HotwheelsSisyphus: variable keyword arguments

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
-actually is, which is our new subclass. In such cases, a unique method name like myclassInit should be used.
 ::
 
-One easy way to copy the arguments passed to the instance variables when creating a class is to use link::Classes/Object#*newCopyArgs::.  This
-method will copy the arguments to the instance variables in the order that the variables were defined in the class,
-starting from the parent classes and working it's way down to the current class.
+Although this approach might seem complex, the benefit is that the new instance variables can be added to either class and it will always be clear which one you meant to assign to.
+
+Should it be possible to remove instance members in the future,
+it is good to require only keyword argument be used.
```
HotwheelsSisyphus: keyword arguments (plural)

again, subjunctive with "that" is probably easier for the ESL folks.

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
-	var a,b,c;
+Foo {
+	*new { |... args, kwargs| 
+		if(args.empty.not) { Error().throw };
+		^super.superPerformArgs(\newCopyArgs, #[], kwargs);
+	}
+}
+::
+
+Another benefit is that the instance is never in an incomplete state, either all the instance members have been assigned to, or the object does not exist.
```
HotwheelsSisyphus: I'd use a colon here:

Another benefit is that the instance is never in an incomplete state: Either all the instance members have been assigned to, or the object does not exist.

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
+	derivedInit { |newA, newB, newC|
+		// do initialisation with all the arguments
+		this.init(newA, newB) // call Base.init
+		c = newC; // assign to Derived's c
 	}
 }
 ::
 
+Warning:: 
+If both the parent and child class define an code::init:: method, the child class will override the parent, meaning it will no longer be possible to call that implementation.
```
HotwheelsSisyphus: Base and Derived instead of parent and child?

HelpSource/Reference/Functions.schelp
```diff
@@ -8,25 +8,43 @@ section:: Introduction
...
-{ arg a, b, c;  var d;   d = a * b; c + d }
+{ |a, b, c|
+	var d;
+	d = a * b;
+	c + d
+}
 ::
 
-Functions are not evaluated immediately when they occur in code, but are passed as values just like integers or strings.
+Functions are not evaluated immediately when they occur in code, but are themselves object that can be passed around.
```
HotwheelsSisyphus: objects

HelpSource/Reference/Functions.schelp
```diff
@@ -8,25 +8,43 @@ section:: Introduction
...
 
-A function may be evaluated by passing it the code::value:: message and a list of arguments.
+A function may be evaluated by passing it the code::value:: message, or its shorthand code::.()::, and a list of arguments.
```
HotwheelsSisyphus: A Function object may be evaluated ...

maybe .value() for clearer parallelism to .()

HelpSource/Reference/Functions.schelp
```diff
@@ -8,25 +8,43 @@ section:: Introduction
...
 code::
-f = { arg a, b; a + b };
+f = { |a, b| a + b };
 f.value(4, 5).postln;
-f.value(10, 200).postln;
+f.(4, 5).postln; // same as above
+f.(10, 200).postln;
 ::
 
 An empty function returns the value nil when evaluated.
```
HotwheelsSisyphus: An empty Function object

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
 ::
 
-If you want all the arguments put in an Array
+See link::Reference/Literals:: for more information.
+
+
+
+section:: Variable Arguments
+
+Variable arguments are how function catch an unknown number of arguments or keyword arguments.
```
HotwheelsSisyphus: code::Function::s or Function objects
(plural anyway, uppercase Function)

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
 
-If you want all the arguments put in an Array
+See link::Reference/Literals:: for more information.
+
+
+
+section:: Variable Arguments
+
+Variable arguments are how function catch an unknown number of arguments or keyword arguments. 
+There are two possible variable argument, the first for positional arguments and second for keyword arguments.
```
HotwheelsSisyphus: There are two possible variable arguments, the first for positional arguments, and the second for keyword arguments.

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
+// catches all arguments
+g = { |...args| "args: %".format(args) };
+
+g.(1, 2, 3) // args: [1, 2, 3]
+g.() // args: []
 ::
 
-In general arguments may be initialized to literals or expressions, but in the case of Function:play or SynthDef:play, they may only be initialized to literals.
+The second variable argument catches all keyword arguments.
+It put them in an link::Classes/Array:: of key-value pairs.
```
HotwheelsSisyphus: puts

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
 
-{ arg freq = Array.geom(4, 100, 3); Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
+f.(1, 2, 3, meow: 10, woof: 20) // args: [1, 2, 3], kwargs: [meow, 10, woof, 20]
 
-// but this is:
-{ arg freq =  #[ 100, 300, 900, 2700 ]; Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
 ::
 
-See link::Reference/Literals:: for more information.
+The use of keyword variable arguments is often paired with link::Classes/Object#-performArgs::.
```
HotwheelsSisyphus: variable keyword arguments 
(?)

HelpSource/Classes/IdentityDictionary.schelp
```diff
@@ -1,11 +1,15 @@
 CLASS::IdentityDictionary
 summary::associative collection mapping keys to values
-related::Classes/Environment, Classes/Event
+related::Classes/Environment, Classes/Event, Guides/ObjectPrototyping
 categories::Collections>Unordered
 
 DESCRIPTION::
 An IdentityDictionary is an associative collection mapping keys to values. Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
 
+An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::). They behave like objects whose methods can be programmed at runtime.
```
HotwheelsSisyphus: I'd make that into one sentence (below); alternatively, use "These" instead of "They"

An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::), which behave like objects whose methods can be programmed at runtime.

HotwheelsSisyphus (commented):
Just suggestions really, I mostly went typo-hunting... so this concerns the prose more than code.
 
I think the child/parent words could be replaced with other language (subclass/superclass) because that seems to be more current in sc, but I don't think it's a huge issue or impedes understanding.

HelpSource/Reference/Functions.schelp
prko: parentheses?
```
(
{ |a, b, c|
	var d;
	d = a * b;
	c + d
}
)
```

HelpSource/Reference/Functions.schelp
prko: How about adding both?
```
by passing it the code::value():: or code::value:: message, or its shorthand code::.()::,
```

HelpSource/Reference/Functions.schelp
prko: then add `link::`?
```
An empty link::Classes/Function:: object
```

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
 Pipe style, default value is an expression:
 code::{ |x = (10.rand)| ... } :: link::#[2]::
 
-If the last argument in the list is preceded by three dots (an ellipsis), then all the remaining arguments that were passed will be assigned to that variable as an link::Classes/Array::. Arguments must be separated by commas.
 
-examples:
+In general arguments may be initialized to literals or expressions, but in the case of Function:play or SynthDef:play, they may only be initialized to literals.
```
prko:
```
general arguments may be initialized to literals or expressions, but in the case of link::Classes/Function#-play:: or link::Classes/SynthDef#-play::, they may only be initialized to literals.
```

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
 code::
-{ arg a, b, c=3; } // is equivalent to:
+// this is okay:
 
-{ |a, b, c=3| }
+{ arg a = Array.geom(4, 100, 3); a * 4 }.value;
```
prko, line 79: How about this?
```
(
{ arg a = Array.geom(4, 100, 3); 
	a * 4 
}.value;
)

// this is also okay:
(
{ |a = (Array.geom(4, 100, 3))|
	a * 4 
}.value;
)
```

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
 
-{ arg x='stop', y, z=0; } // these args are initialised
+// this is not:
 
-{ arg a, b, c ... d; } // any arguments after the first 3 will be assigned to d as an Array
+{ arg freq = Array.geom(4, 100, 3); Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
+
+// but this is:
+{ arg freq =  #[ 100, 300, 900, 2700 ]; Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
```
prko:
```
{ arg freq = #[ 100, 300, 900, 2700 ]; Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // produces sound!
```

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
+
+
+
+section:: Variable Arguments
+
+Variable arguments are how function catch an unknown number of arguments or keyword arguments. 
+There are two possible variable argument, the first for positional arguments and second for keyword arguments.
+They are marked by an ellipsis after any normal arguments.
+
+The first variable argument catches all remaining arguments in an link::Classes/Array::
```
prko: I feel the following might be better:
``` 
The first variable argument (the positional variable argument)
```
However, I am not sure...

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
+
+// catches all arguments
+g = { |...args| "args: %".format(args) };
+
+g.(1, 2, 3) // args: [1, 2, 3]
+g.() // args: []
 ::
 
-In general arguments may be initialized to literals or expressions, but in the case of Function:play or SynthDef:play, they may only be initialized to literals.
+The second variable argument catches all keyword arguments.
```
prko, line 113: How about this?
```
The second variable argument (the keyword variable argument) catches all keyword arguments.
```

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
+f.(1, 2, 3, meow: 10, woof: 20) // args: [1, 2, 3], kwargs: [meow, 10, woof, 20]
 
-// but this is:
-{ arg freq =  #[ 100, 300, 900, 2700 ]; Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
 ::
 
-See link::Reference/Literals:: for more information.
+The use of keyword variable arguments is often paired with link::Classes/Object#-performArgs::.
+
+You cannot reassign the value of the variable arguments even if they share the same name.
```
prko: How about this?
```
You cannot reassign the value of the variable arguments inside the keyword variable arguments, even if they share the same name.
```

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
+code::
+f = { |...args, kwargs| "args: %, kwargs: %".format(args, kwargs) };
+f.(args: 10, kwargs: 20) // args: [], kwargs: [args: 10, kwargs: 20]
+::
+
 
 anchor::[1]::
```
prko, line 141: `subsection:: [1] Literal argument defaults` can be linked as `link::# [1] Literal argument defaults`.

So, `anchor::[1]::` seems to be unnecessary here, I think.

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,82 @@ code::{ |x = 1| ... } :: link::#[1]::
...
+f.(2);  // prints 2
 
-f.value;   // prints 1
+f.();   // prints 1
 
-f.value(nil);  // prints nil
+f.(nil);  // prints nil
 ::
 
 anchor::[2]::
```
prko, line 152: as above, I think `anchor::[2]::` is not needed.

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+var cat = makeSpeaker.("meow");
+var dog = makeSpeaker.("woof");
+
+cat.speak("loudly").postln; // I say "meow" loudly.
+dog.speak("quitely").postln; // I say "woof" quitely.
+)
+::
+
+
+section:: Private Instance Variables (Members)
```
prko, line 74: Could you add an expression of what (Member) here means?

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+
+
+You cannot create nor edit a class without recompiling the language... but what if you need that?
+
+Object prototyping is a middle ground that lets you build class like structures at runtime.
+It works by taking advantage of the class link::Classes/IdentityDictionary:: (or more generally its child link::Classes/Event::) and the special method link::Classes/Object#-doesNotUnderstand::.
+While there are some oddities with this approach, it is extremely powerful.
+
+
+section:: Instance Variables (Members)
```
prko, line 12: Could you add an expression of what (Member) here means?

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
+::
+
+In this case note that code::super.new:: calls the method code::new:: on the superclass and returns a new object.
 
-	// this is a normal constructor method
-	*new { | arga, argb, argc |
-		^super.new.init(arga, argb, argc)
+subsection:: Instance Variables and Initialisation Logic
+
+Instance can have members (instance variables), where each instance of the class can have a different value for each member.
```
prko: Could you add an expression of what `members` here mean?

prko:
>@prko does this help? Specifically the bit in Functions.schelp

Yes, it's extremely helpful!
This is exactly the kind of documentation I always want to read in SC!

HelpSource/Reference/Functions.schelp
telephon: I don't think we need to put this on several lines.

HelpSource/Reference/Functions.schelp
telephon: Why did you suggest to change "In general, arguments" to "general arguments"?

HelpSource/Reference/Functions.schelp
telephon: Yes, we need a better name for the keyword arguments in the function. A shorter one would be nice though. Maybe "inner keyword arguments"?

HelpSource/Guides/ObjectPrototyping.schelp
telephon: Hey, we can't explain everything! Of course we could just leave out "member". It was meant to make it easier to understand, not harder …

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: Also, this was already there, I just moved it around.

HelpSource/Guides/ObjectPrototyping.schelp
JordanHendersonMusic: @prko the reason why I use the terms 'members' and 'methods' is because, if the user has any programming experience, they are significantly more likely to know the terms 'member' and 'method' than 'instance variable' and 'message' which are smalltalk specific terms — smalltalk kinda faded away after the 90s when javascript won the webwars and object orient programming went in a slightly different direction.

HelpSource/Guides/ObjectPrototyping.schelp
JordanHendersonMusic: > I hate it anyway.

the feeling is shared.

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: I think that might be a bit long here, I don't want to get into sc syntax too much, just deal with the specifics of functions

HelpSource/Guides/ObjectPrototyping.schelp
prko: I see, then may be as follows?
```
section: Private Instance Variables (Referred to as 'Members' in Some Languages)
```

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: Since this is a reference document for 'Function', the link and capitalisation throughout the document could be too much — it is already at the top of the file.

prko: Ah, ok! thanks

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: That is too repetitive.

HelpSource/Reference/Functions.schelp
prko: Oh, sorry it was copy-paste mistake.

The correct sentence would be:
>In general arguments may be initialized to literals or expressions, but in the case of link::Classes/Function#-play:: or link::Classes/SynthDef#-play::, they may only be initialized to literals.

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: I think this is okay because there are examples right after

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: I think the example below clarifies what is meant, this seems like a tricky thing to word, I'd probably draw a distinction using the term 'call–sight'.

HelpSource/Reference/Functions.schelp
telephon: Please, be careful not to add more work. Your comments are useful, but if the respondent has to invest time to double-check, it is lost again. The best here would be to only suggest the part that is changed, thus ` link::Classes/Function#-play:: or link::Classes/SynthDef#-play::`. Thank you @prko !

HelpSource/Reference/Functions.schelp
telephon: yes, good.
you mean call-site?

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: I can't get that to work, I think it is fine as is

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: how about now? I also added another sentence above and an example below

HelpSource/Reference/Functions.schelp
capital-G: I think the suggest change feature of GH is actually really helpful for smaller remarks as they hightlight the proposed changes in line and applying or dismiss the proposed change from within the browser, see https://haacked.com/archive/2019/06/03/suggested-changes/ 
I think this is a good workflow for minor edits such as typos or formatting issues.

HelpSource/Reference/Functions.schelp
prko: Oh, sorry: 
There should no white space after #:
```
link::#[1] Literal argument defaults.
```
If it still does not work, please refer to: https://github.com/supercollider/supercollider/blob/ddd98743ff97e35423614e2a4b4f9200ac3c4faa/HelpSource/Classes/Bus.schelp#L116-L117

https://doc.sccode.org/Classes/Bus.html#Asynchronous%20Control%20Bus%20Methods

JordanHendersonMusic: Does that work inside of a subsection though? It also isn't my work just what was there already.

prko: >Does that work inside of a subsection though? 

Yes, I have just retested.

>It also isn't my work just what was there already.

Ah, sorry!
Then leave it as it is!

capital-G: @prko PR reviews should only cover the actually touched code of the PR in order to keep things focused. If you notice something unrelated that should be changed please file the changes via another PR.

prko: Sorry, I will do!

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
+}
+::
+
+These can be assigned in two ways.
+Through code::newCopyArgs:: or through an code::init:: method.
+The former is recommend and discussed first.
+
+Rather than calling the method code::new:: a call to link::Classes/Object#*newCopyArgs:: can be used to initialise members.
+
+The code below allows the call sight to decide the initial value of code::c:: and defaults code::a:: to 10 and code::b:: to 20.
```
cdbzb: typo: `site` not `sight`

JordanHendersonMusic: thanks!

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,87 @@ code::{ |x = 1| ... } :: link::#[1]::
...
-If you want all the arguments put in an Array
+See link::Reference/Literals:: for more information.
+
+
+
+section:: Variable Arguments
+
+Variable arguments are how functions catch an unknown number of arguments or keyword arguments. 
+There are two possible variable arguments, the first for positional arguments, and second for keyword arguments.
+They are marked by an ellipsis after any normal arguments.
```
cdbzb: should mention that they need to be separated be a comma?

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,87 @@ code::{ |x = 1| ... } :: link::#[1]::
...
-{ arg freq = Array.geom(4, 100, 3); Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
+f.(1, 2, 3, meow: 10, woof: 20) // args: [1, 2, 3], kwargs: [meow, 10, woof, 20]
 
-// but this is:
-{ arg freq =  #[ 100, 300, 900, 2700 ]; Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
 ::
 
-See link::Reference/Literals:: for more information.
+The use of variable keyword arguments is often paired with link::Classes/Object#-performArgs::.
+
```
cdbzb, line 128: an example here might be helpful!

HelpSource/Reference/Functions.schelp
```diff
@@ -8,25 +8,43 @@ section:: Introduction
...
+f.(10, 200).postln;
 ::
 
-An empty function returns the value nil when evaluated.
+An empty function object returns the value nil when evaluated.
 code::
 {}.value.postln;
 ::
 
+Also the arguments can be called by name.
```
cdbzb: perhaps "Arguments can also be provided by name"?

HelpSource/Reference/Functions.schelp
```diff
@@ -8,25 +8,43 @@ section:: Introduction
...
+code::
+f = { |a, b| a + b };
+f.(a: 10, b: 2) // 12
+
+f.(b: 3, a: 10) // 13, order does not matter for keywords
+f.(3, b: 4) // 7, 'a' is always first argument
+
+f.(3, a: 4) // Error: Duplicate keyword 'a' at position 0.
+
+f.(a: 3) // Error: binary operator '+' failed, b is nil and you cannot add a number and 'nil'
```
cdbzb: I think both of these `nil`s should be the same - not sure what the stye guide suggests for keywords...

HelpSource/Reference/Functions.schelp
```diff
@@ -8,25 +8,43 @@ section:: Introduction
...
+::
+
 A function can be thought as a machine able to perform a task on demand, e.g. a calculator. The calculator can receive input (args) and can output a value, the result of the performed operations. The function definition can then be thought as the building of the calculator: once built, the calculator does nothing until it is requested to work (by passing the value method to a function).
```
cdbzb, line 48: missing 'of': "a function can be thought of as etc" (while we're here....)

JordanHendersonMusic: Not my work.

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: You can just click on the link, there is a clear example over there.

HelpSource/Guides/ObjectPrototyping.schelp
JordanHendersonMusic: I think this is fine as is.

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: Seems overkill to me, but I've done it.

HelpSource/Reference/Functions.schelp
JordanHendersonMusic: I've fixed the nils. I don't understand the second part of of that sentence.

HelpSource/Classes/Function.schelp
```diff
@@ -3,6 +3,8 @@ summary::Implements a function
 categories::Core>Kernel
 related::Reference/Functions, Classes/AbstractFunction, Classes/FunctionDef
 
+This document deals with the usage of function objects, for a more general introduction and documentation of syntax, see link::Reference/Functions::.
```
HotwheelsSisyphus: This document deals with the usage of Function objects. For a more general introduction and documentation of syntax, see link::Reference/Functions::.

HelpSource/Classes/IdentityDictionary.schelp
```diff
@@ -1,11 +1,15 @@
...
-related::Classes/Environment, Classes/Event
+related::Classes/Environment, Classes/Event, Guides/ObjectPrototyping
 categories::Collections>Unordered
 
 DESCRIPTION::
 An IdentityDictionary is an associative collection mapping keys to values. Keys and values can be arbitrary objects, but for keys, often a link::Classes/Symbol:: is used.
 
+An IdentityDictionary can be used to implement prototype objects (link::Guides/ObjectPrototyping::), which behave like objects whose methods can be programmed at runtime.
+
+
```
HotwheelsSisyphus: max 1 empty line

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+summary:: Writing runtime 'classes'
+related:: Classes/IdentityDictionary
+
+
+You cannot create nor edit a class without recompiling the language... but what if you need that?
+
+Object prototyping is a middle ground that lets you build class–like structures at runtime.
+It works by taking advantage of the class link::Classes/IdentityDictionary:: (often the subclass link::Classes/Event:: is used) and the special method link::Classes/Object#-doesNotUnderstand::.
+While there are some oddities with this approach, it is extremely powerful.
+
```
HotwheelsSisyphus, line 11: max 1 empty line

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
+title:: Object Prototyping
+categories:: Guides
+summary:: Writing runtime 'classes'
+related:: Classes/IdentityDictionary
+
```
HotwheelsSisyphus, line 5: max 1 empty line

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+code::
+a = (meow: 10)
+a.woof = 4;
+a.woof // 4
+a.meow // 10
+::
+
+section:: Methods
+
+By assigning a function to a key, you can call it just as a normal instance method would be called.
```
HotwheelsSisyphus: By assigning a function to a key, you can call that function as if it were a normal instance method.

HelpSource/Guides/ObjectPrototyping.schelp
```diff
@@ -0,0 +1,142 @@
...
+
+cat.speak("loudly").postln; // I say "meow" loudly.
+dog.speak("quietly").postln; // I say "woof" quietly.
+)
+::
+
+
+section:: Private Instance Variables (Members)
+
+These can be implemented inside the constructor function as normal variables, however, you don't add them to the dictionary.
```
HotwheelsSisyphus, line 76: I actually don't easily understand what "you don't ..." means. Is it just that you should use a different approach than for the constructor, and should NOT do this:
```
 (
 var makeSpeaker = { |word|
 	(
 		word: word,
       secret: "default secret", //what "you don't" do
 		speak: { |self, adverb| "I say \"%\" %.".format(self.word, adverb) },
 		speak_secret: { |self| "My secret is \"%\".".format(secret.value) }
 	)
 };
 
 var p1 = makeSpeaker.();
 p1.secret = "actual secret"; // what "you don't" do?
 p2.secret = "another secret"; // what "you don't" do?
 )
 ```
 
 If it's that, then clarify accordingly:
 
Instance variables can be implemented inside the constructor function as normal variables; note that they should not be added to the dictionary, because [quick five word explanation].

Comment (HotwheelsSisyphus) [reply HelpSource/Guides/ObjectPrototyping.schelp:76]: and maybe , depending on how important you think it is, some short negative example might be instructive (but should be clearly marked as a "DON'T DO THIS" case)

File: HelpSource/Reference/Functions.schelp [hunk 67]
```diff
@@ -53,52 +71,87 @@ code::{ |x = 1| ... } :: link::#[1]::
...
+f.(1, 2, 3, meow: 10, woof: 20) // args: [1, 2, 3], kwargs: [meow, 10, woof, 20]
 
-// but this is:
-{ arg freq =  #[ 100, 300, 900, 2700 ]; Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // silence
 ::
 
-See link::Reference/Literals:: for more information.
+The use of variable keyword arguments is often paired with link::Classes/Object#-performArgs::.
+
+While you can name the variable arguments anything, they always refers to the unmatched positional and keyword arguments, so it is often best to name them code::args:: and code::kwargs::.
```
HotwheelsSisyphus: Maybe: 
While you can pick any name for the variable arguments, ...

HelpSource/Guides/WritingClasses.schelp
```diff
@@ -161,54 +161,148 @@ MyClass : SomeSuperclass {
...
+Warning:: 
+If both the code::Base:: and code::Derived:: class define an code::init:: method, the code::Derived:: class will override the code::Base::'s implementation, 
+meaning it will no longer be possible to call that implementation.
+In such cases, a unique method name like code::myclassInit:: should be used, or better yet, a specialised name that indicates what kind of initialization is performed.
+::
+
+While the code::init:: pattern is the most flexible, care must be taken as inside the code::init:: functions,
+the instance is in an incomplete state and other methods might not work as expected;
+for example, this may become confusing when combined with threads and asynchronous execution, for example, when interacting with the server.
+
```
HotwheelsSisyphus, line 287: through 305: max 1 empty line at a time

HelpSource/Reference/Functions.schelp
```diff
@@ -53,52 +71,87 @@ code::{ |x = 1| ... } :: link::#[1]::
...
 
-{ arg a, b, c ... d; } // any arguments after the first 3 will be assigned to d as an Array
+// but this is:
+{ arg freq =  #[ 100, 300, 900, 2700 ]; Mix(SinOsc.ar(freq, 0, 0.1)) }.play; // produces sound!
 ::
 
-If you want all the arguments put in an Array
+See link::Reference/Literals:: for more information.
+
+
```
HotwheelsSisyphus: max 1 empty line

HotwheelsSisyphus:
Btw I think this can be merged if Jordan doesn't have the time to address the nitpicks I've brought up in the last review. Maybe I misunderstood the "request for review"; I didn't mean to slow down the process. It's totally fine with me if the private instance variables thing gets fixed at a (much) later date.

JordanHendersonMusic:
@HotwheelsSisyphus sorry, I've been quite busy and I forget about this. If this is okay as is, then I'd rather not make those changes. 

Rather than making comments like you have, you can instead suggest an edit on github, which means all I have to do is click accept. It probably takes you less time to suggest an edit than writing a comment describing what you think should happen.

JordanHendersonMusic:
Thanks!

Okay this is done then.. don't forget to squash!

PR #6996, opened 2025-06-20
Let's consider macOS universal build... again

dyfer:
## Purpose and Motivation

The official Qt6 binaries that we use for the releases are compiled for both arm64 and x64 architectures. The only thing that prevented us from building a universal binary of SC on macOS was the unavailaility of dual-architecture dependencies: fftw, libsndfile, and readline.

I was previously experimenting with adding a universal "triplet" for vcpkg, but that doesn't work with some of the libraries which use autotools. 

Another option is to build packages for both architectures and then combine them manually. This is what this PR does... This is not the most elegant solution, but it seems to work. The script is embedded directly in the vcpkg install step, which makes the workflow file a bit bigger. Luckily, the script is not nearly as extensive as what we [used to use for making universal binaries of homebrew libs](https://gist.github.com/dyfer/6c83905d4593750105897e51e87ec345).

As a bonus, we have one less macOS job running concurrently, and when cache is available, it takes a similar amount of time as the other macOS builds. 

The practical changes are: 
- fftw, libsndfile, and readline are now installed with vcpkg
- their dependencies are embedded (they are built as static libs), that's why we don't have e.g. libogg in the bundle as a separate lib
- portaudio is built from our source tree (so that there's one less package to keep track of, but this could come from vcpkg as well if we'd like)
- [FindReadline.cmake](https://github.com/supercollider/supercollider/blob/48ff7d923d7503c86da1b201776096ec2640e99f/cmake_modules/FindReadline.cmake) script is updated to not blindly use brew prefix command, but to check where the header/library files are located (this was needed to enable finding readline in vcpkg on macOS)

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -13,32 +13,21 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - job-name: "arm64  with ctest"
+          - job-name: "universal arm64 x64 with ctest"
             os-version: "15"
             xcode-version: "16.0"
             qt-version: "6.7.3" # will use qt from aqtinstall
             qt-modules: 'qtwebengine qtwebchannel qtwebsockets qtpositioning'
             deployment-target: "11"
-            cmake-architectures: "arm64"
-            homebrew-packages: "libsndfile readline fftw portaudio"
-            vcpkg-packages: ""
-            vcpkg-triplet: ""
-            extra-cmake-flags: "" # -D SC_VERIFY_APP=ON # verify app doesn't seem to work with official qt6
-            artifact-suffix: "macOS-arm64"
-            ctest: true
-
-          - job-name: "x64 with ctest"
-            os-version: "13"
-            xcode-version: "15.2"
-            qt-version: "6.7.3" # will use qt from aqtinstall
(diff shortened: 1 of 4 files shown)


dyfer:
Merge note: don't squash :)
After the review I'll squash GHA additions into one commit - I think the cmake readline change should be in a separate commit.

capital-G (approved):
LGTM!
Thanks for putting the shell script into a dedicated file - makes reviewing much easier if syntax highlighting becomes available.

So, IIUC: use vcpkg to install dependencies for both architectures and afterwards use `lipo` to merge them into a shared universal lib which will then be used during building/linking?

Only thing left here: Does it affect the signing/notarization process in any way? Probably not, right?
I assume the arm64 part is working since it is used within the test pipeline. Have you tested the x64 part as well? I don't have fast internet r/n and can't download the file :/

edit: Feel free to merge marcin! :)

dyfer:
Yes, I've run the intel build locally - it starts and makes sound :)

I rebased to squash GHA commits - merging!

PR #6997, opened 2025-06-22
classlib: DoesNotUnderstandError: refactor suggestions

telephon:
## Purpose and Motivation

The DoesNotUnderstandError suggestions sometimes are over-confident, because:
- they only give one suggestion even if there may be others that match almost as well
- they suggest that there has been a typo where there may have been other reasons
- the errorString contained the suggestion, which one may want to keep separate

## Types of changes

This improves the DoesNotUnderstandError posting:
- it now suggests up to 4 similar names
- it uses a neutral language that doesn't speculate about the reason
- The errorString itself now does not contain the suggestion
- As an experiment to be evaluated over time, we sometimes post class suggestions if they are appropriate.
- for better introspection, some new methods have been introduced
- the implementation of DoesNotUnderstand is refactored

I took the liberty to remove an instance variable called `suggestedCorrection`, because it makes no sense anymore (we have different kinds of suggestions)

So this is a breaking change if you happen to be using it (rather unlikely).

For this, I took the occasion to add a better interface to `Class`, which allows us to retrieve all responing methods of an object through its class (`respondingMethods`). A unit test for this is added as well.

Maybe the expression "Closest matching message found for the receiver:" is not yet optimal, suggestions welcome!

```supercollider

[1, 2, 3].collkt

// before

^^ ERROR: Message 'collkt' not understood.
Perhaps you misspelled 'collect', or meant to call 'collkt' on another receiver?
RECEIVER: [1, 2, 3]

// now:

^^ ERROR: Message 'collkt' not understood.
RECEIVER: [1, 2, 3]
Closest matching message found for the receiver:
	collect


[1, 2, 3].coll

// before

^^ ERROR: Message 'coll' not understood.
Perhaps you misspelled 'poll', or meant to call 'coll' on another receiver?
RECEIVER: [1, 2, 3]

// now:

^^ ERROR: Message 'coll' not understood.
RECEIVER: [1, 2, 3]
Closest matching messages found for the receiver:
	poll, call, dpoll, fill

```

```supercollider
// Sometimes, we get suggestions for classes:

// examples
View.new.drawingEnabled;
^^ ERROR: Message 'drawingEnabled' not understood.
RECEIVER: a View

Closest matching classes found for the selector:
	UserView, QUserView


```

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

## Merge notes
- Release Notes: Improve the DoesNotUnderstandError posting by providing more than one suggestion and rephrasing the string. Note that the suggestion is not part of the error string anymore.

### DIFF
--- a/HelpSource/Classes/String.schelp
+++ b/HelpSource/Classes/String.schelp
@@ -489,6 +489,33 @@ code::
 "These are several words which are fish".findAll("fish").postln;
 ::
 
+method::findSimilarIn
+
+From a list of strings, find similar strings, sorted by edit distance and limited by similarity.
+
+code::
+"hi".findSimilarIn(["ho", "hu?", "hugs", "Hola", "else"], 2); // [ho, hu?]
+"hi".findSimilarIn(["ho", "hu?", "hugs", "Hola", "else"], 3); // [ho, hu?, hugs, Hola]
+"hi".findSimilarIn(["ho", "Hi", "hugs", "Hola", "else"], 3, prioritizeCapitalization: false); // [ho, Hi, hugs]
+"hi".findSimilarIn(["ho", "Hi", "hugs", "Hola", "else"], 3, prioritizeCapitalization: true); // [Hi, ho, hugs, Hola]
+"hi".findSimilarIn(["ho", "Hi", "hugs", "Hola", "else"], 4, minSimilarity: nil); // [Hi, ho, hugs, Hola, else]
+"hi".findSimilarIn(["ho", "Hi", "hugs", "Hola", "else"], 4, minSimilarity: 0.01); // [Hi, ho, hugs, Hola]
+::
+
+argument::array
+An array of strings to find similar strings in.
+
+argument::maxEditDistance
+The largest link::Classes/SequenceableCollection#editDistance#edit distance:: that is accepted as a match.
+
+argument::minSimilarity
+Keep only results whose link::Classes/SequenceableCollection#similarity#similarity:: is larger than this value. If set to code::nil::, all values are accepted.
+
+argument::prioritizeCapitalization
+If true, calculated the edit distances on lower case versions.
+
+
+
 method::contains
 Returns a link::Classes/Boolean:: indicating if the String contains string.
 code::
--- a/SCClassLibrary/Common/Collections/String.sc
+++ b/SCClassLibrary/Common/Collections/String.sc
@@ -261,6 +261,25 @@ String[char] : RawArray {
 		_String_FindBackwards
 		^this.primitiveFailed
 	}
+	findSimilarIn { |array, maxEditDistance, minSimilarity, prioritizeCapitalization = true|
+		var names, editDistances, bestMatchIndices, searchFor;
+		if(prioritizeCapitalization) {
+			names = array.collect { |x| x.asString.toLower };
+			searchFor = this.toLower;
+		} {
+			names = array.collect { |x| x.asString };
+			searchFor = this;
+		};
+
+		editDistances = names.collect(editDistance(_, searchFor));
+		bestMatchIndices = editDistances.order;
+		bestMatchIndices = bestMatchIndices.select { |i|
+			maxEditDistance.isNil or: { editDistances[i] <= maxEditDistance }
+			and:
+			{ minSimilarity.isNil or: { similarity(names[i], searchFor) >= minSimilarity } }
+		}
+		^array[bestMatchIndices]
+	}
 	endsWith { arg string;
 		^this.contains(string, this.size - string.size)
 	}
@@ -487,7 +506,6 @@ String[char] : RawArray {
 	}
 	asAbsolutePath {
 			// changed because there is no need to create a separate object
-			// when String already knows how to make an absolute path
 		^this.absolutePath;  // was ^PathName(this).asAbsolutePath
 	}
 
--- a/SCClassLibrary/Common/Core/Error.sc
+++ b/SCClassLibrary/Common/Core/Error.sc
@@ -162,16 +162,16 @@ DoesNotUnderstandError : MethodError {
 			if(methodSuggestions.notEmpty) {
 				plural = if(methodSuggestions.size > 1) { "s" } { "" };
 				suggestedCorrection = methodSuggestions.first;
-				methodSuggestions = methodSuggestions.join(", ");
+				methodSuggestions = methodSuggestions.join("\n\t");
 				suggestion = suggestion ++
-				"\nClosest matching message% found for the receiver:\n\t%\n".format(plural, methodSuggestions);
+				"\nMessage% understood by the receiver with a similar name:\n\t%\n".format(plural, methodSuggestions);
 			};
 			classSuggestions = this.classSuggestions.keep(4);
 			if(classSuggestions.notEmpty) {
 				plural = if(classSuggestions.size > 1) { "s" } { "" };
-				classSuggestions = classSuggestions.join(", ");
+				classSuggestions = classSuggestions.join("\n\t");
 				suggestion = suggestion ++
-				"\nObject% which understand the selector % derive from:\n\t%\n".format(plural, selector, classSuggestions)
+				"\nObject% which understand the selector '%' derive from:\n\t%\n".format(plural, selector, classSuggestions)
 			}
 		}
 	}
--- a/SCClassLibrary/Common/Core/Error.sc
+++ b/SCClassLibrary/Common/Core/Error.sc
@@ -157,6 +157,7 @@ DoesNotUnderstandError : MethodError {
 	init {
 		var methodSuggestions, classSuggestions, plural;
 		if(receiver.notNil and: { selector.notNil }) {
+
 			methodSuggestions = this.methodSuggestions.keep(4);
 			if(methodSuggestions.notEmpty) {
 				plural = if(methodSuggestions.size > 1) { "s" } { "" };
@@ -177,7 +178,7 @@ DoesNotUnderstandError : MethodError {
 
 	methodSuggestions {
 		var names = receiver.class.respondingMethods.collect(_.name);
-		^selector.asString.findSimilarStringsIn(names, maxEditDistance: 2)
+		^selector.asString.findSimilarIn(names, maxEditDistance: 2)
 	}
 
 	classSuggestions {
--- a/SCClassLibrary/Common/Core/Error.sc
+++ b/SCClassLibrary/Common/Core/Error.sc
@@ -168,10 +168,9 @@ DoesNotUnderstandError : MethodError {
 			};
 			classSuggestions = this.classSuggestions.keep(4);
 			if(classSuggestions.notEmpty) {
-				plural = if(classSuggestions.size > 1) { "s" } { "" };
 				classSuggestions = classSuggestions.join("\n\t");
 				suggestion = suggestion ++
-				"\nObject% which understand the selector '%' derive from:\n\t%\n".format(plural, selector, classSuggestions)
+				"\nObjects which understand the selector '%' derive from:\n\t%\n".format(selector, classSuggestions)
 			}
 		}
 	}
@@ -182,7 +181,7 @@ DoesNotUnderstandError : MethodError {
 	}
 
 	classSuggestions {
-		^Object.findRespondingSubclasses(selector).collect(_.name)
+		^Object.findRespondingUpperSubclasses(selector).collect(_.name)
 	}
 
 	errorString {
--- a/SCClassLibrary/Common/Core/Error.sc
+++ b/SCClassLibrary/Common/Core/Error.sc
@@ -169,11 +169,11 @@ DoesNotUnderstandError : MethodError {
 				if(classSuggestions.size < 8) {
 					classSuggestions = classSuggestions.join("\n\t");
 					suggestion = suggestion ++
-					"\nObjects which understand the selector '%' derive from:\n\t%"
+					"\nObjects which respond to the selector '%' derive from:\n\t%"
 					.format(selector, classSuggestions)
 				} {
 					suggestion = suggestion ++
-					"\nMany other objects understand the selector '%' (found % superclasses)."
+					"\nMany other objects respond to the message '%' (found % superclasses)."
 					.format(selector, classSuggestions.size)
 				}
 			}
--- a/SCClassLibrary/Common/Core/Error.sc
+++ b/SCClassLibrary/Common/Core/Error.sc
@@ -145,7 +145,7 @@ ShouldNotImplementError : MethodError {
 }
 
 DoesNotUnderstandError : MethodError {
-	var <>selector, <>args, <>keywordArgumentPairs, <suggestedCorrection, suggestion = "";
+	var <>selector, <>args, <>keywordArgumentPairs, suggestion = "";
 	*new { arg receiver, selector, args, keywordArgumentPairs;
 		^super.new(nil, receiver)
 		.selector_(selector)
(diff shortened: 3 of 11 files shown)


JordanHendersonMusic:
You've got some failing tests, other than that, I think this is a great addition and the implementation looks fine to me.

cdbzb:
would it not make sense to sort the `Other classes that respond` list with `editDistance` as well?

JordanHendersonMusic:
I think in the case of returning a class name, what we probably want to do is return only the classes that don't have a superclass.

SCClassLibrary/Common/Core/Error.sc
```diff
@@ -157,26 +157,35 @@ DoesNotUnderstandError : MethodError {
...
 			methodNames = methods.collect { |x| x.name.asString.toLower };
+
 			// we compare lower-case versions to prioritize capitalization mistakes
 			lowerCaseSelector = selector.asString.toLower;
-			editDistances = methodNames.collect(_.editDistance(lowerCaseSelector));
-			minIndex = editDistances.minIndex;
+			editDistances = methodNames.collect(editDistance(_, lowerCaseSelector));
+			bestMatchIndices = editDistances.order;
+
+			bestMatchIndices = bestMatchIndices.keep(4);
```
JordanHendersonMusic: Why only the top 4?

SCClassLibrary/Common/Core/Error.sc
```diff
@@ -157,26 +157,35 @@ DoesNotUnderstandError : MethodError {
...
+
 			// Edit distance of 3 chosen arbitrarily; also, filter out completely dissimilar matches
```
JordanHendersonMusic: This comment is now incorrect as it's now 2

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -73,6 +73,19 @@ Class {
 		};
 		^nil
 	}
+	respondingMethods {
+		var all = Array.new, selectors = IdentitySet.new;
+		this.superclassesDo { arg class;
```
JordanHendersonMusic, line 78: Does the style guide require pipes?

JordanHendersonMusic (changes requested):
Why not just return all with a low edit distance? Why limit the count at all?

SCClassLibrary/Common/Core/Error.sc
telephon: seemed like a good number? Would you have preferred some more?

SCClassLibrary/Common/Core/Kernel.sc
telephon: I follow the rule that I keep the same format as the rest of the file unless I change more, then I'd adjust the rest of the file as well. I can do that if we agree this should be changed.

telephon:
> I think in the case of returning a class name, what we probably want to do is return only the classes that don't have a superclasses, otherwise the list could be huge!
> 
> It also doesn't take into consideration unique method, nor object prototyping, nor anything else that overloads doesnotunderstand... So it really can never be perfect.

Yes, this is why I hesitate. We could add it in the develop version (given that enough people use it) and then, if we get annoyed by the suggestions, remove it again.

telephon:
OK, good, all safety measures are in place.

HelpSource/Classes/String.schelp
```diff
@@ -489,6 +489,30 @@ code::
 "These are several words which are fish".findAll("fish").postln;
 ::
 
+method::findSimilarStringsIn
```
JordanHendersonMusic: Do we need to use the word 'string' again, could find similar in be enough?

HelpSource/Classes/String.schelp
```diff
@@ -489,6 +489,30 @@ code::
...
+code::
+"hi".findSimilarStringsIn(["ho", "hu?", "hugs", "Hola", "else"], 2); // [ho, hu?]
+"hi".findSimilarStringsIn(["ho", "hu?", "hugs", "Hola", "else"], 3); // [ho, hu?, hugs, Hola]
+"hi".findSimilarStringsIn(["ho", "Hi", "hugs", "Hola", "else"], 3, prioritizeCapitalization: false); // [ho, Hi, hugs]
+"hi".findSimilarStringsIn(["ho", "Hi", "hugs", "Hola", "else"], 3, prioritizeCapitalization: true); // [Hi, ho, hugs, Hola]
+"hi".findSimilarStringsIn(["ho", "Hi", "hugs", "Hola", "else"], 4, keepOnlySimilar: false); // [Hi, ho, hugs, Hola, else]
+"hi".findSimilarStringsIn(["ho", "Hi", "hugs", "Hola", "else"], 4, keepOnlySimilar: true); // [Hi, ho, hugs, Hola]
+::
+
+argument::collection
```
JordanHendersonMusic: Collection kind of implies a class, does this work with all collections, or just array?

HelpSource/Classes/String.schelp
```diff
@@ -489,6 +489,30 @@ code::
...
+A collection of strings to find similar strings in.
+
+argument::maxEditDistance
+The largest edit distance that is accepted as a match.
```
JordanHendersonMusic: Perhaps link to where edit distance is explained?

SCClassLibrary/Common/Collections/String.sc
```diff
@@ -486,8 +505,8 @@ String[char] : RawArray {
 		^PathName(this).asRelativePath(relativeTo)
 	}
 	asAbsolutePath {
-			// changed because there is no need to create a separate object
```
JordanHendersonMusic, line 489: Since there is a pr dealing with paths in flight, let just remove this in case it causes a clash.

SCClassLibrary/Common/Core/Error.sc
```diff
@@ -116,7 +116,7 @@ PrimitiveFailedError : MethodError {
 
 	*new { arg receiver;
 		^super.new(Thread.primitiveErrorString, receiver)
-			.failedPrimitiveName_(thisThread.failedPrimitiveName)
```
JordanHendersonMusic, line 119: What was already there was correct, same for the rest in this file.

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -40,7 +40,7 @@ Class {
 			classesInited.add(aClass);
 
 			if(aClass.isMetaClass.not and: { aClass.class.findMethod(\initClass).notNil }, {
-					aClass.initClass;
```
JordanHendersonMusic, line 43: This one can stay

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -170,6 +183,32 @@ Class {
...
+		var list = [];
+		subclasses.do { arg class;
+			if(class.instancesRespondTo(selector)) {
+				list = list.add(class);
+			};
+			list = list.addAll(class.findRespondingSubclasses(selector))
+		};
+		^list
+	}
+	findRespondingUpperSubclasses { arg selector;
```
JordanHendersonMusic, line 196: I'm confused by this name, could you explain it please

lang/LangPrimSource/PyrPrimitive.cpp
```diff
@@ -2279,6 +2279,61 @@ int prObjectRespondsTo(struct VMGlobals* g, int numArgsPushed) {
     return errNone;
 }
 
+int prInstancesOfClassRespondTo(struct VMGlobals* g, int numArgsPushed) {
+    PyrSlot *a, *b;
+    PyrClass* classobj;
+    PyrMethod* meth;
+    PyrSymbol* selector;
+    int index;
```
JordanHendersonMusic: I know this is the same as the surrounding function, but this style is generally agreed to be bad, could you declare the variables where you initialize them please.

HelpSource/Classes/String.schelp
telephon: I suppose not ...

HelpSource/Classes/String.schelp
telephon: We need something that is compatible with `order` and `at` passing indices. Dictionaries don't have the second. So it will be a sequenceable collection.

HelpSource/Classes/String.schelp
telephon: done

HelpSource/Classes/String.schelp
telephon: -> array

HelpSource/Classes/String.schelp
telephon: done,

telephon:
Actually I jsut saw that 

```
bestMatchIndices = bestMatchIndices.select { |i|
			maxEditDistance.isNil or: { editDistances[i] <= maxEditDistance }
			and:
			{ keepOnlySimilar.not or: { similarity(names[i], searchFor) > 0 } }
		}
```

is redundant, calculating the distance twice

SCClassLibrary/Common/Core/Kernel.sc
telephon: So, maybe I need to change the name. When traversing the subclasses, it finds the highest level of classes where a method is implemented. So, less structurally seen, it is the class that implements the method.

findRespondingImplementingClasses maybe?

lang/LangPrimSource/PyrPrimitive.cpp
telephon: done, I always keep the style the same, in this case I've changed it for both primitives.

SCClassLibrary/Common/Collections/String.sc
telephon: no clue how this got in.

SCClassLibrary/Common/Core/Kernel.sc
JordanHendersonMusic: One issue is that the parent class might implement it with 'subclass responsibility'. Perhaps the word 'declared' is better? It's a difficult one.

telephon:
> Why not just return all with a low edit distance? Why limit the count at all?

I just took this from the existing implementation, actually.

SCClassLibrary/Common/Core/Kernel.sc
telephon: Done.

SCClassLibrary/Common/Collections/String.sc
telephon: ah I see, automatic indentation. I've removed the change.

SCClassLibrary/Common/Core/Kernel.sc
telephon: hm, yes, this is a good point.

telephon: maybe this should be checked in the primitive?

telephon: But since this is a problem of `respondsTo` already, 
```
EZGui.new.respondsTo(\prSetViewParams) // true
```
... and the implementation may theoretically be complicated, we better leave it.

telephon:
> > Why not just return all with a low edit distance? Why limit the count at all?
> 
> I just took this from the existing implementation, actually.

It makes sense in so far as a typo can't be wrong on each character. I think we should fine tune this behaviour in practice. Maybe the similarity check is not needed for the error post, maybe the class suggestion is too rare to be worth it. Then we should remove it again fefore the next release.

SCClassLibrary/Common/Core/Kernel.sc
telephon: With regards to the name of the two methods:

- `findAllRespondingSubclasses`
- `findNearestRespondingSubclasses`

@JordanHendersonMusic What do you think?

JordanHendersonMusic:
> > > Why not just return all with a low edit distance? Why limit the count at all?

> > I just took this from the existing implementation, actually.

> It makes sense in so far as a typo can't be wrong on each character. I think we should fine tune this behaviour in practice.

What I'm thinking is that, if we limit it to an edit distance of 3 (or whatever), and there are 10 responses, there is no guarantee it would be in the first n. Since they are all equally likely, why not print them all, it is unlikely that this list would be too long. 

---

I think it might be more useful for the similar class suggestion to only print the most abstract class where it is first declared. I.e., it tells you what *kind* of things it should be, not the exact class.

Here is an example.

```
View.new.lastForWhich

Closest matching classes found for the selector:
	Set, List, Event, Pair
```

The suggestion only tells us classes where `lastForWhich` has an implementation, that also doesn't have a subclass. I think it would be better to just return an error message saying something along the lines of: 
```
View.new.lastForWhich

View does not implement 'lastForWhich', but instances derived from the following do:
	Collection
```

See, the current suggestion misses out the most common collection, `Array`.

telephon:
> See, the current suggestion misses out the most common collection, `Array`.
> 
> This could be implemented by walking down from Object's subclasses until you find a class that has the method implemented, then you no longer need to check its subclasses.

It uses `findRespondingSubclasses` and then looks for names that aare similar. The point is that we still assume that there was a typo, not some other kind of error. Maybe this is a wrong assumption.

```supercollider
Object.findRespondingSubclasses(\lastForWhich).collect(_.name)
// returns
[Collection, Set, Dictionary, IdentityDictionary, LabelsDictionary, Environment, PLbindefEnvironment, Event, PLbindefParEnvironment, TempPLbindefParEnvironment, FuzzyDictionary, FuzzySet, NodeMap, ProxyNodeMap, TwoWayDictionary, HierSchDict, IdentitySet, OrderedIdentitySet, Range, MultiLevelIdentityDictionary, LibraryBase, Archive, Library, Halo, Interval, TwoWayIdentityDictionary, ObjectTable, Array2D, Pair, Bag, IdentityBag, SequenceableCollection, RingBuffer, LinkedList, Order, SparseArray, List, SortedL...etc...
```

JordanHendersonMusic:
Ah no I see why the limit is as it is. 

---

Yes, I don't think printing every class is useful, only the most abstract class, because the list will be huge and include other abstract classes. In the case of lastOfWhitch, `Collection` seems like the correct thing to print.

```
View.new.lastForWhich

Closest matching classes found for the selector:
	Set, List, Event, Pair
```
The above seems to imply that only set, list, event and pair support that method, that isn't true, and the user probably meant to pass an array.

In the description example, I don't think it should print QView, this would also accomplish that.

SCClassLibrary/Common/Core/Kernel.sc
JordanHendersonMusic: Nearest to what though? Nearest to object, how about, find most abstract responding classes?

Likewise, why not, find all responding classes? Why use the word subclass?

JordanHendersonMusic:
Right I see why I was confused. There are (at least) four possibilities:

1. Selector incorrectly spelt.
2. Class name incorrectly spelt
3. Wrong class provided, but selector valid.
4. There is also wrong class and invalid selector.

The current implementation provides info assuming 2. I don't think that is very common, because you must have spelt a valid class for it to compile: `Obhect` is a syntax error.

I think the implementation should print info for 3 instead, in which case, using the 'upperclasses' method makes most sense.

We could walk both graphs for 4, but I'm happy to leave it.

telephon:
Yes, exactly. So let's test for a while the combination 1 + 3 and then decide what we find appropriate. I'll make the chenges.

JordanHendersonMusic:
### 1
`123.sd`
```
^^ ERROR: Message 'sd' not understood.
RECEIVER: 123

Messages understood by the receiver with a similar name:
	+
	-
	*
	odd




```


### 2
`Object().pow(2)`
```
^^ ERROR: Message 'pow' not understood.
RECEIVER: an Object

Messages understood by the receiver with a similar name:
	post
	do
	poll

Objects which understand the selector 'pow' derive from:
	Number
	Symbol
	AbstractFunction
	SequenceableCollection




````


### 3
`Point().sign`
```
^^ ERROR: Message 'sign' not understood.
RECEIVER: Point(0, 0)

Messages understood by the receiver with a similar name:
	size
	fin

Objects which understand the selector 'sign' derive from:
	SimpleNumber
	Symbol
	AbstractFunction
	SequenceableCollection





```

---

I like this, it looks really good!

There seems to be quite a lot of extra whitespace at the bottom of these.


1's recommendations don't really make sense. I attempted to type in 'as' but missed by one space on the keyboard, obviously this is keyboard specific, but the operators seem more unlikely than the text — I doubt there is anything that can be done about that.

File: SCClassLibrary/Common/Core/Error.sc [hunk 12]
```diff
@@ -192,7 +202,9 @@ DoesNotUnderstandError : MethodError {
 		if(protectedBacktrace.notNil, { this.postProtectedBacktrace });
 		this.dumpBackTrace;
 		// this.adviceLink.postln;
-		"^^ %\nRECEIVER: %\n\n\n".postf(this.errorString, receiver);
+		"\n^^ %\nRECEIVER: %\n".postf(this.errorString, receiver);
+		suggestion.postln;
+		"\n\n\n".post;
```
JordanHendersonMusic: Do we need these new lines, I'd rather not have them if possible.

telephon: I'm reducing them. We keep two, so errors can still be separated visually when scrolling up.

JordanHendersonMusic:
How about this...

If the lowest edit distance has more than 4, print all the suggestions.
Otherwise print only the first 4.

This way, the most likely is always represented.

telephon:
You mean for method suggestions, right? For classes, we don't sort, because we don't assume a typo.

telephon:
Currently, I print all that are similar > 0.5, there seem to be not too many.

SCClassLibrary/Common/Core/Error.sc
```diff
@@ -157,27 +155,35 @@ DoesNotUnderstandError : MethodError {
...
+			if(methodSuggestions.notEmpty) {
+				plural = if(methodSuggestions.size > 1) { "s" } { "" };
+				methodSuggestions = methodSuggestions.join("\n\t");
+				suggestion = suggestion ++
+				"\nMessage% with a similar name understood by the receiver:\n\t%\n".format(plural, methodSuggestions);
+			};
+			classSuggestions = Object.findRespondingUpperSubclasses(selector).collect(_.name);
+			if(classSuggestions.notEmpty) {
+				if(classSuggestions.size < 8) {
+				classSuggestions = classSuggestions.join("\n\t");
```
JordanHendersonMusic: Indentation is off here

JordanHendersonMusic (changes requested):
Okay I get not wanting to over complicate things, I think this pr is a big improvement anyway. 

Just one small thing...

JordanHendersonMusic (approved):

PR #7004, opened 2025-06-27
scsynth: improve audio thread priority setting on macOS

dyfer:
## Purpose and Motivation

I observed, that it is somewhat easy to create dropouts in scsynth when e.g. application windows are moved. That makes me think that maybe the thread priorities are not set appropriately.

I tried to leverage the `thread_set_priority_rt()` function for this. It has _slightly_ improved the situation on my system (M1pro). Test code below.

I'm not sure if this is the best sollution, or if the `computation` and `constraint` values should be adjusted further.

<details><summary>Example code I used for stressing scsynth</summary>

```supercollider
(
Server.scsynth;
s.options.memSize_(2 ** 20); // needed for the load test
s.options.hardwareBufferSize_(128);

~numSynths = 500;
s.waitForBoot{
	SynthDef(\sinVerb, { arg out = 0, amp = 0.01;
		Out.ar(out, GVerb.ar(SinOsc.ar(Rand(400, 700))) * amp);
	}).add;

	~allSynths ?? {~allSynths = []};
	~allSynths = ~allSynths.addAll(
		~numSynths.collect({
			s.sync;
			Synth(\sinVerb, [\amp, -60.dbamp]);
		})

	)
}
)

~allSynths.do(_.free)

// macOS 14, CPU Apple M1pro, 500 synths
// this PR: peak around 90-100%, avg around 72%; no or minimal dropouts when moving windows
// 3.13: peak around 90-100%, avg around 85%; no dropouts when idle; dropouts when moving windows
```

</details>

Regarding the implementation: note that I had to move the thread creation to after the driver was initialized - this was needed in order to get the samplerate, which is needed to set the deadlines.

EDIT:
I have also tested this on a 2015 MacBook Air (Intel). 
For 100 synths from the example, SC 3.13 produced constant dropouts; this build produced only minimal and occasional dropouts.

EDIT 2:
~~Note that before SC 3.14 we were inadvertently dropping the thread priority for the audio thread...~~ We were dropping the thread NRT thread, apparently. See https://github.com/supercollider/supercollider/pull/6043. 

## Types of changes

- New feature (?)

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/server/scsynth/SC_CoreAudio.cpp
+++ b/server/scsynth/SC_CoreAudio.cpp
@@ -107,7 +107,15 @@ static void syncOSCOffsetWithTimeOfDay() {
 }
 
 static void resyncThreadFunc() {
-    /* keep the default thread priority */
+// on macOS, indicate that the thread is important
+// on other platforms, keep the default thread priority
+#    ifdef __APPLE__
+    // on macOS, set the thread priority to user-interactive
+    int result = pthread_set_qos_class_self_np(QOS_CLASS_USER_INTERACTIVE, -10);
+    if (result != 0) {
+        scprintf("SC_AudioDriver: could not set resync thread priority to user-interactive: %s\n", strerror(result));
+    }
+#    endif
 
     while (true) {
         sleep(20);
@@ -339,8 +347,26 @@ SC_AudioDriver::~SC_AudioDriver() {
 }
 
 void SC_AudioDriver::RunThread() {
-    /* NB: on macOS we just keep the default thread priority */
-#ifdef NOVA_TT_PRIORITY_RT
+#ifdef __APPLE__
+    // on macOS, set the thread priority to user-interactive
+    int result = pthread_set_qos_class_self_np(QOS_CLASS_USER_INTERACTIVE, 0);
+    if (result != 0) {
+        scprintf("SC_AudioDriver: could not set thread priority to user-interactive: %s\n", strerror(result));
+    }
+#    ifdef NOVA_TT_PRIORITY_PERIOD_COMPUTATION_CONSTRAINT
+    int blockSize = mWorld->mBufLength;
+    uint32_t ns_per_block = 1e9 / mSampleRate * blockSize;
+    uint32_t computation = ns_per_block * 0.4;
+    uint32_t constraint = ns_per_block * 0.8;
+    // printf("Verifying values: mSampleRate = %f, blockSize = %d, period = %d, computation = %d, constraint = %d\n",
+    // mSampleRate, blockSize, ns_per_block, computation, constraint);
+
+    bool success = nova::thread_set_priority_rt(ns_per_block, computation, constraint, true);
+    if (!success) {
+        scprintf("SC_AudioDriver: could not set thread priority with period computation constraint\n");
+    }
+#    endif // NOVA_TT_PRIORITY_PERIOD_COMPUTATION_CONSTRAINT
+#elif defined(NOVA_TT_PRIORITY_RT)
     int priority = nova::thread_priority_interval_rt().first;
     nova::thread_set_priority_rt(priority);
 #endif
@@ -408,10 +434,6 @@ void SC_ScheduledEvent::Perform() {
 }
 
 bool SC_AudioDriver::Setup() {
-    mRunThreadFlag = true;
-    SC_Thread thread(std::bind(&SC_AudioDriver::RunThread, this));
-    mThread = std::move(thread);
-
     int numSamples;
     double sampleRate;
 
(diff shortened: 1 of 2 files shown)


telephon:
> Btw this PR also made a difference when testing on older intel machine. On the surface it seems like a desirable feature, but I'm not very familiar with the intricacies of scheduling etc. It would be great to get feedback from someone who has more experience with it, especially on macOS.

I haven't seen the issue on my intel mac so far, but I'll be running this branch for a while to see if I notice anything positive (or negative).

dyfer:
Thanks @telephon 
Just to be clear: the behavior that I observed is when reaching the max CPU load and slight dropouts start to appear. With this PR I observed the slight dropouts to be minimized, both on M1 and Intel chips.

dyfer:
Also see #7009 for a possible sclang timing improvement.

redFrik:
i did some tests and came to the conclusion that the SuperCollider 3.13.0 i had (Built from tag 'Version-3.13.0' [3188503]) was the problem.  it is nearly unusable on my M1 Pro Sonoma - lots of timing glitches - even under the lightest of loads.
test code was basically just a couple of Pbinds with server latency 0.05, sampleRate 48K, blockSize 64.

with SuperCollider 3.14.0-rc1 i cannot make the same code produce any late messages. also this PR and [#7009](https://github.com/supercollider/supercollider/pull/7009) feel stable.
with your extreme examples i notices differences, but they're subtle.

so my big issue all along was that 3.13.0 build.  thanks and sorry for diverting the discussion.

dyfer:
> with SuperCollider 3.14.0-rc1 i cannot make the same code produce any late messages

Yes, this was changed in https://github.com/supercollider/supercollider/pull/6043. However, that removed raising the priorities... which this PR tries addresses.

BTW I noticed that PR also removed raising the priority for the resync thread, I added it here.

server/scsynth/SC_CoreAudio.cpp
```diff
@@ -339,8 +347,26 @@ SC_AudioDriver::~SC_AudioDriver() {
...
+    }
+#    ifdef NOVA_TT_PRIORITY_PERIOD_COMPUTATION_CONSTRAINT
+    int blockSize = mWorld->mBufLength;
+    uint32_t ns_per_block = 1e9 / mSampleRate * blockSize;
+    uint32_t computation = ns_per_block * 0.4;
+    uint32_t constraint = ns_per_block * 0.8;
+    // printf("Verifying values: mSampleRate = %f, blockSize = %d, period = %d, computation = %d, constraint = %d\n",
+    // mSampleRate, blockSize, ns_per_block, computation, constraint);
+
+    bool success = nova::thread_set_priority_rt(ns_per_block, computation, constraint, true);
```
Spacechild1, line 364: *As this is the NRT thread*, we want a thread priority that is high, but still below the audio thread. This is what `int priority = nova::thread_priority_interval_rt().first;` attempts to do on Linux and Windows. We need to do a similar thing on macOS. AFAICT, your code essentially sets the same priority as the audio thread, which doesn't seem right.

Spacechild1:
@timblechmann How would you set thread priorities on macOS in 2025? We need a mid-high priority for the NRT and network thread and high priority for Supernova's DSP helper threads.

server/scsynth/SC_CoreAudio.cpp
dyfer: Thanks! Where's the audio thread? What's its priority?

~~The proposed implementation definitely improves scsynth behavior on macOS/M1 (less dropouts / more headroom) - I'm not saying it's right but I'm reinforcing that it did _something_.~~ Maybe not...

dyfer:
> No, this was about the _NRT thread_.

Sorry for my misunderstanding!

dyfer:
Hm... on testing today, I can't see the difference between the current develop build and this PR. Maybe I was chasing a red herring...

Spacechild1:
> Maybe I was chasing a red herring...

Yeah, I can't see how this PR could possible reduce audio dropouts, so the likely explanation is that it doesn't :)

dyfer:
Sorry for the noise, everybody. 
I was mistakenly comparing this against SC 3.13 (which _is_ more prone to dropouts on M1).

Closing for now.

server/scsynth/SC_CoreAudio.cpp
timblechmann: > Where's the audio thread? What's its priority?

on macos, the audio thread is managed by coreaudio. it does not have a "priority", but "time constraints", which coreaudio derives from the audio buffers size

PR #7009, opened 2025-06-28
sclang macos: use QoS for thread priorities

dyfer:
## Purpose and Motivation

This PR tries to implement thread priorities in **sclang** using a more modern API.

I observed subjectively that the proposed changes make the internal SC scheduling less prone to disruption when performing other tasks (switching applications, moving windows, bringing up Mission Control, etc).

I used the following test code borrowed from another issue:
```supercollider
(
s.waitForBoot({
	SynthDef(\imp, { OffsetOut.ar(0, Impulse.ar(0)); FreeSelf.kr(Impulse.kr(0) * -12.dbamp); }).add;
	s.sync;
	t = Task({
		10000.do({ arg i;
			Synth(\imp);
			0.05.wait;
		});
	});

	t.start;
})
)
t.stop;

s.quit
```
When running this code, I proceeded to move SC window around, switch applications, activate Mission Control, minimize windows etc etc. The irregularities are exacerbated when the background activity occurs, though seems to be much less so with this PR.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
Interesting, the changes in this PR seem to introduce some test failures, presumably due to different timing of messages. Needs more investigating...

lang/LangPrimSource/PyrSched.cpp
```diff
@@ -568,36 +500,7 @@ SCLANG_DLLEXPORT_C void schedRun() {
...
-
-    // param.sched_priority = 70; // you'll have to play with this to see what it does
-    // pthread_setschedparam (gSchedThread, policy, &param);
-
-    pthread_getschedparam(gSchedThread.native_handle(), &policy, &param);
-
-    // post("param.sched_priority %d\n", param.sched_priority);
-#endif // __APPLE__
-
+    // NOTE: on macOS, we set the thread priority in schedRunFunc using the QoS API.
```
Spacechild1, line 503: Why not also on Linux (with `pthread_self()`)? IMO it would be better if the thread priority code were in the same place.

dyfer:
> But what has changed between SC 3.13 and SC 3.14 that could explain the difference in behavior?

https://github.com/supercollider/supercollider/pull/6043

I think this was all a red herring - I was comparing my experiments against 3.13, instead of 3.14/develop. Closing for now.

Issue #7010, opened 2025-06-29
Redefining nil Usage and Sentinel Handling in SuperCollider


prko:
## Motivation

While discussing [this SuperCollider forum topic](https://scsynth.org/t/set-and-nil/11913), I noticed that the 'nil' object in SuperCollider has multiple functionalities: it indicates uninitialised data, signifies invalid values, marks terminal states (e.g. end-of-stream) and acts as a sentinel for empty slots in collections such as 'Set' and 'Dictionary'. This multipurpose usage increases ambiguity and introduces silent errors or inconsistent behaviour when working with collections and optional values.

To improve clarity, debugging and semantic consistency, I believe that 'nil' should exclusively represent uninitialised or absent values and should no longer function as a sentinel or error state indicator.

## Description of Proposed Feature

According to the documentation, Nil is a class with a single instance, 'nil', which is used for

- uninitialised data
- bad or invalid values
- terminal values (e.g. end-of-stream markers).

Additionally, as noted in this forum reply, 'nil' is used as an internal sentinel value to denote empty slots in hash-based collections such as 'Set' and 'Dictionary', and their subclasses. This ambiguous use of 'nil' introduces the following issues:

- Developers cannot store 'nil' as a legitimate value in 'Set' (this results in a runtime error).
- In Dictionary, assigning nil deletes a key, even if the user intends to store it explicitly.
- APIs return 'nil' for missing values, but users cannot always distinguish this from 'nil' that has been stored.

These cases reduce the transparency of the system and force developers to write defensive code to differentiate between 'absent', 'intentionally empty', and 'deleted'.

I propose revisiting the semantics of 'nil' to ensure the separation of intent is clear.

- 'Nil' should be used solely to indicate an 'uninitialised/absent value'.
- Introduce internal-only sentinel objects (e.g. UndefinedSlot) for internal slots.
- Use explicit error signalling (e.g. exceptions or Either-like wrappers) for failed lookups.

## Plan for Implementation

This seems to be a task that I cannot complete correctly. However, I have the following plan:

Currently:

- nil is used internally in Set, Dictionary, etc., as the sentinel for unoccupied slots.
- Functions often return nil to indicate failure or the absence of values (e.g. findMatch, asClass).
- In a dictionary, assigning nil deletes the key rather than storing nil.

Proposed plan:

[...]

PR #7011, opened 2025-06-29
Add nil's behaviour in Set and Nil documentation

prko:
## Purpose and Motivation
This documentation update reflects the current behaviour of nil in SuperCollider, which serves multiple purposes: representing uninitialised values, terminal markers, and acting as an internal sentinel within collections. While this multipurpose usage may be reconsidered in the future, the current update ensures the existing implementation is accurately described.
If nil is later restricted to represent only uninitialised values, the affected documentation should be revised accordingly. See #7010.

## Types of changes

- Documentation
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Nil.schelp
+++ b/HelpSource/Classes/Nil.schelp
@@ -4,17 +4,23 @@ summary::Represents uninitialized data
 
 description::
 
-Nil has a single instance named nil and is used to represent uninitialized data,
+Nil has a single instance named code::nil:: and is used to represent uninitialized data,
 bad values, or terminal values such as end-of-stream.
+Internally, code::nil:: also serves as a sentinel value, used to mark empty or unoccupied slots
+in hash-based data structures such as code::Set::, and its subclasses.
+Because of this internal role, code::nil::—despite being a valid object—cannot be stored
+as an element in code::Set:: or any of its subclasses.
+Attempting to do so will result in a runtime error.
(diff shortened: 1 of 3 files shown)


telephon (changes requested):
"Set class::" here is a typo.

prko:
>"Set class::" here is a typo.
>Was it? I think the uppercase is correct here

We should distinguish the following:
- `Set`: 
This refers to the class itself.
- `a set`: 
This refers to an instance of the set class. This can also be phrased as `an instance of the class code::Set::` in Set documentation file or `an instance of the class link::Classes/Set::`.
- `a Set`:
This should generally be avoided, as it blends object and class references; phrasing such as a set or an instance of Set is clearer and more idiomatic in documentation.

Please let me know if this is incorrect. This type of summary might be more appropriate in a different section.

HotwheelsSisyphus:
Yep, that is exactly what I meant. I don't think the summary you posted needs to go into the docs, I think it will be clearer now with the new capitalization (the old version of the help file was really not great in that respect, so thanks for these improvements)!
I can't comment on the content regarding nil etc., as I haven't followed that discussion enough; but as far as Intelligibility etc. are concerned, this looks good to me.

dyfer:
@telephon have you concerns been addressed?

telephon (approved):

PR #6900, opened 2025-05-27
Disable obsolete audio APIs on Windows

dyfer:
## Purpose and Motivation

WASAPI should be the default non-ASIO option on Windows. This PR removes MME, DirectSound, and WDM-KS APIs from the Windows build.

Are we ready to do this? Is there any benefit for using the obsolete APIs on Windows?

I also updated the docs, including a note on WASAPI resampling that reflects the change in #6899.

## Types of changes

- Documentation
- ~~Bug fix (if counting using MME as the default device as a bug...)~~ as of SC 3.14, we are already using WASAPI for default device selection
- New feature (as in - less superfluous posting when booting the server)
- Breaking change (if someone was explicitly using a MME/DS/WDM-KS device)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #6775, opened 2025-04-26
Update SC_PortAudio.cpp

prko:
## Purpose and Motivation
On Windows, each item in the device list is excessively long when presented in a single line, which significantly reduces readability
```
-> localhost
Booting server 'localhost' on address 127.0.0.1:57110.

Device options:
  - MME : Microsoft Sound Mapper - Input   (device #0 with 2 ins 0 outs)
  - MME : Microphone (High Definition Aud   (device #1 with 2 ins 0 outs)
  - MME : Microsoft Sound Mapper - Output   (device #2 with 0 ins 2 outs)
  - MME : Speakers (High Definition Audio   (device #3 with 0 ins 8 outs)
  - Windows DirectSound : Primary Sound Capture Driver   (device #4 with 2 ins 0 outs)
  - Windows DirectSound : Microphone (High Definition Audio Device)   (device #5 with 2 ins 0 outs)
  - Windows DirectSound : Primary Sound Driver   (device #6 with 0 ins 2 outs)
  - Windows DirectSound : Speakers (High Definition Audio Device)   (device #7 with 0 ins 8 outs)
  - ASIO : Magix Low Latency 2016   (device #8 with 2 ins 8 outs)
  - ASIO : ReaRoute ASIO (x64)   (device #9 with 16 ins 16 outs)
  - Windows WASAPI : Speakers (High Definition Audio Device)   (device #10 with 0 ins 2 outs)
  - Windows WASAPI : Microphone (High Definition Audio Device)   (device #11 with 2 ins 0 outs)
  - Windows WDM-KS : Microphone (HD Audio Microphone)   (device #12 with 2 ins 0 outs)
  - Windows WDM-KS : Speakers (HD Audio Speaker)   (device #13 with 0 ins 8 outs)

Requested devices:
  In:
  - (default)
  Out:
  - (default)

Selecting default system input/output devices

Booting with:
  In: MME : Microphone (High Definition Aud
  Out: MME : Speakers (High Definition Audio
  Sample rate: 44100.000
  Latency (in/out): 0.013 / 0.091 sec
SC_AudioDriver: sample rate = 44100.000000, driver's block size = 64
SuperCollider 3 server ready.
Requested notification messages from server 'localhost'
localhost: server process's maxLogins (1) matches with my options.
localhost: keeping clientID (0) as confirmed by server process.
Shared memory server interface initialized
```
This PR changes this as follows:
```
-> localhost
Booting server 'localhost' on address 127.0.0.1:57110.

Device options:
[...]

PR #7018, opened 2025-06-29
Improve posting when booting the server

prko:
## Purpose and Motivation
On Windows and Linux, sclang does not display the index and quotation mark for audio device names.
```
Exception in World_New: boost::interprocess::bad_alloc

Device options:
- MME : Microsoft 사운드 매퍼 - Input
  (2 ins, 0 outs)
- MME : Microphone(High Definition Audi
  (2 ins, 0 outs)
- MME : Microsoft 사운드 매퍼 - Output
  (0 ins, 2 outs)
- MME : Speakers(High Definition Audio 
  (0 ins, 8 outs)
- Windows DirectSound : 주 사운드 캡처 드라이버
  (2 ins, 0 outs)
- Windows DirectSound : Microphone(High Definition Audio Device)
  (2 ins, 0 outs)
- Windows DirectSound : 주 사운드 드라이버
  (0 ins, 2 outs)
- Windows DirectSound : Speakers(High Definition Audio Device)
  (0 ins, 8 outs)
- ASIO : Magix Low Latency 2016
  (2 ins, 8 outs)
- ASIO : ReaRoute ASIO (x64)
  (16 ins, 16 outs)
- Windows WASAPI : Speakers(High Definition Audio Device)
  (0 ins, 2 outs)
- Windows WASAPI : Microphone(High Definition Audio Device)
  (2 ins, 0 outs)
- Windows WDM-KS : Microphone (HD Audio Microphone)
  (2 ins, 0 outs)
- Windows WDM-KS : Speakers (HD Audio Speaker)
  (0 ins, 8 outs)

Requested devices:
  In:
  - (default)
  Out:
  - (default)

Selecting default system input/output devices

Booting with:
  In: Windows WASAPI : Microphone(High Definition Audio Device)
  Out: Windows WASAPI : Speakers(High Definition Audio Device)
  Sample rate: 48000.000
  Latency (in/out): 0.022 / 0.024 sec
SC_AudioDriver: sample rate = 48000.000000, driver's block size = 64
SuperCollider 3 server ready.
localhost: server process's maxLogins (1) matches with my options.
Server: possible boot delay.
On some Windows-based machines, Windows Defender sometimes delays server boot by one minute.
You can add scsynth.exe and/or supernova.exe *processes* to Windows Defender exclusion list to avoid this delay. It's safe.
```
```
jackdmp 1.9.21
Copyright 2001-2005 Paul Davis and others.
Copyright 2004-2016 Grame.
Copyright 2016-2022 Filipe Coelho.
jackdmp comes with ABSOLUTELY NO WARRANTY
This is free software, and you are welcome to redistribute it
under certain conditions; see the file COPYING for details
JACK server starting in realtime mode with priority 10
self-connect-mode is "Don't restrict self connect requests"
audio_reservation_init
Acquire audio card Audio0
creating alsa driver ... hw:0|hw:0|1024|2|48000|0|0|nomon|swmeter|-|32bit
configuring for 48000Hz, period = 1024 frames (21.3 ms), buffer = 2 periods
ALSA: final selected sample format for capture: 32bit integer little-endian
ALSA: use 2 periods for capture
ALSA: final selected sample format for playback: 32bit integer little-endian
ALSA: use 2 periods for playback
JackDriver: client name is 'SuperCollider'
JackDriver: connected  system:capture_1 to SuperCollider:in_1
JackDriver: connected  system:capture_2 to SuperCollider:in_2
JackDriver: connected  SuperCollider:out_1 to system:playback_1
JackDriver: connected  SuperCollider:out_2 to system:playback_2
SuperCollider 3 server ready.
JackDriver: max output latency 42.7 ms
localhost: server process's maxLogins (1) matches with my options.
```

On macOS, sclang displays them:
```
VSTPlugin 0.6-pre1
Found 0 LADSPA plugins
Number of Devices: 10
   0 : "LG TV SSCR"
   1 : "Live Gamer Ultra 2.1-Audio"
   2 : "CalDigit Thunderbolt 3 Audio"
   3 : "CalDigit Thunderbolt 3 Audio"
   4 : "BlackHole 2ch"
   5 : "MacBook Pro Microphone"
   6 : "MacBook Pro Speakers"
   7 : "ZoomAudioD"
   8 : "Aggregate Device"
   9 : "Multi-Output Device"

   Streams: 1
      0  channels 1

"MacBook Pro Speakers" Output Device
   Streams: 1
      0  channels 2

SC_AudioDriver: sample rate = 44100.000000, driver's block size = 512
SuperCollider 3 server ready.
localhost: server process's maxLogins (1) matches with my options.
```
I think the quotation mark is convenient when copy and pasting the device name, but it would be better to unify the formatting.  
This PR removes index and quotation marks for audio device when booting server.
I would add the channel numbers, but I could not find such things in code.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation


dyfer:
If we are to unify formatting, I'd myself rather put the quotation marks around device names in the portaudio backend (to indicate that backend name is part of the name), but I think it would be better to poll more users about it, as this is a personal preference. I'd ask similarly whether ppl want to include device numbers or not (I don't feel strongly either way) - these are not used for anything anyway.

prko:
Would you be open to reconsidering the idea of conducting a survey?
- In PR #6775, I suggested displaying the device index, but I ultimately agreed with your view that including the index may not be ideal.
- Using quotation marks to indicate the device name is indeed a convenient and user-friendly approach.

Therefore, I think it makes sense to include quotation marks in the device name while omitting the device index.

One more thing:
I believe it would be helpful to indicate each device’s input and output channel numbers. For reference, when booting Supernova on macOS, sclang posts the following information:
```
Supernova booting
Available Audio Devices:
- LG TV SSCR
  (0 ins, 2 outs)
- Live Gamer Ultra 2.1-Audio
  (4 ins, 0 outs)
- CalDigit Thunderbolt 3 Audio
  (0 ins, 2 outs)
- CalDigit Thunderbolt 3 Audio
  (2 ins, 0 outs)
- BlackHole 2ch
  (2 ins, 2 outs)
- MacBook Pro Microphone
  (1 ins, 0 outs)
- MacBook Pro Speakers
  (0 ins, 2 outs)
- ZoomAudioDevice
  (2 ins, 2 outs)
- Aggregate Device
  (5 ins, 2 outs)
- Multi-Output Device
  (0 ins, 2 outs)

VSTPlugin 0.6-pre1
Found 0 LADSPA plugins
Requested audio devices:
  In: Fireface 800 (3F0)
  Out: Fireface 800 (3F0)

Selecting default system input/output devices

Opening audio devices:
  In: MacBook Pro Microphone
  Out: MacBook Pro Speakers
  Sample rate: 44100.000
  Latency (in/out): 0.249 / 0.031 sec
Audio devices opened successfully
Supernova ready
```

prko:
When unifying the server boot message format, we should also take into account the following variations across platforms:

- **Windows**:  
  `Device options:`

- **macOS**:  
  `Number of Devices:  %d`

- **Supernova**:  
  `Available Audio Devices:`

Additionally, while the message `Supernova booting` is shown during Supernova startup, there is currently no equivalent message when booting `scsynth`. I have not yet identified the most appropriate location to display a `Scsynth booting` message.

I have been looking for a suitable survey platform to gather user preferences. The most straightforward approach would be to allow participants to select their preferred formatting. However, I have not been able to find any survey tool that supports **multi-line checkbox options**.

capital-G:
I don't know if the different formatting style creates problems for any users, but if we really want to have this uniform we could/should(?) probably refactor the printing code into a dedicated class in order to reduce the maintenance over time.

dyfer:
First off - @prko thank you for your diligence and care about SC!

Feel free to discuss/review/merge, but I feel that for me there are more pressing issues to attend to. 

Also, I _personally_ tend to like the slight differences between scsynth and supernova - it's easier to distinguish them when booting. But maybe that's not important.

prko:
> I don't know if the different formatting style creates problems for any users, 

For general purposes, I don't believe they would cause significant problems.
However, when presenting concepts in a classroom setting—  
– especially where most students are new to the field,  
– and where some are using macOS while others are on Windows—  

it would be preferable for the post window output to be consistent across platforms and sound engines, as this would help reduce the time needed to explain variations specific to each system.

Such consistency would also make tutorial development much smoother. If each platform or engine outputs slightly different information, these discrepancies must be addressed, which inevitably lengthens the explanatory text unnecessarily.

---

>Also, I personally tend to like the slight differences between scsynth and supernova - it's easier to distinguish them when booting. But maybe that's not important.

The distinction between Supernova and Scsynth is valuable. That said, simply including a line such as `Supernova booting` or `Scsynth booting` at the very top should suffice, and the rest of the output could be standardised accordingly.

---

Initially, my aim was merely to remove the device index. However, unifying the output might not be a bad idea.

> ... if we really want to have this uniform we could/should(?) probably refactor the printing code into a dedicated class in order to reduce the maintenance over time.

Creating a dedicated class could be a valid solution, though it seems to require considerable effort:  
– removing the corresponding code from the C++ layer,  
– developing a new class,  
– and handling related structural changes.

Simply modifying the existing code would be a more time-efficient solution. I’ll explore whether it’s possible to construct the channel numbers for booting devices on macOS. Alternatively, removing channel numbers from other platforms and from Supernova could also be a viable option. If that proves feasible, I’ll proceed with that approach.

However, if creating a dedicated class is required, it may not be something I can undertake on my own at this time, as I currently lack experience in that area.

dyfer:
> Creating a dedicated class could be a valid solution, though it seems to require considerable effort:
> – removing the corresponding code from the C++ layer,

I'm pretty sure @capital-G meant a C++ class. We can't do this in sclang.

adcxyz:
> First off - @prko thank you for your diligence and care about SC!

fully agree! 

Personally, I find the supernova post on macOS quite good as is, 
and it could be improved with a few small additions; 
then it would be the model toward which to unify the others.
```
Supernova booting
Available Audio Devices:
0: BlackHole 16ch (16 ins, 16 outs)
1: BlackHole 64ch (64 ins, 64 outs)
2: MacBook Pro Microphone (1 ins, 0 outs)
3: MacBook Pro Speakers (0 ins, 2 outs)

Found 0 LADSPA plugins

Requested audio devices: 
  In: (default)
  Out: (default)

Selecting default system input/output devices

Opening audio devices:
  In: MacBook Pro Microphone
  Out: MacBook Pro Speakers
  Sample rate: 44100.000
  Latency (in/out): 0.136 / 0.030 sec
```


My wishes here would be:
```
// make the boot line and server program more visible with bold:
*** Booting server 'localhost' on address 127.0.0.1:57110.
*** Supernova booting
// device names are directly usable as settings in code, so write them in "":
Available Audio Devices:
0: "BlackHole 16ch" (16 ins, 16 outs)
1: "BlackHole 64ch" (64 ins, 64 outs)
2: "MacBook Pro Microphone" (1 ins, 0 outs)
3: "MacBook Pro Speakers" (0 ins, 2 outs)

Found 0 LADSPA plugins
// useful to know
Requested audio devices: 
  In: (default)
  Out: (default)

Selecting default system input/output devices

// again write them as strings, and add number of channels 
// they are actually booted with:
Opening audio devices:
  In: "MacBook Pro Microphone", 1 in
  Out: "MacBook Pro Speakers", 2 outs
// and these are great to have! 
  Sample rate: 44100.000
  Latency (in/out): 0.136 / 0.030 sec
```

prko:
@adcxyz 
Thank you for your suggestion.
I tried to incorporate your feedback into commit https://github.com/supercollider/supercollider/pull/7018/commits/742a4b0a20c5d3581b77f3ec992dcbc41088d698 as much as I could.

However, I was unable to implement the following features:
- in and out latency
- requested device

As @capital-G mentioned, I believe implementing these features would require a dedicated C++ class.

Below is the current output shown in the post window on macOS:
```
Server.scsynth
s.quit
s.boot
////////////////
'/quit' message sent to server 'localhost'.
-> localhost
-> localhost


*** Booting server 'localhost' on address 127.0.0.1:57111.
VSTPlugin 0.6-pre1
Found 0 LADSPA plugins

*** Scsynth booting
10 Available Audio Devices:
- "LG TV SSCR"
  (0 ins, 2 outs)
- "Live Gamer Ultra 2.1-Audio"
  (4 ins, 0 outs)
- "CalDigit Thunderbolt 3 Audio"
  (0 ins, 2 outs)
- "CalDigit Thunderbolt 3 Audio"
  (2 ins, 0 outs)
- "BlackHole 2ch"
  (2 ins, 2 outs)
- "MacBook Pro Microphone"
  (1 ins, 0 outs)
- "MacBook Pro Speakers"
  (0 ins, 2 outs)
- "ZoomAudioD"
  (2 ins, 2 outs)
- "Aggregate Device"
  (5 ins, 2 outs)
- "Multi-Output Device"
  (0 ins, 2 outs)

Booting with:
  In: "MacBook Pro Microphone" (Streams: 1 , 0), 1 outs
  Out: "MacBook Pro Speakers" (Streams: 1 0), 2 ins)
  Sample rate: 44100.000000
  SC_AudioDriver's block size: 512
SuperCollider 3 server (scsynth) ready.

localhost: server process's maxLogins (1) matches with my options.
```
```
Server.supernova
s.quit
s.boot
////////////////
'/quit' message sent to server 'localhost'.
-> localhost
-> localhost


*** Booting server 'localhost' on address 127.0.0.1:57111.

*** Supernova booting
10 Available Audio Devices:
- "LG TV SSCR"
  (0 ins, 2 outs)
- "Live Gamer Ultra 2.1-Audio"
  (4 ins, 0 outs)
- "CalDigit Thunderbolt 3 Audio"
  (0 ins, 2 outs)
- "CalDigit Thunderbolt 3 Audio"
  (2 ins, 0 outs)
- "BlackHole 2ch"
  (2 ins, 2 outs)
- "MacBook Pro Microphone"
  (1 ins, 0 outs)
- "MacBook Pro Speakers"
  (0 ins, 2 outs)
- "ZoomAudioDevice"
  (2 ins, 2 outs)
- "Aggregate Device"
  (5 ins, 2 outs)
- "Multi-Output Device"
  (0 ins, 2 outs)

VSTPlugin 0.6-pre1
Found 0 LADSPA plugins
Requested audio devices:
  In: "Fireface 800 (3F0)"
  Out: "Fireface 800 (3F0)"

Selecting default system input/output devices

Booting with:
  In: "MacBook Pro Microphone", 2 in(s)
  Out: "MacBook Pro Speakers", 2 out(s)
  Sample rate: 44100.000
  Latency (in/out): 0.249 / 0.031 sec
Audio devices opened successfully, and server booted successfully.
SuperCollider 3 server (supernova) ready.

localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
node id 2 already in use
node id 2 already in use
node id 2 already in use
node id 2 already in use
node id 1 already in use
node id 1 already in use
node id 1 already in use
node id 1 already in use
```

dyfer:
@prko please set up automated formatting in your environment - probably using [pre-commit](https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit) will be the easiest.

telephon:
> * Since device numbers are not used anywhere programmatically, I'd have a slight preference for just a dash (`- "MacBook microphone"`)

Why not leave out the dash, then you can just copy the line and paste it into your code.

dyfer:
> Why not leave out the dash, then you can just copy the line and paste it into your code.

Having some kind of list delimiter is better for readability, IMO. OTOH, it's easier to just copy the line when there are no other characters... but if we do include quotation marks, that identifying what to copy exactly.

### Current posting (before this PR)

```
Supernova booting
Available Audio Devices:
- BlackHole 16ch
  (16 ins, 16 outs)
- BlackHole 2ch
  (2 ins, 2 outs)
- BlackHole 64ch
  (64 ins, 64 outs)
- MacBook Pro Microphone
  (1 ins, 0 outs)
- MacBook Pro Speakers
  (0 ins, 2 outs)

VSTPlugin 0.6
Found 0 LADSPA plugins
Requested audio devices:
...
```
### Quotations added
```
Supernova booting
Available Audio Devices:
- "BlackHole 16ch"
  (16 ins, 16 outs)
- "BlackHole 2ch"
  (2 ins, 2 outs)
- "BlackHole 64ch"
  (64 ins, 64 outs)
- "MacBook Pro Microphone"
  (1 ins, 0 outs)
- "MacBook Pro Speakers"
  (0 ins, 2 outs)

VSTPlugin 0.6
Found 0 LADSPA plugins
Requested audio devices:
...
```
### Quotations and indices added
```
Supernova booting
Available Audio Devices:
0: "BlackHole 16ch"
   (16 ins, 16 outs)
1: "BlackHole 2ch"
   (2 ins, 2 outs)
2: "BlackHole 64ch"
   (64 ins, 64 outs)
3: "MacBook Pro Microphone"
   (1 ins, 0 outs)
4: "MacBook Pro Speakers"
   (0 ins, 2 outs)

VSTPlugin 0.6
Found 0 LADSPA plugins
Requested audio devices:
...
```
### Quotations added, no list delimiter or index
```
Supernova booting
Available Audio Devices:
"BlackHole 16ch"
  (16 ins, 16 outs)
"BlackHole 2ch"
  (2 ins, 2 outs)
"BlackHole 64ch"
  (64 ins, 64 outs)
"MacBook Pro Microphone"
  (1 ins, 0 outs)
"MacBook Pro Speakers"
  (0 ins, 2 outs)

VSTPlugin 0.6
Found 0 LADSPA plugins
Requested audio devices:
...
```
### Quotations added, dash, ins/outs in the same line
```
Supernova booting
Available Audio Devices:
- "BlackHole 16ch" (16 ins, 16 outs)
- "BlackHole 2ch" (2 ins, 2 outs)
- "BlackHole 64ch" (64 ins, 64 outs)
- "MacBook Pro Microphone" (1 ins, 0 outs)
- "MacBook Pro Speakers" (0 ins, 2 outs)

VSTPlugin 0.6
Found 0 LADSPA plugins
Requested audio devices:
...
```

The last one would be my personal preference.

dyfer:
> This is good on macOS... On Windows:

I like it on both platforms. Windows names can be nominally longer than a single line anyway (when the post window is smaller on the side) so they would be cut or wrapped regardless whether the ins/outs are in the same line or not.

I really prefer the fewer lines in the output.

---

BTW, I wonder if we should consider another option, after all:

### Quotations added, no list delimiter or index, ins/outs in the same line

```
Supernova booting
Available Audio Devices:
"BlackHole 16ch" (16 ins, 16 outs)
"BlackHole 2ch" (2 ins, 2 outs)
"BlackHole 64ch" (64 ins, 64 outs)
"MacBook Pro Microphone" (1 ins, 0 outs)
"MacBook Pro Speakers" (0 ins, 2 outs)

VSTPlugin 0.6
Found 0 LADSPA plugins
Requested audio devices:
...
```

On Windows that would save a couple of characters from the long device name - not a huge difference, but worth noting:

```
"ASIO : Magix Low Latency 2016" (2 ins, 8 outs)
"ASIO : ReaRoute ASIO (x64)" (16 ins, 16 outs)
"Windows WASAPI : Speakers(High Definition Audio Device)" (0 ins, 2 outs)
"Windows WASAPI : Microphone(High Definition Audio Device)" (2 ins, 0 outs)
```

(I removed MME/DS/WDM devices from the list, as we'll likely remove them eventually)

prko:
OK!

I have realised the following parts of  the audio device names on Windows are defined in external libraries:

- Windows WDM-KS
https://github.com/PortAudio/portaudio/blob/147dd722548358763a8b649b3e4b41dfffbcfbb6/src/hostapi/wdmks/pa_win_wdmks.c#L3850
- Windows DirectSound: 
https://github.com/PortAudio/portaudio/blob/147dd722548358763a8b649b3e4b41dfffbcfbb6/src/hostapi/dsound/pa_win_ds.c#L1217
- Windows WASAPI:
https://github.com/PortAudio/portaudio/blob/147dd722548358763a8b649b3e4b41dfffbcfbb6/src/hostapi/wasapi/pa_win_wasapi.c#L2352

Is it possible to remove the word "Windows" from the SuperCollider package?

for example:

From
```
        if (mDeviceInOut[0] != paNoDevice) {
            // avoid to allocate the 128 virtual channels reported by the portaudio library for ALSA "default"
            mInputChannelCount =
                std::min<size_t>(mWorld->mNumInputs, Pa_GetDeviceInfo(mDeviceInOut[0])->maxInputChannels);
            fprintf(stdout, "  In: \"%s\" %d in(s)\n", GetPaDeviceName(mDeviceInOut[0]).c_str(), Pa_GetDeviceInfo(mDeviceInOut[0])->maxInputChannels);
        } else {
            mInputChannelCount = 0;
        }
```

to
```
if (mDeviceInOut[0] != paNoDevice) {
    // avoid to allocate the 128 virtual channels reported by the portaudio library for ALSA "default"
    mInputChannelCount =
        std::min<size_t>(mWorld->mNumInputs, Pa_GetDeviceInfo(mDeviceInOut[0])->maxInputChannels);

    std::string deviceName = GetPaDeviceName(mDeviceInOut[0]);
    size_t pos = deviceName.find("Windows");
    if (pos != std::string::npos) {
        deviceName.erase(pos, std::string("Windows").length());
    }

    fprintf(stdout, "  In: \"%s\" %d in(s)\n", deviceName.c_str(), Pa_GetDeviceInfo(mDeviceInOut[0])->maxInputChannels);
} else {
    mInputChannelCount = 0;
}
```
The code is not tested. It is just an AI-suggestion.

If we could remove `Windows `, the device names are shorter:
```
- "DirectSound : Microphone(High Definition Audio Device)" (2 ins, 0 outs)
- "DirectSound : Speakers(High Definition Audio Device)" (0 ins, 8 outs)
- "WASAPI : Speakers(High Definition Audio Device)" (0 ins, 2 outs)
- "WASAPI : Microphone(High Definition Audio Device)" (2 ins, 0 outs)
- "WDM-KS : Microphone (HD Audio Microphone)" (2 ins, 0 outs)
- "WDM-KS : Speakers (HD Audio Speaker)" (0 ins, 8 outs)
```

dyfer:
> Is it possible to remove the word "Windows" from the SuperCollider package?

Not really, this is defined in PortAudio. Also, this would break backwards compatibility in some sense - a device previously named "Windows WASAPI : xxx" would now be "WASAPI : xxx", which would make programmatic device selection fail.

I don't think it's feasible to remove `Windows` from the backend name for the time being. It is technically possible, but the implementation would have to be somewhat hacky (string replacements), I think.

dyfer:
> Is it possible to remove the word "Windows" from the SuperCollider package?

BTW I gave this some more thought... I think we can consider it and make it at least partially backwards compatible. But let's revisit it at a later time, after 3.14 is out and maybe after merging #6900

prko:
I reported the followings:
https://github.com/PortAudio/portaudio/issues/1050
https://github.com/PortAudio/portaudio/issues/1058

prko:
The latest status:

(On macOS, there is `streams`.  Why does this only exist on macOS?)

## macOS
### scsynth
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Scsynth booting
8 Available Audio Devices:
- "BlackHole 16ch"
  (16 ins, 16 outs)
- "BlackHole 2ch"
  (2 ins, 2 outs)
- "BlackHole 64ch"
  (64 ins, 64 outs)
- "MacBook Pro Microphone"
  (1 ins, 0 outs)
- "MacBook Pro Speakers"
  (0 ins, 2 outs)
- "ZoomAudioD"
  (2 ins, 2 outs)
- "Aggregate Device"
  (1 ins, 2 outs)
- "Multi-Output Device"
  (0 ins, 2 outs)

Booting with:
  In: "MacBook Pro Microphone" (Streams: 1 , 0), 1 outs
  Out: "BlackHole 2ch" (Streams: 1 0), 2 ins)
  Sample rate: 44100.000000
  SC_AudioDriver's block size: 512
SuperCollider 3 server (scsynth) ready.

localhost: server process's maxLogins (1) matches with my options.
```
### supernova
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Supernova booting
8 Available Audio Devices:
- "BlackHole 16ch"
  (16 ins, 16 outs)
- "BlackHole 2ch"
  (2 ins, 2 outs)
- "BlackHole 64ch"
  (64 ins, 64 outs)
- "MacBook Pro Microphone"
  (1 ins, 0 outs)
- "MacBook Pro Speakers"
  (0 ins, 2 outs)
- "ZoomAudioDevice"
  (2 ins, 2 outs)
- "Aggregate Device"
  (1 ins, 2 outs)
- "Multi-Output Device"
  (0 ins, 2 outs)

Requested audio devices:
  In: "(default)"
  Out: "(default)"

Selecting default system input/output devices

Booting with:
  In: "MacBook Pro Microphone", 3 in(s)
  Out: "BlackHole 2ch", 4 out(s)
  Sample rate: 44100.000
  Latency (in/out): 0.249 / 0.013 sec
Audio devices opened successfully, and server booted successfully.
SuperCollider 3 server (supernova) ready.

localhost: no maxLogins info from server process.
```
## Windows
### scsynth
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Scsynth booting
3 Available Audio Devices:
- "ASIO : Magix Low Latency 2016"
  (2 ins, 8 outs)
- "ASIO : ReaRoute ASIO (x64)"
  (16 ins, 16 outs)
- "Windows WASAPI : Speakers(High Definition Audio Device)"
  (0 ins, 2 outs)

Requested devices:
  In:
  - (default)
  Out:
  - (default)

Selecting default system output device
WARNING: no input device selected, input is disabled.

Booting with:
  Out: "Windows WASAPI : Speakers(High Definition Audio Device)" 2 out
  Sample rate: 48000.000
  Latency (in/out): 0.000 / 0.022 sec
  Sample rate: 48000.000000
  SC_AudioDriver's block size: 64
SuperCollider 3 server (scsynth) ready.

localhost: server process's maxLogins (1) matches with my options.
```
### Supernova
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Supernova booting
3 Available Audio Devices:
- "ASIO : Magix Low Latency 2016"
  (2 ins, 8 outs)
- "ASIO : ReaRoute ASIO (x64)"
  (16 ins, 16 outs)
- "Windows WASAPI : Speakers(High Definition Audio Device)"
  (0 ins, 2 outs)

Requested audio devices:
  In: "(default)"
  Out: "(default)"

Selecting default system output device
WARNING: no input device selected, input is disabled.

Booting with:
```
## Linux
### scsynth
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.
SC FFT global init: cosTable initialised.
JackDriver: client name is 'SuperCollider'
  Sample rate: 48000.000000
  SC_AudioDriver's block size: 1024
JackDriver: connected  Built-in Audio Analog Stereo:capture_FL to SuperCollider:in_1
JackDriver: connected  Built-in Audio Analog Stereo:capture_FR to SuperCollider:in_2
JackDriver: connected  SuperCollider:out_1 to Built-in Audio Analog Stereo:playback_FL
JackDriver: connected  SuperCollider:out_2 to Built-in Audio Analog Stereo:playback_FR
SuperCollider 3 server (scsynth) ready.

localhost: server process's maxLogins (1) matches with my options.
```
### supernova
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.
SC FFT global init: cosTable initialised.

*** Supernova booting
Warning: cannot raise thread priority
SuperCollider 3 server (supernova) ready.
Warning: cannot raise thread priority

localhost: no maxLogins info from server process.
```

dyfer:
Thanks for the update - can we close it then?

prko:
> Thanks for the update - can we close it then?

I’ve updated this PR to indicate that we should either bring it to a conclusion or have another developer open a new PR to address the issue.

Could you clarify how far you intend to go in closing this PR?
I’m happy for it to be closed if that’s your preference — I simply want to ensure that the underlying issue is properly resolved, either here or in a subsequent PR.

I also believe that unifying this behaviour would reduce unnecessary explanation time when teaching students, and it would help to save space when documenting SuperCollider behaviour. For these reasons, I consider this change both necessary and beneficial.

dyfer:
It would be helpful if you provided some opinion instead of presenting some data and essentially asking others to read into details and deduce what might need to be done.

To me, the printing is now very similar on macOS and Windows.
It is by necessity different on Linux, where we use Jack and don't get access to device setting. It may get closer if we implement a native pipewire backend, but that's a separate issue.

@prko, please clarify what you think is still needed here. Since you originally opened this issue, it makes sense for you to indicate whether it has been addressed or not.

prko:
The list style of the audio devices is now good and consistent across platforms and server programmes.

However, the next part is slightly inconsistent:

- macOS scsynth:
   ```
   Booting with:
     In: "MacBook Pro Microphone" (Streams: 1 , 0), 1 outs
     Out: "BlackHole 2ch" (Streams: 1 0), 2 ins)
     Sample rate: 44100.000000
     SC_AudioDriver's block size: 512
   SuperCollider 3 server (scsynth) ready.
   ```
- macOS Supernova:
   ```text
      Requested audio devices:
     In: "(default)"
     Out: "(default)"
   
   Selecting default system input/output devices
   
   Booting with:
     In: "MacBook Pro Microphone", 3 in(s)
     Out: "BlackHole 2ch", 4 out(s)
     Sample rate: 44100.000
     Latency (in/out): 0.249 / 0.013 sec
   Audio devices opened successfully, and server booted successfully.
   SuperCollider 3 server (supernova) ready.
   ```
- Windows scsynth:
   ```text
   Requested devices:
     In:
     - (default)
     Out:
     - (default)
   
   Selecting default system output device
   WARNING: no input device selected, input is disabled.
   
   Booting with:
     Out: "Windows WASAPI : Speakers(High Definition Audio Device)" 2 out
     Sample rate: 48000.000
     Latency (in/out): 0.000 / 0.022 sec
     Sample rate: 48000.000000
     SC_AudioDriver's block size: 64
   SuperCollider 3 server (scsynth) ready.
   ```
- Wiindows supernova
   ```text
   Requested audio devices:
     In: "(default)"
     Out: "(default)"
   
   Selecting default system output device
   WARNING: no input device selected, input is disabled.
   
   Booting with:
   ```

Currently, Supernova on Windows does not boot, so I cannot compare its output precisely. However, I assume it is very similar to macOS Supernova and Windows scsynth.

So:

1. macOS scsynth uses a noticeably different format:
   - It does not include the `Requested audio devices:` section.
   - It does not show the `Selecting default system output device` line.
   - It does not report input/output latency.
   - It includes `(Streams: xx yy)`, which does not appear in macOS Supernova. I think `(Streams: xx yy)` could be omitted.

2. macOS Supernova does not report the block size.

---

What I would suggest is one of the following:

A. Merge this PR as it is, and then open a new PR to unify the differing formats mentioned above.
<- for me the simplest way if another developer opens a new pr

B. Continue working within this PR to unify the formats.
<-doable

C. Write a completely new implementation so that a dedicated C++ class handles all of this in a consistent way.
<- can try with ai assistance

dyfer:
> A. Merge this PR as it is, and then open a new PR to unify the differing formats mentioned above.

In general I think this PR should be wrapped up and merged soon.

> However, the next part is slightly inconsistent:

Yes. I still believe the user impact for this is minimal. 

A few specifics:

- I agree streams info on macOS can be omitted
- we could add "requested device" part to macOS - but I'm torn whether we really want that info at all; it is informative, but it leads to extra posting and data duplication (device name would appear a total of 3 times: once when listed, once if it was requested, and once if it is used). 
- the Portaudio backend (used on Windows in both servers, and in supernova on macOS) reports latency but not hardware buffer size. Reporting hardware buffer size would be nice, but I'm not sure how feasible it is -> I think this is less likely achievable
- we could consider adding latency info on macOS (scsynth), but the coreaudio backend needs to be rewritten (uses old APIs) so let's postpone that
- supernova on macOS uses portaudio which leads to posting consistent with Windows but not macOS scsynth - eventually I'm hoping to write a native coreaudio backend for supernova so again, I would postpone any considerations between supernova and scsynth _on macOS_.

In general, I think we could consider separate PR(s) for removing streams info and considering printing requested device. Other matters touch on underlying implementation and I don't think they are practical to solve, currently.

However, I expressed my preference for keeping the number of ins/outs in the same line as the device name. With larger list of devices, the list is more likely to not fit in the height of the post window (this is definitely the case on my system). Would you consider removing newline before the ins/outs posting? I think the addition of quotation marks makes is clear what is and what is not a device name.

prko:
Done in https://github.com/supercollider/supercollider/pull/7018/commits/d436fed737a9975b9a2f0f7fd3efa498c4025f43

---

### macOS scsynth
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.
SCDoc: Indexing help-files...

*** Scsynth booting
12 Available Audio Devices:
- "SAMSUNG"   (0 ins, 2 outs)
- "Live Gamer Ultra 2.1-Audio"   (4 ins, 0 outs)
- "CalDigit Thunderbolt 3 Audio"   (0 ins, 2 outs)
- "CalDigit Thunderbolt 3 Audio"   (2 ins, 0 outs)
- "BlackHole 16ch"   (16 ins, 16 outs)
- "BlackHole 2ch"   (2 ins, 2 outs)
- "BlackHole 64ch"   (64 ins, 64 outs)
- "MacBook Pro Microphone"   (1 ins, 0 outs)
- "MacBook Pro Speakers"   (0 ins, 2 outs)
- "ZoomAudioD"   (2 ins, 2 outs)
- "Aggregate Device"   (5 ins, 2 outs)
- "Multi-Output Device"   (0 ins, 2 outs)

Booting with:
  In: "MacBook Pro Microphone" (Streams: 1 , 0), 1 outs
  Out: "BlackHole 2ch" (Streams: 1 0), 2 ins)
  Sample rate: 44100.000000
  SC_AudioDriver's block size: 512
SCDoc: Indexed 2997 documents in 1.36 seconds
SuperCollider 3 server (scsynth) ready.

localhost: server process's maxLogins (1) matches with my options.
```
### macOS supernova
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Supernova booting
12 Available Audio Devices:
- "SAMSUNG"   (0 ins, 2 outs)
- "Live Gamer Ultra 2.1-Audio"   (4 ins, 0 outs)
- "CalDigit Thunderbolt 3 Audio"   (0 ins, 2 outs)
- "CalDigit Thunderbolt 3 Audio"   (2 ins, 0 outs)
- "BlackHole 16ch"   (16 ins, 16 outs)
- "BlackHole 2ch"   (2 ins, 2 outs)
- "BlackHole 64ch"   (64 ins, 64 outs)
- "MacBook Pro Microphone"   (1 ins, 0 outs)
- "MacBook Pro Speakers"   (0 ins, 2 outs)
- "ZoomAudioDevice"   (2 ins, 2 outs)
- "Aggregate Device"   (5 ins, 2 outs)
- "Multi-Output Device"   (0 ins, 2 outs)

Requested audio devices:
  In: "(default)"
  Out: "(default)"

Selecting default system input/output devices

Booting with:
  In: "MacBook Pro Microphone", 3 in(s)
  Out: "BlackHole 2ch", 4 out(s)
  Sample rate: 44100.000
  Latency (in/out): 0.249 / 0.013 sec
Audio devices opened successfully, and server booted successfully.
SuperCollider 3 server (supernova) ready.

localhost: no maxLogins info from server process.
```

### Windows scsynth
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Scsynth booting
3 Available Audio Devices:
- "ASIO : Magix Low Latency 2016"   (2 ins, 8 outs)
- "ASIO : ReaRoute ASIO (x64)"   (16 ins, 16 outs)
- "Windows WASAPI : Speakers(High Definition Audio Device)"   (0 ins, 2 outs)

Requested devices:
  In:
  - (default)
  Out:
  - (default)

Selecting default system output device
WARNING: no input device selected, input is disabled.

Booting with:
  Out: "Windows WASAPI : Speakers(High Definition Audio Device)" 2 out
  Sample rate: 48000.000
  Latency (in/out): 0.000 / 0.022 sec
  Sample rate: 48000.000000
  SC_AudioDriver's block size: 64
SuperCollider 3 server (scsynth) ready.

localhost: server process's maxLogins (1) matches with my options.
```

### Windows supernova
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Supernova booting
3 Available Audio Devices:
- "ASIO : Magix Low Latency 2016"   (2 ins, 8 outs)
- "ASIO : ReaRoute ASIO (x64)"   (16 ins, 16 outs)
- "Windows WASAPI : Speakers(High Definition Audio Device)"   (0 ins, 2 outs)

Requested audio devices:
  In: "(default)"
  Out: "(default)"

Selecting default system output device
WARNING: no input device selected, input is disabled.

Booting with:
```

---

I think this is a bit confusing. How about when the in/out posting comes before the device name?
### Alternative format A
```text
- (  0 ins,   2 outs) "SAMSUNG"
- (  4 ins,   0 outs) "Live Gamer Ultra 2.1-Audio" 
- (  0 ins,   2 outs) "CalDigit Thunderbolt 3 Audio"
- (  2 ins,   0 outs) "CalDigit Thunderbolt 3 Audio"
- ( 16 ins,  16 outs) "BlackHole 16ch"  
- (  2 ins,   2 outs) "BlackHole 2ch"   
- ( 64 ins,  64 outs) "BlackHole 64ch"
- (128 ins, 128 outs) "a device with 128ch"
- (  1 ins,   0 outs) "MacBook Pro Microphone"
- (  0 ins,   2 outs) "MacBook Pro Speakers"   
- (  2 ins,   2 outs) "ZoomAudioDevice"   
- (  5 ins,   2 outs) "Aggregate Device"   
- (  0 ins,   2 outs) "Multi-Output Device"
- (  2 ins,   8 outs) "ASIO : Magix Low Latency 2016" 
- ( 16 ins,  16 outs) "ASIO : ReaRoute ASIO (x64)"
- (  0 ins,   2 outs) "Windows WASAPI : Speakers(High Definition Audio Device)"
```

### Alternative format B

```text
- (0 ins, 2 outs)	  "SAMSUNG"
- (4 ins, 0 outs)	  "Live Gamer Ultra 2.1-Audio" 
- (0 ins, 2 outs)	  "CalDigit Thunderbolt 3 Audio"
- (2 ins, 0 outs)	  "CalDigit Thunderbolt 3 Audio"
- (16 ins, 16 outs)	  "BlackHole 16ch"
- (2 ins, 2 outs)	  "BlackHole 2ch" 
- (64 ins, 64 outs)	  "BlackHole 64ch"
- (128 ins, 128 outs)	  "a device with 128ch"
- (1 ins, 0 outs)	  "MacBook Pro Microphone"
- (0 ins, 2 outs)	  "MacBook Pro Speakers" 
- (2 ins, 2 outs)	  "ZoomAudioDevice" 
- (5 ins, 2 outs)	  "Aggregate Device" 
- (0 ins, 2 outs)	  "Multi-Output Device"
- (2 ins, 8 outs)	  "ASIO : Magix Low Latency 2016" 
- (16 ins, 16 outs)	  "ASIO : ReaRoute ASIO (x64)"
- (0 ins, 2 outs)	  "Windows WASAPI : Speakers(High Definition Audio Device)"
```

If neither of the alternative formats works for you, or if you have no further suggestions, please feel free to merge it.

dyfer:
I'd prefer listing the device name first - I think name is the most relevant information so should be listed first, especially if the later part is cut off in a smaller window.

Do we really need 3 spaces between the name and ins/outs? I think that with the quotes in place it would still be readable with a single space.

SCClassLibrary/Common/Control/ServerStatus.sc
```diff
@@ -326,7 +326,7 @@ ServerStatusWatcher {
 		server.sendMsg("/notify", flag.binaryValue, server.clientID);
 
 		if(flag){
-			"Requested notification messages from server '%'\n".postf(server.name)
+			"\nRequested notification messages from server '%'\n".postf(server.name)
```
dyfer: Why do we add a newline at the beginning? We don't typically do that.

If we really want a space after the server ready message, we could add two newlines there... But i'm not sure if that's needed.

capital-G:
Quoting myself: https://github.com/supercollider/supercollider/pull/7018#issuecomment-3027560123

> [...] we could/should(?) probably refactor the printing code into a dedicated class in order to reduce the maintenance over time.

I think the cleanest approach would be to have something like a `PrintInfo` class in `common` which accepts some information about available audio interfaces via a struct. This way we would have one spot where the printing gets done in case we think synchronizing them is worth the effort.

Patching the individual implementations will only lead to divergence again over time.

I still think this is nice to have but not a high priority.
Ideally a LSP could auto-complete the available interfaces in case this is needed, which I think is providing much more help for the user in everyday situations.

dyfer:
> I think the cleanest approach would be to have something like a `PrintInfo` class in `common` which accepts some information about available audio interfaces via a struct. This way we would have one spot where the printing gets done in case we think synchronizing them is worth the effort.
> 
> Patching the individual implementations will only lead to divergence again over time.


Yes, it would be the cleanest, but the most code is in extracting this info from individual backends. Printing is duplicated but that code is minimal and IMO manageable to keep in sync, more or less. We're not adding new backends every month :)
 

> I still think this is nice to have but not a high priority. Ideally a LSP could auto-complete the available interfaces in case this is needed, which I think is providing much more help for the user in everyday situations.

Well, the whole sclang-server duality is rather clumsy when it comes to managing audio devices. From the logical standpoint, listing devices while the server is booting is useless/too late - the information from the device listing needs to be applied _before_ booting the server. I personally am used to the server listing devices when booting and would not necessarily want to remove it, but from the logical standpoint that information is provided too late - one needs to restart the server if a change is needed....

OTOH there's no strict connection between the backend used by the server and the language (they match, mostly, but they may not). 

And then there's a matter using the server with other clients, which may not have a way of listing audio devices. I thought that maybe we should add a flag to the server for listing devices in a serialized manner (without booting), and that should be the way to inquire what's available for all clients, not only sclang.

----

I _don't think_ that reworking the whole process is a task for this PR. As @prko insists on unifying the posting as much as possible, I think we can do that more or less as suggested in this PR.

SCClassLibrary/Common/Control/ServerStatus.sc
prko: I follow your opinion. Thanks! Done in https://github.com/supercollider/supercollider/pull/7018/commits/da87349aaaf8efc303e5284506093c7f4b7c0557

server/scsynth/SC_CoreAudio.cpp
```diff
@@ -589,8 +589,33 @@ bool SC_CoreAudioDriver::DriverSetup(int* outNumSamplesPerCallback, double* outS
...
+            AudioObjectGetPropertyData(devices[i], &chPropAddr, 0, NULL, &propertySize, inputBufList);
+
+            int inChs = 0;
+            for (UInt32 j = 0; j < inputBufList->mNumberBuffers; ++j)
+                inChs += inputBufList->mBuffers[j].mNumberChannels;
+            free(inputBufList);
+
+            chPropAddr.mScope = kAudioDevicePropertyScopeOutput;
+            AudioObjectGetPropertyDataSize(devices[i], &chPropAddr, 0, NULL, &propertySize);
+            AudioBufferList* outputBufList = (AudioBufferList*)malloc(propertySize);
```
JordanHendersonMusic: Why are we using malloc?

prko: # MS Copillot says:

---

The short answer: **because you don’t know the size of the `AudioBufferList` at compile time**, and CoreAudio requires you to allocate exactly the amount of memory the system reports.

---

## 🧠 **Why `malloc` is required here**

### 1. **`AudioObjectGetPropertyDataSize` returns a runtime‑determined size**
The number of buffers and the size of the `AudioBufferList` depend on the audio device.  
Different devices have different numbers of output channels, different buffer configurations, etc.

So you first ask CoreAudio:

```c
AudioObjectGetPropertyDataSize(devices[i], &chPropAddr, 0, NULL, &propertySize);
```

This gives you the exact number of bytes needed for the `AudioBufferList`.

### 2. **`AudioBufferList` is a variable‑length struct**
`AudioBufferList` looks like this:

```c
struct AudioBufferList {
    UInt32      mNumberBuffers;
    AudioBuffer mBuffers[1]; // variable length!
};
```

That `mBuffers[1]` is a classic C trick:  
the struct *pretends* to have one buffer, but in reality you must allocate enough space for however many buffers the device actually has.

So you cannot do:

```c
AudioBufferList outputBufList;  // WRONG
```

because you don’t know how many buffers it needs.

### 3. **Therefore you must allocate dynamically**
```c
AudioBufferList* outputBufList = (AudioBufferList*)malloc(propertySize);
```

This allocates exactly the correct size for the device’s buffer list.

---

## 🧩 **Why not use `new` or stack allocation?**

- Stack allocation (`AudioBufferList buf;`) is impossible because the size is unknown.
- `new` would work in C++, but this code is C, and CoreAudio examples traditionally use `malloc`.
- CoreAudio APIs expect you to manage the memory manually.

---

# Google Gemini says:

The reason `malloc` was used here is due to the design of Apple's CoreAudio `AudioBufferList` struct. It contains a variable-length array (`mBuffers`), meaning the required memory size fluctuates at runtime depending on the device's configuration. Because of this, we have to call `AudioObjectGetPropertyDataSize` to get the exact required byte size (`propertySize`) and then allocate that memory dynamically.

However, you make a great point. Using raw `malloc` in C++ requires manual `free` calls, which can easily lead to memory leaks in the event of an early return or an exception. It is definitely safer to replace it with a modern C++ RAII approach to ensure the memory is cleaned up automatically.

I will update the code to use `std::vector` (or `std::unique_ptr`) to handle this safely.

**Proposed Fix (using `std::vector`):**

```cpp
chPropAddr.mScope = kAudioDevicePropertyScopeOutput;
AudioObjectGetPropertyDataSize(devices[i], &chPropAddr, 0, NULL, &propertySize);

// Use std::vector to manage memory automatically
std::vector<uint8_t> bufferData(propertySize);
AudioBufferList* outputBufList = reinterpret_cast<AudioBufferList*>(bufferData.data());

```

---

*Tip: `std::vector` is generally the standard and most readable approach for this in modern C++, but if the project heavily prefers smart pointers for this kind of raw byte allocation, you can swap the code block with the `std::unique_ptr` version provided in the previous answer.*

prko: Let me know what I should do!

capital-G: > Why are we using malloc?

tldr - this is common practice w/ macOS API, see

https://github.com/supercollider/supercollider/blob/2f0803bcd2e551564e3fef8d5075816cbb685cd4/server/scsynth/SC_CoreAudio.cpp#L999-L1004

I remember asking jmc about this as well ;)

JordanHendersonMusic: @prko I'm not reading that.

> tldr - this is common practice w/ macOS API, see

Do you know why? If 'just because' this seems like a very bad practice!

prko: >>tldr - this is common practice w/ macOS API, see

>Do you know why? If 'just because' this seems like a very bad practice!

I think this question could be discussed in a new issue or on https://www.scsynth.org/, so that we can keep this PR focused.

If there are no objections, could we move forward with this PR?

capital-G: > Do you know why? If 'just because' this seems like a very bad practice!

Isn't this a common practice with C based APIs? You pass a pointer where the results get written into? Instead of providing many functions, apple has a single function, maybe easier for byte-masking and adding undocumented features?

JordanHendersonMusic: Yes but SC_CoreAudio is C++. 

This suggestion from gemmini looks almost right to me.

```cpp
chPropAddr.mScope = kAudioDevicePropertyScopeOutput;
AudioObjectGetPropertyDataSize(devices[i], &chPropAddr, 0, NULL, &propertySize);

// Use std::vector to manage memory automatically
std::vector<uint8_t> bufferData(propertySize);
auto* bufList = reinterpret_cast<AudioBufferList*>(bufferData.data());

```

Also we shouldn't really be using `NULL`, either `0` or `nullptr`.

prko: I noticed 21 instances of malloc in this file. Should we address all of them in this PR, or would it be better to open a new one? If we decide to fix all of them in this PR, I would welcome one or more co-authors to help! Please let me know specifically which of the 21 malloc calls need to be updated.

prko: I think it would be better to merge this PR as it is.
Regarding `malloc` and `NULL`, it might be best for someone familiar with C++ to address those issues.

prko: @capital-G @JordanHendersonMusic @dyfer
Could I mark this thread as resolved?
As the scope of this PR is just to improve posting when booting the server, I think it would be best to move forward with merging this as it is.
Since there hasn't been any further discussion on updating the pre-existing `malloc` and `NULL` usage, the PR seems to be stuck right now. If we need to refactor those into modern C++, it might be better handled in a separate issue. Let me know if it's okay to resolve this and proceed!

capital-G: IMO it is okay to do the allocation like this - this is common behavior in the existing code and I think it is better to follow consistent w/ existing approaches and then do a PR which refactors this approach.

I am still hesitant on the general approach of this PR though b/c, as said multiple times, I think this should be done in a printing class b/c this will definitely diverge again, but I don't really think that constructing such a class is also the best spend effort currently b/c of sparse review resources.
Since dyfer has a different stance on that, I'll pass the handling of the PR over to him ;)

JordanHendersonMusic: I won't be the one refactoring this, so feel free to ignore me.... But I'd be quite annoyed to know that the clearer more readable solution was discussed and opted against. 

Its about ensuring that the code you write *means* exactly what it does. Raw pointers can mean many things, this is literally why rust needs to exists. Being consistent with code that is hard to read is not a good idea.

prko: >I am still hesitant on the general approach of this PR though b/c, as said multiple times, I think this should be done in a printing class b/c this will definitely diverge again

>Being consistent with code that is hard to read is not a good idea.

I agree, but my programming ability isn’t at that level yet.
I’m sorry for trying to make changes beyond my skill, but I still think this change is better than leaving it as it is.

capital-G: > I won't be the one refactoring this, so feel free to ignore me.... But I'd be quite annoyed to know that the clearer more readable solution was discussed and opted against.

I think inconsistencies across the code base are a greater nuisance than having legacy approaches around - I think if you don't agree w/ this then we have to agree to disagree on that topic ;)
E.g. if you do a refactor to change e.g. to smart pointers on `HiddenWorld` (which came up lately), it seems better to me to do this in one go as otherwise I have to look up all the special cases, which just creates additional mental load.
The target should be to have a tested, safe and modern approach across the code base, but these are all PRs by themselves ;) So I think the current approach of talking w/ CoreAudio is fine in regards to the scope of this PR.
My otherwise already stated objection to the printing approach of this PR remains though^^

JordanHendersonMusic: > it seems better to me to do this in one go as otherwise I have to look up all the special cases, which just creates additional mental load.

If it's a pointer you need to look at all the places it's is used. This is because a pointer can be: an owning pointer which you must free, a shared resource which you just conditionally free, a reference which you should never free; it can also already be null, or an array of of items with a size,  or perhaps point outside of the array....

If it's a unique ptr or a vector (as in this pr) you don't need to think about this. You just copy or move it.

By adding more pointers you are making whomever has to refactor this do this error prone work of investigating what the pointer actually means in this specific case. 

I don't care about consistency when what your are trying to be consistent with is very hard to maintain and there are simple c++ structures that can exactly express your intent.

I think this pr should not refactor the code, but just add it's use with a vector. If somehow refractors late, refactoring the vector case is obvious because it explicitly says how the memory is used.


@prko, the reason a less familiar programmer should free uncomfortable refactoring this is ***precisely*** because it uses raw pointers. If it was vectors, you can freely copy them without worrying. You are adding to your problem.

Again I won't be the one to refactor this, so you can ignore me, but I strongly disagree with the reasons for doing this.

prko: Please take a look at commit 98bb369.

I attempted to reflect @JordanHendersonMusic’s feedback. I’m not entirely sure about this change, but I can launch SC without any problem, and both scsynth and supernova boot successfully. Also, I can hear sound!

Please let me know what I should do to get this PR merged.

```text
*** Booting server 'localhost' on address 127.0.0.1:57110.
Found 0 LADSPA plugins

*** Scsynth booting
6 Available Audio Devices:
- "BlackHole 2ch" (2 ins, 2 outs)
- "MacBook Pro Microphone" (1 ins, 0 outs)
- "MacBook Pro Speakers" (0 ins, 2 outs)
- "ZoomAudioD" (2 ins, 2 outs)
- "Aggregate Device" (5 ins, 2 outs)
- "Multi-Output Device" (0 ins, 2 outs)

Booting with:
  In: "MacBook Pro Microphone" (Streams: 1 , 0), 1 outs
  Out: "BlackHole 2ch" (Streams: 1 0), 2 ins)
  Sample rate: 48000.000000
  SC_AudioDriver's block size: 512
SuperCollider 3 server (scsynth) ready.
localhost: server process's maxLogins (1) matches with my options.
```
```text
*** Booting server 'localhost' on address 127.0.0.1:57110.

*** Supernova booting
6 Available Audio Devices:
- "BlackHole 2ch" (2 ins, 2 outs)
- "MacBook Pro Microphone" (1 ins, 0 outs)
- "MacBook Pro Speakers" (0 ins, 2 outs)
- "ZoomAudioDevice" (2 ins, 2 outs)
- "Aggregate Device" (5 ins, 2 outs)
- "Multi-Output Device" (0 ins, 2 outs)

Found 0 LADSPA plugins
Requested audio devices:
  In: "(default)"
  Out: "(default)"

Selecting default system input/output devices

Booting with:
  In: "MacBook Pro Microphone", 2 in(s)
  Out: "BlackHole 2ch", 2 out(s)
  Sample rate: 48000.000
  Latency (in/out): 0.233 / 0.012 sec
Audio devices opened successfully, and server booted successfully.
SuperCollider 3 server (supernova) ready.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
localhost: no maxLogins info from server process.
node id 2 already in use
node id 2 already in use
node id 2 already in use
node id 2 already in use
node id 2 already in use
node id 2 already in use
node id 1 already in use
node id 1 already in use
node id 1 already in use
node id 1 already in use
node id 1 already in use
node id 1 already in use
```

prko:
You can find a screenshot with comments showing the status up to commit [98bb369](https://github.com/supercollider/supercollider/pull/7018/commits/98bb3695033fb9b0b3aeb18583e7c2109483c890) in [issue #7647 (comment)](https://github.com/supercollider/supercollider/issues/7647#issuecomment-5337419883).

The screenshot includes notes on the current state of the changes and helps illustrate the progress made so far.

Issue #7020, opened 2025-06-30
selector symbols may segfault sclang


telephon:
All primitives that take a symbol as a method selector crash when a class name is passed. This is of course accidental, but still we may want to fix this.

The reason is that the method table lookup calculation from the symbol returns an out of bounds index when the symbol is a class name:

https://github.com/supercollider/supercollider/blob/bc06e6cd7a7c33d6a96488e7ab0129989de52333/lang/LangPrimSource/PyrPrimitive.cpp#L2242

This was noticed in #7016.

## Steps to reproduce

```supercollider
[].perform(\Foo);
[].perform(\Object); // crashes
[].doesNotUnderstand(\Object) // crashes
```

## Solution
This could be avoided by adding a size check for index into the method table.

Currently, there is none:
```c++
 selector = slotRawSymbol(b);
 index = slotRawInt(&classobj->classIndex) + selector->u.index;
 meth = gRowTable[index];
```

JordanHendersonMusic:
There's no overhead with either a small function like that, nor the optional type.

dyfer:
@telephon have this issue been addressed in [#7021](https://github.com/supercollider/supercollider/pull/7021)?

PR #7021, opened 2025-06-30
sclang: protect method lookup from out of bounds index

### LINKED ISSUE #7016: respondsTo crashes with class name as selector
The primitive `prObjectRespondsTo` is called, and the offset in the method table is probably wrong:

https://github.com/supercollider/supercollider/blob/bc06e6cd7a7c33d6a96488e7ab0129989de52333/lang/LangPrimSource/PyrPrimitive.cpp#L2242

https://github.com/supercollider/supercollider/blob/bc06e6cd7a7c33d6a96488e7ab0129989de52333/lang/LangPrimSource/PyrPrimitive.cpp#L2257

## Steps to reproduce

```supercollider
[].respondsTo(\Foo) // false
[].respondsTo([\plot, \scramble]) // true
[].respondsTo(\Object) // crashes
[].respondsTo([\plot, \Object]) // crashes
```

we'd expect it to return `false`.

`Interpreter has crashed or stopped forcefully. [Exit code: 11]`

telephon:
When the method selector is a symbol that is a class name, method lookup may be out of bounds. We check if the symbol is a monster and return.

This fixes #7016 and #7020.

```supercollider
[].respondsTo(\Object)
[].doesNotUnderstand(\Object)
[].perform(\Object);
[].performList(\Object);
[].performArgs(\Object);
```

## Discussion

This adds an extra check in a central place of method dispatch.
We should discuss if this is the best way to go, and make sure this is really the most performant way.

The community hasn't noticed this bug for 20 years, so is it worth checking for the monster every method call? Maybe yes, maybe no.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing

### DIFF
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -118,6 +118,10 @@ HOT void sendMessageImpl(VMGlobals* g, PyrSymbol* selector, PyrSlot* recvrSlot, 
     // In cases of redirect and super calls the function to be recursively call, this is implemented with a goto.
 
 lookup_again:
+    if ((selector->flags & sym_Class) != 0) {
+        post("A method selector can't be a class name\n");
+        return;
+    };
     // The variables classobj and selector are updated when the goto is triggered, causing these to change.
     auto method = gRowTable[slotRawInt(&classobj->classIndex) + selector->u.index];
     auto methodRaw = METHRAW(method);
@@ -342,8 +346,8 @@ void doesNotUnderstand(VMGlobals* g, PyrSymbol* selector, std::int64_t numArgsPu
         std::rotate(rargBegin, rargBegin + 1, rargEnd);
     }
     auto selectorSlot = g->sp - numArgsPushed + 1;
-
     const auto callIndex = slotRawInt(&classObject->classIndex) + s_doesNotUnderstand->u.index;
+
     auto method = gRowTable[callIndex];
 
     if (tryUniqueMethod(g, method, receiverSlot, selectorSlot, numArgsPushed, numKeyArgsPushed))
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -118,11 +118,12 @@ HOT void sendMessageImpl(VMGlobals* g, PyrSymbol* selector, PyrSlot* recvrSlot, 
     // In cases of redirect and super calls the function to be recursively call, this is implemented with a goto.
 
 lookup_again:
-    if ((selector->flags & sym_Class) != 0) {
-        post("A method selector can't be a class name\n");
-        return;
-    };
     // The variables classobj and selector are updated when the goto is triggered, causing these to change.
+    if ((selector->flags & sym_Class) != 0) {
+        // if selector is a class name, a lookup in gRowTable would be indexed out of bounds
+        doesNotUnderstand(g, selector, numArgsPushed, numKeyArgsPushed);
+        return;
+    }
     auto method = gRowTable[slotRawInt(&classobj->classIndex) + selector->u.index];
     auto methodRaw = METHRAW(method);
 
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
We don't need the unlikely stuff. It's a super micro optimisation that should only be used in tight loops.

dyfer:
Thanks @telephon !

As this is potentially a bigfix for 3.14 (IIUC), it should be rebased on 3.14.

telephon:
OK, so the question is whether this way of fixing it is good (I'll remove the UNLIKELY macros again), or if there should be a function as jordan suggested.

telephon:
I don't know how to fix `sendMessageImpl` – something is different there. @JordanHendersonMusic do you have an idea?

telephon:
> That is weird.

seems that I have solved it by not returning an error ...

> Is it always falling over a particular case, or in all cases - if so could you print the name? I don't see how it could fail in all cases.

I would have done it, but I don't know how to extract a string from the selector.

JordanHendersonMusic:
@capital-G it is a long standing bug (since the first commit I think), so it would be good to get this in as soon as possible... but it only occurs in rare situations and is quite easily avoided, so it wouldn't be the end of the world if it waited until 15.

telephon:
ok, this is ready to be discussed (see changed description above).

telephon:
P.S. I'll write tests once we agree to go on with this.

lang/LangSource/PyrMessage.cpp
```diff
@@ -119,6 +119,11 @@ HOT void sendMessageImpl(VMGlobals* g, PyrSymbol* selector, PyrSlot* recvrSlot,
 
 lookup_again:
     // The variables classobj and selector are updated when the goto is triggered, causing these to change.
+    if ((selector->flags & sym_Class) != 0) {
+        // if selector is a class name, a lookup in gRowTable would be indexed out of bounds
+        doesNotUnderstand(g, selector, numArgsPushed, numKeyArgsPushed);
```
JordanHendersonMusic, line 124: I don't think it is right to call doesnotunderstand here.

This would be a side effect.
```
a = (\Foo: 10);
a.Foo // syntax error
a.perform(\Foo) // 10
```
It is odd that perform and perform syntax diverge.

I think it is best to post an error and fail.

JordanHendersonMusic:
I don't think you need tests, this is pretty straightforward.

lang/LangSource/PyrMessage.cpp
telephon: In order to fail you'd have to return an error, but we are in a `void` function. Just returning doesn't end up doing the right thing. 

The code you post does the right thing in this branch, i think. That is, it doesn't change its behaviour from what it used to be.

Also this works correctly:
```
a = (\Object: 10);
a.perform(\Object) // 10

// ... and this doesn't crash:
a = [];
a.perform(\Object) // error


```
>It is odd that perform and perform syntax diverge.

not sure what you meant?

JordanHendersonMusic: Ah I see, never mind!

telephon:
@dyfer @capital-G what do you think, should this go into 3.14? If the change is acceptable, we can switch the branch again, or just merge for 3.15.

dyfer:
Since it doesn't fix a regression from 3.13 and we've already released rc1, and the bug is not easily triggered, my preference would be for it to go to 3.15. But I don't oppose if others feel strongly that it should go to 3.14.

telephon:
I think it is ok to merge it for 3.15 – it is a very old bug and nobody will be hurt by it.

dyfer:
When squash merging this, please make sure to remove reference to issue 7007, as that is [not the correct issue number](https://github.com/supercollider/supercollider/pull/7021#issuecomment-3027945797).

Issue #7006, opened 2025-06-28
Drop Qt5 compatibility


capital-G:
3.14 introduced Qt6 support, and since Qt5 has already reached its EOL (https://www.qt.io/blog/qt-5.15-extended-support-for-subscription-license-holders), I don't see much reason to still maintain Qt5 compatibility within the codebase. If people rely on Qt5, they should probably stick to 3.13.
Since we also include the Qt browser this also creates either some security problems or a more complex build.

Although the Qt5 compatibility does not result in massive amounts of code, it just clutters up the code and also creates support issues (see https://github.com/supercollider/supercollider/issues/7003).
One of the reasons why the SC codebase is hard to dig through is IMO also because of legacy support (e.g. we still have some powerbook optimizations within the UGen codebase...).
This just piles up over the time, and now would probably the time to remove it.

## ToDo

* [ ] https://github.com/supercollider/supercollider/issues/6824#issuecomment-3682463686 - still some connections and ifdefs are in place to maintain qt5 support

telephon:
What may need to be checked: I heard that linuxes on RaspberryPi do not yet support QT6, which could mean that scide won't be usable anymore on this platform. I am not sure if this is true, maybe @redFrik do you know more?

redFrik:
Qt6 works just fine on Raspberry Pi. Our newly updated RPi build instructions even promote it.  https://github.com/supercollider/supercollider/blob/3.14/README_RASPBERRY_PI.md

telephon:
thank you!

9Skrip:
thats bad now unusable on windows 7 but still supports powerbooks

jamshark70:
> thats bad now unusable on windows 7 but still supports powerbooks

The powerbook comment isn't about Qt -- Apple had ditched PowerPC chips long before we had Qt in SC.

9Skrip:
here is the project to make [windows 7 compatibility patches](https://github.com/crystalidea/qt6windows7) https://github.com/crystalidea/qt6windows7

dyfer:
Some relevant discussion:

https://github.com/supercollider/supercollider/pull/7593#issuecomment-4815867051
https://github.com/supercollider/supercollider/pull/7593#issuecomment-4819928875
https://github.com/supercollider/supercollider/pull/7593#issuecomment-4820297582
https://github.com/supercollider/supercollider/pull/7593#issuecomment-4820526537
[...]

Issue #7003, opened 2025-06-26
IDE Crashing on Launch. Mac 15.5 Sequoia , M1 Max


wortsampler:
## Environment

* SuperCollider version: *3.15.0-Dev*
* Operating system: *macOS 15.5 on M1 Max*
* Other details (Qt version, audio driver, etc.):  ```qt@5: stable 5.15.16``` and ```qt: stable 6.9.0```

## Steps to reproduce
```brew unlink qt@5```

```cmake -G Xcode -DCMAKE_PREFIX_PATH=`brew --prefix qt@5` ..```

and: 

```cmake --build . --target install --config RelWithDebInfo  ```
## Expected vs. actual behavior

Building succeeds with no errors.
When i try to open the IDE it crashes instantly.  

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace SIGNAL, Code 6 Abort trap: 6
Terminating Process:   SuperCollider [77261]

Application Specific Information:
abort() called


Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x185a15388 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x185a4e88c pthread_kill + 296
2   libsystem_c.dylib             	       0x185957c60 abort + 124
3   QtCore                        	       0x10305c828 0x103050000 + 51240
4   QtCore                        	       0x103073978 0x103050000 + 145784
5   QtCore                        	       0x1032dacdc QMessageLogger::fatal(char const*, ...) const + 96
6   QtGui                         	       0x102b391f8 QGuiApplicationPrivate::createPlatformIntegration() + 7672
7   QtGui                         	       0x102b39254 QGuiApplicationPrivate::createEventDispatcher() + 36
8   QtCore                        	       0x1030d50d0 QCoreApplicationPrivate::init() + 732
9   QtGui                         	       0x102b34530 QGuiApplicationPrivate::init() + 76
10  QtWidgets                     	       0x1016e761c QApplicationPrivate::init() + 60
11  QtWidgets                     	       0x1016e75cc QApplication::QApplication(int&, char**, int) + 140
12  SuperCollider                 	       0x100c76374 0x100c6c000 + 41844
13  dyld                          	       0x1856aeb98 start + 6076

Thread 1:
[...]

Issue #6579, opened 2024-12-22
3.14 release


capital-G:
SuperCollider 3.13.0 was released in February of 2023 and since then we had some great additions, fixes and improvements which are probably useful to distribute them to a wider audience through an official release.

Having the [SuperCollider symposium](https://scsynth.org/t/supercollider-symposium-2025/9911) on 13-15 March 2025 labelled *The Future of SuperCollider*, I'd propose to use this event as an opportunity to release the next major version, which would be `3.14.0` :)

The time horizon is ambitious but the event is a great opportunity, and quoting @dyfer from https://github.com/supercollider/supercollider/issues/6547#issuecomment-2499303785

> [...]and then focus on working towards 3.14 sooner than later.

The main questions we should discuss

* [ ] When should we "seal-off" the develop branch for an RC?
* [ ] What open issues/PRs do you consider for to the next major release?
* [ ] Should we target the symposium date for an RC or for an actual major release? 

Feel free to add issues/PR to the milestone 3.14 (see https://github.com/supercollider/supercollider/milestone/25) or use this issue to discuss the scope of the next major release. I created a new GH project @ https://github.com/orgs/supercollider/projects/2/views/1 in order to obtain a better overview and will try to manage this according to the comments here and elsewhere - right now it is rather empty but this is about to change.

Also please state if you agree or disagree with the proposition of a mid-March release of 3.14.0 so we can get an overview within the dev-community about this.
It would be also great if you could state your capacities until March to see if we have enough capacity within the community for such an endeavor.

Also see https://github.com/supercollider/supercollider/wiki/Releasing-cycle-and-process for the formalities of a release (which suggests a major release every 3-6 months which seems long overdue)

cdbzb:
I'll defer to more active folks but I think this would be *great* and I'd be happy to do at least some testing

prko:
For me, it would be great if the following PRs and issues could be implemented:
- #6501
- #6401
- #6566
- #6563
- #6554
- #6549
- #6545
- #6537 
- #6536 
- #6534 
- #6479
A related issue seems to have been removed.... I cannot find some of my posts that belong to it.
- #6539
- #6415 
- #6409 
- #6406 
- #6394
- #6385 
- #6379
- #6369 
- #6371
- #6366
- #6315
- #6313
- #6312 
- #6294 
- #6291 
- #6283
- #6282 
- #6278 
- #6265
#6304
- #6207
- #6171
- #6168
- #6154
- #6153
- #6141
[...]

PR #7025, opened 2025-07-02
clarified use of supernova and fixed build instructions with qt5

wortsampler:
## Purpose and Motivation
As discovered in https://github.com/supercollider/supercollider/issues/7003 there was a typo in the qt5 building instructions in the macOS readme. 

Also clarified the use of supernova and added a stronger suggestion to drop Qt5 (EOL). 
See also https://github.com/supercollider/supercollider/issues/7006

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/README_MACOS.md
+++ b/README_MACOS.md
@@ -66,10 +66,9 @@ Build instructions
     cd SuperCollider
     mkdir -p build
     cd build
-    # the default way is to build SC with supernova: 
-    cmake -G Xcode -DSUPERNOVA=ON ..
-    # or, if you want to build without it:
     cmake -G Xcode  ..
+    # or, if you want to build with supernove:
+    cmake -G Xcode -DSUPERNOVA=ON ..
     # then start the build
     cmake --build . --target install --config RelWithDebInfo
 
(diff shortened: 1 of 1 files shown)


dyfer:
Just to clarify - one of the reasons supernova is not on by default, at least on macOS, is because it needs additional dependencies (fftw and portaudio) comparing to scsynth, IIUC.

I have plans to address this and turn it on by default on all platforms... but that should be discussed separately, I think.

I'd prefer it this PR only addressed the typo for the Qt6 unlink. The supernova flag could be then removed in `cmake -G Xcode -DCMAKE_PREFIX_PATH=`brew --prefix qt@5` ..` for consistency, if desired.

dyfer (approved):
Thanks! LGTM.

capital-G:
Thx for the contribution <3

Its already merged, but shouldn't this had 3.14 as a target? Should we cherry pick?

dyfer:
> Its already merged, but shouldn't this had 3.14 as a target? Should we cherry pick?

IMO the change on `develop` is more visible, as this is what shows up when you visit the repo on github (unless you change branch of course). We can still consider cherry-picking it, but I don't think it's that important?

capital-G:
Ideally it should be in both IMO so it is properly versioned. If e.g. #7006 becomes a thing there wouldn't be a notice about qt5 for 3.14 anymore.

capital-G:
> Hm, we probably need a better system than a label to keep track of cherry pick for closed PRs...

Maybe https://github.com/supercollider/supercollider/issues/6579#issuecomment-3029430351 will do for now?

wortsampler:
thanks!

wortsampler:
> Its already merged, but shouldn't this had 3.14 as a target? Should we cherry pick?
what would be the right way to do it - or how do you do it? would 3.14 be the upstream branch? but it should also be in the dev branch, shouldn't it?

PR #6464, opened 2024-09-08
QT: remove deprecations for QT 5.15

dyfer:
## Purpose and Motivation

This PR provides updates and fixes for the deprecations up to Qt 5.15.
This is an "intermediary" PR, extracted from the Qt6 update branch for easier review. It generally does not include Qt6 code explicitly, even though commits do mention qt6...

Thanks @jrsurge for starting this work back in the day! A number of commits were extracted from @jrsurge's qt6 branch.

This PR also raises the minimum supported QT to 5.15; however I'd wait with updating documentation until the Qt6 updates are finalized.

Please review carefully! I'm not confident in C++ nor in Qt :)

## Types of changes

- Bug fix (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- n/a Updated documentation
- [x] This PR is ready for review

PR #7030, opened 2025-07-03
QtCollider: AbstractStepValue widgets: improve wheelEvent scroll behavior + fix regression

### LINKED ISSUE #7008: Scroll behavior in Slider GUIs changed from 3.13 to 3.14 (Linux/Sway/Wayland)
## Environment

* SuperCollider version: 3.14-rc1
* Operating system: ArchLinux 6.15.4-arch2-1
* Other details (Qt version, audio driver, etc.): Qt 6.9.1, Wayland, Sway

## Steps to reproduce

When opening a `Slider().front` in 3.13.1 the touchpad scrolling behavior is smooth and responds to acceleration (you can switch values precisely or make big shifts depending on how fast you scroll). In 3.14-rc1 this is not the case anymore, the scroll jump of values doesnt change accordingly to the speed of movement. This is unfortunate because the old behavior can be much more expressive.

I don't have other OS to try if this is only a Linux/Wayland issue or general change with Qt6.

elgiano:
## Purpose and Motivation

Fixes #7008

When migrating from deprecated .delta() to .angleDelta()/.pixelDelta() (9d10e196ce263faff1f9951d0db012a948ffbc87)
there were some errors in scaling the delta values, and some probably typos turning vertical scaling to horizontal.

`angleDelta()` is a direct translation of the old deprecated `delta()`, so I think we could skip checking for `pixelDelta()` in widgets like QcSlider and just use `angleDelta()`, to get the old behavior back. I'm not sure if using pixelDelta gives any better precision/smoothness in this case. However, for now I've kept also `pixelDelta`, and changed it's scaling to a value that gives me the same feel as angleDelta (and as in 3.13).

Can someone on a mac try it and report how does scrolling feel between this PR, 3.13 and 3.14-rc1? For me on linux (X11/Wayland) it's now the same across all versions.

While at it, I also added support for wheelEvent in other widgets. Full list of affected widgets:
- QcSlider: adjusted
- QcNumberBox: adjusted
- ScIDE AudioStatusBox (accepts scrolling for adjusting server volume): adjusted
- QcSlider2D: added
- QcKnob: added

Note: this is not about Qt6, as .angleDelta and .pixelDelta were introduced in Qt5, and .delta was deprecated already then

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -24,6 +24,7 @@
 #include <QHBoxLayout>
 #include <QWheelEvent>
 
+#include "QtCollider/widgets/QcAbstractStepValue.h"
 #include "main.hpp"
 
 namespace ScIDE {
@@ -138,7 +139,7 @@ void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, b
 
 void AudioStatusBox::wheelEvent(QWheelEvent* event) {
     // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
-    double delta = event->modifiers().testFlag(Qt::AltModifier) ? event->angleDelta().x() : event->angleDelta().y();
+    const double delta = getNormalizedScrollRatio(event, 0).y();
     if (delta) {
         if (delta > 0 || event->inverted())
             emit increaseVolume();
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -138,7 +138,7 @@ void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, b
 
 void AudioStatusBox::wheelEvent(QWheelEvent* event) {
     if (!event->angleDelta().isNull()) {
-        if (event->angleDelta().y() > 0)
+        if (event->angleDelta().y() > 0 || event->inverted())
             emit increaseVolume();
         else
             emit decreaseVolume();
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -141,7 +141,7 @@ void AudioStatusBox::wheelEvent(QWheelEvent* event) {
     // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
     const double delta = getNormalizedScrollRatio(event, 0).y();
     if (delta) {
-        if (delta > 0 || event->inverted())
+        if (delta > 0)
             emit increaseVolume();
         else
             emit decreaseVolume();
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -139,7 +139,7 @@ void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, b
 
 void AudioStatusBox::wheelEvent(QWheelEvent* event) {
     // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
-    const double delta = getNormalizedScrollRatio(event, 0).y();
+    const double delta = getScrollSteps(event).y();
     if (delta) {
         if (delta > 0)
             emit increaseVolume();
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -138,7 +138,7 @@ void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, b
 
 void AudioStatusBox::wheelEvent(QWheelEvent* event) {
     if (!event->angleDelta().isNull()) {
-        if (event->angleDelta().x() > 0)
+        if (event->angleDelta().y() > 0)
             emit increaseVolume();
         else
             emit decreaseVolume();
--- a/editors/sc-ide/widgets/audio_status_box.cpp
+++ b/editors/sc-ide/widgets/audio_status_box.cpp
@@ -137,8 +137,10 @@ void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, b
 }
 
 void AudioStatusBox::wheelEvent(QWheelEvent* event) {
-    if (!event->angleDelta().isNull()) {
-        if (event->angleDelta().y() > 0 || event->inverted())
+    // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
+    double delta = event->modifiers().testFlag(Qt::AltModifier) ? event->angleDelta().x() : event->angleDelta().y();
+    if (delta) {
+        if (delta > 0 || event->inverted())
             emit increaseVolume();
         else
             emit decreaseVolume();
--- a/QtCollider/widgets/QcAbstractStepValue.h
+++ b/QtCollider/widgets/QcAbstractStepValue.h
@@ -41,37 +41,34 @@ private:
     double _altScale;
 };
 
-static QPointF getNormalizedScrollRatio(const QWheelEvent* e, QSizeF pixelRatio) {
+static QPointF getScrollSteps(const QWheelEvent* e) {
+    // use pixelDelta on systems that support it, except x11 (Qt docs say it's unreliable)
     const auto platform = QGuiApplication::platformName();
-    const bool isX11 = platform == "xcb";
-    // angleDelta: relative to 360deg; pixelDelta: relative to widget size
-    QPointF ratio;
-    // Force using angleDelta on X11 (advice from Qt docs)
-    // the check for pixelRatio.isEmpty is to allow forcing angleDelta by passing pixelRatio = 0
-    if (e->pixelDelta().isNull() || pixelRatio.isEmpty() || isX11) {
-        ratio.setX(e->angleDelta().x() / 8. / 360.);
-        ratio.setY(e->angleDelta().y() / 8. / 360.);
-    } else {
-        ratio.setX(e->pixelDelta().x() * pixelRatio.width());
-        ratio.setY(e->pixelDelta().y() * pixelRatio.height());
-    }
+    const bool isX11 = platform == "x11";
+    // angleDelta: map 15deg/step, pixelDelta: map 10px/step (found empirically to match angleDelta)
+    QPointF steps =
+        (e->pixelDelta().isNull() || isX11) ? QPointF(e->angleDelta()) / 120. : QPointF(e->pixelDelta()) / 10.;
+
+    // Use angleDelta (and not pixelDelta) for consistency with SC versions before 3.14
+    // QPointF ratio(e->angleDelta().x() / 120., e->angleDelta().y() / 120.);
+
+    if (!e->pixelDelta().isNull())
+        printf("scroll: angleDelta/12(%d,%d) pxDelta(%d,%d) px/step(%f) steps(%f,%f)\n", (e->angleDelta() / 12.).x(),
+               (e->angleDelta() / 12.).y(), e->pixelDelta().x(), e->pixelDelta().y(),
+               (e->pixelDelta().y() != 0) ? e->pixelDelta().y() * 120. / e->angleDelta().y() : 0, steps.x(), steps.y());
 
     // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
     // only on linux (x11 and wayland) and Windows, not on macos
     if (e->modifiers().testFlag(Qt::AltModifier) && (isX11 || platform == "wayland" || platform == "windows"))
-        ratio = ratio.transposed();
+        steps = steps.transposed();
 
     // Note: on some platforms (x11, wayland) "natural scrolling" is not detectable: inverted() returns always false
     if (e->inverted()) {
-        ratio.setY(ratio.y() * -1);
+        steps.setY(steps.y() * -1);
     } else {
         // Normally horiz scroll produces positive delta values if the wheel is moved to the left (Qt docs)
-        ratio.setX(ratio.x() * -1);
+        steps.setX(steps.x() * -1);
     };
 
-    return ratio;
-}
-
-static QPointF getNormalizedScrollRatio(const QWheelEvent* e, double pixelRatio) {
-    return getNormalizedScrollRatio(e, QSizeF(pixelRatio, pixelRatio));
+    return steps;
 }
--- a/QtCollider/widgets/QcAbstractStepValue.h
+++ b/QtCollider/widgets/QcAbstractStepValue.h
@@ -21,6 +21,9 @@
 
 #pragma once
 
+#include <QWheelEvent>
+#include <QGuiApplication>
+
 class QcAbstractStepValue {
 protected:
     QcAbstractStepValue(): _shiftScale(100.f), _ctrlScale(10.f), _altScale(0.1f) {}
@@ -37,3 +40,37 @@ private:
     double _ctrlScale;
     double _altScale;
 };
+
+static QPointF getNormalizedScrollRatio(const QWheelEvent* e, QSizeF pixelRatio) {
+    const auto platform = QGuiApplication::platformName();
+    const bool isX11 = platform == "xcb";
+    // angleDelta: relative to 360deg; pixelDelta: relative to widget size
+    QPointF ratio;
+    // Force using angleDelta on X11 (advice from Qt docs)
+    // the check for pixelRatio.isEmpty is to allow forcing angleDelta by passing pixelRatio = 0
+    if (e->pixelDelta().isNull() || pixelRatio.isEmpty() || isX11) {
+        ratio.setX(e->angleDelta().x() / 8. / 360.);
+        ratio.setY(e->angleDelta().y() / 8. / 360.);
+    } else {
+        ratio.setX(e->pixelDelta().x() * pixelRatio.width());
+        ratio.setY(e->pixelDelta().y() * pixelRatio.height());
+    }
+
+    // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
+    // only on linux (x11 and wayland) and Windows, not on macos
+    if (e->modifiers().testFlag(Qt::AltModifier) && (isX11 || platform == "wayland" || platform == "windows"))
+        ratio = ratio.transposed();
+
+    // Normally horiz scroll produces positive delta values if the wheel is moved to the left (Qt docs)
+    ratio.setX(ratio.x() * -1);
+
+    // Note: on some platforms (x11, wayland) "natural scrolling" is not detectable: inverted() returns always false
+    if (e->inverted())
+        ratio.setY(ratio.y() * -1);
+
+    return ratio;
+}
+
+static QPointF getNormalizedScrollRatio(const QWheelEvent* e, double pixelRatio) {
+    return getNormalizedScrollRatio(e, QSizeF(pixelRatio, pixelRatio));
+}
--- a/QtCollider/widgets/QcAbstractStepValue.h
+++ b/QtCollider/widgets/QcAbstractStepValue.h
@@ -61,12 +61,13 @@ static QPointF getNormalizedScrollRatio(const QWheelEvent* e, QSizeF pixelRatio)
     if (e->modifiers().testFlag(Qt::AltModifier) && (isX11 || platform == "wayland" || platform == "windows"))
         ratio = ratio.transposed();
 
-    // Normally horiz scroll produces positive delta values if the wheel is moved to the left (Qt docs)
-    ratio.setX(ratio.x() * -1);
-
     // Note: on some platforms (x11, wayland) "natural scrolling" is not detectable: inverted() returns always false
-    if (e->inverted())
+    if (e->inverted()) {
         ratio.setY(ratio.y() * -1);
+    } else {
+        // Normally horiz scroll produces positive delta values if the wheel is moved to the left (Qt docs)
+        ratio.setX(ratio.x() * -1);
+    };
 
     return ratio;
 }
--- a/QtCollider/widgets/QcAbstractStepValue.h
+++ b/QtCollider/widgets/QcAbstractStepValue.h
@@ -49,14 +49,6 @@ static QPointF getScrollSteps(const QWheelEvent* e) {
     QPointF steps =
         (e->pixelDelta().isNull() || isX11) ? QPointF(e->angleDelta()) / 120. : QPointF(e->pixelDelta()) / 10.;
 
-    // Use angleDelta (and not pixelDelta) for consistency with SC versions before 3.14
-    // QPointF ratio(e->angleDelta().x() / 120., e->angleDelta().y() / 120.);
-
-    if (!e->pixelDelta().isNull())
-        printf("scroll: angleDelta/12(%d,%d) pxDelta(%d,%d) px/step(%f) steps(%f,%f)\n", (e->angleDelta() / 12.).x(),
-               (e->angleDelta() / 12.).y(), e->pixelDelta().x(), e->pixelDelta().y(),
-               (e->pixelDelta().y() != 0) ? e->pixelDelta().y() * 120. / e->angleDelta().y() : 0, steps.x(), steps.y());
-
     // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
     // only on linux (x11 and wayland) and Windows, not on macos
     if (e->modifiers().testFlag(Qt::AltModifier) && (isX11 || platform == "wayland" || platform == "windows"))
--- a/QtCollider/widgets/QcAbstractStepValue.h
+++ b/QtCollider/widgets/QcAbstractStepValue.h
@@ -44,8 +44,9 @@ private:
 static QPointF getScrollSteps(const QWheelEvent* e) {
     // use pixelDelta on systems that support it, except x11 (Qt docs say it's unreliable)
     const auto platform = QGuiApplication::platformName();
-    const bool isX11 = platform == "x11";
-    // angleDelta: map 15deg/step, pixelDelta: map 10px/step (found empirically to match angleDelta)
+    const bool isX11 = platform == "xcb";
+    // angleDelta: map 15deg/step (conventional)
+    // pixelDelta: map 10px/step (found empirically to match angleDelta on wayland)
     QPointF steps =
         (e->pixelDelta().isNull() || isX11) ? QPointF(e->angleDelta()) / 120. : QPointF(e->pixelDelta()) / 10.;
 
--- a/QtCollider/widgets/QcKnob.cpp
+++ b/QtCollider/widgets/QcKnob.cpp
@@ -22,7 +22,6 @@
 #include "QcKnob.hpp"
 #include "../QcWidgetFactory.h"
 #include "../style/routines.hpp"
-#include "widgets/QcAbstractStepValue.h"
 
 #include <QMouseEvent>
 #include <QPainter>
--- a/QtCollider/widgets/QcKnob.cpp
+++ b/QtCollider/widgets/QcKnob.cpp
@@ -31,7 +31,7 @@ using namespace QtCollider;
 
 QC_DECLARE_QWIDGET_FACTORY(QcKnob);
 
-QcKnob::QcKnob(): Style::Client(this), _value(0.f), _step(0.01), _mode(0), _centered(false) {
+QcKnob::QcKnob(): Style::Client(this), _value(0.f), _step(0.01), _mode(0), _centered(false), scrollRemainder(0.) {
     setFocusPolicy(Qt::StrongFocus);
     setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);
     setAttribute(Qt::WA_AcceptTouchEvents);
@@ -78,10 +78,7 @@ void QcKnob::wheelEvent(QWheelEvent* e) {
     const double size = qMin(width(), height()) * PI;
     const double pixelStep = size > 0. ? 1. / size : 0.;
     double step = qMax(_step, pixelStep);
-
-    const double scrollRatio = getNormalizedScrollRatio(e, pixelStep).y();
-
-    double numSteps = scrollRatio / step;
+    double numSteps = getScrollSteps(e).y();
     modifyStep(&step);
     // accumulate fractional numSteps to help scrolling through big steps
     scrollRemainder = std::modf(numSteps + scrollRemainder, &numSteps);
--- a/QtCollider/widgets/QcKnob.cpp
+++ b/QtCollider/widgets/QcKnob.cpp
@@ -80,8 +80,11 @@ void QcKnob::wheelEvent(QWheelEvent* e) {
     // use angleDelta for X11 (advice from Qt docs)
     const bool isX11 = QGuiApplication::platformName() == "xcb";
     // angleDelta: relative to 360deg; pixelDelta: relative to widget size
-    double scrollRatio =
-        (e->pixelDelta().isNull() || isX11) ? e->angleDelta().y() / 8. / 360. : e->pixelDelta().y() * pixelStep;
+    const QPointF scrollRatioXY = (e->pixelDelta().isNull() || isX11) ? e->angleDelta().toPointF() / 8. / 360.
+                                                                      : e->pixelDelta().toPointF() * pixelStep;
+
+    // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
+    double scrollRatio = e->modifiers().testFlag(Qt::AltModifier) ? scrollRatioXY.x() : scrollRatioXY.y();
 
     // note: on some platforms (x11, wayland) "natural scrolling" is not detectable: inverted() returns always false
     if (e->inverted())
--- a/QtCollider/widgets/QcKnob.cpp
+++ b/QtCollider/widgets/QcKnob.cpp
@@ -71,6 +71,31 @@ void QcKnob::mouseMoveEvent(QMouseEvent* e) {
     _prevPos = e->pos();
 }
 
+static const double PI = 3.14159265358979323846264338327950288419717;
+
+void QcKnob::wheelEvent(QWheelEvent* e) {
+    const double size = qMin(width(), height()) * PI;
+    const double pixelStep = size > 0. ? 1. / size : 0.;
+    double step = qMax(_step, pixelStep);
+    // use angleDelta for X11 (advice from Qt docs)
+    const bool isX11 = QGuiApplication::platformName() == "xcb";
+    // angleDelta: relative to 360deg; pixelDelta: relative to widget size
+    double scrollRatio =
+        (e->pixelDelta().isNull() || isX11) ? e->angleDelta().y() / 8. / 360. : e->pixelDelta().y() * pixelStep;
+
+    // note: on some platforms (x11, wayland) "natural scrolling" is not detectable: inverted() returns always false
+    if (e->inverted())
+        scrollRatio *= -1;
+
+    // convert ratio to number of steps (totSteps = 1/step)
+    double numSteps = scrollRatio / step;
+    modifyStep(&step);
+    // accumulate fractional numSteps to help scrolling through big steps
+    scrollRemainder = std::modf(numSteps + scrollRemainder, &numSteps);
+    setValue(_value + numSteps * step);
+    Q_EMIT(action());
+}
+
 void QcKnob::paintEvent(QPaintEvent*) {
     using namespace QtCollider::Style;
     using QtCollider::Style::Ellipse;
--- a/QtCollider/widgets/QcKnob.cpp
+++ b/QtCollider/widgets/QcKnob.cpp
@@ -22,6 +22,7 @@
 #include "QcKnob.hpp"
 #include "../QcWidgetFactory.h"
 #include "../style/routines.hpp"
+#include "widgets/QcAbstractStepValue.h"
 
 #include <QMouseEvent>
 #include <QPainter>
@@ -77,20 +78,9 @@ void QcKnob::wheelEvent(QWheelEvent* e) {
     const double size = qMin(width(), height()) * PI;
     const double pixelStep = size > 0. ? 1. / size : 0.;
     double step = qMax(_step, pixelStep);
-    // use angleDelta for X11 (advice from Qt docs)
-    const bool isX11 = QGuiApplication::platformName() == "xcb";
-    // angleDelta: relative to 360deg; pixelDelta: relative to widget size
-    const QPointF scrollRatioXY = (e->pixelDelta().isNull() || isX11) ? e->angleDelta().toPointF() / 8. / 360.
-                                                                      : e->pixelDelta().toPointF() * pixelStep;
 
-    // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
-    double scrollRatio = e->modifiers().testFlag(Qt::AltModifier) ? scrollRatioXY.x() : scrollRatioXY.y();
+    const double scrollRatio = getNormalizedScrollRatio(e, pixelStep).y();
 
-    // note: on some platforms (x11, wayland) "natural scrolling" is not detectable: inverted() returns always false
-    if (e->inverted())
-        scrollRatio *= -1;
-
-    // convert ratio to number of steps (totSteps = 1/step)
     double numSteps = scrollRatio / step;
     modifyStep(&step);
     // accumulate fractional numSteps to help scrolling through big steps
--- a/QtCollider/widgets/QcKnob.hpp
+++ b/QtCollider/widgets/QcKnob.hpp
@@ -61,6 +61,7 @@ protected:
     virtual void mousePressEvent(QMouseEvent*);
     virtual void mouseMoveEvent(QMouseEvent*);
     virtual void paintEvent(QPaintEvent*);
+    virtual void wheelEvent(QWheelEvent*);
 
 private:
     double value(const QPoint&);
@@ -70,4 +71,5 @@ private:
     int _mode;
     QPoint _prevPos;
     bool _centered;
+    double scrollRemainder;
 };
--- a/QtCollider/widgets/QcNumberBox.cpp
+++ b/QtCollider/widgets/QcNumberBox.cpp
@@ -312,7 +312,9 @@ void QcNumberBox::mouseMoveEvent(QMouseEvent* event) {
 
 void QcNumberBox::wheelEvent(QWheelEvent* event) {
     if (scroll && isReadOnly() && _valueType == Number) {
-        double delta = event->angleDelta().y() / 120.f;
+        // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
+        double delta = event->modifiers().testFlag(Qt::AltModifier) ? event->angleDelta().x() : event->angleDelta().y();
+        delta /= 120.;
         if (event->inverted())
             delta *= -1.;
         scrollRemainder = std::modf(delta + scrollRemainder, &delta);
--- a/QtCollider/widgets/QcNumberBox.cpp
+++ b/QtCollider/widgets/QcNumberBox.cpp
@@ -310,11 +310,9 @@ void QcNumberBox::mouseMoveEvent(QMouseEvent* event) {
 }
 
 void QcNumberBox::wheelEvent(QWheelEvent* event) {
-    if (scroll && isReadOnly() && _valueType == Number && event->angleDelta().x()) {
-        const QPointF delta = event->pixelDelta().isNull()
-            ? event->angleDelta() / 8.f // scaled to return steps of 15
-            : event->pixelDelta() * 0.25f; // this matches old scaling of delta
-        stepBy(delta.x() > 0 ? 1 : -1, scrollStep);
+    const auto delta = event->angleDelta().y();
+    if (scroll && isReadOnly() && _valueType == Number && delta != 0) {
+        stepBy(delta > 0 ? 1 : -1, scrollStep);
         Q_EMIT(action());
     }
 }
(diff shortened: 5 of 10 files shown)


dyfer:
Here's are my tests on M1 macbook pro with internal touchpad:
In comparison to SC 3.13, this PR makes the scroll more sensitive. 
Note that current develop has the scroll less sensitive than 3.13.
I think that with inertia (having the scroll continue after a quicker "flick" of the touchpad) this PR might be too sensitive? However, scrolling without inertia seems good to me.

Scroll without inertia (sliding across full vertical space of the touchpad):

https://github.com/user-attachments/assets/38f42a71-76ab-4ec1-abb2-816c2fcdc5a4



Scroll with inertia (trying to use similar speed on the touchpad on each of the sliders, but that's of course subjective)


https://github.com/user-attachments/assets/5a8b1a50-fc8c-467c-9df2-cbb06505294f

dyfer:
BTW I looked at Qt documentation for this, as I thought there should be a more direct translation from a distance on the touchpad to what's on the screen to get a natural feel. [Qt docs](https://doc.qt.io/qt-6/qwheelevent.html#pixelDelta) suggest to use `numPixels` when available, and `numDegrees` otherwise:

> 
> ```cpp
> void MyWidget::wheelEvent(QWheelEvent *event)
> {
>     QPoint numPixels = event->pixelDelta();
>     QPoint numDegrees = event->angleDelta() / 8;
> 
>     if (!numPixels.isNull()) {
>         scrollWithPixels(numPixels);
>     } else if (!numDegrees.isNull()) {
>         QPoint numSteps = numDegrees / 15;
>         scrollWithDegrees(numSteps);
>     }
> 
>     event->accept();
> }
> ```

dyfer:
More thoughts on this:

For something like a slider, it makes sense to use `pixelData` _directly_ if we can move the slider by the actual num of pixels (we'd need to calculate that).
As for accurate scaling for non-slider objects, I think we'd need to poll screen pixel density for this - that'd give us a relation between the size of physical gesture and the amount of change for a number, possibly consistent across platforms/configurations. So in that case maybe it's better to stick to `angleData` at the cost of smoothness?

dyfer:
Ah, OK. Let's do 

> I'm just not sure if we should go for it now or later :D

Let's do that later maybe. There are a few more things to consider, like handling inverted scrolling (in my tests I had to swipe _down_ to make the slider go up, as I use the "natural scrolling" direction on my system).


> Sounds reasonable, but I'm not sure I would want to go that path, because QSlider has steps. It seems consequential to me to only use numSteps and thus angleDelta.

I'm not sure if the steps are really in the way here. I believe that visual slider should move according to movement?

Also, for non-slider objects, I think it makes sense to stick to angle data, I think this should also be smooth if the hardware/OS supports it.

dyfer:
Another point of information: I was trying to see how scroll to change value is implemented in other macOS audio apps... and it seems that "scroll to change value" is typically _not_ implemented. I'm not saying we should do it one way or the other, but just to raise awareness of what seems to be the "typical" macOS look&feel.

elgiano:
Thanks, I haven't noticed `pixelStep()`, I'm pushing a new version with:
```cpp
double step = qMax(_step, pixelStep());
modifyStep(&step);
const double scrollSteps = e->pixelDelta().isNull() ? e->angleDelta().y() / 120.0 : e->pixelDelta().y();
double dval = scrollSteps * step;
setValue(_value + dval);
```
That is, dval is `pixelDelta() * pixelStep()`. It works well on wayland, and I suppose also on mac but:
- it becomes definitely too fast (unusable) when using default ctrlScale and shiftScale
- on X11, at least on my system, it's too fast. Maybe I have to take into account pixelDensity, but I also see this note from [Qt](https://doc.qt.io/qt-6/qwheelevent.html#pixelDelta):
> Note: On X11 this value is driver-specific and unreliable, use angleDelta() instead.

What do you think?

dyfer:
> Note: On X11 this value is driver-specific and unreliable, use angleDelta() instead.

Ugh, in that case maybe an ifdef for X11? Or is there a Qt flag that tells us if we're in X11 at runtime?

elgiano:
It should be possible to check `QGuiApplication::platformName == "xcb"`  to detect x11.
On the other hand, I don't think I would like to make ctrl_scale and shift_scale unusable when scrolling... have you tried pressing ctrl or shift while scrolling?

dyfer:
I tried the build https://github.com/supercollider/supercollider/pull/7030/commits/50ff352cb2e5be483ea58dbcd76529ee4df0c090

I see two issues: 
- The scroll is a bit too sensitive now. I assume that it should take a similar space on trackpad to drag the slider (click + drag) and two finger scroll, right? The two finger scroll is about 2-3x faster/larger that dragging the handle...
- mouse wheel scroll is waaay to fast. I think the issue is that `step = qMax(_step, pixelStep());` should only be used if pixelDelta is used, but I assume that when mouse is used, then angleDelta is used. That issue was not introduced in this PR, but I think it should be addressed, what do you think?

dyfer:
Well, looking at the code agin, I don't know if using pixelStep is wrong. 

I guess the question is whether we set the scroll wheel sensitivity based on gui size, or not. Intuitively I'd think it makes sense to set it based on the _number of revolutions_ of the scroll wheel,  i.e. one full revolution = full range (or whatever?) of the slider, but maybe that's not a good look and feel? In that case we shouldn't use pixelStep from scroll wheel.
For the touchpad scroll I'd think it should take the same about of space on the touchpad to drag the slider vs two-finger scroll, but that assumption may be wrong.

elgiano:
> Well, looking at the code agin

yeah, this code is agin' :D

I don't have a mouse wheel around to try, so I don't know if pixelDelta is available or not in that case. But theoretically: if `angleDelta * 8` is the number of degrees, and we are calculating 15 degrees per step, we have 24 steps for a full rotation. Default step is 0.1, so it should take around a half rotation (150deg) to cover the full range. We can make that larger, for instance 36 deg/step, so that a full rotation maps to the full range.

I tried another way to calculate the increment from pixelDelta:
```
deltaPixels / totalPixels = deltaSteps / totalSteps
totalSteps = 1 / step

deltaSteps = deltaPixels / totalPixels / step
```
Then I modify step (so that alt_scale and ctrl_scale still work), and get the increment as deltaSteps * modifiedStep.

This for me gives a scroll speed comparable with dragging, on wayland. It is faster than it was in 3.13, but feels more intuitive. On X11 on the other hand, is way too fast, while angleDelta is too slow (especially now that it became slower)...

Let me know how it works for you! I'll sleep on it. We can also just revert everything to angleDelta so to reproduce the 3.13 behavior, and take this later, if it takes too long.

elgiano:
ah interesting thanks!
Could you try to force the use of angleDelta (and perhaps adjust the scale to something like 360 / 2) and see how it all feels?

dyfer:
@elgiano I will try but at a later time. Sorry for the delay. Thanks for working on this!

dyfer:
In [4d7f769](https://github.com/supercollider/supercollider/commit/4d7f769f5aec2d64327944d1777bcecb633a238d) the touchpad feels exactly right for the slider 👍.

For the mouse, I have two comments:

- I feel the movement is generally too small; for a default size Slider (`Slider().front`) I need to make to make approx 12-14 finger movements to go through the full range; if the slider is taller, like screen height, it takes many more turns. `const double degreesPerStep = isMouse ? 4. : 1.;` seemed like a better value, where I needed ~4-5 gestures to move through the default-sized slider
- Do we want the mouse operation to be scaled to the screen resolution, as it is currently, as opposed to slider's full range? With the current implementation the number of turns of the mousewheel is fixed to number of pixels on screen; i.e. larger slider will take more turns to move. Do you feel we should change it so that we always get the same number of turns for the full range of the slider, regardless of the size? In that case `pixelStep` should only be used for the touchpad input, I guess.

dyfer:
All right, after more testing, here are my thoughts on macOS:
- pixelDelta emitted from the mousewheel includes speed-based scaling (faster rotation -> more jump). I _think_ this is more consistent with how other UI elements work on macOS, as it felt more natural, so based on that _I think it'd be beneficial to keep using pixelDelta when available_
  - fast mousewheel movement is a bit too fast in that case, but I think it's okay; slow movement is small enough; and faster movement when pressing alt/option seems to be accurate enough, at least subjectively
- I think that for using angleDelta, it would be better to scale the full range to the number of rotations. I think something like two rotations of the mousewheel equal about 5 finger gestures and that seems to feel good for me for the full range of the slider. I tested this by forcing the condution below with `if (e->pixelDelta().isNull() || isX11 || true)`. Could be also set a bit smaller per step. What do you think?

```cpp
void QcSlider::wheelEvent(QWheelEvent* e) {
    double step;
    double numSteps;
    const bool isX11 = QGuiApplication::platformName() == "xcb";

    // use angleDelta for X11 (advice from Qt docs)
    if (e->pixelDelta().isNull() || isX11) {
        step = qMax(_step, 1.0 / 360.0);
        numSteps = e->angleDelta().y() / 8;
    } else {
        step = qMax(_step, pixelStep());
        // for pixelDelta, map full range to widget dimension (= 1/pixelStep)
        // scrollPx / totPx = scrollSteps / totSteps (totSteps = 1.0 / step)
        numSteps = e->pixelDelta().y() * pixelStep() / step;
    }

    modifyStep(&step);
    setValue(_value + step * numSteps);
    Q_EMIT(action());
}
```

elgiano:
Thanks for all the testing and thoughts!
I see one last problem, when a user sets a rather big step (e.g. 0.25).
- pixelDelta: a single scroll has now to cover at least 0.25 of the widget's size to cause an increment (feels slow or even broken, note that the widget doesn't keep track of the total scroll of the current gesture, so only fast gestures would work, which are hard to calibrate)
- angleDelta: a small gesture now triggers a big increment. it becomes too fast, the opposite of pixelDelta...

When the step is small we want the widget to be slower, when the step is big we want it to be faster... I'm trying to find a suitable mapping... if anyone has any ideas please let me know <3

dyfer:
https://github.com/supercollider/supercollider/pull/7030/commits/6bc691a89d57abcda26b9ef670ed6847cdee45a3 feels great to me on macOS - both as is, and when forcing to use `angleDelta`, e.g. mousewheel (by default mousewheel is reported through `pixelDelta` on macOS). They do feel different, but I believe both behaviors are good.

dyfer:
BTW NumberBox feels a bit too reactive with touchpad on macOS, but it think it's acceptable. With the default `NumberBox().front` it's hard to move it by 1, it typically moves by at least 2; however holding alt/option one gets finer resolution so _think_ it's fine.

dyfer (approved):
I think this is good as is.

Feel free to double check how NumberBox feels with a touchpad on other platforms - I think it feels similarly to SC 3.13 right now on macOS, even though it's hard to change the value by a single step.

editors/sc-ide/widgets/audio_status_box.cpp
```diff
@@ -138,7 +138,7 @@ void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, b
 
 void AudioStatusBox::wheelEvent(QWheelEvent* event) {
     if (!event->angleDelta().isNull()) {
-        if (event->angleDelta().x() > 0)
+        if (event->angleDelta().y() > 0)
```
dyfer: Could you also add inversion here based on `event->inverted()`?

dyfer (dismissed):

editors/sc-ide/widgets/audio_status_box.cpp
dyfer: Thanks!

editors/sc-ide/widgets/audio_status_box.cpp
```diff
@@ -137,8 +137,10 @@ void AudioStatusBox::onServerRunningChanged(bool running, const QString&, int, b
 }
 
 void AudioStatusBox::wheelEvent(QWheelEvent* event) {
-    if (!event->angleDelta().isNull()) {
-        if (event->angleDelta().y() > 0 || event->inverted())
+    // If Alt is pressed, Qt swaps scroll axis: undo it because we use alt to change scale
+    double delta = event->modifiers().testFlag(Qt::AltModifier) ? event->angleDelta().x() : event->angleDelta().y();
+    if (delta) {
+        if (delta > 0 || event->inverted())
```
dyfer: Ah, this is incorrect - now the volume _always_ goes up. ~~Shouldn't it be something like `(delta > 0) * event->inverted().not`? (not the right syntax probably)~~ I don't think that's correct either.

dyfer:
The changes with alt seem incorrect on macOS; most widgets stop moving. On macOS shift is the key that should turn vertical scroll into horizontal.

With Slider2D, this behavior is interesting:
- with touchpad, pressing alt inverts axas (definitely incorrect :) )
- with mousewheel, pressing alt turns vertical scroll into horizonstal - that's okay, but it should happen with shift, not with alt on macOS

dyfer:
> So, did I understand correctly that with the last changes you get axes _inversion_ but not swapping? Does a Slider (1D) move if you scroll it horizontally?

Nope, sorry! Typo.
I get axes swapping with alt. But that's not correct on macOS.
On macOS, shift should swap vertical axis for horizontal, only for the mousewheel (at least this is how scrolling windows with a mouswheel worsk).

elgiano:
Ok, this is a bit of a problem... since we use both shift and alt for alternate scales, we need platform-specific fixes... I'll look more into it

dyfer:
> Does Shift actually swap axes on your mac though?

It does in other apps, where the window can be scrolled both horizontally and vertically, but only for the mousewheel , i.e. mousewheel scrolls window up and down shift+mousewheel scrolls left/right. 
Shift does not change the behavior of the touchpad of course, as touchpad has both horizontal and vertical scrolling.
I can see this behavior e.g. in Chrome, Safari, and SC IDE post window (with word wrap off). 

So all in all, we shouldn't have the _alt_ fixes active on macOS. I'd need to check pn Windows though.

I think I know what's happening with shift on macOS: 
Currently the mousewheel movement of Slider, Knob etc _stops_ when I press shift. I think that's because it changes the scroll axis to the horizontal one (it does not affect movements from the touchpad though; i.e. shift only speeds it up). I don't think there's a clean way to address this though, as both mouse and touchpad come through pixelDelta... unless we use a switch for the mouse device?

So for Slider2D, when I tested before the alt changes were pushed, the behavior with the mousewheel was: when pressing shift, the axis switched to horizontal; but we also increased the step, so the value jumped all the way to left/right (step too large to be useful).

elgiano:
Ok, great, it's starting to become clearer now. Now Alt doesn't affect macos anymore, we need to test it on Windows though... and I've abstracted a function to get scrollRatio, now it's all in one place.

As for the scales: I think it would make sense to make the default scale like 10 times smaller, so that shift and ctrl scales can be used for bigger movements and alt for very fine tunings, and it also would address @bgola's comment. On the other hand I think the pixelDelta() scaling is a quite intuitive criterion (1px scroll -> 1px step). What do you think?

dyfer:
Thanks! Notes for https://github.com/supercollider/supercollider/pull/7030/commits/105f6502b18a6285b0d8202d60b37e5a26e8ea72 on macOS:

Slider, Knob, behave fine with touchpad and mousewheel. Behavior with shift is still noop (more below).

Slider2D: with natural scrolling on, the horizontal axis is now inverted! I think [this](https://github.com/supercollider/supercollider/pull/7030/files#diff-2d13f34154d38777e5066da6796eed224e472905ffcd1970e7881fb2a15b84f4R65) should only happen when the controls are not inverted.

BTW the Volume control still scrolls only up, the `event->inverted()` [here](https://github.com/supercollider/supercollider/pull/7030/files#diff-ff90868f663ef239becc59e4cb451b83de3359e04c49a6ac111dcf745963c040R144) needs to be removed.

Regarding shift:
I propose that we disable increasing the step with shift (and ctrl? is that supposed to change the scrolling speed as well?). On macOS scrolling has built-in scaling based on the gesture speed, and additionally for touchpad only, has inertia. This makes it easy and intuitive to make larger gestures, so I don't think having shift/ctrl is useful. Additionally, shift changes scroll direction so it it prevents vertical widgets from scrolling when pressed (and I don't think we need to work around this).

On more note: it looks like the UI design on macOS these days generally discourages using scroll for changing values. None of Apple apps I tested (Logic Pro etc) allows changing values with a scroll. I'm not saying we should disable it on macOS, but I think we only need to make it work "well enough", not be counterintuitive, not necessarily have all the same functionality as other platforms. In my mind that means that we can forego "shift" on macOS, for example. But that's just my personal opinion.

elgiano:
I have a stronger and stronger feeling that we're over-engineering a seldom used feature... and also that we shouldn't change the behavior when all we wanted to do was to fix a regression.

The "original" behavior is to use angleDelta, assume 15deg/step and apply the number of steps. When no step is defined, it defaults to pixelStep, and it feels very slow because the step is very small. I would accept this because it fits with the metaphor and also because it would be the same behavior as in 3.13.

I would restore this behavior and accept that it feels slow when no step is provided. Alt and Ctrl work on all platforms to use a bigger or smaller step (and I would accept that shift doesn't work on macos, since it seems the axes swap is done by system and not by Qt). So I would also keep the "Alt" and "inverted" countermeasures put so far.

dyfer:
> I agree it feels close to right, but I see three problems:
> 
> * way we are changing the default feeling too much, and it's not configurable

Yeah, the non-configurable part is unfortunate.

> * we are changing the relationship between scroll and steps (from "a fixed amount of scroll applies a step" to "a certain visual proportion applies the same proportional increment")

For the slider, when steps were not defined, weren't we effectively using the visual proportion though? 

> * in dealing with angleDelta and pixelDelta we face a lot of platform-specific differences and implementation details (e.g. `angleDelta = 1.2 * pixelDelta`, but it looks like on a macos touchpad it should be `a = 2 * p`)

Yeah, the platform-specific differences are fragile.

> So, I would like to go back to the previous paradigm, and to use only angleDelta. After all these are not scrolling contents, but adjustable stepping controls.

Okay. This will possibly degrade experience on macOS (angleDelta doesn't have the speed-based scaling and inertia IIUC), but OTOH on macOS scrolling with touchpad seems not really present in the UI paradigm so maybe we don't need to care about it.

> I also thought it would make sense to have a visual paradigm when the step is not defined

Again, isn't this what we had before?

> We could also scale down pixelDelta to be comparable more or less with angleDelta (say, `angleDelta = 1.5 * pixelDelta`) but I don't know how much we could rely on such conversion.

If we use pixelDelta, the current scaling feels good, so I wouldn't change that... after all, pixelDelta should refer to the visual paradigm. But maybe it's okay to revert to angleDelta

More comments:
- I'd keep the "inverted" additions, so that the movement works always in the correct direction
- I'd keep the current implementation around somewhere (maybe a branch on the main repo, linked in the PR's description) for a possible future consideration. In particular, I think we could re-consider using pixelDelta for Slider2D - that one looks like a potentially expressive controller with a touchpad. OTOH, maybe one should use HID for such use case anyway...

elgiano:
I would keep all the collateral improvements: both the inverted and the Alt key thing.

I copied this branch to a new branch on main repo: https://github.com/supercollider/supercollider/tree/topic/qt-scroll-widgets-visual

For this PR, as a last attempt to keep pixelDelta, we can try to map 10px to a step. It seems pretty much equivalent to angleDelta on my system, but I suspect it would be different on mac.

elgiano:
I applied the changes to restore 3.13 behavior, plus the pixelDelta scaling attempt.
I also left in a debug print, among the other things it prints an estimate of px/step: on my system it varies between 7.5 and 15, thus I chose 10 as a reasonable average. I wonder what it prints on macos

dyfer:
Hi @elgiano 



> Almost: we were mapping 15 degrees from angleDelta to a pixelStep. That is: a fixed amount of scroll to an increment that depends on widget size: `degreesScrolled / 15 * pixelStep`, i.e `angleDelta / 120 / widgetSize`.

Hm really? I thought that [this line](https://github.com/supercollider/supercollider/pull/7030/files#diff-480d87f75c40eaca8c16adeba1eace52c326c93034c1c8cda02e753eca4fee57L117) used `pixelDelta` if available? It's not really important, I'm just trying to understand the previous state.





> I applied the changes to restore 3.13 behavior, plus the pixelDelta scaling attempt.

Great, thanks!

> I also left in a debug print, among the other things it prints an estimate of px/step: on my system it varies between 7.5 and 15, thus I chose 10 as a reasonable average. I wonder what it prints on macos

Here's the output on macOS. However, it shows up inconsistently in the post window (doesn't post for a while, then posts a bunch of lines at once) so I'm not sure what movement it actually referred to.

<details>

```
-> a Slider
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,-1) pxDelta(1,-8) px/step(60.000000) steps(0.100000,-0.800000)
scroll: angleDelta/12(0,-2) pxDelta(1,-12) px/step(60.000000) steps(0.100000,-1.200000)
scroll: angleDelta/12(0,-2) pxDelta(1,-9) px/step(60.000000) steps(0.100000,-0.900000)
scroll: angleDelta/12(0,-1) pxDelta(1,-8) px/step(60.000000) steps(0.100000,-0.800000)
scroll: angleDelta/12(0,-1) pxDelta(1,-7) px/step(60.000000) steps(0.100000,-0.700000)
scroll: angleDelta/12(0,-1) pxDelta(1,-6) px/step(60.000000) steps(0.100000,-0.600000)
scroll: angleDelta/12(0,-1) pxDelta(1,-5) px/step(60.000000) steps(0.100000,-0.500000)
scroll: angleDelta/12(0,-1) pxDelta(1,-5) px/step(60.000000) steps(0.100000,-0.500000)
scroll: angleDelta/12(0,-1) pxDelta(0,-4) px/step(60.000000) steps(0.000000,-0.400000)
scroll: angleDelta/12(0,-1) pxDelta(0,-4) px/step(60.000000) steps(0.000000,-0.400000)
scroll: angleDelta/12(0,-1) pxDelta(0,-3) px/step(60.000000) steps(0.000000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(0,-3) px/step(60.000000) steps(0.000000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(0,-3) px/step(60.000000) steps(0.000000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(0,-3) px/step(60.000000) steps(0.000000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(-1,-4) px/step(60.000000) steps(-0.100000,-0.400000)
scroll: angleDelta/12(0,-1) pxDelta(-1,-4) px/step(60.000000) steps(-0.100000,-0.400000)
scroll: angleDelta/12(0,-1) pxDelta(-1,-3) px/step(60.000000) steps(-0.100000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(-1,-3) px/step(60.000000) steps(-0.100000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(-1,-3) px/step(60.000000) steps(-0.100000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(-1,-3) px/step(60.000000) steps(-0.100000,-0.300000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(-1,-1) px/step(60.000000) steps(-0.100000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-2) px/step(60.000000) steps(0.000000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-2) px/step(60.000000) steps(-0.100000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(-1,-1) px/step(60.000000) steps(-0.100000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(-1,-1) px/step(60.000000) steps(-0.100000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(-1,-1) px/step(60.000000) steps(-0.100000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,1) px/step(60.000000) steps(0.000000,0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,1) px/step(60.000000) steps(0.000000,0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,1) px/step(60.000000) steps(0.000000,0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,1) px/step(60.000000) steps(0.000000,0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,1) px/step(60.000000) steps(0.000000,0.100000)
scroll: angleDelta/12(0,1) pxDelta(-1,3) px/step(60.000000) steps(-0.100000,0.300000)
scroll: angleDelta/12(0,2) pxDelta(-1,9) px/step(60.000000) steps(-0.100000,0.900000)
scroll: angleDelta/12(0,2) pxDelta(-1,13) px/step(60.000000) steps(-0.100000,1.300000)
scroll: angleDelta/12(0,2) pxDelta(-1,14) px/step(60.000000) steps(-0.100000,1.400000)
scroll: angleDelta/12(0,2) pxDelta(-1,10) px/step(60.000000) steps(-0.100000,1.000000)
scroll: angleDelta/12(0,1) pxDelta(0,8) px/step(60.000000) steps(0.000000,0.800000)
scroll: angleDelta/12(0,2) pxDelta(0,14) px/step(60.000000) steps(0.000000,1.400000)
scroll: angleDelta/12(0,2) pxDelta(0,14) px/step(60.000000) steps(0.000000,1.400000)
scroll: angleDelta/12(0,3) pxDelta(0,15) px/step(60.000000) steps(0.000000,1.500000)
scroll: angleDelta/12(0,3) pxDelta(0,15) px/step(60.000000) steps(0.000000,1.500000)
scroll: angleDelta/12(0,3) pxDelta(0,15) px/step(60.000000) steps(0.000000,1.500000)
scroll: angleDelta/12(0,2) pxDelta(0,14) px/step(60.000000) steps(0.000000,1.400000)
scroll: angleDelta/12(0,3) pxDelta(0,16) px/step(60.000000) steps(0.000000,1.600000)
scroll: angleDelta/12(0,3) pxDelta(0,16) px/step(60.000000) steps(0.000000,1.600000)
scroll: angleDelta/12(0,2) pxDelta(0,14) px/step(60.000000) steps(0.000000,1.400000)
scroll: angleDelta/12(0,2) pxDelta(0,13) px/step(60.000000) steps(0.000000,1.300000)
scroll: angleDelta/12(0,2) pxDelta(0,12) px/step(60.000000) steps(0.000000,1.200000)
scroll: angleDelta/12(0,2) pxDelta(0,11) px/step(60.000000) steps(0.000000,1.100000)
scroll: angleDelta/12(0,2) pxDelta(0,10) px/step(60.000000) steps(0.000000,1.000000)
scroll: angleDelta/12(0,2) pxDelta(0,9) px/step(60.000000) steps(0.000000,0.900000)
scroll: angleDelta/12(0,1) pxDelta(0,8) px/step(60.000000) steps(0.000000,0.800000)
scroll: angleDelta/12(0,1) pxDelta(0,8) px/step(60.000000) steps(0.000000,0.800000)
scroll: angleDelta/12(0,1) pxDelta(0,7) px/step(60.000000) steps(0.000000,0.700000)
scroll: angleDelta/12(0,1) pxDelta(0,6) px/step(60.000000) steps(0.000000,0.600000)
scroll: angleDelta/12(0,1) pxDelta(0,6) px/step(60.000000) steps(0.000000,0.600000)
scroll: angleDelta/12(0,1) pxDelta(0,5) px/step(60.000000) steps(0.000000,0.500000)
scroll: angleDelta/12(0,1) pxDelta(0,5) px/step(60.000000) steps(0.000000,0.500000)
scroll: angleDelta/12(0,1) pxDelta(0,5) px/step(60.000000) steps(0.000000,0.500000)
scroll: angleDelta/12(0,1) pxDelta(0,4) px/step(60.000000) steps(0.000000,0.400000)
scroll: angleDelta/12(0,1) pxDelta(0,4) px/step(60.000000) steps(0.000000,0.400000)
scroll: angleDelta/12(0,1) pxDelta(0,4) px/step(60.000000) steps(0.000000,0.400000)
scroll: angleDelta/12(0,1) pxDelta(0,3) px/step(60.000000) steps(0.000000,0.300000)
scroll: angleDelta/12(0,1) pxDelta(0,3) px/step(60.000000) steps(0.000000,0.300000)
scroll: angleDelta/12(0,1) pxDelta(0,3) px/step(60.000000) steps(0.000000,0.300000)
scroll: angleDelta/12(0,1) pxDelta(0,3) px/step(60.000000) steps(0.000000,0.300000)
scroll: angleDelta/12(0,0) pxDelta(0,2) px/step(60.000000) steps(0.000000,0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,2) px/step(60.000000) steps(0.000000,0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,-1) pxDelta(-1,-5) px/step(60.000000) steps(-0.100000,-0.500000)
scroll: angleDelta/12(0,-3) pxDelta(-1,-15) px/step(60.000000) steps(-0.100000,-1.500000)
scroll: angleDelta/12(0,-3) pxDelta(-2,-19) px/step(60.000000) steps(-0.200000,-1.900000)
scroll: angleDelta/12(0,-4) pxDelta(-2,-23) px/step(60.000000) steps(-0.200000,-2.300000)
scroll: angleDelta/12(0,-4) pxDelta(-2,-25) px/step(60.000000) steps(-0.200000,-2.500000)
scroll: angleDelta/12(0,-4) pxDelta(-2,-21) px/step(60.000000) steps(-0.200000,-2.100000)
scroll: angleDelta/12(0,-3) pxDelta(-2,-15) px/step(60.000000) steps(-0.200000,-1.500000)
scroll: angleDelta/12(0,-5) pxDelta(0,-30) px/step(60.000000) steps(0.000000,-3.000000)
scroll: angleDelta/12(0,-6) pxDelta(0,-33) px/step(60.000000) steps(0.000000,-3.300000)
scroll: angleDelta/12(0,-6) pxDelta(0,-34) px/step(60.000000) steps(0.000000,-3.400000)
scroll: angleDelta/12(0,-6) pxDelta(0,-34) px/step(60.000000) steps(0.000000,-3.400000)
scroll: angleDelta/12(0,-6) pxDelta(0,-34) px/step(60.000000) steps(0.000000,-3.400000)
scroll: angleDelta/12(0,-6) pxDelta(0,-33) px/step(60.000000) steps(0.000000,-3.300000)
scroll: angleDelta/12(0,-6) pxDelta(0,-36) px/step(60.000000) steps(0.000000,-3.600000)
scroll: angleDelta/12(0,-6) pxDelta(0,-35) px/step(60.000000) steps(0.000000,-3.500000)
scroll: angleDelta/12(0,-5) pxDelta(0,-32) px/step(60.000000) steps(0.000000,-3.200000)
scroll: angleDelta/12(0,-5) pxDelta(0,-29) px/step(60.000000) steps(0.000000,-2.900000)
scroll: angleDelta/12(0,-4) pxDelta(0,-26) px/step(60.000000) steps(0.000000,-2.600000)
scroll: angleDelta/12(0,-4) pxDelta(0,-24) px/step(60.000000) steps(0.000000,-2.400000)
scroll: angleDelta/12(0,-4) pxDelta(0,-22) px/step(60.000000) steps(0.000000,-2.200000)
scroll: angleDelta/12(0,-3) pxDelta(0,-20) px/step(60.000000) steps(0.000000,-2.000000)
scroll: angleDelta/12(0,-3) pxDelta(0,-19) px/step(60.000000) steps(0.000000,-1.900000)
scroll: angleDelta/12(0,-3) pxDelta(0,-17) px/step(60.000000) steps(0.000000,-1.700000)
scroll: angleDelta/12(0,-3) pxDelta(0,-16) px/step(60.000000) steps(0.000000,-1.600000)
scroll: angleDelta/12(0,-2) pxDelta(0,-14) px/step(60.000000) steps(0.000000,-1.400000)
scroll: angleDelta/12(0,-2) pxDelta(0,-13) px/step(60.000000) steps(0.000000,-1.300000)
scroll: angleDelta/12(0,-2) pxDelta(0,-12) px/step(60.000000) steps(0.000000,-1.200000)
scroll: angleDelta/12(0,-2) pxDelta(0,-11) px/step(60.000000) steps(0.000000,-1.100000)
scroll: angleDelta/12(0,-2) pxDelta(0,-10) px/step(60.000000) steps(0.000000,-1.000000)
scroll: angleDelta/12(0,-2) pxDelta(0,-9) px/step(60.000000) steps(0.000000,-0.900000)
scroll: angleDelta/12(0,-1) pxDelta(0,-8) px/step(60.000000) steps(0.000000,-0.800000)
scroll: angleDelta/12(0,-1) pxDelta(0,-8) px/step(60.000000) steps(0.000000,-0.800000)
scroll: angleDelta/12(0,-1) pxDelta(0,-7) px/step(60.000000) steps(0.000000,-0.700000)
scroll: angleDelta/12(0,-1) pxDelta(0,-7) px/step(60.000000) steps(0.000000,-0.700000)
scroll: angleDelta/12(0,-1) pxDelta(0,-6) px/step(60.000000) steps(0.000000,-0.600000)
scroll: angleDelta/12(0,-1) pxDelta(0,-5) px/step(60.000000) steps(0.000000,-0.500000)
scroll: angleDelta/12(0,-1) pxDelta(0,-5) px/step(60.000000) steps(0.000000,-0.500000)
scroll: angleDelta/12(0,-1) pxDelta(0,-5) px/step(60.000000) steps(0.000000,-0.500000)
scroll: angleDelta/12(0,-1) pxDelta(0,-5) px/step(60.000000) steps(0.000000,-0.500000)
scroll: angleDelta/12(0,-1) pxDelta(0,-4) px/step(60.000000) steps(0.000000,-0.400000)
scroll: angleDelta/12(0,-1) pxDelta(0,-4) px/step(60.000000) steps(0.000000,-0.400000)
scroll: angleDelta/12(0,-1) pxDelta(0,-4) px/step(60.000000) steps(0.000000,-0.400000)
scroll: angleDelta/12(0,-1) pxDelta(0,-3) px/step(60.000000) steps(0.000000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(0,-3) px/step(60.000000) steps(0.000000,-0.300000)
scroll: angleDelta/12(0,-1) pxDelta(0,-3) px/step(60.000000) steps(0.000000,-0.300000)
scroll: angleDelta/12(0,0) pxDelta(0,-2) px/step(60.000000) steps(0.000000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-2) px/step(60.000000) steps(0.000000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-2) px/step(60.000000) steps(0.000000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-2) px/step(60.000000) steps(0.000000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-2) px/step(60.000000) steps(0.000000,-0.200000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,0) pxDelta(0,-1) px/step(60.000000) steps(0.000000,-0.100000)
scroll: angleDelta/12(0,1) pxDelta(0,5) px/step(60.000000) steps(0.000000,0.500000)
scroll: angleDelta/12(0,9) pxDelta(0,53) px/step(60.000000) steps(0.000000,5.300000)
scroll: angleDelta/12(0,38) pxDelta(1,230) px/step(60.000000) steps(0.100000,23.000000)
scroll: angleDelta/12(0,46) pxDelta(0,273) px/step(60.000000) steps(0.000000,27.300000)
scroll: angleDelta/12(0,42) pxDelta(0,250) px/step(60.000000) steps(0.000000,25.000000)
scroll: angleDelta/12(0,39) pxDelta(0,231) px/step(60.000000) steps(0.000000,23.100000)
scroll: angleDelta/12(0,36) pxDelta(0,216) px/step(60.000000) steps(0.000000,21.600000)
scroll: angleDelta/12(0,34) pxDelta(0,202) px/step(60.000000) steps(0.000000,20.200000)
scroll: angleDelta/12(0,32) pxDelta(0,190) px/step(60.000000) steps(0.000000,19.000000)
scroll: angleDelta/12(0,28) pxDelta(0,169) px/step(60.000000) steps(0.000000,16.900000)
scroll: angleDelta/12(0,27) pxDelta(0,159) px/step(60.000000) steps(0.000000,15.900000)
scroll: angleDelta/12(0,26) pxDelta(0,154) px/step(60.000000) steps(0.000000,15.400000)
scroll: angleDelta/12(0,25) pxDelta(0,149) px/step(60.000000) steps(0.000000,14.900000)
scroll: angleDelta/12(0,24) pxDelta(0,144) px/step(60.000000) steps(0.000000,14.400000)
scroll: angleDelta/12(0,23) pxDelta(0,139) px/step(60.000000) steps(0.000000,13.900000)
scroll: angleDelta/12(0,22) pxDelta(0,134) px/step(60.000000) steps(0.000000,13.400000)
scroll: angleDelta/12(0,22) pxDelta(0,129) px/step(60.000000) steps(0.000000,12.900000)
scroll: angleDelta/12(0,21) pxDelta(0,124) px/step(60.000000) steps(0.000000,12.400000)
scroll: angleDelta/12(0,20) pxDelta(0,119) px/step(60.000000) steps(0.000000,11.900000)
scroll: angleDelta/12(0,19) pxDelta(0,113) px/step(60.000000) steps(0.000000,11.300000)
scroll: angleDelta/12(0,18) pxDelta(0,109) px/step(60.000000) steps(0.000000,10.900000)
scroll: angleDelta/12(0,18) pxDelta(0,105) px/step(60.000000) steps(0.000000,10.500000)
scroll: angleDelta/12(0,17) pxDelta(0,100) px/step(60.000000) steps(0.000000,10.000000)
scroll: angleDelta/12(0,16) pxDelta(0,96) px/step(60.000000) steps(0.000000,9.600000)
scroll: angleDelta/12(0,15) pxDelta(0,90) px/step(60.000000) steps(0.000000,9.000000)
scroll: angleDelta/12(0,14) pxDelta(0,85) px/step(60.000000) steps(0.000000,8.500000)
scroll: angleDelta/12(0,14) pxDelta(0,81) px/step(60.000000) steps(0.000000,8.100000)
scroll: angleDelta/12(0,13) pxDelta(0,77) px/step(60.000000) steps(0.000000,7.700000)
scroll: angleDelta/12(0,12) pxDelta(0,72) px/step(60.000000) steps(0.000000,7.200000)
scroll: angleDelta/12(0,11) pxDelta(0,68) px/step(60.000000) steps(0.000000,6.800000)
scroll: angleDelta/12(0,10) pxDelta(0,62) px/step(60.000000) steps(0.000000,6.200000)
scroll: angleDelta/12(0,10) pxDelta(0,58) px/step(60.000000) steps(0.000000,5.800000)
scroll: angleDelta/12(0,9) pxDelta(0,55) px/step(60.000000) steps(0.000000,5.500000)
scroll: angleDelta/12(0,9) pxDelta(0,51) px/step(60.000000) steps(0.000000,5.100000)
scroll: angleDelta/12(0,8) pxDelta(0,47) px/step(60.000000) steps(0.000000,4.700000)
scroll: angleDelta/12(0,7) pxDelta(0,43) px/step(60.000000) steps(0.000000,4.300000)
scroll: angleDelta/12(0,7) pxDelta(0,41) px/step(60.000000) steps(0.000000,4.100000)
scroll: angleDelta/12(0,6) pxDelta(0,37) px/step(60.000000) steps(0.000000,3.700000)
scroll: angleDelta/12(0,6) pxDelta(0,34) px/step(60.000000) steps(0.000000,3.400000)
scroll: angleDelta/12(0,5) pxDelta(0,32) px/step(60.000000) steps(0.000000,3
```
</details>


I think the slower scroll on macOS with touchpad is fine, with inertia etc it's easy enough to change the value I think, even if it doesn't correspond to the physical distance on screen.

Also, we were hoping with @capital-G to release 3.14.0-rc2 in about 24h... (after 18:00 UTC on Monday), do you think this is good in the current form (minus the printing of course)? Do we need more testing on Linux - X11 or Waylwand?

Also, I tested this build on Windows 10. Scrolling slow, but I think it's okay. Curiously, inverted scroll setting is not detected (i.e. when turned on, the widgets scroll in opposite direction).

elgiano:
Ok, thanks! I didn't expect such a difference in px/step, but since you say it feels ok I would leave it as it is.
I'll clean up a bit and push the final version before tomorrow!

> Hm really? I thought that [this line](https://github.com/supercollider/supercollider/pull/7030/files#diff-480d87f75c40eaca8c16adeba1eace52c326c93034c1c8cda02e753eca4fee57L117) used pixelDelta if available? It's not really important, I'm just trying to understand the previous state.

This line was changed in 9d10e196ce263faff1f9951d0db012a948ffbc87, which is the commit that replaced the deprecated `delta()`. Although that commit is from 2021, it was merged only in https://github.com/supercollider/supercollider/pull/6464, for 3.14. Here is what happened: https://github.com/supercollider/supercollider/commit/9d10e196ce263faff1f9951d0db012a948ffbc87#diff-480d87f75c40eaca8c16adeba1eace52c326c93034c1c8cda02e753eca4fee57L117
Before that change, in 3.13, we were only using angleDelta / 120. After, we were using either angleDelta / 8160 or pixelDelta / 120. Although I believe pixelDelta / 120 works ok on mac, this made scrolling on X and Wayland much much slower.

dyfer:
> Ok, thanks! I didn't expect such a difference in px/step,

I think this happens when you "fling" fingers over touchpad to quickly scroll.
But again, because of the delayed posting (I wonder what's that about?) I couldn't confirm for sure.

dyfer:
> I meant the difference in px/step = 60 (which never changes in your example), compared to the values I get on Wayland (~10) and X (120, but anyway we are not using pxDelta on X).

Ah, got it.

elgiano:
I had a typo in the last commit (checking "x11" instead of "xcb").
Also, I didn't understand that scrolling became much faster on macos, it doesn't on wayland... and that's the difference between 10 px/step and 60 px/step. The behavior with defined steps it's also the same as in 3.13.

I think in the future if we want to use pixelDelta, we have to to switch to the visual paradigm, which also makes sense with high-res scrolling to make small or big adjustments. But I also don't want to keep going in circles :D so if this behavior is good enough for you, let's go with it and eventually change it in a later release.

Alternatively, we can still go back to using only angleDelta to match perfectly the 3.13 behavior.

dyfer:
@elgiano this is good for me as is, if you can vouch for behavior on Linux. I don't want to keep going in circles either. 
I think we could consider the visual scrolling paradigm, but let's leave that for the future.
This PR makes scrolling operational - "good enough" - on all platforms IIUC (I haven't tested on Linux).

capital-G:
What a ride - thanks gianluca and marcin for tackling this <3

PR #6972, opened 2025-06-11
Support for NO_LIBSNDFILE builds

passyur:
## Purpose and Motivation

Fixes #6103 and allows for builds with the NO_LIBSNDFILE flag. Though not a recommended flag to toggle on, NO_LIBSNDFILE still seems worth having for debug purposes and enjoys support from scsynth. It does not have QtCollider support (due to SoundfileView), nor supernova support. This PR adds both.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation
- [ ] This PR is ready for review

PR #7035, opened 2025-07-07
standardize clarifying comments for preprocessor conditionals (redux)

passyur:
## Purpose and Motivation

Invalidated commits from #7034, so those changes are here now.

Originally mentioned in https://github.com/supercollider/supercollider/pull/6972#discussion_r2180677561.

For readability, it would be nice if comments used a consistent style for clarifying the branch taken for preprocessor `#ifdef`'ing. There are several places where this isn't the case. The style this PR 'adheres' to follows these sort of patterns

```c
#ifdef MACRO

#else // not MACRO (or !MACRO)

#endif // MACRO
```

```c
#ifndef MACRO

#else /* MACRO */

#endif /* not MACRO (or !MACRO) */
```

```c
#ifdef MACRO_1

#elif MACRO_2

#elif MACRO_3

#else // !MACRO_1 && !MACRO_2 && !MACRO_3

#endif // MACRO_1 || MACRO_2 || MACRO_3
```

Comment style is only relevant as a matter of consistency within the file itself (so if the file uses C-style for most preprocessor comments, then all preprocessor comments should be C-style, and so on).

## Types of changes

- Style

## To-do list

### DIFF
--- a/SCDoc/SCDoc.tab.cpp
+++ b/SCDoc/SCDoc.tab.cpp
@@ -1229,7 +1229,7 @@ int yydebug;
 # define YY_SYMBOL_PRINT(Title, Kind, Value, Location)
 # define YY_STACK_PRINT(Bottom, Top)
 # define YY_REDUCE_PRINT(Rule)
-#endif /* !YYDEBUG */
+#endif /* YYDEBUG */
 
 
 /* YYINITDEPTH -- initial size of the parser's stacks.  */
--- a/SCDoc/SCDoc.tab.cpp
+++ b/SCDoc/SCDoc.tab.cpp
@@ -1229,7 +1229,7 @@ int yydebug;
 # define YY_SYMBOL_PRINT(Title, Kind, Value, Location)
 # define YY_STACK_PRINT(Bottom, Top)
 # define YY_REDUCE_PRINT(Rule)
-#endif /* YYDEBUG */
+#endif /* !YYDEBUG */
 
 
 /* YYINITDEPTH -- initial size of the parser's stacks.  */
--- a/lang/LangSource/Bison/lang11d_tab.cpp
+++ b/lang/LangSource/Bison/lang11d_tab.cpp
@@ -1615,7 +1615,7 @@ int yydebug;
 # define YY_SYMBOL_PRINT(Title, Kind, Value, Location)
 # define YY_STACK_PRINT(Bottom, Top)
 # define YY_REDUCE_PRINT(Rule)
-#endif /* YYDEBUG */
+#endif /* !YYDEBUG */
 
 
 /* YYINITDEPTH -- initial size of the parser's stacks.  */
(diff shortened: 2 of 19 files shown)


capital-G:
Thx - I think it would be good to have some consistency here in order to avoid confusion and to avoid discussions about styleguides.

At the same time: It seems our formatter clang-format doesn't offer a way to automate this, therefore creating maintenance trouble by pointing out consistency problems if they appear. I'd therefore favor to delete any comments concerning the preprocessor encapsulations. Do you think this could be a way to go?

SCDoc/SCDoc.tab.cpp
```diff
@@ -1229,7 +1229,7 @@ int yydebug;
 # define YY_SYMBOL_PRINT(Title, Kind, Value, Location)
 # define YY_STACK_PRINT(Bottom, Top)
 # define YY_REDUCE_PRINT(Rule)
-#endif /* !YYDEBUG */
+#endif /* YYDEBUG */
```
JordanHendersonMusic: This file is generated by bison, you shouldn't edit it.

lang/LangSource/Bison/lang11d_tab.cpp
```diff
@@ -1615,7 +1615,7 @@ int yydebug;
 # define YY_SYMBOL_PRINT(Title, Kind, Value, Location)
 # define YY_STACK_PRINT(Bottom, Top)
 # define YY_REDUCE_PRINT(Rule)
-#endif /* !YYDEBUG */
+#endif /* YYDEBUG */
```
JordanHendersonMusic: Same here, its a generated file.

passyur:
> clang-format doesn't offer a way to automate this

I see. It seems clang-format's way of handling this is an indent for 'nested' conditionals, hence: [`IndentPPDirectives`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#indentppdirectives). The `.clang-format` currently has `PPDIS_AfterHash`, but I... don't like how that looks. They added `PPDIS_BeforeHash` a few years ago, which I think looks better. Thoughts?

`PPDIS_AfterHash`:
```c
#if FOO
#  if BAR
#    include <foo>
#  endif
#endif 
```

`PPDIS_BeforeHash`:
```c
#if FOO
  #if BAR
    #include <foo>
  #endif
#endif
```

> delete any comments concerning the preprocessor encapsulations

I am in favor of having the formatter 'enforce' the style here, though, since that would alleviate some of the labor from a code reviewer. So, I will remove the comments.

SCDoc/SCDoc.tab.cpp
passyur: Thanks for pointing that out!

passyur:
Closing this since it's pretty old and minor. Might come back to it if I decided I misjudged. 🙃

PR #1333, opened 2015-03-06
Scide/autocomplete help

vdonnefort:
Here's a new feature for scide. This PR adds help informations about selected class when showing autocomplete menu.

PR #7036, opened 2025-07-08
IDE: don't call resize(0, 0) on completion widget

### LINKED ISSUE #7007: scide crashes when running on second monitor (3.14-rc1/Linux/Wayland/Sway)
## Environment

* SuperCollider version: 3.14-rc1
* Operating system: ArchLinux 6.15.4-arch2-1 
* Other details (Qt version, audio driver, etc.): Qt 6.9.1

## Steps to reproduce

Open scide on second monitor, type the name of any class, when it tries to auto-complete the class name, the IDE crashes. 

When running from the terminal, I get this error message:

```
% scide
[destroyed object]: error 0: width and height must be positive and non-zero
```

It doesn't happen 100% of the time, but most of the time (without any changes to how I am running the IDE).

When it does work, the auto-complete help window shows in the wrong location. With 3.13.1 all works as expected.

I will investigate this further, but adding the issue here already in case someone else also is running into this.

dyfer:
## Purpose and Motivation

Fixes #7007, according to a [user report](https://github.com/supercollider/supercollider/issues/7007#issuecomment-3049095148).

But... what does it break instead?

This is a somewhat blind attempt at fixing a crash on Wayland with multiple displays (see #7007 for bug description).

I don't know what was the purpose of `resize(0, 0)` in the first place though. For reference it was introduced in  https://github.com/supercollider/supercollider/commit/7f2db98b85e0cbdda03a55d6beb0ec8166e2ab39 as part of https://github.com/supercollider/supercollider/pull/1333. On the surface the autocomplete hints still show as they have before this PR, as far as I can tell...

I'd appreciate if others could test to make sure this doesn't break any functionality!

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/code_editor/completion_menu.cpp
+++ b/editors/sc-ide/widgets/code_editor/completion_menu.cpp
@@ -58,10 +58,7 @@ CompletionMenu::CompletionMenu(QWidget* parent): PopUpWidget(parent), mCompletio
 
 void CompletionMenu::addItem(QStandardItem* item) { mModel->appendRow(item); }
 
-void CompletionMenu::adapt() {
-    mListView->setFixedWidth(mListView->sizeHintForColumn(0));
-    resize(0, 0);
-}
+void CompletionMenu::adapt() { mListView->setFixedWidth(mListView->sizeHintForColumn(0)); }
 
 void CompletionMenu::addInfo(QString info) {
     mTextBrowser->setText(info);
--- a/editors/sc-ide/widgets/code_editor/completion_menu.cpp
+++ b/editors/sc-ide/widgets/code_editor/completion_menu.cpp
@@ -58,7 +58,10 @@ CompletionMenu::CompletionMenu(QWidget* parent): PopUpWidget(parent), mCompletio
 
 void CompletionMenu::addItem(QStandardItem* item) { mModel->appendRow(item); }
 
-void CompletionMenu::adapt() { mListView->setFixedWidth(mListView->sizeHintForColumn(0)); }
+void CompletionMenu::adapt() {
+    mListView->setFixedWidth(mListView->sizeHintForColumn(0));
+    adjustSize();
+}
 
 void CompletionMenu::addInfo(QString info) {
     mTextBrowser->setText(info);


capital-G:
> I don't know what was the purpose of `resize(0, 0)` in the first place though

In case you haven't checked, quoting https://doc.qt.io/qt-6/qwidget.html#size-prop

> Note: Setting the size to QSize(0, 0) will cause the widget to not appear on screen. This also applies to windows.

dyfer:
Thanks for testing!

> In case you haven't checked, quoting https://doc.qt.io/qt-6/qwidget.html#size-prop

I didn't think to check that, sorry. Thanks for the reference!

So when does this function actually triggers closing the widget? I didn't catch the difference in behavior in this PR.

> Just tested 3.14-rc1 on fedora w/ wayland and couldn't trigger a crash, 

Did you test with a secondary display?

capital-G:
> Did you test with a secondary display?

Yes, but gnome and not sway.

editors/sc-ide/widgets/code_editor/completion_menu.cpp
```diff
@@ -58,10 +58,7 @@ CompletionMenu::CompletionMenu(QWidget* parent): PopUpWidget(parent), mCompletio
 
 void CompletionMenu::addItem(QStandardItem* item) { mModel->appendRow(item); }
 
-void CompletionMenu::adapt() {
-    mListView->setFixedWidth(mListView->sizeHintForColumn(0));
-    resize(0, 0);
```
elgiano, line 63: maybe substitute `resize(0,0)` with `adjustSize()`? Not fundamental though: it doesn't make a difference now because `adapt()`is currently called only before the widget is shown for the first time

elgiano (approved):

editors/sc-ide/widgets/code_editor/completion_menu.cpp
dyfer: I changed it to `adjustSize()`. Would there be any possible side effects of this? I haven't read through the code closely.

@bgola could you test https://github.com/supercollider/supercollider/pull/7036/commits/d3c4e4a9cd8eef5766a5b74f50e7823ba83fac37 on your system?

dyfer: @elgiano do you think we should risk adding this for 3.14.0-rc2 ? From the discussion so far it seems that the original bug only applies to a very specific environment configuration, is hard to reproduce, and I don't have full grasp of what's actually being changed by either removing the resize or changing it to `adjustSize`.

PR #7039, opened 2025-07-09
Server: Add /g_dumpTree end delimiter

josephine-wolf-oberholtzer:
## Purpose and Motivation

It's well known that large node trees can overflow `/g_queryTree` OSC packets. An alternative approach to gathering the complete node tree output is to read it from the server process's stdout. However, there's no way to know _explicitly_ when the output has completed. One can attempt contrived methods of guaranteeing completion: sleeping, dumping the node tree twice, trying to sync, etc. But the simplest means of knowing the dump has completed is to just add an end delimiter to the output.

This PR adds an `END NODE TREE Group {node id}` message to the server output as the final line of `/g_dumpTree` output for both `scsynth` and `supernova`.

PS: Why do I want the node tree output? I use when unit-testing changes to complex server setups.

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/server/supernova/sc/sc_osc_handler.cpp
+++ b/server/supernova/sc/sc_osc_handler.cpp
@@ -1473,6 +1473,7 @@ void g_dump_tree(int id, bool flag) {
     stream << "NODE TREE Group " << id << std::endl;
 
     g_dump_node(stream, *node, flag, 1);
+    stream << "END NODE TREE Group " << id << std::endl;
     log(stream.str().c_str(), stream.str().size());
 }
 
--- a/server/scsynth/SC_Group.cpp
+++ b/server/scsynth/SC_Group.cpp
@@ -107,6 +107,8 @@ void Group_DumpNodeTree(Group* inGroup) {
         child = next;
     }
     tabCount--;
+    if (tabCount == 0)
+        scprintf("END NODE TREE Group %d\n", inGroup->mNode.mID);
 }
 
 void Group_DumpNodeTreeAndControls(Group* inGroup) {
@@ -179,6 +181,8 @@ void Group_DumpNodeTreeAndControls(Group* inGroup) {
         child = next;
     }
     tabCount--;
+    if (tabCount == 0)
+        scprintf("END NODE TREE Group %d\n", inGroup->mNode.mID);
 }
 
 void Group_CalcDumpTree(Group* inGroup) {


josephine-wolf-oberholtzer:
I'd love to write unit tests for this change, but I don't know how. Any guidance would be appreciated.

dyfer:
Hi @josephine-wolf-oberholtzer 
Just FYI we're focusing our resources on getting 3.14.0 out the door for the next couple of weeks.

I don't _think_ we have any guarantees regarding the terminal output of scsynth, so I don't think this will break anything. I personally think this change is fine, but I'd like more eyes on this.

> I'd love to write unit tests for this change, but I don't know how. Any guidance would be appreciated.

Ah, great idea, but I have no idea how to approach a unit test for this...

josephine-wolf-oberholtzer:
> Hi @josephine-wolf-oberholtzer 
> 
> Just FYI we're focusing our resources on getting 3.14.0 out the door for the next couple of weeks.

That's fine - no big rush. I don't know what your branching strategy is, but I'll be happy if this lands on `develop`, which is the branch of SC I typically build in GHA for use in my test suites.

> I don't _think_ we have any guarantees regarding the terminal output of scsynth, so I don't think this will break anything. I personally think this change is fine, but I'd like more eyes on this.

I couldn't find any guarantees anywhere, but happy for others to weigh in.

dyfer:
@josephine-wolf-oberholtzer are you still planning to add tests to this?

josephine-wolf-oberholtzer:
@dyfer I am, I've just been busy with life. Hopefully sometime in the end of September.

josephine-wolf-oberholtzer:
@dyfer I have some basic unit tests for the changes against both `scsynth` and `supernova`. Please let me know if this looks idiomatic.

I used `Pipe` instead of `Server` to work around how `Server`'s use of `unixCmd` hijacks `stdout` to the post window instead of a stream I can read from directly.

testsuite/classlibrary/TestServer_dumpTree.sc
```diff
@@ -0,0 +1,68 @@
...
+			// supernova's /g_dumpTree output is different
+			programAndExpectedOutput.put(
+				"supernova",
+				List[
+					"NODE TREE Group 0",
+					"   0 group",
+					"END NODE TREE Group 0",
+				]
+			);
+		};
```
dyfer: I think the `program` part could be taken from `Server.program` (you need to call `Server.scsynth` / `Server.supernova` class method first - that interface is peculiar).

For the purpose of unit tests there's maybe no need to check if supernova exists, but if you wanted to do that, I'd suggest `File.exists` on the full path (from `Server.program`) - I hope that works, haven't tried.

dyfer:
Thanks! ~~Aside from the comment above, I'll look into the structure of the unit test at a later date (just FYI).~~

Okay, I think I got it.

I think the functionality of `programAndExpectedOutput.keysValuesDo` should go in a helper function (taking server path as an argument) that return the contents of the command output.

Then I'd suggest making two tests:
`test_scsynth_dumpTree` and `test_supernova_dumpTree`
where you get the server path, call the helper function, analyze the output, and assert the result.

See e.g. https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestUGen_RTAlloc.sc for using a helper method in the test suite - it can be named whatever as long as it doesn't start with `test_`, _I think_.

dyfer:
> I used `Pipe` instead of `Server` to work around how `Server`'s use of `unixCmd` hijacks `stdout` to the post window instead of a stream I can read from directly.

Would `.unixCmdGetStdOut` work though?

EDIT: maybe pipe is better after all.

testsuite/classlibrary/TestServer_dumpTree.sc
josephine-wolf-oberholtzer: The only unit test that seems to run against `supernova` is this one: https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestEnvGen_server.sc#L75

I borrowed the pattern from there.

Unless there's some magic that makes every test run twice, once for each server program...?

dyfer: There's no magic and no good coverage for supernova so far.

Thanks for pointing out the pattern - it should be improved for sure.

I think for this particular test it would be great to be explicit about supernova tests and make separate test call for each server, as it deals with specific output of the two programs.

dyfer: BTW looks like the supernova test failed?

josephine-wolf-oberholtzer: 1) Yes, you tend to get warnings about threads and/or locking down memory when you run the server(s) in GHA. I'm working on a fix to filter out these warnings from the captured stdout

2) So I never work in sclang, only in Python (or occasionally other languages), and when I write tests in Python using the [pytest](https://docs.pytest.org/en/stable/) library I make use of parameterized test fixtures which parametrically expand my tests e.g. run every test with `scsynth` and then with `supernova`. That saves me the trouble of writing a separate test for each scenario, or bundling multiple scenarios into the same test using loops: https://github.com/supriya-project/supriya/blob/main/tests/contexts/test_Server_lifecycle.py#L118
\
Haven't the faintest idea how you'd build out that functionality in sclang here, but something to think about.

3) Looks like there's also failures under OSX because neither `scsynth` nor `supernova` are on the shell path. Will also work out a portable fix for this.

dyfer: 1. sounds good
2. yeah, I'm not saying our test framework is great; you asked about being idiomatic (in relation to our tests) so I think this is where my suggestion is coming from. Thank you for making the changes.
3. As I wrote in the previous comment, please use `Server.program` class method, which provides a full path for the server on each platform.

josephine-wolf-oberholtzer:
> > I used `Pipe` instead of `Server` to work around how `Server`'s use of `unixCmd` hijacks `stdout` to the post window instead of a stream I can read from directly.
> 
> Would `.unixCmdGetStdOut` work though?
> 
> EDIT: maybe pipe is better after all.

`.unixCmdGetStdOut` uses a `Pipe` and calls `Pipe.close` after reading - it's effectively synchronous, and doesn't give me the opportunity to send OSC messages. This will hang when used against a running server process, forcing me to put it into a thread (a `Routine`? Dunno what you call things around here), which just makes the whole thing more complicated than using `Pipe` directly.

dyfer:
> `.unixCmdGetStdOut` uses a `Pipe` and calls `Pipe.close` after reading - it's effectively synchronous, and doesn't give me the opportunity to send OSC messages.

You are of course correct, my bad, sorry for suggesting otherwise. Pipe is the only non-blocking way to go.

josephine-wolf-oberholtzer:
One transient error in the Windows tests, unrelated to my changes: https://github.com/supercollider/supercollider/actions/runs/17576593543/job/50020755468?pr=7039

But my new tests pass there.

dyfer:
> One transient error in the Windows tests, unrelated to my changes:

Yes, I've seen it before, but couldn't track it down yet. I'm re-running the windows tests CI.

Pinging "server council members" - @elgiano and @Spacechild1: Could you take a look at this PR?

testsuite/classlibrary/TestServer_dumpTree.sc
```diff
@@ -0,0 +1,64 @@
...
+        if (pipe.notNil) {
+            pipe.close;
+        };
+    }
+
+    getOutput { |program, expectedOutput|
+        var actualOutput = List[],
+            addr = NetAddr("127.0.0.1", 57110),
+            line;
+        pipe = Pipe.new(program ++ ServerOptions.new.asOptionsString, "r");
```
dyfer: Sorry, I just realized something.
57110 should probably be a variable - use it for NetAddr AND as an argument for ServerOptions - `ServerOptions().asOptionsString(port)` - just so there's no ambiguity there.

I guess `ServerOptions().bindAddress_(ip)` could be also specified explicitly. So:
```supercollider
ip = "127.0.0.1";
port = 57110;
addr = NetAddr(ip, port);

pipe = Pipe.new(program ++ ServerOptions().bindAddress_(ip).asOptionsString(port), "r")
```

testsuite/classlibrary/TestServer_dumpTree.sc
```diff
@@ -0,0 +1,64 @@
...
+        addr.sendMsg("/g_dumpTree", 0, 0);
+        addr.sendMsg("/quit");
+        // consume lines until no more remain
+        line = pipe.getLine;
+        while ({ line.notNil }) {
+            // This warning can appear in GitHub Actions runs for supernova
+            if (
+                line != "Warning: cannot raise thread priority",
+                { actualOutput.add(line); }
+            );
```
dyfer: There may be a few more warnings in supernova, e.g. https://github.com/supercollider/supercollider/blob/develop/server/supernova/server/server.cpp#L323

Maybe instead of matching an exact line, it would be reasonable to check for anything starting with "Warning: "?

Also, IMO no need to call out Github Actions on this, these warnings do happen on user machines too sometimes (but feel free to leave it, that's purely cosmetic; it's more of an FYI that it's not explicitly an issue with GHA).

dyfer (changes requested):

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -1473,6 +1473,7 @@ void g_dump_tree(int id, bool flag) {
     stream << "NODE TREE Group " << id << std::endl;
 
     g_dump_node(stream, *node, flag, 1);
+    stream << "END NODE TREE Group " << id << std::endl;
```
Spacechild1, line 1476: This is not really related to this PR, but can you do me a favor and replace
```
log(stream.str().c_str(), stream.str().size());
```
with
```
auto msg = stream.str();
log(msg.c_str(), msg.size());
```
The `str()` method actually returns a `std::string` object, not a reference, so the current code accidentally makes an extra copy of the whole message just to get its size...

Spacechild1 (approved):

Spacechild1:
The Server code looks fine to me! I think this is a benign change, so I have no objections.

> It's well known that large node trees can overflow /g_queryTree OSC packets. An alternative approach to gathering the complete node tree output is to read it from the server process's stdout.

Personally, I find this very brittle, particularly since there is no stability guarantee for scsynth's console output.

Usually, scsynth/sclang deal with the size restriction of UDP packets by sending data via temp files. If we are looking for an *actual* solution for the size limitations of `/g_queryTree`, we could add a new Server command that writes the OSC message to a file. (The size limit of OSC messages is 2 GB, which should be large enough to represent any Server tree :) But this probably the topic for another issue/PR.

server/supernova/sc/sc_osc_handler.cpp
josephine-wolf-oberholtzer: Can we do that in a separate PR?

josephine-wolf-oberholtzer:
@Spacechild1 1) what's wrong with stdout? and 2) why isn't there a guarantee for scsynth/supernova's console output and why shouldn't we codify it?

testsuite/classlibrary/TestServer_dumpTree.sc
josephine-wolf-oberholtzer: You're right, there's probably a couple additional warnings that crop up. I'm not 100% sure they all get announced with `Warning:` but that seems like a reasonable generalization. Will fix.

server/supernova/sc/sc_osc_handler.cpp
Spacechild1: It just feels a bit overkill to make a dedicated PR for such a trivial fix. But if you don't mind...

Spacechild1:
> 1) what's wrong with stdout? and 2) why isn't there a guarantee for scsynth/supernova's console output and why shouldn't we codify it?

There is nothing wrong with stdout per se. It's just that scsynth is not a command line utility that writes to stdout. scsynth's console output is purely informative and we want to have the freedom to change it at any point. The public API is the OSC interface.

Also keep in mind that scsynth and supernova are multi-threaded programs. Threads can print to stdout concurrently, which causes messages to be interleaved.

server/supernova/sc/sc_osc_handler.cpp
josephine-wolf-oberholtzer: OK, I'll do it based on your word.

josephine-wolf-oberholtzer: Done.

testsuite/classlibrary/TestServer_dumpTree.sc
josephine-wolf-oberholtzer: Done.

testsuite/classlibrary/TestServer_dumpTree.sc
josephine-wolf-oberholtzer: Done.

josephine-wolf-oberholtzer:
@Spacechild1 Hear you about the poor interaction between multi-threading and multi-line logging. For my purposes, this is all perfectly fine. Like the PR description says, I'm using this for unit testing. Set up a server, make a set of operations against, validate the (large) node tree. It's not happening during real-world performance. It's happening in very controlled CI/CD environments. Personally, would prefer something struct-log-like, e.g. single-line JSON in and out over stdin/stdout. I don't really see a problem with heavy logging from applications like this. But that's for another implementation I'm sure.

Thank you for the 👍🏻 🙇🏻‍♀️

server/supernova/sc/sc_osc_handler.cpp
Spacechild1: Thanks! :)

Spacechild1:
Ok for me!

elgiano:
Looks good to me too, merged! Thanks @josephine-wolf-oberholtzer for making it and @dyfer and @Spacechild1 for reviewing it!

josephine-wolf-oberholtzer:
Thank you everyone for shepherding my first contribution 💓 ! And my test suite thanks you too 😉.

PR #6994, opened 2025-06-16
servers [windows]: select WASAPI as default device if available

dyfer:
## Purpose and Motivation

Use WASAPI for the default device selection on Windows.

We seem to need this before we can disable old APIs, see https://github.com/supercollider/supercollider/pull/6900#issuecomment-2914914392

## Types of changes

- New feature (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #7051, opened 2025-07-16
Fix default device selection on Windows when no WASAPI devices are available

dyfer:
## Purpose and Motivation

It seems that #6994 introduced an issue: when WASAPI device is not present/available, but ASIO is, ASIO was not selected by default anymore. See https://github.com/supercollider/supercollider/pull/6994#issuecomment-3079941850

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested: I tried this on the [windows tests PR](https://github.com/supercollider/supercollider/pull/6763/commits/1117d19fb145cdd4414841026f90d2a388c57c52) and the [tests now seem to run](https://github.com/supercollider/supercollider/actions/runs/16327638827/job/46122346418#step:6:240).
- [x] All tests are passing

### DIFF
--- a/common/SC_PaUtils.cpp
+++ b/common/SC_PaUtils.cpp
@@ -91,7 +91,8 @@ PaDeviceIndex GetPaDefaultDevice(bool isInput) {
     int hostApiCount = Pa_GetHostApiCount();
     for (int i = 0; i < hostApiCount; ++i) {
         const PaHostApiInfo* apiInfo = Pa_GetHostApiInfo(i);
-        if (apiInfo && apiInfo->type == paWASAPI) {
+        // prefer WASAPI, but only if at least one input or output device is present
+        if (apiInfo && apiInfo->type == paWASAPI && apiInfo->deviceCount > 0) {
             return isInput ? apiInfo->defaultInputDevice : apiInfo->defaultOutputDevice;
             // while this may return noPaDevice, we don't fall back to Pa_GetDefault[Input|Output]Device here
             // because this could lead to selecting devices with different APIs between input and output
@@ -168,6 +169,14 @@ PaError TryGetDefaultPaDevices(PaDeviceIndex* inDevice, PaDeviceIndex* outDevice
             *outDevice = GetPaDefaultDevice(false);
             if (*inDevice != paNoDevice && *outDevice != paNoDevice)
                 fprintf(stdout, "Selecting default system input/output devices\n");
+            else if (*inDevice != paNoDevice && *outDevice == paNoDevice)
+                fprintf(stdout,
+                        "Selecting default system input device\n"
+                        "WARNING: no output device selected, output is disabled.\n");
+            else if (*outDevice != paNoDevice && *inDevice == paNoDevice)
+                fprintf(stdout,
+                        "Selecting default system output device\n"
+                        "WARNING: no input device selected, input is disabled.\n");
         }
     } else {
         // no inputs nor outputs


dyfer:
> You mentioned that there's code that tries to match the APIs (https://github.com/supercollider/supercollider/blob/develop/common/SC_PaUtils.cpp#L121-L122), but this runs _before_ we select default devices.

Ah okay, thanks for pointing this out.

> Actually, we should not do that because we might end up with two devices with different APIs. For example, take a laptop with no built-in microphone input, but a generic ASIO driver: GetPaDefaultDevice would return the ASIO driver as the input device and WASAPI speakers as the output device.

Stepping back, what should be the functional outcome in that case?

- choose ASIO as an input and output device (the only possible choice to have both input and output)?
- choose WASAPI output and no input? (silently disabling input was previously discussed as a non-desirable solution when considering behavior with macOS samplerate mismatch, but I think we _could_ consider it here, with extra posting)
- fail to boot? (we'd get this for free if we just adopt the PR in the current form, but the error would not be informative, I don't think...)

Spacechild1:
> choose ASIO as an input and output device (the only possible choice to have both input and output)?

Not desirable as a default device, because it might just be some virtual ASIO driver. That's the whole reason why switched to WASAPI as the default API.

> choose WASAPI output and disable input? (silently disabling input was previously discussed as a non-desirable solution when considering behavior with macOS samplerate mismatch, but I think we could consider it here, with extra posting)

That's what I would tend towards. We should probably also post a message if the input or output is disabled.

> If that's the case, this PR keeps the old behavior in that edge case, doesn't it?

We *could* just keep the old buggy behavior, but I would prefer not to :)

What about the solution I've proposed in my previous comment? I think we can safely skip WASAPI if there are no devices, but as soon as there is at least one input or output device, it would qualify as the default device.

dyfer:
> > If that's the case, this PR keeps the old behavior in that edge case, doesn't it?
> 
> We _could_ just keep the old buggy behavior, but I would prefer not to :)

Sure! But also I'd say a priority for me would be to fix the bug that we introduced since 3.13 (when no WASAPI devices available) than to fix a bug that nobody complained about :) But yeah, let's try to fix it.

> What about the solution I've proposed in my previous comment? I think we can safely skip WASAPI if there are no devices, but as soon as there is at least one input or output device, it would qualify as the default device.

That's good, but how would it behave when we have only output WASAPI device? We'd get noPaDevice for the input and the server would silently disable input, right?

dyfer:
I'm thinking dedicated message is better. Pushing now.

common/SC_PaUtils.cpp
```diff
@@ -91,7 +91,7 @@ PaDeviceIndex GetPaDefaultDevice(bool isInput) {
     int hostApiCount = Pa_GetHostApiCount();
     for (int i = 0; i < hostApiCount; ++i) {
         const PaHostApiInfo* apiInfo = Pa_GetHostApiInfo(i);
-        if (apiInfo && apiInfo->type == paWASAPI) {
+        if (apiInfo && apiInfo->type == paWASAPI && apiInfo->deviceCount > 0) {
```
Spacechild1, line 95: Maybe add a small comment, e.g.:
```
// prefer WASAPI, but only if at least one input or output device is present.
```

dyfer:
@Spacechild1 I tested this by:
- cherry-picking this commit into [windows testing PR](https://github.com/supercollider/supercollider/pull/6763) - the server boots in an environment with no WASAPI devices
- I've run the build on a Windows 10 machine with both WASAPI and ASIO devices - WASAPI is still chosen as default input/output
- I've tried simulating finding only a single output WASAPI device by [forcing the input to be `noPaDevice`](https://github.com/dyfer/supercollider/commit/a80cf5830a6fb93cc8ea8ae78b3b8f97c5866293) - server booted properly with only the output active; sound output worked fine

Anything else I should test or is this good to go?

I'm guessing we should test it in the wild by releasing 3.14.0-rc3, what do you think? @capital-G is that okay?

capital-G:
Yes, lets do a RC3 - there has also been a major bug which should be fixed for 3.14 -> https://github.com/supercollider/supercollider/pull/7062

dyfer:
@Spacechild1 is there anything left to do in this PR?

Spacechild1 (approved):

Spacechild1:
Sorry, I was AFK. All good from my side!

dyfer:
@capital-G should we merge this and release rc3?

capital-G:
Yes, I merged it and we can go forward with RC3 :)

PR #7052, opened 2025-07-16
Enable and fix more tests

dyfer:
## Purpose and Motivation

This PR enables more tests.

The NodeProxy and TempoClock tests seem to fail, becasue there is an indeterministic and relatively large delay when synchronizing the forked TempoClock with the main routine's clock. The fix moves the time sensitive parts into the TempoClock-controlled routine.

SCDoc test needed to have a mock HelpSource folder added in the user's Extensions so that it's detected by `SCDoc.prRescanHelpSourceDirs`. That seemed only to matter on macOS and Windows though.

## Types of changes

- Bug fix

## To-do list

- [x] Enable tests that pass locally (the ones below pass locally)
- [x] Fix tests that still fail in CI
  - [x] `TestNodeProxy_Server:test_schedAfterFade_notBeforeQuant`
  - [x] `TestSCDoc:test_helpSourceDirs_includedExtensions`
  - [x] `TestTempoClock:test_nextTimeOnGrid_okAfterMeterChange`
- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/testsuite/classlibrary/TestSCDoc.sc
+++ b/testsuite/classlibrary/TestSCDoc.sc
@@ -1,23 +1,43 @@
 TestSCDoc : UnitTest {
 
-    test_helpSourceDirs_excludedExtensions {
-        var result;
-        var extensionPaths = [thisProcess.platform.systemExtensionDir, thisProcess.platform.userExtensionDir];
-        var prevSetting = LanguageConfig.excludeDefaultPaths;
-        LanguageConfig.excludeDefaultPaths = true;
-        result = SCDoc.prRescanHelpSourceDirs.any { |p| extensionPaths.any { |ep| ("^"++ep.escapeChar($\\)).matchRegexp(p) } };
-        LanguageConfig.excludeDefaultPaths = prevSetting;
-        this.assert(result.not, "should not search for extensions' HelpSource when LanguageConfig.excludeDefaultPaths = true");
-    }
+	test_helpSourceDirs_excludedExtensions {
+		var result;
+		var extensionPaths = [thisProcess.platform.systemExtensionDir, thisProcess.platform.userExtensionDir];
+		var prevSetting = LanguageConfig.excludeDefaultPaths;
+		var tmpHelpSourcePath = Platform.userExtensionDir +/+ "HelpSource";
+		var folderWasCreated = false;
+		LanguageConfig.excludeDefaultPaths = true;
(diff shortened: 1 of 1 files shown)


telephon:
`TestNodeProxy_Server:test_schedAfterFade_notBeforeQuant` this one doesn't fail on my machine here.

dyfer:
@jamshark70 I believe you wrote the TempoClock tests, would be be able to take a look to see if my changes maintain tests's validity?

jamshark70:
> I believe you wrote the TempoClock tests, would be be able to take a look to see if my changes maintain tests's validity?

I think it's fine (though I didn't run the tests locally).

dyfer:
@capital-G would you be able to review by any chance? Thanks!

testsuite/classlibrary/TestNodeProxy_Server.sc
```diff
@@ -156,14 +156,22 @@ TestNodeProxy_Server : UnitTest {
 	}
 
 	test_schedAfterFade_notBeforeQuant {
-		var ok = true, earlierThan = 0.01;
+		var ok = true, earlierThan = 0.05, cond = Condition();
```
capital-G, line 159: Nitpick but wouldn't it be better to set the default value of `ok` to false?
IIUC we want to check the order of certain temporarily conditions - so I think the test should default to failing and it would be the tests responsibility to provide the passing value.
Otherwise the test could somehow pass if e.g. the condition has not been awaited correctly for some reason.

dyfer: The basic premise of the test was set up by the original author and I have not changed it.

What the test tests is that the callback does _not_ fire until after certain amount of time has passed; i.e. the test ensures that the callback does not fire too early. I think it's logically tricky to assume what's the failure otherwise...

I guess we could restructures the test so that we actually wait for the callback to fire, compare the time of firing, and then evaluate that against the wait time to determine whether the test passed or not (and also trigger failure if the callback doesn't fire after a certain timeout). However, I don't think it is worth the hassle to rewrite this. What do you think?

> Awesome job detective @dyfer :)

Haha thanks :)

EDIT: and yes, everything passes locally, that's how I fixed the tests in the first place :)

dyfer: Since you wrote that it's up to me whether to change this test, I'd say - let's leave it as is. Merging!

Issue #7044, opened 2025-07-15
Window position is shifted when it's not resizable


dyfer:
## Environment

* SuperCollider version: 3.13.0, 3.15.0-dev
* Operating system: macOS, Windows
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
// compare
Window("resizable", Rect(100, 100, 300, 300), true).front;
Window("not resizable", Rect(100, 100, 300, 300), false).front;
```

## Expected vs. actual behavior

Expected: Windows are shown in the same spot on the screen.
Actual: non-resizable window is shown lower by the size of the title bar.

<img width="327" height="379" alt="Image" src="https://github.com/user-attachments/assets/355d0a94-5e87-42f6-88e5-5fe0e173ece5" />

prko:
>We need to find what is the reason for that shift, likely somewhere in the Qt positioning logic.

sclang often provides multiple ways to achieve the same result. Using `bounds_` resolves this issue.
Of course, the ideal way is to find the reason for the shift when setting the value of the argument `resizable` to `false`...

dyfer:
> Using `bounds_` resolves this issue.

I'd argue that this is not the case.
We don't have a way of knowing the size of the title bar, as it may depend on system settings etc. AFAIU the only proper way to address this is to find where the offset is happening in Qt and fix it there.

JordanHendersonMusic:
Have you tested whether mac is just moving the windows so they don't collide... What happens if you put them next to each other?

JordanHendersonMusic:
Have you tested whether mac is just moving the windows so they don't collide... What happens if you put them next to each other?

PR #7059, opened 2025-07-19
Update Window.schelp

prko:
## Purpose and Motivation
Prior to resolving the issue reported in #7044, it may be beneficial to provide additional explanatory notes in `Window.schelp`.
This pull request introduces supplementary information and guidance to assist users.

## Types of changes

- Documentation
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Window.schelp
+++ b/HelpSource/Classes/Window.schelp
@@ -29,9 +29,54 @@ METHOD:: new
 	argument:: name
 		A String for the text that will be displayed in the title bar. The default is 'panel'.
 	argument:: bounds
-		A Rect specifying position and size of the window. The size does not include the border and title bar. Position is measured from the bottom-left corner of the screen (this is different than link::Classes/View#-bounds::). The default is size 400x400 at position 128x64, but in Qt the window is centered on the screen by default.
+		A link::Classes/Rect:: specifying the position and size of the window in screen coordinates.
+		The size does strong::not:: include the window's border and title bar.
+
+		Position is measured from the strong::bottom-left:: corner of the screen (This differs from link::Classes/View#-bounds::, which uses the top-left corner).
+		This coordinate convention originates from SuperCollider’s early implementation on macOS using the Cocoa GUI, which adopts bottom-left as its origin.
+
+		The default size is 400x400.
+		When using the Qt GUI backend, if no position is specified, the window is centered on the screen at launch (In older backends such as Cocoa, the default position was 128x64).
+		Thus, the approximate default position can be described as:
+
+		teletype::
+x = (screenWidth - 400) / 2
+::
+		teletype::
+y = (screenHeight - 400) / 2
+::
+
+		note::
+			Default position may vary depending on platform, screen resolution, and whether the window argument `resizable` is set to `true` or `false`.
+			::
 	argument:: resizable
-		A Boolean indicating whether this window is resizable by the user. The default is code::true::.
+		A Boolean indicating whether the window is resizable by the user.
+		The default is code::true::.
+		On strong::all platforms::, disabling resizing (code::resizable: false::) can affect the position and layout of the window due to differences in how system-level window frames and decorations are computed.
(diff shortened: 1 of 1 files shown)


prko:
The following screenshot is from the rendered help documentation of `Winodow`:
Screenshot 2025-07-19 at 12 08 05

HotwheelsSisyphus:
Looks good and very readable, but I didn't follow the discussion so I guess it's best if somebody who did looks this over as well. I only changed the backticks around `true` and `false` to code:: ::.

dyfer:
IMO I don't think we should add a comment about `resizable`. There's an issue filed for that - I'd rather fix it...
But if @HotwheelsSisyphus thinks it should be added, then we can do that.

prko:
If #7066 is merged, all the explanations regarding `resizable` and `position` should be removed,

@dyfer 
Thanks for #7066.

HelpSource/Classes/Window.schelp
```diff
@@ -29,9 +29,54 @@ METHOD:: new
 	argument:: name
 		A String for the text that will be displayed in the title bar. The default is 'panel'.
 	argument:: bounds
-		A Rect specifying position and size of the window. The size does not include the border and title bar. Position is measured from the bottom-left corner of the screen (this is different than link::Classes/View#-bounds::). The default is size 400x400 at position 128x64, but in Qt the window is centered on the screen by default.
+		A link::Classes/Rect:: specifying the position and size of the window in screen coordinates.
+		The size does strong::not:: include the window's border and title bar.
+
+		Position is measured from the strong::bottom-left:: corner of the screen (This differs from link::Classes/View#-bounds::, which uses the top-left corner).
+		This coordinate convention originates from SuperCollider’s early implementation on macOS using the Cocoa GUI, which adopts bottom-left as its origin.
```
dyfer: I think line 36 would read better as

`This coordinate convention originates from SuperCollider’s original Cocoa GUI implementation on macOS, which uses a bottom-left origin.`

HelpSource/Classes/Window.schelp
```diff
@@ -29,9 +29,54 @@ METHOD:: new
...
+		The default size is 400x400.
+		When using the Qt GUI backend, if no position is specified, the window is centered on the screen at launch (In older backends such as Cocoa, the default position was 128x64).
+		Thus, the approximate default position can be described as:
+
+		teletype::
+x = (screenWidth - 400) / 2
+::
+		teletype::
+y = (screenHeight - 400) / 2
+::
```
dyfer: I think this should be shorter for a number of reasons:
- there's no practical way of using another GUI backend currently, so I don't think other backends should be mentioned
- ~~the calculation is not really informative; what are `x` and `y` in relation to creating the Window? Since they actually refer to the _center_ of the Window, not the bounds, I think calculating them here isn't helpful and it just adds superfluous text.~~ Sorry, I now see `x` and `y` refer to the corner of the window, but still, I find this calculation unnecessary. IMO "the window is centered" is specific enough.
- "at launch" is inaccurate - "when created" would be better, but also we won't see the window until it's _shown_ with `.front`. So maybe less explicit language would be even better?
- Re: `In older backends such as Cocoa` - that information is unnecessary at this point and I'm not sure if it's accurate - I can't remember what Swing (the other GUI backend) did. Anyway, this is not relevant at this point.

My suggestion for the whole selected section (lines 39-47) is:
`If no position is specified, the window is centered on the screen by default.`.
(Again, no "approximate default position" / teletype section)

prko: thanks , done in https://github.com/supercollider/supercollider/pull/7059/commits/9bff11402a024592ef6d059e3426376013ef3554

HelpSource/Classes/Window.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/7059/commits/9bff11402a024592ef6d059e3426376013ef3554.

PR #7065, opened 2025-07-20
scide: remove explicit window reactivations post-open/save dialogs

passyur:
## Purpose and Motivation

Just noticed this while going through `main_window.cpp`—two todo comments reference a now-fixed (as of Qt 5.1, putatively) issue pertaining to focusing/window activation on macOS. 

This change essentially reverts 52e76de—tested this on Qt 6 and there is no difference w.r.t window focusing with/without the code. (Both exhibit the desired behavior—the main window refocuses after the open/save dialog closes.)

Link to the pertinent Qt bug report: https://bugreports.qt.io/browse/QTBUG-25295.

## Types of changes

- Cleanup

## To-do list

- [x] Code is tested

### DIFF
--- a/editors/sc-ide/widgets/main_window.cpp
+++ b/editors/sc-ide/widgets/main_window.cpp
@@ -995,22 +995,9 @@ bool MainWindow::save(Document* doc, bool forceChoose, bool saveInExtensionFolde
             // filepath with added suffix already exists!
         }
 
-#ifdef Q_OS_MAC
-        QWidget* last_active_window = QApplication::activeWindow();
-#endif
-
-        int result = dialog.exec();
-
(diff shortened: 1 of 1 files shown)


passyur:
Any update on this? @dyfer (sorry if you aren't the right person to ping on this...)

dyfer (approved):
Thanks for the ping @passyur 

This looks fine to me. I tested on macOS CI builds (main build using Qt 6.7 and legacy build using Qt 5.15) - both seem to work fine, the main window stays active after the open/close dialogs.

As I think this is a trivial change, I'll merge this.

dyfer:
Thanks for the PR @passyur !

PR #7082, opened 2025-07-26
Add examples folder to sclang and help home

prko:
## Purpose and Motivation
Supersedes #6981

## Types of changes

- Documentation
- New feature
- 
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Help.schelp
+++ b/HelpSource/Help.schelp
@@ -49,13 +49,22 @@ SECTION::Getting started
 
 These are useful starting points for getting help on SuperCollider:
 
-definitionlist::
-## link::Tutorials/Getting-Started/00-Getting-Started-With-SC##Getting Started tutorial series:: || Get started with SuperCollider
-## link::Guides/Glossary:: || Glossary
-## link::Guides/More-On-Getting-Help:: || Using Help Files effectively and inspecting class definition files to build more understanding
-## link::Browse.html#Tutorials#All tutorials:: || Index of all help files categorized under "Tutorials"
-## link::Guides/UserFAQ:: || Common Errors and FAQ
-## link::https://github.com/supercollider/supercollider/wiki/Tutorials##Community Learning Resources (Tutorials & Publications):: || SuperCollider Tutorials Wiki: Community-curated tutorial index and learning resources
+subsection:: Introduction & Reference Materials
+
+list::
+## teletype::sclang/scd:: example folder
+code::
+ExampleFiles.dir.openOS // Hold down the Shift key and press the Enter key to open the "sclang/scd example folder".
+::
+## link::Tutorials/Getting-Started/00-Getting-Started-With-SC##Getting Started tutorial series::
+## link::Guides/Glossary::
+## link::Guides/More-On-Getting-Help##More On Getting Help:::
+
+Using Help Files effectively and inspecting class definition files to build more understanding
+## link::Browse.html#Tutorials#All tutorials:::
+
+Index of all help files categorized under emphasis::Tutorials::
+## link::Guides/UserFAQ##Common Errors and FAQ::
 ::
 
 subsection:: Getting Sound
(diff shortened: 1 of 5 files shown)


HelpSource/Help.schelp
```diff
@@ -49,13 +49,22 @@ SECTION::Getting started
...
-## link::Guides/More-On-Getting-Help:: || Using Help Files effectively and inspecting class definition files to build more understanding
-## link::Browse.html#Tutorials#All tutorials:: || Index of all help files categorized under "Tutorials"
-## link::Guides/UserFAQ:: || Common Errors and FAQ
-## link::https://github.com/supercollider/supercollider/wiki/Tutorials##Community Learning Resources (Tutorials & Publications):: || SuperCollider Tutorials Wiki: Community-curated tutorial index and learning resources
+subsection:: Introduction & Reference Materials
+
+list::
+## teletype::sclang/scd:: example folder
+code::
+ExampleFiles.dir.openOS // Hold down the Shift key and press the Enter key to open the "sclang/scd example folder".
```
prko: >`ExampleFiles.dir.openOS`

I think `Platform.exampleDir` would be better here. 'ExampleFiles.dir' is somewhat confusing to me.

HotwheelsSisyphus (commented):
Thanks! Looks good to me, but I'd change two things:
1. order of the list items under Intro  & Reference Materials: I'd just put the examples last. 
2. In `teletype::sclang/scd:: example folder`, the formatting looks weird. I'd just not bother with teletype in this case? Also, maybe just write "SuperCollider examples folder"... Total beginners might not yet know what sclang is, and the "scd" may also be confusing (yes the files are scd, but the filetype may also not be so important) 

finally, just dreaming here: Maybe we should make a schelp file that serves as an index for all the examples, where one can load each example file in the IDE by clicking on a link, like it is currently possible to open a class file from the help when clicking on the "Source" link above "see also"? I'm not sure how complicated this would be.

HelpSource/Help.schelp
prko: Done in https://github.com/supercollider/supercollider/pull/7082/commits/091a9011fbea495dc91ad6d6c62a3afc2859c6e9

prko:
>I'd change two things:

Thanks! Done in https://github.com/supercollider/supercollider/pull/7082/commits/1a00586432f63a01c7fafbb33232e1d1dc93edc5
Screenshot 2025-07-30 at 11 24 17


---

>just dreaming here: Maybe we should make a schelp file that serves as an index for all the examples, where one can load each example file in the IDE by clicking on a link, like it is currently possible to open a class file from the help when clicking on the "Source" link above "see also"? 

To do this, 
- One should create a dedicated HTML document file. However, the example folder should be moved to `HelpSource/static`.
- Unless the example folder is moved to `HelpSource/static`, one should make a `sc` file that renders an HTML file using the path `Platform.exampleDir`.

SCClassLibrary/Common/Files/ExampleFiles.sc
```diff
@@ -10,4 +10,8 @@ ExampleFiles {
 	*child {
 		^Platform.resourceDir +/+ "sounds" +/+ "a11wlk01.wav";
 	}
+
+	*dir {
+ 		^(Platform.resourceDir +/+ "examples")
+ 	}
```
capital-G: IMO ExampleFiles should be limited static/media files and not code files? At least it feels strange to me that one member of this class is representing something different (code to execute) than the other members of this class (static media assets). I think it would be sufficient to have the code examples under `Platform.exampleDir` - but probably even better under `Path.exampleDir` as the user doesn't need to be aware that such a directory-path is platform-dependent. But this should then wait for https://github.com/supercollider/supercollider/pull/6755

prko: Thanks! First, I removed the `dir` method from the `ExampleFiles` class in https://github.com/supercollider/supercollider/pull/7082/commits/a12b422bc2e4fa19f24db4b594bd7c208098e107.

Changing `Platform.exampleDir` to `Path.exampleDir` could be done later.

PR #7089, opened 2025-07-29
Plotter: fix steplike data displays

### LINKED ISSUE #5296: plotMode=\bars does not draw the last element

mtmccrea:
## Purpose and Motivation

This pull request improves visual alignment for step-like data. 

There are two type os step-like data: centered respresentation (e.g. `\bars`) and sample-and-hold representation (e.g. `\steps` and `\levels`).

Previously, `\bars` were offset so they didn't align with their domain/x-axis index, and the last value wouldn't be shown (see #5296):
```supercollider
[5, 0, 1, 2, 3, 4].plot.plotMode_([\bars])
````
Screenshot 2025-07-29 at 14 08 21

Now fixed:
Screenshot 2025-07-29 at 14 00 25

While fixing this, various minor issues came up with step-like data, such as partially hidden values and buggy behavior with `editMode` enabled. This led to adding more plot modes, including a centered version of `levels`. Other new modes place markers (dots or points) at the x-position where the data is sampled, as is already done with `plines`, `dfilled`, etc. modes, and to disambiguated centered and hold-like displays. This is in part inspired by the idea of using plots as fader banks (where the data marker would be a "handle").

All available modes are shown here (new modes highlighted):
Screenshot 2025-07-28 at 14 55 15

Additional changes:
- Introduced `prDomainSpec` to manage the padding for plot modes that require it, leaving user-facing `domainSpec` to accurately reflect the spec supplied by the user.
- Check that user-specified mode is valid, warning if not
- Update `plotMode_` to refresh plot specifications dynamically when modes are changed. 
- Increased label width multiplier in `Grid.sc` to avoid cutting off text of tick labels.
- Updated documentation to clarify behavior for step-like plot modes and their visibility constraints. (addresses concerns in #6479)

Thanks @prko for getting this started in #6944.

## Types of changes

Fixes #5296

- Documentation
- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Plotter.schelp
+++ b/HelpSource/Classes/Plotter.schelp
@@ -115,8 +115,15 @@ table::
 ## code::\filled:: || fills the area under a waveform, and assumes a bipolar signal (filled to the zero-crossing)
 ## code::\pfilled:: || code::\filled::, marked also by code::\points::
 ## code::\dfilled:: || code::\filled::, marked also by code::\dots::
-## code::\levels:: || sample-and-hold style horizontal lines from the data point extending over the sample interval
-## code::\steps:: || "step interpolation": code::\levels:: connected by vertical lines
+## code::\levels:: || horizontal lines, sample-and-hold style
+## code::\plevels:: || code::\levels::, with a code::\points:: data markers
+## code::\dlevels:: || code::\levels::, with a code::\dots:: data markers
+## code::\levelsCentered:: || horizontal lines, centered on the data point
+## code::\plevelsCentered:: || code::\levelsCentered::, with a code::\points:: data markers
+## code::\dlevelsCentered:: || code::\levelsCentered::, with a code::\dots:: data markers
+## code::\steps:: || "step interpolation", sample-and-hold style, connected by vertical lines
+## code::\psteps:: || code::\steps::, with a code::\points:: data markers
+## code::\dsteps:: || code::\steps::, with a code::\dots:: data markers
 ## code::\bars:: || bar graph with filled bars
 ::
 
@@ -142,17 +149,12 @@ p.plotMode = \steps;
 /* Preview the available modes */
 (
 s.waitForBoot{
-    var modeList = [
-        \points, \dots, \levels, \steps,
-        \lines, \plines, \dlines,
-        \stems, \pstems, \dstems,
-        \filled, \dfilled, \pfilled, \bars,
-    ];
-    modeList.clump((modeList.size/2).ceil).do{ |modes, i|
-        var frq = 1000, w = 600, h = 800;
+    var modeList = Plotter.modes;
+    modeList.clump((modeList.size/3).ceil).do{ |modes, i|
+        var frq = 3000, w = 500, h = 800;
         p = {
-            LFSaw.ar(frq) ! modes.size
-        }.plot(2.25 * frq.reciprocal);
+            SinOsc.ar(frq) ! modes.size
+        }.plot(0.75 * frq.reciprocal);
 
         0.25.wait; // let Plotter update with data
         p.plotMode_(modes);
--- a/HelpSource/Classes/Plotter.schelp
+++ b/HelpSource/Classes/Plotter.schelp
@@ -141,7 +141,7 @@ are the same, in which case a single link::Classes/Symbol:: is returned.
 discussion::
 code::
 
-p = (0..20).scramble.plot;
+p = (0..20).scramble.plot;  p.parent.alwaysOnTop = true;
 p.plotMode = \points;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: HelpSource/Classes/Plotter.schelp [hunk 1]
```diff
@@ -173,6 +175,29 @@ p.superpose_(true);
...
 
+warning::
+Plot modes  code::\levels::, code::\steps::, and their variants display data in a sample-and-hold manner, so unless the link::#-domainSpecs:: has be set explicitly to extend beyond the last data point, the "step" of the last data point may not be visible.
+code::
+( // Can't see the last data point
+p = 4.collect({ 2.5.rrand(3.0) }).plot.plotMode_(\levels);
+p.parent.alwaysOnTop = true;
+)
+
+// Adding the "point" makes it visible
```
HotwheelsSisyphus, line 188: This is difficult to see even if you know it should be there. Unless we have a way of setting the size of the point markers, I would add something like

`(if you look closely at the right edge of the plot)`

to make this clearer.

Could also help to use non-random data, so folks know where to look. E.g.: `[2,3,4,5].plot.plotMode_(\levels)` for line 182

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -508,24 +575,29 @@ Plot {
...
-			Pen.lineTo(x[i] @ y[i]);
-			Pen.lineTo(x[i + 1] ?? { plotBounds.right } @ y[i]);
+			Pen.lineTo(Point(x[i], y[i]));
+			Pen.lineTo(Point(x[i + 1] ?? { plotBounds.right }, y[i]));
 		};
 		Pen.stroke;
 	}
 
 	bars { |x, y|
+		var gap = halfBinWidthPx * 0.1;
```
HotwheelsSisyphus, line 585: EDIT: ah nvm, of course it's the gap between bars... oh well. Need more coffee.

tried to follow along with the arithmethic in my head and got a bit confused, maybe it makes sense to eliminate the gap var altogether, since it isn't used further down?
Alternatively leave a comment to explain what it refers to.


var xOffset = halfBinWidthPx * -0.9
var barWidth = halfBinWidthPx * 1.8

HotwheelsSisyphus (approved):
Thanks, this is great! I have two small suggestions but that's it.

HelpSource/Classes/Plotter.schelp
mtmccrea: Thanks! I've revised it for clarity, and removed the randomness.

HotwheelsSisyphus:
Slightly OT, something that occurred to me today as I'm messing around with making a Histogram class (mostly because there's currently two ways to draw a histogram that do slightly different things): It would be nice to have another mode like bars but _without_ the gaps (or alternatively, a centered and filled step display). The bars do look good in a histogram display, but it turns out that when there is a larger number of bins, the gaps create sort of a moiré effect... (EDIT: actually, turning smoothing on for the bars mode does a decent job of counteracting that too!) 

I ended up cobbling this together from the other plotModes:
```
	barsNoGap { |x, y|
		var ycoord0 = plotBounds.bottom - (spec.warp.unmap(0) * plotBounds.height);

		Pen.smoothing_(false);
		Pen.moveTo(Point(x[0] - halfBinWidthPx, ycoord0));
		y.size.do { |i|
			Pen.lineTo(Point(x[i] - halfBinWidthPx, y[i]));
			Pen.lineTo(Point(x[i] + halfBinWidthPx, y[i]));
		};
		Pen.lineTo(Point(x.last + halfBinWidthPx, ycoord0));
		Pen.fill
	}
```

It works, for all I can tell... I can push it to this PR if you want, or include it with the histogram PR when I get around to that.
Also open to calling this something else, like StepsFilledCentered or wahtever.

mtmccrea:
Thanks for your suggestion @HotwheelsSisyphus 

> or include it with the histogram PR when I get around to that.

Given the low momentum on this PR, I'd prefer if for this feature to go with your histogram PR, which is a sensible addition.

> Also open to calling this something else, like StepsFilledCentered

The no-gap bars variant would actually be something closer to `levelsCenteredFilled` (steps aren't centered). It could perhaps be called something more simple like `histo`. In a perfect world a few core plot modes would have fill and marker attributes, but as it is we just keep adding to the list.

PR #7091, opened 2025-08-07
Classlib: Dictionary: Disallow 'nil' as Dictionary key

jamshark70:
## Purpose and Motivation

See https://scsynth.org/t/strange-bug-with-dictionary/12162/3

> The code below creates a Dictionary which is mysteriously missing keys 2, 4, 7, 9, 10, 11.

```
(
a = Dictionary.new;

a[nil] = 0;

12.do{|i|
	if(a[i] == nil,{
		a[i] = "yes";
	});
};
a.keysValuesDo{|key, val|
	[key,val].postln;
};
)
```

The approach here is to add a new selector `keyHash`, used by Dictionary to distinguish the purpose of requesting the hash. Nil:hash remains as is to handle cases like `[1, nil, 3].hash`. Dictionary:scanFor now uses keyHash, so that method dispatch by itself detects the faulty case.

Adding an `if(key.isNil)` branch into Dictionary:scanFor would have a small performance impact. This approach should not sacrifice any performance while still catching the error condition.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

Probably should squash merge.

### DIFF
--- a/SCClassLibrary/Common/Collections/Dictionary.sc
+++ b/SCClassLibrary/Common/Collections/Dictionary.sc
@@ -360,7 +360,7 @@ Dictionary : Set {
 	}
 	scanFor { arg argKey;
 		var maxHash = array.size div: 2;
-		var start = (argKey.hash % maxHash) * 2;
+		var start = (argKey.keyHash % maxHash) * 2;
 		var end = array.size-1;
 		var i = start;
 		forBy( start, end, 2, { arg i;
--- a/SCClassLibrary/Common/Core/Nil.sc
+++ b/SCClassLibrary/Common/Core/Nil.sc
@@ -129,4 +129,6 @@ Nil {
 	}
 
 	superclassesDo {}
+
+	keyHash { Error("'nil' is not valid as a Dictionary key").throw }
 }
--- a/SCClassLibrary/Common/Core/Object.sc
+++ b/SCClassLibrary/Common/Core/Object.sc
@@ -224,6 +224,7 @@ Object {
 	basicHash { _ObjectHash; ^this.primitiveFailed }
 	hash { _ObjectHash; ^this.primitiveFailed }
 	identityHash { _ObjectHash; ^this.primitiveFailed }
+	keyHash { _ObjectHash; ^this.primitiveFailed }
 
 	// lazy equality: same as == for objects
 	// "composed" for lazy operands (patterns, UGens)
--- a/testsuite/classlibrary/TestDictionary.sc
+++ b/testsuite/classlibrary/TestDictionary.sc
@@ -38,4 +38,18 @@ TestDictionary : UnitTest {
 		this.assertEquals(dict.trueAt(\k), true, "dictionary trueAt should respond true if its booleanValue is true.");
 	}
 
+	test_Dictionary_disallows_nil_as_key {
+		var dict = Dictionary[1 -> 2];
+		var err;
+		try {
+			dict.put(nil, 3);
+		} { |error|
+			err = error
+		};
+		this.assert(
+			err.notNil and: { err.errorString.contains("'nil' is not valid") },
+			"Dictionary should throw an error when trying to 'put' at nil as a key"
+		)
+	}
+
 }


JordanHendersonMusic:
I think the `if` would be better: adding a method to *all* objects to fix a bug with a single class is unnecessary; since this will *only* be overridden for `nil` it can safely be hard coded; it is possible one day we might have a new collection where you can use `nil` as a key, then `keyHash` doesn't make sense.

If I remember correctly, identity dictionary deletes the entry? I think we should make dictionary conform.

jamshark70:
> I think the if would be better:
> 
> adding a method to all objects to fix a bug with a single class is unnecessary;

It's a kind of SmallTalk-y way, though, to use dispatch rather than branching (and it's faster, see below).

> since this will only be overridden for nil it can safely be hard coded;

TBH I don't strongly object to the branching approach -- but we should be honest that it adds about 15% to the cost of a Dictionary access.

Replicating the `scanFor` logic into a couple of functions, one using a branching opcode to trap nil and the other using keyHash:

```
(
f = { arg array, argKey;
	var maxHash = array.size div: 2;
	var start = ((argKey ?? { Error("key must not be nil").throw })
		.hash % maxHash
	) * 2;
	var end = array.size-1;
	var i = start;
	block { |break|
		forBy( start, end, 2, { arg i;
			var key = array.at(i);
			if ( key.isNil or: { key == argKey }, { break.(i) });
		});
		end = start - 1;
		forBy( 0, start-2, 2, { arg i;
			var key = array.at(i);
			if ( key.isNil or: { key == argKey }, { break.(i) });
		});
		-2
	};
};

g = { arg array, argKey;
	var maxHash = array.size div: 2;
	var start = (argKey.keyHash % maxHash) * 2;
	var end = array.size-1;
	var i = start;
	block { |break|
		forBy( start, end, 2, { arg i;
			var key = array.at(i);
			if ( key.isNil or: { key == argKey }, { break.(i) });
		});
		end = start - 1;
		forBy( 0, start-2, 2, { arg i;
			var key = array.at(i);
			if ( key.isNil or: { key == argKey }, { break.(i) });
		});
		-2
	};
};

h = { |func, tries = 20|
	var sum = 0;
	tries.do {
		var now = Main.elapsedTime;
		func.value;
		sum = sum + Main.elapsedTime - now;
	};
	sum / tries
};
)

d = Dictionary.new;
100000.do { |i| d.put(i, 1000000.rand) };
a = d.array;

// average of 20 trials, to reduce margin of error
h.value { 100000.do { f.(a, 100000.rand) } };
-> 0.10155424975001
-> 0.10226796790002
-> 0.10032913885

h.value { 100000.do { g.(a, 100000.rand) } };
-> 0.087886993399997
-> 0.087901743749995
-> 0.087841677800003
```

Taking the ratio of the middle values, `0.10155424975001 / 0.087886993399997 = 1.1555094311602` or 15-16%.

The check could be moved into `put` rather than `scanFor` so that at least read-only access wouldn't lose performance. But I expect you would still see the ~15% hit when populating a Dictionary.

Now, as I said, I won't object too strongly if there's consensus in favor of the branching way. My opinion is that performance is more important for core collection classes -- it's not good to have to write in the changelog, "We've slowed down Dictionary population by upward of 10%." My registering this opinion doesn't mean that I absolutely 100% insist, *but* I do think we should be cautious about slowing down core collections by more than a couple percentage points.

> it is possible one day we might have a new collection where you can use nil as a key, then keyHash doesn't make sense.

OTOH the new collection can simply not use keyHash.

> If I remember correctly, identity dictionary deletes the entry? 

It does if the *value* is nil. IdentityDictionary errors out if you try to put at nil.

```
d = IdentityDictionary.new;
d.put(nil, 10);
ERROR: Primitive '_IdentDict_Put' failed.
Wrong type.
```

jamshark70:
Actually, though: by modifying `scanFor` directly, rather than benchmarking functions, I found that `.put`'s performance difference is negligible. `.at`'s performance does suffer a bit more, but as noted, that can be avoided by checking nil keys only when `.put`-ting. Checking only `scanFor` removes from the benchmark the other work that `put` does, so that 15% statistic doesn't reflect user-facing impact.

So perhaps this PR can go away and I can do another one with the branching approach.

JordanHendersonMusic:
> Actually, though: by modifying scanFor directly, rather than benchmarking functions, I found that .put's performance difference is negligible. .at's performance does suffer a bit more, but as noted, that can be avoided by checking nil keys only when .put-ting. Checking only scanFor removes from the benchmark the other work that put does, so that 15% statistic doesn't reflect user-facing impact.
So perhaps this PR can go away and I can do another one with the branching approach.

This all sounds good!

PR #7097, opened 2025-08-16
Add Link to Dev Documentation on Help Home Page

prko:
## Purpose and Motivation
There have been multiple requests for access to the most up-to-date help documentation. For example:
- https://scsynth.org/t/development-meetings-scheduling-polls/234/122?u=prko
- https://scsynth.org/t/development-meetings-scheduling-polls/234/124?u=prko

The linked posts above highlight the need for clearer access to documentation built from the development branch.

This PR aims to address those requests by adding a link to the

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Help.schelp
+++ b/HelpSource/Help.schelp
@@ -4,6 +4,18 @@ categories::Help
 
 NOTE:: link::Guides/News-3_14##News in SuperCollider version 3.14:: ::
 
+note::
+The most up-to-date online help documentation for SuperCollider is built weekly from the development branch and is available at:
+
+link::https://dev.docs.supercollider.online/::.
+This version reflects the latest changes and contributions from active development.
+
+For stable release documentation, refer to the main site:
+
+link::https://docs.supercollider.online/::.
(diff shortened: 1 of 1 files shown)


capital-G:
IMO we shouldn't give people the idea that the dev docs are "better" than the bundled docs - the bundled docs are the best docs as the dev docs can be diverge regarding available features which then can lead to some confusion. So in order to avoid this confusion we should only communicate the availability of the dev docs to advanced users who are aware of the diverging feature - therefore I don't think we should advertise the dev docs at the top of the documentation.

Yet I think the online dev docs are great to test/obtain feedback on e.g. new guides by sharing them via their URL using the forum etc.

I created a PR which adds a link to the dev docs in the projects README -> https://github.com/supercollider/supercollider/pull/7103 which I think would be sufficient to make advanced users aware of such a deployment.

prko:
OK, I close this! Thanks!

PR #7101, opened 2025-08-16
GHA macOS: use homebrew qt for share libscsynth build

dyfer:
## Purpose and Motivation

The libscsynth build was taking a longer time due to using qt from aqtinstall. As this build only tests scsynth build configuration, it doesn't matter much which OS and Qt source is used. 

I _think_ that not using aqt install here will also lower our GHA cache usage (1.6 GB per build).

Alternatively, we could disable it altogether, but I don't fully understand myself whether it's useful to have it or not.

## Types of changes

- Maintenance?

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -58,14 +58,13 @@ jobs:
             extra-cmake-flags: "-D SYSTEM_BOOST=ON -D SYSTEM_YAMLCPP=ON"
             ctest: true
 
-          - job-name: "x64 shared libscsynth"
-            os-version: "13"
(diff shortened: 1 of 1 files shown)


capital-G (approved):
lgtm and seems right to start phasing out x64 in favor of arm64 for our macOS builds.

PS: I also don't know the motivation behind shared libscsynth :D Would be great to have some insights here, does @Spacechild1 have an idea what this is for, in order to spread some knowledge?

edit: Feel free to merge by yourself @dyfer

PR #4583, opened 2019-09-18
Use fftw in supernova on macos

dyfer:
## Purpose and Motivation

In the process of trying to find solutions to #4289, it was suggested to use the `FFTW` library on macOS for supernova.

This PR adds build instructions for including FFTW in supercollider built on macOS (for supernova only; scsynth still uses vDSP). 

~~In order to provide static FFTW I am using a custom homebrew formula. This formula is temporarily hosted on my [github](https://github.com/dyfer/homebrew-fftw-sc), but if we decide to accept this PR, it should be moved to a more permanent place (and appropriate portions of this PR should be updated).~~

EDIT: This uses a regular FFTW from homebrew. Thanks to @brianlheim for figuring out how to fix linking with it.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] Updated documentation
- [x] This PR is ready for review
- [x] commit history is cleaned up

Issue #4289, opened 2019-02-05
FFT processing in parallel group on supernova causes loud distortion (macOS)


dyfer:
## Environment

* SuperCollider version: 3.10.3
* Operating system: macOS 10.13.6
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce
This was previous reported as issue with PartConv, but actually seems to apply to FFT processing in general on my machine.

```supercollider
(
//CAREFUL! this will create LOUD DISTORTION
~numSynths = 2;

~syncBetweenSynths = true; //works fine
// ~syncBetweenSynths = false; //THIS WILL CREATE LOUD DISTORTION

Server.supernova;

s.waitForBoot {

	SynthDef(\pt_test, {arg playbuf = 0, irbuf = 0, amp = 0.5, out = 0;
		var sig, chain;
		sig = SinOsc.ar(300, 0, amp);
		chain = FFT(LocalBuf(2048), sig);
		sig = IFFT(chain); // inverse FFT
		Out.ar(out, sig * amp);
	}).add;

	~target = ParGroup();
	s.sync;

	~allSynths = ~numSynths.collect({|inc|
		var synth = Synth(\pt_test, [\amp, -26.dbamp, \out, inc], ~target);
		if(~syncBetweenSynths, {s.sync});
		synth;
	});
}
)

//free everything
(
~target.free;
currentEnvironment.clear;
)
```

## Expected vs. actual behavior

Actual behavior:
When multiple ~~PartConvs~~ synths with FFT are started in a parallel group at the same time, they output loud distortion. If they are started after waiting for the server to allocate resources for the previous Synth, or if it's a regular group, they work fine.

Expected behavior: 
Multiple ~~PartConvs~~ synths with FFT  start simultaneously in a ParGroup without issue.

## Discussion:
It seems to be an issue with the vDSP implementation, as indicated in the [source](https://github.com/supercollider/supercollider/blob/develop/common/SC_fftlib.cpp#L24). Building supernova with the internal `Green FFT` library fixes the problem (although that's arguably not a proper "fix").

patrickdupuis:
I wasn't immediately able to reproduce (Ubuntu, 3.10.2-rc1). I'll try this again later today.

dyfer:
Few more observations:

- it does behave the same for me on macOS with 3.10.0ish, 3.10.1, 3.10.2-rc1
- additionally, it occurs also when more than one PartConv is started within a parallel group, _regardless of timing_, if
[...]

PR #7107, opened 2025-08-18
vDSP: support multiple threads

dyfer:
## Purpose and Motivation

The vDSP (FFT library on macOS) implementation originally was not multithreaded due to optimization strategy using a global buffer for holding the intermediate representation of FFT data. Once that was discovered, we [switched to using FFTW on macOS in supernova](https://github.com/supercollider/supercollider/pull/4583).

See https://github.com/supercollider/supercollider/issues/4289 for code example triggering the original error.

I wanted to switch back to using vDSP as it brings supernova's requirements closer to scsynth on macOS.

### Implementation details:

~~When using fftlib with supernova, in this PR we are allocating as many temporary buffers as there are CPU cores (hardware concurrency). Supernova [caps max number of threads to hardware concurrency](https://github.com/supercollider/supercollider/blob/develop/server/supernova/dsp_thread_queue/dsp_thread.hpp#L179), so we should never run out IIUC, but in case something changed, more buffers will be allocated if a buffer is requested beyond the preallocated pool. That will happen in the audio thread, IIUC, but that's still better than a server crash, right?~~

~~Please review carefully. While I came up myself with the logic that allocates multiple buffers and assigns them per-thread, I was using help from LLMs when implementing it.~~

## Types of changes

- Documentation
- New feature (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/common/SC_fftlib.cpp
+++ b/common/SC_fftlib.cpp
@@ -23,6 +23,9 @@ NOTE:
 vDSP uses a "SplitBuf" as an intermediate representation of the data.
 For speed we keep this global, although this makes the code non-thread-safe.
 (This is not new to this refactoring. Just worth noting.)
+NEW ADDITION:
+Additional logic was used when adapting this for Supernova,
+ensuring that each thread has its own SplitBuf.
 */
 
 #include "clz.h"
@@ -98,7 +101,11 @@ static float* fftWindow[2][SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1];
 
 #if SC_FFT_VDSP
 static FFTSetup fftSetup[SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1]; // vDSP setups, one per FFT size
+#    ifdef SUPERNOVA
+thread_local static COMPLEX_SPLIT splitBuf;
+#    else
 static COMPLEX_SPLIT splitBuf; // Temp buf for holding rearranged data
+#    endif
 #endif
 
 #if SC_FFT_GREEN
@@ -167,6 +174,23 @@ static inline float* scfft_create_fftwindow(int wintype, int log2n) {
 
 static void scfft_ensurewindow(unsigned short log2_fullsize, unsigned short log2_winsize, short wintype);
 
+#if SC_FFT_VDSP
+// initialize splitbuf memory
+void scfft_init_splitbuf() {
+    // vDSP prepares its memory-aligned buffer for rearranging input data.
+    // Note max size here - meaning max input buffer size is these two sizes added together.
+    // vec_malloc used in API docs, but apparently that's deprecated and malloc is sufficient for aligned memory on OSX.
+    splitBuf.realp = (float*)malloc(SC_FFT_MAXSIZE * sizeof(float) / 2);
+    splitBuf.imagp = (float*)malloc(SC_FFT_MAXSIZE * sizeof(float) / 2);
+}
+#endif
+
+void scfft_thread_init() {
+#if SC_FFT_VDSP
+    scfft_init_splitbuf();
+#endif
+}
+
 static bool scfft_global_initialization(void) {
     for (int wintype = 0; wintype < 2; ++wintype) {
         for (int i = 0; i < SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1; ++i) {
@@ -192,11 +216,12 @@ static bool scfft_global_initialization(void) {
             printf("FFT ERROR: Mac vDSP library could not allocate FFT setup for size %i\n", 1 << i);
         }
     }
-    // vDSP prepares its memory-aligned buffer for rearranging input data.
-    // Note max size here - meaning max input buffer size is these two sizes added together.
-    // vec_malloc used in API docs, but apparently that's deprecated and malloc is sufficient for aligned memory on OSX.
-    splitBuf.realp = (float*)malloc(SC_FFT_MAXSIZE * sizeof(float) / 2);
-    splitBuf.imagp = (float*)malloc(SC_FFT_MAXSIZE * sizeof(float) / 2);
+#    ifndef SUPERNOVA
+    // allocate splitBuf memory
+    // on supernova we do that by calling scfft_thread_init() from
+    // realtime_engine_functor::init_thread() and thread_init_functor::operator()
+    scfft_init_splitbuf();
+#    endif
     // printf("SC FFT global init: vDSP initialised.\n");
 #elif SC_FFT_FFTW
     size_t maxSize = 1 << SC_FFT_LOG2_MAXSIZE;
--- a/include/common/SC_fftlib.h
+++ b/include/common/SC_fftlib.h
@@ -64,3 +64,6 @@ void scfft_doifft(scfft* f);
 
 // destroy any resources held internally.
 void scfft_destroy(scfft* f, SCFFT_Allocator& alloc);
+
+// initialize thread local buffers
+void scfft_thread_init();
--- a/server/supernova/server/server.cpp
+++ b/server/supernova/server/server.cpp
@@ -321,6 +321,9 @@ void thread_init_functor::operator()(int thread_index) {
         if (!result)
             std::cout << "Warning: cannot set thread affinity of audio helper thread" << std::endl;
     }
+
+    // initialize thread local buffers
+    scfft_thread_init();
 }
 
 void io_thread_init_functor::operator()() const {
@@ -366,6 +369,9 @@ void realtime_engine_functor::init_thread(void) {
     }
 
     name_current_thread(0);
+
+    // initialize thread local buffers
+    scfft_thread_init();
 }
 
 void realtime_engine_functor::log_(const char* str) {
--- a/README_MACOS.md
+++ b/README_MACOS.md
@@ -46,8 +46,8 @@ Prerequisites:
 - **git, cmake >= 3.12, libsndfile, readline, and qt6 >= 6.2**, installed via homebrew:
   `brew install git cmake libsndfile readline qt@6`
 
-- If you want to build with the *supernova* server, you need **portaudio** and **fftw** packages, which can also be installed via homebrew:
-  `brew install portaudio fftw`
+- If you want to build with the *supernova* server, you need **portaudio** package, which can also be installed via homebrew:
+  `brew install portaudio`
 
 Obtaining the source code
 -------------------------
--- a/editors/sc-ide/CMakeLists.txt
+++ b/editors/sc-ide/CMakeLists.txt
@@ -518,7 +518,6 @@ if(APPLE OR WIN32)
             endif()
         endif()
         if(SUPERNOVA)
-            get_target_property(LINK_FFTW libsupernova SUPERNOVA_FFTW)
             string(APPEND DEPLOY_CMD "
                 -executable=${CONTENTS_DIR}/Resources/supernova
                 -executable=${CONTENTS_DIR}/Resources/plugins/DiskIO_UGens_supernova.scx
--- a/server/scsynth/CMakeLists.txt
+++ b/server/scsynth/CMakeLists.txt
@@ -1,7 +1,8 @@
 # here we choose who provides us with the FFT lib
 # this is done independently for scsynth and supernova
-# on macOS scsynth uses vDSP implementation for FFT
-# which is not thread-safe
+# for historical reasons, as the vDSP implementation
+# wasn't thread-safe in the past
+# and couldn't be used in supernova
 
 if (APPLE)
     add_definitions("-DSC_FFT_VDSP")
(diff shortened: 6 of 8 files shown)


dyfer:
@joshpar would you be able to take a look at this? Thanks!

common/SC_fftlib.cpp
```diff
@@ -98,7 +107,82 @@ static float* fftWindow[2][SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1];
...
+        real = static_cast<float*>(nova::malloc_aligned(size));
+        imag = static_cast<float*>(nova::malloc_aligned(size));
+    }
+
+    void resize(size_t newsize) {
+        if (newsize == size)
+            return; // no-op if size unchanged
+
+        if (real)
+            free(real);
```
Spacechild1: If you allocate with `nova::malloc_aligned`, you need to free with `nova::free_aligned`.

common/SC_fftlib.cpp
```diff
@@ -98,7 +107,82 @@ static float* fftWindow[2][SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1];
...
 static FFTSetup fftSetup[SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1]; // vDSP setups, one per FFT size
-static COMPLEX_SPLIT splitBuf; // Temp buf for holding rearranged data
+// vDSP prepares its memory-aligned buffer for rearranging input data.
+// Note max size here - meaning max input buffer size is these two sizes added together.
+size_t gBufSize = SC_FFT_MAXSIZE * sizeof(float) / 2;
+
+struct BufferPair {
+    float* real = nullptr;
+    float* imag = nullptr;
+    size_t size = 0;
```
Spacechild1: Why are these members public?

common/SC_fftlib.cpp
```diff
@@ -98,7 +107,82 @@ static float* fftWindow[2][SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1];
...
+        assert(real && imag);
+        return { real, imag };
+    }
+
+    ~BufferPair() {
+        if (real)
+            free(real);
+        if (imag)
+            free(imag);
+    }
```
Spacechild1: When you are writing your own destructor, there's a high chance that you also need to write copy constructor/assignment and move constructor/assignment (= "The Rule of Five"). Otherwise you get the compiler generated default versions, which may do the wrong thing. This is definitely the case here.

common/SC_fftlib.cpp
```diff
@@ -98,7 +107,82 @@ static float* fftWindow[2][SC_FFT_LOG2_ABSOLUTE_MAXSIZE_PLUS1];
...
+#    else
+static std::unique_ptr<BufferPair> gBufPair; // Temp buf for holding rearranged data
+#    endif
+
+static inline BufferPair* getBufferPair() {
+#    ifdef SUPERNOVA
+    if (bufPairIndex < 0)
+        bufPairIndex = globalBufAssignmentCounter.fetch_add(1, std::memory_order_relaxed);
+    {
+        std::lock_guard<std::mutex> lock(globalBufPairsMutex);
```
Spacechild1: No mutex locking on the audio thread!

Spacechild1 (changes requested):

common/SC_fftlib.cpp
dyfer: Got it, thanks.
So how do I guard against buffer access? Or is lock not needed since I have individual buffers per thread?

Spacechild1: I think we need a different solution alltogether. Hold on.

common/SC_fftlib.cpp
```diff
@@ -192,11 +276,20 @@ static bool scfft_global_initialization(void) {
...
-    // vec_malloc used in API docs, but apparently that's deprecated and malloc is sufficient for aligned memory on OSX.
-    splitBuf.realp = (float*)malloc(SC_FFT_MAXSIZE * sizeof(float) / 2);
-    splitBuf.imagp = (float*)malloc(SC_FFT_MAXSIZE * sizeof(float) / 2);
+#    ifdef SUPERNOVA
+    {
+        std::lock_guard<std::mutex> lock(globalBufPairsMutex);
+        // Preallocate buffers for a number of threads based on hardware concurrency,
+        // as we don't know how many supernova is actually using
+        // if we run out, more will be allocated on demand
+        size_t kPreallocatedBufCount = std::thread::hardware_concurrency();
```
Spacechild1: `k<Name>` is typically used for compile time constants, not for local variables.

Spacechild1:
The mutex-protected vector of buffers is not ideal... Luckily there is a much simpler solution:

1. Make `static COMPLEX_SPLIT splitBuf` thread-local, but only for `SUPERNOVA`
2. Make a function `scfft_init_thread` that allocates the buffers in the thread-local `splitBuf` instance
3. Call `scfft_init_thread()` in the following places:
    -  `realtime_engine_functor::init_thread()` for the main audio callback resp. the NRT main thread (via `non_rt_functor::init_thread()`)
    - `thread_init_functor::operator()` for the DSP helper threads
4. That should be it.

dyfer:
Wow, thanks so much @Spacechild1 !

> Call `scfft_init_thread()` in the following places:
> * `realtime_engine_functor::init_thread()` for the main audio callback resp. the NRT main thread (via `non_rt_functor::init_thread()`)
> * `thread_init_functor::operator()` for the DSP helper threads

I thought that the fft functions are only available through the plugin interface... Should I `#include SC_fftlib.cpp` in `server.cpp` to be able to call that new function, or is there some other way?

Spacechild1:
> I thought that the fft functions are only available through the plugin interface... 

`scfft_create`, `scfft_destroy`, `scfft_dofft` and `scfft_doifft` are exported via the interface table, but the implementation is naturally on the server side. Nothing is stopping you from adding an ordinary function that is only called by scsynth or supernova.

> Should I #include SC_fftlib.cpp in server.cpp to be able to call that new function, or is there some other way?

`SC_fftlib.cpp` is already compiled into scsynth and supernova.

Spacechild1:
Side note: I have no idea why `scfft_global_initialization` is called at static initialization time and not explicitly like other global initialization functions such as `InterfaceTable_Init()` or `initialize_library()`... I would find it much cleaner to put it into `SC_fftlib.h` and then call it explicitly. But that's not the topic of this PR :)

dyfer:
Just to clarify:

Aside from originally allocating the buffers, there's also `scfft_create` which in turn checks for the buffer size and re-allocates memory if needed (`scfft_ensurewindow`). IIUC that's called outside of the audio thread, so it wouldn't have access to the correct instance of the thread local `splitBuf`. Is that right? If so, how to address this?

Spacechild1:
>  IIUC that's called outside of the audio thread

`scfft_create` resp. `scfft_ensurewindow` are called on the audio thread.

dyfer:
> `scfft_create` resp. `scfft_ensurewindow` are called on the audio thread.

Ah got it, thanks!

include/common/SC_fftlib.h
```diff
@@ -64,3 +64,6 @@ void scfft_doifft(scfft* f);
 
 // destroy any resources held internally.
 void scfft_destroy(scfft* f, SCFFT_Allocator& alloc);
+
+// prepare splitBuf per thread
```
Spacechild1: This is implementation specific. Maybe say something like "initialize thread local buffers" or just "thread-local initialization"

dyfer: Okay. But how would you feel about changing this function  to `scfft_init_splitbuf()` and calling it from the `scfft_global_initialization()` on scsynth and from the thread init on supernova? I'll push a separate commit.

Spacechild1: I like `scfft_init_splitbuf` as an internal helper function, but I think it's too implementation specific to call it outside the initialization routines.

I would keep `scfft_thread_init` or `scfft_init_thread` and do the following:
```
void scfft_thread_init()
{
#if SC_FFT_VDSP
    scfft_init_splitbuf();
#endif
}
```

dyfer: What do you think about https://github.com/supercollider/supercollider/pull/7107/commits/5f510a18bb0b8860d9df0a164feaf0a9f228152b ?

include/common/SC_fftlib.h
```diff
@@ -65,5 +65,5 @@ void scfft_doifft(scfft* f);
 // destroy any resources held internally.
 void scfft_destroy(scfft* f, SCFFT_Allocator& alloc);
 
-// initialize the splitBuf memory for vDSP implementation
-void scfft_init_splitbuf();
+// initialize thread local buffers for vDSP implementation
```
Spacechild1: The callers don't need to know anything about vDSP. It's enough to know that some sort of per-thread initilization is happening.

server/supernova/server/server.cpp
```diff
@@ -322,8 +322,8 @@ void thread_init_functor::operator()(int thread_index) {
             std::cout << "Warning: cannot set thread affinity of audio helper thread" << std::endl;
     }
 
-    // initialize the FFT library
-    scfft_init_splitbuf();
+    // initialize thread local buffers for vDSP implementation
```
Spacechild1: Again, this comment is too specific. I don't think we need a comment here at all. `scfft_thread_init` is pretty self-explanatory.

common/SC_fftlib.cpp
```diff
@@ -176,12 +176,16 @@ static void scfft_ensurewindow(unsigned short log2_fullsize, unsigned short log2
 
 // initialize splitbuf memory
 void scfft_init_splitbuf() {
```
Spacechild1, line 179: I think the whole function needs to wrapped in `#if SC_FFT_VDSP`.

include/common/SC_fftlib.h
Spacechild1: Looks good to me! Just some small notes on the comments.

common/SC_fftlib.cpp
dyfer: ~~Calls to that function (in `scfft_global_initialization` and `scfft_thread_init`) are wrapped in `SC_FFT_VDSP`. Should I really also wrap the function itself?~~ Ah I see, this fails on other platforms.

include/common/SC_fftlib.h
dyfer: I changed the comment. Is it better or should I change it or remove it altogether?

Spacechild1:
This looks good now! I don't have time to test myself, but I don't see any issues. Just make sure to test it both with RT and NRT synthesis.

dyfer:
Thanks! Turns out NRT synthesis was segfaulting... 
I tried to fix it somewhat blindly in https://github.com/supercollider/supercollider/pull/7107/commits/310cd9f9ce7168eb3dd8a24e722aca8979bb2f76. Is that okay? It does seem to work.

server/supernova/server/server.cpp
```diff
@@ -203,6 +203,7 @@ void nova_server::group_free_deep(abstract_group* group) {
 void nova_server::run_nonrt_synthesis(server_arguments const& args) {
 #ifndef NO_LIBSNDFILE
     start_dsp_threads();
+    scfft_thread_init(); // initialize thread local buffers
```
Spacechild1: This fixes the symptom, but it does not solve the underlying issue that the `non_rt_functor::init_thread` is not executed (which is a bug).

I would mimick the behavior of `portaudio_backend::perform` in `sndfile_backend::audio_fn`:
```
public:
    void audio_fn_noinput(size_t frames_per_tick) {
        if (unlikely(!engine_initialized)) {
        {
            engine_functor::init_thread();
            engine_initialized = true;
        }        
        engine_functor::run_tick();
        write_output_buffers(frames_per_tick);
    }

    void audio_fn(size_t frames_per_tick) {
        super::clear_outputs(frames_per_tick);
        read_input_buffers(frames_per_tick);
        audio_fn_noinput(frames_per_tick);
    }

private:
    SndfileHandle input_file, output_file;
    std::size_t read_position;
    int block_size_;
    bool engine_initialized = false; // <- new member
```

dyfer: Great, thank you so much! I wouldn't be able to arrive at this conclusion myself...

NRT synthesis seems to work now.
However, reading from a file in NRT does not work, although it doesn't work in 3.14/develop either, so I think that can be left for another PR, unless you feel differently. I filed the issue as #7108.

Spacechild1: > so I think that can be left for another PR, unless you feel differently

I think so too!

Spacechild1 (approved):

Spacechild1:
The code has been successfully tested by @dyfer. Unless someone has objections, I will go ahead and merge this in the next few days.

PR #7109, opened 2025-08-21
SCLang: add caching to `GetBackTrace`

JordanHendersonMusic:
Thanks to @jamshark70 for reporting and helping with this.

Caches all frames in the back trace. If a duplicate frame is found, just copy in the previous `DebugFrame`, don't make a new one.

Take the following example.
```
~good = { |level|
	if(level > 0) {	~good.(level - 1) } { Error("meow").throw	};
};

try { ~good.(20) }
```
This would perform fine.

However, adding a layer of function indirection would show super-linear memory usage and easily crash my machine.
```
~bad = { |level|
	{
		if(level > 0) {	~bad.(level - 1) } { Error("meow").throw	};
	}.();
};

try { ~bad.(20) }
```

Now, the two example both scale linearly as `level` increases, the bad example scales faster, but it is still linear.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -1774,10 +1774,15 @@ int prDumpBackTrace(struct VMGlobals* g, int numArgsPushed) {
 
 /* the DebugFrameConstructor uses a work queue in order to avoid recursions, which could lead to stack overflows */
 struct DebugFrameConstructor {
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Disclaimer: I am not a ~~lawyer~~ sclang expert, but this looks sensible to me - I appreciate the use of more modern constructs and comments :)

Feel free to merge or request additional reviews.

JordanHendersonMusic:
Thanks for the review!

PR #4516, opened 2019-07-30
Fix network socket defaults

snappizz:
## Purpose and Motivation

fix #4496 and #4497. see those issues for more detailed explanations. basic summary:

- scsynth and supernova listen to all network interfaces by default, which isn't good for security -- it means that by default, everyone on your LAN can crash your server! the defaults should only listen to your own machine
- scsynth supports the `-B` command line option, which should default to 127.0.0.1 and not 0.0.0.0
- supernova doesn't support `-B` at all, and in fact throws an error if you try to give it `-B`. (for this PR i only did enough to prevent the error, not fix the network issue in supernova)
- however `ServerOptions` doesn't expose `-B`, so i added `ServerOptions:bindAddress`

i was on the fence about whether to put this into 3.10.3 or 3.11, since it is a (minor) breaking change and a new feature. however it's also a pretty embarrassing security hole.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #7111, opened 2025-08-25
SCLang & Server: increase server defaults

### LINKED ISSUE #5558: Increase default server memory size
## Motivation

The default memory size for the server is currently 8192, about 8mb. This is extremely conservative by today's standards and I think this leads to some confusion for new(er) users once they want to start loading sound files into buffers for example, which will quickly lead to those 8 megabytes being exhausted. The user then has to figure out the issue, find the startup file / or add the setting to the project at hand and then be able to do what the user wanted to. I think the default should be more modern and then for users who want an extremely small memory footprint they can set the memory size in a startup file, instead of vice-versa as it is now.

related: https://github.com/supercollider/supercollider/issues/5557

I think the value of the memory size could reasonably be raiseed to 128mb or 256mb without causing too much trouble. Not sure what exact sweetspot here is but I guess a lot of laptops nowadays have approximately 4-8GB of ram? 

## Plan for Implementation

Change this line:
https://github.com/supercollider/supercollider/blob/e4743da23948f6ce1cfe7750273814d0f489cc20/SCClassLibrary/Common/Control/Server.sc#L74

JordanHendersonMusic:
## Purpose and Motivation

Increase `memSize`, `maxNodes`, `maxSynthDefs`, and `numWireBufs` in both SCLang and in the world options.

Closes  #5558

## Types of changes

- Bump Defaults

## To-do list

- [x] Code is tested
- [x] All tests are passing

Before merging, we should consult authors of other clients, as by changing the world options, we are altering everyone's default.


include/server/SC_WorldOptions.h
```diff
@@ -38,15 +38,15 @@ struct WorldOptions {
...
+    uint32 mMaxNodes = 8192;
+    uint32 mMaxGraphDefs = 8192;
+    uint32 mMaxWireBufs = 512;
     uint32 mNumAudioBusChannels = 1024;
     uint32 mNumInputBusChannels = 8;
     uint32 mNumOutputBusChannels = 8;
     uint32 mNumControlBusChannels = 16384;
     uint32 mBufLength = 64;
-    uint32 mRealTimeMemorySize = 8192; // in kilobytes
+    uint32 mRealTimeMemorySize = 256000; // in kilobytes
```
Spacechild1: Just beware that increasing the `WorldOptions` defaults might consume too much memory for projects that use multiple scsynth instances.

JordanHendersonMusic: We had a talk about this in the dev meeting, the argument was that if you are running multiple servers, you should be able to change this option yourself. The increase in the default is more to avoid confusing new users.

dyfer: Can we use a power of 2 value here as well? Not a technical requirement, but for consistency with other values...
`262144`
(The same would apply to the `.sc` file)

Spacechild1: > We had a talk about this in the dev meeting, the argument was that if you are running multiple servers, you should be able to change this option yourself. 

Sure, but this has to be communicated clearly because it can be a breaking change. Take a project like https://github.com/asb2m10/plugincollider. If they update SC and recompile, they might be wondering why suddenly every plugin instance uses 256 MB of RAM ;-)

JordanHendersonMusic: Do you have any advice on how to do that? I'm a bit stumped.

Spacechild1: Well, personally, I wouldn't touch the default values in `WorldOptions` in the first place. But if you really want to, just make sure to list it as a (breaking) change in the release notes.

dyfer: Hm... that argument of a project like plugincollider is somewhat convincing to me for not touching the WorldOptions (or maybe not as much). What do you think @capital-G ?

I think maybe WorldOptions values could be increased 4x, while classlib side more (the values suggested in the PR). 

If the value in ServerOptions ends up being different from scsynth defaults, then we could add a comment there noting that these values are different from WorldOptions.

jamshark70: > I think maybe WorldOptions values could be increased 4x, while classlib side more (the values suggested in the PR).

I think that's fair.

So, scsynth: `uint32 mRealTimeMemorySize = 32768;`

Then in ServerOptions:

```
	var <>memSize = 262144;
```

and

```
				memSize: 32768,
```

It's important that the `defaultValues` dictionary match the scsynth defaults, so that the larger memSize value will be written into the options string.

capital-G: > Hm... that argument of a project like plugincollider is somewhat convincing to me for not touching the WorldOptions (or maybe not as much). What do you think @capital-G ?

IMO outside projects should set those values explicitly and not rely on the default values provided by scysnth/supernova if they somehow rely on these values. We don't guarantee that those values are kept static and any kind of version upgrades should be tested.
Such changes should be communicated prominently in the changelog though.

> I think maybe WorldOptions values could be increased 4x, while classlib side more (the values suggested in the PR).

IMO sclang shouldn't provide any kind default values in the first place but take the default values from the scsynth/supernova process, which then can be overwritten upon request or inspected upon request.
Yet I think that having 2 "default" values, one when starting scsynth via sclang and another one w/o starting from sclang, starts to become really confusing and we should settle for one default value in both domains - IMO the proposed ones in the PR.

Spacechild1: > IMO sclang shouldn't provide any kind default values in the first place but take the default values from the scsynth/supernova process, which then can be overwritten upon request or inspected upon request.

I tend to agree. At the same time, it's nice that `ServerOptions` is populated with meaningful values that the user can inspect, instead of everything being `nil`.

However, we could add a primitive for `ServerOptions.new` that simply populates the instance variables from a `WorldOptions` instance.  This way we automatically take the default options from `WorldOptions` *and* the user can see the actual values.

Spacechild1: Anyway, if we use a common default value, 256 MB seems a bit much too me. What are people doing that they need this amount of RT memory? Do you have any numbers of typical RT memory requirements?

(I guess 64 MB could be a reasonable compromise.)

JordanHendersonMusic: > IMO sclang shouldn't provide any kind default values in the first place but take the default values from the scsynth/supernova process, which then can be overwritten upon request or inspected upon request.

I disagree, scsynth is just an audio server, it doesn't know anything about the use case. 

SCLang is a use case, here we can assume there is likely only going to be one audio server, and the user will be running this as the main application, or at least one of the main applications (perhaps something else doing visuals?). Additionally, SCLang is design to be friendly to new users. Using 256mb of memory is reasonable in this case.

Something like PluginCollider is a different use case and should come with there own defaults that better suits.

With this like of reasoning, I don't think that scsynth should have defaults at all, and the client (use case) should provide them, but obviously changing that isn't reasonable. 

I'd be in favour of deleting the changes to the server options and leaving the sclang changes as they are.

Spacechild1: > I disagree, scsynth is just an audio server, it doesn't know anything about the use case.

That's a good point and this has been my initial concern. 256 MB is just too large as a default value. I agree that we should confine this change to the sclang side.

JordanHendersonMusic: Okay I'll remove the server change. This means the two will be different, this seems fine because scsynth's use case is to be a general purpose audio server, sclang's use case is less general.
 
> 256 MB is just too large as a default value.

Too large of a value of sclang's default?

Spacechild1: > Too large of a value of sclang's default?

At least for scsynth. For sclang it's probably fine, since most users only use a single Server instance, but I would still be curious to know how one can realistically use that amount of RT memory.

dyfer: > I would still be curious to know how one can realistically use that amount of RT memory

I think the most common situation is using longer delay lines.
Assuming 1 minute of single channel audio = ~10 MB: take a synth with delay of max 10s, operating on a spatial I signal of 16 channels. That's 160s -> ~26MB per synth. Ten of those, e.g. triggered live, and we hit 256 MB. Not the most common scenario, but absolutely viable.

Spacechild1: > I think the most common situation is using longer delay lines.

Note that this only applies to `DelayN`, `DelayL` and `DelayC`. All other delay objects operate on buffers and therefore don't use RT memory for the actual delay lines. Using the `Delay*` objects for long delay lines is certainly a mistake. (I'm surprised this isn't actually mentioned in the help files! I just filed an issue for this: https://github.com/supercollider/supercollider/issues/7113)

dyfer: > Using the `Delay*` objects for long delay lines is certainly a mistake.

I'd say that's subjective! (even if I partially agree) But my example uses multiple medium-length (subjective...) buffers in a scenario that's totally feasible. When running many synths with delay lines it's a chore to prepare buffers for all of them.

jamshark70: > Note that this only applies to DelayN, DelayL and DelayC. 

Allpass[N/L/C] as well, right?

Spacechild1: Yes! I forgot about those.

telephon: You may have arbitarily many parallel independent synths that need a delay line or some memory, you certainly don't want to go though all the management of buffers for that. There are many cases where you threfore want a `LocalBuf` or a `DelayN` instead. So it isn't a mistake, but something to be aware of, yes.

JordanHendersonMusic:
I've reverted the world options change, just the sc side of things that is changed now.

Spacechild1 (approved):

jamshark70 (dismissed):
Sorry, I know this was "almost done," but...

If there is a difference between the ServerOptions `defaultsValues` and the WorldOptions values, then it's mandatory for that parameter to be included in the `.asOptionsString` result.

But `asOptionsString` uses a discrepancy between a member variable and its corresponding default to know whether to include that parameter or not:

```
		if (memSize !== defaultValues[\memSize], {
			memSize = memSize.asInteger;
			o = o ++ " -m " ++ memSize;
		});
```

The intent here is for sclang always to override scsynth's default, meaning that the current "check !==" approach will not work.

One approach might be to use an association in the collection: `memSize: 8192 -> 256000`, and initialize variables to `theDefault.value`, and then have a method to: 1/ if the default is a simple value, check `!==`; 2/ if the default is an association, check `variable !== defaultValues[key].key`.

JordanHendersonMusic:
@jamshark70 thanks! The whole default values can be removed as it is unnecessary now.

JordanHendersonMusic:
Okay, so this needs another review.

I've rewritten how serveroptions is initlaised and belaoptions. Before it was doing a completely dangerous thing with `instVarPut`, now it uses the new keyword arg feature.

I think the ServerOptions should know at all what the WorldOptions default are, it should always send its values, there is no reason why it wouldn't, and this decreases coupling.

JordanHendersonMusic:
Does anyone know what the purpose of this test is? 
	`TestServerOptions:test_bindAddress_defaultValue (0.0s)`

dyfer:
> Does anyone know what the purpose of this test is?
> `TestServerOptions:test_bindAddress_defaultValue (0.0s)`

I believe this is to ensure that the bind address setting is actually populated to the scsynth command arguments.

Background: we used to not bind to local address by default and scsynth would accept connections from anywhere, including from the network. This was changed as a potential security risk and I believe the tests just ensures that what we set in the class library actually makes it to the server command line.

JordanHendersonMusic:
So it was added in this PR https://github.com/supercollider/supercollider/pull/4516. I think the test arises because it was trying to manage the shared state of the WorldOption's default values, by sending all the defaults every time, this test becomes unnecessary.

I will delete this test and write a new one that checks the options are sent every time.

JordanHendersonMusic:
I thought this would be an easy change....



Okay so I've reworked how these classes are initialised.

I've made all the values be sent where possible, it no longer checks against the defaults (I don't know why it did this in the past? Is there something I am missing?).

I've turned everything into integers where relevant when making the option string (the primitive does this and so does the command line program).

I've deleted the test that check the value wasn't sent when it equalled the default.

SCClassLibrary/Common/Control/Server.sc
```diff
@@ -1,119 +1,63 @@
...
+	var <>pingsBeforeConsideredDead = 5;
 
-	var <>reservedNumAudioBusChannels;
-	var <>reservedNumControlBusChannels;
-	var <>reservedNumBuffers;
-	var <>pingsBeforeConsideredDead;
+	var <maxLogins = 1;
 
-	var <maxLogins;
+	// TODO: None of these are used, why are they here?
```
JordanHendersonMusic: I don't know what these are for, I can't find where they are read.

SCClassLibrary/Common/Control/Server.sc
```diff
@@ -128,115 +72,65 @@ ServerOptions {
...
-			o = o ++ " -s " ++ safetyClipThreshold;
-		});
+		};
+
+		if (Server.program.asString.contains("supernova")) {
+			if (threads.notNil) { o = o ++ " -T " ++ threads.asInteger };
+			if (threadPinning.notNil) { o = o ++ " -y " ++ threadPinning.asBoolean.asInteger };
+		};
+
+		// TODO: what if the server is on a different machine with a different platform?
```
JordanHendersonMusic: This seems like a bug

SCClassLibrary/Common/Control/Server.sc
jamshark70: recHeaderFormat, recSampleFormat, recChannels, recBufSize are used in the Recorder class.

They aren't commandline options because there is no dedicated OSC command for recording, so, recording need not be configured at server boot time. recChannels and recBufSize are used when creating the recording buffer; recHeaderFormat  recSampleFormat are used in the /b_write command that prepares for DiskOut use.

SCClassLibrary/Common/Control/Server.sc
jamshark70: > > TODO: what if the server is on a different machine with a different platform?
>
> This seems like a bug

Is it even possible for sclang on machine A to `unixCmd` onto machine B?

In my 23 years of SC use, I've never heard of anyone trying to do that.

We do have a protocol to connect to a remote server, but this assumes that the remote machine has already booted the server. It's never been the remote client's responsibility to construct a command line options string for a different machine.

jamshark70 (changes requested):
My only remaining request for changes to remove the comment `// TODO: None of these are used, why are they here?`.

I suppose this hinges on the definition of a "server option." Does this term strictly refer to commandline options, or does it refer as well to configuration options that may be used elsewhere? Currently it's the latter definition, and I think this is fine. If it's strongly felt that server options must be commandline options and only those, then there would be a breaking change to move them out. (I don't think it's worth it.)

Removing the default checks does streamline the code. I agree.

SCClassLibrary/Common/Control/Server.sc
JordanHendersonMusic: Right I see! Okay I'll change the comments to reflect.

SCClassLibrary/Common/Control/Server.sc
JordanHendersonMusic: Ah I see! I've never used a remote server.

JordanHendersonMusic: I've removed the comment, but creating ServerOptions and sending them to a remote machine to boot the server doesn't seem that far fetched though.

capital-G:
@jamshark70 can you have another look at this now?

jamshark70 (approved):

testsuite/classlibrary/TestServerOptions.sc
```diff
@@ -1,25 +0,0 @@
-TestServerOptions : UnitTest {
```
capital-G, line 1: Why remove this test?

capital-G (commented):
The code looks good to me, though I personally disagree w/ un-syncing the defaults of scsynth within sclang and the defaults of spawning scsynth from cli with the motivation to make it easier for 3rd party vendors. IMO we should always have the user in mind and since the CLI options of scsynth are already pretty obscure it would be the easiest for the user to implicitly assume that cli-scsynth and sclang-scsynth are essentially the same, because otherwise people have to start digging through all the single character options of scsynth/supernova and look up what the values are within sclang-scsynth through the ServerOptions class.
There is a limited amount of 3rd party vendors which are probably technically skilled enough to set their configuration explicitly when upgrading scsynth and there are probably more users who want to start scsynth through CLI w/o digging through all the configuration defaults of the CLI.

And thanks for making the `ServerOptions` class a tad nicer. I still find the `asOptionsString` method somehow awkward, but maybe this can be less "magical" if we at some point leverage on `boost::program_potions` for scsynth :)
Thinking of this - I think this could be the proper way of increasing the defaults b/c they are communicated more clearly through the CLI interface? I wanted to work on this, but the bela support makes this a bit more complex than I'd like it to be - at least I don't know how to test that it still works correctly with bela and we don't have a person around who uses a bela?

If people think it is fine as is, so w/o upgrading the defaults within scsynth/supernova, people can merge it, but I am not the person who merges it :)

PS: Why did you remove the test though? I think it would be good to keep this around.

testsuite/classlibrary/TestServerOptions.sc
JordanHendersonMusic: The test checked arguments weren't emitted when they matched scsynth's defaults. This pr removes the knowledge of scsynth's defaults.

JordanHendersonMusic:
I think depending on the number of people using scsynth directly, that decoupling the defaults of scsynth to SCLang and the ensuing reduced complexity outway the benefits of ease of use for those users. I suspect the number of CLI users is small compared to the number of SCLang users, but I have no data to back this up.

Other clients should be encouraged to switch to this approach, decoupling themselves from scsynth's defaults. If anyone has any suggestions on how to communicate that, please let me know!

I agree that the scsynth cli is too complex, many of the single arguments variables seem arbitrary and a more complete interface would be welcome, but this would require changing all clients and configurations so I don't think it is worth it.

SCClassLibrary/Common/Control/Server.sc
dyfer: I have booted a server remotely, but not directly through ssh; rather I created an OSC communication with sclang running on the other end and transmitted server options that way. I.e. the communication was done by setting ServerOptions's values from OSC values.

But I agree, it's not that far fetched to imagine an ssh link and the ability to start the server on a different platform.

JordanHendersonMusic: Let's leave that issue for another pr as there is already a bunch to discuss here! I'll make one.

dyfer:
@capital-G are we merging this or do we need more discussion?

capital-G:
> @capital-G are we merging this or do we need more discussion?

I think further discussion wouldn't different a much different result anymore, so I think it is fine to merge as is. Though I am worried by the red macOS pipeline - I couldn't re-trigger the workflow, but we should know that the changes doesn't introduce problems for our CI.

dyfer:
Hm, good point about the CI - it did seem to fail at a similar point in multiple runs... I'll investigate locally.

SCClassLibrary/Common/Control/Server.sc
```diff
@@ -1,119 +1,63 @@
...
-	var <numAudioBusChannels;
-	var <>numControlBusChannels;
-	var <numInputBusChannels;
-	var <numOutputBusChannels;
-	var <>numBuffers;
+	var <>maxNodes = 8192;
+	var <>maxSynthDefs = 8192;
+	var <>protocol = \udp;
+	var <>blockSize = 64;
+	var <>hardwareBufferSize = 0;
```
dyfer: I think this is incorrect: it seems to effectively request the _smallest_ possible buffer size on macOS, instead of the default buffer size. I think it should be `nil` by default, so that the server chooses the default buffer size. 

_Alternatively_, we'd need to change the server behavior so that it interprets `0` as request for the default buffer size, but that may be a breaking change?

On my system (macOS) this results in a buffer size of 15 samples (???) and the test suite indeed breaks with that setting (why it breaks and why it defaults to such a buffer size may be an issue to address in another PR).

dyfer (changes requested):
@JordanHendersonMusic I think I found the issue that caused the testsuite to fail.

JordanHendersonMusic:
@dyfer thanks for looking at this! I'll see what I can do in the new year.

JordanHendersonMusic:
Okie dokie, this should be good now hopefully, all checks are green!

Spacechild1:
Looks like @dyfer has to approve :)

dyfer (approved):
I think this is good now, thanks!

capital-G:
@Spacechild1 I just wonder if the memory bump could actually introduce more cache misses and therefore degrade the performance of the server in general?

If you think this could have an impact I may look into writing some benchmarks to have some comparison.

JordanHendersonMusic:
Have you measured a regression in performance?

capital-G:
Not yet, but I read about cache misses and that these are actually quite impactful and I don't know how the allocator handles fragmentation. I am currently not using the dev version b/c of the API changes :/ But I'll do some comparison later.

Spacechild1:
> @Spacechild1 I just wonder if the memory bump could actually introduce more cache misses and therefore degrade the performance of the server in general?

No. Memory that is never accessed does not take up cache space and therefore can't cause cache misses.

> I don't know how the allocator handles fragmentation.

The amount of RT memory does not change the behavior of the allocator. When you return a chunk of RT memory, the allocator always tries to merge it with adjacent free chunks.

PR #4918, opened 2020-05-09
Remove boost_sync

brianlheim:
## Purpose and Motivation

Fixes #4914 (confirmed locally using the boost from the source release)

I know this looks like a big change, but see #4914 for my reasoning. This is actually quite safe in terms of the amount of code that had to be ported.

There was only one place where i had to make a substantial change, in PyrSched.cpp, and i confirmed locally with a test program that this works as expected.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #7116, opened 2025-08-28
[build] Update Boost integration

### LINKED ISSUE #7114: Cmake Boost implementation needs upgrading?
## Environment

* SuperCollider version: 3.15-dev
* Operating system: macOS and Linux

## Steps to reproduce

```sh
# when building:
cmake -D SYSTEM_BOOST=ON ..

```

<details><summary>Relevant configure output</summary>

```
(...)
-- Found Pthreads: /Applications/Xcode_16.0.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.0.sdk/usr/lib/libpthread.tbd
CMake Warning (dev) at CMakeLists.txt:322 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


CMake Warning at /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1428 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:1553 (_Boost_COMPONENT_DEPENDENCIES)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2165 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:322 (find_package)


-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found Boost: /opt/homebrew/include (found suitable version "1.88.0", minimum required is "1.50.0") found components: thread system filesystem program_options regex test_exec_monitor chrono atomic
(...)
```
</details>

IIUC we should use [`find_package(boost)`](https://cmake.org/cmake/help/latest/module/FindBoost.html) instead of the [current approach](https://github.com/supercollider/supercollider/blob/develop/CMakeLists.txt#L320-L344), though I'm not sure what are the implications of the latter.

dyfer:
## Purpose and Motivation

In this PR:
- for system boost, manually created targets are replaced with the ones provided by boost's "new" (as of ~1.70 or so) cmake integration
- aliases are added for our bundled boost so that these new targets can be used throughout the codebase.
- a custom `Boost::sync` target was added to manage dependency between boost_sync and boost, IIUC
- nova-tt was updated to [remove an unnecessary dependence on boost](https://github.com/timblechmann/nova-tt/pull/9)
- global boost definitions were moved to the `else()` branch of `SYSTEM_BOOST` in the main CMakeLists.txt
- superfluous `boost/chrono` includes were removed.

Fixes #7114 

### Notes
- changes in [external_libraries/CMakeLists.txt](https://github.com/supercollider/supercollider/compare/develop...dyfer:supercollider:topic-boost-cmake?expand=1#diff-53fccc4367b2e00a0b7d14b2c9dc243fe72aa34777244fa8e5c94d658fd84e7c) include moving some lines around for clarity, I can revert those changes if desired
- for clarification I added keywords (`PRIVATE`/`PUBLIC`) for linking throughout [server/plugins/CMakeLists.txt](https://github.com/supercollider/supercollider/compare/develop...dyfer:supercollider:topic-boost-cmake?expand=1#diff-3d1ccc9e1c72318a4384e2c1430e41a1b02019a0352cff7dbe4e6791eac2f9b7) not just for boost stuff; I can revert this if desired
- Minimum boost version was raised to 1.70, but I think we need an even newer version now after the updates for 3.14; I just couldn't find that info readily
- I was trying to move the global definitions (like `BOOST_CHRONO_HEADER_ONLY`) to target definitions, but couldn't get it to work on Windows, so I left them as globals.

## Types of changes

- Bug fix/maintenance

## To-do list

- [x] Code is tested
- [x] All tests are passing


dyfer:
@capital-G would you be able to take a look at this PR?

CMakeLists.txt
```diff
@@ -315,36 +307,25 @@ if (NOT WIN32)
 endif()
 
 if(SYSTEM_BOOST)
-    # https://github.com/supercollider/supercollider/issues/4576#issuecomment-531499017
-    # Boost >= 1.71 ships with its own Boost config file, we ignore it for now.
-    set(Boost_NO_BOOST_CMAKE ON)
+    if(POLICY CMP0167)
+        cmake_policy(SET CMP0167 NEW)
+    endif()
```
capital-G: Can you add a small comment why this is necessary and link to e.g. https://cmake.org/cmake/help/latest/policy/CMP0167.html - this seems really cryptic otherwise

CMakeLists.txt
```diff
@@ -315,36 +307,25 @@ if (NOT WIN32)
...
     set(Boost_USE_MULTITHREADED ON)
-    find_package( Boost 1.50.0 COMPONENTS  thread system filesystem program_options regex test_exec_monitor )
+    find_package(Boost 1.70.0 REQUIRED COMPONENTS thread program_options regex unit_test_framework)
```
capital-G: Replacing `test_exec_monitor` seems fine, but do we actually need to have `unit_test_framework` as a dependency at all? this seems like a dev dependency, not a build depedency - I don't know if we have some way to represent this currently though.

Also interesting that we do not require `boost::system`?

lang/CMakeLists.txt
```diff
@@ -244,8 +242,7 @@ if(SC_HIDAPI)
   endif()
 endif()
 
-target_link_libraries(libsclang boost_thread_lib boost_system_lib boost_regex_lib boost_program_options_lib)
-target_include_directories(libsclang PUBLIC ${boost_include_dirs})
+target_link_libraries(libsclang Boost::thread Boost::program_options Boost::regex Boost::sync)
```
capital-G: maybe make this multi-line so git history becomes more atomic?

```suggestion
target_link_libraries(
  libsclang Boost::thread
  Boost::program_options
  Boost::regex
  Boost::sync
)
```

Please only approve if the indentation is correct - I only eye-balled the indentation from the webui ;)

server/plugins/CMakeLists.txt
```diff
@@ -66,7 +66,8 @@ endif()
 
 foreach(plugin ${plugin_sources})
 	string(REPLACE .cpp "" plugin_name ${plugin} )
-		add_library(${plugin_name} MODULE ${plugin})
+	add_library(${plugin_name} MODULE ${plugin})
+	target_link_libraries(${plugin_name} PRIVATE Boost::headers)
```
capital-G, line 70: I suppose this makes up for the deleted `boost_include_dirs` in its parent folder?

server/plugins/CMakeLists.txt
```diff
@@ -179,6 +180,7 @@ if (SUPERNOVA)
 	foreach(plugin ${plugin_sources})
 		string(REPLACE .cpp "_supernova" plugin_name ${plugin} )
 		add_library(${plugin_name} MODULE ${plugin})
+		target_link_libraries(${plugin_name} PRIVATE Boost::headers)
```
capital-G, line 183: isn't this already covered by line 70?

server/CMakeLists.txt
```diff
@@ -13,7 +13,6 @@ include_directories(${CMAKE_SOURCE_DIR}/external_libraries
 					${CMAKE_SOURCE_DIR}/external_libraries/nova-tt
 					)
 
-include_directories(${boost_include_dirs})
```
capital-G, line 16: We are sure this doesn't impact e.g. the external plugin interface?

capital-G (changes requested):
Wow - thanks for doing this! I think this starts to make our CMake files a bit more easier to comprehend. Ping me when the comments have been resolved and I'll merge it :)

In general I think this PR contains a bit much :D It would have been easier to digest if these would have been multiple annotated commits which would be squashed afterwards. (e.g. make linking private as one commit) - for the next boost extravaganza :)

> superfluous boost/chrono includes were removed.

Props for spotting and fixing the unnecessary import of `chrono`!
some remark: This may affect the content of the `external_libraries/boost` folder when updating to a new boost version since its gets generated according to the header files which are imported in the codebase.
If no other boost lib is depending upon `chrono`, this would lead to the deletion of `chrono` and also its transient dependencies.
I'll think it is to fix this w/ the next boost upgrade, what do you think?

---

> Some note I written for myself, maybe still interesting for others?

I am a bit torn apart on how to move forward with boost since we don't apply/require performance-patches anymore (which were actually in place for chrono and mingw threads -> https://github.com/supercollider/supercollider/blob/086e553340d166fdb7b0adf9bb460a82ace945f6/external_libraries/boost_sc_changes.patch )

a) keep as-is: default to in repo boost and update boost via boost extraction tool
b) make system boost the default boost for building and deprecate provided boost (this could backfire in case boost gets an upgrade which breaks the SC code base which happened!)
~~c) use cmake fetch so boost version bumps are one-liners~~ -> https://github.com/cpm-cmake/CPM.cmake/discussions/448

=> a)/as-is still seems to be most suitable for SC

CMakeLists.txt
dyfer: `unit_test_framework` is used in [server tests](https://github.com/supercollider/supercollider/pull/7116/files#diff-bdbc87e553cdb81bd51f68350106d8cd90a4b6ffe9028c8437f40142500a1281R2). If we want to be very explicit, it should be only included if `ENABLE_TESTSUITE` is `ON`, do you want me to implement that?

[IIUC `boost::system` is not to be linked to anymore](https://www.boost.org/doc/libs/1_70_0/libs/system/doc/html/system.html#:~:text=Boost.System%20is%20now%20header%2Donly.%20A%20stub%20library%20is%20still%20built%20for%20compatibility%2C%20but%20linking%20to%20it%20is%20no%20longer%20necessary.). It's also not available as a component in recent boost.

server/plugins/CMakeLists.txt
dyfer: Yes, AFAIU.

server/CMakeLists.txt
dyfer: I'm not sure. I was under the impression that we only need `target_link_libraries`. But the fact that this builds and runs should be enough maybe?

dyfer:
> In general I think this PR contains a bit much :D It would have been easier to digest if these would have been multiple annotated commits which would be squashed afterwards. 

To clarify:  should I redo the PR breaking it apart into multiple commits (only to be squash-merged afterwards)?

> 
> > superfluous boost/chrono includes were removed.
> 
> Props for spotting and fixing the unnecessary import of `chrono`! some remark: This may affect the content of the `external_libraries/boost` folder when updating to a new boost version since its gets generated according to the header files which are imported in the codebase.

Yes, it likely will.

>  I'll think it is to fix this w/ the next boost upgrade, what do you think?

How would we remember to track this down before the next boost update/extraction?

IMO it's fine to just remove it now and it'll go away from included boost the next time we update.

OR if we want to be super cautious, I can separate this into a new PR and re-run boost extraction tool (with the same version of boost). I think there's one [manual change](https://github.com/supercollider/supercollider/pull/6733) we'd need to apply IIUC...? I don't think re-extracting boost is worth the work though.


> => a)/as-is still seems to be most suitable for SC

Yeah, I'd keep it as is for now. It definitely makes building on Windows easier (how would we handle the [interprocess flag](https://github.com/supercollider/supercollider/pull/7116/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR325) if we used a pre-built boost anyway...?)

I don't think we'll ever completely shed boost, but as some of its features make its way into the standard library, maybe the dependency will become smaller over  time (unless we add boost's tcp/http server to SC or something like that... haha)

server/plugins/CMakeLists.txt
dyfer: Good catch, I'll double check.

CMakeLists.txt
capital-G: Maybe this is actually better while we are at it? But only if this isn't too complex :)

server/CMakeLists.txt
capital-G: Do internal plugins access the server via the same plugin api? If this is the case then yes :) Otherwise it also probably works ;)

capital-G:
> To clarify: should I redo the PR breaking it apart into multiple commits (only to be squash-merged afterwards)?

No, just for the next time :)


> How would we remember to track this down before the next boost update/extraction?

The boost extraction tool should take care of this - and hopefully one of us two will experience the next boost upgrade and maybe we will remember that chrono was removed :) Otherwise the build/CI will verify that this will (still) work - so I don't think we need to track this down somewhere, just wanted to share the info w/ you :)

> IMO it's fine to just remove it now and it'll go away from included boost the next time we update.

I agree.

server/CMakeLists.txt
dyfer: Well, if external plugins need boost headers, then they should take care of that. And they shouldn't link to SC boost anyway.

dyfer:
> > To clarify: should I redo the PR breaking it apart into multiple commits (only to be squash-merged afterwards)?

> No, just for the next time :)

Noted! I'll aim for more modular PRs even if we squash merge them later.

> > How would we remember to track this down before the next boost update/extraction?
> 
> The boost extraction tool should take care of this - and hopefully one of us two will experience the next boost upgrade and maybe we will remember that chrono was removed :)

I guess I misunderstood, I thought you're suggesting not to remove unnecessary chrono includes at this time, so my question was how would we remember to remove chrono includes later if we skip them here?
But I think we're both on the same page now - remove it here and let the script remove them at the next boost upgrade.

> just wanted to share the info w/ you :)

Thanks! We're on the same page then.

EDIT:
`boost_sync` may be [using chrono](https://github.com/supercollider/supercollider/blob/develop/external_libraries/boost_sync/include/boost/sync/support/boost_chrono.hpp#L17) as well (?) So maybe it won't go away immediately. Also, `boost_sync` could possibly be removed once we move to c++20 IIUC, see discussion in #4918.

CMakeLists.txt
dyfer: No too bad, it will be a couple more lines - first creating a list of components, then conditionally adding `unit_test_framework`, then using that list in `find_package`.

server/plugins/CMakeLists.txt
dyfer: Ah, I don't think this is covered by line 70. This is a separate entry for supernova plugins.

CMakeLists.txt
dyfer: done

CMakeLists.txt
dyfer: done

lang/CMakeLists.txt
dyfer: done (I've put `libsclang` and `Boost::thread` in separate lines though... ok?)

dyfer:
@capital-G I think I addressed your review comments, please take a look.

capital-G (approved):
Thanks, lgtm!

PR #7118, opened 2025-08-28
supernova: fix include w/ system boost

tharvik:
## Purpose and Motivation

when compiling w/ system boost (1.88.0) and w/o jack, it fails with
```
SuperCollider-3.14.0-Source/server/supernova/server/server.cpp:278:16: error: ‘tie’ is not a member of ‘boost’; did you mean ‘xtime’?
  278 |         boost::tie(min, max) = thread_priority_interval_rt();
      |                ^~~
      |                xtime
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/supernova/server/server.cpp
+++ b/server/supernova/server/server.cpp
@@ -18,6 +18,8 @@
 
 #include <stdexcept>
 
+#include <boost/tuple/tuple.hpp>
+
 #include "SC_Win32Utils.h"
 
 #include "nova-tt/thread_affinity.hpp"


dyfer:
Thanks @tharvik 
FWIW at least on macOS we build with system boost 1.89 and I haven't seen that issue.
Also, coincidentally I have just been [working on cleaning up boost cmake integration](https://github.com/supercollider/supercollider/pull/7116) and I wonder if it changes anything. Could you test https://github.com/supercollider/supercollider/tree/topic-boost-cmake on your system?

tharvik:
> FWIW at least on macOS we build with system boost 1.89 and I haven't seen that issue.

hum, maybe it's the combination of cmake flags I'm using? I can paste some more info if you want to look more into it.

> Also, coincidentally I have just been [working on cleaning up boost cmake integration](https://github.com/supercollider/supercollider/pull/7116) and I wonder if it changes anything. Could you test https://github.com/supercollider/supercollider/tree/topic-boost-cmake on your system?

haa, yes, I saw it too, and hoped that it would fix my build but to no avail.

dyfer:
> hum, maybe it's the combination of cmake flags I'm using? I can paste some more info if you want to look more into it.

Could you please file this as an issue, instead of pasting here? Include system details, as well as cmake invocation and full output of the configure step (in addition to the error, of course).

Thanks!

dyfer:
I think we should actually modernize the use of tuples (which will remove boost dependency here). Can you try https://github.com/supercollider/supercollider/tree/topic/cpp-structured-bindings ?

dyfer:
@tharvik if you confirm that the linked branch (which is the base for #7119) fixes this, I'd suggest that this PR is closed and superseded by 7119.

tharvik:
> Could you please file this as an issue, instead of pasting here? Include system details, as well as cmake invocation and full output of the configure step (in addition to the error, of course).

issue opened with some way to reproduce it.

> I think we should actually modernize the use of tuples (which will remove boost dependency here).

good idea, boost tuples are no really needed anymore w/ modern c++.

> Can you try https://github.com/supercollider/supercollider/tree/topic/cpp-structured-bindings ?

and that works, well done! closing this PR then in favor of #7119 

thanks for the help!

dyfer:
Thanks @tharvik for reporting!

Issue #1634, opened 2015-08-02
Refactor new object allocation and GCWrite.


scztt:
This is a feature for the extended refactoring discussed in #1476 and #1624.

Some probable goals:
- Provide an alternative interface for the "allocate new object, attach to an existing object, and perform GC Write" pattern for common cases.
- Make various kinds of new obj assignments explicit at compile time (e.g. differentiate stack / obj assignment, make explicit types for unattached / attached new object ptrs). Provide compile-time guarantees of correct behavior where possible.
- Provide asserts for common assumptions about e.g. objects being white/grey/etc.
- Remove the requirement for GC implementation details from primitives (this will make further GC refactoring / replacement easier)
- Provide alternatives to common cases that require pointer arithmetic.

Some extended goals:
- Since the above requires some non-trivial changes to primitives, a general refactor of primitive calling might be justified.

The above should be done in a way that allows primitives to be refactored one-by-one as needed, to allow bug risk / code churn to be isolatable. The goal is to provide safer new patterns, and (1) use them for new primitives or (2) reduce complexity / risk in existing primitives - blanket refactoring of ALL existing primitives should be a long-term goal but not strictly tied to this feature.

muellmusik:
As I've got a little bit of down time, I've started thinking about this again. I think the problems we need to address (or at least prevent in normal usage) are:

1. Make sure that created objects are made reachable between calls to object creation functions.
2. Make sure that objects made reachable by storing in another object get the correct corresponding calls to GCWrite.
3. Ensure that parent object sizes are correctly updated.
4. Make use of GCWriteNew rather than GCWrite where appropriate.
5. Catch any objects created which are unused. (This seems like it must be an error.)
6. Make common errors in things like stack access less likely.

Possible elements of a reworking:

**Track last allocation**: Basically the idea is to keep track of the last allocated object and raise an alarm if another allocation happens before that is made reachable. This could be done in different ways. There are currently some explicit exceptions to this rule to do with arrays (```runGC=false```) which we may need to continue to allow for performance reasons.

**PrimitiveCall Class**: @scztt suggested this in #1476, and a lot of this has already been done in the Qt stuff. We could use this to provide a more convenient and readable way to access the stack and primitive args, set the receiver, handle errors and eliminate a lot of boilerplate. The local PrimitiveCall instance might be a good place to _track the last allocation_. (In general that should at least be local to the VM, as we may want multiple VMs in the future, and I think within the same primitive is not bad.)

[...]

PR #7133, opened 2025-09-07
SCLang: avoid collecting inside primitives

JordanHendersonMusic:
## Purpose and Motivation

When you allocate inside a primitive, if you don't connect that resource to the memory graph the next time you allocate (and call collect) the object might be collected, leaving you with an invalid memory resource. This has caused many bugs in the past. This PR simply holds off calling collect until the primitive has finished.

This has the potential draw back that if you are creating many temporary objects, you might run out of memory. But none of the primitives do this right now, and the ones that do allocate a lot of temporary objects usually disable the collection cycle anyway to avoid having the GC free the objects before they are placed on the stack.

## Types of changes

- Memory management improvement

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/lang/LangSource/GC.h
+++ b/lang/LangSource/GC.h
@@ -124,7 +124,7 @@ public:
         }
     }
 
-    /// Don't called collect immediately, wait until some context has finished, then call collect.
+    /// Don't call collect immediately, wait until some context has finished, then call collect.
     /// This is implemented primarily for use inside primitives, where collecting while creating temporary objects can
     /// lead to them being freed unless care is taken.
     void enterDelayedCollectionContext() {
--- a/lang/LangSource/GC.h
+++ b/lang/LangSource/GC.h
@@ -124,6 +124,21 @@ public:
         }
     }
 
+    // Don't call collect from inside primitives, instead wait until leaving the primitive.
+    // This ensures that objects created inside the primitive are not destroyed by the collection cycle.
+    void pushInsidePrimitive() {
+        insidePrimitive = true;
+        attemptedToCollectInsidePrimitive = false;
+    }
+
+    void popInsidePrimitive() {
+        insidePrimitive = false;
+        if (attemptedToCollectInsidePrimitive) {
+            Collect();
+            attemptedToCollectInsidePrimitive = false;
+        }
+    }
+
     // users should not call anything below.
 
     void Collect();
@@ -207,6 +222,9 @@ private:
     unsigned char mBlackColor, mGreyColor, mWhiteColor, mFreeColor;
     bool mCanSweep;
     bool mRunning;
+
+    bool insidePrimitive { false };
+    bool attemptedToCollectInsidePrimitive { false };
 };
 
 inline void PyrGC::DLRemove(PyrObjectHdr* obj) {
--- a/lang/LangSource/GC.h
+++ b/lang/LangSource/GC.h
@@ -128,13 +128,13 @@ public:
     /// This is implemented primarily for use inside primitives, where collecting while creating temporary objects can
     /// lead to them being freed unless care is taken.
     void enterDelayedCollectionContext() {
-        delayCollection = true;
-        attemptedToCollectWhenDelayed = false;
+        mDelayCollection = true;
+        mAttemptedToCollectWhenDelayed = false;
     }
     void exitDelayedCollectionContext() {
-        delayCollection = false;
-        if (attemptedToCollectWhenDelayed) {
-            attemptedToCollectWhenDelayed = false;
+        mDelayCollection = false;
+        if (mAttemptedToCollectWhenDelayed) {
+            mAttemptedToCollectWhenDelayed = false;
             Collect();
         }
     }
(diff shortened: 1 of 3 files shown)


Spacechild1:
I like this!

> When you allocate inside a primitive, if you don't connect that resource to the memory graph the next time you allocate (and call collect) the object might be collected, leaving you with an invalid memory resource. 

This is indeed a dangerous pitfall! Not disabling the GC for temporary objects is an easy mistake to make. It's just a Boolean argument in `newPyrObject`, `newPyrString`, `newPyrArray`, etc. and as such it's really easy to overlook. I've made this mistake myself in the past and only discovered it by accident.

JordanHendersonMusic:
Yes it is easy to do! 

I've been writing a lot of zig code recently which forces you to think in terms of allocators and to find patterns of memory lifetimes, which made me realise if primitives used something like an area allocator this issue goes away. 

There was a proposal (@muellmusik I think you mentioned this once?) to model this relationship using c++'s object model, like you would in rust, but I think it is a bit overkill. 

You do technically lose the ability to free temporary objects made inside the primitive with this approach,  but SC usually has plenty of free memory available as the startup process allocates a lot and frees it once done, and we don't tend to allocate too much inside the existing primitives anyway.

Spacechild1:
> You do technically lose the ability to free temporary objects made inside the primitive with this approach,

We can always provide an escape hatch for people who (think they) know what there are doing. E.g. a function like `enableGCinPrimitive`.

JordanHendersonMusic:
Sure, nice and easy, it can just disable the 'insidePrimitive' bool and be called at the top of the primitive.

JordanHendersonMusic:
Okay I've added that, I've also renamed these methods because it doesn't seem right that the GC knows about primitives, it is now referred to as a 'delayed collection context'.

muellmusik:
Hey @JordanHendersonMusic! At a quick glance this looks great! My favourite type of solution: One that just handles it for you. 

My only caveat (besides test test test) would be that it's worth considering this within a context of a broader rework of allocation, which I think would still be a good idea as there are a number of similar pitfalls. That said, I can't see any particular issues with that based on a quick look, but it's perhaps worth thinking through.

> @muellmusik I think you mentioned this once?

Yup, I think this was the last version of a PR attempting to look at this broadly: https://github.com/supercollider/supercollider/pull/3384

I laid out my understanding of the problems to address here: https://github.com/supercollider/supercollider/issues/1634#issuecomment-354433378

JordanHendersonMusic:
Thanks!

Re testing...

I've checked this does indeed call collect when leaving a primitive, and have checked there aren't any memory leaks directly resulting from this code. I don't really know how else to test it as the worst that can happen is that collect isn't called until the next primitive call. I think of this change as meaning we no longer have to check all the primitives for this use-after-unintended-free bug, but instead just have to check the delayed  collect works (which it does).

Re the other issues raised in that comment, I think a lot of them aren't really related to allocation directly, and can instead be fixed through helper methods. I've got a bunch of ideas about this... But I don't like any of them because they all involve complex c++ concepts (or worse, macros) which I'm really keen to avoid.

JordanHendersonMusic:
I think this is good to be reviewed, and potentially merged if no one sees any issues. I've done some more testing and can't find any issues, so it really just needs to be tested in the development branch.

muellmusik:
Would it be worth it to add something to the test suite?

Otherwise I think this looks really good.

JordanHendersonMusic:
The only way I can think to test this, which is what I have been doing, is to run a bunch of code (I've been running parts of the sclang unit test suite) and check the memory usage doesn't grow. I don't think that is something that can be reasonably tested using our existing tooling, but I'm happy for suggestions!

muellmusik (approved):

muellmusik:
> I don't think that is something that can be reasonably tested using our existing tooling, but I'm happy for suggestions!

I guess what would be nice would be some test that using the existing or new gc introspection methods just checks that it's working as predicted in a range of cases (i.e. in and out of collections, etc). I suspect that could be a little finicky to get right though, and I don't want it to hold up this PR.

JordanHendersonMusic:
I don't think I'll add anything like that in this pr. But in future I think the GC part of the code base could be reworked to allow for proper unit tests.

JordanHendersonMusic:
@Spacechild1 do you think this is good too? If so I'll merge.

lang/LangSource/GC.h
```diff
@@ -207,6 +229,9 @@ class PyrGC {
     unsigned char mBlackColor, mGreyColor, mWhiteColor, mFreeColor;
     bool mCanSweep;
     bool mRunning;
+
+    bool delayCollection { false };
+    bool attemptedToCollectWhenDelayed { false };
```
Spacechild1: I think these should have a `m` prefix like all the other member variables.

Spacechild1:
Just a small stylistic thing. Otherwise this looks good!

Spacechild1 (approved):

PR #7142, opened 2025-09-18
Update `.asInteger` and `.asFloat` in `String.schelp`

### LINKED ISSUE #7137: Update `.asInteger` and `.asFloat` documentation
## Environment

* SuperCollider version: 3.15.dev.
* Operating system: macOS

## Steps to reproduce

```
"2e3".asFloat// -> 2000.0

"2e3".asInteger //-> 2
```

- Expected: `"2e3".asInteger` should return `2000`.
- actual: `"2e3".asInteger` returns `2`.

prko:
closes #7137

## Purpose and Motivation
The behaviour of `.asInteger` and `.asFloat` for the `String` class is different: `.asInteger` does not treat `e` as an exponent.
This PR adds some explanation about this behaviour.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


telephon (approved):

capital-G:
@supercollider/documentation should this be merged?

HotwheelsSisyphus (changes requested):
I don't think this should be merged as is, I find the explanations a bit unclear; I'll try to suggest a better formulation within the next few hours

HotwheelsSisyphus:
Alright, here's my suggestion. I did not want to push this directly, that would have felt impolite, but I also don't want to waste people's time being nitpicky... uff.
My main concern is that "does not regard the exponent represented as emphasis::e::" is not the most accessible way of formulating this, and also doesn't make the behavior explicit (i.e., what happens if you do pass it a string with an e).

```
discussion::
The method code::.asInteger:: does not understand scientific notation (e.g., teletype::2e3:: for teletype::2000::), 
and simply ignores all characters in the string after its first nonnumeric character 
(excepting signs teletype::+,-:: at the beginning of the string). 
If you use scientific notation, use link::#-asFloat:: instead.
```

```
The method code::.asFloat:: understands scientific notation (e.g., teletype::2e3:: for teletype::2000::,teletype::2e-3:: for teletype::0.002::),
and accepts uppercase teletype::E:: and lowercase teletype::e::.
By contrast, the method link::#-asInteger: does emphasis::not:: accept scientific notation.

```

prko:
@HotwheelsSisyphus  
Thank you — your explanation is much clearer than mine!  
If you’d like, you could directly correct this and be listed as a coauthor of the PR, which I think would be the simplest approach. 
If you’d prefer not to, just let me know — I can either make the correction myself, or revert this part so you can open a new PR.

HotwheelsSisyphus (approved):

prko:
Thanks!

PR #7144, opened 2025-09-21
SCLang: add fast Hungarian Algorithm solving

dantepfer:
## Purpose and Motivation

The Hungarian Algorithm, aka Linear Assignment or Munkres, is useful in music for voice-leading and other applications (see https://en.wikipedia.org/wiki/Hungarian_algorithm). @tedmoore previously implemented a fully SC Class that works well (see https://github.com/tedmoore/Linear-Assignment). However, it's slow for large matrices. The implementation in this commit is in C++ and is much faster. For example, solving a 100x100 matrix with Ted's class takes ~1.5s, while this implementation takes ~0.0005s for the same matrix.

## Types of changes
- New feature

## To-do list
- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Array.schelp
+++ b/HelpSource/Classes/Array.schelp
@@ -419,6 +419,18 @@ code::
 [0.1, [\s_new, \default, -1, 1, 1, \freq, 1961]].asRawOSC; // bundle
 ::
 
+method::hungarianSolve
+Returns the solution to the Linear Assignment problem, using the Hungarian Algorithm, also known as Munkres, in the format [lowest_cost, [assignments]]. Receiver must be a 2-dimensional Array of arrays, but it need not be square, and may contain negative numbers. Implementation is in C++, so it's fast!
+code::
+//example from Wikipedia ("https://en.wikipedia.org/wiki/Hungarian_algorithm")
+[
+   [8, 4, 7],
+   [5, 2, 3],
+   [9, 4, 8]
+].hungarianSolve;
+//returns [15.0, [0, 2, 1]]
+::
+
 subsection:: Bela
 
 method:: belaScope
--- a/HelpSource/Classes/Array.schelp
+++ b/HelpSource/Classes/Array.schelp
@@ -420,40 +420,55 @@ code::
 ::
 
 method::hungarianSolve
-Returns the solution to the Linear Assignment problem, using the Hungarian Algorithm, also known as Munkres, in the format code::[lowest_cost, [assignments]]::.
+Given a two-dimensional array, in which rows represent workers, columns represent tasks, and values represent the cost for each worker to do each job, this returns the assignment of workers to tasks which has the lowest total cost.
+This is known as the Linear Assignment Problem, and here it is solved using the Hungarian Algorithm, also known as Munkres, which is enormously faster than solving it by brute force.
+The result is given in the format code::[lowest_cost, [assignments]]::, where the code::[assignments]:: array maps each row to its assigned optimal column. For example, assignments code::[0, 2, 1]:: means that for the lowest overall cost, worker 0 should do job 0, worker 1 should do job 2, and worker 2 should do job 1.
 Receiver must be a 2-dimensional Array of arrays, but it need not be square (zeroes will be added when necessary), and may contain negative numbers and floats.
-This algorithm is useful in music for voice-leading and other applications, such as assigning notes to instruments.
+This algorithm is useful in music for voice-leading and any other problem, such as assigning notes to instruments, which can be framed as a cost-minimization task.
 
 code::
-// Basic example from Wikipedia ("https://en.wikipedia.org/wiki/Hungarian_algorithm")
+// Basic example from Wikipedia (https://en.wikipedia.org/wiki/Hungarian_algorithm)
+(
 [
    [8, 4, 7],
    [5, 2, 3],
    [9, 4, 8]
 ].hungarianSolve;
-//returns [15.0, [0, 2, 1]]
+//—> [15.0, [0, 2, 1]]
+)
 
-// Voice leading between two chords
+// Musical example: Voice-leading between two chords
 (
-~voiceLead = { |src, trg|
-    var cm = src.collect { |n1| trg.collect { |n2| (n1 - n2).abs } };
-	var asn = cm.hungarianSolve[1];
-    asn.collect { |t, s| [src[s], trg[t]] }
+//You have two chords, and you wish to know which note in the first chord should move to which note in the second chord for optimal voice-leading.
+~sourceChord = [72, 60, 64, 67]; //C
+~targetChord = [67, 74, 65, 59]; //G7
+
+//Build a cost matrix with interval distance from one note to the next as the cost. In other words, the farther the target note from the source note, the more 'expensive' it is for the source note to go to the target note. Other, more sophisticated cost functions are possible; for example, one could build in an extra penalty for leaps.
+~costMatrix = ~sourceChord.collect { |n1| ~targetChord.collect { |n2| (n1 - n2).abs } };
+
+//Take a look at the resulting cost matrix
+"Cost Matrix:".postln; ~costMatrix.do { |r| r.postln };
+// —> [5, 2, 7, 13]
+//    [7, 14, 5, 1]
+//    [3, 10, 1, 5]
+//    [0,  7, 2, 8]
+
+//Solve the cost matrix, i.e. find the assignments of source note to target note that has the lowest overall cost
+~assignments = ~costMatrix.hungarianSolve[1];
+// —> [1, 3, 2, 0]
+
+//Associate each source note with its optimal target note, following the assignments
+~voiceLeading = ~assignments.collect { |targetIndex, sourceIndex| [~sourceChord[sourceIndex], ~targetChord[targetIndex]] };
+
+//Print the results!
+"Best Voice-Leading:".postln; ~voiceLeading.do { |assignment|
+    var sourceNote = assignment[0], targetNote = assignment[1], interval = targetNote - sourceNote;
+    ("Note " ++ sourceNote ++ " → " ++ targetNote ++ " (interval: " ++ interval ++ ")").postln;
 };
-
-~chC = [72, 60, 64, 67]; //C
-~chG = [67, 74, 65, 59]; //G7
-
-~cMat = ~chG.collect { |n1| ~chG.collect { |n2| (n1 - n2).abs } };
-
-~cMat.do { |r| r.postln };
-
-~vLead = ~voiceLead.(~chC, ~chG);
-
-~vLead.do { |p|
-    var src = p[0], trg = p[1], int = trg - src;
-    ("Note " ++ src ++ " → " ++ trg ++ " (interval: " ++ int ++ ")").postln;
-};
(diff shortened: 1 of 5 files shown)


capital-G:
Hey - thanks for this PR, I think this is a really useful feature to have in the core!

Currently there are some formatting issues, you can take a look at https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit how you can use `pre-commit` to get the code in the demanded shape (assuming you are on unxi, use `FULL_CHECK=1 pre-commit run --all-files`) - or you can use clang-format directly.

It would also be great to have some unit tests for this feature with some nasty values like nan, inf, 0, ... just to check that the language doesn't crash or stall on such values. You can take a look at the existing unit test at https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestArray.sc or use the guide https://github.com/supercollider/supercollider/wiki/Unit-Testing-Guide

I think the naming is okay, `Array.hungarianSolve` - what do you think @telephon ?

lang/LangPrimSource/PyrArrayPrimitives.cpp
```diff
@@ -2616,6 +2617,79 @@ int prArrayIsRectangular(struct VMGlobals* g, int numArgsPushed) {
...
+    if (err) return err;
+
+    // Empty -> return [0, []]
+    if (C.empty() || C[0].empty()) {
+        PyrObject* out = newPyrArray(g->gc, 2, 0, true);
+        SetObject(a, out);
+        SetFloat(out->slots + 0, 0.0);
+        out->size = 1;
+
+        PyrObject* asg = newPyrArray(g->gc, 0, 0, true);
```
JordanHendersonMusic, line 2662: Since these gc allocated objs are not connected to the memory graph (not on the stack), there is a chance the GC will delete the resource at the next allocation.

I've got a pr to make this bug impossible, but right now, this code will sometimes crash. You can set that final `true` to a `false` to fix this issue on all GC allocations.

HelpSource/Classes/Array.schelp
```diff
@@ -419,6 +419,18 @@ code::
 [0.1, [\s_new, \default, -1, 1, 1, \freq, 1961]].asRawOSC; // bundle
 ::
 
+method::hungarianSolve
+Returns the solution to the Linear Assignment problem, using the Hungarian Algorithm, also known as Munkres, in the format [lowest_cost, [assignments]]. Receiver must be a 2-dimensional Array of arrays, but it need not be square, and may contain negative numbers. Implementation is in C++, so it's fast!
```
JordanHendersonMusic: Could you also explain, briefly, why this is musically useful? Perhaps with an example?

Also no need to mention how it is implemented.

dantepfer: @capital-G thanks for your feedback! I've addressed both your points in the current commits. 

@JordanHendersonMusic I've extended the help to include musical applications, with an example by @smoge  which I grabbed from a conversation on this topic on the forum.

dantepfer: @JordanHendersonMusic was there another change that you wanted me to make in the C++ code? I got an email to that effect but can't find it here on GitHub. Thanks!

JordanHendersonMusic: No I misread your code then deleted it. The implementation looks fine to me!

HelpSource/Classes/Array.schelp
```diff
@@ -419,6 +419,43 @@ code::
 [0.1, [\s_new, \default, -1, 1, 1, \freq, 1961]].asRawOSC; // bundle
 ::
 
+method::hungarianSolve
+Returns the solution to the Linear Assignment problem, using the Hungarian Algorithm, also known as Munkres, in the format [lowest_cost, [assignments]]. Receiver must be a 2-dimensional Array of arrays, but it need not be square (zeroes will be added when necessary), and may contain negative numbers and floats.
```
capital-G: Should we wrap the return in a code context as it indicates an array?
Also new lines after each sentence are a good practice within scdoc.

Only a suggestion - I am fine also in its current state.

```suggestion
Returns the solution to the Linear Assignment problem, using the Hungarian Algorithm, also known as Munkres, in the format code::[lowest_cost, [assignments]]::.
Receiver must be a 2-dimensional Array of arrays, but it need not be square (zeroes will be added when necessary), and may contain negative numbers and floats.
```

capital-G (approved):
Thanks for revisiting and adjusting! The PR looks now very good from my perspective, only an optional suggestion which is probably too nitpicky.

I'd still wait for some feedback from the @supercollider/class-library team and if they also approve we merge it!

Mergenote: Should be squashed

HelpSource/Classes/Array.schelp
dantepfer: @capital-G Great suggestions both — it took me a sec to realize that the new lines don't actually affect final layout but are just for readability of the scdoc raw text.

JordanHendersonMusic:
My only question is whether this is really necessary as we shouldn't add primitives without consideration. 

Is the SC implementation as good as it could be? Why is it slow? Are we sure it isn't a big O type of issue? Could it be *fast enough*?

I haven't had a close look at the sc implementation and I'm not after benchmarks or anything too in depth, but if we could briefly look at why it isn't fast enough that would help explain why this addition is needed.

dantepfer:
@JordanHendersonMusic Excellent question. First of all, here's the thread at scsynth.org that started me down this path: https://scsynth.org/t/hungarian-algorithm-munkres-in-sc/11371. One takeaway from the thread is that it is non-trivial to write a good Munkres implementation in SC, even when taking a canonical one in another language as an example. 

Beyond that, the main thing to consider is that Munkres is by nature about optimization. When brute-forced, the Assignment Problem it solves requires an amount of computation that grows factorially with the size of the matrix. As an example, you'll see code I wrote in the thread above that brute-forces the problem in SC, and it takes 6 seconds to solve a 10x10 matrix on my M3 Mac that way. 

So the whole purpose of Munkres is to solve the problem as quickly as possible. The class that @tedmoore wrote is excellent, as far as I can tell: a clean and faithful implementation in pure SC of this JavaScript library, which has 103K downloads/month and seems to be the canonical JS version: https://github.com/addaleax/munkres-js.

It's possible that it could be optimized further, but I don't think it'd be easy. Here's a relevant point of comparison:

In Python's canonical scipy library, Munkres used to be implemented in pure Python: https://github.com/scipy/scipy/blob/v0.18.1/scipy/optimize/_hungarian.py#L13-L107

As of scipy 1.4.0, it is now implemented as a compiled C++ library, "to improve performance" (see https://docs.scipy.org/doc//scipy-1.16.2/release/1.4.0-notes.html). 

I wrote a script to compare the interpreted vs compiled C++ Python versions (script attached further below). Even for these two optimized versions, the difference is substantial:

```
Matrix: 100x100, dtype=int64
Python implementation : 12.31 ms, cost=1831
SciPy implementation  : 0.10 ms, cost=1831
Speed ratio (Py/SciPy): 120.93x

Matrix: 400x400, dtype=int64
Python implementation : 606.25 ms, cost=1383
SciPy implementation  : 3.12 ms, cost=1383
Speed ratio (Py/SciPy): 194.38x

Matrix: 700x700, dtype=int64
Python implementation : 3993.06 ms, cost=1297
SciPy implementation  : 10.18 ms, cost=1297
Speed ratio (Py/SciPy): 392.19x
```

I suspect that in the SuperCollider case, where the difference is even more extreme, we're running up against the overhead of SC Arrays in an operation that requires many steps. Optimization of the SC class may speed it up, but I'm quite sure it'll never approach the speed of the C++ implementation. And as mentioned, the whole purpose of Munkres is speed. 

[compare_interpreted_to_compiled_hungarian.py.zip](https://github.com/user-attachments/files/22497812/compare_interpreted_to_compiled_hungarian.py.zip)

JordanHendersonMusic:
@dantepfer thanks that all makes sense! This seems like a good thing to have as a primitive then.

dantepfer:
@capital-G @JordanHendersonMusic I'm curious if there's anything I can do to help shepherd this along to be included into the next version of SC. It's my first time contributing, so please let me know if there's anything I can do to help — or is this already as much in the pipeline as it can be?

capital-G:
Hey - we are still waiting for a review from the @supercollider/class-library team - since we are adding a new functionality to the class library (which can't be removed in the future) it is best to get it as good as possible on the first try - maybe they have some feedback regarding naming/functionality.
Afterwards this PR can be merged.

I also added it to the 3.15 release project.

HelpSource/Classes/Array.schelp
```diff
@@ -419,6 +419,44 @@ code::
 [0.1, [\s_new, \default, -1, 1, 1, \freq, 1961]].asRawOSC; // bundle
 ::
 
+method::hungarianSolve
+Returns the solution to the Linear Assignment problem, using the Hungarian Algorithm, also known as Munkres, in the format code::[lowest_cost, [assignments]]::.
```
JordanHendersonMusic: This line
> Returns the solution to the Linear Assignment problem, using the Hungarian Algorithm, also known as Munkres...

Doesn't really tell the average user of sc (a musician) what and why this method is useful. While useful, I don't think this should be the first thing the user reads.

> ...  in the format code::[lowest_cost, [assignments]]::.

I don't understand what the term 'assignments' means? Is it the indices of the elements returned?

HelpSource/Classes/Array.schelp
```diff
@@ -419,6 +419,44 @@ code::
...
+Receiver must be a 2-dimensional Array of arrays, but it need not be square (zeroes will be added when necessary), and may contain negative numbers and floats.
+This algorithm is useful in music for voice-leading and other applications, such as assigning notes to instruments.
+
+code::
```
JordanHendersonMusic, line 429: Would you mind adding some comments showing what the code returns, at least for the final result? Having to run all the code to figure out if this is useful in a musical setting is a little clumsy, particularly if you are reading the documentation in a browser that doesn't allow code execution.

JordanHendersonMusic (changes requested):
This looks good to me, I agree it would be good to get a once over from those specifically dealing with the class library. 

I've just got a few quibbles with the documentation that I think are easily addressed as I don't think it reads all that clear if you don't already know what this algorithm does.

HelpSource/Classes/Array.schelp
dantepfer: @JordanHendersonMusic Thanks much for this. Looking at this again now, you're right that it could be much clearer. I just submitted a new commit with more general explanation and a much-clarified musical example, complete with printed-out results in comments.

HelpSource/Classes/Array.schelp
dantepfer: @JordanHendersonMusic I've greatly improved this (I hope) in the latest commit!

HelpSource/Classes/Array.schelp
JordanHendersonMusic: Thanks! This looks really good now, great work!

JordanHendersonMusic (approved):

capital-G:
@JordanHendersonMusic @telephon can we merge this?

telephon:
Thank you :)

PR #7146, opened 2025-09-22
build: cleanup portaudio cmake integration

dyfer:
## Purpose and Motivation

This PR cleans up our Cmake PortAudio configuration.

- system portaudio is now only found through pkg-config
- bundled portaudio is exposed as a new target
- linking was updated to use the new target
- EDIT: I moved the `add_subdirectory` call to all places where portaudio is found inside sclang/scsynth/supernova, as I think that's cleaner... but I'm totally open to reverting this; I kept it as a separate commit for easier review.

Note that previously we seemed to have _also_ supported system portaudio on *nix _without pkg-config_. I don't think this scenario is needed these days. It was not documented anyway, AFAICT. Please let me know if you think we still need to support it.

I believe this PR also fixes one _intermittent_ issue I sometimes observed on macOS (where portaudio is used for supernova): sometimes incremental builds would fail because of an issue with linking flags for portaudio.

I removed the copyright info from the `FindPortaudio.cmake` file since it now has no resemblance to the original and is also substantially smaller (3 LOC). Is that the way to go or should I still include old copyright info? 

## Types of changes

- Cleanup
- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
@jrsurge @capital-G what do you think of these changes?
In particular, do you think that adding portaudio to the build tree in three separate places (lang/servers) in 44dd915c62543b2b8cda5df148307c21dfbb4572 makes sense, or should we just go with changes I made in 365c6b5660594ec57cf173fe1a23ce983de75642?

capital-G (approved):
I am not too familiar w/ CMake and the building process of Portaudio, but your PR makes it for me clearer to understand what's going on, so I'll think its good to be merged.

I think it is fair to remove the copyright notice if the code has been replaced.

Feel free to merge or wait for another review by a person w/ more experience regarding CMake :)

Edit: Sorry, you asked which approach I'd prefer - I'd prefer https://github.com/supercollider/supercollider/commit/365c6b5660594ec57cf173fe1a23ce983de75642 since the other looks easy to diverge over time.

dyfer:
> I'd prefer [365c6b5](https://github.com/supercollider/supercollider/commit/365c6b5660594ec57cf173fe1a23ce983de75642) since the other looks easy to diverge over time.

The more I thought about it the more I leaned towards this one as well. Thanks!

PR #7149, opened 2025-09-23
docs/reference/iteration update: explain break, more details re .repeat/loop

HotwheelsSisyphus:
title says it mostly.

Hopefully this is more approachable than what we had before.

for now I've left `do, for and forBy` alone, will look at that at a later stage.

please take look at the indentation in the .block examples. I can't make it look right.

render here:
[Iteration sep 23.pdf](https://github.com/user-attachments/files/22503115/Iteration.sep.23.pdf)

Regarding repeat/loop I tried to do some simplifying, I hope it's not misleading. I still struggle to understand why the docs sometimes refer to a Routine as a Routine and sometimes as a Stream (I do understand that it's a subclass of Stream, but still...)

Also as an aside, not sure what to make of this (from the class library, Core/Function.sc):
```
loop {
		// loop is supported magically by the compiler,
		// thus it can be implemented in terms of itself
		loop { this.value };
	}
```
I guess, sufficiently advanced technology? Does this explain why sc crashes when loop is called outside a Routine?

### DIFF
--- a/HelpSource/Reference/iteration.schelp
+++ b/HelpSource/Reference/iteration.schelp
@@ -1,101 +1,87 @@
 title:: Iteration
 summary:: for, do, loop, repeat
-categories:: Core,Common methods, Streams-Patterns-Events
+categories:: Core,Common methods
 related::  Classes/Object, Classes/Function, Classes/Routine, Classes/Stream, Classes/Pattern
 
-(This page combines material formerly in the Control Structures and loop/repeat
-helpfiles.)
+
+(This page combines material formerly in the Control Structures and loop/repeat helpfiles.)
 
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
Take a look at `while`'s implementation too.

The reason they both look like this is to do with inlining. When the body can be inlined, the compiler provides the implementation (same with `and` and a few others). `while` and `loop` can always be rewritten in an inlined form, so the compiler can always implement it, and no SC Lang implementation is ever needed.

tedmoore (approved):
I like how you've explained this stuff. Also, I think comparing common control structures to the way other languages do things is very often helpful for people coming with some pre-existing coding knowledge.

PR #7150, opened 2025-09-24
Standardize dash usage in schelp files

### LINKED ISSUE #7148: Updating schelp files: replacing `--` (old style dash) text with `—` (Em dash)
In reviewing the schelp files, I’ve noticed that the `--` sequence is being used as a substitute for a dash in non-code text and comments. While this may have been a legacy convention, it’s worth noting that `--` is also a method in sclang, which makes its use in prose and commentary potentially confusing or suboptimal.

Therefore, I intend to replace all instances of `--` (used stylistically as a dash) with the proper Em dash (`—`) for clarity and consistency. I’ll carry out this update at a later point. If anyone has any suggestions or objections, I’d be more than happy to take them into consideration.

A search using the regular expression `[^-<]--(?!(-|>|"|\())` has returned 147 matches across 66 files. As the results also include instances of the method `--`, I’ll need to review each line carefully during the conversion to ensure accuracy.
<details><summary>Details</summary>
<p>

HelpSource/Classes/Collection.schelp:
  632  Return the set of all items which are not elements of both  this and strong::that::.
  633: this -- that
  634  code::

HelpSource/Classes/Compander.schelp:
  20  argument::slopeBelow
  21: Slope of applied amplitude curve if control signal amplitude is below code::thresh::. If > 1.0, the amplitude drops off more quickly the softer the control signal gets; when the control signal is close to 0 amplitude, the output should be exactly zero -- hence, noise gating. Values < 1.0 are possible, but it means that a very low-level control signal will cause the input signal to be amplified, which would raise the noise floor.
  22  

HelpSource/Classes/CompanderD.schelp:
  24  argument::slopeBelow
  25: Slope of the amplitude curve below the threshold. If this slope > 1.0, the amplitude will drop off more quickly the softer the control signal gets; when the control signal is close to 0 amplitude, the output should be exactly zero -- hence, noise gating. Values < 1.0 are possible, but it means that a very low-level control signal will cause the input signal to be amplified, which would raise the noise floor.
  26  

HelpSource/Classes/CondVar.schelp:
  34  code::
  35: // Version 1: `wait` without predicate -- while loop required
  36  (

  66  
  67: // Version 2: `wait` with predicate -- no while loop required
  68  (

HelpSource/Classes/ContiguousBlockAllocator.schelp:
  61  c.alloc(2);  // 5
  62: c.alloc(4);  // nil -- no block of 4 available
  63  

HelpSource/Classes/Date.schelp:
   31  
   32: ## The most typical usage is to create a Date from a year/month/day and possibly hour/minute/second.  All of these properties (except teletype::year::) have default values (teletype::month = 1::, teletype::day = 1::, teletype::hour = 0::, teletype::minute = 0::, teletype::second = 0::). teletype::rawSeconds:: is computed based on the other properties. There is no default year -- if no teletype::year:: is provided, an error is thrown.
   33  

   45  
   46: It is possible to specify dates with "out of range" values that will be resolved into a valid range -- this allows simple date offset calculations to be performed (see example below).
   47  

  123  
  124: If you modify this property, the other properties will not update. See note at link::#-resolve:: -- there is no "resolveFromRawSeconds" method, but you can generate a new self-consistent Date instance by passing the teletype::rawSeconds:: as an argument to link::#*new::.
  125  

  166  
  167: If the instance properties (teletype::year::, teletype::month::, teletype::day::, teletype::hour::, teletype::minute::, teletype::second::) are modified, the Date will be in an inconsistent state -- its teletype::rawSeconds:: won't reflect the modifications, and the properties may fall outside of "normal" ranges (teletype::month: 1-12, day:1-31, hour: 0-23, minute/second: 0-59::).  Also, using the string conversion methods may give strange results. Calling the teletype::resolve:: method will put things back into a consistent state.
  168  

HelpSource/Classes/Demand.schelp:
  40  )
  41: a.set(\t_trig, 1); // do this a few times -- the demand value should increase by 1 each time
  42  a.set(\t_reset, 1);	// goes immediately back to 0

HelpSource/Classes/DiskIn.schelp:
  24  Buffer number
  25: Note:: The Buffer's numFrames must be a power of two and is recommended to be at least 65536 -- preferably 131072 or 262144. Smaller buffer sizes mean more frequent disk access, which can cause glitches. ::
  26  

HelpSource/Classes/DiskOut.schelp:
  40  Buffer.write)
  41: Note:: The Buffer's numFrames must be a power of two and is recommended to be at least 65536 -- preferably 131072 or 262144. Smaller buffer sizes mean more frequent disk access, which can cause glitches. ::
  42  

HelpSource/Classes/Environment.schelp:
  191  
  192: Local variables declared in functions, and class and instance variables, use lexical scope. That is, the context in which they are understood depends on where the declaration is read during compilation. Asynchronous functions -- any function that will execute outside (later than) the current execution flow -- carry their lexically scoped variables with them.
  193  code::

  230  
  231: Most objects are simply returned -- the method call behaves like a getter for any other object.
  232  code::

  254  ::
  255: Environment variables inside a function will refer to the currently active environment -- not to the Environment being addressed. This is to allow the object prototype to interact with the link::#currentEnvironment::.
  256  code::

  269  
  270: Assigning a value into an environment using a setter -- strong::name_():: or strong::.name = ...:: -- posts a warning message if the name is already defined in the class library.
  271  code::

HelpSource/Classes/Exception.schelp:
  10  
  11: The built-in exception types are actually subclasses of link::Classes/Error:: -- see its help file for the hierarchy.
  12  

HelpSource/Classes/FFT.schelp:
  15  code::
  16: // NO: The buffer has two channels -- this will not work
  17  fft = FFT(LocalBuf(2048, 2), aStereoSignal);

HelpSource/Classes/File.schelp:
   90  list::
   91: ## code::true:: -- A new directory was created at code::path::.
   92: ## code::false:: -- A directory already existed at code::path::; a new one was not created.
   93  ::

   99  list::
  100: ## code::true:: -- You can assume that the path no longer exists. (Either it existed and was deleted, or it didn't exist and it still doesn't exist.)
  101: ## code::false:: -- The file could not be deleted (probably a permissions error). Your code should assume that the path still exists.
  102  ::

  108  list::
  109: ## code::true:: -- At least one file was deleted.
  110: ## code::false:: -- No files were deleted.
  111  ::

HelpSource/Classes/IdentityDictionary.schelp:
  38  ## Iteration operations (link::Classes/Dictionary#-do::, link::Classes/Dictionary#-keysValuesDo::) only touch the main-level values.
  39: ## link::Classes/Dictionary#-collect::, link::Classes/Dictionary#-select::, and link::Classes/Dictionary#-reject:: return new dictionaries. The parent and proto will be preserved emphasis::as is:: in the new result, but -- as with code::do:: -- they will not be processed in the iteration, and as with code::copy::, they are identical references.
  40  ::

HelpSource/Classes/Klank.schelp:
  148  
  149: // -- overlap texture
  150  (

  184  
  185: // -- frequency and decay scaling
  186  (

  217  
  218: // -- overlap texture 2
  219  (

  252  
  253: // -- overlap texture 3
  254  (

HelpSource/Classes/LinkClock.schelp:
  52  ## strong::Phase:: || strong::Peer A beats:: || strong::Peer B beats:: || Notes
  53: ## 0 || 32 || -- ||
  54  ## 1 || 33 || -3 || Link calculates -3 for B

HelpSource/Classes/MIDIOut.schelp:
  163  list::
  164: ## code::MIDIOut.newByName("device", "port"):: -- Should be compatible on all systems.
  165: ## code::MIDIOut(index):: -- macOS, Windows (teletype::index:: is the device's index in code::MIDIClient.destinations::). Note that this usage is emphasis::not:: compatible with Linux! macOS and Windows users should not expect Linux users to be able to run this style of connection.
  166: ## code::MIDIOut(0, MIDIClient.destinations[index].uid):: -- Linux only (teletype::index:: is the device's index in code::MIDIClient.destinations::).
  167  ::

HelpSource/Classes/Object.schelp:
  583  
  584: // Set does not understand keysValuesDo -- result is nil
  585  Set[1, 2, 3].tryPerform(\keysValuesDo, { |key, value| [key, value].postln });
  586  
  587: // Error occurs within keysValuesDo -- error is thrown back to halt execution
  588  (a: 1, b: 2, c: 3).tryPerform(\keysValuesDo, { |key, value| [key, value].flippityblargh });

HelpSource/Classes/Pattern.schelp:
  88  
  89: The link::#-play:: method does not return the pattern itself. Instead, it returns the link::Classes/EventStreamPlayer:: object that actually runs the pattern. Control instructions -- stop, pause, resume, play, reset -- should be addressed to the EventStreamPlayer. (The same pattern can play many times simultaneously, using different EventStreamPlayers.)
  90  

HelpSource/Classes/Psync.schelp:
  7  
  8: Psync behaves somewhat like link::Classes/Pfindur:: -- it has a teletype::maxdur:: argument that limits the total duration of the event stream.
  9  

HelpSource/Classes/Quant.schelp:
  23  method::timingOffset
  24: For use with patterns only -- this enables patterns to run slightly ahead of their sounding time on the clock, giving you control over the order in which threads execute.
  25  

  62  
  63: In some cases, you might want two patterns that are sounding at the same time to evaluate in a specific order -- for instance, the second pattern might depend upon data calculated by the first. If they are scheduled on the clock for exactly the same time, you have no control over the order of execution: the second pattern might evaluate first, in which case it would be using stale data for the pattern that should have run first.
  64  

HelpSource/Classes/Rest.schelp:
  48  argument:: value
  49: The Rest instance's numeric value, to be used in math operations. Note that a Rest's value is ignored for most Event keys (assuming the Event does nothing in response to code::.play::). If a Rest appears in a rhythm key (code::dur:: or code::delta::), then the number is the time until the next event. Consequently, numeric Rests are often used for duration -- but there is no requirement that a Rest's value must be a duration.
  50  

HelpSource/Classes/Routine.schelp:
  234  
  235: ⟪ 263 characters skipped ⟫ is 1 and the Routine is waiting for 1 beat, the next "awake" should happen after one second. When rescheduling without a code::quant::, the next "awake" will emphasis::still:: happen after one second -- but the code::beats:: value will adjust for the new clock.
  236  

HelpSource/Classes/Scale.schelp:
  143  Scale.minorPentatonic.pitchesPerOctave; // 12
  144: Scale.ajam.pitchesPerOctave; // 24--this is a quarter-tone scale
  145  Scale.partch_o1.pitchesPerOctave; // 43

  155  code::
  156: Scale.ajam.stepsPerOctave;	// 12 -- quarter-tone scales have normal octaves
  157  ::

HelpSource/Classes/ServerOptions.schelp:
  133  list::
  134: ## code::true:: (default) -- Use the system clock. Timestamped messages will maintain consistent
  135  latency over long sessions, but may not be perfectly sample-accurate.
  136: ## code::false:: -- Use the sample clock. This helps to
  137  support sample-accurate scheduling; however, messaging latency from

HelpSource/Classes/Set.schelp:
   94  
   95: method::symmetricDifference, --
   96  Return the set of all items which are not elements of both this and strong::that::.

  100  symmetricDifference(a, b);
  101: a -- b // shorter syntax
  102  ::

HelpSource/Classes/SoundFile.schelp:
  353  
  354: // It's a proper audio file -- server can load it
  355  b = Buffer.read(s, PathName.tmp +/+ "sf-help.wav");

HelpSource/Classes/SynthDef.schelp:
   47  numberedlist::
   48: ## link::Classes/NamedControl:: -- recommended for flexibility and clarity.
   49  ## Literal arrays (as noted above).

  154  
  155: The exception is very large SynthDefs, where you have a choice between link::#-writeDefFile:: and this method. Even then, the efficiency savings of code::writeOnce:: are only in disk I/O -- both methods build the SynthDef every time they run. ::
  156  

  368  
  369: // Values outside of the array bounds are ignored --
  370  // 0 is dropped

HelpSource/Classes/SynthDesc.schelp:
   41  
   42: For example, by listing argument names and link::Classes/ControlSpec::s under the 'specs' key in the event, the client can use the specs to build a link::Classes/GUI:: allowing control over all the SynthDef's inputs, with sensible range mappings. (The "window" button in the SynthDescLib browser does exactly this -- any ControlSpecs listed in the metadata will be used for the corresponding synth input's slider.)
   43  

   57  ## AbstractMDPlugin || A dummy plug-in, which writes no metadata, so that users who are not interested in metadata will not find extra files in the synthdefs directory.
   58: ## TextArchiveMDPlugin || Writes the metadata as a SuperCollider text archive -- metadata.writeArchive(path). This is the default. Metadata are written only if the SynthDef contains metadata.
   59  ::

  100  // Same SynthDef, but adding metadata
  101: // \freq and \amp exist in the global ControlSpec collection -- Spec.specs
  102  // They are converted to real ControlSpecs using .asSpec

HelpSource/Classes/Task.schelp:
    8  
    9: The purpose of a Task is to separate a Routine's state of execution from its running state (that is, its state of being scheduled on a clock or not, or paused in a link::Classes/CondVar:: or not). Use Task if you expect the process to need to start, stop or resume multiple times while maintaining the execution flow. (This means that Tasks are not 100% interchangeable with Routines -- for many uses, Tasks should be preferred over Routines.)
   10  

  110  
  111: ⟪ 307 characters skipped ⟫k is stopped, it remains scheduled until it wakes up again (based on its wait time). If, during this interval, the task were restarted, there would be two references to the task in the scheduler queue -- a situation that is irrecoverable short of stopping everything with command-period.
  112  

HelpSource/Classes/TempoClock.schelp:
  224  
  225: code::thisThread.clock:: is always the current clock -- so, writing code::thisThread.clock.beatsPerBar = /* number */:: as a matter of habit will guarantee that the meter change only ever applies to the current clock.
  226  

HelpSource/Classes/VDiskIn.schelp:
  16  buffer number
  17: Note:: The Buffer's numFrames must be a power of two and is recommended to be at least 65536 -- preferably 131072 or 262144. Smaller buffer sizes mean more frequent disk access, which can cause glitches. ::
  18  

HelpSource/Guides/Debugging-tips.schelp:
  406  ::
  407: When debugging code that crashes, place a line like this somewhere in the code. If you get the error output, you know that the infinite loop is happening after the error--so move the error.throw later and try again.
  408  If it crashes, you know the infinite loop is earlier. Eventually, after a lot of heartache, you can zero in on the location.
  409  
  410: Here is a rogues' gallery of infinite loop gotchas--things that don't look like infinite loops, but they will kill your code quicker than you can wish you hadn't just pushed the enter key:
  411  code::

  431  
  432: Recursion is often used to walk through a tree structure. Tree structures are usually finite--no matter which branch you go down, eventually you will reach the end. If you have a data structure that is self-referential, you can easily get infinite recursion:
  433  code::

  435  a.put(5, a);	// now one of the items of a is a itself
  436: a.postcs;		// crash--postcs has to walk through the entire collection, which loops on itself
  437  ::

HelpSource/Guides/FFT-Overview.schelp:
  155  
  156: ⟪ 326 characters skipped ⟫operating directly on code::chainA::, then the two link::Classes/IFFT:: units would produce the same signal, which, when added together, would reinforce each other. Instead, the sound is nearly silent -- proving that code::chainB:: is in a different buffer, even though the function does not explicitly create it.
  157  

HelpSource/Guides/News-3_9.schelp:
  221  
  222: strong:: Breaking change: :: Rests in the patterns system have been restructured. Instead of using the code:: isRest :: event property, events are considered rests if one of their properties is a code:: Rest :: object. You must use instances of code:: Rest :: rather than the rest class itself -- use of code:: Rest :: instead of code:: Rest() :: is now deprecated.
  223  

HelpSource/Guides/News-3_11.schelp:
  131  
  132: STRONG:: Breaking change::: code:: Color:asHSV :: could sometimes return NaN -- grayscale colors returned NaN hue, and black returned NaN hue and saturation. Zero values are now returned in these cases, as is the standard (#4369)
  133  

HelpSource/Guides/Order-of-execution.schelp:
  158  (
  159: l = Bus.control(s, 1);  // get a bus for the LFO--not relevant to order-of-exec
  160: b = Bus.audio(s, 2);	// assuming stereo--this is to keep the src->fx chain separate from
  161  				// other similar chains

HelpSource/Guides/OSC_communication.schelp:
   89  ## Int8Arrays become "b" (blob).
   90: ## Chars are added as a raw type tag, appending no data to the OSC packet. This is needed to send code::$N:: (nil), code::$T:: (true), code::$F:: (false), code::$I:: (infinity/impulse). Be careful with this -- if the destination entity doesn't agree that the tag consumes zero bytes, the remaining part of the OSC message may be corrupted!
   91  ::

  111  
  112: If an unrecognized tag is encountered, sclang will make that unrecognized tag into a Char object and add that to the OSC message. It will then skip ahead in the OSC message as if it were reading a string -- it looks for the next null byte and then skips 0-3 bytes for 4-byte alignment.

HelpSource/Guides/ServerTiming.schelp:
  59  
  60: Notice that even though the left and right channel patterns were scheduled for exactly the same time, the events don't complete executing at the same time. Further, the Synth(...) call instructs the server to play the synth immediately on receipt, so the right channel will be lagging behind the left by about 2 ms each event--and not by the same amount each time. Timing, then, is always slightly imprecise.
  61  

HelpSource/Guides/Understanding-Errors.schelp:
  192  ::
  193: code:: #[0, 1, 2] :: is a literal array. Literal arrays cannot be manipulated--they can only be indexed. They cannot be changed internally.
  194  

HelpSource/Guides/UserFAQ.schelp:
  222  NOTE::
  223: If both are true, then the result will be 2, not 1. In some cases, the 2 may not be acceptable. That can be fixed by wrapping the Or in another comparison -- strong::((x > 0) + (y > 0)) > 0:: -- because 2 > 0 evaluates to 1!
  224  ::

  286  
  287: strong::Alternate solution::: For delay units, you may use preallocated delay buffers -- code::Buffer.alloc():: -- and the "Buf" delay units:
  288  code::BufDelayN::, code::BufDelayL::, code::BufDelayC::, code::BufCombL:: etc.

  300  
  301: - Supply a literal array -- code::\#[1, 2, 3]:: -- as the default for the argument name in the function.
  302  This is discussed in link::Classes/SynthDef::'s help file.

  330  
  331: To do steps \#1 and \#2, the SynthDef builder has to know the size of an array argument strong::before:: running the function. That's possible only if it's a literal array: code::\#[1, 2, 3, 4, 5]::. Any other array notation creates the array strong::while running the function:: (step \#3). But then it's too late -- the SynthDef builder already created a non-array control channel for it!
  332  

HelpSource/Guides/UsingMIDI.schelp:
  44  
  45: Any filters that are omitted will match all values -- e.g., for an omni-channel responder, simply leave out a code::chan:: filter.
  46  

HelpSource/Guides/WritingHelp.schelp:
  294      code::
  295:     git clone --recursive https://github.com/{your_username}/supercollider.git
  296      ::

HelpSource/Overviews/SymbolicNotations.schelp:
  113  ## code:: setA - setB :: || difference of sets (elements of setA not found in setB)
  114: ## code:: set -- set :: || symmetric difference:
  115  code::
  116: (setA -- setB) == ((setA - setB) | (setB - setA))
  117  ::

  132  
  133: a -- b
  134  Set[ 2, 9, 3, 4, 8 ]

  277  ## code:: [item, item...] :: || define an link::Classes/Array:: containing given items
  278: ## code:: Set[item, item...] :: || define a link::Classes/Set:: -- any link::Classes/Collection:: class name can be used other than Set
  279  ## code:: #[item, item...] :: || define a literal link::Classes/Array::

  321  ::
  322: ## code:: SomeClass.[index] :: || Equivalent to SomeClass.at(index) -- Instr.at is a good example
  323  ## code:: myObject.method(*array) :: || call the method with the arguments in an array

  361  ## code:: a[..j] :: || same as code:: a.copySeries(0, j) ::
  362: ## code:: a[j..] :: || same as code:: a.copySeries(i, a.size-1) :: (this is OK--Array is finite)
  363  

HelpSource/Reference/Assignment.schelp:
  55  ::
  56: -- or --
  57  code::

HelpSource/Reference/Literals.schelp:
  364  
  365: The outer function includes only the selector code::value::. The other selectors -- code::>::, code::*::, code::-:: -- belong to the inner function definition and don't affect the outer function's number of selectors.
  366  

HelpSource/Reference/Server-Command-Reference.schelp:
  109  table::
  110: ## -1 || turn off locally in the bundle -- error posting reverts to the "permanent" setting for the next message or bundle.
  111  ## -2 || turn on locally in the bundle.
  112  ::
  113: These "temporary" states accumulate within a single bundle -- so if you have nested calls to methods that use bundle-local error suppression, error posting remains off until all the layers have been unwrapped.
  114  If you use ['/error', -1] within a self-bundling method, you should always close it with ['/error', -2] so that subsequent bundled messages will take the correct error posting status.

HelpSource/Reference/ServerPluginAPI.schelp:
  236  ## code::sc_frac::
  237: ## code::sc_log2:: (legacy -- use code::std\::log2(std\::abs(x))::)
  238: ## code::sc_log10:: (legacy -- use code::std\::log10(std\::abs(x))::)
  239: ## code::sc_trunc:: (legacy -- use code::std\::trunc::)
  240  ::

  322  ## code::sc_sqrdif(a, b)::
  323: ## code::sc_atan2(a, b):: (legacy -- use code::std\::atan2::)
  324  ::

HelpSource/Tutorials/A-Practical-Guide/PG_01_Introduction.schelp:
   86  
   87: In SuperCollider, patterns are best for tasks that need to produce sequences, or streams, of information. Often these are numbers, but they don't have to be -- patterns can generate any kind of object.
   88  

  128  ## It's tested and it works. You don't have to debug how Pseries works (whereas, if you write a Routine, you might make a mistake and then have to find it.)
  129: ## With the Routine -- especially if it's complicated -- you will have to decipher it when you come back to the code later. The Pattern states the purpose right there in the code.
  130  ::

  151  
  152: This explains an easy "gotcha" with patterns -- forgetting to turn the pattern into a stream doesn't get the expected result. Since a pattern doesn't have any concept of a current state, calling code::next:: on it is meaningless, so code::next:: does what it does for most objects: return the receiver object itself. The method code::asStream:: creates the stream conforming to the pattern's specification, and calling code::next:: on the stream advances to its next state and returns the new value.
  153  

HelpSource/Tutorials/A-Practical-Guide/PG_02_Basic_Vocabulary.schelp:
   9  
  10: Some of the patterns will be demonstrated with a Pbind construct. This is a taste of things to come -- sequencing sonic events using patterns. Don't worry about how Pbind works just yet... all in good time.
  11  

  13  
  14: Again, the purpose is to start learning the vocabulary of patterns -- like learning new words when studying a human language. You can always come back and look at the rest later.
  15  

HelpSource/Tutorials/A-Practical-Guide/PG_03_What_Is_Pbind.schelp:
  176  
  177: To send only the relevant values to the new Synth, the Event needs to know what controls exist in the SynthDef. This is done by a library of descriptors for SynthDefs; the descriptor is a link::Classes/SynthDesc::, and the library is a link::Classes/SynthDescLib::. The normal methods -- code::.send(s)::, code::.load(s):: -- to communicate a SynthDef to the server do not enter it into the library. As a result, SynthDefs sent this way will not work properly with Pbind. Instead, use different methods that emphasis::store:: the SynthDef into the library.
  178  

  230  
  231: If it's the code::\freq:: key that determines whether the event as a rest or not, why does it work to use it with code::\note::? As noted, keys like code::\degree::, code::\note::, and code::\midinote:: are automatically converted into frequency. The math operations that perform the conversion preserve Symbols intact -- e.g., code::\rest + 1 == \rest:: . So the code::\rest:: value is passed all the way through the chain of conversions so that code::\freq:: in the event ends up receiving code::\rest::.
  232  

  268  	\freq, Pexprand(200, 800, inf),
  269: 	\trig, 1,	// note that this is NOT t_trig -- just \trig
  270  	\delta, 0.125

HelpSource/Tutorials/A-Practical-Guide/PG_04_Words_to_Phrases.schelp:
   34  
   35: But it gets even more fun -- list patterns don't care whether they're enclosing value patterns (as in the previous example) or event patterns. That means you can write a set of Pbind-style patterns, each one representing a phrase, and string them together. This next example is longer, but that's only because of a larger number of phrase patterns. The structure is very simple, though: code::Pxrand([Pbind(), Pmono(), Pmono()...], inf):: . Some of the phrases are written with Pmono to slide between notes.
   36  

  167  
  168: A complicated pattern with lots of embedding can be hard to read because it's more work to separate note-level details from the larger structure. The pattern choosing the phrases -- code::Pxrand(#[repeated, octave, tritone, dim], inf) :: -- is self-explanatory, however, and Psym fills in the details transparently.
  169  

HelpSource/Tutorials/A-Practical-Guide/PG_05_Math_on_Patterns.schelp:
   9  
  10: Some of these operations look like things you would do to an array, but there is a critical difference. Doing math on an array performs the operation on every array item all at once. By contrast, patterns are "lazy" -- they evaluate one value at the time, only when asked, and they only do as much as they need to do to deliver the next value. An operation on a pattern produces another pattern that remembers the work that is to be done. Making a stream out of the composite pattern creates the structure to perform the operation upon request.
  11  

HelpSource/Tutorials/A-Practical-Guide/PG_06b_Time_Based_Patterns.schelp:
   9  
  10: Most of these patterns work by remembering the clock's current time at the moment the pattern is embedded into a value stream. The time value used for calculation is, then, the clock's time at the moment of evaluation minus the starting time -- that is, the number of beats elapsed since the patterns started embedding. If the pattern is embedded several times, the starting time is also reset so that the pattern begins again from the beginning.
  11  

HelpSource/Tutorials/A-Practical-Guide/PG_06c_Composition_of_Patterns.schelp:
   35  
   36: // g · f == g(f(x)) -- f is evaluated first, and its result is passed to g
   37  g.value(f.value(1));

  113  
  114: A third group -- link::Classes/Psetp::, link::Classes/Paddp::, link::Classes/Pmulp:: -- behave slightly differently, nesting pattern evaluation.
  115  

HelpSource/Tutorials/A-Practical-Guide/PG_06d_Parallel_Patterns.schelp:
   9  
  10: The other is to combine them into a parallel stream. The result is a single pattern object that can be played or stopped only as one unit. Some degree of interactive control is lost, but there are times when merging several patterns is necessary -- for instance, converting a pattern into a Score object for NRT rendering.
  11  

  22  
  23: Ppar and its cousins are good for a fixed set of parallel patterns -- that is, you need to know in advance how many patterns will be parallelized. Once the parallel pattern starts, there is no way to add more streams to it. To keep adding streams, use link::Classes/Pspawner:: and link::Classes/Pspawn::. For the purpose of this overview, some basic features will be illustrated in a couple of simple examples. These classes have more capabilities; refer to their help files for specifics.
  24  

  61  		// Pfunc returns the pattern without rendering the stream
  62: 		// -- important for Pspawn
  63  		// See the Pspawn helpfile for other ways to embed patterns

HelpSource/Tutorials/A-Practical-Guide/PG_06g_Data_Sharing.schelp:
  55  
  56: ⟪ 306 characters skipped ⟫ment is used for every stream. In that case, the same environment could also be used in different Penvir patterns, and modifications of the environment by one Penvir would carry over to all the others -- hence its usefulness for sharing data.
  57  

HelpSource/Tutorials/A-Practical-Guide/PG_08_Event_Types_and_Parameters.schelp:
  203  
  204: All of these buffer event types expect the buffer number to be provided. They will not automatically get a buffer number from the server's buffer allocator. A Buffer object is allowed -- you could create the Buffer first using code::Buffer.alloc:: or code::Buffer.new:: and then use this object in the control events. See also link::#Event types with cleanup:: below for other, user-friendlier Buffer control options.
  205  

  449  ## receiver || The object to be modified.
  450: ## args || The list of variable names to set in the receiver. The receiver should have a setter method -- variableName_ -- for each of these. New values will be looked up in the event.
  451  ::

HelpSource/Tutorials/A-Practical-Guide/PG_Cookbook02_Manipulating_Patterns.schelp:
  76  
  77: What we can do instead is base the Pbind on strong::pattern proxies:: -- objects that take the place of a pattern. The link::Classes/PatternProxy:: is a single object that creates a single stream within Pbind, but it looks for its values to the pattern and stream contained inside the proxy. Changing the proxy's pattern replaces the stream, without having to touch the Pbind's closed box.
  78  

HelpSource/Tutorials/A-Practical-Guide/PG_Cookbook03_External_Control.schelp:
  119  
  120: Triggering a pattern based on audio amplitude is a bit trickier -- not because it's harder to play the pattern, but because identifying when the trigger should happen is more involved. The most straightforward way in SuperCollider is to use the link::Classes/Amplitude:: UGen to get the volume of the input signal and compare it to a threshold. Volume can fluctuate rapidly, so the code::releaseTime:: argument of Amplitude is set to a high value. This makes the measured amplitude fall more slowly toward the baseline, preventing triggers from being sent too close together.
  121  

HelpSource/Tutorials/A-Practical-Guide/PG_Cookbook07_Rhythmic_Variations.schelp:
   40  
   41: ⟪ 266 characters skipped ⟫his changes the data in the environment, which then get plugged into Pbind in the code::~pbindPairs.value():: line. That pattern will play through; when it ends, Plazy gives control back to its parent -- Pn, which repeats Plazy.
   42  

  184  		),
  185: 			// default Event checks \freq --
  186  			// if a symbol like \rest or even just \,

HelpSource/Tutorials/A-Practical-Guide/PG_Cookbook08_Swing.schelp:
  163  
  164: There's room also for a slight optimization. In the previous event, we decided whether the next event would need to swing or not. Now, in the current event, we are processing what used to be "next." So we can just copy the old value of teletype::nextShouldSwing:: from last time, instead of redoing the calculation. (Note that this requires code::nextShouldSwing = false:: in the beginning -- because teletype::now:: is always 0 for the first event, and consequently can never swing.)
  165  

HelpSource/Tutorials/A-Practical-Guide/PG_Ref01_Pattern_Internals.schelp:
   98  
   99: link::Classes/Routine::'s code::next:: method takes one argument, which is passed into the stream (Routine). The catch is that the routine doesn't start over from the beginning -- if it did, it would lose its unique advantage of remembering its position and resuming on demand. So it isn't sufficient to receive the argument using the routine function's argument variable.
  100  

  127  
  128: This convention -- receiving the first input value as an argument, and subsequent input values as a result of a method call -- holds true for the code::embedInStream:: method in patterns also. The rules are:
  129  

  208  		var hiVal, loVal;
  209: 			// length.value -- functions allowed for length
  210  			// e.g., Pwhite could give a random number of values for each embed

HelpSource/Tutorials/Getting-Started/14-Scheduling-Events.schelp:
   37  
   38: Note that we have moved to TempoClock, since this is the most commonly used. While there is only one SystemClock, there can be many TempoClocks all running at different speeds, if need be. One TempoClock is the default, accessed by code::TempoClock.default:: -- we will use this throughout. (To save typing, you may wish to assign a TempoClock to a variable, for instance, code::t = TempoClock.default ::.)
   39  

   57  
   58: Inside a scheduled function, you might want to know which clock is running the function. code::thisThread.clock:: tells you this -- don't worry for now about how it knows, just know that you can use this to find out.
   59  

   76  
   77: Nothing happens. That's because "hello" is just a value -- it doesn't do anything. The lesson is that it makes sense to schedule objects that will emphasis::take some action::.
   78  

  108  
  109: That shouldn't scare you off of scheduling -- this is less likely to happen with Routines and Tasks, which you will use more often. But you should be aware of it.
  110  

HelpSource/Tutorials/Getting-Started/15-Sequencing-with-Routines-and-Tasks.schelp:
   31  
   32: Recall that, when you schedule a function on a clock, numbers returned by the function are treated as time values -- specifically, the amount of time until the function should execute again. The same thing happens with numbers yielded by a routine.
   33  

   81  
   82: There you go -- our first sequence.
   83  

  195  unlike the previous infinite sequences, this one stops when it runs out of data.
  196: That's the purpose of the while loop -- it continues only as long as the 'dur'
  197  data stream keeps pumping out values. (See the

  199  
  200: There must be an easier way to write the data streams -- repeatedly writing the same do loop is certainly inconvenient. In fact, there is such a way, covered in the next tutorial: sequencing with patterns.
  201  

  224  
  225: The easiest way to add latency to your outgoing Synths is with the Server strong::makeBundle:: method. Don't worry about how it works for now -- the important thing is that it uses the first value for latency, and runs the messages produced by the function according to that latency.
  226  

HelpSource/Tutorials/Getting-Started/16-Sequencing-with-Patterns.schelp:
   31  
   32: Note that the Pseq is not streamable by itself, but it creates a stream (Routine) when you call asStream on it. This routine can then be used exactly like to any other routine -- the while loop used to read out the stream values is exactly the same for both, even though they are written differently.
   33  

   79  ## link::Classes/Pgeom::(start, grow, length) || geometric series, e.g., 1, 2, 4, 8, 16
   80: ## link::Classes/Pwhite::(lo, hi, length) || random number generator, uses rrand(lo, hi) -- equal distribution
   81: ## link::Classes/Pexprand::(lo, hi, length) || random number generator, uses exprand(lo, hi) -- exponential distribution
   82  ::

  125  	\instrument, \smooth,
  126: 		// MIDI note numbers -- converted automatically to Hz
  127  	\midinote, Pseq([60, 72, 71, 67, 69, 71, 72, 60, 69, 67], 1),
</p>
</details>

prko:
closes: #7148

## Purpose and Motivation
This PR aims to standardize dash usage:

-  En dashes mistakenly used as hyphens are replaced with -.
-  En dashes that should be em dashes are replaced with em dashes (—).
-  Legacy double hyphens (--) are replaced with em dashes (—).
-  Code lines containing method calls with -- remain unchanged.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


prko:
I believe I've reviewed the changed parts thoroughly, and there are no rendering errors caused by this update.  
However, the spacing around em dashes should be standardized, as it's currently inconsistent.  
Personally, I prefer having spaces before and after an em dash.  
However, some prefer no spaces—whether it's a single em dash or a pair enclosing a phrase—and even favor keeping the inner content tightly bound to the em dashes without any surrounding spaces.

HotwheelsSisyphus:
Looks good to me.
Re standardization: Technically we said American spelling/punctuation the style guide—I think—which entails no spaces around the dashes for the most part. _However_, I think this isn't super important and if the text isn't all prose (i.e., alternates with code) might look weird. I think it's best left alone, I don't think people will get into fights about this ;)

HotwheelsSisyphus (approved):

prko:
Thanks!

capital-G:
OT but maybe you can enjoy this article regarding the dash usage these days https://www.nytimes.com/2025/09/18/magazine/chatgpt-dash-hyphen-writing-communication.html?unlocked_article_code=1.qk8.xBPJ.Xby2jWiQ2AZU&smid=url-share - I certainly enjoyed it ;)

prko:
Thank you for sharing such an insightful article. I truly enjoyed the depth and clarity!

Having to manually type out verbose HTML entities like &emsp; or &#8198; is tedious and error-prone.

```

"&#x20;"   // " "	// space. // &#32;

"&hairsp;" // " "	// hair space. // &#8202;

"&#8239;"  // " "	// narrow no-break space.

"&ensp;"   // " "	// en space. // &#8194;

"&emsp;"   // " "	// em space. // &#8195;

"&#8198;"  // " "	// 6-per-em space.

"&emsp14;" // " "	// 4-per-em space. &#8197;

"&emsp13;" // " "	// 3-per-em space. &#8196;

"&thinsp;" // " "	// thin space. &#8201;

"&nbsp;"   // " "	// non breaking space. &#160;

"&puncsp;" // " "	// punctuation space. &#8200;

"&numsp;"  // " "	// figure space. &#8199;

"&#x3164;" // "ㅤ" // Hangul Filler
```
Introducing intuitive, memorable abbreviations for each space type would make word processors and **renderers like SCDocHTMLRenderer** far more user-friendly.

HotwheelsSisyphus:
> OT but maybe you can enjoy this article regarding the dash usage these days 

yeah, but it's also a specifically anglo problem. Ein guter teutscher tut sowas nicht, and the french non plus. (Their punctuation is the best imho.) I somehow got used to the dashes without spaces but still they're up there with the comma/period within quotation marks, as in "terrible," as the things I hate most about the CMOS rules.

dyfer:
Is there a reason this didn't get merged? I noticed this through a still open issue (#7148).

prko:
I do not know why this was not merged....
This PR seems to have been closed because I wrongly removed the corresponding part from my repository.

JordanHendersonMusic:
> am not sure why this PR was not merged, as merging PRs is not part of my responsibilities yet.

Some times it just gets missed. If its been approved but not merged, it is reasonable for the PR author to ping some one, or, what I like to do, is add it to the dev meeting list.

JordanHendersonMusic:
FYI, latex turns -- and --- into the en and em respectively.

prko:
Thanks!
>latex turns -- and --- into the en and em respectively.

This could be very easily done in `SCDocHTMLRenderer`, indeed!

PR #7151, opened 2025-09-25
Add NSMotionUsageDescription for AirPod positional tracking support

capital-G:
## Purpose and Motivation

Allows to access positional tracking data from connected AirPods on macOS via external UGens.
See https://github.com/capital-G/sc_airpods/

Without this PR the server crashes when the plugin gets loaded due to missing permissions.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/cmake_modules/MacOSXBundleInfo.plist.in
+++ b/cmake_modules/MacOSXBundleInfo.plist.in
@@ -151,6 +151,8 @@
 	<string>NSApplication</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>SuperCollider wants to use your microphone or sound input.</string>
+	<key>NSMotionUsageDescription</key>
+	<string>SuperCollider wants to use your airpods to use spatial audio.</string>
 	<key>NSServices</key>
 	<array>
 		<dict>


dyfer (approved):
Cool!

BTW I wonder if there's a way to check for the permissions when plugin loads and not call the APIs if there's no permission?

In a similar vein, when the server doesn't get microphone access, it should post a warning (it currently boots, but there's no sound input of course). I hope to add that when I get around to the coreaudio backend rewrite (we need to migrate to a newer coreaudio API eventually).

capital-G:
> BTW I wonder if there's a way to check for the permissions when plugin loads and not call the APIs if there's no permission?

That is an interesting question - the documentation https://developer.apple.com/documentation/coremotion/cmheadphonemotionmanager?language=objc#overview is IMO a bit unclear in this regard.
For me it is fine if the server crashes at this point b/c it is an external server extension and I think it is also better located outside of core-sc :)

> I hope to add that when I get around to the coreaudio backend rewrite (we need to migrate to a newer coreaudio API eventually).

Oh, good that you mention that - I am now somehow familiar w/ coreaudio/objective c and would have taken a look in regards to device switching, but if you want to do a full overhaul I'll wait w/ that - but I'll happily review any PR :)

dyfer:
> That is an interesting question - the documentation https://developer.apple.com/documentation/coremotion/cmheadphonemotionmanager?language=objc#overview is IMO a bit unclear in this regard. For me it is fine if the server crashes at this point b/c it is an external server extension and I think it is also better located outside of core-sc :)

Well, I agree this change is changing core-sc in relation to an "obscure" 3rd-party plugin, but if there's no better way to go about it, I think the change is innocuous enough to be included. I'll defer to your judgement on this.

> > I hope to add that when I get around to the coreaudio backend rewrite (we need to migrate to a newer coreaudio API eventually).
> 
> Oh, good that you mention that - I am now somehow familiar w/ coreaudio/objective c and would have taken a look in regards to device switching, but if you want to do a full overhaul I'll wait w/ that - but I'll happily review any PR :)

Realistically I'm not sure when I'll get to it though. May be months.
If you want to jump on it, let's have a quick chat, so that I can share what I already thought about in that regard.

PR #7155, opened 2025-09-29
SCLang: Fix bug with kwargs in primitives

### LINKED ISSUE #7152: sclang crashes in `IdentityDictionary.put`
The line `e.put(\cats, \birds, otherAnimals: \dogs)` (where `e` is an `IdentityDictionary`) crashes sclang v.3.14.0. The code, as well as the output, is given below.

Courtesy of Maxwell Miller who discovered this issue and provided this code.

## Environment

* SuperCollider version: 3.14.0
This also crashes on a friend's MacOS laptop (unknown sclang version).

## Steps to reproduce

```supercollider
e = IdentityDictionary.new()
e.put(\cats, \birds, otherAnimals: \dogs)
```

Output:
```
-> IdentityDictionary[]
Interpreter has crashed or stopped forcefully. [Exit code: -1073741819]
```

JordanHendersonMusic:
## Purpose and Motivation

Closes #7152

`().put(\asdf, 123, meow: 23)` Would crash because the stack pointer wasn't decremented when removing the kwargs from the stack, meaning the primitive was called with the stack pointer 5 from the end, but the primitive expected it to be 3. 

I do not know when this bug was introduced, as the changes made recently with the new kwargs shouldn't have affected this.

Two things should be considered in the future.

* Passing a stack pointer of the receiver to primitive, or checking the right number of arguments have been passed. This means rewriting all primitives, so probably isn't worth it, but would be safer in the long run.
* Refactoring the `doPrimitive...` functions because they are hard to work with right now.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -3616,16 +3616,15 @@ void doPrimitive(VMGlobals* g, PyrMethod* meth, int numArgsPushed) {
 }
 
 void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int numKeyArgsPushed) {
-    int i, j, m, diff, err;
-    PyrSlot *pslot, *qslot;
-    int numArgsNeeded, numArgsPushed;
-
+    const auto maybe_gc_sanitycheck = [&]() {
 #ifdef GC_SANITYCHECK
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
Closed in favour of #7166

PR #7160, opened 2025-09-30
classlib: NodeMap: move map to base class

telephon:
## Purpose and Motivation

NodeMap implements unmap, so we expact map, even if this is a synonym

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

Release Notes: NodeMap now implements map (synonym with set), instead of only ProxyNodeMap.

### DIFF
--- a/HelpSource/Classes/NodeMap.schelp
+++ b/HelpSource/Classes/NodeMap.schelp
@@ -20,6 +20,9 @@ set arguments of a node
 method::unset
 remove settings
 
+method::map
+map arguments of a node (synonymous with code::set::, mapping is done automatically when a suitable argument is passed).
+
 method::unmap
 remove mappings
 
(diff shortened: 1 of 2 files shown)


HelpSource/Classes/NodeMap.schelp
```diff
@@ -20,6 +20,9 @@ set arguments of a node
 method::unset
 remove settings
 
+method::map
+map arguments of a node (synonymous with code::set::, mapping is done automatically when a suitable argument is passed).
```
capital-G, line 24: Can you add a link to what "map" refers to in this case?

I don't even know if my suggestion would be correct here, so I think it is always good to provider further resources if we refer to other functionality.

```suggestion
LINK::Classes/Node.html#-map:: arguments of a LINK::Classes/Node:: (synonymous with LINK::Node.html#-set::, mapping is done automatically when a suitable argument is passed).
```

capital-G (approved):
Only a small remark regarding documentation - feel free to discard the comment and merge

JordanHendersonMusic:
Good to merge?

Issue #7106, opened 2025-08-18
Windows: possible issue with unixCmd calls


dyfer:
## Environment

* SuperCollider version: 3.15.0-dev
* Operating system: Windows
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
// not a real reproducer... an attempt to trace back the process
// add TestSuite to the compiled classes, then:
s.boot;
TestProcessCreation.run;
thisProcess.platform.killProcessByID(s.pid); // execute at the right moment...?
```

## Expected vs. actual behavior

I saw a suspicious error in the test runner on Windows in GHA and I want to file it as a separate issue for visibility.

This is an [output of one test run](https://github.com/supercollider/supercollider/actions/runs/17030428210/job/48272639042#step:6:13265):

```

	Suite: TestProcessCreation

...

	Test: test_string_unixCmd_quoteCmdAndArg
Server 'TestOSCBundle' exited with exit code 0.
WARNING: Server 'TestOSCBundle' failed to quit after 3.0 seconds.

FAIL: a TestProcessCreation: test_string_unixCmd_quoteCmdAndArg - quoted command with quoted argument evaluates successfully
Is:
	 nil 
Should be:
	 0


UNIT TESTS FOR 'TestProcessCreation' COMPLETED
There were failures:
a TestProcessCreation: test_string_unixCmd_quoteCmdAndArg - quoted command with quoted argument evaluates successfully
```
It seems that exiting server interfered with the execution of one of the `.unixCmd` tests... Is that even possible?

capital-G:
Is there something we can do about this? I'd really like to get this fixed or at least be removed from the test suite as the CI should always be *green* in order to clearly communicate any new bugs which are introduced via PRs.

Pinging @supercollider/windows

dyfer:
@capital-G this is beyond my expertise level, but I was also curious about fixing it and recently asked an LLM whether something in our [Windows file handling code](https://github.com/supercollider/supercollider/blob/dcaafe904b5832a4ec4e0d78dac6e1708aee117e/common/sc_popen.cpp#L133-L377) could be causing this problem.

It pointed out a few possible issues, including global reference counting for the processes, a possible race condition when adding processes to the process list, and possible issues with registering waiting for the handle. This is beyond my cpp expertise, but I'd love to fix it eventually. I can't dive into this for the time being though...

[...]

PR #7165, opened 2025-10-02
tests: disable process creation tests on windows

dyfer:
## Purpose and Motivation

Temporarily disable `TestProcessCreation -test_string_unixCmd_quoteCmdAndArg` on Windows so that the CI passes for PRs etc.

This does NOT fix the underlying issue, documented in #7106 

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/scripts/test_run_proto_gha.scd
+++ b/testsuite/scripts/test_run_proto_gha.scd
@@ -76,6 +76,12 @@
 		'skipReason': "UnitTest fails when run in GHA on Linux (FIXME)",
 	),
 	(
+		'suite': "TestProcessCreation",
+		'test': "test_string_unixCmd_quoteCmdAndArg",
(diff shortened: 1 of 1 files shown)


dyfer:
This approach for the workaround seems to be effective - the CI passed on multiple tries.

dyfer:
FWIW I'm not 100% sure this is a reliable workaround, as I thought that it is the exiting server process that interacts with one of the process creation tests (so disabling one shouldn't make a difference)... Yet, this branch ran successfully 4 times, between my branch and this PR, so maybe it will provide a viable workaround to make the CI pass more reliably.

capital-G (approved):
Ah - nice to see that we seem both prefer the same way :) And happy to see a green pipeline again <3

LGTM! Feel free to merge this yourself after considering the comment in https://github.com/supercollider/supercollider/issues/7106#issuecomment-3364862777

PR #7166, opened 2025-10-03
SCLang: Refactor primitives and keyword arguments

### LINKED ISSUE #7152: sclang crashes in `IdentityDictionary.put`

### LINKED ISSUE #7369: `asString`'s limit argument is almost never respected
## Environment

* SuperCollider version: all

## Steps to reproduce

```supercollider

\asdf.asString(3);

// doesn't matter what value you give it, always prints the same.
(1..1024).asString(30)

```

limit should be respected and the string capped.

### LINKED ISSUE #7488: keyword argument curiosity
Not sure what this should mean, but it's curious no?

```
sc3> -3.max(7)
-> 7
sc3> -3.max(7, 9)
-> 7
sc3> -3.max(7, 9, 7)
-> 7
sc3> -3.max(7, 9, aNumber: 7)
-> 9
```

Best, R

JordanHendersonMusic:
## Purpose and Motivation

From now on, primitives in c++ and sclang must have the same number of arguments, including variable arguments and variable keyword arguments, it is also a compile error to write a primitive that doesn't exist.

Previously, primitives could be declared with `n` arguments in c++ but `m` arguments in sclang. This was very confusing.

Previously, having a primitive with keyword arguments meant all keyword arguments were passed with the key to the primitive, even if they were declared as positional in sclang. This meant that
```
A {
   *new { |foo... args, kwargs|
          _PrintAllArgs
   }
}
A(foo: 10, bar: 20)
```
would print `[\foo, 10, \bar, 20]` rather than `[10, \bar, 20]`, which is unique behaviour.

Fixes bug `().put(\asdf, 123, meow: 23)` which would crash because the stack pointer wasn't decremented when removing them from the stack. I do not know when this bug was introduced.

Introduces a number of tests to ensure primitives will always be working correctly.

## Review guide

To start, I'd recommend having a look at the tests. 

## Difference

```
// All primitives just return the stack in an array.
TestPrimitive_Obj {
	noKwargsNoVarArgs { |a, b, c|
		_PrimitiveTestNoKwargsNoVarArg;
		^this.primitiveFailed
	}

	noKwargsWithVarArgs { |a, b, c... args|
		_PrimitiveTestNoKwargsWithVarArgs;
		^this.primitiveFailed
	}

	kwargsWithVarArgs { |a, b, c... args, kwargs|
		_PrimitiveTestWithKwargs;
		^this.primitiveFailed
	}
}

TestPrimitive_Obj().kwargsWithVarArgs(a: 1, b: 2, c: 3)
// old
 [a TestPrimitive_Obj, a, 1, b, 2, c, 3]
// new
 [a TestPrimitive_Obj, 1, 2, 3]

TestPrimitive_Obj().noKwargsWithVarArgs(1, 2, c: 3)
// old
 [a TestPrimitive_Obj, 1, 2, c, 3] 
// new
 [a TestPrimitive_Obj, 1, 2, 3] 
```

Previously, `definePrimitiveWithKey` was only meant to implement functions with variable keyword arguments and **no** positional arguments. If the sclang method had positional arguments, and they were specified as a keyword arguments, they would be placed on the end, now they are placed correctly. 

## Types of changes

- Bug fix
- Breaking change ?? Maybe, I am unsure.

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/GUI/Base/QView.sc
+++ b/SCClassLibrary/Common/GUI/Base/QView.sc
@@ -219,7 +219,7 @@ View : QObject {
 		^this.getProperty( \focus );
 	}
 
-	acceptsMouse {
+	acceptsMouse { 
 		_QWidget_AcceptsMouse
 		^this.primitiveFailed;
 	}
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -3618,14 +3618,9 @@ void doPrimitive(VMGlobals* g, PyrMethod* meth, int numArgsPushed) {
 }
 
 void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int numKeyArgsPushed) {
-    const auto maybe_gc_sanitycheck = [&]() {
 #ifdef GC_SANITYCHECK
-        g->gc->SanityCheck();
+    g->gc->SanityCheck();
 #endif
-    };
-
-    maybe_gc_sanitycheck();
-
     const int numNormalArgsGiven = allArgsPushed - (numKeyArgsPushed * 2);
 
     PyrMethodRaw* methraw = METHRAW(meth);
@@ -3634,15 +3629,15 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
     g->primitiveIndex = primIndex - def->base;
     g->primitiveMethod = meth;
 
-    PyrSlot* reciever = g->sp - allArgsPushed + 1;
+    PyrSlot* receiver = g->sp - allArgsPushed + 1;
 
     // Remove kwargs from stack to temporary stack.
     if (numKeyArgsPushed > 0) {
-        PyrSlot* first_kwarg = g->sp - (numKeyArgsPushed * 2) + 1;
+        PyrSlot* firstKwarg = g->sp - (numKeyArgsPushed * 2) + 1;
         for (size_t i { 0 }; i < numKeyArgsPushed * 2; i += 2) {
-            temporaryKeywordStack[i] = first_kwarg[i];
-            assert(first_kwarg[i].isSymbol());
-            temporaryKeywordStack[i + 1] = first_kwarg[i + 1];
+            temporaryKeywordStack[i] = firstKwarg[i];
+            assert(firstKwarg[i].isSymbol());
+            temporaryKeywordStack[i + 1] = firstKwarg[i + 1];
         }
         g->sp -= numKeyArgsPushed * 2;
     }
@@ -3665,9 +3660,8 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
             }
         } else {
             const unsigned int needed = numNeededArgs - numNormalArgsGiven;
-            const auto disFromSPToReciever = std::distance(g->sp, reciever);
             if (maybeReallocStack(g, needed)) {
-                reciever = g->sp + disFromSPToReciever;
+                receiver = g->sp + std::distance(g->sp, receiver);
             }
             PyrSlot* from = slotRawObject(&meth->prototypeFrame)->slots + numNormalArgsGiven - 1;
             PyrSlot* to = g->sp;
@@ -3679,7 +3673,7 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
         }
     }
 
-    int num_variable_kwargs = 0;
+    int numVariadicKwargs = 0;
     // Put keywords back on the stack, overriding what was already there on the stack.
     if (numKeyArgsPushed && methraw->totalNumberArguments) {
         PyrSymbol** argNames = slotRawSymbolArray(&meth->argNames)->symbols;
@@ -3690,7 +3684,7 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
             // We start at one here because you can't override 'this'.
             for (size_t argN { 1 }; argN < methraw->numNormalArguments; ++argN) {
                 if (key == argNames[argN]) {
-                    reciever[argN] = temporaryKeywordStack[keyword + 1];
+                    receiver[argN] = temporaryKeywordStack[keyword + 1];
                     goto found;
                 }
             }
@@ -3701,7 +3695,7 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
                          slotRawSymbol(&slotRawClass(&meth->ownerclass)->name)->name, slotRawSymbol(&meth->name)->name);
                 }
             } else {
-                num_variable_kwargs += 1;
+                numVariadicKwargs += 1;
                 g->sp += 1;
                 g->sp[0] = temporaryKeywordStack[keyword];
                 g->sp += 1;
@@ -3712,12 +3706,12 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
         }
     }
 
-    if (num_variable_kwargs) {
+    if (numVariadicKwargs) {
         g->numpop = numArgsOnStack - 1;
         g->gc->enterDelayedCollectionContext();
         int err;
         try {
-            err = ((PrimitiveWithKeysHandler)def[1].func)(g, numArgsOnStack, num_variable_kwargs);
+            err = ((PrimitiveWithKeysHandler)def[1].func)(g, numArgsOnStack, numVariadicKwargs);
         } catch (std::exception& ex) {
             g->lastExceptions[g->thread] = std::make_pair(std::current_exception(), meth);
             err = errException;
@@ -3759,6 +3753,7 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
             setupForMethod(g, meth, numArgsOnStack, 0);
         }
     }
+
 #ifdef GC_SANITYCHECK
     g->gc->SanityCheck();
 #endif
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -3660,8 +3660,10 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
             }
         } else {
             const unsigned int needed = numNeededArgs - numNormalArgsGiven;
+            // distance is cached here because we are about to reallocate the stack.
+            const auto disFromSPToReciever = std::distance(g->sp, receiver);
             if (maybeReallocStack(g, needed)) {
-                receiver = g->sp + std::distance(g->sp, receiver);
+                receiver = g->sp + disFromSPToReciever;
             }
             PyrSlot* from = slotRawObject(&meth->prototypeFrame)->slots + numNormalArgsGiven - 1;
             PyrSlot* to = g->sp;
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -18,6 +18,7 @@
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 */
 
+#include "PyrErrors.h"
 #include "PyrKernel.h"
 #include "PyrObject.h"
 #include "PyrPrimitive.h"
@@ -37,6 +38,7 @@
 #include "SC_AudioDevicePrim.hpp"
 #include "SC_LanguageConfig.hpp"
 #include "SC_Filesystem.hpp"
+#include "VMGlobals.h"
 
 #include <iterator>
 #include <map>
@@ -3748,17 +3750,72 @@ void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int n
         }
         g->gc->exitDelayedCollectionContext();
 
-    if (err <= errNone)
-        g->sp -= g->numpop;
-    else {
-        // post("primerr %d\n", err);
-        SetInt(&g->thread->primitiveIndex, methraw->specialIndex);
-        SetInt(&g->thread->primitiveError, err);
-        setupForMethod(g, meth, numArgsOnStack, 0);
+        if (err <= errNone)
+            g->sp -= g->numpop;
+        else {
+            // post("primerr %d\n", err);
+            SetInt(&g->thread->primitiveIndex, methraw->specialIndex);
+            SetInt(&g->thread->primitiveError, err);
+            setupForMethod(g, meth, numArgsOnStack, 0);
+        }
     }
 #ifdef GC_SANITYCHECK
     g->gc->SanityCheck();
 #endif
+}
+
+
+namespace sc::primitive_test {
+
+int noKwNoVarArgs(VMGlobals* g, int numArgs) {
+    if (numArgs != 4)
+        return errFailed;
+
+    PyrSlot* a = g->sp - 3;
+    PyrSlot* b = g->sp - 2;
+    PyrSlot* c = g->sp - 1;
+    PyrSlot* d = g->sp;
+
+    auto array = newPyrArray(g->gc, 4, 0, false);
+    array->slots[0] = *a;
+    array->slots[1] = *b;
+    array->slots[2] = *c;
+    array->slots[3] = *d;
+    array->size = 4;
+    *a = PyrSlot::make(array);
+    return errNone;
+}
+
+int noKwWithVarArgs(VMGlobals* g, int numArgs) {
+    if (numArgs < 4)
+        return errFailed;
+    auto array = newPyrArray(g->gc, numArgs, 0, false);
+    std::copy(g->sp - (numArgs - 1), g->sp + 1, array->slots);
+    array->size = numArgs;
+    *(g->sp - numArgs + 1) = PyrSlot::make(array);
+    return errNone;
+}
+
+int kwWithVarArgsSansKw(VMGlobals* g, int numArgs) {
+    if (numArgs < 4)
+        return errFailed;
+    auto array = newPyrArray(g->gc, numArgs, 0, false);
+    std::copy(g->sp - (numArgs - 1), g->sp + 1, array->slots);
+    array->size = numArgs;
+    *(g->sp - numArgs + 1) = PyrSlot::make(array);
+    return errNone;
+}
+
+int kwWithVarArgs(VMGlobals* g, int numArgs, int numKw) {
+    if (numArgs < 4)
+        return errFailed;
+    auto array = newPyrArray(g->gc, numArgs, 0, false);
+    std::copy(g->sp - (numArgs - 1), g->sp + 1, array->slots);
+    array->size = numArgs;
+    *(g->sp - numArgs + 1) = PyrSlot::make(array);
+    return errNone;
+}
+
 }
 
 void initPrimitives() {
@@ -3891,10 +3948,10 @@ void initPrimitives() {
     index = 0;
 
     // tests
-    definePrimitive(base, index++, "_PrimitiveTestNoKwargsNoVarArg", primitiveTests::noKwNoVarArgs, 4, 0);
-    definePrimitive(base, index++, "_PrimitiveTestNoKwargsWithVarArgs", primitiveTests::noKwWithVarArgs, 4, 1);
-    definePrimitiveWithVariableKeys(base, index, "_PrimitiveTestWithKwargs", primitiveTests::kwWithVarArgsSansKw,
-                                    primitiveTests::kwWithVarArgs, 4);
+    definePrimitive(base, index++, "_PrimitiveTestNoKwargsNoVarArg", sc::primitive_test::noKwNoVarArgs, 4, 0);
+    definePrimitive(base, index++, "_PrimitiveTestNoKwargsWithVarArgs", sc::primitive_test::noKwWithVarArgs, 4, 1);
+    definePrimitiveWithVariableKeys(base, index, "_PrimitiveTestWithKwargs", sc::primitive_test::kwWithVarArgsSansKw,
+                                    sc::primitive_test::kwWithVarArgs, 4);
     index += 2;
 
 
--- a/lang/LangPrimSource/PyrPrimitive.cpp
+++ b/lang/LangPrimSource/PyrPrimitive.cpp
@@ -22,18 +22,15 @@
 #include "PyrObject.h"
 #include "PyrPrimitive.h"
 #include "PyrPrimitiveProto.h"
-#include "PyrSignal.h"
 #include "PyrSignalPrim.h"
 #include "PyrMathPrim.h"
 #include "PyrListPrim.h"
-#include "Opcodes.h"
-#include "SC_InlineBinaryOp.h"
+#include "PyrSlot.h"
 #include "PyrMessage.h"
 #include "PyrParseNode.h"
 #include "PyrLexer.h"
 #include "PyrKernelProto.h"
 #include "PyrInterpreter.h"
-#include "PyrArchiverT.h"
 #include "PyrDeepCopier.h"
 #include "PyrDeepFreezer.h"
 #include "InitAlloc.h"
@@ -41,6 +38,7 @@
 #include "SC_LanguageConfig.hpp"
 #include "SC_Filesystem.hpp"
 
+#include <iterator>
 #include <map>
 #include <cstdlib>
 #include <cstring>
@@ -74,24 +72,8 @@ PyrSymbol* s_recvmsg;
 
 void initPatternPrimitives();
 
-typedef struct {
-    PrimitiveHandler func;
-    PyrSymbol* name;
-    unsigned short base;
-    unsigned char numArgs;
-    unsigned char varArgs;
-    unsigned char keyArgs;
-} PrimitiveDef;
 
-typedef struct {
-    int size, maxsize;
-    PrimitiveDef* table;
-} PrimitiveTable;
-
-extern PrimitiveTable gPrimitiveTable;
-
-
-int getPrimitiveNumArgs(int index) { return gPrimitiveTable.table[index].numArgs; }
+int getPrimitiveNumArgs(int index) { return gPrimitiveTable.table[index].numNormalArguments; }
 
 PyrSymbol* getPrimitiveName(int index) { return gPrimitiveTable.table[index].name; }
 
@@ -422,7 +404,7 @@ int prObjectString(struct VMGlobals* g, int numArgsPushed) {
     PyrString* string;
     char str[256];
 
-    a = g->sp;
+    a = g->sp - 1;
     if (IsSym(a)) {
         string = newPyrString(g->gc, slotRawSymbol(a)->name, 0, true);
         SetObject(a, string);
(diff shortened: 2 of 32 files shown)


telephon:
This fixes a lot of subtle interface inconsistencies. It would be interesting if the (explicitly noted) hacky parts would find a settlement. But it would be good if someone else could review this in detail.

JordanHendersonMusic:
> It would be interesting if the (explicitly noted) hacky parts would find a settlement.

Just to be clear on why this exist, the same code is being used in primitives and in the interpreter where the stack behaviour is different. There are some method that do something similar and pass a boolean around to mark this instead, so we should be able to do the same thing.

JordanHendersonMusic:
Closed #7152

dyfer:
From the functional standpoint, I think this is somewhat of an urgent fix and I'd lean towards releasing SC 3.14.1 with it. I'm curious to hear other opinions on this. ~~Also see https://github.com/supercollider/supercollider/issues/7152#issuecomment-3368555319 - this PR fixes interpreter crash for e.g. `true - 1`.~~ EDIT: this issue turned out to be a red herring. Please disregard.

Unfortunately I don't have in-depth knowledge to _meaningfully_ review this.

JordanHendersonMusic:
> `true - 1` 

Okay that is convincing. 

This pr makes it much harder for these issues with primitives to arise again, along with clearing up the relationship between SCLang method and c++ primitives.

JordanHendersonMusic:
Just thinking about this a bit more, if there are quarks out there that use primitives, and they don't have the right number of arguments, the class library will refuse to compile. Is that behaviour we wan't to introduce in a minor change?

Does anyone know of any quarks that use primitives? I think this should be unsafe behaviour as we (as the core language) should be the only ones calling these.

dyfer:
I don't think quarks should call primitives directly, so I don't think this would be an issue and it should be fine to release 3.14.1 with it.

muellmusik:
I would think that would be unsafe behaviour and very bad design. I’d say primitives should not be considered part of the public interface and should only be called via the appropriate method(s).

dyfer:
Yes... and if any issues arise, that would mean that the extensions need to be fixed/corrected so that they only use sclang calls.

JordanHendersonMusic:
Okay as long as we are happy with that position then I can't foresee any other issues

muellmusik:
I think it’s completely sensible. Probably worth documenting as a policy.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -1495,6 +1496,52 @@ void PyrMethodNode::compile(PyrSlot* result) {
     // set primitive
     // optimize common cases
 
+
+    if (mPrimitiveName) {
```
JordanHendersonMusic, line 1591: This is where the primitives are checked against the sclang code.

JordanHendersonMusic:
All the tests are passing now, thanks @dyfer, so this would be a good time to decide what to do with this.

dyfer:
Thanks @JordanHendersonMusic 
A whitespace at the end of two lines have snuck into `PyrInterpreter3.cpp`, can you fix that so that the CI runs?

I think this should be merged, but if possible I'd like someone who understands Cpp and the interpreter better than me to review this... @capital-G @scztt how can we move this forward?

JordanHendersonMusic:
Ah think it snuck in with a rebase!

capital-G:
Maybe someone can have a look at the tests and check if they make sense or if they cover everything described in the PR.
If the tests look fine, we decided in the dev meeting that this is good to be merged.

JordanHendersonMusic:
If I rebase this, do we think this is good to go?

lang/LangSource/PyrParseNode.cpp
```diff
@@ -1525,6 +1526,52 @@ void PyrMethodNode::compile(PyrSlot* result) {
...
+        auto prim = gPrimitiveTable.table[methraw->specialIndex];
+        if (prim.func == undefinedPrimitive) {
+            error("Method %s:%s has an undefined primitive.\n",
```
JordanHendersonMusic, line 1594: This case should be a warning in the next version, but an error in the following while we officially make it an error to reference a primitive outside of the class library.

capital-G:
Dev meeting: 
This is a breaking change in regards to primitives - if the method in sclang doesn't match its primitive it creates a compilation error. It also adds a compilation error when a primitive doesn't exist.
  Christof voiced concerns about making this a compile time error, prefering a runtime error for dynamic error catching - but b/c primitives should not be called outside the class library the compilation error is actually beneficial b/c it points out errors and provides an upgrade path.
  We should be LOUD about this change to the community - maybe we can check if every primitive is exposed via a class library function.
  Marcin: Maybe make it a warning in 3.15 and turn it off in 3.16?
  Decission: Primitive checking comes in place w/ 3.15 and a warning if primitive gets called from quark, with making this not possible in 3.16.
  Needs rebasing.

This can be merged by Jordan.

JordanHendersonMusic:
I've been looking at how to get an error posted when a class not in the class library uses a primitive and I don't think it is easily possible.

The issue is that you can point your language config to any directory and disable the default directory. This means we simply don't know what where the class library is, so can't switch on that. 

The other option would be to switch on the class name, but that would require writing a list of all the classes allowed to access primitives in the C++.

For normal uses, this really doesn't matter and would be fine, but being able to boot SC with a different class library is really useful for development and having to write the names of new core classes somewhere in the C++ would be a pain.


One possible solution would be to create a pragma syntax in the language that only the class library is allowed to use. This could be used useful in many places, but here it could be used to mark the class as allowing primitives.

```
T : Q ##[allow_primitives = true] {
}
```

Then we simply say the use of these pragmas is reserved for the class library, and should never be used in user code, therefore they never can allow primitives in their classes.

Either way, I think this is an extra feature and beyond this PR.


Right now, this PR produces a compile time error when the primitive is declared incorrectly. This is a breaking change, but a good one, because user (and class library code) was previously broken.

One example of this is `(1...1024).asString(32)` which never passed the limit argument on, now it does. It was also pretty easy to get segfaults and weird stack issues.

dyfer:
> I've been looking at how to get an error posted when a class not in the class library uses a primitive and I don't think it is easily possible.

I don't think this is what we need. I think the discussion was on whether we should prevent the class library from compiling already in 3.15 if there's a mismatch between the class and the primitive, or post warning in 3.15 and error out in 3.16+?

JordanHendersonMusic:
> if there's a mismatch between the class and the primitive,

This will probably lead to a segfault though, or just weird behaviour, it is a serious bug.

`(1...1024).asString(32)` this `32` isn't passed to the function for this reason. 

Things are also often fine when you all a primitive normally, but when you use keywords, it can just suddenly break.

Why would you want this, the whole project could just crash mid-performance and all you did was pass an argument or call a method with keywords rather than positional arguments! 

Having a hard error here is of course annoying, but points to a serious flaw that could bring down your live show. If it is possible to ignore, people will.

After the dev meeting my impression is that the warning was for people using primitives outside of the core class library (in quarks and their own classes). This shouldn't be allowed, but previously it we had no opinion on this, so it was fair game. That is what I thought we should introduce slowly. 

I think we should reserve these softer transitory warnings for when we are redefining parts of the core language, not fixing bugs. Otherwise we are going to end up with many of these.

dyfer:
> Having a hard error here is of course annoying, but points to a serious flaw that could bring down your live show. If it is possible to ignore, people will.

Yes, that's fair.

> After the dev meeting my impression is that the warning was for people using primitives outside of the core class library (in quarks and their own classes). This shouldn't be allowed, but previously it we had no opinion on this, so it was fair game. That is what I thought we should introduce slowly.

Got it. I probably misremembered then, sorry.
Yeah, I don't see a clean logical way to prevent user extensions from accessing primitives directly. Documentation is probably the best we can do for now. 

> I think we should reserve these softer transitory warnings for when we are redefining parts of the core language, not fixing bugs. Otherwise we are going to end up with many of these.

Sounds good. Sorry for the noise!

JordanHendersonMusic:
@dyfer I guess we could add a field to the config file in the future were you separately mark where your class library is, but that is a future issue!

It would be great to get a proper review, but if not I think we agreed this is in principle good to go?

SCClassLibrary/Common/GUI/Base/QView.sc
```diff
@@ -219,7 +219,7 @@ View : QObject {
 		^this.getProperty( \focus );
 	}
 
-	acceptsMouse {
+	acceptsMouse {
```
dyfer: Extra whitespace

```suggestion
	acceptsMouse {
```

capital-G:
Just in case you missed it @JordanHendersonMusic https://github.com/supercollider/supercollider/actions/runs/22972106064/job/66691364110?pr=7166#step:13:23

```
4/23 Test  #4: supernova__callback_interpreter_system_run ....***Failed    0.00 sec
Running 3 test cases...
/home/runner/work/supercollider/supercollider/testsuite/server/supernova/callback_interpreter_system.cpp(46): fatal error: in "callback_interpreter_threadpool_test": critical check i == 41 has failed [40 != 41]

*** 1 failure is detected in the test module "Master Test Suite"


      Start  5: supernova__memory-pool_test_run
 5/23 Test  #5: supernova__memory-pool_test_run ...............
```

JordanHendersonMusic:
That's supernova, it occasionally breaks for some reason?

JordanHendersonMusic:
What's the status with this? I can't remember what we agreed about in the dev meeting, but this should be merged, it's just a matter of whether I've overlooked something in the code. 

I can just merge it if no one wants to review it?

capital-G:
I am also a bit confused about its state :D Looking at https://github.com/supercollider/supercollider/wiki/Development-meetings%3A-minutes i found

## 2026-02-02
> https://github.com/supercollider/supercollider/pull/7166 : This is a breaking change in regards to primitives - if the method in sclang doesn't match its primitive it creates a compilation error. It also adds a compilation error when a primitive doesn't exist. Christof voiced concerns about making this a compile time error, prefering a runtime error for dynamic error catching - but b/c primitives should not be called outside the class library the compilation error is actually beneficial b/c it points out errors and provides an upgrade path. We should be LOUD about this change to the community - maybe we can check if every primitive is exposed via a class library function. Marcin: Maybe make it a warning in 3.15 and turn it off in 3.16? Decission: Primitive checking comes in place w/ 3.15 and a warning if primitive gets called from quark, with making this not possible in 3.16. Needs rebasing. This can be merged by Jordan.

## 2026-03-11

> https://github.com/supercollider/supercollider/pull/7166 - Christof will review and approve this PR, there are some minor things

So maybe @Spacechild1 can have a look over this for a review and give a thumbs up or request changes?

Spacechild1:
Just a heads-up: I won't be available for any code reviews before April 21. After that, I will happily take a look. If someone else wants to step in, feel welcome!

JordanHendersonMusic:
@Spacechild1 of course no worries! 

The biggest change here is 'contractual' , which I think we all agree on. Technically this could change the behaviour of code, but only broken code, it's just that people might not have known it was broken.

The c++ is lots of small fixes, and a renaming of the methraw fields because they no longer made sense. 

One questionable thing to note is that the signal binary ops don't work with adverbs, but are required to be declared with an adverb argument to match the primitive Def. This is a bit odd actually and I might revisit this....

JordanHendersonMusic:
@Spacechild1 I don't suppose you have a moment to take a look at this one? I think there was just posted another issue that this changes.

lang/LangPrimSource/PyrPrimitive.cpp
```diff
@@ -3620,29 +3618,106 @@ void doPrimitive(VMGlobals* g, PyrMethod* meth, int numArgsPushed) {
...
+                numArgsOnStack = numNeededArgs;
+            } else {
+                usedNormalVarArgs = true;
+                numArgsOnStack = numNormalArgsGiven;
+            }
+        } else {
+            const unsigned int needed = numNeededArgs - numNormalArgsGiven;
+            const auto disFromSPToReciever = std::distance(g->sp, reciever);
+            if (maybeReallocStack(g, needed)) {
+                reciever = g->sp + disFromSPToReciever;
```
Spacechild1: typo in `reciever`.

Just a small thing, but do we need the `disFromSPToReciever` variable? I find it pretty verbose and it threw me off a bit while reading. It's only used once and I don't think it adds any information since it just describes what's already obvious from the code. (Not to mention that it should be declared inside the if-clause where it's actually used.) I think the following would be more succint:
```cpp
receiver = g->sp + std::distance(g->sp, reciever);
```

lang/LangPrimSource/PyrPrimitive.cpp
```diff
@@ -3620,29 +3618,106 @@ void doPrimitive(VMGlobals* g, PyrMethod* meth, int numArgsPushed) {
...
     g->primitiveIndex = primIndex - def->base;
     g->primitiveMethod = meth;
 
-    if (def->keyArgs && numKeyArgsPushed) {
-        g->numpop = allArgsPushed - 1;
+    PyrSlot* reciever = g->sp - allArgsPushed + 1;
+
+    // Remove kwargs from stack to temporary stack.
+    if (numKeyArgsPushed > 0) {
+        PyrSlot* first_kwarg = g->sp - (numKeyArgsPushed * 2) + 1;
```
Spacechild1: `first_kwarg` should be camel case.

lang/LangPrimSource/PyrPrimitive.cpp
```diff
@@ -3620,29 +3618,106 @@ void doPrimitive(VMGlobals* g, PyrMethod* meth, int numArgsPushed) {
...
+            PyrSlot* to = g->sp;
+            for (size_t i { 0 }; i < needed; ++i) {
+                to[i + 1] = from[i];
+            }
+            g->sp += needed;
+            numArgsOnStack = numNeededArgs;
+        }
+    }
+
+    int num_variable_kwargs = 0;
```
Spacechild1: Should be camel case.

lang/LangPrimSource/PyrPrimitive.cpp
```diff
@@ -3620,29 +3618,106 @@ void doPrimitive(VMGlobals* g, PyrMethod* meth, int numArgsPushed) {
 }
 
 void doPrimitiveWithKeys(VMGlobals* g, PyrMethod* meth, int allArgsPushed, int numKeyArgsPushed) {
-    int i, j, m, diff, err;
-    PyrSlot *pslot, *qslot;
-    int numArgsNeeded, numArgsPushed;
-
+    const auto maybe_gc_sanitycheck = [&]() {
```
Spacechild1: Should be camel case.

Also, do we really need this? AFAICT this lambda is only called once, namely right after it's declared.

Actually, it *could* be used in L 3741 again, but honestly I would just repeat the `#if` clause.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -3996,206 +4039,179 @@ void PyrBlockNode::compile(PyrSlot* slotResult) {
...
-        for (j = 0; j < numArgs; ++j) {
-            if (blockargs[j] == slotRawSymbol(varslot)) {
-                error("Function argument '%s' already declared in %s:%s\n", slotRawSymbol(varslot)->name,
-                      slotRawSymbol(&gCompilingClass->name)->name, slotRawSymbol(&gCompilingMethod->name)->name);
-                nodePostErrorLine((PyrParseNode*)vardef);
-                compileErrors++;
-            }
+    // Ensure no variable or argument in this scope collides.
+    std::vector<PyrSymbol*> encounteredNames {};
+    encounteredNames.reserve(numSlotsForProtoFrame);
```
Spacechild1, line 4078: Since the sclang compiler/interpreter is not reentrant anyway, we *could* make this `static` to avoid repeated heap allocations and deallocations. In that case we wouldn't need to reserve, instead the vector would keep its capacity:
```cpp
static std::vector<PyrSymbol*> encounteredNames {};
encounteredNames.clear();
``` 

Alternatively, we could use `boost::small_vector` (https://www.boost.org/doc/libs/1_68_0/doc/html/boost/container/small_vector.html) to utilize the stack for small vector sizes.

This is probably just a micro-optimization, but it's something that immediately jumped into my head.

Spacechild1 (commented):
Thanks for the PR! As you know, I'm not too familiar with the sclang internals, but it all makes sense on a broad level and I trust you with the details :)

The C++ side looks fine. I just have a few comments/suggestions, mostly stylistic.

Sorry it took so long! Too busy with other stuff...

lang/LangPrimSource/PyrPrimitive.cpp
JordanHendersonMusic: I've already commit a lot of code in snake, I really don't think this matters, the standard is in snake.

lang/LangSource/PyrParseNode.cpp
JordanHendersonMusic: > Since the sclang compiler/interpreter is not reentrant...

At the moment... Literally working on this right now!

I think it's fine as is, this doesn't need to be fast, particularly if we can multithread it!

Spacechild1: > At the moment... Literally working on this right now!

Cool!

> I think it's fine as is, this doesn't need to be fast, particularly if we can multithread it!

Ok. It's also something that can easily be optimized later, if it ever shows up un profiling (which it probably won't)

lang/LangPrimSource/PyrPrimitive.cpp
Spacechild1: > I've already commit a lot of code in snake, 

That's unfortunate, but not an argument for committing more ;-)

> I really don't think this matters, 

Then why do we have code style guides?

> the standard is in snake.

This doesn't matter, we should follow the style of the project. Supernova uses snake case for variables/members, scsynth and sclang use camel case. In fact, you have been using camel case yourself for most other variables/members, that's why I called out these specific instances. This is about consistency.

capital-G: I think it would be great if we would have some "guidance" here through clang-tidy - see https://github.com/supercollider/supercollider/issues/7497

JordanHendersonMusic: Also there are things like `pyr_compile_pool` that have existed since the initial commit. Sclang uses both. Its a mess,  but it doesn't affect readability so I don't think it matters.

> Then why do we have code style guides?

because formatting/indentation does affect how you read the code, and there are many more options. There are less for names, snake, camel/Pascal, and the terrifying SCREAMING_SNAKE, you need to be comfortable reading all of those to write c++. That why I don't think this matters. 

The other reason is that I've spent about a week carefully refactoring almost the whole compiler in preparation for moving the parser into a static library and I've used snake everywhere and I do not have time to change it.

Spacechild1: I think consistency matters, otherwise it gradually becomes a big mess over time. There's simply no reason for mixing camel case and snake case for variables inside the same function. I'm only asking for these 3-4 variables, that's literally a single search+replace operation.

JordanHendersonMusic: Okay that's fine I'll do that. But I'm about to push a big pr that uses snake everywhere, it's actually quite useful to be able to tell which function is old based on the name. I'm slowly rewriting the whole compile so it's quite useful to be able to see what bits haven't been done yet. Eventually the whole thing will be snake and pulled out into static libraries.

I really don't think names create a mess, indentation and things that alter how you read control flow, sure, but not the style the name is written in.

Spacechild1: > Okay that's fine I'll do that. 

Thanks!

> Eventually the whole thing will be snake and pulled out into static libraries.

IMO using snake case for new separate modules is less of a problem, as long it is consistent in itself. My concern is foremost about mixing styles within a module, let alone within the same function.

> I really don't think names create a mess, indentation and things that alter how you read control flow, sure, but not the style the name is written in.

Maybe for you, but I find it quite jarring to read functions where the local variables have different casing for no apparent reason. I don't see why you would treat case style differently from, say, brace style or indentation.

I should also point out that in sclang and scsynth, underscores are generally used to separate a prefix from a variable/function name, e.g. `class_<var>`, `obj_<enum>`, `World_<function>`, `Graph_<function>`, etc. Using actual snake case for variables goes against that convention and undermines the meaning of the underscore.

I general, I always try to emulate the style of the project I'm working on. Stylistic consistency is always more important than my own personal preferences.

(In the same vein, I wouldn't write `for (size_t i { 0 }; ...` in a code base that otherwise does `for (size_t i = 0; ...`, but I intentionally didn't bring that up in my review in fear of sounding overly pedantic :)

JordanHendersonMusic: > I general, I always try to emulate the style of the project I'm working on. Stylistic consistency is always more important than my own personal preferences.

> My concern is foremost about mixing styles within a module

 `pyr_compiler_pool` (a global), the rest of the globals are declared like `gCamel`, then class methods are both camel and pascal (`SymbolTable::MakeNew`), and free functions which are both camel, all lower case `getsym`, and pascal `DumpSimpleBackTrace`, variables are usually in camel, but there are exceptions like `AllocPool::Alloc::candiate_size`.

> I should also point out that in sclang and scsynth, underscores are generally used to separate a prefix from a variable/function name

Sort of. `class_sequenceable_collection`, its all snake, if that where true it would be `class_sequencableCollection`.

Then there is the issue of declaring all variables at the top, mutating everything, global variables everywhere...

The lang directory is a huge mess. Hence why I don't think this matters, but I'll change these, however there is no consistency nor 'normal' in this part of the code base.

I'm rewriting the whole thing and pulling everything I can out into separate static libraries. These libraries will be consistent and use Pascal for types and snake for everything else. Then the stuff left in the langSource dir is basically just there to glue to all together. 


Re braceinitalisation. yeah its overkill. but initialisation is super complicated in C++ and brace is always the safest option so I'd rather just be consistent. Emulating a code base that was written in c++98 in c++17 seems unnecessary.

Spacechild1: > I'm rewriting the whole thing and pulling everything I can out into separate static libraries. These libraries will be consistent and use Pascal for types and snake for everything else. Then the stuff left in the langSource dir is basically just there to glue to all together.

This sounds good!

> Emulating a code base that was written in c++98 in c++17 seems unnecessary.

C++17 doesn't mean you can't use simple assignments in numeric for-loops. I think that's still the most idiomatic way and there is absolutely no safety issue. But again, it's about consistency. If the existing code always used brace initialization in for-loops, I would do as well.

> but initialisation is super complicated in C++ and brace is always the safest option

Not always. It wouldn't be C++ without stupid pitfalls like this:
```cpp
std::vector<int> vec(4); // -> [0, 0, 0, 0]
std::vector<int> vec{4}; // -> [4]
```

JordanHendersonMusic: > std::vector<int> vec{4}; // -> [4]

That is exactly what I'd expect a vector to do though, so for me, it is always safer to do the list/brace initialisation. Hence why I just do it everywhere now, one less thing to think about.

> I think that's still the most idiomatic way and there is absolutely no safety issue.

If you know exactly what type it is or it is a literal.

```
if (std::size_t i = max(); i >=0; --i )
```
```
if (std::size_t i{max()}; i >=0; --i )
```
If `max()` returns an `int` of `-1` you get promotion to `2^64-1`, expect in the second case where it fails to compile.

Sure its an edge case, but you don't have to think about with brace initalisation.

Another option would be auto.

Spacechild1: This is getting pretty off-topic, but why not? :-)

> That is exactly what I'd expect a vector to do though, so for me, 

Really? Why? In the general case,  `A a{4}` is supposed to be synonymous with `A a(4)`, which means: "pass the number 4 to the constructor of `A`". There are already two other completely valid syntax options for constructing a `std::vector` with an initializer list:
```cpp
std::vector<int> vec{{ 4 }}; // -> pass { 4 } to the constructor
std::vector<int> vec = { 4 };
```
The fact that `std::vector<int> vec{4}` calls the initializer list overload is just mind-boggingly stupid because it means that you cannot possibly call the other constructors with the (supposedly) universal brace initialization syntax.

My point is that universal brace initialization is not really universal. If applied mechanically, it can cause compiler errors in the best case and silently change behavior in the worst case.

JordanHendersonMusic: I guess I've never wanted to create a vector with a size, I've always found reserve to be more intuitive. 

The promotion to initialiser list is dumb, but it's the only constructor I ever use, so practically it doesn't bother me - it's actually useful because is no ambiguity here.

Spacechild1: >  I've always found reserve to be more intuitive.

I guess it depends on the use case. When you treat a `std::vector` as a collection of elements, you typically start with an empty vector, possibly `reserve` the space and then push elements. However, if you use `std::vector` as a data buffer (soundfile, image, histogram, etc.) you often start with a specific size and default value (often 0). In latter case, the size/value constructor comes handy.

JordanHendersonMusic: > In latter case, the size/value constructor comes handy.

I'd use assign for that.

The other reason I like braces is because it always does the same constructor and somethings in the standard have so many I get fed up of looking then up!

I think the problem is that constructors are overused. I generally use statics in my code when there is any chance of ambiguity, I don't know why the standard doesn't do that as much.

So yes, that's why I use braces everywhere, it's what I know, and I write less bugs with it, and so I've gotten in to the habit of just using it everywhere because for me there are no down sides.

There are some places where I'd want others to use it more: aggregate init, notably with unions, which is a c++ 20 feature I miss; but also for the narrowing. But I wouldn't try and convince anyone assigning from a literal is bad or anything.

Spacechild1: Fair enough. 

> I'd use assign for that.

I see. That's certainly possible. I prefer
```cpp
std::vector<char> image(width * height, 0);
```
over
```
std::vector<char> image;
image.assign(width * height, 0);
```
but that's certainly a matter of taste.

JordanHendersonMusic: I'm so paranoid about unitialised variables I'd actually do ...
```cpp
std::vector<char> i{};
i.assign(...);
```
I know this doesn't matter with vector, but for other things you never know.

Thanks for reviewing this by the way!

JordanHendersonMusic:
Thanks for reviewing this @Spacechild1! I've made all the changes you have requested.

lang/LangPrimSource/PyrPrimitive.cpp
```diff
@@ -3620,29 +3618,100 @@ void doPrimitive(VMGlobals* g, PyrMethod* meth, int numArgsPushed) {
...
+                g->sp -= dif;
+                numArgsOnStack = numNeededArgs;
+            } else {
+                usedNormalVarArgs = true;
+                numArgsOnStack = numNormalArgsGiven;
+            }
+        } else {
+            const unsigned int needed = numNeededArgs - numNormalArgsGiven;
+            if (maybeReallocStack(g, needed)) {
+                receiver = g->sp + std::distance(g->sp, receiver);
```
Spacechild1: Fortunately, I looked at my comments again. I thought: "wait a second, isn't this line really a no-op?"

Then I looked at your original code and finally understood what it was doing: it cached the distance between `g->sp` and `receiver` *before* potentially reallocating the stack so that it can fixup the `receiver` pointer.

So my suggestion actually broke the code! 🤦‍♂️

Please restore your original code, but add a comment, just to make sure that nobody will make the same mistake again. I'm really sorry for that!

JordanHendersonMusic: Good catch! I'd obviously forgotten it did that too!

Spacechild1: Once you have reverted my bad suggestion (and added the comment), this PR should be ready for merging!

Spacechild1 (approved):

Spacechild1:
IMO this is good to go now. Feel free to merge!

JordanHendersonMusic:
@Spacechild1 Thank you!

Spacechild1:
Thank *you*!

PR #7167, opened 2025-10-04
tests: use actual booleans in the proto file

dyfer:
## Purpose and Motivation

This changes the test proto file to use actual booleans, as suggested in https://github.com/supercollider/supercollider/issues/7106#issuecomment-3364862777

With this change, [this part of the runner](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/sclang_test_runner.scd#L117-L121) could be simplified, should I also adjust that? It works as is...

## Types of changes

- Cleanup

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation


JordanHendersonMusic:
Can you see any reason why they were strings in the first place?

dyfer:
> Can you see any reason why they were strings in the first place?

This was originally a yaml, not an scd file, so IIUC everything had to be a string, interpreted back to sclang.

JordanHendersonMusic:
In that case, I think you should change the runner too.

capital-G:
Thanks for looking into this dyfer :)

> This was originally a yaml, not an scd file, so IIUC everything had to be a string, interpreted back to sclang.

But [yaml supports booleans](https://yaml.org/type/bool.html)^^

> With this change, [this part of the runner](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/sclang_test_runner.scd#L117-L121) could be simplified, should I also adjust that? It works as is...

I think it would be cool if we just apply `.value` on it instead of transferring this to a string. This would allow dynamic (de-)activation of tests, so e.g.

```supercollider
// skip: {thisProcess.platform.name == \osx}
// skip: true

// fallback to false if value is nil
t[\skip].value ?? {false}
```

For me it would be fine if we adjust these lines of the test runner script in this PR as well.

dyfer:
> > This was originally a yaml, not an scd file, so IIUC everything had to be a string, interpreted back to sclang.
> 
> But [yaml supports booleans](https://yaml.org/type/bool.html)^^

Okay, it seems it was me then who screwed up when converting the ~~yaml~~ json file, sorry... :) [The ~~yaml~~ json file indeed had booleans and not strings](https://github.com/supercollider/supercollider/blob/Version-3.13.0/testsuite/scripts/gha_test_run_proto.json#L10).

> I think it would be cool if we just apply `.value` on it instead of transferring this to a string. 

We won't even need `.value` I think? Boolean is already a dictionary entry, though I need to test this. I'll change it.

> This would allow dynamic (de-)activation of tests, so e.g.

Hm, I'm not sure I want to include a default value for skipping? Maybe this is something to consider when the testing script is reworked? Unless you think it's really beneficial.

EDIT: I also have #7052 open, which touches the proto file. I think I'd like to merge that one first and then resolve conflicts in this PR.

capital-G:
> We won't even need `.value` I think? Boolean is already a dictionary entry, though I need to test this. I'll change it.

Evaluating the value with `.value` would allow us to either write down boolean primitives (`false.value` evaluates to `false`), but also to evaluate dynamic functions (`{thisProcess.platform.name == \osx}.value` evaluates to `true`, but only on macos). In the current setup this would allow us to turn off the unix test for windows in an easy manner :)

> Hm, I'm not sure I want to include a default value for skipping?

There already exists a fallback value - currently we return `false` if `t[\skip]` is nil or not set to `"true"`.

https://github.com/supercollider/supercollider/blob/29ca22cb411d4831758cacdac3704882d0569817/testsuite/scripts/sclang_test_runner.scd#L117-L121

IMO it is probably best to be somehow explicit if a test should be skipped, so defaulting to non-skipping seems fine.

dyfer:
> IMO it is probably best to be somehow explicit if a test should be skipped, so defaulting to non-skipping seems fine.

Makes sense, I agree. I'll make the changes.

dyfer:
This PR is now rebased and includes a change to the test runner file.

capital-G (approved):
Awesome :)

Issue #7088, opened 2025-07-28
Scale degree literals in Literals.schelp


HotwheelsSisyphus:
## Environment

* * SuperCollider version: 3.14

## Expected vs. actual behavior

The literals helpfile currently says the following:

> Integer numbers as scale degrees supports accidentals notation by adding the suffixes s for sharp and b for flat. Accidentals are represented as floating point values.
> ```
> 2s == 2.1  // scale degree two, sharp
> 2b == 1.9  // scale degree two, flat
> 2ss == 2.2 // scale degree two, double sharp
> 2bb == 1.8 // scale degree two, double flat
> ```

However, there doesn't seem to be anywhere this is actually usable. I would assume that it should work in Scale.degreeToFreq, for instance:

```supercollider
Scale.major.degreeToFreq(4 - 1, rootFreq: 12.midicps, octave: 4.0).cpsmidi.round(1) 
// -> 65 (scale degree 4 in C major: F)
Scale.major.degreeToFreq(4.1 - 1, rootFreq: 14.midicps, octave: 4.0).cpsmidi.round(1) 
// -> 65  ... but this should be 66, i.e. sharpened scale degree 4, i.e. F#. 
```
The reason it doesn't work there is the wrapAt In common/collections/scale.sc:

```
	degreeToFreq { |degree, rootFreq, octave|
		^this.degreeToRatio(degree, octave) * rootFreq;
	} 

	degreeToRatio { |degree, octave = 0|
		octave = octave + (degree div: degrees.size);
		^this.ratios.wrapAt(degree) * (this.octaveRatio ** octave);
	}
```
The accidental notation suggested in the Literals helpfile lets you notate stuff like the raised scale degree 3 through something like 2s, which is represented by Float 2.1. But that float gets truncated when passed to wrapAt. (This is actually an even bigger problem with lowered scale degrees, where 2b is represented as 1.9 and thus wrapAt is even further off).

## Solution: 

(Unless somebody knows a place in the code where this feature is actually used)

Either
1. Put a Note or Warning in Literals.schelp that this was never really implemented.
or
2. Implement it. Somewhat not so straightforward if one doesn't presuppose 12TET. (Think of harpsichords with split black keys, etc.)... 
3. Or just implement it for 12TET (and say so in the help), which would be something like this:

```
degreeToRatio { |degree, octave = 0| 
                var degDiatonic = degree.round(1);
                var accidentals = (degree - degDiatonic) * 10;
[...]

PR #7168, opened 2025-10-05
Remove scale degrees subsection from Literals.schelp

tedmoore:
## Purpose and Motivation

Removed scale degrees section and examples from Literals.schelp because, apparently, this functionality was never implemented. See Issue #7088 , which this PR would also close

## Types of changes

- Documentation

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

Should close Issue #7088

### DIFF
--- a/HelpSource/Reference/Literals.schelp
+++ b/HelpSource/Reference/Literals.schelp
@@ -93,39 +93,6 @@ code::
 
 Hexidecimal numbers notated with code::0x:: may only be expressed as integers.
 
-subsection::Scale Degrees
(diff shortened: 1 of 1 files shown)


jamshark70:
I believe this PR should be closed, for reasons explained in a comment under #7088 -- https://github.com/supercollider/supercollider/issues/7088#issuecomment-3369690878

capital-G:
There was a consensus at the dev meeting to not delete the documentation but instead add a warning to it that it may not work as intended.

Jordan has some ideas on how this feature could be re-modeled for actual usage, but this would be something for another PR.

PR #7169, opened 2025-10-06
SCDocHTMLRenderer: Insert a Space at Each Line Break

### LINKED ISSUE #7001: SCDoc: Line breaks after tag does not result in a whitespace
## Environment

* SuperCollider version: 3.15-dev
* Operating system: any
* Other details (Qt version, audio driver, etc.): -

## Steps to reproduce

```
detail. If set to code::UnitTest.brief::, and link::#*reportPasses::
is true, passes will be reported, but without any extra details which
```

w/o any whitespace at the end of the first line [results in](https://docs.supercollider.online/Classes/UnitTest.html#*passVerbosity)

Image

See https://github.com/supercollider/supercollider/pull/7000 for an example where this occurs (and has been fixed)

A whitespace should be handled implicitly as a whitespace here like elsewhere in SCDoc.

This could either be an issue with the `SCDocHTMLRenderer` class or within the parser of SCDoc.

prko:
fixes #7001
## Purpose and Motivation

The current implementation of SCDocHTMLRenderer does not insert a space when a new line begins, as outlined in issue #7001. This pull request addresses and resolves that problem.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/SCDoc/SCDocRenderer.sc
+++ b/SCClassLibrary/SCDoc/SCDocRenderer.sc
@@ -563,7 +563,7 @@ SCDocHTMLRenderer {
 				};
 				this.renderChildren(stream, node);
 			},
-			\NL, { }, // these shouldn't be here..
+			\NL, { stream << " "; }, // these shouldn't be here..
 // Plain text and modal tags
 			\TEXT, {
 				stream << this.escapeSpecialChars(node.text);


prko:
This implementation introduces a new issue:  
In the following code snippet, I deliberately broke the line before the closing parenthesis ) to avoid unwanted indentation:

https://github.com/supercollider/supercollider/blob/29ca22cb411d4831758cacdac3704882d0569817/HelpSource/Help.schelp#L41-L42

However, with the implementation in this pull request, a whitespace is now inserted before the ), which is undesirable.

I believe the root of the problem lies in the fact that sclang-style indentation is being applied to schelp documentation. To work around this, I personally resort to using alternative punctuation and typographic symbols such as:

•  Quotation marks: “ ” and ‘ ’
•  Brackets: ❲ ❳ and ❨ ❩
•  Apostrophe and possessive marker: ʼ

However, inputting these non-standard symbols is rather cumbersome—not only within the IDE, but across most software environments.

PR #7175, opened 2025-10-07
plugins: add BufIn to read buffer samples as flat array (like FluidBufToKr)

elgiano:
## Purpose and Motivation

When users want to read Buffer data as a control stream, they need to create as many BufRd UGens as the number of samples they want to read. This pattern is common for example in FluCoMa, where a buffer often contains multi-channel analysis data, and FluCoMa indeed provides the `FluidBufToKr` pseudo-UGen precisely to pull this data out of a buffer and make it available as a control stream for other non-buffer-reading UGens. Having a dedicated UGen saves the overhead of having to spawn, say, 1024 BufRd UGens.

This is a draft, I just wrote the basic functionality for now:
- only KR
- simple test (in a new file)
- drafted documentation

I'm happy to receive comments about pretty much everything, from the name to how to write the inner loop, the docs, or in which file to place the tests.

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/Classes/BufIn.schelp
+++ b/HelpSource/Classes/BufIn.schelp
@@ -1,16 +1,24 @@
 class:: BufIn
 summary:: Reads Buffer samples as flat array
-related:: Classes/BufRd
+related:: Classes/BufRd, Classes/Index
 categories:: UGens>Buffer
 
 Description::
+Outputs the contents of a buffer as a control signal, reading a number of
+consecutive samples as separate channels. Like link::Classes/Index::, but
+for multiple indices as the same time, saving the overhead of making multiple UGens.
 
-Outputs the contents of a buffer as a control signal, reading adjacent samples
-as separate channels.
-
-In comparison to link::Classes/BufRd:: :
+In comparison to link::Classes/BufRd:::
 BufRd has a single reading head that moves through buffer samples, BufIn has
-multiple heads, each staying still on a sample.
+multiple heads, each staying still on a different sample.
+
+BufIn treats multi-channel buffers as a flat, interleaved sequence of values.
+This means that reading for instance 2 samples of a 2-channels buffer will
+output only the first value for each channel. See link::#examples:: below, and
+link::Classes/Buffer#Multichannel Buffers:: for more info.
+
+Only the link::#*kr:: method is implemented, because the value at a single buffer index can't
+change between control periods.
 
 classmethods::
 private:: categories
@@ -21,13 +29,16 @@ Number of output channels, i.e. number of consecutive samples to read from the
 buffer.
 note::
 It doesn't have to match the Buffer's number of samples. If the
-buffer has less samples, the remaining output channels will return 0.::
+buffer has less samples than requested, the exceeding output channels will return 0.::
 
 argument::bufnum
 The index of the buffer to use.
 
 argument::offset
 The sample index to start reading from.
+
+returns::
+An array of output channels, corresponding to the requested buffer samples.
 
 instancemethods::
 private:: init, argNamesInputsOffset, checkInputs
@@ -36,10 +47,8 @@ Examples::
 
 code::
 
-(
 // load some numbers on a buffer
 b = Buffer.loadCollection(s, [34, 61, 104.3, 50]);
-)
 
 // print those numbers
 x = { BufIn.kr(4, b).poll }.play;
@@ -58,5 +67,21 @@ x.free;
 b.loadCollection({rrand(40,100)}!4);
 
 b.free
+::
 
+Values from multi-channel buffers are interleaved. Note that they need to be
+interleaved manually when using link::Classes/Buffer#*loadCollection::
+
+code::
+b = Buffer.loadCollection(s, [[1,2,3], [4,5,6]].lace);
+
+x = { BufIn.kr(6, b).poll }.play;
+// outputs [1, 4, 2, 5, 3, 6]
+// i.e. [ch0_0, ch1_0, ch0_1, ch1_1, ch0_2, ch1_2]
+
+x.free;
+
+// user can de-interleave manually afterwards
+x = { BufIn.kr(6, b).unlace(2).poll }.play;
+// outputs [1, 2, 3, 4, 5, 6]
 ::
--- a/HelpSource/Classes/BufIn.schelp
+++ b/HelpSource/Classes/BufIn.schelp
@@ -17,12 +17,9 @@ This means that reading for instance 2 samples of a 2-channels buffer will
 output only the first value for each channel. See link::#examples:: below, and
 link::Classes/Buffer#Multichannel Buffers:: for more info.
 
-Only the link::#*kr:: method is implemented, because the value at a single buffer index can't
-change between control periods.
-
 classmethods::
 private:: categories
-method:: kr
+method:: ar, kr
 
 argument::numChannels
 Number of output channels, i.e. number of consecutive samples to read from the
--- a/HelpSource/Classes/BufIn.schelp
+++ b/HelpSource/Classes/BufIn.schelp
@@ -12,16 +12,18 @@ In comparison to link::Classes/BufRd:::
 BufRd has a single reading head that moves through buffer samples, BufIn has
 multiple heads, each staying still on a different sample.
 
-BufIn treats multi-channel buffers as a flat, interleaved sequence of values.
-This means that reading for instance 2 samples of a 2-channels buffer will
-output only the first value for each channel. See link::#examples:: below, and
-link::Classes/Buffer#Multichannel Buffers:: for more info.
+For multi-channel buffers, BufIn defaults to reading a single channel, but can
+also read all of them as a flat, interleaved sequence of values, by setting the
+channel argument to -1. Note that the provided number of outputs has to take
+this into account, e.g. to read 2 frames of a 2-channel buffer,
+teletype::numOutputs:: has to be set to 4.
+See link::#examples:: below, and link::Classes/Buffer#Multichannel Buffers:: for more info.
 
 classmethods::
 private:: categories
 method:: ar, kr
 
-argument::numChannels
+argument::numOutputs
 Number of output channels, i.e. number of consecutive samples to read from the
 buffer.
 note::
@@ -31,8 +33,13 @@ buffer has less samples than requested, the exceeding output channels will retur
 argument::bufnum
 The index of the buffer to use.
 
+argument::channel
+Select which channel to read, or teletype::-1:: for reading all of them as an
+interleaved flat series (see link::examples:: below).
+
 argument::offset
-The sample index to start reading from.
+The sample index to start reading from. Exceeding the buffer's number of
+channels will results in zeros being returned.
 
 returns::
 An array of output channels, corresponding to the requested buffer samples.
@@ -66,13 +73,44 @@ b.loadCollection({rrand(40,100)}!4);
 b.free
 ::
 
-Values from multi-channel buffers are interleaved. Note that they need to be
-interleaved manually when using link::Classes/Buffer#*loadCollection::
+Reading a single channel from a multi-channel buffer:
+code::
+b = Buffer.loadCollection(s, [[1,2,3], [4,5,6]].lace);
+
+x = { BufIn.kr(3, b, channel: 0).poll }.play;
+// outputs [1, 2, 3]
+x.free;
+
+x = { BufIn.kr(3, b, channel: 1).poll }.play;
+// outputs [4, 5, 6]
+x.free;
+
+// multichannel expansion
+x = { BufIn.kr(3, b, channel: [0,1]).poll }.play;
+// outputs [[1, 2, 3], [4, 5, 6]]
+x.free;
+
+// note that 'channel' is modulatable
+(
+x = { 
+	var ch = ToggleFF.kr(Impulse.kr(1)).poll(label:'channel')).poll;
+	BufIn.kr(3, b, channel: ch)
+}.play;
+)
+
+x.free;
+
+b.free;
+::
+
+To read all channels at the same time, set teletype::`channel: -1`::. The output
+will be a flat array with all channels interleaved.
+Note that they need to be interleaved manually when using link::Classes/Buffer#*loadCollection::
 
 code::
 b = Buffer.loadCollection(s, [[1,2,3], [4,5,6]].lace);
 
-x = { BufIn.kr(6, b).poll }.play;
+x = { BufIn.kr(6, b, channel: -1).poll }.play;
 // outputs [1, 4, 2, 5, 3, 6]
 // i.e. [ch0_0, ch1_0, ch0_1, ch1_1, ch0_2, ch1_2]
 
@@ -81,4 +119,6 @@ x.free;
 // user can de-interleave manually afterwards
 x = { BufIn.kr(6, b).unlace(2).poll }.play;
 // outputs [1, 2, 3, 4, 5, 6]
+
+b.free
 ::
--- a/HelpSource/Classes/BufIn.schelp
+++ b/HelpSource/Classes/BufIn.schelp
@@ -1,0 +1,62 @@
+class:: BufIn
+summary:: Reads Buffer samples as flat array
+related:: Classes/BufRd
+categories:: UGens>Buffer
+
+Description::
+
+Outputs the contents of a buffer as a control signal, reading adjacent samples
+as separate channels.
+
+In comparison to link::Classes/BufRd:: :
+BufRd has a single reading head that moves through buffer samples, BufIn has
+multiple heads, each staying still on a sample.
+
+classmethods::
+private:: categories
+method:: kr
+
+argument::numChannels
+Number of output channels, i.e. number of consecutive samples to read from the
+buffer.
+note::
+It doesn't have to match the Buffer's number of samples. If the
+buffer has less samples, the remaining output channels will return 0.::
+
+argument::bufnum
+The index of the buffer to use.
+
+argument::offset
+The sample index to start reading from.
+
+instancemethods::
+private:: init, argNamesInputsOffset, checkInputs
+
+Examples::
+
+code::
+
+(
+// load some numbers on a buffer
+b = Buffer.loadCollection(s, [34, 61, 104.3, 50]);
+)
+
+// print those numbers
+x = { BufIn.kr(4, b).poll }.play;
+
+x.free;
+
+// use as control signals for other UGens
+(
+{ 
+	var notes = BufIn.kr(4, b);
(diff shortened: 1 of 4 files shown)


prko:
I understand that FluCoMa is a separate project and not officially part of the SuperCollider core. However, since it already addresses the overhead of spawning multiple BufRd UGens for control-rate access, I wonder whether its functionality could be leveraged more directly.  
Of course, I realise that licensing, build complexity, and long-term maintenance are important factors when considering external integration. Still, I’m curious whether adopting or referencing FluCoMa’s approach might offer a more efficient path forward, especially for users working with large buffers or multi-channel analysis data.

SCClassLibrary/Common/Audio/BufIO.sc
```diff
@@ -111,6 +111,23 @@ RecordBuf : UGen {
 }
 
 
+BufIn : MultiOutUGen {
+	*kr { arg numChannels, bufnum=0, offset=0;
```
capital-G: nitpick, can we use the *new* syntax here?


```suggestion
	*kr { | numChannels, bufnum=0, offset=0 |
```

server/plugins/DelayUGens.cpp
```diff
@@ -1600,6 +1609,37 @@ void RecordBuf_next_10(RecordBuf* unit, int inNumSamples) {
...
+    unit->m_fbufnum = -1e9f;
+    unit->m_failedBufNum = -1e9f;
+
+    SETCALC(BufIn_next);
+    BufIn_next(unit, 1);
+}
+
+void BufIn_next(BufIn* unit, int inNumSamples) {
+    GET_BUF_SHARED
+    uint32 numOutputs = unit->mNumOutputs;
```
capital-G: A more general question: does this actually copy the value of the variable to the stack or is the compiler *smart enough* to replace the occasions with the reference?

If so, wouldn't it be more efficient to store this value in the struct within the constructor in this case? IIUC it is not modulatable.

And since it is not modulatable, we can also make it `const` ;)

server/plugins/DelayUGens.cpp
```diff
@@ -1600,6 +1609,37 @@ void RecordBuf_next_10(RecordBuf* unit, int inNumSamples) {
...
+    uint32 offset = static_cast<uint32>(ZIN0(1));
```
capital-G: ~~uint32 seems correct here but could create some nasty buffer underflows if the input signal goes below 0.
Maybe we should guard this with `.max(0)`?~~

[static_cast seems to clamp any negative int -> uint values](https://cpp.sh/?source=%2F%2F+Example+program%0A%23include+%3Ciostream%3E%0A%23include+%3Cstring%3E%0A%0Aint+main()%0A%7B%0A++std%3A%3Astring+name%3B%0A++std%3A%3Acout+%3C%3C+%22What+is+your+name%3F+%22%3B%0A++getline+(std%3A%3Acin%2C+name)%3B%0A++std%3A%3Acout+%3C%3C+%22Hello%2C+%22+%3C%3C+name+%3C%3C+%22!%5Cn%22%3B%0A%7D) - so safe to ignore this comment or is this compiler dependent? Couldn't find anything on cppreference.com

server/plugins/DelayUGens.cpp
```diff
@@ -1600,6 +1609,37 @@ void RecordBuf_next_10(RecordBuf* unit, int inNumSamples) {
...
+        if (unit->mWorld->mVerbosity > -1 && !unit->mDone && (unit->m_failedBufNum != fbufnum)) {
+            Print("Buffer UGen: no buffer data\n");
+            unit->m_failedBufNum = fbufnum;
+        }
+        ClearUnitOutputs(unit, inNumSamples);
+        return;
+    }
+
+    // we could wrap around buf samples, but no other ugen does it
+    if (offset < 0)
```
capital-G: offset is uint, so it could not go below 0

SCClassLibrary/Common/Audio/BufIO.sc
```diff
@@ -111,6 +111,23 @@ RecordBuf : UGen {
...
+		^this.multiNew('control', numChannels, bufnum, offset)
+	}
+
+	init { arg argNumChannels ... theInputs;
```
capital-G, line 123: Same

```suggestion
	init { | argNumChannels ... theInputs |
```

server/plugins/DelayUGens.cpp
```diff
@@ -1600,6 +1609,37 @@ void RecordBuf_next_10(RecordBuf* unit, int inNumSamples) {
...
+        offset = 0;
+
+    for (size_t ch = offset; ch < numOutputs; ch++) {
+        OUT(ch)[0] = bufData[ch];
```
capital-G: what happens when the buffer has not enough samples?

capital-G (changes requested):
Thanks @elgiano - this is a really nice feature.

Could you

* add an `.ar` version for this as well? I am sure that there are also people who want to profit from the performance gains for some fancy buffer modulation :)
* Add tests for "garbage" values, so e.g. negative offset, overflowing the number of channels compared to the number of samples within the buffer
* What should happen if e.g. offset=-2 with 4 channels? Should we mask the non-existing channels with silence or post an error?

SCClassLibrary/Common/Audio/BufIO.sc
elgiano: nitpick: I also prefer the *new* syntax but I used the old one here because of the rest of this file. Maybe we change them all together one day?

server/plugins/DelayUGens.cpp
elgiano: I'm not sure but I think making it const will trigger the optimization...

server/plugins/DelayUGens.cpp
elgiano: I made it an int, it was actually creating a problem down the line

server/plugins/DelayUGens.cpp
elgiano: the ugen outputs zero. We could also wrap, it would make sense with other buf reading ugens, but it would make it more complicated and I don't know if there's a use case for it

SCClassLibrary/Common/Audio/BufIO.sc
capital-G: valid point - consistency is probably better here.

elgiano:
@prko I think it's a common enough usage pattern to deserve this optimization, even though I can't think of other examples for reading buffers as large multichannel outputs than analysis data and thus FluCoMa, but perhaps PV_MagBuffer.  Elsewhere in SC we use Index or BufRd, but this creates a lot of UGens when reading large numbers of channels. I see this as a similar optimization to what Klang is for SinOsc.

@capital-G thanks for looking at it! 
- I had more or less the same thoughts about tests this morning and I added some.
- After your comment I added also an `ar` method, even though I'm pretty sure a UGen can't see buffer values changing within a control block (the changer must have finished running earlier in the chain). But at least is a convenience function for rate conversions.
- I don't know if we should loop like BufRd: it would add a little more complexity and I don't see a use case for it. For now I opted for the simplest solution but I could change it if we decide otherwise.

dyfer:
> * What should happen if e.g. offset=-2 with 4 channels? Should we mask the non-existing channels with silence or post an error?

I vote for unobstructed reading of random memory addresses! :D j/k

dyfer:
> However, since it already addresses the overhead of spawning multiple BufRd UGens for control-rate access,

@prko, FluCoMa does _not_ address the overhead, but this ugen does. If this is pulled in, FluCoMa should probably eventually transition to using this ugen instead of its less efficient (but the only one currently possible) approach.

dyfer:
@elgiano this is great!

I wonder about the name. One association I have with "...In" is the it could be an "input" to a buffer... which this is not. (And I also wonder about a companion plugin, would it be worth to create a ugen that takes an array of channels and writes to individual buffer samples?)

What do you think about `BufRdArr` or something similar, stemming from `BufRd`? `BufRdArr` is not as elegant as `BufIn`, but IMO the "In" part is ambiguous.

tedmoore:
> ugen that takes an array of channels and writes to individual buffer samples?

<3<3 In FluCoMa, this is the `FluidKrToBuf`

> I wonder about the name.

I'm not sure that `BufToKr` and `KrToBuf` are the best names, but since it's, let's say, fixing, the FluCoMa implementation maybe `BufToStream.kr`, `BufToStream.ar`, `StreamToBuf.kr`, and `StreamTuBuf.ar` clarifies the relationship?

//==============

I wonder if a user accidentally puts in an audio buf with one channel and 48000 samples (or many many many more than that). This would create a stream 48000 channels wide... The FluCoMa UGen doesn't let the user pass a buffer with more than 1000 samples. (However, in my own practice, I remove this limitation...). Maybe there needs to be a way to watch out for this which would probably cause a long hang in trying to compile the synthdef?

HelpSource/Classes/BufIn.schelp
```diff
@@ -0,0 +1,84 @@
...
+Description::
+Outputs the contents of a buffer as a control signal, reading a number of
+consecutive samples as separate channels. Like link::Classes/Index::, but
+for multiple indices as the same time, saving the overhead of making multiple UGens.
+
+In comparison to link::Classes/BufRd:::
+BufRd has a single reading head that moves through buffer samples, BufIn has
+multiple heads, each staying still on a different sample.
+
+BufIn treats multi-channel buffers as a flat, interleaved sequence of values.
```
tedmoore: Rather than treating the buffer as flat and interleaving, I think what might be more general-purpose flexible is always output a single channel and have a modulatable argument that can specify which channel to read. I realize this is basically a matrix transposed version of BufRd, but it means one could stack up a bunch of data points for example prepped to send to the fluid mlp and then just index between them on the server. Even more flexible would be if floats passed to the modulatable channel argument would interpolate between the channels.

elgiano:
> I vote for unobstructed reading of random memory addresses! :D j/k

I once made a plugin for that: [elgiano/NasalDemons](https://github.com/elgiano/nasaldemons)

Back to topic:

> I wonder about the name. One association I have with "...In" is the it could be an "input" to a buffer.

> maybe BufToStream.kr, BufToStream.ar, StreamToBuf.kr, and StreamTuBuf.ar clarifies the relationship


My reasoning was that like In.ar reads from a Bus to a "stream", BufIn.ar reads from a Buffer to a "stream".
I think the word "stream" is problematic though, because in SC it's always related to Patterns.
"Signal" would also have the problem of referring to the Signal object... but maybe it's better than stream in this regard, there are many instances in the docs for audio, control or output signals.

So, perhaps `BufToSignal` / `SignalToBuf`?
Or, out of the blue, `BufToArray`? Somehow seeing `BufToSignal` doesn't make me think much of what this UGen does, its kind of "static" behavior, but maybe it's just me?

elgiano:
> This would create a stream 48000 channels wide
With this UGen you need to specify channels manually, it's not automatically dependent on how many frames the buffer has. If a user writes 48000, it will likely fail when the SynthDef is built for lack of interconnected buffers, perhaps with a long hang, but the user needs to have written that `48000`, so I think it's fine. One can't even use `BufSamples.ir(buf)` here, because the synthdef needs to know at compile time how many channels to allocate.

There is still the case where one writes `b.numFrames`, making a buffer-specific synthDef, but that's an edge case? We could add a check in sclang for not exceeding wireBufs, I'm looking into it just out of curiosity. Currently s.options.numWireBufs is 64 on my system, but I can still go up to 610 channels, so I need to dig a little more.

tedmoore:
> With this UGen you need to specify channels manually, it's not automatically dependent on how many frames the buffer has.

I like this.

> There is still the case where one writes `b.numFrames`, making a buffer-specific synthDef, but that's an edge case?

Can't save everyone all the time. Per above, I think it's probably good enough having the user required to specify a number makes it so the interface is raising awareness of what is happening. It makes sure the user is saying what they want.

> We could add a check in sclang for not exceeding wireBufs, I'm looking into it just out of curiosity.

Could do. Not necessary in my opinion. If a user get's the "not enough wireBufs" error they can increase it. I do find that a lot of users don't know what this error means (and what the wire bufs are), so it would be very lovely and polite if this object threw an error to increase the general knowledge of how this stuff works and how to fix this error, but I don't know that that's you're responsibility right here.

dyfer:
> > There is still the case where one writes `b.numFrames`, making a buffer-specific synthDef, but that's an edge case?
> 
> Can't save everyone all the time. Per above, I think it's probably good enough having the user required to specify a number makes it so the interface is raising awareness of what is happening. It makes sure the user is saying what they want.

I'd like to be able to both specify the number of frames read ( -> resulting channels), or have it propagated from the buffer. In the latter case one case just use BufFrames.kr, so it's just up to the new plugin to accept this value (.ir only I guess, because it needs to be fixed for the duration of the synth... OR is that not feasible?)


> Or, out of the blue, `BufToArray`?

I like `BufToArray`. I'd argue against "Signal", as this is also a language-side data representation.
While we are not discussing an inverse ugen, I think the name would do well to be "reversible" (e.g. BufToArray vs ArrayToBuf) if an inverse ugen would be considered later. 

> My reasoning was that like In.ar reads from a Bus to a "stream", BufIn.ar reads from a Buffer to a "stream".

Yes, that makes sense... but if we were to consider the opposite direction, that would be BufOut, which also doesn't read very clear to me. I think `BufToArray` is the closest so far to me for clarity, though BufIn is more elegant and shorter.

tedmoore:
> (.ir only I guess, because it needs to be fixed for the duration of the synth... OR is that not feasible?)

I think it's not possible because the width of the resulting kr stream needs to be known when the synthdef is compiled not just when initialized (which is when .ir reports).

elgiano:
Maybe BufToArray is the most straightforward after all.

For me BufOut works though: Out writes to a bus, ScopeOut to a scope, BufOut to a buffer. But perhaps this is something that makes sense only because of these internal references among these ugens...

I'm fine with both, have nothing against BufToArray other than its length. But clarity is of course more important.

P.S. ted is right, we can't use BufFrames.ir in there because we need to know the number of channels when compiling the synthdef, like BufRd

dietcv:
I would like if the Ugen name keeps a reference to BufRd / BufWr. Maybe something like BufDataRd / BufDataWr.
I think wrap makes no sense when wanting to read all the selected data via offset at once (wrap always has this time dependency imo), clip should be fine.

HelpSource/Classes/BufIn.schelp
dietcv: I guess the interpolation can be implemented with the cubicinterp function from SC_SndBuf.h. (i have already used that for my graindelay ugen).

tedmoore:
What about BufChannelRd which would correspond to BufChannelWr. It is reading all (or some subset of) the samples in a channel.

HelpSource/Classes/BufIn.schelp
elgiano: I thought I had replied to this but can't find the reply anymore...
I think interpolation can be a subsequent stage, or even another UGen, like what `IndexL` is to `Index`. Regarding channels, I'm thinking: let's add a `channel` argument to specify a single channel to be extracted, but that can also be `-1` (can't send `nil` to server), to get the interleaved flat array. 
Reading multiple selected channels will be possible through multichannel expansion, at the cost of making a separate UGen for each channel. If someone wants to read all channels in the most efficient way, they could specify channel = -1.

As a default, I would say channel 0. I think I would also have to rename the first argument from numChannels to numFrames or numOutputs, because it can be confusing to talk about channels in two different ways at the same time.

tedmoore: i think that all sounds great.

elgiano:
> What about BufChannelRd which would correspond to BufChannelWr.

I think BufChannelRd would give the impression of a BufRd that reads only a selected channel. We miss the part where the UGen is reading all samples out of a channel *at once*. Maybe something like BufDump in this sense? Although it can have a "destructive" connotation, and I can't think of a "complementary" ugen (BufUndump? BufLoad?)

> keeps a reference to BufRd / BufWr. Maybe something like BufDataRd / BufDataWr.

I think Data is kind of vague here, where Array is at least more connected to the array you get out of it? I also thought of BufRdArray and BufWrArray, but here I get a feeling that RdArray would read an array to a buf, instead of reading a buf as an array...

In the meanwhile I added the 'channel' argument and next I will make the complementary UGen

dietcv:
I just wanted to add after all these name suggestions, that figuring out a name definetly helps to understand the purpose of the ugen not only for using, but also for developing it.

prko:
Perhaps something along the following lines (some overlap with suggestions already made):

- `Buf2Array`
- `BufToArray`
- `BufArray`
- `BufRdArray`
- `BufRdAsArray`

`BufDump` also seems appealing, although it suggests that the output might include more than just the array itself. For example, it could additionally provide:

- The file path associated with the Buffer instance prior to conversion
- The buffer number of the Buffer instance prior to conversion
- The channel count of the Buffer instance prior to conversion
- The frame count of the Buffer instance prior to conversion
- The sample rate of the Buffer instance prior to conversion
- The duration of the Buffer instance prior to conversion

tedmoore:
> I think BufChannelRd would give the impression of a BufRd that reads only a selected channel.

that's a good point

dyfer:
I like BufRdArray (or BufRdAsArray). I don't feel it's very "elegant", but I think it's more accurate/descriptive than others.

cdbzb:
Each Buffer frame becomes a channel - So how about Frames2Channels or BufferFrames ?

dyfer:
> Each Buffer frame becomes a channel - So how about Frames2Channels or BufferFrames ?

Frames2Channels is nice, but I'd like to have "Buf" in the name (like with Playbuf, BufRd etc)

I don't think BufferFrames would work, as we have `BufFrames` which reports the number of frames in the buffer.

HelpSource/Classes/BufIn.schelp
telephon: switching between multichannel expansion and not-expansion usually is specified with  a Ref, like in Klank. Would that be useful here?

cdbzb:
BufFrames2Channels describes it completely I think! Maybe too long though... but nice to be able to understand without checking docs...

telephon:
I find `BufFrames2Channels` confusing, because `BufFrames` returns the number of frames. I found best things like `BufRdArray` or `BufRdChannels`.

tremblap:
Dear all. Indeed I am now also stuck with the overhead, so I was about to reinvent the wheel when @tedmoore pointed me to this. Obviously, FluCoMa will decommission the fluidbuf2kr (and hopefully kr2buf) if/when this PR comes off. Let me know if I can help in any way.

dyfer:
I like the latest suggestions on the name (`BufRdArray` or `BufRdChannels`), could we move forward with the PR? it would be a wonderful addition.

Issue #7181, opened 2025-10-11
Series creation shorthand (a,b..c) behavior when step (b-a)  is not a multiple of 0.5


HotwheelsSisyphus:
(3.15 dev version, macos 12.7.6, intel)

Just noticed this today, haven't found it documented anywhere. I'd consider this an actual "bug" but this one is most likely beyond me to fix, because it involves a syntax shorthand and I don't know where those are defined, most likely not in the class library? Possibly related to float arithmetics somehow and floor/ceil business around 0. It doesn't affect the explicit Array.series method because that one does not specify the end for the series, I imagine the bug relates to how the size is being calculated when the shorthand is used.

This seems only to happen with floats in specific cases, which makes it even more insidious...

Also: Changing that behavior would alter a lot of people's code I imagine?... At least I use this a lot.

Anyway
```
(-0.95,-0.9..-0.7)//-> [-0.95, -0.9, -0.85, -0.8, -0.75, -0.7]  includes end value
(-1,-0.95..-0.7) // -> [-1.0, -0.95, -0.9, -0.85, -0.8, -0.75] does not include it
(-2,-1.95..-1.7) //  does not include it
(-2.05,-2..-1.7)// includes it

(-3,-2.95..-2.7) // includes it
(-3,-2.9..-2.7).wrapAt(-1) / ->-2.8 ... does not (!?)

// another interesting one::
(-1.5,-1.01..1.44).wrapAt(-1) // -> 1.44 (as expected)
(-1.5,-1.01..1.93).wrapAt(-1) // -> 1.44  // step 0.49
(-2.5,-2.01..0.44).wrapAt(-1) //-> -0.05
(-2.5,-2.01..0.93).wrapAt(-1) // -> 0.44
```
I guess one would want to write proper tests for that one to figure out when exactly it happens...

HotwheelsSisyphus:
here's a bit of testing. Apparently it's not limited to that [-2 , -1] range after all, and happens basically with any step size not a multiple of 0.5
```
(

var test = { |start, step, prec=10|
        var end = (start+(50*step));
        var x = (start, (start+step) .. end).last;
        if((x - end).abs > (step/prec).abs) {
           [start, step] //
        }
    };
var starts = Array.series(2000, -10, 0.01);
var steps = Array.series(300, -1.5, 0.01);
~results = {|a,b| test.(starts[a],steps[b])}.dup([2000, 300]);
[...]

PR #7186, opened 2025-10-13
prSimpleNumberSeries (a,b..c) size calculation adjusted to offset float error. 

HotwheelsSisyphus:
To address #7181, where floats don't play nice with (a,b..c) (i.e., SimpleNumber.series).
EDIT (Jan 23): this is now good to be reviewed.

Summary (Oct 18): 
My changes concern the size calculation of series for floats with step size != 0 only.
The way this was done previously, for a series `(first, second..last`), where `step = second - first` and `range = last - first` is this:
Divide `range` by `step` and truncate; then add 1 to that. (I.e., `(int)(range/step + 1))`
That guaranteed that the final calculated value was never higher than the last argument, because that value for an array size n would be` first + (step * (size-1))`, which is the same as `first + (trunc(range/step) * step)`. But that also resulted in a brutal and a bit unpredictable chopping off of the last value when`step` (or `first`) had an inconvenient floating point representation. 

This PR adjust the calculation: it now multiplies `range/step` by `(1 + 2.pow(-49))` to compensate for floating point error.

This  for the following behavior:
```
//unchanged: 
(0.0, 2.0..7.0) // -> [0, 2, 4, 6] as before;
//  even, as an extreme example (from the added tests):
(0.001, 0.002.. 999.999999999998).size // -> 999999 (not 1e6!)
// this is a very artificial example, meant to illustrate that the error compensation is at the end of the day fairly small, 
// even where the array size isn't.  

// changed: cases where the exclusion of the last element was due to floating point error:
(-1.5, -1.45..1.5).last // -> 1.5; would have been 1.45 before
(-1.5, -1.45..-1.35).last // -> -1.35; would have been -1.4 before.
```

<details><summary>older musings, resolved</summary>
<p>
The rewrite here starts with a size is that is (often) 1 too little: and then checks if another element can be added that is within the range plus some tolerance, which is relative to step (to allow for floating point error). If yes, it increments the size by 1. If not, it doesn't.
Basically what remains to be done is:
Settle on tolerance and the way to calculate it - currently I have it set at a millionth of the step size, but I pulled that fraction out of thin air more or less. Gotta start somewhere.
Write some more tests, and weed out some of the more redundant ones I put in there.
Mabye mention it in the documentation.
</p>
</details>

### DIFF
--- a/HelpSource/Classes/SimpleNumber.schelp
+++ b/HelpSource/Classes/SimpleNumber.schelp
@@ -510,7 +510,10 @@ returns::
 An link::Classes/Array::.
 
 discussion::
-The last value may not be included in the result if the step size does not divide evenly into the range of the series.
+The last value will not be included in the result if the step size does not divide evenly into the range of the series. 
+(E.g., code:: series(5, 7, 10):: returns code::[5, 7, 9]::.)
+While this method internally compensates for floating point error in its size calculation,
+consider using link::Classes/Array#*series:: if stability of size is very important.
 
 method:: seriesIter
 Create a link::Classes/Routine:: that iterates over an arithmetic series from teletype::this:: over strong::second:: to strong::last::.
--- a/HelpSource/Classes/SimpleNumber.schelp
+++ b/HelpSource/Classes/SimpleNumber.schelp
@@ -509,11 +509,46 @@ code::
 returns::
 An link::Classes/Array::.
 
-discussion::
-The last value will not be included in the result if the step size does not divide evenly into the range of the series. 
-(E.g., code:: series(5, 7, 10):: returns code::[5, 7, 9]::.)
-While this method internally compensates for floating point error in its size calculation,
-consider using link::Classes/Array#*series:: if stability of size is very important.
+note::
+strong::Size of resultant Array:::
+If all arguments are link::Classes/Integer::s and the step size (teletype::second - this::) does not evenly subdivide the range (teletype::last - this::) of the series, 
+the teletype::last:: argument will not be included in the result.
+
+code:: series(5, 7, 10) // -> [5, 7, 9]::
+
+If arguments include link::Classes/Float::s, the behavior will be the same: 
+
+code::series(1.0, 2.0, 3.999999999) // -> [1.0, 2.0, 3.0]::
+
+In previous versions of SuperCollider (up to 3.14), cases could be encountered where floating point error resulted in omission of the final element:
+
+code::(1.5, 1.45..1.35).last // (SC 3.14) -> [1.5, 1.45, 1.4]; 1.35 omitted!:: 
+
+This is no longer the case as of version 3.15, as the method now compensates for its own floating point error in its calculation of the size.
+footnote:: 
+Note that the link::#-series:: method cannot possibly compensate for the floating point error that might have arisen in the calculations of the arguments themselves. 
+Therefore, in numerically extreme cases, it might be better to use an integer series with a multiplier, or  link::Classes/Array#*series::, where the size of the array can be directly given as an argument.
+code::
+( 
+//extremely contrived example of the kind of arguments
+// that contain enough floating point error to affect .series's size calculation:
+var one = (exp(log(17/11) + log(13/7)) * 7 * 11 / 13 /17).pow(2);
+var one_thousandths = 1/(one * 1000);
+var two_thousandths = (one + one) * one/1000;
+var size = (one_thousandths, two_thousandths, 1000/one).size;
+[one, one_thousandths, two_thousandths].collect (_.asStringPrec(20)).postln;
+size.postln;
+) // -> 999999; but algebraically, size should be 1000000
+
+(
+//for numerically complicated cases where size of the array is important, use Array.series 
+Array.series(1000000, one_thousandths, one_thousandths);
+// or an  integer series with a multiplier:
+(1..1000000) * one_thousandths
+)
+::
+::
+::
 
 method:: seriesIter
 Create a link::Classes/Routine:: that iterates over an arithmetic series from teletype::this:: over strong::second:: to strong::last::.
--- a/lang/LangPrimSource/PyrMathPrim.cpp
+++ b/lang/LangPrimSource/PyrMathPrim.cpp
@@ -885,7 +885,16 @@ int prSimpleNumberSeries(struct VMGlobals* g, int numArgsPushed) {
         if (step == 0.f) {
             size = 1;
         } else {
-            size = (int)((last - first) / step) + 1;
+            double range = last - first;
+            double tolerance = 0.000001; // 1e-6; pretty arbitrary choice. Discuss?
+            size = static_cast<int>(round(range / step)); // this will often be 1 too few
+            // size * step represents the value of a hypothetical additional element;
+            // but we only add it if it is within the allowed tolerance.
+            if (fabs(size * step) <= fabs(range + step * tolerance)) {
+                ++size;
+            }
+            // step and range will be of same sign (if not, error below);
+            // fabs() is only to not have to change comparison operator for positive and negative step/range.
         }
 
         if ((size < 1) || ((step >= 0) && (last < first)) || ((step <= 0) && (last > first))) {
--- a/lang/LangPrimSource/PyrMathPrim.cpp
+++ b/lang/LangPrimSource/PyrMathPrim.cpp
@@ -886,15 +886,9 @@ int prSimpleNumberSeries(struct VMGlobals* g, int numArgsPushed) {
             size = 1;
         } else {
             double range = last - first;
-            double tolerance = 0.000001; // 1e-6; pretty arbitrary choice. Discuss?
-            size = static_cast<int>(round(range / step)); // this will often be 1 too few
-            // size * step represents the value of a hypothetical additional element;
-            // but we only add it if it is within the allowed tolerance.
-            if (fabs(size * step) <= fabs(range + step * tolerance)) {
-                ++size;
-            }
-            // step and range will be of same sign (if not, error below);
-            // fabs() is only to not have to change comparison operator for positive and negative step/range.
+            double size_w_error = range / step;
+            // the multiplier below is for float error compensation and equals 1 + 2.pow(-49)
+            size = static_cast<int>(0x1.0000000000008p0 * size_w_error) + 1; // cast truncates
         }
 
         if ((size < 1) || ((step >= 0) && (last < first)) || ((step <= 0) && (last > first))) {
--- a/testsuite/classlibrary/TestSimpleNumber.sc
+++ b/testsuite/classlibrary/TestSimpleNumber.sc
@@ -241,19 +241,42 @@ TestSimpleNumber : UnitTest {
 	}
 
 	test_series {
-		var first = 0;
-		var step = 2.0001;
-		var last = 8;
-		var arr = first.series(step, last);
-		this.assert(arr.last <= last, "SimpleNumber:series should not produce an array whose last value is greater than the specified 'last' argument.");
-
-		first = 1;
-		arr = first.series(first, first);
-		this.assert(arr.size == 1, "SimpleNumber:series Int types with first == last and step == 0 should return an array of [ first ]");
-
-		first = 1.1;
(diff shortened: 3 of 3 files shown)


dyfer:
@HotwheelsSisyphus please note that `TestSimpleNumber -test_series` fails with this change.

HotwheelsSisyphus:
yup, that makes sense. was worth a try anyway.
I figure here someone more experienced needs to make a call as to the acceptable tolerance. 

The test is (0,2.0001..8) so it should reasonably get caught by this error: `arguments do not form an arithmetic progression.` (line 892)... 

One could add a condition to that error check along the lines of `(abs(last - (size - 1) * step) - abs(first)) > [some very small number]`. There's probably best practices and design patterns for this that I don't know.

JordanHendersonMusic:
I'd start by writing more tests, including the one that is broken. There are many paths in this primitive and it would be good to cover all of them, essentially a test for each if statement.

HotwheelsSisyphus:
I'm looking into the testing... taking this as an opportunity to understand floats a bit better.

But generally I figured one would want a behavior something like this:
1. if the calculated final element is close enough (according to some tolerance) to the argument provided as `last`, simply accept that close-enough final element;
2. if it's not, throw an error ("not an arithmetic progression"); or post a warning, and exclude the `last` element.

I imagine the tolerance, if one wants one, is best set relative to step size.
But I'm not so sure about the error vs. warning thing... I wasn't aware that for instance `(0,2..7)` is currently permitted (returns `[0,2,4,6]` without warning or error).  I had just never tried it.

JordanHendersonMusic:
> I imagine the tolerance, if one wants one, is best set relative to step size.

I believe you can calculate the error using something called the floating point epsilon. Not 100% sure how you'd do that though! Epsilon times number of steps maybe?

I don't know if an error is necessary if the element doesn't exactly match, as you can just define the sequence to be less than or equal to it. This is what it does right now and the error can be reserved for sequences that don't approach the final value.

There might be more issues like this which is why I'd write more tests first to find as many as possible.

HotwheelsSisyphus:
> something called the floating point epsilon

My (limited) understanding is that epsilon is either used simply as a synonym for tolerance, or it refers to the machine epsilon / unit in the last place (Kahan: ulp(x) is the gap between the two floating-point numbers nearest to x, even if x is one of them) , which is a very low level thing that we probably don't need here. I think there's a function for that in the boost library though.

I'd argue in favor of setting tolerance as a proportion of step.
 Reason: 
 [1, 2, 3, 4 + 1e-6]
 is "more ok" than 
 [1e-5, 2e-5, 3e-5, 4.1e-5]. 
Maybe number of elements can figure in here as well, but I think the step size is the most important because it's the one thing that contains clues about what the relevant precision and order of magnitude is.

> I don't know if an error is necessary 

Right, after thinking about that a bit more I agree.

HotwheelsSisyphus:
(pushed this mainly to test the changes I made, and lo, it doesn't do the thing it's meant to do.)

dyfer:
@HotwheelsSisyphus you know that you can run tests locally, right? Just add the `testsuite/classlibrary` folder from the source to the compiled paths and run the test with `TestSimpleNumber.run`.

HotwheelsSisyphus:
@dyfer thanks... yes, the issue is more that I don't have a decent build environment set up on the machine I'm currently on. So I can only test the primitives there as they are in the develop branch, not my rewrite attempts in this branch... I'll figure this out locally but it's a bit involved because that laptop is old enough that the relevant homebrew packages are no longer provided pre-built, etc...

dyfer:
@HotwheelsSisyphus ah got it. 

<details><summary>If you want to try building on an older OS...</summary>

If you are determinet to get SC to build on an older OS where homebrew packages are not available, I'd suggested trying the vcpkg route. You can get libsndfile, fftw, and readline from vcpkg (if you build without supernova you don't need portaudio, and if you do need it, you can use `-D SYSTEM_PORTAUDIO=OFF` to build with the bundled one). We don't have this well documented, but this is what we do for the legacy build in CI. After installing vcpkg and relevant packages, set VCPKG_ROOT env variable and maybe other cmake flags (see [here](https://github.com/supercollider/supercollider/blob/develop/.github/workflows/build_macos.yml#L189-L190)). We may or may not need to use one of the triplets we ship in SC, I'm not sure (they set static/shared libs to what I found is working in our case).

For Qt, you can install it with the official installer or aqt-install, possibly adding it to the PATH (I'm not sure if that's needed or if these tools do that for you).

There's a short [description for VCPKG on Windows](https://github.com/supercollider/supercollider/blob/develop/README_WINDOWS.md#readline-support), aside from shell and OS specific stuff (e.g. the triplet) it should be similar on macOS.

I hope this helps!
</details>

HotwheelsSisyphus:
The tests pass now! I figure that the remaining CI checks probably don't have anything to do with my changes? But who knows. Will try to build it later on my other computer to see.

Anyway, changes concern the size calculation of series for floats with step size != 0 only.
The way this was done previously, for a series `(first, second..last)`, where `step = second - first` and `range = last - first` is this:
Divide range by step and truncate; then add 1 to that. (I.e., (int)(range/step + 1))
That guaranteed that the final calculated value was never higher than the `last` argument, because that value for an array size n would be `first + (step * (size-1))`, which is the same as `first + trunc(range/step) * step`. But that also resulted in a brutal and a bit unpredictable chopping off of the last value when the step had an inconvenient floating point representation.

The rewrite here starts with a size is that is (often) 1 too little and then checks if another element can be added that is within the range plus some tolerance. If yes, it increments the size by 1. If not, it doesn't.

Basically what remains to be done is:
1. Settle on tolerance and the way to calculate it - currently I have it set at a millionth of the step size.
2. Write some more tests, and weed out some of the more redundant ones I put in there.

dyfer:
macOS and Windows builds [here](https://github.com/supercollider/supercollider/actions/runs/18602399264/job/53043877907?pr=7186) failed for unrelated reasons (just to confirm - FYI)

prko:
>`(-1.5, -1.45..1.5).last // -> 1.5; would have been 1.45 before`

This is quite neat.  
I had assumed it would be the user’s responsibility, since the following also works:
`(-1.5, -1.45.trunc(0.05) ..1.5).last // -> 1.5`

dyfer:
> I figure that the remaining CI checks probably don't have anything to do with my changes?

@HotwheelsSisyphus if you rebase the PR on current develop all builds should work - just FYI.

HotwheelsSisyphus:
> I had assumed it would be the user’s responsibility

Basically everything remains the user's responsibility, because afaik the license disavows "fitness for any particular purpose"... :P 
Nonetheless, it seems like an easy and harmless quality of life fix in hindsight that saves users some typing. There have been other options before, like Array.series etc...

telephon:
It seems that the value of tolerance is fairly arbitrary (or case-dependent) ?

HotwheelsSisyphus:
Yes, exactly. It can strictly speaking only be decided with respect to some use case. 

However, two points:
1. I’d argue that tolerance = 0 (which is the same as current behavior before this PR) is _also_ a value of tolerance, and one that results in numerically « unstable »  behavior (namely with regard to the length of the series). 
2. The step size does contain some minimal information about what the relevant order of magnitude is. I.e., if the step size is 2, then a few millionths won’t make a relevant difference to code execution; if the step size is a millionth, then of course a few millionth will make difference.. 

So IMO, there should be a nonzero tolerance, and it should be relative to (i.e., some fraction of) the step size; what remains to be decided is what fraction exaclty (i.e., `stepsize * 1e-6`, or `stepsize * 1e-7` etc.). 
For this, one could compare some number crunching libraries like numpy, pandas, R or MATLAB or whatever. I will look into that.

HotwheelsSisyphus:
Short update: I've been thinking about this more than is strictly healthy, and the thing is, there isn't really an abstract solution for this. Obviously functions like `numpy.isclose` exist but their defaults are quite arbitrary. Numpy and matlab have functions called `linspace`, but none of these are specified by supplying step size _and_ end value. (I.e., you specify start, stop, num, and the step value is calculated for you). 
There is also `numpy.arange`, but a) this has similar stability problems (for different reasons, apparently); and b) the end value is generally not included.

I also tried to figure out the upper bound on the error, which is very doable on paper, but in practice I haven't got it to work, at least not in a sclang mockup. Also it does seem like overkill and adds extra calculations.


-------------------
Right now I'm leaning toward a simpler approach: Since the problem is the size calculation, and arises because `range`/`step` is floored on values that  (because of float errors) come out very close to, but still below, integer values, it may be sensible, instead of (pseudocode) 
```
size = floor(range/step) + 1
// for example
size = floor(3.0/0.050000001) + 1 // -> 60
```
to write 
```
size = floor(range/step * 1.0000001) + 1
// for example
size = floor(3.0/0.050000001 * 1.0000001) + 1 // -> 61

```
This is enough to push these cases that one naively feels "should work", like that (-1.5,-1.45..1.5), into the next higher size; and it doesn't require an extra conditional. It _will_ of course add extra elements to series of size 1e8 or larger (or however many zeroes I put there), but I figure if you have that many elements in an array, you're probably not worried about one extra. (Also, isn't there a maximum size on arrays anyway?). 

-------------------
I don't think this is terribly urgent as far as sc is concerned, I mostly just want to stop thinking about it.

HotwheelsSisyphus:
Long story short, I think this is finally good for review now. I tried to actually calculate the upper bound on the error, but in the end that didn't compensate enough, so I probably calculated wrong. Anyway, to never have to think about this again, the added compensation is `(range/step) * epsilon * 8` (when I would have thought that 4 or 5 epsilon would be enough). 

I've opted for multiplying the `range/step` by $1 + 2^{-49}$. This seems to get most problem cases but otherwise keeps the current behavior.

HelpSource/Classes/SimpleNumber.schelp
```diff
@@ -510,7 +510,10 @@ returns::
 An link::Classes/Array::.
 
 discussion::
-The last value may not be included in the result if the step size does not divide evenly into the range of the series.
+The last value will not be included in the result if the step size does not divide evenly into the range of the series. 
+(E.g., code:: series(5, 7, 10):: returns code::[5, 7, 9]::.)
+While this method internally compensates for floating point error in its size calculation,
+consider using link::Classes/Array#*series:: if stability of size is very important.
```
telephon: Thank you for looking into it carefully.

Since this syntax is the main way to produce series of values, we should consider this only if we have some way to make it reproducible (across systems and over time).

The minimum would be to have a way that calculates the series length for the given values.

How does NumPy do it?

HotwheelsSisyphus: > The minimum would be to have a way that calculates the series length for the given values.

huh? this is exactly what the code is doing. Please look at it again.

It should be reproducible across systems and time (that's the point of IEEE754), and I added unit tests that should return errors if ever that changes. Perhaps I didn't formulate this quite the right way. By "stable" I meant numerically stable, in the sense that small changes in some x never produce large changes in some f(x). This does not mean indeterministic. We cannot guarantee stability of size because we cannot know what values users are going to put there as arguments.
I can choose different wording for the documentation.

> How does NumPy do it?

as I wrote above:
> Short update: I've been thinking about this more than is strictly healthy, and the thing is, there isn't really an abstract solution for this. Obviously functions like numpy.isclose exist but their defaults are quite arbitrary. Numpy and matlab have functions called linspace, **but none of these are specified by supplying step size and end value. (I.e., you specify start, stop, num, and the step value is calculated for you).**
> There is also numpy.arange, but a) this has similar stability problems (for different reasons, apparently); and b) the end value is generally not included.

telephon: > It should be reproducible across systems and time (that's the point of IEEE754), and I added unit tests that should return errors if ever that changes. Perhaps I didn't formulate this quite the right way.

Ah ok, I misunderstood, sorry!



> > The minimum would be to have a way that calculates the series length for the given values.
> 
> huh? this is exactly what the code is doing. Please look at it again.

I meant that if the method's behaviour were not reproducible, we would need something that gives us the information without having to calculate the series itself:

```
seriesLength { |second, last|
    <return the length of the series you will get>
}
```

But since you say that it is reproducible then it should be fine, I think.

PR #7189, opened 2025-10-18
SCLang: Gluons: implement ffi interface

JordanHendersonMusic:
## Purpose and Motivation

Create an a way of calling dynamic libraries from inside sc.

See `include/gluons_ffi_interface` for the c api.

There are also tests using an inbuilt library, `TestGluon.run`, and some c tests using a testing library.

## Current state of this PR

Here is some code that currently works.

```supercollider
~l = Gluon.open("...");
~l.foo(2.0, 4.2); // 6.2 
~l.close

// or
Gluon.with("...") { |l|
   l.foo(2.0, 4.2)
};
```

Here is the implementation of `foo`
```cpp
sc_gluon_out_param_tag_v1 add_f64(sc_gluon_library_data_v1_t library_data,
                                  sc_gluon_callback_object_v1_t maybe_callback_data, sc_gluon_param_v1_t* in_params,
                                  uint32_t num_in_params, sc_gluon_out_param_or_maybe_diagnostic_v1* out_param) {
    if (num_in_params != 2) {
        out_param->maybe_diagnostic = "wrong number of in params";
        return sc_gluon_error_with_non_owned_diagnostic;
    }

    const sc_gluon_param_v1_t& p1 = in_params[0];
    const sc_gluon_param_v1_t& p2 = in_params[1];

    if (p1.tag != sc_gluon_f64) {
        out_param->maybe_diagnostic = "param 1 is not an f64";
        return sc_gluon_error_with_non_owned_diagnostic;
    }

    if (p2.tag != sc_gluon_f64) {
        out_param->maybe_diagnostic = "param 2 is not an f64";
        return sc_gluon_error_with_non_owned_diagnostic;
    }

    const auto r = p1.data.f64 + p2.data.f64;

    out_param->out_param.data.f64 = r;
    out_param->out_param.tag = sc_gluon_f64;
    out_param->out_param.owns_data = false;
    out_param->out_param.size = 1;

    return sc_gluon_produced_param;
}

```

Additionally, callbacks can be used.

```supercollider
Gluon.with("...") { |l|
   l.with_call_back("hello", callback: { |count, a|
      [count, a].postln
  });
}; // returns 1.2

// waits for 1
// prints [1, "hello"]
// waits for 1
// prints  [2, "hello"]

```

Here is the implementation of that.
```cpp

sc_gluon_out_param_tag_v1 with_call_back(sc_gluon_library_data_v1_t library_data,
                                         sc_gluon_callback_object_v1_t maybe_callback_data,
                                         sc_gluon_param_v1_t* in_params, uint32_t num_in_params,
                                         sc_gluon_out_param_or_maybe_diagnostic_v1* out_param) {
    auto* lib = static_cast<LibraryData*>(library_data);

    if (num_in_params != 1) {
        out_param->maybe_diagnostic = "wrong number of in params";
        return sc_gluon_error_with_non_owned_diagnostic;
    }

    std::thread([=, param_copy = sc_gluon_copy_param_v1(in_params[0])]() mutable {
        using namespace std::chrono_literals;
        std::this_thread::sleep_for(1s);

        if (maybe_callback_data) {
            param_t ps[2] = { create_param(1), sc_gluon_copy_param_v1(param_copy) };
            lib->callback_f(maybe_callback_data, ps, 2);
        }

        std::this_thread::sleep_for(1s);

        if (maybe_callback_data) {
            param_t ps[2] = { create_param(2), sc_gluon_copy_param_v1(param_copy) };
            lib->callback_f(maybe_callback_data, ps, 2);
        }

        sc_gluon_free_param_if_owned_v1(param_copy);

        if (maybe_callback_data)
            lib->release_callback_f(maybe_callback_data);

        return;
    }).detach();

    out_param->out_param = create_param(1.2);
    return sc_gluon_produced_param;
}
```

## Types of changes
- New feature

## To-do list

- [x] Implement windows version of `dlopen` and friends  ??? Current the CI is doing something odd.
- [x] Test the param type more, currently, only basics types have been tested, these could leak if they are done wrong.
- [x] Test callbacks
- [x] Figure out if the headers should be in their own repo — No
- [x] Investigate returnFromBlock at the top of the stack, doesn't seem to cause any issues, so why did it post a warning?  Answer: it causes subtle stack issues, I've made it a fatal error now as it will eventually crash the interpreter.

### DIFF
--- a/include/gluon_ffi_interface/sc_gluon_v1_entry_points.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_entry_points.h
@@ -1,0 +1,26 @@
+#pragma once
+#include "sc_gluon_v1_types.h"
+
+#ifdef __cplusplus
+#    define SC_GLUON_EXTERN extern "C"
+#endif
+
+#define SC_GLUON_EXPORT SC_GLUON_EXTERN __attribute__((__visibility__("hidden")))
+
+// Required
+SC_GLUON_EXPORT uint32_t sc_gluon_version();
+
+// Required
+SC_GLUON_EXPORT sc_gluon_library_data_v1_t
+sc_gluon_load_library(sc_gluon_do_callback_v1_f, sc_gluon_release_callback_object_v1_f,
+                      struct sc_gluon_function_declarations_v1_t** const out_decls, uint32_t* out_size);
+
+
+// Optional
+SC_GLUON_EXPORT void sc_gluon_post_load_library(sc_gluon_library_data_v1_t,
+                                                struct sc_gluon_function_declarations_v1_t* decls_to_be_freed,
+                                                uint32_t decls_size);
+
+
+// Optional
+SC_GLUON_EXPORT void sc_gluon_unload_library(sc_gluon_library_data_v1_t);
--- a/include/gluon_ffi_interface/sc_gluon_v1_entry_points.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_entry_points.h
@@ -5,7 +5,7 @@
 #    define SC_GLUON_EXTERN extern "C"
 #endif
 
-#define SC_GLUON_EXPORT SC_GLUON_EXTERN __attribute__((__visibility__("hidden")))
+#define SC_GLUON_EXPORT SC_GLUON_EXTERN __attribute__((__visibility__("default")))
 
 // Required
 SC_GLUON_EXPORT uint32_t sc_gluon_version();
--- a/include/gluon_ffi_interface/sc_gluon_v1_entry_points.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_entry_points.h
@@ -1,3 +1,5 @@
+// SuperCollider foreign function interface.
+// Author Jordan Henderson - JordanHendersonMusic 2025
 #pragma once
 #include "sc_gluon_v1_types.h"
 
@@ -5,15 +7,21 @@
 #    define SC_GLUON_EXTERN extern "C"
 #endif
 
-#define SC_GLUON_EXPORT SC_GLUON_EXTERN __attribute__((__visibility__("default")))
+#if defined(_WIN32) || defined(_WIN64) || defined(WIN32) || defined(WIN64)
+#    define SC_GLUON_EXPORT SC_GLUON_EXTERN __declspec(dllexport)
+#else
+#    define SC_GLUON_EXPORT SC_GLUON_EXTERN __attribute__((__visibility__("default")))
+#endif
 
 // Required
 SC_GLUON_EXPORT uint32_t sc_gluon_version();
 
 // Required
-SC_GLUON_EXPORT sc_gluon_library_data_v1_t
-sc_gluon_load_library(sc_gluon_do_callback_v1_f, sc_gluon_release_callback_object_v1_f,
-                      struct sc_gluon_function_declarations_v1_t** const out_decls, uint32_t* out_size);
+SC_GLUON_EXPORT uint8_t sc_gluon_load_library(sc_gluon_param_v1_t* in_params, uint32_t num_in_params,
+                                              sc_gluon_do_callback_v1_f do_callback,
+                                              sc_gluon_release_callback_object_v1_f release_callback,
+                                              sc_gluon_function_declarations_v1_t** const out_decls,
+                                              uint32_t* out_decls_size, sc_gluon_library_data_v1_t* out_library_data);
 
 
 // Optional
--- a/include/gluon_ffi_interface/sc_gluon_v1_types.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_types.h
@@ -135,10 +135,10 @@ union sc_gluon_data_v1 {
 
 struct sc_gluon_param_v1_t {
     union sc_gluon_data_v1 data;
+    uint32_t size; // For non-array parameters, this is ignored, but set to '1' as a good practice.
     enum sc_gluon_param_tag_v1 tag;
     bool owns_data; // Indicates that the param owns some heap allocated data, and will be deallocated when the lifetime
                     // of the parameter ends.
-    uint32_t size; // For non-array parameters, this is ignored, but set to '1' as a good practice.
 };
 
 union sc_gluon_out_param_or_maybe_diagnostic_v1 {
--- a/include/gluon_ffi_interface/sc_gluon_v1_types.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_types.h
@@ -179,3 +179,44 @@ inline void sc_gluon_free_param_v1(struct sc_gluon_param_v1_t param) {
         }
     }
 }
+
+inline struct sc_gluon_param_v1_t sc_gluon_copy_param_data_v1(struct sc_gluon_param_v1_t param) {
+    switch (param.tag) {
+    case sc_gluon_param_array: {
+        // Recursive case.
+        sc_gluon_data_v1 d {};
+        d.param_array = (sc_gluon_param_v1_t*)malloc(param.size * sizeof(sc_gluon_param_v1_t));
+        for (uint32_t i = 0; i < param.size; ++i)
+            d.param_array[i] = sc_gluon_copy_param_data_v1(param.data.param_array[i]);
+
+        return { d, param.size, sc_gluon_param_array, true };
+    }
+    case sc_gluon_u8_array: {
+        sc_gluon_data_v1 d {};
+        d.u8_array = (uint8_t*)malloc(param.size * sizeof(uint8_t));
+        memcpy(d.u8_array, param.data.u8_array, param.size * sizeof(uint8_t));
+        return { d, param.size, sc_gluon_u8_array, true };
+    }
+    case sc_gluon_f64_array: {
+        sc_gluon_data_v1 d {};
+        d.f64_array = (double*)malloc(param.size * sizeof(double));
+        memcpy(d.f64_array, param.data.f64_array, param.size * sizeof(double));
+        return { d, param.size, sc_gluon_f64_array, true };
+    }
+    case sc_gluon_f32_array: {
+        sc_gluon_data_v1 d {};
+        d.f32_array = (float*)malloc(param.size * sizeof(float));
+        memcpy(d.f32_array, param.data.f64_array, param.size * sizeof(float));
+        return { d, param.size, sc_gluon_f32_array, true };
+    }
+    case sc_gluon_char_array: {
+        sc_gluon_data_v1 d {};
+        d.character_array = (char*)malloc(param.size * sizeof(char));
+        memcpy(d.character_array, param.data.character_array, param.size * sizeof(char));
+        return { d, param.size, sc_gluon_char_array, true };
+    }
+    default: {
+        return param;
+    }
+    }
+}
--- a/include/gluon_ffi_interface/sc_gluon_v1_types.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_types.h
@@ -3,7 +3,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <stdlib.h>
-
+#include <cstring>
 /*
 
 SuperCollider foreign function interface.
--- a/include/gluon_ffi_interface/sc_gluon_v1_types.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_types.h
@@ -111,8 +111,7 @@ enum sc_gluon_param_tag_v1 {
     sc_gluon_u8_array,
     sc_gluon_f64_array,
     sc_gluon_f32_array,
-    sc_gluon_char_array,
-    sc_gluon_symbol_value_array,
+    sc_gluon_char_array, // Note, this is not a null terminated string, but an array with a size.
     //
     sc_gluon_param_array,
 };
@@ -124,12 +123,11 @@ union sc_gluon_data_v1 {
     char character;
     bool boolean;
     void* raw_pointer;
-    intptr_t symbol_value;
+    uint64_t symbol_value;
     uint8_t* u8_array;
     double* f64_array;
     float* f32_array;
     char* character_array;
-    intptr_t* symbol_value_array;
     struct sc_gluon_param_v1_t* param_array;
 };
 
@@ -173,10 +171,6 @@ inline void sc_gluon_free_param_v1(struct sc_gluon_param_v1_t param) {
             }
             case sc_gluon_char_array: {
                 free(param.data.character_array);
-                break;
-            }
-            case sc_gluon_symbol_value_array: {
-                free(param.data.symbol_value_array);
                 break;
             }
             default:
--- a/include/gluon_ffi_interface/sc_gluon_v1_types.h
+++ b/include/gluon_ffi_interface/sc_gluon_v1_types.h
@@ -1,18 +1,15 @@
-
+// SuperCollider foreign function interface.
+// Author Jordan Henderson - JordanHendersonMusic 2025
 #pragma once
 #include <stdint.h>
-#include <stdbool.h>
 #include <stdlib.h>
 #include <cstring>
 /*
 
-SuperCollider foreign function interface.
-Author Jordan Henderson - JordanHendersonMusic 2025
-
 There are three parts to this header:
     1. Entry points, these are functions that will be called from the sc host. They must be named accordingly.
-    2. The definition of the function signature. When implementing a function to be called from sc, it must have this
-        signature.
+    2. The definition of the foreign function signature. When implementing a function to be called from sc, it must have
+        this signature.
     3. The parameter definition. To avoid exposing sc internals (which may change), an intermediate format is given for
         arguments, this means not all supercollider types can be passed to foreign functions.
 
@@ -23,82 +20,107 @@ The three types of entry point are 'version', 'load', 'post_load', and 'unload'.
 
     In 'load', a pointer to an array of structures describing what functions the library offers must be set.
     A name, an implementation pointer, the number of params expected (-1 if accepts any number), and a
-    bool indicating whether the function accepts a callback must all be provided. A pointer to some data
-    for the library may be returned, this will be passed to each function allowing for library state without using
-    globals. Additionally, the function provides a callback function, which should be stored in the case a function
-    wishes to return execution back to sc from another thread.
+    uint8_t (bool) indicating whether the function accepts a callback must all be provided. A pointer to some data
+    for the library may be returned via an out param, this will be passed to each function allowing for library state
+    without using globals. Additionally, the function provides a callback function, which should be stored in the case a
+    function wishes to return execution back to sc from another thread. Parameters from supercollider can also be passed
+    in at this stage. This function must return a parameter (even nil) to indicate a successful load. This parameter is
+    NOT passed back to supercollider.
 
     In 'post_load' you are provided the opportunity to deallocate the array of function descriptions.
 
     In 'unload' you can deallocate the library state.
 
 
-The functions accepts the library state, a nullable pointer to some callback object (this is a type erased
-PyrObject* of which `value' will be called upon), the input parameter array, and an output parameter or potential
-diagnostic. Additionally, it must return a tag indicating whether an output parameter or diagnostic has been produced.
+The foreign functions accepts the library state, a nullable pointer to some callback object, the input parameter array,
+and an output parameter or potential diagnostic. Additionally, it must return a tag indicating whether an output
+parameter or diagnostic has been produced.
 
-Parameters are essentially tagged unions that when containing allocated resources indicate their size and whether they
+Parameters are tagged unions that when containing allocated resources indicate their size and whether they
 own their memory.
 
-All input parameters that have allocated resources (arrays) will be freed after the function has finished if they 'own'
-their data. To prevent this, the 'owns_data' member may be set to false, thereby taking ownership of the memory.
+All input parameters that have allocated resources (arrays) will be freed after the foreign function has finished if
+they 'own' their data. To prevent this, the 'owns_data' member may be set to 0, thereby taking ownership of the memory.
 This memory can either be freed manually at a later date, or returned as a part of an output parameter, ensuring the
 output parameter 'owns' the data.
 
 Input parameters who do not own their data have unknown lifetimes, the data will persist for the duration of the
-function call, but may be freed immediately after, or by sc's garbage collector at an unknown point in time.
+foreign function call, but may be freed immediately after, or by sc's garbage collector at an unknown point in time.
+Mutating 'non-owned' data will allows you to directly alter the contents of structures like Signals without creating
+copies.
 
-When passing any 'owned' memory back to the sc host, it must be allocate with malloc as free will be called on it.
-Alternatively, you can pass unowned memory back to the host and no attempt to deallocate will be made.
+
+When passing parameters back to supercollider, if the memory is marked as 'owned' no copy will be made and supercollider
+will take ownership of it. Ensure that all memory stored inside of a parameter is valid to be deallocated with 'free'.
+If in doubt, set the 'owns_data' flag to 0, and supercollider will copy the data.
 
 */
 
+////// Report which version of the gluon ffi interface used..
 typedef uint32_t (*sc_gluon_version_f)();
 
+////// Pointer to custom user defined library data
 typedef void* sc_gluon_library_data_v1_t;
-typedef void* sc_gluon_callable_object_v1_t;
-typedef void (*sc_gluon_do_callback_v1_f)(sc_gluon_callable_object_v1_t, struct sc_gluon_param_v1_t* params,
+
+////// Pointer to some gc reserved supercollider object
+typedef void* sc_gluon_callback_object_v1_t;
+
+////// Function to trigger callback object in supercollider, can pass parameters, must not be called on the main thread.
+typedef void (*sc_gluon_do_callback_v1_f)(sc_gluon_callback_object_v1_t, struct sc_gluon_param_v1_t* params,
                                           uint32_t num_params);
-typedef void (*sc_gluon_release_callback_object_v1_f)(sc_gluon_callable_object_v1_t);
 
-//////////// ENTRY POINTS FOR LIBRARY.
+////// Release the callback object allowing the garbage collector to clean it up, must not be called on the main thread.
+typedef void (*sc_gluon_release_callback_object_v1_f)(sc_gluon_callback_object_v1_t);
 
+
+//////////// MAIN ENTRY POINTS FOR LIBRARY.
+
+////// Load library
 // Required. Must be called: sc_gluon_load_library
-typedef sc_gluon_library_data_v1_t (*sc_gluon_load_library_v1_f)(
-    sc_gluon_do_callback_v1_f, sc_gluon_release_callback_object_v1_f,
-    struct sc_gluon_function_declarations_v1_t** const out_decls, uint32_t* out_size);
+// Return 0 for successful load.
+typedef uint8_t (*sc_gluon_load_library_v1_f)(struct sc_gluon_param_v1_t* in_params, uint32_t num_in_params,
+                                              sc_gluon_do_callback_v1_f do_callback,
+                                              sc_gluon_release_callback_object_v1_f release_callback,
+                                              struct sc_gluon_function_declarations_v1_t** const out_decls,
+                                              uint32_t* out_decls__size, sc_gluon_library_data_v1_t* out_library_data);
 
+////// Post load library
 // Optional. Must be called: sc_gluon_post_load_library.
 typedef void (*sc_gluon_post_load_library_v1_f)(sc_gluon_library_data_v1_t,
                                                 struct sc_gluon_function_declarations_v1_t* decls_to_be_freed,
                                                 uint32_t decls_size);
 
+////// Post unload library
 // Optional. Must be called: sc_gluon_unload_library.
 typedef void (*sc_gluon_unload_library_v1_f)(sc_gluon_library_data_v1_t);
 
-////////////  FUNCTION PTRS
 
+//////////// Foreign function signature
+////// Functions that will be called in supercollider must have this signature.
+
+////// Used to report whether some function produce some parameter to pass back to supercollider, or whether an error
+/// occurred.
 enum sc_gluon_out_param_tag_v1 {
     sc_gluon_produced_param,
     sc_gluon_error_with_owned_diagnostic,
     sc_gluon_error_with_non_owned_diagnostic,
-    sc_gluon_error_without_diagnostic,
 };
 
 typedef enum sc_gluon_out_param_tag_v1 (*sc_gluon_function_v1_f)(
-    sc_gluon_library_data_v1_t library_data, sc_gluon_callable_object_v1_t maybe_callback_data,
+    sc_gluon_library_data_v1_t library_data, sc_gluon_callback_object_v1_t maybe_callback_data,
     struct sc_gluon_param_v1_t* in_params, uint32_t num_in_params,
     union sc_gluon_out_param_or_maybe_diagnostic_v1* out_param);
 
-
-//////////// PARAMS
-
+////// Used to report back to supercollider what functions the library implements.
+////// See sc_gluon_load_library_v1_f
 struct sc_gluon_function_declarations_v1_t {
     const char* name;
     sc_gluon_function_v1_f ptr;
     int32_t num_parms; // negative means accepts any number of arguments.
-    bool accepts_callback;
+    uint8_t accepts_callback;
 };
+
+//////////////// PARAMETERS
 
 enum sc_gluon_param_tag_v1 {
     sc_gluon_nil,
@@ -115,13 +137,12 @@ enum sc_gluon_param_tag_v1 {
     //
     sc_gluon_param_array,
 };
-
 union sc_gluon_data_v1 {
     int32_t nil_;
     int32_t i32;
     double f64;
     char character;
-    bool boolean;
+    uint8_t boolean;
     void* raw_pointer;
     uint64_t symbol_value;
     uint8_t* u8_array;
@@ -135,22 +156,25 @@ struct sc_gluon_param_v1_t {
     union sc_gluon_data_v1 data;
     uint32_t size; // For non-array parameters, this is ignored, but set to '1' as a good practice.
     enum sc_gluon_param_tag_v1 tag;
-    bool owns_data; // Indicates that the param owns some heap allocated data, and will be deallocated when the lifetime
-                    // of the parameter ends.
+    uint8_t owns_data; // Indicates that the param owns some heap allocated data, and will be deallocated when the
+                       // lifetime of the parameter ends.
 };
 
+////// Holds either an parameter to pass back to supercollider, or a null terminated diagnostic string.
 union sc_gluon_out_param_or_maybe_diagnostic_v1 {
     struct sc_gluon_param_v1_t out_param;
+    // sc_gluon_out_param_tag_v1 indicates whether this is an owned pointer. Null terminated. Can be nullptr.
     const char* maybe_diagnostic;
 };
+
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // HELPERS
 
-inline void sc_gluon_free_param_v1(struct sc_gluon_param_v1_t param) {
+inline void sc_gluon_free_param_if_owned_v1(struct sc_gluon_param_v1_t param) {
     if (param.tag == sc_gluon_param_array) {
         for (uint32_t i = 0; i < param.size; ++i)
-            sc_gluon_free_param_v1(param.data.param_array[i]);
+            sc_gluon_free_param_if_owned_v1(param.data.param_array[i]);
 
         if (param.owns_data)
             free(param.data.param_array);
@@ -180,40 +204,40 @@ inline void sc_gluon_free_param_v1(struct sc_gluon_param_v1_t param) {
     }
 }
 
-inline struct sc_gluon_param_v1_t sc_gluon_copy_param_data_v1(struct sc_gluon_param_v1_t param) {
+inline struct sc_gluon_param_v1_t sc_gluon_copy_param_v1(struct sc_gluon_param_v1_t param) {
     switch (param.tag) {
     case sc_gluon_param_array: {
         // Recursive case.
         sc_gluon_data_v1 d {};
         d.param_array = (sc_gluon_param_v1_t*)malloc(param.size * sizeof(sc_gluon_param_v1_t));
         for (uint32_t i = 0; i < param.size; ++i)
-            d.param_array[i] = sc_gluon_copy_param_data_v1(param.data.param_array[i]);
+            d.param_array[i] = sc_gluon_copy_param_v1(param.data.param_array[i]);
 
-        return { d, param.size, sc_gluon_param_array, true };
+        return { d, param.size, sc_gluon_param_array, 1 };
     }
     case sc_gluon_u8_array: {
         sc_gluon_data_v1 d {};
         d.u8_array = (uint8_t*)malloc(param.size * sizeof(uint8_t));
         memcpy(d.u8_array, param.data.u8_array, param.size * sizeof(uint8_t));
-        return { d, param.size, sc_gluon_u8_array, true };
+        return { d, param.size, sc_gluon_u8_array, 1 };
     }
     case sc_gluon_f64_array: {
         sc_gluon_data_v1 d {};
         d.f64_array = (double*)malloc(param.size * sizeof(double));
         memcpy(d.f64_array, param.data.f64_array, param.size * sizeof(double));
-        return { d, param.size, sc_gluon_f64_array, true };
+        return { d, param.size, sc_gluon_f64_array, 1 };
     }
     case sc_gluon_f32_array: {
         sc_gluon_data_v1 d {};
         d.f32_array = (float*)malloc(param.size * sizeof(float));
         memcpy(d.f32_array, param.data.f64_array, param.size * sizeof(float));
-        return { d, param.size, sc_gluon_f32_array, true };
+        return { d, param.size, sc_gluon_f32_array, 1 };
     }
     case sc_gluon_char_array: {
         sc_gluon_data_v1 d {};
         d.character_array = (char*)malloc(param.size * sizeof(char));
         memcpy(d.character_array, param.data.character_array, param.size * sizeof(char));
-        return { d, param.size, sc_gluon_char_array, true };
+        return { d, param.size, sc_gluon_char_array, 1 };
     }
     default: {
         return param;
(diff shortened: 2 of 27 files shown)


JordanHendersonMusic:
@dyfer oh master of the arcane CMake arts... 

I don't suppose you have any ideas how to get the path of a library inside of CMake? Do I need to 'install' it?

dyfer:
> I don't suppose you have any ideas how to get the path of a library inside of CMake? Do I need to 'install' it?

I think this is out of my depth... what exactly do you need? Please note that I have a very minimal understanding of the whole FFI mechanism.

JordanHendersonMusic:
Oh I just need to compile some code into a library, and get the path so I can pass it to a c test (an executable).

capital-G:
This is some truly awesome work Jordan! Maybe I can try porting #6615 to use the FFI? Would this help you?

JordanHendersonMusic:
It would, but maybe not just yet. There are probably some basic errors that I can catch with some simpler testing - I wouldn't want you to waste your time!

JordanHendersonMusic:
@dyfer  actually, don't worry, I've decided to go a different way for inbuilt testing.

JordanHendersonMusic:
@capital-G actually I think this is quite stable now, I still haven't got the windows builds working yet, and the apple CI builds aren't working either (but they *should* work).

capital-G:
Had a call with Jordan who walked me through the steps how to use this in a C++ project (thx for this), so I thought I'd share how to use the FFI in a bootstrap project.
I think this acts as a good intro for the documentation that Jordan wrote.

`CMakeLists.txt`
```cmake
cmake_minimum_required(VERSION 3.12)
project(sclang_websocket)
set(CMAKE_CXX_STANDARD 17)

# set file extension
set(CMAKE_SHARED_LIBRARY_SUFFIX ".gluon")
set(CMAKE_SHARED_LIBRARY_PREFIX "")

# make sure this is a SHARED library!
add_library(sclang_websocket SHARED
    src/helloWorld.cpp
)

target_include_directories(sclang_websocket PUBLIC
    ${SC_SRC_PATH}/include/lang
)
```

`src/helloWorld.cpp`
```cpp
#include "sc_ffi_v1.h"
#include "sc_ffi_version.h"

// stores our declarations we want to expose to sclang
sc_ffi_function_declarations_v1_t gDeclarations[1];

// the actual C-function we want to expose
sc_ffi_out_param_tag_v1 helloWorld(
    sc_ffi_library_data_v1_t libraryData,  // use this to pass along a potential state into the function
    sc_ffi_callable_object_v1_t maybeCallbackData,
    sc_ffi_param_v1_t* inParameters,
    uint32_t numInParameters,
    sc_ffi_out_param_or_maybe_diagnostic_v1* outParam  // the data that we return to sclang
)
{
    std::cout << "HELLO WORLD <3" << std::endl;

    // we are running inside the sclang vm here, so no blocking operations
    // but you can spawn a thread and use the callback functionality :) See Jordans example

    outParam->out_param.tag = sc_ffi_nil;  // choose appropriate return type from enum
    outParam->out_param.owns_data = false;
    outParam->out_param.size = 1;
    outParam->out_param.data.nil_ = {};  // this is how we represent nil

    return sc_ffi_produced_param;  // choose enum to tell the language if the op succeeded or not
}

// since we want to expose the functions we need to use the C context
extern "C" {

    // this function needs to be implemented
    uint32_t sc_ffi_version() { return 1; }

    // the sc_ffi_load_library function we need to implement
    sc_ffi_library_data_v1_t  sc_ffi_load_library(
        sc_ffi_do_callback_v1_f doCallbackFunction,
        sc_ffi_release_callback_object_v1_f releaseCallbackObject,
        sc_ffi_function_declarations_v1_t** const outDeclarations,
        uint32_t* outSize
    ) {
        // attach the declarations to our local copy
        *outDeclarations = gDeclarations;
        *outSize = 1;

        // specify the first declaration
        sc_ffi_function_declarations_v1_t& declaration = gDeclarations[0];
        declaration.accepts_callback = false;  // simple for now
        declaration.name = "helloWorld";  // name of the function that will be exposed on sclang side
        declaration.num_parms = 0;  // number of parameters our function will consume
        declaration.ptr = helloWorld;  // pointer to our c function

        std::cout << "FFI loaded our lib" << std::endl;

        return nullptr;
    }
}

```

Build, and then in sclang one can do

```supercollider
*** Welcome to SuperCollider 3.15.0-dev. *** For help type cmd-d.
sc3> ~y = FFILibrary.open("/Users/scheiba/github/sclang_websocket/build/libsclang_websocket.dylib");

FFI loaded our lib
-> a FFILibrary
sc3> ~y.helloWorld()

HELLO WORLD <3
-> nil
sc3> 
```

Worked a bit with it so far and I really like it!
No stackPointer arithmetic which can be screwed up, the callback functionality is really powerful and everything is easily debugable!
I am looking forward to use this for some nice things :)

capital-G:
Should we settle for a unified file name extension for this? I heard that using no file extension is actually problematic on linux machines since every needs to get inspected in a directory for server extensions, and we probably want to ship `sc` code along with a sclang extension.
I'll also take this language extension into account for the design of the package manager, currently called baryon.

We would need something from the atomic/elementary particle namespace. While we are at it, we can maybe also come up with an alternative name for server extensions? 

Looks like a bright future is ahead in sc land :)

JordanHendersonMusic:
We could call them gluons?

>A gluon (/ˈɡluːɒn/ GLOO-on) is a type of massless elementary particle that mediates the strong interaction between quarks, acting as the exchange particle for the interaction. Gluons are massless vector bosons, thereby having a spin of 1.[7] Through the strong interaction, gluons bind quarks into groups according to quantum chromodynamics (QCD), forming hadrons such as protons and neutrons.

Also, writing a library like this often involves some 'glue' code that links SC to the actual use case. I think it is kinda cute!

JordanHendersonMusic:
Fyi, I edited your example, you had a couple too many directory includes in the cmake.

capital-G:
I think I came across an error here.

When I tried to load the FFI within the `*initClass` like

```supercollider
WebSocket {
	classvar ffi;

	*initClass {
		ffi = FFILibrary.open(
			pathToLibrary: "/Users/scheiba/Library/Application Support/SuperCollider/Extensions/sclang_websocket/sclang_websocket.gluon"
		);
	}
}
```

 it fails with

```
compiling class library (debug build)...
	Found 882 primitives.
	Compiling directory '/Applications/SuperCollider-dev.app/Contents/Resources/SCClassLibrary'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/Extensions'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/Bjorklund'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/SafetyNet'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/ChordSymbol'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/ServerOptionsGui'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/MathLib'
	Compiling directory '/Users/scheiba/github/own/sc-grids'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/adclib'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/miSCellaneous_lib'
	Compiling directory '/Users/scheiba/github/DanDanSC'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/OSCMapper'
	Compiling directory '/Users/scheiba/github/JSONlib'
	Compiling directory '/Users/scheiba/github/sc-mackie'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/Wavesets'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/WavesetsEvent'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/JITLibExtensions'
	Compiling directory '/Users/scheiba/Library/Application Support/SuperCollider/downloaded-quarks/Influx'
	Compiling directory '/Users/scheiba/github/JITstage'
	Compiling directory '/Users/scheiba/github/nth'
	Compiling directory '/Users/scheiba/github/webRTCgui'
	Compiling directory '/Users/scheiba/supercollider/livecoding-karlsruhe'
	numentries = 2096132 / 38010584 = 0.055
	8332 method selectors, 4562 classes
	method table size 35758368 bytes, big table size 304084672
	Number of Symbols 21854
	Byte Code Size 766290
	compiled 989 files in 0.24 seconds

Info: 19 methods are currently overwritten by extensions. To see which, execute:
MethodOverride.printAll

compile done
FFI loaded our lib
ERROR: Message 'put' not understood.
RECEIVER:
   nil
ARGS:
   Integer -2147483647
Instance of FFILibrary {    (0x128144588, gc=20, fmt=00, flg=00, set=02)
  instance variables [1]
    id : Integer -2147483647
}
KEYWORD ARGUMENTS:
CALL STACK:
	Nil:handleError
		arg this = nil
	Thread:handleError
	Object:throw
	Object:doesNotUnderstand
		arg this = nil
		arg selector = 'put'
		arg args = [*2]
		arg kwargs = [*0]
	Meta_FFILibrary:open
		arg this = <instance of Meta_FFILibrary>
		arg pathToLibrary = "/Users/scheiba/Library/Appli..."
		var id = -2147483647
		var lib = <instance of FFILibrary>
	Meta_WebSocketClient:initClass
		arg this = <instance of Meta_WebSocketClient>
	Meta_Class:initClassTree
		arg this = <instance of Meta_Class>
		arg aClass = <instance of Meta_WebSocketClient>
		var implementsInitClass = nil
	ArrayedCollection:do
		arg this = [*423]
		var i = 36
	Meta_Class:initClassTree
		arg this = <instance of Meta_Class>
		arg aClass = <instance of Meta_Object>
		var implementsInitClass = nil
	Process:startup
		arg this = <instance of Main>
		var time = 365.7850105
	Main:startup
		arg this = <instance of Main>
		var didWarnOverwrite = false

^^ ERROR: Message 'put' not understood.
RECEIVER: nil

Messages with a similar name understood by the receiver:
	push
	post

Many other objects respond to the message 'put' (found 29 superclasses).

```

Loading the same file within the interpreter after all class library setup works flawlessly.

```supercollider
(
~y = FFILibrary.open(
	pathToLibrary: "/Users/scheiba/Library/Application Support/SuperCollider/Extensions/sclang_websocket/sclang_websocket.gluon"
);
)
FFI loaded our lib
-> a FFILibrary
```

JordanHendersonMusic:
I think you have to init the FFI library class before using it in initClass.

JordanHendersonMusic:
A wider question, is how can we distribute these? I don't really know all that much about this. But I guess you could put the compiled library inside of a quark's repository and distribute the library binary like that.

Is security a concern here? Could/should we use a checksum to ensure you have downloaded the right library? You could be required to pass a checksum value to `Gluon.load`.

capital-G:
> I think you have to init the FFI library class before using it in initClass.

Right, thanks - I was looking for it in the documentation but couldn't find it in https://docs.supercollider.online/Guides/WritingClasses.html, only found it in https://docs.supercollider.online/Classes/Class.html#*initClassTree after you gave a hint - maybe this could be added to the writing class guide? What do you think @supercollider/documentation?

> A wider question, is how can we distribute these? I don't really know all that much about this. But I guess you could put the compiled library inside of a quark's repository and distribute the library binary like that.
> 
> Is security a concern here? Could/should we use a checksum to ensure you have downloaded the right library? You could be required to pass a checksum value to `Gluon.load`.

I think that would responsibility of a package manager - you shouldn't/wouldn't want to distribute binaries via git, especially if they come in a wide variety of architectures/os/ABI. (e.g. the websocket gluon is around 14 mb - times 4 platforms times each release results in a big repo really soon).
I'll try to finish the package manager before mid-December, b/c it would be really nice to have as well for the 3.15 server plugin API change.

---

Do you think it would make sense to provide a higher level C++ API as well? This wouldn't need to be as stable as the C ABI as we could change the interface structure w/o changes in the resulting binary format. Or maybe we wait til we go for C++20, since constexpr could be nice for templating such an API?

JordanHendersonMusic:
> I'll try to finish the package manager before mid-December, b/c it would be really nice to have as well for the 3.15 server plugin API change.

That would be awesome, perhaps share what you have so far at the next dev meeting, it seems to always be last on the agenda!?

> Do you think it would make sense to provide a higher level C++ API as well? 

There is no reason why we couldn't provide one for a variety of languages. I think that should go in a different pr, but I don't mind making a few. 

Usually, FFI language bindings are in separate repos from the rest of the code base. This avoids introducing new language into the main repo and allows for relaxed rules around this like c++ versions.

dyfer:
This is amazing work Jordan!

> Should we settle for a unified file name extension for this?

my 2c (from the forum chat):
"Gluons" is a really cute name and I like it _as a name itself_.
However, I'm generally confused by new names that are not self-explanatory, like ruby's "gems".
I'd personally prefer a generic term like "language plugins", but I'm happy to go along with "gluons" for "more character" if others prefer that.

Also, I think musicians are likely to be familiar with VST/AU/other plugins, but I'm guessing they are less likely to be familiar with elementary particles and making a connection between them and the SC world... it is quite poetic though.

JordanHendersonMusic:
@dyfer I don't mind changing it back, however, it probably isn't a good idea to distribute these without some class to do the initialisation, meaning they would be a part of the quark. They aren't really something the average user would see, it is more something an implementer would use.

dyfer:
> meaning they would be a part of the quark.

Ah, okay. So we wouldn't really have Quarks and Language Plugins (+ associated classes) separation? Originally I thought they would be separate (in naming convention) but maybe that's not a good idea.

(For the record, I personally would prefer "Sclang extensions" instead of "quarks", but there's no point changing that).

JordanHendersonMusic:
Essentially every library will have an initialisation phase, which you'd probably want to do in a class constructor, so while you could have them as a separate thing to download, I don't think any one will do that.

The distinction between gluons and quarks is something a library author would care about, the user would never even see the term 'gluon'.

@capital-G this is why I wondered about how we could distribute these with the quarks, because they seem like an addition to that system in my mind.

capital-G:
> (For the record, I personally would prefer "Sclang extensions" instead of "quarks", but there's no point changing that).

But those are actually class library extensions, not sclang extensions ;)

> @capital-G this is why I wondered about how we could distribute these with the quarks, because they seem like an addition to that system in my mind.

Yeah - I am starting to think it may be better to get rid of distinctions like quark vs server extension and just have one package format which can contain one or more of

* server extensions / ? / compiled
* class library extensions / quarks / source code
* sclang extensions / gluon / compiled

So with this new package manager/format the user wouldn't need to care about distinction of quark/gluon/..., it would only become a nomenclature within the SC world.

dyfer:
> > (For the record, I personally would prefer "Sclang extensions" instead of "quarks", but there's no point changing that).
> 
> But those are actually class library extensions, not sclang extensions ;)

I meant sclang as in supercollider language, not the interpreter, but I guess that's inaccurate :) yes, of course I meant classlib extensions. 

BTW I find sclang to be a bit confusing as a name, because it can refer to both the interpreter as well as the supercollider language (syntax). I wonder if there's a way to use a clearer naming scheme. 

> it probably isn't a good idea to distribute these without some class to do the initialisation, meaning they would be a part of the quark.

Server extensions are also distributed with classes. My initial thought was that we'd have 3 separate families of extensions: quarks, server plugins (shipped with necessary classes, as they are currently), and language extensions (also shipped with classes).

However:

> Yeah - I am starting to think it may be better to get rid of distinctions like quark vs server extension and just have one package format which can contain one or more of
> 
> * server extensions / ? / compiled
> * class library extensions / quarks / source code
> * sclang extensions / gluon / compiled

I think this is the best way forward. I remember new users being very confused as to the difference between server plugins and quarks ("I installed sc3-plugins but don't see functionality xyz..." - "you need a _quark_ for that" - "???" or vice versa) and I believe a single concept of "supercollider extension" would be better. If/once we get the package manager, I think it should handle all extensions in one place (however we call them), so there's a single place to get e.g. jitlib extensions, Tartini pitch tracker, and ... the upcoming websockets language extension?

include/gluon_ffi_interface/sc_gluon_v1_entry_points.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include "sc_gluon_v1_types.h"
+
+#ifdef __cplusplus
+#    define SC_GLUON_EXTERN extern "C"
+#endif
+
+#define SC_GLUON_EXPORT SC_GLUON_EXTERN __attribute__((__visibility__("hidden")))
```
capital-G: Shouldn't this be public/default?


```suggestion
#define SC_GLUON_EXPORT SC_GLUON_EXTERN __attribute__((__visibility__("default")))
```

JordanHendersonMusic: yes, thanks!

capital-G:
I ported #6615 to use this gluon ffi interface into a dedicated repo https://github.com/capital-G/sclang_websocket (still wip).
The relevant glue code can be found at https://github.com/capital-G/sclang_websocket/blob/main/src/ws_gluon.cpp - with some C++ magic it is probably possible to drastically reduce the amount of code since there is lots of boilerplate r/n.

I think the approach of this FFI is really really good b/c as an extension developer I don't have to care about stack pointers (or screw them up...), it is clear how to pass data or events between my extension and sclang runtime, and I don't have to understand all the finesse of the GC. So it is much easier and safer.
The ability to provide callbacks is really powerful and makes it really easy to resolve async events in the FFI and pass its results back into sclang (e.g. the event of receiving a message within a websocket connection).

The only limitation I have encountered was that it is only possible to provide one callback function.  At the same time this did not require any compromises in functionality, but required a specific design for the FFI and how to init/call it - but I think this is fine?
E.g. when a websocket server accepts a new connection I want to register the new connection in sclang and also register the callbacks `onMessage` and `onConnectionStateChange` - since these are time critical events and can happen very shortly after a new connection has been accepted.
I [called the `listenerNewConnectionCallback` from C++](https://github.com/capital-G/sclang_websocket/blob/145890624265d00527fbbddc315a368d1359a99b/src/ws_gluon.cpp#L290-L298) which [creates a new WebSocketConnection object on the sclang side](https://github.com/capital-G/sclang_websocket/blob/145890624265d00527fbbddc315a368d1359a99b/src/Classes/WebSocket.sc#L56-L64) and [registers the callbacks in its `init` method](https://github.com/capital-G/sclang_websocket/blob/145890624265d00527fbbddc315a368d1359a99b/src/Classes/WebSocket.sc#L87-L90).
If in the meantime I already got a connection reset or a message I wouldn't be able to capture those events from within sclang as the heap allocation of a new connection happens?

---

tldr: We can start to extend sclang w/o making it part of the sclang runtime - which is probably the single most interesting thing of sclang development since splitting server and language!

JordanHendersonMusic:
Thanks glad you like it!

> ... and I don't have to understand all the finesse of the GC

> The only limitation I have encountered was that it is only possible to provide one callback function. At the same time this did not require any compromises in functionality, but required a specific design for the FFI and how to init/call it - but I think this is fine?

So I've been thinking about changing this so you can pass callbacks as normal parameters, this issue is that it complicates the gc and makes things more error prone. 

First we'd have to move to manually capturing the callback object in the library code, but since you could have nested `param_arrays` of callback objects, you'd have to walk this structure to capture them all, and make sure you release them. I think this just complicates things.

Instead... what you could do is pass a single callback and call it with an id indicating what has changed, then do some more dispatching in sc lang. 

I'm also going to write a little cpp helper interface for this because it is a little complicated to construct the param objects.

JordanHendersonMusic:
So I've added a ctest which creates a library and loads it as a part of the CI.

There is also a few cpp helper functions to make the creation of the parameters easier. I tried to make a nice class interface, but it is just too complex because the parameters may or may not own the data, and trying to avoid copies just becomes hard work.

include/gluon_ffi_interface/sc_gluon_v1_types.h
```diff
@@ -0,0 +1,222 @@
...
+    union sc_gluon_out_param_or_maybe_diagnostic_v1* out_param);
+
+
+//////////// PARAMS
+
+struct sc_gluon_function_declarations_v1_t {
+    const char* name;
+    sc_gluon_function_v1_f ptr;
+    int32_t num_parms; // negative means accepts any number of arguments.
+    bool accepts_callback;
```
JordanHendersonMusic: Change bool to uint33_t

include/gluon_ffi_interface/sc_gluon_v1_types.h
```diff
@@ -0,0 +1,222 @@
...
+    //
+    sc_gluon_param_array,
+};
+
+union sc_gluon_data_v1 {
+    int32_t nil_;
+    int32_t i32;
+    double f64;
+    char character;
+    bool boolean;
```
JordanHendersonMusic: No bool

include/gluon_ffi_interface/sc_gluon_v1_types.h
```diff
@@ -0,0 +1,222 @@
...
+    float* f32_array;
+    char* character_array;
+    struct sc_gluon_param_v1_t* param_array;
+};
+
+struct sc_gluon_param_v1_t {
+    union sc_gluon_data_v1 data;
+    uint32_t size; // For non-array parameters, this is ignored, but set to '1' as a good practice.
+    enum sc_gluon_param_tag_v1 tag;
+    bool owns_data; // Indicates that the param owns some heap allocated data, and will be deallocated when the lifetime
```
JordanHendersonMusic: No bool

JordanHendersonMusic:
@capital-G I've just broken your websocket implementation — sorry! 

Now the load_library function accepts parameters, meaning you don't need an 'init' function for the library, hopefully this reduces some of the boilerplate.

capital-G:
> @capital-G I've just broken your websocket implementation — sorry!
> 
> Now the load_library function accepts parameters, meaning you don't need an 'init' function for the library, hopefully this reduces some of the boilerplate.

np, I enjoy providing some feedback on the iteration by providing a real-world example. I hope I'll come around to adjusting it this weekend.

BTW - I noticed that the FFI branch breaks the SHM interface on macOS and produces some errors during interpreter start

```
compiling class library...
ERROR: objc[93998]: Class QT_ROOT_LEVEL_POOL__THESE_OBJECTS_WILL_BE_RELEASED_WHEN_QAPP_GOES_OUT_OF_SCOPE is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4710) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c710). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
ERROR: objc[93998]: Class KeyValueObserver is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4738) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c738). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
objc[93998]: Class WeakPointerLifetimeTracker is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4788) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c788). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
objc[93998]: Class RunLoopModeTracker is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e47d8) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c7d8). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
objc[93998]: Class QDarwinPermissionHandler is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4850) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c850). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
objc[93998]: Class QMetalLayer is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtGui.framework/Versions/A/QtGui (0x102dbcb28) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtGui.framework/Versions/A/QtGui (0x1180f8b28). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
```

and

```
Cannot connect to shared memory: Cannot connect to shared memory
ERROR: Primitive '_ServerShmInterface_connectSharedMem' failed.
Failed.
RECEIVER:
Instance of ServerShmInterface {    (0x178012678, gc=B8, fmt=00, flg=00, set=02)
  instance variables [2]
    ptr : nil
    finalizer : nil
}

PROTECTED CALL STACK:
	Meta_MethodError:new	0x138646800
		arg this = PrimitiveFailedError
		arg what = Failed.
		arg receiver = a ServerShmInterface
	Meta_PrimitiveFailedError:new	0x13864e880
		arg this = PrimitiveFailedError
		arg receiver = a ServerShmInterface
	Object:primitiveFailed	0x13817b340
		arg this = a ServerShmInterface
	Server:connectSharedMemory	0x139550f80
		arg this = localhost
		var id = 57110
	Server:bootInit	0x139554100
		arg this = localhost
		arg recover = false
	a FunctionDef	0x139552440
	a FunctionDef	0x139588a40
	Routine:prStart	0x138e9c380
		arg this = a Routine
		arg inval = 84.3443625

CALL STACK:
	MethodError:reportError
		arg this = <instance of PrimitiveFailedError>
	< closed FunctionDef >
		arg error = <instance of PrimitiveFailedError>
	Function:try
		var result = <instance of PrimitiveFailedError>
	< FunctionDef in Method Scheduler:seconds_ >
		arg time = 86.227410709
		arg item = <instance of Routine>
	< FunctionDef in Method SequenceableCollection:pairsDo >
		arg i = 0
	Integer:forBy
		arg this = 0
		arg endval = 0
		arg stepval = 2
		var i = 0
		var j = 0
		arg this = [*2]
	Scheduler:seconds_
		arg newSeconds = 86.227441709
	Meta_AppClock:tick
	Process:tick
		arg this = <instance of Main>
^^ The preceding error dump is for ERROR: Primitive '_ServerShmInterface_connectSharedMem' failed.
Failed.
RECEIVER: a ServerShmInterface
```

I am somehow sure that I don't get these problems on normal dev branch, but I am on a 1 week old version of this branch.

lang/LangSource/GluonFFIV1.cpp
```diff
@@ -0,0 +1,614 @@
...
+#ifdef _WIN32
+#    include "SC_Win32Utils.h"
+#    include "SC_Codecvt.hpp"
+#else
+#    include <dlfcn.h>
+#endif
+
+using namespace sc_gluon;
+
+sc_gluon_param_v1_t slot_to_param(PyrSlot slot);
```
JordanHendersonMusic, line 30: Self referring arrays will cause an infinite loop, throw an error.

dyfer:
> BTW - I noticed that the FFI branch breaks the SHM interface on macOS and produces some errors during interpreter start
> 
> ```
> compiling class library...
> ERROR: objc[93998]: Class QT_ROOT_LEVEL_POOL__THESE_OBJECTS_WILL_BE_RELEASED_WHEN_QAPP_GOES_OUT_OF_SCOPE is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4710) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c710). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
> ERROR: objc[93998]: Class KeyValueObserver is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4738) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c738). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
> objc[93998]: Class WeakPointerLifetimeTracker is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4788) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c788). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
> objc[93998]: Class RunLoopModeTracker is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e47d8) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c7d8). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
> objc[93998]: Class QDarwinPermissionHandler is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtCore.framework/Versions/A/QtCore (0x1036e4850) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtCore.framework/Versions/A/QtCore (0x11879c850). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
> objc[93998]: Class QMetalLayer is implemented in both /Applications/SuperCollider-dev.app/Contents/Frameworks/QtGui.framework/Versions/A/QtGui (0x102dbcb28) and /opt/homebrew/Cellar/qt/6.9.1/lib/QtGui.framework/Versions/A/QtGui (0x1180f8b28). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
> ```

IIUC these are not related to FFI or other SC code itself - IIUC this is unavoidable if we ship Qt libs inside the App bundle when  the user has also Qt installed with homebrew. I _think_ that if you unlink (or maybe uninstall...) homebrew's qt, these should go away. Also, I don't think these are related to the SHM interface, but I may be wrong.

JordanHendersonMusic:
Some notes for myself from the dev meeting — thanks @Spacechild1 :
1.This raises issues with symbol collisions with our libraries, what do we export? Should be nothing.
2. You can't free memory across the boundary, fix this!

Spacechild1:
First, I want to applaud @JordanHendersonMusic for taking on this very important work!

I have already made some comments on the dev meetings, but I thought it would be good to write it down. I'm very busy till the end of November, but then I will have more time again to look at these things.

To give some background: I have extensively worked with the Lua C API and recently also with the Java JNI. I also dabbled in the Python C API and various other scripting language APIs. I have to admit that I find the proposed Gluon API a bit idiosyncratic compared to the ones I've used so far. This isn't necessarily a criticism, just an observation.

There are two things that I immediately noticed:

---

I think what we *actually* need is a C API around the primitive system. As a Class author, I don't want to be responsible for dynamically loading a module; this should be done automatically by the VM, similar to Server plugins.

I just want to register a few custom primitives in the entry point of my module, so I can use them in my Class(es):
```sc
Foo {
    bar { |baz|
        _Foo_bar
        ^this.primitiveFailed
    }
}
```

Loading C modules at runtime is neat, but do we really need this?

---

What I also noticed is the use of tagged unions for function arguments and return values. Personally, I would rather use API functions for accessing function arguments and returning values. I think these are more flexible and extendable. They also help to keep memory allocations on the sclang side. (As I've mentioned in the dev meeting, you can't `malloc()` something on one side and `free()` it on the other side of the plugin boundary.) Finally, they can often be implemented without marshalling, i.e. by directly working with actual VM objects.

Just to give an example for such a (fictional) API:
```c
sc_error_t my_function(sc_env* env, void* user)
{
    // check number of arguments
    size_t arg_count = env->get_argument_count();
    if (arg_count < 3) { return sc_error_bad_argument; }
    
    // get first argument as integer
    sc_object_t obj = env->get_argument(0);
    int int_arg = env->obj_is_integer(obj ) ? env->obj_as_integer(obj ) : 0;
    
    // get second argument as string
    obj = env->get_argument(1);
    const char* str_arg = env->obj_is_string(obj) ? env->obj_as_string(obj);
   
    // third argument is an array
    obj = env->get_argument(2);
    if (env->obj_is_array(obj)) {
        sc_array_t array = env->obj_as_array(obj);
        size_t size = env->get_array_size(array);
        // iterate over array
        for (size_t i = 0; i < size; ++i) {
            sc_object_t slot = env->array_get_slot(array, i);
            double f = env->obj_is_float(slot) ? env->obj_as_float(slot) : 0.f;
        }
    }

    // we might even support the new kwargs feature (if possible):
    size_t kwargs_count = env->get_kwargs_count();
    for (size_t i = 0; i < kwargs_count; ++i) {
        const char* env->get_kwargs_name(i);
        sc_object_t env->get_kwargs_value(i);
        // etc.
    }

    // return an array of 3 floats.
    // Note how we don't have to malloc() or free() any objects!
    // Objects are allocated with API functions and ownership is transferred to sclang.
    sc_array_t array = env->create_array(3);
    env->array_set_slot_float(array, 0, 10.0);
    env->array_set_slot_float(array, 1, 20.0);
    env->array_set_slot_float(array, 2, 30.0);
    env->return_object(array);

    return sc_error_none;
}
```

Notice how simple the signature a C function is! It just takes a pointer to the VM environment (`env`) and a pointer to user data. This is very similar to a C function in Lua:
```c
typedef int (*lua_CFunction) (lua_State *L);
```
All the actual complexity resides in the API functions.

There are generally two approaches for making API functions available to a C module:
1. export regular C functions from a DLL. Examples: Lua, Python, Pd, Max/MSP
2. pass a table of function pointers. Examples: JNI, SC server plugin API, REAPER extensions

---

I would suggest anyone who is interested in this topic to have a look at the Lua C API (https://www.lua.org/pil/contents.html#P4) and the JNI (https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#interface-function-table). They both work very well, but they make very different design decisions. I think a lot can be learned from these.

Just to be clear: I absolutely don't say that we have to follow the design of the Lua C API or the JNI or the (fictional) API I've posted above! I just think that it's important to evaluate different API designs before we settle on a particular implementation.

---

One thing I definitely have to think about is the whole callback business. But I can't really do that before December.

JordanHendersonMusic:
Thanks for taking the time to write this! I'll have a little rethink and get back with an update. 

Just quickly...

1. I think passing an env argument is a good idea
2. I didn't consider that the host and library might be using different malloc and free --- having an env argument solves this issue
4. I think loading it as a module makes managing the state a lot easier, particularly since you'd have to free anything from the library on a SC hook (recompile class library), and you could easily be left with dangling pointers there. For the use case of just wanting a pure function, it is a bit of overhead though.

I think we will need a `block for callback` function in which you can allocate args and trigger the callback (from a helper thread). As the GC needs to be paused well you do this.

For now, that means this PR is going to change, so no one else should spend too much time reviewing the fine details.

I did try looking into Lua's implementation but only found luaFFI, which is a part of luaJIT.... so maybe this should just be called the `sc_c_api`?

dyfer:
Not sure if this is relevant, but as I was looking through old branches in the repo, by chance I came across @scztt 's test implementation of an ffi...? https://github.com/supercollider/supercollider/commit/2eeb23005ad5b64a89a812d809920a4f2fb3f24f
Linking it here just in case it's useful in any way.

Spacechild1:
Just a quick reply:

> I think loading it as a module makes managing the state a lot easier, particularly since you'd have to free anything from the library on a SC hook (recompile class library)

For global state, this could be done in the module's exit point, similar to the `unload` function in SC server plugins. When you recompile the Class Library, you can simply reinitialize all loaded modules by first calling the exit function and then again the entry function. You don't actually have to unload the DLL/SO itself. (Although it can be useful to have an option to unload plugin modules before recompiling the Class Library, especially during plugin development.)

The more interesting question is how to reliably release individual resources, e.g. a network socket. Unfortunately, sclang does not have user-defined finalizers and we typically have to manually call `free` methods. *I think this issue applies to any C API for sclang.*

The simple solution would be to require plugins to keep track of resources in their global state, so they have to release all remaining resources in the module's exit function.

The more involved solution would be to add support for user defined finalizers, which could call into primitives.

> I did try looking into Lua's implementation but only found luaFFI, which is a part of luaJIT....

Just for reference, here's how you register a C function resp. write a C module: https://www.lua.org/pil/26.2.html

> so maybe this should just be called the sc_c_api?

Or just `sclang_*` :)

Spacechild1:
> I think we will need a block for callback function in which you can allocate args and trigger the callback (from a helper thread). As the GC needs to be paused well you do this.

I will just quickly outline how this can be done in a Lua-style API.

Let's assume I want to write a C function that starts a thread and then posts one or more messages back to the main thread. This is how I would do it in Lua:

1. pin the callback function resp. receiving object by creating a "reference" (see https://www.lua.org/pil/27.3.2.html)
2. store the "reference" and a copy of the arguments and send it to the thread
3. in the thread, post a message with the result back to the event loop
4. when the message is dispatched, get the receiver from the "reference" and call it with the result
5. release the "reference" when it's no longer needed

---

We could follow a similar approach in sclang. Here's what we would need for that:

- Language APIs typically offer some way to (temporarily) pin objects to prevent them being garbage-collected. In Lua that would be the "reference" system (see above). In JNI this can be done with the `NewGlobalRef` and `DeleteGlobalRef` functions (https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#newglobalref). Let's imagine two API functions:
  ```c
  sc_object_t arg = env->get_argument_object(0);
  sc_object_t ref = env->pin_object(obj);
  // ...
  env->unpin_object(ref);
  ```

- I would propose something like a `postMessage` API function which would just pass an arbitrary object to the main scheduler thread:
  ```c
  typedef sc_error_t (*sc_callback_t)(sc_env*, void* message, sc_bool cancelled)

  // in function table:
  sc_error_t (*postMessage)(void *message, sc_callback_t func, void *token);
  sc_error_t (*cancelMessage)(void *token);
  ```

  <details><summary>Side note</summary>

  In fact, this would also be handy for some built-in primitives. For example, incoming UDP packets and MIDI messages are currently dispatched by locking the interpreter and sending a message with `runInterpreter`. This has two drawbacks:
  - the UDP or MIDI message has to wait for the global interpreter lock
  - the network/MIDI thread is blocked for the whole duration of `runInterpreter`, which may take an arbitrary amount of time.

  It's also important to point out that locking the interpreter on a secondary thread **can deadlock** when a primitive wants to join that thread.
  
  `postMessage`, on the other hand, would just post a message to a queue and notify the main scheduler thread. The scheduler would then pop messages from the queue and invoke its callback functions. In the meantime, the network/MIDI thread can continue to receive messages.

  </details>

- In the message callback you can then send a message to some object. Let's imagine the following API functions:
  ```c
  // push receiver
  env->push_object(receiver);
  // push three arguments
  env->push_integer(1);
  env->push_float(3.0);
  env->push_string("foo");
  // call method
  env->send_message("someSelector");
  ```

- Finally, we need a way to get/return opaque pointers. In Lua these are called light userdata (https://www.lua.org/pil/28.5.html). In our case that would be `get_argument_pointer`, `array_set_slot_pointer`, `array_get_slot_pointer` and `return_pointer`.

---

Now I will show how these building blocks can be pieced together in a minimal (fictional) example. Let's assume we want to write a `MySensor` class to interface with some sensor. The `open` method should open the sensor on a given COM port, after which we periodically receive sensor values on a dedicated thread. We want to forward these sensor values to the instance method `prHandleData` on the main scheduler thread. The `close` method should close the sensor and cancel any pending messages.

```c
// context for our class and the sensor's thread function
struct MySensorContext {
    sc_env_t env;
    sc_object_t self;
    sensor_t sensor;
};

// sensor data sent from the sensor thread to the scheduler thread
struct MySensorData {
    sc_object_t self;
    float x, y, z;
};

// callback for postMessage() API function. This is called on the main scheduler thread.
void MySensor_handleData(sc_env_t *env, void *msg, sc_bool cancelled)
{
    if (!cancelled) {
        // call "prHandleData" instance method with "x", "y" and "z" arguments
        auto data = (MySensorData*)msg;
        env->push_object(data->self);
        env->push_float(data->x);
        env->push_float(data->y);
        env->push_float(data->z);
        env->send_message("prHandleData");
    }
    delete msg;
}

// callback for fictional sensor library. This is called on the sensor's receive thread.
void receiveSensorData(const sensor_data_t* data, void *user)
{
    auto ctx = (MySensorContext*)user;
    // send sensor data to main scheduler thread
    auto msg = new MySensorData;
    msg->self = ctx->self;
    msg->x = data->x;
    msg->y = data->y;
    msg->z = data->z;
    // use 'ctx' as token so we can cancel the message, see MySensor_close().
    env->postMessage(ctx->env, msg, MySensor_handleData, ctx);
}

// _MySensor_open primitive
sc_error_t MySensor_open(sc_env_t *env, void *user)
{
    // get strong reference to the receiver
    sc_object_t self = env->pin_object(env->get_receiver());
    // first argument is COM port
    if (!env->get_argument_type() != sc_type_integer) { return sc_error_bad_argument; }
    int port = env->get_argument_integer(0);
    // create and start sensor
    auto ctx= new MySensorContext;
    ctx->env = env;
    ctx->self = self;
    ctx->sensor = sensor_create();
    // let's assume this starts an internal thread and 'receiveSensorData' is called for every sensor message
    ctx->sensor->start(port, receiveSensorData, ctx);
    // return context as opaque pointer. The caller stores the pointer so it can later use it to close the sensor.
    env->return_pointer(ctx);
    return sc_error_none;
}

// _MySensor_close primitive
sc_error_t MySensor_close(sc_env_t *env, void *user)
{
    // get context from first argument
    auto ctx = (MySensorContext)env->get_argument_pointer(0);
    // close and destroy sensor. This also joins the receive thread.
    sensor_destroy(ctx->sensor);
    // Now we can cancel all pending messages.
    // Note that we have passed 'ctx' as our token, see receiveSensorData().
    env->cancelMessage(ctx);
    // unpin message receiver
    env->unpin_object(ctx->self);
    delete ctx;
    return sc_error_none;
}
```

In the example above, we forward the sensor data to a private instance method "prHandleData", but this could also be a user provided callback object. In that case, we would just have to send the "value" message instead.

---

Again, I just want to show how things could look with a more traditional C API. Unfortunately, I can't really participate in a deeper discussion until December. I just wanted to get these things off my head so I can go back and focus on the stuff I actually have to do :-D I think there's quite a bit of material and ideas to digest. Happy to discuss further in December!

JordanHendersonMusic:
So the plan is to first rework the threading situation, using some kind of message posting as @Spacechild1 has suggested. Then implementing the C API will be much simpler. This is likely a 3.16 change.

PR #7190, opened 2025-10-18
[build] guard homebrew deployment fix

dyfer:
## Purpose and Motivation

It seems that the deployment in macdeployqt shipped with homebrew has been fixed...

~~Something to consider is whether we try to check for exact Qt version to apply the fix... But homebrew generally pushes for latest versions so maybe it's not worth it.~~ I changed this so that it only applies to Qt6 lower than 6.9.3. It's still a workaround, but it should work fine for both older and current Qt installs from homebrew.

~~I haven't tested this locally yet.~~

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/editors/sc-ide/CMakeLists.txt
+++ b/editors/sc-ide/CMakeLists.txt
@@ -494,7 +494,7 @@ if(APPLE OR WIN32)
             ")
         endif()
         set(VERIFY_CMD "include(BundleUtilities)")
-        if("${QT_VERSION_MAJOR}" GREATER_EQUAL 6 AND SC_USE_QTWEBENGINE)
+        if("${QT_VERSION_MAJOR}" GREATER_EQUAL 6 AND "${Qt6_VERSION}" VERSION_LESS "6.9.3" AND SC_USE_QTWEBENGINE)
             # Fix deployment with qt from homebrew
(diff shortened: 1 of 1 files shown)


dyfer:
@capital-G do you think this is okay? It's a fragile workaround one way or another, but this seems to work for both older and newer Qt from homebrew.

JordanHendersonMusic (approved):

JordanHendersonMusic:
This is the fix for the following error right?
```
CMake Error at editors/sc-ide/cmake_install.cmake:93 (file):
  file COPY cannot duplicate symlink

    /opt/homebrew/opt/qt/lib/QtQuickWidgets.framework

  at

    /Users/runner/work/supercollider/supercollider/build/Install/SuperCollider/SuperCollider.app/Contents/Frameworks/QtQuickWidgets.framework

  because: A directory already exists at that location
  editors/cmake_install.cmake:44 (include)
  cmake_install.cmake:63 (include)


Command PhaseScriptExecution failed with a nonzero exit code
```

In that case, I'm in favour of merging this even if it is a bit hacky, as it will allow me to continue testing my FFI implementation.

dyfer:
> This is the fix for the following error right?

Yes it is

dyfer:
I'll merge this so that the CI passes again.

Issue #7192, opened 2025-10-18
`.interpret` should optionally respect local scope


prko:
## Motivation
Currently, `.interpret` evaluates a string in the current environment, but it does not have access to local variables or function arguments.  
This makes it cumbersome to use `.interpret` inside functions, because one must assign values to global variables as a “bridge”:
```supercollier
(
y = "[	
cos(x),
cos(-pi/2 + x),
sin(x),
sin(pi/2 + x)
]";
(0, 0.1 .. 2pi).collect{ |i| // a bridge element
	x = i; // this is a bridge between y and (0, 0.1 .. 2pi)
	y.interpret
}.flop.plot("Phase, Cosine, Sine", Window.availableBounds)
.axisLabelX_(
	y[1..(y.lastIndex-1)]
	.split($,)
	.collect{ |code| code.asString.replaceRegexp("\n", "") }
)
.setProperties(\labelFont, (Font().size_(64)));
)
```
## Description of Proposed Feature
Introduce an optional argument (e.g. `\local`) to `.interpret` so that it can evaluate the string in the local scope of the function:
```supercollider
(
y = "[	
cos(x),
cos(-pi/2 + x),
sin(x),
sin(pi/2 + x)
]";
(0, 0.1 .. 2pi).collect{ |x| // x instead of y
	y.interpret(\local)// \local lets sclang to interpret this line in the scope of the local function.
}.flop.plot("Phase, Cosine, Sine", Window.availableBounds)
.axisLabelX_(
	y[1..(y.lastIndex-1)]
	.split($,)
	.collect{ |code| code.asString.replaceRegexp("\n", "") }
)
.setProperties(\labelFont, (Font().size_(64)));
)
```
BY doing this:
- Removes the need for awkward global “bridge” assignments.
- Makes .interpret more useful for educational examples.
- Keeps code cleaner and less error-prone.

## Plan for Implementation

It is unclear whether this is feasible within sclang’s current compiler and scoping design. If possible, I would be interested in exploring an implementation. I would appreciate guidance from core developers.

jamshark70:
I don't think there will be any way to make local `var`s accessible within other compilation scopes.

```
var abc = 1;

"{ abc.postln }".interpret;
```

... is not the same as

```
var abc = 1;

{ abc.postln }  // open function
```

... because, in the second case, it's one invocation of the compiler (so `abc` exists everywhere within this compiler invocation's scope). In the first case, interpret` is a second invocation of the compiler:

[...]

Issue #7164, opened 2025-10-02
Developer meetings


capital-G:
Developer meetings are held at irregular intervals, usually monthly, where open PRs, issues, and any other developer or organizational-related topics are discussed. Everyone is welcome to these meetings, regardless of seniority as the only requirement is curiosity. These meetings also serve as an good opportunity for newcomers to get started in the development of SuperCollider.
The content of the meeting is what the people present want to discuss - so it is a good place to ask how to move forward with a stuck PR or ask how to approach a certain problem. Additionally, we will also use the meeting to tackle bigger topics such as the organization of the next release.

It is also possible to add issues or PRs to the https://github.com/orgs/supercollider/projects/8 project.

The date for the next developer meeting is voted via <https://scsynth.org/t/development-meetings-scheduling-polls>. After the vote is finalized, the date of the next meeting is announced via <https://scsynth.org/t/development-meetings-schedule>.

If you don’t have a user account in the forum but still want to be informed about the next developer meeting via GitHub, it will also be announced via this issue. You can subscribe to this issue by using the Notifications menu on the right of the issue to receive notifications about the next meeting via GitHub.
This is a locked GitHub issue, which is only used to communicate the date of the next developer meeting such that it is fine to subscribe to this issue.

Additionally, the developer community uses the Discourse forum chat at <https://scsynth.org/chat/c/supercollider-development/45> instead of Slack, Discord, etc., which acts as a fast way of communication and can also be helpful for small questions. Access to this chat requires a forum user account.

capital-G:
Next dev meeting is @ [2025-10-06 18:00 - 20:00 UTC](https://www.inyourowntime.zone/2025-10-06_20.00_Europe.Berlin) via <https://video.imm.rsh-duesseldorf.de/sc-dev-meeting>

tedmoore:
Just a heads up. I'll likely be a bit late.

capital-G:
Next dev meeting is @ [2025-11-09 18:00 UTC](https://www.inyourowntime.zone/2025-11-09_19.00_Europe.Berlin) via <https://video.imm.rsh-duesseldorf.de/sc-dev-meeting>

capital-G:
Next dev meeting is @ [2025-12-07 20:00 UTC](https://www.inyourowntime.zone/2025-12-07_21.00_Europe.Berlin) via <https://video.imm.rsh-duesseldorf.de/sc-dev-meeting>

capital-G:
[...]

PR #7203, opened 2025-10-24
sclang: rework name resolution 

JordanHendersonMusic:
This has become a bigger change than expected.

Essentially the issue was that while we can emit errors, a lot of the code actually makes sense, it just differs from what supercollider is currently doing (which is confusing and wrong).

For that reason, there are breaking changes in this PR, namely, you cannot have inline mutually recursive functions.

```supercollider
var f = {g.()};
var g = {f.()};
```

After this PR (and from version 3.16), this will have to be...
```supercollider
var f, g;
f = {g.()};
g = {f.()};
``` 

Please also see the new language specification help documentation I have made, it is just a start, but will be really useful to have some informal descriptions of what the language should do, rather than relying on what it happens to do.

## Purpose and Motivation

See the opening of Lucile Nihlen's keynote. @lnihlen.

https://www.youtube.com/watch?v=SM1nnjeHluI

TLDR:
```supercollider
var f = {
	arg a1 = x1;
	var x1 =2 + 2;
	a1
}; // returns nil, not 4.

var f = {
	arg a1 = x1;
	var x1 = 4;
	a1
} // returns 4
```

Here is another one that interacts with interpreter variables.

```supercollider
s; // server

f = {
	arg c = s;  // this 's' refers to the 's' on the next line.
	var s = 1;
	c
}

f.() // returns 1, not the server
```

This is a feature that shouldn't have ever existed as it causes far more problems that it solves, if a user is using this as as feature, they have one more version to change it before it is made invalid behaviour.

---

This PR changes the name resolution so you cannot access variables and arguments before they are declared.

To mitigate this breaking change, in version 3.15 an error message is produced informing the user they need to update.

From version 3.16, this will discount the variable that hasn't been declared yet, and instead look it up in the parent scope.

Here is what this looks like.

---

### 1

```
{ arg a = b, b = 10; a }.()

ERROR: Accessing the named identifier 'b' before it was declared.
From the next version of SuperCollider, this will be an error and your code will fail to compile or behave differently.
Please fix the code before updating.
For more information on this change see scdoc://VersionUpdateGuide/ArgAndVarOrdering.html
  in interpreted text
  line 1 char 11:

  { arg a = b, b = 10; a }.() 
             
-> 10

```

### 2

Like 1, but uses expression not literal. All other examples behave the same as this if you replace the initialisation with an expression.

```
{ arg a = b, b = 2 + 2; a }.()

ERROR: Accessing the named identifier 'b' before it was declared.
From the next version of SuperCollider, this will be an error and your code will fail to compile or behave differently.
Please fix the code before updating.
For more information on this change see scdoc://VersionUpdateGuide/ArgAndVarOrdering.html
  in interpreted text
  line 1 char 11:

  { arg a = b, b = 2 + 2; a }.() 
             
-> nil
```
             
### 3

```
{ var x = y; var y; x }.()

ERROR: Accessing the named identifier 'y' before it was declared.
From the next version of SuperCollider, this will be an error and your code will fail to compile or behave differently.
Please fix the code before updating.
For more information on this change see scdoc://VersionUpdateGuide/ArgAndVarOrdering.html
  in interpreted text
  line 1 char 11:

  { var x = y; var y; x }.() 
             
-> nil
```

### 4

```
{ var x = x; x }.()

ERROR: Using the variable 'x' in its own initialiser.
From the next version of SuperCollider this will be an error and your code will fail to compile or behave differently.
Please fix the code before updating.
For more information on this change see scdoc://VersionUpdateGuide/ArgAndVarOrdering.html
  in interpreted text
  line 1 char 11:

  { var x = x; x }.() 
             
-> nil
```

### 5

This one produces no error.
This is to allow recursive functions.
```
{ var f = { f }; f }.()
```

### 6

```
{ var x = y; var y = 10; x }.()

ERROR: Accessing the named identifier 'y' before it was declared.
From the next version of SuperCollider this will be an error and your code will fail to compile or behave differently.
Please fix the code before updating.
For more information on this change see scdoc://VersionUpdateGuide/ArgAndVarOrdering.html
  in interpreted text
  line 1 char 11:

  { var x = y; var y = 10; x }.() 
             
-> 10

```

## Behavior in 3.16

```
{ arg a = b, b = 10; a }.()
-> nil
```

Why?? `b` is an interpreter variable!
```
b = \meow;
{ arg a = b, b = 10; a }.()
-> \meow
```

All of the possible errors produce the same error message because they are all trying to lookup the name strictly **before** the current initialization.

```
{ arg a1 = b1, b1 = 10; a1 }.()

ERROR: Variable 'b1' not previously declared.
  in interpreted text
  line 1 char 13:

  { arg a1 = b1, b1 = 10; a1 }.() 
               
-> nil
```

### DIFF
--- a/SCClassLibrary/JITLib/GUI/EnvirGui.sc
+++ b/SCClassLibrary/JITLib/GUI/EnvirGui.sc
@@ -20,7 +20,9 @@ EnvirGui : JITGui {
 
 	removeReplaceKey { |replacer|
 		var pv = this.viewForParam(replacer);
-		var replaced = replaceKeys.removeAt(replaced);
+		// TODO: this looks wrong, what was the intention here?	
+		// Note that 'replaced' is not used!
+		var replaced = replaceKeys.removeAt(nil);
 		if (pv.notNil) { pv.label_(replacer) };
 	}
 
--- a/lang/LangSource/PyrParseNode.cpp
+++ b/lang/LangSource/PyrParseNode.cpp
@@ -37,11 +37,14 @@
 #include <stdlib.h>
 #include <stdarg.h>
 #include <cctype>
+#include <unordered_map>
 #include "PredefinedSymbols.h"
+#include "SC_Version.hpp"
 #include "SimpleStack.h"
 #include "SC_LanguageConfig.hpp"
 #include "SC_Codecvt.hpp"
 #include "SpecialSelectorsOperatorsAndClasses.h"
+#include "SC_VersionSwitches.hpp"
 
 namespace fs = std::filesystem;
 
@@ -69,6 +72,9 @@ PyrClass* gCompilingClass = nullptr;
 PyrMethod* gCompilingMethod = nullptr;
 PyrBlock* gCompilingBlock = nullptr;
 PyrBlock* gPartiallyAppliedFunction = nullptr;
+
+enum struct InitStates { NotStarted, Started, Finished };
+std::unordered_map<PyrBlock*, std::vector<InitStates>> gNamedIdentifierInitStates {};
 
 bool gIsTailCodeBranch = false;
 bool gTailIsMethodReturn = false;
@@ -132,8 +138,8 @@ void compileQMsg(PyrParseNode* arg1, PyrParseNode* arg2);
 void compileQQMsg(PyrParseNode* arg1, PyrParseNode* arg2);
 void compileXQMsg(PyrParseNode* arg1, PyrParseNode* arg2);
 void compileSwitchMsg(PyrCallNode* node);
-void compileAssignVar(PyrParseNode* node, PyrSymbol* varName, bool drop);
-void compilePushVar(PyrParseNode* node, PyrSymbol* varName);
+void compileAssignVar(PyrParseNode* node, PyrSymbol* varName, bool drop, bool ignoreInitErrors);
+void compilePushVar(PyrParseNode* node, PyrSymbol* varName, bool ignoreInitErrors = false);
 bool isAnInlineableBlock(PyrParseNode* node);
 bool isAnInlineableAtomicLiteralBlock(PyrParseNode* node);
 bool isAtomicLiteral(PyrParseNode* node);
@@ -155,8 +161,8 @@ void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* super
 bool isThisObjNode(PyrParseNode* node);
 int conjureSelectorIndex(PyrParseNode* node, PyrBlock* func, bool isSuper, PyrSymbol* selector, int* selType);
 Byte conjureLiteralSlotIndex(PyrParseNode* node, PyrBlock* func, PyrSlot* slot);
-bool findVarName(PyrBlock* func, PyrClass** classobj, PyrSymbol* name, int* varType, int* level, int* index,
-                 PyrBlock** tempfunc);
+bool findVarName(PyrParseNode* node, PyrBlock* func, PyrClass** classobj, PyrSymbol* name, int* varType, int* level,
+                 int* index, PyrBlock** tempfunc, bool ignoreInitErrors = false);
 void countClassVarDefs(PyrClassNode* node, int* numClassMethods, int* numInstMethods);
 void dumpNodeList(PyrParseNode* node);
 int compareCallArgs(PyrMethodNode* node, PyrCallNode* cnode, int* varIndex, PyrClass* specialClass);
@@ -253,18 +259,19 @@ struct FindVarNameResult {
     PyrClass* classobj;
 };
 
-std::optional<FindVarNameResult> findVarName(PyrBlock* func, PyrClass* classobjC, PyrSymbol* varName) {
+std::optional<FindVarNameResult> findVarName(PyrParseNode* node, PyrBlock* func, PyrClass* classobjC,
+                                             PyrSymbol* varName, bool ignoreInitErrors = false) {
     int level, index, varType;
     PyrClass* classobj = classobjC;
     PyrBlock* tempfunc;
 
-    if (findVarName(func, &classobj, varName, &varType, &level, &index, &tempfunc))
+    if (findVarName(node, func, &classobj, varName, &varType, &level, &index, &tempfunc, ignoreInitErrors))
         return FindVarNameResult { level, index, varType, tempfunc, classobj };
     else
         return std::nullopt;
 }
 
-void compilePushVar(PyrParseNode* node, PyrSymbol* varName) {
+void compilePushVar(PyrParseNode* node, PyrSymbol* varName, bool ignoreInitErrors) {
     if (std::isupper(varName->name[0])) {
         if (compilingCmdLine && varName->u.classobj == nullptr) {
             error("Class not defined.\n");
@@ -288,7 +295,7 @@ void compilePushVar(PyrParseNode* node, PyrSymbol* varName) {
         PushSpecialValue.emit(OpSpecialValue::False);
     } else if (varName == s_nil) {
         PushSpecialValue.emit(OpSpecialValue::Nil_);
-    } else if (const auto result = findVarName(gCompilingBlock, gCompilingClass, varName)) {
+    } else if (const auto result = findVarName(node, gCompilingBlock, gCompilingClass, varName, ignoreInitErrors)) {
         const FindVarNameResult findResult = *result;
         switch (findResult.varType) {
         case varInst:
@@ -1155,7 +1162,7 @@ int compareCallArgs(PyrMethodNode* node, PyrCallNode* cnode, int* varIndex, PyrC
         PyrClass* classobj;
 
         classobj = gCompilingClass;
-        varFound = findVarName(gCompilingBlock, &classobj, slotRawSymbol(&nameNode->mSlot), &varType, &varLevel,
+        varFound = findVarName(node, gCompilingBlock, &classobj, slotRawSymbol(&nameNode->mSlot), &varType, &varLevel,
                                varIndex, nullptr);
         if (!varFound)
             return methNormal;
@@ -1327,6 +1334,8 @@ void PyrMethodNode::compile(PyrSlot* result) {
     methraw->popSize = numSlots - 1;
     firstKeyIndex = numArgs + numVariableArgs + numKwArgs;
 
+    gNamedIdentifierInitStates[method] = std::vector<InitStates>(numSlots, InitStates::NotStarted);
+
     numArgNames = methraw->posargs;
 
     if (numSlots == 1) {
@@ -1487,6 +1496,15 @@ void PyrMethodNode::compile(PyrSlot* result) {
         }
     }
 
+    // If there are no initialiser expression, which would involve using the name resolution, then set all the
+    // gNamedIdentifiers to finished. This is done here because some optimised types of methods never actually call
+    // compile on their body. If there are initialiser expression, then the methType is about to be set to methNormal
+    // and it will be compiled as expected.
+    if (!hasVarExprs) {
+        for (auto& v : gNamedIdentifierInitStates[method])
+            v = InitStates::Finished;
+    }
+
     methType = methNormal;
     if (hasVarExprs) {
         methType = methNormal;
@@ -1634,16 +1652,26 @@ void PyrMethodNode::compile(PyrSlot* result) {
         compile_body:
             SetTailIsMethodReturn mr(false);
             PyrSlot dummy;
+            const auto numTotalArguments = methraw->posargs;
             if (mArglist) {
                 vardef = mArglist->mVarDefs;
                 for (i = 1; i < numArgs; ++i, vardef = (PyrVarDefNode*)vardef->mNext) {
+                    gNamedIdentifierInitStates[method][i] = InitStates::Started;
                     vardef->compileArg(&dummy);
+                    gNamedIdentifierInitStates[method][i] = InitStates::Finished;
                 }
+                // No need to compile, you can't set them.
+                if (mArglist->mRest)
+                    gNamedIdentifierInitStates[method][numArgs] = InitStates::Finished;
+                if (mArglist->mKeywordArgs)
+                    gNamedIdentifierInitStates[method][numArgs + 1] = InitStates::Finished;
             }
             if (mVarlist) {
                 vardef = mVarlist->mVarDefs;
                 for (i = 0; i < numVars; ++i, vardef = (PyrVarDefNode*)vardef->mNext) {
+                    gNamedIdentifierInitStates[method][numTotalArguments + i] = InitStates::Started;
                     vardef->compile(&dummy);
+                    gNamedIdentifierInitStates[method][numTotalArguments + i] = InitStates::Finished;
                 }
             }
             COMPILENODE(mBody, &dummy, true);
@@ -1651,11 +1679,13 @@ void PyrMethodNode::compile(PyrSlot* result) {
         installByteCodes((PyrBlock*)method);
     }
 
+
     if (!oldmethod) {
         addMethod(gCompilingClass, method);
     }
 
     gCompilingMethod = nullptr;
+    gNamedIdentifierInitStates.erase(method);
     gCompilingBlock = nullptr;
     gPartiallyAppliedFunction = nullptr;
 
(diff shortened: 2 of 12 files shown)


prko:
Will this affect the following implementation in https://github.com/supercollider/supercollider/issues/7192#issuecomment-3431066871?
```
String {
	...
	interpret { |... args| ^thisProcess.interpreter.interpret(this, *args); }
	interpretPrint { |args| ^thisProcess.interpreter.interpretPrint(this, *args); }
	...
}
```

lnihlen:
If you're going to use my work I feel it's important to set the context of this example from my talk. This was intended as an example of the kind of change to the SuperCollider language that I didn't feel comfortable making *without extensive community discussion,* because of the incompatibility concerns and "unexpected consequences" when interacting with legacy code.

prko:
From my viewpoint,
```
(
arg a = 1, b = 2, c = a + b, d = x;
var x = a + b;
d;
)
```
It should return a do-not-understand error for the following reasons:

1.  At the moment of defining `d = x`, the value of `x` is nil.
2. `nil` cannot be added to an integer.

This breaking change could introduce many unexpected behaviours in code that previously worked differently.

JordanHendersonMusic:
> This was intended as an example of the kind of change to the SuperCollider language that I didn't feel comfortable making without extensive community discussion, because of the incompatibility concerns and "unexpected consequences" when interacting with legacy code.

I definitely agree there should be a conversation, this was the only way I could think to start that! And thanks for the talk, I really enjoyed it!

However, I think this particular issue is very minor, I've never seen code using this, and because it reorders it differently depending on whether it is a literal or expression, it is unstable to the point of being not useful. It is also an easy thing to fix and standing in the way of one of the most requested features, inline variables. 


On the general point about supercollider being an archive of author's works, I think it is too much of an ask from a user to expect a tiny group of non-paid contributors to take on the task of an archivist as well as trying to improve and maintain sc. Obviously we shouldn't break code without reason, and should provide simple steps to allow authors to update their work over a long period of time. But the write once run forever with no upkeep mentality seems unreasonable. 

This is why I'm suggesting a warning, give the user base a year or so to adapt their code, then move ahead with the breaking change.


@prko 

> Will this affect the following implementation in

That doesn't seem related to me, there is no code in the class library that triggers this warning.

> It should return a do-not-understand error for the following reasons:

This code uses x before it is defined, SC always evaluates code top to bottom left to right... You are suggesting something similar to time travel which would be very confusing. It would also be annoying to implement and introduce another issue where there are cycles of dependencies between variables.

JordanHendersonMusic:
@prko I've changed the examples (and added one) to make it clean these aren't interpreter variables which might make it more obvious how the compiler sometimes decided to time travel.

lnihlen:
> > This was intended as an example of the kind of change to the SuperCollider language that I didn't feel comfortable making without extensive community discussion, because of the incompatibility concerns and "unexpected consequences" when interacting with legacy code.
> 
> I definitely agree there should be a conversation, this was the only way I could think to start that! And thanks for the talk, I really enjoyed it!
> 
> However, I think this particular issue is very minor, I've never seen code using this, and because it reorders it differently depending on whether it is a literal or expression, it is unstable to the point of being not useful. It is also an easy thing to fix and standing in the way of one of the most requested features, inline variables.
> 
> On the general point about supercollider being an archive of author's works, I think it is too much of an ask from a user to expect a tiny group of non-paid contributors to take on the task of an archivist as well as trying to improve and maintain sc. Obviously we shouldn't break code without reason, and should provide simple steps to allow authors to update their work over a long period of time. But the write once run forever with no upkeep mentality seems unreasonable.
> 
> This is why I'm suggesting a warning, give the user base a year or so to adapt their code, then move ahead with the breaking change.

You're quite welcome for the talk, I'm glad you enjoyed it. You make good points here. I was saying in the talk that SC needed a governance model, and I think I had some grand vision of a very community-driven process like had been proposed with the RFCs. But "governance by maintainers" is pragmatic and it allows the language to move forward. To be fair, it's also more democratic than the BDFL model I have in Hadron currently, if only for lack of contributors.

capital-G:
IMO this is more a resolution of undefined behavior than a breaking change (still ofc a breaking change).

One motivation of establishing the council structure is that big decisions like these can be managed since there is a council, in this case of the [sclang council](https://github.com/supercollider/supercollider/wiki/Council-structure#sclang), should manage a discussion and is in the end "allowed" to pull the trigger on a decision based on the discussion.

Maybe we should discuss this in the dev meeting how such a discussion should look like? Maybe we can take the "best bits" of the RFC procedure and try out how it works? Maybe this particular issue is a good test balloon since it is a bigger change but probably w/o really big consequences.

muellmusik:
FWIW, I have never seen code like this, would not think the result predictable, and would expect it to probably be undefined behaviour.

> I've changed the examples (and added one) to make it clean these aren't interpreter variables which might make it more obvious how the compiler sometimes decided to time travel.

I was going to suggest that as well. This works as expected and makes some sense to me:

```
x = 2;

{
	arg d1 = x;
	d1;
}.value // returns 2
```
I'm prepared to be convinced otherwise if there's something I'm unaware of, and of course keep discussing, but for the moment I don't see any reason not to fix this.

lnihlen:
Well, lexical scoping certainly adds some additional wrinkles to inline declaration, for example this is valid sclang code:

```supercollider
method {
  var f = { var d1 = x; d1 };

  // .... other code

  var x = 2;
}
```

If we allow inline declaration, this could arguably be a compiler error. `x` is being referenced in `f` before it is declared. Not before it's being used, however. But in other languages supporting lexical scoping, this would be flagged as bad code. What happens if `x` or `f` are class or instance variables?

I was trying to make a couple of points in the talk, perhaps not very well. The first is that even something appearing as "simple" and uncontroversial like inline variables can actually have a really profound impact on your user base. And it's not just obviously wrong code constructed for rhetorical purposes, this will have impact on some relatively mundane code as shared above.

The second point was that the situation is actually worse than that. Right now, sclang initializes arguments and local variables in two phases. The slot literals get copied into a "prototype" array during compilation (strings and arrays, being objects, are marked as "read only") and are then copied again during stack frame initialization. Any non-trivial expression initializers are compiled and injected at the start of the method.

It means that I can create really silly examples that look like use-before-declaration but are actually valid, typically by using local variables to initialize arguments. But it *also* means that *interpreter implementation details have leaked into the language abstraction.* And that's the real problem, because now we can't change the implementation without the likelihood of breaking users.

But the last time I tried to commit code to sclang I got pushback and my PR was ultimately rejected because I was trying to fix a broken integer parser (IIRC) and there was concern that people were relying on the broken behaviour. It seems the dev culture has changed significantly and has much higher risk tolerance, so this will be interesting to see how it unfolds.

JordanHendersonMusic:
>  If we allow inline declaration, this could arguably be a compiler error. x is being referenced in f before it is declared. Not before it's being used, however. But in other languages supporting lexical scoping, this would be flagged as bad code. What happens if x or f are class or instance variables?

My intention is to ensure SC strictly follows top to bottom, left to right. This is step one.

> ... can actually have a really profound impact on your user base. And it's not just obviously wrong code constructed for rhetorical purposes, this will have impact on some relatively mundane code as shared above.

I thought this was expressed very clearly in your talk, but the scale of the impact seems like it is going to be small, unpredictable, but still small. No one is going to rely on this regularly, because the behaviour changes whether it is a literal or expression (unstable), so it should only occur infrequently, meaning it will be easy to fix. 

I'd also like to point out we don't have any direct evidence someone has relied on this beyond these nonsense examples. 

Recently we made a similar breaking change in keyword arguments and being able to set the `...args` variable directly, which you can no longer do. @jamshark70 I believe this broke some of your code, do you have any thoughts on this, or perhaps on how such breaking changes might be communicated better?

lnihlen:
My only complaint about kwargs as an interpreter implementer I haven't been able to find detailed documentation about the specific change to the language that was made. Do you have that handy?

muellmusik:
> But the last time I tried to commit code to sclang I got pushback and my PR was ultimately rejected because I was trying to fix a broken integer parser (IIRC) and there was concern that people were relying on the broken behaviour. It seems the dev culture has changed significantly and has much higher risk tolerance, so this will be interesting to see how it unfolds.

I guess one has to take these things case by case (I don’t recall the one you mention), and of course one can reasonably make an argument that breaking code should be done only rarely. For myself I sometimes think we worry about it a little too much.

I thought your talk was really great btw @lnihlen! If I’d been in the room I would have suggested that this issue is maybe something that you should just not worry about and fix. The potential downside seems very small. There’s also the tradition in SC of major versions being big breaking changes / rethinks, and I think a brand new interpreter implementation would be an appropriate occasion for that! (Sorry drifting off topic but thought this was worth saying.)

JordanHendersonMusic:
@lnihlen 

> My only complaint about kwargs as an interpreter implementer I haven't been able to find detailed documentation about the specific change to the language that was made. Do you have that handy?

Everything I did was in the PR, that was my first big PR ever (come from a music background) so there is a strong chance I should have done some more documenting! I'm quite happy to talk about it though and maybe create some documentation if you think it necessary, but let's carry on that conversation somewhere else, issue, pr, or forum, I don't mind.

> ...this issue is maybe something that you should just not worry about and fix. 

re-'fix'
 
@muellmusik 
So I'd much rather jump straight to inline variables and give hard compiler errors when the user tries to do this. However, there is this case ...
```supercollider
{
   arg a1 = s;
   var s = 1;
   a1
}
```
...which currently returns 1, but if I just went ahead and made this behaviour in the way we'd all expect (no time travel), it should return the server. My concern isn't outright breaking people's code (compiler warnings are easy to track down and fix), it is subtly changing the code from one functioning form into another — this is very hard to debug and will annoy people far more than an error/warning.

By posting the warning for a single release, there is a transition period before we change the behaviour of the code, giving people time to adapt. Perhaps, it should be an error instead?

muellmusik:
@JordanHendersonMusic Sorry when I said just fix it, that was a response to @lnihlen’s concern about compatibility. I think a warning is for sure the right thing to do in this case. I’m just aware that I haven’t seen an example of this that isn’t ‘silly’, or at least probably bad design, like both using interpreter bars and locally declaring vars with the same name in the same function. But there _could_ be something so a warning seems good.

(If there was a major version I think it would be okay just to break stuff. People would move projects to the new version on the understanding that they should be checked and SC3 could still be maintained if needed.)

Thanks for taking this on!

jamshark70:
> Recently we made a similar breaking change in keyword arguments and being able to set the ...args variable directly, which you can no longer do. @jamshark70 I believe this broke some of your code, do you have any thoughts on this, or perhaps on how such breaking changes might be communicated better?

Oh right, that case was like this -- an attempt to use keyword addressing on a `... x` arbitrary-length argument:

```
f = { |a ... b| [a, b].postln };

f.(b: 3);
WARNING: keyword arg 'b' not found in call to function.
[nil, []]
-> [nil, []]
```

... where previous versions would have printed `[nil, [3]]`.

My case happened to be doing non-critical area, so I'm not going to worry about it (and it's not directly related to this issue, so, here isn't the right place to go into it).

~~~

For the current issue, I broadly agree with the rule "top to bottom, left to right," and I also agree that breaking that rule isn't really a sensible usage to preserve.

capital-G:
> (If there was a major version I think it would be okay just to break stuff. People would move projects to the new version on the understanding that they should be checked and SC3 could still be maintained if needed.)

Thats interesting b/c I always understood that one goal of SC is to support old code b/c there are cultural artifacts building upon this which probably should be preserved.

Additionally, I am not really sure if we even require a major version bump (see python2->python3 fiasco). Having the sclang FFI (#7189 ) and hopefully soon a proper package manager would allow us to reduce ties of maintaining while still keeping compatibility and allow us to develop new features or change interfaces within the class library. E.g. we also changed the init behavior of some UGens which changed/corrected their outcome - w/ a package manager we could still keep those "broken" versions available upon request.
If we want even more resilience we would need to introduce namespaces such that two versions of the same dependency could co-exist w/ each other.

Instead of a major bump for sclang we probably could use of something like rust editions, see https://doc.rust-lang.org/edition-guide/editions/ - this would require maintenance effort on sclang side but I think it would be worth it b/c it would allow us to introduce new and breaking features w/o breaking old code.

Maybe we can discuss/coordinate this further in the [next dev meeting](https://github.com/supercollider/supercollider/issues/7164#issuecomment-3453872325)?

telephon:
As far as I can tell, we are perhaps less risk-averse, but still very careful. It is important to stay backwards compatible as much as possible. As far as I can tell, it would have to be a major feature that would justify a breaking change. But there are boundary cases, where hidden "features" can be exploited. I think that ironing out such wrinkels doesn't come with such a high risk, especially if the changes necessary to fix broken code are small and the error clear.

The thing with the number fix was (too) subtle, if I remember correctly ... (?)

lang/LangSource/PyrParseNode.cpp
```diff
@@ -48,6 +48,8 @@ namespace fs = std::filesystem;
 
 using namespace Opcode;
 
+bool allowVariableLoopup = true;
```
dyfer: "Loopup"?

lang/LangSource/PyrParseNode.cpp
```diff
@@ -4173,6 +4179,7 @@ int conjureConstantIndex(PyrParseNode* node, PyrBlock* func, PyrSlot* slot) {
     return constants->size - 1;
 }
 
+
```
dyfer, line 4447: unnecessary whitespace?

dyfer (commented):
Just double checking spelling

JordanHendersonMusic:
Sorry about the delay!

Anyone have any objections? There has been a lot of conversation here, but no one has actually objected.

telephon (dismissed):

JordanHendersonMusic:
I'm going to change how this is implemented so that we automatically get the error in 3.16 . But we agreed to merge.

dyfer (dismissed):
just to avoid accidental merge - waiting to change the implementation of the warning by @JordanHendersonMusic

JordanHendersonMusic:
Okay so I've changed the behaviour here.

Given this...
```
(
var x1 = 10;
var f = {
	arg a1 = x1;
	var x1 = 2 + 2;
	a1
};
f.()
);
```

In version 3.15, this will return ...

```
ERROR: Accessing a variable from the current function inside an argument initaliser is not allowed.
In future supercollider versions the variable will be looked up in the parent scope.
Please fix your code before updating.
-> nil
```

Which is the old behaviour.

However, from version 3.16, this will return...

```
10
```

Edit: just realised I need to check `var x = x;` I think right now this is nil, but it should also look up `x` in the parent scope.


---

This is a change from what we have discussed, but I think it makes the most sense. Now, the compiler just assumes you can never access the variables in the current function while initialising arguments and skips them.

This means that without the `var x1 = 10;` line, you will get a variable not found error (from 3.16).

prko:
Thank you!

Not all users run the most recent release of SuperCollider; in fact, we still see forum posts from people using considerably older versions. For this reason, it might be preferable for 3.15 to produce the same result as 3.16, while displaying a clear message before the result, or summarising the change in the NEWS file as usual. This would avoid different behaviour across versions and help users on older releases understand the transition without unexpected discrepancies.

SCClassLibrary/JITLib/GUI/EnvirGui.sc
```diff
@@ -20,7 +20,8 @@ EnvirGui : JITGui {
 
 	removeReplaceKey { |replacer|
 		var pv = this.viewForParam(replacer);
-		var replaced = replaceKeys.removeAt(replaced);
+		// TODO: this looks wrong, what was the intention here?	
+		var replaced = replaceKeys.removeAt(nil);
```
JordanHendersonMusic, line 24: Does anyone know what this was supposed to do? I think this change has already found one bug!

dyfer: I don't know, but maybe @telephon has some idea?

JordanHendersonMusic:
This PR is a big change, could break peoples code if they have relied on a variable before its definition, but I think worth it. Without it, having `var`s in the middle of functions #7381 would cause all sorts of chaos! 

The approach here is to generate a warning in the next sc version, before moving to the new behaviour in the following. By that time, we can also enable the inline vars.

I think this PR is good for review.

SCClassLibrary/JITLib/GUI/EnvirGui.sc
telephon: It looks like a typo, and probably should be
```
var replaced = replaceKeys.removeAt(replacer);
```

@adcxyz – could you please check? Thanks!!

dyfer:
TODO before merging:
- review [NameResolution.schelp](https://github.com/supercollider/supercollider/pull/7203/changes#diff-1be90513bf20869c8f31bdad9adc990ba22b9ddb87e39ec521987c49709c4c2c)
- compile all the quarks to see if required changes are trivial or not

SCClassLibrary/JITLib/GUI/EnvirGui.sc
JordanHendersonMusic: This is also a bit confusing because the variable **replaced** is never used.

PR #7205, opened 2025-10-25
sclang: introduce AbstractObject

JordanHendersonMusic:
Introduce abstract object changing the class hierarchy.

Note, AbstractObject is by definition an **incomplete** object, using it directly will break the interpreter, that is the whole point, it is the user's responsibility to implement all these features. 

This was @telephon's idea from an RFC from way back, and I only just got around to implementing it.

This PR depends on #7206.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
+++ b/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
@@ -1,18 +1,18 @@
-// ExperimentalAbstractObject represent the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
+// AbstractObjectExperimental represent the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
 // This make it very useful as a wrapper class and to implement some specific features.
 
 // !!! --- PLEASE NEVER ADD A NORMAL PASCAL CASE METHOD TO ABSTRACT OBJECT --- !!!
-// The whole point of this class is to be minimal, ExperimentalAbstractObjects are not Objects, allowing those who subclass it to create their own definition of what an object is and isn't.
+// The whole point of this class is to be minimal, AbstractObjectExperimentals are not Objects, allowing those who subclass it to create their own definition of what an object is and isn't.
 
-ExperimentalAbstractObject {
+AbstractObjectExperimental {
 
 	/////// 1. Core and Interpreter functionality, do not override.
 
 	*new { |maxSize = 0| _BasicNew; ^this.sc_abstract_object_primitive_failed }
 	*newCopyArgs { | ... args, kwargs | _BasicNewCopyArgsToInstVars; ^this.sc_abstract_object_primitive_failed }
 
-	// ExperimentalAbstractObject support identity, this is required by the interpreter and cannot actually be overridden.
-	// When you put ExperimentalAbstractObjects into an identity dictionary the interpreter uses its own implementation, this should match the sclang implementation or things will get confusing.
+	// AbstractObjectExperimental support identity, this is required by the interpreter and cannot actually be overridden.
+	// When you put AbstractObjectExperimentals into an identity dictionary the interpreter uses its own implementation, this should match the sclang implementation or things will get confusing.
 	// Identity operators are also provided, this isn't strictly necessary, but we usually think of identity and these operators as a set.
 	identityHash { _ObjectHash; ^this.sc_abstract_object_primitive_failed }
 	=== { arg obj; _Identical; ^this.sc_abstract_object_primitive_failed }
@@ -20,7 +20,7 @@ ExperimentalAbstractObject {
 
 	// Interpreter will engage in undefined behavior if this method does not throw.
 	// It is not possible to implement a boolean abstract object as the compiler assumes its internal slot layout.
-	mustBeBoolean {  MustBeBooleanError(nil, "ExperimentalAbstractObject").throw }
+	mustBeBoolean {  MustBeBooleanError(nil, "AbstractObjectExperimental").throw }
 
 	// It is not possible to implement a 'nil' abstract object due to the compiler inlining the definition of these when used inside 'if' control flow.
 	// You cannot override these methods.
@@ -37,8 +37,8 @@ ExperimentalAbstractObject {
 	// Note: Interpreter will segfault if this isn't provided so a default is given.
 	// You will need to override doesNotUnderstand to anything useful will abstract objects,
 	//   hence why it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
-	// "ExperimentalAbstractObject" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
-	doesNotUnderstand { |selector ...args, kwargs| SubclassResponsibilityError("ExperimentalAbstractObject", thisMethod, ExperimentalAbstractObject).throw }
+	// "AbstractObjectExperimental" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
+	doesNotUnderstand { |selector ...args, kwargs| SubclassResponsibilityError("AbstractObjectExperimental", thisMethod, AbstractObjectExperimental).throw }
 
 
 
@@ -47,7 +47,7 @@ ExperimentalAbstractObject {
 	// These methods may be used *only* to implement other abstract objects, do not call them from outside the definition of an abstract object.
 	// They are here so that other's may use primitive implementations, without making the primitives themselves publicly available.
 
-	sc_abstract_object_primitive_failed { PrimitiveFailedError("ExperimentalAbstractObject").throw }
+	sc_abstract_object_primitive_failed { PrimitiveFailedError("AbstractObjectExperimental").throw }
 
 	sc_abstract_object_hash { _ObjectHash; ^this.sc_abstract_object_primitive_failed }
 	sc_abstract_object_copy_shallow { _ObjectShallowCopy; ^this.sc_abstract_object_primitive_failed }
@@ -55,7 +55,7 @@ ExperimentalAbstractObject {
 	sc_abstract_object_copy_deep { _ObjectDeepCopy; ^this.sc_abstract_object_primitive_failed }
 
 	// Yielding — must be implemented as primitives as special methods are provided.
-	// Note that idle is not provided because it requires calling 'value' on the ExperimentalAbstractObject'.
+	// Note that idle is not provided because it requires calling 'value' on the AbstractObjectExperimental'.
 	sc_abstract_object_yield {	_RoutineYield; ^this.sc_abstract_object_primitive_failed }
 	sc_abstract_object_always_yield { _RoutineAlwaysYield; ^this.sc_abstract_object_primitive_failed }
 	sc_abstract_object_yield_and_reset { |reset = true| _RoutineYieldAndReset; ^this.sc_abstract_object_primitive_failed }
@@ -63,7 +63,7 @@ ExperimentalAbstractObject {
 	// This is often necessary in abstract objects to determine if we have an abstract object, use isKindOf in normal use.
 	sc_abstract_object_is_kind_of { |aClass| _ObjectIsKindOf; ^this.sc_abstract_object_primitive_failed }
 
-	// Needed when we want to do the normal method calling in an ExperimentalAbstractObject.
+	// Needed when we want to do the normal method calling in an AbstractObjectExperimental.
 	// In many cases, the 'perform*' messages should defer to these as most of the time, you actually want to implement doesNotUnderstand.
 	sc_abstract_object_perform_args { |selector, args, kwargs|	_ObjectPerformArgs;	^this.sc_abstract_object_primitive_failed }
 	sc_abstract_object_super_perform_args { |selector, args, kwargs| _ObjectSuperPerformArgs; ^this.sc_abstract_object_primitive_failed }
--- a/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
+++ b/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
@@ -1,5 +1,5 @@
 // AbstractObjectExperimental represent the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
-// This make it very useful as a wrapper class and to implement some specific features.
+// This make it useful as a wrapper class and to implement some specific features.
 
 // !!! --- PLEASE NEVER ADD A NORMAL PASCAL CASE METHOD TO ABSTRACT OBJECT --- !!!
 // The whole point of this class is to be minimal, AbstractObjectExperimentals are not Objects, allowing those who subclass it to create their own definition of what an object is and isn't.
@@ -32,11 +32,11 @@ AbstractObjectExperimental {
 
 
 
-	/////// 2. Main interface, override this method.
+	/////// 2. Main interface, override these method.
 
 	// Note: Interpreter will segfault if this isn't provided so a default is given.
 	// You will need to override doesNotUnderstand to anything useful will abstract objects,
-	//   hence why it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
+	//   hence it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
 	// "AbstractObjectExperimental" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
 	doesNotUnderstand { |selector ...args, kwargs| SubclassResponsibilityError("AbstractObjectExperimental", thisMethod, AbstractObjectExperimental).throw }
 
@@ -45,7 +45,7 @@ AbstractObjectExperimental {
 	/////// 3. Helper methods and primitive implementations
 
 	// These methods may be used *only* to implement other abstract objects, do not call them from outside the definition of an abstract object.
-	// They are here so that other's may use primitive implementations, without making the primitives themselves publicly available.
+	// They are here so that other objects may use primitive implementations, without making the primitives themselves publicly available.
 
 	sc_abstract_object_primitive_failed { PrimitiveFailedError("AbstractObjectExperimental").throw }
 
--- a/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
+++ b/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
@@ -1,5 +1,5 @@
 // AbstractObjectExperimental represents the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
-// This make it useful as a wrapper class and to implement some specific features.
+// This makes it useful as a wrapper class and to implement some specific features.
 
 // !!! --- PLEASE NEVER ADD A NORMAL PASCAL CASE METHOD TO ABSTRACT OBJECT --- !!!
 // The whole point of this class is to be minimal.
--- a/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
+++ b/SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
@@ -1,0 +1,74 @@
+// ExperimentalAbstractObject represent the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
+// This make it very useful as a wrapper class and to implement some specific features.
+
+// !!! --- PLEASE NEVER ADD A NORMAL PASCAL CASE METHOD TO ABSTRACT OBJECT --- !!!
+// The whole point of this class is to be minimal, ExperimentalAbstractObjects are not Objects, allowing those who subclass it to create their own definition of what an object is and isn't.
+
+ExperimentalAbstractObject {
+
+	/////// 1. Core and Interpreter functionality, do not override.
+
+	*new { |maxSize = 0| _BasicNew; ^this.sc_abstract_object_primitive_failed }
+	*newCopyArgs { | ... args, kwargs | _BasicNewCopyArgsToInstVars; ^this.sc_abstract_object_primitive_failed }
+
+	// ExperimentalAbstractObject support identity, this is required by the interpreter and cannot actually be overridden.
+	// When you put ExperimentalAbstractObjects into an identity dictionary the interpreter uses its own implementation, this should match the sclang implementation or things will get confusing.
+	// Identity operators are also provided, this isn't strictly necessary, but we usually think of identity and these operators as a set.
+	identityHash { _ObjectHash; ^this.sc_abstract_object_primitive_failed }
+	=== { arg obj; _Identical; ^this.sc_abstract_object_primitive_failed }
+	!== { arg obj;_NotIdentical; ^this.sc_abstract_object_primitive_failed }
+
+	// Interpreter will engage in undefined behavior if this method does not throw.
+	// It is not possible to implement a boolean abstract object as the compiler assumes its internal slot layout.
+	mustBeBoolean {  MustBeBooleanError(nil, "ExperimentalAbstractObject").throw }
+
+	// It is not possible to implement a 'nil' abstract object due to the compiler inlining the definition of these when used inside 'if' control flow.
+	// You cannot override these methods.
+	isNil { ^false }
+	notNil { ^true }
+	? { |obj| ^this }
+	?? { |obj| ^this }
+	!? { |obj| ^obj.value(this) }
+
+
+
+	/////// 2. Main interface, override this method.
+
+	// Note: Interpreter will segfault if this isn't provided so a default is given.
+	// You will need to override doesNotUnderstand to anything useful will abstract objects,
+	//   hence why it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
+	// "ExperimentalAbstractObject" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
+	doesNotUnderstand { |selector ...args, kwargs| SubclassResponsibilityError("ExperimentalAbstractObject", thisMethod, ExperimentalAbstractObject).throw }
+
+
+
+	/////// 3. Helper methods and primitive implementations
+
+	// These methods may be used *only* to implement other abstract objects, do not call them from outside the definition of an abstract object.
+	// They are here so that other's may use primitive implementations, without making the primitives themselves publicly available.
+
+	sc_abstract_object_primitive_failed { PrimitiveFailedError("ExperimentalAbstractObject").throw }
+
+	sc_abstract_object_hash { _ObjectHash; ^this.sc_abstract_object_primitive_failed }
+	sc_abstract_object_copy_shallow { _ObjectShallowCopy; ^this.sc_abstract_object_primitive_failed }
+	sc_abstract_object_copy_immutable { _ObjectCopyImmutable; ^this.sc_abstract_object_primitive_failed }
+	sc_abstract_object_copy_deep { _ObjectDeepCopy; ^this.sc_abstract_object_primitive_failed }
+
+	// Yielding — must be implemented as primitives as special methods are provided.
+	// Note that idle is not provided because it requires calling 'value' on the ExperimentalAbstractObject'.
+	sc_abstract_object_yield {	_RoutineYield; ^this.sc_abstract_object_primitive_failed }
+	sc_abstract_object_always_yield { _RoutineAlwaysYield; ^this.sc_abstract_object_primitive_failed }
+	sc_abstract_object_yield_and_reset { |reset = true| _RoutineYieldAndReset; ^this.sc_abstract_object_primitive_failed }
+
+	// This is often necessary in abstract objects to determine if we have an abstract object, use isKindOf in normal use.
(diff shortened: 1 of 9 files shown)


JordanHendersonMusic:
@telephon what do you think?

This differs from your proposal as Meta_Object's superclass is now Meta_AbstractObject. This is because class methods couldn't be looked up from AbstractObject in Object without this, and it seems to make the implementation easier.

I've also taken a hard stance on having as few methods in AbstractObject as possible, I know your rfc proposed a few, but I don't think they are necessary as you won't every use AbstractObject directly, instead you will subclass it, where you can decide if you opt into those methods, or defer with doesNotUnderstand.

telephon:
This is great! I like that you reduced the "stupidity" of the object and also included the class, that was my intention, too, I just had tried to make the [Abstract Object RFC](https://github.com/telephon/rfcs/blob/master/rfcs/0002-abstract-object.md) as "minimally invasive" and "maximally feasible" as possible. All the better that this is here now.

Maybe we should go through the https://github.com/telephon/Neutral quark to check what we want to have as part of the core subclasses?

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -56,6 +43,17 @@ Object {
 		^this.primitiveFailed
 	}
 
+
```
telephon: just semantically it may be clearer to perform these calls on "super"

SCClassLibrary/Common/Core/Object.sc
```diff
@@ -501,18 +471,9 @@ Object {
...
-	}
-	alwaysYield {
-		_RoutineAlwaysYield
-		^this.primitiveFailed
-	}
-	yieldAndReset { arg reset = true;
-		_RoutineYieldAndReset
-		^this.primitiveFailed
-	}
+	yield {	^this.sc_abstract_object_yield }
```
telephon: could you briefly explain why we can't or shouldn't write it here as defined in AbstractObject?
```
yield {	_RoutineYield; ^super.sc_abstract_object_primitive_failed }
```

telephon:
One (optional) feature that would make this even more useful would be to have a primitive that forwards a method call to another object one-to-one. Then layers of delegations cost very little, and one can freely wrap objects in delegators (and those in other ones), to add interfaces. I had noted that under "Method forwarding".

Do you think this would be hard to do? We already have optimised primitives for forwarding, don't we?

JordanHendersonMusic:
Thanks for taking a look! If there is time I'm going to talk about this at the next dev meeting as it is quite a large change. The proto object is where I have been spending most of my time. It seems to work very well, I even have a debug wrapper that lets you step through message calls to the held object.

Regarding exactly what should be in abstract object, my opinion is that it should contain nothing. Not even enough to print to the post window, because as the interpreter or kernel changes, these requirements will eventually change and we never want to be in a situation where we need an `EvenMoreAbstractObject`. This means it is always the user's job to turn an abstract object into a 'real' object.

Right now, all you need to do to create a wrapper is add a constructor, implement does not understand (and `mustBeBoolean`), and it works, except in control flow inlining, which I think is unfixable as I believe the jump byte codes are actually incompatible with the whole concept of non-local returns.

Regards faster delegation. I see what you mean, however, I think it will be functionally identical to performArgs, we still have to check for default arguments, too many or too few, keywords, variables args and keyword arguments. The fact is, supercollider's message calling convention is just complicated and slow. 

Now... I have some ideas to speed this up in certain cases. 

```
F { foo{ |...a, k| ^held.performArgs(\bar, a, k) } }
```
In this, highly specific case where we don't do anything to the arguments, it actually isn't necessary to do any argument management, just change the receiver.
```
F { foo{ |...a, k|
   doSomething();
   ^held.performArgs(\bar, a, k) 
} }
```
Likewise here, the arguments can just stay on the stack while `doSomething` is called (this could lead to more stack usage).

The issue is that the moment you introduce a named argument, we need to do the argument logic again.

telephon:
> Regarding exactly what should be in abstract object, my opinion is that it should contain nothing.

I absolutely agree. As I said, my care with regards to what needs to be kept was just to make it possible at all. Great that you found a way to reduce it to (almost) nothig.

> In this, highly specific case where we don't do anything to the arguments, it actually isn't necessary to do any argument management, just change the receiver.

My idea only concerned the case where you can forward the whole bunch of selector, args and kwargs to a different objects, without looking into it and of course without changing anything about them.

If I remember correctly, there are already some optimisations in place to forward a method call to the superclass. All there is to be done (if this is correct) would be to allow one more argument (the receiver) and then pass to that one instead of the superclass.

The receiver would then do the checking if the signature matches.

JordanHendersonMusic:
Yes I think that optimisation is a good idea and I'll have a look, should be quite easy! 

We might also consider a syntax to make this forwarding easier...

```
F {
   var h, g;
   foo, bar -> h;
   goo, gar -> g;
}
```
Or something? I think someone suggested something similar once on here, but I can't find it.

telephon:
Actually, I am not sure if we need extra syntax? It would be an internal method of delegator objects which you wouldn't have to touch.

I was wondering about the `#(...)` as well, if not this could simply be a method of `IdentityDictionary`, like `().delegate` or `().proto`. Simply because there may be many kinds of prototype objects, and we won't want a syntax for each. But perhaps it is justified, I also like it!

JordanHendersonMusic:
There are a two reasons why I like the extra delegate syntax. 

1. It makes it absolutely explicit what it does. 
2. This makes it very easy to implement in the compiler. Right now, the whole lexing, parsing and compiling is muddled and needed a major rewrite. I have started this, but it is a big old job!

But I agree the syntax is unnecessary.

We could also explore something like delegate to object when the message matches an interface.
```
F{ 
   var h;
   SomeInterfaceClass -> h;
}
```
Which automatically goes and generates the messages contained in `SomeInterfaceClass`. Might get a bit confusing with inheritance, but could be manageable!


--- 

I think `#(...)` for proto object is a good idea because most people try to use event `(...)` for object prototyping. This makes it easy to convert. I think this one is used enough to make it worth it.

I could rename proto object to something like `Structure`, or `Klass`? It is essentially an implementation of supercollider flavour Oop in supercollider.

I am particularly pleased about how inheritance of these protoobjevts turned out and that you can override doesnotundersnd 

```
p = #(doesNotUnderstand: \meow);
p.foo // \meow

base = #(speak: {|s| "I say" ++ s.speakImpl } );
child = #(super: base);
child.speakImpl = \meow;
child.speak // I say meow
```

telephon:
Re the extra forward syntax. Why not:

```
fooMessage { |a, b, c|
    ^someObject.forwardMethod(thisMethod)
}
```

Or even:

```
fooMessage { |a, b, c|
    ^someObject.forwardThisMethod
}
```

This says what it does.

telephon:
> I think `#(...)` for proto object is a good idea because most people try to use event `(...)` for object prototyping. This makes it easy to convert. 

well, my point was that `(...).proto` is just as easy and could be adapted to various variants. Don't you think this may need to be accounted for? Also you may want to take an existing dict and convert it to a prototype. 

> I think this one is used enough to make it worth it.

If it turns out to be that way, yes! And I am not saying that it won't be.

Also, I noticed that the ProtoObject doesn't support multiple inheritance like IdentityDictionary (using `proto`).

<hr>

So perhaps, we could make it two steps: the pure AbstractObject as one pull request and all infrastructure that uses it in a second one?

Of course I'll also help with getting this done in one block, but it may have a lot of subtle ramifications.

JordanHendersonMusic:
Okay I think splitting this up is a good idea then. Abstract object and the hierarchy changes are quite contained --- took a while to figure out because the code is very fragile in this part, but the change is actually quite simple.

I'd like to create a few use case to go with it, protobject being one, but a debug object with break points would be another.

I'll also create a separate pr for any forwarding optimisation. But I do think they should all be considered as a pack, so I'll make a multi issue thing.

---

So I've gotten a little confused about how we are using the term 'proto' and 'protobject'.

I'm not thinking of a set of protoobjects, are you?

I'm not thinking about 'object prototyping' either. This is because it isn't very popular and most people end up just using it to implement class - this is what JavaScript did and they introduced the class keyword.

I'm essentially trying to do the same thing. Create a way to make runtime 'classes'. 

Having a method to convert from a dictionary to one of these 'class' objects should be possible, but shouldn't be the main way you make them, as they aren't really dictionaries.

---

Re the optimisation of method forwarding. The way I'm thinking about doing it will not work with those examples. When you call a method, all the arguments are on the stack.
```
a.f(1, b: 2)
Stack: [a, 1, b, 2]
```
But inside 'a.f' they need to be rearranged, assigned in the correct places, allocated into arrays, copied around... This is the slow bit.

If we are directly forwarding, perhaps with some code that gets evaluated before, we can just reach into the stack and change the receiver - no argument faff.

The issue is that you cannot use, name, or touch the arguments. Now it is possible to walk the entire ast of the method to figure his out... But this is complicated. By introducing a new syntax where you can't even access the arguments, this becomes trivial, and would be quite a performance boost.

```
F {
   var h:

   h <- foo, bar { |selector| 
      ("before" ++ selector).postln
   } { |selector| 
      ("After" ++ selector).postln
   }
}

F(h).foo

"Before foo"
... runs h.foo
"After foo"

```

telephon:
> Okay I think splitting this up is a good idea then. Abstract object and the hierarchy changes are quite contained --- took a while to figure out because the code is very fragile in this part, but the change is actually quite simple.
> 
> I'd like to create a few use case to go with it, protobject being one, but a debug object with break points would be another.

Sure, we just need to have a detailed discussion about these, because we won't be able to change them later.

> I'll also create a separate pr for any forwarding optimisation. But I do think they should all be considered as a pack, so I'll make a multi issue thing.

Good!

> I'm essentially trying to do the same thing. Create a way to make runtime 'classes'.

Yes. If we use a dictionary internally as a method lookup, however, we'll have an option to use the multiple inheritance provided by dictionary (which does implement a version of object prototyping with some caveats), so that it may be good to take this into account.

> Having a method to convert from a dictionary to one of these 'class' objects should be possible, but shouldn't be the main way you make them, as they aren't really dictionaries.

I don't quite understand, maybe I have been unclear – we are also use `r` to define a routine, despite the fact that a function is not a routine: `r { 1.yield }`.

Maybe we should arrange a meeting where we go through the work I've done so far? Then we could decide together what should go into the use cases.

I've used the "Neutral" quark to collect ideas about the future AbstractObject.

> Re the optimisation of method forwarding. 
> ...
> The issue is that you cannot use, name, or touch the arguments. Now it is possible to walk the entire ast of the method to figure his out... But this is complicated. By introducing a new syntax where you can't even access the arguments, this becomes trivial, and would be quite a performance boost.

My idea was the following. Let's call `C(O)` a context object that implements an interface for an object `O`. When sent a message, it will try use it to select a method in `C`, or it will otherwise forward it to `O`. 

Now, if this idea is to be used freely, we don't want to create a lot of overhead each time we send a message to an object wrapped in a context. For example, deep inside a system, you may have $C_1(C_2(C_3(C_4(O))))$. So the idea is to forward blindly, and assume that the signature is up to the receiver. No messing with the stack apart from replacing the receiver. This is what I thought.

JordanHendersonMusic:
@telephon I've split this up into separate issues #7217 because there is a **lot** to consider!

I'm going to hide all the conversation so far as I think it will make it easier to focus on the implementation here. There are quite a few method implementations in AbtractObject and Object that should be considered fully.

I've made a specific issue for the message forwarding  #7216

JordanHendersonMusic:
Notes to self from dev meeting — thanks @telephon!
1. Turn this into `PreviewAbstractObject` that way we can ship the core functionality before working out all of the kinks. This ***must*** be completely revertible, breaking code that ***directly*** touches `PreviewAbstractObject` only.

telephon:
Great! Let's call it somethimng that starts with AbstractObject, e.g. `AbstractObjectPreliminary` or `AbstractObjectPreview`. Then one can find it with autocomplete.

capital-G:
@JordanHendersonMusic one question that came through my mind: If we introduce this change, would it still be possible to load a 3.14 class library with a 3.15 interpreter?
While we can't have upwards compatibility regarding the interpreter/class library (i.e. we do add primitives ocasionally), it is still nice to have backwards compatibility in case we need that.

JordanHendersonMusic:
No it will not be possible.

I was going to mention this last night but the moment passed....

The issue with having the class library as a 'package' and allowing the user to pick which one they want, is that the class library and the interpreter are intimately linked — you cannot have one without the other. This is just the way sc has been designed. For this reason I don't think your approach in baryon will work (with regards to the class library only, the server stuff sounds excellent!).

There was a suggestion to refactor the class library into smaller 'modules'. This way, the 'core' module would be something you get with  your interpreter, and the everything else can be modular. The issue is that the sc class library  is quite coupled and separating these classes in chunks isn't going to be trivial, and will probably involve breaking changes.

capital-G:
> There was a suggestion to refactor the class library into smaller 'modules'. This way, the 'core' module would be something you get with your interpreter, and the everything else can be modular. The issue is that the sc class library is quite coupled and separating these classes in chunks isn't going to be trivial, and will probably involve breaking changes.

Thanks for clarifying that.
I think one way forward could be indeed to create smaller chunks of the class library and its primitives through Gluons (once Gluons and a package manager have been stabilized), making SC more a distribution like debian which would allow to entangle everything over time - and people could start developing libraries which are replacing existing ones "in place" (see e.g. JitLib gui v2)

I think a very cool way would be if we at some point allow for wasm binaries as gluons and server extensions - I don't know if this would influence any design decisions currently? My idea would be to add something like https://github.com/bytecodealliance/wasm-micro-runtime and if we load a wasm binary we use this to run the gluon code. This would allow to use wasm binaries as a fallback if no system binaries are available, which is probably a cool thing regarding longevity.

JordanHendersonMusic:
> and its primitives through Gluons (once Gluons and a package manager have been stabilized),

Anything that calls a primitive needs to be in core and coupled to the interpreter. Having a C api is a way of achieving performance gains, but can't be used to replace all primitives. Actually, calling a primitive (which can't be replaced by a c function call) is a pretty good requirement to be in core and anything that doesn't do that could be considered for removal. There are a few exceptions to this for classes like `RawPointer` or `Frame` which don't 'do' anything on the sc side, but are used just to represent compiler internals that you can query.

The issue that will arise with moving primitives to a c api is that only relatively basic types can be exposed. We don't want classes to be exposed as that requires us to share the internal layout of objects (which is private to the interpreter). This is quite a serious limitation and will bloat 'core'.

>  JitLib gui v2

The jitlib stuff is perhaps one of the cases where there is the least coupling and this would be simple, but in general, I don't think separating out the class library is going to be easy — sc just hasn't been designed with this in mind.... Not to say it isn't possible or that it shouldn't be done, more just to wish you luck!

You might be able to have the wasm interpreter in a c library too?

Anyway, we should probably move this conversation to another issue!

JordanHendersonMusic:
So I've reworked this and made it smaller. It would be good to really consider what method I've put in AbstractObject and the relationship with Object.

Here is a quick little quark playground with some examples in. It might be useful if others contributed to that and we tried to do a bunch of testing with all of them. 

https://github.com/JordanHendersonMusic/AbstractObjectPlayground 

Is it possible to do unit testing in a quark?

telephon:
> Is it possible to do unit testing in a quark?

yes of course. You just add the test classes.

SCClassLibrary/Common/Core/AbstractObject.sc
```diff
@@ -1,54 +1,41 @@
...
-//          Additionally, this message is REQUIRED to throw an error as the jump byte codes have no way to skip the true and false branches.
 AbstractObject {
+	/////// 1. Core and Interpreter functionality, do not override.
 	*new { |maxSize = 0| _BasicNew; ^this.sc_abstract_object_primitive_failed }
 	*newCopyArgs { | ... args, kwargs | _BasicNewCopyArgsToInstVars; ^this.sc_abstract_object_primitive_failed }
 
-	// Hashing, AbstractObject support identity, this is required by the interpreter and cannot actually be overridden.
-	// When you put AbstractObjects into an identity dictionary the interpreter uses its own implementation, this should match or things will get confusing.
+	// AbstractObject support identity, this is required by the interpreter and cannot actually be overridden.
+	// When you put AbstractObjects into an identity dictionary the interpreter uses its own implementation, this should match the sclang implementation or things will get confusing.
```
telephon: line 13 Since there are different reasons for having identity, this can be left out.

line 14 I wouldn't think of identity as a set. The comment is a bit confusing, could be left out as well.

I think you can leave out 13 and 14.

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -1,54 +1,41 @@
...
-	// AbstractObject does NOT support dependencies (it can be added in a subclass but must register with Object.dependantsDictionary)
-	// AbstractObject does NOT support unique methods, these are easy to implement in a subclass, and unnecessary in ProtoObject.
-	// AbstractObject does NOT support gc reporting nor does it expose any compiler internal methods.
-	// AbstractObject does NOT support archiving, this can be implemented in the subclass.
-	// AbstractObject does NOT support switch as it requires equality.
+	// Interpreter will engage in undefined behavior if this method does not throw.
+	mustBeBoolean {  MustBeBooleanError(nil, "AbstractObject").throw }
+
+
+	/////// 2. Main interface, override this method.
```
telephon, line 35: "override these methods"?

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -1,54 +1,41 @@
...
+
+	// Interpreter will segfault if this isn't provided.
+	// Note: you will need to override this to anything useful will abstract objects,
+	//   hence why it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
```
telephon, line 39: "hence it throws"

SCClassLibrary/Common/Core/AbstractObject.sc
```diff
@@ -1,54 +1,41 @@
...
+	// Interpreter will engage in undefined behavior if this method does not throw.
+	mustBeBoolean {  MustBeBooleanError(nil, "AbstractObject").throw }
+
+
+	/////// 2. Main interface, override this method.
+
+	// Interpreter will segfault if this isn't provided.
+	// Note: you will need to override this to anything useful will abstract objects,
+	//   hence why it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
+	// "AbstractObject" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
```
telephon: This is already the case in:

```
mustBeBoolean {  MustBeBooleanError(nil, "AbstractObject").throw }
```

Maybe explain in the first occurence?

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -1,54 +1,41 @@
...
+	// "AbstractObject" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
+	doesNotUnderstand { |selector ...args, kwargs| SubclassResponsibilityError("AbstractObject", thisMethod, AbstractObject).throw }
+
+
+	/////// 3. Helper methods and primitive implementations
 
-	// Helper methods and primitive implementations.
+	// These methods may be used *only* to implement other abstract objects, do not call them from outside the definition of an abstract object.
+	// They are here so that other's may use primitive implementations, without making the primitives themselves publicly available.
```
telephon, line 48: other's -> other objects (?)

SCClassLibrary/Common/Core/AbstractObject.sc
```diff
@@ -1,54 +1,41 @@
...
+	/////// 3. Helper methods and primitive implementations
 
-	// Helper methods and primitive implementations.
+	// These methods may be used *only* to implement other abstract objects, do not call them from outside the definition of an abstract object.
+	// They are here so that other's may use primitive implementations, without making the primitives themselves publicly available.
 
-	// These errors don't actually pass on 'this', this is because printing 'this' requires the object to be able to respond.
-	// If we passed 'this', it is likely it get the interpreter stuck in an infinite loop trying to print an error.
-	// You may wish to reimplement these to pass 'this' if you know that printing is a safe operation.
+	// "AbstractObject" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
```
telephon: maybe explain only once.

SCClassLibrary/Common/Core/AbstractObject.sc
```diff
@@ -65,12 +52,11 @@ AbstractObject {
...
 
 	// Needed when we want to do the normal method calling in an AbstractObject.
-	// We do not provide 'tryPerform' as it required 'respondsTo', which is bad design as it doesn't allow 'doesNotUnderstand' to take effect.
-	// Since 'multiChannelPerform' and 'performWithEnvir' do not require primitives, they are not provided either.
 	// In most cases, the 'perform*' messages should defer to these as most of the time, you actually want to implement doesNotUnderstand.
 	sc_abstract_object_perform_args { |selector, args, kwargs|	_ObjectPerformArgs;	^this.sc_abstract_object_primitive_failed }
 	sc_abstract_object_super_perform_args { |selector, args, kwargs| _ObjectSuperPerformArgs; ^this.sc_abstract_object_primitive_failed }
 	sc_abstract_object_perform_msg { |msg| _ObjectPerformMsg; ^this.sc_abstract_object_primitive_failed }
 	sc_abstract_object_perform { |selector ... args| _ObjectPerform; ^this.sc_abstract_object_primitive_failed }
+	// Note that this is used to implement syntax, it is very strange to change this.
```
telephon: this comment is hard to understand.

lang/LangSource/PyrObject.cpp
```diff
@@ -1606,9 +1616,14 @@ void initClasses() {
...
 
-    memcpy(slotRawObject(&class_class->classptr->iprototype)->slots, slotRawObject(&class_class->iprototype)->slots,
+    memcpy(slotRawObject(&class_abstract_object_meta->iprototype)->slots,
+           slotRawObject(&class_class->iprototype)->slots, sizeof(PyrSlot) * classClassNumInstVars);
+    memcpy(slotRawSymbolArray(&class_abstract_object_meta->instVarNames)->symbols,
+           slotRawSymbolArray(&class_class->instVarNames)->symbols, sizeof(PyrSymbol*) * classClassNumInstVars);
+
+    memcpy(slotRawObject(&class_class_meta->iprototype)->slots, slotRawObject(&class_class->iprototype)->slots,
            sizeof(PyrSlot) * classClassNumInstVars);
-    memcpy(slotRawSymbolArray(&class_class->classptr->instVarNames)->symbols,
```
telephon, line 1611: So the changes from `class_class->classptr` to `class_class_meta` are (somewhat) unrelated cleanups, right? This variable wasn't used consistently in the previous version.

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -0,0 +1,62 @@
+// AbstractObject represent the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
+// This make it very useful as a wrapper class and to implement some specific features.
```
telephon, line 2: line 2 can be left out (or just the "very" removed)

JordanHendersonMusic:
I've renamed it to ExperimentalAbstractObject. This PR just needs some documentation now. Which I will try and do in the new year.

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -0,0 +1,74 @@
+// ExperimentalAbstractObject represent the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
+// This make it very useful as a wrapper class and to implement some specific features.
+
+// !!! --- PLEASE NEVER ADD A NORMAL PASCAL CASE METHOD TO ABSTRACT OBJECT --- !!!
+// The whole point of this class is to be minimal, ExperimentalAbstractObjects are not Objects, allowing those who subclass it to create their own definition of what an object is and isn't.
+
+ExperimentalAbstractObject {
```
telephon, line 7: Can we call it `AbstractObjectExperimental` ? then it will be found more easily later. It is ok like as it is as well.

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -0,0 +1,74 @@
...
+
+	/////// 1. Core and Interpreter functionality, do not override.
+
+	*new { |maxSize = 0| _BasicNew; ^this.sc_abstract_object_primitive_failed }
+	*newCopyArgs { | ... args, kwargs | _BasicNewCopyArgsToInstVars; ^this.sc_abstract_object_primitive_failed }
+
+	// ExperimentalAbstractObject support identity, this is required by the interpreter and cannot actually be overridden.
+	// When you put ExperimentalAbstractObjects into an identity dictionary the interpreter uses its own implementation, this should match the sclang implementation or things will get confusing.
```
telephon, line 15: 15/16 can be left out.

telephon (commented):
just some cleaning up. I'll also test the branch in a next step, so I only comment now.

lang/LangSource/PyrObject.cpp
JordanHendersonMusic: That's right, it is just easier to refer to it by name rather than through a member pointer. Up on line 1574.

JordanHendersonMusic: That's right, it is just easier to refer to it by name rather than through a member pointer. Up on line 1574.

telephon: I also think when a new name is introduced, it should be used. If there are instances of the signifier and the signified used arbitrarily next to each other, it will always prompt the question what the signifier really means: you suspect that there is a difference that is not there.

telephon:
@JordanHendersonMusic what do you think, should we try to finish this one? It would be good to have some time to test it before we publish the new version. 

I had a few little things, but they are really just naming.and a bit of documentation.

JordanHendersonMusic:
Ah yes... this one...

Honestly I don't know. 

The problem is that using this 100% correctly is going to be very hard due to all the subtle methods you have to provide. 

It isn't abstract object that has the issue, it's all the classes that might inherit from it. 

One example is performBinaryOp, should this go through doesNotUnderstand or not? If it does, it will break.

telephon:
Well, sure, of course. But I think the point was that we try it now. I have the quark that builds on the assumptions of something close to this one, and the first thing could then be to make a version that uses it. Then we'll see. If it all doesn't pan out, we can still move back later – it is experimental.

telephon:
As long as the garbage collector etc. works, it should be fine.

telephon (approved):

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -0,0 +1,74 @@
+// AbstractObjectExperimental represent the minimum interface needed to get the interpreter to work (it will cause the IDE to post an error as it doesn't support printing).
+// This make it very useful as a wrapper class and to implement some specific features.
```
JordanHendersonMusic:
```suggestion
// This make it useful as a wrapper class and to implement some specific features.
```

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -0,0 +1,74 @@
...
+	// You cannot override these methods.
+	isNil { ^false }
+	notNil { ^true }
+	? { |obj| ^this }
+	?? { |obj| ^this }
+	!? { |obj| ^obj.value(this) }
+
+
+
+	/////// 2. Main interface, override this method.
```
JordanHendersonMusic:
```suggestion
	/////// 2. Main interface, override these method.
```

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -0,0 +1,74 @@
...
+
+	// Note: Interpreter will segfault if this isn't provided so a default is given.
+	// You will need to override doesNotUnderstand to anything useful will abstract objects,
+	//   hence why it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
```
JordanHendersonMusic:
```suggestion
	//   hence it throws a SubclassResponsibilityError rather than a DoesNotUnderstandError.
```

SCClassLibrary/Common/Core/ExperimentalAbstractObject.sc
```diff
@@ -0,0 +1,74 @@
...
+	// "AbstractObjectExperimental" is passed as a string here, rather than 'this', as 'this' may not support the 'toString' method, which would cause the error handling to get stuck in a loop.
+	doesNotUnderstand { |selector ...args, kwargs| SubclassResponsibilityError("AbstractObjectExperimental", thisMethod, AbstractObjectExperimental).throw }
+
+
+
+	/////// 3. Helper methods and primitive implementations
+
+	// These methods may be used *only* to implement other abstract objects, do not call them from outside the definition of an abstract object.
+	// They are here so that other's may use primitive implementations, without making the primitives themselves publicly available.
```
JordanHendersonMusic:
```suggestion
	// They are here so that other objects may use primitive implementations, without making the primitives themselves publicly available.
```

JordanHendersonMusic:
@telephon thanks for that!!

PR #7206, opened 2025-10-26
sclang: fix keyword arguments bug

JordanHendersonMusic:
## Purpose and Motivation

In the previous kwargs commit, we had agreed you couldn't set the 'args' variable directly, i.e., this shouldn't work.
```
f = { |...args, kwargs|
    [args, kwargs]
};
f.(args: 10) != 10 // should be [[], [args: 10]]
```
This sometimes worked (it worked in a block as above, but didn't work in a method), instead the 'args' parameter would be omitted from the 'kwargs' array.

This was a bug (off by one error) made by me, but it is a simple fix. I've also added a bunch more testing to ensure this won't happen again.

----

EDIT by @dyfer:

To clarify, if someone comes back to this: the issue can only be reproduced in a method of a class. With a following class defined:

```supercollider
KwargsTest {
	foo { |...args, kwargs|
		^[args, kwargs]
	}
}
```
This is the correct behavior:
```supercollider
KwargsTest().foo(args: 10); // returns [[], [args: 10]]
```
Before this PR, the method would incorrectly return `[[], []]`.

----

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -402,10 +402,10 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
     int keywordArgumentSize = 0;
 
     const auto findKeywordArgIndex = [=](const PyrSymbol* argName) -> std::optional<uint32_t> {
-        for (size_t namei = 0; namei < methNumNormArgs; ++namei) {
-            const auto indexOffsetFromThis = namei + (isMethod ? 1 : 0);
-            if (methArgNames[indexOffsetFromThis] == argName)
-                return indexOffsetFromThis;
+        for (size_t namei = (isMethod ? 1 : 0); namei < methNumNormArgs; ++namei) {
+            const auto considering = methArgNames[namei];
+            if (considering == argName)
+                return namei;
         }
         return std::nullopt;
     };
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
@dyfer @capital-G I strongly think this should be a 3.14.1 fix!

prko:
> `f.(args: 10) != 10 // should be [[], [args: 10]]`

Regarding `[args: 10]`: 
I believe this may be misleading. It ought to be one of the following:
- `[args, 10]`
  This seems preferable. SuperCollider automatically converts `[args: 10]` into `[args, 10]`.
- `(args: 10)`
    unlikely...
- `(args: [10])`
    unlikely...
- `[args, [10]]`
    unlikely...

JordanHendersonMusic:
@prko I don't understand your comment. The code should return `[[], [args: 10]]` ... which is the same as `[[], [\args, 10]]`

prko:
```
f.(args: 10)
```
returns:
```
-> [[], [args, 10]]
```
---
```
f.(args: 10).postln
```
returns:
```
[[], [args, 10]]
-> [[], [args, 10]]
```
---
Also...
```
[\a: 3] == [\a, 3] // -> true
```
```
[\a: 3] // -> [a, 3]
[\a, 3] // -> [a, 3]

[\a: 1, \b: 2] // -> [a, 1, b, 2]
```
---

>@prko I don’t understand your comment. The code should return [[], [args: 10]], which is the same as [[], [\args, 10]].

As you mentioned, they are indeed equivalent, and you are correct.

However, I believe it is clearer to show exactly what sclang returns, because otherwise one step in the process is omitted.

dyfer (approved):

capital-G (approved):

PR #7219, opened 2025-11-02
sclang, classlib: ProtoObject, a better runtime 'class' implementation

### LINKED ISSUE #7214: Event is a poor choice to implement object prototyping/runtime classes with
## Steps to reproduce

```supercollider
(numChannels: \meow).numChannels // 1
```

We need a way to build runtime class like things, doesn't have to be object prototyping (javascript has this but most people just use it to implement classes), but it should work as expected. There are numerous issues on the forum about this.

JordanHendersonMusic:
Depends on #7205 so the change log looks much bigger than expected.

## Purpose and Motivation

Closes #7214

We need some way to make runtime class like object. Previously many people have used `Event`, but it was not made for this and has many subtle issues.

This is my proposal for how this should be done.

Here is a bunch of examples of how this looks. I've opted to use the `#(...)` syntax as it makes converting from existing `Event` based solutions to the new one very easy.

```
a = #(meow: 1);
a.meow \\ 1
```

```
a = #(numChannels: \meow);
a.numChannels \\ \meow --- with event this would have returned 1
```

```
a = #(meow: 10)
a.woof = 4;
a.woof // 4
a.meow // 10
```

```
a = #(meow: 10, protoObjectLock: true)
a.woof = 10; // Prints error
```

```
p = #(speak: {|self| "I say %".format(self.word) });
b = #(super: p, word: \meow);

b.speak // I say meow
```
```
a = #(doesNotUnderstand:   "I'm sorry Dave, I'm afraid I can't do that."} );
a.shutdown //  "I'm sorry Dave, I'm afraid I can't do that."
```

```
a = #(value: \meow );
a.(); // \meow
```

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


JordanHendersonMusic:
closing because the abstract object needs way more work before doing this.

PR #7220, opened 2025-11-02
Document object interface by splitting it into tiers

### LINKED ISSUE #7218: Categorise and document the object interface in tiers
## Motivation

Object is huge.

Some of the messages are hard requirements of  the interpreter and not implementing them in the correct way leads to ***undefined behaviour***.
* `doesNotUnderstand` if this isn't implemented, then the interpreter will either segfault or call the wrong method.
* `mustBeBoolean` this method is **required** to throw #7208
* `identityHash`, `===`, and `!==` is also a part of this tier as collections like identity dictionary implement it in the back-end meaning you can't actually change it. If this isn't defined, it won't cause undefined behaviour.

If other messages aren't implemented the core interpreter will crash — this isn't undefined, but affect core functionality the user is *probably* expecting to work. Crashes often include infinite loops and other hard locks.

Printing
* `asString`
* `storeOn`

Error reporting/exceptiong. 
Note these take place in a **no-throw** context as you cannot throw while already throwing.
* `dump*
* `isException`, informs sc whether an exception has occured.

Then there is functionality like `yield`.

...

For any object to be a complete `Object` it must implement all of the functionality, but with `AbstractObject` we need to be able to talk about and write code with *incomplete* objects, allowing the class author to fulfil the `Object` interface how they see fit. 

Lots of documentation, probably as comments in the code. 

Some refactoring, introducing very explicit messages for some of the top tier messages.

E.g. 
Should `asString` really be the message we call when we want to post something to the post window? Are we really doing a full string conversion?

My suggestion would be something like this...
```
Object {
   asStringForPostWindow { ^asString }
   asStringForExceptionPrint { ^asString } // this one is no throw!
   ...
}
```

JordanHendersonMusic:
## Purpose and Motivation

Closes #7218
Depends on #7205

The only changes in this PR of note are in `Object.sc`. See https://github.com/supercollider/supercollider/pull/7220/commits/265b84a42f6db19ade77e1aaf71746e5e6439f26

Object is huge, this breaks a part the methods into tiers depending on how close they are to the vm and interpreter.

The following methods are introduced to make the distinction between the core behaviour of the interpreter and the Object interface cleaner:
*  `asStringForPostWindow` { ^this.asString } 
* * Printing into the post window
* `asStringForError` { ^this.asString }
* * Printing for errors. These messages CANNOT throw. If an Object's asString method can throw, the following should be overloaded to a safe implementation.
* `asStringForErrorPostln` { ^this.dump }

## Types of changes

- Documentation through comments
- Reordering methods

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Core/AbstractObject.sc
+++ b/SCClassLibrary/Common/Core/AbstractObject.sc
@@ -1,0 +1,76 @@
+// AbstractObject is often not valid in many standard sc contexts, e.g., it doesn't support asString so can't be shown in the post window.
+// As a result, it isn't supposed to be used directly, but as a way to implement other objects that do work in standard sc contexts, ProtoObject and AbstractWrapper both do this.
+// There is also an issue with 'keywords', sc lang doesn't have any, but being able to 'yield' or 'copy' an object isn't really a part of the object interface, but more a fundamental feature of how users expect the language to work.
+// This is a fundamental awkwardness with smalltalk and strictly OOP languages.
+// For this reason, a few of these methods are implemented here, again, AbstractObject is not supposed to be used by itself, but inherited from — AbstractObject is an *incomplete* object.
+// There is an additional issue with primitives, to help combat this the following 'sc_abstract_object*' methods are introduced to easily enable abstract objects to opt into certain expected behaviour,
+//     note, this doesn't cover all standard sc behaviour, just cases that are impossible to implement otherwise.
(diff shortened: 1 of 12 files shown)


JordanHendersonMusic:
Changed my mind, think this would be better in some documentation elsewhere.

PR #7223, opened 2025-11-03
Add support for multi‑line strings in HelpBrowser’s code editor

### LINKED ISSUE #7210: Code Blocks in Help Browser Fail to Parse "" Pairs Spanning Multiple Lines
## Environment

* Operating system: macOS

## Steps to reproduce

1. Install `DynGen` by following the instructions at https://github.com/capital-G/DynGen.
2. In the help documentation, locate the fifth code block above the note.
3. Evaluate the first code snippet in that block.
4. Then, evaluate the second code snippet in the same block.

- Expected: no error
- Actual:
   ```
   ERROR: syntax error, unexpected '(', expecting end of file
     in interpreted text
     line 34 char 1:
   
     (
     ^
     Ndef(\g, {
   -> nil
   ```

The same code block behaves as expected in the code editor. Please refer to the attached video for reference:
https://www.dropbox.com/scl/fi/ywfq3ztm8h6o8vooybyto/Parsing-Issue-with-Code-Blocks-in-Help-Browser.mov?rlkey=bndlsdctugqcbxuqr85lk3swx&st=dfltxmqu&dl=0

I'm reporting this issue here rather than in the repository, as I believe it stems from the schelp system itself.

prko:
fixes: #7210
## Purpose and Motivation

Currently, the code editor in the HelpBrowser does not accept multi‑line strings, which results in syntax errors when using Cmd/Ctrl+Enter, as illustrated in #7210.  
This PR resolves the issue.

However, the colour of multi‑line strings remains black, and this should also be addressed.  
I believe that a proper implementation of """ would be the best solution for this.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/HelpSource/static/editor.js
+++ b/HelpSource/static/editor.js
@@ -18,7 +18,7 @@ const init = () => {
             { regex: /^\$\\?./, token: 'char' },
             { regex: /^~\w+/, token: 'env-var' },
             { regex: /^\/\/[^\r\n]*/, token: 'comment single-line-comment' },
-            { regex: /"(?:[^\\]|\\.)*?(?:"|$)/, token: 'string' },
+            { regex: /"(?:[^\\]|\\.)*?"/, token: 'string' },
             { regex: /^[-.,;#()\[\]{}]/, token: 'text punctuation' },
             { regex: /\/\*/, push: 'comment', token: 'comment multi-line-comment' },
             { regex: /^[+\-*/&\|\^%<>=!?]+/, token: 'text operator' },


dyfer:
@capital-G @JordanHendersonMusic I haven't really worked with CodeMirror. How does this PR look to you?

JordanHendersonMusic:
It's impossible to write an SC lexer in regex, the correct solution would be to use the new one, but I am unsure if that is possible because this is JavaScript.

dyfer:
Okay, so what should we do with the code editor in HTML?

prko:
This seems to be resolved by https://github.com/supercollider/supercollider/pull/7546.

JordanHendersonMusic:
> Okay, so what should we do with the code editor in HTML?

Do you mean JavaScript?

Well we need to see if we can compile the lexer to wasm and use it. If not, it will always be slightly broken.

dyfer:
> Do you mean JavaScript?

Yes :)



> Well we need to see if we can compile the lexer to wasm and use it. If not, it will always be slightly broken.

That's what I was thinking too (slightly broken).  My question was how to proceed further, I guess next on the list would be trying the wasm approach...?

JordanHendersonMusic:
Yeah, I'd actually be in favour of not even trying to fix this regex thing. The more you look at it, the more broken you will realise it is. It's just not worth the energy to try and fix this.

Same holds for the ides syntax highlighting.

If it turns out we can't use wasm for whatever reason then we might have to reassess.

I'm pretty busy right now, but if someone wants to know how to use the lexer, check out the standalone in the sc_lexer directory. Should be straight forward (I hope)!

PR #7224, opened 2025-11-03
sclang: non recursive method macro of double dispatch

### LINKED ISSUE #7225: Double dispatch methods `performBinar..*` are broken as they don't 'perform'
## Environment

* SuperCollider version:
* Operating system:

## Steps to reproduce

```supercollider

2.performBinaryOpOnSimpleNumber('+', 2) // errors

```

This should work, perform methods should perform. In fact, the design of this does not work with doesNotUnderstand, in sc, we must let methods fail. It is impossible to write the `performBinary*` in small talk languages. Then only way to break this recursion is by checking for it. 

I propose we rewrite all double dispatch. We need a non-recursive method for this.

Then double dispatch can be implemented like this

```supercollider
+ { |other, adverb|
	_Add; // primitive escape, this is the exit path
	^this.performBinaryDoubleDispatch('+', other, adverb)
}

invertBinaryOperands #non_recursive { |selector, other, adverb|
	^other.performList(selector, this, adverb)
}

performBinaryDoubleDispatch { |selector, other, adverb|
	var r;
	try {
		r = this.invertBinaryOperands('+', other, adverb)
	} { |er|
		if (er.isKindOf(RecursionError)){
			r = other.as(this.class).performList('+', other, adverb)
		} { 
			er.throw
		}
	};
	^r
}
```

JordanHendersonMusic:
## Purpose and Motivation
Closes #7225

The existing way of doing double dispatch through the performBinaryOp* methods doesn't work because they don't actually 'perform', likewise, in sc, you can't know if something will perform ahead of time, as doesNotUnderstand hasn't been called.

This is very important as we move towards AbstractObject as being able to call a perform* method and it actually perform, going through doesNotUnderstand, is imperative.

Implements a special kind of method that will throw an error if recursion is detected. This simply allows us to flip the arguments in binary operators and try them without getting stuck.

This adds method macros as a new syntax.

```
Object {
   invertForBinaryCall #non_recursive {  .... }
}
```

This inserts bytecodes to, stop tail call recursion, then walk up the frame, making sure we haven't entered this method before.

This PR also removes `mustBeBoolean` and introduces a new method `fatalInterpreterError` which is called when something (like invalid recursion) has occurred and it is impossible to continue.

## Types of changes

- Bug fix
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Core/Object.sc
+++ b/SCClassLibrary/Common/Core/Object.sc
@@ -16,6 +16,15 @@ Object {
 		^this.primitiveFailed
 	}
 
+	// This method is REQUIRED to throw (or perform some non-local return) as executing the following instructions in the frame will lead to undefined behaviour.
+	fatalInterpreterError { |nameOfError|
+		switch(nameOfError)
+		{ 'mustBeBoolean' } { MustBeBooleanError(this, nil).throw }
+		{ 'invalidRecursion' } { Error("Invalid recursion was discovered").throw } {
+			Error("Some unknown fatal error has occured: " ++ nameOfError).throw
+		}
+	}
+
 	// debugging and diagnostics
 	dump {
 		_ObjectDump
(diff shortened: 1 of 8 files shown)


JordanHendersonMusic:
closing might come back to this, but with the reduced abstract object design this isn't needed.

PR #7227, opened 2025-11-03
TempoClock.beats_ forward to TempoClock.default.beats_

HotwheelsSisyphus:
As stated in the summary; brought up by @jamshark70 on the forum (https://scsynth.org/t/misleading-help-bit-in-tempoclock-forwarding/12526). If there are any good reasons to _not_ also have the setter forward to the default TempoClock (I can't think of any, but who knows), this would imply a doc change instead of this classlib change - please let me know if that is the case.   

Also organizes the other lines in the Class methods that forward to default, so this is less of a jumble hopefully.

the actual addition is this:
```
	*beats_ { |beats|  ^TempoClock.default.beats_(beats) }
```
The other changed lines are just cosmetics: The order in which the methods were listed was pretty random, so I grouped/reordered them a bit while I was in there. I double checked to make sure I didn't actually cut a method there.

### DIFF
--- a/SCClassLibrary/Common/Core/Clock.sc
+++ b/SCClassLibrary/Common/Core/Clock.sc
@@ -370,39 +370,42 @@ elapsed time is whatever the system clock says it is right now. elapsed time is 
 	}
 
 	// these methods allow TempoClock to act as TempoClock.default
+	*isRunning { ^this.default.isRunning }
 	*stop { TempoClock.default.stop	}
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Thanks for the fix!

I double-checked if everything has been transferred correctly.

Next time it would be cool if both tasks, adding the `beats` setter and re-arranging the order would be two distinct commits, this makes reviewing a bit easier :)

Issue #7217, opened 2025-11-02
AbstractObject Project


JordanHendersonMusic:
This is an issue to collect the existing issues and PRs surrounding AbstractObject and what it entails. 

Feel free to discuss other ideas/feature/improvements here, then create a new issues and I'll add them to the list.

Issues
* [ ] #3467 
* [ ] #7214
* [ ] #7215
* [ ] #7216
* [ ] #7218
* [ ] #7208
* [ ]  #7225
* [x] #7226

PRs:
* [ ] #7205
* [ ] #7219
* [ ] #7224
* [ ] #7228

dyfer:
In the interest of keeping discussions separate, I'm putting my response to https://github.com/supercollider/supercollider/pull/7228#issuecomment-3489571190 here (feel free to move it elsewhere)

I'm not necessarily holding on to the notion that we have to have a reliable `respondsTo` method, but I was wondering whether the mechanism of object prototyping through overriding `doesNotUnderstand` is something we want to keep moving forward as the only/default mechanism. What if AbstractObject had `addInstanceMethod` / `removeInstanceMethod` special methods, which would collect additional methods in a defined/known place (a Dictionary or something)? That would allow for more insight into the methods, and possibly have `respondsTo` query them as well.

Again, this is not strictly about `respondsTo`, but about overriding error handling. When I first learned that the way to extend a class is to override `doesNotUnderstand` error, it felt a bit like a hack to me. The fact that we touch error handling when trying to extend the interface and that we need to re-throw the error through custom logic feels fragile, albeit it does work.

JordanHendersonMusic:
The `doesNotUnderstand` trick is one of the key features of a smalltalk, I think it should have been called `runtimeMethodDispatch` or something else because it isn't at the error state yet.

There is no reason why you couldn't make `respondsTo` work in a protoobject by implementing it — ProtoObject has a dictionary and you could absolutely make it work. 

The issue is that sometimes it isn't possible to do this and it imposes yet another axiomatic requirement on all objects. AbstractObject is supposed to have the absolute minimum interface... compared with Objects 400 odd methods. 

But calling `respondsTo` isn't necessary, as you always catch, so it should just be avoided. 

We also haven't addressed cases like this...
```
F {
   foo { DoesNotUnderstand(..).throw }
}
```
Now `F.respondsTo(\foo)` is true, and it throws an error... what if it throws a `ShouldNotImplementError`? The only way to reliably figure it out, is to call the message.

[...]

PR #7228, opened 2025-11-04
classlib: limit use of `respondsTo` and post warning

JordanHendersonMusic:
## Purpose and Motivation

`respondsTo` is bad smalltalk design as it doesn't allow of `doesNotUnderstand` to take effect. While this is currently a minor issue in the class library, as we move towards AbtractObject, this will become a major pain.

This PR removes the vast majority of uses from the class library as they are trivially replaceable with `tryPerform`, the remaining places require dependency inversion to solve.

Also makes `respondsTo` post a warning when used, this can be disabled.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Object.schelp
+++ b/HelpSource/Classes/Object.schelp
@@ -98,6 +98,13 @@ code::
 
 method::respondsTo
 
+warning:: 
+This method is NOT recommended as it breaks polymorphism and doesn't allow the effects of code::doesNotUnderstand:: to take place.
+Using this method means the code is subtly broken in ways that will be hard to predict!
+It will post a warning when called.
+Instead see code::tryPerform:: which in the vast majority of use cases is the correct approach.
+::
+
 Answer a link::Classes/Boolean:: whether the receiver understands the message selector or array of message selectors.
 
 code::
@@ -110,6 +117,9 @@ code::
 
 argument::aSymbol
 A selector name. Must be a link::Classes/Symbol::.
+
+argument::postWarning
+A boolean, can be used to disable the warning if it is absolutely necessary.
 
 method::isKindOf
 
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -544,7 +544,7 @@ SynthDef {
 			var inputs, ugenName;
 			if (ugen.inputs.notNil) {
 				inputs = ugen.inputs.collect {|in|
-					if (in.respondsTo(\dumpName)) { in.dumpName }{ in };
+					in.tryPerform(\dumpName) ?? { in }
 				};
 			};
 			[ugen.dumpName, ugen.rate, inputs].postln;
--- a/SCClassLibrary/Common/Audio/SynthDefOld.sc
+++ b/SCClassLibrary/Common/Audio/SynthDefOld.sc
@@ -90,10 +90,7 @@ SynthDefOld : SynthDef {
 			}
 		} { // catch
 			arg e;
-			if (file.respondsTo(\close)) {
-				file.close;
-			};
-			Error("SynthDefOld: could not write def: %".format(e.what())).throw;
+			file.tryPerform(\close) ??  { Error("SynthDefOld: could not write def: %".format(e.what())).throw };
 		}
 	}
 
--- a/SCClassLibrary/Common/Audio/UGen.sc
+++ b/SCClassLibrary/Common/Audio/UGen.sc
@@ -390,11 +390,11 @@ UGen : AbstractFunction {
 		CheckBadValues.perform(this.methodSelectorForRate, this, id, post);
 	}
 
-	*methodSelectorForRate { arg rate;
+	*methodSelectorForRate { |rate|
 		if(rate == \audio,{ ^\ar });
 		if(rate == \control, { ^\kr });
 		if(rate == \scalar, {
-			if(this.respondsTo(\ir),{
+			if(this.respondsTo(\ir, postWarning: false),{
 				^\ir
 			},{
 				^\new
@@ -463,11 +463,14 @@ UGen : AbstractFunction {
 		if (rate == \demand, { ^3 });
 		^0 // scalar
 	}
+
+	// TODO: this method should be implemented in subclasses instead.
+	// Using respondsTo like this breaks polymorphism and is bad smalltalk design.
 	methodSelectorForRate {
 		if(rate == \audio,{ ^\ar });
 		if(rate == \control, { ^\kr });
 		if(rate == \scalar, {
-			if(this.class.respondsTo(\ir),{
+			if(this.class.respondsTo(\ir, postWarning: false),{
 				^\ir
 			},{
 				^\new
--- a/SCClassLibrary/Common/Collections/Association.sc
+++ b/SCClassLibrary/Common/Collections/Association.sc
@@ -5,8 +5,9 @@ Association : Magnitude {
 		^super.newCopyArgs(key, value)
 	}
 
-	== { arg anAssociation;
-		^anAssociation respondsTo: \key and: { key == anAssociation.key }
+	== { |other|
+		var otherKey = other.tryPerform(\key); 
+		^otherKey !? { otherKey == key } ?? { false }
 	}
 
 	hash {
--- a/SCClassLibrary/Common/Collections/Dictionary.sc
+++ b/SCClassLibrary/Common/Collections/Dictionary.sc
@@ -23,9 +23,7 @@ Dictionary : Set {
 		^nil
 	}
 	trueAt { arg key;
-		var val = this.at(key);
-		if(val.respondsTo(\booleanValue).not) { ^false };
-		^val.booleanValue
+		^this.at(key).tryPerform(\booleanValue) ?? {false}
 	}
 	falseAt { arg key;
 		^this.trueAt(key).not
@@ -551,10 +549,7 @@ IdentityDictionary : Dictionary {
 
         if (selector.isSetter) {
             selector = selector.asGetter;
-            if(this.respondsTo(selector)) {
-                warn(selector.asCompileString
-                    + "exists as a method name, so you can't use it as a pseudo-method.")
-            };
+			// Impossible to tell if you can or can't use a method name ahead of time.
             ^this[selector] = args[0];
         };
 
--- a/SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
+++ b/SCClassLibrary/Common/Collections/EnvironmentRedirect.sc
@@ -193,7 +193,7 @@ EnvironmentRedirect {
 	// networking
 	dispatch_ { arg disp;
 		dispatch = disp;
-		if(disp.respondsTo(\envir_)) { disp.envir_(this) };
+		disp.tryPerform(\envir_, this)
 	}
 	envir_ { arg argEnvir;
 		envir = argEnvir;
--- a/SCClassLibrary/Common/Collections/Event.sc
+++ b/SCClassLibrary/Common/Collections/Event.sc
@@ -193,19 +193,13 @@ Event : Environment {
 				note: #{
 					(~degree + ~mtranspose).degreeToKey(
 						~scale,
-						~scale.respondsTo(\stepsPerOctave).if(
-							{ ~scale.stepsPerOctave },
-							~stepsPerOctave
-						)
+						~scale.tryPerform(\stepsPerOctave) ?? {~stepsPerOctave}
 					);
 				},
 				midinote: #{
-					(((~note.value + ~gtranspose + ~root) /
-						~scale.respondsTo(\stepsPerOctave).if(
-							{ ~scale.stepsPerOctave },
-							~stepsPerOctave) + ~octave - 5.0) *
-						(12.0 * ~scale.respondsTo(\octaveRatio).if
-							({ ~scale.octaveRatio }, ~octaveRatio).log2) + 60.0);
+					((~note.value + ~gtranspose + ~root) /
+						(~scale.tryPerform(\stepsPerOctave) ?? {~stepsPerOctave} + ~octave - 5.0) 
+						* (12.0 * (~scale.tryPerform(\octaveRatio) ?? { ~octaveRatio }).log2) + 60.0);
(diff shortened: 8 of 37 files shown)


dyfer:
> While this is currently a minor issue in the class library, as we move towards AbtractObject, this will become a major pain.

I don't have a bigger picture of the move towards AbstractObject so I take your word for it... On the face of it, I feel like introspecting availability of a method without calling it seems useful. It would be great to get clarification what is problematic about it.

This seems like a rather intrusive change, so I'd like to make sure we understand all the implications.

BTW this is also [used in our test runner](https://github.com/JordanHendersonMusic/supercollider/blob/topic/limit-respondsTo-usage/testsuite/scripts/sclang_test_runner.scd#L88). If this PR proceeds, that should also be addressed. @scztt I believe you originally wrote that code, I wonder if you have any feedback on this PR?

If we decide this is indeed a desired change, I don't think the method signature should change (re: added warning). Rather, the method should be deprecated and post a deprecation warning.

JordanHendersonMusic:
> I feel like introspecting availability of a method without calling it seems useful.

This is fundamentally impossible in SC because we support doesNotUnderstand. 

A silly example...
```
F {
   doesNotUnderstand { |selector|
      if(selector === foo and: 0.5.coin ) {
         ^1
      };
      DoesNotUnderstandError(...).throw
   }
}
```
Does F respond to \foo?

Obviously this is silly, but what if we add a method in response to something external? You'd need to simulate the entire universe to answer the question - or get a time machine.

This severely impacts the ability to create wrapper classes, that extend or intercept messages.

What respondsTo should be called is 'object responds to message and is implemented as a static method'.

dyfer:
> What respondsTo should be called is 'object responds to message and is implemented as a static method'.

Yes...

I see it's tricky to make this work.

Is the idea that `doesNotUnderstand` overloading is going to be used more widely, beyond Event and runtime object prototyping?

I'd argue that fixing `tryPerform` when it still can't differentiate method not found and method returning nil is not ideal. But maybe there's no other way...

JordanHendersonMusic:
> Is the idea that doesNotUnderstand overloading is going to be used more widely, beyond Event and runtime object prototyping?

Yes, this is the basis of abstract object, it will not work unless we fully commit to this idea.

This is a more general issue with 'nil' and what it means.

However, I think it is quite minor. There was maybe one case in the class lib where this was needed and it was just as easy to wrap it directly in a try block.

dyfer:
Given the conversation here and in https://github.com/supercollider/supercollider/issues/7217:

Is [this behavior](https://github.com/supercollider/supercollider/pull/7228/files#diff-7a8d29e7dcf2197ffb96e0301a92f14f767bf1d925b70ff9aff470c949c81d25R131) included only for backward compatibility? if it is, then I think it unnecessarily makes the functionality fragile. My suggestion would be:

1. Introduce a new, fixed `tryPerform` (`tryPerformNew` or something like that?) that catches all errors, and use it throughout the codebase (except cases where it's not possible currently). Also document that respondsTo and tryPerform should not be used in new code, but don't post warnings or deprecate anything just yet. I'd limit this PR to this step.
2. Do whatever is needed to be able to proceed to step 3 (??) - would that mean adopting AbstractObject?
3. Fix the remaining cases to use `tryPerformNew`
4. (Optionally) deprecate `respondsTo` and `tryPerform`. (I personally think it would also be okay to keep them for a while and _not_ deprecate them, just documented as "don't use")

dyfer:
Ah, I think I understand the proposed implementation of tryPerform and the inability to distinguish error from a method returning nil.

How about introducing a new method that returns a boolean, but actually tries to run the method, discarding the result?

```supercollider

	tryPerformBool { |...args, kwargs|
	 	var r;
		try {
			this.performArgs(args[0], args[1..], kwargs);
			r = true;
		} { |er|
			r = false;
		};
		^r
	}
// or
	tryPerformBool { |...args, kwargs|
	 	var r = false;
		try {
			this.performArgs(args[0], args[1..], kwargs);
			r = true;
		};
		^r
	}
```
In any case, I think I'd still suggest to proceed roughly according to the previous comment, so:

1. Use either `tryPerform` or `tryPerformBool` (or however we'd call it) throughout the codebase (except cases where it's not possible currently). Also document that respondsTo should not be used in new code, but don't post warnings or deprecate anything just yet. I'd limit this PR to this step.
2. Do whatever is needed to be able to proceed to step 3 (??) - would that mean adopting AbstractObject?
3. Fix the remaining cases to use `tryPerform` or `tryPerformBool`
4. (Optionally) deprecate `respondsTo`. (I personally think it would also be okay to keep it for a while and _not_ deprecate it, just documented as "don't use")

telephon:
- In terms of reflection, it is absolutely ok semantically to call `respondsTo`. 
- it is a good step to not use respondsTo in the class library to make it more universal
-  but`x.tryPerform(sel) ?? { ... }` is not really the right replacement, because it may successfully call the method and still return nil, which would then call the function instead of returning the correct nil value. "Returning nil" is not the same as failing to implement an interface.
- the problem about using `try` instead would be similar: any error in the call (independently where it came from) would cause the replacement to be returned.

I do see that `doesNotUnderstand` may delegate absolutely anywhere, but it could also be that good proto object design to implement `respondsTo` correctly. A little bit similar to what we do with `==` and `hash`. This would be a reasonable protocol.

For example, if I have an extendible object, it would look up if the extension has the method, otherwise pass on to the object it holds. If I have a proto object, I would, by default, simply look if it implements the method. If I want a more advanced delegative behaviour of doesNotUnderstand, then I am responsible to also implement a reasonable respondsTo. In the extreme case, I could make it return `true` for all selectors and take care of it in the object.

In short, I feel it just moves the bump in the carpet elsewhere. Just thinking aloud – as I said I like making the existing code base more extendible.

telephon:
Also, just to note, `tryPerform` may possibly change (and break) state, which can't be undone.

JordanHendersonMusic:
> In terms of reflection, it is absolutely ok semantically to call respondsTo.

Okay, but it is implemented as `object responds to message that is implemented as a method`.

> For example, if I have an extendible object, it would look up if the extension has the method, otherwise pass on to the object it holds. If I have a proto object, I would, by default, simply look if it implements the method. If I want a more advanced delegative behaviour of doesNotUnderstand, then I am responsible to also implement a reasonable respondsTo. In the extreme case, I could make it return true for all selectors and take care of it in the object. .. .but it could also be that good proto object design to implement respondsTo correct.

Yes this could be done, however, I don't think it could be done for **all** abstract objects.
One example is something like `Maybe`, where the value isn't assigned until later and the result of `respondsTo` cannot be known.... which takes us back to control flow!

> butx.tryPerform(sel) ?? { ... } is not really the right replacement

It was in 90% (60/66) of cases contained in the class library. After flicking through the quarks, it appears to be a even a little higher.

> ...it may successfully call the method and still return nil, which would then call the function instead of returning the correct nil value. "Returning nil" is not the same as failing to implement an interface.

@dyfer has a nice suggestion for that. It seems to only be relevant when you want to discard the value so it could also be called `tryPerformDiscard`. 

However, the inverse is also true, the method may exist and purposefully throw a DoesNotUnderstandError or ShouldNotImplementError. 

In that case, `respondsTo` *is* semantically wrong, because the definition of not supporting an interface isn't not having an entry in the method table, hence why I disagree with the follow... 

> In terms of reflection, it is absolutely ok semantically to call respondsTo.

There is something similar here to inlining control flow... Everything is a message, but  `respondsTo` is a message about the messaging system itself; to respond correctly, we must escape the messaging system ­— I think there is some parallel here to Russel's paradox?

Hence, why I think we should remove its use from the class library and discourage it through documentation and a warning (this PR), and consider longer term how to remove it.

telephon:
Just for reference, I copy here the new implementation of `tryPerform` which should replace `respondsTo`:


```supercollider
tryPerform { |...args, kwargs|
	 	var r;
		try {
			r = this.performArgs(args[0], args[1..], kwargs)
		} { |er|
			if(er.isKindOf(DoesNotUnderstandError) and: { er.selector === args[0]} ) {
				r = nil; // Failure. Old interface returned nil. If the method returns nil to, there is no way to tell the difference.
			} {
				er.throw;
			}
		}
		^r
	}
```

Two immediate issues that would need to be addressed:
- since this actually performs the message rather than asking the object if it may, it may change hidden state that is not possible to change back
- the test `if(er.isKindOf(DoesNotUnderstandError) and: { er.selector === args[0]} )` may in some (possibly rare) cases be wrong, because the error may have arisen for a different reason.
- when implementing `==` via tryPerform, `hash` would have to be implemented in exactly the same way, otherwise the system would (really logically) become inconsistent.

I'll answer your issues above a little later.

telephon:
> One example is something like `Maybe`, where the value isn't assigned until later and the result of `respondsTo` cannot be known.... which takes us back to control flow!

Yes, but the result of `tryPerform` can also not be known until later. This is what I meant with moving the bump under the carpet.



> However, the inverse is also true, the method may exist and purposefully throw a DoesNotUnderstandError or ShouldNotImplementError.
> 
> In that case, `respondsTo` _is_ semantically wrong, because the definition of not supporting an interface isn't not having an entry in the method table, hence why I disagree with the follow...
> 
> > In terms of reflection, it is absolutely ok semantically to call respondsTo.

Yes, this may mean that `respondsTo` is not currently correctly implemented.



> There is something similar here to inlining control flow... Everything is a message, but `respondsTo` is a message about the messaging system itself; to respond correctly, we must escape the messaging system ­— I think there is some parallel here to Russel's paradox?

A system like sclang is inconsistent from a logical point of view and it is easy to construct paradoxes in it. It (arguably) need to be inconsistent to be expressive enough. This also means that "meta-questions" and "questions" are on the same level: classes are just objects. I agree, you are correctly pointing out that this concerns control flow, which is here not entirely up to the receiver. So we perhaps need to find a better way to let the call site direct control flow depending on information about the objects concerned. Or better, let the object specify control flow in the call site.

In any way, the current use of `try` is not yet the solution (for the reasons in the previous post).

Great to have this coversation, btw.

JordanHendersonMusic:
> A system like sclang is inconsistent from a logical point of view and it is easy to construct paradoxes in it.

> Yes, this may mean that respondsTo is not currently correctly implemented.

My point is that it is *impossible* to implement. You need to reflect on what the code will do, which requires simulating the program, the external interfaces it is connected to, and everything else in the universe.

To get rid of this consistency, I think we should move people away from using this method. Since ~90% of use cases are can simply be replaced with `tryPerform`, this seems completely reasonable to me?

Re. `tryPerform`'s implementation:

> since this actually performs the message rather than asking the object if it may, it may change hidden state that is not possible to change back

As >90%  of use cases actually did want to evaluate the method, and the rest could be replaced (perhaps not due to backwards compatibility) with a simple dependency inversion. In fact, I'm pretty certain that **all** other possible use cases can be implemented by extending the underlying interface.

So this is actually what this method should do, it is what 9 out of 10 times people want to do.

> the test if(er.isKindOf(DoesNotUnderstandError) and: { er.selector === args[0]} ) may in some (possibly rare) cases be wrong, because the error may have arisen for a different reason.

Thinking about this a bit more.... it should also compare the receiver via identity.
Then, some `AbstractObject` should implement `tryPerform` by comparing it (with equality) against what ever it is trying to perform on. This does require abstract objects to implement yet another method... however, this one (unlike `respondsTo`) is actually possible to write.

> when implementing == via tryPerform, hash would have to be implemented in exactly the same way, otherwise the system would (really logically) become inconsistent.

Isn't that the case already though, that `==` and `hash` must agree?

telephon:
> My point is that it is _impossible_ to implement. You need to reflect on what the code will do, which requires simulating the program, the external interfaces it is connected to, and everything else in the universe.

Let's see it this way: `respondsTo` is a question about some property of the receiver (be it a delegator or not). We could go through the class library and see what property this is. For example, one of those could be, for a UGen, `implementsAudioRate`. This would have to be made true explicitly where it is true. Of course this is more overhead, but it is explicit. Any abstract object could / should implement it if it wants to take part in audio rate stuff. This example makes clear that currently, a class *signals* that property through implementing the respective method. 

> To get rid of this consistency, I think we should move people away from using this method. Since ~90% of use cases are can simply be replaced with `tryPerform`, this seems completely reasonable to me?

If `tryPerform` would be implementable correctly, yes, but I have my doubts. We haven't yet talked about the overhead it has, but let's keep this aside for now.




> Thinking about this a bit more.... it should also compare the receiver via identity. Then, some `AbstractObject` should implement `tryPerform` by comparing it (with equality) against what ever it is trying to perform on. This does require abstract objects to implement yet another method... however, this one (unlike `respondsTo`) is actually possible to write.

When the message is passed on via doesNotUnderstand and further down an object does really not understand the message, the receiver will not be the abstract object, but that object it was called on.

> > when implementing == via tryPerform, hash would have to be implemented in exactly the same way, otherwise the system would (really logically) become inconsistent.
> 
> Isn't that the case already though, that `==` and `hash` must agree?

In the change you made, it isn't: you forward the decision about `==`, but not the retrieval of `hash`.

telephon:
> Thinking about this a bit more.... it should also compare the receiver via identity.
> Then, some `AbstractObject` should implement `tryPerform` by comparing it (with equality) against what ever it is trying to perform on. This does require abstract objects to implement yet another method... however, this one (unlike `respondsTo`) is actually possible to write.

This might be a path -- although I am really not sure it is simple enough. 

It would not allow recursive calls, however.

```supercollider
PassOn : AbstractObject {
	var <>func;
    doesNotUnderstand { |selector ... args, kwargs|
		func.valueArgs(args, kwargs)
	}
}

Caller {
	maybeCallMe { |obj|
		^obj.tryPerform(\foo) ?? { obj }
	}	
}

f = Array.new;
a = PassOn.func_({ | ... args| f.foo(args) })
x = Caller.new.maybeCallMe(a);
```

Then x will not fail (f does not understand foo).

If we make sure in tryPerform to throw if the receiver is not the initial one, let's suppose:

```supercollider
f = Array.new;
a = PassOn.func_({ | ... args| if(condition) { a.foo(args) } }); // conditional recursion
x = Caller.new.maybeCallMe(a);
```

There may be other cases that I have overlooked.

PR #7233, opened 2025-11-11
TestServer_clientID_booted: increase timeouts

dyfer:
## Purpose and Motivation

I noticed lately that some of the TestServer_clientID_booted tests failed...

This PR increases timeouts so that we wait longer for the server to finish booting and quitting.

In particular, [here](https://github.com/supercollider/supercollider/actions/runs/19241584129/job/55025336955#step:7:13376) it seems like the "watchdog" timer fired before the `doWhenBooted` - this message should've appeared before the `PASS` messages (here's the [test source](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestServer_clientID_booted.sc#L131-L165)).
I'm not 100% sure this was the case, but I generally find the AppClock (which we run the testsuite with) in the CI to be extremely unstable so all timing requirements need to be very loose (TempoClock seems much better, but we can't run the whole testsuite with it). The longer timeout doesn't necessarily increase the testing time - the test still proceeds as soon as the server confirms its readiness. But it seems that sometimes in the CI this takes longer than the previously suggested 3 seconds.
I extended other wait times just in case. Again, if everything works as expected, we shouldn't ever reach them.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestServer_clientID_booted.sc
+++ b/testsuite/classlibrary/TestServer_clientID_booted.sc
@@ -99,7 +99,7 @@ TestServer_clientID_booted : UnitTest {
 
 		cond = Condition.new;
 
-		timer = fork { 3.wait; cond.unhang };
+		timer = fork { 10.wait; cond.unhang };
 		server.doWhenBooted {
 			"% - server login timed out.\n".postf(thisMethod);
 			timer.stop;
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Looks like a good idea. Feel free to merge after successful CI @dyfer

PR #7234, opened 2025-11-11
test_oscillators_startAtCorrectPhase: add target

dyfer:
## Purpose and Motivation

`{}.loadToFloatArray` was missing a target and it was using a default server, instead of the one designated for this test. 
I looked over the testsuite and it seems to have been the only place where the target argument for `{}.loadToFloatArray` was missing.

I believe this made the testsuite end up with an extra default server object dangling at the end... (not sure).

## Types of changes

- Documentation
- Bug fix (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation


capital-G (approved):
Well spotted! Feel free to merge when pipeline passes.

PR #7244, opened 2025-11-17
build: remove Intel compiler flags

dyfer:
## Purpose and Motivation

Does anybody know if we still need these settings for the Intel compiler? They were added 15 years ago.
I don't think we've ever officially supported it and at least the `-std=c++11` part is outdated.

## Types of changes

- Cleanup

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation


dyfer:
@timblechmann I think you were the last person to touch support for the intel compiler. Do you have any objections to removing it? Is there a reason to keep it around?

JordanHendersonMusic (approved):
Supporting gcc, clang and msvc plus all the os and c++ library combinations is already a lot of work. I think this is okay,

PR #7246, opened 2025-11-18
Recorder-related documentation changes (Recorder, Server, ServerOptions)

HotwheelsSisyphus:
<details><summary>Classlib change moved to 7249</summary>
<p>Recorder.prRecord created the recording synth on target `0` when no node was explicitly specified.

This will always translate to Group 0 on the *default* server, which creates problems when trying to record on any other server. 
I changed `0` to ~~`server.asTarget`~~ `RootNode(server)`, where `server` is an instance variable (each Recorder is assigned to a server). ~~`asTarget` then translates to the default Group on that server.~~ (.asTarget gets called in Synth.new anyway).

<details><summary>More or less settled after help from @cdbzb: `0` is `RootNode(server.default)`. </summary>
<p>
 Maybe consider before merging: `0.asTarget` returns `Group(0)` on the default server, whereas after my change (to `server.asTarget`), the corresponding expression returns `Group(1)`. 
I did not find a way to make it return Group(0). Is there any good reason one would _want_ Group(0) here? From what I understand, the default group is meant to be Group(1). But what is Group(0) supposed to be?
</p>
</details>
</p>
</details>
Will add more doc changes in a bit!

improved (?) indentation in the very first example
Removed an unused (an unexciting) synthdef; 
changed "SynthDefName" to \synthdefname
added note to the node argument of Recorder.record.

## Purpose and Motivation
I'm discovering the joys of using multiple servers/cpu cores. 

## Types of changes

- Documentation
- Bug fix

### DIFF
--- a/HelpSource/Classes/Recorder.schelp
+++ b/HelpSource/Classes/Recorder.schelp
@@ -10,7 +10,8 @@ code::
 (
 { SinOsc.ar(
 	SinOsc.ar(
-		XLine.kr(1, 100, 5)).exprange(*XLine.kr([20, 800], [7000, 200], 10)
+		XLine.kr(1, 100, 5)
+		).exprange(*XLine.kr([20, 800], [7000, 200], 10)
     )
    ) * 0.1
 
@@ -80,7 +81,10 @@ argument:: numChannels
 The number of output channels to record.
 
 argument:: node
-The link::Classes/Node:: to record immediately after. By default, this is the default group 1.
+The link::Classes/Node:: to record immediately after.
+By default, this is the default group 1.
+While the node ID can be represented by an link::Classes/Integer::, 
+this should avoided when using more than one server simultaneously (see link::Reference/asTarget::). 
 
 argument:: duration
 If set, this limits recording to a given time in seconds.
@@ -157,20 +161,15 @@ section::Examples
 code::
 // something to record
 (
-SynthDef("bubbles", { |out|
+SynthDef(\bubbles, { |out|
 	var f, sound;
 	f = LFSaw.kr(0.4, 0, 24, LFSaw.kr([8, 7.23], 0, 3, 80)).midicps; // glissando function
 	sound = CombN.ar(SinOsc.ar(f, 0, 0.04), 0.2, 0.2, 4); // echoing sine wave
 	Out.ar(out, sound);
 }).add;
-
-SynthDef("tpulse", { |out = 0, freq = 700, sawFreq = 440.0|
-	Out.ar(out, SyncSaw.ar(freq, sawFreq, 0.1))
-}).add;
-
 )
 
-x = Synth.new("bubbles");
+x = Synth(\bubbles);
 
 s.prepareForRecord; // if you want to start recording on a precise moment in time, you have to call this first.
 
@@ -192,7 +191,7 @@ code::
 thisProcess.platform.recordingsDir = "/home/user/";
 
 // instantiate the Recorder
-r = Recorder.new(s);
+r = Recorder(s);
 
 // record into a flac file
 r.recHeaderFormat = "flac";
--- a/SCClassLibrary/Common/Control/Recorder.sc
+++ b/SCClassLibrary/Common/Control/Recorder.sc
@@ -138,7 +138,7 @@ Recorder {
 	/* private implementation */
 
 	prRecord { |bus, node, dur|
-		recordNode = Synth.tail(node ? RootNode(server), synthDef.name, [\bufnum, recordBuf, \in, bus, \duration, dur ? -1]);
+		recordNode = Synth.tail(node ? 0, synthDef.name, [\bufnum, recordBuf, \in, bus, \duration, dur ? -1]);
 		recordNode.register(true);
 		recordNode.onFree { this.stopRecording };
 		if(responder.isNil) {
--- a/SCClassLibrary/Common/Control/Recorder.sc
+++ b/SCClassLibrary/Common/Control/Recorder.sc
@@ -138,7 +138,7 @@ Recorder {
 	/* private implementation */
 
 	prRecord { |bus, node, dur|
-		recordNode = Synth.tail(node ? server.asTarget, synthDef.name, [\bufnum, recordBuf, \in, bus, \duration, dur ? -1]);
+		recordNode = Synth.tail(node ? RootNode(server), synthDef.name, [\bufnum, recordBuf, \in, bus, \duration, dur ? -1]);
 		recordNode.register(true);
 		recordNode.onFree { this.stopRecording };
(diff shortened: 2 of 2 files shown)


cdbzb:
Group 0 is the RootNode - from the help: A RootNode is the Group with the nodeID of 0 which is always present on
each Server and represents the root of that server's node tree.

It is always playing, and always running, cannot be freed, or moved
anywhere.

HotwheelsSisyphus:
Ah, thanks! That would make sense, and it may be what the 0 in here was intended for. The doc for RootNode also mentions that not adding nodes to RootNode  "provides a known basic node order and protects functionality like Server.record, Server.scope, etc."
~~But otoh, I just checked and `Rootnode(s).asTarget` is not equal to `0.asTarget`. ~~
EDIT: nah, I guess that relation is just not implemented. They do have the same nodeID on the same server, which should be enough.

telephon:
It would be better if the two kinds of changes you made were in separate pull requests. Changing this zero to RootNode is unproblematic, but should anyone have an issue at some later time and search for it, this would make it easier to find. But not a big issue.

HotwheelsSisyphus:
Oh, no problem, I can do that. By the "kinds of changes", I assume you meant the classlib and the help changes?

HelpSource/Classes/Recorder.schelp
dyfer, line 18: If you're fixing indentation/formatting, then I think this should be
```supercollider
{ 
	SinOsc.ar(
		SinOsc.ar(
			XLine.kr(1, 100, 5)
		).exprange(*XLine.kr([20, 800], [7000, 200], 10))
	) * 0.1
}.play;
```
...or leave it as is.

HelpSource/Classes/Recorder.schelp
```diff
@@ -80,7 +81,10 @@ argument:: numChannels
 The number of output channels to record.
 
 argument:: node
-The link::Classes/Node:: to record immediately after. By default, this is the default group 1.
+The link::Classes/Node:: to record immediately after.
+By default, this is the default group 1.
+While the node ID can be represented by an link::Classes/Integer::, 
+this should avoided when using more than one server simultaneously (see link::Reference/asTarget::).
```
dyfer, line 87: "should avoided" -> "should be avoided"

dyfer (changes requested):
Thanks! 
Aside from the two minor suggestions, I have a general comments/questions for the helpfile changes:
- re: `By default, this is the default group 1.` in the suggested change - is that true? Isn't this actually group 0 / root node?
- what would we think about removing all integer designations of the target in the Recorder helpfile and instead replacing them with `Group` / `RootNode` and similar objects? I don't think using integers directly should be encouraged anymore. I understand it may be better to do this in a separate PR.

HelpSource/Classes/Recorder.schelp
HotwheelsSisyphus: Right. Now I ended up messing with it even more and I figure the most elegant way to deal with this is variables:
```
(
{
    var fmFreq = XLine.kr(1,100,5);
	var fmDepth = XLine.kr([20, 800], [7000, 200], 10);         
	SinOsc.ar(
		SinOsc.ar(fmFreq)
			.exprange(*fmDepth)
    ) * 0.1
}.play;
s.record(duration: 10);
)
```
does that check out? At least it hurts my brain significantly less.

telephon:
> By default, this is the default group 1.

This means, the recorder comes _after_ the group 1, and in fact the default is that it will be added to the tail of the node given in the argument, which by default is 0.

So this may be a better explanation than the one that was and still is in the help page.

dyfer:
> So this may be a better explanation than the one that was and still is in the help page.

Yes, IIUC the functionality is that it will be placed _at the end_ of the root node, not explicitly after group 1 (though the result is the same unless one starts putting synths outside of the default group); the default node value is 0.

HotwheelsSisyphus:
> * re: `By default, this is the default group 1.` in the suggested change - is that true? Isn't this actually group 0 / root node?

> This means, the recorder comes after the group 1, and in fact the default is that it will be added to the tail of the node given in the argument, which by default is 0.

Yes, I overlooked this yesterday... I was going to change it to the following: (haven't yet pushed it):
```
The link::Classes/Node:: at the tail of which a temporary recording node will be added. 
When left unspecified, this will default to the servers's link::Classes/RootNode::.
```
--------------
> * what would we think about removing all integer designations of the target in the Recorder helpfile and instead replacing them with `Group` / `RootNode` and similar objects? I don't think using integers directly should be encouraged anymore. I understand it may be better to do this in a separate PR.

Yes, I will scan it again in that respect, but I think there are not in fact taht many.

Re separate PR: I was going to separate the classlib and help components here anyway (I'll revert the classlib change in here and submit it separately). So might as well do it here

HotwheelsSisyphus:
>  (though the result is the same unless one starts putting synths outside of the default group)

I think, for instance, the SafetyNet limiter happens after Group 1 but not after the RootNode.

HelpSource/Classes/Recorder.schelp
dyfer: Variables are fine, but I'd prefer to stick to the default indentation that's applied by the IDE. I see no point doing this by hand, just copy-paste into the IDE and press TAB:
```supercollider
{
	var fmFreq = XLine.kr(1,100,5);
	var fmDepth = XLine.kr([20, 800], [7000, 200], 10);         
	SinOsc.ar(
		SinOsc.ar(fmFreq)
		.exprange(*fmDepth)
	) * 0.1
}.play;
```

EDIT: Alternatively keep `SinOsc.ar(fmFreq).exprange(*fmDepth)` in a single line

HotwheelsSisyphus: oh absolutely...
I use scnvim normally and it happens that I forget to check that stuff! Also I simply misremembered the guidelines here.

HotwheelsSisyphus: ah but now I double checked, and found this in the style guidelines, so I may not have misremembered after all. 
So this is maybe a discrepancy/ambiguity between style guidelines and IDE indentation:

> In long chains of method calls on the same object, it may be beneficial to split the method call across two lines. The dot should be on the second line, not the first, and it should be indented one level. Don't put whitespace between the dot and the method name.
> ```supercollider
> Button()
>    .states_([["blorp", nil, nil]])
>    .action_({
>        "hey hey hey".postln
>    });
> ```

For this one I'll just put the method call on the same line so we can move on.

dyfer: Thanks for pointing this out. Yes, it seems like a discrepancy, would you be able to file that as a new issue? If not, I'd be happy to do it. Thanks.

HotwheelsSisyphus:
sure, will do so later today!

PR #7248, opened 2025-11-21
cmake: set/update compiler hardware optimizations

dyfer:
## Purpose and Motivation

Previously we didn't explicitly use compiler optimizations for newer CPU instructions and features (like SSE4).

I reorganized and updated compiler settings optimizing for modern hardware:
- GCC/Clang - Linux (and MinGW):
  - for GCC 12+ and Clang 12+ on x86_64 I used "architecture level" matching Haswell processors (2012-era)
    - for older compilers I default to SSE4 which I believe maps to 2009-era CPUs; I don't think it's needed to target anything older than that
  - for Arm64 I used `armv8-a` but I'm not sure how safe/conservative that is
  - optimizations for older 32-bit ARM (v6 and v7 - Rpi 1 - 2 - 3) were left in place, though using a newer CMake conventions; they are untested, unfortunately
 - Apple Clang:
   - I used "CPU family" approach for arm64 and "architecture level" for x86_64; for the latter the default is now set to Haswell processors (Apple hardware from 2013 onward); note that in the legacy build I set this to an earlier value (Ivy Bridge) to catch one generation of hardware that the legacy build can still run on (2012 MacBook Pro running macOS 10.15)
   - for arm64 (Apple CPU) the default value `apple-m1` doesn't change anything; however exposing it allows taking potential advantage of newer Apple chips
-  Windows MSVC:
   - I added the newer `SSE4.2` optimization, which is a relatively recent addition to MSVC (previously the first "higher" option that the default SSE2 was AVX); I didn't use AVX optimizations since some [lower-end chips from as late as 2021](https://en.wikipedia.org/wiki/Tremont_(microarchitecture)) still don't support it. If we decided to create a Windows legacy build, maintaining the basic SSE4.2 optimization, we could consider optimizing the main build for newer CPUs with e.g. AVX512.

FYI I haven't tested the actual performance of these changes.

IIUC these settings do _not_ affect explicit uses of SSE/NEON instructions; e.g. in the `nova-simd` library.

## Types of changes

- Documentation
- New feature (?)

## To-do list

- [x] Code is tested - partially:
  - [ ] Tested for performance changes 
  - [ ] Tested for building on Raspberry Pis
  - [x] The existing CI configurations build and run
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -43,7 +43,7 @@ jobs:
             # homebrew-uninstall: "readline" # use this instead when cross-compiling for x86_64 on arm64
             # vcpkg-packages: "readline portaudio libsndfile fftw3" # use this instead when cross-compiling for x86_64 on arm64
             vcpkg-triplet: x64-osx-release-supercollider # required for build-libsndfile
-            extra-cmake-flags: ""
+            extra-cmake-flags: "-D CLANG_X86_64_ARCH=ivybridge"
             artifact-suffix: "macOS-x64-legacy" # set if needed - will trigger artifact upload
             ctest: false
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,14 +106,6 @@ option(USE_CCACHE "Use ccache if available." ON)
 
 option(FFT_GREEN "Use internal 'Green' FFT lib rather than FFTW. (Not recommended.)" OFF)
 
-if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^AMD64|EM64T|x86_64|X86|i386|i686")
-    option(SSE "Compile with support for SSE instructions." ON)
-    option(SSE2 "Compile with support for SSE2 instructions." ON)
-else() # non-x86 platforms do not have SSE
-    set(SSE OFF)
-    set(SSE2 OFF)
-endif()
-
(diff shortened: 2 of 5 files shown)


JordanHendersonMusic:
I don't know about this approach, seems quite complicated and stateful. Could we use cmake presets instead?

dyfer:
I haven't used cmake presents. Could you elaborate how that would be helpful here?

dyfer:
I think I have an idea how to migrate this to presets. I'll probably submit it as another PR. Turning this to draft for now.

JordanHendersonMusic:
I don't think you can auto detect with presets. So if that is needed maybe this is fine?

dyfer:
> I don't think you can auto detect with presets.

Yeah, that's correct. I thought that if no preset is specified, then we'd default to close to what we have right now (SSE2 or SSE4 on x86_64). Presets would give a nicer way to have multiple options to choose from. 

But maybe that's not as important.

Alternatively, without the presets, I think I'd like to expose a single option in cmake, like `SC_COMPILER_ARCH_FLAGS` or similar and set it conditionally using the logic from this PR. The difference from the current state of the PR would be that the flag would be a complete compiler flag (like `/arch:SSE4.2` or `-Xarch_x86_64 -march=haswell -Xarch_arm64 -mcpu=apple-m1`), so it would be possible to cleanly override it for any configuration/compiler/platform.

PR #7249, opened 2025-11-23
Recorder.prRecord: default node: replace integer 0 with RootNode(server)

HotwheelsSisyphus:
as discussed in #7246; separating the classlib change from the help changes, which are getting more involved.

integer 0 will default to the RootNode on the _default_ server, even if the recorder is supposedly on another server. 
This changes this to default to the RootNode of the recorder's server, even when that is not the default server.

This is a non-breaking change.

### DIFF
--- a/SCClassLibrary/Common/Control/Recorder.sc
+++ b/SCClassLibrary/Common/Control/Recorder.sc
@@ -138,7 +138,7 @@ Recorder {
 	/* private implementation */
 
 	prRecord { |bus, node, dur|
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks @HotwheelsSisyphus !
I think this is a straightforward change. Would someone from @supercollider/class-library be able to review this?

jamshark70 (approved):
LGTM

telephon:
Thanks!

PR #7253, opened 2025-11-23
3.14.1

capital-G:
Merging 3.14.1 into main as part of the release.


dyfer (approved):

capital-G:
We don't understand why yields the error message

grafik

so we created #7256 instead.

PR #7259, opened 2025-11-24
Merge 3.14.1 main back into develop

capital-G:
Last part of the release process of 3.14.1 - using a custom branch to avoid branch protection problems.

This is a normal PR - I think this is fine, but we cherry picked from develop - so I don't know what happens with the cherry picked commits in between ;)


dyfer (approved):
Thanks so much!

The changes look good.

_I don't think_ the cherry-picked commits will be an issue... but we'll see :)

Merging!

PR #7262, opened 2025-11-26
build: reorganize compile options

dyfer:
## Purpose and Motivation

I'm working on a few aspects of CMake cleanup and I'm splitting work between multiple PRs for easier review.

This PR reorganizes compile options so that they are defined in the same section of the main CMakeLists.txt.

This PR is meant to be a mostly cosmetic and it should not change anything for supported configurations. However, it does remove options that I _think_ are outdated (flags for older MinGW and macOS/GCC).

## Types of changes

- Cleanup

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -413,7 +413,7 @@ if (SC_MEMORY_DEBUGGING)
 endif()
 
 if(MINGW)
-	add_compile_options("$<$<COMPILE_LANGUAGE:C,CXX>:-mstackrealign>")
+	add_compile_options(-mstackrealign)
 endif()
 
 # support for building on Raspberry Pi 1/2/3 and BBB
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,8 +341,6 @@ endif()
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 	add_compile_options(
 		-Werror=return-type      # Error out if function definitions lack a return type
-		-fschedule-insns2        # Optimize instruction scheduling
-		-fomit-frame-pointer     # Optimize by omitting the frame pointer where possible
 		-fno-math-errno          # Optimize math functions by not setting errno
 		-fno-signaling-nans      # Assume no signaling NaNs for better optimization
 		-fsigned-zeros           # Maintain distinction between -0.0 and +0.0
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,32 +76,8 @@ if (${CMAKE_COMPILER_IS_GNUCXX})
             OUTPUT_STRIP_TRAILING_WHITESPACE
     )
 
-    add_compile_options(
-        -Werror=return-type      # Error out if function definitions lack a return type
-        -fschedule-insns2        # Optimize instruction scheduling
-        -fomit-frame-pointer     # Optimize by omitting the frame pointer where possible
-        -fno-math-errno          # Optimize math functions by not setting errno
-        -fno-signaling-nans      # Assume no signaling NaNs for better optimization
-        -fsigned-zeros           # Maintain distinction between -0.0 and +0.0
-        -fno-associative-math    # Prevent reordering of floating-point operations
-    )
-
 	if(${_gcc_version} VERSION_LESS 4.8)
 		message(FATAL_ERROR "SuperCollider requires at least gcc-4.8 when compiled with gcc.")
-	endif()
-
-	if(APPLE)
-        execute_process(
-                COMMAND ${CMAKE_CXX_COMPILER} --version
-                OUTPUT_VARIABLE _gcc_version_info
-                OUTPUT_STRIP_TRAILING_WHITESPACE
-        )
-		if ("${_gcc_version_info}" MATCHES "Apple")
-			add_definitions("-fpascal-strings")
-		endif()
-		add_definitions("-D_REENTRANT")
-	elseif(NOT (WIN32 AND NOT CYGWIN))
-		add_definitions("-pthread")
 	endif()
 
 elseif(${CMAKE_CXX_COMPILER} MATCHES icpc)
@@ -119,10 +95,6 @@ if(APPLE)
 	endif()
 elseif(UNIX)
 	add_definitions(-DSC_DATA_DIR="${CMAKE_INSTALL_PREFIX}/share/SuperCollider")
-endif()
-
-if(WIN32)
-    add_definitions(-DWIN32_LEAN_AND_MEAN -DNOMINMAX -D_WIN32_WINNT=0x0600)
 endif()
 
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -361,15 +333,40 @@ if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
...
-        # the default
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
-        set(CMAKE_CXX_LINK_FLAGS "${CMAKE_CXX_LINK_FLAGS} -stdlib=libc++")
-    endif()
+#############################################
+# other compiler settings
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+	add_compile_options(
+		-Werror=return-type      # Error out if function definitions lack a return type
```
JordanHendersonMusic, line 343: I think we want the return type error on clang too.

CMakeLists.txt
```diff
@@ -361,15 +333,40 @@ if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
...
+		-fschedule-insns2        # Optimize instruction scheduling
```
JordanHendersonMusic: GCC seems to enable this at 02 so we don't need it.
```
-fschedule-insns ¶
If supported for the target machine, attempt to reorder instructions to eliminate execution stalls due to required data being unavailable. This helps machines that have slow floating point or memory load instructions by allowing other instructions to be issued until the result of the load or floating-point instruction is required.

Conventionally enabled at optimization levels -O2 and -O3. However, many targets override this behavior. For example, on x86, it is disabled at all levels, while on AArch64, it is enabled only at -O3.

-fschedule-insns2 ¶
Similar to -fschedule-insns, but requests an additional pass of instruction scheduling after register allocation has been done. This is especially useful on machines with a relatively small number of registers and where memory load instructions take more than one cycle.

Enabled at levels -O2, -O3, -Os.
```

CMakeLists.txt
```diff
@@ -361,15 +333,40 @@ if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
...
+		-fomit-frame-pointer     # Optimize by omitting the frame pointer where possible
```
JordanHendersonMusic: Gcc does this at 01, and it only saves a single register, which isn't a big deal on 64 but systems. Probably don't need this.
```
fomit-frame-pointer ¶
Omit the frame pointer in functions that don’t need one. This avoids the instructions to save, set up and restore the frame pointer; on many targets it also makes an extra register available.

On some targets this flag has no effect because the standard calling sequence always uses a frame pointer, so it cannot be omitted.

Note that -fno-omit-frame-pointer doesn’t guarantee the frame pointer is used in all functions. Several targets always omit the frame pointer in leaf functions.

Enabled by default at -O1 and higher.
```

CMakeLists.txt
```diff
@@ -361,15 +333,40 @@ if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
...
+		-fno-math-errno          # Optimize math functions by not setting errno
+		-fno-signaling-nans      # Assume no signaling NaNs for better optimization
+		-fsigned-zeros           # Maintain distinction between -0.0 and +0.0
+		-fno-associative-math    # Prevent reordering of floating-point operations
```
JordanHendersonMusic, line 347: Why wouldn't we want this on clang?

CMakeLists.txt
dyfer: I tried that, but it actually errors out with nova-simd on arm64 ([here](https://github.com/timblechmann/nova-simd/blob/master/vec/vec_neon.hpp#L229-L242) I believe). It's on my radar. 
I'd like to move the options to clang (i.e. introduce functional changes) in a separate PR, if that's okay?

CMakeLists.txt
dyfer: Again, this PR is meant as cleanup only. I'd like to leave adding options to clang side for a separate PR.

CMakeLists.txt
dyfer: This is beyond my expertise so I'm happy to defer to your judgement. 
I just wanted to point out that supernova also turns this on explicitly (which is redundant and I suggested removing it in a separate PR) - are you sure there's no performance aspect to removing it? The fact that it's enabled on -O1 suggests that indeed there's no impact, but I wanted to double check.

CMakeLists.txt
dyfer: Sounds good, I'll remove it.

CMakeLists.txt
JordanHendersonMusic: Sure!

CMakeLists.txt
JordanHendersonMusic: Well there would be a performance impact if it was set to -fno-omit-frame-pointer, but since it is on at -O1, this means it only affect the performance of debug builds... but you want the frame pointer then so you get proper backtraces.

Clang also does the same thing
```
-fomit-frame-pointer, -fno-omit-frame-pointer
Omit the frame pointer from functions that don’t need it. Some stack unwinding cases, such as profilers and sanitizers, may prefer specifying -fno-omit-frame-pointer. On many targets, -O1 and higher omit the frame pointer by default. -m[no-]omit-leaf-frame-pointer takes precedence for leaf functions
```
https://clang.llvm.org/docs/ClangCommandLineReference.html

CMakeLists.txt
JordanHendersonMusic: Okay, but lets circle back to this later then because this could mean gcc and clang builds give different results for basic arithmetic.

dyfer: Yes, I'm actually submitting a fix to nova-simd right now.

If that gets approved, I could add moving flags to common gcc/clang section and include the submodule update in this PR (as a separate commit). Or submit another one later on, let me know.

CMakeLists.txt
JordanHendersonMusic: FYI, I'd use something like `__builtin_unreachable` to fix that in nova-simd.

Or 
```
if defined(_MSC_VER) && !defined(__clang__) // MSVC
    __assume(false);
#else // GCC, Clang
    __builtin_unreachable();
#endif
```

JordanHendersonMusic: Moved conversation over to the PR.

CMakeLists.txt
JordanHendersonMusic: We'd also want to consider what MSVC does. I'm not that familiar with it but everything ought to agree on simple arithmetic.

I'd say keep it for another pr, its quite a complicated change and would be good to get a nice commit message for it. 

Once the tests pass this can be merged.

dyfer:
Sorry for another push - I simplified a compile option for mingw.

JordanHendersonMusic (approved):
Good to merge

PR #4119, opened 2018-10-09
cmake: turn off -ffast-math for gcc

brianlheim:
Purpose and Motivation
----------------------

This causes fp math to behave incorrectly
Instead, enable optimizations for the worst offenders (no signaling nan, no math errno)

Fixes #4116

Types of changes
----------------

- Bug fix (non-breaking change which fixes an issue)
- Breaking change (possibly, needs evaluation)

Checklist
---------

- [x] All tests are passing
- [x] If necessary, new tests were created to address changes in PR, and tests are passing
- [x] Updated documentation, if necessary
- [x] This PR is ready for review

Remaining Work
--------------

Tests should be run after building with this change to ensure nothing
else in the main language test suite has broken as a result of this
change.

PR #7263, opened 2025-11-27
build: add some gcc flags to clang

dyfer:
## Purpose and Motivation

This adds flags that I _think_ are valid for both compilers to the clang section.

I don't know how to enable something like `-Werror=return-type` on MSVC... or if any of the other flags have equivalents.

## Types of changes

- New feature (?)

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -342,11 +342,8 @@ SET(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 	add_compile_options(
-		-Werror=return-type      # Error out if function definitions lack a return type
 		-fno-math-errno          # Optimize math functions by not setting errno
 		-fno-signaling-nans      # Assume no signaling NaNs for better optimization
-		-fsigned-zeros           # Maintain distinction between -0.0 and +0.0
-		-fno-associative-math    # Prevent reordering of floating-point operations
 	)
 
 	if(APPLE)
@@ -357,6 +354,11 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+	if(NOT APPLE)
+		add_compile_options(
+			-fno-math-errno      # On Apple and most BSDs, math library functions never set errno, and so -fno-math-errno is the default
+		)
+	endif()
 	if(NOT SC_CLANG_USES_LIBSTDCPP)
 		# the default
 		add_compile_options("-stdlib=libc++")
@@ -365,6 +367,12 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+	add_compile_options(
+		-Werror=return-type      # Error out if function definitions lack a return type
+		-fsigned-zeros           # Maintain distinction between -0.0 and +0.0
+		-fno-associative-math    # Prevent reordering of floating-point operations
+	)
+
 	if(FORTIFY)
 		add_compile_definitions("_FORTIFY_SOURCE=2")
 	endif()
@@ -391,6 +399,8 @@ if(MSVC)
 # caused by some versions of libsndfile's C++ header redefining nullptr as NULL.
 # See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
   add_compile_options("/Zc:__cplusplus")
+
+# for math operations, we expect the default /fp:precise flag to be used, so that we match settings for gcc/clang
 
 # _ENABLE_ATOMIC_ALIGNMENT_FIX prevents the build from breaking when VS2015 update 2 upwards are used
 # see http://boost.2283326.n4.nabble.com/lockfree-ENABLE-ATOMIC-ALIGNMENT-FIX-for-VS-2015-Update-2-td4685955.html
--- a/server/plugins/CMakeLists.txt
+++ b/server/plugins/CMakeLists.txt
@@ -57,10 +57,6 @@ if(APPLE OR WIN32)
 	set(CMAKE_SHARED_MODULE_SUFFIX ".scx")
 endif()
 
-if (${CMAKE_COMPILER_IS_GNUCXX})
-	add_definitions(-fno-finite-math-only)
-endif()
-
 foreach(plugin ${plugin_sources})
 	string(REPLACE .cpp "" plugin_name ${plugin} )
 	add_library(${plugin_name} MODULE ${plugin})


CMakeLists.txt
```diff
@@ -340,11 +340,8 @@ endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 	add_compile_options(
-		-Werror=return-type      # Error out if function definitions lack a return type
 		-fno-math-errno          # Optimize math functions by not setting errno
 		-fno-signaling-nans      # Assume no signaling NaNs for better optimization
```
JordanHendersonMusic, line 346: I think we want both of these on clang too.

CMakeLists.txt
```diff
@@ -363,6 +360,12 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+	add_compile_options(
+		-Werror=return-type      # Error out if function definitions lack a return type
+		-fsigned-zeros           # Maintain distinction between -0.0 and +0.0
+		-fno-associative-math    # Prevent reordering of floating-point operations
```
JordanHendersonMusic, line 373: We should also specify mscv too.
I ***believe*** the correct flag would be /fp:precise 

https://learn.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior?view=msvc-170

CMakeLists.txt
JordanHendersonMusic: But no-signaling-nans is called 

no-signaling-math[¶](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fsignaling-math)

CMakeLists.txt
dyfer: /fp:precise seems to be the default, should we still set it?

CMakeLists.txt
dyfer: no-math-errno was causing compilation issues, like:

```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk/usr/include/c++/v1/cmath:322:5: error:

      <cmath> tried including <math.h> but didn't find libc++'s <math.h> header. This usually means that your header search paths
      are not configured properly. The header search paths should contain the C++ Standard Library headers before any C Standard
      Library, and you are probably using compiler flags that make that not be the case.
  322 | #   error <cmath> tried including <math.h> but didn't find libc++'s <math.h> header. \
      |     ^

fatal error: too many errors emitted, stopping now [-ferror-limit=]
```

Any thoughts on this?

And should I add `no-signaling-math` to clang only?

CMakeLists.txt
JordanHendersonMusic: I think there should at least be a comment here saying that is what we support and that if we ever add something, we need to make sure msvc does the same thing.

CMakeLists.txt
JordanHendersonMusic: Ah okay this is a little complicated..

clang does not have `-fno-signaling-nans`

ignore what I said about `no-signaling-math` that is wrong, don't use that.

Perhaps just set `-fno-signaling-nans` on GCC. 

Re `no-math-errno `

> On some targets, math library functions never set errno, and so -fno-math-errno is the default. This includes most BSD-derived systems, including Darwin.

https://clang.llvm.org/docs/UsersManual.html

I think this means we want to set it, unless we are on apple where it seems to cause an error!?

JordanHendersonMusic (changes requested):
I think this looks good, just to check though that the submodule update is supposed to go in this PR?

dyfer:
> just to check though that the submodule update is supposed to go in this PR?

Yes! That pulls in the [change in nova-simd which fixed building on clang/arm64](https://github.com/timblechmann/nova-simd/pull/12) with `-Werror=return-type`. The submodule also had a couple of cosmetic changes since the last time we updated it, that's why the diff is a little larger.

JordanHendersonMusic (approved):
okay looks good

dyfer:
I also noticed that we set [`-fno-finite-math-only`](https://github.com/supercollider/supercollider/blob/develop/server/plugins/CMakeLists.txt#L61) for the plugins for GCC.
- should we touch it in this PR and apply it to Clang as well?
- if so, should that flag stay for plugins only or be used throughout?

JordanHendersonMusic:
Actually I think it should be removed because it is the default. 


>-ffinite-math-only
Allow optimizations for floating-point arithmetic that assume that arguments and results are not NaNs or +-Infs.
This option is not turned on by any -O option besides -Ofast since it can result in incorrect output for programs that depend on an exact implementation of IEEE or ISO rules/specifications for math functions. It may, however, yield faster code for programs that do not require the guarantees of these specifications.
The default is -fno-finite-math-only.

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

I think it is pretty safe to just say we do not support -Ofast, only -O3, with the few tweaks we have already added.


Seems it was originally added all the way back in 2010 https://github.com/supercollider/supercollider/commit/018a3b5a69b428c7ff8c0888737d067250abe59b 

so the gcc defaults might have changed since then which might explain why it was added in the first place.

dyfer:
> Seems it was originally added all the way back in 2010 [018a3b5](https://github.com/supercollider/supercollider/commit/018a3b5a69b428c7ff8c0888737d067250abe59b)
> 
> so the gcc defaults might have changed since then which might explain why it was added in the first place.

Ah, I see. In that case it might've been added because [we used to compile with -ffast-math](https://github.com/supercollider/supercollider/commit/018a3b5a69b428c7ff8c0888737d067250abe59b#diff-34dc2e44b14fc26c8651b204eaf2195382705e6526c6143edaa60a2e464c5e45R5), which was eventually removed in https://github.com/supercollider/supercollider/pull/4119.

So should I remove -fno-finite-math-only flag in this PR?

JordanHendersonMusic:
Ah that makes more sense! 

Yes I think since you are touching the maths stuff, it should go here.

dyfer:
I updated the PR. Still good to merge? (once CI passes)

CMakeLists.txt
dyfer: I added it to clang with a conditional switch disabling it on apple.

CMakeLists.txt
dyfer: I added the comment

dyfer:
@JordanHendersonMusic just to double check, can I merge this after also removing flags from [server/plugins/CMakeLists.txt](https://github.com/supercollider/supercollider/pull/7263/files#diff-3d1ccc9e1c72318a4384e2c1430e41a1b02019a0352cff7dbe4e6791eac2f9b7)?

JordanHendersonMusic:
Yes I think that will be fine since it was never enabled - merge away!

PR #7267, opened 2025-11-30
cmake: set/update compiler hardware optimizations, take 2

dyfer:
## Purpose and Motivation

This is a slightly different (cleaner?) implementation of #7248. It exposes a single cmake flag with all hardware-related compiler flags, which I hope makes it easier to use and maintain.

The rest of the description, copied from 7248, since the functionality is basically the same:

Previously we didn't explicitly use compiler optimizations for newer CPU instructions and features (like SSE4), but only SSE/SSE2.

I reorganized and updated compiler settings optimizing for modern hardware:

- GCC/Clang - Linux (and MinGW):
  - for GCC 11+ and Clang 12+ on x86_64 I used "architecture level" matching Haswell processors (2012-era)
    - for older compilers I default to SSE4 which I believe maps to 2009-era CPUs; I don't think it's needed to target anything older than that
  - for Arm64 I used `armv8-a` but I'm not sure how safe/conservative that is
  - optimizations for older 32-bit ARM (v6 and v7 - Rpi 1 - 2 - 3) were left in place, though using a newer CMake conventions; they are untested, unfortunately
 - Apple Clang:
   - I used "CPU family" approach for arm64 and "architecture level" for x86_64; for the latter the default is now set to Haswell processors (Apple hardware from 2013 onward); note that in the legacy build I set this to an earlier value (Ivy Bridge) to catch one generation of hardware that the legacy build can still run on (2012 MacBook Pro running macOS 10.15)
   - for arm64 (Apple CPU) the default value `apple-m1` doesn't change anything; however exposing it allows taking potential advantage of newer Apple chips
-  Windows MSVC:
   - I added the newer `SSE4.2` optimization, which is a relatively recent addition to MSVC (previously the first "higher" option that the default SSE2 was AVX); I didn't use AVX optimizations since some [lower-end chips from as late as 2021](https://en.wikipedia.org/wiki/Tremont_(microarchitecture)) still don't support it. If we decided to create a Windows legacy build, maintaining the basic SSE4.2 optimization, we could consider optimizing the main build for newer CPUs with e.g. AVX512.
 
FYI I haven't tested the actual performance of these changes.

IIUC these settings do not affect explicit uses of SSE/NEON instructions; e.g. in the nova-simd library.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested - at least the configurations we build in the CI
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -43,7 +43,7 @@ jobs:
             # homebrew-uninstall: "readline" # use this instead when cross-compiling for x86_64 on arm64
             # vcpkg-packages: "readline portaudio libsndfile fftw3" # use this instead when cross-compiling for x86_64 on arm64
             vcpkg-triplet: x64-osx-release-supercollider # required for build-libsndfile
-            extra-cmake-flags: ""
+            extra-cmake-flags: "-D SC_COMPILER_ARCH_FLAGS='-march=ivybridge'"
             artifact-suffix: "macOS-x64-legacy" # set if needed - will trigger artifact upload
             ctest: false
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,14 +106,6 @@ option(USE_CCACHE "Use ccache if available." ON)
 
 option(FFT_GREEN "Use internal 'Green' FFT lib rather than FFTW. (Not recommended.)" OFF)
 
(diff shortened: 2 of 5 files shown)


JordanHendersonMusic (approved):
Looks much better, thanks for taking the time to clean this up!

PR #7271, opened 2025-12-01
PyrObject: add MSVC hints

dyfer:
## Purpose and Motivation

MSVC was complaining about a GCC-specific instruction. This PR adds an equivalent for MSVC.

I looked up a solution for this - I don't have much expertise here. Please review carefully.

## Types of changes

- Bug fix (avoids compilation warning under MSVC)
- New feature (For MSVC?)

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/lang/LangSource/PyrObject.cpp
+++ b/lang/LangSource/PyrObject.cpp
@@ -54,6 +54,14 @@
 #    include <parallel/algorithm>
 #endif
 
+#if defined(__GNUC__) || defined(__clang__)
+#    define PRAGMA_IVDEP _Pragma("GCC ivdep")
+#elif defined(_MSC_VER)
+#    define PRAGMA_IVDEP __pragma(loop(ivdep))
+#else
+#    define PRAGMA_IVDEP
+#endif
+
 
 PyrClass* gClassList = nullptr;
 int gNumSelectors = 0;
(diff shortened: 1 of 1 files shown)


lang/LangSource/PyrObject.cpp
```diff
@@ -1053,7 +1061,7 @@ static void binsortClassRows(PyrMethod const** bigTable, const ColumnDescriptor*
         PyrMethod const** row = bigTable + j * numSelectors;
         memcpy(temprow, row, numSelectors * sizeof(PyrMethod*));
 
-#pragma GCC ivdep
+        PRAGMA_IVDEP
         for (int i = 0; i < numSelectors; ++i)
             row[i] = temprow[sels[i].selectorIndex];
```
JordanHendersonMusic, line 1066: This is fine the memory can't alias

lang/LangSource/PyrObject.cpp
```diff
@@ -1347,7 +1355,7 @@ static size_t fillClassRow(const PyrClass* classobj, PyrMethod** bigTable, boost
     if (superclassobj) {
         PyrMethod** superrow = bigTable + slotRawInt(&superclassobj->classIndex) * gNumSelectors;
 
-#pragma GCC ivdep
+        PRAGMA_IVDEP
         for (int i = 0; i != gNumSelectors; ++i) {
```
JordanHendersonMusic, line 1359: Also for, can only alias if it class is it's own super class - there is always a meta class in SC.

JordanHendersonMusic (approved):
Okay this seems fine, do you know what triggered the warning?

dyfer:
> Okay this seems fine, do you know what triggered the warning?

I'm not sure if I understand the question?
Previously `#pragma GCC ivdep` was not guarded with a preprocessor definition so MSVC would [complain](https://github.com/supercollider/supercollider/actions/runs/19834278728/job/56827781601#step:14:861) that it doesn't understand it - "unknown pragma". 
This PR replaces GCC/Clang-only definition with one for both GCC/Clang and for MSVC.

JordanHendersonMusic:
Ah, so this isn't a new issue caused by cmake changes!? That makes sense, okay, merge ahead.

PR #7272, opened 2025-12-02
Add Guide to .scd File Evaluation and Execution: AutoRun, Load, CLI Methods, etc

### LINKED ISSUE #7268: Documentation Request: Guide for .scd Document Evaluation/Execution (SC-IDE & CLI)
## Motivation
There are recurring questions on the forum regarding how to evaluate or execute .scd documents in SC-IDE and from the command line (terminal, DOS prompt).  
For example: [Evaluating a complete .scd file](https://scsynth.org/t/evaluating-a-complete-scd-file/12617?u=prko).
Currently, it is not easy to reach the relevant help documentation without knowing about .autoRun.  
A clear guide documentation would help prevent repeated questions and improve usability.

I propose adding a summary and comparison documentation that covers:
• `.autorun`
• `.load`
• `.loadRelative`
• using sclang in CLI
• (Possibly) other editor-specific methods. In the mentioned thread, emacs usage was discussed, but I am not sure if it is necessary to include. Emacs users are usually advanced and may not rely on help documentation.

## Plan for Implementation
If this idea is considered worthwhile, I am willing to draft an initial proposal. Although, judging from past experience, it seems quite likely that someone may only use my draft as a reference and then open a new PR instead. Still, it would be a good exercise for me and could provide inspiration for whoever finalises the documentation.

prko:
close #7268
## Purpose and Motivation

There are recurring questions on the forum regarding how to evaluate or execute .scd documents in SC-IDE and from the command line (terminal, DOS prompt).
For example: [Evaluating a complete .scd file](https://scsynth.org/t/evaluating-a-complete-scd-file/12617?u=prko).
Currently, it is not easy to reach the relevant help documentation without knowing about .autoRun.
A clear guide documentation would help prevent repeated questions and improve usability.

Description of Proposed Feature
I propose adding a summary and comparison documentation that covers:
- `/*RUN*/` when `.autorun` is `true`
- `.load`
- `.loadRelative`
- Compile String or File
- using sclang in CLI

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


prko:
After testing all the code snippets in the documentation on macOS, Windows 11 and Ubuntu, I updated the documentation. I am also attaching the rendered PDF:
[Evaluating and Executing .scd Files (SC-IDE & CLI) _ SuperCollider 3.15.0-dev Help.pdf](https://github.com/user-attachments/files/23915467/Evaluating.and.Executing.scd.Files.SC-IDE.CLI._.SuperCollider.3.15.0-dev.Help.pdf)

prko:
Oh... 
The default CSS does not print the document correctly. The following uses my custom CSS file:
[Evaluating and Executing .scd Files (SC-IDE & CLI) _ SuperCollider 3.15.0-dev Help.pdf](https://github.com/user-attachments/files/23915534/Evaluating.and.Executing.scd.Files.SC-IDE.CLI._.SuperCollider.3.15.0-dev.Help.pdf)

@tedmoore
Would you be interested in the Pdf difference as well?

HotwheelsSisyphus:
Thanks, very welcome addition! I'll try to look this over properly within the next few days. For now, two quick questions:

1. Have you maybe discussed the "information"-quasi-tag (that you use repeatedly) with @tedmoore already? I'm not against adding new tags, but we should probably consider how that would work in the new system, and what to call them. I'm not sure if "information" is all that descriptive (as almost everything can be considered "information"), we could think of something better maybe. (the same goes for "note", really.) Something to think about: how would you explain the difference between "information" and "note" tags? 

2. "SCD Documents": I think I am more used to seeing file name extensions in lower case, e.g. "txt documents", "exe file", or whatever. I'm not sure if there's a convention on that. Maybe @capital-G knows?

EDIT: Also, so I don't forget it later: The startup file (on all platforms) lives in userConfigDir, which on MacOS defaults to the unix XDG_CONFIG when that is set (e.g. when using scnvim from within zsh); only if the XDG dir is not set, the userConfigDir forwards to the mac appSupportDir instead (that's what happens in the scide). I've been meaning to document that, too...

capital-G:
> 2\. "SCD Documents": I think I am more used to seeing file name extensions in lower case, e.g. "txt documents", "exe file", or whatever. I'm not sure if there's a convention on that. Maybe @capital-G knows?

Lowercase is the way to go :)

prko:
>1. Have you maybe discussed the "information"-quasi-tag (that you use repeatedly) with @tedmoore already? I'm not against adding new tags, but we should probably consider how that would work in the new system, and what to call them. I'm not sure if "information" is all that descriptive (as almost everything can be considered "information"), we could think of something better maybe. (the same goes for "note", really.) Something to think about: how would you explain the difference between "information" and "note" tags?

No, I have not yet discussed the tags such as Information, note, warning, etc. with others.  
I do recall that @capital-G mentioned something similar last year:

- https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html

- https://myst-parser.readthedocs.io/en/v0.17.1/syntax/optional.html#syntax-admonitions

However, my memory was mistaken — there is no information tag in those documents referenced by @capital-G.

I simply chose to use information because I find the long yellow box less readable.

---

>>2. "SCD Documents": I think I am more used to seeing file name extensions in lower case, e.g. "txt documents", "exe file", or whatever. I'm not sure if there's a convention on that. Maybe @capital-G knows?
>
>Lowercase is the way to go :)

I have amended it in commit [89ab39f](https://github.com/supercollider/supercollider/pull/7272/commits/89ab39f7a14e3ab164aa972407b0890c4057254c).

--
>EDIT: Also, so I don't forget it later: The startup file (on all platforms) lives in userConfigDir, which on MacOS defaults to the unix XDG_CONFIG when that is set (e.g. when using scnvim from within zsh); only if the XDG dir is not set, the userConfigDir forwards to the mac appSupportDir instead (that's what happens in the scide). I've been meaning to document that, too...

Ehm..., I mostly use SC‑IDE, so this is new to me. It would indeed be very helpful if it were documented. Thank you!

prko:
I slightly revised the documentation:
[Evaluating and Executing scd Files (SC-IDE & CLI) _ SuperCollider 3.15.0-dev Help.pdf](https://github.com/user-attachments/files/23962987/Evaluating.and.Executing.scd.Files.SC-IDE.CLI._.SuperCollider.3.15.0-dev.Help.pdf)

prko:
I slightly revised the documentation (also replaced the information tag with the tip tag):
[Evaluating and Executing scd Files (SC-IDE & CLI) _ SuperCollider 3.15.0-dev Help.pdf](https://github.com/user-attachments/files/23973250/Evaluating.and.Executing.scd.Files.SC-IDE.CLI._.SuperCollider.3.15.0-dev.Help.pdf)

tedmoore:
thanks @prko and others here. this is a great addition! 🎉

prko:
@tedmoore Thanks!

I have replaced the term method(s) with option(s) or ways, as I believe that method could be confusing for beginners, given its specific use in object‑oriented programming to describe a class method.

[Evaluating and Executing scd Files (SC-IDE & CLI) _ SuperCollider 3.15.0-dev Help.pdf](https://github.com/user-attachments/files/24014015/Evaluating.and.Executing.scd.Files.SC-IDE.CLI._.SuperCollider.3.15.0-dev.Help.pdf)

prko:
I have two questions:
1. Should `thisProcess.interpreter.executeFile(path)` be included? 
   (my thought: it depends... maybe yes?)
2. Should the following be included?
    - `ServerBoot.add`
    - `ServerTree.add`
    - `ServerQuit.add`

   (my thought: I cannot judge it correctly... maybe no?)

Please refer to https://scsynth.org/t/how-to-load-startup-scd-path-automatically-on-book/12662?u=prko

prko:
I tested `thisProcess.interpreter.executeFile(path, arg1, arg2)` and found no clear advantage compared with load. It may be that my understanding does not encompass its intended use. 

So...,
incorporating `thisProcess.interpreter.executeFile(path ... args)` is currently beyond my capacity.

HotwheelsSisyphus:
> I tested `thisProcess.interpreter.executeFile(path, arg1, arg2)` and found no clear advantage compared with load. It may be that my understanding does not encompass its intended use.
> 
> So..., incorporating `thisProcess.interpreter.executeFile(path ... args)` is currently beyond my capacity.

It might be good to actually look at the implementations for the things you are documenting. It's a bit of sleuthing but there (in common/collection/string.sc and common/core/Kernel.sc) you can see that

a. `"string".load` simply forwards to `thisProcess.interpreter.executeFile("string")`; and
b. `thisProcess.interpreter.executeFile(path...args)` contains a protected call to `thisProcess.interpreter.compileFile(path).valueArray(args)`.

So in this case, I'd say that `String.load` is a shortcut to `Interpreter.executeFile`, but which doesn't allow you to send additional arguments.  what are the arguments for? Well, they get passed to whatever `compileFile` returns (which could be a function, for example). 

You can find implementations through the help browser (upper left, below heading: E.g. "Source: Kernel.sc" in the [web version](https://doc.sccode.org/Classes/Interpreter.html), except in the scide the link will actually work and open that file), or when it's methods, you can also grep the classlibrary directory. I do this all the time when I try to document stuff. Sometimes, trial and error is enough to figure out what something does, but looking at the source is really the only way to know for sure.

prko:
@HotwheelsSisyphus 
>a. "string".load simply forwards to `thisProcess.interpreter.executeFile("string");` and
b. `thisProcess.interpreter.executeFile(path...args)` contains a protected call to `thisProcess.interpreter.compileFile(path).valueArray(args)`.

I had checked these before posting my previous message.

Well…  

On this basis, I was not able to construct a meaningful use case.
My understanding:

1. To use `thisProcess.interpreter.executeFile(path...args)`, the corresponding .scd file should be a function without the enclosing curly braces `{}`.  
   For example:
   ```
   |a, b|
   "Arguments received: % and %".format(a, b).postln;
   a + b
   ```
2. Then, to execute that file:
   ```thisProcess.interpreter.executeFile("functionTest.scd".resolveRelative, 10, 20);```

However, I cannot envisage a practical use case for this. That is why I am unable to provide a convincing code example. For me, load and loadRelative seem preferable to `thisProcess.interpreter.executeFile(path...args)`.

HotwheelsSisyphus:
Sorry, I misunderstood.
 
 > I cannot envisage a practical use case for this
 
I don't know, maybe in some setup contexts where you pass certain parameters (e.g., info about hardware, number of channels) to some more complicated code? Anyway it seems reasonable to me that, if there's a way to run some code, you should also be able to pass parameters to that code. Evidently it doesn't mean that one needs to use it all the time. Just use the addition example you provided...

prko:
No need to apologise. I described it rather obscurely; I should have written it more concretely.

> (e.g., info about hardware, number of channels) to some more complicated code?

Yes, it could be… but I still cannot think of a good example. I asked about this in the forum.

prko:
I added an example using thisProcess.interpreter.executeFile.  
I also reformatted the schelp file.  
I believe this PR is ready for review.

HotwheelsSisyphus:
> Should the following be included?
>
> ServerBoot.add
> ServerTree.add
> ServerQuit.add

I think it would a good idea to mention them, though not separately. Something like: 

"Subclasses of AbstractSystemAction and AbstractServerAction implement .add methods; for instance, 
`StartUp.add { thisProcess.interpreter.executeFile("someFile.scd") } `"
or 
`ServerBoot.add { thisProcess.interpreter.executeFile("someFile.scd") } `
or
`CmdPeriod.add { thisProcess.interpreter.executeFile("someFile.scd") } `

(or some other example. I haven't tested those yet!)


------------------------------------
> I believe this PR is ready for review.

could you also attach a pdf? Thanks!

prko:
@HotwheelsSisyphus 

I have revised the documentation to reflect your suggestions.  

The code in the `Event hooks (CmdPeriod, ShutDown, ServerBoot, ServerTree, ServerQuit)` section has been tested on macOS, while the other code examples have been tested on macOS, Ubuntu, and Windows.

I am also attaching the PDF versions:

- using custom css:
   [Executing and Loading .scd Files and SystemAction Hooks (SC-IDE & CLI) custom css.pdf](https://github.com/user-attachments/files/24213676/Executing.and.Loading.scd.Files.and.SystemAction.Hooks.SC-IDE.CLI.custom.css.pdf)

- using default css (Every first line on each page is hidden by the floating menu.):
   [Executing and Loading .scd Files and SystemAction Hooks (SC-IDE & CLI) default css.pdf](https://github.com/user-attachments/files/24213679/Executing.and.Loading.scd.Files.and.SystemAction.Hooks.SC-IDE.CLI.default.css.pdf)

prko:
@HotwheelsSisyphus

I have revised the documentation to reflect your suggestions.

The code in the `Event hooks (CmdPeriod, ShutDown, ServerBoot, ServerTree, ServerQuit)` section has been tested on macOS, while the other code examples have been tested on macOS, Ubuntu, and Windows.

I am also attaching the PDF versions:

- using custom css:
   [Executing and Loading .scd Files and SystemAction Hooks (SC-IDE & CLI) custom css.pdf](https://github.com/user-attachments/files/24213971/Executing.and.Loading.scd.Files.and.SystemAction.Hooks.SC-IDE.CLI.custom.css.pdf)

- using default css (Every first line on each page is hidden by the floating menu.):
   [Executing and Loading .scd Files and SystemAction Hooks (SC-IDE & CLI) default css.pdf](https://github.com/user-attachments/files/24213972/Executing.and.Loading.scd.Files.and.SystemAction.Hooks.SC-IDE.CLI.default.css.pdf)

prko:
Please note that this pull request will require revision after PR #7291 is merged.

As I am not a native English speaker, it may be more efficient if the reviewer could directly suggest the necessary changes. I would be very happy for this pull request to be co‑authored, as I believe working together on my draft will be the most effective way forward.

HotwheelsSisyphus:
hi, just wanted to say I'm getting back to this now. It's a bit tricky because I think the current structure is a bit confusing, so it isn't really about the English (which seems fine for the most part), but more about what gets discussed where in what depth, etc. 

Just a few notes, more tomorrow hopefully:
1. No need to list all the SystemActions in the overview at the start. I'd list the superclass and maybe one example.
(btw, the startup file is not the same as the StartUp system action; startup file gets executed first.)
2. Otoh, we don't need a huge example for every little thing in _this_ file; for instance, one can simply mention/link the startup file here and, for instance, add a larger example in the Startup file helpfile.
3. There's probably a more elegant way to deal with the example scd files:
a) either make them on the spot with File.open in the Platfrom.defaultTempDir or
b) include the files in questions somewhere in the help directory. 
4. it seems the explanation about PATH takes a lot of space. As this isn't strictly a supercollider thing, maybe we can find some other resource to link to that explains this? Also, I think people who want headless operation probably can be assumed to be familiar with this stuff already.

HotwheelsSisyphus:
So I'm now starting to move some of your material regarding the startup file into Reference/StartUpFile.schelp. I'll make a separate PR for that and name you as coauthor. 

Quick question though: What's the default UserConfigDir on windows? I don't have a windows installation ready to try it out. The current startupfile.schelp says:
"C:\\SuperCollider\\startup.scd"
which seems kind of weird to me (\\ twice in a path?) but I haven't used windows in a while...

prko:
> So I'm now starting to move some of your material regarding the startup file into Reference/StartUpFile.schelp. I'll make a separate PR for that and name you as coauthor.

Thanks!

---

>Quick question though: What's the default UserConfigDir on windows? 

`Platform.userConfigDir`  
returns:
→ `C:\Users\prko\AppData\Local\SuperCollider`  
In general: `~/AppData/Local/SuperCollider`

In `startup.scd`, opened via the File menu (“Open startup file ...”),  
evaluating `Document.current.path` returns:
→ `C:/Users/prko/AppData/Local/SuperCollider/startup.scd`
In general: `~/AppData/Local/SuperCollider/startup.scd`?

---

Thanks also for the notes — I agree the overview doesn’t need to list all the `SystemActions`. Mentioning the superclass and one clear example should be enough there.

> 1. No need to list all the SystemActions in the overview at the start. I'd list the superclass and maybe one example.  
> (btw, the startup file is not the same as the StartUp system action; startup file gets executed first.)  
> 2. Otoh, we don't need a huge example for every little thing in this file; for instance, one can simply mention/link the startup file here and, for instance, add a larger example in the Startup file helpfile.

Good point about distinguishing the startup file from the `StartUp` system action; I’ll make sure that’s clarified.

> 3. There's probably a more elegant way to deal with the example scd files:  
> a) either make them on the spot with `File.open` in `Platform.defaultTempDir` or  
> b) include the files in question somewhere in the help directory.

I also agree we don’t need to overload this file with long examples. A short mention and a link to the dedicated help file seems more elegant.  
For the example `.scd` files, I like your suggestion: either generate them on the fly with `File.open` in `Platform.defaultTempDir`, or include them in the help directory.

> 4. It seems the explanation about PATH takes a lot of space. As this isn't strictly a SuperCollider thing, maybe we can find some other resource to link to that explains this? Also, I think people who want headless operation probably can be assumed to be familiar with this stuff already.

When I tried headless operation for the first time, it wasn’t easy since there were no instructions. I spent a lot of time figuring it out. Nowadays, it might be easier if AI gives a correct answer, but I haven’t tested that. I think we could preserve the explanation for beginners…

HotwheelsSisyphus:
> When I tried headless operation for the first time, it wasn’t easy since there were no instructions. I spent a lot of time figuring it out. Nowadays, it might be easier if AI gives a correct answer, but I haven’t tested that. I think we could preserve the explanation for beginners…

I mainly thought we might be able to find tutorial that explains these things, so there is less for us to maintain. But I looked around yesterday and it turns out the "official" docs for these things are a bit too concise for beginners. and I'm also not sure we want to link to some stackexchange explanation (even though those are often very good)...  so yeah, let's keep that. 

I've never directly used the sclang CLI for anything other than -h, but I think that every `sclang filename.scd` will start a new sclang process. Maybe a way to send files to an existing sclang process should also be included, which I think is done by piping it stdin, but I'm not sure off the top of my head of the right way to do that. mkfifo?

prko:
> so yeah, let's keep that.

Thanks!

> Maybe a way to send files to an existing sclang process should also be included, which I think is done by piping it stdin, but I'm not sure off the top of my head of the right way to do that. mkfifo?

Before writing my opinion here, I should understand more clearly what you mean by "a way to send files to an existing sclang process." Do you mean something like the following?

1. SC-IDE is running on a machine. (sclang A)
2. On the same machine, a user launches `sclang path/filename.scd` in a terminal window, so a new sclang process is started. (sclang B)
3. On the same machine, the user wants to evaluate a complete `.scd` file using either sclang A or sclang B via a terminal command.

I have never considered case No. 3.  

My experience has only been sending short code lines from 'sclang B' to 'sclang A' as OSC messages, either between different machines or on the same machine (mainly sclang <-> Max). Of course, 'sclang' listens to the incoming OSC message and evaluates it automatically via the `.interpret` method.

It seems that `mkfifo` exists only on macOS and Linux. If it were possible to open a document in an already running application without knowing its PID, that would be amazing.

prko:
Could you let me know the current status of this PR when you have a chance? The initial feedback seemed positive, and I have tried to address the requested changes, along with some additional cleanup. Since then, I have not seen any further response, so I just wanted to check in.

HotwheelsSisyphus:
oh sorry, I'll try to look at it again these days...

PR #7273, opened 2025-12-02
msvc: change silenced warnings

dyfer:
## Purpose and Motivation

This PR silences MSVC warnings, resulting in a much cleaner build.

I'm not sure if all of these should be silenced - maybe we should fix some of them?

`_ENABLE_ATOMIC_ALIGNMENT_FIX` doesn't seem to be needed anymore.

`_CRT_SECURE_NO_WARNINGS _SCL_SECURE_NO_WARNINGS` seem to be silenced by the flags anyway. EDIT: that's the case only for external libraries; in the main codebase we'd want to fix those anyway.

## Types of changes

- Cleanup

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation


JordanHendersonMusic:
> I'm not sure if all of these should be silenced - maybe we should fix some of them?

Is it possible to disable warnings for the external libraries only? It would be great if we could make all warnings into errors in our code.

dyfer:
> Is it possible to disable warnings for the external libraries only?

Yes, I think so. Would you like me to do that, and for the main codebase disable only zero-sized array, removing other suppressions?

JordanHendersonMusic:
I think that would be the right think to do. Some/most of those warnings are things we really ought to fix.

JordanHendersonMusic:
I think the unreferenced parameter could be silenced too.

dyfer:
> I think the unreferenced parameter could be silenced too.

 Do you mean `4101 unreferenced local variable` or `4102 unreferenced label` or both?

EDIT: Sorry, I see it now. Unreferenced parameter is another warning (4100).
Still, should 4101 and 4102 be silenced?

JordanHendersonMusic:
Sorry for not getting back to you. Yes I think this is good!

JordanHendersonMusic (approved):

JordanHendersonMusic:
Separately, it might be worth considering things like yaml-cpp, they are using deprecated features, what should we do about that?

CMakeLists.txt
```diff
@@ -401,26 +401,14 @@ endif()
...
-    /wd4309 # '=' : truncation of constant value
-    /wd4800 # forcing value to bool 'true' or 'false' (performance warning)
-    /wd4996 # The POSIX name for this item is deprecated.
+    /wd4200 # nonstandard extension used: zero-sized array in struct/union
   )
 
-# this flag causes MSVC to correctly report __cplusplus, which prevents a compiler error
-# caused by some versions of libsndfile's C++ header redefining nullptr as NULL.
-# See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
-  add_compile_options("/Zc:__cplusplus")
```
timblechmann, line 417: you probably want to keep this flag: without it __cplusplus is *always* defined to 199711L

https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

dyfer: Gotcha. I misunderstood when I looked it up - I thought that on newer MSVC and from c++17 on it was _automatically_ defined as a newer version.

dyfer:
> Separately, it might be worth considering things like yaml-cpp, they are using deprecated features, what should we do about that?

It would be good to update our 3rd-party libraries from time to time, but it doesn't seem we have enough dev power to do it atm.

I don't have strong opinion whether we should silence compilation errors from outdated libraries or not. 

EDIT: I added note to the warnings silenced for external libraries, so that we know what to look at

dyfer:
@JordanHendersonMusic what do you think about the state of the PR after the latest changes?

PR #7278, opened 2025-12-05
Recorder.prRecord: addAction becomes \addAfter if target node isKindOf(Synth)

HotwheelsSisyphus:
(I promise this is my last Recorder-related pr.)

Recorder currently always uses the `\addToTail` addAction for the recording synth, on the assumption that the target is a **Group** (even if the argument is called "node"); it may be useful occasionally to specify a specific **Synth** to record after. Currently, supplying a Synth as `node` arg fails, because Synths don't have tails; only groups do. 

**This simply checks whether the target node `isKindOf(Synth)`, and if it is not, everything is as before (\addToTail); if it _is_ a Synth, \addAfter instead.** 

## Purpose and Motivatio
Mainly I don't want to have to explain in the docs that the `node` argument needs to be, in fact, a group. And renaming the argument might break things. 
Also, presumably user-friendly, if user happens to think in terms of Synths rather than groups.

Possible niche application: 
SomeSignal (Dry) -> SomeFxInsert (Wet), where we want to record dry and wet simultaneously and the insert uses a ReplaceOut (i.e., you wouldn't be able to simply record separate busses). 

## Types of changes
- New feature

### DIFF
--- a/SCClassLibrary/Common/Control/Recorder.sc
+++ b/SCClassLibrary/Common/Control/Recorder.sc
@@ -138,7 +138,11 @@ Recorder {
 	/* private implementation */
 
 	prRecord { |bus, node, dur|
-		recordNode = Synth.tail(node ? RootNode(server), synthDef.name, [\bufnum, recordBuf, \in, bus, \duration, dur ? -1]);
+		recordNode = Synth(synthDef.name, 
+						args: [\bufnum, recordBuf, \in, bus, \duration, dur ? -1],
+						target: node ? RootNode(server), 
+						addAction: if(node.isKindOf(Synth), \addAfter, \addToTail)
+		);
 		recordNode.register(true);
 		recordNode.onFree { this.stopRecording };
 		if(responder.isNil) {


JordanHendersonMusic:
Why can't we just make `addToTail` on synths the same as `after`? Would this break anything?

HotwheelsSisyphus:
> Why can't we just make addToTail on synths the same as after? Would this break anything?

I hadn't thought of that! Easy to do  I think, but a wider-ranging change.. I permit myself to mess with the Recorder class because I think of that as probably not performance critical, i.e., it's probably not part of people's pieces; the Synth class on the other hand very much is... 

Anyway, adding this to Synth.new would do ( + 2 turns head into before and tail into after):
```
		if(target.isKindOf(Synth) and: {addActionID < 2}) {addActionID = addActionID + 2};

```
Maybe there is a faster way though? Any different with `target.class == Synth` for example?

As for what this could break, idk. Currently one gets a
```
FAILURE IN SERVER /s_new Group 1001 not found
```
when supplying a Synth as another Synth's target with addToTail (so nothing to break there), but maybe there are some other issues I'm not aware of.

telephon:
> Why can't we just make `addToTail` on synths the same as `after`? Would this break anything?

I think this is a good idea. What it may "break" is that things work without failure in a slightly different way than expected? But I'd be for it.

HotwheelsSisyphus:
I've looked at it for slightly longer and it's definitely more than just one line that needs to be added (even though mostly just copypasting that one line into the different variants of the `*new` method, i.e. `newMsg`, `newPaused`, etc.). 
Otoh, I also tried figuring out where those things are handled in the server (scsynth/SC_MiscCmds.cpp), and it's not necessarily cleaner there (provided I've even found the right place), because different addActions result in different error checks on the target, not the other way around.

So a language side fix is probably better - I'll maybe just rename and reuse this PR, or should I open a different one?

PR #7281, opened 2025-12-08
build: remove NO_GPL3 option

### LINKED ISSUE #7264: Remove `NO_GPL3` flag?
## Motivation

While cleaning up cmake options, I started investigating the `NO_GPL3` flag.

Licencing.html says: 

> SuperCollider was originally published under the terms of _version 2 or later_ of the GPL, but the application now includes some GPL3-licensed code. If you have a need specifically for GPL2 compatibility then it is possible to recompile SuperCollider from source without the GPL3 elements.

Cmake setup has the following text: `NO_GPL3 "Disable GPL3 code, for pure-GPL2 situations. (Not recommended.)`

We've had many contributions over the years and most source files specify that they are GPL 2, but some do specify GPL3. Examples: [1](https://github.com/supercollider/supercollider/blob/37cd92c5adf4093ac918e69be1abc65869fb6878/SCDoc/SCDoc.tab.hpp#L10), [2](https://github.com/supercollider/supercollider/blob/37cd92c5adf4093ac918e69be1abc65869fb6878/lang/LangSource/Bison/lang11d_tab.h#L10).

[Supercollider as a whole is GPL 3](https://github.com/supercollider/supercollider/blob/develop/COPYING#L2). BTW does that mean that files without the copyright notice are automatically GPL 3?

## Description of the Issue

The flag in question currently only [disables `SC_VIM` extension](https://github.com/supercollider/supercollider/blob/37cd92c5adf4093ac918e69be1abc65869fb6878/editors/CMakeLists.txt#L15-L17), which is GPL3. However, it does not exclude other GPL3 code (like the one mentioned earlier) from the build.

## Plan for Implementation

It would take extra work to make `NO_GPL3` flag exclude all of the GPL3 code. Also I'm not sure if it's even possible at this point.

I propose to remove the `NO_GPL3` flag. However, I don't know if there are other implications in doing so.

dyfer:
## Purpose and Motivation

We have a mix of GPL2 and GPL3 code at this point, it's likely not possible nor practical to exclude all GPL3 code.

## Types of changes

- Documentation
- Cleanup

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,7 +121,6 @@ if(WIN32)
 else(WIN32)
     option(NO_AVAHI "Disable Avahi support. (Not recommended.)" OFF)
 endif(WIN32)
-option(NO_GPL3 "Disable GPL3 code, for pure-GPL2 situations. (Not recommended.)" OFF)
 
 option(SCLANG_SERVER "Build with internal server." ON)
 
@@ -425,10 +424,6 @@ endif(MSVC)
 
 if(MINGW)
 	add_compile_options(-mstackrealign)
-endif()
-
-if (NO_GPL3)
-	add_compile_definitions(NO_GPL3)
 endif()
 
 if (SC_MEMORY_DEBUGGING)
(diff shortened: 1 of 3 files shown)


capital-G (approved):

capital-G:
SC switched to GPL 3.0 16 years ago via 2b581dfc266f6b39bb1cb591e293a2c872e1f56a - so every contribution since then w/o dedicated copyright disclaimer has been made with a GPL 3.0 license, which makes it not really possible to compile SC with GPL 2.0 exclusively - I therefore approved and merged this PR.

PR #7282, opened 2025-12-08
Switch to type safe qt connections within IDE

### LINKED ISSUE #6824: Replace "old-style"

capital-G:
## Purpose and Motivation

Closes #6824

This PR replaces many (but not all) connections with their implicit "SIGNAL/SLOT" syntax with a more type safe variant where we pass function pointers.
This hopefully reduces runtime errors while also improves clarity within the codebase and also allows to be clear where each signal is used. Though I am unsure which kind of checks the SIGNAL/SLOT syntax was performing internally, if it was performed during runtime or compile time.

While writing this I did not use XCode as a generator but Ninja. When using XCode I had errors all over the place - for the old and the new syntax, see

grafik

If anyone can give me a hint on why and how to fix that - this would be greatly appreciated! I do not see this errors in VSCode w/ XCode, though I also don't see any errors when I create wrong connections. Maybe @timblechmann or @jleben can provide some advice what the best setup is w/o relying on the Qt Creator?

There are two connections which I did not replace, mainly

https://github.com/supercollider/supercollider/blob/develop/editors/sc-ide/widgets/help_browser.cpp#L101-L102

https://github.com/supercollider/supercollider/blob/develop/editors/sc-ide/widgets/help_browser.cpp#L458

I think those could be broken b/c the types do not match in any way - maybe we should fix those? Though I did not want to make this PR become more excessive than it already is.

I also did not touch any connections made through https://github.com/supercollider/supercollider/blob/develop/editors/sc-ide/core/sig_mux.cpp - I do not really understand what this is doing as there is no documentation, so it is probably best to delay its implementation in another PR.

Additionally, there is a strange bug I am experiencing, though I am unsure if this got introduced by this change:
* If I open the Qt IDE with a debugger, it crashes
* If I open the Qt IDE when moved to `/Applications` it crashes

Otherwise this runs fine (so running it any other place and w/o a debugger), so I am a bit unsure where the problem is here. Would therefore be great if someone could build this on their local computer and verify this.

Regarding review: Please double check if I did not screw up any signals.

E.g. 

```diff
-connect(&mTmpCoalTimer, SIGNAL(timeout()), this, SLOT(onTmpCoalUsecs()));
+connect(&mTmpCoalTimer, &QTimer::timeout, this, &Document::onTmpCoalUsecs);
```

the 2nd and 4th parameters are crucial here - they should be the same, though only typed now.

edit: While looking through the code I noticed that we have some memory leaks within the IDE - we should fix those by making resources member variables instead of allocating new objects which are never freed.

## Types of changes

- Maintenance

### DIFF
--- a/editors/sc-ide/widgets/settings/sclang_page.cpp
+++ b/editors/sc-ide/widgets/settings/sclang_page.cpp
@@ -53,17 +53,17 @@ SclangPage::SclangPage(QWidget* parent): QWidget(parent), ui(new Ui::SclangConfi
     connect(ui->activeConfigFileComboBox, &QComboBox::currentTextChanged, this,
             &SclangPage::changeSelectedLanguageConfig);
 
-    connect(ui->sclang_add_configfile, SIGNAL(clicked()), this, SLOT(dialogCreateNewConfigFile()));
-    connect(ui->sclang_remove_configfile, SIGNAL(clicked()), this, SLOT(dialogDeleteCurrentConfigFile()));
+    connect(ui->sclang_add_configfile, &QToolButton::clicked, this, &SclangPage::dialogCreateNewConfigFile);
+    connect(ui->sclang_remove_configfile, &QToolButton::clicked, this, &SclangPage::dialogDeleteCurrentConfigFile);
 
-    connect(ui->sclang_add_include, SIGNAL(clicked()), this, SLOT(addIncludePath()));
-    connect(ui->sclang_add_exclude, SIGNAL(clicked()), this, SLOT(addExcludePath()));
+    connect(ui->sclang_add_include, &QToolButton::clicked, this, &SclangPage::addIncludePath);
+    connect(ui->sclang_add_exclude, &QToolButton::clicked, this, &SclangPage::addExcludePath);
 
-    connect(ui->sclang_remove_include, SIGNAL(clicked()), this, SLOT(removeIncludePath()));
-    connect(ui->sclang_remove_exclude, SIGNAL(clicked()), this, SLOT(removeExcludePath()));
+    connect(ui->sclang_remove_include, &QToolButton::clicked, this, &SclangPage::removeIncludePath);
+    connect(ui->sclang_remove_exclude, &QToolButton::clicked, this, &SclangPage::removeExcludePath);
 
-    connect(ui->sclang_post_inline_warnings, SIGNAL(stateChanged(int)), this, SLOT(sclangConfigChanged()));
-    connect(ui->sclang_exclude_default_paths, SIGNAL(stateChanged(int)), this, SLOT(sclangConfigChanged()));
+    connect(ui->sclang_post_inline_warnings, &QCheckBox::stateChanged, this, &SclangPage::markSclangConfigDirty);
+    connect(ui->sclang_exclude_default_paths, &QCheckBox::stateChanged, this, &SclangPage::markSclangConfigDirty);
 
     connect(ui->sclang_port, SIGNAL(valueChanged(int)), this, SLOT(showLanguageRestartDialogOnSave()));
 }
--- a/editors/sc-ide/widgets/settings/sclang_page.cpp
+++ b/editors/sc-ide/widgets/settings/sclang_page.cpp
@@ -65,7 +65,11 @@ SclangPage::SclangPage(QWidget* parent): QWidget(parent), ui(new Ui::SclangConfi
     connect(ui->sclang_post_inline_warnings, &QCheckBox::stateChanged, this, &SclangPage::sclangConfigChanged);
     connect(ui->sclang_exclude_default_paths, &QCheckBox::stateChanged, this, &SclangPage::sclangConfigChanged);
 
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     connect(ui->sclang_port, SIGNAL(valueChanged(int)), this, SLOT(showLanguageRestartDialogOnSave()));
+#else
+    connect(ui->sclang_port, &QSpinBox::valueChanged, this, &SclangPage::showLanguageRestartDialogOnSave);
+#endif
 }
 
 SclangPage::~SclangPage() { delete ui; }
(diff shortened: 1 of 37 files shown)


capital-G:
There seems to be an error on windows 32 bit regarding

https://github.com/capital-G/supercollider/blob/d2f5c7d2db8ce3096cec8eab74ae886b10c854ba/editors/sc-ide/widgets/editor_box.cpp#L55-L56

and one error on macOS 64 bit legacy

https://github.com/capital-G/supercollider/blob/d2f5c7d2db8ce3096cec8eab74ae886b10c854ba/editors/sc-ide/widgets/help_browser.cpp#L111

edit: Both builds use Qt 5.15.2, so there is probably a different API/Types in place for Qt5 - though it is very interesting that both platforms show different errors?

capital-G:
@dyfer do we still support Qt5 for 3.15? See #7006, which would make the errors disappear w/o patching/researching.
I am fine either way.

dyfer:
TL;DR I can't decide 🫣

My internal desire is to support older OSs as much as we can, but with limited dev resources I don't want us to bend over backwards either.
In practical terms, removing Qt 5.15 support will block us from building SC for Windows older than 10 (Win 7/8); but we're not doing that for the releases anymore anyway, so maybe there's no need for it?
I'm not sure how much removing support for Qt 5 would influence building for embedded platforms, like RPi (esp the older models).
With Qt5 removed we won't be able to build for 32-bit Windows anymore, thought that's likely not needed, aside from testing 32-bit builds...
We can still have a macOS legacy build with older Qt6 that would still support the oldest macOS we otherwise can target (10.15). We'd just loose QuartzComposerView, but that's obsolete functionality anyway.

capital-G:
I'd say lets see how much work it is to get it done within Qt6 - it seems desirable to stick to it b/c it does not seem to be too complicated to maintain compatibility currently, but the gains are still there.

Have you tested if you can install this build and open w/o crashes? Would be interested if it is my machine or if I screwed something up.

capital-G:
@dyfer I think we should delay this until we remove qt5 - it seems qt5 does not really like it and since I get all these errors i CLion I can not even tell which could be candidates for fixing regarding qt5 and which would be fine.

dyfer:
@capital-G I think we only needed one more revert to the old style for Qt5 and then it builds: https://github.com/dyfer/supercollider/commit/cdd5b1613fd28cb521c6ae5d48f840a5f8d908a2

I think it would be worthwhile to merge this in!

(I haven't gone through all the connections in detail though - yet)

editors/sc-ide/widgets/settings/sclang_page.cpp
```diff
@@ -53,17 +53,17 @@ SclangPage::SclangPage(QWidget* parent): QWidget(parent), ui(new Ui::SclangConfi
...
-    connect(ui->sclang_remove_exclude, SIGNAL(clicked()), this, SLOT(removeExcludePath()));
+    connect(ui->sclang_remove_include, &QToolButton::clicked, this, &SclangPage::removeIncludePath);
+    connect(ui->sclang_remove_exclude, &QToolButton::clicked, this, &SclangPage::removeExcludePath);
 
-    connect(ui->sclang_post_inline_warnings, SIGNAL(stateChanged(int)), this, SLOT(sclangConfigChanged()));
-    connect(ui->sclang_exclude_default_paths, SIGNAL(stateChanged(int)), this, SLOT(sclangConfigChanged()));
+    connect(ui->sclang_post_inline_warnings, &QCheckBox::stateChanged, this, &SclangPage::markSclangConfigDirty);
+    connect(ui->sclang_exclude_default_paths, &QCheckBox::stateChanged, this, &SclangPage::markSclangConfigDirty);
 
     connect(ui->sclang_port, SIGNAL(valueChanged(int)), this, SLOT(showLanguageRestartDialogOnSave()));
```
dyfer, line 69: ~~It seems we need to rebase anyway~~ - could you also update this to the new syntax?

capital-G: IIRC there was also some compilation error on qt5 - I could guard this w/ the Qt5 ifdef, then we will have a smoother transition from Qt5->Qt6?

capital-G:
> @capital-G I think we only needed one more revert to the old style for Qt5 and then it builds: [dyfer@cdd5b16](https://github.com/dyfer/supercollider/commit/cdd5b1613fd28cb521c6ae5d48f840a5f8d908a2)

Thanks for giving a push here - I already gave up after I saw compilation errors popping up one after another instead of all at once^^

editors/sc-ide/widgets/settings/sclang_page.cpp
dyfer: Ah I see...
If you don't mind, I think an update + keeping the old style for qt5 would be great!

dyfer: BTW I wanted to check this locally and since I wasn't sure where to look I asked an LLM - it suggested defining the overload like so:

```cpp
connect(ui->sclang_port, QOverload<int>::of(&QSpinBox::valueChanged), this, &SclangPage::showLanguageRestartDialogOnSave);
```
This seems to build and run locally with qt5.
I'm not sure if this is the correct way to go and/or if we should do this in other places where objects have multiple overloads? Do you know whether this is an idiomatic solution?

capital-G: Yeah, I saw this in Qt documentation but I found the code so *ugly* that I didn't want to go down this path^^
Ifdef looks cleaner for me b/c once we drop Qt5 we only have sleek connections ;)

dyfer: Okay, great! Yeah, it doesn't look particularly pleasing... let's stick with an ifdef

capital-G:
If the pipeline goes green it should be good and you can go through all the lines marcin ;) hf

dyfer:
> you can go through all the lines marcin

That'll have to wait :)... but I hope to get to it eventually!

dyfer (approved):
Thanks! 

I went through the changes as well as I could, I think it all checks out. I also ran Document tests locally for good measure (they are not run in the CI since we're not running the IDE).

Feel free to merge!

capital-G:
Thanks for double-checking marcin, much appreciated!

PR #7285, opened 2025-12-12
Update portmidi library

dyfer:
## Purpose and Motivation

This PR replaces older `portmidi` code with a current submodule (the tip of the master branch which happened to be tagged as [v2.0.8](https://github.com/PortMidi/portmidi/commits/v2.0.8), AFAICT). This greatly reduces MSVC warnings produced by portmidi.

After looking at MSVC warnings generated by 3rd party libraries, I thought I'd try to see if updating some of them would be a way to remove the warnings. 

This may need a general discussion whether we want to replace older 3rd party libraries with updated code and proper git submodules. I'm not sure what would be the concerns though?

The submodule is in a sub-subdirectory, to cleanly re-set build flags (similarly to how we handle portaudio).

Note that we use portmidi on Windows only. I don't believe our test suite tests MIDI functionality on Windows.
I minimally checked a build from this branch to confirm that MIDI still works on an actual Windows 10 machine (note in, note out, controller in etc), but I didn't do any stress/reliability/latency testing or comparison.

## Types of changes

- Maintenance

## To-do list

- [x] Code is tested (minimally)
- [x] All tests are passing
- [ ] Updated documentation


capital-G:
> This may need a general discussion whether we want to replace older 3rd party libraries with updated code and proper git submodules. I'm not sure what would be the concerns though?

I also thought about this and I think the only concern is that you want to build a 10 year old version of SuperCollider you probably have better luck if everything is self-contained instead of depending on another resource. But this is the only reason I can think of as long as we require the whole lib and do not modify it.

On the other side: Our repo becomes smaller and also updating becomes easier.

Checking the changelog (upwards of https://github.com/PortMidi/portmidi/blob/101dac9455e2718512c94e24cbcae6a6f34b908b/CHANGELOG.txt#L31) doesn't seem to reveal any concerning breaking changes.
https://github.com/PortMidi/portmidi/releases/tag/v2.0.1 states why there was a major version bump

> PortMidi has a new extended API with Pm_CreateVirtualInput(), Pm_CreateVirtualOutput(), and Pm_DeleteVirtualDevice() so we are calling this major version 2.

I think this is good-to-be-merged - but maybe someone from the people w/ a longer C++ background can chime in here and state their experience? @scztt ? :)
We could also add the dev-meeting flag, such that it will be resolved at the latest by then?

capital-G (approved):

dyfer:
> I also thought about this and I think the only concern is that you want to build a 10 year old version of SuperCollider you probably have better luck if everything is self-contained instead of depending on another resource.

Yes, I guess that's a consideration, but I think this is what the release source tarballs are for - they include all the code, including submodules, as it was at the time of the release.

PR #7289, opened 2025-12-16
Specify sclang port within IDE settings

capital-G:
value gets stored in the `sc_ide_conf.yaml` and defaults to 57120 if not present in the yaml file

## Purpose and Motivation

Allows to specify which port sclang should use and stores this value in the `scide_ide_conf.yaml`

image

Sadly I am not the best in Qt Designer, so this looks quite off.... I decided to make it right-bound for some reason.

This came up in https://scsynth.org/t/default-sclang-port-change/12715

## Types of changes

- New feature

## To-do list

Please test b/c it works for me outside of `/Application`, but once I move `SuperCollider.app` into `/Application` sc-ide crashes - but this is also my current behavior on `develop` branch - I may have to clear my build dir...

### DIFF
--- a/editors/sc-ide/widgets/settings/sclang_page.cpp
+++ b/editors/sc-ide/widgets/settings/sclang_page.cpp
@@ -64,6 +64,8 @@ SclangPage::SclangPage(QWidget* parent): QWidget(parent), ui(new Ui::SclangConfi
 
     connect(ui->sclang_post_inline_warnings, SIGNAL(stateChanged(int)), this, SLOT(markSclangConfigDirty()));
     connect(ui->sclang_exclude_default_paths, SIGNAL(stateChanged(int)), this, SLOT(markSclangConfigDirty()));
+
+    connect(ui->sclang_port, SIGNAL(valueChanged(int)), this, SLOT(markSclangConfigDirty()));
 }
 
 SclangPage::~SclangPage() { delete ui; }
--- a/editors/sc-ide/widgets/settings/sclang_page.cpp
+++ b/editors/sc-ide/widgets/settings/sclang_page.cpp
@@ -73,6 +73,11 @@ void SclangPage::load(Manager* s) {
 
     ui->autoStart->setChecked(s->value("autoStart").toBool());
     ui->runtimeDir->setText(s->value("runtimeDir").toString());
+    if (s->value("sclangPort").isNull()) {
+        ui->sclang_port->setValue(57120);
+    } else {
+        ui->sclang_port->setValue(s->value("sclangPort").toInt());
+    }
 
     QStringList availConfigFiles = availableLanguageConfigFiles();
     QString configSelectedLanguageConfigFile = s->value("configFile").toString();
@@ -96,6 +101,7 @@ void SclangPage::store(Manager* s) {
     s->beginGroup("IDE/interpreter");
     s->setValue("autoStart", ui->autoStart->isChecked());
     s->setValue("runtimeDir", ui->runtimeDir->text());
+    s->setValue("sclangPort", ui->sclang_port->value());
     s->setValue("configFile", ui->activeConfigFileComboBox->currentText());
     s->endGroup();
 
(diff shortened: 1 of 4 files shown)


dyfer:
Thanks so much! This is great.

However, I think we should also trigger a language reboot dialog when we change the port, similarly to changing the config file. The logic for that may be a bit cumbersome, looking at how the config file dialog is triggered now - we obviously don't want to trigger two dialogs if both the config file and the port are changed...

I think the dialog text could stay the same as it is now.

EDIT the GHA build works fine for me whether it's copied to `/Applications` or not

capital-G:
> I think the dialog text could stay the same as it is now.

~~Well, it needs to be a different dialog b/c we do not need just a re-compile but also a restart. But I think I can also add this, using lambda functions can reduce the code footprint.~~ => We indeed restart the language and not just recompile.
Will do :)

> EDIT the GHA build works fine for me whether it's copied to `/Applications` or not

Thanks for this info :)

editors/sc-ide/widgets/settings/sclang_page.cpp
```diff
@@ -64,6 +64,8 @@ SclangPage::SclangPage(QWidget* parent): QWidget(parent), ui(new Ui::SclangConfi
 
     connect(ui->sclang_post_inline_warnings, SIGNAL(stateChanged(int)), this, SLOT(markSclangConfigDirty()));
     connect(ui->sclang_exclude_default_paths, SIGNAL(stateChanged(int)), this, SLOT(markSclangConfigDirty()));
+
+    connect(ui->sclang_port, SIGNAL(valueChanged(int)), this, SLOT(markSclangConfigDirty()));
```
dyfer: Sorry, I didn't fully follow the logic - would this trigger writing to the config file also, even though no changes were made?

capital-G: Only the moment you touch/modify the box. If you bump it up and then down to its original value it will therefore still count as changed.

dyfer: Yeah, but what I'm saying is that it would be best not to overwrite the sclang.conf if we don't actually change _that_.

capital-G: Why not? When we open the dialog the settings get parsed and when we close it we write the state of the UI back into the file, so this seems fine to me? Also we do not trigger the dialog twice.

dyfer: I was specifically talking about sclang.conf - that only has include/exclude paths etc. IIUC.
The port number is written in the ide conf file (right?) and yes, that one should be overwritten. 
But I didn't check the logic closely whether both files get overwritten at the same time or only when relevant changes happen?

telephon:
What happens when the port is not available?

editors/sc-ide/widgets/settings/sclang_page.cpp
capital-G: now I see what you are talking about - I indeed rewrote the sclang_conf file, but I don't think that this is a big problem in the bigger scope of the IDE ;)

I adjusted the PR now such that changing the port number will trigger the "restart language?" dialog but not write the sclang_conf file.
I did some code maintenance which makes the things a bit clearer.

Once we have the new connect syntax when dropping Qt6 we can also use lambda functions which should reduce the necessary methods.

While I were at it - I prefixed the member variables with `m`.

capital-G:
> What happens when the port is not available?

The PR only makes the port explicit when starting sclang via IDE - it does not check if the port is free or in use, this is sclang behavior and not part of this PR.

edit: actually unsure about the current behavior regarding blocked ports
<details>
Yet, to write it down: sclang currently tries out the next 10 ports implicitly IIRC until one port is available - I personally don't really like this behavior since I expect that

```
sclang --port 57120
```

is afterwards available via `57120` and not `57121`.
This should be handled via exit codes and the IDE/invoker, not by the language process itself.
</details>

editors/sc-ide/widgets/settings/sclang_page.cpp
capital-G: I tested it but please verify if you also think that the behavior is as expected :)

capital-G:
This PR took a bit longer than expected, but beyond GUI design I think we now have a somehow good implementation and also cleared up some other stuff along the way.

dyfer (approved):
Thanks! 
Works as advertised. Thanks for cleaning up the logic!

Feel free to merge!

dyfer:
IMO sclang port choosing behavior is something to look into separately from this PR.

PR #7291, opened 2025-12-17
Interpreter.executeFile, String.load now take args and kwargs.

HotwheelsSisyphus:
## Purpose and Motivation
Before this PR, String.load didn't take arguments, but Interpreter.executeFile, to which it forwards, did; and neither took kwargs. This fixes that.
This came up in a discussion and seemed like it would be easy enough to do. 

## Types of changes

- Documentation
- Tests added
- New feature
- Not a breaking change

## To-do list

- remaining question: Is there some guideline for where exactly within the "assets" folder in the testsuite/classlib the individual assets should go? Currently I have:

```zsh
├── TestQuark
│   └── TestQuark.quark
└── TestString.scd
```
 The only other item currently in there is a Folder TestQuark, which sort of suggests that each UnitTest class gets its own folder for the assets; otoh, the testquark test does explicitly use that folder rather than the file that's in there. I can move the .scd to a TestString folder if reviewers think that's better.

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


JordanHendersonMusic:
Should line 884 of the help be `...args` instead? I can't remember exactly what scdoc and the style guide say here regarding the argument tag for variable args.

Otherwise this looks good, thanks!

HotwheelsSisyphus:
> Should line 884 of the help be `...args` instead? I can't remember exactly what scdoc and the style guide say here regarding the argument tag for variable args.

You're right, the scdoc syntax states that it should be `... args` (with a space), and I also overlooked the space in the class files (which doesn't seem to matter for sclang but does for scdoc).

Anyway, I now reshuffled the order of methods between loadPaths and loadRelative; they have the same signature so we can reuse the arg description, but there's an example for loadPaths that, if it's between loadPaths and loadRelative, makes it less than obvious that the arg descriptions apply to both methods. 
This will likely mess up the diffs, so here's a render of the result: 

**changed order of methods**
changed order of methods

**the new subsection**
the new subsection


I'm also noticing that the example is really not ideal, because it uses loadPaths to load just _one_ file... I'll try to improve that in a separate commit.

JordanHendersonMusic:
So when I did the kwargs I never changed scdoc to reflect this, nor did I change/update  any of the style guidelines.... This should be addressed at some point as kwargs usage grows.

For this pr, I *think* it would be a good idea to do `...kwargs` in the argument tag as well. I believe it renders correctly.

HotwheelsSisyphus:
yes, it does render correctly!

HotwheelsSisyphus:
but wait a minute: scdoc throws a warning 
```
Method -loadPaths has arg named 'kwargs', but doc has 'argument:: ... kwargs'.
```
But if I try to change this in the class library, the library won't compile (also isn't that simply wrong anyway?):
```
parse error:
  	load { |... args, ... kwargs|
  	etc.
 ```
 so I figure the scdoc warning should be suppressed for that case. I don't know how to do that and whether it's worth the effort... or if I we should just leave the arg tag without the ellipsis.

**Comment by JordanHendersonMusic** [comment]
> but wait a minute: scdoc throws a warning

Ah so that's a bug that needs fixing, or we could decide that the docs shouldnt show the ellipse for kwargs.

For now, I don't think it matters what you do, it is only a minor issue and when it is fixed, this will post a warning so we can find it later.

Is the rest of the pr stable now?

**Comment by HotwheelsSisyphus** [comment]
> Is the rest of the pr stable now?

it is now! the last few commits (i.e., yesterday and today) were just
a) adding a space in the class library (... args)
b) adding the ... to the arg tags in the docs
c) reshuffling the docs & and updating the example.

**Review by JordanHendersonMusic — COMMENTED** [review]
So there are some subtle issues here that are quite hard to address.

Consider a file as follows:
```supercollider
|pathName|  
pathName + 1 
```
and its use:
```supercollider
"path".load(pathName: 1)
```
What should happen here?

Essentially, all methods that have `...args, kwargs` should *not* have any other arguments, unless they intent to capture those arguments.

The signature of String:load meets this criteria, but its implementation of passing this along to Intrepreter:executeFile causes the argument 'pathName' to be caught and overriden (or attempted to, what it actually does is irrelevant because it isn't what is expected).

This becomes more problematic when we start using loadPaths and loadRelative as they all capture different arguments, loadRelative calls loadPaths. which calls executeFile, so the arguments caught there are: 'warn, action, and pathName'.

---

I think there are two solutions here:

* String:load and Interpreter:executeFile

`executeFile` could be re-implemented (under a new name) in terms of another method that doesn't take `pathname` as an argument, just uses the first variable argument. Then String:load and the old `executeFile` could call this new method.

* String:loadPaths and String:loadRelative

I think these methods shouldn't use the variable argument syntax at all, but take an array of arguments and an array of kwargs pairs (just remove the elipse). This isn't elegant, but seems like the only way to make these methods fully adhere to what we'd expect.

testsuite/classlibrary/TestString.sc
```diff
@@ -1,5 +1,21 @@
 TestString : UnitTest {
 
+    // ------ load (execute) a scd file ----------------------
+
+    // forwarding of args/kwargs to Interpreter.executeFile/compileFile
+	test_load_scd_file_w_args_and_kwargs {
+		var expected = "interior crocodile alligator got chevrolet theatre";
+        var path = PathName(thisMethod.filenameSymbol.asString).pathOnly +/+ "assets" +/+ "TestString.scd";
```
JordanHendersonMusic, line 8: Would be good to add a test that the file exists with a nice error message in case someone in the future moves it by accident.

HotwheelsSisyphus:
Right. Although this seems like a very exceptional case... 
Since this would involve an api change anyway (at least for loadPaths/loadRelative), maybe we should try to have similar signatures for all the String methods? I.e., have them all take args, kwargs arrays without varargs:
```
String.load(argsArray,kwargsArray)
String.loadPaths(warn, action, argsArray, kwargsArray)
String.loadRelative(warn, action, argsArray, kwargsArray)
```

--------------------------

as for forwarding executeFile to something else (say prExecuteFile), I currently don't understand how that helps. Maybe I misunderstood, e.g.
```
executeFile { | pathName ... args, kwargs |  
    this.prExecuteFile([pathName ++ args], kwargs)     
}
prExecuteFile { | ... args, kwargs |
   this.compileFile(args[0]).valueArgs(args[1..], kwargs)
   }
```
because there the outer executeFile will still capture a pathName keyword...

or did you mean something like this: 
```
executeFile  { | ... args, kwargs |
	var i = kwargs !?  _.indexOf(\pathName); 
	// i will be nil if there are no kwargs or if \pathName is not in them.
	var path;
	if(i.isNil { 
		path = args[0]; 
		args = args[1..] 
	} {
		path = kwargs[i+1]
	};
	this.compileFile(path).valueArgs(args, kwargs);
}
```
that's a bit contrived but would at least be backward compatible both with cases where executeFile is called with a pathName keyword argument, and with cases where it's the first positional arg...
but maybe I just misunderstood and there's a more straightforward solution.

JordanHendersonMusic:
If string:load calls prExexuteFile then it doesn't have the issue, previously it took no args so we get to define it completely, so it should be correct where possible. This way executeFile stays the same - we can't remove the argument name.

Making loadPaths and loadRelative be arrays seems fine. It doesn't seem likely you'd want to send the same arguments to all files anyway so probably won't be used.

HotwheelsSisyphus:
ah, I get it now. So should `executeFile` stay as is, i.e.
``` 
executeFile(pathName ... args)
```
or should I add the kwargs there?
```
executeFile(pathName ... args, kwargs)
```
(I favor the latter.)

JordanHendersonMusic:
Yes the latter, you can always add kwargs once there is already args.

testsuite/classlibrary/TestString.sc
HotwheelsSisyphus: did so, thanks!

JordanHendersonMusic (approved):
Looks good now! Annoying this ended up being harder than expected, but I think it's important not to have a surprising API here.

JordanHendersonMusic:
Would be good to get a second review on this one.

PR #7300, opened 2025-12-31
Post interpreter start command to post window

capital-G:
## Purpose and Motivation

Prints the command how `sclang` gets invoked from within the IDE to the post window.

```
Starting interpreter with "/Users/scheiba/github/supercollider/build/editors/sc-ide/Debug/SuperCollider.app/Contents/Resources/../MacOS/sclang -l /Users/scheiba/Library/Application Support/SuperCollider/sclang_conf.yaml -i scqt"
```

image

This is probably a good resource for debugging in case the interpreter does not start and came up during https://scsynth.org/t/strange-sclang-problem-in-scide/12781

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/editors/sc-ide/core/sc_process.cpp
+++ b/editors/sc-ide/core/sc_process.cpp
@@ -154,6 +154,12 @@ void ScProcess::startLanguage(void) {
     if (!workingDirectory.isEmpty())
         setWorkingDirectory(workingDirectory);
 
+    QString startMessageContent = sclangCommand;
+    for (auto& argument : sclangArguments) {
+        startMessageContent = startMessageContent + " " + argument;
+    };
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic (approved):

JordanHendersonMusic:
Good idea

nuss:
Just because my post about my recent problem with sclang quitting (resp., not coming up at all) immediately on startup was mentioned:

1. The problem wasn't about the config that I was using. The problem was caused by a path to a non-existing "Runtime Directory" (see screenshot)
2. The problem seems to have gone (already fixed?) in SC 3.15-dev. If an invalid "Runtime Directory" is given it simply gets ignored and the interpreter comes up normally.
Bildschirmfoto zu 2026-01-01 21-34-46

PR #7306, opened 2026-01-02
Improve discoverability of `|==|` by adding `==` in UGen.schelp

prko:
## Purpose and Motivation

The current documentation in `UGen.schelp` places the explanation of `==` within the entry for `|==|`. 
This arrangement presumes that readers are already familiar with both concepts, which makes the information 
less accessible. In practice, users who only know `==` may find it difficult to discover `|==|`, as the 
connection is not immediately obvious.

By adding a dedicated note under `==`, this pull request improves discoverability and clarity. It ensures 
that users searching for `==` are guided towards the correct usage of `|==|` when comparing signals, while 
retaining the distinction that `==` compares UGen objects themselves. This change aligns the documentation 
with common expectations, reduces potential confusion, and makes the learning process more straightforward 
for new users.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/UGen.schelp
+++ b/HelpSource/Classes/UGen.schelp
@@ -358,7 +358,7 @@ Outputs 1, representing true for negative, 0 representing false for positive or 
 
 method:: ==
 note::
-code::==:: compares UGen objects by identity.
+code::==:: compares UGen objects by equality.
 Use code::|==|:: to compare the signals they produce.
 ::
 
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/UGen.schelp
```diff
@@ -356,9 +356,18 @@ Outputs 1, representing true for positive, 0 representing false for negative or
 method::isNegative
 Outputs 1, representing true for negative, 0 representing false for positive or zero input values.
 
+method:: ==
+note::
+code::==:: compares UGen objects by identity.
```
telephon: "by equality" is more correct, because this is what `==` wants to say (in this case, it happens to be implemented as identity).

prko: Yes! I was confused. Sorry, and thank you for the correction.

telephon (approved):

PR #7307, opened 2026-01-02
sclang: nan tagged fix

### LINKED ISSUE #7296: sclang crashes on arm64 linux
## Environment

* SuperCollider version:3.15.0-dev [f0cbfda18]
* Operating system: Ubuntu24.04.3. ARM64 (via parallels)

## Steps to reproduce
1. build SC in termnal:
```shell
sudo apt-get install   build-essential   cmake   libjack-jackd2-dev   libsndfile1-dev   libfftw3-dev   libxt-dev   libavahi-client-dev   libudev-dev

sudo apt-get install   git   libasound2-dev   libicu-dev   libreadline6-dev   libudev-dev   pkg-config   libncurses5-dev   jackd2

sudo apt-get install   qt6-base-dev   qt6-base-dev-tools   qt6-tools-dev qt6-tools-dev-tools   qt6-declarative-dev   libqt6gui6   libqt6printsupport6   libqt6svgwidgets6   libqt6websockets6-dev   libqt6webenginecore6   libqt6webenginecore6-bin   qt6-webengine-dev   qt6-webengine-dev-tools   libqt6webchannel6-dev   libqt6opengl6-dev   libqt6svg6-dev   linguist-qt6   qt6-l10n-tools   libglx-dev libgl1-mesa-dev   libvulkan-dev   libxkbcommon-dev   libxcb-xkb-dev

git clone --recurse-submodules https://github.com/SuperCollider/SuperCollider.git

cd SuperCollider/

mkdir build && cd build

cmake ..

make

sudo make install

sudo ldconfig
```
2. Execute SC-IDE

### Expected
sclang should boot without error.

### actual
```
compiling class library...
Interpreter has quit. [Exit code: 11]
```

### LINKED ISSUE #7301: Make ctests print on failure

Should be as simple as adding `--rerun-failed --output-on-failure`. 

Otherwise it would be good to print the log on failure.

JordanHendersonMusic:
## Purpose and Motivation

Closes  #7301

Prints what went wrong in the ctest.

Closes #7296

Fixes crash in arm64 - I did not understand the pointer spec before.

Fixes casting issue for nans.

Tidys up some macros.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/PyrSlot.h
+++ b/lang/LangSource/PyrSlot.h
@@ -42,10 +42,11 @@ A PyrSlot is an 8-byte value which is either a double precision float or a
 #ifdef __SIZEOF_POINTER__
 #    define SIZEOF_POINTER __SIZEOF_POINTER__
 #else // MSVC doesn't define __SIZEOF_POINTER__
-#    ifdef _WIN32
-#        define SIZEOF_POINTER 4
-#    else
+#    ifdef _WIN64
 #        define SIZEOF_POINTER 8
+#    else // For some reason, 64 bit windows defines _WIN32 ... this code is very fragile as there is no cross platform
+          // way to do this.
+#        define SIZEOF_POINTER 4
 #    endif
 #endif
 
--- a/lang/LangSource/PyrSlot.h
+++ b/lang/LangSource/PyrSlot.h
@@ -91,12 +91,16 @@ inline void unreachable() {}
 // -Inf = 1111 1111 1111 0000 000000000000000000000000000000000000000000000000 = FFF0 0000 0000 0000
 
 // If exponents are all 1 and fraction isn't 0, its some type of nan (if fraction is zero, it is an inf)
-// sNaN = 0111 1111 1111 0000 000000000000000000000000000000000000000000000001 = 7FF0 0000 0000 0001
-// qNaN = 0111 1111 1111 1000 000000000000000000000000000000000000000000000001 = 7FF8 0000 0000 0001
-// qNaN = 1111 1111 1111 1000 000000000000000000000000000000000000000000000001 = FFF8 0000 0000 0001
-//        | sign is irrelevant to nan
-// qNaN = 1111 1111 1111 1111 111111111111111111111111111111111111111111111111 = FFFF FFFF FFFF FFFF
+// sNaN    = 0111 1111 1111 0000 000000000000000000000000000000000000000000000001 = 7FF0 0000 0000 0001
+// qNaN    = 0111 1111 1111 1000 000000000000000000000000000000000000000000000001 = 7FF8 0000 0000 0000
+// qNaN(1) = 0111 1111 1111 1000 000000000000000000000000000000000000000000000001 = 7FF8 0000 0000 0001
+// qNaN(2) = 0111 1111 1111 1000 000000000000000000000000000000000000000000000001 = 7FF8 0000 0000 0002
+// qNaN    = 1111 1111 1111 1000 000000000000000000000000000000000000000000000001 = FFF8 0000 0000 0001
+//           | sign is irrelevant to nan
+// qNaN    = 1111 1111 1111 1111 111111111111111111111111111111111111111111111111 = FFFF FFFF FFFF FFFF
 //                     still qNaN
+
+// Quiet nans can be produced through std::nan("0"), std::nan("1"). The only nan that can be stored in a slot is "0".
 
 
 // Signaling NaN (sNaN) is used to signal some floating point error has occurred.
@@ -124,7 +128,6 @@ inline void unreachable() {}
 // If the data is less than 48 bits, additional bits can be used to create further tags.
 
 namespace details {
-static constexpr uint64_t safeNaN = 0x7FF8000000000001;
 
 // cpp reference
 template <class To, class From>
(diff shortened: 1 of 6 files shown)


capital-G:
Thanks for tackling this!

I ran the tests on an arm 64 test runner and it displayed some further problems (see https://github.com/capital-G/supercollider/actions/runs/20660350455/job/59321973082 and https://github.com/capital-G/supercollider/actions/runs/20660350455/job/59327318865, but I think this is already way better now. I haven't checked if the failed tests are due to a faulty server or something more fundamental regarding byte codes etc.

How should we proceed here regarding review @JordanHendersonMusic ? Is it necessary that this gets reviewed properly by some person or should we pull it in as is?

JordanHendersonMusic:
Looks like it is a serial port test error. I don't know if that has anything to do with nan tagging, it could be doing something with bad nan values, unlikely though, and since we haven't been running the aarch64 tests it might have been broken before. 

Re review. This was a pretty simple misunderstanding on my part of how pointers work. No real need to review that bit.

If anyone knows how you can, in a cross platform way, know whether you are building on 32 or 64 bit in a macro that would be good - can't believe that's so hard...

I think I'm going to take this moment to add some more tests, so don't merge yet. I'll do this tomorrow morning.


I still think we should make it so when the ctests fail, the build step of the ci fails. I understand why that isn't desirable for the other tests, but ideally all these interpreter tests should be static_asserts, the only reason I can't do that is because of msvc (the usual culprit). Alternatively we could have two test phases... I don't know how to do that in cmake but believe it is possible.

capital-G:
> If anyone knows how you can, in a cross platform way, know whether you are building on 32 or 64 bit in a macro that would be good - can't believe that's so hard...

Maybe @timblechmann has a good idea for that?

> I still think we should make it so when the ctests fail, the build step of the ci fails.

I think this would be a good decision, but for this the tests should pass. I think we currently do not stop b/c the windows ctests fail.

JordanHendersonMusic:
Okay so I've made it so if the tests fail, the ci fails, but not on windows. This is ugly, but we really shouldn't ignore these tests. The mingw tests are completely broken, I don't know why. The interpreter and sclang tests all work on msvc (something is wrong with supernova there). 

Assuming the class library tests all pass, I think this is good.

lang/LangSource/PyrSlot.h
```diff
@@ -39,17 +39,25 @@ A PyrSlot is an 8-byte value which is either a double precision float or a
...
 #include "PyrSymbol.h"
 
-#if (__SIZEOF_POINTER__ == 8) || defined(__x86_64__) || defined(_M_X64) || defined(__LP64__) || defined(_WIN64)
-#    define POINTER_NEEDS_PADDING 0
+#ifdef __SIZEOF_POINTER__
+#    define SIZEOF_POINTER __SIZEOF_POINTER__
+#else // MSVC doesn't define __SIZEOF_POINTER__
+#    ifdef _WIN64
+#        define SIZEOF_POINTER 8
+#    else // For some reason, 64 bit windows defines _WIN32 ... this code is very fragile as there is no cross platform
```
Spacechild1: Yes, `_WIN32` is always defined and `_WIN64` is only defined for 64-bit targets (x64 or arm64). But yes, this is a bit brittle because the absence of `__SIZEOF_POINTER__` does not really imply that we are compiling with MSVC.

At the very least, you should do the following:
```
#if defined(__SIZEOF_POINTER__)
#    define SIZEOF_POINTER __SIZEOF_POINTER__
#elif defined(_WIN32)
#    ifdef _WIN64
#        define SIZEOF_POINTER 8
#    else
#        define SIZEOF_POINTER 4
#else
#    error "unsupported compiler"
#endif
``` 

Here's a more exhaustive check:
```
#ifdef __SIZEOF_POINTER__
#    define SIZEOF_POINTER __SIZEOF_POINTER__
#elif defined(__x86_64__) || defined(__aarch64__) || defined(_WIN64)
#    define SIZEOF_POINTER 8
#else
#    define SIZEOF_POINTER 4
#endif

// verify the pointer size
static_assert(SIZEOF_POINTER == sizeof(void*), "unexpected pointer size");
```

JordanHendersonMusic:
Okay should be good. Sorry about the complete mess of the commit history, I found the only way to test all compilers and archs is to use the CI, which runs really slow on my github account — still slow here, but better.

lang/LangSource/PyrSlot.h
```diff
@@ -190,35 +198,17 @@ template <typename T> struct MaybePadPointerTo64Bits {
...
+            // There is no need to mask when the pointer is padded as the mask doesn't touch the data.
             return ptr;
         } else {
             const auto r = reinterpret_cast<uintptr_t>(ptr);
-            if (r & (1ULL << 47)) {
-                return reinterpret_cast<T>(r | (~Masks::pointer));
-            } else {
-                return reinterpret_cast<T>(r & Masks::pointer);
-            }
+            return reinterpret_cast<T>(r & Masks::pointer);
```
Spacechild1, line 239: This looks all good so far!

But why did you change that line? AFAICT the previous version was correct because it tried to maintain the canonical form through sign extension of bit 47 (https://en.wikipedia.org/wiki/X86-64#Canonical_form_addresses).

(At least on Linux, bit 47 should never be 1 because that address range is reserved for the kernel. In that case, checking bit 47 is not necessary and we can always set the top 16 bits to 0. Not sure about Windows and macOS, though...)

JordanHendersonMusic: I thought it was correct too... But it didn't work on aarch64 (Mac). That bit was set but not the rest.

I'm assuming we will never get one of these kernel pointers on any operating system now - I think that's reasonable.. it seems to work.

Spacechild1: >  But it didn't work on aarch64 (Mac). That bit was set but not the rest.

Aha! aarch64 apparently uses a different addressing scheme where bits 48-63 are either 0 (= user space) or 1 (= kernel space), independent of the value of bit 47! Here's a nice write-up: https://muxup.com/2023q4/storing-data-in-pointers

This means that on both x86-64 and aarch64 the topmost 16 bits will always be zero in user space and we can just mask them off. The old code was actually wrong on aarch64 (because bit 47 can actually be set in user space) and the new code is correct.

Can you maybe add a comment about canonical addresses on x86-64 and aarch64 somewhere?

JordanHendersonMusic: That is a nice article, thanks!

Okay will do.

JordanHendersonMusic: Alright, I've expanded the comments and added some links. Also I suggested a way for us to support kernel space pointers if we really need to in the future.

Spacechild1: Thanks for the comments, these are great!

Spacechild1 (approved):

prko:
Thanks!

Issue #7313, opened 2026-01-08
Regex captures beyond end of line character $


capital-G:
## Environment

* SuperCollider version: 3.15.0-dev (Built from branch 'ide-port-gui' [059f1e19c8])
* Operating system: osx 15.6.1 (AArch64)
* Other details
	* Qt version: 6.9.3
	* argv: 
	* server: scsynth
	* includes: JITLibExtensions, Influx, adclib, miSCellaneous_lib, classlibrary

## Steps to reproduce

```supercollider
(
~code = "
some; // test
should still appear;
";

// regex matches newline $ but continues consumption of regex
~code.replaceRegexp("\/\/.*$", "").postln;
-> 
some; 
)
```

## Expected vs. actual behavior

It is apparent that regex is in multi-line mode `m` since otherwise it wouldn't capture `//` - but it shouldn't capture beyond the end of line as the next line does not match the regex.

<img width="391" height="306" alt="Image" src="https://github.com/user-attachments/assets/62363b25-db88-4177-a789-370d4e172bc6" />

prko:
On my end, this seems to work:
```
~code.replaceRegexp("\/\/[^\n]*", "").postln;
```
However, it would be great if \/\/.*$ also worked, as I remember struggling with this before. I thought it might be due to my incomplete understanding of regular expressions.

jamshark70:
> but it shouldn't capture beyond the end of line as the next line does not match the regex.

It looks to me like SC's regexp interprets `$` as end of string rather than end of line.

I don't think this is entirely unreasonable. `$` as EOL makes sense when the text is expected to span multiple lines (text editors or files). In a text editor, documents are fewer and longer; in code, strings are numerous and shorter. Clearly SC's regexp biased toward a larger number of shorter strings.

The question is what to do about it. Switching away from multiline mode would break compatibility with existing code, so I think it would be risky to treat this as a plain-and-simple bug and just change it. Another approach might be to add a flag somewhere, so you can decide behavior you want. (Since it's unknown from outside whether a string represents a single line or a whole document, it would be reasonable to let the user choose.)

Or, handle it by splitting the string into lines before searching.

```
(
~code = "
some; // test
should still appear;
";

// regex matches newline $ but continues consumption of regex
~code
.split($\n)
.collect(_.replaceRegexp("\/\/.*$", ""))
.join($\n)
.postln;
)
```

... which could be packaged into a method or a helper function.

prko:
More tests:

```
(
[...]

PR #7311, opened 2026-01-07
Add Buffer Playback Guide

prko:
## Purpose and Motivation

This PR brings together several frequently asked topics related to `PlayBuf`, `BufRd`, and buffer bank creation.
With the exception of not using `PathName`, I have listed only the simplest approaches based on my own experience. I recognise, however, that other developers may consider some of these methods inadvisable.
Please let me know if there are any inefficient approaches that should be removed.

The table of contents of the documentation is as follows:
- Description
- Buffer Playback Basics
  - PlayBuf vs. BufRd
  - Buffer.read vs. Buffer.readChannel
- Preparing and Querying Buffers
- Strategies for Reverse Playback
  - Using PlayBuf
  - Using BufRd
- Handling Channel Count Mismatches
  - Strategy 1: Dynamic SynthDef Generation
    - Using PlayBuf
    - Using BufRd
  - Strategy 2: Using a Function Handling Synthdefs
    - Using PlayBuf
    - Using BufRd
  - Strategy 3: Stereo balancing with Buffer.readChannel and Balance2
    - Using PlayBuf
    - Using BufRd
  - Strategy 4: Downmixing Multichannel Buffers (Splay)
- Managing Buffer Banks
  - Create buffer arrays in a directory
    - Create buffer arrays with supported sound files in a directory
    - Create buffer arrays with specific sound file format in a directory
  - Create buffer bank in a directory and its subdirectories
    - Prepare Files
    - Create buffer arrays
    - Create buffer arrays using Recursive Loading (Hierarchy)
    - Create buffer arrays using Recursive Loading (without Hierarchy)
    - Recursive Loading into Dictionaries
    - Accessing Buffers: Array vs. IdentityDictionary

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [ ] This PR is ready for review

PR #7308, opened 2026-01-02
Add SC_as_Plugin.schelp guide + Help Home shortcut [Part 1 of 3, external audio integration]

prko:
Closes: #7298 when all three documents have been written.

## Purpose and Motivation
This is the first part of the documentation described in #7298.
I will add the other two guidance documents after receiving feedback on this PR.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #7315, opened 2026-01-10
Add examples of both greedy and non‑greedy regular expressions to the String section

prko:
## Purpose and Motivation
This PR adds examples of both greedy and non‑greedy regular expressions to the String section, as referenced in #7313.  
Including these examples will save users time and effort when working with regular expressions. I spent some time exploring this myself, and found @JordanHendersonMusic’s explanation particularly clear and helpful.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/String.schelp
+++ b/HelpSource/Classes/String.schelp
@@ -367,6 +367,8 @@ subsection:: Regular expressions
 
 The String class provides access to the boost library's regular expression functions. Boost's default uses Perl settings. (Currently, there is no hook to override the regex style.) Syntax details may be found at link::https://www.boost.org/doc/libs/1_69_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html::.
 
+Regular expressions in SuperCollider use greedy matching by default. This means quantifiers like code::*:: and code::+:: will match as much text as possible while still allowing the overall pattern to succeed. This can lead to unexpected results when working with multi-line strings. See link::#Greedy_vs_Non-Greedy::.
+
 Note carefully the argument order:
 
 List::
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
Looks good, my only question is whether this is in the correct place, and whether we need a whole new file for regexs. There is a lot of information here already.

For now I'm happy to leave it unless others think it is necessary?

prko:
Considering the structure of the documentation, it is in the wrong place. We may need additional regular expression documentation for SC, since the current String documentation is extremely long.

From the perspective of user accessibility and convenience, however, the current location is not bad. It is practical, as users who refer to the main paragraph of the Regular Expression section can easily find the example. The explanation of greedy matching explicitly directs them to the replaceRegexp example, while other methods also point back to the main Regular Expression description. In this way, even when users look up methods other than replaceRegexp, they are guided to the same central example and explanation.

If the current official documentation contained sufficient information, issue #7313 would not have been written.

JordanHendersonMusic:
Sorry @prko I can't really follow what you've written.

Do you agree with me that this would be better elsewhere, in a dedicated regex file? And do you want to/can you be bothered right now to change it? If not, I'll merge as this looks good.

prko:
>Do you agree with me that this would be better elsewhere, in a dedicated regex file? 

Yes, I do.

>And do you want to/can you be bothered right now to change it?

I would create a new schelp file, but unfortunately I am unable to do so at the moment.
I am currently revising issue #7311. Drafting the initial version did not take very long, but the revision process is proving rather time‑consuming. Once that is complete, I must also finish issue #7308.

If I were to take this on, it would likely be several weeks before I could do so.

JordanHendersonMusic (approved):

HotwheelsSisyphus:
I'm not sure if this should go in the String class reference, to be honest... I'm also more in favor of a separate **tutorial** file. (I think that's something we haven't really discussed enough, but it would be good if there were a clearer distinction between class references and tutorials. Readers want different things from these two kinds of help files; I find the class reference is often overloaded with examples. I'm guilty of adding to that myself, though).  As for **reference**, the link to the the boost reference is really all we need, I think. 

at any rate, please don't use the table:: tag just to make a box around things. Imho it would be better to just make a separate box:: tag, because then it's clear what the intention is (to make a box around things). I can look into that.

prko:
The `table::` tag has been removed.
At present, we do not have a box:: tag. I believe that introducing new tags is not advisable, given that schelp2 is already being developed. The functionality of `box::` could, in my view, be implemented directly within schelp2. What are your thoughts on this, @capital‑G and @tedmoore?

prko:
@HotwheelsSisyphus 
I’ve removed the table you pointed out. When you have a moment, could you please review the changes again and consider approving this PR?

PR #7284, opened 2025-12-12
coc: no ai slop

JordanHendersonMusic:
Introduce a 'no ai slop' section to the code of conduct. This text is (slightly) adapted from luka's response on the forum, which is in turn a adaptation of @capital-G's proposal.

https://scsynth.org/t/add-no-ai-slop-to-community-guidelines/12672/18?u=jordan

@jamshark70 has suggested the following procedure:

1. This PR be reviewed and edited until a consensus is reached.
2. This then be shared over on the forum (and where ever else).
3. Only then should this be merged.

Please avoid turning this into a discussion on AI in general (perhaps open a new issue), this is about halting the proliferation of 'low effort'/'ai slop' throughout the community.

PR #7319, opened 2026-01-11
Introduce rich text editor

capital-G:
**DISCLAIMER**: This PR was vibe coded! This was my first try to figure out how vibe-coding is like. It took around 1 hour in total (with 3 sessions b/c I was hitting the quota of each 5hr session easily)  to guide and generate using claude code within CLion. Afterwards I looked over every LOC and tried to understand the change which took another 4-5 hours (I also changed/re-factored/tweaked some stuff while at it), so consider this reviewed by me, but this still needs another pair of eyes - especially the dynamic casting and if we can not incooperate this better into the type level, but I am not deep enough into the editor and C++ to see the possibilities.
I wanted to use this feature as a test balloon to see what models are capable of these days (I must say, I am really impressed).
I had this idea for a longer time and looked into this but stepped away from it b/c the editor only separates between highlighted and non-highlighted code through a binary flag, which appears in all kind of situations, which needed to be turned into an enum and which created all sorts of type problems as well. I think I could have written this by myself in around 3-4 days of work, so I saved some time using an LLM.
I think this could be a good test candidate for #7284 - I am also fine if this is considered AI slop and we do not merge it, but I think w/ a disclaimer and since I reviewed every LOC I wouldn't consider it slop. I hope this doesn't break the flood gates, but I think it is better to understand/evaluate/test how these models can be applied rather than having a total stance against them - they probably will not go away...
Since the SC codebase has probably been scraped by these models already, I think it was fine to let an LLM operate on it.

## Purpose and Motivation

This (re-)introduces a new file format `scr` and a rich-text editor for this file format which allows to format the font-family and some font-attributes (bold/italic/underline/strike/color) of a selected text region.

grafik

The editing toolbar at the top is only visible when a `scr` document is opened.
The normal syntax highlighting can be applied by selecting the code and pressing `cmd shift c` (or through a menu action).
The code needs to be indented manually though! Also, automatic region evaluation is not existing yet - one has to select the region manually, normal evaluation only evaluates the line of the cursor.

The file-format is based on Qts rich text format which is based on HTML, so a `scr` file can also be viewed in the browser.

grafik

A quick view at `head test.scr`

```html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN" "http://www.w3.org/TR/REC-html40/strict.dtd">
<html><head><meta name="qrichtext" content="1" /><meta charset="utf-8" /><style type="text/css">
p, li { white-space: pre-wrap; }
hr { height: 1px; border-width: 0; }
li.unchecked::marker { content: "\2610"; }
li.checked::marker { content: "\2612"; }
</style></head><body style=" font-family:'EB Garamond'; font-size:13pt; font-weight:400; font-style:normal;">
<p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;"><span style=" font-size:110pt; font-weight:700; color:#f20404;">Hello</span></p>
<p style=" margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;"><span style=" font-family:'Hack Nerd Font Mono'; font-size:18pt; text-decoration: underline;">we can have now</span></p>
<p style="-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px; font-family:'Hack Nerd Font Mono'; font-size:18pt;"><br /></p>
```

Since formatting and a themed editor are a opposite features your mileage may vary when opening an `scr` file in the browser or an editor with a different theme.
Hence, I would not include `scr` files into a repo as it makes reviewing hard and b/c of the themed IDE this can create all kind of problems

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/cmake_modules/MacOSXBundleInfo.plist.in
+++ b/cmake_modules/MacOSXBundleInfo.plist.in
@@ -124,6 +124,24 @@
 			<key>LSTypeIsPackage</key>
 			<false/>
 		</dict>
+		<dict>
+			<key>CFBundleTypeExtensions</key>
+			<array>
+				<string>scr</string>
+			</array>
+			<key>CFBundleTypeIconFile</key>
+			<string>scd</string>
+			<key>CFBundleTypeName</key>
+			<string>SuperCollider Rich Text</string>
+			<key>CFBundleTypeMIMETypes</key>
+			<array>
+				<string>text/html</string>
+			</array>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>LSTypeIsPackage</key>
+			<false/>
+		</dict>
 	</array>
 	<key>CFBundleExecutable</key>
 	<string>${MACOSX_BUNDLE_BUNDLE_NAME}</string>
@@ -196,6 +214,24 @@
 				<array>
 					<string>sc</string>
 					<string>scd</string>
+				</array>
+			</dict>
+		</dict>
+		<dict>
+			<key>UTTypeConformsTo</key>
+			<array>
+				<string>public.text</string>
+				<string>public.html</string>
+			</array>
+			<key>UTTypeDescription</key>
+			<string>SuperCollider Rich Text</string>
+			<key>UTTypeIdentifier</key>
+			<string>org.supercollider.scr</string>
+			<key>UTTypeTagSpecification</key>
+			<dict>
+				<key>public.filename-extension</key>
+				<array>
+					<string>scr</string>
 				</array>
 			</dict>
 		</dict>
--- a/editors/sc-ide/CMakeLists.txt
+++ b/editors/sc-ide/CMakeLists.txt
@@ -92,6 +92,8 @@ set ( ide_moc_hdr
     widgets/code_editor/overlay.hpp
     widgets/code_editor/autocompleter.hpp
     widgets/code_editor/completion_menu.hpp
+    widgets/code_editor/rich_editor.hpp
+    widgets/code_editor/rich_toolbar.hpp
     widgets/settings/dialog.hpp
     widgets/settings/general_page.hpp
     widgets/settings/sclang_page.hpp
@@ -144,6 +146,8 @@ set ( ide_src
     widgets/code_editor/overlay.cpp
     widgets/code_editor/autocompleter.cpp
     widgets/code_editor/completion_menu.cpp
+    widgets/code_editor/rich_editor.cpp
+    widgets/code_editor/rich_toolbar.cpp
     widgets/settings/dialog.cpp
     widgets/settings/general_page.cpp
     widgets/settings/sclang_page.cpp
--- a/editors/sc-ide/core/doc_manager.cpp
+++ b/editors/sc-ide/core/doc_manager.cpp
@@ -38,8 +38,9 @@
 
 using namespace ScIDE;
 
-Document::Document(bool isPlainText, const QByteArray& id, const QString& title, const QString& text):
+Document::Document(DocumentType docType, const QByteArray& id, const QString& title, const QString& text):
     mId(id),
+    mDocType(docType),
     mDoc(new QTextDocument(text, this)),
     mTitle(title),
     mIndentWidth(4),
@@ -64,9 +65,12 @@ Document::Document(bool isPlainText, const QByteArray& id, const QString& title,
     if (mTitle.isEmpty())
         mTitle = tr("Untitled");
 
-    mDoc->setDocumentLayout(new QPlainTextDocumentLayout(mDoc));
+    // Rich text documents use QTextDocumentLayout (default), others use QPlainTextDocumentLayout
+    if (docType != RichText)
+        mDoc->setDocumentLayout(new QPlainTextDocumentLayout(mDoc));
 
-    if (!isPlainText)
+    // Only SuperCollider documents get syntax highlighting
+    if (docType == SuperCollider)
         mHighlighter = new SyntaxHighlighter(mDoc);
 
     connect(Main::instance(), &Main::applySettingsRequest, this, &Document::applySettings);
@@ -74,14 +78,17 @@ Document::Document(bool isPlainText, const QByteArray& id, const QString& title,
     applySettings(Main::settings());
 }
 
-void Document::setPlainText(bool set_plain_text) {
-    if (isPlainText() == set_plain_text)
+void Document::setDocumentType(DocumentType docType) {
+    if (mDocType == docType)
         return;
+
+    mDocType = docType;
 
     delete mHighlighter;
     mHighlighter = 0;
 
-    if (!set_plain_text)
+    // Only SuperCollider documents get syntax highlighting
+    if (docType == SuperCollider)
         mHighlighter = new SyntaxHighlighter(mDoc);
 }
 
@@ -263,9 +270,9 @@ DocumentManager::DocumentManager(Main* main, Settings::Manager* settings):
     loadRecentDocuments(settings);
 }
 
-Document* DocumentManager::createDocument(bool isPlainText, const QByteArray& id, const QString& title,
+Document* DocumentManager::createDocument(Document::DocumentType docType, const QByteArray& id, const QString& title,
                                           const QString& text) {
-    Document* doc = new Document(isPlainText, id, title, text);
+    Document* doc = new Document(docType, id, title, text);
     mDocHash.insert(doc->id(), doc);
 
     QStandardItem* item = new QStandardItem(doc->title());
@@ -279,6 +286,14 @@ Document* DocumentManager::createDocument(bool isPlainText, const QByteArray& id
 
 void DocumentManager::create() {
     Document* doc = createDocument();
+
+    connect(doc->textDocument(), &QTextDocument::contentsChanged, doc, &Document::storeTmpFile);
+    syncLangDocument(doc);
+    Q_EMIT(opened(doc, 0, 0));
+}
+
+void DocumentManager::createRichDocument() {
+    Document* doc = createDocument(Document::RichText);
 
     connect(doc->textDocument(), &QTextDocument::contentsChanged, doc, &Document::storeTmpFile);
     syncLangDocument(doc);
@@ -331,11 +346,16 @@ Document* DocumentManager::open(const QString& path, int initialCursorPosition, 
 
     closeSingleUntitledIfUnmodified();
 
-    const bool fileIsPlainText = !(info.suffix() == QStringLiteral("sc") || (info.suffix() == QStringLiteral("scd"))
-                                   || (info.suffix() == QStringLiteral("schelp")));
+    auto docType = getDocumentTypeFromFileSuffix(info.suffix());
 
-    Document* doc = createDocument(fileIsPlainText, id);
-    doc->mDoc->setPlainText(decodeDocument(bytes));
+    Document* doc = createDocument(docType, id);
+
+    // Rich text documents are stored as HTML
+    if (docType == Document::RichText) {
+        doc->mDoc->setHtml(decodeDocument(bytes));
+    } else {
+        doc->mDoc->setPlainText(decodeDocument(bytes));
+    }
     doc->mDoc->setModified(false);
     doc->mFilePath = filePath;
     QString fileTitle = info.fileName();
@@ -406,7 +426,8 @@ void DocumentManager::restore() {
             MainWindow::instance()->showStatusMessage(tr("Cannot open file for reading: %1").arg(path));
         QByteArray bytes(file.readAll());
         file.close();
-        Document* doc = createDocument(false, QByteArray(), QFileInfo(path).baseName(), decodeDocument(bytes));
+        Document* doc =
+            createDocument(Document::SuperCollider, QByteArray(), QFileInfo(path).baseName(), decodeDocument(bytes));
         doc->mTmpFilePath = path;
         syncLangDocument(doc);
         Q_EMIT(opened(doc, 0, 0));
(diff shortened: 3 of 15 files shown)


prko:
Thank you very much for this impressive implementation. 

I tested it on Windows, and it works beautifully (pasting a copied section of a help document preserves the original formatting!). 
Inserting images is supported by dragging them (even from the help browser or web browser), but when copying text and an image from the help browser, the image is omitted when pasting.

I truly appreciate the effort that has gone into it.

I have just a small comment and a couple of questions:
- It seems that a new keyboard shortcut for single‑line evaluation might be helpful, as Shift + Enter does not currently work when using the rich text editor. Ctrl/Cmd + Enter seems to evaluate the single-line code.

Q1. Is this implementation connected with the new help system?

Q2. Do you plan to add support for style (paragraph style and font style), text alignment (left, centre, right, and justify), and vertical line spacing (leading)? I ask because Mathematica notebooks offer these features, and it would be wonderful if this environment could eventually support something similar.
Screenshot 2026-01-13 at 14 51 37

---

added:

I tested further, and I think:

- Additional toolboxes may not be necessary. If detailed editing is required, we can use other text editors such as Word, LibreOffice, and so on, since repasting text from another editor preserves its formatting in SC‑IDE’s rich text editor. If I remember correctly, the editor used before SC‑IDE worked in the same way: it did not offer toolboxes like the editors mentioned above, but the text format was preserved when copying and pasting.
- The quality of the results produced by the AI varies. In my experience, it often does not work as intended and frequently produces errors, although sometimes it is helpful. I would say that if the result does reflect the author’s intention, it should not be disrespected, because the author must undoubtedly have worked hard to realise it. Therefore, it ought to be accepted by the community.

capital-G:
Thanks for the feedback prko, and also confirming that this works under windows!

I may have to check regarding images - we should probably enforce relative paths on those, otherwise there is no way that this will work.
We could also take a look into embedding them as blob, but the PR is already quite heavy as-is, so I'd defer this to another PR.

> * It seems that a new keyboard shortcut for single‑line evaluation might be helpful, as Shift + Enter does not currently work when using the rich text editor. Ctrl/Cmd + Enter seems to evaluate the single-line code.

I don't know what shift + enter does, I only use cmd + enter, but we could probably add support for ctrl + enter as well?

> Q1. Is this implementation connected with the new help system?

Nope. SCDocs 2.0 is intended to be markdown based to be serializable, RTF is too "rich" and unstructured for that ;)

> Q2. Do you plan to add support for style (paragraph style and font style), text alignment (left, centre, right, and justify), and vertical line spacing (leading)? I ask because Mathematica notebooks offer these features, and it would be wonderful if this environment could eventually support something similar.

I think font style is supported? Alignment and vertical line height could be added through PRs.
I don't want to include structures such as headlines etc, as rich text is not a format which is targeted at being serializable, that what markdown is for.

What I do would like to add is to make the whole document scriptable from within sclang - I think this could result in some neat meta-performances or new text-interfaces (e.g. step sequencer where active step is bold).

> * Additional toolboxes may not be necessary. If detailed editing is required, we can use other text editors such as Word, LibreOffice, and so on, since repasting text from another editor preserves its formatting in SC‑IDE’s rich text editor.

That is good to know, really nice that this works!

> * The quality of the results produced by the AI varies. In my experience, it often does not work as intended and frequently produces errors, although sometimes it is helpful. I would say that if the result does reflect the author’s intention, it should not be disrespected, because the author must undoubtedly have worked hard to realise it. Therefore, it ought to be accepted by the community.

From a quality POV I was really content with this, but it still doesn't make reviewing easier and also feels a lot of joy is taken away - it doesn't feel like I accomplished something here.

telephon:
> I don't know what shift + enter does,

Shift enter evaluates the current line or selection only.

prko:
> I don't know what shift + enter does, I only use cmd + enter, but we could probably add support for ctrl + enter as well?

At present, the behaviour of Ctrl+Enter in the rich text editor does not appear to be the same as in the SCD text editor. The rich text editor seems not to parse parentheses in the way the SCD editor does.

---

I believe Shift+Enter cannot be used to evaluate code, since in the rich text editor it already serves its own function, namely to insert a line break without starting a new paragraph.

Spacechild1:
Cool!

> This PR was vibe coded! ... Afterwards I looked over every LOC and tried to understand the change which took another 4-5 hours (I also changed/re-factored/tweaked some stuff while at it), so consider this reviewed by me

"Vibe coding" means generating code without any review. What you did is typically called "AI-assisted coding" or "agentic coding". It is certainly not slop.

prko:
Just an optional thought: 
It might be helpful to provide a dedicated box for the code section (or vice versa for comment or text part).

In the screenshot below, the box was copied and pasted from the help home page’s table. 

I wasn’t able to copy and paste a table from an RTF or a DOCX file in Microsoft Word. Bulleted and numbered list items can be copied and pasted, although the indentation needs to be adjusted manually:

Screenshot 2026-02-11 at 18 54 35

telephon:
> It might be helpful to provide a dedicated box for the code section (or vice versa for comment or text part).

Why do you want this box? 
The motivation behind rich text for code (at least mine) is that code becomes more like ordinary writing. For me, this is desirable, because it lowers the boundary to cross from one to the other (in either way). Adding a box would seem to reinforce the boundary.

telephon:
The first time I tried, the fonts were not found. maybe it needed some restart or wait time. Now the fonts are found and the basic functionality seems to work fine.

It would be good if we could set and get the relevant properties from the `Document` class.
If you **select** and hit tab, autocolorisation and formatting could be activated.
Also, copy and paste style would be a useful shortcut (on macos this is alt-cmd-c and alt-cmd-v).

prko:
>Why do you want this box?

To make it easier to visually distinguish code from comments.
A dedicated box may not be necessary — this could also be achieved by:

- using a different background color, or
- using a different font style for code.

However, there may be issues with automatic code parsing.
In `.scd` files, parsing only works correctly when code and comments are properly structured from the very beginning of the file.

In the `.scr` format, using plain text without comment markers would make this impossible.
If the code section needs to remain untouched, the comment symbols could be rendered in a much smaller font size; size 4 is the smallest available.
Comments that appear on the same line as code, however, do not require their symbols to be reduced in size.
Screenshot 2026-02-13 at 08 23 46

telephon:
Ah I wouldn't have thought that we'd want an scr file interpreted as a whole. I am not sure if this is really needed. But indeed it is useful to have a more or less standard way to communicate that something is meant as code.

Currently, we are communicating what belongs together by parentheses. So perhaps a parentheis at the beginning of a line could signal that what comes after is code, until it closes again.

prko:
>Ah I wouldn't have thought that we'd want an scr file interpreted as a whole. I am not sure if this is really needed. 

You are absolutely right. There is no need to use an .scr file in order to interpret the code as a whole. The .scr files will only be required when preparing material for presentations or lectures.

telephon:
> The .scr files will only be required when preparing material for presentations or lectures.

At the time when supercollider still had rich text, it was used for all kinds of purposes, especially as a part of general artistic and compositional practice. So in particular it was the code that was formatted.

@redFrik wrote nice pieces that modified their own code formatting.

prko:
`scr` doesn’t seem to be a unique extension. On my system (macOS), it happens to be associated with GIMP by default. It’s not really a problem, but I thought it might be worth mentioning.

dyfer:
A couple thoughts on the file format:
- .scr seems to also be an extension of Windows Screen Saver files. How about we use something like `.scdr`?
- could exporting as "proper" html be an option? or is it already that and one only needs to change the extension?
- are there particular things you'd like feedback or review on @capital-G to move this forward? aside from @telephon's suggestions

prko:
>.scr seems to also be an extension of Windows Screen Saver files. How about we use something like `.scdr`?

`.scr` was the file extension for Windows screensavers in Windows 95 and later versions, but I’m not sure whether it is still commonly used today.  
Given modern conventions, I would choose `.scdx` rather than `.scr` or `.scdr`. Consider how other applications updated their formats:  
- Finale: `.mus` → `.musx`  
- Microsoft Word: `.doc` → `.docx`  
- Microsoft Excel: `.xls` → `.xlsx`  
- Microsoft PowerPoint: `.ppt` → `.pptx`

joshpar:
I’m pretty sure the “x” at the end of those extensions was because the formats are
Compressed xml. In some cases I know you can change the extension to .zip, decompress and view the xml.

dyfer:
Yeah, "x" in those formats is for XML. "r" (for Rich) seems appropriate here imo.
Eg. A counterexample to "x" is .rtf -> .rtfd (.... Directory)

prko:
SCDR seems to be the best!  Thanks!

JordanHendersonMusic:
So I do have some serious concerns about regions and the scd format. We do not have a real definition of what a region is.... Consider that it is really common for people to not put semicolons at the end of the line. 

This makes syntax highlighting, or any kind of tooling really hard! 

Regions and one line snippets can be added to the grammar... I'm working on rewriting it as we speak but will be getting to this bit later.

I don't think we should add another file format until we fix this....

Or, in the new format, get rid of regions and require all lines end in a semicolon for the time being until this is sorted. 


The worst thing possible would be if this behaviour differed from the original, then we could never solve this properly as we'd have to encompass both behaviours.

capital-G:
The thing that I am most concerned with is in regards to theme/styling - even major rich text editors such as google docs do not provide theming, b/c it becomes rather awkward what is "relative" and what is "absolute" styling in regards to color, which is absolutely necessary for us b/c of syntax highlighting. The html screenshot in the initial post demonstrates this problem and I can't think of a problem how this could be solved while still being version-independent/future-proof.
E.g. one way could be to add the metadata of the theme that this was written in, but then this format becomes versioned which is even worse (or we have to think of a migration path which involves engineering efforts and locks the format in an undesirable way IMO).
If this ever gets merged, I'd really advocate to make this a niche feature and not in any way a default way of writing sc b/c the documents become essentially not shareable since they are tied to a theme of the editor. Also rich text documents are not really suitable for git related applications.

I am a bit unsure about how to proceed here: there hasn't been much feedback on this feature and it introduces lots of problems/headaches - I'd also rewrite the code from hand b/c I think the AI written code here is rather verbose and can probably written more optimized/concise.

If someone is interested in this feature maybe the person could propose how the stated problems could be resolved, otherwise I have the tendency to close this PR b/c I don't know how to fix the resulting problems in a good way.

JordanHendersonMusic:
Could this be a markdown document instead? Doesn't that achieve similar results?

capital-G:
> Could this be a markdown document instead? Doesn't that achieve similar results?

Not really - markdown is not rich text. The current format is closer to a subset of html and allows to change color and font of the document - this is quite nice b/c one could pair/export it with a wasm build and one could do some really interesting things if one could script it like `Document`, but this would be another PR.

prko:
@capital-G

> If someone is interested in this feature maybe the person could propose how the stated problems could be resolved, otherwise I have the tendency to close this PR b/c I don't know how to fix the resulting problems in a good way.

To address the theming and styling issues mentioned above, introducing a dedicated block or box for SuperCollider code could be a viable solution. This might also help with the region-detection issues that @JordanHendersonMusic pointed out.

My idea is as follows:

---

### **For SuperCollider code areas (SC-CODE)**

The colours, fonts, font size, and paragraph properties should be strictly determined by the active SC-IDE theme.
When a user attempts to change any of these properties through the “font settings” or “paragraph settings” dialog, they must first check an option such as **“Unsynchronise from SC-IDE Theme.”**

Once this option is checked, SC-IDE would display a warning:

```text
By changing any of these items, SC-CODE will no longer be synchronised with the SC-IDE theme.
To resynchronise, please uncheck this option.

```

This keeps the default behaviour fully theme-synchronised and avoids locking the document format to a specific theme, while still allowing advanced users to override styling if they explicitly choose to.

---

### **For non-code areas (comments, general text)**

These parts do not need to follow SC-IDE theme rules.
They should remain relatively free and should **not** include an “Unsynchronise from SC-IDE Theme” option.

This way, we can maintain the seamless writing experience that @telephon advocated earlier, while keeping the actual code technically manageable.

If this still does not resolve the issues you mentioned, then apologies for the unhelpful idea — but I hope it provides at least one possible direction.

---

@JordanHendersonMusic

> So I do have some serious concerns about regions and the scd format. We do not have a real definition of what a region is... Consider that it is really common for people to not put semicolons at the end of the line.
> This makes syntax highlighting, or any kind of tooling really hard!

As someone who often uses semicolons only as delimiters, I didn’t fully realise how difficult syntax highlighting becomes when semicolons are optional. I tend to think of code without a semicolon as a single evaluation unit. For example, the following consists of two evaluation units:

```supercollider
(
a = 1;
b = 2;
c = a + b
)

c

```

However, many SC code snippets — including those in SCDoc `code::` tags — are not enclosed in parentheses, and semicolon usage is not consistent.

Would syntax highlighting become easier if the following were treated as delimiters for evaluation units?

* A line without a semicolon followed by a blank line
* A code snippet enclosed by parentheses without a semicolon at the end

---

### Added:

Recently, I noticed that some AI editors automatically create a code block when the user types three backticks(```):
```
```
Users can then type code or text inside that area.
It would be great if SCDR could provide a similar feature.

JordanHendersonMusic:
I have plans for region detect... I *think* I can make it a part of the grammar (bison has some cool error.recovery which might make this work).

It's only relevant in this pr if it adds a new definition of what a region is - which it should not do.

Issue #6145, opened 2023-11-30
Convert server plugin API to C


Spacechild1:
## Motivation

Currently, the server plugin API is actually a C++ API. This has the significant downside that you cannot write server plugins in other languages without at least some C++ glue code.

Since most languages can interface with C, a pure C API would make it easier to develop server plugins in other languages (C, Rust, Zig, Fortran, etc.)

There are two distinct, but related aspects:

1. languages that understand C headers (C, C++, Obj-C, D, Zig, etc.) can work directly with our header files

2. all other languages (e.g. Rust) can reimplement our plugin API by following the C ABI

## Plan for Implementation

The required changes necessarily break the ABI, so we need to bump the plugin API version and recompile existing plugins. However, we should try to maintain *source compatibility* as much as possible, i.e. existing C++ plugins should still compile.

In general, these are the required changes:

1. replace `bool` data members and function arguments with `int`
    Only requires recompilation.

4. replace reference functions parameters with pointers, e.g. `struct FifoMsg&` --> `struct FifoMsg *`
    This would obviously break C++ source compatibility.

    If we can guarantee that reference and pointer arguments are interchangable on the ABI level, we might do this conditionally for non-C++ code.

    Otherwise we would have to break source compatibility for all API functions with reference parameters.
    AFAICT, only the following functions would be affected:
    ```
    bool (*fSendMsgFromRT)(World* inWorld, struct FifoMsg& inMsg);
    bool (*fSendMsgToRT)(World* inWorld, struct FifoMsg& inMsg);
    struct scfft* (*fSCfftCreate)(size_t fullsize, size_t winsize, SCFFT_WindowFunction wintype, float* indata,
                                  float* outdata, SCFFT_Direction forward, SCFFT_Allocator& alloc);
    void (*fSCfftDestroy)(scfft* f, SCFFT_Allocator& alloc);
    bool (*fGetScopeBuffer)(World* inWorld, int index, int channels, int maxFrames, ScopeBufferHnd&);
    void (*fPushScopeBuffer)(World* inWorld, ScopeBufferHnd&, int frames);
    void (*fReleaseScopeBuffer)(World* inWorld, ScopeBufferHnd&);
    ```
    Assuming that most people actually use the corresponding macros, we may get away with changing their definition.
    For example:
    `#define SendMsgFromRT (*ft->fSendMsgFromRT)`
    would become
[...]

Issue #4438, opened 2019-06-06
SCFFT_Allocator not ABI compatible


Spacechild1:
The Supercollider plugin interface uses a class `SCFFT_Allocator` in the interface methods `fSCfftCreate` and `fSCfftDestroy`.

Unfortunately, `SCFFT_Allocator` is an abstract virtual base class and calling virtual methods across DLL boundaries should be avoided (unless you use COM techniques*). Different compilers can have different vtable layouts and your plugin may fail/crash in mysterious ways because the wrong virtual method is getting called. Practical example: recent GCC uses two (!) virtual methods for the virtual destructor while MSVC uses one, this means that subsequent virtual methods will be offset by 1.

The allocator should rather be a C struct like:

```
typedef void *(*AllocFunc)(void * user, size_t size);
typedef void (*AsyncStageFn)(void * user, void *ptr);

struct SCFFT_Allocator {
    void *mUser;
    AllocFunc mAlloc;
    FreeFunc mFree;
};

class SCWorld_Allocator : public SCFFT_Allocator {
public:
    SCWorld_Allocator(InterfaceTable* ft, World* world)
        // casting function pointers and calling them without casting back to the original type
        // is technically UB, but in this case it works because World * and void * have the same size
        : mUser((void *)world), mAlloc((AllocFunc)(ft->fRTAlloc)), mFree((FreeFunc)(ft->fRTFree)) {}
};
```

This could be done for the next time the plugin API version changes. UGens using `SCWorld_Allocator` just have to recompiled, but if they use their own subclass of `SCFFT_Allocator `(unlikely), they must change some code.

dyfer:
I came across a misterious [issue with PartConv](https://github.com/supercollider/supercollider/issues/4289) (it uses FFT). This goes beyond my understanding of C++ but I was wondering whether anyone has idea whether this could be related?

Spacechild1:
I don't think this is related. This issue only affects plugins compiled with a different compiler than scsynth/supernova, so it shouldn't affect core plugins.

dyfer:
Thank you for clarifying!

sonoro1234:
How is that we dont have crashes because of this?

Spacechild1:
[...]

PR #7329, opened 2026-01-20
Convert plugin API to C

Spacechild1:
## Purpose and Motivation

Convert the plugin API to pure C while maintaining full source compatibility.

The general idea is that all public structs and functions should follow the C ABI so that the plugin API can be reimplemented in other languages (most notably Rust and Zig).

NOTE: Supernova is not supported yet!

---

I have removed all C++-isms from all public structs and interface table functions. Some macros and many inline functions/methods are still C++ only; third-party developers are expected to translate these to their preferred language. (This includes C!)

I have added a new test (`testsuite/plugins/c_api_test.cpp`) that tries to build a simple plugin in pure C. If someone tries to add C++-isms, the test will fail.

---

Addresses https://github.com/supercollider/supercollider/issues/6145 and https://github.com/supercollider/supercollider/issues/4438

## Types of changes

- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


Spacechild1:
I know, the change set looks massive, but the changes are mostly trivial. Still, this is probably the largest of several upcoming plugin API changes (https://github.com/supercollider/supercollider/issues/6102).

Once this is ready and merged, I can work on the remaining parts, particularly https://github.com/supercollider/supercollider/issues/5347 and https://github.com/supercollider/supercollider/pull/6702.

Spacechild1:
This should now be ready!

I can successfully compile the built-in plugins, sc3 plugins and my own plugins (which also cover the more esoteric parts of the plugin API). I would say that source compatibility has been achieved 🎯 

I also ran some tests on the Server, in particular with the FFT UGens, since I had to change `SCFFT_Allocator` from a C++ abstract base class to a C struct. AFAICT everything still works as expected.

JordanHendersonMusic:
> Some macros and many inline functions/methods are still C++ only; third-party developers are expected to translate these to their preferred language. (This includes C!)

Would you mind just expanding on this? Does this mean these macros don't work in c right now?

Spacechild1:
> Would you mind just expanding on this? Does this mean these macros don't work in c right now?

Yes, some macros like `GETBUF` don't work because they assume a certain class layout that is not possible to achieve in C.

Also, there are quite a few structs/classes with inline member functions, like `sc_msg_iter`, `RGen`, `Complex`, `Polar`. Obviously, the member functions cannot be called from C, but the structs themselves have a pure C layout. This is important because the data itself is shared between the Server and the plugin.

Macros and inline (member) functions are utilities that do not affect the ABI. Personally, I would treat C the same as Rust and Zig in the sense that these utilities need to be translated to the respective language.

JordanHendersonMusic:
Ah I see, I was confused because we are using the same source code for the C headers (which are consumed by other languages) and the C++ headers (which are not)!

Might it be worth separating these out, so we have pure C source code, then the C++ includes some other file with its specific stuff in? That would mean existing plugins would have to include another header I guess...

Spacechild1:
> Might it be worth separating these out, so we have pure C source code, then the C++ includes some other file with its specific stuff in?

Yes, it would be great to separate them more cleanly. I haven't even tried because I didn't want to add too many changes. (The change set is already large enough.) I might give it a try once this PR is merged. Fortunately, we could also do it at a later point because it would not affect the API/ABI.

BTW, the separation is not exactly trivial because many structs also have (inline) member functions. For now, I just wrapped the member functions in `#ifdef __cplusplus` blocks. If we want pure C headers and also keep source compatibility with existing C++ plugins, we'd have to create pure C structs that are then inherited by their corresponding C++ structs. For example:

```c++
/*------------------ SC_C_API.h ----------------------*/

struct RGen_ {
    uint32 s1, s2, s3; // random generator state
};

#ifndef __cplusplus
typedef struct RGen_ RGen;
#endif

struct sc_msg_iter_ {
    const char* data;
    const char* rdpos;
    const char* endpos;
    const char* tags;
    int32 size;
    int32 count;
};

#ifndef __cplusplus
typedef struct sc_msg_iter_ sc_msg_iter;
#endif

// etc.

/*--------------- RGen.h ------------------*/

#include "SC_C_API.h"

struct RGen : RGen_ {
    void init(uint32 seed);
    // etc.
};

/*------------ sc_msg_iter.h --------------*/

#include "SC_C_API.h"

struct sc_msg_iter : sc_msg_iter_ {
    sc_msg_iter();
    sc_msg_iter(int inSize, const char* inData);
    // etc.
};

```

> That would mean existing plugins would have to include another header I guess...

Not really because the C++ headers would just include the C headers.

JordanHendersonMusic:
> we'd have to create pure C structs that are then inherited by their corresponding C++ structs. For example....

Oh that's quite nice, I didn't think of that annoyingly...

> Fortunately, we could also do it at a later point because it would not affect the API/ABI.

Okay sounds good. In that case this all looks pretty good at a glance. Are there any areas where you think we should pay more attention when reviewing?

Spacechild1:
> Are there any areas where you think we should pay more attention when reviewing?

The vast majority of the PR just consists of:
- adding `#ifdef __cplusplus` blocks
- replacing `inline` with the new `SC_INLINE` macro in public headers so they can be used by C plugins
- consistently use explicitly sized types (e.g. `int32` instead of `int`)
- typedef `struct World` to `World` and replace `struct World*` with `World*` whenever possible.
  (Most of the other structs already have typedefs; I have no idea why `World` did not.)
  
These are all trivial changes, but in combination they touch a lot of files...

---
  
The following three tasks have been a bit tricky because of the requirement to maintain source compatibility:
  1. come up with a typedef for `SCBool`:
    https://github.com/supercollider/supercollider/pull/7329/changes#diff-36c119617edca0acaeddd0c171fb5b4c95a964dc7972317fba72eb5591caf588R32-R39
  2. replace reference arguments with pointer arguments:
    https://github.com/supercollider/supercollider/pull/7329/changes#diff-14bc16a251f875fccb3f70df84b0fbb63cce760a42d35e43d6a5253cbcee8b91R165-R171
    https://github.com/supercollider/supercollider/pull/7329/changes#diff-14bc16a251f875fccb3f70df84b0fbb63cce760a42d35e43d6a5253cbcee8b91R230-R238
  3. convert `SCFFT_Allocator` from a pure abstract base class to a pure C struct:
    https://github.com/supercollider/supercollider/pull/7329/changes#diff-8b4b944b4c0f314b12757dabdc6b01f60d5a81d36aa9d195c3c279a6ba88f047L40-R47
    https://github.com/supercollider/supercollider/pull/7329/changes#diff-14bc16a251f875fccb3f70df84b0fbb63cce760a42d35e43d6a5253cbcee8b91R253-R264
    https://github.com/supercollider/supercollider/pull/7329/changes#diff-446966f442114620525b1d0e473a9fcdf5b1cbec6adef4f99b16002c4475a3bcR276-R280
    https://github.com/supercollider/supercollider/pull/7329/changes#diff-446966f442114620525b1d0e473a9fcdf5b1cbec6adef4f99b16002c4475a3bcL471-R472

I think the solutions above are adequate and they have been successfully tested, but if someone wants to take a deeper look, these are the spots with the non-trivial fixes.

Spacechild1:
Any more comments on this? Do you think I can merge this? @capital-G @dyfer? It would be great to have this as the basis for my other PRs to avoid lots of merge conflicts.

include/plugin_interface/sc_msg_iter.h
```diff
@@ -36,31 +36,36 @@ inline const char* OSCstrskip(const char* str) {
...
     elem64 slot;
     slot.i = ((int64)sc_ntohl(*(uint32*)inData) << 32) + (sc_ntohl(*(uint32*)(inData + 4)));
     return slot.f;
 }
 
 struct sc_msg_iter {
-    const char *data, *rdpos, *endpos, *tags;
-    int size, count;
+    const char* data;
+    const char* rdpos;
```
capital-G, line 62: Can we make this non-const? See

https://github.com/capital-G/DynGen/pull/47/changes#r2612027681

capital-G:
Can you add that the C-API example gets built in the CI? Maybe this could be implemented via ctest?

> Any more comments on this? Do you think I can merge this? @capital-G @dyfer? It would be great to have this as the basis for my other PRs to avoid lots of merge conflicts.

One thing I would like to ask - there is no way of maintaining binary-support with legacy plugins? We do lots of stuff regarding legacy support and plugins are the most important bit here - at least for me. OTOH: at least with the new API we have source compability and can extend it for future extesions w/o breaking binary support (until we go for 64  bits on the server^^).

Other than that: I can't wait to run my python django server in a real time thread via a scsynth server c-extensions ;) Really looking forward to that change!

I think its fine merging this to move forward here! :) Maybe Jordan can give it an additional review - b/c I don't know enough about C++/C to review this properly

edit: Oh - if you want - some doxygen docs would be really appreciated and would make the API so much easier to work with :)

JordanHendersonMusic:
> Can you add that the C-API example gets built in the CI? Maybe this could be implemented via ctest?

This was the only thing I was going to suggest. 

Code looks p pretty straight forward to me.

include/plugin_interface/sc_msg_iter.h
Spacechild1: I think you misunderstand my comment. It's the `char*` argument in `fDoAsynchronousCommand` that should be made `const`, not the other way round!

Spacechild1:
> One thing I would like to ask - there is no way of maintaining binary-support with legacy plugins?

No.

> OTOH: at least with the new API we have source compability and can extend it for future extesions w/o breaking binary support (until we go for 64 bits on the server^^).

We already have source compability. But you are right that the new API can be extended without breaking ABI compatibility.

> Can you add that the C-API example gets built in the CI? Maybe this could be implemented via ctest?

The example does get built on C-API: https://github.com/supercollider/supercollider/actions/runs/21226569416/job/61075094502?pr=7329#step:9:729

---

Side note: I just noticed that some of the server tests fail on Windows, but the CI passes nevertheless. Examples:
https://github.com/supercollider/supercollider/actions/runs/21226569416/job/61075094464?pr=7329#step:20:68
https://github.com/supercollider/supercollider/actions/runs/21226569416/job/61075094443?pr=7329#step:20:69
https://github.com/supercollider/supercollider/actions/runs/21226569416/job/61075094459?pr=7329#step:20:71
(With Msys2 *all* tests fail...)

Spacechild1:
> edit: Oh - if you want - some doxygen docs would be really appreciated and would make the API so much easier to work with :)

This is definitely on my roadmap!

Spacechild1:
Once someone approves this PR I'll go ahead and merge it.

dyfer:
> Side note: I just noticed that some of the server tests fail on Windows, but the CI passes nevertheless. Examples:

FIY some ctests fail on Windows, we haven't fixed them yet so for now the ctest step just continues despite failures...
https://github.com/supercollider/supercollider/pull/6896#issuecomment-2915677193

JordanHendersonMusic:
> Side note: I just noticed that some of the server tests fail on Windows, but the CI passes nevertheless. Examples:

Yes this has always been a problem. I've got a pr that makes the ci fail when the c tests fail, but I disabled this for windows as one of the targets we don't support, and the other windows build seems to have a problem with the testing program itself - either that or everything is broken...

Looks like the plugin is built, sorry for missing that! 

Going forward, do you think there is any value in actually doing some testing with the plugin? Making sure it can be loaded and produces the right output and whatnot? Could cover the whole API. This would be a different pr of course.

JordanHendersonMusic (approved):

Spacechild1:
> Going forward, do you think there is any value in actually doing some testing with the plugin?

I'm not sure. C and C++ plugins use the very same pure C structs and pure C functions. I can't really imagine how these would be compiled or called differently by a C and C++ compiler. After all, the very point of using the C ABI is to prevent these kind of incompatibilites.

The one important thing is to avoid (re)introducing C++-isms into the public structs and functions. This is already covered by `c_api_test.c` because it wouldn't compile in that case.

capital-G:
> The example does get built on C-API: https://github.com/supercollider/supercollider/actions/runs/21226569416/job/61075094502?pr=7329#step:9:729

But why is it part of the build step and not part of the ctest suite? Ideally this should only be built in a test step and not everytime someone compiles SC.

edit: I don't have any experience in ctest, so this is only a question from the outside from my experience in testing python etc.

JordanHendersonMusic:
>  But why is it part of the build step and not part of the ctest suite? Ideally this should only be built in a test step and not everytime someone compiles SC.

I believe (but cmake isn't my strong suit) that we build this target because we build all targets in the ci. 

I don't think we will build this if you only build the ctests but I'm not too sure how to add a dependency without using the add_ctest command. I don't think we need to actually check anything in the ctest, but it would be good to make sure it is built as a part of the tests, so that might be worth doing?

There are also issues with updating the ctest, they don't recompile automatically, but that's something else.

Spacechild1:
> but it would be good to make sure it is built as a part of the tests, so that might be worth doing?

The C++ test programs are only built when `ENABLE_TESTSUITE=ON` (which is the default):

https://github.com/supercollider/supercollider/blob/2ad172350566d87036d6180b0ed5dccb21c6bda0/CMakeLists.txt#L605-L607

If `ENABLE_TESTSUITE` is `OFF`, the test programs will not be built and no tests will be added.

> But why is it part of the build step and not part of the ctest suite?

Because building targets is part of the build step. The test step is for actually running the test.

I have to admit that I never used ctest myself, but the common pattern seems to be this:
```cmake
#include(CTest)

# ...

if (BUILD_TESTING)
    add_executable(foo_test foo_test.exe)
    add_test(NAME foo_test COMMAND foo_test)
endif()
```
The `CTest` module defines a `BUILD_TESTING` cache variable and calls `enable_testing()`. The `add_test` command adds a test to the test suite, but only if `BUILD_TESTING` is `ON`. See also https://cmake.org/cmake/help/latest/command/add_test.html#command:add_test. By guarding `add_executable` with `BUILD_TESTING`, we only build the test programs when testing is enabled.

There are all kinds of variations to this. SuperCollider uses its `ENABLE_TESTSUITE` option to conditionally include the testsuite directory, see above.
Opus sets `BUILD_TESTING` based on its `OPUS_BUILD_TESTING` option and uses it to conditionally add executables and tests see https://github.com/xiph/opus/blob/a3f0ec02b34281c38aefdaf4bbbd5787ad641d62/CMakeLists.txt#L34-L39 resp.
https://github.com/xiph/opus/blob/a3f0ec02b34281c38aefdaf4bbbd5787ad641d62/CMakeLists.txt#L689-L704

PR #7332, opened 2026-01-21
sclang: fix large allocation bug,

### LINKED ISSUE #7331: changed size for INT_MAX_BY_PyrSlot causes crashes with simpleNumberSeries
## Environment

macos sequoia 15.7.3

bug occurs on: 3.15 devt "bleeding edge" build downloaded yesterday(b7dffed), also when building locally

no bug:  current release 3.14.1, and an earlier 3.15 devt.( e8514ec; but this is also on a different machine, macos 12.7.6)

## Steps to reproduce

```supercollider
(1.. 139217728); //will crash the interpreter, but should throw an error
```

expected: throw error: 
```
prSimpleNumberSeries: array size 134217728 exceeds limit (134217727)
ERROR: Primitive '_SimpleNumberSeries' failed.
```
actual: crashes interpreter.

## possibly reason
I think the limit 
```
const int INT_MAX_BY_PyrSlot = INT_MAX / sizeof(PyrSlot);
```
 is somehow twice as large as it should be (i.e., 268435454 rather than 134217727) . I noticed this when messing around with `lang/langPrimSource/PyrMathPrim.cpp` for an unrelated PR.

Presumably something has changed with the Slots?

JordanHendersonMusic:
## Purpose and Motivation

Closes #7331
```supercollider
(1.. 134917722) // crash
``` 

Why.

`PyrGC:Allocate` was not allocating enough memory as it was using the sizeclass to recalculate the requested amount of bytes, this is fine in all cases expect when the required memory exceeds the size classes (2^27). With the introduction of nan boxing, we have doubled the possible size of the array we can represent, meaning this bug has become apparent.

Make internal calculation use int64 as these overflow in extreme cases.

Add tests to ensure we can always have arrays of sizes `2^28 -1`, which is the maximum we can represent. Why is it the maximum. We store the item count **and** the num bytes as an int32, that means we need to be able to represent the value `itemCount * sizeof(PyrSlot)`. 

Additionally, there was a bug in `SweepBigObjects` where if you called it multiple times, only the first one would sweep. Obviously this was added as a performance tweak, but it is actually a bug as the only way to reset it is to perform a full gc flip, which can make you easily run out of RAM as they occur rarely. This will have a performance impact when allocating large objects, but seems reasonable to avoid consuming all memory on the system.

## Types of changes

- Major bug fix in allocators :crocodile: 

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/GC.cpp
+++ b/lang/LangSource/GC.cpp
@@ -24,6 +24,7 @@
 #include "PyrObjectProto.h"
 #include "PyrSymbol.h"
 #include "InitAlloc.h"
+#include <limits>
 #include <string.h>
 #include <stdexcept>
 #include "PyrLexer.h"
@@ -342,13 +343,16 @@ HOT PyrObject* PyrGC::New(size_t inNumBytes, std::int64_t inFlags, std::int64_t 
 #endif
 
     // obtain size info
-    const int64_t alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
-    const int64_t numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
-    const int64_t numSlots = numSlotsMaybe0 < 1 ? 1 : numSlotsMaybe0;
-    const int64_t unboundedSizeclass = LOG2CEIL(numSlots);
-    const int64_t sizeclass = sc_min(unboundedSizeclass, kNumGCSizeClasses - 1);
-
-    const int64_t credit = 1LL << sizeclass;
+    const size_t alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
+    const size_t numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
+    const size_t numSlots = numSlotsMaybe0 < 1 ? 1 : numSlotsMaybe0;
+    assert(numSlots <= std::numeric_limits<int>::max());
+    // LOG2CEIL only accepts int as it uses compiler builtins.
+    const int unboundedSizeclass = LOG2CEIL(static_cast<int>(numSlots));
+    const int sizeclass = std::min<int>(unboundedSizeclass, kNumGCSizeClasses - 1);
+    assert(sizeclass >= 0);
+    static_assert(kNumGCSizeClasses < 64);
+    const uint64_t credit = 1ULL << sizeclass;
     mAllocTotal += credit;
     mNumAllocs++;
 
--- a/lang/LangSource/GC.cpp
+++ b/lang/LangSource/GC.cpp
@@ -351,20 +351,21 @@ HOT PyrObject* PyrGC::New(size_t inNumBytes, std::int64_t inFlags, std::int64_t 
     // LOG2CEIL currently only accepts ints, so we make sure that numSlots does not exceed
     // the maximum sizeclass value (which is guaranteed to fit into an int32).
     static constexpr auto maxSizeClass = kNumGCSizeClasses - 1;
+
     static constexpr size_t maxCredit = 1ULL << (maxSizeClass);
     static_assert(maxCredit <= std::numeric_limits<int32>::max());
 
-    const size_t credit = (numSlots > maxCredit) ? maxCredit : NEXTPOWEROFTWO(static_cast<int32>(numSlots));
+    // If numSlots is smaller than maxSizeClass, it is also guaranteed to fit into an int32
+    // and so it is safe to call LOG2CEIL (which currently only accepts 32-bit integers)
+    const int32 sizeclass = (numSlots >= maxCredit) ? maxSizeClass : LOG2CEIL(static_cast<int32>(numSlots));
+
+    // If the object fits inside a size band, credit is the true number of slots allocated.
+    // If is doesn't, then this is dealt with separately through classification as a 'large' object.
+    const size_t credit = 1ULL << sizeclass;
 
     mAllocTotal += credit;
     mNumToScan += credit;
     mNumAllocs++;
-
-    // If numSlots is outside int32 range, we can't cast it to perform LOG2CEIL (which only works on ints), so return
-    // max sizeclass.
-    const int32 sizeclass = (numSlots > std::numeric_limits<int32>::max())
-        ? maxSizeClass
-        : std::min<int32>(LOG2CEIL(static_cast<int32>(numSlots)), maxSizeClass);
 
     // Now we can allocate.
     PyrObject* obj = Allocate(inNumBytes, sizeclass, inRunCollection);
--- a/lang/LangSource/GC.cpp
+++ b/lang/LangSource/GC.cpp
@@ -334,30 +334,26 @@ void PyrGC::BecomeImmutable(PyrObject* inObject) { inObject->obj_flags |= obj_im
 void DumpBackTrace(VMGlobals* g);
 
 HOT PyrObject* PyrGC::New(size_t inNumBytes, std::int64_t inFlags, std::int64_t inFormat, bool inRunCollection) {
-    PyrObject* obj = nullptr;
-
-    if (inFlags & obj_permanent) {
+    if (inFlags & obj_permanent)
         return NewPermanent(inNumBytes, inFlags, inFormat);
-    }
 
 #ifdef GC_SANITYCHECK
     SanityCheck();
 #endif
 
     // obtain size info
+    const int32 alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
+    const int32 numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
+    const int32 numSlots = numSlotsMaybe0 < 1 ? 1 : numSlotsMaybe0;
+    const int32 unboundedSizeclass = LOG2CEIL(numSlots);
+    const int32 sizeclass = sc_min(unboundedSizeclass, kNumGCSizeClasses - 1);
 
-    int32 alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
-    int32 numSlots = alignedSize / sizeof(PyrSlot);
-    numSlots = numSlots < 1 ? 1 : numSlots;
-    int32 sizeclass = LOG2CEIL(numSlots);
-    sizeclass = sc_min(sizeclass, kNumGCSizeClasses - 1);
-
-    int32 credit = 1LL << sizeclass;
+    const int32 credit = 1LL << sizeclass;
     mAllocTotal += credit;
     mNumAllocs++;
 
     mNumToScan += credit;
-    obj = Allocate(inNumBytes, sizeclass, inRunCollection);
+    PyrObject* obj = Allocate(inNumBytes, sizeclass, inRunCollection);
 
     obj->obj_format = inFormat;
     obj->obj_flags = inFlags & 255;
--- a/lang/LangSource/GC.cpp
+++ b/lang/LangSource/GC.cpp
@@ -345,22 +345,33 @@ HOT PyrObject* PyrGC::New(size_t inNumBytes, std::int64_t inFlags, std::int64_t 
     // obtain size info
     const size_t alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
     const size_t numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
-    const size_t numSlots = numSlotsMaybe0 < 1 ? 1 : numSlotsMaybe0;
-    assert(numSlots <= std::numeric_limits<int>::max());
-    // LOG2CEIL only accepts int as it uses compiler builtins.
-    const int unboundedSizeclass = LOG2CEIL(static_cast<int>(numSlots));
-    const int sizeclass = std::min<int>(unboundedSizeclass, kNumGCSizeClasses - 1);
-    assert(sizeclass >= 0);
-    static_assert(kNumGCSizeClasses < 64);
-    const uint64_t credit = 1ULL << sizeclass;
+
+    const size_t numSlots = std::max<size_t>(alignedSize / sizeof(PyrSlot), 1);
+
+    // LOG2CEIL currently only accepts ints, so we make sure that numSlots does not exceed
+    // the maximum sizeclass value (which is guaranteed to fit into an int32).
+    static constexpr auto maxSizeClass = kNumGCSizeClasses - 1;
+    static constexpr size_t maxCredit = 1ULL << (maxSizeClass);
+    static_assert(maxCredit <= std::numeric_limits<int32>::max());
+
+    const size_t credit = (numSlots > maxCredit) ? maxCredit : NEXTPOWEROFTWO(static_cast<int32>(numSlots));
+
     mAllocTotal += credit;
+    mNumToScan += credit;
     mNumAllocs++;
 
-    mNumToScan += credit;
+    // If numSlots is outside int32 range, we can't cast it to perform LOG2CEIL (which only works on ints), so return
+    // max sizeclass.
+    const int32 sizeclass = (numSlots > std::numeric_limits<int32>::max())
+        ? maxSizeClass
+        : std::min<int32>(LOG2CEIL(static_cast<int32>(numSlots)), maxSizeClass);
+
+    // Now we can allocate.
     PyrObject* obj = Allocate(inNumBytes, sizeclass, inRunCollection);
 
     obj->obj_format = inFormat;
     obj->obj_flags = inFlags & 255;
+    // The size of the object is set to zero and the data inside the object is uninitialized memory.
     obj->size = 0;
     obj->classptr = class_object;
     obj->gc_color = mWhiteColor;
--- a/lang/LangSource/GC.cpp
+++ b/lang/LangSource/GC.cpp
@@ -344,7 +344,6 @@ HOT PyrObject* PyrGC::New(size_t inNumBytes, std::int64_t inFlags, std::int64_t 
 
     // obtain size info
     const size_t alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
-    const size_t numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
 
     const size_t numSlots = std::max<size_t>(alignedSize / sizeof(PyrSlot), 1);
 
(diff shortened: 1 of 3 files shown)


lang/LangSource/GC.cpp
```diff
@@ -334,30 +334,26 @@ void PyrGC::BecomeImmutable(PyrObject* inObject) { inObject->obj_flags |= obj_im
...
 #ifdef GC_SANITYCHECK
     SanityCheck();
 #endif
 
     // obtain size info
+    const int64_t alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
+    const int64_t numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
+    const int64_t numSlots = numSlotsMaybe0 < 1 ? 1 : numSlotsMaybe0;
+    const int64_t unboundedSizeclass = LOG2CEIL(numSlots);
+    const int64_t sizeclass = sc_min(unboundedSizeclass, kNumGCSizeClasses - 1);
```
Spacechild1: Just a question: why `int64_t` and not `size_t`?

First, I think we're only with unsigned numbers.

Second, `inNumBytes` is already `size_t` and I don't think we are dealing numbers that would exceed `size_t`.

Spacechild1: Similar for all other places where you're using `int64_t`.

JordanHendersonMusic: We did overflow an i32 and on 32 bit systems size_t is close to that, but I don't know if it is possible to even allocate that much memory on a 32 bit system. I'd rather avoid thinking about this if possible by just making it a fixed width integer for the internal calculations. I'll make it unsigned though.

Spacechild1: I don't know. To me it feels wrong to use `size_t` as the alloc size argument and then use `uint64_t` internally. I don't see how you would be able to allocate more memory than what fits into a `size_t`.

Of course, overflowing a `int32` on a 32-bit system is very much possible.

Spacechild1: > but I don't know if it is possible to even allocate that much memory on a 32 bit system.

Depends on the OS. On 32-bit Linux, 1 GB is reserved for the kernel. On Windows, 2 GB is for the kernel and 2 GB for user space; howerver, when linking with `/LARGE_ADDRESS_AWARE` the user space address space can be increased to 3 GB on 32-bit Windows and 4 GB on 64-bit Windows.

lang/LangSource/GC.cpp
```diff
@@ -334,30 +334,26 @@ void PyrGC::BecomeImmutable(PyrObject* inObject) { inObject->obj_flags |= obj_im
 void DumpBackTrace(VMGlobals* g);
 
 HOT PyrObject* PyrGC::New(size_t inNumBytes, std::int64_t inFlags, std::int64_t inFormat, bool inRunCollection) {
```
Spacechild1, line 337: Side note: I have no idea why the `inFlags` and `inFormat` arguments are `int64_t`... Same with `NewFrame` and `NewFinalizer`. I know, this is not your code, but it could be fixed on the way.

lang/LangSource/GC.cpp
Spacechild1: What I'm trying to say is: if we want to prevent overflow with large allocation requests, we actually have to do a range check. There are two options:

1. do it on the callsite, i.e. *before* we actually try to allocate memory. Once we're in `PyrGC::New`, it's already too late because we went through a `size_t`. This would allow us to handle too large allocation requests gracefully. However, we would have to determine all the callsites that could possibly do huge allocations.

2. We could make the `inNumBytes` argument a `uint64_t` and do the following:
    ```cpp
    // allocSize is calculated with uint64_t
    if (allocSize > std::numeric_limits<size_t>::max()) {
        throwMemfailed(inNumBytes);
    }
    // after that, everything can be size_t
    ```
    This would be much simpler to implement. The downside is that the application probably just crashes.   But at least the user sees what has gone wrong.

Finally, we could also combine 1. and 2. and only do 1. on specific callsites that are likely to cause overflow, e.g. certain Array methods.

lang/LangSource/GC.cpp
JordanHendersonMusic: I made them `int64_t` because they were originally `long long` (or just `long` I forget) and was removing variable width types — `long` and `long long` being different between windows and everyone else. In that PR I made everything the largest type just to keep the behaviour the same, if the intention of revisiting these as needed.

I'd rather keep this PR to just fixing the bug, which is the allocator allocating less than requested, and deal with these later because we should also discuss things like memory layout when these types are used in the PyrObjectHdr — e..g, should the item count be signed or not, should it be variable width (`size_t`), or should we keep the size and alignment of the header the same across architectures?

lang/LangSource/GC.cpp
JordanHendersonMusic: I like 2.

However, the limiting factor isn't size_t, it is the item count and byte size that is stored in the PyrObjectHdr, which are both i32s. This means the item limit is actually `INT_MAX / sizeof(PyrSlot)` — the linked issue shows one place where this is used. I do think we should look again at what types sc internally uses for counts, but (as I mentioned in the other comment) would rather leave that if possible as this is relatively simple bug.

I'll change this code to use size_t and see what happens.

Spacechild1: > However, the limiting factor isn't size_t, it is the item count and byte size that is stored in the PyrObjectHdr, which are both i32s.

Ah, yes. That is a whole other issue. I think we have already discussed this in the past.

IMO we should get rid of all `int32` size variables/members in sclang. That part of sclang has never really been ported to 64-bit. In fact, we might even support 48-bit signed integers (or maybe more) on the language level. That would give as an integer range of -2.8e+014 - 2.8e+014. That's roughly 280 trillions, which should be enough for most practical use cases. It is certainly large enough not to overflow allocation requests on 64-bit systems.

But that would be a separate (and huge) project.

lang/LangSource/GC.cpp
Spacechild1: We're not really talking about item counts here! `inFlags` are just flags (that actually fit into a single byte). `inFormat` is just an enum. There is no reason for these to be a 64-bit integer. These could just be `uint32`. (Ideally, we would use proper typedefs that are defined together with the flags and format enum so we actually know they have the appropriate size.)

But yes, let's leave this to another PR.

JordanHendersonMusic: Oh I completely agree, its just that they already were (on some systems) when I found them, so thought best not to change it at the time.

lang/LangSource/GC.cpp
JordanHendersonMusic: Right so I've move this to `size_t`, except where we call LOG2CEIL as this is only defined for `int` and uses compiler builtins and inline assembly to implement (clz) — I'd rather not touch that here, but we should probably look at these some other time.

I've also made the PyrGC::Collect a uint64 then check it doesn't overflow the num to scan count, which has to be signed as the rest of the code subtracts and checks for less than zero.

lang/LangSource/GC.cpp
```diff
@@ -334,30 +335,29 @@ void PyrGC::BecomeImmutable(PyrObject* inObject) { inObject->obj_flags |= obj_im
...
-    sizeclass = sc_min(sizeclass, kNumGCSizeClasses - 1);
-
-    int32 credit = 1LL << sizeclass;
+    const size_t alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
+    const size_t numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
+    const size_t numSlots = numSlotsMaybe0 < 1 ? 1 : numSlotsMaybe0;
+    assert(numSlots <= std::numeric_limits<int>::max());
+    // LOG2CEIL only accepts int as it uses compiler builtins.
+    const int unboundedSizeclass = LOG2CEIL(static_cast<int>(numSlots));
+    const int sizeclass = std::min<int>(unboundedSizeclass, kNumGCSizeClasses - 1);
```
Spacechild1: Instead of an assertion we could do an actual runtime check, so the code would also work on 64-bit systems. We just need to cap `numSlots` to the max. size class value, which is guaranteed to fit into an `int`, instead of capping the resulting size class:
```cpp
const size_t numSlots = std::max<size_t>(alignedSize / sizeof(PyrSlot), 1);
// LOG2CEIL currently only accepts ints, so we make sure that numSlots does not exceed
// the max. sizeclass value (which is guaranteed to fit into an int32)
static constexpr size_t maxCredit = 1ULL << (kNumGCSizeClasses - 1);
static_assert(maxCredit <= std::numeric_limits<int32>::max());
size_t credit;
if (numSlots > maxCredit) {
    credit = maxCredit;
} else {
    int32 sizeclass = LOG2CEIL(static_cast<int32>(numSlots));
    assert(sizeclass >= 0 && sizeclass < kNumGCSizeClasses);
    credit = 1ULL << static_cast<size_t>(sizeclass);
}
```
I hope I didn't mess anything up, but I think you get the idea.

lang/LangSource/GC.h
```diff
@@ -149,7 +149,7 @@ class PyrGC {
     // users should not call anything below.
 
     void Collect();
-    void Collect(int32 inNumToScan);
+    void Collect(uint64 inNumToScan);
```
Spacechild1, line 152: this should really be `size_t`

lang/LangSource/GC.h
```diff
@@ -220,11 +220,11 @@ class PyrGC {
     PyrObjectHdr mGrey;
 
     int32 mPartialScanSlot;
-    std::int64_t mNumToScan;
+    int64 mNumToScan;
     int32 mNumGrey;
 
-    int32 mFlips, mCollects, mAllocTotal, mScans, mNumAllocs, mStackScans, mNumPartialScans, mSlotsScanned,
-        mUncollectedAllocations;
+    int64 mAllocTotal;
```
Spacechild1, line 226: This will never be negative so this should be `size_t`.

(Most of the other members should probably be `size_t` (or `ptrdiff_t`) as well, but let's keep that for another PR :)

lang/LangSource/GC.h
```diff
@@ -220,11 +220,11 @@ class PyrGC {
     PyrObjectHdr mGrey;
 
     int32 mPartialScanSlot;
-    std::int64_t mNumToScan;
+    int64 mNumToScan;
```
Spacechild1, line 223: I would rather make this `ptrdiff_t`.

lang/LangSource/GC.h
```diff
@@ -38,7 +38,7 @@ const int kMaxPoolSet = 7;
 const int kNumGCSizeClasses = 28;
 const int kFinalizerSet = kNumGCSizeClasses;
 const int kNumGCSets = kNumGCSizeClasses + 1;
-const std::int64_t kScanThreshold = 256LL;
+const uint64_t kScanThreshold = 256LL;
```
Spacechild1, line 41: better: `size_t`. You could also remove the redundant (and wrong) `LL` suffix

Spacechild1:
Sorry for being so nitpicky about the types, but if we are going to change the types of some members, let's pick the correct type from the start.

lang/LangSource/GC.cpp
Spacechild1: Side note, the second branch can actually be simplified like this:
```cpp
} else {
    credit = NEXTPOWEROFTWO(static_cast<int32>(numSlots));
}
```
Makes it also more clear what's going on.

JordanHendersonMusic:
I don't agree right now with making `mNumToScan` or `mAllocTotal` variable width.... this is how I'm understanding this code but could be wrong....


I don't think these values represent a quantity of memory, but the number of slots returned from the GC over some time period. `mNumToScan` is since the last collection cycle, but the value passed into `PyrGC::Collect` could be over any period. Importantly, these slots might have the same memory address, meaning the amount of memory allocated could be very small but the number needed to be scanned far greater.


Here is a PR from a while ago where `mNumToScan` as an `int` overflowed  https://github.com/supercollider/supercollider/pull/6261/changes

This happened before running out of system memory, even 32 bit limit.

If (my) memory serves... this happens because some (?) call frames are not (always) freed by doing a collection cycle, but  returned to the free list when the interpreter (or was it thread) returns from it. — I can't remember exactly what optimisation was taking place here but I think something unusual/clever was happening with frames that causes the `mNumToScan` to be significantly large that the amount of memory actually allocated from the os.

I think it would be safest to make all count of slots `uint64`s.

lang/LangSource/GC.cpp
JordanHendersonMusic: I don't think this will work because when `numSlots` is greater than `maxCredit` we will not scan enough objects in the next collection cycle. This will lead to more allocations and less recycling.

Spacechild1:
> I don't think these values represent a quantity of memory, but the number of slots returned from the GC over some time period

You're absolutely right! Sorry for that! Let's keep it `uint64` then.

lang/LangSource/GC.cpp
Spacechild1: I'm not sure I understand. Currently we only use `numSlots` to calculate the `sizeclass`, which in turn is capped to `kNumGCSizeClasses - 1`. This means that `credit` would never be larger than `maxCredit` anyway.

Unless I'm completely confused, my code should be logically equivalent to yours. The only difference is that it should be safe for huge allocations on 64-bit systems; your version would abort with an assertion failure.

JordanHendersonMusic: @Spacechild1 I've copied your code in here, would you mind just checking it all looks good?

lang/LangSource/GC.cpp
```diff
@@ -334,33 +335,43 @@ void PyrGC::BecomeImmutable(PyrObject* inObject) { inObject->obj_flags |= obj_im
...
+    mNumToScan += credit;
     mNumAllocs++;
 
-    mNumToScan += credit;
-    obj = Allocate(inNumBytes, sizeclass, inRunCollection);
+    // If numSlots is outside int32 range, we can't cast it to perform LOG2CEIL (which only works on ints), so return
+    // max sizeclass.
+    const int32 sizeclass = (numSlots > std::numeric_limits<int32>::max())
+        ? maxSizeClass
+        : std::min<int32>(LOG2CEIL(static_cast<int32>(numSlots)), maxSizeClass);
```
Spacechild1: I think the check should really be:
```cpp
// If numSlots is smaller than maxSizeClass, it is also guaranteed to fit into an int32 and so it is safe to call LOG2CEIL
 const int32 sizeclass = (numSlots >= maxSizeClass) ? maxSizeClass : LOG2CEIL(static_cast<int32>(numSlots);
```

That's the point of ```static_assert(maxCredit <= std::numeric_limits<int32>::max());```.

---

However, in my code I forgot that we actually need the sizeclass later. My bad! In that case, we shouldn't call `NEXTPOWEROFTWO` *and* `LOG2CEIL`, as the former can be derived from the latter.

I would propose the following instead:
```cpp
static constexpr size_t maxSizeClass = kNumGCSizeClasses - 1;
static constexpr size_t maxCredit = 1ULL << maxSizeClass;
static_assert(maxCredit <= std::numeric_limits<int32>::max());
// If numSlots is smaller than maxSizeClass, it is also guaranteed to fit into an int32
// and so it is safe to call LOG2CEIL (which currently only accepts 32-bit integers)
const size_t sizeclass = (numSlots >= maxCredit) ? maxSizeClass : LOG2CEIL(static_cast<int32>(numSlots));
const size_t credit = 1ULL << sizeclass;
mAllocTotal += credit;
mNumToScan += credit;
mNumAllocs++;

// Now we can allocate.
PyrObject* obj = Allocate(inNumBytes, sizeclass, inRunCollection);
```

JordanHendersonMusic: okay should be good then? I also added more comments explaining what credit is.

JordanHendersonMusic: I also made the sizeband an int as that is what is returned from log2ceil, and what is passed to allocate. It is now promoted with the shift to sizet

lang/LangSource/GC.cpp
```diff
@@ -334,33 +335,44 @@ void PyrGC::BecomeImmutable(PyrObject* inObject) { inObject->obj_flags |= obj_im
...
         return NewPermanent(inNumBytes, inFlags, inFormat);
-    }
 
 #ifdef GC_SANITYCHECK
     SanityCheck();
 #endif
 
     // obtain size info
+    const size_t alignedSize = (inNumBytes + kAlignMask) & ~kAlignMask; // 16 byte align
+    const size_t numSlotsMaybe0 = alignedSize / sizeof(PyrSlot);
```
Spacechild1: I think `numSlotsMaybe0` is a leftover.

lang/LangSource/GC.cpp
```diff
@@ -334,33 +335,44 @@ void PyrGC::BecomeImmutable(PyrObject* inObject) { inObject->obj_flags |= obj_im
...
+
+    // LOG2CEIL currently only accepts ints, so we make sure that numSlots does not exceed
+    // the maximum sizeclass value (which is guaranteed to fit into an int32).
+    static constexpr auto maxSizeClass = kNumGCSizeClasses - 1;
+
+    static constexpr size_t maxCredit = 1ULL << (maxSizeClass);
+    static_assert(maxCredit <= std::numeric_limits<int32>::max());
+
+    // If numSlots is smaller than maxSizeClass, it is also guaranteed to fit into an int32
+    // and so it is safe to call LOG2CEIL (which currently only accepts 32-bit integers)
```
Spacechild1: Since we are at it, you could remove the `(which currently only accepts 32-bit integers)` part because that is already mentioned in the comment above.

lang/LangSource/GC.cpp
Spacechild1: Looks good! Just two little things, then this should be ready to be merged.

Spacechild1 (approved):

Spacechild1:
From my side this is ready for merging now! Thanks!

JordanHendersonMusic:
Hmm, the 32 bit windows c test fails when trying to allocate the largest array possible. I suppose this is okay, because it is the system running out of ram? I'm not too sure what to do in this case. 

Either way, I need to disable the test because it hangs until it timeouts.

Spacechild1:
This is expected. Unless you program very defensively, a 32-bit program typically crashes if you run out of memory. Note that the available memory is significantly less than 4 GB because a large part is reserved for the kernel. A 32-bit program on 64-bit Windows only gets 2 GB of addressable memory. (You can get the full 4 GB if you build the EXE with `IMAGE_FILE_LARGE_ADDRESS_AWARE` flag.)

JordanHendersonMusic:
Right okay, so I guess the best way forward is just to disable that test when compiling for 32 architectures.

PR #7334, opened 2026-01-21
Scale.directory, Tuning.directory: nicer formatting for output

HotwheelsSisyphus:
this was supposed to be a very minimal change here to the `Scale.directory` and `Tuning.directory` methods. but here goes another half an hour:

output before change: \ key: name
i.e., a confusing space between the \ and the key.

e.g.,
```
\ ahirbhairav: Ahirbhairav
\ ajam: Ajam
\ atharKurd: Athar Kurd
\ augmented: Augmented
\ augmented2: Augmented 2
\ bartok: Bartok
\ bastanikar: Bastanikar
\ bayati: Bayati
\ bhairav: Bhairav
\ chinese: Chinese
\ chromatic: Chromatic
```
output after:  an ok looking table that even includes a header.
```
-> key / method selector  name of scale
\aeolian                  Aeolian
\ahirbhairav              Ahirbhairav
\ajam                     Ajam
\atharKurd                Athar Kurd
\augmented                Augmented
\augmented2               Augmented 2
\bartok                   Bartok
\bastanikar               Bastanikar
etc.

```
Found this while poking around in the french language section of scsynth. A new user seems to have been confused as to what the selector should have been (e.g. for Scale.major, they thought it might be Scale.Major). Although I think this was probably a more general misunderstanding of the syntax, the unnecessary space certainly doesn't help. All this could also be documented better but I'm trying to be more focused with my contribution, so that's for another time.

(https://scsynth.org/t/a-propos-de-gamme/8981)

### DIFF
--- a/SCClassLibrary/Common/Collections/Scale.sc
+++ b/SCClassLibrary/Common/Collections/Scale.sc
@@ -164,7 +164,7 @@ Scale {
 
 	*directory {
 		var columnHeaders = ["key / method selector   name of scale"];
-		^(columnHeaders ++ this.names.collect({ |k| $\\ ++ k.asString.padRight(25) ++ all.at(k).name } )).join("\n")
+		^(columnHeaders ++ this.names.collect({ |k| $\\ ++ k.asString.padRight(24) ++ all.at(k).name } )).join("\n")
 	}
 
 	storeArgs { ^[degrees, pitchesPerOctave, tuning, name] }
--- a/SCClassLibrary/Common/Collections/Scale.sc
+++ b/SCClassLibrary/Common/Collections/Scale.sc
@@ -163,7 +163,9 @@ Scale {
 	}
 
 	*directory {
-		^this.names.collect({ |k| "\\%: %".format(k, all.at(k).name) }).join("\n")
+		var columnHeaders = ["key / method selector   name of scale"];
+		var spaces = {|symbol| String.fill(25 - symbol.asString.size, $ )};
+		^(columnHeaders ++ this.names.collect({ |k| "%% %%".format($\\, k, spaces.(k), all.at(k).name) })).join("\n")
 	}
 
 	storeArgs { ^[degrees, pitchesPerOctave, tuning, name] }
@@ -290,7 +292,9 @@ Tuning {
 	}
 
 	*directory {
-		^this.names.collect({ |k| "\\%: %".format(k, all.at(k).name) }).join("\n")
+		var columnHeaders = ["key / method selector   name of tuning"];
+		var spaces = {|symbol| String.fill(25 - symbol.asString.size, $ )};
+		^(columnHeaders ++ this.names.collect({ |k| "%% %%".format($\\, k, spaces.(k), all.at(k).name) })).join("\n")
 	}
 }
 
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic (commented):
These aren't that efficient, but I don't think it matters. Looks good to me!

SCClassLibrary/Common/Collections/Scale.sc
```diff
@@ -163,7 +163,9 @@ Scale {
 	}
 
 	*directory {
-		^this.names.collect({ |k| "\\ %: %".format(k, all.at(k).name) }).join("\n")
+		var columnHeaders = ["key / method selector   name of scale"];
+		var spaces = {|symbol| String.fill(25 - symbol.asString.size, $ )};
+		^(columnHeaders ++ this.names.collect({ |k| "%% %%".format($\\, k, spaces.(k), all.at(k).name) })).join("\n")
```
prko:
```
^this.names.collect({ |k| "\\" ++ ("%: %".format(k, all.at(k).name)) }).join("\n")
```

HotwheelsSisyphus: I included the header intentionally, and that's why I also removed the colon.

prko: Ah, yes — your version is better than my suggestion.

Sorry for not reading the code carefully enough. At first, I thought it was making the code longer, but your version is actually better for readability.

HotwheelsSisyphus:
Meanwhile I just couldn't leave it alone, so I updated the solution to something more readable now that I'm hip to String.padRight... not sure how much more efficient this is but I also don't think it matters.

prko:
I am not sure about the exact runtime efficiency, but the current code seems better than the previous version.

There is just one small thing I would like to mention: at the moment, the header in `Scale.directory` seems to contain one extra space compared with the other lines.
```
Scale.directory

-> key / method selector   name of scale
\aeolian                  Aeolian
\ahirbhairav              Ahirbhairav
```
```
Tuning.directory

-> key / method selector   name of tuning
\bp                        Bohlen-Pierce
\catler                    Jon Catler
```

HotwheelsSisyphus:
Thanks, fixed (on the second try)

HotwheelsSisyphus:
I'm not quite sure why those macos tests keep failing. Some formatting specialty maybe? It's not obvious to me from the logs.

dyfer:
CI failures are random and not related to this change, AFAICT. 

Anybody cares to approve this?

dyfer:
Also, are we sure we want to include a header? That may be confusing for programmatical use (but I'm not sure if these methods are to be used as such)

HotwheelsSisyphus:
> Also, are we sure we want to include a header? That may be confusing for programmatical use

Good point - I just removed it, I don't think it's terribly important, better to just move things forward here. I also noticed that I hadn't updated the implementation for Tuning.directory, only for Scale.directory... so I changed that too now.

prko:
Would it be better to input a tab for programmatic use?

```
^this.names.collect({ |k| $\\ ++ k.asString.padRight(16) ++ "\t" ++ all.at(k).name
```

PR #7337, opened 2026-01-23
class library: fix -absolutePath and -isAbsolutePath overrides for Windows

Spacechild1:
`String -absolutePath` is currently broken on Windows. Most notably, a path like `C:/foo` is not detected as an absolute path because the method (incorrectly) only assumes backward slashes. As a consequence, the path will be resolved to the current working directory and you get a bogus path like `C:\Program Files\SuperCollider-3.14.0/C:/foo`.

Note that Windows supports both forward and backward slashes in path names. This is also explicitly stated in the SC docs, see the section "Pathname Support" in String.schelp:

> On Windows, both forward slash "/" and backward slash "\\" are path separators.

In fact, the logic in `String -absolutePath` and `PathName -isAbsolutePath` had several flaws and I made the following fixes:
1. make sure that both methods use the exact same logic for detecting an absolute path
2. use the official MS docs as a reference (https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats); most importantly, support forward slashes!
3. a leading forward slash is not "unix-style", it is specified by MS to be relative to the current drive. It can also be a backward slash. Finally, successive slashes are allowed, they are simply ignored.
4. `standardizePath` actually works on Windows, so let's use it.

I've used the following test:
```sc
"C:/test".absolutePath; // -> C:/test
"C:\\test".absolutePath; // -> C:\test
"\\\\system07\\test".absolutePath; // UNC path -> \\system07\test
"/test".absolutePath; // -> C:/test
"\\test".absolutePath; // -> C:\test
"~/test".absolutePath; // -> C:\Users\Christof\test
"test".absolutePath; // C:\Program Files\SuperCollider-3.14.0\test

PathName("C:/test").isAbsolutePath; // -> true
PathName("C:\\test").isAbsolutePath; // -> true
PathName("\\\\system07\test").isAbsolutePath; // UNC path -> true
PathName("/test").isAbsolutePath; // -> false
PathName("\\test").isAbsolutePath; // -> false
// Note: the following returns true because PathName internally calls standardizePath
PathName("~/test").isAbsolutePath; // -> true
PathName("test").isAbsolutePath; // -> false
```

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Platform/windows/SystemOverwrites/overwrites.sc
+++ b/SCClassLibrary/Platform/windows/SystemOverwrites/overwrites.sc
@@ -1,31 +1,29 @@
 + String {
+	// See https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats for reference.
 	absolutePath {
-		var first, sep;
-		sep = thisProcess.platform.pathSeparator;
-			// 2 styles of absolute path:
-			// C:\... or \\machine_name\...
-		if((this[0].isAlpha and: { this[1] == $: and: { this[2] == sep } })
-				or: { this[0] == sep and: { this[1] == sep } }) {
+		// First check if the path is already absolute.
+		// There are 3 styles of absolute paths:
+		// 1. C:\..., 2. C:/...  and 3. \\<machine>\... (= UNC path)
+		if((this[0].isAlpha and: { this[1] == $: and: { this[2] == $\\ or: { this[2] == $/ } } })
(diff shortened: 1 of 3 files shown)


JordanHendersonMusic:
Would you mind making some unit tests for this method?

Spacechild1:
Added unit tests!

telephon (approved):
The code reads good to me.

PR #7339, opened 2026-01-23
sclang: fix duplicate keys in inlined switch crash

### LINKED ISSUE #1005: Switch statements with duplicate test cases make SCLang crash
This makes SCLang crash:
```supercollider
a = \one;
(
b = switch (a,
    \one, {1},
    \two, {2},
    \three, {3},
    \four, {4},
    \two, {2}
);
)
```
If a is any other case though it works.

---

3.6.6 on Snow Leopard

JordanHendersonMusic:
Only check inlined switches as the crash occurs when the jump offset is incorrectly calculated. Also, there is no way to peek into the generic uninlined cases so this is the best we can do here.

## Purpose and Motivation

Closes #1005

```supercollider
a = \one;
(
b = switch (a,
    \one, {1},
	\two, {2},
    \three, {3},
    \four, {4},
	\two, {2}
);
)
// crash
```

Bytecode jump offsets are calculated incorrectly. Now we just check there are no duplicate keys.

## Types of changes

- Bug fix

## Error Results

```supercollider
switch (1,
	\one, {1},
	\two, {2},
	\three, {3},
	\four, {4},
	\two, {2}
);
```
```
ERROR: Duplicate key 'two' found in inlined switch statement.
  in interpreted text
  line 5 char 5:

  	\two, {2},
       
  	\three, {3},
-> nil
```

```supercollider
switch (1,
	1, {1},
	1, {2},
	2, {3},
	3, {4},
	4, {2}
);
```
```
ERROR: Duplicate key '1' found in inlined switch statement.
  in interpreted text
  line 3 char 2:

  	1, {1},
    
  	1, {2},
-> nil
```

```supercollider
switch (1,
	inf, {1},
	inf, {2},
	2, {3},
	3, {4},
	4, {2}
);
```
```
ERROR: Duplicate key 'inf' found in inlined switch statement.
  in interpreted text
  line 3 char 4:

  	inf, {1},
      
  	inf, {2},
-> nil
```

```supercollider
switch (1,
	-inf, {1},
	2, {3},
	3, {4},
	4, {2},
	-inf, {2},
);
```
```
ERROR: Duplicate key '-inf' found in inlined switch statement.
  in interpreted text
  line 3 char 5:

  	-inf, {1},
       
  	2, {3},
-> nil
```

I don't know why the line posting result is so ugly, but I think that is another issue.
## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/PyrParseNode.cpp
+++ b/lang/LangSource/PyrParseNode.cpp
@@ -2885,44 +2885,30 @@ void compileSwitchMsg(PyrCallNode* node) {
         int offset = 0;
         int lastOffset = 0;
 
-        std::vector<PyrSlot> visitedKeys;
+        // Maintain a record of the visited keys, if there are duplicates, post a compiler error.
+        using Key = std::pair<PyrSlot, PyrParseNode*>;
+        std::vector<Key> visitedKeys {};
         visitedKeys.reserve(numArgs);
-        const auto hasDuplicateKeyPostErrorAndPrepareForReturn = [&](auto node, PyrSlot key) -> bool {
-            if (const auto duplicateKey = std::find(visitedKeys.begin(), visitedKeys.end(), key);
-                duplicateKey != visitedKeys.end()) {
-                error("Duplicate key found in inlined switch statement\n");
-                nodePostErrorLine(node);
-                compileErrors++;
-                return true;
-            } else
-                return false;
-        };
 
         for (; argnode; argnode = nextargnode) {
             nextargnode = argnode->mNext;
             if (nextargnode != nullptr) {
                 ByteCodes byteCodes = compileSubExpressionWithGoto((PyrPushLitNode*)nextargnode, 0x6666, true);
+                PyrPushLitNode* keyargnode = (PyrPushLitNode*)argnode;
 
                 PyrSlot* key;
-                PyrSlot value;
-                SetInt(&value, offset);
-                PyrPushLitNode* keyargnode = (PyrPushLitNode*)argnode;
                 if (isAtomicLiteral(argnode)) {
                     key = &keyargnode->mSlot;
-                    if (hasDuplicateKeyPostErrorAndPrepareForReturn(keyargnode, *key))
-                        return;
+                    visitedKeys.push_back({ *key, keyargnode });
                 } else {
                     PyrBlockNode* bnode = (PyrBlockNode*)slotRawPtr(&keyargnode->mSlot);
                     PyrDropNode* dropnode = (PyrDropNode*)bnode->mBody;
                     PyrPushLitNode* litnode = (PyrPushLitNode*)dropnode->mExpr1;
                     key = &litnode->mSlot;
-                    if (hasDuplicateKeyPostErrorAndPrepareForReturn(litnode, *key))
-                        return;
+                    visitedKeys.push_back({ *key, litnode });
                 }
 
-                visitedKeys.push_back(*key);
-
-                int index = arrayAtIdentityHashInPairs(array, key);
+                const int index = arrayAtIdentityHashInPairs(array, key);
                 PyrSlot* slot = array->slots + index;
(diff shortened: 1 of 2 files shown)


lang/LangSource/PyrParseNode.cpp
```diff
@@ -2713,6 +2713,20 @@ void compileSwitchMsg(PyrCallNode* node) {
         int absoluteOffset = byteCodeLength(gCompilingByteCodes);
         int offset = 0;
         int lastOffset = 0;
+
+        std::vector<PyrSlot> visitedKeys;
+        visitedKeys.reserve(numArgs);
+        const auto hasDuplicateKeyPostErrorAndPrepareForReturn = [&](auto node, PyrSlot key) -> bool {
```
telephon: Despite being detailed, this name (`hasDuplicateKeyPostErrorAndPrepareForReturn`) kept me busy trying to understand. Since the code is there right below, it may be enough to call it `checkForDuplicateKeys`? 

Alternatively, there could be a function `hasDuplicateKeys` which just does the check, and a little bit of code duplication at the call site where the return actually happens. 


```
if (hasDuplicateKeys(keyargnode, *key)) {
               error("Duplicate key found in inlined switch statement\n");
                nodePostErrorLine(node);
                compileErrors++;
                return;
}

Comment (JordanHendersonMusic) [reply lang/LangSource/PyrParseNode.cpp]: ...yes this is confusing on second glance, I'll rewrite!

Comment (JordanHendersonMusic) [reply lang/LangSource/PyrParseNode.cpp]: @telephon  I think I've addressed your concern now, it should be much clearer!

File: lang/LangSource/PyrParseNode.cpp [hunk 2]
```diff
@@ -2929,6 +2935,60 @@ void compileSwitchMsg(PyrCallNode* node) {
             }
         }
 
+        const auto keyLessThan = [](const Key& lhs, const Key& rhs) -> bool {
+            return lhs.first.rawBytes() < rhs.first.rawBytes();
+        };
+        std::sort(visitedKeys.begin(), visitedKeys.end(), keyLessThan);
+
+        // Note, we don't need to use the proper pyrslot ==, (or < if there was one) operator as that respects nans and
+        // other double comparison rules. Here we know the user doesn't have a nan in their inlineable switch, as it is
```
telephon, line 2944: There is a way to make a nan, namely to just write `0/0`.

JordanHendersonMusic: That's not a literal though so it can't be inlined.

telephon: ok, then I suppose it is fine!  It may be worth a comment?

lang/LangSource/PyrParseNode.cpp
JordanHendersonMusic, line 2857: @telephon what I'll do instead is flip this check with an early return, that way it's more obvious we are only considering the inlined case.

telephon (approved):

mtmccrea:
This may be contributing to a new sclang crash: #7476

lang/LangSource/PyrSlot.h
```diff
@@ -319,6 +319,9 @@ union PyrSlot {
         // Doubles have odd comparison rules, otherwise compare the raw data.
         return (lhs.isDouble() && rhs.isDouble()) ? lhs.getDouble() == rhs.getDouble() : lhs.u_raw == rhs.u_raw;
     }
+    [[nodiscard]] friend inline bool operator<(PyrSlot lhs, PyrSlot rhs) noexcept {
+        return (lhs.isDouble() && rhs.isDouble()) ? lhs.getDouble() < rhs.getDouble() : lhs.u_raw <= rhs.u_raw;
```
JordanHendersonMusic, line 323: @mtmccrea ah, typo, both should be less than, not less than equal.

PR #7340, opened 2026-01-23
sclang: compile error when duplicate var name is found

### LINKED ISSUE #1201: duplicate instance variable definitions are not detected
The following class definition won't fail although it should:

`````
Test {
	var variable;
	var variable;

}
````

### LINKED ISSUE #3263: Duplicate instance variables in class not throwing error on compile, as they should
```
TestClass{
	var <>items, listview, modifier, <button, defaultAction, winPosition,<value,
	           items, listview, modifier, <button, defaultAction, winPosition,<value;
}
```
compiles without error, but ist should throw an error. 
Using the variables in the class then causes random problems. 

SuperCollider3.9dev
SC-cca12ff02a774a9ea212e8883551d3565bb24a6f.zip

### PR DESCRIPTION
**Pull request by JordanHendersonMusic**
## Purpose and Motivation

Closes #1201
Closes #3263

Old
```supercollider
TestClass {
  var <var1 = 2;
  var <var1 = 3;
}
TestClass().var1 // doesn't work
```

New
```
ERROR: Found duplicate instance variable name 'var1'
  in file '/home/jordan/.repo/SuperCollider/SCClassLibrary/Common/Core/Object.sc'
  line 2 char 16:

    var <var1 = 2;
                  
    var <var1 = 3;
ERROR: Library has not been compiled successfully.
```

Also works for classvars and const vars.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing

### DIFF
--- a/lang/LangSource/PyrParseNode.cpp
+++ b/lang/LangSource/PyrParseNode.cpp
@@ -18,6 +18,9 @@
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
 */
 
+#include "PyrObject.h"
+#include "PyrSlot.h"
+#include "PyrSymbol.h"
 #include "SCBase.h"
 #include "PyrParseNode.h"
 #include "PyrLexer.h"
@@ -28,6 +31,8 @@
 #include "PyrKernelProto.h"
 #include "PyrObjectProto.h"
 #include "GC.h"
+#include <algorithm>
+#include <iterator>
 #include <new>
 #include <string>
 #include <optional>
@@ -565,13 +570,13 @@ PyrClass* getNodeSuperclass(PyrClassNode* node) {
 }
 
 void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* superclassobj) {
-    PyrVarListNode* varlist;
-    PyrVarDefNode* vardef;
-    PyrSlot *islot, *cslot, *kslot;
-    PyrSymbol **inameslot, **cnameslot, **knameslot;
-    PyrClass* metaclassobj;
-    PyrMethod* method;
-    PyrMethodRaw* methraw;
+    PyrVarListNode* varlist = nullptr;
+    PyrVarDefNode* vardef = nullptr;
+    PyrSlot *islot = nullptr, *cslot = nullptr, *kslot = nullptr;
+    PyrSymbol **inameslot = nullptr, **cnameslot = nullptr, **knameslot = nullptr;
+    PyrClass* metaclassobj = nullptr;
+    PyrMethod* method = nullptr;
+    PyrMethodRaw* methraw = nullptr;
     int instVarIndex, classVarIndex;
 
     // copy superclass's prototype to here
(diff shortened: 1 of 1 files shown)


lang/LangSource/PyrParseNode.cpp
```diff
@@ -783,6 +788,48 @@ void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* super
             break;
         }
     }
+
+    // Is vector faster than a set? Since symbols are just pointers, one would hope for really good auto vectorization.
+    // That and it is unlikely we will have that many members?
```
Spacechild1: Good question... `std::unordered_set` will have lots of tiny memory allocations because it's node based.

(Boost 1.81 added open addressing hash containers, e.g. https://www.boost.org/doc/libs/latest/libs/unordered/doc/html/unordered/reference/unordered_flat_set.html. Unfortunately, we are still at 1.70.)

A hash set would certainly simplify the code. You could try to compare `std::vector` with `std::unordered_set` and see if it has a measurable effect on the Class Library compile time.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -783,6 +788,48 @@ void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* super
...
+    auto findDuplicateName =
+        [names = std::vector<PyrSymbol*>()](const PyrSymbolArray* array) mutable -> std::optional<PyrSymbol*> {
```
Spacechild1, line 794: Nice trick with the lambda capture! I have to remember that :)

lang/LangSource/PyrParseNode.cpp
```diff
@@ -783,6 +788,48 @@ void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* super
...
+        if (array == nullptr)
+            return std::nullopt; // Arrays can be null, meaning, empty.
+        if (names.capacity() < array->size)
```
Spacechild1: This if-check is redundant. `reserve` has no effect if the required size is smaller than the current capacity. It never shrinks the vector.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -783,6 +788,48 @@ void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* super
...
+
+        std::copy(array->symbols, array->symbols + array->size, std::back_inserter(names));
+        std::sort(names.begin(), names.end());
+        const auto maybe_duplicate = std::adjacent_find(names.begin(), names.end());
+
+        if (maybe_duplicate != names.end()) {
+            return { *maybe_duplicate };
+        }
+
+        names.clear();
```
Spacechild1: Shouldn't you also clear the vector on the happy path?

I think it's better to just clear it once at the beginning of the function.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -783,6 +788,48 @@ void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* super
...
+    // Is vector faster than a set? Since symbols are just pointers, one would hope for really good auto vectorization.
+    // That and it is unlikely we will have that many members?
+    auto findDuplicateName =
+        [names = std::vector<PyrSymbol*>()](const PyrSymbolArray* array) mutable -> std::optional<PyrSymbol*> {
+        if (array == nullptr)
+            return std::nullopt; // Arrays can be null, meaning, empty.
+        if (names.capacity() < array->size)
+            names.reserve(array->size);
+
+        std::copy(array->symbols, array->symbols + array->size, std::back_inserter(names));
```
Spacechild1: Personally, I would rather do:
```cpp
names.insert(names.end(), array->symbols, array->symbols + size);
```
This way you wouldn't need to call `reserve` because `insert` already does it for you.

lang/LangSource/PyrParseNode.cpp
```diff
@@ -783,6 +788,48 @@ void fillClassPrototypes(PyrClassNode* node, PyrClass* classobj, PyrClass* super
...
+
+        if (maybe_duplicate != names.end()) {
+            return { *maybe_duplicate };
+        }
+
+        names.clear();
+        return std::nullopt;
+    };
+
+    if (const auto duplicate = findDuplicateName(slotRawSymbolArray(&classobj->instVarNames)); duplicate) {
```
Spacechild1: That `; duplicate` is not necessary in this case.

lang/LangSource/PyrParseNode.cpp
Spacechild1: Aha, I guess it's because we return from the outer function if `findDuplicateName` returns a result. I still think clearing the vector on the top would be clearer.

JordanHendersonMusic: I'll add the clear at the beginning, definitely cleaner!

lang/LangSource/PyrParseNode.cpp
JordanHendersonMusic: Ah, it originally returned an iterator before I refactored it. Yes this can go.

JordanHendersonMusic:
Technically this is a breaking change.
```
A { 
   var var1;
}
B : A {
   var var1;
}
```

This no longer compiles. It didn't work before, so would have caused bug, but now it is an outright compiler error.

JordanHendersonMusic:
@Spacechild1 I think I've addressed your feedback, thanks!

Spacechild1 (approved):

dyfer:
I don't disagree with this PR, but I think it prevents the current release of sc3-plugins from working (well just a single class, but still): https://github.com/supercollider/sc3-plugins/pull/415

I wonder how many other classes in the wild will have issues here.

Also, is https://github.com/supercollider/supercollider/issues/7348 caused by this or unrelated?

JordanHendersonMusic:
First example. That's a good thing! That code was just wrong.

dyfer:
Based on my experience encountering the duplicate var name error and seeing it in a couple dozen places in my own SC install, between my classes, quarks, and non-public extensions, I was thinking of the following:

As this change will likely cause at least some user extensions to prevent the library from compiling, would we want to consider turning this into warnings (or even posting "errors") but still allowing the class library to compile (as before) in SC 3.15, and postponing the current behavior (class lib doesn't compile, interpreter exits) until SC 3.16? That way we would have a chance to warn most everyone the change is coming, instead of just stopping the library from compiling.

Even if we fix up all the listed quarks, there will be user code that will prevent SC from starting and that can be quite frustrating when it comes "out of the blue". 

Maybe this is something to discuss at the next dev meeting?

JordanHendersonMusic:
Sounds good to me! 

I really didn't think this would be that much of an issue because it does some weird behaviour that I suspect would have been causing bugs.

There is also the pr about initialised vars which would then have the same behaviour. I think we should automate this error posting by comparing against the current version. Is the current semver set as a macro define by cmake somewhere?

dyfer:
> There is also the pr about initialised vars which would then have the same behaviour. I think we should automate this error posting by comparing against the current version.

Sounds good!

FWIW I thought we'd just comment out `compileErrors++;` statements with a note to uncomment them for 3.16, but if you have a cleaner solution then we should go for that.

> Is the current semver set as a macro define by cmake somewhere?

Not a macro, but the version is set in https://github.com/supercollider/supercollider/blob/develop/common/SC_Version.hpp.in, populated from https://github.com/supercollider/supercollider/blob/develop/SCVersion.txt

PR #7344, opened 2026-01-25
sclang: disable inlining in switch when a case is nil

### LINKED ISSUE #4178: switch statement: inconsistent behavior with arrays and default function.
Environment
-----------

* Your SuperCollider version: 3.10, 3.10-beta2
* Your operating system and version: macOS 10.12.6 

Steps to reproduce (for bugs)
-----------------------------

```supercollider
/*  case 1  */
(
35.do{
	// var test = nil;   // works
	// var test = 35;    // works
	// var test = 5;     // works, default func
	var test = [35, 72]; // inconsistent results when test is an array
	
	switch (test,
		nil, { "matched to nil" },
		35, { 35 },
		{ "default: ".post; test }
	).postln;
}
)

/*  case 2  */
(
35.do{
	// var test = nil;   // works
	// var test = 35;    // works
	// var test = 5;     // works, default func
	// var test = [35, 72]; // works, catches match
	var test = [5, 7]; // this now works with the default function, if there's an array in the switch statement cases
	
	switch (test,
		nil, { "matched to nil" },
		35, { 35 },
		[35, 72], { [35, 72] }, // comment this out and it'll break
		{ "default: ".post; test }
	).postln;
}
)
```

Expected Behavior
-----------------
When the switch contains no match to `test`, it should return the result of the default function.

Current Behavior
----------------
Case 1:
When `test` is an array, it's caught only inconsistently by the default function, otherwise answers the `nil` case.  Results vary on repeated executions.

Case 2:
If there is an array in the switch cases, `test` can be an array and work properly. If there is no array in the switch cases, `test` fails if it's an array.  Results vary on repeated executions.

JordanHendersonMusic:
## Purpose and Motivation

Closes #4178

Run this a bunch, it is random.
```
switch([], nil, { $0 }, { $X });
```

This is because the default case is stored as a `nil` in the inlined jump offset array so when the hash of the array (`[]`) (aka its address) is used to offset into the jump offsets, it picks either branch depending on what it encounters first.

This PR disables inlining when there is a literal nil as a case.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/PyrParseNode.cpp
+++ b/lang/LangSource/PyrParseNode.cpp
@@ -20,7 +20,6 @@
 
 #include "PyrObject.h"
 #include "PyrSlot.h"
-#include "PyrSymbol.h"
 #include "SCBase.h"
 #include "PyrParseNode.h"
 #include "PyrLexer.h"
@@ -40,11 +39,8 @@
 #include <stdlib.h>
 #include <stdarg.h>
 #include <cctype>
-#include "InitAlloc.h"
 #include "PredefinedSymbols.h"
 #include "SimpleStack.h"
-#include "PyrPrimitive.h"
-#include "SC_Win32Utils.h"
 #include "SC_LanguageConfig.hpp"
 #include "SC_Codecvt.hpp"
 #include "SpecialSelectorsOperatorsAndClasses.h"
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
@lnihlen I wondered what you thought about this? Fixing this indeterminacy while attempting to keep the inlined optimisation in-tact is quite challenging...

lnihlen:
> @lnihlen I wondered what you thought about this? Fixing this indeterminacy while attempting to keep the inlined optimisation in-tact is quite challenging...

I'm always for improved diagnostics, so from that perspective this seems great.

One thing is that control flow keywords like `if`, `while`, `switch` are not actually keywords, and in fact any SC programmer can author methods with those names and add them to any class. So I wonder from a language compatibility standpoint if somebody has a random method out there called `switch` that will confuse this code.

For Hadron I've already decided to make the control flow keywords actual keywords, so users won't be allowed to name methods with those names. I think we should commit to control flow being first-class language construct, and I would argue with inlining we're kind of halfway there already.

Nice work

JordanHendersonMusic:
Thanks for taking a look, I don't want to accidentally change something here that makes your work invalid or more difficult by changing the language subtly!

I suppose the other option would be to go the other way, to make the inclusion of `nil` as a case force the switch to not be inlined. This would completely preserve backwards compatibility and work correctly as the bug here is specifically the combination of inlined and a nil case. It would be a performance regression for people who use `nil`, but that's better than a compiler warning for them, and both are far better than the broken behaviour we have now.

User defined switch statements wouldn't make a difference because the interpreter only calls them when it isn't inlined at the call sight, so if a user had that working, they were already avoiding this bug.

Imo, the issue isn't so much that there are user defined control flow and there is a difference between inlined and non-inlined, it is that the rules are completely arbitrary and simply down to the current state of the compiler optimisations. E.g., with `var` 'hoisting' a large amount of currently uninlineable functions could be inlined, it's quite easy to do with another compiler pass, would offer great performance, but also break so much code. If there were simpler rules, say `x.if(...)` is always dispatched, and `if(x) {...}` is always inlined then I think this would be manageable and provide the best of both worlds.

Spacechild1 (dismissed):

JordanHendersonMusic:
Lets just hang off on merging this, as after sleeping on it, I think it might be better to make `nil` is switches okay, but not inlined. I need to do some testing first though.

JordanHendersonMusic:
Okay that is done. Now there is no breaking change and nil if allowed in a switch statement.

I've added some tests.

And I've introduced a new function that returns the literal value.

testsuite/classlibrary/TestSwitch.sc
```diff
@@ -0,0 +1,40 @@
+TestSwitch : UnitTest {
+	test_nil {
+		// This is indeterminate, but quite likely, 50 was more than enough to trigger this bug in practice.
```
lnihlen, line 3: Have you tested this without your patch? What's the fail rate?

lang/LangSource/PyrParseNode.cpp
```diff
@@ -2291,6 +2285,79 @@ bool isAtomicLiteral(PyrParseNode* node) {
...
+        return std::nullopt;
+
+    const auto& block = static_cast<const PyrBlockNode&>(maybeBlock);
+
+    // Having arguments and variables mean we can't inline it, therefore, it isn't a literal.
+    if (block.mArglist || block.mVarlist) {
+        if constexpr (Post == MaybePostNonInlineableWarning::True) {
+            gNumUninlinedFunctions += 1;
+            if (SC_LanguageConfig::getPostInlineWarnings()) {
+                post("WARNING: FunctionDef contains variable declarations and so"
```
lnihlen: The error only complains about variable declarations but not argument declarations, but either will generate the error message.

testsuite/classlibrary/TestSwitch.sc
JordanHendersonMusic:
```
100.collect{
	String.fill(50) { switch([], nil, { $0 }, { $X }) }.as(IdentityBag).contents[$0]
}.mean / 50 * 100
```
24%

lang/LangSource/PyrParseNode.cpp
JordanHendersonMusic: Ah yes I just copied this bit of code over from before.

lnihlen (approved):

PR #7346, opened 2026-01-25
sclang: fix lexer ambiguity with scale degrees and whitespace/infix binary ops

### LINKED ISSUE #2568: Language syntax ambiguities for "method:" selector notation
screen shot 2016-12-18 at 6 36 41 pm

The first line evaluates to 16, and the second line throws an error because the `16s` is immediately understood as 16.1, leaving the garbage method selector `qrsum:`. This could be a problem with the language syntax, or the IDE highlighting, or both.

JordanHendersonMusic:
## Purpose and Motivation

Closes #2568

`16spow: 2` is parsed as (16s).pow(2). This should be an error. 

This is technically a breaking change, but seems very unlikely that someone will be relying on it.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/PyrLexer.cpp
+++ b/lang/LangSource/PyrLexer.cpp
@@ -681,8 +681,7 @@ digits_1: /* number started with digits */
             goto accidental2;
         if (std::isalpha(d)) {
             yytext[yylen] = 0;
-            post("encountered an accidental '%s' with unknown trailing text. Ensure there is whitespace after "
-                 "accidentals.\n",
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
> IMO we should reject all use of infix method names that are not preceded by whitespace. The compiler bug is that it doesn't consistently fail this case.

@jamshark70 

I've done something different here, I've made it not possible to parse scale degree literals with trailing text that doesn't match the scale indicator (either 's' or 'b'). I figured this would be less of a breaking change as it doesn't invalidate code that uses this.

PR #7352, opened 2026-01-28
cmake: remove old version variables

dyfer:
## Purpose and Motivation

Cleanup.

I believe we don't need these vars in cmake, but I'm not 100% sure.

## Types of changes

- Cleanup

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


capital-G (approved):
I also couldn't find any references to it elsewhere and since we the pipeline is green, this is good to go.

Always appreciate your work on making CMake a nicer place, thx Marcin <3

PR #7353, opened 2026-01-29
sclang: make duplicate class vars a warning in 3.15, error from 3.16

JordanHendersonMusic:
## Purpose and Motivation

Turns recent duplicate class var into a warning for version 3.15, but a compilation error from 3.16. Also introduces a way to make this easier as this is something we might want to use going forward.  

Now the output is
```
ERROR: Found duplicate instance variable name 'name'
  in file '/home/jordan/.local/share/SuperCollider/downloaded-quarks/wslib/wslib-classes/Main Features/File management/FolderLibrary.sc'
  line 97 char 10:

        var <name;
            
  
WARNING: From version 3.16.0 onwards the preceding error will be a compilation failure, please fix the code before updating.

```

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/common/SC_Version.hpp.in
+++ b/common/SC_Version.hpp.in
@@ -25,12 +25,17 @@
 
 // Does not support tweaks.
 // Useful for branches in the code as this inherits tuple's lexigraphical comparisons.
+// Initialise with {Major, Minor, Patch}.
 struct SemanticVersion : public std::tuple<uint32_t, uint32_t, uint32_t> {
-    using std::tuple<uint32_t,uint32_t, uint32_t>::tuple;
+    using std::tuple<uint32_t, uint32_t, uint32_t>::tuple;
+
+    [[nodiscard]] constexpr uint32_t major() const { return std::get<0>(*this); } 
+    [[nodiscard]] constexpr uint32_t minor() const { return std::get<1>(*this); } 
+    [[nodiscard]] constexpr uint32_t patch() const { return std::get<2>(*this); } 
 
     [[nodiscard]] std::string asString() const {
         std::stringstream out;
-        out << std::get<0>(*this) << "." << std::get<1>(*this) << "." << std::get<2>(*this);
+        out << major() << "." << minor() << "." << patch();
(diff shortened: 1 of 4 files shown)


JordanHendersonMusic:
@dyfer like this perhaps?

JordanHendersonMusic:
We discussed improving the error message providing the user with instructions on how to fix this. 

We also talked about about having a url link in the post window — perhaps this could be a link to the docs (URL/URI handler in qt?). There is a bigger discussion here in how do we apply this to other errors.

Otherwise, the content here, was agreed and should be merged before 3.15 and made a policy of all future compiler errors.

JordanHendersonMusic:
@capital-G I know we talked about improving the error message in the dev meeting, but I'd like to merge this as is for now, wait for #7363, and then redo all the compiler error messages with real formatting, enumerate all the compiler errors, and give them clickable links. This way, we can ensure all the errors have a standard formatting.

What do you think?

JordanHendersonMusic:
Can this be merged, I'm waiting on this for a few other PR. Note that this doesn't change the warning about duplicate class vars, it only introduces the mechanism to allow specifying an future version in which the warning becomes an error.

dyfer (approved):
I think this is good, thanks!

Re: implementation of the version check: I have only basic understanding of the cpp side. Considering this is rather cosmetic in general - if you think the implementation is good, I think this can be merged.

dyfer:
> Note that this doesn't change the warning about duplicate class vars,

Are you sure? Doesn't this PR turn the class lib compilation error from duplicate vars into a warning, at least for SC 3.15?

JordanHendersonMusic:
Sorry I meant it doesn't improve the warning message, which was something else that was suggested.

JordanHendersonMusic:
@dyfer, I just added some comments, a couple of helper methods to document, and move the declaration to a header  so it can be used elsewhere. I'll merge tomorrow once the tests pass.

PR #7355, opened 2026-01-30
docs: clarify integer vs. float default values in UGen arguments

### LINKED ISSUE #7343: Inconsistent default argument types between method call aid (including documentation’s method signature section) and documentation's method argument section
## Environment

## Steps to reproduce

1. Open SuperCollider IDE.
2. Type `PlayBuf.ar(` or `BufRd.ar(` and trigger the method call aid (autocomplete/help popup).
3. Observe the default argument values shown in the method signature section.
4. Compare these with the argument description section in the documentation.
5. Notice that loop and startPos are shown as floats in the call aid/signature, but described as integers in the argument section.

### Expected behavior:
Default values in the method call aid and method signature section should match the documented types in the argument description section.
Integer arguments such as loop and startPos should be displayed as integers (0, 1, etc.), not floats (0.0, 1.0).

### Actual behavior:
There are inconsistencies between the method call aid (and the documentation’s method signature section) and the argument description section of the documentation, particularly regarding default argument values and their types.

#### Examples
The _method call aid_ and _method signature section of the documentation_ show these values as floats, while the argument description section explains them as integers (explicitly or implicitly).
- For `PlayBuf.ar` and `BufRd.ar`:
   - In the argument description section, `loop` is documented as 0 or 1 (integer).
   - In the method call aid and method signature section, `loop`'s default is shown as 1.0 (float).
- For PlayBuf:
   - `startPos` is semantically an integer (frame index).
      - In the method call aid and method signature section, it appears as a float.
      - In the argument description section, there is no explicit information, but an integer is appropriate.
   - `doneAction` in `PlayBuf.ar` is consistently shown as an integer in both places, so no issue there.

prko:
Closes #7343.

## Purpose and Motivation

This PR clarifies a common point of confusion for beginners regarding the display of default values for UGen arguments in the SuperCollider IDE (Method Call Assistance) and the Method Signature section of the documentation.

There, default values are often displayed as floats (e.g., `0.0`), even when the argument is conceptually an integer (e.g., `loop: 0`).

This change adds an explanation to the `UGen` documentation and the `Terms for Beginners` guide. It clarifies that this discrepancy reflects the underlying architecture of the audio server (scsynth and supernova), where UGen arguments are treated as 32-bit floating-point values during signal processing. It emphasizes that integer and float notations are functionally equivalent in this context.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation


JordanHendersonMusic:
Arguments do not have to be passed to the server, you could switch on their value and send an altogether different ugen. 

If an integer default is shown as a float in the ide, that's a bug. Let's not document it, let's fix it.

prko:
@JordanHendersonMusic 
Oh! Thanks!!
I’ll close this PR once the other one is opened.

prko:
>If an integer default is shown as a float in the ide, that's a bug. Let's not document it, let's fix it.

As discussed in #7561, this behaviour will not be changed. It would be sensible to document it. Most users are unlikely to need the explanation, but newcomers — particularly those approaching SC as a way into electroacoustic music — might hesitate briefly, and this note could help to clarify things for them.

JordanHendersonMusic:
I'm unsure about this. The documentation is not a lesson plan, it's a contract with the user - we must ensure we never/rarely break what is in the contract. Additionally, every word written is a development and maintenance burden. I just don't see the benefit for all of these words unfortunately.

PR #7356, opened 2026-01-31
Keep consistency between getting and setting the s.options.device

bgola:
## Purpose and Motivation

Currently if you set different `s.options.inDevice` and `s.options.outDevice`, when you get the `s.options.device` it returns an array consisting of `[inDevice, outDevice]`. 

The problem is that if you try to set `s.options.device` to an array, this will fail when booting because `s.options.device_` is  setting both `inDevice` and `outDevice` to the value it is set, so both `inDevice` and `outDevice` are set as the array. Example:

```
s.options.device = ["Test", "Test2"]
-> a ServerOptions
s.options.inDevice.postln;
[Test, Test2]
s.options.outDevice.postln;
[Test, Test2]
```

It becomes a bigger issue because if you do something like:

```
s.options.inDevice = "Test";
s.options.outDevice = "Test2";
```

And then do:
`s.options.device = s.options.device;`

This will fail to boot, which I believe is a bad behavior. 

In this PR I add the possibility of setting the `s.options.device` as a two-element array, so it is compatible with the getter pattern, keeping the method consistent. I also updated the documentation.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/Classes/ServerOptions.schelp
+++ b/HelpSource/Classes/ServerOptions.schelp
@@ -36,7 +36,7 @@ method:: blockSize
 The number of samples in one control period. The default is 64.
 
 method:: device
-A String that allows you to choose a sound device to use as input and output. The default, code::nil::, will use the system's default input and output device(s). See link::Reference/AudioDeviceSelection:: for more details.
+A String or a 2-element Array of Strings that allows you to choose a sound device to use as input and output. The default, code::nil::, will use the system's default input and output device(s). See link::Reference/AudioDeviceSelection:: for more details.
 
 method:: inDevice
 A String that allows you to choose an input sound device. The default, code::nil::, will use the system's default input device. See link::Reference/AudioDeviceSelection:: for more details.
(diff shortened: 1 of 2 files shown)


telephon (approved):
Thanks you, I didn't know this was part of the spec.

JordanHendersonMusic:
My only point about this is the use of `isKindOf` in that it means `List` and other collections won't work. 

Or, what if we accepted symbols in the future, and someone passed a `SymbolArray`? This wouldn't work because `SymbolArray` isn't a kind of `Array`. 

I don't see a nice way around this, as the problem stems from:
1. Considering `String` to natively be a collection,
2. The interface of this method.

Perhaps we'd need a `ServerDevice` class so you can do all of these things...
```
s.options.device.input = "...";
s.options.device = [...]; // setter turns it into a ServerDevices
```

I don't know if this is overkill, but it might solve some of the other issues regarding queries?

telephon:
Actually what we have been needing quite often is a reliable method interface that will always return an Array. I'm calling it asArray2, because asArray is already assumed.

```supercollider
Object {
   asArray2 { ^[this]
}

Array {
    asArray2 { ^this }
}

SequenceableCollection {
    asArray2 { ^this.asArray }
}
```

Then the method might look like:

```supercollider
device_ { |dev|
	dev = dev.asArray2;
	inDevice = dev.wrapAt[0];
	outDevice = dev.wrapAt[1];
}
```

Issue #7342, opened 2026-01-24
Support `.scx` plugin extension on Linux


Spacechild1:
## Motivation

For whatever reason, UGen plugins on Linux use the generic `.so` extension instead of `.scx`, which used on Windows on macOS.

First, this is an inconsistency that I really don't like.

Second, and more importantly, it means that you cannot ship libraries together with the plugin because they have the same extension and the Server searches recursively.

Fortunately, there is a workaround: put the `.so` files in a folder called `scide_<something>` or `ignore`.

## Description of Proposed Feature

However, this feels rather hacky. Therefore I would propose to introduce `.scx` as a supported plugin extension on Linux.

Changing the plugin extension is not a problem with the built-in plugins and sc3-plugins, because they are always tied to a specific release. The actual problem are third-party plugins. One important goal of the plugin API changes (https://github.com/supercollider/supercollider/issues/6102) is to maintain source compatibility, i.e. third-party developers should only have to recompile their plugins, but they shouldn't have to change their code. If we change the plugin extension, everyone would have to update their build scripts. We have to keep supporting the `.so` extension, at least for a while, to provide an update path.

## Plan for Implementation

Here's a possible interim solution: If the server detects at least one `.scx` file in a particular folder, it will not look for any `.so` in that same folder (or any of its subfolders). This means that people can already start shipping `.so` next to `.scx` files.

capital-G:
This is a very good idea! IIUC: the server tries to open each .so file which is really not nice - it is much better to guide the server here by using the `scx` extension.

IMO we could simply switch to `.scx` with 3.15 - all plugins need to be rebuilt anyway and renaming it to `.scx` can also be done by the user manually if the build script did not get updated.

We should probably offer an "upgrade" document which explains the motivation behind the API breakage and also what kind of small things should be adjusted?

Spacechild1:
I'm trying to maintain source compatibility and I think this includes the build scripts. However, changing the plugin extension is a one line change in the `CMakeLists.txt`, so maybe we can do a breaking changes in this case.

Personally, I would still prefer soft deprecation, using the strategy outlined in the PR description. We could also print a warning for every `.so` that is found and ask the user to rename the extension to `.scx`. We might eventually remove support for `.so` files in a later SC version.

The problem with hard deprecation is that you wouldn't be able to build the same (third-party) plugin for different SC versions without modifying the build script. (Not everyone will immediately upgrade to SC 3.15.)

---

[...]

PR #7360, opened 2026-02-04
various Server plugin changes/fixes

### LINKED ISSUE #6040: Windows: allow developers to ship .dll files next to .scx extensions
When developing new extensions, there might be the need of shipping them with external .dll / .so / .dynlib files to link against. Dynamically setting the `rpath` for Unix is quite simple, and it can be done when building the extension itself. On Windows, this is not possible to do. Instead, it must be done with a call to [SetDllDirectory](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setdlldirectoryw) before calling `LoadLibrary`, which SC is currently not doing in its `PlugIn_Load` function.

The current workaround, which is admittedly quite ugly, is to prompt the users of the extension to copy the .dll next to the `scsynth` executable, following Windows' way of shipping dlls.

Would it be worth to add these changes to allow developers to ship .dll files next to .scx?

Spacechild1:
## Purpose and Motivation

This PR fixes several issues regarding Server plugins. The motivation is to ensure a uniform behavior across platforms. Most importantly:
1. use the same plugin extension (`.scx`) on all platforms
2. allow to ship shared library dependencies (`.dll`, `.dylib`, `.so`) together with the plugin

I have added a simple test plugin in `testsuite/server/plugins` that implements a function in a shared library. I can now successfully load this plugin on all three platforms!

## Changes

- Linux: change the plugin extension from `.so` to `.scx`. This means people can finally ship shared library dependencies together with their UGen plugins, without having to put them in special folders. **This is a breaking change!** However, I have provided the following upgrade path:
    
    If a `.so` file is in the same directory as a `.scx` file, or in any subdirectory, we assume it is a dependency of the `.scx` file and therefore ignore it. Otherwise we try to open the `.so` file as a plugin and post a warning to the console. We also post a (one-time) message that tells the user what they can do to avoid the warning.
    
    We can remove this workaround in SC 3.16 or SC 3.17
    
- Windows: call `SetDllDirectory` with the current plugin directory so that plugins can find DLLs in the same directory

- Linux/scsynth: fix segfault when `dlopen` fails.

- scsynth: only unload plugins after all Nodes have been destroyed.

- Supernova: follow symlinks when loading plugins and ignore special folders to ensure the same behavior as scsynth.

Closes #6040 and #7342.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/scsynth/SC_Lib_Cintf.cpp
+++ b/server/scsynth/SC_Lib_Cintf.cpp
@@ -77,7 +77,11 @@ SC_LibCmd* gCmdArray[NUMBER_OF_COMMANDS];
 #endif
 
 void initMiscCommands();
-static bool PlugIn_LoadDir(const fs::path& dir, bool reportError, bool didFindPlugin = false);
+#ifdef LINUX_PLUGIN_WORKAROUND
+static bool PlugIn_LoadDir(const fs::path& dir, bool reportError, bool foundScxFile = false);
+#else
+static bool PlugIn_LoadDir(const fs::path& dir, bool reportError);
+#endif
 std::vector<void*> open_handles;
 #ifdef __APPLE__
 void read_section(const struct mach_header* mhp, unsigned long slide, const char* segname, const char* sectname) {
@@ -390,7 +394,13 @@ static bool PlugIn_Load(const fs::path& filename) {
 #endif // _WIN32
 }
 
-static bool PlugIn_LoadDir(const fs::path& dir, bool reportError, bool didFindPlugin) {
+#ifndef LINUX_PLUGIN_WORKAROUND
+static bool PlugIn_LoadDir(const fs::path& dir, bool reportError) {
+#else
+// if 'foundScxFile' is true, it means that we have alredy found a .scx file in one of the parent
+// directories. In this case, we treat all .so files as shared libraries and ignore them.
+static bool PlugIn_LoadDir(const fs::path& dir, bool reportError, bool foundScxFile) {
+#endif
     std::error_code ec;
     fs::directory_iterator iter(dir, fs::directory_options::follow_directory_symlink, ec);
     fs::directory_iterator end;
@@ -435,7 +445,7 @@ static bool PlugIn_LoadDir(const fs::path& dir, bool reportError, bool didFindPl
 
         if (fs::is_regular_file(iter->status()) && path.extension() == SC_PLUGIN_EXT) {
             PlugIn_Load(path);
-            didFindPlugin = true;
+            foundScxFile = true;
         }
 
         iter.increment(ec);
@@ -455,8 +465,8 @@ static bool PlugIn_LoadDir(const fs::path& dir, bool reportError, bool didFindPl
             const fs::path& path = entry.path();
 
             if (fs::is_directory(path)) {
-                PlugIn_LoadDir(path, reportError, didFindPlugin);
-            } else if (!didFindPlugin && path.extension() == ".so") {
+                PlugIn_LoadDir(path, reportError, foundScxFile);
+            } else if (!foundScxFile && path.extension() == ".so") {
                 // No .scx plugins were found in this directory or any parent directory -> assume the .so file is a
                 // plugin.
                 if (PlugIn_Load(path)) {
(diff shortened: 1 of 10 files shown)


server/scsynth/SC_Lib_Cintf.cpp
```diff
@@ -65,8 +66,18 @@ HashTable<struct PlugInCmd, Malloc>* gPlugInCmds = nullptr;
...
+// parent folders we simply ignore the .so file because we can assume it is a dependency.
+// Otherwise we load the .so file and post a warning.
+// NOTE: we should remove this workaround in the future!
+#ifdef __linux__
+#    define LINUX_PLUGIN_WORKAROUND
+#endif
+
 void initMiscCommands();
-static bool PlugIn_LoadDir(const fs::path& dir, bool reportError);
+static bool PlugIn_LoadDir(const fs::path& dir, bool reportError, bool didFindPlugin = false);
```
JordanHendersonMusic: `didFindPlugin`

I'm a bit confused as to what this argument does? Particularly since you write to it on line 438 when doing the linux work-around. Could it be renamed to something clearer?

If we are loading a directory then I'd expect a 'did find plugin' argument to be an out parameter.

Spacechild1: Maybe `didFindScxFile` is clearer?

JordanHendersonMusic: That implies it returns to the call sight whether or not it found a plugin, i.e., `bool& didFindScxFile`.

Is it `hasAlreadySeenFileWithSCXExtentionInDir`?

Spacechild1: > That implies it returns to the call sight whether or not it found a plugin

Why? The name says nothing about whether it's an in or out parameter. However, the fact that it's a value, and not a reference, makes this pretty clear.

> Is it hasAlreadySeenFileWithSCXExtentionInDir?

You really like long variable names :-D What if I just add a comment above the function? Then `didFindScxFile` should be sufficient, I think.

Spacechild1: Is it clearer with 5ae8786?

JordanHendersonMusic: > The name says nothing about whether it's an in or out parameter. 

I read it like ... 'did find plugin?'. 

If I'm right about what it does, then perhaps, `hasSeenScxFile` would work? 

Comment would work to.

Spacechild1: Aha, it's a grammer issue. `didFindScxFile` indeed reads like a question. So I guess `foundScxFile` should do it.

JordanHendersonMusic: > `foundScxFile`

Sounds good!

Spacechild1: What about 523f84b?

server/supernova/sc/sc_ugen_factory.cpp
```diff
@@ -215,22 +225,82 @@ bool sc_plugin_container::run_cmd_plugin(World* world, const char* name, struct
...
+    for (const auto& entry : fs::directory_iterator(dir, options)) {
+        if (fs::is_directory(entry.path())) {
+            load_plugin_folder(entry.path(), found_scx_file);
+        } else if (!found_scx_file && entry.path().extension() == ".so") {
+            // No .scx plugins were found in this directory or any parent directory -> assume the .so file is a plugin.
+            load_plugin(entry.path());
+            std::cout << "*** WARNING: '" << SC_Codecvt::path_to_utf8_str(entry.path()).c_str()
+                      << "': the .so extension has been deprecated!" << std::endl;
+
+            static bool didWarn = false;
```
JordanHendersonMusic, line 292: I'm not super familiar with supernova... is there any chance of multiple threads accessing this static at the same time?

Edit: it wouldn't be the end of the world because its just a bool and the write is a single instruction, worse that would happen is you get multiple posts, but it would be good to know.

server/scsynth/SC_Lib_Cintf.cpp
JordanHendersonMusic: Looks good, much clearer.

server/supernova/sc/sc_ugen_factory.cpp
Spacechild1: No. Only UGen calc functions can run in parallel.

JordanHendersonMusic: Thanks!

JordanHendersonMusic (approved):
Merge after CI.

PR #7361, opened 2026-02-04
Fix Pdup crash when n < 1 and add regression test

### LINKED ISSUE #7095: Pdup lack of check causes sclang crash
## Environment

* SuperCollider version: 3.14
* Operating system: MacOS

## Steps to reproduce

```supercollider
// Correct use of Pdup. Plays fine:
(
s.waitForBoot{

Pbind(
    \dur, Pdup(8, 1/8),
).play;

}
)
// If you get the two arguments switched by accident, sclang crashes:
(
s.waitForBoot{
Pbind(
    \dur, Pdup(1/8, 8),
).play;
}
)

```

The *new method of Pdup should check if `n` is a whole number before trying to construct, to avoid crashing.

TheSanjBot:
## Purpose and Motivation

This PR fixes a crash in `Pdup` that occurs when the repeat count stream yields
values smaller than 1, for example when the arguments are accidentally swapped
(`Pdup(1/8, 8)`).

The fix follows the solution discussed in #7095 by ensuring that events are only
duplicated when `n >= 1`, while preserving existing behavior for valid values.

A regression test is added to ensure this case no longer crashes `sclang`.

Fixes #7095.

## Types of changes

- Bug fix
- Tests

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation (not required)


SCClassLibrary/Common/Streams/FilterPatterns.sc
```diff
@@ -620,7 +620,8 @@ Pdup : FilterPattern {
 		while {
 			(inevent = stream.next(event)).notNil
 		} {
-			if((nn = nstream.next(event)).notNil) {
+			nn = nstream.next(event);
+			if(nn.notNil and: { nn >= 1 }) {
 				nn.abs.do {
```
HotwheelsSisyphus:
```suggestion
				nn.do {
```
(because the abs is unnecessary with the added test)

testsuite/classlibrary/TestPdup.sc
```diff
@@ -0,0 +1,16 @@
...
+        var err;
+
+        err = {
+            Pbind(
+                \dur, Pdup(1/8, 8)
+            ).asStream.next;
+        }.try { |e| e };
+
+        // If it crashed, err will be an Error
+        this.assert(err.isKindOf(Error).not);
```
HotwheelsSisyphus:
```suggestion
        this.assert(err.isKindOf(Error).not, "Pdup should not crash if the first argument is < 1");
```

HotwheelsSisyphus (approved):
Thanks! Looks good, the suggestions are more of the "might as well" type. Btw I think there's also an assertNoException method for the UnitTest class which does the same thing.

A question though (also to @supercollider/class-library team ): Should we make Pdup print or throw a warning if n < 1?

telephon:
> A question though (also to @supercollider/class-library team ): Should we make Pdup print or throw a warning if n < 1?

No, I think it should be a feature to be able to skip a pattern for n < 1.

telephon:
But it should be documented in the helpfile. @TheSanjBot could you do this, please? Thanks!

testsuite/classlibrary/TestPdup.sc
```diff
@@ -0,0 +1,16 @@
+TestPdup : UnitTest {
+
+    test_swappedArguments_doesNotCrash {
+        var err;
+
+        err = {
+            Pbind(
+                \dur, Pdup(1/8, 8)
+            ).asStream.next;
+        }.try { |e| e };
```
JordanHendersonMusic: Oooooo... returning a value from try is long term broken. You need to create a variable and assign to it inside the block instead.

Sorry about this, it is a really complex thing to fix!

JordanHendersonMusic (changes requested):

testsuite/classlibrary/TestPdup.sc
TheSanjBot: > Oooooo... returning a value from try is long term broken. You need to create a variable and assign to it inside the block instead.

I’ve updated the test to avoid relying on the return value of try.

JordanHendersonMusic:
Thanks! Looks good to me, @telephon all good?

telephon:
Ah maybe the example usage from https://github.com/supercollider/supercollider/issues/7095#issuecomment-3207662013 could be added as a test or/and to the help.

testsuite/classlibrary/TestPdup.sc
```diff
@@ -1,16 +1,20 @@
 TestPdup : UnitTest {
 
     test_swappedArguments_doesNotCrash {
-        var err;
+        var err = nil;
 
-        err = {
+        {
```
HotwheelsSisyphus: maybe this is quicker to read:
```supercollider
var func = {     
    Pbind(\dur, Pdup(1/8, 8))
        .asStream.next;
};
this.assertNoException(func, "Pdup should not crash if the first argument is < 1")
```

HotwheelsSisyphus: simile for the other test

TheSanjBot: > simile for the other test

Thanks — updated the test to use assertNoException. There’s only one test in this file currently, so this should cover it.

JordanHendersonMusic (approved):

PR #7363, opened 2026-02-05
scide: make urls clickable in the post window

### LINKED ISSUE #7362: ScIDE: make post window recognise URI formatted lines and open them nicely on click
## Motivation

Clickable links in the post window:

```
file://path/blah.scd:32:10
```

Opens file at line 32, character 10

```
scdoc://Classes/Array:somemethod...
```

Internally, these are turned into sclang code in qt and passed back to the running sclang instance by passing the protocol, the address, and extra information like so...

```
file://path/blah.scd:32:10

... on click turns into...

PostWindowURIHandler(\file, "path/blah.scd", 32, 10)
```

One can then register their own URI handlers 
```
PostWindowURIHandler.register(\hello, {
   |address... other|
   "Got a hello URI %".format(address).postln
});

"hello://fromThePostWindow".postln;
//^ then you click on the highlighted link in the post window
-> Got a hello URI fromThePostWindow
```

---

There are questions, like, should we have a default system handler? This would make things like mailto:// and https:// work - should the latter be opened by qt browser or the systems?

... but I don't really care about this for now and am happy to leave to unimplemented, being able to link to help documentation in compile error messages (and runtime) is the motivation here.

This should also follow the standard of the uri spec, with all the names of the uri being passed correctly to sclang. I need to add that here, but `QUrl` does this for us, and the spec is on Wikipedia.

## Plan for Implementation
 
Shouldn't be hard. I'll have a look when/if I find the time.

Could use `QUrl` class to do the parsing of the string in c++ and the `URI` class in scdoc.sc

JordanHendersonMusic:
## Purpose and Motivation
Closes #7362

```
"scdoc://Classes/Array.html#-put"
```

dddd

When you click on the link it takes you to the `put` method of `Array`.

This also works on files...

```
file://Some/Abs/Path/test.scd:10
```
which will open the file and set the cursor to the tenth character.

Additionally, one can register their url handler simply.

```
PostWindowURLHandler.register(\bang, {|url| ...  });

"bang://blah"
```

This only works in ScIDE.

I ***think*** the URL is in an acceptable standard format? We use `file://` internally to open the help docs so that is what I based it off.

I primarily intent to use this for clickable links in error messages.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -15,11 +15,13 @@ ScIDE {
 		Class.initClassTree(PostWindowURLHandler);
 
 		PostWindowURLHandler.register(\scdoc, {|url|
-			ScIDE.openHelpUrl("file://" ++ SCDoc.helpTargetDir +/+ url[8..])
+			var prefix = "scdoc://";
+			ScIDE.openHelpUrl("file://" ++ SCDoc.helpTargetDir +/+ url[prefix.size..])
 		});
 
 		PostWindowURLHandler.register(\file, {|url|
-			var parts = url[7..].split($:);
+			var prefix = "file://";
+			var parts = url[prefix.size..].split($:);
 			{ Document.open("/" ++ parts[0], parts[1] ?? { 0 }, parts[2] ?? [0]) }.fork(AppClock)
 		});
 
@@ -457,7 +459,7 @@ PostWindowURLHandler {
 	}
 
 	*prDefault { |url|
-		"PostWindowURLHandler does not know how to handle the URL '%'".format(url).postln
+		" PostWindowURLHandler does not know how to handle the URL '%'".format(url).warn.postln
 	}
 }
 
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -11,6 +11,17 @@ ScIDE {
 		subListSorter = { | a b | a[0].perform('<', b[0]) };
 
 		Class.initClassTree(Server);
+		Class.initClassTree(SCDoc);
+		Class.initClassTree(PostWindowURLHandler);
+
+		PostWindowURLHandler.register(\scdoc, {|url|
+			ScIDE.openHelpUrl("file://" ++ SCDoc.helpTargetDir +/+ url[8..])
+		});
+
+		PostWindowURLHandler.register(\file, {|url|
+			var parts = url[7..].split($:);
+			{ Document.open("/" ++ parts[0], parts[1] ?? { 0 }, parts[2] ?? [0]) }.fork(AppClock)
+		});
 
 		StartUp.add {
 			if (this.connected) {
@@ -423,6 +434,30 @@ ScIDE {
 	*prConnect {|ideName|
 		_ScIDE_Connect
 		this.primitiveFailed
+	}
+}
+
+PostWindowURLHandler {
+	classvar registered;
+
+	*initClass {
+		Class.initClassTree(IdentityDictionary);
+		registered = IdentityDictionary();
+	}
+
+	*register { |name, fn| registered[name] = fn }
+	
+	*new { |scheme, url|
+		var found = registered[scheme.asSymbol];
+		^if(found.isNil){
+			this.prDefault(url)
+		} {
+			found.(url)
+		}
+	}
+
+	*prDefault { |url|
+		"PostWindowURLHandler does not know how to handle the URL '%'".format(url).postln
 	}
 }
 
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -459,7 +459,7 @@ PostWindowURLHandler {
 	}
 
 	*prDefault { |url|
-		" PostWindowURLHandler does not know how to handle the URL '%'".format(url).warn.postln
+		"PostWindowURLHandler does not know how to handle the URL '%'".format(url).warn
 	}
 }
 
--- a/editors/sc-ide/widgets/main_window.cpp
+++ b/editors/sc-ide/widgets/main_window.cpp
@@ -166,6 +166,8 @@ MainWindow::MainWindow(Main* main): mMain(main), mClockLabel(0), mDocDialog(0) {
     connect(this, &MainWindow::evaluateCode, main->scProcess(), &ScProcess::evaluateCode);
     // Interpreter: post output
     connect(main->scProcess(), &ScProcess::scPost, mPostDocklet->mPostWindow, &PostWindow::post);
+
+    connect(mPostDocklet->mPostWindow, &PostWindow::handleClickedURL, main->scProcess(), &ScProcess::evaluateCode);
     // Interpreter: monitor running state
     connect(main->scProcess(), &ScProcess::stateChanged, this, &MainWindow::onInterpreterStateChanged);
     // Interpreter: forward status messages
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -351,6 +351,22 @@ void PostWindow::wheelEvent(QWheelEvent* e) {
 #endif
 }
 
+void PostWindow::mousePressEvent(QMouseEvent* e) {
+    clickedAnchor = (e->button() & Qt::LeftButton) ? anchorAt(e->pos()) : QString();
+    QPlainTextEdit::mousePressEvent(e);
+}
+
+void PostWindow::mouseReleaseEvent(QMouseEvent* e) {
+    // Make sure that the release event was over the same anchor it started on.
+    if (e->button() & Qt::LeftButton && !clickedAnchor.isEmpty() && anchorAt(e->pos()) == clickedAnchor) {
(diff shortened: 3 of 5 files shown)


editors/sc-ide/widgets/post_window.cpp
```diff
@@ -188,29 +191,56 @@ QString PostWindow::symbolUnderCursor() {
...
+                    }
+
+                    // Put space back in, if not last word.
+                    if (w + 1 == words_count) {
+                        break;
+                    } else {
+                        cursor.insertText(" ", line_format);
+                    }
+                }
+            } else {
```
capital-G: Just a minor note: I always prefer to have the trivial/shorter case the beginning if case such that the matching if condition is closer to the code

editors/sc-ide/widgets/main_window.cpp
```diff
@@ -166,6 +166,8 @@ MainWindow::MainWindow(Main* main): mMain(main), mClockLabel(0), mDocDialog(0) {
     connect(this, &MainWindow::evaluateCode, main->scProcess(), &ScProcess::evaluateCode);
     // Interpreter: post output
     connect(main->scProcess(), &ScProcess::scPost, mPostDocklet->mPostWindow, &PostWindow::post);
+
+    connect(mPostDocklet->mPostWindow, &PostWindow::handleClickedURL, main->scProcess(), &ScProcess::evaluateCode);
```
capital-G, line 170: I think it would be a bit clearer if `handleClickedURL` would call `Main::scProcess()->evaluateCode()` directly, instead of re-directing this call through a postWindow signal? At least I was initially a bit confused where `handleClickedURL` gets implemented.

SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -426,6 +437,30 @@ ScIDE {
 	}
 }
 
+PostWindowURLHandler {
```
capital-G, line 442: Needs documentation ;)

SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -11,6 +11,17 @@ ScIDE {
...
 		Class.initClassTree(Server);
+		Class.initClassTree(SCDoc);
+		Class.initClassTree(PostWindowURLHandler);
+
+		PostWindowURLHandler.register(\scdoc, {|url|
+			ScIDE.openHelpUrl("file://" ++ SCDoc.helpTargetDir +/+ url[8..])
+		});
+
+		PostWindowURLHandler.register(\file, {|url|
+			var parts = url[7..].split($:);
```
capital-G: its a bit obvious in context, but it wouldn't hurt to add a comment what the hardcoded `7` is actually removing here

SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -11,6 +11,17 @@ ScIDE {
 		subListSorter = { | a b | a[0].perform('<', b[0]) };
 
 		Class.initClassTree(Server);
+		Class.initClassTree(SCDoc);
+		Class.initClassTree(PostWindowURLHandler);
+
+		PostWindowURLHandler.register(\scdoc, {|url|
```
capital-G, line 17: why does the ScIDE need to know of postWindowURLHandler? Wouldn't it be sufficient to init this in `PostWindowURLHandler` itself, such that ScIDE doesn't need to know anything about the class?

SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -426,6 +437,30 @@ ScIDE {
...
+		var found = registered[scheme.asSymbol];
+		^if(found.isNil){
+			this.prDefault(url)
+		} {
+			found.(url)
+		}
+	}
+
+	*prDefault { |url|
+		"PostWindowURLHandler does not know how to handle the URL '%'".format(url).postln
```
capital-G: Should we prefix this with `error` or use `warn`?

editors/sc-ide/widgets/post_window.hpp
```diff
@@ -61,8 +62,26 @@ class PostWindow : public QPlainTextEdit {
     QSize minimumSizeHint() const { return QSize(50, 50); }
     QString symbolUnderCursor();
 
+    void mousePressEvent(QMouseEvent* e) override {
+        clickedAnchor = (e->button() & Qt::LeftButton) ? anchorAt(e->pos()) : QString();
```
capital-G: maybe add a comment why we need to store this? It is explained in `mouseReleaseEvent`, b ut maybe it seems suitable to add a comment above `clickedAnchor` such that an IDE can display the comment when hovering, making it a bit clearer when looking at this 10 years from now.

capital-G (changes requested):
Thanks for this neat feature Jordan - I think there are some minor things and questions.

SCClassLibrary/scide_scqt/ScIDE.sc
JordanHendersonMusic: Originally I was going to have the handlered shared between all ides, but I couldn't find a directory to put the class in.

editors/sc-ide/widgets/main_window.cpp
JordanHendersonMusic: I haven't really dug into the qr stuff, and I know you have so if you think this should happen then I'm fine with it, but I find using globals quite confusing in signals. I find it much easier to see how the program works when you can see all the connections in one place. 

Perhaps the name of this signal could be better though!?

editors/sc-ide/widgets/post_window.hpp
JordanHendersonMusic: Okay done

SCClassLibrary/scide_scqt/ScIDE.sc
JordanHendersonMusic: Okay I've added some

SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -457,7 +459,7 @@ PostWindowURLHandler {
 	}
 
 	*prDefault { |url|
-		"PostWindowURLHandler does not know how to handle the URL '%'".format(url).postln
+		" PostWindowURLHandler does not know how to handle the URL '%'".format(url).warn.postln
```
capital-G: Why the whitespace and warn and postln?

```suggestion
		"PostWindowURLHandler does not know how to handle the URL '%'".format(url).warn
```

capital-G (approved):
~~There is one small thing left - once this has been addressed it is ready to be squashed and merged.~~

This is almost ready to be merged, but there are still things which should be addressed.

SCClassLibrary/scide_scqt/ScIDE.sc
capital-G: Couldn't we add something like this to the class init of `PostWindowURLHandler`?

```supercollider
if(Platform.ideName==\scqt, {
  registerScDoc();
});
```

capital-G (changes requested):

editors/sc-ide/widgets/main_window.cpp
capital-G: I see. We currently don't enforce/follow this, the signals are already spread across the IDE.

I'd prefer to move this closer to the actual implementation such that the whole post window logic is kept together and we simply accept that any widget can evaluate sclang code.
Ideally the IDE helps here to find any calls to evaluate in case we need to refactor it.

Up to you, thats just my 2ct :)

SCClassLibrary/scide_scqt/ScIDE.sc
JordanHendersonMusic: This won't work because `SCIde` needs to be inited before `PostWindowURLHandler`, but we don't know which IDE is being used, therefore we can't call `Class.initClassTree`.

JordanHendersonMusic:
@capital-G do you think this is alright now? I'm waiting on this to implement a bunch of compiler errors in other PRs.

capital-G (approved):
LGTM - feel free to merge!

dyfer:
Where do we emit `file://path` links? 

It seems that `file://path` is actually [incorrect](https://en.wikipedia.org/wiki/File_URI_scheme#Number_of_slash_characters); it should either be `file:///path` or `file:/path` for local paths.

JordanHendersonMusic:
> Where do we emit file://path links?

In other PRs. The compiler does it mostly. 

> It seems that file://path is actually [incorrect](https://en.wikipedia.org/wiki/File_URI_scheme#Number_of_slash_characters); it should either be file:///path or file:/path for local paths.

I didn't know that, both look very wrong to my eye! I don't oppose changing it to `:/` if you think it is important?

dyfer:
> I didn't know that, both look very wrong to my eye! I don't oppose changing it to `:/` if you think it is important?

I think both correct ways should be handled, and they are not currently. I haven't done enough research to suggest one or the other, but if both are correct, then I think both should be handled.

I don't oppose to _also_ support the seemingly incorrect `file://` for clickable links, but we should should emit proper syntax ourselves. 

Personally `file:/` seems more suspicious to me, so I'd suggest emitting `file:///` ourselves. But again, this is not based on any actual knowledge/experience, only on occasional occurrence of these in the wild.

PR #7366, opened 2026-02-08
sclang: large reworking of errors and large refactor of lang/

JordanHendersonMusic:
This pr will be broken into multiple smaller ones where possible to make reviewing easier. This might take a while and somethings have to come in large chunks. 

One thing I've been keen on doing is not adding to the technical debt we have (e.g., excessive use of global mutable variables) but to do this properly.

I've put this here to show where this is going.

Part one
#7394 - rewrite of lexer so token location is correct emitted, provides many other benefits, including a 'one true' lexer, extendible with templates, so we don't need the (at least) 3 currently in this repo. Also opens up the possibility of unicode support, string interpolation, dedicated DSL support...etc.

Part two
Grammar refactor, move lexer-parser communication to use `yylval` and `yyloc` as our current implementation relies on unsafe globals, also refactor parse node constructors so they are safe.

Part three
Rework main interpret entry point to accept a filename, location within the file, along with the source, this will mean IDEs need to update. Store source code of all functions with the function def, not just closed functions.

Part four 
Errors and bytecode serialisation on SC side. This will mostly be supercollider code. Potentially bring in unicode to the class library in some way.

## Purpose and Motivation

Supercollider's error messages are bad, this PR put backtraces in and highlights the code. This is something I've been trying to do for a long time now and have only just become familiar enough with the code base to be able to get this to work — it has been quite a bit of work!

This is the current state of this PR.

```supercollider
(
~f = { |i, action| 
	if (i == 0) { 
		action.() 
	} { 
		~f.(i - 1, action) 
	} 
};

~f.(4, { 3.asdf })
)
```

***This shows the most verbose output, there are options to disable parts of the trace.***

```
ERROR: a DoesNotUnderstand error has occurred.
3 did not understand the message 'asdf'

Call Stack
────────────
  ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
  ┃ 1: Function                                                          ┃
  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
       FunctionDef:
       ─────────────────────────────────────────────────────────
         Selectors: [a FunctionDef, a FunctionDef]
         Constants: [f]
         Bytecodes:
         ──────┬───────────────────────────────────────────────────────
           7   │ PushSpecialNumber One 
           8   │ PushLiteralX [Index 1] 
           9   │ TailCallReturnFromFunction 
           10 →│ SendSpecialMsg 3 [SpecialSelectors value] 

     SourceCode:
     ─────┬───────────────────────────────────────────────────────
       1  │ (
       2  │ ~f = { |i, action| 
       3  │ 	if (i == 0) { 
       4  │ 		action.() 
       5  │ 	} { 
       6  │ 		~f.(i - 1, action) 
       7  │ 	} 
       8  │ };
       9  │ 
       10→│ ~f.(1, { 3.asdf })
          │ ^^^^^^^^^^^^^^^^^^
  ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
  ┃ 2: Function                                                          ┃
  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
     Arguments: i=1, action=a Function
       FunctionDef:
       ─────────────────────────────────────────────────────────
         Constants: [f]
         Bytecodes:
         ──────┬───────────────────────────────────────────────────────
           12  │ PushOneAndSubtract 
           13  │ PushTempZeroVar [VariableIndex, 1] 
           14  │ TailCallReturnFromFunction 
           15 →│ SendSpecialMsg 3 [SpecialSelectors value] 

     SourceCode:
     ─────┬───────────────────────────────────────────────────────
       2  │ ~f = { |i, action| 
       3  │ 	if (i == 0) { 
       4  │ 		action.() 
       5  │ 	} { 
       6 →│ 		~f.(i - 1, action) 
          │ 		^^^^^^^^^^^^^^^^^^
  ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
  ┃ 3: Function                                                          ┃
  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
     Arguments: i=0, action=a Function
       FunctionDef:
       ─────────────────────────────────────────────────────────
         Constants: [f]
         Bytecodes:
         ──────┬───────────────────────────────────────────────────────
           4   │ JumpIfFalse [Uint [16, 1, 0], Uint [16, 0, 7]] 
           5   │ PushTempZeroVar [VariableIndex, 1] 
           6   │ TailCallReturnFromFunction 
           7  →│ SendSpecialMsg 1 [SpecialSelectors value] 

     SourceCode:
     ─────┬───────────────────────────────────────────────────────
       2  │ ~f = { |i, action| 
       3  │ 	if (i == 0) { 
       4 →│ 		action.() 
          │ 		^^^^^^^^^
  ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
  ┃ 4: Function                                                          ┃
  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
       FunctionDef:
       ─────────────────────────────────────────────────────────
         Selectors: [asdf]
         Bytecodes:
         ──────┬───────────────────────────────────────────────────────
           1   │ PushInteger8 [Int [8, 0, 3]] 
           2   │ TailCallReturnFromFunction 
           3  →│ SendMsg 1 [SelectorIndex 0] 

     SourceCode:
     ──────┬───────────────────────────────────────────────────────
       10 →│ ~f.(1, { 3.asdf }
           │          ^^^^^^
  ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
  ┃ 5: Object:doesNotUnderstand                                          ┃
  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
     Arguments: this=3, selector=asdf ...args=[], kwargs=[]
       FunctionDef:
       ─────────────────────────────────────────────────────────
         Selectors: [throw, DoesNotUnderstandError]
         Bytecodes:
         ──────┬───────────────────────────────────────────────────────
           3   │ PushTempZeroVar [VariableIndex, 1] 
           4   │ PushTempZeroVar [VariableIndex, 2] 
           5   │ PushTempZeroVar [VariableIndex, 3] 
           6  →│ SendSpecialMsg 5 [SpecialSelectors new] 

     SourceCode:
     ───────┬───────────────────────────────────────────────────────
       351  │ 	doesNotUnderstand { |selector ...args, kwargs|
       352 →│ 		DoesNotUnderstandError(this, selector, args, kwargs).throw;
            │ 		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

ERROR: '3' did not understand the message 'asdf'

Messages with a similar name understood by the receiver:
	asin
	absdif
	as
	asRef
```

FYI, if you read this on mobile, the formatting is messed up. I've used unicode arrows and boxes, don't know if that is a good idea, but it looks good in the post window!

There are many many formatting issues to be resolved in the error printing, but it appears to be stable.

## Types of changes

- All parsenodes can now calculate their location in the text.
- FunctionDefs now store the entire source file, not just their little piece.
- Bytecodes now emit their location in the source file into the function def.
- Rename PyrBlock to PyrFunctionDef because blocks are syntax and not all of them are valid function defs (inlined blocks are never turned into function defs), there was also ambiguity with the naming as the class library called 'blocks' 'functiondefs'.
- Rework the error printing of doesnotunderstand to print backtraces highlight the source code.
- Make doesNotUnderstand calls not be tail called, otherwise we can't see the code the user has called.
- rename the lexer character variables because they were ambiguous as to what they actually refereed to.
- completely rework the relationship between lexer and parser, removing incorrect use of globals in favour of `yylval` and `yyloc`.
- remove some unnecessary grammar rules.
- refactor parts of the compiler that were too complicated to read.
- renaming of a few variables as their names no longer made sense or were ambiguous.

## To-do list

much...

Get the clickable links in the terminal PR merged.

Get all functiondefs to know their file name PR merged.

Implement clickable links to the source code.

Rework compiler error messages so they look similar-ish to the run time errors.


capital-G:
This is such a great feature - thanks so much Jordan! <3

One thing I asked myself: Should we "reverse" the call stack? Assuming that one runs into an error - currently I have to scroll up to the start position to understand which context I have and start to scroll down from there.
If we reverse the order, I could just simply scroll up and incrementally understand the error until I am reaching the end - this has the advantage that I don't have to scroll to a certain position in order to start reading the error.

What do you think about this?

JordanHendersonMusic:
Ah, that is what I have done I think! Actually, I'm a little confused as to which way up you mean? are you saying the old way way wrong, and my way is right, or my way is wrong and the old way was right?

The post window, and most terminals, always autoscroll to the bottom, this way you read it from the bottom to top —looks odd on github, but I think it makes sense in the IDE. 

I'm also going to have clickable links to the source code soon.

This is what you actually see
```
	6: Function
───────────────┬───────────────────────────────────────────────────────
		11→    │~f.(4, { 3.asdf })
		       │        ^^^^^^^


ERROR: '3' did not understand the message 'asdf'

Messages with a similar name understood by the receiver:
	asin
	absdif
	as
	asRef
```

capital-G:
Oh sorry - I somehow misread that - I got confused by the "0" stack which just shows the whole function - but indeed you are right. I should get more sleep before commenting on PRs ;)

lang/LangSource/PyrParseNode.cpp
```diff
@@ -4163,10 +4148,27 @@ void PyrBlockNode::compile(PyrSlot* slotResult) {
...
 
-    if ((!gFunctionCantBeClosed && gFunctionHighestExternalRef == 0) || mIsTopLevel)
+    if ((!gFunctionCantBeClosed && gFunctionHighestExternalRef == 0) || mIsTopLevel) {
         SetNil(&fdef->contextDef);
+        fdef->isClosed = PyrSlot::make(true);
 
-    gCompilingBlock = prevBlock;
+        // The logic here for getting the true start of the function is flawed as it requires the function begin with a
+        // curly and is set in the lexer using a stack contains opening and closing bracket positions. This doesn't work
+        // because block nodes do not have to begin with a curlies, a file can just contain a arguments, vars, or just
```
telephon: a file can just contain arguments, vars,

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -460,8 +460,10 @@ FunctionDef {
 	var fullSourceCode; // this is the complete file! Use codeCharacterLocations to index into it
 	var <codeCharacterLocations; // array of ints, twice as many as in code, used to represent the start of the byte code in and the end.
 	var <codeSizes; // size in bytes of each byte code.
+	var <isClosed;
+	var closedSourceCode;
```
telephon: maybe better call it sourceCodeClosed

telephon:
Beautiful, this has been an open issue for ages!

telephon:
```
	6: Function
───────────────┬───────────────────────────────────────────────────────
		11→    │ ~f.(4, { 3.asdf })
		       │        ^^^^^^^
```

cosmetics: add one space after the pipe before the code

JordanHendersonMusic:
This isn't going to work with unicode unless we make unicode a part of the class library, or we do all the formatting in C++, which is an option.

JordanHendersonMusic:
I've added automatic bytecode destructuring and conversion into a sclang friendly format so you can walk it. The formatting of this is currently a work in progress, but all the information is there.

Also there are global options to change the output
```
DebugFramePrinter {
	classvar <showSourceCode = true;
	classvar <showBytecode = false;
	classvar <reverseStack = true;
	classvar <stackTraceMax = 10;
...
```

SCClassLibrary/Common/Core/Kernel.sc
```diff
@@ -463,6 +463,8 @@ FunctionDef {
 	var <isClosed;
 	var closedSourceCode;
 
+	// TODO: there is a difference between getting the code with the whitespace before it, used for printing errors, and to re-create the function.
+	// Do we really need to use the source code to recreate a function?
```
telephon: This was the purpose of keeping the sourceCode.

telephon: ... and of course for introspection, having the sourceCode with the object.

JordanHendersonMusic: You can always recreate a 'closed' function from its bytecodes, selector, and constants array, in other words, you can just serialise the functiondef directly. 

The only time you wouldn't want to do this is when serialising this to disk, as it exposed interpreter details, but in that case, you might as well save the file?


Also, what is and isn't a closed function is suprising:

```
(
var nl;
{ nl }.isClosed // false
)

(
var nl;
#{ nl }.isClosed // true
)

(
{ nl }.isClosed // true
)
```

This is because closed functions are compiled inside of the interpreter class, meaning they have access too all the `Interpreter` variables and `Object` classvars and consts. 

```
#{ preProcess = nil };
```

Re introspection, the way in which supercollider assigned the source code only work with functions made with explicit curly braces.
```
( _ + 1).isClosed // false
```
despite the fact that
```
{ |a| a + 1 }.isClosed // true
```

telephon: > Re introspection, the way in which supercollider assigned the source code only work with functions made with explicit curly braces.

This is probably an omission.

The methods `asCompileString` (or `cs`) are often used to recreate the source code, both for reading and reinterpretation. The limitations are known, but can be avoided if necessary by creating an outer scope.

```
( _ + 1).cs // -> { | args | "open Function" }
{( _ + 1)}.cs // -> {( _ + 1)}
```

JordanHendersonMusic: Okay I've been at this all day, trying to make  `( _ + 1).def.sourceCode == "( _ + 1)` and it doesn't work.

`a = _ + 1; a.def.sourceCode == "_ + 1"` this is done.

JordanHendersonMusic:
I don't know how this is going to be reviewed...

Perhaps it would be useful if I talked about how the lexer/parser/compiler worked at one of the dev meetings, why the old approach using globals to pass information between them was incorrect due to the lookahead nature of LA(1)LR parsing and how bison gives us the tools to solve it.

SCClassLibrary/Common/Core/Kernel.sc
JordanHendersonMusic: Took two days.... but now ` (    _ +    1).def.sourceCode == "(    _ +    1)"`

JordanHendersonMusic:
Okay so everything I have done so far has been backwards compatible, however, the main exception classes have quite poor interfaces as all the methods all print or otherwise have side effects, rather than return a value. This means we can't get the result of an error, making it hard to test and impossible to generate a big report of what all the errors look like. What I'd like to be able to do is have the CI create a document with the current state of the error messages in. This way, we can track how they improve (or regress) over time. I also intend to do this for compiler errors.

How do we feel about a breaking change in the `Exception` classes? @supercollider/class-library

SCClassLibrary/Common/Core/Kernel.sc
JordanHendersonMusic: FYI sourceCode should work again, and now works on all functions, and anything that generates a functiondef/method.

lang/LangSource/PyrSlot.h
```diff
@@ -459,6 +463,49 @@ union PyrSlot {
...
+        static_assert(I < NumFields);
+        static_assert(I != 0);
+        return m_fields[I];
+    }
+
+    // not marking this private otherwise we have to write constructors, just don't touch this.
+    BackingType m_fields[NumFields] {}; // zeros these.
+};
+
+
```
JordanHendersonMusic: TODO: just noticed this and the reimplementation of the METHRAW isn't used, that can be another PR.

HotwheelsSisyphus:
(Maybe tangentially relevant here, but not sure if realistically fixable or worth the hassle: #7430 - where any stderr coming back to sclang from a unixCmd automatically becomes an error, even if it is actually of a different severity level)

JordanHendersonMusic:
It's not quite relevant but is a really good point! Generally how we deal with IO in sc isnt ideal. We follow the Smalltalk idea where the whole language is an operating system - so it's capable of everything. It would be much nicer if you set these buffers up at launch globally, similar to how you define POSIX like functions in javascript when doing wasm. Top down rather than bottom up. But that's a big rewrite and all primitives - maybe one day!

PR #7368, opened 2026-02-08
Document non-standard behavior for trunc and div; specify standard methods

HotwheelsSisyphus:
EDIT: seemed more sensible to close this one and re-open from another branch.

<details><summary>deleted from older PR description</summary>
<p>

This PR adds three methods.
(I've gone ahead and started down this route, but I don't mind reverting it if necessary.)

The other suggestions included in the doc are:
- `divide`: The name is in analogy to trunc/truncate, where div/trunc are the supercollider versions, divide and truncate the "standard" versions. 

- `roundDown` (optional alias for the old trunc; `roundUp` exists already). This is maybe unnecessary but it is probably a better name for what sc's trunc actually does. 

Since these haven't been discussed it would be cool to have some kind of feedback whether this is a good idea or not. If not, I can limit this PR to trunc/ate.
At any rate these are additions, hence should be non-breaking.

TODO:
Implementation. The "dones" are tentative, in that I expect to have to do some debugging and probably overlooked a lot of things.
If I understand correctly how this works, I'Il need to:
- add methods in the class library (done)
- define the primitive in `PyrPrimitive.cpp` (done)
- assign the opcode in `PyrMathOps.cpp` (done)
- (and possibly define a `sc_truncate` in `SC_InlineBinaryOp.h`, if it's to be in analogy to sc's current `trunc`. )
- UGens  (done)
- and then actually get it to build and work... Currently, all the builds with ctest fail, etc. 

</p>
</details>

### DIFF
--- a/HelpSource/Classes/SimpleNumber.schelp
+++ b/HelpSource/Classes/SimpleNumber.schelp
@@ -41,7 +41,18 @@ method:: mod
 Modulo
 
 method:: div
-Integer Division
+Divide teletype::this:: by teletype::aNumber:: and rounds to next-lower integer.
+For positive integers strong::only::, this is equivalent to standard integer division.
+Warning::
+This is a non-standard implementation of integer division,
+which has been preserved in SuperCollider for backward compatibility reasons,
+and is equivalent to code::floor(this/aNumber)::.
+
+For standard integer division, use code::truncate(this/aNumber):: or link::#-divide::
+::
+
+method:: divide
+Standard integer division, divides teletype::this:: by teletype::aNumber:: and rounds towards 0.
 
 method:: **
 Exponentiation
@@ -71,7 +82,10 @@ method:: round
 Round to multiple of aNumber
 
 method:: roundUp
-Round up to a multiple of aNumber. For roundDown use link::Classes/SimpleNumber#-trunc#trunc::.
+Round up to a multiple of aNumber.
+
+method:: roundDown
+Round down to a multiple of aNumber. 
 
 method:: smallButNotZero
 Check if the value is closer to zero than a threshold, but not zero.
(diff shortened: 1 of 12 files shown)


HotwheelsSisyphus:
ok, this builds (at least in some cases and locally for me)  but sclang crashes instantly with a segfault. I figure there's probably something about the opCodes that I did wrong (or something that I should have done and didn't do, etc.). 
(I can't pretend to really understand how this works, most of these changes were just modidfied copypasting. it was worth a try...)

@JordanHendersonMusic, cam you spot anything obvious at first glance?

JordanHendersonMusic:
My advice... don't touch those enums or anything labeled 'special'... best to think of it like a carefully balanced stack of fine china. 

Hyperbole aside.. the numerical value of the enums are used elsewhere in the code as numbers. So changing the order or amount can easily cause erroneous behaviour.

You can implement a new normal primitive in sclang to get the behaviour you want. The server stuff is best done with a new ugen.

HotwheelsSisyphus:
Thanks for the advice... yeah, I just noticed those opcode numbers were hardcoded...
so yes, this is definitelly beyond my abilities. Interesting though. 

I just figured that truncation is such a basic thing that would best be dealt with at the source rather than doing some class library side fix. That seems a bit roundabout, e.g. ` { floor(abs(x)) * sign(x) }` is just more operations than trunc. 

I guess for my local experiments I'll see if I can actually replace the old trunc/div primitives with "corrected" versions and build the old ones on the class library side.

HotwheelsSisyphus:
<details><summary>resolved</summary>
<p>
I'm messing with this  on and off, primitives work fine but not so much UGens, specifically adding extra UGens rather than changing the existing ones. 

NOTE: this question is based on the code in another branch of mine, [this one](https://github.com/supercollider/supercollider/compare/develop...HotwheelsSisyphus:supercollider:trunc-experiments).
 Look at the changes there to `Cmakelists.txt` and the added `truncate.cpp`. At present I don't even care how well the UGen even works... but I just can't get it to install to begin with it.


If I try to add `Truncate` as an additional UGen, the library compiles and the server boots fine, 
`exception in GraphDef_Recv: UGen 'Truncate' not installed`
I suspect I'm doing something stupid/wrong in the CMakeLists.txt, because I haven't ever touched that stuff before. I tried to model the approach a bit on the most recent UGen that was added, @capital-G's LinkUGen (#6947). 
</p>
</details> 

EDIT: nvm, I was just running the wrong scsynth binary, because on linux `Server.program` is by default set to `exec scsynth` so it wasn't using the one that actually _had_ those plugins. Well. good to know.

PR #7372, opened 2026-02-18
FreqScope singleton behavior: change test in *new so instance can be accessed after creation

HotwheelsSisyphus:
newest update on this PR, 20/02/2026:

Keeping this PR minimal for now; might do more FreqScope changes later on but that seems more complicated, one would have to refactor the new method quite a bit. 

So the actual changes are:

1.  classvar `instance`
2. `getInstance` method 
3. `new` posts a warning when `scopeOpen = true`.

FreqScope is meant to only have one instance, but after creation of that one instance, there is no way to programmatically  access that instance if it hasn't been assigned to something when it was first instantiated. This PR  changes the *new method to ~~return the single instance not only when it is created, but also if it has already been created earlier~~ post a warning that points user to the getInstance method if an instance of FreqScope already exists.

## Purpose and Motivation
Problem:
```
Freqscope() // open the freqscope but forget to assign it (or its instance vars)
// try to assign it now:
f = Freqscope();
f; // nil
```
In these cases, one has to restart the interpreter to be able to access the scope.  (Maybe there are other ways, but they're not terribly obvious to me).

## Types of changes

See code:
Previously *new checked whether classvar `scopeOpen` is True, and if it was, did nothing. If it was not true, it built the instance, returned it, and set `scopeOpen` to True.
 That's good for not making lots of new Freqscopes, but bad for the above cases. 

<details><summary>questions resolved in discussion</summary>
<p>

With this PR, this is changed to simply return the instance if `scopeOpen` is true. This won't create new freqscopes but lets you use the .new method to access the existing instance.

## Questions (that of course occurred to me only after the fact)

Is it even reasonable to use the *new method (i.e., constructor) for this (i.e., returning the isntance if it exists) ? 

I checked with the [wiki](https://en.wikipedia.org/wiki/Singleton_pattern) article on Singletons, and the examples in there have public .getInstance methods that call the private constructors if necessary and otherwise return the instance. This is a bit like what *new is doing in the present case, except it's called `new` rather than `getInstance`.

Should the `classvar singleInstance` have a getter method (i.e., be `classvar <singleInstance`)?

</p>
</details>

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
@@ -333,163 +333,163 @@ FreqScope {
 				"Only one instance of FreqScope can exist simultenously.\n"
 				"To access the current instance, use:\n\n"
 				"FreqScope.getInstance\n\n"
-			)
-		} {
-			//make scope
+			);
+			^nil
+		}
+		//make scope
 
-			scopeColor = scopeColor ?? { Color.new255(255, 218, 000) };
+		scopeColor = scopeColor ?? { Color.new255(255, 218, 000) };
 
-			scopeOpen = true;
+		scopeOpen = true;
 
-			server = server ? Server.default;
+		server = server ? Server.default;
 
-			rect = Rect(0, 0, width, height);
-			pad = [30, 48, 14, 10]; // l,r,t,b
-			font = Font.monospace(9);
-			freqLabel = Array.newClear(12);
-			freqLabelDist = rect.width/(freqLabel.size-1);
-			dbLabel = Array.newClear(17);
-			dbLabelDist = rect.height/(dbLabel.size-1);
+		rect = Rect(0, 0, width, height);
+		pad = [30, 48, 14, 10]; // l,r,t,b
+		font = Font.monospace(9);
+		freqLabel = Array.newClear(12);
+		freqLabelDist = rect.width/(freqLabel.size-1);
+		dbLabel = Array.newClear(17);
+		dbLabelDist = rect.height/(dbLabel.size-1);
 
-			nyquistKHz = server.sampleRate;
-			if( (nyquistKHz == 0) || nyquistKHz.isNil, {
-				nyquistKHz = 22.05 // best guess?
-			},{
-				nyquistKHz = nyquistKHz * 0.0005;
-			});
+		nyquistKHz = server.sampleRate;
+		if( (nyquistKHz == 0) || nyquistKHz.isNil, {
+			nyquistKHz = 22.05 // best guess?
+		},{
+			nyquistKHz = nyquistKHz * 0.0005;
+		});
 
-			setFreqLabelVals = { arg mode, bufsize;
-				var kfreq, factor, halfSize;
+		setFreqLabelVals = { arg mode, bufsize;
+			var kfreq, factor, halfSize;
 
-				factor = 1/(freqLabel.size-1);
-				halfSize = bufsize * 0.5;
-
-				freqLabel.size.do({ arg i;
-					if(mode == 1, {
-						kfreq = (halfSize.pow(i * factor) - 1)/(halfSize-1) * nyquistKHz;
-					},{
-						kfreq = i * factor * nyquistKHz;
-					});
-
-					if(kfreq > 1.0, {
-						freqLabel[i].string_( kfreq.asString.keep(4) ++ "k" )
-					},{
-						freqLabel[i].string_( (kfreq*1000).asInteger.asString)
-					});
-				});
-			};
-
-			setDBLabelVals = { arg db;
-				dbLabel.size.do({ arg i;
-					dbLabel[i].string = (i * db/(dbLabel.size-1)).asInteger.neg.asString;
-				});
-			};
-
-			window = Window("Freq Analyzer", resizable: false)
-			.bounds_(rect.resizeBy(pad[0] + pad[1] + 4, pad[2] + pad[3] + 4))
-			.moveToBottom;
+			factor = 1/(freqLabel.size-1);
+			halfSize = bufsize * 0.5;
 
 			freqLabel.size.do({ arg i;
-				freqLabel[i] = StaticText(window, Rect(pad[0] - (freqLabelDist*0.5) + (i*freqLabelDist), pad[2] - 10, freqLabelDist, 10))
-					.font_(font)
-					.align_(\center)
-				;
-				StaticText(window, Rect(pad[0] + (i*freqLabelDist), pad[2], 1, rect.height))
-					.string_("")
-				;
+				if(mode == 1, {
+					kfreq = (halfSize.pow(i * factor) - 1)/(halfSize-1) * nyquistKHz;
+				},{
+					kfreq = i * factor * nyquistKHz;
+				});
+
+				if(kfreq > 1.0, {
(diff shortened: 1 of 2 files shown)


SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
```diff
@@ -313,7 +313,8 @@ FreqScopeView {
 }
 
 FreqScope {
-	classvar <scopeOpen;
+	classvar <scopeOpen = false;
+	classvar singleInstance;
```
JordanHendersonMusic: This is exactly how I'd do the singleton pattern. Keeping it private.

However it might make sense to have a public getter as it does most clearly communicate what is happening - returning the singleton from new is more of a hack.

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
```diff
@@ -322,7 +323,9 @@ FreqScope {
 		var setFreqLabelVals, setDBLabelVals;
 		var nyquistKHz;
 		busNum = busNum.asControlInput;
-		if(scopeOpen != true, { // block the stacking up of scope windows
+		if(scopeOpen) {
+			^singleInstance 
+		} {
```
JordanHendersonMusic: You could remove the else braces and remove a whole level of indentation here. 

```
if(open) { ^instance };
....
```

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
```diff
@@ -322,7 +323,9 @@ FreqScope {
 		var setFreqLabelVals, setDBLabelVals;
 		var nyquistKHz;
 		busNum = busNum.asControlInput;
-		if(scopeOpen != true, { // block the stacking up of scope windows
+		if(scopeOpen) {
```
JordanHendersonMusic: So this is discarding all the arguements passed in, this means the instance won't get updated.

JordanHendersonMusic:
Perhaps it would be better to throw an error asking the user to close the old window?

Then we could have a method *newOrClose (or something) that also closes any open windows.

telephon:
Good to fix as it is now, but …

I think it should never have been a singleton. 

One may want several of them with different settings, and at least one needs one per server. It should instead behave like `Stethoscope`.

prko:
@HotwheelsSisyphus 
Are you referring to the following behaviour?
```
FreqScope() // -> a FreqScope
f = FreqScope(); // -> FreqScope
f; // -> FreqScope
f.window // ERROR: Message 'window' not understood.

Window.closeAll

f = FreqScope(); // -> a FreqScope
f; // -> a FreqScope
f.window // -> a Window
f.window.close // the window is closed.

f = FreqScope(); // -> a FreqScope
f; // -> a FreqScope
f.window // -> a Window
g = FreqScope(); // -> FreqScope
g; // -> FreqScope
g.window // ERROR: Message 'window' not understood.
f.window // -> a Window
f.window.close // the window is closed.
```

As far as I remember:

- `s.freqscope` has historically been a singleton window. Calling `FreqScope()` does not create a new instance if one already exists. To create additional frequency scopes, users should use `FreqScopeView`.

- `s.scope` has also been a singleton. To open multiple scopes, users should use `Stethoscope`.

- `s.meter` behaved differently in the past: each evaluation created a new window. Since SC 3.13.0, however, it has also been a singleton, and users should use `ServerMeterView` to create additional meters.

- `s.plotTree` followed the same pattern as `s.meter`. It is now a singleton as of SC 3.14.0, and users should use `NodeTreeView` to create additional tree views.

These behaviours appear consistent across the different server GUI tools.

I’m mentioning them here only because I’m a bit concerned that this existing consistency might unintentionally change.

HotwheelsSisyphus:
> @HotwheelsSisyphus Are you referring to the following behaviour?

yes. I realize now that of course I could have done `FreqScope.scopeOpen = false` and then called `new` again, but still , this behavior threw me a bunch of times when I was trying to figure out how to, e.g., set the frame rate on the qt side; also because currently, exactly nothing happens when one tries to call FreqScope.new a second time. So at least a warning might be in order in that case. 
 
> I’m mentioning them here only because I’m a bit concerned that this existing consistency might unintentionally change. 

This PR will not change whether this class is a singleton or not (if you're worried, look at the code). If anything, it tries to make it more of a textbook case of a singleton. I think Julian's comment was more with respect to a long-term perspective, not with respect to this particular PR.

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
HotwheelsSisyphus: right. That's also not ideal, but from looking at it briefly it looks like it might be easier to just kill the current instance and make a new one with the new arguments rather than setting the arguments on the existing instance. 

(Because the positions of the buttons and the size of the window etc. are a bit interdependent; if you change the size, you gotta change the buttons, so you might as well start from scratch)

HotwheelsSisyphus:
So the last update did the following:
1. add methods: newReplace, getInstance
2. new posts a warning when scopeOpen = true (rather than returning the instance).

The odd thing is that I can now produce cases where I get
```
FreqScope.scopeOpen // -> false
// but then
FreqScope.new() // -> WARNING: Blablabla (i.e., the warning message that should only happen when scopeOpen = true)
```
Aaand also, sometimes I get more than one scope window which defeats the purpose of the whole thing.

But so far I've not been able to systematize when exactly this happens. (I.e., not always! Maybe has to do with what is assigned to vars?) I will play around with it more tomorrow.

HotwheelsSisyphus:
New update: 
reverted the addition of the newReplace method, which I couldn't get to work.
Keeping this PR minimal for now; might do more FreqScope changes later on but that seems more complicated, one would have to refactor the new method quite a bit. 

So the actual changes are:

1.  classvar `instance`
2. `getInstance` method 
3. `new` posts a warning when `scopeOpen = true`.

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
```diff
@@ -313,16 +313,28 @@ FreqScopeView {
...
+		^instance
+	}
+	
 	*new { arg width=522, height=300, busNum=0, scopeColor, bgColor, server;
 		var rect, scope, window, pad, font, freqLabel, freqLabelDist, dbLabel, dbLabelDist;
 		var setFreqLabelVals, setDBLabelVals;
 		var nyquistKHz;
 		busNum = busNum.asControlInput;
-		if(scopeOpen != true, { // block the stacking up of scope windows
+		if(scopeOpen) {
```
JordanHendersonMusic, line 330: Just to make this more readable - because this method is huge...

Could you reformat this like...

```
if(scopeOpen) {
   warn(...).postln;
   ^nil;
};
```

And then deindent everything that follows, removing the final set of brackets.

JordanHendersonMusic (changes requested):

SCClassLibrary/Common/GUI/PlusGUI/Control/FreqScope.sc
HotwheelsSisyphus: done!

JordanHendersonMusic (approved):
That formatter issue with the methods on the new line really needs fixing doesn't it!

dyfer:
@JordanHendersonMusic this has been sitting for a while - is it fine to merge this?

Issue #6589, opened 2025-01-14
Quarks GUI version list does not reflect update until window is closed & re-opened


daveriedstra:
## Environment

* SuperCollider version: 3.13.0
* Operating system: macOS, Linux
* Other details (Qt version, audio driver, etc.): n/a

## Steps to reproduce

* Start with out-of-date quark installed (ie, a new tagged version is available)
* Open Quarks GUI
* Click the name of your out-of-date quark
* Note the available versions
* Click "Check for updates" and wait for update check to complete
* Observe that the available versions of your quark have not changed
* Close and re-open the Quarks GUI
* Observe that the available versions of your quark are now up-to-date

## Expected vs. actual behavior

*Expected:* Quark version list is immediately updated after update check
*Actual:* Quarks GUI window must be closed and re-opened for version list to reflect updates

mariagnovais:
Hi @daveriedstra I would like to work on this issue! Could you please assign it to me?

capital-G:
> Hi [@daveriedstra](https://github.com/daveriedstra) I would like to work on this issue! Could you please assign it to me?

We don't assign issues - but it is good that you mentioned that you want to work on it. Looking forward to your PR!

PR #7374, opened 2026-02-18
QuarksGui:  refresh displayed version list after check for update (#6589 solved)

mariagnovais:
Issue (#6589):
After running "Check for updates", the Quarks GUI updated the directory data but did not refresh the displayed version list for the currently selected quark. The updated versions only appeared after closing and reopening the Quarks window.

Changes:
Added a refreshDirectoryView method to QuarksGui. This method: Clears the TreeView, resets the internal quarkRows cache, rebuilds the GUI state, restores the previously selected quark. The second change was replaced this.update with this.refreshDirectoryView after a successful directory update.

### DIFF
--- a/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
+++ b/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
@@ -640,22 +640,18 @@ QuarkRowView {
 	refreshDirectoryView {
 	var selectedPath, newQuark;
 
-	// remember what is selected in detail view (if any)
 	selectedPath = detailView.model.notNil.if({ detailView.model.localPath }, { nil });
 
-	// hard reset like closing + reopening the window
 	treeView.clear;
 	quarkRows = Dictionary.new;
 	initialized = false;
 
-	// rebuild rows from the updated model directory
 	this.update;
 
-	// restore selection so detail view + Versions tree rebuilds too
 	if(selectedPath.notNil) {
 		newQuark = model.all.detect({ |q| q.localPath == selectedPath });
 		if(newQuark.notNil) {
-			detailView.model = newQuark; // triggers detailView.update via model_
+			detailView.model = newQuark; 
 		} {
 			detailView.model = nil;
 		};
--- a/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
+++ b/SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
@@ -628,10 +628,7 @@ QuarkRowView {
 		isInstalled = quark.isInstalled;
 		btn.value = isInstalled.binaryValue;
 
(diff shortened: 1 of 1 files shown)


SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
```diff
@@ -628,12 +628,31 @@ QuarkRowView {
...
-		// this column has invisible text
-		// its used to sort the rows so that installed quarks are at the top
 		treeItem.setString(0, isInstalled.if("Y", { quark.isDownloaded.if("N", "") }));
-		// 1 is the install button. its not possible to sort by this column
 		treeItem.setString(2, quark.name ? "");
 		treeItem.setString(3, (quark.version ? "").asString);
 		treeItem.setString(4, (quark.summary ? "").replace(Char.nl.asString," ").replace(Char.tab.asString, ""));
 	}
+
+	refreshDirectoryView {
```
HotwheelsSisyphus, line 637: move to QuarksGui class

SCClassLibrary/Common/Quarks/GUI/QuarksGui.sc
```diff
@@ -628,12 +628,31 @@ QuarkRowView {
...
+	var selectedPath, newQuark;
```
HotwheelsSisyphus, line 638: and observe indentation levels of other methods:  i.e., 638-658 should be indented one more level.

HotwheelsSisyphus (changes requested):
Hi, Thanks for this!
It seems you accidentally put the method in the wrong class. Easy fix. 
And then there's teh indentation. just verify that it looks the same way as the other methods.

More important though is to actually test your code before submitting. This is pretty quick with the class library, if you set sclang_conf.yaml in your userConfigDir to include the classlibrary folder contained in your local sc repo. 
Then you only need to save changes to the .sc file and do "recompile class library" in the ide.

For now I can only verify that, once the method is moved to the QuarksGui class, the class library compiles fine. I don't currently have enough quarks installed here to produce the actual issue that this PR fixes... maybe someone else can help out there.

HotwheelsSisyphus:
Hi, may I ask why you closed the PR? 
I hope you weren't discouraged by the requested changes! 
They're maybe 2 minutes of work. It would be a pity to just leave it there since you've already figured most of it out... the reviews aren't meant to discourage people from contributing.

Another option,  if you're not sure about a  PR, is converting it to draft.
 Also, a lot of PRs in this repo are just sitting around open but somewhat unfinished, so that's also normal if you don't have time to finish it right now. Though ideally in that case you'd leave a short comment to that effect.

Thanks in any case!

dyfer:
@mariagnovais are you planning to continue working on this PR?

dyfer:
@mariagnovais ping

Issue #7376, opened 2026-02-19
3.14.1 fails to build with boost 1.87.0


tgurr:
## Environment

* SuperCollider version: 3.14.1
* Operating system: Linux
* Other details: Qt 6.10.2
* GCC: 15.2.0

## Steps to reproduce
1. have boost 1.87.0 installed
2. try to build supercollider 3.14.1

## Expected vs. actual behavior

```
[ 62%] Building CXX object lang/CMakeFiles/libsclang.dir/LangSource/SC_LanguageClient.cpp.o
[...]

PR #7379, opened 2026-02-23
scsynth: CoreAudio: replace boost::optional with std::optional

### LINKED ISSUE #7221: Replace `boost::optional` with `std::optional`
Found e.g. in 
https://github.com/supercollider/supercollider/blob/4ac149a818ee87a5da97c6651b93bb5487003193/server/scsynth/SC_CoreAudio.h#L28

See https://www.boost.org/doc/libs/latest/libs/optional/doc/html/boost_optional/std_comp.html - probably similar to `std::filesystem`

anirban6996:
## Purpose and Motivation

Fixes #7221 

This PR modernizes the audio core by replacing `boost::optional` with C++17's native `std::optional` in `server/scsynth/SC_CoreAudio.h`. 

Since the codebase now supports C++17, removing this specific Boost dependency reduces compile-time overhead and brings the code closer to modern standard library practices. 

I have verified the codebase using `grep` to ensure no other instances of `boost::optional` remain, and the project compiles 100% cleanly on macos 26

## Types of changes

- Refactoring / Tech Debt (Modernizing to C++17 standard)

## To-do list

- [x] Code is tested (Compiled `scsynth` cleanly locally)
- [x] All tests are passing (Local build hit 100%)
- [ ] Updated documentation (N/A - purely internal C++ change)

### DIFF
--- a/server/scsynth/SC_CoreAudio.cpp
+++ b/server/scsynth/SC_CoreAudio.cpp
@@ -1160,7 +1160,7 @@ bool SC_CoreAudioDriver::DriverSetup(int* outNumSamplesPerCallback, double* outS
 
     *outNumSamplesPerCallback = mHardwareBufferSize / outputStreamDesc.mBytesPerFrame;
     if (mExplicitSampleRate) {
-        *outSampleRate = mExplicitSampleRate.get();
+        *outSampleRate = mExplicitSampleRate.value();
     } else {
         *outSampleRate = outputStreamDesc.mSampleRate;
     }
--- a/server/scsynth/SC_CoreAudio.h
+++ b/server/scsynth/SC_CoreAudio.h
@@ -25,7 +25,7 @@
 #include "OSC_Packet.h"
 #include "SC_SyncCondition.h"
 #include "PriorityQueue.h"
-#include <boost/optional.hpp>
+#include <optional>
 
 #include <SC_Lock.h>
 
@@ -160,7 +160,7 @@ protected:
     int mNumSamplesPerCallback;
     uint32 mPreferredHardwareBufferFrameSize;
     uint32 mPreferredSampleRate;
-    boost::optional<uint32> mExplicitSampleRate;
+    std::optional<uint32> mExplicitSampleRate;
     double mBuffersPerSecond;
     double mAvgCPU, mPeakCPU;
     int mPeakCounter, mMaxPeakCounter;


JordanHendersonMusic:
> the project compiles 100% cleanly on a Linux environment (Ubuntu 24.04 / GCC 13.3) 

These changes only affect mac os, hence why the CI is failing.

Spacechild1:
First, thanks for the PR!

> and the project compiles 100% cleanly on a Linux environment (Ubuntu 24.04 / GCC 13.3) after this drop-in replacement.

You should really compile on the system that is affected by the changes. In this case, it's macOS, not Linux.

This is the error:
```
/Users/runner/work/supercollider/supercollider/server/scsynth/SC_CoreAudio.cpp:1163:46: error: no member named 'get' in 'std::optional<unsigned int>'
 1163 |         *outSampleRate = mExplicitSampleRate.get();
```
`boost::optional` uses `get()` whereas `std::optional` uses `value()`. Therefore it's not really a drop-in replacement. Just change the code to `*outSampleRate = mExplicitSampleRate.value();`

anirban6996:
I will try compiling on my macbook, but I am out and my macbook is at home...I will try to compile it Tommorow when I get back home

anirban6996:
Screenshot 2026-02-24 at 11 40 35 AM

I changed `*outSampleRate = mExplicitSampleRate.get();` to `*outSampleRate = mExplicitSampleRate.value();` and it does compile and but i got lots of deprecated warnings.....

Screenshot 2026-02-24 at 11 51 23 AM

but the boost external library is still there too

Spacechild1:
I mean, `boost::optional` is used inside boost, so naturally we can't get rid of it entirely.

But at least we're not using `boost::optional` in our own code anymore.

Spacechild1 (approved):

anirban6996:
> I mean, `boost::optional` is used inside boost, so naturally we can't get rid of it entirely.
> 
> But at least we're not using `boost::optional` in our own code anymore.

Yes I understand it is an external library...but my question was that...is that library really needed after we replaced our own code...or that library is still used in other places of our own code....that was my real question actually...

Spacechild1:
Boost itself is used in quite a few places.

capital-G:
Thanks for the PR! :)

> I mean, `boost::optional` is used inside boost, so naturally we can't get rid of it entirely.

Well - we should have checked if this is actually the case ;) Boost brings its own dependency management tool called `bcp`, which eliminates (or adds) dependencies if they are not used anymore.

I just re-run it on boost 1.90 and we still have the `optional` directory, so there is still some other boost dependency which uses the `boost::optional` instead of `std::optional`. But this is fine since boost is self-contained.

The reason why I wanted to get rid of `boost::optional`: Optional has become part of the C++ standard library, and as it is often the case, boost acts as an incubator for new C++ features. But once such a library become mainline, boost also tends to remove its own implementation in favor of the std one.
By switching to the std implementation of `optional`, we won't run into problems once `boost::optional` gets removed (this is what happened with `boost::filesystem`). And as seen in https://github.com/supercollider/supercollider/issues/7376 and https://github.com/supercollider/supercollider/pull/6477 - there is often not the possibility to just stay on an older boost version.

edit: See e.g.
https://github.com/supercollider/supercollider/blob/71813a99bbb5a465f37b1d3cc0df98277679ae90/external_libraries/boost/boost/program_options/detail/value_semantic.hpp#L169
we use `program_options` to parse CLI arguments.

I can't see any usage of `boost::optional` in our code. I searched for `boost::optional` and `boost/optional`.

Spacechild1:
> The reason why I wanted to get rid of boost::optional: Optional has become part of the C++ standard library, and as it is often the case, boost acts as an incubator for new C++ features. 

It used to :) It seems like this role has been taken over by the Beman project: https://bemanproject.org/

> But once such a library become mainline, boost also tends to remove its own implementation in favor of the std one.

That's not true! I don't think boost ever removed libraries after they have been adopted by the C++ standard. After all, they want to support existing projects and C++ libraries are not always a 100% drop-in replacement.

> By switching to the std implementation of optional, we won't run into problems once boost::optional gets removed (this is what happened with boost::filesystem). 

Boost did *not* remove `boost::filesystem`, we just replaced `boost::filesystem` with `std::filesystem` because we have officially switched to C++17 and `std::filesystem` became widely available.

anirban6996:
Thanks for taking the time out and explaining everything ;)....I am trying to understand how these things work... from the explanation I understood that we still need other tools of boost for other things ( like `boost::program_options` ) but we don't have `boost::optional` in our own code (the other tools might use `boost::optional` internally) so we cannot remove it completely and boost is the incubation lab for future features which might get added to the standard library.

anirban6996:
> It used to :) It seems like this role has been taken over by the Beman project: https://bemanproject.org

Oh so beman project will be the new boost of the future I see

PR #7381, opened 2026-02-23
sclang: `var` in middle of function

JordanHendersonMusic:
## Purpose and Motivation

This is a work in progress, but the following does currently work as expected.
I'm making a PR so people can see what I'm intending with the other PRs, and so this work doesn't get lost some how.

This is a feature I've wanted for years and the lack of it really annoys me...

```
// silly place holder syntax while I migrate.
var foo = #;{
   |f ...a, k|
    a.postln;
    k.postln;
    var bar = 10;
    bar = bar + f
};
foo.(10, 2, meow: 10);
// [2]
// [meow: 10]
// -> 20 
```

This PR is built on top of the error code #7366 , this is because it did a major refactor that makes working with the parser/compiler much easier.

This PR is actually only a couple of hundred lines.

This PR also exposed all sorts of bizzare and unwanted behaviour in the name resolution, if we want this PR, we also need #7203.

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


Spacechild1:
Wow, I have wanted this for years! Currently, the only other language I know that requires users to declare all variables upfront is C89 :)

Do you know why this limitation existed in the first place? Are there any tradeoffs?

---

Just a side note: as a Quark/Extension author there are still many situations where I would have to declare variables upfront to avoid inline warnings. Just to give a silly little example:
```
(
// this causes an inline warning:
f = { |i|
	if (i > 0) {
		var a = i * 0.5;
		var b = i * 2.0;
		a + b;
	} { i }
}
)

(
// this is fine
f = { |i|
	var a, b;
	if (i > 0) {
		a = i * 0.5;
		b = i * 2.0;
		a + b;
	} { i }
}
)
```

What should I do in this case? Tell users to disable inline warnings?

What is the state of inline warnings? Are they on or off by default?

JordanHendersonMusic:
> Do you know why this limitation existed in the first place?

I think Smalltalk required you to do this.

I also believe there were no var initialisers, and SC added these later, hence why the order of execution is a mess, it was just an oversight.

You example of inline warnings is more of a social issue... 

If we could agree that `if` is a keyword and not a message, removing ugen.:if, then I can make the compile automatically lift those declarations outside. It's not hard to do... Well first I'll have to rewrite the compiler but I seem to be doing that anyway...

 I believe the warning is off by default though. The ci requires the class library to not have any as a part of a test.

jamshark70:
> Do you know why this limitation existed in the first place?

Perhaps because James McCartney is a solo developer, not a language-design committee with a dev team, and he was also responsible for a whole audio engine and suite of plugins, so some "nice to have" things didn't make it :wink:

JordanHendersonMusic:
Also, we don't have stack based variables — variables don't have stack semantics. If they did, it would be trivial to declare variables where ever you want.

This is odd at first, but actually really desirable as it make implementing continuations and exceptions far easier. Since sc is a language for music, having all functions be co-routines which can be forked and/or paused makes a lot of sense. Otherwise we'd need a special `async` keyword like other languages and introduce the whole coloured function issue.

All our variables have frame semantics and live as long as the frame does, not as long as the stack. This requires declaring them. The natural place to do this is at the top of the block. This pr loops over the body of the function to find them all.

> not a language-design committee with a dev team

... and working 25-30 years ago, since which language design as a field has matured a great deal!

Spacechild1:
> Perhaps because James McCartney is a solo developer, not a language-design committee with a dev team, and he was also responsible for a whole audio engine and suite of plugins, so some "nice to have" things didn't make it

I didn't mean this a criticism of JMC, I just thought there were some performance considerations when he made this decision, just like the static class compilation (which enables O(1) method lookup).

> ... and working 25-30 years ago, since which language design as a field has matured a great deal!

I don't know a single scripting language other than sclang that requires variables to be declared upfront. Compiled languages also moved away from this long time ago. (IIRC the original reason for declaring variables at the top of a block was to reduce the memory usage of the compiler.) Even 30 years ago, sclang was the exception in this regard. That's why I thought there must have been a particular reason (compilation speed, runtime performance, etc.)

> Also, we don't have stack based variables — variables don't have stack semantics. If they did, it would be trivial to declare variables where ever you want.
> ...
> This is odd at first, but actually really desirable as it make implementing continuations and exceptions far easier.
> Otherwise we'd need a special async keyword like other languages and introduce the whole coloured function issue.

Ok, so this is probably the actual reason: it's easier to implement closures and coroutines this way.

I just want to point out that Lua solves this quite elegantly with their "upvalue" technique. This is explained in chapter 5 ("Functions and Closures") in https://www.lua.org/doc/jucs05.pdf:

> Lua uses a structure called an upvalue to implement closures. Any outer local variable is accessed indirectly through an upvalue. The upvalue originally points to the stack slot wherein the variable lives (Figure 4, left). When the variable goes out of scope, it migrates into a slot inside the upvalue itself (Figure 4, right). Because access is indirect through a pointer in the upvalue, this migration is transparent to any code that reads or writes the variable. Unlike its inner functions, the function that declares the variable accesses it as it accesses its own local variables: directly in the stack.

For more details, see https://www.cs.tufts.edu/~nr/cs257/archive/roberto-ierusalimschy/closures-draft.pdf.

JordanHendersonMusic:
Crafting interpreters also uses lua's upvalue model in their implementation - great book!

It definitely impacts compilation speed as it has to look over every line of code in the block to visit the variables... but ought to be minimal, and only happens at compile time, lookup is completely resolved at compile time so runtime isn't affected.

Spacechild1:
Great book indeed! I'm not sure if this covered/mentioned in the book, but Wren pulls of few tricks to get very fast member lookup and method calls while still supporting to define classes at runtime: https://wren.io/performance.html

JordanHendersonMusic:
> This lets Wren avoid the extra checking most languages need to do at runtime to handle too few or too many arguments being passed to a method. In Wren, it’s not syntactically possible to call a method with the wrong number of arguments.

This is our issue, we do message sending in which you can do anything!

We do the rest, apart from the single pass compiler, which is what this pr does, adds another but brief pass over a subset of the tree.

Spacechild1:
> This is our issue, we do message sending in which you can do anything!

But how is message passing different from method calls? When we send a message to an object, don't we just look up the corresponding method in the object's class? Currently, the lookup is done via a global method table, but I don't quite see why could not switch to copy-down-inheritance. (I'm not saying we really should, just asking out of curiousity.)

JordanHendersonMusic:
Oh the lookup is super fast. We actually store the offset into the method table inside the symbol of the method name, and an additional larger offset inside the symbol of the class name. It's basically ...

```
gMethods[class.offset + method.offset]
```
It's very clever and heavily relies on a static compile time class definition, just like wren! I know the author does some music (they have a YouTube) so I wonder if they have heard of supercollider? I dare '@' them!

The bit that is really slow is arguement matching, we can have too few, too many, keywords, keywords that don't match, keywords that collide, can get defaults from environments, and variable args and variable kwargs!

Issue #2378, opened 2016-10-05
Unary operator trunc gives incorrect results for negative numbers


Beechside:
The unary operator trunc should always round towards zero.

-3.3.trunc should give -3 but SC gives -4
-3.3.trunc(0.25) should give -3.25 but SC gives -3.5

I know of no language where trunc always rounds towards negative infinity so believe this to be a bug in SC.

I realise that I can use -3.3.trunc(-1) and -3.3.trunc(-0.25) to get the correct results but that seems to be going against the "standard" definition of trunc.

Sciss:
SuperCollider defines many mathematically operations in a "non-standard" way. Given that trunc has worked this way since 2003 or so, I don't think it's warranted to change its operation. Apart from this general reservation, I'm also not aware that "trunc" is defined anywhere as a standard mathematical operation.

Trunc is defined as follows (my copy in Scala):

```scala
def trunc (a: Float, b: Float): Float = if (b == 0) a else (math.floor(a / b) * b).toFloat
```

This is a very clear and unambiguous definition.

Beechside:
I hope you don't mind my saying that your example is a clear implementation, rather than a clear definition ;o)

I can understand the desire to protect existing code and will support if that is the broad consensus.

Nonetheless, the act of truncation is to remove digits, not to round or to floor.  Removing digits always tends towards zero.  As for a "standard" definition, you would find no universal algorithmic consensus anywhere from from wiki to Knuth but truncation is a very unambiguous process which I believe to have a clear mathematical parallel.

I appreciate your speedy reply - thank you from a github newcomer, even if I am an SC veteran...

vivid-synth:
Other languages do agree with @Beechside's description:

Haskell:
```
> truncate (-3.3)
-3
```

Ruby:
```
irb(main):003:0> -3.3.truncate
=> -3
```

Python:
```
>>> import math
>>> math.trunc(-3.3)
-3
```

Javascript:
```
Math.trunc(-3.3)
-3
```

I'm maybe not the best advocate because I don't use sclang as much as scsynth, but I'd be open to the idea of "fixing" how ```trunc``` works if there isn't much code that uses it with negative numbers, and if everyone is notified that it's changed. Not sure it's worth it on balance, but it's worth discussing.

vivid-synth:
[...]

PR #7387, opened 2026-02-27
trunc(ate), div(ide): document behavior and implement "fixed" versions.

HotwheelsSisyphus:
Thought it would be better to reopen this on another branch. Basically waiting for instructions now.
<details><summary>The story so far
</summary>
<p>

As discussed in that other issue (#2378). 

`div` actually has a similar problem, but we did not discuss it explicitly.
(i.e., `a.div(b) == floor(a/b)` but is normally equivalent to `std::trunc(a/b)`).  

We had a little "vote" in https://github.com/supercollider/supercollider/issues/2378#issuecomment-3623515421, and there are now 2 votes for `truncate` and 0 for `trunc2` as a name for the "standard" method, though I recall that this was tied 2-2 as some point.
(That quorum is, uh, suboptimal, but what can ya do.) 

</p>
</details> 

#### I think I've now figured out how to do this but I do need some direction/decision by more experienced folks as to _what_ exactly to do.
We've agreed on non-breaking changes, i.e., only  **adding** things rather than changing interfaces (i.e., not simply "fixing" the existing trunc). So at any rate, the resulting solution should not break code written in sclang. So that doesn't need to be discussed. The question is really about the Server side only. 

There are two main options:
1. Simply **add** new, "fixed" UGens (truncate and divide)
2. **Replace** the respective operations with corrected ones, and reimplement the backward-compatible trunc on the class library side.

1 of course is simplest. It just seems a bit inelegant and adds more code. Sort of a shantytown approach.
2 replaces some code in the server and adds very little to the class library.  

For 2, there would be 2 options: 
- a) Internally rename the operations in question (i.e., opTUGenrunc becomes opTruncate), such that a .truncate method would  call a _Truncate primitive). 
- b) don't rename them, only correct the math. I think this could become confusing.

Within 1, two more questions:
- a) For `truncate`, the question is:
    - Should it be a unary or a binary op?
        -  if it is to be strictly analogous to sc's `trunc` , then it would a binary op, with two extra multiplications. (Sc's `trunc` is currently `std::floor(a/b) * b`; a binary op `truncate` would be `std::trunc(a/b) *b`. ) 
         -  if it is to behave like c-style `trunc`, then it would be unary. 
- b) Should the extra code for that UGen be in a separate file, or at the end of the respective naryOpUGens file?

I think that's roughly it.

### DIFF
--- a/HelpSource/Classes/SimpleNumber.schelp
+++ b/HelpSource/Classes/SimpleNumber.schelp
@@ -41,7 +41,18 @@ method:: mod
 Modulo
 
 method:: div
-Integer Division
+Divide teletype::this:: by teletype::aNumber:: and rounds to next-lower integer.
+For positive integers strong::only::, this is equivalent to standard integer division.
+Warning::
+This is a non-standard implementation of integer division,
+which has been preserved in SuperCollider for backward compatibility reasons,
+and is equivalent to code::floor(this/aNumber)::.
+
+For standard integer division, use code::truncate(this/aNumber):: or link::#-divide::
+::
+
+method:: divide
+Standard integer division, divides teletype::this:: by teletype::aNumber:: and rounds towards 0.
 
 method:: **
 Exponentiation
@@ -71,7 +82,10 @@ method:: round
 Round to multiple of aNumber
 
 method:: roundUp
-Round up to a multiple of aNumber. For roundDown use link::Classes/SimpleNumber#-trunc#trunc::.
+Round up to a multiple of aNumber.
+
+method:: roundDown
+Round down to a multiple of aNumber. 
 
 method:: smallButNotZero
 Check if the value is closer to zero than a threshold, but not zero.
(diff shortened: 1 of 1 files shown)


telephon:
Thank you for this overview! For the dev meeting in may, we'll write up a strategy (with @capital-G and you could also join!).

telephon:
As a direct reply for the part 2: 
add to the end of the respective naryOpUGens file.  

There is a slightly difficult trade-off to be made with respect to unary vs. binary op.
- if we want to provide an easy option for people to switch between the two behaviours, then the interface should be the same for both
- if we want to indicate that this is an implementation of the c-interface, it should be unary. This has the advantage of being slightly more efficient.

I tend towards the first one, and (still, sorry!) towards naming it `trunc2` for symmetry reasons The documentation could be:

- trunc: asymmetric truncation
- trunc2: symmetric truncation (as is standard in c, but with an extra argument)

PR #7392, opened 2026-03-02
docs: Expand Window.schelp

prko:
## Purpose and Motivation

This PR enhances the SuperCollider `Window.schelp` help documentation by adding comprehensive examples for setting window bounds.

Window bounds configuration is a common source of confusion for beginners, particularly the distinction between screen-relative and window-relative coordinate systems. This PR addresses that issue by providing detailed, well-structured examples with clear explanations.

The additions include:
- Hard-coded examples demonstrating direct coordinate usage and the difference between `Window.screenBounds` and `Window.availableBounds`
- Flexible examples showcasing practical methods such as `Rect.aboutPoint`, `insetBy`, and centre-point calculations
- Comparative explanations of coordinate system origins (bottom-left for windows vs. top-left for views)

These examples serve as practical reference material, enabling users to write more robust and maintainable code for window positioning across different screen resolutions and operating systems.
 
## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Window.schelp
+++ b/HelpSource/Classes/Window.schelp
@@ -366,6 +366,7 @@ v.background = Color.gray;
 t = TextView(v, Rect(10, 10, 80, 30));
 c = StaticText(x, Rect(90, 30, 50, 25));
 c.string = "test";
+x.front;
 )
 ::
 
--- a/HelpSource/Classes/Window.schelp
+++ b/HelpSource/Classes/Window.schelp
@@ -404,15 +404,15 @@ subsubsection:: Adjusting bounds from top-left
 To position a window relative to the top, subtract from the height.
 
 code::
-/* Hard‑coded example (using a fixed pixel offset): */
+/* Example that positions the window using a fixed pixel offset: */
 (
-x = Window("hard-coded position",
+x = Window("fixed-offset position"",
     Rect(100, Window.screenBounds.height - 200, 300, 100)
 );
 x.front;
 )
 
-/* Scaling avoids hard‑coded (fixed‑pixel) offsets: */
+/* Scaling the offset relative to the screen height: */
 (
 ~factor = 0; // 0: top, 1: bottom
 ~rectHeight = 100;
--- a/HelpSource/Classes/Window.schelp
+++ b/HelpSource/Classes/Window.schelp
@@ -336,7 +336,7 @@ w = Window.new("test");
 w.front;
 )
 
-/* Hard‑coded example (using a fixed pixel offset): */
+// Set window position using a fixed pixel offset:
 (
 w.bounds = Rect(100, 100, 400, 400); // 100 avoids collision with OS GUI elements.
 w.front;
--- a/HelpSource/Classes/Window.schelp
+++ b/HelpSource/Classes/Window.schelp
@@ -327,13 +327,13 @@ it is generally safer to scale positions relative to the screen or available are
 
 subsubsection:: Adjusting bounds from bottom-left
 
-Window coordinates use a bottom-left origin. The following examples show how to
+link::Classes/Window:: class coordinates use a bottom-left origin. The following examples show how to
 adjust bounds after creating a window.
 
 code::
 /* Create a window at the default position and size: */
 (
-w = Window("test");
+w = Window.new("test");
 w.front;
 )
 
@@ -341,7 +341,7 @@ w.front;
 w.bounds_(Rect(0, 0, 400, 400)).front; // May conflict with OS GUI elements at the bottom.
 
 (
-w.bounds = Rect(0, 100, 400, 400); // 100 avoids collision with OS GUI elements.
+w.bounds = Rect(100, 100, 400, 400); // 100 avoids collision with OS GUI elements.
 w.front;
 )
 
@@ -355,12 +355,35 @@ w.bounds_(Rect(0, Window.availableBounds.height * ~factor, 400, 400));
 
 (
 ~factor = 0.035; // 0: bottom, 1: top
-w.bounds_(Rect(0, Window.screenBounds.height * ~factor, 400, 400));
+w.bounds_(Rect(100, Window.screenBounds.height * ~factor, 400, 400));
 // Window.screenBounds includes the full display area.
 // Use this when OS GUI elements are hidden.
 )
 
 w.close;
+::
+
+note::
+The origin differs from the coordinate system used inside a link::Classes/View:: or link::Classes/Window:: class, which use a top-left origin:
+
+code::
+(
+x = Window("test", Rect(100, 100, 300, 100)); // origin: bottom-left
+
+b = Button(x, Rect(90, 0, 50, 25));           // origin: top-left
+v = View(x, Rect(0, 0, 100, 50));             // origin: top-left
+v.background = Color.gray;
+
+t = TextView(v, Rect(10, 10, 80, 30));        // origin: top-left
+
+c = StaticText(x, Rect(90, 30, 50, 25));      // origin: top-left
+c.string = "test";
+
+x.front;
+)
+
+x.close;
+::
 ::
 
 subsubsection:: Adjusting bounds from top-left
@@ -370,23 +393,32 @@ To position a window relative to the top, subtract from the height.
 code::
 /* Hard‑coded example (using a fixed pixel offset): */
 (
-x = Window.new("hard-coded position",
-Rect(100, Window.screenBounds.height - 200, 300, 100)
+x = Window("hard-coded position",
+    Rect(100, Window.screenBounds.height - 200, 300, 100)
 );
 x.front;
 )
 
 /* Scaling avoids hard‑coded (fixed‑pixel) offsets: */
 (
-var factor = 0; // 0: top, 1: bottom
-var rectHeight = 100;
-var titleHeight = 32; // The title bar height varies depending on OS and display setting.
-var normalizedPosition = Window.screenBounds.height * factor + rectHeight + titleHeight;
+~factor = 0; // 0: top, 1: bottom
+~rectHeight = 100;
+~titleHeight = 32; // The title bar height varies depending on OS and display setting.
+~screenHeight = Window.screenBounds.height;
+~normalizedOffset = ~screenHeight * ~factor + ~rectHeight + ~titleHeight;
 
-y = Window.new("scaled position",
-Rect(100, Window.screenBounds.height - normalizedPosition, 300, rectHeight)
+y = Window("scaled position",
+    Rect(100, ~screenHeight - ~normalizedOffset, 300, ~rectHeight)
 );
 y.front;
+)
+
+(
+~yHeight = y.bounds.height;
+~normalizedOffset  = ~screenHeight * 0.1 + ~yHeight + ~titleHeight;
+y.bounds_(
+	Rect(100, ~screenHeight - ~normalizedOffset, y.bounds.width, ~yHeight)
+).front
 )
 
 x.close;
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,90 @@ w.front;
...
+
+// This bottom-left origin differs from the coordinate system used within a View and a
+// Window, where the origin is at the top-left:
+(
+b = Button(x, Rect(90, 0, 50, 25));
+v = View(x, Rect(0, 0, 100, 50));
+v.background = Color.gray;
+t = TextView(v, Rect(10, 10, 80, 30));
+c = StaticText(x, Rect(90, 30, 50, 25));
+c.string = "test";
```
HotwheelsSisyphus, line 377:
```suggestion
c.string = "test";
x.front;
```

HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,90 @@ w.front;
...
+y.front;
+)
+// Window.availableBounds.height may differ from Window.screenBounds.height
+// depending on the OS (e.g., macOS menu bar, Windows taskbar).
+// This makes the top margin more accurate in practice.
+::
+
+code::
+/* Changing bounds after window creation */
+x.bounds_(Rect(100, 300, 300, 100));
```
HotwheelsSisyphus:
```suggestion
x.bounds_(Rect(100, 300, 300, 100)).front;
```

HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,90 @@ w.front;
 
 subsection:: Setting Bounds
 
+subsubsection:: Hard code examples
```
HotwheelsSisyphus: I would just leave out the subsubsection headings. ("hardcoded" vs. "flexible") ... I see what  you mean now, but I think the wording here is slightly confusing.

HotwheelsSisyphus (changes requested):
Thanks! I do think these examples are helpful, but I'd suggest reworking them, and discarding the "hard-coded"/"flexible" opposition, which I find confusing here. 
To handle different display resolutions, I'd just multiply (scale) instead of add (offset), by which I mean this:
```
// your example
x = Window.new("test", Rect(100, Window.screenBounds.height - 300, 300, 100));
```
could become this:
```
x = Window.new("test", Rect(100, Window.screenBounds.height * someFactor, 300, 100));
```
and similar changes elsewhere.

I'd also suggest removing redundancies; we don't need to show all the four combinations of insetBy/center with availableBounds/screenBounds. 
If this were a tutorial or guide, this would be different and you could be more expansive; but this file is technically class reference.

prko:
> To handle different display resolutions, I'd just multiply (scale) instead of add (offset), by which I mean this:
> 
> ```
> // your example
> x = Window.new("test", Rect(100, Window.screenBounds.height - 300, 300, 100));
> ```
> 
> could become this:
> 
> ```
> x = Window.new("test", Rect(100, Window.screenBounds.height * someFactor, 300, 100));
> ```
> 
> and similar changes elsewhere.

```
x = Window.new("test", Rect(100, Window.screenBounds.height - 300, 300, 100));
```
This code was written by @mtmccrea, so it would be best to wait for his feedback before removing it.
Alternatively, we could keep both examples in parallel if that is acceptable.

prko:
@HotwheelsSisyphus 
Thanks again for your input!

I’ve rewritten the whole section, and I’d appreciate it if you could let me know whether this version reads better.

prko:
@HotwheelsSisyphus 
Thanks again for your input!

I’ve rewritten the whole section, and I’d appreciate it if you could let me know whether this version reads better.

mtmccrea:
> This code was written by @mtmccrea, so it would be best to wait for his feedback before removing it.

Either a factor or pixel offset is reasonable, I don't have a preference. But I also don't think either should be "avoided" as the comment in the example implies.

prko:
Thanks!

@mtmccrea
>Either a factor or pixel offset is reasonable, I don't have a preference. But I also don't think either should be "avoided" as the comment in the example implies.

Instead of the current sentences:
```
When positioning windows, remember that screen sizes and available display areas
vary across systems. Instead of using a fixed pixel offset (a hard‑coded value),
it is generally safer to scale positions relative to the screen or available area.
```
how about the following?
```
Because screen dimensions and available display areas can vary across systems—and 
even within the same system depending on user settings—window positions and sizes may 
be specified using fixed pixel values or by scaling them relative to the screen. 
Both approaches are valid; choose the method that best matches the layout behavior you want to support.
```

mtmccrea:
I would suggest something like:

> Because screen sizes vary, scaling window coordinates relative to the screen bounds can offer flexibility when using your GUI on different systems.

I agree with @HotwheelsSisyphus that avoiding redundancies is preferred. For example, it could be OK to omit the pixel offset example, since that's pretty straightforward.

prko:
@mtmccrea 
Thank you very much.
I’ve updated the text as you suggested, and I’ve also removed one of the code lines in the pixel‑offset examples.
(I kept the other example because it explains how to avoid a collision with the OS GUI.)

HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,140 @@ w.front;
...
+
 code::
-// use screenbounds for precise placement from the top
+/* Create a window at the default position and size: */
+(
+w = Window.new("test");
+w.front;
+)
+
+/* Hard‑coded example (using a fixed pixel offset): */
```
HotwheelsSisyphus:
```suggestion
// Set window position using a fixed pixel offset:
```

HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,140 @@ w.front;
...
+(
+w.bounds = Rect(100, 100, 400, 400); // 100 avoids collision with OS GUI elements.
+w.front;
+)
+
+/* Scaling avoids hard‑coded (fixed‑pixel) offsets: */
```
HotwheelsSisyphus:
```suggestion
// Set window position (and size) with respect to available pixels in display's resolution:
```

HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,140 @@ w.front;
...
+(
+~factor = 0.037; // 0: bottom, 1: top
+w.bounds_(Rect(0, Window.availableBounds.height * ~factor, 400, 400)).front;
```
HotwheelsSisyphus, line 346: I would adjust _all_ of the bounds this way. I'd put the availableBounds in a var and then multiply them with factors for the respective dimensions. But I'm not sure if that is maybe too much to explain here ( e.g., why the indexing [2,3,2,3], etc.;).
```
~factors = [0, 0.037, 0.3, 0.3] // fractions of respective dimensions
~bds = Window.availableBounds.asArray[[2,3,2,3]]
w.bounds_(~factors * ~bds)
```

HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,140 @@ w.front;
...
+subsubsection:: Applying margins
+
+Use code::Rect.insetBy(h, v):: to apply consistent margins from the
+edges of the available display area.
+
+code::
+(
+w = Window("window with margins",
+    Window.availableBounds.insetBy(300, 200)
+    // Rect.insetBy(h: hMargin, v: vMargin)
```
HotwheelsSisyphus: remove this line. you've already explained the args for Rect.insetBy above.

HelpSource/Classes/Window.schelp
```diff
@@ -320,15 +321,140 @@ w.front;
...
+subsubsection:: Centering a window
+
+Use code::Rect.aboutPoint(point, dx, dy):: to center a window
+around a point. Use code::Window.screenBounds:: when OS GUI elements are hidden.
+
+code::
+(
+w = Window("centered window",
+    Rect.aboutPoint(Window.availableBounds.center, 200, 100)
+    // Rect.aboutPoint(point: center, dx: halfWidth, dy: halfHeight)
```
HotwheelsSisyphus: remove this line (redundant)

HotwheelsSisyphus (changes requested):
Thanks for the reminder about this PR!

General things, I did not flag all instances of these, please do look for them yourself.

1. Please remove the words "hard-coded". I you can change something at runtime like the bounds of a window, it's simply _not_ hard-coded:
```
// example of what hardcoded means, roughly:
Platform.userConfigDir = "my very special path" 
// -> Error: msg not understood
i.e., can't change it at runtime.

// whereas
w = Window();
w.bounds = [3,4,5,6];
// no error, can be changed.
```

See my first two suggestions for alternative formulations.

2. please rework the line breaks in accordance with the schelp style guidelines. (roughly, one sentence per line, avoid breaks in middle of sentences). No need to adhere to this slavishly everywhere but the present PR doesn't conform to it at all.

3. for single line comments, use //, not /*

HelpSource/Classes/Window.schelp
prko: Thanks!

HelpSource/Classes/Window.schelp
prko: Thanks!

HelpSource/Classes/Window.schelp
prko: Done in https://github.com/supercollider/supercollider/pull/7392/commits/4ac9a2e68f50e8681aa03cf3b639e26d53c950a9.

I think removing the part you consider redundant makes this less beginner-friendly, but I'm happy to defer to your judgement so that this PR can be merged more easily.

HelpSource/Classes/Window.schelp
prko: Done in https://github.com/supercollider/supercollider/pull/7392/commits/4ac9a2e68f50e8681aa03cf3b639e26d53c950a9.

I think removing the part you consider redundant makes this less beginner-friendly, but I'm happy to defer to your judgement so that this PR can be merged more easily.

HelpSource/Classes/Window.schelp
prko: OK, done in https://github.com/supercollider/supercollider/pull/7392/commits/4ac9a2e68f50e8681aa03cf3b639e26d53c950a9

If my attempt does not seem quite right, would you mind revising it in the way you think works best?

I may simply have a slightly different approach to writing documentation. I tend to include a bit more explanation and a few more examples so that beginners can follow the text and code more easily. My impression is that others in the developer group, including you, generally prefer a simpler and more elegant style. I can see the value in that as well.

HotwheelsSisyphus: Thanks! This particular comment wasn't really meant as something you _have_ to change, but 
I think your newest addition is good!

In general - and I've mentioned this to you before - I think it's important to distinguish (at least) between tutorials and general class reference. For tutorials, no problem to be more expansive, whereas, for reference, the preference is to keep it concise. (Window.schelp is in the `Classes` directory, so it should be considered class reference; whereas stuff in the `Guides` or  `Tutorials` dir would not be considered "class reference". Traditionally though, the distinction has been murky in supercollider docs.)

In general though, I think sheer repetition doesn't cut it in a written document, didactically speaking. I.e., writing things twice in a document isn't the same as repeating things in class or saying things multiple times in a youtube tutorial, because people do not scan written documents the same way that they listen to someone speaking. Judging from my own teaching experience, learners (or at least musicians) tend to get overwhelmed by large amounts of text, so if there is less text to parse to begin with, that may be a good thing).

HotwheelsSisyphus (approved):
Thanks, this is good to merge now I think!

HelpSource/Classes/Window.schelp
prko: >Judging from my own teaching experience, learners (or at least musicians) tend to get overwhelmed by large amounts of text, so if there is less text to parse to begin with, that may be a good thing).

This is also true!!!!

Thanks!

prko:
@dyfer @capital-G 
I think this is ready to be merged. What are your thoughts?

dyfer:
I can still see some instances of the work "hard-coded", is this intended?

prko:
Sorry, fixed!

dyfer:
Thanks!

Issue #6846, opened 2025-05-13
Python's ''' or """ in sclang


prko:
## Motivation
I think `.quote` can be omitted if sclang supports Python's `'''` or `"""`.

It will increase productivity and simplify the code lines.

## Description of Proposed Feature

- `"tell your" + "friends".quote + "not to tread onto the lawn"`-> `'''"tell your "friends" not to tread onto the lawn"'''`
- It will also be useful to `openOS`, `unixCmd`, `systemCmd`, `unixCmdGetStdOut` etc.

## Plan for Implementation
Could Python's `'''` or `"""` be implemented in sclang using a class method definition?
From my understanding it seems not to be possible, but could someone give me a hint on how to implement it in sclang without C++?

If modifying C++ is unavoidable, I cannot implement it myself. Sorry.

jamshark70:
> Could Python's ''' or """ be implemented in sclang using a class method definition?

Preprocessor could do it (see my new SC book chapter :wink: ) but I doubt we want to rely on the preprocessor for core syntax.

JordanHendersonMusic:
This would be an addition to the bison grammar and the lexer. This would probably be do-able, but the main lexing function is a bit of a handful as it if filled with gotos. 

If we were considering as an addition, it would be worth thinking about f strings too, these would be harder to implement, but much more beneficial.

JordanHendersonMusic:
Fyi, I have no intention of implementing this. Just use escape characters. I am working towards f strings though.

prko:
Thanks!

PR #7394, opened 2026-03-03
sclang: new lexer

JordanHendersonMusic:
## Purpose and Motivation

Our lexer, parser, compile are all muddled and very hard to use, both because of the coupling, and due to the large amount of global mutable state.

This is the lexer part of fixing this. 

I will also be doing a grammar refactor (it is currently a part of the error rework).

This is a little piece of that project.

This lexer is also ready for use with unicode so we can do stuff like  `∑(a)`.

We also have at least 3 lexers in this repo, the main one, one for skipping brackets, and the syntax highlighting one, the latter of which is often wrong. After this, there will only be one lexer because it is templated with no dependencies on sc internals. In fact, you could even compile it to web assembly if you wanted.

Currently this works pretty well.

There is a fuzz tester too that compares the old against the new, seems to work and not produce any more errors.

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing


JordanHendersonMusic:
I've tested this against Lucile's very big corpus, there are some failures, but in my opinion, it is the code that is wrong. Basically, when the source has invalid ASCII (a number less than 0), the new lexer interpreters this as the start of a multi-byte unicode codepoint, so it consumes many bytes, or throws an invalid UTF8 exception. There are a few of these in wslib and VirtualGamelan. IMO, this is an error in the code and I'm not even sure how these bytes got there in the first place.

JordanHendersonMusic:
An example of the lexing errors as they are now. Parsing errors will be done in another PR when the ast is refactored to pass on the correct source code location.
```
sc3> (



1 + 1

]

Lexing Error:
    1 │ (
      │ ^ Parenthises opened here...
    2 │ 
    3 │ 


    5 │ 1 + 1
    6 │ 
    7 │ ] 
      │ ^ ...was expected to be closed here with a ')'.


ERROR: syntax error, unexpected BADTOKEN, expecting DOTDOT or ':' or ',' or ')'
-> nil
```

telephon:
Great, this better error reporting is very useful.

> the new lexer interpreters this as the start of a multi-byte unicode codepoint, so it consumes many bytes, or throws an invalid UTF8 exception.

We should make sure that illegal whitespace characters are replaced before they hit the lexer. You often don't notice it on a US keyboard, but on various others, it happens a lot. On a German keyboard, you may hit option-8 for `{` and space while option is still pressed, then you get the whitespace character ` `, which should not invalidate your code, but does on some consoles.

JordanHendersonMusic:
Do you know what that character is called? I'm having issues googling it.

I'd rather add support for the space (assuming it's valid unicode). That way we don't need to do replacements, which depending on where the code came from can be annoying.

JordanHendersonMusic:
Ah never mind, it's the nbsp, I've actually already added support for this one so it just works as a normal space!

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+}
+
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// details
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+namespace details {
+
+template <typename CRTP> struct TextLocationLocation {
```
lnihlen: Why CRTP here? Is this because you are trying to enforce that all `T` implement `TextLocationLocation`? And you don't have constraints until C++20? This seems like overkill for a type that is likely to have very few implementations.

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+};
+struct FileCodeRange : details::TextLocationRange<FileCodeLocation> {
+    using TextLocationRange<FileCodeLocation>::TextLocationRange;
+    using TextLocationRange<FileCodeLocation>::operator=;
+};
+
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Codepoint --- represents a unicode code point. T
+// Node: we do not work with graphemes, so all multicodepoint unicode must be handled by the logic in the lexer.
```
lnihlen: s/Node/Note

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+static constexpr CodePoint invalid_utf8_flag = std::numeric_limits<CodePoint>::max();
+
+// Returns code point and number of utf8 bytes consumed
+inline std::tuple<CodePoint, std::uint8_t> char_sequence_to_codepoint(const char* source, size_t pos,
```
lnihlen: Is there no std C++ support for utf-8? Or a lightweight library for this? Seems like we shouldn't need to roll our own unicode support here to this extent.

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+    // returns begin and size
+    [[nodiscard]] std::tuple<const char*, std::size_t> source_range(const SourceCodeRange& range) const;
+
+    struct LineIter {
+        const CodePointStream& cps;
+        std::size_t start_line, end_line;
+        std::size_t current_line { start_line };
+        [[nodiscard]] inline std::optional<std::tuple<const char*, std::size_t, std::size_t>> operator()();
+    };
+    friend struct LineIter;
```
lnihlen: Since LineIter is declared within this class why does it need to be a friend? This seems like a violation of encapsulation.

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+
+    // Does not clear the new line cache. Goes back to the beginning.
+    void reset();
+
+
+private:
+    FileCodeLocation source_start_in_file; // Allows us to get the location in the file from the source code location.
+    SourceCodeLocation next; // Iterator through code, points to the next not the current.
+    const char* source; // Text, may not be null terminated. Only access this inside read.
+    size_t source_length; // Text length.
```
lnihlen: nit: I think it reads a little cleaner to give each of these comments their own line above the member declaration.

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+        const auto offset36 = std::max<int>(0, std::min<int>(36, radix)) - 11;
+
+        // TODO: I'd like to change this part of the language.
+        // We are only accepting valid radixs, that means what happens after can become another token..
+        //
+        //
+        //      |*| ---- integer '345'
+        // 2r012345
+        // |^^^|
+        //   |------------- floatradix '2r012'
```
lnihlen: Wouldn't this be an `intradix`?

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+        //
+        //      |**| ---- keybinop def:
+        // 2rabcdef:
```
lnihlen: `2rab` is not a valid radix number

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+    }
+
+#define literal_case(c)                                                                                                \
+    case (c):                                                                                                          \
+        if (auto r = action.template process<static_cast<TokenType>((c))>({ token_start, stream.end_token() }))        \
+            return *r;                                                                                                 \
+        else                                                                                                           \
+            goto discard_token
+
+    // Simple character literals
```
lnihlen: I called these "delimiters" in Hadron because I reserve the term "literals" to for actual literal values. Might help to pick a different term than literal here

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1381 @@
...
+        const SourceCodeLocation end = [&]() {
+            if (is_lower(first) || is_upper(first) || first == '_')
+                // Symbol began with alpha or underscore, therefore, can contain alpha, underscore, and numbers
+                return stream.advance_while(
+                    [](auto c) { return is_lower(c) || is_upper(c) || is_numeric(c) || c == '_'; });
+            if (is_numeric(first))
+                // Symbol began with a digit, therefore can only contain digits... why??
+                return stream.advance_while(is_numeric);
+
+            // This is weird, if it isn't either of the above, emit an empty symbol.
```
lnihlen: "this is weird" seems extraneous - an empty symbol is valid sc

lang/LangSource/Bison/lang11d
```diff
@@ -1456,26 +1456,6 @@ rwslotdef		: rwspec name
 					{ $$ = (intptr_t)newPyrVarDefNode((PyrSlotNode*)$2, (PyrParseNode*)$4, $1); }
 				;
 
-litdictslotdef	: listliteral ':' listliteral
```
lnihlen, line 1459: while I agree this should happen, can we move to a separate PR?

testsuite/lexer/compare_file/compare_file.cpp
```diff
@@ -0,0 +1,127 @@
+#include <sstream>
```
lnihlen: Just a general file comment - do we need to check in the test code comparing the old and new lexer? What's the plan to clean this up?

lnihlen (changes requested):
This is a massive PR so I don't know how thorough a review I'll be able to commit to. If there are pieces you particulary want feedback on please direct my attention to those.

Overall it largely looks great, this is a big step up for one of the crustier parts of the code base.

I would say this code is a "step up" on the level of C++ feature usage from what I see in most sc code and so idk how maintainable/approachable/readable this will be for others. I've tried to push back on some of the complexity when I didn't see it adding a lot of benefit, for example CRTP.

JordanHendersonMusic:
@lnihlen thanks for taking a look at this 😁 I'll try and address your points over this coming week!

include/lexer/lexer.hpp
JordanHendersonMusic: Good point, I've removed this in favour of a little bit of copy and paste.

Originally, I had the base class to avoid code duplication and to make them strong types. Then I needed to make them comparable, but need the derived class to only be comparable with instances of its own type, not its siblings — it doesn't make sense to compare a file location with a source code location (without converting first). This was just a way to enforce that.

testsuite/lexer/compare_file/compare_file.cpp
JordanHendersonMusic: Good point, I think I'll delete the fuzz tester and the a-b tests just before this is merged.

include/lexer/lexer.hpp
JordanHendersonMusic: I'm not aware of any in the std, beyond the deprecated conversions. 

The smallest library I've been able to find would be utf8proc - https://github.com/JuliaStrings/utf8proc , which is used by the Julia language, but even that seems like a lot of this one function. I've read it and updated the code based on some of their's as this didn't handle a couple of edge cases.

Perhaps if we need more unicode support in the future including that might be a worth it? 

What do you think?

JordanHendersonMusic:
@lnihlen I've updated the code accordingly reran the tests and everything seems good.

>  If there are pieces you particulary want feedback on please direct my attention to those.

Mainly readability of the lexer.hpp file as the tests should handle behaviour. I know I've used 'more' C++ that what is common in the code base, I think that it is okay as long as its for a reason (the CRTP use was too much).

The code in the PyrLexer file is mostly a mess. This is because I had to add this awkward shim to make the parser work with this lexer. I intended to fix this (and hopefully the rest of this file) in a separate PR.

include/lexer/lexer.hpp
lnihlen: I think sometimes the point of dependencies is less about code size minimization and more about limiting the scope of future maintenence to only things that this software is unique in being able to do. Said perhaps better, it's less about optimizing code and more about optimizing *focus*, keeping the focus of sclang on being a great compiler/interpreter, and limiting the number of "and alsos" that we have to add to that focus.

That said, if adding a dependency like this gives you too much heartburn, I'd suggest breaking this functionality out to a separate file, to at least help with the abstraction barrier between utf-8 processing and lexing.

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1407 @@
...
+*/
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Here is the public api and documentation. Everything in the 'details' namespace is private, do not use.
+
+// Unicode code point
+using CodePoint = std::uint32_t;
+inline constexpr CodePoint ascii_to_codepoint(char c);
+
+// Location and range in source code and file.
```
lnihlen: I note you never articulate in a comment what the difference between file and source locations are.

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1407 @@
...
+    return static_cast<CodePoint>(c);
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// CodePointStream - Walks through source code.
+// Its job is to return information about where we are in the source code, allow conversions to locations in the file
+//      and to return the 'character'.
+// It does not discard anything.
+// It cannot undo.
+// Annoyingly, it has state, because when in a command line context, the first token must be Interpret,
```
lnihlen: Could you provide two implementations of CodePointStream, one that provides the token and one that doesn't?

include/lexer/lexer.hpp
```diff
@@ -0,0 +1,1407 @@
...
+    ErSymbolQuoteUnclosed,
+    ErMultilineCommentUnclosed,
+    ErStringUnclosed,
+    ErMissingExponent,
+    ErInvalidToken,
+    ErInvalidUTF8,
+
+    ErFirstUserDefinedError = 5119
+};
+
```
lnihlen: I don't have a better suggestion top of mind, so this feedback seems a bit unfair, but it strikes me as odd that for the amount of thought and care you've given to the Lexer API, e.g. applying `[[nodiscard]]` and `const` everywhere appropriate, the Tokens are so brittle that anyone fiddling with the numeric values of the enums will break many of the functions below.

At minimum I'd suggest leaving a comment at the top of the enums warning folks to double-check the functions below if they change any numeric values. In Rust I just used nested enumerated types to give the same top-level grouping of tokens that you achieve here with numeric ranges.

I wonder, if there's a more C++-y way of adding some safety to making the enum to function mapping less brittle?

JordanHendersonMusic: Been thinking about this the past few days...
Aside from reflection (spent some time with zig a few months ago!), this is the simplest I could come up with.

```

\\ starts at zero, size of 4
Static constexpr tuple subrange1{get_range({}, 4)};
\\ starts after subrange1, with a size of 1024
Static constexpr tuple subrange2{get_range(subrange1, 1024)};


enum {
    a = get<1>(subrange1),
    b = get<2>(subrange1),
    c = ...,
    d = ...,
    e = get<1>(subrange2),
    f = get<2>(subrange2),
    ...1022 more entries
},
```

This way we check the ranges don't overlap as get_range creates them sequentially. Checking an enum entry is within a range becomes trivial. And the enum values are never out of range because get enforces this at compile time.

You could still break this if you tried, but you'd have to not follow the existing pattern so it would be obvious where the issue was.

I haven't tested this, but I think it's all constexpr with the compilers we have...

JordanHendersonMusic: Okay that didn't work. Instead I've used delimiters in the enum (because you can have different entries with the same value). So long as any future developers don't alter the START and END delimiters, nothing should break. Unless an entry needs to be in two separate categories...

lnihlen: Ah, I just remembered. Other compilers use X-Macros: https://hackaday.com/2023/03/16/the-x-macro-a-historic-preprocessor-hack/ (just search for "C++ X Macro" for more) to solve this exact problem.

JordanHendersonMusic: Hmm, I can't see how the X macro pattern would solve the issue of grouping or having a hierarchy of enum values. It would solve the string printing later.

include/lexer/lexer.hpp
JordanHendersonMusic: I don't quite understand what you are asking as the CodePointStream is a fancy iterator, it only returns CodePoints. To get the tokens you need to call the main lexer function.

include/lexer/lexer.hpp
JordanHendersonMusic: I've pulled this out into a separate file. I don't have an issue with including utf8proc, but it would require reworking a large amount of the cmake, turning this into a library, rather than a standalone header (+std). I'd rather avoid doing that for now, but might think about it in the future when/if needed.

include/lexer/lexer.hpp
JordanHendersonMusic: done.

JordanHendersonMusic:
@lnihlen I've pulled this out from a header only file to a static library. It also depends on utf8proc. I think this is much clearer! The code is the same, just how it is presented has changed. 

Just got the error printing to put back in.

include/lexer/lexer.hpp
JordanHendersonMusic: I quite like the delimiters in the enum, it is a bit 'C', rather than C++ but I think it works quite well.

langutils/sc_lexer/include/lexer.hpp
```diff
@@ -0,0 +1,499 @@
...
+        const int radix = str_to_int_for_radix(radix_str_b, sz, 10);
+        const auto offset10 = std::max<int>(0, std::min<int>(10, radix)) - 1;
+        const auto offset36 = std::max<int>(0, std::min<int>(36, radix)) - 11;
+
+        // TODO: I'd like to change this part of the language.
+        // We are only accepting valid radixs, that means what happens after can become another token..
+        //
+        //     |**| ---- integer '2345'
+        // 2r012345
+        // |^^|------------- floatradix '2r01'
```
lnihlen: nit: this is also an intradix

langutils/sc_lexer/include/tokens.hpp
```diff
@@ -0,0 +1,215 @@
...
+// TokenType
+// An enum covering all the types of things the lexer sees.
+// These are defined in range, DO NOT CHANGE THE RANGE DELIMITERS, just insert new cases inbetween them.
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// clang-format off
+enum struct TokenType : unsigned int {
+    EndOfFile = 0,
+
+    START_OF_IDENTIFIERS,
```
lnihlen, line 16: Really like where this ended up, nice job

lnihlen (approved):
LGTM, great work here Jordan!

JordanHendersonMusic:
@lnihlen thanks for reviewing this, the code has definitively improved as a result!

I've just got to fix the source code locations in the compiler (they are a mess), which will make the tests pass, and then I'll merge this.

prko:
I was too late.... Does contain this PR `'''` or `"""` discussed in #6846?

dyfer:
@JordanHendersonMusic I noticed a minor thing with posting the error (I'm guessing it's from this PR but I haven't looked closely):

When I try to run
```supercollider
var ęą = 123;
```

I get
```
Lexing Error:
Error: /path/to/SuperCollider.app/Contents/Resources/SCClassLibrary/deprecated/3.12/deprecated-3.12.sc:1:5
    1 │ var ęą = 123; 
      ┆     ^
ERROR: syntax error, unexpected BADTOKEN, expecting NAME or WHILE
  in interpreted text
  line 1 char 5:

-> nil
```

Note the line with the class library file - the error certainly does not come from there.

Otherwise the error looks fantastic :)

<details><summary>Here's the old error (before lexer refactor) for the same invocation</summary>

```
ERROR: syntax error, unexpected '=', expecting NAME or WHILE
  in interpreted text
  line 1 char 10:

  var ęą = 123; 
           ^
-> nil
```
</details>

Do you want me to file it as a separate issue?

JordanHendersonMusic:
> Do you want me to file it as a separate issue?

I've actually already fixed this in https://github.com/supercollider/supercollider/pull/7558

The reason this error and a few others have occured isn't an issue with the lexer per say, but because understand sc's compiler state is very hard — it uses many many globals, but they are all defined in different files and mutated at different times.

dyfer:
Thanks!

PR #7395, opened 2026-03-04
Synthdef version 3

Spacechild1:
## Purpose and Motivation

This PR introduces a new SynthDef version (v3). The only difference to the current version (v2) is that every synth definition starts with a `size (int32` field that tells the total size of the definition in bytes (including the field itself).

This means we can extend SynthDefs in the future by adding more fields without having to bump the version. Parsers can use the byte size to check how many of these new (optional) fields are present.

This is needed as a first step for #7402 because I want to add the block size and resample factor as fields to the SynthDef.

This PR also contains the following improvements and bug fixes:

scsynth:
- `ReadWriteMacros.h`: improve read functions (add bounds checking, avoid undefined behavior, correct return types) + add tests
- `GraphDef`: remove unnecessary code duplication between v1 and v2
- `GraphDef`: add missing bound checking in parsing functions!
- `GraphDef`: prevent memory leaks when exception is thrown
- modernize `BufColorAllocator` code

supernova:
- use `std::memcpy` for standard-conformant type punning
- skip variant data

sclang:
- fix path concatenation in `SynthDef -load` and `SynthDef -store`
- suppress error message in `WindowsPlatform - clearMetadata`
- check magic number and version number in `SynthDesc` parsing
- fix constructor of `SynthDefOld`
- fix error handling in `SynthDef -writeDef`
- skip variant data in `SynthDescs -readSynthDef`

## Types of changes

- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/Synth-Definition-File-Format.schelp
+++ b/HelpSource/Reference/Synth-Definition-File-Format.schelp
@@ -6,14 +6,15 @@ related:: Classes/SynthDef
 Synth definition files are read by the synth server and define collections of unit generators and their connections. These files are currently written by  the SuperCollider language application, but theoretically could be written by any program.
 Such a program would need knowledge of the SC unit generators and their characteristics, such as number of inputs and outputs and available calculation rates. The code to write these files is open and available in the SuperCollider language app.
 
-In the current SuperCollider 3.6 development version, there are two versions of SynthDef:
+In the current SuperCollider 3.15 development version, there are three versions of SynthDef:
 
 list::
 ## strong::SynthDef:: - The original SynthDef file format, backward compatible with .scsyndef files saved using earlier versions of SuperCollider.
 ## strong::SynthDef2:: - A revised version that extends SynthDef's capacity.
+## strong::SynthDef3:: - A another revision that adds new fields and makes the SynthDef format extendable.
 ::
 
-This document describes SynthDef2's format. See link::#Original SynthDef format:: for the differences between SynthDef and SynthDef2.
+This document describes SynthDef3's format. See link::#SynthDef2 format:: and link::#Original SynthDef format:: for the differences to the SynthDef2 resp. SynthDef1 format.
 
 section:: Basic types
 All data is stored big endian. All integers, unless otherwise noted, are signed. All data is packed, not padded or aligned.
@@ -37,6 +38,7 @@ list::
 
 a strong::synth-definition:: is :
 list::
+## int32 - the size of the synth definition in bytes (including this field)
 ## pstring - the name of the synth definition
 ## int32 - number of constants (K)
 ## [float32] * K - constant values
@@ -91,6 +93,8 @@ list::
 ## [float32] * P - variant initial parameter values
 ::
 
+note::Developers can extend the SynthDef3 format by adding new field to the emphasis::end:: of the synth definition. Parsers can use the size field to determine how many of these new fields are present.::
+
 section:: File Format as Tree
 
 soft::added by Jonatan Liljedahl::
@@ -102,6 +106,7 @@ tree::
 ## int16 - number of synth definitions in this file (D).
 ## [ strong::synth-definition:: ] * D
     tree::
+    ## int32 - the size of the synth definition in bytes (including this field)
     ## pstring - the name of the synth definition
     ## int32 - number of constants (K)
     ## [float32] * K - constant values
@@ -146,6 +151,10 @@ tree::
         ::
     ::
 ::
+
+section:: SynthDef2 format
+
+The SynthDef2 format omits the size field at the beginning of a synth definition as well as any fields after the variant specs.
 
 section:: Original SynthDef format
 
--- a/HelpSource/Reference/Synth-Definition-File-Format.schelp
+++ b/HelpSource/Reference/Synth-Definition-File-Format.schelp
@@ -31,7 +31,7 @@ section:: File Format
 a strong::synth-definition-file:: is :
 list::
 ## int32 - four byte file type id containing the ASCII characters: "SCgf"
-## int32 - file version, currently 2.
+## int32 - file version, currently 3.
 ## int16 - number of synth definitions in this file (D).
 ## [ strong::synth-definition:: ] * D
 ::
@@ -102,7 +102,7 @@ soft::added by Jonatan Liljedahl::
 subsection::synth-definition-file
 tree::
 ## int32 - four byte file type id containing the ASCII characters: "SCgf"
-## int32 - file version, currently 2.
+## int32 - file version, currently 3.
 ## int16 - number of synth definitions in this file (D).
 ## [ strong::synth-definition:: ] * D
     tree::
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -302,6 +302,7 @@ SynthDef {
 		this.asArray.writeDef(stream);
 		^stream.collection;
 	}
+
 	writeDefFile { arg dir, overwrite(true), mdPlugin;
 		var desc, defFileExistedBefore;
 		if((metadata.tryPerform(\at, \shouldNotSend) ? false).not) {
@@ -323,11 +324,22 @@ SynthDef {
 		}
 	}
 
-	writeDef { arg file;
+	writeDef { arg file, version=3;
 		// This describes the file format for the synthdef files.
-		var allControlNamesTemp, allControlNamesMap;
+		var allControlNamesTemp, allControlNamesMap, byteSize, startPos, endPos;
 
 		try {
+			if (version < 2 or: { version > 3 }) {
+				Error("version number out of range").throw
+			};
+
+			if (version > 2) {
+				// save current stream position so we can calculate the SynthDef size.
+				startPos = file.pos;
+				// the SynthDef size in bytes. This will be set at the very end.
+				file.putInt32(0);
+			};
+
 			file.putPascalString(name.asString);
 
 			this.writeConstants(file);
@@ -394,6 +406,17 @@ SynthDef {
 					};
 				};
 			};
+
+			if (version > 2) {
+				// calculate total size in bytes
+				endPos = file.pos;
+				byteSize = endPos - startPos;
+				// overwrite the size field in the beginning
+				file.pos = startPos;
+				file.putInt32(byteSize);
+				// restore stream position
+				file.pos = endPos;
+			}
 		} { // catch
 			arg e;
 			Error("SynthDef: could not write def: %".format(e.what())).throw;
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -628,14 +628,14 @@ SynthDef {
 			// should remember what dir synthDef was written to
 			dir = dir ? synthDefDir;
 			this.writeDefFile(dir);
-			server.sendMsg("/d_load", dir ++ name ++ ".scsyndef", completionMsg)
+			server.sendMsg("/d_load", dir +/+ name ++ ".scsyndef", completionMsg)
 		};
 	}
 
 	// write to file and make synth description
 	store { arg libname=\global, dir(synthDefDir), completionMsg, mdPlugin;
 		var lib = SynthDescLib.getLib(libname);
-		var file, path = dir ++ name ++ ".scsyndef";
+		var file, path = dir +/+ name ++ ".scsyndef";
 		if(metadata.falseAt(\shouldNotSend)) {
 			protect {
 				var bytes, desc;
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -375,8 +375,7 @@ SynthDef {
 
 					varname = name ++ "." ++ varname;
 					if (varname.size > 32) {
-						Post << "variant '" << varname << "' name too long.\n";
-						^nil
+						Error("variant '%' name too long".format(varname)).throw;
 					};
 					varcontrols = controls.copy;
 					pairs.pairsDo { |cname, values|
@@ -385,9 +384,8 @@ SynthDef {
 						if (cn.notNil) {
 							values = values.asArray;
 							if (values.size > cn.defaultValue.asArray.size) {
-								postf("variant: '%' control: '%' size mismatch.\n",
-									varname, cname);
-								^nil
+								Error("variant: '%' control: '%' size mismatch"
+									.format(varname, cname)).throw;
 							}{
 								index = cn.index;
 								values.do {|val, i|
@@ -395,9 +393,8 @@ SynthDef {
 								}
 							}
 						}{
-							postf("variant: '%' control: '%' not found.\n",
-								varname, cname);
-							^nil
+							Error("variant: '%' control: '%' not found"
+								.format(varname, cname)).throw;
 						}
 					};
 					file.putPascalString(varname);
@@ -417,8 +414,7 @@ SynthDef {
 				// restore stream position
 				file.pos = endPos;
 			}
-		} { // catch
-			arg e;
+		} { arg e; // catch
 			Error("SynthDef: could not write def: %".format(e.what())).throw;
 		}
 	}
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -330,7 +330,7 @@ SynthDef {
 
 		try {
 			if (version < 2 or: { version > 3 }) {
-				Error("version number out of range").throw
+				Error("version number" + version + "out of range").throw
 			};
 
 			if (version > 2) {
--- a/SCClassLibrary/Common/Collections/Collection.sc
+++ b/SCClassLibrary/Common/Collections/Collection.sc
@@ -664,7 +664,7 @@ Collection {
                 outliers, this.size, min, max
             )
         };
-        
+
 		^freqs;
 	}
 
@@ -712,12 +712,12 @@ Collection {
 
 	// Synth support
 
-	writeDef { | file |
+	writeDef { | file, version=3 |
 		file.putString("SCgf");
-		file.putInt32(2); // file version
+		file.putInt32(version); // file version
 		file.putInt16(this.size); // number of defs in file.
 
-		this.do { | item | item.writeDef(file); }
+		this.do { | item | item.writeDef(file, version); }
 	}
 
 	writeInputSpec { | file, synthDef |
--- a/SCClassLibrary/Common/Collections/Collection.sc
+++ b/SCClassLibrary/Common/Collections/Collection.sc
@@ -713,6 +713,10 @@ Collection {
 	// Synth support
 
 	writeDef { | file, version=3 |
+		if (version < 2 or: { version > 3 }) {
+			Error("version number" + version + "out of range").throw
+		};
+
 		file.putString("SCgf");
 		file.putInt32(version); // file version
 		file.putInt16(this.size); // number of defs in file.
--- a/SCClassLibrary/Platform/windows/WindowsPlatform.sc
+++ b/SCClassLibrary/Platform/windows/WindowsPlatform.sc
@@ -32,7 +32,8 @@ WindowsPlatform : Platform {
 	}
 	clearMetadata { |path|
 		path = path.splitext[0].do({ |chr, i| if(chr == $/) { path[i] = $\\.asAscii } });
-		"del %%.*meta%".format(34.asAscii, path, 34.asAscii).systemCmd;
+		// suppress error message if no files match the given pattern
+		"del \"%.*meta\" 2>nul".format(path).systemCmd;
 	}
 
 	killProcessByID { |pid, force = true, subprocesses = true|
--- a/server/scsynth/ReadWriteMacros.h
+++ b/server/scsynth/ReadWriteMacros.h
@@ -71,17 +71,17 @@ inline void readData(FILE* file, char* outData, size_t inLength) {
         throw std::runtime_error("readData: read != inLength");
 }
 
-inline int32 readInt8(char*& buf) {
+inline int32 readInt8(const char*& buf) {
     int32 res = *buf++;
     return res;
 }
 
-inline uint32 readUInt8(char*& buf) {
+inline uint32 readUInt8(const char*& buf) {
     uint8 res = (uint8)*buf++;
     return (uint32)res;
 }
 
-inline int32 readInt16_be(char*& buf) {
+inline int32 readInt16_be(const char*& buf) {
     uint16 c = readInt8(buf);
     uint16 d = readInt8(buf);
 
@@ -89,7 +89,7 @@ inline int32 readInt16_be(char*& buf) {
     return (int16)res;
 }
 
-inline int32 readInt32_be(char*& buf) {
+inline int32 readInt32_be(const char*& buf) {
     int32 a = readInt8(buf);
     int32 b = readInt8(buf);
     int32 c = readInt8(buf);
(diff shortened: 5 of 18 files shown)


server/scsynth/ReadWriteMacros.h
```diff
@@ -71,45 +71,57 @@ inline void readData(FILE* file, char* outData, size_t inLength) {
...
+    return res;
 }
 
-inline int32 readInt16_be(char*& buf) {
-    int32 c = readInt8(buf);
-    int32 d = readInt8(buf);
+inline int32 readInt16_be(const char*& buf, const char* end) {
+    checkBufferSpace(buf, end, 2);
+    uint16 c = buf[0];
+    uint16 d = buf[1];
```
JordanHendersonMusic, line 85: I am always unsure what integer promotion does in these case. We are going from a signed char to an unsigned 16 bit int. Does this behave the same as before, as we converted to a signed value?

server/scsynth/ReadWriteMacros.h
```diff
@@ -71,45 +71,57 @@ inline void readData(FILE* file, char* outData, size_t inLength) {
...
+    checkBufferSpace(buf, end, 4);
+    uint32 a = buf[0];
+    uint32 b = buf[1];
+    uint32 c = buf[2];
+    uint32 d = buf[3];
+    buf += 4;
 
-    int32 res = ((a & 255) << 24) | ((b & 255) << 16) | ((c & 255) << 8) | (d & 255);
-    return res;
+    uint32 res = ((a & 255) << 24) | ((b & 255) << 16) | ((c & 255) << 8) | (d & 255);
```
JordanHendersonMusic, line 100: Do we need all this masking? They came from a char, they can't be bigger than 255.

JordanHendersonMusic (commented):
World also be great to have some tests of the new format. Perhaps you could easily adapt some of the existing synthdefs in the test suite and make them compile in both versions?

server/scsynth/ReadWriteMacros.h
Spacechild1: Yes. Going from a smaller to a larger integer type always keeps the bit representation of the original value. (Internally, the `char` is first converted to `uint8` and then promoted to `uint16`.)

Also note that `char` may be signed or unsigned, depending on the platform. On ARM, `char` is generally unsigned (except on macOS/iOS).

Spacechild1:
> World also be great to have some tests of the new format. Perhaps you could easily adapt some of the existing synthdefs in the test suite and make them compile in both versions?

The new format is used by default. I think what we need to test is the *old* format. I will take some SynthDefs, compile them in both versions, read them back as `SynthDescs` and then compare the results.

server/scsynth/ReadWriteMacros.h
Spacechild1: Yes, the masking is necessary. The reason is that when converting from `char` to `uint32` the result can be larger than 255 when the `char` value was negative (because of two's complement). Masking removes these extra 1s:

-127 -> `uint32` -> 1111111111111111111111111110000001 (= 4294967169) ->& 255 -> 10000001 (= 129)

Alternatively, I would have to explicitly cast `buf[N]` to `uint8` before storing in `uint32`. What's better in your opinion? Either way, I should add a comment that says that the masking resp. the casts are really necessary.

server/scsynth/ReadWriteMacros.h
Spacechild1: > (Internally, the char is first converted to uint8 and then promoted to uint16.)

D'oh, it's actually the other way around! Promotion from (signed) char to `uint16` is done in two steps:
1. sign extend `char` to `int16`
2. convert `int16` to `uint16` with modulo arithmetic.

That's also why the masking below is necessary.

server/scsynth/ReadWriteMacros.h
JordanHendersonMusic: Is using memcpy easier here then? That's what bit_cast does. 

I don't know what is better, I just think bit_cast is generally clearer, sign extension and integer promotion confuses me, never mind arch differences.

Re signed/unsigned char, can you do an `if constexpr`. I'm more asking here because I'm doing something similar to get my new lexer to work with utf8 - so we can have unicode operators and whatnot.

JordanHendersonMusic:
>  I think what we need to test is the old format. 

Yes sorry that's what I meant. Can the class lib produce the old format?

Spacechild1:
> Can the class lib produce the old format?

Yes, it can! That's exactly why I added the `version` argument to the `writeDef` method.

server/scsynth/ReadWriteMacros.h
Spacechild1: > Is using memcpy easier here then? That's what bit_cast does.

The issue is that with `memcpy` the behavior depends on the endianess of the host machine. You'd need to check the endianess at compile time. If the host machine is little endian, you need to read the bytes in the inverse direction (= byte swap), otherwise you have to read them in order. Practically, all the platforms we care about are little endian, but still... We *could* add some platform specific defines for the host endianess, but I think this out of scope for this PR.

The nice thing about the bitshifting approach is that it works regardless of endianess. Personally, I would leave it like it is. I only really changed the type from `int32` / `int16` to `uint32` / `uint16` because left-shifting negative integers is actually undefined behavior! The current code should be well-defined on all platforms.

Once we upgrade to C++20, we could use `std::endian` + `std::memcpy`:
```cpp
inline int32 readInt32_be(const char*& buf, const char* end) {
    checkBufferSpace(buf, end, 4);
    int32 result;
    if constexpr (std::endian::native == std::endian::little) {
        // byte-swap
        char temp[4];
        temp[0] = buf[3];
        temp[1] = buf[2];
        temp[2] = buf[1];
        temp[3] = buf[0];
        std::memcpy(&result, temp, 4);
    } else {
        std::memcpy(&result, buf, 4);
    }
    return result;
}
```

In C++23 you can also use `std::byteswap`:
```cpp
inline int32 readInt32_be(const char*& buf, const char* end) {
    checkBufferSpace(buf, end, 4);
    int32 result;
    std::memcpy(&result, buf, 4);
    if constexpr (std::endian::native == std::endian::little) {
        return std::byteswap(result);
    } else {
        return result;
    }
}
```

> Re signed/unsigned char, can you do an if constexpr.

What for? The behavior is the same anyway.

JordanHendersonMusic: Nice, thanks for explaining!

Spacechild1: I just added C++ tests for all deserialization functions, just to make sure everything works as expected.

I did not test the `File*` overloads, but they are not really used anywhere. Maybe I can just remove these?

Spacechild1:
EDIT: oh my, turns out that `SynthDefOld -new` was completely wrong. Apparently nobody ever tried to actually create an instance of this class :-D

---

@JordanHendersonMusic I've tried to write some Unit tests to check that SynthDef version 1 and version 2 are still working.

For this purpose I'm trying to create SynthDef version 1 with the `SynthDefOld` version:
```sc
(
SynthDefOld(\test_v1_1, { |out, value=1| Out.kr(out, value) })
)
```
However, I get the following error:
```
ERROR: Message 'def' not understood.
RECEIVER:
   nil
ARGS:
KEYWORD ARGUMENTS:
PATH: C:/Users/Christof/Documents/SuperCollider/synthdef-v3-test.scd

PROTECTED CALL STACK:
	Meta_MethodError:new	0000016F36BE3BC0
		arg this = DoesNotUnderstandError
		arg what = nil
		arg receiver = nil
	Meta_DoesNotUnderstandError:new	0000016F36BE5F80
		arg this = DoesNotUnderstandError
		arg receiver = nil
		arg selector = def
		arg args = []
		arg keywordArgumentPairs = []
	Object:doesNotUnderstand	0000016F36BA6A80
		arg this = nil
		arg selector = def
		arg args = nil
		arg kwargs = nil
	SynthDef:addControlsFromArgsOfFunc	0000016F35FA2C80
		arg this = a SynthDefOld
		arg func = nil
		arg rates = nil
		arg skipArgs = 0
		var def = nil
		var names = nil
		var values = nil
		var argNames = nil
		var specs = nil
	SynthDef:buildUgenGraph	0000016F35FA2700
		arg this = a SynthDefOld
		arg func = nil
		arg rates = nil
		arg prependArgs = []
		var result = nil
		var saveControlNames = nil
	a FunctionDef	0000016F35FA1000
	Function:prTry	0000016F35077AC0
		arg this = a Function
		var result = nil
		var thread = a Thread
		var next = nil
		var wasInProtectedFunc = false
```
This is because the graph function passed to `addControlsFromArgsOfFunc` is `nil`. However, I do not understand why this is happening. I've added `postln` statements to all methods between `SynthDefOld - new` and `SynthDef -addControlsFromArgsOfFunc`. In the former, the graph function argument is still valid, but already in `SynthDef -build`, the graph function argument becomes `nil`! I have no idea what's going on... Can you maybe help me?

Spacechild1:
The arm clang linux builds currently fail because of an (accidental) C++20-ism in OpcodeDetails.h: https://github.com/supercollider/supercollider/actions/runs/22724943805/job/65897468678?pr=7395#step:9:2966. Ping @JordanHendersonMusic

Spacechild1:
Added client-side and server-side tests!

JordanHendersonMusic:
The error is here. 
```
/home/runner/work/supercollider/supercollider/testsuite/server/scsynth/readwrite_test.cpp:10:24: error: non-constant-expression cannot be narrowed from type 'int8' (aka 'signed char') to 'char' in initializer list [-Wc++11-narrowing]
   10 |     char buffer[1] = { a };
      |                        ^
/home/runner/work/supercollider/supercollider/testsuite/server/scsynth/readwrite_test.cpp:10:24: note: insert an explicit cast to silence this issue
   10 |     char buffer[1] = { a };
      |                        ^
      |                        static_cast<char>( )
```

You talking about the byte thing? Yes I need to fix this, but really I want to get rid of the amount of warnings and make any new one errors because it's hard to notice when you make a new one.

Spacechild1:
> The error is here.

D'oh! I got confused by all the warnings. BTW, this is a nice demonstration how `char` is only unsigned on Linux ARM :)

telephon:
>Parsers can use the byte size to check how many of these new (optional) fields are present.
Just one question: if we were to add fields later, could they be of any type, or is there a constraint?

Spacechild1:
> Just one question: if we were to add fields later, could they be of any type, or is there a constraint?

They can be any type! As long as new fields are always added at the end of synth-definition, parsers can simply check if there are bytes left before trying to parse these new (optional) fields.

SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -628,14 +647,14 @@ SynthDef {
 			// should remember what dir synthDef was written to
 			dir = dir ? synthDefDir;
 			this.writeDefFile(dir);
-			server.sendMsg("/d_load", dir ++ name ++ ".scsyndef", completionMsg)
+			server.sendMsg("/d_load", dir +/+ name ++ ".scsyndef", completionMsg)
```
JordanHendersonMusic, line 650: Does this mean we didn't have the directory before? Is there a chance this could break someone's project then?

SCClassLibrary/Common/Collections/Collection.sc
```diff
@@ -664,7 +664,7 @@ Collection {
                 outliers, this.size, min, max
             )
         };
-
```
JordanHendersonMusic, line 667: White space creep

SCClassLibrary/Platform/windows/WindowsPlatform.sc
```diff
@@ -32,7 +32,8 @@ WindowsPlatform : Platform {
 	}
 	clearMetadata { |path|
 		path = path.splitext[0].do({ |chr, i| if(chr == $/) { path[i] = $\\.asAscii } });
-		"del %%.*meta%".format(34.asAscii, path, 34.asAscii).systemCmd;
+		// suppress error message if no files match the given pattern
+		"del \"%.*meta\" 2>nul".format(path).systemCmd;
```
JordanHendersonMusic, line 36: I don't know what this does, would you mind commenting on it?

server/scsynth/SC_GraphDef.cpp
```diff
@@ -242,96 +201,46 @@ typedef struct IndexMap {
...
-    }
-}
-
-// ver 1
-
-inline static void calcParamSpecs1(GraphDef* graphDef, char*& buffer) {
-    if (graphDef->mNumParamSpecs) {
-        int hashTableSize = NEXTPOWEROFTWO(graphDef->mNumParamSpecs);
+inline static void calcParamSpecs(GraphDef* graphDef, const char*& buffer, const char* end, int version) {
+    uint32 numSpecs = graphDef->mNumParamSpecs;
```
JordanHendersonMusic, line 205: I know we don't have any guidelines, but my preference is for everything that doesn't change to be declared const, but this is mostly coming from working in the Lang where mutable variables are overused. Since this one is long lasting, I think it's much easier to read if it is const, but I wouldn't hold up the pr if you think this is trivial.

server/scsynth/SC_GraphDef.cpp
```diff
@@ -726,108 +564,44 @@ void GraphDef_Dump(GraphDef* inGraphDef) {
...
-                stack = nullptr;
-                throw std::runtime_error("buffer coloring error: reallocation during release failed.");
-            } else {
-                stack = tmpstack;
-            }
-            memset(stack + stackMaxSize, 0, stackMaxSize * sizeof(int16));
-            stackMaxSize *= 2;
-        }
-        stack[stackPtr++] = inIndex;
+    if (--mRefs[inIndex] == 0) {
```
JordanHendersonMusic, line 603: More of a personal thing, but this syntax confused me at first, I've never seen the combination of index and pre-decrement before now.

server/supernova/sc/sc_synthdef.cpp
```diff
@@ -116,25 +119,46 @@ int32_t read_int(const char*& buffer, const char* buffer_end, int size) {
...
-#else
-            ret.emplace_back(buffer, buffer_end, version);
-#endif
-        } catch (std::exception const& e) { std::cout << "Exception when reading synthdef: " << e.what() << std::endl; }
+    try {
+        std::vector<sc_synthdef> result;
+
+        // check header ('SCgf')
+        int32 magic = read_int32(buffer, buffer_end);
+        if (magic != (('S' << 24) | ('C' << 16) | ('g' << 8) | 'f'))
```
JordanHendersonMusic, line 127: Might consider pulling this magic check into a function as it is used a lot. Could also get rid of the magic number of 4 in a few lines too.

server/supernova/sc/sc_synthdef.cpp
```diff
@@ -116,25 +119,46 @@ int32_t read_int(const char*& buffer, const char* buffer_end, int size) {
...
+                const char* synthdef_end = buffer + synthdef_size - 4;
+                if (synthdef_end > buffer_end)
+                    throw std::runtime_error("wrong synthdef size");
+
+                result.emplace_back(buffer, synthdef_end, version);
+
+                buffer = synthdef_end;
+            }
+        } else {
+            for (int i = 0; i != definition_count; ++i) {
```
JordanHendersonMusic, line 152: Can definition count ever be negative?

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: On the contrary! This change actually makes sure that you always get a valid path string. The previous code was broken for `dir` strings that didn't have a trailing slash. (The default dir does have a trailing slash, that's why it happened to work.)

SCClassLibrary/Common/Collections/Collection.sc
Spacechild1: This just removes some trailing whitespace. Is this a bad thing?

SCClassLibrary/Platform/windows/WindowsPlatform.sc
Spacechild1: the only real change here is the addition of `2>nul`. As the comment says, this suppresses error messages if no files match the given pattern.

(The `del` command posts an error message if the given pattern does not match any string. However, in this case we just want to any files of the given pattern, no matter if they exist or not.)

The reason we're only started to see these error messages with SC 3.15-dev is https://github.com/supercollider/supercollider/pull/6971

server/supernova/sc/sc_synthdef.cpp
Spacechild1: No, but `definition_count` is a signed type (here: `int16`). The original code also used `int` here.

server/supernova/sc/sc_synthdef.cpp
Spacechild1: It is used exactly once in scsynth and supernova respectively. For me this doesn't really count as "a lot" :)

server/scsynth/SC_GraphDef.cpp
Spacechild1: I just tried to match the style of the existing code base, which doesn't really use `const` values.

server/scsynth/SC_GraphDef.cpp
Spacechild1: Here I just kept the existing code. The syntax didn't really confuse me: we reference an array element and then we pre-decrement it.

SCClassLibrary/Common/Audio/SynthDef.sc
JordanHendersonMusic: Ah the default had one already, that makes sense then!

server/supernova/sc/sc_synthdef.cpp
JordanHendersonMusic: Okay if it can't be negative that's fine then. I try and put these kind of facts in as assets, they function sort of like comments to express pre and post conditions.

Spacechild1:
Are there any general concerns about this PR?

JordanHendersonMusic:
The only thing I can think of is whether there are any other changes we might want to make to the spec while updating?

Perhaps the p string could take an unsigned size value instead of signed one? That's the only thing I can think of...

Spacechild1:
> The only thing I can think of is whether there are any other changes we might want to make to the spec while updating?

I thought about that too, but I can't really think of anything important. We can still make changes after merging this PR, as long as we do it before the release of SC 3.15.

One thing that came to my mind: we could add the number of local soundbuffers to the SynthDef. This way we can actually allocate the `SndBuf` array as part of the `Graph` structure itself, like we do with all the other arrays. The `MaxLocalBufs` UGen is just a hack that wouldn't have been necessary if the old SynthDef format was extendable. With this PR, we can avoid such hacks in the future.

(Actually, I recently discovered that the whole `LocalBuf` system suffers from an off-by-one error that can crash the Server. Since I want to fix the logic anyway, it might be a good chance to also retire `MaxLocalBufs`. We probably still have to keep it around for backwards compatibility with SynthDef v2...)

> Perhaps the p string could take an unsigned size value instead of signed one?

AFAICT the size value is already unsigned.

HelpSource/Reference/Synth-Definition-File-Format.schelp
JordanHendersonMusic, line 26: Ah yes you are right, it is unsigned, its just call `int8` rather than `uint8`.

JordanHendersonMusic:
> Since I want to fix the logic anyway, it might be a good chance to also retire MaxLocalBufs. We probably still have to keep it around for backwards compatibility with SynthDef v2...)

Okay do you want to put that change in this PR or another? It would be annoying to have MaxLocalBufs in version 3, then have to create a version 4 to remove it. If you want to do it in another, would you mind making an issue and put it in the 3.15 project so it doesn't get forgotten about.

Other than that, I think this PR looks good.

HelpSource/Reference/Synth-Definition-File-Format.schelp
Spacechild1: Yes, but with the word **unsigned** in bold :)

Spacechild1:
> Okay do you want to put that change in this PR or another?

In another PR.

> would you mind making an issue and put it in the 3.15 project so it doesn't get forgotten about.

Here you go: https://github.com/supercollider/supercollider/issues/7403

testsuite/server/scsynth/readwrite_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+BOOST_AUTO_TEST_CASE(readInt8_test) {
+    int8 a = -120;
+
+    char buffer[1] = { (char)a };
+    const char* ptr = buffer;
+    const char* end = buffer + 1;
+
+    int8 b = readInt8(ptr, end);
+    BOOST_REQUIRE(a == b);
```
JordanHendersonMusic: Minor point: 

Its probably better to use BOOST_TEST(). Require will halt the test if it fails, the other will continue, since these tests are independent, it makes sense to not halt.

JordanHendersonMusic (approved):
I think this is good, its in a bit of the code base I am less familiar with though.

I do wonder about the testing, its quite minimal. Since the change was only to the size field, this seems fine, but in future it might be nice to have some more complete testing. Ideally, you shouldn't be able to crash the server, so we might consider doing so fuzz testing to just chuck bytes at it.

testsuite/server/scsynth/readwrite_test.cpp
Spacechild1: Thanks, I didn't know this! Done!

Spacechild1:
>  Ideally, you shouldn't be able to crash the server, so we might consider doing so fuzz testing to just chuck bytes at it.

Yeah, some fuzz testing could be nice. One idea I had is to create a SynthDef and then progressively remove bytes from the end and send the resulting corrupted SynthDefs to the Server. I might give this a try.

JordanHendersonMusic:
I've been using libfuzz (which come with clang) over here, super easy to setup. Found a bunch of odd things with it.

https://github.com/JordanHendersonMusic/supercollider/blob/0f62cfbdfaa1e3a3583bbc99b72c30535babc60f/testsuite/lexer/fuzz/lexer_token_fuzz.cpp

https://github.com/JordanHendersonMusic/supercollider/blob/0f62cfbdfaa1e3a3583bbc99b72c30535babc60f/testsuite/lexer/fuzz/CMakeLists.txt

Spacechild1:
I've added tests for corrupted SynthDef data, for the Client (`SynthDescLib-readStream`) and for both Servers. No crashes so far.

HelpSource/Reference/Synth-Definition-File-Format.schelp
JordanHendersonMusic: This should be updated to '3'.

Spacechild1: Thanks! Done.

Spacechild1:
@telephon @capital-G @dyfer Does this PR need any further discussion? Or can I merge it?

(Once this PR is merged, I can update the reblocking PR.)

telephon:
Fine for me!

capital-G:
> @telephon @capital-G @dyfer Does this PR need any further discussion? Or can I merge it?
> 
> (Once this PR is merged, I can update the reblocking PR.)

Regarding a versioned network protocol which is expandable, I could not not think of protobuf - are you familiar w/ their best practices - most of them are obvious, but maybe good to check them out -> https://protobuf.dev/best-practices/dos-donts/ 

I haven't read the code, but if you think it is fine (after skimming over the protobuf doc) you should merge it - any quirks will be discovered by usage in the dev branch. This is what the dev branch should be for :)
And thanks for doing the work! Also, I always appreciate more tests :)

Pinging @josephine-wolf-oberholtzer as she is probably interested in this PR and can maybe provide some further ideas regarding extensions to the SynthDef format?

Spacechild1:
I'll go ahead and merge this now so I can move on with the reblocking PR. After all, this only adds a size field before each synth definition. There doesn't seem to be any general concern about introducing a new SynthDef version.

If someones has ideas/wishes for SynthDef3, they can make an issue or PR. For brainstorming, there is already https://github.com/supercollider/supercollider/issues/7403.

Thanks to @JordanHendersonMusic for the extensive review!

PR #7396, opened 2026-03-04
Fix no discard warnings in slot conversions

JordanHendersonMusic:
## Purpose and Motivation
Satisfy warning of discarding error values in slot conversion.

This uses asserts as no error checking previously took place.

## Types of changes

- warning fixing

## To-do list

- [x] Code is tested


JordanHendersonMusic:
@Spacechild1 here is the new one.

Spacechild1 (approved):

PR #7397, opened 2026-03-04
sclang: delete implicit casting in slot constructor

JordanHendersonMusic:
## Description

You should always know what type you are putting in the slot, therefore, you should be explicit. C++ is not explicit, but does offer a way to enforce this by deleting a generic template function.

Also, adds ctests that fail to compile, ensuring this behavior doesn't creep into the code base. This is a bit weird, but check our the `pyrslot_*.cpp` files and the cmake list for how that works. Ideally, these would be `static_asserts` with `std::invokable` but I couldn't get it to work.

## Purpose and Motivation

Safety. Does your integer cast in the way you expect? What about your enum? What about your pointer? This attempts to remove the ambiguity.

### DIFF
--- a/lang/LangSource/PyrSlot.h
+++ b/lang/LangSource/PyrSlot.h
@@ -335,26 +335,25 @@ public:
         return { PrivateTag(), Tags::charTag, static_cast<uint64_t>(details::bit_cast<uint8_t>(c)) };
     }
 
-    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr* o) {
+    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr* o) noexcept {
         return { PrivateTag(), Tags::objHdrTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(o)) };
     }
-    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr& o) {
+    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr& o) noexcept {
         return { PrivateTag(), Tags::objHdrTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(&o)) };
     }
-    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr o) = delete;
 
-    [[nodiscard]] inline static PyrSlot make(PyrSymbol* o) {
+    [[nodiscard]] inline static PyrSlot make(PyrSymbol* o) noexcept {
         return { PrivateTag(), Tags::symTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(o)) };
     }
-    [[nodiscard]] inline static PyrSlot make(PyrSymbol& o) {
+    [[nodiscard]] inline static PyrSlot make(PyrSymbol& o) noexcept {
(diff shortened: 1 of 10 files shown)


lang/LangSource/PyrSlot.h
```diff
@@ -333,23 +334,43 @@ union PyrSlot {
...
         return { PrivateTag(), Tags::ptrTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(o)) };
     }
+    [[nodiscard]] inline static PyrSlot make() noexcept { return {}; }
     [[nodiscard]] inline static PyrSlot make(PyrNil) noexcept { return {}; }
     [[nodiscard]] inline static PyrSlot make(bool b) noexcept {
         return { PrivateTag(), b ? Tags::trueTag : Tags::falseTag };
     }
 
+    // This is a bit hard to read (SFINAE), but basically says 'If you aren't derived from PyrObjectHdr, we will not do
+    // any casts for you'. One day, when we upgrade to c++ 20, this can be replaced by a simple concept.
```
capital-G, line 367: should we start an issue for "what to do/improve when we switch to c++20"

lang/LangSource/PyrSlot.h
```diff
@@ -333,23 +334,43 @@ union PyrSlot {
...
+    }
+    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr o) = delete;
+
+    [[nodiscard]] inline static PyrSlot make(PyrSymbol* o) {
         return { PrivateTag(), Tags::symTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(o)) };
     }
+    [[nodiscard]] inline static PyrSlot make(PyrSymbol& o) {
+        return { PrivateTag(), Tags::symTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(&o)) };
+    }
+    [[nodiscard]] inline static PyrSlot make(PyrSymbol o) = delete;
```
JordanHendersonMusic, line 352: Delete this deleted function.

lang/LangSource/PyrSlot.h
```diff
@@ -333,23 +334,43 @@ union PyrSlot {
...
-    [[nodiscard]] inline static PyrSlot make(struct PyrObjectHdr* o) {
+
+    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr* o) {
         return { PrivateTag(), Tags::objHdrTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(o)) };
     }
-    [[nodiscard]] inline static PyrSlot make(struct PyrSymbol* o) {
+    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr& o) {
+        return { PrivateTag(), Tags::objHdrTag, static_cast<uint64_t>(reinterpret_cast<uintptr_t>(&o)) };
+    }
+    [[nodiscard]] inline static PyrSlot make(PyrObjectHdr o) = delete;
```
JordanHendersonMusic, line 344: And delete this

JordanHendersonMusic:
Since this is just stricter type checking, and we talked about this in the last dev meeting, could someone just give a review so I can merge?

capital-G (approved):
Here you go @JordanHendersonMusic

JordanHendersonMusic:
Thanks! 😊

Issue #7388, opened 2026-02-28
Default server binary on linux


HotwheelsSisyphus:
I recently noticed (the hard way) that on linux, the default for `Server.program` (set in `UnixPlatform`) is `"exec scsynth"`, i.e. it looks in the $PATH. 
This causes trouble when there exist multiple builds of sc on the same machine. For instance, I was scratching my head about why some UGen wasn't found in one build, only to notice that of course it was using the wrong scsynth binary... 

I wonder if it's reasonable to change the default to something like:
```
Server.program = "exec $(dirname $(readlink -f /proc/%/exe))/scsynth".format(thisProcess.pid)
```
I.e., check which binary is being used for sclang and use the scsynth in the same directory.
Which would assume that the user **wants** to use the scsynth that is in the same directory as the sclang currently being used; I think this is generally a reasonable assumption; and it makes it easier to handle multiple builds on the same computer.
And like the previous default for Server.program, it can of course be overwritten.

Opinions? I'm relatively new to linux I have to say, so there may be things I haven't thought about. But the problem with not-so-self-contained installations on linux seems to be acknowledged here: 

https://github.com/supercollider/supercollider/wiki/Path-searching

jamshark70:
I think it's worth doing if it's to ensure compatible behavior across platforms.

Currently:

- macOS: Run the scsynth process from within the app bundle: `Server.program = "exec %/scsynth".format((Platform.resourceDir +/+ "../Resources").shellQuote);`
- Windows: scsynth process from the resource directory: `Server.program = (Platform.resourceDir +/+ "scsynth.exe").quote;`
- Linux: scsynth process from $PATH (no version specificity)

The macOS command is a solid precedent to do something similar for Linux. On that basis alone, I'd support the change.

I don't have access to a Windows SC machine right now, so I'm not sure if the resource directory is specific to each installed version. I guess it probably is...? If so, then Linux is the odd one out and definitely should be updated. If not, then this should be addressed in Win also.

HotwheelsSisyphus:
I've looked at `common/SC_Filesystem_win.cpp`, andthis bit sets the windows resource dir:
```
Path SC_Filesystem::defaultResourceDirectory() {
    WCHAR buf[MAX_PATH];
    GetModuleFileNameW(nullptr, buf, MAX_PATH); 
    return Path(buf).parent_path();
}
```
[...]

PR #7398, opened 2026-03-04
Linux: Server.program to use scsynth bin from the same dir as sclang bin.

HotwheelsSisyphus:
For uniformity across platforms (analogous to win behavior) and less trouble with multiple builds on some systems:

From #7388: 

I recently noticed (the hard way) that on linux, the default for Server.program (set in UnixPlatform.sc) is "exec scsynth", i.e. it looks in the $PATH.
This causes trouble when there exist multiple builds of sc on the same machine. For instance, I was scratching my head about why some UGen wasn't found in one build, only to notice that of course it was using the wrong scsynth binary...

I wonder if it's reasonable to change the default to something like:
```
Server.program = "exec $(dirname $(readlink /proc/%/exe))/scsynth".format(thisProcess.pid)
```
I.e., check which binary is being used for sclang and use the scsynth in the same directory.
Which would assume that the user wants to use the scsynth that is in the same directory as the sclang currently being used; I think this is generally a reasonable assumption; and it makes it easier to handle multiple builds on the same computer.
And like the previous default for Server.program, it can of course be overwritten.

---------------------

Alternative, without asking the shell:
```
Server.program = "exec %/scsynth".format((Platform.resourceDir +/+ "../../bin").shellQuote);
```

I think the other suggestion is slightly better because afaik there is no easy way to build sc that would spread the binaries across multiple folders(?). I.e., `scide, sclang, scsynth` are reliably in the same folder, but the path to them from `resourceDir` might be different if `SC_DATA_DIR` has been set during build.  It could be that `SC_DATADIR`, which optionally sets the `resourceDir` at build time, might  be different from the install path (which determines where the binaries go). 

------------------------------

Compare win behavior:
`WindowsPlatform.sc` uses the scsynth in resourceDir 
```
Server.program = (Platform.resourceDir +/+ "scsynth.exe").quote;
```

and `resourceDir is defined as the one that sclang is in.
`SC_Filesystem_win.cpp`:
```
Path SC_Filesystem::defaultResourceDirectory() {
    WCHAR buf[MAX_PATH];
    GetModuleFileNameW(nullptr, buf, MAX_PATH);
    return Path(buf).parent_path();
}
```
OSX is more complicated but also sets resource dir to current path of sclang, and 
```
Server.program = "exec %/scsynth".format((Platform.resourceDir +/+ "../Resources").shellQuote);
```
where the scsynth isn't directly in the resourceDir but anyway within the same app package.

-------------

## types of change

- classlib, using shell commands from the gnu coreutils
- Breaking change (technically; for people who rely on $PATH/scsynth)
(what one **could** do to smoothe over that change is check whether a scsynth at the new default location exists, and if not, fall back to PATH and emit a warning)

## TODO

- Update documentation - [this](https://github.com/supercollider/supercollider/wiki/Path-searching)  mostly, which needs to be updated anyway
- I'm not terribly confident in my shell skillz, so someone who is should look this over. Of course it works fine for me, but idk what could go wrong etc.

### DIFF
--- a/SCClassLibrary/Platform/linux/LinuxPlatform.sc
+++ b/SCClassLibrary/Platform/linux/LinuxPlatform.sc
@@ -14,8 +14,9 @@ LinuxPlatform : UnixPlatform {
 
 		helpDir = this.systemAppSupportDir++"/Help";
 
-		// Server setup
-		Server.program = "exec $(dirname $(readlink /proc/%/exe))/scsynth".format(thisProcess.pid);
+		// Server setup. first looks for scsynth in the dir containing the sclang executable;
+		// if nothing is found, falls back to PATH
+		Server.program = "PATH=$(dirname $(readlink /proc/$PPID/exe)):$PATH; exec scsynth";
 
 		// Score setup
 		Score.program = Server.program;
--- a/SCClassLibrary/Platform/linux/LinuxPlatform.sc
+++ b/SCClassLibrary/Platform/linux/LinuxPlatform.sc
@@ -15,7 +15,7 @@ LinuxPlatform : UnixPlatform {
 		helpDir = this.systemAppSupportDir++"/Help";
 
 		// Server setup
-		Server.program = "exec scsynth";
+		Server.program = "exec $(dirname $(readlink /proc/%/exe))/scsynth".format(thisProcess.pid);
 
 		// Score setup
 		Score.program = Server.program;


capital-G:
The current linux has IMO a certain advantage: it is possible to change/influence the used scsynth from outside the language (e.g. I want this for the package manager - but I'll solve this through another PR).
But I see how this can create trouble when having multiple versions of SuperCollider installed (which would require something like virtual environments like in python - "yay...").

BTW - isn't  `/proc/self/exe` sufficient in this case?
Additionally: Instead of hardcoding the scsynth path, we could pre-pend the path of `sclang` to the `PATH` variable and then execute scsynth - in case it is available next to the `sclang` dir, this will be taken - if not it will be searched for in the `PATH` variable. Maybe this creates less headaches for linux maintainers who split sclang/scsynth/scide?

So something like

```shell
PATH=$(dirname $(readlink /proc/self/exe)):$PATH exec scsynth
```

HotwheelsSisyphus:
>  change/influence the used scsynth from outside the language

Oh, I simply hadn't thought about that...  

> BTW - isn't /proc/self/exe sufficient in this case?

apparently not, for all I can tell ... it gets you the path to the binary of the readlink process, not sclang. Maybe there are other, better ways to do this... as mentioned I'm not super confident in my shell skills.

> Additionally: Instead of hardcoding the scsynth path, we could pre-pend the path of sclang to the PATH variable and then execute scsynth 

maybe I'm missing something, but how would you then override _that_ externally? Because sclang prepends to PATH  immediately before calling exec scsynth. One would have to separate the set-scsynth path command from the exec scsynth  command...

capital-G:
> > BTW - isn't /proc/self/exe sufficient in this case?
> 
> apparently not, for all I can tell ... it gets you the path to the binary of the readlink process, not sclang. Maybe there are other, better ways to do this... as mentioned I'm not super confident in my shell skills.

Ah - we would probably need to write this to a dummy variable in this case - but I think at this point we should simply go w/ the pid approach you have chosen :)

> > Additionally: Instead of hardcoding the scsynth path, we could pre-pend the path of sclang to the PATH variable and then execute scsynth
> 
> maybe I'm missing something, but how would you then override _that_ externally? Because sclang prepends to PATH immediately before calling exec scsynth. One would have to separate the set-scsynth path command from the exec scsynth command...

It is not possible to overwrite that. I mentioned this as it allows for fallback scsynth locations provided by the system - maybe this makes some setups easier? In the future I'd like to introduce something like `SCLANG_PATH`, `SCSYNTH_PATH` etc which allow to make their paths explicit and would take precedence if present - but this is another PR :)

Spacechild1:
Generally, this is a good idea! I had stumbled upon the same issue just recently when I tried to test a local develop build and the IDE would keep starting the system-installed `scsynth`...

> PATH=$(dirname $(readlink /proc/self/exe)):$PATH exec scsynth

This sounds like a good idea!

Stupid question: is `sclang` actually started from the same directory as the IDE? I'm too lazy to get my Linux machine to test.

HotwheelsSisyphus:
> is sclang actually started from the same directory as the IDE? 

Yes, at least by default, the build puts all the binaries in the same directory, and same for the packages I've tried. I don't know if there is maybe some build flag that would change that, but if so it's not terribly obvious.

Btw, that would be the nicest thing: To be able to set the paths for the binaries (and other paths, e.g. the userConfigDir, which is currently very hardcoded) somewhere in a menu in the IDE...

re the $PATH option: 
export doesn't persist, but String.setenv does. Is something like this reasonable in principle?
```
startup {
var prependSclangDirToPATH = "echo $(dirname $(readlink /proc/$PPID/exe)):$PATH";
"PATH".setenv(prependSclangDirToPATH.unixCmdGetStdOut);
Server.program = "exec scsynth"
}
```
I.e., the sclang dir is prepended to PATH at startup _but_ PATH can still be changed before the server starts. So (if this works), one should be able to get the default _and_ change it externally as well.

-------

more about the /proc/self/exe:
$PPID is what we were looking for!
$PID is what one thinks it is (pid of the shell process); $PPID is pid of parent (sclang).

capital-G:
> export doesn't persist, but String.setenv does. Is something like this reasonable in principle?

Thats the very idea to not have this persistent but only apply this within one command. We don't want to tinker with the `PATH` variable of the user beyond the command.
This creates real trouble when having multiple SC versions as they now start competing with each other.

> I.e., the sclang dir is prepended to PATH at startup _but_ PATH can still be changed before the server starts. So (if this works), one should be able to get the default _and_ change it externally as well.

Not really - you pre-pend to the `PATH` variable and `scsynth` will be searched for iterative through all paths specified via `PATH`. Since you put something to the top, there is no way to influence this from outside the language.

IMO it is best to patch the `PATH` variable for the `scsynth` command as specified above and later introduce an environment variable which allows to explicitly set the path, which falls back to searching for scsynth, sclang etc. using `PATH`.

> Btw, that would be the nicest thing: To be able to set the paths for the binaries (and other paths, e.g. the userConfigDir, which is currently very hardcoded) somewhere in a menu in the IDE...

Sounds like a good idea and shouldn't be too hard.

HotwheelsSisyphus:
Oh I see, then I think I just misunderstood your initial comment:

>  it is possible to change/influence the used scsynth from outside the language

How would you do this then? I assumed that by this you meant prepending to (or otherwise changing) sclang's PATH. 
My last suggestion had roughly this scenario in mind:
1. sclang inherits parent's PATH = [x,y,z]
2. sclang prepends its own bin directory, so PATH = [ b,x,y,z]
3. someone "changes/influences the used scsynth from outside the language" by setting sclangs PATH to whatever:
E.g. by prepending a specific directory  (PATH = [a,b,x,y,z]) or the whole PATH of the parent, again (sclang's PATH=[x,y,z,b,x,y,z])

capital-G:
> How would you do this then?

E.g. by having an env variable which takes precedence - but this would be another PR

```supercollider
var scsynthPath = "SCSYNTH_PATH".getenv;
Server.program = if(scsynthPath.notNil, {
	"exec %".format(scsynthPath);
}, {
	"PATH=$(dirname $(readlink /proc/self/exe)):$PATH exec scsynth";
});
```

> 3. someone "changes/influences the used scsynth from outside the language" by setting sclangs PATH to whatever:
>        E.g. by prepending a specific directory  (PATH = [a,b,x,y,z]) or the whole PATH of the parent, again (sclang's PATH=[x,y,z,b,x,y,z])

This is not how env variables work - they are not globally synced but instead inherited - thats why sourcing (i.e. `source ~/.zshrc`) is necessary. The only way to pass env variables to scide/sclang is setting them before starting scide/sclang.

HotwheelsSisyphus:
oh I see, now I get it! thanks for explaining.

SCClassLibrary/Platform/linux/LinuxPlatform.sc
```diff
@@ -14,8 +14,9 @@ LinuxPlatform : UnixPlatform {
 
 		helpDir = this.systemAppSupportDir++"/Help";
 
-		// Server setup
-		Server.program = "exec $(dirname $(readlink /proc/%/exe))/scsynth".format(thisProcess.pid);
+		// Server setup. first looks for scsynth in the dir containing the sclang executable;
+		// if nothing is found, falls back to PATH
+		Server.program = "PATH=$(dirname $(readlink /proc/$PPID/exe)):$PATH; exec scsynth";
```
capital-G, line 19: are you sure about the `;`?
I think this should be written w/o it b/c you want to modify the `PATH` variable for `exec scsynth` - by adding a `;` it will be split into two statements, so `PATH` is not modified for `exec scsynth` - but this is what we want.

edit: please double-check on this - I don't have a linux machine^^

HotwheelsSisyphus: yes, I did test that commit, and it works (with the semicolon).
The way I understand it, each `.unixCmd` (like the string here) is its own shell process, and within that, variables do persist. They just don't persist across separate `unixCmd`s. 
[UnixSupport](https://doc.sccode.org/Overviews/Methods.html#storeOn)

capital-G: Oh TIL env variables are persist across commands, so this is fine :)

capital-G (approved):

PR #7401, opened 2026-03-06
cmake: add folder property to all targets.

JordanHendersonMusic:
This makes IDE show the list of targets in a nice way, otherwise it is just a giant list.

## Purpose and Motivation

Creates folders for all targets, making is easier to find what you are looking for.

Also does the same for tests.

sc_targets
sc_targegts-tseetes

tets

## Types of changes

- CMake/IDE improvement.


capital-G (approved):
Good idea!
Also works in CLion

grafik

Should we maybe move the content of the main folder to the top level? Was looking for sclang for a moment^^

grafik

Feel free to adapt my suggestion or merge if you prefer yours :)

JordanHendersonMusic:
That's a good idea! I'll do that.

Issue #7370, opened 2026-02-15
Wrong behavior of previousPowerOfTwo and isPowerOfTwo methods


Spacechild1:
`a.previousPowerOf(2)` returns wrong results:

- if `a` is already a power-of-2, it should return the number itself, instead it returns the previous power-of-2.
  `nextPowerOf` behaves correctly in this case.
- ~~`1.previousPowerOf(2)` returns `0.5`, which is just wrong~~
  Actually this is correct because the method is not confined to integers. 0.5 is 2^(-1) and -1 comes before 0 (= 2^0 = 1)
- `0.previousPowerOf(2)` returns `0.0`, but `0.0` is not a power-of-2. (What should it return in this case?)
- `a.previousPowerOf(2)` returns `nan` for negative values. Maybe that's expected.

---

~~Also, `0.isPowerOfTwo` wrongly returns `true`.~~ This has been fixed by https://github.com/supercollider/supercollider/pull/7409

## Environment

* SuperCollider version: 3.14
* Operating system: Windows 10
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
4.previousPowerOf(2); // -> 2.0 - wrong!
4.nextPowerOfTwo// -> 4 - correct!
4.nextPowerOf(2)// -> 4.0 - correct!

0.previousPowerOf(2); // -> 0.0 - not a power-of-2

-1.previousPowerOf(2); // -> nan

// ------------------------------

0.isPowerOfTwo; // -> true - wrong!
```

HotwheelsSisyphus:
$0.5 = 1/2 = 2^{-1}$, so that bit isn't really outlandish; but since 1 is also a power of two, it should return that. 
I think the method needs to be corrected to 
```
pow(base, floor(log(this)/log(base))) 
```
as for how to handle negative bases, no idea. maybe just throw an error. Or return imaginary numbers, idk. 

the primitive isPowerOfTwo boils down to
```
(x & (x - 1)) == 0 
``` 
so that explains the 0. Maybe also test for x > 0 before calling the primitive?

Spacechild1:
>  `0.5 == 1 / 2 ==  2^(-1) , so that bit isn't really outlandish; 

Ah, you're right! I don't think about negative exponents. That at least explains the result.

> but since 1 is also a power of two, it should return that.

Exactly.

> the primitive isPowerOfTwo boils down to
> `(x & (x - 1)) == 0`
[...]

PR #7402, opened 2026-03-07
Synth reblocking and upsampling

Spacechild1:
This is a revamp of #6702. See the latter for previous discussions.

## Purpose and Motivation

This PR allows Synths to run at different block sizes or samplerates. This enables two important techniques that haven't been possible so far:
- tight feedback loops, down to a single sample
- upsampling + filtering to remove/reduce aliasing, e.g. for distortion, hard-syncing, osciallators and any other type of synthesis that is prone to aliasing

---

### For users

There are two new objects that can be used in SynthDefs:
- `Reblock`: the constructor receives the desired block size, either as a fixed number or as a Synth argument/control. If the value is 0, the Server block size is used. The new block size must be a power of two. Currently, it cannot be larger than the Server block size, i.e. we only support *downblocking*. This may change in the future.
- `Resample`: the constructor receives the resample factor, either as a fixed number or as a Synth argument/control. 1.0 means no resampling. The factor must be a power of two. Currently, it cannot be smaller than 1.0, i.e. we only support *upsampling*. This may change in the future.

*NOTE*: Supernova currently does not support Synth arguments/controls as arguments to `Reblock` and `Resample`, i.e. the block size and resample factor must be constants. This may change in the future.

Here are two practical examples:
```sc
// 1. Karplus-String with small block size
(
SynthDef(\delay, { |out, blockSize|
    var source, local;

    Reblock(blockSize);

    source = Decay.ar(Impulse.ar(2), 0.1) * WhiteNoise.ar(0.2);
    local = LocalIn.ar(2) + [source, 0]; // read feedback, add to source
    local = DelayC.ar(local, 0.5, MouseX.kr.pow(2) * 0.5); // delay sound

    // reverse channels to give ping pong effect, apply decay factor
    LocalOut.ar(local.reverse * MouseY.kr.pow(0.5).clip(0, 0.999));

    Out.ar(out, local);
}).add
)

// min. delay length: Server block size
Synth(\delay);
// try with smaller block sizes:
Synth(\delay, [ blockSize: 16 ]);
Synth(\delay, [ blockSize: 4 ]);
Synth(\delay, [ blockSize: 1 ]);

// 2. anti-aliasing with upsampling + filtering
(
SynthDef(\hardsync, { |out, upsample=1.0|
     var sig, cutoff = 18000;

    Resample(upsample);
    
    sig = SyncSaw.ar(MouseY.kr(1, 1000, 2), MouseX.kr(100, 4000, 2));
    sig = LPF.ar(sig, cutoff);
    sig = LPF.ar(sig, cutoff);
    Out.ar(out, sig * 0.05 ! 2);
}).add;
)

// lots of aliasing
Synth(\hardsync); 
// watch aliasing disappear with increasing upsample factors
Synth(\hardsync, [ upsample: 2.0 ]);
Synth(\hardsync, [ upsample: 4.0 ]);
Synth(\hardsync, [ upsample: 8.0 ]);
```

---

### For plugin authors

On the implementation side, this adds new `mBufRate` and `mFullRate` members to `Graph`. For "normal" Synths, these just point to the global `Rate` members in `World`. If the Synth is reblocking and/or upsampling, it has its own `Rate` instances.

UGens "just work" as long as they consistently use the `SAMPLERATE`, `BUFLENGTH`, `FULLRATE`, `FULLBUFLENGTH` etc. macros in `SC_Unit.h` resp. the corresponding C++ methods in `SC_Plugin.hpp`, because they now refer to the Graph's rates instead of the global World rates.

The only UGens that need to aware of reblocking and upsampling are I/O UGens because they are responsible for reblocking/resampling when reading from / writing to Busses. (Actually, this applies to all UGens that use Busses, but there are very few third-party plugins that do, as far as I am aware.)

---

### Implementation

This PR adds four new fields to the synth definition in SynthDef version 3:

- int32 - the block size. 0 = use Server block size. -1 = take the value from the given control index (see next field)
- int32 - the block size control index. Only used if the block size value above is -1
- float32 - the resample factor. -1.0 = take the value from the given control index (see next field)
- int32 - the resample control index. Only used if the resample factor value above is -1.0

Similarly, the (private) GraphDef` structure has four new fields:
```cpp
int32 mBlockSize;
uint32 mBlockSizeIndex;
float32 mResampleFactor;
uint32 mResampleIndex;
```

In `Graph_Ctor`, this is used to set the block size and resample factor either from the values itself or from the specified controls. 

In Supernova, we currently only accept literal values due to the way Synths are created. Currently, all Synth arguments are only parsed and set *after* the Synth has been fully constructed. At this point it is too late for setting the block size or sample rate. If the user attempts to use controls, we post a warning and use the default values. In the future, we may support using controls by refactoring the Synth construction code.

**NOTE**: this feature requires a break in the plugin API because I had to make changes to the (public) `Graph` struct. Fortunately, the plugin API has already been bumped in 3.15-dev because of #7329, so this PR should be included in 3.15 to avoid breaking the API again. See also this meta-issue: https://github.com/supercollider/supercollider/issues/6102

---

### Testing

It would be great if you could test this PR by running your existing SynthDefs at different combinations of block sizes and upsampling factors and check if they still work/sound as expected.

Here are two important guide lines to avoid false positives:

- the Synth sounds different at lower block sizes -> please check that this isn't also the case when setting the global Server block size to the same value (ideally with the official SC build)
- the Synth sounds different when upsampled -> please check that this isn't also the case when running the Server at the corresponding global sample rate  (ideally with the official SC build)

UGens might behave weirdly in a reblocked/upsampled Synth, but they should never crash the Server.

There have been no real issues reported so far with the original PR (#6702).

## Types of changes

- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Reblock.schelp
+++ b/HelpSource/Classes/Reblock.schelp
@@ -6,9 +6,9 @@ related:: Classes/Resample, Classes/SynthDef
 DESCRIPTION::
 For certain algorithms it is necessary to run a link::Classes/Synth:: at a small block size. Instead of decreasing the block size of the whole Server, which could be rather wasteful, you can use the Reblock object inside a link::Classes/SynthDef:: to only change the block size of specific Synths. (Keep in mind that very small block sizes will increase the CPU usage.)
 
-The block size can be a fixed number, or it can be set dynamically at runtime via a Synth argument/control.
+The block size can be a constant number or it can be set dynamically via a Synth control (except for Supernova). See link::#*new:: for more information.
 
-note::For technical reasons, Supernova only supports fixed block sizes.::
+note::For technical reasons, Supernova requires a fixed resampling factor. Trying to use a Synth control results in a warning.::
 
 For example, this can be used to implement very short feedback loops:
 
@@ -46,6 +46,16 @@ METHOD:: new
 Specify the block size of the Synth. You should only create this once per SynthDef. Duplicate instances generate a warning and are ignored.
 
 ARGUMENT:: blockSize
-the block size. A non-zero block size must be a power of two. If the value is 0, the Server block size will be used.
+the block size. This can either be a constant number (e.g. 1) or a Synth control. With the latter, the block size can be set independently for each Synth. However, it cannot be modulated after the Synth has been instantiated on the Server.
 
-Currently, only downblocking is supported, i.e. the block size must be smaller than or equal to the Server block size. Otherwise you get a warning.
+The block size must be a power of two. A value of 0 means that the Server block size should be used (= no reblocking).
+
+Currently, only downblocking is supported! A block size greater than the Server block size is ignored and results in a warning.
+
+DISCUSSION::
+
+Note that the Reblock object is not a UGen and there are certain restrictions on the types of arguments it accepts. In particular, you cannot use UGen expressions. (UGens are only evaluated after the Synth has been initialized. At this point, the block size must have already been set.) For example, the following does not work and will throw an error:
+code::
+// futile attempt at generating a random block size:
+Reblock(2 ** IRand(0, 4)); // -> throws an error
+::
--- a/HelpSource/Classes/Reblock.schelp
+++ b/HelpSource/Classes/Reblock.schelp
@@ -54,7 +54,7 @@ Currently, only downblocking is supported! A block size greater than the Server 
 
 DISCUSSION::
 
-Note that the Reblock object is not a UGen and there are certain restrictions on the types of arguments it accepts. In particular, you cannot use UGen expressions. (UGens are only evaluated after the Synth has been initialized. At this point, the block size must have already been set.) For example, the following does not work and will throw an error:
+Note that the Reblock object is strong::not:: a UGen and there are certain restrictions on the types of arguments it accepts. In particular, you cannot use UGen expressions. (UGens are only evaluated after the Synth has been initialized. At this point, the block size must have already been set.) For example, the following does not work and will throw an error:
 code::
 // futile attempt at generating a random block size:
 Reblock(2 ** IRand(0, 4)); // -> throws an error
--- a/HelpSource/Classes/Reblock.schelp
+++ b/HelpSource/Classes/Reblock.schelp
@@ -1,0 +1,51 @@
+TITLE:: Reblock
+summary:: Change the block size of a SynthDef/Synth
+categories:: Server>Abstractions
+related:: Classes/Resample, Classes/SynthDef
+
+DESCRIPTION::
+For certain algorithms it is necessary to run a link::Classes/Synth:: at a small block size. Instead of decreasing the block size of the whole Server, which could be rather wasteful, you can use the Reblock object inside a link::Classes/SynthDef:: to only change the block size of specific Synths. (Keep in mind that very small block sizes will increase the CPU usage.)
+
+The block size can be a fixed number, or it can be set dynamically at runtime via a Synth argument/control.
+
+note::For technical reasons, Supernova only supports fixed block sizes.::
+
+For example, this can be used to implement very short feedback loops:
+
+code::
+// Karplus-Strong with different block sizes
+(
+SynthDef(\delay, { |out, blockSize|
+    var source, local;
+
+	Reblock(blockSize);
+
+    source = Decay.ar(Impulse.ar(2), 0.1) * WhiteNoise.ar(0.2);
+    local = LocalIn.ar(2) + [source, 0]; // read feedback, add to source
+	local = DelayC.ar(local, 0.5, MouseX.kr.pow(2) * 0.5); // delay sound
+
+    // reverse channels to give ping pong effect, apply decay factor
+	LocalOut.ar(local.reverse * MouseY.kr.pow(0.5).clip(0, 0.999));
+
+	Out.ar(out, local);
+}).add
+)
+
+// by default, the min. delay length is the Server block size
+Synth(\delay);
+// try with smaller block sizes:
+Synth(\delay, [ blockSize: 16 ]);
+Synth(\delay, [ blockSize: 4 ]);
+Synth(\delay, [ blockSize: 1 ]);
+::
+
+CLASSMETHODS::
+
+METHOD:: new
+
+Specify the block size of the Synth. You should only create this once per SynthDef. Duplicate instances generate a warning and are ignored.
+
+ARGUMENT:: blockSize
+the block size. A non-zero block size must be a power of two. If the value is 0, the Server block size will be used.
+
+Currently, only downblocking is supported, i.e. the block size must be smaller than or equal to the Server block size. Otherwise you get a warning.
--- a/HelpSource/Classes/Resample.schelp
+++ b/HelpSource/Classes/Resample.schelp
@@ -53,7 +53,7 @@ Currently, only upsampling is supported! Any attempt at downsampling (e.g. a fac
 
 DISCUSSION::
 
-Note that the Resample object is not a UGen and there are certain restrictions on the types of arguments it accepts. In particular, you cannot use UGen expressions. (UGens are only evaluated after the Synth has been initialized. At this point, the resampling factor must have already been set.) For example, the following does not work and will throw an error:
+Note that the Resample object is strong::not:: a UGen and there are certain restrictions on the types of arguments it accepts. In particular, you cannot use UGen expressions. (UGens are only evaluated after the Synth has been initialized. At this point, the resampling factor must have already been set.) For example, the following does not work and will throw an error:
 code::
 // futile attempt at generating a random resampling factor:
 Resample(2 ** IRand(0, 4)); // -> throws an error
--- a/HelpSource/Classes/Resample.schelp
+++ b/HelpSource/Classes/Resample.schelp
@@ -1,0 +1,51 @@
+TITLE:: Resample
+summary:: Change the sample rate of a SynthDef/Synth
+categories:: Server>Abstractions
+related:: Classes/Reblock, Classes/SynthDef
+
+DESCRIPTION::
+For certain applications it is beneficial to run a link::Classes/Synth:: at a higher sample rate. Instead of increasing the sample rate of the whole Server, which could be very wasteful, you can use the Resample object inside a link::Classes/SynthDef:: to only change the sample rate of specific Synths.
+
+For example, with a factor of 4.0, the Synth will run at four times the Server's sample rate. Keep in mind that upsampling increases the CPU usage roughly proportionally to the upsampling factor.
+
+The resample factor can be a fixed number, or it can be set dynamically at runtime via a Synth argument/control.
+
+note::For technical reasons, Supernova only supports fixed resample factors.::
+
+For example, this can be used to implement anti-aliasing via oversampling + filtering:
+
+code::
+// reduce aliasing with upsampling + filtering
+(
+SynthDef(\hardsync, { |out, upsample=1.0|
+	var sig, cutoff = 18000;
+
+	Resample(upsample);
+
+	sig = SyncSaw.ar(MouseY.kr(1, 1000, 2), MouseX.kr(100, 4000, 2));
+
+	// crude anti-alisaing filter
+	sig = LPF.ar(sig, cutoff);
+	sig = LPF.ar(sig, cutoff);
+	Out.ar(out, sig * 0.05 ! 2);
+}).add;
+)
+
+// lots of aliasing
+Synth(\hardsync);
+// watch aliasing disappear with increasing upsample factors
+Synth(\hardsync, [ upsample: 2.0 ]);
+Synth(\hardsync, [ upsample: 4.0 ]);
+Synth(\hardsync, [ upsample: 8.0 ]);
+::
+
+CLASSMETHODS::
+
+METHOD:: new
+
+Specify the resample factor of the Synth. You should only create this once per SynthDef. Duplicate instances generate a warning and are ignored.
+
+ARGUMENT:: factor
+the resample factor, either as a constant number or as a Synth argument resp. link::Classes/NamedControl::. A value of 0.0 means no resampling (= a factor of 1.0).
+
+Currently, only upsampling is supported, i.e. the factor must be greater than or equal to 1.0 (= no resampling). Otherwise you get a warning.
--- a/HelpSource/Classes/Resample.schelp
+++ b/HelpSource/Classes/Resample.schelp
@@ -8,16 +8,15 @@ For certain applications it is beneficial to run a link::Classes/Synth:: at a hi
 
 For example, with a factor of 4.0, the Synth will run at four times the Server's sample rate. Keep in mind that upsampling increases the CPU usage roughly proportionally to the upsampling factor.
 
-The resample factor can be a fixed number, or it can be set dynamically at runtime via a Synth argument/control.
+The resampling factor can be a constant number or it can be set dynamically via a Synth control (except for Supernova). See link::#*new:: for more information.
 
-note::For technical reasons, Supernova only supports fixed resample factors.::
+note::For technical reasons, Supernova requires a fixed resampling factor. Trying to use a Synth control results in a warning.::
 
-For example, this can be used to implement anti-aliasing via oversampling + filtering:
+For example, you can use upsampling + filtering to reduce the effects of aliasing:
 
 code::
-// reduce aliasing with upsampling + filtering
 (
-SynthDef(\hardsync, { |out, upsample=1.0|
+SynthDef(\hardsync, { |out, upsample|
 	var sig, cutoff = 18000;
 
 	Resample(upsample);
@@ -31,9 +30,9 @@ SynthDef(\hardsync, { |out, upsample=1.0|
 }).add;
 )
 
-// lots of aliasing
+// no upsampling -> lots of aliasing
 Synth(\hardsync);
-// watch aliasing disappear with increasing upsample factors
+// watch aliasing disappear with increasing upsampling factors
 Synth(\hardsync, [ upsample: 2.0 ]);
 Synth(\hardsync, [ upsample: 4.0 ]);
 Synth(\hardsync, [ upsample: 8.0 ]);
@@ -43,9 +42,19 @@ CLASSMETHODS::
 
 METHOD:: new
 
-Specify the resample factor of the Synth. You should only create this once per SynthDef. Duplicate instances generate a warning and are ignored.
+Specify the resampling factor of the Synth. You should only create this once per SynthDef. Duplicate instances generate a warning and are ignored.
 
 ARGUMENT:: factor
-the resample factor, either as a constant number or as a Synth argument resp. link::Classes/NamedControl::. A value of 0.0 means no resampling (= a factor of 1.0).
+the resampling factor. This can either be a constant number (e.g. 2.0) or a Synth control. With the latter, the resampling factor can be set independently for each Synth. However, it cannot be modulated after the Synth has been instantiated on the Server.
 
-Currently, only upsampling is supported, i.e. the factor must be greater than or equal to 1.0 (= no resampling). Otherwise you get a warning.
+The resampling factor must be power of two. A value of 0.0 is the same as 1.0 (= no resampling).
+
+Currently, only upsampling is supported! Any attempt at downsampling (e.g. a factor of 0.5) is ignored and results in a warning.
+
+DISCUSSION::
+
+Note that the Resample object is not a UGen and there are certain restrictions on the types of arguments it accepts. In particular, you cannot use UGen expressions. (UGens are only evaluated after the Synth has been initialized. At this point, the resampling factor must have already been set.) For example, the following does not work and will throw an error:
+code::
+// futile attempt at generating a random resampling factor:
+Resample(2 ** IRand(0, 4)); // -> throws an error
+::
--- a/SCClassLibrary/Common/Audio/SynthDef.sc
+++ b/SCClassLibrary/Common/Audio/SynthDef.sc
@@ -9,6 +9,8 @@ SynthDef {
 	var <>constants;
 	var <>constantSet;
 	var <>maxLocalBufs;
+	var <reblock;
+	var <resample;
 
 	// topo sort
 	var <>available;
@@ -79,6 +81,8 @@ SynthDef {
 		controls = nil;
 		controlIndex = 0;
 		maxLocalBufs = nil;
+		reblock = nil;
+		resample = nil;
 	}
 	buildUgenGraph { arg func, rates, prependArgs;
 		var result;
@@ -405,6 +409,12 @@ SynthDef {
 			};
 
 			if (version > 2) {
+				// reblock
+				(this.reblock ?? { Reblock() }).writeDef(file);
+
+				// resample
+				(this.resample ?? { Resample() }).writeDef(file);
+
 				// calculate total size in bytes
 				endPos = file.pos;
 				byteSize = endPos - startPos;
@@ -449,6 +459,27 @@ SynthDef {
 		^true
 	}
 
+	reblock_ { arg obj;
+		if (obj.isKindOf(Reblock).not) {
+			Error("wrong argument to reblock_").throw;
+		};
+		if (reblock.isNil) {
+			reblock = obj;
+		} {
+			"ignore duplicate Reblock object".warn;
+		};
+	}
+
+	resample_ { arg obj;
+		if (obj.isKindOf(Resample).not) {
+			Error("wrong argument to resample_").throw;
+		};
+		if (resample.isNil) {
+			resample = obj;
+		} {
+			"ignore duplicate Resample object".warn;
+		};
+	}
 
 	// UGens do these
 	addUGen { arg ugen;
@@ -739,4 +770,95 @@ SynthDef {
 		^synth
 	}
 
+}
+
+Reblock {
+	var <>synthDef;
+	var <>blockSize;
+	var <>controlIndex;
+
+	*new { arg blockSize=0;
+		^super.new.init(blockSize);
+	}
+
+	init { arg blockSize;
+		synthDef = UGen.buildSynthDef;
+		if (blockSize.isKindOf(SimpleNumber)) {
+			blockSize = blockSize.asInteger;
+			// 0 means no reblocking
+			if ((blockSize != 0) and: { blockSize.isPowerOfTwo.not }) {
+				Error("block size % is not a power of two".format(blockSize)).throw;
+			};
+			this.blockSize = blockSize;
+		} {
+			if (blockSize.isKindOf(OutputProxy)) {
+				controlIndex = blockSize.controlIndex
+			};
+			if (controlIndex.isNil) {
+				Error("Reblock requires a number or Synth Control").throw;
+			}
+		};
+		if (synthDef.notNil) {
+			synthDef.reblock = this;
+		}
+	}
+
+	writeDef { arg stream;
+		if (controlIndex.notNil) {
+			// use control index
+			stream.putInt32(-1);
+			stream.putInt32(controlIndex);
+		} {
+			// constant block size
+			stream.putInt32(blockSize);
+			stream.putInt32(0);
+		}
+	}
+}
+
+Resample {
+	var <>synthDef;
+	var <>factor;
+	var <>controlIndex;
+
+	*new { arg factor=1.0;
+		^super.new.init(factor);
+	}
+
+	init { arg factor;
+		synthDef = UGen.buildSynthDef;
+		if (factor.isKindOf(SimpleNumber)) {
+			factor = factor.asFloat;
+			// 0.0 means no resampling
+			if (factor == 0.0) { factor = 1.0 };
+			// make sure that 'factor' is a power of two (with positive or negative exponent)
+			if ((factor.asInteger.isPowerOfTwo.not) and:
+				{ factor.reciprocal.asInteger.isPowerOfTwo.not }) {
+				Error("resample factor % is not a power of two".format(factor)).throw;
+			};
+			this.factor = factor;
+		} {
+			if (factor.isKindOf(OutputProxy)) {
+				controlIndex = factor.controlIndex
+			};
+			if (controlIndex.isNil) {
+				Error("Resample requires a number or Synth Control").throw;
+			}
+		};
+		if (synthDef.notNil) {
+			synthDef.resample = this;
+		}
+	}
+
+	writeDef { arg stream;
+		if (controlIndex.notNil) {
+			// control index
+			stream.putFloat(-1.0);
+			stream.putInt32(controlIndex);
+		} {
+			// constant factor
+			stream.putFloat(factor);
+			stream.putInt32(0);
+		}
+	}
 }
--- a/include/server/ErrorMessage.hpp
+++ b/include/server/ErrorMessage.hpp
@@ -71,7 +71,9 @@ std::string apiVersionMismatch(std::string const& utf8Filename, int const expect
     assert(expectedVersion != actualVersion);
 
     // both 1 and 2 were introduced in 3.6
-    static map<int, string> const scVersionForAPIVersion = { { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" } };
+    static map<int, string> const scVersionForAPIVersion = {
+        { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" }, { 4, "3.14" }
+    };
 
     stringstream message;
     message << "ERROR: API version mismatch: " << utf8Filename << "\n";
--- a/include/server/ErrorMessage.hpp
+++ b/include/server/ErrorMessage.hpp
@@ -72,7 +72,7 @@ std::string apiVersionMismatch(std::string const& utf8Filename, int const expect
 
     // both 1 and 2 were introduced in 3.6
     static map<int, string> const scVersionForAPIVersion = {
-        { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" }, { 4, "3.14" }
+        { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" }, { 4, "3.15" }
     };
 
     stringstream message;
--- a/server/scsynth/SC_MiscCmds.cpp
+++ b/server/scsynth/SC_MiscCmds.cpp
@@ -746,9 +746,8 @@ SCErr meth_d_free(World* inWorld, int inSize, char* inData, ReplyAddress* inRepl
     return kSCErr_None;
 }
 
-
-SCErr meth_s_new(World* inWorld, int inSize, char* inData, ReplyAddress* inReply);
-SCErr meth_s_new(World* inWorld, int inSize, char* inData, ReplyAddress* /*inReply*/) {
+// 'argtype' is false for setn type args
+SCErr meth_s_do_new(World* inWorld, int inSize, char* inData, bool argtype) {
     SCErr err;
     sc_msg_iter msg(inSize, inData);
     int32* defname = msg.gets4();
@@ -770,7 +769,7 @@ SCErr meth_s_new(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRep
         Group* group = Msg_GetGroup(inWorld, msg);
         if (!group)
             return kSCErr_GroupNotFound;
-        err = Graph_New(inWorld, def, nodeID, &msg, &graph, true); // true for normal args
+        err = Graph_New(inWorld, def, nodeID, &msg, &graph, argtype);
         if (err)
             return err;
         if (!graph)
@@ -781,7 +780,7 @@ SCErr meth_s_new(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRep
         Group* group = Msg_GetGroup(inWorld, msg);
         if (!group)
             return kSCErr_GroupNotFound;
-        err = Graph_New(inWorld, def, nodeID, &msg, &graph, true);
+        err = Graph_New(inWorld, def, nodeID, &msg, &graph, argtype);
         if (err)
             return err;
         Group_AddTail(group, &graph->mNode);
@@ -790,7 +789,7 @@ SCErr meth_s_new(World* inWorld, int inSize, char* inData, ReplyAddress* /*inRep
         Node* beforeThisNode = Msg_GetNode(inWorld, msg);
         if (!beforeThisNode)
             return kSCErr_NodeNotFound;
-        err = Graph_New(inWorld, def, nodeID, &msg, &graph, true);
+        err = Graph_New(inWorld, def, nodeID, &msg, &graph, argtype);
         if (err)
             return err;
         Node_AddBefore(&graph->mNode, beforeThisNode);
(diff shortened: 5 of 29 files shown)


SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -740,3 +798,60 @@ SynthDef {
...
+
+	*new { arg blockSize=0;
+		^super.new.init(blockSize);
+	}
+
+	init { arg blockSize;
+		var synthDef = UGen.buildSynthDef;
+		case { blockSize.isKindOf(SimpleNumber) } {
+			blockSize = blockSize.asInteger;
+			if (blockSize < 0) {
```
telephon: this test isn't needed, because negative numbers are not powers of two.

SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -740,3 +798,60 @@ SynthDef {
...
+		^super.new.init(factor);
+	}
+
+	init { arg factor;
+		var synthDef = UGen.buildSynthDef;
+		case { factor.isKindOf(SimpleNumber) } {
+			factor = factor.asFloat;
+			// make sure that 'factor' is a power of two (with positive or negative exponent)
+			if ((factor <= 0.0)
+				or: { (factor >= 1.0) and: { factor.asInteger.isPowerOfTwo.not } }
```
telephon: as far as I can tell, these disjunctions can be expressed in a single predicate: `factor.asInteger.abs.isPowerOfTwo.not`

SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -740,3 +798,60 @@ SynthDef {
...
+Resample {
+	var <>factor;
+	var <>controlIndex;
+
+	*new { arg factor=1.0;
+		^super.new.init(factor);
+	}
+
+	init { arg factor;
+		var synthDef = UGen.buildSynthDef;
```
telephon: There should be a test that warns if you are not in a SynthDef, and therefore `UGen.buildSynthDef.isNil`.

SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -740,3 +798,60 @@ SynthDef {
...
+			factor = factor.asFloat;
+			// make sure that 'factor' is a power of two (with positive or negative exponent)
+			if ((factor <= 0.0)
+				or: { (factor >= 1.0) and: { factor.asInteger.isPowerOfTwo.not } }
+				or: { (factor < 1.0) and: { factor.reciprocal.asInteger.isPowerOfTwo.not } }) {
+				Error("resample factor must be a power of two").throw;
+			};
+			this.factor = factor;
+		} { factor.isKindOf(OutputProxy) and: { factor.source.isKindOf(Control) }} {
+			this.controlIndex = factor.source.specialIndex + factor.outputIndex;
```
telephon: would be good as a method of OutputProxy:

```supercollider
OutputProxy {
	
	controlIndex {
		if(source.isKindOf(Control).not) { ^nil };
		^source.specialIndex + outputIndex
	}
	
}
```

telephon (commented):
a few suggestions for simplification

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: Great idea! Actually, I think we could also use `controlIndex` in the `controlName` method!

Currently, `controlName` uses the following code to compute the control index:
```sc
var counter = 0, index = 0;
this.synthDef.children.do { |ugen|
    if (this.source.synthIndex == ugen.synthIndex) { index = counter + this.outputIndex; };
	if (ugen.isKindOf(Control)) { counter = counter + ugen.channels.size; };
};
^synthDef.controlNames.detect { |c| c.index == index };
 ```
This is a linear scan over *all* UGens in the current SynthDef, which is pretty wasteful. I *think* this can be replaced with a call to the `controlIndex` method:
```sc
var index = this.controlIndex;
^synthDef.controlNames.detect { |c| c.index == index };
```
AFAICT, the control index is always `source.specialIndex + outputIndex` (if the source is a kind of `Control`).

I've also noticed that `controlName` does not really check whether the source itself is really a kind of `Control`.

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: I don't think so. The second predicate (`{ (factor < 1.0) and: { factor.reciprocal.asInteger.isPowerOfTwo.not } }`) actually checks for power of two with negative exponent (e.g. 0.5 or 0.25).

However, I can remove the `factor <= 0.0` check. (Negative values will not be power of twos and I actually want to allow 0.0 as a default value, similar to `Reblock`.)

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: I do not necessarily disagree. However, UGens currently do not warn about this. Here's the code for `UGen-addToSynth`:
```sc
addToSynth {
    synthDef = buildSynthDef;
    if (synthDef.notNil, { synthDef.addUGen(this) });
}
```
If we decide to print a warning, I think we should do it consistently.

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: In theory, the check can be reduced to the following:
```
if (factor.asInteger.isPowerOfTwo.not and: { factor.reciprocal.asInteger.isPowerOfTwo.not })
```
The problem is that there's currently a bug in `isPowerOfTwo` where it wrongly returns `true` for 0. See https://github.com/supercollider/supercollider/issues/7370

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: I just went ahead and added a warning. Later we might decide to also post a similar warning in `UGen-addToSynth`.

JordanHendersonMusic: I find it really useful to build ugen graphs outside the synthdef, as you can query the channel count with shape, without building the graph

Spacechild1: That's interesting! I guess we can discuss this briefly in the dev meeting :) (Will be there in a moment)

dyfer: I also query ugen graphs for number of channels etc so I don't think we should post a warning just because the graph is outside of the synthdef.

Spacechild1:
TODO: bump the plugin API temporarily to avoid crashing existing third-party plugins on the develop branch.

SCClassLibrary/Common/Audio/SynthDef.sc
telephon: yes, please do take the occasion to make a good interface. 

It is usually when we need to do something we realise that an interface is needed, and then is also a good moment, because it is used immediately.

Spacechild1:
TODO: throw an Error on duplicate `Reblock` / `Resample` objects + add tests

capital-G:
Can you add some tests? Also, the test should cover when specifying `Reblock` and `Resample` multiple times? Or e.g. having `Reblock` in loop like `8.collect{|i| Resample(2.0); SinOsc.ar(i)}` or multi-channel expansion?

IMO instead of error we should maybe move towards a warning?

Spacechild1:
~~Ok. let's make it a warning. Only the first `Reblock` / `Resample` object is considered, duplicate objecs post a warning. I will add a note to the Help files.~~

Turns out I had already implemented the warning, it was just in another method :-D Now I've also documented the behavior in the Help files.

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: Removed the warning.

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: I tried to call `controlName` inside the ugen graph func and print the output. That's the only place where we actual have access to the OutputProxies. For all I can see, the `controlIndex` method really does the same thing as the loop, so I have replace the latter with the former.

telephon (commented):
Just a thought: perhaps it would be clearer and more explicit if the SynthDef had the four variables of `Reblock` and `Resample` as (private) instance variables.

Then, 

- either the logic for setting them goes into SynthDef instance methods (and the Reblock/Resample class just calls this on `UGen.buildSynthDef`).
- or, the logic remains in the Reblock/Resample class and it sets the values in SynthDef.


Or, the SynthDef could ask the Reblock/Resample class to write themselves to the file which is passed to it by the SynthDef as it writes it.

I am not convinced that this is better, but I wanted to throw it in anyway.

Spacechild1:
> Or, the SynthDef could ask the Reblock/Resample class to write themselves to the file which is passed to it by the SynthDef as it writes it.

Thanks! I've now added `writeDef` methods to `Reblock` and `Resample` which is responsible for writing the SynthDef fields. This is much cleaner now, I think. I wanted to do this initially, but I didn't know how to elegantly handle the case of `resample` or `reblock` being `nil`. Only now I realized I could just use default-constructed temporary instances :)

HelpSource/Classes/Resample.schelp
```diff
@@ -0,0 +1,51 @@
...
+summary:: Change the sample rate of a SynthDef/Synth
+categories:: Server>Abstractions
+related:: Classes/Reblock, Classes/SynthDef
+
+DESCRIPTION::
+For certain applications it is beneficial to run a link::Classes/Synth:: at a higher sample rate. Instead of increasing the sample rate of the whole Server, which could be very wasteful, you can use the Resample object inside a link::Classes/SynthDef:: to only change the sample rate of specific Synths.
+
+For example, with a factor of 4.0, the Synth will run at four times the Server's sample rate. Keep in mind that upsampling increases the CPU usage roughly proportionally to the upsampling factor.
+
+The resample factor can be a fixed number, or it can be set dynamically at runtime via a Synth argument/control.
```
dyfer: Do I understand correctly that "fixed number" refers to SynthDef creation (`SynthDef().add`) and "dynamically at runtime" to `Synth()` called with an argument? And also that it's not modulatable? 
We can improve the docs later, I don't want to burden Christof much beyond the functionality of the PR, but I wanted to leave this comment while I noticed it.

If my understanding is correct, then I'd rephrase this specifically like:
```suggestion
The resample factor can be set statically when creating the SynthDef (code::SynthDef().add::), or dynamically per instance when playing the Synth (code::Synth()::). However, it cannot be modulated once the synth starts playing.
```

HelpSource/Classes/Reblock.schelp
```diff
@@ -0,0 +1,51 @@
+TITLE:: Reblock
+summary:: Change the block size of a SynthDef/Synth
+categories:: Server>Abstractions
+related:: Classes/Resample, Classes/SynthDef
+
+DESCRIPTION::
+For certain algorithms it is necessary to run a link::Classes/Synth:: at a small block size. Instead of decreasing the block size of the whole Server, which could be rather wasteful, you can use the Reblock object inside a link::Classes/SynthDef:: to only change the block size of specific Synths. (Keep in mind that very small block sizes will increase the CPU usage.)
+
+The block size can be a fixed number, or it can be set dynamically at runtime via a Synth argument/control.
```
dyfer: Similarly to resampling, I wonder if this is cleaner (if my understanding is correct...):
```suggestion
The block size can be set statically when creating the SynthDef (e.g. code::SynthDef().add::), or dynamically per instance when playing the Synth (code::Synth()::), except in Supernova (see below). However, it cannot be modulated once the synth starts playing.
```

HelpSource/Classes/Reblock.schelp
```diff
@@ -0,0 +1,51 @@
...
+
+note::For technical reasons, Supernova only supports fixed block sizes.::
```
dyfer: Along the lines of my other comments
```suggestion
note::For technical reasons, Supernova only supports fixed block sizes set as constants in a code::SynthDef::.::
```

dyfer (commented):
I have some documentation suggestions - not a priority, we can do this later.

Reblock and Resample are _not_ modulatable once the Synth starts, correct? (Supernova limitations notwithstanding)

server/scsynth/SC_MiscCmds.cpp
```diff
@@ -746,9 +746,8 @@ SCErr meth_d_free(World* inWorld, int inSize, char* inData, ReplyAddress* inRepl
     return kSCErr_None;
 }
 
-
-SCErr meth_s_new(World* inWorld, int inSize, char* inData, ReplyAddress* inReply);
-SCErr meth_s_new(World* inWorld, int inSize, char* inData, ReplyAddress* /*inReply*/) {
+// 'argtype' is false for setn type args
+SCErr meth_s_do_new(World* inWorld, int inSize, char* inData, bool argtype) {
```
dyfer, line 750: Are changes in `SC_MiscCmds.cpp` related to reblocking/resampling directly, or a cleanup, or some other bugfix? Sorry I don't know this code well.

HelpSource/Classes/Reblock.schelp
Spacechild1: What about the following:

> For technical reasons, Supernova requires a fixed block size for every SynthDef, i.e. the `blockSize` argument must be a constant number.

HelpSource/Classes/Resample.schelp
Spacechild1: Yeah, the wording is not entirely clear. I also left out some important information. 

> However, it cannot be modulated once the synth starts playing.

Thanks, that's definitely worth pointing out! However, it's really "once the Synth has been created", not "once the Synth starts playing". That's a subtle difference. (A Synth can be created in a paused state, in which case it may only start playing long after it has been created.)

Here's another shot:

> The resample factor can either be a constant number (e.g. 2.0) or it can be a Synth argument / NamedControl. With the latter, the resample factor can be set independently for each Synth. However, it cannot be modulated after the Synth has been created.
>
> Note that you cannot use any expressions on Synth controls. (These are implemented as UGens and as such are only evaluated after the Synth has been initialized.) For example, the following does not work and will throw an error:
>```sc
>var upsample = \upsample.ir(0.0);
>// cannot do any expressions on the `upsample` control:
>Resample(if (upsample != 0.0, 4.0, 1.0)); // -> throws an error
>// instead the control must be passed directly to Resample:
>Resample(upsample);
>```

server/scsynth/SC_MiscCmds.cpp
Spacechild1: Ah, that is a left over form the old PR where the resample factor and block size was part of a new `/s_newEx` command. I've added `meth_s_do_new` to avoid code duplication. I've also used this to share code between `/s_new` and `/s_newargs`. Apparently, I have kept the latter part for the new PR.

In this context, it is really an unrelated cleanup. Should I remove it?

dyfer: I think it would be best to remove it from this PR, if that's not too much trouble - or at least have it as a separate commit, clearly labelled.

Spacechild1: Moved it to a separate commit.

HelpSource/Classes/Resample.schelp
telephon: This example is misleading:
1. in a ugen graph, the expression `upsample != 0.0` will always return false, where `upsample` is a UGen.
2. You should never be obliged to pass a control rather than a UGen, if the value is just sampled at initialisation. You should be able to write `Resample((2 ** IRand(0, BlockSize.ir.log2)))`, just as you write `Resample(upsample)`.

There is no reason to limit this to a control.
Don't you think?

dyfer: This is better!
Building on this, I have a thought to remove the ambiguity due to the possibility of creating a Synth representation in the language only:

```
However, it cannot be modulated after the Synth has been instantiated on the server.
```

dyfer: Also, if the rate (`.ir`) is not specified, would the control still work?
```
SynthDef(\name, {arg upsample = 1;
    Resample(upsample);
});
```
~~And does upsample of 0.0 make sense (re: your previous example)? Isn't that a multiplier?~~ I guess your previous example is just to show an invalid attempt to avoid passing a 0 as resampling factor, right?
Still, what happens if resampling is accidentally set to 0?

HelpSource/Classes/Reblock.schelp
dyfer: Yes, I like it, it's more concise.

HelpSource/Classes/Resample.schelp
Spacechild1: >  in a ugen graph, the expression upsample != 0.0 will always return false, where upsample is a UGen.

Sorry, I meant to write `upsample |!=0| 0.0`. But that's actually redundant since the `if` method considers any non-zero output as truthy. So it should have really been:
```sc
// cannot do any expressions on the `upsample` control:
Resample(if (upsample, 4.0, 1.0)); // -> throws an error
```

> You should never be obliged to pass a control rather than a UGen, if the value is just sampled at initialisation. 

In the case of `Reblock` and `Resample` you are. Maybe I should make it more clear that this differs from the behavior of "normal" UGens.

> There is no reason to limit this to a control.

I gave the reason in the original post: "These are implemented as UGens and as such are only evaluated after the Synth has been initialized."

I guess this was too minimal. What about the following:

"These are implemented as UGens and as such are only evaluated after the Synth has been initialized. At this point, the resampling factor must have already been set."

The point is that the UGens are only initialized in the first call to the calc function -- *after* the Synth has been created! This includes any init-rate expressions.

> You should be able to write Resample((2 ** IRand(0, BlockSize.ir.log2))), just as you write Resample(upsample).

No, this wouldn't work. You can't set the resampling factor while calculating the UGen graph. The resampling factor (and block size) must be set *before* the UGens are calculated. That's why it must be either a constant number or a control index.

Spacechild1: > Also, if the rate (.ir) is not specified, would the control still work?

Sure! Synth arguments are explicitly mentioned. You can use `|resample|`, `\resample.kr` or `\resample.ir`, they all just create instances of `Control`. The calc rate is only relevant to the control *UGen*, i.e. the UGen that outputs the control value. The actual Synth controls are just a single flat array of numbers that have no associated rate. 

> I guess your previous example is just to show an invalid attempt to avoid passing a 0 as resampling factor, right?

Not really :-D I meant to write the following:
```sc
Resample(if (upsample, 4.0, 1.0)); // -> throws an error
``` 
`upsample` would a quasi-boolean Synth argument. If "true" (!= 0.0), the resampling factor should be 4.0, otherwise it should be 1.0 (= no resampling). This doesn't work because the Synth control is not directly passed as an argument to `Resample`.

> Still, what happens if resampling is accidentally set to 0?

This is mentioned in the docs: 0.0 is treated the same as 1.0 (= no resampling).

HelpSource/Classes/Reblock.schelp
dyfer: How about: `(...) must be a constant number; i.e. it may not be a synth control.`

HelpSource/Classes/Resample.schelp
dyfer: > This is mentioned in the docs: 0.0 is treated the same as 1.0 (= no resampling).

Hm, don't want to split hairs, but that seems both convenient and arbitrary :) I guess it's OK if we don't currently allow Resample factors `< 1.0`, but logically I'd expect a factor of 0.0 to produce silence and warn  instead of reverting to 1.0. OR... process and output only first sample of the synth?? That would be I think accurate, but impractical...

Also I just thought that it could be useful to allow undersampling when one needs to process something at a rate slightly higher than control rate but lower than audio rate, but that's something to consider later, _maybe_. That would be a very narrow use case anyway (like bio-signals, for example; BUT control rate is typically fine for those).

Thanks for the other explanations.

> "These are implemented as UGens and as such are only evaluated after the Synth has been initialized. At this point, the resampling factor must have already been set."

I like this! It gives a good description for those with some familiarity with SC.

Spacechild1: > but logically I'd expect a factor of 0.0 to produce silence and warn instead of reverting to 1.0. 

A resample factor of 0.0 is impossible, so we can use it as a reserved value, similar to how a block size of 0 means "use the Server block size". Now, it is indeed a bit redundant as a factor of 1.0 already means "no resampling". I did this for two reasons:
1. If a SynthDef uses no resampling or reblocking, you can just zero all four fields
2. You don't have to provide a default value for the SynthDef argument / NamedControl:
```sc
// 'resample' has an implicit default value of 0.0.
// Instead of making this an error, we interpret it as 1.0 (= no resampling), just like a block size of 0 means "no reblocking"
SynthDef(\test, { |out, resample, blockSize|
    Resample(resample);
    Reblock(blockSize);
    Out.ar(out, SinOsc.ar(440));
}).add;
```

> OR... process and output only first sample of the synth?? That would be I think accurate, but impractical...

Yes, theoretically one could interpret a factor of 0.0 as downsampling by infinity (2**(-inf) -> 0.0). There's only one slight issue: the number of samples and the size of the audio buffers for the first (and only) process tick would have to be infinite :-D

telephon: > No, this wouldn't work. You can't set the resampling factor while calculating the UGen graph. The resampling factor (and block size) must be set _before_ the UGens are calculated. That's why it must be either a constant number or a control index.

I see, yes, this makes sense. Interesting boundary case, so they come between the controls and the ugens? How about audio rate controls?

dyfer: > There's only one slight issue: the number of samples and the size of the audio buffers for the first (and only) process tick would have to be infinite :-D

Just a small detail haha

Okay, the default value reasoning makes sense.

Spacechild1: > I see, yes, this makes sense. Interesting boundary case, so they come between the controls and the ugens? 

I think you could see it that way.

> How about audio rate controls?

All rates are supported. The only thing that does not work is a NamedControl with lag values (unless control-rate and fixed) because this wraps the actual Control UGen in a Lag UGen.

On the Server side, the rate does not matter. In fact, Synth controls themselves do not have rates, only the mappings do. The mapping rate is only relevant to the `AudioControl` UGen.

telephon: Hm, ok, this is a very delicate case, it will be important to make clear that `Resamble` and `Reblock` are not UGens and do not follow the logic.

If we wanted to make this really clear, we would make it an interface purely of the SynthDef, like:
```
UGen.buildSynthDef.reblock(8);
UGen.buildSynthDef.resample(4);
```

I do see that this is not beautiful code.

What about making a class that returns the SynthDef directly?

```
ThisSynthDef {
    *new { ^UGen.buildSynthDef }
}
```

Then we could write:

```
ThisSynthDef.reblock(8);
ThisSynthDef.resample(4);
```
(alternatively `BuildSynthDef`)

And it would also be useful in other cases where you need information from the build synth def.

Just a thought! I would be also fine with the classes, but there is a slight issue with the similarity to UGens.

Spacechild1: > Then we could write:
> ```sc
>ThisSynthDef.reblock(8);
> ThisSynthDef.resample(4);
>```

Wouldn't you have to actually create an instance?
```sc
ThisSynthDef().reblock(8);
ThisSynthDef().resample(4);
```

I would rather use a class method of `SynthDef`:
```sc
SynthDef.thisDef.reblock(8);
SynthDef.thisDef.resample(4);
```

However, this doesn't really change anything about the restrictions of the argument type. It only makes the code longer...

People might still try to do things like this:
```sc
SynthDef.thisDef.reblock(2 ** IRand(0, BlockSize.ir.log2);
```

I think we only have to make it clear in the docs that `Resample` and `Reblock` are not UGens. Personally, I wouldn't worry too much. I think most people will only ever use constant numbers or Synth arguments / NamedControls.

muellmusik: > What about making a class that returns the SynthDef directly?
> 
> ```
> ThisSynthDef {
>     *new { ^UGen.buildSynthDef }
> }
> ```

No strong views on this idea, but this seems like a special case style-wise? A class method of SynthDef seems more consistent (```SynthDef:*buildSynthDef``` or whatever). Or even a pseudo-variable? ```thisSynthDef```, though that's probably overkill? None of these seem ideal admittedly.

Spacechild1: One nice thing about a `SynthDef-reblock` and `SynthDef-blockSize` method is that you can also just call them on the resulting SynthDef (but only for constant factors / block sizes):
```sc
SynthDef(\test, { |out|
    Out.ar(out, SinOsc.ar(440));
}).reblock(1).add;
```
However, setting the argument via a Synth control still seems rather awkward in comparison:
```sc
SynthDef(\test, { |out, blockSize|
    SynthDef.thisDef.reblock(blockSize);
    Out.ar(out, SinOsc.ar(440));
}).add;
```

I still prefer the `Reblock` and `Resample` objects.

telephon: So what about:
```supercollider
SynthDef(\test, { |out, blockSize|
    thisSynthDef.reblock(blockSize);
    Out.ar(out, SinOsc.ar(440));
}).add;
```

Spacechild1: This would require a change in the language. I don't think that we should add a new global pseudo-variable just for this purpose. And again, the restrictions on the argument type still apply and would have to be documented. I don't really see the advantage.

Spacechild1: I could image having `reblock` and `resample` as class methods (that act on the currently building SynthDef):
```sc
SynthDef(\test, { |out, blockSize|
    SynthDef.reblock(blockSize);
    Out.ar(out, SinOsc.ar(440));
}).add;
```

It still doesn't magically solve the problem of the argument type restrictions. Therefore I would still opt for the less verbose solution of having dedicated objects.

One advantage of the `Reblock` and `Resample` objects is that the documentation would be in dedicated files. The help file for `SynthDef` is already pretty large...

telephon: > It still doesn't magically solve the problem of the argument type restrictions.

I don't think that this is a problem at all if the interface says what it does. The separate classes look good and I aesthetically prefer them. In the end, they are just wrappers for methods of `SynthDef`.

I think the closest comparison is `NamedControl` and `LocalIn` or `LocalBuf`, also the `InfoUGen` subclasses are somewhat related.

If you give a good description why they can't take ugen graphs as argument, but why they can take pure controls as argument, it should be fine.

Ah one other possibility: allow numbers and symbols. If passed a symbol, it refers to the respective control name.

Spacechild1: > If you give a good description why they can't take ugen graphs as argument, but why they can take pure controls as argument, it should be fine.

Yes, that's what I think.

> Ah one other possibility: allow numbers and symbols. If passed a symbol, it refers to the respective control name.

I actually thought about that. One the plus side, it would remove the ambiguity because you can't pass *any* UGen. However, I found it too awkward in comparison, especially with NamedControls:
```sc
SynthDef(\test, {
    \resample.kr;
    Resample(\resample);
    // ...
}).add;
```
VS
```sc
SynthDef(\test, {
    Resample(\resample.kr);
    // ...
}).add;
```
Passing a symbol to `Resample` *could* automatically generate a NamedControl if it doesn't exist yet, but that's too much magic for my taste. Also, with the Symbol argument it is not immediately clear what it does whereas passing a Synth control directly is pretty much self-explanatory (at least to me :)

Spacechild1: > I think the closest comparison is NamedControl and LocalIn or LocalBuf, also the InfoUGen subclasses are somewhat related.

They all create actual UGens. I don't think there's any prior art here. It's definitely worth discussing the interface because we might get similar features in the future.

I've already floated an idea for passing Symbols as UGen arguments (see https://github.com/supercollider/supercollider/issues/7403#issuecomment-4022975253 and below). In that case, there could be a `DefSym` object that lets you add additional Symbols to the SynthDef's symbol table. (This would allow setting "symbol" arguments via Synth controls.) This follows the design of `Resample` and `Reblock`. Alternatively, it could also be implemented as a class method: `SynthDef-defSym`.

We should keep the bigger picture in mind. What kind of interface do want for specifying *SynthDef attributes*?

JordanHendersonMusic: > This would require a change in the language. I don't think that we should add a new global pseudo-variable just for this purpose.

You can just do this (you have to update the C++ as well, but its really minor).
```supercollider
Object {
   classvar thisSynthDef;
}
```

*This is not an endorsement of this, just wanted to say its trivial if really needed.*

---

I like this, literally can't be a control if it isn't inside the function.

```supercollider
SynthDef(\test, Re(sample: 4, block: 1) {
   Out.ar(...)
}).add
```
I floated this earlier but no one mentioned it.

---


Or another option might be to use the default value of the argument and just not have a UGen at all, i.e., having a control named `resample` or `blockSize` simply changes the value.

It is a bit magic, but since we want `ir` rules, isn't too suprising.

```supercollider
SynthDef(\test, { 
   \resample.ir(4) // has a side effect of changing the sample rate.
...
}).add
```

Spacechild1: > I like this, literally can't be a control if it isn't inside the function.

People *should* be able to use controls. The point is just that controls must be passed directly, i.e. you can't do any math on them (because this would result in a UGen expression). I still think this is an edge case that can be avoided with good documentation.

telephon: In my view, the conclusion of all this is:
- the main interface is an interface of a SynthDef, and it can be applied after the SynthDef has been instantiated (therefore messages to the instance)
- we want to have a simple way to access this interface from within a SynthDef's function that can be externally controlled.

For a simple way to access this interface, there are these possibilities:
1. simplify the access to the SynthDef, e.g. by `thisSynthDef`. 
       **Pro**: understandable, useful also for other things (Question: are there many other introspective tasks that could benefit?), **Con**: no autocomplete, cluttered documentation of SynthDef
2. make classes whose `new` method sets the values in `UGen.buildSynthDef` when called, return nil.
       **Pro**: easy to remember and looks good, **Con**: looks like a UGen, potentially confusing, because it is something completely different, namely sugar for a setter.
3. make a control name that does this automatically when it exists.
      **Pro**: easy to remember and looks good, **Con**: no autocomplete, and somewhat obscure in functionality


I am still undecided, hope this helps toward a good decision! Any opinons?

Spacechild1: Thanks for the summary!

> make classes whose new method sets the values in UGen.buildSynthDef when called, return nil.

Just a slight correction: the `Reblock` and `Resample` constructors do not return `nil`, they create a valid instance. The `init` method then passes that instance to the SynthDef (if `UGen.buildSynthDef` is not `nil`). In theory, the constructors *could* just set some members in the SynthDef, but creating actual objects helps with the logic and allows to encapsulate the serialization code.

> make a control name that does this automatically when it exists.

Can you tell more about this? This hasn't come up so far and I can't quite imagine what it would look like.

telephon: But I see that as the implementation is now, it is added to the SynthDef and kept there as a place to keep the information and functionality. In the ugen function itself, it is useless and should not be used.

I meant actually that `Reblock.new` **should** return `nil` since it is just there to pass a message to the SynthDef conveniently - it can still be present in the SynthDef. Just a thought to minimise wrong use.

>Can you tell more about this? This hasn't come up so far and I can't quite imagine what it would look like.

I was referring to @JordanHendersonMusic 's idea above:

> Or another option might be to use the default value of the argument and just not have a UGen at all, i.e., having a control named `resample` or `blockSize` simply changes the value.

Spacechild1: > I meant actually that Reblock.new should return nil since it is just there to pass a message to the SynthDef conveniently - it can still be present in the SynthDef.

This is not really practical because then I can't create default-constructed instances in the `SynthDef-writeDef` method

> Just a thought to minimise wrong use.

I don't see the issue. `Reblock` and `Resample` are not valid UGen inputs (`Object -isValidUGenInput` returns `false`). If you try to use them as UGen inputs, you get a clear error message:
```sc
(
SynthDef(\boo, { |out, blockSize|
	Out.kr(out, Reblock(blockSize));
}).add;
)
```
->
```
Out arg: 'channelsArray' has bad input: a Reblock
 ARGS:
   bus: an OutputProxy OutputProxy
   channelsArray: a Reblock Reblock
SynthDef boo build failed
ERROR: Out arg: 'channelsArray' has bad input: a Reblock
```

I don't see how returning `nil` would improve things. On the contrary: now the error message would complain about a `nil` input and the user would be wondering where that comes from. Even worse: `nil` is convertible to Array so this can lead to even more cryptic error messages.

Not to mention that returning `nil` from a constructor is just weird.

> I was referring to @JordanHendersonMusic 's idea above:

Ah sorry, I think I overlooked this. That's too much magic for my taste. Reserving Synth argument names for special purposes feels pretty hacky to me. Also, it's not discoverable at all. How would the user find the documentation? IMO this is not a reasonable option.

---

At this point, I think it boils down to only two options:
1. `SynthDef` methods (e.g. `SynthDef -reblock` and `SynthDef -resample`) + a global pseudo-variable `thisSynthDef`
  **Pro**: understandable, useful also for other things (Question: are there many other introspective tasks that could benefit?),
  **Con**: no autocomplete, cluttered documentation of SynthDef
2. `Reblock` and `Resample` classes that register themselves in the currently building SynthDef (`UGen.buildSynthDef`)
  **Pro**: easy to remember and looks good. Documentation can live in dedicated help files. The SynthDef serialization logic is encapsulated in the objects, i.e. they know how to write themselves (`-writeDef`)
  **Con**: looks like a UGen, potentially confusing, because it is something completely different.
  
Regarding the argument that `Reblock` and `Resample` might look like UGens: What is the actual concern here? What are the subtle mistakes that the user could make (that wouldn't be possible with the other option)?

(BTW, the documentation now explicitly mentions that these classes are not UGens.)

dyfer: Even though a bit inconsistent with existing practices (similarity to a UGen), I'd vote for `Resamble` and `Reblock` classes. Clear error handling makes it a good choice IMO.

Spacechild1: To be fair, the error handling would be very similar in both options. For me the deciding advantages of the `Reblock` and `Resample` objects over the `SynthDef` methods are (in descending order of importance):
1. documentation (with examples) can be in dedicated files
2. less typing and aesthetically more pleasing (very subjective, I know :)
3. the serialization code is encapsulated in the `writeDef` methods

I just realized that probably one reason why I don't find the `Reblock` and `Resample` objects so unusual is that it reminds me of Pd's `[block~]` object. This object looks like a UGen (in Pd terms: signal object), but it doesn't do any computation. It's only purpose is to specify the block size, resampling factor and overlap.

Spacechild1:
Ok, I have now added Unit tests for every single IO UGen. *All tests pass!*

I basically compare the output of two Synths of the same SynthDef where one Synth is resampled resp. reblocked and the other is not. This makes sure that resampling or reblocking does not alter the behavior of any IO UGen. The tests actually helped to catch a few remaining issues.

From my side the PR would be ready. There are still some discussions about the client-side interface (see https://github.com/supercollider/supercollider/pull/7402#discussion_r2951939037) but at the moment I'm sensing a tendency towards the current implementation (`Resample` and `Reblock` classes).

So far I haven't heard any concerns regarding the Server-side implementation. If we want to change the client-side interface, we can still do that before we release SC 3.15. It would be mostly a matter of copy-and-paste. Therefore I would suggest to merge this PR rather sooner than later so that more people will actually try it out in practice.

But first this PR needs a (positive) review :)

include/server/ErrorMessage.hpp
```diff
@@ -71,7 +71,9 @@ std::string apiVersionMismatch(std::string const& utf8Filename, int const expect
     assert(expectedVersion != actualVersion);
 
     // both 1 and 2 were introduced in 3.6
-    static map<int, string> const scVersionForAPIVersion = { { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" } };
+    static map<int, string> const scVersionForAPIVersion = {
+        { 1, "3.6.0" }, { 2, "3.6.0" }, { 3, "3.9.0" }, { 4, "3.14" }
```
JordanHendersonMusic, line 75: Should this be 3.15?

SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -740,3 +771,94 @@ SynthDef {
 	}
 
 }
+
+Reblock {
+	var <>synthDef;
+	var <>blockSize;
+	var <>controlIndex;
+
+	*new { arg blockSize=0;
```
JordanHendersonMusic, line 780: Given than blockSize cannot be nil (it throws in init), is it better to do `blockSize (0)`?

SCClassLibrary/Common/Audio/SynthDef.sc
```diff
@@ -740,3 +771,94 @@ SynthDef {
...
+		}
+	}
+}
+
+Resample {
+	var <>synthDef;
+	var <>factor;
+	var <>controlIndex;
+
+	*new { arg factor=1.0;
```
JordanHendersonMusic, line 824: Similar here, factor can't be nil, perhaps it is better to use `factor (1.0)`?

JordanHendersonMusic (commented):
Couple of queries, otherwise this seems ready to merge.

include/server/ErrorMessage.hpp
Spacechild1: Thanks, done!

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: Ah, you mean in case someone explicitly passes `nil`? There are so many places in the Class Library that use "simple" default values without explicitly checking for `nil`. I'd rather follow the existing style and hope for a general solution to this problem.

JordanHendersonMusic: Or implicitly, those are the real issue!

```
Foo.new(f.());


{ |a|
   Foo.new(a
};

var a = ...;
Foo.new(a);
```

If `a (1)` means ` a = a ?? {1}`, that a is never nil, then `a = 1` must mean that `nil` is valid in the function and therefore the code should be expecting it.


> I'd rather follow the existing style and hope for a general solution to this problem.

If there isn't one, then we are going to have to accept the existing 'style' is wrong, potentially changing a lot of code... 

It's also not simply a style issue either, because it has different behaviour. 

---

Given this error occurs everywhere in the class library and in the quarks (I think I've seen two places where this is used correctly...) this isn't really the place to discuss this. But I think your reluctance to change it interesting and probably mirrored by others!

JordanHendersonMusic (approved):

JordanHendersonMusic:
Looks good to merge imo. Appreciate the throughoutness of the test!

SCClassLibrary/Common/Audio/SynthDef.sc
Spacechild1: > It's also not simply a style issue either, because it has different behaviour.

Of course, `blockSize = (0.0)` would be the correct behavior. However, almost no code in the Class Library does it, so I'd rather be deliberately wrong, but consistent. The whole thing has to be fixed one way or another, but it's not clear yet what this fix will be eventually.

Spacechild1:
Thanks for the review! @dyfer Should I wait for the dev meeting or can I just merge it right away?

Issue #5939, opened 2022-12-20
Plugin API Extension: `SendReply`


weefuzzy:
## Motivation

Extending the Plugin API to allow sending OSC replies to clients directly from the Command FIFO thread. 

## Description of Proposed Feature

The Plugin API allows plugins to run custom jobs on the Command FIFO (i.e. not the audio thread) via `fDoAsyncronousCommand`. However, there is no publicly exposed way to send a reply to clients from this thread  – the relevant function (in `scsynth`), `SendReply` is not in the `InterfaceTable`. As a result one has to dispatch a reply back via the audio thread. This then creates problems for clients waiting on replies because an additional asynchronous indirection is introduced.  

This proposes that the Plugin API is extended with three functions: `SendReply` and two extra to copy and delete a `ReplyAddr` that is hidden behind a `void*`. Something like

```c
void  fSendReply(void* inReplyAddr, char* inBuf, int inSize); 
void* fCopyReplyAddress(void* inreply); 
void  fDeleteReplyAddress(void* inreply); 
```

In [FluCoMa](https://github.com/flucoma/flucoma-sc) we've ended up doing this 'illegally' by linking to the `ReplyAddr` structure and making these functions ourselves, because the pain for users outweighed the pain of doing something brittle and bad. Following some discussion in https://github.com/supercollider/supercollider/issues/5347 and https://github.com/flucoma/flucoma-sc/issues/149, it looks like the coming bump in Boost versions will result in our comeuppance, hence making an official request.  

## Plan for Implementation

The main thing is there being agreement on a mechanism for extending the plugin API robustly. If a mechanism were available, or in flight, I would happily and gratefully do the implementation for these functions. 

For `scsynth` the implementation is trivial and FluCoMa's functions, which have been out in the world for some years without problems, can be transplanted directly. `SendReply` merely forwards to an already existing internal function in `SC_ReplyImpl.hpp`, and the other two functions are just casts and copies. In C++, from our code: 

```c++
void* copyReplyAddress(void* inreply)
{
  if (!inreply) return nullptr;
  ReplyAddress* reply = new ReplyAddress();
  *reply = *(static_cast<ReplyAddress*>(inreply));
  return reply;
}

void deleteReplyAddress(void* inreply)
{
  if (!inreply) return;
  delete static_cast<ReplyAddress*>(inreply);
}

void SendReply(void* inReplyAddr, char* inBuf, int inSize)
{
[...]

PR #7405, opened 2026-03-10
Plugin command and unit command extensions

### LINKED ISSUE #7359: Add DoAsyncUnitCmd API function
Currently, we have to (ab)use `DoAsynchronousCmd` to perform asynchronous unit commands.

It would like to add a `DoAsyncUnitCmd` API function:

```
typedef SCBool (*UnitCmdStageFn)(Unit* inUnit, void* cmdData, void* replyAddr);

void (*fDoAsyncUnitCmd)(Unit* inUnit, void* replyAddr, void* cmdData,
        UnitStageFn stage2, // stage2 is non real time
        UnitStageFn stage3, // stage3 is real time - completion msg performed if stage3 returns true
        UnitStageFn stage4, // stage4 is non real time - sends done if stage4 returns true
        AsyncFreeFn cleanup, int32 completionMsgSize, const void* completionMsgData);
```

Note that the function takes a pointer to the Unit. The Server would then keep the Unit alive as long as there are pending asynchronous Unit commands. This can be done by adding a reference count to the `Graph`.

This would solve a very annoying lifetime issue where a Unit could be destroyed while an async command is still running. Instead, freeing the Graph would only decrement the reference count and remove it from the owning Node. When a Unit command finishes, it decrements the Graph's reference count as well. The one who decrements to zero actually frees the Graph.

When a Graph is freed while a unit is running, the `inUnit` argument of `UnitCmdStageFn` in stage3 (RT) is set to `nullptr`. This means the Unit command function only has to check that argument to know that it can safely exchange data with the Unit.

Spacechild1:
## Purpose and Motivation

This PR adds the following extensions to the plugin API:

- `DoAsynchronousCommandEx`: this is the same as `DoAsynchronousCommand`, except it takes function pointers of type `AsyncStageFnEx` instead of `AsyncStageFn`. Contrary to the latter, `AsyncStageFnEx` also passes the reply address. This is necessary for future planned extensions that allow users to send arbitrary messages back to the client from the various stage functions. See also #5939.

- `DefineUnitCmdEx`: this is the same as `DefineUnitCmd`, except that it takes a `UnitCmdFuncEx` instead of `UnitCmdFunc`.
Contrary to the latter, `UnitCmdFuncEx` also passes the reply address. This is necessary so that users can call `DoAsyncUnitCommand` (see below)

- `DoAsyncUnitCommand`: this is similar to `DoAsynchronousCommandEx`, with the following important differences:
  - `AsyncUnitStageFn` takes a `Unit*` instead of a `World*` so that the user can access the Unit in the various stage functions. Typically, this is used in stage3 (= RT) to safely exchange data between the Unit and the user data struct.
  - `DoAsyncUnitCommand` keeps the owning Synth alive for the full duration of the command. This is done by incrementing/decrementing the reference count.
    - scsynth: I have added a `mRefCount` field to the `Graph` structure together with `Graph_AddRef` and `Graph_Release` functions.
    - supernova: incidentally, `server_node` already has its own reference counting mechanism.
    
    *This fixes a long-standing and critical lifetime issue with using async commands inside Units!*

  - Before we call the stage3 function (= RT), we check if the Graph/synth still has a parent. If not, this means that it has been freed by the user. In this case we pass NULL as the `Unit*` so that the user can detect the situation and safely cancel the operation. Note that they can still continue with the stage4 function, e.g. to free any resources they just allocated.

  - I had to slightly change the way Graphs/synths are freed:
    - scsynth: Previously, `Node_Delete` would directly call `Graph_Dtor`. Now it calls the new `Graph_Delete` function which only decrements the refcount. If the new refcount is not zero, it means there are still pending unit commands. In this case we only remove the Node from the Server tree and release the Node ID, but we do not actually destroy the Graph yet! `Graph_Dtor` is only called when the reference count goes to zero, i.e. after the last pending unit command has finished.

    - supernova: When a synth is freed by the user (directly or indirectly), the node gets removed from the node tree. This automatically decrements its refcount. Previously, `server::remove_node()` would always call `sc_synth::finalize()` on all synth children, which destroys all the units. However, we actually want to keep the units alive as long as they are being referenced by pending unit commands. Instead we simply call `finalize()` directly in `~sc_synth()`, i.e. right before the synth gets destroyed. (The one who decrements the refcount to zero destroys the synth.)

  Closes #7359.

I have updated the async plugin command and unit command examples in `DemoUGens.cpp`.

---

Minor fixes/changes/additions I made on the way:

- scsynth: the cleanup function passed to `DoAsynchronousCommand` can now be be `NULL`. (This was already the case in supernova.) The same is true for `DoAsynchronousCommandEx`.

- supernova: make `server_node` refcount non-atomic to avoid unnecessary atomic operations. The refcount is only ever incremented/decremented on the RT thread via operations on the node tree. Since these operations are not thread safe anyway, there is no reason why the refcount should be atomic.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/scsynth/SC_SequencedCommand.cpp
+++ b/server/scsynth/SC_SequencedCommand.cpp
@@ -1594,3 +1594,95 @@ template <typename StageFn> void AsyncPlugInCmd_<StageFn>::Stage4() {
     if (result && mCmdName && mReplyAddress.mReplyFunc != null_reply_func)
         SendDone(mCmdName);
 }
+
+///////////////////////////////////////////////////////////////////////////
+
+SCErr PerformAsyncUnitCommand(
+    Unit* inUnit, void* replyAddr, const char* cmdName, void* cmdData,
+    AsyncUnitStageFn stage2, // stage2 is non real time
+    AsyncUnitStageFn stage3, // stage3 is real time - completion msg performed if stage3 returns true
+    AsyncUnitStageFn stage4, // stage4 is non real time - sends done if stage4 returns true
+    AsyncFreeFn cleanup, int completionMsgSize, const void* completionMsgData) {
+    if (!Graph_HasParent(inUnit->mParent)) {
+        // If Graph_HasParent() returns false, it means that the graph has been removed. This can only
+        // happen if DoAsyncUnitCommand() is called in a Unit destructor (which is not allowed).
+        // This check is important because it makes sure that we don't increment a reference count
+        // that has already gone to zero!
+        scprintf("ERROR: cannot call DoAsyncUnitCommand() in a Unit destructor!\n");
+        return kSCErr_Failed;
+    }
+
+    void* space = World_Alloc(inUnit->mWorld, sizeof(AsyncUnitCmd));
+    ReturnSCErrIfNil(space);
+    AsyncUnitCmd* cmd = new (space) AsyncUnitCmd(inUnit, (ReplyAddress*)replyAddr, cmdName, cmdData, stage2, stage3,
+                                                 stage4, cleanup, completionMsgSize, completionMsgData);
+    if (!cmd)
+        return kSCErr_Failed;
+    if (inUnit->mWorld->mRealTime)
+        cmd->CallNextStage();
+    else
+        cmd->CallEveryStage();
+    return kSCErr_None;
+}
+
+///////////////////////////////////////////////////////////////////////////
+
+AsyncUnitCmd::AsyncUnitCmd(
+    Unit* inUnit, ReplyAddress* inReplyAddress, const char* cmdName, void* cmdData,
+    AsyncUnitStageFn stage2, // stage2 is non real time
+    AsyncUnitStageFn stage3, // stage3 is real time - completion msg performed if stage3 returns true
+    AsyncUnitStageFn stage4, // stage4 is non real time - sends done if stage4 returns true
+    AsyncFreeFn cleanup, // cleanup is called in real time
+    int completionMsgSize, const void* completionMsgData):
+    SC_SequencedCommand(inUnit->mWorld, inReplyAddress),
+    mUnit(inUnit),
+    mCmdName(cmdName),
+    mCmdData(cmdData),
+    mStage2(stage2),
+    mStage3(stage3),
+    mStage4(stage4),
+    mCleanup(cleanup),
+    mAlive(true) {
+    // make sure that the owning Graph is kept alive for the whole duration of the command!
+    Graph_AddRef(inUnit->mParent);
+
+    if (completionMsgSize > 0 && completionMsgData) {
+        mMsgSize = completionMsgSize;
+        mMsgData = (char*)World_Alloc(mWorld, mMsgSize);
+        ThrowIfNil(mMsgData);
+        memcpy(mMsgData, completionMsgData, mMsgSize);
+    }
+}
+
+AsyncUnitCmd::~AsyncUnitCmd() {
+    if (mCleanup)
+        mCleanup(mWorld, mCmdData);
+    // finally release the owning Graph
+    Graph_Release(mUnit->mParent);
+}
+
+void AsyncUnitCmd::CallDestructor() { this->~AsyncUnitCmd(); }
+
+bool AsyncUnitCmd::Stage2() {
+    bool result = !mStage2 || (mStage2)(mUnit, mCmdData, &mReplyAddress);
+    return result;
+}
+
+bool AsyncUnitCmd::Stage3() {
+    // Check whether the owning Graph has been removed in the meantime. If yes, we pass
+    // nullptr as the Unit pointer so that the stage function can detect and properly
+    // handle the situation. For example, it may still have to release resources on stage4.
+    mAlive = Graph_HasParent(mUnit->mParent);
+    bool result = !mStage3 || (mStage3)(mAlive ? mUnit : nullptr, mCmdData, &mReplyAddress);
+    // Only send completition message if the Graph is still alive!
+    if (result && mAlive)
+        SEND_COMPLETION_MSG;
+    return result;
+}
+
+void AsyncUnitCmd::Stage4() {
+    bool result = !mStage4 || (mStage4)(mUnit, mCmdData, &mReplyAddress);
+    // Only send /done message if the Graph has been alive in stage3.
+    if (result && mAlive && mCmdName && mReplyAddress.mReplyFunc != null_reply_func)
+        SendDone(mCmdName);
+}
--- a/server/scsynth/SC_SequencedCommand.cpp
+++ b/server/scsynth/SC_SequencedCommand.cpp
@@ -20,6 +20,9 @@
 
 
 #include "SC_SequencedCommand.h"
+#include "SC_GraphDef.h"
+#include "SC_BufGen.h"
+#include "SC_Unit.h"
 #include "SC_CoreAudio.h"
 #include "SC_Errors.h"
 #include "scsynthsend.h"
@@ -1510,11 +1513,33 @@ SCErr PerformAsynchronousCommand(
 
 ///////////////////////////////////////////////////////////////////////////
 
-AsyncPlugInCmd::AsyncPlugInCmd(
+SCErr PerformAsynchronousCommandEx(
+    World* inWorld, void* replyAddr, const char* cmdName, void* cmdData,
+    AsyncStageFnEx stage2, // stage2 is non real time
+    AsyncStageFnEx stage3, // stage3 is real time - completion msg performed if stage3 returns true
+    AsyncStageFnEx stage4, // stage4 is non real time - sends done if stage4 returns true
+    AsyncFreeFn cleanup, int completionMsgSize, const void* completionMsgData) {
+    void* space = World_Alloc(inWorld, sizeof(AsyncPlugInCmdEx));
+    ReturnSCErrIfNil(space);
+    AsyncPlugInCmdEx* cmd = new (space) AsyncPlugInCmdEx(inWorld, (ReplyAddress*)replyAddr, cmdName, cmdData, stage2,
+                                                         stage3, stage4, cleanup, completionMsgSize, completionMsgData);
+    if (!cmd)
+        return kSCErr_Failed;
+    if (inWorld->mRealTime)
+        cmd->CallNextStage();
+    else
+        cmd->CallEveryStage();
+    return kSCErr_None;
+}
+
+///////////////////////////////////////////////////////////////////////////
+
+template <typename StageFn>
+AsyncPlugInCmd_<StageFn>::AsyncPlugInCmd_(
     World* inWorld, ReplyAddress* inReplyAddress, const char* cmdName, void* cmdData,
-    AsyncStageFn stage2, // stage2 is non real time
-    AsyncStageFn stage3, // stage3 is real time - completion msg performed if stage3 returns true
-    AsyncStageFn stage4, // stage4 is non real time - sends done if stage4 returns true
+    StageFn stage2, // stage2 is non real time
+    StageFn stage3, // stage3 is real time - completion msg performed if stage3 returns true
+    StageFn stage4, // stage4 is non real time - sends done if stage4 returns true
     AsyncFreeFn cleanup, // cleanup is called in real time
     int completionMsgSize, const void* completionMsgData):
     SC_SequencedCommand(inWorld, inReplyAddress),
@@ -1524,7 +1549,7 @@ AsyncPlugInCmd::AsyncPlugInCmd(
     mStage3(stage3),
     mStage4(stage4),
     mCleanup(cleanup) {
-    if (completionMsgSize && completionMsgData) {
+    if (completionMsgSize > 0 && completionMsgData) {
         mMsgSize = completionMsgSize;
         mMsgData = (char*)World_Alloc(mWorld, mMsgSize);
         ThrowIfNil(mMsgData);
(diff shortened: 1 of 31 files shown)


Spacechild1:
I made sure to split the PR into self-contained meaningful commits. If you feel intimidated by the large change set, I'd recommend reading through the individual commits.

include/plugin_interface/sc_msg_iter.h
```diff
@@ -69,18 +74,62 @@ struct sc_msg_iter {
...
+    /** @brief Skips a blob argument. */
     void skipb();
-    size_t remain() { return (size_t)(endpos - rdpos); }
 
-    char nextTag(char defaultTag = 'f') { return tags ? tags[count] : defaultTag; }
+    /** @brief returns the number of remaining bytes */
+    size_t remain() const { return (size_t)(endpos - rdpos); }
+
+    /** @brief Returns the next tag, or '\0' if the end of tags is reached */
+    char nextTag(char defaultTag = 'f') const { return tags ? tags[count] : defaultTag; }
```
JordanHendersonMusic: Why use default values here rather than an optional? I see how this is okay for string, but in the literal cases you actually don't have a reliable way to check for an incorrect tag.

Also, in your docs you refer to the `argument` but that sounds like the argument of the function, not the next piece of data in the buffer. 

nextTag(char)'s documentation is wrong, it returns the default.

include/plugin_interface/sc_msg_iter.h
```diff
@@ -298,13 +347,34 @@ inline void sc_msg_iter::getb(char* outArray, size_t arraySize) {
...
+            data = rdpos + 4;
+            rdpos += sizeof(int32);
+        } else if (tags[count] == 'm') {
+            len = 4;
+            data = rdpos;
+        }
+    }
+    size_t len4 = (len + 3) & (size_t)-4;
+    rdpos += len4;
+    count++;
```
JordanHendersonMusic, line 298: Is it right to increment the count if the tag is invalid?

server/scsynth/SC_SequencedCommand.cpp
```diff
@@ -1523,32 +1548,138 @@ AsyncPlugInCmd::AsyncPlugInCmd(
...
+}
 
-void AsyncPlugInCmd::CallDestructor() { this->~AsyncPlugInCmd(); }
+template <typename StageFn> void AsyncPlugInCmd_<StageFn>::CallDestructor() { this->~AsyncPlugInCmd_(); }
 
-bool AsyncPlugInCmd::Stage2() {
-    bool result = !mStage2 || (mStage2)(mWorld, mCmdData);
+template <typename StageFn> bool AsyncPlugInCmd_<StageFn>::Stage2() {
+    bool result;
+    if constexpr (std::is_same_v<StageFn, AsyncStageFnEx>)
```
JordanHendersonMusic, line 1570: Have you thought about things like reference types? I think this is fine if it is always function pointers (or member function pointers).

server/scsynth/SC_SequencedCommand.h
```diff
@@ -508,28 +507,54 @@ class SendReplyCmd : public SC_SequencedCommand {
...
 ///////////////////////////////////////////////////////////////////////////
 
+// AsyncPlugInCmd and AsyncPlugInCmdEx only differ in the type of stage function.
+template <typename StageFn> class AsyncPlugInCmd_ : public SC_SequencedCommand {
+public:
+    AsyncPlugInCmd_(World* inWorld, ReplyAddress* inReplyAddress, const char* cmdName, void* cmdData,
+                    StageFn stage2, // stage2 is non real time
+                    StageFn stage3, // stage3 is real time - completion msg performed if stage3 returns true
+                    StageFn stage4, // stage4 is non real time - sends done if stage4 returns true
+                    AsyncFreeFn cleanup, int completionMsgSize, const void* completionMsgData);
```
JordanHendersonMusic, line 518: I was wondering if there was anyway to avoid having these comments and encode this information with two distinct strong types (more than an alias), one call `StageFnRT` and the other `StageFnNRT`. What do you think?

server/supernova/sc/sc_osc_handler.cpp
```diff
@@ -3726,65 +3733,160 @@ void handle_asynchronous_command(World* world, const char* cmdName, void* cmdDat
...
+        throw std::runtime_error("cannot call DoAsyncUnitCommand() in a Unit destructor!");
+
+    // make sure that the owning synth is kept alive for the whole duration of the command!
+    // This also means we can safely use the synth pointer inside the callbacks.
+    synth->add_ref();
+
+    cmd_dispatcher<realtime>::fire_system_callback(
+        [=, message = std::move(message), endpoint = std::move(endpoint)]() mutable {
+            // stage 2 (NRT thread)
+            bool result2 = !stage2 || stage2(unit, cmd_data, endpoint.get());
```
JordanHendersonMusic, line 3829: You might consider flipping this function with an early return as the indentation is a little much here due to the callbacks.

server/supernova/sc/sc_synth.cpp
```diff
@@ -30,13 +30,14 @@
...
     const bool rt_synthesis = world.mRealTime;
 
-    mNode.mWorld = &sc_factory->world;
-    rgen.init((uint32_t)(uint64_t)this);
-
     /* initialize sc wrapper class */
+    mNode.mWorld = &world;
+    mNode.mID = node_id;
+
+    rgen.init((uint32_t)(uint64_t)this);
```
JordanHendersonMusic: I know you didn't write this cast, but if you know what it does, would you mind changing it to be a little clearer? It is trying to convert the address of this into a uint32? Does this work on 32 bit?

Maybe like this? And let the function do the conversion to`uint32`
```
reinterpret_cast<intptr_t>(this)
```

server/supernova/sc/sc_synth.cpp
```diff
@@ -281,38 +284,46 @@ void sc_synth::map_control_buses_audio(unsigned int slot_index, int audio_bus_in
...
     } else {
         log_printf("unit index %d out of range!\n", unit_index);
     }
 }
 
+void sc_synth::queue_unit_cmd(sc_msg_iter* args, detail::endpoint_ptr const& endpoint) {
+    // NOTE: we are not in a calc function, so we don't need to use rt_alloc()!
+    // log_printf("queue unit command\n");
+    void* mem = rt_pool.malloc(sizeof(sc_unit_cmd) + args->size);
+    sc_unit_cmd* cmd = new (mem) sc_unit_cmd {};
```
JordanHendersonMusic, line 313: Should we check the allocation was successful?

server/supernova/sc/sc_ugen_factory.hpp
```diff
@@ -35,11 +37,17 @@ namespace nova {
...
 struct sc_unitcmd_def : public named_hash_entry {
-    const UnitCmdFunc func;
+    const std::variant<UnitCmdFunc, UnitCmdFuncEx> func;
 
-    sc_unitcmd_def(const char* cmd_name, UnitCmdFunc func): named_hash_entry(cmd_name), func(func) {}
+    template <typename Func> sc_unitcmd_def(const char* cmd_name, Func func): named_hash_entry(cmd_name), func(func) {}
 
-    void run(Unit* unit, struct sc_msg_iter* args) { (func)(unit, args); }
+    void run(Unit* unit, sc_msg_iter* args, detail::endpoint_ptr const& endpoint) {
+        if (std::holds_alternative<UnitCmdFuncEx>(func)) {
```
JordanHendersonMusic: Might be cleaner to use 
```
if(auto fn = std::get_if<T>(&func))
   (*fn)(....);
```
hold alternative can do conversions in some cases.

JordanHendersonMusic (commented):
I haven't thought through the bigger picture yet, and I don't know enough about the thread model on the server to comment on that. But just from going through this, it all looks very clear. 

My main point is about the osc msg iter, I've found bugs with this before where the tags weren't ones we supported, i.e., someone passing a symbol rather than a string, or a double rather than a float, and then everything breaks. It might be worth completely rethinking the API.

server/scsynth/SC_SequencedCommand.cpp
Spacechild1: `StageFn` is always a function pointer (`AsyncStageFn` or `AsyncStageFnEx`)

server/scsynth/SC_SequencedCommand.h
Spacechild1: We can't really use strong types since these are plain C function pointers and they have the same signature. All these arguments are provided by the UGen plugin through the C API.

server/supernova/sc/sc_osc_handler.cpp
Spacechild1: Good idea! Done.

server/supernova/sc/sc_synth.cpp
Spacechild1: Hah, good catch! Yes, `reinterpret_cast<intptr_t>(this)` is more appropriate here. Fixed up.

server/supernova/sc/sc_synth.cpp
Spacechild1: Good catch! Done.

server/supernova/sc/sc_ugen_factory.hpp
Spacechild1: Thanks, I forgot about `get_if`! That's much nicer!

> hold alternative can do conversions in some cases.

This would surprise me! Do you have any sources on this?

server/scsynth/SC_SequencedCommand.cpp
Spacechild1: Note that the template argument for `AsyncPluginInCmd_` is provided explicitly:
```
using AsyncPlugInCmd = AsyncPlugInCmd_<AsyncStageFn>;
using AsyncPlugInCmdEx = AsyncPlugInCmd_<AsyncStageFnEx>;
```
Any other template argument would fail with a compile time error.

include/plugin_interface/sc_msg_iter.h
Spacechild1: Thanks for catching that! This is taken from the original `getb()` overload, which must only be called when the OSC argument is really a blob/MIDI.

However, it would be nice if we could just call this unconditionally. Will fix that and update the docs.

server/scsynth/SC_SequencedCommand.cpp
JordanHendersonMusic: Ah, yes, I didn't read that close enough!

server/supernova/sc/sc_ugen_factory.hpp
JordanHendersonMusic: Just looked it up and it doesn't do conversions - I thought it turned things like `string` into a `string_view` if you asked. Must have mis-remebered. Weird naming on it though.


`get_if` is perhaps cleaner.

server/scsynth/SC_SequencedCommand.h
JordanHendersonMusic: Hmm yes I guess there isn't a good way to do this, alias would give a name, but be too weak - comments seem fine then!

server/supernova/sc/sc_ugen_factory.hpp
Spacechild1: I changed to `get_if`.

> Just looked it up and it doesn't do conversions

Yes, that's what I thought. After all, the very purpose of `std::holds_alternative` is to check if a variant member has a particular type. Implicit conversion would completely defeat this.

include/plugin_interface/sc_msg_iter.h
Spacechild1: > Why use default values here rather than an optional? I see how this is okay for string, but in the literal cases you actually don't have a reliable way to check for an incorrect tag.

This removes lots of verbose error checking on the user side. It also handles the case of missing/omitted arguments. If you want to "properly" handle wrong argument types, you can check the type with `nextTag()` before calling any of these methods.

I agree that the API could be better, but we can't really change it because we need to maintain source compatibility. Somewhere in the future we may add a modern replacement, e.g.
```cpp
class OSCArgStream {
    OSCArgStream(const sc_msg_iter& iter);
    
    std::optional<int32> getInt32();
    std::optional<float> getFloat();
    std::optional<const char*> getString();
    // etc.
};
```

> Also, in your docs you refer to the argument but that sounds like the argument of the function, not the next piece of data in the buffer.

Thanks! Yes, that's ambigious. I've improved the wording.

Spacechild1: > nextTag(char)'s documentation is wrong, it returns the default.

Not really. `nextTag` only returns `defaultTags` if there are no type tags. `tags` is the type tag string. If all arguments have been consumed, `tags[count]` returns the `\0` character of the type tag string.

In reality, all modern OSC implementations use type tags. This is what the OSC 1.0 spec has to say:
> An OSC message consists of an OSC Address Pattern followed by an OSC Type Tag String followed by zero or more OSC Arguments.
>
> Note: some older implementations of OSC may omit the OSC Type Tag string. Until all such implementations are updated, OSC implementations should be robust in the case of a missing OSC Type Tag String.
https://ccrma.stanford.edu/groups/osc/spec-1_0.html

Since the OSC 1.0 spec has been released 24 years ago, it might be safe to assume that all OSC applications now use type tags.

The behavior of OSC messages without timetags is very vague and potentially dangerous. Maybe I can go ahead and remove support for this? This can be done safely, e.g. by setting `tags` to an empty string. (In this case, the user would not get any arguments, but at least there would be no crashes.)

Spacechild1: Have a look at the latest commit. I have now added a `skiparg` method that can skip any OSC argument (including the "non-standard" ones).

Previously, `geti`, `getf`, `getd` and  `gets` could only deal with 'i', 'f', 'd', 's' and 'b' arguments; any other types would lead to data corruption. Now these methods simply call `skiparg` on wrong argument types and return the default value. I have also extended this to `gett` and `getb`. This means you can now call any of these methods on any types and never corrupt the argument stream.

JordanHendersonMusic: Thanks for addressing this, I think your changes are very clear!

One thing, nextTag looks like it might consume the osc data, it's written with a bunch of methods that do, perhaps add this to the comment, or rename it something like `peekTag`.

> This removes lots of verbose error checking on the user side. It also handles the case of missing/omitted arguments. If you want to "properly" handle wrong argument types, you can check the type with nextTag() before calling any of these methods.

Ah yes, I wasn't really thinking about this as a use case, but that makes sense.

> The behavior of OSC messages without timetags is very vague and potentially dangerous. Maybe I can go ahead and remove support for this?

I think so, but maybe that's another pr if it might cause issues.

include/plugin_interface/sc_msg_iter.h
```diff
@@ -248,63 +351,110 @@ inline const char* sc_msg_iter::gets(const char* defaultValue) {
...
+        len = OSCstrlen(rdpos);
+        break;
+    case 'b':
         len = (size_t)OSCint(rdpos);
         rdpos += sizeof(int32);
-    } else if (tags[count] == 'm')
-        len = 4;
+        break;
+    default:
+        // all other types have no data
```
JordanHendersonMusic: So the problem I can see is if the user passes in a custom/invalid tag type. We can't skip, because we don't know it's size. Can we throw in that case?

Spacechild1:
@JordanHendersonMusic Since the `sc_msg_iter` fixes turned out be larger than expected, so I'll move them to a dedicated PR: https://github.com/supercollider/supercollider/pull/7407. Could you take your last questions to the new PR?

JordanHendersonMusic (approved):
Cool in that case this is all makes sense to me!

I don't know much about the threading model here, @Spacechild1 you might be the most active member who does, so if you think this is good then feel free to merge, if you want to wait for another review that's cool.

Spacechild1:
Thanks! I still want to add some tests. Then I might just go ahead and merge this.

PR #7406, opened 2026-03-10
sclang: parser: remove unused PyrLitDictNode

### LINKED ISSUE #7375: lang: literal dictionaries do not work.
## Environment

* SuperCollider version: all
## Steps to reproduce

What does this equal?
```
#[ (foo: 1) ] 
```

Clearly, it shouldn't be
```
[ nil ]
```
 (which it is).

This is because the implementation of `PyrLitDictNode` is empty! It appears this is an unimplemented feature.

I suggest we remove it!

hayakzan:
## Purpose and Motivation

Fixes #7375

This PR removes the unused parser node `PyrLitDictNode`.

`PyrLitDictNode` currently has empty `compile()` and `dump()` implementations.

The grammar rule `dictlit2` exists only to construct `PyrLitDictNode`, so it was removed along with the `listliteral` alternative that referenced it.

The parser `lang11d_tab.cpp` was regenerated after updating `lang11d`.

I first removed only the `PyrLitDictNode` declarations/definitions and confirmed that the build still failed because `lang11d` generates parser code that calls `newPyrLitDictNode`. That showed `dictlit2` is not actually an independent feature path, it is the entry point that constructs `PyrLitDictNode`, so removing it is part of making the node removal compile cleanly. 

While exploring this area I noticed `litDictPut` appears unused, but I left that untouched for this PR to keep the scope narrow.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested 
- [x] All tests are passing 
- [ ] Updated documentation (N/A - internal C++ changes)


JordanHendersonMusic:
Excellent, thanks!! 

Could you also remove `lictDictPut` if it is unused. No point having it if literal dictionarys don't exist.

It might also be referenced in the dump byte codes file.

Don't worry about the commit history here, some of the code base has a useful history, this part is basically just the initial commit from 25 years ago.

hayakzan:
Thank you! It looks like it was only in `PyrParseNode.cpp`. Removed.

JordanHendersonMusic:
I think this is good, I just want to check the generated file matches what my bison produces! I'll try and do that this week.

JordanHendersonMusic (approved):
Great thanks for this!

PR #7407, opened 2026-03-10
improve/fix sc_msg_iter

Spacechild1:
This PR attempts to make several improvements and fixes to the `sc_msg_iter` class which is part of the public plugin API:

- document the behavior of every method with Doxygen comments

- add new `getb()` overload that returns a *view* of the blob

- `gets()` should return default value when there are no remaining arguments

- `getbsize()` and `remain()` should be const

- add `hasArgs()` method. This is more reliable than `remain()` if you want to check for remaining arguments.

- make sure that `gett()`, `geti()`, `getf()`, `getd()`, `gets()` and `getb()` can handle any argument type, including the non-standard types. If the argument type cannot be reasonably converted to the wanted type, the argument is skipped and the provided default value is returned.

- add `skiparg()` function that skips the next OSC argument (of any type)

- `getb()`: fix wrong size argument to `memcpy()`

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/include/plugin_interface/sc_msg_iter.h
+++ b/include/plugin_interface/sc_msg_iter.h
@@ -57,6 +57,20 @@ SC_INLINE float64 OSCdouble(const char* inData) {
     return slot.f;
 }
 
+struct sc_msg_blob {
+    const void* data;
+    size_t size;
+};
+
+/** @class Read the remaining OSC arguments from an OSC message.
+ *
+ *  This is used by bufgen, plugin command or unit command functions.
+ *
+ *  @attention Before SC 3.15, geti(), getf(), getd() and gets() would lead to
+ *  unexpected behavior if the OSC argument was anything other than 'i', 'f', 'd' or 's'.
+ *  Similarly, gett() and getb() did not correctly handle wrong argument types.
+ *  Since SC 3.15, these functions correctly handle all argument types.
+ */
 struct sc_msg_iter {
     const char* data;
     const char* rdpos;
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
Yes this looks good, I was just wondering if the user passed either a custom tag type or an invalid one by mistake, we wouldn't be able to skip data as we can't know it's size. Would it be possible to throw instead?

Spacechild1:
> Yes this looks good, I was just wondering if the user passed either a custom tag type or an invalid one by mistake, we wouldn't be able to skip data as we can't know it's size. Would it be possible to throw instead?

Good point!

My initial thought was also to just throw an exception. However, on second thought I'm a bit sceptical:
1. the very point of the `get*` methods is that you don't need any error handling. Now you need to at least wrap everything in a try-catch block
2. we currently don't use exceptions in the plugin API
3. not everybody likes to use exceptions

I think once we detect an unknown argument type, the stream can be considered corrupted. The only thing we really want is not crash the Server.

Maybe it's enough to just return the default value and *not* advance the typetag counter? Any subsequent calls to this or other methods would just output the default value as well.

If someone wants to do proper type/error checking, they would call `nextTag`. In this case, they would also be able to detect bad arguments and take the appropriate measures.

What do you think?

JordanHendersonMusic:
Another option would be to validate all tags in one go. Perhaps at construction, or the user could validate the stream for the next N arguments. 

I think this is used in the language too (can't  remember right now, will look tomorrow), but there exceptions are fine.

I think your idea of just keep out putting the default seems fine, it's quite a rare thing anyway, and as long as it doesn't crash, I think it's fine.

This would be a good thing to fuzz test.

---

If you are implementing a command and except argument of tag A but get B, this API will silently return the default value. I see the use of this, although the silence means the user cannot get a response. But the other 'error' you might want to handle is when the user has missed an argument in the middle. Say you are given tags A B D, but expect A B C D, the given final value will be replaced by Cs default, and D will also default, despite the given D being valid.

I think in the general case it's better to give the user the tags as an array, and let them deal with all error and unexpected tags. This means we shouldn't have to do any error checking as the call sight checks everything at once.

Something like...
```
if (auto r = msg.expectConvertible<'f', 'f'>()){
   r.get<0>() + r.get<1>()
} 
```
At this point though, it's a whole API redesign. 

I don't know how to avoid exceptions and keep the old one, but I also dislike wrapping everything in try catch blocks. Can this be caught in server code, rather than in the plugin, in nice way? Or does that have issues as it crosses a c ABI boundary? This is why I think your idea, although not ideal, is a pretty good compromise.

Spacechild1:
> Another option would be to validate all tags in one go. Perhaps at construction, or the user could validate the stream for the next N arguments.

Validating all typetags on the Server side before dispatching the messages might be an interesting solution! After all, the checking should be done for *all* incoming OSC messages -- not only plugin/unit commands -- to prevent users from crashing the Server with malformed messages. (`sc_msg_iter` is used internally by scsynth. Supernova, however, uses the oscpack library instead.) The advantage is that we can abort malformed Server commands early on and print an informative error message.

The check could simply we done in the `init` method. Here we can actually throw an exception since this is typically only called by the Server, not by the user.

> Say you are given tags A B D, but expect A B C D, the given final value will be replaced by Cs default, and D will also default, despite the given D being valid.

Yes, that's an issue. I think it depends on the situation. If the plugin author is 100% in control of the OSC arguments (e.g. because these are created by sclang methods), they can get away with not checking the typetags. As soon as the OSC arguments can come from the user (and they are not validated client-side), you should probably check them. `sc_msg_iter` supports both approaches.

> I think in the general case it's better to give the user the tags as an array,

Raw OSC messages are not suitable for random access because you need to parse and skip all arguments 0-N to know where the data for argument N+1 starts. If you want to keep it zero cost, the API must essentially be stream based. That's why oscpack has `osc::ReceivedMessageArgumentStream` and `osc::ReceivedMessageArgumentIterator`, but no random argument access.

Many high-level OSC libraries do provide random argument access, but that's because they pre-process the raw OSC messages and store them in a new (typically heap-allocated) form.

Spacechild1:
Two other things I have in mind:

- `isNumber()`: returns whether the next argument is numeric ('i', 'h', 'f', 'd' or 'c').
   Sometimes you just cannot provide a unique default value to `geti()` and friends that can be used to check for wrong argument types. In this case `isNumber()` would be very helpful.

- `isString()`: returns whether the next argument is a string ('s' or 'S'). Not sure if this is actually needed. Does anyone actually use the 'S' tag?

Spacechild1:
@JordanHendersonMusic should we take this PR as the opportunity to remove support for untyped OSC messages? `sc_msg_iter::init` could simply throw an exception if there are no typetags, as part of the verification process. This means that untyped OSC messages would never make it to the plugin in the first place. Instead the Server prints an error message.

JordanHendersonMusic:
Putting tag validation and checking they are typed in `init`, and throwing if not, sounds like a good idea. Just mark it as `noexcept(false)` or something.

I'd be in favour of both those `is*` methods. Perhaps also an `isFloating` and `isInt`? Since we cast to the smaller types `f32` & `i32` the user should have only passed larger ones by mistake, not because they need the range. The string/symbol has come up here before when talking about the class library and how it converts all osc messages to symbols.

>Raw OSC messages are not suitable for random access because you need to parse and skip all arguments 0-N to know where the data for argument N+1 starts.

Aren't the tags consecutive and fixed size  in memory though? This makes them random access. 

The little snippet I suggested would stream the arguments into a tuple after checking all the tags. Strings and whatnot would be views/spans into the buffer.

PR #7409, opened 2026-03-12
Fix: ISPOWEROFTWO returning true for 0

### LINKED ISSUE #7408: `0.isPowerOfTwo` returns true.
## Environment

* SuperCollider version: dev
* Operating system: - 
* Other details (Qt version, audio driver, etc.): -

## Steps to reproduce

```supercollider
0.isPowerOfTwo;
-> true
```

0 is not a power of two and it should return `false`.

The primitive calls

https://github.com/supercollider/supercollider/blob/613f26f9549693f1d3032145048f1fc14a863981/include/common/clz.h#L102

This function gets also used in lots of checks regarding "power of two sizes" (e.g. fft window size) - see https://github.com/search?q=repo%3Asupercollider/supercollider%20ISPOWEROFTWO&type=code

Santoshhh-25:
Fixes #7408

0.isPowerOfTwo currently returns true because the bit trick
(x & (x - 1)) == 0 also evaluates to true for x = 0.

This patch ensures the input is positive before applying the
bit check by adding x > 0.

### DIFF
--- a/include/common/clz.h
+++ b/include/common/clz.h
@@ -99,7 +99,7 @@ SC_INLINE int32 NUMBITS(int32 x) { return 32 - CLZ(x); }
 SC_INLINE int32 LOG2CEIL(int32 x) { return 32 - CLZ(x - 1); }
 
 // is x a power of two
-SC_INLINE SCBool ISPOWEROFTWO(int32 x) { return (x & (x - 1)) == 0; }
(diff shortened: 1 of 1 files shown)


Spacechild1 (approved):

Spacechild1:
Thanks!

PR #7411, opened 2026-03-14
Fix crash when receiving malformed OSC UDP packets

### LINKED ISSUE #7351: Interpreter crash when sending certain non-osc UDP packets
Certain kind of non-osc udp messages seem to crash sclang. I've encountered the behaviour when trying to send OSC from Bitwig to SuperCollider, and have been able to reproduce the crash with a small PD patch.

## Environment

- SuperCollider version: Built from branch 'develop' [fb0f85a924]
- Operating system: MacOs 14.7.2

## Steps to reproduce

First start SuperCollider/sclang and ensure that `NetAddr.langPort` is set to `57120`.

Download, open and run the following pd patch: https://gist.github.com/jpburstrom/94c78a0600900695fa2bb631c9744003

```
[ connect localhost 57120 ]  1. Connect 
        [ 0 (                2. Send non-osc message
|       /
[ netsend -u -b ]
```

When sending the message, the interpreter crashes.

SuperCollider should not crash, but probably silently ignore the message. A possible fix would be to validate the message before trying to process it. I've sketched out this for myself, and it seems to work (https://github.com/jpburstrom/supercollider/commit/291f07c6eac0d71032ab813b8df4690099300f78). If this looks like a good solution, I could make a pull request. But I'm unsure if an invalid packet should be handled in some way?

There are other similar open issues (#7327, #4561, #5267), but this issue and fix is probably not related to that since those messages would pass as valid OSC messages with the current validation function. 

Link to crash log: https://gist.github.com/jpburstrom/173a558965a4e48d4086bd4714f2d49d

Santoshhh-25:
Fixes #7351
Solution
Add validation in ProcessOSCPacket to ensure that incoming packets begin with a valid OSC prefix (/ for OSC messages or #bundle for OSC bundles).
Packets that do not match the expected OSC structure are ignored instead of being passed to the OSC parser.

Testing

1. Reproduced the crash by sending malformed UDP packets to the language port (NetAddr.langPort).

2. Verified that the interpreter previously crashed when receiving invalid packets.

3. After the fix, malformed packets are ignored and sclang continues running normally.

4. Confirmed that valid OSC messages are still processed correctly.

### DIFF
--- a/lang/LangPrimSource/OSCData.cpp
+++ b/lang/LangPrimSource/OSCData.cpp
@@ -67,7 +67,7 @@ bool gUseDoubles = false;
 
 InternalSynthServerGlobals gInternalSynthServer = { nullptr, kNumDefaultSharedControls, gDefaultSharedControls };
 
-std::unique_ptr<InPort::UDP> gUDPport {};
+std::unique_ptr<InPort::UDP> gUDPport { };
 
 PyrString* newPyrString(VMGlobals* g, char* s, int flags, bool runGC);
 
--- a/lang/LangPrimSource/OSCData.cpp
+++ b/lang/LangPrimSource/OSCData.cpp
@@ -823,18 +823,10 @@ void PerformOSCMessage(int inSize, const char* inData, PyrObject* replyObj, int 
 
 // takes ownership of inPacket
 void ProcessOSCPacket(std::unique_ptr<OSC_Packet> inPacket, int inPortNum, double time) {
-    // Validate Packet before Processing;
+    // Guard against null/empty packets
     if (!inPacket || !inPacket->mData || inPacket->mSize <= 0) {
         return;
     }
-
-    char* data = inPacket->mData.get();
-
-    // OSC message must start with '/' or "#bundle";
-    if (data[0] != '/' && strncmp(data, "#bundle", 7) != 0) {
-        return;
-    }
-
     // post("recv '%s' %d\n", inPacket->mData, inPacket->mSize);
     const auto isBundle = IsBundle(inPacket->mData.get());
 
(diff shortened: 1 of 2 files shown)


lang/LangPrimSource/OSCData.cpp
```diff
@@ -67,7 +67,7 @@ bool gUseDoubles = false;
 
 InternalSynthServerGlobals gInternalSynthServer = { nullptr, kNumDefaultSharedControls, gDefaultSharedControls };
 
-std::unique_ptr<InPort::UDP> gUDPport {};
+std::unique_ptr<InPort::UDP> gUDPport { };
```
capital-G:
```suggestion
std::unique_ptr<InPort::UDP> gUDPport {};
```

please don't touch unrelated LOC ;)

If you have clang-format related issues, you should try out our pre-commit setup, see https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit

capital-G (commented):
Thanks for the PR - can you specify which kind of tests you run b/c I am sure that the problem described in https://github.com/supercollider/supercollider/issues/7351 is not covered by the proposed check.

Santoshhh-25:
I've been working on this same issue in #7411. Regarding your question about whether invalid packets should be handled — I think silently ignoring them is the right approach since they are not valid OSC messages. However, I'd suggest combining both approaches:

Early null/size validation (defensive check)
IsMessage() check before PerformOSCMessage() (your fix)

I'm updating my PR to incorporate your IsMessage() approach along with the additional null checks. Happy to collaborate if you'd like to combine our work.

capital-G:
> I've been working on this same issue in #7411. Regarding your question about whether invalid packets should be handled — I think silently ignoring them is the right approach since they are not valid OSC messages. However, I'd suggest combining both approaches:
> 
> Early null/size validation (defensive check)

Why should this ever be `null`? This is code which gets called a lot, so unnecessary statements should be omitted and we don't have any indications that this could be null.

> IsMessage() check before PerformOSCMessage() (your fix)

I haven't proposed a fix and I wouldn't call `IsMessage` in this case.

> I'm updating my PR to incorporate your IsMessage() approach along with the additional null checks.

The referenced issue #7351 is currently not solved by this PR.

jpburstrom:
> I haven't proposed a fix and I wouldn't call `IsMessage` in this case.

I think @Santoshhh-25 refers to my original bug report (#7351), where I linked to a sketch I made (https://github.com/jpburstrom/supercollider/commit/291f07c6eac0d71032ab813b8df4690099300f78). This fixed the problem on my end (malformed messages doesn't crash supercollider anymore), but with my very limited knowledge of C++ and the codebase, I'm unsure if this is a good long term solution.

PR #7413, opened 2026-03-15
sclang: redefine that passing nil as arg means reset to default

### LINKED ISSUE #7385: sclang: passing nil to arguments with initialisers behaves differently depending on syntax.
## Steps to reproduce

```supercollider

f = { arg a1 = 1; a1 };
g = { arg a1=(1); a1 };
h = { arg a1(1); a1 };
i = { arg a1((1)); a1 };
j = { | a1 1| a1 };

f.(nil) // nil
g.(nil) // 1
h.(nil) // 1
i.(nil) // 1
j.(nil) // nil

```

This is because PyrVarDefNode::hasExpr reports that (even if it is a literal), the node has an expression if there are brackets. It still initialses the functiondef prototypeframe correctly, it just also does an extra `a = a ?? {1};`

This only occurs when initialised with a literal, if it was an expression it would always return the result of the expression.

If you call the function with no arguments, then they all agree it is `1`.

It should be the same.

I think the correct thing to do here is to make the argument the value when passing nil explicitly in all cases.

JordanHendersonMusic:
## Purpose and Motivation

Closes #7385

Fixes bug when explicitly passing `nil` would override the default argument value when it was a literal.

Adds documentation regarding what the intended behaviour is from now on.

This aligns sc's behaviour with lua. Usually python is the standard, but you can't call functions with too few arguments there so the issue doesn't arise.

Old behaviour.

```supercollider
f = { arg a1 = 1; a1 };
g = { arg a1=(1); a1 };
h = { arg a1(1); a1 };
i = { arg a1((1)); a1 };
j = { | a1 1| a1 };

f.(nil) // nil
g.(nil) // 1
h.(nil) // 1
i.(nil) // 1
j.(nil) // nil
```

After this PR, passing `nil` means, 'reset to default', therefore all the above return `1`.

One of the places where this is very annoying is here...

```supercollider
f { |a =10| ^a };
g { |a| ^this.f(a) };
```

... as `g`'s `a` is `nil` and this overrides `f`'s default,  so `g() == nil`.

Note, this is a breaking change in two cases:
- When the argument was declared as a literal, if it was an expression, or had brackets (?), you could not pass `nil` and the default was used (correctly).
- `f.(10, a1: nil)` You could also get the argument to be nil if you had a collision. Now this means, 'reset to default', and therefore returns `1`.  See documentation for examples.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [ ] Update `NodeProxy:clear`.
- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Audio/UGen.sc
+++ b/SCClassLibrary/Common/Audio/UGen.sc
@@ -64,17 +64,17 @@ UGen : AbstractFunction {
 	}
 	exprange { arg lo = 0.01, hi = 1.0;
 		^if (this.signalRange == \bipolar) {
-			this.linexp(-1, 1, lo, hi, nil)
+			this.linexp(-1, 1, lo, hi, \none)
 		} {
-			this.linexp(0, 1, lo, hi, nil)
+			this.linexp(0, 1, lo, hi, \none)
 		};
 	}
 
 	curverange { arg lo = 0.00, hi = 1.0, curve = -4;
 		^if (this.signalRange == \bipolar) {
-			this.lincurve(-1, 1, lo, hi, curve, nil)
+			this.lincurve(-1, 1, lo, hi, curve, \none)
 		} {
-			this.lincurve(0, 1, lo, hi, curve, nil)
+			this.lincurve(0, 1, lo, hi, curve, \none)
 		};
 	}
 
@@ -189,6 +189,9 @@ UGen : AbstractFunction {
 			},
 			\max, {
 				^this.min(max);
+			},
+			\none, {
+				^this;
 			}
 		);
 		^this
--- a/SCClassLibrary/Common/Control/Buffer.sc
+++ b/SCClassLibrary/Common/Control/Buffer.sc
@@ -306,7 +306,7 @@ Buffer {
 	}
 
 	// risky without wait
-	getToFloatArray { arg index = 0, count = -1, wait = 0.01, timeout = 3, action;
+	getToFloatArray { arg index = 0, count, wait = 0.01, timeout = 3, action;
 		var refcount, array, pos, getsize, resp, done = false;
 
 		pos = index = index.asInteger;
--- a/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
@@ -187,8 +187,9 @@ NodeProxy : BusPlug {
 
 
 
-	putAll { | list, index = (0), channelOffset = 0 |
+	putAll { | list, index, channelOffset = 0 |
 		channelOffset = channelOffset.asArray;
+		index = index ?? {0};
 		if(index.isSequenceableCollection) {
 			max(list.size, index.size).do { |i|
 				this.put(index.wrapAt(i), list.wrapAt(i), channelOffset.wrapAt(i))
--- a/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
@@ -47,19 +47,19 @@ NodeProxy : BusPlug {
 
 	isPlaying { ^group.isPlaying }
 
-	free { | fadeTime, freeGroup = true |
+	free { | fadeTime = (this.fadeTime), freeGroup = true |
 		var bundle, freetime;
 		var oldGroup = group;
 		if(this.isPlaying) {
 			bundle = MixedBundle.new;
-			if(fadeTime.notNil) {
+			if(fadeTime != 0) {
 				bundle.add([15, group.nodeID, "fadeTime", fadeTime]) // n_set
 			};
 			this.stopAllToBundle(bundle, fadeTime);
 			if(freeGroup) {
 				oldGroup = group;
 				group = nil;
-				freetime = (fadeTime ? this.fadeTime) + (server.latency ? 0) + 1e-9; // delay a tiny little
+				freetime = fadeTime + (server.latency ? 0) + 1e-9; // delay a tiny little
 				server.sendBundle(freetime, [11, oldGroup.nodeID]); // n_free
 			};
 			bundle.send(server);
--- a/HelpSource/LanguageSpecification/ArgumentPassing.schelp
+++ b/HelpSource/LanguageSpecification/ArgumentPassing.schelp
@@ -1,0 +1,48 @@
+title:: Argument Passing
+summary:: How arguments are passed to function/methods
+categories:: LanguageSpecification
+
+SuperCollider supports positional arguments, keyword arguments, environmental argument lookup, and default arguments.
+These features all combine to sometimes create confusing results. 
+This document provides the rules for how one can expect these conflicts to be ressolved.
+
+section:: Nil and defaults
+
+One major source of confusion is what code::nil:: means at the call sight.
+
+In previous SuperCollider versions the behaviour changed depending on subtle things like the syntax used to declare default argument values. 
+This has since been removed, now code::nil:: has a precise meaning...
+
+code::nil:: means to reset the argument to the default value. If no default value is provided, this is code::nil:: itself.
+
+Here are some examples.
+
+code::
+f = { |a = 1| a };
+
+f.(); // 1
+
+f.(10); // 10
+f.(a: 10); // 10
+
+f.(nil) // 1
+f.(a: nil) // 1
+
+f.(10, a: nil) // 1 - also posts warning about argument collision
+::
+
+
+This must hold true regardless of how the default argument is declared or internally evaluated.
+
+
+section:: Argument Collisions
+
+When positional arguments (arguments without keywords) and keyword arguments collide, the right most argument is the final value.
+
+code::
+f = { |a = 1| a };
+f.(10, a: nil, a: 20, a: 11) // 11
+::
+
+To break this example down: first code::a:: is set to the default of code::1::, then to code::10::, followed by returning to code::1:: as code::nil:: resets to the default, before setting to code::20::, followed by code::11::.
+Ultimately, this means all but the final keyword assignment is irrelevent and the code should just be code::f.(a: 11):: or code::f.(11)::.
--- a/SCClassLibrary/Common/Audio/Env.sc
+++ b/SCClassLibrary/Common/Audio/Env.sc
@@ -219,13 +219,15 @@ Env {
 		^releaseNode.notNil
 	}
 
-	asMultichannelSignal { arg length = 400, class = (Signal);
+	asMultichannelSignal { arg length = 400, class;
 		var multiChannelArray = this.asMultichannelArray;
 		var channelCount = multiChannelArray.size;
 
 		var totalDur = this.totalDuration;
 
 		length = max(length, levels.size);
+
+		class = class ?? {Signal};
 
 		^multiChannelArray.collect { |chan|
 			var signal, ratio;
--- a/SCClassLibrary/Common/Audio/SynthDefOld.sc
+++ b/SCClassLibrary/Common/Audio/SynthDefOld.sc
@@ -165,11 +165,12 @@ SynthDefOld : SynthDef {
 
 + Object {
 
-	writeDefFileOld { arg name, dir, overwrite = (true);
+	writeDefFileOld { arg name, dir, overwrite;
 
 		StartUp.defer { // make sure the synth defs are written to the right path
 			var file;
 			dir = dir ? SynthDef.synthDefDir;
+			overwrite = overwrite ? true;
 			if (name.isNil) { Error("missing SynthDef file name").throw } {
 				name = dir +/+ name ++ ".scsyndef";
 				if(overwrite or: { pathMatch(name).isEmpty })
--- a/SCClassLibrary/Common/Control/GraphBuilder.sc
+++ b/SCClassLibrary/Common/Control/GraphBuilder.sc
@@ -20,8 +20,8 @@ GraphBuilder {
 		})
 	}
 
-	*makeFadeEnv { arg fadeTime = (0.02);
-		var dt = NamedControl.kr(\fadeTime, fadeTime);
+	*makeFadeEnv { arg fadeTime;
+		var dt = NamedControl.kr(\fadeTime, fadeTime ?? {0.02});
 		var gate = NamedControl.kr(\gate, 1.0);
 		var startVal = (dt <= 0);
 
--- a/SCClassLibrary/Common/Control/Spec.sc
+++ b/SCClassLibrary/Common/Control/Spec.sc
@@ -41,8 +41,9 @@ ControlSpec : Spec {
 	var <minval, <maxval, <warp, <step, <>default, <>units, >grid;
 	var <clipLo, <clipHi;
 
-	*new { arg minval = (0.0), maxval = (1.0), warp = ('lin'), step = (0.0), default, units, grid;
-		^super.newCopyArgs(minval, maxval, warp, step, default ? minval, units ? "", grid).init
+	*new { arg minval, maxval, warp, step, default, units, grid;
+		// In version 3.16, these args can simply be `args minval = 0.0`, including `default = minval`.
+		^super.newCopyArgs(minval ?? {0.0}, maxval ?? {1.0}, warp ?? {'lin'}, step ?? {0.0}, default ?? {minval}, units ?? {""}, grid).init
 	}
 
 	*newFrom { arg similar;
--- a/SCClassLibrary/Common/Core/Object.sc
+++ b/SCClassLibrary/Common/Core/Object.sc
@@ -705,11 +705,12 @@ Object {
 		^this.performBinaryOpOnSomething(aSelector, thing, adverb)
 	}
 
-	writeDefFile { arg name, dir, overwrite = (true);
+	writeDefFile { arg name, dir, overwrite;
 
 		StartUp.defer { // make sure the synth defs are written to the right path
 			var file;
 			dir = dir ? SynthDef.synthDefDir;
+			overwrite = overwrite ? true;
 			if (name.isNil or: { name.asString.isEmpty }) { Error("missing SynthDef file name").throw } {
 				name = dir +/+ name ++ ".scsyndef";
 				if(overwrite or: { pathMatch(name).isEmpty })
--- a/SCClassLibrary/Common/Core/Thread.sc
+++ b/SCClassLibrary/Common/Core/Thread.sc
@@ -18,8 +18,9 @@ Thread : Stream {
 	var <executingPath, <oldExecutingPath;
 	var rescheduledTime;
 
-	*new { arg func, stackSize = (512);
-		^super.new.init(func, stackSize)
+	*new { arg func, stackSize;
+		// In version 3.16 this can simply be `arg stackSize = 512`;
+		^super.new.init(func, stackSize ?? { 512 })
 	}
 	init { arg argFunc, argStackSize = 512;
 		_Thread_Init
--- a/SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Control/server-scope.sc
@@ -1,6 +1,8 @@
 + Server {
-	scope { arg numChannels, index = 0, bufsize = 4096, zoom = (1), rate = \audio;
+	scope { arg numChannels, index = 0, bufsize = 4096, zoom, rate = \audio;
 		numChannels = numChannels ?? { if (index == 0) { options.numOutputBusChannels } { 2 } };
+
+		zoom = zoom ?? { 1 };
 
 		if(scopeWindow.isNil) {
 			scopeWindow = Stethoscope(this, numChannels, index, bufsize, zoom, rate, nil,
--- a/SCClassLibrary/Common/Streams/Stream.sc
+++ b/SCClassLibrary/Common/Streams/Stream.sc
@@ -337,13 +337,14 @@ PauseStream : Stream {
 
 	isPlaying { ^stream.notNil }
 
-	play { arg argClock, doReset = (false), quant;
+	play { arg argClock, doReset, quant;
 		// check 'quant' first, before changing any state variables
 		// if there is an error, this object should be unaffected
 		var tempClock = argClock ? clock ? TempoClock.default;
 		var time = quant.asQuant.nextTimeOnGrid(tempClock);
 
 		if (stream.notNil, { "already playing".postln; ^this });
+		doReset = doReset ?? { false };
 		if (doReset, { this.reset });
 		clock = tempClock;
 		streamHasEnded = false;
@@ -560,13 +561,14 @@ EventStreamPlayer : PauseStream {
 
 	asEventStreamPlayer { ^this }
 
-	play { arg argClock, doReset = (false), quant;
+	play { arg argClock, doReset, quant;
 		// check 'quant' first, before changing any state variables
 		// if there is an error, this object should be unaffected
 		var tempClock = argClock ? clock ? TempoClock.default;
 		var time = quant.asQuant.nextTimeOnGrid(tempClock);
 
 		if (stream.notNil, { "already playing".postln; ^this });
+		doReset = doReset ?? { false };
 		if (doReset, { this.reset });
 		clock = tempClock;
 		streamHasEnded = false;
--- a/SCClassLibrary/Common/UnitTesting/UnitTest.sc
+++ b/SCClassLibrary/Common/UnitTesting/UnitTest.sc
@@ -427,8 +427,9 @@ UnitTestResult {
 
 	var <testClass, <testMethod, <message, <details, <debug;
 
-	*new { |testClass, testMethod, message = "", details, debug|
-		^super.newCopyArgs(testClass ? this, testMethod ? thisMethod, message, details, debug)
+	*new { |testClass, testMethod, message, details, debug|
+		// In version 3.16 this can simply be |message = ""|.
+		^super.newCopyArgs(testClass ? this, testMethod ? thisMethod, message ?? {""}, details, debug)
 	}
 
 	report { |brief=false|
--- a/SCClassLibrary/JITLib/ProxySpace/InBus.sc
+++ b/SCClassLibrary/JITLib/ProxySpace/InBus.sc
@@ -310,11 +310,11 @@ Monitor {
 		this.playNToBundle(bundle, outs, amps, ins, vol, fadeTime, group, addAction, multi: multi)
 	}
 
-	newGroupToBundle { | bundle, target, addAction=(\addToTail) |
+	newGroupToBundle { | bundle, target, addAction |
 		target = target.asGroup;
 		group = Group.basicNew(target.server);
 		group.isPlaying = true;
-		bundle.add(group.newMsg(target, addAction));
+		bundle.add(group.newMsg(target, addAction ?? {\addToTail}));
 		NodeWatcher.register(group);
 	}
 
--- a/common/SC_Version.hpp.in
+++ b/common/SC_Version.hpp.in
@@ -18,6 +18,7 @@
  *    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
  */
 
+#pragma once
 #include <string>
 #include <sstream>
 #include <cstdint>
--- a/lang/LangSource/PyrInterpreter3.cpp
+++ b/lang/LangSource/PyrInterpreter3.cpp
@@ -2287,7 +2287,9 @@ HOT void Interpret(VMGlobals* g) {
                     auto protoframe = meth->prototypeFrame.getPyrObjType<PyrObject>();
                     if (index < numArgsPushed) {
                         if (sp[index].isNil() && !protoframe->slots[index].isNil())
-                            v_errors::PassingNilToLiteralInitArg::print_error();
+                            v_errors::PassingNilToLiteralInitArg::print_error(
+                                meth->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
+                                meth->name.getSymbol()->name);
                         slotCopy(sp, sp + index);
                     } else {
                         slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
@@ -2415,7 +2417,9 @@ HOT void Interpret(VMGlobals* g) {
                     auto protoframe = meth->prototypeFrame.getPyrObjType<PyrObject>();
                     if (index < numArgsPushed) {
                         if (sp[index].isNil() && !protoframe->slots[index].isNil())
-                            v_errors::PassingNilToLiteralInitArg::print_error();
+                            v_errors::PassingNilToLiteralInitArg::print_error(
+                                meth->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
+                                meth->name.getSymbol()->name);
                         slotCopy(sp, sp + index);
                     } else {
                         slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
--- a/lang/LangSource/PyrInterpreter3.cpp
+++ b/lang/LangSource/PyrInterpreter3.cpp
@@ -40,6 +40,8 @@
 #include <signal.h>
 #include "SpecialSelectorsOperatorsAndClasses.h"
 
+#include "VersionedErrors.h"
+
 #include <float.h>
 #define kBigBigFloat DBL_MAX
 #define kSmallSmallFloat DBL_MIN
@@ -2276,10 +2278,21 @@ HOT void Interpret(VMGlobals* g) {
             case methReturnArg:
                 sp -= numArgsPushed - 1;
                 index = methraw->specialIndex; // zero is index of the first argument
-                if (index < numArgsPushed && !(sp + index)->isNil())
-                    slotCopy(sp, sp + index);
-                else
-                    slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
+                if constexpr (v_errors::PassingNilToLiteralInitArg::do_new()) {
+                    if (index < numArgsPushed && !(sp + index)->isNil())
+                        slotCopy(sp, sp + index);
+                    else
+                        slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
+                } else {
+                    auto protoframe = meth->prototypeFrame.getPyrObjType<PyrObject>();
+                    if (index < numArgsPushed) {
+                        if (sp[index].isNil() && !protoframe->slots[index].isNil())
+                            v_errors::PassingNilToLiteralInitArg::print_error();
+                        slotCopy(sp, sp + index);
+                    } else {
+                        slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
+                    }
+                }
                 break;
 
             case methReturnInstVar:
@@ -2393,10 +2406,20 @@ HOT void Interpret(VMGlobals* g) {
                 sp = g->sp;
                 sp -= numArgsPushed - 1;
                 index = methraw->specialIndex; // zero is index of the first argument
-                if (index < numArgsPushed && !(sp + index)->isNil()) {
-                    slotCopy(sp, sp + index);
+                if constexpr (v_errors::PassingNilToLiteralInitArg::do_new()) {
+                    if (index < numArgsPushed && !(sp + index)->isNil())
+                        slotCopy(sp, sp + index);
+                    else
+                        slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
                 } else {
-                    slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
+                    auto protoframe = meth->prototypeFrame.getPyrObjType<PyrObject>();
+                    if (index < numArgsPushed) {
+                        if (sp[index].isNil() && !protoframe->slots[index].isNil())
+                            v_errors::PassingNilToLiteralInitArg::print_error();
+                        slotCopy(sp, sp + index);
+                    } else {
+                        slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
+                    }
                 }
                 break;
 
--- a/lang/LangSource/PyrInterpreter3.cpp
+++ b/lang/LangSource/PyrInterpreter3.cpp
@@ -2276,7 +2276,7 @@ HOT void Interpret(VMGlobals* g) {
             case methReturnArg:
                 sp -= numArgsPushed - 1;
                 index = methraw->specialIndex; // zero is index of the first argument
-                if (index < numArgsPushed)
+                if (index < numArgsPushed && !(sp + index)->isNil())
                     slotCopy(sp, sp + index);
                 else
                     slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
@@ -2393,7 +2393,7 @@ HOT void Interpret(VMGlobals* g) {
                 sp = g->sp;
                 sp -= numArgsPushed - 1;
                 index = methraw->specialIndex; // zero is index of the first argument
-                if (index < numArgsPushed) {
+                if (index < numArgsPushed && !(sp + index)->isNil()) {
                     slotCopy(sp, sp + index);
                 } else {
                     slotCopy(sp, &slotRawObject(&meth->prototypeFrame)->slots[index]);
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -449,10 +449,14 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
         size_t i { 0 };
         for (PyrSlot* it { pushedArgs }; it < (pushedArgs + numNormalArgsAdded); ++it, ++i) {
             if (it->isNil() && !proto->slots[i].isNil()) {
-                auto m = callFrame->method.getPyrObjType<PyrMethod>();
-                v_errors::PassingNilToLiteralInitArg::print_error(
-                    m->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
-                    callFrame->method.getPyrObjType<PyrMethod>()->name.getSymbol()->name);
+                if (isMethod) {
+                    auto m = callFrame->method.getPyrObjType<PyrMethod>();
+                    v_errors::PassingNilToLiteralInitArg::print_error(
+                        m->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
+                        callFrame->method.getPyrObjType<PyrMethod>()->name.getSymbol()->name);
+                } else {
+                    v_errors::PassingNilToLiteralInitArg::print_error("Function", "value");
+                }
             }
             outCallFrameStack[i] = *it;
         }
@@ -485,10 +489,14 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
                     *(outCallFrameStack + *argIndex) = *argValue;
             } else {
                 if (argValue->isNil() && !proto->slots[*argIndex].isNil()) {
-                    auto m = callFrame->method.getPyrObjType<PyrMethod>();
-                    v_errors::PassingNilToLiteralInitArg::print_error(
-                        m->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
-                        callFrame->method.getPyrObjType<PyrMethod>()->name.getSymbol()->name);
+                    if (isMethod) {
+                        auto m = callFrame->method.getPyrObjType<PyrMethod>();
+                        v_errors::PassingNilToLiteralInitArg::print_error(
+                            m->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
+                            callFrame->method.getPyrObjType<PyrMethod>()->name.getSymbol()->name);
+                    } else {
+                        v_errors::PassingNilToLiteralInitArg::print_error("Function", "value");
+                    }
                 }
                 *(outCallFrameStack + *argIndex) = *argValue;
             }
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -169,12 +169,26 @@ lookup_again:
         if (numKeyArgsPushed > 0)
             applyKeywords();
         g->sp -= numArgsPushed - 1; // Remove all args but the receiver.
-        const auto argIndex = methodRaw->specialIndex; // Zero is index of the first argument.
-        if (argIndex < numArgsPushed && !(g->sp + argIndex)->isNil())
-            slotCopy(g->sp, g->sp + argIndex);
-        else
-            slotCopy(g->sp, &slotRawObject(&method->prototypeFrame)->slots[argIndex]);
-        return;
+
+        const auto index = methodRaw->specialIndex; // Zero is index of the first argument.
+
+        if constexpr (v_errors::PassingNilToLiteralInitArg::do_new()) {
+            if (index < numArgsPushed && !(g->sp + index)->isNil())
+                slotCopy(g->sp, g->sp + index);
+            else
+                slotCopy(g->sp, &slotRawObject(&g->method->prototypeFrame)->slots[index]);
+        } else {
+            auto protoframe = g->method->prototypeFrame.getPyrObjType<PyrObject>();
+            if (index < numArgsPushed) {
+                if (g->sp[index].isNil() && !protoframe->slots[index].isNil())
+                    v_errors::PassingNilToLiteralInitArg::print_error(
+                        g->method->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
+                        g->method->name.getSymbol()->name);
+                slotCopy(g->sp, g->sp + index);
+            } else {
+                slotCopy(g->sp, &slotRawObject(&g->method->prototypeFrame)->slots[index]);
+            }
+        }
     }
 
     case methReturnInstVar: { // Return instance variable.
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -19,6 +19,7 @@
 */
 
 #include "PyrMessage.h"
+#include "PyrKernel.h"
 #include "PyrPrimitiveProto.h"
 #include "PyrInterpreter.h"
 #include "PyrPrimitive.h"
@@ -433,8 +434,12 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
     } else {
         size_t i { 0 };
         for (PyrSlot* it { pushedArgs }; it < (pushedArgs + numNormalArgsAdded); ++it, ++i) {
-            if (it->isNil() && !proto->slots[i].isNil())
-                v_errors::PassingNilToLiteralInitArg::print_error();
+            if (it->isNil() && !proto->slots[i].isNil()) {
+                auto m = callFrame->method.getPyrObjType<PyrMethod>();
+                v_errors::PassingNilToLiteralInitArg::print_error(
+                    m->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
+                    callFrame->method.getPyrObjType<PyrMethod>()->name.getSymbol()->name);
+            }
             outCallFrameStack[i] = *it;
         }
     }
@@ -465,8 +470,12 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
                 else
                     *(outCallFrameStack + *argIndex) = *argValue;
             } else {
-                if (argValue->isNil() && !proto->slots[*argIndex].isNil())
-                    v_errors::PassingNilToLiteralInitArg::print_error();
+                if (argValue->isNil() && !proto->slots[*argIndex].isNil()) {
+                    auto m = callFrame->method.getPyrObjType<PyrMethod>();
+                    v_errors::PassingNilToLiteralInitArg::print_error(
+                        m->ownerclass.getPyrObjType<PyrClass>()->name.getSymbol()->name,
+                        callFrame->method.getPyrObjType<PyrMethod>()->name.getSymbol()->name);
+                }
                 *(outCallFrameStack + *argIndex) = *argValue;
             }
 
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -29,6 +29,7 @@
 #include "PredefinedSymbols.h"
 #include "PyrObjectProto.h"
 #include "SCBase.h"
+#include "VersionedErrors.h"
 
 #include <optional>
 #include <csetjmp>
@@ -425,8 +426,18 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
 
     // Number that will actually be pushed to resulting call frame.
     const auto numNormalArgsAdded = std::min<uint32>(numNormalSuppliedArgs, methNumNormArgs);
-    // Replace arguments if they are not nil.
-    replaceNotNil(outCallFrameStack, pushedArgs, pushedArgs + numNormalArgsAdded);
+
+    if constexpr (v_errors::PassingNilToLiteralInitArg::do_new()) {
+        // Replace arguments if they are not nil.
+        replaceNotNil(outCallFrameStack, pushedArgs, pushedArgs + numNormalArgsAdded);
+    } else {
+        size_t i { 0 };
+        for (PyrSlot* it { pushedArgs }; it < (pushedArgs + numNormalArgsAdded); ++it, ++i) {
+            if (it->isNil() && !proto->slots[i].isNil())
+                v_errors::PassingNilToLiteralInitArg::print_error();
+            outCallFrameStack[i] = *it;
+        }
+    }
 
     if (numNormalSuppliedArgs > methNumNormArgs && methHasVarArg) {
         // Too many args.
@@ -444,13 +455,20 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
 
         // If found in method.
         if (const auto argIndex = findKeywordArgIndex(argKeyword); argIndex.has_value()) {
-            if (argValue->isNil())
-                // Put it back to the default value. Passing nil is the same as saying, reset to default.
-                // It is important to use the proto default here in the case of a previous positional arguments being
-                // present and having overriden the value.
-                *(outCallFrameStack + *argIndex) = proto->slots[*argIndex];
-            else
+            // Found keyword.
+            if constexpr (v_errors::PassingNilToLiteralInitArg::do_new()) {
+                // Passing nil is the same as saying, reset to default.
+                // It is important to use the proto default here in the case of a previous positional arguments
+                // being present and having overriden the value.
+                if (argValue->isNil())
+                    *(outCallFrameStack + *argIndex) = proto->slots[*argIndex];
+                else
+                    *(outCallFrameStack + *argIndex) = *argValue;
+            } else {
+                if (argValue->isNil() && !proto->slots[*argIndex].isNil())
+                    v_errors::PassingNilToLiteralInitArg::print_error();
                 *(outCallFrameStack + *argIndex) = *argValue;
+            }
 
             // SC docs show that in the case of colliding args, the last one added should always be the result.
             if (*argIndex < numNormalArgsAdded) {
--- a/lang/LangSource/PyrMessage.cpp
+++ b/lang/LangSource/PyrMessage.cpp
@@ -23,6 +23,7 @@
 #include "PyrInterpreter.h"
 #include "PyrPrimitive.h"
 #include "PyrListPrim.h"
+#include "PyrSlot.h"
 #include "PyrSymbol.h"
 #include "GC.h"
 #include "PredefinedSymbols.h"
@@ -166,11 +167,11 @@ lookup_again:
         if (numKeyArgsPushed > 0)
             applyKeywords();
         g->sp -= numArgsPushed - 1; // Remove all args but the receiver.
-        const auto classIndex = methodRaw->specialIndex; // Zero is index of the first argument.
-        if (classIndex < numArgsPushed)
-            slotCopy(g->sp, g->sp + classIndex);
+        const auto argIndex = methodRaw->specialIndex; // Zero is index of the first argument.
+        if (argIndex < numArgsPushed && !(g->sp + argIndex)->isNil())
+            slotCopy(g->sp, g->sp + argIndex);
         else
-            slotCopy(g->sp, &slotRawObject(&method->prototypeFrame)->slots[classIndex]);
+            slotCopy(g->sp, &slotRawObject(&method->prototypeFrame)->slots[argIndex]);
         return;
     }
 
@@ -410,9 +411,22 @@ void prepareArgsForExecute(VMGlobals* g, PyrBlock* block, PyrFrame* callFrame, s
         return std::nullopt;
     };
 
+    // Put defaults on the stack first.
+    std::copy(proto->slots, proto->slots + methNumNormArgs, outCallFrameStack);
+
+    // Override defaults with envir, if provided.
+    if (optionalEnvirLookup)
+        // ident dicts can't have nil inside them, so no need to check the default.
+        for (auto i = 0; i < methNumNormArgs; ++i) {
+            PyrSlot keyslot;
+            SetSymbol(&keyslot, methArgNames[i]);
+            identDict_lookupNonNil(optionalEnvirLookup, &keyslot, calcHash(&keyslot), &outCallFrameStack[i]);
+        }
+
     // Number that will actually be pushed to resulting call frame.
     const auto numNormalArgsAdded = std::min<uint32>(numNormalSuppliedArgs, methNumNormArgs);
-    std::copy(pushedArgs, pushedArgs + numNormalArgsAdded, outCallFrameStack);
+    // Replace arguments if they are not nil.
+    replaceNotNil(outCallFrameStack, pushedArgs, pushedArgs + numNormalArgsAdded);
 
     if (numNormalSuppliedArgs > methNumNormArgs && methHasVarArg) {
         // Too many args.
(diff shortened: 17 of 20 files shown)


telephon:
No, no, wait, we really may want to keep the possibility to explicitly pass nil in order to override the default argument.

The semantic of default arguments is not "if the argument passed is nil", but "if no argument was passed". It is quite risky to change this.

The special syntax `{ |x (2)| ... }` as opposed to `{ |x=2| ... }` was introduced to allow this behaviour when you need it. Already at that time, one could have instead changed the way it works, but this was not done, and I suppose it was for the reason above.

JordanHendersonMusic:
I very much disagree with this. 
Here is the existing behaviour.

```supercollider
{ | a = 2| a }.(nil) // nil
{ arg a = 2; a }.(nil) // nil
{ | a 2 | a }.(nil) // nil

{ | a (2) | a }.(nil) // 2
{ arg a = 1 + 1; a }.(nil) // 2
{ |a = (1 + 1)| a }.(nil) // 2
{ | a = (2)| a }.(nil) // 2
{ arg a = (2); a }.(nil) // 2
``` 

As you can see, the current behaviour is as follows:

Passing `nil` means reset the argument to its default, in all cases ***expect*** when argument is a literal and the syntax exactly matches one of three above.

Having things behave differently depending on whether the value is a literal or an expression is not good as this is an internal details and quite complicated to deduce.

For example `2` is a literal, but `(2)` is not.

This also heavily impacts any future constant propagation. 


It should either reset to default, or overide the default to nil, in ***all cases***. As this is baked into the bytecode, I do not think this was ever intended, and changing this to work with non-literals would be a very large breaking change.

> The special syntax { |x (2)| ... } as opposed to { |x=2| ... } was introduced to allow this behaviour when you need it.

This doesn't appear to be correct given the example above.

JordanHendersonMusic:
*Sorry for all the edits...!!!*

Also note that we don't distinguish between
```
f.(nil)
```
and 
```
f.( g.() )
```
Where `g` might return `nil`.

This means the call-sight doesn't know if the default value will ever be used to not, as it depends on the syntax used to declare the default. 
Hence the following two statements are not so readily distinguishable and plays with the ambiguity of whether nil is no value or the absence of one...
> The semantic of default arguments is not "if the argument passed is nil", but "if no argument was passed". It is quite risky to change this.

From the perspective of the call sight I think this behaviour makes no sense at all. When you set a default, you are saying, 'when no value is passed, make it this value', since `nil` is not a value, its the same as saying 'when this is `nil`, make it this value'.

From the perspective of the function body one of the key benefits of declaring default arguments is that they cannot be `nil`. This is because when the callsight doesn't declare an argument, you get `nil`. If defaulted arguments can be `nil`, and as `nil` is essentially useless (because it isn't a value), you have to check it again before you use it.
```
{ |a 1|
a = a ?? { 'a second default' };
``` 


Identity dictionary also provides some issues here. `a[\key] = nil` is the same as deleting the value. Here, `nil` means the absence of a value.

This becomes problematic when we consider `peformEnvir` as again, `nil` and the absence of a value are indistinguishable.

jamshark70:
Hm. I see the point, and sometimes you do have to break stuff, but *guaranteed* somebody somewhere will find somewhere in their code that used to work for them in the past, e.g. where `nil` was used as a sentinel, no longer works, and because of the nature of this change, it will be difficult and confusing to find it. (Agreed that `nil` is a poor choice for a sentinel, but "well you shouldn't have" is not what users are going to want to hear.)

And it's not feasible to ship an automated tool to find potential defaults whose behavior might change, because most argument defaults are literals.

I don't necessarily disagree with the change; I just wish there were a way to soften the blow for users. I think the probability is quite high that the user impact has been underestimated here.

dyfer:
> I think the probability is quite high that the user impact has been underestimated here.

I agree with this! We need to be careful with a change like this, since it WILL change users' code behavior silently, even if the original behavior is highly undesirable. Similarly to @jamshark70 I'm not strictly opposed to the change, but I think we need to discuss it more.

JordanHendersonMusic:
> And it's not feasible to ship an automated tool to find potential defaults whose behavior might change, because most argument defaults are literals.

Actually it's possible to generate a warning when this happens... But the check has to happen at **every** function call, for every argument. That's a major performance hit! This could be toggled on and off though. 

I will have a look and see if I can do the warning to error thing here. It will have performance implications though.

My main concern is that many basic compiler optimisations that we probably want to do in the future, e.g. constant folding ( turning `1+1` into `2`) changes an expression into at literal... Which changes what `nil` means at the call sight - that's just weird.

Whether or not something is a literal should have no impact on the behaviour of the code.

JordanHendersonMusic:
I've updated this to post a warning in 3.15 and do the new behaviour in 3.16

```
sc3> f = { arg a1 = 1; a1 };
-> a Function
sc3> f.(nil)
ERROR: in version 3.16 and onwards, passing nil to a default initialised argument will have different behaviour.
See **** for more information and for how to update your code.
-> nil
sc3> 
```

I need to get the clickable links in the post window working then the `***` will show a help document link telling the user how to fix this.

One issue, this will print every single time a message is passed. This can easily spam the post window, but I don't see a way around this because you want to post it for every unique occurrence in the source code, but that information isn't available here.

JordanHendersonMusic:
Okay so the problem with the errors/warnings is that we can't tell when the user is aware of this and has done it properly.

This occurs in Thread.new 
```
Thread *new { arg func, stackSize = (512); }
```
Here, `stackSize` is set in the prototype frame, and the nil check byte code is emitted.

I wonder how many people know that the correct way to get an initialised argument to not be nil is to wrap the literal in brackets?

---

Another place the warning is posted is in UnitTestResult.
Here message is initalised as `arg message = ""`, however the code does not perform any `nil` checks, I imagine the author assumed it would never be `nil`.

However this is okay, because later on it checks `message.size > 0`, and `nil.size == 0`. 

I think this works more by coincidence than design.

JordanHendersonMusic:
So the cases that I have changed in the class library are mostly cases where we do ***not*** want to pass nil, i.e., where this feature is not wanted. The vast majority of these were added by @telephon to fix bugs caused by this feature.

To provide evidence that function authors expect nil when there is a default argument, I think you'd want to find some code like....

```supercollider
{ |a = 10|
    if (a.isNil) {
        ...
    }
}
```
Or some other way that both the default is set and `nil` is handled.

Edit: I've found a couple of these, but they appear to be very simple fixes.

---

The places where I have found this behaviour used are in UGen:linexp and linecurve, NodeProxy:clear, and Buffer:getToFloatArray.

In UGen, this was using `nil` to mean, don't clip, provided a default clip of minmax. This is a simple fix done by passing `\none` as as clip value. Ideally, the default would have been removed, but that would be a breaking change.

NodeProxy:clear's fadeTime is a little more complicated.

Buffer:getToFloatArray was using `-1` as a sentinel and performing a `nil` check, this is completely redundant.

SCClassLibrary/JITLib/ProxySpace/NodeProxy.sc
JordanHendersonMusic, line 68: @telephon is this the same as before? I'm a bit confused because the actually warning is generated in `clear` for the argument `fadeTime`, but all the places it is passed to are set to `this.fadeTime`. I was wondering if you could take a look?

telephon:
Sorry, this is a lot. I don't get the point. The semantics is clear and works as expected:

```
f = { |x (4)| x }; f.(nil); // should be 4
f = { |x=4| x }; f.(nil); // should be nil
```

First we should know why we want to change such a basic functionality. The potential for subtle errors is massive, and so far I don't see the gain.

JordanHendersonMusic:
Yes this has become a lot, sorry about that, honestly I find this behaviour so surprising I assumed it was a simple bug.

> First we should know why we want to change such a basic functionality. 

Tldr: I think it's useless and confusing.

Because it means all code that uses `x = 4` must check for `nil`. This happens very rarely in the class library, meaning all the other code is wrong - that's a lot of incorrect code. It also means all my code that uses default args is wrong.

It means the following doesn't work.
```
f = { |a=2, b| ... };
g = { |a| f.(a, 10) };
```
How do you get `f`'s `a` to default from inside `g`? Note that I've placed `a` at the beginning!

I also haven't seen a use case where you actually benefit from having a defaulted argument be `nil`. The one case where this was obvious in the class library is UGen linexp, but you can trivially replace `nil` with `\none`. The only place this would have used is when the type of the argument can't contain a sentinel value, but in that case, it is unlikely you'd want a default argument.

I've never seen a language do this - would love to know if there are any. Lua seems the closest to our behaviour.

It is a unique meaning assigned to `nil`, as it usually means the absence of something. If a function returns either a value or `nil` and you want to pass this along, you need to call the function in a different way depending on whether you want the default or not.
```
f.() !? { |a| g.(a) } ?? { g.() }
```

The semantics are not clear `x = (2)` is not the same as `x = 2`. Yet in all other cases `2 === (2)`. Changing this, which is called constant propagation, is something we want to do as it has huge performance advantages, but as it stands, this will be a breaking change.  

It also makes `x = ""` irrelevant as you now must handle `nil` as well as `""`.. thing is, the default default value is `nil`, so you might as well have not specified `""`.

telephon:
Yes, I understand the reasoning, and I understand that for some applications, it is quite annoying. 
Consider that `nil` is an object, and sometimes it represents a particular behaviour.

Changing it would make it impossible to do this:
```
f = { |id, freq=440, latency=0.01|
     s.sendBundle(latency, "/n_set", id, freq)
};

f.(1001, 500, latency:nil); // where nil latency means send without time stamp.
```
For this reason, latency 0 is not the same as latency nil.

Maybe there is another solution how to pass on arguments from one function to another and distinguish between:

- nil args, and 
- unsupplied args.

In any case, it would make a massive breaking change, so we'll need to discuss this in a meeting.

JordanHendersonMusic:
> In any case, it would make a massive breaking change, so we'll need to discuss this in a meeting.

Sure!

> Consider that nil is an object

I disagree, it might inherit from `Object`, but it means the lack of a value. If you don't set a default and don't pass an argument, you get `nil`, i.e., the absence of a value.

---

So I think the example there is bad design. Consider this 
```
f.(10001, 500, latencys[\foo])
``` 
If `\foo` doesn't exist you get completely different behaviour. Hence why I don't think this has a use because the function could do one of three things:
* either omit the default arg — the default default is `nil`. 
* have another argument called `omitTimestamp` and check the latency is zero or less.
* use -1 to mean no timestamp.

---

Although I disagree with its usefulness and intended use, the real issue is three fold: it is so easy to do this and that is surprising because no other language does this, it is the default throughout the class library although the presence of `nil` is very rarely handled, and I cannot find any documentation on it (consider FunctionDef's). 

This is why I think removing it is a good idea, because I think the majority of places where it has been used are bugs.

However, if we *really* want to keep this behaviour, might I suggest a change in syntax to something far more explicit and impossible to do by mistake.

How about?

```
{ | a ?= 1|
   ...
}
```

This aligns with other languages where the question mark means an optional value.

dyfer:
I understand that certain aspects of SC may be bad design, but this particular change will likely wreak havoc. Julian's example is likely a relatively common one - overriding the default latency with no latency for immediate instantiation. I definitely used nil to re-set an arg to nil in my code (I'm happy to concede it's bad design - but it's one that's certainly in use) - certainly not as a bug.

If we want a cleaner interface, I think we may need to consider how provide the ability to revert argument to a default value, and how to override it with no value (nil), using different objects or keywords

FWIW in my mind the bug is the other way around - that nil does _not_ override non-literals - I'd expect it to do so...

One path forward would be to deprecate passing nil as arg and accept either only uppercase `Nil` (to set any kind of arg into `nil`) or a new object like `Default` (?) to reset to value to default when requested.

JordanHendersonMusic:
> FWIW in my mind the bug is the other way around - that nil does not override non-literals - I'd expect it to do so...

I don't think this is possible. This is currently implementated by overwriting the value in the call frame directly. If you have an expression (or any non-constant thing), it can't go in the slot at compile time.

Therefore the expression one is implement as `a = a ?? {default}`. This is exactly how you do it in lua.

I don't think it is possible to do what you've suggested without creating a second type of `nil` with it's own semantics. What happens when you put that thing in an identity dictionary?

Hence why I thought this was a bug.

> One path forward would be to deprecate passing nil as arg and accept either only uppercase Nil (to set any kind of arg into nil) or a new object like Default (?) to reset to value to default when requested.

I think this will get messy and won't really solve the problem. I think the explicit opt-in syntax would be the best way `a ?= 2`.

> Julian's example is likely a relatively common one - overriding the default latency with no latency for immediate instantiation. 

You could just as easily use `-1`, or `\noTimestamp`, or literally anything that isn't `nil` or a greater than zero number.

dyfer:
> > FWIW in my mind the bug is the other way around - that nil does not override non-literals - I'd expect it to do so...
> 
> I don't think this is possible. 

Good to know. 

> I don't think it is possible to do what you've suggested without creating a second type of `nil` with it's own semantics. What happens when you put that thing in an identity dictionary?

Yeah, probably not feasible.

> I think this will get messy and won't really solve the problem. 

Yeah, I can see that.


> > Julian's example is likely a relatively common one - overriding the default latency with no latency for immediate instantiation.
> 
> You could just as easily use `-1`, or `\noTimestamp`, or literally anything that isn't `nil` or a greater than zero number.

Got it, yes, this indeed works in the case of bundle latency. Good point for further considerations. Thanks!

jamshark70:
> You could just as easily use -1, or \noTimestamp, or literally anything that isn't nil or a greater than zero number.

I can confirm that symbols would be a good workaround (using Pd to sniff the raw bytes):

```
s.makeBundle(\noTimestamp, { x = Synth(\default) });
print: 0 0 0 0 35 98 117 110
print: 100 108 101 0 0 0 0 0
print: 0 0 0 1 0 0 0 36
print: 0 0 0 9 44 115 105 105
print: 105 105 0 0 100 101 102 97
print: 117 108 116 0 0 0 3 232
print: 0 0 0 0 0 0 0 1

s.makeBundle(nil, { x = Synth(\default) });
print: 0 0 0 0 35 98 117 110
print: 100 108 101 0 0 0 0 0
print: 0 0 0 1 0 0 0 36
print: 0 0 0 9 44 115 105 105
print: 105 105 0 0 100 101 102 97
print: 117 108 116 0 0 0 3 233
print: 0 0 0 0 0 0 0 1
```

The packets are identical, so latency `nil` and latency `\noTimestamp` behave equivalently.

However negative latency will calculate a timestamp in the past, and thus produce "late" warnings.

~~

Are there any conspicuous failures in the test suite when run with this change? (Absence of failures wouldn't be proof that the change is safe, of course, but it might suggest that the change is less dangerous than we might suppose. Meanwhile, failures would give us some indication where to look.)

SCClassLibrary/Common/Audio/UGen.sc
```diff
@@ -64,17 +64,17 @@ UGen : AbstractFunction {
 	}
 	exprange { arg lo = 0.01, hi = 1.0;
 		^if (this.signalRange == \bipolar) {
-			this.linexp(-1, 1, lo, hi, nil)
```
JordanHendersonMusic, line 67: @jamshark70 here was one fix - as you can see it is very simple.

SCClassLibrary/Common/Control/Buffer.sc
JordanHendersonMusic, line 314: @jamshark70 this one wasn't a failure, but useless, since it checked for nil anyway, defaulting to -1 was pointless.

JordanHendersonMusic:
I actually didn't test the sendBundle thing, I just meant from the perspective of the function any other value would work and you could check for it, but it's good that symbols work! 

That's my point about calling this useless, you don't need it, if the functions actually does accept every possible value in the class library, make a new class for your sentinel! But it's more likely -1, 0, \none, or something like that will do just fine.

Reserving `nil` to mean the absence of a value is important, or we will need to create another nil type for when we want the absence of `nil`.

Having an absence of latency is the same as zero latency, it's just that send bundle does something special in this case. Really it should be a separate argument, it even a different function to send without a timestamp... but a symbol seems quite clear for backwards compatibility/updating with ease.

telephon:
Yes, there are indeed cases where already now `nil` can't stand for nothing properly speaking. In demand ugens, for example, it is `NaN` (logical, because we are in the domain of floats), but in patterns "nothing" is ambivalent (it normally means stop yielding, but you may want to override empty values further up until new ones come in). A boundary case is `Rest` (subclass of Operand).

The represnetation of nothing has been a long standing issue. Once we have AbstractObject, we may be able to solve this better.

JordanHendersonMusic:
> In demand ugens,

A great example of where this is confusing!

```supercollider
~make_dseq = { |list, repeat| Dseq(list, repeat) };

SynthDef(\test, {
	~make_dseq.([1, 2])
})

// ERROR: Dseq arg: '0' has bad input: nil

ListDUGen : DUGen {
	*new { arg list, repeats = 1;
		^this.multiNewList(['demand', repeats] ++ list)
	}
}
```

I would expect it to default to a repeat of `1`, not `nil`.


My point isn't that passing `nil` is bad. 

For example I think these all make sense.
```
1 + nil == 1;
2 * nil == 2;
```
This is because promoting `nil` into the identity of the function means to not perform the function.
```
+ { |other = 0| ... }
* { |other = 1| ... }
```

The issue is specifically when `nil` is used to mean something other than what is provided in the default argument.  If no default is given (which is impossible because the default default is `nil`), then obviously passing `nil` is fine and is equivalent to omitting the argument at the call sight.

The `Dseq` case is interesting because there is no reason why `nil` can't be turned into a repeat of `1` — it is invalid, and SuperCollider often tries to avoid errors! In fact, I think that is a bug and should be fixed! Of course the fix is to use `repeats = (1)`... hence why I think the changed proposed in this PR will fix more code than it breaks as there are very few correct usages of `argument = X`.

> Once we have AbstractObject

I don't see how that is relevant.

Spacechild1:
This discussion is really interesting. I think the problem boils down to the combination of two language features:
1. function/method arguments can have default values
2. missing arguments default to `nil`

There is another language that has both features: JavaScript!

```js
function foo(a = 1, b = 2, c) {
    console.log(a, b, c)
}

foo(3, 4) // -> 1, 2, undefined
foo(3) // -> 3, 2, undefined
foo() // -> 1, 2, undefined
foo(undefined, undefined) // -> 1, 2, undefined
```

Note how explicitly passing `undefined` yields the default arguments! That's exactly the behavior that @JordanHendersonMusic (and I, for that matter) would expect. I'm not saying that JS is the pinnacle of great language design, but this is something they definitely got right.

I completely agree with Jordan that the current behavior of sclang is just wrong from a language design perspective, for all the reasons he has given. The fact that the behavior actually depends on how the default argument is being declared (literal vs expression) makes it even worse!

To sum it up: in the context of a function, `nil` means the absence of an argument and it should always be treated as such. This means that passing `nil` as an argument should have the same effect as omitting that argument!

However, I agree with the others that we must be careful here because there is certainly quite a bit of code out there that relies on the current behavior. This is a tricky situation...

JordanHendersonMusic:
@Spacechild1 thanks for the summary, it is really useful to have fresh input on a complex issue such as this! 

Regards how painful this might be.... 

I definitely underestimated this because I thought it was a bug at first!!!

Since then I've added the warning to error message that gets triggered when the user passes nil to a function with a default. This obviously occurs at run time, rather than compile time. 

Currently, there is no caching, so it can spam the post window - I think I'll change this. 

Also, there are some false negative, when the user had used the brackets. This is harder to change, possible, but means making the compiler store extra information in the function definition object. This might be useful in other situations though so I think I'll do it.

Additionally, I want to make a nice bit of documentation accessible through a clickable link in the post window.

On issue with this approach this that, in the case of using this incorrectly, the users code is already broken, but in 3.16 it will be fixed without them doing anything. That means what ever fix they implement is only temporarily needed and their old code will become correct again.

If any one has any other ideas about how to mitigate the pain of this, I'm all ears!

Finally I think it's worth considering the places where this fixes bugs / improves behaviour - this change brings both pain and pleasure! Given the novelty of this, I believe there are more incorrect usage that correct, so there should be more fixes, but these will be silent and users won't notice them.

Perhaps the way forward is for me to implement all the warnings and errors properly, and for people with large old code bases to try it, and then have a chat about how painful this was in a developer meeting? We might even consider merging it just to see what happens? Unfortunately I don't have such a large collection, and I know I've never used this on purpose.

jamshark70:
> Having an absence of latency is the same as zero latency

Strictly speaking, not exactly. 0 latency means that the server should push the sound out of the speakers at the moment defined by the language's logical time + 0 sec. Because network transmission time is always non-zero, this is always late. Nil latency means to execute the message at the next opportunity, without complaining about it.

What happens in the case of symbols is that `slotDoubleVal()` fails (in `prNetAddr_SendBundle()`, so the "logical time + latency" calculation is skipped, and the bundle is sent with the special "immediate" value defined in the OSC standard. I.e., any value for latency that can't be cast to a double will be treated as "immediate."

Getting afield of the main topic -- anyway. I find myself in the middle. I've also wondered at times why literal defaults are overridden by nil -- seemed odd to me. But changing it will take extensive testing and careful review of the class library. This has been started, but I'm not convinced that it's been completed yet (though it certainly could be finished, at some point).

lnihlen:
The problem from where I see it is that `nil` has been doing two jobs here as far as communicating user intent:

 - to specify the literal `nil` value
 - to request the callee to provide a default value for this (likely positional) argument

We don't have a way at either compile or runtime to disabiguate these two intentions.

I think that changing the calling semantics, even as "uneven" as they are with expression initializers vs. literal initializers, amounts to essentially an interpreter "ABI" change which I expect will result in wide-ranging negative user impact. The same code as written will now work differently depending only on sclang version. This seems like a bad situation to me, and I don't think this is a good change for the project.

I can think of two other ways that we can fix this that don't break existing code:

1. Deprecate expression initalizers. The users loses nothing in terms of code functionality, it's always possible to specify that same `if arg ?? { initialize };` block at the top of the method. This also cleans up some of the weirdness around referencing local variables in argument initalizing code, etc, so arguably gets us a bit closer to inline variable declarations. Furthermore, this means no behaviour/ABI changes, and we can warn that these are going away, then error on them in a subsequent version. That solves the "bug" filed about this quirk of SC language design. The big plus here is that the refactor still works as expected on older code, so no loss of compatibility.

2. Disambiguate the two intentions. What if we created a new sentinel value, maybe `default` or we could perhaps use the curry symbol `_` to indicate the default value should be used (hot take, needs more thought), or even the double-curry `__` for extra spice. Then `nil` always means literal `nil`, we don't change the ABI, just the AST.

This is the product of a half hour of thought on my part. I imagine we could come up with even a few more alternatives together as a group at the dev meeting. Both of these resolve the inconsistency without introducing a behaviour change to existing code, and I think should be explored before we pursue this alternative further.

JordanHendersonMusic:
> The same code as written will now work differently depending only on sclang version

Yes this is something we've agreed upon in dev meetings before. We have a version where a warning is posted the behaviour is going to change, then in the next version, it does. This means assuming the user runs their code once every 18 months, they have an easy obvious upgrade path. If they have no interest in this small act of maintenance, they shouldn't upgrade - many users already don't! For these small language changes (only one use in the test suite of this issue) this seems pragmatic.

Re the initialiser ordering, we agreed last meeting on the other PR that redefines this. I just need to make the error messages nicer. Busy this week might have time this weekend maybe.

As I see, this feature is used in two places: explicitly, where you just replace it with another value; or implicitly, where I think the majority of uses are and that these are bugs.
```
f = { |a = 1, b| ... }
f.(nil); \\ explicit 
g = {|a, b| f.(a, b) }
g.() \\ implicit, a should be 1.
```
The `default` or `_` suggestion doesn't fix the implict use. You still have to do a nil check in `g`. ... Unless we made `default` an object not a keyword and defaulted `g`'s args to `default` - but that brings with it a lot of complexity as it's a new literal type.

This is exactly what happened in the Dseq example, the arguments was defined `a = 1`, but `nil` has no meaning in the function body and it produces an error. Supercollider tries to avoid errors when there is a reasonable alternative, which is what the default value is for.

That is why I think this is worth fixing, it only occurs correctly in a very small number to places, is easy to change where it does, and fixes many bugs.

JordanHendersonMusic:
Briefly re the `sendBundle` example,
```supercollider
	sendBundle { arg time ... args;
		_NetAddr_SendBundle
		^this.primitiveFailed;
	}
```
It has no default arguments so you can still do `s.sendBundle(nil)`. The only issue occurs when you wrap it and provide a default.

```supercollider
f = { |time = 0.01 ...a|
   s.sendBundle(time, *a)
}
```
In this case you will never get the lack of a timestamp. But this is an easy fix, either remove the default argument, or do a check for a special value.
```supercollider
// time: number or \noTimestamp 
f = { |time = 0.01 ...a|
   s.sendBundle( if (time == \noTimestamp) { nil } { time }, *a)
}
```
Alternatively, this would be the rare place where you might want to use the syntax I proposed earlier saying that `time` will be `0.01` or `nil`. But again, I think this is confusing and should be avoided.
```supercollider
// time: number or explicit nil
f = { |time ?= 0.01 ...a|
   s.sendBundle( time, *a)
}
```


---

@jamshark70 
> This has been started, but I'm not convinced that it's been completed yet (though it certainly could be finished, at some point).

Once I've:
* Got the lexer pr merged
* Redone the lexer/bison code and refactored pyrparsenodes (already done as a draft in the error pr)
* Added line number information to each bytecode (done in the error pr)

We can have a separate build of sclang with code coverage. This will let us see what parts of the class library we aren't testing. Not fool-proof, but a marked improvement over the current situation.

telephon:
> > The same code as written will now work differently depending only on sclang version
> 
> Yes this is something we've agreed upon in dev meetings before. We have a version where a warning is posted the behaviour is going to change, then in the next version, it does.

We are working out a document that makes explicit of how to deal with method interface changes. As far as I can remember, we didn't agree that versions will contain deep breaking changes, but that the next version will allow itself a little more than our usual very strict procedure.

But maybe I wasn't there when this was discussed.

telephon:
> In this case you will never get the lack of a timestamp. But this is an easy fix, either remove the default argument, or do a check for a special value.

Of course, you are completely right that code can be changed. But I am pretty sure it will break heaps of libraries people have written some time and rarely use. Once they use them, the result will be a subtle bug. 

Don't get me wrong, I find the reasoning convincing, but the strategy of solving it is too invasive as far as I am concerned. Maybe @lnihlen 's idea is on the right track?

JordanHendersonMusic:
> Once they use them

In the next version they would get an error like...
```
Error: passing nil to an argument with a default will change behaviour in the next version. 
This has occured in 'file://blah.sc:34:20' while calling the method 'scdoc://Foo:bar'.
For more information on how to fix this, see: 'scdoc://errors.schelp/#argnil'.
```

Where all those URLs are clickable links that take to straight to the source code or to the help docs.

There's nothing subtle about that!

My issue with having a default keyword is that it forces this confusing, useless behaviour on everyone, whereas right now, it only occurs as an edge case. This doesn't improve the language, I think it makes it - explicit sure - but overall worse.

jamshark70:
> Perhaps the way forward is for me to implement all the warnings and errors properly, and for people with large old code bases to try it, and then have a chat about how painful this was in a developer meeting?

Hey, I'm willing to kick the tires on that (though "old" code bases includes, like, last week :wink: ).

Not difficult, at a non-critical time, to build with this PR and improv with my system a bit. That spans a *lot* of my code.

lnihlen:
> > Once they use them
> 
> In the next version they would get an error like...
> 
> ```
> Error: passing nil to an argument with a default will change behaviour in the next version. 
> This has occured in 'file://blah.sc:34:20' while calling the method 'scdoc://Foo:bar'.
> For more information on how to fix this, see: 'scdoc://errors.schelp/#argnil'.
> ```
> 
> Where all those URLs are clickable links that take to straight to the source code or to the help docs.
> 
> There's nothing subtle about that!

Sure, but on the next version the error is removed. So users new to SC, trying to use a 12 year old (or older) Quark (directory.txt is full of them) will now have broken library code they didn't even write, and no diagnostics telling them where they went wrong.

So I maintain my assertion that the likelihood that this behaviour change is going to ruin some user's day is very high, particularly in the context of an ecosystem of widely-shared 20+ year old code. 

> 
> My issue with having a default keyword is that it forces this confusing, useless behaviour 

This seems like a subjective value judgement. Not saying its wrong, just pointing out that your opinion on this behaviour is not universally held, so others may weigh the benefits of this change against potential user disruption differently.

I'm all for behaviour changes and moving the language forward. I want us to find ways to do that where old code fails predictably, however. Compilation error messages are a great way to draw user attention to interpreter changes. 

I never thought Hadron, which has no users and therefore no expectations, would end up being the more conservative of the two projects. This seems odd to me, like a strange inversion of priorities between the two projects.

JordanHendersonMusic:
>  Sure, but on the next version the error is removed. So users new to SC, trying to use a 12 year old (or older) Quark (directory.txt is full of them) will now have broken library code they didn't even write, and no diagnostics telling them where they went wrong.

So I really hope @capital-G's package manager will let us remove these unmaintained quarks from general use. In my mind, the problem is using unmaintained code - I know this happens a lot, but it's a problem that needs a solution.


This thinking out loud a little here...

While I'm against versions because it will introduce a large maintenance burden...

Because this behaviour is quite contained we could introduce a compatibility mode, and store this on the function def.

```
Foo {
   ##legacyCallingConvention


   bar { ... }
}
```
Now all methods in Foo get this legacy flag and when we call one, we use the old behaviour.

Now there are complexities with this, because functions created inside the methods would need the flag too, this happens at runtime. However you can return these functions outside of this legacy context, which would be confusing.

Perhaps it is enough to limit the legacy behaviour to just the method itself? This should substantially limit negative effects, although not completely.

Note this still requires the class be marked as legacy, is this really that much harder that fixing the error message? Particularly since it's use is quite infrequent.

I still think that asking all quark author to check their code works once every 18-24 months is reasonable. Not fixing this bad default because someone used it 20 years ago is paid in an increase in difficulty faced by new users - without whom, SC slowly disappears. 

---


>Not difficult, at a non-critical time, to build with this PR and improv with my system a bit. That spans a lot of my code

Thanks, that's would be really useful! But maybe not until I've fixed up the error message!

lnihlen:
I was just re-iterating my arguments from before as I don't feel they've been heard. Removing that post.

There are very few people who understand the interpreter internals well enough to make these kind of substantial changes. SuperCollider remains a do-acracy, however, where those with the time/energy/capability to make the changes are generally given much latitude to do so.

Since I'm not an active SuperCollider developer I will leave it here. I've raised my objections and concerns. I am an SC _language_ developer, however, and I don't think this is a good change for the language. Of course Hadron needs compatibility with sclang in order to be useful. But I will not enjoy the day that I implement this change in Hadron. Particularly because it will result in additional function call overhead, and I already have a considerable performance gap to close.

prko:
>After this PR, passing nil always resets the argument to its default value.

What should users do in cases where they genuinely need to pass nil as an actual value?

At the moment, I don’t see a documented or recommended way to express “I want this argument to be nil”, as opposed to “I want to reset this argument to its default”.
Has this been discussed, or is there a proposed alternative syntax or sentinel value?

JordanHendersonMusic:
> There are very few people who understand the interpreter internals well enough to make these kind of substantial changes. 

@lnihlen 

I seem to recall you raised a similar objection to the keyword arguments change on the forum a few years ago!

I don't think this is because the code is difficult.

Since SC was written almost 30 years ago, most of this has become textbook. Crafting interpreters is directly applicable to almost all of this code. Exceptions being the GC, where we do a little bit often to have consistent latency, and heap allocated call frames to make continuations easier. This is is old well established technology.

What makes this code hard is that it's undocumented, is in need of modernising, and a general good old tidy. Great example being the lexer where there are three global variables for the filename all subtlety different, and the global state is accessed, mutated, *and defined* across many files, including the parse node constructors.

I'm actively trying to fix this with other PRs: to make this part of the code more accessible, and to encourage people to dig into this side of the project. I don't think it's healthy in the long run, nor in the spirit of the project, to label everything in the `lang` dir *'keep out: here be dragons'*.

I also want to improve testing because changes in this part of the code base do have far reaching consequences. One thing I'm planning is code coverage in sclang, so our unit tests can actually cover the class library - this will probably be a separate build though. I also really like your idea of a corpus of code!

Re performance: I don't know what hardon is doing, but for us, this change (minus the error posting) *should* actually give a performance boost as we will only need to do `a ?? d` when there is an expression. The nil check-then-copy can even be vectorised for larger numbers of arguments. Sure we have to do this at every call, but it's in c++ not in the interpreter. I'll have to benchmark it to find out.

---

I think I need to get some numbers on how often this feature is actually used *intentionally*. Before then, everything is a little hypothetical. If this turns out to be a heavily used feature, then of course I'll close this pr - I've spent significantly more time responding here than I have writing code! 


---

@prko

> What should users do in cases where they genuinely need to pass nil as an actual value?

Nil isn't a value, it's the absence of one. `(a: 1)[\a] = nil` deletes the entry.

Nil is also the *default* default argument, so the presence of a default argument means we are not expecting this to be nil.

You can always pass nil if there isn't a default argument - either explicitly, or by omission.

This pr makes it so you can't pass nil if there is a default argument, it just turns into the default. Which already happened with all but a few very specific (but common) syntaxes. It also aligns SC with other languages, namely js.

prko:
Thank you for your kind and detailed explanation.

>This pr makes it so you can't pass nil if there is a default argument, it just turns into the default. 

I believe there may be cases where a user intentionally provides a default value, but occasionally needs to pass `nil` explicitly.

For example:
```
(
f = { |x = 1|
    x
    !? { Pbind(\degree, Pseq((1..8), x), \dur, 0.08).play }
        ?? { "no music".postln }
    }
)
f.()      // default: plays once
f.(2)     // plays twice
f.(nil)   // plays nothing
```

And without a default argument:
```
(
g = { |x|
    x
    !? { Pbind(\degree, Pseq((1..8), x), \dur, 0.08).play }
        ?? { "no music".postln }
    }
)
g.(1)   // plays once
g.(2)   // plays twice
g.()    // default: plays nothing
```
In such cases, the ability to pass `nil` deliberately can be meaningful, as it allows the user to distinguish between “use the default” and “treat this as an explicit nil”.

JordanHendersonMusic:
```
f.(0) // plays nothing
```

`nil` just means zero here, better to be explicit. 

No need for the nil that way because thats the default behaviour of `Pseq`. Hence my argument that this is useless.

If we really need this behaviour, I've proposed the special syntax ` x ?= 1`, but I don't think it actually has a use beyond making updating old code easy. There is always a more explicit way to achieve what is wanted.

These points have already been addressed above, there is a lot of text, but it is all there.

lnihlen:
> > There are very few people who understand the interpreter internals well enough to make these kind of substantial changes.
> 
> @lnihlen
> 
> I seem to recall you raised a similar objection to the keyword arguments change on the forum a few years ago!

That's not an objection so much as an appeal to your sense of fairness and justice. As language developers we are in a position of power in this community. I applaud your efforts to reduce the technical debt and obfuscation in the code base. That doesn't change the situation today.

Because of that power, I feel the language devs have a responsibility to put their subjectivities aside as best they can and consider the overall community. You don't like current behaviour, and you're willing to risk potential serious disruption to the user community to satisfy your opinion. I'm trying to appeal to your ethical sensibilities to ask you to consider paths that may be less harmful.

There are many language design decisions in SC that I think are terrible. I have faithfully implemented them in Hadron. I'd love for there to be a process to evolve the langauge but we have to consider inclusion of our user base in every move.

That's why I've raised this same argument before, because this is a pattern of behaviour for both of us that highlights a difference in underlying values. I think the SC user community is the most valuable asset of SC. There are musicians and artists who have learned just enough SClang to make their art. Lacking a CS or engineering background, that learning has likely been painful. Why would we want to risk a subtle break their code, and thus break their art, and offer them no recourse?

> 
> I don't think this is because the code is difficult.
> 
> Since SC was written almost 30 years ago, most of this has become textbook. Crafting interpreters is directly applicable to almost all of this code. Exceptions being the GC, where we do a little bit often to have consistent latency, and heap allocated call frames to make continuations easier. This is is old well established technology.
> 
> What makes this code hard is that it's undocumented, is in need of modernising, and a general good old tidy. Great example being the lexer where there are three global variables for the filename all subtlety different, and the global state is accessed, mutated, _and defined_ across many files, including the parse node constructors.
> 
> I'm actively trying to fix this with other PRs: to make this part of the code more accessible, and to encourage people to dig into this side of the project. I don't think it's healthy in the long run, nor in the spirit of the project, to label everything in the `lang` dir _'keep out: here be dragons'_.

Whole-heartedly agree. Many of the design principles of Hadron treat this as fundamental - everyone should be able to hack on any part of the system that they like. I still would assert that language developers are going to remain a minority in this community, perhaps for many other reasons outside of "crusty old C++". If that kept people from projects there would not be much open source in the world.

> 
> I also want to improve testing because changes in this part of the code base do have far reaching consequences. One thing I'm planning is code coverage in sclang, so our unit tests can actually cover the class library - this will probably be a separate build though. I also really like your idea of a corpus of code!

Happy to share the corpus with you, if you have a Codeberg account I can easily provide you access, just get in touch. I think it might be valuable to run the new lexer through it, and whatever other compiler bits end up developing. I've found a number of surprises for Hadron this way.

> 
> Re performance: I don't know what hardon is doing, but for us, this change (minus the error posting) _should_ actually give a performance boost as we will only need to do `a ?? d` when there is an expression. The nil check-then-copy can even be vectorised for larger numbers of arguments. Sure we have to do this at every call, but it's in c++ not in the interpreter. I'll have to benchmark it to find out.

IIUC we _already_ only do `a ?? d` in expression initializers. What I'm talking about is the replacement of a `memcpy` for the provided arguments into the callee stack from with a scan checking for nullity. Even vectorized, which can be hard to do where the input size is unknown, it will be slower than a blit.

I've made good progress on designing and writing a benchmark suite specifically for SC interpreters. It's not done, but has enough there that the results have been useful to me when optimizing Hadron. I haven't published it as a quark yet but I actively maintain it at https://codeberg.org/hadron/LangBench. I wrote a rust program in https://codeberg.org/hadron/tools that can compare different sclang builds in statisticaly robust fashion across multiple program runs, thus lowering experimental noise from things like ASLR. Happy to help you get setup and running that benchmarking tool on this PR if you like.

> 
> I think I need to get some numbers on how often this feature is actually used _intentionally_. Before then, everything is a little hypothetical. If this turns out to be a heavily used feature, then of course I'll close this pr - I've spent significantly more time responding here than I have writing code!

Hence the corpus efforts. But absent actual statistics we have to make a judgement call on possibilities. This PR makes it possible for client code to silently fail, now we're arguing about likelihoods. I find any likliehood of that unacceptable. Others do not. Agree this has been a big time sink for many including myself. This underlying difference in values/standards I think is the root of the problem.

However, there is no set of rules for governing SC lang evolution. I think this is unfortunate, because that might be a systemic way for both of us to find agreement, or at least compromise, and feel that justice had been served. Absent that, as I said earlier, SC is a "do-ocracy", this PR and many others represent an impressive body of work on your part, I'm appreciative and respectful of all of that, so I don't want to take more of your time/energy arguing.

> 
> @prko
> 
> > What should users do in cases where they genuinely need to pass nil as an actual value?
> 
> Nil isn't a value, it's the absence of one. `(a: 1)[\a] = nil` deletes the entry.
> 
> Nil is also the _default_ default argument, so the presence of a default argument means we are not expecting this to be nil.
> 
> You can always pass nil if there isn't a default argument - either explicitly, or by omission.
> 
> This pr makes it so you can't pass nil if there is a default argument, it just turns into the default. Which already happened with all but a few very specific (but common) syntaxes. It also aligns SC with other languages, namely js.

JordanHendersonMusic:
> so I don't want to take more of your time/energy arguing.

I'm going to sit and reflect on what you've written here because it deserved some thought, but conversing with you on these issues has been a pleasure — time well spent!

> IIUC we already only do a ?? d in expression initializers. What I'm talking about is the replacement of a memcpy for the provided arguments into the callee stack from with a scan checking for nullity. 

Yes, but if the user has written `a = 1`, which means ` a = 1 or nil`, they need to then write the nil check themselves, expect in the very rare case it is intentional, or when the checks they are doing happen to be implemented as methods on `Nil` (do they check this?). In the case of `Dseq`, there should be a nil check there! This mitigates the majority of the performance gains, and only decreasing the safety/validity of their code. 

---

Today I discovered the ruby actually lets you pass nil and override the default... however it doesn't allow you to pass too few arguments and throws a runtime error, mitigating this problem, which only arises because the *default* default is `nil`.

Spacechild1:
> Today I discovered the ruby actually lets you pass nil and override the default... however it doesn't allow you to pass too few arguments and throws a runtime error, mitigating this problem, which only arises because the default default is nil.

The same is true in Python, btw.

telephon:
The question is, what is the difference in intent between `f.()` and `f.(nil)` ? So far, in sclang there has been a difference in assumed intent. The discussion is whether this distinction should be abandoned in the future.

capital-G:
Sorry - I haven't read everything of this discussion so far, just though I'd also drop by by here ;)

One thing that we could do is introduce an object `Undefined : Nil` where `undefined` behaves the same as `nil`, but it would respond to `isUndefined`. This would allow to pass `nil` explicitly to functions and is also how JS resolves to its default parameters

```javascript
function foo(bar = 42) {
  console.log(bar);
}

bar();  // 42
bar(undefined);  // 42
bar(null);  // null 
bar(0);  // 0
```

Picking up the given example

```supercollider
~f = { |a =10| b.postln; a };
~g = { |a| ~f.(a) };


~g.(); // 10 | "a" becomes undefined => passed to f as undefined => default parameter of `f` gets picked up
~g.(nil); // nil
```

The main change: Not defined parameters become `undefined` which is transparent to `nil` (except a `isUndefined` message?) and `undefined` will resolve to a default parameter when passed around. If it never resolves to a default parameter, it will just behave like `nil`, which is the current behavior.
This would ofc also introduce a breaking change but I think with less disruption (`nil` can still passed explicitly which I assume is in use) by being more explicit instead of only changing implicit behavior?
I don't know if this introduces further problems down the line... I am not a language hacker, just a user ;)

Regarding objects/statements as default arguments, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments - this is also a foot gun in Python and should be avoided.

sclang has many quirks and IMO we shouldn't take users the gun away to shoot into their own foot just for the sake of consistency, "expected" behavior or new features - [*although practicality beats purity*](https://peps.python.org/pep-0020/) also applies here. (btw - [JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features) could maybe be a good inspiration for sclang in general? It also has it quirks due to its past, but there is an agreement in implementations to (almost) never break existing JS code).

---

Although I am not a fan of corpo talk, we could maybe look out for "lower hanging fruits"/"apply the pareto principle" rather than follow such a thorny topic altogether? Although I agree that a change is desirable and makes more sense to me, it is always the question of how much trouble this resolves for our users.

E.g. changing to inline variable declaration also changes (undefined) behavior, but I can explain anyone what the advantage is (as probably everybody has run into it), while the implicit downside of it requires to dig into an (awesome) video from the SC symposium in order to understand what is now different due to this change.
I think more users would be open to try out 3.15 if "it has inline var declarations, though it may break your code" rather than "there is now an undefined value which is like nil but allows you to pass default parameters of functions around - it may break your code though" [insert you wouldn't get it meme here].

OT: Other high pay-offs features of sclang according to capital-G: sclang-C-Extension (imagine the ability to circumvent all this awkward buffer wrapping/passing in flucoma^^ and websockets^^) and language inspection/LSP (think this really help newcomers, which we should care more about IMO). 

---

> So I really hope @capital-G's package manager will let us remove these unmaintained quarks from general use. In my mind, the problem is using unmaintained code - I know this happens a lot, but it's a problem that needs a solution.

I don't think that magnet (the package manager) can solve this - this is a social issue, one of ownership in particular, instead of technical issue. One thing that magnet does is being more explicit about your environment and make it easier to replicate this environment in which code runs.
It doesn't get rid of the problem people not updating their quarks which are in use by users - we would need to have a way to re-claim ownership within the package manager - I don't think we should go down this road, at least w/ our current community/moderation status^^

---
OT:

> Nil isn't a value, it's the absence of one. `(a: 1)[\a] = nil` deletes the entry.

One of the most annoying design decission in sclang in my everyday usage - but did you know that the JSON parser allows you to circuvment this? ;)

```supercollider
"{\"foo\": null}".parseJSON.cs;
-> Dictionary[("foo" -> nil)]

~y = (() ++ ~x).cs; 
-> ($c: $t, $ : $n, $o: $o, $[: $(, 
  $r: $y, $": $ , $): $], $D: $i, $n: $a, 
  $i: $l, $-: $>)
```

The JSON parser is a hot mess (everything except null becomes a string) and should be fixed, but this would also break user code, though by replacing a faulty implementation w/ a correct one.

JordanHendersonMusic:
> One thing that we could do is introduce an object Undefined : Nil where undefined behaves the same as nil

Edit: scratch this following paragraph - it's possible.

You can't inherit from a literal - it's not an object so has no where to store a class ptr. This means we'd have to duplicate all the methods of Nil, including user defined ones, and alter all the byte codes that operate on nil. Not practical unfortunately! Otherwise this could work.

Also remember the default default is nil, so changing that effects every function, this change only affects functions defined like `x = 1` where the call sight happens to pass nil.

> Regarding objects/statements as default arguments, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments - this is also a foot gun in Python and should be avoided.

Our arguments are immutable so this is okay. We do this 
```python
def append_to(element, to=None):
    if to is None:
        to = []
    to.append(element)
    return to
```
**Except when it's a literal define like x = 1!!** Where we omit the `is None` check allowing you to reassign `to` back to `None` - hence this problem.

The benefit is that doing this 
```
f = { |a, b| g.(a, b) }
```
actually uses `g`s defaults. Right now it doesn't use the defaults if defined like `x = 1` and to literal, but does in all other cases!! The benefit is that you don't have to read the source code of all the functions you are calling, neither do you have to know about this odd behaviour to simply wrap a function.

Yes, the inline thing is a much more obvious benefit, but I still think this subtly affects users - I've already shown how `DSeq` is broken, there are hundreds of classes like this. The only solution right now is to copy and paste the arguments into the wrapper function.

> The JSON parser is a hot mess

I think I'm in love 💕

capital-G:
> You can't inherit from a literal - it's not an object so has no where to store a class ptr. This means we'd have to duplicate all the methods of Nil, including user defined ones, and alter all the byte codes that operate on nil. Not practical unfortunately! Otherwise this could work.

Just babbling:  `nil` is in the end an opcode on the stack IIUC - couldn't we enhance this `nil` opcode w/ a flag outside of the current op-code-range (e.g. opcode are limited w/ 8 bits, but we could "enhance" them with bits 9 and 10? a bit like in nan tagging?) - the additional bits only gets checked when a function gets called in the vm since this is only where we need to check if it is undefined or nil?
Or just add it as another singleton next to `nil` which simply redirects everything except the function call opcode?

Though I think I have to read more crafting interpreters before I can provide you proper inspirations^^

prko:
Thank you for the work on this PR and for the detailed discussion so far.

I’m not opposed to breaking changes in principle, and I’ve had to update my own code for genuine bug fixes in the past: <https://github.com/prko/Notator/issues/1#issuecomment-3744904264>.

As I understand it, the PR’s goal is to make the calling semantics consistent: when an argument has a default, explicitly passing `nil` will always mean “use the default” (and the current literal‑vs‑expression inconsistency would disappear). I can see why that is attractive from an implementation and language‑design perspective.

My concern is primarily about user impact and ecosystem compatibility.

1) This is effectively an ABI/behaviour change at the language level: the same source code may behave differently depending only on the sclang version. Even if a warning is posted in 3.15, in 3.16 the warning goes away and the behaviour changes—so users who run older Quarks or shared code (which they did not write) may end up with changed behaviour without any diagnostic pointing them to the cause.

2) In the SuperCollider ecosystem, there is no automated mechanism to notify authors of older Quarks or code snippets (e.g. those shared on sccode.org), and many of those projects are unmaintained. This makes it difficult to rely on “authors will update their code” as a mitigation strategy. It also suggests that Tutorials and SCHelp examples would need a careful audit to avoid teaching patterns that become subtly version‑dependent.

3) From the perspective of many non‑developer users, cross‑language arguments (e.g. “JavaScript/Lua do X”) are not especially persuasive. What matters most to them is whether their existing SuperCollider code—and the libraries they depend on—continues to behave the same way after an upgrade.

Related to my earlier question in this thread: if passing `nil` to a defaulted argument is redefined to always mean “use the default”, what is the recommended way to express the other intent—i.e. “I explicitly want this argument to be `nil`”? Even if the view is that “nil is the absence of a value”, in practice some APIs currently use `nil` as a meaningful sentinel distinct from the default.

Given all of the above, I’m wondering whether the benefits of this change outweigh the likely disruption, and whether a less invasive disambiguation mechanism (e.g. an explicit ‘default’ sentinel or syntax) could be worth exploring.

JordanHendersonMusic:
@capital-G I've edited my post, undefined could work! We'd have to do some compiler magic and it would be the first inbuilt type like this, but it could work. Shouldn't be too hard either, although does involve introducing a new keyword, potentially an even bigger breaking change than this depending on how frequently this is *intentionally* used.

@prko

> “I explicitly want this argument to be nil”

Default arguments mean, 'nil has no meaning here, instead use - blah'. 
```
+ { | rhs = 0 | ... }
* { | rhs = 1 | ... }
```
If you want extra behaviour, which is what we talked about with send bundle, use literally any other value. A symbol is often a good choice. Handling old code is the issue here, but it not an issue for new code, because this feature has no unique use.

There is only one instance of this triggered in the test suite, changing it is was simply a matter of changing `nil` to `\none`, no other change was needed. 

It is only with the `x = LITERAL` syntax where it secretly means `x = LITERAL or nil`. This behaviour is not triggered if you write `x = (1)`. Hence why `f.(nil)` would often be the same as just `f.()`.

> in practice some APIs currently use nil as a meaningful sentinel distinct from the default

Do you have examples!?

Spacechild1:
> what is the recommended way to express the other intent—i.e. “I explicitly want this argument to be nil”? Even if the view is that “nil is the absence of a value”, in practice some APIs currently use nil as a meaningful sentinel distinct from the default.

I would say the intent is wrong. If an argument has a default value, it should never be `nil` in the first place. If you want `nil` as a sentinel, don't provide a default value. If you do want a default value, pick another sentinel.

It's basically the same with Events: you can't use `nil` as a sentinel value because `nil` is used to denote the absence of a key.

One part where the whole thing really breaks down (and Jordan has already mentioned this) is `performWithEnvir` or `valueEnvir`: it is not possible to explicitly pass `nil` as an argument that way, simply because a key with value `nil` cannot exist in an environment. I think this clearly demonstrates the logical flaw of allowing `nil` to override default arguments.

As Jordan said, it would be good to collect of code examples, especially in Quarks, that actually do this, so we can get a feeling of how much of a breaking change this would be.

lnihlen:
> I would say the intent is wrong. If an argument has a default value, it should never be `nil` in the first place. If you want `nil` as a sentinel, don't provide a default value. If you do want a default value, pick another sentinel.

If software engineers only ever used semantics offered them in a language correctly, we'd live in a very different world. It's not hard to imagine folks using `nil` as a sentinel value. Yes, a flag is more appropriate but [Hyrum's Law](https://www.hyrumslaw.com/) particulary in a dynamically-typed language with a non-engineering user base (no shade) dictates that this is very likely happened.

The argument that this is "bad code" doesn't make the harm go away. We have to respect all of our users, even those that use SC in ways we don't think are elegant or intuitive. This is why I keep asking folks to set their value judgements aside and consider the well-being of the community over lofty opinions of "correct" usage.

Spacechild1:
> It's not hard to imagine folks using nil as a sentinel value.

Sure, but there are cases where it just doesn't work or doesn't make sense. I think it's worth pointing this out to prevent users from making that mistake.

> The argument that this is "bad code" doesn't make the harm go away. We have to respect all of our users, even those that use SC in ways we don't think are elegant or intuitive. 

I agree. I never really argued for a change in the language, I just wanted to help people understand why the current behavior is logically wrong and how it can be problematic. Quoting my very first post in this thread:

> However, I agree with the others that we must be careful here because there is certainly quite a bit of code out there that relies on the current behavior. This is a tricky situation...

I take backwards compatibility quite seriously, exactly for the reasons you've pointed out in this thread.

prko:
@Spacechild1 
>I would say the intent is wrong. If an argument has a default value, it should never be nil in the first place. If you want nil as a sentinel, don't provide a default value. If you do want a default value, pick another sentinel.

Thank you for correcting me!

@JordanHendersonMusic 
>>in practice some APIs currently use nil as a meaningful sentinel distinct from the default
>
>Do you have examples!?

In VSCode, I tried to find the usage of `nil` with `([^:]:|[^=!]=|[^'"\]\)|}],)(| )\bnil\b(?!;)` in the latest SC build's resource folders on macOS.
At a glance, it does not seem to have such examples.

I am searching for the same results in the default downloaded-quarks folder; VSCode keeps searching endlessly and shows no results.

Anyway, I remember trying some algorithms using `nil`, `??`, and `!?` in SC.
Perhaps my memory is conflated, since I can’t point to a concrete example right now.
So, at least in the current codebase, I couldn’t find a clear case at first glance where `nil` is intentionally used as a sentinel distinct from the default.

Given that, I find the arguments of @Spacechild1 and @JordanHendersonMusic are convincing in this context, and I’m inclined to reconsider my previous assumption.
That said, I still find @lnihlen’s perspective reasonable, and it might be worth keeping in mind for future API design discussions.

JordanHendersonMusic:
>  consider inclusion of our user base in every move.

> We have to respect all of our users, even those that use SC in ways we don't think are elegant or intuitive. This is why I keep asking folks to set their value judgements aside and consider the well-being of the community over lofty opinions of "correct" usage.

>  because this is a pattern of behaviour for both of us that highlights a difference in underlying values. I think the SC user community is the most valuable asset of SC. There are musicians and artists who have learned just enough SClang to make their art. Lacking a CS or engineering background, that learning has likely been painful.

@lnihlen 

I complete agree with everything you've said here!

I think the difference between our views might come from that you are caring for established supercollider users, those who have been through this 'painful' process, while I'm trying to improve the situation for new users. A significant proportion of sc users are students who are both learning supercollider, sound design, building their artistic practice, and computer programming — all at once! Having taught supercollider to students (postgrads), I know they find this a difficult task (I did!).  These weird edge cases make this situation worse. What is more, these users are almost never represented in online discussions and are drastically underrepresented, yet make up a large portion of our user base.

I'm not arguing that we should remove this behaviour because I think its incorrect, I think we should remove it because: it makes learning the language harder (can't write wrappers easily; unique compared to other languages), doesn't add anything of value (easily replaced by a different sentinel), only works in arcane situations (weird syntax/literals vs expressions), often used incorrectly (code that doesn't work with `nil` yet uses `arg x = 1`), and is probably (?) infrequently used (one in test suite found so far).

The work that I've been trying to do on sc is mostly driven around improving the new user experience, improving error messages is one of these places, trying to stomp out these unnecessary hurdles to learn the language is another. I am trying to make the language more inclusive by making it easier (or at least, have less thorny edges to worry about).

My view is that those with large established code bases should accept *some small level* of pain if they want to stay updated to improve the situation for new users, as ultimately the community withers without them. This isn't ideal, but I can't see another way to improve the situation for new users and students?

The question I have regarding this PR is just how much pain will this actually cause to established users, because this is of course a balance.

lnihlen:
> I think the difference between our views might come from that you are caring for established supercollider users, those who have been through this 'painful' process, while I'm trying to improve the situation for new users. A significant proportion of sc users are students who are both learning supercollider, sound design, building their artistic practice, and computer programming — all at once! Having taught supercollider to students (postgrads), I know they find this a difficult task (I did!). These weird edge cases make this situation worse. What is more, these users are almost never represented in online discussions and are drastically underrepresented, yet make up a large portion of our user base.

I think this is an excellent point. I think the best (but not perfect) way to represent the perspective and needs of new users is to listen to educators and understand the issues they see when teaching students new to the language. I didn't know you were approaching this work from that perspective, my assumption was that your approach was similar to mine, a language nerd. Of course, both can be true :).

> My view is that those with large established code bases should accept _some small level_ of pain if they want to stay updated to improve the situation for new users, as ultimately the community withers without them. This isn't ideal, but I can't see another way to improve the situation for new users and students?
> 
> The question I have regarding this PR is just how much pain will this actually cause to established users, because this is of course a balance.

I have no issue with causing Quark maintainers some pain in order to make the language more accessible to newcomers. I just wish we could, in all cases where we are causing any user pain, have that pain take the form of a deterministic compilation error. Code that we as a dev team have broken should not compile.

But it's an imperfect world we live in. And it's true that code like this will break, at least for a release or so, which is better than nothing.

Thank you for taking the time to talk through my concerns on this PR. I really appreciate the care you've taken to be thoughtful and clear. I approve this change. I don't know if that meets the criteria for merge but LGTM.

lnihlen (approved):

jamshark70:
> What should users do in cases where they genuinely need to pass nil as an actual value?

One convenient solution would be to pass a reference to nil.

```
(
f = { |sometimesShouldBeNil(2)|
	sometimesShouldBeNil.dereference
};
)

f.()
-> 2

f.(nil)
-> 2

f.(`nil)
-> nil
```

This wouldn't help if the argument also sometimes needs to be a reference -- and to keep its status as a reference -- but I'm willing to bet that such cases basically don't exist (whereas we already identified some cases where just passing a nil is useful).

muellmusik:
In the spirit of the day I propose we create a new SuperNil.On 1 Apr 2026, at 02:52, H. James Harkins ***@***.***> wrote:﻿jamshark70 left a comment (supercollider/supercollider#7413)

What should users do in cases where they genuinely need to pass nil as an actual value?

One convenient solution would be to pass a reference to nil.
(
f = { |sometimesShouldBeNil(2)|
	sometimesShouldBeNil.dereference
};
)

f.()
-> 2

f.(nil)
-> 2

f.(`nil)
-> nil

This wouldn't help if the argument also sometimes needs to be a reference -- and to keep its status as a reference -- but I'm willing to bet that such cases basically don't exist (whereas we already identified some cases where just passing a nil is useful).

telephon:
Since the main inconvenience is that you want to be able to distinguish between "nothing" and "unsupplied", mainly for being able to do argument forwarding in a convenient way, it would make sense to have a value that behaves like `nil`, but explicitly means "unsupplied".

So that 
```
g = { |x=1| x };
f = { |x| g.(x) }; 
f.(nil); // returns nil because it calls g.(nil)
f.(); // calls g.(unsupplied) which is equivalent to g.() and returns 1
```

At least this seems like a typical reification of a semantic concept.

JordanHendersonMusic:
Changing the default default argument will affect all code. 

Right now, the default looks like `x = nil`, changing it to `x = unsupplied` is a huge breaking change.

What happens if you put `unsupplied` into an identity dictionary?  How does this new type compare with `nil`. Adding these new fundamental types is altering supercollider's 'ontologoy' so to speak. I would not feel comfortable doing that. 

What should the default value of an initialized variable be? Why make args and vars different?

>  Since the main inconvenience is that you want to be able to distinguish between "nothing" and "unsupplied", 

Do you mean the main inconvenience to accepting this PR? Also, 'I' do not want to be able to distinguish between this, and haven't seen an example of where someone does. This conversation so far has been focused on using `nil` as a sentinel, I understand this use, but I don't not see why you'd ever want to draw a distinction between what is ultimately syntax.

Do you have an example?

telephon:
The word "you" was not meant as a personal address, but as a synony of "we". But please keep in mind that it is intended as being part of an open discussion.

I am aware that there may be consequences that need to be considered, and that it may be a bad idea. If "unsupplied" behaves exactly like `nil` but denotes the extra information of an explicit user intended omission, I see less of a problem currently.

> What should the default value of an initialized variable be? Why make args and vars different?

This should be nil because you cannot dynamically supply a variable with a value analogously to argument forwarding.

> > Since the main inconvenience is that you want to be able to distinguish between "nothing" and "unsupplied",
> 
> Do you mean the main inconvenience to accepting this PR?

No the inconvenience that motivates this PR. As I understand it, the main inconvenience is how argument forwarding works (`{ |x| g.(x) }`, where g has a default arg).

As far as I know, it was always an intentional choice in sclang to interpret `f.(nil)` as the caller's intention to pass `nil` and not using the default value. Using the default value is written as `f()`. The only problem is the above case of forwarding.

JordanHendersonMusic:
> As far as I know, it was always an intentional choice in sclang to interpret f.(nil) as the caller's intention to pass nil

Based on the code, I think it arises out of an optimisation where `a ?? {}` is omitted when it's a literal as the value is stored in the frame. I don't know the history here as it's in the initial commit.


> No the inconvenience that motivates this PR. As I understand it, the main inconvenience is how argument forwarding works ({ |x| g.(x) }, where g has a default arg)

Ah, I understand! Yes that's a problem, but I think the biggest problem is that no one seems to use `a=1` correctly. They must write the body to work when a is nil.

Literally the other day @HotwheelsSisyphus ran into this in a pr, I've made this mistake many times, and there are many places in the class library where this will cause a bug. Most of the places where `a(1)` are used are there to fix issues caused by this - most of these prs were authored by you Julian! -, there are also places where people use `a=1` check for nil and then *reassign* the default, it's like the authors are saying 'no please I really did mean what I wrote!'.

If we, the authors of the class library, can't use this correctly, how do we expect students and those from musical backgrounds? That is what I view the biggest issues here as, it's confusing, it's very weird to switch on syntax, but most of all, it's the default, meaning this is forced on all users.

If we actually do want this, I've suggested `a ?= 1` as a new syntax, this way, it's not the default. Again, I think this is useless and we shouldn't switch on syntax, and the same behaviour can be achieved by just not having a default, but if it is absolutely necessary to preserve this, this syntax would allow that when needed.

telephon:
Really, the current behaviour is not unintended, it is an intended feature of sclang. I've often explained it to students and there was never a problem in that. It is just that sometimes you get it wrong.

But yes, sure we can think about it.



>>    As far as I know, it was always an intentional choice in sclang to interpret f.(nil) as the caller's intention to pass nil

> Based on the code, I think it arises out of an optimisation where a ?? {} is omitted when it's a literal as the value is stored in the frame. I don't know the history here as it's in the initial commit.

Yesterday I asked James, he argued that it was intended. He argued that if the user passes nil, we should assume that this is intentional and not that it should be replaced by a default -- and if you need a different behaviour, there is the `{ |x (1)| ` syntax.

I see why one would want to have it differently, but this is the way it has been all along.

JordanHendersonMusic:
> Yesterday I asked James, he argued that it was intended. He argued that if the user passes nil, we should assume that this is intentional and not that it should be replaced by a default

The issue i have is that this isn't consistent with how `nil` is used elsewhere.

Consider this (ignoring that buffer isn't a literal for now...)

```
var loadBuffer = { |path| if(path.notAudioFile) {
      nil 
   } { 
      Buffer(path) 
   }
};
var defaultBuffer = Buffer(...);
var playBuffer = { | buf = defaultBuffer|
   ...
};

playBuffer.(loadBuffer.("Wrong path"))
```

Here, it is clearly intended for the default to be used, not nil. The user has not passed nil, they called a function. I can see how you might want to pass nil when the user types `nil`, but not here, and not when you passing arguments along.

You cannot tell the difference between the absence of a value and an unspecified one. This feature isn't not consistent with the rest of the language. 

It's also impossible to write bytecode that knows whether the argument was specified or not, which is why this doesn't work with non-literals.

Consider this example and what g must do to correctly pass on the values, preserving the unspecified state.
```
f = { |a=1, b=2| ...}
g = { |a, b| 
   if(b.isNil) { 
      if(a.isNil) {f.()} { f.(a) }
   } {
      if(a.isNil) {f.(b : b)} {f.(a, b)}
   }
}
```
Now consider a third argument...

What's worse is this still doesn't even really work because it doesn't work when the user does `g.(nil)`.... 

Sure, you can do it with `g ={|...a, k| f.valueArgs(a,k)`, but now you don't get to see the arguments in the signature, which is often very useful!

I also think it's problematic you can 'unspecify' an argument in the middle without using keyword arguments.

telephon:
Thanks!

 assume 
```
f = { |a=1,b=2| [a, b] };
g = { |a,b| f.(a, b) };
```

Now if we had a -- hypithetical -- second `nil` value that externally behaves exactly like nil. Only from the backend, within a function call, you could distinguish it as `unspecified`. Then, if the default-default-arguments of a function (i.e. the default argument when no default argument is supplied), here, the defaults of `g`, were such an `unspecified` nil, this value would be passed to the call `f.(unspecified, unspecified)`, and would lead to the defaults of `f` being used.

This is the behaviour that you want, right?

And if you call `g(nil)`, then this would return `[nil, 2]`, as expected.

JordanHendersonMusic:
> And if you call g.(nil), then this would return [nil, 2], as expected.

I wouldn't expect this - hence the initial pr where I thought this was a bug, and all the many mistaken uses of this.

Your solution doesn't work for functions that return an optional value or nil (like an event lookup). 

It would also be very hard to keep this in the backend, it would probably leak out into user code somehow.

---

My point with the previous post was to try and show how this feature is broken in SC right now, to the point where it can't be used. This means we aren't talking about deleting behaviour, but removing an unimplemented feature, similar to how literal dictionarys were recently removed.

---

I still think `f = {|a ?= 1, b ?=2| ... }` is a good solution because it is applied at the small scope, at the functuon call sight, and does not introduce a new primitive type throughout all code.

JordanHendersonMusic:
Just to recap from the dev meeting as some participants here weren't at the meeting.

None of these options are good.

Introducing a new type of nil requires being consistent with the following ways nil can be checked for:
* The C++ function `IsNil`
* A byte comparison with an existing nil.
* The supercollider function `isNil`
* The bytecodes `isNilJump` and similar used to implement `??` and friends

This means checking all the source code in the `lang/` directory and even then, there will probably be logical inconsistencies with existing code — dangerous. This approach does however solve the issue and would be a great solution if we were redesigning the language.

Removing the feature would potentially be okay if we could do it at compile time, we can't therefore this will be messy.

Basically, it is this way and isn't likely to change, people need to know the difference, and probably not use `a = 1` unless they really meant to.

I started a forum post about this subject: https://scsynth.org/t/the-difference-between-defaults-arguments-with-equals-sign-and-brackets-arg-a-1-vs-arg-a-1/13261/1

JordanHendersonMusic:
Just found a bug in the CI caused by this.

If you assert without a message then the default isn't applied .....
```supercollider

SomeTest : UnitTest {
   test_blah { this.assert(false) }
}

UnitTestResult {
	*new { |testClass, testMethod, message="", details, debug|
                                           ^^not applied
		^super.newCopyArgs(testClass ? this, testMethod ? thisMethod, message, details, debug)
	}
}
```
...and then this fails...
```supercollider
newFailures.do {|fail|
	test[\results].add((
		\pass: false,
		\test: ("" ++ fail.message.split($\n)[0])[0..1000],
		\reason: (fail.message.split($\n)[1..])[0..1000]
	))
};
```
... as `fail.message` is `nil` which doesn't respond to `split`.

The assert without a commit message happened in `TestSerialPort`.

dyfer:
> The assert without a commit message happened in `TestSerialPort`.

I previously noticed that `TestSerialPort` would fail when moving unit tests to Ubuntu 24.04 in CI (that's why Linux tests still run on 22.04), could that be the issue? I don't see how this would behave differently on different OSes.

JordanHendersonMusic:
The serial port behaves differently yes (don't know why), but it is the failure and calling  `nil.split` which bricked the test runner.

Issue #7130, opened 2025-09-05
Some shortcuts do not work in code region in Help Browser


prko:
## Environment

* SuperCollider version: 3.14.0
* Operating system: macOS
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```
1. Quit the server 
2. Type SinOsc in the IDE editor
3. Press CMD + D
4. Go to `Example` section, and click the fourth line (an empty line) inside the code tag area.
5. Press CMD + B
6. Type s.boot in a blank line and Press CMD + enter. 
7. Press CMD + M
8. Press CMD + Shift + M
9. Press CMD + Alt + M
10. Press CMD + M
11. Press CMD + Shift + M
12. Press CMD + Alt + M
13. Press CMD + T
14. Press CMD + Shift + T
15. Press CMD + Alt + T
16. Go to the third line beginning with `{ SinOsc.ar`.
17. Press Shift + enter
18. Press CMD + .
19. Press CMD + enter
20. Press CMD + .
21. Double-click SinOsc and Press CMD + I
22. Press CMD + Shift + P
23. Press CMD + Shift + L
24. Double-click `play` and Press CMD + D
```
### Expected
All shortcuts should work.
### Actual
Some shortcuts do not work:
- CMD + B
- CMD + M
- CMD + Shift + M
- CMD + Alt + M
- CMD + T
- CMD + Shift + T
- CMD + Alt + T
- CMD + I (reported in #6820)
- CMD + Shift + P (reported in #6820)
- CMD + Shift + L
- CMD + D

Some of them work outside of the code region inside the Help Browser on macOS.
However, they work in Windows at least, and they worked on older versions of SC on macOS if I remember correctly.

prko:
`Cmd + D` has been added because, while `Ctrl + D` works on Windows, the equivalent shortcut did not work on macOS.

prko:
There is a difference between Linux and Windows:
- To see help inside the code region of the Help browser:
   - Linux: Place the cursor on the word
   - Windows: Double-click the word.

--- 

On some Linux systems such as Ubuntu, `ctrl + alt + t` does not call the node tree window but opens a new terminal window. This is not only within the code region of the Help browser but also in the code editor.

capital-G:
IIUC this is about precedence of capturing keys or passing them upwards. IIRC this is always a bit tricky b/c of multiplatform support. You can take a look at e.g. https://github.com/supercollider/supercollider/pull/6762 for more information about it :)

[...]

PR #7414, opened 2026-03-15
Topic/enlarge shrink font

### LINKED ISSUE #7373: Unable to change font size in IDE Post window (shortcuts and context menu unresponsive)
## Environment

* Operating system: All (macOS 15.7.4, Ubuntu 22.04.4, Windows 11)

## Steps to reproduce

```supercollider
// No specific code required; the issue occurs directly in the IDE.

// 1. Open the SuperCollider IDE
// 2. Attempt to change the Post window font size using:
//    - Ctrl + +
//    - Ctrl + -
//    - Ctrl + 0
// 3. Alternatively, right‑click in the Post window and select:
//    - Enlarge Font
//    - Shrink Font
//    - Reset Font Size
```

**Expected:**  
The Post window font size should increase, decrease, or reset when using the keyboard shortcuts or the context‑menu options.

**Actual:**  
None of the font size controls has any effect.  
No error messages are shown.

**Note:**
On macOS, no keyboard shortcut appears when right‑clicking in the Post window.

stefandiaconu:
## Purpose and Motivation

Fixes #7373 

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing

### DIFF
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -96,13 +96,13 @@ void PostWindow::createActions(Settings::Manager* settings) {
     mActions[ZoomIn] = ovrAction = new OverridingAction(tr("Enlarge Font"), this);
     ovrAction->setIconText("+");
     ovrAction->setStatusTip(tr("Enlarge post window font"));
-    connect(ovrAction, &QAction::triggered, this, [this]() { zoomIn(); });
+    connect(ovrAction, &QAction::triggered, this, &PostWindow::zoomIn);
     ovrAction->addToWidget(this);
 
     mActions[ZoomOut] = ovrAction = new OverridingAction(tr("Shrink Font"), this);
     ovrAction->setIconText("-");
     ovrAction->setStatusTip(tr("Shrink post window font"));
-    connect(ovrAction, &QAction::triggered, this, [this]() { zoomOut(); });
+    connect(ovrAction, &QAction::triggered, this, &PostWindow::zoomOut);
     ovrAction->addToWidget(this);
 
     mActions[ResetZoom] = ovrAction = new OverridingAction(tr("Reset Font Size"), this);
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -96,16 +96,19 @@ void PostWindow::createActions(Settings::Manager* settings) {
     mActions[ZoomIn] = ovrAction = new OverridingAction(tr("Enlarge Font"), this);
     ovrAction->setIconText("+");
     ovrAction->setStatusTip(tr("Enlarge post window font"));
+    ovrAction->setShortcutVisibleInContextMenu(true);
     connect(ovrAction, &QAction::triggered, this, [this]() { zoomIn(); });
     ovrAction->addToWidget(this);
 
     mActions[ZoomOut] = ovrAction = new OverridingAction(tr("Shrink Font"), this);
     ovrAction->setIconText("-");
     ovrAction->setStatusTip(tr("Shrink post window font"));
+    ovrAction->setShortcutVisibleInContextMenu(true);
     connect(ovrAction, &QAction::triggered, this, [this]() { zoomOut(); });
     ovrAction->addToWidget(this);
 
     mActions[ResetZoom] = ovrAction = new OverridingAction(tr("Reset Font Size"), this);
+    ovrAction->setShortcutVisibleInContextMenu(true);
     connect(ovrAction, &QAction::triggered, this, &PostWindow::resetZoom);
     ovrAction->addToWidget(this);
 
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -96,19 +96,16 @@ void PostWindow::createActions(Settings::Manager* settings) {
     mActions[ZoomIn] = ovrAction = new OverridingAction(tr("Enlarge Font"), this);
     ovrAction->setIconText("+");
     ovrAction->setStatusTip(tr("Enlarge post window font"));
-    ovrAction->setShortcutVisibleInContextMenu(true);
     connect(ovrAction, &QAction::triggered, this, [this]() { zoomIn(); });
     ovrAction->addToWidget(this);
 
     mActions[ZoomOut] = ovrAction = new OverridingAction(tr("Shrink Font"), this);
     ovrAction->setIconText("-");
     ovrAction->setStatusTip(tr("Shrink post window font"));
-    ovrAction->setShortcutVisibleInContextMenu(true);
     connect(ovrAction, &QAction::triggered, this, [this]() { zoomOut(); });
     ovrAction->addToWidget(this);
 
     mActions[ResetZoom] = ovrAction = new OverridingAction(tr("Reset Font Size"), this);
-    ovrAction->setShortcutVisibleInContextMenu(true);
     connect(ovrAction, &QAction::triggered, this, &PostWindow::resetZoom);
     ovrAction->addToWidget(this);
 
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -96,13 +96,13 @@ void PostWindow::createActions(Settings::Manager* settings) {
     mActions[ZoomIn] = ovrAction = new OverridingAction(tr("Enlarge Font"), this);
     ovrAction->setIconText("+");
     ovrAction->setStatusTip(tr("Enlarge post window font"));
-    connect(ovrAction, &QAction::triggered, this, &PostWindow::zoomIn);
+    connect(ovrAction, &QAction::triggered, this, [this]() { zoomIn(); });
     ovrAction->addToWidget(this);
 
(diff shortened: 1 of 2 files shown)


capital-G (commented):
Hey thanks for the PR! :)

I think it would be best to set this on a global level and not just for specific values within a context menu?

You can take a look at https://doc.qt.io/qt-6/qt.html#ApplicationAttribute-enum on how to set this as a global flag, so it also affects the HelpWindow and the Editor window.

You can set this value in the mainwindow file.

prko:
Thanks!
@capital-G 
Then, can #7130 also be fixed?

capital-G:
> Then, can #7130 also be fixed?

#7130 is unrelated to these changes, see https://github.com/supercollider/supercollider/issues/7130#issuecomment-4064276425 for more information :)

stefandiaconu:
@capital-G, I completely agree with that. I looked for entry point of the application but I found the wrong one, SCDoc entry point. I need to get more failiar with the source code.

I agree to put it in `mainwindow.cpp`, I can do it first line in the constructor initialization or last line. But after finding the right entry point for the applicatin, I am proposing to add it in `main_function.cpp`, right after `QApplication app(argc, argv);`. I tested it quickly and it does show the shortcuts for the other options from context menu, not only enlarge, shrink and reset font.

I am happy to put it either in `main_windows.cpp` or `main_function.cpp`. for me it would make more sense in `main_function.cpp`. Let me know your thoughts.

HotwheelsSisyphus:
I wonder, could this also be made to work with a trackpad pinch-to-zoom gesture? Pinch to zoom currently works for the help browser (where it changes the zoom factor, though probably not the font size itself). But not for the editor and post win...
At any rate, thanks!

capital-G:
> I agree to put it in `mainwindow.cpp`, I can do it first line in the constructor initialization or last line. But after finding the right entry point for the applicatin, I am proposing to add it in `main_function.cpp`, right after `QApplication app(argc, argv);`. I tested it quickly and it does show the shortcuts for the other options from context menu, not only enlarge, shrink and reset font.

Yes - `main_function` looks like a good place to set this attribute!


> I wonder, could this also be made to work with a trackpad pinch-to-zoom gesture? Pinch to zoom currently works for the help browser (where it changes the zoom factor, though probably not the font size itself). But not for the editor and post win...

A bit OT, but this feature exists (at least) for macOS - see https://github.com/supercollider/supercollider/pull/6563
I don't have access to a Linux touchpad, so I can't test it on it, but maybe you can use the code to adapt it for Linux as well?

HotwheelsSisyphus:
> A bit OT

sorry... but thanks for the link! I keep forgetting that I'm on a different OS now... anyway, I mentioned this here because I thought it might be related and hadn't found an open issue (probably because said issue has been resolved, oops... oh well). I might look into it at some point but if it's already working on mac that's probably good enough.

dyfer:
@HotwheelsSisyphus trackpad pinch to zoom implementation shouldn't be macos specific - it _should_ work on other platforms if 1) the platform and 2) Qt both support it. However, support for touchpad gestures varies among platforms. Can you confirm with SC 3.14 or current dev whether it works or not on your system?

EDIT I guess support for pinch gestures is different on Windows and Linux and we likely would need additional logic to handle other pinch methods depending on the platform (Windows vs Linux X11 vs Linux Wayland).

stefandiaconu:
@capital-G , thanks for confirmation. I have updated the PR.

editors/sc-ide/widgets/post_window.cpp
```diff
@@ -96,13 +96,13 @@ void PostWindow::createActions(Settings::Manager* settings) {
     mActions[ZoomIn] = ovrAction = new OverridingAction(tr("Enlarge Font"), this);
     ovrAction->setIconText("+");
     ovrAction->setStatusTip(tr("Enlarge post window font"));
-    connect(ovrAction, &QAction::triggered, this, &PostWindow::zoomIn);
+    connect(ovrAction, &QAction::triggered, this, [this]() { zoomIn(); });
```
capital-G: can you revert these two lines? Or was the change intentional?

editors/sc-ide/widgets/post_window.cpp
```diff
@@ -96,13 +96,13 @@ void PostWindow::createActions(Settings::Manager* settings) {
...
     ovrAction->addToWidget(this);
 
     mActions[ZoomOut] = ovrAction = new OverridingAction(tr("Shrink Font"), this);
     ovrAction->setIconText("-");
     ovrAction->setStatusTip(tr("Shrink post window font"));
-    connect(ovrAction, &QAction::triggered, this, &PostWindow::zoomOut);
+    connect(ovrAction, &QAction::triggered, this, [this]() { zoomOut(); });
```
capital-G: same as above

capital-G (changes requested):
Except the two unrelated LOC this looks good

editors/sc-ide/widgets/post_window.cpp
stefandiaconu: The change is intentional. Using lambda will use the default `steps` paramaeter, which is set to 1. Using `&PostWindow::zoomIn` the `steps` is 0. That's why the enlarging of font was not working. This change should fix it.

editors/sc-ide/widgets/post_window.cpp
stefandiaconu: Same as above

editors/sc-ide/widgets/post_window.cpp
capital-G: I can zoom in/out in the post window using cmd+/cmd-, the menu and also the touchpad, so I think the existing code works?
Also, the lambda function would invoke the same behavior as the existing signal, but in the new code we would capture `this` - this is mostly fine but we only use this pattern rarely and I'd prefer to have some consistencies within the signals of the IDE (see https://github.com/supercollider/supercollider/pull/7282)

stefandiaconu: @capital-G I completely agree being consistent. The bug raised was mentioning that the enlarge/shrink font is not working, either using context menu or cmd+/cmd-. When I run the source code using CLion on macOS I can see the same behaviour as described in the bug raised. Now here's the twist. I downloaded SuperCollider 3.14.1 universal bynary, macOS 11+ and that functionality works just fine. So I reverted the commit that uses lambda and is ready for review.

capital-G (approved):
Hey - I think we, or at least I, got a bit mixed up here - big sorry! There seems to be two issues

a) No shortcuts are shown in dialog menu (which this PR fixes)
b) Keyboard shortcuts and dialog menu actions don't work for zoomIn / zoomOut - I didn't know these were broken, so I got confused in the PR review - sorry for this.

It is always good to keep things atomic, so since this PR fixes a) I'll merge it - thanks for this and congrats to the first merged PR! :)
If you want to tackle b), please create another PR :)

Regarding b) - we have a signal which is wired up like this

https://github.com/supercollider/supercollider/blob/b365366bc6bba71a8c2208ad820b6d2bcf2e2c38/editors/sc-ide/widgets/post_window.cpp#L99

but the signature of `zoomIn` actually also has a parameter in its signature, which defaults to `1` in the header

https://github.com/supercollider/supercollider/blob/b365366bc6bba71a8c2208ad820b6d2bcf2e2c38/editors/sc-ide/widgets/post_window.cpp#L249

Maybe this creates some problems? It is probably worth to attach a debugger to the zoomIn method ;)

stefandiaconu:
> Hey - I think we, or at least I, got a bit mixed up here - big sorry! There seems to be two issues
> 
> a) No shortcuts are shown in dialog menu (which this PR fixes) b) Keyboard shortcuts and dialog menu actions don't work for zoomIn / zoomOut - I didn't know these were broken, so I got confused in the PR review - sorry for this.
> 
> It is always good to keep things atomic, so since this PR fixes a) I'll merge it - thanks for this and congrats to the first merged PR! :) If you want to tackle b), please create another PR :)
> 
> Regarding b) - we have a signal which is wired up like this
> 
> https://github.com/supercollider/supercollider/blob/b365366bc6bba71a8c2208ad820b6d2bcf2e2c38/editors/sc-ide/widgets/post_window.cpp#L99
> 
> but the signature of `zoomIn` actually also has a parameter in its signature, which defaults to `1` in the header
> 
> https://github.com/supercollider/supercollider/blob/b365366bc6bba71a8c2208ad820b6d2bcf2e2c38/editors/sc-ide/widgets/post_window.cpp#L249
> 
> Maybe this creates some problems? It is probably worth to attach a debugger to the zoomIn method ;)

No problem. I will create another PR. I did debug it and the `zoonIn` parameter is 0. After I searched on Google, the issue is that the parameter default value is not preserved when called via a function pointer. Hence my approach with lambda. I will look to see if there's a different way of doing this.

capital-G:
> Hence my approach with lambda. I will look to see if there's a different way of doing this.

Just use a lambda at this point, this is totally fine - maybe add a comment about that the function pointer doesn't preserve default arguments, didn't know about this as well.

stefandiaconu:
> > Hence my approach with lambda. I will look to see if there's a different way of doing this.
> 
> Just use a lambda at this point, this is totally fine - maybe add a comment about that the function pointer doesn't preserve default arguments, didn't know about this as well.

Cool. Will push something tonight 👍

PR #7416, opened 2026-03-15
Rename good first issues to good-first-issues

capital-G:
In order to avoid tracking of *good first issues* from outside resources, I renamed it to *good-first-issues*.

This requires also some modification on the README.md

## Purpose and Motivation

See forum chat

## Types of changes

- Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Read our [contributing guide](https://github.com/supercollider/supercollider/wik
 
 ### Good First Issues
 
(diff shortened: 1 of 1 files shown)


dyfer (approved):
I haven't looked at the forum but I think the change is fine.
Would we need to do anything else repo-wise?
Also, how does this "avoid tracking" from outside?

telephon:
It just reduces visibility, because the old one was gighubwide default.

dyfer:
Ah, got it.

PR #7418, opened 2026-03-18
Convolution UGen fixes

Spacechild1:
## Purpose and Motivation

The Convolution UGens wrongly used `uint32` for buffer numbers, which caused issues with negative numbers.

This PR makes sure that all buffer numbers are signed. It also fixes the logic in the `ConvGetBuffer()` helper function to account for negative numbers.

Finally, it fixes the `test_allUGens_allocFail` test method of `TestUGen_RTAlloc` by passing an actual valid Buffer instead of -1 to the Convolution UGens. Otherwise the UGens constructors would return before attempting to allocate any RT memory.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/server/plugins/Convolution.cpp
+++ b/server/plugins/Convolution.cpp
@@ -244,22 +244,31 @@ void Convolution_next(Convolution* unit, int numSamples) {
 
 
 // include local buffer test in one place
-static SndBuf* ConvGetBuffer(Unit* unit, uint32 bufnum, const char* ugenName, int inNumSamples) {
+// TODO: reevaluate the error handling behavior. Should the Unit really stop processing just
+// because it (temporarily) received an invalid buffer number? Probably not.
+static SndBuf* ConvGetBuffer(Unit* unit, int32 bufnum, const char* ugenName, int inNumSamples) {
     SndBuf* buf;
     World* world = unit->mWorld;
 
-    if (bufnum >= world->mNumSndBufs) {
-        int localBufNum = bufnum - world->mNumSndBufs;
-        Graph* parent = unit->mParent;
-        if (localBufNum <= parent->localMaxBufNum) {
-            buf = parent->mLocalSndBufs + localBufNum;
+    // first check if bufnum is positive!
+    if (bufnum >= 0) {
+        if (bufnum >= world->mNumSndBufs) {
+            int localBufNum = bufnum - world->mNumSndBufs;
+            Graph* parent = unit->mParent;
+            if (localBufNum <= parent->localMaxBufNum) {
+                buf = parent->mLocalSndBufs + localBufNum;
+            } else {
+                if (unit->mWorld->mVerbosity > -1)
+                    Print("%s: invalid buffer number (%d).\n", ugenName, bufnum);
+                goto handle_failure;
+            }
         } else {
-            if (unit->mWorld->mVerbosity > -1)
-                Print("%s: invalid buffer number (%d).\n", ugenName, bufnum);
-            goto handle_failure;
+            buf = world->mSndBufs + bufnum;
         }
     } else {
-        buf = world->mSndBufs + bufnum;
+        if (unit->mWorld->mVerbosity > -1)
+            Print("%s: invalid buffer number (%d).\n", ugenName, bufnum);
+        goto handle_failure;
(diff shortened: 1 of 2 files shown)


server/plugins/Convolution.cpp
```diff
@@ -1074,11 +1077,12 @@ void Convolution3_next_a(Convolution3* unit) {
 
     if (unit->m_prevtrig <= 0.f && curtrig > 0.f) {
         uint32 framesize_f = unit->m_framesize * sizeof(float);
-        float fbufnum = ZIN0(1);
-        uint32 bufnum = (int)fbufnum;
+        int32 bufnum = (int32)ZIN0(1);
         SndBuf* buf = ConvGetBuffer(unit, bufnum, "Convolution3", numSamples);
-        LOCK_SNDBUF_SHARED(buf);
-        memcpy(unit->m_inbuf2, buf->data, framesize_f);
+        if (buf) {
```
JordanHendersonMusic: Should this do anything if it fails?

server/plugins/Convolution.cpp
```diff
@@ -551,7 +559,7 @@ void Convolution2L_next(Convolution2L* unit, int numSamples) {
     unit->m_pos += numSamples;
 
     if (unit->m_prevtrig <= 0.f && curtrig > 0.f) {
-        uint32 bufnum = (int)ZIN0(1);
```
JordanHendersonMusic, line 554: I'm always a bit concerned by c casts, particularly when combined with macros.

Are we sure this won't try to do something silly, like cast a float* to an int on 32 bit systems? I.e., taking the address as an int?

server/plugins/Convolution.cpp
Spacechild1: Good catch! Actually, I can just return because `ConvGetBuffer` basically deactivates the Unit. This would be in line with similar places. I've updated the commit accordingly.

server/plugins/Convolution.cpp
Spacechild1: In general, you have a point about using c-style casts with macro functions where you don't immediately see the return type. 

In this case, we know that `ZIN0` returns a float and not a `float*`. Otherwise, the thing would crash every single time. There is no real potential for a subtle bug here.

JordanHendersonMusic: Okay cool, if you're happy that can't happen then so am I!

JordanHendersonMusic:
This all looks good, and is a welcome fix, hopefully this stops these from failing in the ci intermittently!

Are you intending to do more on this? If not, I'll approve and this can be merged.

Spacechild1:
> Are you intending to do more on this? If not, I'll approve and this can be merged.

For now I just wanted to fix the CI.

There are some sketchy things in these UGens. For example, the error handling in `GetConvBuf` doesn't really makes sense to me. (Why do we want to stop the whole UGen just because we temporarily got a bad buffer number?) I've added a comment, hoping that someone will eventually fix this in the future :)

server/plugins/Convolution.cpp
```diff
@@ -1117,16 +1125,13 @@ void Convolution3_next_k(Convolution3* unit) {
 
     uint32 framesize_f = unit->m_framesize * sizeof(float);
 
-
     if (unit->m_prevtrig <= 0.f && curtrig > 0.f) {
-        float fbufnum = ZIN0(1);
-        uint32 bufnum = (int)fbufnum;
+        int32 bufnum = (int32)ZIN0(1);
         SndBuf* buf = ConvGetBuffer(unit, bufnum, "Convolution3", 1);
-        if (!buf)
```
JordanHendersonMusic, line 1125: The audio rate one does the return, why not here?

JordanHendersonMusic (changes requested):

server/plugins/Convolution.cpp
Spacechild1: Thanks! I missed that one. Done!

JordanHendersonMusic (approved):
Merge once tests pass, thanks for this!

Spacechild1:
Thanks for the quick review!

Issue #6736, opened 2025-04-16
Window Bounds Inconsistency: Wrong top/bottom Argument Name for Windows/macOS; Incremental Offset in Linux


prko:
## Environment

* SuperCollider version: 3.14_dev
* Operating system: Linux

## Steps to reproduce

```supercollider
(
Window("", Rect(0, 0, 200, 200)).front;
Window("", Rect(200, 200, 200, 200)).front;
Window("", Rect(200, 200, 150, 150)).front
)
```
Window Bounds Inconsistency: wrong top/bottom argument name for Win/Mac; Offset on Linux
### Expected: 
The behavior should remain consistent across all platforms.

### Actual: 
The results in Windows and macOS show slight differences, which are understandable due to platform-specific nuances. However, the behavior observed in Linux cannot be considered acceptable for the following three reasons:

1. In Windows and macOS, the `top` argument in `Rect(left, top, width, height)` for the `Window` class does not function correctly and instead represents the `bottom` value.  
2. On Linux, the `top` argument in `Rect(left, top, width, height)` works as expected and refers to the correct value.  
3. In Linux, repeated execution of the same code leads to an incremental addition of offset values in both horizontal and vertical dimensions of the window bounds.  

While discrepancies between Windows and macOS can be attributed to operating system differences, the behavior on Linux indicates a unique bug. Specifically, the incremental offset introduced upon each window creation requires immediate attention.

For further clarification, kindly refer to the attached video demonstration:
https://www.dropbox.com/scl/fi/ixr30ncpsjdys6ht9ipds/Screen-Recording-2025-04-16-at-10.56.33.mov?rlkey=i8tvpfhyyc24c8eu53qaib3l5&st=b1camynn&dl=0

prko:
SuperCollider 3.13.0 on Fedora also has strange, inconsistent behaviour in this regard:
https://www.dropbox.com/scl/fi/jd73wzhhxohrhs38xe46h/Screen-Recording-2025-04-16-at-22.31.37.mov?rlkey=cr8kkiwo0dmv0szwx3wcxnh4b&st=1t79rosz&dl=0

SimonDeplat:
I can't reproduce the issue on Ubuntu 22.04 LTS (KDE, _not_ using Wayland).

First, the `top` value refers to the bottom of the screen in my case, so `Window("", Rect(0, 0, 200, 200)).front;` appears in the bottom left corner of the screen. Is that what you different from what you mentioned ?

[...]

PR #7421, opened 2026-03-20
Helpbrowser: redirect http links to system browser

HotwheelsSisyphus:
Clicking on web links now opens a menu that gives user the option to open the link in either system default web browser or the web-unsafe built-in help browser.
I made sure this works both in the SCIDE helpbrowser and the one created through HelpBrowser.new(). 

added classvars:
redirectEnabled: whether to open that menu at all
redirectPattern: the regex pattern that determines what gets redirected.

added methods:
prShouldRedirect (a test whether to redirect at all)
prRedirectMenu (code for the actual menu)
prIDEorHelpBrowser (we need slightly different behavior depending on whether you're working with the IDE help browser or an instance of HelpBrowser).

code is tested and documented.

<details><summary>out of date or resolved</summary>
<p>

add *redirectToSystemBrowser class method to HelpBrowser class, 
which is used in *goTo (for IDE Helpbrowser) and 
webView.onLinkActivated in HelpBrowser.init.

Tested both and they work, for all I can tell. 
What they don't (and probably can't) do is bring the system browser to front, 
so instead they post to the post window to let the user know that the web site has indeed been opened. 
I wonder if there's an easy way to make this message a different color, or make it bold? I could see the message getting overlooked, and people not realizing that the site has opened in another application. 
(Otoh, this is neither a warning nor an error, so I don't want to use either .warn or .error)

@prko suggested that the forwarding to system browser should not happen with loopback addresses. For now I haven't added this (basically just another test), because I'm not sure why one would use this in the helpbrowser (I had to look up what loopback addresses even are). But no problem to add that if someone gives me an example.

The .replaceRegex to ensure https may be unnecessary, I can imagine that the scdoc parser already does something similar... 

What we don't have with this PR is a way to force opening websites in the helpbrowser (ide or not), say by ctrl-clicking or something. This isn't something I know how to do off the bat, but probably we don't need it anyway?

</p>
</details>

### DIFF
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -34,7 +34,7 @@ table::
 ::
 
 classmethods::
-private:: getOldWrapUrl, initClass, prRedirectToSystemBrowser
+private:: getOldWrapUrl, initClass, prShouldRedirect, prRedirectMenu, prIDEorHelpBrowser
 
 method:: instance
 The singleton HelpBrowser instance.
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -38,28 +38,36 @@ private:: getOldWrapUrl, initClass, prShouldRedirect, prRedirectMenu, prIDEorHel
 
 method:: instance
 The singleton HelpBrowser instance.
+emphasis::Does not concern the SCIDE's built-in helpbrowser.::
 
 method:: new
 Create a new HelpBrowser instance with given home URL.
+emphasis::Does not concern the SCIDE's built-in helpbrowser.::
 
 method:: defaultHomeUrl
 Get or set the default home URL.
 
 method:: openNewWindows
+emphasis::Does not concern the SCIDE's built-in helpbrowser.::
 Get or set the default for "open in new windows" toggle.
 When set to teletype::false:: (default), this class behaves like a singleton class.
 
 method:: redirectEnabled
-When set to true (default), redirects links that match code::redirectPattern:: to the system's default Web browser.
-Disable at your own risk: SC's help browser is not web safe.
+When set to teletype::true:: (default), clicking on links whose target matches code::redirectPattern:: will open a menu giving the user the choice to open the link in the system's default Web browser (recommended),
+or in SuperCollider's own help browser (strong::not:: recommended for remote links, because not safe for web browsing).
+warning::Disable at your own risk: SC's help browser is not web safe.
+
+This concerns both the SCIDE's built-in help browser and help browser instances created
+through code::HelpBrowser.new()::.::
 
 method:: redirectPattern
-Get or set the regex pattern (a link::Classes/String#Regular expressions#String::) that determines which links are forwarded to the system's default browser.
-When the link matches this pattern, it will be opened in the system's default browser (provided code::redirectEnabled:: is teletype::true::.)
-By default, this is set to code::"^http":: (i.e., it redirects all links that begin with teletype::http::).
+Get or set the regex pattern (a link::Classes/String#Regular expressions#String::) that determines which links will open the menu described above (link::#*redirectEnabled::).
+By default, this is set to code::"^http":: (i.e., it opens the menu all links that begin with teletype::http::).
+To exclude addresses such as teletype::localhost:: from this behavior, append a link::https://en.wikipedia.org/wiki/Regular_expression#Assertions#negative lookahead::, e.g.: code::"^https?:\/\/(?!localhost|127\..*)"::.
 
 method:: goTo
-Go to url with singleton instance or a new window, depending on the code::openNewWindows:: setting.
+When using the SCIDE, go to teletype::url:: in the built-in helpbrowser;
+otherwise, go to url with singleton code::HelpBrowser:: instance or a new window, depending on the code::openNewWindows:: setting.
 
 method:: openHelpFor
 Open the relevant help page for given text in the singleton HelpBrowser instance.
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -34,7 +34,7 @@ table::
 ::
 
 classmethods::
-private:: getOldWrapUrl, initClass
+private:: getOldWrapUrl, initClass, prRedirectToSystemBrowser
 
 method:: instance
 The singleton HelpBrowser instance.
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -34,7 +34,7 @@ table::
 ::
 
 classmethods::
-private:: getOldWrapUrl, initClass, prShouldRedirect, prRedirectMenu, prIDEorHelpBrowser
+private:: getOldWrapUrl, initClass, init, prShouldRedirect, prOpenRedirectMenu, prOpenInIDEorHelpBrowser
 
 method:: instance
 The singleton HelpBrowser instance.
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -48,6 +48,15 @@ Get or set the default home URL.
 method:: openNewWindows
 Get or set the default for "open in new windows" toggle.
 
+method:: redirectEnabled
+When set to true (default), redirects links that match code::redirectPattern:: to the system's default Web browser.
+Disable at your own risk: SC's help browser is not web safe.
+
+method:: redirectPattern
+Get or set the regex pattern (a link::Classes/String#Regular expressions#String::) that determines which links are forwarded to the system's default browser.
+When the link matches this pattern, it will be opened in the system's default browser (provided code::redirectEnabled:: is teletype::true::.)
+By default, this is set to code::"^http":: (i.e., it redirects all links that begin with teletype::http::).
+
 method:: goTo
 Go to url with singleton instance or a new window, depending on the code::openNewWindows:: setting.
 
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -47,6 +47,7 @@ Get or set the default home URL.
 
 method:: openNewWindows
 Get or set the default for "open in new windows" toggle.
+When set to teletype::false:: (default), this class behaves like a singleton class.
 
 method:: redirectEnabled
 When set to true (default), redirects links that match code::redirectPattern:: to the system's default Web browser.
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -45,7 +45,7 @@ HelpBrowser {
 	}
 
 	*prRedirectMenu { |url, openInSCIDE|
-		var info = MenuAction("Choose how to open external link:\n").enabled_(false);
+		var info = MenuAction("Choose how to open external link:").enabled_(false);
 		var domain = MenuAction(if(url.size < 75) {url} {url.replaceRegexp("https?:\/\/|(/.*)", "")} ).enabled_(false);
 		var separator = MenuAction().separator_(true);
 		var systemDefaultBrowser = MenuAction("Open in system default browser");
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -4,6 +4,10 @@ HelpBrowser {
 	classvar <>openNewWindows = false;
 	classvar <>scrollStep = 40;
 	classvar <>scrollPageStep = 350;
+	classvar <>redirectEnabled = true;
+	// redirect link addresses that match redirectPattern to default system browser
+	classvar <>redirectPattern = "^http";
+	// all http gets redirected
 
 	var <>homeUrl;
 	var <window;
@@ -38,11 +42,10 @@ HelpBrowser {
 
 
 	*prRedirectToSystemBrowser { |url|
-			var redirectPattern = "^https?:\/\/(?!127.*|localhost)";
 			url = url.asString;
-			// matches all http except for loopback/localhost (IPv4)
-			if(redirectPattern.matchRegexp(url)) {
+			if(redirectEnabled and: { redirectPattern.matchRegexp(url) }) {
 				url.openOS;
+				(url ++ " opened in system browser").postln;
 				^true
 			} {
 				^false
@@ -53,7 +56,7 @@ HelpBrowser {
 		var ideClass = \ScIDE.asClass;
 		var redirected = HelpBrowser.prRedirectToSystemBrowser(url);
 		case
-		{ redirected } { (url ++ " opened in system browser").postln; }
+		{ redirected } { ^nil }
 		{ ideClass.notNil } { ideClass.openHelpUrl(url) }
 		{ this.front.goTo(url) };
 	}
@@ -253,10 +256,8 @@ HelpBrowser {
 
 		webView.onLinkActivated = {|wv, url|
 			var redirected, newPath, oldPath;
-			case
-			{ HelpBrowser.prRedirectToSystemBrowser(url) } { (url ++ " opened in system browser").postln; }
-			{ this.redirectTextFile(url) } { ^nil }
-			{
+			redirected = this.redirectTextFile(url) or: { HelpBrowser.prRedirectToSystemBrowser(url) };
+			if (not(redirected)) {
 				#newPath, oldPath = [url,webView.url].collect {|x|
 					if(x.notEmpty) {x.findRegexp("(^\\w+://)?([^#]+)(#.*)?")[1..].flop[1][1]}
 				};
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -38,10 +38,10 @@ HelpBrowser {
 
 
 	*prRedirectToSystemBrowser { |url|
+			var redirectPattern = "^https?:\/\/(?!127.*|localhost)";
 			url = url.asString;
-			redirectPattern = "^https?:\/\/(?!127.*|localhost)";
 			// matches all http except for loopback/localhost (IPv4)
-			if(redirectPattern.matchRegex(url)) {
+			if(redirectPattern.matchRegexp(url)) {
 				url.openOS;
 				^true
 			} {
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -36,13 +36,24 @@ HelpBrowser {
 		^super.new.init( aHomeUrl ? defaultHomeUrl, newWin ?? { openNewWindows } );
 	}
 
+
+	*redirectToSystemBrowser { |url|
+			if(url.contains("http")) {
+				url.replaceRegexp("https?", "https").openOS;
+				(url ++ " opened in system browser").postln;
+				^true
+			} {
+				^false
+			}
+	}
+
 	*goTo {|url|
 		var ideClass = \ScIDE.asClass;
-		if ( ideClass.notNil ) {
-			ideClass.openHelpUrl(url);
-		}{
-			this.front.goTo(url);
-		}
+		var redirected = HelpBrowser.redirectToSystemBrowser(url);
+		case
+		{ redirected } { ^nil }
+		{ ideClass.notNil } { ideClass.openHelpUrl(url) }
+		{ this.front.goTo(url) };
 	}
 
 	*front {
@@ -240,7 +251,7 @@ HelpBrowser {
 
 		webView.onLinkActivated = {|wv, url|
 			var redirected, newPath, oldPath;
-			redirected = this.redirectTextFile(url);
+			redirected = this.redirectTextFile(url) or: { HelpBrowser.redirectToSystemBrowser(url) };
 			if (not(redirected)) {
 				#newPath, oldPath = [url,webView.url].collect {|x|
 					if(x.notEmpty) {x.findRegexp("(^\\w+://)?([^#]+)(#.*)?")[1..].flop[1][1]}
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -4,10 +4,10 @@ HelpBrowser {
 	classvar <>openNewWindows = false;
 	classvar <>scrollStep = 40;
 	classvar <>scrollPageStep = 350;
-	classvar <>redirectEnabled = true;
+
 	// redirect link addresses that match redirectPattern to default system browser
+	classvar <>redirectEnabled = true;
 	classvar <>redirectPattern = "^http";
-	// all http gets redirected
 
 	var <>homeUrl;
 	var <window;
@@ -44,7 +44,7 @@ HelpBrowser {
 		^(redirectEnabled and: { redirectPattern.matchRegexp(url) })
 	}
 
-	*prRedirectMenu { |url, openInSCIDE|
+	*prOpenRedirectMenu { |url, useSCIDE|
 		var info = MenuAction("Choose how to open external link:").enabled_(false);
 		var domain = MenuAction(if(url.size < 75) {url} {url.replaceRegexp("https?:\/\/|(/.*)", "")} ).enabled_(false);
 		var separator = MenuAction().separator_(true);
@@ -54,15 +54,14 @@ HelpBrowser {
 		var menu = Menu(info, domain, separator, systemDefaultBrowser, helpBrowser, closeMenu);
 		var removeDependants = { [systemDefaultBrowser, helpBrowser, closeMenu].do(_.removeDependant)  };
 		systemDefaultBrowser.addDependant { |action, what| if (what == \triggered) { url.openOS; removeDependants.() } };
-		helpBrowser.addDependant { |action, what| if (what == \triggered) { this.prIDEorHelpBrowser(url, openInSCIDE); removeDependants.() } };
+		helpBrowser.addDependant { |action, what| if (what == \triggered) { this.prOpenInIDEorHelpBrowser(url, useSCIDE); removeDependants.() } };
 		^menu.front;
 	}
 
-	*prIDEorHelpBrowser { | url, openInSCIDE |
-		// when openInSCIDE = false, will open in a HelpBrowser instance.
+	*prOpenInIDEorHelpBrowser { | url, useSCIDE(true) |
+		// when useSCIDE = false, will open in a HelpBrowser instance.
 		var ideClass = \ScIDE.asClass;
-		openInSCIDE ?? { openInSCIDE = true };
-		if(ideClass.notNil and: openInSCIDE) {
+		if(ideClass.notNil and: useSCIDE) {
 			ideClass.openHelpUrl(url)
 		} {
 			this.front.goTo(url)
@@ -71,9 +70,9 @@ HelpBrowser {
 
 	*goTo {|url|
 		if(this.prShouldRedirect(url)) {
-			this.prRedirectMenu(url)
+			this.prOpenRedirectMenu(url)
 		} {
-			this.prIDEorHelpBrowser(url)
+			this.prOpenInIDEorHelpBrowser(url)
 		}
 	}
 
@@ -275,7 +274,7 @@ HelpBrowser {
 			var newPath, oldPath;
 			if( not(this.redirectTextFile(url)) ) {
 				if( HelpBrowser.prShouldRedirect(url)) {
-					HelpBrowser.prRedirectMenu(url, openInSCIDE: false)
+					HelpBrowser.prOpenRedirectMenu(url, useSCIDE: false);
 				} {
 					#newPath, oldPath = [url,webView.url].collect {|x|
 					if(x.notEmpty) {x.findRegexp("(^\\w+://)?([^#]+)(#.*)?")[1..].flop[1][1]}
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -37,10 +37,12 @@ HelpBrowser {
 	}
 
 
-	*redirectToSystemBrowser { |url|
-			if(url.contains("http")) {
-				url.replaceRegexp("https?", "https").openOS;
-				(url ++ " opened in system browser").postln;
+	*prRedirectToSystemBrowser { |url|
+			url = url.asString;
+			redirectPattern = "^https?:\/\/(?!127.*|localhost)";
+			// matches all http except for loopback/localhost (IPv4)
+			if(redirectPattern.matchRegex(url)) {
+				url.openOS;
 				^true
 			} {
 				^false
@@ -49,9 +51,9 @@ HelpBrowser {
 
 	*goTo {|url|
 		var ideClass = \ScIDE.asClass;
-		var redirected = HelpBrowser.redirectToSystemBrowser(url);
+		var redirected = HelpBrowser.prRedirectToSystemBrowser(url);
 		case
-		{ redirected } { ^nil }
+		{ redirected } { (url ++ " opened in system browser").postln; }
 		{ ideClass.notNil } { ideClass.openHelpUrl(url) }
 		{ this.front.goTo(url) };
 	}
@@ -251,8 +253,10 @@ HelpBrowser {
 
 		webView.onLinkActivated = {|wv, url|
 			var redirected, newPath, oldPath;
-			redirected = this.redirectTextFile(url) or: { HelpBrowser.redirectToSystemBrowser(url) };
-			if (not(redirected)) {
+			case
+			{ HelpBrowser.prRedirectToSystemBrowser(url) } { (url ++ " opened in system browser").postln; }
+			{ this.redirectTextFile(url) } { ^nil }
+			{
 				#newPath, oldPath = [url,webView.url].collect {|x|
 					if(x.notEmpty) {x.findRegexp("(^\\w+://)?([^#]+)(#.*)?")[1..].flop[1][1]}
 				};
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -45,19 +45,23 @@ HelpBrowser {
 	}
 
 	*prRedirectMenu { |url, openInSCIDE|
+		var info = MenuAction("Choose how to open external link:\n").enabled_(false);
+		var domain = MenuAction(if(url.size < 75) {url} {url.replaceRegexp("https?:\/\/|(/.*)", "")} ).enabled_(false);
+		var separator = MenuAction().separator_(true);
 		var systemDefaultBrowser = MenuAction("Open in system default browser");
-		var helpBrowser = MenuAction("Open in sclang HelpBrowser (unsafe)");
-		var closeMenu = MenuAction("Do not open");
-		var menu = Menu(systemDefaultBrowser, helpBrowser, closeMenu);
+		var helpBrowser = MenuAction("Open in SuperCollider's help browser (unsafe)");
+		var closeMenu = MenuAction("Cancel");
+		var menu = Menu(info, domain, separator, systemDefaultBrowser, helpBrowser, closeMenu);
 		var removeDependants = { [systemDefaultBrowser, helpBrowser, closeMenu].do(_.removeDependant)  };
 		systemDefaultBrowser.addDependant { |action, what| if (what == \triggered) { url.openOS; removeDependants.() } };
 		helpBrowser.addDependant { |action, what| if (what == \triggered) { this.prIDEorHelpBrowser(url, openInSCIDE); removeDependants.() } };
 		^menu.front;
 	}
 
-	*prIDEorHelpBrowser { | url, openInSCIDE = true |
+	*prIDEorHelpBrowser { | url, openInSCIDE |
 		// when openInSCIDE = false, will open in a HelpBrowser instance.
 		var ideClass = \ScIDE.asClass;
+		openInSCIDE ?? { openInSCIDE = true };
 		if(ideClass.notNil and: openInSCIDE) {
 			ideClass.openHelpUrl(url)
 		} {
--- a/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
+++ b/SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
@@ -40,25 +40,37 @@ HelpBrowser {
 		^super.new.init( aHomeUrl ? defaultHomeUrl, newWin ?? { openNewWindows } );
 	}
 
+	*prShouldRedirect {|url|
+		^(redirectEnabled and: { redirectPattern.matchRegexp(url) })
+	}
 
-	*prRedirectToSystemBrowser { |url|
-			url = url.asString;
-			if(redirectEnabled and: { redirectPattern.matchRegexp(url) }) {
-				url.openOS;
-				(url ++ " opened in system browser").postln;
-				^true
-			} {
-				^false
-			}
+	*prRedirectMenu { |url, openInSCIDE|
+		var systemDefaultBrowser = MenuAction("Open in system default browser");
+		var helpBrowser = MenuAction("Open in sclang HelpBrowser (unsafe)");
+		var closeMenu = MenuAction("Do not open");
+		var menu = Menu(systemDefaultBrowser, helpBrowser, closeMenu);
+		var removeDependants = { [systemDefaultBrowser, helpBrowser, closeMenu].do(_.removeDependant)  };
+		systemDefaultBrowser.addDependant { |action, what| if (what == \triggered) { url.openOS; removeDependants.() } };
+		helpBrowser.addDependant { |action, what| if (what == \triggered) { this.prIDEorHelpBrowser(url, openInSCIDE); removeDependants.() } };
+		^menu.front;
+	}
+
+	*prIDEorHelpBrowser { | url, openInSCIDE = true |
+		// when openInSCIDE = false, will open in a HelpBrowser instance.
+		var ideClass = \ScIDE.asClass;
+		if(ideClass.notNil and: openInSCIDE) {
+			ideClass.openHelpUrl(url)
+		} {
+			this.front.goTo(url)
+		}
 	}
 
 	*goTo {|url|
-		var ideClass = \ScIDE.asClass;
-		var redirected = HelpBrowser.prRedirectToSystemBrowser(url);
-		case
-		{ redirected } { ^nil }
-		{ ideClass.notNil } { ideClass.openHelpUrl(url) }
-		{ this.front.goTo(url) };
+		if(this.prShouldRedirect(url)) {
+			this.prRedirectMenu(url)
+		} {
+			this.prIDEorHelpBrowser(url)
+		}
 	}
 
 	*front {
@@ -153,6 +165,7 @@ HelpBrowser {
 		var x, y, w, h;
 		var str;
 
+		if (not(openNewWindows)) { singleton = this };
 		homeUrl = aHomeUrl;
 
 		winRect = Rect(0, 0, 800, (Window.screenBounds.height * 0.8).floor);
@@ -255,18 +268,21 @@ HelpBrowser {
 		webView.onLoadFailed = { this.stopAnim };
 
 		webView.onLinkActivated = {|wv, url|
-			var redirected, newPath, oldPath;
-			redirected = this.redirectTextFile(url) or: { HelpBrowser.prRedirectToSystemBrowser(url) };
-			if (not(redirected)) {
-				#newPath, oldPath = [url,webView.url].collect {|x|
-					if(x.notEmpty) {x.findRegexp("(^\\w+://)?([^#]+)(#.*)?")[1..].flop[1][1]}
-				};
-
-				if(newPath != oldPath && openNewWin) {
-						HelpBrowser.new(newWin:openNewWin).goTo(url);
+			var newPath, oldPath;
+			if( not(this.redirectTextFile(url)) ) {
+				if( HelpBrowser.prShouldRedirect(url)) {
+					HelpBrowser.prRedirectMenu(url, openInSCIDE: false)
 				} {
-					this.goTo(url);
-				};
+					#newPath, oldPath = [url,webView.url].collect {|x|
+					if(x.notEmpty) {x.findRegexp("(^\\w+://)?([^#]+)(#.*)?")[1..].flop[1][1]}
+					};
+			
+					if(newPath != oldPath && openNewWin) {
+							HelpBrowser.new(newWin:openNewWin).goTo(url);
+					} {
+						this.goTo(url);
+					}
+				}
 			}
 		};
 


SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,13 +36,24 @@ HelpBrowser {
 		^super.new.init( aHomeUrl ? defaultHomeUrl, newWin ?? { openNewWindows } );
 	}
 
+
+	*redirectToSystemBrowser { |url|
+			if(url.contains("http")) {
```
capital-G: imo contains is too broad - we are only intrested in the beginning - otherwise a class containing `http` may also get redirected to the outside

```suggestion
			if(url.asString.beginsWith("http")) {
```

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,13 +36,24 @@ HelpBrowser {
...
+				url.replaceRegexp("https?", "https").openOS;
```
capital-G: You can't just add a `s` to `http` and expect it to work as http and https are two distinct services. It is responsibility of the server to redirect a http request to a https one these days, but we can't assume that every server has https support.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,13 +36,24 @@ HelpBrowser {
...
+				(url ++ " opened in system browser").postln;
```
capital-G: imo this method shouldn't print as it isn't responsible for this action. printing of an action should happen where the action takes place :)

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,13 +36,24 @@ HelpBrowser {
 		^super.new.init( aHomeUrl ? defaultHomeUrl, newWin ?? { openNewWindows } );
 	}
 
+
+	*redirectToSystemBrowser { |url|
```
capital-G: consider making this a private method so we can break it again in the future as it shouldn't be used by the user directly.

capital-G (commented):
Some comments to consider

prko:
>@prko suggested that the forwarding to system browser should not happen with loopback addresses. For now I haven't added this (basically just another test), because I'm not sure why one would use this in the helpbrowser (I had to look up what loopback addresses even are). But no problem to add that if someone gives me an example.

One concrete use case for keeping loopback addresses inside the HelpBrowser is local content rendering.

I am planning to publish my Quark (https://github.com/prko/Notator) in 2026 or 2027, and one of its goals is to display MusicXML files directly inside SuperCollider. This is possible via a WebView, and the HelpBrowser could also be used for this purpose. In such a setup, a local HTTP server is required, because at the moment there is no way for the HelpBrowser to access local files directly. In this case, forwarding loopback addresses (e.g. localhost or 127.0.0.1) to the system browser would break the intended workflow.

More generally, I would also argue that sccode.org should ideally be accessible within the HelpBrowser. This would allow users to browse and evaluate SuperCollider code directly, without copying and pasting snippets into the editor, which aligns well with the HelpBrowser’s role as an interactive learning and exploration environment.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: ah right!
but since prko asked for loopback not to be redirected, I figured I'd use a regex for this.  Although this might no be very readable:

```
                       url = url.asString;
			redirectPattern = "^https?:\/\/(?!127.*|localhost)";
			// matches all http except for loopback/localhost (IPv4)
			if(redirectPattern.matchRegex(url)) {
                       }
```
I'm a bit unsure about the :// part, maybe there are other ways to write that separator that I am not aware of.

capital-G:
Thanks @prko - haven't thought about that. The current implementation would not allow to access any external resources anymore - maybe we should add a keyword argument called something along `allowExternalBrowser: false` (probably w/ better naming?), which for calls from the Qt-IDE itself (so, when clicking on a link) is set to true, but it wouldn't affect existing code in any way.

HotwheelsSisyphus:
I should note that on my linux, String.openOS throws some errors in the post window (though it still does what it's supposed to do in this case, namely open in system browser). I assume this has nothing to do with SC and more with some absolutely imponderable combinations of settings, DE, window managers, etc. in linux. (The bit about locale is probably because mine is set to french because I'm silly...). 
I get some sort of "error" or warning for just about every receiver (different filetypes, paths...) for .openOS, but it still does open as intended.
<details><summary>example error</summary>
<p>

```
/usr/bin/flatpak run:206110): dconf-WARNING **: 16:11:36.726: unable to open file '/etc/dconf/db/site': Failed to open file “/etc/dconf/db/site”: open() failed: No such file or directory; expect degraded performance
ERROR: 
(process:2): Gtk-WARNING **: 16:11:36.796: Locale not supported by C library.
	Using the fallback 'C' locale.
ERROR: 
(cobalt:2): Gdk-WARNING **: 16:11:36.800: Failed to read portal settings: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface ?org.freedesktop.portal.Settings? on object at path /org/freedesktop/portal/desktop

(/usr/bin/flatpak run:206152): dconf-WARNING **: 16:11:36.927: unable to open file '/etc/dconf/db/site': Failed to open file “/etc/dconf/db/site”: open() failed: No such file or directory; expect degraded performance
Ouverture dans une session de navigateur existante.
```

</p>
</details> 


But just to be safe, maybe someone can check what happens on Windows/MacOS? (No need to download this PR branch, this should be a problem with String.openOS, if there is a problem at all).

prko:
@HotwheelsSisyphus  
Could you please share the SuperCollider code that produces those errors?

In the past, there were issues related to certain special characters in strings passed to `.openOS`. Those characters could behave differently depending on the operating system, which meant that some websites could not be opened consistently across platforms. As far as I understand, this issue has since been resolved. However, I am still having difficulty understanding the error examples you mentioned, and I am not sure which specific code paths are producing them.

For additional context, you can also find several closed issues related to `.openOS` in the SuperCollider GitHub issue tracker by searching for `is:issue state:closed openOS`. I previously reported one such issue myself, which was related to OS‑dependent handling of certain characters in URLs.

@capital-G  
Thank you for your consideration.

In addition, I would like to suggest that the following SuperCollider‑related sites could reasonably be opened within the built‑in HelpBrowser rather than being forwarded to the system browser:

- https://scsynth.org/
- https://supercollider.github.io/examples
- https://docs.supercollider.online/
- https://github.com/supercollider/supercollider
- https://www.youtube.com/@SuperColliderOnline

Allowing these sites to remain inside the HelpBrowser would make it more convenient for users to explore SuperCollider‑specific resources. In particular, it would enable users to evaluate code examples directly within the HelpBrowser, without needing to copy and paste them into the editor, which aligns well with the HelpBrowser’s role as an interactive learning and exploration environment.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,13 +36,26 @@ HelpBrowser {
 		^super.new.init( aHomeUrl ? defaultHomeUrl, newWin ?? { openNewWindows } );
 	}
 
+
+	*prRedirectToSystemBrowser { |url|
+			var redirectPattern = "^https?:\/\/(?!127.*|localhost)";
```
prko: `var redirectPattern = "^https?:\/\/(?!127.*|localhost)(?!.*(?:scsynth|[Ss]uper[Cc]ollider|sccode|sc140)(?!.*[Dd]ownload))";`

How about this? This resolves the issue discussed in
https://github.com/supercollider/supercollider/pull/7421#issuecomment-4101046018.
It also allows web pages containing `Download` or `download` to be opened in the system’s default web browser.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,13 +36,26 @@ HelpBrowser {
...
+			url = url.asString;
+			// matches all http except for loopback/localhost (IPv4)
+			if(redirectPattern.matchRegexp(url)) {
+				url.openOS;
```
capital-G: IMO the printing of `(url ++ " opened in system browser").postln;` should happen here

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
capital-G: IMO this is not working - i.e.

```supercollider
"^https?:\/\/(?!127.*|localhost)".matchRegexp("http://127.0.0.1");
-> false
```

I think it is simpler if you invert the truth value of the regex and avoid negative lookahead - and also think of matching the `file` prefix

```
^(https?:\/\/((127\.0\.0\.1)|localhost)|file).*
```

capital-G (commented):
> * https://scsynth.org/
> * https://supercollider.github.io/examples
> * https://docs.supercollider.online/
> * https://github.com/supercollider/supercollider
> * https://www.youtube.com/@SuperColliderOnline

I strongly disagree w/ allowing some websites implicitly - the web browser is essentially always outdated and provides a security risk for the user when using it for the web. IMO we have some responsibility for the user here. JS exploits are a thing and WebKit is heavily under attack, being an entrance to closed systems such as iOS, PlayStation, Switch, etc.

Although - I think one thing we could do is make this configurable by the user by making the regex a documented `classvar` which defaults to the present regex and can be adapted in e.g. a `startup` file. This would allow users to accept a risk explicitly, but protect all other users.

Additionally we should still allow the user to navigate the  HelpBrowser to a URL in the web if explicitly set (at least I sometimes use this to display a QR code during live coding performances^^).
Maybe we should wrap all the checking by a classvar boolean called something like `redirectExternalResourcesToSystemBrowser` (please don't use this long name....), which defaults to `true` but can be set to `false` to allow any kind of website access through the help browser.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: > not working - i.e.
> ```
> "^https?:\/\/(?!127.*|localhost)".matchRegexp("http://127.0.0.1");
> -> false
> ```

This is the intended result though - it should _not_ redirect loopback addresses.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: ok, agree, and had it like that earlier. but I made the change in response to your previous comment:

> imo this method shouldn't print as it isn't responsible for this action. printing of an action should happen where the action takes place :)

HotwheelsSisyphus:
> Although - I think one thing we could do is make this configurable by the user by making the regex a documented classvar which defaults to the present regex and can be adapted in e.g. a startup file. This would allow users to accept a risk explicitly, but protect all other users.

I think this is the best solution. so `redirectPattern` will become a classvar, and for the boolean I suggest something like `redirectEnabled`?

In that case, I think it's also reasonable to make the default pattern simpler, i.e. just `^http` instead of `^https?:\/\/(?!127.*|localhost)`. Prko can then simply set it to the more complicated pattern in his quark.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: and including `file` is unnecessary - all we want is for that prefix _not_ to match.
the pattern is meant to match http(s) prefixes (minus those of loopback addresses). That is what gets redirected, and nothing else.

HotwheelsSisyphus: (unless there's a good reason to redirect file:: prefixes to the browser).

prko:
>>Although - I think one thing we could do is make this configurable by the user by making the regex a documented classvar which defaults to the present regex and can be adapted in e.g. a startup file. This would allow users to accept a risk explicitly, but protect all other users.
>
>I think this is the best solution. so redirectPattern will become a classvar, and for the boolean I suggest something like redirectEnabled?

I agree — making the redirect pattern configurable is a very good direction, especially for users who are familiar with classvars, startup files, and the `HelpBrowser` class. However, the current behavior still seems to overlook the author’s intention in some cases.

Since introducing a new tag format has already been rejected, I would like to suggest an alternative that stays within the existing `link::...::` syntax.

Some documentation authors may want certain links to open inside the HelpBrowser (for example, pages like *sccode.org* are often more readable there). To support this intention, we could extend the logic in `HelpBrowser.prRedirectToSystemBrowser` so that:

- **if the first character inside a `link::...::` target is a specific prefix (e.g. `i.`)**,  
  then the link is opened in the HelpBrowser instead of the system browser. For example:

To open a web page in the HelpBrowser:
```SCDoc
link::i.https://sccode.org/::
```

To open a web page in the system default browser:
```SCDoc
link::https://sccode.org/::
```

This keeps the current tag format unchanged, respects the author’s intention, and avoids introducing new syntax.  
It also provides a simple and explicit mechanism for authors who want to opt into HelpBrowser rendering.

Does this seem reasonable to you?

HotwheelsSisyphus:
To be honest, this sounds like you're asking for a bespoke solution for something you want to do in your code... I don't know what @capital-G thinks, but I think we should probably not do that, and instead prioritize security here.

Especially since for your quark, you could simply overwrite the *goTo and *prRedirect methods in whichever way you see fit, by placing an ext-HelpBrowser.sc file within a SystemOverwrites folder within the directory of your quark. 

That file would then read
```
+ HelpBrowser {
*goTo { 
    / /whatever you like
}

*prRedirectToSystemBrowser {
    //whatever you like
}
}
```

prko:
> Especially since for your quark, you could simply overwrite the *goTo and *prRedirect methods in whichever way you see fit, by placing an ext-HelpBrowser.sc file within a SystemOverwrites folder within the directory of your quark.

Thank you for the suggestion — technically speaking, placing an ext-HelpBrowser.sc file inside a SystemOverwrites folder would indeed work.

From my experience, though, overriding core classes or methods is something that should be approached with great caution. While it can be useful for experimentation or for very specific workflows, it often leads to unexpected behaviour, especially for curious users who like to explore different parts of the system. In the long run, such overrides can become a source of confusion, wasted time, or even discourage users who are still learning their way around SuperCollider.

For this reason, I feel it is generally better to avoid relying on SystemOverwrites unless absolutely necessary, and to prefer solutions that keep the default class library intact.

HotwheelsSisyphus:
Please name a concrete case of this "confusion and wasted time" resulting from systemoverwrites. Apologies, but I genuinely can't tell if this is a legitimate concern or something your LLM hallucinated.
Generally if you only have five or so quarks/extensions installed, there is unlikely to be much conflict in the overwrites (and who else overwrites the helpbrowser anyway?), and if a beginner installs more than five quarks or extensions, then **that** is probably the source of "confusion and wasted time", not the systemoverwrite. 

Also, you can just set a different regex pattern for the redirectPattern, such that all the links you want to keep inside the Helpbrowser are excluded (i.e., not redirected). You already started doing that above, so I don't see why we need to change the code again. Just have your quark set a different redirectPattern.

prko:
>Please name a concrete case of this "confusion and wasted time" resulting from systemoverwrites. Apologies, but I genuinely can't tell if this is a legitimate concern or something your LLM hallucinated.

To clarify, the concern is genuinely mine — it is not something produced by an LLM. The phrase “overriding core classes” was suggested by the LLM; in my original wording I simply referred to “built‑in classes”.

I can name at least one concrete case, and I could probably find several more by reviewing issues I have filed in the past. However, describing them in detail here might unintentionally single out the authors involved, which I would prefer to avoid.

In general, conflicts arising from installing multiple Quarks are not unusual. The most common problems come from duplicated class definitions, but there have also been cases — admittedly less frequent, yet still real — where overwritten classes caused unexpected behaviour. These situations can be confusing for users who enjoy experimenting with the system, and they often lead to time spent tracing problems back to an override. This difficulty is compounded by the fact that many Quarks are not thoroughly documented, which can make identifying the source of a conflict particularly challenging.

> You already started doing that above, so I don't see why we need to change the code again.

Regarding your comment about not needing to change the code again: my concern is about respecting the author’s intention when a help page is designed to be read and executed directly (for example, on sites such as sccode.org). The current implementation focuses on user customisation, which means there is no way to preserve the author’s intended behaviour in such contexts.

Of course, if this PR is merged, I will follow the new behaviour. Unless someone raises the same concern again in the future, I do not intend to object further. As you said, we should proceed in the way you consider appropriate.

HotwheelsSisyphus:
Thanks for your response. I think now I finally understand what you mean by the "author's intended behavior": What the author of the respective help file wanted when they put the links there... sorry for being so slow to pick up on that.

Anyway, I think in this case it's still better to handle this globally (via the `redirectPattern`) than locally (per individual link, as your suggested solution with the prefix). Otherwise, we build in the kind of security risk that redirecting to system browser is supposed to remove, and in a somewhat unpredictable matter that's hard to manage: We'd have to look at each link individually to tell if it will be redirected or not. Whereas with the `redirectPattern`, there's so to speak one big, central switch which you can flip to get the safe or unsafe behavior. 

Otoh, I'm not that well versed in security issues either, so maybe I'll defer to @capital-G 's judgment in this...

prko:
Before further perspectives are added, I’d like to clarify one practical point from the user side — and here “user” includes both people who read the documentation and those who write or maintain it.

The SC HelpBrowser is not just a documentation viewer; it also provides the ability to evaluate sclang code directly. This is an essential part of how many users explore examples, test ideas, and write or refine documentation.

If web pages containing SuperCollider code examples (such as sccode.org or similar resources) are always redirected to the system browser, users lose the ability to evaluate code snippets inside the integrated environment. This reduces the usefulness of the HelpBrowser as an interactive learning and authoring tool.

Additionally, when reading or sharing SuperCollider code on GitHub, the HelpBrowser is often the most efficient environment. Of course, WebView can also be used, but there is usually no need for that. Users should not have to copy code from a system browser into SuperCollider or VSCode, or perform the reverse process, just to test or verify examples. Keeping such pages inside the HelpBrowser preserves a smoother workflow.

I’m mentioning this because the decision about which URLs should remain inside the HelpBrowser has a significant impact on how people learn, experiment, and contribute documentation within the SC ecosystem.

capital-G:
Maybe as a compromise: Instead of implicitly redirecting to the system browser on external links, we could ask the user if the link should be opened in the help browser via a popup dialog?
This makes the implementation more complex though if we use Qt for that, but we maybe could just use JavaScript for that, which would probably be simpler? 
If a link is clicked, it is checked if it begins with `http` - if so, display a dialog via JS which ask if the external link should be opened in the HelpBrowser or the SystemBrowser. In the latter case, a `os://<url-that-has-been-clicked>` is prepended and sclang can then call the SystemBrowser with the given url.
Regarding how to open a link in a system browser, maybe https://docs.python.org/3/library/webbrowser.html will have some information on the dos and donts.

As mentioned, the HelpBrowser is not secure and intended to browse the internet. I don't want to use this as a reason to remove features, but we should inform the user about this.

Sorry that the PR takes so long @HotwheelsSisyphus ...

HotwheelsSisyphus:
It almost looks like this can be done in sclang, except maybe not because of some async stuff I can't figure out.
```
	*prRedirectToSystemBrowser { |url|
		var return;
		url = url.asString;
		if(redirectEnabled and: { redirectPattern.matchRegexp(url) }) {
			Menu(
				MenuAction("Open in System Browser", { url.openOS; return = true }),
				MenuAction("Open in HelpBrowser (unsafe)", { return = false }),
			).front;
			^return; // I think this gets returned before being assigned a value. what do?
		} {
			^false
		}
	}
```

prko:
Thanks for the investigation and the PR — I really appreciate the effort.

While testing I found that `MenuAction` callbacks are not working on macOS 15.7.4+ (see #7436). I also hit an `ERROR: Non Boolean in test.` at this line:
```supercollider
{ redirected } { ^nil }   // case expects boolean tests
```
in
```supercollider
	*goTo {|url|
		var ideClass = \ScIDE.asClass;
		var redirected = HelpBrowser.prRedirectToSystemBrowser(url);
		case
		{ redirected } { ^nil } // <- this line produces the error.
		{ ideClass.notNil } { ideClass.openHelpUrl(url) }
		{ this.front.goTo(url) };
	}
```

so `prRedirectToSystemBrowser` alone didn’t resolve the issue for me.

To keep the change minimal and cross‑platform, I reverted the other suggested edits and changed only `*goTo` to present a small menu and attach callbacks directly to the `MenuAction`s. This version works for me on Windows 11 (develop build) and macOS 10.14.6 (SC 3.13.0):

```supercollider
	*goTo {|url|
		var ideClass = \ScIDE.asClass;
		var systemDefaultBrowser = MenuAction("Open in system default browser");
		var helpBrowser = MenuAction("Open in sclang HelpBrowser (unsafe)");
		var menu = Menu(systemDefaultBrowser, helpBrowser);
		systemDefaultBrowser.addDependant { |action, what| if (what == \triggered) { url.openOS; } };
		helpBrowser.addDependant { |action, what| if (what == \triggered) { ideClass.openHelpUrl(url); } };
		if (ideClass.notNil) {
			if(url.beginsWith("http")) {
				menu.front;
			} {
				ideClass.openHelpUrl(url);
			}
		}{
			this.front.goTo(url);
		}
	}
```
If that approach sounds acceptable, could @HotwheelsSisyphus reflect this change in your PR?
Thanks again — happy to iterate on this.

(The only problem is on macOS 15.7.4+ as described above...)

prko:
Occasionally, the menu fails to close. The following implementation also resolves the issue:
```
	*goTo {|url|
		var ideClass = \ScIDE.asClass;
		var systemDefaultBrowser = MenuAction("Open in system default browser");
		var helpBrowser = MenuAction("Open in sclang HelpBrowser (unsafe)");
		var closeMenu = MenuAction("Do not open");
		var menu = Menu(systemDefaultBrowser, helpBrowser, closeMenu);
		var removeDependants = { [systemDefaultBrowser, helpBrowser, closeMenu].do(_.removeDependant)  };
		systemDefaultBrowser.addDependant { |action, what| if (what == \triggered) { url.openOS; removeDependants.() } };
		helpBrowser.addDependant { |action, what| if (what == \triggered) { ideClass.openHelpUrl(url); removeDependants.() } };
		closeMenu.addDependant { |action, what| if (what == \triggered) { removeDependants.() } };
		if (ideClass.notNil) {
			if(url.beginsWith("http")) {
			menu.front;
			} {
				ideClass.openHelpUrl(url);
			}
		}{
			this.front.goTo(url);
		}
	}
```

HotwheelsSisyphus:
@prko, Thanks for the testing! 
I basically took over your code, just rearranged it a bit so it can also be used from a HelpBrowser instance (not just the IDE).

prko:
Thank you!

I think it would be better if the menu items — especially the third one — were more informative.  
For example, instead of simply “Do not open”, something like:

**“Do not open this external link”**

might make the intention clearer.  
As it is now, users may wonder why this option appears at all.

I’m also getting the following error when selecting **“Open in sclang HelpBrowser (unsafe)”**:

```
ERROR: Non Boolean in test.
RECEIVER:
   nil
CALL STACK:
	MethodError:reportError
		arg this = <instance of MustBeBooleanError>
	Nil:handleError
		arg this = nil
		arg error = <instance of MustBeBooleanError>
	Thread:handleError
		arg error = <instance of MustBeBooleanError>
	Object:throw
		arg this = <instance of MustBeBooleanError>
	Object:mustBeBoolean
		arg this = nil
	Meta_HelpBrowser:prIDEorHelpBrowser
		arg this = <instance of Meta_HelpBrowser>
		arg url = "https://supercollider.github..."
		arg openInSCIDE = nil
	< FunctionDef in Method Meta_HelpBrowser:prRedirectMenu >
		arg action = <instance of MenuAction>
		arg what = 'triggered'
	Function:update
		arg obj = <instance of MenuAction>
		arg what = 'triggered'
		arg args = [*1]
	< FunctionDef in Method Set:do >
		arg item = <instance of Function>
	ArrayedCollection:do
		arg this = [*4]
		var i = 1
	Set:do
		arg this = <instance of IdentitySet>
		var i = 0
	Object:changed
		arg this = <instance of MenuAction>
		arg what = 'triggered'
		arg moreArgs = [*1]
	AbstractMenuAction:onTriggered
		arg this = <instance of MenuAction>
		arg b = false
^^ The preceding error dump is for ERROR: Non Boolean in test.
RECEIVER: nil
```

To resolve this, I integrated `prRedirectToSystemBrowser` inside `goTo`.  
If this error is resolved and `MenuAction` works again on macOS, we can move forward from here!

HotwheelsSisyphus:
Did you clone the commit I pushed today? Because I renamed the method, there is no longer a `prRedirectToSystemBrowser`. 

As I said, I did rearrange things a bit, and we won't get this error any more because the conditional now works differently.

HotwheelsSisyphus:
Fwiw, the added methods:
prShouldRedirect (a test whether to redirect at all) 
prRedirectMenu (code for the actual menu)
prIDEorHelpBrowser (we need slightly different behavior depending on whether you're working with the IDE help browser or an instance of HelpBrowser).

Maybe we can just replace the third button text with "cancel", which I think is pretty universally intelligble.
And I think there could be another (non-clickable) line in the menu that shows just the domain of the link). Seems like "separator" is the only way to do this...

Rough sketch (won't align properly here, sry):
```
       var domain = (url.replaceRegexp("https?:\/\/(www.)?|(/.*)", "");         
      var info = MenuAction(domain).separator_(true);
       var systemDefaultBrowser = MenuAction("Open in system default browser");
		var helpBrowser = MenuAction("Open in sclang HelpBrowser (unsafe)");
		var closeMenu = MenuAction("Cancel");
		var menu = Menu(info, systemDefaultBrowser, helpBrowser, closeMenu);

```

prko:
> Did you clone the commit I pushed today? Because I renamed the method, there is no longer a prRedirectToSystemBrowser.

By `prRedirectToSystemBrowser`, I was referring to the old implementation.  
Sorry for the confusing wording.

> Rough sketch (won't align properly here, sry):

I also tested your sketch, but the error still occurs on my end:

> ERROR: Non Boolean in test.

I’m attaching a video showing exactly what I did (you can see the error in the recording):
https://www.dropbox.com/scl/fi/z002gf397gi32uha8vgxr/HelpBrowser.mov?rlkey=qbpblp8kwna0pa1vfx4452m3u&st=yh2xhq2k&dl=0

prko:
Oh, I found the reason for the error:

```
	*goTo {|url|
		if(this.prShouldRedirect(url)) {
			this.prRedirectMenu(url)
		} {
			this.prIDEorHelpBrowser(url)
		}
	}
```
It should be:
```
	*goTo {|url|
		if(this.prShouldRedirect(url)) {
			this.prRedirectMenu(url, true)
		} {
			this.prIDEorHelpBrowser(url)
		}
	}
```

---

Regarding the menu text, how about the following?
```
	*prRedirectMenu { |url, openInSCIDE|
		// var domain = (url.replaceRegexp("https?:\/\/(www.)?|(/.*)", ""));
		var info = MenuAction("Choose how to open this external link:" + url).enabled_(false);
		// var info = MenuAction("External link detected:" + url).enabled_(false);
		var separator = MenuAction.separator;
		var systemDefaultBrowser = MenuAction("Open in system default browser");
		var helpBrowser = MenuAction("Open in sclang HelpBrowser (unsafe)");
		var cancelMenu = MenuAction("Cancel");
		var menu = Menu(info, separator, systemDefaultBrowser, helpBrowser, cancelMenu);
		var removeDependants = {
			[systemDefaultBrowser, helpBrowser, cancelMenu].do(_.removeDependant)
		};
		systemDefaultBrowser.addDependant { |action, what|
			if (what == \triggered) {
				url.openOS;
				removeDependants.() }
		};
		helpBrowser.addDependant { |action, what|
			if (what == \triggered) {
				this.prIDEorHelpBrowser(url, openInSCIDE);
				removeDependants.() }
		};
		^menu.front;
	}
```

I think exposing `http://`, `https://`, and `www.` is useful, but the full URL might be too long.  I’ll follow your decision on this.

Anyway, it seems to be working now.

--- 

Added:

One more design suggestion:
For the menu text, how about the following?

```
	*prRedirectMenu { |url, openInSCIDE|
		// var domain = (url.replaceRegexp("https?:\/\/(www.)?|(/.*)", ""));
		var info = MenuAction("Choose how to open this external link:\n").enabled_(false);
		// var info = MenuAction("External link detected:").enabled_(false);
		var separator = MenuAction.separator;
		var systemDefaultBrowser = MenuAction("Open in system default browser");
		var helpBrowser = MenuAction("Open in sclang HelpBrowser (unsafe)");
		var cancelMenu = MenuAction("Cancel");
		var menu = Menu(
			info,
			MenuAction(url).enabled_(false),
			separator,
			systemDefaultBrowser,
			helpBrowser,
			cancelMenu
		);
		var removeDependants = {
			[systemDefaultBrowser, helpBrowser, cancelMenu].do(_.removeDependant)
		};
		systemDefaultBrowser.addDependant { |action, what|
			if (what == \triggered) {
				url.openOS;
				removeDependants.() }
		};
		helpBrowser.addDependant { |action, what|
			if (what == \triggered) {
				this.prIDEorHelpBrowser(url, openInSCIDE);
				removeDependants.() }
		};
		^menu.front;
	}
```
Screenshot 2026-03-28 at 15 50 46

HotwheelsSisyphus:
Thanks!

Ah, I'd overlooked that non boolean... I guess default args don't work the way I thought they would. Oh well.  Fixed that now. Could you test the new commit and see if you're ok with it? I hope I haven't overlooked anything this time.

JordanHendersonMusic:
@HotwheelsSisyphus re the default args, we recently had a huge discussion about this @
#7413! 

Tldr: today, you should do `someArg = (true)` right now - it's very confusing. Only use `someArg = true` when you actually want the user to be able to pass `nil` sometimes (which you almost never do).

prko:
I tested it on Windows 11, and it now works!
Thank you!!!

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,12 +40,40 @@ HelpBrowser {
 		^super.new.init( aHomeUrl ? defaultHomeUrl, newWin ?? { openNewWindows } );
 	}
 
-	*goTo {|url|
+	*prShouldRedirect {|url|
+		^(redirectEnabled and: { redirectPattern.matchRegexp(url) })
+	}
+
+	*prRedirectMenu { |url, openInSCIDE|
+		var info = MenuAction("Choose how to open external link:\n").enabled_(false);
```
prko:
```suggestion
		var info = MenuAction("Choose how to open external link:").enabled_(false);
```

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,12 +40,40 @@ HelpBrowser {
...
+		var domain = MenuAction(if(url.size < 75) {url} {url.replaceRegexp("https?:\/\/|(/.*)", "")} ).enabled_(false);
```
prko, line 49: Wow, this is truly impressive!

prko:
I’m really satisfied with the latest commit!
Previously, if you needed to download a file while viewing SC‑related webpages in the HelpBrowser, you had to copy the URL, open it in an external browser, and download it manually.
Now, you can simply choose ‘Open in system default browser’ from the menu, and the file downloads automatically while the browser pops to the front. It makes the whole process smooth and seamless.

Screenshot 2026-03-29 at 9 00 53 PM

HelpSource/Classes/HelpBrowser.schelp
```diff
@@ -47,6 +47,16 @@ Get or set the default home URL.
 
 method:: openNewWindows
 Get or set the default for "open in new windows" toggle.
+When set to teletype::false:: (default), this class behaves like a singleton class.
+
+method:: redirectEnabled
+When set to true (default), redirects links that match code::redirectPattern:: to the system's default Web browser.
```
HotwheelsSisyphus:
```suggestion
When set to true (default), clicking on links whose target matches code::redirectPattern:: will open a menu giving the user the choice to open the link in the system's default Web browser (recommended), 
or in SuperCollider's own help browser (strong::not:: recommended for remote links, because not safe for web browsing).
```

HelpSource/Classes/HelpBrowser.schelp
```diff
@@ -47,6 +47,16 @@ Get or set the default home URL.
...
+Disable at your own risk: SC's help browser is not web safe.
+
+method:: redirectPattern
+Get or set the regex pattern (a link::Classes/String#Regular expressions#String::) that determines which links are forwarded to the system's default browser.
```
HotwheelsSisyphus:
```suggestion
Get or set the regex pattern (a link::Classes/String#Regular expressions#String::) that determines which links will open the menu described above (link::#*redirectEnabled::). 
```

HelpSource/Classes/HelpBrowser.schelp
```diff
@@ -47,6 +47,16 @@ Get or set the default home URL.
...
+When the link matches this pattern, it will be opened in the system's default browser (provided code::redirectEnabled:: is teletype::true::.)
```
HotwheelsSisyphus:
```suggestion
```

HotwheelsSisyphus:
Thanks! Alright, now basically just waiting for another reviewer. Btw I added @prko as a co-author in the latest commit, just so that doesn't get forgotten when we squash and merge.

prko:
> Btw I added @prko as a co-author in the latest commit, just so that doesn't get forgotten when we squash and merge.

Much appreciated!

I tested the latest commit with SuperCollider 3.13.0 on macOS, using a local modification that replaces `replaceRegexp` with `replace` (since `replaceRegexp` is not available in 3.13.0) in order to check the menu functionality. Everything works perfectly.
If `Menu` and `MenuAction` also work on macOS with SC 3.14.1 and SC 3.15.0‑dev, that would be ideal.

prko:
I tested it on Ubuntu, and there is an issue with the menu position.
The menu does not appear at the mouse-click position. I think this could be related to #6736.
Screenshot 2026-03-31 at 10 44 57

HotwheelsSisyphus:
Huh, this is interesting. For me (ubuntu studio, kde plasma, wayland) the menu reliably appears in the middle of the screen (not next to the link), which I think is acceptable. Are you on X11 or wayland? Kde? Gnome? (etc.) These things seems to matter in linuxland...

At any rate, I don't think this is within the purview of the present PR... can't fix everything.

prko:
I think my Ubuntu system does not have KDE Plasma:
```bash
echo $XDG_SESSION_TYPE
wayland
```
```bash
parallels@ubuntu-gnu-linux-24-04-3:~/Desktop$ echo $XDG_CURRENT_DESKTOP
ubuntu:GNOME
```
- **Firmware Version:**                            26.3.0 (57392)
- **OS Name:**                                     Ubuntu 24.04.4 LTS
- **OS Build:**                                    (null)
- **OS Type:**                                     64-bit
- **GNOME Version:**                               46
- **Windowing System:**                            Wayland
- **Kernel Version:**                              Linux 6.17.0-19-generic

Yes, fixing this is outside of this PR.

dyfer:
@HotwheelsSisyphus @prko @capital-G what else do we need in this PR?

prko:
This appears ready for merging. The behaviour on Linux when using Wayland will need to be handled additionally, preferably in a separate PR.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -4,6 +4,10 @@ HelpBrowser {
 	classvar <>openNewWindows = false;
 	classvar <>scrollStep = 40;
 	classvar <>scrollPageStep = 350;
+	classvar <>redirectEnabled = true;
+	// redirect link addresses that match redirectPattern to default system browser
+	classvar <>redirectPattern = "^http";
+	// all http gets redirected
```
capital-G: Why is some comment above and some below - IMO the below one is a bit redundant

We could also think about naming this `redirectRegex`, but i am fine either way.


```suggestion
	// redirect link addresses that match redirectPattern to default system browser
	classvar <>redirectPattern = "^http";
```

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,12 +40,40 @@ HelpBrowser {
...
+		var helpBrowser = MenuAction("Open in SuperCollider's help browser (unsafe)");
+		var closeMenu = MenuAction("Cancel");
+		var menu = Menu(info, domain, separator, systemDefaultBrowser, helpBrowser, closeMenu);
+		var removeDependants = { [systemDefaultBrowser, helpBrowser, closeMenu].do(_.removeDependant)  };
+		systemDefaultBrowser.addDependant { |action, what| if (what == \triggered) { url.openOS; removeDependants.() } };
+		helpBrowser.addDependant { |action, what| if (what == \triggered) { this.prIDEorHelpBrowser(url, openInSCIDE); removeDependants.() } };
+		^menu.front;
+	}
+
+	*prIDEorHelpBrowser { | url, openInSCIDE |
```
capital-G: `openInSCIDE` sounds like a boolean flag, which later defaults to `true` and is not always specified. so maybe just declare its default implicit value in its signature?
Maybe also name it `useIDE`, b/c `openIn` is a bit uncommon as variable name setup?

Also the name could be a bit better, e.g. `prOpen` or something w/ goto or navigate?

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -239,18 +272,21 @@ HelpBrowser {
...
-				if(newPath != oldPath && openNewWin) {
-						HelpBrowser.new(newWin:openNewWin).goTo(url);
+			var newPath, oldPath;
+			if( not(this.redirectTextFile(url)) ) {
+				if( HelpBrowser.prShouldRedirect(url)) {
+					HelpBrowser.prRedirectMenu(url, openInSCIDE: false)
 				} {
-					this.goTo(url);
-				};
+					#newPath, oldPath = [url,webView.url].collect {|x|
```
capital-G, line 279: why don't we declare `newPath` and `oldPath` only in this scope where it is needed?
Makes reading a bit clearer to me b/c otherwise I have to look a head into the `not` branch to figure out for what these variables are actually used.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,12 +40,40 @@ HelpBrowser {
 		^super.new.init( aHomeUrl ? defaultHomeUrl, newWin ?? { openNewWindows } );
 	}
 
-	*goTo {|url|
+	*prShouldRedirect {|url|
+		^(redirectEnabled and: { redirectPattern.matchRegexp(url) })
+	}
+
+	*prRedirectMenu { |url, openInSCIDE|
```
capital-G: see `prIDEorHelpBrowser`.
Also this could be named `prShowRedirectMenu` b/c this is not there to get/construct the redirect menu, but performs the `.front` operation on it.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -137,6 +169,7 @@ HelpBrowser {
 		var x, y, w, h;
 		var str;
 
+		if (not(openNewWindows)) { singleton = this };
```
capital-G, line 171: Was the singleton implementation broken or why was this modified?

capital-G (commented):
Only found some minor things for some optimizations, but this could already go in as-is.

HotwheelsSisyphus:
Hi, sorry for the long absence (moved back to Vienna to be unemployed here instead) - from my end I had thought of it as mostly finished but I'll go over it again with Dennis's comments...

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: yes (iirc, only for the helpbrowser as Qt object, not when used in the IDE).

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: sure  will do. I think this is because I had originally tried to keep the changes more minimal

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: yes, I think there was some mixup because I was using the old style of default (openInSCIDE=true) in the arg declaration, and that hadn't worked. I'll use the new style which presumably will

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: ah yes, in my mind each comment was assoicated to previous line; fine to remove

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: wait, actually it might make sense to keep the variable declaration out of the if-statement because then it will inline.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: the method _does_ "get/construct" the menu, though. 
I could change this slightly if you think it's more reasonable, and put the .front method in the call site, i.e. we return `^menu` here and below we call `this.prRedirectMenu(args).front` if you think that this is clearer.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
HotwheelsSisyphus: but now I'm not sure if the change to OnLinkActivated is even relevant... that method isn't called anymore when I click links in a HelpBrowser instance... (instead it simply seems to be .goTo) ... but I swear that a few months back, it was (and worked!).  I'm testing on commit 21347a84c on Kubuntu. I will want to figure this out before this gets merged...

Maybe something has changed in qt or js on the system that affects SC behavior? ...

HotwheelsSisyphus:
@capital-G: I'm a bit stumped by the `webview` issue (only relevant if one isn't using the SCIDE help browser: It appears that `.onLinkActivated` just never registers, even though `.overrideNavigation` returns true. 
Can you see if you can reproduce this or if it's just something on my end that has broken the Qt? 

```
// using this branch;
h = HelpBrowser();
// now click on some link - expected behavior is that the menu should open;
// actual behavior is that the HelpBrowser instance simply navigates to the link.
//still it seems that the webview was initialized correctly:
v = h.window.view.children[7] // a WebView
v.onLinkActivated // will return a function, even though I can't get anything out of .def.sourceCode.postcs;
v.overrideNavigation // will return true.
// so somehow the onLinkActivated never gets triggered. (verifiable by putting a postln in the function inside the init method)
// however setting the function afte object initialization seems to work:
v.onLinkActivated = { |wv, func| "navigation is being overriden".postln };
// now attempt some navigation: should post some things but not navigate.
```

If I can't fix this part, then I'll just revert the changes that concern the non-SCIDE Helpbrowser and leave them for another day... a pity though. I am starting to doubt myself but I was quite sure that this had worked back in March; now it doesn't, even though I haven't touched my SC installation since then and only did system updates.


<details><summary>but it works fine in this example adapted from WebView.schelp</summary>
<p>

```
// A simple web browser
(
var browser, webview, reloadStop, prev, next, urlBox, go;

webview = WebView()
    .minSize_(300@200);
reloadStop = Button()
    .states_([["※"], ["◙"]])
    .fixedSize_(36@28);
prev = Button()
    .states_([["⇦"]])
    .fixedSize_(36@28);
next = Button()
    .states_([["⇨"]])
    .fixedSize_(36@28);
urlBox = TextField()
    .minWidth_(100);
go = Button()
    .states_([["⌘"]])
    .fixedSize_(36@28);

reloadStop.action = {
    |v|
    if (v.value == 1) {
        webview.reload(true);
    } {
        webview.stop;
    }
};

prev.action = { webview.back };
next.action = { webview.forward };
urlBox.action = { webview.url = urlBox.string };
go.action = { webview.url = urlBox.string };

webview.onUrlChanged = {
    |view, url|
    urlBox.string = url;
};

webview.onLoadStarted = {
    reloadStop.value = 1;
    urlBox.background = Color.grey(0.4);
};

webview.onLoadFinished = {
    reloadStop.value = 0;
    urlBox.background = Color.grey(0.2);
};

webview.onLoadFailed = {
    reloadStop.value = 0;
    urlBox.background = Color.red(1, 0.2);
};

/////////////  my addition
webview.onLinkActivated = { |wv, url| "navigation is being overridden".postln };
webview.overrideNavigation = true;
////////////
browser = View(bounds: 900@700).layout_(VLayout(
    HLayout(
        prev, reloadStop, next,
        urlBox,
        go
    ),
    webview
));

browser.front;

urlBox.valueAction = "http://supercollider.github.io/";
)
```
</p>
</details>

HotwheelsSisyphus:
nevermind, turns out I am extremely stupid: I had a systemoverride for the .init method (to add a keyDownAction), so I wasn't actually running the right code...

HotwheelsSisyphus:
Should be good to merge now! Please ignore my previous comments about `onLinkActivated`, I was a bit out of it and had forgotten about my overrides.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
```diff
@@ -36,12 +40,39 @@ HelpBrowser {
...
+		var helpBrowser = MenuAction("Open in SuperCollider's help browser (unsafe)");
+		var closeMenu = MenuAction("Cancel");
+		var menu = Menu(info, domain, separator, systemDefaultBrowser, helpBrowser, closeMenu);
+		var removeDependants = { [systemDefaultBrowser, helpBrowser, closeMenu].do(_.removeDependant)  };
+		systemDefaultBrowser.addDependant { |action, what| if (what == \triggered) { url.openOS; removeDependants.() } };
+		helpBrowser.addDependant { |action, what| if (what == \triggered) { this.prOpenInIDEorHelpBrowser(url, useSCIDE); removeDependants.() } };
+		^menu.front;
+	}
+
+	*prOpenInIDEorHelpBrowser { | url, useSCIDE(true) |
```
capital-G, line 61: why do you use the init constructor(?) writing instead of the more common default constructor, i.e.


```suggestion
	*prOpenInIDEorHelpBrowser { | url, useSCIDE=true |
```

capital-G (approved):
Nice! Lgtm, feel free to merge! :)

I have one question/remark regarding the default param - feel free to choose either way, but I'd be interested in the reasoning behind the current path.

SCClassLibrary/Common/GUI/tools/HelpBrowser.sc
JordanHendersonMusic: No this is the right way, with = it can be nil.

capital-G: Ah thx for the info! Can you share some link/docs where this gets explained? I think writing `useSCIDE=true` is rather canonical, I have not come across the `(true)` writing in any code yet. I thought the latter was more relevant to mutable variables, like the same problem as in python.

Should we adapt the class library in that case? B/c there `=` is used everywhere.

dyfer: Wasn't this heavily discussed in https://github.com/supercollider/supercollider/pull/7413?
I agree that the `arg(val)` notation is un(der?)documented and I personally never used it, but of course it's fine if it's intentional in this case.

HotwheelsSisyphus: it's intentional indeed. 
```
a = { | x(1) | x };
b = { | y | a.(y) };
b.() // should return 1; whereas if 
a = { |x=1| x };
b.() // will return nil - we don't want that
```
So the point is to resolve to default only at the last moment, but with the arg=val syntax the unassigned args from calling functions will be interpreted as "deliberate" nils and cause trouble in the test.

capital-G: > it's intentional indeed.

my 2ct: IMO this is not good design as the default value isn't exposed/propagated to the user interface in that case.
The `()` arguments are nice when wrapping functions, but I found it a bit excessive to be used here.

But as mentioned, if you think it is fine as-is, feel free to merge :)

prko: https://dev.docs.supercollider.online/Reference/Functions.html#[2]%20Expression%20argument%20defaults

I think the standard notation is `*prOpenInIDEorHelpBrowser { | url, useSCIDE=(true) |`.

JordanHendersonMusic: > The () arguments are nice when wrapping function

So we have coloured functions? Ones that can be wrapped and ones that can't. No thank you!

> I think the standard notation is *prOpenInIDEorHelpBrowser { | url, useSCIDE=(true) |.

This is two types of initialisers, I think this is the most confusing.

> I have not come across the (true) writing in any code yet

Literally just fixed a bug that brought down the ci because of this... Well it was `("")`. If you see `arg x=true;.... if(x)...` that will throw if wrapped, it should *never* be = because you can only have bools in if statements.

JordanHendersonMusic: > my 2ct: IMO this is not good design as the default value isn't exposed/propagated to the user interface in that case.

This is a failing of the tooling, not the language. We shouldn't we writing code to make the tools do what we want, we should fix the tools.

prko: >>I think the standard notation is *prOpenInIDEorHelpBrowser { | url, useSCIDE=(true) |.

>This is two types of initialisers, I think this is the most confusing.

What I meant was this:

Between `{ | x = (1) | x }` and `{ | x(1) | x }`, the form `{ | x = (1) | x }` seems to be the standard one, while `{ | x(1) | x }` feels more like a shorthand.

I did mean that both `{ | x = (1) | x }` and `{ | x = 1 | x }` are standard.

Sorry for the confusing wording.

capital-G: > > The () arguments are nice when wrapping function
> 
> So we have coloured functions? Ones that can be wrapped and ones that can't. No thank you!

That is indeed a good argument, though I still don't like that you can not pass `nil` explicitly to these functions, so it is also not the allround solution to fix this issue in general.
So it seems as a rule of thumb: If the argument is valid to be nil, use `=`, and otherwise use `()`?

Will lead to some inconsistencies with the existing code base, but IMO that can be improved over time and something we should adapt to.

Though I share the view of prko here, we haven't documented that `|foo(true)|` is valid syntax, [the docs](https://docs.supercollider.online/Reference/Functions.html#Arguments) only provide `foo=(true)`, and this communicates much clearer that it is a (wrapped) default assignment.
Or we add to the docs that `foo(true)` is also valid. I don't have any preference here, but we should adapt a consistent style within the class library.

https://github.com/supercollider/supercollider/wiki/Style-Guidelines%3A-SuperCollider#parameter-lists suggests `foo = (true)`.

> This is a failing of the tooling, not the language. We shouldn't we writing code to make the tools do what we want, we should fix the tools.

I don't understand how this is related to a tooling issue: In Python it is really common to propagate/repeat kwargs - it is somehow annoying, but simply provides a clearer dev experience IMO b/c the interface tells already a lot about the implementation, no need to bring up the docs or the source code or trace how the default argument gets resolved eventually.

JordanHendersonMusic: > seems to be the standard one, while { | x(1) | x } feels more like a shorthand.

It's not, it's in the grammar as the correct way to do this. If anything, `x=(...)` is the odd one as it behaves exactly as `x(...)`. You use `x=(...)` when you have a non-literal value, which does the same behaviour of `x(...)`... So the equals sign is unnecessary.

JordanHendersonMusic: > That is indeed a good argument, though I still don't like that you can not pass nil explicitly to these functions, so it is also not the allround solution to fix this issue in general.
So it seems as a rule of thumb: If the argument is valid to be nil, use =, and otherwise use ()?

Yes! See here...

https://scsynth.org/t/the-difference-between-defaults-arguments-with-equals-sign-and-brackets-arg-a-1-vs-arg-a-1/13261

> I don't understand how this is related to a tooling issue

The language wasn't designed with tooling in mind. This is a shame. But trying to adapt the language to the tooling isnt the right approach. Instead I propose something like..
```
\\#doc:
\\#argument: a ....
foo { |a, b|
   ...
}
```
(Just random syntax) But we can borrow from any of the other in text documentation approaches that exist.

prko: >It's not, it's in the grammar as the correct way to do this. If anything, x=(...) is the odd one as it behaves exactly as x(...). You use x=(...) when you have a non-literal value, which does the same behaviour of x(...)... So the equals sign is unnecessary.

@JordanHendersonMusic 
Thanks! It would be nice if the Functions reference included at least one sentence explaining this.

I was confused because of the following line in the documentation:
https://github.com/supercollider/supercollider/blob/06e0206d912326196dd1af78a6fcc4411032f573/HelpSource/Reference/Functions.schelp#L202

This comment gave me the impression that using `=` is the standard form, so I didn’t realise that `x(...)` is actually the canonical grammar and `x=(...)` is just an alternative.

Sorry for the confusing wording.

HotwheelsSisyphus: oh wow, didn't expect that to be so controversial...
Anyway, it seems there are at least two threads of discussion here:

1 what do in this case? A maybe uncontroversial way to settle this maybe would be to remove the default value, rewrite the test and rename the arg such that it tests whether ` useSCIDE.isNil`. Whether this is better design idk; also note that these are all private functions that the user isn't actually meant to interface with.
2 general issue to settle: what default arg style in class library?
3. design issue, independent of sclang - how to deal with propagation of defaults. Personally I just don't know what the recommended practice is here and am happy to take some advice.

I don't want to cut discussion short on 2 and 3, but maybe 2 is better discussed in a separate issue?

capital-G: 1) I think your current path is the correct one, sorry for creating so much noise.
2) Indeed, maybe take this over to https://github.com/supercollider/supercollider/issues/7385 (though we could wait to settle for `foo=(true)` or `foo(true)` via the style-guidelines to have more consistency, see https://github.com/supercollider/supercollider/pull/7421#discussion_r3535273975 ?)
3) Lets also discuss this in https://github.com/supercollider/supercollider/issues/7385 

We don't have to be so rigid within private implementations, so feel free to merge as is - we can later adapt changes if necessary.

JordanHendersonMusic: > so feel free to merge as is

I was just waiting to do that, I didn't want to cut off the conversation ;)

PR #7438, opened 2026-03-30
polishing code in QT Collidr folder for UI Maintainance

tiagomoraismorgado:
This commit does basically aim at maitaining the overall folder for QtCollider. I tried to do it in a decent way. The commit may not be 100% clean and 100% legit. i did my best, now feel free to accept it. if there are any issues, that´s fine as well

### DIFF
--- a/QtCollider/primitives/prim_QObject.cpp
+++ b/QtCollider/primitives/prim_QObject.cpp
@@ -349,7 +349,7 @@ QC_LANG_PRIMITIVE(QObject_SetEventHandler, 4, PyrSlot* r, PyrSlot* a, VMGlobals*
         return errWrongType;
     int eventType = QtCollider::get(a + 0);
     PyrSymbol* method = 0;
-    slotSymbolVal(a + 1, &method);
+    if (slotSymbolVal(a + 1, &method))
     Synchronicity sync = IsTrue(a + 2) ? Synchronous : Asynchronous;
     bool enabled = IsTrue(a + 3);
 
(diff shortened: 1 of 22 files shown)


QtCollider/primitives/prim_QObject.cpp
```diff
@@ -349,8 +349,7 @@ QC_LANG_PRIMITIVE(QObject_SetEventHandler, 4, PyrSlot* r, PyrSlot* a, VMGlobals*
         return errWrongType;
     int eventType = QtCollider::get(a + 0);
     PyrSymbol* method = 0;
-    if (slotSymbolVal(a + 1, &method))
```
JordanHendersonMusic, line 352: All these are no discard as they can fail, you've just made a bunch of warnings. Why change them?

QtCollider/primitives/prim_QObject.cpp
```diff
@@ -350,7 +350,6 @@ QC_LANG_PRIMITIVE(QObject_SetEventHandler, 4, PyrSlot* r, PyrSlot* a, VMGlobals*
     int eventType = QtCollider::get(a + 0);
     PyrSymbol* method = 0;
     if (slotSymbolVal(a + 1, &method))
-        assert(false);
```
JordanHendersonMusic, line 353: This isn't right either, now the line after this doesn't run if `slotSymbolVal` succeeds. There's also the rest of these changes like this that are wrong.

Are you new to c++? Did you use ai to help with this pr?

PR #7441, opened 2026-04-05
Group.queryTree: implement optional arg, updated example in Group.schelp

HotwheelsSisyphus:
Group.query tree does not take arguments since 2007, ~~so the current method description and the example are outdated~~ but now it does. This PR
* adds the option to pass an arg
* ~~removes the example~~ simplifies the example
* _sligthly_ updates the method description
*  also formatting changes to method desc of dumpTree

Types of changes:
- Documentation
- class library

I think this still qualifies as a minor change, even though it's more than a quick fix... I'll leave the PR up until next weekend,  ~~then I'll merge it myself unless someone disagrees or just merges it themselves~~ nah I forgot I still need an approving review before I can merge.

Thanks!

### DIFF
--- a/HelpSource/Classes/Group.schelp
+++ b/HelpSource/Classes/Group.schelp
@@ -147,7 +147,7 @@ method:: queryTree
 Near-duplicate of code::.dumpTree.:: Queries the server for a message describing this group's node subtree.
 Sends a server message of the type teletype::/g_queryTree::; the server's reply is automatically parsed in the post window.
 When you use the code::.queryTree:: method, you cannot customize this parsing; this is only possible when sending the server command directly.
-See teletype::/g_queryTree:: in the link::Reference/Server-Command-Reference#g_queryTree#Server Command Reference::.
+See teletype::/g_queryTree:: in the link::Reference/Server-Command-Reference#/g_queryTree#Server Command Reference::.
 
 
 Examples::
--- a/HelpSource/Classes/Group.schelp
+++ b/HelpSource/Classes/Group.schelp
@@ -144,11 +144,13 @@ If code::postControls:: is true, then the current Control (arg) values for any s
 The default is false.
 
 method:: queryTree
-Near-duplicate of code::.dumpTree.:: Queries the server for a message describing this group's node subtree.
-Sends a server message of the type teletype::/g_queryTree::; if no argument is provided, the server's reply is automatically parsed in the post window.
-Alternatively, a link::Classes/Function:: can be provided as an teletype::action:: argument, for instance: 
-code:: 
-Group().queryTree( { |msg| msg.postln });
+Queries the server for a message describing this group's node subtree.
+Similar to to code::.dumpTree::, but with the difference that code::.queryTree:: permits
+custom handling of the server's reply message when an teletype::action:: argument is provided. 
+When no argument is provided to code::.queryTree::, the output resembles that of code::.dumpTree::.
+code::
+Group().queryTree; // default reply handler.
+Group().queryTree( { |msg| msg.postln }); // custom (but very boring) reply handler.
 ::
 See teletype::/g_queryTree:: in the link::Reference/Server-Command-Reference#/g_queryTree#Server Command Reference::.
 
--- a/HelpSource/Classes/Group.schelp
+++ b/HelpSource/Classes/Group.schelp
@@ -139,10 +139,15 @@ method:: deepFree, deepFreeMsg
 Free all Synths in the group, and all Synths in any enclosed groups, but do not free this group or any of its enclosed groups.
 
 method:: dumpTree
-Post a representation of this group's node subtree. (Sends a message of the type g_dumpTree.) If code::postControls:: is true, then the current Control (arg) values for any synths contained in this group will be posted as well. The default is false.
+Post a representation of this group's node subtree. (Sends a message of the type teletype::g_dumpTree::.)
+If code::postControls:: is true, then the current Control (arg) values for any synths contained in this group will be posted as well.
+The default is false.
 
 method:: queryTree
-Queries the server for a message describing this group's node subtree. (Sends a message of the type g_queryTree.) This reply is passed to the action function as an argument. See g_queryTree in link::Reference/Server-Command-Reference:: or the example below for information on how the reply is structured.
+Near-duplicate of code::.dumpTree.:: Queries the server for a message describing this group's node subtree.
+Sends a server message of the type teletype::/g_queryTree::; the server's reply is automatically parsed in the post window.
+When you use the code::.queryTree:: method, you cannot customize this parsing; this is only possible when sending the server command directly.
+See teletype::/g_queryTree:: in the link::Reference/Server-Command-Reference#g_queryTree#Server Command Reference::.
 
 
 Examples::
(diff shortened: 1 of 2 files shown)


muellmusik:
Thanks for picking this up. Vague memories of this in the distant past...

I think a nicer change would be to restore the option to provide your own parser function. I think that was the original intent, but in any case it would be a useful interface. 

Maybe in this case it's the implementation that's got it wrong and the documentation that's correct! :-)

HotwheelsSisyphus:
Hm, that's doable too... But I figured if the user can write their own parser function, they can simply  `s.sendMsg("/g_queryTree", etc.)` and set up their own OSCFunc?

HelpSource/Classes/Group.schelp
```diff
@@ -139,10 +139,15 @@ method:: deepFree, deepFreeMsg
...
+Post a representation of this group's node subtree. (Sends a message of the type teletype::g_dumpTree::.)
+If code::postControls:: is true, then the current Control (arg) values for any synths contained in this group will be posted as well.
+The default is false.
 
 method:: queryTree
-Queries the server for a message describing this group's node subtree. (Sends a message of the type g_queryTree.) This reply is passed to the action function as an argument. See g_queryTree in link::Reference/Server-Command-Reference:: or the example below for information on how the reply is structured.
+Near-duplicate of code::.dumpTree.:: Queries the server for a message describing this group's node subtree.
+Sends a server message of the type teletype::/g_queryTree::; the server's reply is automatically parsed in the post window.
+When you use the code::.queryTree:: method, you cannot customize this parsing; this is only possible when sending the server command directly.
```
JordanHendersonMusic: This sentence is a bit too fancy. Maybe, 'to customise...' instead?

HotwheelsSisyphus: Thanks! I might remove the sentence anyway once I get @muellmusik's suggestion to work...

HotwheelsSisyphus:
Ok, so I did as @muellmusik suggested. The default action function starting on line 360 is unchanged from the previous version,  I just indented it one level and now the diffs look misleading.

muellmusik:
Thanks for doing this.

> But I figured if the user can write their own parser function, they can simply `s.sendMsg("/g_queryTree", etc.)` and set up their own OSCFunc?

Yes true, though the idea way back in the day was that we would have 'object style' interfaces so that users didn't have to deal directly with OSC messages. Your change is consistent with that, and I think meets the original intention. I have a feeling this was sort of half finished and/or a forgotten discussion.

HotwheelsSisyphus:
@muellmusik, no problem, it was easy enough to do... but maybe I could talk you into reviewing the pr (the class library code mainly, as this is more sensitive)?

HelpSource/Classes/Group.schelp
```diff
@@ -139,10 +139,18 @@ method:: deepFree, deepFreeMsg
...
 
 method:: dumpTree
-Post a representation of this group's node subtree. (Sends a message of the type g_dumpTree.) If code::postControls:: is true, then the current Control (arg) values for any synths contained in this group will be posted as well. The default is false.
+Post a representation of this group's node subtree. (Sends a message of the type teletype::g_dumpTree::.)
+If code::postControls:: is true, then the current Control (arg) values for any synths contained in this group will be posted as well.
+The default is false.
 
 method:: queryTree
-Queries the server for a message describing this group's node subtree. (Sends a message of the type g_queryTree.) This reply is passed to the action function as an argument. See g_queryTree in link::Reference/Server-Command-Reference:: or the example below for information on how the reply is structured.
+Near-duplicate of code::.dumpTree.:: Queries the server for a message describing this group's node subtree.
```
muellmusik: Be slightly better to focus on the difference here, rather than state near duplicate. One just posts to standard out, the other sends a reply which (thanks to you!) can be custom handled. Suggest rewording.

HelpSource/Classes/Group.schelp
```diff
@@ -139,10 +139,18 @@ method:: deepFree, deepFreeMsg
...
+Sends a server message of the type teletype::/g_queryTree::; if no argument is provided, the server's reply is automatically parsed in the post window.
+Alternatively, a link::Classes/Function:: can be provided as an teletype::action:: argument, for instance: 
+code:: 
+Group().queryTree( { |msg| msg.postln });
```
muellmusik: I think we need a better example of how to parse this. But perhaps we can make this easier...

SCClassLibrary/Common/Control/Node.sc
```diff
@@ -350,53 +350,59 @@ AbstractGroup : Node {
...
-									if(msg[i + 4 + j].isMemberOf(Symbol), {
-										(msg[i + 4 + j] ++ ": ").post;
+	queryTree { | action |
+		// FIXME: this only works in scsynth. in supernova, the handle_g_queryTree primitive fails
+		// it needs to be passed an endpoint arg, and this somehow doesn't happen.
+		
+		// when a custom action is provided,
+		// suppress server-failed-to-respond warning by setting done to true
+		var resp, done = action.notNil;
+		action = action ?? {
```
muellmusik, line 360: I'd suggest a better approach here would be to parse this into a more SClang friendly data structure. This could then do the default post, or be passed to the action func. 

This could take a number of forms. A higher level approach would be some form of object designs, but I think maybe nested dictionaries would be better and lightweight.

```
queriedGroupID -> IdentityDictionary[ 
    nestedGroup -> IdentityDictionary[ 
        synthID -> IdentityDictionary[
            \def -> defName,
            \controls -> IdentityDictionary[
                controlName1 -> value,
                controlName2 -> value,
...
                ]
            ],
        ...
    ]
]
```

Or something like that. Sorry I realise that this has ballooned a bit. I can have a go at that in the coming days if you like. Would be worth it to get this right.

muellmusik (changes requested):

HelpSource/Classes/Group.schelp
HotwheelsSisyphus: good point, will do

HelpSource/Classes/Group.schelp
HotwheelsSisyphus: I deliberately used the most simple possible example here; the details of what to do with the msg imho would be better placed in the server command reference than in this file.

I can also reinstate the previous example, but that's just the same as what the class library does, so it's kind of superfluous.

SCClassLibrary/Common/Control/Node.sc
HotwheelsSisyphus: Thanks, tbh I don't want to touch this is this PR, as this was originally intended to be a quick doc fix; but you're very welcome to do so in a separate PR.

muellmusik: Can do. Should we close this one then? I guess it doesn't change much in the end then.

HotwheelsSisyphus: If by close, you mean "approve", then yes. It wasn't meant to change much, just chipping away at small things that seem like easy fixes? Otherwise, this particular method isn't really that important, and it overlaps quite a bit with queryAllNodes, dumpTree etc. Just figured that it's preferable for docs and class library to agree.

muellmusik: Yeah, fair enough. I don't feel the solution for the action is quite what we need though, which I think just leaves some small formatting changes? 

I do appreciate the frustration that can come from holding things up while waiting for related matters to be discussed. I'm travelling for gigs at the moment, but will try to take a look at this now. Otherwise there'd be no harm in merging this, but as a further solution for the action seems likely, I think it's cleaner not to have alternate versions that may never see release in the commit log. So I'd appreciate a little more time to look at this, thanks.

HotwheelsSisyphus: No worries, I don't see it as urgent either, and definitely not high priority. I just can't really contribute on the level of improving the default action... also there's a seemingly endless amount of these half-finished methods on SC, personally I'd rather aim for consistency with docs than the best possible solution.

PR #7447, opened 2026-04-12
update yaml-cpp to 0.9.0

capital-G:
## Purpose and Motivation

Update yaml-cpp to most recent version 0.9.0. Current provided version is 0.5.3 which is 11 years old. The version was probably used to avoid to tie SC to C++11 (since this is the last version to run w/o C++11), but SC has already reached C++17 now.
This should get rid of some compiler warnings regarding deprecation of some C++ methods on some platforms/compilers.

I therefore also updated our fork https://github.com/supercollider/yaml-cpp to include the newest commits upstream.

Once the pipeline for all platforms goes green I think it is fine to merge this.

edit: apparently there are some linking errors w/ windows regarding some DLL exposure? If someone w/ some windows skill/machine can fix this, would be greatly appreciated. Drafted until a fix for windows has been found.

```
 D:\a\supercollider\supercollider\external_libraries\yaml-cpp\src\node_data.cpp(16,40): error C2720: 'YAML::detail::node::m_amount': '__declspec(dllimport)' storage-class specifier illegal on members [D:\a\supercollider\supercollider\build\external_libraries\yaml.vcxproj]
  (compiling source file 'libyamlcpp.cpp')
  
D:\a\supercollider\supercollider\external_libraries\yaml-cpp\src\node_data.cpp(16,40): error C2491: 'YAML::detail::node::m_amount': definition of dllimport static data member not allowed [D:\a\supercollider\supercollider\build\external_libraries\yaml.vcxproj]
  (compiling source file 'libyamlcpp.cpp')
  
D:\a\supercollider\supercollider\external_libraries\yaml-cpp\src\node_data.cpp(18,31): warning C4273: 'YAML::detail::node_data::empty_scalar': inconsistent dll linkage [D:\a\supercollider\supercollider\build\external_libraries\yaml.vcxproj]
  (compiling source file 'libyamlcpp.cpp')
      D:\a\supercollider\supercollider\external_libraries\yaml-cpp\include\yaml-cpp\node\detail\node_data.h(84,29):
      see previous definition of 'YAML::detail::node_data::empty_scalar'

```

## Types of changes

- Maintenance

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation


dyfer:
Thanks @capital-G !

A while back I was trying to update yaml-cpp and encountered this issue as well. I don't have a fix though, but I think we should redo the cmake implementation to add yaml target with `add_subdirectory` instead of adding the file directly. I'll look into it.

Also, do you think we could take this opportunity to switch to using the upstream repo instead of our fork?

capital-G:
I found that the [changelog of 0.8.0](https://github.com/jbeder/yaml-cpp/releases/tag/0.8.0) says

> Significant CMake changes, especially to Windows builds

Also, updating this isn't such a high priority for me anymore - I needed a good JSON lib in C++ for the LSP, but yaml-cpp is subpar for this (see https://github.com/jbeder/yaml-cpp/issues/941, which may explain some strange behavior on our side regarding the json parser of sclang).
For now I am using boost-json.

> Also, do you think we could take this opportunity to switch to using the upstream repo instead of our fork?

Good idea!

dyfer:
> For now I am using boost-json.

In that case, should we consider moving to boost-json altogether?

EDIT: i.e. instead of updating this submodule and updating cmake implementation

capital-G:
> In that case, should we consider moving to boost-json altogether?

Well, yaml is a superset of json, so this probably won't cut it and we have to keep yaml around since we are using it for the language config. And boost doesn't have yaml support. It is a pity that yaml-cpp doesn't have good json support - but it will just be another boost dependency ;)

BTW - the more time I spend w/ yaml, the more I dislike it - it is just too complex b/c it allows for gotos/references/links, has confusing type castings, and most of the time  a "json w/ comments" would be probably sufficient for most use cases.
Most tools use [`toml`](https://toml.io/) these days - it is easy to write a parser for it and is less complex, which also results in safer ways to use toml.
I just found a good blogpost which goes into the problems w/ yaml https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell

dyfer:
One thing that bothered me was that we don't have writing primitives in sclang for either yaml or json (right?). I think separating json parsing and adding write support for it would be an improvement. I think we could move our config to json then maybe? Even the help [indicates](https://doc.sccode.org/Classes/String.html#-parseJSON) that there was at some point a desire to have a proper json-only parser.

But since we probably would need yaml reading support indefinitely, I'll take a look at the build issue here.

capital-G:
> One thing that bothered me was that we don't have writing primitives in sclang for either yaml or json (right?).

I written things down in #7451 - feel free to add

> I think separating json parsing and adding write support for it would be an improvement. I think we could move our config to json then maybe? Even the help [indicates](https://doc.sccode.org/Classes/String.html#-parseJSON) that there was at some point a desire to have a proper json-only parser.

I think using json for config isn't a good choice either b/c there is no way to comment (out) in json.
Since we started using yaml, we probably can't remove it anymore, so maybe just stick w/ it.

dyfer:
@capital-G I used your commit and pushed other changes in https://github.com/supercollider/supercollider/pull/7453. Is it okay to close this?

capital-G:
> @capital-G I used your commit and pushed other changes in #7453. Is it okay to close this?

ofc - thanks for looking into it!

PR #7449, opened 2026-04-14
sclang: fix non-local return

### LINKED ISSUE #232: 'try { loop ... }' doesn't unwind the stack properly if an error occurs within the loop

### LINKED ISSUE #7448: Lang Crashes periodically with what looks like a GC issue
I've seen a few of these now with crashes at ```AllocPool::Alloc(unsigned long)``` or in ```PyrGC```. Below are three dumps from yesterday. EDIT: Now have a reproducer

## Environment

* SuperCollider version: 3.14.1
* Operating system: MacOS 15.7.4
* Other details (Qt version, audio driver, etc.): Qt 6.7.3

## Steps to reproduce
This crashes pretty reliably. Boiled it down from some more complex code I was running.

```(
var bufCol, count = 0;

bufCol = IdentityDictionary();

SkipJack({
	count.postln;
	count = count + 1;
	100.do({try { bufCol[Pdef(\doesNotExist).source[\buf].source.bufnum] } });
}, 0.001);
)
```

Crash dumps: 

https://gist.github.com/muellmusik/e77cce4afaf0053e335445e3e589145f
https://gist.github.com/muellmusik/312fefb54c5d01ff37b28c2005c4d01a
https://gist.github.com/muellmusik/ef1d76ca28ecd462d2380cab7c962283

JordanHendersonMusic:
## Purpose and Motivation

Non local returns don't unwind the stack.

Closes #7448

Closes #232  (I think...)

## Types of changes

- Bug fix
- Breaking change? I don't think so, but technically people might have relied on the borked behaviour.

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/lang/LangSource/PyrInterpreter3.cpp
+++ b/lang/LangSource/PyrInterpreter3.cpp
@@ -39,6 +39,7 @@
 #include <string.h>
 #include <signal.h>
 #include "SpecialSelectorsOperatorsAndClasses.h"
+#include "VMGlobals.h"
 
 #include <float.h>
 #define kBigBigFloat DBL_MAX
@@ -511,14 +512,24 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
 #    pragma GCC optimize("-fno-gcse")
 #endif
 
+// Enable this to print each opcode and the stack size
+#if 0
+#    include <iostream>
+#    define LogOpcode(NAME)                                                                                            \
+        { std::cout << NAME.name << ' ' << g->gc->StackDepth() << '\n'; };
+#else
+#    define LogOpcode(NAME)
+#endif
 
 /// Pass the name of an Opcode struct, e.g. InterpretOpcode(SendMsgX)
 /// Setups of a switch case and a label for the computed gotos.
 #define InterpretOpcode(NAME)                                                                                          \
     case NAME.code:                                                                                                    \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME);
 
-#define InterpretExtendedOpcode(NAME) case Extended::NAME.code:
+#define InterpretExtendedOpcode(NAME)                                                                                  \
+    case Extended::NAME.code:                                                                                          \
+        LogOpcode(NAME);
 
 /// Pass the name of an Opcode struct that has 16 variants in the second nibble, e.g. InterpretOpcode16(PushInstVar)
 /// Setups of a switch case and a label for the computed gotos.
@@ -539,7 +550,7 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
     case NAME.codeOffset<13>():                                                                                        \
     case NAME.codeOffset<14>():                                                                                        \
     case NAME.codeOffset<15>():                                                                                        \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME)
 
 #define InterpretOpcode15(NAME)                                                                                        \
     case NAME.codeOffset<0>():                                                                                         \
@@ -557,7 +568,7 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
     case NAME.codeOffset<12>():                                                                                        \
     case NAME.codeOffset<13>():                                                                                        \
     case NAME.codeOffset<14>():                                                                                        \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME)
 
 #define InterpretOpcode14(NAME)                                                                                        \
     case NAME.codeOffset<0>():                                                                                         \
@@ -574,7 +585,7 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
     case NAME.codeOffset<11>():                                                                                        \
     case NAME.codeOffset<12>():                                                                                        \
     case NAME.codeOffset<13>():                                                                                        \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME)
 
 #define InterpretOpcode7(NAME)                                                                                         \
     case NAME.codeOffset<0>():                                                                                         \
@@ -584,7 +595,7 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
     case NAME.codeOffset<4>():                                                                                         \
     case NAME.codeOffset<5>():                                                                                         \
     case NAME.codeOffset<6>():                                                                                         \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME)
 
 #define InterpretOpcode8(NAME)                                                                                         \
     case NAME.codeOffset<0>():                                                                                         \
@@ -595,7 +606,7 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
     case NAME.codeOffset<5>():                                                                                         \
     case NAME.codeOffset<6>():                                                                                         \
     case NAME.codeOffset<7>():                                                                                         \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME)
 
 #define InterpretOpcode9(NAME)                                                                                         \
     case NAME.codeOffset<0>():                                                                                         \
@@ -607,14 +618,14 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
     case NAME.codeOffset<6>():                                                                                         \
     case NAME.codeOffset<7>():                                                                                         \
     case NAME.codeOffset<8>():                                                                                         \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME)
 
 #define InterpretOpcode4(NAME)                                                                                         \
     case NAME.codeOffset<0>():                                                                                         \
     case NAME.codeOffset<1>():                                                                                         \
     case NAME.codeOffset<2>():                                                                                         \
     case NAME.codeOffset<3>():                                                                                         \
-        opcode_label_##NAME:
+        opcode_label_##NAME : LogOpcode(NAME)
 
 HOT void Interpret(VMGlobals* g) {
     // byte code values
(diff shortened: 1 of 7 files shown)


dyfer:
This is a huge improvement, thanks @JordanHendersonMusic for taking this on!

> Closes #232 (I think...)

Yes, it seems so! At the very least, the reproducer from @telephon now functions as expected: 
```supercollider
x = [nil]; "hello".postf(try { max(x, 1) }); // hello-> hello
```

This is way beyond my understanding for a proper review, but it has been such a long-standing issue and I wanted to express my appreciation for the fix.

testsuite/classlibrary/TestNonLocalReturn.sc
```diff
@@ -0,0 +1,122 @@
...
+		var r = try { nil.(1, 2, 3, this.tail_call_that_throws_with_things_on_stack) };
+
+		this.assertEquals(r, nil);
+
+		after = this.stackDepth;
+		this.assertEquals(before, after, "Non local returns should not leave things on the stack");
+	}
+
+
+	tail_call{ ^1 + 1 }
```
dyfer:
```suggestion
	tail_call { ^1 + 1 }
```

testsuite/classlibrary/TestNonLocalReturn.sc
```diff
@@ -0,0 +1,122 @@
...
+	tail_call_that_throws { ^Error("throwing").throw }
+	tail_call_that_throws_with_things_on_stack {
+		// Yes this is weird because 1 is on the stack.
+		// As of 2026, Function.forkIfNeeded does this.
```
dyfer:
```suggestion
		// As of SC 3.15, Function.forkIfNeeded does this.
```

dyfer (commented):
I just have a couple of minor/formatting suggestions, feel free to disregard

JordanHendersonMusic (commented):
The changes to the allocator method is something @Spacechild1 and I did, it's just copy and pasted from another method. It's a bug I found when doing this.

JordanHendersonMusic:
> This is way beyond my understanding for a proper review, but it has been such a long-standing issue and I wanted to express my appreciation for the fix.

Thanks!

To be honest, there isn't much to review. What happened before was wrong. If you see here https://github.com/supercollider/supercollider/issues/7448#issuecomment-4243038009 I've described exactly what is going wrong.

Really it's a question of whether the tests are adequate and capture all the edge cases. I think I need to add some that return from the block rather than the method. Right now I'm only applying the fix when you return from a method - which I think makes sense, but should check with tests! It might also be good to test yield, and perhaps some custom non-local return use. 

This does open up the possibility of using non-local returns more generally, there are some bizarre things you can do!

dyfer:
> To be honest, there isn't much to review. 

...maybe, if one understands how sclang works internally :)

> Really it's a question of whether the tests are adequate and capture all the edge cases.

Just in case it's related: we also have an issue when e.g. manipulating large ~~dictionaries~~ lists. I don't think I ever filed it because I don't exactly understand what's going on... seems to be related to GC: when the dictionary is above certain size and its contents are being manipulated, it seems to be getting other data... See code and comments below:

<details>

```supercollider
(
var allItems;
var numItems = 1000; // numbers below 300 will likely not cause the error

allItems = numItems.collect({ Dictionary() }).as(List);

// ~allItems = allItems; // !!!!! uncommenting this gets rid of the error...

allItems.do({|thisDict, inc|
	var newDict;

	newDict = Dictionary.newFrom((\inc: inc));
	allItems.add(newDict); // add to the List

	allItems.do({|t|
		if(t.isKindOf(Dictionary).not) {
			"%: detected a non-dictionary entry after add: %".format(inc, t.asCompileString).error;
			// this.halt;
		};
	});

	allItems.remove(thisDict); // current item from the list
	allItems.do({|t|
		if(t.isKindOf(Dictionary).not) {
			"%: detected a non-dictionary entry after remove: %".format(inc, t.asCompileString).error;
			// this.halt;
		};
	});
});
)
```

</details>

JordanHendersonMusic:
There's an issue for that one, was just looking at it!

Yeah... I have no idea what is happening there. I doubt it's GC related though. I think it's stack related but can't find where it goes wrong and why the size has something to do with it.

lnihlen:
Have you tested the performance overhead of this? Otherwise LGTM

lang/LangSource/PyrInterpreter3.cpp
```diff
@@ -511,14 +512,24 @@ static inline void checkStackDepth(VMGlobals* g, PyrSlot* sp) {
 #    pragma GCC optimize("-fno-gcse")
 #endif
 
+// Enable this to print each opcode and the stack size
```
lnihlen: Do you want to leave this in? Should it be in a separate PR?

lnihlen (approved):
Overall LGTM. Let me know if you want help benchmarking this change with LangBench.

lang/LangSource/PyrInterpreter3.cpp
JordanHendersonMusic: Hmm yes should probably be separate because I want to remove all the current print debugging and redo it with a cleaner interface. Most of the old commented-out code doesn't work anymore anyway... Perhaps this could replaced with flags in the build system, or even just a header, just so they are all in once place.

JordanHendersonMusic:
Thanks! No I haven't yet tested it. I'll try your bench some time this week and get back to you on that

telephon (approved):

JordanHendersonMusic:
@lnihlen Well I ran the LangBench... and assuming number_of_runs wants to be high, then I'm confused because either this improves performance by 100%, or something else is happening...

```
// develop
--- start of json ---
{
"awfy.macro.richards": {
        "wall_clock_time": 60.050890193,
        "number_of_runs": 1017
},
"awfy.micro.queens": {
        "wall_clock_time": 60.001482897999,
        "number_of_runs": 13323
},
"awfy.micro.bounce": {
        "wall_clock_time": 60.000706273,
        "number_of_runs": 38802
},
"awfy.micro.permute": {
        "wall_clock_time": 60.004194268999,
        "number_of_runs": 7405
},
"awfy.macro.havlak": {
        "wall_clock_time": 61.722593091,
        "number_of_runs": 16
},
"awfy.micro.storage": {
        "wall_clock_time": 60.002134415003,
        "number_of_runs": 7009
},
"awfy.micro.mandelbrot": {
        "wall_clock_time": 60.000003883042,
        "number_of_runs": 16737422
},
"awfy.micro.nbody": {
        "wall_clock_time": 60.000013532021,
        "number_of_runs": 2726825
},
"awfy.macro.cd": {
        "wall_clock_time": 60.45111086,
        "number_of_runs": 150
},
"awfy.micro.sieve": {
        "wall_clock_time": 60.003033341003,
        "number_of_runs": 15213
},
"awfy.micro.list": {
        "wall_clock_time": 60.000893714004,
        "number_of_runs": 52184
},
"awfy.micro.towers": {
        "wall_clock_time": 60.004773875,
        "number_of_runs": 13270
},
"awfy.macro.deltablue": {
        "wall_clock_time": 60.006101054997,
        "number_of_runs": 6352
}
}
--- end of json ---
```

```
// this pr
--- start of json ---
{
"awfy.macro.richards": {
        "wall_clock_time": 60.014470259,
        "number_of_runs": 1003
},
"awfy.micro.queens": {
        "wall_clock_time": 60.001484546003,
        "number_of_runs": 24307
},
"awfy.micro.bounce": {
        "wall_clock_time": 60.000311722996,
        "number_of_runs": 64244
},
"awfy.micro.permute": {
        "wall_clock_time": 60.000221626998,
        "number_of_runs": 14643
},
"awfy.macro.havlak": {
        "wall_clock_time": 61.082240919,
        "number_of_runs": 26
},
"awfy.micro.storage": {
        "wall_clock_time": 60.000504857,
        "number_of_runs": 12328
},
"awfy.micro.mandelbrot": {
        "wall_clock_time": 60.000001458032,
        "number_of_runs": 26579729
},
"awfy.micro.nbody": {
        "wall_clock_time": 60.000006117006,
        "number_of_runs": 4262146
},
"awfy.macro.cd": {
        "wall_clock_time": 60.115723143,
        "number_of_runs": 336
},
"awfy.micro.sieve": {
        "wall_clock_time": 60.001248423004,
        "number_of_runs": 36473
},
"awfy.micro.list": {
        "wall_clock_time": 60.000000225001,
        "number_of_runs": 62825
},
"awfy.micro.towers": {
        "wall_clock_time": 60.003460929996,
        "number_of_runs": 12470
},
"awfy.macro.deltablue": {
        "wall_clock_time": 60.004091739,
        "number_of_runs": 12231
}
}
--- end of json ---

```

muellmusik:
Thanks @JordanHendersonMusic so much for fixing this longstanding bug! 🙏

Issue #7451, opened 2026-04-17
Improve and fix JSON implementation


capital-G:
## Motivation

This came up in https://github.com/supercollider/supercollider/pull/7447#issuecomment-4264159833 and is probably good to its own issue.

Currently we have a (broken or at least unreliable) way of loading JSON files and no easy way of writing JSON files - e.g. parsing

```json
{
  "x": 42,
  "y": null,
  "z": false
}
```

results in

```supercollider
~j = "{\"x\": 42, \"y\": null, \"z\": false}".parseJSON;

~j[\x] // nil
~j["x"] // 42
~j["x"].class // String (!)

~j["z"].class; // String (!)

~j.keys // -> Set[y, x, z]
// dictionary with nil as value(!)
~j.cs; // -> Dictionary[("z" -> "false"), ("x" -> "42"), ("y" -> nil)]

```

In general, there are some incompatibilities between sclangs dicitionary and the JSON format (which is based on the JavaScript object capabilities).

But there are still points which should be changed:

* The values for JSON should be preserved instead of implicitly converted to strings - though this would be a breaking change, but since the existing implementation is broken, I think we don't have another way to get around this. This could be an upstream bug (see https://github.com/jbeder/yaml-cpp/issues/941 ) - I haven't looked into the implementation on sclang side yet.
* A JSON null should be represented through a custom type, e.g. `JSONNull` which mimics (or even derives?) from `Nil`, since allowing `nil` as a value in a sclang dictionary would be a major breakage.
* Add a `asJSON` method/message which generates a valid JSON

## Plan for Implementation

There is https://github.com/musikinformatik/JSONlib/blob/main/classes/JSONlib.sc but I would write it quite differently today, i.e. using a `asJson` as a message like scztt did in https://github.com/scztt/LanguageServer.quark/blob/develop/JSON.sc, which is more smalltalk-ish IIUC.
Using a primitive would have some advantages such as that it is actually a valid json or easily pretty-printed.

Python has a quite good approach to this IMO by passing values through an optional function for en/decoding at runtime

Custom encoder:
```python
import json
def custom_json(obj):
    if isinstance(obj, complex):
        return {'__complex__': True, 'real': obj.real, 'imag': obj.imag}
    raise TypeError(f'Cannot serialize object of {type(obj)}')

json.dumps(1 + 2j, default=custom_json)
```

Custom decoder
```python
import json
def as_complex(dct):
    if '__complex__' in dct:
[...]

PR #7453, opened 2026-04-17
Update yaml-cpp

dyfer:
## Purpose and Motivation

This PR builds on top of #7447.

I updated CMake integration so that we can just use `add_subdirectory` and I adjusted surrounding cmake code to use the proper target name. I also removed hiding relevant msvc warnings for the old yaml-cpp library (AFAICT, the updated library is not generating these warnings). Finally, I pointed the submodule to the upstream repository.

EDIT: I added small modernization to the cmake implementation so that we link with a target only instead of adding include directories.

## Types of changes

- Maintenance

## To-do list

- [x] Code is tested
- [x] All tests are passing
- n/a Updated documentation


JordanHendersonMusic:
Hmm, I think we should have a suite of JSON tests in the testsuite before touching this stuff. If only so we know what we break/fix. Otherwise, I think this code is good.

dyfer:
I have rebased this PR so that it passes the yaml/json tests I added in https://github.com/supercollider/supercollider/pull/7457

JordanHendersonMusic:
Okay cool, so nothing changes in those test. Is there the possibility that anything might change? Have you looked at the change log between the versions?

capital-G:
> Have you looked at the change log between the versions?

I have reviewed the changelog, and they changed some internals for better typing (requiring C++11 etc) and fixed some minor bugs and security bugs - so shouldn't be a problem. Have in mind that JSON is already broken/unreliable in sclang (see #7451) and needs a proper fix at some point.

JordanHendersonMusic (approved):

JordanHendersonMusic:
> so shouldn't be a problem

That is just what I wanted to check!

PR #7457, opened 2026-04-19
testsuite: test yaml/json conversion

dyfer:
## Purpose and Motivation

Introduce tests for conversion from YAML/JSON.

Suggested in https://github.com/supercollider/supercollider/pull/7453#issuecomment-4273447670

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestString.sc
+++ b/testsuite/classlibrary/TestString.sc
@@ -224,4 +224,67 @@ TestString : UnitTest {
 		var reconstructed = large.asCompileString.interpret;
 		this.assert(large == reconstructed, "A large string's compileString should interpret back to itself");
 	}
+
+	// ------- YAML/JSON -----------------------------------------------
+
+	test_parseYaml_arrays {
+		var yaml = "---\n- first\n- second\n- third";
+		var result = ["first", "second", "third"];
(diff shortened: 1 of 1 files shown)


dyfer:
@capital-G @JordanHendersonMusic what do you think of these tests?

Full transparency: I used an LLM to get me examples of yaml/json to use, as I'm not as familiar with them. Anything we should change or add?

JordanHendersonMusic (approved):
Looks good! Merge away!

JordanHendersonMusic:
I do have questions about things like types here... But if this is just to update the library, it's fine for now.

dyfer:
re: types: I'm not convinced by the current implementation, but these tests just reflect the current state of it. At least we'll know when we change it if we decide to do so.

capital-G:
Thx marcin!

In case you need further testing, here is a more throughout testsuite https://github.com/musikinformatik/JSONlib/blob/main/Tests/TestJSONlib.sc

But we are testing a broken implementation^^ ;)

dyfer:
> But we are testing a broken implementation^^ ;)

Yes we are - for now.

Issue #7182, opened 2025-10-12
Adding items to large Lists sometimes breaks


dyfer:
## Environment

* SuperCollider version: 3.15-dev, 3.14, 3.13
* Operating system: macOS
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

```supercollider
(
var allItems;
var numItems = 2000; // lower numbers, like 300, will likely not cause the error

allItems = numItems.collect({ Dictionary() }).as(List);

// ~allItems = allItems; // !!!!! uncommenting this gets rid of the error...

allItems.do({|thisDict, inc|
	var newDict;
	
	newDict = Dictionary.newFrom((\inc: inc));
	allItems.add(newDict); // add to the List
	
	allItems.do({|t|
		if(t.isKindOf(Dictionary).not) {
			"%: detected a non-dictionary entry after add: %".format(inc, t.asCompileString).error;
		};
	});
	
	allItems.remove(thisDict); // current item from the list
	allItems.do({|t|
		if(t.isKindOf(Dictionary).not) {
			"%: detected a non-dictionary entry after remove: %".format(inc, t.asCompileString).error;
		};
	});
});
)
```

## Expected vs. actual behavior

Expected: list should be populated with Dictionaries only.

Actual: with large lists, other objects leak into the List. The code above posts the message with the object in question being rendered as a compile string:

```755: detected a non-dictionary entry after add: [nil, nil, 'inc', 755]```

(To clarify: I'm posting an error message, it's not an actual error being thrown).

This is an issue that #7180 works around for the test runner.

While I understand that manipulating the List while it's being iterated may be a bad practice... BUT it seems to work unless we hit some memory size. Is this a garbage collection issue?

Also, as noted in the reproducer, assigning the list to an environment variable gets rid of the error... Which I find rather mysterious.

JordanHendersonMusic:
Does this happen on 3.14?

dyfer:
Yes, it's an old bug, I've seen it in the test runner ever since we started running all the tests (albeit infrequently). 
To be sure, I double checked and the reproducer is valid in both 3.13 and 3.14.

JordanHendersonMusic:
Thanks, I will try and figure out what's going on here when I get some time. Redoing kitchen floor this week!

JordanHendersonMusic:
I haven't been able to find the cause, but doing `allItems.copy.do({|thisDict, inc|...` at the top seem to fix the issue.

dyfer:
[...]

PR #7463, opened 2026-04-22
classlib: implement `indexOf` for `List`

JordanHendersonMusic:
## Purpose and Motivation

This fixes an unknown bug #7182 by making `List` defer to its array for indexOf, rather than using a custom `do` loop and equality comparison.

This doesn't actually solve the bug, because I still don't know why it happens. Stack corruption due to unusual extended opcodes? maybe, perhap GC graph related?

Nevertheless, this is a large performance gain for this method.
## Types of changes

- Performance gain
- Bug fix, but more by accident as I still don't know exactly why.

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Collections/List.sc
+++ b/SCClassLibrary/Common/Collections/List.sc
@@ -34,6 +34,7 @@ List : SequenceableCollection {
 	}
 
 	// accessing
+	indexOf { arg item; ^array.indexOf(item) }
 
 	at { arg i; ^array.at(i) }
 	clipAt { arg i; i = i.asInteger.clip(0, this.size - 1); ^array.at(i) }


dyfer:
> Nevertheless, this is a large performance gain for this method.

This looks good in theory, but does it actually improve performance? The code below produces similar result with and without this PR in my tests.

```supercollider
(
bench{
	var allItems;
	var numItems = 20000;
	
	allItems = numItems.collect({ Dictionary() }).as(List);
	
	allItems.do({|thisDict, inc|
		allItems.remove(thisDict); 
		// allItems.removeAt(inc); // this is faster, but still similar between vanilla SC and this PR
	});
}
)
```

If it doesn't improve the performance, do we want to include it? It conceals the related bug, but does not fix it... what do you think?

JordanHendersonMusic:
Hmm I never checked but it's a primitive call Vs using do and a function call. I'd expect quite a drastic performance change. Perhaps a 'tighter' benchmark is needed?

```
l = List( {10000.rand}!999999 );

bench{
   10000.do{ |i|
      l.indexOf(I);
   }
}
```

I'm away from pc right now so can't check until tomorrow.

dyfer:
SORRY! I was modifying the wrong .sc file. This is indeed faster.

```
(
l = List.newFrom( {10000.rand}!999999 );

bench{
   10000.do{ |i|
      l.indexOf(i);
   }
}
)
```
Previous implementation: 5.09s
This PR: 0.05s

JordanHendersonMusic:
> Previous implementation: 5.09s
This PR: 0.05s

Ha! Didn't expect that much!

dyfer:
If we merge this, the reproducer for #7182 will change. I originally thought that we won't be able to reproduce it, but I managed to trigger the issue by changing numItems, putting an array in the dictionary and (!) assigning the list to an environment variable. Putting it here so that #7182 can be amended if we merge this.

```supercollider
( // reproducer for #7182 after merging #7463
var allItems;
var numItems = 5000;

allItems = numItems.collect({ Dictionary() }).as(List);

~allItems = allItems; // this time it's needed to trigger the issue

allItems.do({|thisDict, inc|
	var newDict;

	newDict = Dictionary.newFrom((\inc: inc ! 500));
	allItems.add(newDict); // add to the List

	allItems.do({|t|
		if(t.isKindOf(Dictionary).not) {
			"%: detected a non-dictionary entry after add: %".format(inc, t.asCompileString).error;
			// this.halt;
		};
	});

	allItems.remove(thisDict); // current item from the list
	
	allItems.do({|t|
		if(t.isKindOf(Dictionary).not) {
			"%: detected a non-dictionary entry after remove: %".format(inc, t.asCompileString).error;
			// this.halt;
		};
	});
});
)
````

dyfer (approved):
Now that I found another reproducer for 7182, I think we can merge this. The performance gain is massive!

PR #7466, opened 2026-04-23
 Clarify f.(*array) vs f.(array) in function call examples in SymbolicNotations.schelp

prko:
## Purpose and Motivation

For beginners, clarify the difference between `f.(*array)` and `f.(array)` in the function call examples.

The current example shows how to pass arguments from an array with `*`, but it does not directly contrast that form with passing the array as a normal single argument.

Adding that comparison makes the behavior clearer and helps prevent confusion about what `*` changes in a function call.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Overviews/SymbolicNotations.schelp
+++ b/HelpSource/Overviews/SymbolicNotations.schelp
@@ -314,10 +314,14 @@ definitionlist::
 ## code:: f.(*argList) :: || evaluate the function with the arguments in an array
 ## code:: f.(anArgName: value) :: || keyword addressing of function or method arguments
 code::
-f = { |a, b| a * b };
-f.(2, 4);
-f.(*[2, 4]);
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic (approved):

JordanHendersonMusic:
Thanks! It's a little weirder that that though....


This does nothing.

```
f.(2, *4) \\ [2, 4]
```
It only expand when the class of the object is exactly an array of list. It doesn't work with symbol array for example. 


https://sc.dennis-scheiba.com/c/b8aa6a5bd39f93c5c2ecc71620c0a5ff

I think it should work with pretty much any collection...


You can also change the behaviour....
```
Object{
    performList { ^\meow }
}
f.(*2) // [\meow]
```

But let's not talk about that!

PR #7469, opened 2026-04-24
Correct `Platforam.helpDir`

prko:
## Purpose and Motivation
Currently, `Platform.helpDir` points to a non-existent folder.
This PR changes it so that:

- `Platform.helpDir` points to `Platform.userAppSupportDir +/+ "Help"`
- `Platform.helpSourceDir` points to `Platform.resourceDir +/+ "HelpSource"`

## Types of changes

- Bug fix
- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/Platform.schelp
+++ b/HelpSource/Classes/Platform.schelp
@@ -38,11 +38,10 @@ subsection:: Directories and filesystem stuff
 method:: classLibraryDir
 location of the bundled class library
 
-method:: helpSourceDir
-location of the bundled help files
-
 method:: helpDir
-Location of the HTML help documents rendered from code::Platform.helpSourceDir::
+Returns the system-wide directory containing the help source files.
+This method will be deprecated.
+Use link::Classes/SCDoc#*helpSourceDir:: for the help source files and link::Classes/SCDoc#*helpTargetDir:: for the directories containing the rendered HTML files.
 
 method:: systemAppSupportDir
 system application support directory
--- a/HelpSource/Classes/Platform.schelp
+++ b/HelpSource/Classes/Platform.schelp
@@ -38,8 +38,11 @@ subsection:: Directories and filesystem stuff
 method:: classLibraryDir
 location of the bundled class library
 
-method:: helpDir
+method:: helpSourceDir
 location of the bundled help files
+
+method:: helpDir
+Location of the HTML help documents rendered from code::Platform.helpSourceDir::
 
 method:: systemAppSupportDir
 system application support directory
--- a/SCClassLibrary/Platform/Platform.sc
+++ b/SCClassLibrary/Platform/Platform.sc
@@ -4,7 +4,7 @@ Platform {
 	// IDE actions
 	classvar <>makeServerWindowAction, <>makeSynthDescWindowAction, <>openHelpFileAction, <>openHTMLFileAction;
 
-	var <classLibraryDir, <>recordingsDir, features;
+	var <classLibraryDir, <helpDir, <>recordingsDir, features;
 	var <>devpath;
 
 	*initClass {
--- a/SCClassLibrary/Platform/Platform.sc
+++ b/SCClassLibrary/Platform/Platform.sc
@@ -13,7 +13,7 @@ Platform {
 
 	initPlatform {
 		classLibraryDir = thisMethod.filenameSymbol.asString.dirname.dirname;
-		helpDir = thisMethod.filenameSymbol.asString.dirname.dirname.dirname ++ "/Help";
+		helpDir = thisProcess.platform.userAppSupportDir +/+ "Help";
 		features = IdentityDictionary.new;
 		recordingsDir = this.userAppSupportDir +/+ "Recordings";
 	}
@@ -31,6 +31,7 @@ Platform {
 	// directories
 	*classLibraryDir { ^thisProcess.platform.classLibraryDir }
 	*helpDir { ^thisProcess.platform.helpDir }
+	*helpSourceDir { ^thisMethod.filenameSymbol.asString.dirname.dirname.dirname +/+ "HelpSource"}
 
 	userHomeDir {
 		_Platform_userHomeDir
--- a/SCClassLibrary/Platform/Platform.sc
+++ b/SCClassLibrary/Platform/Platform.sc
@@ -4,7 +4,7 @@ Platform {
 	// IDE actions
 	classvar <>makeServerWindowAction, <>makeSynthDescWindowAction, <>openHelpFileAction, <>openHTMLFileAction;
 
-	var <classLibraryDir, <helpDir, <>recordingsDir, features;
+	var <classLibraryDir, <>recordingsDir, features;
 	var <>devpath;
 
 	*initClass {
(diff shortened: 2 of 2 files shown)


prko:
I think `Platform.helpDir` and `Platform.helpSourceDir` are necessary, but we would need to discuss it.

HelpSource/Classes/Platform.schelp
```diff
@@ -38,9 +38,12 @@ subsection:: Directories and filesystem stuff
 method:: classLibraryDir
 location of the bundled class library
 
-method:: helpDir
+method:: helpSourceDir
 location of the bundled help files
 
+method:: helpDir
```
HotwheelsSisyphus, line 41:
```suggestion
method:: helpTargetDir
```

HotwheelsSisyphus: and add a line
```suggestion
private:: helpDir
```
so it's no longer shown in the docs (but deliberately leave it in .sc for backward compatibility)

SCClassLibrary/Platform/Platform.sc
```diff
@@ -31,6 +31,7 @@ Platform {
 	// directories
 	*classLibraryDir { ^thisProcess.platform.classLibraryDir }
 	*helpDir { ^thisProcess.platform.helpDir }
+	*helpSourceDir { ^thisMethod.filenameSymbol.asString.dirname.dirname.dirname +/+ "HelpSource"}
```
HotwheelsSisyphus:
```suggestion
	*helpSourceDir { ^SCDoc.helpSourceDir }
```

HotwheelsSisyphus (commented):
Thanks, but there is already SCDoc.helpTargetDir..., and in the interest of maintainability we shouldn't be defining _more_ paths. 

See my suggestions, where Platform takes the values of these var from SCDoc... 
I think helpTargetDir might be a better name than helpDir, but this can be discussed.

HotwheelsSisyphus:
Ah whoops, sorry I'll have to rethink this - my (now-deleted) suggestions don't work because SCDoc takes these paths from Platform to begin with. 

So it would be better to initialize all those paths in the Platform.initClass, and adapt SCDoc accordingly.

prko:
@HotwheelsSisyphus Thanks!

When I opened this PR, I momentarily forgot about `SCDoc.helpSource` and `SCDoc.helpTarget`. I have renamed the PR and corrected `Platform.helpDir`. We do not need `Platform.helpSource`.

HotwheelsSisyphus:
Ok, I thing this has gotten a bit confused, sorry for causing some of this confusion!

Three separate things to settle:
1. Do we want to keep .helpDir around? Since it was broken up to now, we might as well remove it. The CI probably fails because one of us removed the variable and there's still some other bit of code that wants to call that variable; which we can identify and change.
2. If we keep it, (I'm neutral on this) should it point to the "Source" or "Target" dir for help? I think the "target" maybe makes more sense, because that dir (in the file system) is actually named `/Help`, whereas HelpSourceDir is, well, `helpSource`.  
3.  A question of dependency/initialization: There should ideally be _one_ place where all these dirs are first initialized, and the Platform class is probably the right one; all other classes should init the respective dir variables _from_ Platform (to prevent disagreement between classes about the paths). This would involve a change to SCDoc.sc (which doesn't necessarily need to go into this PR).

prko:
> 1. Do we want to keep .helpDir around?

I do not think it is particularly useful, as it seems to point to a non-existent folder that may date back to before the SC-IDE was introduced.

> 2. If we keep it, (I'm neutral on this) should it point to the "Source" or "Target" dir for help?

From its name, I would expect it to refer to the same folder as `SCDoc.helpTargetDir`.
However, the help documentation may suggest that it should refer to the same folder as `SCDoc.helpSourceDir`, especially because it uses the word “bundled”:
https://github.com/supercollider/supercollider/blob/a27ec6c01ac78fba2967c136ba8cfc94414d1c61/HelpSource/Classes/Platform.schelp#L41-L42

> 3. A question of dependency/initialization: There should ideally be one place where all these dirs are first initialized, and the Platform class is probably the right one; all other classes should init the respective dir variables from Platform (to prevent disagreement between classes about the paths). This would involve a change to SCDoc.sc (which doesn't necessarily need to go into this PR).

I agree with that view.

---

It would probably be worth hearing what other developers think as well.

telephon:
> > 1. Do we want to keep .helpDir around?
> 
> I do not think it is particularly useful, as it seems to point to a non-existent folder that may date back to before the SC-IDE was introduced.

It shouldn't be deprecated, I think. We will be able to keep using it. I think it should point to the html versions of the rendered files.

In older versions of sc it pointed to the folder of the rtf-files with the help (so to the files that can be read using extenrnal applications as well as sc).

HotwheelsSisyphus:
> It shouldn't be deprecated, I think. We will be able to keep using it. 

I'm not sure how to read this statement. (Did you mean to write "we should" rather than "we will"?) 

Off Topic: (At any rate, deprecation, afaik, does _not_ mean we wouldn't be able to keep using it, but rather that there would be a warning to the effect of "if you write new code in 2026, please use [alternative method/dir] instead". 
It's true that most software eventually completely eliminates the deprecated bits, but sc isn't most software, so we could just have them stick around forever and throw warnings. And have a global language setting somewhere for a "legacy" mode that disables these warnings...?)

At any rate, if neither SCDoc.helpTargetDir nor Platform.helpDir are deprecated, this would mean that there are two names for the same path; imho, this isn't ideal; one would like new users to at least be told which one is the more commonly used name these days.

telephon:
In this particular case, deprecation may be not too bad, since the methods are probably used not so much in code but more in set ups. 

But otherwise, depecation is often not worth the noise it causes: in other words, the overhead of having two ways of doing something is less than the overhead to organise the changes across time and people.

capital-G:
> In this particular case, deprecation may be not too bad, since the methods are probably used not so much in code but more in set ups.
> 
> But otherwise, depecation is often not worth the noise it causes: in other words, the overhead of having two ways of doing something is less than the overhead to organise the changes across time and people.

As a proposal: should we move deprecations which we will not planning to remove/break from being a runtime warning to a documentation thing (i.e. by adding a `deprecated::` primitive as seen in  https://github.com/supercollider/supercollider/issues/7474 ), which then could be displayed in the help file and also add a comment to existing code in the ide (via e.g. a LSP)?

See e.g. in rust

grafik

This doesn't introduce noise while running code (which is probably really annoying if it is not "my" code which is creating the warning but a dependency), but nudges the user who is editing the source file to "fix it" and also avoids people to pick it up from then on.

JordanHendersonMusic:
> As a proposal: should we move deprecations which we will not planning to remove/break from being a runtime warning to a documentation thing (i.e. by adding a deprecated:: primitive as seen in #7474 ), which then could be displayed in the help file and also add a comment to existing code in the ide (via e.g. a LSP)?

I think this should be done with pragmas.

```
A {
   ##[deprecated]
   foo { ... }
}
```
This way, we can toggle whether a warning is printed, and the language server can simply query the functiondef - imo, it shouldn't parse the docs.

dyfer:
We can discuss deprecation mechanics in a new issue.
How to proceed with this PR?

PR #7470, opened 2026-04-25
ClassLibrary/lang: Add mtime instance method to File (fixes #2526)

### LINKED ISSUE #2526: [feature request] File:mtime as instance method
I'm going to break my own rule about feature requests and muse aloud: why, if I've already opened a File, do I have to do this to get its modified timestamp?

```
File.mtime(myFile.path)
```

instead of:

```
myFile.mtime
```

?

Super low-priority, but IMO valid.

catchafallingstar:
Purpose and Motivation
Fixes #2526.

This PR adds an mtime instance method to the File class, allowing users to query the modification time of an open file object directly. Previously, this was only available as a class method requiring a file path.

Types of changes
New feature

To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Files/File.sc
+++ b/SCClassLibrary/Common/Files/File.sc
@@ -109,6 +109,10 @@ File : UnixFILE {
 		fileptr = nil;
 		openFiles.remove(this);
 	}
+	mtime {
+		_FileInstanceMTime
+		^this.primitiveFailed
+	}
 	length { // returns the length of the file
 		_FileLength;
 		^this.primitiveFailed;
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
I think it would be better to implement this in terms of `File.mtime`. Generally, it's best to not add primitives unless absolutely necessary. What do you think?

dyfer:
@catchafallingstar are you interested in incorporating feedback for this PR?

catchafallingstar:
Yes! Sorry I was looking for jobs earlier this month, I will continue working on this!

dyfer:
ping @catchafallingstar 

As @JordanHendersonMusic suggested, a class library implementation may be sufficient... and it would be much simpler, something like:

```supercollider
	mtime {
		^File.mtime(this.realpath)
	}
```

PR #7471, opened 2026-04-25
Try to make Buffer method arguments (`numChannels`, `index`, and `numFrames`-related arguments) consistent

### LINKED ISSUE #7462: Fractional numFrames causes Buffer:getToFloatArray to fail
## Environment

* SuperCollider version: any current (this part of the code hasn't changed in a long time)
* Operating system: any
* Other details (Qt version, audio driver, etc.): not relevant

## Steps to reproduce

```supercollider
// 6890.625 = 10 * 44100 / 64
b = Buffer.alloc(s, 6890.625, 10);

// actually, you don't need any data
a = { RecordBuf.ar(Array.fill(10, { WhiteNoise.ar }), b, loop: 0, doneAction: 2); Silent.ar(1) }.play;

b.getToFloatArray(wait: -1, timeout: 10, action: { |data| data.postln });

FAILURE IN SERVER /b_getn index out of range
WARNING: Buffer-streamToFloatArray failed!

b.free;
b = Buffer.alloc(s, 6890, 10);

b.getToFloatArray(wait: -1, timeout: 10, action: { |data| data.postln });

(no failure)
```

There's not a good reason for the one to fail and the other to succeed; they should both succeed.

The question is whether there is any benefit to preserving the fraction in the buffer object, when the server-side buffer can't have a fractional number of frames. I think at one point there was an argument that the client side buffer object should keep the fraction (which is why we cast to integer in buffer messages but not in buffer constructors), but since there are other ramifications, maybe we should revisit that.

prko:
fixes #7462 

## Purpose and Motivation

I tried adding `.asInteger` in all possible places, but it produced the following error:
```
ERROR: Message 'asInteger' not understood.
RECEIVER:
   nil
ARGS:
KEYWORD ARGUMENTS:

PROTECTED CALL STACK:
	Meta_MethodError:new	0x1533ed940
		arg this = DoesNotUnderstandError
		arg what = nil
		arg receiver = nil
	Meta_DoesNotUnderstandError:new	0x1533efd00
		arg this = DoesNotUnderstandError
		arg receiver = nil
		arg selector = asInteger
		arg args = []
		arg keywordArgumentPairs = []
	Object:doesNotUnderstand	0x1747c7640
		arg this = nil
		arg selector = asInteger
		arg args = nil
		arg kwargs = nil
	Meta_Buffer:new	0x159cd9680
		arg this = Buffer
		arg server = localhost
		arg numFrames = nil
		arg numChannels = nil
		arg bufnum = 0
		arg sampleRate = 48000.0
	a FunctionDef	0x15adc6e80
		arg i = 0
	Integer:do	0x31148dbc0
		arg this = 100
		arg function = a Function
		var i = 0
	Meta_Collection:fill	0x15adc6a80
		arg this = Array
		arg size = 100
		arg function = a Function
		var obj = []
	Meta_SayBuf:prepRRBufs	0x17342e900
		arg this = SayBuf
		arg server = localhost
		arg numBufs = 100
	Function:doOnServerBoot	0x152a371c0
		arg this = a Function
		arg server = localhost
	ArrayedCollection:do	0x329363a00
		arg this = [a Function, a Function]
		arg function = a Function
		var i = 1
	List:do	0x32932ed00
		arg this = List[a Function, a Function]
		arg function = a Function
	Meta_AbstractServerAction:performFunction	0x1731c6080
		arg this = ServerBoot
		arg server = localhost
		arg function = a Function
	Meta_AbstractServerAction:run	0x1731c6540
		arg this = ServerBoot
		arg server = localhost
		var selector = doOnServerBoot
	a FunctionDef	0x329469a00
	Routine:prStart	0x15a374900
		arg this = a Routine
		arg inval = 42657.560064459

CALL STACK:
	< closed FunctionDef >
	Function:try
		var result = <instance of DoesNotUnderstandError>
	< FunctionDef in Method Scheduler:seconds_ >
		arg time = 42657.560064459
		arg item = <instance of Routine>
	< FunctionDef in Method SequenceableCollection:pairsDo >
		arg i = 0
	Integer:forBy
		arg this = 0
		arg endval = 0
		arg stepval = 2
		var i = 0
		var j = 0
		arg this = [*2]
	Scheduler:seconds_
		arg newSeconds = 42657.560113542
	Meta_AppClock:tick
	Process:tick
		arg this = <instance of Main>

^^ ERROR: Message 'asInteger' not understood.
RECEIVER: nil

Message with a similar name understood by the receiver:
	isInteger

Many other objects respond to the message 'asInteger' (found 9 superclasses).
```
After removing `.asInteger` from `Buffer.new`, I was able to avoid this particular error. However, the same error still occurred when evaluating `.free` (an instance method of `Buffer`).

So, in the end, I applied this change only to `Buffer.alloc`.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Control/Buffer.sc
+++ b/SCClassLibrary/Common/Control/Buffer.sc
@@ -60,14 +60,14 @@ Buffer {
 
 	allocRead { arg argpath, startFrame = 0, numFrames = -1, completionMessage;
 		path = argpath;
-		this.startFrame = startFrame;
-		server.listSendMsg(this.allocReadMsg(argpath, startFrame, numFrames.asInteger, completionMessage))
+		this.startFrame = startFrame.asInteger;
+		server.listSendMsg(this.allocReadMsg(argpath, startFrame.asInteger, numFrames.asInteger, completionMessage))
 	}
 
 	allocReadChannel { arg argpath, startFrame = 0, numFrames = -1, channels, completionMessage;
 		path = argpath;
-		this.startFrame = startFrame;
-		server.listSendMsg(this.allocReadChannelMsg(argpath, startFrame, numFrames.asInteger, channels,
+		this.startFrame = startFrame.asInteger;
+		server.listSendMsg(this.allocReadChannelMsg(argpath, startFrame.asInteger, numFrames.asInteger, channels,
 			completionMessage))
 	}
 
@@ -79,14 +79,14 @@ Buffer {
 	allocReadMsg { arg argpath, startFrame = 0, numFrames = -1, completionMessage;
 		this.cache;
 		path = argpath;
-		this.startFrame = startFrame;
+		this.startFrame = startFrame.asInteger;
 		^["/b_allocRead", bufnum, path, startFrame.asInteger, (numFrames ? -1).asInteger, completionMessage.value(this)]
 	}
 
 	allocReadChannelMsg { arg argpath, startFrame = 0, numFrames = -1, channels, completionMessage;
 		this.cache;
 		path = argpath;
-		this.startFrame = startFrame;
+		this.startFrame = startFrame.asInteger;
 		completionMessage !? { completionMessage = [completionMessage.value(this)] };
 		^["/b_allocReadChannel", bufnum, path, startFrame.asInteger, (numFrames ? -1).asInteger] ++ channels ++ completionMessage
 	}
@@ -98,7 +98,7 @@ Buffer {
 		bufnum ?? { bufnum = server.nextBufferNumber(1) };
 		^super.newCopyArgs(server, bufnum)
 			.doOnInfo_(action).cache
-			.allocRead(path, startFrame, numFrames.asInteger, {|buf|["/b_query", buf.bufnum] })
+			.allocRead(path, startFrame.asInteger, numFrames.asInteger, {|buf|["/b_query", buf.bufnum] })
 	}
 
 	read { arg argpath, fileStartFrame = 0, numFrames = -1, bufStartFrame = 0, leaveOpen = false, action;
@@ -115,7 +115,7 @@ Buffer {
 		bufnum ?? { bufnum = server.nextBufferNumber(1) };
 		^super.newCopyArgs(server, bufnum)
 			.doOnInfo_(action).cache
-			.allocReadChannel(path, startFrame, numFrames.asInteger, channels,
+			.allocReadChannel(path, startFrame.asInteger, numFrames.asInteger, channels,
 				{|buf|["/b_query", buf.bufnum]})
 	}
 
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks for the PR! 
Sorry, but the PR is not very clear as to which cases it does and does not fix. The issue also does not clearly lists all the cases where the issue occurs.
- could you provide a possibly comprehensive list of examples that fail due to float arguments in the Buffer class? (this could also go in the issue if you prefer)
- could you specify which of those are actually being fixed in this PR in its current state?

prko:
> could you provide a possibly comprehensive list of examples that fail due to float arguments in the Buffer class? (this could also go in the issue if you prefer)

Not at the moment, unfortunately. I may have run into similar problems in the past, but I did not file issues for them at the time, and I do not remember the specific cases well enough to list them now. So I do not currently have a comprehensive list of failing cases.

When I took a closer look at the `Buffer` documentation, my initial assumption was that frame-index-related arguments and `numFrames`-related arguments should not accept arbitrary non-integer float values. Values like `100.0` seem reasonable, since they are effectively integers written in float form, whereas values like `100.4`, `1000.8`, or `6890.625` seem questionable in this context.

However, based on many parts of the `Buffer` documentation (though not all), these values appear to be truncated automatically, so this seems to be accepted behaviour in many cases.

While trying to improve the approach I described in my comment on issue #7462, I ran into a new error that I was not able to resolve, as described in the opening post of this PR.

To keep this PR focused and easier to review, I decided not to include those broader changes here.

> could you specify which of those are actually being fixed in this PR in its current state?

At the moment, this PR only fixes the error reproduced by the example code in #7462:

```supercollider
// 6890.625 = 10 * 44100 / 64
b = Buffer.alloc(s, 6890.625, 10);

// actually, you don't need any data
a = { RecordBuf.ar(Array.fill(10, { WhiteNoise.ar }), b, loop: 0, doneAction: 2); Silent.ar(1) }.play;

b.getToFloatArray(wait: -1, timeout: 10, action: { |data| data.postln });
```

If I get the chance, I would be happy to revisit the broader issue in a separate follow-up.

---

I left many of the commits in place to show the steps I went through, the error I was not able to resolve, and the changes I had tried. However, I realise that this may have obscured what this PR actually fixes. Sorry about that.

dyfer:
> I left many of the commits in place to show the steps I went through, the error I was not able to resolve, and the changes I had tried.

Got it!

I think it would be good to trace this back and apply consistently where needed. Just leaving this comment here in case someone is able to pick this up. Let's keep this open for now, though eventually I'd prefer it is there was a comprehensive PR addressing all or most cases in the Buffer class.

PR #7472, opened 2026-04-26
Add `asInteger` to `numFrames` & `numChannels` of `Buffer.alloc` & `Buffer.getToFloatArray`

### LINKED ISSUE #7462: Fractional numFrames causes Buffer:getToFloatArray to fail

prko:
fixes #7462 

## Purpose and Motivation

This PR resolves the error triggered by evaluating the following example code from #7462:

```supercollider
b = Buffer.alloc(s, 6890.625, 10);

// actually, you don't need any data
a = { RecordBuf.ar(Array.fill(10, { WhiteNoise.ar }), b, loop: 0, doneAction: 2); Silent.ar(1) }.play;

b.getToFloatArray(wait: -1, timeout: 10, action: { |data| data.postln });
```
However, if this example code is not expected to throw an error without truncating `6890.625`, then I have misunderstood the issue, and this PR should probably be closed.

As far as I understand it, there are two possible ways to address the issue:

1. [Add `.asInteger` to `var count` in the `getToFloatArray` method](https://github.com/supercollider/supercollider/commit/6c39387de2c8a66b3d5f753e6ac972755abb10ed).  
   At the moment, the implementation applies `.asInteger` to `(numFrames * numChannels)`, which may lead to an incorrect calculation.

2. [Add `.asInteger` to `numFrames` and `numChannels` in `Buffer.alloc`](https://github.com/supercollider/supercollider/commit/ac71924dfa49ec9e10f1878eb20beafa240ce8b3).  
   The documentation says that `numFrames` and `numChannels` are truncated, but the corresponding method does not currently do this. This commit would make the implementation consistent with the documentation.

I put the second approach in a separate PR (#7471), but my impression is that it is better left as a draft for now. That is why I opened this new PR instead.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Control/Buffer.sc
+++ b/SCClassLibrary/Common/Control/Buffer.sc
@@ -312,7 +312,7 @@ Buffer {
 		pos = index = index.asInteger;
 		// treat -1 and nil the same
 		if(count == -1 || count.isNil) {
-			count = (numFrames * numChannels).asInteger - index;
+			count = (numFrames.asInteger * numChannels.asInteger).asInteger - index;
 		};
 		array = FloatArray.newClear(count);
 		refcount = (count / 1633).roundUp;
--- a/SCClassLibrary/Common/Control/Buffer.sc
+++ b/SCClassLibrary/Common/Control/Buffer.sc
@@ -31,8 +31,8 @@ Buffer {
 		^super.newCopyArgs(
 			server,
 			bufnum,
-			numFrames,
-			numChannels,
+			numFrames.asInteger,
+			numChannels.asInteger,
 			sampleRate
 		).alloc(completionMessage).cache
 	}


prko:
I would appreciate feedback on whether this PR could be merged after making the necessary changes, or if it should be closed due to a misunderstanding.

PR #7473, opened 2026-04-28
Add `ratiocent` and `centratio` methods

prko:
## Purpose and Motivation
This PR adds `ratiocent` and `centratio`.

Ideally, as with the other unary operators, these methods would be implemented as primitives. In that case, the SC methods would delegate to corresponding primitives such as `_RatioCent` and `_CentRatio`. Doing so would also require adding the new method names to the lexer and providing the corresponding C++ implementations.

For the moment, however, I have taken a simpler approach by reusing the existing midiratio and ratiomidi methods.

My impression is that this may not be regarded as the preferred long-term implementation. Even so, I thought it might still be useful to put together a minimal working version first and offer it as a starting point for discussion.

## Types of changes

- Documentation
- New feature
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/AbstractFunction.schelp
+++ b/HelpSource/Classes/AbstractFunction.schelp
@@ -121,6 +121,8 @@ a = Pseries(220, 220, inf).cpsmidi; a.asStream.nextN(12); // overtone series as 
 
 method::midiratio
 method::ratiomidi
+method::centratio
+method::ratiocent
 method::ampdb
 method::dbamp
 method::octcps
--- a/HelpSource/Classes/SequenceableCollection.schelp
+++ b/HelpSource/Classes/SequenceableCollection.schelp
@@ -753,7 +753,7 @@ subsection::Math Support - Unary Messages
 
 All of the following messages send the message link::#-performUnaryOp:: to the receiver with the unary message selector as an argument.
 
-method::neg, reciprocal, bitNot, abs, asFloat, ceil, floor, frac, sign, squared, cubed, sqrt, exp, midicps, cpsmidi, midiratio, ratiomidi, ampdb, dbamp, octcps, cpsoct, log, log2, log10, sin, cos, tan, asin, acos, atan, sinh, cosh, tanh, rand, rand2, linrand, bilinrand, sum3rand, distort, softclip, coin, even, odd, isPositive, isNegative, isStrictlyPositive, real, imag, magnitude, magnitudeApx, phase, angle, rho, theta, asFloat, asInteger
+method::neg, reciprocal, bitNot, abs, asFloat, ceil, floor, frac, sign, squared, cubed, sqrt, exp, midicps, cpsmidi, midiratio, ratiomidi, centratio, ratiocent, ampdb, dbamp, octcps, cpsoct, log, log2, log10, sin, cos, tan, asin, acos, atan, sinh, cosh, tanh, rand, rand2, linrand, bilinrand, sum3rand, distort, softclip, coin, even, odd, isPositive, isNegative, isStrictlyPositive, real, imag, magnitude, magnitudeApx, phase, angle, rho, theta, asFloat, asInteger
 
 method::performUnaryOp
 Creates a new collection of the results of applying the selector to all elements in the receiver.
--- a/HelpSource/Classes/SimpleNumber.schelp
+++ b/HelpSource/Classes/SimpleNumber.schelp
@@ -953,6 +953,14 @@ method:: ratiomidi
 Convert a ratio to an interval in semitones.
 returns:: an interval in semitones
 
+method:: centratio
+Convert an interval in cents to a ratio.
+returns:: a ratio
+
+method:: ratiocent
+Convert a ratio to an interval in cents.
+returns:: an interval in cents
+
 method:: ampdb
 Convert a linear amplitude to decibels.
 
--- a/HelpSource/Overviews/Operators.schelp
+++ b/HelpSource/Overviews/Operators.schelp
@@ -108,6 +108,12 @@ Convert an interval in MIDI notes into a frequency ratio.
 method:: ratiomidi
 Convert a frequency ratio to an interval in MIDI notes.
 
+method:: centratio
+Convert an interval in cents into a frequency ratio.
+
+method:: ratiocent
+Convert a frequency ratio to an interval in cents.
+
 method:: dbamp
 Convert decibels to linear amplitude.
 
--- a/SCClassLibrary/Common/Collections/SequenceableCollection.sc
+++ b/SCClassLibrary/Common/Collections/SequenceableCollection.sc
@@ -793,6 +793,8 @@ SequenceableCollection : Collection {
 	midicps { ^this.performUnaryOp('midicps') }
 	cpsmidi { ^this.performUnaryOp('cpsmidi') }
 	midiratio { ^this.performUnaryOp('midiratio') }
+	ratiocent { ^this.performUnaryOp('ratiocent') }
+	centratio { ^this.performUnaryOp('centratio') }
 	ratiomidi { ^this.performUnaryOp('ratiomidi') }
 	ampdb { ^this.performUnaryOp('ampdb') }
 	dbamp { ^this.performUnaryOp('dbamp') }
--- a/SCClassLibrary/Common/Core/AbstractFunction.sc
+++ b/SCClassLibrary/Common/Core/AbstractFunction.sc
@@ -48,6 +48,8 @@ AbstractFunction {
 	cpsmidi { ^this.composeUnaryOp('cpsmidi') }
 	midiratio { ^this.composeUnaryOp('midiratio') }
 	ratiomidi { ^this.composeUnaryOp('ratiomidi') }
+	ratiocent { ^this.composeUnaryOp('ratiocent') }
+	centratio { ^this.composeUnaryOp('centratio') }
 	ampdb { ^this.composeUnaryOp('ampdb') }
 	dbamp { ^this.composeUnaryOp('dbamp') }
 	octcps { ^this.composeUnaryOp('octcps') }
--- a/SCClassLibrary/Common/Core/Symbol.sc
+++ b/SCClassLibrary/Common/Core/Symbol.sc
@@ -124,6 +124,8 @@ Symbol {
 	cpsmidi { ^this }
 	midiratio { ^this }
 	ratiomidi { ^this }
+	centratio { ^this }
+	ratiocent { ^this }
 	ampdb { ^this }
 	dbamp { ^this }
 	octcps { ^this }
--- a/SCClassLibrary/Common/Math/Signal.sc
+++ b/SCClassLibrary/Common/Math/Signal.sc
@@ -235,6 +235,8 @@ Signal[float] : FloatArray {
 	//cpsmidi { _CPSMIDI; ^this.primitiveFailed }
 	//midiratio { _MIDIRatio; ^this.primitiveFailed }
 	//ratiomidi { _RatioMIDI; ^this.primitiveFailed }
+	//centratio { _MIDIRatio; ^this.primitiveFailed }
+	//ratiocent { _RatioMIDI; ^this.primitiveFailed }
 	//ampdb { _AmpDb; ^this.primitiveFailed }
 	//dbamp { _DbAmp; ^this.primitiveFailed }
 	//octcps { _OctCPS; ^this.primitiveFailed }
--- a/SCClassLibrary/Common/Math/SimpleNumber.sc
+++ b/SCClassLibrary/Common/Math/SimpleNumber.sc
@@ -25,6 +25,8 @@ SimpleNumber : Number {
 	cpsmidi { _CPSMIDI; ^this.primitiveFailed }
 	midiratio { _MIDIRatio; ^this.primitiveFailed }
 	ratiomidi { _RatioMIDI; ^this.primitiveFailed }
+	centratio { ^(this / 100).midiratio }
+	ratiocent { ^(this.ratiomidi * 100) }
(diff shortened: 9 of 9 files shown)


capital-G (changes requested):
Thanks - IMO it is fine to do this within the language and w/o a primitive - makes the source code more transparent for the user and we can always later decide to create optimization for this.
The only question is if we need the `centratio` / `ratiocent` method - IMO this is fine b/c cent is so essential (although it is just `/100`).

I'd appreciate some tests for it though - I don't doubt that your implementation is correct, but it is just good to test as much as possible so we know that we don't break anything here in the future, see e.g.

https://github.com/supercollider/supercollider/blob/a27ec6c01ac78fba2967c136ba8cfc94414d1c61/testsuite/classlibrary/TestSimpleNumber.sc#L259-L264

and

https://github.com/supercollider/supercollider/blob/a27ec6c01ac78fba2967c136ba8cfc94414d1c61/testsuite/classlibrary/TestCoreUGens.sc#L277-L284

Otherwise this is fine and ready to be merged after the tests has been added.

---

I thought that we should maybe have some way to (automatically) document that certain methods were only introduced in e.g. 3.15 - b/c I'd be hesitant to use newer methods in Quarks for compatibility issues. I'll make an issue about this.

HotwheelsSisyphus:
I don't know what others think, but personally, I disagree with this addition (though not strongly); I would consider this to be "method bloat". The existing conversion methods are justified because they aren't linear and involve offsets that are difficult to remember; but if we add .ratiocent, we might as well add things such as  `SimpleNumber.multiplyByOneHundred`, implementation `{ |x| ^(x * 100) }` or `SimpleNumber.addOne`, implementation `{ |x| ^(x + 1) }`... I don't mean to be mean, I just don't see how this adds to anyone's quality of life. (So at the very least, an example of where this would be better then just writing out `.ratiomidi * 100` would be appreciated).

dyfer:
I also am not convinced this is a needed addition - multiplication/division by 100 is simple enough IMO.
But if others think this is useful, I won't strongly oppose.

prko:
@capital-G
Thank you for the feedback. I’ll add the test code to UnitTest.sc once the other developers agree.

---

@HotwheelsSisyphus 

> So at the very least, an example of where this would be better then just writing out .ratiomidi * 100 would be appreciated.

@dyfer 
>I also am not convinced this is a needed addition - multiplication/division by 100 is simple enough IMO.

SuperCollider already provides `ratiomidi` and `midiratio`, so `ratiocent` and `centratio` wouldn’t introduce any new mathematical capabilities. Cent conversions can already be done using logarithms or MIDI conversions:

```supercollider
// Existing approaches
(var ratio = 2; ratio.ratiomidi * 100;)
(var ratio = 2; 1200 * log2(ratio);) // shorter and clearer than using ratiomidi

(var cents = 1200; (cents / 100).midiratio;)
(var cents = 1200; 2.pow(cents / 1200);) // shorter and clearer than using midiratio
```

The question is whether cent-based operations are common enough in musical practice to justify dedicated methods. Here are a few reasons why I believe they are:

**1. Semantic clarity**
Tuning work often deals with cents directly. Existing approaches require an extra translation step, which can obscure the musical intent:

```supercollider
~commaSyntonicRatio = 81/80;

// Indirect
~commaSyntonicCents = ~commaSyntonicRatio.ratiomidi * 100;

// Direct
~commaSyntonicCents = ~commaSyntonicRatio.ratiocent;
~commaSyntonicCents = 1200 * log2(~commaSyntonicRatio); // uses a mathematical expression rather than an sclang method
```
The direct version expresses the musical idea much more clearly.

**2. Symmetry in usage**
*(This perspective was suggested by an AI assistant, but I think it is worth considering.)*
`ratiocent` and `centratio` form a clean inverse pair:

```supercollider
// Proposed:
ratio.ratiocent
cents.centratio

// Current:
ratio.ratiomidi * 100
(cents / 100).midiratio
// or
1200 * log2(ratio)
2.pow(cents / 1200)
```
This symmetry improves readability and reduces the chance of small calculation errors.

**3. Workflow cohesion**
In longer examples or tutorials, cent-focused code becomes significantly easier to follow:

```supercollider
(
// 1. Define commas as ratios
~commaSyntonicRatio = 81/80;
~commaSchismaRatio = 32805/32768;

// 2. Convert to cents
~commaSyntonicCents = ~commaSyntonicRatio.ratiocent;
~commaSchismaCents = ~commaSchismaRatio.ratiocent;

// 3. Convert cents back to ratios for DSP
~detunedPair = { |cents|
    var ratio = cents.centratio;
    // (cents / 100).midiratio;
    // 2.pow(cents / 1200);
    SinOsc.ar(440 *[1, ratio]) * 0.1
};
)

x = ~detunedPair.play(args: [\cents, ~commaSchismaCents]);
x.release;
```
This keeps the code strictly within the vocabulary of tuning practice.

**4. Usability (HCI perspective)**
For users who primarily think in cents, having to mentally apply `* 100` or `/ 100` adds a small but recurring cognitive burden. Dedicated methods would remove that friction. 

Of course, users who prefer MIDI-based workflows can continue using the existing methods. Some may prefer working directly with `midiratio` and `ratiomidi`, treating MIDI values as quantities where the integer part represents semitones and the fractional part represents cent deviations. However, a value like `1.000625`—representing 1 semitone plus 0.0625 cents—is arguably less readable than expressing the exact same idea directly in cents. In this sense, `ratiocent` and `centratio` offer a much more intuitive alternative.

*(As an aside, while writing this, I wondered if names like `ratiosemitone` or `semiratio` might be clearer than `ratiomidi` and `midiratio`—though I do not plan to open a separate PR for that.)*

**Conclusion**
While `ratiocent` and `centratio` do not fill a mathematical gap, they greatly improve clarity and reduce cognitive load in cent-oriented tuning code. Whether this ergonomic benefit justifies adding the methods is ultimately up for discussion.

Regarding the implementation itself, instead of routing through MIDI:
```supercollider
centratio { ^(this / 100).midiratio }
ratiocent { ^(this.ratiomidi * 100) }
```
I suggest the following to explicitly show the mathematical relationship between ratio and cents:
```supercollider
ratiocent { ^(1200 * log2(this)) }
centratio { ^(2.pow(this/1200)) }
```

prko:
Erm...,I realised that the following code
```
(
~commaSyntonicRatio = 81/80;
~commaSyntonicCents = ~commaSyntonicRatio.ratiocent;

~detunedPair = { |cents|
    var ratio = cents.centratio;
    SinOsc.ar(440 * [1, ratio]) * 0.1
};
)

x = ~detunedPair.play(args: [\cents, ~commaSyntonicCents]);
x.release;
```
returns the following error:
```
-> a Function
ERROR: Operator 'centratio' applied to a UGen is not supported in scsynth

PROTECTED CALL STACK:
	BasicOpUGen:operator_	0x112454c80
		arg this = an UnaryOpUGen
		arg op = centratio
	UnaryOpUGen:init	0x112456f80
		arg this = an UnaryOpUGen
		arg theOperator = centratio
		arg theInput = an OutputProxy
	a FunctionDef	0x113bbcd98
		sourceCode = "{ |cents|
    var ratio = cents.centratio;
    // (cents / 100).midiratio;
    // 2.pow(cents / 1200);
    SinOsc.ar(440 * [1, ratio]) * 0.1
}"
		arg cents = an OutputProxy
		var ratio = nil
	SynthDef:buildUgenGraph	0x379c75140
		arg this = SynthDef:temp__42
		arg func = a Function
		arg rates = nil
		arg prependArgs = []
		var result = nil
		var saveControlNames = [ControlName  P 0 i_out scalar 0]
	a FunctionDef	0x136f3d700
		arg i_out = an OutputProxy
		var result = nil
		var rate = nil
		var env = nil
	SynthDef:buildUgenGraph	0x379c75140
		arg this = SynthDef:temp__42
		arg func = a Function
		arg rates = nil
		arg prependArgs = []
		var result = nil
		var saveControlNames = nil
	a FunctionDef	0x379c73a40
	Function:prTry	0x381c0a2c0
		arg this = a Function
		var result = nil
		var thread = a Thread
		var next = nil
		var wasInProtectedFunc = false
	
CALL STACK:
	Exception:reportError
		arg this = <instance of Error>
	Nil:handleError
		arg this = nil
		arg error = <instance of Error>
	Thread:handleError
		arg error = <instance of Error>
	Object:throw
		arg this = <instance of Error>
	Function:protect
		var result = <instance of Error>
	SynthDef:build
		arg this = <instance of SynthDef>
		arg ugenGraphFunc = <instance of Function>
		arg rates = nil
		arg prependArgs = nil
	Function:play
		arg target = <instance of Group>
		arg outbus = 0
		arg fadeTime = 0.02
		arg addAction = 'addToHead'
		arg args = [*2]
		var server = <instance of Server>
		var def = nil
		var synth = nil
		var bytes = nil
		var synthMsg = nil
	< closed FunctionDef >  (no arguments or variables)
		var res = nil
		var code = "x = ~detunedPair.play(args: ..."
		var doc = nil
	Process:interpretPrintCmdLine
		arg this = <instance of Main>
^^ The preceding error dump is for ERROR: Operator 'centratio' applied to a UGen is not supported in scsynth

```
However, the following works:
```
(
~commaSyntonicRatio = 81/80;
~commaSyntonicCents = ~commaSyntonicRatio.ratiocent;
~detunedPair = { |cents|
    var ratio = (cents / 100).midiratio;
    SinOsc.ar(440 * [1, ratio]) * 0.1
};
)

x = ~detunedPair.play(args: [\cents, ~commaSyntonicCents]);
x.release;
```

---

Anyway, I’ll go with the others’ opinion!

capital-G:
I also had the feeling that this may be "bloat", at the same time we already have plenty of *lightweight* functions (probably not the best argument) and also that cent is a pretty common in a musical context.
If prko thinks this would be a good addition, I think it is fine to add it.

edit: One advantage of having a method for this is the ability to chain methods, e.g.
```smalltalk
// requires a bracket
(30.ratiomidi * 100).reciprocal
// method chaining
30.ratiocent.reciprocal
```

Regarding the UGen error: Maybe `ratiocent { ^this.performUnaryOp('ratiocent') }` in `SequenceableCollection.sc` should rather be `^(this.ratiomidi * 100)` and vice versa - but I don't know about the specifics regarding simplenumber/signal/ugen methods.

@prko I have a feeling that you used AI for formulating https://github.com/supercollider/supercollider/pull/7473#issuecomment-4343107396 I personally would prefer less usage of AI - it may seem perfect but verbose - and I prefer non-perfect but personal, makes interacting more fun :)

JordanHendersonMusic:
Since cents is just a linear factor of midi, I don't think it's needed. If it was something more complex than a multiple, then I'd agree. 

Just thinking out loud, if chaining is a concern... I'm pretty confident we could add this to the language `2.*(2)`. I'll look into it more at some point. It's currently invalid syntax so should be okay.

prko:
@capital-G 
Thank you for your opinions!

---

Here are my thoughts, written in my own wording (AI is used only to control English oddities):
- From the moment of opening this PR until yesterday:
   - I proposed this PR from the user's convenience (I think this is an HCI perspective, considering UX) to construct code from the perspectives of musicians and music theorists.
   - The coding experience is simpler with this than using `midiratio` and `ratiomidi`: less typing, fewer characters, and simplicity of completed code. Reducing typing increases productivity and make the workflows smoother.
  - Readability of extremely small values in cents is easier:
      -  `0.125.centratio` vs `0.00125.midiratio`
      -  `1.0001.ratiocent`: `-> 0.17311474931345`
         `1.0001.ratiomidi`: `-> 0.0017311474931345`
   - While writing this PR, I thought changing the name from `ratiomidi` to `ratiosemitone`, `ratiosemi` or `ratiost` is semantically better, and vice versa. However, I do not want to open a new PR regarding this.
- Currently:
   Using the formulae directly is not bad. 
  - It is shorter than using `midiratio` and `ratiomidi`. 
  - The formulae are fundamental knowledge when learning basic theory required to know this field:
  - examples:
    - `1200 * log2(3/2)`
      `(3/2).ratiocent`
      `(3/2).ratiomidi * 100`
    - `2.pow(20 / 1200)`
      `20.centratio`
      `(20 / 100).midiratio`

telephon:
I think for method chaining it is practical. No big argument required, we just need to decide how much it is needed. The example really is something like `(2/100).midiratio` vs `2.centratio`. Something extra needs to be learned ("what does cent mean"), and if what is learned is good, why not.

jamshark70:
> ERROR: Operator 'centratio' applied to a UGen is not supported in scsynth

It needs an AbstractFunction implementation as well.

Adding a new backend math operator implementation in the server is complicated, but a pseudo-ugen implementation would be fine in this case (and I'd suggest piggybacking off of `midiratio` to reduce the number of ugens).

I personally think that `centratio` / `ratiocent` is edging a bit in the direction of bloat.

prko:
Regarding @jamshark70's point about the UGen implementation, it seems not to be difficult to resolve (e.g., by adding an `AbstractFunction` implementation). **Thanks for pointing that out!**

However, before updating the code, I think the priority is to decide whether we actually want to keep this PR alive based on its conceptual value.

I believe one of the strengths of higher‑level programming languages is that they allow us to focus on the conceptual intent of a task without having to express all the underlying mathematical steps explicitly. In this sense, adding methods such as `centratio` and `ratiocent` helps bring the musical/construction idea to the foreground rather than the numerical operations required to achieve it. This provides a clearer and more convenient interface for practical work, and I think it represents a meaningful improvement to the language.

(If we agree on this direction, we could even consider adding `midicent` and `centmidi` later to complete the vocabulary, though that is secondary for now.)

 If there is a consensus that this ergonomic benefit is worthwhile, I will gladly fix the UGen issue and continue with the PR. However, I sense that adding these methods might not be widely supported at the moment. **If there are no further positive opinions in the coming weeks, I will go ahead and close this PR. Thanks!**

PR #7477, opened 2026-05-01
cmake: fix sclang final build

dyfer:
## Purpose and Motivation

Unity build aka final build for the sclang target was previously broken.

I fixed it by excluding Bison file from it, adding some `#pragma once` guards and moving one cpp definition to a header file.
I also used the modern cmake functionality to make the unity build - I plan to do this throughout the project in separate PRs.

I think that cmake and preprocessor changes are straightforward, but please partiularly review the `dInfinity = std::numeric_limits<double>::infinity();` change - I'm not sure if this is the best way to do it and it was suggested by a LLM. Also, is there a better place to put it?

How I tested this:
```sh
cmake -D FINAL_BUILD=ON ..
cmake --build . --target sclang --config RelWithDebInfo
```

EDIT: we don't make unity builds in CI - should we? maybe one of the linux builds?

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/QtCollider/primitives/primitives.h
+++ b/QtCollider/primitives/primitives.h
@@ -19,6 +19,8 @@
  *
  ************************************************************************/
 
+#pragma once
+
 #include <PyrPrimitive.h>
 #include <VMGlobals.h>
 #include <PyrSlot.h>
--- a/common/SC_Version.hpp.in
+++ b/common/SC_Version.hpp.in
@@ -18,6 +18,8 @@
  *    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
  */
 
+#pragma once
+
 #include <string>
 #include <sstream>
 #include <cstdint>
--- a/lang/CMakeLists.txt
+++ b/lang/CMakeLists.txt
@@ -187,11 +187,14 @@ endif()
 include(../SCDoc/CMakeLists.txt)
 list(APPEND sclang_sources ${SCDOC_SRCS})
 
-if(0 AND FINAL_BUILD) # sclang final-builds are broken
-	CREATE_FINAL_FILE(libsclang_final.cpp ${sclang_sources})
-	add_library(libsclang STATIC libsclang_final.cpp ${sclang_parser_source})
-else()
-	add_library(libsclang STATIC ${sclang_sources} ${sclang_parser_source})
+add_library(libsclang STATIC ${sclang_sources} ${sclang_parser_source})
+
+if(FINAL_BUILD)
+	set_source_files_properties(${sclang_parser_source} PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+	set_target_properties(libsclang PROPERTIES 
+		UNITY_BUILD ON
+		UNITY_BUILD_BATCH_SIZE 16 
+	)
 endif()
 
 target_compile_definitions(libsclang PRIVATE YYSTACK_USE_ALLOC)
--- a/lang/LangPrimSource/OSCData.cpp
+++ b/lang/LangPrimSource/OSCData.cpp
@@ -618,8 +618,6 @@ static PyrInt8Array* MsgToInt8Array(sc_msg_iter& msg, bool runGC) {
     return obj;
 }
 
-static const double dInfinity = std::numeric_limits<double>::infinity();
-
 // Convert raw OSC message to Array.
 static PyrObject* ConvertOSCMessage(int inSize, const char* inData) {
     if ((inSize & 3) != 0) {
--- a/lang/LangPrimSource/PyrSched.cpp
+++ b/lang/LangPrimSource/PyrSched.cpp
@@ -55,7 +55,6 @@
 #    include <sys/time.h>
 #endif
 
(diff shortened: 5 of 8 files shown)


JordanHendersonMusic:
> dInfinity = std::numeric_limits<double>::infinity();

This is fine, the move from static to inline would change the address I believe. But we only care about the.value.

 I'd love to.review this... But I have no idea what this is supposed to do!

dyfer:
> I'd love to.review this... But I have no idea what this is supposed to do!

Do you mean - what my change is supposed to do or what `dInfinity = std::numeric_limits<double>::infinity()` is supposed to do?

JordanHendersonMusic:
No, what is a unity build and why would you use one?

dyfer:
My understanding is that unity build is meant to speed up compilation times (but not incremental builds) by combining all sources into a single file (well, in case of the cmake implementation, in a few larger files, according to batch size, IIUC). 

Do we actually need it? I'm not sure. We have it as an option in our cmake and it was broken (for sclang), now it works.

I guess adding pragma once is a reasonable cleanup regardless, right?

We could remove unity build function altogether, if we don't think it's worthwhile. Otherwise, we should update the implementation to modern cmake (as I proposed in this PR).

JordanHendersonMusic (approved):
I don't really see the point of these unity builds as comp times don't seem to be that slow down me. But the changes look fine, so merge away if you think it's worth while!

dyfer:
Thanks @JordanHendersonMusic !

@capital-G any opinion on whether we should keep unity builds around? I'm not sure if I remember correctly, but I thought that @jrsurge advocated for them as some point (sorry if I'm misremembering, James).

capital-G:
> @capital-G any opinion on whether we should keep unity builds around?

Some relevant commits:

* https://github.com/dyfer/supercollider/commit/80196bb3ca54f349a3c8ca539b5dfc8613e360d7
* https://github.com/dyfer/supercollider/commit/ff71d3b1909ed86ee38cbab68c20908d26aee5df

Seems it was broken for 15 years and only worked for 1 year?
If it was broken and nobody remembers what it is actually used for I think we could remove it - maybe wait for 1 week for jrsurge?
Was this more a clean-up procedure or are you interested in this unity/final build?

dyfer:
> Seems it was broken for 15 years and only worked for 1 year?

Only for the `sclang` target. `scsynth` and `supernova` targets have working unity builds (albeit using a 15-year-old implementation). Also it is currently disabled for the IDE (probably needs similar fixes like sclang) and it seems it was never implemented for the plugins.

https://github.com/dyfer/supercollider/actions/runs/25260776067/job/74070177191#step:9:1

> Was this more a clean-up procedure or are you interested in this unity/final build?

Hm, a little of both, but started as a clean up definitely.
I'm not sure if people are actually using it.
It seems that for an uncached build, it seems to cut down build time in CI on Linux from [8 minutes](https://github.com/dyfer/supercollider/actions/runs/25230066959/job/73983289405#step:9:1) to [6 minutes](https://github.com/dyfer/supercollider/actions/runs/25260776067/job/74070177191#step:9:1)(*) and I guess we would get a bit more if we add unity build functionality to the ide and the plugins. Still, most speed benefit comes from using ccache (that cuts build time sometimes to under 1 minute on Linux), though it doesn't always work (cache expires etc).

Honestly, I'm not sure what to do about it. We could:
- fix them and switch most builds in CI to unity builds to save resources
- fix them and switch only a few builds in the CI to unity builds to keep testing whether they work
- remove the functionality altogether from the cmake
- leave it in, but not test - i.e. keep it potentially broken (I'd prefer not to do that though)

FWIW it's also currently broken on Windows, I'd look into it if we want to maintain it.

(*) here's a [branch](https://github.com/dyfer/supercollider/tree/sclang-final-build-default) where I turned on unity builds `ON` by default on all platforms.

JordanHendersonMusic:
That's not a bad build time increase, I'm going to be busy for the next few days, but using ftime-trace  https://clang.llvm.org/docs/analyzer/developer-docs/PerformanceInvestigation.html might be a good idea to see where build time is actually spent.

capital-G:
Can such a build still be debugged? I could imagine that combining everything to one file could create some difficulties of mapping source code to instruction(?)

jrsurge:
I'd recommend at least experimenting with unity builds, particularly for CI builds.

It's an approach for reducing the number of translation units, which can cut down unnecessary recompilations. It can also help the linker by having fewer things to link.

It can expose things like ODR violations, and transitive includes, but can make internal linkage more difficult to reason about (things that were shielded by a translation unit boundary maybe aren't anymore, etc.)

Unity builds should be optional - locally, incremental builds might prove more efficient. It's important for developers to be able to switch, and CI should likely have the switch turned on.

They can still be debugged, yeah.

In short, there are tradeoffs, but worth looking at

dyfer:
I started looking at fixing sclang msvc unity build and it's a bit more tricky due to macro clashes. I got it to work but I want to see if there's a way to do it better, that will be a separate PR.

I propose to merge this PR so that we can move forward exploring unity builds. Is that okay with everyone?

dyfer:
Merging!

PR #7479, opened 2026-05-01
build: enable supernova on all platforms by default

dyfer:
## Purpose and Motivation

Supernova previously was enabled by default only on Linux/BSD, but not on macOS or Windows, although we have been providing supernova in the releases for the latter platforms for a while.

This PR changes the default so that supernova is built by default on all platforms.

One minor build change: I also changed the default on macOS to build portaudio from source. This removes the need to additionally install portaudio from homebrew.
Note that we have already been using the built-in portaudio (for supernova) in macOS releases for some time - it's much easier to obtain a universal binary that way. It's just that in this PR it becomes a default setting.

I checked the main macOS and Windows builds and supernova is still included.

## Types of changes

- New feature ? not really

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [x] Updated documentation

### DIFF
--- a/README_MACOS.md
+++ b/README_MACOS.md
@@ -46,9 +46,6 @@ Prerequisites:
 - **git, cmake >= 3.12, libsndfile, readline, and qt6 >= 6.2**, installed via homebrew:
   `brew install git cmake libsndfile readline qt@6`
 
-- If you want to build with the *supernova* server, you need **portaudio** package, which can also be installed via homebrew:
-  `brew install portaudio`
-
 Obtaining the source code
 -------------------------
 
@@ -67,8 +64,6 @@ Build instructions
     mkdir -p build
     cd build
     cmake -G Xcode ..
-    # or, if you want to build with supernova:
-    cmake -G Xcode -DSUPERNOVA=ON ..
     # then start the build
     cmake --build . --target install --config RelWithDebInfo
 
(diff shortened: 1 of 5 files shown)


README_MACOS.md
```diff
@@ -221,10 +216,10 @@ Common arguments to control the build configuration are:
 
   * Build the *supernova* server:
 
-    `-DSUPERNOVA=ON`
+    `-DSUPERNOVA=ON # ON by default`
 
-    Using supernova requires the `portaudio` audio backend, so you need to install it
-    (Homebrew and MacPorts both provide packages).
+    Using supernova requires the `portaudio` audio backend, which will be built from source by default.
```
capital-G: Maybe we should "reverse" this now that the default has changed? e.g.

> Starting with 3.15, supernova will be also built as default in macOS.
> To not build supernova, set the configure variable `-DSUPERNOVA=OFF`.
>
> Using supernova requires the portaudio audio backend, which will be built from source by default. In order to use portaudio installed via Homebrew, additionally set the -DSYSTEM_PORTAUDIO=ON flag.
> 
> Note: Supernova adds an alternative server executable and a supernova version of each plugin is built. If you also use the sc3-plugins package, make sure to compile them with supernova support too.
>
> Within SC you will be able to switch between scsynth and supernova by evaluating one of:
> 
> ```supercollide
> Server.supernova; // use supernova
> Server.scsynth; // use scsynth - default
> s.boot; // start the server
> ```
>
>Check sc help for `ParGroup` to see how to make use of multi-core hardware.

capital-G (approved):
Technical changes looks good, though I think we should adapt the readme a bit more.
I added a suggestion how the docs could be modified accordingly.
Feel free to ignore my suggestion though.

README_MACOS.md
dyfer: Thanks! I adapted your suggestions.

PR #7481, opened 2026-05-02
build/sclang check parser errors in pre commit

JordanHendersonMusic:
## Purpose and Motivation

Does as the name suggests, add a pre-commit hook that runs when the grammar file is changed, if there are any errors/warning, it fails.

Move the shell script to tools/.

Also adds the shell script as a vscode task.

Also remove unused rules, these changes to the grammar are trivial, no need to review as this feature was already removed in another PR, this was just missed (which isn't possible now with the pre-commit hook).

This turns any reduce/shift warnings into errors, meaning we can be that little bit more confident about grammar changes — an incy wincy bit more confident...

## Types of changes

- Build step
- Remove redundant grammar rule

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,8 @@ repos:
 
     -   id: bison
         name: Run Bison parser checking for any errors
-        entry: tools/generate_parser_bison.sh
-        files: lang/LangSource/Bison/lang11d
+        # NOTE: this command is a copy from tools/generate_parser_bison.sh
+        entry: bison -Werror --defines=lang/LangSource/Bison/lang11d_tab.h --output=lang/LangSource/Bison/lang11d_tab.cpp lang/LangSource/Bison/lang11d
+        files: lang/LangSource/Bison/*
         pass_filenames: false
-        language: script
+        language: system
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@ repos:
           - "c++"
         language: python
         additional_dependencies: ['clang-format==14.0.6']
+
     -   id: clang_lint
         name: Lint C++ code
         entry: python tools/clang-format.py lint
@@ -22,9 +23,17 @@ repos:
           - "c++"
         language: python
         additional_dependencies: ['clang-format==14.0.6']
+
     -   id: sc_empty_blank_lines
         name:  Check trailing blank lines in sclang
         entry: python tools/remove_trailing_blank_lines.py
         pass_filenames: true
         files: '\.sc((d|help)?)$'
         language: python
+
+    -   id: bison
+        name: Run Bison parser checking for any errors
+        entry: tools/generate_parser_bison.sh
+        files: lang/LangSource/Bison/lang11d
+        pass_filenames: false
+        language: script
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,10 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Generate Parser With Bison",
+            "type": "shell",
+            "command": "tools/generate_parser_bison.sh"
+        }
+    ]
+}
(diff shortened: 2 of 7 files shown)


JordanHendersonMusic:
@capital-G is this pre-commit okay? I haven't written one before.

capital-G:
> @capital-G is this pre-commit okay? I haven't written one before.

There are currently two problems:

1) This only runs on unix systems (shell script)
2) This requires bison as a dependency

So anyone editing the file on windows would need to turn off pre-commit for that commit and need bison installed (btw - is the bison-output version dependent?). The latter is probably fine, but we could move the command from the shell script also into pre-commit and leave a comment to sync those?

Otherwise I think this is fine.
Have you tested if pre-commit detects an outdated `lang11d_tab.cpp` using this?

JordanHendersonMusic:
> This requires bison as a dependency

I *think* (don't know anything about this) that this only runs when you edit the grammar. This means you need bison to generate the parser, so this isn't an issue as you always want to run bison after editing the file.

> This only runs on unix systems (shell script)

> but we could move the command from the shell script also into pre-commit and leave a comment to sync those?

I'd much rather have one source of truth on this, you should never turn the errors off (why this isn't the default....).

So this means that windows users could never run the script they always had to type this out...

What's the best way to make this work on windows and Unix then?

> Have you tested if pre-commit detects an outdated lang11d_tab.cpp using this?

No I haven't, not too sure how to do that! Any thoughts?

capital-G:
> > This requires bison as a dependency
> 
> I _think_ (don't know anything about this) that this only runs when you edit the grammar. This means you need bison to generate the parser, so this isn't an issue as you always want to run bison after editing the file.

Yes, therefore I think it is fine as mentioned.

> > but we could move the command from the shell script also into pre-commit and leave a comment to sync those?
> 
> I'd much rather have one source of truth on this, you should never turn the errors off (why this isn't the default....).
> 
> So this means that windows users could never run the script they always had to type this out...
> 
> What's the best way to make this work on windows and Unix then?

Worse - this would probably fail and they could never complete the commit - even when everything is as it should be, since sh is not available and therefore the command fails.

I think the best is, as mentioned, to duplicate the command since there is no unified scripting language. You could try to cramp it into a python script, since we already have a dependency for python w/ pre-commit, but I think to just duplicate the command (as shell and in pre-commit) is the easiest way - add a comment to it and it should stay somehow in sync.

> > Have you tested if pre-commit detects an outdated lang11d_tab.cpp using this?
> 
> No I haven't, not too sure how to do that! Any thoughts?

Edit `lang/LangSource/Bison/lang11d`, stage the file and run `pre-commit run` - this should now fail since the files are out of sync.

Should we maybe also add `lang/LangSource/Bison/lang11d_tab.cpp` as a file which triggers the pre-commit check? Currently it would be possible to edit `lang11d_tab.cpp` and therefore de-sync it w/ `lang11d` (you shouldn't do that in the first place, but a check wouldn't hurt?), which I think is the intention of this check.

JordanHendersonMusic:
Hmm, another option would be to make bison a required dependency and generate the files using cmake.

This would mean the generated files could actually be excluded from the git history.

JordanHendersonMusic:
> Should we maybe also add lang/LangSource/Bison/lang11d_tab.cpp as a file which triggers the pre-commit check? Currently it would be possible to edit lang11d_tab.cpp and therefore de-sync it w/ lang11d (you shouldn't do that in the first place, but a check wouldn't hurt?), which I think is the intention of this check.

I've made it run the check when ever anything in the Bison directory changes. This solves this issue.

I think everything else is addressed?

capital-G (approved):
Once the merge conflict has been resolved, this can be merged.

I am rather curious if now a `pre-commit run --all-files` will always run the `bison` parser - but maybe this is not such a big issue after all, bison doesn't seem like such a big/complex dependency - we could always remove it if it creates problems. Important bit would be, that its output is deterministic across versions.

JordanHendersonMusic:
> Important bit would be, that its output is deterministic across versions.

So I don't think it is right now, but that's because we are using it in an unsupported way (global variables). I've fixed a bit of this with the lexer, next step should finish this.

Then the behaviour should be the same, even if the code differs.

PR #7483, opened 2026-05-08
Remove 255-arg limit for SynthDefs

jamshark70:
## Purpose and Motivation

https://scsynth.org/t/increasing-the-limits-classvars-function-args-and-serveroptions/13234 pointed out that, even when using NamedControls, SynthDefs are limited to 255 control names, not by anything structural in SynthDef, but rather due to an optimization in SynthDesc: the way that msgFunc uses function arguments to access event keys en masse.

An optimization should not impose limits that would otherwise not be present. (labeling this as a bug fix for that reason).

This change clumps control names into blocks of 250.

- Normal case: If <= 250, the behavior is unchanged. `msgFunc` is generated exactly as before.
- Large-def case: If > 250, `makeMsgFunc` constructs multiple message functions and uses an outer-level function to merge their results into a single array.

```
// normal case: no change
SynthDescLib.at(\default).msgFunc.postcs; ""

#{ arg out, freq, amp, pan;
	var	x9218BE44 = Array.new(8);
	out !? { x9218BE44.add('out').add(out) };
	freq !? { x9218BE44.add('freq').add(freq) };
	amp !? { x9218BE44.add('amp').add(amp) };
	pan !? { x9218BE44.add('pan').add(pan) };
	x9218BE44
}

// big case
SynthDef('300_controls', {
	var ctl = Control.names(
		Array.fill(300, { |i| ("ctl" ++ i).asSymbol })
	).kr(Array.fill(300, { |i| i/300 }));
	Out.ar(0, DC.ar(0))
}).add;

SynthDescLib.at('300_controls').msgFunc.postcs; ""

#{
	var yAD7B1D38 = Array(600);
	yAD7B1D38 = yAD7B1D38.addAll(#{ arg ctl0, ctl1, ctl2, /* snip... */ ctl248, ctl249;
	var	xAD7B1D38 = Array.new(500);
	ctl0 !? { xAD7B1D38.add('ctl0').add(ctl0) };
	ctl1 !? { xAD7B1D38.add('ctl1').add(ctl1) };
	ctl2 !? { xAD7B1D38.add('ctl2').add(ctl2) };
	/* snip... */
	ctl248 !? { xAD7B1D38.add('ctl248').add(ctl248) };
	ctl249 !? { xAD7B1D38.add('ctl249').add(ctl249) };
	xAD7B1D38
}.valueEnvir);
	yAD7B1D38 = yAD7B1D38.addAll(#{ arg ctl250, ctl251, /* snip... */ ctl298, ctl299;
	var	xAD7B1D38 = Array.new(100);
	ctl250 !? { xAD7B1D38.add('ctl250').add(ctl250) };
	ctl251 !? { xAD7B1D38.add('ctl251').add(ctl251) };
	/* snip... */
	ctl298 !? { xAD7B1D38.add('ctl298').add(ctl298) };
	ctl299 !? { xAD7B1D38.add('ctl299').add(ctl299) };
	xAD7B1D38
}.valueEnvir);
	yAD7B1D38;
}
```

Large case: One `#{ ... }.valueEnvir` covers the first 250 control names, then another, and another if needed. All control names are covered (the unit test verifies this), and it's still using the optimized `valueEnvir` approach rather than a `do` loop. (Yes, it looks dumb when printed out, but readability isn't an issue, and it's good to avoid the overhead of `do`.)

In the normal case, there should be no performance impact on event.play (because `msgFunc` has not changed at all). Of course, for a large number of control names, performance would be slower anyway. I think this design minimizes performance impact.

Also, we never had a unit test for msgFunc, so I added one here.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation n/a

### DIFF
--- a/SCClassLibrary/Common/Audio/SynthDesc.sc
+++ b/SCClassLibrary/Common/Audio/SynthDesc.sc
@@ -436,9 +436,11 @@ SynthDesc {
 	}
 
 	makeMsgFunc {
-		var	string, comma=false;
+		var	string, strings, comma = false;
 		var	names = IdentitySet.new,
 		suffix = this.hash.asHexString(8);
+		var index = 0, count;
+		var return;
 		// if a control name is duplicated, the msgFunc will be invalid
 		// that "shouldn't" happen but it might; better to check for it
 		// and throw a proper error
@@ -451,11 +453,15 @@ SynthDesc {
 					.format(name).warn;
 					comma = true;
 				} {
-					names.add(name);
+					if(msgFuncKeepGate or: { name != \gate }) {
+						names.add(name);
+					};
 				};
 			};
 		});
-		if(names.size > 255) { Error("A synthDef cannot have more than 255 control names.").throw };
+		if(names.size == 0) {
+			^ #{ Array.new }
+		};
 		// reusing variable to know if I should continue or not
 		if(comma) {
 			"\nYour synthdef has been saved in the library and loaded on the server, if running.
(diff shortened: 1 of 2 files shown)


SCClassLibrary/Common/Audio/SynthDesc.sc
```diff
@@ -436,82 +436,124 @@ SynthDesc {
 	}
 
 	makeMsgFunc {
-		var	string, comma=false;
```
JordanHendersonMusic, line 439: There is some odd whitespace after the 'var' here

SCClassLibrary/Common/Audio/SynthDesc.sc
```diff
@@ -436,82 +436,124 @@ SynthDesc {
...
+					}
+				} {
+					controlName = controlNames[index + scanned];
+					name = controlName.name.asString;
+					if (name != "?") {
+						if (name == "gate") {
+							hasGate = true;
+							if(msgFuncKeepGate) {
+								if (comma) { stream << ", " } { comma = true };
+								stream << name;
```
JordanHendersonMusic: This code is heavily indented! Can you think of an easy way to reduce this?

testsuite/classlibrary/TestSynthDescMsgFunc.sc
```diff
@@ -0,0 +1,100 @@
+TestSynthDescMsgFunc : UnitTest {
```
JordanHendersonMusic, line 1: Given that we haven't tested this before, do you think it might be worth testing whether these high numbered arguments work? Is there a limit on the server's side?

JordanHendersonMusic (changes requested):
This looks good, thanks! A few queries.

You could add the commit that does the formatting to the list of ignored commits in the blame. Not a big deal though.

Note: Do not squash.

SCClassLibrary/Common/Audio/SynthDesc.sc
jamshark70: Leftover from the Cocoa editor days, I think... fixed.

SCClassLibrary/Common/Audio/SynthDesc.sc
jamshark70: > Can you think of an easy way to reduce this?

Actually, yes... `case` is basically `elif` in other languages. Also I removed one while loop (we can do both the argument list and the function body at the same time).

testsuite/classlibrary/TestSynthDescMsgFunc.sc
jamshark70: > testing whether these high numbered arguments work?

They do.

> Is there a limit on the server's side?

I can't see a practical limit. "Synth Definition File Format" says "int32 - number of parameters (P)" and "int32 - number of parameter names (N)" but of course you'll hit UDP packet size limits long before that.

Concrete test:

```
(
bench {
	var n = 5000;
	SynthDef(\bigctl, {
		var ctl = Control.names(
			Array.fill(n, { |i| ("ctl" ++ i).asSymbol })
		).kr(Array.fill(n, { |i| 0 }));
		Out.kr(1000, ctl.sum)
	}).add;
}
)
time to run: 7.2861927690001 seconds.
// ^^ lol -- so already nobody is going to do this

x = Synth(\bigctl);

Bus(\control, 1000, 1, s).getSynchronous
-> 0.0

x.set(\ctl4999, 1.0);

Bus(\control, 1000, 1, s).getSynchronous
-> 1.0
```

So the server has successfully set the last of 5000 control names.

JordanHendersonMusic (approved):
Thanks!

jamshark70:
Wait just a sec, don't merge yet, I noticed something.

jamshark70:
OK, I think it's fixed. I inadvertently dropped an if condition at the end. I think it would have been possible for it to generate a msgFunc like `#{ arg ; var x333333; x333333 }`. Now it will generate the function only if it actually added at least 1 control name into the code strings.

So let's let it do the checks, and then go.

JordanHendersonMusic:
Do you want to write a unit test for that? Or do you think this is good as is?

jamshark70:
> Do you want to write a unit test for that? Or do you think this is good as is?

Actually upon a bit more testing... yes, I do need to update that.

It turns out to be impossible to get into that condition, but I found a tiny bug for this (incredibly rare) case.

The only case where it could possibly happen is like this (which, let's be real, nobody is going to do this):

```
(
SynthDef(\testempty, { arg gate = 1;
	Out.kr(1000, gate)
}).add;
)
```

SynthDesc by default excludes `gate` from the msgFunc. So in this unique case you have a `controls` collection that is not empty, but whose entries will all be excluded. So the msgFunc should be `#{ Array.new }`.

There is already a check for this, near the top of `makeMsgFunc`, but I wrote `^ #{ Array.new }` instead of `msgFunc = #{ Array.new }; ^this`. So msgFunc ends up being nil.

Rare case, yeah, but let me close the gap, and I think it is worth a unit test, to catch the case where somebody removes that `if(names.size == 0)`.

Later today though.

jamshark70:
OK, I checked in those changes.

PR #7484, opened 2026-05-08
server: do not skip .so files in 'top-level' directories

Spacechild1:
This fixes an issue with system-wide extensions where the plugins are installed in the same location as the core plugins. Since the latter already use the new .scx extensions, any plugins with the old .so extension would be silently skipped.

Initially reported by @jamshark70 on the SC forum: https://scsynth.org/t/3-15-so-extension-has-been-deprecated/13245

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/scsynth/SC_Lib_Cintf.cpp
+++ b/server/scsynth/SC_Lib_Cintf.cpp
@@ -68,9 +68,12 @@ SC_LibCmd* gCmdArray[NUMBER_OF_COMMANDS];
 
 // The generic .so extension has been deprecated in SC 3.15 so developers can ship shared libraries
 // along their plugins. We do the following to provide a smooth upgrade path:
-// If there has been a plugin with the SC_PLUGIN_EXT extension in the same folder or one of its
-// parent folders we simply ignore the .so file because we can assume it is a dependency.
+// If there has been a plugin with the SC_PLUGIN_EXT extension in the same folder or one of its parent
+// folders we simply ignore the .so file because we can assume it is a dependency.
 // Otherwise we load the .so file and post a warning.
+// We have to make an exception for 'top-level' search paths because system-wide extensions install their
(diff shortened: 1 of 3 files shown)


JordanHendersonMusic (approved):
Would be nice if all plugins put their libraries in a directory so you know where they came from later... but since we can't enforce that now, this seems like a reasonable alternative to me! 

I haven't tested this, but assuming some one has, this can be merged.

Spacechild1:
> Would be nice if all plugins put their libraries in a directory so you know where they came from later...

Yeah, throwing all the plugin binaries in a single directory is pretty messy. I think each extension should install into its own subdirectory.

> I haven't tested this, but assuming some one has, this can be merged.

Cool! I will wait for feedback from @jamshark70. I've successfully tested on Linux with scsynth and supernova, but you never know...

jamshark70:
> I will wait for feedback from @jamshark70.

Maybe tomorrow...? Because I have to rebuild for 3.15 (I'm not ready yet to use 3.15 dev as my main install).

dyfer:
@jamshark70 would you be able to test this?

jamshark70:
> @jamshark70 would you be able to test this?

Now I see warnings for sc3-plugins files:

```
WARNING: '/usr/local/lib/SuperCollider/plugins/HOADecLebedev504.so': the .so extension has been deprecated!
WARNING: '/usr/local/lib/SuperCollider/plugins/HOAPanLebedev262.so': the .so extension has been deprecated!
WARNING: '/usr/local/lib/SuperCollider/plugins/HOAMirror8.so': the .so extension has been deprecated!
WARNING: '/usr/local/lib/SuperCollider/plugins/HOAEncLebedev504.so': the .so extension has been deprecated!

... etc.
```

So I think this is fixed.

PR #7485, opened 2026-05-10
Add missing Size.schelp referenced by `Image` and `QPenPrinter` schelp files

prko:
## Purpose and Motivation

Two help files in the SuperCollider documentation reference `link::Classes/Size::`, but the corresponding help file `Size.schelp` does not exist:
- https://github.com/supercollider/supercollider/blob/3747a49c645c39792b367ed4b89ef9386cd6a1f0/HelpSource/Classes/Image.schelp#L86
- https://github.com/supercollider/supercollider/blob/3747a49c645c39792b367ed4b89ef9386cd6a1f0/HelpSource/Classes/QPenPrinter.schelp#L85

This results in a broken help link and an error in the help browser.  
This PR adds the missing `Size.schelp` file to resolve the issue.

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


HelpSource/Classes/Size.schelp
```diff
@@ -0,0 +1,71 @@
...
+The height component. Typically an link::Classes/Integer:: or an link::Classes/Float::, but accepts any built-in subclasses of link::Classes/Number::.
+
+returns:: A new link::Classes/Size:: instance.
+
+
+instancemethods::
+
+METHOD:: width
+
+METHOD:: storeArgs
```
JordanHendersonMusic, line 36: what does this do?

HelpSource/Classes/Size.schelp
```diff
@@ -0,0 +1,71 @@
...
+~aSize1.postln
+
+~aSize2 = Size(100, 200);
+
+~aSize1 == ~aSize2 // -> false
+
+~aSize1.asRect     // -> Rect(0, 0, 50, 100)
+
+~aSize1.asPoint     // -> Point(50, 100)
+::
```
JordanHendersonMusic, line 90: I know sc doesn't follow this rule so far... but would you mind putting semicolons at the end of each line so that the entire code block is valid sc code?

HelpSource/Classes/Size.schelp
prko: `.storeArgs` returns an array with the width and height: `[width, height]`
I do not know the purpose of this....
I kept all methods in the automatically generated SCDoc template.

HelpSource/Classes/Size.schelp
prko: OK, done in https://github.com/supercollider/supercollider/pull/7485/commits/dbdf8fcc247d81de60fc7b3c54d8e43e28cc9c43.
Thanks!

HelpSource/Classes/Size.schelp
JordanHendersonMusic: Could you add that? I think the rest of the methods are self-explanatory

prko: Of course! Done in https://github.com/supercollider/supercollider/pull/7485/commits/259ececef191a9ef63be2e09e2b16a1798e1f843

JordanHendersonMusic (approved):

JordanHendersonMusic:
Thanks!

jamshark70:
BTW -- not quite on topic, but I always wondered. Why do we need both Size() and Point()? (I.e., I think we don't.)

JordanHendersonMusic:
Yeah I agrree, it is also not a size, but more of an extent... can't change it now though!

prko:
Probably yes. Here’s a practical comparison from my side:
```
~size = Size(300, 500);
~w1 = Window("~size", ~size.asRect).front
```
```
~point = Point(300, 500);
~w2 = Window("~point", ~point.asRect).front
```
Both produce the same result.

PR #7489, opened 2026-05-13
fix sclang unity builds in msvc

dyfer:
## Purpose and Motivation

This PR fixes unity build for sclang under MSVC. 
This was achieved by excluding files from the unity build. The need to exclude multiple files comes from naming/macro clashes with Windows headers, AFAIU

Additionally, I've turned on unity build for a single non-release build on each platform.

I plan to fix unity builds for other components, and once that is done, we can decide if we want to use unity builds by default or not.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/lang/CMakeLists.txt
+++ b/lang/CMakeLists.txt
@@ -189,6 +189,21 @@ add_library(libsclang STATIC ${sclang_sources} ${sclang_parser_source})
 
 if(FINAL_BUILD)
 	set_source_files_properties(${sclang_parser_source} PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+    set_source_files_properties(${CMAKE_SOURCE_DIR}/SCDoc/lex.scdoc.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+    if(WIN32)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/common/SC_Filesystem_win.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/common/SC_PaUtils.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/common/SC_Win32Utils.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/lang/LangPrimSource/PyrFilePrim.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/lang/LangPrimSource/PyrPlatformPrim.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/lang/LangPrimSource/PyrSched.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/lang/LangPrimSource/SC_AudioDevicePrim.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/lang/LangPrimSource/SC_ComPort.cpp PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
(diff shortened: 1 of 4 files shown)


lang/CMakeLists.txt
```diff
@@ -189,6 +189,21 @@ add_library(libsclang STATIC ${sclang_sources} ${sclang_parser_source})
 
 if(FINAL_BUILD)
 	set_source_files_properties(${sclang_parser_source} PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
```
JordanHendersonMusic, line 191: Why is all this necessary exactly?

dyfer: Because in the SC codebase we use macros and definitions that are also used in Windows headers (e.g. `IN`, `OUT`, `TokenType`, `YYSTYPE` - some of these are from Bison). If we compile files with those defitions in one batch with files including Windows headers (whether directly or indirectly, e.g. through boost/asio), the compilation fails. 

I decided to simply exclude the files that end up eventually including windows headers from the batch, but there may be other/better ways to deal with this. 
For macros, [this is one way](https://github.com/supercollider/supercollider/blob/develop/common/SC_Win32Utils.h#L25-L45) to deal with the issue, but we'd need such guards (or `#pragma push_macro`/`#pragma pop_macro`) around all the direct and indirect windows includes, which may be impractical. Also I don't know how to deal with the enum conflicts (like TokenType).

EDIT: outside of Windows, it seems we have some conflicting definitions inside our codebase, e.g. `YYLTYPE`, that's why some files are excluded from unity builds on all platforms.

JordanHendersonMusic: Thanks for explaining this!

JordanHendersonMusic:
So I have a concern about unity builds based on your explanation.

Right now, I find it a common occurrence to have different compilers, build systems and os behave differently. This can be due to GCC and clang not agreeing, or more commonly msvc strictly adhering to the standard, sometimes it's due to different oss and the system libraries... Many reasons.

I am a bit concerned that this will add yet another way that the code will work on my machine (doing incremental compilation) and not on the ci, as debugging based on GitHub's output is not a pleasant experience!

dyfer:
These are valid concerns! Yet another configuration that may behave differently is definitely something to be weary of.

FWIW I don't suggest that we move _all_ CI to a unity build... but if it is significantly faster, we could move _most_ of CI, and leave one non-unity build for testing. I think it is worth discussing before we move this further.

@jrsurge do you have any suggestions here? Adding another build configuration that may behave slightly differently is a concern and we don't want to create more maintenance burden.

JordanHendersonMusic:
>  I think it is worth discussing before we move this further.

It sounds like you want to run this as an experiment? It that case, since reverting something in the build system is so easy, I say just do it and see what happens!

No need to discuss something we don't have enough information about yet!


Fyi, the enum conflict is due to a namespace issue, if you add `using TokenType = SC::lex::TokenType` the issue disappears.

The problems with the yy* vars seems real though, the bison generator makes a tone of globals, we have two grammars, sclang and the docs, I *believe* this necessitates odr violations??

JordanHendersonMusic (approved):

dyfer:
Thanks!

> Fyi, the enum conflict is due to a namespace issue, if you add `using TokenType = SC::lex::TokenType` the issue disappears.

Hm, are you sure? The issue is that _our_ definition leaks into Windows headers, IIUC, e.g.:

```
2026-05-12T18:24:03.5811014Z C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\um\winnt.h(12059,5): error C2365: 'TokenType': redefinition; previous definition was 'typedef' [D:\a\supercollider\supercollider\build\lang\libsclang.vcxproj]
2026-05-12T18:24:03.5812750Z (compiling source file '../../lang/CMakeFiles/libsclang.dir/Unity/unity_1_cxx.cxx')
```

As for the macro conflicts, do you think they should be solved in the code (pop/push macro) or is it fine to just exclude these files from the unity build like I'm proposing right now?

JordanHendersonMusic:
> The issue is that our definition leaks into Windows headers, IIUC, e.g.:

Ah yes I'm still not 100% sure how they work. Assuming it's the TokenType from the new lexer, then it's in a namespace so can't collide (unless they also have a `sc::lex` namespace). So if all the files are being stitched together, then we might just have to remove a `using` declaration somewhere.


I've only ever used undef with macros, but it's with a go? 

What I'm concerned about is constantly adding to this list. We should try to get everything working correctly first... but if you just wanna see what happens and if it affects development then I'm also fine with merging now!

dyfer:
> We should try to get everything working correctly first...

Do you mean - unity build working correctly by excluding files? Or fixing files so that they we don't populate conflicting names/macros into windows headers?

JordanHendersonMusic:
Fixing the files

PR #7493, opened 2026-05-15
Add vcpkg manifest mode

dyfer:
## Purpose and Motivation

Background: we started using vcpkg in CI as an alternative source of packages first on Windows to get the readline library. Then we also started to use vcpkg for macOS libraries in order to deploy SC for older systems, as well as to provide universal arm64/x86_64 binaries (the latter is done by manually assembling "universal" libraries). All this was done in vcpkg "classic" mode, i.e. by installing vcpkg packages system-wide and then point cmake to use them. A more modern way to use vcpkg is in "manifest mode", where packages needed by a project are specified in the manifest file and are automatically installed during the cmake configure step.

This PR adds manifest file so that vcpkg can be used in manifest mode, i.e. automatically install dependencies (libsndfile, fftw, readline). By default, these will be used from vcpkg only on Windows and only when vcpkg is detected, but can be enabled on other platforms as well (we use it in CI on macOS).

Note: the main Windows build changes the source for fftw and libsndfile - they are now installed with vcpkg, as opposed to the manual install. I think that since we need to use vcpkg to get readline, we should switch to treating vcpkg as the default dependencies source on Windows. However, we still keep the ability to use the manually installed fftw and libsndfile. I'm certainly open to discussing this change, I don't want to force the use of vcpkg.

Other adjustments to the CI pipeline include disabling manifest mode for the macOS universal build.

~~EDIT: I also added portaudio from vcpkg, as suggested by @sadguitarius, so that users don't need to install asio headers manually (these are fetched as a vcpkg dependency for portaudio). We can easily revert to using portaudio from our source, but for now using the one from vcpkg seems the easiest, if we already use vcpkg for other dependencies.~~

~~EDIT 2: I realized that using portaudio from vcpkg causes legacy windows audio APIs (MME, DirectSound, WDKMS) to be exposed in scsynth again. In that case, I set the CI build to still use the built-in portaudio... but I think that for the local builds using portaudio from vcpkg is still easier because it avoids the need to manually download and install asio sdk.~~

EDIT 3: After merging #7651, I left the portaudio from our sources in, since we will now automatically download ASIO SDK and it maintains our configuration of audio APIs.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation (~~note: the vcpkg change is most notable for windows, but windows readme needs a complete rewrite. That will be a separate PR~~ Windows readme update submitted as #7641 )


capital-G (approved):
Please merge @dyfer

dyfer:
This should also be accompanied by adding a statically-linked fftw to sc3-plugins, but we should wait until https://github.com/supercollider/sc3-plugins/pull/435 is resolved, which may prompt the need to release sc3-plugins 3.14.1, which should use the current fftw implementation in sc3-plugins.

EDIT: submitted https://github.com/supercollider/sc3-plugins/pull/446

dyfer:
@capital-G I added portaudio from vcpkg since your last review.
Otherwise there seems to be a consensus that it's fine to go ahead with vcpkg manifest mode, IIUC.

dyfer:
I just realized that using portaudio from vcpkg causes legacy windows audio APIs (MME, DirectSound, WDKMS) to be exposed again (see #6900). In that case, I set the CI build to still use the built-in portaudio... but I think that for the local builds using portaudio from vcpkg is still easier. 

I'm open to suggestions :)

dyfer:
I've kept portaudio from our sources after all (see the updated description).

Ready for review :)

vcpkg.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "supercollider",
+  "version": "0.0.1",
+  "features": {
+    "fftw3": {
+      "description": "Fast Fourier Transform library",
+      "dependencies": [
+        "fftw3"
```
capital-G, line 8: shouldn't we pinpoint the versions, or at least add a restriction that no major version jumps are accepted? Do you think we should  add dependabot for this so we get notified on new major versions? See https://learn.microsoft.com/en-us/vcpkg/concepts/dependabot

.github/workflows/build_macos.yml
```diff
@@ -37,13 +37,10 @@ jobs:
             qt-modules: 'qtwebengine'
             deployment-target: "10.15"
             cmake-architectures: x86_64
-            homebrew-packages: "readline portaudio"
-            vcpkg-packages: "libsndfile"
```
capital-G, line 41: so we are switching from homebrew to vcpkg for x86 macos builds - is there an advantage to this or simply preparing for cross compiling when x64 runners are gone?

cmake_modules/FindPortaudio.cmake
```diff
@@ -1,6 +1,19 @@
...
+elseif(TARGET portaudio)
+	add_library(PortAudio::PortAudio ALIAS portaudio)
+elseif(TARGET portaudio::portaudio)
+	add_library(PortAudio::PortAudio ALIAS portaudio::portaudio)
+elseif(TARGET unofficial::portaudio::portaudio)
+	add_library(PortAudio::PortAudio ALIAS unofficial::portaudio::portaudio)
+else()
+	# this should only be run on systems with pkg-config installed
+	find_package(PkgConfig REQUIRED)
+	pkg_check_modules(PORTAUDIO_SYSTEM_LIB REQUIRED IMPORTED_TARGET portaudio-2.0)
```
capital-G, line 15: we pinpoint a version here but don't do so in the manifest or what is the 2.0 referring to here?

capital-G (commented):
some questions, but otherwise this is a nice addition - especially like the clarity of the cmake code :)

.github/workflows/build_macos.yml
dyfer: The only switch from homebrew to vcpkg is for readline here (libsndfile was from vcpkg previously, but now we use the manifest mode instead of running `vcpkg install libsndfile`).
Vcpkg allows explicitly controlling macOS deployment target. Readline from homebrew seemed to have worked fine for the legacy build, but IIUC this was only by accident - readline didn't seem to touch any of the newer APIs. By using vcpkg version, we explicitly control which macOS version it is built for.

cmake_modules/FindPortaudio.cmake
dyfer: portaudio-2.0 is what was there before (this part is as it was before, only now it is inside an else clause so it shows up as a new addition.

It seems that 2.0 refers to the API version and it relates to "V19" version of Portaudio: https://github.com/PortAudio/portaudio/wiki/V19ReleasePlan

vcpkg.json
dyfer: I haven't thought of library versions... I think it will be years before any version jump happens, but if you think it's reasonable, I can take a look into pinpointing versions...

This would only matter if the API would change.

capital-G: Maybe it is not that necessary - we don't (i.e. can't) do this on homebrew, so it is probably also fine to roll along w/ vcpkg. Once it creates problems, we can add a barrier.

capital-G (approved):

capital-G:
Feel free to merge @dyfer :)

Issue #7207, opened 2025-10-26
Improve searchability of `kwargs` in the documentation


prko:
In SuperCollider 3.14, support for keyword arguments (kwargs) was introduced. This change is mentioned in the 3.14 News section, which also references the corresponding GitHub issue and pull request. However, the current help system does not return relevant results when searching for “kwargs”.

As a result, users who are not already aware of the feature may find it difficult to discover through the documentation alone. To improve usability, it would be helpful if:

- The term kwargs appeared in the main help files where keyword arguments are explained.
- Cross-references were added from the News entry to the relevant help pages.
- Search indexing ensured that queries for “kwargs” lead to the appropriate documentation.

This would make the feature more visible and accessible, particularly for new users who rely on the help browser’s search function.

**While this is partly a limitation of the current help system’s core functionality, it also highlights the importance of writing help documentation in a way that ensures new features can be discovered using existing search capabilities.**

PR #7495, opened 2026-05-16
Standardise `key–value pair` usage and `argument terminology` across documentation

### LINKED ISSUE #7491: Document the key–value pair syntax within Arrays/Lists and clarify its distinction from Event literals
## Motivation
After reading the discussion here:
https://scsynth.org/t/array-vs-event-sintax-for-argument-passing-in-synth/13222?u=prko

I realised that the syntax for key–value pairs within Arrays or Lists is not clearly documented.
This syntax is easily confused with Event literals, and the actual behaviour is not obvious, particularly for users who might expect [\a: 3] to behave like (a: 3).

The following examples illustrate the current behaviour:
```
a = [\a, 3]        // -> [a, 3]
b = [\a: 3]        // -> [a, 3]
c = [\a: 3, \b]    // -> [a, 3, b]
d = [\a: 3: \b]    // ERROR: syntax error, unexpected ':', expecting ']'
```
These examples show that:

- `[\a: 3]` is not an Event literal but an Array using a special key–value pair syntax.
- The rules governing this syntax (what is valid and what is not) seem not to be fully documented.
- This can lead to confusion, especially when passing arguments to Synths.

Adding the explanation on the Array or List with key-value pairs, including the example above.

## Plan for Implementation
I am willing to contribute the documentation update, but I am unsure where the most appropriate location would be.
Currently, related references appear as follows:
```
7 results - 5 files

HelpSource/Classes/CallOnce.schelp:
  44  argument:: kwargs
  45: Array of key-value pairs for keyword arguments. See link::Classes/Function#-value::.
  46  

HelpSource/Classes/Function.schelp:
  174  argument::pairs
  175: Array or List with key-value pairs.
  176  

HelpSource/Classes/GridLayout.schelp:
  48  
  49:     The simplified syntax for placing key-value pairs into an array comes handy (see link::Reference/Syntax-Shortcuts#Creating Arrays with key-value pairs::, and the example below).
  50  

HelpSource/Classes/LineLayout.schelp:
  33  
  34: You can assign a strong::stretch factor:: and/or strong::alignment:: to an item by wrapping it into an array, followed by pairs of ('stretch', factor) and/or ('align', alignment). 'stretch' and 'align' may be abbreviated with 's' and 'a'. Simplified syntax for placing key-value pairs into an array comes handy (see link::Reference/Syntax-Shortcuts#Creating arrays with key-value pairs::, and the example below). For possible alignment values see link::Reference/gui_alignments::.
  35  

HelpSource/Classes/Object.schelp:
  570  argument:: pairs
  571: Array or List with key-value pairs.
  572  discussion::
```
Guidance on the best place to add this information would be appreciated.

prko:
closes: #7491, #7207

---

**Note:** This PR is now larger than initially intended.  
It began with clarifying key–value pair usage, but it now also standardises
argument terminology throughout the documentation for consistency.
See the explanation in https://github.com/supercollider/supercollider/pull/7495/commits/d4a8ff716065274e7d61e1cfb6feb84699eb3b40. 

----

## Purpose and Motivation

Improves the documentation of key–value pair syntax for several subclasses of link::Classes/Collection::.
The explanation is written with beginners in mind, inspired by a recent question on the scsynth forum.
See #7491.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Reference/Functions.schelp
+++ b/HelpSource/Reference/Functions.schelp
@@ -75,7 +75,7 @@ note::
 The use of three arguments is often paired with link::Classes/Object#-performArgs::.
 ::
 
-An argument list immediately follows the open curly bracket of a function definition. An argument list either begins with the reserved word code::arg::, or is contained between two vertical bars. If a function takes no arguments, then the argument list may be omitted.
+An argument list immediately follows the open curly bracket of a function definition. An argument list either begins with the reserved word code::arg::, or is contained between two vertical bars. If a function takes no arguments, then the argument list must be omitted.
 
 While you can pick any name for the emphasis::variadic positional arguments:: and emphasis::arbitrary keyword arguments::, they always refer to the unmatched emphasis::variadic positional arguments:: and emphasis::arbitrary keyword arguments::, so it is often best to name them code::args:: and code::kwargs::.
 
--- a/HelpSource/Reference/Functions.schelp
+++ b/HelpSource/Reference/Functions.schelp
@@ -52,25 +52,61 @@ image::functions.png#Functions::
 
 section:: Arguments
 
+Functions may take several kinds of arguments.
+SuperCollider language supports three categories of function arguments in the following order of an argument list:
+
+numberedlist::
+## Declared Arguments
+
+fixed, named arguments that may have default values.
+
+## Positional Variable Arguments
+
+a catch‑all for additional positional arguments.
+
+## Arbitrary Keyword Arguments
+
+a catch‑all for additional keyword arguments.
+::
+
+note::
+The use of three arguments is often paired with link::Classes/Object#-performArgs::.
+::
+
 An argument list immediately follows the open curly bracket of a function definition. An argument list either begins with the reserved word code::arg::, or is contained between two vertical bars. If a function takes no arguments, then the argument list may be omitted.
 
-Names of arguments in the list may be initialized to a default value using the following syntax forms. Arguments which are not explicitly initialized will be set to nil if no value is passed for them.
+While you can pick any name for the positional variable and arbitrary keyword arguments, they always refer to the unmatched positional and keyword arguments, so it is often best to name them code::args:: and code::kwargs::.
 
-"arg" style, default value is a literal:
+
+Names of arguments in the list may be initialized to a default value.
+Arguments which are not explicitly initialized will be set to code::nil:: if no value is passed for them.
+The following subsections describe each argument type in detail.
+
+
+subsection:: Declared Arguments
+
+Declared arguments are written first in the argument list.
+Names of arguments in the list may be initialized to a default value using the following syntax forms (literals or expressions):
+
+list::
+## "arg" style, default value is a literal:
 code::{ arg x = 1; .... } :: link::#[1]::
 
-"arg" style, default value is an expression:
+## "arg" style, default value is an expression:
 code::{ arg x = 10.rand; ... } :: link::#[2]::
 
-"arg" style, default value is a literal but you want to treat it like an expression:
+## "arg" style, default value is a literal but you want to treat it like an expression:
 code::{ arg x = (2); ... } :: link::#[2]::
 
-Pipe style, default value is a literal:
+## Pipe style, default value is a literal:
 code::{ |x = 1| ... } :: link::#[1]::
 
-Pipe style, default value is an expression:
+## Pipe style, default value is an expression:
 code::{ |x = (10.rand)| ... } :: link::#[2]::
+::
 
+Declared arguments behave differently depending on whether the default is a literal or an expression.
+Literal defaults are stored in the link::Classes/FunctionDef::, while expression defaults are evaluated at call time only when the passed value is code::nil::.
 
 In general arguments may be initialized to literals or expressions, but in the case of link::Classes/Function#-play:: or link::Classes/SynthDef#-play::, they may only be initialized to literals.
 code::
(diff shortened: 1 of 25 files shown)


HelpSource/Reference/Key-Value-Pairs.schelp
```diff
@@ -8,9 +8,24 @@ SECTION::Motivation
...
+x = { |a = 0, b = 0| a + b };
+
+// positional argument:
+x.(1, 2)
+
+// keyword argument (key-value pair):
+x.(b: 1)
+x.(b: 1, a: 2)
+::
+It is unrelated to the collecton-based key–value pair syntax described in this document.
```
JordanHendersonMusic: Not if you use kwargs.

prko:
```
This syntax uses keyword arguments (kwargs) for method calls. While it looks similar, it is a different mechanism from the collection‑based key–value pair syntax discussed in this document.
```
How about this then?

JordanHendersonMusic: It is the same mechanism though 
```
{ | ...a, k| k }.(a:2, b:2) == [\a, 2, \b, 2]
```

prko: Sorry! Several weeks ago, I reverted my SC3.15.0-dev to an older version which is compatible with the current official SC3plugin. So all the documentation I committed this spring was deleted in my local copy of SC. 
Now I updated each documentation...

Also, this part is changed. Please see https://github.com/supercollider/supercollider/pull/7495/commits/3601cea7cad2550fdd6c1fb233c4047068012d61

prko:
@JordanHendersonMusic 

I have reviewed most of the help files related to key–value pairs and keyword arguments.
The expression keyword arguments is highly context‑dependent and, in some cases, ambiguous.

Up to SC 3.13, this term appears to have referred exclusively to the keyword‑argument syntax used in function calls.
With SC 3.14, however, arbitrary keyword arguments are now permitted after variable arguments.
Because of this change, it has become necessary to distinguish clearly between:

- keyword arguments used in function‑call syntax, and

- keyword arguments belonging to the third group of function parameters.

This PR aims to resolve the resulting ambiguity and standardise the terminology across the documentation.

---


Unify terminology for keyword arguments and function argument categories across documentation

The term “keyword arguments” has become ambiguous in the documentation due to the recent introduction of “arbitrary keyword arguments” (|... args, kwargs|) in Function definitions.

Historically, The SuperCollider Book and earlier documentation used “keyword arguments” exclusively to mean keyword argument calls (message‑passing syntax), without ambiguity.

This commit standardises terminology across all relevant documents to remove confusion and ensure consistent usage.

- The overloaded term “keyword arguments” is now split into:
  - “keyword argument calls” (message‑passing syntax: f.(x: 1, y: 2))
  - “arbitrary keyword arguments” (Function definition: |... args, kwargs|)

- Likewise:
  - “positional arguments” has been clarified as “positional argument calls”

- The terminology for function arguments has been standardised as follows:
  - declared arguments
  - variable arguments
    - variadic positional arguments
    - arbitrary keyword arguments

All occurrences found in the documentation have been updated for consistency.

If the community prefers alternative terminology, the following substitutions remain entirely possible:
  - “keyword argument calls” → “keyword arguments”
  - “positional argument calls” → “positional arguments”
  - “declared arguments”, “variadic positional arguments”, and “arbitrary keyword arguments” may be replaced with any agreed‑upon terms.

This commit provides a consistent baseline from which such future adjustments may be made.

HelpSource/Reference/Functions.schelp
```diff
@@ -52,94 +52,84 @@ image::functions.png#Functions::
...
+a catch‑all for additional keyword arguments.
+::
 
-{ arg a = Array.geom(4, 100, 3); a * 4 }.value;
+note::
+The use of three arguments is often paired with link::Classes/Object#-performArgs::.
+::
 
-// this is not:
+An argument list immediately follows the open curly bracket of a function definition. An argument list either begins with the reserved word code::arg::, or is contained between two vertical bars. If a function takes no arguments, then the argument list may be omitted.
```
JordanHendersonMusic: The argument list **must** be omitted in the case of `{ | | }`.

JordanHendersonMusic:
> “keyword argument calls”

This term seems a little off to me. Perhaps just 'keyword arguments'?

Then at the call sight we have positional arguments and keyword arguments.

> “arbitrary keyword arguments”

I think this one should be 'variadic keyword arguments'.

HelpSource/Reference/Functions.schelp
prko: Yes, thanks! Done in https://github.com/supercollider/supercollider/pull/7495/commits/ea845bab8def2dfb1df00c7de2fd676f57f72c95

prko:
>>“keyword argument calls”
>
>This term seems a little off to me. Perhaps just 'keyword arguments'?
>Then at the call sight we have positional arguments and keyword arguments.

OK, I can revert it, and it is better not to change the term has used for a long time.

---

>>“arbitrary keyword arguments”
>
>I think this one should be 'variadic keyword arguments'.

The term “arbitrary keyword arguments” is introduced by @telephon in the news in SC3.14:
- https://github.com/supercollider/supercollider/blob/35546310964134d4a0f79217ab879a8669dfd284/HelpSource/Guides/News-3_14.schelp#L10
-  https://github.com/supercollider/supercollider/blob/35546310964134d4a0f79217ab879a8669dfd284/HelpSource/Guides/News-3_14.schelp#L48

If @telephon does not object, in this PR’s last status:

- revert:
   - currently: “keyword argument calls” → will be: “keyword arguments”
   - currently: “positional argument calls” → will be: “positional arguments”

- change:
   - currently: “arbitrary keyword arguments” → will be: “variadic keyword arguments”

- keep:
   - “variadic positional arguments”

I will not touch the [supercollider/HelpSource/Guides/News-3_14.schelp](https://github.com/supercollider/supercollider/blob/35546310964134d4a0f79217ab879a8669dfd284/HelpSource/Guides/News-3_14.schelp).

prko:
Otherwise, in the final status of this PR, we could:
- revert:
   - currently: “keyword argument calls” → will be: “keyword arguments”
   - currently: “positional argument calls” → will be: “positional arguments”
- change:
   - currently: “variadic positional arguments” → will be: “ arbitrary positional arguments”
- keep:
   - “arbitrary keyword arguments”

prko:
The term `arbitrary keyword arguments` used in the SC 3.14 release notes is very clear and easy to understand. However, using only `keyword arguments` in the documentation can be confusing. In most programming contexts, `keyword arguments` are understood specifically in contrast to `positional arguments` in function calls, so readers may misinterpret the term.

For consistency across SC versions and to avoid ambiguity, I think continuing to use `arbitrary keyword arguments` would be clearer.

HotwheelsSisyphus:
I agree (with using the term "arbitrary kwargs" for SC), but it's also helpful to include the reference to the wider technical term (variadic) somewhere (just once in the glossary and maybe in Function.schelp, **not** everywhere)

E.g., add "arbitrary kwargs" to glossary:
```
Arbitrary keyword arguments: Keyword arguments supplied via { | ... args, kwargs | }, where the number of keyword arguments is indefinite (variadic).
```

prko:
I think you mean `Functions.schelp`? This PR already has explanations there.

In my opinion, `Glossary.schelp` should include more technically oriented and SC‑specific terms.

```SCDoc
## Variadic positional arguments || Positional arguments supplied via code::{ | ... args | }::, where the number of positional arguments is indefinite. See link::Reference/Functions#Variadic positional arguments::.

## Arbitrary keyword arguments || Keyword arguments supplied via code::{ | ... args, kwargs | }::, where the number of keyword arguments is indefinite (variadic). See link::Reference/Functions#Arbitrary Keyword Arguments::.
```

Let’s see what other developers say.

PR #7503, opened 2026-05-20
build: fix precommit for bison

JordanHendersonMusic:
Sometimes the pre commit will give an error about incorrect use of a regex, this fixes that by using the correct glob.

To format all files you now run `FULL_CHECK=1 pre-commit run --hook-stage manual --all-files`

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,7 @@ repos:
         name: Run Bison parser checking for any errors
         # NOTE: this command is a copy from tools/generate_parser_bison.sh
         entry: bison -Werror --defines=lang/LangSource/Bison/lang11d_tab.h --output=lang/LangSource/Bison/lang11d_tab.cpp lang/LangSource/Bison/lang11d
-        files: lang/LangSource/Bison/*
+        files: lang/LangSource/Bison/.+
         pass_filenames: false
         language: system
+        stages: [pre-commit, pre-merge-commit, pre-push]


JordanHendersonMusic:
Thought I tested this thoroughly, obviously not...

dyfer:
So how do we check whether this is correct?

JordanHendersonMusic:
Oh I've ran this a bunch on my other pr, I think this is correct now! I thought it was regex, not a glob, which I didn't even know was a thing...

To check it on a local machine just edit any file in the bison dir.

dyfer:
To test this, I tried changing contents of `lang/LangSource/Bison/lang11d_tab.cpp`.
However, I got the same result (expected behavior) on `develop` as well as in this PR:

```sh
Run Bison parser checking for any errors.................................Failed
- hook id: bison
- files were modified by this hook
```

What is the situation where the current implementation fails and this PR fixes?

-----------

Also, macOS comes with an old version of bison preinstalled. When I first tried committing a file in the bison directory, I got:

```
Run Bison parser checking for any errors.................................Failed
- hook id: bison
- exit code: 1

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/bison: invalid option -- W
Try `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/bison --help' for more information.
```

The solution was the install bison from `homebrew` and then specifically add it to the `PATH` (it's not added by default, as to not replace the system-installed one). But maybe there's a way to make the pre-commit to work with the old version as well? Idk, I know nothing about bison...

<details>

```
==> Pouring bison--3.8.2.arm64_sonoma.bottle.tar.gz
==> Caveats
bison is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have bison first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/bison/bin:$PATH"' >> ~/.zshrc

For compilers to find bison you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/bison/lib"
```
</details>

I think the bison part needs to be documented on the [wiki section of pre-commit](https://github.com/supercollider/supercollider/wiki/Pre%E2%80%90commit). I also wonder if that documentation could be easier to find... I usually search for it in the GH search bar.

JordanHendersonMusic:
> What is the situation where the current implementation fails and this PR fixes?

Ah! Yes I didn't actually explain what is wrong, Dennis messaged me about this so I just assummed.

Sometimes the precommit will give an error telling you not to use a regex in the file attribute, that is what this PR fixes.

---

> The solution was the install bison from homebrew

This seems correct, you can't edit the files manually, they need to be generated by bison from the grammar. I believe we specify a minimum bison version somewhere? What version is preinstalled?

> But maybe there's a way to make the pre-commit to work with the old version as well? 

I don't think so, the options here to generate the file and turn them warnings into errors, without that, the precommit step is not all the useful. The aim is to stop bad grammar changes being made.

> I think the bison part needs to be documented on the

You don't need bison to run precommit, you only need bison if you edit the files, and since they are generated by bison (except from the grammar) you need it anyway. I was kinda assuming the user would already have bison in their path.

dyfer:
> Sometimes the precommit will give an error telling you not to use a regex in the file attribute, that is what this PR fixes.

Got it. Could you please update the first comment? May be useful when searching the archive in the future.

Since I can't reproduce the issue, I'll let Dennis approve this.

> > The solution was the install bison from homebrew
> 
> This seems correct, you can't edit the files manually, they need to be generated by bison from the grammar. I believe we specify a minimum bison version somewhere? What version is preinstalled?

On my macOS 14 bison that comes with the system (or Xcode? not sure) is version 2.3.

```sh
~ bison -V
bison (GNU Bison) 2.3
```

> > I think the bison part needs to be documented on the
> 
> You don't need bison to run precommit, you only need bison if you edit the files, and since they are generated by bison (except from the grammar) you need it anyway. I was kinda assuming the user would already have bison in their path.

The issue is that Bison use is not documented _at all_. There's no mention of it in the project docs, helpfiles, or the wiki. It's only mentioned in the code itself. I am only guessing what it is for, based on your work, looking at the code (occasionally), discussion in the PRs and at the dev meetings.

JordanHendersonMusic:
> On my macOS 14 bison that comes with the system (or Xcode? not sure) is version 2.3.

That was released in 2006. I actually don't know what version we are running in the ci... I'll check that. It shouldn't matter for language stability what version we use, but if it generates even slightly different source code the ci will fail. I don't think this is that much of an issue right now because bison is so old and very stable.

> The issue is that Bison use is not documented at all. 

Oh! I didn't know that! 

It takes the grammar file and turns it into the function yyparse using a big jump table.

dyfer:
> That was released in 2006.

Yeah, macOS comes with some old libraries by default (readline is another example).

> I actually don't know what version we are running in the ci...

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#installed-apt-packages

On Ubuntu 24.04 it seems to be 3.8. Not sure if it's installed by default on the macOS or Windows runners though.

JordanHendersonMusic:
@dyfer @capital-G 

I've made pre-commit work with the formatter now, you run `FULL_CHECK=1 pre-commit run --hook-stage manual --all-files`

If we need to do this for anything else, you just copy the `stages` written in the bison rule.

This is quite hacky, but seems to work well.

.pre-commit-config.yaml
```diff
@@ -35,6 +35,7 @@ repos:
         name: Run Bison parser checking for any errors
         # NOTE: this command is a copy from tools/generate_parser_bison.sh
         entry: bison -Werror --defines=lang/LangSource/Bison/lang11d_tab.h --output=lang/LangSource/Bison/lang11d_tab.cpp lang/LangSource/Bison/lang11d
-        files: lang/LangSource/Bison/*
+        files: lang/LangSource/Bison/.+
         pass_filenames: false
         language: system
+        stages: [pre-commit, pre-merge-commit, pre-push]
```
capital-G, line 41: Why not use simply `stages: [manual]`?

As said, I think since we can't control the version of bison I don't think it is good to have this part of pre-commit, see https://github.com/supercollider/supercollider/issues/7526#issuecomment-4586487531

PR #7504, opened 2026-05-20
sclexer: enable all warnings and make them all errors

JordanHendersonMusic:
## Purpose and Motivation

Enables all warnings and makes them all errors. About as safe as C++ can get.

### DIFF
--- a/cmake_modules/CompilerWarnings.cmake
+++ b/cmake_modules/CompilerWarnings.cmake
@@ -3,7 +3,7 @@
 #
 # https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
 
-function(set_project_warnings project_name)
+function(set_target_warnings target_name)
   set(MSVC_WARNINGS
       /W4 # Baseline reasonable warnings
       /w14242 # 'identifier': conversion from 'type1' to 'type1', possible loss of data
--- a/cmake_modules/CompilerWarnings.cmake
+++ b/cmake_modules/CompilerWarnings.cmake
@@ -1,0 +1,77 @@
+# This file is copied from here! https://github.com/lefticus/cpp_weekly/blob/master/cmake/CompilerWarnings.cmake
+# from here:
+#
+# https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
+
+function(set_project_warnings project_name)
+  set(MSVC_WARNINGS
+      /W4 # Baseline reasonable warnings
+      /w14242 # 'identifier': conversion from 'type1' to 'type1', possible loss of data
+      /w14254 # 'operator': conversion from 'type1:field_bits' to 'type2:field_bits', possible loss of data
+      /w14263 # 'function': member function does not override any base class virtual member function
+      /w14265 # 'classname': class has virtual functions, but destructor is not virtual instances of this class may not
+              # be destructed correctly
+      /w14287 # 'operator': unsigned/negative constant mismatch
+      /we4289 # nonstandard extension used: 'variable': loop control variable declared in the for-loop is used outside
+              # the for-loop scope
+      /w14296 # 'operator': expression is always 'boolean_value'
+      /w14311 # 'variable': pointer truncation from 'type1' to 'type2'
+      /w14545 # expression before comma evaluates to a function which is missing an argument list
+      /w14546 # function call before comma missing argument list
+      /w14547 # 'operator': operator before comma has no effect; expected operator with side-effect
+      /w14549 # 'operator': operator before comma has no effect; did you intend 'operator'?
+      /w14555 # expression has no effect; expected expression with side- effect
+      /w14619 # pragma warning: there is no warning number 'number'
+      /w14640 # Enable warning on thread un-safe static member initialization
+      /w14826 # Conversion from 'type1' to 'type_2' is sign-extended. This may cause unexpected runtime behavior.
+      /w14905 # wide string literal cast to 'LPSTR'
+      /w14906 # string literal cast to 'LPWSTR'
+      /w14928 # illegal copy-initialization; more than one user-defined conversion has been implicitly applied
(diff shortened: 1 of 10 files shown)


JordanHendersonMusic:
@dyfer What do you think about the CMake changes here? I think this is a good way to go about it?

dyfer:
> @dyfer What do you think about the CMake changes here? I think this is a good way to go about it?

I don't know cmake _that_ well :) It looks good to me, but I have no reference for exporting compile settings into a cmake file and a function.

In msvc this PR just turns `4289` into an error, the rest are warnings, is that what you intended?

EDIT: all warnings are indeed turned into errors using a separate flag.

JordanHendersonMusic:
> is that what you intend?

Absolutely not. Actually I might have to disable msvc because it seems to be giving me a false positive that I cannot get rid off...

Edit: realised this sounded harsh, I meant I haven't intended to do that, but might have!

cmake_modules/CompilerWarnings.cmake
```diff
@@ -0,0 +1,77 @@
...
+      -Wpedantic # warn if non-standard C++ is used
+      -Wconversion # warn on type conversions that may lose data
+      -Wsign-conversion # warn on sign conversions
+      -Wnull-dereference # warn if a null dereference is detected
+      -Wdouble-promotion # warn if float is implicit promoted to double
+      -Wformat=2 # warn on security issues around functions that format output (ie printf)
+  )
+
+  set(CLANG_WARNINGS ${CLANG_WARNINGS} -Werror)
+  set(MSVC_WARNINGS ${MSVC_WARNINGS} /WX)
```
JordanHendersonMusic, line 52: @dyfer I think this one makes all warnings error

dyfer: Ah got it, makes sense. Sorry I missed it!

cmake_modules/CompilerWarnings.cmake
```diff
@@ -0,0 +1,77 @@
+# This file is copied from here! https://github.com/lefticus/cpp_weekly/blob/master/cmake/CompilerWarnings.cmake
+# from here:
+#
+# https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
+
+function(set_project_warnings project_name)
```
dyfer: Shouldn't this function and argument be called target instead of a project? it does call `target_compile_options` in the end.
```suggestion
function(set_target_warnings target_name)
```

JordanHendersonMusic: Yes it should! I just copied this from the guy who does c++ weekly.

langutils/sc_lexer/CMakeLists.txt
```diff
@@ -8,7 +10,9 @@ add_library(sc_lexer STATIC
 
 target_include_directories(sc_lexer PUBLIC include)
 
+
 target_link_libraries(sc_lexer PRIVATE utf8proc)
+set_project_warnings(sc_lexer)
```
JordanHendersonMusic:
```suggestion
set_target_warnings(sc_lexer)
```

langutils/sc_lexer/standalone/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
+include(${CMAKE_SOURCE_DIR}/cmake_modules/CompilerWarnings.cmake)
+
 add_executable(sc_lexer_standalone main.cpp)
 target_link_libraries(sc_lexer_standalone PRIVATE sc_lexer)
-
+set_project_warnings(sc_lexer_standalone)
```
JordanHendersonMusic:
```suggestion
set_target_warnings(sc_lexer_standalone)
```

cmake_modules/CompilerWarnings.cmake
dyfer: You'd also need to change the variable name at the bottom of the file...

dyfer: I'd also consider changing `PROJECT_WARNINGS` to something else, but since this is local it's not as important.

dyfer:
Sorry for tossing it around...
This is good (once the variable name gets fixed) but I wonder about the function/file name

We set some global compiler flags, which would be good to convert to target options (in a separate PR) and in that context, maybe there's a better name for the function/file in this PR, like `set_target_compiler_strict_warnings` or `set_target_compiler_strict_errors`. I image turning our [current flags](https://github.com/supercollider/supercollider/blob/develop/CMakeLists.txt#L358-L421) into a function like `set_target_compiler_flags` or similar.

What do you think?

JordanHendersonMusic:
Sounds good! I do think we should start all new projects in strict mode, but there's basically no chance of getting the old code to work under it. 

'strict warnings' reads good to me, although it also does warnings as errors, but seems good enough.

I actually didn't know you could pull things out into files and define functions in cmake, we might be able to simplify a lot this way?

dyfer:
I didn't know either! Yes, some stuff can certainly be simplified - some of the stuff that we set globally now but would be better set as target parameters.

dyfer:
Cmake looks good to me!
Do we need cpp review before merging?

JordanHendersonMusic:
Nope, its just basic stuff.

langutils/sc_lexer/CMakeLists.txt
```diff
@@ -8,7 +10,9 @@ add_library(sc_lexer STATIC
 
 target_include_directories(sc_lexer PUBLIC include)
 
+
```
dyfer: Could you remove this whitespace change?

dyfer (approved):
CMake part looks good to me!
(if possible, please remove the whitespace change before merging)

langutils/sc_lexer/CMakeLists.txt
JordanHendersonMusic:
```suggestion
```

PR #7508, opened 2026-05-21
build: msvc add diagnostic:caret

JordanHendersonMusic:
## Purpose and Motivation

before:

```
example.cpp
<source>(7): error C2220: the following warning is treated as an error
<source>(7): warning C4189: 'i': local variable is initialized but not referenced
```
after
```
example.cpp
<source>(7,24): error C2220: the following warning is treated as an error
    const std::uint8_t i {0};
                       ^
<source>(7,24): warning C4189: 'i': local variable is initialized but not referenced
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -404,6 +404,7 @@ if(MSVC)
     /wd4101 # unreferenced local variable
     /wd4102 # unreferenced label
     /wd4200 # nonstandard extension used: zero-sized array in struct/union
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
This is particularly useful when msvc fails during ci for some reason or another. We *could* just add it there if there is a reason to not do it here.

dyfer (approved):
Oh, that seems super useful! Thanks

PR #7510, opened 2026-05-22
sclexer: redefine newline, making \r, \v, and \f errors, make all newlines in literals '\n'

### LINKED ISSUE #7520: sclang: major bug, editors change the content of strings and other text literal (CRLF)
TLDR:
Sometimes, if different text editors/operating system were used to write each string...
```supercollider
"meow
" 
!=
"meow
";
```

Worst still, on windows...
```
A {
   *str{^"meow
"}
}

A.str != "meow
";
```
Yes, whether you define a string in a class file or in a scd file changes it's content!

## Steps to reproduce

***Type out*** this code in scide, on linux and mac this will print `10`, i.e., the `\n` character.  
```supercollider
a = "hello
";
a[5].asInteger.postln;
```

***Type*** the exact same code into Notepad++ on windows and this will return `13`, the `\r` character.

This is because windows uses CRLF (`\r\n`) for its new lines. SCIDE doesn't do this, but other text editors are free to do as they please.

Note: a similar thing can also happen if you open a really old mac os file (before osx) as they used `\r` for new lines.

This doesn't just apply to strings...
```supercollider
a = $
;
a2 = $\
;
b = ' hello\
meow\
';
```
Above are some more cases where this is confusing.
In fact the character literal is particularly confusing because you don't know if it is a space or a new line.

This is a huge bug because depending on what editor and operating system you use, strings which **appear** equal are not! This is particularly an issue if people are working together over a network!

Why is the class library string different from the cmd mode one? Well, when loading the class library we use `std::filesystem::file.open` which automatically concerts `\n` to `\r\n` on windows, but scide does not do this cmd code, it enforces Unix style endings `\n`.****

Using whitespace to specify code is a terrible idea, because you can't see it.

#### 1
Make all strings and symbols convert `\r\n` into `\n`, note: this **does not** apply if the user literally writes `"\r\n"`. Also, forbid \r, \v, and \f in any text literal, if you want these, use the escape character.  
#### 2
Forbid the following syntaxs as it is impossible to tell if it is a newline, a carriage return, or a space (without highlighting the text or inspecting the raw bytes). Instead, this should be a syntax error, or a space (if one is present).

```
$
;
$\
;
``` 

If you want a new line do `$\n` or `$\r`.

This is a breaking change, but I think it is a good one because right now this is impossible to read (because it is whitespace, invisible text), and makes collaborating dangerous.

---

Side note, I think the fact you can escape literals in symbols is in fact a bug and should be removed..
```
'line one\
line two\
line three';
```
But this might be a large breaking change.

JordanHendersonMusic:
This PR does three things:

1. Make all instances of bare `\r`, `\v` or `\f` errors. These are not allowed in the source code because editors cannot agree on how to display them. This means old Mac code (pre osx) needs updating.

1. All literals that contain new lines are turned into `\n` (even if they are `\r\n`) — see #7520 for a description of why this is bad.

1. Syntax for the character literal cannot contain a newline 
```
$
;
```
This is not allowed, you instead must do `$\n`. Note: `$ ` is still allowed as this is a space.

---

In the old lexer there was no definition of a new line, this led to the parser/compiling thinking code was on a line it wasn't.
As an example, it counted '\r' as a line break, but it also counted '\n', meaning the windows new line '\r\n' was counted twice.
Likewise, you could use the vertical tab '\v' or formfeed '\f' as whitespace, but neither of these were counted as new lines.
This gets really confusing as certain IDEs would/wouldn't display this as a new line.
For example, SCIDE would display a little squre with a question mark in for the vertical tab, but the terminal will display the vertical tab correctly (like a new line).
This means you could have single line comments over several lines in certain editors....

This PR defines new lines to be either '\n', '\r\n', or a couple of specific unicode new lines (these are unambiguious and if an editor supports utf8 and doesn't display them as a newline, it is wrong, not us).
The bare carriage return '\r', vertical tab '\v' and form feed '\f' are all made illegal and cannot appear in any source code.

This is **very** important if we want to know where in the file our source code comes from, which is a requirement for both good error messages and a debugger!

---

## Purpose and Motivation

For an example of the issues caused by fix 2, see #7520 

An example of issue 1 from the sc3 plugins is as follows...

The file BiquadEq.sc in scide
```
/*
RMEQ {
/*
Dattoro
Effect Design
Part 1: Reverberator and Other Filters
JAES, vol 45, no 9, 1997

	Synth.fftScope({
		n = WhiteNoise.ar(0.1);
		f = MouseX.kr( 0.1, 24 );
		b = MouseY.kr( -12, 12 ) * 2;
		[
			RegaliaMitraEQ.ar( n, 5000, f.midiratio - 1, b),
			RBJPeakingEQ.ar( n, 5000, f.midiratio - 1, b )
		]
	})
	
	Synth.fftScope({RegaliaMitraEQ.ar( WhiteNoise.ar(-20.dbamp), MouseY.kr(100, 1000, \exponential), 12.midirq, MouseX.kr(-20, 20))})	
	scope({RegaliaMitraEQ.ar( WhiteNoise.ar(-60.dbamp), LFNoise2.kr(1, 50, 300), LFNoise2.kr( 5, 6, 15).midiratio - 1, LFNoise1.kr( 2.1, 12, 30))})	.




... blah blah blah it looks good!
```
printing in the terminal using `less` , using neovim, and displaying in github
```
/*
RMEQ {^M/*^MDattoro^MEffect Design^MPart 1: Reverberator and Other Filters^MJAES, vol 45, no 9, 1997^M  Synth.fftScope({^M              n = WhiteNoise.ar(0.1);^M            f = MouseX.kr( 0.1, 24 );^M             b = MouseY.kr( -12, 12 ) * 2;^M         [^M                     RegaliaMitraEQ.ar( n, 5000, f.midiratio - 1, b),^M                   RBJPeakingEQ.ar( n, 5000, f.midiratio - 1, b )^M                ]^M     })^M    ^M      Synth.fftScope({RegaliaMitraEQ.ar( WhiteNoise.ar(-20.dbamp), MouseY.kr(100, 1000, \exponential), 12.midirq, MouseX.kr(-20, 20))})    ^M      scope({RegaliaMitraEQ.ar( WhiteNoise.ar(-60.dbamp), LFNoise2.kr(1, 50, 300), LFNoise2.kr( 5, 6, 15).midiratio - 1, LFNoise1.kr( 2.1, 12, 30))})      ^M*/^M  *ar { arg in, freq=440, rq=0.1, dbgain=0;^M          ^M              var wc, w2, w1, dw, q, beta, sr, y;^M           var b0, b1, b2, a0, a1, a2, az, k;^M                 ^M              k = dbgain.dbamp;^M                             ^M              sr = 2pi / SampleRate.ir;^M             wc = sr * freq;                              // freq in radians^M            dw = rq * wc;                           // bandwidth in radians^M               dw = (dw/2).tan;                     // delta is only used to compute <A7>^M         beta = (1 - dw) / (1 + dw);     // <A7>^M               y = cos(wc).neg;                             // lattice coefficient^M                ^M              b0 = beta;^M            b1 = y * (1 + beta);^M          b2 = 1;^M                    ^M              a0 = 1;^M               a1 = b1;        // y * (1 + beta);^M            a2 = beta;^M            ^M//            az = SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg );^M              az = SOS.ar( in, b0, b1, b2, a1.neg, a2.neg );^M        ^M              ^( k * (in - az) ) + (in + az) * 0.5^M       }^M}
*/
^M/*^MImplementation of filters as described at:^M<http://www.harmony-central.com/Computer/Programming/Audio-EQ-Cookbook.txt>^M*/^M/*^MPeakingEQ {^M         /*^M            scope({RBJPeakingEQ.ar( WhiteNoise.ar(-60.dbamp), MouseY.kr(100, 1000, \exponential), 12.midiratio - 1, MouseX.kr(0, 45))})          ^M              scope({RBJPeakingEQ.ar( WhiteNoise.ar(-60.dbamp), LFNoise2.kr(1, 50, 300), LFNoise2.kr( 5, 6, 15).midiratio - 1, LFNoise1.kr( 2.1, 12, 30))})        ^M      */^M    *ar { arg in, freq=440, rq=0.1, dbgain=0, mul=1, add=0;^M               var b0, b1, b2, a0, a1, a2;^M                var a, omega, sin, cos, alpha, a0n;^M           var aovera, atimesa;^M          a = dbgain.dbamp;       ^M              omega = 2pi * freq/SampleRate.ir;^M          sin = omega.sin;^M              cos = omega.cos;^M              alpha = sin * rq / 2;^M         aovera = (alpha/a);^M                atimesa = (alpha*a);^M          ^M              a0 = 1 + aovera;^M              a1 = -2 * cos;^M                a2 = 1 - aovera;^M                   b0 = 1 + atimesa;^M             b1 = a1;^M              b2 = 1 - atimesa;^M             ^M              a0n = a0.neg;^M         ^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0n, a2/a0n, mul, add )^M       }^M}^M*/^MNotch  {^M    /*^M            Synth.fftScope({RBJNotch.ar( WhiteNoise.ar(0.1), 4400, MouseX.kr(0.01, 2))}) ^M              Synth.fftScope({RBJNotch.ar( SinOsc.ar(LFTri.kr(0.1, 1000, 1100), 0, 0.1), 500, MouseX.kr(0.01, 2))})   ^M           */^M    *ar { arg in, freq=440, rq=0.1, mul=1, add=0;^M         var b0, b1, b2, a0, a1, a2;^M           var a, omega, sin, cos, alpha, beta, q;^M            ^M              omega = 2pi * freq/SampleRate.ir;^M             sin = omega.sin;^M              cos = omega.cos;^M              alpha = rq * sin/2;^M                a0 = 1 + alpha;^M               a1 = -2 * cos;^M                a2 = 1 - alpha;^M               b0 = 1;^M                    b1 = a1;        // -2 * cos;^M          b2 = 1;^M               ^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )^M    }^M}

LowShelf  {^M   /*^M            Synth.fftScope({RBJLowShelf.ar( WhiteNoise.ar(0.1), 400, 1, MouseX.kr(-24, 24))})       ^M      */^M    *ar { arg in, freq=440, shelfslope=1, dbgain=0, mul=1, add=0;^M              var am1, ap1, bs, am1cos, ap1cos, q;^M          var b0, b1, b2, a0, a1, a2;^M                var a, omega, sin, cos, alpha, beta;^M          ^M              a = dbgain.dbamp;       ^M              omega = 2pi * freq/SampleRate.ir;^M                  sin = omega.sin;^M              cos = omega.cos;^M              beta = a.sqrt * ((a+a.reciprocal) * (shelfslope.reciprocal - 1) + 2).sqrt;^M//               beta = (((a.squared + 1) / shelfslope) - (a - 1).squared).sqrt;^M//             q = ((a + a.reciprocal) * (shelfslope.reciprocal - 1) + 2).sqrt.reciprocal;^M//              q.postln;^M//           beta = a.sqrt / q;^M            am1 = a-1;^M            ap1 = a+1;^M                 am1cos = am1 * cos;^M           ap1cos = ap1 * cos;^M           bs = beta * sin;^M              ^M              a0 = ap1 + am1cos + bs;^M                    a1 = -2 * ( am1 + ap1cos );^M           a2 = ap1 + am1cos - bs;^M               b0 = a * ( ap1 - am1cos + bs );^M               b1 = 2 * a * ( am1 - ap1cos );^M             b2 = a * ( ap1 - am1cos - bs );^M               ^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )^M }^M}
^MHighShelf  {^M        /*^M            Synth.fftScope({RBJHighShelf.ar( WhiteNoise.ar(0.1), 11000, 1, MouseX.kr(-6, 6))})      ^M      */^M    *ar { arg in, freq=440, shelfslope=1, dbgain=0, mul=1, add=0;^M              var am1, ap1, bs, am1cos, ap1cos;^M             var b0, b1, b2, a0, a1, a2, a0n;^M           var a, omega, sin, cos, alpha, beta, q;^M               ^M              a = dbgain.dbamp;       ^M              omega = 2pi * freq/SampleRate.ir;^M          sin = omega.sin;^M              cos = omega.cos;^M//            beta = (((a.squared + 1) / shelfslope) - (a - 1).squared).sqrt;^M            beta = a.sqrt * ((a+a.reciprocal)*(shelfslope.reciprocal-1)+2).sqrt;^M//                q = ((a + a.reciprocal) * (shelfslope.reciprocal - 1) + 2).sqrt.reciprocal;^M//              beta = a.sqrt / q;^M            am1 = a-1;^M            ap1 = a+1;^M            am1cos = am1 * cos;^M                ap1cos = ap1 * cos;^M           bs = beta * sin;^M              ^M              a0 = ap1 - am1cos + bs;^M               a1 = 2 * ( am1 - ap1cos );^M         a2 = ap1 - am1cos - bs;^M               b0 = a * ( ap1 + am1cos + bs );^M               b1 = -2 * a * ( am1 + ap1cos );^M                    b2 = a * ( ap1 + am1cos - bs );^M               ^M              a0n = a0.neg;^M         ^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0n, a2/a0n, mul, add )^M       }^M}

// old verbose names
RegaliaMitraEQ : RMEQ {}
//RBJPeakingEQ : PeakingEQ {}
RBJNotch : Notch {}^MRBJHighShelf : HighShelf {}
RBJLowShelf : LowShelf {}
```

printing in the terminal using `cat`
```
/*
JAES, voscope({RegaliaMitraEQ.ar( WhiteNoise.ar(-60.dbamp), LFNoise2.kr(1, 50, 300), LFNoise2.kr( 5, 6, 15).midiratio - 1, LFNoise1.kr( 2.1, 12, 30))}/      }ar { ar^( k * (in - az) ) + (in + az) * 0.5 a2.neg );g, a2/a0.neg );tto compute �
*/
PeakingE/*{.harmscope({RBJPeakingEQ.ar( WhiteNoise.ar(-60.dbamp), LFNoise2.kr(1, 50, 300), LFNoise2.kr( 5, 6, 15).midiratio - 1, LFNoise1.kr( 2.1, 12}otch  {}ar { ar^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )0, 0.1), 500, MouseX.kr(0.01, 2))})

}/wShelf}ar { ar^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )ciprocal;24))})
}/ghShel}ar { ar^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0n, a2/a0n, mul, add )qrt.reciprocal;, 6))})

// old verbose names
RegaliaMitraEQ : RMEQ {}
//RBJPeakingEQ : PeakingEQ {}
RBJHighShelf : HighShelf {}
RBJLowShelf : LowShelf {}
```

As you can see, the second option displays the bare carriage returns as `^M`, but the third actually does this properly and goes back to the beginning of the line and overwrites the text, essentially writing everything on one line...

There is a similar issue regarding vertical tab, where this doesn't end a comment and is allowed in quoted symbols. IDEs may choose to render this on a new line (correctly).
```
\\ a comment ? more comment
```
or
```
\\ a comment 
           more comment
```

This is bad because it not only means we don't know what line anything is on, but what is an isn't **read*** by the user as code changes.

## Types of changes

- Bug fix
- Breaking change

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation ??? 
- [ ]  Need to retest error printing and line number logic because that changed.

### DIFF
--- a/langutils/sc_lexer/include/lexer.hpp
+++ b/langutils/sc_lexer/include/lexer.hpp
@@ -262,6 +262,23 @@ decltype(auto) lexer_digits(CodePointStream& stream, Action& action, SourceCodeL
 
 }
 
+inline void consume_new_lines(CodePointStream& stream) {
+    while (true) {
+        const auto [c, n] = stream.peek_n<2>().characters;
+        if (is_single_newline(c)) {
+            stream.advance();
+            continue;
+        } else if (c == '\r' && n == '\n') {
+            stream.advance();
+            stream.advance();
+            continue;
+        }
+        return;
+    }
+}
+
+
+// Before calling this, you must ensure validate_source_format has been called.
 template <typename Action> auto lexer(CodePointStream& stream, Action& action) noexcept -> typename Action::Output {
     using namespace details;
 
@@ -290,11 +307,22 @@ next_token : {
         return *r;
     }
 
-    if (is_newline(code_point)) {
-        stream.advance_while(is_newline);
-        return_orelse_next_token(action.template process<TokenType::NewLine>({ token_start, stream.end_token() }));
+    if (code_point == '\r') {
+        const auto n = stream.peek();
+        if (n == '\n') {
+            stream.advance(); // consume '\n'
+            consume_new_lines(stream);
+            return_orelse_next_token(action.template process<TokenType::NewLine>({ token_start, stream.end_token() }));
+        } else {
+            return_orelse_next_token(action.template process<TokenType::ErFoundInvalidNewlineCharaceter>(
+                { token_start, stream.end_token() }));
+        }
     }
 
+    if (is_single_newline(code_point)) {
+        consume_new_lines(stream);
+        return_orelse_next_token(action.template process<TokenType::NewLine>({ token_start, stream.end_token() }));
+    }
 
     if (code_point == '\t') {
         stream.advance_while(is_tab);
@@ -306,7 +334,6 @@ next_token : {
         return_orelse_next_token(action.template process<TokenType::Space>({ token_start, stream.end_token() }));
     }
 
-    // ASCII tokens. Their ASCII value is their token type value.
     switch (code_point) {
     case '^':
         return_orelse_next_token(
@@ -333,6 +360,10 @@ next_token : {
         return_orelse_next_token(action.template process<TokenType::CloseSquare>({ token_start, stream.end_token() }));
     case '}':
         return_orelse_next_token(action.template process<TokenType::CloseCurly>({ token_start, stream.end_token() }));
+    case '\v':
+    case '\f':
+        return_orelse_next_token(
+            action.template process<TokenType::ErFoundInvalidNewlineCharaceter>({ token_start, stream.end_token() }));
     }
 
     if (code_point == '#') {
@@ -351,37 +382,54 @@ next_token : {
 
         // '/*'
         if (p == '*') {
-            stream.advance();
-            CodePoint it_1 = 0, it = 0;
+            stream.advance(); // consume '*';
             size_t level = 1;
-            stream.advance_while([&](auto c) {
-                it_1 = it;
-                it = c;
-                if (it_1 == '/' && it == '*') {
-                    level += 1;
-                    it = 0; // consume
-                    return true;
-                }
-                if (it_1 == '*' && it == '/') {
+            while (true) {
+                const auto [c, n] = stream.peek_n<2>().characters;
+                if (c == '*' && n == '/') {
+                    stream.advance();
+                    stream.advance();
                     level -= 1;
-                    it = 0; // consume
-                    return level != 0;
+                    if (level == 0) {
+                        return_orelse_next_token(
+                            action.template process<TokenType::MultilineComment>({ token_start, stream.end_token() }));
+                    }
+                    continue;
+                } else if (c == '/' && n == '*') {
+                    stream.advance();
+                    stream.advance();
+                    level += 1;
+                    continue;
+                } else if ((c == '\r' && n != '\n') || c == '\v' || c == '\f') {
+                    stream.advance();
+                    return_orelse_next_token(action.template process<TokenType::ErFoundInvalidNewlineCharaceter>(
+                        { token_start, stream.end_token() }));
+                } else if (c == 0) {
+                    return_orelse_next_token(action.template process<TokenType::ErMultilineCommentUnclosed>(
+                        { token_start, stream.end_token() }));
+                } else {
+                    stream.advance();
+                    continue;
                 }
-                return true;
-            });
-            const auto delimit = stream.advance();
-            if (delimit == 0) {
-                return_orelse_next_token(action.template process<TokenType::ErMultilineCommentUnclosed>(
-                    { token_start, stream.end_token() }));
             }
-            return_orelse_next_token(
-                action.template process<TokenType::MultilineComment>({ token_start, stream.end_token() }));
         }
 
         // '//'
         else if (p == '/') {
-            stream.advance_while([](auto c) { return !is_newline(c); });
-            return_orelse_next_token(action.template process<TokenType::Comment>({ token_start, stream.end_token() }));
+            while (true) {
+                const auto [c, n] = stream.peek_n<2>().characters;
+                if ((c == '\r' && n != '\n') || c == '\v' || c == '\f') {
+                    stream.advance();
+                    return_orelse_next_token(action.template process<TokenType::ErFoundInvalidNewlineCharaceter>(
+                        { token_start, stream.end_token() }));
+                } else if ((c == '\r' && n == '\n') || is_single_newline(c) || c == 0) {
+                    return_orelse_next_token(
+                        action.template process<TokenType::Comment>({ token_start, stream.end_token() }));
+                } else {
+                    stream.advance();
+                    continue;
+                }
+            }
         }
 
         // Its a binary operator.
@@ -400,13 +448,26 @@ next_token : {
     if (code_point == '$') {
         if (stream.peek() == '\\') {
             stream.advance(); // consume '\'
-            stream.advance(); // consume whatever happens after it.
+        }
+        const auto [c, n] = stream.peek_n<2>().characters;
+        if (c > 127 || c < 0) {
+            // Can only be ascii.
+            return_orelse_next_token(
+                action.template process<TokenType::ErUnicodeInCharacter>({ token_start, stream.end_token() }));
+        }
+        if (c == '\r' && n == '\n') {
+            stream.advance(); // consume \r
+            stream.advance(); // consume \n
+            return_orelse_next_token(action.template process<TokenType::ErFoundInvalidNewlineCharaceter>(
+                { token_start, stream.end_token() }));
+        } else if (c == '\r' || c == '\v' || c == '\f' || is_single_newline(c) || c == 0 || is_tab(c)) {
+            stream.advance(); // consume white space
+            return_orelse_next_token(action.template process<TokenType::ErFoundInvalidNewlineCharaceter>(
+                { token_start, stream.end_token() }));
+        } else {
+            stream.advance(); // consume
             return_orelse_next_token(action.template process<TokenType::Ascii>({ token_start, stream.end_token() }));
         }
-
-        stream.advance(); // consume whatever character comes after the '$'
-
-        return_orelse_next_token(action.template process<TokenType::Ascii>({ token_start, stream.end_token() }));
     }
 
     if (is_starting_identifier(code_point) || code_point == '_')
@@ -422,21 +483,32 @@ next_token : {
 
     // Strings
     if (code_point == '"') {
-        const auto end = stream.advance_while([escaped = false](auto c) mutable {
-            if (c == '\\' && !escaped) {
+        bool escaped { false };
+        while (true) {
+            const auto [c, n] = stream.peek_n<2>().characters;
+            if ((c == '\r' && n != '\n') || c == '\v' || c == '\f') {
+                // bad newline
+                stream.advance();
+                return_orelse_next_token(action.template process<TokenType::ErFoundInvalidNewlineCharaceter>(
+                    { token_start, stream.end_token() }));
+            } else if (c == '\\' && !escaped) {
                 escaped = true;
-                return true;
+                stream.advance();
+                continue;
+            } else if (c == '"' && !escaped) {
+                // end of string.
+                stream.advance();
+                return_orelse_next_token(
+                    action.template process<TokenType::StringLine>({ token_start, stream.end_token() }));
+            } else if (c == 0) {
+                return_orelse_next_token(
+                    action.template process<TokenType::ErStringUnclosed>({ token_start, stream.end_token() }));
+            } else {
+                escaped = false;
+                stream.advance();
+                continue;
             }
-            if (c == '"' && !escaped)
-                return false;
-            escaped = false;
-            return true;
-        });
-
-        if (stream.advance() != '"')
-            return_orelse_next_token(action.template process<TokenType::ErStringUnclosed>({ token_start, end }));
-
-        return_orelse_next_token(action.template process<TokenType::StringLine>({ token_start, stream.end_token() }));
+        }
     }
 
     // Symbol that begin with a '\'. Note: first character is used to alter what is acceptable in the following.
@@ -459,24 +531,52 @@ next_token : {
 
     // Symbol quotes 'asdf'
     if (code_point == '\'') {
-        const auto content_end = stream.advance_while([escape = false](auto c) mutable {
-            if (is_newline(c) && !escape) // you can escape the new line characters
-                return false;
-            if (c == '\'' && !escape)
-                return false;
-            if (c == '\\' && !escape) {
-                escape = true;
-                return true;
-            }
-            escape = false;
-            return true;
-        });
-        if (stream.peek() != '\'')
-            return_orelse_next_token(
-                action.template process<TokenType::ErSymbolQuoteUnclosed>({ token_start, content_end }));
+        // NOTE: should you be able to escape these newline literal characters?
 
-        stream.advance();
-        return_orelse_next_token(action.template process<TokenType::SymbolQuote>({ token_start, stream.end_token() }));
+        bool escape { false };
+        while (true) {
+            const auto [c, n] = stream.peek_n<2>().characters;
+
+            if ((c == '\r' && n != '\n') || c == '\v' || c == '\f') {
+                // bad newline
+                stream.advance();
+                return_orelse_next_token(action.template process<TokenType::ErFoundInvalidNewlineCharaceter>(
+                    { token_start, stream.end_token() }));
+            } else if (((c == '\r' && n == '\n') || is_single_newline(c)) && !escape) {
+                // Valid new line, but not esacped.
+                stream.advance();
+                return_orelse_next_token(
+                    action.template process<TokenType::ErSymbolQuoteUnclosed>({ token_start, stream.end_token() }));
+            } else if (c == '\'' && escape) {
+                escape = false;
+                stream.advance();
+                continue;
+            } else if (c == '\'' && !escape) {
+                stream.advance();
+                return_orelse_next_token(
+                    action.template process<TokenType::SymbolQuote>({ token_start, stream.end_token() }));
+            } else if (c == '\\') {
+                escape = !escape;
+                stream.advance();
+                continue;
+            } else if (c == 0) {
+                return_orelse_next_token(
+                    action.template process<TokenType::ErSymbolQuoteUnclosed>({ token_start, stream.end_token() }));
+            } else if (c == '\r' && n == '\n' && escape) {
+                stream.advance();
+                stream.advance();
+                escape = false;
+                continue;
+            } else if (is_single_newline(c) && escape) {
+                stream.advance();
+                escape = false;
+                continue;
+            } else {
+                stream.advance();
+                escape = false;
+                continue;
+            }
+        }
     }
 
     // These are the control codes, throw them away!
--- a/lang/LangSource/PyrLexer.cpp
+++ b/lang/LangSource/PyrLexer.cpp
@@ -608,6 +608,13 @@ private:
         bool escaped = false;
         auto from_it = start;
         for (; from_it != end; from_it += 1) {
+            if (from_it[0] == '\r' && from_it[1] == '\n') {
+                temp_buffer.push_back('\n');
+                // It doesn't matter if you esacpe the newline character or not, it does nothing.
+                escaped = false;
+                from_it += 1; // skip '\n' aswell
+                continue;
+            }
             if (*from_it == '\\' && !escaped) {
                 escaped = true;
                 // don't write
@@ -689,7 +696,7 @@ void print_error_line(const char* filepath, const char* txt, size_t txt_len, sc:
                 if (cp_iter.current_location() <= selection_start) {
                     ss << sc::lex::codepoint_as_whitespace(*cp);
                 } else if (cp_iter.current_location() <= selection_end) {
-                    const auto w { sc::lex::codepoint_width(*cp) };
+                    const auto w = std::max<std::uint8_t>(sc::lex::codepoint_width(*cp), 1);
                     for (size_t i { 0 }; i < w; ++i)
                         ss << '^';
                 } else
@@ -837,9 +844,11 @@ struct GlobalBisonLexerState {
                 const auto desc = std::string { "This multiline comment does not have a closing */." };
                 print_error_line(gCompilingFileSym->name, char_stream.source, char_stream.source_length, o.range,
                                  desc.c_str());
-            }
-
-            else {
+            } else if (o.is(TokenType::ErFoundInvalidNewlineCharaceter)) {
+                print_error_line(gCompilingFileSym->name, char_stream.source, char_stream.source_length, o.range,
+                                 "\'\\v\' and \'\\f\' are no longer supported, please replace them with a new line "
+                                 "character \'\\n\'.");
+            } else {
                 print_error_line(gCompilingFileSym->name, char_stream.source, char_stream.source_length, o.range);
             }
         }
@@ -976,10 +985,13 @@ int yylex() {
                     str += '\v';
                 else
                     str += *b;
-                escaped = false;
+            } else if (b[0] == '\r' && b[1] == '\n') {
+                str += '\n';
+                b += 1;
             } else {
                 str += *b;
             }
+            escaped = false;
         }
 
         prev = out;
@@ -1850,6 +1862,7 @@ SCLANG_DLLEXPORT_C void runLibrary(PyrSymbol* selector) {
 
 bool startLexer(PyrSymbol* fileSym, const fs::path& p, int startPos, int endPos, int lineOffset) {
     const char* filename = fileSym->name;
+    gCompilingFileSym = fileSym;
 
     gCompilinTextLen = -1;
 
--- a/lang/LangSource/PyrParseNode.cpp
+++ b/lang/LangSource/PyrParseNode.cpp
@@ -4227,7 +4227,16 @@ void PyrBlockNode::compile(PyrSlot* slotResult) {
         SetNil(&block->contextDef);
 
         PyrString* string = newPyrStringN(compileGC(), location.size(), flags, false);
-        memcpy(string->s, gCompilingText + location.begin.absolute, location.size());
+        const auto end = location.end.absolute;
+        for (size_t read { location.begin.absolute }, write { 0 }; read < end; ++read, ++write) {
+            const auto c = gCompilingText[read];
+            if (c == '\r') {
+                ++read;
+                // This should be enforced by the lexer.
+                assert(gCompilingText[read] == '\n');
+            }
+            string->s[write] = gCompilingText[read];
+        }
         SetObject(&block->sourceCode, string);
     }
 
--- a/langutils/sc_lexer/include/codepoint.hpp
+++ b/langutils/sc_lexer/include/codepoint.hpp
@@ -33,10 +33,9 @@ inline constexpr CodePoint invalid_utf8_flag { std::numeric_limits<CodePoint>::m
 
 
 // Note: these are supercollider's defintions, not unicode's.
-[[nodiscard]] bool is_newline(CodePoint c) noexcept;
+[[nodiscard]] bool is_single_newline(CodePoint c) noexcept;
 [[nodiscard]] bool is_space(CodePoint c) noexcept;
 [[nodiscard]] bool is_tab(CodePoint c) noexcept;
-[[nodiscard]] bool is_whitespace(CodePoint c) noexcept;
 [[nodiscard]] bool is_control_code(CodePoint c) noexcept;
 [[nodiscard]] bool is_lower(CodePoint c) noexcept;
 [[nodiscard]] bool is_upper(CodePoint c) noexcept;
--- a/langutils/sc_lexer/include/codepoint_stream.hpp
+++ b/langutils/sc_lexer/include/codepoint_stream.hpp
@@ -6,6 +6,7 @@
 
 #include <text_location.hpp>
 #include <codepoint.hpp>
+#include <type_traits>
 
 namespace sc::lex {
 
@@ -41,7 +42,6 @@ class CodePointStream {
         std::size_t current_line_number { 0 };
         // Doesn't work with tabs, fix this once the main parser no longer needs this.
         std::size_t current_column_count { 0 };
-        bool prev_was_newline { false };
         void update(CodePoint next, std::uint8_t sz) noexcept;
     } state {};
 
@@ -139,12 +139,28 @@ public:
 
     // Null terminator is NEVER accepted as a predicate.
     template <typename Predicate> std::size_t advance_while_count(Predicate&& predicate) {
-        auto discard_null_then_predicate = [&](auto c) {
-            return (c == 0 || state.next_byte_offest >= source_length) ? false : predicate(c);
-        };
-        std::size_t i { 0 };
-        for (auto c = peek(); discard_null_then_predicate(c); c = advance_and_peek(), ++i) {}
-        return i;
+        if constexpr (std::is_invocable_v<Predicate, CodePoint, CodePoint>) {
+            auto discard_null_then_predicate = [&](auto c, auto n) {
+                return (c == 0 || state.next_byte_offest >= source_length) ? false : predicate(c, n);
+            };
+            std::size_t i { 0 };
+            while (true) {
+                auto p = peek_n<2>();
+                if (discard_null_then_predicate(p.at<0>(), p.at<1>())) {
+                    advance();
+                    ++i;
+                } else {
+                    return i;
+                }
+            }
+        } else {
+            auto discard_null_then_predicate = [&](auto c) {
+                return (c == 0 || state.next_byte_offest >= source_length) ? false : predicate(c);
+            };
+            std::size_t i { 0 };
+            for (auto c = peek(); discard_null_then_predicate(c); c = advance_and_peek(), ++i) {}
+            return i;
+        }
     }
 
     template <typename Predicate> SourceCodeLocation advance_while(Predicate&& predicate) {
--- a/langutils/sc_lexer/include/source_utils.hpp
+++ b/langutils/sc_lexer/include/source_utils.hpp
@@ -22,6 +22,7 @@ public:
     CodePointIterator& operator=(const CodePointIterator&) = delete;
 
     std::optional<CodePoint> forwards() noexcept;
+    // Note: this requires valid utf8.
     std::optional<CodePoint> backwards() noexcept;
 
     void jump(std::int64_t m) noexcept { txt_iter += m; }
--- a/langutils/sc_lexer/include/text_location.hpp
+++ b/langutils/sc_lexer/include/text_location.hpp
@@ -26,12 +26,12 @@ struct SourceCodeLocation {
 
 // A location point inside an entire file
 struct FileCodeLocation {
-    [[nodiscard]] bool operator==(const SourceCodeLocation& o) const noexcept;
-    [[nodiscard]] bool operator!=(const SourceCodeLocation& o) const noexcept;
-    [[nodiscard]] bool operator<(const SourceCodeLocation& o) const noexcept;
-    [[nodiscard]] bool operator>(const SourceCodeLocation& o) const noexcept;
-    [[nodiscard]] bool operator<=(const SourceCodeLocation& o) const noexcept;
-    [[nodiscard]] bool operator>=(const SourceCodeLocation& o) const noexcept;
+    [[nodiscard]] bool operator==(const FileCodeLocation& o) const noexcept;
+    [[nodiscard]] bool operator!=(const FileCodeLocation& o) const noexcept;
+    [[nodiscard]] bool operator<(const FileCodeLocation& o) const noexcept;
+    [[nodiscard]] bool operator>(const FileCodeLocation& o) const noexcept;
+    [[nodiscard]] bool operator<=(const FileCodeLocation& o) const noexcept;
+    [[nodiscard]] bool operator>=(const FileCodeLocation& o) const noexcept;
 
     // Offset as a byte index into the text.
     std::size_t absolute { 0 };
@@ -47,6 +47,12 @@ struct SourceCodeRange {
     [[nodiscard]] static SourceCodeRange range(SourceCodeRange left, SourceCodeRange right);
     [[nodiscard]] std::size_t size() const;
     [[nodiscard]] std::size_t line_count() const;
+    [[nodiscard]] bool operator==(const SourceCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator!=(const SourceCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator<(const SourceCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator>(const SourceCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator<=(const SourceCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator>=(const SourceCodeRange& o) const noexcept;
 
     SourceCodeLocation begin, end;
 };
@@ -56,7 +62,12 @@ struct FileCodeRange {
     [[nodiscard]] static FileCodeRange range(FileCodeRange left, FileCodeRange right);
     [[nodiscard]] std::size_t size() const;
     [[nodiscard]] std::size_t line_count() const;
-
+    [[nodiscard]] bool operator==(const FileCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator!=(const FileCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator<(const FileCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator>(const FileCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator<=(const FileCodeRange& o) const noexcept;
+    [[nodiscard]] bool operator>=(const FileCodeRange& o) const noexcept;
     FileCodeLocation begin, end;
 };
 
--- a/langutils/sc_lexer/include/tokens.hpp
+++ b/langutils/sc_lexer/include/tokens.hpp
@@ -137,6 +137,10 @@ enum struct TokenType : unsigned int {
         ErMissingExponent,
         ErInvalidToken,
         ErInvalidUTF8,
+        ErUnicodeInCharacter,
+
+        ErFoundInvalidNewlineCharaceter, // '\v' or '\f', replace with '\n'
+
         ErUnknown,
 
         START_OF_USER_DEFINED_ERRORS,
--- a/langutils/sc_lexer/src/codepoint.cpp
+++ b/langutils/sc_lexer/src/codepoint.cpp
@@ -13,6 +13,7 @@ namespace sc::lex {
     if (width_or_error < 0)
         // Note: all these '1' need to be initalised like std::uint8_t{1} because MSVC thinks '1' is an int and can't
         // see the conversion is fine.
+        // Zero is valid ONLY when dealing with the null terminator.
         return { invalid_utf8_flag, std::uint8_t { 1 } };
     else
         return { cp, width };
@@ -37,10 +38,9 @@ namespace sc::lex {
     CodePoint cp;
     const auto width_or_error = utf8proc_iterate(c, count, &cp);
     const auto width { static_cast<std::uint8_t>(width_or_error) };
-    if (width_or_error < 0)
-        return { invalid_utf8_flag, std::uint8_t { 1 } };
-    else
-        return { cp, width };
+    // Zero is valid ONLY when dealing with the null terminator.
+    return width_or_error < 0 ? std::tuple<CodePoint, std::uint8_t> { invalid_utf8_flag, std::uint8_t { 1 } }
+                              : std::tuple<CodePoint, std::uint8_t> { cp, width };
 }
 
 [[nodiscard]] std::uint8_t codepoint_size(CodePoint uc) noexcept {
@@ -87,11 +87,12 @@ namespace sc::lex {
         ;
 }
 
-[[nodiscard]] bool is_newline(CodePoint c) noexcept {
+// Does not work for \r\n
+[[nodiscard]] bool is_single_newline(CodePoint c) noexcept {
+    // Carriage return does not count.
+    // Code that used a bare carriage return for a new line must be updated.
+    // Likewise vertical tab and form feed are not counted, replace these with a '\n'.
     return c == '\n' // line feed
-        || c == '\v' // vertical tab (not classified as a tab)
-        || c == '\f' // form feed
-        || c == '\r' // carrage return
         || c == 0x0085 // next line
         || c == 0x2028 // line separator
         || c == 0x2029 // paragraph separator
@@ -99,8 +100,6 @@ namespace sc::lex {
 }
 
 [[nodiscard]] bool is_tab(CodePoint c) noexcept { return c == '\t'; }
-
-[[nodiscard]] bool is_whitespace(CodePoint c) noexcept { return is_newline(c) || is_space(c) || is_tab(c); }
 
 [[nodiscard]] bool is_control_code(CodePoint c) noexcept {
     return (1 <= c && c <= 8) || (14 <= c && c <= 31) || c == 127;
--- a/langutils/sc_lexer/src/codepoint_stream.cpp
+++ b/langutils/sc_lexer/src/codepoint_stream.cpp
@@ -1,4 +1,5 @@
 #include "codepoint_stream.hpp"
+#include "codepoint.hpp"
 
 namespace sc::lex {
 
@@ -26,15 +27,12 @@ CodePointStream::CodePointStream(const char* src, std::size_t src_len, FileCodeL
 }
 
 void CodePointStream::State::update(CodePoint next, std::uint8_t sz) noexcept {
-    if (prev_was_newline) {
+    if (is_single_newline(next)) {
         current_line_number += 1;
         current_column_count = 0;
-        prev_was_newline = false;
     } else {
-        // TODO: this is wrong and should be removed.
-        current_column_count += 1; // codepoint_width(next);
+        current_column_count += 1;
     }
-    prev_was_newline = is_newline(next);
     next_byte_offest += sz;
 }
 
--- a/langutils/sc_lexer/src/source_utils.cpp
+++ b/langutils/sc_lexer/src/source_utils.cpp
@@ -54,7 +54,7 @@ std::optional<LineIter> LineIter::make(const char* txt_start, const char* txt_en
 std::optional<LineIter::Result> LineIter::forwards() noexcept {
     if (previous_dir == Backward) {
         const auto [cp, sz] = cp_iter.current_codepoint();
-        if (is_newline(cp)) {
+        if (is_single_newline(cp)) {
             cp_iter.forwards();
             current_line += 1;
         }
@@ -65,7 +65,7 @@ std::optional<LineIter::Result> LineIter::forwards() noexcept {
 
     bool ends_in_newline_char = false;
     for (auto r = cp_iter.forwards(); r; r = cp_iter.forwards()) {
-        if (is_newline(*r)) {
+        if (is_single_newline(*r)) {
             ends_in_newline_char = true;
             break;
         }
@@ -85,9 +85,9 @@ std::optional<LineIter::Result> LineIter::backwards() noexcept {
     if (previous_dir == Forwards) {}
     previous_dir = Backward;
     const auto end = cp_iter.current_location();
-    bool ends_in_newline_char = is_newline(std::get<0>(cp_iter.current_codepoint()));
+    bool ends_in_newline_char = is_single_newline(std::get<0>(cp_iter.current_codepoint()));
     for (auto r = cp_iter.backwards(); r; r = cp_iter.backwards()) {
-        if (is_newline(*r)) {
+        if (is_single_newline(*r)) {
             cp_iter.forwards(); // undo
             break;
         }
--- a/langutils/sc_lexer/src/text_location.cpp
+++ b/langutils/sc_lexer/src/text_location.cpp
@@ -5,12 +5,10 @@ using namespace sc::lex;
 // assuming the line number and column are correct and that the source text is the same means we only need to comare
 // the absolute byte offset.
 [[nodiscard]] bool SourceCodeLocation::operator==(const SourceCodeLocation& o) const noexcept {
-    return absolute == o.absolute;
+    return absolute == o.absolute && lineNumber == o.lineNumber && column == o.column;
 }
 
-[[nodiscard]] bool SourceCodeLocation::operator!=(const SourceCodeLocation& o) const noexcept {
-    return absolute != o.absolute;
-}
+[[nodiscard]] bool SourceCodeLocation::operator!=(const SourceCodeLocation& o) const noexcept { return !(*this == o); }
 
 [[nodiscard]] bool SourceCodeLocation::operator<(const SourceCodeLocation& o) const noexcept {
     return absolute < o.absolute;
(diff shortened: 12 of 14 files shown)


JordanHendersonMusic:
Thinking about this a little more, it might be better to forbid carriage returns unless followed by a newline character, that would allow us to generate an error message.

langutils/sc_lexer/include/lexer.hpp
```diff
@@ -351,37 +382,54 @@ next_token : {
...
+                        return_orelse_next_token(
+                            action.template process<TokenType::MultilineComment>({ token_start, stream.end_token() }));
+                    }
+                    continue;
+                } else if (c == '/' && n == '*') {
+                    stream.advance();
+                    stream.advance();
+                    level += 1;
+                    continue;
+                } else if ((c == '\r' && c != '\n') || c == '\v' || c == '\f') {
```
JordanHendersonMusic:
```suggestion
                } else if ((c == '\r' && n != '\n') || c == '\v' || c == '\f') {
```

prko:
Does contain this PR `'''` or `"""` discussed in #6846?

dyfer:
For class library files, could this be a warning instead of error in SC 3.15, turning into and error from 3.16 on?
What would it take to: 
- for invalid character, print a warning and skip?
- for invalid newline, convert internally to `\n`, and print a warning?

Again, this would be just for one minor version (3.15) to give users time to update code.

JordanHendersonMusic:
The problem is if you have a file full of \r then you add a new line \n, you now have \r\n, which is a single new line. I don't think it's possible to solve this.

What could be done, is that we add a utility program to convert. This could be accessed in one of the ides drop down menus? I think this would be better. 

For invalid utf8 characters we could turn this into a warning. But having the user convert the file is a bit more complex, in most editors you just change the encoding and resave, I don't know how to do this in the ide.

JordanHendersonMusic:
Checking lucile's corpus gives only these files that use the bare `\r`.

```
./sc3-plugins-main/source/JoshUGens/sc/classes/joshfilts.sc
./sc3-plugins-main/source/JoshUGens/sc/classes/AudioMSG.sc
./sc3-plugins-main/source/JoshUGens/sc/classes/Vocoder.sc
./sc3-plugins-main/source/JoshUGens/sc/classes/Phaser.sc
./sc3-plugins-main/source/JoshUGens/sc/classes/Class Files/CheapVerb.sc
./sc3-plugins-main/source/JoshUGens/sc/classes/Class Files/BiquadEQ.sc
./quarks/crucial-library/Players/SimpleTrigger.sc
./quarks/crucial-library/Players/SimpleTriggerGui.sc
./quarks/crucial-library/Players/PlayerMixerGui.sc
./quarks/JITMIDIKtl/UC33Ktl.sc
./quarks/JITMIDIKtl/MPD24Ktl.sc
```

That is 11 out of 2892 aka 0.38%.

JordanHendersonMusic:
There are no occurrences of `\v` or `\f`.

JordanHendersonMusic:
There is another issue not addressed in this PR which might require a different solution.

```supercollider

f = { |a, b|
   a + b
};

f.def.source != "{ |a, b|
   a + b
}";
```

As we want all string literals to have the same newline encoding (`\n` probably), this becomes really confusing if we allow source code to have `\r\n` new lines.

In the above example, the source code works with the file newline type `\r\n`, but you can't create a string literal with that newline type (without doing ` "{ |a, b|\r\n   a + b \rn}`), instead you always get `\n`. 


For this reason, I think the approach in this PR is wrong and instead we should have a prelexer phase that removes all `\r` from the source. This absolutely breaks all files that currently use `\r` for their newlines, but since mac ditched this 25 years ago, only scide editor shows this properly, it is easy to convert (it isn't the code that is borked, just the file, `tr \r \n`), and the occurrence of this is low in the corpus (0.38%), I believe this is right approach as it ensures everything else works as expected.

dyfer:
> we should have a prelexer phase that removes all `\r` from the source

Why can't we have a prelexer phase that:
1. first converts all `\r\n` to `\n`
2. then converts all `\r` to `\n`

and prints a warning? Wouldn't that achieve what I proposed in https://github.com/supercollider/supercollider/pull/7510#issuecomment-4597397496?

JordanHendersonMusic:
1 is fine and works well, it also solves the issue with source code display.

The issue with two is that all the editors disagree on whether \r is a newline.

So what should this code do?
```
// 1 +1; \r NuclearMissiles().launch;
'meow'.postln
```

2 would launch the missiles, which would agree with scide's display of them code. 

```
// 1 +1; 
NuclearMissiles().launch;
'meow'.postln
```

But everyone else, including here on GitHub, would see this

```
// 1 +1; ^M NuclearMissiles().launch;
'meow'.postln
```
And be confused as to why that happen.

Same holds for \v.

dyfer:
> The issue with two is that all the editors disagree on whether \r is a newline.

Sure. But currently sclang treats it as a newline, right?
In that case I suggest we keep that functionality, but add a warning. Eventually we can turn it into an error and not load a file with such character at all, but I'd do that after we post a warning about this for one version of SC.

prko:
I have many extensions and Quarks installed (though not all of them).
A VSCodium search of `~/Application Support/SuperCollider/` shows zero occurrences of `\f` or `\v`.

I understand that the goal of this PR is to eliminate ambiguous control characters (`\r`, `\v`, `\f`) because different editors render them inconsistently, and automatic conversion could change the meaning of code. I fully agree that these characters should not appear in source files.

Still, considering that many extensions, including Quarks, are no longer maintained, I wonder whether a **one‑time migration tool** (separate from the compiler) could help users clean up legacy files without silently changing semantics.

For example, when sclang boots and compiles the library, it could:

- **Step 1. Show information:**
   ```text
   Unsupported control characters (`\r|\f|\v`) were found in several files under ~/Application Support/SuperCollider/:
   - downloaded-quarks/atk-sc3/Classes/extPathName.sc
   - downloaded-quarks/BBCut/classes/cutprocs/statbal/statbal.sc
   - ...
   - ...
   ```

- **Step 2. Ask the user once whether to automatically convert all occurrences:**
   ```text
   Would you like to run a one‑time migration to replace them with \n?
   ```
This way, the user is not asked per file — the decision is made once, and the entire library is migrated in a single step.

When SC-IDE was introduced, sclang automatically converted old help files into schelp files.  
Perhaps a similar tool could ease the transition for users with legacy files, while keeping the compiler strict as proposed.

Here is the list of `\r(?!\n)` occurrences in `~/Application Support/SuperCollider`:
<details><summary>Details</summary>
<pre>
988 results - 21 files

downloaded-quarks/atk-sc3/Classes/extPathName.sc:
  31  //	to be both ergonomic and comprehensive, providing both classic and novel algorithms
  32: //	to creatively manipulate and synthesise complex Ambisonic soundfields.
//
  33  //	The tools are framed for the user to think in terms of the soundfield kernel. By

downloaded-quarks/BBCut/classes/cutprocs/statbal/statbal.sc:
  1: //This file is part of The BBCut Library. Copyright (C) 2001  Nick M.Collins distributed under the terms of the GNU General Public License full notice in file BBCutLibrary.help

//N.M.Collins 24/10/01 

//quick version - no normalisation step, but then avoids any controversy
//over what the normalisation step should be
//no worry over how many non zero weights

StatBalStream : Stream {

    var <>states,<>weights,<>heterogeneity,<>length,<counter,<stats,<>accent=1;

    *new { arg states, weights, heterogeneity=0.0, length=inf;  

        ^super.new.states_(states).weights_(weights).heterogeneity_(heterogeneity).length_(length).reset
    }
    
    next {  //v is expected increment
            var choice,best,temp,actual;
            
            ^(counter >= length).if {
                nil
            } {
                
                //stat balance procedure, no normalisation
                
                //create expected increments, choosing least as candidate
                choice = weights.minIndex { |val, i|
                    stats[i] + ((accent + (heterogeneity * 1.0.rand)) / val)
                } ?? { states.size.rand };
                
                //actual increment
                actual = accent / weights[choice];
                    
                //update statistics, no normalisation
                stats[choice] = stats[choice] + actual;
                    
                counter = counter + 1;  
                states[choice];
            };
    }
    
    reset {
        counter = 0;        
        stats = Array.fill(states.size,{0.0}); 
    }

}

downloaded-quarks/BBCut/classes/cutprocs/statbal/statbalnorm.sc:
  1: //This file is part of The BBCut Library. Copyright (C) 2001  Nick M.Collins distributed under the terms of the GNU General Public License full notice in file BBCutLibrary.help

 //created by Nick Collins 10/10/01 
//this is fully general version- 
//quick versions StatBalStream,Pstatbal with no normalisation, no zero weights! 
//my own normalisation step- discussed with Charles Ames- he no longer wants to think about it...

//possible improvement- search takes first occuring state if draw in expected increments- can improve this to random choice? 
//not without extra computational effort- also pointless in longterm unless hetero is zero

//specific streams class and patterns class

//if weight < 0.00001 taken as zero (floating point 0.0 must be tested thus)
//zero weighted states do not have their statistics changed in normalisation
//so may 'turn off' an option for a while

//assumes that states and weights arrays are the same size
//change weights, states, heterogeneity, length, accent whenever desired, but don't change size of weights array= size of states array

StatBalNormStream : Stream {
	var <>states,<>weights,<>heterogeneity,<>length,<counter,<stats,<>accent=1;

*new { arg states, weights, heterogeneity=0.0, length=inf;	

		^super.new.states_(states).weights_(weights).heterogeneity_(heterogeneity).length_(length).reset
	}
	
	next {	//v is expected increment
			var choice,best,numnonzero,temp,actual,deduct;
			
			^if (counter >= length, {
				nil
			},{
				
				//stat balance procedure
				
				//create expected increments, choosing least as candidate
				choice=states.size.rand;
				
				best=inf;
				
				numnonzero=0;
				
				weights.do(
				{
				arg val,i;
				
				if(val>0.00001,
				{
				temp= stats.at(i)+((accent+(heterogeneity*(1.0.rand)))/val);
				
				if(temp<best,{best=temp;choice=i;});
			
				numnonzero= numnonzero+1;
				}
				)
				
				}
				);
				
				
				if(numnonzero>0,{
				//actual increment
				actual= accent/weights.at(choice);
					
				
				//normalise, updating statistics
				
				//could take numnonzero-1, don't deduct from incremented stat
				deduct= actual/numnonzero;
				
				weights.do(
				{
				arg val,i;
				
				if(i==choice,
				{
				stats.put(i,stats.at(i)+actual-deduct)
				},
				{
				if(val>0.00001,
				{
				stats.put(i,stats.at(i)-deduct)
				}
				)
				
				}
				)
				
				}
				)
				
				}
				);
					
				counter = counter + 1;	
				states.at(choice)
				
			})	
	}
	
	reset {
	counter = 0;		
	stats= Array.fill(states.size,{0.0}); 
	}

}			

downloaded-quarks/BBCut/classes/cutprocs/statbal/statbalpattern.sc:
  1: //This file is part of The BBCut Library. Copyright (C) 2001  Nick M.Collins distributed under the terms of the GNU General Public License full notice in file BBCutLibrary.help

Pstatbal : Pattern {	
	var <>states,<>weights, <>heterogeneity,<>length;
	*new { arg states,weights,heterogeneity=0.0,length=inf;
		^super.new.states_(states).weights_(weights).heterogeneity_(heterogeneity).length_(length)
	}	
	asStream {	 
		^StatBalStream.new(states,weights,heterogeneity,length)
	}
}

Pstatbalnorm : Pattern {	
	var <>states,<>weights, <>heterogeneity,<>length;
	*new { arg s,w,het=0.0,len=inf;
		^super.new.states_(s).weights_(w).heterogeneity_(het).length_(len)
	}	
	asStream {	 
		^StatBalNormStream.new(states,weights,heterogeneity,length)
	}
} 

downloaded-quarks/crucial-library/Players/PlayerMixerGui.sc:
  1: 
PlayerMixerGui : AbstractPlayerGui {

	guiBody { arg layout;
		this.durationGui(layout);
		model.players.do({ arg pl;
			pl.gui(layout.startRow);	
		});
	}
	
}

downloaded-quarks/crucial-library/Players/SimpleTrigger.sc:
  1: 
// thanks to Maurizio for help working on this
SimpleTrigger : SynthlessPlayer {

	var <>action, <>spec, <value;

	*new { arg spec=\trig;
		^super.new.spec_(spec.asSpec).init
	}
	init {
		value = spec.maxval;
	}
	trig { arg val;
		action.value(this);
		value = val ? spec.maxval;
		this.changed(this);
	}

	synthArg { ^spec.minval }
	rate { ^\control }
	addToSynthDef {  arg synthDef, name;
		synthDef.addTr(name, this.synthArg);
	}
	instrArgFromControl { arg control; ^control }
	makePatchOut {
		patchOut = UpdatingScalarPatchOut(this, enabled: false);
	}
	didStop {
		super.didStop;
		patchOut.free;
		patchOut = nil;
	}

	guiClass { ^SimpleTriggerGui }
}

downloaded-quarks/crucial-library/Players/SimpleTriggerGui.sc:
  1: 
SimpleTriggerGui : AbstractPlayerGui {

	guiBody { arg layout;
		ActionButton(layout, model.name ? "trig", { model.trig;});
	}

}

downloaded-quarks/Ctk/Ctk classes/LP.sc:
    4  
    5: \header {
  tagline = ##f
  composer = ##f
}

\paper {
  paper-width = 7\cm
  paper-height = 7\cm
  top-margin = 0.1\cm
  left-margin = 0.1\cm

}
    6  */

  353  		this.addLyVar(\useSpatialNotation, 
  354: 			"useSpatialNotation = {
	\\override Score.SpacingSpanner #'uniform-stretching = ##t
	\\override Score.SpacingSpanner #'strict-note-spacing = ##t
  355  	\\set Score.proportionalNotationDuration = #(ly:make-moment 1 64)
  356  	\\override Staff.Glissando #'breakable = ##t
  357: 	}");
		this.addLyVar(\hideNote, 
  358: 			"hideNote = {\\once \\override NoteHead #'transparent = ##t
	\\once \\override NoteHead #'no-ledgers = ##t
	\\once \\override Stem #'transparent = ##t
	\\once \\override Beam #'transparent = ##t
  359: 	\\once \\override Accidental #'transparent = ##t
	}");
		this.addLyVar(\timeNotation, 
  360: 			"timeNotation = {\\once \\override Stem #'transparent = ##t
	\\once \\override Glissando #'thickness = 5
	\\once \\override Beam #'transparent = ##t
	}");
		}
  361  	}

  796  
  797: lowerLower = \new Voice = "lowerLower" {
  \voiceTwo
  \override Glissando #'thickness = 5 
  \override Stem #'transparent = ##t 
  \override Glissando #'bound-details #'right #'Y = #2
  a4 \glissando s s \once \override NoteHead #'transparent = ##t a4 s s bf4 \glissando c4
}
  798  

  800  
  801: \header {
  tagline = ##f
  composer = ##f
}

\paper {
  paper-width = 14\cm
  paper-height = 7\cm
  top-margin = 0.1\cm
  left-margin = 0.1\cm

}

\include "english.ly"

lowerLower = \new Voice = "lowerLower" {
  \voiceTwo
  \override Glissando #'thickness = 5 
  \override Stem #'transparent = ##t 
  \override Glissando #'bound-details #'right #'Y = #2
  a4 \glissando s s \once \override NoteHead #'transparent = ##t a4 s s 
  \once   \override Glissando #'bound-details #'right #'Y = #2.5
  bf4 \glissando 
  \once   \override Glissando #'bound-details #'right #'Y = #3
  c'4 \glissando s s 
  bf8
  a4 \glissando s s 
  \once \override NoteHead #'transparent = ##t a4
}
  802  

  804  
  805:   \once   \override Glissando #'bound-details #'right #'padding = #2.5   


\score {
	\new Staff = "bass" <<
	  \override Staff.TimeSignature #'transparent = ##t
	  \override Staff.BarLine #'transparent = ##t
	  \clef bass
	  \time 4/4
	  \lowerLower
	>>

}
  806  */

downloaded-quarks/DayTimer/DayTimer.sc:
  1: DayTimer { 
	classvar <all, <top;
	var <dates, <skip, <name;
	
	*new { |name, id, test, func| 
		var res = this.at(name); 
		if (id.notNil) { res.putDate(id, [test, func]) };
		^res;
	}
	
	*at { |name| 
		^all[name] ?? { ^super.new.init(name) };
	}
	
	init { |inname|
		name = inname;
		dates = ();
		all.put(name, this);
		skip = SkipJack({ this.checkList; }, 1, name: name, autostart: false);
	}
	
	*initClass { 
		all = ();
		top = this.new(\top);
	}
	
	*start { 	top.skip.start }
	*stop { 	top.skip.stop }
	
	*putDate { |id, test, func| top.putDate(id, test, func) }
	*removeAt { |id| ^top.removeAt(id) }

	start { 	skip.start }
	stop { 	skip.stop }

	putDate { |id, test, func| dates.put(id, [test, func]) }
	removeAt { |id| ^dates.removeAt(id) }
	
	checkList { 
		var date = Date.getDate;
		var timeNow = date.hour * 60 + date.minute * 60 + date.second;

		dates.keysValuesDo { |id, pair| 
			var test, func; #test, func = pair; 
			try { 
				if (test.isArray) { test = (test * [3600, 60, 1]).sum };
				if (test.isNumber) { 
					if (test == timeNow) { func.value(date, timeNow) }
				} { 
				//	"test is a function...".postln;
					if (test.value(date) == true) { func.value }
				}
			} { |err| 
				("DayTimer(\\" ++ name ++ ") failed to execute a function:").warn; 
				func.postcs;
				
				err.reportError;
			}
		};
	}
}

downloaded-quarks/ddwEQ/EQspecs.sc:
   1  
   2: MultiEQ {
	// container for specifications for an EQ bank
   3: 	// creates an individual synthdef for each band and allows placing on server
	// H. James Harkins, Jan 2003 - jamshark70@dewdrop-world.net
	// based on Joseph Anderson's Regalia-Mitra EQ classes
	// ftp://ftp.music.washington.edu/Public/jla/RM_Filters.sit
	
	classvar	numArgs = 4,	// must equal the number of eq parms: type, freq, k, rq
			<>sampleRate = 44100;
   4: 	
	var	<spec,	// List of EQBand
		<numChannels,
   5  		<target, <bus, <addAction, <mul = 1,

   7  		<>editor;
   8: 	
	*new {
   9: 			  // takes an argument list; .init calls .parseArgs to parse the list
		arg numChan = 1 ... z;
		
		^super.new.init(numChan, z)
	}
	
	init { arg numChan ... z;
  10: 		numChannels = numChan;
		spec = this.parseArgs(z);
	}
	
	parseArgs {  // turn argument list into array of parameter arrays
		arg ... z;	// argument list; may or may not be wrapped in an array
		
		var oneSpec, allSpecs;

		z = z.flat;	// remove all inner array wrapping

		(z.size == 0).if({
			^List.new		// empty list for no specs
		}, {
			allSpecs = List.new;		// initialize allSpecs
			
			z.do({ arg item;	// iterate over arguments
				item.isSymbol.if({	// a symbol indicates a new spec
  11  						// if oneSpec is nil, this is the first one so don't add
  12: 					oneSpec.isNil.if(nil,
						{ 		// else add it
  13  							allSpecs.add(

  16  							);
  17: 						}
					);
					oneSpec = List.with(item);	// then initialize new spec
				}, {
					oneSpec.add(item);		// other types, add to onespec
				});
			});
			
			^allSpecs.add(
  18  				EQBand.newFromArray(oneSpec, numChannels, sampleRate)
  19  					.parent_(this)
  20: 			);  // add the last spec
	
		});
	}
	
	add {		// add one or more filters
		arg ... z;		// same syntax for arglist as .new
		var newbands;
  21  		

  30  		spec = spec ++ newbands;
  31: 		
		^this
	}
  32  	

  36  		i.notNil.if({ this.removeAt(i, updateGUI) });
  37: 	}
	
	removeAt {		// remove a filter
		arg i, updateGUI;	// index to filter
  38  		var band;

  42  
  43: 		^this
	}
	
	at {		// index a filter; returns one band
		arg i;
		^spec.at(i)
	}
	
	play { arg targ, b, addAct = \addToTail, amp = 1;
  44  		target = targ;

  59  	
  60: 	asString { 
		var out;
		out = "MultiEQ.new(" ++ numChannels ++ ", ";   // initialize string
		spec.do( {		// loop through filters
			arg s, i;
			(i > 0).if({ out = out ++ ", " });  // skip comma for first eq
			out = out ++ "\\" ++ s.type
				++ ", " ++ s.freq.round(0.001)
				++ ", " ++ s.k.ampdb.round(0.001) ++ ".dbamp"
				++ ", " ++ s.rq.round(0.001);
		});
		^(out ++ ")")
	}
	
	edit { arg targ, b, addAct = \addToTail, amp = 1;	// just like .play
  61  			// assumes your synth or process is already playing

downloaded-quarks/GamelanSynthesis/Classes/VGSamp.sc:
  155  		Task { 
  156: 			var stillEmpties; 
	
			0.2.wait; 
  157  
  158: 			bufenv !? { bufenv.clear(true) };
			bufenv = BufEnvir(server);
			
  159  			

downloaded-quarks/MathLib/classes/LaTeX/LaTeX.sc:
  112  
  113: \begin{tabular}{l l l l }  \\  
$0$	&$1$	&$2$	&$3$  \\  
  \hline  $4$	&$5$ 	&$6$	&$7$  \\  
  114: \cline{1-1} \cline{3-4}
$8$	&$9$	&$10$	&$11$  \\  
  \hline  $12$	&$13$	&$14$	&$15$  \\  
\end{tabular}
  115  */

downloaded-quarks/MathLib/classes/MarkovSet/FuzzyDictionary.sc:
   4  
   5: FuzzyDictionary : IdentityDictionary {

	put { arg key, item;
		super.put(key, super.at(key).add(item))
	}
	
	at { arg key;
		var list = super.at(key);
		list !? { ^list.choose };
   6: 		^nil
	}
	
	removeAt { arg key, item;
		var list = super.at(key);
   7  		list !? { 

  14  				^item 
  15: 		};
		^nil
	}

	choose {
		var list = super.choose;
		list !? { ^list.choose };
  16: 		^nil
	}
  17  	

  39  	}
  40: 	*/
	
	doAt { arg key, function;
		super.at(key).do(function);
	}
	
	keyAt { arg key;
		^super.at(key)
	}
  41  	

  44  		super.put(key, item)
  45: 	}
	
}
  46  

downloaded-quarks/Morse/MorseDict.sc:
   1: 		// based on hairi vogel's morse class
MorseDict {
   2: 	
classvar <codes, <strings, <letters;
   3: 
	*initClass {
   4  

   7  		codes =  Event[ 	
   8: 			$a -> [0,1],
			$b -> [1,0,0,0],
			$c -> [1,0,1,0],
			$d -> [1,0,0],
			$e -> [0],
			$f -> [0,0,1,0],
			$g -> [1,1,0],
			$h -> [0,0,0,0],
			$i -> [0,0],
			$j -> [0,1,1,1],
			$k -> [1,0,1],
			$l -> [0,1,0,0],
			$m -> [1,1],
			$n -> [1,0],
			$o -> [1,1,1],
			$p -> [0,1,1,0],
			$q -> [1,1,0,1],
			$r -> [0,1,0],
			$s -> [0,0,0],
			$t -> [1],
			$u -> [0,0,1],
			$v -> [0,0,0,1],
			$w -> [0,1,1],
			$x -> [1,0,0,1],
			$y -> [1,0,1,1],
			
   9: 			$z -> [1,1,0,0],
			$0 -> [1,1,1,1,1],
			$1 -> [0,1,1,1,1],
			$2 -> [0,0,1,1,1],
			$3 -> [0,0,0,1,1],
			$4 -> [0,0,0,0,1],
			$5 -> [0,0,0,0,0],
			$6 -> [1,0,0,0,0],
			$7 -> [1,1,0,0,0],
			$8 -> [1,1,1,0,0],
			$9 -> [1,1,1,1,0],
  10: 			
			$. -> [0,1,0,1,0,1],
			$, -> [1,1,0,0,1,1],
			$: -> [1,1,1,0,0,0],
  11: 			$? -> [0,0,1,1,0,0],
			$' -> [0,0,1,1,0,0],
  12  			$- -> [1,0,0,0,0,1],

  30  
  31: 			$  -> [2]				// stop - is that so? 
		];	
  32  		

  36  			letters.put(this.signs(key).asSymbol, key);
  37: 		}
	}
  38  	

  49  	*at { arg code; 
  50: 		 if (code.isKindOf(Symbol).not) { code = code.asAscii.toLower };
 			^codes.at(code) ? [];
	}
  51  	

  53  	
  54: 	*postKeys { this.keys.printcsAll.postcs }
}

downloaded-quarks/NetLib/classes/extStringCleaning.sc:
  1: + Char {
	isClean {
		^this.isPrint or: { this.ascii.inclusivelyBetween(9, 13) }
	}
}

+ String {

	countGremlins { ^this.findGremlins.size }

	findGremlins {
		var indices = [];
		this.do { |ch, i|
			if (ch.isClean.not) { indices = indices.add(i) }
		};
		^indices
	}

	clean { |space=true|	// strip all non-crossplatform ascii chars.
		^if (space)		// either add blankspaces
			{ this.collect { |ch| if (ch.isClean) { ch } { $  }; } }
						// or not.
			{ this.select(_.isClean) }
	}
}

+ Document {
	clean { |space=true|
		var cleanedString;
		cleanedString = this.string.clean(space);
		if (cleanedString != this.string, {
			this.string = cleanedString;
			"" ++ this.class + this.title + "swapped for clean string.".postln;
		});
	}
	*makeCleanerWindow {
		var w;
		w = SCWindow("cleaner", Rect(0,0,140, 60)).front;
		w.view.decorator = FlowLayout(w.bounds.copy.left_(0).top_(0));
		SCButton(w, Rect(0,0,120,20))
			.states_([["clean current doc"]])
			.action_ { Document.current.clean(false) };
		SCButton(w, Rect(0,0,120,20))
			.states_([["current doc gremlins"]])
			.action_ { var gr; gr = Document.current.string.findGremlins;
				if (gr.notEmpty,
					{ Document.current.selectRange(gr[0], 1).front },
					{ Document.current.selectRange(nil, 0).front; "None!".postln; }
				);
			};
	}
}

+ Class {
	*showGremlins {
		var scfiles = Set.new;
		var file, string, badIndices, count;
		Class.allClasses.do { |cl| scfiles.add(cl.filenameSymbol); };

		scfiles.as(Array).do { |name|
			file = File(name.asString, "r");
			if (file.isOpen, {
				badIndices = file.contents.findGremlins;
				count = badIndices.size;
				if (count > 0, { ("" + name + ":" + count + "at:" + badIndices).postln });
				file.close;
			}, {
				("could not open file: " + name).postln;
			});
		};
	}
}

Extensions/SC3plugins/JoshUGens/classes/AudioMSG.sc:
  1: AudioMSG : UGen {
	*ar { arg in, index=0pi, mul = 1.0, add=0.0;
//		^SinOsc.ar( 0, ( in*0.5pi ) + SinOsc.ar( 0, in*2pi, index ), mul, add )
  2  		^this.multiNew('audio', in, index).madd(mul, add);
  3: 	}
}
  4  

Extensions/SC3plugins/JoshUGens/classes/joshfilts.sc:
  1: ZLPF {
	*ar { arg in, fc=440, damping=0.707, mul=1, add=0;

		var b0, b1, b2, a0, a1, a2;
		var c, cc, twodC;
		var out;
		
		out = in;
		
		damping.asArray.do({ arg df;
			
			c = (pi * fc/SampleRate.ir).tan.reciprocal;
			cc = c.squared;
			twodC = 2 * df * c;
	
			b0 = (1 + twodC + cc).reciprocal;
			b1 = 2 * b0;
			b2 = b0;
	
			a0 = 1;
			a1 = 2 * b0 * (1 - cc);
			a2 = b0 * (1 - twodC + cc);
	
			out = SOS.ar( out, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )
		});
		^out
	}
}


ZHPF {
	*ar { arg in, fc=440, damping=0.707, mul=1, add=0;

		var b0, b1, b2, a0, a1, a2;
		var c, cc, twodC;
		var out;
		
		out = in;
		
		damping.asArray.do({ arg df;
			
			c = (pi * fc/SampleRate.ir).tan;
			cc = c.squared;
			twodC = 2 * df * c;
	
			b0 = (1 + twodC + cc).reciprocal;
			b1 = -2 * b0;
			b2 = b0;
	
			a0 = 1;
			a1 = 2 * b0 * (cc - 1);
			a2 = b0 * (1 - twodC + cc);
	
			out = SOS.ar( out, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )
		});
		^out
	}
}


BLPF {
	*ar { arg in, order=4, freq=440, mul=1, add=0;
		var df, out;
		order = order.asInteger;
		df = Dictionary[
			2->2.sqrt.reciprocal,
			4->[0.924, 0.383],
			6->[0.966, 0.707, 0.259],
			8->[0.981, 0.831, 0.556, 0.195],
			10->[0.988, 0.891, 0.707, 0.454, 0.156]
		].at(order);
		df.isNil.if({^(this.asString+"invalid order:" +order).error});
		out = (df.asArray.at(0) > 0).if({ZLPF.ar( in, freq, df)},{in});
		^out * mul + add
	
	}
}

BHPF {
	*ar { arg in, order=4, freq=440, mul=1, add=0;
		var df, out;
		order = order.asInteger;
		df = Dictionary[
			2->2.sqrt.reciprocal,
			4->[0.924, 0.383],
			6->[0.966, 0.707, 0.259],
			8->[0.981, 0.831, 0.556, 0.195],
			10->[0.988, 0.891, 0.707, 0.454, 0.156]
		].at(order);
		df.isNil.if({^(this.asString+"invalid order:" +order).error});
		out = (df.asArray.at(0) > 0).if({ZHPF.ar( in, freq, df)},{in});
		^out * mul + add
	
	}
}


LRLPF {
	*ar { arg in, order=4, freq=440, mul=1, add=0;
		var df, out;
		out = in;
		2.do({out = BLPF.ar( out, order/2, freq)})
		^out * mul + add
	}
}


LRHPF {
	*ar { arg in, order=4, freq=440, mul=1, add=0;
		var df, out;
		out = in;
		2.do({out = BHPF.ar( out, order/2, freq)});
		^out * mul + add
	
	}
}

Extensions/SC3plugins/JoshUGens/classes/Phaser.sc:
  1: Phaser {

	*help { arg method, argument;
		^Switch(method.asSymbol)
			.case(\note, {"\nWritten by CK 1/3/02. first version"})
			.case(\short_desc, {"Phaser with phase argument"})
			.case(\long_desc, {"Sawtooth oscillator. Output ranges from (0 to mul starting at phase) + add."})
			.caseIncludes([\ar,\kr],{
				Switch(argument.asSymbol)
					.case( 'freq', {"frequency in Hertz"} )
					.case( 'phase', {"starting phase (between 0 and mul)"} )
					.case( 'mul', {"maximum value"} )
					.result
			})
			.case(\examples, {"\nEXAMPLES:\n"
"\n(\nSynth.plot({ Phaser.ar( 500, 0.5, 1 ) })\n)"			
"\n\n(\nSynth.plot({ Phaser.ar( 500, 1, 2 ).fold2(1) })\n)"			
			})
			.case(\nil,{HELP(this)})
			.resultStr
	}
	
	*kr { arg freq = 440.0, phase=0, mul = 1.0, add = 0.0;
		var halfmul;
		halfmul = mul/2;
		^LFSaw.kr( freq, halfmul, mul + phase ).wrap2(halfmul) + halfmul + add;
	}
	*ar { arg freq = 440.0, phase=0, mul = 1.0, add = 0.0;
		var halfmul;
		halfmul = mul/2;
		^LFSaw.ar( freq, halfmul, mul + phase ).wrap2(halfmul) + halfmul + add;
	}
}

Extensions/SC3plugins/JoshUGens/classes/Vocoder.sc:
  1: // car is "pitch you hear" (like a saw wave, or sound), mod is voice, num is number of bands
// to be spaced over the threshhold, , hpf and scale add back some of mod
Vocoder {
	
	
	
	*ar { 	arg car, mod, num, low=100, high=5000, q=0.02, hpf=5000, hpfscal=0.05, outscal=25;
	
			var width, cf, hf,  out, filtmod, filtcar, tracker, ratio;
	
	
		out = Mix.arFill(( num + 1 ), { arg i; 
					
					ratio = (( high / low)**num.reciprocal );
	
				 	cf =  ( ratio**i) * low;
				 	
					filtmod = BPF.ar( mod, cf, q);
				
					tracker = Amplitude.kr(filtmod);
				
					filtcar = BPF.ar( car, cf, q);	
				
	 
					( outscal * ( filtcar * tracker ));
					});
  2  
  3  		hf = HPF.ar(HPF.ar( mod, hpf), hpf);
  4: 
		
		^out + ( hpfscal * hf )}
  5  

Extensions/SC3plugins/JoshUGens/classes/Class Files/BiquadEQ.sc:
   1  /*
   2: RMEQ {
/*
Dattoro
Effect Design
Part 1: Reverberator and Other Filters
JAES, vol 45, no 9, 1997

	Synth.fftScope({
		n = WhiteNoise.ar(0.1);
		f = MouseX.kr( 0.1, 24 );
		b = MouseY.kr( -12, 12 ) * 2;
		[
			RegaliaMitraEQ.ar( n, 5000, f.midiratio - 1, b),
			RBJPeakingEQ.ar( n, 5000, f.midiratio - 1, b )
		]
	})
	
	Synth.fftScope({RegaliaMitraEQ.ar( WhiteNoise.ar(-20.dbamp), MouseY.kr(100, 1000, \exponential), 12.midirq, MouseX.kr(-20, 20))})	
	scope({RegaliaMitraEQ.ar( WhiteNoise.ar(-60.dbamp), LFNoise2.kr(1, 50, 300), LFNoise2.kr( 5, 6, 15).midiratio - 1, LFNoise1.kr( 2.1, 12, 30))})	

*/
	*ar { arg in, freq=440, rq=0.1, dbgain=0;
		
		var wc, w2, w1, dw, q, beta, sr, y;
		var b0, b1, b2, a0, a1, a2, az, k;
		
		k = dbgain.dbamp;
				
		sr = 2pi / SampleRate.ir;
		wc = sr * freq;				// freq in radians
		dw = rq * wc;				// bandwidth in radians

		dw = (dw/2).tan;			// delta is only used to compute �
		beta = (1 - dw) / (1 + dw);	// �
		y = cos(wc).neg;			// lattice coefficient
		
		b0 = beta;
		b1 = y * (1 + beta);
		b2 = 1;
		
		a0 = 1;
		a1 = b1;	// y * (1 + beta);
		a2 = beta;
		
//		az = SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg );
		az = SOS.ar( in, b0, b1, b2, a1.neg, a2.neg );
	
		^( k * (in - az) ) + (in + az) * 0.5
	}
}
   3  */
   4: 
/*
Implementation of filters as described at:
<http://www.harmony-central.com/Computer/Programming/Audio-EQ-Cookbook.txt>
*/
/*
PeakingEQ {
	/*
		scope({RBJPeakingEQ.ar( WhiteNoise.ar(-60.dbamp), MouseY.kr(100, 1000, \exponential), 12.midiratio - 1, MouseX.kr(0, 45))})	
		scope({RBJPeakingEQ.ar( WhiteNoise.ar(-60.dbamp), LFNoise2.kr(1, 50, 300), LFNoise2.kr( 5, 6, 15).midiratio - 1, LFNoise1.kr( 2.1, 12, 30))})	
	*/

	*ar { arg in, freq=440, rq=0.1, dbgain=0, mul=1, add=0;
		var b0, b1, b2, a0, a1, a2;
		var a, omega, sin, cos, alpha, a0n;
		var aovera, atimesa;

		a = dbgain.dbamp;	
		omega = 2pi * freq/SampleRate.ir;
		sin = omega.sin;
		cos = omega.cos;
		alpha = sin * rq / 2;

		aovera = (alpha/a);
		atimesa = (alpha*a);
		
		a0 = 1 + aovera;
		a1 = -2 * cos;
		a2 = 1 - aovera;

		b0 = 1 + atimesa;
		b1 = a1;
		b2 = 1 - atimesa;
		
		a0n = a0.neg;

		^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0n, a2/a0n, mul, add )
	}

}
*/
Notch  {
	/*
		Synth.fftScope({RBJNotch.ar( WhiteNoise.ar(0.1), 4400, MouseX.kr(0.01, 2))})	
		Synth.fftScope({RBJNotch.ar( SinOsc.ar(LFTri.kr(0.1, 1000, 1100), 0, 0.1), 500, MouseX.kr(0.01, 2))})	
	*/

	*ar { arg in, freq=440, rq=0.1, mul=1, add=0;

		var b0, b1, b2, a0, a1, a2;
		var a, omega, sin, cos, alpha, beta, q;
		
		omega = 2pi * freq/SampleRate.ir;
		sin = omega.sin;
		cos = omega.cos;
		alpha = rq * sin/2;


		a0 = 1 + alpha;
		a1 = -2 * cos;
		a2 = 1 - alpha;

		b0 = 1;
		b1 = a1;	// -2 * cos;
		b2 = 1;

		^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )
	}
}
   5  
   6: LowShelf  {
	/*
		Synth.fftScope({RBJLowShelf.ar( WhiteNoise.ar(0.1), 400, 1, MouseX.kr(-24, 24))})	
	*/

	*ar { arg in, freq=440, shelfslope=1, dbgain=0, mul=1, add=0;
		var am1, ap1, bs, am1cos, ap1cos, q;

		var b0, b1, b2, a0, a1, a2;
		var a, omega, sin, cos, alpha, beta;
		
		a = dbgain.dbamp;	
		omega = 2pi * freq/SampleRate.ir;
		sin = omega.sin;
		cos = omega.cos;

		beta = a.sqrt * ((a+a.reciprocal) * (shelfslope.reciprocal - 1) + 2).sqrt;
//		beta = (((a.squared + 1) / shelfslope) - (a - 1).squared).sqrt;
//		q = ((a + a.reciprocal) * (shelfslope.reciprocal - 1) + 2).sqrt.reciprocal;
//		q.postln;
//		beta = a.sqrt / q;

		am1 = a-1;
		ap1 = a+1;
		am1cos = am1 * cos;
		ap1cos = ap1 * cos;
		bs = beta * sin;
		
		a0 = ap1 + am1cos + bs;
		a1 = -2 * ( am1 + ap1cos );
		a2 = ap1 + am1cos - bs;

		b0 = a * ( ap1 - am1cos + bs );
		b1 = 2 * a * ( am1 - ap1cos );
		b2 = a * ( ap1 - am1cos - bs );

		^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0.neg, a2/a0.neg, mul, add )
	}
}
   7: 
HighShelf  {
	/*
		Synth.fftScope({RBJHighShelf.ar( WhiteNoise.ar(0.1), 11000, 1, MouseX.kr(-6, 6))})	
	*/

	*ar { arg in, freq=440, shelfslope=1, dbgain=0, mul=1, add=0;
		var am1, ap1, bs, am1cos, ap1cos;

		var b0, b1, b2, a0, a1, a2, a0n;
		var a, omega, sin, cos, alpha, beta, q;
		
		a = dbgain.dbamp;	
		omega = 2pi * freq/SampleRate.ir;
		sin = omega.sin;
		cos = omega.cos;

//		beta = (((a.squared + 1) / shelfslope) - (a - 1).squared).sqrt;
		beta = a.sqrt * ((a+a.reciprocal)*(shelfslope.reciprocal-1)+2).sqrt;
//		q = ((a + a.reciprocal) * (shelfslope.reciprocal - 1) + 2).sqrt.reciprocal;
//		beta = a.sqrt / q;

		am1 = a-1;
		ap1 = a+1;
		am1cos = am1 * cos;
		ap1cos = ap1 * cos;
		bs = beta * sin;
		
		a0 = ap1 - am1cos + bs;
		a1 = 2 * ( am1 - ap1cos );
		a2 = ap1 - am1cos - bs;

		b0 = a * ( ap1 + am1cos + bs );
		b1 = -2 * a * ( am1 + ap1cos );
		b2 = a * ( ap1 + am1cos - bs );
		
		a0n = a0.neg;

		^SOS.ar( in, b0/a0, b1/a0, b2/a0, a1/a0n, a2/a0n, mul, add )
	}
}
   8  

  11  //RBJPeakingEQ : PeakingEQ {}
  12: RBJNotch : Notch {}
RBJHighShelf : HighShelf {}
  13  RBJLowShelf : LowShelf {}

Extensions/SC3plugins/JoshUGens/classes/Class Files/CheapVerb.sc:
  1: CheapVerb {
	*ar { arg in, decayTime=1, numDelays=6, mul=1.0, add=0.0;
		var out, numCh, maxDelayTime=0.05;
		out = in;
		numCh = in.size;
		numDelays.do({ 
			out = AllpassN.ar( 
				out, 
				maxDelayTime, 
				(numCh > 0).if({
						// if in is multi channel, make seperate delay time for each channel
						Array.fill( numCh, { arg i; maxDelayTime.rand })
					},{
						maxDelayTime.rand
				}), 
				decayTime
			) 
		});
		^out * mul + add
	}
}
</pre>
</details>

prko:
When converting \r, we should also consider that multiple empty lines may appear consecutively:
https://github.com/supercollider/supercollider/pull/7538/changes/644e345cb283f6c2a8f1c3d77cab903ec07e9022

JordanHendersonMusic:
@dyfer if we have to run the conversion to load the file in the lexer you might as well get the user to convert the file there and then? Particularly if it is very simple.


The down side with both approaches is that not only newlines change, but so does the content of strings that contain bare literal crs. But we will also need to convert crlf into lf so this shouldn't be too much of a problem unless that code specifically relies on checking for \r or \r\n

Also, on windows, some strings on some editors will change size \r\n to \n. That could break something - but that code would have only worked on windows anyway so I don't think it is too much of an issue.

dyfer:
> @dyfer if we have to run the conversion to load the file in the lexer you might as well get the user to convert the file there and then? Particularly if it is very simple.

Uhm... I don't think so, at least not in all cases.
If the file is in a version-controlled repository that the user has no/limited control over (i.e. a Quark), then changing the contents and leaving the uncommitted changes leaves it in a "broken" state (from the user perspective - they can't update the quark etc).

The strings changing size is definitely an issue, but I don't think there's an ideal "clean" way to proceed. IMO we should make a warning and accept that certain features may be broken (like locating where exactly are we in the code or what are the sizes of strings).

--- 

Ultimately I want to avoid the scenario like this: "I updated SuperCollider and now it doesn't start". This is where we are right now (with the recent changes in develop) and it's too much friction, I think.
IMO we should - at least for one minor version - accept all line feed characters that were accepted previously, and print a warning. _Maybe_ also provide a conversion option.... but how would that even work e.g. in sclang in the command line? I think posting is the best we can do. And maybe add a conversion option in the IDE (or the conversion happens implicitly when opening the file and saving it... thought that also can pose problems).

JordanHendersonMusic:
I will submit prs to all quarks to fix this. The user may have to disable all quarks and reinstall them after updating - unless they updated before.

There are such a small number of files this is completely reasonable to do.

> The strings changing size is definitely an issue, but I don't think there's an ideal "clean" way to proceed. IMO we should make a warning and accept that certain features may be broken (like locating where exactly are we in the code or what are the sizes of strings).

Hard disagree, right now the content of the code depends on the *author's* os and editor, and *your* editor. That's ridiculous. Its a bug.

---

Okay I'll just do a conversion in a prelexing stage and post a warning. Got to do this anyway to remove the crlf.

prko:
>I will submit prs to all quarks to fix this. The user may have to disable all quarks and reinstall them after updating - unless they updated before.

This approach would work only if all quark authors respond within a couple of months, but in my experience many of them don’t — some haven’t responded for years. A more realistic solution would be to add an automatic conversion mechanism to sclang itself, so that these cases are detected and handled automatically.

Submitting PRs to all quarks to fix this is a huge amount of work, and I really admire your dedication to doing it the orthodox way. I just worry that the outcome may not match the effort, given how many quarks are effectively unmaintained.

dyfer:
> I will submit prs to all quarks to fix this.

I admire the effort, but I don't think this is possible :) There are quarks that are not listed in the quarks [directory](https://github.com/supercollider-quarks/quarks/blob/master/directory.txt) - you can install a quark from any git url. While maybe there aren't as many out there as e.g. vst plugins, it is an open ecosystem.

> > The strings changing size is definitely an issue, but I don't think there's an ideal "clean" way to proceed. IMO we should make a warning and accept that certain features may be broken (like locating where exactly are we in the code or what are the sizes of strings).
> 
> Hard disagree, right now the content of the code depends on the _author's_ os and editor, and _your_ editor. That's ridiculous. Its a bug.

I agree that's ridiculous and that it's a bug. 
But I would like to find a solution that doesn't prevent SC from starting, based on the installed extensions, _at least in the interim_. Is it possible? I'm not sure if we are discussing technical limitations or conceptual disagreement.

JordanHendersonMusic:
Right but it's only quarks written on classic Mac os, so before 2001 or so. When were quarks released as a feature?

> prevent SC from starting

That's a little different issue. The \r is currently preventing them from starting, I'm just going to convert internally from \r and \r\n to \n, that solves that. I'll change this pr to do that sometime this week.

The issue with the newline in the string won't stop SC from booting, and it only affects class library code authored on windows not checked into a git repo (as git converts).

Sorry if I'm a bit touchy about this. But I'm planning to implement good errors, a call stack debugger (like a step debugger in vscode) and an actual step debugger... But we can't do that unless we know where the code is located, and a part of that is agreeing on what a string is, and what a newline is, which requires the editor to also be in agreement.

dyfer:
> That's a little different issue. The \r is currently preventing them from starting, I'm just going to convert internally from \r and \r\n to \n, that solves that. I'll change this pr to do that sometime this week.

Okay! This was my primary concern. If the library compiles, that's great.

> The issue with the newline in the string won't stop SC from booting, and it only affects class library code authored on windows not checked into a git repo (as git converts).

So the library compiles, but if there's a method etc that tries to load a string with a wrong newline, that would be an error, correct?

JordanHendersonMusic:
> So the library compiles, but if there's a method etc that tries to load a string with a wrong newline, that would be an error, correct?

No, I think it is best to auto convert to \n. Emitting an error would be annoying because you can't see the problem, we also don't have file encoding options for scide. If you have notepad++ there is an option to show the newline types. And when you press enter on windows you always get crlf so it would be hard to change.

This way, we define that strings only contain \n as their newline character and can convert from any newline format. We already do this for all runtime code authored on scide and all code in the class library as git converts to \n I believe (there are some settings somewhere).

The issue comes if someone has relied on the fact a string might use \r (very unlikely because it would have been written around 2002) or the size of a string when you get the crlf (in class library not in version control on windows). Both these seem unlikely to occur.

In a different pr I also want to forbid all the other ASCII stuff you can't see (except \n, ' ', and \t) and introduce a new escape sequence that lets you type them - modeled after perl regex - `"meow \x0D"` which turns into the byte 0D. Since \x isn't used (compiler just turns it into x) this shouldn't be a breaking change... Unless someone has erroneously written \x instead of x in their code, but we can omit a compile time warning->error if the conversion isn't possible.

Issue #7509, opened 2026-05-22
Compiling *.sc files with unicode symbols or bare ^M carriage returns fails


sadguitarius:
## Environment

* SuperCollider version: main branch build
* Operating system: Windows
* Other details (Qt version, audio driver, etc.): Qt 6.11.1

## Steps to reproduce

I found this when building sc3-plugins against the main branch. Currently, the '§' characters in BiquadEQ.sc, the '©' character in tuningScale.sc and the numerous bare ^M style carriage returns in JoshUGens cause sclang to hang when compiling extensions directory.

I have a fork of sc3-plugins with the offending characters removed/replaced ([here](https://github.com/sadguitarius/sc3-plugins/tree/unicode-fix)) and sclang boots up fine. At least in the case of the bare carriage returns, this is also a bug in sc3-plugins, but I'd imagine there should be some kind of error raised in sclang instead of it just hanging completely.

JordanHendersonMusic:
Is the the dev branch or the 'main' branch? If it's dev, I've broken something... Although it is actually a good thing, and was going to make a pr to do exactly this.

JordanHendersonMusic:
Ah I see it is dev!

JordanHendersonMusic:
Okay so the unicode is a misnomer here as you do get a proper error (this will be improved in a future PR), the examples given are in comments however, which all work as expected.

```
// delta is only used to compute §
// © 2003 Lance Putnam
```
This is all fine.

The issue is the carriage returns.

This is actually inevitable as the old lexer didn't care about white space, whereas the new one does — this is to enable the creation of a formatter down the line. This means you need to be able to count the new line characters. On windows this is a both '\r\n', on unix and mac > osx its a '\n', but on mac pre osx it is just a '\r'. This means it isn't possible to count newline without knowing what operating system the **author** originally wrote the code in... I'd going to put forward a pr that discards all '\r'. Therefore the old code does need to be updated.

Additionally, we have the same issue with a vertical tab '\v' (this is in the vocoder file). I'm classifying this as a newline character. Previously we didn't have a definition of whitespace (never mind a newline), but it was discarded (like whitespace) if it was at the beginning of a loop, but you couldn't use it to finish a one line comment or a symbol — very confusing an inconsistent! 

Note: github can't display this code,  sc ide (aka qt)  and vscode both struggle. It is time to update!

sadguitarius:
[...]

PR #7513, opened 2026-05-25
sclexer: fix console encoding for Windows in lexer standalone

sadguitarius:
## Purpose and Motivation

Partial fix for #7509. On Windows, the console defaults to 16-bit wide char encoding, so it's necessary to convert stdin to UTF-8 before passing to the lexer. The console output codepage must also be changed to UTF-8 to properly display all characters.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/langutils/sc_lexer/standalone/main.cpp
+++ b/langutils/sc_lexer/standalone/main.cpp
@@ -23,9 +23,9 @@ template <typename Stream> void print(Stream& s, SourceCodeRange r) {
 std::string utf8_encode(const std::wstring& wstr) {
     if (wstr.empty())
         return std::string();
-    int size_needed = WideCharToMultiByte(CP_UTF8, 0, &wstr[0], static_cast<int>(wstr.size()), NULL, 0, NULL, NULL);
+    int size_needed = WideCharToMultiByte(CP_UTF8, 0, wstr.data(), static_cast<int>(wstr.size()), NULL, 0, NULL, NULL);
     std::string strTo(static_cast<size_t>(size_needed), 0);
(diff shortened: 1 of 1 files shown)


langutils/sc_lexer/standalone/main.cpp
```diff
@@ -14,17 +17,43 @@ template <typename Stream> void print(Stream& s, SourceCodeRange r) {
       << r.end.column << "]}";
 }
 
+#ifdef _MSC_VER
+// utility function to convert default Windows stdin encoding to UTF-8
+// https://stackoverflow.com/questions/215963/how-do-you-properly-use-widechartomultibyte
+std::string utf8_encode(const std::wstring& wstr) {
+    if (wstr.empty())
+        return std::string();
+    int size_needed = WideCharToMultiByte(CP_UTF8, 0, &wstr[0], static_cast<int>(wstr.size()), NULL, 0, NULL, NULL);
```
JordanHendersonMusic: Rather than `&wstr[0]` it should be `wstr.data()` same with all the places where you get a pointer to the first element.

JordanHendersonMusic (changes requested):
Thanks for this! There's just one little API thing.

langutils/sc_lexer/standalone/main.cpp
sadguitarius: Oops, I'd just grabbed that function wholesale from stackoverflow. Fixed!

JordanHendersonMusic (approved):

PR #7519, opened 2026-05-28
Update fftw and libsndfile cmake modules

dyfer:
## Purpose and Motivation

I'd like to update the scripts that search for fftw and libsndfile. This is mostly useful for Windows and would allow fftw and sndfile to be found in either `Program Files` or next to the SC source, including the current implementation.
I think this is a straightforward change, or rather a cleanup, aiming to simplify build setup on Windows. At the same time, there might be an edge case when cross compiling SC (eg. for x86 or arm64 on the x86_64 host) where a library for the wrong architecture would be discovered first. Still, I think this is a remote possibility.

The main improvement here is that the folders containing the libraries can now have a version suffix (e.g. `libsndfile-1.2.2`) and the library will still be found. 

The PR removes from the libsndfile search paths the "Mega-Nerd" locations which were used by the older libsndfile Windows installer from 2017 (v.1.0.28).

This PR also enables linking to fftw's .dll library under mingw.

These changes are extracted from #7493 for easier review. 

I'm separately working on a Windows readme update to update instructions for building on Windows, so the documentation changes are not included in this PR.

## Types of changes

- New feature (sort of)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation - forthcoming in a separate PR

### DIFF
--- a/cmake_modules/FindFFTW3f.cmake
+++ b/cmake_modules/FindFFTW3f.cmake
@@ -1,33 +1,51 @@
 # Find single-precision (float) version of FFTW3
 
-FIND_PATH(
+set(FFTW_HINT_PATHS_WIN "")
+if(WIN32)
+  file(GLOB FFTW_HINT_PATHS_WIN 
+    "$ENV{ProgramFiles}/fftw*"
+    "$ENV{ProgramFiles\(x86\)}/fftw*"
+    "$ENV{ProgramW6432}/fftw*"
+    "${CMAKE_SOURCE_DIR}/../${CMAKE_LIBRARY_ARCHITECTURE}/fftw*"
+    "${CMAKE_SOURCE_DIR}/../fftw*"
+  )
+endif()
+
+if(MINGW) # enable linking to dlls directly on mingw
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic:
> or next to the SC source,

Why would we want that?

dyfer:
> Why would we want that?

This is where the libraries were supposed to go according to the [Windows README](https://github.com/supercollider/supercollider/blob/develop/README_WINDOWS.md#arranging-the-components) for quite some time.

IIUC, there's no standard location for libraries on Windows, so putting them next to SC source makes as much or as little sense as putting them in `c:\Program Files` IMO.

I also updated the PR description which clarifies the additions.

----

I think it's time we suggest using `vcpkg` for all Windows dependencies (that's the easiest way to get a functional Readline on Windows, AFAIU), but maybe that's a discussion for the upcoming Windows readme rewrite PR which I'm also working on.

JordanHendersonMusic (approved):

JordanHendersonMusic:
Thanks that makes sense, feel free to merge if your happy then!

PR #7525, opened 2026-05-29
Refactor Post Window return output: replace `-> `, fix SC‑IDE colour, and improve return markers

### LINKED ISSUE #7524: Inconsistent Post window return highlighting
## Motivation
While reading #7520, the following behaviour of the SC-IDE came to mind:
Multi-line return values in the Post window highlight only the first line:
   - Example 1:
      ```supercollider
      (
      'line one\
      line two\
      line three'
      .postcs
      )
      ```
      returns:
      Image
      
   - Example 2:
      ```supercollider
      (
      "line one\
      line two\
      line three".postcs
      )
      ```
      returns:
      Image

I am accustomed to these behaviours, but it would be preferable if the highlighting were consistent.

it would be preferable if all lines belonging to the final printed result in the Post window were highlighted, rather than only the first line.

## Plan for Implementation
To my knowledge, this could be addressed by modifying the relevant primitives, which is currently beyond my ability.

prko:
fixes: #7524
## Purpose and Motivation
Currently, the multi‑line output in the Post window displays the green colour only for the first line after `->`, as shown below:
Screenshot 2026-05-29 at 14 50 21 copy

This pull request changes that behaviour so that all subsequent lines belonging to the same result block are displayed in green:
Screenshot 2026-05-29 at 14 44 35 copy
 test code:
```supercollider
(
"line one\
line two\
line three".postcs
)

(
'line one\
line two\
line three'
.postcs
)

("1
2
3
4
5
"
)
```
----

I would like to submit this pull request with full respect for the SuperCollider Code of Conduct and with appreciation for the concerns raised regarding the use of AI‑assisted development.

To be transparent: I did not manually review every individual line of the AI‑generated changes. Instead, I followed the AI‑guided workflow and focused my efforts on thorough testing. My intention was not to shift the burden of verification onto reviewers, but to ensure—within my current technical abilities—that the resulting behaviour was correct and consistent across platforms.

Although I am not familiar with C++, working through this issue has been a meaningful learning experience for me. Even if my understanding remains limited, the process has helped me gain a clearer sense of how this part of the codebase behaves, and I am grateful for the opportunity to learn through practical experimentation.

I repeatedly built and tested the results on an M1 Max Mac, a virtualised Windows environment, and Ubuntu 24.04.3 ARM64. I also downloaded and tested the GitHub Actions artefacts several times to confirm that the changes functioned reliably. The following builds were produced and confirmed to run successfully on macOS, Windows, and Linux:

### ✔ macOS (Universal)
SuperCollider‑cc9f597d5ab1224b289a8858db87b8c447dd0693‑macOS‑universal.dmg  
[https://github.com/prko/supercollider/actions/runs/26616098521/artifacts/7282519077](https://github.com/prko/supercollider/actions/runs/26616098521/artifacts/7282519077)

### ✔ Ubuntu 22.04 (gcc12)
SuperCollider‑cc9f597d5ab1224b289a8858db87b8c447dd0693‑ubuntu‑22.04‑gcc12  
[https://github.com/prko/supercollider/actions/runs/26616098521/artifacts/7282499904](https://github.com/prko/supercollider/actions/runs/26616098521/artifacts/7282499904)

### ✔ Windows 64‑bit
SuperCollider‑cc9f597d5ab1224b289a8858db87b8c447dd0693‑win64  
[https://github.com/prko/supercollider/actions/runs/26616098521/artifacts/7282530155](https://github.com/prko/supercollider/actions/runs/26616098521/artifacts/7282530155)

Since I have personally verified that the resulting builds run correctly on all three major platforms, I hope that my testing can be of practical help. I also hope that this reduces the burden on reviewers to validate the behaviour across different operating systems.

In addition, I confirmed that the following code snippets behave identically with and without this patch applied:

```
Buffer.read(s, ExampleFiles.child).play

post("-> [1, 2,\n");   // mInResultBlock = true
post("  3, 4,\n");     // remains true (success style) -> OK
post("ERROR: something\n"); // previously: ERROR printed in success style
post("  5, 6]\n");     // also printed in success style
```

I also asked DeepSeek to explain the modified code to me, and although it described the logic as only partially sound, all of the examples it raised as potential concerns behaved correctly in my testing. Based on this, my current understanding is that the changes are acceptable, though I remain open to correction.

I would also appreciate it if the reviewers could take into account that my previous pull request—although seemingly simpler than this one—did not succeed despite my efforts. In contrast, this contribution has undergone much more extensive testing and iteration, and I hope that it will be more useful to the project.

If any part of this contribution requires revision, clarification, or a different approach, I am very willing to adjust it. I appreciate the guidance offered by the maintainers and reviewers, and I welcome any feedback that can help improve this work.

**If, for any reason, this contribution cannot be accepted in its current form, I would be very glad if someone else could build upon the ideas here and create a new pull request so that this improvement may still find its way into the project.**

Thank you for your time and consideration.

----

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -206,11 +206,12 @@ void PostWindow::post(const QString& text) {
     }
 
     const auto lines = text.split("\n");
-    const size_t line_count = lines.size();
+    const auto line_count = lines.size();
     QTextCharFormat result_format = formatForPostLine("->");
 
     for (size_t i { 0 }; i < line_count; ++i) {
-        const QString& line = lines[i];
+        const auto line = lines[i];
+        const auto line_format = formatForPostLine(line);
         cursor.movePosition(QTextCursor::End);
 
         if (line.startsWith("->"))
@@ -225,7 +226,7 @@ void PostWindow::post(const QString& text) {
         } else {
             // words are just text separated by spaces.
             const auto words = line.split(" ");
-            const size_t words_count = words.size();
+            const auto words_count = words.size();
             for (size_t w { 0 }; w < words_count; ++w) {
                 const auto& word = words[w];
                 cursor.movePosition(QTextCursor::End);
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -199,46 +199,39 @@ void PostWindow::post(const QString& text) {
     if (text == "\n") {
         cursor.movePosition(QTextCursor::End);
         cursor.insertText(text, currentFormat);
+        mInResultBlock = false;
 
         if (scroll)
-            emit(scrollToBottomRequest());
+            emit scrollToBottomRequest();
         return;
     }
 
-    const auto ends_with_new_line = text.back() == '\n';
     const auto lines = text.split("\n");
-    const auto line_count = lines.size();
+    const int line_count = lines.size();
 
-    // ★ 이 호출이 "결과 블록"인지 한 번만 판단
-    const bool is_result_block = !lines.isEmpty() && lines.first().startsWith("->");
+    QTextCharFormat result_format = formatForPostLine("->");
 
-    // 결과 블록일 때 쓸 포맷을 미리 준비
-    QTextCharFormat result_format;
-    if (is_result_block) {
-        result_format = formatForPostLine("->");
-    }
-
-    for (size_t i { 0 }; i < line_count; ++i) {
-        const auto line = lines[i];
-
-        // ★ 결과 블록이면 모든 줄에 result_format,
-        //   아니면 기존처럼 줄마다 formatForPostLine(line)
-        const auto line_format = is_result_block ? result_format : formatForPostLine(line);
+    for (int i = 0; i < line_count; ++i) {
+        const QString& line = lines[i];
 
         cursor.movePosition(QTextCursor::End);
+
+        if (line.startsWith("->"))
+            mInResultBlock = true;
+
+        QTextCharFormat line_format = mInResultBlock ? result_format : formatForPostLine(line);
 
         if (!line.contains("://")) {
             cursor.insertText(line, line_format);
         } else {
             const auto words = line.split(" ");
-            const auto words_count = words.size();
-            for (size_t w { 0 }; w < words_count; ++w) {
+            const int words_count = words.size();
+            for (int w = 0; w < words_count; ++w) {
                 const auto& word = words[w];
                 cursor.movePosition(QTextCursor::End);
 
-                if (const auto maybe_url = QUrl(word, QUrl::ParsingMode::StrictMode);
-                    maybe_url.isValid() && word.contains("://")) {
-                    cursor.insertHtml(QString("<a href='") + word + QString("'>") + word + QString("<\\a>"));
+                if (const auto maybe_url = QUrl(word, QUrl::StrictMode); maybe_url.isValid() && word.contains("://")) {
+                    cursor.insertHtml("<a href='" + word + "'>" + word + "</a>");
                 } else {
                     cursor.insertText(word, line_format);
                 }
@@ -250,10 +243,13 @@ void PostWindow::post(const QString& text) {
 
         if (i + 1 != line_count)
             cursor.insertText("\n", line_format);
+
+        if (line.isEmpty())
+            mInResultBlock = false;
     }
 
     if (scroll)
-        emit(scrollToBottomRequest());
+        emit scrollToBottomRequest();
 }
 
 QTextCharFormat PostWindow::formatForPostLine(QString line) {
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -200,20 +200,17 @@ void PostWindow::post(const QString& text) {
         cursor.movePosition(QTextCursor::End);
         cursor.insertText(text, currentFormat);
         mInResultBlock = false;
-
         if (scroll)
-            emit scrollToBottomRequest();
+            emit(scrollToBottomRequest());
         return;
     }
 
     const auto lines = text.split("\n");
-    const int line_count = lines.size();
-
+    const size_t line_count = lines.size();
     QTextCharFormat result_format = formatForPostLine("->");
 
-    for (int i = 0; i < line_count; ++i) {
+    for (size_t i { 0 }; i < line_count; ++i) {
         const QString& line = lines[i];
-
         cursor.movePosition(QTextCursor::End);
 
         if (line.startsWith("->"))
@@ -221,26 +218,32 @@ void PostWindow::post(const QString& text) {
 
         QTextCharFormat line_format = mInResultBlock ? result_format : formatForPostLine(line);
 
+        // If there is some text that looks like a URI and contains '://' then turn it into a html anchor so we can
+        // click it.
         if (!line.contains("://")) {
             cursor.insertText(line, line_format);
         } else {
+            // words are just text separated by spaces.
             const auto words = line.split(" ");
-            const int words_count = words.size();
-            for (int w = 0; w < words_count; ++w) {
+            const size_t words_count = words.size();
+            for (size_t w { 0 }; w < words_count; ++w) {
                 const auto& word = words[w];
                 cursor.movePosition(QTextCursor::End);
 
-                if (const auto maybe_url = QUrl(word, QUrl::StrictMode); maybe_url.isValid() && word.contains("://")) {
-                    cursor.insertHtml("<a href='" + word + "'>" + word + "</a>");
+                if (const auto maybe_url = QUrl(word, QUrl::ParsingMode::StrictMode);
+                    maybe_url.isValid() && word.contains("://")) {
+                    cursor.insertHtml(QString("<a href='") + word + QString("'>") + word + QString("<\\a>"));
                 } else {
                     cursor.insertText(word, line_format);
                 }
 
+                // Put space back in, if not last word.
                 if (w + 1 != words_count)
                     cursor.insertText(" ", line_format);
             }
         }
 
+        // Don't write a new line in the final case.
         if (i + 1 != line_count)
             cursor.insertText("\n", line_format);
 
@@ -249,7 +252,7 @@ void PostWindow::post(const QString& text) {
     }
 
     if (scroll)
-        emit scrollToBottomRequest();
+        emit(scrollToBottomRequest());
 }
 
 QTextCharFormat PostWindow::formatForPostLine(QString line) {
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -44,7 +44,7 @@
 
 namespace ScIDE {
 
-static const QString RESULT_MARKER = QString::fromUtf8(u8"\u200B\u200C\u200D\u2060");
+static const QString RESULT_MARKER = QString::fromUtf8(u8"\u2060\u2060");
 
 PostWindow::PostWindow(QWidget* parent): QPlainTextEdit(parent) {
     setReadOnly(true);
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -196,10 +196,12 @@ void PostWindow::post(const QString& text) {
     const bool scroll = mActions[AutoScroll]->isChecked();
     QTextCursor cursor(document());
 
+    mInResultBlock = false;
+
     if (text == "\n") {
         cursor.movePosition(QTextCursor::End);
         cursor.insertText(text, currentFormat);
-        mInResultBlock = false;
+
         if (scroll)
             emit(scrollToBottomRequest());
         return;
@@ -213,8 +215,10 @@ void PostWindow::post(const QString& text) {
         const auto line = lines[i];
         cursor.movePosition(QTextCursor::End);
 
-        if (line.startsWith("->"))
+        bool isResultLine = line.startsWith("->");
+        if (isResultLine) {
             mInResultBlock = true;
+        }
 
         QTextCharFormat line_format = mInResultBlock ? result_format : formatForPostLine(line);
 
@@ -246,9 +250,6 @@ void PostWindow::post(const QString& text) {
         // Don't write a new line in the final case.
         if (i + 1 != line_count)
             cursor.insertText("\n", line_format);
-
-        if (line.isEmpty())
-            mInResultBlock = false;
     }
 
     if (scroll)
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -211,7 +211,6 @@ void PostWindow::post(const QString& text) {
 
     for (size_t i { 0 }; i < line_count; ++i) {
         const auto line = lines[i];
-        const auto line_format = formatForPostLine(line);
         cursor.movePosition(QTextCursor::End);
 
         if (line.startsWith("->"))
--- a/editors/sc-ide/widgets/post_window.cpp
+++ b/editors/sc-ide/widgets/post_window.cpp
@@ -44,6 +44,8 @@
 
 namespace ScIDE {
 
+static const QString RESULT_MARKER = QString::fromUtf8(u8"\u200B\u200C\u200D\u2060");
+
 PostWindow::PostWindow(QWidget* parent): QPlainTextEdit(parent) {
     setReadOnly(true);
     setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOn);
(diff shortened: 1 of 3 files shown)


JordanHendersonMusic:
There are many unrelated changes in the code. Could you go through and revert all the changes that don't contribute to this pr?

This is the problem with ai, it just does random crap that you didn't ask it to. This makes reviewing hard and pollutes the git history.

The actual change here is alright, but it should be about 4 lines or code.

prko:
OK, I will do! Thamks1

editors/sc-ide/widgets/post_window.cpp
```diff
@@ -199,28 +199,33 @@ void PostWindow::post(const QString& text) {
...
+
         // If there is some text that looks like a URI and contains '://' then turn it into a html anchor so we can
         // click it.
         if (!line.contains("://")) {
             cursor.insertText(line, line_format);
         } else {
             // words are just text separated by spaces.
             const auto words = line.split(" ");
-            const auto words_count = words.size();
+            const size_t words_count = words.size();
```
capital-G: unnecessary change

editors/sc-ide/widgets/post_window.cpp
```diff
@@ -199,28 +199,33 @@ void PostWindow::post(const QString& text) {
...
+        mInResultBlock = false;
         if (scroll)
             emit(scrollToBottomRequest());
         return;
     }
 
-    const auto ends_with_new_line = text.back() == '\n';
     const auto lines = text.split("\n");
-    const auto line_count = lines.size();
+    const size_t line_count = lines.size();
```
capital-G: unnecessary change

editors/sc-ide/widgets/post_window.cpp
```diff
@@ -199,28 +199,33 @@ void PostWindow::post(const QString& text) {
...
+    QTextCharFormat result_format = formatForPostLine("->");
+
     for (size_t i { 0 }; i < line_count; ++i) {
-        const auto line = lines[i];
-        const auto line_format = formatForPostLine(line);
+        const QString& line = lines[i];
```
capital-G: unnecessary change

editors/sc-ide/widgets/post_window.hpp
```diff
@@ -119,6 +119,9 @@ private slots:
     /// Will be empty when no anchor has been clicked
     /// Used in conjuction with anchorAt(QPos)
     QString clickedAnchor;
+
+    // State for multi-line result block formatting
```
capital-G: docs are written w/ 3 `///` iirc

```suggestion
    /// State for multi-line result block formatting
```

capital-G (commented):
I am a bit unsure about the last

```cpp
        if (line.isEmpty())
            mInResultBlock = false;
```

Is `text` always terminated by `\n` or not?

prko:
>I am a bit unsure about the last
>
>        if (line.isEmpty())
>            mInResultBlock = false;
>Is text always terminated by \n or not?

Oh, an important question!
Here are two examples showing that SuperCollider output is not consistently
terminated by `\n`:

```supercollider
"1st line\n\n3rd line\n".post   // ends with newline → produces a trailing empty line
"1st line\n\n3rd line".post     // no trailing newline → no empty line
```
Screenshot 2026-05-30 at 10 24 55

----

After reading @capital-G’s comment, I realised that this PR still does not handle the following two cases correctly:

1. Case: no trailing newline
   ```supercollider
   "1\n\n1".post
   ```
   returns
   Screenshot 2026-05-30 at 07 03 25
   ```text
   1        //// (black)
   
   1-> 1 //// (black)
   
   1        //// (black)
   ```
   Here, the return value (`-> 1`) does not start on a new line, so the PostWindow cannot detect the result block correctly.

2. Case: trailing newline
   ```supercollider
   "1\n\n1\n".post
   ```
   returns
   Screenshot 2026-05-30 at 07 03 12
   ```text
   1
   
   1
   -> 1 //// (green)
   
   1
   ```
   Here, the trailing newline produces an empty line, and the current algorithm treats this empty line as the end of the result block, which is incorrect.

## Possible directions
### To resolve the first case:
sclang (class level or primitive level) should ensure that a return value always begins on a new line. If the return value does not start on a new line, sclang should insert a line break before printing it.

### To resolve the second case:
The PostWindow should not rely on `line.isEmpty()` to detect the end of a
result block. Instead, sclang should send the complete output buffer to the
IDE without splitting or interpreting line breaks prematurely.

Does this interpretation make sense?

editors/sc-ide/widgets/post_window.cpp
prko: Thanks! Done with https://github.com/supercollider/supercollider/pull/7525/commits/d812f9be771965697a7db0bfefc9a678d4c3e00a

prko:
@capital-G 

Thank you for the correction.
To use AI effectively for C++, I need to learn the basics of C++ first.

editors/sc-ide/widgets/post_window.cpp
prko: Thanks! I’ve updated the code in https://github.com/supercollider/supercollider/pull/7525/commits/29cf274688c684db468a3972b6e1a4002943efde.

editors/sc-ide/widgets/post_window.cpp
prko: Thanks! I’ve updated the code in https://github.com/supercollider/supercollider/pull/7525/commits/29cf274688c684db468a3972b6e1a4002943efde.

editors/sc-ide/widgets/post_window.cpp
prko: Thanks! I’ve updated the code in https://github.com/supercollider/supercollider/pull/7525/commits/29cf274688c684db468a3972b6e1a4002943efde, but an error occurred. I might have made a mistake.

JordanHendersonMusic:
> Is `text` always terminated by `\n` or not?


I seem to recall that you can have a long piece of text it can get split over multiple buffers, if that is the case then the answer would be no.

@prko good idea to test with larger inputs!

editors/sc-ide/widgets/post_window.cpp
prko: Resolved in https://github.com/supercollider/supercollider/pull/7525/commits/a2d8759e99c95e5e4266868a07abdad60ff44487

prko:
With [a2d8759](https://github.com/supercollider/supercollider/pull/7525/commits/a2d8759e99c95e5e4266868a07abdad60ff44487), it looks like the unnecessary AI-generated changes have been cleaned up. We still need to address the three issues that were mentioned:
1. ```
    if (line.isEmpty())
        mInResultBlock = false;
    ```
2. ```
   1
   
   1
   -> 1 //// (green)
   
   1
   ```
3.  ```
    1        //// (black)
    
    1-> 1 //// (black)
    
    1        //// (black)
    ```

capital-G:
> To use AI effectively for C++, I need to learn the basics of C++ first.

I think you could use AI to let you explain what it does along the way: ask *how* and *why* things are done, rather than letting it things do for you. I don't mind reviewing code that was generated w/ AI by you as long as I have the feeling that you get something out of it :)
At the same time: The are many "levels" of C++ - and Qt C++ is quite manageable, so I think it is a good entrance :) 

I think the problem of determining the end of a result block either results in some quirky formatting issues or solving the [halting problem](https://en.wikipedia.org/wiki/Halting_problem) ;)

IIRC, the `->` prefix is created by sclang itself

https://github.com/supercollider/supercollider/blob/4d4e9ec7bcb8e58aea85283cc2138864b83589e7/SCClassLibrary/Common/Core/Kernel.sc#L705-L720

One way to resolve this could be to add an (invisible?) termination symbol so we could know for sure when the return value has been finished printing.
Or - probably the actual clean but more heavy implementation: switch to ansii codes https://en.wikipedia.org/wiki/ANSI_escape_code
Currently the post window is **not** a terminal emulator, and a terminal emulator is actually a somehow complex dependency, but we could maybe look into implementing some basic behavior? Could be a nice feature IMO.

prko:
https://github.com/supercollider/supercollider/pull/7525/commits/916670e6480dc4b8bda4e1af55a2a8b3c20e5e25: This commit resolves two of the previously mentioned issues:

1. ```
    if (line.isEmpty())
        mInResultBlock = false;
    ```
2. ```
   1

   1
   -> 1 //// (green)
   
   1    //// (green)
   ```

However, the following problem still remains:
-  ```
    1        //// (black)
    
    1-> 1 //// (black)
    
    1        //// (black)
    ```
To fix this, I attempted to modify the following section but was unsuccessful:
https://github.com/supercollider/supercollider/blob/4d4e9ec7bcb8e58aea85283cc2138864b83589e7/editors/sc-ide/widgets/post_window.cpp#L211-L249

AI suggested using `line.indexOf("->")` instead of `line.startsWith("->")` to detect a PostWindow return appearing in the middle of a line. However, this approach is problematic because key–value pairs also use `->`.

Would it be possible to change `->` to `LLR:\n` in the following location?https://github.com/supercollider/supercollider/blob/4d4e9ec7bcb8e58aea85283cc2138864b83589e7/SCClassLibrary/Common/Core/Kernel.sc#L719

`LLR` stands for last line return, and it is also easy for beginners to understand.

The output would then look like this:

--- 

```supercollider
(
"line one\
line two\
line three".postcs
)
```
returns:
```text
"line one
line two
line three"
LLR:       // (green)
line one.  // (green)
line two   // (green)
line three // (green)
```

---

```supercollider
(
'line one\
line two\
line three'
.post
)
```
```text
line one          // (black)
line two          // (black)
line threeLLR: // line three (black) LLR: (green)
line one         // (green)
line two         // (green)
line three      // (green) 
```

---

```supercollider
("1
2
3
4

5
"
)
```
```text
LLR:  // (green)
1  // (green)
2  // (green)
3  // (green)
4  // (green)

5   // (green)
```
A screenshot from a locally built SC on Ubuntu:
Screenshot 2026-05-31 at 01 26 27

As you see through the screenshot, I confirmed that the modification works as intended in my local build.
Since this change introduces a new output marker (`LLR:`), I am sharing the idea first to check whether it aligns with the project’s direction.
If there are no objections, I can submit it as a commit.

Note:  
`LLR:` does not conflict with any existing syntax or operators in SuperCollider, so it should be safe to introduce as a marker.
Since it is visually subtle compared to the current `-> `, we could also consider extending it (e.g., `LLR::\n` or `LLR →\n`) to improve readability, depending on the project’s preference.

JordanHendersonMusic:
Alternatively, sclang could just call a primitive when it outputs the result of an expression which is different from the normal post, there should also be one for errors, then the language client just determines what to do with it unambiguously.

prko:
@capital-G 
> One way to resolve this could be to add an (invisible?) termination symbol so we could know for sure when the return value has been finished printing.

Applying this idea to `->`:
Since I have successfully used `LLR:` (in `.interpretPrintCmdLine`) together with `indexOf` (in `post_window.cpp`), we could instead use some invisible characters. Because the last-line return is already distinguished by colour, we no longer need a visible marker like `->` or `LLR:`. Moreover, the `-> ` indentation looks awkward when multiple lines are returned.

@JordanHendersonMusic 
>Alternatively, sclang could just call a primitive when it outputs the result of an expression which is different from the normal post, there should also be one for errors, then the language client just determines what to do with it unambiguously.

Yes, I agree! That approach would make the output types unambiguous and remove the need for any textual markers entirely.

prko:
I have several questions related to my previous post:

1. In `post_window.cpp`, we seem to use the same characters like this:
   ```c++
   line.startsWith(u8"\u200B\u200C\u200D\u2060\uFEFF")
   ```
   and
   ```c++
   int idx = line.indexOf(u8"\u200B\u200C\u200D\u2060\uFEFF");
   ```
   However, in sclang, using the escape sequence "\u200B\u200C\u200D\u2060\uFEFF" does not work.
To use these invisible characters, we need to insert the actual characters directly, for example:
   ```supercollider
   ("​‌‍⁠﻿" ++ res).postln;
   ```
   Is this acceptable?
2. Would it be possible to use visible markers such as `"–>"` (en-dash), `"—>"` (em-dash), `"⟶"`,`"⟹"`, `"→"`, `"⇒"` or , `"⇨"` instead of `"->"`?

prko:
With commit https://github.com/supercollider/supercollider/pull/7525/commits/4c16f638b8a8b42f851f6cdc70dbad08b42b2a02, the Post Window no longer prefixes returned values with `-> `. Instead, it marks the return value using a sequence of invisible characters:  `u8"\u200B\u200C\u200D\u2060"`
Here is a screenshot:
Screenshot 2026-05-31 at 08 01 04

We could also add a horizontal rule 
```supercollider
("​‌‍⁠" ++ res + "\n──────────────────────────────").postln;
```
after the return value to make it more visually distinct:
Screenshot 2026-05-31 at 10 03 07

prko:
I tested this further in the macOS terminal.

1. The non‑ASCII arrow characters (⟶ ⟹ → ⇒ ⇨) do not look good visually.
2. There is no syntax coloration for the return marker defined here:  
   <https://github.com/supercollider/supercollider/blob/4d4e9ec7bcb8e58aea85283cc2138864b83589e7/SCClassLibrary/Common/Core/Kernel.sc#L719>
3. Therefore, starting the return value with invisible characters is not a good idea.
4. `"->"` is still ambiguous when programming or manipulating strings, because several classes use it for key–value pairs.
5. I think combining a verbal label with symbols and a newline is a reasonable approach. (The first‑line indentation caused by "-> " looks visually odd.) For example:

   ```supercollider
   ("* LAST RETURN *\n" ++ res ++ "\n──────────────────────────────").postln;
   ```
   Screenshot 2026-05-31 at 10 54 30
   The horizontal lines are very helpful when debugging, as shown in the screenshot.

prko:
I tested several additional characters, and I think the following options work better than the others:

- ```prompt
   ("⮑\n" ++ res ++ "\n──────────────────────────────")`
   ```
- ```
   ("↳\n" ++ res ++ "\n──────────────────────────────")`
   ```
- ```
   ("*LAST*\n" ++ res ++ "\n──────────────────────────────")`

Note that the shape of `⮑` differs between macOS and other operating systems.

Screenshots from Windows 11:
Screenshot 2026-05-31 at 15 36 48 copy
Screenshot 2026-05-31 at 15 36 48
Screenshot 2026-05-31 at 15 44 15

---

I think this PR would benefit from being co‑authored, as there are many aspects to consider.

prko:
Strangely, SC‑IDE on macOS renders `"⮑"` as a rectangular fallback glyph in both the code editor and the post window, as shown below:
Screenshot 2026-05-31 at 19 08 10
Because of this, `"⮑"` cannot be used.
`"↳"` displays correctly and works well.

JordanHendersonMusic:
Sorry I just don't think this is worth it. What if the users has a string with that character in, does that mess up all formatting afterwards? Parsing the output just isn't the correct way to solve this.

prko:
>What if the users has a string with that character in, does that mess up all formatting afterwards?

Yes.  
I think the strings and characters I mentioned are rarely used in `.scd` files, but your observation is correct.

prko:
>Sorry I just don't think this is worth it. What if the users has a string with that character in, does that mess up all formatting afterwards? Parsing the output just isn't the correct way to solve this.

However, a specific sequence of invisible characters followed by the characters below will never be used:

- Display: `"​‌‍⁠﻿↳\n"`
   Characters used: `"\u200B\u200C\u200D\u2060\uFEFF↳\n"`

- Display: `"​‌‍⁠﻿*LAST*\n"`
   Characters used: `"\u200B\u200C\u200D\u2060\uFEFF*LAST*\n"`

- Display: `"​‌‍⁠﻿::LAST::\n"`
   Characters used: `"\u200B\u200C\u200D\u2060\uFEFF::LAST::\n"`

Using this kind of combination will never occur in typical `.scd`, `.sc`, or `.schelp` files.

prko:
After further testing, I believe that the commit
<https://github.com/supercollider/supercollider/pull/7525/commits/a9cc3d76626e3d0f6341bf7aa335d098daf1cec1>
currently provides the best behavior.

With this commit applied, sclang prints the output in both the SC-IDE and the terminal as follows:
`u8"\u2060\u2060" ++ "::LAST::>\n"`

The issue with this combination is that both `\u2060` and `\u2060\u2060` introduce a gap—similar to a whitespace—when they appear as the first character of a line in the macOS terminal and the Windows command prompt.
On the Ubuntu terminal, this gap does not appear:

- macOS 
  - SC-IDE
     Screenshot 2026-06-01 at 03 21 27
  - terminal
     Screenshot 2026-06-01 at 03 52 30

 - Windows 
    - SC-IDE
       Screenshot 2026-06-01 at 03 46 38
   - terminal
     Screenshot 2026-06-01 at 03 47 17

- Ubuntu 
  - SC-IDE
     Screenshot 2026-06-01 at 03 55 15
  - terminal
     Screenshot 2026-06-01 at 03 53 53

---

I hope that the commit <https://github.com/supercollider/supercollider/pull/7525/commits/a9cc3d76626e3d0f6341bf7aa335d098daf1cec1>,
for which the screenshots are attached in this post, can be accepted.

For comparison:

The commit <https://github.com/supercollider/supercollider/commit/4c16f638b8a8b42f851f6cdc70dbad08b42b2a02> does not include any visible character or string.

The commit <https://github.com/supercollider/supercollider/commit/916670e6480dc4b8bda4e1af55a2a8b3c20e5e25> fixes only the colour (see details in <https://github.com/supercollider/supercollider/pull/7525#issuecomment-4583442726>).

If none of these approaches is considered acceptable, this PR can simply remain as a draft for future work.

capital-G:
I personally don't like making the output this noisy and I know some coding-performances which would get broken by such an implementation, so I would veto this.

IMO, as stated, I think using ANSI colors is the cleanest approach (and is what other programming languages are using), followed by using an invisible character - e.g. ascii 0x04 https://en.wikipedia.org/wiki/End-of-Transmission_character

Before start implementing anything like this, please wait for feedback from Jordan since I think he knows best what would be the proper way of solving this.

capital-G (changes requested):

prko:
Sorry for merging the branch — that was my mistake.

>IMO, as stated, I think using ANSI colors is the cleanest approach (and is what other programming languages are using), followed by using an invisible character - e.g. ascii 0x04 https://en.wikipedia.org/wiki/End-of-Transmission_character

I agree with your point.
After testing, `::LAST::>\n` does feel a bit awkward, and `::>\n` is cleaner.
But ideally, I also think it would be best to avoid visible markers entirely and rely on colour only, using an invisible character if needed.
The `End‑of‑Transmission character` seems like a reasonable option, but I’m open to whatever approach the maintainers consider most appropriate.

JordanHendersonMusic:
> I personally don't like making the output this noisy

I agree.

>  I think using ANSI colors is the cleanest approach

Problem is when the editor doesn't know how to display these and then you get a mess again. Some of github's output looks like a mess due to these in the ci.

> ascii 0x04

I think this is a good idea, but there is a similar problem where we really want to highlight the error messages in red.
Right now that is done by checking to see whether the line begins with `ERROR:`, this is useful for multi-line errors. This could also be done with the terminator character.

I think a better solution would be for sclang to write these different types of messages to different streams, then the language client can write them to different places if requested to do so, or colour/format them as needed. This removes the ambiguity. This means the errors can go straight to stderr if wanted.

If people relied on some particular format of this, it would be quite easy to add a backwards compatibly mode, since all the changes would happen in the language client and can be controlled through the config file. 

While this PR does fix the issue, I think it detracts in other ways. It might have been more time efficient to have created an issue and proposed a solution before beginning work.

prko:
Thanks for the feedback. That makes sense. I’ll close this PR and open an issue so we can discuss the approach properly.

PR #7535, opened 2026-06-01
Update `sounds`, `sounds/*` etc

prko:
## Purpose and Motivation
In some files, paths such as`sounds`, `sounds/*, "sounds/a11wlk01.wav" are still used.
This PR updates those paths and introduces  `ExampleFiles.dir`.
I tested all changes except for [testsuite/classlibrary/TestNetAddr_Server.sc](https://github.com/supercollider/supercollider/compare/develop...prko:topic/fix_foldername_sounds?expand=1#diff-464e5ec81fdea08a7f82b8f7a4b654e96dcb23daca68e745855f80b9aaf427d6).

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/PathName.schelp
+++ b/HelpSource/Classes/PathName.schelp
@@ -155,6 +155,54 @@ myPath = PathName.new(ExampleFiles.child.dirname);
 myPath.allFolders.postln;
 )
 ::
+discussion::
+code::
+(
+var test_FilePath_original       = PathName(ExampleFiles.child);
+var test_BasePath                = ("~/Downloads/test__" ++ Date.seed).standardizePath;
+var test_FolderPath_wo_extension = test_BasePath +/+ test_FilePath_original.extension;
+var test_FolderPath_w_extension  = test_BasePath +/+ test_FilePath_original.extension ++ ".folder";
+var test_FilePath_w_extension    = test_BasePath +/+ test_FilePath_original.fileName;
+var test_FilePath_wo_extension   = test_BasePath +/+ test_FilePath_original.fileNameWithoutExtension;
+
+[test_BasePath, test_FolderPath_wo_extension, test_FolderPath_w_extension].do { |path|
+	File.mkdir(path)
+};
+
+[test_FilePath_w_extension, test_FilePath_wo_extension].do { |path|
+	File.copy(test_FilePath_original.fullPath, path)
+};
+
+(
+"FILE (with extension):\n\tPathName("
+++ test_FilePath_w_extension ++ ").allFolders\n\t\t"
+++ PathName(test_FilePath_w_extension).allFolders
+++ "\n\t\t     <- All folder names are included."
+
+++ "\n\nFILE (without extension):\n\tPathName("
+++ test_FilePath_wo_extension ++ ").allFolders\n\t\t"
+++ PathName(test_FilePath_wo_extension).allFolders
+++ "\n\t\t     <- All folder names are included."
+
+++ "\n\nFOLDER (without extension):\n\tPathName(" ++ test_FolderPath_wo_extension
+++ ").allFolders\n\t\t"
+++ PathName(test_FolderPath_wo_extension).allFolders
+++ "\n\t\t     <- The folder name \"wav\" is excluded."
+
+++ "\n\nFOLDER (with extension):\n\tPathName(" ++ test_FolderPath_w_extension
+++ ").allFolders\n\t\t"
+++ PathName(test_FolderPath_w_extension).allFolders
+++ "\n\t\t     <- The folder name \"wav.folder\" is excluded.\n"
+).postln;
+
+test_BasePath.openOS;
+
+~test_BasePath = test_BasePath
+/*
+File.deleteAll(~test_BasePath) // Evaluate it to remove the test_ files/foler.
+*/
+)
+::
 
 method::diskName
 if path is an absolute path, returns the disk name; else a blank string.
--- a/HelpSource/Classes/PathName.schelp
+++ b/HelpSource/Classes/PathName.schelp
@@ -138,7 +138,6 @@ myPath.filesDo{ |afile| afile.postln };
 ::
 
 method::allFolders
-note::
 This method has known reliability issues and is generally discouraged.
 Use instead link::Classes/String#-dirname:: and link::Classes/String#-split:: in combination:
 code::
@@ -156,62 +155,17 @@ teletype::
 -> [, Imaginary, Directory]
 ::
 code::
-"D:/Directory/fish.rtf".dirname.split        // on Windows
+"D:/Directory/fish.rtf".dirname.split         // on Windows
 ::
 returns:
 teletype::
 -> [D:, Directory]
 ::
-::
-
-Does a string comparison based on LINK::Classes/Platform#*pathSeparator::, and will return all entries which are surrounded by this separator.
-Since it solely relies on string comparison, this will not check the context of the filesystem to verify if the returned entries are actually folders.
-
 
 method::diskName
 
-note::
 This method has known reliability issues and is generally discouraged.
 It remains available for legacy use cases, and the examples below illustrate how to use it when necessary.
-::
-
-if path is an absolute path, returns
-list::
-## the top-level directory of the path on Linux and macOS,
-## the drive letter on Windows
-::
-else a blank string.
-code::
-( // note the / at the start is excluded
-var myPath;
-myPath = PathName.new(ExampleFiles.child.replaceRegexp("^/", ""));
-myPath.diskName;
-)
-::
-returns on Linux:
-teletype::
--> usr
-::
-returns on macOS (The result below assumes that SuperCollider.app is located in the teletype::Applications:: folder.):
-teletype::
--> Applications
-::
-returns on Windows:
-teletype::
--> C:
-::
-
-code::
-( // note the / at the start
-var myPath;
-myPath = PathName.new( "/test" +/+ ExampleFiles.child);
-myPath.diskName;
-)
-::
-returns on all platforms:
-teletype::
-->
-::
 
 method::+/+
 Path concatenation operator - useful for avoiding doubling-up slashes unnecessarily.
--- a/HelpSource/Classes/PathName.schelp
+++ b/HelpSource/Classes/PathName.schelp
@@ -142,9 +142,9 @@ Does a string comparison based on LINK::Classes/Platform#*pathSeparator::, and w
 Since it solely relies on string comparison, this will not check the context of the filesystem to verify if the returned entries are actually folders.
 
 code::
-/* Test with a file */
+/* a separator at the beginning; no separator at the end */
 
-// a separator at the beginning; no separator at the end:
+// test with a file:
 (
 var filePath = ExampleFiles.child.debug;
 var myPath = PathName.new(filePath);
@@ -152,32 +152,7 @@ myPath.postln;
 myPath.allFolders; // includes the top-level directory; excludes a11wlk01.wav
 )
 
-(// separators at the beginning and at the end:
-var filePath = (ExampleFiles.child ++ "/").debug;
-var myPath = PathName.new(filePath);
-myPath.postln;
-myPath.allFolders; // includes both the top-level directory and a11wlk01.wav
-)
-
-// no separators at the beginning or the end:
-(
-var filePath = (ExampleFiles.child.replaceRegexp("^/", "")).debug;
-var myPath = PathName.new(filePath);
-myPath.postln; // PathName(filePath) adds a separator at the beginning when there is no separator at the end
-myPath.allFolders; // includes the top-level directory; excludes a11wlk01.wav
-)
-
-// no separator at the beginning; a separator at the end:
-(
-var filePath = (ExampleFiles.child.replaceRegexp("^/", "") ++ "/").debug;
-var myPath = PathName.new(filePath);
-myPath.postln; // PathName(filePath) does not add a separator at the beginning when there is one at the end
-myPath.allFolders; // excludes the top-level directory; includes a11wlk01.wav
-)
-
-/* Test with a folder*/
-
-// a separator at the beginning; no separator at the end:
+// test with a folder:
 (
 var folderPath = ExampleFiles.child.dirname.debug;
 var myPath = PathName.new(folderPath);
@@ -185,7 +160,18 @@ myPath.postln;
 myPath.allFolders; // includes the top-level directory; excludes the bottom-most folder
 )
 
-// separators at the beginning and at the end:
+
+/* separators at the beginning and at the end */
+
+// test with a file:
+(
+var filePath = (ExampleFiles.child ++ "/").debug;
+var myPath = PathName.new(filePath);
+myPath.postln;
+myPath.allFolders; // includes both the top-level directory and a11wlk01.wav
+)
+
+// test with a folder:
 (
 var folderPath = (ExampleFiles.child.dirname ++ "/").debug;
 var myPath = PathName.new(folderPath);
@@ -193,20 +179,58 @@ myPath.postln;
 myPath.allFolders; // includes both the top-level directory and the bottom-most folder
 )
 
-// no separators at the beginning or the end:
+
+/* no separators at the beginning or the end */
+
+// test with a file:
+(
+var filePath = (ExampleFiles.child.replaceRegexp("^/", "")).debug;
+var myPath = PathName.new(filePath);
+myPath.postln;
+
+myPath.allFolders; 
+// linux & macOS: excludes both the top-level directory and a11wlk01.wav
+// Windows: includes the top-level directory; excludes a11wlk01.wav
+// See the PathName method .diskName
+)
+
+// test with a folder:
 (
 var folderPath = (ExampleFiles.child.dirname.replaceRegexp("^/", "")).debug;
 var myPath = PathName.new(folderPath);
-myPath.postln; // PathName(folderPath) adds a separator at the beginning when there is no separator at the end
-myPath.allFolders; // includes the top-level directory; excludes the bottom-most folder
+myPath.postln;
+
+myPath.allFolders; 
+// linux & macOS: excludes both the top-level directory and the bottom-most folder
+// Windows: includes the top-level directory; excludes the bottom-most folder
+// See the PathName method .diskName
 )
 
-// no separator at the beginning; a separator at the end:
+
+/* no separator at the beginning; a separator at the end */
+
+// test with a file:
+(
+var filePath = (ExampleFiles.child.replaceRegexp("^/", "") ++ "/").debug;
+var myPath = PathName.new(filePath);
+myPath.postln;
+
+myPath.allFolders; 
+// linux & macOS: excludes the top-level directory; includes a11wlk01.wav
+// Windows: includes both the top-level directory and a11wlk01.wav
+// See the PathName method .diskName
+)
+
+// test with a folder:
 (
 var folderPath = (ExampleFiles.child.dirname.replaceRegexp("^/", "") ++ "/").debug;
 var myPath = PathName.new(folderPath);
-myPath.postln; // PathName(folderPath) adds a separator at the beginning; PathName(folderPath) removes the separator at the end
-myPath.allFolders; // includes the top-level directory; excludes a11wlk01.wav
+myPath.postln;
+
+myPath.allFolders; 
+// linux & macOS: excludes the top-level directory; includes the bottom-most folder
+// Windows: includes both the top-level directory and the bottom-most folder
+// See the PathName method .diskName
 )
 ::
 
@@ -224,24 +248,28 @@ var filePath_wo_extension   = basePath +/+ filePath_original.fileNameWithoutExte
 [basePath, folderPath_wo_extension, folderPath_w_extension].do { |path| path.mkdir };
 
 [filePath_w_extension, filePath_wo_extension].do { |path|
-	File.copy(filePath_original.fullPath, path)
+    File.copy(filePath_original.fullPath, path)
 };
 
 [
-	["FILE with extension:\n\t", filePath_w_extension],
-	["FILE without extension:\n\t", filePath_wo_extension],
-	["FOLDER without extension:\n\t", folderPath_wo_extension],
-	["FOLDER with extension:\n\t", folderPath_w_extension],
+    ["FILE with extension:\n\t", filePath_w_extension],
+    ["FILE without extension:\n\t", filePath_wo_extension],
+    ["FOLDER without extension:\n\t", folderPath_wo_extension],
+    ["FOLDER with extension:\n\t", folderPath_w_extension],
 ].do { |array|
-	var code = "PathName(" ++ array[1].quote ++ ").allFolders";
-	(array[0] ++ code ++ "\n\t\t" ++ code.interpret.asString).postln;
-	(
-		if(array[0].contains("FILE")) {
-			"\t\t     <- All folder names are included."
-		} {
-			"\t\t     <- The folder name \"" ++ array[1].basename++ "\" is excluded."
-		}
-		++ "\n").postln
+	var path = Platform.case(
+		\windows, { array[1].replace("/", "\\").replace("\\", "\\\\") },
+		{ array[1] }
+	);
+	var code = "PathName(" ++ path.quote ++ ").allFolders";
+    (array[0] ++ code ++ "\n\t\t" ++ code.interpret.asString).postln;
+    (
+        if(array[0].contains("FILE")) {
+            "\t\t     <- All folder names are included."
+        } {
+            "\t\t     <- The folder name \"" ++ array[1].basename++ "\" is excluded."
+        }
+        ++ "\n").postln
 };
 
 basePath.openOS;
@@ -252,7 +280,12 @@ File.deleteAll(~basePath) // Evaluate this to remove the files and folders creat
 ::
 
 method::diskName
-if path is an absolute path, returns the root directory on the storage; else a blank string.
+if path is an absolute path, returns
+list::
+## the top-level directory of the path on linux and macOS,
+## the drive letter on Windows
+::
+else a blank string.
 code::
 ( // note the / at the start is excluded
 var myPath;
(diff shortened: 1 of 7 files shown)


SCClassLibrary/Common/Files/ExampleFiles.sc
```diff
@@ -10,4 +10,8 @@ ExampleFiles {
 	*child {
 		^Platform.resourceDir +/+ "sounds" +/+ "a11wlk01.wav";
 	}
+
+	*dir {
+		^Platform.resourceDir +/+ "sounds";
+	}
```
capital-G: There was some discussion during the creation of `ExampleFiles` if it should only cover sound files or not, and there was no clear settlement on the issue - therefore hardcoding it to a `sounds` directory seems to settle this discussion, but I don't see how much gets gained by doing so.

Additionally, this could lead people to use directory "arithmetics", which was the idea to get rid of this in the first place. By using this class we could move the example files elsewhere w/o breaking peoples code, which is not possible w/ the exposed `dir`.

prko: Yes, I agree with you. For this reason, I am also not sure whether introducing `ExampleFiles.dir` is a good choice. It was essentially a side‑effect‑like decision made to replace the non‑existent paths `"sounds/"`, `"sounds/*"`, and `"MyDisk/SC 2.2.8 f/Sounds/FunkyChicken"`. The only real advantage is achieving consistent documentation, and I am not sure that benefit outweighs the conceptual ambiguity it introduces. Since I opened this PR with documentation consistency in mind, I am open to modifying it based on further feedback from other developers.

dyfer: The test suite also assumes this class only points to individual files. I don't think we should be adding other methods here, at least for now.

prko: OK! Done in <https://github.com/supercollider/supercollider/pull/7535/commits/a10e5e181e194d94c6b32bd7d7c9737485d287eb>.

Thanks for your opinion. I was unsure about the idea, and your comments helped clarify the direction.

prko:
A note:

`"/MyDisk/SC 2.2.8 f/Sounds/FunkyChicken"`
This path looks reasonable for macOS and Linux users.
However, Windows users who have never used macOS or Linux may find it confusing.

HelpSource/Classes/PathName.schelp
```diff
@@ -138,11 +138,20 @@ myPath.filesDo{ |afile| afile.postln };
 ::
 
 method::allFolders
-returns a list of all the folder names contained in the pathname itself.
+Returns all folder names in the path.
+If the path ends with a file, all folders in the path are returned.
+If the path ends with a folder, the final folder (the folder itself) is not included.
```
capital-G: This is actually not correct, given the path `/foo/bar`

```supercollider
PathName("/hello/foo").allFolders; // -> List[hello]
PathName("/hello/foo/").allFolders; // -> List[hello, foo]
```

`foo` can either be a folder or a file [at least in the first writing] - I would actually expect `PathName` to check if `foo` is a file or a folder and return its result accordingly. [In fact, `/hello/foo` does not exist on my system, so it couldn't determine if this is a folder or a file.] I don't really understand the purpose of this method?

This is an implementation error/flaw though, but maybe we can document this that this can lead to unexpected results?

capital-G (commented):
Any ideas how to move forward here @prko? I don't really know what to do about the implementation flaw, but I think your text at least improves it from its current state.

Otherwise the PR looks good to me

HelpSource/Classes/PathName.schelp
prko: >This is actually not correct, given the path /foo/bar
>```
>PathName("/hello/foo").allFolders; // -> List[hello]
>PathName("/hello/foo/").allFolders; // -> List[hello, foo]
>```

I think adding a slash at the end of path should be avoided: [38f82cc](https://github.com/supercollider/supercollider/pull/7535/commits/38f82cc9be9c21176802b6bb3f51fcf737dfb813).

----

I think we could include an example code in the discussion section.
How about adding a code to test it as I have done in [2e25270](https://github.com/supercollider/supercollider/pull/7535/commits/2e252702b0945c3450de5b2df04faacc7ae4a3ed)?

Test result on my end:
Screenshot 2026-06-09 at 14 59 02

I think the following code snippet is better: 
- [27c8103](https://github.com/supercollider/supercollider/pull/7535/commits/27c810341c5da61d781fe82634f1f213972ec4d0)
- [6d5fbed](https://github.com/supercollider/supercollider/pull/7535/commits/6d5fbed3af181352cca8985d55f63bfd5f6ea4dd)
(Post window posts almost the same messages.)

HelpSource/Classes/PathName.schelp
```diff
@@ -138,27 +138,93 @@ myPath.filesDo{ |afile| afile.postln };
 ::
 
 method::allFolders
-returns a list of all the folder names contained in the pathname itself.
+Returns a list of all folder names in the pathname.
```
capital-G: I am not sure that the added text is actually that helpful.

The thing that should be communicated: This check is only based on the existence of `/` and is therefore pretty useless on unix systems, where `/foo/bar` and `/foo/bar/` are both equally valid path names [though this can be disputed, see https://stackoverflow.com/a/53874287]. I think the same even applies to windows.

Also, `PathName("foo/bar").allFolders` returns `[]` which is maybe unexpected.

IIUC, the implementation is equivalent to `"/foo/bar".split.drop(1).drop(-1);`.

One suggestion.

> Does a string comparison based on LINK::Classes/Platform#*pathSeparator::, and will return all entries which are surrounded by this separator.
> Since it solely relies on string comparison, this will not check the context of the filesystem to verify if the returned entries are actually folders.

IMO this method should be deprecated b/c its current implementation is flawed and it could only be fixed through a complete change of implementation and idea of `PathName`.

prko: I reflected your opinion in 8fe992b.

Additionally, after reading your comment on this part, I found a problem described in #7568.

I hope #6755 can be finalized and included before releasing 3.15.0.

prko: I tried to fix the inconsistency and explain the os-specific bebavoiur in [ede00b2](https://github.com/supercollider/supercollider/commit/ede00b209e19ca5602d81ca44bed9df148d6ac91).

prko:
PathName is somewhat inconvenient to use, and today I realised why I often had problems with it. I tried to fix this. I know that some core developers want to deprecate it and introduce a new class or extend the String class methods. Even after we have better alternatives, I believe these strange behaviours should either be fixed or properly documented.

HelpSource/Classes/PathName.schelp
```diff
@@ -138,30 +138,185 @@ myPath.filesDo{ |afile| afile.postln };
...
+
+~basePath = basePath /*
+File.deleteAll(~basePath) // Evaluate this to remove the files and folders created for the test. */
+)
+::
+
+method::diskName
+if path is an absolute path, returns
+list::
+## the top-level directory of the path on linux and macOS,
```
dyfer: For consistency:
```suggestion
## the top-level directory of the path on Linux and macOS,
```

dyfer:
Aside from a minor suggestion from me, do we need anything else in this PR?

HelpSource/Classes/PathName.schelp
prko: Thanks!  Done in https://github.com/supercollider/supercollider/commit/9afb125dba8cd77a9fd1806afee6878ff75e345e

HelpSource/Classes/PathName.schelp
```diff
@@ -138,30 +138,185 @@ myPath.filesDo{ |afile| afile.postln };
...
+
 code::
+/* a separator at the beginning; no separator at the end */
+
+// test with a file:
 (
-var myPath;
-myPath = PathName.new("MyDisk/SC 2.2.8 f/Sounds/FunkyChicken");
-myPath.allFolders.postln;
+var filePath = ExampleFiles.child.debug;
```
capital-G: I don't know if it is necessary to provide all the edge cases here. People should simply avoid using this method b/c it has a broken implementation.

HelpSource/Classes/PathName.schelp
```diff
@@ -138,30 +138,185 @@ myPath.filesDo{ |afile| afile.postln };
...
+// linux & macOS: excludes the top-level directory; includes the bottom-most folder
+// Windows: includes both the top-level directory and the bottom-most folder
+// See the PathName method .diskName
 )
 ::
 
-method::diskName
-if path is an absolute path, returns the disk name; else a blank string.
+discussion::
+When the path does not end with a seperator, the method handles the path consistently, regardless of whether the final component contains a file extension or not.
```
capital-G: This implementation is IMO also broken - what diskname should e.g. `/foo` yield? We shouldn't therefore document every edgecase but simply tell people to avoid using it.

SCClassLibrary/Common/Files/PathName.sc
```diff
@@ -6,7 +6,28 @@ PathName {
 	classvar <>tmp;
 
 	*new { | path = "" |
-		^super.newCopyArgs(path.standardizePath)
+		var standardPath = path.standardizePath;
```
capital-G: what is happening here? Please avoid fixing this implementation, it is broken beyond repair and we should not touch it w/o good reason to do so.

capital-G (commented):
IMO we shouldn't demonstrate the faultiness of each implementation through rather excessive examples. People who are familiar w/ path will become aware that the methods are somehow not making sense - other users should be communicated to "maybe not use this".
E.g. `diskName` can not be derived from a PathName alone in Unix, so the question is why this was added in the first place and not part of a `WindowsPathName` subclass.

And I also did not understand why the `PathName` constructor was modified.

SCClassLibrary/Common/Files/PathName.sc
prko: To resolve #7568.

prko: I reverted it.

HelpSource/Classes/PathName.schelp
prko: I followed your opinion.

HelpSource/Classes/PathName.schelp
prko: I followed your opinion.

prko:
>And I also did not understand why the PathName constructor was modified.

The modification was originally intended to make the behavior of `.allFolders` and `.diskName` more consistent.
This change has been reverted, as using `.dirname.split` provides a simpler and more predictable alternative.
The documentation has been updated to reflect this.

```supercollider
"/Imaginary/Directory/fish.rtf".dirname.split
-> [, Imaginary, Directory]

PathName("/Imaginary/Directory/fish.rtf").allFolders
-> List[Imaginary, Directory]

"Imaginary/Directory/fish.rtf".dirname.split
-> [Imaginary, Directory]

PathName("Imaginary/Directory/fish.rtf").allFolders
-> List[Directory]
```
I added some documentation to clarify this behaviour.

p.s.:
In my own work, I have not used either method at all.

prko:
If the latest commit is still not acceptable, I can take the following approaches:

```
method::allFolders
(Probably not ideal due to inconsistent behavior in modern SuperCollider.)

method::diskName
(Probably not ideal due to inconsistent behavior across modern OSes.)
```

**I leave the decision entirely to the reviewers, and I appreciate your consideration.**

prko:
@dyfer @capital-G 
I think this is ready to be merged. What are your thoughts?

JordanHendersonMusic:
I'm unsure about documenting the bad behaviour, if the user sees this in the documentation, they will expect it to never change. Documentation is a contract of stability from the user's perspective. I think it's enough to say it is depreciated, and if we must show why, this should be done in another document, perhaps a reference?

HelpSource/Classes/SoundFile.schelp
```diff
@@ -283,11 +283,11 @@ A link::Classes/String:: indicating the format of the sample data which was read
...
-## strong:: format :: || strong:: notes:: || strong::supported headers:: (partial list) 
-## "int8", "int16", "int24", "int32" || integer formats || "AIFF", "WAV", "RF64", "W64", "CAF", "FLAC" 
-## "float" || floating-point format (won't clip above 0dB) || "AIFF", "WAV", "RF64", "W64", "CAF" 
-## "mulaw", "alaw" || U-law and A-law encoding|| "WAV", "W64" 
-## "vorbis" || "Vorbis" compressed format || "OGG" 
+## strong:: format :: || strong:: notes:: || strong::supported headers:: (partial list)
+## "int8", "int16", "int24", "int32" || integer formats || "AIFF", "WAV", "RF64", "W64", "CAF", "FLAC"
+## "float" || floating-point format (won't clip above 0dB) || "AIFF", "WAV", "RF64", "W64", "CAF"
+## "mulaw", "alaw" || U-law and A-law encoding|| "WAV", "W64"
+## "vorbis" || "Vorbis" compressed format || "OGG"
```
dyfer: Could you please remove this whitespace change?

prko: Thanks, done in [2282999](https://github.com/supercollider/supercollider/pull/7535/commits/2282999862e58ef5324014e7be69b5139f3c769b)

prko:
For `.diskname`, I kept only the cautionary explanation.
For `.allFolders`, I left the alternative method chains.

HelpSource/Classes/PathName.schelp
```diff
@@ -138,30 +138,34 @@ myPath.filesDo{ |afile| afile.postln };
 ::
 
 method::allFolders
-returns a list of all the folder names contained in the pathname itself.
+This method has known reliability issues and is generally discouraged.
+Use instead link::Classes/String#-dirname:: and link::Classes/String#-split:: in combination:
 code::
```
prko: I think this alternative method chain is user-friendly, but core developers seem to not agree....

What should I remove from this PR to be merged? 
- from the line 143 to 163
- from the line 142 to 163

HelpSource/Classes/PathName.schelp
```diff
@@ -138,30 +138,34 @@ myPath.filesDo{ |afile| afile.postln };
 ::
 
 method::allFolders
-returns a list of all the folder names contained in the pathname itself.
+This method has known reliability issues and is generally discouraged.
+Use instead link::Classes/String#-dirname:: and link::Classes/String#-split:: in combination:
```
capital-G: IMO the best thing here would be to add a remark about the unexpected behavior and otherwise keep it as-is.


```suggestion
returns a list of all the folder names contained in the pathname itself.

note::
This returns different results depending on a trailing path separator, e.g. code::PathName("/foo/bar").allFolders:: and code::PathName("/foo/bar/").allFolders::, which are the same valid unix paths, return different arrays.
Use with caution.
::

```

HelpSource/Classes/PathName.schelp
```diff
@@ -138,30 +138,34 @@ myPath.filesDo{ |afile| afile.postln };
...
-)
 
-( // note the / at the start
-var myPath;
-myPath = PathName.new("/MyDisk/SC 2.2.8 f/Sounds/FunkyChicken");
-myPath.diskName.postln;
-)
-::
+This method has known reliability issues and is generally discouraged.
+It remains available for legacy use cases, and the examples below illustrate how to use it when necessary.
```
capital-G: I think it would be good to keep the existing docs (and adapt the new paths) and describe what it actually does, something along

> On Windows platforms, this will return the disk name for absolute paths.
> Otherwise, and on unix platforms, this will return an empty string.

capital-G (commented):
Sorry for my forth-and-back.


I think it is best to keep the documentation somehow minimal and precise and use examples in a similar manner.

E.g. on `PathName.allFolders` you are documenting the behavior of `dirname` and `split` as an alternative, but instead of giving 3 examples you could simply provide one, such as

```supercollider
"/foo/bar/".split(Platform.pathSeparator)
```

or

```supercollider
"/foo/bar/".split(Platform.pathSeparator).reject({|x| x.size==0})
```

I think this PR could focus on adding a remark why this may work not as expected (maybe through a minimal example) and what to do about it, maybe linking to other methods.

The other changes, i.e. `sc` files look good to me.

HelpSource/Classes/PathName.schelp
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/7535#:~:text=Update%20PathName.schelp-,6453343,-Merge%20info.

HelpSource/Classes/PathName.schelp
```diff
@@ -139,28 +139,31 @@ myPath.filesDo{ |afile| afile.postln };
...
-)
+ExampleFiles.child.split(Platform.pathSeparator).reject({|x| x.size==0})
+::
 ::
 
 method::diskName
-if path is an absolute path, returns the disk name; else a blank string.
+
+This method has known reliability issues and is generally discouraged.
+It remains available for legacy use cases, and the examples below illustrate how to use it when necessary:
```
capital-G: You should not delete the prior documentation, as now it is hidden in the examples what this will return which is the essential part of the docs. Also I wouldn't call it deprecated, the whole `PathName` class is just wonky, to say at least.

The description could be

>On Windows platforms, this will return the disk name for absolute paths.
>Otherwise, and on unix platforms, this will return an empty string.

And, if necessary, add a small example, though I think this is unnecessary

```
code::
PathName("D:/foo/bar/").diskName.cs;
PathName("/foo/bar").diskName.cs;
::
```

HelpSource/Classes/PathName.schelp
```diff
@@ -139,28 +139,31 @@ myPath.filesDo{ |afile| afile.postln };
...
+Use with caution.
+
+As an alternative, a combination of link::Classes/String#-dirname:: and link::Classes/String#-split:: may provide more predictable results:
 code::
-(
-var myPath;
-myPath = PathName.new("MyDisk/SC 2.2.8 f/Sounds/FunkyChicken");
-myPath.allFolders.postln;
-)
+ExampleFiles.child.split(Platform.pathSeparator).reject({|x| x.size==0})
```
capital-G: This returns

```
-> [Applications, SuperCollider-3.14.app, Contents, Resources, sounds, a11wlk01.wav]
```

which is wrong as `a11wlk01.wav` is not a folder! There is no way of resolving this w/o filesystem access, which `PathName` does not use.

I'd simply keep the note

> This returns different results depending on a trailing path separator, e.g. code::PathName("/foo/bar").allFolders:: and code::PathName("/foo/bar/").allFolders::, which are the same valid unix paths, return different arrays.
> Use with caution.

but don't expand further on it, e.g.

> As an alternative, a combination of link::Classes/String#-dirname:: and link::Classes/String#-split:: may provide more predictable results:

since this is also not a drop-in replacement as seen above.

capital-G (commented):
still some things.

HelpSource/Classes/PathName.schelp
prko: Thanks! Done in https://github.com/supercollider/supercollider/pull/7535/commits/2404f0e38615829216ae11d160d6a73f98d8af82

HelpSource/Classes/PathName.schelp
prko: Thanks, done in https://github.com/supercollider/supercollider/pull/7535/commits/2404f0e38615829216ae11d160d6a73f98d8af82.

---

```supercollider
ExampleFiles.child.split(Platform.pathSeparator).reject({ |x| x.size == 0 })
```

This code is the example you provided.  
The code I wrote in the previous commit was:

```supercollider
"/Imaginary/Directory/fish.rtf".dirname.split   // on Linux and macOS
```

and

```supercollider
"D:/Directory/fish.rtf".dirname.split           // on Windows
```

So, when `.dirname` is inserted into the code you suggested, the result becomes the same as expected when using `.allFolders`.

----

I do not understand why the examples use a trailing `/` at the end of the path. In the previous documentation, there were no `allFolders` examples that ended with `/`.  
I am following your suggestion since I am not familiar with conventions in other programming languages, but it is still difficult to understand the reasoning. When the path does not end with `/`, the behavior does not seem as problematic as described.

Most users rely primarily on the help documentation, so introducing an example that uses an undocumented trailing `/` may cause confusion.  
Additionally, I am not sure why Windows users should be presented with a Unix‑style pathname in this context.

capital-G: Well, `dirname` should only be applied when it is clear that one is working on a file, which is a bad assertion.

> I do not understand why the examples use a trailing / at the end of the path. In the previous documentation, there were no allFolders examples that ended with /.
> I am following your suggestion since I am not familiar with conventions in other programming languages, but it is still difficult to understand the reasoning. When the path does not end with /, the behavior does not seem as problematic as described.

`/foo/bar/` and `/foo/bar` are both valid unix path names to folders and both used equally often - one works for `.allFolders`, the other does not work.

> Additionally, I am not sure why Windows users should be presented with a Unix‑style pathname in this context.

I don't know how window paths are treated in sc - I ocassionally see `/` as path separator on windows(!) in sclang code instead of `\\`, so I don't know how paths are expressed properly.

Again, sorry for the forth-and-back - this is a broken implementation and this is the reason why I would just prefer to abandon it instead of working and documenting around its errors ;)

HelpSource/Classes/PathName.schelp
```diff
@@ -139,27 +139,36 @@ myPath.filesDo{ |afile| afile.postln };
...
 ::
 
+note::
+This returns different results depending on a trailing path separator, e.g. code::PathName("/foo/bar").allFolders:: and code::PathName("/foo/bar/").allFolders::, which are the same valid unix paths, return different arrays.
+Use with caution.
+
+As an alternative, a combination of link::Classes/String#-dirname:: and link::Classes/String#-split:: may provide more predictable results:
+code::
+ExampleFiles.child.dirname.split(Platform.pathSeparator).reject({|x| x.size==0})
+::
```
capital-G: IMO we should remove this as this is just as broken b/c this only returns a proper result if the `PathName` heads to a file and not to a folder.


```suggestion
```

HelpSource/Classes/PathName.schelp
```diff
@@ -139,27 +139,36 @@ myPath.filesDo{ |afile| afile.postln };
...
-myPath.diskName.postln;
-)
-
-( // note the / at the start
-var myPath;
-myPath = PathName.new("/MyDisk/SC 2.2.8 f/Sounds/FunkyChicken");
-myPath.diskName.postln;
+myPath = PathName.new(ExampleFiles.child);
+myPath.diskName;
 )
```
capital-G: Maybe just use

```
PathName.new(ExampleFiles.child).diskName.cs;
```

But IMO this is not clear as it could be, as `D:/foo/bar` would be a proper example, but then again, I don't know how windows path separators are handled properly in sc. But maybe we should just skip this is also broken...

prko: on mac: 
```
-> ""
```
on Windows:
```
-> "C:"
```

Done in https://github.com/supercollider/supercollider/pull/7535/commits/e1e3aa26c352a8a6f4b6f9240beefe8545f1bcde.

Thanks!

prko: I replaced `.cs` with `.asCompileString`.
See #7627.

(In my work, I only use `.cs`....)

prko:
@dyfer , @capital-G
Please let me know if there’s anything else that should be corrected.

HelpSource/Classes/PathName.schelp
```diff
@@ -19,7 +19,7 @@ method::new
 argument::path
```
JordanHendersonMusic, line 19: This file is about how to construct a path from a string, I think this change obfuscates that.

HelpSource/Classes/SoundFile.schelp
```diff
@@ -182,7 +182,7 @@ The raw array must be a link::Classes/FloatArray:: or link::Classes/Signal::, wi
 code::
 (
 f = SoundFile.new.headerFormat_("AIFF").sampleFormat_("int16").numChannels_(1);
-f.openWrite("sounds/sfwrite.aiff");
+f.openWrite("~/Downloads/sfwrite.aiff".standardizePath);
```
JordanHendersonMusic, line 185: Why downloads?

prko: `"~/Documents"` and `"~/Desktop"` can be confusing because they may refer either to a local path or to a cloud‑synchronised location.  

On macOS, these folders are often redirected to iCloud Drive when sync is enabled. For example:  
- Local path: `/Users/<username>/Documents`  
- iCloud path: `/Users/<username>/Library/Mobile Documents/com~apple~CloudDocs/Documents`  
  
On Windows, OneDrive performs a similar redirection:  
- Local path: `C:\Users\<username>\Documents`  
- Cloud path: `C:\Users\<username>\OneDrive\Documents`  
 
Because of this virtualisation, files may not physically reside on the local disk — they can be cached or downloaded on demand from the cloud. This can lead to unpredictable behaviour when writing or reading files programmatically.  
  
By contrast, `"~/Downloads"` is generally much safer, as it almost always points to a truly local directory that is not synchronised with cloud services by default.  
  
It’s also worth noting that `Platform.resourceDir` is not intended for file writing — it points to the application’s resource directory, which may be read‑only or protected by the operating system.  
  
For clarity and reliability, it’s therefore better to use an explicitly local directory (for example, `"~/Downloads"` or a dedicated project folder) rather than relying on `"~/Documents"`, `"~/Desktop"`, or `Platform.resourceDir`.

HelpSource/Classes/PathName.schelp
prko: I understand your point that this section is meant to demonstrate how to construct a path from a string.

However, the previous literal string — **`"MyDisk/SC 2.2.8 f/Sounds/FunkyChicken"`** — is based on a legacy classic Mac OS path format from the early 2000s. This format is highly unfamiliar and can be very confusing for modern users, especially those on Windows.

I believe it is much more helpful for users to see an example that reflects the actual path conventions of their own operating system. By using **`ExampleFiles.child`**, the documentation dynamically generates a valid, OS-appropriate path format for each user.

This approach helps them clearly understand how paths are constructed on their specific platform, rather than struggling to interpret an outdated, platform-specific string. I hope this clarifies the intention behind the change!

HelpSource/Classes/SoundFile.schelp
prko: @JordanHendersonMusic 
Would you mind suggesting an alternative folder that aligns with the points I mentioned, if `~/Downloads` isn’t appropriate?

HelpSource/Classes/PathName.schelp
prko: @JordanHendersonMusic 

If my explanation wasn’t convincing enough, please let me know how I can clarify it further.

PR #7537, opened 2026-06-01
cmake: don't install utf8proc

dyfer:
## Purpose and Motivation

The recently added library `utf8proc` was running its own installation when building the `install` target of SC, which resulted in utf8proc library being copied to the installation destination. This PR disables the library's installation.

Note: if we need to add libraries and manipulate their flags, we tend to put them in a subdirectory, like with [portaudio](https://github.com/supercollider/supercollider/tree/develop/external_libraries/portaudio). I haven't moved the repo in this case, since it sometimes messes up people's local repositories (one needs to resync submodules or something like that), but if we feel that is warranted, I can also do it in this PR.

## Types of changes

- Bug fix (kind of)

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/external_libraries/CMakeLists.txt
+++ b/external_libraries/CMakeLists.txt
@@ -27,6 +27,7 @@ if(MSVC) # hide warnings from 3rd party libraries
 	)
 endif()
 
+set(UTF8PROC_INSTALL OFF CACHE BOOL "Install utf8proc" FORCE)
 add_subdirectory(utf8proc)
 
 set_property(TARGET utf8proc PROPERTY FOLDER 3rdparty)


JordanHendersonMusic:
Ah didn't know this was happening! Thanks! I thought I'd added the library correctly? If you want it move it that's fine with me.

JordanHendersonMusic (approved):

dyfer:
>  I thought I'd added the library correctly? 

Yes, I think so.
But also the library will run whatever it wants in its cmake script. Typically there's an option to disable the install step exactly for the situation we have here.

I'm not sure it's worth it to move the library - the "issue" is that we set `UTF8PROC_INSTALL` variable for all remaining libraries in `external_libraries` - but it's unlikely that name is used elsewhere. @capital-G do you have an opinion one way or another?

capital-G:
> I'm not sure it's worth it to move the library - the "issue" is that we set `UTF8PROC_INSTALL` variable for all remaining libraries in `external_libraries` - but it's unlikely that name is used elsewhere. @capital-G do you have an opinion one way or another?

So the only reason to have a sub-directory within the sub-directory is to have isolated vars b/c of folder boundaries? Sonuds a bit hacky to me^^ But IMO we are fine as-is - once it creates any trouble, we can always adjust, but it doesn't seem likely [famous last words].

Feel free to merge!

dyfer:
> So the only reason to have a sub-directory within the sub-directory is to have isolated vars b/c of folder boundaries? Sonuds a bit hacky to me^^

Yeah, I don't know if there's a better way to handle this. Essentially I don't think there's any other way to override the settings of the library than force a cache var.

PR #7540, opened 2026-06-03
Update build_linux.yml: Add Ubuntu ARM64 artifact upload

prko:
## Purpose and Motivation

This PR updates `build_linux.yml` to include uploading of Ubuntu ARM64 build artefacts.  
The workflow already builds ARM64 binaries, but the resulting artefacts were not being uploaded, unlike the x86_64 builds.

This PR is AI‑assisted. Based on the following comment, I believe this change is logical, and I have tested that the resulting build launches correctly on my Ubuntu system:  
[https://github.com/supercollider/supercollider/blob/926827c607415e8c4333b6dfe891a3cd64038818/.github/workflows/build_linux.yml#L39](https://github.com/supercollider/supercollider/blob/926827c607415e8c4333b6dfe891a3cd64038818/.github/workflows/build_linux.yml#L39)

The only issue I currently have is described in the form of a question here:  
[https://scsynth.org/t/how-can-i-start-sc-ide-from-the-terminal-with-a-customised-class-library-location-and-customised-platform-resourcedir/13348?u=prko](https://scsynth.org/t/how-can-i-start-sc-ide-from-the-terminal-with-a-customised-class-library-location-and-customised-platform-resourcedir/13348?u=prko)  
However, I believe this PR itself is correct.

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -102,6 +102,7 @@ jobs:
             use-syslibs: false
             shared-libscsynth: false
             ctest: true
+            artifact-suffix: "ubuntu-24.04-arm-clang18"
 
     name: Ubuntu ${{ matrix.os-version }} ${{ matrix.c-compiler }} ${{ matrix.job-name }}
     env:


dyfer:
Hi @prko, the change is fine, technically. But it would be good to test if this artifact is actually usable. 

~~Have you tried downloading the artifact and running it locally?~~ EDIT: the PR description suggests that you did, in that case this is fine.

Note: We upload the other linux artifact because we use it to run tests, I'm not sure if it's ever used by users to run locally (also see my reply on the forum about Linux packages and standalone builds).

prko:
Last week, I downloaded an Ubuntu artifact on Parallels Desktop running on an Intel Mac.
I was able to launch scide by double‑clicking it, and I assumed it was working. (I don’t recall which PR it was.) At the time, I didn’t realise that scide was running based on `/usr/local/share/SuperCollider`.

I’ll use the Linux build only to check whether the tests pass.
For actual use, I’ll build it myself.

However, it would be great if an AppImage were supported for Linux.

dyfer:
Thanks. Sorry if I'm not understanding correctly - does it mean you still need the arm64 artifact to be uploaded or should this be closed?

EDIT and yes AppImage would be nice. We discussed that at some point but I think there were some issues... we may revisit that.

prko:
This can be closed! Thanks!

PR #7534, opened 2026-06-01
Fix keyboard shortcuts and font reset behavior in HelpBrowser

prko:
closes: #7426, #7130, #6820, #6629

**NOTE:**
This PR was assisted by GitHub Copilot.

## Purpose and Motivation

This PR fixes the following:

- **Resetting font size in HelpBrowser**  
  - Before:  
    Pressing `Cmd + 0` (macOS) or `Ctrl + 0` (Windows) reduced the font size instead of resetting it.  
  - With this PR:  
    Pressing `Cmd + 0` (macOS) or `Ctrl + 0` (Windows) correctly resets the font size.

- **Keyboard shortcuts in code editors in HelpBrowser on macOS**  
  The following shortcuts now work as expected:  
  - `Cmd + B`  
  - `Cmd + M`  
  - `Cmd + Shift + M`  
  - `Cmd + Alt + M`  
  - `Cmd + T`  
  - `Cmd + Shift + T`  
  - `Cmd + Alt + T`  
  - `Cmd + I`  
  - `Cmd + Shift + P`  
  - `Cmd + Shift + L`  
  - `Cmd + D`

The following issues are **not** fixed by this PR:

- On macOS, `Cmd + R` in the HelpBrowser code editors does not refresh the document. 
- On macOS, `Cmd + 0`, `Cmd + =`, and `Cmd + -` do not change the font size inside the code editor areas.  
  (These shortcuts work correctly outside the code editor.)

These issues should be addressed in a separate PR.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] This PR is ready for review

PR #7521, opened 2026-05-28
[draft and test using AI] Enable keyboard shortcuts in the code editor in HelpBrowser

prko:
closes: #7130 #6820 #6629

## Purpose and Motivation
Some keyboard shortcuts in the code editor in HelpBrowser do not work.
I asked this first to MS Copilot, then asked it again using the answer from MS Copilot to GitHub Copilot.
I am uploading this to test the artefacts for Windows, macOS and Ubuntu.

## Types of changes

- Documentation
- Bug fix
- New feature
- Breaking change

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation
- [ ] This PR is ready for review

PR #7542, opened 2026-06-04
Fix HelpBrowser key shortcuts

### LINKED ISSUE #7539: Fix inconsistent HelpBrowser keyboard shortcuts across platforms
## Environment

* SuperCollider version: 3.15.dev
* Operating system: all

## Steps to reproduce
1. Launch SC  
2. Open the HelpBrowser and click **Search**  
3. Type **SinOsc**  
4. Click **SinOsc**  
5. Try the following key shortcuts both **inside** the code editor in HelpBrowser and **outside** the code editor in HelpBrowser:
   - `Cmd/Ctrl + R`
   - `Cmd/Ctrl + =`
   - `Cmd/Ctrl + +`
   - `Cmd/Ctrl + -`
   - `Cmd/Ctrl + 0`

### Expected behaviour
- **Cmd/Ctrl + R**
  - **Linux & macOS:** Refresh HelpBrowser (inside/outside code editor)  
  - **Windows:** No action  
- **Cmd/Ctrl + =**
  - **All platforms:** Increase font size (inside/outside code editor)  
- **Cmd/Ctrl + +**
  - **All platforms:** Increase font size (inside/outside code editor)  
- **Cmd/Ctrl + 0** and **Cmd/Ctrl + -**
  - **All platforms:** Change font size **inside/outside** the code editor in HelpBrowser  

### Actual behaviour
- **Cmd/Ctrl + R**
  - **Linux (Ubuntu tested):** Refresh HelpBrowser (inside/outside code editor)  
  - **macOS:** Refresh HelpBrowser only outside the code editor (**It should also work inside the code editor**)
  - **Windows:** Triggers “revoke find and replace” in the IDE editor window (**no action expected**)
- **Cmd/Ctrl + =**
  - **macOS:** Increases font size  
  - **Linux & Windows:** No action  
  **Note:** In web browsers on Windows and Ubuntu, `Ctrl + =` *does* increase font size. There is no reason for SC HelpBrowser to accept this shortcut only on macOS.  
- **Cmd/Ctrl + +**, **Cmd/Ctrl + 0** and **Cmd/Ctrl + -**
  - **macOS:** Changes font size only outside the code editor (**It should also work inside the code editor**)
  - **Linux & Windows:** Changes font size **inside/outside** the code editor

prko:
For the current state of this PR, see https://github.com/supercollider/supercollider/pull/7542#issuecomment-4639939418.

---

closes: #7539

Modifying `HelpBrowser.cpp` and `overriding_action.hpp` in this PR was assisted by some AIs, and I reviewed it based on the experience from #7534.

## Purpose and Motivation
This PR fixes most of the inconsistent behaviours described in #7539.

- **Cmd/Ctrl + R**
  - **Linux & macOS:** Refreshes the HelpBrowser (both inside and outside the code editor)
  - **Windows:** No action

- **Cmd/Ctrl + =**, **Cmd/Ctrl + 0**, **Cmd/Ctrl + -**  
  - **All platforms:** Adjust font size (both inside and outside the code editor)

**Additionally, navigating to the previous and next pages now works consistently across all platforms, both inside and outside the code editor.**

## Types of changes
- Bug fix

## To‑do list
- [x] Code is tested  
- [x] All tests are passing  

### DIFF
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -24,7 +24,7 @@ table::
     ## soft::G:: || go to bottom 
     ## soft::g:: || go to top
     ## soft::shift + j:: or soft::ctrl + minus:: || zoom out
-    ## soft::shift + k:: or soft::ctrl + equal/plus:: || zoom in
+    ## soft::shift + k:: or soft::ctrl + equal:: || zoom in
     ## soft::/:: or soft::ctrl + f:: || search in page
     ## soft::F3:: || open Search page
     ## soft::F5:: || reload page
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -21,10 +21,10 @@ table::
     ## soft::ctrl + u:: || scroll more lines up
     ## soft::h:: or soft::alt + left arrow:: || go back
     ## soft::l:: or soft::alt + right arrow:: || go forward
-    ## soft::G:: || go to bottom 
+    ## soft::G:: || go to bottom
     ## soft::g:: || go to top
     ## soft::shift + j:: or soft::ctrl + minus:: || zoom out
-    ## soft::shift + k:: or soft::ctrl + plus:: || zoom in
+    ## soft::shift + k:: or soft::ctrl + equal/plus: || zoom in
     ## soft::/:: or soft::ctrl + f:: || search in page
     ## soft::F3:: || open Search page
     ## soft::F5:: || reload page
--- a/HelpSource/Classes/HelpBrowser.schelp
+++ b/HelpSource/Classes/HelpBrowser.schelp
@@ -21,10 +21,10 @@ table::
     ## soft::ctrl + u:: || scroll more lines up
     ## soft::h:: or soft::alt + left arrow:: || go back
     ## soft::l:: or soft::alt + right arrow:: || go forward
-    ## soft::G:: || go to bottom
+    ## soft::G:: || go to bottom 
     ## soft::g:: || go to top
     ## soft::shift + j:: or soft::ctrl + minus:: || zoom out
-    ## soft::shift + k:: or soft::ctrl + equal/plus: || zoom in
+    ## soft::shift + k:: or soft::ctrl + equal/plus:: || zoom in
     ## soft::/:: or soft::ctrl + f:: || search in page
     ## soft::F3:: || open Search page
     ## soft::F5:: || reload page
--- a/editors/sc-ide/core/util/overriding_action.hpp
+++ b/editors/sc-ide/core/util/overriding_action.hpp
@@ -48,9 +48,7 @@ public:
         if (event->key() >= Qt::Key_Shift && event->key() <= Qt::Key_Alt)
             return QKeySequence();
 
-        Qt::KeyboardModifiers modifiers = event->modifiers() & ~Qt::KeypadModifier;
-
-        return QKeySequence(modifiers | event->key());
+        return QKeySequence(event->modifiers() | event->key());
     }
 
 protected:
--- a/editors/sc-ide/core/util/overriding_action.hpp
+++ b/editors/sc-ide/core/util/overriding_action.hpp
@@ -48,6 +48,8 @@ public:
         if (event->key() >= Qt::Key_Shift && event->key() <= Qt::Key_Alt)
             return QKeySequence();
 
+        Qt::KeyboardModifiers modifiers = event->modifiers() & ~Qt::KeypadModifier;
+
         return QKeySequence(event->modifiers() | event->key());
     }
 
--- a/editors/sc-ide/core/util/overriding_action.hpp
+++ b/editors/sc-ide/core/util/overriding_action.hpp
@@ -50,7 +50,7 @@ public:
 
         Qt::KeyboardModifiers modifiers = event->modifiers() & ~Qt::KeypadModifier;
 
-        return QKeySequence(event->modifiers() | event->key());
+        return QKeySequence(modifiers | event->key());
     }
 
 protected:
--- a/editors/sc-ide/core/util/overriding_action.hpp
+++ b/editors/sc-ide/core/util/overriding_action.hpp
@@ -48,8 +48,6 @@ public:
         if (event->key() >= Qt::Key_Shift && event->key() <= Qt::Key_Alt)
             return QKeySequence();
 
-        Qt::KeyboardModifiers modifiers = event->modifiers() & ~Qt::KeypadModifier;
-
         return QKeySequence(event->modifiers() | event->key());
     }
 
--- a/editors/sc-ide/core/util/overriding_action.hpp
+++ b/editors/sc-ide/core/util/overriding_action.hpp
@@ -29,14 +29,14 @@ namespace ScIDE {
 
 class OverridingAction : public QAction {
 public:
-    OverridingAction(QObject* parent): QAction(parent) { setShortcutContext(Qt::WidgetWithChildrenShortcut); }
+    OverridingAction(QObject* parent): QAction(parent) { setShortcutContext(Qt::WindowShortcut); }
 
     OverridingAction(const QString& text, QObject* parent): QAction(text, parent) {
-        setShortcutContext(Qt::WidgetWithChildrenShortcut);
+        setShortcutContext(Qt::WindowShortcut);
     }
 
     OverridingAction(const QIcon& icon, const QString& text, QObject* parent): QAction(icon, text, parent) {
-        setShortcutContext(Qt::WidgetWithChildrenShortcut);
+        setShortcutContext(Qt::WindowShortcut);
     }
 
     void addToWidget(QWidget* widget) {
--- a/editors/sc-ide/core/util/overriding_action.hpp
+++ b/editors/sc-ide/core/util/overriding_action.hpp
@@ -29,14 +29,14 @@ namespace ScIDE {
 
 class OverridingAction : public QAction {
 public:
-    OverridingAction(QObject* parent): QAction(parent) { setShortcutContext(Qt::WindowShortcut); }
+    OverridingAction(QObject* parent): QAction(parent) { setShortcutContext(Qt::WidgetWithChildrenShortcut); }
 
     OverridingAction(const QString& text, QObject* parent): QAction(text, parent) {
-        setShortcutContext(Qt::WindowShortcut);
+        setShortcutContext(Qt::WidgetWithChildrenShortcut);
     }
 
     OverridingAction(const QIcon& icon, const QString& text, QObject* parent): QAction(icon, text, parent) {
-        setShortcutContext(Qt::WindowShortcut);
+        setShortcutContext(Qt::WidgetWithChildrenShortcut);
     }
 
     void addToWidget(QWidget* widget) {
--- a/editors/sc-ide/widgets/help_browser.cpp
+++ b/editors/sc-ide/widgets/help_browser.cpp
@@ -334,21 +334,6 @@ bool HelpBrowser::eventFilter(QObject* object, QEvent* event) {
             }
             break;
         }
-        case QEvent::ShortcutOverride: {
-            // check if any widget action shortcut matches the observed keyEvent
-            // if yes, capture the event, else, bubble up the event
-            auto keyEvent = static_cast<QKeyEvent*>(event);
-
-            auto sequence = OverridingAction::keySequence(keyEvent);
-
-            for (int i = 0; i < ActionCount; ++i) {
-                if (mActions[i] && mActions[i]->shortcuts().contains(sequence)) {
-                    event->accept();
-                    return true;
-                }
-            }
-            break;
-        }
         default:
             break;
         }
--- a/editors/sc-ide/widgets/help_browser.cpp
+++ b/editors/sc-ide/widgets/help_browser.cpp
@@ -132,11 +132,15 @@ HelpBrowser::HelpBrowser(QWidget* parent): QWidget(parent) {
 
 void HelpBrowser::onPageLoad() {
     mLoadProgressIndicator->stop();
+    // add these actions to weview's renderer, to capture shift+enter and possibly other swallowed shortcuts
     static_cast<OverridingAction*>(mActions[EvaluateRegion])->addToWidget(mWebView->focusProxy());
     static_cast<OverridingAction*>(mActions[Evaluate])->addToWidget(mWebView->focusProxy());
-    if (mWebView && mWebView->focusProxy()) {
-        mWebView->focusProxy()->installEventFilter(this);
-    }
+    static_cast<OverridingAction*>(mActions[ZoomIn])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[ZoomOut])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[ResetZoom])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Reload])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Back])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Forward])->addToWidget(mWebView->focusProxy());
 }
 
 void HelpBrowser::createActions() {
@@ -172,9 +176,21 @@ void HelpBrowser::createActions() {
     mWebView->pageAction(QWebEnginePage::Copy)->setShortcut(QKeySequence::Copy);
     mWebView->pageAction(QWebEnginePage::Paste)->setShortcut(QKeySequence::Paste);
 
-    mActions[Back] = mWebView->pageAction(QWebEnginePage::Back);
-    mActions[Forward] = mWebView->pageAction(QWebEnginePage::Forward);
-    mActions[Reload] = mWebView->pageAction(QWebEnginePage::Reload);
+    // proxy page actions to avoid shortcuts conflicts with main window
+    // note that we assign shortcuts here as they don't depend on IDE settings
+    auto proxyPageAction = [this](QAction* pageAction) {
+        // OverridingAction limits shortcut context to this widget
+        auto ovrAction = new OverridingAction(pageAction->icon(), pageAction->text(), this);
+        connect(ovrAction, &OverridingAction::triggered, pageAction, &QAction::trigger);
+        // disable pageAction shortcut and assign it to ovrAction instead
+        ovrAction->setShortcut(pageAction->shortcut());
+        pageAction->setShortcut(QKeySequence());
+        ovrAction->addToWidget(this);
+        return ovrAction;
+    };
+    mActions[Back] = proxyPageAction(mWebView->pageAction(QWebEnginePage::Back));
+    mActions[Forward] = proxyPageAction(mWebView->pageAction(QWebEnginePage::Forward));
+    mActions[Reload] = proxyPageAction(mWebView->pageAction(QWebEnginePage::Reload));
 }
 
 void HelpBrowser::applySettings(Settings::Manager* settings) {
@@ -297,51 +313,8 @@ bool HelpBrowser::helpBrowserHasFocus() const {
 }
 
 bool HelpBrowser::eventFilter(QObject* object, QEvent* event) {
-    if (object == mWebView || (mWebView && object == mWebView->focusProxy())) {
+    if (object == mWebView) {
         switch (event->type()) {
-        case QEvent::ShortcutOverride: {
-            auto keyEvent = static_cast<QKeyEvent*>(event);
-
-#    ifdef Q_OS_WIN
-            if (keyEvent->key() == Qt::Key_R && (keyEvent->modifiers() & Qt::ControlModifier)) {
-                event->accept();
-                if (mActions[Reload])
-                    mActions[Reload]->trigger();
-                return true;
-            }
-#    endif
-
-#    ifdef Q_OS_MAC
-            if ((keyEvent->key() == Qt::Key_Plus || keyEvent->key() == Qt::Key_Equal)
-                && (keyEvent->modifiers() & Qt::ControlModifier)) {
-                event->accept();
-                if (mActions[ZoomIn])
-                    mActions[ZoomIn]->trigger();
-                return true;
-            }
-#    endif
-            if (keyEvent->key() == Qt::Key_Minus && (keyEvent->modifiers() & Qt::ControlModifier)) {
-                event->accept();
-                if (mActions[ZoomOut])
-                    mActions[ZoomOut]->trigger();
-                return true;
-            }
-            if (keyEvent->key() == Qt::Key_0 && (keyEvent->modifiers() & Qt::ControlModifier)) {
-                event->accept();
-                if (mActions[ResetZoom])
-                    mActions[ResetZoom]->trigger();
-                return true;
-            }
-            auto sequence = OverridingAction::keySequence(keyEvent);
-            for (int i = 0; i < ActionCount; ++i) {
-                if (mActions[i] && mActions[i]->shortcuts().contains(sequence)) {
-                    event->accept();
-                    return true;
-                }
-            }
-            break;
-        }
-
         case QEvent::MouseButtonPress: {
             QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
             switch (mouseEvent->button()) {
@@ -353,6 +326,21 @@ bool HelpBrowser::eventFilter(QObject* object, QEvent* event) {
                 return true;
             default:
                 break;
+            }
+            break;
+        }
+        case QEvent::ShortcutOverride: {
+            // check if any widget action shortcut matches the observed keyEvent
+            // if yes, capture the event, else, bubble up the event
+            auto keyEvent = static_cast<QKeyEvent*>(event);
+
+            auto sequence = OverridingAction::keySequence(keyEvent);
+
+            for (int i = 0; i < ActionCount; ++i) {
+                if (mActions[i] && mActions[i]->shortcuts().contains(sequence)) {
+                    event->accept();
+                    return true;
+                }
             }
             break;
         }
--- a/editors/sc-ide/widgets/help_browser.cpp
+++ b/editors/sc-ide/widgets/help_browser.cpp
@@ -63,6 +63,10 @@ using namespace QtCollider;
 #    ifdef Q_OS_WIN
 QAction* HelpBrowser::winCtrlRBlocker = nullptr;
 #    endif
+#    ifdef Q_OS_MAC
+QAction* HelpBrowser::macCmdPlusBlocker = nullptr;
+// QAction* HelpBrowser::macCmdEqualBlocker = nullptr;
+#    endif
 
 HelpBrowser::HelpBrowser(QWidget* parent): QWidget(parent) {
     QRect availableScreenRect = qApp->primaryScreen()->availableGeometry();
@@ -146,10 +150,6 @@ void HelpBrowser::onPageLoad() {
     static_cast<OverridingAction*>(mActions[Back])->addToWidget(mWebView->focusProxy());
     static_cast<OverridingAction*>(mActions[Forward])->addToWidget(mWebView->focusProxy());
 
-    // mActions[ZoomIn]->setShortcutContext(Qt::WindowShortcut);
-    // mActions[ZoomOut]->setShortcutContext(Qt::WindowShortcut);
-    // mActions[ResetZoom]->setShortcutContext(Qt::WindowShortcut);
-
 #    ifdef Q_OS_WIN
     if (!winCtrlRBlocker) {
         winCtrlRBlocker = new OverridingAction(this);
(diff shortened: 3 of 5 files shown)


prko:
With [Update overriding_action.hpp](https://github.com/supercollider/supercollider/pull/7542/commits/b4ac8c77fd21deec16d8e2563387aeb31040b1ad), “Ctrl + =” and “Ctrl + +” enlarge the font size on both Ubuntu and Windows.
On macOS, “Cmd + =” enlarges the HelpBrowser font size, while “Cmd + +” enlarges the SC‑IDE code editor font size. I think this behavior is acceptable.

I will try blocking “Cmd + +” in the HelpBrowser so that it no longer enlarges the SC‑IDE code editor font size.

prko:
I am giving up on resolving the issue where `Cmd + +` in the HelpBrowser enlarges the font size in the SC‑IDE code editor.  
If other developers also cannot find any clues after  https://github.com/supercollider/supercollider/pull/7542/commits/b4ac8c77fd21deec16d8e2563387aeb31040b1ad,   https://github.com/supercollider/supercollider/pull/7542/commits/b4ac8c77fd21deec16d8e2563387aeb31040b1ad should be the final one. Please let me know.

For `help_browser.cpp` and `overriding_action.hpp`, I followed the AI’s guidance and applied the necessary changes, removing anything that seemed unnecessary. At this point, reviewers should check whether the code is clean and ready to be considered final.

HelpSource/Classes/HelpBrowser.schelp
```diff
@@ -21,10 +21,10 @@ table::
     ## soft::ctrl + u:: || scroll more lines up
     ## soft::h:: or soft::alt + left arrow:: || go back
     ## soft::l:: or soft::alt + right arrow:: || go forward
-    ## soft::G:: || go to bottom 
+    ## soft::G:: || go to bottom
```
dyfer: Unnecessary whitespace change, can you remove this?

HelpSource/Classes/HelpBrowser.schelp
```diff
@@ -21,10 +21,10 @@ table::
...
     ## soft::g:: || go to top
     ## soft::shift + j:: or soft::ctrl + minus:: || zoom out
-    ## soft::shift + k:: or soft::ctrl + plus:: || zoom in
+    ## soft::shift + k:: or soft::ctrl + equal/plus: || zoom in
```
dyfer: I think there's one missing `:` after "plus"?

dyfer (dismissed):
I don't know about the Qt changes, just caught a couple of cosmetic ones.

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -135,6 +135,26 @@ void HelpBrowser::onPageLoad() {
...
+    static_cast<OverridingAction*>(mActions[Reload])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Back])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Forward])->addToWidget(mWebView->focusProxy());
+
+    mActions[ZoomIn]->setShortcutContext(Qt::WindowShortcut);
+    mActions[ZoomOut]->setShortcutContext(Qt::WindowShortcut);
+    mActions[ResetZoom]->setShortcutContext(Qt::WindowShortcut);
+
+#    ifdef Q_OS_WIN
+    static QAction* winCtrlRBlocker = nullptr;
```
capital-G: Can you make this a private static member variable and add some small docstring/explanation, why we need this blocker variable?

I am no C++ expert, but a static variable within a function looks somehow off to me, but maybe this is common practice after all?

editors/sc-ide/core/util/overriding_action.hpp
```diff
@@ -48,6 +48,8 @@ class OverridingAction : public QAction {
         if (event->key() >= Qt::Key_Shift && event->key() <= Qt::Key_Alt)
             return QKeySequence();
 
+        Qt::KeyboardModifiers modifiers = event->modifiers() & ~Qt::KeypadModifier;
```
capital-G: This looks like an unused variable

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -191,12 +211,31 @@ void HelpBrowser::applySettings(Settings::Manager* settings) {
 
     mActions[DocClose]->setShortcut(settings->shortcut("ide-document-close"));
 
-    mActions[ZoomIn]->setShortcut(settings->shortcut("editor-enlarge-font"));
+    QList<QKeySequence> zoomInShortcuts;
+    zoomInShortcuts.append(settings->shortcut("editor-enlarge-font"));
 
-    mActions[ZoomOut]->setShortcut(settings->shortcut("editor-shrink-font"));
+#    ifdef Q_OS_MAC
+    zoomInShortcuts.append(QKeySequence(Qt::META | Qt::Key_Equal)); // Cmd + =
```
capital-G: Wouldn't this be a "us-only" keyboard shortcut (b/c capslocked `=` is `+` on us layout iirc?) - I don't know what the common approach is here, maybe the Qt docs can provide some best practices for this?

You could reduce this platform-separation code by declaring something like a `mMetaKey` member variable, which for macOS results in `Qt::Meta` and for other to `Qt::CTRL`.

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -191,12 +211,31 @@ void HelpBrowser::applySettings(Settings::Manager* settings) {
...
+    zoomInShortcuts.append(QKeySequence(Qt::META | Qt::SHIFT | Qt::Key_Equal)); // Cmd + Shift + =  → +
+    zoomInShortcuts.append(QKeySequence(Qt::META | Qt::Key_Plus)); // Cmd + + (keypad)
+    zoomInShortcuts.append(QKeySequence(Qt::META | Qt::SHIFT | Qt::Key_Plus)); // Cmd + Shift + +
+    zoomInShortcuts.append(QKeySequence("Meta++"));
```
capital-G: How is this different from  `zoomInShortcuts.append(QKeySequence(Qt::META | Qt::Key_Plus));`?

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -191,12 +211,31 @@ void HelpBrowser::applySettings(Settings::Manager* settings) {
 
     mActions[DocClose]->setShortcut(settings->shortcut("ide-document-close"));
 
-    mActions[ZoomIn]->setShortcut(settings->shortcut("editor-enlarge-font"));
+    QList<QKeySequence> zoomInShortcuts;
+    zoomInShortcuts.append(settings->shortcut("editor-enlarge-font"));
```
capital-G: IMO pressing `Cmd +` on the help browser shouldn't affect the editor font? I would it expect to zoom in on the layout of the help browser?

capital-G (changes requested):
> Adjust font size (both inside and outside the code editor)

IMO zooming on the help browser should be independent of the editor, since the help browser also has text which the text editor can not display (i.e. non-monospaced text). I may want to therefore change the layout w/o changing my own written code.

editors/sc-ide/widgets/help_browser.cpp
prko: I’d also like to hear your thoughts on this. To clarify the current situation: the behaviour is already correct on Windows and Linux. This inconsistent behaviour only occurs on macOS, and it is *not* introduced by this PR.

To avoid confusion, let's define the terms:
- **SC‑IDE Code Editor** → the main code editor in SC‑IDE  
- **HelpBrowser codeMirrorContainer** → the `<div class="codeMirrorContainer">` inside the HelpBrowser HTML

Here is how macOS behaves across different versions:

---

## **SC 3.14 to current (without this PR)**

### When clicking outside `codeMirrorContainer`:
- `Cmd +` → increases font size in the **SC‑IDE Code Editor** - `Cmd =` → increases font size in the **HelpBrowser**

### When clicking inside `codeMirrorContainer`:
- `Cmd +` → increases font size in the **SC‑IDE Code Editor** - `Cmd =` → **does nothing**

---

## **SC 3.13 (and earlier) / With this PR applied**

### When clicking outside `codeMirrorContainer`:
- `Cmd +` → increases font size in the **SC‑IDE Code Editor**
- `Cmd =` → increases font size in the **HelpBrowser**

### When clicking inside `codeMirrorContainer`:
- `Cmd +` → increases font size in the **SC‑IDE Code Editor** - `Cmd =` → increases font size in the **HelpBrowser**

---

## **Summary**

The issue you pointed out (`Cmd +` enlarges SC-IDE code editor font) is an older, pre-existing macOS‑specific problem. This PR simply fixes the regression introduced in SC 3.14 and restores the expected behaviour from SC 3.13 (where `Cmd =` at least works consistently inside the `HelpBrowser codeMirrorContainer`). 

I did initially spend several hours investigating how to prevent `Cmd +` in the HelpBrowser from expanding the code editor text—not only in this PR, but also during my work on #7534 and #7521. However, after exploring various approaches, I realised that a proper fix requires a much deeper architectural change. Therefore, I have left it out of scope for this PR so we can focus on resolving the 3.14 regression first.

editors/sc-ide/widgets/help_browser.cpp
prko: I tried it already before: See Line 219 in [0a33968](https://github.com/supercollider/supercollider/commit/0a3396876cc2f0534e78d68d83142a79a1804ef1)

editors/sc-ide/widgets/help_browser.cpp
prko: >Wouldn't this be a "us-only" keyboard shortcut (b/c capslocked = is + on us layout iirc?) 

Yes, I’m aware of that. I added this extra shortcut because `Cmd +` does not work on my device.

>You could reduce this platform-separation code by declaring something like a mMetaKey member variable, which for macOS results in Qt::Meta and for other to Qt::CTRL.

Yes, I know.
AI suggested this after I repeatedly asked for a cleaner solution.
I left it in place only to show what the AI suggested and what I tried while solving the issue.

HelpSource/Classes/HelpBrowser.schelp
prko: Thank you! Done in [af4dec5](https://github.com/supercollider/supercollider/pull/7542/commits/af4dec5c6e7f3db4c0af0ada6139a047e2e0a94b)

HelpSource/Classes/HelpBrowser.schelp
prko: Thank you! Done in [af4dec5](https://github.com/supercollider/supercollider/pull/7542/commits/af4dec5c6e7f3db4c0af0ada6139a047e2e0a94b)

editors/sc-ide/widgets/help_browser.cpp
prko: Thank you! Done in [af4dec5](https://github.com/supercollider/supercollider/pull/7542/commits/af4dec5c6e7f3db4c0af0ada6139a047e2e0a94b)

editors/sc-ide/core/util/overriding_action.hpp
prko: Thank you! Done in [af4dec5](https://github.com/supercollider/supercollider/pull/7542/commits/af4dec5c6e7f3db4c0af0ada6139a047e2e0a94b)

editors/sc-ide/widgets/help_browser.cpp
prko: Thank you! Done in [af4dec5](https://github.com/supercollider/supercollider/pull/7542/commits/af4dec5c6e7f3db4c0af0ada6139a047e2e0a94b)

prko:
I’ve spent an unexpectedly long time (including today, yesterday, and several earlier attempts) looking deeply into the keyboard‑shortcut issue in the HelpBrowser. Unfortunately, handling `"Cmd" + "Shift" + "="` on the English keyboard layout has turned out to be more complicated than I anticipated, and I haven’t yet been able to find a reliable solution.

Since the `"Cmd" + "Shift" + "="` issue on the English layout is an older, pre‑existing bug rather than something introduced by this PR, I’d like to suggest leaving it outside the scope of this pull request. I think it would be more practical to address it in a separate issue or PR, so that it doesn’t hold up the current platform‑inconsistency fixes.

editors/sc-ide/widgets/help_browser.hpp
```diff
@@ -152,6 +152,19 @@ private slots:
     QAction* mActions[ActionCount];
 
     int mServerPort = 0; // if 0, server is not running
+
+#ifdef Q_OS_WIN
+    // Blocks the default `Ctrl + R` behavior on Windows, which would otherwise
+    // trigger an unwanted refresh action in the embedded browser widget.
+    static QAction* winCtrlRBlocker;
```
capital-G: If you want to stop key events from bubbling up, you should use this pattern

https://github.com/supercollider/supercollider/blob/926827c607415e8c4333b6dfe891a3cd64038818/editors/sc-ide/widgets/lookup_dialog.cpp#L202-L215

You can read the docs about this at https://doc.qt.io/qt-6/eventsandfilters.html#event-filters

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -60,6 +60,14 @@ namespace ScIDE {
 
 using namespace QtCollider;
 
+#    ifdef Q_OS_WIN
+QAction* HelpBrowser::winCtrlRBlocker = nullptr;
+#    endif
+#    ifdef Q_OS_MAC
+QAction* HelpBrowser::macCmdPlusBlocker = nullptr; // does not work
```
capital-G: if it doesn't work, why do we set it below or why does this even exist then?

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -60,6 +60,14 @@ namespace ScIDE {
...
+QAction* HelpBrowser::macCmdEqualBlocker = nullptr; // does not work
+#    endif
```
capital-G: These should be set to nullpointers in the header file.

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -135,6 +143,44 @@ void HelpBrowser::onPageLoad() {
...
     static_cast<OverridingAction*>(mActions[Evaluate])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[ZoomIn])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[ZoomOut])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[ResetZoom])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Reload])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Back])->addToWidget(mWebView->focusProxy());
+    static_cast<OverridingAction*>(mActions[Forward])->addToWidget(mWebView->focusProxy());
+
+#    ifdef Q_OS_MAC
+    mActions[ZoomIn]->setShortcutContext(Qt::WidgetWithChildrenShortcut);
```
capital-G: Why do we set only on macOS the shortcut context? And also, why set it to `WidgetWithChildrenSupport` instead of `WidgetShortcut`, see https://doc.qt.io/qt-6/qt.html#ShortcutContext-enum - or: why isn't the default enough here and we need to specify it?

This could be correct though b/c the web browser has a webView member, but it seems strange that this worked before fine as well?

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -135,6 +143,44 @@ void HelpBrowser::onPageLoad() {
...
+    mActions[ZoomOut]->setShortcutContext(Qt::WidgetWithChildrenShortcut);
+    mActions[ResetZoom]->setShortcutContext(Qt::WidgetWithChildrenShortcut);
+#    endif
+
+#    ifdef Q_OS_WIN
+    if (!winCtrlRBlocker) {
+        winCtrlRBlocker = new OverridingAction(this);
+        winCtrlRBlocker->setShortcut(QKeySequence(Qt::CTRL | Qt::Key_R));
+        winCtrlRBlocker->setShortcutContext(Qt::WidgetWithChildrenShortcut);
```
capital-G: the init of the `winCtrlRBlocker` should be moved into the constructor of the class.

Also, I think it is not necessary to guard this w/ a `Q_OS_WIN` beyond the constructor - simply have this as a member and only for `Q_OS_WIN` set this to a non-nullptr and all other platforms check if the pointer is set. This makes the code less platform dependent, such that when refactoring on macOS the code windows code doesn't get trashed unnoticed that easily.

As said elsewhere, this is probably not needed as the event filter for "[blocking] the default `Ctrl + R` behavior on Windows".

editors/sc-ide/widgets/help_browser.cpp
capital-G: This summary looks a bit too much ai assisted to me, not a fan of that^^

> This PR simply fixes the regression

I'd disagree w/ that - I wouldn't want to have the zoom in shortcut performed on the browser to affect the code editor - I'd argue that the previous behavior is wrong as well, and instead of removing this wiring, you are introducing it here explicitly.

capital-G (changes requested):
I must admit that the code didn't get better since the last review - mainly b/c the code is now much more platform dependent, which is always good to be avoided if possible.

Some comments haven't been addressed, such as the `meta` key. But while performing a small google search I came across that one https://doc.qt.io/archives/qt-5.15/qkeysequence.html#standard-shortcuts - maybe this could be relevant?

Alternatively - why don't we use the shortcut as defined in the editor? This seems cleanest to me overall

Image

I think a good implementation could be to block these shortcuts from bubbling up (see review comment about eventfilter) and wire them up to the respective action on the helpbrowser widget. No platform dependent code necessary anymore, which is the most preferred solution, and the user can re-assign this and it becomes consistent across help browser and code editor.
This is probably not more than 20 loc?

editors/sc-ide/widgets/help_browser.cpp
prko: I hope this doesn’t come across the wrong way. I just want to share my thoughts openly so we can clear up any misunderstandings and move forward effectively.

---
---

>I'd disagree w/ that - I wouldn't want to have the zoom in shortcut performed on the browser to affect the code editor - I'd argue that the previous behavior is wrong as well, and instead of removing this wiring, you are introducing it here explicitly.

I think there might be a slight misunderstanding regarding my intention and what this PR actually does.

Let me try to clarify.

---

>I'd argue that the previous behavior is wrong I'd argue that the previous behavior is wrong as well, and instead of removing this wiring, you are introducing it here explicitly.

If you mean I have not yet succeeded in removing the connection between the browser zoom and the editor zoom, then you are entirely right. That is exactly the limitation I was trying to explain.

However, if the meaning is one or all of the following, then that’s a misunderstanding:
- I think the previous behaviour (pressing `"cmd" + "shift" + "="` in HelpBrowser enlarges the font size in `SC-IDE Code Editor`) is correct.
- The issue that pressing `"cmd" + "shift" + "="` in HelpBrowser enlarges the font size in `SC-IDE Code Editor` was resolved in 3.14 or in 3.15 before #7534. I am reintroducing the wrong behaviour in this PR.

To be absolutely clear:

* I **agree** with you 100% that pressing `"cmd" + "shift" + "="` in the `HelpBrowser` should not affect the `SC-IDE Code Editor`. It is wrong behaviour.
* **I did not introduce this wiring.** This problem has existed long before this PR (in 3.13, 3.14, and likely earlier).
* I spent hours trying to completely remove this coupling, but as I mentioned, it seems to require a much deeper architectural change that is currently beyond my skill level. That is why I suggested treating it as a separate, out-of-scope issue rather than letting it block the other fixes here.

---

>This summary looks a bit too much ai assisted to me, not a fan of that^^

I did use AI to help soften my phrasing and clean up my English (e.g., it suggested the phrase *"a proper fix requires a much deeper architectural change"* and it removed a sentence including "regression" which I then rewrote). However, it is strongly based on my own writing.

prko:
Regarding regressions, here is why I consider some behaviours in 3.14/3.15-dev to be a regression compared to 3.13 when interacting inside the `<div class="codeMirrorContainer"> region in HelpBroweser`:

1. Pressing `"cmd" + "0"`
  - In SC3.13:
     Resets the font size in `HelpBrowser`.
  - In SC3.14 and in SC3.15.0-dev (before #7534, since #7534 fixed this):
     Shrinks the font size in `HelpBrowser`.
2. Pressing `"cmd" + "="`, `"cmd" + "-"` and `"cmd" + "0"`  in `<div class="codeMirrorContainer"> region in HelpBroweser`
  - In SC3.13:
     Changes the font size in `HelpBowser`.
  - In SC3.14 and in SC3.15.0-dev (before this PR, since this PR fixes it): 
     No action. It works only outside `<div class="codeMirrorContainer"> in HelpBroweser` works.
3. Code Parsing of SC Code Evaluation in `<div class="codeMirrorContainer"> in HelpBroweser`
  - In SC3.13: 
     No issues.
  - In SC3.14 and SC3.15.0-dev
     Could easily go wrong. See #7544.
4. `HelpBrowser` Stability after SC Code Evaluation in `<div class="codeMirrorContainer"> region in HelpBrowser`
  - In SC3.13: 
     No issues.
  - In SC3.14 and SC3.15.0-dev
     The `HelpBrowser` sometimes goes completely white when evaluating a code block.
 
---

My goal with this PR is simply to restore the working 3.13 behaviors and fix the critical regressions so that the `HelpBrowser` is at least usable while we figure out a better, cleaner way to implement the event filters you suggested.

editors/sc-ide/widgets/help_browser.cpp
prko: To get help from others... Also, it could be used for other developers, because I think now I cannot resolve the issue of `"cmd" + "shift" + "="` on macOS.

prko: Removed in [423e60](https://github.com/supercollider/supercollider/commit/7423e6037fe4c425ca6d7f7603d05c08e100d322)

editors/sc-ide/widgets/help_browser.cpp
prko: This part is removed to use the event filter in [4690746](https://github.com/supercollider/supercollider/pull/7542/commits/46907468abdfa8ff56459ae8bbabd74f26f02964).

editors/sc-ide/widgets/help_browser.cpp
prko: This part is removed in [4690746](https://github.com/supercollider/supercollider/pull/7542/commits/46907468abdfa8ff56459ae8bbabd74f26f02964) and [7423e60](https://github.com/supercollider/supercollider/pull/7542/commits/7423e6037fe4c425ca6d7f7603d05c08e100d322).

editors/sc-ide/widgets/help_browser.cpp
prko: There is no difference in using `WidgetWithChildrenShortcut` and `WidgetShortcut`. Removing this section seems also to be ok. I remove this in [3944541](https://github.com/supercollider/supercollider/pull/7542/commits/3944541e5b2aed9fd12548a559a23d6c0ff1f572)

editors/sc-ide/widgets/help_browser.hpp
prko: I could not configure it. I left this.

prko:
**Current state of this PR:**

- **Fixed**
  - `Cmd/Ctrl` + `= / - / 0` in `HelpBrowser` (both inside and outside the `CodeMirrorContainer`)
    - All platforms: Adjusts the font size correctly.
  - Navigation to previous/next pages now works consistently across all platforms (both inside and outside the `CodeMirrorContainer` in `HelpBrowser`).

- **Partially Fixed**  
  *(Issues present since at least SC 3.13 or earlier)*  
  - `Cmd/Ctrl` + `R` in `HelpBrowser` (both inside and outside the `CodeMirrorContainer`)
    - **FIXED (Linux & macOS):** Refreshes the `HelpBrowser`.
    - **NOT FIXED (Windows):** Still triggers the `Find & Replace` pane in the SC-IDE code editor.
  - `Cmd/Ctrl` + `Shift` + `=` in `HelpBrowser`
    - **macOS:** Increases the font size of the SC-IDE code editor (both inside and outside the `CodeMirrorContainer`).
    - **Linux & Windows:** No functionality.  
      *(Note: In SC 3.15.0-dev, `Ctrl` + `+` was fixed to enlarge the HelpBrowser text, but this PR removes that behavior.)*

**The non-fixed issues are outside the scope of this PR due to my current implementation limitations:**
- On macOS, `Cmd/Ctrl` + `Shift` + `=` increases the font size of the SC-IDE code editor.
- On Windows, `Cmd/Ctrl` + `R` in `HelpBrowser` still triggers the `Find & Replace` pane in the SC-IDE code editor.

If these two issues need to be resolved within this PR, I will need assistance from the coauthors.

prko:
I think it is better to convert this PR to a draft. If someone later finds that this PR can be revived, I would be glad for them to continue. It is also completely fine with me if someone opens a new PR to address the issues discussed here and in #7539.

What I have learned from this PR is that using AI in unfamiliar fields can be convenient, but also quite risky.

My mistakes in this PR:

1. I tried to avoid learning C++ and Qt as much as I could.
2. I relied on AI to solve issues I didn’t fully understand.
3. AI can solve some problems quickly, but it can also ruin work if the user’s attention slips. In unfamiliar disciplines, maintaining steady attention is difficult. Even in familiar disciplines, repetitive work (reading, comparing and judging on revised text and code) can easily cause attention to drift.
4. I focused too much on executing AI‑generated suggestions instead of thinking through the logic myself. This ended up wasting both my time and the reviewer’s time. **I apologise to @capital-G for the time and effort he invested.**
5. In my experiences in general, the idea that AI can reliably solve complex problems is almost a fantasy. It feels similar to buying lottery tickets with a vague expectation of a big win.
6. **To contribute responsibly to SuperCollider development, one needs at least to learn the basics of C++. Since I’m not ready to commit to that, I should avoid opening PRs like this.** What I *can* do is contribute to documentation improvements and the SC class file.

editors/sc-ide/widgets/help_browser.cpp
```diff
@@ -191,10 +197,20 @@ void HelpBrowser::applySettings(Settings::Manager* settings) {
...
+#    else
+    int metaKey = Qt::CTRL;
+#    endif
 
+    zoomInShortcuts.append(QKeySequence(metaKey | Qt::Key_Equal));
+    zoomInShortcuts.append(QKeySequence(metaKey | Qt::Key_Plus));
+    zoomInShortcuts.append(QKeySequence("Ctrl++"));
+
+    mActions[ZoomIn]->setShortcuts(zoomInShortcuts);
+    mActions[ZoomOut]->setShortcut(settings->shortcut("editor-shrink-font"));
```
capital-G: Have you tried "plain" `QKeySequence::ZoomIn` and `QKeySequence::ZoomOut`.
At least it seems off/inconsistent to use the user-provided shortcut for zoom out and some hardcoded values for zooming in ;)

I suggested to use the shortcuts from the editor, but since we are also not using them for refresh etc, it is maybe not bad to use the Qt provided ones?

If you are still interested in the PR, you could give that a try and see how it works out :)

capital-G:
> I think it is better to convert this PR to a draft. If someone later finds that this PR can be revived, I would be glad for them to continue. It is also completely fine with me if someone opens a new PR to address the issues discussed here and in #7539.
> 
> What I have learned from this PR is that using AI in unfamiliar fields can be convenient, but also quite risky.

from my experience: best results w/ AI are achieved when guiding it rather tight - at least if you care about understanding/maintainability of the code, which we care for here ;)
You could use AI as a "teacher" - e.g. "I'd like to implement feature X - can you guide and explain to me each necessary step".

> 3. AI can solve some problems quickly, but it can also ruin work if the user’s attention slips. In unfamiliar disciplines, maintaining steady attention is difficult. Even in familiar disciplines, repetitive work (reading, comparing and judging on revised text and code) can easily cause attention to drift.

Reviewing is also a task ;) Certainly not as fun as coding (IMO), but really necessary and w/ AI this becomes even more important.

> 4. I focused too much on executing AI‑generated suggestions instead of thinking through the logic myself. This ended up wasting both my time and the reviewer’s time. **I apologise to @capital-G for the time and effort he invested.**

No need to apologize here at all! If you learned something along the way, thats fine. I certainly learned something as well :)

> 6. **To contribute responsibly to SuperCollider development, one needs at least to learn the basics of C++. Since I’m not ready to commit to that, I should avoid opening PRs like this.** What I _can_ do is contribute to documentation improvements and the SC class file.

I don't think you do - C++ has some idiosyncrasies, but you can learn those as you go - changes such are these are actually quite manageable IMO - you certainly need to learn something new, but thats the fun part ;)
I can really recommend https://www.learncpp.com/ as a resource, that's how I mostly learned C++.

The code looks much better r/n, and I think some open questions aren't too hard ;)

Regarding the open problems, I took a closer look at this:

`HelpBrowser` derives from `WebView`, which also defines some shortcuts

https://github.com/supercollider/supercollider/blob/90f56761109b62feb8f9fbb02999b11631ba2662/QtCollider/widgets/QcWebView.cpp#L59-L63

See https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts for how these translate to.
But you'll notice, that ctrl+r isn't listed as refresh for windows - so we would have to add this manually.

> macOS: Increases the font size of the SC-IDE code editor (both inside and outside the CodeMirrorContainer)

So when I now press `cmd =` within the help editor, the scd document also changes font size? IMO this is not desirable as stated above, also how this can be avoided.

editors/sc-ide/widgets/help_browser.cpp
prko: > Have you tried "plain" QKeySequence::ZoomIn and QKeySequence::ZoomOut.

Yes:

- https://github.com/prko/supercollider/commit/cade795364b1e0cc1993072ea4237dfec6ce2b6d
- https://github.com/prko/supercollider/commit/aaa3bf2de8695e643a9446d9f0557c428346ee72
- https://github.com/prko/supercollider/commit/c556ee3b11171d91478e4641c1fdc6f0393e9529
- https://github.com/prko/supercollider/commit/4575a817ed7ce0061542254c899b0673fac9b5b3

prko:
Now all except "cmd + shift + =" work.

1. `"Cmd/Ctrl" + "R"`
- Linux, macOS and Windows: Refreshes the HelpBrowser (both inside and outside the code editor)
- On Windows: F5 still refreshes HelpBrowser.

2. `"Cmd/Ctrl" + "="`, `"Cmd/Ctrl" + "0"`, `"Cmd/Ctrl" + "-"`
- All platforms: Adjust font size (both inside and outside the code editor)
- On Windows and Linux: `"Ctrl" + "Shift" + "="` also increases the font size.
- On macOS, I gave up implementing `"Cmd" + "Shift" + "="` to increase the font size.

3. Navigating to the previous and next pages now works consistently across all platforms, both inside and outside the code editor.

editors/sc-ide/widgets/help_browser.cpp
capital-G: @prko I think this has now become obsolete since we are re-wiring the focusProxy events above.
Can you delete this section and see if your new fix is still working?

capital-G (commented):
One thing left to clarify and then I think we are good to get this merged :)

editors/sc-ide/widgets/help_browser.cpp
prko: On macOS, Windows and Ubuntu, I tested HelpBrowser after deleting this part with the artefacts in https://github.com/prko/supercollider/actions/runs/29539586786.

This part seems to be safely removed: https://github.com/supercollider/supercollider/pull/7542/commits/8e9a227cab0b9a4372201d58ee041555ba12c71f

capital-G (approved):
Thx prko, I think this can be merged now :)

capital-G:
Lets see if this will create problems on international keyboards and different keyboard layouts ;)

PR #7545, opened 2026-06-05
sclang/lexer: define all newline literals to be '\n', not the platform default

### LINKED ISSUE #7509: Compiling *.sc files with unicode symbols or bare ^M carriage returns fails

### LINKED ISSUE #7520: sclang: major bug, editors change the content of strings and other text literal (CRLF)

### LINKED ISSUE #7523: Inconsistent sclang output formatting between Windows and macOS/Linux (extra blank lines on Windows)
I’m not sure when this behavior first appeared. I’m currently running build [6314d58] on Windows, but I remember noticing the issue early this year.
Apologies for reporting this so late.

----

SuperCollider’s sclang produces different output formatting on Windows compared with macOS and Linux.  
On Windows, additional blank lines appear between messages, whereas macOS and Linux display the output without these gaps.  
This results in inconsistent behaviour in the IDE’s Post window, including issues with `.postcs` result‑block highlighting.

### Windows example

```
Found 880 primitives.

Compiling directory 'C:\Users\...\SCClassLibrary'

Compiling directory 'C:\ProgramData\SuperCollider\Extensions'

Compiling directory 'C:\Users\...\Extensions'

numentries = 1416224 / 25663200 = 0.055
```

### macOS example

```
Found 879 primitives.
Compiling directory '/Users/.../SCClassLibrary'
Compiling directory '/Users/.../Extensions'
numentries = 860286 / 12957196 = 0.066
```

### Ubuntu example

```
Found 890 primitives.
Compiling directory '/usr/local/share/SuperCollider/SCClassLibrary'
Compiling directory '/home/parallels/.local/share/SuperCollider/Extensions'
numentries = 973484 / 17515328 = 0.056
```

### Why this matters

- The IDE Post window displays the text exactly as sclang emits it.  
  Because Windows inserts extra blank lines, the formatting differs across platforms.
- This affects:
  - readability  
  - the formatting of result blocks generated by .postcs, .post, postln, and other related methods.
  - consistency in tutorials and documentation  
  - the overall user experience  
- The IDE cannot correct this, as it receives pre‑formatted text from sclang.

### Expected behaviour

sclang should handle newlines consistently across all platforms.

### Proposed direction

Standardise sclang’s output so that:

- all platforms use `\n` consistently  
- Windows does not insert additional blank lines  
- the IDE displays identical formatting on every operating system  

This would improve cross‑platform consistency for 3.15.0 and reduce confusion for users and extension developers.

JordanHendersonMusic:
Previously there was a nasty bug where strings with newlines in them depended on the author's os and editor when written in the class library.

This led to this bug that existed on windows when the class was written by a user.
```
A {
  ^str { ^"meow
woof"}
}

A.str != "meow
woof";
```
This is because on windows the newline character is actually two, \r\n, but on unix, and when something is checked into a git repo, and in the sc ide's runtime evaluation mode, the newline character is '\n'.

As well as converting all \r\n into \n, this pr also converts all \r into \n, this is because ancient macs (before 2001) used to use \r as thier newline character and we still have some code that does that.

Further, we redefine that the only whitespace character the ascii literal `$ ` can represent is the ascii space ' '. The error message for this looks as follows:
```
Lexing Error:
Error: /home/jordan/.repo/SuperCollider/SCClassLibrary/SCDoc/SCDoc.sc:1:1
    1 │ $
      ┆ ^ replace with $\n.
    2 │
```

This is done because previously the lexer would turn this runtime code `$` into `$ `, as it would always insert a space at the end of the evaluated code (else it got stuck in an inifitine loop).

Fixes https://github.com/supercollider/supercollider/issues/7520
Fixes https://github.com/supercollider/supercollider/issues/7523
Fixes https://github.com/supercollider/supercollider/issues/7509

## Types of changes

- Documentation
- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


JordanHendersonMusic:
@dyfer what do you think about this? It converts \r and \r\n to \n, and clears up some ambiguity regarding whitespace in the `$` literal so it is always the ascii space. I opted to introduce `\0` as a new escape literal here for completeness. Technically this is a breaking change as previously `\0` would have just turned into `0`. Do you think that is alright?

If so, I'll go ahead and start updating the documentation.

Also, previously if you ended a runtime code evaluation with a `$` it would turn into `$ ` (that's the space right there). I've remove this and turned it into an error.

I also discovered the quoted symbols don't work with escape characters. Here is an interesting one! Not too sure what to do about that as it might be a larger breaking change and probably not worth it.

```
'\n' === 'n' // !!!
```

dyfer:
Thanks @JordanHendersonMusic 

Newline conversions look good to me from the functional standpoint (I haven't looked at the implementation).

One thing I would indeed want to consider further is introducing a new escape sequence `\0`. I'm sorry, but I don't fully understand the need for it. Can you give a complete code example and erroneous behavior where this issue transpires?

JordanHendersonMusic:
Well I added it more for completeness in the lexer, particularly for ASCII literals. I suppose if I added a unicode escape literal instead you could just do "\u0000". This would provide all the functionality and only one potential place to break stuff.

... Actually after writing that out I think it's a better idea. Still it is potentially one breaking change, but means we ought not to need another one like this.

This can do the warning to error thing, so won't become available for a version or so.

I think I will also add a warning if a user attempts to use any non-supported escape character. Although the whitespace ones should stay as hard errors because they are confusing.

This way we can make all the escapes in quoted symbols (expect the newline) warnings too, because they do nothing.

By the way, the reason I was very hesitant to do the warning to error thing before is because previously we were talking about putting this in the lexer. Right now the lexer has no idea what a version is and should stay that way really. This stuff all goes in the compiler, so it's not an issue. 

I really want to try and keep the bits that are pulled out of Lang/ to be version independent and as self contained as possible. The actual compiler can always be a bit of a mess to accommodate these oddities.

dyfer:
Sounds good!

> Well I added it more for completeness in the lexer, particularly for ASCII literals. I suppose if I added a unicode escape literal instead you could just do "\u0000". This would provide all the functionality and only one potential place to break stuff.
> 
> ... Actually after writing that out I think it's a better idea. Still it is potentially one breaking change, but means we ought not to need another one like this.

Yes, that seems to be better, but still, why exactly do we need it?

In any case, a unicode escape sequence would be great, though that would be a breaking change, I think? (i.e. "\u" won't return "u") I'm not opposed, just making sure we all take that into account.

JordanHendersonMusic:
> why exactly do we need it?

Well as this pr makes it impossible to have \r, \v, \f in source code, and later I want to make the rest of the bizarre ancient ASCII crap illegal, the user might want a legitimate way to input these characters.

It is unlikely to come up in general sc use, but when I eventually rework the threading issue and revive the c API, stuff like this might come up.

JordanHendersonMusic:
@dyfer I've left out the proposed \u to hopefully make reviewing easier. Otherwise, hopefully this is all good. It is the least disruptive way possible to do this, it could cause issue if the user has on purposed put a carriage return/vertical tab/form feed in their code. But this seems really unlikely.

HelpSource/Classes/Char.schelp
```diff
@@ -4,7 +4,7 @@ related::Classes/String
 categories:: Core
 
 description::
-An ASCII character represented as a signed 8-bit integer (-128 to 127).
+A Char represented as a signed 8-bit integer (-128 to 127).
```
dyfer: So is it not an ASCII character? is it a Unicode character? "A Char" is probably least descriptive here.

HelpSource/Classes/Char.schelp
dyfer, line 2: If it's not ASCII anymore (?) this should also be updated

HelpSource/Classes/Char.schelp
```diff
@@ -17,9 +17,20 @@ multi-byte encodings like UTF-8 and UTF-16, or the full Latin-1 (ISO 8859-1)
 character set—Chars with negative values are perfectly legal, and may be strung
 together in strings that use these encodings.
 
-The SuperCollider IDE uses UTF-8 to decode and display strings.
+The SuperCollider IDE and UTF-8 to decode and display strings.
```
dyfer: I don't understand this change...?

HelpSource/Reference/Literals.schelp
```diff
@@ -86,9 +86,9 @@ This eliminates ambiguity: if lowercase letters were allowed,
 code::36rA.bitNot:: might be a function call, or it might be
 code::36rA.BITNOT == 10.320080118934::.
 code::
-12r4a.abc // wrong
-12r4a.ABC // works
-12r4A.ABC // better
+12r4a.ab // wrong
+12r4a.AB // works
+12r4A.AB // better
```
dyfer, line 91: Why are these changed?

HelpSource/Reference/Literals.schelp
```diff
@@ -196,6 +196,13 @@ code::
 \not really a symbol  // illegal
 ::
 
+You can also specify integer symbols this way, but if the symbol starts with a digit, it cannot contain letters.
+code::
+\1234;
+\892342534;
+\3456meow // not illegal
```
dyfer: not illegal?

HelpSource/Classes/Char.schelp
```diff
@@ -17,9 +17,20 @@ multi-byte encodings like UTF-8 and UTF-16, or the full Latin-1 (ISO 8859-1)
...
 See link::Classes/String:: for more information.
 
+note::
+From version 3.15 and onwards one can no longer have a whitespace literal (except the space) after the dollar sign.
```
dyfer: I think this should be `From version 3.15 onwards` (i.e. remove "and")

dyfer (commented):
Thanks!

I can't review cpp part of this reliably, but I think the functional changes are good.

I have some comments regarding the help files.

HelpSource/Classes/Char.schelp
JordanHendersonMusic: No it isn't a unicode codepoint either, that's 21 bits. It is just a single byte. Half of which is valid ascii.

HelpSource/Reference/Literals.schelp
JordanHendersonMusic: Because they were invalid. Try running the old stuff and you will get a parsing error because 'c' is an unexpected identitfier. I wrote an issue about this weird bit of syntax from radixs a while a ago.

It's unrelated but just a small thing.

HelpSource/Classes/Char.schelp
JordanHendersonMusic: Oops! I rewrote this and decided against it!

HelpSource/Classes/Char.schelp
JordanHendersonMusic: It never was because it is a char.

... Sorry I realised that isn't useful! It's a whole byte but ASCII is only 7 bits. So it can represent invalid ASCII too.

HelpSource/Classes/Char.schelp
JordanHendersonMusic:
```suggestion
The SuperCollider IDE uses UTF-8 to decode and display strings.
```

HelpSource/Classes/Char.schelp
JordanHendersonMusic:
```suggestion
From version 3.15 onwards one can no longer have a whitespace literal (except the space) after the dollar sign.
```

HelpSource/Reference/Literals.schelp
JordanHendersonMusic:
```suggestion
\3456meow // now illegal
```

HelpSource/Classes/Char.schelp
dyfer: So maybe "A character represented as a signed 8-bit integer (-128 to 127)" ?

JordanHendersonMusic: Actually I think it should be 'A Char is a signed 8-bit integer' ? 

The class is called char and not all valid chars are characters. Really it should just be called Byte.

dyfer: Char's identity was always a mystery to me.
`Char is a signed 8-bit integer` is pretty good, but it differs from a regular integer since it is most commonly represented as a character. 
What else can it represent? When it's value lies outside of ASCII, isn't it essentially a flavor of [extended ASCII](https://en.wikipedia.org/wiki/Extended_ASCII)?

JordanHendersonMusic: Well what it is interpreted as defines how it will.be displayed. The ide assumes strings are utf8, so you can use a char to represent a single byte in a utf8 stream.... But really strings are just arrays of bytes, so you can store whatever you like in it.

Extended ASCII isn't really a thing outside windows anymore.

There is a bit of a naming issue here because ASCII literals **can** only be ASCII ($a), but they are converted to a char, which can represent a large set of values. I think I'll have another flick through this doc and see if some of this can be rephrased.


...I actually just thought of a bug where I need to add a compiler error if someone tries to assign a  unicode character to an ASCII literal.

JordanHendersonMusic:
@dyfer I think I've address all your concerns? Do you think this can be approved now? Don't worry about reviewing the C++ code, the stuff in the lexer is trivial and the rest of the changes are in the old code and are messy because the state there was already messy — I'm going to refactor the whole thing anyway.

langutils/sc_lexer/include/codepoint_stream.hpp
```diff
@@ -46,8 +47,10 @@ class CodePointStream {
     } state {};
 
 public:
-    CodePointStream(const char* src, std::size_t src_len, FileCodeLocation src_start_in_file);
-    CodePointStream() = delete;
+    CodePointStream(NormalisedSource src, FileCodeLocation src_start_in_file) noexcept:
+        source_start_in_file(src_start_in_file),
+        source(std::move(src).steal_string()) {}
+    // CodePointStream(const char* src, std::size_t src_len, FileCodeLocation src_start_in_file);
```
dyfer: Is this supposed to stay here?

dyfer:
I think this is good from the functional standpoint. I have not reviewed C++ but I trust Jordan's judgement.

I'm happy to merge this, @capital-G do you have any opinion here? Basically are there any blind spots we haven't thought of.

langutils/sc_lexer/include/codepoint_stream.hpp
JordanHendersonMusic: Thanks!

JordanHendersonMusic:
> Basically are there any blind spots we haven't thought of.

Please note the ownership is a mess, that is because that side of things is being refactored.

telephon (approved):

PR #7548, opened 2026-06-07
build: expose static plugins as a build option

dyfer:
## Purpose and Motivation

This PR builds on the recently re-enabled static plugins option in the wasm build.

I'd like to propose exposing static plugins option in cmake. I'm not sure if we need it for anything else, but I think this is cleaner from the logical standpoint than the current implementation.

## Types of changes

- Documentation
- New feature (e.g. build scsynth with static plugins on macOS)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/.github/workflows/build_wasm.yml
+++ b/.github/workflows/build_wasm.yml
@@ -41,6 +41,7 @@ jobs:
             -DSUPERNOVA=no \
             -DSC_HIDAPI=no \
             -DNO_LIBSNDFILE=yes \
+            -DSTATIC_PLUGINS=yes \
             -DSC_QT=no \
             -DNO_AVAHI=yes \
             -DSC_ABLETON_LINK=no \
--- a/server/scsynth/CMakeLists.txt
+++ b/server/scsynth/CMakeLists.txt
@@ -147,24 +147,25 @@ elseif (AUDIOAPI STREQUAL bela)
 elseif (AUDIOAPI STREQUAL webaudio)
     add_definitions(
             "-DSC_AUDIO_API=SC_AUDIO_API_WEBAUDIO"
-            "-DSTATIC_PLUGINS"
     )
     list(REMOVE_ITEM scsynth_sources
             SC_ComPort.cpp
             ${CMAKE_SOURCE_DIR}/common/SC_Reply.cpp
     )
+    list(APPEND scsynth_sources
+            SC_WebAudio.cpp
+            ${CMAKE_SOURCE_DIR}/platform/wasm/SC_WebOsc.cpp
+    )
+endif()
+
+if(STATIC_PLUGINS)
     # the plugins are linked statically, their sources must be included
     set(plugin_sources_abs "")
     foreach(plugin ${all_plugin_sources})
         string(CONCAT plugin_abs ${CMAKE_SOURCE_DIR}/server/plugins/ ${plugin} )
         list(APPEND plugin_sources_abs ${plugin_abs})
     endforeach(plugin)
-
-    list(APPEND scsynth_sources
-            SC_WebAudio.cpp
-            ${CMAKE_SOURCE_DIR}/platform/wasm/SC_WebOsc.cpp
-            ${plugin_sources_abs}
-    )
+    list(APPEND scsynth_sources ${plugin_sources_abs})
 endif()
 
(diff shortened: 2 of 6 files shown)


server/scsynth/SC_Lib_Cintf.cpp
```diff
@@ -128,14 +128,20 @@ extern void DynNoise_Load(InterfaceTable* table);
 extern void FFT_UGens_Load(InterfaceTable* table);
 extern void iPhone_Load(InterfaceTable* table);
 
+#    ifndef NO_LIBSNDFILE
 extern void DiskIO_Unload(void);
+#    endif
+#    ifndef NO_X11
 extern void UIUGens_Unload(void);
```
dyfer, line 135: I'm not sure why these functions were previously exposed even in the builds without disk and ui ugens... that's why I added preprocessor guards here.

capital-G: well spotted - this will lead to a merge conflict on a branch I am working on, but better to resolve it here :)

.github/workflows/build_wasm.yml
```diff
@@ -41,6 +41,7 @@ jobs:
             -DSUPERNOVA=no \
             -DSC_HIDAPI=no \
             -DNO_LIBSNDFILE=yes \
+            -DSTATIC_PLUGINS=yes \
```
capital-G, line 44: should we make some build flags for wasm implicit - e.g. setting `supernova=no` within CMakeFile for wasm?
If we go for this, it should be done outside of this PR though.

server/scsynth/CMakeLists.txt
```diff
@@ -228,6 +229,13 @@ if(UNIX AND NOT APPLE)
 	target_compile_definitions(libscsynth PUBLIC "SC_PLUGIN_DIR=\"${LINUX_SC_PLUGIN_DIR}\"")
 endif()
 
+if(STATIC_PLUGINS)
+    target_compile_definitions(libscsynth PRIVATE "-DSTATIC_PLUGINS")
```
capital-G, line 233: care to explain why this is necessary? just curious, want to learn more cmake. Wouldn't `libscsynth` inherit the cache variables from global cmake?

capital-G (approved):
Thanks! lgtm - feel free to merge [although this introduces some merge conflicts on my side, but oh well]

There are some questions, but this is not affecting the implementation here.

.github/workflows/build_wasm.yml
dyfer: I was thinking about that. Seems like cmake preset would be a good way to implement this? I haven't used them, but from what I read it seems like a clean solution to set a number of options based on a platform

server/scsynth/CMakeLists.txt
dyfer: Global cmake `STATIC_PLUGINS` option is a cmake variable, IIUC it doesn't make its way into Cpp preprocessor automatically - `add_definitions()` is what adds definitions to the compilation invocation. `add_definitions` is global(-ish?) and IIUC using definitions associated with targets is cleaner, so that's why I add it to the target here.

Also cmake option could be called something else to differentiate it from the preprocessor definition, do you think we should do that? e.g. `SC_STATIC_PLUGINS` in cmake.

.github/workflows/build_wasm.yml
capital-G: Yeah, lets try those. I used them for some projects now and I think I am a bit torn apart and haven't found them that useful - e.g. you can't overwrite a base preset from which other presets inherit from, but this would be e.g. important for e.g. using a custom qt resolution for macOS etc. And they also start to duplicate things from cmake^^

But lets give them a try - they are optional so we can try if they actually provide some help for some people.

server/scsynth/CMakeLists.txt
capital-G: I was just a bit confused by `-D` - it seems cmake strips this away? And I mixed up that this is a compile definition, I thought this was a compile option :)

IMO it is fine as is :)

dyfer: > I was just a bit confused by `-D` - it seems cmake strips this away

Yes... we use `-D` elsewhere that's why I used it, but later we can clean this up and remove those.

PR #7440, opened 2026-04-01
Add wasm support for sclang

capital-G:
This PR is built on top of https://github.com/supercollider/supercollider/pull/7428 - so it may look more intimidating than it actually is...

A demo of the new webeditor is hosted at https://sc.dennis-scheiba.com/

## Purpose and motivation

quoting from the new readme: 

Interaction with sclang happens via OSC and text I/O, where both kinds are accessible via a JavaScript API.
Like for scsynth, all kind of networking is stripped away since wasm does not provide socket access.
Instead, all incoming OSC messages should be passed as binary [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) to `sendOsc` and all outgoing messages can be received as Uint8 array via `onOsc`.
Take a look at the web editor how to interact with the API.

The current network mock is rather limited and aimed at communicating with the scsynth.wasm server.
To keep the implementation easier, the concept of IP address and port has been removed, which limits the usage of the wasm build to single client/server setups.
This may change in the future though.

### Limitations

* No file access yet.
  Will be available once the best approach for a filesystem in conjunction with scsynth and the browser has been determined.
* DOM interaction is untested, but `JS.runCode` can run JS code in the main thread of the browser, which allows to interact with the DOM.
* Any kind of OSCFunc/OSCdef that goes beyond the server is not implemented.
* Clock-re-synchronization with the server is currently not in-place.
* The clock implementation (SystemClock, AppClock, ...) may be janky in general.

### Source code

Any kind of patching in the source code is guarded by the `__EMSCRIPTEN__` definition.
The most crucial patching has been performed in

* `PyrObject.cpp`: Compilation of the class library is not happening in a boost thread pool but using `std::async` - like the Windows implementation.
  Using the thread pool created termination problem of the thread pool.
* `OSCData.cpp`: WASM does not allow to create sockets, so it is assumed that the network socket is located at `127.0.0.1:57120`.
  Operations such as `NetAddr.broadcastFlag` are no-oped.
* `SC_ComPort.h`: Any reference to a socket is removed

The terminal client `SC_TerminalClient`, which is derived from `LanguageClient`, is replaced with a `SC_WasmClient`, which exposes text and OSC I/O via a JavaScript API using embind.

Additionally, `SC_WebOsc.cpp` is also added to the build, which allows to construct and parse OSC messages.

## Types of changes

[...]

PR #7428, opened 2026-03-21
Add scsynth wasm build

capital-G:
## Purpose and Motivation

Demo is available via https://supercollider.dennis-scheiba.com/ 

Introduces wasm build of scsynth using [`AudioWorklets`](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet), therefore replacing #6569 (and therefore also https://github.com/supercollider/supercollider/pull/5309) which ran on the main thread of the JS engine.
AudioWorklets run in the dedicated real time audio thread of the browser, improving the performance and removing a deprecation warning.

At some point we may want to publish this through npm?

@JordanHendersonMusic promised me to review this if I rewrite the PR in AudioWorklets^^ Hope that is still a thing ;)

Addressing objections in the former PRs:

* CI includes a new `wasm` build target, so any future changes on scsynth are also checked against the wasm build
* Provide an example on how to use the wasm build and deploy it to the internet
* Added documentation on how to build and develop and documented the source code

Additionally, this PR fixes a longstanding bug in the construction of init rate UGens - see 71b9340e1fc6205e03594fcb85659eefb54b770c

Some things which are skipped in this PR to make reviewing easier by reducing the implementation footprint, namely

* Buffer and DiskAccess support
* X11/Mouse UGen support

Credits go to Hanns Holger Rutz / @Sciss  who wrote the original implementation in #5309 , but this implementation is from scratch except some CMake lines and the gcd demo synth.

## Types of changes

- Documentation
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review 

## Merge notes

I'd appreciate a rather soonish review in order to keep things in sync here and avoid resolving merge-conflicts

edit: Since the last commit only covers a README file only a subset of the CI runs - here is a full CI run from my branch https://github.com/capital-G/supercollider/actions/runs/23380871142

PR #7552, opened 2026-06-10
supernova: n_before and n_after using the moved node's parent instead of the target's group

### LINKED ISSUE #5909: Supernova's moveAfter() behaves differently than for scsynth
## Environment

* SuperCollider version: 3.2.12
* Operating system: Ubuntu 22.04

## Steps to reproduce

```supercollider
Server.supernova
s = Server.default()
s.waitForBoot( SynthDef(\test, { Out.ar(10, SinOsc.ar())} ).add; )

g = Group()
h = Group(addAction: 'addToTail')

a = Synth(\test, target: g, addAction: 'addToTail')
b = Synth(\test, target: g, addAction: 'addToTail')
c = Synth(\test, target: g, addAction: 'addToTail')

s.queryAllNodes
/*
NODE TREE Group 0
   0 group
      1 group
         1000 group
            1002 test
            1003 test
            1004 test
         1001 group
*/

a.moveToHead(h)
c.moveAfter(a)
s.queryAllNodes

/*
NODE TREE Group 0
   0 group
      1 group
         1000 group
            1003 test
            1004 test
         1001 group
            1002 test
*/

a.moveToHead(g)
b.moveToHead(h)
a.moveToHead(h)
c.moveAfter(b)
s.queryAllNodes
/*
NODE TREE Group 0
   0 group
      1 group
         1000 group
            1004 test
         1001 group
            1002 test
            1003 test
*/
c.moveAfter(a)
s.queryAllNodes
/*
NODE TREE Group 0
   0 group
      1 group
         1000 group
         1001 group
            1002 test
            1004 test
            1003 test
*/
c.moveAfter(b)
s.queryAllNodes
/*
NODE TREE Group 0
   0 group
      1 group
         1000 group
            1004 test
         1001 group
            1002 test
            1003 test
*/
s.freeAll

```

Except for the 4th query listed above (out of 5), the listing given by `supernova` is different than what I'd expect from `moveAfter()` and different from the output given by `scsynth`.  I'd expect `supernova` to behave in the same way as `scsynth`:

```supercollider
Server.scsynth
s = Server.default()
s.waitForBoot( SynthDef(\test, { Out.ar(10, SinOsc.ar())} ).add; )

g = Group()
h = Group(addAction: 'addToTail')

a = Synth(\test, target: g, addAction: 'addToTail')
b = Synth(\test, target: g, addAction: 'addToTail')
c = Synth(\test, target: g, addAction: 'addToTail')

s.queryAllNodes
/*
NODE TREE Group 0
   1 group
      1000 group
         1002 test
         1003 test
         1004 test
      1001 group
*/

a.moveToHead(h)
c.moveAfter(a)
s.queryAllNodes

/*
NODE TREE Group 0
   1 group
      1000 group
         1003 test
      1001 group
         1002 test
         1004 test
*/

a.moveToHead(g)
b.moveToHead(h)
a.moveToHead(h)
c.moveAfter(b)
s.queryAllNodes
/*
NODE TREE Group 0
   1 group
      1000 group
      1001 group
         1002 test
         1003 test
         1004 test
*/
c.moveAfter(a)
s.queryAllNodes
/*
NODE TREE Group 0
   1 group
      1000 group
      1001 group
         1002 test
         1004 test
         1003 test
*/
c.moveAfter(b)
s.queryAllNodes
/*
NODE TREE Group 0
   1 group
      1000 group
      1001 group
         1002 test
         1003 test
         1004 test
*/
s.freeAll
```

It is unclear to me whether already reported issues: #936 and #3100 are related.

shaygp:
move_before_or_after resolved the destination from the moved node's parent instead of the target's parent. when the node and the target were in different groups this took the wrong branch (it moved the node to the tail of its own group) and also skipped the child count and parent updates, so the node was spliced across lists with a stale parent. a following move could then send the node back to its old group.

now it resolves the target's parent like scsynth does. added a test that compares the resulting tree on scsynth and supernova for cross group n_after and a double move.

closes #5909

### DIFF
--- a/testsuite/classlibrary/TestServer_nodeMove.sc
+++ b/testsuite/classlibrary/TestServer_nodeMove.sc
@@ -1,0 +1,76 @@
+TestServer_nodeMove : UnitTest {
+
+	var pipe;
+
+	tearDown {
+		if (pipe.notNil) {
+			pipe.close;
+		};
+	}
+
+	treeAfter { |program, moves|
+		var order = List[],
+			ip = "127.0.0.1",
+			port = 57110,
+			addr = NetAddr(ip, port),
+			line,
+			match;
+		pipe = Pipe.new(program ++ ServerOptions().bindAddress_(ip).asOptionsString(port), "r");
+		line = pipe.getLine;
+		while ({ line.notNil && line.contains("ready").not }) {
+			line = pipe.getLine;
+		};
+		// build a known tree, then apply the moves and read back the resulting
+		// depth-first node order (root id 0 dropped). /g_new args are:
+		// new group id, add action (0 = head of target, 1 = tail of target), target id.
+		// the result is two sibling groups under the root, each holding two child groups:
+		//   1000 ( 2000 2001 )   1001 ( 3000 3001 )
+		addr.sendMsg("/g_new", 1000, 0, 0);
+		addr.sendMsg("/g_new", 1001, 1, 0);
+		addr.sendMsg("/g_new", 2000, 0, 1000);
+		addr.sendMsg("/g_new", 2001, 1, 1000);
+		addr.sendMsg("/g_new", 3000, 0, 1001);
+		addr.sendMsg("/g_new", 3001, 1, 1001);
+		moves.do { |m| addr.sendMsg(*m) };
+		addr.sendMsg("/g_dumpTree", 0, 0);
+		addr.sendMsg("/quit");
+		line = pipe.getLine;
+		while ({ line.notNil }) {
+			match = line.findRegexp("^\\s+([0-9]+) group");
+			if (match.notEmpty and: { match[1][1].asInteger != 0 }) {
+				order.add(match[1][1].asInteger);
+			};
+			line = pipe.getLine;
+		};
+		pipe.close;
+		^ order
+	}
+
+	// moving 2001 after 3000 should pull it out of group 1000 and place it
+	// inside group 1001, right after 3000.
+	test_moveAfter_acrossGroups {
+		var moves = [["/n_after", 2001, 3000]];
+		[Server.program, Server.program.replace("scsynth", "supernova")].do { |program|
+			this.assertEquals(
+				this.treeAfter(program, moves),
+				List[1000, 2000, 1001, 3000, 2001, 3001],
+				program.basename ++ " /n_after across groups should move the node into the target's group",
+			);
+		};
+	}
+
+	// first move 2000 into group 1001, then move it again within 1001. the second
+	// move must resolve 2000's real (updated) parent, not a stale one, otherwise
+	// supernova sends it back to group 1000.
+	test_moveAfter_doesNotStaleParent {
+		var moves = [["/n_after", 2000, 3000], ["/n_after", 2000, 3001]];
+		[Server.program, Server.program.replace("scsynth", "supernova")].do { |program|
+			this.assertEquals(
+				this.treeAfter(program, moves),
+				List[1000, 2001, 1001, 3000, 3001, 2000],
+				program.basename ++ " a second /n_after should follow the node's real parent, not a stale one",
+			);
+		};
+	}
+
+}
--- a/server/supernova/server/group.hpp
+++ b/server/supernova/server/group.hpp
@@ -218,7 +218,7 @@ public:
 
     template <node_position Relation> static void move_before_or_after(server_node* node, server_node* target) {
         assert((Relation == before) || (Relation == after));
-        abstract_group* target_parent = node->get_parent();
+        abstract_group* target_parent = target->get_parent();
 
         if (Relation == after && target->next_node() == nullptr) {
             // for the sake of simplicity, move the node to the tail of the target's parent group


jamshark70:
I'm not qualified to review C++, but I did want to take a sec to say thanks for this. It's overdue to close some of the gaps where supernova interprets the server OSC spec differently.

(Another related bug is that n_order works [IIRC] oppositely in supernova vs scsynth...)

shaygp:
super glad it helps! i got into supercollider pretty recently and i think it's a cool and important project, so i want to keep putting time into it.
  
that one is up as #7551, same family of bug... supernova was applying the add action to every node in the list instead of only the first so head and after came out reversed (the example in the linked issue didn't change at all). 

i've started looking at the browser/wasm side too (#5224) and i was thinking about decoding audio in JS and copying the floats to avoid porting libsndfile. 

i got deep into the browser side because i believe there's a lot to do there too but also important for me to know if this is something the contributors find important or would my time be better spent elsewhere?

capital-G:
Hey - welcome! Thanks for looking into supernova, this is much appreciated!

> i've started looking at the browser/wasm side too (#5224) and i was thinking about decoding audio in JS and copying the floats to avoid porting libsndfile.

A few days ago a wasm build of scsynth has been merged, see #7428 and #7440. I have a PR where I also built scsynth w/ libsndfile to wasm. Libsndfile adds around 2 mb. Is there a benefit of avoiding libsndfile beyond the added file size?

One thing that hasn't been implemented/fixed yet is the synchronization between sclang and scsynth in wasm. HMU if you are further interested in some wasm stuff :)

shaygp:
thanks, glad to be here!

> A few days ago a wasm build of scsynth has been merged, see #7428 and #7440. I have a PR where I also built scsynth w/ libsndfile to wasm. Libsndfile adds around 2 mb. Is there a benefit of avoiding libsndfile beyond the added file size?

ahhh i just saw that, amazing! the only thing is that decodeAudioData always resamples to the AudioContext rate (usually 48k), so you lose the file's original sample rate and exact data. that's perfectly fine for most stuff but Libsndfile has deterministic decoding (same result across browsers where decodeAudioData can differ since each ships its own decoder and resampler) and preserves the original file characteristics. Libsndfile as the default is great, maybe an optional JS decode helper later. 

the sclang/scsynth sync point sounds really interesting to me though. i found #7440 and the clock re synchronization with the server limitation you list there. be right back : )

testsuite/classlibrary/TestServer_nodeMove.sc
```diff
@@ -0,0 +1,66 @@
...
+		while ({ line.notNil && line.contains("ready").not }) {
+			line = pipe.getLine;
+		};
+		addr.sendMsg("/g_new", 1000, 0, 0);
+		addr.sendMsg("/g_new", 1001, 1, 0);
+		addr.sendMsg("/g_new", 2000, 0, 1000);
+		addr.sendMsg("/g_new", 2001, 1, 1000);
+		addr.sendMsg("/g_new", 3000, 0, 1001);
+		addr.sendMsg("/g_new", 3001, 1, 1001);
+		moves.do { |m| addr.sendMsg(*m) };
```
JordanHendersonMusic, line 34: Could you just add a comment explaining what these should do? It's not entirely obvious what this is testing.

JordanHendersonMusic (changes requested):
One really small thing here, otherwise this is great, thanks!

testsuite/classlibrary/TestServer_nodeMove.sc
shaygp: done, added comments explaining the tree setup and what each test checks, thanks!

testsuite/classlibrary/TestServer_nodeMove.sc
```diff
@@ -0,0 +1,76 @@
...
+			line = pipe.getLine;
+		};
+		// build a known tree, then apply the moves and read back the resulting
+		// depth-first node order (root id 0 dropped). /g_new args are:
+		// new group id, add action (0 = head of target, 1 = tail of target), target id.
+		// the result is two sibling groups under the root, each holding two child groups:
+		//   1000 ( 2000 2001 )   1001 ( 3000 3001 )
+		addr.sendMsg("/g_new", 1000, 0, 0);
+		addr.sendMsg("/g_new", 1001, 1, 0);
+		addr.sendMsg("/g_new", 2000, 0, 1000);
```
dyfer, line 30: Shouldn't we sync before creating subgrups to make sure that parent groups exist?

shaygp: these all go over one socket in order and /g_new is synchronous on the server so a child's parent is already crreated by the time the child is processed. the existing server tests rely on the same udp-on-loopback setup and i ran the full setup as a sanity check with no mismatch.

a sync wouldn't change the one real risk, a dropped packet since it can't recover that either. if you'd prefer a hard delivery guarantee though, i'm happy to switch the control messages to tcp. lmk!

dyfer: Thanks for clarifying!
I think we considered changing the test suite to use tcp, but that did not fully work the last time we tried IIRC (can't remember the exact reasons), so let's not go that route for now.
If `/g_new` indeed is synchronous on the server, then maybe this is fine as is. What we don't have control over is the message ordering with udp, but the likelihood of these being out of order on localhost is probably as negligent as a dropped message... right?

JordanHendersonMusic: Yeah tcp has some issue on the CI and isn't possible right now, it would be good to move everything over at once when (if) it is ever sorted.

I don't think out of order messages is likely on a local host, I think more of the test suite would be (even more) broken if that happened frequently.

JordanHendersonMusic (approved):

PR #7554, opened 2026-06-11
supernova: /notify reply not including the max logins

### LINKED ISSUE #3337: Supernova does not reply with max logins on /notify request
From [this comment](https://github.com/supercollider/supercollider/pull/3275#issuecomment-350616155)

> The /done, /notify format has changed in scsynth, but the corresponding change hasn't happened in supernova.
>
> This is likely to affect only multiclient applications using supernova -- i.e., almost nobody. But, to be really correct, we should fix it.
>
> I guess here: https://github.com/supercollider/supercollider/blob/3.9/server/supernova/sc/sc_osc_handler.cpp#L1116

See related work in https://github.com/supercollider/supercollider/pull/3245

Milestoning as 3.9.x, but would be nice for 3.9.1

shaygp:
scsynth replies to /notify with /done /notify clientID maxLogins but supernova only sent /done /notify clientID. the max-logins option was also never stored in supernova, the parser dropped it, so -l did nothing. now the option is read into the server arguments and the value is sent in the reply like scsynth.

added a test that checks the /done /notify reply on scsynth and supernova includes the max logins, and that the configured -l value is reported back.

closes #3337

### DIFF
--- a/testsuite/classlibrary/TestServer_notifyReply.sc
+++ b/testsuite/classlibrary/TestServer_notifyReply.sc
@@ -1,0 +1,51 @@
+TestServer_notifyReply : UnitTest {
+
+	notifyReply { |program, port, logins = 1|
+		var ip = "127.0.0.1",
+			addr,
+			responder,
+			reply,
+			condvar = CondVar(),
+			gotReply = false,
+			attempts = 0;
+		(program ++ ServerOptions().maxLogins_(logins).bindAddress_(ip).asOptionsString(port)).unixCmd;
+		addr = NetAddr(ip, port);
+		responder = OSCFunc({ |msg| reply = msg; gotReply = true; condvar.signalOne }, '/done', addr);
+		// poll while the server boots: a /notify sent over UDP before the server
+		// is listening is simply dropped, so resend until the /done reply arrives.
+		// normally this takes about a second, the cap is a give-up for a server
+		// that never comes up (25 * 0.2s = 5s).
+		while ({ gotReply.not and: { attempts < 25 } }) {
+			addr.sendMsg("/notify", 1);
+			condvar.waitFor(0.2);
+			attempts = attempts + 1;
+		};
+		responder.free;
+		// resend /quit a few times: a single UDP packet can be dropped while the
+		// server is still finishing its boot, which would leave it running.
+		3.do { addr.sendMsg("/quit"); 0.05.wait };
+		^ reply
+	}
+
+	test_notifyReply_includesMaxLogins {
+		this.assertEquals(
+			(this.notifyReply(Server.program, 57398) ?? { [] }).size,
+			4,
+			"scsynth /done /notify reply should include the client ID and max logins"
+		);
+		this.assertEquals(
+			(this.notifyReply(Server.program.replace("scsynth", "supernova"), 57399) ?? { [] }).size,
+			4,
+			"supernova /done /notify reply should include the client ID and max logins"
+		);
+	}
+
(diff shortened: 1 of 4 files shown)


testsuite/classlibrary/TestServer_notifyReply.sc
```diff
@@ -0,0 +1,45 @@
...
+			condvar = CondVar(),
+			gotReply = false,
+			attempts = 0;
+		(program ++ ServerOptions().maxLogins_(logins).bindAddress_(ip).asOptionsString(port)).unixCmd;
+		addr = NetAddr(ip, port);
+		responder = OSCFunc({ |msg| reply = msg; gotReply = true; condvar.signalOne }, '/done', addr);
+		while ({ gotReply.not and: { attempts < 50 } }) {
+			addr.sendMsg("/notify", 1);
+			condvar.waitFor(0.2);
+			attempts = attempts + 1;
```
JordanHendersonMusic, line 21: So it's a max wait time of 100 seconds? How often does this normally take?

Our ci has a timeout of 20mins and our longest test is about 10 seconds.

JordanHendersonMusic (commented):
One query then this looks good 👍

testsuite/classlibrary/TestServer_notifyReply.sc
```diff
@@ -0,0 +1,51 @@
...
+		responder.free;
+		// resend /quit a few times: a single UDP packet can be dropped while the
+		// server is still finishing its boot, which would leave it running.
+		3.do { addr.sendMsg("/quit"); 0.05.wait };
+		^ reply
+	}
+
+	test_notifyReply_includesMaxLogins {
+		this.assertEquals(
+			(this.notifyReply(Server.program, 57398) ? []).size,
```
JordanHendersonMusic: Not particularly important here as this is a test, but you should avoid `? r` as it is just a less performance version of `?? { r }` as the former always evaluates `r` but the latter only when needed.

shaygp: switched to ?? { [] }. thanks!

dyfer:
Thanks for implementing this!

I have a question about the tests: why are they implemented using serverProgramCmd.unixCmd instead of `server.bootSync`?
Unless I'm missing something, it would be better to have `server.bootSync` in `setUp {}` and `server.quit` in `tearDown {}` methods, like [here](https://github.com/supercollider/supercollider/blob/develop/testsuite/classlibrary/TestBuffer_Server.sc#L5-L14). For the Supernova test, it _should_ work to just call `Server.supernova` beforehand, but I haven't tested that (then I'd call `Server.scsynth` in teardown to reset it).

testsuite/classlibrary/TestServer_notifyReply.sc
```diff
@@ -0,0 +1,51 @@
...
+		while ({ gotReply.not and: { attempts < 25 } }) {
+			addr.sendMsg("/notify", 1);
+			condvar.waitFor(0.2);
+			attempts = attempts + 1;
+		};
+		responder.free;
+		// resend /quit a few times: a single UDP packet can be dropped while the
+		// server is still finishing its boot, which would leave it running.
+		3.do { addr.sendMsg("/quit"); 0.05.wait };
+		^ reply
```
dyfer, line 27:
```suggestion
		^reply
```
AFAICT we generally don't put a space between a caret and the return statement

dyfer:
@shaygp can you consider my suggestion regarding the unit test?

Issue #7410, opened 2026-03-13
Lexer quietly ignores ellipses unicode in place of actual ellipses


lnihlen:
## Environment

* SuperCollider version: 3.14.1
* Operating system: macOS
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce

Filing this now because it looks like it's timely, I hear SC is getting a new lexer soon possibly..

I found this while working on parsing every example of SuperCollider code I can find with Hadron. This is in a Quark, see https://github.com/muellmusik/Utopia/issues/26 for the details.

The following two code examples produce different results while looking visually very similar. They both compile and run without error in SC 3.14.1:

```supercollider
(
f = {|...args| args };
f.(0, 1, 2)
)  // produces `[0, 1, 2]`

(
f = {|… args| args };
f.(0, 1, 2)
)  // produces `0`
```

The second input, some text editor has likely "helpfully" replaced the three periods with the ellipses unicode code point U+2026.

## Expected vs. actual behavior

The second input in invalid SuperCollider code and should produce a compilation error.

JordanHendersonMusic:
Ah! I think it is being accepted by the branch that does something like ...
```
if(!(isprint(c) || isspace(c))) 
   discard character
```
So it's just being thrown away leaving just the args. Thing is, this is undefined anyway because `c` is an int and these functions explicitly only work on `unsigned chars`.

lnihlen:
I think if you aren't in a comment any unrecognized character should at least result in a warning. In Hadron it's an error. Otherwise there's just too much potential for confusing bugs like this, where the code does something very different from how it looks.

lnihlen:
Seeing a similar problem in another quark, where U+2212 (mathematical minus sign) is being used in place of unary negation `-`, and since SC silently swallows it, the number is actually parsed as positive. Like I said in my previous comment, this is a breeding ground for bad bugs.

JordanHendersonMusic:
Yes, this is bad!

I wonder if there is any code out there that was written in extended ASCII? If so, then where it overwrote the controls codes, it is broken, but it would work for the higher values in some places. 

The other place fixing this might create errors is if people actually have control codes in their code. I think that's unlikely though? Probably best to just enforce that code is printable or white space.

This would also need to be addressed if we get (limited) unicode support.

---

Do you want to make a pr, if not, I'll probably get around to it at some point?

PR #7558, opened 2026-06-13
lexer: swallow invalid utf8 for the time being

### LINKED ISSUE #7509: Compiling *.sc files with unicode symbols or bare ^M carriage returns fails

JordanHendersonMusic:
Previously the lexer incorrectly let you have invalid utf8. This is bad. This pr makes it not an error, but still posts the warning.

## Purpose and Motivation

wslib is full of invalid utf8. This makes it not a true compiler error, but still posts the warning. We want these files to be cleaned up because the following is obviously invalid sc code
```supercollider
SCButton(win,18@15)
	.font_(Font("Helvetica", 9))
	.states_([["#"]])
	.action_({|val| 
Ê ÊÊ ÊÊ ÊsynthParams = Array.new; 
Ê ÊÊ ÊÊ ÊsynthParams = synthParams.add(param[1]); // synth params 
Ê ÊÊ ÊÊ ÊsynthParams = synthParams.add(param[3]); // argument values 
Ê ÊÊ ÊÊ ÊsynthParams = ['outbus', bus].add(synthParams.flop).flat.asCompileString; 
Ê ÊÊ ÊÊ Ê("Synth.new(\\" ++ this.name ++ ", " ++ synthParams ++ 
Ê ÊÊ ÊÊ ÊÊ Ê", target: " ++ tgt ++ ", addAction: \\" ++ addAct ++ ")").postln; 
Ê Ê}); 
```

This also allow unicode characters, we will want to remove this in the future but is currently (3.14) allowed.

## Types of changes

Turns error into warning.
Also fixes crash regarding nullptr deref when error occurs during building class dependancy list.
## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/lang/LangSource/PyrLexer.cpp
+++ b/lang/LangSource/PyrLexer.cpp
@@ -770,22 +770,22 @@ struct GlobalBisonLexerState {
                  "Error:\n──────────────────────────────────────────────────────────────────────────────────\n");
             if (o.is(ExtendedErrors::GotCurlyExpectedParen) || o.is(ExtendedErrors::GotSquareExpectedParen)) {
                 if (o.extra_range_of_error) {
-                    print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), *o.extra_range_of_error,
-                                     "Parenthises opened here...");
+                    print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(),
+                                     *o.extra_range_of_error, "Parenthises opened here...");
                     print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), o.range,
                                      "...was expected to be closed here with a ')'.");
                 }
             } else if (o.is(ExtendedErrors::GotCurlyExpectedSquare) || o.is(ExtendedErrors::GotParenExpectedSquare)) {
                 if (o.extra_range_of_error) {
-                    print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), *o.extra_range_of_error,
-                                     "Square bracket opened here...");
+                    print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(),
+                                     *o.extra_range_of_error, "Square bracket opened here...");
                     print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), o.range,
                                      "...was expected to be closed here with a ']'.");
                 }
             } else if (o.is(ExtendedErrors::GotParenExpectedCurly) || o.is(ExtendedErrors::GotSquareExpectedCurly)) {
                 if (o.extra_range_of_error) {
-                    print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), *o.extra_range_of_error,
-                                     "Curly bracket opened here...");
+                    print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(),
+                                     *o.extra_range_of_error, "Curly bracket opened here...");
                     print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), o.range,
                                      "...was expected to be closed here with a '}'.");
                 }
@@ -802,7 +802,8 @@ struct GlobalBisonLexerState {
                 const auto [ptr, sz] = char_stream.source_code_range_to_text(o.range);
                 const std::string example { ptr, sz };
                 const auto desc = std::string { "Expected digits after the 'e', for example '" } + example + "10'.";
-                print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), o.range, desc.c_str());
+                print_error_line(fileName, char_stream.source.c_str(), char_stream.source.size(), o.range,
+                                 desc.c_str());
             }
 
             else if (o.is(TokenType::ErSymbolQuoteUnclosed)) {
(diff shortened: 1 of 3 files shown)


JordanHendersonMusic:
@jamshark70 this fixes the issue you ran into yesterday

dyfer:
Thanks!

> This also allow unicode characters, we will want to remove this in the future but is currently (3.14) allowed.

Can you clarify in what circumstances unicode characters won't be allowed in the future?

JordanHendersonMusic:
> Can you clarify in what circumstances unicode characters won't be allowed in the future?

Well all of them (except inside string literals).

Here is one example https://github.com/supercollider/supercollider/issues/7410 where this causes absolute chaos because sc just pretends it is whitespace!

JordanHendersonMusic:
IMO this is good to go and fixes issue when people update, we can make these invalid characters actually errors in the future as the files need updating.

dyfer:
I've ran this build against my default configuration and it does detect many incorrect UTF8 characters (good/fine).
But is there supposed to be a caret to indicate there the character is? I get something like this:

```
Lexing Error:
Error: /Users/marcin/Library/Application Support/SuperCollider/downloaded-quarks/wslib/wslib-classes/Main Features/EQ/extBHiCut-magResponse.sc:10:10
    8 │ 			{ freqs = (..freqs).linexp(0,freqs-1, 20, 22000); };
    9 │ 		in = 1!freqs.size;
   10 │ 		rqs.do {
      ┆ 		         Invalid UTF8 encountered here, you probably want to delete this.

Lexing Error:
Error: /Users/marcin/Library/Application Support/SuperCollider/downloaded-quarks/wslib/wslib-classes/Main Features/EQ/extBHiCut-magResponse.sc:10:15
    8 │ 		in = 1!freqs.size;
    9 │ 		rqs.do {

```

JordanHendersonMusic:
ahh.. so this is actually a different issue. I'm going to be pushing a new commit to refactor the error printing soon I hope. Essentially there is a little bug in the lexer where it marks the newline character as starting on one line, and ending on the next, when the character itself needs to be all on the previous. I've just been waiting to get these two PR merged because other wise I'm goijng to have to deal with merge conflicts ...

dyfer:
Sounds good. I don't oppose to merging this even if posting needs refinement, but I wanted to make sure that the behavior is expected.

I can't give a proper cpp review, do you need another pair of eyes on the implementation side, or is it enough to focus on the functionality? 

As for the functionality: full disclosure, since this PR deals with cpp and lexer which I know very little about, I used Copilot to help me with the review. It created an overview and suggested follow up points, which I used to increase my understanding of what's going on.

> Also fixes crash regarding nullptr deref when error occurs during building class dependancy list.

How would this crash be triggered? Is this something we could add a test for? Maybe it's worth filing as a (minimal) issue so that there's a trace of it?

Another concern from the LLM was whether the lexer continues predictably after the warning. It seemed so when running on my configuration (which is full of invalid characters), but I don't know if it can be tested any more than that. Do you think there may be any way where lexer does not continue properly after the warning? Is this something I can test?

Finally, it would be good to have a single issue that summarizes the recent Unicode handling changes, the trajectory, and possibly a separate discussion for when are we going to turn these warnings into errors. This is not a requirement for this PR, but something we should do soon. Sorry to throw it your way, but since you've been in the center of these discussions, I think you are in the best position to summarize that issue. The objective would be to have a succinct text that gives the birds-eye view of the changes/timeline (with no or minimal discussion) and links to detailed discussions regarding particular issues. Is this something you could eventually create?

JordanHendersonMusic:
> How would this crash be triggered? Is this something we could add a test for? Maybe it's worth filing as a (minimal) issue so that there's a trace of it?

This probably isn't worth it. See, the compiler is just a bit of a mess when it comes to state management because everything is global. Previously gCompilingFileSym wasn't set when in cli mode, but we are going to want to know this if possible. I've updated this so it isn't null in any case, but it's worth having a null check like I've done in any case (when I refactor this the state will vanish).

> Another concern from the LLM was whether the lexer continues predictably after the warning

Check out the standalone executable, as the compiler just bails when the lexer reports an error,  the standalone continues until the end of the file.

But yes it is fine.


> I can't give a proper cpp review, do you need another pair of eyes on the implementation side, or is it enough to focus on the functionality?

Functionality is good to be honest! The stuff in pyrlexer just isn't worth digging into because I'm trying to rewrite it all (had an issue with unit testing along the way!). 

> Finally, it would be good to have a single issue that summarizes the recent Unicode handling changes, the trajectory, and possibly a separate discussion for when are we going to turn these warnings into errors.

Yes! I'll do that soon.

There ought not be any functionality changes right now. The changes should all happen later, but basically: all files should be valid utf8 (because we need **a** standard and this one is the best); and, no more counting things we don't recognise as whitespace, unicode will no longer be white shape, neither will random control characters or unusual ascii stuff - these should not appear in source code. The **vast** majority of code already does it, there are just some weird cases out there! For the unicode stuff, like using the unicode ellipse over dot dot dot, unicode actually has a nice table somewhere and I'll make the errors print suggestions on what should be written instead... I've seen you've written ... Did you mean ... (That sort of thing).

JordanHendersonMusic:
TODO: add a more descriptive explanation of why this is now broken so the user understands the trade off.

prko:
This looks complete, but merge conflicts are still blocking progress.

JordanHendersonMusic:
This should be good to merge now.

capital-G (approved):
feel free to merge - I only skimmed over it, so if you want an actual review you should wait for someone w/ some lexer xp to review this ;)

PR #7567, opened 2026-06-16
TestServer_dumpTree: switch to NRT server

dyfer:
## Purpose and Motivation

The supernova version of the `TestServer_dumpTree` test seems to hang at times:
https://github.com/supercollider/supercollider/actions/runs/27577542122/job/81532796978#step:7:12155
I was able to recreate failure by putting artificial load on my system and running the test locally. 

```sh
CORES=$(sysctl -n hw.physicalcpu); echo "Running 'yes' on $CORES cores..."; for i in $(seq 1 $CORES); do yes > /dev/null & done; wait; echo "Done."
```
Under load we tend to miss UDP packets and the process never quits, particularly in supernova. Since this test specifically depends on arrival of a single packet to trigger dumpTree and to quit, and since there's no way to unhang sclang when it waits for the input from the pipe, I propose that these tests are switched to using a non-realtime server.

The output parsing is a bit arbitrary (we're waiting for specific strings to appear), but I think it's in line with the original implementation.

The diff is noisy since I applied default IDE's formatting (the original was submitted with spaces not tabs).

BTW looking at another failure, we should probably do the same for the [TestServer_nodeMove test](https://github.com/supercollider/supercollider/actions/runs/27581418135/job/81543585825#step:7:304), which also depends on capturing server's output with a Pipe.

EDIT: it turns out that on Linux we don't get a full output out of supernova - I tried with redirecting logging to a file and it's the same as reading output directly, so I think it's unlikely just an output buffering issue. LLM suggests it's an issue of the logging thread being shut down before it manages to process all output, but I don't know if that's the right track. Leaving this as a draft for now.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestServer_dumpTree.sc
+++ b/testsuite/classlibrary/TestServer_dumpTree.sc
@@ -45,8 +45,7 @@ TestServer_dumpTree : UnitTest {
 				// ignore output until we see matching start
 				if (line.beginsWith("NODE TREE")) {nodeTreeOutputDectected = true};
 				// once we see matching beginning, capture output
-				// strip carriage return because of Windows
-				if (nodeTreeOutputDectected) {actualOutput.add(line.replace("\r", ""))};
+				if (nodeTreeOutputDectected) {actualOutput.add(line)};
 				// stop capturing when we see matching end
 				if (line.beginsWith("END NODE")) {nodeTreeOutputDectected = false};
 				line = logFile.getLine;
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
In #7574 I've added a loop counter which throws an exception if it is exceeded. This stop these from hanging indefinitely and blocking all sc code — a bit of  a temporary fix.

It could be that the server isn't flushing before shutting down as the LLM suggests.

PR #7569, opened 2026-06-16
Add Flatpak build

### LINKED ISSUE #3604: Create flatpak distribution for Linux

capital-G:
## Purpose and Motivation

Closes https://github.com/supercollider/supercollider/issues/3604

Flatpak is a way to install software on linux w/o going through the native package manager of the OS. This leads to beefier binaries and resources-usage, but allows people to use the newest SC version w/o having to build it themselves or wait for their package manager to update SC.
This makes the installation of SC in linux much more easier.

Used https://docs.flatpak.org/en/latest/sandbox-permissions.html and https://docs.flatpak.org/en/latest/module-sources.html for the more "advanced" sandboxing and bundling stuff.

I tested this on my fedora 42 x86 laptop (build and run the flatpak) and on the virtual fedora 42 arm virtual machine via my macbook and it also played back sound.
Please check if you can run this on your linux machine - it should be as easy as downloading the binary artifact and double-click on the file which will install SC via flatpak on your system (assuming that flatpak is installed).
It allowed me to install a flatpak sc beneath a native-build sc - it may be possible to have multiple flatpak sc next to each other by looking into flatpak branches, but I leave this to the linux users.

The next step could be to publish this also via flathub - but I am not too eager about that b/c this requires some synchronization w/ the flathub team b/c we can't simply upload the binaries through our CI and we have to create a dedicated repo for that... Maybe someone who actually uses linux could take care of this ;) I am personally fine w/ providing this as download, maybe we could add this as nightly build download to the website.

Before we merge, we should be sure that this is actually a wanted feature: The feedback I got when I communicated that there will be a flatpak soon was "why flatpak and not [appimage](https://appimage.org/)". This also seems nice, but is even more statically linked which is even more worrying, but I let the linux heads decide on that ;) But after working some time w/ flatpak now, I don't consider it that nice tbh - it feels more like an appstore thing instead of "get this running".

## Types of changes

- New feature

## To-do list

- [ ] Read into jack support, e.g.  https://github.com/flatpak/flatpak/issues/3622#issuecomment-704606905 
- [ ] Test `sclang` and `scsynth` as commands
- [ ] Test multi-channel audio
- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -133,8 +133,15 @@ jobs:
             s3-artifact-extension: 'zip'
             s3-create-latest-link: true # create link to pointing to the "latest" build
 
+          - artifact-suffix: flatpak-x86
+            s3-os-name: flatpak-x86
+            s3-artifact-prefix: '-flatpak-x86'
+            s3-artifact-suffix: ''
+            s3-artifact-extension: 'zip'
+            s3-create-latest-link: true # create link to pointing to the "latest" build; activate only one per branch per s3-os-name
+
     if: github.repository_owner == 'supercollider' && github.event_name != 'pull_request' # run in the main repo, but not on pull requests
-    needs: [lint, version-getter, build-macos, build-windows]
+    needs: [lint, version-getter, build-macos, build-windows, build-flatpak]
     runs-on: ubuntu-latest
     name: 'deploy ${{ matrix.artifact-suffix }} to s3'
     env:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -207,7 +207,7 @@ jobs:
   # release - list of files uploaded to GH release is specified in the *upload* step
   deploy_gh:
     if: startsWith(github.ref, 'refs/tags/') # run on tagged commits
-    needs: [lint, version-getter, build-macos, build-windows]
+    needs: [lint, version-getter, build-macos, build-windows, build-flatpak]
     runs-on: ubuntu-latest
     name: 'deploy release'
     env:
@@ -228,4 +228,6 @@ jobs:
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-macOS-x64-legacy.dmg/*
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64-installer/*
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-x86_64.flatpak/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-aarch64.flatpak/*
           draft: true
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -133,11 +133,12 @@ jobs:
             s3-artifact-extension: 'zip'
             s3-create-latest-link: true # create link to pointing to the "latest" build
 
-          - artifact-suffix: flatpak-x86
-            s3-os-name: flatpak-x86
-            s3-artifact-prefix: '-flatpak-x86'
+          - artifact-suffix: x86_64.flatpak
+            s3-os-name: x86_64-flatpak
+            s3-artifact-prefix: 'x86_64-flatpak '
             s3-artifact-suffix: ''
-            s3-artifact-extension: 'zip'
+            s3-artifact-extension: 'flatpak'
+            artifact-extension: '.flatpak'
             s3-create-latest-link: true # create link to pointing to the "latest" build; activate only one per branch per s3-os-name
 
     if: github.repository_owner == 'supercollider' && github.event_name != 'pull_request' # run in the main repo, but not on pull requests
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -64,6 +64,13 @@ jobs:
       sc-version: ${{ needs.version-getter.outputs.sc-version }}
     secrets: inherit
 
+  build-flatpak:
+    needs: [lint, version-getter]
+    uses: ./.github/workflows/build_flatpak.yml
+    with:
+      sc-version: ${{ needs.version-getter.outputs.sc-version }}
+    secrets: inherit
+
   build-macos:
     needs: [lint, version-getter]
     uses: ./.github/workflows/build_macos.yml
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -133,9 +133,17 @@ jobs:
             s3-artifact-extension: 'zip'
             s3-create-latest-link: true # create link to pointing to the "latest" build
 
-          - artifact-suffix: x86_64.flatpak
-            s3-os-name: x86_64-flatpak
-            s3-artifact-prefix: 'x86_64-flatpak '
+          - artifact-suffix: "x86_64.flatpak"
+            s3-os-name: flatpak-x86_64
+            s3-artifact-prefix: '-x86_64-flatpak'
+            s3-artifact-suffix: ''
+            s3-artifact-extension: 'flatpak'
+            artifact-extension: '.flatpak'
+            s3-create-latest-link: true # create link to pointing to the "latest" build; activate only one per branch per s3-os-name
+
+          - artifact-suffix: "arm64.flatpak"
+            s3-os-name: flatpak-arm64
+            s3-artifact-prefix: '-arm64-flatpak'
             s3-artifact-suffix: ''
             s3-artifact-extension: 'flatpak'
             artifact-extension: '.flatpak'
@@ -237,5 +245,5 @@ jobs:
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64-installer/*
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64/*
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-x86_64.flatpak/*
-            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-aarch64.flatpak/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-arm64.flatpak/*
           draft: true
--- a/platform/linux/online.supercollider.SuperCollider.yml
+++ b/platform/linux/online.supercollider.SuperCollider.yml
@@ -33,18 +33,7 @@ modules:
     config-opts:
       - -DCMAKE_BUILD_TYPE=Release
       - -DAUDIOAPI=jack
-      - -DSC_QT=ON
-      - -DSC_IDE=ON
-      - -DSC_USE_QTWEBENGINE=ON
-      - -DSUPERNOVA=ON
-      - -DSC_ABLETON_LINK=ON
-      - -DSC_HIDAPI=ON
-      - -DSYSTEM_BOOST=OFF
-      - -DSYSTEM_YAMLCPP=OFF
-      - -DNO_AVAHI=OFF
-      - -DINSTALL_HELP=ON
       - -DENABLE_TESTSUITE=OFF
-      - -DSC_SYMLINK_CLASSLIB=OFF
       - -DCMAKE_INSTALL_PREFIX=/app
     sources:
       - type: dir
--- a/platform/linux/online.supercollider.SuperCollider.yml
+++ b/platform/linux/online.supercollider.SuperCollider.yml
@@ -34,7 +34,7 @@ modules:
       - -DAUDIOAPI=jack
       - -DSC_QT=ON
       - -DSC_IDE=ON
-      - -DSC_USE_QTWEBENGINE=ON  # arm does not support qt webengine
+      - -DSC_USE_QTWEBENGINE=ON
       - -DSUPERNOVA=ON
       - -DSC_ABLETON_LINK=OFF
       - -DSC_HIDAPI=ON
--- a/platform/linux/online.supercollider.SuperCollider.yml
+++ b/platform/linux/online.supercollider.SuperCollider.yml
@@ -15,6 +15,7 @@ finish-args:
   - --filesystem=host
   # scsynth uses jack, but pipewire is jack compatible
   - --filesystem=xdg-run/pipewire-0
+  - --system-talk-name=org.freedesktop.Avahi
 
 # remove dev files from build
 cleanup:
@@ -36,11 +37,11 @@ modules:
       - -DSC_IDE=ON
       - -DSC_USE_QTWEBENGINE=ON
       - -DSUPERNOVA=ON
-      - -DSC_ABLETON_LINK=OFF
+      - -DSC_ABLETON_LINK=ON
       - -DSC_HIDAPI=ON
       - -DSYSTEM_BOOST=OFF
       - -DSYSTEM_YAMLCPP=OFF
-      - -DNO_AVAHI=ON
+      - -DNO_AVAHI=OFF
       - -DINSTALL_HELP=ON
       - -DENABLE_TESTSUITE=OFF
       - -DSC_SYMLINK_CLASSLIB=OFF
--- a/platform/linux/online.supercollider.SuperCollider.yml
+++ b/platform/linux/online.supercollider.SuperCollider.yml
@@ -1,0 +1,60 @@
+app-id: online.supercollider.SuperCollider
+runtime: org.kde.Platform
+runtime-version: '6.10'
+sdk: org.kde.Sdk
+base: io.qt.qtwebengine.BaseApp
+base-version: '6.10'
+command: scide
+
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --share=network
+  - --device=all
+  - --filesystem=host
+  # scsynth uses jack, but pipewire is jack compatible
+  - --filesystem=xdg-run/pipewire-0
+
+# remove dev files from build
+cleanup:
+  - /include
+  - /lib/cmake
+  - /lib/pkgconfig
+  - /share/man
+  - '*.a'
+  - '*.la'
+
+modules:
+  - name: supercollider
+    buildsystem: cmake-ninja
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DAUDIOAPI=jack
+      - -DSC_QT=ON
+      - -DSC_IDE=ON
+      - -DSC_USE_QTWEBENGINE=ON  # arm does not support qt webengine
+      - -DSUPERNOVA=ON
+      - -DSC_ABLETON_LINK=OFF
+      - -DSC_HIDAPI=ON
+      - -DSYSTEM_BOOST=OFF
+      - -DSYSTEM_YAMLCPP=OFF
+      - -DNO_AVAHI=ON
+      - -DINSTALL_HELP=ON
+      - -DENABLE_TESTSUITE=OFF
+      - -DSC_SYMLINK_CLASSLIB=OFF
+      - -DCMAKE_INSTALL_PREFIX=/app
+    sources:
+      - type: dir
+        path: ../../
+    post-install:
+      - mv /app/share/applications/SuperColliderIDE.desktop
+        /app/share/applications/online.supercollider.SuperCollider.desktop
+      - desktop-file-edit
+        --set-key=Icon --set-value=online.supercollider.SuperCollider
+        /app/share/applications/online.supercollider.SuperCollider.desktop
+      - mv /app/share/icons/hicolor/scalable/apps/sc_ide.svg
+        /app/share/icons/hicolor/scalable/apps/online.supercollider.SuperCollider.svg
+      - install -Dm644 $FLATPAK_BUILDER_BUILDDIR/icons/sc_ide_128.png
+        /app/share/icons/hicolor/128x128/apps/online.supercollider.SuperCollider.png
--- a/.github/workflows/build_flatpak.yml
+++ b/.github/workflows/build_flatpak.yml
@@ -15,11 +15,11 @@ jobs:
         include:
           - arch: x86_64
             runner: ubuntu-24.04
-            jobname: "x86 64bit"
+            job-name: "x86 64bit"
             artifact-suffix: "x86_64.flatpak"
           - arch: aarch64
             runner: ubuntu-24.04-arm
-            jobname: "arm64"
+            job-name: "arm64"
             artifact-suffix: "arm64.flatpak"
 
     steps:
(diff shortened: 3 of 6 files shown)


dyfer:
> "why flatpak and not [appimage](https://appimage.org/)"

Hahaha that was my question too!

I agree that active Linux users should help make these decisions. My intuition is that ultimately providing both Flatpak and Appimage would be great.

EDIT: a quick test on debian 12 in a VM (aarch64):
I installed flatpak with apt and then ran `flatpak install SupeCollder.....flatpak`
Result:

```
error: The application online.supercollider.SuperCollider/aarch64/master requires the runtime org.kde.Platform/aarch64/6.10 which was not found
```
Do we need to depend on KDE in the flatpak build? One can certainly run Qt apps under other desktop environemnts.

capital-G:
Some things I haven't checked (b/c I don't know how to uninstall custom built SC in linux): is sclang and scsynth available as commands after installation?

> Do we need to depend on KDE in the flatpak build?

Ah no - but flatpak is a bit more "barebones" than one expect (e.g. I'd expected it to catch the dependencies automatically from a global index).

You have to add the index, can you try again after adding flathub as package index to flatpak, see https://wiki.archlinux.org/title/Flatpak#Add_a_repository? Though I am unsure about the `--user` flag here^^

```shell
flatpak remote-add --if-not-exists --user flathub https://dl.flathub.org/repo/flathub.flatpakrepo
```


> > "why flatpak and not [appimage](https://appimage.org/)"
> 
> Hahaha that was my question too!

IIUC:

Flatpak: Distributed like in an app store, app has an unique identifier like in java. You catch dependencies from a package index. This also feels like a native app in e.g. KDE/gnome
AppImage: The whole app is self contained and can be clicked on like through `chmod +x`, you have to take care of adding this to your desktop environment as an app and add e.g. path variables.

platform/linux/online.supercollider.SuperCollider.yml
```diff
@@ -0,0 +1,60 @@
...
+      - -DSUPERNOVA=ON
+      - -DSC_ABLETON_LINK=OFF
+      - -DSC_HIDAPI=ON
+      - -DSYSTEM_BOOST=OFF
+      - -DSYSTEM_YAMLCPP=OFF
+      - -DNO_AVAHI=ON
+      - -DINSTALL_HELP=ON
+      - -DENABLE_TESTSUITE=OFF
+      - -DSC_SYMLINK_CLASSLIB=OFF
+      - -DCMAKE_INSTALL_PREFIX=/app
```
dyfer, line 37: I'm not sure about specifying all the flags here, I'd be inclined to align it better with build flags in Linux CI and possibly leaving more at default values.

Specifically:
- `# arm does not support qt webengine` - I don't think that's true? I installed this flatpak on virtual arm64 machine and the web browser was present
- why turning Ableton Link OFF? it is ON by default
- why NO_AVAHI?

platform/linux/online.supercollider.SuperCollider.yml
```diff
@@ -0,0 +1,60 @@
...
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --share=network
+  - --device=all
+  - --filesystem=host
+  # scsynth uses jack, but pipewire is jack compatible
+  - --filesystem=xdg-run/pipewire-0
+
+# remove dev files from build
```
dyfer, line 20: Why do we want to do that? Shouldn't we be able to build plugins against SC from flatpak?
(I know we don't ship them on Windows and macOS, but I think it would be nice, in fact)

.github/workflows/actions.yml
dyfer, line 120: I think we should add Flatpak to the releases as well?

dyfer: If we add flatpak to the release artifacts, we should add `build-flatpak` to this list as well

dyfer:
After adding the flathub repository this totally worked on Debian 12. That's fantastic!

As for Flatpak vs AppImage I vote for both (eventually) - but I'm also not a regular Linux user, so...

I guess the next step would be to gather more feedback from the Linux users, correct?

platform/linux/online.supercollider.SuperCollider.yml
capital-G: > * `# arm does not support qt webengine` - I don't think that's true? I installed this flatpak on virtual arm64 machine and the web browser was present

Oh, this is a leftover... I worked on arm emulator and I thought it was an arm issue but I fixed it by using `base: io.qt.qtwebengine.BaseApp`.

I had some problems w/ service discovery b/c of the sandbox, which AVAHI and Link is relying on, so I deactivated those. There is probably some way to request access to this from within the sandbox? I didn't want to spend too much time on it w/o having a "this seems something that we would like to have".

platform/linux/online.supercollider.SuperCollider.yml
capital-G: Yeah, I was also surprised but this is how you do things in flatpak -> https://docs.flatpak.org/en/latest/manifests.html#cleanup

.github/workflows/actions.yml
```diff
@@ -221,4 +228,6 @@ jobs:
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-macOS-x64-legacy.dmg/*
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64-installer/*
             ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-win64/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-x86_64.flatpak/*
+            ${{ env.INSTALL_PATH }}/${{ env.ARTIFACT_FILE_PREFIX }}-aarch64.flatpak/*
```
capital-G: @dyfer is it correct to use the `/*` here as well? Btw, I couldn't find `win64-installer` in our artifacts.

dyfer: > @dyfer is it correct to use the `/*` here as well? 

IIRC Yes - I think this is because GHA artifacts are in double zips, so in this job they get unpacked to installpath/zipname/zipfile.zip. 

> Btw, I couldn't find `win64-installer` in our artifacts.

Installer is only created for releases, IIRC.

Also, it's fine to have extra paths here, e.g. for builds we disabled but may want to bring back etc (e.g. macOS-arm64).

platform/linux/online.supercollider.SuperCollider.yml
dyfer: Got it, in that case it's fine I guess :)

platform/linux/online.supercollider.SuperCollider.yml
dyfer: I agree that these functions are probably less used than other areas of SC, but they have been part of SC for some time. 

If they are disabled in flatpak, we need to be loud and clear in the docs about it (not sure what the best place would be, probably Linux readme? and the first release where we introduce flatpak)

platform/linux/online.supercollider.SuperCollider.yml
```diff
@@ -0,0 +1,60 @@
...
+command: scide
+
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --share=network
+  - --device=all
+  - --filesystem=host
+  # scsynth uses jack, but pipewire is jack compatible
```
dyfer, line 16: should we keep this comment?

platform/linux/online.supercollider.SuperCollider.yml
capital-G: Ah sorry, with

> I didn't want to spend too much time on it w/o having a "this seems something that we would like to have".

I referred to Flatpak, no the implementation of AVAHI/Link. If you think we should go for a flatpak, then I can look further into Link/AVAHI.

platform/linux/online.supercollider.SuperCollider.yml
capital-G: Unsure - it may be obvious but it is a bit strange since we currently do not support pipewire beyond the jack/pipewire shim.
I think it doesn't hurt having it.

platform/linux/online.supercollider.SuperCollider.yml
dyfer: Ah got it.
And just to clarify on my end, we _could_ consider shipping flatpak with these off, but we would need to document it.
IMO we should go for flatpak package, it's a straightforward way of getting current SC on Linux. We could additionally consider AppImage. (but again, I'm not a Linux user).

platform/linux/online.supercollider.SuperCollider.yml
dyfer: Yeah, I was thinking it was a bit obvious, but it is to me where I'm aware of SC, jack, and pipewire.
Fine to leave it in then.

nuss: If I may I'd like to to make a remark on my experiences with audio applications installed via flatpak, using jack/pipewire. I'm currently on debian 13, _not_ using a standard pipewire/jack configuartion (i.e. not using `pipewire-jack` but instead running pipwire on top of jack  (`jackd2`)). I have though experiences with installing apps  running pipewire-jack on debian 12. 
What I can say is that apps (e.g. Audacity, Reaper) installed through flatpak are routed differently into jack: In the following Screenshot from the QJackCtl patchbay you can see that the portaudio inputs to Audacity (3.7.3, installed with apt) are being created **within** the jack client 
audacity_3 7 3_aptitude
... whereas using Audacity 3.7.8 installed from flatpak gets its portaudio inputs **within pipewire through Jack SOURCE** (screenshot from qpwgraph) audacity_3 7 8_flatpak
This goes for Reaper just as well.

The obvious disadvantage, though one can somehow say audio over jack "is working", Jack SOURCE will only provide two channels, all audio going to or coming from jack enabled programs will be mixed down to two channels. 

I've tried installing apps from flatpak on an environment where jack capabilities were provided through pipewire-jack rather than jack (using debian 12) but I never got it to work properly there either (can't remember exactly anymore but I believe it was the same reason. I vaguely remember Audacity didn't work at all using the jack backend...).

Maybe you're aware of all this already - I don't know. Maybe this place in the config is exactly the place where this peculiarity can be tweaked?

capital-G: Flatpak does not support native jack, only pipewire is supported - see https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/work_items/1001#note_323461581
Additionally, IIRC, jack is already removed from most distros in favor of pipewire (at least >= ubuntu 24.04 and >= debian 13).

Flatpak is sandboxed which comes w/ its own difficulties and limitations, we can hope that the situation improves over time.
Though I read that by adding `--filesystem=xdg-run/pipewire-0` it is possible to have multiple audio i/o channels.

Can you test the flatpak on your local machine? I only have superficial knowledge about jack/pipewire/linux audio, so it would be great to have some feedback from you.

nuss: I would do so if I knew how to test a flatpak build locally. (I really hope it will be possible to still install Jack manually after debian 13...)

capital-G: > I would do so if I knew how to test a flatpak build locally.

You can download it from the CI ;) Scroll down to artifacts here https://github.com/supercollider/supercollider/actions/runs/27755911383?pr=7569 and then you can download the flatpak.

I may have to look into https://github.com/flatpak/flatpak/issues/3622#issuecomment-704606905 as well such that jack gets removed as much as possible from the flatpak.

nuss: Ah, thanks! So, this is the result of my first attempt to run the flatpak:
```
$ flatpak run online.supercollider.SuperCollider 
Qt: Session management error: Could not open network socket
[2:21:0619/140940.773508:ERROR:bus.cc(408)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: Datei oder Verzeichnis nicht gefunden
xkbcommon: ERROR: [XKB-338] Couldn't find file "symbols/eurkey" in include paths
xkbcommon: ERROR: [XKB-338] 1 include paths searched:
xkbcommon: ERROR: [XKB-338] 	/usr/share/X11/xkb
xkbcommon: ERROR: [XKB-661] Couldn't process include statement for 'eurkey'
xkbcommon: ERROR: [XKB-769] Abandoning symbols file "(unnamed map)"
xkbcommon: ERROR: Failed to compile xkb_symbols
xkbcommon: ERROR: [XKB-822] Failed to compile keymap
[2:20:0619/140940.818465:FATAL:xkb_keyboard_layout_engine.cc(661)] Keymap file failed to load: eurkey
```
... `eurkey` is my current keyboard layout. Haven't experienced any problems with it running any flatpack-installed software so far...

capital-G: Ah, thx for reporting. This seems like an upstream problem, see https://github.com/flathub/io.qt.qtwebengine.BaseApp/issues/402 and https://github.com/zealdocs/zeal/issues/1890

Don't know what to do about that though... Didn't run into problems in Fedora44/Wayland/Dvorak layout.

nuss: I'll try with a different keyboard layout (after lunch)

nuss: ok, I've uninstalled the eurkey keyboard now. I can run SC now. I booted the server, unfortunately with the result that I've seen with all other flatpak audio apps: SC's input is routed through JACK Source,to jack  its output through JACK Sink to system output:
flatpak_supercollider_routing_qpwgraph
As you can see I can easily create in and output channels but sink and source only provide two channels. Also ins and outs on sink and source are differently encoded - they're not called `capture_1`,`capture_2` but `capture_FL`, `capture_FR`. I guess you could have more channels but they are meant for surround encoding, not SC channels 0, 1,  2, 3, 4, etc.  - I think SC shouldn't get routed this way.
Also I've installed SC flatpak on some other machine that's currently standing around here, running pipewire-jack, not jackd2 in debian 12. Here SC installed from flatpak didn't boot the server at all!:
```
*** Welcome to SuperCollider 3.15.0-dev. *** For help press Ctrl-D.
file "/home/stefan/.var/app/online.supercollider.SuperCollider/data/SuperCollider/Help/scdoc_version" does not exist.
SCDoc: refresh static files
SCDoc: Indexing help-files...
SCDoc: Indexed 1396 documents in 1.19 seconds
ERROR: terminate called without an active exception
```
I don't understand why audio is handled that way in flatpak installed apps. I needed e.g. to install Reaper in my former day job, running under debian. I ended up installing Reaper manually (installer from their website) as the program installed from flatpak turned out to be pretty useless b/c of its audio routing.

capital-G: Hey, thanks for the feedback. I don't really get why the server wouldn't boot - "it works on my machine" w/ fedora 44.

```
Starting interpreter with "sclang -i scqt" in ""
SC FFT global init: cosTable initialised.
compiling class library...
	Found 884 primitives.
	Compiling directory '/app/share/SuperCollider/SCClassLibrary'
	Compiling directory '/home/dennis/.var/app/online.supercollider.SuperCollider/data/SuperCollider/Extensions'
	numentries = 865224 / 13116984 = 0.066
	5708 method selectors, 2298 classes
	method table size 13910184 bytes, big table size 104935872
	Number of Symbols 12943
	Byte Code Size 411538
	compiled 333 files in 1.37 seconds

Info: 4 methods are currently overwritten by extensions. To see which, execute:
MethodOverride.printAll

compile done
Couldn't set realtime scheduling priority 1: Operation not permitted
Class tree inited in 0.13 seconds


*** Welcome to SuperCollider 3.15.0-dev. *** For help press Ctrl-D.
Couldn't set realtime scheduling priority 1: Operation not permitted
SCDoc: Indexing help-files...
SCDoc: Indexed 1396 documents in 1.33 seconds
-> localhost
SC FFT global init: cosTable initialised.
JackDriver: client name is 'SuperCollider'
JackDriver: connected  Studio 1824c Analog Surround 7.1:capture_FL to SuperCollider:in_1
JackDriver: connected  Studio 1824c Analog Surround 7.1:capture_FR to SuperCollider:in_2
JackDriver: connected  SuperCollider:out_1 to Studio 1824c Analog Surround 7.1:playback_FL
JackDriver: connected  SuperCollider:out_2 to Studio 1824c Analog Surround 7.1:playback_FR
SuperCollider 3 server ready.
localhost: server process's maxLogins (1) matches with my options.
```

Maybe some sandboxing regarding audio is not working? Or do you have the jack/pipewire server started?

And qjackctl looks like this for me like this, which I think is fine?

Screenshot From 2026-07-08 13-17-42

I don't know how we could fix things beyond that - i.e. how multiple channels are displayed/organized within jack/pipewire is a jack setting iirc - maybe @functionaljerk can chime in here?

One thing I am not too happy about: flatpak can only expose files via a desktop file, so running sclang is possible but only via

```
 flatpak run --command=sclang online.supercollider.SuperCollider
```

people could add some alias to their `~./zshrc` (or else), but this can not be documented properly through flatpak itself - but I'll try to document this somewhere...

nuss: That screenshot looks good. Are you using "Analog Surround 7.1" intentionally? If you connect more than `out_1` and `out_2` from SuperCollider to subsequent playback ports, what comes out of your speakers? Will the output  Surround 7.1 encoded? That looks a bit odd. But maybe it's just the naming.

capital-G: I don't use 7.1 and also don't use linux everyday - IIRC there is a jack setting/config how a multi-channel device will be "resolved"/"displayed".

> Will the output Surround 7.1 encoded?

What would be 7.1 encoding in that case? .1 could have a lowpass applied, but I don't think that jack does things like this?

capital-G:
@dyfer @nuss do you think we should merge this and publish this as an experimental way of running sc under linux?
It would be great if you could double-check the s3-release upload marcin!
I added ableton link and avahi support and it seems to work.

Once this gets merged I'd also like to include the download links to the website, similar to

https://github.com/supercollider/supercollider.github.io/blob/50083a3356d73daa679b09df23f01b95b6996162/_data/downloads.yml#L43-L48

platform/linux/online.supercollider.SuperCollider.yml
dyfer: I still don't see a good reason to setting the default flags (I _think_ that's everything except `CMAKE_BUILD_TYPE`, `ENABLE_TESTSUITE`, and `CMAKE_INSTALL_PREFIX`?), but it's rather cosmetic, so if you prefer to keep them, then it's fine.

capital-G: reduced it down to

```
      - -DCMAKE_BUILD_TYPE=Release
      - -DAUDIOAPI=jack
      - -DENABLE_TESTSUITE=OFF
      - -DCMAKE_INSTALL_PREFIX=/app
```

dyfer:
> @dyfer @nuss do you think we should merge this and publish this as an experimental way of running sc under linux?

Yes, I'm in favor of this!

> It would be great if you could double-check the s3-release upload marcin!

The code looks good.
Testing this for real involves creating a release. Could you create a test release on your fork to check if the artifacts are uploaded properly? Creating a release on a fork is less disruptive than creating it on the main repo.

In a similar vein, you could test s3 upload by just pushing your branch to the main repo. But I don't think there will be any issues there and if something comes up after merging the PR we can quickly fix it before adding the download link to the website.

> I added ableton link and avahi support and it seems to work.

Thanks!

capital-G:
> Could you create a test release on your fork to check if the artifacts are uploaded properly? Creating a release on a fork is less disruptive than creating it on the main repo.

How do I do this? I thought merging to main would do this, but apparently not b/c the owner is not supercollider

https://github.com/capital-G/supercollider/actions/runs/29036327499

Do I have to adjust the `actions.yml` file? This is really cumbersome as it takes really long to get to the deploy step on forks.

dyfer:
> How do I do this? I thought merging to main would do this, but apparently not b/c the owner is not supercollider

Just create a release on your fork: https://github.com/capital-G/supercollider/releases
Specify the tag to go on the `flatpack-build` branch. 

BTW IIUC for a test release you can use any tag, doesn't have to be `Version-x.y.z`.

EDIT: no need to adjust actions.yml

capital-G:
Thx marcin!

Here is the deploy release action: https://github.com/capital-G/supercollider/actions/runs/29039602201/job/86196714699
Here are the release artifacts: https://github.com/capital-G/supercollider/releases/tag/flatpak

dyfer:
> Here are the release artifacts: https://github.com/capital-G/supercollider/releases/tag/flatpak

Nice! It works!

Also this reveals a discrepancy in naming of the Windows installer vs all other artifacts... but it's not a big deal (and def not for this PR haha)

dyfer (approved):
I think we should go for it.

@nuss do you think there are any major issues to address before we merge this?

nuss:
I'd say go for it! I still remain a bit skeptical about that JACK routing thing - I have seen this within _all_  JACK capable audio applications installed over flatpak - but I can't really test it any further at the moment. Let's see...

capital-G:
image

🫠

platform/linux/online.supercollider.SuperCollider.yml
FunctionalJerk: I successfully installed the flatpak and everything seems to be working fine on my system (Ubuntu 24.04, pipewire, XFCE). 

I get these Errors when booting the server: 
```
*** Welcome to SuperCollider 3.15.0-dev. *** For help press Ctrl-D.
ERROR: qt.qpa.theme.gnome: dbus reply error: [ "org.freedesktop.DBus.Error.UnknownMethod" ] "No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop"
qt.qpa.theme.gnome: dbus reply error: [ "org.freedesktop.DBus.Error.UnknownMethod" ] "No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop"
file "/home/dizzy/.var/app/online.supercollider.SuperCollider/data/SuperCollider/Help/scdoc_version" does not exist.
```  

But so far I don't see them affecting anything.

prko:
Flatpak is amazing!

---

I had two problems: 

---

```shell
parallels@ubuntu-gnu-linux-24-04-3:~/Downloads$ flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo

Note that the directories 

'/var/lib/flatpak/exports/share'
'/home/parallels/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

error: Can't load uri https://dl.flathub.org/repo/flathub.flatpakrepo: While fetching https://dl.flathub.org/repo/flathub.flatpakrepo: [60] SSL peer certificate or SSH remote key was not OK
```
However, I could resolve this by the assistance of AI:
```
sudo apt update
sudo apt install --reinstall ca-certificates
sudo update-ca-certificates
```

---

Then,
```shell
sudo timedatectl set-ntp true
```

I am not sure if it should be documented.

PR #7572, opened 2026-06-17
Allow passing scsynth and supernova paths in language config adding infrastructure for easier building and testing

### LINKED ISSUE #7573: extend language config allowing things to be passed to the runtime - pass path of server
## Motivation

The server path to the runtime. 

Very useful for testing.

Make main sclang functions take an options arg.

The is an API change to libsclang.

How to access this information I'm undecided on. 

1. Have an unordered_map in vmglobals and access this with a primitive taking a key.

1. Have an option intrinsic class stored in vmglobals.

First is easy to expand, but means the keys are runtime values, not known at compile time. Second is easier to use incorrectly, but means adding pinning to the GC - we should do this anyway.

JordanHendersonMusic:
## Purpose and Motivation

> ...  I just want to be able to run the test suite on the current build, I don't want anything to touch path and I don't want it to look in my system config or even user. I don't want to install anything. I just want to pass all information in with one config file.

closes #7573 

Also adds a bunch of vscode tasks. There are two visible, one that runs the CI in the build dir and the other just launches sclang.

This cuts out the ide which is often a pain to debug for some reason and breakpoints in the language aren't respected. 

## Types of changes

- New feature


JordanHendersonMusic:
Hmmm, this doesn't work as it doesn't use the server's from the build directory...
Meaning it is only useful for testing sclang.

JordanHendersonMusic:
Okay I think we need to be able to pass in the path of scsynth and supernova in the config file, and pass an options struct when we boot sclang. Currently we are doing something odd and look in the dirrectory the executeable is found for the server process (this is done inside sclang).

I think we should also add some meta data on the tests so we can just run the language ones or only the server ones.

capital-G:
Is there an advantage of making this a vscode task over a CMake target which could be picked up by other IDEs?

One thing that got me to avoid going for this was its runtime: A testsuite run takes around 10 minutes, and most of the time it is easier to configure the IDE and just run the necessary tests.

JordanHendersonMusic:
> Is there an advantage of making this a vscode task over a CMake target

Cmake is unnecessarily hard.

> One thing that got me to avoid going for this was its runtime: 

I'm going to do another commit that labels all server tests so you can omit them. These take the most time.

It's very useful when you are working on the language.

JordanHendersonMusic:
> Cmake is unnecessarily hard.


Just to expand...

I could spend hours trying to get this to work in cmake, but il not many people will be able to maintain it. I think the build system is already very complex and we should be cautious of furthering this complexity.

dyfer:
IMO CMake is not that hard ;) and could be used by other tools than vscode. Our cmake files are still quite extensive and a bit messy, I agree. Still, I don't necessarily oppose this...

Finding server by sclang is somewhat of a separate issue... we do this differently on different platforms, according to platform specifics, but it's gear towards finding it after installation, it's doesn't take into account running scsynth/sclang from the build directories, though I thought that there was a note about this _somewhere_.... 

Improving and modularizing the test runner is an orthogonal task to this - currently we have two ways of running tests: the full test runner, and running them using `UnitTest.runAll`/`TestClass_testName.run`, after adding the testsuite to the sclang compiled paths. Maybe we need a birds-eye view approach to take into account both how to run tests and how to integrate them at the same time?

JordanHendersonMusic:
> Maybe we need a birds-eye view approach to take into account both how to run tests and how to integrate them at the same time?

Sure, but I just want to be able to run the test suite on the current build, I don't want anything to touch path and I don't want it to look in my system config or even user. I don't want to install anything. I just want to pass all information in with one config file. This pr sets it to the build_sclang.cfg, this is made by cmake and can pass all information needed this way. Imo, this is cleaner than actually doing the setup and running in cmake.

The issue I run into time and time again is adding dependencys to ctests, so you can ensure sclang is built before you run a test when calling it as a command. Build systems are literally just a graph,  this should be trivial.

JordanHendersonMusic:
... well this PR took a turn.


@dyfer I haven't tested it fully, but I think this works nicely! One issue is that we have been modifying the string to change to supernova.... this is horrid, but I've done my best to mitigate any issues.

CMakeLists.txt
```diff
@@ -581,24 +581,24 @@ endif()
...
+configure_file(languageConfigFiles/docRenderer.cfg.in ${CMAKE_CURRENT_BINARY_DIR}/languageConfigFiles/docRenderer.cfg)
 
 if(SC_DOC_RENDER)
 	file(GLOB_RECURSE SCDocSources RELATIVE HelpSource .*[^~])
 	file(GLOB_RECURSE SCDocClasses RELATIVE SCClassLibrary/SCDoc *.sc)
 
 	add_custom_target(doc ALL
 		COMMAND sclang
-			-l ${CMAKE_CURRENT_BINARY_DIR}/build_sclang.cfg
+			-l ${CMAKE_CURRENT_BINARY_DIR}languageConfigFiles/docRenderer.cfg
```
dyfer: isn't there a `/` missing before `languageConfigFiles`?

SCClassLibrary/Common/Control/Server.sc
```diff
@@ -1355,12 +1357,32 @@ Server {
...
+
+	*supernova { this.program = this.supernovaProgram }
+
+	*scsynthProgram {
+		var p = RuntimeOptions.scsynthPath;
+		var segments;
+		p !? {
+			segments = this.program.split($ );
+			segments[0] = p;
+			^segments.join($ );
```
dyfer: Is this to accommodate `exec` prefix or for something else?

lang/LangSource/PyrLexer.cpp
```diff
@@ -1782,6 +1782,7 @@ SCLANG_DLLEXPORT_C bool compileLibrary(bool standalone) {
         SC_LanguageConfig::readLibraryConfig(standalone);
     }
 
+
```
dyfer: whitespace

.vscode/tasks.json
```diff
@@ -10,7 +10,7 @@
       "label": "Reset CI results directory",
       "type": "shell",
       "command": "rm -r ci_results || true && mkdir ci_results",
-      "hide": true,
+      "hide": true
```
dyfer, line 13: is this change needed/desired?

dyfer:
> ... well this PR took a turn.

Haha 

> this is horrid, but I've done my best to mitigate any issues.

Yeah, I don't like it either, but we haven't come up with anything better up to this point...

dyfer (commented):
A few preliminary comments

SCClassLibrary/Common/Control/Server.sc
JordanHendersonMusic: I've rewritten this, but it is to work with.. `"PATH=$(dirname $(readlink /proc/$PPID/exe)):$PATH; exec scsynth"` in LinuxPlatform

.vscode/tasks.json
JordanHendersonMusic: its an autoformat change, best let vscode do what it wants!

SCClassLibrary/Common/Control/Server.sc
```diff
@@ -1355,14 +1357,40 @@ Server {
...
+		);
 	}
 
-	*supernova {
-		this.program = this.program.replace("scsynth", "supernova")
+	*supernovaProgram { |supernovaPath (RuntimeOptions.supernovaPath), oldProgram (this.program)|
+		^this.prReplaceProgram(
+			oldProgram,
+		 	if (supernovaPath === \undefined) { nil } { supernovaPath },
+		  	supernovaPath,
```
JordanHendersonMusic:
```suggestion
```

capital-G:
I found that the language config file has multiple issues (non-versioned, yaml is really bad for editing via unix tools,...) and I found it one of the most annoying things in SC and would like to remove it instead of embracing it further (ideally the magnet package file will make it obsolete nonetheless...)

IMO using program options (https://github.com/supercollider/supercollider/pull/6640) instead of language config is the preferred way since it easily allows to scope an environment: it is documented and can also be set via env variables, which is what almost all programming environments use (for a good reason IMO). In #6802 I also favored using env variables instead, which seem less intrusive to me.

> IMO CMake is not that hard ;)

+1 - it is definitely worth learning since it is the build tool we use and while I often think "it sucks", it is way better than any other build tool I worked with.
I also used vscode for a long time, but I found its capabilities really bad regarding building (e.g. use dynamic binary paths).
But cmake really excels at this - cmake knows where each binary is, and that running the tests depends on scsynth, supernova and sclang - which is really convenient way and is accessible then by any IDE.

JordanHendersonMusic:
> and I found it one of the most annoying things in SC and would like to remove it instead of embracing it further

Well there are two parts to the language config, there is the file, and the c++ struct/global. You will always need to pass information to runtime, so the struct must stay in some form or another. I can also add command line options if you like, but I think keeping the language config file is a good idea as this is quite useful for storing all the different configurations. 

>  I also favored using env variables instead, which seem less intrusive to me.

I'm dislike env vars because they are essentially just global state, this way, everything is explicit.

If some one wants to rewrite this in cmake, feel free to do so. whether the vscode stuff is merged or not doesn't really bother me, I'm going to be keeping this because it is exceptionally useful, just thought others might want it to.

> it is way better than any other build tool I worked with.

Zig's is the best I've ever used, because it is just zig code.

Also, I don't really think we should be using cmake to launch an executable, that's going to be the ide's job.

JordanHendersonMusic:
Merged into the other pr

Issue #5421, opened 2021-04-09
LinkClock tests sometimes fail


dyfer:
## Environment

* SuperCollider version: 3.11.2
* Operating system: macOS 10.14
* Other details (Qt version, audio driver, etc.):

## Steps to reproduce
Setup:
- add `testsuite/classlibrary` to your sclang path
- recompile library before running each test

Results:
- `test_newFromTempoClock_reschedulesOldClockQueue` - some iterations error out:
```supercollider
(
fork {
	~suite = TestLinkClock();
	20.do({
		~suite.test_newFromTempoClock_reschedulesOldClockQueue;
		0.1.wait;
	});
}
)
```
Result:
<details>

```supercollider
...
PASS: a TestLinkClock: new - starting a LinkClock with newFromTempoClock should reschedule stream players
Is:
	 a LinkClock 
Should be:
	 a LinkClock
ERROR: clock is not running.
ERROR: Primitive '_TempoClock_Sched' failed.
Failed.
RECEIVER:
Instance of TempoClock {    (0x1192c1b98, gc=7C, fmt=00, flg=00, set=03)
  instance variables [7]
    queue : instance of Array (0x10ffe2400, size=7, set=8)
    ptr : nil
    beatsPerBar : Float 4.000000   00000000 40100000
    barsPerBeat : Float 0.250000   00000000 3FD00000
    baseBarBeat : Float 0.000000   00000000 00000000
    baseBar : Float 0.000000   00000000 00000000
    permanent : false
}
PATH: /Volumes/data/Dokumenty/2020-2021/supercollider workspace/LinkClock failures.scd

PROTECTED CALL STACK:
	Meta_MethodError:new	0x115871f40
		arg this = PrimitiveFailedError
		arg what = Failed.
		arg receiver = a TempoClock
	Meta_PrimitiveFailedError:new	0x115878500
		arg this = PrimitiveFailedError
		arg receiver = a TempoClock
	Object:primitiveFailed	0x112f4eac0
		arg this = a TempoClock
	SimpleNumber:schedBundleArrayOnClock	0x116e49640
		arg this = 0.016
		arg clock = a TempoClock
		arg bundleArray = [ [ 15, 1016, gate, 0 ] ]
		arg lag = 0.0
		arg server = localhost
		arg latency = nil
		var sendBundle = a Function
	a FunctionDef	0x117315bc0
		sourceCode = "#{
					var tempo, server, eventTypes, parentType;

[...]

PR #7574, opened 2026-06-18
Rework Exceptions, UnitTest and CI refactor

### LINKED ISSUE #7577: sclang: try doesn't differentiate between throwing and error and returning one.
## Environment

* SuperCollider version: all

## Steps to reproduce

```supercollider
try { Error() } { 1.exit }
```

Didn't throw anything, this shouldn't terminate the program, but it does.

### LINKED ISSUE #7578: Threads should inherit their exceptionHandlers
## Motivation

```supercollider
try { ... } { |err| "I got an error".postln };
```
No matter what goes in the `...`, if an error occurs the handler should be called.

This is really important in the testsuite where we want to capture all exceptions thrown by a tests... if an exception leaves the bounds of the tests, it is an error. Inside the tests, if you are throwing, you do not want it escaping.

## Plan for Implementation

```cpp
void initPyrThread(VMGlobals* g, PyrThread* thread, PyrSlot* func, int stacksize, PyrInt32Array* rgenArray,
   ...
    if (g->thread) {
        thread->exceptionHandler = g->thread->exceptionHandler;
        gc->GCWrite(thread, &g->thread->exceptionHandler);
    }
```
```supercollider
Function {
	protect { arg handler, tryHandler;
		var result;
		result = this.prTry(tryHandler);
		if (result.isException and: {result.wasThrown}) {
			handler.value(result);
			result.throw;
		}{
			handler.value; // argument should be nil if there was no exception.
			^result
		};
	}

	try { |handler| ^this.prTry(handler) }

	prTry {  |handler|
		var result, thread = thisThread;
		var next = thread.exceptionHandler,
		wasInProtectedFunc = Exception.inProtectedFunction;
		thread.exceptionHandler = {|error|
			thread.exceptionHandler = next; // pop
			// If we are throwing from a child thread, then we run the handler and halt the child thread.
			if (thisThread !== thread) { 
				handler.value(error);
				^this.halt;
			} {
				^handler.value(error)
			}
		};
		Exception.inProtectedFunction = true;
		result = this.value;
		Exception.inProtectedFunction = wasInProtectedFunc;
		thread.exceptionHandler = next; // pop
		^result;
	}
}
```
```supercollider
Exception {
	var <>what, <>protectedBacktrace, <>path, <wasThrown=false;

	throw {
		wasThrown = true;
		^super.throw;
	}
}
```

### LINKED ISSUE #7608: Restarting Pprotect bricks sclang
## Environment

* SuperCollider version: 3.15
* Operating system:

## Steps to reproduce

```supercollider
p = Pprotect( Routine { 0.01.yield; Error().throw }, {"CAUGHT".postln})
p.play;
p.play;
```

https://sc.dennis-scheiba.com/c/6af7e99c4d76ba3ecaed009a89c4f2b3

No bricks.

JordanHendersonMusic:
## Purpose and Motivation

There is no way to catch all exceptions. This makes it impossible to avoid turning sclang into a brick when making a CLI tool. This PR fixes that by ensuring all (well sometimes all) routines pass their exceptions back to whomever called them.

Rework unit testing, adding features flags so you can filter them.

### Exceptions
Examples of new behaviour.

```supercollider
try {
   fork { Error().throw };
   Error().throw
} {  "Caught".postln };
// prints Caught twice
```

```supercollider
var r = try { Routine { \moo.yield; Error().throw } } { "Will not be called".postln };
try { r.next } { "not called".postln } // \moo;
try { r.next } { "is called".postln } // "is called"
```

```supercollider
p = Prout {
   \meow.yield;
   Error().throw
};

try { p.play } { "Catches the error".postln };
// yields \meow, then prints "Catches the error"
```

This makes `Pprotect` useless, which is good, because it is seriously broken #7608 and its implementation is very confusing.

Additionally this allows throwing arbitrary objects. 
```supercollider
try { 1.throw }  {  |er| er } === 1
```

There will undoubtedly be some breaking changes, particularly if users touched `Thread:exceptionHandler` (which they should **never** do because they will corrupt everything). Nevertheless, I think they will be small and only involve error paths, and the upside is that we can now writing code with the certainty that we can catch exceptions.

Pprotect should probably be removed in the future, and all the complexity in the schedulers where the clock/streams need to mark themselves as having streamErrors can also disappear with this.

### UnitTest

Flags have been added, letting you filter. Output has been expanded. Failures are added if a test throws an exception *outside* of its scope.

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation


testsuite/classlibrary/TestCoreUGens.sc
```diff
@@ -5,6 +5,7 @@ TestCoreUGens.run
 */
 TestCoreUGens : UnitTest {
 
+
```
JordanHendersonMusic:
```suggestion
```

capital-G (commented):
Why don't we use sub-classes for this? E.g. `ServerUnitTest`, `DocumentUnitTest`. Most testing frameworks I work with differentiate the context they need by different classes - this would also allow us to make the test context more consistent.

Having an implicit member variable `features` seem a bit easy to miss IMO?

JordanHendersonMusic:
> Why don't we use sub-classes for this?

because we don't have multiple inheritance so each test can only have one feature tagged.

> Having an implicit member variable features seem a bit easy to miss IMO?

Yes, but its the only way to allow multiple features that are defined in the class, this is better than adding it to a separate text file.

capital-G:
> because we don't have multiple inheritance so each test can only have one feature tagged.

But we could abstract the current approach of flags through classes? IMO this provides a better dev experience regarding autocomplete etc. Also, the tests you modified don't have multiple features - and unit test should be isolated, so it an edge case that they cover multiple areas.
And flags such as `isSlow`(?) could be `classvars` of `AbstractUnitTest` base class which can be filtered by a test runner.

JordanHendersonMusic:
>  IMO this provides a better dev experience regarding autocomplete etc. 

I just stuck some feature on that were useful for me right now. Being a ugen, iougen, and depending on abelton link are all independant (well iougen is a subset of ugen, but abelton link is independant). Using classes for this is going to be a complete mess.

> And flags such as isSlow(?) could be classvars of AbstractUnitTest base class which can be filtered by a test runner.

classvars in abstractunittest would be shared across everything that inherits from it. This wont' work. Also, we don't have an instance of the class when it comes to filtering, we only have the meta class.

dyfer:
Thanks for this.

I have a strong request here: let's consider the need to specify which tests to run in the CI, the way we currently do using https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/test_run_proto_gha.scd

BTW that file was based on the the [original implementation](https://github.com/supercollider/qpm/blob/main/qpm/scscripts/test_runner.scd) of the [test runner](https://github.com/supercollider/supercollider/blob/develop/testsuite/scripts/sclang_test_runner.scd) which I think was designed by @scztt and built into (now obsolete) [qpm](https://github.com/supercollider/qpm).

We have two ways of running the tests: UnitTest.runAll, and `sclang_test_runner.scd`. The latter has the ability choose which tests to run, log progress, measure test time, and it constantly updates a results file, so that results can be retrieved even if sclang tails completely. I think we should bring this functionality to UnitTest.runAll eventually (in a separate PR).

As for selecting which tests to run, I don't insist on following the way `test_run_proto_gha.scd` is structured right now, but it's one way to select which tests are running. I think that tests configurations should be stored in separate files, as opposed to the class library. I'd imagine we'd ship some "presets" (like no server tests) and also have a directory for users to specify their own sets if needed.

What do you think of storing test sets in some kind of configuration file(s)?

JordanHendersonMusic:
> What do you think of storing test sets in some kind of configuration file(s)?


> let's consider the need to specify which tests to run in the CI, the way we currently do using


So I think this is the wrong approach. If a family of tests (say ones that use the document class) don't work on the ci then we should filter out all tests that touch that class. If we keep a list of tests we just dump whatever doesn't work in there. This means we never really dig into why. Good example is linkclock, it's borked on my machine too! How long has this been the case?  The condvar/condition stuff should work, there is absolutely no reason why not.

We should only be removing tests that can't work on the ci because a classes doesn't exist there or for some platform specific reason. This can be described in a feature and then filtered out - document is one example.

When it comes to adding a new test, how many configuration files will that author have to edit? If users have their own, how will these be updated! This way, you just add the meta data and the test ends up in the correct list.

> I think we should bring this functionality to UnitTest.runAll eventually (in a separate PR).

I'd rather it be a different method, but bringing the functionality into the class library sounds good!

telephon:
Off my head, there are three different approaches: 

1. one method that returns symbols or classes as signal (the one suggested here)
2. several predicates implemented as methods (rturning a boolean)
3. an object that organises tests into groups (could be a class or just a dictionary in a file)

The variant (3) would be along the lines from what we have now, and it is more decoupled. The disadvantage is that you have to explicitly add tests it these collections.

The disadavntage of (1) is that it adds one class per feature. As it is right now, it is not yet vey "speaking", because "feature" can mean a lot of different things and the class names do not represent what they do. 

```supercollider
UnitTest.runAll(UnitTestFeatures.serverBoot);
```

Some variants of the interface, independently of implementation

```supercollider
// more speaking
UnitTest.runSelected([\needsServer, \needsNetwork]);

// more obvious
UnitTest.runList(
    UnitTest.select([\needsServer, \needsNetwork]);
);

// the most general:
UnitTest.runList(
    UnitTest.all.select { |test| test.needsServer or: { test.needsNetwork } }
);
```




Also maybe a good way forward would be to decide:  How many of these "features" do we currently need? Will this number grow quickly?

dyfer:
> If we keep a list of tests we just dump whatever doesn't work in there. This means we never really dig into why. 

Hey, no need to take swipes at broken tests :) We know the current state is not ideal (*). Many devs have spent time trying to fix as many as we could. It's just a time sink: not only one needs to figure out the exact logic and original author's intent, but also evaluate if the test is broken or if the functionality is broken (sometimes). Finally, some tests only/mostly fail when they are run in the (unoptimized for audio/desktop) CI environment and it's difficult to recreate the failures locally; or the failures are very indeterminate.

Also, nobody suggests adding new tests to the list. But I agree that the current approach mixes two issues currently: excluding tests of certain category that e.g. can't be run in a CLI context (i.e. Document tests) and tests that are broken one way or another.

---
(*) There was a period of time when tests were being added to the repository, but they weren't run in the CI. That left a lot of technical debt that we still haven't "paid back". I don't believe we merged any test that outright fail since the testsuite has been running in the CI (though a few of them have been randomly failing).

EDIT:

> Good example is linkclock, it's borked on my machine too! How long has this been the case? 

About 5 years since we started tracking it, so maybe a little more: https://github.com/supercollider/supercollider/issues/5421

JordanHendersonMusic:
> The variant (3) would be along the lines from what we have now, and it is more decoupled. The disadvantage is that you have to explicitly add tests it these collections.

I don't think we want to decouple tests and their features (or whatever we call them), because they are *their's*.

> The disadavntage of (1) is that it adds one class per feature.

I don't understand this comment? Tests can have many features, they return an array of them, and features are just symbols. The class methods just return symbols - it's basically an enum.

> As it is right now, it is not yet vey "speaking", because "feature" can mean a lot of different things and the class names do not represent what they do.

> // more obvious
UnitTest.runList(
    UnitTest.select([\needsServer, \needsNetwork]);
);

That's a great suggestion, I think this is perfect, but I'd use the enum class to make the symbols. 

The third also reads well, but makes the implementation a bit harder.

> Finally, some tests only/mostly fail when they are run in the (unoptimized for audio/desktop) CI environment and it's difficult to recreate the failures locally; or the failures are very indeterminate.

So very few of the server tests work on my machine (optimised) when run in the test runner, port in use errors, so it isn't just on the VM. My point is by having a do not run list, is that we might have prematurely added tests to this list.


There is probably some issue with sclang s threading that needs looking at...

> Hey, no need to take swipes at broken tests :) 

Just to be clear, I think the unittest class is really excellent. My issue it that we've prioritised getting a green tick on the ci, rather than displaying our errors, which makes it appear that we don't have any. This is obviously really useful for developing new features, but the errors should be slightly annoying otherwise they don't get fixed. Friction is sometimes useful. If we need to disable some specific known naughty tests when we make a release then that seems fine, but for the Dev branch, I think having these error be annoying is good. We could even have a separate ci step that doesn't report if a known flakey test has failed.

Edit: in theory we could get this flaky test list by polling GitHub issue with a certain tag. This is probably too hard to implement. I also don't think 'flakeyness' should be a feature. This is a good time for a separate list imo. So the difference between this and the current approach is that we rule out some test by category, these are not run in ci, then we have a list labeling tests as either flakey or expected to fail. Flakey test can fail without causes the ci to fail, if expected to fail tests pass, this causes the ci to fail.

JordanHendersonMusic:
> Also maybe a good way forward would be to decide: How many of these "features" do we currently need? Will this number grow quickly?

If they are just symbols, I don't think it's that much of an issue. As long as we don't go crazy. If this was done with inheritance these would grow at `(n^2 + n)/2` (if orthogonal). Adding features does mean going through all the tests and labelling them correctly, this took me about 5 mins, so it's not too long, but long enough to make you think twice.

telephon:
> That's a great suggestion, I think this is perfect, but I'd use the enum class to make the symbols.

Why don't you just make them methods of the test?

```
needsServer { ^true }
needsNetwork { ^true }
```

dyfer:
Thinking more about this, I like the idea of the features more and more, thanks Jordan for pushing it this way.

Another thing that could be done when tests are labelled with features, is that we could manually split the testsuite into multiple segments ran in parallel (well at least two, the ones requiring the server, and everything else).

Off the top of my head, in case it's useful for labelling:

The categories I can think of are:
- needsGUI
- needsWebEngine
- needsDocument (needsIDE? would we differentiate this for different editors potentially?)
- needsScsynth
- needsSupernova
  - currently we don't really run tests with supernova, except a few that explicitly run against both servers
  - switching between scsynth and supernova for all tests has not been attempted AFAIR - I guess it should work to do `Server.supernova; TestName.run` but I'm not sure
  - eventually, if resources are available, maybe we should probably run the full server-using portion of the test suite against both servers to ensure compatibility?
- needs...External? or a more granular division
  - I'm thinking of Serial, LinkClock, HID (not tested currently) etc

Finally, it may be useful to be able to combine features using both inclusion and exclusion? I.e. if one wants to run tests that need scsynth but don't need GUI (I think there are a few that need both).

EDIT: there's also a matter of tests for extensions... These are not currently ran at all (for the few tests that made their way to sc3-plugins), but it is something to maybe keep in mind when we refine the solution here. Once the package manager and sclang extensions come around, it would be good for the extensions to be able run their tests as well (whether in CI or locally), though it may be orthogonal to this discussion.

JordanHendersonMusic:
I'm going to push a new pr in a day or two. It moves the script into UnitTest.runAll and adds a printing class (basically produces the same output). I've also: added more try blocks and if the tests throw it adds a failure result, added a way to tag tests as having 'knownIssues', made a language config / flag to enable any uncaught exceptions to terminate the sclang instance (useful if the testsuites.itself fails, and useful in many cli contexts), given tests time limits so we don't hang if the server fails to boot, and probably a couple of things I can't remember right now. It seems to be working quite well at the moment.

@telephon yup that's much nicer. One thing will be presets, there will have to be some way to return a filter function. 

@dyfer lots of testsuites do the regex things, but I've never really understood it and I don't think we really used it? It requires the names across suites to follow some kind of convention.

> Finally, it may be useful to be able to combine features using both inclusion and exclusion? I.e. if one wants to run tests that need scsynth but don't need GUI (I think there are a few that need both).

Julian's third example I think is the most flexible and makes what you want something like...
```
UnitTest.runList(UnitTest.all.select{|t|
   t.hasSCSynth and: t.hasGUI.not
});
```

Or even shorter 
```
UnitTest.runFilter {|t|
   t.hasSCSynth and: t.hasGUI.not
}
```

One down side is that this isn't at the method level, but at the class, I don't really think this is an issue, but it's something to be aware of.

> currently we don't really run tests with supernova, except a few that explicitly run against both servers

Yes this is an issue, but I think we should actually remove the supernova tests. Instead, I think the setUp should be given a server options instance, then we can run all the server tests with a bunch of different options because there shouldn't be any difference... Well if there is, then the test can query and switch, but in most cases there shouldn't be. Implementing this is probably something for another time though.

JordanHendersonMusic:
Okay I think this is looking pretty good (check out the histogram). What do we think about this approach?

SCClassLibrary/Common/UnitTesting/UnitTest.sc
```diff
@@ -55,50 +83,147 @@ UnitTest {
...
+			} {
+				this.prRunTestMethodImpl(method, report);
+			};
+
+			this.prRunTestMethodHandleKnowIssue(method, report);
 			if(report) { this.class.report };
+
+			durations ?? { durations = () };
+			durations[this.class] ?? { durations[this.class] = () };
+			durations[this.class][method.name] = Date.localtime.rawSeconds - start.rawSeconds;
```
JordanHendersonMusic, line 165:
```suggestion
			durations[this.class][method.name] = Date.localtime.rawSeconds - start.rawSeconds;
			"% time taken = %". format(method.name, durations[this.class][method.name]).postln;
```

telephon:
I like it, this reads good! Is there anything that you were unsure about?

dyfer:
This is some great work, @JordanHendersonMusic ! It's a larger PR so I'd like to review in full, will take more time.

One housekeeping question/request:
Would you be able to rebase/squash commits so that 1) sclang changes (quit on throw, passing variables, server path handling) are in one or two commits, 2) vscode etc changes are in another commit and 3) unit test refactor is in yet another commit? Maybe even separate changes to the actual tests (like linkclock changes, moving the plot function) into a separate commit?

JordanHendersonMusic:
@dyfer thanks! I really want to make sure all the work you and other have done investigating and curating these tests is visible — this is why all the known issue have urls to issues! 

Yes that sounds like a good idea regarding commits! I think things are pretty much settled down to make this do-able.

There is this issue here: https://github.com/supercollider/supercollider/issues/7578 which makes all this code quite a bit harder. You see, right now it actually isn't possible to write code in such a way that all exceptions are caught, this is because threads don't inherit their exception handler. The issue for this PR and other people who want to write command line tools is you want all uncaught exceptions to kill the program (else it becomes a potato). One place this comes up is in Pprotect. Here, we want to throw and check the clean up happens, but one important strategy in ensuring the test don't hang for ever is to wrap all tests in try blocks and add the exceptions as failure. Now this happens on a different thread, so it isn't such an issue so we can't currently catch it, but at the same time we want all uncaught to end the program... I don't think we should let tests throw exceptions outside of their scope, and that means we need to wrap everything that can throw inside a try (which requires inheriting exceptions handlers).

I've noticed two other issues. We can't actually ensure all threads have finished once the test returns, and there seems to be a gc related issue regarding clocks on the cpp side of things where if you make a temporary clock and trigger the GC, then get the clock to do something (like link clock which responds to external stuff), the underlying memory for the sclang-side clock gets recycled, usually turning into a frame.

dyfer:
Thanks for all the work and explanation again, this is so great.

> I really want to make sure all the work you and other have done investigating and curating these tests is visible

Thanks, I appreciate it!

> There is this issue here: #7578 which makes all this code quite a bit harder. You see, right now it actually isn't possible to write code in such a way that all exceptions are caught, this is because threads don't inherit their exception handler. 

Being able to address this is awesome. I thought of this many times after realizing that there's no way to prevent sclang from just sitting there in CI.

> there seems to be a gc related issue regarding clocks on the cpp side of things where if you make a temporary clock and trigger the GC, then get the clock to do something (like link clock which responds to external stuff), the underlying memory for the sclang-side clock gets recycled, usually turning into a frame.

YES I've seen that in LinkClock failures. Such a mystery (esp when not understanding the inside of sclang) - THANK YOU for catching that!

dyfer:
Wow, an [ASCII HISTOGRAM](https://github.com/supercollider/supercollider/actions/runs/27972471669/job/82782904673?pr=7574#step:6:14839)!!! 🤩👏

testsuite/scripts/run_classlibrary_unittest.scd
```diff
@@ -0,0 +1,5 @@
+UnitTest.runAll { 
+    |durations, passes, failures, knownIssues, success|
```
dyfer: Our [style guide says](https://github.com/supercollider/supercollider/wiki/Style-Guidelines%3A-SuperCollider#rule-place-the-parameter-list-on-the-same-line-as-the-opening-curly-bracket-of-a-function-or-method) to put the arguments in the same line as the opening bracket.

Similarly in other places, please.

capital-G:
> Why don't we use sub-classes for this? E.g. `ServerUnitTest`, `DocumentUnitTest`. Most testing frameworks I work with differentiate the context they need by different classes - this would also allow us to make the test context more consistent.
> 
> Having an implicit member variable `features` seem a bit easy to miss IMO?

Sorry to be a party pooper, but I still stand with this and is based on my experience on writing tests.
I think implementing a classmethod screams for being missed (and is IMO not a good abstraction) and I am almost sure will create compatibility/legacy issues down the line b/c we have to touch every test instead of adjusting its base abstraction.
We are forced to make everything a class in sclang, so it feels really awkward for me to not make use of inheritance in that case like almost every other test framework is doing. And "merged" behavior is also an anti-pattern in regards to unit-tests, which should be as atomic as possible - though this could be done though if sclang would have something like mixins or decorators ;)

Also, while I appreciate the efforts around making the test results nicer, I think we should rely on outside tools which can help us here instead of re-inventing some wheels. I have somewhere an (unfinished?) fork where I was implementing [JUnit reporting](https://github.com/testmoapp/junitxml) which would be nice in combination w/ e.g. https://github.com/mikepenz/action-junit-report

I know that SCs implicit path assumptions create lots of trouble, but I **strongly** belief that we should not add stuff to the language config yaml! The YAML format is terrible in regards to automatic editing via unix tools and I know for sure that it will create conflicts (and confusion for the user) when multiple versions of scsynth/supernova are present on the system.
I am a big fan of env variables and they are used pretty much everywhere b/c they just behave like scopes in programming, which really helps a lot in automated environments. I can't think of an interpreter which doesn't use env variables to locate paths of other programs, so I am not sure why we should go a different path here. [The lookup of the sclang/scsynth/supernova paths within the package manager magnet relies exclusively on env variables]

Although I am no fan of the startup file, this could be one of the instances where it is actually useful - for CI we could simply resolve it via `.getenv` and users could have their local path set, but this would only affect advanced users explicitly and not everyone implicitly.

> Off my head, there are three different approaches:
> 
>   1. one method that returns symbols or classes as signal (the one suggested here)
>   2. several predicates implemented as methods (rturning a boolean)
>   3. an object that organises tests into groups (could be a class or just a dictionary in a file)

I'd vote for 3, which would allow for something like

```smalltalk
UnitTest.all.select({|test| test.name.contains("hello")}).run;
```

`UnitTest` could implement some default subsets of tests to run and additionally, each `UnitTest` could implement a `shouldRun` method (which has a default implementation of true), such that a unit test can decide by itself if it should be run in the current environment or not (e.g. we could disable tests in CI via `"SC_RUNNING_IN_CI".getenv.notNil`) - this couples the test execution condition to the code of the test.

> If we need to disable some specific known naughty tests when we make a release then that seems fine, but for the Dev branch, I think having these error be annoying is good. We could even have a separate ci step that doesn't report if a known flakey test has failed.

The CI is rather "binary" - either it is successful or not. Having a non-successful CI is problematic in multiple ways - e.g. introducing a new failed test will not get noticed, as well as other build issues since the CI is simply "red" and not being able to make a release w/ a red pipeline. Therefore I think it is fine to skip tests in CI. ofc it would be fine to have as few divergence as possible between full testsuite and CI testsuite, but I doubt that this will be realistic.

> Just to be clear, I think the unittest class is really excellent.

I think our testing framework could benefit from some features

* there is no way to mock/patch implementations and also no test fixtures generator - this leads to some awkward and unperformant test situations
* be able to re-run the changed tests w/o recompilation leads to a situation where writing tests is pretty tiresome and results in not the best test coverage
* No `ServerTest` class
* It would be also great to flag/wrap tests instead of doing this on a class level, but this would require something like decorators/macros or abstract object?

JordanHendersonMusic:
> I think implementing a classmethod screams for being missed

Good job it's trivial to change, unlike changing base class.

> automatic editing via unix tools 

What does this mean and why is it relevant? Ymal is pretty widespread, this is a part of sc and won't be going away because people probably depend on it. You can even edit it directly in SC so changes there break backwards compat in a major way.

> will create conflicts (and confusion for the user) when multiple versions of scsynth/supernova are present on the system.

How? This solves that issue because you just tell it the path explicitly.

Unused entries in the ymal file are ignored.

>adjusting its base abstraction.

Also a compatibility/legacy issue.

> Also, while I appreciate the efforts around making the test results nicer, I think we should rely on outside tools which can help us here instead of re-inventing some wheels.

We can just do both. This adds a oncomplete method to the unit tests, just send the information somewhere else.


> I am a big fan of env variables and they are used pretty much everywhere b/c they just behave like scopes in programming

Until you don't know who set it and where. Particularly problematic when you have lists. There is a compiler flag on CXX_flags on my system that I cannot find who sets it.... It drives me up the wall!

> The CI is rather "binary" - either it is successful or not... Therefore I think it is fine to skip tests in CI. ofc it would be fine to have as few divergence as possible between full testsuite and CI testsuite, but I doubt that this will be realistic.

This pr has moved one since then. It is good to run the tests, even if they fail, because this means we check they don't throw!!! This pr ignores the results of tests with issues.

capital-G:
> > I think implementing a classmethod screams for being missed
> 
> Good job it's trivial to change, unlike changing base class.

Well - then you have to go into every class instead of changing the base.

> > automatic editing via unix tools
> 
> What does this mean and why is it relevant? Ymal is pretty widespread, this is a part of sc and won't be going away because people probably depend on it. You can even edit it directly in SC so changes there break backwards compat in a major way.

YAML is indeed pretty widespread but no major tool has adapted it in the last 10 years b/c it is a god awful format (look at the [abysmal spec](https://github.com/yaml/yaml-spec/blob/main/spec/1.2/markdown/spec.md) and all the [security flaws](https://nvd.nist.gov/vuln/search#/nvd/home?keyword=yaml&resultType=records) that grow out of it^^). Basic libraries such as editing w/o rewriting the whole file (e.g. edit some tree but keep the comments) are often not existent.
It is really hard to edit yaml files w/ confidence w/ e.g. sed or some basic string manipulation.
This is the reason why I implementing passing env variables to the language config b/c otherwise it was awful to specify which class library should be used.

I also don't see why this would be a backwards breaking change at some point - IMO we should go for "isolated, deterministic and reproducible environments" which are described on a "project/folder" level. While we could keep the global yaml around for some time, we ideally replace it with something which describes the environment properly.
I tried to use YAML for that but I just gave up, it just consumed too much of my time and didn't want to start writing a yaml parser from scratch^^

> > will create conflicts (and confusion for the user) when multiple versions of scsynth/supernova are present on the system.
> 
> How? This solves that issue because you just tell it the path explicitly.

There will be for sure people who will put SC into a versioned folder will specify the absolute path to that. Once a new version replaces the old one, the path will not resolve and this will lead to problems.
The question is: What should be persistent across scsynth/sclang/scide version changes and what should be modified to account for the new version? This is IMO an unsolved problem of the language config format and is a reason why I'd vote to just abandon it instead of embracing it.

> Unused entries in the ymal file are ignored.

Yes, but a proper config would be versioned such that we can be sure what and what not should be present and also how to migrate to a version or fall back to some different parsing logic.

> > Also, while I appreciate the efforts around making the test results nicer, I think we should rely on outside tools which can help us here instead of re-inventing some wheels.
> 
> We can just do both. This adds a oncomplete method to the unit tests, just send the information somewhere else.

Cool - also no problems with that nice histogram [props!], just wanted to let you know of this stuff which we could use some other tools as well before you start implementing a GH reporting bot for the test results ;)

> > I am a big fan of env variables and they are used pretty much everywhere b/c they just behave like scopes in programming
> 
> Until you don't know who set it and where. Particularly problematic when you have lists. There is a compiler flag on CXX_flags on my system that I cannot find who sets it.... It drives me up the wall!

OT: Have you searched for the content of your CXX flags on your file system? ;) I am not a linux expert, but IIUC you should be able to determine your environment variables solely by going through your `~/.zshrc` [insert link to arch wiki^^]

JordanHendersonMusic:
> IMO we should go for "isolated, deterministic and reproducible environments" which are described on a

So just replace the yaml file with json? I don't understand your objection here. Yaml is crap, fine, but without specifying loads of command line arguments (which I've added too) you need some file. This is what we've got right now. Turning this into json is pretty trivial regardless of whether there are a couple more options.

> This is the reason why I implementing passing env variables to the language config b/c otherwise it was awful to specify which class library should be used.

I'd rather this be json. Just make the cpp command line option use a different parser when it has something with a json extension.

Also, I strongly think if you have different versions flying around the class library server and everything else should all be in the folder with the project. Other wise we are just recreating the worst bits of python.

This could be done with a single command line flag (without config file) and then sclang fills in the Lang config struct with all the relevant information from the cwd, erroring out if it can't find what it needs. Qt ide then just switches to work on projects.

> What should be persistent across scsynth/sclang/scide version changes and what

Don't use mismatching server and language versions. The only way this arises is in the build dir. With the project, there is no need for a language config file, everything is in the dir.

JordanHendersonMusic:
> Well - then you have to go into every class instead of changing the base.

Yes this is good. People should stop and think twice before adding tags they aren't going to use, it's a development burden. 

Also, I'd you want to grow for them, right now is trivial, with a inheritance approach, you have to search the hierarchy.

Again, what you are proposing is potentially a n^2 + n /2 approach if these are all orthogonal. Since the tests all overlap, very few are atomic, you will need a lot of these combinations. I know they should be atomic... But this is an unrealistic suggestion at this point.


Sorry, I still think this is a bad idea and will cause a mess because you are trying to model a simple set inclusion using inheritance, which famously doesn't model a subset relationship, but a super set. If we had multiple inheritance this could work - but I'd still be against it.

JordanHendersonMusic:
@dyfer all done!

JordanHendersonMusic:
There are two (maybe three) remaining issues:

> 	[✘]	TestQuark:test_runQuarkInstallHooks

This one fails on Linux only, don't know why.


Windows seems to not respet the exit code, seems something else is overriding it.

The layout of the histogram seems messed on GitHub (might just be mobile) I haven't observed this on pc so will double check.

dyfer:
> Windows seems to not respet the exit code,

I've seen that too in CI, but I don't know why that is. There seems to be some inconsistent behavior when running `0.exit` from the main thread as opposed to from a routine - the latter seems to hang, and eventually exit if a subsequent call to sclang is made (tested in the IDE):

```supercollider
(
{
	"Hello and bye".postln;
	1.wait;
	0.exit;
}.fork(clock: AppClock)
)
```

JordanHendersonMusic:
Csharp does something similar to this, but it tags methods not classes and using attributes.



```csharp
namespace MSTestNamespace
{
    [TestClass]
    public class UnitTest1
    {
        [TestMethod, Priority(1), TestCategory("CategoryA")]
        public void TestMethod1()
        {
        }
    }
}
```


Here's junit with tags
```java
@Test
@Tag("IntegrationTest")
public void testAddEmployeeUsingSimpelJdbcInsert() {
}

@Test
@Tag("UnitTest")
public void givenNumberOfEmployeeWhenCountEmployeeThenCountMatch() {
}
```

MATLAB does tags like this 
```matlab
methods (Test, TestTags = {'System','FeatureA'})
        function testE (testCase)
            % test code
        end
    end
```

Go does something similar with build tags
```go
//+build test_all func_test
```



So yes, I think this is fine. 

I could pretty easily add attributes to the language (but that's overkill for this).

JordanHendersonMusic:
Do that instead.
```supercollider
PTest : UnitTest {
    *flags { ^[UnitTestFlags.needsServerBoot] }
    *needsServerBoot{ ^flags.includes(UnitTestFlags.needsServerBoot) }
}
```

JordanHendersonMusic:
TODO:

https://www.boost.org/doc/libs/1_44_0/doc/html/boost/program_options/parse_environment_id900805.html 

just this not the file to pass new information

JordanHendersonMusic:
Okay so this has yet again grown. Turns out, it isn't possible to write code that doesn't break without reworking the exception handling mechanism. Take a look at the description, I've rewritten it. I've also tried to keep a clean history here so the commits should make sense.


There are two existing issues: one with a test, I'm pretty certain it just can't find the right path; and another where windows doesn't respect the return code (it succeeded, but reports a failure incorrectly).

SCClassLibrary/Common/Core/Thread.sc
```diff
@@ -18,15 +18,26 @@ Thread : Stream {
 	var <executingPath, <oldExecutingPath;
 	var rescheduledTime;
 
-	*new { arg func, stackSize = (512);
-		^super.new.init(func, stackSize)
+	// When **creating** a thread, the default is to inherit all exception handlers from the thread that created it.
+	// This **can** be overridden in the thread the **resumes** it depending on the way it is resumed, e.g., 
+	//		the scheduler behaves differently from other resumptions as we don't really want to consume that thread into the calling, but instead, just *awaken* it.
```
JordanHendersonMusic: note to self, these comments are outdated.

JordanHendersonMusic:
I also added a problem matcher to the vscode task. When I get back to the compiler update, which I will start after this, it can be updated to actually point to the `assert` statement.
errr

JordanHendersonMusic:
@dyfer I think this is about as good as it can get. Its bigger than before because sc exception mechanism was lacking the features necessary to write code that doesn't crash! I've had to disable windows CI step results because there is some serious bug in sclang where the return code isn't respected (#7614). There is also some bug with `Quark` and `LanguageConfig` that only happens on linux (#7610). 

I haven't done any documentation changes yet, but will do if this can be agreed.

Also, we should think about removing `Pprotect` in the future because it is now unnecessary.
```supercollider
protect { p.play } { "always gets called".postln };
try { p.play } { "gets called if thrown" };

// or if you need to pass on the pattern

Prout { |i| protect { p.embedInStream(i) } { "always gets called".postln } }.play
```

dyfer:
Thanks @JordanHendersonMusic, this is amazing work.

At the risk of asking too much, I'm thinking that the cleanest path forward is to split this PR into a few different PRs so that we can properly review these changes. Here's a suggestion:

A separate PR for error checking, including
- https://github.com/supercollider/supercollider/pull/7574/changes/c330f7f56b3b3621b653918b18dc4505bb353961
- https://github.com/supercollider/supercollider/pull/7574/changes/26e31153323953903316fddbf1ac4acb59d51fe4
- https://github.com/supercollider/supercollider/pull/7574/changes/1a301cc1b47bd00e0e20eb054c87360c359255e2
- https://github.com/supercollider/supercollider/pull/7574/changes/cf66e0d3242b2b64c58a57a7f03e7fa683bd7701
- https://github.com/supercollider/supercollider/pull/7574/changes/cf66e0d3242b2b64c58a57a7f03e7fa683bd7701

A separate PR for actual test changes, including
- https://github.com/supercollider/supercollider/pull/7574/changes/4f4330e79dffe10b7a197688225901ee5ce550fe
- https://github.com/supercollider/supercollider/pull/7574/changes/56174d5d9df25cd66ed37c2c25ccbdc0dcc71cfa
- https://github.com/supercollider/supercollider/pull/7574/changes/e44117829e5f0a78fdc07dac3adb0ba03a5ba218
- https://github.com/supercollider/supercollider/pull/7574/changes/b4b69605acb97747ba6af71a74712fb86320b2bb

And a separate PR for any remaining commits, like, path corrections, vscode etc.
Alternatively, these remaining commits could go into the second test changes PR. However, I'd definitely like to separate error checking to get more eyes on these changes. Would that be okay with you? it's a bit of cherry-picking, but I think it will make the review easier.

JordanHendersonMusic:
sure, `rebase -i` to the rescue!

JordanHendersonMusic:
Okay so doing the final pr for the actual unit tests is a bit hard because it depends on all the others. So I'll just leave it  until the others can be sorted

JordanHendersonMusic:
Another reason why the vscode tasks are great is when you are using worktrees and want to version the class library.

PR #7579, opened 2026-06-22
Fix boost include for macOS

### LINKED ISSUE #3099: scsynth: shared memory access prevented, crashes server.
starting with 40a65392eea6b652c0f798fee970783d6cb20b1b, I get a

```
Cannot connect to shared memory: Cannot connect to shared memory
ERROR: Primitive '_ServerShmInterface_connectSharedMem' failed.
Failed.
```

when booting any server e.g. via `s.boot`.

Since the hanging process reportedly chokes ScIDE if localhost is spawned, a good method is to either run a server from the commandline, or to do

```
g = Server(\ext, NetAddr("192.168.23.3", 57112); 
g.boot;
```

where you need to replace `"192.168.23.3"` with the IP your computer is known by on the net.

the previous commit bb844145697de2aa90aaf05d093fbddffd6630b1 works.

capital-G:
Including libsndfile poisoned the include directories on macOS such that the homebrew boost was included before the boost that supercollider specifies. This resulted in a (possible) boost-version mismatch between scsynth and sclang, which created runtime problems for the SHM, such that the scope did not work anymore. By adding SYSTEM to the target include, the manually specified boost via cmake will be considered over the implicit (yet still fallback) boost headers from the libsndfile include.

Quoting the CMAKE docs @ https://cmake.org/cmake/help/latest/command/target_include_directories.html

> If SYSTEM is specified, the compiler will be told the directories are meant as system include directories on some platforms. This may have effects such as suppressing warnings or skipping the contained headers in dependency calculations (see compiler documentation). Additionally, system include directories are searched after normal include directories regardless of the order specified.

It is really strange that the `SNDFILE_INCLUDE_DIR` on homebrew resolves to `/opt/homebrew/include`, potentially poisoning all our includes!
I am not proficient enough in linking etc to tell if this is expected behavior for include directories.

```shell
➜  Downloads ls -alh /opt/homebrew/include | grep libsndfile
lrwxr-xr-x@   1 scheiba  admin    46B Jul 20  2025 sndfile.h -> ../Cellar/libsndfile/1.2.2_1/include/sndfile.h
lrwxr-xr-x@   1 scheiba  admin    47B Jul 20  2025 sndfile.hh -> ../Cellar/libsndfile/1.2.2_1/include/sndfile.hh

➜  Downloads ls -alh /opt/homebrew/include | grep boost
lrwxr-xr-x@   1 scheiba  admin    38B Apr 10 12:27 boost -> ../Cellar/boost/1.90.0_1/include/boost

➜  Downloads ls -alh /opt/homebrew/include | grep \.h | wc -l
     335
```

I used claude to help me through this - took me this way around 2 hours instead of multiple days to figure that one out... dynamic linking - yay^^

Closes https://github.com/supercollider/supercollider/issues/3099

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation


dyfer:
> It is really strange that the `SNDFILE_INCLUDE_DIR` on homebrew resolves to `/opt/homebrew/include`, potentially poisoning all our includes!
> I am not proficient enough in linking etc to tell if this is expected behavior for include directories.

Based on quick search linking the general homebrew include is expected, as it prevents build configuration breaking (needing to be recreated) every time homebrew updates packages. Is that reasonable behavior? I'm not sure, re-running  cmake configure is not a big deal IMO. But since I couldn't find any advice to link to individual packages' include folders, let's keep it this way.

dyfer (approved):
Thanks! What a find, thanks for tracking this down. This works on my system and fixes SHM for the local build. 

I don't have experience with the `SYSTEM` flag, but based on the manual and LLM advice it seems fine.

I think this can be merged.

capital-G:
> Based on quick search linking the general homebrew include is expected, as it prevents build configuration breaking (needing to be recreated) every time homebrew updates packages. Is that reasonable behavior?

I'd expect that homebrew would also offer an isolated libsndfile folder which is version independent and then symlinked against the actual versioned libsndfile folder. But don't get me started on homebrew^^

PR #7581, opened 2026-06-23
Add search paths to FindSndfile.cmake

dyfer:
## Purpose and Motivation

It seems that libsndfile from homebrew was found on macOS on Apple Silicon only because `/opt/homebrew` is added to global search paths. This PR explicitly adds them to the search paths.

I'm not sure if we really need this, but it seems to be a cleaner (if not strictly necessary) approach.

I thought it is related to #7579, but it is not.

## Types of changes

- Maintenance...?

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/cmake_modules/FindSndfile.cmake
+++ b/cmake_modules/FindSndfile.cmake
@@ -20,11 +20,17 @@ elseif (SNDFILE_INCLUDE_DIR AND SNDFILE_LIBRARY)
 elseif (APPLE)
 
   find_path(SNDFILE_INCLUDE_DIR sndfile.h
-    HINTS /usr/local/opt/libsndfile/include
+    HINTS 
+      /opt/homebrew
+      /usr/local/opt/libsndfile
+    PATH_SUFFIXES include
   )
 
   find_library(SNDFILE_LIBRARY NAMES libsndfile.dylib libsndfile.a
-    HINTS /usr/local/opt/libsndfile/lib
+    HINTS 
+      /opt/homebrew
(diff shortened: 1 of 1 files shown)


cmake_modules/FindSndfile.cmake
```diff
@@ -20,11 +20,17 @@ elseif (SNDFILE_INCLUDE_DIR AND SNDFILE_LIBRARY)
 elseif (APPLE)
 
   find_path(SNDFILE_INCLUDE_DIR sndfile.h
-    HINTS /usr/local/opt/libsndfile/include
+    HINTS 
+      /opt/homebrew
+      /usr/local/opt/libsndfile
```
capital-G, line 25: wouldn't you want the include directory here? Or does this work recursively? I assumed the latter as well?

```suggestion
      /opt/homebrew/include
      /usr/local/opt/libsndfile/include
```

dyfer: PATH_SUFFIXES below adds that, IIUC. We just do it elsewhere like this, it seems to be idiomatic but I'm not sure if it makes much difference in our case either way. I was trying to make it look closer to the rest of the file.

dyfer: @capital-G do you think this needs any changes?

capital-G: kk - seems good in that case, though a bit confusing IMO :D

capital-G (approved):

PR #7582, opened 2026-06-23
Fix BufRd interpolation documentation

### LINKED ISSUE #7575: BufRd audio rate overshoots at interpolation 0 (BUT: no interpolation is in fact 1)
## Steps to reproduce

```supercollider
b = Buffer.alloc(s, 128, 1);
d = ({ 1.0.rand2 ! 8 } ! 16).flat;
b.sendCollection(d);
g = { BufRd.ar(1, b, LFSaw.ar(200).range(0, BufFrames.kr(b)), interpolation: 0) };
g.plot; 
```

Image

shouldn't overshoot.

capital-G:
Fallback interpolation is cubic interpolation in the BufRd implementation, not no interpolation. See https://github.com/supercollider/supercollider/issues/7575#issuecomment-4750843064

Closes #7575

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Classes/BufRd.schelp
+++ b/HelpSource/Classes/BufRd.schelp
@@ -39,7 +39,7 @@ argument::interpolation
 1 means no interpolation, 2 is linear, 4 is cubic interpolation. 
 (The numbers one, two, and emphasis::four:: correspond to the number of consecutive
 samples needed to compute each type of interpolation.)
-Any other values for this argument will silently default to no interpolation.
+Any other values for this argument will silently default to cubic interpolation.
 
 instancemethods::
 private:: init, argNamesInputsOffset, checkInputs


JordanHendersonMusic:
How about, 4 is the default is sclang, but 0 does actually means none?

capital-G:
> How about, 4 is the default is sclang, but 0 does actually means none?

We shouldn’t change default values or behavior if it isn't inherently flawed in order to not modify any existing sounds / composition.

JordanHendersonMusic:
Well if we wanted to add another option, right now, everything means cubic (bar a couple of options) - isn't that flawed?

I get 0 being cubic, but -42?

capital-G:
> Well if we wanted to add another option, right now, everything means cubic (bar a couple of options) - isn't that flawed?
> 
> I get 0 being cubic, but -42?

What would you suggest? Cubic interpolation is the most useful of all available. I still think we should not adjust the sonic outcome of others people code.

JordanHendersonMusic:
I'd suggest:

0: cubic 
1: opt1 
2: opt2 
3: opt3

Anything else: undefined - do not use (and document this).

This preserve the behaviour when users have specified one of the valid options, even keeping the default. It changes none of the real choices, and if users want to be silly and use the value 42 we don't care what happens (so long as it doesn't crash).

I think we should adopt this approach across all these types of options. Having a fallback is fine, explicitly telling users it is a valid option pins us in.

dyfer:
> I'd suggest:
> 
> 0: cubic 1: opt1 2: opt2 3: opt3

As @capital-G pointed out in the linked issue, the numbers in BufRd correspond to the number of samples needed for the interpolation, hence 1, 2, and 4.

We shouldn't change the current behavior, it may not be "logical" but it's not broken.

dyfer (approved):

JordanHendersonMusic:
@dyfer 

Ah okay, I don't have an opinion on *those* settings then. I just don't think we should *document* that options other than 0, 1, 2 & 4 do something specific. They should just be left as 'undefined' so we leave ourselves open to anything we might want in the future. Right now, we map 1 to linear, 2 to quad, and literally every other number to cubic.

dyfer:
IMO the current (in this PR) documentation is fine, but we could add while that the use of numbers other than 1, 2, and 4 defaults to cubic now, it may possibly change in the future - to discourage use of other values. But I'd be fine leaving this as is.

dyfer:
Also, the length of the discussion becomes disproportionate to the suggested change :)

telephon (approved):

telephon:
We should throw an error for values larger than 4 and smaller than zero maybe? All other values are already taken, right?

dyfer:
> All other values are already taken, right?

In terms of current implementation, both 0 and 3 behave the same as 4 (default).

This check would need to be in the processing loop in cpp, right? That's tricky, no?

JordanHendersonMusic:
> This check would need to be in the processing loop in cpp, right? That's tricky, no?

Yes this is not a good idea in audio code and why I suggested we document providing a value not supported as undefined behaviour (but we secretly just provide a default but dont document it).

This text guarantees that from now until the end of time we will not add a new value without a breaking change.

To be clear, I don't really think this is an issue for this.... Unless with @Spacechild1's resampling there is a need for different interpolation approaches... I'm trying to think beyond this to all cases where we switch on a value.

telephon:
> This text guarantees that from now until the end of time we will not add a new value without a breaking change.

Well, then let's change the wording now.

capital-G:
> Yes this is not a good idea in audio code and why I suggested we document providing a value not supported as undefined behaviour (but we secretly just provide a default but dont document it).
> 
> This text guarantees that from now until the end of time we will not add a new value without a breaking change.

I don't see any difference for the end user here - we can't just change behavior and say "you shouldn't have used that, that's UB" - thats Hyrums Law IIUC. That works in corporate environments w/ type checkers etc, but not in SC where we want to provide as few restrictions and as much continuity as possible to the user.

Changing existing simply to be avoided, be it documented or undocumented - but better have it documented, though fallback values can be expected to break over time, i.e. if we introduce a new value.
For both kinds we should have a good reason to change existing behavior - and we certainly did changes when we observed a faulty implementation.

JordanHendersonMusic:
> Well, then let's change the wording now.

I think we should have a scrub through all code and make this change all at once. Clearly saying it is ub to pass a value outside the range.

This way, if we need to add a new value to any of these we don't have to have that awkward discussion about who's code we are going to break because this would have been documented for several versions/years by that point.

Importantly, we wouldn't actually change any code now (so user code would behave the same) only documentation.

PR #7583, opened 2026-06-24
Remove legacy Cocoa GUI documentation (SCImage, SCImageFilter, etc.)

prko:
## Purpose and Motivation

This PR removes several obsolete .schelp files that document classes which no longer exist in SuperCollider.
These classes belonged to the old Cocoa GUI system and were removed when the Qt GUI was introduced.

The following documentation files refer to classes that are no longer part of the codebase:
- SCImage
- SCImageFilter
- SCImageKernel
- SCContainerView
- SCDragView
- SCEnvelopeEdit

Keeping these files may confuse new users, so they should be removed.

## Types of changes

- Documentation

## To-do list

- [x] Updated documentation


dyfer (approved):
I think it's fine to remove these, I don't believe this functionality will be brought back in any way.

Issue #7584, opened 2026-06-24
SuperCollider 3.15 meta issue


capital-G:
## Proposed Timeplan

Code-freeze: End of July 2027

Release: September 2027

## Project

The release is tracked mostly via https://github.com/orgs/supercollider/projects/6

This meta-issues help to keep track of some things which are linked to other resources and/or require a certain order.

## Considering before 3.15

* [ ] https://github.com/supercollider/sc3-plugins/pull/435 - this is a regression of 3.13 - should we make a sc3-plugins 3.14.2 release w/ old API w/ this fix or should this be 3.15 only fix?
* [ ] https://github.com/supercollider/supercollider/issues/6102 - plugin interface changes
* [ ] help browser issues indicated by @prko:
  * [x] https://github.com/supercollider/supercollider/issues/6895
  * [ ] https://github.com/supercollider/supercollider/pull/7546#issuecomment-4680304521
  * [ ] https://github.com/supercollider/supercollider/pull/7542
  * [x] https://github.com/supercollider/supercollider/issues/7380

## Stuff to do before release

* [ ] Link FFTW statically in sc3-plugins (see https://github.com/supercollider/supercollider/issues/7543)
* [ ] Fix release script

prko:
Sorry to rewrite this here again. **I am posting this because I am concerned that the following issues may remain unresolved, and I believe they significantly affect beginners as well as users who rely heavily on documentation in their workflows**:

**1. HelpBrowser shows a blank page**  
The HelpBrowser intermittently renders a blank page when its CodeMirror content is evaluated or even scrolled, which makes the documentation difficult to use.
This issue was introduced in **3.14**. See **#6895**.

**2. CodeMirror evaluation interference**  
Code evaluation inside the HelpBrowser’s CodeMirror container is affected by the document currently opened in the main Code Editor.  
When the cursor in the main Code Editor is placed inside code, that code is evaluated instead — on all platforms.  
These two contexts should remain isolated.  
This issue appears to have been introduced by the recent **JavaScript‑based evaluation changes** made to resolve the evaluation problems in 3.14.  
See: [https://github.com/supercollider/supercollider/pull/7546#issuecomment-4680304521](https://github.com/supercollider/supercollider/pull/7546#issuecomment-4680304521)

**3. Keyboard shortcuts in HelpBrowser**  
Several keyboard shortcuts do not work inside the HelpBrowser, or conflict with global editor shortcuts.  
[...]

PR #7593, opened 2026-06-26
qt6: update type to typeId

dyfer:
## Purpose and Motivation

I'm going through our codebase to remove qt deprecations. The changes here are the most "noisy" so I'd like them to be reviewed separately.

I used a LLM to find these fixes, but I didn't put in any of the code optimizations it suggested, just replaced the qt objects.

I personally would like to still keep supporting Qt 5.15, hence all the ifdefs. But if other devs really would like to drop it, we could discuss that as well.

## Types of changes

- Maintanance

## To-do list

- [ ] Code is tested - I'm not sure how to test these changes...
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/QtCollider/QObjectProxy.cpp
+++ b/QtCollider/QObjectProxy.cpp
@@ -173,7 +173,11 @@ bool QObjectProxy::invokeMethod(const char* method, PyrSlot* retSlot, PyrSlot* a
     // construct the return data object
     QGenericReturnArgument rarg;
     const char* rtype_name = mm.typeName();
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     int rtype_id = QMetaType::type(rtype_name);
+#else
+    int rtype_id = QMetaType::fromName(rtype_name).id();
+#endif
 
     MetaValue returnVal;
 
--- a/QtCollider/QcSignalSpy.h
+++ b/QtCollider/QcSignalSpy.h
@@ -67,7 +67,11 @@ public:
         QList<QByteArray> params = mm.parameterTypes();
 
         for (int i = 0; i < params.count(); ++i) {
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
             int type = QMetaType::type(params.at(i).constData());
+#else
+            int type = QMetaType::fromName(params.at(i)).id();
+#endif
             if (type == QMetaType::Void)
                 qcErrorMsg(QString("QObject:connect: Don't know how to handle '%1', "
                                    "use qRegisterMetaType to register it.")
--- a/QtCollider/widgets/QcGraph.cpp
+++ b/QtCollider/widgets/QcGraph.cpp
@@ -181,7 +181,11 @@ void QcGraph::setCurves(const QVariantList& curves) {
         const QVariant& data = curves[i];
         QcGraphElement::CurveType type;
         double curvature;
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
         if (data.type() == QVariant::Int) {
+#else
+        if (data.typeId() == QMetaType::Int) {
+#endif
             type = (QcGraphElement::CurveType)data.toInt();
             curvature = 0.0;
         } else {
--- a/editors/sc-ide/core/settings/serialization.cpp
+++ b/editors/sc-ide/core/settings/serialization.cpp
@@ -214,19 +214,24 @@ static void writeTextFormat(const QTextCharFormat& fm, YAML::Emitter& out) {
 }
 
 static void writeValue(const QVariant& var, YAML::Emitter& out) {
-    switch (var.type()) {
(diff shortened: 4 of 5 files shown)


JordanHendersonMusic:
My question here is whether supporting qt6 *or* suppressing these warnings is worth the messy macro ifs.

We've already seen there are a bunch of issues in the help browser and stuff with upgrading to Qt 6 (highlighting not working). There is no guarantee that fixing these in 6 won't cause a regression in 5. That's why I'd rather stop supporting 5.

The other concern I have is if we stop supporting 5, how will we know to delete these checks?

One thing we can add is this...
```
static_assert(
   SC_Version < Semantic version{3, 16, 0},
   "Do we still support qt 5? if so, update the version number, if not, do a search form macros remove them."
);
```

This way, the macro can't be left around if we stop supporting 5.

dyfer:
> My question here is whether supporting qt6 _or_ suppressing these warnings is worth the messy macro ifs.

One reason to maintain Qt5 compatibility is to be able to create a 32-bit. Once we remove Qt5 support, we won't be able to build 32-bit SC with the IDE (but we can still build a headless version, I think). Qt 6 also drops support for older OSs, but there's no practical difference for us at this point, given other updates (unless someone is specifically building for an older system on Windows or Linux).

> We've already seen there are a bunch of issues in the help browser and stuff with upgrading to Qt 6 (highlighting not working). 

Is there an issue filed for this?

> There is no guarantee that fixing these in 6 won't cause a regression in 5. That's why I'd rather stop supporting 5.

Totally valid.
We state that there's "[limited support](https://github.com/supercollider/supercollider#platform-support)" for 5, so personally I wouldn't mind if things like cosmetic stuff slightly break, while the build still generally works. But this might not be an optimal approach, idk.

> The other concern I have is if we stop supporting 5, how will we know to delete these checks?

IMO it's easy enough to do a global search for `QT_VERSION_CHECK` and remove them by hand. Before this PR we have about 45 of those in the codebase, with this PR and other qt cleanup it would probably get to around 55. They should be removed in a single commit and I'd be happy to do it, if we decide to remove qt5 support altogether.

JordanHendersonMusic:
> One reason to maintain Qt5 compatibility is to be able to create a 32-bit. 

What systems needs 32bit and Qt? Wasm is the only 32bit system I'm aware of.

dyfer:
> What systems needs 32bit and Qt?

AFAIU old Raspberry Pis (1, 2, Zero, and Zero W). We also build 32bit on Windows, so if someone needs it, they can build it... but it's unlikely. 

EDIT: the original Bela platform seems to be 32-bit as well, but I'm not sure if we support that or the newer 64-bit platform; also I haven't checked if it uses the IDE or just the CLI builds.

JordanHendersonMusic:
I thought the pis couldn't use qt either?

dyfer:
> I thought the pis couldn't use qt either?

RPi runs a full fledged OS with a GUI, so I don't see why not. But I haven't done it myself, and arguably these old Pis are slow and impractical, esp for desktop use.

JordanHendersonMusic:
I just checked the readme instructions and the Bela ones have disabled qt.

> the project can only compile the QtWebEngine library under a 64-bit OS, which means that the help browser in the IDE, and the HelpBrowser and WebView classes in sclang, are unavailable for a 32-bit OS.

Ah I misremembered, it is just a few parts.

JordanHendersonMusic (approved):

dyfer:
Sorry, I should've pointed the qt5 discussion to https://github.com/supercollider/supercollider/issues/7006

prko:
@dyfer 
>>We've already seen there are a bunch of issues in the help browser and stuff with upgrading to Qt 6 (highlighting not working).
>>
>Is there an issue filed for this?

- https://github.com/supercollider/supercollider/issues/7584#issuecomment-4815149059
- https://github.com/supercollider/supercollider/issues/7584#issuecomment-4819092755
- https://github.com/supercollider/supercollider/issues/7584#issuecomment-4819739817

---- 

added:

>I just checked the readme instructions and the Bela ones have disabled qt.
>
>the project can only compile the QtWebEngine library under a 64-bit OS, which means that the help browser in the IDE, and the HelpBrowser and WebView classes in sclang, are unavailable for a 32-bit OS.

I think Bela and Raspberry Pi users generally don’t depend on SC-IDE.
Raspberry Pi is often run headlessly, making SC-IDE unnecessary, and even when using the desktop locally on the machine or remotely via VNC, the environment isn’t ideal for frequent use of the HelpBrowser.
Bela is controlled through a web-based interface instead.

In my own workflow with both devices, my MacBook Pro served as the main machine. I accessed the other machines remotely, used SC-IDE only on my laptop, and pasted code into the remote systems as needed.

dyfer:
Since we're not planning to remove support for qt5 in 3.15, I'll merge this as is.

PR #7595, opened 2026-06-27
Topic/add gersemi cmake formatter

JordanHendersonMusic:
## Purpose and Motivation

Auto format all the cmake.

before
```cmake
if(UNIX AND NOT APPLE)
    install(DIRECTORY include/common include/plugin_interface include/server include/lang
        DESTINATION ${CMAKE_INSTALL_PREFIX}/include/SuperCollider
	FILES_MATCHING PATTERN "*.h" PATTERN "*.hpp"
    )
    install(FILES SCVersion.txt
        DESTINATION ${CMAKE_INSTALL_PREFIX}/include/SuperCollider
    )
endif()
```
after
```cmake
if(UNIX AND NOT APPLE)
    install(
        DIRECTORY
            include/common
            include/plugin_interface
            include/server
            include/lang
        DESTINATION ${CMAKE_INSTALL_PREFIX}/include/SuperCollider
        FILES_MATCHING
        PATTERN "*.h"
        PATTERN "*.hpp"
    )
    install(
        FILES SCVersion.txt
        DESTINATION ${CMAKE_INSTALL_PREFIX}/include/SuperCollider
    )
endif()
``` 

TODO: add to ignored if we like.


editors/sc-ide/CMakeLists.txt
```diff
@@ -156,16 +169,13 @@ set ( ide_src
     widgets/util/volume_widget.cpp
     widgets/util/status_box.cpp
     widgets/style/style.cpp
-
     ${CMAKE_SOURCE_DIR}/common/SC_TextUtils.cpp
     ${CMAKE_SOURCE_DIR}/common/SC_Filesystem_macos.cpp
     ${CMAKE_SOURCE_DIR}/common/SC_Filesystem_win.cpp
     ${CMAKE_SOURCE_DIR}/common/SC_Filesystem_unix.cpp
     ${CMAKE_SOURCE_DIR}/common/SC_Filesystem_iphone.cpp
-
```
dyfer, line 165: Do we want empty lines here to be removed? They do separate logical groups of files. It's probably fine, but I wanted to point it out.

JordanHendersonMusic: I think comments would achieve the same goal. I haven't done any manual editing, this is literally the default output - I think it looks great!

dyfer: I agree it does look great!

capital-G:
If we use this, we should also include this file https://github.com/supercollider/supercollider/blob/develop/external_libraries/CMakeLists.txt

dyfer:
> If we use this, we should also include this file https://github.com/supercollider/supercollider/blob/develop/external_libraries/CMakeLists.txt

and also SCVersion.txt

JordanHendersonMusic:
I was a bit unsure about the external cmake one because I couldn't figure out a quick way to write the exclude filter in precommit

dyfer:
Another detail: we may want to separate uppercase/lowercase changes into a commit separate from the whitespace change (and add both to [.git-blame-ignore-revs](https://github.com/supercollider/supercollider/blob/develop/.git-blame-ignore-revs)) , in order to make sure that git indeed properly maps the changes to previous commits.

dyfer:
Do we want to proceed with this? I'm in favor, but if we decide for a smaller change using #7594, that would be fine too.

I'm slightly concerned that editorconfig is orthogonal to clang-format and this cmake formatter, though if editorconfig will consistently be a subset of the rules imposed by the other formatters, then maybe that's okay.

PR #7598, opened 2026-06-28
Update qt6 deprecations

dyfer:
## Purpose and Motivation

When building against qt6 (tested with 6.11), we get many notifications of deprecated functions. This PR removes all of them as far as I can tell.

It seems there are other qt-related warnings are about [overridden functions](https://github.com/supercollider/supercollider/actions/runs/28212234133/job/83576002289#step:16:10040), but that's for a separate PR.

The fixes were suggested by an LLM, but they seem straightforward and I haven't observed any issues.

## Types of changes

- Maintenance
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/QtCollider/widgets/image_painter.h
+++ b/QtCollider/widgets/image_painter.h
@@ -123,7 +123,7 @@ struct ImagePainter {
         case AlignVCenter:
             rect.moveTop(targetRect.top() + targetRect.height() / 2 - rect.height() / 2);
             break;
-        case AlignRight:
+        case AlignBottom:
             rect.moveBottom(targetRect.bottom());
             break;
         default:
--- a/editors/sc-ide/widgets/multi_editor.cpp
+++ b/editors/sc-ide/widgets/multi_editor.cpp
@@ -222,7 +222,13 @@ void EditorTabBar::mouseDoubleClickEvent(QMouseEvent* event) {
 }
 
 void EditorTabBar::showContextMenu(QMouseEvent* event) {
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     mTabUnderCursor = tabAt(event->pos());
+    QPoint globalPos = event->screenPos().toPoint();
+#else
+    mTabUnderCursor = tabAt(event->position().toPoint());
+    QPoint globalPos = event->globalPosition().toPoint();
+#endif
 
     QMenu* menu = new QMenu(this);
     // Cannot have a close tab action if we are not over a tab
@@ -234,7 +240,7 @@ void EditorTabBar::showContextMenu(QMouseEvent* event) {
         menu->addAction(tr("Close Tabs to the Right"), this, SLOT(onCloseTabsToTheRight()));
     }
 
-    menu->popup(event->screenPos().toPoint());
+    menu->popup(globalPos);
 }
 
 void EditorTabBar::onCloseTab() {
--- a/QtCollider/QWidgetProxy.cpp
+++ b/QtCollider/QWidgetProxy.cpp
@@ -478,7 +478,11 @@ bool QWidgetProxy::interpretDragEvent(QObject* o, QEvent* e, QList<QVariant>& ar
         if (!internal)
             interpretMimeData(data, args);
     } else {
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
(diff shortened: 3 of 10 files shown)


QtCollider/widgets/image_painter.h
```diff
@@ -123,7 +123,7 @@ struct ImagePainter {
         case AlignVCenter:
             rect.moveTop(targetRect.top() + targetRect.height() / 2 - rect.height() / 2);
             break;
-        case AlignRight:
+        case AlignBottom:
```
dyfer, line 126: This seems to have been a bug, but I don't know how it transpired...?

JordanHendersonMusic: Do they have the same numeric value?

editors/sc-ide/widgets/multi_editor.cpp
```diff
@@ -233,8 +237,11 @@ void EditorTabBar::showContextMenu(QMouseEvent* event) {
         menu->addAction(tr("Close Other Tabs"), this, SLOT(onCloseOtherTabs()));
         menu->addAction(tr("Close Tabs to the Right"), this, SLOT(onCloseTabsToTheRight()));
     }
-
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     menu->popup(event->screenPos().toPoint());
+#else
+    menu->exec(event->globalPosition().toPoint());
+#endif
```
JordanHendersonMusic: This one seems questionable, are we sure global positioning is the same as screen space in all cases? I'm not entirely sure what these are.

QtCollider/widgets/image_painter.h
dyfer: They do not. However, this is in the vertical alignment switch, so AlignRight is not valid. Here's the compilation warning: https://github.com/supercollider/supercollider/actions/runs/28331461949/job/83930301141#step:16:11473

editors/sc-ide/widgets/multi_editor.cpp
dyfer: I'll check, thanks.

Also, I didn't realize that I changed popup into exec, I need to check what that implies.

dyfer: The [docs](https://doc.qt.io/qt-6/qmouseevent-obsolete.html#screenPos) explicitly say to replace screenPos with globalPosition, so I think that replacement is fine.

QtCollider/widgets/image_painter.h
JordanHendersonMusic: Cool let's see what this subtly changes!

JordanHendersonMusic (approved):

QtCollider/widgets/image_painter.h
JordanHendersonMusic, line 40: @dyfer okay they aren't what I expected... They do have the same value! That means they will behave exactly the same. It thought these were Qt enums.

JordanHendersonMusic (changes requested):

JordanHendersonMusic (approved):

editors/sc-ide/widgets/multi_editor.cpp
dyfer: I updated this so that there's only a single ifdef switch and also we keep `menu->popup()`.

LLM also suggested that we are unnecessarily accumulating popup menus with each click, until the tab is destroyed... but that would be something to address in a separate PR, if that indeed is an issue (I only have vague understanding).

QtCollider/widgets/image_painter.h
dyfer: Ah, got it! I was looking at https://doc.qt.io/qt-6/qt.html#AlignmentFlag-enum, my bad.

dyfer:
So... OK to merge? I'm not sure if the conversations can be considered resolved or not.

QtCollider/widgets/image_painter.h
JordanHendersonMusic: Yes me to!

dyfer:
@JordanHendersonMusic thanks for merging.
Even thought the individual commits were small, they were separated on purpose so that individual changes can be easily inspected. Do we always want to squash in such cases?

Issue #4283, opened 2019-01-31
Quarks.install("wrong.url") behaves incorrectly


dyfer:
## Environment

* SuperCollider version: 3.9.3, 3.10.1
* Operating system: macOS 10.13
* Other details (Qt version, audio driver, etc.): git installed with xcode command-line tools

## Steps to reproduce

```supercollider
Quarks.install("http://wrong.name");
// output:
// Installing wrong
// wrong installed
// -> Quark: wrong[]

// after recompiling library
// WARNING: Could not open directory: '/Users/<username>/Library/Application Support/SuperCollider/downloaded-quarks/wrong'
```

## Expected vs. actual behavior

Expected behavior: 
If the path is incorrect, the installation should fail with an error and additional compile paths should not be added. It would be useful is there is a clear messages that the URL is wrong.

Currently, the installation fails silently, and an incorrect path is added to the compile directory list.

gusano:
@dyfer Thanks for spotting!
The issue here is that [there is no way to know what `Pipe.callSync(...)` exit code is](https://github.com/supercollider/supercollider/blob/develop/SCClassLibrary/Common/Files/File.sc#L217-L230), therefore cloning a valid or an invalid URL will yeld the same...
I'll have a closer look.

gusano:
@dyfer I don't have much time and I don't think there's a *good* way to solve this (if any) :|

jamshark70:
> The issue here is that there is no way to know what Pipe.callSync(...) exit code is,

That's actually not true.

https://github.com/supercollider/supercollider/pull/4115

But this PR was rejected because I should have proposed to deprecate callSync and replace it with a method that handles errors, but instead I had asked about it on the mailing list, gotten only one bit of advice (to keep the existing behavior and just add an error code) and then failed review... so the issue is still unfixed.

I dunno, maybe I'll take it up again. Or maybe someone else will. Anyway the PR shows how to get the error code.

PR #7603, opened 2026-06-29
classlib: add "quiet" flag to git commands

### LINKED ISSUE #7602: Git status updates are prepended with `ERROR: `
## Environment

* SuperCollider version: develop
* Operating system: macOS

## Steps to reproduce

```supercollider

Quarks.install("JoshMisc"); // or any other quark that isn't already downloaded

```

Actual behavior:
```
Installing JoshMisc
ERROR: Cloning into '/Users/marcin/Library/Application Support/SuperCollider/downloaded-quarks/JoshMisc'...
Adding path: /Users/marcin/Library/Application Support/SuperCollider/downloaded-quarks/JoshMisc
JoshMisc installed
```
Expected:
Either display `Cloning into '/Users/marcin/Library/Application Support/SuperCollider/downloaded-quarks/JoshMisc'...` on its own without the "ERROR: " message, or don't display this line at all, which is the behavior in SC 3.13.

I believe this is the result of a two factors:
- we now print stderr messages from sclang in the IDE (https://github.com/supercollider/supercollider/pull/6971)
- git outputting its status messages through stderr

Possible solution:
- add `-q` flag to make git commands quiet (aside from actual errors) - this would match behavior in older SC, before we were printing stderr
- add control to unixCmd, Pipe etc whether stderr is forwarded as an error or regular output (??? I'm not sure this is feasible)
- ...?

dyfer:
## Purpose and Motivation

This adds `-q` (`quiet`) flag to the git commands. It should prevent from status updates being reported by git to stderr (and thus being displayed with the `ERROR: ` note in the IDE.

IMO displaying status messages would be nice, but we don't have an easy way to differentiate between status update and an error, as `git` outputs status updates to stderr. In this PR I _believe_ actual errors would still be reported/posted (even if we [can't catch them](https://github.com/supercollider/supercollider/issues/4283)...).

Fixes #7602, if only in appearance.

## Types of changes

- Bug fix (more like a workaround...)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Quarks/Git.sc
+++ b/SCClassLibrary/Common/Quarks/Git.sc
@@ -11,20 +11,21 @@ Git {
 	clone { |url|
 		this.git([
 			"clone",
+			"--quiet", // suppresses progress report on stderr
 			url,
 			thisProcess.platform.formatPathForCmdLine(localPath)
 		], false);
 		this.url = url;
 	}
 	pull {
-		this.git(["pull", "origin", this.defaultBranchName ? ""])
+		this.git(["pull", "--quiet", "origin", this.defaultBranchName ? ""])
 	}
 	checkout { |refspec|
-		this.git(["checkout", refspec])
+		this.git(["checkout", "--quiet", refspec])
 	}
 	fetch {
 		tags = remoteLatest = nil;
-		this.git(["fetch"]);
+		this.git(["fetch", "--quiet"]);
 	}
 	isDirty {
 		^this.git(["--no-pager diff HEAD --"]).size != 0


SCClassLibrary/Common/Quarks/Git.sc
```diff
@@ -10,21 +10,21 @@ Git {
 	}
 	clone { |url|
 		this.git([
-			"clone",
+			"clone -q",
```
capital-G: shouldn't these flags be separated, i.e.
```suggestion
			"clone",
			"--quiet", // suppresses progress report on stderr
```

Otherwise this looks good to me

dyfer:
All right, I tested installing quarks on macOS and Windows, seems to work fine!

capital-G (approved):

PR #7616, opened 2026-07-04
classlibrary: ensure forkIfNeeded respects the requested clock

JordanHendersonMusic:
## Purpose and Motivation

forkIfNeeded should fork if the clocks don't match.

```
fork {
   {
        ... I need ClockA,  previously this would not work.
   }.forkIfNeeded(ClockA)
}
```

I haven't updated any docs, because I assumed this is a bug.

A part of https://github.com/supercollider/supercollider/pull/7574

## Types of changes

- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation

## Release Note (added after merge)

when using `forkIfNeeded` passing a clock as argument that is not the same as the current thread's clock, we need to fork. This was previously not done, now it is. This may subtly change some behaviour.

### DIFF
--- a/SCClassLibrary/Common/Core/Function.sc
+++ b/SCClassLibrary/Common/Core/Function.sc
@@ -158,7 +158,7 @@ Function : AbstractFunction {
 		// We should fork if the clocks don't match.
 		// We should also fork if the quants/stackSize don't match, but that doesn't seem possible.
 		if(thisThread.isKindOf(Routine) 
-			and: { clock !? { clock == thisThread.clock} ?? {true} },
+			and: { clock.isNil or: { clock == thisThread.clock }},
 			this, 
 			{ ^this.fork(clock, quant, stackSize) }
 		);
--- a/SCClassLibrary/Common/Core/Function.sc
+++ b/SCClassLibrary/Common/Core/Function.sc
@@ -155,7 +155,13 @@ Function : AbstractFunction {
 	}
 
 	forkIfNeeded { arg clock, quant, stackSize;
-		if(thisThread.isKindOf(Routine), this, { ^this.fork(clock, quant, stackSize) });
+		// We should fork if the clocks don't match.
+		// We should also fork if the quants/stackSize don't match, but that doesn't seem possible.
+		if(thisThread.isKindOf(Routine) 
+			and: { clock !? { clock == thisThread.clock} ?? {true} },
+			this, 
+			{ ^this.fork(clock, quant, stackSize) }
+		);
 		^thisThread;
 	}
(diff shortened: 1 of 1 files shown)


dyfer:
May I request @jamshark70 and @telephon to chime in here? This seems like a logical change to me, but I wonder if there are any interactions with Patterns etc that I'm not anticipating...

jamshark70:
I think, go ahead. If `forkIfNeeded` isn't going to respect the clock argument, then it shouldn't have a clock argument (but there's a good reason to have it, so it should be respected).

I can't think offhand of a case where a pattern genuinely forks. Ppar, Pspawner etc go to extraordinary lengths to avoid forking! I.e., the pattern contract is that the stream's behavior should be self-contained within its own thread; `fork` would be discouraged.

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -155,7 +155,13 @@ Function : AbstractFunction {
 	}
 
 	forkIfNeeded { arg clock, quant, stackSize;
-		if(thisThread.isKindOf(Routine), this, { ^this.fork(clock, quant, stackSize) });
+		// We should fork if the clocks don't match.
+		// We should also fork if the quants/stackSize don't match, but that doesn't seem possible.
+		if(thisThread.isKindOf(Routine)
```
telephon, line 160: This still reads very opaque. Better formulate in terms of boolean ops if possible. E.g. like this:

```
thisThread.isKindOf(Routine) and: { clock.isNil or: { clock == thisThread.clock }})
```

telephon (changes requested):
In general, please, yes!

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -155,7 +155,13 @@ Function : AbstractFunction {
 	}
 
 	forkIfNeeded { arg clock, quant, stackSize;
-		if(thisThread.isKindOf(Routine), this, { ^this.fork(clock, quant, stackSize) });
+		// We should fork if the clocks don't match.
+		// We should also fork if the quants/stackSize don't match, but that doesn't seem possible.
+		if(thisThread.isKindOf(Routine) 
+			and: { clock !? { clock == thisThread.clock} ?? {true} },
```
JordanHendersonMusic:
```suggestion
			and: { clock.isNil or: { clock == thisThread.clock }},
```

JordanHendersonMusic:
@telephon should be good now!

telephon (approved):

PR #7617, opened 2026-07-04
classlibrary: rework exceptions 

### LINKED ISSUE #7577: sclang: try doesn't differentiate between throwing and error and returning one.

### LINKED ISSUE #7578: Threads should inherit their exceptionHandlers

### LINKED ISSUE #7608: Restarting Pprotect bricks sclang

JordanHendersonMusic:
A part of #7574

 **I have not yet documented this, I'll do that once we agree on behaviour.**

## Purpose and Motivation

It was not possible to write code the catches all exceptions, this means the unit test library cannot be implemented in sc.

This adds a new 'deep' mode to try, which will catch all lexically enclosed exceptions, even if they are thrown from another thread.

```

try(deep: true) {
   fork { 1.wait; Error().throw }
} {
   "I still get called, even though it was forked when in 'deep' mode".postln;
}

```

---

Other changes.

Returning an error from try is no longer considered the same as throwing
it.
```supercollider
try { Error() } { "meow" }  // used to 'meow', no longer
```

You can now throw anything (like you could before) but it is now passed
to the catch block.
```supercollider
try { 1.throw } { "meow" } // horrary it 'meows'!
```

## Types of changes

- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [ ] Updated documentation


SCClassLibrary/Common/Core/Thread.sc
```diff
@@ -169,17 +175,19 @@ Routine : Thread {
 	}
 
 	// PRIVATE
+	// This method is called in the interpreter from C++ code whenever any thread/clock (e.g., LinkClock) wishes to awaken a task.
 	awake { arg inBeats, inSeconds, inClock;
 		if(rescheduledTime.isNil) {
 			clock = inClock;
-			^this.next(inBeats)
+			^this.next(inBeats);
```
JordanHendersonMusic:
```suggestion
			^this.next(inBeats)
```

SCClassLibrary/Common/Streams/Patterns.sc
```diff
@@ -665,16 +667,12 @@ Pprotect : FilterPattern {
...
-		if(result.isException) {
-			func.value(result, thisThread);
-			result.throw;  // to nested error handlers
-		} {
-			// here, `protect` calls the error handler
-			// but Pprotect has a slightly different meaning
-			^result
-		}
+	embedInStream { |inval|
+		// Despite this class being called protect, it doesn't actually do what protect does.
```
JordanHendersonMusic:
```suggestion
		// Despite this class being called protect, it doesn't actually do what protect does as it doesn't call `func` on the success path.
```

SCClassLibrary/Common/Streams/Stream.sc
```diff
@@ -345,6 +345,9 @@ PauseStream : Stream {
 
 		if (stream.notNil, { "already playing".postln; ^this });
 		if (doReset, { this.reset });
+
```
JordanHendersonMusic:
```suggestion
```

testsuite/classlibrary/TestPprotect.sc
```diff
@@ -1,64 +1,62 @@
...
+		var catches = 0;
+		var cond = CondVar();
+
+		var pat = Pprotect(Routine{Error("meow").throw}, { 
+			catches = catches + 1;
+			cond.signalOne;
+		 });
+
+		// Pprotect rethrows the exception, so we must catch it again to stop it escaping the test function.
+		// This is quite odd because it only runs the handler when an error occurs, so its really like a try that always throws (why?).
```
JordanHendersonMusic:
```suggestion
```

testsuite/classlibrary/TestPprotect.sc
```diff
@@ -1,64 +1,62 @@
...
+		try { pat.play } { 
+			catches = catches + 1;
+			cond.signalOne;
+		};
+		this.assert(cond.waitFor(0.2, { catches == 2 }), "catches == 2 and didn't timeout. Pprotect should clear the stream's exceptionHandler" );
```
JordanHendersonMusic:
```suggestion
		this.assert(cond.waitFor(0.2, { catches == 2 }), "catches == 2 and didn't timeout. Pprotect should catch exceptions, and so should a try around the `p.play`." );
```

testsuite/classlibrary/TestPprotect.sc
```diff
@@ -1,64 +1,62 @@
...
+				caught = caught + 1; 
+			}
 		);
-
-		stream = pat.play;
-		condvar.waitFor(0.1);
-
+ 
+		try { stream = pat.play } { caught = caught + 1};
+		this.assert(condvar.waitFor(0.2, { caught.debug(\caught) == 2 }), "TIMEOUT or caught should be 1");
```
JordanHendersonMusic:
```suggestion
		this.assert(condvar.waitFor(0.2, { caught == 2 }), "TIMEOUT or caught should be 1");
```

testsuite/classlibrary/TestPprotect.sc
```diff
@@ -1,64 +1,62 @@
...
+ 
 		wasReset = false;
 		stream.reset;
-		stream.play;
-		condvar.waitFor(0.1);
-
+		try { stream.play } { caught = caught + 1};
+		this.assert(condvar.waitFor(0.2, {caught.debug(\caught) == 4}), "TIMEOUT or caught shoudl be 2");
```
JordanHendersonMusic:
```suggestion
		this.assert(condvar.waitFor(0.2, {caught == 4}), "TIMEOUT or caught should be 4");
```

SCClassLibrary/Common/Streams/Stream.sc
```diff
@@ -465,16 +467,16 @@ Task : PauseStream {
 	// optimization:
 	// Task builds its own Routine, so 'protect' here
 	*new { arg func, clock;
-		var new = super.new(
+		var self = super.new(
```
JordanHendersonMusic:
```suggestion
		var new = super.new(
```

SCClassLibrary/Common/Streams/Stream.sc
```diff
@@ -465,16 +467,16 @@ Task : PauseStream {
...
 			Routine({ |inval|
 				protect {
 					func.value(inval)
 				} {
-					new.streamError
+					self.streamError
```
JordanHendersonMusic:
```suggestion
					new.streamError
```

SCClassLibrary/Common/Streams/Stream.sc
```diff
@@ -465,16 +467,16 @@ Task : PauseStream {
...
 				};
 			}
 		), clock);
-		^new
+		^self
```
JordanHendersonMusic:
```suggestion
		^new
```

JordanHendersonMusic:
Any one interested in this? It is quite a big change, but I'm pretty sure it is the only way to ensure we can actually catch all exceptions, and therefore write software without the risk of turning sclang into a brick — important in CLI mode when waiting for something to run or when distributing a piece to play as a 'standalone'.

dyfer:
@telephon @jamshark70 would you be able to take a look at this? Thanks!

telephon:
I have no time to look into it currently. Just a though, have you considered that an error thrown within a task leaves the taks in the right condition? The error needs to be passed up the chain of threads.

Tdef(\x, {  Error().throw }).play;

JordanHendersonMusic:
I have and I think I've done what you are suggesting? Id recommend trying this out but as far as I can tell this...


```
try { x.play } { "caught".postln}
```
... Always gets caught because it's set in the event stream player.

telephon:
If I remember the test is whether this will be true in all contexts:

```
(
fork {
	var result = false;
	Tdef(\x, { 1.wait; Error().throw }).play;
	1.2.wait;
	Tdef(\x, { 1.wait; result = true }); // starts again
	2.wait;
	result.postln
}
)
```

JordanHendersonMusic:
Yes it will*. When an exception handler is called from a thread (a) that didn't define it, it just halts the throwing thread (a). It only returns from the try expression when the throwing thread is the same as the one that defined the try block.

This does mean that the handler can be called multiple times, which is new.


* If it isnt true, it's a bug.

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -177,38 +177,40 @@ Function : AbstractFunction {
...
-		wasInProtectedFunc = Exception.inProtectedFunction;
-		thread.exceptionHandler = {|error|
-			thread.exceptionHandler = next; // pop
-			^error
-		};
+	try { |handler| 
+		var result;
+		var wasInProtectedFunc = Exception.inProtectedFunction;
+
+		thisThread.exceptionHandler = thisThread.exceptionHandler.add({ |error|
```
ddw-js-70, line 195: Shouldn't this be `addFunc`? There's `handler.value` below, but if `handler` is a normal array of functions, then this will not call the functions:

```
a = nil.add({ |x| x * 2 }).add({ |x| x - 1000 });
a.value(100);
-> [a Function, a Function]

b = nil.addFunc({ |x| x * 2 }).addFunc({ |x| x - 1000 });
b.value(100);
-> [200, -900]
```

BTW let's please not change the behavior of `.value` on an array of functions. We have a class to handle an array of functions that does get evaluated by `.value`: `FunctionList`. Use that, do not break something fundamental like `.value`. (I'm assuming this PR does change the behavior, because otherwise all of your tests would fail by the exception handlers not getting evaluated. I might have missed something, though.)

SCClassLibrary/Common/Core/Function.sc
```diff
@@ -151,7 +151,7 @@ Function : AbstractFunction {
 	get { arg prevVal; ^prevVal }
 
 	fork { arg clock, quant, stackSize;
-		^Routine(this, stackSize).play(clock, quant);
+		^Routine(this, stackSize, thisThread.exceptionHandler.asArray.copy).play(clock, quant);
```
ddw-js-70, line 154: Hm, per my other comment about FunctionList, this gets a bit tricky. If FunctionList is used (IMO it should be) then `thisThread.exceptionHandler` could be:

- nil: no-op is OK here
- a Function: Can't copy (IIRC?) and no need to force it to be a function list here.
- a FunctionList: Should copy.

Since this pattern appears in a few places, we probably need a dedicated method to handle it by polymorphism.

ddw-js-70 (commented):
Consider this "request changes" -- I didn't have time this week to restore access to my real account.

Also, I have to admit that I haven't deeply, deeply read every line of this PR.

I would like to ask, what is the precedent from other languages for forked threads inheriting exception handlers from the parent thread? There's an assumption here that inheriting exception handlers is obviously the only correct way to do things. I'm not so sure it's that obvious, but I'd be persuaded by "literally nobody does it the (old) SC way." Another way to look at it is that a forked thread is an independent activity that just happens to launch from a certain location, but that the location from which it's launched doesn't affect the task's independence. (An inconsistency in this suggestion is: I depend on threads inheriting the environment from the parent, so why not inherit the exception handler? I'm aware of that... I don't have a strong objection to the new way; I just think it might be more of a gray area than the original PR comment states.)

If `Pprotect` is no longer needed, then... is `Prout {|i| protect { p.embedInStream(i) } { ... }}` *really* an improvement? It seems like it's expanding the syntax with no gain in usability for the end user. TBH I'd just write my own PatternProtect in ddwPatterns, because I don't want to deal with extra tokens for this purpose.

JordanHendersonMusic:
> I would like to ask, what is the precedent from other languages for forked threads inheriting exception handlers from the parent thread?

Well most languages have async functions, all functions as async in SC, and because we use them everywhere, I'm not sure a comparison would be fair.

Edit: it's not actually clear what you/SC means by a forked thread.

>There's an assumption here that inheriting exception handlers is obviously the only correct way to do things.

SC already did this... But only when calling resume. In fact, the implementation of the schedules reimplemented the exception handler for nil, without the halting. 

One thing this pr doesn't work well with in redefining nil.handle exception... But that never worked anyway, because once the thread paused, the next time it throws it goes through the scheduler, which copies out the implementation of nil.


This pr just expands this to include the fork case.

> Another way to look at it is that a forked thread is an independent activity that just happens to launch from a certain location, but that the location from which it's launched doesn't affect the task's independence. 

Yes this is a logical issue, but in practice you just wrap the fork in another try and print the error. Then it behaves exactly the same as you have suggested.

> If Pprotect is no longer needed, then... is Prout {|i| protect { p.embedInStream(i) } { ... }} really an improvement? It seems like it's expanding the syntax with no gain in usability for the end user. TBH I'd just write my own PatternProtect in ddwPatterns, because I don't want to deal with extra tokens for this purpose.

The problem is that pprotect doesn't protect, it's a try - it is also easy to break if you resume it after it has thrown. We could create a new class that just returns the prout in the constructor.

The alternative, which is how I've seen it be used is when you protect a player, now you can just do `protect { p.play } { ... }`.

SCClassLibrary/Common/Core/Function.sc
JordanHendersonMusic: Only the final handler is evaluated and then popped, not all of them. If the user wants the previous handler to be evaluated, they must throw.

SCClassLibrary/Common/Core/Function.sc
JordanHendersonMusic: See above, this is a stack, not a function list.

SCClassLibrary/Common/Core/Thread.sc
```diff
@@ -76,8 +84,13 @@ Thread : Stream {
 		_PrimName
 		^this.primitiveFailed
 	}
-	handleError { arg error;
-		(exceptionHandler ? parent).handleError(error)
+
+	handleError { |error| 
+		^if (this.exceptionHandler.size != 0)  {
+			this.exceptionHandler.pop.handleError(error)
```
JordanHendersonMusic, line 90: @ddw-js-70 - @jamshark70 

This is where the function is popped from the stack.

ddw-js-70:
> Well most languages have async functions, all functions as async in SC, and because we use them everywhere, I'm not sure a comparison would be fair.

I have to admit here, I have no idea what you mean by this. "All functions are async"?

> > There's an assumption here that inheriting exception handlers is obviously the only correct way to do things.
> 
> SC already did this... But only when calling resume.

Also no idea what you mean here. It's possible that I'm just being exceptionally thick, but, some of these messages look like things that you've spent a lot of time thinking about, so they're very clear in your mind. Not equally clear to everyone else.

> Edit: it's not actually clear what you/SC means by a forked thread.

`fork` IMO implies that it's been scheduled and it's running under its own power on some clock. I've always taken this to mean that the new thread is not under control of the block that spawned it.

Taking the first case in #7574:

```
try {
   fork { Error().throw };
   Error().throw
} {  "Caught".postln };
// (new behavior) prints Caught twice
```

... it simply would never have occurred to me, in any way, shape or form, that the error handler would fire twice. The fork thing goes off on its own path. By the time it starts executing, the try block has already finished. The exception handler is gone. This, to me, is clear and logical.

I can see the benefit of the proposal, and I'm not shutting it down. I think in the long run, I'm probably okay with the change. I'm perhaps less certain than you that the old way is obviously terrible and the new way is obviously correct.

> The problem is that pprotect doesn't protect, it's a try

Is it, though?

```
p = Pprotect(
	Prout {
		1.0.yield;
		(1.0 + nil).yield;
	},
	{ |er| er.errorString.debug("pprotect handler") }
);

q = Pseq([p, Pseries(0, 1, 8)], 1).asStream;

q.next;  // 1.0
q.next;  // error thrown
q.next;  // nil -- thread is dead
```

It *uses* prTry but it rethrows the exception and the thread is terminated at that point. That's `protect` behavior, not `try`. `try` would continue with the Pseries.

OK, let's have a look at the Prout way:

```
// Prout way
o = Prout {
	1.0.yield;
	(1.0 + nil).yield;
};

p = Prout { |in| protect { o.embedInStream(in) } { |er| er.errorString.debug("pprotect handler") } };

q = Pseq([p, Pseries(0, 1, 8)], 1).asStream;

q.next;  // 1.0
q.next;  // error thrown
q.next;  // nil -- thread is dead
```

Zero difference in behavior.

The Prout way, btw, does not save us from #7608.

```
o = Routine { 0.01.yield; Error().throw };

p = Prout { |in| protect { o.embedInStream(in) } { |er| er.errorString.debug("pprotect handler") } };

p.play;
p.play;  // crash bang, we used Prout but we're still dead
```

So I feel like the horribleness of Pprotect is being exaggerated. (Possibly this exaggeration is coloring my view of the other, more important parts of the PR.)

I can agree, btw, that the Prout way is clearer and we could replace the current Pprotect with that:

```
	embedInStream { arg inval;
		^protect {
			pattern.embedInStream(inval)
		} { |exc|
			func.value(exc, thisThread);
		}
	}
```

(which I don't remember if we tried it like that before, and there was some problem necessitating the more convoluted implementation...? If there's no exception, `protect` returns the result of the function block, so the prTry chicanery may simply be unnecessary.)

SCClassLibrary/Common/Core/Thread.sc
ddw-js-70: OK, understood.

JordanHendersonMusic:
> I have to admit here, I have no idea what you mean by this. "All functions are async"?

In many other languages, not all functions allocate a frame. Functions which must allocate a frame, because they can be paused and resumed, are labeled async. In SC, all functions can be paused and resumed because all functions allocate a frame (expected tail call opts, but that's different...).

>>> There's an assumption here that inheriting exception handlers is obviously the only correct way to do things.
>>SC already did this... But only when calling resume.
>Also no idea what you mean here. 

```
f = Routine { loop { 10.0.rand.yield } }; \\ A
try {
   f.next;  \\ B
}
```
At 'B' the thread create at 'A' throws any errors into the thread of 'B'. This is done by walking up the thread calling hierarchy.

> ... it simply would never have occurred to me, in any way, shape or form, that the error handler would fire twice. 

Yes this single case is odd. But consider this...
```
try { f.() } { "Caught".post };
```
... and consider possible implementations of `f`.
```
\\ A
f = {
   Routine.play { 1.throw };
   2
};

\\ B
f = {
   var condVar = ...;
   var result;
   fork { 
      ...
      result = 1;
      condvar.signal
   };
   condVar.wait;
   r
};
```

The point of this pr is that without this, it is not possible to write code such that the exceptions in 'B' are caught. This means the entire unit test system is fundamentally broken, and it is impossible to write any cli code because you cannot catch the exception in 'B' (nor 'A').

Given we do a large amount of communication with the server in the style of 'B' this is a pretty serious issue.


> I'm perhaps less certain than you that the old way is obviously terrible and the new way is obviously correct.

It's not obvious. The obvious thing would be to have promises that return values and propagate errors... but we can't do that without breaking the language/rendering basically everyone's code incorrect. 

This is the least damaging way I could find to do this as, as far as I can see, the only downside is having to wrap fork in a try when you truly do mean to disconnect.

> So I feel like the horribleness of Pprotect is being exaggerated.

Yes, I don't really care about it, this PR is about being able to write code that doesn't turn sclang into a brick...

...But you are looking at the wrong example here. It's the success path where it behaves like try.
```
protect { ... } { "meow".postln}; \\ A
Pprotect { ... } { "meow".postln }; \\ B
```
'A' always meows, 'B' only does when there is an error, hence it is a try. We can't change this, even the class library duplicates the logic into both branches.


The reason why I like this so much is because when you write a cli tool now, you literally wrap everything like this...
```
try {
   ... Do the thing ...
} { |er| 
   er.reportError;
   1.exit;
}
```
Now it will never sit there doing nothing, hanging the whole workflow. Yes it does mean you need to wrap all legitimate places where you truly want to disconnect a thread from the hierarchy in another try... But that's pretty easy to do!

ddw-js-70:
> In SC, all functions can be paused and resumed because all functions allocate a frame.

OK... So, any function in SC *could* be paused and resumed, but is not necessarily going to do that in fact. I feel like that's a different statement from saying that all functions are asynchronous, but, OK.

> ```
> f = Routine { loop { 10.0.rand.yield } }; \\ A
> try {
>    f.next;  \\ B
> }
> ```
> At 'B' the thread create at 'A' throws any errors into the thread of 'B'. This is done by walking up the thread calling hierarchy.

Thinking about it overnight, I realize that this is the core of the issue: Should exception handlers follow temporal scope (current behavior), or lexical scope (proposed new behavior)?

And as context for the next couple of paragraphs, I'm going to suggest that in some cases you might want temporal scope, and other cases you might want lexical scope. So it should be up to the user to decide what they want, not to have one or the other forced upon all cases.

By temporal scope, I mean that at the moment `f.next` is called, an exception handler is in force. "B" catches the error from "A" because B's exception handler has not yet expired by virtue of exiting the try block. So I don't see it as an exception being thrown from one thread into another (independent) thread. `f.next` could be any other method call that can throw -- `try` cannot exit until that method call returns, hence any exceptions thrown during the evaluation of that method call must be caught at "B." When `try` exits, the exception handler is gone and any post-hoc side effects (e.g. scheduled activities) are not subject to an exception handler that is gone.

> ```
> \\ A
> try {
>    Routine.play { 1.throw };
>    2
> };
> ```

What you're proposing is that the entire Routine.play should be under control of an exception handler put in place by `try`. That's absolutely desirable for some, probably a lot, of cases.

But I'm not 100% convinced that it's desirable for *all* possible cases.

> This means the entire unit test system is fundamentally broken, and it is impossible to write any cli code because you cannot catch the exception in 'B' (nor 'A').

I would rephrase this. It is possible to catch exceptions in a forked thread if it puts its own exception handler in place.

You are making a correct argument that, especially in the unit test framework, it is not at all reasonable to expect test contributors to remember to write a lot of exception handlers everywhere, in any forked or scheduled anything -- basically, "it's fine as long as you never screw up" is a sad joke.

Sooooo... Is there a way to make this optional? Like, a `deepTry` / `deepProtect` method, and leave the existing try / protect alone?

It looks a bit like: "I found something I don't quite like, and the only way to deal with that is to rip the guts out and break behavior that other users might be relying on." I think the solution could be more narrowly tailored to the specific cases that you've mentioned (which are completely valid cases!).

JordanHendersonMusic:
> Should exception handlers follow temporal scope (current behavior), or lexical scope (proposed new behavior)?

Yes the new behaviour is lexical, but the old behaviour is not temporal, it's still lexical, but has a slightly different meaning.

Forking, in this pr follows *the current* lexical scope, before, it was as if you had written the fork at the top scope. Time has nothing to do with it.

Edit: ah I misunderstood you here. This pr doesn't change the behaviour of next/resume, these are generators, other languages draw a distinction between threads/forking and generators. SC implements them using the same mechanism, but this pr doesn't change this.


> Sooooo... Is there a way to make this optional? Like, a deepTry / deepProtect method, and leave the existing try / protect alone?

Yes, but let's not. By creating two types of try we just have twice the problems. The goal here is to write software that when an exception is thrown, you can catch it. Leaving the old behaviour means we can't do that with changing all code ever written.

> But I'm not 100% convinced that it's desirable for all possible cases.

Regards leaving the existing behaviour, you literal just write, `try { p.play } { |er| er.report error }`. We could easily add a helper.
```
fork(detach: true) {
   ...
}

p.play(detach: true)
```

This seems reasonable for something that happens rarely and let's us all write software that doesn't turn SC into a brick when running cli tools or standalones.

It also makes the default way the 'safe' way, in that exceptions can be caught, and if the user really does want to disconnect, they are in charge of making sure the exceptions are caught... 

To be honest.. I can't really see why you'd ever want to disconnect. It is always better to wrap it in another try.

ddw-js-70:
Anyway, at the end of the day, I'm not standing in the way of this. I think on balance, it's a good change. I find myself a little bit allergic to the overstatement of the case. Making the test suite more robust in the face of errors is by itself enough of a reason to do this; language such as "entire unit test system is fundamentally broken" and "impossible to write any cli code" -- (sorry) -- puts me off. (Although I suppose I indulge in such language from time to time.)

One reason why it puts me off is that a hyperbolic statement of a problem calls for invasive changes, and those changes are implicitly seen as mandatory and urgent. That is implicitly an argument to de-prioritize backward compatibility, without making that argument explicit (and it's not the only time that a discussion has trended in that direction). I would really prefer to settle down and look at specific cases in a more dispassionate way.

For myself, I tend to think that all other things being equal, large breaking changes should be reserved for the next major version, while point-releases should prioritize (not necessarily enforce in every case) backward compatibility. If the argument is that point releases should break backward compatibility mainly because SC4 is hypothetical for the foreseeable future, maybe a better solution is to work concretely on SC4. That is, if we feel we can/should/have to make large-ish breaking changes in the SC3 line, this might point to broader issues with governance and project management that impede SC4 planning.

In this specific case, I don't think it will be that hard for us to adapt to the new behavior, and the test suite is a solid reason to go ahead.

JordanHendersonMusic:
> Is there a way to make this optional? Like, a deepTry / deepProtect method, and leave the existing try / protect alone?

After sitting on this, I think something like this is necessary as currently protect can be triggered more than once... Which makes no sense. Thanks!

JordanHendersonMusic:
So this pr has three commits now. 

One is adding a primitive that checks for out of context return. 

Two makes the syntax `msg(keyword: value) { ...blocks... }` valid. Previously, you had to have a positional arguement, `msg(2, keyword: value) { ... }`. This seems like an oversight and is a trivial change.

Three is the change to the exception handlers which introduces deep try, `try(deep: true) { ... }`, this catches exceptions from other threads allowing you to catch all exceptions.

Would some one mind saying yay or nay? I'll do docs after that?

PR #7619, opened 2026-07-05
classlibrary: basic extention testing

JordanHendersonMusic:
## Purpose and Motivation

Adds basic tests for extension methods.

## Types of changes

- New test 
## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestExtension.sc
+++ b/testsuite/classlibrary/TestExtension.sc
@@ -1,0 +1,24 @@
+TestExtension_Base {
+	*foo { ^\foo }
+	*bar { ^\bar }
+}
+
+TestExtension_Derived : TestExtension_Base {
+	*foo { ^\fooDerived }
+}
+
+TestExtension : UnitTest {
+	test_extensions_basic {
+		this.assertEquals(TestExtension_Base.meow, \meow, "basic extension works.");
+		this.assertEquals(TestExtension_Base.foo, \foo2, "basic extension override works.");
+	}
+	test_extensions_inheritance {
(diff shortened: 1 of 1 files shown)


testsuite/classlibrary/TestExtension.sc
```diff
@@ -0,0 +1,24 @@
...
+	*foo { ^\fooDerived }
+}
+
+TestExtension : UnitTest {
+	test_extensions_basic {
+		this.assertEquals(TestExtension_Base.meow, \meow, "basic extension works.");
+		this.assertEquals(TestExtension_Base.foo, \foo2, "basic extension override works.");
+	}
+	test_extensions_inheritance {
+		this.assertEquals(TestExtension_Derived.foo, \fooDerived, "inherited extension doesn't override.");
```
dyfer: shouldn't the comment be "does override" or "should override"?

JordanHendersonMusic: No an extension override to a *base* class should be overridded by the derived class..... that is still really confusing though!

dyfer: Got it. Then I'd prefer `extension override to a base class should be overridden by the derived class` as that's clearer, if still a bit confusing.

JordanHendersonMusic:
```suggestion
		this.assertEquals(TestExtension_Derived.foo, \fooDerived, "extension override to a base class should be overridden by the derived class.");
```

dyfer (approved):
Thanks!

PR #7621, opened 2026-07-05
ci: run sclang ci even if some unrelated builds fail

JordanHendersonMusic:
## Purpose and Motivation

Even if some build fails, the next step should still run if possible.

TODO: find a way to break one of the unneeded builds and test this! Could possibly just merge this and see what happens when one eventually breaks as this is very low stakes.

## Types of changes

- CI

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -9,9 +9,11 @@ on:
 jobs:
   build-linux:
     runs-on: ubuntu-${{ matrix.os-version }}
+    continue-on-error: ${{ matrix.not-needed-for-ci-test }}
     strategy:
       fail-fast: false
       matrix:
+        not-needed-for-ci-test: [true]
         include:
 
           - job-name: ""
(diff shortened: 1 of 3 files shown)


dyfer:
Thank for this!

I'm not sure if this will work, but I'd prefer to avoid double negation and just use a `true` flag:

```
runs-on: ubuntu-${{ matrix.os-version }}
continue-on-error: ${{ ! matrix.needed-for-running-tests }}
```
and
```
artifact-suffix: "ubuntu-22.04-gcc12"
needed-for-running-tests: true
```

To break one of the builds, I think you can for example change one of the linux builds to an unavailable compiler version, e.g. 
```
          - job-name: ""
            os-version: "22.04"
            c-compiler: "gcc-12312312312"
            cxx-compiler: "g++-12312312312"
            use-syslibs: false
            shared-libscsynth: false
            ctest: false
```

dyfer:
Ah, it seems that this tested itself with the macOS and Windows failures... haha

dyfer:
I tried some things and it seems #7622 is a better approach.

JordanHendersonMusic:
> ${{ ! matrix.needed-for-running-tests }}

Ah, didn't know how to do a 'not'! That's nicer

PR #7622, opened 2026-07-06
GHA: always run tests

dyfer:
## Purpose and Motivation

This seems to be an optimal approach:
- tests will always run, even if one of the matrix jobs fails, unless the run is cancelled
- gha will still report general failure (❌ in the job's overall status) if non-testing build fails
  - this is not the case in #7621, see https://github.com/supercollider/supercollider/actions/runs/28755578849

Here's a run where one of the Linux builds failed, but tests still ran: https://github.com/dyfer/supercollider/actions/runs/28771313520

## Types of changes

- New feature (?)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -88,6 +88,7 @@ jobs:
   test-linux:
     needs: [version-getter, build-linux]
     uses: ./.github/workflows/test_linux.yml
+    if: ${{ !cancelled() }}
     with:
       sc-version: ${{ needs.version-getter.outputs.sc-version }}
     secrets: inherit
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
This means the test will try to run even if the build for it fails right?

dyfer:
Yes, but I believe it will fail due to missing artifact, so I think that's ok.

JordanHendersonMusic (approved):

JordanHendersonMusic:
Merge away!

PR #7553, opened 2026-06-11
Resolve ambiguous help example syntax causing CodeMirror parsing issues

prko:
## Purpose and Motivation
fixes typos related to https://github.com/supercollider/supercollider/issues/7544#issuecomment-4675837216.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #7624, opened 2026-07-06
Fix help browser code editor highlighting

dyfer:
## Purpose and Motivation

This fixes the issue with treating commented out parentheses as valid ones:
```supercollider
(
1.postln;
2.postln;
)
3.postln; // <<<< this was still selected when double clicking or evaluating the previous block with cmd-enter, causing a syntax error due to missing semicolon beforehand
// )
```

I just asked an LLM to figure out how to fix this. I don't think we want to spend too much effort on maintaining the web code editor, so if this is confirmed to work (I only tested on macOS), then I think it should be okay.

EDIT: @prko tested on macOS and [Windows](https://github.com/supercollider/supercollider/pull/7553#issuecomment-4900832821) and confirmed this seems to be working.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/HelpSource/static/editor.js
+++ b/HelpSource/static/editor.js
@@ -81,10 +81,10 @@ const selectRegion = (options = { flash: true }) => {
         let token = editor.getTokenTypeAt(cursor) || ''
         if (cursorLeft.hitSide)
             return cursorLeft
+        if (token.match(/(comment|string|symbol|char)/))
+            return findLeftParen(cursorLeft)
         let ch = editor.getLine(cursorLeft.line)
             .slice(cursorLeft.ch, cursorLeft.ch+1)
-        if (token.match(/^(comment|string|symbol|char)/))
-            return findLeftParen(cursorLeft)
         if (ch === ')')
             return findLeftParen(findLeftParen(cursorLeft))
         if (ch === '(')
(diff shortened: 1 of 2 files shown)


dyfer:
@capital-G what do you think of this change?

capital-G:
> @capital-G what do you think of this change?

I don't know a thing about lexers, maybe @JordanHendersonMusic can take a quick look if this provides an improvement over the status quo?

dyfer:
This is not about the lexer, lexing behavior is unchanged. This seemed to be a logic issue when deciding what to select.

JordanHendersonMusic:
The big issue here is that the `*.scd` file format does not contain valid sc code and doesn't define how invalid code is handled. Regions are just one example of where this messes up.

There is literally no definition of what a region is anywhere, so whatever this PR does is as correct as anything else.

We need to a proper definition and this should be written in c++ so the ide sends the whole file and says, evaluate the region that includes this point (where the cursor is). This way, editors don't need to keep re-implementing this nasty piece of code.

If you wanna merge this I really don't care (so long as it doesn't crash or something silly). Have you tried some more obscure examples with code with incorrect syntax in...
```
((((

( 1 + 1 );
```

(Do we even need to support such things in the help browser?)

> This is not about the lexer. This seemed to be a logic issue when deciding what to select.

No, its about the parser (or the fact this should be included in the parser but isn't). You are implementing a parser here (well editing one).

dyfer:
> Have you tried some more obscure examples with code with incorrect syntax in...
> 
> ```
> ((((
> 
> ( 1 + 1 );
> ```

Just to make this clear: 
This PR fixes an issue with a _correct_ syntax (when unbalanced parentheses are included in a comment). 

In this specific case, putting cursor in the line with (1+1) and pressing shift-enter evaluates the line. 
But pressing cmd-enter tries to evaluate the whole block and fails...

JordanHendersonMusic (approved):

JordanHendersonMusic:
Okay that sounds good then! and this is only in the help browser too so I don't think we need to worry about incorrext syntax.

dyfer:
Thanks! In that case I'll merge this.

PR #7626, opened 2026-07-07
Keep IDE in focus

### LINKED ISSUE #3756: The Mac OS X menu bar disappears at startup when the help-browser is docked.
The menu bar is available only if the help browser exists and is detached.

This is annoying because by starting SuperCollider on double-clicking an .scd file, it is not possible to load the project by making "Evaluate File".
Some users that are only musiciens think that the program is broken, unless they think to click in the window.

This behavior happens in versions 3.7.2, 3.8.1 and 3.9.3, but not in 3.6.6.
The problem doesn't exist in Windows.

dyfer:
## Purpose and Motivation

This PR introduces two workarounds (?) to keep the code editor in focus:
- it brings the focus back after recompilation (fixes #3756)
- ~~it brings the focus back after~~ it keeps the editor focus when the help browser loads

Tested on macOS. I'm not sure how focusing behaves on ~~Windows and~~ various Linux environments. EDIT: I tested on Windows, it seems to be working fine (the focus ~~is brought back after~~ stays in the editor when loading the help).

Changes were implemented with a LLM help.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested:
  - [x] on macOS
  - [x] on Windows
  - [ ] on Linux
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/editors/sc-ide/core/sc_process.cpp
+++ b/editors/sc-ide/core/sc_process.cpp
@@ -368,6 +368,13 @@ void ScProcess::onResponse(const QString& selector, const QString& data) {
     else if (selector == QStringLiteral("classLibraryRecompiled")) {
         mCompiled = true;
         emit classLibraryRecompiled();
+
+        // bring the ide window to focus
+        static bool firstStart = true;
+        if (firstStart && MainWindow::instance()) {
+            MainWindow::instance()->raise();
+            firstStart = false;
+        }
     }
 
     else if (selector == QStringLiteral("requestCurrentPath"))
--- a/editors/sc-ide/widgets/help_browser.cpp
+++ b/editors/sc-ide/widgets/help_browser.cpp
@@ -366,8 +366,6 @@ void HelpBrowser::onScResponse(const QString& command, const QString& data) {
     mWebView->load(urlString);
 
     HelpBrowserDocklet* helpDock = MainWindow::instance()->helpBrowserDocklet();
-    if (helpDock)
-        helpDock->focus();
 
     emit urlChanged();
 }


editors/sc-ide/core/sc_process.cpp
```diff
@@ -368,6 +368,12 @@ void ScProcess::onResponse(const QString& selector, const QString& data) {
     else if (selector == QStringLiteral("classLibraryRecompiled")) {
         mCompiled = true;
         emit classLibraryRecompiled();
+
+        // bring the ide window to focus
+        if (MainWindow::instance()) {
+            MainWindow::instance()->raise();
+            MainWindow::instance()->activateWindow();
+        }
```
JordanHendersonMusic, line 377: what if the main window wasn't the last window in focus?

editors/sc-ide/widgets/main_window.cpp
```diff
@@ -224,6 +225,17 @@ MainWindow::MainWindow(Main* main): mMain(main), mClockLabel(0), mDocDialog(0) {
...
+    // bring focus back to the IDE after the help browser finishes loading
+    connect(mHelpBrowserDocklet->browser(), &HelpBrowser::loadFinished, this, [this]() {
+        static bool initialStartup = true;
+        if (initialStartup) {
+            QTimer::singleShot(0, this, &MainWindow::focusCodeEditor);
+            initialStartup = false;
+        }
+    });
+#endif // SC_USE_QTWEBENGINE
+
```
JordanHendersonMusic: This is something that always annoys me! Thanks for fixing this!

editors/sc-ide/core/sc_process.cpp
dyfer: Good question. Currently we always bring the IDE to focus...  so if the recompile is triggered from the help browser, the ide will focus afterwards (which is maybe not ideal).

One issue I just noticed that is that if one triggers the recompile and then switches away from the IDE (before the recompile finishes), the IDE would pop back up after the recompile.

How about I add a flag so this only happens on the first start of sclang inside the IDE?

editors/sc-ide/widgets/main_window.cpp
dyfer: Thanks!

BTW I realized we don't really need to defer this action if it's triggered from the webpage reload, so I removed the timer.

editors/sc-ide/core/sc_process.cpp
dyfer: I tried adding the flag so this only happens on the first boot. Subsequent sclang starts don't seem to steal focus, at least on macOS...

dyfer: Now I can only observe focus "stealing" if I initiate starting the IDE, then shift focus to another app; the IDE would pop back up in focus after sclang starts.

However, this seems to match behavior of other apps that take a while to start.

JordanHendersonMusic (approved):

dyfer:
Do we need to test on Linux? Or can I merge this?

dyfer:
@JordanHendersonMusic I have changed the implementation for the help browser focus issue. It seems that we [don't need to explicitly set focus](https://github.com/supercollider/supercollider/pull/7626/changes#diff-5d16e30e64ff050d5b52d5a15ea93e0840da67d6ea93a67ca64b799869129d4bL369-L370) to the help browser. With that change, the focus never leaves the editor, so there's no need for a callback. The help docklet still seems to come to focus when one triggers help from the IDE. Am I missing something here, or is this approach fine?

editors/sc-ide/widgets/main_window.cpp
dyfer: Please note that I changed the approach to this altogether. Take a look if that's okay.

dyfer:
Also, we could consider wrapping the [sclang focus fix](https://github.com/supercollider/supercollider/pull/7626/changes#diff-b0fe6d5ca9796c566fa98c99704922d51c96ae8296afb9053207d13b05cbe073R373-R377) in a macOS platform switch - it doesn't seem to make a difference on Windows, and I'm guessing it's also not strictly needed on Linux...?

capital-G (approved):
Thanks for fixing this marcin, much appreciated!

Tested on fedora 44, did not create any problems, so I think this can be merged as-is.

PR #7632, opened 2026-07-10
Display method docs in autocomplete menu

### LINKED ISSUE #7629: Display inline docs for methods
There is already the displaying of the class docs within the auto-complete dialog of Classes

Image

I propose to also display the docs of methods in the autocomplete dialog of methods in a similar manner (an implementation sketch is at https://github.com/capital-G/supercollider/tree/inline-method-docs)

https://github.com/user-attachments/assets/3f54409f-be72-46fb-9cfd-ff130fadde04

One change I did here was to explode the same methods instead of collapsing them. This is the current behavior

Image

and this would be the behavior w/ the change in place

Image

This is one change I am unsure about - I could continue collapsing this and only display it on the implementation resolving (see below), but I think having the ability to look into the documentation is actually helpful in that case, although this may lead to further scrolling down in other cases

Image

Before an implementation in a PR gets discussed I thought we may want to discuss if this is a desirable feature (IMO yes) and also get some feedback about how this feature should be.
A future LSP would provide the same information, though via LSP instead of resloving this via Qt code ;) But this is a low hanging fruit, so I think it would be fine to use this until the LSP has arrived.

capital-G:
Closes #7629.

## Purpose and Motivation

This PR displays the docs of methods in the autocomplete dialog of methods in a similar manner to the already existing class docs

https://github.com/user-attachments/assets/3f54409f-be72-46fb-9cfd-ff130fadde04

One change I did here was to explode the same methods instead of collapsing them. This is the current behavior

old | new
--- | ---
Image | Image

This introduces some (excessive?) scrolling if one wants to skim through all available methods, but I'd argue that b/c of the minimum of three chars this has never been intended for that usage.
This may change w/ some proper introspection from a future LSP, which is hopefully due in 3.16.

But IMO already a completion like this is very helpful in everyday usage

image

## Types of changes

- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/editors/sc-ide/widgets/code_editor/autocompleter.cpp
+++ b/editors/sc-ide/widgets/code_editor/autocompleter.cpp
@@ -757,12 +757,13 @@ void AutoCompleter::fillClassHelp(DocNode* node, QString& infos) {
                 .arg(mCompletion.menu->currentText());
 }
 
-void AutoCompleter::fillMethodHelp(DocNode* node, QString& infos, const QString& methodName, const QString& className) {
+void AutoCompleter::fillMethodHelp(const DocNode& node, QString& infos, const QString& methodName,
+                                   const QString& className) {
     QString instanceMethods = getMethodDocs(node, methodName, false);
     infos = QStringLiteral("<h4>%1-%2</h4>%3").arg(className).arg(methodName).arg(instanceMethods);
 }
 
-void AutoCompleter::fillClassMethodHelp(DocNode* node, QString& infos, const QString& methodName,
+void AutoCompleter::fillClassMethodHelp(const DocNode& node, QString& infos, const QString& methodName,
                                         const QString& className) {
     QString instanceMethods = getMethodDocs(node, methodName, true);
     infos = QStringLiteral("<h4>%1#%2</h4>%3").arg(className).arg(methodName).arg(instanceMethods);
@@ -798,10 +799,10 @@ void AutoCompleter::updateCompletionMenuInfo() {
         fillClassHelp(node, infos);
         break;
     case MethodCompletion:
-        fillMethodHelp(node, infos, methodName, className);
+        fillMethodHelp(*node, infos, methodName, className);
         break;
     case ClassMethodCompletion:
-        fillClassMethodHelp(node, infos, methodName, className);
+        fillClassMethodHelp(*node, infos, methodName, className);
         break;
     case InvalidCompletion:
         break;
@@ -1228,29 +1229,29 @@ QString AutoCompleter::parseClassElement(DocNode* node, QString element) {
     return QString();
 }
 
-bool isMatchingMethod(DocNode* node, QString methodName) {
-    if (QString(node->id) == "METHODNAMES" && node->n_childs > 0 && node->children[0]->text == methodName) {
+bool isMatchingMethod(const DocNode& node, QString methodName) {
+    if (QString(node.id) == "METHODNAMES" && node.n_childs > 0 && node.children[0]->text == methodName) {
         return true;
     }
-    for (int i = 0; i < node->n_childs; i++) {
-        if (isMatchingMethod(node->children[i], methodName))
+    for (int i = 0; i < node.n_childs; i++) {
+        if (isMatchingMethod(*node.children[i], methodName))
             return true;
     }
     return false;
 }
 
-QString AutoCompleter::getMethodDocs(DocNode* node, QString methodName, bool isClassMethod) {
-    if (QString(node->id) == (isClassMethod ? "CMETHOD" : "IMETHOD") && isMatchingMethod(node, methodName)
-        && node->n_childs >= 2) {
+QString AutoCompleter::getMethodDocs(const DocNode& node, QString methodName, bool isClassMethod) const {
+    if (QString(node.id) == (isClassMethod ? "CMETHOD" : "IMETHOD") && isMatchingMethod(node, methodName)
+        && node.n_childs >= 2) {
         QString str;
         // first child is method name, which we want to skip here
         // we want straight access to the body of the method doc
-        parseClassNode(node->children[1], &str);
+        parseClassNode(node.children[1], &str);
         return str;
     }
 
-    for (int i = 0; i < node->n_childs; i++) {
-        QString ret = getMethodDocs(node->children[i], methodName, isClassMethod);
+    for (int i = 0; i < node.n_childs; i++) {
+        QString ret = getMethodDocs(*node.children[i], methodName, isClassMethod);
         if (!ret.isEmpty())
             return ret;
     }
--- a/editors/sc-ide/widgets/code_editor/autocompleter.cpp
+++ b/editors/sc-ide/widgets/code_editor/autocompleter.cpp
@@ -1235,7 +1235,7 @@ bool isMatchingMethod(DocNode* node, QString methodName) {
     return false;
 }
 
-QString AutoCompleter::getMethodDocs(DocNode* node, QString methodName, bool isClassMethod = false) {
+QString AutoCompleter::getMethodDocs(DocNode* node, QString methodName, bool isClassMethod) {
     if (QString(node->id) == (isClassMethod ? "CMETHOD" : "IMETHOD") && isMatchingMethod(node, methodName)
         && node->n_childs >= 2) {
         QString str;
--- a/editors/sc-ide/widgets/code_editor/autocompleter.cpp
+++ b/editors/sc-ide/widgets/code_editor/autocompleter.cpp
@@ -763,10 +763,10 @@ void AutoCompleter::fillMethodHelp(DocNode* node, QString& infos, const QString&
     infos = QStringLiteral("<h4>%1-%2</h4>%3").arg(className).arg(methodName).arg(instanceMethods);
 }
 
-void AutoCompleter::fillClassMethodHelp(DocNode* node, QString& infos, const QString& methodName) {
-    // @todo derive via meta class
-    QString instanceMethods = parseClassElement(node, "CLASSMETHODS");
-    infos = QStringLiteral("<h4>Class Method %1</h4>%2").arg(methodName).arg(instanceMethods);
+void AutoCompleter::fillClassMethodHelp(DocNode* node, QString& infos, const QString& methodName,
+                                        const QString& className) {
+    QString instanceMethods = getMethodDocs(node, methodName, true);
+    infos = QStringLiteral("<h4>%1#%2</h4>%3").arg(className).arg(methodName).arg(instanceMethods);
 }
 
 void AutoCompleter::updateCompletionMenuInfo() {
@@ -775,6 +775,11 @@ void AutoCompleter::updateCompletionMenuInfo() {
         return;
     }
     auto className = klass->name.get();
+    // strip Meta_ prefix for resolution of docs
+    if (className.startsWith("Meta_")) {
+        className.remove(0, 5);
+    }
+
     DocNode* node = parseHelpClass(findHelpClass(className));
     if (!node) {
         mCompletion.menu->addInfo(QString());
@@ -793,7 +798,7 @@ void AutoCompleter::updateCompletionMenuInfo() {
         fillMethodHelp(node, infos, methodName, className);
         break;
     case ClassMethodCompletion:
-        fillClassMethodHelp(node, infos, methodName);
+        fillClassMethodHelp(node, infos, methodName, className);
         break;
     case InvalidCompletion:
         break;
@@ -1250,8 +1255,6 @@ QString AutoCompleter::getMethodDocs(DocNode* node, QString methodName, bool isC
 
 void AutoCompleter::parseClassNode(DocNode* node, QString* str) {
     QString id = node->id;
-
-    std::cout << node->id << std::endl;
 
     if (id == "NOTE")
         str->append("<br><br>Note:<br>");
--- a/editors/sc-ide/widgets/code_editor/autocompleter.cpp
+++ b/editors/sc-ide/widgets/code_editor/autocompleter.cpp
@@ -760,13 +760,13 @@ void AutoCompleter::fillClassHelp(DocNode* node, QString& infos) {
 void AutoCompleter::fillMethodHelp(const DocNode& node, QString& infos, const QString& methodName,
                                    const QString& className) {
     QString instanceMethods = getMethodDocs(node, methodName, false);
-    infos = QStringLiteral("<h4>%1-%2</h4>%3").arg(className).arg(methodName).arg(instanceMethods);
+    infos = QStringLiteral("<h4><code>%1-%2</code></h4>%3").arg(className).arg(methodName).arg(instanceMethods);
 }
 
 void AutoCompleter::fillClassMethodHelp(const DocNode& node, QString& infos, const QString& methodName,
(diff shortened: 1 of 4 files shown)


JordanHendersonMusic:
As I've mentioned, I'm aiming for an LSP in 3.16, so I'd like to hold off until then because this behaviour might have to be reverted.

capital-G:
> As I've mentioned, I'm aiming for an LSP in 3.16, so I'd like to hold off until then because this behaviour might have to be reverted.

I think this PR is rather miniscule and worth alone for one release (which would be 3.15) based on how much it improves the visibility of the docs - at least it helps me a lot during everyday coding.
The changes can be removed easily and we can collect some feedback on how to display docs from the users.

Once we have a LSP, much of the Qt IDE needs to be rewritten nonetheless (e.g. syntax highlighter, formatter, introspection, etc), so I think these 100 LOC are not such a big obstacle and I can also take the task of re-adapting it for the LSP.

editors/sc-ide/widgets/code_editor/completion_menu.cpp
```diff
@@ -88,6 +88,14 @@ const ScLanguage::Method* CompletionMenu::currentMethod() {
     return item ? item->data(MethodRole).value<const ScLanguage::Method*>() : 0;
 }
 
+ScLanguage::Class* CompletionMenu::currentClass() {
+    // @todo use std::optional as return value
```
JordanHendersonMusic: Do you want to do this? If not you could just remove this, pointers should always be assumed to be optional, otherwise they would be a reference.

editors/sc-ide/widgets/code_editor/completion_menu.hpp
```diff
@@ -58,6 +58,7 @@ class CompletionMenu : public PopUpWidget {
     void setCompletionRole(int role);
     QString currentText();
     const ScLanguage::Method* currentMethod();
+    ScLanguage::Class* currentClass();
```
JordanHendersonMusic: can this be const? Then method that is...
```suggestion
    ScLanguage::Class* currentClass() const;
```

editors/sc-ide/widgets/code_editor/autocompleter.hpp
```diff
@@ -132,8 +132,15 @@ private slots:
     static QString findHelpClass(QString klass);
     static DocNode* parseHelpClass(QString file);
     static QString parseClassElement(DocNode* node, QString element);
+    QString getMethodDocs(DocNode* node, QString methodName, bool isClassMethod);
```
JordanHendersonMusic: I think node should be a reference here instead because you don't check for nil, also it can be const.
```suggestion
    QString getMethodDocs(const DocNode& node, QString methodName, bool isClassMethod);
```
Can the method be const as well?

```suggestion
    QString getMethodDocs(const DocNode& node, QString methodName, bool isClassMethod) const;
```

editors/sc-ide/widgets/code_editor/autocompleter.cpp
```diff
@@ -1189,12 +1229,45 @@ QString AutoCompleter::parseClassElement(DocNode* node, QString element) {
...
+    if (QString(node->id) == (isClassMethod ? "CMETHOD" : "IMETHOD") && isMatchingMethod(node, methodName)
+        && node->n_childs >= 2) {
+        QString str;
+        // first child is method name, which we want to skip here
+        // we want straight access to the body of the method doc
+        parseClassNode(node->children[1], &str);
+        return str;
+    }
+
+    for (int i = 0; i < node->n_childs; i++) {
```
JordanHendersonMusic:
```suggestion
    for (size_t i = 0; i < node->n_childs; i++) {
```

editors/sc-ide/widgets/code_editor/autocompleter.cpp
```diff
@@ -1189,12 +1229,45 @@ QString AutoCompleter::parseClassElement(DocNode* node, QString element) {
     return QString();
 }
 
+bool isMatchingMethod(DocNode* node, QString methodName) {
+    if (QString(node->id) == "METHODNAMES" && node->n_childs > 0 && node->children[0]->text == methodName) {
+        return true;
+    }
+    for (int i = 0; i < node->n_childs; i++) {
```
JordanHendersonMusic:
```suggestion
    for (size_t i = 0; i < node->n_childs; i++) {
```

editors/sc-ide/widgets/code_editor/autocompleter.cpp
```diff
@@ -751,23 +744,69 @@ void AutoCompleter::onCompletionMenuFinished(int result) {
...
+    QString infos;
+
+    auto methodName = mCompletion.menu->currentText();
+    // setter methods are appended with `_`, but SCDoc maps does not have documentation for a setter method,
+    // so we fall back to the getter method.
+    if (methodName.endsWith("_")) {
+        methodName.chop(1);
+    }
+
+    switch (mCompletion.type) {
```
JordanHendersonMusic:
```suggestion
    auto methodName = mCompletion.menu->currentText();
    // setter methods are appended with `_`, but SCDoc maps does not have documentation for a setter method,
    // so we fall back to the getter method.
    if (methodName.endsWith("_")) {
        methodName.chop(1);
    }
    
    QString infos;
    switch (mCompletion.type) {
```
Moving the `infos` closer to the switch makes it clearer where this is filled.

editors/sc-ide/widgets/code_editor/autocompleter.cpp
```diff
@@ -507,7 +507,9 @@ CompletionMenu* AutoCompleter::menuForClassCompletion(CompletionDescription cons
 
     for (ClassMap::const_iterator it = matchStart; it != matchEnd; ++it) {
         Class* klass = it->second.data();
-        menu->addItem(new QStandardItem(klass->name));
+        auto item = new QStandardItem(klass->name);
+        item->setData(QVariant::fromValue(klass), CompletionMenu::ClassRole);
+        menu->addItem(item);
```
JordanHendersonMusic:
```suggestion
        auto* item = new QStandardItem(klass->name);
        menu->addItem(item);
        item->setData(QVariant::fromValue(klass), CompletionMenu::ClassRole);
```
I *believe* qt will walk the graph from the application to free memory. If the `setData` throws an exception, the new object isn't connected to the graph and will leak.

It is also customary to use `auto*` when it is a pointer, this makes it clear you aren't making a copy, same as `auto&`.

editors/sc-ide/widgets/code_editor/autocompleter.hpp
```diff
@@ -132,8 +132,15 @@ private slots:
...
     static void parseClassNode(DocNode* node, QString* str);
 
+    // fill info box w/ docs
+
+    void fillClassHelp(DocNode* node, QString& infos);
+    void fillMethodHelp(DocNode* node, QString& infos, const QString& methodName, const QString& className);
+    void fillClassMethodHelp(DocNode* node, QString& infos, const QString& methodName, const QString& className);
```
JordanHendersonMusic: You don't actually need to pass by reference here because `QStrings` implement copy on write so are cheap to pass by value... but I don't think this really matters and has already been done in the code base.

editors/sc-ide/widgets/code_editor/autocompleter.cpp
```diff
@@ -751,23 +744,69 @@ void AutoCompleter::onCompletionMenuFinished(int result) {
...
+    if (!examples.isEmpty())
+        examples.prepend("<h4>Examples</h4>");
+    // MSVStudio 2013 does not concatenate multiple QStringliterals ("""") properly
+    // see http://blog.qt.io/blog/2014/06/13/qt-weekly-13-qstringliteral/
+    infos = QStringLiteral("<h4>%1</h4>%2%3<p><a href=\"%4\">go to help</a>")
+                .arg(parseClassElement(node, "SUMMARY"))
+                .arg(parseClassElement(node, "DESCRIPTION"))
+                .arg(examples)
+                .arg(mCompletion.menu->currentText());
+}
```
JordanHendersonMusic: why doesn't this just return a QString?

editors/sc-ide/widgets/code_editor/autocompleter.cpp
```diff
@@ -751,23 +744,69 @@ void AutoCompleter::onCompletionMenuFinished(int result) {
...
+
+void AutoCompleter::fillClassMethodHelp(DocNode* node, QString& infos, const QString& methodName,
+                                        const QString& className) {
+    QString instanceMethods = getMethodDocs(node, methodName, true);
+    infos = QStringLiteral("<h4>%1#%2</h4>%3").arg(className).arg(methodName).arg(instanceMethods);
+}
+
 void AutoCompleter::updateCompletionMenuInfo() {
-    DocNode* node = parseHelpClass(findHelpClass(mCompletion.menu->currentText()));
+    auto klass = mCompletion.menu->currentClass();
```
JordanHendersonMusic: does this work?
```suggestion
    const auto* klass = mCompletion.menu->currentClass();
```

editors/sc-ide/widgets/code_editor/autocompleter.cpp
capital-G: I sometimes find it easier to reason about ownership when I pass an already existing resource into a function instead of letting the function create the resource for me.
This also allows to modify any existing infos, though r/n this string is always empty.

editors/sc-ide/widgets/code_editor/completion_menu.cpp
capital-G: ah thx - was just an idea, but yeah, optional doesn't seem that useful when working w/ pointers.

editors/sc-ide/widgets/code_editor/autocompleter.hpp
capital-G: I think using a const ref might be cleaner here, but this would create some inconsistencies w/ the current implementation which is using `DocNode*` everywhere.
Any advice what should be preferred here? Consistency or "cleanness"?

editors/sc-ide/widgets/code_editor/autocompleter.cpp
capital-G: Should we really do this? `node->n_childs` is `int`, not `size_t`, so I thought going w/ `int` would be the preferred way in that case?

editors/sc-ide/widgets/code_editor/autocompleter.hpp
capital-G: I don't get why we use `QString` in the first place over the native `std::string` - I just use it like I would use a normal string.

capital-G:
@JordanHendersonMusic applied most changes except for `size_t` and `DocNode*` - see discussion.

editors/sc-ide/widgets/code_editor/autocompleter.hpp
JordanHendersonMusic: Qt requires utf16... its horrid.

You want to use them whenever doing qt stuff.

editors/sc-ide/widgets/code_editor/autocompleter.cpp
JordanHendersonMusic: oh is it!! That's dumb. okay then leave it.

editors/sc-ide/widgets/code_editor/autocompleter.cpp
JordanHendersonMusic: > I sometimes find it easier to reason about ownership when I pass an already existing resource into a function instead of letting the function create the resource for me.

That isn't good C++ (in most cases).... it is good C though!

There are benefits of doing it the way you have done, you can chose whether the object lives on the stack or on the heap... but since QStrings are always heap allocated this is meaningless here.

> when I pass an already existing resource 

the QString doesn't 'exist' as a resource yet... it is basically just an empty pointer. It is like you have written this...
```cpp
void assignInt(int* loc) {
 *loc = 2;
}
int y 
int* x = &y;
assignInt(x);
```
just do 
```cpp
const auto y = getInt();
```

It sounds like you are not yet used to move semantics and RVO?

> This also allows to modify any existing infos, though r/n this string is always empty.

No it doesn't because you call the move assignment operator. `infos = QStringLiteral("<h4>%1</h4>%2%3<p><a href=\"%4\">go to help</a>")`

editors/sc-ide/widgets/code_editor/autocompleter.hpp
JordanHendersonMusic: I'd write it correctly and leave the old code to be wrong. Its a new function so you control the call sight and can easily de-reference it. Alternatively you could add a null check... but the function is pretty meaningless if you pass in a nullptr.

editors/sc-ide/widgets/code_editor/autocompleter.cpp
capital-G: > It sounds like you are not yet used to move semantics and RVO?

I think move semantics somehow, but haven't heard of RVO yet.
Maybe it is just adapting bad style from the SC code base ;)

I switched now to a `QString AutoCompleter::fillClassHelp(DocNode* node) const` signature.

> > This also allows to modify any existing infos, though r/n this string is always empty.
> 
> No it doesn't because you call the move assignment operator. `infos = QStringLiteral("<h4>%1</h4>%2%3<p><a href=\"%4\">go to help</a>")`

Well, you could mutate it in that case, its up to the implementation, thats why I think this is actually a somehow nicer implementation, especially since we can't make use of the auto keyword allocation b/c of the switch case.
But lets stick w/ the easy/straight-forward one!

capital-G:
@JordanHendersonMusic applied the suggested changes, thanks for this, much appreciated! :)

editors/sc-ide/widgets/code_editor/autocompleter.cpp
JordanHendersonMusic: > I think move semantics somehow, but haven't heard of RVO yet.


RVO (return value optimisation, or copy-elision) is when you return a new object from a function, the compiler guarantees it won't make a copy to the call sight.

```cpp
T makeT() { return T{}; } // < make a resource
T my_t = makeT(); // < RVO means we don't copy it into to 'my_t'
```
In other words, the compiler turns the code above into the following code...
```cpp
T initT(void* t) { new (t) T{}; }
T my_t; // unitialisied;
initT(my_t); 
```

> Maybe it is just adapting bad style from the SC code base ;)

Yeah, this isn't a good code base to learn from!

JordanHendersonMusic (approved):

editors/sc-ide/widgets/code_editor/autocompleter.cpp
```diff
@@ -751,24 +744,69 @@ void AutoCompleter::onCompletionMenuFinished(int result) {
     // quitCompletion("cancelled");
 }
 
+QString AutoCompleter::fillClassHelp(DocNode* node) const {
+    QString examples = parseClassElement(node, "EXAMPLES");
+    if (!examples.isEmpty())
+        examples.prepend("<h4>Examples</h4>");
+    // MSVStudio 2013 does not concatenate multiple QStringliterals ("""") properly
+    // see http://blog.qt.io/blog/2014/06/13/qt-weekly-13-qstringliteral/
```
dyfer, line 752: This seems to be an outdated comment, maybe there's no need to include it in this function?

PR #7635, opened 2026-07-10
Make http:// and https:// link-click functional in the Post window

### LINKED ISSUE #7634: PostWindowURLHandler does not know how to handle the URL
## Environment

* SuperCollider version:
   * Windows: 3.15.0-dev [1062f4f] (downloaded on July 7, 2026)
   * macOS: 3.15.0-dev [a00e3e7] (downloaded on July 4, 2026)
* Operating system: Windows, macOS (not tested on Linux, but it should be the same.)

## Steps to reproduce

1. Evaluate the following line: `"https://github.com/supercollider/supercollider"`
2. Click the underlined string in the Post window:
    `-> https://github.com/supercollider/supercollider`

### Expected
HelpBrowser or SystemBrowser should open the link.

### Actual
```
WARNING: PostWindowURLHandler does not know how to handle the URL 'https://github.com/supercollider/supercollider'
```

prko:
fixes: #7634
## Purpose and Motivation

This PR makes http:// and https:// links clickable in the Post window.
Before this PR. clicking http:// and https:// was warned as below:
>WARNING: PostWindowURLHandler does not know how to handle the URL 'https://github.com/supercollider/supercollider'

## Types of changes

- Bug fix
- New feature

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -26,7 +26,6 @@ ScIDE {
 		});
 
 		PostWindowURLHandler.register(\http, {|url|
-			var prefix = "http://";
 			HelpBrowser.goTo(url)
 		});
 
--- a/SCClassLibrary/scide_scqt/ScIDE.sc
+++ b/SCClassLibrary/scide_scqt/ScIDE.sc
@@ -27,12 +27,12 @@ ScIDE {
 
 		PostWindowURLHandler.register(\http, {|url|
 			var prefix = "http://";
-			PostWindowURLHandler.prOpenRedirectMenu(url)
+			HelpBrowser.goTo(url)
 		});
 
 		PostWindowURLHandler.register(\https, {|url|
 			var prefix = "http://";
-			PostWindowURLHandler.prOpenRedirectMenu(url)
+			HelpBrowser.goTo(url)
 		});
 
 		StartUp.add {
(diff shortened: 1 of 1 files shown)


SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -25,6 +25,16 @@ ScIDE {
 			{ Document.open("/" ++ parts[0], parts[1] ?? { 0 }, parts[2] ?? [0]) }.fork(AppClock)
 		});
 
+		PostWindowURLHandler.register(\http, {|url|
+			var prefix = "http://";
+			PostWindowURLHandler.prOpenRedirectMenu(url)
+		});
```
JordanHendersonMusic, line 30: Why can't this just be `{|url| HelpBrowser.open(url) }`?

prko: I adopted the idea from HelpBrowser: when clicking a URL, the user can choose whether to open it in HelpBrowser or in the system’s default web browser. I think this is cool, but I’m not sure if it’s the best approach. I could simply use `{ |url| HelpBrowser.open(url) }`, which would also work without any issues.

Note: At the moment, when clicking a download link, the user can choose where to open it, and selecting the system’s default web browser downloads the linked file. This is convenient. I saw the message  
`WARNING: PostWindowURLHandler does not know how to handle the URL`  
when clicking the link in the following warning in the Post window:

```
ERROR: API version mismatch: /Users/prko/Library/Application Support/SuperCollider/Extensions/SC3plugins/HOAUGens/HOAMirror3.scx
    This plugin is not compatible with SuperCollider >=3.15
    The plugin has not been loaded; please find or compile a newer version.
    Releases of sc3-plugins can be downloaded from https://github.com/supercollider/sc3-plugins/releases
    (Plugin's API version: 3. Expected: 6)
```

JordanHendersonMusic: Simple is better here, this is meant as a small class that just dispatches elsewhere.

prko: Thanks, done in https://github.com/supercollider/supercollider/pull/7635/commits/dc9ca6f93e07eb57be7fb23baa4b58f858321908.

The functionality is the same as before with just a simple line.
I had thought the menu might pop up in a strange position or even off‑screen, but that was simply a mistaken assumption without proper reasoning. Thanks!

SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -25,6 +25,16 @@ ScIDE {
 			{ Document.open("/" ++ parts[0], parts[1] ?? { 0 }, parts[2] ?? [0]) }.fork(AppClock)
 		});
 
+		PostWindowURLHandler.register(\http, {|url|
+			var prefix = "http://";
```
JordanHendersonMusic:
```suggestion
```
You don't need these!

SCClassLibrary/scide_scqt/ScIDE.sc
```diff
@@ -25,6 +25,16 @@ ScIDE {
...
+			HelpBrowser.goTo(url)
+		});
+
+		PostWindowURLHandler.register(\https, {|url|
+			var prefix = "http://";
```
JordanHendersonMusic:
```suggestion
```

JordanHendersonMusic (changes requested):

prko:
@JordanHendersonMusic Thanks for the corrections. I don’t understand why I didn’t remove the unnecessary var definitions.

JordanHendersonMusic (approved):

JordanHendersonMusic:
Thanks!

Issue #6654, opened 2025-03-03
Pgbrown step distribution is skewed


HotwheelsSisyphus:
Caveat: My grasp of stats is a bit sketchy, so someone with a firmer footing should definitely double check this one.

But the distribution in its current implementation clearly isn't uniformly logarithmic; the downward steps will be larger always, because `(1 - x).log.abs > (1 + x).log.abs`.

## Environment

* SuperCollider version: 3.14 dev

## Steps to reproduce

Class code in Streams/Pattern.sc:

```supercollider
Pgbrown : Pbrown {
	calcNext { arg cur, step;
		^cur * (1 + step.xrand2)
	}
}
```
Run that function a few times to see how p (current implementation) quickly approaches 0 whereas r does not:

```supercollider
(
var step = 1/4;
n = 99999;
r = p = 1;
n.do({
    //this is calcNext
    p = p * (1+step.xrand2);
    //and this is (I think) what it should be
    r = r * exprand((1+step).reciprocal, 1+step)
    });
r.postln;
p.postln;
)
```

## Expected vs. actual behavior

Expected: "Steps" in either direction should be of same (logarithmic) size. Presumably (I'm not sure here), product of multiplication factors should not show any clear tendencies (like r in the above snippet).

Actual: Downward steps tend to be larger, as a consequence, product of multiplication factors approaches 0 as the number of steps goes to infinity.

## Proposed fix:
```supercollider
Pgbrown : Pbrown {
	calcNext { arg cur, step;
		^cur *  exprand((1+step).reciprocal, 1+step)
        }
}
```
or even better, replace arg step with `ratio` that corresponds to (1+step) to make this clearer.

##Two questions:
a. does the proposed fix check out?
b. should this even be changed, given that there might be pieces/code out there that use Pgbrown in its current form? Or just put a warning in the docs and suggest people use Pbrown.linexp instead?

Can do a PR once I have confirmation on either point.

PR #7636, opened 2026-07-12
Pgbrown: make .calcNext properly geometric

HotwheelsSisyphus:
as discussed in issue #6654. Currently (before this PR) the distribution skews downward, which it shouldn't.
Note that the "wraparound" (that one gets if the return value from `calcNext` is outside the limits `lo, hi`) will still be linear., not geometric. I thought about fixing that too, but since this is handled by the `embedInStream` method, this is more complicated; and likely not worth the hassle; it only happens with unreasonable step sizes anyway.
So I opted for a very minimal fix, easier to review.

## Purpose and Motivation

see #6654

## Types of changes

- Bug fix

### DIFF
--- a/SCClassLibrary/Common/Streams/Patterns.sc
+++ b/SCClassLibrary/Common/Streams/Patterns.sc
@@ -492,7 +492,7 @@ Pbrown : Pattern {
 
 Pgbrown : Pbrown {
 	calcNext { arg cur, step;
-		^cur * (1 + step.xrand2)
+		^cur * exprand((1 + step).reciprocal, 1 + step)
 	}
 }
 


JordanHendersonMusic:
Could you make a quick graph of each distribution? Would be nice to be able to see how big the error was and how yours has fixed it.

HotwheelsSisyphus:
Well, it's the distribution on the step size that is skewed, hence this will only become really apparent on iteration. So even small errors matter. I think I explained this relatively well in the original issue, where there is some demo code. I'll try to whip together some histograms...

HotwheelsSisyphus:
ok here we go (fixed now). The `.ratiomidi` is just to map the (pseudo-) geom distribution back to linear so it's more evident whether it's uniform or not. 

```

[Array.exprand(1e6, 4/5, 5/4).ratiomidi.histo(1e3), Array.rand(1e6, 0.75, 1.25).ratiomidi.histo(1e3)].plot(name: "red: after this PR; blue: before this PR").superpose_(true).plotColor_([Color.red, Color.blue])
Array.rand(1e6, 0.75, 1.25).ratiomidi.mean // around -0.2 
Array.exprand(1e6, 4/5, 5/4).ratiomidi.mean // around 0 (better)


```
image

(Unfortunately the .histo method doesn't relabel the domain; so ignore the x-axis labels).

HotwheelsSisyphus:
bah I used .rand2, should have used .rrand for the demo - disregard previous post please. The following is the fixed one:

ok here we go: The `.ratiomidi` is just to map the (pseudo-) geom distribution back to linear so it's more evident whether it's uniform or not. 

```
[Array.exprand(1e6, 4/5, 5/4).ratiomidi.histo(1e4,-5,5), Array.rand(1e6, 0.75, 1.25).ratiomidi.histo(1e4, -5, 5)]
.plot(name: "green: this PR, red: Before this PR")
.superpose_(true).plotColor_([Color.green, Color.red])

Array.rand(1e6, 0.75, 1.25).ratiomidi.mean // around -0.2 
Array.exprand(1e6, 4/5, 5/4).ratiomidi.mean // around 0 (better)


```
image

(Unfortunately the .histo method doesn't relabel the domain; so ignore the x-axis labels).

JordanHendersonMusic:
Why are the ranges different? green from 1000-10000 but red from 0 -10000

HotwheelsSisyphus:
because, as I mention in the issue, $(1 - x) < \frac{1}{(1 + x)}$. I.e., when x is 0.25, the lower bound for the linear range from which the step size is drawn, [1 - 0.25, 1 + 0.25] is 0.75; for the geometric distribution to be uniform, the limits for the step sizse need to be reciprocals of one another, i.e.[(1/(1 + .25), 1 + .25], which equals [4/5, 5/4] =  [0.8, 1.25]. So the lower bound there is 0.8.

HotwheelsSisyphus:
btw here's a cleaner looking comparison. The top bit is what the distribution should be (i.e., after this pr), the bottom is what it is currently. (They're not random numbers but rather just corresponding Array.geom and .series, so the distributions are clearer at smaller sample sizes).
 Both plots scale geometrically in the x-axis, and the top one is symmetric about 1 and uniform (in that geometric scale), while the bottom one is not uniform. 

image

```
// to make the example:
r = 1e6; 
x = [Array.geom(r+1, 4/5, (5/4).pow(2/r)), Array.series(r+1, 0.75, 0.5/r)]

// the following will only work with my Histogram PRs.
x.histo(1e4, 2/3, 3/2, \exp).plot
```
--------

 To explain why this happens, but with different numbers: 
 The geometric mean "midway" between 4 and 16 is 8 (because 4  * 2= 8  but also 8 * 2 = 16 ). But there's twice as many integers between 8 and 16 than there are between 8 and 4. Hence if you take the series 4 .. 16 and plot it geometrically it will skew right: there are more numbers on the right of 8 than to the left.
 
Now in this particular case that right skewness is offset by the left bound's being in the wrong place, so it actually skews ever so slightly left (as demonstrated in the issue). To stick with the example, this is something like putting the left bound at 0 because it's "the same distance" from 8 as 8 is from 16. The issue is just that this is linear distance, and we want geometric distance.

PR #7649, opened 2026-07-21
help:Documented SortedCollection function callback

suetanvil:
## Purpose and Motivation

I noticed an annoying omission in the documentation for `SortedCollection`.  This fixes that.

## Types of changes

- Documentation

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [X] Updated documentation

## Merge notes

I am not currently equipped to compile SuperCollider, so I was not able to confirm that this doesn't break anything.  If that's a problem, I can get back to you once I've done that. I'm hoping that this is minor enough that you can just run it through CI.

### DIFF
--- a/HelpSource/Classes/SortedList.schelp
+++ b/HelpSource/Classes/SortedList.schelp
@@ -7,6 +7,15 @@ CLASSMETHODS::
 method::new
 Creates a SortedList with the initial capacity given by strong::size:: and a comparison strong::function::. The SortedList can contain equal / identical elements more than once.
 
+The function must accept two arguments and return true if and only if the first argument comes before the second argument in the sort order.  If omitted, defaults to {|a, b| a < b}.  (Note that this is slightly different from what SequentialCollection.sort does by default; it uses the '<=' operator instead.)
+
+code::
+t = SortedList.new(8, {|a,b| a.first < b.first });
+t.add([10, "last"]);
+t.add([1, "first"]);
+t.postln;
+::
+
 INSTANCEMETHODS::
 
 private::indexForInserting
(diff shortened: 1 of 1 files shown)


HelpSource/Classes/SortedList.schelp
```diff
@@ -7,6 +7,15 @@ CLASSMETHODS::
 method::new
 Creates a SortedList with the initial capacity given by strong::size:: and a comparison strong::function::. The SortedList can contain equal / identical elements more than once.
 
+The function must accept two arguments and return true if and only if the first argument comes before the second argument in the sort order.  If omitted, defaults to {|a, b| a < b}.  (Note that this is slightly different from what SequentialCollection.sort does by default; it uses the '<=' operator instead.)
```
HotwheelsSisyphus:
```suggestion
The function must accept two arguments and return true if and only if the first argument comes before the second argument in the sort order.  
If omitted, defaults to code::{|a, b| a < b}::.  
(Note that this is slightly different from what link::Classes/SequentialCollection#-.sort:: does by default, which uses the code::<=:: operator instead.)
```

HotwheelsSisyphus (changes requested):
Thanks for this! Looks good to me, I just added some formatting tags in accordance with our style guidelines for (new contributions to) the help... I think all you'll need to do is to click on "accept suggestion".

The linebreaks won't show in the rendered docs but are handy for revisions/diffs. Obviously most of the extant schelp files don't adhere to all those style guidelines so it's not terribly obvious that any such style guidelines do in fact exist ;)

HelpSource/Classes/SortedList.schelp
suetanvil: Thanks.  I just copied the original style and assumed that was what was expected.  This is much more readable.

suetanvil:
Thanks.  I agree with this change and only kept the original formatting because that seemed like the norm.

HotwheelsSisyphus (approved):
Thanks! I'll merge this now.

PR #7650, opened 2026-07-21
gha windows: use MSVC 2026

dyfer:
## Purpose and Motivation

Should we use MSVC 2026 for the Windows build?

Note: currently this causes the `scsynth_test_run` ctest to timeout. @JordanHendersonMusic do you have any idea about that? I know that ctests on Windows generally need work...

This PR is more of a question than a suggestion, hence I'm submitting it as a draft for now.

If we don't feel this is the right time to do this, I'm happy close this without merging.

## Types of changes

- Maintenance

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -31,15 +31,15 @@ jobs:
             # create-installer: ${{ startsWith(github.ref, 'refs/tags/') }}
             # installer-suffix: "win32-installer"
 
-          - job-name: "64-bit MSVC 2022 with ctest"
+          - job-name: "64-bit MSVC 2026 with ctest"
             fftw-arch: "x64"
             cmake-arch: "x64"
-            os-version: "2022"
+            os-version: "2025-vs2026"
             qt-version: "6.8.3"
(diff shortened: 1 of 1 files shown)


JordanHendersonMusic:
> Should we use MSVC 2026 for the Windows build?

I think we should keep the old one. We can add the new one as well if you like?

> Note: currently this causes the scsynth_test_run ctest to timeout

I have been pretending the windows c tests don't exist.

dyfer:
> We can add the new one as well if you like?

I don't think it's necessary ATM, as it will extend the windows build times due to ctest timeout. 

We can postpone migrating to MSVC 2026 until later.

dyfer:
Hm, maybe we should add msvc 2026 after all (not necessarily as the release build, may be an additional one), as that's would match our support guarantees... but the ctest timeout would need to be somehow addressed.

JordanHendersonMusic:
I don't have a windows dev build setup so can't really look into it. I tried for a while to get good reporting of errors in the crest results... But have up.

PR #7641, opened 2026-07-16
README_WINDOWS.md: update build instructions

dyfer:
## Purpose and Motivation

General update for the somewhat outdated Windows build instructions.

Includes documentation for  #7493.

WIP - open for feedback :)

## Types of changes

- Documentation

## To-do list

- [n/a] Code is tested
- [n/a] All tests are passing
- [x] Updated documentation
- [x] This PR is ready for review

PR #7651, opened 2026-07-21
Download ASIO headers in cmake

dyfer:
## Purpose and Motivation

On Windows, PortAudio needs ASIO headers in order to build the ASIO backend (practically required for low latency operation on Windows). ASIO headers can't be distributed with other projects for licensing reasons.

There a few ways to handle asio sdk download:
- require automatic downloading and extracting to a specific location in our source tree (current solution)
- automatic download in newer version of PortAudio (currently in their `master` branch, but not part of a release yet)
- automatic download in our cmake (this PR)
- switching to PortAudio from vcpkg (provides an option to download asio sdk automatically as a dependency). However, this approach exposes obsolete backends, which we [disabled](https://github.com/supercollider/supercollider/pull/6900) in the SC distribution.

The upcoming version of PortAudio provides automatic downloading of ASIO headers. Assuming we don't want to update our PortAudio code to a non-release version, I propose to add this automatic downloading functionality to our cmake to simplify building on Windows _for the time being_ (until a new version of PortAudio is released).

I was thinking about this for some time, but more recently when working on #7493, which explores the idea of using PortAudio from vcpkg (which does provide an option to automatically download asio sdk). However, I realized that providing our own version of PortAudio is still beneficial, since we can disable obsolete backends, as indicated above.

For reference, here's the [download code in upstream PortAudio](https://github.com/PortAudio/portaudio/blob/master/CMakeLists.txt#L208-L217), which I based the proposed implementation off of.

BTW the headers will only be downloaded once. [Here's a run where I left downloading the headers manually](https://github.com/dyfer/supercollider/actions/runs/29858405234/job/88745697682#step:13:43), and downloading in cmake is properly skipped.

What do you think?

## Types of changes

- Maintenance
- New feature (?)

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation - I'm not updating the documentation in this PR, since I'm reworking Windows docs separately in #7641

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -67,7 +67,6 @@ jobs:
     env:
       BUILD_PATH: ${{ github.workspace }}/build
       INSTALL_PATH: ${{ github.workspace }}/build/Install
-      LIBS_DOWNLOAD_PATH: ${{ github.workspace }}/../3rd-party
       FFTW_INSTALL_DIR: "C:/Program Files/fftw"
       ARTIFACT_FILE: "SuperCollider-${{ inputs.sc-version }}-${{ matrix.artifact-suffix }}"
       INSTALLER_FILE: "SuperCollider-${{ inputs.sc-version }}-${{ matrix.installer-suffix }}"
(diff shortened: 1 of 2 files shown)


JordanHendersonMusic (approved):
looks good

PR #7653, opened 2026-07-23
Histogram class

HotwheelsSisyphus:
I noticed that there are currently two slightly different ways to draw Histograms in sc:
Collection.histo.plot and Collection.plotHisto, where the latter looks a bit better.

So I figured I'd make a Histogram class on the basis of those two that would unify the histogramming in sc.

With this PR, collection.histo.plot and collection.plotHisto both become aliases
of Histogram.new(collection).plot.

The class and its methods were originally mostly reshuffled material from the above collection methods, but I've tinkered with that a fair bit to enable the following:
- exponential (or even arbitrary) scaling of domain via warp
- will also take arrays of arrays as data arg, roughly like .plot does
- some niceties regarding special cases (i.e., when all the data are integers).
- rule-of-thumb auto calculation of number of bins when not provided (data.size.sqrt).

- 
Anyway, looks better and should be a non-breaking change.
also added some tests, but they are outdated. let's see if they still pass ... 

## Types of changes

- New feature

## To-do list

**TODO: determine where in the class library this file should sit. Is `common/collections` ok?**
TODO: schelp update in Collection.schelp for plotHisto and histo methods!
~~TODO: some aliasing that happens with plotmode \bars on the plotter side when there are lots of bins, I had a simple fix for it but I forgot where I saved that.~~

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/Collections/Histogram.sc
+++ b/SCClassLibrary/Common/Collections/Histogram.sc
@@ -8,93 +8,100 @@ Histogram  {
     var <plotter;
 
     *new { | data, n_bins, min, max, x_warp(\lin) |
-		if(data.isEmpty) { MethodError("Histogram.new: data is empty"); ^nil };
-        data = Histogram.prCheckData(data);
+		if(data.isEmpty) { MethodError("Histogram.new: data is empty").throw; };
         ^super.newCopyArgs(data, n_bins, min, max, x_warp)
             .prUpdateHistogram;
 	}
 
-    *prCheckData { |data|
-        if(not(data[0].isCollection)) {
-            data = [data]; // prepare for loop in prMakeFreqsArray
-        } {
-            if(data.collect(_.size).differentiate[1..].sum > 0) {
-                warn("Histogram.new: data arrays not of same sample size; \
-                    comparison may not be sensible")
-            };
-        };
-        ^data
-    }
-
     prUpdateHistogram {
         domainSpec = this.prPrepareDomainSpec;
-        freqsArray = this.prMakeFreqsArray;
-
-        // the data points should align with the middle of each bin,
-        // not at its left boundary.
-        // so we offset the Plotter's domain by half a bin width:
-        // UNLESS the plotmode does this already!? --> Verify
-        domain = domainSpec.map(Array.series(n_bins, 1 / (2  * n_bins), 1/n_bins));
+        if(data.maxDepth == 1) {
+            #freqsArray, outliers = this.prMakeFreqsArray(data);
+        } {
+            #freqsArray, outliers = data.collect (this.prMakeFreqsArray(_)).flop;
+        };
+        domain = domainSpec.map(Array.series(n_bins, 0, 1/(n_bins)));
+        // the values in domain give the lower bounds of each bins.
+        // this means that domain.last will be < max.
         yspec = [0, freqsArray.flat.maxItem * 1.05, \lin].asSpec;
         plotter !? { this.plot };
         ^this
     }
 
     prPrepareDomainSpec {
-        var range, binWidth;
-        n_bins = n_bins ?? { data[0].size.sqrt.ceil };
-        min = min ?? { data[0].minItem };
-		max = max ?? { data[0].maxItem };
+        // flat copy: mem abuse?
+        var range, binWidth, new_n_bins, new_max, dataflat = data.flat;
+        n_bins = n_bins ?? { data.shape.last.sqrt.ceil };
+        min = min ?? { dataflat.minItem };
+		max = max ?? { dataflat.maxItem };
         range = max - min;
-        if(data.every(_.every(_.isInteger)) and: { x_warp === \lin }) {
-            if (n_bins > range) {
-                n_bins = range.asInteger
-            } { // binWidth needs to be integer;
-                // so we round up n_bins
-                binWidth = (range/n_bins).ceil;
-                n_bins = (range/binWidth).ceil.asInteger;
+        if(x_warp === \lin and: { dataflat.every(_.isInteger) } ) {
+            range = range + 1; // this +1 is for the max (e.g.: 4 - 2 = 2, but interval [2, 4] includes 3 integers)
+            if (n_bins >= range) {
+                new_n_bins = range.asInteger;
+            } { // binWidth should be integer.
+                // if not, there may be artifacts where certain bins contain more integers than others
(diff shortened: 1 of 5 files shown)


telephon (commented):
Great! This should be there.
I wonder, is there a reason you don't use the class `Bag`?

HotwheelsSisyphus:
Bag is the multiset class, right? If so, it would be unordered, but histogram bins tend to be ordered... unless that's not what you meant. 

But the real reason I didn't think about what class to use for the bins was that I modeled this on the previous implementation (see the deleted bits in Collection.sc) - where `freqs` is just an Array. Is there some possible performance bonus  if I use some other class, say Bag?

SCClassLibrary/Common/Collections/Histogram.sc
```diff
@@ -0,0 +1,152 @@
...
+
+}
+
++ Collection {
+    // from Collection
+    histo { |n_bins, min, max, x_warp|
+        ^Histogram.new(this, n_bins, min, max, x_warp)
+    }
+    // from PlotView.sc
+    plotHisto { |n_bins, min, max, x_warp|
```
dyfer, line 149: Technically both methods are a breaking change, since they now return a different object than before. I think this may be okay, given the cleaner approach etc, but I wanted to flag this for a possible consideration when reviewing.

One time-honored (albeit inelegant) solution would be to have e.g. `histo` return an array of freqs (as before), as well as something like `histo1` (haha) returning a `Histogram`.

Again, it may be okay as is, but I want us to consider this change.

HotwheelsSisyphus: uff, of course you're technically correct, and I had simply overlooked that... thanks for pointing this out!  I had only looked at the class library and the docs where, iirc, histo _always_ appears as `histo.plot`, so it seemed like nobody cared about the array of freqs. I haven't looked in quarks or on sccode.org... I can do that when I have time.

in my view not having multiple histo-related things is a selling point of this PR, so somewhat predictably I would be very much against adding a `histo1` method or similar, unless it's `histo666420691312` or something similarly edifying.  

If the above (where .histo returns a Histogram object rather than an array) turns out a serious point of contention, I'd rather leave .histo as is then (or change implementation to return Histogram().freqsArray, which should be non-breaking), though I can't say I like that very much...

HotwheelsSisyphus: Although, 

> both methods are a breaking change

I can totally follow wrt `histo`, but with `plotHisto` I'm not so sure I understand. (This is more about what "breaking change" actually means in detail:  and I might not be up to speed in this respect. Is it the change in function signature? The change in defaults?) 

... because what's being returned is a plotter object, as before, and (excepting the silly special case for integers, which in hindsight I might just remove), this should be fully backward compatible _unless_ the user accidentally supplies a fourth argument (which would, before this PR, have been ignored, with this PR, interpreted as a warp).

EDIT: Oh, found another change: The default for n_bins. So yes, that would change behavior when n_bins is not supplied explicitly. If necessary we can change this back to make this less breaking...

dyfer: In my comment I was only referring to what is being returned.

> ... because what's being returned is a plotter object,

Hm, is it? My understanding it that now `plotHisto` returns a [`Histogram`](https://github.com/supercollider/supercollider/pull/7653/changes#diff-196cee51b631cb6298df7d15826dcb72dc924a8f3d7b07e32fd5fdb01a001bfaR150) object, not a [`Plotter`](https://github.com/supercollider/supercollider/pull/7653/changes#diff-e1761c0464556b95db7a72cad794319025b9a0e733a37605cad47e08af32b613L1452).

> histo always appears as histo.plot, so it seemed like nobody cared about the array of freqs

Again, I'm not saying this breaking change is unacceptable, but it _is_ a breaking change and it _will_ break someone's code. We can't assume how users used the method; if they assigned the result to a variable, it will now contain something different. If a certain functionality is there, the assumption is that someone has used it. 

If we want to avoid breaking people's code (albeit breaking may be minor), the old methods should return the same object/data as before. While `histo1` is a  bit ugly, a fully spelled name could be an alternative maybe? I.e. `.histo` returning an array of freqs and `.histogram` returning a Histogram object.

Adding a new argument is absolutely fine, but the existing args should not be changed (or renamed, ideally, so `.histo`'s and `.plotHisto`'s first arg should still be `steps`).

HotwheelsSisyphus: Here's what I see in line 150 (plotHisto implementation):

```
        ^Histogram.new(this, n_bins, min, max, x_warp).plot;
```
I.e., it returns whatever Histogram.plot returns... and that's a plotter.

HotwheelsSisyphus: I don't feel strongly about the variable names. Ok with changing that if necessary. Although "steps" to me doesn't seem super intuitive for a histogram...

Not terribly important in this case, but occasionally there are strange naming choices where a change might be worthwhile...  

So just taking this an example, could a viable option in such cases be to include the old name as a kwarg?
E.g. in the present case,
```
histo {  | n_bins, min, max ,x_warp, steps |
n_bins = n_bins ?? steps ?? { some default }
// etc.
}
``` 
such that  these calls would work out equivalently:
```
histo(100);
histo(steps: 100);
histo(n_bins: 100)
```

dyfer: > I.e., it returns whatever Histogram.plot returns... and that's a plotter.

Ah, my bad! I missed the .plot at the end. If that returns a Plotter, then that's of course fine.

As for the argument names, I'm not saying they are good, but only that we should consider backwards compatibility.

> So just taking this an example, could a viable option in such cases be to include the old name as a kwarg?

That's an interesting approach, I haven't thought of that. My intuition says that it makes the function signature unnecessarily messy, but yes it does provide backwards compatibility while introducing a better name for the argument. I wonder if @telephon has an opinion here? I trust his judgement more on language syntax intricacies.

<details><summary>The old arg name could even be deprecated?</summary>

If we wanted to discourage the use of the old argument, we could post a deprecation warning when `steps` arg is used. That could provide a path to ultimately remove it... though I'm not sure if this is worth all the hassle.

</details>

Back to return values, ultimately I'd lean towards just keeping the old return value for `.histo` (an array of freqs). If someone wants to get a Histogram as a result, then they should just do `Histogram(<data>)`. I.e. I don't think adding a second method that would return a Histogram is needed. What would you think about that approach?

PR #7654, opened 2026-07-25
Add warning against bundling external Extensions inside Quarks

prko:
## Purpose and Motivation

Recently, I encountered a Quark that included an external Extension both as a
declared dependency and also bundled the Extension inside its own directory:
https://scsynth.org/t/performative-a-toolkit-for-creating-performance-instruments-and-designing-parameters-in-supercollider/13461/11?u=prko

This caused unnecessary errors, forcing users to remove the Extension from the
SC-IDE Preferences path and manually delete the folder in Finder or any equivalent
file manager.

This PR adds a warning to address this problematic practice among some Quark authors.

## Types of changes

- Documentation

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/HelpSource/Guides/UsingQuarks.schelp
+++ b/HelpSource/Guides/UsingQuarks.schelp
@@ -202,7 +202,7 @@ Quarks.addFolder("~/supercollider/quarks");
 Now you can (un)install your own packages from your local folder.
 
 warning::
-If your Quark relies on any extension that is not a Quark, list it as a dependency in your README instead of relying on the Quarks to install that extension.
+If your Quark relies on any extension that is not a Quark, list it as a dependency in your README instead of relying on the Quarks system to install that extension.
 ::
 
 Managing your code with git is optional, but you should consider using it early on. Even if you do not intend to share your code with anyone else, git provides a backup system and a time machine if you break something.
--- a/HelpSource/Guides/UsingQuarks.schelp
+++ b/HelpSource/Guides/UsingQuarks.schelp
@@ -211,11 +211,7 @@ Quarks.addFolder("~/supercollider/quarks");
 Now you can (un)install your own packages from your local folder.
 
 warning::
-Do not include plugins, extensions, or classes distributed by others directly inside
-your Quark folder. SuperCollider loads Extensions globally, so placing an external
-Extension inside a Quark can cause duplicate class errors for users who have already
-installed that Extension separately, according to that extension’s installation
-instructions.
+Do not include plugins, extensions, or classes distributed by others directly inside your Quark folder. SuperCollider loads Extensions globally, so placing an external Extension inside a Quark can cause duplicate class errors for users who have already installed that Extension separately, according to that extension’s installation instructions.
 
 Example:
 
@@ -225,8 +221,7 @@ ERROR: duplicate Class found: 'SomeClassName'
 ~/Library/Application Support/SuperCollider/downloaded-quarks/YourQuark/classes/...
 ::
 
-If your Quark depends on an external Extension, declare it as a dependency in your
-README file or in the Quark metadata, rather than bundling the Extension itself.
+If your Quark depends on an external Extension, declare it as a dependency in your README file or in the Quark metadata, rather than bundling the Extension itself.
 ::
 
 Managing your code with git is optional, but you should consider using it early on. Even if you do not intend to share your code with anyone else, git provides a backup system and a time machine if you break something.
--- a/HelpSource/Guides/UsingQuarks.schelp
+++ b/HelpSource/Guides/UsingQuarks.schelp
@@ -215,7 +215,6 @@ ERROR: duplicate Class found: 'SomeClassName'
 
 If your Quark depends on an external Extension, declare it as a dependency in your
 README file or in the Quark metadata, rather than bundling the Extension itself.
-
 ::
 
 Managing your code with git is optional, but you should consider using it early on. Even if you do not intend to share your code with anyone else, git provides a backup system and a time machine if you break something.
--- a/HelpSource/Guides/UsingQuarks.schelp
+++ b/HelpSource/Guides/UsingQuarks.schelp
@@ -28,6 +28,18 @@ list::
 
 section:: Installing/uninstalling/updating quarks
 
+warning::
+soft::&hsp;::Some Quarks are not actively maintained and may cause errors during compilation or usage. The most frequently encountered errors are:
+
+list::
+## teletype::ERROR: duplicate Class found::
+## teletype::ERROR: Found duplicate instance variable name::
+::
+
+If you encounter such issues, you can report the problem by creating an issue in the Quark's repository. More proactive users are encouraged to fork the repository, resolve the issue, and open a Pull Request.
+
+In these situations, rather than relying solely on the built-in documentation, you may need to independently troubleshoot the problem. We recommend actively seeking help on link::https://scsynth.org:: or other social network services.
+::
 
 subsection::using the GUI
 
--- a/HelpSource/Guides/UsingQuarks.schelp
+++ b/HelpSource/Guides/UsingQuarks.schelp
@@ -29,7 +29,7 @@ list::
 section:: Installing/uninstalling/updating quarks
 
 warning::
-soft::&hsp;::Some Quarks are not actively maintained and may cause errors during compilation or usage. The most frequently encountered errors are:
+Some Quarks are not actively maintained and may cause errors during compilation or usage. The most frequently encountered errors are:
 
 list::
 ## teletype::ERROR: duplicate Class found::
--- a/HelpSource/Guides/UsingQuarks.schelp
+++ b/HelpSource/Guides/UsingQuarks.schelp
@@ -29,12 +29,7 @@ list::
 section:: Installing/uninstalling/updating quarks
 
 warning::
-Some Quarks are not actively maintained and may cause errors during compilation or usage. The most frequently encountered errors are:
-
-list::
-## teletype::ERROR: duplicate Class found::
-## teletype::ERROR: Found duplicate instance variable name::
-::
+Some Quarks are not actively maintained and may cause errors during compilation or usage.
 
 If you encounter such issues, you can report the problem by creating an issue in the Quark's repository. More proactive users are encouraged to fork the repository, resolve the issue, and open a Pull Request.
 
--- a/HelpSource/Guides/UsingQuarks.schelp
+++ b/HelpSource/Guides/UsingQuarks.schelp
@@ -198,6 +198,26 @@ Quarks.addFolder("~/supercollider/quarks");
 
 Now you can (un)install your own packages from your local folder.
 
+warning::
+Do not include plugins, extensions, or classes distributed by others directly inside
+your Quark folder. SuperCollider loads Extensions globally, so placing an external
+Extension inside a Quark can cause duplicate class errors for users who have already
+installed that Extension separately, according to that extension’s installation
+instructions.
+
+Example:
+
+teletype::
+ERROR: duplicate Class found: 'SomeClassName'
+~/Library/Application Support/SuperCollider/Extensions/SomeExtension/classes/...
+~/Library/Application Support/SuperCollider/downloaded-quarks/YourQuark/classes/...
+::
+
+If your Quark depends on an external Extension, declare it as a dependency in your
+README file or in the Quark metadata, rather than bundling the Extension itself.
+
+::
+
 Managing your code with git is optional, but you should consider using it early on. Even if you do not intend to share your code with anyone else, git provides a backup system and a time machine if you break something.
 
 code::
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks for flagging this, @prko.

While these may be a correct warnings, I'm not sure if this is worth the addition, as it creates more text to process by the users. I don't think that bundling other quarks within quarks is a widespread problem - I haven't experienced it myself.

Additionally you indicate that `The most frequently encountered errors are: (...) ## teletype::ERROR: duplicate Class found::`. IMO that's an arbitrary evaluation - the duplicate class is most common when mixing extensions installed by Quarks system with manual installation, not strictly from quarks install itself.

I think that the addition of `Some Quarks are not actively maintained and may cause errors during compilation or usage.` is worthwhile, but I'm not convinced the extra text beyond that is very useful - and again, it increases the amount of text that users need to process.

That's just my personal opinion :)

prko:
@dyfer Thanks, I removed the part you mentioned. 

Please let me know if there are more parts to delete.

>I don't think that bundling other quarks within quarks is a widespread problem - I haven't experienced it myself.

In the link in the opening post of this PR, the Quark author wrote an extension as a dependency. The author seems not to intend to change it based on his answer.
Before this experience, I had never thought of opening this PR. (I first experienced that a Quark installs a non-Quark external within downloaded-quarks folder.)

dyfer:
> @dyfer Thanks, I removed the part you mentioned.
> 
> Please let me know if there are more parts to delete.

I _personally_ would _only_ keep `Some Quarks are not actively maintained and may cause errors during compilation or usage.` and remove _all_ other text. 

The sentence `If you encounter such issues, you can report the problem by creating an issue in the Quark's repository. More proactive users are encouraged to fork the repository, resolve the issue, and open a Pull Request.` could maybe stay as well, but I think that for those with no experience of contributing this isn't enough to get them involved, but for those who already know, it is redundant. Again, my personal opinion.


> The author seems not to intend to change it based on his answer. 

I'm not sure if that's correct, or at least it's not about the author I would think of.

IMO the issue is the manual installation instructions in fosc, rather than the dependency indication in the Monolith quark. There's no need to install extensions manually when they can be installed with `Quarks.install("url")`. The only thing missing is a quark file and updating readme in fosc. Are you planning to open an issue or a PR in the fosc repository to address this?

---

Still, adding multiple sentences to the documentation distributed with SC just because _one_ quark causes issues doesn't seem justified to me. A general warning is worth adding - a similar sentence was one of the first things @joshpar said in the SuperCollider class he taught when talking about quarks :) Let's save everyone's time going into discussing this and focus on a solution rather than increasing documentation with explanation of corner cases.

prko:
Is this now ok?

--- 

>Are you planning to open an issue or a PR in the fosc repository to address this?

No, I’ve already submitted several PRs related to FOSC, and I’m currently waiting for the authors’ responses.
Furthermore, this problem stems from the author’s choice to make SC install an extension—rather than a Quark—into the `downloaded-quarks` folder.
Consequently, the users who installed the Quark are the ones who suffer from this mistake.

dyfer:
> Is this now ok?

There's still additional text in lines 204+...

> Furthermore, this problem stems from the author’s choice to make SC install an extension—rather than a Quark—into the `downloaded-quarks` folder. 

Was there any indication or discussion that manual installation is an intentional decision?

prko:
> There's still additional text in lines 204+...

Those lines are the core of this PR. I’ve shortened the text.
Is it acceptable now?

---

>Furthermore, this problem stems from the author’s choice to make SC install an extension—rather than a Quark—into the downloaded-quarks folder.
>
>Was there any indication or discussion that manual installation is an intentional decision?

It’s unclear what you mean by “manual installation.” Do you mean manual installation of FOSC or of Performative?

Here is my understanding:

1. FOSC was released several years ago as an Extension (not as a Quark), so users are expected to install it manually if they want to use it.
2. Recently, **Performative** was announced as a Quark. When installing Performative, SC installs **FOSC** into `Platform.userAppSupportDir +/+ "SuperCollider/downloaded-quarks/Fosc"`. This results in **duplicate class errors**.
4. I reported this issue here:  
   https://scsynth.org/t/performative-a-toolkit-for-creating-performance-instruments-and-designing-parameters-in-supercollider/13461/5?u=prko  
   The author’s response is here:  
   https://scsynth.org/t/performative-a-toolkit-for-creating-performance-instruments-and-designing-parameters-in-supercollider/13461/9?u=prko  
   He does not seem to be addressing the issue.
5. Further discussion can be found here:  
   https://scsynth.org/t/quark-paths-performative-and-fosc/13493

dyfer:
> > There's still additional text in lines 204+...
> 
> Those lines are the core of this PR. I’ve shortened the text. Is it acceptable now?

Yes, it's fine for me. Thanks!

> 
> > Furthermore, this problem stems from the author’s choice to make SC install an extension—rather than a Quark—into the downloaded-quarks folder.
> > Was there any indication or discussion that manual installation is an intentional decision?
> 
> It’s unclear what you mean by “manual installation.” Do you mean manual installation of FOSC or of Performative?

By manual installation I mean "download an extension and manually place it in the Extensions folder".

For the purpose of this discussion, I consider `Quarks.install("url")` as the recommended (non-manual) way of installing extensions.

> 
> Here is my understanding:
> 
> 1. FOSC was released several years ago as an Extension (not as a Quark), so users are expected to install it manually if they want to use it.

It should just be a quark. It does NOT have to be listed in the quarks directory though. Any git repo can be installed with `Quarks.install("url")`, but the repo should include a quark file.

dyfer:
...and I filed a PR for Fosc: https://github.com/n-armstrong/fosc/pull/13

prko:
>...and I filed a PR for Fosc: https://github.com/n-armstrong/fosc/pull/13

Thanks!

---

My assumption was that developers creating new Quarks or extensions would respect the FOSC author’s original choice, given that FOSC was already published. Ideally, the author of Performative would have resolved this issue before releasing it. As a regular user, I expect this kind of consideration from developers, even though SC is an open‑source ecosystem.

dyfer:
> My assumption was that developers creating new Quarks or extensions would respect the FOSC author’s original choice, given that FOSC was already published. Ideally, the author of Performative would have resolved this issue before releasing it. As a regular user, I expect this kind of consideration from developers, even though SC is an open‑source ecosystem.

I understand that this creates friction, but please be reasonable with expectations towards developers of small-scale / personal extensions. There's a balance between user-friendliness and developer's resources.

IMO manual installation is very inconvenient, so I understand opting for installation using the Quarks system instead. But I also understand your point about respecting original author intention (though I personally think that manual installation by copying the files should be discouraged).

HelpSource/Guides/UsingQuarks.schelp
```diff
@@ -198,6 +201,10 @@ Quarks.addFolder("~/supercollider/quarks");
 
 Now you can (un)install your own packages from your local folder.
 
+warning::
+If your Quark relies on any extension that is not a Quark, list it as a dependency in your README instead of relying on the Quarks system to install that extension.
```
JordanHendersonMusic, line 205: What kind of 'extension' can a quark rely on that isn't a quark?

prko: Well, any kind?

I opened this because of `Performative`, which relies on `FOSC` extension.
@jamshark70 suggested a solution for this in 
https://scsynth.org/t/quarks-and-extensions/13493/15?u=prko

I could update this PR according to @jamshark70  's suggestion and @dyfer 's suggestion...

JordanHendersonMusic:
(in response to the comment I just made about what is an extensions that isn't a quark)... Oh I see.

Yeah, I think this is redundant and should be removed. Fosc should be a quark - that's the 'package' manger we have. I'd rather see effort put into addressing why the author of fosc didn't make this a quark to begin with, and fixing what ever issue they encountered that prevented them from doing so. Creating quarks should be easy-peasy.

HelpSource/Guides/UsingQuarks.schelp
dyfer: The underlying issue is that extensions can be installed in multiple ways:
- as Quarks
  - an edge case of this is a git repo installed as a quark (`Quarks.install("<url>")`), when it doesn't have a quark file and does not provide any metadata; that's the case of how `Fosc` has been pulled in as a dependency
- by manually copying files to the Extensions folder
- by adding a directory to the `sclang_conf.yaml` file (in IDE preferences, typically)

All these are independent and don't have any duplicate class/dependency checking between each other, until sclang tries to compile the library.

That could be spelled out in the docs better, possible? I didn't check whether it is or not, currently.

---
As for this change, I also think that `Fosc` dependency is a singular instance and does not warrant adding this warning to the docs.

prko: Well, before the recent 3.15.0-dev, I installed the following extensions:

- Extensions/DanceGestureAnalyser-master/
- Extensions/DynGen/
- Extensions/FluidCorpusManipulation/
- Extensions/fosc/
- Extensions/interpolation.sc
- Extensions/mi-UGens/
- Extensions/nUPIC_26_PUBLIC-26_1.0.0/
- Extensions/OversamplingOscillators/
- Extensions/PortedPlugins/
- Extensions/sc_faust/
- Extensions/SC3plugins/
- Extensions/SuperClean/
- Extensions/TXModular_089_sources/
- Extensions/VSTPlugin/
- Extensions/World_World/

I think any of these could become dependencies of the Quark package in the future. (I will not write any Quark relying on one of these.)

dyfer: Sure. For example sc3plugins is a dependency of the atk quark. BUT the issue is not that the problem can't exist - it's more like it's an edge case and thus not worth (IMO) bloating the docs. Most of the extensions you listed are simply not installable as `Quarks.install("url")` so it's not possible to list them as automatically installed depdendency and have the same issue as with fosc.

The point is that this is an edge case and thus not worth the added text.

prko: I don’t agree with your viewpoint, and I actually think this kind of stance is one of the reasons why newcomers and non‑advanced users feel discouraged from using SuperCollider. However, it seems that I cannot change this, so I will close the PR.

prko:
I close this.

JordanHendersonMusic:
@prko

> I don’t agree with your viewpoint, and I actually think this kind of stance is one of the reasons why newcomers and non‑advanced users feel discouraged from using SuperCollider.

If I may share my view in why I think this friction keeps reoccurring...

I think you are working on trying to fix really important issues, but occasionally reaching for the wrong tool (documentation). 

The best solution is one that makes the issue impossible to occur again. Documentation is always the worst solution to a problem because it requires the users read them and that we maintain it... Yet it is sometimes necessary when all else fails. 

There are legacy cases of this issue occurring, but if focs is relatively modern, it should be a quark. If we can find out why it isn't, fix that problem, then we can solve this for ever - with out requiring the user read a single line. 

To be clear, I'm saying a well designed system needs minimal documentation because it is impossible to use incorrectly. If we can make a problem impossible, we should.

The other serious problem with documentation is that when a new users does something wrong, the answer we give is 'read the docs, duh'. That's not a nice basis for a community, and requires new users read a increasingly large tome before doing anything, which is impossible, so they will always feel alienated, and hence come to view SC as arcane.

Unfortunately it already is arcane, but let's try and reverse that together?

prko:
>The best solution is one that makes the issue impossible to occur again. 

That sounds ideal, and I completely agree.

>Documentation is always the worst solution to a problem because it requires the users read them and that we maintain it... 
>The other serious problem with documentation is that when a new users does something wrong, the answer we give is 'read the docs, duh'. That's not a nice basis for a community, and requires new users read a increasingly large tome before doing anything, which is impossible, so they will always feel alienated, and hence come to view SC as arcane.

To be honest, I’ve been thinking that SC’s documentation could be improved, especially when compared with Max.

>Unfortunately it already is arcane, but let's try and reverse that together?

My abilities don’t cover as wide a scope as those of other developers, but I can contribute in the ways I’m able, such as:
- cross‑platform testing
- improving documentation
- writing small SC code examples (scd and sc)

In any case, I’d like to help where I can.
If each of us contributes a little, I believe SC can become a much more welcoming environment.

JordanHendersonMusic:
> To be honest, I’ve been thinking that SC’s documentation could be improved, especially when compared with Max.

Absolutely, I don't think any one would disagree with that!

But the best code/system/ecosystem is one that doesn't need documentation because it's so obvious and impossible to use incorrectly... Or at least really hard to use incorrectly.

We can improve the docs by changing the code to make the problem disappear, and then *deleting* the documentation.

I think we should always try this first because the docs are already huge, far bigger than what we can reasonably maintain. 

One thing that helps combat the difficulties maintaining the docs would be a better search system and adding more links between pages --- 
I've seen you've done some work on this in the past!

It's also worth noting that max/pd has a very different type system to us. Their types are the cables and their objects functions, so max has yellow stripey, grey (which is a summed type of only a few others), blue stripey, and whatever colour the graphics tensor one is. This reduced type system makes documentation a lot easier, but their language a lot less expressive than ours.

PR #7656, opened 2026-07-25
Plot - centered stepwise plotmodes on geometric domainSpecs

HotwheelsSisyphus:
Noticed this when I did my histogram PR: 
The current implementation of the stepwise-centered plot modes introduced by @mtmccrea only works when the domainSpec is linear; for geometric (i.e. /exp) domainSpecs, the x - location of the data points is off,  because of the  halfBinWidthPx calculation  and in the case of the \bars plotmode, this results in bars that are too thin, too thick, what have you. 

The reason is a muddling of the data's x coordinate (e.g., the independent variable x of a function y(x)) and the plot's x coordinate (the horizontal component of the input to Pen, in pixels). But because the plot's x coordinate is linear, this only really becomes apparent when the data's x coordinate is not.

The variable previously named `domainPad` (which I rename to plotXPad) actually is more easily specified with respect _not_ to the data's x coordinate (`domain`, `domainSpec`) but with respect to the plot's x coordinate, which is equivalent to `Array.interpolation(domain.size) * plotBounds.width`, aka `(0..size-1)/(size - 1) * plotBounds.width`, i.e. some fraction of the plotBounds.

So the thing that needed to be changed is this:
```
domainPad = { plotter.domain.differentiate.drop(1).minItem / 2 } 

```
(i.e., half the difference between the first two items)
which needed to be divided by domainRange further down to express a fraction of
the plotBounds:
```
halfBinWidthPx = (domainPad / domainRange) * plotBounds.width
```

the problem is the detour via domain, which only works when said domain is linear.
turns out the detour is unnecessary, as we can express the fraction without recourse to the domain:

```
plotXPad = (value.size).reciprocal/2
halfBinWidthPx = plotXPad * plotBounds.width
```
this also means that the domainRange variable (not used anywhere else) becomes unnecessary, and that it becomes easier to deal with geometric domainSpecs.

additional changes reflect the fact that domainXPad is now a fraction in [0,1] rather than a magnitude in the domain, which permitted some simplifications.

What _still_ doesn't work (and is not worth fixing imo, though I've thought about it) is when domain values are irregular with respect to some warp, because then a fixed halfBinWidthPx won't do. As an example: If domainspec is linear but there should be bars at x = 1, 2, 4, 8, 16... then bin width is itself dependent on x position. This is doable in principle but I doubt anyone cares.

- Tested with the "centered stepwise" examples from Plotter.schelp. 

## Some pictures:
```
x = Array.geom(25, 1, 25.pow(1/24))
y = 2 * x
y.plot.plotMode_(\bars).domainSpecs_([[1,25,\exp]])
```

Before:
image

After: 
image

### DIFF
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -12,7 +12,7 @@ Plot {
 	var <>labelMargin = 2;  // margin around tick or axis labels
 	var <>borderMargin = 3; // margin separating the edge of the view from its inner elements
 	var <>hideLabelsHeightRatio = 1.2, <>hideLabelsWidthRatio = 1.5; // hide labels below plot:labels ratio
-	var domainPad = 0.0, halfBinWidthPx = 0.0; // for steplike data width
+	var plotXPad = 0.0, halfBinWidthPx = 0.0; // for steplike data width
 	var valueCache, resolution;
 
 	*initClass {
@@ -225,22 +225,20 @@ Plot {
 		domainSpec = sp;
 
 		// update private domain specs to give visual padding to the beginning and end of the data
-		if(value.size > 1) {
-			domainPad = if(this.hasCenteredSteplikeDisplay) {
-				if(plotter.domain.isNil)
-				{ 0.5 } // domain step defaults to 1
-				{ plotter.domain.differentiate.drop(1).minItem / 2 }
+		// this padding is a fraction of the plotBounds (e.g. 1/200 for data of size 100)
+		plotXPad = if(this.hasCenteredSteplikeDisplay) {
+				(value.size).reciprocal/2
 			} {
 				0.0
 			};
-
+		if(value.size > 1) {
 			prDomainSpec = domainSpec.copy;
 			prDomainSpec
-			.minval_(domainSpec.minval - domainPad)
-			.maxval_(domainSpec.maxval + domainPad);
+				.minval_(domainSpec.warp.map(0 - plotXPad))
+				.maxval_(domainSpec.warp.map(1 + plotXPad));
+			// warp.map can take args outside [0, 1], which spec.map would clip
 		} {
 			prDomainSpec = domainSpec;
-			domainPad = if(this.hasCenteredSteplikeDisplay) { 0.5 } { 0.0 };
 		};
 
 		if(gridOnX and: { prDomainSpec.notNil }) {
@@ -398,7 +396,7 @@ Plot {
 	}
 
 	domainCoordinates { |size|
-		var vals, domainRange;
+		var vals; // linear in [0, 1]
 
 		if (plotter.domain.notNil) {
 			vals = prDomainSpec.unmap(plotter.domain);
@@ -406,15 +404,11 @@ Plot {
 			if (size == 1 or: { domainSpec.range == 0.0 }) {
 				vals = 0.5.dup(size) // put the values in the middle of the plot
 			} {
-				vals = prDomainSpec.unmap(
-					Array.interpolation(size, domainSpec.minval, domainSpec.maxval)
-				);
+				vals = Array.interpolation(size, plotXPad, 1 - plotXPad)
 			}
 		};
 
-		domainRange = prDomainSpec.range;
-		if(domainRange == 0) { domainRange = 1 };
-		halfBinWidthPx = domainPad / domainRange * plotBounds.width;
+		halfBinWidthPx = plotXPad * plotBounds.width;
 
 		^plotBounds.left + (vals * plotBounds.width);
 	}
@@ -587,7 +581,7 @@ Plot {
 		var barWidth = halfBinWidthPx * 2 - (2 * gap);
 		var centery = 0.linlin(this.spec.minval, this.spec.maxval, plotBounds.bottom, plotBounds.top, clip: nil);
 
-		Pen.smoothing_(false);
+		Pen.smoothing_(true);
 
 		y.size.do { |i|
 			var p = x[i] @ y[i];
--- a/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
+++ b/SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
@@ -576,12 +576,13 @@ Plot {
 	}
 
 	bars { |x, y|
-		var gap = halfBinWidthPx * 0.1;
+		// if gaps would be < 1 pixel, don't draw gaps at all.
+		var gap = if(halfBinWidthPx >= 5) { halfBinWidthPx * 0.1 } { 0 };
 		var xOffset = halfBinWidthPx.neg + gap;
 		var barWidth = halfBinWidthPx * 2 - (2 * gap);
 		var centery = 0.linlin(this.spec.minval, this.spec.maxval, plotBounds.bottom, plotBounds.top, clip: nil);
 
-		Pen.smoothing_(true);
+		Pen.smoothing_(false);
 
 		y.size.do { |i|
 			var p = x[i] @ y[i];


prko:
The following plotModes exhibit the behaviour described in this PR:

* levelsCentered
* dlevelsCentered
* plevelsCentered
* bars

With this PR applied, these modes also alter their x‑ruler behaviour.

As a result, all plotModes now share the same x‑ruler, except for the four modes listed above, which adjust their width. This makes the overall behaviour more consistent.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -587,7 +581,7 @@ Plot {
 		var barWidth = halfBinWidthPx * 2 - (2 * gap);
 		var centery = 0.linlin(this.spec.minval, this.spec.maxval, plotBounds.bottom, plotBounds.top, clip: nil);
 
-		Pen.smoothing_(false);
```
dyfer, line 590: Is this change intentional? I _may_ have performance implications when e.g. animating the plot, though I haven't tested that.

HotwheelsSisyphus: it is intentional but I don't mind changing it back if you think it's a bad idea. 
I've noticed that with a larger number of bars, the smoothing gets rid of  an aliasing effect that is there otherwise. Also I figured since whatever the smoothing algorithm is, it's probably the kind of parallel linear algebra that modern hardware is supposedly pretty good at anyway because we're supposed to be using LLMs all the time... no idea if that reasoning actually makes sense to people who understand this stuff (I don't really).

dyfer: I'm not sure if this is a bad idea - I don't have enough insight. Let's leave it in for now, unless others have a different preference.

HotwheelsSisyphus: fwiw, I just tried another approach: turning the smoothing off but not drawing the gap between the `\bars` if said gap would be less than a pixel (or half a pixel) wide. (I.e., `if(gap < x) { gap = 0 }`, `x` to be determined. That seems to do the trick and presumably causes less overhead than the smoothing, while simultaneously leaving the bars looking neat when there is enough space for them, and not aliasing when there is not.

[Screencast_20260728_194258.webm](https://github.com/user-attachments/assets/dfab468b-d021-47b6-a024-070df8afce58)

dyfer: Oh, I like that!

HotwheelsSisyphus: alright then, I just pushed that change!

mtmccrea:
I think there is some confusion about the functionality of `domainSpec`. It just controls the mapping of the domain values (x-values of the data). Unless set explicitly, the domain values default to [0, .. data.size-1]. So if you remap them visually with an exponential `domainSpec`, you would expect a nonlinear distribution of data on the x-axis.

So IIUC, for a correct test of what you're going for you'd need to set the `domain` explicitly:

```supercollider

x = Array.geom(25, 1, 25.pow(1/24))
y = 2 * x
p = y.plot.plotMode_(\bars).domainSpecs_([[1,25,\exp]])
p.domain = x

```

If you run that you'll see what I think you expect to be a proper distribution of data.

You'll also see that we lose the inter-bar gap and plot padding at the left and right of the plot. Which could be looked at more closely.

I can't follow up much right now, but hopefully this understanding of the domain redirects any changes you might want to make.

I'd also say that an exponential spec on the domain for these plot modes is likely to be an edge case, so may not be worth a significant revision of the drawing logic, and if that's the case, we could at least make note of any unsupported or unexpected behavior in the docs of those plot modes.

HotwheelsSisyphus:
> likely to be an edge case, so may not be worth a significant revision of the drawing logic
 
 I disagree.  spectrograms, for instance? Hardly an edge case in computer music. Yes you can map cpsmidi but you don't always want to. Granted, it's probably more common if you deal with distributions and histograms rather than functions, but still I see no reason not to include the functionality. (Otoh, perfectly understandable if you don't have time to review now; urgent this is not).
 
I also don't think this is "a significant revision" of the drawing logic. It's essentially a refactor with some tweaks. On the other hand, if it _is_ "significant" it seems odd to then _also_ argue that "naah we don't need this, you can already do this if you also set `domain`. "

To address that latter argument: Yeah sure, but if I want to quickly plot something, I don't want to have to _both_ supply domainSpec _and_ domain. You can (sort of) build the domain given a domainSpec, but not the other way around (unless you want the plotter to also do curve-fitting... ); so in my mind this is a quality of life improvement. 

to repeat a bit what I said in my initial pr description, I think the relevant confusion is not about what the domain functionality does as opposed to the domainSpec, but rather that there is a muddled distinction in the (pre-PR) code between _data_ coordinates (which includes both x's and y's and can be in whatever scale and unit) and _display_ coordinates (i.e., stuff on the screen, in pixels, also x and y; this is where the domainPad/plotXpad should be, because it is about the width of things on the _screen_, not about something in the data) . To understand where I'm coming from, consider how matplotlib solves this - they even have a bunch of intermediate coordinate systems between data coords and display: 

https://matplotlib.org/stable/users/explain/artists/transforms_tutorial.html

mtmccrea:
Just circling back briefly, sorry I can't do an in-depth look atm.

> spectrograms, for instance?

Spectrograms, being fft-based are a linear decomposition of the spectrum. This is why the bin width changes with frequency on many fft-based logarithmic plots (see FreqScope). That would require a variable bar width, which isn't currently supported, but could be.

Your before/after plots in the description show two different data sets – identical (y-)values but different x-values. So it seems this PR implicitly changes x-values, which should be independent of the domain spec/range/warp.

I agree that it could be beneficial to calculate the bar width, gap, and pad based on unmapped values (range [0, 1]. But I see atm you're calculating the padding based on values.size. Ideally padding would be based on the x-value of the first and last data points and the domain range. I realize it's currently only based on the x-distance between the first and second data point, which was added to get the most obvious use case of near-uniform distros across the x-axis to work.

It would be helpful if we could frame what needs to be done in terms of a missing feature or a bug, with desired behavior and a code example or two.

For example, you may be proposing that the automatic domain distribution of data follows the spec warping when the domain has not otherwise been explicitly set by the user.

Thanks for working through this!

HotwheelsSisyphus:
whoops, of course you're right about the DFT...
My actual motivation for looking into this was wanting to show a histogram of a geometrically uniform distribution (e.g., exprand) without having to do a .cpsmidi on the data. (#7636). Maybe that is indeed slightly more of an edge case ... but still, it doesn't seem outlandish to me to want this.

> For example, you may be proposing that the automatic domain distribution of data follows the spec warping when the domain has not otherwise been explicitly set by the user.

exactly. And this happens to be the behavior with linear specs already... example:

```
// exciting graph of a sine
y = { |x| sin(x/500 * 2pi) }.dup(500);
y.plot.domainSpecs_([[0, 2]])
```
This is probably what led me to assume that, as long as we want x-coords that are equidistant, one needn't specify `domain` separately. (As distinct from cases like, idk, `domain = [2, 7, 7.3, 27]` or whatever). I also rarely ever _actually_ graph things that simply have 0..n as the domain, and (as you can see above), that has been working just fine using domainSpecs. Hence my expectation that the same work for \exp warps.

---------------------------

At any rate my understanding is that the PR does not change "values" (as in, `.domain` ) but simply the location where these values are being displayed, and this should happen only if the `domain` is nil. I might however want to test this more thoroughly.

> I see atm you're calculating the padding based on values.size. Ideally padding would be based on the x-value of the first and last data points and the domain range. 

I was extrapolating here from the linear case, and the prior state of the code. In the linear case (and when domainSpec.range simply equal to the first and last element in domain), we _can_ use `size` to get the same results as before, and it seemed more elegant/legibile to me. 

An issue I notice now with this is with `domain`s that are smaller than the `domainSpec.range`, (e..g domain [5..20] and domainSpec [1, 24])... I'll need to think about this.

HotwheelsSisyphus:
update with some heat-addled thoughts  re this part:

> domains that are smaller than the domainSpec.range, (e..g domain [5..20] and domainSpec [1, 24])

(in fact, in these cases we simply don't _need_ to pad!... but anyway:)
fairly straightforward but less concise:
```
plotXPad = domainSpec.unmap(domain.wrapAt([0,1,-2,-1]))
    .differentiate[1,3]/2
```

raises the question though under what conditions one actually does need a separate right and left pad.  Because if the bins are _not_ meant to be of equal size on the screen, the approach wit halfbinWidthPx won't work either way.

-----------------
so I also looked into the other thing where one does not assume the bins to be of equal horizontal size...the most straightforward way to implement this is to treat the x's as bin boundaries, not centers;  the question that this brings up is that one effectively needs one more x data point than y data points... if there are n y-values, there will be n+1 bin boundaries, unless there is some sort of exception rule where one could say, e.g., the rightmost bin boundary will be taken from the domainSpec instead. E.g., 
```
domain =  [0, 0.8, 1.5, 1.8, 2.0, 2.5, 3.1, 4.0, 4.7, 5.3]; 
domainSpec = [0, 6.0]; 
```
then a \bins plotmode could be something like this (not tested yet!)

```
	bins { |x, y|
		var gap = if(halfBinWidthPx >= 5) { halfBinWidthPx * 0.05 } { 0 };
		// gap needs to be independent of any one bin width here,
		// but maybe somehow relate to the width of the smallest bin? -> TBD
		var specwarp = this.spec.warp;
		var centery = if(specwarp === \lin ) {specwarp.unmap(0) * plotBounds.height} { 0 }  ;
                //centery will only really work in \lin warps, so ignore it otherwise.

		Pen.smoothing_(false);

		y.size.do { |i|
			var xLeftYCenter = x[i]@0;
			var xRightYBarEnd = (x[i + 1] ?? { (1 - xPlotPad) * plotBounds.width })@y[i]; 
                         // 1 - xPlotPad would be equal to prDomainSpec.unmap(domainSpec.maxval)

			Pen.addRect( 
				Rect.fromPoints(xleftYCenter, xrightYBarEnd)
					.insetBy(gap, 0)
					.moveBy(0, centery);
			)
		}

	}
```
using those Rect methods (`fromPoints, insetBy, moveBy`) because I find it a bit easier to glean what's going on. Not sure if this is true for other readers, though.

mtmccrea:
I see a few separate issues emerging which would be actionable new features:

1. A more intelligent padding on the left and/or right of datapoints

2. support non-uniform bin widths for step-like plot modes.

3. support equidistant data spacing when automatically assigning domain values on non-linear domainSpecs (e.g. \exp)

Would you agree these are the relevant feature requests you've brought up? Would you like to separate them out? Again, driving new features with concrete use cases add clarity to the intention and the usefulness of each proposal.

Some thoughts on each:

(1) seems useful for different plotting scenarios, fairly easy to reason about:
 - Decide how best to calculate the pad based on the data's distribution on the x-axis
 - Omit padding if the range of the domain already exceeds the min/max data domain point

(2) is an aesthetic tweak that would likely require user input on the behavior (we may not always want space-filling steps) or a dedicated plot mode(s).

(3) For uniform spanning in `\exp` warping, this may be straightforward, I'm not sure about other spec warping (I'm not familiar with the algo). More generally, this feature would require keeping track of whether the domain was user- or auto-assigned to know how to handle assigning the domain if the domain spec changes. From what I see so far, adding this rather nuanced implementation for added convenience on the user end may not be worth it when the user can just assign the domain of the data themselves?

---

As I mentioned before, we might be wary of building more logic into the plot modes — edge cases tend to proliferate and the codebase is already getting a bit tough to reason through. Not trying to deter the additions, but I think we're already seeing how it's not as straightforward as initially thought...

HotwheelsSisyphus:
Well, in terms of the "issues" you isolate, this PR is meant to address (3) (I mean, see title and original PR description), and everything else I've brought up (e.g., padding) is just either a) me thinking aloud (the  comment earlier today) or b) is being brought up because its implementation is, in my eyes, the cause for why things didn't work the way I thought they were meant to. (the original PR description)

Focusing on that (3) and addressing your concerns.

> this feature would require keeping track of whether the domain was user- or auto-assigned to know how to handle assigning the domain if the domain spec change

the `Plot` instance nowhere assigns `domain` if said domain is not already assigned. 

Also, this PR did not propose changing behavior in this respect; rather, it generalized a preexisting behavior (if domain is nil, create x-values on the basis of the domainSpec, which latter defaults to linear.). 

This part IS straightforward IMO. I may not have framed the whole thing ideally from the get go but I was somewhat hoping that commenters would actually look at the five or so lines that have been changed. To be honest I'm not convinced this has come to pass. Maybe it would be better to move the discussion to comments on individual lines of code; let's abstain from writing further novels here.

mtmccrea:
OK, just sticking to (3) — preferring a uniform spacing of data points regardless of the spec's warp — I'll reiterate my position from the beginning: I disagree with this misunderstanding of how the domain spec works.

I don't see a rationale for uniform spacing irregardless of the spec's warp other than at some point you expected this. I've asked for examples so that I might understand where you're coming from and you haven't offered any more than your original cursory example, which I modified to do what I thought you wanted with one extra line.

The follow-up use case you proposed, a frequency spectrum, actually wasn't valid and is in fact a good example of why you would expect non-uniform bin spacing.

With your proposed change, if you plot data and switch between linear and non-linear domains, you wouldn't see any change in the plot, which is counterintuitive and violates the role of a warped domain. 
```supercollider
p = 5.collect{ rrand(1.0, 5) }.plot.plotMode_(\dlevelsCentered);
// toggle between these
p.domainSpecs_([1, 20].asSpec);
p.domainSpecs_([1, 20, \exp].asSpec);
```

An unspecified domain (`nil`) defaults to uniform spacing on a linear domain because it's by far the most common plotted data: a signal sampled evenly in time where the domain are sample indices. I cannot offhand think of a sensible default domain for arbitrary warping and you have not proposed one.

Furthermore, while I did review the code, I didn't comment on it because i think the premise still wasn't on firm ground (and it was making additional changes to bar padding, which wasn't the focus of this PR). I pulled the change, tested it and found it breaks current functionality with regard to where data falls on the domain:

Before:
Screenshot 2026-08-05 at 19 30 19

This PR:
Screenshot 2026-08-05 at 19 29 57

Others may disagree and want to review this further.

SCClassLibrary/Common/GUI/PlusGUI/Math/PlotView.sc
```diff
@@ -582,7 +576,8 @@ Plot {
 	}
 
 	bars { |x, y|
-		var gap = halfBinWidthPx * 0.1;
+		// if gaps would be < 1 pixel, don't draw gaps at all.
+		var gap = if(halfBinWidthPx >= 5) { halfBinWidthPx * 0.1 } { 0 };
```
mtmccrea, line 580: I think this is useful!

mtmccrea:
> the Plot instance nowhere assigns domain if said domain is not already assigned.

You're right to flag this — the linear, index-based domain is the implicit default. The `domain` variable remains `nil` to indicate that it has not been explicitly set, by design.

HotwheelsSisyphus:
Ok, at this point I finally decide that this is not worth my time and yours. I'd have been happy to fix some of the actual details such as the examples in the penultimate comments, but I'm running up against a wall of "I disagree with this on principle" with a PR that was meant to do something pretty much uncontroversial. Thanks for your time and bye.

mtmccrea:
Please feel free to leave this open for others to review if you still feel that this change is beneficial.

PR #7657, opened 2026-07-25
Make doc heading sticky but collapse upon scrolling

### LINKED ISSUE #7655: Expose Current Documentation Title Always Even Though Page is scrolled down
## Motivation
When scrolling through the SuperCollider help documentation, the title of the current page disappears. This makes it difficult to know which document is being viewed when navigating long pages.

The documentation viewer should keep the current page title visible (e.g., pinned at the top) even when the user scrolls down.

Image
Image

## Plan for Implementation
~~I may implement it by editing `SCDocHTMLRenderer` if it can be placed to the left of the `Table of Contents`.~~ (`SCDocHTMLRenderer` does not contain the menubar.)
However, the best place seems to be the right side of  the `Help browser` or the left side of the `Home` on the `Help browser bar`.

capital-G:
## Purpose and Motivation

Closes https://github.com/supercollider/supercollider/issues/7655

Makes the class name/header sticky and collapse it upon scrolling such that it doesn't consume too much screen estate. Adds even more JS though ;/

Maybe prko can test this on some site, i.e. search, home, browse, etc - on my quick test I hadn't run into any issues

https://github.com/user-attachments/assets/687b38a0-1a3b-4a00-b8a0-0c15178dfb56

## Types of changes

- New feature

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/HelpSource/static/scdoc.css
+++ b/HelpSource/static/scdoc.css
@@ -54,8 +54,13 @@ p {
 
 a {
     /* The menu bar takes some pixels which we need to respect
-    when jumping to an anchor. */
-    scroll-margin-top: var(--menu-bar-height);
+    when jumping to an anchor.
+    Additionally, some offset is added b/c of the dynamic height
+    of the collapsing header (see .header.collapsed).
+    At some point, this should be dynamically linked via a css property
+    instead of hardcoding heights...
+     */
+    scroll-margin-top: calc(var(--menu-bar-height) + 70px);
 }
 
 a:link, a:visited, a:link:hover {
--- a/HelpSource/static/scdoc.css
+++ b/HelpSource/static/scdoc.css
@@ -125,7 +125,7 @@ a.subclass_toggle:hover {
     padding: 0.5em 0;
     background-color: var(--color-bg);
     box-shadow: 0em 0em 0.25em var(--color-fg-200);
-    z-index: 2;
+    z-index: 3;
 }
 
 #nav {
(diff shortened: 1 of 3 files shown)


JordanHendersonMusic:
I'm not against this, but do wonder what kind of performance implications this could have - I'm thinking about raspberry pi's and whatnot here ...

capital-G:
I'd say that the performance impact is rather neglectable - css animations are pretty efficient, especially on webkit, and we are scrolling (and therefore re-rendering) nonetheless. If you observer any kind of stuttering we could back off here, but I'd be surprised.

prko:
I tested it

- on macOS 15.7.7 on a physical machine,
- on Windows 11 in a virtual machine via Parallels Desktop, and
- on Ubuntu in a virtual machine via Parallels Desktop.

The pages I tested are: Home, SinOsc, JSONlib (only on macOS 15.7.7), Browser, Class Tree, Methods, Documents, and Classes.

It works smoothly on all platforms!

Thank you for this.
The way you implemented it is something I hadn’t imagined, and it’s truly beautiful.

JordanHendersonMusic (approved):

prko:
Ah, one more thing.

When selecting an item from the table of contents, the first two lines of the target section are hidden behind the sticky header. This offset should be taken into account when automatically scrolling to the selected item from the table of contents.


Google Gemini suggests:
https://github.com/supercollider/supercollider/blob/33403268729b52f92029117c205d23529d1f7e4a/HelpSource/static/scdoc.css#L620-L625
```css
:target {
    border-bottom: 6px solid var(--color-warning-bg);
    scroll-margin-top: calc(var(--menu-bar-height) + 50px); 
}
```
I tested this on all platforms, and it works, but I am not sure if this is the best way to do it.

---
added:

80px is better:
```css
    scroll-margin-top: calc(var(--menu-bar-height) + 80px); 
```

dyfer:
I like the feature!

But yes, as @prko indicated, the target when selecting an anchor from TOC needs to now be offset...

capital-G:
Thanks for testing @prko, haven't noticed this before, but fixed it now by adding a constant of 70px. I think this solution is now hacky b/c we are hardcoding heights which can vary if people apply custom styling.
If others think this is still desirable, we could merge it, but ~~I think it isn't worth the technical debt introduced by it in relation to custom styling that people may have in place.~~ it introduces technical debt due to the constant and it may break others people custom styling - ooth, this is just an anchor.

PR #7659, opened 2026-07-27
Update Linux compilers in GHA

dyfer:
## Purpose and Motivation

Test building SC against newer compilers.

Here's the logic:
- build against the last 4 versions of gcc and clang but additionally keep the oldest version previously in the matrix
- I have not touched the gcc-12 / Ubuntu 22.04 build since this is what we run the testsuite with; the serial port tests seem to fail with Ubuntu 24.04 so I have not changed that configuration
- we can still drop testing with old compilers if maintaining compatibility becomes burdensome

## Types of changes

- Maintenance

## To-do list

- [x] Code is tested
- [ ] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -16,14 +16,6 @@ jobs:
 
           - job-name: ""
             os-version: "22.04"
-            c-compiler: "gcc-9"
-            cxx-compiler: "g++-9"
-            use-syslibs: false
-            shared-libscsynth: false
-            ctest: false
-
-          - job-name: ""
-            os-version: "22.04"
             c-compiler: "gcc-11"
             cxx-compiler: "g++-11"
             use-syslibs: false
@@ -54,35 +46,27 @@ jobs:
             use-syslibs: false
             shared-libscsynth: true
             ctest: false
+
+          - job-name: ""
+            os-version: "26.04"
+            c-compiler: "gcc-15"
+            cxx-compiler: "g++-15"
+            use-syslibs: false
+            shared-libscsynth: false
+            ctest: false
+
+          - job-name: "with ctest"
+            os-version: "26.04"
+            c-compiler: "gcc-16"
+            cxx-compiler: "g++-16"
+            use-syslibs: false
+            shared-libscsynth: false
+            ctest: true
             
           - job-name: ""
             os-version: "22.04"
             c-compiler: "clang-11"
             cxx-compiler: "clang++-11"
-            use-syslibs: false
-            shared-libscsynth: false
-            ctest: false
-
-          - job-name: ""
-            os-version: "24.04"
-            c-compiler: "clang-15"
-            cxx-compiler: "clang++-15"
-            use-syslibs: false
-            shared-libscsynth: false
-            ctest: false
-
-          - job-name: ""
-            os-version: "24.04"
-            c-compiler: "clang-16"
-            cxx-compiler: "clang++-16"
-            use-syslibs: false
-            shared-libscsynth: false
-            ctest: false
-
-          - job-name: ""
-            os-version: "24.04"
-            c-compiler: "clang-17"
-            cxx-compiler: "clang++-17"
             use-syslibs: false
             shared-libscsynth: false
             ctest: false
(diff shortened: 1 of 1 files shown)


dyfer:
@JordanHendersonMusic did you mention that in your recent PRs you encountered some compatibility issues with compiler versions? In any case, do you think it's okay to update the build matrix like this, or maybe this PR should wait until your sclang/lexer/parser PRs get merged?

capital-G (approved):

JordanHendersonMusic:
I think the clang stuff is okay, but let's keep gcc11, the old gccs are the ones I've had issues with.

To be honest, anything that works on an old compiler ought to work on the new one. It's the other way around that's the issue. So I don't really see much benefit in this.

dyfer:
> I think the clang stuff is okay, but let's keep gcc11, the old gccs are the ones I've had issues with.

To clarify, should I remove gcc9 and keep gcc11 instead? I'd like to generally keep last four + one older.

At this point we certainly don't need to support gcc9.

> So I don't really see much benefit in this.

The general idea is to build against what we [claim to support](https://github.com/supercollider/supercollider/wiki/Platform-Support#toolchains) (+ some extras, if that's not too much hassle). Is this something we should change/discuss?

JordanHendersonMusic:
> At this point we certainly don't need to support gcc9.

I've had things that don't work on just gcc9. I'd love to remove it, but we can only do that if we are really certain we don't need it.

Just for context, gcc9 came out in 2019, but got its final update in 2022. Yet we only support c++17, not even 20.

> The general idea is to build against what we [claim to support](https://github.com/supercollider/supercollider/wiki/Platform-Support#toolchains)

I agree with this, but it is different from what we currently support so could cause breaking changes.

If you think this is okay (I don't know too much about all this) then lets test on all compilers we claim to support and remove the rest.


Also of note here is apple clang.. I believe it is sometimes different from normal clang.


*its annoying that c++ doesn't just have one compiler!!*

dyfer:
> > At this point we certainly don't need to support gcc9.
> 
> I've had things that don't work on just gcc9. I'd love to remove it, but we can only do that if we are really certain we don't need it.

We don't claim to support it (well, we do in the readme, but it's outdated). I think it's fine to remove it.

> 
> Just for context, gcc9 came out in 2019, but got its final update in 2022. Yet we only support c++17, not even 20.
> 
> > The general idea is to build against what we [claim to support](https://github.com/supercollider/supercollider/wiki/Platform-Support#toolchains)
> 
> I agree with this, but it is different from what we currently support so could cause breaking changes.

It's different because we don't aggressively keep up with the toolchain releases :)

FWIW we should probably start building with msvc2026 (not necessarily for the release, but just to make sure it builds...)

> 
> If you think this is okay (I don't know too much about all this) then lets test on all compilers we claim to support and remove the rest.

I personally always aim to support older stuff, but we shouldn't bend ourselves backwards if there's an issue. In the future, when you experience issues with compilers outside of our declared support, please flag it in the PR so that we can consider removing it from the CI etc.

> 
> Also of note here is apple clang.. I believe it is sometimes different from normal clang.

Yeah, that's why we also list supported xcode versions.

dyfer:
I removed the gcc-9 job. OK to merge?

JordanHendersonMusic (approved):
Cool, this make things easier, hopefully it doesn't cause any issues!

.github/workflows/build_linux.yml
```diff
@@ -148,7 +156,10 @@ jobs:
...
 
       - name: install qt from apt
-        run: sudo apt-get install --yes qt6-base-dev qt6-base-dev-tools qt6-tools-dev qt6-tools-dev-tools qt6-declarative-dev libqt6gui6 libqt6printsupport6 libqt6svgwidgets6 libqt6websockets6-dev libqt6webenginecore6 qt6-webengine-dev qt6-webengine-dev-tools libqt6webchannel6-dev libqt6opengl6-dev libqt6svg6-dev linguist-qt6 qt6-l10n-tools  libglx-dev libgl1-mesa-dev libvulkan-dev libxkbcommon-dev libxcb-xkb-dev
+        run: |
+          sudo apt-get install --yes qt6-base-dev qt6-base-dev-tools qt6-tools-dev qt6-tools-dev-tools qt6-declarative-dev libqt6gui6 libqt6printsupport6 libqt6svgwidgets6 libqt6webenginecore6 qt6-webengine-dev qt6-webengine-dev-tools linguist-qt6 qt6-l10n-tools libglx-dev libgl1-mesa-dev libvulkan-dev libxkbcommon-dev libxcb-xkb-dev libqt6opengl6-dev libqt6svg6-dev
+          # some package names differ between ubuntu 24.04 and 26.04 so we try both
+          sudo apt-get install -y libqt6webchannel6-dev libqt6websockets6-dev || sudo apt-get install -y qt6-webchannel-dev qt6-websockets-dev
```
capital-G, line 162: @dyfer @jamshark70 Maybe this should also be documented in the linux readme? See https://scsynth.org/t/libqt6websockets6-dev/13518

PR #7665, opened 2026-07-29
Improve SerialPort tests

dyfer:
## Purpose and Motivation

This PR enables serialport tests on macOS and provides preliminary support for testing on Windows.

Testing Serial Port on Windows yields some issues so they are not enabled in the CI as part of this PR. I plan to work on that later.

If desired, I can remove the explicit windows parts from this PR... but IMO they may as well go in for now.

## Types of changes

- Tests
- New feature (socat on macOS)

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [n/a] Updated documentation

### DIFF
--- a/testsuite/classlibrary/TestSerialPort.sc
+++ b/testsuite/classlibrary/TestSerialPort.sc
@@ -11,6 +11,8 @@ TestSerialPort : UnitTest {
 	var skipSerialTests;
 	var input;
 	var output;
+	var socatPath;
+	var com0comPath;
 
 	const kBufferSize = 8192;
 
@@ -24,8 +26,8 @@ TestSerialPort : UnitTest {
 		this.destroyPorts();
 	}
 
-	// Return true if we are on Windows or socat is not installed.
-	// This method memoizes its results.
+	// Return if socat (macOS/Linux) or com0com (Windows) is not installed.
+	// This method memorizes its results.
 	// FIXME find better way to annotate/skip tests
 	skipSerialTests {
 		if(skipSerialTests.notNil) {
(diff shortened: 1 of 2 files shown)


capital-G (approved):
I haven't deeply inspected the code, but if it works it works I guess - so lgtm and feel free to merge.
Though some docs about socat and com0com would also be appreciated - I guess these are programs to mock serial ports?

There are some changes on the tests for SerialPort in the CI, which is what we want here.

dyfer:
> Though some docs about socat and com0com would also be appreciated - I guess these are programs to mock serial ports?

Yeah... though their setup looks a bit different.

Re: documentation - I was basically going off of what the source and the looking up what the tool does etc I guess I can add my findings to the comments - OTOH these are just tests so is it worth it?

testsuite/classlibrary/TestSerialPort.sc
```diff
@@ -267,17 +354,25 @@ TestSerialPort : UnitTest {
...
+		"Written bytes: %\n".postf(written);
+
+		watchdog = fork { 5.wait; cond.test_(true).signal; "TIMEOUT!".warn };
 
 		// spin until all data has been read
 		while { (rxErrs == 0) and: cond.test.not } { rxErrs = in.rxErrors; 0.01.wait; };
 
 		this.assert(rxErrs > 0);
 
+		watchdog.stop();
```
JordanHendersonMusic, line 374: Rather than spinning up another thread, can we check the elapsed time in loop instead?

JordanHendersonMusic (commented):
One question, otherwise I think this is good. I agree that some comments describing what these tools do would be useful, like a sentence or two.

testsuite/classlibrary/TestSerialPort.sc
dyfer: I think I corrected this.

dyfer:
> Though some docs about socat and com0com would also be appreciated - I guess these are programs to mock serial ports?

Yeah, I was simply going off of https://github.com/supercollider/supercollider/pull/7665/changes#diff-e70d213cc4d1e6206ddf4d3aadcc6d155e36f7af24a11a3f5cf35d60d7a6389fR6-R7, which, while minimal, IMO gives enough clues on what's needed.

JordanHendersonMusic:
Cool, are we happy to merge?

dyfer:
@capital-G are you okay with this or should I add more description? If so, what do you think is needed?

capital-G:
> @capital-G are you okay with this or should I add more description? If so, what do you think is needed?

I still don't have too much of a clue what those programs are doing (I'd guess mocking a com port?^^), but I think in case those tests create problems a google search will tell me ;)

feel free to merge

dyfer:
> I still don't have too much of a clue what those programs are doing (I'd guess mocking a com port?^^), but I think in case those tests create problems a google search will tell me ;)

Me neither, but reading https://github.com/dyfer/supercollider/blob/e4e0adcd73a5c1c599a0484af4a53ee8c4237505/testsuite/classlibrary/TestSerialPort.sc#L6-L7 was enough for me to understand which tools are needed to mock the port, which I think is enough here.

I'll merge this, thanks!

PR #7667, opened 2026-07-31
Recorder: Use String instead of Symbol for file path in Post window

prko:
## Purpose and Motivation

**File paths are typically used as Strings, and using Symbols for file paths is uncommon. This PR improves workflow efficiency by ensuring that copied file paths behave as expected. `.openOS` below is just an example of using pasted file paths from the Post window.**

This PR fixes an issue in several recorder‑related log messages where the percent sign indicating the path of the recorded file was printed using single quotes, for example:
```supercollder
"Recording channels % ... \npath: '%'\n".postf(recordBuf.path);
"Resumed recording ...\npath: '%'\n".postf(recordBuf.path);
"... paused recording.\npath: '%'\n".postf(recordBuf.path);
```
When users copy the printed path from the Post window
```text
-> localhost
Preparing recording on 'localhost'
Recording channels [0, 1] ... 
path: '/Users/prko/Music/SuperCollider Recordings/SC_260731_131303.wav'
Recording Stopped: (SC_260731_131303.wav)
```
 and attempt to open it via:
```supercollider
'/Users/prko/Music/SuperCollider Recordings/SC_260731_131303.wav'.openOS // <- <copied path>.openOS;
```
SuperCollider interprets the copied value as a Symbol, not a String, which results in the following error:
```text
ERROR: Message 'openOS' not understood.
RECEIVER:
   Symbol '/Users/.../SC_260731_131303.wav'
```
Only String objects respond to openOS, so this behaviour prevents users from opening recorded files directly from the copied output.

To resolve this, the percent sign is now printed inside double quotes:
```supercollider
"Recording channels % ... \npath: \"%\"\n".postf(recordBuf.path);
"Resumed recording ...\npath: \"%\"\n".postf(recordBuf.path);
"... paused recording.\npath: \"%\"\n".postf(recordBuf.path);
```

## Types of changes

- Improve workflow

## To-do list

- [x] Code is tested
- [x] All tests are passing

### DIFF
--- a/SCClassLibrary/Common/Control/Recorder.sc
+++ b/SCClassLibrary/Common/Control/Recorder.sc
@@ -35,7 +35,7 @@ Recorder {
 			if(this.isRecording.not) {
 				this.prRecord(bus, node, duration);
 				this.changedServer(\recording, true);
-				"Recording channels % ... \npath: '%'\n"
+				"Recording channels % ... \npath: \"%\"\n"
 				.postf(bus + (0..this.numChannels - 1), recordBuf.path);
 			} {
 				if(paused) {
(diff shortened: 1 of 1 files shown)


capital-G (approved):
Thx prko!

PR #7671, opened 2026-08-01
Fix dangling connection for tcp disconnect on scsynth

### LINKED ISSUE #5337: Scsynth doesn't clean up TCP connections

capital-G:
This issue came up some days ago w/ a performance w/ multiple clients.

When a tcp client disconnected, it was never removed from the `/notify` workflow (aka `mUsers` on the hidden world) which became an issue b/c this resulted writing messages into a closed socket. When closing the tcp socket, the server will now remove the client similar to `/notify 0`.

Fixes #5337

Contrary to the default way of a very sparse documentation, I added some doxygen docs for it to make the intention clear and explain any quirks.
The implementation is somehow linked to `NotifyCmd::Stage2` so I linked both implementations via documentation. At some point this should be abstracted away in its own class since (hidden) world is already quiet entangled and verbose.

I'd appreciate if people confirm this fixes it on various platforms, I only tested macOS. Please take a look into the issue on how to perform the test, if it works, no (excessive) printing of `Bad file descriptor` should appear.

## Types of changes

- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

### DIFF
--- a/server/scsynth/SC_ComPort.cpp
+++ b/server/scsynth/SC_ComPort.cpp
@@ -44,6 +44,7 @@
 
 #include "nova-tt/semaphore.hpp"
 #include "nova-tt/thread_priority.hpp"
+#include "oscpack_1_1_0/osc/OscException.h"
 
 #ifdef USE_RENDEZVOUS
 #    include "Rendezvous.h"
@@ -498,7 +499,9 @@ SC_TcpConnection::~SC_TcpConnection() {
     AudioDriver(mWorld)->SendMsgFromEngine(msg);
 
     // now close the socket
-    socket.close();
+    try {
+        socket.close();
+    } catch (boost::system::system_error& e) { printf("ERROR: Could not close TCP socket: %s\n", e.what()); }
     mParent->connectionDestroyed();
 }
 
--- a/server/scsynth/SC_ComPort.cpp
+++ b/server/scsynth/SC_ComPort.cpp
@@ -38,6 +38,8 @@
 #include <boost/enable_shared_from_this.hpp>
 #include <boost/typeof/typeof.hpp>
 
+#include "SC_CoreAudio.h"
+#include "SC_FifoMsg.h"
 #include "SC_Lock.h"
 
 #include "nova-tt/semaphore.hpp"
@@ -48,7 +50,9 @@
 #endif
 
 
+// forward declarations
 bool ProcessOSCPacket(World* inWorld, OSC_Packet* inPacket);
+void World_RemoveClient(FifoMsg* msg);
 
 namespace scsynth {
 
@@ -332,6 +336,11 @@ public:
         boost::asio::ip::tcp::no_delay noDelayOption(true);
         socket.set_option(noDelayOption, error);
 
+        mClientIdentification.mProtocol = kTCP;
+        mClientIdentification.mPort = socket.remote_endpoint().port();
+        mClientIdentification.mSocket = socket.native_handle();
+        mClientIdentification.mAddress = socket.remote_endpoint().address();
+
         // first message must be the password. 4 tries.
         bool validated = mWorld->hw->mPassword[0] == 0;
         for (int i = 0; !validated && i < 4; ++i) {
@@ -368,6 +377,9 @@ private:
     int32 OSCMsgLength;
     char* data;
     class SC_TcpInPort* mParent;
+
+    /** acts as the identification within SC world, which is necessary when deregistering upon disconnect. */
+    ReplyAddress mClientIdentification;
 
     void handleLengthReceived(const boost::system::error_code& error, size_t bytes_transferred) {
         if (error) {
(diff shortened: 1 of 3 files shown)


server/scsynth/SC_ComPort.cpp
```diff
@@ -48,7 +50,9 @@
 #endif
 
 
+// forward declarations
```
JordanHendersonMusic, line 53: Could you explain in the comment why you are forward declaring, we know what a forward declare is.

server/scsynth/SC_ComPort.cpp
```diff
@@ -477,7 +489,18 @@ class SC_TcpInPort {
...
+    // we use fifomsg to get access to the stage2 thread lock.
+    // the callback function removes its passed data, so we make a copy of our reply address.
+    FifoMsg msg;
+    msg.Set(mWorld, World_RemoveClient, nullptr, new ReplyAddress(mClientIdentification));
+    AudioDriver(mWorld)->SendMsgFromEngine(msg);
+
+    // now close the socket
+    socket.close();
+    mParent->connectionDestroyed();
+}
```
JordanHendersonMusic, line 505: Can any of this throw, and is it likely to do so? Throwing in destructors can be weird as they are called when something else is already being thrown.

server/scsynth/SC_SequencedCommand.cpp
```diff
@@ -1268,6 +1268,7 @@ bool NotifyCmd::Stage2() {
         SendDoneWithVarArgs(&mReplyAddress, "/notify", "ii", clientID, (int)hw->mMaxUsers);
 
     } else {
+        // keep this in sync w/ `World_RemoveClient` implementation for TCP de-registration via disconnect
```
JordanHendersonMusic, line 1271: Seems like this is likely to be forgotten, why not just make it a function? It's quite complex as there are several structures here to interact with.

server/scsynth/SC_ComPort.cpp
capital-G: The function I am referring to is declared in a cpp file, which I think should/can not be included. I can remove the comment though.

server/scsynth/SC_ComPort.cpp
capital-G: Good idea, I'll check on that

server/scsynth/SC_SequencedCommand.cpp
capital-G: I am not too eager to write a function for that (though it is probably better than the status of this PR), but I'd prefer to abstract this away in its own `ClientManager` class. Right now there are 3 member variables in hidden world which you need to keep in sync through the function - ideally these should be hidden by a class abstraction, accessed via methods.

Would you "greenlit" such a class or do you think this would be a wrong approach?

server/scsynth/SC_ComPort.cpp
JordanHendersonMusic: Then can you declare it in the header too. It is weird to include from just the source.

It's like accessing a private variable.

capital-G: Well yeah, but all the functions of `SC_SequencedCommand` are declared in the `include` section, making them a public facing API.
Since this is a private implementation, I wouldn't expose it to clients via the API. I could make a new hpp for this, but it seems overkill - and it is already used for `ProcessOSCPacket` as well.

JordanHendersonMusic: Hm. The include folder is a bit unfortunate really. Okay leave it then.

server/scsynth/SC_SequencedCommand.cpp
JordanHendersonMusic: The world or hidden world could hold all IO managers, yes I think this sound fine, but should be another PR. Putting these into functions will make writing that PR easier.

server/scsynth/SC_ComPort.cpp
capital-G: I catched the possible exception of `socket.close` now.

server/scsynth/SC_SequencedCommand.cpp
capital-G: Then let's get this merged and then we can merge get #7676 merged ;)

server/scsynth/SC_ComPort.cpp
JordanHendersonMusic: do you want to rethrow it after doing ` mParent->connectionDestroyed();`?

capital-G: I thought we shouldn't throw in a destructor?

JordanHendersonMusic: Ah no you are right! Looks good then!

capital-G:
@JordanHendersonMusic should we merge this or should I move this into a function? Seems a bit redundant since #7676 is available IMO ;)

JordanHendersonMusic (approved):

PR #7672, opened 2026-08-02
plugins: Median: clip length to 1 to fix out-of-bounds server crash

### LINKED ISSUE #7645: Applying `Median` with a value < 1 Hard Crashes the Server
Reprex:

```sclang
s.waitForBoot({
    var def = SynthDef(\median, { arg out = 0;
		var signal = Median.ar(0.5, SinOsc.ar());
        Out.ar(out, signal.dup);
    }).add;
	
    s.sync;
    def.play;
});
```
```
-> localhost
```
tested on 3.14.1, Windows

Obviously 0.5 for a median filter is an invalid argument, but it could be reasonable for the server to gracefully exit and throw an error or output the original signal like it does for `1` instead of crashing. The above behavior also happens for values like `-1`.

vlerdf:
## Purpose and Motivation

Fixes #7645.

`Median_Ctor` clips the length argument with a lower bound of 0, so `Median.ar(0.5, ...)` ends up with `m_medianSize == 0`. In `Median_InsertMedian` that makes `last` -1, the ageing loop never runs, and `pos` is still -1 when the loop ends. It then reads `m_medianValue[pos - 1]` and writes `m_medianValue[pos]`, both out of bounds, and scsynth dies. Negative lengths clip to 0 and hit the same path.

Clipping to 1 instead. Median.schelp already gives the valid range as 1 to 31, and Pitch guards its own median with `if (unit->m_medianSize > 1)`, so nothing was meant to run below that. Length 1 is already the pass-through case: `pos` becomes 0, neither while loop runs, and the input sample comes back unchanged.

## Types of changes

- Documentation
- Bug fix

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

I confirmed the overflow and the fix with AddressSanitizer and the plugin builds clean, but I couldn't get the class library suite to run locally (headless sclang never finished booting the server here), so CI is the first real run of the new test. Leaving that box unchecked until it goes green.

### DIFF
--- a/testsuite/classlibrary/TestFilterUGens.sc
+++ b/testsuite/classlibrary/TestFilterUGens.sc
@@ -167,16 +167,23 @@ TestFilterUGens : UnitTest {
 	// A length below 1 left the insertion index at -1 in Median_InsertMedian,
 	// which then indexed the median buffer out of bounds and took the server
 	// down with it. Median's own help file documents the valid range as 1 to 31.
-	test_Median_length_below_one_does_not_crash_server {
-		[0.5, 0, -1].do { |length|
-			{ Median.ar(length, DC.ar(0)) }.play(target: server);
-		};
+	test_medianLengthBelowOne {
+		var condition = CondVar();
+		var signal;
 
-		server.sync;
+		{ Median.ar(length: [0.5, 0, -1], in: DC.ar(0)) }.loadToFloatArray(
+			duration: 0.01,
+			target: server,
+			action: { |array| signal = array; condition.signalOne }
+		);
 
 		this.assert(
-			server.serverRunning and: { server.unresponsive.not },
-			"server should survive Median.ar with a length below 1"
+			condition.waitFor(5),
+			"Median.ar with a length below 1 should render instead of taking the server down"
+		);
+		this.assert(
+			signal.notNil and: { signal.every { |sample| sample == 0.0 } },
+			"a length below 1 should clip to 1, which passes the input through unchanged"
 		);
 	}
 
--- a/testsuite/classlibrary/TestFilterUGens.sc
+++ b/testsuite/classlibrary/TestFilterUGens.sc
@@ -172,7 +172,7 @@ TestFilterUGens : UnitTest {
 			{ Median.ar(length, DC.ar(0)) }.play(target: server);
 		};
 
-		0.5.wait;
+		server.sync;
 
 		this.assert(
 			server.serverRunning and: { server.unresponsive.not },
--- a/testsuite/classlibrary/TestFilterUGens.sc
+++ b/testsuite/classlibrary/TestFilterUGens.sc
@@ -164,6 +164,22 @@ TestFilterUGens : UnitTest {
 		this.assertEquals(result, 1, "Integral of a single 1 should be 1");
 	}
 
+	// A length below 1 left the insertion index at -1 in Median_InsertMedian,
+	// which then indexed the median buffer out of bounds and took the server
+	// down with it. Median's own help file documents the valid range as 1 to 31.
+	test_Median_length_below_one_does_not_crash_server {
+		[0.5, 0, -1].do { |length|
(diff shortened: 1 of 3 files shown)


testsuite/classlibrary/TestFilterUGens.sc
```diff
@@ -164,6 +164,22 @@ TestFilterUGens : UnitTest {
...
 
+	// A length below 1 left the insertion index at -1 in Median_InsertMedian,
+	// which then indexed the median buffer out of bounds and took the server
+	// down with it. Median's own help file documents the valid range as 1 to 31.
+	test_Median_length_below_one_does_not_crash_server {
+		[0.5, 0, -1].do { |length|
+			{ Median.ar(length, DC.ar(0)) }.play(target: server);
+		};
+
+		0.5.wait;
```
JordanHendersonMusic: Can we rewrite this without a wait? A sync should do it?

JordanHendersonMusic (dismissed):
Thanks! One small thing.

testsuite/classlibrary/TestFilterUGens.sc
vlerdf: Done.
One thing: sync waits on a Condition with no timeout, so a regression would hang the test instead of failing with the assert message. I don't mind either way, but say if you'd prefer a bounded wait.

testsuite/classlibrary/TestFilterUGens.sc
```diff
@@ -164,6 +164,22 @@ TestFilterUGens : UnitTest {
 		this.assertEquals(result, 1, "Integral of a single 1 should be 1");
 	}
 
+	// A length below 1 left the insertion index at -1 in Median_InsertMedian,
+	// which then indexed the median buffer out of bounds and took the server
+	// down with it. Median's own help file documents the valid range as 1 to 31.
+	test_Median_length_below_one_does_not_crash_server {
```
capital-G: please use lowerCamelCase for method names.

Also `does not crash server` is probably a bit verbose here - a test name should primarily cover what gets tested, the outcome of a non-crashing server is given ;)

testsuite/classlibrary/TestFilterUGens.sc
```diff
@@ -164,6 +164,22 @@ TestFilterUGens : UnitTest {
...
+		[0.5, 0, -1].do { |length|
+			{ Median.ar(length, DC.ar(0)) }.play(target: server);
```
capital-G: this leaves a dangling synth on the server.

You could instead do something along these lines, which makes use of multi channel expansion to speed things up and actually checks the values of the signal, while also having a timeout awareness.

```supercollider
(
{
	var server = s;
	var success = false;
	var condition = CondVar();
	var lengths = [1.0, 2.0];
	
	{Median.ar(length: lengths, in: DC.ar(0.0))}.loadToFloatArray(duration: 0.01, action: {|sig|
		success = sig.any({|x| x-0.0 < 0.01});
		condition.signalOne;
	});
	if(condition.waitFor(2.0), {
		["Test was success?", success].postln;
	}, {
		"Timeout!".postln;
	});

}.fork;
)
```

capital-G (changes requested):

testsuite/classlibrary/TestFilterUGens.sc
vlerdf: Done, both points. Also switched the assertion to every with an exact zero check, since with DC.ar(0) and the length clipped to 1 the passthrough should give exact zeros.
Side benefit: waitFor also fixes the timeout concern @JordanHendersonMusic and I discussed above, the test now fails with a message instead of hanging if the server goes down.

vlerdf: @capital-G test-linux fails on TestArray:test_arraystats, which is array normalization and unrelated to this. Worth a re-run?

capital-G (approved):
Thanks :)

PR #7676, opened 2026-08-03
Introduce client manager class for scsynth

capital-G:
## Purpose and Motivation

Based on #7671

Instead of having the client involved code scattered around the server implementation with multiple data members which need to be kept in sync, this PR introduces a higher level API through a class without sacrificing performance.
This reduces the complexity of `HiddenWorld` by moving all client related code into this manager.
By using a class instead of a struct, we can also follow RAII, hopefully making the code clearer as well.

One thing this PR also introduces is the enforcement of a const `ReplyAddress` in some places. This became necessary because one only obtains a const reference while iterating over a `std::map`, but this also seems a cleaner implementation to me.
This introduces changes to the plugin API though! This is unfortunate, but we are already touching the plugin API for 3.15, so it may be a good opportunity as any to do this adjustment here?

This whole PR is maybe a bit unnecessary, but maybe this could be used as a starting point to make scsynth better documented and easier to understand by bundling abstractions?

This may be even introducing some performance improvements b/c the old code used `for (auto addr : *inWorld->hw->mUsers)` which, IIUC, would do a copy here instead of using a reference of `for (auto& addr : *inWorld->hw->mUsers)` - though I am unsure on that?

## Types of changes

- Refactoring

## To-do list

- [x] Code is tested
- [x] All tests are passing
- [x] Updated documentation

Merge note: Squash

### DIFF
--- a/include/common/SC_Reply.h
+++ b/include/common/SC_Reply.h
@@ -22,4 +22,4 @@
 #pragma once
 
 struct ReplyAddress;
-typedef void (*ReplyFunc)(struct ReplyAddress* inReplyAddr, char* inBuf, int inSize);
+typedef void (*ReplyFunc)(const ReplyAddress* inReplyAddr, char* inBuf, int inSize);
--- a/server/scsynth/SC_ClientManager.cpp
+++ b/server/scsynth/SC_ClientManager.cpp
@@ -1,1 +1,21 @@
 #include "SC_ClientManager.h"
+
+#include "SC_CoreAudio.h"
+#include "SC_FifoMsg.h"
+#include "SC_HiddenWorld.h"
+#include "SC_World.h"
+
+static void removeClientCallback(FifoMsg* msg) {
+    auto* world = msg->mWorld;
+    auto* address = static_cast<ReplyAddress*>(msg->mData);
+
+    world->hw->mClientManager->removeClient(*address);
+
+    delete address;
+}
+
+void ClientManager::removeClientDefer(World* world, const ReplyAddress& client) {
+    FifoMsg msg;
+    msg.Set(world, removeClientCallback, nullptr, new ReplyAddress(client));
+    AudioDriver(world)->SendMsgFromEngine(msg);
+}
--- a/server/scsynth/SC_ClientManager.cpp
+++ b/server/scsynth/SC_ClientManager.cpp
@@ -1,0 +1,1 @@
+#include "SC_ClientManager.h"
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -30,12 +30,7 @@ public:
      * @brief checks a given password and returns true if the password is valid.
      * If no password has been set, this will return true for any password.
      */
-    bool checkPassword(const std::string& password) const {
-        if (mPassword) {
-            return mPassword == password;
-        }
-        return true;
-    }
+    bool checkPassword(const std::string& password) const { return mPassword ? password == *mPassword : true; }
 
     /// returns true if the server has set a password
     bool hasPassword() const { return mPassword.has_value(); };
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -11,13 +11,18 @@ class ClientManager {
     /// maximum number of clients that the server accepts
     uint32 mMaxUsers;
 
+    /// since this gets accessed in the real time thread, this should only be accessed via
+    /// in protected environments such as stage1, stage2, ... or in the real time thread.
     std::map<ReplyAddress, uint32> mClientDict;
 
     /// the password necessary for login.
-    std::optional<std::string> mPassword;
+    std::optional<std::string> mPassword = {};
 
 
 public:
+    /**
+     * @param maxUsers maximum number of clients that can connect to the server
+     */
     ClientManager(uint32 maxUsers): mMaxUsers(maxUsers) {}
 
     /**
@@ -29,7 +34,7 @@ public:
             return mPassword == password;
         }
         return true;
-    };
+    }
 
     /// returns true if the server has set a password
     bool hasPassword() const { return mPassword.has_value(); };
@@ -37,18 +42,25 @@ public:
     /// set password for server login
     void setPassword(const std::optional<const std::string>& password) { mPassword = password; }
 
-    /// returns true iff the server has a free slot for a new client
-    bool clientSlotFree() const { return mClientDict.size() < mMaxUsers; }
-
     /// returns maximum number of allowed users for this server
     uint32 getMaxUsers() const { return mMaxUsers; }
 
+    /**
+     * @brief Returns the dict of all registered clients and their ID.
+     * Only access this in (N)RT locked environments!
+     */
     const std::map<ReplyAddress, uint32>& getClients() const { return mClientDict; }
 
-    /// returns true iff passed client was present and got removed
+    /**
+     * @brief returns true iff passed client was present and got removed
+     * Only call this in (N)RT locked environments!
+     */
     bool removeClient(const ReplyAddress& client) { return mClientDict.erase(client); }
 
-    /// returns a value iff client has an id and has therefore been registered
+    /**
+     * @brief returns a value iff client has an id and has therefore been registered
+     * Only call this in (N)RT locked environments!
+     */
     std::optional<uint32> getClientID(const ReplyAddress& address) const {
         auto it = mClientDict.find(address);
         if (it == mClientDict.end()) {
@@ -58,7 +70,8 @@ public:
     }
 
     /**
-     * @brief registers a client with
+     * @brief registers a client with an optional given ID.
+     * Only call this in (N)RT locked environments!
      */
     std::optional<uint32> registerClient(const ReplyAddress& address, std::optional<uint32> requestedID = {}) {
         auto id = getNextClientID(requestedID);
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -10,9 +10,6 @@ struct World;
  * @brief Manages client access of the server such as password check, client id issuing and handling tcp disconnects.
  */
 class ClientManager {
-    /// maximum number of clients that the server accepts
-    uint32 mMaxUsers;
-
     /// since this gets accessed in the real time thread, this should only be accessed via
     /// in protected environments such as stage1, stage2, ... or in the real time thread.
     std::map<ReplyAddress, uint32> mClientDict;
@@ -20,6 +17,8 @@ class ClientManager {
     /// the password necessary for login.
     std::optional<std::string> mPassword {};
 
+    /// maximum number of clients that the server accepts
+    uint32 mMaxUsers;
 
 public:
     /**
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -65,7 +65,7 @@ public:
         return true;
     }
 
-    ClientIDDict& getClientIDDict() { return mClientIDdict; }
+    uint32 getClientID(const ReplyAddress& address) const { return mClientIDdict.at(address); }
 
     void addClientId(uint32 clientID, ReplyAddress address) { mClientIDdict.insert(std::make_pair(address, clientID)); }
 
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -1,8 +1,10 @@
 #pragma once
 #include <deque>
 #include <map>
+#include <set>
 
 #include "SC_Types.h"
+#include "SC_ReplyImpl.hpp"
 
 typedef std::map<struct ReplyAddress, uint32> ClientIDDict;
 typedef std::deque<int> ClientIDs;
@@ -12,15 +14,28 @@ typedef std::set<ReplyAddress> Clients;
  * @brief Manages client access of the server such as password check, client id issuing and handling tcp disconnects.
  */
 class ClientManager {
-    // Clients mUsers;
-    // ClientIDs mAvailableClientIDs;
-    // ClientIDDict mClientIDdict;
+    /// maximum number of clients that the server accepts
+    uint32 mMaxUsers;
+
+    Clients* mUsers;
+    ClientIDs* mAvailableClientIDs;
+    ClientIDDict* mClientIDdict;
 
     /// the password necessary for login.
     std::optional<std::string> mPassword;
 
 
 public:
+    ClientManager(uint32 maxUsers):
+        mMaxUsers(maxUsers),
+        mUsers(new Clients()),
+        mAvailableClientIDs(new ClientIDs()),
+        mClientIDdict(new ClientIDDict()) {
+        for (int i = 0; i < mMaxUsers; i++) {
+            mAvailableClientIDs->push_back(i);
+        }
+    }
+
     /// checks a given password. if password has not been set, this will always return false,
     /// so check with has password before.
     bool checkPassword(const std::string& password) const { return mPassword == password; };
@@ -30,4 +45,10 @@ public:
 
     /// set password for server login
     void setPassword(const std::string& password) { mPassword = password; }
+
+    /// returns true iff the server has a free slot for a new client
+    bool clientSlotFree() { return 0 < mMaxUsers; }
+
+    /// returns maximum number of allowed users for this server
+    uint32 getMaxUsers() { return mMaxUsers; }
 };
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -66,11 +66,10 @@ public:
      * Only call this in (N)RT locked environments!
      */
     std::optional<ClientID> getClientID(const ReplyAddress& address) const {
-        auto it = mClientDict.find(address);
-        if (it == mClientDict.end()) {
-            return {};
+        if (auto it = mClientDict.find(address); it != mClientDict.end()) {
+            return it->second;
         }
-        return it->second;
+        return std::nullopt;
     }
 
     /**
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -18,7 +18,7 @@ class ClientManager {
     std::map<ReplyAddress, uint32> mClientDict;
 
     /// the password necessary for login.
-    std::optional<std::string> mPassword = {};
+    std::optional<std::string> mPassword {};
 
 
 public:
--- a/server/scsynth/SC_ClientManager.h
+++ b/server/scsynth/SC_ClientManager.h
@@ -1,14 +1,8 @@
 #pragma once
-#include <deque>
 #include <map>
-#include <set>
 
 #include "SC_Types.h"
 #include "SC_ReplyImpl.hpp"
-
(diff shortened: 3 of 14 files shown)


server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
...
+class ClientManager {
+    /// maximum number of clients that the server accepts
+    uint32 mMaxUsers;
+
+    /// since this gets accessed in the real time thread, this should only be accessed via
+    /// in protected environments such as stage1, stage2, ... or in the real time thread.
+    std::map<ReplyAddress, uint32> mClientDict;
+
+    /// the password necessary for login.
+    std::optional<std::string> mPassword = {};
```
JordanHendersonMusic: Don't need the equals, you are calling the copy constructor here. Just do `std::optional<std::string> mPassword {};`

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
...
+
+
+public:
+    /**
+     * @param maxUsers maximum number of clients that can connect to the server
+     */
+    ClientManager(uint32 maxUsers): mMaxUsers(maxUsers) {}
```
JordanHendersonMusic, line 31: Technically all single argument constructors should be explicit (very minor point here,  can ignore if you wish).

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
...
+#include "SC_ReplyImpl.hpp"
+
+struct World;
+
+/**
+ * @brief Manages client access of the server such as password check, client id issuing and handling tcp disconnects.
+ */
+class ClientManager {
+    /// maximum number of clients that the server accepts
+    uint32 mMaxUsers;
```
JordanHendersonMusic: personally I'd move this member after the mPassword, that way you don't have extra padding.

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
...
+     * @param maxUsers maximum number of clients that can connect to the server
+     */
+    ClientManager(uint32 maxUsers): mMaxUsers(maxUsers) {}
+
+    /**
+     * @brief checks a given password and returns true if the password is valid.
+     * If no password has been set, this will return true for any password.
+     */
+    bool checkPassword(const std::string& password) const {
+        if (mPassword) {
```
JordanHendersonMusic: could be a ternary ` return mPassword ?  *mPassword == password : true`

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
...
+            return mPassword == password;
+        }
+        return true;
+    }
+
+    /// returns true if the server has set a password
+    bool hasPassword() const { return mPassword.has_value(); };
```
JordanHendersonMusic, line 40: I like marking const functions `[[nodiscard]]`, it isn't really necessary here... but is very useful if you want to ensure someone always checks a returned error code! Feel free to ignore.

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * @brief Manages client access of the server such as password check, client id issuing and handling tcp disconnects.
+ */
+class ClientManager {
+    /// maximum number of clients that the server accepts
+    uint32 mMaxUsers;
+
+    /// since this gets accessed in the real time thread, this should only be accessed via
+    /// in protected environments such as stage1, stage2, ... or in the real time thread.
+    std::map<ReplyAddress, uint32> mClientDict;
```
JordanHendersonMusic: What does this Uint32 mean here?

Consider doing a using statement.

```
using WhatEverItMeans = uint32;

using ClientMap = std::map<ReplayAddress, WhatEverItMeans>;

ClientMap = mClientDict;
```

JordanHendersonMusic: Also, `std::map` is much slower than `std::unordered_map` do we need the ordering?

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
+#pragma once
```
JordanHendersonMusic, line 1: Since this is C++ code, the header should be `.hpp`

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,112 @@
...
+     * @brief returns a value iff client has an id and has therefore been registered
+     * Only call this in (N)RT locked environments!
+     */
+    std::optional<uint32> getClientID(const ReplyAddress& address) const {
+        auto it = mClientDict.find(address);
+        if (it == mClientDict.end()) {
+            return {};
+        }
+        return it->second;
+    }
```
JordanHendersonMusic: Using the capturing if lets you write this cleaner.

```
if(auto it = mClientDict.find(address); it != mClientDict.end())
   return it->second;
else
   return std::nullopt;
```

The nullopt if preference, but I think its clearer about the intent  here.

server/scsynth/SC_ComPort.cpp
```diff
@@ -163,7 +166,7 @@ static void udp_reply_func(struct ReplyAddress* addr, char* msg, int size) {
         printf("%s\n", errc.message().c_str());
 }
 
-static void tcp_reply_func(struct ReplyAddress* addr, char* msg, int size) {
+static void tcp_reply_func(const ReplyAddress* addr, char* msg, int size) {
     // Write size as 32bit unsigned network-order integer
```
JordanHendersonMusic, line 171: Can you go around to all these functions and make sure null checks are performed for addr and msg. Or add `assert(addr)` if it is a precondition.

server/scsynth/SC_ComPort.cpp
```diff
@@ -369,6 +377,9 @@ class SC_TcpConnection : public boost::enable_shared_from_this<SC_TcpConnection>
     char* data;
     class SC_TcpInPort* mParent;
 
+    /** acts as the identification within SC world, which is necessary when deregistering upon disconnect. */
+    ReplyAddress mClientIdentification;
```
JordanHendersonMusic: Best to place large members first in the struct.

server/scsynth/SC_SequencedCommand.cpp
```diff
@@ -1214,72 +1214,29 @@ int NotifyCmd::Init(char* inData, int inSize) {
...
-                                        hw->mClientIDdict->at(mReplyAddress));
-                scprintf("/notify : already registered\n");
-                return false;
+        if (auto clientID = hw->mClientManager->getClientID(mReplyAddress)) {
+            SendFailureWithIntValue(&mReplyAddress, "/notify", "notify: already registered\n", *clientID);
+            scprintf("/notify : already registered\n");
+        } else {
+            if (auto newClientId = hw->mClientManager->registerClient(mReplyAddress)) {
+                SendDoneWithVarArgs(&mReplyAddress, "/notify", "ii", newClientId.value(),
+                                    (int)hw->mClientManager->getMaxUsers());
```
JordanHendersonMusic, line 1227: I'd use a static cast here personally.

server/scsynth/SC_HiddenWorld.h
```diff
@@ -94,21 +95,15 @@ typedef MsgFifoNoFree<NodeEndMsg, 1024> NodeEndsFifo;
...
 struct HiddenWorld {
     class AllocPool* mAllocPool;
     IntHashTable<struct Node, AllocPool>* mNodeLib;
     GrafDefTable* mGraphDefLib;
-    uint32 mMaxUsers;
-    Clients* mUsers;
-    ClientIDs* mAvailableClientIDs;
-    ClientIDDict* mClientIDdict;
+
+    ClientManager* mClientManager;
```
JordanHendersonMusic, line 104: Surely this would be better as a unique ptr?

server/scsynth/SC_World.cpp
```diff
@@ -332,13 +332,8 @@ World* World_New(WorldOptions* inOptions) {
...
         hw->mGraphDefLib = new HashTable<struct GraphDef, Malloc>(&gMalloc, inOptions->mMaxGraphDefs, false);
         hw->mNodeLib = new IntHashTable<Node, AllocPool>(hw->mAllocPool, inOptions->mMaxNodes, false);
-        hw->mUsers = new Clients();
-        hw->mMaxUsers = inOptions->mMaxLogins;
-        hw->mAvailableClientIDs = new ClientIDs();
-        for (int i = 0; i < hw->mMaxUsers; i++) {
-            hw->mAvailableClientIDs->push_back(i);
-        }
-        hw->mClientIDdict = new ClientIDDict();
+        hw->mClientManager = new ClientManager(inOptions->mMaxLogins);
```
JordanHendersonMusic, line 335: Why new?

JordanHendersonMusic (commented):
Just some thoughts on the C++ side of things.

I don't know too much about the architecture, but it seems the lifetimes of world and hidden world could also do with a rewrite using smart ptrs.

server/scsynth/SC_ClientManager.h
capital-G: Will probably ignore it, but can you provide some insight why this should be done?

server/scsynth/SC_ClientManager.h
capital-G: tend to ignore it, but can you provide some insights on `[[nodiscord]]`? I thought this was there to give the compiler some hints for optimizations?

server/scsynth/SC_ClientManager.h
capital-G: Not sure, but when I use `unordered_map` I obtain  `call to implicitly-deleted default constructor of 'std::hash<ReplyAddress>'`

server/scsynth/SC_ClientManager.h
capital-G: scsynth follows `.h` and `.cpp` - should we still deviate from it in that case?

server/scsynth/SC_ComPort.cpp
capital-G: Should we do this here? I think the PR is already big enough and I don't want to go into the rabbit hole of all scsynth message functions ;)

server/scsynth/SC_SequencedCommand.cpp
capital-G: agreed, but this is copy pasted from the previous implementation. I think it would be cleaner to go over all the c-style casts in messaging w/ another PR.

server/scsynth/SC_HiddenWorld.h
capital-G: Probably yes, but I think we should aim for consistency w/ the existing code base here. Moving the members behind smart pointers would be a PR on its own.

One thing: We could make `HiddenWorld` a class (to get RAII), but what would be the motivation of making it a pointer in the first place? Is it more to improve upon caching and keep the (potentially) passed around object as small as possible for the cache?

server/scsynth/SC_World.cpp
capital-G: I actually don't know, see https://github.com/supercollider/supercollider/pull/7676#discussion_r3706626869 for a possible why?

capital-G:
Thx for the review Jordan, much appreciated!

include/common/SC_Reply.h
```diff
@@ -22,4 +22,4 @@
 #pragma once
 
 struct ReplyAddress;
-typedef void (*ReplyFunc)(struct ReplyAddress* inReplyAddr, char* inBuf, int inSize);
+typedef void (*ReplyFunc)(const ReplyAddress* inReplyAddr, char* inBuf, int inSize);
```
capital-G, line 25: @Spacechild1 do you think this is okay?
I think it is semantically correct, though it could break the plugin ABI (probably no problem currently) but also plugin code?

server/scsynth/SC_ClientManager.h
JordanHendersonMusic: Not the compiler. It generates a warning if the user doesn't assign it to a variable. It's really important if you return an error code that the call sight needs to handle it. It is a mistake if a user calls a cosnt method and doesn't do something with the returned value, so it should be the default. It doesn't change behaviour at all.

server/scsynth/SC_HiddenWorld.h
JordanHendersonMusic: > Probably yes, but I think we should aim for consistency w/ the existing code base here.

I worry this just adds to the problem though.

capital-G: IMO it is best to keep consistency, even if it means to follow some not best practices. This allows for easier refactoring and creates less mental overhead until the fix is applied b/c there is one way of doing things (instead of 2, and at some points, possibly 3^^).

server/scsynth/SC_ClientManager.h
JordanHendersonMusic: It stops you doing these casting...
```
// given...
uint32 f() { return 1; }

ClientManager c = f();

void g(ClientManager cm) { ... }
g(1);
``` 
Instead you must be explicit and call its constructor.
```
ClientManager c { f() };
g({1});
```

server/scsynth/SC_HiddenWorld.h
JordanHendersonMusic: > This allows for easier refactoring

unique ptr follows raii. You just copy it wherever. This current approach requires very careful reading of how the resource is managed. 

smart pointer are specifically designed so you don't have to think about the memory as much.

The current approach is an increase in (more of the same) mental overhead.

include/common/SC_Reply.h
Spacechild1: `const` does not break the *ABI* (= application binary interface) because it's only meaningful to the compiler. However, it can break source compatibility (= the API). Fortunately, `ReplyFunc` is never used by plugins.

`ReplyFunc` can be used by `libscsynth` clients, but since we currently don't make any API/ABI guarantees about `scsynth` (and I want do a refactor in the near future), this is not an issue.

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,109 @@
...
+    void setPassword(const std::optional<const std::string>& password) { mPassword = password; }
+
+    /// returns maximum number of allowed users for this server
+    uint32 getMaxUsers() const { return mMaxUsers; }
+
+    /**
+     * @brief Returns the dict of all registered clients and their ID.
+     * Only access this in (N)RT locked environments!
+     */
+    const std::map<ReplyAddress, ClientID>& getClients() const { return mClientDict; }
```
Spacechild1, line 52: I would use a typedef for `std::map<ReplyAddress, ClientID>`, e.g. `ClientMap`.

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,109 @@
...
+        return id;
+    }
+
+private:
+    bool isClientIDTaken(ClientID id) const {
+        for (const auto& [addr, clientID] : mClientDict) {
+            if (clientID == id)
+                return true;
+        }
+        return false;
```
Spacechild1, line 95: I mean, this *works*, but IMO it would be cleaner to keep the assigned client IDs in a `std::unordered_set`.

Currently, the loop in `getNextClientID` has quadratic complexity. I know, it's not performance critical code in any way, but it feels a bit sloppy.

server/scsynth/SC_ClientManager.h
```diff
@@ -0,0 +1,109 @@
...
+struct World;
+
+using ClientID = uint32;
+
+/**
+ * @brief Manages client access of the server such as password check, client id issuing and handling tcp disconnects.
+ */
+class ClientManager {
+    /// since this gets accessed in the real time thread, this should only be accessed via
+    /// in protected environments such as stage1, stage2, ... or in the real time thread.
```
Spacechild1, line 18: > /// since this gets accessed in the real time thread, this should only be accessed via

`mClientDict` should *not* be accessed in the realtime thread!

In general, it is accessed in the NRT thread, but there is one exception, `GraphDef_DeleteMsg`, which looks like a mistake (not by you!)

> in protected environments such as stage1, stage2, ... or in the real time thread.

This doesn't make any sense. `stage1` is the RT thread, `stage2` is the NRT thread. Both can't be true.

server/scsynth/SC_ClientManager.cpp
```diff
@@ -0,0 +1,21 @@
...
+    world->hw->mClientManager->removeClient(*address);
+
+    delete address;
+}
+
+void ClientManager::removeClientDefer(World* world, const ReplyAddress& client) {
+    FifoMsg msg;
+    msg.Set(world, removeClientCallback, nullptr, new ReplyAddress(client));
+    AudioDriver(world)->SendMsgFromEngine(msg);
+}
```
Spacechild1, line 21: I really wouldn't put this in `ClientManager`. Deferring should be done on the callsite.

Also, in the comment you're saying
> This can be called from any thread (though not RT thread!)
but this completely backwards. `SendMsgFromEngine` is used to send a message *from* the RT thread to the NRT thread.

AFAICT this is only called in a single spot, namely `~SC_TcpConnection`. Since a `SC_TcpConnection` is destroyed on the network (ASIO) thread, we can't call `SendMsgFromEngine`. In fact, the only thing we have to do is lock the NRT thread and call `removeClient()`.

server/scsynth/SC_ComPort.cpp
```diff
@@ -477,7 +490,15 @@ class SC_TcpInPort {
     }
 };
 
-SC_TcpConnection::~SC_TcpConnection() { mParent->connectionDestroyed(); }
+SC_TcpConnection::~SC_TcpConnection() {
+    ClientManager::removeClientDefer(mWorld, mClientIdentification);
```
Spacechild1, line 494: `SC_TcpConnection` is destroyed on the network (ASIO) thread. There is no need to defer anything! Instead just lock the NRT thread and directly call `removeClient`. See also https://github.com/supercollider/supercollider/pull/7676/changes#r4066429458.

server/scsynth/SC_GraphDef.cpp
```diff
@@ -429,7 +429,7 @@ SCErr GraphDef_DeleteMsg(World* inWorld, GraphDef* inDef) {
     packet.addtag('s');
     packet.adds((char*)inDef->mNodeDef.mName);
 
-    for (auto addr : *inWorld->hw->mUsers) {
+    for (const auto& [addr, clientID] : inWorld->hw->mClientManager->getClients()) {
```
Spacechild1, line 432: This is actually a mistake, the client map must only be accessed on the NRT thread (e.g. stage2 of async commands)!

The solution is simple: just move this whole part to `DeleteGraphDefMsg::Perform()` (You'll have to add a `World* mWorld` member to `DeleteGraphDefMsg`). Since `Perform()` is called on the NRT thread, you can get rid of `SendReplyCmd_d_removed` and directly call `SendReply()`.

server/scsynth/SC_HiddenWorld.h
Spacechild1: This is fine. `World` and `HiddenWorld` are currently directly malloc'd and freed, i.e. no C++ constructors or destructors are called, so `unique_ptr` wouldn't work.

That being said, I'm currently working on a big refactoring of scsynth internals. One of my goals is to replace all these raw pointers with values or smart pointers.

This PR can follow the existing pattern and I'll later do the cleanup.

Spacechild1 (changes requested):

PR #7689, opened 2026-08-14
classlib: 'delete' the newCopyArgs method on QObject preventing a segfaut

### LINKED ISSUE #7685: Segfault when calling layout_ on subclass of View
## Environment

* SuperCollider version: 3.14.1
* Operating system: Linux (Debian 13)
  QT Version:  6.8.2

## Steps to reproduce
I have this simple class which defines a subclass of `View`.  
```supercollider
// this version fails - causing a segfault on creating a new instance
C : View {
	var <a;

	*new  { |a|
		^super.newCopyArgs(a: a).init
	}

	init {
		this.layout_(VLayout());
	}
}

// works
C : View {
	var <a;

	*new  { |a|
		^super.new.init(a)
	}
		
	init { |key|
		a = key;
		this.layout_(VLayout());
	}
}
```

Version 1 of the class above shouldn't segfault but throw a `PrimitiveFailedError` (if there's something wrong within code logic)

JordanHendersonMusic:
## Purpose and Motivation

Closes #7685

Initialising a subclass of qobject without calling the 'new' method will result in a segfault.

## Types of changes

- Bug fix

## To-do list

- [ ] Code is tested
- [ ] All tests are passing
- [ ] Updated documentation

### DIFF
--- a/SCClassLibrary/Common/GUI/Base/QObject.sc
+++ b/SCClassLibrary/Common/GUI/Base/QObject.sc
@@ -48,6 +48,10 @@ QObject {
 		^super.new.initQObject( className, argumentArray );
 	}
 
+	*newCopyArgs { |...args, kwargs| 
+		Error("Please replace call to 'newCopyArgs' with a call to 'new'. All QObjects must call the 'new' method.").throw
+	}
+
 	*heap { ^heap.copy }
 
 	initQObject{ arg className, argumentArray;


JordanHendersonMusic:
@nuss What do you think of this? Actually getting a PrimitiveError (as you suggested) doesn't seem like the simplest thing to do.

nuss:
@JordanHendersonMusic - hmm... hoenstly,I don't know. I now (kindof) understand why using `newCopyArgs` isn't the best choice when subclassing `View` (maybe `superPerformArgs` as described in the [writing classes guide](https://docs.supercollider.online/Guides/WritingClasses.html#Instance%20Variables%20and%20Initialisation%20Logic) would be a better choice?).  The [`SCViewHolder` helpfile](https://docs.supercollider.online/Classes/SCViewHolder.html) explicitly mentions that subclassing a view class isn't always the best choice - I guess I found one example why (the `View` helpfile, however, doesn't mention anything about that, unfortunately). I thought there's something wrong about the `_QWidget_SetLayout` primitive that gets called internally by `layout_` but, well, if that's the only reason why `layout_` fails? I haven't come across this problem in any other situation...

JordanHendersonMusic:
> maybe superPerformArgs...

Yes that can work, but it involves rewriting all the class hierarchy and will probably be a breaking change.

> thought there's something wrong about the _QWidget_SetLayout primitive that gets called internally by layout_ but, well, if that's the only reason why layout_ fails? 

The new method registers the QObject with some structure internal to Qt collider. The lookup happens when calling layout (or anything qt related) and then crashes because it doesn't exist. The only way to register the object is to call the constructor (well, prConstruct but it's private), so it just isn't possible to use newCopyArgs here.

> I haven't come across this problem in any other situation

I have unfortunately! And I've ran into accidentally overriding the init method. Constructors need a rethink, probably with some new syntax... Something like...
```
B{
   var bar;
   B(bar) : bar(bar.postln) { $B.postln }
}
A : B {
   var foo;
   A(foo, bar) : foo(foo.postln), super(bar) { $A.postln }
} 

A('foo', 'bar')
```
Which then prints `foo, bar, B, A`, so up when doing var assignments, and down the hierarchy when evaluating methods. This is what Dart does.

nuss:
Slowly it feels like I've opened a can of worms...

JordanHendersonMusic:
Sorry, yeah, I think it's pretty bad there is this much complexity in simply making an object... Anyway, I think this pr is probably the best we can do right now (or something variant of it).

nuss:
Not your fault... thanks (and sorry) for taking your time to explain the problem and effort to provide a solution. Honestly I'm lacking the expertise to assess a solution for such a complex problem. Personally I don't think removing `newCopyArgs` from `QObject` wouldn't harm my code but I don't know who else might be affected.

JordanHendersonMusic:
Sure, it will effect others, but there code is highly likely to segfault and ruin their performance. This kind of breaking changes can be painful, but I think are always worth it.

capital-G (approved):

